PR #2416, opened 2017-07-05
Adding a Fronius Solar power metrics plugin

BrainDoctor:
Produces charts such as this:
fronius_charts

PR #2448, opened 2017-07-12
add plugin for Stiebel Eltron heat pump system

BrainDoctor:
My next plugin ;)

1. Node.js plugin
2. Readme page section in node.d/readme.md
3. Default configuration in conf.d/node.d/stiebeleltron.conf.md
4. Updated infographic. Had to move the lowest section down a bit, to provide space for this and future plugins
5. I didn't update the "Add charts" wiki yet. Should I split the "UPS and power" section into "UPS" and "Household appliances"? Or just a new section? Which links would get broken? I'm thinking of moving the solar power plugins and this one to "household", which would make more sense to me.

I'm aware that the plugin is collecting values based on a rather large JSON configuration (definition of charts, dimensions etc.). I decided to do it in this way so that other people are able to collect different values in their systems. I cannot provide a plugin that is working for every setup, since I cannot test them (there are air/water systems, geothermal probes, etc). But I can make it adaptive.

Also the metrics are collected using RegEx. Not optimal but there is no direct API (AFAIK) and performance-wise they are just fine (some 0.9% spikes on a Intel Xeon). The slowest component is the device response itself (usually around 1s, sometimes it takes up to 4s).


ktsaou:
Nice!

> I'm thinking of moving the solar power plugins and this one to "household"

I like that too.

I haven't reviewed it yet. Give me some time.

One note: for the images you add to the infographic, please copy and paste the image to a comment like this one. The problem is that third party sites update their images and as time passes the infographic breaks and needs maintenance. If you just paste the image to github and use the github URL at the infographic, our interactive page (https://my-netdata.io/infographic.html) uses the github CDN for image delivery which is usually very fast and available.

BrainDoctor:
You mean pasting the logo here:
logo-stiebel-eltron
and use this link?

BrainDoctor:
I updated the infographic with the stiebel logo from here and the fronius from PR #2416 in the same commit. I know, I was too lazy to create a new PR ;)

ktsaou:
I am ready to merge this.
Do you agree?

ktsaou:
this one?

BrainDoctor:
Yes, unless you want to wait for the refactor

ktsaou:
Your call... whatever you like...

BrainDoctor:
ok lets merge this, so we have at least a working version. Have you decided what to do about the wiki pages? household category etc.

ktsaou:
household seems good...

ktsaou:
merged it! thank you!

PR #2459, opened 2017-07-13
adapted prometheus formatting

ktsaou:
https://github.com/firehol/netdata/pull/2436#issuecomment-315002915

I did some research to check if `as collected` could possibly use the format of `average`. I found out that for most of the cases, even if we collect metrics from multiple sources, the metrics that are added to the same chart happen to have exactly the same algorithm, multiplier and divider.

I found only 3 charts that this is not true. A redis one, and 2 netdata internal ones (for monitoring netdata itself).

Also, I checked if netdata autodetection incrementally adds dimensions to charts (so that adding an incompatible dimension will suddenly make the chart dimensions heterogeneous). I found that most dynamic additions to a chart, are controlled by the exact same source, which is totally unlikely to change the algorithm, multiplier or divider.

The key problem remains statsd synthetic charts, that are user controlled.

So, I added some code that formats homogeneous charts in `as collected` mode with the format `average` uses. This is still dynamic, meaning that netdata will automatically switch format if the chart becomes heterogeneous, but I think the probability of this happening is close to zero.

Also, metrics are named `*_total` for counters, `*_average` for averages and `*_sum` for sums.

Finally, for `average` mode, the units are in the metrics name.

### DIFF
--- a/src/backend_prometheus.c
+++ b/src/backend_prometheus.c
@@ -67,6 +67,38 @@ static inline size_t prometheus_label_copy(char *d, const char *s, size_t usable
     return n;
 }
 
+static inline char *prometheus_units_copy(char *d, const char *s, size_t usable) {
+    const char *sorig = s;
+    char *ret = d;
+    size_t n;
+
+    *d++ = '_';
+    for(n = 1; *s && n < usable ; d++, s++, n++) {
+        register char c = *s;
+
+        if(!isalnum(c)) *d = '_';
+        else *d = c;
+    }
+
+    if(n == 2 && sorig[0] == '%') {
+        n = 0;
+        d = ret;
+        s = "_percent";
+        for( ; *s && n < usable ; n++) *d++ = *s++;
+    }
+    else if(n > 3 && sorig[n-3] == '/' && sorig[n-2] == 's') {
+        n = n - 2;
+        d -= 2;
+        s = "_persec";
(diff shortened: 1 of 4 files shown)


ktsaou:
@candlerb I would appreciate a test to make sure we haven't broken anything.

candlerb:
LGTM on brief test.

~~~
  - job_name: 'netdata-scrape'

    metrics_path: '/api/v1/allmetrics'
    params:
      format: [prometheus]
      source: [as-collected]
      prefix: [raw]

    static_configs:
      - targets: ['10.12.254.116:19999']

  - job_name: 'netdata-scrape-normalized'

    metrics_path: '/api/v1/allmetrics'
    params:
      format: [prometheus]
      prefix: [norm]

    static_configs:
      - targets: ['10.12.254.116:19999']
~~~

I can compare, for example, `irate(raw_disk_io_total[2m])/2` with `norm_disk_io_kilobytes_persec_average`.  I get the same number of time series on each.  The graphs are in rough agreement.

ktsaou:
thanks!
My only concern is if there is a limit on the length of the metric name at prometheus. With the addition of units the metric names are long.
Do you know?

ktsaou:
coverity scan is happy too with this commit.

ktsaou:
merged!

PR #2478, opened 2017-07-18
Fix script to build release correctly on OSX

philwhineray:
The sed command used to ignore build artifacts when checking for
unpackaged files was using expressions not supported by BSD sed.

Replaced them with simpler working versions and removed a couple of
erroneous copy-paste lines from the firehol/firehol repo.


ktsaou:
nice!

PR #2518, opened 2017-07-25
Fix parsing of mdstat output

JeffHenson:
When an operation is in progress, there could be two spaces before the
percentage when it's less than 10%.


ktsaou:
thank you! I don't use this plugin, so I have requested @l2isbad is check.

l2isbad:
@ktsaou all is 👍 

thx @JeffHenson

ktsaou:
perfect! merged!

PR #2525, opened 2017-07-27
Improve tomcat chart

Wing924:
Improvement:
- Support Apache+AJP Protocol (reverse proxy).
- Add detailed memory usage.
- Add processing_time.

new charts:


python.d/tomcat.chart.py
```diff
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 # Description: tomcat netdata python.d module
 # Author: Pawel Krupa (paulfantom)
+# Update author: Wing924
 
 from base import UrlService
 from re import compile
```
l2isbad: re package not needed

python.d/tomcat.chart.py
```diff
@@ -1,9 +1,11 @@
...
+import xml.etree.ElementTree as ET
 
 try:
     from urlparse import urlparse
```
l2isbad: urlparse not needed

python.d/tomcat.chart.py
```diff
@@ -16,61 +18,141 @@
...
-        netloc = urlparse(self.url).netloc.rpartition(':')
-        if netloc[1] == ':': port = netloc[2]
-        else: port = 80
-        
-        self.regex_jvm = compile(r'<jvm>.*?</jvm>')
-        self.regex_connector = compile(r'[a-z-]+%s.*?/connector' % port)
-        self.regex = compile(r'([\w]+)=\\?[\'\"](\d+)\\?[\'\"]')
-        
-        return UrlService.check(self)
+        return UrlService.check(self) and self._get_data() != None
```
l2isbad: check method not needed (looks like the [UrlService check](https://github.com/firehol/netdata/blob/master/python.d/python_modules/base.py#L568-L591) is ok for this plugin)

python.d/tomcat.chart.py
```diff
@@ -16,61 +18,141 @@
...
+                    if conn.get('name') == self.connector_name:
+                        connector = conn
+                        break
+            else:
+                connector = xml.find('connector')
+
+            memory = jvm.find('memory')
+            data['free']  = memory.get('free')
+            data['total'] = memory.get('total')
+            data['used']  = int(data['total']) - int(data['free'])
```
l2isbad: Not sure about this. Dict `get` returns the result or `None`. In case of `None` `int(None)` will raise the Exception.

python.d/tomcat.chart.py
```diff
@@ -16,61 +18,141 @@
...
+            ["tenured_max", 'max', "absolute", 1, 1048576]
+        ]},
 }
 
-
 class Service(UrlService):
     def __init__(self, configuration=None, name=None):
         UrlService.__init__(self, configuration=configuration, name=name)
         self.url = self.configuration.get('url', "http://127.0.0.1:8080/manager/status?XML=true")
+        self.connector_name = self.configuration.get('connector_name', None)
```
l2isbad: You have to update the module configuration file also

python.d/tomcat.chart.py
Wing924: It also check if status page is valid :)

python.d/tomcat.chart.py
Wing924: I added documentation to conf file.

python.d/tomcat.chart.py
Wing924: I found `data['used']` is not be used. So I simply delete this line.

l2isbad:
@Wing924 
`urlparse` import is not needed

Wing924:
@l2isbad I removed unused imports.

l2isbad:
looks ok for me @ktsaou

python.d/tomcat.chart.py
```diff
@@ -3,74 +3,145 @@
...
-        if data:
-            jvm = self.regex_jvm.findall(data) or ['']
-            connector = self.regex_connector.findall(data) or ['']
-            data = dict(self.regex.findall(''.join([jvm[0], connector[0]])))
-        
-        return data or None
+        data = None
+        raw_data = self._get_raw_data()
+        if raw_data:
+            xml = ET.fromstring(raw_data)
```
l2isbad: this one will fail if data is not xml (`url` without `?XML=true`). So i suggest to wrap it in `try/except` (`ET.ParseError`) and add a debug info to warn the user we need xml (`self.debug('TEXT')`)

l2isbad: i mean `xml = ET.fromstring(raw_data)` part

Wing924: I've fixed it.

Wing924:
@l2isbad When do you plan to merge this? :)

l2isbad:
@ktsaou 
^^

ktsaou:
oops! I missed this!
Sorry!

Issue #2446, opened 2017-07-12
Enable keepalive on UrlService


Wing924:
UrlService open http connection every seconds.
Because keepalive is off, TCP sockets are not reused.
It cause so many TIME_WAIT sockets.

I think make keepalive on can improve performance.

ktsaou:
@l2isbad have a look please.

l2isbad:
I am not sure this can be implemented using urllib2 w/o additional packages or something.

ktsaou:
hm... this is a big problem though. I hadn't realized we work like that...
I read about urllib3 and other alternatives.
Why do they bother you?

l2isbad:
Well, the main goal was to make it work on all installs out from the box without additional installing i think. urllib3 is not default package on debian <= 7

ktsaou:
hm... yes we have to support older systems.

Can we use urllib3 only when it is available?
Can we ship urllib3 with netdata? (if it is python-only without any dependencies, this could be easy)

l2isbad:
I like the idea to switch to the urllib3

ktsaou:
ok, I mark as enhancement, assigned to you.
Thanks!

ktsaou:
merged it

Issue #2174, opened 2017-05-10
Count TCP connections per port


povesteam:
Is it possible to extract number of connections per listening port?

For example, something similar with this graph, but with 4 lines (labeled 22, 80, 443, 3000) if there are processes listening for ssh, http, https and one node.

![image](https://cloud.githubusercontent.com/assets/506532/25898935/aba071c4-3596-11e7-88dc-43585b7dc649.png)

ktsaou:
Hi.

Unfortunately no. But there is something similar (and probably more useful):

![image](https://cloud.githubusercontent.com/assets/2662304/25917017/624a6064-35cf-11e7-8db4-1659adbed343.png)

This chart shows the number of sockets open per application group. There are also similar charts for Users and Groups.

You can control the application groups, by editing `apps_groups.conf`.

ktsaou:
The reason for not being able to do it per port, is that netdata will have read all the system sockets, which as you may understand there may be several thousands of them.

This is a CPU intensive operation I am trying to avoid. Though, this is a quite common request, so probably at some point we will have to do it, with the ability to enable or disable it as needed.

ktsaou:
Also, keep in mind the chart I posted counts also the listening sockets.

povesteam:
Hi @ktsaou, thank you for explaining and providing alternative.
I will try to configure `apps_groups.conf` and come back here with the result.

povesteam:
It works when I add `node: node` to `apps_groups.conf`, but it counts all connections to those processes, including listen, incoming and outgoing.

I would want to count only established connections on the listening port.
In this example, I would want to see **4** connections for port 4000 and **7** connections for port 4001 (even the total of **11** would be ok).

Current behaviour includes outgoing database connections and the listening socket that you told me about, which gives a total of **16**.

```
tcp        0      0 0.0.0.0:4000            0.0.0.0:*               LISTEN      8852/node
tcp        0      0 127.0.0.1:4000          127.0.0.1:36320         ESTABLISHED 8852/node
tcp        0      0 127.0.0.1:4000          127.0.0.1:48075         ESTABLISHED 8852/node
tcp        0      0 127.0.0.1:4000          127.0.0.1:33103         ESTABLISHED 8852/node
tcp        0      0 127.0.0.1:4000          127.0.0.1:36921         ESTABLISHED 8852/node
tcp        0      0 127.0.0.1:4001          127.0.0.1:56316         ESTABLISHED 10744/node
[...]

PR #2528, opened 2017-07-28
new chart: netstat

Wing924:
It collect tcp connection state from netstat command.


python.d/netstat.chart.py
```diff
@@ -0,0 +1,54 @@
...
+        raw_data = self._get_raw_data()
+        if not raw_data:
+            return None
+        ptn = re.compile(r'\s+')
+        data = {}
+        for key in ['ESTABLISHED', 'SYN_SENT', 'SYN_RECV', 'FIN_WAIT1', 'FIN_WAIT2', 'TIME_WAIT', 'CLOSE', 'CLOSE_WAIT', 'LISTEN', 'CLOSING']:
+            data[key] = 0
+
+        for line in raw_data:
+            tk = ptn.split(line)
```
l2isbad: Do we really need re package here? Looks like simple line.split() will do the work. Can you check it?

l2isbad:
@Wing924 check codacy error

python.d/netstat.chart.py
Wing924: @l2isbad 
This is sample output form netstat:
```
tcp4       0      0  10.49.49.61.55709      40.100.155.2.443       ESTABLISHED
tcp4       0      0  10.49.49.61.55708      40.100.155.2.443       ESTABLISHED
tcp4      31      0  10.49.49.61.55679      192.30.255.113.443     CLOSE_WAIT
```
Results contain many spaces. Can you tell me how to split without regex?

l2isbad:
```python
'hello world            and              goodbuy            !'.split()
['hello', 'world', 'and', 'goodbuy', '!']
```

python.d/netstat.chart.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+# Description: example netdata python.d module
+# Author: Pawel Krupa (paulfantom)
```
l2isbad: really? 😄

l2isbad:
thx @Wing924 
So it's only works for ipv4.

I am not an expert but looks like we can extract the same info from `/proc/net/tcp` which is much faster.

`man 5 proc`
```
       /proc/net/tcp
              Holds a dump of the TCP socket table.  Much of the information is not of use apart from debugging.   The  "sl"  value  is  the  kernel  hash  slot  for  the  socket,  the
              "local_address"  is  the local address and port number pair.  The "rem_address" is the remote address and port number pair (if connected).  "St" is the internal status of
              the socket.  The "tx_queue" and "rx_queue" are the outgoing and incoming data queue in terms of kernel memory usage.  The "tr",  "tm->when",  and  "rexmits"  fields  hold
              internal information of the kernel socket state and are useful only for debugging.  The "uid" field holds the effective UID of the creator of the socket
```

[St value](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/net/tcp_states.h?id=HEAD) (internal status of the socket)

@ktsaou ?

ktsaou:
#2174

I will try to come up with a plugin to do this in C.
I haven't decided it yet how it should be done though:

1. A plugin inside netdata daemon
2. A feature of apps.plugin
3. something completely new

Ideally, I would like to have this:

1. count sockets globally (like this plugin)
2. count sockets per application (in apps.plugin, but the presentation is somewhat complex - we will need a chart per status with all the applications as dimensions)
3. count sockets of one or more interesting ports (i.e. generate a chart per port?)

For sure, we need an **application watchdog** in netdata. Something that will monitor everything about an application or a pid.

I think we should do this in `apps.plugin`... this will also allow us to have sockets per User and sockets per User Group.

What do you think?

Wing924:
@ktsaou 
>I think we should do this in apps.plugin

I agree with it because it's the fastest way to do so and can provide more detailed info.

> count sockets globally (like this plugin)

I use this plugin and found many bugs in my system such as #2446. I think it is also important for other users.

> count sockets per application (in apps.plugin, but the presentation is somewhat complex - we will need a chart per status with all the applications as dimensions)

I think count established sockets and create 2 charts: connect from and connect to is useful and simple.
If destination address is self, it is  connect to, otherwise connect from.

> count sockets of one or more interesting ports (i.e. generate a chart per port?)

I think generating a chart per listening port is useful.

Wing924:
@l2isbad I changed data source from netstat command to /proc/net/tcp. Can you review it?

l2isbad:
@Wing924 i think it will be implemented by @ktsaou in C

ktsaou:
Yes, I will. This is too heave to implemented in python.

Issue #2520, opened 2017-07-25
non-superuser postgres job not working if there is another superuser-job configured


BrainDoctor:
So I have this in `/etc/netdata/python.d/postgres.conf`
```yaml
socket:
    name     : 'local'
    user     : 'postgres'
    database : 'postgres'

jira:
    name     : 'Jira'
    database : 'jira'
    user     : 'jira'
    host     : 'jira.intern'
    port     : 5433
    password : '*****'
    database_poll : 'jira'

owncloud:
    name     : 'Owncloud'
    database : 'owncloud'
    user     : 'owncloud'
    host     : 'cloud.intern'
    port     : 5432
    password : '*****'
    database_poll : 'owncloud'

tcpipv4:
    name     : 'local'
    database : 'postgres'
    user     : 'postgres'
    host     : '127.0.0.1'
    port     : 5432

tcpipv6:
    name     : 'local'
    database : 'postgres'
    user     : 'postgres'
    host     : '::1'
    port     : 5432

```
The dashboard only shows Jira:
![2017-07-25_210908](https://user-images.githubusercontent.com/12159026/28589226-87adc4c4-717d-11e7-9508-5a0f522479e8.png)

error.log:
```
2017-07-25 20:55:13: python.d ERROR: postgres_socket Failed to connect to {'user': 'postgres', 'database': 'postgres'}. Error: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?

2017-07-25 20:55:13: python.d ERROR: postgres_socket check() failed - disabling job
2017-07-25 20:55:13: python.d ERROR: DISABLED: postgres/socket
2017-07-25 20:55:13: python.d INFO: CHECKED OK: postgres_jira
2017-07-25 20:55:13: python.d INFO: RENAMED: postgres_Jira , from postgres_jira
2017-07-25 20:55:13: python.d INFO: CHECKED OK: postgres_owncloud
2017-07-25 20:55:13: python.d INFO: RENAMED: postgres_Owncloud , from postgres_owncloud
2017-07-25 20:55:13: python.d ERROR: postgres_tcpipv4 Failed to connect to {'port': 5432, 'host': '127.0.0.1', 'user': 'postgres', 'database': 'postgres'}. Error: could not connect to server: Connection refused
        Is the server running on host "127.0.0.1" and accepting
        TCP/IP connections on port 5432?

[...]

PR #2538, opened 2017-07-30
postgres.chart.py: Fix superuser bug.

facetoe:
Resolve issue raised in #2520 that led to non superuser instances
attempting to execute queries that require superuser privs.


ktsaou:
thanks!

PR #2539, opened 2017-07-30
rename and move network interfaces to container section

ktsaou:
This PR renames and moves container `veth*` interfaces to the section of the container.

It works like this:

1. When a container is first encountered, a new external program (`cgroup-network`) is called to find the network interfaces that belong to this container.

2. The cgroups plugin then registers a network interface rename.

3. For all network interfaces of the system, a mechanism to rename the charts has been implemented. This mechanism is hooked when interfaces are first encountered and also before each data collection iteration (of course only when there is something to rename).

The result of the above is that `net.veth12345678` now becomes `cgroup_CONTAINER.net_eth0`. The same happens for all charts of the network interface.

The result is the following, for each container:

image

### about `cgroup-network`

`cgroup-network` is a little binary that can find the network interfaces of a cgroup (container) given a pid in it.
This is to be used for moving the `veth*` network interfaces under each container section of the dashboard.

Once installed, run:

```sh
/usr/libexec/netdata/plugins.d/cgroup-network -p PID
```

Example:

```sh
# ps fax | grep graphite | head -n 1 | cut -d ' ' -f 1
20773
# /usr/libexec/netdata/plugins.d/cgroup-network -p 20773
veth96bda88 eth0
```


ktsaou:
Now the question is: **what is the best way to move the network interfaces to the containers section?**

Options:

1. Just move the chart at the proper section, without changing its name. That is the network interface will continue to be called `net.veth96bda88`, but instead of it being rendered under the **Network Interfaces** section, it will be moved at the container section.

   **PROS**: it seems simple (just a tag in the C code and some javascript magic to move it).
   **CONS**: the metrics will be inconsistent across container restarts - it will be impossible to use it at backends.

2. Rename the chart on the fly, using the interface name as it appears inside the container. So `net.veth96bda88` will become `cgroup_graphite.net_eth0`.

   **PROS**: the best consistency as far as metric names are concerned
   **CONS**: renaming a chart does not exist yet. The interface will exist for a few seconds before it is renamed, so backends will see a few entries with the `veth*` name.

3. Duplicate the chart, so it will exist as both names.

   **PROS**: it will be available in both places at the same time
   **CONS**: Twice the memory and disk for the same metrics. I don't have a clue how to do this at the core of netdata...

@titpetric @l2isbad @agross what do you think?

ktsaou:
@titpetric it would also be interesting to see if `cgroup-network` can work (switch namespaces) when it runs in a privileged container.

ktsaou:
hm... there is another issue: when we move the interface to the container section, all the metrics should be reversed (ie input should be output, and output should input).

This probably means that quite a few changes have to be made to flip all the dimensions in the network charts...

agross:
Thanks for the opportunity to participate in the decision! I'm on vacation until Friday with little internet access, please don't expect my input earlier than that.

ktsaou:
I chose and implemented No 2 - network interfaces are renamed.

ktsaou:
Check it live at http://build.my-netdata.io

agross:
Looking good! 👍

vlvkobal:
@ktsaou you have added `netdev_rename_device_add()` and `netdev_rename_device_del()`. They are declared and defined in `plugin_proc.h` and `proc_net_dev.c` respectively. As long as `plugin_proc.h` and `proc_net_dev.c` are excluded from compilation on FreeBSD and macOS, netdata can not be compiled there at the moment. Is it OK to exclude `sys_fs_cgroup.c` too as long as cgrops are Linux specific?

ktsaou:
> Is it OK to exclude sys_fs_cgroup.c too as long as cgrops are Linux specific?

Oops! Sorry! Yes, of course.
Would you like to make a PR?

PR #2541, opened 2017-08-01
renaming fronius dimensions to lowercase.

BrainDoctor:
Most chart dimensions seem to be lower case. Renamed the dimensions to be more consistent with the existing naming convention.

Unit and system tested.


ktsaou:
perfect!

PR #2582, opened 2017-08-10
Add guide for accessing web dashboards

sangyun-han:
Users can't find how to access web dashboard before check default port using 'netdata -h' 
If there is a guide for access dashboard in READM, it is good to user.

Issues #2592


ktsaou:
thanks!

PR #2588, opened 2017-08-11
Reduce notification e-mail spam score.

### LINKED ISSUE #2587: Improve spam score of netdata email notifications
In some cases netdata notification emails have to go through some antispam (even though it is probably a good thing to have them only run in some internal network, or whitelist these kinds of emails, in some cases this is not possible). 

According to rspamd I get the following : 

```
X-Spamd-Result: default: False [5.00 / 30.00]
 MIME_HEADER_CTYPE_ONLY(2.00)
 FORGED_SENDER(0.30)
 MIME_HTML_ONLY(0.20)
 R_MISSING_CHARSET(2.50)
```

We can't do anything about FORGED_SENDER, but I guess the others could be fixed.

Ferroin:
This adds Content-Disposition and Content-Transfer-Encoding headers to the HTML and text/plain e-mail bodies, thus eliminating one of the issues that tools like Spamassassin check for.

The value `inline` for the Content-Disposition is largely ignored by most e-mail clients these days, but may cause some older clients to display both parts of the message.  It's usage here is consistent with what most e-mail clients set when sending both HTML and text/plain components.

The value `8bit` for the Content-Type-Encoding clarifies that there is unescaped binary data in the text, but that it is limited to 998 character lines and only uses CR and LF at the end of the lines.  `7bit` would work also when dealing with sending the messages in English, and technically be more widely compatible, but we can't assume that the message content will not contain 8-bit characters because certain components (specifically, at least the host name and chart name) may contain Unicode codepoints above 127.

Ideally, we also need to add `charset=` parameters to the Content-Type header for both components, but that requires a bit more work because it's non-trivial to translate the locale's character set to an IANA character set.

Partially fixes #2587

### DIFF
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -105,6 +105,13 @@ SEND_EMAIL="YES"
 DEFAULT_RECIPIENT_EMAIL="root"
 # to receive only critical alarms, set it to "root|critical"
 
+# Optionally specify the encoding to list in the Content-Type header.
+# This doesn't change what encoding the e-mail is sent with, just what
+# the headers say it was encoded as.
+# This shouldn't need to be changed as it will almost always be
+# autodetected from the environment.
+#EMAIL_CHARSET="UTF-8"
+
 
 #------------------------------------------------------------------------------
 # pushover (pushover.net) global notification options
--- a/plugins.d/alarm-notify.sh
+++ b/plugins.d/alarm-notify.sh
@@ -1402,6 +1402,8 @@ This is a MIME-encoded multipart message
 
 --multipart-boundary
 Content-Type: text/plain
+Content-Disposition: inline
+Content-Transfer-Encoding: 8bit
 
 ${host} ${status_message}
 
@@ -1418,6 +1420,8 @@ Notification generated on ${this_host}
 
 --multipart-boundary
 Content-Type: text/html
+Content-Disposition: inline
+Content-Transfer-Encoding: 8bit
 
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" style="font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; box-sizing: border-box; font-size: 14px; margin: 0; padding: 0;">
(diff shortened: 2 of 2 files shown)


ktsaou:
Thanks!

I think charset can be set to iso8859-1 or anything that has the ASCII set in it. netdata is not unicode aware and currently everything is in English. So, iso8859-1 is a good choice.

Do you want to add this too before merging it?

ktsaou:
Regarding the charset, we could also add an option to `health_alarm_notify.conf` so that users could set their own if the internal default is not good enough for them.

As I see it, only the hostname could be problematic (since we do not control it).

ivanaxel:
Thanks!

I think charset can be set to iso8859-1 or anything that has the ASCII set in it. netdata is not unicode aware and currently everything is in English. So, iso8859-1 is a good choice.

Do you want to add this too before merging it?

ktsaou:
yes, please do.

Ferroin:
I'll be happy to add it, but still I'm not 100% certain that it's correct.  While netdata itself is not unicode aware, it's not netdata that's composing the message, it's the shell that's running the script, and that will almost always be unicode aware.

I also contend that the chart name may not be ASCII either.  At minimum, I think it's possible for the fping plugin to produce chart names that aren't ASCII if it's told to check hosts with non-punycode encoded hostnames.

Ferroin:
OK, just did a bit more digging.  I hadn't realized that MIME charset names are case insensitive (and I feel like a bit of an idiot because of this, seeing as I already knew that the MIME type is case insensitive), which, combined with verifying that the glibc, musl, and uClibc locale definitions use MIME charset names, means it's actually trivial to pull this information out of the environment.  I'll look at getting that added in tomorrow with a commit message clarifying the discrepancy in this commit message.  I'll also see about having a way to override the charset from `health_alarm_notify.conf`.

Additionally, just to clarify my earlier comment, we _should_ be find just making sure we match the character set we claim in the MIME headers to the system locale setting.  For both the hostname, and chart names, netdata appears to do no special handling for a particular encoding, which in turn means that the values will already be encoded to match the system local.

Ferroin:
OK, just pushed out a commit with the encoding related changes.  That part has only been very lightly tested, as I don't really have the infrastructure to verify it.

Ferroin:
Forgot to mention, with this new commit added, this should completely fix #2587.

ktsaou:
Thanks!
I can't verify it either. What do you think? merge it?
I think we can merge it if the emails are delivered (and I don't see anything that could break it). Could you please test basic email delivery works (to gmail for example)?

Ferroin:
I've since double checked on one of my systems that throws a warning every night (nightly backups tend to send the disk backlog into the 6+ digit range), and it appears to work.

I'd be more worried with how e-mail clients handle it though than MTA's, this can all impact how the client renders it.  On that note, Thunderbird, GMail, and Mutt display no apparent differences in handling with these changes.

ktsaou:
Ok, we can merge it then. Presentation is something we can fix later. What should never happen is prevent alarms from being sent...

PR #2596, opened 2017-08-14
Exclude sys_fs_cgroup.c from compilation on FreeBSD and macOS

vlvkobal:
Fixes `netdata` compilation issue on FreeBSD and macOS after #2539


ktsaou:
thanks!

PR #2599, opened 2017-08-14
allow HTTP/2 for custom_log_format in web_log

Funzinator:
some web servers log HTTP2 requests as `HTTP/2` instead of `HTTP/2.0`. When using a custom log
format, requests with HTTP/2 can't be matched using the default dictionary.

This commit makes the `.0` part optional for custom logs


ktsaou:
merged it.
thanks!

@l2isbad I didn't requested a review. This seems pretty simple.

Issue #2544, opened 2017-08-01
How to read temperature without sensors in FreeBSD?


theAlchemist-dev:
Hello,

I am using your amazing piece of software on my `FreeBSD 10.3` server.
When reading the documentation, the sensor data should be read with `sensors`?

FreeBSD has no `sensors` command IMO, but I do have `kldload coretemp` with readout by `sysctl`:

`# kldload coretemp`
`# sysctl -a |grep temper`
`dev.cpu.3.temperature: 35.0C`
`dev.cpu.2.temperature: 35.0C`
`dev.cpu.1.temperature: 30.0C`
`dev.cpu.0.temperature: 30.0C`

Am I missing some configuration?
The CPU frequencies are being read, so the required `Python cpufreq` module is working, right?

Edit: `netdata version: 1.7`

ktsaou:
hm... I think is this not supported out of the box.

@vlvkobal is that something we can support on FreeBSD?

@theAlchemist-dev you can create a simple bash plugin to execute these commands and turn the output to something netdata can accept. Use this file as a starting point: https://github.com/firehol/netdata/blob/master/charts.d/example.chart.sh

vlvkobal:
I'm on the road for a few days more. I'll be able to look at it in a week.

vlvkobal:
@theAlchemist-dev I can't get CPU temperature on VM. Thus, I had to write it blindly on the basis of `man`. Could you try a code from `git clone --depth=1 --branch coretemp https://github.com/vlvkobal/netdata.git`. Is it working?

theAlchemist-dev:
Thanks for the help, I did not had time yet to try it myself.

@vlvkobal I have installed the repo and I got netdata running.
However, I don't see the temperature anywhere in the web UI.

I am checking the error.log now, but I am not sure what to search for.
Any advice on where I should look :)?

vlvkobal:
Does `grep temperature error.log` show something?

theAlchemist-dev:
Nope, sorry, I have done that, but I cannot find anything with `temp` other than `hddtemp`

theAlchemist-dev:
For the record, I have installed `netdata` using the `netdata-installer.sh` script. That should use your edited `make file` right?

vlvkobal:
You can do `make clean` before `netdata-installer.sh` to be shure that all is rebuilt from scratch.

theAlchemist-dev:
@vlvkobal Good one, sadly it makes no difference.

[...]

PR #2605, opened 2017-08-17
Add CPU temperature chart to FreeBSD plugin

vlvkobal:
Solves #2544. Unfortunately, coretemp sysctls are quite heavy calls.


ktsaou:
nice!

> heavy calls

heavy? cpu intensive?

Will the plugin disable itself when running on VPS (ie no useful info from these heavy calls)?

vlvkobal:
Module eats about 7 milliseconds/run for 4 CPUs. Yes, it disables itself on VPS and when `coretemp` kernel module isn't running.

ktsaou:
merged it.
(I thought I did - sorry for the delay)

PR #2357, opened 2017-06-18
add chrony monitory

fooltux:

PR #2639, opened 2017-08-27
Statistics for precision of time data of local chrony

domschl:
Following the suggestions in #2357, here a module for chrony based on ExecutableService.

Please comment.

### DIFF
--- a/python.d/chrony.chart.py
+++ b/python.d/chrony.chart.py
@@ -61,6 +61,15 @@ CHARTS = {
         ]}
 }
 
+CHRONY = [('Frequency', 'frequency', 1e3),
+          ('Last offset', 'lastoffset', 1e9),
+          ('RMS offset', 'rmsoffset', 1e9),
+          ('Residual freq', 'residualfreq', 1e3),
+          ('Root delay', 'rootdelay', 1e9),
+          ('Root dispersion', 'rootdispersion', 1e9),
+          ('Skew', 'skew', 1e3),
+          ('System time', 'timediff', 1e9)]
+
 
 class Service(ExecutableService):
     def __init__(self, configuration=None, name=None):
@@ -69,15 +78,6 @@ class Service(ExecutableService):
         self.command = "chronyc -n tracking"
         self.order = ORDER
         self.definitions = CHARTS
-
-    CHRONY = [('Frequency', 'frequency', 1e3),
-              ('Last offset', 'lastoffset', 1e9),
-              ('RMS offset', 'rmsoffset', 1e9),
-              ('Residual freq', 'residualfreq', 1e3),
-              ('Root delay', 'rootdelay', 1e9),
-              ('Root dispersion', 'rootdispersion', 1e9),
-              ('Skew', 'skew', 1e3),
-              ('System time', 'timediff', 1e9)]
 
     def _get_data(self):
         """
@@ -99,7 +99,7 @@ class Service(ExecutableService):
             if len(value) > 0:
                 parsed[key] = value.split()[0]
 
-        for key, dim_id, multiplier in self.CHRONY:
+        for key, dim_id, multiplier in CHRONY:
             try:
                 data[dim_id] = int(float(parsed[key]) * multiplier)
             except (KeyError, ValueError):
--- a/python.d/chrony.chart.py
+++ b/python.d/chrony.chart.py
@@ -76,17 +76,25 @@ class Service(ExecutableService):
         :return: dict
         """
         try:
-            chrony_dict = {ln.split(':', 1)[0].strip(): ln.split(':', 1)[1].strip().split(' ')[0]
-                           for ln in self._get_raw_data()[1:]}
-            return {'timediff': int(float(chrony_dict['System time']) * 1e9),
-                    'lastoffset': int(float(chrony_dict['Last offset']) * 1e9),
-                    'rmsoffset': int(float(chrony_dict['RMS offset']) * 1e9),
-                    'rootdelay': int(float(chrony_dict['Root delay']) * 1e9),
-                    'rootdispersion': int(float(chrony_dict['Root dispersion']) * 1e9),
-                    'skew': int(float(chrony_dict['Skew']) * 1e3),
-                    'frequency': int(float(chrony_dict['Frequency']) * 1e3),
-                    'residualfreq': int(float(chrony_dict['Residual freq']) * 1e3)
-                    }
+            lines = self._get_raw_data()
+            if lines is not None:
+                chrony_dict = {}
+                for line in lines:
+                    lparts = line.split(':', 1)
+                    value = lparts[1].strip().split(' ')[0]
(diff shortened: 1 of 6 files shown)


ktsaou:
@domschl thanks!
Waiting for @l2isbad to review this.

python.d/chrony.chart.py
```diff
@@ -0,0 +1,92 @@
...
+        self.definitions = CHARTS
+
+    def _get_data(self):
+        """
+        Format data received from shell command
+        :return: dict
+        """
+        try:
+            chrony_dict = {ln.split(':', 1)[0].strip(): ln.split(':', 1)[1].strip().split(' ')[0]
+                           for ln in self._get_raw_data()[1:]}
```
l2isbad: `self._get_raw_data()` [returns](https://github.com/firehol/netdata/blob/master/python.d/python_modules/base.py#L919-L921) `None` if there is any problem.
```python
>>> None[1:]
TypeError: 'NoneType' object has no attribute '__getitem__'
```

Also dict comprehensions doesn't work in python 2.6

domschl:
@l2isbad thx for the input! Please have a look, if that addresses the points: I've added checks for None-return vals of _get_raw_data() and expanded the dict comprehension.

domschl:
Added some more sanity checks in case chronyc changes output format.

l2isbad:
The whole method in `try/except` block doesn't looks cool 😄 

I install chrony and the output from "chronyc -n tracking":

```
Reference ID    : 89.109.251.22 (89.109.251.22)
Stratum         : 2
Ref time (UTC)  : Mon Sep  4 08:57:56 2017
System time     : 0.000184203 seconds slow of NTP time
Last offset     : -0.000071091 seconds
RMS offset      : 0.001949425 seconds
Frequency       : 17.784 ppm fast
Residual freq   : -0.004 ppm
Skew            : 0.242 ppm
Root delay      : 0.111796 seconds
Root dispersion : 0.002530 seconds
Update interval : 1029.9 seconds
Leap status     : Normal
```
Or 

```
Reference ID    : 127.127.1.1 ()
Stratum         : 10
Ref time (UTC)  : Mon Sep  4 09:18:40 2017
System time     : 0.000000000 seconds fast of NTP time
Last offset     : +0.000000000 seconds
RMS offset      : 0.000000000 seconds
Frequency       : 17.784 ppm fast
Residual freq   : +0.000 ppm
Skew            : 0.000 ppm
Root delay      : 0.000000 seconds
Root dispersion : 0.000001 seconds
Update interval : 0.0 seconds
Leap status     : Not synchronised
```

Do we need to check "Leap status"? If not this is my example (not tested but should work)


```python
CHRONY = [('Frequency', 'frequency', 1e3),
          ('Last offset', 'lastoffset', 1e9),
          ('RMS offset', 'rmsoffset', 1e9),
          ('Residual freq', 'residualfreq', 1e3),
          ('Root delay', 'rootdelay', 1e9),
          ('Root dispersion', 'rootdispersion', 1e9),
          ('Skew', 'skew', 1e3),
          ('System time', 'timediff', 1e9)]
....

def _get_data(self):
    if not raw_data:
        return None
    
    raw_data = (line.split(':', 1) for line in raw_data)
    parsed, data = dict(), dict()

    for line in raw_data:
        try:
            key, value = (x.strip() for x in line)
        except ValueError:
            continue
        parsed[key] = value.split()[0]

    for key, dim_id, multiplier in CHRONY:
        try:
            data[dim_id] = int(float(raw[key]) * multiplier)
        except KeyError:
            continue

    return data or None

```

python.d/chrony.chart.py
```diff
@@ -0,0 +1,108 @@
...
+        self.definitions = CHARTS
+
+    CHRONY = [('Frequency', 'frequency', 1e3),
+              ('Last offset', 'lastoffset', 1e9),
+              ('RMS offset', 'rmsoffset', 1e9),
+              ('Residual freq', 'residualfreq', 1e3),
+              ('Root delay', 'rootdelay', 1e9),
+              ('Root dispersion', 'rootdispersion', 1e9),
+              ('Skew', 'skew', 1e3),
+              ('System time', 'timediff', 1e9)]
```
l2isbad: Move the `CHRONY` list outside the Service class.

python.d/chrony.chart.py
```diff
@@ -0,0 +1,108 @@
...
+
+        raw_data = (line.split(':', 1) for line in raw_data)
+        parsed, data = dict(), dict()
+
+        for line in raw_data:
+            try:
+                key, value = (l.strip() for l in line)
+            except ValueError:
+                continue
+            if len(value) > 0:
```
l2isbad: In what cases do we need this check?

domschl:
@l2isbad thanks again! I've implemented your suggestions with three small changes:
* typo: `int(float(raw[key]) * multiplier)`   needs to be `int(float(parsed[key]`...
* `value.split()[0]` crashes on `value=''` with index out-of-range, fixed with check, in case chrony sends a line ending with ':' (e.g. 'Frequency :').
* `float(parsed[key])` crashes, if `parsed[key]` is non-numeric, added exception-type (e.g. if chrony sends 'Frequency : not initialized').

python.d/chrony.chart.py
domschl: you are too fast ;-) S.b.

python.d/chrony.chart.py
domschl: will do...

python.d/chrony.chart.py
l2isbad: > value.split()[0] crashes on value=''

then
```python
if value:
    ...
```
?
No reason to len() (if i understand it right :smile:)

domschl: right you are! Done.

l2isbad:
looks ok @ktsaou 
thx @domschl 
And if you have time please do:
1. Add chrony to https://github.com/firehol/netdata/wiki/Add-more-charts-to-netdata
2. https://github.com/firehol/netdata/blob/master/python.d/README.md

l2isbad:
And
3. https://github.com/firehol/netdata/blob/master/conf.d/python.d.conf

And you have to update Makefiles or it will not be installed:
1. https://github.com/firehol/netdata/blob/master/conf.d/Makefile.am
2. https://github.com/firehol/netdata/blob/master/python.d/Makefile.am

domschl:
Concerning 'leap status': this just gives textual information about possible upcoming leap seconds. This can only happen twice a year, and isn't something that can be plotted in a useful way. Possible values are: `Normal`, `Insert second`, `Delete second` or `Not synchronized`. So for now, I don't do anything with those.

Makefiles & wik, doc: I will do the updates.

domschl:
Updates to Makefile, conf and documentation done. Please let me know if there's anything else to be done.

@l2isbad : thanks a lot for the help and guidance.

l2isbad:
@ktsaou ready

ktsaou:
merged! thanks!

PR #2642, opened 2017-08-28
Pull MYSQL config settings for extra alerting possibilities

Rockstar04:
This PR adds flexibility in generating custom graphs and alerts based off the configuration of a MySQL server. My use case was to alert when the server was nearing the `max_connection` limit.


ktsaou:
@Rockstar04 thanks!
Waiting for @l2isbad is review this.

python.d/mysql.chart.py
```diff
@@ -224,6 +247,7 @@ def slave_running(value):
     'threads': {
         'options': [None, 'mysql Threads', 'threads', 'threads', 'mysql.threads', 'line'],
         'lines': [
+            ['var_max_connections', 'max', 'absolute'],
```
l2isbad: Do we need this on chart?
I see its can be[ hidden](https://github.com/firehol/netdata/wiki/External-Plugins#dimension) (`giving the keyword hidden will make this dimension hidden, it will take part in the calculations but will not be presented in the chart`)

Rockstar04: I was on the fence since it's technically not a thread (just the limit) bit I think it helps give the chart more context, and you can quickly glance at it and gauge how close you are to reaching your connection limit.

Rockstar04: Sorry for the double reply, I'm on a road trip with the family so I can look for myself, but is there a way to add this as a gauge? Set the peak for the gauge as the max_connection_limit and the needle would be the number of connection threads.

l2isbad: >  it helps give the chart more context, and you can quickly glance at it and gauge how close you are to reaching your connection limit.

sounds good 👍

python.d/mysql.chart.py
```diff
@@ -10,9 +10,32 @@
...
+GLOBAL_VARS = [
+ 'back_log',
+ 'binlog_cache_size',
+ 'host_cache_size',
+ 'innodb_buffer_pool_size',
+ 'innodb_log_buffer_size',
+ 'innodb_log_file_size',
+ 'join_buffer_size',
+ 'key_buffer_size',
+ 'max_connections',
```
l2isbad: Since you are using `max_connections` only i suggest to comment out all other list members

Rockstar04: That's alright a much slimmed down list of all the possible config values. I left the ones I thought others would find interesting and/or that could be added to other charts in a similar way to max_connection_limit was. Commenting them out would be a good compromise, it would make clear to others that we can fetch any config value, but (I assume) lower the memory and processing requirements for values we aren't currently using.

Rockstar04: Sorry for the typos there, I'm trying to reply from mobile while getting the house ready for company to come over :P

l2isbad: > That's alright a much slimmed down list

I know. But there is no reason to [loop](https://github.com/firehol/netdata/pull/2642/files#diff-f580f78a2d9a19e60f501e7c13470e0eL425) through the whole list if you are not using them at the moment.

Rockstar04:
@l2isbad I was poking around the JS for gauges, is there a way we can set the gauge max to the `var_max_connections` variable ?

l2isbad:
>   I was poking around the JS for gauges, is there a way we can set the gauge max to the var_max_connections variable ?

@ktsaou

ktsaou:
> is there a way we can set the gauge max to the var_max_connections variable

Currently no. I have considered this, but it needs a protocol enhancement to support them. I have opened #2688 for me to remember to do it.

There are a few alternatives through. Take disk space for example. The plugin exposes a chart with 2 dimensions `avail` and `used` in a `stacked` chart. So the user can see the total and is also able to have gauges with total, avail or used. It is not super flexible, but it is a good workaround.

ktsaou:
You could also provide a chart with utilization percentage. This way, the max will always be known: 100%.

Rockstar04:
Awesome. I'll look into the options and get something put together. Once this is out of the way I have at least 2 more PR's I'll be submitting as well.

Rockstar04:
Gauges from our Master MySQL Server:
master

Gauges from our Slave MySQL Server:
slave

What would you guys like to see from this so we can get something merged, and revisit it later (possibly after #2688)?

ktsaou:
I am ok with this functionality (aesthetically I would prefer gauges, but I guess we do this later).

The alarm should use the new chart (the %, not the variable that is not there yet). Also there could be another alarm (probably `to: silent` if this is not fatal) for the innodb buffer.

For the code review I think @l2isbad should review it.

Rockstar04:
I updated them to all use gauges, can you elaborate on what you want for the alarm? Instead of doing `calc: ($Threads_connected / $var_max_connections) * 100` use `calc: $Max_connections_ratio` ?

ktsaou:
> can you elaborate on what you want for the alarm? Instead of doing calc: ($Threads_connected / $var_max_connections) * 100 use calc: $Max_connections_ratio ?

ok, I just checked the charts you create:

1. you have added 2 dimensions `var_max_connections` and `Max_connections_ratio` to the `threads` chart. These should be removed. All the dimensions of the chart, should be of the same context (same things).

2. you have added dimension `Innodb_buffer_pool_utilization` to `innodb_buffer_pool_bytes`. It should be removed too.

3. a new chart should be created `connections_utilization` that should include dimension `Max_connections_ratio` as `utilization` with algorithm `absolute` and units `percentage`.

4. a new chart should be created `innodb_buffer_pool_utilization` that should include dimension `Innodb_buffer_pool_utilization` as `utilization` with algorithm `absolute` and units `percentage`.

Then the alarms and `dashboard_info.js` should use these charts.

(sorry for these late changes - I just stepped into the code).

Rockstar04:
I don't agree with any of those charting changes. The whole point of this PR was to give more context to existing charts, not provide the end user even more data overload. Adding charts for the sake of charts is not something I want my name on. I've updated the database servers I manage to pull updates from the fork I made so I can keep the charts and graphs that are useful to me.

ktsaou:
I am sorry you feel like that.

netdata charts have a meaning. You can't add a dimension of `percentage` to a chart of `bytes`. It will really confuse people and it will break almost everything about backends. People will get frustrated if netdata cannot keep some consistency in this area.

Internally, netdata has some structures to mark charts as "detailed" or "hidden". These are not exposed to the dashboard yet, but the idea is that there can be charts that are not as important. Check the `uptime` chart for example - it is a stupid chart and eventually it will be hidden.

I think the information you have added is useful. 

Of course it is your call.

Thank you anyway!

ktsaou:
I just recalled we have this functionality:

```
<div data-netdata="system.ram"
    data-dimensions="used|buffers|active|wired"
    data-append-options="percentage"
    data-chart-library="easypiechart"
    data-title="Used RAM"
    data-units="%"
    data-easypiechart-max-value="100"
    data-width="9%"
    data-after="-480"
    data-points="480"
    data-colors="#66AA00">
```

This is used RAM chart on the dashboard.

image

So, you request from netdata the percentage of any dimension of a chart, vs the chart total. So, if there is a chart that properly breaks down connections and innodb buffer pool size, then netdata can calculate percentages.

proper break down is this:

- `used`
- `reserved`
- `free`

So, the total is the sum of all dimensions. `free` can be hidden (not visible anywhere).

If we update the existing charts to show a proper breakdown of the resource utilization, then the percentages can be calculated on the fly.

PR #2657, opened 2017-08-30
support domain name sockets

### LINKED ISSUE #2656: Feature Request: Add the ability to listen on a unix domain socket instead of binding to an IP address.
As of right now, netdata only allows serving it's HTTP interface over a regular TCP/IP connection.  This works fine in most cases, but has some limitations, most notably:
1. Access control for local users is complicated.
2. Performance when proxying netdata through another local web server is not as good as it could be.

Adding the ability to listen on a unix domain socket would solve both issues because:
1. You can then use regular POSIX filesystem permissions (and also ACL's) to control local access.
2. You don't incur a round-trip through the TCP and IP stacks when for each request.

Providing this functionality is pretty standard for web applications that expect to be proxied through another web server in regular usage, and connecting to a web application that provides this type of access is well supported in most standard web servers.

Adding this would also provide a much more secure and robust way to handle #2526.

ktsaou:
netdata now supports unix domain sockets:

1. for the web API (use `bind to = unix:/path/to/netdata.sock`). This allows web servers in front of netdata to use a socket file while proxying netdata.

2. for statsd

3. for sending and receiving metrics between netdata netdata servers (if they are on the same host)

4. for pushing metrics to backends

Fixes #2656

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -80,6 +80,7 @@
 #include <sys/time.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <sys/un.h>
 #include <time.h>
 #include <unistd.h>
 #include <uuid/uuid.h>
--- a/src/main.c
+++ b/src/main.c
@@ -172,7 +172,7 @@ void kill_childs()
         // it is detached
         // pthread_join(w->thread, NULL);
 
-        w->obsolete = 1;
+        WEB_CLIENT_IS_OBSOLETE(w);
     }
 
     int i;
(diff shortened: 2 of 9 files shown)


ktsaou:
API use of `unix:` sockets (i.e. netdata is the server, listening on them) is tested and to works.

client use of `unix:` sockets (streaming, backends, etc) is implemented but it is untested. If anyone can test this, please report success or failure.

elfgoh:
Found this gem of a feature @ktsaou . Thanks for creating this feature!

PR #2684, opened 2017-09-06
web_log plugin: localhost added to the "address" regex"

l2isbad:
partially fix https://github.com/firehol/netdata/issues/2670

Not sure about matching "hostnames" as addresses by default, but it can be easily added in `custom regex` if needed.


ktsaou:
merged!
thanks!

PR #2699, opened 2017-09-08
Update bind_rndc.conf

fooltux:
fix bug

### DIFF
--- a/conf.d/health.d/bind_rndc.conf
+++ b/conf.d/health.d/bind_rndc.conf
@@ -1,5 +1,5 @@
  alarm: bind_rndc_stats_file_size
-      on: bind_rndc.stats_size
+      on: bind_rndc_named._namedstats_size
    units: megabytes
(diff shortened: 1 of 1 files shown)


conf.d/health.d/bind_rndc.conf
```diff
@@ -1,5 +1,5 @@
  alarm: bind_rndc_stats_file_size
-      on: bind_rndc.stats_size
+      on: bind_rndc_named._namedstats_size
```
l2isbad: Must be `template` not `alarm`. `on: bind_rndc.stats_size` is correct.


```
 template: bind_rndc_stats_file_size
      on: bind_rndc.stats_size
   units: megabytes
   every: 60
    calc: $stats_size
    warn: $this > 512
    crit: $this > 1024
    info: Bind stats file is very large! Consider to create logrotate conf file for it!
      to: sysadmin
```

PR #2704, opened 2017-09-09
added CPU core and package thermal throttling support

ktsaou:
This extends netdata to monitor CPU thermal events, throttling the CPUs.

2 new charts have been added:

1. Per core throttling count (how many times the CPU core has been throttled due to its thermal conditions)
2. Package throttling count (how many times the CPU package has been throttled due to the thermal conditions of any of its cores) - this is disabled be default, since it the information it provides is inferior to per core throttling count.

Also, cpufreq is replicated into netdata (from python), however the python version can use cpufreq statistics and estimate the average cpu frequency over the iteration period, so the internal one is disabled.

### DIFF
--- a/coverity-scan.sh
+++ b/coverity-scan.sh
@@ -35,9 +35,11 @@ make clean || exit 1
 
 "${covbuild}" --dir cov-int make -j${cpus} || exit 1
 
+echo >&2 "Compressing data..."
 tar czvf netdata-coverity-analysis.tgz cov-int || exit 1
 
-curl --form token="${token}" \
+echo >&2 "Sending analysis..."
+curl --progress-bar --form token="${token}" \
   --form email=costa@tsaousis.gr \
   --form file=@netdata-coverity-analysis.tgz \
   --form version="${version}" \
--- a/src/proc_interrupts.c
+++ b/src/proc_interrupts.c
@@ -18,10 +18,10 @@ struct interrupt {
 
 // since each interrupt is variable in size
 // we use this to calculate its record size
-#define recordsize(cpus) (sizeof(struct interrupt) + (cpus * sizeof(struct cpu_interrupt)))
+#define recordsize(cpus) (sizeof(struct interrupt) + ((cpus) * sizeof(struct cpu_interrupt)))
 
 // given a base, get a pointer to each record
-#define irrindex(base, line, cpus) ((struct interrupt *)&((char *)(base))[line * recordsize(cpus)])
+#define irrindex(base, line, cpus) ((struct interrupt *)&((char *)(base))[(line) * recordsize(cpus)])
 
 static inline struct interrupt *get_interrupts_array(size_t lines, int cpus) {
     static struct interrupt *irrs = NULL;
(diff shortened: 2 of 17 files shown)


ktsaou:
@l2isbad, @tycho have a look at this.

The top chart is from the new code inside netdata. The bottom one comes from python.d.plugin (cpufreq):

image

I understand this is a snapshot of the current CPU frequency, however it is strange that netdata usually reports higher CPU clocks. It should be random, but it is not.

The python plugin has support for `cpufreq/stats/time_in_state`, but I can't find any server with this available. So, both plugins work with `cpufreq/scaling_cur_freq`.

tycho:
@ktsaou In order to have `cpufreq/stats/time_in_state` you need to have the `cpufreq-stats` module loaded or built into the kernel (my kernel has `CONFIG_CPU_FREQ_STAT=y`).

I don't see an appreciable difference between the two graphs you show. They're reading the same data source, and it's a relatively worthless instantaneous snapshot of the selected CPU frequency. The frequency it reads will vary wildly depending on the `cpufreq` governor in use as well, because the governor chooses when it's time to scale up or down and each governor has its own opinion about when is best. If you try with the `performance` governor, the frequencies will flat line at the max frequency. If you use the `conservative` governor, you'll likely see a lot less variability, but system performance will be poorer because `conservative` doesn't switch out of low power states very often. You can try this by running either:
```
$ cpupower frequency-set -g conservative
```
or, if you don't have `cpupower` installed:
```
$ for GOV in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do echo -n conservative > $GOV; done
```
Honestly trying to get the graphs to look any good without `cpufreq-stats` turned on is going to be difficult and of questionable benefit. IMO, not worth it.

tycho:
Also, it looks like the "accurate mode" code is broken in the Python version somehow. I'm trying to figure out why.

ktsaou:
thanks! yes, I concluded the same.

I have a few systems with `CONFIG_CPU_FREQ_STAT=y`, but none is showing up `cpufreq/stats/time_in_state`. I also modprobed the module. No luck.

Any ideas how to enable it?

example 1:

```sh
# uname -a
Linux boxe 4.9.44-1-lts #1 SMP Thu Aug 17 14:28:40 CEST 2017 x86_64 GNU/Linux
# zcat /proc/config.gz | grep CONFIG_CPU_FREQ_STAT
CONFIG_CPU_FREQ_STAT=y
CONFIG_CPU_FREQ_STAT_DETAILS=y
# find /sys -name time_in_state
< nothing >
```

example 2:

```sh
# uname -a
Linux box 4.4.26-gentoo #1 SMP PREEMPT Sun Nov 13 00:21:59 EET 2016 x86_64 Intel(R) Celeron(R) CPU J1900 @ 1.99GHz GenuineIntel GNU/Linux
# zcat /proc/config.gz | grep CONFIG_CPU_FREQ_STAT
CONFIG_CPU_FREQ_STAT=m
CONFIG_CPU_FREQ_STAT_DETAILS=y
# lsmod | grep cpufreq
cpufreq_stats          16384  0
# find /sys -name time_in_state
< nothing >
```

any ideas?

tycho:
What governor do you have selected?

```
$ find /sys -name 'scaling_governor' | xargs cat
```

ktsaou:
powersave in both systems.

I am trying to mode these plugin into the netdata to remove the python dependency. Also, netdata is very efficient in this (it keeps the files open while reading `/sys`, so it is a lot faster than python). If I can enable `time_in_state` I could finish it in netdata and obsolete the python ones.

tycho:
Oh, I bet I know what it is. I bet it's that ridiculous `intel_pstate` driver. Try booting with `intel_pstate=disable` on your kernel command line.

ktsaou:
yes, I am using the `intel_pstate` driver on all systems.
Shall I permanently disable it? Why is it ridiculous and how the heck is it the default?

tycho:
I keep it disabled on all my systems. It has bitten me in the past (both at work where and at home). It makes really bad decisions about when to change P-states, and causes all kinds of extra latency. Very bad for performance-sensitive workloads. Aside from making poor P-state decisions, it also doesn't support `cpufreq-stats` which is lame.

I figured out what was wrong with `cpufreq.chart.py`. Looks like commit 0bf73332fb9a43db97f96fdbda06a2e887b1f800 broke it. Here's a fix: https://github.com/tycho/netdata/tree/fix-cpufreq-accurate-mode

tycho:
Hmm, I think my fix needs work actually. It should keep trying the accurate mode in case it becomes available (e.g. switching from the `schedutil` governor which doesn't support `cpufreq-stats` to another one that does should still allow you to see accurate data)

tycho:
Better fix:
https://github.com/tycho/netdata/commit/61dbe88ef1b1d8d927c6eb13bd00046113841cd2

ktsaou:
ok, make a PR to merge it and I'll try to replicate it in C.

ktsaou:
oh, it just a removal of 2 lines? ok, I have it.

ktsaou:
got the PR. thanks!

tycho:
Did `intel_pstate=disable` make `time_in_state` show up for you? 😄

ktsaou:
Just rebooted and yes, now I have `acpi-cpufreq` and `time_in_state` files.

And I can see the difference (the 2nd chart is python based, accurate):

image

ktsaou:
However, on my laptop, the CPU frequency I can now get is 1500MHz. Strange for a CPU with base frequency 2.8GHz and turbo 3.4GHz...

```sh
# cat cpuinfo_max_freq 
2801000

# cat scaling_max_freq 
1500000

# cat scaling_available_frequencies 
2801000 2800000 2700000 2500000 2400000 2200000 2100000 1900000 1700000 1600000 1500000 1400000 1200000 1100000 900000 800000 
```

tycho:
Yeah, it's supposed to be the "average" frequency since the last read, but it's still not quite right. If you're a bit late/early reading the file, the number will vary by a few MHz. Easy to repro: `killall -SIGSTOP python; sleep 1.5; killall -SIGCONT python` while netdata is running will cause the graph to go a bit nuts.

A better way to do it would be to keep the values from the previous read. On the 2nd read, subtract out all the previous times. Then figure out the average frequency from that... Going to try and implement that in the Python version now.

tycho:
Fixed!

Demo here: https://netdata.uplinklabs.net/#menu_cpu_submenu_cpufreq;theme=slate;help=true;update_always=true

It's not super readable, but it works: https://github.com/tycho/netdata/commit/865aedf6bfd26d96b887c92be68bcbadc2a0f7eb

tycho:
Oh, I think I missed your point earlier. Yes, you can clock way under your rated clock speed. That's what's neat about P-states. If I'm annoyed with my CPU fans going on or my laptop burning my legs off while watching a Twitch stream (which is always a CPU hungry process) I can just forcibly underclock the CPU by setting a low maximum P-state (e.g. cap it at 1GHz). The Twitch stream still plays just fine, but it doesn't turbo boost or eat nearly as much power to do it.

ktsaou:
hm... but somehow it ignores values above 1.5GHz:

```sh
# cat scaling_available_frequencies 
2801000 2800000 2700000 2500000 2400000 2200000 2100000 1900000 1700000 1600000 1500000 1400000 1200000 1100000 900000 800000 

# echo "2801000" >scaling_max_freq

# cat scaling_max_freq 
1500000
```

any ideas?

tycho:
ACPI might be capping it. You can usually fix that with:
```
$ echo 1 > /sys/module/processor/parameters/ignore_ppc
```

ktsaou:
But it respects values below 1.5GHz:

```sh
# echo "1400000" >scaling_max_freq 
# cat scaling_max_freq 
1400000
```

ktsaou:
> ACPI might be capping it. You can usually fix that with:

yes, that did the trick: `echo 1 > /sys/module/processor/parameters/ignore_ppc`

tycho:
Awesome! Enjoy having better performance now. 😉

ktsaou:
> Awesome! Enjoy having better performance now

thnaks! One final note: what about turbo? So, now it is limited to the max base frequency of the processor...

tycho:
Turbo is represented in a funny way. Note this strange frequency listed in your "scaling_available_frequencies":

**2801000** 2800000 2700000 2500000 2400000 2200000 2100000 1900000 1700000 1600000 1500000 1400000 1200000 1100000 900000 800000 

That weird 2801MHz number is a magic value indicating turbo boost. This does give you a nice way to enable/disable turbo. If you set `scaling_max_freq` to 2800MHz, you will never use a turbo boost frequency. Great for when you're looking for reproducible benchmarking results, or simply to conserve energy.

Unfortunately this means that you can't see what frequency you are turboing into with the `time_in_state` file, but I don't think there's a nice in-kernel way to see that at the moment anyway.

If you want to see what turbo boost frequency you're actually hitting, you can watch the output of `turbostat -d`. The source of `turbostat` is [in the kernel tree](https://github.com/torvalds/linux/tree/master/tools/power/x86/turbostat) if you don't have a package available for it, or want to adapt it for netdata (it could replace the current `cpufreq` code for x86/x86_64).

ktsaou:
nice!

ktsaou:
Turbostat output is nice:

```
turbostat: cpu3 jitter 26083 8533
Core	CPU	Avg_MHz	Busy%	Bzy_MHz	TSC_MHz	IRQ	SMI	C1	C1E	C3	C6	C7s	C8	C9	C10	C1%	C1E%	C3%	C6%	C7s%	C8%	C9%	C10%	CPU%c1	CPU%c3	CPU%c6	CPU%c7	CoreTmp	PkgTmp	GFX%rc6	GFXMHz	Totl%C0	Any%C0	GFX%C0	CPUGFX%	Pkg%pc2	Pkg%pc3	Pkg%pc6	Pkg%pc7	Pkg%pc8	Pkg%pc9	Pk%pc10	PkgWatt	CorWatt	GFXWattRAMWatt	PKG_%	RAM_%
-	-	63	4.06	1562	2808	8637	280	1603	689	136	538	3	2733	0	2358	7.84	1.22	0.17	1.29	0.01	25.36	0.00	60.30	16.61	0.17	1.78	77.38	50	53	96.20	350	27.86	17.66	3.60	1.03	16.63	40.30	0.00	0.00	0.00	0.00	0.00	4.98	0.92	0.06	1.30	0.00	0.00
0	0	73	4.00	1816	2808	916	35	235	52	2	28	0	174	0	257	20.84	0.19	0.01	0.83	0.00	15.39	0.00	58.77	24.84	0.04	1.14	69.98	49	53	96.20	350	27.86	17.66	3.60	1.03	16.62	40.30	0.00	0.00	0.00	0.00	0.00	4.98	0.92	0.06	1.30	0.00	0.00
0	4	63	4.20	1509	2808	1024	35	212	103	18	72	0	483	0	301	7.47	0.93	0.12	1.10	0.00	33.50	0.00	52.97	24.63
1	1	44	3.06	1436	2808	969	35	141	54	10	107	0	310	0	284	4.87	1.49	0.05	2.00	0.00	23.16	0.00	65.65	12.74	0.14	2.64	81.41	50
1	5	70	4.24	1660	2808	1095	35	191	88	22	76	1	408	0	267	4.21	1.39	0.48	1.43	0.00	27.56	0.00	61.02	11.57
2	2	61	4.11	1497	2808	848	35	187	56	14	35	0	248	0	344	6.81	1.79	0.34	0.95	0.00	21.27	0.00	64.91	16.41	0.31	1.62	77.55	47
2	6	51	3.76	1362	2808	1744	35	196	109	13	63	0	285	0	299	7.75	1.57	0.09	1.38	0.00	20.54	0.00	65.24	16.76
3	3	64	4.37	1470	2808	905	35	171	148	45	69	0	459	0	328	3.78	1.09	0.15	1.09	0.00	34.81	0.00	55.06	13.15	0.18	1.72	80.58	45
3	7	81	4.77	1689	2808	1136	35	270	79	12	88	2	366	0	278	7.00	1.32	0.15	1.50	0.04	26.67	0.00	58.78	12.74
```

I wonder what is `jitter`, that is reported like an issue...

tycho:
One nice thing about the turbostat stuff is that it'll work regardless of what P-state governor you have selected, or whether you use `intel_pstate` or not, etc. One downside is that it needs to read MSRs, which means you a) need the "msr" module loaded or built in to the kernel, and b) need to be able to read /dev/cpu/*/msr (so, probably running as root).

The jitter thing appears to mean that the cost of reading the APERF/MPERF MSRs differed by >2x for some reason.

ktsaou:
> One downside is that it needs to read MSRs, which means you a) need the "msr" module loaded or built in to the kernel, and b) need to be able to read /dev/cpu/*/msr (so, probably running as root).

hm... we can easily turn `turbostat` to an external setuid / setcap plugin, that will push all these nice metrics to netdata. Merging it into netdata seems inappropriate though (it is quite big and complex).

It would be interesting to detect if the machine is a physical or virtual and automatically enable this on physical servers...

tycho:
It may just be enough to see if you can read APERF/MPERF, and not worry about whether it's virtualized or not. If you can read it, you're either on bare metal or you're in a VM that is permitting you to read that data. Some of the largest Amazon EC2 instance types support guest P-state/C-state control (and thus support reading APERF/MPERF). So even though it's a VM doesn't mean you should ignore the turbo data.

PR #2712, opened 2017-09-10
cpufreq.chart.py: use much more accurate accounting for average CPU frequency

tycho:
The way I was tracking it before was bad because if we were too early or late reading the time_in_state file, we might get really bogus results.

Now we remember the last snapshot we saw, and show the average frequency *since we last refreshed*.

### DIFF
--- a/python.d/cpufreq.chart.py
+++ b/python.d/cpufreq.chart.py
@@ -38,20 +38,31 @@ class Service(SimpleService):
         data = {}
 
         if self.accurate_exists:
-            elapsed = time.time() - self.timetable['last']
-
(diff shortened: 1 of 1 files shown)


tycho:
I realize this code will be dead eventually when the C version appears, but the C version should probably have a good source of inspiration...

ktsaou:
thanks!



> I realize this code will be dead eventually when the C version appears, but the C version should probably have a good source of inspiration...

Well, we have to do the C code better and `turbostat` changed my view. So, I think this will stay there for some time...

ktsaou:
merged it.

PR #2714, opened 2017-09-10
RFC: implement simple 'turbostat' python.d chart module

tycho:
**This isn't intended to be merged for now.** I'd like @ktsaou to see this and try it out, though, since we were looking at turbostat recently.

This module requires that `turbostat` be installed. It also requires that `/dev/cpu/*/msr` be readable by whatever user netdata runs the Python modules under, and that `turbostat` have the `CAP_SYS_RAWIO` capability. This is sufficient for testing:
```
# setcap cap_sys_rawio=ep /usr/bin/turbostat
# chmod +r /dev/cpu/*/msr
```

You can see this in action here: https://netdata.uplinklabs.net/#menu_cpu_submenu_turbostat;theme=slate;help=true

There are a bunch of other metrics we could plug into this, but I wanted to proof-of-concept this first. In whatever implementation we end up using, I'd prefer if we don't repeatedly shell out to turbostat. Ideally we'd just start turbostat and continuously stream and parse its output. Unfortunately the output modes for turbostat other than `--Dump` aren't very machine-readable.

### DIFF
--- a/collectors/python.d.plugin/turbostat/turbostat.chart.py
+++ b/collectors/python.d.plugin/turbostat/turbostat.chart.py
@@ -1,0 +1,269 @@
+# -*- coding: utf-8 -*-
+# Description: turbostat netdata python.d module
+# Author: Steven Noonan (tycho)
+
+import copy
+import os
+import struct
+import subprocess
+import time
+
+disabled_by_default = True
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+_BASE10_LINES = set(['CPU', 'core', 'IRQ', 'SMI', 'package'])
+
+CHART_TEMPLATES = {
+    'avg_mhz': {
+        'options': [None, 'Average CPU clock frequency, including idle time', 'MHz', 'turbostat', 'turbostat', 'line'],
+        'lines': [
+            # lines are created dynamically in `check()` method
+        ]},
+    'busy_mhz': {
+        'options': [None, 'Average CPU clock frequency, when busy', 'MHz', 'turbostat', 'turbostat', 'line'],
+        'lines': [
+            # lines are created dynamically in `check()` method
+        ]},
+    'power': {
+        'options': [None, 'Power utilization', 'Watts', 'turbostat', 'turbostat', 'line'],
+        '_divisor': 100,
+        '_per': 'package',
+        '_replace': [ ('_watts', '') ],
+        '_metrics': ['pkg_watts', 'ram_watts', 'gfx_watts'],
+        'lines': [
+            # lines are created dynamically in `check()` method
+        ]},
+}
+
+MSR_RAPL_POWER_UNIT = 0x606
+DEVNULL = open(os.devnull, 'wb')
+
+def rdmsr(cpu, address):
+    # HACK: Doing this for debugging. Very likely wrong for anything but my system.
+    if address == MSR_RAPL_POWER_UNIT:
+        return 0x00000000000a0e03
+
+    # NOTE: Requires CAP_SYS_RAWIO since Linux 3.7-ish
+    with open('/dev/cpu/' + str(cpu) + '/msr', 'rb', buffering=0) as msr:
+        msr.seek(address)
+        data = msr.read(8)
+        v, = struct.unpack('<Q', data)
+        return v
+
+
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = []
+        self.definitions = {}
+        self.fake_name = "cpu"
+        self.assignment = {}
+        self.last_turbostat = None
+        self.last_turbostat_time = 0
+        self.rapl = {}
+
+    def _parse_stat_line(self, line):
+        line = line.split(b': ', 1)
+
+        # Line has no delimiter, probably a debug message of some kind. Ignore it.
+        if len(line) == 1:
+            return None
+
+        name, value = line[0].decode('utf-8'), line[1]
+
+        if name == 'CPU':
+            return name, int(value.split()[0])
+
+        try:
+            if name in _BASE10_LINES:
+                value = int(value, 10)
+            else:
+                value = int(value, 16)
+        except:
+            return None
+
+        return name, value
+
+    def _invoke_turbostat(self):
+        cpus = {}
+        proc = subprocess.Popen(['turbostat', '--Dump'], stdout=subprocess.PIPE, stderr=DEVNULL)
+        cpu = None
+        now = time.time()
+        for line in proc.stdout:
+            line = line.rstrip()
+
+            if not line:
+                cpu = None
+                continue
+
+            stat = self._parse_stat_line(line)
+            if stat is None:
+                #self.alert("Failed to parse line: '%s'" % (line,))
+                continue
+
+            name, value = stat
+
+            if cpu is None and name != 'CPU':
+                continue
+
+            if name == 'CPU':
+                cpu = {}
+                cpus[value] = cpu
+
+            cpu[name] = value
+
+        return now, cpus
+
+    def _dict_delta(self, before, after):
+        result = {}
+        for key, value in before.items():
+            try:
+                result[key] = after[key] - value
+            except TypeError:
+                # Probably not a numeric type. Don't worry about it.
+                pass
+
+        return result
+
+    def _get_data(self):
+        data = {}
+
+        now, turbostat = self._invoke_turbostat()
+
+        for cpuidx, stats in turbostat.items():
+            last_stats = self.last_turbostat[cpuidx]
+
+            delta = self._dict_delta(last_stats, stats)
+
+            tsc = delta['TSC']
+            aperf = delta['aperf']
+            mperf = delta['mperf']
+
+            time = now - self.last_turbostat_time
+            avg_mhz = aperf / time / 1e3
+            busy_mhz = (tsc / 1e3) * aperf / mperf / time
+
+            cpuname = 'cpu%d' % (cpuidx,)
+            data[cpuname + '_avg_mhz'] = avg_mhz
+            data[cpuname + '_busy_mhz'] = busy_mhz
+
+            pkgidx = self.assignment[cpuname]['package']
+            pkgname = 'pkg%d' % (pkgidx,)
+
+            if pkgname + '_ram_watts' not in data and pkgidx in self.rapl:
+                power_units, energy_units = self.rapl[pkgidx]
+
+                ram = max(0, delta['Joules RAM'])
+                pkg = max(0, delta['Joules PKG'])
+                gfx = max(0, delta['Joules GFX'])
+
+                data[pkgname + '_ram_watts'] = ram * energy_units / time * 100.0
+                data[pkgname + '_pkg_watts'] = pkg * energy_units / time * 100.0
+                data[pkgname + '_gfx_watts'] = gfx * energy_units / time * 100.0
+
+        self.last_turbostat_time = now
+        self.last_turbostat = turbostat
+
+        return data
+
+    def check(self):
+        split_by = None
+        try:
+            split_by = str(self.configuration['split_by'])
+            if split_by not in ['logical', 'core', 'package']:
+                self.error("Value '%s' for 'split_by' configuration option isn't a valid choice, ignoring" % (split_by,))
+                split_by = None
+        except (KeyError, TypeError):
+            self.alert("No 'split_by' option specified. Not dividing CPUs up by topology.")
+
+        try:
+            self.last_turbostat_time, self.last_turbostat = self._invoke_turbostat()
+        except:
+            self.error("Could not invoke turbostat, disabling.")
+            return False
+
+        for cpuidx, stats in self.last_turbostat.items():
+            cpuname = 'cpu%d' % (cpuidx,)
+            pkgidx = stats['package']
+            self.assignment[cpuname] = {
+                'package': pkgidx,
+                'core': stats['core'],
+            }
+
+            if pkgidx not in self.rapl:
+                try:
+                    msr = rdmsr(cpuidx, MSR_RAPL_POWER_UNIT)
+                except IOError:
+                    # We don't have CAP_SYS_RAWIO and can't rdmsr :(
+                    continue
+                power_units = 1.0 / (1 << (msr & 0xF))
+                energy_units = 1.0 / (1 << ((msr >> 8) & 0x1F));
+                self.rapl[pkgidx] = (power_units, energy_units)
+
+        if len(self.assignment) == 0:
+            self.error("Could not find any CPUs in turbostat dump")
+            return False
+
+        order = []
+
+        packages_seen = []
+
+        for chart, template in CHART_TEMPLATES.items():
+            for cpuname in sorted(self.assignment, key=lambda v: int(v[3:].split('_')[0])):
+                assignment = self.assignment[cpuname]
+
+                cpuidx = int(cpuname[3:])
+                pkgidx = assignment['package']
+                pkgname = 'pkg%d' % (pkgidx,)
+                coreidx = assignment['core']
+                corename = 'core%d' % (coreidx,)
+
+                if split_by == 'package':
+                    suffix = pkgname
+                elif split_by == 'core':
+                    suffix = pkgname + '_' + corename
+                elif split_by == 'logical':
+                    suffix = cpuname
+                else:
+                    suffix = None
+
+                chartname = chart
+                if suffix is not None:
+                    chartname += '_' + suffix
+
+                if chartname not in self.definitions:
+                    self.definitions[chartname] = copy.deepcopy(template)
+                    order.append((chartname, (pkgidx, coreidx, cpuidx)))
+
+                metrics = template.get('_metrics', [chart])
+                per = template.get('_per', 'logical')
+                divisor = template.get('_divisor', 1000)
+                substitutions = template.get('_replace', [])
+
+                prefix = cpuname
+                nickname_prefix = ''
+
+                if per == 'package':
+                    if pkgidx in packages_seen:
+                        continue
+
+                    prefix = pkgname
+                    if suffix is None:
+                        nickname_prefix = pkgname + '_'
+
+                for metric in metrics:
+                    nickname = cpuname
+                    if per == 'package':
+                        nickname = metric
+                    for term, replace in substitutions:
+                        nickname = nickname.replace(term, replace)
+                    self.definitions[chartname]['lines'].append([prefix + '_' + metric, nickname_prefix + nickname, 'absolute', 1, divisor])
+
+                if per == 'package':
+                    packages_seen.append(pkgidx)
+
+
+        self.order = [name for name, topology in sorted(order, key=lambda v: v[1])]
+
+        return True
--- a/collectors/python.d.plugin/Makefile.am
+++ b/collectors/python.d.plugin/Makefile.am
@@ -96,6 +96,7 @@ include squid/Makefile.inc
 include tomcat/Makefile.inc
 include tor/Makefile.inc
 include traefik/Makefile.inc
+include turbostat/Makefile.inc
 include unbound/Makefile.inc
 include uwsgi/Makefile.inc
 include varnish/Makefile.inc
--- a/collectors/python.d.plugin/python.d.conf
+++ b/collectors/python.d.plugin/python.d.conf
@@ -97,6 +97,7 @@ nginx_log: no
 # traefik: yes
 # tomcat: yes
 # tor: yes
+# turbostat: yes
 unbound: no
 # uwsgi: yes
 # varnish: yes
--- a/collectors/python.d.plugin/turbostat/Makefile.inc
+++ b/collectors/python.d.plugin/turbostat/Makefile.inc
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# THIS IS NOT A COMPLETE Makefile
+# IT IS INCLUDED BY ITS PARENT'S Makefile.am
+# IT IS REQUIRED TO REFERENCE ALL FILES RELATIVE TO THE PARENT
+
+# install these files
+dist_python_DATA       += turbostat/turbostat.chart.py
+dist_pythonconfig_DATA += turbostat/turbostat.conf
+
+# do not install these files, but include them in the distribution
+dist_noinst_DATA       += turbostat/Makefile.inc
--- a/collectors/python.d.plugin/turbostat/turbostat.conf
+++ b/collectors/python.d.plugin/turbostat/turbostat.conf
@@ -1,0 +1,30 @@
+# netdata python.d.plugin configuration for turbostat
+#
+# This file is in YaML format. Generally the format is:
+#
+# name: value
+#
+
+# ----------------------------------------------------------------------
+# Global Variables
+# These variables set the defaults for all JOBs, however each JOB
+# may define its own, overriding the defaults.
+
+# update_every sets the default data collection frequency.
+# If unset, the python.d.plugin default is used.
+# update_every: 1
+
+# priority controls the order of charts at the netdata dashboard.
+# Lower numbers move the charts towards the top of the page.
+# If unset, the default for python.d.plugin is used.
+# priority: 60000
+
+# penalty indicates whether to apply penalty to update_every in case of failures.
+# Penalty will increase every 5 failed updates in a row. Maximum penalty is 10 minutes.
+# penalty: yes
+
+# autodetection_retry sets the job re-check interval in seconds.
+# The job is not deleted if check fails.
+# Attempts to start the job are made once every autodetection_retry.
+# This feature is disabled by default.
+# autodetection_retry: 0


tycho:
Interesting things to try when watching these charts:

- Run an intense single-threaded CPU workload
- Run an intense multi-threaded CPU workload
- Different cpufreq governors: powersave, ondemand, conservative, performance, schedutil
- Turbo/non-turbo `scaling_max_freq` values (e.g. on my system it's `3600000` for non-turbo and `3601000` for turbo)
- Enable/disable deep C-states (will affect the maximum turbo clock multiplier)

If you need some kind of CPU workload for testing, I recommend my [n-body simulation](https://github.com/tycho/nbody). It can be invoked as follows... just vary the `OMP_NUM_THREADS` value if you want single/multi-threaded workloads (powers of two up to your logical CPU count are most interesting in a hyperthreaded system):
```
## If you are linked against the LLVM OpenMP runtime:
$ env KMP_AFFINITY=scatter,verbose,granularity=fine OMP_NUM_THREADS=1 \
   ./nbody --bodies 128

## Or if you're linked against the GCC OpenMP runtime:
$ env OMP_PLACES=cores OMP_DISPLAY_ENV=TRUE OMP_NUM_THREADS=1 \
   ./nbody --bodies 128
```

Another factor to note with turbo is that your maximum turbo ratio will vary depending on whether your workload is using a single core or multiple, and whether it's simple integer instructions, SSE instructions, or AVX instructions (which have varying effects on power utilization). If you run an n-body simulation built with `-march=core2` you will very likely see a differet maximum turbo frequency than with `-march=core-avx2`.

tycho:
Another thing to play with, @ktsaou:

- Try changing the MSR_IA32_ENERGY_PERF_BIAS value. You can do this with `cpupower set -b <N>` where N is a value from `0` to `15`, where `0` is "maximum performance" and `15` is "maximum energy efficiency". In my experience, changing this MSR only did something meaningful on laptops, with `intel_pstate` disabled (for some reason, `intel_pstate` refuses to respect this value, another reason I have it turned off). When set to `0`, my Skylake i7-6600U would clock up to 3200MHz, and when set to `15`, it clocked down to 800MHz. I keep it at around `11` (2700MHz) or `12` (2300MHz) to keep it very cool and silent.

ktsaou:
This is very interesting!

I see turbostat has a lot more columns. A lot with `Watt` appended, a few with `GFX`, and probably the most interesting `PKG_%` and `RAM_%`.

Do you know what all these are?

I mean if `PGK_%` is the real % utilization of the cpu, with 100% being the maximum the cpu can give, this is a lot better than `system.cpu`.

Similarly for `RAM_%`. Does this have anything to do with RAM bandwidth utilization? If yes, we have a huge hit here.

You know, while trying to understand why a server is slow, if we could somehow have an indication of congestion at its basic components, like RAM bandwidth congestion, PCIe bandwidth congestion, etc. we could provide very interesting alarms...

tycho:
These are the columns I have on one of my machines:

Core    CPU     Avg_MHz Busy%   Bzy_MHz TSC_MHz IRQ     SMI     C1      C1E     C3      C6      C7s     C8      C9      C10     C1%     C1E%    C3%     C6%     C7s%    C8%     C9%     C10%    CPU%c1  CPU%c3  CPU%c6  CPU%c7  CoreTmp PkgTmp  GFX%rc6 GFXMHz  Totl%C0 Any%C0  GFX%C0  CPUGFX% Pkg%pc2 Pkg%pc3 Pkg%pc6 Pkg%pc7 Pkg%pc8 Pkg%pc9 Pk%pc10 PkgWatt CorWatt GFXWatt RAMWatt PKG_%   RAM_%

On this machine in particular, `PKG_%` and `RAM_%` are zeros, so you can't always trust those values to be available. On another, `PKG_%` is available, but `RAM_%` is not.

The `Pkg%pcN` values indicate package-level C-state residency. If all cores/threads are in a low C-state, the package itself can enter a lower C-state. Here's a diagram from AnandTech showing what package C-states do, and rules about entering them (note that the C-state definitions are hardware-specific -- this diagram is for Haswell ULT SKUs):

Package C-states

The `PKG_%` and `RAM_%` fields are referring to RAPL (Running Average Power Limit) for the package and DRAM. The value for `PKG_%` is derived from MSR_PKG_PERF_STATUS, which is described in the Intel SDM volume 3 as follows:

> MSR_PKG_PERF_STATUS is a read-only MSR. It reports the total time for which the package was throttled due to the RAPL power limits. Throttling in this context is defined as going below the OS-requested P-state or T-state. It has a wrap-around time of many hours. The availability of this MSR is platform specific (see Chapter 2, “Model-Specific Registers (MSRs)” in the Intel® 64 and IA-32 Architectures Software Developer’s Manual, Volume 4).

If all CPUs are busy (i.e. in C0) and there's no throttling going on, then `PKG_%` will be 100%. If something like the `intel_rapl` or `intel_powerclamp` driver kicks in and throttles the CPU, then `PKG_%` will drop. So you can basically think of `PKG_%` as a percentage of package TDP.

l2isbad:
What's the state of the PR? Do we need this @ktsaou ?

ktsaou:
We can close this.
@tycho 	we will be working with eBPF to collect more precise processor utilization information. It would be great if you could also help us design it properly.

We will start with just calculating IPC...

tycho:
@ktsaou I don't see how that would provide equivalent information. It's useful to know the current power utilization and clock frequency, and AFAIK you can't get that from eBPF.

tycho:
See here for a demo of what the module in this PR provides:

https://netdata.uplinklabs.net/#menu_cpu_submenu_turbostat;theme=slate

I don't think you can get that from eBPF.

ktsaou:
hm... my bad then! Sorry!

Could you please rebase this to have it merged?

tycho:
Working on it, though I am currently out of town for the U.S. Thanksgiving holiday and will probably have to take a look at this early next week -- but I do need some input before that anyway.

This module still has a major caveat of requiring the ability to read MSRs. I'll explain where this module currently is at, and hopefully start a conversation about how to do things properly.

### The Requirements
In order for `turbostat` to function as a non-root user under netdata, we need:
* Read access to the `/dev/cpu/*/msr` files (e.g. via a group)
* `CAP_SYS_RAWIO` on the `turbostat` binary itself
* The netdata user needs to be in the group with read access to `/dev/cpu/*/msr`

Alternatively, we could have netdata run `turbostat` as root (e.g. via `sudo turbostat`) and bypass all the above requirements.

### My Current (Ugly) Solution

In my case I have a udev rule to grant the group `msr` read permissions for the `/dev/cpu/*/msr` files:
```
$ cat /usr/lib/udev/rules.d/99-dev-cpu-msr.rules
SUBSYSTEM=="msr", GROUP="msr", MODE="0640"
```
Since the above sets the `/dev/cpu/*/msr` files as readable by group `msr`, I've added the `netdata` user to the `msr` group.

And finally I have `CAP_SYS_RAWIO` set for the `turbostat` binary:
```
$ setcap cap_sys_rawio=ep /usr/bin/turbostat
```

### The Problem

The big question is: **how do we integrate all the changes needed to make the module work?** One option is to simply *not* integrate these changes in netdata itself, and provide documentation on how to do it.

Changing `CAP_SYS_RAWIO` on the `turbostat` binary is tricky for end users. It's something the package manager should own, but I doubt the various Linux distro maintainers can be persuaded to set `CAP_SYS_RAWIO` by default for the `turbostat` binary. If a user does `setcap` on the binary and the package gets updated, the capability gets reverted and the user would have to hand-add the capability again. To solve this for my machines, I forked the Arch `linux-tools` package and added an [`install` script for the `turbostat` package](https://git.uplinklabs.net/steven/projects/archlinux/ec2/ec2-packages.git/tree/linux-tools/turbostat.install). But not everyone can do that.

I'd strongly advise **against** trying to write our own `turbostat` binary (i.e. just so we could control `CAP_SYS_RAWIO` or run it setuid root), because there are all sorts of special cases depending on which hardware generation the CPU is (and architectural differences like AMD vs Intel). The folks who wrote `turbostat` know what they're doing.

Even if we have our own `turbostat` implementation (or binary) with the appropriate capability flag, we would still need read access to MSRs. From what I'm aware, the only way to read those in userspace is via the `/dev/cpu/*/msr` files provided by the msr module (or built-in module, depending on kernel config). The `RDMSR` instruction is a privileged instruction, so the kernel's got to handle that.

### A Better Solution?
One simple way to handle all this might be to have netdata invoke turbostat with `sudo turbostat` and just document that end users should allow the netdata user to passwordless sudo to root for just the turbostat binary, e.g. an `/etc/sudoers` line of something like:
```
netdata ALL=(root) NOPASSWD: /usr/bin/turbostat
```
That would eliminate the need for turbostat to have `CAP_SYS_RAWIO` and for `/dev/cpu/*/msr` to be readable by anything other than root -- no new udev rules, no new system groups, and no new packaging requirements for `turbostat`. It seems like the simplest way forward, but I admit it's kind of a sad way to approach it.

### Discuss!
I'd like to hear thoughts about how to handle this.

ktsaou:
For `fping`, I created a patch which I submitted to the authors and has been merged into the official `fping`. The patch allows `fping` to output data in netdata format.

If we do the same for `turbostat`, and our patch is accepted, we could just write a wrapper script to do the preparation work and then let `turbostat` push the metrics to netdata.

For `tc`, a patch was not possible, so I provided a shell script wrapper and netdata parses the output of `tc` in `C`. This was required since the output of `tc` is actually massive and I wanted it to be the fastest possible. Of course, in this case we couldn't avoid the frequent forks to execute `tc`.

So, I think the optimal would be either of these 2:

1. Provide a wrapper to call turbostat with sudo, and leave turbostat permanently running (not with `--Dump`), and a C plugin to parse the output of turbostat and push it to netdata.

or

2. Make a patch to `turbostat` to have it output netdata external plugins output. In this case, we don't need a dedicated C plugin for it. The external plugins API will just parse the output like it does for all the plugins.

tycho:
I rebased the turbostat module patch onto a0c21fe1d63cfaac741c66a5bc1ff7d827c0ebca. It's working on a few different hosts that I have here.

We still need to figure out the best path forward for this feature.

I'd really rather not submit a patch for turbostat, because it already outputs data in a reasonable format. I think the only cost to be concerned about is the `sudo` overhead, which is rather high. Invoking directly with the appropriate permissions/capabilities is very low-overhead and works, but it's a big pain to configure properly.

ktsaou:
> I'd really rather not submit a patch for turbostat, because it already outputs data in a reasonable format. I think the only cost to be concerned about is the `sudo` overhead, which is rather high. Invoking directly with the appropriate permissions/capabilities is very low-overhead and works, but it's a big pain to configure properly.

I think the best solution would be to start `turbostat` only once, instead of spawning it every second. So, instead of running `turbostat --Dump` to run `turbostat -i 1`.

If `turbostat` could output data in netdata format, a netdata plugin would not be needed. We could just start `turbostat` as a netdata plugin. If we can't do this, then there are 2 solutions:

1. write a turbostat plugin in `C` that parses the output of `turbostat -i 1`
or
2. write a script that runs `turbostat -i 1 | awk ...` to convert the output of turbostat to netdata format. This would be ugly when reviewing the process tree (`ps fax`), but it will still be better than forking `turbostat` every second.

tycho:
I really don't think reading the output of `turbostat -i N` is the way to go. That data is human-readable. We shouldn't rely on it being in any particular format, and even if we parse it correctly we'll lose precision in our data. The data from `--Dump` is raw key/value pairs that we can easily parse.

ktsaou:
I agree. This is why I suggested to make a PR to turbostat, so that it could dump raw data while in a loop (or even output netdata plugin data).

tycho:
OK, maybe a two-pronged approach:

* Submit a patch to upstream turbostat to allow `-i N` to be used with `--Dump`
* Write a C wrapper for pre-patch versions of turbostat which just executes `turbostat --Dump` on regular intervals (i.e. behaving like the post-patch version of turbostat would, but with a bit higher overhead)

Depending on which version of turbostat is available, we could run the crappy execute-on-a-loop wrapper or we could run turbostat directly.

There's also the issue of privileges, so maybe the wrapper could be used in both pre and post-patch cases, and either it will exec `turbostat -i N --Dump` or loop over invocations of `turbostat --Dump`. Then we provide instructions which state that our wrapper needs to have a line in `sudoers` for netdata to invoke it.

cakrit:
@tycho I don't see any progress here. While @ktsaou's suggestion is for an ideal solution, I did a test and I see that turbostat consistently takes under 50ms real time on my laptop. So I suggest you take another, final look at your code, add a README.md and we have @ilyam8 review and merge this. If there are any concerns regarding performance, you can have the default update every be 5sec instead of 1 in the config file and people can always change it or even disable the plugin.

@ilyam8 do you agree?

ilyam8:
> or even disable the plugin

I think it should be disable by default. All collectors that uses popen is disabled by default (this one executed `turbostat --Dump` binary and parses the output).

cakrit:
Sounds good @ilyam8. So do a review requesting these changes and, if @tycho can't progress it, you can take the code and do a PR yourself. CLA is already signed.

collectors/python.d.plugin/turbostat/turbostat.chart.py
```diff
@@ -0,0 +1,267 @@
...
+# Author: Steven Noonan (tycho)
+
+import copy
+import os
+import struct
+import subprocess
+import time
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
```
ilyam8: Please disable by default the module

`disabled_by_default` = True

This should be mentioned in the module readme, in order to enable the module user have to explicitly enable it in the `python.d.conf`

tycho: I'm not sure where the best place to note that is.

cakrit:
Closing due to inactivity. Please comment if you wish to continue with it and we'll reopen it.

satcom886:
I'd really like to see this implemented. Few of my systems don't show overclocked and boost frequencies in the `cpu.cpufreq` chart. This would solve my problems and I suspect the problems of many others.

ilyam8:
@satcom886 quick way is to download the script and put it to the python.d dir

> https://raw.githubusercontent.com/tycho/netdata/implement-turbostat-module/collectors/python.d.plugin/turbostat/turbostat.chart.py

(dont forget to remove `disabled_by_default = True
`)

tycho:
For anyone who wants to run it, be aware that there are a several prerequisites to be able to do so:
- `/dev/cpu/*/msr` must be readable by the netdata user. I did this via a [udev rule](https://git.uplinklabs.net/steven/projects/archlinux/ec2/ec2-packages.git/tree/linux-tools/99-dev-cpu-msr.rules), after adding a system group called `msr` and adding the `netdata` user to it.
- `turbostat` binary needs `cap_sys_rawio+ep`. This can be a pain to maintain if you set it manually and are using distribution packages, though you could probably set up some package manager hooks to set the capabilities any time its package gets updated.
- Your netdata systemd service definition needs `CapabilityBoundingSet=CAP_SYS_RAWIO`

satcom886:
@ilyam8 Thanks, I know that. I tried it prior to writing my comment and I loved it! It would just be a little nicer to have the plugin ready to go without having to manually download it.

The permissions thing is a little complicated, but wasn't a problem for me. I would probably force myself to make a script that would set up the group and udev rule.

Asheboy:
Thanks @tycho, this is exactly what I wanted and is great. I found these instructions by @foundObjects helpful too, https://github.com/netdata/netdata/issues/3710#issuecomment-473711184. Only thing it is missing is setting `CapabilityBoundingSet=CAP_SYS_RAWIO` on the systemd unit file.

SteveBox0:
@tycho I am getting the below error form your script. Any suggestions?

```
/usr/libexec/netdata/plugins.d/python.d.plugin turbostat debug
2022-06-14 04:28:56: python.d INFO: plugin[main] : using python v3
2022-06-14 04:28:56: python.d DEBUG: plugin[main] : looking for '/etc/netdata/python.d.conf'
2022-06-14 04:28:56: python.d INFO: plugin[main] : '/etc/netdata/python.d.conf' was not found
2022-06-14 04:28:56: python.d DEBUG: plugin[main] : looking for '/usr/lib/netdata/conf.d/python.d.conf'
2022-06-14 04:28:56: python.d DEBUG: plugin[main] : '/usr/lib/netdata/conf.d/python.d.conf' is loaded
2022-06-14 04:28:56: python.d DEBUG: plugin[main] : looking for 'pythond-jobs-statuses.json' in /var/lib/netdata
2022-06-14 04:28:56: python.d DEBUG: plugin[main] : loading '/var/lib/netdata/pythond-jobs-statuses.json'
2022-06-14 04:28:56: python.d DEBUG: plugin[main] : '/var/lib/netdata/pythond-jobs-statuses.json' is loaded
2022-06-14 04:28:56: python.d DEBUG: plugin[main] : [turbostat] looking for 'turbostat.conf' in ['/etc/netdata/python.d', '/usr/lib/netdata/conf.d/python.d']
2022-06-14 04:28:56: python.d WARNING: plugin[main] : [turbostat] 'turbostat.conf' was not found
2022-06-14 04:28:56: python.d INFO: plugin[main] : [turbostat] built 1 job(s) configs
2022-06-14 04:28:56: python.d CRITICAL: turbostat[turbostat] : No 'split_by' option specified. Not dividing CPUs up by topology.
2022-06-14 04:28:56: python.d ERROR: turbostat[turbostat] : Could not find any CPUs in turbostat dump
2022-06-14 04:28:56: python.d INFO: plugin[main] : turbostat[turbostat] : check failed
2022-06-14 04:28:56: python.d INFO: plugin[main] : no jobs to serve
2022-06-14 04:28:56: python.d INFO: plugin[main] : exiting from main...
```

```
$ turbostat --Dump
turbostat version 20.09.30 - Len Brown <lenb@kernel.org>
CPUID(0): AuthenticAMD 0x10 CPUID levels; 0x80000023 xlevels; family:model:stepping 0x19:21:0 (25:33:0)
CPUID(1): SSE3 MONITOR - - - TSC MSR - HT -
CPUID(6): APERF, No-TURBO, No-DTS, No-PTM, No-HWP, No-HWPnotify, No-HWPwindow, No-HWPepp, No-HWPpkg, No-EPB
CPUID(7): No-SGX
RAPL: 234 sec. Joule Counter Range, at 280 Watts
turbostat: fopen /dev/cpu_dma_latency
: Permission denied
current_driver: acpi_idle
current_governor: menu
current_governor_ro: menu
cpu12: POLL: CPUIDLE CORE POLL IDLE
cpu12: C1: ACPI FFH MWAIT 0x0
cpu12: C2: ACPI IOPORT 0x414
cpu12: cpufreq driver: acpi-cpufreq
cpu12: cpufreq governor: performance
cpufreq boost: 1
cpu0: MSR_RAPL_PWR_UNIT: 0x000a1003 (0.125000 Watts, 0.000015 Joules, 0.000977 sec.)

$ turbostat -v
turbostat version 20.09.30 - Len Brown <lenb@kernel.org>

$ turbostat
turbostat version 20.09.30 - Len Brown <lenb@kernel.org>
CPUID(0): AuthenticAMD 0x10 CPUID levels; 0x80000023 xlevels; family:model:stepping 0x19:21:0 (25:33:0)
CPUID(1): SSE3 MONITOR - - - TSC MSR - HT -
CPUID(6): APERF, No-TURBO, No-DTS, No-PTM, No-HWP, No-HWPnotify, No-HWPwindow, No-HWPepp, No-HWPpkg, No-EPB
CPUID(7): No-SGX
RAPL: 234 sec. Joule Counter Range, at 280 Watts
turbostat: fopen /dev/cpu_dma_latency
: Permission denied
current_driver: acpi_idle
current_governor: menu
current_governor_ro: menu
cpu12: POLL: CPUIDLE CORE POLL IDLE
cpu12: C1: ACPI FFH MWAIT 0x0
cpu12: C2: ACPI IOPORT 0x414
cpu12: cpufreq driver: acpi-cpufreq
cpu12: cpufreq governor: performance
cpufreq boost: 1
cpu0: MSR_RAPL_PWR_UNIT: 0x000a1003 (0.125000 Watts, 0.000015 Joules, 0.000977 sec.)
Core    CPU     Avg_MHz Busy%   Bzy_MHz TSC_MHz IRQ     POLL    C1      C2      POLL%   C1%     C2%     CorWatt PkgWatt
-       -       6       0.15    3846    3693    4673    29      82      4525    0.00    0.02    99.85   0.59    18.61
0       0       6       0.15    3804    3693    166     0       21      160     0.00    0.09    99.78   0.04    18.61
0       12      1       0.01    3688    3693    34      0       0       33      0.00    0.00    99.99
1       1       2       0.05    3779    3693    80      0       2       76      0.00    0.00    99.96   0.13
1       13      15      0.42    3642    3693    948     0       2       945     0.00    0.00    99.65
2       2       2       0.04    3627    3693    81      0       0       79      0.00    0.00    99.97   0.04
2       14      6       0.17    3849    3693    112     0       0       94      0.00    0.00    99.85
3       3       4       0.11    3675    3693    116     0       0       113     0.00    0.00    99.90   0.08
3       15      30      0.69    4382    3693    395     0       0       381     0.00    0.00    99.35
4       4       10      0.27    3630    3693    394     0       3       387     0.00    0.00    99.76   0.08
4       16      4       0.11    3614    3693    254     0       0       253     0.00    0.00    99.92
5       5       3       0.07    3620    3693    164     0       15      153     0.00    0.01    99.93   0.03
5       17      3       0.08    3604    3693    141     0       0       138     0.00    0.00    99.94
8       6       3       0.08    3611    3693    111     0       0       106     0.00    0.00    99.93   0.04
8       18      9       0.22    3961    3693    217     0       0       210     0.00    0.00    99.80
9       7       6       0.15    3730    3693    130     0       1       122     0.00    0.00    99.87   0.03
9       19      3       0.07    3616    3693    76      0       0       76      0.00    0.00    99.94
10      8       3       0.08    3828    3693    106     0       0       102     0.00    0.00    99.93   0.02
10      20      1       0.03    3607    3693    59      0       0       58      0.00    0.00    99.98
11      9       6       0.15    3685    3693    165     0       1       160     0.00    0.00    99.86   0.06
11      21      9       0.25    3648    3693    573     29      37      540     0.01    0.33    99.46
12      10      5       0.12    3746    3693    108     0       0       103     0.00    0.00    99.89   0.02
12      22      2       0.04    4052    3693    61      0       0       60      0.00    0.00    99.97
13      11      4       0.12    3795    3693    144     0       0       140     0.00    0.00    99.90   0.03
13      23      1       0.03    3593    3693    38      0       0       36      0.00    0.00    99.98
^CCore  CPU     Avg_MHz Busy%   Bzy_MHz TSC_MHz IRQ     POLL    C1      C2      POLL%   C1%     C2%     CorWatt PkgWatt
-       -       7       0.18    3866    3693    1979    22      25      1929    0.00    0.03    99.82   0.49    17.94
0       0       3       0.10    3662    3693    69      0       0       68      0.00    0.00    99.92   0.02    17.94
0       12      1       0.02    3591    3693    20      0       0       20      0.00    0.00    99.99
1       1       1       0.03    3592    3693    26      0       1       25      0.00    0.00    99.98   0.10
1       13      31      0.80    3875    3693    399     0       0       397     0.00    0.00    99.30
2       2       2       0.05    3605    3693    41      0       0       41      0.00    0.00    99.97   0.03
2       14      8       0.22    3829    3693    90      0       0       88      0.00    0.00    99.81
3       3       4       0.11    3669    3693    45      0       0       45      0.00    0.00    99.91   0.13
3       15      42      0.96    4348    3693    163     0       0       158     0.00    0.00    99.10
4       4       9       0.26    3606    3693    143     0       0       142     0.00    0.00    99.79   0.06
4       16      4       0.12    3615    3693    106     0       0       105     0.00    0.00    99.92
5       5       3       0.08    3622    3693    68      0       2       65      0.00    0.11    99.84   0.04
5       17      6       0.15    3670    3693    70      0       0       69      0.00    0.00    99.88
8       6       3       0.08    3608    3693    52      3       7       42      0.00    0.32    99.62   0.02
8       18      9       0.23    3869    3693    91      0       0       89      0.00    0.00    99.79
9       7       6       0.17    3602    3693    64      0       0       62      0.00    0.00    99.85   0.02
9       19      2       0.05    3621    3693    32      0       0       31      0.00    0.00    99.96
10      8       2       0.07    3617    3693    38      0       0       38      0.00    0.00    99.95   0.01
10      20      2       0.05    3589    3693    34      0       1       33      0.00    0.00    99.97
11      9       7       0.19    3741    3693    87      7       5       76      0.00    0.32    99.51   0.04
11      21      9       0.23    3712    3693    180     12      8       176     0.02    0.00    99.81
12      10      2       0.06    3598    3693    40      0       0       40      0.00    0.00    99.96   0.00
12      22      2       0.07    3597    3693    37      0       0       37      0.00    0.00    99.95
13      11      3       0.08    3625    3693    59      0       0       58      0.00    0.00    99.96   0.01
13      23      3       0.09    3597    3693    25      0       1       24      0.00    0.00    99.93
$

```

tycho:
@SteveBox0 Not sure what's wrong in your case. I mean, there are some things that suggest you need to run as root:
```
turbostat: fopen /dev/cpu_dma_latency
: Permission denied
```

But basically the output *should* look something like this with `turbostat --Dump`:
```
t 0x5632783d8490, c 0x5632783da930, p 0x5632783d5a10
CPU: 0 flags 0x6
TSC: 0007EA57D2AD6EC6
aperf: 000017DBCC5818DD
mperf: 0000164942F1CF0C
c1: 0000000000000000
IPC: 1299
IRQ: 33188736
core: 0
c3: 0000000000000000
c6: 0000000000000000
c7: 0000000000000000
DTS: 0C
Joules: 3A204259
mc6_us: 0000000000000000
package: 0
Weighted cores: 0000000000000000
Any cores: 0000000000000000
Any GFX: 0000000000000000
CPU + GFX: 0000000000000000
pc2: 0000000000000000
pc8: 0000000000000000
pc9: 0000000000000000
pc10: 0000000000000000
cpu_lpi: 0000000000000000
sys_lpi: 0000000000000000
Joules PKG: 0
Joules COR: 0
Joules GFX: 0
Joules RAM: 0
Throttle PKG: 0
Throttle RAM: 0
PTM: 0C

t 0x5632783d85c8, c 0x5632783da930, p 0x5632783d5a10
CPU: 12 flags 0x0
TSC: 0007EA57D2AF509A
aperf: 000016329A3BB987
mperf: 000014AB9BA879C8
c1: 0000000000000000
IPC: 1299
... etc ...
```
You could try using `strace` to figure out what it's trying to look at and not finding with `--Dump`. Maybe `/dev/cpu/*/msr` are missing and you need the `msr` kernel module loaded?

SteveBox0:
@tycho 
Thank you for your quick reply.

Running as root only changes the output by a very small amount.
```
~# turbostat --Dump
turbostat version 20.09.30 - Len Brown <lenb@kernel.org>
CPUID(0): AuthenticAMD 0x10 CPUID levels; 0x80000023 xlevels; family:model:stepping 0x19:21:0 (25:33:0)
CPUID(1): SSE3 MONITOR - - - TSC MSR - HT -
CPUID(6): APERF, No-TURBO, No-DTS, No-PTM, No-HWP, No-HWPnotify, No-HWPwindow, No-HWPepp, No-HWPpkg, No-EPB
CPUID(7): No-SGX
RAPL: 234 sec. Joule Counter Range, at 280 Watts
/dev/cpu_dma_latency: 100 usec (constrained)
current_driver: acpi_idle
current_governor: menu
current_governor_ro: menu
cpu13: POLL: CPUIDLE CORE POLL IDLE
cpu13: C1: ACPI FFH MWAIT 0x0
cpu13: C2: ACPI IOPORT 0x414
cpu13: cpufreq driver: acpi-cpufreq
cpu13: cpufreq governor: performance
cpufreq boost: 1
cpu0: MSR_RAPL_PWR_UNIT: 0x000a1003 (0.125000 Watts, 0.000015 Joules, 0.000977 sec.)
```

msr seems to be loaded from a quick look. 
```
~# lsmod | grep "msr             "
msr                    16384  0

~# ls -l /dev/cpu/*/msr
crw-r----- 1 root msr 202,  0 Jun 14 05:54 /dev/cpu/0/msr
crw-r----- 1 root msr 202, 10 Jun 14 05:54 /dev/cpu/10/msr
crw-r----- 1 root msr 202, 11 Jun 14 05:54 /dev/cpu/11/msr
crw-r----- 1 root msr 202, 12 Jun 14 05:54 /dev/cpu/12/msr
crw-r----- 1 root msr 202, 13 Jun 14 05:54 /dev/cpu/13/msr
crw-r----- 1 root msr 202, 14 Jun 14 05:54 /dev/cpu/14/msr
crw-r----- 1 root msr 202, 15 Jun 14 05:54 /dev/cpu/15/msr
crw-r----- 1 root msr 202, 16 Jun 14 05:54 /dev/cpu/16/msr
crw-r----- 1 root msr 202, 17 Jun 14 05:54 /dev/cpu/17/msr
crw-r----- 1 root msr 202, 18 Jun 14 05:54 /dev/cpu/18/msr
crw-r----- 1 root msr 202, 19 Jun 14 05:54 /dev/cpu/19/msr
crw-r----- 1 root msr 202,  1 Jun 14 05:54 /dev/cpu/1/msr
crw-r----- 1 root msr 202, 20 Jun 14 05:54 /dev/cpu/20/msr
crw-r----- 1 root msr 202, 21 Jun 14 05:54 /dev/cpu/21/msr
crw-r----- 1 root msr 202, 22 Jun 14 05:54 /dev/cpu/22/msr
crw-r----- 1 root msr 202, 23 Jun 14 05:54 /dev/cpu/23/msr
crw-r----- 1 root msr 202,  2 Jun 14 05:54 /dev/cpu/2/msr
crw-r----- 1 root msr 202,  3 Jun 14 05:54 /dev/cpu/3/msr
crw-r----- 1 root msr 202,  4 Jun 14 05:54 /dev/cpu/4/msr
crw-r----- 1 root msr 202,  5 Jun 14 05:54 /dev/cpu/5/msr
crw-r----- 1 root msr 202,  6 Jun 14 05:54 /dev/cpu/6/msr
crw-r----- 1 root msr 202,  7 Jun 14 05:54 /dev/cpu/7/msr
crw-r----- 1 root msr 202,  8 Jun 14 05:54 /dev/cpu/8/msr
crw-r----- 1 root msr 202,  9 Jun 14 05:54 /dev/cpu/9/msr
```

Here is an strace for "turbostat --Dump". I am running Kernel 5.18.3 on an AMD CPU and remember some msr related security changes in 5.* but dont think it should impact this information. 
https://pastebin.com/0Jdu08ss

tycho:
@SteveBox0 Maybe try compiling the latest Turbostat from source? It's in the linux kernel tree under tools/. I seem to recall something a few months back where the `--Dump` argument wasn't being handled correctly or something like that.

SteveBox0:
@tycho 
Same output excluding the version changed from "20.09.30" to "21.05.04". This tool was originally created for Intel CPUs maybe  the output is not the same for AMD CPUs.  

```
# ./turbostat --Dump
turbostat version 21.05.04 - Len Brown <lenb@kernel.org>
CPUID(0): AuthenticAMD 0x10 CPUID levels
CPUID(1): family:model:stepping 0x19:21:0 (25:33:0) microcode 0x0
CPUID(0x80000000): max_extended_levels: 0x80000023
CPUID(1): SSE3 MONITOR - - - TSC MSR - HT -
CPUID(6): APERF, No-TURBO, No-DTS, No-PTM, No-HWP, No-HWPnotify, No-HWPwindow, No-HWPepp, No-HWPpkg, No-EPB
CPUID(7): No-SGX
RAPL: 234 sec. Joule Counter Range, at 280 Watts
/dev/cpu_dma_latency: 100 usec (constrained)
current_driver: acpi_idle
current_governor: menu
current_governor_ro: menu
cpu14: POLL: CPUIDLE CORE POLL IDLE
cpu14: C1: ACPI FFH MWAIT 0x0
cpu14: C2: ACPI IOPORT 0x414
cpu14: cpufreq driver: acpi-cpufreq
cpu14: cpufreq governor: performance
cpufreq boost: 1
cpu0: MSR_RAPL_PWR_UNIT: 0x000a1003 (0.125000 Watts, 0.000015 Joules, 0.000977 sec.)
```

The turbostat seems to be producing the data required to graph anyway. What would need to be changed to the python script for it to work with this configuration? E.g. turbostat -q -n 1

```
Core  CPU     Avg_MHz Busy%   Bzy_MHz TSC_MHz IPC     IRQ     POLL    C1      C2      POLL%   C1%     C2%     CorWatt PkgWatt
-       -       41      0.93    4430    3693    1.21    6480    109     511     2168    0.00    0.16    98.94   1.92    22.51
0       0       9       0.24    3685    3693    0.27    180     0       2       174     0.00    0.00    99.80   0.36    22.51
0       12      136     2.97    4581    3693    1.23    808     18      61      79      0.03    1.04    96.01
1       1       31      0.69    4543    3693    1.69    155     0       0       35      0.00    0.00    99.33   0.10
1       13      6       0.15    3627    3693    0.63    60      0       0       59      0.00    0.00    99.87
2       2       13      0.34    3868    3693    0.79    111     0       0       104     0.00    0.00    99.70   0.05
2       14      3       0.08    4019    3693    0.95    33      0       1       32      0.00    0.00    99.94
3       3       5       0.14    3626    3693    0.36    50      0       0       50      0.00    0.00    99.88   0.07
3       15      17      0.38    4406    3693    1.18    39      0       0       34      0.00    0.00    99.64
4       4       39      0.87    4465    3693    1.61    162     0       0       51      0.00    0.00    99.15   0.11
4       16      10      0.25    4062    3693    0.83    49      0       0       45      0.00    0.00    99.78
5       5       149     3.23    4601    3693    1.42    985     14      79      85      0.01    1.35    95.45   0.41
5       17      19      0.49    3865    3693    0.71    172     0       0       166     0.00    0.00    99.55
8       6       10      0.28    3666    3693    0.27    127     18      4       120     0.02    0.00    99.75   0.06
8       18      5       0.14    3673    3693    0.52    60      0       0       46      0.00    0.00    99.87
9       7       8       0.22    3653    3693    0.22    179     0       0       177     0.00    0.00    99.82   0.17
9       19      90      2.13    4226    3693    1.85    175     0       0       153     0.00    0.00    97.92
10      8       121     2.62    4622    3693    0.86    725     2       204     86      0.00    0.14    97.27   0.27
10      20      128     2.75    4662    3693    1.18    969     20      76      75      0.03    0.37    96.90
11      9       1       0.03    3637    3693    0.26    20      0       0       20      0.00    0.00    99.98   0.06
11      21      20      0.54    3783    3693    0.46    250     1       1       240     0.00    0.00    99.52
12      10      9       0.25    3775    3693    0.90    89      22      5       77      0.01    0.11    99.66   0.21
12      22      113     2.47    4584    3693    1.33    898     14      71      114     0.02    0.56    96.99
13      11      34      0.76    4476    3693    0.90    111     0       6       75      0.00    0.26    99.00   0.06
13      23      10      0.26    3690    3693    0.88    73      0       1       71      0.00    0.00    99.76
```

Also, looks like both version of turbostat should support the AMD CPU.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=33eb82251af9be47a625ca1578f44e596a3a0ca9

tycho:
Oh I forgot that I am carrying a patch in the turbostat that I use, since --Dump has been broken on AMD for a while:
https://git.uplinklabs.net/steven/ec2-packages/src/branch/master/linux-tools/unbreak-turbostat-dump.patch
```diff
diff --git a/tools/power/x86/turbostat/turbostat.c b/tools/power/x86/turbostat/turbostat.c
index 47d3ba895d6d..1b2c0d9e968a 100644
--- a/tools/power/x86/turbostat/turbostat.c
+++ b/tools/power/x86/turbostat/turbostat.c
@@ -2225,9 +2225,8 @@ int get_counters(struct thread_data *t, struct core_data *c, struct pkg_data *p)
 		p->rapl_dram_perf_status = msr;
 	}
 	if (do_rapl & RAPL_AMD_F17H) {
-		if (get_msr_sum(cpu, MSR_PKG_ENERGY_STAT, &msr))
-			return -13;
-		p->energy_pkg = msr;
+		/* Unbreaks turbostat --Dump */
+		p->energy_pkg = 0;
 	}
 	if (DO_BIC(BIC_PkgTmp)) {
 		if (get_msr(cpu, MSR_IA32_PACKAGE_THERM_STATUS, &msr))
```

SteveBox0:
@tycho 
Thanks this works but breaks PkgWatt that you could see was working in 5.18.

Using 5.19 turbostat on 5.18 i was able to get both to work correctly. 
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/power/x86/turbostat/turbostat.c?h=v5.19-rc2

PR #2715, opened 2017-09-10
cpuidle.chart.py: only tickle known-to-be-idle CPUs

tycho:
I noticed that the cpuidle.chart.py script run time was spiking whenever all the CPUs were busy:

2017-09-10 14_21_53-tobii highlighter

In retrospect this makes perfect sense. Busy CPUs would be slower to respond to some new thread being scheduled onto them via `sched_setaffinity`.

The fix is pretty straightforward. We only need to tickle CPUs we know are idle -- i.e. those which have counters that aren't changing in `/proc/schedstat`.


ktsaou:
perfect! thanks!

PR #2716, opened 2017-09-10
cpuidle.chart.py: re-read schedstat after tickling idle CPUs

tycho:
Oops, minor mistake. We also need to re-read `/proc/schedstat` once we've awoken the idlers, otherwise we'll have gaps in the data:

2017-09-10 14_56_15-tobii highlighter


ktsaou:
ok. nice!

PR #2742, opened 2017-09-15
Add python module ntp

rda0:
The module **ntp** uses SimpleService class and the command `ntpq -c "readlist 0"` to read `ntpd`'s system variables and a combination of the commands `ntpq -c associations` (to get the current association ids of the peers in `check()`) and the command `ntpq -c "readlist <associd>"` to read the peer variables.

If `ntpd` is restarted these association ids change, which results in returning no more data for peers.

The custom dashboard can be enabled by adding the following line in `netdata.conf`:

    [web]
    custom dashboard_info.js = dashboard_info_custom.ntp.js

The module is disabled by default. To enable it, edit `python.d.conf` and set to:

    ntp: yes

### DIFF
--- a/python.d/ntp.chart.py
+++ b/python.d/ntp.chart.py
@@ -3,8 +3,8 @@
 # Author: Sven Mäder (rda)
 
 from base import SimpleService
-from subprocess import Popen, PIPE
 import re
+import ntpq
 
 # default module values
 update_every = 10
@@ -22,21 +22,7 @@ ORDER = [
     'precision',
     'stratum',
     'tc',
-    'mintc',
-    'peer_stratum',
-    'peer_precision',
-    'peer_rootdelay',
-    'peer_rootdisp',
-    'peer_hmode',
-    'peer_pmode',
-    'peer_hpoll',
-    'peer_ppoll',
-    'peer_headway',
-    'peer_offset',
-    'peer_delay',
-    'peer_dispersion',
-    'peer_jitter',
-    'peer_xleave'
+    'mintc'
 ]
 
 CHARTS = {
@@ -100,9 +86,7 @@ CHARTS = {
 class Service(SimpleService):
     def __init__(self, configuration=None, name=None):
         SimpleService.__init__(self, configuration=configuration, name=name)
-        self.ntpq = self.find_binary('ntpq')
-        self.associd_list = list()
-        self.peer_dict = dict()
+        self.sock, self.sockaddr = ntpq.get_socket('127.0.0.1')
         self.rgx_sys = re.compile(
                 r'.*stratum=(?P<stratum>[0-9.-]+)'
                 r'.*precision=(?P<precision>[0-9.-]+)'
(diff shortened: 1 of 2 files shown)


l2isbad:
Super heavy to be merged.
N number of Popen calls (depends on number of peers) per iteration.

```
lgz@lgz-homepc[~]ntpq -c associations

ind assid status  conf reach auth condition  last_event cnt
===========================================================
  1 39748  931a   yes   yes  none   outlier    sys_peer  1
  2 39749  964a   yes   yes  none  sys.peer    sys_peer  4
  3 39750  944a   yes   yes  none candidate    sys_peer  4
  4 39751  941a   yes   yes  none candidate    sys_peer  1
```

I see there is [python-ntplib](https://pypi.python.org/pypi/ntplib/) in pure python.

And [ntpq](https://gist.github.com/pcn/8f65b81a9157f8dc649e2834b31fd223) implementaion  in python (which uses ntplib).

The problem is lack of `python3-ntplib` in debian old stable repo (debian 8).

ktsaou:
I agree with @l2isbad 
@rda0 we have to be very careful. This plugin will introduce a significant overhead to the system.

rda0:
Hi @l2isbad and @ktsaou 

Thank you for your feedback and this fantastic monitoring tool.

I agree it is heavy, but we were in need of those graphs to debug ntp on some servers, so this was acceptable for us.

I made a new version using the libraries mentioned by @l2isbad, just to see if I get it working.
The module only gets the system variables at the moment.
Sorry, my python knowledge and especially using libraries is very limited.
I could not find out how to use `python3-ntplib` instead of `python-ntplib` in the module.

Ferroin:
I think I might have found an alternative method for gathering this data that doesn't require third-party libraries or calling out to CLI tools.

The NTP daemon can be configured to save a wide variety of statistics during normal operation, and statistics on the peers are one of the options for what to save.  Quoting the `ntp.conf` manual page directly:
```
Enables recording of peer statistics information.  This includes statistics records of all peers of a NTP
server and of special signals, where present and configured.  Each valid update appends a line of the
following form to the current element of a file generation set named peerstats:
    48773 10847.650 127.127.4.1 9714 -0.001605376 0.000000000 0.001424877 0.000958674

The first two fields show the date (Modified Julian Day) and time (seconds and fraction past UTC
midnight).  The next two fields show the peer address in dotted-quad notation and status, respectively.
The status field is encoded in hex in the format described in Appendix A of the NTP specification RFC
1305.  The final four fields show the offset, delay, dispersion and RMS jitter, all in seconds.
```

It doesn't say what frequency it actually writes the stats out at (which is interesting, since all the other types explicitly list the intervals), but I suspect it dumps a line for each peer every time it talks to that peer.  It obviously doesn't give as much info as the `readlist` command for `ntpq` does, but I think if you also add the `rawstats` statistics collection too, it might cover everything you need (`rawstats` gives you info on the 4 timestamps for every message).

The downside to this approach is that writing out the statistics is disabled completely by default (and has no automatic cleanup of old files _and_ doesn't appear to play nice with log rotation), and the path is largely user configurable, so it would require some setup in both `ntp.conf` and `netdata.conf` to make this work.

rda0:
Thanks a lot for the feedback @Ferroin 

It's really funny, because I just started working on this today after months of inactivity. I deleted my netdata fork some time ago, when I received feedback from @l2isbad and @ktsaou, that the module was far too heavy. Therefore I created a [third party module repo](https://github.com/rda0/netdata-ntp) for everyone who might be still interested in this.

The new version from today does not need any libraries. It just connects to `ntpd` using an UDP socket to read the system variables (equivalent to `ntpq -c readvar`), now with minimal overhead (hopefully).
I tried to use `SocketService` to get the data, but I could not find out how I could connect to an UDP socket. If anyone could shed some light on this, that would be great.

However, I think in some situations, especially for debugging of time critical services, it would be interesting to also have some data of the other peers as well. Maybe I can extend the current version to get this data in the same way as I did for the system variables. Otherwise I will definitely have a look at what you proposed.

In which man page exactly did you find this information? I could not find it on debian stretch, but I found it [here](http://doc.ntp.org/4.2.4/monopt.html).

Please let me know if I should create another pull request..

l2isbad:
> The new version from today does not need any libraries.

sounds great

> I tried to use SocketService to get the data, but this is not possible because ntpd only listens on UDP and the netdata plugin only implements TCP.

So we have to implement it in ```SocketService```?
I think the problem is https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/FrameworkServices/SocketService.py#L118
```socket.SOCK_STREAM``` is hardcoded

Have you tried to change the value to ```socket.SOCK_DGRAM```. Does it work after this change?

Anyway feel free to make a PR  👍

**Comment by rda0** [comment]
> So we have to implement it in `SocketService`?

That would be great.

> Have you tried to change the value to `socket.SOCK_DGRAM`?. Does it work after this change?

Yeas it does :)

Something like this would work:

```
diff --git a/python.d/python_modules/bases/FrameworkServices/SocketService.py b/python.d/python_modules/bases/FrameworkServices/SocketService.py
index 90631df..ae8127d 100644
--- a/python.d/python_modules/bases/FrameworkServices/SocketService.py
+++ b/python.d/python_modules/bases/FrameworkServices/SocketService.py
@@ -14,6 +14,7 @@ class SocketService(SimpleService):
         self.host = 'localhost'
         self.port = None
         self.unix_socket = None
+        self.dgram_socket = False
         self.request = ''
         self.__socket_config = None
         self.__empty_request = "".encode()
@@ -115,7 +116,11 @@ class SocketService(SimpleService):
                 if self.__socket_config is not None:
                     self._connect2socket()
                 else:
-                    for res in socket.getaddrinfo(self.host, self.port, socket.AF_UNSPEC, socket.SOCK_STREAM):
+                    if self.dgram_socket:
+                        sock_type = socket.SOCK_DGRAM
+                    else:
+                        sock_type = socket.SOCK_STREAM
+                    for res in socket.getaddrinfo(self.host, self.port, socket.AF_UNSPEC, sock_type):
                         if self._connect2socket(res):
                             break
```

l2isbad:
> The downside to this approach is that writing out the statistics is disabled completely by default (and has no automatic cleanup of old files and doesn't appear to play nice with log rotation), and the path is largely user configurable, so it would require some setup in both ntp.conf

That's how `smartd_log` works. I thinks it's acceptable.

rda0:
I made a new version of [ntp.chart.py](https://github.com/rda0/netdata-ntp/blob/master/python.d/ntp.chart.py). It now also includes the peer variables. This should work with the default `ntp.conf`. All data is read via UDP socket.

It is a bit a hassle with the NTP Control Message protocol. I need to make several UDP requests, depending on the number of peers. But this is all done in `check()`. After that in `_get_data()` it only makes 2 UDP requests. It always gets the systemvars but only the peervars for one peer in each invocation of `_get_data()`. It looks like this works, as netdata shows no gaps in the graphs. This should be no problem, because there will not be new useful data every second. The minimal poll exponent (`minpoll`) accepted by `ntpd` is `3` (8s), the default is `6` (64s). The poll exponent increases over time of up to `maxpoll`, which is by default set to `10` (1024s).

This version at the moment uses `SimpleService` with its own `_get_raw_data()` method, but it should be easily adaptable to use `SocketService` with the patch mentioned erlier.

What do you think?

l2isbad:
> easily adaptable to use SocketService with the patch mentioned erlier

please do

rda0:
New PR #3421 added.

ktsaou:
ok, I am closing this in favor of #3421

PR #2749, opened 2017-09-17
prevent duplicate chart names when streaming; 

### LINKED ISSUE #2718: go_expvar issues/improvements
Hello,

I have found two issues with my go_expvar python.d plugin that I would like to fix, but I need some pointers:

1) Charting very small values

I have recently started measuring response times in my app, which are in the microsecond range. The app is returning them as seconds, like this:

`"input.httplisten9448.request_duration.p50": 2.5883469939240174e-05`

It then gets rounded down to 0 during processing. If a metric is set to the `float` type in go_expvar plugin, the plugin multiplies the value by 1000 and then sets the line divisor to 1000, however, this is not enough for these values. I was thinking of multypling the values by let's say 10^12 in the plugin, and then setting the line divisor accordingly (if it's not overriden by the user). Would that be okay or is there a better way to go about this? Maybe let the user supply the internal multiplier value by config?

Let me know what do you think.

2) Duplicated chart ID's

I'm kind of stumped by this - for some reason, one of the dot-separated parts in the chart ID is getting duplicated for some charts, which then messes up other metrics backend (influx in my case).

This is the chart info from netdata dashboard for a chart that's having this issue:

`runtime: number of goroutines (go_expvar_stblogd.go_expvar_stblogd.num_goroutines)`

Another chart, (which is okay):

`pipeline: message rate (go_expvar_stblogd.message_rate)`

But, when I run the `python.d` plugin directly, the chart and dimension config looks okay:

```
CHART go_expvar_stblogd.runtime_goroutines num_goroutines 'runtime: number of goroutines' goroutines runtime expvar.runtime.goroutines line 60007 1
DIMENSION runtime_goroutines goroutines absolute 1 1

CHART go_expvar_stblogd.message_rate message_rate 'pipeline: message rate' req/s pipeline expvar.pipeline.reqrate line 60008 1
DIMENSION message_rate_input input incremental 1 1
DIMENSION message_rate_filter filter incremental 1 1
DIMENSION message_rate_output output incremental 1 1

...

BEGIN go_expvar_stblogd.runtime_goroutines 999975
SET runtime_goroutines = 752
END
BEGIN go_expvar_stblogd.message_rate 999975
SET message_rate_input = 129777748
SET message_rate_filter = 131840251
SET message_rate_output = 131840232
END
```

I have no idea where the duplicate part of ID is coming from.

@ktsaou or @l2isbad, I'd appreciate any help with either of these issues.

Thanks

ktsaou:
fixes #2718

### DIFF
--- a/src/plugins_d.c
+++ b/src/plugins_d.c
@@ -217,32 +217,46 @@ inline size_t pluginsd_process(RRDHOST *host, struct plugind *cd, FILE *fp, int 
             st = NULL;
         }
         else if(likely(hash == CHART_HASH && !strcmp(s, PLUGINSD_KEYWORD_CHART))) {
-            int noname = 0;
             st = NULL;
 
-            if((words[1]) != NULL && (words[2]) != NULL && strcmp(words[1], words[2]) == 0)
-                noname = 1;
-
-            char *type = words[1];
-            char *id = NULL;
-            if(likely(type)) {
-                id = strchr(type, '.');
-                if(likely(id)) { *id = '\0'; id++; }
-            }
-            char *name = words[2];
-            char *title = words[3];
-            char *units = words[4];
-            char *family = words[5];
-            char *context = words[6];
-            char *chart = words[7];
-            char *priority_s = words[8];
+            char *type           = words[1];
+            char *name           = words[2];
+            char *title          = words[3];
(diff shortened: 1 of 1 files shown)


l2isbad:
fixed

Issue #2636, opened 2017-08-27
Support access control lists for controlling accesses to netdata


ktsaou:
For the web API port, support a simple pattern matching for filtering the client IPs accessing netdata.

This should eliminate the need for binding netdata to multiple ports and extensive firewalling.

Supported access list patterns:

- [x] `[web].allow connections from` to limit the clients accessing everything about netdata.

For the allowed hosts, the following additional checks should be implemented:

- [x] `netdata.conf`: `[web].allow badges from` to limit the clients accessing badges
- [x] `netdata.conf`: `[web].allow streaming from` to limit the clients allowed to stream
- [x] `netdata.conf`: `[web].allow netdata.conf from` to limit the clients that can get `netdata.conf`
- [x] `netdata.conf`: `[registry].allow from` to limit the clients allowed to access the dashboard
- [x] `stream.conf`: `[API_KEY].allow from` to limit the clients allowed to push metrics for a specific API KEY
- [x] `stream.conf`: `[MACHINE_GUID].allow from` to limit the clients allowed to push metrics for a specific machine

All the above will be netdata simple patterns (i.e. strings with `*` as wildcard and `!` for negative matches).

ktsaou:
Wiki updates:

- [x] nginx: https://github.com/firehol/netdata/wiki/Running-behind-nginx#limit-direct-access-to-netdata

PR #2750, opened 2017-09-17
add web API port access list

ktsaou:
This PR adds support for applying access control to the web API ports.

This access control is global, meaning that it is applied when a client is attempting to connect. For filtered out clients, will not be able the send even a single byte to netdata.

It is configured in `netdata.conf`, like this:

```
[web]
    global API allow from = 127.* ::1 *
```

The default is shown above. It is redundant, meaning that `*` matches `127.*` and `::1`, but it is a good default for customization, since to allow `10.*` the users should change it to `127.* ::1 10.*`, to allow both localhost and `10.*`.

When the access list is denying access, it logs `DENIED ACCESS to client 'IP'` to `error.log`.

> **IMPORTANT**: This setting is applied to all the functions of all web ports netdata listens. This includes streaming, but it does not include statsd.

#2636

---

also reformatted `access.log`. Now it also logs streaming operations.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -83,6 +83,8 @@ void web_server_config_options(void) {
     web_x_frame_options = config_get(CONFIG_SECTION_WEB, "x-frame-options response header", "");
     if(!*web_x_frame_options) web_x_frame_options = NULL;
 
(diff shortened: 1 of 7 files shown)


mikeweilgart:
I see a comment in the config file from the latest netdata build:

>        # option 'global API allow from' is not used.

What should I make of this?  Where is the ACL functionality now?

mikeweilgart:
Is it now supposed to be "allow connections from" instead?

ktsaou:
yes

PR #2751, opened 2017-09-17
Deploy using new firehol-automation user

philwhineray:
Automated release uploads from Travis will subsequently be associated with this user, rather than my account.


ktsaou:
thanks!

PR #2847, opened 2017-10-05
Units scaling and multiple timezones

### LINKED ISSUE #2698: timezone setting on the dashboard
@glatapoui suggested at #1544 to add an option to switch timezone at the browser, as a feature for easily correlating netdata views with others sources (e.g. log files).

This is a nice one.

### LINKED ISSUE #2745: allow user settings to control the units of the charts at the dashboard
netdata has hard-coded units. So:

- disk I/O is in `kilobytes/s`
- disk size is in `GB`
- memory size is in `MB`
- network bandwidth in is `kilobits/s`
- temperatures in Celcius

The ultimate goal is to allow users to change these on the dashboard.

A simple solution would be to provide a `data-value-formatter="function(){ ... }"` at the charts, to allow the caller overwrite the presentation of the numbers.

To achieve a higher abstraction, another setting should be supported: `data-convert-units-to="wanted units"`. This should automatically change the formatter function and the units of the chart. It could also apply some error checking to ensure the wanted units can be applied to this chart.

Then, on the dashboard, a new settings tab should be added to allow the user set the desired units for each type of metrics.

The above should fix: #2425, #376, #1433, #616

Using the same mechanism arbitrary changes to the presentation of values could be supported. This should fix: #1842 point 6 (uptime) and #2723

ktsaou:
fixes #2745
fixes #2698 

This PR adds:

1. auto-scaling of units at the dashboard
2. units-conversions (e.g. temperatures)
3. time zones conversions at the dashboard

## new dashboard settings

screenshot from 2017-10-10 23-54-01

---

The dashboard does dynamic units conversion, on the fly, like the following. Note that the chart units change, based on the value of the selected dimension:

peek 2017-10-06 22-58

---

With dynamic unit conversion, certain cases may seem awkward. Like this:

image

Custom dashboards can add `data-common-units="NAME"`. All the charts with the same `NAME` will get the same units.

---

Tests to be done:

- [x] test it works for both auto-scaling and original.
- [x] test it works when switching / enabling / disabling dimensions at the legend.
- [x] test it works when zooming in / out the charts.
- [x] test it works when hovering over the chart, after zooming in/out and enabling / disabling dimensions.
- [x] test custom dashboards can overwrite this per chart (`data-desired-units="auto"` or `data-desired-units="original"`).
- [x] test custom dashboards can give their desired scale (`data-desired-units="GB"`, when the chart is by default on `MB`).
- [x] support `data-common-units="NAME"`
- [x] test how it interacts with existing dashboard that have overwritten the units with `data-units="UNITS"`.
- [x] support converting Celsius to Fahrenheit. Another setting should be introduced to control the desired temperature units (or even a new tab called `Units` that all unit conversions should be maintained at).
- [x] support time zone conversions

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -17,6 +17,7 @@ char *netdata_configured_cache_dir   = NULL;
 char *netdata_configured_varlib_dir  = NULL;
 char *netdata_configured_home_dir    = NULL;
 char *netdata_configured_host_prefix = NULL;
+char *netdata_configured_timezone    = NULL;
 
 int enable_ksm = 1;
 
--- a/src/common.h
+++ b/src/common.h
@@ -238,6 +238,7 @@ extern char *netdata_configured_cache_dir;
 extern char *netdata_configured_varlib_dir;
 extern char *netdata_configured_home_dir;
 extern char *netdata_configured_host_prefix;
+extern char *netdata_configured_timezone;
 
 extern void netdata_fix_chart_id(char *s);
 extern void netdata_fix_chart_name(char *s);
--- a/src/main.c
+++ b/src/main.c
@@ -496,6 +496,87 @@ static void get_netdata_configured_variables() {
     get_system_pid_max();
 }
 
+static void get_system_timezone(void) {
+    // avoid flood calls to stat(/etc/localtime)
+    // http://stackoverflow.com/questions/4554271/how-to-avoid-excessive-stat-etc-localtime-calls-in-strftime-on-linux
+    const char *tz = getenv("TZ");
+    if(!tz || !*tz)
+        setenv("TZ", config_get(CONFIG_SECTION_GLOBAL, "TZ environment variable", ":/etc/localtime"), 0);
+
+    char buffer[FILENAME_MAX + 1];
+    const char *timezone = NULL;
+    ssize_t ret;
+
+    // use the TZ variable
+    if(tz && *tz && *tz != ':') {
+        timezone = tz;
+        // info("TIMEZONE: using TZ variable '%s'", timezone);
+    }
+
+    // use the contents of /etc/timezone
+    if(!timezone && !read_file("/etc/timezone", buffer, FILENAME_MAX)) {
+        timezone = buffer;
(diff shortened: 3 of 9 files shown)


Rockstar04:
I just applied this to a few of my servers and it looks and works great. I don't have any custom dashboards, so I can't easily help check off those last boxes though.

ktsaou:
@Rockstar04 thanks!
Did netdata detect the right time zone of your servers properly? Which distros did you check?

Ferroin:
Doesn't appear to break anything on 64-bit Gentoo, but I didn't check the timezone stuff thoroughly. It appears to have correctly detected the time zones on my servers, but they're all set to the same time zone, and I didn't check with a browser on a system using a different time zone.

Rockstar04:
Times and timezones were all correct. All of our servers are 14.04/16.04 Ubuntu and a single  Debian Stretch server.

https://imgur.com/ciHTvDc

ktsaou:
@Rockstar04 what I am interested is the 2 timezones presented here:

image

Especially the `server reported timezone`.

Rockstar04:
capture

America/Chicago is our correct timezone. Side note, is [globa] supposed to be [global] ?

Ferroin:
Did some more checking, and I'm seeing correct times and timezones, even when setting random timezones on the client system.

ktsaou:
> America/Chicago is our correct timezone. Side note, is [globa] supposed to be [global] ?

yeap, that is a typo...

ktsaou:
completed the tests.
everything seems good.

ktsaou:
image

image

ktsaou:
image

image

sagar-chandarana:
@ktsaou please update wiki with some examples at  https://github.com/firehol/netdata/wiki/Custom-Dashboards

ktsaou:
about settings related to units scaling on custom dashboards?

sagar-chandarana:
Yes exactly. `data-desired-units`

ktsaou:
ok, added a lot. This still far from complete...
I also added links to the code that shows per chart-library settings.

If anyone is willing to contribute some documentation in this area, please do so...

PR #2854, opened 2017-10-08
Possible typo? Not sure

JoeBanks13:


ktsaou:
Thanks. But I don't think it is a typo. Changelog is probably a name we use to refer to the changes log of our software.

PR #2857, opened 2017-10-09
add tcp ListenOverflows and ListenDrops chart and alarm

### LINKED ISSUE #2855: detect overflow of TCP LISTEN queue
Hi,
I'd like to propose adding the _TcpExtListenOverflows_ counter to the chart "TCP Handshake Issues".
### Reason behind this:
I'm currently investigating a problem on one of our machines where the server (Java-based) can't keep up with the amount of incoming connections. New connections fill up the LISTEN queue until it finally overflows. The result is that connections get dropped. I'd like to monitor this condition in a chart and add an alarm to this.
For those interested in the background, this post nicely descibes what is behind this counter: https://perfchron.com/2015/12/26/investigating-linux-network-issues-with-netstat-and-nstat/
And here: http://veithen.github.io/2014/01/01/how-tcp-backlog-works-in-linux.html
### More ideas:
I also thought if it makes sense to monitor the length of the LISTEN queue directly (Recv-Q in `netstat -n -a` for LISTEN connections). But a) this would require parsing through all open connections which, as you correctly state in #2169 is not a good idea performance-wise, and b) the length of this queue can be configured by applications so we would not know when it is full.
But I guess a temporary increase of queued connections is not a problem, only when connections start to be dropped massively. That is a condition I'd like to be notified of.

What do you think? Is this an addition that would make sense? And by the way, keep up the good work, I love this tool!

ktsaou:
fixes #2855 

Added a new chart: `ipv4.tcplistenissues` with dimensions `ListenOverflows` and `ListenDrops`.

> This chart detects if any listening TCP socket on the host, has overflown, or it drops connections. This is system-wide: any listening TCP socket, of any application.

The chart will not be shown if these kernel counters are zero. It will be enabled automatically if it is found non-zero at any point (it is collected via `/proc/net/netstat` every second). If you need to enable it even if it is zero, edit netdata.conf and set:

```
[plugin:proc:/proc/net/netstat]
	TCP listen issues = yes
```

2 alarms have been added, one for `ListenOverflows` and one for `ListenDrops` that detect if there is any overflow or drop in the last minute (they run every 10 seconds).

screenshot from 2017-10-09 23-04-05

The alarms will automatically be attached when the chart is active.

### DIFF
--- a/conf.d/Makefile.am
+++ b/conf.d/Makefile.am
@@ -102,6 +102,7 @@ dist_healthconfig_DATA = \
     health.d/softnet.conf \
     health.d/squid.conf \
     health.d/swap.conf \
+    health.d/tcp_listen.conf \
     health.d/tcp_resets.conf \
     health.d/udp_errors.conf \
     health.d/varnish.conf \
--- a/conf.d/health.d/tcp_listen.conf
+++ b/conf.d/health.d/tcp_listen.conf
@@ -1,0 +1,27 @@
+# -----------------------------------------------------------------------------
+# tcp listen sockets issues
+
+   alarm: 1m_ipv4_tcp_listen_overflows
+      on: ipv4.tcplistenissues
+      os: linux
+   hosts: *
+  lookup: sum -60s unaligned absolute of ListenOverflows
+   units: overflows
+   every: 10s
+    crit: $this > 0
+   delay: up 0 down 5m multiplier 1.5 max 1h
+    info: the number of TCP listen socket overflows during the last minute
+      to: sysadmin
+
+   alarm: 1m_ipv4_tcp_listen_drops
+      on: ipv4.tcplistenissues
+      os: linux
+   hosts: *
+  lookup: sum -60s unaligned absolute of ListenDrops
+   units: drops
+   every: 10s
+    crit: $this > 0
+   delay: up 0 down 5m multiplier 1.5 max 1h
+    info: the number of TCP listen socket drops during the last minute
+      to: sysadmin
+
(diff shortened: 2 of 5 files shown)


ktsaou:
@vlvkobal this would also be interesting for FreeBSD / macOS servers.

ardenxu:
@ktsaou   i have enable the config as following , why still no chart for listen overflow, see detail below.

1. config:

[plugin:proc:/proc/net/netstat]
        bandwidth = auto
        input errors = auto
        multicast bandwidth = auto
        broadcast bandwidth = auto
        multicast packets = auto
        broadcast packets = auto
        ECN packets = auto
        **TCP listen issues = yes**
        TCP reorders = auto
        TCP SYN cookies = auto
        TCP out-of-order queue = yes
        TCP connection aborts = yes
        TCP memory pressures = yes
        TCP SYN queue = yes
        TCP accept queue = yes
        filename to monitor = /proc/net/netstat

2. dashboard screen snapshot:

image

thanks for help.

ilyam8:
the option is

> TCP accept queue

https://github.com/netdata/netdata/blob/14622c0478294b2e4477ae373bd85d2012cbd12f/collectors/proc.plugin/proc_net_netstat.c#L136

https://github.com/netdata/netdata/blob/14622c0478294b2e4477ae373bd85d2012cbd12f/collectors/proc.plugin/proc_net_netstat.c#L218-L221

___

Check `Networking Stack`->`tcp` on the dasbhoard it is very likely you have there `TCP Accept Queue Issues (ip.tcp_accept_queue)` chart

ardenxu:
based on code, looks like TCP listen issues has been DEPRECATED,  TCP accept queue is the flag to turn on monitor for tcp queue overflow count. there is no issue here. sorry for any interrupt.

ardenxu:
@liyam8 thanks for echo,  also found out the flag based on code, thanks for echo.

Issue #2382, opened 2017-06-22
memory mode map initialization slow when database is too big


ktsaou:
When the database is too big (several GBs) and `memory mode = map` is used, netdata initializes the whole memory with zeros, which results in all these GBs to be flushed to disk. We should be able to detect these are just initialized data and avoid this duplicate initialization (so the files on disk will not be re-written - they will remain sparse files).

Similarly, when a slave host is offline and reconnects, netdata either initializes all its database to zero, or it fills in a gap (depending on the duration it was offline compared to the duration of the database). Depending on the duration the slave was offline, this again leads to huge amounts of data being flushed to disk.

Ferroin:
For the initialization side of things, had you considered using fallocate() prior to mapping the file?  That would eliminate the need to zero-fill the memory (dependent on what flags you use), and on filesystems that support unwritten extents (almost everything used on Linux these days, as well as ZFS and I think HFS+ too), it would be far faster, while also guaranteeing (unless you're using BTRFS) that you'll be able to write out all the data.

Ferroin:
As a more specific performance example, fallocate() creating a 1G file on BTRFS on an SSD takes about 0.008s on my system, while doing the same by writing out null bytes in 4k chunks takes roughly 1.1s.

ktsaou:
The way it is now, a sparse file is created (just a byte is written at the wanted file size - almost zero time) and then it is `mmap()`'d.

But then netdata detects that the memory mapped from that file, does not contain something it can use (old data), and fills the entire memory block with zeros with `memset()`, which results in huge amounts of data being flushed to disk.

This optimization is about passing back a flag that states that this was a new file, to prevent the call to `memset()`.

The same `memset()` is made if the data loaded do not overlap with the current time-range of the database (ie. if the database is 1 hour long, the last point in the file is older than 1 hour).

Ferroin:
Ah, that makes sense.  It may still be worth having the option to fallocate instead though, since if that data isn't written out, you have no guarantee that any arbitrarily written values until the database history limit won't have errors when trying to write-back (the initial zeroing of the file also makes sure the space is allocated for it).  I don't do enough low-level stuff in C to know how exactly ENOSPC is handled in a write to an mmaped file though, so I may just be worrying about nothing.

GitStoph:
![image](https://user-images.githubusercontent.com/20820226/27540285-4702f24a-5a45-11e7-9578-dc8ba1f49541.png)
[...]

PR #2869, opened 2017-10-11
faster chart gaps; 

ktsaou:
#2382 

This PR inserts gaps into the charts using `memset()` instead of custom loops.


gmosx:
@ktsaou is this deprecated?

cakrit:
@mfundul please review this. If it needs changes, then we close this one and you create a new one.

mfundul:
This is an optimization on the old database code path that is not applicable for `memory mode = dbengine` but it has its usefulness. I say let's keep this code in mind for this sprint since it's related to #6103.

PR #2871, opened 2017-10-12
Fix compilation errors for FreeBSD 12.0

vlvkobal:
Fixes https://github.com/firehol/netdata/pull/1321#issuecomment-333936592

### DIFF
--- a/src/freebsd_sysctl.c
+++ b/src/freebsd_sysctl.c
@@ -79,6 +79,9 @@ struct __vmmeter {
 	u_int v_interrupt_free_min;
 	u_int v_free_severe;
 };
+typedef struct __vmmeter vmmeter_t;
+#else
+typedef struct vmmeter vmmeter_t;
 #endif
 
 // --------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/freebsd_sysctl.c
```diff
@@ -27,6 +27,59 @@
 
 int system_pagesize = PAGE_SIZE;
 int number_of_cpus = 1;
+#if __FreeBSD_version >= 1200029
```
ktsaou: Is this a typo?
They removed the structure after this version?

src/freebsd_sysctl.c
```diff
@@ -856,7 +909,11 @@ int do_system_ram(int update_every, usec_t dt) {
     (void)dt;
     static int mib_active_count[4] = {0, 0, 0, 0}, mib_inactive_count[4] = {0, 0, 0, 0}, mib_wire_count[4] = {0, 0, 0, 0},
                mib_cache_count[4] = {0, 0, 0, 0}, mib_vfs_bufspace[2] = {0, 0}, mib_free_count[4] = {0, 0, 0, 0};
+#if __FreeBSD_version < 1200029
```
ktsaou: You can avoid these `#ifdef` statements with a `typedef`. I suggest to `typedef` the right one at the struct definition, to keep the code clean...

src/freebsd_sysctl.c
vlvkobal: They changed and hid it. They also defined `__vmmeter` as a workaround in `vmstat` utility.

vlvkobal:
Thank you for the hint. I've changed #ifdefs to typedefs.

ktsaou:
merged! thanks!

PR #2875, opened 2017-10-13
Fix compilation errors for Mac OS X 10.5

vlvkobal:
Fixes https://github.com/firehol/netdata/issues/1915#issuecomment-333384278 and https://github.com/firehol/netdata/pull/2850#issuecomment-335258743


ktsaou:
nice!

PR #2879, opened 2017-10-14
extend external plugins to collect variables, plugin and module names

### LINKED ISSUE #2688: allow plugins to push variables to netdata
Allow plugins to push static variables to netdata.

For example, a plugin may be willing to push the max connections allowed, the max memory allowed, etc.

Such netdata variables can be used in netdata alarms and there is already an API that exposes to dashboards.

Internally netdata already supports such variables, though they are not pushed to remote netdata when streaming metrics.

ktsaou:
## `plugin` and `module` parameters to the external plugin API

- [x] Added 2 more parameters to `CHART` plugin lines: `plugin` (parameter No 11) and `module` (parameter No 12). Both are optional. `plugin` will be filled with the name of the plugin run by netdata, if empty. `module` will remain empty.
- [x] expose `plugin` and `module` to the API
- [x] transfer `plugin` and `module` with streaming
- [x] render `plugin` and `module` on the `date` tooltip on the dashboard
- [x] define `plugin` and `module` for all netdata internal modules (linux, freebsd and macos)
- [x] update the wiki

tooltip on the date:
image

panned or zoomed chart out of focus:
image

---

## support custom variables at the external plugins API

Variables will allow the plugins to provide static values to alarms. The variables are host specific so, use names wisely. Do not use `-` (hyphen, minus) in variables. This is an operator in alarm expressions. You can use `.` and `_` though.

Also, its does not matter where in the output these are set. You can mix them with all other commands, although for clarity I suggest to put them at the very beginning (before any `CHART`).

- [x] support `VARIABLE name = value` at the plugins API
- [x] transfer `VARIABLE name = value` while streaming, on every re-connect
- [x] transfer `VARIABLE name = value` while streaming, without a re-connect (i.e. in the middle of the communication, as soon as a variable is set, on slaves and proxies).
- [x] update the wiki

Unfortunately, this change will break compatibility a bit. Newer plugins will not work with older netdata and newer slaves will not be able to stream data to older netdata. The other way around will work though (older plugins and older slaves, will be able to work with newer netdata).

---

## units auto-scaling fixes

- [x] make sure the ylabel of dygraphs does not show wrong units. This was introduced with auto-scaling of units. There were 2 problems:

   i. the units shown at the y-axis were wrong. This has been fixed.
   ii. the units shown at the y-axis were wrong, when a subset of the dimensions were selected. The only viable solution is to hide the y-axis legend of the chart. The reason is that when dimensions are selected, the dashboard uses the chart library to calculate the y-range of the selected ones. But this is done after requesting an update of the chart. So, to render the right units, a re-draw would be necessary. To avoid it, we hide the y-axis label when a subset of the dimensions are selected.

---

## optimizations

- [x] saved a few hundreds of index lookups per second, by re-organizing a bit the code

---

## other bug fixes

- [x] Fixed 2 IPv6 charts that were showing flipped received/sent data.
- [x] Fixed a memory leak that was not freeing custom VARIABLES when hosts are removed (was not significant up to now, because streaming was not pushing variables - so this leak was only observable when netdata exits - i.e. for the database of `localhost`).

---

fixes #2688


ktsaou:
@l2isbad I think this is ready. Now you can pass `plugin`, `module`, and host variables.

ktsaou:
ok, made all possible tests, on linux, with/without streaming, with/without cleaning up obsolete hosts, on freebsd 10.3 and 11.0 on macOS.

This is stable.

ktsaou:
The only problem, is that slaves, proxies and masters will have to be updated all. The best way is to first update the masters, then the proxies, then the slaves.

ktsaou:
hm... and now I think of it, there should be another kind of variables: **chart local variables**.

To illustrate the problem: the way it is now (variables are host-wide), if a netdata monitors 2 mysql severs, there is be no way to setup an alarm template that will use the proper host variable for each mysql server. The only way to setup such alarms in this case, is to manually configure an alarm for each host.

hm...

ktsaou:
I will open a new issue for **chart local variables**.
There are too many changes in this PR, and I think it will slow us down if not merged now.
So, I am merging it.

PR #2886, opened 2017-10-16
support chart custom variables at the API and streaming

### LINKED ISSUE #2884: Support chart-local variables
With PR #2879 we added support for host-global variables.
We should also support chart-local variables.

ktsaou:
fixes #2884 

The `VARIABLE` lines of external plugins, now follow this template:

```
VARIABLE [SCOPE] name = value
```

The optional `SCOPE` can be:

- `GLOBAL` or `HOST` to define the variable at the host level.
- `LOCAL` or `CHART` define the variable at the chart level.

The position of the `VARIABLE` line, sets its scope too.

So, defining a `VARIABLE` before any `CHART`, or between `END` and `BEGIN` (outside any chart), sets `GLOBAL` scope, while defining a `VARIABLE` just after a `CHART` or a `DIMENSION`, or within the `BEGIN` - `END` block of a chart, sets `LOCAL` scope.

~~This is untested currently.~~ tested it.

### DIFF
--- a/src/health.h
+++ b/src/health.h
@@ -412,7 +412,8 @@ extern void rrdcalc_unlink_and_free(RRDHOST *host, RRDCALC *rc);
 extern void rrdcalctemplate_free(RRDCALCTEMPLATE *rt);
 extern void rrdcalctemplate_unlink_and_free(RRDHOST *host, RRDCALCTEMPLATE *rt);
 
-extern int  rrdvar_callback_for_all_variables(RRDHOST *host, int (*callback)(void *rrdvar, void *data), void *data);
+extern int  rrdvar_callback_for_all_host_variables(RRDHOST *host, int (*callback)(void *rrdvar, void *data), void *data);
+extern int rrdvar_callback_for_all_chart_variables(RRDSET *st, int (*callback)(void *rrdvar, void *data), void *data);
 
 #ifdef NETDATA_HEALTH_INTERNALS
 #define RRDVAR_MAX_LENGTH 1024
--- a/src/health.h
+++ b/src/health.h
@@ -11,7 +11,8 @@ typedef enum rrdvar_type {
     RRDVAR_TYPE_COLLECTED               = 3,
     RRDVAR_TYPE_TOTAL                   = 4,
     RRDVAR_TYPE_INT                     = 5,
-    RRDVAR_TYPE_CALCULATED_ALLOCATED    = 6  // a custom variable, allocate on purpose (ie. not inherited from charts)
+    RRDVAR_TYPE_CALCULATED_ALLOCATED    = 6  // a custom variable, allocated on purpose (ie. not inherited from charts)
+                                             // used only for custom host global variables
 } RRDVAR_TYPE;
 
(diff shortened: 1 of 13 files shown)


ktsaou:
cc: @l2isbad (for the new plugin)

l2isbad:
@ktsaou 
About global. Are they really global or global per module/job?

About local. 

```python
 'system': {
        'options': [None, "Chrony System Time Deltas", "microseconds", 'system', 'chrony.system', 'area'],
        'lines': [
            ['timediff',   'system time', 'absolute', 1, 1000]
        ],
	    'variables': {'var_a': 3, 'var_b': 2, 'var_c': 3}
	}
```

?

l2isbad:
Is it allowed to change the values ​​of variables in runtime?

ktsaou:
> Is it allowed to change the values ​​of variables in runtime?

Yes, any time.

> Do we need global variables for python.d.plugin or only local vars?

I think it depends on the module. If the module monitors multiple instances of something, the best is local variables. If the module monitors the system itself, then global variables might be better.

For chrony, for example, if the module can monitor only 1 instance of chrony, either global or local will do.

To simplify things, we could say, to use local everywhere. I can't think of any limitation with this strategy.

l2isbad:
> To simplify things, we could say, to use local everywhere. I can't think of any limitation with this strategy.

Agree.

How it will work. Lets take https://github.com/firehol/netdata/pull/2642 as example.

```python
    'threads': {
        'options': [None, 'mysql Threads', 'threads', 'threads', 'mysql.threads', 'line'],
        'lines': [
            ['Threads_connected', 'connected', 'absolute'],
            ['Threads_created', 'created', 'incremental'],
            ['Threads_cached', 'cached', 'absolute', -1, 1],
            ['Threads_running', 'running', 'absolute'],
        ],
        variables: {var_max_connections: None}}
```
If value is None variable will not be created (pushed to netdata) at create() (if value is digit it will) but it will be added to the chart variables.

To update/set variable it's value should be returned in the dict that the `get_data()` method returns (with all other data). Once the variable is assigned a value it will be pushed to netdata in next update()

Insertion another variable in runtime to chart will be implemented too.

@Rockstar04 what do you think?

Rockstar04:
That seems like a workable solution to me.

ktsaou:
Yes, it seems nice!
I will try to merge this PR asap. It does not introduce more breaking changes. It only extends the previous PR.

ktsaou:
For the variable names, since they will be used at alarms, I suggest instead of `var_max_connections` to use `mysql_max_connections`, or just `max_connections`.

Remember we will have to use them like this:

```
calc: $connections * 100 / $max_connections
units: %
warn: $this > 80
crit: $this > 90
```

so, something easy to remember is best.

Since this will be chart local, it will also be accessible as `$chart.name.max_connections` (i.e. `$mysql_local.connections.max_connections`.

ktsaou:
done! this is ready and it works smoothly!
I'll merge it once travis completes.

ktsaou:
updated the wiki too.

ktsaou:
merged!

PR #2892, opened 2017-10-18
Add tcp ListenOverflows to FreeBSD plugin

vlvkobal:
Implements #2857 on FreeBSD. I can't find data equivalent to ListenDrops in FreeBSD, but as far as I understand, ListenOverflows is a major concern.
It's ready to merge.


ktsaou:
nice!

vlvkobal:
It's not a mistake. It was suspicious at first for me too. But comments for the variable in the header and `netstat` utility state that it handles overflows.

ktsaou:
ok, perfect!

PR #2893, opened 2017-10-18
web_log: resp_length is integer or "-"

racciari:
find_regex_custom function describes 'resp_length' can be integer or "-", but it wasn't the case. Made the change accordingly on the same base as 'bytes_sent'.


ktsaou:
thanks!

@l2isbad please review this.

ktsaou:
merged! thanks!

PR #2900, opened 2017-10-19
support multiple plugin directories

### LINKED ISSUE #2009: Allow adding plugins in /etc/netdata

ktsaou:
fixes #2009 

netdata now supports multiple external plugin directories. They are configured via the same variable, but now it accepts a space separated list of directories:

```
[global]
    plugins directory = "/usr/libexec/netdata/plugins.d" "/etc/netdata/custom-plugins.d"
```

By default netdata adds the directory `custom-plugins.d` inside the configs directory. Custom plugins can be made executable, with suffix `.plugin` and placed in this directory.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -243,6 +243,7 @@ typedef enum rrdcalc_status {
 extern char *netdata_configured_hostname;
 extern char *netdata_configured_config_dir;
 extern char *netdata_configured_log_dir;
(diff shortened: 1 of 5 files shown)


cransom:
@ktsaou I ran some testing and this looks like exactly what I was looking for. thanks!

PR #2906, opened 2017-10-21
Add kernel same-page merging Python plugin

FedericoCeratto:
File descriptors are cached to avoid open/close on every update.


ktsaou:
Thanks!

@FedericoCeratto this already exists, at the memory section. It is inside netdata, written in C.

PR #2907, opened 2017-10-21
Add declarative style example plugin

FedericoCeratto:
Align example with the style used in most plugins


l2isbad:
example.chart.py rewritten in https://github.com/firehol/netdata/pull/2872

PR #2908, opened 2017-10-21
Add Netdata plugin for Tor

FedericoCeratto:


ktsaou:
Thanks!
@l2isbad please review.

l2isbad:
I will do it after https://github.com/firehol/netdata/pull/2872 is merged

python.d/tor.chart.py
```diff
@@ -0,0 +1,128 @@
...
+
+        try:
+            if not STEM_AVAILABLE:
+                raise Exception("The stem library is missing")
+            self._conf = self.load_conf()
+            self._authenticate()
+            self._ready = True
+        except Exception as e:
+            self.error(str(e))
+            self._ready = False
```
l2isbad: Please move all logic to the `check() `method.
Like
```python
def check(self):
    if not STEM_AVAILABLE:
        self.error('The stem library is missing')
        return False
    ...
    ...
```

python.d/tor.chart.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def load_conf(self):
+        fn = os.path.join(NETDATA_CONFIG_DIR, "tor.conf")
+        with open(fn) as f:
+            j = json.load(f)
+        return j
```
l2isbad:
```python
    self.configuration = configuration  # <= configuration
```

python.d/tor.chart.py
```diff
@@ -0,0 +1,128 @@
...
+            self._controller = stem.control.Controller.from_port(port=port)
+        else:
+            socket_path = self._conf.get('socket_path', '')
+            self._controller = stem.control.Controller.from_socket_file(
+                path=socket_path)
+
+        try:
+            self._controller.authenticate(password=password)
+        except stem.connection.AuthenticationFailure as e:
+            self.error('Authentication failed ({})'.format(e))
```
l2isbad: change to (py 2.6 compatibility)
```python
self.error('Authentication failed ({0})'.format(e))
```

python.d/tor.chart.py
```diff
@@ -0,0 +1,128 @@
...
+#   "socket_path": "/tmp/REPLACEME",
+# }
+#
+# You can use "port" or "socket" as "connect_method"
+# The "port" and "socket_path" params will be used accodingly
+
+import json
+import os
+
+from base import SimpleService
```
l2isbad: change to
```python
```

python.d/tor.chart.py
```diff
@@ -0,0 +1,128 @@
...
+    # Running from CLI
+    s = Service(
+        name=NAME,
+        configuration=dict(
+            priority=priority,
+            retries=retries,
+            update_every=1
+        ),
+    )
+    s.update(1)
```
l2isbad: No needed.

You can run python.d.plugin in debug mode.
https://github.com/firehol/netdata/wiki/Add-more-charts-to-netdata#configuring-plugins

```
/usr/libexec/netdata/plugins.d/python.d.plugin 1 debug mysql
```

candrews:
_bump_

I'd love to see this get in - perhaps this effort can be revived?

FedericoCeratto:
@candrews updated

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+# default module values
+# update_every = 10
+priority = 90000
+retries = 60
+
+ORDER = ['traffic']
+CHARTS = {
+    'traffic': {
+        'options': [None, "Tor traffic", 'bytes/s', 'Tor traffic',
+                    'tor.traffic', 'line'],
```
l2isbad: please make it one line

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+        'lines': [
+            ["read", "read", "incremental"],
+            ["write", "write", "incremental"],
```
l2isbad: please use `'` instead of `"`

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class Service(SocketService):
+    """Provide netdata service for Tor"""
+    def __init__(self, configuration=None, name=None):
+        SocketService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+
+    def _authenticate(self):
```
l2isbad: don't make it protected

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+        connect_method = self.configuration.get('connect_method', 'port')
+        password = self.configuration.get('password', None)
```
l2isbad: i suggest to move all these `smth = self.configuration.get` to `init` (`self.smth =`)

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        self._parse_config()
+        if not STEM_AVAILABLE:
+            self.error("The stem library is missing")
+            return False
+
+        try:
+            self._authenticate()
+            return True
+        except Exception as e:
```
l2isbad: To broad exception

And i see `try/except` already in `self._authenticate()`. Why do we need it here?

python.d/tor.chart.py
l2isbad: method name is `authenticate`, but it performs connection and authorization.

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+            self.error(str(e))
+            return False
+
+    def get_metric(self, name):
+        try:
+            return self._controller.get_info(name)
+        except Exception:
```
l2isbad: to broad exception

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        try:
+            tr = self.get_metric('traffic/read')
+            tw = self.get_metric('traffic/written')
+            data = dict(
+                read=int(tr),
+                write=int(tw)
+            )
+            return data
+        except Exception as e:
```
l2isbad: to broad exception

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+    def _authenticate(self):
+        connect_method = self.configuration.get('connect_method', 'port')
+        password = self.configuration.get('password', None)
+        if connect_method == 'port':
+            port = self.configuration.get('port', 9051)
+            self._controller = stem.control.Controller.from_port(port=port)
+        else:
+            socket_path = self.configuration.get('socket_path', '')
+            self._controller = stem.control.Controller.from_socket_file(
+                path=socket_path)
```
Ferroin: Line length we're enforcing is 120 characters, so you don't need a line continuation here.

python.d/tor.chart.py
```diff
@@ -0,0 +1,109 @@
...
+        """Get host metrics
+        :return: dict
+        """
+        try:
+            tr = self.get_metric('traffic/read')
+            tw = self.get_metric('traffic/written')
+            data = dict(
+                read=int(tr),
+                write=int(tw)
+            )
```
Ferroin: Please update the above to the standard dictionary literal syntax.  It should end up being something like this:

```
data = {
    'read': int(tr),
    'write': int(tw),
}
```

python.d/tor.chart.py
Ferroin: Still too broad.  The issue here is that catching `Exception` also catches a number of errors that should never be caught, most notably syntax errors (`SyntaxError` is a subclass of `Exception`), but also things like `ValueError`, `AttributeError`, and `TypeError`.

python.d/tor.chart.py
```diff
@@ -0,0 +1,104 @@
...
+        'options': [None, 'Tor traffic', 'bytes/s', 'Tor traffic', 'tor.traffic', 'line'],
+        'lines': [
+            ['read', 'read', 'incremental'],
+            ['write', 'write', 'incremental'],
+        ]
+    }
+}
+
+
+class Service(SocketService):
```
l2isbad: btw why do you inherit from `SocketService` ?

FedericoCeratto: _parse_config is missing in SimpleService

l2isbad: yes, but do you really need it? what for?

ktsaou:
This need a rebase now.

candrews:
We're so close!

ktsaou:
> We're so close!

It depends on @l2isbad 	
So, @l2isbad 	please review.

l2isbad:
i suggest to use `SimpleService` as a parent class. I see no need in `SocketService`.

FedericoCeratto:
@l2isbad the PR is updated. I'm using SocketService instead of SimpleService because the latter is missing _parse_config

l2isbad:
> because the latter is missing _parse_config

Yeah, ok. Why do you need `_parse_config` ?

I still suggest to rewrite it using `SimpleService`.

In addition to that i suggest to rewrite `reconnect` functionality in a more simple way.

We need:

-  connect
-  reconnect
-  is_alive


We  use this pattern in many modules. 

Example:

https://github.com/netdata/netdata/blob/5bcaae918eaa3ded10000429906bba56da2b8d11/collectors/python.d.plugin/rethinkdbs/rethinkdbs.chart.py#L206-L231

One more thing - i see no need in protected methods, please use public methods

PR #2915, opened 2017-10-23
added field for source device in pushbullet sender

user501254:
This allows for clearly setting the source of alarm, uniquely identifiable by the unique device id when mentioned.
Also helps when running netdata on multiple machines. Using this, alerts from same devices can be grouped together in Pushbullet's interface.

In the below screenshot, notice that the notifications originates from "RaspberryPi" and are hence consolidated under it separately (don't mix with other user pushes). 
screenshot from 2017-10-24 02-20-34

Related information about [creating devices](https://docs.pushbullet.com/#create-device) for Pushbullet.

### DIFF
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -161,6 +161,9 @@ SEND_PUSHBULLET="YES"
 PUSHBULLET_ACCESS_TOKEN=""
 DEFAULT_RECIPIENT_PUSHBULLET=""
 
+# Device iden of the sending device. Optional.
+SOURCE_DEVICE=""
+
 
 #------------------------------------------------------------------------------
 # Twilio (twilio.com) SMS options
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -162,7 +162,7 @@ PUSHBULLET_ACCESS_TOKEN=""
 DEFAULT_RECIPIENT_PUSHBULLET=""
 
 # Device iden of the sending device. Optional.
-SOURCE_DEVICE=""
+PUSHBULLET_SOURCE_DEVICE=""
 
 
 #------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ktsaou:
Thanks!
Does pushbullet work when the device is empty? I mean, people will update their netdata and the alarms should work without reconfiguring this. So, for this to be merged, it has to work with an unset device.

ktsaou:
I also see that the text says: the source of alarm is ... and it states the script itself instead of the alarm file. There is something wrong there (not your fault, but could you please fix it?).

user501254:
Yes, you can still push without specifying the `PUSHBULLET_SOURCE_DEVICE=""` ie. when `"source_device_iden": ""` (is blank). Under this condition pushes will appear as they heave till now, mixed among other pushes.
I will look into https://github.com/firehol/netdata/pull/2915#issuecomment-338808364 once I get home.

user501254:
So I checked https://github.com/firehol/netdata/pull/2915#issuecomment-338808364 and feel that's how it's supposed to be for [test alarms](https://github.com/firehol/netdata/wiki/alarm-notifications#testing-notifications). Other real alarms show the line number and configuration file just fine.

user501254:
On another note, we can make the push notification look much pretty and functional if we send the push as a `"type": "link"`. Here's a sample comparison, you can see my new approach on the right:
screenshot from 2017-10-25 00-03-19  screenshot from 2017-10-25 00-05-24.

How this is better:
1. definitely looks better since the overly large link is partially hidden
2. maintains all the information that it had previously
3. on mobile if the new push is tapped from the pull down notifications shade, it will directly take you to the weblink of the alert in the browser and not to the pushbullet app instead.

Let me know your thoughts on this and if I can do these changes in the same PR or create a separate new one.

ktsaou:
I like it. Can you adapt it?

user501254:
@ktsaou 
Done. I also added a timestamp for the event in message body.
screenshot from 2017-10-26 01-02-55

ktsaou:
ok nice.

One little thing: please move the date below family and above the alarm source. The date is a detail compared to severity, chart and family (since these indicate what happened).

Then we are ready to merge.

user501254:
Moved the date

ktsaou:
merged it.
thank you!

user501254:
Welcome!
Let me know in case of any issues related to this by mentioning me.

Issue #2865, opened 2017-10-11
potgresql: After upgrading to postgres 10 logs are flooded with errors


skibbipl:
Hello,
```
2017-10-11 08:43:30 UTC [4960-1] postgres@postgres ERROR:  could not open directory "pg_xlog/archive_status": No such file or directory
2017-10-11 08:43:30 UTC [4960-2] postgres@postgres STATEMENT:
        SELECT
            CAST(COUNT(*) AS INT) AS file_count,
            CAST(COALESCE(SUM(CAST(archive_file ~ $r$\.ready$$r$ as INT)), 0) AS INT) AS ready_count,
            CAST(COALESCE(SUM(CAST(archive_file ~ $r$\.done$$r$ AS INT)), 0) AS INT) AS done_count
        FROM
            pg_catalog.pg_ls_dir('pg_xlog/archive_status') AS archive_files (archive_file);
```
In postgresql 10 pg_xlog directory was renamed to `pg_wal`, but netdata has `pg_xlog` hardcoded in python script:
https://github.com/firehol/netdata/blob/d2c4f8190fbd2e7e9877df9a7fd59ddf9525e354/python.d/postgres.chart.py#L65
I guess there should be some setting to set the directory for newest postgresql.

Regards

ktsaou:
@l2isbad @facetoe please check this.

skibbipl:
https://wiki.postgresql.org/wiki/New_in_postgres_10#Renaming_of_.22xlog.22_to_.22wal.22_Globally_.28and_location.2Flsn.29

anayrat:
Looks at https://github.com/OPMDG/check_pgactivity we have different queries depending on PG Version.

facetoe:
Hi guys, I've submitted a PR to fix this issue.

l2isbad:
thx @facetoe

facetoe:
No worries! As far as I can tell this directory name can't be changed by any configuration option so this patch should do the job.

ktsaou:
merged it!

PR #2917, opened 2017-10-24
postgres.chart.py: Support new WAL directory name.

facetoe:
In versions <= 10 the WAL directory was called 'pg_xlog', and from 10
onwards has been renamed to 'pg_wal'.

Raised in #2865

### DIFF
--- a/python.d/postgres.chart.py
+++ b/python.d/postgres.chart.py
@@ -62,7 +62,7 @@ SELECT
     CAST(COALESCE(SUM(CAST(archive_file ~ $r$\.ready$$r$ as INT)), 0) AS INT) AS ready_count,
     CAST(COALESCE(SUM(CAST(archive_file ~ $r$\.done$$r$ AS INT)), 0) AS INT) AS done_count
 FROM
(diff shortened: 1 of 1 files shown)


python.d/postgres.chart.py
```diff
@@ -62,7 +62,7 @@
     CAST(COALESCE(SUM(CAST(archive_file ~ $r$\.ready$$r$ as INT)), 0) AS INT) AS ready_count,
     CAST(COALESCE(SUM(CAST(archive_file ~ $r$\.done$$r$ AS INT)), 0) AS INT) AS done_count
 FROM
-    pg_catalog.pg_ls_dir('pg_xlog/archive_status') AS archive_files (archive_file);
+    pg_catalog.pg_ls_dir('{}/archive_status') AS archive_files (archive_file);
```
l2isbad: Python 2.6.6
```python
>>> '{}'.format(1)
ValueError: zero length field name in format
>>> '{0}'.format(1)
'1'
```

facetoe: Ahh, easy fix. Patch incoming. Good catch!

ktsaou:
merged it.
thanks!

PR #2927, opened 2017-10-25
calculate total system disk I/O by aggregating the I/O of all physical disks

### LINKED ISSUE #2877: zfs writes are not monitored
Hi,

Just created a simple stripe zfs pool with
zpool create -f zfspool sdb sdc sdd

on a Centos7 machine. 

And ran bonnie++ on it. The netdata showed read activity instead of write (but lower throughput than in reality).

Anyone else seen a similar problem?

Thanks

ktsaou:
fixes #2877

The total disk I/O reported by netdata came from `/proc/vmstat`, counters `pgpgin` and `pgpgout`:

- `pgpgin` is the amount of memory (in KB) that the kernel paged in from disk.
- `pgpgout` is the amount of memory (in KB) that the kernel paged out to disk.

It seems this does not include ZFS bandwidth.

So, this PR renames the old `system.io` to `system.pgpgio` and adds a new `system.io` that calculates its values as the sum of the all the physical disks of the system (it excludes partitions and disks with slaves).

### DIFF
--- a/src/proc_diskstats.c
+++ b/src/proc_diskstats.c
@@ -348,6 +348,8 @@ int do_proc_diskstats(int update_every, usec_t dt) {
 
     size_t lines = procfile_lines(ff), l;
 
+    collected_number system_read_kb = 0, system_write_kb = 0;
+
     for(l = 0; l < lines ;l++) {
         // --------------------------------------------------------------------------
         // Read parameters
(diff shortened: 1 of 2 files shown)


ktsaou:
There is a side effect with this:

RAID devices, now count the bandwidth for all disks. So, the kernel may be writing 1MB, but depending on the RAID setup, netdata will report a higher value.

PR #2946, opened 2017-10-30
cgroup-network can now detect libvirt tun/tap interfaces

### LINKED ISSUE #2545: libvirt plugin
netdata already monitors libvirt VMs through cgroups.
For each VM it can get CPU, Memory and Disk I/O, as seen by the host.

However, libvirt provides an API that can expose:

- CPU (same with cgroups),
- Memory (it can provide the amount of memory used inside the VM, while cgroups can only see the total allocated)
- Disk (I think it is the same with cgroups),
- Network (this is unique - it cannot be fetched via cgroups)

Additionally, libvirt is able to provide information about all the VMs running, their statuses, etc.

ktsaou:
and move them to cgroup section, much like it did for veth interfaces.

There is now a new external helper script that can also query virsh, docker, lxc, etc.
This script is executed by `cgroup-network`, so it runs as `root`.

The result is now that VMs also get their network interfaces at their section:

screenshot from 2017-10-31 01-32-43

fixes  #2545

### DIFF
--- a/plugins.d/Makefile.am
+++ b/plugins.d/Makefile.am
@@ -12,6 +12,7 @@ dist_plugins_SCRIPTS = \
 	alarm-notify.sh \
 	alarm-test.sh \
 	cgroup-name.sh \
+	cgroup-network-helper.sh \
 	charts.d.dryrun-helper.sh \
 	charts.d.plugin \
 	fping.plugin \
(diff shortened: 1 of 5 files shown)


l2isbad:
do we still need https://github.com/firehol/netdata/issues/2545 ?

PR #2951, opened 2017-11-01
minor fixes: netdata.spec; zfs l2hits; gauge gradient colors

### LINKED ISSUE #2902: Gradient color problem
Hello, i'm trying to put gradient color on gauges but i'm not able to do:

			div data-netdata="system.cpu"
                        data-host="http://xxx/netdata/server5/"
                        data-common-max="system.cpu"
                        data-decimal-digits="0.0"
                        data-title="System CPU"
			data-chart-library="gauge"
                        data-gauge-max-value="100"
			data-gauge-generate-gradient="[0, 70, 100]"
			data-gauge-gradient-percent-color-0="#4fce00"
			data-gauge-gradient-percent-color-70="#ffb200"
			data-gauge-gradient-percent-color-100="#ff0000"
                        data-units="%"
                        data-width="9%"
                        data-height="100%"
                        data-after="-300"
                        data-points="300"
                        /div

With or without the options, the chart stay with the same default color, its only changes if i put data-colors="#228B22" but its static and not dynamic like i wanted...

### LINKED ISSUE #2945: l2arc accounting for misses as hits
Hello, I just quickly skimmed through the sources and thought this looked out of place:

```
// l2 reads
unsigned long long l2hit = arcstats.l2_hits + arcstats.l2_misses;
```

I wouldn't expect l2hit to be the sum of hits and misses; is the name slightly misleading or is one of these variables supposed to aim at something else?

Thanks

https://github.com/firehol/netdata/blob/502c33810fcec4fd5e5171b1d7ab6b23365bfcbf/src/zfs_common.c#L27

ktsaou:
This PR updates `netdata.spec` with the permission on the missing setuid netdata plugins.

---

It also fixes a wrong calculation of ZFS l2hits, misses and reads. fixes #2945 

---

It also restores gradient colors for gauge charts. fixes #2902


ktsaou:
@alonbl I would be grateful if you could review this netdata.spec change.

alonbl:
H,

This suid and caps usage is growing, as usual I do not like it, it becomes a severe security issue of this component.

My repeated recommendation is to have separate daemons for each privilege escalation.

If you do not want to have this, at least create a user/group for netdata and permit only this user/group to execute these privilege escalation scripts.

However, as privilege escalation is used too much, probably better to just assign a user with all capabilities and run everything in this context. But again, the web interface should probably split out of the core.

For this specific patch, looks ok if we ignore security.

Alon

ktsaou:
ok, thanks.

> For this specific patch, looks ok if we ignore security.

As I have said before, I really don't see how splitting netdata into 2 daemons would improve security. I think the opposite will happen: we will have a lot more code running with escalated privileges.

netdata today, attempts to minimize the code running escalated. Only the absolutely necessary do this, without exceptions. The whole daemon runs unprivileged and only very specific plugins are split from netdata into external programs and run escalated.

As you suggest, we should only allow group netdata to run these privileged plugins. So, should we change the ownership of these plugins to `root:netdata` with permissions `4550` ?

alonbl:
On 2 November 2017 at 12:17, Costa Tsaousis <notifications@github.com> wrote:
>
> As I have said before, I really don't see how splitting netdata into 2 daemons would improve security. I think the opposite will happen: we will have a lot more code running with escalated privileges.

I do not think you are correct...

> netdata today, attempts to minimize the code running escalated. Only the absolutely necessary do this, without exceptions. The whole daemon runs unprivileged and only very specific plugins are split from netdata into external programs and run escalated.

We discussed this already... I must put  my notice, as I would not
approve this approach internally.

> As you suggest, we should only allow group netdata to run these privileged plugins. So, should we change the ownership of these plugins to root:netdata with permissions 4550 ?

Yes, you must make sure that only netdata can access these privilege
escalation modules.

ktsaou:
> I do not think you are correct...

I am sorry, I still don't get it. Probably I haven't understood what you propose. But, I can't find even a single tiny benefit from turning external setuid programs, into a separate daemon. I actually find a lot of drawbacks, both in usability and security.

I have spent 2 hours evaluating different cases, to understand your view. But, I can't.

I have evaluated 3 cases:

## Case 1: split escalated and non-escalated functions

1. Have all the escalated data collection plugins running as a separate daemon
2. Have all the non-escalated data collection plugins running with the web server daemon

In this case, a lot of plugins would have to be moved from the non-escalated to the escalated daemon, just because they need a little tiny info that is available only when escalated (cgroups is one example, that needs to run `cgroup-network` as root). 

**This actually means that a lot more code will be now running escalated.**

## Case 2: split data collection and web

1. a data collection daemon, escalated
2. a data collection daemon, non-escalated
3. a web server

Like above, a lot of plugins would have to be moved from the non-escalated to the escalated data collection daemon. Again, **this actually means that a lot more code will be running escalated.**

But, now the web server is separated. We could say that this is a better setup in terms of security because the web server is separate. **But this can only be true if the web server is running as a different user than the non-escalated data collection server.** If they run as the same user, security is the same.

### Case 3: run escalated plugins with setcap/setuid

Today, external escalated plugins cannot accept commands from netdata. They perform a hard coded job, pushing metrics to netdata, without reading even a single byte from netdata. So, the unprivileged netdata server cannot actually use them to do anything bad.

I find this a lot more secure. A lot less code runs escalated. This code is isolated, meaning that it cannot accept commands from netdata.

---

If you believe my thinking is wrong, please provide a use case that it fails compared to what you have in mind. I am very keen to understand what you propose.

PR #2957, opened 2017-11-02
added support for flock notification webhooks

tvarsis:
Implemented compatible alarm notifications for flock.com webhooks similar to how the other webhooks are implemented.


ktsaou:
Nice! Please update the wiki (similar to what already exists for slack).
Once this is done, I think we can merge this.

ktsaou:
merged
thanks!

ktsaou:
for the infographic:

image

PR #2976, opened 2017-11-07
Switch from pd-send to curl for pagerduty notifications

HackToHell:
Currently uses pd-send which needs to be installed on the server to work, now this works with just curl and using the pager duty WebAPI.

https://v2.developer.pagerduty.com/v2/docs/trigger-events

### DIFF
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -346,10 +346,6 @@ KAFKA_SENDER_IP=""
 
 #------------------------------------------------------------------------------
 # pagerduty.com notification options
-#
-# pagerduty.com notifications require the pagerduty agent to be installed and 
-# a "Generic API" pagerduty service.
-# https://www.pagerduty.com/docs/guides/agent-install-guide/
 
 # multiple recipients can be given like this:
 #              "<pd_service_key_1> <pd_service_key_2> ..."
--- a/plugins.d/alarm-notify.sh
+++ b/plugins.d/alarm-notify.sh
@@ -607,24 +607,13 @@ done
 # check kafka
 [ -z "${KAFKA_URL}" -o -z "${KAFKA_SENDER_IP}" ] && SEND_KAFKA="NO"
 
-# check pagerduty.com
-# if we need pd-send, check for the pd-send command
-# https://www.pagerduty.com/docs/guides/agent-install-guide/
-if [ "${SEND_PD}" = "YES" ]
-    then
-    pd_send="$(which pd-send 2>/dev/null || command -v pd-send 2>/dev/null)"
-    if [ -z "${pd_send}" ]
-        then
-        error "Cannot find pd-send command in the system path. Disabling pagerduty.com notifications."
-        SEND_PD="NO"
-    fi
-fi
 
 # if we need curl, check for the curl command
 if [ \( \
            "${SEND_PUSHOVER}"    = "YES" \
         -o "${SEND_SLACK}"       = "YES" \
         -o "${SEND_FLOCK}"       = "YES" \
+        -o "${SEND_PD}"          = "YES" \
         -o "${SEND_DISCORD}"     = "YES" \
         -o "${SEND_HIPCHAT}"     = "YES" \
         -o "${SEND_TWILIO}"      = "YES" \
@@ -652,6 +641,7 @@ if [ \( \
         SEND_KAVENEGAR="NO"
         SEND_KAFKA="NO"
         SEND_CUSTOM="NO"
+        SEND_PD="NO"
     fi
 fi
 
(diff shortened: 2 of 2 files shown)


ktsaou:
Thanks!

@jimcooley please have look.

I don't know if we should support pagerduty using both methods. So, if `pd-send` is available use that, otherwise fallback to `curl`. I don't know though.

Your call guys...

HackToHell:
I think we should have curl because pd-send is a glorified wrapper around the Web API. Also when it's deployed via docker, it's easier to bundle curl and support multiple services. 

On the other hand uses pd-send means we don't have to maintain any API changes with regard to pagerduty itself. (That's pretty rare).

cc @jimcooley

ktsaou:
I totally agree with you.

Let's wait a bit for @jimcooley to review your changes.

jimcooley:
This looks great as an in-place swap from pd-agent to curl.

Two thoughts:
1. pd-agent handles queuing and retrying in case there are connectivity issues.  I'm assuming that curl will timeout and drop the notification if it can't reach the PagerDuty api.
2. This still uses the v1 events API from pagerduty.  I'm not sure if you want to make the switch to v2 along with the switch to curl.  https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2

ktsaou:
> pd-agent handles queuing and retrying in case there are connectivity issues. I'm assuming that curl will timeout and drop the notification if it can't reach the PagerDuty api.

hm... I can probably make netdata run again `alarm-notify.sh` if it fails, but netdata only knows the final return code of `alarm-notify.sh`. If `alarm-notify.sh` manages to send the alarm via any channel, it returns success to netdata.

I should probably re-think `alarm-notify.sh` a bit. I would be great if it could somehow know of failed notifications and retry them.

Until then... your call for pd...

ktsaou:
btw, probably using `pd-send` when it is available and fallback to `curl` when it is not there, is the best choice.

Supporting v2 is also nice. If you want, you can add a setting, so people could choose what they want.

HackToHell:
@ktsaou Yeah that makes sense let me add the fall back code to curl, can we add retries in case of timeout s or 500 to the `docurl` function ?

ktsaou:
> can we add retries in case of timeout s or 500 to the docurl function

I think curl returns `000` is case it cannot send the request (network is down).
Yes, you can add timeout and handle both `500` and `000`. But please set a `CURL_TIMEOUT` variable, so that this is configurable.

ktsaou:
You can use:

- `CURL_TIMEOUT` for controlling the timeout for each retry
- `CURL_RETRIES` for controlling the number of retries `docurl()` will do

Set them at the top of `alarm-notify.sh` and add them commented at `health_alarm_notify.conf`, so that people will be able to uncomment and change.

jimcooley:
Before this PR is merged, changes from #3549 should be incorporated.

ktsaou:
#3549 is not merged...

elisiariocouto:
Hello guys. Do you have any plans to merge this? I'm currently using [titpetric/netdata](https://github.com/titpetric/netdata) docker image and I'm trying to setup PagerDuty notifications but `pd-send` is not in the image. I've already tried `netdata/netdata` docker images but they also don't have `pd-send` in them.

If you need some help let me know

Ferroin:
I think probably yes.  As of right now it should just need a rebase to handle the recent restructuring of the source directory.

I would suggest waiting to push anything though until GitHub re-enables webhooks though (see https://blog.github.com/2018-10-21-october21-incident-report/ for more info), as until that happens the checks won't all run properly.

jimcooley:
Now that github is back online, this could use a rebase and merge conflict before being merged.  I'm on board with that merge.  I'll take a crack at the merge conflicts.

jimcooley:
@HackToHell - I have the merge conflicts resolved locally.  Would you be okay if I pushed those to your fork in order to keep the changes in this PR?  I would need permissions to push commits to your forked repo.

Alternatively, I could create another fork and another PR.  I haven't really collaborated on someone else's fork and PR, so maybe there's another way.

jimcooley:
Actually, I'll fork the fork and create a PR into the @HackToHell fork for review.

@HackToHell - please let me know if you are available to review and merge a PR with the rebase and merge conflicts.

elisiariocouto:
Hello @jimcooley. Sorry to bother but I guess @HackToHell is not active on GitHub. Do you want to fork on `netdata/netdata` instead and create a Pull Request?

jimcooley:
@elisiariocouto - Thanks for the nudge.  I'll go ahead and create a new PR with these changes that can be directly merged.

jimcooley:
I've created #4673 to replace this PR.  However, if @HackToHell isn't active anymore, we may not be able to pass the CLA test.

elisiariocouto:
I've forked the repo, reproduced @HackToHell changes and opened #4694.
I know this is not the perfect solution, but this is a blocker to me and might be to someone else as well.

PR #2980, opened 2017-11-07
Add charts for Galera replication

spinitron:
with galera operating

Tested on MariaDB 10.1.26 with `wsrep_on = ON` and `OFF`.

With `wsrep_on = OFF` but disabled, one wsrep status variable, wsrep_local_bf_aborts, is present (a bug?) but zero. So the galera chart section shows up with one chart, as shown below. I'd prefer if no galera charts were displayed in this case but I don't know how to do that.

wsrep_on=OFF

I haven't tested it, but I imagine that servers without galera available (e.g. Mariadb ≤ 10.0 or other MySQL derivatives), no galera charts and the galera chart section do not appear.


ktsaou:
thanks!

@l2isbad please review this.

l2isbad:
@spinitron Chart is present if module can get value for **at least one dimension**
https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L211-L239

l2isbad:
@ktsaou does `bytes/s` works with units scaling?

spinitron:
i doubt it. i should learn about unit scaling and update the pr

spinitron:
i doubt it. i should learn about unit scaling and update the pr

spinitron:
i doubt it. i should learn about unit scaling and update the pr

spinitron:
i doubt it. i should learn about unit scaling and update the pr

ktsaou:
hm... bytes/s is probably too low.

If this is disk I/O, I suggest to set it to `KB/s` and supply a divisor of 1024 to the chart.
If it is network I/O, I suggest to set it to `kilobits/s` and supply a divisor of 1000 to the chart.

Also, I suggest to change this to an `area` chart.

tom--:
It appears that the Galera "Replicated bytes" chart y-axis scales automatically. I've seen it with MB/s and KB/s.

I made the changes suggested anyway.

ktsaou:
thanks.
@l2isbad ready to merge?

l2isbad:
yes

ktsaou:
merged! thanks!

PR #2999, opened 2017-11-11
enable partial match for tomcat connector

### LINKED ISSUE #2996: Connector tomcat not collect full stats (only 4 over 8)
I install netdata ok
I use tomcat plugin ok but I only see 4 graphics of memory and not 4 graphs (access/bandwidth/processing_time/threads)

In my tomcat log all ok, except these graphs show me: was removed due to non updating.

In my netdata config:

localhost:
name: ...
url: ...
user: ...
pass: ...
connector: 'http-nio-8080'

localipv4:
name: ...
url: ...
user: ...
pass: ...
connector: 'http-nio-8080'

localipv6:
name: ...
url: ...
user: ...
pass: ...
connector: 'http-nio-8080'

My http://localhost:8080/managet/status?XML=true show me a full xml stats with all metrics and I don't understand why python plugin not is successed.

I debugged tomcat.chart.py and I detected that line 104 (https://github.com/firehol/netdata/blob/master/python.d/tomcat.chart.py) 

...
...
 if conn.get('name') == self.connector_name:
...
..

conn.get('name') has value "http-nio-8080" but self.connector_name has value http-nio-8080

There is a double quote before and after connector name.

I change:
      if conn.get('name').replace('"','') == self.connector_name:
by
       if conn.get('name') == self.connector_name:

and all ok.

Wing924:
fix #2996

### DIFF
--- a/python.d/tomcat.chart.py
+++ b/python.d/tomcat.chart.py
@@ -101,7 +101,7 @@ class Service(UrlService):
             connector = None
             if self.connector_name:
                 for conn in xml.findall('connector'):
-                    if conn.get('name') == self.connector_name:
+                    if self.connector_name in conn.get('name'):
                         connector = conn
                         break
             else:


ktsaou:
merged it. Sorry for the delay. I missed this somehow...

PR #3009, opened 2017-11-13
Support for dnsdist

nobody-nobody:
I wrote a plugin for my dnsdist installation.
I am new to netdata so I took the powerdns plugin as base.
Maybe a dev can review, modify and merge (if useful) it...

https://dnsdist.org/

dnsdist

Best regards

### DIFF
--- a/python.d/dnsdist.chart.py
+++ b/python.d/dnsdist.chart.py
@@ -2,54 +2,83 @@
 from json import loads
 from bases.FrameworkServices.UrlService import UrlService
 
-ORDER = ['latency', 'cache', 'acl', 'noncompliant', 'queries', 'health']
+ORDER = ['queries', 'queries_dropped', 'answers', 'backend_responses', 'backend_commerrors', 'backend_errors', 'cache', 'servercpu', 'servermem', 'query_latency', 'query_latency_avg']
 CHARTS = {
-	'latency': {
-		'options': [None, 'response latency', 'value', 'latency', 'dnsdist.latency', 'area'],
-		'lines': [
-			['latency-slow', '> 1sec', 'incremental'],
-			['latency100-1000', '100-1000ms', 'incremental'],
-			['latency50-100', '50-100ms', 'incremental'],
-			['latency10-50', '10-50ms', 'incremental'],
-			['latency1-10', '1-10ms', 'incremental'],
-			['latency0-1', '< 1ms', 'incremental']
-	]},
-	'cache': {
-		'options': [None, 'cache performance', 'value', 'cache', 'dnsdist.cache', 'area'],
-		'lines': [
-			['cache-hits', 'hits', 'incremental'],
-			['cache-misses', 'misses', 'incremental']
-	]},
-	'acl': {
-		'options': [None, 'access-control-list events', 'value', 'acl', 'dnsdist.acl', 'area'],
-		'lines': [
-			['acl-drops', 'drop by acl', 'incremental'],
-			['rule-drop', 'drop by rule', 'incremental'],
-			['rule-nxdomain', 'nxdomain by rule', 'incremental'],
-			['rule-refused', 'refused by rule', 'incremental']
-	]},
-	'noncompliant': {
-		'options': [None, 'noncompliant data', 'value', 'noncompliant', 'dnsdist.noncompliant', 'area'],
-		'lines': [
-			['empty-queries', 'empty queries', 'incremental'],
-		    ['no-policy', 'no policy', 'incremental'],
-			['noncompliant-queries', 'noncompliant queries', 'incremental'],
-			['noncompliant-responses', 'noncompliant responses', 'incremental']
-	]},
 	'queries': {
-		'options': [None, 'queries', 'value', 'queries', 'dnsdist.queries', 'area'],
+		'options': [None, 'Client queries received', 'queries/s', 'queries', 'dnsdist.queries', 'line'],
 		'lines': [
-			['queries', 'queries', 'incremental'],
-			['rdqueries', 'recursive queries', 'incremental'],
+			['queries', 'all', 'incremental'],
+			['rdqueries', 'recursive', 'incremental'],
+			['empty-queries', 'empty', 'incremental']
+	]},
+	'queries_dropped': {
+		'options': [None, 'Client queries dropped', 'queries/s', 'queries', 'dnsdist.queries_dropped', 'line'],
+		'lines': [
+			['rule-drop', 'rule drop', 'incremental'],
+			['dyn-blocked', 'dynamic block', 'incremental'],
+			['no-policy', 'no policy', 'incremental'],
+			['noncompliant-queries', 'non compliant', 'incremental'],
+			['acl-drops', 'acl', 'incremental'],
+	]},
+	'answers': {
+		'options': [None, 'Answers statistics', 'answers/s', 'answers', 'dnsdist.answers', 'line'],
+		'lines': [
+			['self-answered', 'self answered', 'incremental'],
+			['rule-nxdomain', 'nxdomain', 'incremental', -1],
+			['rule-refused', 'refused', 'incremental', -1],
+			['trunc-failures', 'trunc failures', 'incremental', -1]
+	]},
+	'backend_responses': {
+		'options': [None, 'Backend responses', 'responses/s', 'backends', 'dnsdist.backend_responses', 'line'],
+		'lines': [
 			['responses', 'responses', 'incremental']
 	]},
-	'health': {
-		'options': [None, 'health', 'value', 'health', 'dnsdist.health', 'area'],
+	'backend_commerrors': {
+		'options': [None, 'Backend Communication Errors', 'errors/s', 'backends', 'dnsdist.backend_commerrors', 'line'],
 		'lines': [
-			['downstream-send-errors', 'ds send errors', 'incremental'],
-			['downstream-timeouts', 'ds timeouts', 'incremental'],
-			['servfail-responses', 'servfail responses', 'incremental'],
-			['trunc-failures', 'trunc failures', 'incremental']
+			['downstream-send-errors', 'send errors', 'incremental']
+	]},
+	'backend_errors': {
+		'options': [None, 'Backend error responses', 'responses/s', 'backends', 'dnsdist.backend_errors', 'line'],
+		'lines': [
+			['downstream-timeouts', 'timeout', 'incremental'],
+			['servfail-responses', 'servfail', 'incremental'],
+			['noncompliant-responses', 'non compliant', 'incremental']
+	]},
+	'cache': {
+		'options': [None, 'Cache performance', 'answers/s', 'cache', 'dnsdist.cache', 'area'],
+		'lines': [
+			['cache-hits', 'hits', 'incremental'],
+			['cache-misses', 'misses', 'incremental', -1]
+	]},
+	'servercpu': {
+		'options': [None, 'DNSDIST server CPU utilization', 'ms/s', 'server', 'dnsdist.servercpu', 'stacked'],
+		'lines': [
+			['cpu-sys-msec', 'system state', 'incremental'],
+			['cpu-user-msec', 'user state', 'incremental']
+	]},
+	'servermem': {
+		'options': [None, 'DNSDIST server memory utilization', 'MiB', 'server', 'dnsdist.servermem', 'area'],
+		'lines': [
+			['real-memory-usage', 'memory usage', 'absolute', 1, 1048576]
+	]},
+	'query_latency': {
+		'options': [None, 'Query latency', 'queries/s', 'latency', 'dnsdist.query_latency', 'stacked'],
+		'lines': [
+			['latency0-1', '1ms', 'incremental'],
+			['latency1-10', '10ms', 'incremental'],
+			['latency10-50', '50ms', 'incremental'],
+			['latency50-100', '100ms', 'incremental'],
+			['latency100-1000', '1sec', 'incremental'],
+			['latency-slow', 'slow', 'incremental']
+	]},
+	'query_latency_avg': {
+		'options': [None, 'Average latency for the last N queries', 'ms/query', 'latency', 'dnsdist.query_latency_avg', 'line'],
+		'lines': [
+			['latency-avg100', '100', 'absolute'],
+			['latency-avg1000', '1k', 'absolute'],
+			['latency-avg10000', '10k', 'absolute'],
+			['latency-avg1000000', '1000k', 'absolute']
 	]}
 }
 
--- a/python.d/dnsdist.chart.py
+++ b/python.d/dnsdist.chart.py
@@ -2,7 +2,7 @@
 from json import loads
 from bases.FrameworkServices.UrlService import UrlService
 
-ORDER = ['queries', 'queries_dropped', 'answers', 'backend_responses', 'backend_commerrors', 'backend_errors', 'cache', 'servercpu', 'servermem', 'query_latency', 'query_latency_avg']
+ORDER = ['queries', 'queries_dropped', 'packets_dropped', 'answers', 'backend_responses', 'backend_commerrors', 'backend_errors', 'cache', 'servercpu', 'servermem', 'query_latency', 'query_latency_avg']
 CHARTS = {
 	'queries': {
 		'options': [None, 'Client queries received', 'queries/s', 'queries', 'dnsdist.queries', 'line'],
@@ -17,7 +17,11 @@ CHARTS = {
 			['rule-drop', 'rule drop', 'incremental'],
 			['dyn-blocked', 'dynamic block', 'incremental'],
 			['no-policy', 'no policy', 'incremental'],
-			['noncompliant-queries', 'non compliant', 'incremental'],
+			['noncompliant-queries', 'non compliant', 'incremental']
+	]},
+	'packets_dropped': {
+		'options': [None, 'Packets dropped', 'packets/s', 'packets', 'dnsdist.packets_dropped', 'line'],
+		'lines': [
 			['acl-drops', 'acl', 'incremental']
 	]},
 	'answers': {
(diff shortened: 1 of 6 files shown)


python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+from json import loads
+from bases.FrameworkServices.UrlService import UrlService
+
+ORDER = ['latency_rel', 'latency_abs', 'cache_rel', 'cache_abs', 'acl_abs', 'invalid_abs', 'queries_rel', 'queries_abs', 'health_abs']
+CHARTS = {
+	'latency_rel': {
+		'options': [None, 'Average latency', 'value', 'Latency', 'dnsdist.latency_rel', 'area'],
```
l2isbad: family should be in lowercase (`Latency`). chart type should be `stacked` imo.

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+		'lines': [
+			['latency-slow', '> 1sec', 'incremental'],
+			['latency100-1000', '100-1000ms', 'incremental'],
+			['latency50-100', '50-100ms', 'incremental'],
+			['latency10-50', '10-50ms', 'incremental'],
+			['latency1-10', '1-10ms', 'incremental'],
+			['latency0-1', '< 1ms', 'incremental']
+	]},
+	'latency_abs': {
+		'options': [None, 'Absolute latency', 'value', 'Latency', 'dnsdist.latency_abs', 'line'],
```
l2isbad: Do you really need absolute values? Seem ratio will be better 

family `Latency` -> `latency`
chart type `line` -> `stacked`
dimension algorithm `absolute` -> `percentage-of-absolute-row` (the % of this value compared to the total of all dimensions)

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+		'lines': [
+			['latency-slow', '> 1sec', 'absolute'],
+			['latency100-1000', '100-1000ms', 'absolute'],
+			['latency50-100', '50-100ms', 'absolute'],
+			['latency10-50', '10-50ms', 'absolute'],
+			['latency1-10', '1-10ms', 'absolute'],
+			['latency0-1', '< 1ms', 'absolute']
+	]},
+	'cache_rel': {
+		'options': [None, 'Relative cache stats', 'value', 'Cache', 'dnsdist.cache_rel', 'area'],
```
l2isbad: family `Cache` -> `cache`
chart type `area` -> `stacked`

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+		'lines': [
+			['cache-hits', 'hits', 'incremental'],
+			['cache-misses', 'misses', 'incremental']
+	]},
+	'cache_abs': {
+		'options': [None, 'Absolute cache stats', 'value', 'Cache', 'dnsdist.cache_abs', 'area'],
```
l2isbad: family `Cache` -> `cache`
chart type `area` -> `stacked`
dimension algorithm `absolute` -> `percentage-of-absolute-row`

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+		'lines': [
+			['cache-hits', 'hits', 'absolute'],
+			['cache-misses', 'misses', 'absolute']
+	]},
+	'acl_abs': {
+		'options': [None, 'AccessControlList stats', 'value', 'ACL', 'dnsdist.acl_abs', 'line'],
```
l2isbad: family `ACL` -> `acl`

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+		'lines': [
+			['acl-drops', 'drop by acl', 'absolute'],
+			['rule-drop', 'drop by rule', 'absolute'],
+			['rule-nxdomain', 'rnxdomain by rule', 'absolute'],
+			['rule-refused', 'refused by rule', 'absolute']
+	]},
+	'noncompliant_abs': {
+		'options': [None, 'Noncompliant data', 'value', 'Invalid', 'dnsdist.noncompliant_abs', 'line'],
```
l2isbad: family `Invalid` -> `invalid`

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+		'lines': [
+			['empty-queries', 'empty queries', 'absolute'],
+		    ['no-policy', 'no policy', 'absolute'],
+			['noncompliant-queries', 'noncompliant queries', 'absolute'],
+			['noncompliant-responses', 'noncompliant responses', 'absolute']
+	]},
+	'queries_rel': {
+		'options': [None, 'Relative query stats', 'value', 'Queries', 'dnsdist.queries_rel', 'line'],
```
l2isbad: family `Queries` -> `queries`

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+		'lines': [
+			['queries', 'queries', 'incremental'],
+			['rdqueries', 'rd queries', 'incremental'],
+			['responses', 'responses', 'incremental']
+	]},
+	'queries_abs': {
+		'options': [None, 'Absolute query stats', 'value', 'Queries', 'dnsdist.queries_abs', 'line'],
```
l2isbad: family `Queries` -> `queries`

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+		'lines': [
+			['queries', 'queries', 'absolute'],
+			['rdqueries', 'rd queries', 'absolute'],
+			['responses', 'responses', 'absolute']
+	]},
+	'health_abs': {
+		'options': [None, 'Health', 'value', 'Health', 'dnsdist.health_abs', 'line'],
```
l2isbad: family `Health` -> `health`

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,96 @@
...
+
+	def _get_data(self):
+		data = self._get_raw_data()
+		if not data:
+			return None
+			
+		rdict = dict()
+		jdata = loads(data)
+		for d in jdata:
+			rdict[str(d)] = jdata[str(d)]
```
l2isbad: Just return loads result

```python
return loads(data)
```

nobody-nobody: Well... it was late :)
Did some testing before and did not realize I dont need this all at the end.

python.d/dnsdist.chart.py
nobody-nobody: Area looked more pretty to me than stacked but this is only my personal pref

python.d/dnsdist.chart.py
nobody-nobody: Yes, maybe we can drop absolute latency.
Only useful for long time stats and for that there are better solutions than netdata.

nobody-nobody:
Looks good, will stick to lowercase in the future.
Is there a guideline about which shape to use? I just took what seemed to fit best to me...

Is there a possibility to parse plugin config parameters in this file (did not find info about that)?
So everyone could select which metrics/shape he prefers... 

Thanks for the review.

l2isbad:
> Is there a possibility to parse plugin config parameters in this file (did not find info about that)?
> So everyone could select which metrics/shape he prefers...

I don't think we need such options. Anyway all job configuration parameters are stored in `self.configuartion`.

I may be wrong but all `total_abs` (latency, cache, quieries) chart makes no sense for me in current implementation. I think we should drop them or replace with something meaningful (ratio as example).
@ktsaou ?

nobody-nobody:
Alright. I fixed those coding style issues and included most of your suggestions.
All absolute charts are gone (more useful for long time monitoring).
Still unsure about "area" vs "stacked" but thats not my call.

l2isbad:
@nobody-nobody 
please do
1. update makefiles (https://github.com/firehol/netdata/blob/master/python.d/Makefile.am, https://github.com/firehol/netdata/blob/master/conf.d/Makefile.am)
2. python.d [readme](https://github.com/firehol/netdata/blob/master/python.d/README.md)
3. wiki [page ](https://github.com/firehol/netdata/wiki/Add-more-charts-to-netdata)

optional
4. alerts and dashboard icon/info (example - https://github.com/firehol/netdata/pull/2977/files)

l2isbad:
And please fix codacy error 😄 (Trailing whitespace line 66)

l2isbad:
@ktsaou the rule "all chart dimensions must have same units" is not true for this module. If you have any suggestion how dimensions should be grouped please suggest.

nobody-nobody:
Will do tomorrow if I can find some time (if not than on Thursday).
I also wrote a plugin for powerdns-recursor to make the powerdns family complete.
Will review both tomorrow and adapt to your coding style... and hunt down that nasty whitespace :)

nobody-nobody:
Alright, updated (whitespace is gone now :) )
Could you please review again?

I guess I wait modifying the wiki page until it is officially merged?

PS:
Will create a push request for pdns-recursor once we are done with this one.

ktsaou:
ok, I tried to group them a bit, based on this https://dnsdist.org/statistics.html

## Client Queries

### Queries

title: `Client queries received`
units: `queries/s`,
algorithm: `incremental`
type: `line`
family: `queries`

**queries** with name `all`
Number of received queries.

**rdqueries** with name `recursive`
Number of received queries with the recursion desired bit set.

**empty-queries** with name `empty` and multiplier `-1`
Number of empty queries received from clients.

### Queries Dropped

title: `Client queries dropped`
units: `queries/s`,
algorithm: `incremental`
type: `line`
family: `queries`

**rule-drop** as `rule drop`
Number of queries dropped because of a rule.

**dyn-blocked** as `dynamic block`
Number of queries dropped because of a dynamic block.

**no-policy** as `no policy`
Number of queries dropped because no server was available.

**noncompliant-queries** as `non compliant`
Number of queries dropped as non-compliant.

**acl-drops** as `acl`
The number of packets dropped bacause of the ACL.

> I am not sure **acl-drops** is about client queries. However, what else could be? So, I decided that its place is under queries.

---

## Answers

title: `Answers statistics`
units: `answers/s`,
algorithm: `incremental`
type: `line`
family: `answers`

> The problem here is that the server is not reporting the whole number of answers it sends. It only reports a subset of the answers.

**self-answered**
Number of self-answered responses.

**rule-nxdomain** with multiplier `-1`
Number of NXDomain answers returned because of a rule.

**rule-refused** with multiplier `-1`
Number of Refused answers returned because of a rule.

**trunc-failures**  with multiplier `-1`
Number of errors encountered while truncating an answer.

---

## Backends

family: `backends` on all charts in this section

### Backend Responses

title: `Backend Responses`
units: `responses/s`,
algorithm: `incremental`
type: `line`

**responses**
Number of responses received from backends.


### Backend Communication Errors

title: `Backend Communication Errors`
units: `errors/s`,
algorithm: `incremental`
type: `line`

**downstream-send-errors** as `send errors`
Number of errors when sending a query to a backend.


### Backend Errors

title: `Backend Error Responses`
units: `responses/s`,
algorithm: `incremental`
type: `line`

**downstream-timeouts** as `timeout`
Number of queries not answer in time by a backend.

**servfail-responses** as `servfail`
Number of servfail answers received from backends.

**noncompliant-responses** as `non compliant`
Number of answers from a backend dropped as non-compliant.

---

## Cache performance

title: `Cache performance`
units: `answers/s`,
algorithm: `incremental`
type: `area`
family: `cache`

**cache-hits** (name `hits`, multiplier: `1`)
Number of times an answer was retrieved from cache.

**cache-misses** (name `misses`, multiplier: `-1`)
Number of times an answer was not found in the cache.

---

## Server Peformance

family: `server` on all charts in this section

### CPU utilization

title: `DNSDIST server CPU utilization`
units: `ms/s`,
algorithm: `incremental`
type: `stacked`

**cpu-sys-msec**
Milliseconds spent by dnsdist in the “system” state.

**cpu-user-msec**
Milliseconds spent by dnsdist in the “user” state.

### Memory utilization

title: `DNSDIST server memory utilization`
units: `MB` (I don't know the units of the original metric, so please check),
algorithm: `absolute`
type: `area`

**real-memory-usage**
Current memory usage.


### Dynamic Blocks

**dyn-block-nmg-size**
Number of dynamic blocks entries.

### File Descriptors

**fd-usage**
Number of currently used file descriptors.

### Uptime

**uptime**
Uptime of the dnsdist process, in seconds.


---

## Latency

### Queries Latency

units: `queries/s`
algorithm: `incremental`
type: `stacked`

**latency0-1** as `1ms`
Number of queries answered in less than 1 ms.

**latency1-10** as `10ms`
Number of queries answered in 1-10 ms.

**latency10-50** as `50ms`
Number of queries answered in 10-50 ms.

**latency50-100** as `100ms`
Number of queries answered in 50-100 ms.

**latency100-1000** as `1sec`
Number of queries answered in 100-1000 ms.

**latency-slow** as `slow`
Number of queries answered in more than 1 second.


### Average Latency

title: `Average latency for the last N queries`
units: `ms/query`
algorithm: `absolute`
type: `line`

**latency-avg100** as `100`
Average response latency of the last 100 packets.

**latency-avg1000** as `1k`
Average response latency of the last 1000 packets.

**latency-avg10000** as `10k`
Average response latency of the last 10000 packets.

**latency-avg1000000** as `1000k`
Average response latency of the last 1000000 packets.

nobody-nobody:
Okay, just did some changes based on your input.

Regarding "real-memory-usage":
Values is in bytes. I set a divisor to show MiB.
Regarding "acl-drops":
Those are all kinds (not necessary only queries) of packets which are disallowed by acl.

Regarding "total number of answers":
I guess this could be calculated based on other fields but not sure if 100% accurate so maybe its better to leave this field for now...

What about dyn-block-nmg-size, fd-usage, Uptime?
You did not specify more information (e.g. titel) so I guess you are fine with not including them?
I think the ratio between "could be useful" / "wasted space" is not that good...

I was a bit in a hurry so I hope I did not overlook someting while testing.

ktsaou:
> Regarding "acl-drops":
> Those are all kinds (not necessary only queries) of packets which are disallowed by acl.

Yes, I understood that, but what else they could be? Is it possible to drop backend traffic with an acl?
Anyway, if this is not queries, I guess it should be at its own chart.

> Regarding "total number of answers":
> I guess this could be calculated based on other fields but not sure if 100% accurate so maybe its better to leave this field for now...

hm... probably, but this is not a good strategy, as it would require for the plugin to know how exactly the server works, it will most probably present wrong information as the server evolves, etc.

> What about dyn-block-nmg-size, fd-usage, Uptime?

If you add dnsdist process names to `/etc/netdata/apps_groups.conf`, apps.plugin will provide quite some information about them, including cpu usage, memory, virtual memory, number of processes, threads, sockets, files, pipes, disk activity, memory activity (page faults), etc.

For `dyn-block-nmg-size`, I don't know what exactly it is, so your call.

The same for `uptime`. At some point in future I think apps.plugin will offer something like a process watchdog, that will include this information too, based on what the kernel knows.

Is there a way to check the final result? If you have it installed privately, just post a few screenshots.

@l2isbad please review.

python.d/dnsdist.chart.py
```diff
@@ -0,0 +1,97 @@
...
+			['cache-misses', 'misses', 'incremental', -1]
+	]},
+	'servercpu': {
+		'options': [None, 'DNSDIST server CPU utilization', 'ms/s', 'server', 'dnsdist.servercpu', 'stacked'],
+		'lines': [
+			['cpu-sys-msec', 'system state', 'incremental'],
+			['cpu-user-msec', 'user state', 'incremental']
+	]},
+	'servermem': {
+		'options': [None, 'DNSDIST server memory utilization', 'MiB', 'server', 'dnsdist.servermem', 'area'],
```
l2isbad: `MB` for megabytes

l2isbad:
@ktsaou in general this is ready.

l2isbad:
> Is there a way to check the final result? If you have it installed privately, just post a few screenshots.

@nobody-nobody

nobody-nobody:
Hi, sorry I did not have time the last days.
There is a screenshot attached. Maybe a bit small but I had to zoom out (40%) to catch all graphs.

Also I did so minor changes:
- Added dnsdist to apps_groups
- Moved "acl-drops" into a own chart.

dnsdist

ktsaou:
It seems a lot better.
Could you please also provide a snapshot? If you don't want to attach it here, you can send it to my email: costa@tsaousis.gr

nobody-nobody:
Snapshot (replaced hostname) is attached.
But I was not able to load it (all charts were blank)... hope it works for you...?

[dnsdist.zip](https://github.com/firehol/netdata/files/1552517/dnsdist.zip)

ktsaou:
thanks!
I think this is very nice.

@l2isbad are you ok to merge?

@nobody-nobody are there any pending items related to documentation? Keep in mind you can add information above each dashboard sub-section and each chart, probably with some descriptive text and links to related documentation at the dnsdist site. Some high quality text, generally makes the dashboard a lot more professional. Your call...

Once you are ready, I think we can merge this.

l2isbad:
> are you ok to merge?

yes

ktsaou:
perfect!

@nobody-nobody let me know when you are ready...

nobody-nobody:
Yeah I thought about adding some description but... not sure which level of detail is needed.
If someone uses dnsdist in his network he should have (hopefully) at least basic knowledge about all those fields. And if he misses something he can check out the official documentation.
I guess all charts are pretty clear of what they are showing and I think its more valuable to have a "clean" view instead of additional (maybe useless) text :)
But this is just my own feeling about that. So if you want to have more text... its your baby so you should have the final call.
I am not using this plugin anymore (I disable python completely and went for an native plugin approach for my monitoring targets to save some resources and to be more flexible) but I will finish this one as good as possible so let me know if you would like to see something improved.

ktsaou:
ok, merged.

PR #3025, opened 2017-11-16
added -x for radclient binary call

wungad:
There are not stats from radiusd with freeradius 3.x server if -x is missing.


l2isbad:
thx @wungad 
please add 
```python
return None
```
after
https://github.com/firehol/netdata/blob/master/python.d/freeradius.chart.py#L93

wungad:
Done.

l2isbad:
@ktsaou ready

ktsaou:
merged! thanks!

PR #3033, opened 2017-11-18
Dynamic highlight of time-frames using the ALT or Control key

### LINKED ISSUE #2356: [BUG] Wrong selection in side menu after clicking an entry
Version: v1.6.0-553-g0d75bc6b_rolling
Browser: Safari Version 10.1.1 (12603.2.4)

Problem:
After clicking an entry in the Side menu the wrong menu item gets selected. (see Screenshot)

bildschirmfoto 2017-06-18 um 10 51 11

Memory was clicked and page jump works but the item above (processes) gets the selected menu item. This issue appears with all items.

ktsaou:
Pressing the ALT key and selecting a timeframe at a chart, highlights the same area on all charts, and shows at the top the selected timeframe.

The selected timeframe:

1. is added at the URL hash, so that hitting F5 will preserve it.
2. is transferred among netdata servers, via the registry
3. is maintained when panning and zooming, even if the timeframe is not visible
4. can be dismissed by hitting the X at the overlay

screenshot from 2017-11-18 03-42-15

screenshot from 2017-11-18 03-44-32

---

Fixes a bug in scrollspy that was selecting the wrong menu item; fixes #2356 

---

Rewrote the page scroll handlers to support passive events. There is still an issue with chrome that is somehow randomly delaying events by 2 seconds.

Now netdata renders the charts when the window does not have focus, but the page is scrolled.

Scrolling `sync` and `async` modes have been reworked. netdata scrolling is amazing again! Especially on firefox v57, `sync` scrolling is wow!

---

Registry items now change URL every time the `my-netdata` menu is pressed. This allows propagating the URL hash (pan and zoom, highlighted timeframe, etc) to other netdata servers, by pressing right mouse button on a registry item and selecting `open in new window/tab`.

---

many snapshot improvements:

1. now it allows only one snapshot save at a time (previously a double click on the **Export** button triggered 2 saves at the same time.

2. now it allows only one snapshot load at a time (like before)

3. fixed an issue in snapshots that prevented certain charts to be saved in the snapshot if they had outstanding network connections to be refreshed (now the connections are aborted)

### DIFF
--- a/web/dashboard.js
+++ b/web/dashboard.js
@@ -2326,6 +2326,7 @@ var NETDATA = window.NETDATA || {};
             that.element_chart = null; // the element with the chart
             that.element_legend = null; // the element with the legend of the chart (if created by us)
             that.element_legend_childs = {
+                content: null,
                 hidden: null,
                 title_date: null,
                 title_time: null,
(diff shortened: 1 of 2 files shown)


ktsaou:
peek 2017-11-19 19-39

ktsaou:
peek 2017-11-19 20-10

PR #3050, opened 2017-11-22
isc_dhcpd epoch

wungad:
Something to scratch off to-do list:
- lease_end_time in epoch is now correctly parsed
- infinite lease time (0xFFFFFFFF) causes dhcpd to output "ends: never" in the lease block


wungad:
fixed codacy warnings and and created new request.

PR #3055, opened 2017-11-22
snapshots and highlight improvements

ktsaou:
1. the overlay of the highlighted time-frame now allows resetting the view back the one that was used to highlight the time-frame. ~~pending: fix this functionality for snapshots (it fails).~~ (fixed)

2. loading snapshots can also be done by dropping files at the load snapshot modal.

3. loading snapshots that were taken from a netdata behind some proxy did not work. Now it works. ~~There is still a problem though, that the chartRegistry is maintaining somehow a stripped URL.~~ (fixed)

4. all tooltips are now bootstrap tooltips.

5. alarms modal now shows a thumbs up icon when no alarms are raised.

6, re-wrote the highlighted time-frame functionality to solve all corner cases - it is now a lot cleaner.

### DIFF
--- a/web/dashboard.js
+++ b/web/dashboard.js
@@ -1154,14 +1154,7 @@ var NETDATA = window.NETDATA || {};
             this.charts = {};
         },
 
-        fixid: function(id) {
-            return id.replace(/:/g, "_").replace(/\//g, "_");
-        },
-
         add: function(host, id, data) {
-            host = this.fixid(host);
-            id   = this.fixid(id);
-
             if(typeof this.charts[host] === 'undefined')
                 this.charts[host] = {};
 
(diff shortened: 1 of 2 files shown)


l2isbad:
@ktsaou 
highlight is not visible on stacked charts

screenshot_20171123_104940

ktsaou:
Yes, I know.

more intense white (I don't like it):

screenshot from 2017-11-23 03-57-44


darker highlight (I don't like it):

screenshot from 2017-11-23 03-59-12


Lowered the opacity of stacked charts, only when a highlight is active:

screenshot from 2017-11-23 04-02-26

but it remains quite dim.

The problem is on just a few charts that are full and constant. So I though it is pretty acceptable the way it is now.

Another solution would be to add two vertical very intense black or white lines at the left and the right of the highlighted area, probably only on stacked charts, but I haven't had the chance to try it yet.

l2isbad:
> The problem is on just a few charts that are full and constant. So I though it is pretty acceptable the way it is now.
> 
> Another solution would be to add two vertical very intense black or white lines at the left and the right of the highlighted area, probably only on stacked charts, but I haven't had the chance to try it yet.

👍

PR #3066, opened 2017-11-27
dashboard scroll improvement and more

ktsaou:
Browsers (especially chrome) throttle `setTimeout()` when intensive page work is in place. This often results in a 2-second timeout to load a chart when scrolling.

To work-around this issue, netdata now uses a custom `setTimeout()` function that works with `requestAnimationFrame()`, which is guaranteed to fire 60 times per second (60 fps).

So now, as far as the browser fires this event, netdata charts get updated on time.

This change allowed the `sync` scrolling function of netdata to work mid-scrolling. So, while the page is scrolled, charts get updated. On firefox, it seems the browser disables javascript during scroll, so this is not entirely achieved (it seems it depends on the speed of the scrolling).

To improve things even more, netdata now provides a few hints to the browser about the visible charts, in hope that the browser will move each chart at its own compositing space at the GPU. My (visual) tests on slow computers showed that this improved hover significantly.

---

Under Linux GPU accelerated rendering is disabled by default on chrome. This is quite a big issue, especially when the browser is fullscreen on HiDPi screens. There is no workaround. Scrolling on such a display is problematic, until GPU acceleration is enabled.

It can be enabled by `chrome://flags` and setting:

1. Override software rendering list `ignore-gpu-blacklist`
2. GPU rasterization `enable-gpu-rasterization`

There is also "Zero-copy rasterizer" `enable-zero-copy`, which seems unsupported (compile time).

You can verify your settings at `chrome://gpu`:

screenshot from 2017-11-27 02-07-50

---

Page scrolling is now prevented when a modal is open.

---

netdata now stops updating the charts, while a modal is open.

### DIFF
--- a/web/dashboard.css
+++ b/web/dashboard.css
@@ -44,6 +44,8 @@ body {
     display: inline-block;
     overflow: hidden;
 
+    transform: translate3d(0,0,0);
+
     /* required for child elements to have absolute position */
     position: relative;
 
@@ -53,6 +55,8 @@ body {
 .netdata-container-gauge {
     display: inline-block;
     overflow: hidden;
+
+    transform: translate3d(0,0,0);
 
     /* required for child elements to have absolute position */
     position: relative;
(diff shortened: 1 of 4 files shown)


ktsaou:
The codacy error is false positive. The function it complains about, is 5 lines long - no complexity at all.

PR #3114, opened 2017-12-09
simple pattern enhancements and pattern based statsd synthetic charts dimensions

### LINKED ISSUE #2785: synthetic statsd charts with pattern-based dimensions
Synthetic statsd charts can only be based on a fixed list of dimensions defined at start time.
In some scenarios, the actual possible dimensions are not known or simply infinite. It would make sense to have:

```
# replace 'mychart' with the chart id
# the chart will be named: myapp.mychart
[mychart]
	name = mychart
	title = my chart title
	family = my family
	context = chart.context
	units = tests/s
	priority = 91000
	type = area
	dimension = myapp.collection.*.metric1
```

which would gather
```
myapp.collection.item1.metric1
myapp.collection.item2.metric1
myapp.collection.item3.metric1
myapp.collection.otheritem.metric1
...
```

including metrics that do not exist when the configuration file is loaded.
The context could then be further used to define an alarm on all these metrics.

ktsaou:
- [x] Use `\ ` (backslash + space) to match a space in simple patterns.
- [x] Simple patterns can now extract the wildcarded part from from the matched string. This means that `string.*` can extract the string that the `*` matches.
- [x] Statsd synthetic charts now support pattern based dimensions. Prepend the keyword `pattern` as the first parameter of the `dimension` line (the rest of the parameters remain the same). The metric name will now be considered a simple pattern list.

   Example:

```
dimension = pattern my.*.counter '' last 1 1
```

  The above will match `my.little.counter` and name it `little`, `my.big.counter` and name it `big`, etc.

fixes #2785 

- [x] Statsd synthetic charts now support a dictionary for naming dimensions. So, if you need to name statsd metric `my.little.counter` to `key events` on all synthetic charts, you can add it at the `[dictionary]` section, like this:

```
[app]
   ...

[dictionary]
   my.little.counter = key events

[chart1]
   ...
   dimension = my.little.counter '' last 1 1

[chart2]
   ...
   dimension = my.little.counter '' events 1 1
```
   
   The dimensions of both charts will be named `key events`.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -771,18 +771,19 @@ int main(int argc, char **argv) {
 
                             const char *heystack = argv[optind];
                             const char *needle = argv[optind + 1];
+                            size_t len = strlen(needle) + 1;
+                            char wildcarded[len];
 
-                            SIMPLE_PATTERN *p = simple_pattern_create(heystack
-                                                                      , SIMPLE_PATTERN_EXACT);
-                            int ret = simple_pattern_matches(p, needle);
+                            SIMPLE_PATTERN *p = simple_pattern_create(heystack, SIMPLE_PATTERN_EXACT);
+                            int ret = simple_pattern_matches_extract(p, needle, wildcarded, len);
                             simple_pattern_free(p);
 
                             if(ret) {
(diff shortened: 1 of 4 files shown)


ktsaou:
updated the wiki too: https://github.com/firehol/netdata/wiki/statsd

PR #3125, opened 2017-12-11
do not match a zero value with null color; 

### LINKED ISSUE #3123: Cannot display absolute value in badge coloured by original value
I have a badge that I want to display as an absolute value, but want to colour it depending on what the original value was e.g. green if it was positive, red if it was negative. We don't have any colour-blind people viewing my dashboard ...

Currently the `value_color` option applies against the displayed value, so a URL of the form:

`/api/v1/badge.svg?&chart=mychart&dimension=mydim&options=abs&value_color=grey:null|%235cb85c>0|%23cb3935`

will only ever be grey or green, never red. It would be good to have an option like `options=display_absolute` so only the display value is affected.

### LINKED ISSUE #3124: Zero values are considered null in badge colour conditions
If you have a badge `value_color` like:

`value_color=grey:null|%235cb85c>=0|%23cb3935`

then a zero value will be treated as null instead of obeying the second condition. `null` should IMO only refer to when data isn't available for the dimension.

I'm guessing the null check is simply doing a truth test and both empty data and zero evaluate to false.

ktsaou:
Before this PR badge colors were incorrectly matching zero values as `null`.

fixes #3124

Also added equal comparison with `=` (was using only `:`) and not equal comparison with `!=` and `<>`.

---

Added API option `display-absolute` (and `display_absolute`) to use the signed value at color calculation, but display the absolute value on the badge.

fixes #3123

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1000,6 +1000,7 @@ int main(int argc, char **argv) {
         // --------------------------------------------------------------------
         // create the listening sockets
 
+        web_client_api_v1_init();
         web_server_threading_selection();
 
         if(web_server_mode != WEB_SERVER_MODE_NONE)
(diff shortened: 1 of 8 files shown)


magao:
Fantastic. Really looking forward to 1.9.

PR #3136, opened 2017-12-13
both installers download configuration netdata

ktsaou:
and provide a sane default when this is not possible.
also the static installer lsm states the correct version of netdata.

### DIFF
--- a/installer/functions.sh
+++ b/installer/functions.sh
@@ -640,6 +640,84 @@ install_netdata_logrotate() {
 }
 
 # -----------------------------------------------------------------------------
+# download netdata.conf
+
+fix_netdata_conf() {
+    local owner="${1}"
+
+    if [ "${UID}" -eq 0 ]
+        then
+        run chown "${owner}" "${filename}"
+    fi
+    run chmod 0664 "${filename}"
+}
+
+generate_netdata_conf() {
+    local owner="${1}" filename="${2}" url="${3}"
+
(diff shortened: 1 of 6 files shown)


tisc0:
\o/ testing tomorrow

tisc0:
Tested that : 

>  curl -Ss https://my-netdata.io/kickstart-static64.sh -o netdata-kickstart-static64.sh
>  bash netdata-kickstart-static64.sh --dont-wait
Doesn't seem to be concerned by it, since the install is in /opt by default. At least no /etc/netdata DIR is created.

ktsaou:
the static 64bit installer is built using `makeself`, which unfortunately uses a different set of command line switched. makeself prompts for acceptable of the license. It can by accepted by adding the `--accept` command line switch. Run it with `-h` to check its other options.

tisc0:
Found it... I misused /etc/profile.d to put some script started everytime I open a session... and also when scripts like yours are really working in the system. 
The exit 1 came from a script of mine in between, thanks bash -x (We do better when we've slept a bit).
The weird is not anymore, victory of the Light again \o/

ktsaou:
> But you might have changed something

Whatever I changed is recorded... ( open-source software does that :smile: ) but I don't think so. I don't recall changing anything on the installers since v1.8, before this PR.

tisc0:
I admit, don't know much that deep of your code...  :)
Getting the liberty to close this issue..
Thanks again !

PR #3140, opened 2017-12-15
add chart for redis evited keys

bbaronSVK:


python.d/redis.chart.py
```diff
@@ -18,10 +18,15 @@
 #             'unix_socket': None
 #          }}
 
-ORDER = ['operations', 'hit_rate', 'memory', 'keys', 'net', 'connections', 'clients', 'slaves', 'persistence',
+ORDER = ['operations', 'hit_rate', 'memory', 'keys', 'net', 'connections', 'clients', 'eviction', 'slaves', 'persistence',
          'bgsave_now', 'bgsave_health']
 
 CHARTS = {
+    'eviction':{
+        'options': [None, 'Evicted keys', 'keys/s', 'evicted', 'redis.evictios', 'line'],
```
l2isbad: `Evicted Keys`
Are you sure about `absolute`? If so then `keys/s` is wrong should be `keys`.

bbaronSVK:
You are right, i fix it.

python.d/redis.chart.py
```diff
@@ -18,10 +18,15 @@
 #             'unix_socket': None
 #          }}
 
-ORDER = ['operations', 'hit_rate', 'memory', 'keys', 'net', 'connections', 'clients', 'slaves', 'persistence',
+ORDER = ['operations', 'hit_rate', 'memory', 'keys', 'net', 'connections', 'clients', 'eviction', 'slaves', 'persistence',
          'bgsave_now', 'bgsave_health']
 
 CHARTS = {
+    'eviction':{
+        'options': [None, 'Evicted keys', 'keys', 'evicted', 'redis.evictios', 'line'],
```
l2isbad: sorry for not being clear. `keys/s` is ok when algorithm is `incremental` and wrong if `absolute`. Netdata applies interpolation to find the exact value at second-boundaries only for `incremental` values, so `/s` is misleading for `absolute`.

Now units should be `keys/s` 😄

l2isbad: https://github.com/firehol/netdata/pull/1758#issuecomment-279153641

python.d/redis.chart.py
```diff
@@ -18,10 +18,15 @@
 #             'unix_socket': None
 #          }}
 
-ORDER = ['operations', 'hit_rate', 'memory', 'keys', 'net', 'connections', 'clients', 'slaves', 'persistence',
+ORDER = ['operations', 'hit_rate', 'memory', 'keys', 'net', 'connections', 'clients', 'eviction', 'slaves', 'persistence',
          'bgsave_now', 'bgsave_health']
 
 CHARTS = {
+    'eviction':{
+        'options': [None, 'Evicted keys', 'keys', 'evicted', 'redis.evictios', 'area'],
```
l2isbad: > redis.evictios

typo

ktsaou:
merged it

PR #3159, opened 2017-12-20
Update BTRFS chart descriptions.

Ferroin:
This adds some info on what the values in the graphs should ideally look like, and a bit on how to keep them that way, namely:

* Suggest some degree of slack space that should be present in the physical usage chart.  The particular 5% reference is based on my own experience.  Also refers to `btrfs balance` as the way to help keep things that way, with a reference to the man page for the balance command.
* Suggestions for data and metadata are based on the ideal standard of having no more than one chunk partially full (Data chunks are nominally 1-5GB, metadata are nominally 256MB to 1GB, but need more slack space to account for the Global Reserve).
* The info added for system chunks gives descriptions of expected behavior only.  The system chunks only store the tree that tracks the locations of chunk allocations on the FS, so they are generally not an issue (16K of space here holds enough room for approximately 663 chunks to be tracked, which translates to roughly 500 GB of allocated space).
* I've also removed references to RAID in the descriptions.  Strictly speaking, BTRFS replication and striping are _not_ RAID, and there's been a lot of noise recently on the BTRFS mailing list regarding this and the confusion the profile naming has caused, so I would suggest not referring to things as RAID.

If desired, I can update things with more specific advice on fixing things, but that is liable to get very long and involved, and I personally don't see much value in doing so (given that the best way to fix things is somewhat dependent on the design of the system in question and exactly how things have failed).

### DIFF
--- a/web/dashboard_info.js
+++ b/web/dashboard_info.js
@@ -1670,18 +1670,18 @@ netdataDashboard.context = {
     },
 
     'btrfs.disk': {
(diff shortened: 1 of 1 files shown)


Ferroin:
Forgot to mention in the PR description, this was requested in the discussion around PR #3150

ktsaou:
Thanks!

> The system chunks only store the tree that tracks the locations of chunk allocations on the FS

nice! I searched for it info without luck.

Could you please add such short descriptions of `data`, `metadata` and `system` above the corresponding charts?

Ferroin:
Done, albeit in a way that tries not to sound quite so technical.

I also normalized spacing between sentences so that things look a bit nicer in the code itself.

ktsaou:
perfect! thank you!

PR #3162, opened 2017-12-20
statsd: support dimension flags, like in plugins

ktsaou:


ktsaou:
updated the wiki too: https://github.com/firehol/netdata/wiki/statsd and the changelog https://github.com/firehol/netdata/issues/3143

PR #3166, opened 2017-12-21
Add ceph plugin

lets00:
This pull request add ceph plugin support.

image
I would like suggestions on how to improve this code (first time that I collaborate with this amazing project).

Actually, this plugin collects the following metrics:
- general status
- pool usage
- pool objects' number
- osd usage

The precision supported are: KB, MB, GB and TB.
Ceph versions tested:
-  Luminous
- Jewel


conf.d/python.d/ceph.conf
```diff
@@ -0,0 +1,5 @@
+update_every : 10
+local:
+  config_file: '/etc/ceph/ceph.conf'
+  keyring_file: '/etc/ceph/ceph.client.admin.keyring'
+  presentation_data: 'KB'
```
l2isbad: 1. I suggest to move` update_every : 10` to the module file (`.py`).
2. `keyring_file`. Does it world readable? Do we need to add netdata to `ceph` group or something?
3. `presentation_data`. Netdata has units scaling feature `KB-MB-GB`. I think data should be in KB always.

And please copy any module conf file and modify it for ceph. It should contain all the stuff.

python.d/ceph.chart.py
```diff
@@ -0,0 +1,183 @@
...
+        # Set default presentation_data to GB if it does not defined and declare
+        # presentation_value with PRESENTATION_MAP value.
+        if not self.presentation_data:
+            self.presentation_data = 'GB'
+        self.presentation_value = PRESENTATION_MAP[self.presentation_data]
+        try:
+            self.cluster.connect()
+        except Exception as error:
+            self.error(error)
+            return False
```
l2isbad: Do we need to support reconnect?
like
https://github.com/firehol/netdata/blob/master/python.d/beanstalk.chart.py#L239-L245

l2isbad:
gj @lets00

python.d/ceph.chart.py
```diff
@@ -0,0 +1,183 @@
...
+    def _get_general(self):
+        """
+        Get ceph's general usage
+        :return: dict
+        """
+        info = self.cluster.get_cluster_stats()
+        # Values are get in bytes, convert it
+        return {'general_total': float(info['kb']),
+                'general_usage': float(info['kb_used']),
+                'general_available': float(info['kb_avail'])
```
l2isbad: https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L312-L316

> Only integer values are collected. If you want to push fractional values, multiply this value by 100 or 1000 and set the DIMENSION divider to 1000.

python.d/ceph.chart.py
```diff
@@ -0,0 +1,183 @@
+# -*- coding: utf-8 -*-
+# Description: ceph netdata python.d module
+# Author: Luis Eduardo (lets00)
+
+import rados
```
l2isbad: wrap it in `try/except`
like
https://github.com/firehol/netdata/blob/master/python.d/beanstalk.chart.py#L5-L9

python.d/ceph.chart.py
lets00: I forgot to remove the float casting. I will fix this.

python.d/ceph.chart.py
lets00: Ok, this avoid ImportError Exception. I will fix this.

conf.d/python.d/ceph.conf
lets00: Hi @l2isbad , thanks for the tips/suggestions. I really appreciate it!
1. I will put this on module file.
2. You are right. Moreover, you need change keyring access to group users can be used this file (I add netdata user and change permission file to 640). I need document it in some place. First, I will document on config file, and add a new session on (https://github.com/firehol/netdata/blob/master/python.d/README.md) explain ceph.
3. I define this variable because the default value is KB. In test environment, my ceph cluster contains 8GB and the presantation data is huge number. In my production environment, my ceph cluster contains +100TB and was impractible the huge number that graph shows. If no one presentation_data is defined, for default, it show this data on KB. But I will test again in my production environment to grab screenshots.

python.d/ceph.chart.py
lets00: I will see if this is possible.

conf.d/python.d/ceph.conf
l2isbad: > I define this variable because the default value is KB. In test environment, my ceph cluster contains 8GB and the presantation data is huge number. In my production environment, my ceph cluster contains +100TB and was impractible the huge number that graph shows. If no one presentation_data is defined, for default, it show this data on KB. But I will test again in my production environment to grab screenshots.

> When set to Scale Units the values shown will dynamically be scaled (e.g. 1000 kilobits will be shown as 1 megabit). Netdata can auto-scale these original units: kilobits/s, kilobytes/s, KB/s, KB, MB, and GB. When set to Fixed Units all the values will be rendered using the original units maintained by the netdata server. 

So units should be `KB` not `usage(whatever)`

lets00: I see. But, when I get data on KB, the graph does not convert to other unity (like MB, GB, TB). For example: I caught Total usage of cluster on KB and my cluster have some 32/33 GB. The image shows how data is presented:
image

33.140.660 is not well presented data. I could present it on another scale (GB, for example: 32,9). This was the reason to I use this approach.

How do I set Scale Units? Do you have some example? Thank you for the help.

l2isbad: I see units is `usage` which is wrong. 

`usage` => `KB`

And dimensions should be `free` and `used` (instead `Total Usage Free`) and chart type 'stacked' 

`line` => `stacked`

```python
'general': {
    'options': [None, 'Ceph General Options', 'KB', 'general', 'ceph.general', 'stacked'],
    'lines': []
```

lets00: I replace ```usage``` => ```KB``` and works to me (Auto-scaling). Thank you. I am ending last issues to be fix.

l2isbad:
@lets00 i see `btrfs` stuff was deleted. Please do a rebase.
To add `netdata` to `ceph` group please update:
1. https://github.com/firehol/netdata/blob/master/installer/functions.sh#L721-L743
2. https://github.com/firehol/netdata/blob/master/netdata-installer.sh#L1101-L1108

k0ste:
I have a couple of thoughts on this. I believe that you do not need to add pools statistics. If `netdata` is started on 50 hosts it turns out that 50 times the same metric is collected. Information "about the cluster" is already available in the `ceph-mgr` [prometeus module](//github.com/ceph/ceph/blob/master/src/pybind/mgr/prometheus/module.py). So really it will be nice if `netdata` plugin will care about `ceph-mgr` labels and provide easy access ceph-osd metrics. "about the cluster" metrics I think is the second part of `netdata` metrics and should not be enabled by default, because usual osd-hosts don't have client access to cluster and we don't need metrics overhead I described above.

I believe that `netdata` must have information about osd_host metrics:

```yaml
netdata:
- osd:
  - osd_usage_bytes
  - osd_total_bytes
- osd_perf:
  - osd_commit_latency
  - osd_iowait
  - osd_util
  - osd_read_bytes
  - osd_write_bytes
  - osd_iops_read
  - osd_iops_write
```

Yes, most of them is usual disk usage. But, if watch this in `netdata` you should always care about "what disk is what osd". Some folks use dm-cache\lvm\bcache, so at this situation "what is osd" is not easy to scroll `netdata` page.
Also ceph-mgr honor labes for node_exporter. So also not easy to access to `netdata` labels.
If node exporter used, you can use disk perf metrics like this:

```python
node_disk_bytes_written and on (device,instance) ceph_disk_occupation{ceph_daemon="osd.0"}
```

And with netdata:

First, we need relabel netdata label 'family' (`sda`, `sdb`, etc) to 'device' label, because prometheus can't join another label on query: https://github.com/prometheus/prometheus/issues/2204

```yaml
- job_name: 'netdata'                     
  honor_labels: true                      
  metrics_path: '/api/v1/allmetrics'      
  params:                                 
    format: [prometheus]                  
  static_configs:                         
  - targets:                              
    - '192.168.100.11:19999'              
    labels:                               
      instance: 'ceph-osd0'               
      group: 'ceph_hosts'                           
  metric_relabel_configs:                 
  - source_labels:                        
    - 'family'                            
    separator: ;                          
    target_label: 'device'                
    action: 'replace'                     
```

Second is query:

```python
(netdata_disk_io_kilobytes_persec_average{dimension="writes"} and on (instance,device) ceph_disk_occupation{ceph_daemon=~"$osd"}) * -1
```

`netdata` use `negative` values for network/disk sent/write metrics. And this is very good for `netdata` dashboard, but very bad (imho) for Grafana. So I use multiplier to convert negative value to positive.

lets00:
Hi @k0ste, really good, thank you for the suggest. I have some comments to made it:

> I have a couple of thoughts on this. I believe that you do not need to add pools statistics. If netdata is started on 50 hosts it turns out that 50 times the same metric is collected.

Thats is good think when we need install netdata in all ceph machines with ceph module active in all machines, but, for example, in my production environment, I have ceph client (using ceph cli) in a group of same machine (3 ceph-monitors). So, I collect all information about the ceph cluster in a unique machine in this group. In others machines (ceph-osds) I not caught information about ceph cluster.

One way to implement it (that I can see) is:

- In ceph-mons, I could expose all information about the ceph cluster (Pools, osd usage, status, etc) and in ceph-osds nodes just present informations about osds (avoid repeated information).

> Yes, most of them is usual disk usage. But, if watch this in netdata you should always care about "what disk is what osd". Some folks use dm-cache\lvm\bcache, so at this situation "what is osd" is not easy to scroll netdata page.

I agree. Is most difficult just seeing dist status know if is be used is a OSD or not (or what disk is what osd). I will see how to obtain this metrics. Maybe I can use the disk metric code used in ```netdata```

> Information "about the cluster" is already available in the ceph-mgr prometeus module. So really it will be nice if netdata plugin will care about ceph-mgr labels and provide easy access ceph-osd metrics. "about the cluster" metrics I think is the second part of netdata metrics and should not be enabled by default, because usual osd-hosts don't have client access to cluster and we don't need metrics overhead I described above.

The ceph-mgr is good but it does not available in versions before kraken (in kraken, is optional). In Luminous, it is required (http://docs.ceph.com/docs/master/mgr/). My major environment (production) is Jewel version and not support ```ceph-mgr```. In this code, I do not use ```ceph-mgr``` to obtain metrics, but I will see what I could use ```ceph-mgr``` to obtain this metrics that you suggest (could be more easy that consult rados lib) using ```prometheus``` or ```telegraph``` (if they collect this metrics and it is better than collect without this plugins).

python.d/ceph.chart.py
```diff
@@ -0,0 +1,165 @@
...
+        Create dynamically charts options
+        set on YAML file
+        :return: None
+        """
+        # Lines
+        for pool in self._get_df()['pools']:
+            self.definitions['pool_usage']['lines'].append([pool['name'],
+                                                            pool['name'],
+                                                            'absolute'])
+            self.definitions['pool_objects']['lines'].append(["obj_{}".format(pool['name']),
```
l2isbad:
```python
"obj_{}".format(pool['name'])
```
is `ValueError` in py2.6, please do `{0}` or `{name}`

python.d/ceph.chart.py
```diff
@@ -0,0 +1,165 @@
...
+        return pool_usage
+
+    def _get_pool_objects(self):
+        """
+        Process raw data into pool usage dict information
+        :return: A pool dict with pool name's key and object numbers
+        """
+        pool_objects = {}
+        for df in self._get_df()['pools']:
+            pool_objects["obj_{}".format(df['name'])] = df['stats']['objects']
```
l2isbad: same here

python.d/ceph.chart.py
```diff
@@ -0,0 +1,165 @@
...
+retries = 60
+
+ORDER = ['general', 'pool_usage', 'pool_objects', 'osd_usage']
+
+CHARTS = {
+    'general': {
+        'options': [None, 'Ceph General Options', 'KB', 'general', 'ceph.general', 'stacked'],
+        'lines': [
+            ['general_usage', 'Usage', 'absolute', 1, 1024],
+            ['general_available', 'Available', 'absolute', 1, 1024]
```
l2isbad: Dimension order is a thing. 

> This logic currently says that colors are in pairs: positive, negative, positive, negative, etc. Positive can be anything (green, blue, bright yellow, etc), but negative have more red in them (red, brown, dark purple, etc).

`Available` should be the first (green). And please make it lowercase.
like
`avail`
`used`

l2isbad:
Are you done @lets00 ?

lets00:
@l2isbad Why the name "local" appear after the "Ceph" name?

image

l2isbad:
@lets00 
because of
https://github.com/lets00/netdata/blob/ceph/conf.d/python.d/ceph.conf#L73

If it's always `local` and there is no auto-detection jobs i suggest to change

```yaml
local:
  config_file: '/etc/ceph/ceph.conf'
  keyring_file: '/etc/ceph/ceph.client.admin.keyring'
```
by

```yaml
config_file: '/etc/ceph/ceph.conf'
keyring_file: '/etc/ceph/ceph.client.admin.keyring'
```

lets00:
@l2isbad is done. But I do not know why Codacy /PR up this error, because I just remove a field on ceph.conf and before CIs voted ok.

ktsaou:
This codacy error is ignorable.

@lets00 nice job!
@l2isbad let me know when you believe this can be merged.

python.d/ceph.chart.py
```diff
@@ -0,0 +1,164 @@
...
+
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.config_file = self.configuration.get('config_file')
+        self.keyring_file = self.configuration.get('keyring_file')
+        self.cluster = rados.Rados(conffile=self.config_file,
+                                   conf=dict(keyring=self.keyring_file))
```
l2isbad: Сan this cause an exception?
I mean if there is no `self.config_file` or `self.keyring_file` or `chmod 640 /etc/ceph/ceph.client.admin.keyring` not done.

python.d/ceph.chart.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        if not CEPH:
+            self.error('rados module is needed to use ceph.chart.py')
+            return False
+        if not (self.config_file and self.keyring_file):
+            self.error('config_file and/or keyring_file is not defined')
+            return False
+        try:
+            self.cluster.connect()
+        except Exception as error:
```
l2isbad: I think `self.cluster.connect()` should cause only rados related exceptions (d) so should be
```python
 except rados.Error as error
```

lets00: Makes sense. I will change this to treat only rados exception.

python.d/ceph.chart.py
l2isbad: that is what i got
```python
>>> import rados
>>> cluster = rados.Rados(conffile=None, conf=dict(keyring=None))
  File "rados.pyx", line 525, in rados.Rados.__init__ (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:6445)
  File "rados.pyx", line 425, in rados.requires.wrapper.validate_func (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:4657)
  File "rados.pyx", line 571, in rados.Rados.__setup (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:7612)
  File "rados.pyx", line 424, in rados.requires.wrapper.validate_func (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:4634)
  File "rados.pyx", line 412, in rados.requires.check_type (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:4010)
TypeError: val must be basestring
>>> cluster = rados.Rados(conffile=str(), conf=dict(keyring=str()))
  File "rados.pyx", line 525, in rados.Rados.__init__ (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:6445)
  File "rados.pyx", line 425, in rados.requires.wrapper.validate_func (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:4657)
  File "rados.pyx", line 568, in rados.Rados.__setup (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:7414)
  File "rados.pyx", line 425, in rados.requires.wrapper.validate_func (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:4657)
  File "rados.pyx", line 631, in rados.Rados.conf_read_file (/build/ceph-HiUCwo/ceph-10.2.5/src/build/rados.c:8591)
rados.Error: error calling conf_read_file: error code 22
```

l2isbad: And
```python
>>> cluster = rados.Rados(conffile='/etc/network/interfaces', conf=dict(keyring=str()))
>>> 
```
😄

lets00: Oh, I got confused. Sorry about that =(. I tested this on passed, but not remember that this could be occour. 
In fact, the exception is only raised on self.cluster.connect() when the file (/etc/ceph/keyring...) does not correct permission. If conffile or keyring is None, an exception is raised before (when you create the cluster object). I will fix this.

k0ste:
> @l2isbad is done.

@lets00 it will be nice to see per pool client io stat: `ceph osd pool stats --format=json-pretty`

```json
[
    {
        "pool_name": "rbd_pool",
        "pool_id": 1,
        "recovery": {},
        "recovery_rate": {},
        "client_io_rate": {
            "read_bytes_sec": 982764,
            "write_bytes_sec": 1618003,
            "read_op_per_sec": 557,
            "write_op_per_sec": 212
        }
    }
]
```

And also osd performance dump `ceph osd perf --format=json-pretty`

```json
{
    "osd_perf_infos": [
        {
            "id": 53,
            "perf_stats": {
                "commit_latency_ms": 3,
                "apply_latency_ms": 3
            }
        },
        {
            "id": 26,
            "perf_stats": {
                "commit_latency_ms": 0,
                "apply_latency_ms": 1
            }
        },
        {
            "id": 27,
            "perf_stats": {
                "commit_latency_ms": 2,
                "apply_latency_ms": 2
            }
        }
    ]
}
```

lets00:
Cool @k0ste, great feature! I will see if there are this fields on Jewel and Luminous and implement it. Rados interface must be offer this values. After, I can verify in my production environment this values in graph. Nice!

l2isbad:
still wip, right @lets00 ?

lets00:
Yes @l2isbad. I had to stop a while ago, but I am finishing now.

lets00:
@l2isbad some more observation?

python.d/ceph.chart.py
```diff
@@ -0,0 +1,308 @@
...
+        commit_latency = 0
+
+        for pool_rw_io_b in pool_stats:
+            read_bytes_sec = read_bytes_sec + pool_rw_io_b['client_io_rate'].get('read_bytes_sec', 0)
+            write_bytes_sec = write_bytes_sec + pool_rw_io_b['client_io_rate'].get('write_bytes_sec', 0)
+            read_op_per_sec = read_op_per_sec + pool_rw_io_b['client_io_rate'].get('read_op_per_sec', 0)
+            write_op_per_sec = write_op_per_sec + pool_rw_io_b['client_io_rate'].get('write_op_per_sec', 0)
+        for perf in osd_perf['osd_perf_infos']:
+            apply_latency = apply_latency + perf['perf_stats']['apply_latency_ms']
+            commit_latency = commit_latency + perf['perf_stats']['commit_latency_ms']
```
l2isbad: Use `+=`, like
```python
commit_latency += perf['perf_stats']['commit_latency_ms']
```

l2isbad:
And look at codacy errors. 
You can fix `Method could be a function` by adding  decorator
```python
@staticmethod
```
`Unused variable '__class__'` ignorable.

lets00:
Some images:
Ceph general:
image

image

Pools:
image

OSD:
image
image

l2isbad:
@lets00  gj 👍 

> This branch has conflicts that must be resolved

looks like it's time for one more rebase 😄 

And last but not least - can you suggest some alarms?

ktsaou:
This is nice! 👍 

One very minor suggestion:

@lets00 you have added info above the charts like this: `Shows osd usage...`, `Shows ...`, `Shows ...`, etc.

I suggest to switch it to `OSD space allocation...` (i.e. remove `Shows`).

It would also be nice if you could add a sentence more to explain what is this and how this information should be interpreted. Very basic things, like what is OSD and why should someone care about its space allocation.

Similarly for the rest of the charts...

The idea is to help people understand...

lets00:
@l2isbad I not use alarm on netdata, but I will see how implement it in netdata (If you have an example,  it would help me).

lets00:
@ktsaou Ok. I will describe better the information about charts and how this information should be interpreted. Actually, the description is very short. I will solve this. Thank you for the tips!

l2isbad:
@lets00 
https://github.com/firehol/netdata/blob/master/conf.d/Makefile.am#L74-L79

lets00:
@l2isbad Added.

I do not add this references (health.d/ceph.conf and python.d/ceph.conf) on [configs.signatures](https://github.com/firehol/netdata/blob/b4408fded4075271fedb85441bc9d8602d068da9/configs.signatures) file because that I do not know how generate this hash (I suppose that this is a hash generate maybe using md5).

l2isbad:
@ktsaou ^^

ktsaou:
To update the configs signatures, you can run:

```sh
./netdata-installer.sh get_git_config_signatures
```

Keep in mind this will checkout all versions of the configs from the git log, overwriting any uncommitted changed you have - so, commit everything before running it.

lets00:
@ktsau. Done.
Is this made to ensure authenticity?

ktsaou:
> Is this made to ensure authenticity?

This is used to check if the installed config files have been changed by the user. If they have, they are not overwritten during netdata updates.

So, are we ready? merge?

lets00:
@ktsaou yes, I'm finished. You can merge it.

ktsaou:
merged! Thank you!

ktsaou:
This module is missing from https://github.com/firehol/netdata/wiki/Add-more-charts-to-netdata

ktsaou:
I added it to the wiki menu and to the infographic.

PR #3170, opened 2017-12-23
the dimensions list of API calls now supports netdata simple patterns

ktsaou:
There are cases we need to say, "return all dimensions except this and that".

Before this PR, the only way would be to give all the dimensions that should be returned in a comma or pipe delimited list to the `dimensions=` API parameter.

But what if we don't know them? (i.e. the chart has been created with pattern based matches from statsd metrics).

This PR converts the `dimensions=` API parameter to support netdata simple patterns. So, we can say:

```
dimensions=!dim1,!dim2,*
```

The above returns all dimensions except `dim1` and `dim2`.

The match is against both dimension IDs and dimension names, which means that should now know all the dimension names. So 2 more API options have been added `match-ids` and `match-names`. By default both are enabled (to emulate the previous behavior). But if we set them, only the ones set are used.

So:

```
options=match-ids&dimensions=!dim1,!dim2,*
```

Will return all the dimensions that their IDs do not match `dim1` and `dim2`.

---

alarm badges can now be copied to clipboard:

screenshot from 2017-12-24 00-25-14

the 3 action buttons are (left to right):

1. jump to chart
2. copy badge URL to clipboard
3. copy to clipboard an `<embed>` html fragment for an auto-refreshing badge that can added on any web page.

---

added a global dashboard array variable, to allow custom netdata dashboards show alarms for one or more roles (so, there can be custom dashboards, with web alarm notifications, for specific roles).


ktsaou:
Images for the wiki page: https://github.com/firehol/netdata/wiki/Custom-Dashboard-with-Confluence

screenshot from 2017-12-25 00-46-20

copy-embed-badge

screenshot from 2017-12-25 00-55-18

screenshot from 2017-12-25 00-56-58

screenshot from 2017-12-25 01-06-10

screenshot from 2017-12-25 01-07-10

screenshot from 2017-12-25 01-13-15

screenshot from 2017-12-25 01-14-09

screenshot from 2017-12-25 01-25-05

screenshot from 2017-12-25 01-26-04

screenshot from 2017-12-25 01-30-14

screenshot from 2017-12-25 01-45-39

final-confluence

final-confluence

PR #3180, opened 2017-12-25
support option --dont-start-it on static 64 bit installer; 

ktsaou:
fixes https://github.com/firehol/binary-packages/issues/3


ktsaou:
updated the installation wiki page too.

PR #3184, opened 2017-12-27
fix deadlock on exit of static netdata build

ktsaou:
For some reason, musl-libc deadlocks when netdata exits. netdata calls `exit()` but musl-libc waits forever.

```
#0  __wait () at src/thread/__wait.c:14
#1  0x00007f871ce90873 in __lockfile () at src/stdio/__lockfile.c:10
#2  0x00007f871ce92620 in close_file () at src/stdio/__stdio_exit.c:11
#3  0x00007f871ce9266b in __stdio_exit_needed () at src/stdio/__stdio_exit.c:19
#4  0x00007f871cdec8bb in exit () at src/exit/exit.c:32
#5  0x00007f871ce0a27b in netdata_cleanup_and_exit ()
#6  0x00007f871ce0dfd4 in main ()
```

I traced the problem and found that musl-libc does not like the fact that netdata detaches all threads and the main process exits while all the threads are still running.

I tried to call `pthread_cancel` on all the threads running. The problem however is that if this is called on an exited thread, it crashes with SIGSEGV. So, the exit of threads has to be synchronized somehow.

This PR, attempts to synchronize all threads, by implementing:

1. All threads become non-cancelable while they hold a lock.
2. All cleanup is now local to each thread - even for threads that spawn other threads all the cleanup logic is closer to the thread that needs it.
3. To exit netdata, the main thread cancels the threads it spawned.

So the exit procedure is now: a) the main thread cancels the threads it spawned. These threads cancel the threads they spawned, etc, until all are cancelled. All threads that hold a lock (any lock) defer their cancellation until all the locks they hold are released. Once everything is stopped, netdata exits.

Fixes: https://github.com/firehol/binary-packages/issues/4

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,7 @@ set(NETDATA_LINUX_FILES
         src/sys_kernel_mm_ksm.c
         src/sys_fs_cgroup.c
         src/sys_fs_btrfs.c
-        )
+        src/threads.c src/threads.h)
 
 set(NETDATA_COMMON_FILES
         src/adaptive_resortable_list.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,8 @@ set(NETDATA_LINUX_FILES
         src/proc_uptime.c
         src/sys_kernel_mm_ksm.c
         src/sys_fs_cgroup.c
-        src/sys_fs_btrfs.c)
+        src/sys_fs_btrfs.c
+        )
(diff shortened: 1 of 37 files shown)


ktsaou:
ok. This is stable.

With this PR, netdata has abstracted threads and locks. This means that we are getting closer to provide a true cross-platform abstraction layer around netdata that will allow netdata to be ported to more platforms.

PR #3201, opened 2018-01-02
Include the current netdata version in the HTTP 'Server' header.

mahlonsmith:
Super simplistic patch to include the current netdata version within HTTP contexts.


ktsaou:
merged! thank you!

PR #3206, opened 2018-01-03
allmetrics?format=json now uses metrics filtering like the rest of the backends

### LINKED ISSUE #3198: Chart filtering for allmetrics endpoint (json and shell formats)
Hey Costa!

I have a couple of enhancement requests for the `[backend]` section of the netdata configuration.

Currently, the "send charts matching" attribute nicely filters collected data, but it only does so for the prometheus format.  How difficult would it be to allow that filtering to function for all formats supported by the *allmetrics* API endpoint?

Secondarily (or perhaps, alternatively):  That endpoint allows overriding server configuration of specific attributes (names, prefix).  It would be exceptionally nice to expose the `send charts matching` behavior as a parameter to the allmetrics endpoint as well.

I realize I can fetch directly at the `chart` endpoint, but that only works when I know the chart name in advance -- there's quite a few that are dynamic, that necessitate the allmetrics fetching -- only to throw away all the rest of the data that I'm not interested in -- roughly 1MB of JSON, * N servers, * X requests/min... it adds up really fast.

Something like:

    http://example.org:19999/api/v1/allmetrics?format=json&filter=disk_space.*%20system.load

Thanks for any feedback!

ktsaou:
; fixes #3198

### DIFF
--- a/src/rrd2json.c
+++ b/src/rrd2json.c
@@ -306,7 +306,7 @@ void rrd_stats_api_v1_charts_allmetrics_json(RRDHOST *host, BUFFER *wb) {
     // for each chart
     RRDSET *st;
     rrdset_foreach_read(st, host) {
-        if(rrdset_is_available_for_viewers(st)) {
+        if(backends_can_send_rrdset(BACKEND_SOURCE_DATA_AVERAGE, st)) {
             rrdset_rdlock(st);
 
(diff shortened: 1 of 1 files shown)


ktsaou:
Closing this PR: Check https://github.com/firehol/netdata/issues/3198#issuecomment-354933888 for the reasons.

PR #3217, opened 2018-01-04
make web/version.txt more readable

Wing924:
AS IS: `2cff2b7b780d92929f2fd40d0c1549c46e5c3ae8`
TO BE:` v1.9.0-113-g2cff2b7`

### DIFF
--- a/web/Makefile.am
+++ b/web/Makefile.am
@@ -119,7 +119,7 @@ dist_webdnt_DATA = \
 
 version.txt:
 	if test -d "$(top_srcdir)/.git"; then \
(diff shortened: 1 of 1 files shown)


ktsaou:
This will break the update function of the dashboard.
The hash is used there.

Why do you need this change?

Wing924:
I currently use `curl http://<myserver>:19999/api/v1/charts | jq .version` to get netdata version for service discovery but its output is too long. I think it's better to have some simple endpoints to get that. 
If web/version.txt can't be changed, something like `/api/v1/version` is also good for me :)

PR #3224, opened 2018-01-04
Fix icons not being consistent size

xPaw:
This was initially done in 9941e2f5829426f2fa9bbb9db61f9f59511673b5 but broke when icons changed to SVG version.


ktsaou:
merged!
thanks!

PR #3226, opened 2018-01-04
cgroup-network search for PIDs in sub-cgroups; 

### LINKED ISSUE #3199: Why does cgroup stats for LXC containers show stats for some containers, but not others?
I'm running LXD 2.20 on Ubuntu Xenial and Netdata is pleasantly aware of my containers. However, not all containers have the same stats available! Why?

They always have CPU and RAM stats available, while IO Read/Write, Network Read/Write and Disk space utilization is hit and miss. In the sense that on a 30 container system, almost all of them have one or more charts/stats missing.

Does anybody have an explanation for this behavior? I haven't kept a close eye on them, so I cannot rule out this is usage-based somehow (i.e. one thought I had was that maybe those stats didn't show if there was simply no data to show)

ktsaou:
also it works even when there are no veth interfaces at the system (for VMs, `vnetXX` interfaces)

fixes #3199


ktsaou:
netdata-overview

PR #3230, opened 2018-01-05
fix init script in order to stop netdata ASAP

Wing924:
improve firehol/binary-packages#4

### DIFF
--- a/system/netdata-init-d.in
+++ b/system/netdata-init-d.in
@@ -21,6 +21,68 @@ STOP_TIMEOUT="60"
 
 LOCKFILE=/var/lock/subsys/$DAEMON
 
+# Stop netdata processes
+# Usage: kill_netdata [-p pidfile] [ -d delay] {program}
+kill_netdata()
+{
+    local RC base pid pid_file= delay
+    RC=0; delay=3
+    if [ "$1" = "-p" ]; then
+	pid_file=$2
+	shift 2
+    fi
+    if [ "$1" = "-d" ]; then
+	delay=$2
+	shift 2
+    fi
+
+    # Save basename.
+    base=${1##*/}
+
+    # Find pid.
+    __pids_var_run "$1" "$pid_file"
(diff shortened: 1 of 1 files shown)


Wing924:
@ktsaou Can you take a look at this PR?

ktsaou:
Sorry for delay.

Well, it seems good.

However, this init file is used by many systems. Check this: https://github.com/firehol/netdata/blob/91db73ef803fc8bbdba891227cc126dff8d9b672/installer/functions.sh#L479-L486

Do you believe this is good for all of them?

It is relatively "safe" to merge it, since this file will be installed only on new installations (the installer does not overwrite init scripts, to allow the admins customize them), but generally, some testing needs to be done.

What do you think?

Wing924:
I only tested on RHEL 6 and CentOS 6. I don't have other OS. Can you test it on other OSs?
We need to test carefully on these 2 lines: ` __pids_var_run "$1" "$pid_file"` and `pid="$(__pids_pidof "$1")"`, since it seems private functions in /etc/init.d/functions. I'm not sure other OS have these functions.

ktsaou:
Same here. I don't have access to all other systems.

Can we make it more portable?

The installer has a portable `pidof` (works on all Linux including busybox, freebsd and macos): https://github.com/firehol/netdata/blob/c3fb56027a41ba71f604733bec2d08ac30058ea0/installer/functions.sh#L144-L161

The location of the pid file should be known by the init script.

Keep in mind the installer uses the following to kill all running netdata (this is careful enough when running at a host, not to kill a netdata running in a container): https://github.com/firehol/netdata/blob/c3fb56027a41ba71f604733bec2d08ac30058ea0/installer/functions.sh#L565-L638

I think however, this is probably overkill, since the pid file should be known. The only additional check would be to call `pidisnetdata PID` to make sure the pid is really a netdata and not a stale pid currently used by another process.

PR #3242, opened 2018-01-07
added MemAvailable from /proc/meminfo; 

### LINKED ISSUE #3241: Question: (MemTotal - Memory Used) vs. Memory Available.
Referring to the previous Q&A in issue #1034 ("Help required to match data to of system.ram to free -m"):

Netdata provides system RAM metrics of Free, Used, Cached and Buffers.
Per closed issue #1034, Used is calculated as:

unsigned long long Used = MemTotal - MemFree - Cached - Buffers;

I have been using this "Used" metric as a percentage of MemTotal (calc'd through Netdata provided metrics as the sum of free, used, cached and buffers) to track how close the system is to being out of memory (OOM) when it approaches and hits 100% Used.

However, while (MemTotal - Used) does seem to be close to the OS provided metric of MemAvailable most of the time based on observation, they do report different results.

From the kernel commit comment at [https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773), MemAvailable does a calculation that's a little more involved that what (MemTotal - Used) amounts to, and that calculation may change in the future.

For those of us who would like to track available system memory closely and accurately, wouldn't it be good to provide the OS calculated MemAvailable metric as well since that seems to differ by a least a small extent to the estimate of (MemTotal - Used)?

Or is there a better way to look at this than I am doing?  Or maybe I'm missing something?

Thanks!

John

ktsaou:
fixes #3241 

Added a new chart called `mem.available`, that shows the kernel estimated available RAM for userspace allocations, without causing swapping.

screenshot from 2018-01-07 03-38-30

And an alarm:

screenshot from 2018-01-07 03-39-05

### DIFF
--- a/conf.d/health.d/ram.conf
+++ b/conf.d/health.d/ram.conf
@@ -27,11 +27,11 @@
       on: mem.available
       os: linux
    hosts: *
-    calc: ($avail + $used_ram_to_ignore) * 100 / ($system.ram.used - $used_ram_to_ignore + $system.ram.cached + $system.ram.free)
+    calc: ($avail + $used_ram_to_ignore) * 100 / ($system.ram.used + $system.ram.cached + $system.ram.free + $system.ram.buffers)
    units: %
    every: 10s
     warn: $this < (($status >= $WARNING)  ? ( 5) : (10))
     crit: $this < (($status == $CRITICAL) ? (10) : ( 5))
    delay: down 15m multiplier 1.5 max 1h
-    info: estimated amount of RAM available for processes
+    info: estimated amount of RAM available for userspace processes, without causing swapping
       to: sysadmin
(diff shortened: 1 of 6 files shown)


l2isbad:
@ktsaou 
have a look
screenshot_20180109_151704

Memory charts prio.
https://github.com/firehol/netdata/pull/3242/files#diff-cee94df25e4d04fdebd61d28825701baL171

I think it should be right after CPUs section.

ktsaou:
thanks! making a PR now

PR #3259, opened 2018-01-10
revert by default to multi-threaded web server ; the static ones brea…

ktsaou:
…ks streaming


ktsaou:
The issues has been fixed with #3260

PR #3262, opened 2018-01-11
Fixed bug with Alerta notification. I forgot to add /alert method to …

kattunga:
…webhook url.

Better explication about API URL


ktsaou:
merged

PR #3266, opened 2018-01-11
handle incoming traffic on UDP sockets

ktsaou:
ok, this is embarrassing.

PR #3248 broke also statsd UDP data reception.
This PR fixes it.


cis2131:
Thank you.. It works :-)

PR #3270, opened 2018-01-11
strip environment on cgroup-network;

### LINKED ISSUE #3269: privilege escalation via cgroup-network plugin
In my RPM installation of version 1.9.0, there is a setuid binary, "cgroup-network",  which provides at least two ways for any local user to gain root access:

- Shell command injection here: https://github.com/firehol/netdata/blob/9625c5f599e5c7a1bb58b387443690679d3febc8/src/cgroup-network.c#L445

- Unsafe PATH here: https://github.com/firehol/netdata/blob/46603d84e8cff18cb80bcaceb90abed82686ab89/plugins.d/cgroup-network-helper.sh#L25

ktsaou:
fixes #3269

### DIFF
--- a/plugins.d/cgroup-network-helper.sh
+++ b/plugins.d/cgroup-network-helper.sh
@@ -22,7 +22,10 @@
 
 # -----------------------------------------------------------------------------
 
-export PATH="${PATH}:/sbin:/usr/sbin:/usr/local/sbin"
+# the system path is cleared by cgroup-network
+export PATH="/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
+[ -x /etc/profile ] && source /etc/profile
+
 export LC_ALL=C
 
 PROGRAM_NAME="$(basename "${0}")"
@@ -72,7 +75,6 @@ debug() {
 
 [ -z "${NETDATA_PLUGINS_DIR}"  ] && NETDATA_PLUGINS_DIR="$(dirname "${0}")"
 [ -z "${NETDATA_CONFIG_DIR}"   ] && NETDATA_CONFIG_DIR="$(dirname "${0}")/../../../../etc/netdata"
-[ -z "${NETDATA_CACHE_DIR}"    ] && NETDATA_CACHE_DIR="$(dirname "${0}")/../../../../var/cache/netdata"
 
 # -----------------------------------------------------------------------------
 # parse the arguments
--- a/src/cgroup-network.c
+++ b/src/cgroup-network.c
@@ -589,17 +589,16 @@ int main(int argc, char **argv) {
         call_the_helper(pid, NULL);
     }
     else if(!strcmp(argv[1], "--cgroup")) {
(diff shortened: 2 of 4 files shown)


src/cgroup-network.c
```diff
@@ -472,8 +480,36 @@ void call_the_helper(const char *me, pid_t pid, const char *cgroup) {
...
+}
 
-    freez(m);
+int verify_path(const char *path) {
+    struct stat sb;
+
+    char c;
+    const char *s = path;
+    while((c = *s++)) {
+        if(c == '$' || c == '`') {
```
norcnorc: If the goal is to prevent subshells in shell commands, it might be necessary to forbid '<' and '>' also: Bash supports `<(command)` and `>(command)`.

src/popen.c
```diff
@@ -113,6 +113,42 @@ FILE *mypopen(const char *command, volatile pid_t *pidptr)
...
+    for(i = (int) (sysconf(_SC_OPEN_MAX) - 1); i > 0; i--)
+        if(i != STDIN_FILENO && i != STDERR_FILENO && i != pipefd[PIPE_WRITE]) close(i);
+
+    // move the pipe to stdout
+    if(pipefd[PIPE_WRITE] != STDOUT_FILENO) {
+        dup2(pipefd[PIPE_WRITE], STDOUT_FILENO);
+        close(pipefd[PIPE_WRITE]);
+    }
+
+    execle("/bin/sh", "sh", "-c", command, NULL, env);
```
norcnorc: The risk of command injection could be reduced by eliminating the "/bin/sh" here: It's a whole extra shell process that's not doing any useful work except splitting the `command` apart into the pieces from which we've just assembled it. If the helper script path and its arguments were passed in as separate strings (or an array of them, for `execvple`), then no shell ever has to parse them (assuming all uses of them in the helper script itself are properly quoted).

plugins.d/cgroup-network-helper.sh
```diff
@@ -22,7 +22,10 @@
 
 # -----------------------------------------------------------------------------
 
-export PATH="${PATH}:/sbin:/usr/sbin:/usr/local/sbin"
+# the system path is cleared by cgroup-network
+export PATH="/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
```
norcnorc: The `#!/usr/bin/env bash` at the top of the file seems to be getting a default `PATH` from somewhere (or it wouldn't be able to find `bash`!) so maybe this line isn't needed after all?

ktsaou: I think is the default shell (`/bin/sh`) used by `cgroup-network` that does that. Before that line above, the PATH is `/usr/local/bin:/usr/bin` (at least on my fedora 27 - I added `set >/tmp/log` in the script to find it out).

ktsaou:
Another solution would be to just walk through the environment at `cgroup-network` and replace all illegal characters in all available variables. This means we could still allow the the variables to pass-through, but no injection would be possible.

PR #3275, opened 2018-01-12
Alerta notifier.

kattunga:
family value was missing in event, this may cause automatic correlation to fail with some alarms.


ktsaou:
merged. thanks!

PR #3285, opened 2018-01-14
unit scaling improvements; 

### LINKED ISSUE #3176: Add bytes and milliseconds to scalable units
Those are standard units to measure.. (statsd timers are in milliseconds)

ktsaou:
fixes #3176 

This PR adds more units scaling:

- `MB` to `B`
- `GB` to `KB`
- `GB` to `B`
- several conversions above `TB` and up to `ZB`

In general, there is no meaning to scale more. netdata supports numbers with 7 fractional digits, so scaling more than 10mil times is meaningless.

This PR also adds convertible `milliseconds`. ms is tricky to scale (from a presentation perspective). So, this PR does the following:

- when `ms` is below 1000, it shows `MS.x`, like `1.5` or `345.0` - so the max visible resolution is `0.1` ms

screenshot from 2018-01-15 00-59-27

- when `ms` is above 1000 and below 60000 (1 min), it shows `S.ms` like `5.12` or `34.17` or `345.02`, so the max visible resolution is `10` ms

screenshot from 2018-01-15 01-00-09

- when `ms` is above 60000, it shows `M:SS.ms`, like `6:32.14`, again the max visible resolution is `10` ms

screenshot from 2018-01-15 01-00-34


ktsaou:
edited above the description 0.01 secs  = 10 ms, not 100 ms.

PR #3288, opened 2018-01-15
system.ram plugin: treat Slab memory as Cached

amichelic:
Hi,
the system.ram plugin seems (on Linux) to treat slab memory as "used" while the "free" command shows it as "cache".
system ram-high_slab
This is what free shows:
```
# free -h
              total        used        free      shared  buff/cache   available
Mem:           125G         45G        1.4G        868M         78G         29G
Swap:           63G        6.6G         57G
```

The man page of free says that cache is calculated from Cached and Slab in /proc/meminfo, netdata only uses the Cached field.

For your reference, this is the content of /proc/meminfo at (approx.) the time I captured the screenshot above:
[meminfo.txt](https://github.com/firehol/netdata/files/1631199/meminfo.txt)

The memory isn't actually "used" - after running "echo 2 > /proc/sys/vm/drop_caches" the red bar on netdata gets much smaller. I think netdata should include Slab as well for cached memory. This pull request contains my suggestions for the change.

Thanks,
Adalbert

### DIFF
--- a/src/proc_meminfo.c
+++ b/src/proc_meminfo.c
@@ -137,7 +137,8 @@ int do_proc_meminfo(int update_every, usec_t dt) {
     // --------------------------------------------------------------------
 
     // http://stackoverflow.com/questions/3019748/how-to-reliably-measure-available-memory-in-linux
-    unsigned long long MemUsed = MemTotal - MemFree - Cached - Buffers;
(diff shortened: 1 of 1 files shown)


ktsaou:
merged!
thank you for fixing this!

amichelic:
Looks nice, thank you!

PR #3297, opened 2018-01-16
use mlock() and msync() to lock netdata parts into RAM

ktsaou:
This PR is an attempt to improve memory mode `map`. It uses `mlock()` to lock the currently used part of the database into RAM, which `msync()` every a few minutes.

This is how it works:

1. netdata tries to distribute all `mlock()` and `msync()` calls through time. By default it uses `pagesize / sizeof(storage_number)` which on x86/x64 systems leads to 1024 seconds. So, netdata evenly distributes the `mlock()` and `msync()` call through out this time.

   I experimented with 60 seconds and 600 seconds too. These led to increased I/O due to the frequency of the `msync()` calls. There is a setting that controls this `[global].sync to disk every = 1024`.

2. each chart, together with all its dimensions, is synced just once on every such period.

3. The `mlock()` and `msync()` calls are following the area of the RRD files, netdata works on. So, on an empty database, the first area is locked, then as data are written the previous sections are unlocked and synced and new sections get locked. This is done at page boundaries.

**However, all these seem do not make any difference.** The opposite actually: due to `msync()` the system is experiencing more I/O.

I have added a few more charts to monitor this locking and syncing. They are at the netdata section, `mlock` sub-section:

screenshot from 2018-01-16 22-09-09

The last chart above, shows that netdata syncs at most 500KB per second.

However the system still flushes up to 100MB:

screenshot from 2018-01-16 22-09-33

I has been said that `msync()` is actually faulty in Linux: https://lwn.net/Articles/502612/
However recent kernels seem to have patches that make it work properly: https://github.com/torvalds/linux/commit/7fc34a62ca4434a79c68e23e70ed26111b7a4cf8

I any one understands what is the problem, please post.

Currently this is not mergeable. It is useless.

### DIFF
--- a/src/global_statistics.c
+++ b/src/global_statistics.c
@@ -8,7 +8,16 @@ volatile struct global_statistics global_statistics = {
         .bytes_sent = 0,
         .content_size = 0,
         .compressed_content_size = 0,
-        .web_client_count = 1
+        .web_client_count = 1,
+
+        .memory_locked_current_count = 0,
+        .memory_locked_current_bytes = 0,
+        .memory_locked_count = 0,
+        .memory_locked_bytes = 0,
+        .memory_unlocked_count = 0,
+        .memory_unlocked_bytes = 0,
+        .memory_synced_count = 0,
+        .memory_synced_bytes = 0
 };
 
 netdata_mutex_t global_statistics_mutex = NETDATA_MUTEX_INITIALIZER;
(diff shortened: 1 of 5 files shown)


Ferroin:
The `msync()` man page on my system doesn't say that it won't sync more than the requested region, just that the requested region will be synced.  Given what I know about how writeback works on Linux, I would actually expect more than the requested region to be synced _if_ the areas immediately surrounding it are also dirty.  I don't think that's what's going on here, but maybe that will help someone else piece together what's going on...

ktsaou:
I wouldn't mind if it wrote more, as long as the automatic kernel writback is postponed.

I have 2 problems:

1. Although the memory has an `mlock()`, the kernel does not care and pauses randomly netdata to sync it back to disk, which results to gaps at the charts. This is magnified, because the kernel does this in batches (it is not spread over time - all the netdata dirty pages are flushed together - so only an SSD can do it on time).

2. Although I `msync(MS_SYNC)` pages ahead of time, evenly over time, hopping that the kernel will also spread its syncing, the kernel does not care and still flushes all the pages together, even those that have been flushed just seconds ago.

So, unless we find a solution, working with `mmap` for bigger data sets is a actually a no go...

ktsaou:
I think however, this PR should be merged for the `mlock()` part. This part ensures that critical netdata memory parts will not be swapped out when memory is not `map`.

So, disable the `msync()` part and keep the `mlock()` part.
What do you think?

Ferroin:
I do think the `mlock()` part should go in, possibly with an option to control exactly how much gets locked (for example, on my systems where I run with 6 or 12 hour histories, I would love to be able to lock the last hour's worth of data so that displaying charts doesn't need to fault anything in).

As far as the background writeback, I'm not sure there's any way to completely prevent that on Linux, short of opening the file with `O_DIRECT` and (possibly) using AIO, but that will completely bypass any buffering done by the kernel as well, so netdata would need to buffer things itself and dispatch them in batches.  Perhaps people working with really big data sets might benefit from a new memory mode that works mostly like save but writes things out every few pages in a `memory mode = mmap` compatible format to a file opened with `O_DIRECT`?

ktsaou:
> I do think the mlock() part should go in, possibly with an option to control exactly how much gets locked (for example, on my systems where I run with 6 or 12 hour histories, I would love to be able to lock the last hour's worth of data so that displaying charts doesn't need to fault anything in).

Well, this PR already locks the minimum. So, for each metric only the currently working pages are locked. It is already optimal. The memory locked is irrelevant to the size of the database. It only depends on the number of charts and metrics.

There is a chart that visualizes the locked memory. This is from a server with 2.200 metrics and 275 charts (it locked 10MB):

image

Ferroin:
The amount locked isn't relevant to netdata's data collection, but it can when people are looking at the charts.  I'm not 100% certain how many pages are needed to display the default graphed time period for each metric, but if you're locking less than that, there is a possibility that trying to view the dashboard could trigger swap activity (as the data gets pulled in from swap), or possibly even kill netdata if memory pressure is really high.

Being able to extend the amount of historical data that gets kept locked would additionally help those of us who tend to prefer to look further back than the default make sure we don't have to hit swap to get data.

cakrit:
We won't be pursuing this route any more, the new dbengine will make memory modes map and save obsolete.

PR #3312, opened 2018-01-19
Nginx plus

l2isbad:
~~WIP~~ (not a `nginx plus` user)

fixes https://github.com/firehol/netdata/issues/3049

[snapshot](https://github.com/firehol/netdata/files/1645530/netdata-workPc-20180119-153357-480.gz)

@kmhagan

### DIFF
--- a/conf.d/Makefile.am
+++ b/conf.d/Makefile.am
@@ -50,6 +50,7 @@ dist_pythonconfig_DATA = \
     python.d/mongodb.conf \
     python.d/mysql.conf \
     python.d/nginx.conf \
+    python.d/nginx_plus.conf \
     python.d/nsd.conf \
     python.d/ovpn_status_log.conf \
     python.d/phpfpm.conf \
--- a/conf.d/Makefile.am
+++ b/conf.d/Makefile.am
@@ -98,6 +98,7 @@ dist_healthconfig_DATA = \
     health.d/net.conf \
     health.d/netfilter.conf \
     health.d/nginx.conf \
+    health.d/nginx_plus.conf \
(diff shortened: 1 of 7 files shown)


ktsaou:
this is nice!
If I recall correctly, only the streaming charts are missing. Right?

Also, I see auto-detection is disabled. Will it interfere with the plane `nginx` plugin?

l2isbad:
> If I recall correctly, only the streaming charts are missing. Right?

https://github.com/firehol/netdata/issues/3049#issuecomment-349435486

> Also, I see auto-detection is disabled. Will it interfere with the plane nginx plugin?

No, but i don't know what default url should be. [http://127.0.0.1:8080/status](https://www.nginx.com/products/nginx/live-activity-monitoring/)?

ktsaou:
> No, but i don't know what default url should be. http://127.0.0.1:8080/status?

It is not the standard http/https port? So nginx+ opens a new port for statistics?

l2isbad:
you are right

l2isbad:
@ktsaou we can merge it or wait @kmhagan. Maybe he has some comments as a `nginx plus` user.

l2isbad:
@kmhagan current implementation doesn't support adding upstreams/upstream peers/servers zones in runtime (`netdata` restart required). Since it's initial version i think it's ok and we can add the support later 😄

ktsaou:
So, merge?

l2isbad:
yes

ktsaou:
merged! perfect!

ktsaou:
infographic logo: 
image

Issue #3267, opened 2018-01-11
enable system alarms on freebsd


ktsaou:
netdata supports many system alarms. However these alarms are only applied to linux by default.

it would be nice to verify the alarms that work on freebsd too.

This is a simple job:

Each alarm has this little line: https://github.com/firehol/netdata/blob/092f260d7ea1256d2802dc49ed6f71004a2029af/conf.d/health.d/disks.conf#L14

If the alarm works on freebsd, it should be changed to:

```
   os: linux freebsd
```

of course, to do so, we need to verify the alarms do work on freebsd. This is also simple: install netdata on freebsd, add `freebsd` on all those `os` lines and check the alarms modal.

There may be cases that the chart dimensions the alarm needs, are different under freebsd. In this case we should copy and paste the alarm and adapt it a bit, to work on freebsd.

> I won't do this work. If anyone is willing to spare some time on it, please do so. If you need help, just post here.

Ferroin:
I can't exactly double check things (while I have a FreeBSD test system, I don't have the time or patience to set up tests for all the existing alarms), but I've got a pretty good idea of what alarms are likely to work as-is on FreeBSD.  In particular, I would expect the following pre-bundled alarms to just work:

* Apache
* Netdata backend checks
* Beanstalkd
* BIND rndc
* CouchDB
* Elasticsearch
* HAProxy
* ISC DHCPD
* LightHTTPD
* Memcached
* MongoDB
* MySQL
* named
* NGinx
* PostgreSQL
* Redis
* Squid
* Varnish
* web_log

I would expect FPing, IPFS and Retroshare to work if they have working FreeBSD ports (I'm not sure if they do or not).

I would expect BTRFS and mdstat to never work (I doubt the underlying software they care about will ever get a FreeBSD port).  Similar with netfilter, QoS, and the softnet stuff, as netfilter and softnet interrupts are really Linux things and the QoS data collection is rather system specific (it might be possible to write up a pf module for netdata though, and similar logic could then be used for these).

I would expect ZFS to probably work fine if it's collecting data through the normal ZFS API's (I think it is, but I have near zero experience with ZFS outside of using FreeNAS, so i don't know for certain).

As far as the core system metrics, it really just depends on the required metrics being collected.  The CPU usage and wait alarms should work fine out of the box, as should the RAM usage alarms and most of the disk alarms.

ktsaou:
The following alarms are currently only for Linux:

```sh
# cd /etc/netdata/health.d
[...]

Issue #3253, opened 2018-01-09
Alarm badge link escaping for disk paths in default dashboard


Dennis4b:
(On FreeBSD, but doesn't matter I think).

The alarm badges in the default dashboard for disks are rendered with the following code:
`var badge_url = NETDATA.alarms.server + '/api/v1/badge.svg?chart=' + alarm.chart + '&alarm=' + alarm.name + '&refresh=auto';`

yielding for example the following URL for the disk space of a root partition:
`/app/api/v1/badge.svg?chart=disk_space./&alarm=disk_fill_rate&refresh=auto`

The problem is the `/` (forward slash) in the path, causing the badge to not work.

The actual metric, which is used in charts, is named `disk_space._` (i.e. path separators are replaced with underscores), so I assume there is an escaping step missing somewhere, but I'm not familiar enough with the code to speculate what the correct place to fix this is.

Note also that the "jump to chart" link with action `scrollToChartAfterHidingModal` does work because the chart_id is flattened with `NETDATA.name2id` so the jump goes to the `div` with `id ` `chart_disk_space__` either way.

ktsaou:
Hi,

you are right. this is a bug.

however the bug is that the HTTP request parser should not consider that `/` as a path. It should have stopped parsing `/` as path at the `?`.

Charanjet:
Can anyone please help me to create new alarms and new charts i m looking for a complete guidance.

ktsaou:
traced it to find the issue.

The bug is at the internal web server of netdata. It parses the URLs like this:

1. a request comes

2. it checks it received the whole of it (it searches for `\r\n\r\n` - i.e. an empty line as defined in the spec)

3. BUG 1: it decodes the whole URL. This is a bug since any `/` or `?` URL encoded will be decoded before actually parsing the URL. To work properly it should decode the URL part by part (i.e. first split it in its parts and then decode each part of it). **The way it is now, URL encoding cannot be used to encode characters that are URL control characters (`/`, `?`, `&`, `%`).** Even when URL encoded, these character will still play an important role in URL parsing (although they are escaped).

4. BUG 2: it does not separate URL `path` from URL `query string`. It treats the whole URL as a continuous string, separated by `/?&`, which of course leads to bugs like the one we face here.

So, the right flow should be like this:

1. receive the request

2. make sure it received the whole of it

3. split the path from the query string

4. URL decode the path to find the API command or the filename to be served

5. split the query string in its parts (separated with `&`)

[...]

PR #3337, opened 2018-01-25
Enable compatible FreeBSD alarms

arch273:
I went through #3267 , enabling, adapting, and verifying the 'Linux-only' alarms.

A few comments:
disks.conf - Appears to work, but a known bug with the badges is already reported.  I enabled this for now.  See #3253 
ipc.conf - FreeBSD doesn't have ipc.semaphores.max or semaphores.arrays.max, making this unusable.
softnet.conf - I think qdrops is a roughly equivalent choice.
swap.conf - I don't know of a good way to calculate ram_in_swap for FreeBSD.  Maybe there's a reasonable way to do this?
tcp_conn.conf - No metric for tcp_max_connections nor am I sure how this is calculated on Linux
tcp_listen.conf  - This does work, but the FreeBSD package doesn't include the default as an example as it does for every other value.  I used this to test:
[plugin:freebsd:net.inet.tcp.stats]
        TCP listen issues = yes

I didn't see any immediately obvious way to enable these for FreeBSD:
entropy.conf - Fortuna (random) only ever blocks during system boot AFAIK, so probably not needed for FreeBSD?  See https://www.freebsd.org/cgi/man.cgi?query=random&sektion=4
tcp_mem.conf, tcp_orphans.conf, qos.conf - No metrics for these.

Test evidence from the alarms modal:
https://docs.google.com/document/d/e/2PACX-1vRh5MYFoMBe678-FTf77sZIoZI-hOfdbfLmVb6flMu85BXziAL9yV1qi9ROZM-Udjm28NbSZW5T6i46/pub

### DIFF
--- a/conf.d/health.d/cpu.conf
+++ b/conf.d/health.d/cpu.conf
@@ -39,3 +39,17 @@ template: 20min_steal_cpu
    delay: down 1h multiplier 1.5 max 2h
     info: average CPU steal time for the last 20 minutes
       to: sysadmin
+
+## FreeBSD
+template: 10min_cpu_usage
+      on: system.cpu
+      os: freebsd
+   hosts: *
+  lookup: average -10m unaligned of user,system,interrupt
+   units: %
+   every: 1m
+    warn: $this > (($status >= $WARNING)  ? (75) : (85))
+    crit: $this > (($status == $CRITICAL) ? (85) : (95))
+   delay: down 15m multiplier 1.5 max 1h
+    info: average cpu utilization for the last 10 minutes (excluding nice)
+      to: sysadmin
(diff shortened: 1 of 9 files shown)


Ferroin:
FWIW, the entropy tracking isn't technically needed on Linux either, it's just there because there are a very large number of people who have essentially zero understanding of how `/dev/random` and `/dev/urandom` work on Linux and as a result think they need to track it.

ktsaou:
Wow! nice!
I'll review this within the week end!

Thank you!

ktsaou:
This is very nice!
I am merging it.
I think, the suggestions you made have to be discussed with @vlvkobal.

PR #3344, opened 2018-01-25
d3pie chart library

ktsaou:
This PR adds interactive pie / donut charts to netdata.

I have optimized a bit [d3pie](https://github.com/benkeen/d3pie) for speed (actually, removed most transitions). Other than that, it is pretty much the same with the original.

d3pie


ktsaou:
now yet used by the default dashboard

PR #3350, opened 2018-01-27
added API parameter gpoints=N to allow emulating different data collection periods

ktsaou:
The netdata API already supports a parameter called `points`. This sets the number of points that must be returned from a `data` API call.

So, if the query asks for 1 hour of data (3600 seconds), setting `points=60` will reduce the 3600 data points to 60 points, using the calculation specified at `method=` (`average`, `sum`, etc). So, with `method=sum` this call will return 60 points, each summing 60 points of the database, ie. it will return `X / min`.

The above works well, until the chart is zoomed by hand. If the chart is zoomed-in to let's say 1200 seconds, netdata will now return again `points=60`, but this time aggregating only 20 database points, which completely destroys the per-minute expectation.

To fix it, this PR adds another parameter `gpoints=N`. This works only when `method=average`. Specifying `gpoints=60` will force netdata to always aggregate in multiples of 60 data collection points. So, no matter how the chart is zoomed, the result will always be per-minute.

Keep in mind `gpoints=A` and `points=B` can work together. So, with `method=average` and one hour of data, we can set `gpoints=60` and `points=10`, to have netdata return 10 points, each with the per-minute average of 60 data collection points (i.e. for each returned points netdata will sum 360 database points and divide them by 6 to get the per-minute average).

Of course, setting `gpoints` to anything above 1, will force the chart to be aligned.

Using `gpoints` we can now switch chart units. Of course the units label is not changed automatically, but the data will adapt.

netdata will apply this to all types of metrics (`absolute`, `incremental`, etc), so this option should used only when there is meaning to do so.

### DIFF
--- a/src/health.c
+++ b/src/health.c
@@ -428,6 +428,7 @@ void *health_main(void *ptr) {
                                                   , rc->after
                                                   , rc->before
                                                   , rc->group
+                                                  , 0
                                                   , rc->options
                                                   , &rc->db_after
                                                   , &rc->db_before
--- a/src/rrd2json.c
+++ b/src/rrd2json.c
@@ -1413,7 +1413,7 @@ static RRDR *rrdr_create(RRDSET *st, long n)
     return r;
 }
 
-RRDR *rrd2rrdr(RRDSET *st, long points, long long after, long long before, int group_method, int aligned)
+RRDR *rrd2rrdr(RRDSET *st, long points, long long after, long long before, int group_method, long group_points, int aligned)
 {
     int debug = rrdset_flag_check(st, RRDSET_FLAG_DEBUG)?1:0;
     int absolute_period_requested = -1;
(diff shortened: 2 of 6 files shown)


ktsaou:
btw, with the addition of PR #3344 (d3pie library) it would be nice to have pie/donut charts for e.g. web server response codes. However, presenting a pie chart with per-second resolution is problematic. For busy servers it would probably be ok, but for servers that are not that busy, the chart will most probably be empty most of the time.

This PR allows adding such pie or gauge charts with different units. So, the main chart will be per-second, but the pie chart will present the same info per minute.

ktsaou:
hm... I think this PR is wrong.

It should be `gtime=seconds`, so that the queries will be fixed, even if data collection frequency is not 1.

ktsaou:
replaced with #3351

PR #3360, opened 2018-01-28
Add new dimension in fronius plugin: solar_consumption

ccremer:
rel_SelfConsumption indicates the consumption in % when using grid AND solar power income in regard to Load.

The new dimension considers solar power as the only source, which is more interesting to people when doing statistics.

`solar_consumption` is `0%` if there is no solar power (night or snowfall). It is `100%` if the load is greater than solar power. If the panels produce double the load, the consumption is `50%`.


ktsaou:
nice!

PR #3363, opened 2018-01-28
global XSS protection for netdata

ktsaou:
netdata suffers from XSS in these cases:

1. loading snapshots
2. accessing remote servers via the `;server=URL` URL hash parameter
3. registry data

This PR adds a global XSS protection for netdata.

I have received an email from a folk, explaining this in detail to me. I hope he will join this discussion.

Also, XSS cannot be solved completely, since netdata actually relies on it in certain cases, this PR makes sure netdata has basic protection against it.

So, here is what it does:

## snapshots

When snapshots are loaded, the data of the snapshot are checked for XSS and properly escaped before hitting the screen. The snapshot however includes also javascript functions that control various aspects of the dashboard. I have added a note at the snapshot modal to state the fact:

image

## registry

The registry requests are now always checked for XSS.

## remote servers (via `;server=URL` at the URL hash)

The dashboard checks all API calls for XSS. When loading such a dashboard, this modal appears:

screenshot from 2018-01-30 00-30-05

### DIFF
--- a/src/health_json.c
+++ b/src/health_json.c
@@ -2,8 +2,12 @@
 #include "common.h"
 
 static inline void health_string2json(BUFFER *wb, const char *prefix, const char *label, const char *value, const char *suffix) {
-    if(value && *value)
-        buffer_sprintf(wb, "%s\"%s\":\"%s\"%s", prefix, label, value, suffix);
+    if(value && *value) {
+        buffer_sprintf(wb, "%s\"%s\":\"", prefix, label);
+        buffer_strcat_htmlescape(wb, value);
+        buffer_strcat(wb, "\"");
+        buffer_strcat(wb, suffix);
+    }
     else
         buffer_sprintf(wb, "%s\"%s\":null%s", prefix, label, suffix);
 }
--- a/src/health_json.c
+++ b/src/health_json.c
@@ -31,7 +31,6 @@ static inline void health_alarm_entry2json_nolock(BUFFER *wb, ALARM_ENTRY *ae, R
                     "\t\t\"exec_code\": %d,\n"
                     "\t\t\"source\": \"%s\",\n"
                     "\t\t\"units\": \"%s\",\n"
-                    "\t\t\"info\": \"%s\",\n"
                     "\t\t\"when\": %lu,\n"
                     "\t\t\"duration\": %lu,\n"
                     "\t\t\"non_clear_duration\": %lu,\n"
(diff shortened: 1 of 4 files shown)


Ferroin:
So, this is just my thoughts, but I would make checking on `/api/v1/data` optional.  Most people probably won't care much, but some are likely to care very much (after all, that does give out potentially sensitive information about the system).

ktsaou:
ok, did it. Check the updated description on github.

Ferroin:
I think the modal is a great way to approach this.  The description you give in the modal is also really good (it conveys the important information in a way that should be easy even for non-technical users to understand).

ktsaou:
With this addition, the only possible XSS should be by modifying the `dashboard_info.js` (and its custom version).

Unfortunately, netdata relies on javascript code downloaded from remote netdata servers or included in the snapshots to render all the elements of the dashboard (i.e. the text above the charts, the gauge charts at section headings, etc).

Ferroin:
Documenting that clearly somewhere would be good, but I can't think of any way around it myself.

ktsaou:
ok, let me check - I think I have a solution:

1. On snapshots, the javascript core is not yet run when the modal is shown - so the user has the option to accept the risk and load the snapshot or not.

   I think I can add a tick here, to enable/disable running code from the snapshot. When the code from the code is not run, we could still run the beautification code from the server the dashboard is served.

2. On remote servers, the javascript code has already run when the modal is shown. **This is a problem**, since the user is not given the choice.

   A similar solution could be applied here. Move the modal before the remote code is run and give the user the option to run it or not. Again, when the remote beautification code is not run, we could run the corresponding code from the server the dashboard is served.

I think we have a solution...

ktsaou:
fixed the typos... typing too fast...

ktsaou:
ok, this is it.

I hate ticks, so I didn't add any.

- When loading snapshots, either accept the risk, or don't load it.

- When loading remote servers, disabling xss protection loads remote javascript code too, while enabling xss protection does not run any remote javascript code.

So, I think this can be merged.

ktsaou:
btw, the modal has changed. Check the description above for a screenshot.

Ferroin:
+1 on the new modal (not quite as concise, but gives a much better description).

The only other thing I can think of is whether or not we'll remember the decision per-server to disable XSS checking (with the caveat that I haven't been able to check if the updated code does so or not).

ktsaou:
> The only other thing I can think of is whether or not we'll remember the decision per-server to disable XSS checking (with the caveat that I haven't been able to check if the updated code does so or not).

Well, I don't think this is needed. The modal will appear only when the hash URL parameter `;server=` is added and this can only happen by hand.
Generally, this feature is needed only when you want to view a newer dashboard on an older netdata. If the versions are in sync, you can access the server directly.

ktsaou:
merged! we have XSS protection now!

PR #3365, opened 2018-01-29
Fixes a bug in recent fronius plugin change

ccremer:
One could say I didn't test it enough ;) But now it works as it should.


ktsaou:
merged. thanks!

PR #3387, opened 2018-02-04
d3pie optimizations

ktsaou:
d3pie used `d3.select()` and `d3.selectAll()` too much. This made it quite slow when the page has a lot of elements.

This PR caches several selections and when this is not easily possible, it uses `pie.svg.select()` and `pie.svg.selectAll()` to constrain the searches within the chart.

d3pie should now be quite usable on large pages, although there is still room for improvement...


ktsaou:
Updated the developer of d3pie: https://github.com/benkeen/d3pie/issues/151

PR #3398, opened 2018-02-07
Added support of opsgenie.com alerting

Mitrofanov:


ktsaou:
thanks! nice work!
Is it tested? I don't have comments, so if it is tested, we can merge it.

Mitrofanov:
Hi, yes it's already works in Prod. One thing, it's not supports multiple recipients as soon as there are no such functionality in OpsGenie.

ktsaou:
> it's not supports multiple recipients as soon as there are no such functionality in OpsGenie.

hm... should it them send multiple notifications? I mean, if opsgenie cannot accept a list of recipients, then we should send multiple notifications, one for each recipient. Right?

Mitrofanov:
So, Opsgenie can accept recipients, but we don't use this functional in our setup. So if i had a time, i can add such functionality here. Let me know if this is necessary. Thanks

ktsaou:
hm... I think the way it is now, it silently looses recipients. I would appreciate it if you could adapt it to support multiple recipients. All the notification methods of netdata currently do, so please do so.

cakrit:
Closing due to inactivity. Please comment if you wish to continue with this and we'll reopen it.

tinyhammers:
I know this is old, but I'd like this if it's still on the table. Might give it a test and add to my config anyway.
Hope you're well and staying safe.

cakrit:
@tinyhammers do you think are able to extend alarm-notify.sh yourself? If not, perhaps just create a new feature request and link it to this PR.

PR #3400, opened 2018-02-07
Add charts for postgres

anayrat:
Hello,

This PR add several charts for PostgreSQL: 

  * [x] : Rename chart BGWRITER to CHECKPOINTER because it is checkpointer process which triggers checkpoints (https://www.postgresql.org/docs/current/static/wal-configuration.html).
  * [x] : Rename "writer_scheduled" to "checkpoint_scheduled", same for writer_requested.
  * [x] : New BGWRITER chart according to https://www.postgresql.org/docs/10/static/monitoring-stats.html#PG-STAT-BGWRITER-VIEW
  * [x] :  New chart for autovacuum, it counts workers and their activity VACUUM, VACUUM ANALYZE, ANALYZE, VACUUM FREEZE
  * [x] : Replication delta. Maybe we should put chart below, I did not find how to change the position (python beginner inside)?
  * [x] : Wal chart renamed to archive wal
  * [x] : New chart wal files : number of WAL files, current number of written WAL, current number of recycled WAL
  * [x] : Add temps files stats per database

Heavily inspired by https://github.com/OPMDG/check_pgactivity

Here the result:
# BGWRITER

selection_020

# Autovacuum

selection_021

# Replication delta

selection_022

# Wal

selection_023

# Temp files

selection_024

Regards,

### DIFF
--- a/python.d/postgres.chart.py
+++ b/python.d/postgres.chart.py
@@ -32,6 +32,8 @@ METRICS = dict(
               'tup_updated',
               'tup_deleted',
               'conflicts',
+              'temp_files',
+              'temp_bytes',
               'size'],
     BACKENDS=['backends_active',
               'backends_idle'],
@@ -39,11 +41,20 @@ METRICS = dict(
                  'index_size'],
     TABLE_STATS=['table_size',
                  'table_count'],
+    WAL=['written_wal',
+         'recycled_wal',
+         'total_wal'],
     ARCHIVE=['ready_count',
              'done_count',
              'file_count'],
-    BGWRITER=['writer_scheduled',
-              'writer_requested'],
+    BGWRITER=['checkpoint_scheduled',
+              'checkpoint_requested',
+              'buffers_checkpoint',
+              'buffers_clean',
+              'maxwritten_clean',
+              'buffers_backend',
+              'buffers_alloc',
+              'buffers_backend_fsync'],
     LOCKS=['ExclusiveLock',
            'RowShareLock',
            'SIReadLock',
@@ -52,17 +63,41 @@ METRICS = dict(
            'AccessShareLock',
            'ShareRowExclusiveLock',
            'ShareLock',
-           'RowExclusiveLock']
+           'RowExclusiveLock'],
+    AUTOVACUUM=['analyze',
+                'vacuum_analyze',
+                'vacuum',
+                'vacuum_freeze',
+                'brin_summarize'],
+    STANDBY_DELTA=['sent_delta',
+                       'write_delta',
+                       'flush_delta',
+                       'replay_delta']
 )
 
 QUERIES = dict(
+    WAL="""
+SELECT
+  count(*) as total_wal,
+  count(*) FILTER (WHERE type = 'recycled') AS recycled_wal,
+  count(*) FILTER (WHERE type = 'written') AS written_wal
+FROM
+  (SELECT wal.name,
+     pg_{0}file_name(pg_current_{0}_{1}()),
+     CASE WHEN wal.name > pg_{0}file_name(pg_current_{0}_{1}()) THEN 'recycled'
+       ELSE 'written'
+     END AS type
+  FROM pg_catalog.pg_ls_dir('pg_{0}') AS wal(name)
+  WHERE name ~ '^[0-9A-F]{{24}}$'
+  ORDER BY (pg_stat_file('pg_{0}/'||name)).modification, wal.name DESC) sub;
+""",
     ARCHIVE="""
 SELECT
     CAST(COUNT(*) AS INT) AS file_count,
     CAST(COALESCE(SUM(CAST(archive_file ~ $r$\.ready$$r$ as INT)), 0) AS INT) AS ready_count,
     CAST(COALESCE(SUM(CAST(archive_file ~ $r$\.done$$r$ AS INT)), 0) AS INT) AS done_count
 FROM
-    pg_catalog.pg_ls_dir('{0}/archive_status') AS archive_files (archive_file);
+    pg_catalog.pg_ls_dir('pg_{0}/archive_status') AS archive_files (archive_file);
 """,
     BACKENDS="""
 SELECT
@@ -86,28 +121,37 @@ WHERE relkind = 'i';""",
     DATABASE="""
 SELECT
   datname AS database_name,
-  sum(numbackends) AS connections,
-  sum(xact_commit) AS xact_commit,
-  sum(xact_rollback) AS xact_rollback,
-  sum(blks_read) AS blks_read,
-  sum(blks_hit) AS blks_hit,
-  sum(tup_returned) AS tup_returned,
-  sum(tup_fetched) AS tup_fetched,
-  sum(tup_inserted) AS tup_inserted,
-  sum(tup_updated) AS tup_updated,
-  sum(tup_deleted) AS tup_deleted,
-  sum(conflicts) AS conflicts,
-  pg_database_size(datname) AS size
+  numbackends AS connections,
+  xact_commit AS xact_commit,
+  xact_rollback AS xact_rollback,
+  blks_read AS blks_read,
+  blks_hit AS blks_hit,
+  tup_returned AS tup_returned,
+  tup_fetched AS tup_fetched,
+  tup_inserted AS tup_inserted,
+  tup_updated AS tup_updated,
+  tup_deleted AS tup_deleted,
+  conflicts AS conflicts,
+  pg_database_size(datname) AS size,
+  temp_files AS temp_files,
+  temp_bytes AS temp_bytes
 FROM pg_stat_database
 WHERE datname IN %(databases)s
-GROUP BY datname;
+;
 """,
     BGWRITER="""
 SELECT
-  checkpoints_timed AS writer_scheduled,
-  checkpoints_req AS writer_requested
-FROM pg_stat_bgwriter;""",
-   LOCKS="""
+  checkpoints_timed AS checkpoint_scheduled,
+  checkpoints_req AS checkpoint_requested,
+  buffers_checkpoint * current_setting('block_size')::numeric buffers_checkpoint,
+  buffers_clean * current_setting('block_size')::numeric buffers_clean,
+  maxwritten_clean,
+  buffers_backend * current_setting('block_size')::numeric buffers_backend,
+  buffers_alloc * current_setting('block_size')::numeric buffers_alloc,
+  buffers_backend_fsync
+FROM pg_stat_bgwriter;
+""",
+    LOCKS="""
 SELECT
   pg_database.datname as database_name,
   mode,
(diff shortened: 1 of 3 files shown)


anayrat:
FYI, I plan to send other PR to add charts. Mainly inspired by https://github.com/OPMDG/check_pgactivity

l2isbad:
@facetoe  please have a look

anayrat:
Would you prefer I send one PR per chart or one PR with several charts ?

l2isbad:
> one PR with several charts 

^^

facetoe:
Yep, changing the name makes sense. And adding those additional charts would be awesome.

python.d/postgres.chart.py
```diff
@@ -214,12 +270,42 @@
...
+        'lines': [
+            ['autovacuum_max_worker', 'max workers', 'absolute'],
+            ['analyze', 'analyze', 'absolute'],
+            ['vacuum', 'vacuum', 'absolute'],
+            ['vacuum_analyze', 'vacuum_analyze', 'absolute'],
+            ['vacuum_freeze', 'vacuum_freeze', 'absolute'],
+            ['brin_summarize', 'brin_summarize', 'absolute']
+        ]},
+    'secondary_delta': {
+        'options': [None, 'Secondary delta', 'Kbytes', 'replication delta', 'postgres.secondary_delta', 'line'],
```
l2isbad: >  Netdata can auto-scale these original units: kilobits/s, kilobytes/s, KB/s, KB, MB, and GB

`Kbytes` => `KB`

anayrat:
I think, I have finished. This PR is ready to be reviewed 😄

python.d/postgres.chart.py
```diff
@@ -208,17 +291,52 @@
...
+            ['vacuum_freeze', 'vacuum_freeze', 'absolute'],
+            ['brin_summarize', 'brin_summarize', 'absolute']
+        ]},
+    'secondary_delta': {
+        'options': [None, 'Secondary delta', 'kilobytes/s', 'replication delta', 'postgres.secondary_delta', 'line'],
+        'lines': [
+            ['sent_delta', 'sent_delta', 'absolute', 1, 1024],
+            ['write_delta', 'write_delta', 'absolute', 1, 1024],
+            ['flush_delta', 'flush_delta', 'absolute', 1, 1024],
+            ['replay_delta', 'replay_delta', 'absolute', 1, 1024]
```
l2isbad: if `delta` is needed in dimension i suggest to use space instead of underscore. It looks better on the dashboard.
```pyhon
 ['replay_delta', 'replay delta', 'absolute', 1, 1024]
```
Same true for all new dimensions with underscore.

Another problem: netdata applies interpolation to find the exact value at second-boundaries only for `incremental `values.  `/s` is misleading for `absolute` (there is no guarantee that the module runs once a second).

l2isbad:
gj @anayrat 👍 

in general all the changes looks ok to me, but i am not a postrgresql user.
@facetoe can you look one more time? what do you think?

anayrat:
Thanks @l2isbad . I fixed label and unit.

python.d/postgres.chart.py
```diff
@@ -208,17 +291,52 @@
...
+        'options': [None, 'Autovacuum workers', 'workers', 'autovacuum', 'postgres.autovacuum', 'line'],
         'lines': [
-            ['writer_scheduled', 'scheduled', 'incremental'],
-            ['writer_requested', 'requested', 'incremental']
+            ['autovacuum_max_worker', 'max workers', 'absolute'],
+            ['analyze', 'analyze', 'absolute'],
+            ['vacuum', 'vacuum', 'absolute'],
+            ['vacuum_analyze', 'vacuum_analyze', 'absolute'],
+            ['vacuum_freeze', 'vacuum_freeze', 'absolute'],
+            ['brin_summarize', 'brin_summarize', 'absolute']
```
l2isbad: Are you sure all that stuff should be in one chart?

> autovacuum_max_workers (integer)

`Specifies the maximum number of autovacuum processes (other than the autovacuum launcher) that may be running at any one time. The default is three. This parameter can only be set at server start.`

why do we need to chart it?

anayrat: > Are you sure all that stuff should be in one chart?
Yes, it shows us autovacuum activity. Then, it give information how to tune autovacuum and other settings...

> (autovacuum_max_workers) why do we need to chart it?
You are right, it is not necessary to chart it. Otherwise we should also chart max_connections. I will remove it.

I will also remove "_" in autovacuum labels.

l2isbad:
@anayrat 
keep in mind any chart can have variables
https://github.com/firehol/netdata/wiki/External-Plugins#variable

l2isbad:
just found

> Replication delta. Maybe we should put chart below, I did not find how to change the position (python beginner inside)

@anayrat 
please post the dashboard postgress menu

anayrat:
selection_025

I exported a snapshot : 
[netdata-adrien-pc-20180211-112146-480.zip](https://github.com/firehol/netdata/files/1714250/netdata-adrien-pc-20180211-112146-480.zip)

l2isbad:
> Maybe we should put chart below

what position do you want?

anayrat:
Maybe after all "db *" charts?

l2isbad:
this shoud work
https://github.com/anayrat/netdata/blob/postgres/python.d/postgres.chart.py#L552

change to
```python
position = order.index('database_size')
order.insert(position, chart_name)
```

anayrat:
Thanks @l2isbad, it works as expected!

ktsaou:
this is nice! let me know when it is ready to merge.

l2isbad:
lets wait for @facetoe 
if he is agrees with all changes we can merge it

facetoe:
Looks good to me. Nice work @anayrat!

anayrat:
FYI, I plan to split bgwriter chart in more charts. In order to separate checkpoint, bgwriter, backend and read/write activity.

ktsaou:
nice! so @anayrat merge now or wait for more changes?
since this PR is approved, it is your decision...

anayrat:
please wait for lasts changes (split chart).  Do you want I squash all my commits?

ktsaou:
> please wait for lasts changes (split chart). Do you want I squash all my commits?

ok. yes, it would be nice...

anayrat:
As I said, I splitted bgwritter charts (thanks @gleu). Here is the result : 

selection_027
I splitted charts, because pg_stat_bgwritter reports differrents statistics : 
  * Buffers allocated
  * Buffers written by checkpointer
  * Buffers written by bgwriter
  * Buffers written by backends: there is two metrics:
    * Buffers written by backends, unit is "buffer"
    * Number of times backends have to do their own fsync. It has no unit. I wonder if I should split this line to another chart?
  * Maxwritten : it is the number of time maxwritten threshold is reached, it has no unit.

I assume buffer size is default (8Kb) to compute unit in Kilobytes.

@facetoe thoughts?

If you agree with theses changes, I will squash commits.

Thanks

l2isbad:
@anayrat 
https://github.com/firehol/netdata/pull/3328#issuecomment-359663441
it would be great if you added descriptions to the charts (as an power postgresql user)

anayrat:
@l2isbad good idea! I will do it!

anayrat:
I tried to change `dashboard_info.js` but I have this error: "Cannot load required JS library: http://X.X.X.X:19999/dashboard_info.js?v20180120-1"

Even if I remove all my changes?

Is there a way to clear netdata cache? I tried to clear brower's cache without success :'(

ktsaou:
Open the above URL on your browser and hit control-F5 or shift-F5 to have the browser refresh it.

ktsaou:
Or edit index.html and set the correct date on the URL.

anayrat:
Which date should I put?

anayrat:
Ok, I made a typo in `dashboard_info.js. CI reveals to me :)

anayrat:
It seems finished for me. Here a snapshot. 
[postgres.zip](https://github.com/firehol/netdata/files/1725599/postgres.zip)

And a screenshot : 
netdata-postgres

web/dashboard_info.js
```diff
@@ -984,6 +984,57 @@ netdataDashboard.context = {
...
+        '</ul>' +
+        'For more information see <a href="https://www.postgresql.org/docs/current/static/wal-configuration.html" target="_blank">WAL Configuration</a>.'
+    },
+    'postgres.autovacuum': {
+        info: 'PostgreSQL databases require periodic maintenance known as vacuuming. For many installations, it is sufficient to let vacuuming be performed by the autovacuum daemon.' +
+        'For more information see <a href="https://www.postgresql.org/docs/current/static/routine-vacuuming.html#AUTOVACUUM" target="_blank">The Autovacuum Daemon</a>.'
+    },
+    'postgres.standby_delta': {
+        info: 'Streaming replication delta.<ul>' +
+        '<li><strong>sent_delta:</strong> rReplication delta sent to standby.</li>' +
```
l2isbad: a typo? rReplication

ktsaou:
nice!

 I saw your snapshot and the only thing that bothers me is this:

image

`Temps` should be `Temp`, but I think `files` and `size` should not be on the same chart. These are 2 charts, one counting files and other the size of the files. Generally each chart should include dimension with homogeneous units.

Can you split it in 2 charts?

web/dashboard_info.js
anayrat: I fixed it. Thanks!

anayrat:
@ktsaou I tried to split temp chart, but I don't understand why "Disk blocks" and "Current connection" are between Temp charts?
temp

ktsaou:
> I don't understand why "Disk blocks" and "Current connection" are between Temp charts?

I think the order they are placed in the code, defines that. @l2isbad can you help?

l2isbad:
Yes. Every [chart ](https://github.com/firehol/netdata/wiki/External-Plugins#chart)has priority (relative `priority` of the charts as rendered on the web page, lower numbers make the charts appear before the ones with higher numbers)

Charts are created in [create()](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/FrameworkServices/SimpleService.py#L113-L124) by [self.charts.add_chart()](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L114-L129)

Default `self.charts` priority is 60 000
First chart in self.order has 60 000, second 60 001 and so on.

anayrat:
Yes, but db_stat_temp_files is juste after db_stat_temp_bytes : 
https://github.com/anayrat/netdata/blob/284a734f318b015bf631c1e4f26d63bbb9c5078f/python.d/postgres.chart.py#L216

Maybe the problem is here : 
https://github.com/anayrat/netdata/blob/284a734f318b015bf631c1e4f26d63bbb9c5078f/python.d/postgres.chart.py#L562

l2isbad:
@anayrat 
have a look

i install your branch to debug the problem and...

```
2018-02-15 04:19:06: python.d ERROR: postgres: socket: update() unhandled exception: function pg_current_wal_lsn() does not exist
LINE 3:   pg_wal_lsn_diff(pg_current_wal_lsn() , sent_lsn) AS sent_d...
                          ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.

2018-02-15 04:19:06: python.d DEBUG: postgres: socket: update => [FAILED] (elapsed time: -, retries left: 59)
```

anayrat:
Oh right, view definition has changed in Postgres 10. I should do the same trick as https://github.com/anayrat/netdata/blob/284a734f318b015bf631c1e4f26d63bbb9c5078f/python.d/postgres.chart.py#L439

anayrat:
@l2isbad Can you try again? I made some changes to handle differences between previous versions and Postgres 10.

l2isbad:
```
2018-02-15 19:23:15: python.d ERROR: postgres: socket: update() unhandled exception: function pg_stat_file(text, boolean) does not exist
LINE 14:   ORDER BY (pg_stat_file('pg_xlog/'||name,true)).modificatio...
                     ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

anayrat:
@l2isbad sorry :(. I will try with all Postgres supported version (9.3 -> 10).

anayrat:
@l2isbad I try with these Postgres versions:
* 9.3
* 9.4
* 9.5
* 9.6
* 10

I also changed postgres.conf. Thanks again for your review.

l2isbad:
@anayrat 
not it work 👍 

about the order
1. please add `db_stat_connections` to ORDER
2. change
https://github.com/anayrat/netdata/blob/postgres/python.d/postgres.chart.py#L455
to
```python
for chart_name in [name for name in self.order if name.startswith('db_stat')]:
```
3. actual order is reverse order from
```python
ORDER = ['db_stat_transactions', 'db_stat_blks', 'db_stat_tuple_returned', 'db_stat_tuple_write', 'db_stat_temp_bytes', 'db_stat_temp_files',
```

`db_stat_temp_files` first, db_stat_temp_bytes' second and so on

anayrat:
Good catch for db_stat_connections, it was missing on master! Thanks for the advice for order. It works now.

l2isbad:
are you done @anayrat ?

anayrat:
Yes, I pushed last changes for order and db_stat_connections

l2isbad:
@anayrat thx 😄 

@ktsaou ready

ktsaou:
merged! great work!

PR #3421, opened 2018-02-13
Add python module ntpd using SocketService

rda0:
Modify `SocketService.py` to allow use of `socket.SOCK_DGRAM` using `self.dgram_socket = True`.
Setting this will force to use an UDP socket. An alternative would be to try UDP if a connection to TCP fails.

The module needs to read the raw bytes in the NTP Control Message Protocol header once during `check()`. This does not work with the method `_receive()` in `SocketService`, which decodes the response. Therefore I have implemented the methods `_receive_bytes()` and `_get_raw_data_bytes()` in the module, which are identical to the methods in `SocketService` except for that they return the raw bytes without decoding. It would be possible to integrate this by using a property like `self._raw_bytes`, adding an optional variable to the two methods, or some other way, but I am not sure what would be the "best" way to do this.

I tried to use multiple jobs, to possibly also monitor external hosts, but this does not work correctly at the moment. I don't know why, but when another external host job is added in addition to `local`, it overwrites/interferes with the charts for `local`. I have not found any clues in the examples, that would point to special handling of the charts in conjunction with multiple jobs.

Looking forward for your feedback.

### DIFF
--- a/python.d/ntp.chart.py
+++ b/python.d/ntp.chart.py
@@ -237,6 +237,7 @@ class Service(SocketService):
         self.peers = dict()
         self.request = str()
         self.retries = 0
+        self.all_peers = self.configuration.get('all_peers', False)
 
         peer_filter = self.configuration.get('peer_filter', r'127\..*')
         try:
@@ -276,7 +277,10 @@ class Service(SocketService):
             return None
 
         data.update(self.system.get_data(raw))
-        
+
+        if not self.all_peers:
+            return data
+
         # TODO: run self.find_new_peers() every N runs?
         if not self.peers or self.retries > 8:
             self.find_new_peers()
@@ -289,7 +293,7 @@ class Service(SocketService):
                 else:
                     self.retries += 1
 
-        return data or None
+        return data
 
     def find_new_peers(self):
         new_peers = dict((p.real_name, p) for p in self.get_peers())
--- a/python.d/ntp.chart.py
+++ b/python.d/ntp.chart.py
@@ -1,12 +1,11 @@
 # -*- coding: utf-8 -*-
 # Description: ntp netdata python.d module
 # Author: Sven Mäder (rda0)
+# Author: Ilya Mashchenko (l2isbad)
 
-import socket
 import struct
 import re
 
-from itertools import cycle
 from bases.FrameworkServices.SocketService import SocketService
 
 # default module values
@@ -36,292 +35,137 @@ ORDER = [
     'sys_rootdisp',
     'sys_stratum',
     'sys_tc',
-    'sys_precision'
+    'sys_precision',
+    'peer_offset',
+    'peer_delay',
+    'peer_dispersion',
+    'peer_jitter',
+    'peer_xleave',
+    'peer_rootdelay',
+    'peer_rootdisp',
+    'peer_stratum',
+    'peer_hmode',
+    'peer_pmode',
+    'peer_hpoll',
+    'peer_ppoll',
+    'peer_precision'
 ]
 
 CHARTS = {
     'sys_offset': {
-        'options': [None, "Combined offset of server relative to this host", "ms", 'system', 'ntp.sys_offset', 'area'],
+        'options': [None, 'Combined offset of server relative to this host', 'ms', 'system', 'ntp.sys_offset', 'area'],
         'lines': [
             ['offset', 'offset', 'absolute', 1, PRECISION]
         ]},
     'sys_jitter': {
-        'options': [None, "Combined system jitter and clock jitter", "ms", 'system', 'ntp.sys_jitter', 'line'],
+        'options': [None, 'Combined system jitter and clock jitter', 'ms', 'system', 'ntp.sys_jitter', 'line'],
         'lines': [
             ['sys_jitter', 'system', 'absolute', 1, PRECISION],
             ['clk_jitter', 'clock', 'absolute', 1, PRECISION]
         ]},
     'sys_frequency': {
-        'options': [None, "Frequency offset relative to hardware clock", "ppm", 'system', 'ntp.sys_frequency', 'area'],
+        'options': [None, 'Frequency offset relative to hardware clock', 'ppm', 'system', 'ntp.sys_frequency', 'area'],
         'lines': [
             ['frequency', 'frequency', 'absolute', 1, PRECISION]
         ]},
     'sys_wander': {
-        'options': [None, "Clock frequency wander", "ppm", 'system', 'ntp.sys_wander', 'area'],
+        'options': [None, 'Clock frequency wander', 'ppm', 'system', 'ntp.sys_wander', 'area'],
         'lines': [
             ['clk_wander', 'clock', 'absolute', 1, PRECISION]
         ]},
     'sys_rootdelay': {
-        'options': [None, "Total roundtrip delay to the primary reference clock", "ms", 'system', 'ntp.sys_rootdelay', 'area'],
+        'options': [None, 'Total roundtrip delay to the primary reference clock', 'ms', 'system',
+                    'ntp.sys_rootdelay', 'area'],
         'lines': [
             ['rootdelay', 'delay', 'absolute', 1, PRECISION]
         ]},
     'sys_rootdisp': {
-        'options': [None, "Total root dispersion to the primary reference clock", "ms", 'system', 'ntp.sys_rootdisp', 'area'],
+        'options': [None, 'Total root dispersion to the primary reference clock', 'ms', 'system',
+                    'ntp.sys_rootdisp', 'area'],
         'lines': [
             ['rootdisp', 'dispersion', 'absolute', 1, PRECISION]
         ]},
     'sys_stratum': {
-        'options': [None, "Stratum (1-15)", "1", 'system', 'ntp.sys_stratum', 'line'],
+        'options': [None, 'Stratum (1-15)', '1', 'system', 'ntp.sys_stratum', 'line'],
         'lines': [
             ['stratum', 'stratum', 'absolute', 1, PRECISION]
         ]},
     'sys_tc': {
-        'options': [None, "Time constant and poll exponent (3-17)", "log2 s", 'system', 'ntp.sys_tc', 'line'],
+        'options': [None, 'Time constant and poll exponent (3-17)', 'log2 s', 'system', 'ntp.sys_tc', 'line'],
         'lines': [
             ['tc', 'current', 'absolute', 1, PRECISION],
             ['mintc', 'minimum', 'absolute', 1, PRECISION]
         ]},
     'sys_precision': {
-        'options': [None, "Precision", "log2 s", 'system', 'ntp.sys_precision', 'line'],
+        'options': [None, 'Precision', 'log2 s', 'system', 'ntp.sys_precision', 'line'],
         'lines': [
             ['precision', 'precision', 'absolute', 1, PRECISION]
         ]}
 }
 
-# Dynamic charts templates
-PEER_PREFIX = 'peer'
+PEER_CHARTS = {
+    'peer_offset': {
+        'options': [None, 'Filter offset', 'ms', 'peers', 'ntp.peer_offset', 'line'],
+        'lines': [
+        ]},
+    'peer_delay': {
+        'options': [None, 'Filter delay', 'ms', 'peers', 'ntp.peer_delay', 'line'],
+        'lines': [
+        ]},
+    'peer_dispersion': {
+        'options': [None, 'Filter dispersion', 'ms', 'peers', 'ntp.peer_dispersion', 'line'],
+        'lines': [
+        ]},
+    'peer_jitter': {
+        'options': [None, 'Filter jitter', 'ms', 'peers', 'ntp.peer_jitter', 'line'],
+        'lines': [
+        ]},
+    'peer_xleave': {
+        'options': [None, 'Interleave delay', 'ms', 'peers', 'ntp.peer_xleave', 'line'],
+        'lines': [
+        ]},
+    'peer_rootdelay': {
+        'options': [None, 'Total roundtrip delay to the primary reference clock', 'ms', 'peers',
+                    'ntp.peer_rootdelay', 'line'],
+        'lines': [
+        ]},
+    'peer_rootdisp': {
+        'options': [None, 'Total root dispersion to the primary reference clock', 'ms', 'peers',
+                    'ntp.peer_rootdisp', 'line'],
+        'lines': [
+        ]},
+    'peer_stratum': {
+        'options': [None, 'Stratum (1-15)', '1', 'peers', 'ntp.peer_stratum', 'line'],
+        'lines': [
+        ]},
+    'peer_hmode': {
+        'options': [None, 'Host mode (1-6)', '1', 'peers', 'ntp.peer_hmode', 'line'],
+        'lines': [
+        ]},
+    'peer_pmode': {
+        'options': [None, 'Peer mode (1-5)', '1', 'peers', 'ntp.peer_pmode', 'line'],
+        'lines': [
+        ]},
+    'peer_hpoll': {
+        'options': [None, 'Host poll exponent', 'log2 s', 'peers', 'ntp.peer_hpoll', 'line'],
+        'lines': [
+        ]},
+    'peer_ppoll': {
+        'options': [None, 'Peer poll exponent', 'log2 s', 'peers', 'ntp.peer_ppoll', 'line'],
+        'lines': [
+        ]},
+    'peer_precision': {
+        'options': [None, 'Precision', 'log2 s', 'peers', 'ntp.peer_precision', 'line'],
+        'lines': [
+        ]}
+}
 
-PEER_DIMENSIONS = [
-    ['offset', 'Filter offset', 'ms'],
-    ['delay', 'Filter delay', 'ms'],
-    ['dispersion', 'Filter dispersion', 'ms'],
-    ['jitter', 'Filter jitter', 'ms'],
-    ['xleave', 'Interleave delay', 'ms'],
-    ['rootdelay', 'Total roundtrip delay to the primary reference clock', 'ms'],
-    ['rootdisp', 'Total root dispersion to the primary reference clock', 'ms'],
-    ['stratum', 'Stratum (1-15)', '1'],
-    ['hmode', 'Host mode (1-6)', '1'],
-    ['pmode', 'Peer mode (1-5)', '1'],
-    ['hpoll', 'Host poll exponent', 'log2 s'],
-    ['ppoll', 'Peer poll exponent', 'log2 s'],
-    ['precision', 'Precision', 'log2 s']
-]
 
+class Base:
+    regex = re.compile(r'([a-z_]+)=((?:-)?[0-9]+(?:\.[0-9]+)?)')
 
-class Peer(object):
-    """
-    Class to hold peer data required in _get_data
-    """
-    def __init__(self, peer_id, name, request):
-        self.id = peer_id
-        self.name = name
-        self.request = request
-
-
-class Service(SocketService):
-    def __init__(self, configuration=None, name=None):
-        SocketService.__init__(self, configuration=configuration, name=name)
-        self.port = 'ntp'
-        self.dgram_socket = True
-        self.request_systemvars = None
-        self.peers = None
-        self.peer_error = 0
-        self.regex_srcadr = re.compile(r'srcadr=([A-Za-z0-9.-]+)')
-        self.regex_data = re.compile(r'([a-z_]+)=([0-9-]+(?:\.[0-9]+)?)(?=,)')
-        self.order = None
-        self.definitions = None
-
-    def create_charts(self):
-        """
-        Creates the charts dynamically.
-        Checks ntp for available peers.
-        Adds all peers whith valid data.
-        """
-        # Create systemvars charts
-        self.order = list(ORDER)
-        self.definitions = dict(CHARTS)
-
-        # Get peer ids
-        self.request = self.get_header(0, 'readstat')
-        peer_ids = self.get_peer_ids(self._get_raw_data(raw=True))
-
-        # Get peers
-        peers = self.get_peers(peer_ids)
-
-        # Create peer charts
-        if peers:
-            charts = dict()
-
-            for dimension in PEER_DIMENSIONS:
-                chart_id = '_'.join([PEER_PREFIX, dimension[0]])
-                context = '.'.join(['ntp', chart_id])
-                title = dimension[1]
-                units = dimension[2]
-                lines = list()
-
-                for peer in peers:
-                    unique_dimension_id = '_'.join([peer.name, dimension[0]])
-                    line = [unique_dimension_id, peer.name, 'absolute', 1, PRECISION]
-                    lines.append(line)
-                charts[chart_id] = dict()
-                charts[chart_id]['options'] = [None, title, units, 'peers', context, 'line']
-                charts[chart_id]['lines'] = lines
-
-            self.order += ['_'.join([PEER_PREFIX, d[0]]) for d in PEER_DIMENSIONS]
-            self.definitions.update(charts)
-            self.peers = cycle(peers)
-        else:
-            self.peers = None
-
-    def get_peers(self, peer_ids):
-        """
-        Figures out the possible local domain name.
-        Queries each peer once to get data for charts.
-        Replace the peer srcadr with the possible hostname.
-        Returns all peers whith valid data.
-        """
-        if peer_ids:
-            peer_ids.sort()
-
-        domain = None
-
-        # Get the local domain name
-        if self.peer_names:
-            try:
-                hostname = socket.gethostname()
-                fqdn = socket.getfqdn()
-                if fqdn.startswith(hostname):
-                    domain = fqdn[len(hostname):]
-            except socket.error:
-                self.error('Error getting local domain')
-
-        peers = list()
-
-        # Get peer data
-        for peer_id in peer_ids:
-            request = self.get_header(peer_id, 'readvar')
-            self.request = request
-            raw = self._get_raw_data()
-            if not raw:
-                continue
-
-            data = self.get_data_from_raw(raw)
-            if not data:
-                continue
-
-            match_srcadr = self.regex_srcadr.search(raw)
-            if not match_srcadr:
-                continue
-
-            name = match_srcadr.group(1)
-            match_peer_filter = self.regex_peer_filter.search(name)
-            if match_peer_filter:
-                continue
-
-            if domain:
-                try:
-                    name = socket.gethostbyaddr(name)[0]
-
-                    match_peer_filter = self.regex_peer_filter.search(name)
-                    if match_peer_filter:
-                        continue
-
-                    if len(name) > len(domain) and name.endswith(domain):
-                        name = name[:-len(domain)]
-                except (IndexError, socket.error):
-                    self.error('Failed to reverse lookup address')
-
-            name = name.replace('.', '-')
-
-            peers.append(Peer(peer_id, name, request))
-
-        return peers
-
-    def check(self):
-        """
-        Checks if we can get valid systemvars.
-        If not, returns None to disable module.
-        """
-        self._parse_config()
-        self.peer_names = self.configuration.get('peer_names', True)
-        peer_filter_custom = self.configuration.get('peer_filter', r'127\..*')
-
-        try:
-            self.regex_peer_filter = re.compile(r'^((0\.0\.0\.0)|({0}))$'.format(peer_filter_custom))
-        except re.error as error:
-            self.error('Pattern compile error: {0}'.format(error))
-            return None
-
-        self.request_systemvars = self.get_header(0, 'readvar')
-        self.request = self.request_systemvars
-        raw_systemvars = self._get_raw_data()
-
-        if not self.get_data_from_raw(raw_systemvars):
-            return None
-
-        self.create_charts()
-
-        return True
-
-    def _get_data(self):
-        """
-        Gets systemvars data on each update.
-        Gets peervars data for only one peer on each update.
-        Total amount of _get_raw_data invocations per update = 2
-        """
-        data = dict()
-
-        self.request = self.request_systemvars
-        raw_systemvars = self._get_raw_data()
-        data.update(self.get_data_from_raw(raw_systemvars))
-
-        if self.peers:
-            peer = next(self.peers)
-            self.request = peer.request
-            raw_peervars = self._get_raw_data()
-            data.update(self.get_data_from_raw(raw_peervars, peer))
-
-        if not data:
-            self.error("No data received")
-            return None
-
-        return data
-
-    def get_data_from_raw(self, raw, peer=None):
-        """
-        Extracts key=value pairs with float/integer from ntp response packet data.
-        """
-        data = dict()
-        try:
-            data_list = self.regex_data.findall(raw)
-
-            for data_point in data_list:
-                key, value = data_point
-                if peer:
-                    dimension = '_'.join([peer.name, key])
-                else:
-                    dimension = key
-                data[dimension] = int(float(value) * PRECISION)
-        except (ValueError, AttributeError, TypeError):
-            self.error("Invalid data received")
-            return None
-
-        # If peer returns no valid data, probably due to ntpd restart,
-        # then wait 10 seconds and re-initialize the peers and charts
-        if not data and peer:
-            self.error('Peer error: No data received')
-            self.peer_error += 1
-
-            if (self.peer_error * self.update_every) > 10:
-                self.error('Peer error count exceeded, re-creating charts.')
-                self.create_charts()
-                self.peer_error = 0
-                self.create()
-
-        return data
-
-    def get_header(self, associd=0, operation='readvar'):
+    @staticmethod
+    def get_header(associd=0, operation='readvar'):
         """
         Constructs the NTP Control Message header:
          0                   1                   2                   3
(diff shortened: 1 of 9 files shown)


l2isbad:
> optional variable to the two methods

this^^

something like 

```python
def _get_raw_data(self, raw=False):
...
...
    data = self._receive(raw)


def _receive(self, raw=False):
...
...
    data += buf.decode('utf-8', 'ignore') if not raw else buf
```

python.d/ntp.chart.py
```diff
@@ -0,0 +1,434 @@
+# -*- coding: utf-8 -*-
+# Description: ntp netdata python.d module
+# Author: Sven Mäder (rda0)
+
+import socket
+import struct
+import re
+
+from itertools import cycle
+from base import SocketService
```
l2isbad:
```python
```
[base.py](https://github.com/firehol/netdata/blob/master/python.d/python_modules/base.py) will be deleted in the future

python.d/ntp.chart.py
```diff
@@ -0,0 +1,434 @@
...
+
+    def create_charts(self):
+        """
+        Creates the charts dynamically.
+        Checks ntp for available peers.
+        Adds all peers whith valid data.
+        """
+        # Create systemvars charts
+        self.order = ORDER
+        self.definitions = CHARTS
```
l2isbad:
```python
self.order = list(ORDER)
self.definitions = dict(CHARTS)
```

python.d/ntp.chart.py
```diff
@@ -0,0 +1,434 @@
...
+            peer_filter = r'^((0\.0\.0\.0)|(' + str(self.configuration['peer_filter']) + r'))$'
+        except (KeyError, TypeError):
+            self.error('error parsing peer_filter')
+            peer_filter = r'^((0\.0\.0\.0)|(127\..*))$'
+        self.regex_peer_filter = re.compile(peer_filter)
+
+        try:
+            self.peer_names = bool((self.configuration['peer_names']))
+        except (KeyError, TypeError):
+            self.peer_names = True
```
l2isbad:
```python
self.peer_names = bool((self.configuration.get('peer_names', True)))  # default is True
```

python.d/ntp.chart.py
```diff
@@ -0,0 +1,434 @@
...
+        self.peer_error = 0
+        self.regex_srcadr = re.compile(r'srcadr=([A-Za-z0-9.-]+)')
+        self.regex_data = re.compile(r'([a-z_]+)=([0-9-]+(?:\.[0-9]+)?)(?=,)')
+        self.order = None
+        self.definitions = None
+
+        try:
+            peer_filter = r'^((0\.0\.0\.0)|(' + str(self.configuration['peer_filter']) + r'))$'
+        except (KeyError, TypeError):
+            self.error('error parsing peer_filter')
```
l2isbad: use dict.get here. try/except not needed

then

```python
try:
    regex = re.compile(pattern)
except re.error as error:
    self.error('Pattern compile error. Using defaults.')
    regex = re.compile(default_pattern)
```

python.d/ntp.chart.py
rda0: Awesome, that fixed the job problem right? I would never have seen that..

python.d/ntp.chart.py
```diff
@@ -0,0 +1,381 @@
...
+        peer_filter_start = r'^((0\.0\.0\.0)|('
+        peer_filter_end = r'))$'
+        peer_filter_default = r'127\..*'
+        peer_filter_custom = str(self.configuration.get('peer_filter', peer_filter_default))
+
+        try:
+            self.regex_peer_filter = re.compile(peer_filter_start + peer_filter_custom + peer_filter_end)
+        except re.error as error:
+            self.error('Pattern compile error: %s, Using defaults.' % str(error))
+            self.regex_peer_filter = re.compile(peer_filter_start + peer_filter_default + peer_filter_end)
```
l2isbad: actually i think we should not use defaults in case of pattern compile error. It should warn user about error and exit (return None). What do you think?
if you agree with me move `peer_filter_` stuff to the `check()`

```python
def check():
    peer_filter_custom = self.configuration.get('peer_filter', r'127\..*')
    try:
        self.regex_peer_filter = re.compile(r'^((0\.0\.0\.0)|({0}))$'.format(peer_filter_custom))
    except re.error as error:
        self.error('Pattern compile error: {0}'.format(error))
        return None
```

python.d/ntp.chart.py
```diff
@@ -0,0 +1,381 @@
...
+
+    def create_charts(self):
+        """
+        Creates the charts dynamically.
+        Checks ntp for available peers.
+        Adds all peers whith valid data.
+        """
+        # Create systemvars charts
+        self.order = list(ORDER)
+        self.definitions = dict(CHARTS)
```
l2isbad: please move it to `init()`

l2isbad:
@rda0 
Does it work after stop/restart `ntp.service`?

python.d/ntp.chart.py
```diff
@@ -0,0 +1,381 @@
...
+        # then wait 10 seconds and re-initialize the peers and charts
+        if not data and peer:
+            self.error('Peer error: No data received')
+            self.peer_error += 1
+
+            if (self.peer_error * self.update_every) > 10:
+                self.error('Peer error count exceeded, re-creating charts.')
+                self.create_charts()
+                self.peer_error = 0
+                self.create()
```
l2isbad: i `self.create()` and `self.create()` here. Not sure this is the right way to re-initialize the peers and charts.

l2isbad: please explain why we need to re-initialize something.

python.d/ntp.chart.py
rda0: Yes, this is better, I'll change it.

python.d/ntp.chart.py
rda0: I am not sure as well. I just tried `self.create()` and it worked. What would be the right way to re-create or modify the charts?

I choose to re-creates all charts in this way to be as simple as possible. In theory only the peer_charts need to be modified/removed.

Reason:

This is needed for the `peer_charts` when `ntpd` is restarted. To get data for a peer, we need to make a request which includes the `peer_id` (or in ntp terms "association_id"). Unfortunately by design, these peer_ids (random integers) change on every `ntpd` restart, even if no configuration change was made. This means after a peer_error (when no data is returned) we need to request again all peer_ids. Then we need to re-create the peer charts, because if one or more pools are configured (or maybe if a configured server address is behind DNS round robin), there will most probably be new peers (with different ip addresses) after a restart, even if no configuration change was made.

rda0: How is it possible to add new dimensions to charts at run-time?

After a restart, we could just request the new peer_ids and add the new dimensions (with possible new IP-addresses/host-names). Any obsolete dimensions should be cleaned up after some time, right?

rda0: And yes, it should work after a restart of `ntpd.service`. The `system_vars` should always work and almost immediately. But the `peer_vars` need a few seconds to be initialized, because `ntpd` first has to poll its peers. If we request the new peer_ids before `ntpd` has properly initialized its peers, there will be no peer_charts shown anymore. This is why I choose to just wait 10s after a peer_error (but maybe there is a better way). We could also periodically get the peer_ids and check if some of the ids have changed.

Ferroin:
Out of curiosity, what NTP implementations has this been tested with?

On Linux you've got at least three implementations you might be running against:

* The reference implementation (used by most of the pool.ntp.org servers, from http://www.ntp.org).
* OpenNTPD (part of the OpenBSD project, found at http://www.openntpd.org/).
* Chrony (an independent implementation that lacks some of the less frequently used NTP modes but adds a lot of local features like support for hardware level timekeeping,, found at https://chrony.tuxfamily.org/).

If it only works with some of them, it might be worth adding a check to have it disable itself if it detects that it won't work.

l2isbad:
@rda0 i tried your module and `peers` part was completely broken.
I [modified your script](https://gist.github.com/l2isbad/8a16b2505d1ca6a9b29f075397a171f6)
(it has all functionality besides `peer_names: yes`)

it can handle adding/removing peers, updating peers ids, updating charts in runtime (`ntpd` restart, stop/start).

Please have a look.
I think we can use it as a base.

rda0:
@Ferroin it is only tested with the reference implementation from [www.ntp.org](http://www.ntp.org).

Looks like NTP Control Messages are not implemented by OpenNTPD and Chrony. It fails (time-out) during `check`, the same happens if targeted at the Chrony's command port `323`:

```python
2018-02-16 11:35:56: python.d ERROR: ntp: ntp: socket to "::1" port 123: failed to receive response: timed out
2018-02-16 11:35:56: python.d ERROR: ntp: ntp: Traceback (most recent call last):
  File "/opt/netdata/usr/libexec/netdata/python.d/python_modules/bases/FrameworkServices/SocketService.py", line 178, in _receive
    buf = self._sock.recv(4096)
timeout: timed out

2018-02-16 11:35:56: python.d DEBUG: ntp: ntp: closing socket
2018-02-16 11:35:56: python.d DEBUG: ntp: ntp: final response: 
2018-02-16 11:35:56: python.d ERROR: ntp: ntp: check() => [FAILED]
```

@Ferroin, @l2isbad should the module name be changed to `ntpd`, to be more specific?

l2isbad:
I think yes

rda0:
@l2isbad thank you for the [modified script](https://gist.github.com/l2isbad/8a16b2505d1ca6a9b29f075397a171f6)

Looks great. I will just add it like this, if that is ok.

About updating all peer_charts on every run: Is that not too expensive, having N socket connections on every run? I will try to make some tests next week on our ntp servers, which have 7 peers and low end hardware.

l2isbad:
> Is that not too expensive
Well, i don't know. Not expensive according `apps.plugin` and module `runtime` chart.

it can be `enabled/disabled` via `conf.file`.

To be honest i don't know are there any reason to chart all peers or we can chart just one peer?

We can chart one peer by default. What do you think?

But for sure it can't be done your way - get data in cycle manner on every run.

Ferroin:
@rda0 I agree with @l2isbad, unless you plan on implementing support for fetching stats from chrony (I don't think there's any reasonable way to do so with openntpd).

Regardless of that, I think some slightly better detection could be done.  I think it might be worth the effort to have an initial check that an NTP daemon is running at all by sending a request as an SNTP client and seeing if that succeeds.  If that works, you know you've got an NTP server to talk to on the given address, and can probably safely assume that a timeout on your first attempt to read stats means it doesn't support fetching stats.

Also, I forgot about ntimed (which probably also does not currently support stats, and has a somewhat odd architecture where stratum 1, lower strata, and client systems all have separate components) and NTPsec (which is ultimately a fork of the reference implementation, and may or may not support stats).

Ferroin:
@l2isbad The only case where it makes sense to chart only one peer is if there is only one peer.  The NTP source selection algorithm kind of depends on having multiple peers to choose from if you don't trust your upstream time servers (and you shouldn't unless you run them yourself and can verify the entire network path from you to them), and as a result it's rather important to know what all of them are doing.  Additionally though, the only sensible way to select a peer to plot is to plot data from whichever one is selected as the current time source,  but that changes over time in most cases, sometimes quite regularly (the two stratum 2 servers I run for pool.ntp.org are both syncing off of 6 stratum 1 servers, and they switch around pseudo-randomly roughly every 15 minutes), and thus you will likely see seemingly random jumps to different values.

On a client node (that is, one which doesn't serve time to any other systems), it might make sense to just not track peers at all (they will usually change hourly anyway in most cases), but on a server, it's kind of important to keep track of them, especially if you're trying to provide time service to a large number of systems.  Given that most people are likely to be running client systems, I think the default should probably be just tracking the local stats, and then let the user opt-in to tracking all the peers too.

l2isbad:
> The only case where it makes sense to chart only one peer is if there is only one peer.

I feel how stupid was my question 😢 

@Ferroin many thanks!

> Given that most people are likely to be running client systems, I think the default should probably be just tracking the local stats, and then let the user opt-in to tracking all the peers too.

[implemented](https://gist.github.com/l2isbad/8a16b2505d1ca6a9b29f075397a171f6#file-ntp-chart-py-L276-L277)
@rda0

rda0:
> default should probably be just tracking the local stats, and then let the user opt-in to tracking all the peers too

@Ferroin, @l2isbad thank you, I wanted to propose exactly that.

>  it might be worth the effort to have an initial check that an NTP daemon is running at all

`ntpd` sends its version number in the response, we could check this:

```
version="ntpd 4.2.8p4@1.3265-o Thu Sep  7 20:43:09 UTC 2017 (1)",
processor="x86_64", system="Linux/4.4.0-112-generic", leap=0, stratum=2,
precision=-23, rootdelay=0.142, ...
```

Ferroin:
@rda0 But do you get a timeout or an actual failure code if the stats commands aren't supported by the implementation?  My point about checking with a single SNTP client query was that it would let you more easily differentiate whether the first collection attempt timing out was from not having a server running, or the server not supporting the required NTP mode.

rda0:
> do you get a timeout or an actual failure code if the stats commands aren't supported by the implementation

@Ferroin it is only a timeout, no response at all (from OpenNTPD and Chrony, haven't tested ntimed yet)

> a single SNTP client query would let you more easily differentiate whether the first collection attempt timing out was from not having a server running, or the server not supporting the required NTP mode

Could you provide an example of such an SNTP query?

Ferroin:
I don't know the protocol well enough myself to give a good example.  What I'm talking about though is querying the server like `ntpdate` would when trying to get the time, but just sending a single query.  

SNTP versus NTP doesn't really matter for this as they're the same on the wire, I'm just in the habit of referring to one-shot queries like this as SNTP because they technically are (the difference is entirely on the nodes themselves, SNTP is locally stateless for clients, and thus simpler to implement).

rda0:
>  My point about checking with a single SNTP client query was that it would let you more easily differentiate whether the first collection attempt timing out was from not having a server running, or the server not supporting the required NTP mode.

@Ferroin why do you think we should do this additional check?

When the connection times out, either something is listening on this port and is not responding or the packet was dropped or lost somewhere on the way. In the first two cases, there is nothing we can do. For the case that the packet was lost we could try it again and possibly get a response. If nothing is listening on the port (no server running), we should get a `connection refused` message. If we get a response, which actually contains data that we can use, we assume it is `ntpd` talking to us and show the charts. We could check the version string in the response, to find out if it really is `ntpd`, but I am not sure if this is required.

From my tests it looks like at least Chrony and OpenNTPD do not implement the Control Message Protocol. I don't think that any other implementation is using this. But I am just speculating here.

For information:

Control Messages were first introduced as optional in NTPv2 ([rfc1119](https://www.rfc-editor.org/rfc/rfc1119.pdf)) and are also in NTPv3 ([rfc1305](https://tools.ietf.org/html/rfc1305)). They were removed from NTPv4 ([rfc5905](https://tools.ietf.org/html/rfc5905)) and are now separately specified as draft [draft-ietf-ntp-mode-6-cmds-03](https://tools.ietf.org/html/draft-ietf-ntp-mode-6-cmds-03).

Ferroin:
@rda0 And now I feel like an idiot...  I"m used to dealing with firewalls that just drop UDP traffic to ports with nothing listening instead of sending back an ICMP message to say the port is closed, but that should never be the case for the loopback interface.

Unless you plan on supporting gathering statistics from remote nodes, the check isn't needed at all, and even for remote nodes it should probably just be optional if it gets implemented.

python.d/ntpd.chart.py
```diff
@@ -0,0 +1,385 @@
...
+        raw = self._get_raw_data()
+        if not raw:
+            return None
+
+        data.update(self.system.get_data(raw))
+
+        if not self.all_peers:
+            return data
+
+        # TODO: run self.find_new_peers() every N runs?
```
rda0: @l2isbad I am not sure if we need to do this. This would be required for the case, when a new peer was added. Maybe @Ferroin can give us some input on this? I have no idea if a new peer can appear without a restart. According to @Ferroin it can happen if a pool is configured. If this always means, that another peer was removed at the same time, then we do not need some extra code, because we would get no data from at least one peer.

rda0:
> Unless you plan on supporting gathering statistics from remote nodes

@Ferroin This is supported. But I don't know how useful it is. Maybe if you don't want to install netdata on the remote host. But I want netdata everywhere, I would install it in my fridge if it would be possible ;)

python.d/ntpd.chart.py
Ferroin: There are a couple of cases where it might matter, namely:

1. If you're using multicast or broadcast client mode (not very common except on big networks with no Windows computers), the total number of peers may change over time because you track everyone who's broadcasting the time.
2. If you're using manycast client mode, the number of peers can also change over time (even less common than broadcast or multicast modes) because you associate with whoever responds.
3. If the server is reconfigured and restarted between collection cycles (for example, moving from one network to another when using DHCP to find NTP servers), the number of peers can change.
4. If you are using a pool of servers like pool.ntp.org, the total number of peers won't change, but who you're talking to as peers may change (and thus you may see new association ID"s added and old ones removed).
5. If the daemon can't resolve a hostname it's configured to use as a server when it goes to do a lookup, that peer gets removed, but it may get re-added if a subsequent lookup succeeds.  Interestingly however, this is not the case for peers who just constantly timeout (they just stay stuck in the .INIT. state).

Ferroin:
@rda0 Even if it's supported, I think it's probably overkill to make the check.  The big thing I was worried about was wasting time trying to poll the local host if there was nothing there, but because the loopback interface won't drop UDP packets to closed ports without sending back an ICMP port unreachable message, it doesn't matter.

rda0:
@Ferroin I am sorry, now I understand what you are worried about. The timeout (I think 5s) we run into, when there is something on the port, which does not respond. I don't know if this is bad..

Ferroin:
The only cases where it should happen though are if there is something listening and just dropping packets, or if there's a paranoid firewall that drops packets to closed ports without sending the appropriate ICMP error back.

python.d/ntpd.chart.py
rda0: Wow, thank you :) I don't understand most of it but I think that means we should at least periodically check for new peer_ids or just always check it on every update (maybe one more socket connection does not make the difference).

rda0:
Off topic:

@Ferroin if you are interested, on the [OpenLDAP mailing list](http://www.openldap.org/lists/openldap-technical/201709/msg00059.html) you can read why I started creating the module. With the netdata graphs I found out, that when the ntp servers are on the same subnet, you will possibly have a poor clock accuracy due to ARP requests, which does not happen if you have routers between client and server.

l2isbad:
> periodically check for new peer_ids

this can be done by adding one more check in
https://github.com/rda0/netdata/blob/python-module-ntp/python.d/ntpd.chart.py#L285

```python
if self.runs_counter % RESCAN_INTERVAL == 0 ......

```

`runs_counter` added in https://github.com/firehol/netdata/pull/3432

python.d/ntpd.chart.py
```diff
@@ -0,0 +1,385 @@
...
+            peers_to_update = set(self.peers) & set(new_peers)
+            peers_to_add = set(new_peers) - set(self.peers)
+
+            for peer_name in peers_to_remove:
+                self.hide_old_peer_from_charts(self.peers[peer_name])
+                del self.peers[peer_name]
+
+            for peer_name in peers_to_update:
+                if new_peers[peer_name].id != self.peers[peer_name].id:
+                    self.peers[peer_name].rebuild_request(new_peers[peer_name].id)
```
l2isbad: looking through the code i found that
`peers_to_update` and `for peer_name in peers_to_update:` not needed.

```python
self.peers.update(new_peers)
```

do the job

(`Peer.rebuild_request` should also be deleted)

python.d/ntpd.chart.py
Ferroin: Case 4 and case 5 are the ones we would need to worry about.  1 and 2 are not likely to be widely used on server systems (they're too prone to interference, and have accuracy issues), and 3 is not something I would personally expect most netdata users to experience with any regularity.

Overall though, yes, I think it probably should just check on every update (or at least every few updates).

rda0:
@l2isbad please review especially the last commit 9e884ca. I modified `get_data()` to also get data from peers after we call `find_new_peers()`. And I moved the `self.peer_*` configuration to and `self.definitions.update(PEER_CHARTS)` to `check()`. Please correct me if this is not good.

>  added in #3432 (wait until it's merged and do rebase)

Do you want me to do this?

```
git checkout master
git fetch upstream
git merge upstream/master
git checkout python-module-ntp
git rebase master
```

python.d/ntpd.chart.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def check(self):
+        """
+        Checks if we can get valid systemvars.
+        If not, returns None to disable module.
+        """
+        self._parse_config()
+
+        if self.show_peers:
+            self.definitions.update(PEER_CHARTS)
```
l2isbad: Not needed. Please move to `init()`.
Just do update without any checks

python.d/ntpd.chart.py
```diff
@@ -0,0 +1,383 @@
...
+                self.error('Compile pattern error (peer_filter) : {0}'.format(error))
+                return None
+
+            try:
+                self.peer_rescan = int(self.configuration.get('peer_rescan', peer_rescan))
+                if self.peer_rescan <= 0:
+                    raise ValueError('int > 0 expected: {0}'.format(self.peer_rescan))
+            except ValueError as error:
+                self.error('Value error (peer_rescan) : {0}'.format(error))
+                return None
```
l2isbad: please move to init()
```python
self.configuration.get('peer_rescan', 60)
```

That is all. All checks are not needed.

python.d/ntpd.chart.py
```diff
@@ -0,0 +1,383 @@
...
+        raw = self._get_raw_data()
+        if not raw:
+            return None
+
+        data.update(self.system.get_data(raw))
+
+        if not self.show_peers:
+            return data
+
+        if self.runs_counter == 1 or self.runs_counter % self.peer_rescan or self.retries > 8:
```
l2isbad: I suggest to use
```python
if not self.peers or self.runs_counter % self.peer_rescan == 0 or self.retries > 8
```
`not self.peers` - first run
`self.runs_counter % self.peer_rescan == 0` - every self.peer_rescan runs
to run it on every update `self.peer_rescan` should be 1

l2isbad:
> Do you want me to do this?

yes

python.d/ntpd.chart.py
rda0: @l2isbad OK. I did this check, because I saw when you put `peer_rescan: 0` in the config it will result in an unhandled exception:

```
python.d ERROR: ntpd: ntpd: update() unhandled exception: integer division or modulo by zero
```

rda0: @l2isbad and leave `peer_filter` check in `check()` or also move it to `init()`?

l2isbad: `update() unhandled exception` is ok :smile: 

by the description of the option it is clear that it's must be `> 1` (you can add comment about it in conf file)

> peer_filter check in check()

well, do as you want. If you feel it should be in `check()` leave it in `check()` :+1:

l2isbad:
@rda0 one more thing
please add new module to makefiles
1. [ntpd.conf](https://github.com/firehol/netdata/blob/master/conf.d/Makefile.am#L28-L70)
2. [ntpd.chart.py](https://github.com/firehol/netdata/blob/master/python.d/Makefile.am#L14-L59)

rda0:
> please add new module to makefiles

@l2isbad done. And thanks a lot for teaching me, I hope the next PR will be a bit better. Also many thanks to @Ferroin.

ktsaou:
@rda0 could you please attach a netdata snapshot? (zip it and you can directly drop it here).

rda0:
[screenshots.zip](https://github.com/firehol/netdata/files/1741686/screenshots.zip)

ktsaou:
thanks!
I meant a snapshot not a screenshot.

This button on the dashboard: 
image

It would be fantastic to add information about each chart. The work you have done is **wow**!

ktsaou:
Here are images you posted (to get github URLs):

ntpd_peers

ntpd_system

rda0:
> I meant a snapshot not a screenshot.

My fault :joy: Thanks for the screenshot of the button.

[snapshots.zip](https://github.com/firehol/netdata/files/1741887/snapshots.zip)

rda0:
> It would be fantastic to add information about each chart

@ktsaou Yes, I agree. I think I won't have time this week, maybe next week I'll add some more information.

ktsaou:
> I think I won't have time this week, maybe next week I'll add some more information.

It is ok. If @l2isbad approves this, we can merge it and add info later.

l2isbad:
gj @rda0  :+1:

ktsaou:
merged it!

l2isbad:
@rda0 and the last but not least

update the [wiki pages](https://github.com/firehol/netdata/wiki/Add-more-charts-to-netdata) (page and sidebar `Python Modules`). There is no rush of course :smile:

ktsaou:
@rda0 I have a few suggestions to make:

## default config

I think this is a good default for most systems:

```
localhost:
  name: 'local'
  host: 'localhost'
  port: '123'
  show_peers: no

localhost_ipv4:
  name: 'local'
  host: '127.0.0.1'
  port: '123'
  show_peers: no

localhost_ipv6:
  name: 'local'
  host: '::1'
  port: '123'
  show_peers: no
```

## units

There are a few peer charts with units `1`:
image

I think the units should be `stratum`, `hmode`, `pmode` depending on the chart.

## stratum 16

I think stratum 16 peers should be excluded:

screenshot from 2018-02-22 00-22-56

```sh
# ntpq -pn 127.0.0.1
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
-194.116.168.41  194.177.210.54   2 u  936 1024  377   13.671    1.804  14.778
*194.177.210.54  .GPS.            1 u  312 1024  377   10.879    0.208   0.205
-193.239.214.226 193.93.167.241   3 u  205 1024  377   17.911    1.792   0.470
-195.167.30.249  87.120.164.97    2 u  602 1024  377   11.825    0.988   0.343
-155.207.113.227 122.231.59.246   2 u  107 1024  377   19.000   -0.035   0.532
 207.171.17.42   .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 174.127.124.192 .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 193.243.171.138 .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 85.234.197.21   .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 198.105.223.32  .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 199.188.48.59   .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 46.234.32.107   .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 173.255.139.202 .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 50.31.240.214   .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 82.94.190.206   .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 37.123.115.71   .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 85.214.25.217   .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 46.29.176.73    .XFAC.          16 u    - 1024    0    0.000    0.000   0.000
 94.242.223.210  17.253.52.125    2 u  20d 1024    0   56.728    1.023   0.000
+62.38.83.25     194.177.210.54   2 u 1015 1024  377   10.780    0.087   1.112
+62.38.83.26     194.177.210.54   2 u  929 1024  377   11.616    0.007   0.986
 127.127.1.0     .LOCL.          10 l  21d   64    0    0.000    0.000   0.000
```

## peers cleanup

For this I am not sure, so I would like to discuss it a bit:

Since peers may be added at any time, I think the plugin will continuously adding peers, without ever cleaning up. Unfortunately, even netdata cannot remove dimensions. It can only cleanup entire charts (by re-sending the `CHART` with the option `obsolete`).

So, it is probably better to provide a family of charts per peer, and mark them as obsolete when they are no longer needed. If we do it like that, netdata will cleanup.

This solution would add a **family** per peer (ie. `peer XXX`), with the following charts (this is an example - I just grouped them by their units):

1. `peer_timings_XXX` with dimensions `offset`, `delay`, `dispersion`, `jitter`, `xleave`, `root_delay`, `root_dispersion`
2. `peer_modes_XXX`, with dimensions `stratum`, `hmode`, `pmode`.
3. `peer_exponents_XXX`, with dimensions `host poll`, `peer poll` and `precision`

Of course, I am not exactly sure for the meanings of the dimensions, so may be a better grouping can be made and instead of 3 charts, we may add 5 or 6 with a different grouping.

I understand we will be changing the presentation because of a technical limitation, which is generally a bad idea. However if we don't do it like that the only other option we have, is to disable peers by default (as I suggested in the configs above), until we update netdata to cleanup individual dimensions too (it is somewhat complex to do, so I don't expect this to be done soon enough).

What do you think?

l2isbad:
> is to disable peers by default

I see default is [don't show peers](https://github.com/firehol/netdata/blob/master/python.d/ntpd.chart.py#L236)

rda0:
@ktsaou 

> I think this is a good default for most systems

Why is this needed? I thought, that `SocketService` tries to connect to all socket combinations returned by `getaddrinfo()`. Is this for the case of no proper name resolution for `localhost`?

> I think the units should be stratum, hmode, pmode depending on the chart.

I will change it.

> I think stratum 16 peers should be excluded

OK, how? By setting the value to `0`? Strange, [ntpq documentation](http://doc.ntp.org/current-stable/ntpq.html) states, stratum should be 1-15.

> peers cleanup

So this does not actually clean up the dimensions, just hide them?

```python
            for peer_name in peers_to_remove:
                self.hide_old_peer_from_charts(self.peers[peer_name])
                del self.peers[peer_name]
```

> solution would add a family per peer

I was also thinking about doing it like that, but I was not sure what is better. I think the grouping you proposed is good.

For the meanings of the peer dimensions I am also not sure what they mean. I am also not sure if they are all really that important and will ever give any useful info at all. This includes `stratum`, `hmode`, `pmode`, `hpoll`, `ppoll` and `precision`. I think these variables are unlikely to change very often. Maybe the `*poll` variables are most likely to ever change for most peers but I don't know if that can be interesting.

l2isbad:
> So this does not actually clean up the dimensions, just hide them?

That is why `hide` in the method name :smile: 
https://github.com/firehol/netdata/issues/3017#issuecomment-347548512

> I was also thinking about doing it like that, but I was not sure what is better. I think the grouping you proposed is good.

how many charts it will produce?

rda0:
When I updated [conf.d/python.d.conf](https://github.com/firehol/netdata/pull/3454/files#diff-c39197a3fbfba20ea2919cd264bff593) in #3454, I saw that the `chrony` module is disabled by default.

I also got feedback from one of my collegues @patschbo, he would prefer the `ntpd` module to be disabled on hosts not intended as "NTP servers". And I think he is right. All of our hosts (workstations and servers) are running with an `ntpd` daemon, which is configured to serve time to anyone. But none of them actually serve as an "NTP server host", except for our two "NTP servers" with GPS clock. The hosts also only fetch the time from the two "NTP servers", not from eachother.

Maybe it would be better to disable the module by default so it can be enabled on demand?

Ferroin:
@rda0 I think it really depends on more than whether or not the system is a time server to determine if the system stats matter enough to track them.  If you're doing cluster computing for example, it can be _very_ bad if your jitter, offset, or clock frequency wander get too high because you need to stay in tight sync with the other cluster members.  I do think disabling by default does make some sense though, as it really doesn't' matter in most cases.

@ktsaou Check my first comment on #3454 for info on what the various peer dimensions mean.  As far as what's important you generally have:

* `peer_offset` Somewhat important, but not critically so.
* `peer_delay` Also somewhat important, but again not critically so.
* `peer_dispersion` Extremely important to the internal workings of peer selection.
* `peer_jitter` Also extremely important to peer selection, and additionally a good indicator of time quality from the peer.
* `peer_xleave` Not sure about this one.
* `peer_rootdelay` Same as `peer_delay`
* `peer_rootdisp` Very important to the peer selection algorithm, and also a good indicator of remote time quality.
* `peer_stratum`  Non-critical, but still useful information (what stratum a remote peer is partially dictates reasonable values for some of the important metrics).
* `peer_hmode` and `peer_pmode` Same as `peer_stratum`
* `peer_hpoll` and `peer_ppoll` Not hugely important unless you're debugging issues with the ntp daemon.
* `peer_precision` Also not sure about this one.

Of these, the `*poll` values will follow a very specific progression in most cases (start at minpoll, increase over fixed intervals to maxpoll), so they probably can be dropped.  The `*mode` values should not change once they are determined, but knowing what type of peer a given peer is is useful in many cases.  The stratum value can change, but is not likely to do so in most cases (a given peer's stratum is one more than the stratum of the peer it is synchronizing from, with higher stratum values generally correlating to lower quality time sources, which is why uninitialized associations have a stratum of 16, and the local reference clock has a stratum of 10 or 12 (though using the local reference clock is a Bad Idea (TM) in a significant majority of cases)).  Changes to the other values should be infrequent (occurring at 2^`peer_hpoll` seconds intervals in most cases), but most of them are useful to track.

ktsaou:
@Ferroin very detailed and accurate as always! Thank you!

So, regarding the questions:

> Why is this needed? I thought, that SocketService tries to connect to all socket combinations returned by getaddrinfo(). Is this for the case of no proper name resolution for localhost?

I have seen all types of misconfigurations. For example, localhost to be resolved to `::1` but ipv6 to be disabled, daemons to bind to `::1` but `localhost` to resolve only to ipv4, etc. I have even seen a system that `ntpq -p` does not work, but `ntpq -p 127.0.0.1` does...

So, as a general guideline (and since python.d supports it in a very nice way), we add all combinations manually. Only one will be used.

> When I updated conf.d/python.d.conf in #3454, I saw that the chrony module is disabled by default.

As a general guide, all netdata plugins that can support autodetection should be enabled by default. There are 2 exceptions to this:

1. the plugin is "heavy", meaning that it requires a lot of CPU or RAM, it forks too much or generally wastes resources.

2. the plugin may be "harmful", meaning that we have good indications that it may under conditions cause application servers to crash or enter infinite loops. So the plugin itself is probably excellent, but its effects on the users systems are harmful. This is the case of chrony. Certain versions of chrony on CentOS are buggy and consume 100% of a core when queried for stats the way the chrony netdata plugin queries them. It was enabled by default before these reports.

To my understanding it is better to enable everything by default (given the rules above), and let the users disable the ones they don't need. This is actually one key differentiation of netdata: it enables everything it can - by default, without any configuration.

So, if this plugin is light and not harmful, I suggest to let it shine...

---

I am confused about the peers. Do you now know how to obsolete the peers that are no longer used?

Ferroin:
So, on the note of how much impact this has, I ran some tests overnight to slam my systems with as much NTP traffic as possible, and saw  near zero performance difference with Netdata collecting just the host stats (the actual difference in client requests served was in parts-per-million, and any latency differences were so small I couldn't measure them).  When it was collecting peer stats, it had about a 0.1% impact on throughput (that is, how many client requests got serviced in a given period of time went down by about 0.1%), and caused small (single-digit millisecond) spikes in latency about once a second.

Given that, I would personally say the impact is low enough that it can be reasonably enabled by default to collect just local stats, and collecting peer stats could potentially be enabled by default too if it weren't for the handling of peers that have been removed.

ktsaou:
wow! nice artifacts and justification!

rda0:
Sorry for my late reply and thanks a lot for all the inputs and explanations.

I'll sum-up the TODOs or what is unclear for me:

1. Update default config as outlined by @ktsaou with `localhost`, `localhost_ipv4` and `localhost_ipv6`
2. Update chart units with unit `1` to respective chart values
3. Hide uninitialized peers (stratum `16`) by filtering in `get_peers()`
4. Leave module enabled by default
5. Update chart descriptions with inputs from @Ferroin #3454 
6. Keep all dimensions for now
7. Peer charts is unclear, we have 2 options (how should we proceed?)
7.a. Disable peers by default, leave it as it is for the moment until netdata has the option to obsolete dimensions
7.b. Possibly enable peers by default, group by a family of charts per peer (one submenu per peer), obsolete charts of removed peers.

From what I understand obsoleting charts is possible with `self.charts[id].obsolete()`.

I am not sure if it would be a good solution to to group all charts with unit `ms` together in one chart (generally low values would appear much flatter in comparison to generally higher values). From what I see with a few test peers are these value ranges:

- peer_offset: 0.01 - 0.6
- peer_delay: 0.1 - 5
- peer_dispersion: 1 - 4
- peer_jitter: 0.01 - 0.1
- peer_xleave: 0.02 - 0.04
- peer_rootdelay: 0.2 - 20
- peer_rootdisp: 0.4 - 60

In option `7.a.` I like, that the dimensions are in one chart and can be easily compared.

Additional notes to `hpoll`, `ppoll` and `xleave`:

- The `*poll` variables are described in [rfc5905 section 13](https://tools.ietf.org/html/rfc5905#section-13), but at the moment I do not understand if we can just use one of them. From what I understand, they could differ in the case of network errors. It looks like during a poll, `hpoll` is increased and after the reply packet is received `ppoll` is increased.

- `xleave` is used for interleaved mode (used only in NTP symmetric and broadcast modes), from what I understand it is an internal variable to calculate the most accurate actual on wire delays by measuring the delay introduced on the system (depending also on the CPU). It is described in [NTP Interleaved Modes](http://doc.ntp.org/current-stable/xleave.html)

When I make more modifications, should I open a new pull request with a new branch or can I just use this branch and with a new pull request?

l2isbad:
@rda0 

> From what I understand obsoleting charts is possible with self.charts[id].obsolete().

All modules has [chart_cleanup](https://github.com/firehol/netdata/blob/master/conf.d/python.d/beanstalk.conf#L42-L50) option and default value is 10.

[auto obsolete](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/FrameworkServices/SimpleService.py#L220-L221) implemented in `SimpleService`

ktsaou:
@rda0 your list is good.

I suggest to have peer charts disabled by default.

Obsoleting dimensions is not going to happen soon, so we have to decide how big is this problem for people willing to enable peers. I don't know. If peers will change frequently, I think pivoting the charts is required.

Regarding the grouping of peer charts, I agree with you - they will not be very nice due to value diff between the dimensions. So, I suggest to keep the important metrics alone in their charts and the not so important groupped by unit. However, I can't tell what is important and what is not.

> When I make more modifications, should I open a new pull request with a new branch or can I just use this branch and with a new pull request?

This is entirely your decision...

Ferroin:
On the note of peer charts, the more I think about it, the more it only makes sense to even care about them if you're serving time to another system or if top quality time is mission critical.  In both cases, you should either be defining a static peer list with peers you actively trust, or using a local stratum 0 time reference.  In either case, peers should not change during normal operation of the system.  As a result, I think it probably makes sense to just leave peer charts disabled by default, and not worry too much about obsoleting dimensions (with the caveat that the memory leak it implies should be prominently documented somewhere).

ktsaou:
> . As a result, I think it probably makes sense to just leave peer charts disabled by default, and not worry too much about obsoleting dimensions (with the caveat that the memory leak it implies should be prominently documented somewhere).

I think this should be documented at the config file that describes the option to enable peer charts.

@l2isbad a question: if charts cleanup kicks in at the python plugin, then if after some time the module restores operations, would the plugin resent the chart definition to netdata and continue collecting data?

l2isbad:
Charts operations controlled by chart [flags](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L371-L376)
if chart `created` is `False` or `push` is `True` plugin will send chart definitions to `netdata`.

Сhart `penalty`  > chart `cleanup`:in this case the module [will forward the charts with the obsolete option](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L250-L253) (only for `created` charts) and will set `obsolete` flag to `True`.

All charts with `True` obsoleted flag are [skipping in update](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/FrameworkServices/SimpleService.py#L217-L219).

Chart methods which restores obsolete chart:
1. [add](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L186-L198) new dimension
2. [hide/unhide](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L200-L205) dimension

(i see they just use chart [refresh](https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L255-L258))

So no, plugin doesn't check on each iteration if there is new data for obsolete charts.

l2isbad:
Actually it's easy to add auto refresh for obsoleted charts.

ktsaou:
> Actually it's easy to add auto refresh for obsoleted charts.

can you do it?

Since this feature works on all python charts, if let's say I monitor an apache and I stop it (stop apache) for 1 minute, then it won't restart chart updates when I restart apache. I'll have to restart netdata too. Right?

If this is the case, then the way it works now is buggy...

l2isbad:
> I monitor an apache and I stop it (stop apache) for 1 minute, then it won't restart chart updates when I restart apache. I'll have to restart netdata too. Right?

No.

The chart will be suppressed **only if** it has not been updated `chart_cleanup` iterations in a row **and** module `get_data()` didn't fail (unhandled exception or returns `None` or any other value besides not empty `dict`).

In case of `apache stop` `get_data() `returns `None` and there is no penalty for charts.

l2isbad:
> can you do it?

in https://github.com/firehol/netdata/pull/3491

PR #3434, opened 2018-02-17
apps.plugin now reads /proc/PID/status;

### LINKED ISSUE #3405: How to display all users on graphs in the Users section
Is it possible to have all the users running processes on the graphs in the Users section?
I seem to be failing in finding the right configuration file which governs this section.
I can see users like www-data or root but not all the users. For example smbd from the samba suite launches a process for every user using that very system user. I have around 20 samba users on this production server and I am trying to see the resources they use individually, especially the disk throughput.
Any insight will be greatly apreciated as I spent a couple of hours compiling samba with profiling to get the netdata graph the samba metrics, but that is not offering what I need.

Regards,
Dan

ktsaou:
The following changes have been made to apps.plugin:

1. apps.plugin did not recognize the user of a process, if the process was started as root and then switched to another user. The problem was that it was using the ownership of `/proc/PID/stat` to detect the user and group. After this PR, the plugin reads user and group information from the contents of `/proc/PID/status`. fixes #3405

2. apps.plugin now does not use `/proc/PID/statm`. The same information is available in `/proc/PID/status`, so the later is now used.

3. apps.plugin now reports swap usage per process / user /group. Since the plugin now reads `/proc/PID/status`, it also extracts `VmSwap` from it.

Unfortunately, `/proc/PID/status` is a big file, this means that even after the optimizations (removed `statm` and a `stat()` call to get ownership), apps.plugin is now less resource efficient (uses more CPU and memory).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -248,6 +248,11 @@ if FREEBSD
 apps_plugin_SOURCES += \
 	plugin_freebsd.h \
 	$(NULL)
+else
+apps_plugin_SOURCES += \
+	adaptive_resortable_list.c \
+	adaptive_resortable_list.h \
+	$(NULL)
 endif
 
 apps_plugin_LDADD = \
--- a/src/adaptive_resortable_list.c
+++ b/src/adaptive_resortable_list.c
@@ -96,9 +96,14 @@ void arl_begin(ARL_BASE *base) {
     }
 #endif
 
-    if(unlikely(base->added || base->iteration % base->rechecks) == 1) {
+    if(unlikely(base->iteration > 0 && (base->added || (base->iteration % base->rechecks) == 0))) {
+        int wanted_equals_expected = ((base->iteration % base->rechecks) == 0);
+
+        // fprintf(stderr, "\n\narl_begin() rechecking, added %zu, iteration %zu, rechecks %zu, wanted_equals_expected %d\n\n\n", base->added, base->iteration, base->rechecks, wanted_equals_expected);
+
         base->added = 0;
-        base->wanted = 0;
+        base->wanted = (wanted_equals_expected)?base->expected:0;
+
         ARL_ENTRY *e = base->head;
         while(e) {
             if(e->flags & ARL_ENTRY_FLAG_FOUND) {
@@ -107,7 +112,7 @@ void arl_begin(ARL_BASE *base) {
                 e->flags &= ~ARL_ENTRY_FLAG_FOUND;
 
                 // count it in wanted
-                if(e->flags & ARL_ENTRY_FLAG_EXPECTED)
+                if(!wanted_equals_expected && e->flags & ARL_ENTRY_FLAG_EXPECTED)
                     base->wanted++;
 
             }
@@ -155,10 +160,11 @@ void arl_begin(ARL_BASE *base) {
 
 // register an expected keyword to the ARL
 // together with its destination ( i.e. the output of the processor() )
-ARL_ENTRY *arl_expect(ARL_BASE *base, const char *keyword, void *dst) {
+ARL_ENTRY *arl_expect_custom(ARL_BASE *base, const char *keyword, void (*processor)(const char *name, uint32_t hash, const char *value, void *dst), void *dst) {
     ARL_ENTRY *e = callocz(1, sizeof(ARL_ENTRY));
     e->name = strdupz(keyword);
     e->hash = simple_hash(e->name);
+    e->processor = (processor)?processor:base->processor;
     e->dst = dst;
     e->flags = ARL_ENTRY_FLAG_EXPECTED;
     e->prev = NULL;
@@ -198,7 +204,7 @@ int arl_find_or_create_and_relink(ARL_BASE *base, const char *s, const char *val
 
         // run the processor for it
         if(unlikely(e->dst)) {
-            base->processor(e->name, hash, value, e->dst);
+            e->processor(e->name, hash, value, e->dst);
             base->found++;
         }
 
@@ -254,8 +260,10 @@ int arl_find_or_create_and_relink(ARL_BASE *base, const char *s, const char *val
     if(unlikely(!base->next_keyword))
         base->next_keyword = base->head;
 
-    if(unlikely(base->found == base->wanted))
+    if(unlikely(base->found == base->wanted)) {
+        // fprintf(stderr, "FOUND ALL WANTED 1: found = %zu, wanted = %zu, expected %zu\n", base->found, base->wanted, base->expected);
         return 1;
+    }
 
     return 0;
 }
--- a/src/adaptive_resortable_list.h
+++ b/src/adaptive_resortable_list.h
@@ -51,6 +51,9 @@ typedef struct arl_entry {
 
     uint8_t flags;          // ARL_ENTRY_FLAG_*
 
+    // the processor to do the job
+    void (*processor)(const char *name, uint32_t hash, const char *value, void *dst);
+
     // double linked list for fast re-linkings
     struct arl_entry *prev, *next;
 } ARL_ENTRY;
@@ -102,7 +105,8 @@ extern void arl_free(ARL_BASE *arl_base);
 
 // register an expected keyword to the ARL
 // together with its destination ( i.e. the output of the processor() )
-extern ARL_ENTRY *arl_expect(ARL_BASE *base, const char *keyword, void *dst);
+extern ARL_ENTRY *arl_expect_custom(ARL_BASE *base, const char *keyword, void (*processor)(const char *name, uint32_t hash, const char *value, void *dst), void *dst);
+#define arl_expect(base, keyword, dst) arl_expect_custom(base, keyword, NULL, dst)
 
 // an internal call to complete the check() call
 extern int arl_find_or_create_and_relink(ARL_BASE *base, const char *s, const char *value);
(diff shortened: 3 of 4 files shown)


ktsaou:
@vlvkobal I don't know how to add swap usage per process on FreeBSD. I would appreciate it if you could have a look.

ktsaou:
Unfortunately apps.plugin is now about 20% more "expensive" in term of CPU:

screenshot from 2018-02-17 15-03-10

(similarly for RAM, but it is ignorable, from 2,5MB to 3,2 MB).

ktsaou:
@fredericopissarra can you help me optimize this a bit - I am out of ideas.

The purpose is to parse `/proc/PID/status` in the fastest possible way. apps.plugin reads this file, every second for every pid found running. So, this is repeated hundreds to thousands of times per second.

To parse all proc files, netdata uses https://github.com/firehol/netdata/blob/master/src/procfile.c and https://github.com/firehol/netdata/blob/master/src/procfile.h - these are used everywhere.

procfile read the file with just one `read()` and then creates an array of lines. Each line has an array of words. Pointers are kept directly in the buffer read to allow random access to any line and any word. Of course, allocations are avoided - procfile remembers the sizes of lines and words and re-uses them on each iteration.

I have traced the performance penalty to `procfile_parser()`. This goes through the whole buffer and creates the structures needed for lines and words. https://github.com/firehol/netdata/blob/64e130fac39f577abea6560e555003287951b1f0/src/procfile.c#L142-L272

I found that using `if else if else if` is faster than `switch()`. Any ideas how to make this faster?

ktsaou:
I noticed the total swap used reported is more than the sum of all the swap reported to processes:

```sh
# find the total swap used on the system using /proc/meminfo
# echo "SwapUsed: $(( $(awk '/^SwapTotal:/{printf $2}END{ print ""}' /proc/meminfo) - $(awk '/^SwapFree:/{printf $2}END{ print ""}' /proc/meminfo) ))"
SwapUsed: 11520

# find the total swap used by all processes
# for file in /proc/*/status ; do awk '/VmSwap:/{printf $2}END{ print ""}' $file; done | { total=0; while read kb; do total=$((total + kb)); done; echo $total; }
7252
```

I guess the system is reserving swap for other purposes too...

fredericopissarra:
MAYBE if you give a little hand to the compiler, switch will be, in fact, faster then nested if..then.elses. And, making complex routines inline tends to puts more pressure on L1i cache, maybe causing a lot of cache misses a slowing things down.

Also, I've notice the routines pfwords_add() and pflines_add() are using reallocz using the same pointer as source and destination... what will happen if reallocz fails? The same for pfwords_new() and pflines_new().

Here some changes I HOPE will help you:

```c
// FIX: Sometimes, small funcions like this can be made faster if not inline.
//      Reason: To make a better use of L1i cache.
__attribute__((noinline))
static void _procfile_addword(char *s, size_t *w, pfwords *ff, char *t, size_t l)
{
  *s = '\0';
  ff->words = pfwords_add(ff->words, t);
  ff->lines->lines[l].words++;
  w++;
}

// FIXME: The problem with making big functions like this 'inline' is
//        the pressure on L1i cache. Maybe this 'inline' hint should not be used here.
static inline void procfile_parser(procfile *ff)
{
  // debug(D_PROCFILE, PF_PREFIX ": Parsing file '%s'", ff->filename);

  char *s,    // our current position.
       *e,    // the terminating nul.
       *t;    // the first char of a quoted or parenthesized string

  e = (s = t = ff->data) + ff->len;

  // the look up array to find our type of character
  PF_CHAR_TYPE *separators = ff->separators;

  char quote = 0;       // the quote character - only when in quoted string

  size_t
    l = 0,              // counts the number of lines we added
    w = 0,              // counts the number of words we added
    opened = 0;         // counts the number of open parenthesis

  ff->lines = pflines_add(ff->lines, w);

  // 'likely', here, is inocuous (i think!)...
  while (s < e)
  {
    // we are not at the end
    PF_CHAR_TYPE ct = separators[(size_t)(*s)];

    // MAYBE giving a little hand to the compiler, switch can be faster then nested if..then..else...
    // Must be measured...
    switch (ct)
    {
    case PF_CHAR_IS_SEPARATOR:
      // FIX: Boolean AND is better for short-ciruit!
      if (!quote && !opened)
      {
        // FIXME: likely or unlikely?
        if (s != t)
          _procfile_addword(s, &w, ff, t, l);

        t = s + 1;
      }
      // NOTE: No 'break' here.

    case PF_CHAR_IS_WORD:
      s++;
      break;

    case PF_CHAR_IS_NEWLINE:
      // end of line
      _procfile_addword(s, &w, ff, t, l);

      // debug(D_PROCFILE, PF_PREFIX ":   ended line %d with %d words", l, ff->lines->lines[l].words);

      ff->lines = pflines_add(ff->lines, w);
      l++;
      break;

    case PF_CHAR_IS_QUOTE:
      // FIXME: Instead of testing 2 expressions, one at a time can create a better code.
      //        Sometimes simplier ways are better than compelx ones.
      if (!quote)
      {
        if (s == t)
        {
          quote = *s;
          t = s + 1;
        }
      }
      else
      {
        if (quote == *s)
        {
          quote = 0;
          _procfile_addword(s, &w, ff, t, l);
          t = s + 1;
        }
      }

      s++;
      break;

    case PF_CHAR_IS_OPEN:
      if (s == t)
      {
        opened++;
        t = s + 1;
      }
      else
        if (opened)
          opened++;
      s++;
      break;

    case PF_CHAR_IS_CLOSE:
      if (opened)
      {
        opened--;

        if (!opened)
        {
          _procfile_addword(s, &w, ff, t, l);
          t = s + 1;
        }
      }

      s++;
      break;

    default:
      fatal("Internal Error: procfile_readall() does not handle all the cases.");
    }
  }

  if (likely(s > t && t < e))
  {
    // the last word
    if (likely(ff->len < ff->size))
      *s = '\0';
    else
    {
      // we are going to loose the last byte
      ff->data[ff->size - 1] = '\0';
    }

    ff->words = pfwords_add(ff->words, t);
    ff->lines->lines[l].words++;
  }
}
```

ktsaou:
thanks! Checking it now....

> Also, I've notice the routines pfwords_add() and pflines_add() are using reallocz using the same pointer as source and destination... what will happen if reallocz fails? The same for pfwords_new() and pflines_new().

netdata exits when any memory allocation fails. These are calls to `reallocz()`, not `realloc()`.

ktsaou:
Well, I did various tests without any significant improvements.

I general:

1. if else if else, seems same with a swtich. Based on this http://lazarenko.me/switch, I preferred to keep the conditional version.
2. the `noinline` function does not seem to help either - I guess the cost of a function call is quite high.

So I tried to:

- eliminate memory look ups - I refactored a bit the functions to avoid frequent references to array and structure members, especially the ones with many hops (ie. `ff->lines->lines[line].words++`).

- eliminate function parameters and function return values when possible.

- eliminate local variables in `procfile_parse()` (to save registers)

- prefer inline functions hopping that high end systems will benefit from function inlining due to their larger processor caches

It is in PR #3436, however even after all those, cpu utilization is pretty much the same.

fredericopissarra:
Ok... I think when you use multiple references as ``ff->lines->lines[line].words``, with optimizations enabled, the compiler will apply pointer aliasing to ``ff->lines->lines``. So, this kind of memory lookup is not too critical. But, of course, it is always a good idea to give the compiler a hand...

In my experience ``switch`` tends to create better code then nested ifs (sometimes the compiler builds a jump table)... The cited article, maybe, doesn't account for compiler optimizations (-O2 or -O3)...

As I said, inlining big functions puts pressure to L1i cache, and its size is more or less standard in every modern 80x86 processors (32 KiB - the remaining 32 KiB is for L1d (data) cache). On 'high-end' 80x86 processors we have mode L2 cache, even L3 cache, but what matters to realtime performance is L1. That's why I said "sometimes" a function call can perform better then inlining, avoiding cache evictions, or minimizing them. It can perform even better if the called function is less then 64 bytes long (will fit in a single cache line).

Yet, it's a good idea to avoid function calls with multiple arguments! :)

But, maybe the algorithm is the problem. Instead scanning the string character by character, separate every "symbol" with strtok_r or strsep can perform better.

ktsaou:
> But, maybe the algorithm is the problem. Instead scanning the string character by character, separate every "symbol" with strtok_r or strsep can perform better.

I thought that too. Now the function is splitting every word found, even if the line is not actually used. So, I did a test, and removed all parsing. Still no big difference.

Then it case to be that the cpu reported is like this:

image

So, 80% is system and only 20% is user.

The following are from a server, before and after adding `/proc/pid/status` reads:

image

As you can see both system and user got increased, but mainly the issue is system.

user increased by 0.5ms/sec:

image

but system increased by (almost) 2ms/sec:

image

If I understand this right, all our optimizations affect the 20% only - so they are ignorable.

Is it possible that the time taken by the kernel to print `/proc/PID/status` gets accumulated in our system cpu usage? If this is the case, the only way to improve this significantly, is to prevent a few reads entirely.

ktsaou:
btw, I did a benchmark with your `poor man cycle counting` (https://ghostbin.com/paste/k5fz6) and found that:

1. by not inlining `pfwords_add()`, the parser is consistently 1% more expensive compared to the inlined version

2. the `swtich()` method is 22% more expensive compared to the `if else if else` method.

(for the test I iterated 1 mil times to have consistent / stable measurements).

PR #3436, opened 2018-02-18
procfile optimizations

ktsaou:
This is as good as it can get.

I believe proc file parsing cannot be made faster without per-file specific heuristics (which I strongly believe we should avoid).

So, this PR eliminates a few function calls, a few function call parameters, a few memory lookups and a few local variables, while parsing proc files.

The procfile routines significantly influence apps.plugin - more than half of its execution time is inside `procfile_parser()`. The addition of `/proc/PID/status` significantly affected its cpu usage, however I think we can't do anything more to help it. This is the fastest it can get (without heuristics).


ktsaou:
The cost of `/proc/PID/status` is huge:

apps plugin

PR #3440, opened 2018-02-19
Fixed IDE warning that tracked files are ignored in .gitignore

ccremer:
I was just about to get started in developing a new plugin, when I got this warning.

The previous .gitignore entries were ignoring both `/node_modules` and `/node.d/node_modules`, however `/node.d/node_modules` contains netdata specific modules that shouldn't get ignored (I think). The fix now only ignores `/node_modules`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,8 +69,8 @@ web/index_new.html
 web/version.txt
 
 # related to karma/javascript/node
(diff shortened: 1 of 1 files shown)


ktsaou:
merged - thanks!

PR #3444, opened 2018-02-20
isc_dhcp fix: do not include leases without "ends"

l2isbad:
fixes https://github.com/firehol/netdata/issues/3442

@HammaBamma 
please check it

### DIFF
--- a/plugins.d/python.d.plugin
+++ b/plugins.d/python.d.plugin
@@ -316,10 +316,10 @@ class Plugin(object):
                 job.checked = True
                 continue
             if not job.is_autodetect() or ok is None:
-                job.error('check() => [FAILED]')
+                job.info('check() => [FAILED]')
                 self.delete_job(job)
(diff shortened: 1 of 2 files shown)


ktsaou:
let me know when ready to merge...

HammaBamma:
It works on my primary server as expected, just not on the backup server. I'm going to look further into the isc-dhcpd-server configs, to see why it does the backup state, and if there's a method for it to do some other output. Also, I'd be curious to see if the output changes once the server fails over, and then the script would read it correctly, and start outputting. Thus I could just add both service checks, and one as 0 until it's failed over.. ? Who knows... But thanks for this awesome tool and the quick replies.. I'm going to implement this for a few of my upcoming events and evangelize this!

l2isbad:
@ktsaou merge then

ktsaou:
merged! thanks!

PR #3447, opened 2018-02-20
Add port check plugin

ccremer:
* Is it efficient?
  I don't know, probably not so, since checks for multiple ports are executed sequentially
* Is it secure?
  Depends on your firewall, it may trigger portscan protection
* Why?
  1. For simple use cases where nagios etc. would introduce another unnecessary tool and maintenance
  2. API access for custom dashboards
  3. When using a minimal netdata installation in docker-compose file as slave for health checking a docker image (I'm working on it).

I would like to add alarm templates, but I don't know how (expression) and what a sensible default would be. Any quick ideas? E.g. warning when error code > 0 for last 20 sec, critical over last 1 min. To catch service flapping maybe take averages (e.g. average err code of >0.2) ?

portcheck


ktsaou:
> Is it efficient?
> I don't know, probably not so, since checks for multiple ports are executed sequentially

I believe python.d.plugin runs multiple jobs in parallel. So, it would be best to configure multiple port checks as multiple jobs . @l2isbad can help you more with that.

> I would like to add alarm templates, but I don't know how (expression) and what a sensible default would be. Any quick ideas?

1. gaps should fire an alarm
2. I generally compare the average of the last hour to the average of the last minute. If the minute average is 2x - 3x the hourly average I think it should fire an alarm

The expressions are very simple. Once you have this working, attach here a snapshot and I'll help.

ccremer:
> I believe python.d.plugin runs multiple jobs in parallel. So, it would be best to configure multiple port checks as multiple jobs

I just made these changes in a4327d6. Since I created a dimension for each error type, I figured that multiple jobs are better suited than a single job with multiple ports.

For the alarms I'm going for the error types. e.g. require an 90% uptime of the `success` dimension during the last 1 or 2 minutes. The error type dimensions are binary (1 or 0), so I can directly calculate an average percentage here and still give different alarms depending on error type. I'm leaving the latency chart as a reference for now, since the accuracy of the measurements are apparently not so good (python) #3448 

Are compound alarms possible? e.g. if there are 4 alarm templates for each error type (timeout, unexpected content or status code, ...), but send only one notification if one or more sub-alarms trigger?

ktsaou:
> Are compound alarms possible? e.g. if there are 4 alarm templates for each error type (timeout, unexpected content or status code, ...), but send only one notification if one or more sub-alarms trigger?

not yet, but this is a nice idea...

ccremer:
I made similar alarms to the one in httpcheck. However, can you help me figuring out a sensible latency alarm?

ccremer:
I somehow didn't receive notification about your comment in httpcheck. Please ignore the above for now

l2isbad:
If i [understand it right](https://github.com/firehol/netdata/pull/3448#issuecomment-368266777) this one will be implemented in C.

ktsaou:
> If i understand it right this one will be implemented in C.

Yes, it will eventually, but it does not harm to have alternatives... the whole netdata evolves as time passes...

ccremer:
Please review this too.

ccremer:
bump?

ktsaou:
@l2isbad is this ok to be merged?

ktsaou:
merged! thank you!

PR #3454, opened 2018-02-22
Update documentation for ntpd

rda0:
Documentation updated as requested in #3421.

I added descriptions for the system charts only. Because the peer charts are not shown by default and I do not know what those charts mean :smile: 

The wiki is updated as well.

### DIFF
--- a/conf.d/python.d.conf
+++ b/conf.d/python.d.conf
@@ -56,6 +56,7 @@ go_expvar: no
 # nginx: yes
 # nginx_plus: yes
 # nsd: yes
+# ntpd: yes
 
 # nginx_log has been replaced by web_log
 nginx_log: no
--- a/conf.d/python.d/ntpd.conf
+++ b/conf.d/python.d/ntpd.conf
@@ -58,7 +58,10 @@
 #
 # host: 'localhost'       # the host to query
 # port: '123'             # the UDP port where `ntpd` listens
-# show_peers: no          # use `yes` to show peer charts
+# show_peers: no          # use `yes` to show peer charts. enabling this
+#                         # option is recommended only for debugging, as
+#                         # it could possibly imply memory leaks if the
+#                         # peers change frequently.
 # peer_filter: '127\..*'  # regex to exclude peers
 #                         # by default local peers are hidden
 #                         # use `''` to show all peers.
--- a/conf.d/python.d/ntpd.conf
+++ b/conf.d/python.d/ntpd.conf
@@ -71,3 +71,21 @@
 # ----------------------------------------------------------------------
 # AUTO-DETECTION JOBS
 # only one of them will run (they have the same name)
+
+localhost:
+  name: 'local'
+  host: 'localhost'
+  port: '123'
+  show_peers: no
+
+localhost_ipv4:
+  name: 'local'
+  host: '127.0.0.1'
+  port: '123'
+  show_peers: no
+
+localhost_ipv6:
+  name: 'local'
+  host: '::1'
+  port: '123'
+  show_peers: no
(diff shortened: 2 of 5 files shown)


Ferroin:
For the peer stats:

* `peer_ofset` is the offset of the peer clock relative to the system clock in milliseconds.  Smaller values here weight peers more heavily for selection _after_ the initial synchronization of the local clock.  For a system providing time service to other systems, these should be as low as possible (I aim for +/- 5ms on the two systems I have in the ntp.org server pool).
* `peer_delay` is the RTT for communication with that peer.  In essence, this is the timing you would get back from pinging that peer.  This isn't as critical as either the offset or jitter, but is still factored into the selection algorithm (because as a general rule lower delay means more accurate time).  In most cases, it should be below 100ms.
* `peer_dispersion` is a measure of the estimated error between the peer and the local system.  Lower values here are better.  If I understand correctly, this has to do with things like the frequency offset.
* `peer_jitter` is the essentially a remote estimate of that peers `system_jitter` value.  Lower values here weigh highly in favor of peer selection, and this is a good indicators of overall quality of a given time server (good servers will have values not exceeding single digit milliseconds here, with high quality stratum one servers regularly having sub-millisecond jitter).
* `peer_xleave` I'm not quite sure about this one, but I think it has something to do with the processing delay between a packet being queued and actually being physically sent by the network hardware.
* `peer_rootdelay` For a stratum 1 server, this is the access latency for the reference clock.  For lower stratum servers, it's the sum of the `peer_delay` and `peer_rootdelay` for the system they are syncing off of.  Similarly to `peer_delay`, lower values here are technically better, but have limited influence in peer selection.
* `peer_rootdisp` Is the same thing as `peer_rootdelay`, but measures accumulated `peer_dispersion` instead of accumulated `peer_delay`.
* `peer_stratum` Shows what stratum each peer is.  Should never change in most cases.
* `peer_hmode` and `peer_pmode`  These give info about what mode the packets being sent to and received from a given peer are.  Mode 1 is symmetric active (both the local system and the remote peer have each other declared as peers in `/etc/ntp.conf`), Mode 2 is symmetric passive (only one side has the other declared as a peer), Mode 3 is client, Mode 4 is server, and Mode 5 is broadcast (also used for multicast and manycast operation).  Modes 0, 6, and 7 should never be seen here.
* `peer_hpoll` and `peer_ppoll` log2 representation of the polling interval in seconds.  I'm not quite sure what the difference here is between the two values (they're the same for all peers on all my servers).
* `peer_precision` Not sure about this one either.

ktsaou:
@rda0 do you have any pending items about ntp?
I want to release netdata v1.10 and I don't want to distribute this incomplete.

rda0:
@ktsaou sorry for the delay. I updated `ntpd` documentation and module code as agreed in https://github.com/firehol/netdata/pull/3421#issuecomment-369007351 (item 1 to 6). Unfortunately I do not have enough free time at the moment to do item 7. Would it be OK for you if we change that later?

l2isbad:
@rda0 
Lets [add peer charts](https://github.com/firehol/netdata/blob/master/python.d/ntpd.chart.py#L228) only if [show peers](https://github.com/firehol/netdata/blob/master/python.d/ntpd.chart.py#L236) is `True` (it produces a lot of error logs `...was suppressed due to non updating...`)

ktsaou:
I think that if we ship this with peer charts disabled by default, it would be ok for most users.

rda0:
> Lets add peer charts only if show peers is True

@l2isbad done :+1:

ktsaou:
merged - thank you!

Issue #3435, opened 2018-02-17
IRC notifications


io7m:
Hello!

I notice from reading the wiki that there are a lot of ways to send notifications to the current trendy messaging services (Discord, Slack, etc).

Is there no way to send alerts to an IRC channel for those of us still living in the previous century?

ktsaou:
Is there a command line that can post an iirc message to a channel? If yes, the fix is very simple.

You can edit https://github.com/firehol/netdata/blob/master/plugins.d/alarm-notify.sh and copy-paste-rename one the existing methods to call that command line.

Once you do this, please make a PR.

Ferroin:
I think supporting irker would be nice (and actually should be insanely easy to do), but it would also be nice to not need yet another daemon running somewhere if that is at all possible.

The problem is that IRC is very much a stateful connection-oriented protocol designed to idle as efficiently as possible at the cost of connection setup.  It's not the type of protocol that you can easily just connect to send a single message and then disconnect, and expect to do so at a potentially very high frequency.  irker works as well as it does because it maintains a persistent connection to the server.

io7m:
There is this thing:

https://tools.suckless.org/ii/

It exposes a filesystem-based IRC client. Sending a notification to a channel is achieved by writing a message to a named pipe in the filesystem.

I'll do a bit of experimenting.

ktsaou:
yes

PR #3458, opened 2018-02-22
Added IRC Notifications support.

manosf:
Added IRC Notification support per issue #3435.


plugins.d/alarm-notify.sh
```diff
@@ -316,6 +317,13 @@ DEFAULT_RECIPIENT_EMAIL="root"
 EMAIL_CHARSET=$(locale charmap 2>/dev/null)
 declare -A role_recipients_email=()
 
+# irc configs
+IRC_NICKNAME=
+IRC_REALNAME=
+DEFAULT_RECIPIENT_IRC=
+DEFAULT_IRC_NETWORK="irc.freenode.net"
```
ktsaou: I think `DEFAULT_` is not needed here.

plugins.d/alarm-notify.sh
```diff
@@ -631,6 +651,9 @@ done
 # check kafka
 [ -z "${KAFKA_URL}" -o -z "${KAFKA_SENDER_IP}" ] && SEND_KAFKA="NO"
 
+# check irc
+[ -z "${DEFAULT_IRC_NETWORK}" -o -z "${DEFAULT_RECIPIENT_IRC}" ] && SEND_IRC="NO"
```
ktsaou: Here too.

plugins.d/alarm-notify.sh
```diff
@@ -659,6 +682,7 @@ if [ \( \
         -o "${SEND_PUSHBULLET}"  = "YES" \
         -o "${SEND_KAFKA}"       = "YES" \
         -o "${SEND_CUSTOM}"      = "YES" \
+        -o "${SEND_IRC}"         = "YES" \
```
ktsaou: You don't depend on `curl`, so `SEND_IRC` should not be in this list.

plugins.d/alarm-notify.sh
```diff
@@ -678,6 +702,7 @@ if [ \( \
         SEND_KAVENEGAR="NO"
         SEND_KAFKA="NO"
         SEND_CUSTOM="NO"
+        SEND_IRC="NO"
```
ktsaou: Same here.

plugins.d/alarm-notify.sh
```diff
@@ -1443,6 +1469,39 @@ EOF
...
+        CLEAR)    color="good" ;;
+        *)        color="#777777" ;;
+    esac
+
+    if ["${SEND_IRC}" = "YES" -a ! -z "${nickname}" -a ! -z "${realname}" -a ! -z "${channels}" -a ! -z "${network}" -a ! -z "${servername}"]
+    then
+        for channel in ${CHANNELS}
+        do
+            sentcheck=$(echo -e "USER ${NICKNAME} guest ${REALNAME} ${SERVERNAME}\nNICK ${NICKNAME}\nJOIN ${CHANNEL}\nPRIVMSG ${CHANNEL}> :${MESSAGE}\nQUIT\n" \ | nc ${NETWORK} 6667)  
+            if [ $(echo ${sentcheck} | grep --fixed-strings --count ":${nickname} MODE ${nickname} :+i") -gt 0 ]
```
ktsaou: I suggest to set a variable with this calculation, so that you can print it below instead of `httpcode` (which is not used).

plugins.d/alarm-notify.sh
```diff
@@ -1443,6 +1469,39 @@ EOF
...
+            then
+                info "sent irc notification for: ${host} ${chart}.${name} is ${status} to '${channel}'"
+                sent=$((sent + 1))
+            else
+                error "failed to send irc notification for: ${host} ${chart}.${name} is ${status} to '${channel}', with HTTP error code ${httpcode}."
```
ktsaou: `httpcode` is not used

plugins.d/alarm-notify.sh
```diff
@@ -1657,6 +1716,16 @@ SENT_KAFKA=$?
 send_pd "${to_pd}"
 SENT_PD=$?
 
+# -----------------------------------------------------------------------------
+# send the irc message
+
+send_irc "${IRC_NICKNAME}" "${IRC_REALNAME}" "${to_irc}" "${DEFAULT_IRC_NETWORK}" "${host}" "${host} ${status_message} - ${name//_/ } - ${chart} ----- ${alarm}
```
ktsaou: `DEFAULT_IRC_NETWORK` should be `IRC_NETWORK`

ktsaou:
nice! I made a few comments.
Please also adapt https://github.com/firehol/netdata/blob/master/conf.d/health_alarm_notify.conf

manosf:
I made the changes that you requested and fixed my mistakes and oversights.
I also changed the approach of error detection to comply with IRC's reply codes
in order to be easier to detect and properly inform the user.
Thank you for your comments!

conf.d/health_alarm_notify.conf
```diff
@@ -392,6 +399,34 @@ SEND_PD="YES"
...
+#------------------------------------------------------------------------------
+# irc notification options
+#
+# irc notifications require only the nc utility to be installed. 
+
+# multiple recipients can be given like this:
+#              "<irc_channel_1> <irc_channel_2> ..."
+
+# enable/disable sending irc notifications
+SEND_PD="YES"
```
ktsaou: I think this should be `SEND_IRC="YES"`

plugins.d/alarm-notify.sh
```diff
@@ -1443,6 +1467,48 @@ EOF
...
+
+    case "${status}" in
+        WARNING)  color="warning" ;;
+        CRITICAL) color="danger" ;;
+        CLEAR)    color="good" ;;
+        *)        color="#777777" ;;
+    esac
+
+    if ["${SEND_IRC}" = "YES" -a ! -z "${nickname}" -a ! -z "${realname}" -a ! -z "${channels}" -a ! -z "${network}" -a ! -z "${servername}"]
+    then
```
ktsaou: I think it needs a space before `]`.

plugins.d/alarm-notify.sh
```diff
@@ -1443,6 +1467,48 @@ EOF
...
+
+send_irc() {
+    local NICKNAME="${1}" REALNAME="${2}" CHANNELS="${3}" NETWORK="${4}" MESSAGE="${5}" sent=0 channel color send_alarm reply_codes error
+
+    case "${status}" in
+        WARNING)  color="warning" ;;
+        CRITICAL) color="danger" ;;
+        CLEAR)    color="good" ;;
+        *)        color="#777777" ;;
+    esac
```
ktsaou: please move this `case` block, inside the `if` below. There is no meaning for this to be running if the irc method is not enabled.

plugins.d/alarm-notify.sh
```diff
@@ -1443,6 +1467,48 @@ EOF
...
+            send_alarm=$(echo -e "USER ${NICKNAME} guest ${REALNAME} ${SERVERNAME}\nNICK ${NICKNAME}\nJOIN ${CHANNEL}\nPRIVMSG ${CHANNEL}> :${MESSAGE}\nQUIT\n" \ | nc ${NETWORK} 6667)  
+            reply_codes=$(echo ${send_alarm} | cut -d ' ' -f 2 | grep -o '[0-9]*')
+            for code in ${reply_codes}
+            do
+                if [ "${code}" -ge 400 -a "${code}" -le 599 ]
+                then
+                    error "failed to send irc notification for: ${host} ${chart}.${name} is ${status} to '${channel}', with error code ${code}."
+                    error=1
+                fi
+            done
```
ktsaou: Please move the `error` line outside the loop. You can write it like this:

```sh
error=0
for code in ${reply_codes}
do
    [ "${code}" -ge 400 -a "${code}" -le 599 ] && error=1 && break
done

if [ ${error} -eq 0 ]
then
   info "success ..."
else
   info "failure ..."
fi
```

manosf:
Made all adjustments you proposed and fixed the mistakes I overlooked earlier.
Thank you again for the comments!

ktsaou:
ok, this is good! nice work.

I think it can be merged. Have you tested it?

You will also need to create a simple wiki page, like this: https://github.com/firehol/netdata/wiki/slack-notifications and link it to the wiki menu.

To add an screenshot (please do), upload the image as a comment here, so that github will give the image a URL, and then copy-paste the image markdown into the wiki page.

You can also add IRC to the infographic (I can do this if it is too complex for you).
Here is how:

1. Go to https://www.draw.io/
2. login with github and select "open existing diagram", then navigate to `diagrams/netdata-overview.xml`  in your fork of the repo
3. to add an IRC image, first upload it to a comment on this PR. It is important to use the github CDN for the images, because third party URLs break all the time, and if you add the image itself to draw.io it will be added into the saved xml file, making it huge. So, upload first the image here and then copy its URL and paste it (the URL) to draw.io (it will turn the URL into an image)
4. re-arrange the notification images to make room for your IRC image
5. right click on the image and select "edit link" and paste the URL of the wiki page you added
6. save it - it will be added to this PR.
7. make a PNG of it (File->Export->PNG) and upload it as a comment here, so that we can all see your work...

btw, this diagram is interactive and we use it as an index to documentation: https://my-netdata.io/infographic.html (your changes will automatically appear to that URL once this PR is merged).

manosf:
Sorry about the late answer, it took me some time but the tests are successful after some minor tweaks!

Regarding the testing itself, in order to provide a screenshot for the wiki page, I only used the test option of the alarm-notify.sh as seen below. If you find the output lacking, I could possibly generate some more custom notifications in order to screenshot more statuses and alerts. I will start making the wiki page as soon as you let me know.

As for the infographic, I believe there won't be a problem, thank you for your explicit instructions!

IRCCloud web client:
deepinscreenshot_select-area_20180227224119

Irssi:
deepinscreenshot_select-area_20180227225734

ktsaou:
Perfect. Just use smaller windows for screenshots, so that they will be visible (the wiki does not zoom the screenshots when clicked).

ktsaou:
btw, you haven't submitted the minor fixes to make it work.

`alarm-notify.sh` and its config are somewhat sensitive files. If we accidentaly break them, people that auto-update their netdata will silently not recieve notifications. This should not happen. So, please make sure you have tested it well...

manosf:
@ktsaou I retested the method today before pushing, in order to verify that there are indeed no more changes needed.
I created the wiki page (https://github.com/firehol/netdata/wiki/IRC-notifications) and updated the infographic as you instructed. I believe everything is working as intended!

netdata

ktsaou:
ok, perfect. I also added it to the wiki menu:
image

ktsaou:
merged! thank you!

PR #3462, opened 2018-02-23
monitor hugepages; 

### LINKED ISSUE #3461: Support Infos about HugePages
Hi,
I monitor some Oracle Database Systems which use Hugepages:

cat /proc/meminfo |grep -i Hugep
AnonHugePages:    368640 kB
**HugePages_Total:   53342
HugePages_Free:    11904**
HugePages_Rsvd:        0
HugePages_Surp:        0
**Hugepagesize:       2048 kB**

Those are reserved at front and show up as used RAM:

image

Would be awsome if used could be separated to show HugePages_Total and Free.
So I could see who much is used by other stuff and If I have lot's of Free HugePages that are "wasted" because can't be used by other stuff.

ktsaou:
fixes #3461 

The kernel supports hugepages with multiple sizes. This PR does not support this kind of setup. In order of it to be supported, we should use `/sys/kernel/mm/hugepages` in which there is a directory for each size of hugepages supported.

screenshot from 2018-02-23 15-07-26


h0lzi:
Well hugepage size depends on what your CPU supports. And you can set Hugepagesize
to only one vaule iirc
 and Hugepagesize of  2048 kB
ist the default and the recommended Size e.g for Oracle Databases.

Ferroin:
@h0lzi You can actually set up and statically allocate multiple sizes with options on the kernel command-line.  On modern x86-64, you can have 2MB or 1GB huge pages (though the 1GB ones are almost never used).  Other architectures also have multiple size options, though they are generally much smaller.  Such usage is very rare though, and was not part of the original implementation (hence `/proc/meminfo` assuming you only use one size).

@ktsaou Have you tested this at all on a kernel with THP enabled?  Based on my own understanding of all of this, this will just show empty charts in that case, as all the values you're checking except `Hugepagesize` will be zero, while the `AnonHugePages`, `ShmemHugePages`, and `ShmemPmdMapped` values will instead show the relevant data.  The sysfs interface also won't show anything for THP usage.

ktsaou:
@Ferroin no I haven't (what is THP?). This PR requires `Hugepages_Total` and `Hugepagesize` to be non-zero in order to show the chart.

I saw the huge pages variables you mention, but I don't have clue what they are. Can you help?

ktsaou:
oh I got it. THP is transparent huge pages... Yet, no I haven't.

h0lzi:
Well my Systems I monitor have both. Even though it's not recommended to have both. I need the old non THP stuff though so I'm fine with this. the THP stuff is a different case I think

ktsaou:
ok, it is pretty easy to add the other metrics too, so if you can help me understand what to present and how to group them into charts, I can do it.

Ferroin:
Personally, I would handle THP and statically allocated hugepages (what this PR already covers) as separate charts, largely because how they get accounted is different.  For THP, the values get accounted as part of the regular used memory, and the stats in `/proc/meminfo` give info on how many are allocated (that is, total amount of space allocated to transparent huge pages).  `AnonHugePages` is the number used for anonymous mappings, while `ShmemHugePages` is shared mappings (and I think the `AnonPmdMapped` has something to do with how many are shared, but I'm not sure).  Both values are shown as kB, but might as well be the number of pages because they'll always be a multiple of `Hugepagesize`.

As far as exact presentation, I'm not quite sure what would be best.  I use THP on all my systems (it's actually a reasonably effective way to get back some of the performance lost by using the Meltdown mitigations in the kernel), but I've never done much as far as tracking actual usage.

ktsaou:
ok, here it is in PR #3465

PR #3463, opened 2018-02-23
 Alpine docker image improvements and hardening

ccremer:
* Now comes with fping and curl installed (there were some errors about missing curl command), fping just adds 130 kb to the image
* Installs user `uid=101(netdata) gid=101(netdata)`
* chmod/chown of files and folders

I'm not sure if switching to non-root user will cause problems in containers. It seems to be working in my case. I'm running with
```
docker run -d --cap-add SYS_PTRACE -v /proc:/host/proc:ro -v /sys:/host/sys:ro -p 19999:19999 -u netdata:netdata firehol/netdata:alpine
```

~~I'm also not sure about the USER directive in the docker file. People that derive from this image may not be able to install additional packages without root (I think the USER directive also applies to child images). But at least it's "netdata user"-ready. On the other hand, the alpine image is quite young, not many people will be using it by now.~~

I realized there is a `[global].run as user = netdata` config. I'm assuming that netdata will switch to this user if it exists, even if commented out?

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -13,17 +13,14 @@ WORKDIR /opt/netdata.git
 # Install source
 RUN chmod +x ./netdata-installer.sh && \
     sync && sleep 1 && \
-    ./netdata-installer.sh --dont-wait --dont-start-it && \
-    sync && sleep 1 && \
-# Compile fping
-    /usr/libexec/netdata/plugins.d/fping.plugin install
+    ./netdata-installer.sh --dont-wait --dont-start-it
 
 ################################################################################
 FROM alpine:edge
 
 # Reinstall some prerequisites
 RUN apk --no-cache add lm_sensors nodejs libuuid python py-mysqldb \
-                       py-psycopg2 py-yaml netcat-openbsd jq curl
+                       py-psycopg2 py-yaml netcat-openbsd jq curl fping
 
 # Copy files over
 COPY --from=builder /usr/share/netdata   /usr/share/netdata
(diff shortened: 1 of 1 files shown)


ktsaou:
I see alpine has fping available:

```sh
# apk add fping
(1/1) Installing fping (4.0-r0)
Executing busybox-1.26.2-r9.trigger
OK: 376 MiB in 99 packages
```

and that version is compatible with netdata:

```sh
# fping --help
Usage: fping [options] [targets...]

...

   -N, --netdata      output compatible for netdata (-l -Q are required)

...
```

So, no need to install it from source...

ccremer:
I wasn't aware of that^^ ofc apk is faster
fping binary from pkg was on a different location though. I moved it so that the config does not need modification. I can confirm that it works though.
When starting netdata in docker, I see
```
2018-02-25 00:00:50: netdata INFO  : MAIN : Adjusted my Out-Of-Memory (OOM) score from 0 to 1000.
2018-02-25 00:00:50: netdata ERROR : MAIN : Cannot adjust netdata scheduling policy to idle (5), with priority 0. Falling back to nice (errno 38, Function not implemented)
```
netdata is working fine AFAIK. Is that scheduling just docker or alpine stuff and kinda expected?

ktsaou:
> Is that scheduling just docker or alpine stuff and kinda expected?

It is a libc issue (musl) - they haven't implemented that call.

So, merge?

ccremer:
Yes, I think so.

ktsaou:
merged! thank you!

PR #3485, opened 2018-02-28
[updater] only restart when previously running

das7pad:
Updating netdata (e.g. headless via cron) should not start netdata in case it was not running before.


ktsaou:
thanks!

PR #3491, opened 2018-03-01
python SimpleService: auto refresh obsolete charts added

l2isbad:
requested in https://github.com/firehol/netdata/pull/3421#issuecomment-369424744


ktsaou:
ok, thanks!
(although I had the impression it was not required after all)

PR #3496, opened 2018-03-01
Fix notification DoS introduced by PR 3490.

Ferroin:
PR #3490 added the `--insecure` option to the cURL calls made by alarm-notify.sh to disable TLS certificate validation and make things work without user intervention if cURL can't find the local TLS certificate store.

However, disabling TLS certificate validation opens up a rather trivial to exploit mechanism to block notification delivery.  By hijacking the outbound HTTPS connections made by cURL for notification delivery and then scripting the responses that would be provided by the actual services cURL is trying to connect to, an attacker can persistently block the delivery of notifications through mechanisms other than e-mail and IRC.

It also allows trivial disclosure of potentially sensitive information about the state of the system including disclosing what software is running on the system and potentially certain aspects of it's configuration, which may aid an attacker.  At a minimum, this can provide more reliable feedback to an attacker that a DoS attack on one of the services provided by the system is working.

Here, we change the unconditional use of `--inescure` to be dependent on the value of a variable in `health_alarm_notify.conf`, and make that variable default to not enabling the option while putting a big warning right next to it about the security implications.  Most users should never need to disable TLS certificate validation, so this should be a reasonable default in terms of usability.

This has so far had only the bare minimum of testing, but it appears to work correctly.

In theory, the changes to `alarm-notify.sh` could be simplified because of the explicit use of bash as the shell (using `$(parameter:*word}` type expansion), but doing so would make the code less readable and would result in semantics of the option in `health_alarm_notify.conf` being a bit odd.

This PR does not currently have an associated issue in the bug tracker as I wanted to make sure I had a fix before posting about a potential security bug.

### DIFF
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -62,6 +62,22 @@ curl=""
 nc=""
 
 #------------------------------------------------------------------------------
+# extra options for external commands
+#
+# In some cases, you may need to change what options get passed to an
+# external command.  Such cases are covered here.
+
+# Control TLS certificate validation for cURL.
+# If set to yes, disable certificatie validation.
+# Otherwise, cURL will validate TLS certificates normally.
+# ******************************WARNING****************************************
+# Disabling this validation by setting the below variable to 'yes'
+# makes it trivial for someone to block notification delivery through most
+# mechanisms other than email and IRC, and may additionally leak sensitive
+# information about the system state.
+curl_insecure="no"
+
+#------------------------------------------------------------------------------
 # NOTE ABOUT RECIPIENTS
 #
 # When you define recipients (all types):
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -67,15 +67,13 @@ nc=""
 # In some cases, you may need to change what options get passed to an
 # external command.  Such cases are covered here.
 
-# Control TLS certificate validation for cURL.
-# If set to yes, disable certificatie validation.
-# Otherwise, cURL will validate TLS certificates normally.
-# ******************************WARNING****************************************
-# Disabling this validation by setting the below variable to 'yes'
-# makes it trivial for someone to block notification delivery through most
-# mechanisms other than email and IRC, and may additionally leak sensitive
-# information about the system state.
-curl_insecure="no"
+# Extra options to pass to curl.  In most cases, you shouldn't need to add anything
+# to this.  If you're having issues with HTTPS connections, you might try adding
+# '--insecure' here, but be warned that it will make it much easier for
+# third-parties to block notification delivery, and may allow disclosure
+# of potentially sensitive information.
+curl_options=""
+#curl_options="--insecure"
 
 #------------------------------------------------------------------------------
 # NOTE ABOUT RECIPIENTS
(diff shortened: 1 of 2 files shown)


ktsaou:
ok. I agree.

However, the "denial" part is not right. If there is a man-in-the-middle, then he can "deny" the notification with and without certificate validation.

The only issue that exists is the possible leak of information.
So, we have to choose among:

1. protect the notification information at all cost, even if this means that the notification will not be sent
2. make the most to dispatch the notification, even if this means the notification content itself may be leaked

When I made PR #3490, I chose 2. Considering your view (that notification content is sensitive), I now believe I was wrong. For a widely spread software solution like netdata, when we have a choice, we should pick the most secure default, even for such a tiny bit.

So, I am merging this.

conf.d/health_alarm_notify.conf
```diff
@@ -61,6 +61,22 @@ curl=""
...
+
+# Control TLS certificate validation for cURL.
+# If set to yes, disable certificatie validation.
+# Otherwise, cURL will validate TLS certificates normally.
+# ******************************WARNING****************************************
+# Disabling this validation by setting the below variable to 'yes'
+# makes it trivial for someone to block notification delivery through most
+# mechanisms other than email and IRC, and may additionally leak sensitive
+# information about the system state.
+curl_insecure="no"
```
ktsaou: I suggest to make this:

```
# curl_options="--insecure"
```

That is:

1. commented
2. ready to be uncommented

Also, the comment above about "blocking notification delivery" when certificate verification is disabled, I believe is wrong. Only the "leak" is valid.

ktsaou: Also please adapt the script accordingly for this change.

ktsaou:
I have requested a minor change to make this functionality a bit more generic. Let me know when you are done.

Ferroin:
With validation enabled, you have to get a certificate that at least looks (that is, it has to be either signed by a trusted CA, or by a certificate marked for signing that itself meets these criteria).  While it is possible to do this, it's not trivial (you're talking about shelling out in excess of 500 USD for this type of thing, not including the effort required to find someone who will sell you a forged (or otherwise fake) certificate and properly covering your tracks).

In contrast, if validation is disabled, anybody with a computer and a copy of OpenSSL can put together a certificate that will work, which means blocking notifications becomes essentially free.

I'll have an update for the way it works shortly.

Ferroin:
Updated.  I've got it as a separate commit right now, but if needed I can convert it all to be a single commit.

conf.d/health_alarm_notify.conf
```diff
@@ -61,6 +61,20 @@ curl=""
...
+#
+# In some cases, you may need to change what options get passed to an
+# external command.  Such cases are covered here.
+
+# Extra options to pass to curl.  In most cases, you shouldn't need to add anything
+# to this.  If you're having issues with HTTPS connections, you might try adding
+# '--insecure' here, but be warned that it will make it much easier for
+# third-parties to block notification delivery, and may allow disclosure
+# of potentially sensitive information.
+curl_options=""
```
ktsaou: perfect!
Please move this to the script, so that if the config is missing, the script will not inherit something from the shell.

Ferroin: Actually, the line could in theory just be completely removed.  Undefined variables expand to an empty string in all POSIX compliant shells including bash, and the default value should be an empty string.

ktsaou: ok, I'll do it.

Issue #1749, opened 2017-02-09
Password protect netdata dashboard?


bcacbe:
Is there a way to password protect the netdata dashboard?

l2isbad:
No
https://github.com/firehol/netdata/issues/70

And yes. You always can authenticate access to netdata via nginx/apache/haproxy

ktsaou:
I am closing this in favour of #70.
We already have authentication in our todo list.
Thanks!

Issue #120, opened 2016-04-01
Is there no password protection?


sankalpsingha:
I want to use this to keep a monitor on my server. Is there no way of password protecting the interface?

Currently anyone with the knowledge of my `server-ip:19999` can access the interface. 

Or am I missing something? :open_mouth:

Kubuxu:
See https://github.com/firehol/netdata/issues/101 and https://github.com/firehol/netdata/wiki/Running-behind-nginx#enable-authentication

sankalpsingha:
Yup I see that now. But I think that not every one would want to keep it behind an nginx upstream and a small authentication can be made in the application itself. Just as a basic security feature. Just my two cents..

ktsaou:
You are right. However authentication without SSL is not that much of a security.
So, SSL is needed first, but it is not my first priority.
If you really need this now, I suggest to pass netdata through another web server.

mcnewton:
We have over 150 servers and having found this project it would be nice to run it on them all. But we can't do that without any sort of protection, and we wouldn't install nginx or apache on everything. By default we'd use iptables anyway, but if that's accidentally disabled this becomes wide open. So some sort of added access control would be nice at some point.

ktsaou:
If your servers are not directly on the internet (i.e. they are in a DMZ), you can:

1. disable port 19999 at your DMZ firewall.
2. have just one nginx to proxy all netdata installations. I added info at the wiki: https://github.com/firehol/netdata/wiki/Running-behind-nginx#access-multiple-netdata-servers-via-one-nginx

I really understand this is feature is needed. Please work around it, until it is coded in netdata.

mcnewton:
We have all servers directly on the Internet (behind firewall, but there is internal access from a large number of untrusted machines, so essentially "on the Internet") :-).

But thanks for the link, it's a useful nginx config! Absolutely no pressure from me to get the feature. But something would be useful one day.

gueff:
You may be interested in this little howto, explaining how to install netdata and run it behind an apache with Basic Auth: 
https://blog.ueffing.net/post/2016/04/03/how-to-install-firehol-netdata-on-a-linux-ubuntu-14-04-box-and-run-it-behind-an-apache2-proxy-with-basic-auth-protection/

tx7:
Hi,
First, I want to say thanks to the creator of this plugin; it's awesome.
I'm currently running behind Nginx.

My question is that I've setup with SSL and password protected when I go my site:
[...]

Issue #101, opened 2016-03-31
Any way to add a http basic auth? Or password protect the page?


CodeBrauer:
Are there any built in features for that?
Some charts maybe leaks information that shouldn't be readable for anyone.

hauleth:
You could use Nginx frontend with HTTP Basic Auth to achieve that.

CodeBrauer:
@hauleth But using a webserver would only work as a proxy - on the port netdata was started, the data is still visible to anyone, isn't it?

hauleth:
If you block it in `iptables

CodeBrauer:
Okay, this should work, thanks!

CodeBrauer:
This works with Apache 2.4 and `mod_proxy`&`mod_proxy_http` enabled

```apache
<VirtualHost *:80>
    ServerName netdata.example.com
    ServerAdmin webmaster@localhost
    ProxyPreserveHost On
    ProxyRequests Off
    ProxyVia Off
    ProxyPass / http://127.0.0.1:19999/
    ProxyPassReverse / http://127.0.0.1:19999/

    ErrorLog ${APACHE_LOG_DIR}/error.log
    CustomLog ${APACHE_LOG_DIR}/access.log combined
</VirtualHost>
```

Add htaccess and htpasswd, and done.

MarjanSadat:
hi and thanks for netdata,
is it possible to have an authentication like this before seeing netdata??

![77777777](https://user-images.githubusercontent.com/31684355/48693050-6fc8b080-ebed-11e8-9123-4ed71ccd6241.png)

PR #3501, opened 2018-03-02
Add support for HTTP basic authentication

FooBarWidget:
This pull request adds builtin support for HTTP basic authentication.

The main use case for HTTP basic authentication support built directly into Netdata is to protect the master <-> slave stream endpoint. HTTP basic authentication prevents slaves from querying data from the master.

Netdata can be put behind an Nginx reverse proxy which performs HTTP basic authentication, but the master <-> slave stream endpoint is not real HTTP and so cannot be reverse proxied behind Nginx.

See also this discussion (starting 25 Feb 2018): https://github.com/firehol/netdata/issues/2043

Basic authentication can be activated by setting netdata.conf, "[web]" section, "basic authentication = username:password". Not setting it or setting it to the empty value means that HTTP basic authentication is disabled.

Special care has been taken to ensure that the string comparison function is not vulnerable to timing attacks, by not early terminating.

HTTP basic authentication applies to all HTTP verbs except STREAM.


FooBarWidget:
People in issues #101, #70, #120, #1749 and #732 may be interested in this pull request.

Ferroin:
Just from a cursory look, the implementation appears to be sound (and it's far cleaner than what I would have come up with), but I don't see much value in this right now because Netdata does not support serving the dashboard over TLS.  Unless you've got some form of encryption at the transport level, you're just tossing plain-text authentication tokens (with no replay or reuse protection and no actual verification) across the wire, which provides exactly zero benefit to security because they're trivial for an attacker to grab and reuse (and in fact can be a bad thing for security because it can make you think things are more secure than they actually are).

FooBarWidget:
> Unless you've got some form of encryption at the transport level

I do. I am using stunnel. See https://github.com/firehol/netdata/issues/2043#issuecomment-368562569.

Ferroin:
My point was about netdata in general, not any specific setup.  I'd actually argue that from the master<->slave side of things it might be better to just make that real-enough HTTP to proxy it with NGinx (and then you would be able to filter actual endpoints and limit on specific methods, which is arguably much more secure than using passwords), but I have a feeling that @ktsaou probably feels differently (and I can't really blame him if that's the case, such a change would be a hard break in the protocol).

I'm not against this change myself (Who knows, maybe having this will be enough incentive for someone to actually implement native TLS support in netdata?), but I personally think there should be a prominent warning in place about not messing with this unless you're already encapsulating the connection over a properly secured channel.

FooBarWidget:
Agreed, I also think it is better if the stream endpoint can be proxied through Nginx because it's more flexible that way.

Also agreed about the warning.

FooBarWidget:
Can you make a decision with regards to this? I honestly don't care whether it's going to be HTTP basic authentication or making the stream endpoint real HTTP, I just want to achieve my security goal.

Ferroin:
Strictly speaking, it's @ktsaou or possibly @l2isbad who would need to make such a decision.  As stated before, I personally think that changing things to make the streaming protocol HTTP compliant enough to proxy it through nginx is the better approach, but just adding HTTP Basic Authentication is probably more likely to get merged in a timely fashion since it's less likely to break things.

ktsaou:
this is a nice PR! Thank you!

But, I think it cannot be merged as-is. The main problem is that we need a distributed authentication mechanism. netdata is a distributed app and I think we should invest in scaling out as much as possible.

Give me a few days to come back with proper reasoning and probably a plan to do this right.

Regarding the original issue, of forbidding netdata slaves from querying the master, there is already an option to deny access based on IP:

```
[web]
  allow dashboard from = *
```

You can add negative expressions, like:

```
[web]
  allow dashboard from = !172.16.* *
```

So, if the slaves are at 172.16.*, they will be denied access to the dashboard (and the charts). Everyone else will be accepted.

FooBarWidget:
> this is a nice PR! Thank you!
> But, I think it cannot be merged as-is. The main problem is that we need a distributed authentication mechanism. netdata is a distributed app and I think we should invest in scaling out as much as possible.
> Give me a few days to come back with proper reasoning and probably a plan to do this right.

Thanks, I look forward to your thoughts.

> Regarding the original issue, of forbidding netdata slaves from querying the master, there is already an option to deny access based on IP:

Unfortunately this does not work for me since I am using stunnel to secure the master <-> slave connection. From the point of view of the master, all connections come from 127.0.0.1 since that's where the stunnel exits.

ktsaou:
> Unfortunately this does not work for me since I am using stunnel to secure the master <-> slave connection. From the point of view of the master, all connections come from 127.0.0.1 since that's where the stunnel exits.

btw, you have to try `gvpe`. This it magic...  a mesh VPN, a virtual ethernet that connects all your servers. I have a wrapper to simplify the setup: https://github.com/firehol/netdata-demo-site/tree/master/gvpe

Ferroin:
[tinc](https://www.tinc-vpn.org/) might also be worth looking into as an alternative to stunnel.  I've not used gvpe myself, so i can't really compare it to that, but I have tested tinc in comparison to stunnel, and the performance is about the same (there's marginally more overhead on-the wire for tinc, but about the same resource usage on each node for either), but it does give you real IP's you can filter on.

l2isbad:
tinc is super great :+1: :+1: :+1:

ktsaou:
tinc has a different purpose. The technology underneath tinc and gvpe is almost identical (and actually gvpe uses tinc interfaces code), but:

1. tinc is intended to provide a scaleable private **ad-hoc** mesh network. So, it supports invitations, dynamic routing, etc. This is "adaptable" and hosts can join any time without re-configuration. Tinc has been designed to provide a private internet, over the public internet, that users can join any time without re-configuring the whole VPN.

2. gvpe, on the other hand, was designed to merge physical LANs (it can transparently merge 2 or more physical LANs into one, including parts of the layer-2 communication). It is "pre-configured" and "fixed", there is no adaptability at all (apart from some routing fallback). It is also extremely secure (members of the network cannot decrypt the traffic when this traffic is not for them, even when it passes through them).

To my understanding gvpe is the ultimate cloud virtual LAN tool. The fact that it requires reconfiguration to add more members may seem like a weakness, but it actually is an important feature. It allows creating a VPN without any kind of back-door.

FooBarWidget:
Has a decision been made yet? I've been postponing deploying netdata all this time.

ktsaou:
I am very sorry for the delay. I have a personal issue these days. I hope it will be solved in less than a week and I'll be back up to speed...

ktsaou:
btw, I really like this: https://www.joyfulbikeshedding.com/blog/2018-03-22-netdata-simple-server-monitoring.html

I'll look on your points about security in detail...

ktsaou:
ok, here I am.

To add authentication to netdata, it is important to do it in a way that will promote its distributed philosophy. Users that make occasional use of netdata may be satisfied with just a simple auth prompt at a server, but I think this is not sufficient. We should focus to all those that really want to manage a large infrastructure with netdata.

netdata is distributed by nature. This is its key strength. But it is also a major weakness at the same time. Strength because it allows collecting thousands of metrics in real-time - something unique and pretty impossible (due to its cost) to do it at scale with any other solution. Weakness because it is somewhat tricky to provide a few simple features, like authentication.

Whatever we do with authentication, we have to allow these scenarios to work:

1. multi-user - sharing just one password is a no go. I would love to support PAM, so that folks that have a proper central user management system (LDAP or NIS) will be able to manage access to all their netdata centrally.

2. custom dashboards with charts coming from multiple netdata server should work with single sign on. So, users will login once for all the netdata servers that contribute charts to a custom dashboard.

3. custom dashboards without authentication should also be possible. For example, I have many custom dashboards that monitor specific functions for specific teams or customers. These dashboards are already hosted in a protected environment, so I would probably need specific API keys/tokens that would allow bypassing user authentication for the specific referrer and for specific charts. Of course this is outside the scope of this PR. I am only mentioning it here because it also requires some centralization (i.e. API tokens should be centrally managed for the whole infrastructure).

4. The nature of netdata is such that could allow a more relaxed authentication. I am a big fan of the authentication method used by slack for years (i.e. login by typing your email, receive an email, click a link found in it, da-da you are logged in). No passwords. Then we only need to whitelist the emails that should have access to our servers.

  If this is too relaxed for some, we could add authenticator app support. So, verify your email and provide an OTP. Still no password. Nothing to steal.

  Of course there are more distributed and more secure authentication methods. The most widely used is OAUTH2.  So, we could login with google, microsoft, or github (I wouldn't support social networks, although even this is possibe). Then on our netdata we will whitelist the accounts we want to have access.

Using any of the above distributed methods there is still a question: where this authentication should take place? On each netdata?

I think not.

Each netdata should just trust a central entity (like the registry), that will take care of the following:

1. central configuration of the users allowed
2. central configuration of the API keys/tokens
3. single sign-on for all our netdata servers

So, I strongly believe we should come with a simple protocol for each netdata to trust a central place and then integrate authentication to this central place.

Since you are fluent in C, I would love to work with you for this cause.
What do you think?

FooBarWidget:
Sorry for the late reply too.

I think it is a good idea. It will definitely help scenarios at scale.

However, I do not have the time to commit to helping with such a big vision.

ktsaou:
ok,

@FooBarWidget do I have your permission to use pieces of your code when I will implement this.

kachkaev:
Hi @ktsaou 👋  Are you still interested in looking at this?

aadityabagga:
@kachkaev, you can bind netdata on localhost and view it via apache proxy with http basic auth (this is what i do).

Details here: https://github.com/firehol/netdata/wiki/Running-behind-apache

kachkaev:
Thanks for the link @aadityabagga. Yeah I know it is possible to spin up a reverse proxy in front of netdata and it would do the auth, I'm just wondering if this extra element can be replaced by a simple built-in feature.

IMO the best ‘selling point’ of netdata is that apart from Prometheus + Grafana, this tool can be kicked off on a new server in seconds by a newbies like me. Simple HTTP auth in `netdata.conf` looks like a very useful feature for a person who has just created their first ever VM to host a simple website. Setting up reverse proxies, firewalls, DMZs and so is certainly better, but this the next level of complexity is what you would want to avoid at the very start of the sysadmin journey.

aadityabagga:
@kachkaev got you.

In that case, it can be run without authentication!
(in my opinion)

cakrit:
What @ktsaou described in this PR is already in the roadmap for [netdata.cloud](https://blog.netdata.cloud/posts/netdata-cloud-announcement/). Closing this one, as the implementation is/will be quite different.

PR #3543, opened 2018-03-13
Missing commas in elasticsearch.chart.py.

NeonSludge:
This pull request fixes two Elasticsearch charts that were not receving data:

- bulk threadpool queue chart
- total flushes chart


ktsaou:
thank you!

PR #3544, opened 2018-03-13
New charts for Elasticsearch

NeonSludge:
This pull request adds several charts for Elasticsearch translog/indices segments statistics and JVM buffer pools utilization.
These metrics are often very helpful when evaluating an Elasticsearch node health.

### DIFF
--- a/python.d/elasticsearch.chart.py
+++ b/python.d/elasticsearch.chart.py
@@ -34,12 +34,33 @@ NODE_STATS = [
     'indices.refresh.total_time_in_millis',
     'indices.flush.total',
     'indices.flush.total_time_in_millis',
+    'indices.translog.operations',
+    'indices.translog.size_in_bytes',
+    'indices.translog.uncommitted_operations',
+    'indices.translog.uncommitted_size_in_bytes',
+    'indices.segments.count',
+    'indices.segments.memory_in_bytes',
+    'indices.segments.terms_memory_in_bytes',
+    'indices.segments.stored_fields_memory_in_bytes',
+    'indices.segments.term_vectors_memory_in_bytes',
+    'indices.segments.norms_memory_in_bytes',
+    'indices.segments.points_memory_in_bytes',
+    'indices.segments.doc_values_memory_in_bytes',
+    'indices.segments.index_writer_memory_in_bytes',
+    'indices.segments.version_map_memory_in_bytes',
+    'indices.segments.fixed_bit_set_memory_in_bytes',
     'jvm.gc.collectors.young.collection_count',
     'jvm.gc.collectors.old.collection_count',
     'jvm.gc.collectors.young.collection_time_in_millis',
     'jvm.gc.collectors.old.collection_time_in_millis',
     'jvm.mem.heap_used_percent',
     'jvm.mem.heap_committed_in_bytes',
+    'jvm.buffer_pools.direct.count',
+    'jvm.buffer_pools.direct.used_in_bytes',
+    'jvm.buffer_pools.direct.total_capacity_in_bytes',
+    'jvm.buffer_pools.mapped.count',
+    'jvm.buffer_pools.mapped.used_in_bytes',
+    'jvm.buffer_pools.mapped.total_capacity_in_bytes',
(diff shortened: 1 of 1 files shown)


python.d/elasticsearch.chart.py
```diff
@@ -166,13 +189,69 @@
...
+            ['indices_segments_memory_in_bytes', 'total', 'absolute', 1, 1048567],
+            ['indices_segments_terms_memory_in_bytes', 'terms', 'absolute', 1, 1048567],
+            ['indices_segments_stored_fields_memory_in_bytes', 'stored_fields', 'absolute', 1, 1048567],
+            ['indices_segments_term_vectors_memory_in_bytes', 'term_vectors', 'absolute', 1, 1048567],
+            ['indices_segments_norms_memory_in_bytes', 'norms', 'absolute', 1, 1048567],
+            ['indices_segments_points_memory_in_bytes', 'points', 'absolute', 1, 1048567],
+            ['indices_segments_doc_values_memory_in_bytes', 'doc_values', 'absolute', 1, 1048567],
+            ['indices_segments_index_writer_memory_in_bytes', 'index_writer', 'absolute', 1, 1048567],
+            ['indices_segments_version_map_memory_in_bytes', 'version_map', 'absolute', 1, 1048567],
+            ['indices_segments_fixed_bit_set_memory_in_bytes', 'fixed_bit_set', 'absolute', 1, 1048567]
```
l2isbad: Please replace underscores with spaces in dimension names
like `fixed_bit_set` => `fixed bit set`

python.d/elasticsearch.chart.py
```diff
@@ -166,13 +189,69 @@
...
+        ]},
     'jvm_mem_heap': {
         'options': [None, 'JVM Heap Currently in Use/Committed', 'percent/MB', 'memory usage and gc',
                     'elastic.jvm_heap', 'area'],
         'lines': [
             ['jvm_mem_heap_used_percent', 'inuse', 'absolute'],
             ['jvm_mem_heap_committed_in_bytes', 'commit', 'absolute', -1, 1048576]
```
l2isbad: Please split this chart into two.
Dimensions has different units..

python.d/elasticsearch.chart.py
NeonSludge: Done.

python.d/elasticsearch.chart.py
NeonSludge: Created a new chart for JVM usage represented in bytes.

python.d/elasticsearch.chart.py
```diff
@@ -166,12 +190,74 @@
...
+        'options': [None, 'Total Number Of Indices Segments', 'count', 'indices segments',
+                    'elastic.index_segments_count', 'line'],
+        'lines': [
+            ['indices_segments_count', 'segments', 'absolute']
+        ]},
+    'index_segments_memory': {
+        'options': [None, 'Indices Segments Memory Usage', 'MB', 'indices segments',
+                    'elastic.index_segments_memory', 'area'],
+        'lines': [
+            ['indices_segments_memory_in_bytes', 'total', 'absolute', 1, 1048567],
```
l2isbad: about this
i suggest
1. remove `total` from `index_segments_memory` chart
2. if total is needed as chart please add another one

about `index_segments_memory` are you sure about `area`? maybe `stacked`?

NeonSludge: Yes, this part definitely needed some adjustments.
It appears that the most logical way to do this would be: 
1. Create an `index_segments_memory` stacked chart that has every metric from `indices.segments.` except for `indices.segments.index_writer_memory_in_bytes`. This is the total amount of memory needed to support all existing segments.
2. Create a separate chart with just `indices.segments.index_writer_memory_in_bytes`: this shows the amount of data held in Lucene's `IndexWriter` and not yet written to new on-disk segments.

This way the `total` metric is unnecessary.
I've pushed the necessary changes.

l2isbad:
@NeonSludge gj :+1:

ktsaou:
merged. thank you!

PR #3549, opened 2018-03-14
Pagerduty enhancement to avoid incident duplication.

tarantegui:
We have 2 netdata masters (for high availability) sending alarms through pagerduty. With this enhancement we can avoid alarm duplication.

Feedback is appreciated!

### DIFF
--- a/plugins.d/alarm-notify.sh
+++ b/plugins.d/alarm-notify.sh
@@ -1008,7 +1008,7 @@ send_pd() {
             ${pd_send} -k ${PD_SERVICE_KEY} \
                        -t ${t} \
                        -d "${d}" \
-                       -i ${alarm_id} \
+                       -i ${host}:${chart}:${name} \
                        -f 'info'="${info}" \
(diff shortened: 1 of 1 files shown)


ktsaou:
nice!

I don't use pagerduty myself. Anyone using it? I don't know if this will break anything, so some feedback is need before merging it.

@HackToHell, @jimcooley have a look...

jimcooley:
That change looks fine to me, especially given the scenario you describe where the same alarm gets a different alarm id from two different masters.  I couldn't think of a scenario where `${host}:${chart}:${name}` could identify more than one alarm.

@ktsaou - We're still heavily using netdata with pagerduty alarms.

This looks good to me.  I approve.

jimcooley:
Additional thoughts:
If #2976 moves forward, the same change will be needed there.
If we start using the [v2 events api](https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2), then `${host}:${chart}:${name}` should become the `dedup_key `.

PR #3558, opened 2018-03-19
add response_time histogram for web_log

Wing924:
Histogram is very useful, especially in backend. We can use histogram to calc φ-quantiles, define SLA, etc.
[HISTOGRAMS AND SUMMARIES (prometheus)](https://prometheus.io/docs/practices/histograms/)
[Cumulative histogram](https://en.wikipedia.org/wiki/Histogram#Cumulative_histogram)
## Charts

Because dashboard don't support histogram chart currently, I choose line chart. 
Line chart is not good for present histogram.

## complexity
* Time complexity: `O(nk)`
* Memory complexity: `O(k)`
where `n` is count of log lines, `k` is count of histgram buckets.


python.d/web_log.chart.py
```diff
@@ -347,8 +358,30 @@ def create_web_charts(self, match_dict):
...
         if 'resp_time' not in match_dict:
             self.order.remove('response_time')
+            self.order.remove('response_time_hist')
         if 'resp_time_upstream' not in match_dict:
             self.order.remove('response_time_upstream')
+            self.order.remove('response_time_upstream_hist')
+
+        # Add 'response_time_hist' and 'response_time_upstream_hist' charts if is specified in the configuration
+        histogram = self.configuration.get('histogram', None)
+        if type(histogram) is list:
```
l2isbad: please use 
```python
isinstance(value, list)
```

python.d/web_log.chart.py
```diff
@@ -347,8 +358,30 @@ def create_web_charts(self, match_dict):
...
+            upstream_hist_lines = self.definitions['response_time_upstream_hist']['lines']
+            for i, le in enumerate(histogram):
+                hist_key = "response_time_hist_%d" % i
+                upstream_hist_key = "response_time_upstream_hist_%d" % i
+                hist_lines.append([hist_key, str(le), 'incremental', 1, 1])
+                upstream_hist_lines.append([upstream_hist_key, str(le), 'incremental', 1, 1])
+
+            hist_lines.append(["response_time_hist_%d" % len(histogram), '+Inf', 'incremental', 1, 1])
+            upstream_hist_lines.append(["response_time_upstream_hist_%d" % len(histogram), '+Inf', 'incremental', 1, 1])
+        elif histogram is not None:
```
l2isbad:
```python
elif not histogram:
```

python.d/web_log.chart.py
```diff
@@ -347,8 +358,30 @@ def create_web_charts(self, match_dict):
...
+            self.error("expect histogram list, but was {}".format(type(histogram)))
```
l2isbad:
```python
but was {0}"
```

python.d/web_log.chart.py
```diff
@@ -903,6 +951,18 @@ def get_timings(timings, time):
...
     timings['count'] += 1
 
+def get_hist(index, buckets, time):
+    """
+    :param index:   histogram index (Ex. [10, 50, 100, 150, ...])
+    :param buckets: histogram buckets
+    :param time:    time
+    :return: None
+    """
+    for i in xrange(len(index)-1, -1, -1):
```
l2isbad: use `range`

ktsaou:
merged! thank you!

PR #3566, opened 2018-03-21
Add disk inode usage rate health alarms (like disk fill rate)

amichelic:
I noticed that health alarms for inode usage rate were missing, so I added them (that's copy'n'paste from the disk fill rate block with some changes).


ktsaou:
nice! thank you!

PR #3577, opened 2018-03-26
Update freebsd_getifaddrs.c for system.net

RicardoSette:
Suggestion of code to validate if a interface is physical in FreeBSD, unfortunately is not easy as Linux, for FreeBSD there is no flag to indicate that the interface is physical.

We had some "false positives" with optical interfaces, often because of GBICs not homologated for FreeBSD, so we started to control physical interfaces in NETDATA using CONFIG_SECTION_GETIFADDRS, much simpler and much more accurate.

We added in the section the "set physical interfaces for system.net" attribute where DEFAULT_PHYSICAL_INTERFACES is "igb* ix* cxl* in* ixl* ixlv* bge* ixgbe*".

Now the identification of physical interfaces is controlled by the settings, avoiding other "false positive".

### DIFF
--- a/src/freebsd_getifaddrs.c
+++ b/src/freebsd_getifaddrs.c
@@ -141,19 +141,22 @@ static struct cgroup_network_interface *get_network_interface(const char *name) 
 int do_getifaddrs(int update_every, usec_t dt) {
     (void)dt;
 
-#define DELAULT_EXLUDED_INTERFACES "lo*"
+#define DEFAULT_EXCLUDED_INTERFACES "lo*"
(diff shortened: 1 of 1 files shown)


ktsaou:
nice work! give me 1-2 more days to test it a bit.

RicardoSette:
Hi @ktsaou ,

Added feature system.packages and updated system.net to FreeBSD. Discard pull 3577 and use 3893

https://github.com/firehol/netdata/pull/3893

ktsaou:
Nice!
Please also sign the CLA by adding your name to CONTRIBUTORS.md
If this works for you, I'll merge it. It seems good for me.

PR #3584, opened 2018-03-28
Minor updates to the BTRFS chart help text.

Ferroin:
* Corrected a typo in the `btrfs.system` chart description.
* Added a some info to the `btrfs.disk` chart description to explain that not all the space listed as `unallocated` may be usable.  The exact reason for this is somewhat long to explain (in short, it's possible to get a combination of device sizes and chunk profiles that make it impossible to use 100% of the space on all the devices in all cases.  For example, a 3-device setup using the `raid1` profile will usually not be able to use 100% of the space on all the devices unless the devices are sized correctly (two devices have to be the same size and the third has to be equal to twice their size minus the size of the system chunk) _and_ added to the volume in a specific order (you have to start with the two smaller ones to force the system chunk to be on them and then add the larger one).


web/dashboard_info.js
```diff
@@ -1905,7 +1905,7 @@ netdataDashboard.context = {
...
+        info: 'Physical disk usage of BTRFS. The disk space reported here is the raw physical disk space assigned to the BTRFS volume (i.e. <b>before any RAID levels</b>). BTRFS uses a two-stage allocator, first allocating large regions of disk space for one type of block (data, metadata, or system), and then using a regular block allocator inside those regions. <code>unallocated</code> is the physical disk space that is not allocated yet and is available to become data, metdata or system on demand. When <code>unallocated</code> is zero, all available disk space has been allocated to a specific function. Healthy volumes should ideally have at least five percent of their total space <code>unallocated</code>. You can keep your volume healthy by running the <code>btrfs balance</code> command on it regularly (check <code>man btrfs-balance</code> for more info).  Note that some of the spac elisted as `unallocated` may not actually be usable if the volume uses devices of different sizes.'
```
ktsaou: thanks!

Please change the markdown code quotes to `<code></code>`.

Ferroin: @ktsaou Updated as requested, sorry I forgot about needing actual HTML elements instead of MD (I'm apparently too used to using MD given that pretty much everything I use for chat uses it).

ktsaou:
thank you!

PR #3599, opened 2018-04-03
Add syslog support to alarm-notify.sh.

Ferroin:
This uses the `logger` command (found on all modern Linux and BSD systems)
to allow `alarm-notify.sh` to send netdata alarm messages to a local
or remote syslog server.  By default, logging is done to the local6
facility at a warning log level, with critical alarms being logged at
the err level.

This provides a mechanism of easily generating a running log of all
netdata alarm events on most standard UNIX systems, and when used with
full-featured logging daemons like rsyslog or syslong-ng can be used to
archive such messages in a wide variety of alternative formats.

There are a few limitations currently:

 * It can only log to one location.  This can be easily worked around in
 most syslog daemon's configurations, as almost all of them support some
 form of output multiplexing.
 * The default settings will usually not get the results most people
 would want, but because of the sometimes very significant differences
 in default syslog configurations from distribution to distribution,
 there's not really any way to make the defaults any saner.
 * While it is possible to log directly to a remote syslog server, there
 is no quick and easy wet up for this, it requires the user to manually
 compose the extra options that need to be passed to `logger`.

Syslog messages generated by this will look something like this:

    Apr  3 09:00:30 localhost netdata: Netdata notification on wild-karde at Tue Apr 3 09:00:00 EDT 2018: WARNING, out of disk space time = 5h

Lightly tested on Linux, no testing on BSD yet but based on FreeBSD documentation it should work there as-is.

As mentioned in the limitations above, the default settings are almost certainly not what most people will want (depending on the distribution, you'll either get all the notifications in the main syslog log file (`/var/log/syslog` or `/var/log/messages` in most cases), or you won't get any at all), so it won't send messages unless explicitly enabled (which is extra important as the necessary command is present in the default install of all major Linux distributions, as well as FreeBSD and it's derivatives).

I've been using a custom notification mechanism similar to this for a while now on my VPS nodes so that I have an easy way to correlate netdata alarms with other log entries (for example, making it easier to correlate data collection alarms for my web servers with specific sets of requests).   I've found this to be extremely useful on a couple of occasions, and would much rather not need to go through custom alert notifiers for it, and I expect other people may find it useful too.

### DIFF
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -15,6 +15,7 @@
 # - sms messages to your cell phone or any sms enabled device (messagebird.com)
 # - notifications to users on pagerduty.com
 # - messages to your irc channel on your selected network
+# - messages to a local or remote syslog daemon
 #
 # The 'to' line given at netdata alarms defines a *role*, so that many
 # people can be notified for each role.
@@ -61,6 +62,11 @@ curl=""
 # If not found, irc notifications will be silently disabled.
 nc=""
 
+# The full path of the logger command.
+# If empty, the system $PATH will be searched for it.
+# If not found, syslog notifications will be silently disabled.
+logger=""
+
 #------------------------------------------------------------------------------
 # extra options for external commands
 #
@@ -73,6 +79,13 @@ nc=""
 # third-parties to block notification delivery, and may allow disclosure
 # of potentially sensitive information.
 #curl_options="--insecure"
+
+# Extra options to pass to logger.  If you just want to send to your
+# local system logs, you can leave this empty and everything should work.
+# If you want to send to a remote syslog server, you need to specify the
+# remote server here with the `--host` option, and may need additional
+# options to select the protocol.  See `man logger` for more information.
+#logger_options=""
 
 #------------------------------------------------------------------------------
 # NOTE ABOUT RECIPIENTS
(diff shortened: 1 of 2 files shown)


l2isbad:
> It can only log to one location.

sounds like no go.

@Ferroin What about to write a python proxy to eliminate limitations of invoking `logger` directly from shell?

Ferroin:
Also, I do plan to add a wiki page once this gets merged to give some explanation of the whole log level and facility thing, an example of how to send direct to a remote syslog server, and some example configurations for the big three syslog implementations (classic (BSD and most other UNIX systems), rsyslog, and syslog-ng) covering some of the more likely use cases, namely:

 1. Log all the messages from netdata to their own file but nowhere else.
 2. Forward all the messages from netdata to a remote syslog server.
 3. 1 and 2 combined (log locally to a special file, and forward to a remote server too).
 4. Displaying notifications on the console for alerts.
 5. Explicitly forward messages to the systemd journal (using rsyslog).

ktsaou:
hm... do you want me to syslog alarms from netdata? I can do this and the alarms can be logged before applying the delays notifications get (i.e real-time).

I can add a few netdata.conf settings to allow some configuration...

What do you think?

ktsaou:
Well... generally I think we should move (most of) alarm-notify.sh inside netdata. The problem is that script won't be run if the system is not capable of running anything, but netdata is still running and able to dispatch syslog events.

I agree though it is simpler in `alarm-notify.sh`.

What would be interesting is to allow different notification roles to generate different syslog entries. This could be possible with minimal changes to this PR...

Ferroin:
I can look at adding that, though it may be next week before I have anything to push out that's been tested.

In theory, if we _really_ want to handle dispatch to multiple syslog daemons, that shouldn't be too hard either (it could easily be per role too), but I'll wait on that until there's some consensus here about whether we want it or not.  I'm still of the opinion that we should leave the multiplexing to the syslog daemon itself, as that is the standard location for configuring things like relaying.

ktsaou:
hm... I don't know about multiple syslog servers, but it would be nice to have a different prefix text and possibly a different facility or priority. Like having recipients in this format:

```
facility[.level][@server[:port]][:prefix]
```

Where:

- `facility` is one of `auth`, `authpriv`, `cron`, `daemon`, `ftp`, `kern`, `local0` to `local7`, `lpr`, `mail`, `news`, `syslog`, `user`, `uucp`, `security`.

- `level` is one of `emerg`, `alert`, `crit`, `err`, `warning`, `notice`, `info`, `debug`, `panic`, `error`, `warn`. Default should `warn` for warnings, `crit` for critical and `info` for clear notifications (so the default is dynamic).

- `server` can be any IPv4 or IPv6 IP or a hostname. If it is IPv6 it should be enclosed in brackets `[xx:xx:...:xx]`.

- `port` is any port (default 514)

- `prefix` is any single word (without spaces) to prefix the message

Ferroin:
I like the general design, with one further limitation

Instead of allowing specifying just facility or just log level, it has to be both or neither.  This means we can just pass the whole string to the left of the @ (or the / if there's no server specified) to `logger`, and therefore avoid having to validate it in the script.  Punting the validation to `logger` helps with practical compatibility, as there are some UNIX systems that use different names for some of the levels and facilities.

I can get to work on this, but as mentioned it may be next week before I have anything.

Ferroin:
OK, I just pushed an updated version that uses the role recipient variables to handle configuration using the format suggested by @calmh.  It's not particularly efficient (but parsing complex strings in shell script never is), and it's only been very lightly tested (I've verified that the parsing works correctly for a bunch of expected configurations, and that it still logs correctly to the local syslog daemon), but it appears to work correctly.

In addition to this, there's a couple of other changes bundled in:

* The log message format has been updated.  The message itself will look like:

    `prefix WARNING on wild-karde at Tue Apr 3 09:00:00 EDT 2018: out of disk space time = 5h`

   This of course still needs to be updated a bit to include the chart info, but I'm not quite sure what the best place for it would be.
* I've removed the options to configure the log level globally in favor of the user configuring it per-recipient.  I don't envision many cases where different dynamic levels will be needed.

Beyond the issue with the log message, the only other things I can think of that might be needed are making it more efficient (an in-line Python program for the parsing would probably run faster, but would also look pretty out-of-place), and possibly the option to use an ISO 8601 timestamp in place of the RFC 1123 date, though that may be better as a separate PR as I'd like to have that option for emails too.

calmh:
Wat

Ferroin:
Oh, whoops, wrong mention, sory @calhm, meant @ktsaou...

ktsaou:
@Ferroin I need a couple of days more to check this.
Sorry for the delay...

Ferroin:
@ktsaou Other than review (and the aforementioned decision about the log format, which I would love some input from you and/or @l2isbad about), is there anything else that needs done here?

Ferroin:
I've updated this to put the chart name after the date, but just bef0re the formatted alarm value, as that seems to be the most sensible place for it.  If something else is desired, just let me know.

As far as the date formatting I mentioned earlier, I'm just going to do a separate PR for that once this gets merged.

ktsaou:
I am good to merge this.
@Ferroin I think you have done a excellent job! (as always).

Let's see what people would think of it...

So merge?

Ferroin:
I assume you would like a page added to the Wiki about how to configure this once it's merged given that the config is somewhat different from most other notification channels?

ktsaou:
> I assume you would like a page added to the Wiki about how to configure this once it's merged given that the config is somewhat different from most other notification channels?

yes please, just like the other notification methods have...

Ferroin:
This should be good to go too.

ktsaou:
perfect! thank you!

PR #3601, opened 2018-04-03
update openrc init script

prometheanfire:
quote variables where able
switch to more posix double and
remove unused variables

### DIFF
--- a/system/netdata-openrc.in
+++ b/system/netdata-openrc.in
@@ -3,82 +3,55 @@
 # The user netdata is configured to run as.
 # If you edit its configuration file to set a different
 # user, set it here too, to have its files switch ownership
-: ${NETDATA_OWNER:=netdata:netdata}
-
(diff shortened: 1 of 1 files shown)


prometheanfire:
tested this, it's much more openrcified

ktsaou:
Thanks!

@candrews I think this is ok. Do you have any comments?

Ferroin:
Not exactly the person you mentioned, but I will comment that from the perspective of a Gentoo user who has had to deal with his share of init scripts, this does look a lot nicer and is much more consistent in terms of style with most of the scripts shipped as part of OpenRC.

prometheanfire:
I did run this by him on irc.

    12:38 <           candrews@> prometheanfire, thanks! I don't use openrc, so this improvement is much appreciated

ktsaou:
merged!

PR #3607, opened 2018-04-05
Use new API for IPFS

tancou:
Optimize request number and patch json parsing.

Getting number of pinned files could timeout  when there are too many pinned files.


l2isbad:
oops, forgot about it

ktsaou:
merged it.
thank you!

PR #3621, opened 2018-04-10
postgres module: TypeError fix in query_stats_

l2isbad:
fixes https://github.com/firehol/netdata/issues/3614 plus some identation fixes


ktsaou:
@l2isbad you rock!

PR #3633, opened 2018-04-16
statsd improvements

ktsaou:
1. statsd sets now show zeros when no data are collected
2. statsd histograms and timers were wasting some resources due to unnecessary mutex access on every flush - so when there are many thousands of  histograms and timers there was increased CPU usage of the charting thread
3. several minor optimizations have been applied to lower the CPU usage of the charting thread when there are several millions of statsd metrics collected.
4. When there are several millions of statsd metrics, the charting thread used constant but quite high CPU. The reason was that each second it needed to loop through all the metrics. Now, netdata statsd tracks which statsd metrics are charted (i.e. they are useful) and completely disables all its functions for the not useful ones (i.e. it does not actually parse the collected values and it does not flush them).

### DIFF
--- a/src/statsd.c
+++ b/src/statsd.c
@@ -1649,14 +1649,14 @@ static inline void statsd_flush_gauge(STATSD_METRIC *m) {
     debug(D_STATSD, "flushing gauge metric '%s'", m->name);
 
     int updated = 0;
-    if(m->count && !m->reset) {
+    if(unlikely(!m->reset && m->count)) {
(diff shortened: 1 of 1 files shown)


ktsaou:
ok, fixed the issue with the millions of not-useful statsd metrics too. netdata statsd now loops only through the useful charts (i.e. the ones that appear in charts).

ktsaou:
tested and seems good. merging.

PR #3645, opened 2018-04-22
UrlService decode fix

l2isbad:
fixes https://github.com/firehol/netdata/issues/3641
@ktsaou 
should work on `py2` and `py3` after the fix


ktsaou:
ok, let's see.
Thank you!

PR #3664, opened 2018-04-27
Add a plugin for Spigot Minecraft server stats.

Ferroin:
This adds a plugin which tracks rudimentary stats from a Spigot
Minecraft server using the remote console protocol.  It provides charts
for the number of currently connected users, as well as the number of
ticks the server is processing each second (which correlates strongly
with server-side lag).

The plugin makes use of a Python implementation of the remote console
protocol originally written by Barnaby Gale and released under an MIT
license.  This bundles the module with netdata as it's tiny (less than
100 lines of code) and is not packaged in any distributions I know of.

The default settings are to connect to the local system on the standard
remote console port (25575) without a password, which allows for trivial
setup (users just need to set `enable-rcon` to true in the
server.properties file for their Spigot server).

Currently, this plugin uses the SimpleService framework instead of the
SocketService plugin, which actually makes implementation marginally
simpler.  In theory, it could probably be re-implemented using the
SocketService, but I very much doubt it would be any more efficient than
it is now (we already keep the socket open and connected persistently so
we don't have to re-authenticate, and we make the protocol synchronous
so it's trivial to determine that we've recieved everything because a
read returns no new data).

Example image from an idle server:
image

The 'Minecraft Users' chart is pretty self explanatory, it literally just tracks how many users are online (which requires far more text processing than it should).

The 'SpigotMC Ticks Per Second' chart takes a bit more explanation.  Minecraft natively processes events at a rate of 20Hz.  Each cycle of processing is called a 'tick'.  Heavily loaded servers may however take longer than the requisite 50ms to process one 'tick', in which case you get the in-game equivalent of time-dilation (the server never drops ticks to catch up, it just starts processing the next one immediately if the previous tick took more than 50ms, so in-game time runs slower than real-time).  Spigot provides a command for retrieving running averages of how many ticks were processed in a second over the past 1, 5 and 15 minutes.  Lower values on this chart indicate higher levels of lag in the server itself (usually because too many things are happening at once for it to keep up).  Typical values for a normal server that has sufficiently good hardware are between 20.00 and 19.95.

Lightly tested on my own (mostly unutilized) server.

### DIFF
--- a/python.d/python_modules/third_party/mcrcon.py
+++ b/python.d/python_modules/third_party/mcrcon.py
@@ -1,0 +1,90 @@
+# Minecraft Remote Console module.
+#
+# Copyright (C) 2015 Barnaby Gale
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to
+# deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+# sell copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies of the Software and its documentation and acknowledgment shall be
+# given in the documentation and software packages that this Software was
+# used.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import socket
+import select
+import struct
+import time
+
+
+class MCRconException(Exception):
+    pass
+
+
+class MCRcon(object):
+    socket = None
+
+    def connect(self, host, port, password):
+        if self.socket is not None:
+            raise MCRconException("Already connected")
+        self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.socket.connect((host, port))
+        self.send(3, password)
+
+    def disconnect(self):
+        if self.socket is None:
+            raise MCRconException("Already disconnected")
+        self.socket.close()
+        self.socket = None
+
+    def read(self, length):
+        data = b""
+        while len(data) < length:
+            data += self.socket.recv(length - len(data))
+        return data
+
+    def send(self, out_type, out_data):
+        if self.socket is None:
+            raise MCRconException("Must connect before sending data")
+
+        # Send a request packet
+        out_payload = struct.pack('<ii', 0, out_type) + out_data.encode('utf8') + b'\x00\x00'
+        out_length = struct.pack('<i', len(out_payload))
+        self.socket.send(out_length + out_payload)
+
+        # Read response packets
+        in_data = ""
+        while True:
+            # Read a packet
+            in_length, = struct.unpack('<i', self.read(4))
+            in_payload = self.read(in_length)
+            in_id, in_type = struct.unpack('<ii', in_payload[:8])
+            in_data_partial, in_padding = in_payload[8:-2], in_payload[-2:]
+
+            # Sanity checks
+            if in_padding != b'\x00\x00':
+                raise MCRconException("Incorrect padding")
+            if in_id == -1:
+                raise MCRconException("Login failed")
+
+            # Record the response
+            in_data += in_data_partial.decode('utf8')
+
+            # If there's nothing more to receive, return the response
+            if len(select.select([self.socket], [], [], 0)[0]) == 0:
+                return in_data
+
+    def command(self, command):
+        result = self.send(2, command)
+        time.sleep(0.003) # MC-72390 workaround
+        return result
--- a/python.d/spigotmc.chart.py
+++ b/python.d/spigotmc.chart.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+# Description: spigotmc netdata python.d module
+# Author: Austin S. Hemmelgarn (Ferroin)
+
+import socket
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+from third_party import mcrcon
+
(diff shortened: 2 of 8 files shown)


python.d/spigotmc.chart.py
```diff
@@ -0,0 +1,78 @@
...
+            # '§6TPS from last 1m, 5m, 15m: §a19.99, §a19.99, §a19.99\n'
+            # The values we care about are the three numbers after the :
+            tmp = raw.split(':')[1].split(',')
+            data['tps1'] = float(tmp[0].lstrip()[2:]) * PRECISION
+            data['tps5'] = float(tmp[1].lstrip()[2:]) * PRECISION
+            data['tps15'] = float(tmp[2].lstrip().rstrip()[2:]) * PRECISION
+        except (mcrcon.MCRconException, socket.error):
+            data['tps1'] = None
+            data['tps5'] = None
+            data['tps15'] = None
```
l2isbad: Not sure these `None` makes any sense

python.d/python_modules/third_party/mcrcon.py
```diff
@@ -0,0 +1,90 @@
...
+        out_length = struct.pack('<i', len(out_payload))
+        self.socket.send(out_length + out_payload)
+
+        # Read response packets
+        in_data = ""
+        while True:
+            # Read a packet
+            in_length, = struct.unpack('<i', self.read(4))
+            in_payload = self.read(in_length)
+            in_id, in_type = struct.unpack('<ii', in_payload[:8])
```
l2isbad: unused variable `in_type`

l2isbad:
Seems plugin doesn't support reconnect after restart of the sever. @Ferroin do you think it is not needed?

l2isbad:
please add new module to [python.d.conf](https://github.com/firehol/netdata/blob/master/conf.d/python.d.conf)

python.d/spigotmc.chart.py
Ferroin: I was under the impression from looking at the other Python plugins that returning `None` says that that value couldn't be collected.  It doesn't make much sense here to report anything if we failed (reporting 0 would be potentially misleading).

python.d/python_modules/third_party/mcrcon.py
Ferroin: Huh, that it is.  I hadn't really looked at this code in depth, just pulled it over from the upstream repo.  I'll get that taken care of.

Ferroin:
Hmm, I had forgotten about handling a reconnect.  Given that, it may actually make more sense to just re-do this using `SocketService` and send the authentication each time, although that may take me quite a while.

Also, I could have sworn I forgot about some file that needed updated, looks like `python.d.conf` is it.  I think this can probably be safely enabled by default, given that it has a working `check()` method and the check appears to be reasonably cheap (though that could probably be improved.

It will be Monday before I get a chance to update any of this, though I probably won't have anything then that handles reconnection.

python.d/spigotmc.chart.py
l2isbad: just don't add a key if it is missing

l2isbad:
[beanstalk](https://github.com/firehol/netdata/blob/master/python.d/beanstalk.chart.py) module has custom reconnect

python.d/spigotmc.chart.py
Ferroin: Ah, OK, that makes sense.

Ferroin:
@l2isbad Thanks for the pointer to the beanstalkd module.  I hadn't noticed the custom reconnect stuff there.

I've just pushed a commit that fixes your two comments on the code and adds the module to python.d.conf.

It also adds an initial attempt at reconnect support, though that doesn't appear to be completely working yet, most of the time when it would be used, things hang hard in the plugin (one of the python threads pegs out at 100% CPU usage, the plugin itself stops collecting data, and the stats for it stop updating as well).  Until that gets fixed, this shouldn't be merged, but I'm hoping someone might have some insights as to what's going wrong given that my preliminary debugging has turned up absolutely nothing.

Ferroin:
OK, the auto-reconnection stuff works now, but unfortunately what I had to do to make it work reliably means that the plugin only works on Linux systems.

In short, the issue is that for some reason the first call to `mcrcon.MCRcon.command()` after the remote end disconnecting hangs in the first `socket.recv()` call even if there's a timeout set.  I'm not quite sure why that is the case, but I've added a work-around that checks if the socket is in the `CONNECTED` TCP state or not, and tries to reconnect if it isn't.  This appears to reliably fix things (there might be a race condition with the server initiating a shutdown after that check while we're trying to get data back, but I've not seen it when testing this (Spigot takes a couple of seconds to shut down, and the first thing it does is close the remote console port, so I think it should be fine), and handling that would require adding more such checks, which would in turn make collection very expensive.

As far as the 'Linux only' part, it has to do with differences between the Linux and BSD networking stack handling of the `TCP_INFO` socket option.  On Linux, the call directly returns the structure containing the data we want, up to the number of bytes specified by the third argument to `getsockopts()`.  This makes it very easy to handle from Python.  On BSD however, you have to pass a pointer to a pre-allocated structure that the call will fill, and the call only returns success or failure status, which makes it a bit harder to handle in Python (and since I don't have a BSD system to test on right now, I'm not really comfortable trying to code it).

As of right now, the plugin's `check` method will bail if it's not running on a Linux system, but if someone wants to add BSD support, it wouldn't be too difficult to just remove that and move the check to a helper call that does the right thing based on the system type.

Ferroin:
Hmm, this appears to be a lot more expensive than I had realized to run.  Peak execution time is over 100ms, and the average looks to be about 75ms.

Do we want this disabled by default given how expensive it is, or should we just leave it enabled by default because it's so specific to a particular use case?

Ferroin:
Or, alternatively, should we bump up the default collection interval?  I doubt that statistics per-second are particularly useful for most people.

l2isbad:
> Do we want this disabled by default given how expensive it is

depends on cpu usage  i think

> Or, alternatively, should we bump up the default collection interval?

>  doubt that statistics per-second are particularly useful for most people.

Sounds reasonble

Ferroin:
CPU usage is pretty low (less than 1%), it's just that it takes at least double digit milliseconds to retrieve the data from the Spigot server.  Given that, and a bit of further digging that seems to suggest that it has minimal impact on the Spigot server itself, I think it's probably safe to just leave it enabled.

As far as collection period, I think 5 seconds is probably reasonable.  It looks like the TPS averages aren't updated every second anyway (though I'm not completely sure how often they actually are updated), and user counts typically don't change much second to second even on big servers (and most people seem to only care about how many people are on at a given time, not second by second values), so there's not much point in collecting second-by-second anyway.  I'll push a commit to update the collection period to 5 seconds, and then this should be ready to merge.

Ferroin:
Wanted to make sure this was working correctly and see how the changes impacted performance.  It looks like everything is working fine (at least, on the Spigot Minecraft server I'm running), so I think this should be ready to merge @ktsaou.

ktsaou:
hm... a conflicting change. Could you please solve it?

Ferroin:
Yeah, I should have remembered that that would conflict.

PR #3670, opened 2018-05-01
Add a plugin for the Unbound DNS server.

Ferroin:
Unbound is a caching, recursive DNS resolver primarily used as a local DNS cache for Linux client systems, and to provide basic DNS service for a local network from a router.  It provides a bunch of statistics over it's remote control interface for tools like Netdata to collect.

This PR first adds rudimentary TLS encapsulation and client authentication support to `SocketService`, which is needed to allow TCP based communication with the Unbound remote control interface (they insist on using TLS with an X.509 client certificate for authentication if you're not using a UNIX socket for the connection).  This adds a conditional dependency on the `ssl` module from the Python standard library, if this module isn't found (and it may not be on some embedded systems where space is at a premium, because it depends on a big library and is pretty big itself) then there just won't be any TLS support, but everything else will continue to work as normal.  TLS support comes with 3 new keys for the `SocketService` configuration:

* `ssl`: A simple boolean toggle controlling whether to use SSL/TLS or not.
* `ssl_key`: Specifies the path to an X.509 private key to use for client authentication.
* `ssl_cert`: Specifies the path to an X.509 certificate to be used for client authentication.

In the current default case of not using TLS support, this adds about 1.5-2.5ms to the initialization time of plugins using `SocketService` on most of my systems.  I've verified that this works correctly (that is, just like the existing `SocketService` module on systems that don't have the Python `ssl` module, and that existing modules work the same as before (other than the slightly higher initialization time mentioned above).

The plugin is then added by the second commit.  It currently collects the following stats:

* Query rates, including cache hit and miss rates, how many were dropped due to rate limiting, how many served expired data still in the cache, how many times items were pre-fetched into the cache and how many triggered upstream recursion.
* Various stats about the internal request processing list, including the average and maximum size over the previous collection period, the current number of requests in the list, how many are user requests (as opposed to internal requests such as pre-fetching requests), and how many requests had to be dropped because the list was full (which should obviously be zero most of the time).
* Average and median (50th percentile) processing times for upstream recursive requests.
* Optionally (if you turn on extended stats in both Unbound and the plugin), how many items are in each of the caches.

Unbound actually provides more stats than what this plugin collects, but most of them are largely only of interest to developers or individuals who are debugging low-level issues with Unbound's internals (things like exact memory usage of the various components, or per-thread stats).  Most of these (other than the per-thread stats) can be added pretty easily if desired, but I didn't include them here because I have no use for them.

The plugin supports auto-detection of most of the required settings when connecting to the local host if you tell it where the config for Unbound is, although this is dependent on `unbound.conf` being parseable as YAML (technically, it _is_YAML file, except Unbound allows tabs for indentation, and the upstream config example (and therefore almost every distribution config) use tabs).

The plugin itself is disabled by default as it requires some non-trivial changes from the default Unbound configuration to set up and enable the remote control interface, as well as some mucking about with permissions to get it all working correctly (although the permissions changes can be avoided by running Netdata either as root or as the same user as the Unbound server process).  

Example graphs from one of my systems (which has very little DNS load and as a result doesn't show most of the stats (because they're all just zero)):
image

Lightly tested on half a dozen of my own systems using a variety of remote-control interface configurations.

### DIFF
--- a/python.d/python_modules/bases/FrameworkServices/SocketService.py
+++ b/python.d/python_modules/bases/FrameworkServices/SocketService.py
@@ -4,6 +4,13 @@
 
 import socket
 
+try:
+    import ssl
+except:
+    _SSL_SUPPORT = False
+else:
+    _SSL_SUPPORT = True
+
 from bases.FrameworkServices.SimpleService import SimpleService
 
 
@@ -16,6 +23,9 @@ class SocketService(SimpleService):
         self.unix_socket = None
         self.dgram_socket = False
         self.request = ''
+        self.ssl = False
+        self.cert = None
+        self.key = None
         self.__socket_config = None
         self.__empty_request = "".encode()
         SimpleService.__init__(self, configuration=configuration, name=name)
@@ -56,10 +66,24 @@ class SocketService(SimpleService):
             self.__socket_config = None
             return False
 
+        if self.ssl:
+            try:
+                self.debug('Encapsulating socket with SSL')
+                self._sock = ssl.wrap_socket(self._sock,
+                                             keyfile=self.key,
+                                             certfile=self.cert,
+                                             server_side=False,
+                                             cert_reqs=ssl.CERT_NONE)
+            except (socket.error, ssl.SSLError) as error:
+                self.error('Failed to wrap socket.')
+                self._disconnect()
+                self.__socket_config = None
+                return False
+
         try:
             self.debug('connecting socket to "{address}", port {port}'.format(address=sa[0], port=sa[1]))
             self._sock.connect(sa)
-        except socket.error as error:
+        except (socket.error, ssl.SSLError) as error:
             self.error('Failed to connect to "{address}", port {port}, error: {error}'.format(address=sa[0],
                                                                                               port=sa[1],
                                                                                               error=error))
@@ -248,6 +272,21 @@ class SocketService(SimpleService):
                 self.port = int(self.configuration['port'])
             except (KeyError, TypeError):
                 self.debug('No port specified. Using: "{0}"'.format(self.port))
+
+            if _SSL_SUPPORT:
+                try:
+                    self.ssl = bool(self.configuration['ssl'])
+                except (KeyError, TypeError):
+                    self.debug('No SSL preference specified, not using SSL.')
+                    self.ssl = False
+                else:
+                    try:
+                        self.key = str(self.configuration['ssl_key'])
+                        self.cert = str(self.configuration['ssl_cert'])
+                    except (KeyError, TypeError):
+                        self.debug('No SSL client certificate configuration found.')
+                        self.key = None
+                        self.cert = None
 
         try:
             self.request = str(self.configuration['request'])
--- a/python.d/python_modules/bases/FrameworkServices/SocketService.py
+++ b/python.d/python_modules/bases/FrameworkServices/SocketService.py
@@ -281,8 +281,8 @@ class SocketService(SimpleService):
                 self.debug('No SSL preference specified, not using SSL.')
 
             if self.ssl and _SSL_SUPPORT:
-                self.key = str(self.configuration.get('ssl_key'))
-                self.cert = str(self.configuration.get('ssl_cert'))
+                self.key = self.configuration.get('ssl_key')
+                self.cert = self.configuration.get('ssl_cert')
                 if not self.cert:
                     # If there's not a valid certificate, clear the key too.
                     self.debug('No valid SSL client certificate configuration found.')
--- a/python.d/python_modules/bases/FrameworkServices/SocketService.py
+++ b/python.d/python_modules/bases/FrameworkServices/SocketService.py
@@ -273,24 +273,27 @@ class SocketService(SimpleService):
             except (KeyError, TypeError):
                 self.debug('No port specified. Using: "{0}"'.format(self.port))
 
-            try:
-                self.ssl = bool(self.configuration['ssl'])
-                if self.ssl and not _SSL_SUPPORT:
-                    self.warning('SSL requested but not SSL module found, disabling SSL support.')
-                    self.ssl = False
-            except (KeyError, TypeError):
-                if _SSL_SUPPORT:
-                    self.debug('No SSL preference specified, not using SSL.')
+            self.ssl = bool(self.configuration.get('ssl', self.ssl))
+            if self.ssl and not _SSL_SUPPORT:
+                self.warning('SSL requested but not SSL module found, disabling SSL support.')
                 self.ssl = False
+            if _SSL_SUPPORT and not self.ssl:
+                self.debug('No SSL preference specified, not using SSL.')
 
             if self.ssl and _SSL_SUPPORT:
-                try:
-                    self.key = str(self.configuration['ssl_key'])
-                    self.cert = str(self.configuration['ssl_cert'])
-                except (KeyError, TypeError):
-                    self.debug('No SSL client certificate configuration found.')
+                self.key = str(self.configuration.get('ssl_key'))
+                self.cert = str(self.configuration.get('ssl_cert'))
+                if not self.cert:
+                    # If there's not a valid certificate, clear the key too.
+                    self.debug('No valid SSL client certificate configuration found.')
                     self.key = None
                     self.cert = None
+                elif not self.key:
+                    # If a key isn't listed, the config may still be
+                    # valid, because there may be a key attached to the
+                    # certificate.
+                    self.notice('No SSL client key specified, assuming it\'s attached to the certificate.')
+                    self.key = None
 
         try:
             self.request = str(self.configuration['request'])
--- a/python.d/python_modules/bases/FrameworkServices/SocketService.py
+++ b/python.d/python_modules/bases/FrameworkServices/SocketService.py
@@ -7,9 +7,9 @@ import socket
 try:
     import ssl
 except ImportError:
-    _SSL_SUPPORT = False
+    _TLS_SUPPORT = False
 else:
-    _SSL_SUPPORT = True
+    _TLS_SUPPORT = True
 
 from bases.FrameworkServices.SimpleService import SimpleService
 
@@ -23,7 +23,7 @@ class SocketService(SimpleService):
         self.unix_socket = None
         self.dgram_socket = False
         self.request = ''
-        self.ssl = False
+        self.tls = False
         self.cert = None
         self.key = None
         self.__socket_config = None
@@ -66,15 +66,15 @@ class SocketService(SimpleService):
             self.__socket_config = None
             return False
 
-        if self.ssl:
+        if self.tls:
             try:
-                self.debug('Encapsulating socket with SSL')
+                self.debug('Encapsulating socket with TLS')
                 self._sock = ssl.wrap_socket(self._sock,
                                              keyfile=self.key,
                                              certfile=self.cert,
                                              server_side=False,
                                              cert_reqs=ssl.CERT_NONE)
-            except (socket.error, ssl.SSLError) as error:
+            except (socket.error, ssl.TLSError) as error:
                 self.error('Failed to wrap socket.')
                 self._disconnect()
                 self.__socket_config = None
@@ -83,7 +83,7 @@ class SocketService(SimpleService):
         try:
             self.debug('connecting socket to "{address}", port {port}'.format(address=sa[0], port=sa[1]))
             self._sock.connect(sa)
-        except (socket.error, ssl.SSLError) as error:
+        except (socket.error, ssl.TLSError) as error:
             self.error('Failed to connect to "{address}", port {port}, error: {error}'.format(address=sa[0],
                                                                                               port=sa[1],
                                                                                               error=error))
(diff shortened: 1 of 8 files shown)


Ferroin:
OK, so Codacy is going to continue complaining about there being no protocol specified in the `ssl.wrap_socket()` call this adds to `SocketService`.  For Python2 interoperability reasons, we would need to specify SSLv23 as the protocol, but that's the synonymous with the default in all Python versions Netdata supports, and therefore is kind of superfluous.  Additionally, that default behavior it provides is actually what we want here (namely, it tries to find out what the highest protocol version both client and server support is, and uses that).

If there's a strong desire to change this, I'll be happy to, but otherwise I think it should just stay as-is.

Ferroin:
Oh, also, the complaints about access before assignment on lines 116, 118, and 122 are bogus.  The attributes it's complaining about get assigned by the `self._parse_config()` call (not sure why it's confused though, maybe it doesn't like that I'm not using `super()` or directly calling through `SocketService`?).

python.d/unbound.chart.py
```diff
@@ -0,0 +1,158 @@
...
+    def _get_data(self):
+        raw = self._get_raw_data()
+        data = dict()
+        tmp = dict()
+        for line in raw.splitlines():
+            stat = line.split('=')
+            tmp[stat[0]] = stat[1]
+        for item in STAT_MAP.keys():
+            if item in tmp.keys():
+                data[STAT_MAP[item]] = float(tmp[item])
```
l2isbad: https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/charts.py#L318-L322

So if you really need `float` value you have to multiply it to 100/1000/etc in the code and divide at charts.

python.d/python_modules/bases/FrameworkServices/SocketService.py
```diff
@@ -16,6 +23,9 @@ def __init__(self, configuration=None, name=None):
         self.unix_socket = None
         self.dgram_socket = False
         self.request = ''
+        self.ssl = False
```
l2isbad: Well, are you sure about `ssl` attr? I think it is always true if `self.cert and self.key`.

python.d/python_modules/bases/FrameworkServices/SocketService.py
```diff
@@ -249,6 +273,21 @@ def _parse_config(self):
...
+                    self.ssl = False
+                else:
+                    try:
+                        self.key = str(self.configuration['ssl_key'])
+                        self.cert = str(self.configuration['ssl_cert'])
+                    except (KeyError, TypeError):
+                        self.debug('No SSL client certificate configuration found.')
+                        self.key = None
+                        self.cert = None
+
```
l2isbad: I suggest to avoid stacking `try/except` blocks. 

I think this is enough: (yeah, `SocketService` should be rewritten)
```python
 if _SSL_SUPPORT:
   self.key = self.configuration['ssl_key']
   self.cert = self.configuration['ssl_cert']
```

python.d/unbound.chart.py
```diff
@@ -0,0 +1,158 @@
+# -*- coding: utf-8 -*-
+# Description: unbound netdata python.d module
+# Author: Austin S. Hemmelgarn (Ferroin)
+
+import os
+import yaml
```
l2isbad: Netdata shipped with `yaml` packages for py2, py3.
If the only thing you need is `yaml.load` i suggest to use 
https://github.com/l2isbad/netdata/blob/master/python.d/python_modules/bases/loaders.py#L32-L50

python.d/unbound.chart.py
```diff
@@ -0,0 +1,158 @@
...
+        if self.unix_socket:
+            self.debug('Using unix socket: {0}'.format(self.unix_socket))
+            for key in self.definitions.keys():
+                self.definitions[key]['options'][4] = 'Local'
+        else:
+            self.debug('Connecting to: {0}:{1}'.format(self.host, self.port))
+            self.debug('Using key: {0}'.format(self.key))
+            self.debug('Using certificate: {0}'.format(self.cert))
+            for key in self.definitions.keys():
+                self.definitions[key]['options'][4] = self.host
```
l2isbad: Huge `__init__` is a bad thing. Please move all logic to separate method(s)/check

python.d/unbound.chart.py
Ferroin: Hmm, forgot about that.  This may take a bit more thought, as some of the stuff _needs_ to be `floats` (the request timing info), some of it really shouldn't be (the cache item counts, you can't have half an item in cache), and some I'm not quite sure about (I don't know if the query counters are supposed to be floats or not).

python.d/python_modules/bases/FrameworkServices/SocketService.py
Ferroin: Client certificates aren't inherently mandatory though (think HTTPS for an example of this), and matching on the key and certificate being defined would mandate generating a certificate to use SSL/TLS at all (and some stuff which doesn't require client authentication with certificates will still try to validate the client cert, so requiring a client certificate for such services makes setup _really_ complicated).

python.d/unbound.chart.py
Ferroin: Oh, completely missed that.  I'll get it updated to use that then.

python.d/unbound.chart.py
Ferroin: normally I would agree, and I actually originally had everything from line 105 to 123 in a separate method, but I was having significant trouble getting it to work and it was just easier to push it into `__init__`.  I intend to get that taken care of, I just wanted to get the current (mostly working other than the thing below about `float` conversion) code out there for some extra review (and I'm glad I did).

python.d/unbound.chart.py
Ferroin: On second inspection, it looks like that won't be quite enough, as I kind of need to sanely catch parsing errors (especially since they're reasonably likely due to tab indentation).

Would a construct like what's used in the file you linked to import the requisite libraries work here?

l2isbad: Should the `check()` return true if an `yaml` parse error occurs? (If config exist and readable but has wrong `yaml` syntax)

Ferroin: Yes, in general it probably should.  The actual check being made is that a connection works.  If we don't have any permission to access the unix socket, or we don't have a valid client certificate configured, the connect will fail, and in turn the check will fail.

The attempt to parse the unbound config is simply there to try and auto-fill config, but anything specified manually (except the port currently, which gets overridden either way if we try to parse the unbound config, though that probably needs to change) will override that.

If for example you manually specify the certificate and key, but specify a bogus unbound config (or the config parsing fails because of the tab s most distros have in their default unbound config), we should just fall back to trying on port 8953 on the loopback interface.

python.d/python_modules/bases/FrameworkServices/SocketService.py
Ferroin: The intent here was to avoid wasting time with the client certificate stuff if we're not using SSL, but still provide slightly more specific debug info (there's a pretty big difference between not using SSL at all and not having a client certificate), though looking back at it it's kind of non-obvious what it's doing (`else` clauses on `try`/`except` blocks are kind of rarely used).

I think I'm going to just rework this whole section anyway so that it doesn't silently ignore a request to use SSL if there is no `ssl` module.

l2isbad: i mean simple `self.key = self.configuration.get('ssl_key')` does same work as `try/except`
you can use `self.debug` ofc if you need
```python
 if _SSL_SUPPORT:
   self.key = self.configuration.get('ssl_key')
   if not self.key:
       self.debug("......")
   self.cert = self.configuration.get('ssl_cert')
```

python.d/unbound.chart.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _get_data(self):
+        raw = self._get_raw_data()
+        data = dict()
+        tmp = dict()
+        for line in raw.splitlines():
+            stat = line.split('=')
+            tmp[stat[0]] = stat[1]
+        for item in STAT_MAP.keys():
+            if item in tmp.keys():
```
l2isbad: `.keys()` is not needed, just do `for item in DICT` (same for `if item in DICT`)
this is not the only place you use `dict.keys()`

python.d/python_modules/bases/FrameworkServices/SocketService.py
Ferroin: Ah, OK, that makes much more sense.  Kind of surprised I didn't realize that.

python.d/unbound.chart.py
Ferroin: This is mostly a habit from where I work (most of our programmers don't do much Python coding, and stuff like this helps a bit clarity-wise for some of them).  I'll push a commit to convert all of these shortly.

l2isbad: I am not a programmer but that is a bad habbit. `dict.keys()` produces a new list (py2) and `in` list is much slower than `in` dict.

python.d/python_modules/bases/FrameworkServices/SocketService.py
```diff
@@ -249,6 +273,28 @@ def _parse_config(self):
...
+            self.ssl = bool(self.configuration.get('ssl', self.ssl))
+            if self.ssl and not _SSL_SUPPORT:
+                self.warning('SSL requested but not SSL module found, disabling SSL support.')
+                self.ssl = False
+            if _SSL_SUPPORT and not self.ssl:
+                self.debug('No SSL preference specified, not using SSL.')
+
+            if self.ssl and _SSL_SUPPORT:
+                self.key = str(self.configuration.get('ssl_key'))
+                self.cert = str(self.configuration.get('ssl_cert'))
```
l2isbad: i believe `str` is not needed. How else can you set the key/cert? 
And `str(None)` is always True.

python.d/unbound.chart.py
```diff
@@ -0,0 +1,167 @@
...
+        # The unbound control protocol is always SSL encapsulated
+        # unless it's used over a UNIX socket, so enable SSL _before_
+        # doing the normal SocketService initialization.
+        configuration['ssl'] = True
+        self.port = 8935
+        SocketService.__init__(self, configuration, name)
+        self.ext = self.configuration.get('extended', None)
+        self.ubconf = self.configuration.get('ubconf', None)
+        self.order = ORDER
+        self.definitions = CHARTS
```
l2isbad: Since `order`/`definitions` can be modified you have to copy `ORDER` and `CHARTS`
(`from copy import deepcopy` for `CHARTS`)

python.d/unbound.chart.py
```diff
@@ -0,0 +1,167 @@
...
+            self.port = 8953
+
+    def check(self):
+        # We need to check that auth works, otherwise there's no point.
+        self._connect()
+        result = bool(self._sock)
+        self._disconnect()
+        return result
+
+    def _check_raw_data(self, data):
```
l2isbad: make it `@staticmethod` (remove `self`, add decorator)

l2isbad:
About `codacy` errors.
Lets replace `if not key: key = value` with `key = key or value` and see if it still complains after this change.

l2isbad:
are you done @Ferroin ?

Ferroin:
I think so, the only thing left not quite working is applying the info text to charts for jobs other than a default one, but I'm not hugely worried about that, I would assume people who have gone through getting this set up are smart enough to look at the Unbound documentation for info on what the stats mean.

Ferroin:
Just saw #3678.  I think it would probably be good to match the key names this PR adds for the `SocketService` SSL/TLS support with the ones added by that to `UrlService`.  Should I push that as an extra commit here, or would it be better to make that a separate PR?

l2isbad:
> push that as an extra commit here

^^

Ferroin:
OK, with that added (and the stupid mistake I made doing batch conversion corrected), this should be good to go.

l2isbad:
update makefiles @Ferroin

ktsaou:
Is this ready to be merged?

python.d/python_modules/bases/FrameworkServices/SocketService.py
```diff
@@ -249,6 +273,28 @@ def _parse_config(self):
...
+                if not self.cert:
+                    # If there's not a valid certificate, clear the key too.
+                    self.debug('No valid TLS client certificate configuration found.')
+                    self.key = None
+                    self.cert = None
+                elif not self.key:
+                    # If a key isn't listed, the config may still be
+                    # valid, because there may be a key attached to the
+                    # certificate.
+                    self.notice('No TLS client key specified, assuming it\'s attached to the certificate.')
```
l2isbad: `self.notice` ?

Ferroin: Strictly speaking it's not an error as far as the `ssl` module is concerned, as long as there's a key attached to the certificate file (which is as simple as appending the PEM encoded data to the end of the certificate file).  Unfortunately, there's not really any easy way for us to validate that there is a key attached to the certificate (unless you want me to put a bare-bones PEM parser in here).

l2isbad: i mean what is `self.notice` ? logger has no `notice` method

ktsaou:
merged it!

PR #3671, opened 2018-05-01
Update .gitignore to ignore `.swp` files created by Vim

Ferroin:
Discovered that this wasn't already the case when I almost accidentally committed such a file while working on PR #3670.

Vim (and some other Vi derivatives) create these files while editing other files, they can safely be ignored.


ktsaou:
thank you!

PR #3709, opened 2018-05-10
Added Neo4j support

wadael:
Added support for Neo4j, the graph database. www.neo4j.com

Indicators are:
* total number of nodes
* number of nodes for top5 labels
* number of nodes for top5 relations

Probably not perfect implementation (I'm a python noob).

### DIFF
--- a/conf.d/python.d/neo.conf
+++ b/conf.d/python.d/neo.conf
@@ -6,4 +6,4 @@ neo:
     pwd : 'password'
     update_every : 30
     priority: 22000 # the JOB's order on the dashboard
-    retries: 10       # the JOB's number of restoration attempts
+    retries: 5       # the JOB's number of restoration attempts
--- a/conf.d/python.d/neo.conf
+++ b/conf.d/python.d/neo.conf
@@ -1,0 +1,9 @@
+neo:
+    name : 'neo'
+    host : 'localhost'
+    port : '7687' # port where Bolt protocol is listening
+    user : 'neo4j'
+    pwd : 'password'
+    update_every : 30
+    priority: 22000 # the JOB's order on the dashboard
+    retries: 10       # the JOB's number of restoration attempts
--- a/python.d/README.md
+++ b/python.d/README.md
@@ -2512,11 +2512,13 @@ Displays key metrics for the Neo4j graph database (www.neo4j.com)
    * number of nodes for the top5 labels
    * number of nodes for the top5 relations
 
-Requires the neo4j-module (JS driver) to access the wanted server.
-By default, listens to Neo4j on localhost, accessible with neo4j/password credentials 
+Requires the neo4j-driver module (JS driver) to access the wanted server.
+Users can install the neo4j-driver package by running the /usr/libexec/netdata/python-modules-installer.sh script (with sudo) and restart the netdata service (sudo service netdata restart).
+
+By default, listens to a Neo4j instance on localhost, accessible with neo4j/password credentials. You can update the credentials in the  /etc/netdata/python.d/neo.conf file. 
 
 ## configuration   
-See the neo.conf file to change the settings
+See the /etc/netdata/python.d/neo.conf file to change the settings
 ```
 neo:
     name : 'neo'
--- a/python.d/README.md
+++ b/python.d/README.md
@@ -2504,3 +2504,29 @@ apache_log:
 Module has preconfigured jobs for nginx, apache and gunicorn on various distros.
 
 ---
+
+# neo
+
+Displays key metrics for the Neo4j graph database (www.neo4j.com) 
+   * Total number of nodes
+   * number of nodes for the top5 labels
+   * number of nodes for the top5 relations
+
+Requires the neo4j-module (JS driver) to access the wanted server.
+By default, listens to Neo4j on localhost, accessible with neo4j/password credentials 
+
+## configuration   
+See the neo.conf file to change the settings
+```
+neo:
+    name : 'neo'
+    host : 'localhost'
+    port : '7687' # port where Bolt protocol is listening
+    user : 'neo4j'
+    pwd : 'password'
+    update_every : 30
+    priority: 22000 # the JOB's order on the dashboard
+    retries: 10       # the JOB's number of restoration attempts
+```
+
+---
--- a/python.d/neo.chart.py
+++ b/python.d/neo.chart.py
@@ -1,0 +1,121 @@
+# -*- coding: utf-8 -*-
+# Description:  netdata python.d module for monitoring Neo4j
+# Author: Jerome Baton, @wadael , copied from the example of Pawel Krupa (paulfantom)
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+try:
+    from neo4j.v1 import GraphDatabase
+    NEO4JDRIVER = True
+except ImportError:
+    NEO4JDRIVER = False
+
+
+ORDER = ['monnodes','monlabels','monrels']
+
+CHARTS = {
+    'monnodes': {
+        'options': [None, 'Nodes count', 'Nodes', 'Node count', 'neo4j', 'line'],
+        'lines': [
+             ['id', 'name', 'absolute', 1, 1, 'hidden']
+        ]
+    },
+    'monrels': {
+        'options': [None, 'Relations count', 'Nodes', 'Most present relations', 'neo4j', 'line'],
+        'lines': [
+             ['id', 'name', 'absolute', 1, 1, 'hidden']
+        ]
+    },
+    'monlabels': {
+        'options': [None, 'Relations count', 'Nodes', 'Most present labels', 'neo4j', 'line'],
+        'lines': [
+            ['id', 'name', 'absolute', 1, 1, 'hidden']
+        ]
+    }
+
+
+}
+
+update_every = 10
+priority = 99000
+retries = 10
+
+
+class Service(SimpleService):
+
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.neosession = None
+        self.alive = True
+
+    def __exit__(self, exc_type, exc_value, traceback):
+       if NEO4JDRIVER:
+            self.neosession.close()
+            self.neodriver.close()
+
+    def check(self):
+        if not NEO4JDRIVER:
+            self.error("the 'neo4j-driver' module is needed to use neo.chart.py. See the script python-modules-installer.sh in /usr/libexec/netdata/python.d")
+            return False
+
+        self.connect()
+        return True
+
+    def get_data_from_cypher_query(self, chartName, data, cypherQuery):
+        try:
+            with self.neosession.begin_transaction() as tx:
+                for neorecord in tx.run(cypherQuery):
+                    if neorecord["label"] not in self.charts[chartName]:
+                        self.charts[chartName].add_dimension([ neorecord["label"] ])
+                    data[ neorecord["label"] ]  = neorecord["cnt"]
+                    tx.close()
+        except Exception:
+            return None
+
+    def get_data(self):
+
+        """
+        :return: dict
+        """
+        if not self.is_alive():
+            return None
+
+        if not NEO4JDRIVER:
+            # self.error('neo4j-driver not available. Please install.')
+            return None
+
+        data = dict()
+
+        # nodes (total)
+        self.get_data_from_cypher_query( 'monnodes',data, "MATCH (n) RETURN 'Total' AS label, count(n) AS cnt")
+
+        # relations (top 5)
+        self.get_data_from_cypher_query( 'monrels',data,  "MATCH (n)-[r]-(m) RETURN type(r) as label, count(r) AS cnt ORDER BY cnt DESC LIMIT 5")
+
+        # labels (top 5)
+        self.get_data_from_cypher_query( 'monlabels',data,  "MATCH (a) WITH DISTINCT LABELS(a) AS temp, COUNT(a) AS tempCnt UNWIND temp AS label RETURN label, SUM(tempCnt) AS cnt ORDER BY cnt DESC LIMIT 5")
+        return data
+
+    def connect(self):
+        host = self.configuration.get('host')
+        boltport = self.configuration.get('port','7687')
+        uri = "bolt://" + host + ":" + boltport
+        self.neodriver = GraphDatabase.driver(uri, auth=(self.configuration.get('user'), self.configuration.get('pwd') ))
+
+        if self.neosession == None:
+            self.neosession = self.neodriver.session()
+
+    def reconnect(self):
+        try:
+            self.neosession = self.neodriver.session()
+            self.alive = True
+            return True
+        except Exception:
+            return False
+
+    def is_alive(self):
+        if not self.alive:
+            return self.reconnect()
+        return True
--- a/python.d/neo.chart.py
+++ b/python.d/neo.chart.py
@@ -36,11 +36,6 @@ CHARTS = {
 
 }
 
-update_every = 10
-priority = 99000
-retries = 10
-
-
 class Service(SimpleService):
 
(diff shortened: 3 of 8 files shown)


Ferroin:
No personal experience with Neo4J, so i can't really test it, but:

* You need entries in the various Makefiles to get these to install as part of the standard install (https://github.com/firehol/netdata/blob/master/python.d/Makefile.am and https://github.com/firehol/netdata/blob/master/conf.d/Makefile.am).
* Not sure if it's mandated, but even if not, it's good practice to put basic documentation for the plugin in https://github.com/firehol/netdata/blob/master/python.d/README.md.
* Watch out for trailing whitespace.  Check the Codacy details about this.
* Take a look at how the other python.d plugin config files look.  Yours doesn't have to look exactly the same, but at least talking about what options are provided and what they do is a good idea.

On the note of the merge conflict in `web/dashboard_info.js`, it should be trivial to resolve without a rebase (just keep all the existing changes, and merge yours in).

I'll have some further comments in the code itself shortly.

python.d/neo.chart.py
```diff
@@ -0,0 +1,97 @@
...
+# if you want to add a chart:
+#     insert it in ORDER
+#     copy paste and rename a chart in CHARTS
+#     code your query to return two results per line (label , value)
+
+
+# Warning
+#    charts take time to appear in the browser, also dont forget to clean the cache
+#    see /var/log/netdata/error.log to get some info and know if your code is running
+#    (python.d exited with -1  is bad news)
```
Ferroin: The comments here are probably unneeded, as the case is the same for any of the python.d plugins.

python.d/neo.chart.py
```diff
@@ -0,0 +1,97 @@
+# -*- coding: utf-8 -*-
+# Description:  netdata python.d module for monitoring Neo4j
+# Author: Jerome Baton, copied from the example of Pawel Krupa (paulfantom)
+
+
+# if you want to add a chart:
+#     insert it in ORDER
+#     copy paste and rename a chart in CHARTS
+#     code your query to return two results per line (label , value)
```
Ferroin: Long term, it probably makes sense to change this so that the user doesn't have to edit the plugin itself, but can just modify the configuration file to get additional dimensions.

python.d/neo.chart.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+# Warning
+#    charts take time to appear in the browser, also dont forget to clean the cache
+#    see /var/log/netdata/error.log to get some info and know if your code is running
+#    (python.d exited with -1  is bad news)
+
+
+from bases.FrameworkServices.SimpleService import SimpleService
+from neo4j.v1 import GraphDatabase
```
Ferroin: Two things here:

* Make sure this dependency is documented when you add plugin info to the README.md file.
* This needs to fail safely so you get a reasonable error message.  Take a look at how the beanstalkd plugin (https://github.com/firehol/netdata/blob/master/python.d/beanstalk.chart.py) handles it's external dependencies, it wraps the import in a try block and sets a global variable based on the result (success or failure), and then checks that global variable in it's `check()` method and spits out an error message there if the import failed.

conf.d/python.d/neo.conf
```diff
@@ -0,0 +1,13 @@
+# this conf files must be put in etc/netdata/python.d
+# 2018, jbaton
+
+neo:
+    name : 'neo'
+    host : 'localhost'
+    boltport : '7687'
```
Ferroin: This should ideally just be `port` so it matches up with all the other plugins that make socket connections.

python.d/neo.chart.py
```diff
@@ -0,0 +1,97 @@
...
+        # nodes (total)
+        self.get_data_from_cypher_query( 'monnodes',data, neosession, "MATCH (n) RETURN 'Total' AS label, count(n) AS cnt")
+
+        # relations (top 5)
+        self.get_data_from_cypher_query( 'monrels',data, neosession, "MATCH (n)-[r]-(m) RETURN type(r) as label, count(r) AS cnt ORDER BY cnt DESC LIMIT 5")
+        
+        # labels (top 5)
+        self.get_data_from_cypher_query( 'monlabels',data, neosession, "MATCH (a) WITH DISTINCT LABELS(a) AS temp, COUNT(a) AS tempCnt UNWIND temp AS label RETURN label, SUM(tempCnt) AS cnt ORDER BY cnt DESC LIMIT 5")
+       
+        neosession.close()
```
Ferroin: How expensive is it to re-create a session each time you want to fetch data?  Is this a requirement of the protocol itself, or is it possible to reuse the session?

wadael:
Hi, I followed your kind comments.

I have an issue, it seems that the deployed python-modules-installler is not the one at the root of the project, I added the installation of neo4j-driver. It is absent after running netdata-installer.sh

The JS is also an issue, must improve my git-fu

wadael:
I have updated the JS file with the latest from master via the good old copypaste method, did my edits and pushed. Still conflicting.
Looks like I cant do it from my side. Can you ?

Ferroin:
The current merge issue is because you are re-adding things that were already in master (lines 52-57, 369-379, and 2102-2125).  At this point, you will likely have to manually rebuild the branch on top of the current master in your local repository by doing the following:

Export the changes as a patch using git diff:

        git diff 0d968c3 HEAD > ../patch

Manually edit the patch to remove the hunks you're patching in by hand from master in `web/dashboard_info.js`.
    
Delete your current branch, switch to your local master, update from the upstream repo, and then create a new branch with the same name as the current one:

        git checkout master
        git branch -D 3688
        git pull git://github.com/firehol/netdata master
        git checkout -b 3688

Manually apply the patch to the working tree:

       patch -p1 < ../patch

That will complain about hunks failing for `web/dashboard_info.js`.  The blocks for the individual Neo4J graphs should already be there, you should just need to manually add the bit for Neo4J in the `netdataDashboard.menu` section (lines 380-384 in the copy that's part of the current PR) and delete the `.rej` file that the patch command generates.

Once you've got that resolved, double check the patched working tree and commit it (ideally using the same commit message you did for the first commit for the old copy of the branch).  At that point, test it to make sure everything works (keep the patch around until you've done this).

After testing it locally, you can update this PR with the new branch by doing a force push to your fork of the netdata repository with:

    git push -f origin 3688

wadael:
@Ferroin thank you for your kind help and greatly appreciated explanations.

If you want to try Neo4j, use Docker as explained on page https://hub.docker.com/_/neo4j/

docker run \
    --publish=7474:7474 --publish=7687:7687 \
    --volume=$HOME/neo4j/data:/data \
    neo4j

then go to http://localhost:7474  and in the input bix on top, type 
   :play movies  
and click the arrow

go to the second slide, click the QL, it get copied in the input box, click the arrow.
Enjoy the graph returned. See the charts, now with more than 0 elements.

conf.d/python.d/neo.conf
```diff
@@ -0,0 +1,9 @@
+neo:
+    name : 'neo'
+    host : 'localhost'
+    port : '7687' # port where Bolt protocol is listening
+    user : 'neo4j'
+    pwd : 'password'
+    update_every : 30
+    priority: 22000 # the JOB's order on the dashboard
```
Ferroin: Not sure if it's an issue, but this is set different from what you have in the module itself.  Also, if it's set there, you don't have to have it set here.

Ferroin:
Not sure what's up with the Travis build, perhaps @ktsaou might have some ideas there?

conf.d/python.d/neo.conf
wadael: so I will remove updat_every and priority from the code. Thanks for the spotting. Did I say already I learned the little Python I know with this plugin's plugin :)

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+    NEO4JDRIVER = True
+except ImportError:
+    NEO4JDRIVER = False
+
+
+ORDER = ['monnodes','monlabels','monrels']
+
+CHARTS = {
+    'monnodes': {
+        'options': [None, 'Nodes count', 'Nodes', 'Node count', 'neo4j', 'line'],
```
l2isbad: `family` should be lower case, same for `units`

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+        'lines': [
+             ['id', 'name', 'absolute', 1, 1, 'hidden']
```
l2isbad: Reason for `hidden` dimensions?

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+       if NEO4JDRIVER:
+            self.neosession.close()
+            self.neodriver.close()
+
+    def check(self):
+        if not NEO4JDRIVER:
+            self.error("the 'neo4j-driver' module is needed to use neo.chart.py. See the script python-modules-installer.sh in /usr/libexec/netdata/python.d")
+            return False
+
+        self.connect()
```
l2isbad: Any exceptions here?

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+class Service(SimpleService):
+
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.neosession = None
+        self.alive = True
+
+    def __exit__(self, exc_type, exc_value, traceback):
```
l2isbad: Why is it here?

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def get_data_from_cypher_query(self, chartName, data, cypherQuery):
+        try:
+            with self.neosession.begin_transaction() as tx:
+                for neorecord in tx.run(cypherQuery):
+                    if neorecord["label"] not in self.charts[chartName]:
+                        self.charts[chartName].add_dimension([ neorecord["label"] ])
+                    data[ neorecord["label"] ]  = neorecord["cnt"]
+                    tx.close()
+        except Exception:
```
l2isbad: make `exception` specific
in case of  `exception` we have to set `self.alive` to `False`

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if self.neosession == None:
+            self.neosession = self.neodriver.session()
+
+    def reconnect(self):
+        try:
+            self.neosession = self.neodriver.session()
+            self.alive = True
+            return True
+        except Exception:
```
l2isbad: make `exception` specific

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # labels (top 5)
+        self.get_data_from_cypher_query( 'monlabels',data,  "MATCH (a) WITH DISTINCT LABELS(a) AS temp, COUNT(a) AS tempCnt UNWIND temp AS label RETURN label, SUM(tempCnt) AS cnt ORDER BY cnt DESC LIMIT 5")
+        return data
+
+    def connect(self):
+        host = self.configuration.get('host')
+        boltport = self.configuration.get('port','7687')
+        uri = "bolt://" + host + ":" + boltport
+        self.neodriver = GraphDatabase.driver(uri, auth=(self.configuration.get('user'), self.configuration.get('pwd') ))
```
l2isbad: i think you can move this to `__init__`

I may be wrong but looks like `self.neosession = self.neodriver.session()` is the thing where exception can arise
So we need not `connect`, just return `self.reconnect()` from `check()`. `self.alive` must be `False` from start.

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+
+        """
+        :return: dict
+        """
+        if not self.is_alive():
+            return None
+
+        if not NEO4JDRIVER:
+            # self.error('neo4j-driver not available. Please install.')
+            return None
```
l2isbad: not needed here. I see same in `check`

python.d/README.md
```diff
@@ -2504,3 +2504,31 @@ apache_log:
...
+## configuration   
+See the /etc/netdata/python.d/neo.conf file to change the settings
+```
+neo:
+    name : 'neo'
+    host : 'localhost'
+    port : '7687' # port where Bolt protocol is listening
+    user : 'neo4j'
+    pwd : 'password'
+    update_every : 30
```
l2isbad: i suggest to move `update_every` to module if you think 30 should be default

python.d/neo.chart.py
```diff
@@ -0,0 +1,116 @@
...
+        data = dict()
+
+        # nodes (total)
+        self.get_data_from_cypher_query( 'monnodes',data, "MATCH (n) RETURN 'Total' AS label, count(n) AS cnt")
+
+        # relations (top 5)
+        self.get_data_from_cypher_query( 'monrels',data,  "MATCH (n)-[r]-(m) RETURN type(r) as label, count(r) AS cnt ORDER BY cnt DESC LIMIT 5")
+
+        # labels (top 5)
+        self.get_data_from_cypher_query( 'monlabels',data,  "MATCH (a) WITH DISTINCT LABELS(a) AS temp, COUNT(a) AS tempCnt UNWIND temp AS label RETURN label, SUM(tempCnt) AS cnt ORDER BY cnt DESC LIMIT 5")
```
l2isbad: Well, i think you can add some classes to the top and use them here, like:

```python
class Monnodes:
    name = "monnodes"
    request = "MATCH (n) RETURN 'Total' AS label, count(n) AS cnt"

....

self.get_data_from_cypher_query(Monnodes.name, data, Monnodes.request)
```

python.d/neo.chart.py
Ferroin: @wadael Just a bit of background on the whole `except Exception:` thing since you're reasonably new to Python:

1. This is exactly equivalent to just putting `except:`, so if you _really_ need to catch _everything_ without reporting on what happened, do that instead.
2. Catching everything is almost always a bad idea.  Syntax errors are also a subclass of `Exception`, as are things like passing incorrect variable types to functions (`TypeError`), and trying to access non-existent indexes in iterables or keys in dictionaries (`IndexError` and `KeyError`).  As a result, just catching all types of exceptions can hide real issues with the code itself, which makes debugging much more complicated.  Being specific with what you are actually handling with the except clause is also good practice as it makes sure that things you didn't expect and thus don't properly handle result in actual tracebacks.

In this case, you probably want to specify whatever the Neo4J module's main error class is.

python.d/neo.chart.py
wadael: Hi sorry for the delay.
The reason is I saw this in an example and I am pretty confident it did not work without.

python.d/neo.chart.py
wadael: I liked the idea to write a destructor again :)  
Also, saw this in some tut. I knew close to zero Python before that plugin of plugin.
Thought it was hygienic to have a destructor

python.d/README.md
wadael: I'll do that as maintenance when I'll understand this sentence and packages in python

l2isbad:
yo @wadael 

If you want this is being merged let's continue to work on it

ktsaou:
@l2isbad 	please decide if this is going to be merged.
I see @wadael 	has not signed the CLA, so unless he does, we can't use his code.

l2isbad:
And it looks like it won't

ktsaou:
closed it.
If @wadael wishes to continue working on it, or signs the CLA we could proceed.

wadael:
Hi
Sorry for my mutism. Wont expose reasons here.

I'll need directions on the changes you require. Consider me a python noob.

No issue for me to sign the cla. How ?

You can write to me directly wadael at gmail com

Cheers

cakrit:
Hi @wadael, a button to sign the CLA will appear if you create a new PR. It's one of the automated steps. You can then either proceed yourself with the help of @ilyam8, or let him take your code to finish it. 
Thanks for your help.

wadael:
Hi @cakrit @ilyam8,
I give my current code to the project. 
Please improve and merge it.

PR #3724, opened 2018-05-14
Fix the new date handling in alarm-notify.sh

Ferroin:
The original commit unintentionally capitalized the name of the `date_format` variable in plugins.d/alarm-notify.sh, which caused the new date handling support to not work at all (but thankfully did not break any existing functionality).

This corrects that issue, so that the new date handling support works correctly.

Great example of why you should just copy whole files from the system you developed changes on into your git repo instead of manually copying the changes...


ktsaou:
oops! I didn't notice this either...

PR #3734, opened 2018-05-18
OpenVPN Plugin: Fix for IPv6 connections & ignore unauthenticated Clients

vpnable:
The regex for matching the client list did depend on an IPv4 address, so if a client was connected to the OpenVPN server via IPv6, it got ignored.

So I modified the regex to also match IPv6 addresses; and also ignored clients with the Common Name `UNDEF`, which are unauthenticated clients.

( if you get DoS-ed, the plugin currently reports hundreds of `UNDEF` clients, see https://openvpn.net/archive/openvpn-users/2004-08/msg00116.html )

### DIFF
--- a/python.d/ovpn_status_log.chart.py
+++ b/python.d/ovpn_status_log.chart.py
@@ -107,8 +107,10 @@ class Service(SimpleService):
 
         data = dict(users=0, bytes_in=0, bytes_out=0)
         for row in raw_data:
-            row = ' '.join(row.split(',')) if ',' in row else ' '.join(row.split())
-            match = self.regex['tls'].search(row)
+            columns = row.split(',') if ',' in row else row.split()
+            if columns[0] == 'UNDEF': continue # see https://openvpn.net/archive/openvpn-users/2004-08/msg00116.html
+
+            match = self.regex['tls'].search(' '.join(columns))
             if match:
                 match = match.groupdict()
                 data['users'] += 1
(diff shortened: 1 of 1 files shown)


l2isbad:
i checked your regex and for some reason it takes ~25s to process (!) one line for `py2`, and ~ 19s for `py3` with 100% 1 core usage.

```python
import re

r = re.compile(r'(?:(?:[0-9a-f]|:){1,4}(:(?:[0-9a-f]{0,4})*){1,7}|(?:\d{1,3}(?:\.\d{1,3}){3}(?::\d+)?)) (?P<bytes_in>\d+) (?P<bytes_out>\d+)')

s = "xxx,2001:db8:85a3:8d3:1319:8a2e:370:7348:47562,11933,7589,Mon May 21 20:41:45 2018"

print(r.search(' '.join(s.split(','))).groupdict())
```

```bash
xxx@yyy[~]time python test.py 
{'bytes_in': '11933', 'bytes_out': '7589'}

real	0m24.295s
user	0m24.024s
sys	0m0.268s
xxx@yyy[~]time python3 test.py 
{'bytes_in': '11933', 'bytes_out': '7589'}

real	0m18.946s
user	0m18.588s
sys	0m0.348s
```

i suggest to simplify it to 

```python
r'(?:[0-9a-f:]+|(?:\d{1,3}(?:\.\d{1,3}){3}(?::\d+)?)) (?P<bytes_in>\d+) (?P<bytes_out>\d+)'
```

what do you think?

vpnable:
Oops, thanks for checking; yes, this seems to be working fine. Updated Pull Request.

ktsaou:
merged! thank you!

PR #3747, opened 2018-05-23
 fix btrfs over bcache sector size and minor apps.plugin config changes

### LINKED ISSUE #3746: btrfs shows wrong disk space when the filesystem has sector size 4k but the logical disk sector size is 512B
I just installed an 8TB disk on a system.

The disk is reported like this:

```sh
# gdisk -l /dev/sdc
GPT fdisk (gdisk) version 1.0.3

Partition table scan:
  MBR: not present
  BSD: not present
  APM: not present
  GPT: not present

Creating new GPT entries.
Disk /dev/sdc: 15628053168 sectors, 7.3 TiB
Model: WDC WD80EFZX-68U
Sector size (logical/physical): 512/4096 bytes
Disk identifier (GUID): E222C434-6D21-439D-ADB4-9724602C1D25
Partition table holds up to 128 entries
Main partition table begins at sector 2 and ends at sector 33
First usable sector is 34, last usable sector is 15628053134
Partitions will be aligned on 2048-sector boundaries
Total free space is 15628053101 sectors (7.3 TiB)

Number  Start (sector)    End (sector)  Size       Code  Name
```

So 7.3 TB usable. I have not added any partitions. What is interesting is this:

> Sector size (logical/physical): 512/4096 bytes

I chose to use this disk with bcache. Initialized it with this:

```sh
# make-bcache --block 4k --bucket 2M -B /dev/sdc
```

This created the bcache device `/dev/bcache0`, which I formatted with btrfs:

```sh
# mkfs.btrfs -L 8TB-1 -f /dev/bcache0
```

The bug is that netdata believes the size of the disk if 8 times that:

image

This is what netdata does to find the size:

```
disk="bcache0"
btrfs total disk size in bytes = /sys/block/${disk}/size * /sys/block/${disk}/queue/hw_sector_size
```

These files have the following content:

```sh
# cat /sys/block/${disk}/size /sys/block/${disk}/queue/hw_sector_size 
15628053152
4096
```

They give 58.2 TB when multiplied.

These are all the values in the `queue` directory:

```sh
# for x in /sys/block/${disk}/queue/*; do echo "$(basename ${x}): $(cat "${x}")";  done
add_random: 0
chunk_sectors: 0
dax: 0
discard_granularity: 512
discard_max_bytes: 2199023255040
discard_max_hw_bytes: 2199023255040
discard_zeroes_data: 0
hw_sector_size: 4096
io_poll: 0
io_poll_delay: 0
iostats: 0
logical_block_size: 4096
max_discard_segments: 1
max_hw_sectors_kb: 2147483647
max_integrity_segments: 0
max_sectors_kb: 2147483647
max_segments: 256
max_segment_size: 4294967295
minimum_io_size: 4096
nomerges: 0
nr_requests: 128
optimal_io_size: 0
physical_block_size: 4096
read_ahead_kb: 128
rotational: 0
rq_affinity: 0
scheduler: none
write_cache: write back
write_same_max_bytes: 0
write_zeroes_max_bytes: 0
zoned: none
```

I don't see anything useful here.

---

I believe that the problem is that kernel reports wrong `size` for the bcache. It should have aligned the size to the sector size of this device.

To validate this:

```sh
# for x in /sys/block/${disk}/slaves/*; do echo "$(basename ${x}): size $(cat "${x}/size"), hw_sector_size: $(cat "${x}/queue/hw_sector_size")";  done
sdc: size 15628053168, hw_sector_size: 512
```

So, netdata should use the `hw_sector_size` of the slaves when there are slaves, and use the one it does today, only when there are no slaves for the underlying block devices. Though, if you carefully check the numbers, the underlying disk has a few more sectors (16) compared to the ones offered by bcache to btrfs.

To my understanding though, this is probably a kernel bug. The kernel reports wrong number of sectors (`size`) for the bcache device.

---

@Ferroin what do you think? Am I missing something?

ktsaou:
fixes #3746 (wrong btrfs space calculation when used over bcache and the underlying has 512 bytes logical sector size, but bcache has been configured for 4k sector size).

- added a few more matches to `X` apps plugin target to support browser sandboxes.
- systemd is now matched as `other`, so that its unmatched children are also reported as `other` (before this, all non-matches apps found running under systemd were reported as `system`)


ktsaou:
ok, let's merge it then.
I updated the description to reflect the commits

PR #3767, opened 2018-05-29
Remove the old obsolete web interface.

Ferroin:
General cleanup as discussed in PR #3606.

As stated by @ktsaou there, this works, but it's not really useful anymore.  Given that you have to go out of your way to access it, I'd say it's pretty likely that essentially nobody is using it these days (especially since it's multiple years out of date at this point).

In theory, this could have been part of an upcoming PR I'm preparing to get licensing info into all of Netdata's files, but unlike that this is quick and easy, and should be pretty uncontroversial (or at least, if it is controversial, it will be for different reasons).


ktsaou:
merged!
thank you!

Issue #2956, opened 2017-11-02
[RFE] send notifications repeatedly until the alert is resolved


pschiffe:
Hello,

nagios has this feature, that if there is an alert, notification email is send repeatedly every hour or so (configurable) until the alert is resolved. I would like to see this feature in netdata as well, as sometimes long lasting alerts can be lost during short lasting alerts (if there are occasional spikes e.g. during backups or so). I know that I can probably super-tinker with alerts to configure them in a way that I receive the notifications only when really necessary, but I believe that this new feature would be useful as well. I'm thinking about global configuration (on/off, how often) and per alert as well.

I've checked the documentation, but didn't find anything related. Please let me know if the feature already exists.

Thanks.

Ferroin:
I don't know of such a feature in netdata, but I agree that it would be a good thing to have, especially if there's per-alarm configuration.  Netdata is in fact the only monitoring software I know of that doesn't seem to have this feature.

ktsaou:
Yeap, this is needed.

pschiffe:
bump to increase priority. with more alerts this is getting more and more important. I'm not able to create a PR, but I can help to test, or provide any information necessary.

Ferroin:
I believe Grafana does support recurring notifications, but I don't know how easy it is to set up notifications in Grafana (I looked into Grafana bit some time ago when I was using collectd instead of netdata, but never got very deep because at the time it was more effort than it was worth for my use case).

ktsaou:
How this should work?

Just adding a `repeat: DURATION` would suffice?

Ferroin:
I like the idea of having just a simple `repeat` key, with two caveats:

* It would be _really_ nice to have some global config for this that can be overridden by individual alarms so that people only have to change things in one place if they just want this globally.
* It would also be kind of nice to be able to specify a different frequency for `CRITICAL` alarms, as those are usually more urgent and it therefore makes sense to have them pester people more.  Having this part default to just matching the interval for `WARNING` messages makes sense though.

ktsaou:
ok, I started coding this.

I added `repeat: [DEFAULT_VALUE] [warning VALUE] [critical VALUE]`, where value is a duration (i.e. `1m`, `10s`, `1d`, etc so it is quite flexible - it also support fractionals).

I also added 2 netdata.conf options under `[health]` for providing default repeat rates for all alarms.

The only thing missing is to repeatedly send the notifications. I have the following question:

[...]

PR #3768, opened 2018-05-29
Add the option to thread email notifications.

Ferroin:
This adds a new configuration option in health_alarm_nofity.conf to insert In-Reply-To and References headers in email notifications.  The contents of these headers are composed in a deterministic way from the chart name, alarm name, and host name.

This allows email clients to sort netdata alert emails into groups based on the combination of chart, alarm, and host, which can help significantly when dealing with a large number of alerts.  It also provides something for scripts that might process and filter thes emails to match on that's a bit easier to parse than the subject line.

We default to not enabling this so that people who have their e-mail clients configured for threaded reading don't see an unexpected change in behavior.

Strictly speaking, this is technically a misuse of the headers, as they are supposed to only reference the Message-ID header of an existing e-mail.  However, I know of no e-mail clients that enforce standards compliant behavior (as it would break threading picked up partway through a thread), so we should be safe.

Minimally tested (using the test-alert.sh script and running for a short while on a server where I synthetically induced lots of alerts).

I've been thinking about doing this for a while now, as it _really_ simplifies keeping tracking of alerts via email (open the thread, and the last message is the last state the alarm was in, so earlier messages only matter for historical reasons).  I was reminded of this by recent activity in #2956, and after mentioning it there, decided to just take the time to implement it.

### DIFF
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -171,6 +171,14 @@ DEFAULT_RECIPIENT_EMAIL="root"
 # autodetected from the environment.
 #EMAIL_CHARSET="UTF-8"
 
+# You can also have netdata add headers to the message that will
+# cause most e-mail clients to treat all notifications for a given
+# chart+alarm+host combination as a single thread.  This can help
+# simplify tracking of alarms, as it provides an easy wway for scripts
+# to corelate messages and also will cause most clients to group all the
+# messages together.  THis is off by default.
+#EMAIL_THREADING="YES"
+
 
 #------------------------------------------------------------------------------
 # pushover (pushover.net) global notification options
--- a/plugins.d/alarm-notify.sh
+++ b/plugins.d/alarm-notify.sh
@@ -326,6 +326,7 @@ declare -A role_recipients_custom=()
 EMAIL_SENDER=
 DEFAULT_RECIPIENT_EMAIL="root"
 EMAIL_CHARSET=$(locale charmap 2>/dev/null)
+EMAIL_THREADING=
 declare -A role_recipients_email=()
 
 # irc configs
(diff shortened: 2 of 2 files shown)


ktsaou:
nice!

I suggest to enable it by default and let people disable it if they have issues with it. This way, all new users will get threaded emails without any additional action. Also, I don't think this change will introduce any issues for old users. It is just a presentation.

What do you think?

pschiffe:
Great! I think as well that enabling it by default would be more beneficial than harmful.

Ferroin:
I'm largely indifferent about whether it should be enabled by default or not.  My primary reasoning for having it disabled is that most mail readers that have threading support will display threads collapsed by default, so there would be a significant visible difference in how the messages are listed.  I personally don't mind such changes myself, but though there might be people out there who do.

ktsaou:
> I'm largely indifferent about whether it should be enabled by default or not. My primary reasoning for having it disabled is that most mail readers that have threading support will display threads collapsed by default, so there would be a significant visible difference in how the messages are listed. I personally don't mind such changes myself, but though there might be people out there who do.

You are right, however all threading clients also do something else. They change the priority (the sorting) of the entire thread based on the last message received. So, if the thread keeps coming on top (or bottom depending on the client), there is something new in it. I think the number of incidents per alarm is not that important. After all, my intention was to eliminate them (this is why I added dynamic hysteresis, etc).

Anyway, not such a big deal. I think we can change it later too.

Personally, I think this is the way it should work in the first place, so I would enable it by default.

Ferroin:
Perhaps leave it off by default for now, and switch it after the next release (with something in the release notes saying that it will become default behavior)?  I'm largely just trying to head off the inevitable 'Who moved my cheese?' issues that you get when something suddenly changes in the user interface.

ktsaou:
ok, let's do it that way.

PR #3772, opened 2018-05-31
Licensing cleanup.

Ferroin:
This is an attempt at a global cleanup of the licensing info for Netdata, to make it more consistent with standards for licensing (like actually having each file tagged), and hopefully make it more friendly to FOSSA so that PR #3606 can actually be merged.

It takes the approach of using SPDX style licensing tags in each file.

The first two commits are general cleanup to make things a bit more consistent with how other projects that use SPDX style licensing do things.  The third commit is just a general cleanup I noticed while working on this.

The fourth commit is the big one here.  It tries to add a GPLv3+ SPDX license tag to every file that is part of Netdata and not redistributed from an external third party (and supports comments of some form, because you kind of can't tag files otherwise).  It specifically does not add such tags to most config files (they are technically GPL-3.0 licensed, but almost nobody tags them no matter what licensing scheme they use), and avoids some files that I wasn't trivially able to determine the provenance of.  I may have missed a few files also.  This particular commit looks huge, but it's actually just one or two line changes in every file it touches.

Beyond this, there are two other things to potentially take care of:

* A lot of the copyrights and authorship tags are out of date.  Ideally, these need updated, but doing so is not quite as trivial as just adding license tags.  Assuming @ktsaou wants this done, I can do it, but it will take a while, as I'll have to go through the Git history manually to do it (and as a result it will probably be a separate PR).
* Ideally, we should have third-party code tagged as well (In addition to the boilerplate already there) and have the associated licenses in the `LICENSES` directory, but I wanted to make sure the rest of this was acceptable first.  Assuming it is, I'll take care of this point and either add it as a part of this PR or a new one.


ktsaou:
This is nice!

I agree with everything you wrote above.

Can you check why travis fails? Something broke...

Ferroin:
I apparently forgot to remove the COPYING file from the build infrastructure.  I'll add that to the list of things to include, though I may not get back to this until next week.

Ferroin:
And also I apparently mistagged `avl.c`, as I just used a simple script to add the tags under `src`.

Ferroin:
OK, commits pushed for the libavl code and fixing the build.  I'm going to get started on tagging the third-party code (since that will be easier to do quickly than figuring out the licensing for the other files this isn't touching yet), but probably won't have that done until Monday.

Ferroin:
OK, there's the commits for all the third-party stuff explicitly called out in `LICENSE-REDISTRIBUTED.md`.

Ferroin:
OK, I've gotten almost everything now, the remaining files that don't have tags fall into one of eight categories:

* The Dockerfiles, which I've opened an issue in the upstream repo asking about the license for.
* A few things in the tests directory (which I accidentally added newlines to the end of somehow, despite not tagging them) which I'm not quite sure of the source for.
* A few of the m4 macros, which I can't find conclusive licensing information for.
* File formats I don't recognize, and therefore can't tag properly.
* Files that just don't normally get tagged with a license (config files for example).
* Empty files (AFAIUI, these don't need to be tagged).
* Binary files, which can't be tagged.
* File formats which don't support comments (mostly JSON), and therefore can't be tagged.

Ideally, first four items need addressed, but for now this should be sufficient.  Ideally, I would like to see how FOSSA reacts to this (it's supposed to be fixing most of the issues there so we can properly integrate after all), but I'm not quite sure how to do that.

Ferroin:
@ktsaou Unless you see any issues here, this is probably ready to merge.

ktsaou:
@Ferroin you have done a magnificent job!
You are awesome!

ktsaou:
@Ferroin can you see netdata on fossa.io?

ktsaou:
hm... this had another side-effect. Github uses the LICENSE file to detect the license of the project.
Now we appear as unknown.
Check the netdata license at CNCF: https://landscape.cncf.io/grouping=no&license=open-source&selected=netdata&sort=stars

Ferroin:
Odd, it should be following the symlink.  I'd actually say this is a bug with GitHub itself TBH.

Ferroin:
FWIW, it will also check LICENSE.md, but ours doesn't look like any of the standard license texts, so it isn't pulling anything from there.

ktsaou:
Any ideas how to fix it?

Ferroin:
The fact that it doesn't follow symlinks should be reported to the GitHub developers, not sure how to do that though.

As far as a short term fix, if we just replace LICENSE.md with the GPL-3.0 text, GitHub should detect the license correctly.

ktsaou:
Can you try it?

Ferroin:
Yeah, i'll have a PR up shortly.

Issue #3752, opened 2018-05-25
python plugins got behind by 5 seconds


ktsaou:
@l2isbad I have a server on Azure running 1.9.0-119-gb94d726, with uptime 140 days.
On this server, all python plugins are behind in clock by 4-5 seconds.

An identical server (they are clustered, same uptime, same netdata version), python plugins are behind by 1 second.

The charts' clock is controlled by the collector, using the microseconds argument to `BEGIN` lines.
Is it possible that time is lost due to some rounding of microseconds in python?
Could you please check?

ktsaou:
I also checked a server running v1.10.0-131-g95e50ae_rolling
It seems python plugins are also behind by 1 sec.

In all these servers python 2 is used.

Ferroin:
I've seen drift like this too, though to a much lesser degree, and I don't remember which version of Python was being used.

l2isbad:
Interesting

> s it possible that time is lost due to some rounding of microseconds in python?

`time.time()`

`Return the current time in seconds since the Epoch. Fractions of a second may be present if the system clock provides them.`

Function return 6 or 7 numbers after the decimal point on my PCs.

https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/FrameworkServices/SimpleService.py#L178

Substraction result is multiplied by `1e6`  before `int()`, so numbers are not lost.

l2isbad:
Maybe the problem with `next_run` calculation. But looks like it is the same as https://github.com/firehol/netdata/wiki/External-Plugins#writing-plugins-properly

Ferroin:
OK, just took a look at the code.  I'm betting it's because it's using `time.sleep()`.  This function isn't reliable for precise timing purposes in Python.  Realistically, the `time.sleep()` function behaves like this:

1. If we get a signal before the time is up (and we're running on a Python version before 3.5 or the signal handler throws an exception), wake up anyway.
2. Otherwise, sleep for at least as long as asked for, and then tell the scheduler that we're ready to run again.  Note that this _won't_ preempt anything else that's running, it just tells the scheduler to consider this task as something that can run next time it has to pick a task to run.

In other words, you may sleep less than the requested time, but usually you'll sleep slightly more than the time you asked for, with the exact positive offset depending on how busy the rest of the system is, and the way we're using `time.sleep()` will cause that offset to build up over time.

[...]

PR #3776, opened 2018-06-01
Py2 monotonic time

l2isbad:
@ktsaou 

https://github.com/firehol/netdata/issues/3752

i have a question about `clock_gettime(CLOCK_MONOTONIC_RAW, ctypes.pointer(t)) != 0`

if i understand it right i have to [check](https://github.com/l2isbad/netdata/blob/py2_monotonic_time/python.d/python_modules/bases/monotonic.py#L29) it [only once on start](https://github.com/l2isbad/netdata/blob/py2_monotonic_time/python.d/python_modules/bases/FrameworkServices/SimpleService.py#L12-L14), not on every call, right?

### DIFF
--- a/plugins.d/python.d.plugin
+++ b/plugins.d/python.d.plugin
@@ -15,10 +15,6 @@ from re import sub
 from sys import version_info, argv
 from time import sleep
 
-try:
-    from time import monotonic as time
-except ImportError:
-    from time import time
 
 PY_VERSION = version_info[:2]
 PLUGIN_CONFIG_DIR = os.getenv('NETDATA_CONFIG_DIR', os.path.dirname(__file__) + '/../../../../etc/netdata') + '/'
--- a/python.d/Makefile.am
+++ b/python.d/Makefile.am
@@ -80,6 +80,7 @@ dist_bases_DATA = \
     python_modules/bases/collection.py \
     python_modules/bases/loaders.py \
     python_modules/bases/loggers.py \
+    python_modules/bases/monotonic.py \
     $(NULL)
 
 bases_framework_servicesdir=$(basesdir)/FrameworkServices
--- a/python.d/python_modules/bases/FrameworkServices/SimpleService.py
+++ b/python.d/python_modules/bases/FrameworkServices/SimpleService.py
@@ -4,11 +4,14 @@
 # Author: Ilya Mashchenko (l2isbad)
 
 from threading import Thread
+from time import sleep
 
 try:
-    from time import sleep, monotonic as time
+    from time import monotonic as time
 except ImportError:
-    from time import sleep, time
+    from bases.monotonic import AVAILABLE, time
+    if not AVAILABLE:
+        from time import time
 
 from bases.charts import Charts, ChartError, create_runtime_chart
 from bases.collection import OldVersionCompatibility, safe_print
--- a/python.d/python_modules/bases/monotonic.py
+++ b/python.d/python_modules/bases/monotonic.py
@@ -2,7 +2,7 @@
 import ctypes
 # import os
 
-CLOCK_MONOTONIC_RAW = 4  # see <linux/time.h>
+CLOCK_MONOTONIC = 1  # see <linux/time.h>
 
 
 class _Timespec(ctypes.Structure):
@@ -19,7 +19,7 @@ _clock_gettime.argtypes = [ctypes.c_int, ctypes.POINTER(_Timespec)]
 
 def time():
     t = _Timespec()
-    if _clock_gettime(CLOCK_MONOTONIC_RAW, ctypes.pointer(t)) != 0:
+    if _clock_gettime(CLOCK_MONOTONIC, ctypes.pointer(t)) != 0:
         # errno_ = ctypes.get_errno()
         # raise OSError(errno_, os.strerror(errno_))
         return None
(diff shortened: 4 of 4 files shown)


ktsaou:
> if i understand it right i have to check it only once on start, not on every call, right?

We use `clock_gettime()` to read the time. So, to get the time in `t` you have to make that check every time.

In C we use `CLOCK_MONOTONIC = 1`, not `CLOCK_MONOTONIC_RAW = 4`. Which is trange, since the man page says `CLOCK_MONOTONIC` is affected by ntp adjustments...

Check https://github.com/firehol/netdata/blob/master/src/clocks.c for the clocks we use in C.

Check also my last comment on #3752

ktsaou:
About the check I think I don't understand how this works:

```python

AVAILABLE = bool(time())

from bases.monotonic import AVAILABLE, time
    if not AVAILABLE:
        from time import time
```

But I understand you use this to check if the `time()` is available. I guess only once is enough...

l2isbad:
[original code](https://stackoverflow.com/a/1205762)

```python
import ctypes, os

CLOCK_MONOTONIC_RAW = 4 # see <linux/time.h>

class timespec(ctypes.Structure):
    _fields_ = [
        ('tv_sec', ctypes.c_long),
        ('tv_nsec', ctypes.c_long)
    ]

librt = ctypes.CDLL('librt.so.1', use_errno=True)
clock_gettime = librt.clock_gettime
clock_gettime.argtypes = [ctypes.c_int, ctypes.POINTER(timespec)]

def monotonic_time():
    t = timespec()
    if clock_gettime(CLOCK_MONOTONIC_RAW , ctypes.pointer(t)) != 0:
        errno_ = ctypes.get_errno()
        raise OSError(errno_, os.strerror(errno_))
    return t.tv_sec + t.tv_nsec * 1e-9
```

i am curious about
```python
    if clock_gettime(CLOCK_MONOTONIC_RAW , ctypes.pointer(t)) != 0:
        errno_ = ctypes.get_errno()
        raise OSError(errno_, os.strerror(errno_))
```

so `clock_gettime(CLOCK_MONOTONIC_RAW , ctypes.pointer(t))` sometimes can return not zero and this is error.

I mean under what conditions can it arise? should i check it on evey call to `monotonic_time()`  or it is enough to check once.

> In C we use CLOCK_MONOTONIC = 1, not CLOCK_MONOTONIC_RAW = 4

i see `use CLOCK_MONOTONIC_RAW==4 (since Linux 2.6.28; Linux-specific) to avoid NTP adjustments` and ` see <linux/time.h>`  in the comments

ktsaou:
> I mean under what conditions can it arise? should i check it on evey call to monotonic_time() or it is enough to check once.

from `man clock_gettime`:

```
RETURN VALUE
       clock_gettime(), clock_settime(), and clock_getres() return 0 for success, or -1 for failure
       (in which case errno is set appropriately).

ERRORS
       EFAULT tp points outside the accessible address space.

       EINVAL The clk_id specified is not supported on this system.

       EPERM  clock_settime() does not have permission to set the clock indicated.
```

So, `EPERM` is not valid for this (we don't set the time). `EINVAL` is probably the only possible error. 
I see the see code does not log anything about clock_gettime() failing. I'll add some logs there to make sure we never get such errors.

l2isbad:
> EINVAL is probably the only possible error.

nice, so looks like this PR safe to merge

ktsaou:
ok. I suggest to use `CLOCK_MONOTONIC` to match what netdata already does.

I think I get why we have used `CLOCK_MONOTONIC` instead of `CLOCK_MONOTONIC_RAW`:

netdata gets the realtime clock at the beginning and then it increments it by the microseconds supplied. If we don't get ntp adjustments into account, then the charts will never get aligned with wall clock. All ntp adjustements will be missing from it, so after some time the charts will show the wrong time.

The only way for the charts to have the right time is to use `CLOCK_MONOTONIC`. This means that netdata time can get out of sync with wall clock, only if someone manually adjusts the clock.

So, my suggestion is to use `CLOCK_MONOTONIC`. If this is wrong, we have to change it everywhere.

ktsaou:
@l2isbad have you tested this under both python v2 and v3?
Is it ok to merge?

l2isbad:
These changes only affect `py2`. `py3` has `monotonic` in `time` package (I don't know which monotonic it uses, raw or not)

ktsaou:
merged it

ktsaou:
thank you!

amichelic:
Commit 20bbce31336b860af48a918046b7defd31be5ee8 seems to break Python plugins on FreeBSD; at least on this system:
```$ uname -mrs 
FreeBSD 10.4-RELEASE-p10 amd64
$ python --version
Python 2.7.15
```

After updating to the latest git version, python plugins no longer returned data (neither when running from netdata, nor when running directly in debug mode) (this isn't 100% true - once in a while (several minutes) a record is returned). I narrowed the problem down to the mentioned commit, reverting it makes the plugins run fine again.

Researching a bit more revealed that FreeBSD is using different constants for CLOCK_MONOTONIC (from /usr/include/time.h): 
```#define CLOCK_VIRTUAL   1
//...
#define CLOCK_MONOTONIC 4
```

So the Python code is now using CLOCK_VIRTUAL, which is defined as "for time that increments only when the CPU is running in user mode on behalf of the calling process". The old value 4 was actually good on FreeBSD, because it referenced CLOCK_MONOTONIC.

ktsaou:
oops! @l2isbad check this.

PR #3777, opened 2018-06-01
Add rocket.chat notifications

Hermsi1337:
I implemented notifications for rocketchat. 
My manual tests went very fine.

Nevertheless, I'm not quite sure if I got every possible config parameter. 
Therefore I would really appreciate feedback of you guys.

Regards,
Hermsi

### DIFF
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -332,6 +332,25 @@ SLACK_WEBHOOK_URL=""
 # roles):
 DEFAULT_RECIPIENT_SLACK=""
 
+#------------------------------------------------------------------------------
+# rocketchat (rocket.chat) global notification options
+
+# multiple recipients can be given like this:
+#                  "CHANNEL1 CHANNEL2 ..."
+
+# enable/disable sending slack notifications
+SEND_ROCKETCHAT="YES"
+
+# Login to rocket.chat and create an incoming webhook. You need only one for all
+# your netdata servers (or you can have one for each of your netdata).
+# Without it, netdata cannot send rocketchat notifications.
+ROCKETCHAT_WEBHOOK_URL=""
+
+# if a role's recipients are not configured, a notification will be send to
+# this slack channel (empty = do not send a notification for unconfigured
+# roles):
+DEFAULT_RECIPIENT_ROCKETCHAT=""
+
 
 #------------------------------------------------------------------------------
 # alerta (alerta.io) global notification options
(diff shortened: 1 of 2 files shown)


conf.d/health_alarm_notify.conf
```diff
@@ -332,6 +332,25 @@ SLACK_WEBHOOK_URL=""
...
 DEFAULT_RECIPIENT_SLACK=""
 
+#------------------------------------------------------------------------------
+# rocketchat (rocket.chat) global notification options
+
+# multiple recipients can be given like this:
+#                  "CHANNEL1 CHANNEL2 ..."
+
+# enable/disable sending rocketchat notifications
+SEND_ROCKETCHAT="YES"
```
Ferroin: This should probably default to being off like everything else does

Ferroin:
Beyond the one comment I made about default behavior, this looks fine to me.

conf.d/health_alarm_notify.conf
Hermsi1337: I took example on the SEND_SLACK variable in the same file:
https://github.com/firehol/netdata/blob/master/conf.d/health_alarm_notify.conf#L322

It's also YES:
`SEND_SLACK="YES"`

Or am I missing a point?

Ferroin: Huh, I actually hadn't seen that slack notifications default to on.  That should probably be changed.

The big thing here though is that the check to see if a notification type is enabled happens very early in `alarm-notify.sh`, so turning something off by setting `SEND_*="NO"` avoids almost _all_ processing relating to that notification mechanism, while doing it by leaving the recipients empty has to go through a lot more processing before it decides to do nothing.  IOW, having something disabled with the `SEND_*` variables improves efficiency over disabling it with empty recipients.

Ferroin: OK, NVM, I had forgotten that I had updated all my systems so that all the `SEND_*` variables are set to NO except the ones I want (for the above mentioned performance reasons.

GIven that what you have matches everything else (except the syslog one, but that one really does need to default to being off), you can just ignore my above comments.

Hermsi1337:
Thank you for your approval @Ferroin ! 
I hop we'll see a new release containing these changes soon. :3

ktsaou:
nice work!

I am merging this, but you also have to create a wiki page for it.

klainte:
Hi,
Just wrote a wiki page as you requested (https://github.com/firehol/netdata/wiki/RocketChat-notifications), let me know if that's fine. I did not update the sidebar menu, as this functionality has not been released yet.
I tested the notifications too, working just fine, thanks for the feature !

PR #3785, opened 2018-06-04
Added mcrcon python module to Makefile.am.

Ferroin:
Pretty simple change.  Apparently the fille gets installed anyway for some reason, so technically nothing is broken, it's just good to have this for completeness.


Ferroin:
Correction, it's not getting installed, the CI build just happened to work because it doesn't check that modules run once installed, and it happened to work on my system because I had manually copied in the file on that system while testing.

PR #3786, opened 2018-06-05
Add a plugin for monitoring BOINC clients.

Ferroin:
This adds a python.d module for monitoring Berkeley Open Infrastructure Network Computing clients.  This is the framework utilized by a vast majority of publicly run distributed computing projects (including
SETI@Home and World Community Grid).

The first two commits add a third-party module used to do the actual RPC with the BOINC client.  The original code for this is from https://github.com/MestreLion/boinc-indicator.  That project is unfortunately dead though, so the local changes I've made are not likely to ever get merged back into the upstream code.  For the record, the my changes for this specific code are:

* Merging the low-level and high-level RPC interfaces into one file so it's easier to track.
* Cleaning up the high-level interface a bit so it has a saner way to pass in the hostname and port.
* Removing the superfluous `if __name__ == 'main':` section at the end, since this won't ever be run by itself.
* Silencing the complaints the original code printed about missing attributes that were expected to be missing.

The second commit adds the BOINC module itself.

The module  only tracks the number of tasks in various task states on the system, as almost everything else is trivial to track using other existing plugins (such as apps.plugin).  It utilizes the same RPC mechanism that the official GUI management tool for BOINC clients uses.

In most cases, the data this tracks is not hugely interesting (it doesn't change very often, and generally doesn't change very much on any given system), but there are a handful of situations that it lets us provide alerts for that are of particular interest to users running BOINC on headless systems (namely computational errors, reporting failures, and running out of tasks to run).

Internally, BOINC has 3 state-machines that a given task will advance through, the task state, scheduling state, and process state.  The task state tracks the high-level status of the task (is it new, ready to run, finished running and being reported, hit a computational error, etc). the scheduling state tracks whether the scheduling algorithm says the task can run or not, and the process state tracks the status of the
associated process for the task that does the actual computing.  This module provides charts tracking task counts in each state for each of these state machines, as well as one counting total number of local
tasks, and how many tasks are 'active' (that is, how many are being considered for processing).

We also provide a set of default alarms to alert on compute errors, upload/reporting failures, an empty local task queue, and a low number of active tasks.

Here's a screenshot from the graphs right after a couple of tasks finished processing and the client issued a scheduling request:

boinc-plugin

Does not show any dashboard info because I took it before I got the dashboard info displaying correctly.

### DIFF
--- a/python.d/boinc.chart.py
+++ b/python.d/boinc.chart.py
@@ -139,6 +139,10 @@ class Service(SimpleService):
 
     def is_alive(self):
         if (not self.alive) or \
+        # TODO: This final check should ideally not be needed, but the
+        # boinc_client code does not currenlty handle remote disconnects
+        # gracefully in a threaded environment.  This is the only thing
+        # making this module Linux specific.
            self.client.rpc.sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_INFO, 0) != 1:
             return self.reconnect()
         return True
--- a/python.d/boinc.chart.py
+++ b/python.d/boinc.chart.py
@@ -138,11 +138,11 @@ class Service(SimpleService):
         return self.connect()
 
     def is_alive(self):
-        if (not self.alive) or \
         # TODO: This final check should ideally not be needed, but the
         # boinc_client code does not currenlty handle remote disconnects
         # gracefully in a threaded environment.  This is the only thing
         # making this module Linux specific.
+        if (not self.alive) or \
            self.client.rpc.sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_INFO, 0) != 1:
             return self.reconnect()
         return True
--- a/python.d/boinc.chart.py
+++ b/python.d/boinc.chart.py
@@ -128,7 +128,7 @@ class Service(SimpleService):
 
     def connect(self):
         self.client.connect()
-        self.alive = self.client.connected and self.client.authorized:
+        self.alive = self.client.connected and self.client.authorized
         return self.alive
 
     def reconnect(self):
--- a/python.d/boinc.chart.py
+++ b/python.d/boinc.chart.py
@@ -121,9 +121,6 @@ class Service(SimpleService):
         self.alive = False
 
     def check(self):
-        if platform.system() != 'Linux':
-            self.error('Only supported on Linux.')
-            return False
         return self.connect()
 
     def connect(self):
@@ -138,12 +135,7 @@ class Service(SimpleService):
         return self.connect()
 
     def is_alive(self):
-        # TODO: This final check should ideally not be needed, but the
-        # boinc_client code does not currenlty handle remote disconnects
-        # gracefully in a threaded environment.  This is the only thing
-        # making this module Linux specific.
-        if (not self.alive) or \
-           self.client.rpc.sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_INFO, 0) != 1:
+        if not self.alive:
             return self.reconnect()
         return True
 
--- a/python.d/boinc.chart.py
+++ b/python.d/boinc.chart.py
@@ -6,8 +6,6 @@
 import platform
 import socket
 
-from copy import deepcopy
-
 from bases.FrameworkServices.SimpleService import SimpleService
 
 from third_party import boinc_client
@@ -127,10 +125,7 @@ class Service(SimpleService):
             self.error('Only supported on Linux.')
             return False
         self.connect()
-        if self.client.connected and self.client.authorized:
-            self.alive = True
-        else:
-            self.alive = False
+        self.alive = self.client.connected and self.client.authorized:
         return self.alive
 
     def connect(self):
@@ -142,10 +137,7 @@ class Service(SimpleService):
         except socket.error:
             pass
         self.client.connect()
-        if self.client.connected and self.client.authorized:
-            self.alive = True
-        else:
-            self.alive = False
+        self.alive = self.client.connected and self.client.authorized:
         return self.alive
 
     def is_alive(self):
@@ -157,7 +149,7 @@ class Service(SimpleService):
     def _get_data(self):
         if not self.is_alive():
             return None
-        data = deepcopy(_DATA_TEMPLATE)
+        data = dict(_DATA_TEMPLATE)
         results = []
         try:
             results = self.client.get_tasks()
--- a/python.d/boinc.chart.py
+++ b/python.d/boinc.chart.py
@@ -124,21 +124,18 @@ class Service(SimpleService):
         if platform.system() != 'Linux':
             self.error('Only supported on Linux.')
             return False
-        self.connect()
-        self.alive = self.client.connected and self.client.authorized:
-        return self.alive
+        return self.connect()
 
     def connect(self):
         self.client.connect()
-
-    def reconnect(self):
-        try:
-            self.client.disconnect()
-        except socket.error:
-            pass
-        self.client.connect()
         self.alive = self.client.connected and self.client.authorized:
         return self.alive
+
+    def reconnect(self):
+        # The client class itself actually disconnects existing
+        # connections when it is told to connect, so we don't need to
+        # explicitly disconnect when we're just trying to reconnect.
+        return self.connect()
 
     def is_alive(self):
         if (not self.alive) or \
--- a/python.d/boinc.chart.py
+++ b/python.d/boinc.chart.py
@@ -1,0 +1,178 @@
+# -*- coding: utf-8 -*-
+# Description: BOINC netdata python.d module
+# Author: Austin S. Hemmelgarn (Ferroin)
+# SPDX-License-Identifier: GPL-3.0+
+
+import platform
+import socket
+
+from copy import deepcopy
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+from third_party import boinc_client
+
+
+ORDER = ['tasks', 'states', 'sched_states', 'process_states']
+
+CHARTS = {
+    'tasks': {
+        'options': [None, 'Overall Tasks', 'tasks', 'boinc', 'boinc.tasks', 'line'],
+        'lines': [
+            ['total', 'Total', 'absolute', 1, 1],
+            ['active', 'Active', 'absolute', 1, 1]
+        ]
+    },
+    'states': {
+        'options': [None, 'Tasks per State', 'tasks', 'boinc', 'boinc.states', 'line'],
+        'lines': [
+            ['new', 'New', 'absolute', 1, 1],
+            ['downloading', 'Downloading', 'absolute', 1, 1],
+            ['downloaded', 'Ready to Run', 'absolute', 1, 1],
+            ['comperror', 'Compute Errors', 'absolute', 1, 1],
+            ['uploading', 'Uploading', 'absolute', 1, 1],
+            ['uploaded', 'Uploaded', 'absolute', 1, 1],
+            ['aborted', 'Aborted', 'absolute', 1, 1],
+            ['upload_failed', 'Failed Uploads', 'absolute', 1, 1]
+        ]
+    },
+    'sched_states': {
+        'options': [None, 'Tasks per Scheduler State', 'tasks', 'boinc', 'boinc.sched', 'line'],
+        'lines': [
+            ['uninit_sched', 'Uninitialized', 'absolute', 1, 1],
+            ['preempted', 'Preempted', 'absolute', 1, 1],
+            ['scheduled', 'Scheduled', 'absolute', 1, 1]
+        ]
+    },
+    'process_states': {
+        'options': [None, 'Tasks per Process State', 'tasks', 'boinc', 'boinc.process', 'line'],
+        'lines': [
+            ['uninit_proc', 'Uninitialized', 'absolute', 1, 1],
+            ['executing', 'Executing', 'absolute', 1, 1],
+            ['suspended', 'Suspended', 'absolute', 1, 1],
+            ['aborting', 'Aborted', 'absolute', 1, 1],
+            ['quit', 'Quit', 'absolute', 1, 1],
+            ['copy_pending', 'Copy Pending', 'absolute', 1, 1]
+        ]
+    }
+}
+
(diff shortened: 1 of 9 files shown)


python.d/boinc.chart.py
```diff
@@ -0,0 +1,178 @@
...
+    def is_alive(self):
+        if (not self.alive) or \
+           self.client.rpc.sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_INFO, 0) != 1:
+            return self.reconnect()
+        return True
+
+    def _get_data(self):
+        if not self.is_alive():
+            return None
+        data = deepcopy(_DATA_TEMPLATE)
```
l2isbad: `_DATA_TEMPLATE` is one level `dict` so 
```python
dict(_DATA_TEMPLATE)
#  or _DATA_TEMPLATE.copy() 
```
is enough

python.d/boinc.chart.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def check(self):
+        if platform.system() != 'Linux':
+            self.error('Only supported on Linux.')
+            return False
+        self.connect()
+        if self.client.connected and self.client.authorized:
+            self.alive = True
+        else:
+            self.alive = False
```
l2isbad: i suggest to simplify to
```python
self.alive = self.client.connected and self.client.authorized
```

python.d/boinc.chart.py
```diff
@@ -0,0 +1,178 @@
...
+    def reconnect(self):
+        try:
+            self.client.disconnect()
+        except socket.error:
+            pass
+        self.client.connect()
+        if self.client.connected and self.client.authorized:
+            self.alive = True
+        else:
+            self.alive = False
```
l2isbad: same here

l2isbad:
@Ferroin you rock :+1:

Ferroin:
Well, looks like codeclimate figured out why the dashboard info wasn't showing up.

Also, gotta love the insanity that the Mac builds worked because of VFS case insensitivity...

python.d/boinc.chart.py
Ferroin: Good point.  That first one should be significantly more efficient too.

web/dashboard_info.js
```diff
@@ -2105,6 +2112,22 @@ netdataDashboard.context = {
 
     'unbound.cache': {
         info: 'The number of items in each of the various caches.'
+    },
+
+    'boinc.tasks': {
+        info: 'The total number of tasks and the number of active tasks.  Active tasks are those which are either currently being processed, or are partialy processed but suspended.
```
l2isbad: missing '
(`info: 'text`)

Ferroin: Hmm, I think I've found a bug in Vim's JS syntax highlighting then...

Ferroin:
OK, that last commit should address everything so far, but the dashboard info still doesn't seem to want to show up.

python.d/boinc.chart.py
```diff
@@ -0,0 +1,170 @@
...
+        self.client = boinc_client.BoincClient(host=self.host, port=self.port, passwd=self.password)
+        self.alive = False
+
+    def check(self):
+        if platform.system() != 'Linux':
+            self.error('Only supported on Linux.')
+            return False
+        self.connect()
+        self.alive = self.client.connected and self.client.authorized:
+        return self.alive
```
l2isbad: code right after platform check looks like the same as `self.reconnect` (`connect; set alive; return alive`)
i think you can `return self.reconnect()` after platform check.

(in this case `self.connect` method will not be used anywhere)

what do you think?

Ferroin: Functionally, they're equivalent.  I'm not a big fan of running code in try blocks that I know will fail though, and the way it's currently written is consistent with the layout used for this type of thing in other plugins (both spigotmc and beanstalkd do things this way).  If it should be changed, I'd almost rather:

```python
    def check(self):
        if platform.system() != 'Linux':
            self.error('Only supported on Linux.')
            return False
        return self.connect()

    def connect(self):
        self.client.connect()
        self.alive = self.client.connected and self.client.authorized:
        return self.alive

    def reconnect(self):
        try:
            self.client.disconnect()
        except socket.error:
            pass
        return self.connect()
```

That still minimizes the amount of code, while providing the same set of methods that the other modules do, and avoiding trying to run `self.client.disconnect()` during `check()`.

python.d/boinc.chart.py
```diff
@@ -0,0 +1,170 @@
...
+            self.client.disconnect()
+        except socket.error:
+            pass
+        self.client.connect()
+        self.alive = self.client.connected and self.client.authorized:
+        return self.alive
+
+    def is_alive(self):
+        if (not self.alive) or \
+           self.client.rpc.sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_INFO, 0) != 1:
```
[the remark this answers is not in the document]
Ferroin: Same as in `spigotmc.chart.py`, this is checking to see if the remote end hung up (more accurately, it's checking if the TCP socket is in the CONNECTED state).

Without either having that there, or the system being configured to use particularly unforgiving TCP shutdown settings, we'll hang when we try to send data over the socket if the remote end randomly disconnects.  Realistically, we _might_ still hang if the network connection is interrupted (as opposed to this which will properly catch the remote end gracefully disconnecting), but we can't reliably compensate for that case because these all get run in threads and therefore we can't setup signal delivery for a timeout.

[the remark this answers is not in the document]
Ferroin: Also, note that this isn't needed in other modules like the beanstalkd one because their interface libraries sanely handle this.  Unfortunately, the `boinc_client.py` code does not, and I don't have a good enough understanding of exactly how it works to update it to do so.

python.d/boinc.chart.py
l2isbad: Nice :+1: i like it

Ferroin: OK, I'll put together a commit that updates it to use that style.

l2isbad:
@Ferroin 
I see bionic client [uses rpc](https://github.com/MestreLion/boinc-indicator/blob/master/client.py#L537) thing to [connect/disconnect](https://github.com/MestreLion/boinc-indicator/blob/master/client.py#L549-L560)

rpc connect method desc

```python
    def connect(self, hostname="", port=0, timeout=0):
        ''' Connect to (hostname, port) with timeout in seconds.
            Hostname defaults to None (localhost), and port to 31416
            Calling multiple times will disconnect previous connection (if any),
            and (re-)connect to host.
```

Looks like you don't need to call `discondect` before `connect`

python.d/boinc.chart.py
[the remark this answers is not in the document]
l2isbad: Well, add a `TODO` then. This should be fixed i think :smile:

Ferroin:
Huh, I don't know how I missed that.  That greatly simplifies things.

Also, it looks like I'm having some issue on my particular setup with `dashboard_info.js` getting cached _somewhere_ (though I'm not sure where, my nginx frontend is set to not cache anything from the netdata endpoint, and I'm forcing a validating reload in Chrome when I update things).

l2isbad:
Update version
https://github.com/firehol/netdata/blob/master/web/index.html#L4406

Ferroin:
@l2isbad Ah, that makes sense.  Somewhat ironically, right as I commented about it above, it fully reloaded so I could see the most recent set of changes, and it appears to be working.

The only other thing I can see with it at this point that's a possible issue is the microchip icon, as I'm not quite certain if that's the best choice here, though I can't find any that I think fits any better (calculator maybe?).

python.d/boinc.chart.py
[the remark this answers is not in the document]
Ferroin: Agreed.

l2isbad:
Not sure about icon. Add whatever you think fits better (or looks cooler :smile: )

l2isbad:
`bionic.conf` not added to makefile

l2isbad:
btw what about `runtime` chart? how many ms it needs to get data&

Ferroin:
> boinc.conf not added to makefile

Doh!  I went to all the trouble of writing up a checklist for this stuff, and then forgot to follow it myself!

> btw what about runtime chart? how many ms it needs to get data&

Looks like about 25ms average, 50ms peak for a local BOINC client on my home server, which is actually faster than I expected given that XML processing is involved (though it's not like they even use XML sanely, it's just a flat document with relations expressed by ordering).  I've only done minimal testing for remote clients (I've only checked that it works by running the module by hand in debug mode), so i have no numbers there.

l2isbad:
Please test it with both `python` versions.
https://github.com/firehol/netdata/blob/master/plugins.d/python.d.plugin#L2

Ferroin:
Is there some way to get the `python.d.plugin` to spit out actual stack traces?  I'm finding it somewhat difficult to figure out what's going wrong on Python 3 (well, what exactly the issue is, the problem is that it's passing a string to `hashlib.md5()`, which wants a bytes-like object in Python 3, but the 'obvious' fixes don't seem to work).

Ferroin:
First thing I tried actually.  The problem is that without a stack trace I can't tell whether the error I'm getting is because that doesn't work, or is from somewhere else entirely (I'm inclined to believe it's the second case, but I can't find anywhere else that looks like an issue).

l2isbad:
`/usr/libexec/netdata/plugins.d/python.d.plugin 1 debug trace mysql`

`trace` option adds traceback to the logger
https://github.com/firehol/netdata/blob/master/python.d/python_modules/bases/loggers.py#L52-L65

Catch exception via `try/except` and `self.error` it

Ferroin:
Look like it's almost entirely string formatting issues, I should be able to get through them by the end of the day (though it will probably take me a while, I'm not good at the old string formatting syntax, so I'm going to have to look up some stuff to make sure I get things right when it's not just substituting strings in).

Ferroin:
Not completely working with Python 3 yet, but all the formatting is fixed.

Ferroin:
OK, now it works on both.  The fix is not pretty, but at least it is understandable...

Ferroin:
I'm going to look at doing a pass tomorrow to remove code from `boinc_client.py` that isn't actually being used, as some of the issues Codacy is reporting are there, and there's no point in having it if it's not used.

Of the other Codacy issues, we can ignore the complaint about MD5 (that's what BOINC uses for the auth token, so we're stuck using it), and I'll look into whether the `xml.eltree.ElementTree` complaints can safely be handled without breaking things.

Ferroin:
Also, now that I've looked at things again, I may have to tweak the alarms a bit, state changes are slow, so `sum` doesn't' exactly work as intended (I just had 2 compute errors on my test system (known issue with the project in question on that particular CPU), and it triggered the critical alarm state).

python.d/python_modules/third_party/boinc_client.py
```diff
@@ -0,0 +1,762 @@
...
+
+    def set_mode(self, component, mode, duration=0):
+        ''' Do the real work of set_{run,gpu,network}_mode()
+            This method is not part of the original API.
+            Valid components are 'run' (or 'cpu'), 'gpu', 'network' (or 'net')
+        '''
+        component = component.replace('cpu', 'run')
+        component = component.replace('net', 'network')
+        try:
+            reply = self.rpc.call("<set_{0}_mode><{1}/>\n<duration>{2]</duration>\n</set_{0}_mode>".format(component, RunMode.name(mode).lower(), duration))
```
l2isbad: `{2]`
typo

l2isbad:
> we'll hang when we try to send data over the socket if the remote end randomly disconnects

@Ferroin have a look
`rpc` [connect](https://github.com/MestreLion/boinc-indicator/blob/master/rpc.py#L50-L61) method has `timeout` option which is [30 by default](https://github.com/MestreLion/boinc-indicator/blob/master/rpc.py#L28)

For some reason [BoincClient has no](https://github.com/MestreLion/boinc-indicator/blob/master/client.py#L529-L557) `timeout` args and uses default timeout value.

python.d/python_modules/third_party/boinc_client.py
Ferroin: This is functionally dead code anyway, so I'm planning on removing it.

Ferroin:
I'd been planning to look at that, I just hadn't gotten around to doing so.  Based on what I've seen though, it's second granularity (which is arguably too high for us if this is going to continue having 1 second updates), and using it requires you to check the 'connected' attribute _after_ each RPC call.

l2isbad:
> it's second granularity

i think it is ok. 

> and using it requires you to check the 'connected' attribute after each RPC call

Why? 
https://github.com/Ferroin/netdata/blob/boinc-plugin/python.d/boinc.chart.py#L157-L160
is enough

Ferroin:
NVM, you're right, I was looking at the comments in the `BoincClient` class itself about `BoincClient.connected`.

It looks like the simplest option is to just cut the timeout to 1 second.  I'll need to test though to make sure it properly handles client disconnects.  I'm inclined to believe that it won't handle the situation of the remote end hanging up between RPC calls correctly based on past experience, which would unfortunately mean we still need that `getsockopts()` call.

Ferroin:
OK, looks like it all works with the timeout set down to 0 and the `getsockopts()` check removed.  With this most recent commit, it's no longer Linux specific.

ktsaou:
there is a conflict. Please solve this to have it merged.

Ferroin:
Rebased to resolve the merge conflicts.

python.d/boinc.chart.py
```diff
@@ -0,0 +1,163 @@
+# -*- coding: utf-8 -*-
+# Description: BOINC netdata python.d module
+# Author: Austin S. Hemmelgarn (Ferroin)
+# SPDX-License-Identifier: GPL-3.0+
+
+import platform
```
l2isbad: not used import

ktsaou:
so, merge?

ktsaou:
ok, let's wait @l2isbad to review once more....

l2isbad:
> so, merge?

yes, gj @Ferroin

Ferroin:
@ktsaou Looks like this is ready to go.

ktsaou:
nice!

PR #3787, opened 2018-06-06
Per-thread stats for Unbound.

Ferroin:
Unbound itself provides per-thread versions of almost all the stats we collect (everything we currently collect excpt for the 'extended' stats).  This PR adds the ability to collect this data, and displays it as a set of reduced-height charts below the main (global) charts for each Unbound instance for which it is enabled.  It provides a new configuration key `per_thread`, which if set to yes (defaults to no)  will enable this for the job.

The module queries the monitored instance of Unbound directly (using some rather hackish trickery to leverage the existing data collection code in `SocketService`) to determine how many threads to create charts for during the `check()` function, so it will not respond to changes in the number of threads at runtime, but the number of threads is something which is not likely to change much (if ever), and having this limitation means that the actual collection logic doesn't have to change at all (because we just update the dictionary used for mapping from returned data to dimensions when we create the chart definitions).

The PR also fixes some typos in the Unbound module documentation and config file, and removes the misguided override of the chart families that the code was previously doing.

### DIFF
--- a/python.d/unbound.chart.py
+++ b/python.d/unbound.chart.py
@@ -200,7 +200,6 @@ class Service(SocketService):
             longname = 'Thread {0}'.format(thread)
             charts = dict()
             order = [[], [], []]
-            statmap = dict()
             count = 0
             for item in PER_THREAD_CHARTS:
                 chartname = '{0}{1}'.format(shortname, item)
--- a/python.d/unbound.chart.py
+++ b/python.d/unbound.chart.py
@@ -3,6 +3,7 @@
 # Author: Austin S. Hemmelgarn (Ferroin)
 
 import os
+import sys
 
 from copy import deepcopy
 
@@ -62,6 +63,40 @@ EXTENDED_CHARTS = {
     }
 }
 
+# This is used as a templates for the per-thread charts.
+PER_THREAD_CHARTS = {
+    '_queries': {
+        'options': [None, '{0} Queries Processed', 'queries', 'Queries Processed', 'unbound.threads.queries', 'line'],
+        'lines': [
+            ['{0}_ratelimit', 'Ratelimited', 'absolute', 1, 1],
+            ['{0}_cachemiss', 'Cache Miss', 'absolute', 1, 1],
+            ['{0}_cachehit', 'Cache Hit', 'absolute', 1, 1],
+            ['{0}_expired', 'Expired', 'absolute', 1, 1],
+            ['{0}_prefetch', 'Prefetched', 'absolute', 1, 1],
+            ['{0}_recursive', 'Recursive', 'absolute', 1, 1]
+        ]
+    },
+    '_reqlist': {
+        'options': [None, '{0} Request List', 'items', 'Request List', 'unbound.threads.reqlist', 'line'],
+        'lines': [
+            ['{0}_reqlist_avg', 'Average Size', 'absolute', 1, 1],
+            ['{0}_reqlist_max', 'Maximum Size', 'absolute', 1, 1],
+            ['{0}_reqlist_overwritten', 'Overwritten Requests', 'absolute', 1, 1],
+            ['{0}_reqlist_exceeded', 'Overruns', 'absolute', 1, 1],
+            ['{0}_reqlist_current', 'Current Size', 'absolute', 1, 1],
+            ['{0}_reqlist_user', 'User Requests', 'absolute', 1, 1]
+        ]
+    },
+    '_recursion': {
+        'options': [None, '{0} Recursion Timings', 'seconds', 'Recursive Timings', 'unbound.threads.recursion', 'line'],
+        'lines': [
+            ['{0}_recursive_avg', 'Average', 'absolute', 1, PRECISION],
+            ['{0}_recursive_med', 'Median', 'absolute', 1, PRECISION]
+        ]
+    }
+}
+
+
 # This maps the Unbound stat names to our names and precision requiremnets.
 STAT_MAP = {
     'total.num.queries_ip_ratelimited': ('ratelimit', 1),
@@ -86,6 +121,24 @@ STAT_MAP = {
     'dnscrypt_nonce.cache.count': ('cache_dnscn', 1)
 }
 
+# Same as above, but for per-thread stats.
+PER_THREAD_STAT_MAP = {
+    '{0}.num.queries_ip_ratelimited': ('{0}_ratelimit', 1),
+    '{0}.num.cachehits': ('{0}_cachehit', 1),
+    '{0}.num.cachemiss': ('{0}_cachemiss', 1),
+    '{0}.num.zero_ttl': ('{0}_expired', 1),
+    '{0}.num.prefetch': ('{0}_prefetch', 1),
+    '{0}.num.recursivereplies': ('{0}_recursive', 1),
+    '{0}.requestlist.avg': ('{0}_reqlist_avg', 1),
+    '{0}.requestlist.max': ('{0}_reqlist_max', 1),
+    '{0}.requestlist.overwritten': ('{0}_reqlist_overwritten', 1),
+    '{0}.requestlist.exceeded': ('{0}_reqlist_exceeded', 1),
+    '{0}.requestlist.current.all': ('{0}_reqlist_current', 1),
+    '{0}.requestlist.current.user': ('{0}_reqlist_user', 1),
+    '{0}.recursion.time.avg': ('{0}_recursive_avg', PRECISION),
+    '{0}.recursion.time.median': ('{0}_recursive_med', PRECISION)
+}
+
 
 class Service(SocketService):
     def __init__(self, configuration=None, name=None):
@@ -97,25 +150,24 @@ class Service(SocketService):
         SocketService.__init__(self, configuration, name)
         self.ext = self.configuration.get('extended', None)
         self.ubconf = self.configuration.get('ubconf', None)
+        self.perthread = self.configuration.get('per_thread', False)
+        self.threads = None
         self.order = deepcopy(ORDER)
         self.definitions = deepcopy(CHARTS)
         self.request = 'UBCT1 stats\n'
         self._parse_config()
         self._auto_config()
         self.debug('Extended stats: {0}'.format(self.ext))
+        self.debug('Per-thread stats: {0}'.format(self.perthread))
         if self.ext:
             self.order = self.order + EXTENDED_ORDER
             self.definitions.update(EXTENDED_CHARTS)
         if self.unix_socket:
             self.debug('Using unix socket: {0}'.format(self.unix_socket))
-            for key in self.definitions:
-                self.definitions[key]['options'][4] = 'Local'
         else:
             self.debug('Connecting to: {0}:{1}'.format(self.host, self.port))
             self.debug('Using key: {0}'.format(self.key))
             self.debug('Using certificate: {0}'.format(self.cert))
-            for key in self.definitions:
-                self.definitions[key]['options'][4] = self.host
 
     def _auto_config(self):
         if self.ubconf and os.access(self.ubconf, os.R_OK):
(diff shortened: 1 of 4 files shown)


Ferroin:
Oh, I should probably mention that this is largely for the sake of completeness.  I have no idea whatsoever what practical use the per-thread stats have other than making sure Unbound is balancing queries across threads properly.  Overall, it's not too difficult of a change though, and saves someone from having to add it later when someone inevitably asks for it.

python.d/unbound.chart.py
```diff
@@ -140,11 +192,52 @@ def _auto_config(self):
...
+                order[count].append(chartname)
+                count += 1
+                charts[chartname] = deepcopy(PER_THREAD_CHARTS[item])
+                charts[chartname]['options'][1] = charts[chartname]['options'][1].format(longname)
+                for line in range(0, len(charts[chartname]['lines'])):
+                    charts[chartname]['lines'][line][0] = charts[chartname]['lines'][line][0].format(shortname)
+            self.order = self.order + order[0] + order[1] + order[2]
+            self.definitions.update(charts)
+            for key, value in PER_THREAD_STAT_MAP.items():
+                STAT_MAP[key.format(shortname)] = (value[0].format(shortname), value[1])
```
l2isbad: i suggest to remove all that stuff to a separate function

`PER_THREAD_CHARTS` can be a function `per_thread_charts(thread)` which returns `order` and `chart`:
```python
for tid, _ in  enumerate(self.threads):
    order, charts = per_thread_charts(tid)
    self.order.extend(order)
    self.charts.update(charts)

```

Ferroin: I can do so, though it won't be quite that simple (because I _really_ want to keep the way things get grouped right now, which means a construct like what's currently being done for the order).

Ferroin: Also, it's worth noting that the way it's structured currently makes it trivial to add new dimensions (the plugin isn't tracking _everything_ yet, just the metrics that I know will be non-zero on my own systems, but I do plan to have most of the Unbound stats available through it).  As things are structured here, adding a new per-thread dimension consists of just adding an entry to `PER_THREAD_STAT_MAP`, and a line to the appropriate chart in `PER_THREAD_CHARTS`.

l2isbad: `_generate_perthread_charts` is hard to read.

i see method adds new charts to `self.order`, `self.definitions`
my syggestion is to move all those creating stuff to a separate func like [this](https://github.com/firehol/netdata/blob/master/python.d/beanstalk.chart.py#L116-L162)

am i missing something?

Ferroin: The way the order is being handled is different.  The beanstalkd module just adds all the new charts in whatever order it finds them.  Here, we want a specific order that happens to be different from what we would get if we created the charts for each thread one at a time, and then composes them into the right order.  That's what the stuff messing with the list of lists for the local `order` variable is for.

Put a bit differently, currently, the code results in the following order for charts with two threads:

* Thread 0 Queries
* Thread 1 Queries
* Thread 0 Request List
* Thread 1 Request List
* Thread 0 Recursion Timings
* Thread 1 Recursion Timings

While adapting it to just append the charts for each thread to the order would result in this chart order:

* Thread 0 Queries
* Thread 0 Request List
* Thread 0 Recursion Timings
* Thread 1 Queries
* Thread 1 Request List
* Thread 1 Recursion Timings

The original implementation I had actually gave an order like the second case, and I found it rather hard to read and correlate (with the per-thread metrics, you're usually going to care more about comparing a given dimension to other threads than other metrics for the same thread).

I can still split most of it out to its own function, but I don't know that it's going to improve readability much.  Most likely, it would end up looking like the following (though probably with different naming, and with the second function being outside the `Service` class.

```python
    def _generate_perthread_charts(self):
        for thread in range(0, self.threads):
            charts = dict()
            order = [[], [], []]
            chartinfo, orderinfo = self._generate_perthread_charts2(thread)
            charts.update(chartinfo)
            for i in range(0, len(orderinfo)):
                order[i].append(orderinfo[i])

        self.order = self.order + order[0] + order[1] + order[2]
        self.definitions.update(charts)

    def _generate_perthread_charts2(self, threadnum):
         shortname = 'thread{0}'.format(thread)
         longname = 'Thread {0}'.format(thread)
         charts = dict()
         order = []
         count = 0

         for item in PER_THREAD_CHARTS:
            chartname = '{0}{1}'.format(shortname, item)
            order.append(chartname)
            count += 1
            charts[chartname] = deepcopy(PER_THREAD_CHARTS[item])
            charts[chartname]['options'][1] = charts[chartname]['options'][1].format(longname)

            for line in range(0, len(charts[chartname]['lines'])):
                charts[chartname]['lines'][line][0] = charts[chartname]['lines'][line][0].format(shortname)

            for key, value in PER_THREAD_STAT_MAP.items():
                STAT_MAP[key.format(shortname)] = (value[0].format(shortname), value[1])

        return charts, order
```

l2isbad: i believe netdata groups chart by families (which should be lowercase btw ) so
`[chart1:familyX, chart2:familyY, chart3:familyX, chart4:familyY]` in the dashboard will look like
```
familyX:
  chart1
  chart3
familyY:
  chart2
  chart4
 ```

Comment (l2isbad) [reply python.d/unbound.chart.py]: dimension names should be lowercase too (@ktsaou correct me if I'm wrong)

Comment (l2isbad) [reply python.d/unbound.chart.py]: <details><summary>example implementation (not tested at all)</summary>

```python

def add_keys_to_stat_map(sname):
	for key, value in PER_THREAD_STAT_MAP.items():
		STAT_MAP[key.format(sname)] = value[0].format(sname), value[1]
		
def per_thread_charts(tid):
	sname = 'thread{0}'.format(tid)
	lname = 'Thread {0}'.format(tid)
	order = [
		'{0}_queries'.format(sname),
		'{0}_reqlist'.format(sname),
		'{0}_recursion'.format(sname),
	]
	сharts = {
		order[0]: {
			'options': [None, '{0} Queries Processed'.format(lname), 'queries', 'queries processed', 'unbound.threads.queries', 'line'],
			'lines': [
				['{0}_ratelimit'.format(sname), 'Ratelimited', 'absolute', 1, 1],
				['{0}_cachemiss'.format(sname), 'Cache Miss', 'absolute', 1, 1],
				['{0}_cachehit'.format(sname), 'Cache Hit', 'absolute', 1, 1],
				['{0}_expired'.format(sname), 'Expired', 'absolute', 1, 1],
				['{0}_prefetch'.format(sname), 'Prefetched', 'absolute', 1, 1],
				['{0}_recursive'.format(sname), 'Recursive', 'absolute', 1, 1]
				]
			},
		order[1]: {
			'options': [None, '{0} Request List'.format(lname), 'items', 'request list', 'unbound.threads.reqlist', 'line'],
			'lines': [
				['{0}_reqlist_avg'.format(sname), 'Average Size', 'absolute', 1, 1],
				['{0}_reqlist_max'.format(sname), 'Maximum Size', 'absolute', 1, 1],
				['{0}_reqlist_overwritten'.format(sname), 'Overwritten Requests', 'absolute', 1, 1],
				['{0}_reqlist_exceeded'.format(sname), 'Overruns', 'absolute', 1, 1],
				['{0}_reqlist_current'.format(sname), 'Current Size', 'absolute', 1, 1],
				['{0}_reqlist_user'.format(sname), 'User Requests', 'absolute', 1, 1]
				]
			},
		order[2]: {
			'options': [None, '{0} Recursion Timings'.format(lname), 'seconds', 'recursive timings', 'unbound.threads.recursion', 'line'],
			'lines': [
				['{0}_recursive_avg'.format(sname), 'Average', 'absolute', 1, PRECISION],
				['{0}_recursive_med'.format(sname), 'Median', 'absolute', 1, PRECISION]
				]
			}
		}
	return order, сharts

class Service(SimpleService):
	...
	...
	...
	def add_per_thread_charts(self):
		for tid, _ in  enumerate(self.threads):
			add_keys_to_stat_map(tid)  # <- WARNING! this is wrong (modify global variable in threading env)
			order, charts = per_thread_charts(tid)
			self.order.extend(order)
			self.charts.update(charts)

```
</details>

Ferroin: Huh, you're right, I hadn't realized that.

Ferroin: Also, are you positive about the lowercase requirement?  A number of functionally core plugins aren't sticking to that for at least their dimensions, and it doesn't look like uppercase family names are breaking anything as far as I can see.

Ferroin: Looking further, everything I was looking at uses the name of what it's tracking, completely unmodified, as the name of the dimension.  I'll shift things to match up like that instead (which will change some existing dimension names, but make everything lowercase).

As far as the readability thing, would it be acceptable to keep `PER_THREAD_CHARTS` as a separate data structure, but use named parameters for formatting so it's obvious what's going where?  I've never been a big fan of combining templates for data structures directly into the functions that generate them.

Ferroin: The resultant function would look like:

```python
def _new_perthread_chart(thread):
    sname = 'thread{0}'.format(thread)
    lname = 'Thread {0}'.format(thread)
    charts = dict()
    order = []
    statmap = dict()

    for item in PER_THREAD_CHARTS:
        cname = '{0}{1}'.format(sname, item)
        order.append(cname)
        charts[cname] = deepcopy(PER_THREAD_CHARTS[item])
        charts[cname]['options'][1] = charts[cname]['options'][1].format(longname=lname)

        for line in range(0, len(charts[cname]['lines'])):
            charts[cname]['lines'][line][0] = charts[cname]['lines'][line][0].format(shortname=sname)

    for key, value in PER_THREAD_STAT_MAP.items():
        statmap[key.format(shortname=sname)] = (value[0].format(shortname=sname), value[1])

    return (charts, order, statmap)
```

And the caller would then merge the chart, order, and stat mapping into per-instance versions (nice catch on the `STAT_MAP` thing BTW, I had forgotten when I was originally writing that code that it could be run in parallel threads, and never tested with more than one job, so i never saw any issues).

I kind of understand what you're saying about readability, but for me the version you had posted above that embeds `PER_THREAD_CHARTS` in the function itself is almost as bad because of the amount of indentation combined with the line length (and as mentioned, I very much prefer having the template for a data structure independent of the function which generates that structure from the template).

Ferroin: OTOH, would something like this be better for the first loop in the above function?

```python
for item in PER_THREAD_CHARTS:
    cname = '{0}{1}'.format(sname, item)
    chart = deepcopy(PER_THREAD_CHARTS[item])
    chart['options'][1] = chart['options'][1].format(longname=lname)

    for line in range(0, len(chart['lines'])):
        chart['lines'][line][0] = chart['lines'][line][0].format(shortname='sname')

    charts.update(chart)
    order.append(cname)
```

That makes it a bit clearer exactly what the formatting is being applied to.  In theory, that last loop could be taken even further, but doing so would depend on the specifics of mutability and reference passing in Python, which I would rather avoid as that usually makes things even less readable.

l2isbad:
```
    for key, value in PER_THREAD_STAT_MAP.items():
        statmap[key.format(shortname=sname)] = (value[0].format(shortname=sname), value[1])
```
functionality should be outside of `_new_perthread_chart` function because it is not per thread chart (i mean function with name DO_THIS do DO_THIS_AND_THAT)

l2isbad: Well, i think your example is ok now :smile: :+1:

Ferroin: As far as the loop for the stats map, I'd almost rather just change the name for the function (`_get_perthread_info()` maybe?).  From my perspective at least, it's an integral part of the setup itself for each thread (and it uses variables composed during this function too).

l2isbad: `enumerate` is more pythonic then `range` + `len` :smile:

python.d/unbound.chart.py
```diff
@@ -11,36 +12,36 @@
...
         'lines': [
-            ['ratelimit', 'Ratelimited', 'absolute', 1, 1],
-            ['cachemiss', 'Cache Miss', 'absolute', 1, 1],
-            ['cachehit', 'Cache Hit', 'absolute', 1, 1],
-            ['expired', 'Expired', 'absolute', 1, 1],
-            ['prefetch', 'Prefetched', 'absolute', 1, 1],
-            ['recursive', 'Recursive', 'absolute', 1, 1]
+            ['ratelimit', 'ratelimited', 'absolute', 1, 1],
+            ['cachemiss', 'cache_miss', 'absolute', 1, 1],
+            ['cachehit', 'cache_hit', 'absolute', 1, 1],
```
l2isbad: spaces looks better then underscores in the dashboard

l2isbad: Maybe it is just me. Lets ask @ktsaou

Ferroin: While I would generally agree that spaces look better, on my systems there are more dimensions that use underscores than spaces, and you had commented on preferring consistency.  Also, this may just be me, but having it all lowercase with spaces looks kind of odd to me (as compared to using spaces and making things title case).

ktsaou:
merged!

Issue #3792, opened 2018-06-08
[REQUEST] Add Fleep/webhook notifications


issak:
Fleep offers integration with webhooks, it takes a simple CURL to the webhook address with "from" and "message" arguments, if would be great if it could be added to Netdata

Fleep webhook api documentation: https://fleep.io/blog/integrations/webhooks/

Ferroin:
@issak  I've just created a PR to support Fleep notifications.  If you could test it and make sure it works, that would help to get it merged (I don't use Fleep myself, so I can't really test it).

PR #3794, opened 2018-06-08
Add fleep.io notification support.

Ferroin:
This adds support for sending alarm notifications to fleep.io conversations via webhooks.  It uses the JSON interface to POST a simple message containing information about the alarm, and by default uses the host that raised the alarm as the sender name.

Recipients are specified via the last component of the webhook URL (the only part that's different for each webhook).

I have not tested this myself beyond verifying that it does not break existing notifications, as I don't use Fleep.

Fixes: #3792


ktsaou:
merged!

PR #3799, opened 2018-06-08
Add a python plugin for monitoring power supplies on Linux.

Ferroin:
This allows tracking battery statistics on Linux systems, as well as other energy storage devices that utilize the kernel's power_supply class (although i don't think there are drivers for any other type of energy storage device in the mainline kernel using this class right now).

The primary purpose of this module is twofold:

* To provide a way for battery powered IoT devices to easily alert about a low battery.
* To provide a way for all battery powered devices to alert on some easy to monitor battery health conditions.

It provides up to four charts, one which provides the remaining capacity as a percentage, and three others which report info about charge (in amp-hours), energy (in watt-hours), and voltage, each providing info about the current values, and possibly minimal and maximal values that can be used for computing battery life.

Exact support provided by each individual device varies.  Almost all provide the percentage capacity, but beyond that they may or may not support any or all of the attributes needed for the other three charts (ACPI compliant systems for example support most of the charge related ones, and two of the voltage related values, but none of the energy related ones).

Data collection is done by scanning entries in /sys/class/power_supply. One job must be created for each power supply to be monitored, and there is no autodetection (though the config includes an example that should work to monitor the main battery on most laptops).

This would probably be better off written in C as part of the core, but I ran out of patience trying to implement it there after about 16 hours spent trying to get it working at a rudimentary level.  If someone wants to re-implement this in C and submit a PR for it, let me know and I'll be happy to close this one and help with testing.

Tested on a ACPI compliant laptop.  I unfortunately don't have any other battery powered systems with more exotic hardware I can test this on, though it should Just Work (TM) for anything that uses the power_supply class (except AC power).

P.S. I think i finally managed to put on together without forgetting any files...

### DIFF
--- a/python.d/linux_power_supply.chart.py
+++ b/python.d/linux_power_supply.chart.py
@@ -3,6 +3,7 @@
 # Author: Austin S. Hemmelgarn (Ferroin)
 
 import os
+import platform
 
 from bases.FrameworkServices.SimpleService import SimpleService
 
@@ -117,6 +118,9 @@ class Service(SimpleService):
         self.types = self.configuration.get('charts', 'energy charge capacity voltage').split()
 
     def check(self):
+        if platform.system() != 'Linux':
+            self.error('Only supported on Linux.')
+            return False
         if self.supply is None:
             self.error('No power supply specified for monitoring.')
             return False
--- a/python.d/linux_power_supply.chart.py
+++ b/python.d/linux_power_supply.chart.py
@@ -1,0 +1,159 @@
+# -*- coding: utf-8 -*-
+# Description: Linux power_supply netdata python.d module
+# Author: Austin S. Hemmelgarn (Ferroin)
+
+import os
+
+from bases.FrameworkServices.SocketService import SimpleService
+
+# Everything except percentages is reported as µ units.
+PRECISION = 10 ** 6
+
+# A priority of 90000 places us next to the other PSU related stuff.
+PRIORITY=90000
+
+# We add our charts dynamically when we probe for the device attributes,
+# so these are empty by default.
+ORDER = []
+
+CHARTS = {}
+
+
+def _get_capacity_chart(syspath):
+    # Capacity is measured in percent.  We track one value.
+    options = [None, 'Capacity', '%', 'power_supply', 'power_supply.capacity', 'line']
+    lines = list()
+    attr_now = 'capacity'
+    if _get_sysfs_value(os.path.join(syspath, attr_now)) is not None:
+        lines.append([attr_now, attr_now, 'absolute', 1, 1])
+        return {'capacity': {'options': options, 'lines': lines}}, [attr_now]
+    else:
+        return None, None
+
+def _get_generic_chart(syspath, name, unit, maxname, minname):
+    # Used to generate charts for energy, charge, and voltage.
+    options = [None, name.title(), unit, 'power_supply', 'power_supply.{0}'.format(name), 'line']
+    lines = list()
+    attrlist = list()
+    attr_max_design = '{0}_{1}_design'.format(name, maxname)
+    attr_max = '{0}_{1}'.format(name, maxname)
+    attr_now = '{0}_now'.format(name)
+    attr_min = '{0}_{1}'.format(name, minname)
+    attr_min_design = '{0}_{1}_design'.format(name, minname)
+    if _get_sysfs_value(os.path.join(syspath, attr_now)) is not None:
+        lines.append([attr_now, attr_now, 'absolute', 1, PRECISION])
+        attrlist.append(attr_now)
+    else:
+        return None, None
+    if _get_sysfs_value(os.path.join(syspath, attr_max)) is not None:
+        lines.insert(0, [attr_max, attr_max, 'absolute', 1, PRECISION])
+        lines.append([attr_min, attr_min, 'absolute', 1, PRECISION])
+        attrlist.append(attr_max)
+        attrlist.append(attr_min)
+    elif _get_sysfs_value(os.path.join(syspath, attr_min)) is not None:
+        lines.append([attr_min, attr_min, 'absolute', 1, PRECISION])
+        attrlist.append(attr_min)
+    if _get_sysfs_value(os.path.join(syspath, attr_max_design)) is not None:
+        lines.insert(0, [attr_max_design, attr_max_design, 'absolute', 1, PRECISION])
+        lines.append([attr_min_design, attr_min_design, 'absolute', 1, PRECISION])
+        attrlist.append(attr_max_design)
+        attrlist.append(attr_min_design)
+    elif _get_sysfs_value(os.path.join(syspath, attr_min_design)) is not None:
+        lines.append([attr_min_design, attr_min_design, 'absolute', 1, PRECISION])
+        attrlist.append(attr_min_design)
+    return {name: {'options': options, 'lines': lines}}, attrlist
+
+def _get_charge_chart(syspath):
+    # Charge is measured in microamphours.  We track up to five
+    # attributes.
+    return _get_generic_chart(syspath, 'charge', 'µAh', 'full', 'empty')
+
+def _get_energy_chart(syspath):
+    # Energy is measured in microwatthours.  We track up to five
+    # attributes.
+    return _get_generic_chart(syspath, 'energy', 'µWh', 'full', 'empty')
+
+def _get_voltage_chart(syspath):
+    # Voltage is measured in microvolts. We track up to five attributes.
+    return _get_generic_chart(syspath, 'voltage', 'µV', 'min', 'max')
+
+# This is a list of functions for generating charts.  Used below to save
+# a bit of code (and to make it a bit easier to add new charts).
+GET_CHART = {
+    'capacity': _get_capacity_chart,
+    'charge': _get_charge_chart,
+    'energy': _get_energy_chart,
+    'voltage': _get_voltage_chart
+}
+
+# This opens the specified file and returns the value in it or None if
+# the file doesn't exist.
+def _get_sysfs_value(filepath):
+    try:
+        with open(filepath, 'r') as datasource:
+            return int(datasource.read())
+    except (OSError, IOError):
+        return None
+
+# Certain attributes (*_empty and *_empty_design) are implicitly zero
+# if not provided by the driver.  This function is used to fetch those
+# attributes.
+def _get_sysfs_value_or_zero(filepath):
+    result = _get_sysfs_value(filepath)
+    if result is None:
+        return 0
+    return result
+
+
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.definitions = dict()
+        self.order = list()
+        self.attrlist = list()
+        self.supply = self.configuration.get('supply', None)
+        if self.supply is not None:
+            self.syspath = '/sys/class/power_supply/{0}'.format(self.supply)
+        self.types = self.configuration.get('charts', 'energy charge capacity voltage').split()
+
+    def check(self):
+        if self.supply is None:
+            self.error('No power supply specified for monitoring.')
+            return False
+        if not len(self.types) > 0:
+            self.error('No attributes requested for monitoring.')
+            return False
+        if os.access(self.syspath, os.R_OK):
+            chartset = set(GET_CHART.keys()).intersection(set(self.types))
+            if len(chartset) > 0:
+                charts = dict()
+                attrlist = list()
+                for item in chartset:
+                    chart, attrs = GET_CHART[item](self.syspath)
+                    if chart is not None:
+                        charts.update(chart)
+                        attrlist.extend(attrs)
+                if len(charts) == 0:
+                    self.error('No charts can be created.')
+                    return False
+                self.definitions.update(charts)
+                self.order.extend(sorted(charts.keys()))
+                self.attrlist.extend(attrlist)
+            else:
+                self.error('No valid attributes requested for monitoring.')
+                return False
+        else:
+            self.error('Unable to access {0}'.format(self.syspath))
+            return False
+        return True
+
+    def _get_data(self):
+        data = dict()
+        for attr in self.attrlist:
+            attrpath = os.path.join(self.syspath, attr)
+            if attr.endswith('_min') or attr.endswith('_min_design') or \
+               attr.endswith('_empty') or attr.endswith('_empty_design'):
+                data[attr] = _get_sysfs_value_or_zero(attrpath)
+            else:
+                data[attr] = _get_sysfs_value(attrpath)
+        return data
--- a/python.d/linux_power_supply.chart.py
+++ b/python.d/linux_power_supply.chart.py
@@ -20,19 +20,19 @@ ORDER = []
 CHARTS = {}
 
 
-def _get_capacity_chart(syspath):
+def get_capacity_chart(syspath):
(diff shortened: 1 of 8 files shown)


python.d/linux_power_supply.chart.py
```diff
@@ -0,0 +1,159 @@
+# -*- coding: utf-8 -*-
+# Description: Linux power_supply netdata python.d module
+# Author: Austin S. Hemmelgarn (Ferroin)
+
+import os
+
+from bases.FrameworkServices.SocketService import SimpleService
```
l2isbad: > SocketService import SimpleService

^^

python.d/linux_power_supply.chart.py
```diff
@@ -0,0 +1,159 @@
...
+                self.definitions.update(charts)
+                self.order.extend(sorted(charts.keys()))
+                self.attrlist.extend(attrlist)
+            else:
+                self.error('No valid attributes requested for monitoring.')
+                return False
+        else:
+            self.error('Unable to access {0}'.format(self.syspath))
+            return False
+        return True
```
l2isbad: I suggest to rewrite `check()`
Nested `if else` is always bad. Please avoid it.
```python
if not if os.access(self.syspath, os.R_OK):
    self.error('No valid attributes requested for monitoring.')
    return None

charset = set(GET_CHART.keys()).intersection(set(self.types))
if not charset:
    self.error('No valid attributes requested for monitoring.')
    return None
...
...
...
```
Keep in mind - empty list is always `False`, non empty is always `True`. Do not use `len`.

python.d/linux_power_supply.chart.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def _get_data(self):
+        data = dict()
+        for attr in self.attrlist:
+            attrpath = os.path.join(self.syspath, attr)
+            if attr.endswith('_min') or attr.endswith('_min_design') or \
+               attr.endswith('_empty') or attr.endswith('_empty_design'):
+                data[attr] = _get_sysfs_value_or_zero(attrpath)
```
l2isbad: use `any`
```python
any([
	COND1,
	COND2,
	COND3,
])
```

python.d/linux_power_supply.chart.py
Ferroin: Gah, forgot to update that all the way when I copied over the first few lines form another plugin to save time typing...

python.d/linux_power_supply.chart.py
Ferroin: I actually did not know about the `any()` function, and had completely forgotten that `str.startswith()` could take a tuple.

Ferroin:
Fixed the aforementioned comments, and rebased to deal with the merge conflicts.

ktsaou:
nice!

Are there change requests pending?
Please make sure it is well tested before merge.

python.d/linux_power_supply.chart.py
```diff
@@ -0,0 +1,156 @@
...
+        if self.supply is None:
+            self.error('No power supply specified for monitoring.')
+            return False
+        if not self.types:
+            self.error('No attributes requested for monitoring.')
+            return False
+        if not os.access(self.syspath, os.R_OK):
+            self.error('Unable to access {0}'.format(self.syspath))
+            return False
+        chartset = set(GET_CHART.keys()).intersection(set(self.types))
```
l2isbad: not a big deal but `.keys()` call is not needed

python.d/linux_power_supply.chart.py
```diff
@@ -0,0 +1,156 @@
...
+        for item in chartset:
+            chart, attrs = GET_CHART[item](self.syspath)
+            if chart is not None:
+                charts.update(chart)
+                attrlist.extend(attrs)
+        if len(charts) == 0:
+            self.error('No charts can be created.')
+            return False
+        self.definitions.update(charts)
+        self.order.extend(sorted(charts.keys()))
```
l2isbad: same here

l2isbad:
check it with `py3` @Ferroin

Ferroin:
The system I tested on is using Python 3.6 as the global version, so unless you encountered some issue running it under Python 3.X, I actually need to check Python 2.7.

Ferroin:
Looks like it runs fine under both and reports correctly on my laptop.

As far as thorough testing though, I can't really say that anything beyond what the laptop I have reports works, since I have no other systems to test on that provide this interface.  However, there are only two places it's liable to not work on any arbitrary system, the parsing code (and if that isn't working, the driver is the issue, not us, as it's not properly implementing the API), and the code that's generating the charts (and if that doesn't work, I'll be rather amazed, since all the underlying code for that is being tested on my laptop since charge, energy, and voltage charts all use the same code for chart generation).

ktsaou:
ok. The main goal with new features is to be careful not to introduce a problem on existing setups. If the new feature does not work as expected under all circumstances, but it does not brake something else and it does not misbehave (cpu usage, memory leak, etc), I think it is ok to merge.

So merge?

Ferroin:
Yeah, given that it should be fine to merge.  That said, I do think it would be better off being written in C (it's essentially doing the same type of thing as the BTRFS monitoring code, just tracking the values of some sysfs files, the only complicated part is figuring out what attributes are present and creating charts appropriately), I just don't have the skills to do that myself.

ktsaou:
If I understand this right, it reads from `/sys/class/power_supply/<DEVICE>/`:

1. Capacity: The power supply capacity expressed as a percentage.
  * capacity\_now

2. Charge: The charge for the power supply, expressed as microamphours.
  * charge\_full\_design
  * charge\_full
  * charge\_now
  * charge\_empty
  * charge\_empty\_design

3. Energy: The energy for the power supply, expressed as microwatthours.
  * energy\_full\_design
  * energy\_full
  * energy\_now
  * energy\_empty
  * energy\_empty\_design

2. Voltage: The voltage for the power supply, expressed as microvolts.
  * voltage\_max\_design
  * voltage\_max
  * voltage\_now
  * voltage\_min
  * voltage\_min\_design

hm... I don't see of my systems have these files.
Let me open my laptop and check...

ktsaou:
This is what I have:

```sh
# tree /sys/class/power_supply/*/
/sys/class/power_supply/AC/
├── device -> ../../../ACPI0003:00
├── online
├── power
│   ├── autosuspend_delay_ms
│   ├── control
│   ├── runtime_active_time
│   ├── runtime_status
│   ├── runtime_suspended_time
│   ├── wakeup
│   ├── wakeup_abort_count
│   ├── wakeup_active
│   ├── wakeup_active_count
│   ├── wakeup_count
│   ├── wakeup_expire_count
│   ├── wakeup_last_time_ms
│   ├── wakeup_max_time_ms
│   └── wakeup_total_time_ms
├── subsystem -> ../../../../../../class/power_supply
├── type
└── uevent

/sys/class/power_supply/BAT0/
├── alarm
├── capacity
├── capacity_level
├── charge_full
├── charge_full_design
├── charge_now
├── current_now
├── cycle_count
├── device -> ../../../PNP0C0A:00
├── manufacturer
├── model_name
├── power
│   ├── autosuspend_delay_ms
│   ├── control
│   ├── runtime_active_time
│   ├── runtime_status
│   └── runtime_suspended_time
├── present
├── serial_number
├── status
├── subsystem -> ../../../../../../class/power_supply
├── technology
├── type
├── uevent
├── voltage_min_design
└── voltage_now

6 directories, 40 files
```

Ferroin:
The `AC` one is for the mains power, it can be ignored for the purposes of this (the only interesting thing there is whether it's connected or not).

The `BAT0` directory is for the first ACPI battery device, and is equivalent to what I tested against with this code.

For reference, the values in the charge, energy, and voltage files are expressed in units at a power of 10e-6 (so the 'micro' metric prefix).

Other than ensuring the appropriate scaling, the only special thing going on in the chart generation code is making sure that netdata provides a `_min` or `_empty` dimension matching each `_max` or `_full` dimension, even if the driver does not provide them, so that you can easily write an alarm that computes it's trigger threshold based on the difference between the lowest and highest values.  Beyond the chart generation, the code itself is just some trivial read and report stuff.

Ferroin:
Also, note that some ACPI devices may have more batteries (I've seen laptops that provided for up to _four_  in their firmware), and the exact naming of the battery devices is dependent on the driver being used, hence this providing the option to specify an exact name to monitor.

For configuration, it should work for the default to just track all devices with a name like `BAT*` for which the `present` file contains a `1`, and then provide an option to let people specify arbitrary names to watch.

ktsaou:
> chart generation code is making sure that netdata provides a _min or _empty dimension matching each _max or _full dimension, even if the driver does not provide them

hm... why not offering a percentage chart?

I see:

1. `capacity` is already a percentage (and should be charted)

2. `charge_full` and `charge_now` can be used to calculate `charge_percent` (only this will be charted)

3. `charge_full_design` and `charge_full` can be used to calculate `charge_health_percent` (will also be charted on the same chart with the above or we can provide a stacked chart with 3 dimensions: `full`, `empty`, `lost` - all percentages).

4. `voltage_now` can be charted alone and `voltage_min_design` can be exposed as a chart variable to be used in the alarm in case `voltage_now` goes below it.

I don't have `energy` values on my system.
What do you think?

Ferroin:
There are a couple of issues I see with that approach:

* None of the values are guaranteed to be present.  We can probably assume that the `capacity` attribute will pretty much always be present (I know of no in-tree drivers that don't provide this, even if it's computed by the driver instead of the hardware), but everything else is not guaranteed.  Just like how ACPI batteries don't provide `energy` values, some drivers provide `energy` but not `charge`, some don't provide any of the `_design` values, some may provide non-zero `_empty` and `_min` values, and there are some really stupid ones that only provide `voltage` values and use that to try and represent how charged the battery is (despite that being completely, absolutely wrong).
* The relation between actual charge, energy, or voltage and remaining capacity is usually non-linear.  In most cases, there's a slight exponential fall-off as it gets closer to zero, and the fully 'discharged' value is usually not exactly zero (for most systems, it's the point at which the battery can't sustain enough current to power all the devices in the system).  Given this, percentages for anything other than `capacity` are usually not very useful.
* Anything other than `capacity` is likely to be of primary interest to power-users only (no pun intended).  Most people are likely to just care about effective discharge rate (the slope of the `capacity` graph) and effective capacity (the value of the `capacity` graph).

Given all of that, I think it makes more sense to:
* Just provide the dimensions as-is, albeit scaled to saner exponents (I'd vote for units or possibly milli-units, not the micro-units that are exposed by the API).
* Make sure that each upper-limit dimension (`*max*` and `*full*`) provided by the hardware has a corresponding lower-limit dimension (`*min*`and `*empty*`), even if we just provide a dummy dimension that always reads zero.
* Default to just hiding all the dimensions other than the `capacity` one, which unless I'm mistaken should hide all those charts by default.
* Provide a config option to unhide any combination of `charge`, `energy`, or `voltage` dimensions each as a set (so you have all the `charge` dimensions either hidden or not, same for the other two).

My code does not do the last two, but it could be easily modified to do something similar (just not collect those dimensions).

ktsaou:
Is it practical to assume something like the following?

File|Presence|Units|Comments
:---:|:---:|:---|:---
`capacity`|always|%|Will enable `capacity` chart
`charge_now`|optional|?|Will enable `charge` chart
`charge_full`|optional but may appear if `charge_now` is also present|?|will enable `charge_percent` chart with 3 dimensions `full` and `empty`.
`charge_full_design`|optional but may appear if `charge_full` is also present|?|will enable `lost` (or `weared`) dimension of `charge_percent` chart.

Ferroin:
Regarding what's present, I think it's mostly correct, and if so that can be extended to the `energy_*` and `voltage_*` values.  I'm not 100% certain about that though, as the documentation I based this on (found [here](https://www.kernel.org/doc/Documentation/power/power_supply_class.txt)) does not state explicitly whether or not such relations hold.  The one caveat is that `capacity` _technically_ may not be present if the others are (see the last Q/A item at the bottom of the above linked document for why), but hardware that is like that is rather rare.  I can also think of the possibility of there being a `*_empty_design` value but no `*_empty` value (in which case `*_empty` should probably just report the same value as `*_empty_design`), but I don't know if anything actually does it (it's logically possible, I just can't see any code that actually does it).

As far as the dimensions, I still think we should just expose the raw values (or at least have an option to do so).  The problem is that the relation between all of the charge/energy/voltage values _other_ than the `*_now` ones is entirely dependent on the hardware, and can't reliably be generalized.  For example, comparing `charge_full` to `charge_full_design` isn't a reliable wear indicator, it just tells you how much the capacity is reduced relative to the designed capacity (this is a symptom of wear, but not the only one, and it may not be the one which makes the battery unusable first).

As far as units, as mentioned in previous comments, raw data from sysfs for these is in micro-units.  Full units make a bit more sense for end-user consumption, so just dividing everything (except `capacity`) by 1000000 will give reasonable results.  For some systems it might make sense to instead report milli-units (so 1000 instead of 1000000), but that should probably be an option if we're going to do it.

Ferroin:
@ktsaou Short term, it would be nice to get this merged.  Long term it makes sense to have it in the core, but having something that's working now as-is is still better than having nothing.

I'm going to look at getting this updated to make the charts other than `capacity` not present by default, and add an alarm configuration for the `capacity` chart to trigger when it's really low.

python.d/linux_power_supply.chart.py
```diff
@@ -0,0 +1,167 @@
...
+PRIORITY = 90000
+
+# We add our charts dynamically when we probe for the device attributes,
+# so these are empty by default.
+ORDER = []
+
+CHARTS = {}
+
+
+def _get_capacity_chart(syspath):
```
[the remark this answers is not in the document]
Ferroin: Done.

python.d/linux_power_supply.chart.py
l2isbad: please move the chart creating to a separate function

Ferroin: Said function would just get called by `check()` anyway, because `check()` succeeding is dependent both on the set being created in this line being non-empty, and on the actual set of charts that can be created out of those requested being non-empty.  In essence, generating the charts here is a side effect of what needs to be done to make sure we can collect data.

l2isbad: > Said function would just get called by check() anyway

I know, it's ok

python.d/linux_power_supply.chart.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+# Certain attributes (*_empty and *_empty_design) are implicitly zero
+# if not provided by the driver.  This function is used to fetch those
+# attributes.
+def get_sysfs_value_or_zero(filepath):
+    result = get_sysfs_value(filepath)
+    if result is None:
+        return 0
+    return result
```
l2isbad: just `return get_sysfs_value(filepath) or 0`

l2isbad: Or we can use `get_sysfs_value(filepath) or 0` where we need `get_sysfs_value_or_zero`

Ferroin:
@ktsaou If you could check this it would be appreciated.

ktsaou:
ok, approved it.

Keep in mind though netdata is about presenting meaningful data. We should interpret raw metrics and structure them in the best possible way to help users understand them.

So, this can be merged, but we should revisit it.

Ferroin:
@ktsaou That's a large part of why I updated it to just show capacity by default.  That's the single most useful metric this can provide, and it's immediately understandable to most people.  The other metrics are conditionally useful for battery-powered IoT devices, but I still contend that interpreting them is too dependent on the hardware configuration for Netdata to reasonably do anything beyond what this module is doing.

PR #3815, opened 2018-06-12
Monitor active users and sessions via logind.

Ferroin:
'logind' is a daemon used for tracking active users and sessions on a system.  By querying this, we can easily get counts of users and sessions on the system, and additionally differentiate by session type (graphical, console, or remote).

This module uses the `loginctl` command to query the the service, and does some dead simple text processing on the output to extract the required information.  In theory, we could query the DBus methods directly, but that would get us very little in terms of performance, so it's not worth the significantly higher complexity.

This module requires zero configuration.  The only thing that is required is for the user netdata is running as to be able to call `loginctl list-sessions --no-legend` and get reasonable output back.  However, it will work with any command that provides output in the same format as the above mentioned command, and the user can override the command the same way they normally do for `ExecutableService` derived modules.

`logind` itself is originally a systemd service, but Gentoo provides an independent fork of it (called `elogind`) that does not require systemd.

Example output without an icon or descriptions:

screenshot

Tested on both python 2 and 3 on multiple systems.  Average runtime on the test systems was approximately 8ms, with peak runtime being just over 20ms (though it appears to only rarely go above about 14ms).

### DIFF
--- a/python.d/logind.chart.py
+++ b/python.d/logind.chart.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+# Description: logind netdata python.d module
+# Author: Austin S. Hemmelgarn (Ferroin)
+# SPDX-License-Identifier: GPL-3.0+
+
+import os
+import sys
+
+from bases.FrameworkServices.ExecutableService import ExecutableService
+
+PRIORITY = 40000
+
+ORDER = ['sessions', 'users', 'seats']
+
+CHARTS = {
+    'sessions': {
+        'options': [None, 'Logind Sessions', 'sessions', 'sessions', 'logind.sessions', 'stacked'],
+        'lines': [
+            ['sessions_graphical', 'Graphical', 'absolute', 1, 1],
+            ['sessions_console', 'Console', 'absolute', 1, 1],
+            ['sessions_remote', 'Remote', 'absolute', 1, 1]
+        ]
+    },
+    'users': {
+        'options': [None, 'Logind Users', 'users', 'users', 'logind.users', 'stacked'],
+        'lines': [
+            ['users_graphical', 'Graphical', 'absolute', 1, 1],
+            ['users_console', 'Console', 'absolute', 1, 1],
+            ['users_remote', 'Remote', 'absolute', 1, 1]
+        ]
+    },
+    'seats': {
+        'options': [None, 'Logind Seats', 'seats', 'seats', 'logind.seats', 'line'],
+        'lines': [
+            ['seats', 'Active Seats', 'absolute', 1, 1]
+        ]
+    }
+}
+
+
+class Service(ExecutableService):
+    def __init__(self, configuration=None, name=None):
+        ExecutableService.__init__(self, configuration=configuration, name=name)
+        self.command = 'loginctl list-sessions --no-legend'
+        self.order = ORDER
+        self.definitions = CHARTS
+
+    def _get_data(self):
+        ret = {
+            'sessions_graphical': 0,
+            'sessions_console': 0,
+            'sessions_remote': 0,
+        }
+        users = {
+            'graphical': list(),
+            'console': list(),
+            'remote': list()
+        }
+        seats = list()
+        data = self._get_raw_data()
+
+        for item in data:
+            fields = item.split()
+            if len(fields) == 3:
+                users['remote'].append(fields[2])
+                ret['sessions_remote'] += 1
+            elif len(fields) == 4:
+                users['graphical'].append(fields[2])
+                ret['sessions_graphical'] += 1
+                seats.append(fields[3])
+            elif len(fields) == 5:
+                users['console'].append(fields[2])
+                ret['sessions_console'] += 1
+                seats.append(fields[3])
+
+        ret['users_graphical'] = len(set(users['graphical']))
+        ret['users_console'] = len(set(users['console']))
+        ret['users_remote'] = len(set(users['remote']))
+        ret['seats'] = len(set(seats))
+
+        return ret
--- a/python.d/logind.chart.py
+++ b/python.d/logind.chart.py
@@ -5,7 +5,7 @@
 
 from bases.FrameworkServices.ExecutableService import ExecutableService
 
-PRIORITY = 40000
+prioirty = 59999
 
 ORDER = ['sessions', 'users', 'seats']
 
--- a/python.d/logind.chart.py
+++ b/python.d/logind.chart.py
@@ -3,9 +3,6 @@
 # Author: Austin S. Hemmelgarn (Ferroin)
 # SPDX-License-Identifier: GPL-3.0+
 
-import os
-import sys
-
 from bases.FrameworkServices.ExecutableService import ExecutableService
 
 PRIORITY = 40000
--- a/python.d/logind.chart.py
+++ b/python.d/logind.chart.py
@@ -6,6 +6,7 @@
 from bases.FrameworkServices.ExecutableService import ExecutableService
 
 prioirty = 59999
+disabled_by_default = True
 
 ORDER = ['sessions', 'users', 'seats']
 
--- a/conf.d/Makefile.am
+++ b/conf.d/Makefile.am
@@ -51,6 +51,7 @@ dist_pythonconfig_DATA = \
     python.d/ipfs.conf \
     python.d/isc_dhcpd.conf \
     python.d/litespeed.conf \
+    python.d/logind.conf \
     python.d/mdstat.conf \
     python.d/memcached.conf \
     python.d/mongodb.conf \
--- a/conf.d/python.d.conf
+++ b/conf.d/python.d.conf
@@ -51,7 +51,7 @@ go_expvar: no
 # ipfs: yes
 # isc_dhcpd: yes
 # litespeed: yes
-# logind: yes
+logind: no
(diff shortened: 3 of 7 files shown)


l2isbad:
@ktsaou 
1. Please clarify chart naming convension. I see @Ferroin doesn't like lower case dimensions.
2. I think that there should be a priority strategy. Which priority/priority range should we use for `python.d.` modules.

python.d/logind.chart.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+# Description: logind netdata python.d module
+# Author: Austin S. Hemmelgarn (Ferroin)
+# SPDX-License-Identifier: GPL-3.0+
+
+from bases.FrameworkServices.ExecutableService import ExecutableService
+
+PRIORITY = 40000
```
l2isbad: `PRIORITY` is wrong

https://github.com/firehol/netdata/blob/master/plugins.d/python.d.plugin#L142-L143
https://github.com/firehol/netdata/blob/master/plugins.d/python.d.plugin#L37-L42

Not sure about 40000 btw too far from 60000

Ferroin: Yeah, meant to ask about that.  The point here was to get it just below the `uses` chart, as that seems to me to be the most logical location for it.

Ferroin:
I'm personally against having it much higher than the default.  Running the module is not hugely expensive in terms of processing time, the only expensive part is the process setup for `loginctl` (which actually accounts for almost 80% of the time it takes to _run_ `loginctl` even in a terminal), and even that is not hugely bad (other than the context switching and cache invalidation, but that's going to happen at least once regardless for any plugin).  However, these are not collectable (at least not easily) as 'running' counts, only point-in-time values, and therefore increasing the granularity increases the chances of missing an event (as it is one second granularity is likely to miss some, there are perfectly legitimate reasons for sub-second duration sessions to occur).

That said, as mentioned, we at least in theory could just skip `loginctl` and query the DBus interface directly.  However, that has some significant implications:

1. It will make the code much more complex, and add a extra python dependency which is not guaranteed to be present or usable (if logind is installed, you'll always have `loginctl`, but you may not have python bindings for DBus).
2. The DBus interface has changed at least twice in incompatible ways, but the output of `loginctl` has remained functionally the same.  IOW, we're liable to have things randomly break, and need to support multiple interfaces.

Ferroin:
Rebased to resolve merge conflicts.

Ferroin:
@ktsaou If you could take a look at this, and ideally weigh-in on naming conventions and the whole priority thing, it would be appreciated.

python.d/logind.chart.py
l2isbad: anyway `priority` should be lowercase

l2isbad: 40k is ok for now

Ferroin: Of course I see this right _after_ I push a commit updating it to 59999 so it just floats to the top of the Python plugins...

I actually am interested to hear @ktsaou's input on what would be reasonable for this.  Personally, I think it actually makes the most sense up with the main system charts (that is, at least above the apps.plugin charts), but I'm not sure where up there, or if it even makes sense to put it all the way up there.

l2isbad:
@ktsaou the only thing I doubt is `update_every` 1 for this plugin. It uses `popen` and it runs almost on every distro.

ktsaou:
@Ferroin I agree with @l2isbad that this is too heavy to be run every single second on all systems.

I suggest to merge it disabled by default. Then we can experiment a bit and see if it should enabled with a lower frequency.

@l2isbad will the plugin be enabled if `python.conf` is modified by hand, so that the new file will not be installed and the line `logind: no` is missing?

Ferroin:
@ktsaou Understood, although I still contend that this is significantly less useful at lower collection frequencies.  Strictly speaking, what it's doing is tracking the current number of active entries of whatever type that became active during the past collection period and are still active, so longer collection periods reduce the likelihood of tracking very short-lived sessions (which are actually really common if you're using SSH with specific commands for remote administration).

For now, would you like me to just update the plugin to not run by default?

ktsaou:
@Ferroin yes I agree it would be much better if it runs frequently.

But, up to now we have avoided all plugins that fork frequently. Even the `tc` plugin that is enabled by default runs `tc` once every minute if QoS is not enabled on a system.

So, although I like this plugin, I think that we should disable it by default. The basic info this plugin provides (like which users are logged in), is already provided by apps.plugin at the "User's section" - so I think we will not miss something that important by having it disabled by default.

Think a bit of it. On my system `loginctl` has a dynamic linking like this:

```sh
# ldd $(which loginctl)
	linux-vdso.so.1 (0x00007ffc2b106000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f5fd1230000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f5fd0e74000)
	libsystemd-shared-238.so => /usr/lib/systemd/libsystemd-shared-238.so (0x00007f5fd0a24000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f5fd165a000)
	librt.so.1 => /usr/lib/librt.so.1 (0x00007f5fd081c000)
	libcap.so.2 => /usr/lib/libcap.so.2 (0x00007f5fd0618000)
	libacl.so.1 => /usr/lib/libacl.so.1 (0x00007f5fd040f000)
	libcryptsetup.so.12 => /usr/lib/libcryptsetup.so.12 (0x00007f5fd01bf000)
	libgcrypt.so.20 => /usr/lib/libgcrypt.so.20 (0x00007f5fcfea3000)
	libip4tc.so.0 => /usr/lib/libip4tc.so.0 (0x00007f5fcfc9c000)
	libseccomp.so.2 => /usr/lib/libseccomp.so.2 (0x00007f5fcfa5b000)
	libidn.so.11 => /usr/lib/libidn.so.11 (0x00007f5fcf827000)
	liblzma.so.5 => /usr/lib/liblzma.so.5 (0x00007f5fcf601000)
	liblz4.so.1 => /usr/lib/liblz4.so.1 (0x00007f5fcf3e4000)
	libblkid.so.1 => /usr/lib/libblkid.so.1 (0x00007f5fcf194000)
	libmount.so.1 => /usr/lib/libmount.so.1 (0x00007f5fcef3c000)
	libattr.so.1 => /usr/lib/libattr.so.1 (0x00007f5fced37000)
	libuuid.so.1 => /usr/lib/libuuid.so.1 (0x00007f5fceb30000)
	libdevmapper.so.1.02 => /usr/lib/libdevmapper.so.1.02 (0x00007f5fce8db000)
	libgpg-error.so.0 => /usr/lib/libgpg-error.so.0 (0x00007f5fce6bb000)
	libargon2.so.1 => /usr/lib/libargon2.so.1 (0x00007f5fce4b2000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f5fce2ae000)
	libjson-c.so.3 => /usr/lib/libjson-c.so.3 (0x00007f5fce09f000)
	libpcap.so.1 => /usr/lib/libpcap.so.1 (0x00007f5fcde5b000)
	libudev.so.1 => /usr/lib/libudev.so.1 (0x00007f5fcdc3d000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007f5fcd8a8000)
	libnl-genl-3.so.200 => /usr/lib/libnl-genl-3.so.200 (0x00007f5fcd6a2000)
	libnl-3.so.200 => /usr/lib/libnl-3.so.200 (0x00007f5fcd481000)
	libdbus-1.so.3 => /usr/lib/libdbus-1.so.3 (0x00007f5fcd230000)
	libsystemd.so.0 => /usr/lib/libsystemd.so.0 (0x00007f5fccfa9000)
```

Just executing this plugin will introduce a dramatic overhead. It may not be noticeable given the power of modern computers, but if you pay some attention to minor page faults you will understand this is really bad: thousands of quite complex operations are required, just to execute this little command.

I strongly believe we should do everything in our power to ensure netdata is "lightweight" or "a good citizen" for our servers.

So, as a rule of thumb I suggest we should never enable plugins that fork by default, when they will run unconditionally on all (or most) servers out there.

l2isbad:
> will the plugin be enabled if python.conf is modified by hand

Unfortunately yes. I'll see how it can be fixed.

l2isbad:
https://github.com/firehol/netdata/pull/3869

ktsaou:
> #3869

merged that.

Ferroin:
OK, I'll update this with a commit disabling the plugin by default, add a warning to the documentation that it runs frequently by default, and then probably take a look at the practicality of adding a `DBusService` framework class and adapting this to use that (because that avoids the need to invoke loginctl at all, although it _might_ not be quite as portable as this).

Ferroin:
Updated.  I'll look into the DBus stuff as a separate PR.

Ferroin:
OK, now rebased so that it has the changes from #3869 (there weren't any conflicts, but it couldn't be tested properly without those changes).

ktsaou:
tested it and it seems good.

btw, here is python.d.plugin without and with logind enabled (page faults):

image

ktsaou:
And this is its CPU usage:

image

As you can see, it is quite heavy. Almost 1% cpu increase by itself.

PR #3818, opened 2018-06-13
Replace LICENSE symlink with actual file.

Ferroin:
Should fix poorly coded GitHub license checks (which _should_ be following symlinks inside the repo).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,1 +1,674 @@
-LICENSES/GPL-3.0
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.  We, the Free Software Foundation, use the
+GNU General Public License for most of our software; it applies also to
+any other work released this way by its authors.  You can apply it to
+your programs, too.
+
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
+
+  To protect your rights, we need to prevent others from denying you
+these rights or asking you to surrender the rights.  Therefore, you have
+certain responsibilities if you distribute copies of the software, or if
+you modify it: responsibilities to respect the freedom of others.
+
+  For example, if you distribute copies of such a program, whether
+gratis or for a fee, you must pass on to the recipients the same
+freedoms that you received.  You must make sure that they, too, receive
+or can get the source code.  And you must show them these terms so they
+know their rights.
+
+  Developers that use the GNU GPL protect your rights with two steps:
+(1) assert copyright on the software, and (2) offer you this License
+giving you legal permission to copy, distribute and/or modify it.
+
+  For the developers' and authors' protection, the GPL clearly explains
+that there is no warranty for this free software.  For both users' and
+authors' sake, the GPL requires that modified versions be marked as
+changed, so that their problems will not be attributed erroneously to
+authors of previous versions.
+
+  Some devices are designed to deny users access to install or run
+modified versions of the software inside them, although the manufacturer
+can do so.  This is fundamentally incompatible with the aim of
+protecting users' freedom to change the software.  The systematic
+pattern of such abuse occurs in the area of products for individuals to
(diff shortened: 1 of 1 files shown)


ktsaou:
hm... I think LICENSE.md will have to be the same with LICENSES/GPL-3 for it to work.
Also, does the above change require a Makefile change?

Check also this for the detection process: https://help.github.com/articles/licensing-a-repository/#detecting-a-license

Ferroin:
Unless I'm mistaken, there shouldn't be a Makefile change needed.  From what I can tell from the Ruby code (I am by no means an expert on Ruby, but I know enough to get a general idea what it's doing), it checks the `LICENSE` file first if it exists, so this should be enough to get it to flag Netdata correctly.

ktsaou:
let's see

ktsaou:
hm... no change

Ferroin:
Yeah, I'll pull down a copy of the Ruby code they're using and experiment a bit with it to see if I can figure out what's up here.

Ferroin:
OK, this is weird.  Licensee correctly ID's the LICENSE file as the main license and a perfect match for the GPL-3.

ktsaou:
hm... it may be some caching on github. Let's give it a day... (although when I initialized it, I remember it was updated in real-time).

ktsaou:
still no luck.
Something else is happening. Any ideas?

Ferroin:
I'm not quite sure.  My only thought at this point is that they're ignoring `LICENSE` and just looking at `LICENSE.md`, which the ruby module correctly identifies as something it doesn't recognize.

ktsaou:
Check this:

```sh
# curl -H "Accept: application/vnd.github.black-panther-preview+json" "https://api.github.com/repos/firehol/netdata/community/profile"
{
  "health_percentage": 50,
  "description": "Get control of your servers. Simple. Effective. Awesome! https://my-netdata.io/",
  "documentation": null,
  "files": {
    "code_of_conduct": null,
    "contributing": null,
    "issue_template": null,
    "pull_request_template": null,
    "license": {
      "key": "other",
      "name": "Other",
      "spdx_id": null,
      "url": null,
      "node_id": "MDc6TGljZW5zZTA=",
      "html_url": "https://github.com/firehol/netdata/blob/master/LICENSE"
    },
    "readme": {
      "url": "https://api.github.com/repos/firehol/netdata/contents/README.md",
      "html_url": "https://github.com/firehol/netdata/blob/master/README.md"
    }
  },
  "updated_at": "2018-05-10T15:23:33Z"
}
```

It says `spdx_id: null`...

hm... shall I contact github support?

Ferroin:
I've never looked at the GitHub API's before, but based on the structure of the returned data, I'd expect the `spdx_id` key is set based on the detected licensing, just like the `name` should be.

That said, I would absolutely contact GitHub support, as they're pointing at a file that an up-to-date version of licensee is correctly identifying as GPL-3.0 on my system.

ktsaou:
Sent them an email. Let's see...

ktsaou:
This is what they replied:

>Thanks for getting in touch! I see that you currently have 3 files that begin with LICENSE in the root of the repository:
>
>https://github.com/firehol/netdata/blob/master/LICENSE (GPL v3 License)
>https://github.com/firehol/netdata/blob/master/LICENSE-REDISTRIBUTED.md
>https://github.com/firehol/netdata/blob/master/LICENSE.md
>
>Because your repository contains several licenses, we recommend keeping one of the files (the GPL v3 >License) and renaming the other two so that your license can be detected by our systems. Would such >an approach work for you?
>
>https://help.github.com/articles/licensing-a-repository/#detecting-a-license
>
>All the best,

The problem with `LICENSE-REDISTRIBUTED.md` is that it is linked a lot...

Ferroin:
Linked a lot where exactly?  If it's all internal to the repository, then a change can easily be scripted.

Realistically, it might make sense to just combine `LICENSE.md` and `LICENSE-REDISTRIBUTED.md` into one file with a different name.  Perhaps `REDISTRIBUTING.md`?

ktsaou:
> Linked a lot where exactly?

I mean wiki pages, and github issues.

> Realistically, it might make sense to just combine LICENSE.md and LICENSE-REDISTRIBUTED.md into one file with a different name. Perhaps REDISTRIBUTING.md?

Sounds good...

Ferroin:
OK, I'll look at getting a PR together to do so, and once that gets merged I'll try and get all the links in the Wiki pages updated (can't really do much about the GitHub issues though I suppose).

Ferroin:
Alternatively, would you be OK with just merging`LICENSE.md` and `LICENSE-REDISTRIBUTED.md` into the licensing info section at the end of `README.md`?  That would preserve the information somewhere where at least part of it already is.

ktsaou:
> Alternatively, would you be OK with just mergingLICENSE.md and LICENSE-REDISTRIBUTED.md into the licensing info section at the end of README.md? That would preserve the information somewhere where at least part of it already is.

hm... README.md will become a monster with not so useful info. I don't like this. README.md is a "marketing tool"... :smile: ...so I would avoid having there irrelevant info.

I suggest to merge `LICENSE` with `LICENSE.md` and rename `LICENSE-REDISTRIBUTED.md` to `REDISTRIBUTED.md`.

> Oh, also, I assume you would like the Copyright notices updated to reflect the new company?

Yes, it would be fantastic if you also do this!

I also think we should add CNCF logo and info at the home page, README.md, etc, but I can do this later.

PR #3822, opened 2018-06-14
Add 1-Wire sensor monitoring

dspinellis:
Auto-detect and display data from multiple 1-Wire temperature sensors.
Sensor names are user-configurable.

### DIFF
--- a/python.d/w1sensor.chart.py
+++ b/python.d/w1sensor.chart.py
@@ -1,0 +1,103 @@
+# -*- coding: utf-8 -*-
+# Description: 1-wire temperature monitor netdata python.d module
+# Author: Diomidis Spinellis <http://www.spinellis.gr>
+
+import os
+import re
+from bases.FrameworkServices.SimpleService import SimpleService
+
+# default module values (can be overridden per job in `config`)
+update_every = 5
+
+# Location where 1-Wire devices can be found
+W1_DIR = '/sys/bus/w1/devices/'
+
+# Lines matching the following regular expression contain a temperature value
+RE_TEMP = re.compile(r' t=(\d+)')
+
+ORDER = ['temp']
+
+# id: {
+#     'options': [name, title, units, family (heading 2), context, charttype],
+#     'lines': [
+#         [unique_dimension_name, name, algorithm, multiplier, divisor]
+#     ]}
+CHARTS = {
+    'temp': {
+        'options': [
+            None,
+            '1-Wire Temperature Sensor',
+            'Celsius',
+            'Temperature',
+            'w1sensor.temp',
+            'line'
+        ],
+        # lines are constructed from check
+    }
+}
+
+# Known and supported family members
+# Based on linux/drivers/w1/w1_family.h and w1/slaves/w1_therm.c
+THERM_FAMILY = {
+    '10': 'W1_THERM_DS18S20',
+    '22': 'W1_THERM_DS1822',
+    '28': 'W1_THERM_DS18B20',
+    '3b': 'W1_THERM_DS1825',
+    '42': 'W1_THERM_DS28EA00',
+}
+
+class Service(SimpleService):
+    """Provide netdata service for 1-Wire sensors"""
+    def __init__(self, configuration=None, name=None):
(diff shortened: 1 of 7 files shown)


dspinellis:
1-wire

Here is an image of the resultant dashboard.

python.d/w1sensor.chart.py
```diff
@@ -0,0 +1,103 @@
...
+                identifier = file_name[3:]
+                name = identifier
+                config_name = self.configuration.get('name_' + identifier)
+                if config_name:
+                    name = config_name
+                lines.append(['w1sensor_temp_' + identifier, name,
+                              'absolute', 1, 10])
+            self.definitions['temp']['lines'] = lines
+            return len(self.probes) > 0
+        except Exception as error:
```
l2isbad: Some suggestions:
1. Make `exception` more specific
2. Remove all code from `try/except` block which can not cause an exception

Like:

```python
try:
    filenames = os.listdir(W1_DIR)
except OSError as err:  # i believe we only need OSError here, but i am not sure
    self.error(err)
    return None

...
...
```

python.d/w1sensor.chart.py
```diff
@@ -0,0 +1,103 @@
...
+            try:
+                with open(file_path, 'r') as device_file:
+                    for line in device_file:
+                        matched = RE_TEMP.search(line)
+                        if matched:
+                            # Round to one decimal digit to filter-out noise
+                            value = round(int(matched.group(1)) / 1000., 1)
+                            value = int(value * 10)
+                            data['w1sensor_temp_' + identifier] = value
+            except Exception as error:
```
l2isbad: please make `exception` more specific

l2isbad:
@dspinellis thanks for the contribution :+1:

python.d/w1sensor.chart.py
dspinellis: Done.

python.d/w1sensor.chart.py
dspinellis: Done.

dspinellis:
Thanks!  I pushed a commit addressing your comments. I can then squash the commits when the PR passes code review.

python.d/w1sensor.chart.py
```diff
@@ -0,0 +1,105 @@
...
+            try:
+                with open(file_path, 'r') as device_file:
+                    for line in device_file:
+                        matched = RE_TEMP.search(line)
+                        if matched:
+                            # Round to one decimal digit to filter-out noise
+                            value = round(int(matched.group(1)) / 1000., 1)
+                            value = int(value * 10)
+                            data['w1sensor_temp_' + identifier] = value
+            except OSError as err:
```
l2isbad: add [IOError](https://docs.python.org/2/library/exceptions.html#exceptions.IOError) here

dspinellis:
I squashed my commits following your approval of the changes.

python.d/w1sensor.chart.py
```diff
@@ -0,0 +1,105 @@
...
+# Author: Diomidis Spinellis <http://www.spinellis.gr>
+
+import os
+import re
+from bases.FrameworkServices.SimpleService import SimpleService
+
+# default module values (can be overridden per job in `config`)
+update_every = 5
+priority = 60000
+retries = 60
```
Ferroin: Unless I'm mistaken, you're using the same number of retries as the default, in which case this line can probably be removed.

l2isbad: Yes, [retries and priority](https://github.com/firehol/netdata/blob/master/plugins.d/python.d.plugin#L38-L39) can be removed since it is current default

l2isbad:
@dspinellis please do a rebase

python.d/w1sensor.chart.py
dspinellis: Thank you; I removed them.

dspinellis:
I rebased and squashed the commit to remove the unneeded default values.

ktsaou:
Well done!
Thank you @dspinellis !

PR #3831, opened 2018-06-15
Add the BOINC alarms config to Makefile.am

Ferroin:
@ktsaou Sorry this got missed during the PR for the BOINC module.  I'm still not quite used to working with the autotools build system.  Not technically critical, but would be nice to have merged in the near future.


ktsaou:
thank you!

Issue #3846, opened 2018-06-19
Mongodb with ssl/keyfile does not work


0xicl33n:
Connecting to a server with keyfile encryption causes it to close immediately
```
netdata@spyke2:/etc/netdata/python.d$ /usr/libexec/netdata/plugins.d/python.d.plugin 1 debug trace mongodb
2018-06-19 19:15:57: python.d INFO: plugin: main: Using python 2
2018-06-19 19:15:58: python.d DEBUG: plugin: main: module load source: 'mongodb' => [OK]
2018-06-19 19:15:58: python.d DEBUG: plugin: main: module load config: 'mongodb' => [OK]
2018-06-19 19:15:58: python.d DEBUG: plugin: main: job initialization: 'mongodb scalegrid' => ['OK']
2018-06-19 19:15:58: python.d DEBUG: plugin: main: module status: 'mongodb' => [OK] (jobs: 1)
2018-06-19 19:15:58: python.d ERROR: mongodb: scalegrid: connection closed
2018-06-19 19:15:58: python.d INFO: mongodb: scalegrid: check() => [FAILED]
2018-06-19 19:15:59: python.d INFO: plugin: main: FINISHED
```

`pymongo` [supports SSL and keyfile encrytion](http://api.mongodb.com/python/current/examples/tls.html), but netdata does not. It can be completely transparent of the actual netdata application since it is just a plugin. I will look into adding support for it. The conf file could look like this:

**mongodb.conf:**
```
example:
    name : 'example'
    host : 127.0.0.1
    port : 45801
    user : 'user'
    password : 'password' 
    ssl : 'yes'
    keyfile : '/home/netdata/db.pem'

```

The other option is to  allow us to connect with [mongodb connect strings](https://docs.mongodb.com/manual/reference/connection-string/). [Motor ](http://motor.readthedocs.io/en/stable/)would actually be a much better alternative to pymongo as well since they both have the same syntax

Ferroin:
The relevant plugin code can be found [here](https://github.com/firehol/netdata/blob/master/python.d/mongodb.chart.py).

Ferroin:
Also, if you make any changes to the config format, don't forget to update the config file [here](https://github.com/firehol/netdata/blob/master/conf.d/python.d/mongodb.conf) and the readme info [here](https://github.com/firehol/netdata/blob/master/python.d/README.md) to reflect the changes.

0xicl33n:
```
[...]

PR #3847, opened 2018-06-19
Add support for mongodb connect URI's

0xicl33n:
#3846 

With uris, mongo connection should be this easy

**mongodb.conf:**
```
myDatabase:
    uri = 'mongodb://connection_information'
````

### DIFF
--- a/python.d/mongodb.chart.py
+++ b/python.d/mongodb.chart.py
@@ -650,14 +650,18 @@ class Service(SimpleService):
 
     def _create_connection(self):
         conn_vars = {'host': self.host, 'port': self.port}
+        if hasattr(MongoClient, 'server_selection_timeout'):
+            conn_vars.update({'serverselectiontimeoutms': self.timeout})
         if self.uri:
             try:
-                connection = MongoClient(self.uri)
-                server_status = connection.admin.command('serverStatus')
+                if self.timeout:
+                    connection = MongoClient(self.uri,serverselectiontimeoutms=self.timeout)
+                    server_status = connection.admin.command('serverStatus')
+                else:
+                    connection = MongoClient(self.uri)
+                    server_status = connection.admin.command('serverStatus')
             except PyMongoError as error:
                 return None, None, str(error)
-        elif hasattr(MongoClient, 'server_selection_timeout'):
-            conn_vars.update({'serverselectiontimeoutms': self.timeout})
         try:
             connection = MongoClient(**conn_vars)
             if self.user and self.password:
--- a/python.d/mongodb.chart.py
+++ b/python.d/mongodb.chart.py
@@ -376,6 +376,7 @@ class Service(SimpleService):
         self.password = self.configuration.get('pass')
         self.host = self.configuration.get('host', '127.0.0.1')
         self.port = self.configuration.get('port', 27017)
+        self.uri = self.configuration.get('uri')
         self.timeout = self.configuration.get('timeout', 100)
         self.metrics_to_collect = deepcopy(DEFAULT_METRICS)
         self.connection = None
@@ -649,7 +650,13 @@ class Service(SimpleService):
 
     def _create_connection(self):
         conn_vars = {'host': self.host, 'port': self.port}
-        if hasattr(MongoClient, 'server_selection_timeout'):
+        if self.uri:
+            try:
+                connection = MongoClient(self.uri)
+                server_status = connection.admin.command('serverStatus')
+            except PyMongoError as error:
+                return None, None, str(error)
+        elif hasattr(MongoClient, 'server_selection_timeout'):
             conn_vars.update({'serverselectiontimeoutms': self.timeout})
         try:
             connection = MongoClient(**conn_vars)
--- a/conf.d/python.d/mongodb.conf
+++ b/conf.d/python.d/mongodb.conf
@@ -85,9 +85,9 @@
 #
 #     uri = 'mongodb://USERNAME:PASSWORD@127.0.01:27017/DATABASE_NAME?replicaSet=REPLICA_SET_NAME&ssl=true&ssl_ca_certs=/location/to/db.pem'
 # 
-# The permissions of the keyfile (db.pem or whatever your keyfile is) will need to allow the netdata user owner to read it
+# The permissions of the keyfile (db.pem or whatever your keyfile is) will need to allow the netdata user to read it:
 #
-#     chown netdata:netdata /location/to/db.pem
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: python.d/mongodb.chart.py [hunk 1]
```diff
@@ -649,7 +650,13 @@ def _get_data(self):
...
         conn_vars = {'host': self.host, 'port': self.port}
-        if hasattr(MongoClient, 'server_selection_timeout'):
+        if self.uri:
+            try:
+                connection = MongoClient(self.uri)
+                server_status = connection.admin.command('serverStatus')
+            except PyMongoError as error:
+                return None, None, str(error)
+        elif hasattr(MongoClient, 'server_selection_timeout'):
             conn_vars.update({'serverselectiontimeoutms': self.timeout})
```
l2isbad: > serverSelectionTimeoutMS: (integer) Controls how long (in milliseconds) the driver will wait to find an available, appropriate server to carry out a database operation; while it is waiting, multiple server monitoring operations may be carried out, each controlled by connectTimeoutMS. Defaults to 30000 (30 seconds)

i think `serverselectiontimeoutms` should be setted for `uri` connections too.

0xicl33n: Can do, ill look at it tomorrow.

l2isbad: @0xicl33n i see in `MongoClient` documentation

>         :Parameters:
>           - `host` (optional): hostname or IP address or Unix domain socket
>             path of a single mongod or mongos instance to connect to, or a
>             mongodb URI, or a list of hostnames / mongodb URIs. If `host` is
>             an IPv6 literal it must be enclosed in '[' and ']' characters
>             following the RFC2732 URL syntax (e.g. '[::1]' for localhost).
>             Multihomed and round robin DNS addresses are **not** supported.

So what if replace

https://github.com/firehol/netdata/blob/f47b067b49ee22cc01ac351c7a7c83c136806986/python.d/mongodb.chart.py#L377

with

```python
self.host = self.configuration.get('uri') or self.configuration.get('host', '127.0.0.1')
```

It seems that this will be enough

Ferroin:
Oops, sorry about the double review (it would be nice if Github were a bit smarter about that...).

python.d/mongodb.chart.py
l2isbad: Actually you can use `host` parameter instead of `uri`. @Ferroin what do you think?

l2isbad: i mean `mongodb connect URI's` already supported by current version. All you need is to specify uri as a `host` in the configuration file.

Ferroin: I'm not quite certain that that's the case.  It depends on how MongoClient handles being passed a port in addition to a URI.

At https://github.com/firehol/netdata/blob/master/python.d/mongodb.chart.py#L651, we're unconditionally including the port in the parameters we're putting together to initiate the connection, and that is gauaranteed to be a set value (either the default MongoDB port, or whatever the user supplied).  Given that, we can only keep things as-is iff MongoClient ignores the port parameter when `host` is a URI that includes a port, otherwise we end up not properly supporting the URI syntax.

l2isbad: > we're unconditionally including the port in the parameters

yeah, this should be tested ofc

l2isbad: Ok, i looked at the pymongo code again (`MongoClient` class `python-pymongo 3.6.1-1`)

```python
        for entity in host:
            if "://" in entity:
                res = uri_parser.parse_uri(entity, port, warn=True)
                seeds.update(res["nodelist"])
                username = res["username"] or username
                password = res["password"] or password
                dbase = res["database"] or dbase
                opts = res["options"]
            else:
                seeds.update(uri_parser.split_hosts(entity, port))
```

seems we are safe to include `port`

Ferroin: OK, assuming that it passes testing, that appears to be fine then.  The documentation still needs updated (and I think the changes made by this PR actually do a great job of that), but there should be no code changes needed (though maybe adding a warning if `host` looks like a URI and the config also specifies a value for `port`?).

l2isbad: > The documentation still needs updated

yes

0xicl33n: So then it should be looking more like..

```python
        conn_vars = {'host': self.host, 'port': self.port}
        if self.uri:
            try:
                connection = MongoClient(self.uri,self.timeout)
```
?

0xicl33n:
i dont like how 4df7d27753a38083e11df5e6600978ada6aaf984 looks, its not very clean but it should work

The other idea i had would be like this

```python
    def _create_connection(self):
        conn_vars = {'host': self.host, 'port': self.port}
        if hasattr(MongoClient, 'server_selection_timeout'):
            conn_vars.update({'serverselectiontimeoutms': self.timeout})

        uri_vars = {'uri':self.uri}
        if hasattr(MongoClient, 'server_selection_timeout'):
            uri_vars.update({'serverselectiontimeoutms': self.timeout})

        if self.uri:
            try:
                connection = MongoClient(**uri_vars)
                server_status = connection.admin.command('serverStatus')
            except PyMongoError as error:
                return None, None, str(error)
```
Which mimics existing format, but the problem there is there is no `uri` argument that normally gets passed to `pymongo.client`. So, im not sure how to deal with it

l2isbad:
>  but the problem there is there is no uri argument that normally gets passed to pymongo.client

>     :Parameters:
>       - `host` (optional): hostname or IP address or Unix domain socket
>         path of a single mongod or mongos instance to connect to, or a
>         mongodb URI, or a list of hostnames / mongodb URIs. If `host` is
>         an IPv6 literal it must be enclosed in '[' and ']' characters
>         following the RFC2732 URL syntax (e.g. '[::1]' for localhost).
>         Multihomed and round robin DNS addresses are **not** supported.

^^ `uri` can be passed as a `host` according docs

code from `MongoClient` 
```python
        for entity in host:
            if "://" in entity:
                res = uri_parser.parse_uri(entity, port, warn=True)
                seeds.update(res["nodelist"])
                username = res["username"] or username
                password = res["password"] or password
                dbase = res["database"] or dbase
                opts = res["options"]
            else:
                seeds.update(uri_parser.split_hosts(entity, port))
```

So it should work without any changes, please test it

0xicl33n:
This absolutely works
image
****

Ferroin:
@0xicl33n Updating the documentation to reflect that might be a good idea.  I actually think the documentation changes you had done for this PR were good, so if you want to update just those and submit a separate PR with them, that would be great.

PR #3871, opened 2018-06-22
Update license file names so Github correctly detects our license.

Ferroin:
As discussed in #3818.

@ktsaou I'm going to try and find a way to quickly search all of the wiki for what will now be dead links after merging this, though I'm not too optimistic yet that it will be easily possible (Ideally, I'd just clone the git repo backing the wiki to my local system and check from there, but I'm not sure how to do that).


Ferroin:
Never mind, figured out how to clone the wiki.  Are you sure that anything there is actually referencing `LICENSE-REDISTRIBUTED.md`?  Based on a quick search through the pages with `grep`, I'm not finding _any_ links to it.

ktsaou:
No, I am not sure. Since you checked, I think we are ok.

ktsaou:
license came up again!

PR #3877, opened 2018-06-23
Sign CLA

simonnagl:


ktsaou:
Thank you!
I think a pipe is missing at the end and it will break the table. Could you please add it?

simonnagl:
Done. Please tell me if I should sqash those two commits. Sorry for being careless.

ktsaou:
it is ok. thank you!

PR #3893, opened 2018-06-26
Update freebsd_getifaddrs.c for system.N

RicardoSette:
Suggestion of code to validate if a interface is physical in FreeBSD, unfortunately is not easy as Linux, for FreeBSD there is no flag to indicate that the interface is physical.

We had some "false positives" with optical interfaces, often because of GBICs not homologated for FreeBSD, so we started to control physical interfaces in NETDATA using CONFIG_SECTION_GETIFADDRS, much simpler and much more accurate.

We added in the section the "set physical interfaces for system.net" attribute where DEFAULT_PHYSICAL_INTERFACES is "igb* ix* cxl* in* ixl* ixlv* bge* ixgbe*".

Now the identification of physical interfaces is controlled by the settings, avoiding other "false positive".

=> Added total bandwidth for physical interfaces
=> Added total packets for physical interfaces

### DIFF
--- a/src/freebsd_getifaddrs.c
+++ b/src/freebsd_getifaddrs.c
@@ -142,19 +142,24 @@ static struct cgroup_network_interface *get_network_interface(const char *name) 
 int do_getifaddrs(int update_every, usec_t dt) {
     (void)dt;
 
-#define DELAULT_EXLUDED_INTERFACES "lo*"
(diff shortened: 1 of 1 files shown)


ktsaou:
Commented on the wrong PR.
This seems good for me. So, please sign the CLA and I'll merge it.

PR #3899, opened 2018-06-27
Sign CLA

davidak:
As said in https://github.com/firehol/netdata/pull/3695, it is not my intention to cause trouble for the project, so feel free to relicense it under any open source license you think helps the project to get better.

I still believe GPL is great to protect the freedom of the users and i hope you find a way to keep it.


ktsaou:
Thank you! I agree and I'll try to stay on GPL.

Could you please resolve the conflict? I am on a trip and my internet connection is too bad for this...

Issue #3330, opened 2018-01-23
Notifications to Microsoft Teams


wccropper:
Our company is adding Microsoft Teams. I would like to send alerts to a channel just like Slack, etc. I have the ability to create API urls just like Slack. What would be needed to implement this?

donghwicha:
you can use Prometheus after integration with netdata.  Prometheus has notification feature.

Ferroin:
The code that's actually responsible for sending the alerts is in [alarm-notify.sh](https://github.com/firehol/netdata/blob/master/plugins.d/alarm-notify.sh).

If you can get API urls similar to those used by Slack, it shouldn't be too hard to integrate.  The various `send_` functions in alarm-notify.sh are what actually issues alert (most of them are just wrappers around a call to cURL that submits the required data to the required URL), I'd suggest starting by looking through those in particular.

ktsaou:
@wccropper it should be very simple to do it. As @Ferroin says, it is just copy-paste-adapt.
You can give it a try and make a PR - I'll review this to make sure you did it right.

domacikolaci:
Hi, 

I'm looking at integrating Netdata to Microsoft Teams now, how do we proceed with this? Dont see anything in the Wiki.

Thanks :)

domacikolaci:
Bump <3

ktsaou:
It has been merged in early august.
@Ferroin has a pending task to create the wiki.

However, you can find relevant info at the config file: https://github.com/firehol/netdata/blob/be9507fe8d3bbf6fe9ec91bce653e00675352db6/conf.d/health_alarm_notify.conf#L342-L374

ktsaou:
This has been fixed with #3903

Issue #3489, opened 2018-02-28
Telegram Alarms works only by hand execution.


isca0:
I have deployed netdata on several servers, but i can see the alarms are only sended to telegram when i force to execute it by hand calling the script manually like:

```
/opt/netdata/usr/libexec/netdata/plugins.d/alarm-notify.sh sysadmin hostname 1519681630 1519679963 7 1519845175 host_reachable fping.hostname_latency hostname CRITICAL CLEAR 0 1 14@/opt/netdata/etc/netdata/health.d/fping.conf 328 0 up/down states if the remote host is reachable down up
```
The alarm-notify.sh test 'CRITICAL' works fine on every server.
All alarms are showed fine trought web browser also. 
I dont know if is something i missing here, but i see no erros on logs. 
health_alarm_notify.conf is configured as defaults, just added my telegram bot token and mychat id. 

>What i expected: All alarms automatically sent trought telegram.

My netdata version is netdata 1.9.0-505-gca904596_rolling, installed from kickstart-static64.sh.

ktsaou:
hm... interesting.

netdata sends the alarms while running as user `netdata`. Also, the static build is using a binary `curl` shipped with netdata.

Please try this:

```sh
# become user netdata
sudo su -s /bin/sh netdata

# use the static binaries shipped with netdata
export PATH="/opt/netdata/bin:${PATH}"

# enable debugging
export NETDATA_ALARM_NOTIFY_DEBUG=1

# test the alarms
/opt/netdata/usr/libexec/netdata/plugins.d/alarm-notify.sh test
```

Could you please post the output?

isca0:
God bless you for this DEBUG=1 command... ;) Now i can see the netdata curl is failing...

```
/opt/netdata/bin/curl --data-urlencode disable_notification=true --data-urlencode parse_mode=HTML --data-urlencode disable_web_page_preview=true --data-urlencode $'text=\342\234\205 hostname recovered - <b>test alarm</b>\ntest.chart (test.family)\n<a href="https://registry.my-netdata.io/goto-host-from-alarm.html?host=hostname&chart=test.chart&family=test.family&alarm=test_alarm&alarm_unique_id=1&alarm_id=1&alarm_event_id=3">test alarm (alarm was raised for 3 seconds)</a>\n<i>this is a test alarm to verify notifications work</i>' https://api.telegram.org/bot12456:XXAAHHSHHXBBA/sendMessage\?chat_id=chatidnum
[...]

PR #3903, opened 2018-06-28
adding Microsoft Team support for alert notification

tioumen:
These modifications are related to Microsoft Team support and allow sending alert notifications to MS Team.


plugins.d/alarm-notify.sh
```diff
@@ -218,7 +219,7 @@ fi
 images_base_url="https://registry.my-netdata.io"
 
 # curl options to use
-curl_options=
+curl_options="--insecure"
```
Ferroin: This PR should not change the default behavior of all cURL based notification mechanisms, especially if it's reintroducing a significant security issue (if cURL isn't checking certificates, it becomes trivial to MitM the connections and then ensure that notifications sent using cURL never arrive).

If MS Team integration _really_ requires cURL to not check certificates (and I seriously doubt that it can't be made to work with proper certificate checking), then you should be explicitly passing this option when you call cURL from `send_team()` _and_ making sure it's well documented that you are doing this.

plugins.d/alarm-notify.sh
```diff
@@ -1339,6 +1361,69 @@ send_telegram() {
...
+echo "entering in send_team function"
+    local webhook="${1}" channels="${2}" httpcode sent=0 channel color payload
+
+    [ "${SEND_TEAM}" != "YES" ] && return 1
+
+    case "${status}" in
+        WARNING)  icon="⚠️" && color="FFA500";;
+        CRITICAL) icon="🔥" && color="D93F3C";;
+        CLEAR)    icon="💚" && color="65A677";;
+        *)        icon="♡" && color="0076D7";;
```
Ferroin: It seems like it would be a nice feature to be able to define these icons and colors in the config.  Not really critical, just nice to have.

plugins.d/alarm-notify.sh
```diff
@@ -1339,6 +1361,69 @@ send_telegram() {
...
+                "name": "Netdata",
+                "targets": [
+                { "os": "default", "uri": "http://$(hostname -f):19999" }
+                ]
+            },
+            {
+                "@type": "OpenUri",
+                "name": "Splunk",
+                "targets": [
+                { "os": "default", "uri": "https://your_url.com" }
```
Ferroin: Is this actually supposed to be here?  If so, are you certain it shouldn't be templated and configured by the user from `health_alarm_notify.conf`?

Ferroin:
I can't directly test this myself, but from a quick look at the official documentation from MS, it looks like it's doing things correctly.

Also, this is going to need an update to `CONTRIBUTORS.md`, though I'm not quite sure how @ktsaou wants to handle that since the commit has different authorship from the PR itself.

ktsaou:
> Also, this is going to need an update to CONTRIBUTORS.md, though I'm not quite sure how @ktsaou wants to handle that since the commit has different authorship from the PR itself.

Yes, please make the commit from the same username and add yourself in CONTRIBUTORS.md to sign the CLA.

plugins.d/alarm-notify.sh
tioumen: my bad, I was getting issue related to this #3489 on Centos Linux. I gonna remove this setup.

plugins.d/alarm-notify.sh
tioumen: I would prepare a template for further use. My goal here is to propose template. In the comment you have all the references if you want to adapt it. I hope it's OK for you guys.

conf.d/health_alarm_notify.conf
```diff
@@ -332,6 +333,24 @@ SLACK_WEBHOOK_URL=""
...
 
+#------------------------------------------------------------------------------
+# MS Team (teams.com) global notification options
+# More details are available here regarding the payload syntax options : https://docs.microsoft.com/en-us/outlook/actionable-messages/message-card-reference
+# Online designer : https://acdesignerbeta.azurewebsites.net/
+# multiple recipients can be given like this:
+#                  "CHANNEL1 CHANNEL2 ..."
+
+# enable/disable sending team notifications
+SEND_TEAM="YES"
```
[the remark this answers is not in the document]
tioumen: You're right, I changed them as requested. Hope it's better now.

plugins.d/alarm-notify.sh
Ferroin: I think the part for Netdata is fine, I'm talking specifically about the 'Splunk' thing, with the whole 'https://your_url.com' address.  It looks like it's a left over from testing or example code.

tioumen: You're right, we can remove that part. Done in the next commit.
Actually not only test on my side, I'm pushing metrics to Splunk.

Ferroin:
@tioumen Looks like you accidentally updated the wrong file in that last commit.

tioumen:
@Ferroin I removed the --insecure as curl specific option. Then also 2 lines in the beginning of the file : 
image

Ferroin:
939a54d is removing the option from `conf.d/health_alarm_notify.conf` (and that one shouldn't be removed, it's there so people who need it can just uncomment that line).  The file where the issue I commented on is located is `plugins.d/alarm_notify.sh`.  The commit message from the aforementioned commit seems to indicate you meant to change the correct file (`plugins.d/alarm_notify.sh`), but edited the wrong one by mistake.

plugins.d/alarm-notify.sh
```diff
@@ -220,7 +220,7 @@ fi
 images_base_url="https://registry.my-netdata.io"
 
 # curl options to use
-curl_options="--insecure"
+#curl_options=""
```
Ferroin: Almost, but not quite.  This still needs to not be commented here (yes, it's a bit confusing, but shell scripting wasn't exactly designed for stuff this complex).

plugins.d/alarm-notify.sh
```diff
@@ -1339,6 +1363,62 @@ send_telegram() {
...
+            "@type": "MessageCard",
+            "themeColor": "${color}",
+            "title": "$icon Alert ${status} from netdata for ${host}",
+            "text": "${host} ${status_message}, ${chart} (_${family}_), *${alarm}*",
+            "potentialAction": [
+            {
+                "@type": "OpenUri",
+                "name": "Netdata",
+                "targets": [
+                { "os": "default", "uri": "http://$(hostname -f):19999" }
```
ktsaou: please do not assume the URL is like that. A lot of people proxy netdata via other web server or use VPNs that prevent dns resolution. Also, this will resolve to the master netdata when it processes alarms for a slave.

You can use `${goto_url}` that uses the registry and the netdata unique id, to divert to the proper server.

ktsaou:
nice work.

There are 2 issues to be resolved.

plugins.d/alarm-notify.sh
```diff
@@ -1339,6 +1363,62 @@ send_telegram() {
...
+    [ "${SEND_MSTEAM}" != "YES" ] && return 1
+
+    case "${status}" in
+        WARNING)  icon="${MSTEAM_ICON_WARNING}" && color="${MSTEAM_COLOR_WARNING}";;
+        CRITICAL) icon="${MSTEAM_ICON_CRITICAL}" && color="${MSTEAM_COLOR_CRITICAL}";;
+        CLEAR)    icon="${MSTEAM_ICON_CLEAR}" && color="${MSTEAM_COLOR_CLEAR}";;
+        *)        icon="${MSTEAM_ICON_DEFAULT}" && color="${MSTEAM_COLOR_DEFAULT}";;
+    esac
+
+    for channels in ${channels}
```
ktsaou: this is a typo... it should be:

```
for channel in ${channels}
```

ktsaou: hm... I also don't see the variable `channel` used anywhere...
This means channels are not actually used.
Are you sure this works as expected?

tioumen: I made some changes according to your comment. I'm very disapointed about MS Teams regarding the way to push webhook on it. MS copied Slack but not all the best features ... Let me know if you are OK with my way to implement it. Thanks.

ktsaou:
ok, nice work! thank you for resolving the issues.
So, is this tested? Can we merge it?

tioumen:
I've tested each changes on the 4 tests servers I have. MS Teams notification is working fine even with several different channels. I'll be glad to fix any issues if any.

ktsaou:
merged! Thank you!
Could you please create a wiki page, like the rest of the alarm notification methods?
Generally, you need provide some basic information on how to setup this and a few screenshots of the final result.

ktsaou:
Well, it seems the one you tested was not the one in this PR.
It was giving a bash syntax error:

```sh
webhook=$(echo "${webhook//CHANNEL/${channel}})"
```

Check the final quote. `)` should be outside the quote.
Anyway I fixed it with PR #3958 like this:

```sh
webhook="${webhook//CHANNEL/${channel}}"
```

And prevented also a useless fork ( `$()` forks bash to execute the code ).

tioumen:
@ktsaou I've planned to create Wiki page during my vacations. Will be done before end of August. Thanks for your understanding.

ktsaou:
This fixes #3330

PR #3904, opened 2018-06-28
Sign CLA

Chocobo1:
Just for the record, I was hesitated at first, then I thought if netdata was under other license at the time I contributed, would I still have done it? Probably yes, *as long as netdata is open-source*.


ktsaou:
Thank you!

Issue #3927, opened 2018-07-03
Apcupsd: Connection loss further collects data, but it should stop


ccremer:
If I manually remove the USB cable to a directly attached APC UPS, the `apcupsd` plugin continues to send metrics, but each zeroed. This effectively results in false alarms: If I wanted to check the charge percentage, it will report 0 % on connection loss, even though it might still be working fine at 100%. The plugin should stop sending data at all in this case, so that 
```
template: apcupsd_last_collected_secs
      on: apcupsd.load
    calc: $now - $last_collected_t
   every: 10s
   units: seconds ago
    warn: $this > (($status >= $WARNING)  ? ($update_every) : ( 2 * $update_every))
    crit: $this > (($status == $CRITICAL) ? ($update_every) : (10 * $update_every))
   delay: down 5m multiplier 1.5 max 1h
    info: number of seconds since the last successful data collection
      to: sitemgr
```
actually gets triggered, and not the "charge critical" alarm.

Ferroin:
Interesting.  I wonder if `apcupsd` is officially supposed to report 0% charge under these circumstances.  If so, the alarm may need to be rewritten to assume that repeated 0% charge readings are a problem with data collection (arguably, they're a problem regardless and should just have an alarm anyway, as the other possible causes are a bad battery or a broken UPS).

cakrit:
This was fixed by #3928

ccremer:
No, unfortunately it's not. PR #3928 only added some alarms.
this is the output from apcupsd with UPS connected:
```
root@pve-2:/etc/netdata# apcaccess status localhost
APC      : 001,045,1038
DATE     : 2018-12-17 23:10:21 +0100
HOSTNAME : pve-2
VERSION  : 3.14.14 (31 May 2016) debian
UPSNAME  : pve-2
CABLE    : USB Cable
DRIVER   : USB UPS Driver
UPSMODE  : Stand Alone
STARTTIME: 2018-12-17 23:05:48 +0100
MODEL    : Back-UPS CS 500
STATUS   : ONLINE
LINEV    : 236.0 Volts
LOADPCT  : 34.0 Percent
BCHARGE  : 100.0 Percent
TIMELEFT : 28.5 Minutes
MBATTCHG : 10 Percent
MINTIMEL : 3 Minutes
MAXTIME  : 120 Seconds
OUTPUTV  : 230.0 Volts
SENSE    : Medium
DWAKE    : 0 Seconds
[...]

PR #3928, opened 2018-07-03
Add basic apcupsd alarm templates

ccremer:
I finally took the time and configure the APC UPS correctly. Had one over a year, but it was just in "dumb battery until it runs out" mode.

Anyway, added some alarms. Only the missing collection and charge warnings were actually tested though...

* Warning / Critical when Load is over 80 / 85 % (average over last 10 min)
* Warning / Critical when Charge is under 95 / 15 % (average over last 60 secs, enables fast alarms)
* Warning / Critical when Last collection time is higher than 2 times collection interval (enables fast alarms, with possible false positives)

The reason why I put the charge warning at 95 is, that because normally the charge is 100%. As soon as it goes to battery mode, this will give ppl time to react, if they haven't configured automatic shutdown or other rescue scripts. Those values are all, of course, configurable.

The "last collected" alarm should get triggered, if there wasn't a bug with the apcupsd plugin (#3927)

### DIFF
--- a/conf.d/Makefile.am
+++ b/conf.d/Makefile.am
@@ -88,6 +88,7 @@ healthconfigdir=$(configdir)/health.d
 
 dist_healthconfig_DATA = \
     health.d/apache.conf \
+    health.d/apcupsd.conf \
     health.d/backend.conf \
     health.d/bcache.conf \
     health.d/beanstalkd.conf \
--- a/conf.d/health.d/apcupsd.conf
+++ b/conf.d/health.d/apcupsd.conf
@@ -13,6 +13,8 @@ template: 10min_ups_load
     info: average UPS load for the last 10 minutes
       to: sitemgr
 
+# Discussion in https://github.com/firehol/netdata/pull/3928:
+# Fire the alarm as soon as it's going on battery (99% charge) and clear only when full.
 template: ups_charge
       on: apcupsd.charge
       os: *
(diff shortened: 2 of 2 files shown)


ccremer:
I just fear that there could be some internal rounding errors or measurement inaccuracy that leads to flapping between 99 and 100%. On the other hand, it is running constantly at 100% for a few hours now, never flapped (at 10s intervals). Maybe settle on 98% ? Bear in mind, that in apcupsd you can also specify scripts that should happen when going battery mode.

1 second update interval is actually too fast on my system. the daemon takes around 3 seconds to answer, that's why i choose 2 times @10s updates, that would make 20 secs already). Though I should probably assume that ppl use the defaults...

with the current bug, and 10 secs interval, a single fail to collect values already triggers the charge alarms, since the average is already at 80%

Ferroin:
Assuming that people use the defaults is good.  The current version looks like it's equivalent to what other similar checks for other plugins are doing, so I'd say that's probably the best for now unless there's some really compelling reason to have it behave differently.  Putting some documentation above the alarm that it's liable to trigger false positives would be good though.

As far as rounding errors in the charge reporting, I think that's functionally a non-issue.  We use about a dozen different models of APC UPS where I work, and I've checked charge reporting on each model  (same interface that netdata is using, just a different tool) over an 8 hour interval including cycling back and forth from battery to line power, and have seen no issues with fluttering of the values except when cycling rapidly on and off of battery.  Unless they logic for this was severely overengineered, they're just dividing the remaining charge by the full charge, expressed in at least milliamp-hours, and possibly in microamp-hours and then truncating the result to an integer, which provides very stable results that behave pretty much exactly how people expect it to.

ccremer:
Alright, I take your word for it ;)
If you have multiple models, do you always have the same metrics? Do all models report the temperature for example? I might add more alarms if they are more or less the same metrics over all models.

Ferroin:
I didn't think to check what other than the charge got reported, and I'm out until next Monday.  For the time being though, I think just merging this as is is fine, and then possibly adding some more in a separate PR would be fine.

ktsaou:
merged!

PR #3956, opened 2018-07-11
[WIP] Add support for ThingM Blink(1) notification LED's.

Ferroin:
ThingM makes a rather neat USB connected RGB LED designed for to act as a physical indicator of things like system state.

This adds support to the alerting infrastructure for changing the color of ThingM Blink(1) USB RGB LED devices based on notification state. This allows alerts to trigger physical changes in the appearance of
the system, thus allowing an admin to quickly check if a system has any alerts by simply looking at it.

It uses the following methodology to determine the color for the LED:

1. If any alert is CRITICAL, change the LED to the configured BLINK1_COLOR_CRIT color (red by default).
2. Otherwise, if any alert is a WARNING, change the LED to the configured BLINK1_COLOR_WARN color (yellow by default).
3. Otherwise, change the LED to the configured BLINK1_COLOR_CLEAR color (black by default, which translates to turning off the LED).

This requires both appropriate hardware (which is not expensive, and is also an open-source design), as well as a working copy of the official command-line interface called `blink1-tool`.

The changes make use of a temporary directory for tracking the state of the notifications (this is done efficiently by simply adding empty files to the directory for each notification in a given state other than CLEAR), and for serializing access to the Blink(1) devices (because `blink1-tool` is not concurrent-access safe).  This temporary directory _MUST_ be emptied out when Netdata restarts, otherwise notifications will get 'stuck'.

Recipient strings map to specific devices on the system, and may identify an LED by either it's device number (using a `devnum-NN` format), which is based on enumeration order, or it's serial number (using a `serial-HHHHHHHH` format), which is guaranteed unique for a given device.  Either value can be obtained by running the `blink1-tool --list` command.

Due to the temporary directory requirements, and the fact that that this requires `blink1-tool` to work correctly when run by the user Netdata is run as, this notification mechanism is disabled by default.

-------

**This has not been properly tested yet.**  It's not ready to merge as-is, and may not even work correctly, but I'm posting it here to get some feedback on whether or not there's any practical interest in this given the temporary directory requirement and the rather inflexible way it works combined with the fact that this can be done with custom notifications without needing much work from the end-user.  If there is interest, I'll work on getting it tested and cleaned up so it can be merged, if not, I'll just drop it, but leave the branch up on my fork of Netdata in case someone wants to pick it up and get it working.


ktsaou:
hm... I could export a few shell variables from the C code, before running `alarm-notify.sh`:

```sh
NETDATA_ALARMS_CLEAR=N
NETDATA_ALARMS_WARNING=N
NETDATA_ALARMS_CRITICAL=N
```

So that there will be no need to track the state of the alarms.

I could also run `alarm-notify.sh` for every state (now netdata does not run them for the `UNDEFINED` and `REMOVED` states). I tried to avoid that, because netdata would need to execute `alarm-notify.sh` possibly hundreds of times. So, ideally I would like to avoid this.

---

btw, we have to do something with `alarm-notify.sh`. I becomes huge!

One way could be to split each notification method to its own script file and source it from `alarm-notify.sh`. We could also work a bit the code that finds recipients per notification method to be more generic, so that we will not need to replicate the code with minor changes.

ktsaou:
> NETDATA_ALARMS_CLEAR=N
> NETDATA_ALARMS_WARNING=N
> NETDATA_ALARMS_CRITICAL=N

and we could use this info in emails too (eg. _"this server has 3 alarms in critical state and 2 in warning"_).

Ferroin:
I like the idea of having the variables in alarm-notify.sh to work from., but I'm not sure it's going to be as easy as you might think given that the tracking here is per recipient (so if a given recipient is only getting notifications for a specific role, only notifications going to that role will be tracked for that recipient).  Unfortunately, regardless of whether this needs to be stateful or not, it will still need somewhere to handle locking, as the `blink1-tool` command doesn't handle that itself and the hardware isn't safe against concurrent access.

I also like the idea of splitting out the various alarm mechanisms to their own files.  Something like the following comes to mind as a possible approach to this:

1.  Certain core code like finding recipients gets split out to a function that can then be called as needed by the code used for each method.
2. Each delivery method gets split to it's own file.  These only get sourced if the associated `SEND_*` variables are set to 'YES'.
3. Each delivery method defines it's own `check` function, which is used to determine whether or not all the requirements are met to be able to send notifications via this method.
4. Each delivery method defines it's own `send` function, used to send the notification (similar to what's done already).
5. We standardize how the `send` functions get called.  Currently, this is a mess.  Ideally, the formatting (which is sometimes done outside the function, sometimes in it) should be done in the function itself if things are getting split out like this.

If you can decide on how you want the directory structure to look for this (I'd vote for having a separate `alarm-notify.d` directory for the individual mechanisms), I can look at getting the split taken care of (and generalizing the recipient find code, and some of the other stuff), though it's likely to take me a while (because of the sheer amount of code involved more than anything else).

Ferroin:
Assuming #3960 gets merged, I'll update this to utilize that.

cakrit:
Any chance of moving this forward @Ferroin? It's such a cool feature, I wouldn't mind playing with my son, who asked me to help me learn how to code :)

Ferroin:
@cakrit I do still have interest in this, but I unfortunately don't really have the time as of right now to work on it (I've been really busy on a personal project on top of things being a bit chaotic at work).  If someone else wants to pick this up, I'd be more than happy to help with testing (I've got half-a-dozen of the USB LED modules this is supposed to work with (specifically, the mk2 revision), so I can test complex configurations), and would be happy to provide some advice on working with the CLI tool this is using, but I'm not sure that I'm the best person to push this forwards right now.

Unless you have some objection, I think the best way forwards right now would be to close this and start fresh on a new implementation, as much has changed in alarm-notify.sh since I last touched this, and what's here didn't ever consistently work to begin with.

If someone does do a new implementation, it might be worth looking instead at integrating with the Linux Kernel LED API, which includes support for the particular hardware this was designed to work with, as well as a lot of other hardware options, and thus might be more practically useful to a wider audience.

cakrit:
I created a new issue and left it as 'help wanted' for now. Closing this.

PR #3962, opened 2018-07-13
Add Amazon SNS notification support.

Ferroin:
The Simple Notification Service (SNS) is a reasonably simple message broker service provided by Amazon as part of it's AWS offerings.

SNS utilizes the concept of 'topics' (similar to Netdata's concept of 'roles' for notifications) to control message routing.  Any given topic may have any number of subscribers of any of the following types..

* Email addresses.
* Phone numbers for SMS.
* HTTP or HTTPS web hooks.
* AWS Lambda endpoints.
* AWS SQS endpoints.
* Mobile applications (via various native push notification services).

Topics are represented by Amazon Resource Names (ARN), which are a special type of URI.

Unfortunately, properly signing requests to AWS endpoints is a serious pain in the arse, so we pretty much have to use the CLI interface. Because of the inflexibility of this tool, setup is somewhat painful.

This uses topic ARN's directly as recipients.  SNS does not support delivery to multiple topics in bulk, so a separate call is required for each topic ARN.  Users are provided with the ability to customize the message format used for the notifications.

------

As mentioned above, setup for this is somewhat complicated (not hard per-se, mostly just tedious).  The AWS CLI assumes that you have your configuration in '~/.aws`, and has no way to change or override this securely (you can provide configuration options as environment variables, but that is a bad idea given that the only two configuration options needed for this are security credentials).  I plan to write up a page on the Wiki going over how to set up a system to send notifications via SNS once this gets merged.

I have not yet done any significant testing of this code.  I plan to do so over the weekend, and will report back with either confirmation that it's working, or with updates to get it working when that's done.

### DIFF
--- a/conf.d/health_alarm_notify.conf
+++ b/conf.d/health_alarm_notify.conf
@@ -14,6 +14,7 @@
 # - sms messages to your cell phone or any sms enabled device (twilio.com)
 # - sms messages to your cell phone or any sms enabled device (messagebird.com)
 # - notifications to users on pagerduty.com
+# - notifications to Amazon SNS topics (aws.amazon.com)
 # - messages to your irc channel on your selected network
 # - messages to a local or remote syslog daemon
 # - message to Microsoft Team (thru webhook)
@@ -84,6 +85,11 @@ nc=""
 # If empty, the system $PATH will be searched for it.
 # If not found, syslog notifications will be silently disabled.
 logger=""
+
+# The full path of the aws command.
+# If empty, the system $PATH will be searched for it.
+# If not found, Amazon SNS notifications will be silently disabled.
+aws=""
 
 #------------------------------------------------------------------------------
 # extra options for external commands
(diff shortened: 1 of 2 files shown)


ktsaou:
this is nice!

Ferroin:
I've unfortunately not had a chance to do proper testing of it yet (my weekend ended up being rather chaotic).  I figured that if I can get it working though, it would be a great benefit for anyone using Netdata on Amazon EC2 nodes.

Ferroin:
OK, I've (finally) gotten this properly tested now with the trivial case of an AWS SNS topic with one subscriber configured for email delivery.  Such a situation is not exactly useful (Netdata's built-in email is exponentially nicer), but I can at least confirm this works to deliver notifications to SNS topics.

Sorry about the delay @ktsaou.

Testing was done as follows:

* I created the aforementioned SNS topic on my AWS account and set up an IAM user that just allowed access to SNS.
* I installed the AWS CLI interface and verified that the above mentioned topic and IAM user worked.
* I updated the `netdata` user on my test system to use a dedicated home directory instead of `/`.  In this case, I set things up to use `/var/lib/netdata`.
* Using `sudo`, I ran the `aws configure` command as the `netdata` user and entered the access key and secret key for the above-mentioned IAM user.  This will create a `.aws` directory in the `netdata` user's home directory (this is why you need the above step of relocating the netdata home directory away from `/`) which contains the credentials (with permissions set so that only the `netdata` user can read them).
* Beyond that, I just set up the alert mechanism like I would for testing any other mechanism, and raun `/usr/libexec/netdata/alarm-test.sh`.

If anyone else wants to test this (SNS usage is free below a certain usage limit per-month) and needs help setting up the SNS topic or IAM role, let me know, I'll be happy to post some more info on that (though the AWS documentation provides good examples on the IAM and SNS setup).

Ferroin:
@ktsaou Unless there's objections from someone, I think this is ready to merge.  I ran some extended testing late yesterday evening and it looks like it's properly handling high delivery volumes, multiple recipient topics, and cross-region delivery.  That covers everything I can think of in terms of actual usage.

ktsaou:
merged! thank you!

Ferroin:
I'll get the page for it up on the Wiki within the next few hours.

Issue #3826, opened 2018-06-14
New at netdata and a lot of alarms


chaosm:
Hi, I'm new on vps and I try to solve many problems.

At nedata i had many events (about 400).
Due to this 
```
# sysctl settings are defined through files in
# /usr/lib/sysctl.d/, /run/sysctl.d/, and /etc/sysctl.d/.
#
# Vendors settings live in /usr/lib/sysctl.d/.
# To override a whole file, create a new file with the same in
# /etc/sysctl.d/ and put new settings there. To override
# only specific settings, add a file with a lexically later
# name in /etc/sysctl.d/ and put new settings there.
#
# For more information, see sysctl.conf(5) and sysctl.d(5).

# Minimizing the amount of swapping
vm.swappiness = 20
vm.dirty_ratio = 80
vm.dirty_background_ratio = 5

# Increases the size of file handles and inode cache & restricts core dumps
fs.file-max = 2097152
fs.suid_dumpable = 0

# Change the amount of incoming connections and incoming connections backlog
net.core.somaxconn = 65535
net.core.netdev_max_backlog = 262144

# Increase the maximum amount of memory buffers
net.core.optmem_max = 25165824

# Increase the default and maximum send/receive buffers
net.core.rmem_default = 31457280
net.core.rmem_max = 67108864
net.core.wmem_default = 31457280
net.core.wmen_max = 67108864

# Enable TCP SYN cookie protection
net.ipv4.tcp_syncookies = 1

# Enable IP spoofing protection
net.ipv4.conf.all.rp_filter = 1

# Enable ignoring to ICMP requests and broadcasts request
net.ipv4.icmp_echo_ignore_all = 1
net.ipv4.icmp_echo_ignore_broadcasts = 1

# Enable logging of spoofed packets, source routed packets and redirect packets
net.ipv4.conf.all.log_martians = 1

# Disable IP source routing
net.ipv4.conf.all.accept_source_route = 0

# Disable ICMP redirect acceptance
net.ipv4.conf.all.accept_redirects = 0
```

i lowered the events at about 40-60

and sometimes pick to 150 

![capture1](https://user-images.githubusercontent.com/15524785/41415267-dc576ec2-6ff0-11e8-99bf-4c1d9d3fa0f4.JPG)

I tried many solutions but nothing works better than these settings.
My main problem is these ipv4 drops

[...]

PR #3976, opened 2018-07-17
relaxed network alarms

ktsaou:
an attempt to make network alarms more reasonable:

1. `1m_ipv4_tcp_listen_drops` is now warning for up to 60 packets
2. `1m_ipv4_udp_receive_buffer_errors` and `1m_ipv4_udp_send_buffer_errors` are now leaving the critical state when the alarm zeros (to avoid flip-flop of critical state when the alarms fluctuate around 100.

Fixes: #3826

### DIFF
--- a/conf.d/health.d/tcp_listen.conf
+++ b/conf.d/health.d/tcp_listen.conf
@@ -20,8 +20,9 @@
   lookup: sum -60s unaligned absolute of ListenDrops
    units: drops
    every: 10s
-    crit: $this > 0
+    warn: $this > 0
+    crit: $this > (($status == $CRITICAL) ? (0) : (60))
    delay: up 0 down 5m multiplier 1.5 max 1h
-    info: the number of TCP listen socket drops during the last minute
+    info: the number of TCP listen socket drops during the last minute (includes bogus packets received)
       to: sysadmin
 
--- a/conf.d/health.d/udp_errors.conf
+++ b/conf.d/health.d/udp_errors.conf
@@ -27,7 +27,7 @@
    units: errors
    every: 10s
     warn: $this > 0
-    crit: $this > 100
+    crit: $this > (($status == $CRITICAL) ? (0) : (100))
     info: number of UDP receive buffer errors during the last minute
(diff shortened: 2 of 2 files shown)


ktsaou:
@Ferroin do you agree with these changes?

Ferroin:
I definitely agree on the TCP one.  The other two look sane to me as well (I've never had the issues with them that this is fixing, but I've had such issues with other alarms on occasion, so I entirely understand the motivation).

ktsaou:
Can you help tune the alarms to eliminate false positives and the noisy ones?
If yes, please open a github issue for each one you believe needs fixing, so that we can discuss them and find solutions.

Ferroin:
@ktsaou Yeah, I'd be happy to help with tuning them.  I've actually been meaning to open some issues regarding this, it's just that I keep getting distracted by stuff I consider higher priority (false positive alarms are mostly an annoyance for me, not really a huge problem).

spirogg:
Hi I just updated to latest version and still getting emails 10-20 a day 


ipv4.tcplistenissuesCHART
--
1m ipv4 tcp listen drops = 0.55 drops the number of TCP listen socket drops during the last minute (includes bogus packets received)ALARM
tcpFAMILY
WARNINGSEVERITY
Fri Jul 27 19:27:54 CDT 2018TIME

what is the final fix ? sorry but I read alot and confused how to fix this please

thanks

Spiro

ktsaou:
Well, this PR made it a warning when > 0 and < 60 per minute.

The fact is, the system is dropping packets. The best case is that they are bogus packets the system just rejected. The worst case is that there is a problem with a server you run and packets are dropped.

In many cases of bogus packets, If you enable the iptables connection tracker, the alarm will not trigger (the connection tracker will drop the packets before entering the host), so this alarm will only trigger if you really have a problem.

I am open to suggestions. I am not sure we should just ignore the dropped packets.

@Ferroin I need your help in this. Let's assume we silence this alarm. What will we loose? There is an overflow alarm too, that will trigger if a listening server is not capable of processing all its workload. Is there any other case this alarm is useful?

PR #3987, opened 2018-07-20
Fix config path

roedie:
Since the config is expected to be located in  /etc/netdata it is kind 
of silly to use ../../../../ to get to the right path. Moreover it 
breaks netdata on Debian.

### DIFF
--- a/plugins.d/python.d.plugin
+++ b/plugins.d/python.d.plugin
@@ -18,7 +18,7 @@ from time import sleep
 
 
 PY_VERSION = version_info[:2]
-PLUGIN_CONFIG_DIR = os.getenv('NETDATA_CONFIG_DIR', os.path.dirname(__file__) + '/../../../../etc/netdata') + '/'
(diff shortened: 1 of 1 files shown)


l2isbad:
>  Moreover it breaks netdata on Debian.

debug mode. Netdata works.

In general i agree, path finding in debug mode should be fixed (howerver current  implementration works in most of the cases).

But your PR is wrong, it breaks much more then fixes.

l2isbad:
> Does netdata use different path finding technique when in debug mode?

Year, it looks for env var `NETDATA_CONFIG_DIR` first and if it fails as a fallback it uses `os.path.dirname(__file__) + '/../../../../etc/netdata'`. It works when netdata installed in `/opt/netdata/`

l2isbad:
@roedie 
have a look
https://github.com/firehol/netdata/issues/3854

roedie:
@l2isbad Thx

PR #4006, opened 2018-07-25
Add support for float values for time_multiplier in web_log

### LINKED ISSUE #4003: Web_log doesn't support response times in nanoseconds
Plugin `web_log` only supports response times in microseconds. In order to convert different types of units to microseconds, it provides an extra option (called `time_multiplier`) to multiply the given number. However, it only supports integer type. So if the response time is in nanoseconds, I can't set `time_multiplier` to `0.001`. 

Could you please support the float type for option `time_multiplier` as well? 

https://github.com/firehol/netdata/blob/master/conf.d/python.d/web_log.conf#L87
```
#     custom_log_format:                  # define a custom log format
#          time_multiplier: 1000000       # type <int> - convert time to microseconds
```

https://github.com/firehol/netdata/blob/master/python.d/web_log.chart.py#L632
```python
        resp_time_func = self.configuration.get('custom_log_format', dict()).get('time_multiplier') or 0

        if not isinstance(resp_time_func, int):
            return find_regex_return(msg='Custom log: "time_multiplier" is not an integer')
```

alibo:
It fixes #4003. 

After merging this PR, response times in nanoseconds will be supported as well. You just need to set `time_multiplier` to `0.001`.

### DIFF
--- a/python.d/web_log.chart.py
+++ b/python.d/web_log.chart.py
@@ -629,7 +629,7 @@ class Web:
 
         resp_time_func = self.configuration.get('custom_log_format', dict()).get('time_multiplier') or 0
 
-        if not isinstance(resp_time_func, int):
+        if not isinstance(resp_time_func, (int, float)):
             return find_regex_return(msg='Custom log: "time_multiplier" is not an integer')
 
         try:


alibo:
I'm sorry I forgot to change the comments and the error message. You may want to squash the commits and merge :)

ktsaou:
@alibo please add yourself to CONTRIBUTORS.md, to sign the CLA.
Then we can merge this.

alibo:
@ktsaou Signed :)

PR #4008, opened 2018-07-25
[WIP] Add a python.d module for Yoctopuce sensor units.

### LINKED ISSUE #4002: [Feature Request] Support for Yoctopuce USB sensor modules.
[Yoctopuce](http://www.yoctopuce.com/) produces a line of very well made and rather inexpensive USB connected devices, including a wide range of sensor interfaces.  It would be awesome if Netdata supported reading and charting data from these.

------

This isn't as much a feature request as it is me saying I'm planning on working on this.  I don't currently have hardware I can test with, but I will be purchasing one of their meteorological sensor modules (and possibly an an ALS and thermocouple module too) in the next few weeks, which will give a reasonable base to test the basic code with.

My plan is to write up a python.d module that interfaces with their [VirtualHub](http://www.yoctopuce.com/EN/virtualhub.php) software, which is generally considered to be the correct way for generic projects to access their modules (because their native interfaces are not safe for concurrent access).

Ferroin:
This adds support for monitoring data from various sensor modules produced by Yoctopuce.  These modules are designed for automation and monitoring setups, so they seem like a logical choice to use with Netdata.

This uses the official Yoctopuce Python API to interact with the sensors.  This API is unfortunately not readily available on most distributions, but it's also a C-based (ctypes used to interact with a bundled shared library from the C API), so it isn't something that I think makes sense to include in Netdata.

Support is provided for all the currently manufactured sensor modules that provide a single numerical reading for each contained sensor.

The modules themselves connect to a system over USB.  They are not concurrent access safe though.  Yoctopuce provides a piece of software called Virtual Hub that allows easy mediation of access by multiple programs to the same modules.  By default, we try to connect to an instance of Virtual Hub on the local system.  The Virtual Hub interface also works over the network, and is additionally provided by all of the hardware Hub modules that Yoctopuce provides.

By default, we scan for all supported sensors connected to the hub, and create a dimension for each.  Users can choose to disable this and instead ask for a specific set of sensors as well.

Fixes #4002 

---

This is a work-in-progress.  I've not yet had the opportunity to test it (I will be getting the hardware to do so around this time next month).  I'm hoping for some early review though, and figure that if anyone has hardware themselves that they want to test it.

At the moment, this includes support for eight sensor types:

* Light sensors (all of what Yoctopuce offers are ALS modules, but they might do other light sensors in the future).
* Temperature sensors (including their thermistor, thermocouple, and PT100 modules).
* Humidity sensors (%RH only, because that's all they have sensors for).
* Pressure sensors.
* Current sensors.
* Voltage sensors.
* VOC sensors.
* CO2 sensors.

I will only be able to test a subset of these, but how we interact with the sensors is entirely agnostic of what type of sensor it is (in fact, the all the type specific stuff is encapsulated in the `check()` method, and all of it is computed at runtime), so I'm not too worried about it working once I've got the stuff I will be able to test tested.

In the interest of keeping memory usage down (and because we only need one direct reference to the main class in the module for each type of sensor we support), imports of everything except the core API are done dynamically and encapsulated in such a way that they get unloaded if not needed.  When combined with the fact that the code is type agnostic, the  means that it's trivial to add support for new sensor types (it consists of just adding an item to `ORDER` and a set of entries in `NAME_MAP`).

Autodetection will produce dimension names based on the 'friendly' name for the device.  This will result in dimension names that look like either `THRMCPL1-6D302.temperature1` (for modules with no user defined logical name) or `MyName.sensor` (for modules with a user defined logical name).

Currently, this assumes all the sensors of a given type are using the same units for reporting as the first sensor of that type.  This is generally a reasonable assumption to make, and people will have to go out of their way to break it.


Ferroin:
Rebased to rerun checks with the most recent configuration.  Hopefully, I should finally be able to get hardware within a few weeks to test this.

ktsaou:
nice. please rebase this.

Disabling a module by adding `module: no` to `python.d.conf` will not work.
All those that have modified this file in the past, will not get that little update.

@l2isbad please find a way to have a module disabled by default, without putting it in `python.d.conf`.

l2isbad:
@ktsaou 
we have such option
https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/logind/logind.chart.py#L9

ktsaou:
ok, then @Ferroin please set that option.

Ferroin:
Updated now to use the  new code structure, and actually be disabled by default.

I still have not had a chance yet to test this though, and would like to hold off on merging until I can get it properly tested (which unfortunately is dependent on when I'm able to get the hardware to test it, which should hopefully be soon).

l2isbad:
>  I should have it by the end of the month, and once I get it it should only take me a day or two to get this tested.

Have you gotten it?

cakrit:
Any news on this? We could purchase this hardware too and send it to @ilyam8 to play with it, I'm sure he'll enjoy it! :)

Ferroin:
Last time I tried to get this working, the issue was with getting python.d to properly load the Yoctopuce modules.  As far as I could tell at the time, the dynamic import thing works, it's just an issue with getting Python to properly import modules that aren't in `system.path`, but I didn't look very deep beyond that.  Part of the problem there is that the Yoctopuce Python modules are not available through most distribution's package managers, which adds a layer of complexity to things.

The actual data collection logic works (I've tested that independently of the plugin itself), so the only issues are the above-mentioned import issue and any possible issues in the chart construction.

Just like #3956 though, I unfortunately don't really have the time to work on this right now.  I'll be happy to help when I can (including testing) if someone else wants to work on it.

PR #4024, opened 2018-08-01
python.d: add 'hidden' option to charts

l2isbad:
quick fix #4023

PR #4027, opened 2018-08-02
python plugin for monit (draft)

n0guest:
Hello.

Here is a draft version of python plugin for [Monit](https://mmonit.com/monit/). For now this plugin collects statuses of all checks (process/host/filesystem/etc) except [main system](https://mmonit.com/monit/documentation/monit.html#System). I.e. it's only metrics which aren't exists in Netdata already. Plus it exports few additional metrics for process/hosts checks. But to work correctly it depends on PR #4024 (by @l2isbad), because it uses hidden chart for "processes uptime".

Please, consider including it in upstream. If you have any notes on this code or there is something wrong - let me know and i'll fix.

Thank you.


python.d/monit.chart.py
```diff
@@ -0,0 +1,146 @@
...
+        self._manager = self._build_manager()
+        raw_data = self._get_raw_data()
+        if raw_data:
+            try:
+                xml = ET.fromstring(raw_data)
+            except ET.ParseError:
+                self.debug('%s is not a vaild XML page. Please add "_status?format=xml&level=full" to monit URL.' % self.url)
+                return None
+            return True
+        return None
```
l2isbad: please avoid such netsted constructions, the are always hard to follow

```python
if not raw_data:
    retrun None
....
....
```

n0guest:
Hmm ... looks like Codacy didn't like it. But, to be honest, this plugin was written on base of tomcat.chart.py as the closest example to "parse XML in UrlService class". So, tomcat also wouldn't pass the same test? 0_o

python.d/monit.chart.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _get_data(self):
+        data = None
```
l2isbad: Same here + split  ` _get_data` function into several smaller ones.

```python
if not raw_data:
    return None

parsed = parse(raw_data)
if not parsed:
    return None
...
```

python.d/monit.chart.py
```diff
@@ -0,0 +1,146 @@
...
+            try:
+                xml = ET.fromstring(raw_data)
+            except ET.ParseError:
+                self.debug('%s is not a vaild XML page. Please add "_status?format=xml&level=full" to monit URL.' % self.url)
+                return None
+            data = {}
+
+            for svc_id in DEFAULT_SERVICES_IDS:
+                if svc_id not in DEFAULT_SERVICES_IDS:
+                    continue
```
l2isbad: This doesn't make any sense

n0guest: Yep, it's nonsense, you're right. It was supposed to be configured via conf file. For example, if user don't want to export all checks. I can just remove it for now and implement it later. Ok?

l2isbad: I mean
```for svc_id in DEFAULT_SERVICES_IDS```
```if svc_id not in DEFAULT_SERVICES_IDS```

Same `DEFAULT_SERVICES_IDS`, so ` svc_id` is always in `DEFAULT_SERVICES_IDS`

Comment (n0guest) [reply python.d/monit.chart.py]: > Same DEFAULT_SERVICES_IDS, so svc_id is always in DEFAULT_SERVICES_IDS

It's exactly what i meant. For now i'll just delete 106-107. Ok?

Comment (l2isbad) [reply python.d/monit.chart.py]: Ok

File: python.d/monit.chart.py (hunk 1 above)
Comment (n0guest) [reply python.d/monit.chart.py]: Ok, i'll do.

File: python.d/monit.chart.py (hunk 3 above)
Comment (n0guest) [reply python.d/monit.chart.py]: Removed

File: python.d/monit.chart.py (hunk 1 above)
Comment (n0guest) [reply python.d/monit.chart.py]: Fixed. It's ok now?

File: python.d/monit.chart.py (hunk 2 above)
Comment (n0guest) [reply python.d/monit.chart.py]: Not sure if did exactly how you wanted, but, please, check it in last commit.

File: python.d/monit.chart.py [hunk 4]
```diff
@@ -0,0 +1,146 @@
...
+        self._manager = self._build_manager()
+        raw_data = self._get_raw_data()
+        if not raw_data:
+            return None
+        try:
+            xml = ET.fromstring(raw_data)
+            return True
+        except ET.ParseError:
+            self.debug('%s is not a vaild XML page. Please add "_status?format=xml&level=full" to monit URL.' % self.url)
+            return None
```
l2isbad: i see same code (xml parse) in check and in get_data

i suggest to repace it with `parse` func

then `return bool(parse(raw_data))` in check

n0guest: Ok, lets try again)

python.d/monit.chart.py
```diff
@@ -0,0 +1,146 @@
...
+        'lines': []
+    },
+}
+
+class Service(UrlService):
+    def __init__(self, configuration=None, name=None):
+        UrlService.__init__(self, configuration=configuration, name=name)
+        self.url = self.configuration.get('url', "http://127.0.0.1:2812/_status?format=xml&level=full")
+        self.order = ORDER
+        self.definitions = CHARTS
```
l2isbad: copy `ORDER` and deepcopy `CHARTS`

l2isbad:
Can you post few screenshots?

python.d/monit.chart.py
n0guest: I'm sorry, but i don't understand what you mean. I'm new in python, learned 2 days ago just for purpose of writing this plugin. Could you, please, explain little more details?

n0guest:
NP, how big it should be? It's just for "see how it works" or for wiki?

python.d/monit.chart.py
n0guest: You meant something like this?

```
import copy

self.order = copy.copy(ORDER)
self.definitions = copy.deepcopy(CHARTS)
```

l2isbad: I was wrong, actually it is not needed :smile:

n0guest: Ok, so that's all?

l2isbad: Well, as i said earlier - `get_data` method is hard to read (don't you think so?), can you simplify it?

Ferroin:
Both of the codacy problems are functionally non-issues (fixing them properly would require adding a new dependency to Netdata, which is not exactly something taken lightly, but the problems themselves are non-trivial to exploit, so it's not really a big issue).

Ferroin:
You're missing a handful of necessary changes.  Please check the [new module PR checklist](https://github.com/firehol/netdata/wiki/New-Module-PR-Checklist) and make sure everything on there is taken care of.

Also, please sign the CLA (add your info to the table at the end of the `CONTRIBUTORS.md` file).

Ferroin:
Does anyone here happen to know if urllib supports usage of UNIX sockets?  If not, I'd actually argue for having this use `SocketService` instead, even though it's not quite as trivial to implement that way, because it's not unusual for people using monit to use it over a UNIX socket.

l2isbad:
We can add support for `SocketService` later. Like in `haproxy` module.

l2isbad:
I am sying it because

> i'm new in python, learned 2 days ago just for purpose of writing this plugin
> even though it's not quite as trivial to implement that way,

Ferroin:
@l2isbad _PLEASE_ don't add more multiple-inheritance based polyinstantiation like the HAProxy module does, it's confusing, error-prone, horrible for memory usage, and prevents a couple of otherwise simple memory usage improvements for the whole python.d framework from being viable (no, seriously, I've got a couple of things written up that do improve memory usage, but they break the HAProxy module because of the multiple-inheritance).  I've been trying for about a month now to come up with a way to get rid of the multiple-inheritance in the HAProxy module without making things significantly more confusing or breaking the configuration format.

In this case, I can probably adapt this to use `SocketService` directly once merged without breaking the configuration format.

l2isbad:
I agree with confusing, error-prone, but not with `horrible for memory usage`. What do you mean by `horrible`. Can you show the numbers?

The problem in general is the design itself - plugin imports all modules and there is no opportunities to cleanup it. So it consumes 20-30 MB of ram. Micro tunes do not help much

Ferroin:
Micro-optimizations don't help much for big server systems with lots of memory.  They do help for IoT systems and container usage, as well as some cloud usage, all of which are cases where you're generally memory constrained.  If, for example, the `isc_dhcp` module were to be using `__slots__` for it's various objects, that would translate to about 100 bytes of space savings per lease, which translates to about 1MB of runtime memory savings for a not unusual case of 10 000 leases.

For the specific case of the HAProxy `Service` class, there's about a dozen methods that it ends up with which may or may not be used, each of which being a few kB of runtime memory that may just be dead code.  Yes, that doesn't sound like much, but on stuff like routers that's a sizable percentage of memory.

Ferroin:
Let me explain this a bit differently:  We've got stuff wasting memory.  Some of these things happen to be trivial to fix without breaking anything or adding almost any maintenance overhead.  Given this, we should fix those things, even if the net benefit is minimal, because the cost is essentially zero.

I've actually run the numbers for all the modules we have currently in netdata based on 64-bit CPython 3.6.  We've got multiple megabytes wasted before even doing anything other than instantiating all the `Service` classes just from unused dictionary slots, and that's not even counting any of the third-party code.

l2isbad:
>  suggest to use ``__slots__``
> no maintenance overhead

Oh you! :smile: 

I got your point.

python.d/monit.chart.py
n0guest: Hmm ... Let's make sure we "dot the i's and cross the t's". I'm not a developer, in fact i'm sysadm/sysop/etc. So, if you would be so kind and guide me - i'd be glad to make required changes. In other way - i'll be "shooting in the dark" and wasting both your's and mine time.

Could, please, you provide little more details on this matter?

n0guest:
> You're missing a handful of necessary changes. Please check the new module PR checklist and make sure everything on there is taken care of.
> Also, please sign the CLA (add your info to the table at the end of the CONTRIBUTORS.md file).

@Ferroin, sure do. As soon as i'll pass through "code requirements" above. Until that moment signing or commiting anything could be useless.

python.d/monit.chart.py
l2isbad: > I'm not a developer

So do i.

> Could you, please, provide little more details on this matter?

Ok, use `format` instead of `%`. Then it will be acceptable.

n0guest: Ok, done. Please, check it.

python.d/monit.chart.py
```diff
@@ -111,14 +111,14 @@ def _get_data(self):
                 self.debug("Skipping service from 'System' category, because it's useless in graphs")
                 continue
 
-            xpath_query = "./service[@type='%d']" % (service_id)
-            self.debug("Searching for %s as %s" % (service_category, xpath_query))
+            xpath_query = "./service[@type='{:d}']".format(service_id)
+            self.debug("Searching for {} as {}".format(service_category, xpath_query))
```
l2isbad: Oh, please  give placeholders an explicit positional index, like  `{0} {1}`.  (py2.6 compatibility)

n0guest: Ahah ... Added. And I hope i understood you right)

python.d/monit.chart.py
```diff
@@ -0,0 +1,150 @@
...
+        UrlService.__init__(self, configuration=configuration, name=name)
+        self.url = self.configuration.get('url', "http://127.0.0.1:2812/_status?format=xml&level=full")
+        self.order = ORDER
+        self.definitions = CHARTS
+
+    def parse(self, data):
+        try:
+            xml = ET.fromstring(data)
+        except ET.ParseError:
+            self.debug('{} is not a vaild XML page. Please add "_status?format=xml&level=full" to monit URL.'.format(self.url))
```
l2isbad: here

python.d/monit.chart.py
```diff
@@ -0,0 +1,150 @@
...
+                        dimension_key = 'process_{0}_{1}'.format(subnode, service_name)
+                        if dimension_key not in self.charts['process_' + subnode]:
+                            self.charts['process_' + subnode].add_dimension([dimension_key, service_name, 'absolute'])
+                        data[dimension_key] = int(subnode_value.text)
+
+                if service_category == 'host':
+                    subnode_value = service_node.find('./icmp/responsetime')
+                    if subnode_value == None:
+                        continue
+                    dimension_key = 'host_latency_{}'.format(service_name)
```
l2isbad: and here

python.d/monit.chart.py
```diff
@@ -0,0 +1,150 @@
...
+            return None
+
+        data = {}
+        for service_id in DEFAULT_SERVICES_IDS:
+            service_category = MONIT_SERVICE_NAMES[service_id].lower()
+            if service_category == 'system':
+                self.debug("Skipping service from 'System' category, because it's useless in graphs")
+                continue
+
+            xpath_query = "./service[@type='{:d}']".format(service_id)
```
l2isbad: one more

l2isbad:
> NP, how big it should be? It's just for "see how it works" or for wiki?

I'm just curious, "see how it works".

n0guest:
> I'm just curious, "see how it works".

Here how it looks:

screenshot-20180803-085720

I couldn't find better idea how to visualize on/off states (as "stacked" it looks awful).

python.d/monit.chart.py
n0guest: done

l2isbad:
@Ferroin please have a look, maybe you have some comments.

@n0guest gj, time for [New Module PR Checklist](https://github.com/firehol/netdata/wiki/New-Module-PR-Checklist)

> Optionally, chart information in web/dashboard_info.js. This generally involves specifying a name and icon for the section, and may include descriptions for the section or individual charts.

^^ please do :+1:

python.d/monit.chart.py
```diff
@@ -0,0 +1,150 @@
...
+    'net': {
+        'options': ['interfaces', 'Network interfaces and addresses', 'interfaces', 'network', 'monit.networks', 'line'],
+        'lines': []
+    },
+}
+
+class Service(UrlService):
+    def __init__(self, configuration=None, name=None):
+        UrlService.__init__(self, configuration=configuration, name=name)
+        self.url = self.configuration.get('url', "http://127.0.0.1:2812/_status?format=xml&level=full")
```
Ferroin: I assume that you only support parsing the full XML format?  If that's the case, then at minimum the query string (the `?format=xml&level=full` bit) should be hard-coded in the module and not required as part of the URL in the configuration.  I'd say that the `/_status` part probably should be too, as that's part of the monit API, and not something that's likely to be changed by an end user.

Also, please update the IP address to a host name.  The `UrlService` class will handle that correctly with very little overhead, and that way people who are actually running monit on IPv6 over the loopback interface will actually have IPv6 used.

Essentially, something like this, and then update the default value explained in the config file accordingly:

```
    uri = self.configuration.get('uri', 'http://localhost:2812')
    self.uri = format('{0}/_status?format=xml&level=full', uri)
```

python.d/monit.chart.py
```diff
@@ -0,0 +1,150 @@
...
+        self.order = ORDER
+        self.definitions = CHARTS
+
+    def parse(self, data):
+        try:
+            xml = ET.fromstring(data)
+        except ET.ParseError:
+            self.debug('{0} is not a vaild XML page. Please add "_status?format=xml&level=full" to monit URL.'.format(self.url))
```
Ferroin: Building on the comment above, this should probably be made more generic.

Even if you don't go with hard-coding the path and query string, this should not suggest adding them, because there are plenty of other reasons this error condition could fail.

If you _really_ wanted to do this and not hard-code the URL, then you should be checking the URL itself, and deciding whether to use this or a more generic message based on that.

Ferroin: Also, I'd say this should be more than just a debug message.  It means that you can't collect data at all, so an error or warning makes more sense (while the python.d.plugin code will print an error message if the check fails, having an explanation for _why_ it failed without having to go through manually debugging the plugin is much friendlier for users).

Ferroin:
Besides the couple of specific comments above, this actually looks rather good.  I've tested it on my own systems a bit, and while I can't say that any of the stats it provides will be particularly useful for me, it does work as advertised, and I was actually rather impressed by how well it performs given all the XML parsing it's doing.

n0guest:
> I can't say that any of the stats it provides will be particularly useful for me

We have many installations of Netdata on servers where we're also using them as exporters for Prometheus. So, our usage case - to have "Services status" (running, died, etc) in same Grafana's dashboard where we already have "Services metrics" from Netdata. And, of course, alerts too, based on same "Services statuses".

Ferroin:
@n0guest Oh, believe me, I know there are use cases, it's just that for me, most of the data is not likely to be amazingly useful.

python.d/monit.chart.py
l2isbad: @n0guest 
replace `url` with `host`, `port` and `scheme`, this whould be better i think
```python
scheme = self.configuration.get('scheme', 'http')
host = self.configuration.get('host', 'localhost')
port = self.configuration.get('port', 2812)
self.url = '{0}://{1}:{2}/_status?format=xml&level=full'.format(scheme, host,  port)
```

python.d/monit.chart.py
l2isbad: Since it is `ET.ParseError` `is not a vaild XML page` seem enough, but yes, why not to log the error.

`except ET.ParseError as error:`

n0guest:
> @n0guest gj, time for New Module PR Checklist
> ^^ please do +1

Done. Please, check.

python.d/monit.chart.py
n0guest: >  It means that you can't collect data at all, so an error or warning makes more sense (while the python.d.plugin code will print an error message if the check fails, having an explanation for why it failed without having to go through manually debugging the plugin is much friendlier for users).

Yep, will do. But only one question. Is it ok to use self.warning() for this or there are any limitations for this? I looked into other modules and didn't find any usage of "warning" method. Or this should be an error (as it's "incorrect configuration")?

Please, clarify.

l2isbad: i believe `/_status` thing in `apache`, `nignx`, `hapoxy` is configurable, then it shouldn't be static.

what about `monit`? does it always `_status?format=xml&level=full` ?

python.d/monit.chart.py
l2isbad: i believe `/_status` thing in `apache`, `nignx`, `hapoxy` is configurable, then it shouldn't be static.

what about `monit`? does it always `_status?format=xml&level=full` ?

python.d/monit.chart.py
l2isbad: Yes, for some reason nobody use `warning` :smile: Yes, it is safe. 
But this should be an error.

python.d/monit.chart.py
n0guest: Yes, it's only way. To say more, it's not even documented in [official docs](https://mmonit.com/monit/documentation/monit.html#Service-checks), but mentioned many times in [issues on bitbucket](https://bitbucket.org/tildeslash/monit/issues?q=format%3Dxml). So, it's half-official, but only one known way to get stats from Monit.

PS: At stackoverflow there were mentions, that monit also supports JSON (as format=json), but i can't get it working and proceed from parsing XML.

l2isbad: Ok, another option is to leave it as is (`url') and add a simple check, like

```python
def check(self):
    if not self.url.endswith('_status?format=xml&level=full'):
        self.error('text')
        return False
```

python.d/monit.chart.py
n0guest: Done it. Small example:

```
# head -n 10 log
2018-08-05 12:52:15: python.d INFO: plugin: main: Using python 3
2018-08-05 12:52:16: python.d INFO: monit: localhost: check() => [OK]
2018-08-05 12:52:16: python.d ERROR: monit: remote: Url: http://10.177.20.1:2812/_status?format=xml&level=full. Error: HTTPConnectionPool(host='10.177.20.1', port=2812): Max retries exceeded with url: /_status?format=xml&level=full (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fb7200be780>: Failed to establish a new connection: [Errno 111] Connection refused',))
2018-08-05 12:52:16: python.d INFO: monit: remote: check() => [FAILED]
> /usr/lib/x86_64-linux-gnu/netdata/python.d/monit.chart.py(92)parse()
-> self.error("URL {0} didn't return a vaild XML page. Please check your settings.".format(self.url))
(Pdb) 2018-08-05 12:52:17: python.d ERROR: monit: error: URL http://localhost:19999 didn't return a vaild XML page. Please check your settings.
2018-08-05 12:52:17: python.d INFO: monit: error: check() => [FAILED]
CHART netdata.runtime_monit_local '' 'Execution time for monit_local' 'ms' 'python.d' netdata.pythond_runtime line 145000 1
```

Plus also updated python.d/README.md and conf.d/monit.conf.

python.d/monit.chart.py
n0guest: I did as @Ferroin initially suggested (see 6e023743d5154cc0a21a03d537a1ca7b77f03ab3)

n0guest:
I hope i made all required and suggested changes. Is there anything else?

n0guest:
> Also, please sign the CLA (add your info to the table at the end of the CONTRIBUTORS.md file).

Looks like I missed it. Done now.

n0guest:
So, what now? Are we waiting for a feedback/merge by @ktsaou?

Ferroin:
Yep, he's got the final say on what gets merged or not (though I doubt that he'll object to this).

ktsaou:
merged it.
Thank you!

Ferroin:
Yes, they will be included in the next release.  @ktsaou would be the one to ask about when that will be, but I don't think he's decided for certain yet.

That said, you can run them now by just running the current master branch from GitHub.  It's (usually) stable enough for production usage.

ktsaou:
I think early September we will release 1.11

PR #4030, opened 2018-08-02
Chart MySQL active connections vs connection limit and add health check

amichelic:
This pull request adds a chart which shows:
- the number of currently active (open) connections,
- the configured maximum limit of connections,
- the maximum number of concurrent connection since mysql daemon start.

In addition, a health check is added that starts warning when for than 70% of the available connections slots are used.


ktsaou:
merged it.
Thank you!

PR #4052, opened 2018-08-12
Replace embedded keys and transition my key

philwhineray:
Individual commit logs contain detailed information


philwhineray:
Having a torrid time getting this to build. Looks to be network connectivity problems, finishing up with being unable to connect to ftp.gnu.org (unchanged from previous successful builds) so will try restarting this later...

philwhineray:
Bash FTP is up and running, build now succeeds.

@ktsaou there isn't a rush on merging these, since I don't expect to be tagging netdata, but I'd appreciate their inclusion at a convenient time.

ktsaou:
ok. Do I have to upload my public key somewhere?
Can I test this somehow?

philwhineray:
I replaced your key with its fingerprint and checked it was available to
download, so everything should be ok.

The commit logs have instructions to check I have not tried to defeat the
existing keys.

If there are any problems it will be when you next tag a version but it
should just work.

ktsaou:
ok, merged it.

PR #4055, opened 2018-08-13
proc_interrupts/softirqs parser fix

### LINKED ISSUE #4051: /proc/interrupts plugin memory leak
Recently noticed this odd behavior on one of my ARM64 servers running netdata `v1.10.0-537-g6832ce4_rolling`.

If `plugin:proc:/proc/interrupts` is enabled, netdata memory usage appears to increase until the system OOM killer kills the netdata process. If I refresh the dashboard, the `###` in `"netdata collects every second ### metrics"` appears to be increasing by 5 every data collection interval, +1 per interrupt graph (`system.interrupts`, `cpu.cpu#_interrupts` for CPUs 0-3). After a restart yesterday, it went from ~350 metrics to over 50,000 metrics in a couple hours, eating up over 600MB RAM.

I've been unable to reproduce the behavior on a second server, but the issue persists across complete netdata re-installs on this server alone. Disabling the `/proc/interrupts` metric appears to work as a temporary fix, but any pointers on figuring out what's going wrong would be appreciated. Thanks.

abalabahaha:
For a system with insanely high interrupts/softirqs, netdata appears to parse tokens like `IPI0:1145401553` as a single word, inadvertently creating new dimensions every time the file is parsed.

It looks like a couple other `/proc/*` separators were changed to include the colon (`" \t:"`), but `/proc/interrupts` and `/proc/softirqs` monitors do not (`" \t"`).

Fixes #4051

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -104,3 +104,4 @@ username|name|email (optional)
 @rda0|Sven Mäder|maeder@phys.ethz.ch
 @alibo|Ali Borhani|aliborhani1@gmail.com
 @Nani-o|Sofiane Medjkoune|sofiane@medjkoune.fr
+@abalabahaha|abalabahaha|hi@abal.moe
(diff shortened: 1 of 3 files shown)


Ferroin:
Great catch.  Tested before and after here with some artificial `/proc/interrupts` and `/proc/softirqs` files, and it works as advertised.

One last thing though, please sign the CLA by adding an entry to the end of the table in `CONTRIBUTORS.md`.

ktsaou:
nice finding and fix!
thank you!

PR #4093, opened 2018-08-29
[nginx_plus] fix handling of non-contiguous peer IDs

illes:
`peers_stats()` mistakenly assumed that the `/status` REST response contains peers with zero-based contiguous IDs, which is not the case when upstreams are changed dynamically (via dashboard or API).


ktsaou:
Thank you @illes 
Please sign the netdata CLA by adding yourself to the CONTRIBUTORS.md file of this repo. Just add your name there and commit it like the above, in this PR.

l2isbad:
Seem http://demo.nginx.com/status is no longer working :cry: And i have no chance to test it (`nginx_plus` is non free)

illes:
@l2isbad : the `/status` API has been deprecated in the latest nginx-plus R16 (which [powers the demo site](http://demo.nginx.com/api/3/nginx)).

The [new API](http://demo.nginx.com/swagger-ui/) has a different logic, see e.g. http://demo.nginx.com/api/3/http/upstreams

We will have to consider moving to this new API.

l2isbad:
> We will have to consider moving to this new API.

We need to support both, right?

ktsaou:
> We need to support both, right?

I am afraid so...

PR #4178, opened 2018-09-12
web files should be owned by root:netdata

### LINKED ISSUE #4173: web files should be owned by root:netdata, with permission 0644
I'm having trouble accessing the web server which is running by the user netdata and the files in /usr/share/netdata are owned by root.

```
File '/usr/share/netdata/web//index.html' is owned by user 0 (expected user 411). Access Denied.
```
I see at this [comment](https://github.com/firehol/netdata/issues/52#issuecomment-177582297) that this is a "security rule". How can it be more secure that the webserver owns and can modify its own files? I classify this more like a security hole, because any bug in the webserver software allow a remote user to modify the files to whatever purpose.

ktsaou:
fixes #4173

NOT FINISHED YET

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -699,14 +699,8 @@ NETDATA_GROUP="$(id -g -n ${NETDATA_USER})"
 [ -z "${NETDATA_GROUP}" ] && NETDATA_GROUP="${NETDATA_USER}"
 
 # the owners of the web files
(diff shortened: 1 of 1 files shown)


cakrit:
Is this one still not finished? Perhaps give it to @paulfantom?

ktsaou:
We will remove this check completely with #1394 
So, I am closing this.

PR #4180, opened 2018-09-12
fix for load alarms

ktsaou:


Ferroin:
Oops, sorry I misunderstood how the new variable was supposed to be accessed.

I'm actually curious though, why `alarm` instead of `template` like almost everything else?

ktsaou:
Well, `template` is when you want the same alarm on a number of charts. So one alarm for all your disks, all your mysql servers, all your web_log charts, etc.

The problem with the previous alarms was two-fold:

1. It was using `template`, but this did not harm at all
2. It was using `$load_trigger` cross family. It should use `$system.cpu.load_trigger` or move `load_trigger` to `system.load` chart, which I preferred.

For some exotic reason I have to debug, the variable `$load_trigger` was then evaluated to `NaN`, but somehow the alarm was assuming it was 0. It should not. It should `NaN` the alarm too. I have to test this to find why.

PR #4185, opened 2018-09-13
pythond_small_fixes

### LINKED ISSUE #4089: update_every in postgres plugin
Hi,

I'm having an issue with update_every in the postgres plugin. When setting update_every to 10 in the postgres conf I thought it would run every 10 seconds. But it seems to run every 2 seconds. Am I misunderstanding this?

In the netdata global settings I have:
```
history = 1800
update every = 2
```

But in postgres.conf I have:
```
postgres:
  update_every: 10
  name: 'local'
  database: 'postgres'
  host: '127.0.0.1'
  user: 'myleetuser'
  password: 'verycoolpass'
  port: '5432'
```

l2isbad:
- fixes #4089
- workaround for #3817 (run `gc.collect()` every 300 seconds, can be disabled in `python.d.conf`)
- debug mode fix (repsect enabled/disable module state)

Still can't reproduce it myself but seems it (partially) works. Temporary workaround is better then memory leak

### DIFF
--- a/plugins.d/python.d.plugin
+++ b/plugins.d/python.d.plugin
@@ -17,6 +17,7 @@ from re import sub
 from sys import version_info, argv
 from time import sleep
 
+GC_RUN = True
 GC_COLLECT_EVERY = 300
 
 PY_VERSION = version_info[:2]
@@ -204,6 +205,8 @@ class Plugin(object):
         self.config, error = self.loader.load_config_from_file(PLUGIN_CONFIG_DIR + 'python.d.conf')
         if error:
             Logger.error('"python.d.conf" configuration file not found. Using defaults.')
+        self.do_gc = self.config.get("gc_run", GC_RUN)
+        self.gc_interval = self.config.get("gc_interval", GC_COLLECT_EVERY)
 
         if not self.config.get('enabled', True):
             run_and_exit(Logger.info)('DISABLED in configuration file.')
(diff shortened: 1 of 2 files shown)


plugins.d/python.d.plugin
```diff
@@ -85,7 +87,7 @@ class Job(object):
         self.recheck_every = self.job.configuration.pop('autodetection_retry')
         self.checked = False  # used in Plugin.check_job()
         self.created = False  # used in Plugin.create_job_charts()
-        if OVERRIDE_UPDATE_EVERY:
+        if self.job.update_every < int(OVERRIDE_UPDATE_EVERY):
```
ktsaou: why only when it is smaller? what if it is bigger?

plugins.d/python.d.plugin
```diff
@@ -368,6 +370,10 @@ class Plugin(object):
             self.cleanup()
             self.autodetect_retry()
 
+            # FIXME: https://github.com/firehol/netdata/issues/3817
+            if self.runs_counter % GC_COLLECT_EVERY == 0:
+                gc.collect()
```
ktsaou: I suggest to add a debug here, so that we can see it running in debug mode.

plugins.d/python.d.plugin
```diff
@@ -16,6 +17,7 @@ from re import sub
 from sys import version_info, argv
 from time import sleep
 
+GC_COLLECT_EVERY = 300
```
ktsaou: I suggest to put an option in `python.d.conf` for this so that people can run it more or less frequently, and allow to be disabled.

plugins.d/python.d.plugin
l2isbad: Then it shouldn't be changed.

We have to override job update_every only if global bigger update_every bigger then job update interval.

ktsaou: > We have to override job update_every only if global update_every bigger then job update interval.

Maybe I don't understand what is `OVERRIDE_UPDATE_EVERY`. Can you give me some more info?

Generally, when people say they want a specific `update_every` at the job level, we should respect it, independently of how it compares with the global one. Right?

ktsaou:
let me know when it is ready for merge.

l2isbad:
ready

PR #4194, opened 2018-09-14
nginx_plus: use upstream server IP:port in dimension IDs, not the transient ID

illes:
nginx-plus assigns an internal numeric ID to each upstream server, this ID might change for the "same" upstream server (_same_ meaning identical IP:port) when, e.g.:
- nginx is restarted
- an upstream server is removed and then re-added to upstream via the API

Using the sanitized IP:port in dimension IDs ensures that charts behave as expected.

### DIFF
--- a/python.d/nginx_plus.chart.py
+++ b/python.d/nginx_plus.chart.py
@@ -178,7 +178,7 @@ def web_upstream_charts(wu):
         ]
     }
     for peer in wu:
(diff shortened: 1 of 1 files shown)


l2isbad:
@illes 
I see `name` and `servers` provides same info
```json
{ "server":"10.0.0.1:8080","name":"10.0.0.1:8080"}
```

After your PR
```
charts['web_upstream_{0}_{1}_responses'.format(wu.name, peer.server)]
```

Issue #4233, opened 2018-09-20
Apps plugin: wrong open_sockets counter when fd type changes


vdombrovski:
Bug has been tested on Linux, with netdata 1.9, 1.10 and current master (2087ce00).

Can be reproduced by monitoring the following process:

```c
// mutating_fd.c
#include <stdio.h>
#include <unistd.h>
#include <sys/socket.h>

int main(void)
{
   int fd, sockfd, IVAL = 1e6;

   for (int i=0; i<5; i++) {
      fd = fileno(tmpfile());
      sockfd = socket(AF_INET, SOCK_STREAM, 0);
      usleep(IVAL);
      dup2(fd, sockfd);
      usleep(IVAL);
      close(fd);
      usleep(IVAL);
   }
   return 0;
}
```

```sh
gcc -Wall mutating_fd.c -o mutating_fd
echo "mutating_fd: *mutating_fd*" >> /etc/netdata/apps_groups.conf
systemctl stop netdata
/usr/libexec/netdata/plugins.d/apps.plugin 1 | grep apps.sockets -A 2
```

**Expected behavior (stdout)**:
```
BEGIN apps.sockets 999659
SET mutating_fd = 1
[..]
BEGIN apps.sockets 999659
SET mutating_fd = 1
[..]
BEGIN apps.sockets 999659
SET mutating_fd = 1
```
> sometimes the value can also be 0

Expected behavior would be consistent with lsof output
```sh
lsof -c mutating | grep -c sock
1
```

**Actual behavior**:
```
BEGIN apps.sockets 999659
SET mutating_fd = 1
[..]
BEGIN apps.sockets 999659
SET mutating_fd = 2
[..]
BEGIN apps.sockets 999659
SET mutating_fd = 3
[..]
```

Problem comes from this toggle, which doesn't check that the registered fd is actually of the same type as the original one:

https://github.com/firehol/netdata/blob/2087ce00/src/apps_plugin.c#L1790

**The issue might also affect BSD systems**, but I can't confirm this.

ktsaou:
merged the fix.

PR #4234, opened 2018-09-20
WIP: Apps plugin: fix wrong open_sockets counter when fd type changes

vdombrovski:
This PR is a fix attempt for #4233. The fix is for Linux systems only, as I can't confirm that the problem affects BSD systems.

### DIFF
--- a/src/apps_plugin.c
+++ b/src/apps_plugin.c
@@ -342,6 +342,8 @@ struct pid_stat {
     // kernel_uint_t io_cancelled_write_bytes;
 
     int *fds;                       // array of fds it uses
+    char * *fd_paths;               // array of fd paths
+    int *fd_last_readlink;          // array of last fd readlink check timestamps
     int fds_size;                   // the size of the fds array
 
     int children_count;             // number of processes directly referencing this
@@ -691,7 +693,10 @@ static inline struct pid_stat *get_pid_entry(pid_t pid) {
 
     struct pid_stat *p = callocz(sizeof(struct pid_stat), 1);
     p->fds = callocz(sizeof(int), MAX_SPARE_FDS);
+    p->fd_paths = callocz(sizeof(char *), MAX_SPARE_FDS);
+    p->fd_last_readlink = callocz(sizeof(int), MAX_SPARE_FDS);
     p->fds_size = MAX_SPARE_FDS;
+
 
     if(likely(root_of_pids))
         root_of_pids->prev = p;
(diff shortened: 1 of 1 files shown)


src/apps_plugin.c
```diff
@@ -1760,20 +1760,20 @@ static inline int read_pid_file_descriptors(struct pid_stat *p, void *ptr) {
             p->fds_size = fdid + MAX_SPARE_FDS;
         }
 
-        if(unlikely(p->fds[fdid] == 0)) {
-            // we don't know this fd, get it
+        // get fd information
+        sprintf(fdname, "%s/proc/%d/fd/%s", netdata_configured_host_prefix, p->pid, de->d_name);
```
ktsaou: hm... this means on a system with many thousands of fds, we will do many thousands of sprintfs per second. Have you checked the increased CPU utilization due to this?

I see `readdir()` returns the inode number. We could probably cache that inode number and compare it at the next iteration.

If this works, it will save the `sprintf()` and the `readlink()`.

What do you think?

vdombrovski: I agree. In the current implementation there must be indeed some impact of calling `readlink()` on every single fd for each collection.

I'll try adding an array of inodes to each pid_stat then, much like we currently do with the `fdid->slot` assignation.

ktsaou: Can you please check using your test tool this will work?
It may not report inodes or something.

Also, it may be easier to convert the existing array of ints to an array of structures, with each structure having 2 ints (fd, inode). I can't tell right now which way is better.

vdombrovski: Actually, on second though, I'm having some doubts about this. It has been already specified here: https://github.com/firehol/netdata/blob/5886a64/src/apps_plugin.c#L1787

When a FD number is reused for another link, the inode number stays the same. So we'll always pass the inode check against the cache, even if the link type isn't the same.

ktsaou: hm... this means the best we can do is to cache the filenames, so that we will not do `sprintf` all the time. This will be tricky, since we will have to `strdupz()` and `freez()` these buffers.

Then, we may be able to save a bit on resources, if we don't `readlink()` every single second. So, we could have a timestamp too and execute `readlink()` every 2, 3, or 5 seconds, depending on a global option.

vdombrovski:
Tagged as WIP because all points discussed haven't yet been addressed (notably, `readlink_interval` is not yet configurable).

I ended up using multiple arrays to cache last `readlink` checks and fd names. It would indeed be better to use a `struct` instead, I might add it later.

Please don't hesitate to review the implementation, as I'm not really a seasoned C developer. (Checked functionality, memleaks and GCC warnings still).

ktsaou:
This pull request **fixes 1 alert** when merging c4e412583b62365bd48a63534713cca9247bcfc2 into 9ca9d491d67f1fecc8e8f455d4de8df2a73a371a - [view on LGTM.com](https://lgtm.com/projects/g/firehol/netdata/rev/pr-0b705d971778c8dee12b851081a435227e8e6750)

**fixed alerts:**

* 1 for FIXME comment

---

ktsaou:
ok, fixed this with #4243

ktsaou:
merged the fix.

PR #4242, opened 2018-09-21
Added -NoLog parameter for megacli calls

vobruba-martin:
Running `megacli` without the `-NoLog` parameter leads to generating a `MegaSAS.log` file in the current working directory (for me it is `/etc/netdata/MegaSAS.log`). The file is growing really fast because it is appended after every `megacli `call.


l2isbad:
thx @vobruba-martin 

Cc @petarkozic

ktsaou:
merged! thank you!

PR #4243, opened 2018-09-21
apps.plugin now checks fds for changes, with adaptive caching

### LINKED ISSUE #4233: Apps plugin: wrong open_sockets counter when fd type changes

ktsaou:
fixes #4233 
fixes #4234

apps.plugin monitors the file descriptors of processes and reports their type (file, socket, pipe, etc).
There was an issue: apps.plugin read the fds, but if a process closed and re-opened the same fd number, apps.plugin did not detect the change.

Unfortunately, scanning all the fds of all processes, every single second is resource intensive.
Actually, it doubles the CPU utilization of apps.plugin (which is already quite heavy).

To work around it, this PR does the following:

1. It caches the `/proc` filenames of all open fds. These never change. They are links like `/proc/self/fd/*` shows them. So, once a process opens an fd number, its filename will remain forever in apps plugin, until the process exits. This avoids an `snprintf()` per open file per iteration.

2. apps.plugin now caches links fds resolve to (the results of `readlink()`). But this caching involves some **trust**: as time passes and the fd is unchanged, apps.plugin skips iterations on it.

   It works like this: the first time apps.plugin sees a fd, it will check it at the next iteration too. But then, as iterations continue to find the fd unchanged, apps.plugin skips iterations (uses the cached result), up to 60 seconds.

   So, apps.plugin checks the fd in these iterations 0, 1, 3, 6, 10, 15, 21, 28, etc up to 60 and then every 60.

   This caching has some randomness in it, so that if you start apps.plugin now it won't process all the files together every 60 seconds. So, it tries to distribute them between 60 to 70 seconds each (this happens for the first 60 seconds too).

This is CPU utilization delta, before and after (something like +5 to 10%):

image

This is the caching in action (so when started it did a lot of refreshes, and as files were trusted, it did it less frequently):

image

And this is now changed links found on already open fds:

image

Of course, caching uses some memory (increased from 3.6 to 5 MB):

image

The plugin also detects if files have been changed, using the `/proc` inodes:

image

---

To control this caching, apps.plugin has a new command line option:

```
fds-cache-secs N
```

Where N is the max number of seconds for such a cache.
Caching can be disabled by setting `N = 0`. In this case, apps.plugin will read all the fd, of all processes, on every iteration.

You can configure this setting from `netdata.conf`:

```
[plugin:apps]
	# update every = 1
	command options =  fds-cache-secs 10
```

The above sets the max caching time to 10 seconds (+ randomness of 10 more).

For `fds-cache-secs < 10` the randomness doubles `fds-cache-secs`. So, if you set `fds-cache-secs 2`, the max cache time will be 4 seconds.

For `fds-cache-secs > 10` the randomness is fixed to `10` seconds.


ktsaou:
This pull request **fixes 1 alert** when merging 4178a7503d113e13715ed5d3b44e19c95163b496 into 9ca9d491d67f1fecc8e8f455d4de8df2a73a371a - [view on LGTM.com](https://lgtm.com/projects/g/firehol/netdata/rev/pr-15b5ee81937eab92858ee9e9db5a57d55302038f)

**fixed alerts:**

* 1 for FIXME comment

---

ktsaou:
This pull request **fixes 2 alerts** when merging 6ff3ffc21bc7f7064eae5fed609b1ebcc8003bf5 into 9ca9d491d67f1fecc8e8f455d4de8df2a73a371a - [view on LGTM.com](https://lgtm.com/projects/g/firehol/netdata/rev/pr-ee6f831a813fa45f3ede2e025b6948037b2a5340)

**fixed alerts:**

* 2 for FIXME comment

---

ktsaou:
This pull request **fixes 2 alerts** when merging dcd528b03c3c71107707a34b1490fd68ca0dcc5a into 9ca9d491d67f1fecc8e8f455d4de8df2a73a371a - [view on LGTM.com](https://lgtm.com/projects/g/firehol/netdata/rev/pr-e1edcbc8f0453005140e964c1798c3dbaf5a3491)

**fixed alerts:**

* 2 for FIXME comment

---

PR #4270, opened 2018-09-23
Fix several typos in documentation

Calinou:
There's probably a lot more to fix, but it's better than nothing :slightly_smiling_face:


ktsaou:
Thank you!
Yes, there are a lot more and I sure the wiki is full of typos too...

PR #4281, opened 2018-09-24
tcp syn and accept queue charts and alarms

### LINKED ISSUE #3234: Unexpected ipv4 tcp listen drops
Hi, 
I'm running `netdata` on CentOS 7.4 and almost every day I receive the following message via Telegram:
```
ipv4.tcplistenissues (tcp)
1m ipv4 tcp listen drops = 1 drops
the number of TCP listen socket drops during the last minute
```
I checked every log file on the server, but did not find anything that could be related to the issue. Could you please help me to detect what causes such messages?

In addition, here is `netstat` output:
```
# netstat -anp | grep LISTEN
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      840/sshd            
tcp        0      0 127.0.0.1:8125          0.0.0.0:*               LISTEN      31168/netdata       
tcp        0      0 127.0.0.1:19999         0.0.0.0:*               LISTEN      31168/netdata       
tcp        0      0 127.0.0.1:3306          0.0.0.0:*               LISTEN      1200/mysqld         
tcp6       0      0 :::80                   :::*                    LISTEN      837/httpd           
tcp6       0      0 :::22                   :::*                    LISTEN      840/sshd            
tcp6       0      0 :::443                  :::*                    LISTEN      837/httpd           
tcp6       0      0 ::1:8125                :::*                    LISTEN      31168/netdata       
unix  2      [ ACC ]     STREAM     LISTENING     10514    1/systemd            /run/lvm/lvmpolld.socket
unix  2      [ ACC ]     SEQPACKET  LISTENING     10524    1/systemd            /run/udev/control
unix  2      [ ACC ]     STREAM     LISTENING     16466    1200/mysqld          /var/lib/mysql/mysql.sock
unix  2      [ ACC ]     STREAM     LISTENING     12968    1/systemd            /var/run/dbus/system_bus_socket
unix  2      [ ACC ]     STREAM     LISTENING     7592     1/systemd            /run/systemd/journal/stdout
unix  2      [ ACC ]     STREAM     LISTENING     23225    2521/NetworkManager  /var/run/NetworkManager/private-dhcp
unix  2      [ ACC ]     STREAM     LISTENING     10458    1/systemd            /run/systemd/private
unix  2      [ ACC ]     STREAM     LISTENING     10483    1/systemd            /run/lvm/lvmetad.socket
```

ktsaou:
fixes #3234 

Thanks to @literal, this PR:

1. Adds a new chart `ipv4.tcp_syn_queue` that tracks `TcpExtTCPReqQFullDoCookies` and `TcpExtTCPReqQFullDrop`
2. The chart `ipv4.tcplistenissues` is now renamed to `ip.tcp_accept_queue` that tracks `TcpExtListenOverflows` and `TcpExtListenDrops`
3. The alarm `1m_ipv4_tcp_listen_drops` is now names `1m_tcp_accept_queue_drops` and triggers critical when there is >150min drops.
4. The alarm `1m_ipv4_tcp_listen_overflows` has been renamed to `1m_tcp_accept_queue_overflows`
5. New alarm `1m_tcp_syn_queue_drops` is tracking `TcpExtTCPReqQFullDrop`
6. New alarm `1m_tcp_syn_queue_cookies` is tracking `TcpExtTCPReqQFullDoCookies`
7. Added the explanation given by @literal here: https://github.com/netdata/netdata/issues/3234#issuecomment-423935842 to both the alarm configuration file and above the charts.
8. All charts of `/proc/net/netstat` have now their own category on the dashboard `Networking Stack`, since they track internal functions of the networking stack at the kernel, which are common to both IPv4 and IPv6.

The charts:

image

The alarms:

image


ktsaou:
@literal I am looking forward for your help, so feel free to comment.

You believe `TcpExtListenDrops` should be part of another chart or a chart of its own?

For example, we already have charts for:

1. `ipv4.tcphandshake` with `EstabResets`, `OutRsts`, `AttemptFails`, `SynRetrans`
2. `ipv4.tcperrors` with `InErrs`, `RetransSegs`

So, there 3 possibilities to fix this:

1. Leave it at `ipv4.tcp_accept_queue` and have a better description of it.
2. Move it to another existing chart that better suits it
3. Add a new chart for it

What do you think?

ktsaou:
There is also the possibility to revert the old name of this chart `ipv4.tcplistenissues`.
I thought that renaming to `ipv4.tcp_accept_queue` would allow people to better understand what is happening under the hood. But if this is misleading too, we should probably revert the previous name.

literal:
Thinking about it once again, I wonder whether `TcpExtListenDrops` is a very useful metric at all. It's incremented in a variety of different situations, some of which are harmless while others indicate trouble. And there are even cases when incoming TCP connection requests are discarded without `TcpExtListenDrops` being incremented, e.g. when they time out (after retransmitting the SYN/ACK reply a couple of times).
So perhaps it does more harm confusing users than being a useful piece of information?

Apart from that I agree with your idea to keep the metrics on a single chart, renaming them for more clarity, eg. `synOverflows`, `synCookies` and `acceptDrops`. The chart name `ipv4.tcplistenissues` might be a bit misleading though, because the numbers apply to IPv6 as well.

ktsaou:
So, @Ferroin you say we should keep the alarm about `TcpExtListenDrops`, but only trigger it above a threshold (eg above 60/min)?

@literal yes `ipv4` is wrong here. all `/proc/net/netstat` metrics are both IPv4 and IPv6?
If yes, we could provide another category on the dashboard called "IP Networking" with them, separate from IPv4 and IPv6.

Regarding the chart of `TcpExtListenDrops`, so you say to mix "SYN queue" and "Accept queue" to a single chart. Right?

literal:
@Ferroin

>That example _[SYN timeouts]_ is technically an error on the remote peer's end, not the lo9cal end, hence why it's not tracked by TcpExtListenDrops.

Could as well be caused by any router or firewall along the way or even a bad Netfilter rule on **your** server.

If the same thing happens (i.e. the remote side's ACKs don't reach you) but the party 'eating' the packet is so nice as to inform you about it via an ICMP packet, the discarded SYN **is** counted by `TcpExtListenDrops`. To me that seems inconsistent.

> If it's consistently high though, that's a pretty solid indication that something is wrong, even if the exact cause is 'harmless'.

Just one random scenario: A client has a wrong IP address for a service of yours. It happens to be your network's broadcast address. The client keeps trying to reach your service, but SYN packets to broadcast addresses are always discarded, incrementing `TcpExtListenDrops`.
You will see a high number of `TcpExtListenDrops` even though this time the problem really lies on the remote end. And you will have little chance to find the cause of the listen drops short of watching it in Wireshark while it'shappening.
(I know most routers block broadcast addresses, but not all of them. Also the client could be on the local network. And of course, a server application has to listen on the BC address for this, but most web servers, FTP servers etc. are configured to listen on all addresses by default.)

literal:
> @literal yes ipv4 is wrong here. all /proc/net/netstat metrics are both IPv4 and IPv6?

As far as I can tell, they are.

> If yes, we could provide another category on the dashboard called "IP Networking" with them, separate from IPv4 and IPv6.

That sounds like a good idea to me!

> Regarding the chart of TcpExtListenDrops, so you say to mix "SYN queue" and "Accept queue" to a single chart. Right?

Right. Just an idea, though. I'm not passionate about it. :-)

Ferroin:
> If the same thing happens (i.e. the remote side's ACKs don't reach you) but the party 'eating' the packet is so nice as to inform you about it via an ICMP packet, the discarded SYN is counted by TcpExtListenDrops. To me that seems inconsistent.

Wow, I actually didn't know about that one, so yeah, it is inconsistent.

> Just one random scenario: A client has a wrong IP address for a service of yours. It happens to be your network's broadcast address. The client keeps trying to reach your service, but SYN packets to broadcast addresses are always discarded, incrementing TcpExtListenDrops.
You will see a high number of TcpExtListenDrops even though this time the problem really lies on the remote end. And you will have little chance to find the cause of the listen drops short of watching it in Wireshark while it's happening.

It still tells you something is wrong here.  Yes, it may not be your system, but that does not change the fact that _something_ is wrong.

> (I know most routers block broadcast addresses, but not all of them. Also the client could be on the local network. And of course, a server application has to listen on the BC address for this, but most web servers, FTP servers etc. are configured to listen on all addresses by default.)

Listening on the wildcard address is not the same as listening for all inbound traffic that would be routed to the local system.  A wildcard listen address means you listen on all configured addresses for the local machine, but does not include broadcast or multicast addresses.  So, if you're actually seeing this behavior, there's a bug _somewhere_.

Ferroin:
@ktsaou Yes, although I'd go even higher than that.  Probably something more along the lines of 150/minute (roughly 2.5/second).  Essentially, if things are really bad and the more specific alarms are not triggering, we should still be reporting it, but otherwise we shouldn't risk annoying people.

literal:
@Ferroin 

> A wildcard listen address means you listen on all configured addresses for the local machine, but does not include broadcast or multicast addresses.

You're right. My example was rubbish. ;-)

I'm fine with keeping the `TcpExtListenDrops` metric and the corresponding alarm with a higher threshold. I guess it just boils down to a question of taste - whether you value it higher to avoid confusing users with a somewhat inconsistent catch-all metric or to keep something that might someday be useful for someone.

ktsaou:
Is `/proc/net/sockstat` also both IPv4 and IPv6 ?

ktsaou:
> Is `/proc/net/sockstat` also both IPv4 and IPv6 ?

hm... no it is not. I see `/proc/net/sockstat` and `/proc/net/sockstat6`.

Ferroin:
> hm... no it is not. I see /proc/net/sockstat and /proc/net/sockstat6.

Yeah, there's a somewhat surprising degree of inconsistency in `/proc/net`...

Ferroin:
Overall LGTM right now, but I agree that some of the descriptions around this stuff could use some work.

spinitron:
I silenced 1m_ipv4_tcp_listen_drops because I couldn't understand the root cause or if it is something I need to attend to.

So +1 for more description of what the user might _do_ with the information..

ktsaou:
All charts of `/proc/net/netstat` now appear at their own section "Networking Stack".

image

ktsaou:
> I silenced 1m_ipv4_tcp_listen_drops because I couldn't understand the root cause or if it is something I need to attend to.

Most probably this PR will silence it too. With this PR it triggers only when there are `> 150 packets/min` dropped. For the reasons we could detect, we added separate alarms.

ktsaou:
So, unless you can any comments or suggestions, this is ready.

spinitron:
Afaict, this is a good improvement. I agree: ready.

ktsaou:
@vlvkobal we have renamed a number of charts, from `ipv4` and `ip`, because under linux these chart reflect internal counters of the kernel that apply to both ipv4 and ipv6. These are all the charts appearing here: https://github.com/netdata/netdata/blob/master/src/proc_net_netstat.c

A few of these charts are common under FreeBSD and MacOS. I can rename them the FreeBSD/MacOS ones, but the question is: under FreeBSD/MacOS these metrics are IPv4 only or both IPv4 and IPv6?

vlvkobal:
They are IPv4 only.

ktsaou:
merged!

PR #4289, opened 2018-09-25
Python.d PEP 8 cleanup, modules M

Ferroin:
PEP 8 compliance cleanups for python.d modules with names starting with M.

Relevant: #4167


Ferroin:
Just realized that I had missed the mysql module in this one.  Changes for that have been added.  

@l2isbad, @paulfantom If the two of you could recheck this it would be appreciated.

PR #4348, opened 2018-10-02
Fix phpfpm ipv6 configuration

candrews:
IPv6 literals in URLs must be enclosed in square brackets

Without this change, this error results when running `/usr/libexec/netdata/plugins.d/python.d.plugin 1 debug trace phpfpm`:
```
  File "/usr/libexec/netdata/python.d/python_modules/bases/FrameworkServices/UrlService.py", line 96, in _get_raw_data
    status, data = self._get_raw_data_with_status(url, manager)
  File "/usr/libexec/netdata/python.d/python_modules/bases/FrameworkServices/UrlService.py", line 119, in _get_raw_data_with_status
    redirect=redirect)
  File "/usr/lib/python3.7/site-packages/urllib3/request.py", line 68, in request
    **urlopen_kw)
  File "/usr/lib/python3.7/site-packages/urllib3/request.py", line 89, in request_encode_url
    return self.urlopen(method, url, **extra_kw)
  File "/usr/lib/python3.7/site-packages/urllib3/poolmanager.py", line 310, in urlopen
    u = parse_url(url)
  File "/usr/lib/python3.7/site-packages/urllib3/util/url.py", line 199, in parse_url
    raise LocationParseError(url)
urllib3.exceptions.LocationParseError: Failed to parse: ::1
```

PR #4349, opened 2018-10-02
Fix apache ipv6 configuration

candrews:
IPv6 literals in URLs must be enclosed in square brackets


candrews:
phpfpm has the same problem, see #4348

ktsaou:
thank you!

PR #4373, opened 2018-10-10
Fix unevaluated $used_ram_to_ignore on FreeBSD

openspork:
FreeBSD not taking ARC size into consideration when calculating free RAM.  Copying definition of $used_ram_to_ignore from Linux section resolves.

### DIFF
--- a/collectors/node.d.plugin/node.d.plugin.in
+++ b/collectors/node.d.plugin/node.d.plugin.in
@@ -40,7 +40,7 @@ var util = require('util');
 var http = require('http');
 var path = require('path');
 var extend = require('extend');
-var netdata = require('../../../netdata');
+var netdata = require('netdata');
 
 
 // --------------------------------------------------------------------------------------------------------------------


ktsaou:
You can just append `freebsd` to the linux one at the `os` line.
I think however, this was removed for a reason. Let me search for related issues.

ktsaou:
hm... can't find anything related. So, we can enable this on freebsd.
Ping me when you are ready.

openspork:
> hm... can't find anything related. So, we can enable this on freebsd.
> Ping me when you are ready.

I'm ready -- but don't know what I am ready for.  

Sorry!  This is my first pull request.

ktsaou:
@openspork instead of copying the alarm and adding it a second time, you can just append `freebsd` to the Linux version of the alarm. So it will look like that:

```
 os: linux freebsd
```

Adding the same alarm a second time, would be useful only when the freebsd version of the alarm would differ. But I see they are exactly the same. So we can just apply the linux version to freebsd too.

openspork:
> @openspork instead of copying the alarm and adding it a second time, you can just append `freebsd` to the Linux version of the alarm. So it will look like that:
> 
> ```
>  os: linux freebsd
> ```
> Adding the same alarm a second time, would be useful only when the freebsd version of the alarm would differ. But I see they are exactly the same. So we can just apply the linux version to freebsd too.

@ktsaou Done -- now a single declaration for the used_ram_to_ignore alarm

ktsaou:
yes, thank you.
But now the files in the repo have been moved.... sorry.

Now that file is under `health/health.d/`.
Could you please rebase this PR to merge it?

PR #4391, opened 2018-10-13
modularized all source code

ktsaou:
This PR changes the structure of source tree for almost every piece of code and config file.

This was required to allow more people step into netdata source tree, understand how it works, allow netdata to integrate sub-gits into it, etc.

The new structure follows the key functions of netdata:

- collectors
- backends
- daemon
- database
- health
- libnetdata
- registry
- streaming
- web

Every file of netdata is now under these directories.


ktsaou:
The new directory structure can be browsed here: https://github.com/ktsaou/netdata/tree/plugins3

ktsaou:
> `src/` should remain as a directory for C code only

Why only `C` code should be in `src/`?

The key point here is that every plugin and every module should be in its own directory.

I prefer to have them under `src` in a well organized directory structure so that people can also have visibility of how this works:

```
src/
+ plugins/
    + plugins.d.plugin/      <<< this internal plugin provides the external plugins API
        + python.d.plugin/  <<< this one provides python data collection modules
            + nginx
            + web_log
```

So, you don't like non-C code to be in `src/`? Where should it be and why?

paulfantom:
If you put everything in `src` then what is the point of other directories in top level? Top level directories are the first thing you see when going to github repository, they should reflect modularity of the source code. Putting all code into `src` is just adding one more level of indirection and looks more like sweeping mess under the rug.

Another aspect is the fact that core part written in C is the only one which needs compilation. This also can be reflected by a directory structure. I don't think there is a need to compile whole netdata when developing only apps_plugin. This increase in developer experience could be achieved partly by changing build system and partly by better directory structure.

In my opinion it would be more beneficial if code would be structured in something similar to:
```
web/ (if posssible)
core/
plugins/
  python.plugin/
    modules/
      nginx.py
      web_log.py
  node.plugin/
    modules/
...
```

This way you have clear distinction what is what and where it resides. Also I don't see any need to have directories like `nginx` for modules. As it is now, modules are small single files and they should stay that way. Adding another directory can potentially lead to a change in that situation.

It would be possibly even better if `plugins` directory was structured in a way which shows what those plugins do. Like `input`, `output`, `mangle`, `alert` etc. So maybe even structure like this would be better:
```
web/ (if posssible)
core/
plugins/
  input/
    python.plugin/
      modules/
        nginx.py
        web_log.py
    node.plugin/
      modules/
...
```

TL;DR Group directories first by function then by code type to provide better DevEX and modularity.

----

As for other things:
> info about writing python modules

This should be in `CONTRIBUTING.md` as well as in documentation. If there is a need for it in code, then it should be in `docs/` top level directory

> every file related to nginx, including a README.md

Almost no one reads READMEs which are far down in directory layout. This should be in documentation.

----

Also look at this part:
```
+ plugins/
    + plugins.d.plugin/      <<< this internal plugin provides the external plugins API
```

This is pointless, you put `plugins...` under `plugins`.

paulfantom:
Also regarding build system, good directory layout can lead to less code generation. For example Makefiles for python plugin doesn't need to be automatically created by autoconf/cmake and they can consist of developer tooling. This means that if someone is developing only parts of python code, then he/she doesn't need things like gcc, cmake/autoconf etc. as only thing needed is to have standard tools needed for particular language (python in this case). Not even core part is needed and if it is then can be downloaded from nightly build or run with docker container.

ktsaou:
Then I think the best way is to move `src/*.c` and `src/*.h` to a directory called `daemon` and move everything off `src/`. That is, remove the `src/` directory completely.

ktsaou:
Navigate a bit this: https://github.com/ktsaou/netdata/tree/plugins3/src/plugins
I really like how it is now.

ktsaou:
So, what do you think? Shall I try to remove `src/` completely?
I think this would provide the best visibility on what netdata does.

The list will be like that:

image

With a few additions for installer, contrib, makeself, etc.

ktsaou:
I really like removing it... It will be a great improvement...

ktsaou:
Added `[RFC]` at the title, to let everyone know I am open for ideas about this.

paulfantom:
> Shall I try to remove src/ completely?

I agree. Structure from attached picture looks much cleaner and "friendlier" :smile: Soon I will also take care of top level directories like `packaging`, `makeself` and `contrib` as they have very similar purpose and at least part of their content should be grouped together.

What is `libnetdata`?

cakrit:
I would prefer top level directories to be the pretty standard lib, src, bin, build, conf etc. even if some of them are not necessarily tracked in GitHub. You lose very little if you just add one more level for src and remain future-proof

There won't be a single contributing.md. I don't need the doc for a python plugin to be in the same place as the info on how to extend the core.

Directories for what is now single .py files is necessary and future proof. We will have a separate readme, a separate conf and an image, just to name the ones I am aware of right now. You can't lose by adding structure, you can only regret it later if you don't. It's much easier to collect things in one place if you already have them split, very difficult to do the opposite.

paulfantom:
> I would prefer top level directories to be the pretty standard lib, src, bin, build, conf etc. even if some of them are not necessarily tracked in GitHub. You lose very little if you just add one more level for src and remain future-proof

This separation on `lib`, `src` etc. is good for small, single-language projects, and netdata isn't one.

> You lose very little if you just add one more level for src and remain future-proof

You loose a lot in terms of marketing, community engagement and developer experience. I don't see anyhow this contributes in any way to being future-proof by having all code in `src`. I see it more as non-future-proof and as for "why" - I wrote it in previous comments.

>  It's much easier to collect things in one place if you already have them split, very difficult to do the opposite.

We don't want to have everything in one place! Having all code in one place shows that our app has monolithic architecture which is not the case. Also having everything in one repo causes problems with permissions and community engagement, that's one of the reasons we are under `netdata` organization.

> There won't be a single contributing.md

That's against good practices of creating github repositories and open source projects in general. `CONTRIBUTING.md` should exist as one simple file as the intention of that file is to engage new contributors into taking action and contributing to project by following simple guidelines. This cannot be some distributed multi-page document.

ktsaou:
> What is `libnetdata`

A library of code that all netdata C binaries use. Thinks like parsing files, evaluating expressions, hashing trees, O/S abstraction (threads, locks, pipes), etc.

paulfantom:
I think `plugins` is enough.

ktsaou:
> I think `plugins` is enough.

You mean to rename `plugins.d.plugin` to `plugins`?

It would be `plugins/plugins/python.d.plugin`...

ktsaou:
I could just rename `plugins.d.plugin` to `plugins.d`, so that it would be:

`plugins/plugins.d/python.d.plugin`

ktsaou:
Or I could renamed it to `external`:

`plugins/external/python.d.plugin`

l2isbad:
`external` is nice :+1:

paulfantom:
Why do you need another stage? Why not just `plugins/python.d.plugin`?

cakrit:
> Why do you need another stage? Why not just plugins/python.d.plugin?
I needs to be as shown in the infographic.:
plugins/internal
plugins/external
Question: can we/should we include submodules under plugins/thirdparty as well for these https://netdata-doc.readthedocs.io/en/latest/Third-Party-Plugins/ ?

--

> We don't want to have everything in one place! 
It's what I'm saying, which is why I disagree with the following two:
> I don't see any reason to create more directories to host plugins code.
> Also I don't see any need to have directories like nginx for modules. As it is now, modules are small single files and they should stay that way. Adding another directory can potentially lead to a change in that situation.

--
The discussion on READMEs and CONTRIBUTING.md will continue in #4196

ktsaou:
Finished I think.

You can browse the new directory structure here: https://github.com/ktsaou/netdata/tree/plugins3

ktsaou:
I will continue copying wiki pages to `README.md` of the repo.
I hope you like it! I do!

ktsaou:
Regarding external plugins, well... there is a directory called `plugins.d` in `/usr/libexec/netdata`, so I thought I should name the plugin exactly that.

I don't think we can eliminate it, since there is really an internal plugin called `plugins.d` that implements the external plugins API.

Anyway, I don't have a hard opinion on that. I just find a bit more accurate.

ktsaou:
Guys if you ok this this, approve it please.
We can move more wiki pages here, later...

paulfantom:
@cakrit @ktsaou Need a clear description what is `internal` and what is `external` plugin and why this classification is better than `input`, `output`, etc.

Isn't `daemon` tightly integrated with `database`? If not what is the relation between `backend` and `database`?

This is mouthful: `plugins/plugins.d/python.d.plugin/...`, 3x `plugin` in subsequent directories. I don't like it, but it is not me who maintains those so I don't have strong opinions on this.


Side note: Merging this will break most of current PRs.

cakrit:
> Need a clear description what is internal and what is external plugin and why this classification is better than input, output, etc.

The infographic says that internal plugins run as threads inside NetData, whereas external plugins use plugins.d to get metrics in text format over pipes. Pretty big difference, as I understand it

ktsaou:
> @cakrit @ktsaou Need a clear description what is `internal` and what is `external` plugin and why this classification is better than `input`, `output`, etc.

All plugins in the `plugins/` directory of netdata are `input`:

- `internal` are the ones inside netdata - you never see them on the filesystem or the process tree.
- `external` are the ones you can see inside at the process tree and the `plugins.d` directory ending with `.plugin`.

What other agents (eg `telegraf`) calls `output` is probably closer to what netdata calls `backend` - a better name, to really reflect what netdata does, would be `archiving` or `long term archiving`. You see, we don't need an `output`. We can perfectly work without one.

You can find the backend plugins at the `backends/` directory.

> This is mouthful: `plugins/plugins.d/python.d.plugin/...`, 3x `plugin` in subsequent directories. I don't like it, but it is not me who maintains those so I don't have strong opinions on this.

Well, this is how it really is: `plugins.d.plugin` is an internal data collection plugin. But it does not collect any metrics. Instead it implements the `external` plugins API and forks the external plugins for collecting the metrics.

Actually, for `nginx` module of `python.d.plugin` to work, this is how it gets into netdata:

`nginx` module --(thread of)--> python.d.plugin --(pipe)--> plugins.d.plugin --(thread of)--> netdata

I don't like the long names either, but this is how it works...

> Isn't `daemon` tightly integrated with `database`? If not what is the relation between `backend` and `database`?

One netdata daemon, multiple databases. We may also have different formats of databases. We could move the `database` directory inside the `daemon`, but I think it is pretty important and deserves its own section. In a few months, the `database` directory will have 2 subdirectories: the existing database format, and the new one we will develop.

> Side note: Merging this will break most of current PRs.

I haven't touched much of the code, so a simple `git rebase` should not produce any conflicts.

ktsaou:
btw, once @Ferroin completes #3960 we may split `alarm-notify.sh` into multiple parts, thus creating a third type of plugins: alarm notification plugins, inside the `health/` directory.

So, finally we will have:

- `plugins`  = data collection plugins
- `backends` = metrics archiving plugins
- `health` = alarm notifications plugins
- `database` = database format plugins

ktsaou:
Btw, I would never call `output` the backends. Calling them `output` promotes the use of netdata as a data collection agent. Of course netdata is such an agent, but the backends functionality is primarily needed today because our database is not good enough. Once we have the database we need, backends will be obsolete (we will not remove them, since it allows our users to utilize their existing investments in time-series databases, but their use will be significantly less needed).

paulfantom:
I hope you meant `plugins/health/` not just `health/`.

ktsaou:
> I hope you meant `plugins/health/` not just `health/`.

Well, I don't know. `plugins/health` or `health/plugins` ?

If you project this to the future, we may also have web API plugins, data query plugins, statistical functions plugins, alarm evaluation functions plugins, web dashboard plugins (with js, css, html each), etc.

I am also flirting with idea of linking netdata with libpython, liblua, libnode, etc to allow us hook custom scripted code blocks all over netdata. Where would these plugins go?

So what would be it be better? To have a huge plugins directory with all the functions of netdata in it (this would like like renaming `src/` to `plugins/`), or a more meaningful top level directory with the plugins of each category together with the code and the functionality of that category?

I think `health/plugins` would be appropriate in the long run.
Of course, then `/plugins` is kind of wrong. It should be `/data-collection` or something similar.

paulfantom:
> Well, I don't know. plugins/health or health/plugins ?

If `health` is a plugin and you have a top level directory called `plugins` then for me it is obvious where it should go. That is also connected to why I don't like `plugins/plugins.d` as `plugins.d` doesn't say anything new.

Personally I don't care about internal or external plugins and I would prefer to group them by function which they are providing. Like `health`, `input/collector/whatever`, `backend`, etc. This way you end up with meaningful naming scheme which is future-proof.

One more thing is to clearly say what is actually a plugin as even we are saying `plugin` on a specific collector (like nginx collector) and on hmmm... "plugin" (?) like `python.d.plugin`. This is confusing and I fear that this confusion will be only greater with naming scheme implementing `plugins/plugins.d/`.

paulfantom:
Basically everything which is non-essential is not a plugin.

ktsaou:
> You first have end users and then some of them are converted to contributors.

ok, let's move them to the same directory... It will complicate things for contributors, but having a clear view for users is also important.

l2isbad:
> This is already decided: nginx is a module of python.d.plugin

I  am personally OK with modules. The problem is nobody calls them so. It always was and will be the source of the confusion
like
https://github.com/netdata/netdata/pull/3448#issuecomment-367291283

Ferroin:
Overall, I think the intent behind this is great.  Aside from the already stated benefits (which I like), it also makes it a bit easier to track who's been working on what part of the code from both the CLI and from GitHub, which is a good thing.

Beyond that, I'm kind of indifferent though.  As long as the paths don't get nested too deeply (and they aren't with the current state of the PR), I'm really not that picky about how the source is structured.

ktsaou:
Done.
Please approve it to have it merged!

You can see the current tree here: https://github.com/ktsaou/netdata/tree/plugins3

CMakeLists.txt
```diff
@@ -340,7 +340,7 @@ set(WEB_PLUGIN_FILES
         web/server/web_client.h
         web/server/web_server.c
         web/server/web_server.h
-        )
+        web/server/single/single-threaded.c web/server/single/single-threaded.h web/server/multi/multi-threaded.c web/server/multi/multi-threaded.h web/server/static/static-threaded.c web/server/static/static-threaded.h web/server/web_client_cache.c web/server/web_client_cache.h)
```
paulfantom: why not multiline?

ktsaou: clion did automatically.
I'll fix it on the next PR.

PR #4404, opened 2018-10-16
Added uwsgi plugin

robbert-ef:


l2isbad:
Сurrent version doesn't support adding workers in runtime, do we need this?

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,138 @@
...
+retries = 60
+
+ORDER = ['requests', 'tx', 'avg_rt', 'memory_rss', 'memory_vsz', 'exceptions', 'harakiri', 'respawn', ]
+
+# NOTE: lines are created dynamically in `check()` method
+CHARTS = {
+    'requests': {
+        'options': [None, 'Requests', 'requests/s', 'requests', 'uwsgi.requests', 'stacked'],
+        'lines': [
+            ['requests.%s', 'worker %s', 'incremental']
```
l2isbad: please use underscore instead of dot in dimension_id

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,138 @@
...
+        ]},
```
Ferroin: Please update these so that the formatting matches the other python.d modules.

In short, each chart should look like this:
```
'requests': {
    'options': [None, 'Requests', 'requests/s', 'requests', 'uwsgi.requests', 'stacked'],
    'lines': [
        ['requests.%s', 'worker %s', 'incremental']
    ]
},
```
With the closing braces for the 'lines' list and the chart itself each on their own line indented appropriately.

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,138 @@
...
+}
+
+
+class Service(SocketService):
+    def __init__(self, configuration=None, name=None):
+        SocketService.__init__(self, configuration=configuration, name=name)
+        self.url = self.configuration.get('host', 'localhost')
+        self.port = self.configuration.get('port', 1717)
+        self.order = deepcopy(ORDER)
```
l2isbad: copy is not needed for ORDER

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def read_data(self):
+        """
+        Read data from socket and parse as JSON.
+        :return: (dict) stats
+        """
+        raw_data = self._get_raw_data()
+        if not raw_data:
+            return None
+        return json.loads(raw_data)
```
l2isbad: Are you sure we don't need `try/except` here?

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,138 @@
...
+        'options': [None, 'Respawns', 'respawns', 'respawns', 'uwsgi.respawns', 'line'],
+        'lines': [
+            ['respawn_count.%s', 'worker %s', 'incremental']
+        ]},
+}
+
+
+class Service(SocketService):
+    def __init__(self, configuration=None, name=None):
+        SocketService.__init__(self, configuration=configuration, name=name)
```
Ferroin: Please change this to:
```
super().__init__(configuration=configuration, name=name)
```

It does the same thing in a more future-proof manner and is more consistent with other modules.

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,138 @@
...
+            for worker in stats['workers']:
+                for line in CHARTS[chart]['lines']:
+                    # dimension name and id
+                    dimension_name = line[0] % worker['id']
+                    dimension_id = line[1] % worker['id']
+                    self.definitions[chart]['lines'].append([dimension_name, dimension_id] + line[2:])
+
+        return True
+
+    def _check_raw_data(self, data):
```
Ferroin: This should probably be a static function.

Ferroin:
Aside from the largely formatting related comments I've made, this looks good to me.  I can't test it myself (I don't use uwsgi), but based on a quick look through the uwsgi documentation, it looks to me like it works as advertised.

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
Ferroin: It looks like Codacy thinks so too.  If you just update this line to:

```
@staticmethod
def _check_raw_data(data):
```

That should get rid of the warning from Codacy.

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,138 @@
...
+        ]},
+    'exceptions': {
+        'options': [None, 'Exceptions', 'exceptions', 'exceptions', 'uwsgi.exceptions', 'line'],
+        'lines': [
+            ['exceptions.%s', 'worker %s', 'incremental']
+        ]},
+    'harakiri': {
+        'options': [None, 'Harakiris', 'harakiris', 'harakiris', 'uwsgi.harakiris', 'line'],
+        'lines': [
+            ['harakiri_count.%s', 'worker %s', 'incremental']
```
l2isbad: I am not sure about `worker` prefix.  Do we really need it? (for every chart for evey dim)

l2isbad:
Oh, @robbert-ef thx for contributing first of all :+1:

ktsaou:
Please rebase this.

ktsaou:
The structure of the python plugins has changed a bit to make it easier adding new modules.
Please follow the new structure.

robbert-ef:
> Сurrent version doesn't support adding new workers in runtime, do we need this?

Is this possible using python.d.plugin?

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
robbert-ef: I've added this prefix so it more clear what the chart contains, without it, dimensions would just be `1`, `2`:

image

I'd rather have the `pid` there, but that can change over time in case of respawn and I dont know how to change dimensions run-time.

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
robbert-ef: done

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,138 @@
...
+import json
+from copy import deepcopy
+from bases.FrameworkServices.SocketService import SocketService
+
+# default module values (can be overridden per job in `config`)
+# update_every = 2
+priority = 60000
+retries = 60
+
+ORDER = ['requests', 'tx', 'avg_rt', 'memory_rss', 'memory_vsz', 'exceptions', 'harakiri', 'respawn', ]
```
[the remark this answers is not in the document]
robbert-ef: removed

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
robbert-ef: done

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
robbert-ef: done

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
robbert-ef: done

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def check(self):
+        """
+        Parse configuration, check if redis is available, and dynamically create chart lines data
+        :return: boolean
+        """
+        self._parse_config()
+
+        # Clear dynamic dimensions, these are added during `_get_data()` to allow adding workers at run-time
+        for chart in ('requests', 'tx', 'avg_rt', 'memory_rss', 'memory_vsz'):
```
l2isbad: make a global var for `('requests', 'tx', 'avg_rt', 'memory_rss', 'memory_vsz')`

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,169 @@
...
+            result['requests_%s' % key] = worker['requests']
+            result['tx_%s' % key] = worker['tx']
+            result['avg_rt_%s' % key] = worker['avg_rt']
+
+            # avg_rt is not reset by uwsgi, so reset here
+            if self.last_result.get('requests_%s' % key) == worker['requests']:
+                result['avg_rt_%s' % key] = 0
+
+            result['memory_rss_%s' % key] = worker['rss']
+            result['memory_vsz_%s' % key] = worker['vsz']
```
l2isbad: please use `format` instead of `%`

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,169 @@
...
+import json
+from copy import deepcopy
+from bases.FrameworkServices.SocketService import SocketService
+
+# default module values (can be overridden per job in `config`)
+# update_every = 2
+priority = 60000
+retries = 60
+
+ORDER = ['requests', 'tx', 'avg_rt', 'memory_rss', 'memory_vsz', 'exceptions', 'harakiri', 'respawn']
```
l2isbad: make it change format to
```python
ORDER = [
	'requests',
	'tx',
	'avg_rt',
	'memory_rss',
	'memory_vsz',
	'exceptions',
	'harakiri',
	'respawn',
]
```

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
```diff
@@ -0,0 +1,169 @@
...
+        Parse configuration, check if redis is available, and dynamically create chart lines data
+        :return: boolean
+        """
+        self._parse_config()
+
+        # Clear dynamic dimensions, these are added during `_get_data()` to allow adding workers at run-time
+        for chart in ('requests', 'tx', 'avg_rt', 'memory_rss', 'memory_vsz'):
+            self.definitions[chart]['lines'] = []
+
+        stats = self.read_data()
```
l2isbad: `stats = self.read_data()`
should be called right after `self._parse_config`

collectors/python.d.plugin/uwsgi/uwsgi.chart.py
robbert-ef: personally I think
```result['requests_%s' % key] = worker['rss']```
is more readable than
```result['requests_{}'.format(key)] = worker['rss']```
but will change if that is what you prefer

Comment (l2isbad) [reply collectors/python.d.plugin/uwsgi/uwsgi.chart.py]: https://docs.python.org/release/3.1.5/library/stdtypes.html#old-string-formatting-operations

File: collectors/python.d.plugin/uwsgi/uwsgi.chart.py (hunk 12 above)
Comment (l2isbad) [reply collectors/python.d.plugin/uwsgi/uwsgi.chart.py]: i mean `self.read_data()` return value check

File: collectors/python.d.plugin/uwsgi/uwsgi.chart.py (hunk 10 above)
Comment (l2isbad) [reply collectors/python.d.plugin/uwsgi/uwsgi.chart.py]: or you can use some helper func to make it less verbose

like

```python
def fmt(f, s):
    return '{0}_{1}'.format(f, s)

result['requests_{}'.format(key)] = worker['rss']
#  vs
result[fmt('requests', key)] = worker['rss']
```

robbert-ef: > https://docs.python.org/release/3.1.5/library/stdtypes.html#old-string-formatting-operations
This dates from 2012.
`printf`-style String Formatting was never deprecated and will not be removed afaik.

https://docs.python.org/3/library/stdtypes.html#old-string-formatting
https://stackoverflow.com/questions/13451989/pythons-many-ways-of-string-formatting-are-the-older-ones-going-to-be-deprec

l2isbad:
gj @robbert-ef

ktsaou:
thank you!

PR #4419, opened 2018-10-17
Evaluate $used_ram_to_ignore on FreeBSD

openspork:
Fix un-evaluated $used_ram_to_ignore variable on FreeBSD.


ktsaou:
merged! thank you!

PR #4430, opened 2018-10-18
Feat: detect NIC speed and alarm on each device for net traffic overflow

Ehekatl:
##### Summary

- add a local variable in net.net chart for nic speed
- add alarm template for net traffic overflow on each device

##### Component Name

proc_net_dev

##### Additional Information

screen shot 2018-10-18 at 2 09 48 pm

### DIFF
--- a/collectors/proc.plugin/proc_net_dev.c
+++ b/collectors/proc.plugin/proc_net_dev.c
@@ -436,6 +436,7 @@ int do_proc_net_dev(int update_every, usec_t dt) {
     static int enable_new_interfaces = -1;
     static int do_bandwidth = -1, do_packets = -1, do_errors = -1, do_drops = -1, do_fifo = -1, do_compressed = -1, do_events = -1;
     static char *path_to_sys_devices_virtual_net = NULL;
+    static char *path_to_sys_net_speed = NULL;
 
     if(unlikely(enable_new_interfaces == -1)) {
         char filename[FILENAME_MAX + 1];
@@ -505,15 +506,11 @@ int do_proc_net_dev(int update_every, usec_t dt) {
                 d->virtual = 0;
 
             // set nic speed if present
-            procfile *df = NULL;
-            snprintfz(buffer, FILENAME_MAX, "/sys/class/net/%s/speed", d->name);
-            if(likely(!d->virtual)) df = procfile_open(buffer, " \t,:|", PROCFILE_FLAG_DEFAULT);
-            if(likely(df)) {
-              df = procfile_readall(df);
-              if(likely(df)) {
-                d->speed_max = str2kernel_uint_t(procfile_word(df, 0));
-              }
-              procfile_close(df);
+            if(likely(!d->virtual)) {
+              snprintfz(buffer, FILENAME_MAX, "%s/sys/class/net/%s/speed", netdata_configured_host_prefix, d->name);
+              path_to_sys_net_speed = config_get(CONFIG_SECTION_PLUGIN_PROC_NETDEV, "path to get net device speed", buffer);
+              int ret = read_single_number_file(path_to_sys_net_speed, (unsigned long long*)&d->speed_max);
+              if(ret) error("Cannot read '%s'.", path_to_sys_net_speed);
             }
 
             snprintfz(buffer, FILENAME_MAX, "plugin:proc:/proc/net/dev:%s", d->name);
--- a/collectors/proc.plugin/proc_net_dev.c
+++ b/collectors/proc.plugin/proc_net_dev.c
@@ -66,6 +66,7 @@ static struct netdev {
     kernel_uint_t tcollisions;
     kernel_uint_t tcarrier;
     kernel_uint_t tcompressed;
+    kernel_uint_t speed_max;
 
     // charts
     RRDSET *st_bandwidth;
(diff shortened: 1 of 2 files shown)


collectors/proc.plugin/proc_net_dev.c
```diff
@@ -503,6 +504,18 @@ int do_proc_net_dev(int update_every, usec_t dt) {
...
+            snprintfz(buffer, FILENAME_MAX, "/sys/class/net/%s/speed", d->name);
+            if(likely(!d->virtual)) df = procfile_open(buffer, " \t,:|", PROCFILE_FLAG_DEFAULT);
+            if(likely(df)) {
+              df = procfile_readall(df);
+              if(likely(df)) {
+                d->speed_max = str2kernel_uint_t(procfile_word(df, 0));
+              }
+              procfile_close(df);
+            }
+
```
ktsaou: 3 issues with this code block:

1. We should respect the host prefix of netdata, so that when netdata runs in a container it will be able to find the `/proc` and `/sys` files of the host. This is done like this: https://github.com/netdata/netdata/blob/15408ace0c2a81986026dd66446bb216c9b96eb6/collectors/proc.plugin/proc_net_dev.c#L442-L443

2. All files should pass through `nedata.conf` so that users will be able to override the whole path if they need to. The example above does that.

3. Using `procfile` for that little thing is overkill. You can use this: https://github.com/netdata/netdata/blob/15408ace0c2a81986026dd66446bb216c9b96eb6/collectors/cgroups.plugin/cgroup-network.c#L63-L64

Ferroin:
This needs to filter out virtual devices.  Otherwise it will throw false positives on a regular basis for a wide variety of VPN links.  Most virtual devices default to the various physical-layer related attributes in `/sys/class/net` displaying values equivalent to 10Base-T, which means they report a speed of 10Mb/s regardless of what they can actually sustain.

health/health.d/net.conf
```diff
@@ -1,6 +1,24 @@
...
 
+# -----------------------------------------------------------------------------
+# net traffic overflow
+
+ template: 10s_net_traffic_overflow
+       on: net.net
+       os: linux
+    hosts: *
+ families: *
+   lookup: average -10s unaligned absolute of sent,received
```
ktsaou: I think we need 2 alarms, one for `received` and another for `sent`.
The way this is now, `$this` will be average of the **sum** of `sent` and `average`.

health/health.d/net.conf
```diff
@@ -1,6 +1,24 @@
...
+     calc: $this * 100 / ($nic_speed_max * 1000)
+    units: %
+    every: 10s
+     warn: $this > (($status >= $WARNING)  ? (50) : (70))
+     crit: $this > (($status == $CRITICAL) ? (70) : (90))
```
ktsaou: These conditions combined with the `10s` duration of the `lookup` lines, will make the alarm quite chatty (i.e. it will send alarms frequently).

I suggest to change the `lookup` line to `-1m` and set warning `> 90` and critical to `>95`.

If you do this, also change the title to `1m_XXX`.

Ehekatl: updated duration to 1m, but I made the threshold a little bit lower, since it's kind of rare(and hard) to keep a constant network load at near full speed for the whole minute.

collectors/proc.plugin/proc_net_dev.c
Ehekatl: updated, much cleaner now

Ehekatl:
@Ferroin I did filter with virtual interface, it will create a alarm for each interface though, that should be ok?

Ferroin:
Oh, you're right, I missed that at first, sorry about the confusion.

ktsaou:
On virtual interfaces you can set the variable to 0 and check it at the alarm to disable the check. So, on virtual interfaces and alarm will never trigger.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

Ehekatl:
> check it at the alarm to disable the check
@ktsaou as it now, divided by inf/0 implicitly disable the alarm, how can I explicit disable the alarm by checking variable? Didn't find any doc.

ktsaou:
The problem now is that you don't check if this is a virtual interfaces. So, if this is virtual and the interface somehow has a max speed set, it will trigger (probably) false alarms.

You can check an alarm for `inf` and `nan` like these values are numbers. It is explained here: https://github.com/netdata/netdata/blob/master/health/README.md#expressions

However, I think you could disable the alarms on virtual interfaces with this:

```
calc: ($nic_speed_max > 0) ? ($this * 100 / ($nic_speed_max * 1000)) : ( nan )
```


So, on virtual interfaces it won't even evaluate the alarm.

Ehekatl:
@ktsaou I did check if interface is virtual when read the speed, so on non-virtual interfaces & a successful read, the max speed will be something > 0, otherwise it will always be 0 (on virtual interface or fail to read speed on SR-IOV device, etc)

agree it would be better to explicit check `$nic_speed_max > 0`, but there shouldn't be any false alarms.

ktsaou:
You are right. You did.

ktsaou:
Thank you!

PR #4440, opened 2018-10-19
Indicate FreeIPMI support for FreeBSD

openspork:
Update docs to indicate that FreeIPMI works on FreeBSD

##### Component Name

freeipmi

##### Additional Information

```paste below

```


ktsaou:
Does it work?
have you tried it?

openspork:
Yes, it works fine.  I did not know it was not regarded as working and thought this was just a mistake in the docs.

The FreeBSD port does not have FreeIPMI support enabled by default but it is easily set when installing the port and enabling with "make config" (see port docs: https://www.freshports.org/net-mgmt/netdata).  Once installed I only needed to setuid 4555 the freeipmi.plugin binary.

I just tried reinstalling from GitHub's source -- looking at https://github.com/netdata/netdata/blob/master/CMakeLists.txt I guess this is to be expected.

If FreeIPMI support was added downstream it will probably be very easy to upstream for somebody with development experience.

ktsaou:
Thanks! Merged!
I also updated CMakeLists.txt to build freeipmi.plugin on freebsd, when libipmimonitoring is available.

PR #4443, opened 2018-10-19
modularize the query api

ktsaou:
This PR modularizes the database query engine of netdata.

The new engine supports plugable modules to execute the query. Currently the following methods are supported:

1. `average`
2. `min`
3. `max`
4. `sum`
5. `incremental-sum`
6. `median` (NEW)
7. `ses` simple/single exponential smoothing (NEW)
8. `stddev` standard deviation (NEW)

All these methods can be used in `lookup` lines of alarms and `group` API parameter for charts and badges.

Here is a screenshot of the new averaging methods:

image

## rewrite

This is a re-write of the old engine. The old one, for each query made, was evaluating database values from newer to older. This was actually the reason we haven't added exponential smoothing so far, as these algorithms require access to the data source from the older to the newer values (or otherwise we should cache them in an array to be passed to the statistical functions). This implementation calculates exponential smoothing on the fly.

The new engine walks through from older to newer database values. This required quite a few changes to the calculations made the the API to return the requested timeframe (given that the engine supports **data reduction** and **resampling** of the metrics).

The new code is a lot cleaner and when compiled with with `-DNETDATA_INTERNAL_CHECKS` it also runs many checks to verify the code the stable.

## performance

The new code is marginal faster. A `systems.cpu` query for the `average` calculation with the `static-threaded` web server set to 1 thread, was giving a max of 3.878 queries/sec, now it maxes at 3.973 queries/sec. Also, netdata was able to generate 761Mbps/s of chart data, vs 740Mbps/s with the older engine (using the same single threaded web server - quite impressive).

On the charts below, the left spike is the old code, the right spike is the the new code.

Requests/sec served:

image

Bandwidth sent by netdata:

image


ktsaou:
The codacy reported issue is a false positive.

vlvkobal:
License Identifier is missing in `netdata/web/api/queries/query.c` and `netdata/web/api/queries/query.h`

vlvkobal:
License Identifier is missing in `netdata/web/api/queries/query.c` and `netdata/web/api/queries/query.h`

PR #4460, opened 2018-10-22
Remove duplicated entry and put modules in order in python.d.conf

vladmovchan:
##### Summary
* Remove duplicated `ntpd` entry
* Sort modules in alphabetical order to reduce chance of appearance of duplicated records in future

##### Component Name
`python.d.conf`


l2isbad:
@vladmovchan thx!

PR #4472, opened 2018-10-24
bug-fix: fixed aligned queries that returned no data

ktsaou:
##### Summary

Fixes bug introduced with #4443

Aligned queries were broken under conditions. They returned no data.
This was visible in badges and aligned alarms (none exists, so only custom alarms were affected).

The bug only appeared when the query time-frame was `before=0&after=-something`. Due to a calculation bug, it resulted in zero-duration to be queried.

##### Additional Information

```paste below

```


ktsaou:
@vlvkobal I would appreciate it if you could unit test the query engine. Unfortunately we can't remember all the use cases, and this makes it extremely difficult to make changes to it.

Issue #4473, opened 2018-10-24
Logcheck security alert: netdata : command not allowed ; TTY=unknown ; PWD=/etc/netdata ; USER=root ; COMMAND=validate


lichtamberg:
3 hours ago nearly all of our servers produced a security alert, which contained following information:

`
Oct 24 06:25:50 : netdata : command not allowed ; TTY=unknown ; PWD=/etc/netdata ; USER=root ; COMMAND=validate
`
Since I'm pretty sure our servers haven't been hacked, I think it's more a bug somehow? 
Did netdata update itself and try to execute something with sudo? Any hints appreciated.

Our netdata configuration has not been changed in the last weeks/months, so I'm a bit curious why this happens now?

ktsaou:
Probably your netdata was updated and restarted.

A few netdata plugins execute a `sudo` with validate option to check if they can collect data.

This is needed for:

```sh
# grep -l sudo $(find collectors -name \*.py -o -name \*.js -o -name \*.sh)
collectors/python.d.plugin/adaptec_raid/adaptec_raid.chart.py
collectors/python.d.plugin/samba/samba.chart.py
collectors/python.d.plugin/megacli/megacli.chart.py
collectors/charts.d.plugin/libreswan/libreswan.chart.sh
```

The adaptec module was added yesterday. So, you probably have netdata auto-updated via cron, it updated itself yesterday and the new adaptec module tried to see if it is able to run.

If you don't want this autodetection, you can disable the `adaptec` plugin by editing `python.d.conf` using `/etc/netdata/edit-config python.d.conf`.

@l2isbad I suggest all plugins that require `sudo` to be disabled by default. So, they should run only if the user has explicitly enabled them. Another solution could be to check something else before running `sudo`. For example, libreswan checks that the command `ipsec` is available before running it: https://github.com/netdata/netdata/blob/e5aca2581e137337f384dc5d48ae86593b808bd3/collectors/charts.d.plugin/libreswan/libreswan.chart.sh#L87

Could you please do this?

ktsaou:
I marked this a bug. We should fix this. It is not nice for people to receive such alerts without being aware of why this is needed, or without being obvious that they would need to allow `sudo` for netdata data collection modules.

ktsaou:
Also the adaptec module page does not say anything about this: https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/adaptec_raid

[...]

PR #4477, opened 2018-10-24
Disable python sudo modules by default

l2isbad:
##### Summary

Fixes: #4473
- adaptec_raid fix: check if `arcconf` is available before `sudo` checks
- disable python modules which use `sudo` by default  (adaptec_raid, megacli, samba)
- update readme

##### Component Name

:snake: python.d modules: :snake:

- adaptec_raid
- megacli
- samba

##### Additional Information

```paste below

```


l2isbad:
@petarkozic
explicitly enable `megacli` and `adaptec_raid` in `pythond.d.conf`

PR #4488, opened 2018-10-25
Cleanup dashboard.js (first round)

gmosx:
Performed mild refactoring of dashboard.js as described in #4461

### DIFF
--- a/web/gui/dashboard.js
+++ b/web/gui/dashboard.js
@@ -147,7 +147,7 @@ const NETDATA = window.NETDATA || {};
                     if (Array.isArray(obj) === true) {
                         // console.log('checking array "' + name + '"');
 
-                        const len = obj.length;
+                        let len = obj.length;
                         while (len--)
                             obj[len] = this.object(name + '[' + len + ']', obj[len], ignore_regex);
                     }
@@ -388,6 +388,7 @@ const NETDATA = window.NETDATA || {};
         // backward compatibility
         netdataRegistry = (typeof netdataNoRegistry !== 'undefined' && netdataNoRegistry === false);
     }
+
     if (netdataRegistry === false && typeof netdataRegistryCallback === 'function')
         netdataRegistry = true;
 
@@ -397,8 +398,9 @@ const NETDATA = window.NETDATA || {};
 
     let isSlowDeviceResult = undefined;
     let isSlowDevice = function() {
-        if (isSlowDeviceResult !== undefined)
+        if (isSlowDeviceResult !== undefined) {
             return isSlowDeviceResult;
+        }
 
         try {
             let ua = navigator.userAgent.toLowerCase();
@@ -406,8 +408,7 @@ const NETDATA = window.NETDATA || {};
             let iOS = /ipad|iphone|ipod/.test(ua) && !window.MSStream;
             let android = /android/.test(ua) && !window.MSStream;
             isSlowDeviceResult = (iOS === true || android === true);
-        }
-        catch (e) {
+        } catch (e) {
             isSlowDeviceResult = false;
         }
 
@@ -616,8 +617,7 @@ const NETDATA = window.NETDATA || {};
                 this.clear = function(handle) {
                     return window.cancelAnimationFrame(handle.value);
                 };
-            }
-            else if (window.webkitRequestAnimationFrame) {
+            } else if (window.webkitRequestAnimationFrame) {
                 this.step = function(callback) {
                     return window.webkitRequestAnimationFrame(callback);
                 };
@@ -632,8 +632,7 @@ const NETDATA = window.NETDATA || {};
                         return window.webkitCancelRequestAnimationFrame(handle.value);
                     };
                 }
-            }
-            else if (window.mozRequestAnimationFrame) {
+            } else if (window.mozRequestAnimationFrame) {
                 this.step = function(callback) {
                     return window.mozRequestAnimationFrame(callback);
                 };
@@ -641,8 +640,7 @@ const NETDATA = window.NETDATA || {};
                 this.clear = function(handle) {
                     return window.mozCancelRequestAnimationFrame(handle.value);
                 };
-            }
-            else if (window.oRequestAnimationFrame) {
+            } else if (window.oRequestAnimationFrame) {
                 this.step = function(callback) {
                     return window.oRequestAnimationFrame(callback);
                 };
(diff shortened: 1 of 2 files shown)


web/gui/dashboard.js
```diff
@@ -2458,22 +2494,23 @@ var NETDATA = window.NETDATA || {};
...
+                    return function (value) { 
+                        return value; 
+                    };
                 }
-           }
-           else {
+           } else {
                 // hm... did we forget to implement the new type?
-                console.log('Unmatched unit conversion method for units ' + units.toString());
+                console.log(`Unmatched unit conversion method for units ${units.toString()}`);
```
ktsaou: hm... I didn't know this can be done...

web/gui/dashboard.js
```diff
@@ -2762,6 +2813,12 @@ var NETDATA = window.NETDATA || {};
             console.log(this.id + ' (' + this.library_name + ' ' + this.uuid + '): ' + msg);
         };
 
+        this.debugLog = function(msg) {
+            if (this.debug) {
+                this.log(msg);
+            }
+        }
+
```
ktsaou: This did not exist...

web/gui/dashboard.js
```diff
@@ -3049,14 +3106,15 @@ var NETDATA = window.NETDATA || {};
             that.tm.last_dom_created = 0;
         };
 
-        var createDOM = function() {
-            if(that.enabled === false) return;
+        let createDOM = function() {
```
ktsaou: All other similar functions are `const`. Why this is `let`?

web/gui/dashboard.js
```diff
@@ -2344,32 +2374,40 @@ var NETDATA = window.NETDATA || {};
...
+                    // for (x in this.scalableUnits[units]) {
+                    //     if (this.scalableUnits[units].hasOwnProperty(x)) {
+                    //         let m = this.scalableUnits[units][x];
+                    //         if (m <= max && m > tdivider) {
+                    //             tunits = x;
+                    //             tdivider = m;
+                    //         }
+                    //     }
+                    // }
+                    const sunit = this.scalableUnits[units]
```
ktsaou: missing semicolon.

web/gui/dashboard.js
```diff
@@ -6868,61 +6881,61 @@ var NETDATA = window.NETDATA || {};
...
 
                             return [ axis[0] + foo[0], axis[1] - foo[1] ];
                         }
 
-                        var yAxes = g.yAxisRanges();
-                        var newYAxes = [];
-                        for (var i = 0; i < yAxes.length; i++) {
+                        let yAxes = g.yAxisRanges();
+                        let newYAxes = [];
+                        for (let i = 0; i < yAxes.length; i++) {
```
ktsaou: I guess `for (const` would also do.

web/gui/dashboard.js
```diff
@@ -7380,15 +7382,15 @@ var NETDATA = window.NETDATA || {};
...
-        var colors = state.chartColors();
-        var len = data.result.labels.length;
-        for(var i = 1; i < len ; i++) {
-            var label = data.result.labels[i];
-            var value = data.result.data[index][label];
-            var color = colors[i - 1];
+        let content = [];
+        let colors = state.chartColors();
+        let len = data.result.labels.length;
+        for (let i = 1; i < len ; i++) {
```
ktsaou: I guess `for (const` would also do.

web/gui/dashboard.js
```diff
@@ -8866,31 +8866,31 @@ var NETDATA = window.NETDATA || {};
...
+            let active = {};
+            let data = NETDATA.alarms.current;
 
-            if(typeof data === 'undefined' || data === null)
+            if (typeof data === 'undefined' || data === null)
                 return active;
 
-            for(var x in data.alarms) {
-                if(!data.alarms.hasOwnProperty(x)) continue;
+            for (let x in data.alarms) {
```
ktsaou: I guess `for (const` would also do.

web/gui/dashboard.js
```diff
@@ -1332,9 +1331,13 @@ var NETDATA = window.NETDATA || {};
...
 
             // find the common min
-            var m = min;
-            for(var i in t)
-                if(t.hasOwnProperty(i) && t[i] < m) m = t[i];
+            let m = min;
+            // for (let i in t) {
+            //     if (t.hasOwnProperty(i) && t[i] < m) m = t[i];
+            // }
+            for (let ti of Object.values(t)) {
```
ktsaou: I guess `for (const` would also do.

ktsaou:
Issues that may exist:

## difference between `var` and `let`

According to the docs:

> `var` and `let` are both used for function declaration in javascript but the difference between them is that `var` is function scoped and `let` is block scoped. It can be said that a variable declared with `var` is defined throughout the program as compared to `let`.

I recall there were a few cases that I was getting IDE warnings about duplicate `var` declarations. Normally I would have declared the variable at the parent scope, but this is not required with `var` and the IDE would haven't warn me. Let's hope there is no such case...

## checking booleans explicitly with `===`

According the ECMAScript `if()` calls `ToBoolean(GetValue(exprRef))` http://www.ecma-international.org/ecma-262/5.1/#sec-12.5

I don't know why I had the impression `if(x === true)` is faster than `if(x)`. Based on the description above, no matter how the expression is given, JavaScript will still evaluate it with `ToBoolean(GetValue(expr))`.

This change may still introduce issues, depending on the logic. For example `if(!x)` will return true when `x` is undefined, but `if(x === false)` will return false. Anyway, I hope I never did that (it is stupid to use it like that, I agree).

---

So, let me test it...

ktsaou:
It seems stable. I can't find any issue with this.

ktsaou:
George I am ok to merge this. But at the same time it is not required, since you are also re-organizing `dashboard.js` (based on this PR).

So, I think the best solution is to merge the final one.

As far as I can tell, this seems stable, so it is up to you to decide.

web/gui/dashboard.js
gmosx: Yes, javascript supports interpolation (it's faster too ;-))

gmosx:
Yes, let's merge the final PR, after the split. I will resolve the issues you identified.

web/gui/dashboard.js
```diff
@@ -3729,33 +3781,30 @@ var NETDATA = window.NETDATA || {};
             // of a selection sync and another chart becomes
             // the new master
 
-            if(dont_update !== true && this.isVisible() === true) {
+            if (dont_update !== true && this.isVisible()) {
```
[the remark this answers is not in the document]
gmosx: What is the problem here?

web/gui/dashboard.js
gmosx: Will add it.

(FWIW, semicolons are not really required in JS, unlike C, in my own code I don't add semicolons, but here let's keep them because it's a more standard coding convention)

web/gui/dashboard.js
gmosx: Nope, `i` is mutated (i.e. `i++`).

ktsaou:
closes in favor of #4496

Issue #1236, opened 2016-11-10
snmp, wrong values.


cis2131:
I am trying to messure signal quality in a sat headend.

My OID gives:
.1.3.6.1.4.1.7465.20.2.9.4.4.3.1.2.1.5.1.3 = STRING: "14.000000"

node.d.plugin 1 snmp debug, shows:
2016-10-04 15:19:40: node.d.plugin: DEBUG: snmp: 172.16.10.209: making 2 entries request, max is: 50
2016-10-04 15:19:40: node.d.plugin: DEBUG: snmp: 172.16.10.209: got valid snmp response: [ { oid: '1.3.6.1.4.1.7465.20.2.9.4.4.4.1.2.1.5.1.2',
    type: 4,
    value: <Buffer 31 34 2e 32 35 30 30 30 30> },
  { oid: '1.3.6.1.4.1.7465.20.2.9.4.4.4.1.2.1.5.1.1',
    type: 4,
    value: <Buffer 31 35 2e 35 30 30 30 30 30> } ]
2016-10-04 15:19:40: node.d.plugin: DEBUG: snmp: 172.16.10.209: found snmp value of OIDs 1.3.6.1.4.1.7465.20.2.9.4.4.4.1.2.1.5.1.2 = 14.250000
2016-10-04 15:19:40: node.d.plugin: DEBUG: snmp: 172.16.10.209: found snmp value of OIDs 1.3.6.1.4.1.7465.20.2.9.4.4.4.1.2.1.5.1.1 = 15.500000
2016-10-04 15:19:40: node.d.plugin: DEBUG: snmp: 172.16.10.209: finished snmp with 2 successful and 0 failed values
2016-10-04 15:19:40: node.d.plugin: DEBUG: snmp: 172.16.10.209: processing snmp response (received in 10 ms)
2016-10-04 15:19:40: node.d.plugin: DEBUG: setting current chart to snmp_switch.bandwidth_port1
BEGIN snmp_switch.bandwidth_port1 2917000
SET db = 0x31342e323530303030
END

2016-10-04 15:19:40: node.d.plugin: DEBUG: sent chart snmp_switch.bandwidth_port1
2016-10-04 15:19:40: node.d.plugin: DEBUG: setting current chart to snmp_switch.bandwidth_port2
BEGIN snmp_switch.bandwidth_port2 2917000
SET db = 0x31352e353030303030
END

And the graphs ends up showing 0.

What am I doing wrong?.

[...]

PR #4498, opened 2018-10-26
Fix(snmp): fix parse oidname and santilize dimension name

Ehekatl:
##### Summary

this is a quickfix to make oidname replacement working in node.d snmp plugin.

##### Component Name

node.d snmp plugin

### DIFF
--- a/collectors/node.d.plugin/snmp/snmp.node.js
+++ b/collectors/node.d.plugin/snmp/snmp.node.js
@@ -171,7 +171,7 @@ netdata.processors.snmp = {
 
                     var oid = this.fixoid(dim.oid);
                     var oidname = this.fixoid(dim.oidname);
-                    
+
                     if(__DEBUG === true)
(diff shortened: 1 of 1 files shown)


ktsaou:
@Ehekatl thank you!

This change seems simple, but could you please describe the motivation behind it? Why do we need this? What was the problem you solved?

Ehekatl:
@ktsaou before this change, use oidname in dimension will parse as float and result in `NaN` if oidname it's not a number,  I also sanitize the value to replace space/special character as underscore, make sure the dimension name is valid.

Ehekatl:
actually, I didn't get why there is case that `net_snmp.ObjectType.OctetString` need parse as float...

ktsaou:
oh, I see.

The reason for parsing strings as float: https://github.com/netdata/netdata/issues/1236

I guess you are trying to solve this: https://github.com/netdata/netdata/issues/4512

@agross can you check if this PR fixes your issue?

agross:
@ktsaou Yes, it fixes the issue with `oidname` being treated as a "string parsed as float".

ktsaou:
perfect!
So, I am merging it.

ktsaou:
Thank you @Ehekatl 	!

PR #4508, opened 2018-10-27
do not send duplicate chart names while streaming metrics

### LINKED ISSUE #4468: duplicate name in cgroup if dash present in container name
##### Bug report summary
I have some lxd container and if the container name has one or more dash the metric name will show a duplicated name: 
example:
container name: node-red
cpu metric: cgroup_lxc_node_red.cgroup_lxc_node_red.cpu

if the container name doesn't have a dash this problem is not present
example:
container name: db
cpu metric: cgroup_lxc_db.cpu

##### OS / Environment
ubuntu 18.04/ ubuntu 16.04

ktsaou:
##### Summary

This PR fixes duplicate chart names sent via streaming.
Fixes #4468

##### Additional Information

In PR #2749 we tried to fix that at the master side. Unfortunately, this cannot be fixed reliable at that side, because the `type` chart of the chart name may not match the chart `type` of id (i.e. the user at issue #4468 found a corner case, that names are left intect but hyphens in ids are changed to underscores).

So, this PR attempts to fix it at the slave side.

```paste below

```

### DIFF
--- a/streaming/rrdpush.c
+++ b/streaming/rrdpush.c
@@ -149,12 +149,25 @@ static inline void rrdpush_send_chart_definition_nolock(RRDSET *st) {
 
     rrdset_flag_set(st, RRDSET_FLAG_UPSTREAM_EXPOSED);
 
+    // properly set the name for the remote end to parse it
+    char *name = "";
+    if(unlikely(strcmp(st->id, st->name))) {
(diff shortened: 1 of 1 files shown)


ktsaou:
checked with `vlagrind` too.

PR #4513, opened 2018-10-28
openldap monitoring plugin added

ekartsonakis:
##### Summary
External python plugin for openldap monitoring

##### Component Name
openldap plugin

##### Additional Information
This is an optional plugin turned off by default. Not sure if I have to include whole ldap python module under `python_modules`.

### DIFF
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -34,7 +34,7 @@ CHARTS = {
             ['total_connections', 'connections', 'incremental']
         ]
     },
-  
+
     'bytes_sent': {
         'options': [None, 'Bytes Statistics', 'Bytes', 'ldap', 'monitorCounter', 'line'],
         'lines': [
@@ -150,7 +150,7 @@ class Service(SimpleService):
             self.conn.unbind()
             self.conn = None
             self.alive = False
-        
+
 
     def connect(self):
         try:
@@ -198,6 +198,6 @@ class Service(SimpleService):
                 val = 'NaN'
 
             data[key] = val
-        
+
         return data
 
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -21,33 +21,33 @@ CHARTS = {
     'lines': [
         ['total_connections', None, 'incremental']
      ]},
-    'bytes_sent': { 
+    'bytes_sent': {
     'options': [None, 'Bytes Statistics', 'Bytes', 'ldap', 'monitorCounter', 'line'],
     'lines': [
-        ['bytes_sent', None, 'incremental']  
+        ['bytes_sent', None, 'incremental']
      ]},
     'operations': {
        'options': [None, 'Operations', 'ops', 'ldap', 'monitorCounter', 'line'],
        'lines': [
-        ['completed_operations', None, 'incremental'], 
-        ['initiated_operations', None, 'incremental']  
+        ['completed_operations', None, 'incremental'],
+        ['initiated_operations', None, 'incremental'] 
      ]},
     'referrals_sent': {
     'options': [None, 'Referrals Statistics', 'referalls_sent', 'ldap', 'monitorCounter', 'line'],
     'lines': [
-        ['referrals_sent', None, 'incremental']  
+        ['referrals_sent', None, 'incremental']
      ]},
     'entries_sent': {
     'options': [None, 'Entries Statistics', 'entries_sent', 'ldap', 'monitorCounter', 'line'],
     'lines': [
-        ['entries_sent', None, 'incremental']  
+        ['entries_sent', None, 'incremental']
      ]},
     'ldap_operations': {
     'options': [None, 'Operations', 'bind_operations', 'ldap', 'monitorCounter', 'line'],
     'lines': [
-        ['bind_operations', None, 'incremental'],  
+        ['bind_operations', None, 'incremental'],
         ['search_operations', None, 'incremental'],
-        ['unbind_operations', None, 'incremental'],  
+        ['unbind_operations', None, 'incremental'],
         ['add_operations', None, 'incremental'],
         ['delete_operations', None, 'incremental'],
         ['modify_operations', None, 'incremental'],
@@ -96,7 +96,7 @@ class Service(SimpleService):
         'write_waiters': ('cn=Write,cn=Waiters,cn=Monitor','monitorCounter'),
         'read_waiters': ('cn=Read,cn=Waiters,cn=Monitor','monitorCounter'),
         }
-        
+
         for key in searchlist.keys():
             bind = searchlist[key][0]
             attr = searchlist[key][1]
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -139,6 +139,7 @@ class Service(SimpleService):
         self.port = configuration.get('port')
         self.username = configuration.get('username')
         self.password = configuration.get('password')
+        self.timeout = configuration.get('timeout')
 
         self.alive = False
         self.conn = None
@@ -153,6 +154,7 @@ class Service(SimpleService):
         try:
             self.conn = ldap.initialize('ldap://%s:%s' % (self.server, self.port))
             self.conn.simple_bind(self.username, self.password)
+            self.conn.set_option(ldap.OPT_NETWORK_TIMEOUT, self.timeout)
         except ldap.LDAPError as error:
             self.error(error)
             return False
@@ -176,13 +178,14 @@ class Service(SimpleService):
             return None
 
         data = dict()
-        for key in SEARCH_LIST.keys():
+        for key in SEARCH_LIST:
             bind = SEARCH_LIST[key][0]
             attr = SEARCH_LIST[key][1]
             try:
                 num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr, ])
                 result_type, result_data = self.conn.result(num, 1)
             except ldap.LDAPError as error:
+                self.error(error)
                 self.alive = False
                 return None
 
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -153,8 +153,8 @@ class Service(SimpleService):
     def connect(self):
         try:
             self.conn = ldap.initialize('ldap://%s:%s' % (self.server, self.port))
-            self.conn.simple_bind(self.username, self.password)
             self.conn.set_option(ldap.OPT_NETWORK_TIMEOUT, self.timeout)
+            self.conn.simple_bind(self.username, self.password)
         except ldap.LDAPError as error:
             self.error(error)
             return False
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -139,8 +139,8 @@ class Service(SimpleService):
 
         self.server = configuration.get('server', DEFAULT_SERVER)
         self.port = configuration.get('port', DEFAULT_PORT)
-        self.username = configuration.get('username', None)
-        self.password = configuration.get('password', None)
+        self.username = configuration.get('username')
+        self.password = configuration.get('password')
         self.timeout = configuration.get('timeout', DEFAULT_TIMEOUT)
 
         self.alive = False
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -29,7 +29,7 @@ CHARTS = {
        'options': [None, 'Operations', 'ops', 'ldap', 'monitorCounter', 'line'],
        'lines': [
         ['completed_operations', None, 'incremental'],
-        ['initiated_operations', None, 'incremental'] 
+        ['initiated_operations', None, 'incremental']
      ]},
     'referrals_sent': {
     'options': [None, 'Referrals Statistics', 'referalls_sent', 'ldap', 'monitorCounter', 'line'],
@@ -100,7 +100,7 @@ class Service(SimpleService):
             bind = searchlist[key][0]
             attr = searchlist[key][1]
             num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr,])
-        
+
             try:
                 result_type, result_data = self.conn.result(num, 1)
                 if result_type == 101:
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -14,7 +14,6 @@ from bases.FrameworkServices.SimpleService import SimpleService
 # default module values (can be overridden per job in `config`)
 priority = 60000
 
-
 ORDER = [
     'total_connections',
     'bytes_sent',
@@ -138,7 +137,7 @@ class Service(SimpleService):
         self.port = configuration.get('port')
         self.username = configuration.get('username')
         self.password = configuration.get('password')
-        self.timeout = configuration.get('timeout')
+        self.timeout = configuration.get('timeout', 1)
 
         self.alive = False
         self.conn = None
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -3,10 +3,10 @@
 # Author: Manolis Kartsonakis (ekartsonakis)
 # SPDX-License-Identifier: GPL-3.0+
 
-try: 
-     import ldap 
-     HAS_LDAP = True 
-except ImportError: 
+try:
+     import ldap
+     HAS_LDAP = True
+except ImportError:
      HAS_LDAP = False
 
 from bases.FrameworkServices.SimpleService import SimpleService
@@ -27,7 +27,7 @@ ORDER = [
 ]
 
 CHARTS = {
-  
+
      'total_connections': {
         'options': [None, 'Total Connections', 'connections', 'ldap', 'monitorCounter', 'line'],
         'lines': [
@@ -170,8 +170,8 @@ class Service(SimpleService):
 
 
     def check(self):
-        if not HAS_LDAP: 
-            self.error("'python-ldap' package is needed") 
+        if not HAS_LDAP:
+            self.error("'python-ldap' package is needed")
             return None
 
         return self.connect() and self.get_data()
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -188,7 +188,7 @@ class Service(SimpleService):
             try:
                 num = self.conn.search(dn, ldap.SCOPE_BASE, 'objectClass=*', [attr, ])
                 result_type, result_data = self.conn.result(num, 1)
-            except Exception as error:
+            except ldap.LDAPError as error:
                 self.error("Empty result. Check bind username/password. Message: ",error)
                 self.alive = False
                 return None
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -17,8 +17,6 @@ priority = 60000
 DEFAULT_SERVER = 'localhost'
 DEFAULT_PORT = '389'
 DEFAULT_TIMEOUT = 1
-DEFAULT_USERNAME = ''
-DEFAULT_PASSWORD = ''
 
 ORDER = [
     'total_connections',
@@ -141,8 +139,8 @@ class Service(SimpleService):
 
         self.server = configuration.get('server', DEFAULT_SERVER)
         self.port = configuration.get('port', DEFAULT_PORT)
-        self.username = configuration.get('username', DEFAULT_USERNAME)
-        self.password = configuration.get('password', DEFAULT_PASSWORD)
+        self.username = configuration.get('username', None)
+        self.password = configuration.get('password', None)
         self.timeout = configuration.get('timeout', DEFAULT_TIMEOUT)
 
         self.alive = False
@@ -158,7 +156,8 @@ class Service(SimpleService):
         try:
             self.conn = ldap.initialize('ldap://%s:%s' % (self.server, self.port))
             self.conn.set_option(ldap.OPT_NETWORK_TIMEOUT, self.timeout)
-            self.conn.simple_bind(self.username, self.password)
+            if self.username and self.password:
+                self.conn.simple_bind(self.username, self.password)
         except ldap.LDAPError as error:
             self.error(error)
             return False
@@ -196,7 +195,7 @@ class Service(SimpleService):
             try:
                 if result_type == 101:
                     val = int(result_data[0][1].values()[0][0])
-            except ( ValueError, IndexError) as error:
+            except (ValueError, IndexError) as error:
                 self.debug(error)
                 continue
 
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -11,9 +11,8 @@ update_every = 10
 priority = 60000
 retries = 60
 
-
-
-ORDER = ['total_connections', 'bytes_sent', 'operations', 'referrals_sent', 'entries_sent', 'ldap_operations', 'waiters']
+ORDER = ['total_connections', 'bytes_sent', 'operations',
+         'referrals_sent', 'entries_sent', 'ldap_operations', 'waiters']
 
 CHARTS = {
     'total_connections': {
@@ -80,27 +79,27 @@ class Service(SimpleService):
 
         # Stuff to gather - make tuples of DN dn and attrib to get
         searchlist = {
-        'total_connections':('cn=Total,cn=Connections,cn=Monitor','monitorCounter'),
-        'bytes_sent': ('cn=Bytes,cn=Statistics,cn=Monitor','monitorCounter'),
-        'completed_operations': ('cn=Operations,cn=Monitor','monitorOpCompleted'),
-        'initiated_operations': ('cn=Operations,cn=Monitor','monitorOpInitiated'),
-        'referrals_sent': ('cn=Referrals,cn=Statistics,cn=Monitor','monitorCounter'),
-        'entries_sent': ('cn=Entries,cn=Statistics,cn=Monitor','monitorCounter'),
-        'bind_operations': ('cn=Bind,cn=Operations,cn=Monitor','monitorOpCompleted'),
-        'unbind_operations': ('cn=Unbind,cn=Operations,cn=Monitor','monitorOpCompleted'),
-        'add_operations': ('cn=Add,cn=Operations,cn=Monitor','monitorOpInitiated'),
-        'delete_operations':  ('cn=Delete,cn=Operations,cn=Monitor','monitorOpCompleted'),
-        'modify_operations': ('cn=Modify,cn=Operations,cn=Monitor','monitorOpCompleted'),
-        'compare_operations': ('cn=Compare,cn=Operations,cn=Monitor','monitorOpCompleted'),
-        'search_operations': ('cn=Search,cn=Operations,cn=Monitor','monitorOpCompleted'),
-        'write_waiters': ('cn=Write,cn=Waiters,cn=Monitor','monitorCounter'),
-        'read_waiters': ('cn=Read,cn=Waiters,cn=Monitor','monitorCounter'),
+        'total_connections':('cn=Total,cn=Connections,cn=Monitor', 'monitorCounter'),
+        'bytes_sent': ('cn=Bytes,cn=Statistics,cn=Monitor', 'monitorCounter'),
+        'completed_operations': ('cn=Operations,cn=Monitor', 'monitorOpCompleted'),
+        'initiated_operations': ('cn=Operations,cn=Monitor', 'monitorOpInitiated'),
+        'referrals_sent': ('cn=Referrals,cn=Statistics,cn=Monitor', 'monitorCounter'),
+        'entries_sent': ('cn=Entries,cn=Statistics,cn=Monitor', 'monitorCounter'),
+        'bind_operations': ('cn=Bind,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
+        'unbind_operations': ('cn=Unbind,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
+        'add_operations': ('cn=Add,cn=Operations,cn=Monitor', 'monitorOpInitiated'),
+        'delete_operations':  ('cn=Delete,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
+        'modify_operations': ('cn=Modify,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
+        'compare_operations': ('cn=Compare,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
+        'search_operations': ('cn=Search,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
+        'write_waiters': ('cn=Write,cn=Waiters,cn=Monitor', 'monitorCounter'),
+        'read_waiters': ('cn=Read,cn=Waiters,cn=Monitor', 'monitorCounter'),
         }
 
         for key in searchlist.keys():
             bind = searchlist[key][0]
             attr = searchlist[key][1]
-            num = self.conn.search(bind,ldap.SCOPE_BASE,'objectClass=*',[attr,])
+            num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr,])
         
             try:
                 result_type, result_data = self.conn.result(num, 1)
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -17,6 +17,8 @@ priority = 60000
 DEFAULT_SERVER = 'localhost'
 DEFAULT_PORT = '389'
 DEFAULT_TIMEOUT = 1
+DEFAULT_USERNAME = ''
+DEFAULT_PASSWORD = ''
 
 ORDER = [
     'total_connections',
@@ -139,8 +141,8 @@ class Service(SimpleService):
 
         self.server = configuration.get('server', DEFAULT_SERVER)
         self.port = configuration.get('port', DEFAULT_PORT)
-        self.username = configuration.get('username')
-        self.password = configuration.get('password')
+        self.username = configuration.get('username', DEFAULT_USERNAME)
+        self.password = configuration.get('password', DEFAULT_PASSWORD)
         self.timeout = configuration.get('timeout', DEFAULT_TIMEOUT)
 
         self.alive = False
@@ -181,20 +183,21 @@ class Service(SimpleService):
 
         data = dict()
         for key in SEARCH_LIST:
-            bind = SEARCH_LIST[key][0]
+            dn = SEARCH_LIST[key][0]
             attr = SEARCH_LIST[key][1]
             try:
-                num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr, ])
+                num = self.conn.search(dn, ldap.SCOPE_BASE, 'objectClass=*', [attr, ])
                 result_type, result_data = self.conn.result(num, 1)
-            except ldap.LDAPError as error:
-                self.error(error)
+            except Exception as error:
+                self.error("Empty result. Check bind username/password. Message: ",error)
                 self.alive = False
                 return None
 
             try:
                 if result_type == 101:
                     val = int(result_data[0][1].values()[0][0])
-            except ValueError:
+            except ( ValueError, IndexError) as error:
+                self.debug(error)
                 continue
 
             data[key] = val
--- a/collectors/python.d.plugin/openldap/openldap.chart.py
+++ b/collectors/python.d.plugin/openldap/openldap.chart.py
@@ -1,0 +1,114 @@
+# -*- coding: utf-8 -*-
+# Description: openldap netdata python.d module
+# Author: Manolis Kartsonakis (ekartsonakis)
+# SPDX-License-Identifier: GPL-3.0+
+
+import ldap
+from bases.FrameworkServices.SimpleService import SimpleService
+
+# default module values (can be overridden per job in `config`)
+update_every = 10
+priority = 60000
+retries = 60
+
+
+
+ORDER = ['total_connections', 'bytes_sent', 'operations', 'referrals_sent', 'entries_sent', 'ldap_operations', 'waiters']
+
+CHARTS = {
+    'total_connections': {
+	'options': [None, 'Total Connections', 'connections', 'ldap', 'monitorCounter', 'line'],
+	'lines': [
+		['total_connections', None, 'incremental']
+     ]},
+    'bytes_sent': { 
+	'options': [None, 'Bytes Statistics', 'Bytes', 'ldap', 'monitorCounter', 'line'],
+	'lines': [
+		['bytes_sent', None, 'incremental']  
+     ]},
+    'operations': {
+       'options': [None, 'Operations', 'ops', 'ldap', 'monitorCounter', 'line'],
+       'lines': [
+		['completed_operations', None, 'incremental'], 
+		['initiated_operations', None, 'incremental']  
+     ]},
+    'referrals_sent': {
+	'options': [None, 'Referrals Statistics', 'referalls_sent', 'ldap', 'monitorCounter', 'line'],
+	'lines': [
+		['referrals_sent', None, 'incremental']  
+     ]},
+    'entries_sent': {
+	'options': [None, 'Entries Statistics', 'entries_sent', 'ldap', 'monitorCounter', 'line'],
+	'lines': [
+		['entries_sent', None, 'incremental']  
+     ]},
+    'ldap_operations': {
+	'options': [None, 'Operations', 'bind_operations', 'ldap', 'monitorCounter', 'line'],
+	'lines': [
+		['bind_operations', None, 'incremental'],  
+		['search_operations', None, 'incremental'],
+		['unbind_operations', None, 'incremental'],  
+		['add_operations', None, 'incremental'],
+		['delete_operations', None, 'incremental'],
+		['modify_operations', None, 'incremental'],
+		['compare_operations', None, 'incremental']
+     ]},
+    'waiters': {
+	'options': [None, 'Waiters', 'amount', 'ldap', 'monitorCounter', 'line'],
+	'lines': [
+		['write_waiters', None, 'incremental'],
+		['read_waiters', None, 'incremental']
+     ]},
+}
+
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.conn = ldap.initialize('ldap://%s:%s' % (configuration.get('server'), configuration.get('port')))
+        self.conn.simple_bind(configuration.get('username'), configuration.get('password'))
+
+    def check(self):
+        if self.conn:
+            return True
+        else:
+            return False
+
+    def get_data(self):
+        data = dict()
+
+        # Stuff to gather - make tuples of DN dn and attrib to get
+        searchlist = {
+        'total_connections':('cn=Total,cn=Connections,cn=Monitor','monitorCounter'),
+        'bytes_sent': ('cn=Bytes,cn=Statistics,cn=Monitor','monitorCounter'),
+        'completed_operations': ('cn=Operations,cn=Monitor','monitorOpCompleted'),
+        'initiated_operations': ('cn=Operations,cn=Monitor','monitorOpInitiated'),
+        'referrals_sent': ('cn=Referrals,cn=Statistics,cn=Monitor','monitorCounter'),
+        'entries_sent': ('cn=Entries,cn=Statistics,cn=Monitor','monitorCounter'),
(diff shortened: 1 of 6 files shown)


ktsaou:
nice thank you!

@ekartsonakis you have to solve these 2 issues:

1. You can committed code with an email address that is not linked to your github account. You can to link your email addresses to github, to sign our CLA. Or commit them with an email you have linked to github.

2. LGTM and codacy have identified a problem at line 109 of your python code. Indentation is wrong, so your code will break on python 2.

Your code will be reviewed in detail by @l2isbad. Please follow any suggestions he makes.

ktsaou:
I also see you have not linked `Makefile.inc` to the parent's `Makefile.am`. Please do so, otherwise you module will not be installed with netdata.

ekartsonakis:
Actualy I didn't fix identation. I completely broke it :), let's try again

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,114 @@
...
+        'bind_operations': ('cn=Bind,cn=Operations,cn=Monitor','monitorOpCompleted'),
+        'unbind_operations': ('cn=Unbind,cn=Operations,cn=Monitor','monitorOpCompleted'),
+        'add_operations': ('cn=Add,cn=Operations,cn=Monitor','monitorOpInitiated'),
+        'delete_operations':  ('cn=Delete,cn=Operations,cn=Monitor','monitorOpCompleted'),
+        'modify_operations': ('cn=Modify,cn=Operations,cn=Monitor','monitorOpCompleted'),
+        'compare_operations': ('cn=Compare,cn=Operations,cn=Monitor','monitorOpCompleted'),
+        'search_operations': ('cn=Search,cn=Operations,cn=Monitor','monitorOpCompleted'),
+        'write_waiters': ('cn=Write,cn=Waiters,cn=Monitor','monitorCounter'),
+        'read_waiters': ('cn=Read,cn=Waiters,cn=Monitor','monitorCounter'),
+        }
```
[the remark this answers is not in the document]
Ferroin: And also please indent the contents.

For example, something like this:

```
searchlist = {
    'total_connections':('cn=Total,cn=Connections,cn=Monitor', 'monitorCounter'),
    'bytes_sent': ('cn=Bytes,cn=Statistics,cn=Monitor', 'monitorCounter'),
    'completed_operations': ('cn=Operations,cn=Monitor', 'monitorOpCompleted'),
    'initiated_operations': ('cn=Operations,cn=Monitor', 'monitorOpInitiated'),
    'referrals_sent': ('cn=Referrals,cn=Statistics,cn=Monitor', 'monitorCounter'),
    'entries_sent': ('cn=Entries,cn=Statistics,cn=Monitor', 'monitorCounter'),
    'bind_operations': ('cn=Bind,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
    'unbind_operations': ('cn=Unbind,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
    'add_operations': ('cn=Add,cn=Operations,cn=Monitor', 'monitorOpInitiated'),
    'delete_operations':  ('cn=Delete,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
    'modify_operations': ('cn=Modify,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
    'compare_operations': ('cn=Compare,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
    'search_operations': ('cn=Search,cn=Operations,cn=Monitor', 'monitorOpCompleted'),
    'write_waiters': ('cn=Write,cn=Waiters,cn=Monitor', 'monitorCounter'),
    'read_waiters': ('cn=Read,cn=Waiters,cn=Monitor', 'monitorCounter'),
}
```

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,114 @@
...
+        ['modify_operations', None, 'incremental'],
+        ['compare_operations', None, 'incremental']
+     ]},
+    'waiters': {
+    'options': [None, 'Waiters', 'amount', 'ldap', 'monitorCounter', 'line'],
+    'lines': [
+        ['write_waiters', None, 'incremental'],
+        ['read_waiters', None, 'incremental']
+     ]},
+}
```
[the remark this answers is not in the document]
Ferroin: Indentation for a single chart entry in this list should look like this:

```
    'total_connections': {
        'options': [None, 'Total Connections', 'connections', 'ldap', 'monitorCounter', 'line'],
        'lines': [
            ['total_connections', None, 'incremental']
        ]
    },
```

l2isbad:
yo @ekartsonakis 

if you need help or something isn't clear feel free to ask

l2isbad:
@ekartsonakis hi 

We are interested in this being merged. If you don't have spare time to finish the PR, i can do it. What do you think?

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,145 @@
...
+     HAS_LDAP = False
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+# default module values (can be overridden per job in `config`)
+update_every = 10
+priority = 60000
+retries = 60
+
+ORDER = [
```
l2isbad: please fromat it like this
https://github.com/netdata/netdata/blob/f8e0a57b2e6a77d4b0aac38c864cd02c66cd998b/collectors/python.d.plugin/adaptec_raid/adaptec_raid.chart.py#L19-L24

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,145 @@
...
+try: 
+     import ldap 
+     HAS_LDAP = True 
+except ImportError: 
+     HAS_LDAP = False
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+# default module values (can be overridden per job in `config`)
+update_every = 10
```
l2isbad: Why not 1?

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,145 @@
...
+        'options': [None, 'Total Connections', 'connections', 'ldap', 'monitorCounter', 'line'],
+        'lines': [
+            ['total_connections', None, 'incremental']
+        ]
+    },
+  
+    'bytes_sent': {
+        'options': [None, 'Bytes Statistics', 'Bytes', 'ldap', 'monitorCounter', 'line'],
+        'lines': [
+            ['bytes_sent', None, 'incremental']
```
l2isbad: 1. Please add dimension name.

```python
['bytes_sent', 'sent', 'incremental']
```

^^ please do so for all charts


2. convert bytes to KB

```pyhon
['bytes_sent', None, 'incremental', 1, 1024]
```

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,145 @@
...
+        'options': [None, 'Waiters', 'amount', 'ldap', 'monitorCounter', 'line'],
+        'lines': [
+            ['write_waiters', None, 'incremental'],
+            ['read_waiters', None, 'incremental']
+        ]
+    },
+}
+
+# Stuff to gather - make tuples of DN dn and attrib to get
+searchlist = {
```
l2isbad: 1. make it upper case

2. i suggest to change format to

<details><summary>clicke me!</summary>
<p>

```python
SEARCH_LIST = {
    'total_connections': (
        'cn=Total,cn=Connections,cn=Monitor',
        'monitorCounter',
    ),
    'bytes_sent': (
        'cn=Bytes,cn=Statistics,cn=Monitor',
        'monitorCounter',
    ),
    'completed_operations': (
        'cn=Operations,cn=Monitor',
        'monitorOpCompleted',
    ),
    'initiated_operations': (
        'cn=Operations,cn=Monitor',
        'monitorOpInitiated',
    ),
    'referrals_sent': (
        'cn=Referrals,cn=Statistics,cn=Monitor',
        'monitorCounter',
    ),
    'entries_sent': (
        'cn=Entries,cn=Statistics,cn=Monitor',
        'monitorCounter',
    ),
    'bind_operations': (
        'cn=Bind,cn=Operations,cn=Monitor',
        'monitorOpCompleted',
    ),
    'unbind_operations': (
        'cn=Unbind,cn=Operations,cn=Monitor',
        'monitorOpCompleted',
    ),
    'add_operations': (
        'cn=Add,cn=Operations,cn=Monitor',
        'monitorOpInitiated',
    ),
    'delete_operations':  (
        'cn=Delete,cn=Operations,cn=Monitor',
        'monitorOpCompleted',
    ),
    'modify_operations': (
        'cn=Modify,cn=Operations,cn=Monitor',
        'monitorOpCompleted',
    ),
    'compare_operations': (
        'cn=Compare,cn=Operations,cn=Monitor',
        'monitorOpCompleted',
    ),
    'search_operations': (
        'cn=Search,cn=Operations,cn=Monitor',
        'monitorOpCompleted',
    ),
    'write_waiters': (
        'cn=Write,cn=Waiters,cn=Monitor',
        'monitorCounter',
    ),
    'read_waiters': (
        'cn=Read,cn=Waiters,cn=Monitor',
        'monitorCounter',
    ),
}
```
</p>
</details>

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,145 @@
...
+        self.password = configuration.get('password')
+        
+    def connect(self):
+        try: 
+            print self.server
+            self.conn = ldap.initialize('ldap://%s:%s' % ( self.server, self.port))
+            self.conn.simple_bind(self.username, self.password)
+            return True
+        except:
+            return False
```
l2isbad: We need:
* connect
* reconnect
* alive flag

Thats's example. I haven't tested it. Feel free to use and modify.
<details><summary>click me!</summary>
<p>
```python
        self.order = ORDER
        self.definitions = CHARTS

        self.server = configuration.get('server', 'localhost')
        self.port = configuration.get('port', 389)
        self.user = configuration.get('username')
        self.password = configuration.get('password')

        self.alive = False
        self.conn = None

    def disconnect(self):
        if self.conn:
            self.conn.unbind()
            self.conn = None
            self.alive = False

    def connect(self):
        self.conn = ldap.initialize('ldap://{0}:{1}'.format(self.server, self.port))

        try:
            self.conn.simple_bind(self.user, self.password)
        except ldap.LDAPError as error:
            self.error(error)
            return False

        self.alive = True
        return True

    def reconnect(self):
        self.disconnect()
        return self.connect()

    def check(self):
        if not HAS_LDAP:
            self.error("'python-ldap' package is needed")
            return None

        return self.connect() and self.get_data()

    def get_data(self):
        if not self.alive and not self.reconenct():
            return None
       #  other code
```

</p>
</details>

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,145 @@
...
+        for key in searchlist.keys():
+            bind = searchlist[key][0]
+            attr = searchlist[key][1]
+            num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr,])
+            try:
+                result_type, result_data = self.conn.result(num, 1)
+                if result_type == 101:
+                    val = int(result_data[0][1].values()[0][0])
+                    data[key] = val
+            except Exception as e:
```
l2isbad: Exception is too broad. `ldap.LDAPError` might be enough here.

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,145 @@
...
+                return repr(e)
```
l2isbad: in case of excpetion log it and return `None` (set `alive` flag to `False` if you will use it)

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: I don't think that this change increase readability of the code despite the fact that it is more pythonic.  I tryed something in between.

```
SEARCH_LIST = {
    'total_connections': (
        'cn=Total,cn=Connections,cn=Monitor', 'monitorCounter',
    ),
    'bytes_sent': (
        'cn=Bytes,cn=Statistics,cn=Monitor', 'monitorCounter',
    ),
```

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: to be honest, I don't realy understand when this can be false: 
... `and not self.reconnect():`
Where do we call this method?

l2isbad: There is no guidelines, but this is natural to have _connect, reconnect_ methods for modules which connect to something (module should survive monitored service restart, stop/start). 

>  don't really understand when this can be false ...and not self.reconnect()

```python
if not self.alive and not self.reconenct():
    return None
```

This is always `False` while `alive` flag is `True`.
Module checks `alive` flag and if it's `True` it skips `reconnect` part.
If flag is `False` it calls `reconnect`.

Change `alive` flag to `False` if there is any `ldap` error occurs. This is neede for module to survive monitored service restart, stop/start

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,203 @@
...
+                num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr,])
+                result_type, result_data = self.conn.result(num, 1)
+            except ldap.LDAPError as error:
+                self.error(error)
+
+            try:
+                if result_type == 101:
+                    val = int(result_data[0][1].values()[0][0])
+            except ValueError:
+                val = 'NaN'
```
l2isbad: just `continue`

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,203 @@
...
+
+        data = dict()
+        for key in SEARCH_LIST.keys():
+            bind = SEARCH_LIST[key][0]
+            attr = SEARCH_LIST[key][1]
+            try:
+                num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr,])
+                result_type, result_data = self.conn.result(num, 1)
+            except ldap.LDAPError as error:
+                self.error(error)
```
l2isbad: I am not a `ldap` specialist but i think the only possible reason for exception here is connection problem (service restarted, service anavailable for some reason if it's remote `ldap` instance). correct me if am wrong.

Then we need to change `alive` flag to `False` here and return `None`. module wil attempt to reconnect on next iteration.

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,114 @@
...
+     ]},
+}
+
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.conn = ldap.initialize('ldap://%s:%s' % (configuration.get('server'), configuration.get('port')))
+        self.conn.simple_bind(configuration.get('username'), configuration.get('password'))
```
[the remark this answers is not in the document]
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,114 @@
...
+from bases.FrameworkServices.SimpleService import SimpleService
+
+# default module values (can be overridden per job in `config`)
+update_every = 10
+priority = 60000
+retries = 60
+
+
+
+ORDER = ['total_connections', 'bytes_sent', 'operations', 'referrals_sent', 'entries_sent', 'ldap_operations', 'waiters']
```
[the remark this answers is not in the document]
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: changed to this:
```
try:
if result_type == 101:
    val = int(result_data[0][1].values()[0][0])
except ValueError:
    continue
```

collectors/python.d.plugin/openldap/openldap.chart.py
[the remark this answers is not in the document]
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,203 @@
...
+
+try:
+     import ldap
+     HAS_LDAP = True
+except ImportError:
+     HAS_LDAP = False
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+# default module values (can be overridden per job in `config`)
```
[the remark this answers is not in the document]
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: You are right. I am just starrting to understand the logic behind connection management by the module. Thanks

ekartsonakis: changed with 
```
            try:
                num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr, ])
                result_type, result_data = self.conn.result(num, 1)
            except ldap.LDAPError as error:
                self.alive = False
                return None
```

collectors/python.d.plugin/openldap/openldap.chart.py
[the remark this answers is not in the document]
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done 👍

collectors/python.d.plugin/openldap/openldap.chart.py
l2isbad: Perfect.

All those `[0]`... `IndexError` ? Or it's safe to index `result_data` like that?

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
l2isbad: Looks good to me

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: fixed

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: fixed

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,114 @@
+# -*- coding: utf-8 -*-
+# Description: openldap netdata python.d module
+# Author: Manolis Kartsonakis (ekartsonakis)
+# SPDX-License-Identifier: GPL-3.0+
+
+import ldap
```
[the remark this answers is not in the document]
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
l2isbad: it's not. Have you pushed commits?

l2isbad: Still not fixed

collectors/python.d.plugin/python.d.conf
```diff
@@ -95,3 +95,4 @@ unbound: no
 # varnish: yes
 # w1sensor: yes
 # web_log: yes
+openldap: no
```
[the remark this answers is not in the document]
ekartsonakis: 👍

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: Just not. There was so many opened conversations that confused me. I was closing one by one while fixing them.

ekartsonakis:
when something is marked as outdated?

l2isbad:
@ekartsonakis you almost finished it!

some minor fixes
* please change context for all charts: `monitorCounter` => `openldap.chart_name`, e.g. `openldap.total_connections`
* are you sure all values are incremental? if so units should be `unit/s`, e.g. `connections` => `connections/s` (and `Bytes` => `KB/s`)

Ferroin:
@ekartsonakis GitHub automatically marks as outdated (and closes) any conversations attached to a line of code when a commit is added to the PR that changes that line of code.  The idea is that the next commit touching a line usually fixes whatever comments relating to that line were discussing.

l2isbad:
Last but not least:
1. If remote server isn't responding `simple_bind()` is hanging for 130 seconds.

We need a `timeout` for initialized connection

```python
# timeout should be user configurable, default is 1
self.conn = ldap.initialize('ldap://servername:389')
self.conn.set_option(ldap.OPT_NETWORK_TIMEOUT, self.timeout)
```

2. Do you have a chance to test module against service restart, stop/start? So we can be sure it recovers

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,197 @@
...
+try:
+    import ldap
+    HAS_LDAP = True
+except ImportError:
+    HAS_LDAP = False
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+# default module values (can be overridden per job in `config`)
+disabled_by_default = True
```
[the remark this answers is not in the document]
l2isbad: Yes, we use `disabled_by_default = True` for cpu intensive modules or something like that. 

In general all modules are enabled by default. So please remove `disabled_by_default = True`

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,197 @@
...
+            return None
+
+        data = dict()
+        for key in SEARCH_LIST.keys():
+            bind = SEARCH_LIST[key][0]
+            attr = SEARCH_LIST[key][1]
+            try:
+                num = self.conn.search(bind, ldap.SCOPE_BASE, 'objectClass=*', [attr, ])
+                result_type, result_data = self.conn.result(num, 1)
+            except ldap.LDAPError as error:
```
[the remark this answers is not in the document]
ekartsonakis: I think I did it. Please check it out.

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -153,6 +154,7 @@ def connect(self):
         try:
             self.conn = ldap.initialize('ldap://%s:%s' % (self.server, self.port))
             self.conn.simple_bind(self.username, self.password)
+            self.conn.set_option(ldap.OPT_NETWORK_TIMEOUT, self.timeout)
```
l2isbad: please place it be before `simple_bind` call

ekartsonakis:
If my test is correct, it survived. I run `/usr/libexec/netdata/plugins.d/python.d.plugin debug openldap` while I restarted slapd (openldap server) service and I had the following logs (trimmed) :
```
{...}
2018-11-13 08:45:20: python.d DEBUG: openldap: openldap: update => [OK] (elapsed time: 38, retries left: 60)
2018-11-13 08:45:20: python.d DEBUG: openldap: openldap: sleeping for 9.94556999207 to reach frequency of 10 sec
2018-11-13 08:45:30: python.d ERROR: openldap: openldap: {'desc': "Can't contact LDAP server"}
2018-11-13 08:45:30: python.d DEBUG: openldap: openldap: get_data() returned no data
2018-11-13 08:45:30: python.d DEBUG: openldap: openldap: update => [FAILED] (elapsed time: -, retries left: 59)
2018-11-13 08:45:30: python.d DEBUG: openldap: openldap: sleeping for 9.98792290688 to reach frequency of 10 sec
BEGIN openldap.operations 19993866
SET 'completed_operations' = 49
SET 'initiated_operations' = 69
END
{...}
```

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+
+        self.server = configuration.get('server')
+        self.port = configuration.get('port')
+        self.username = configuration.get('username')
+        self.password = configuration.get('password')
+        self.timeout = configuration.get('timeout')
```
l2isbad: `configuration.get('timeout', 1)`

i suggest `1` as default

l2isbad: add global var `DEFAULT_TIMEOUT` and use it in `configuration.get`

l2isbad:
> If my test is correct, it survived.

Great!

collectors/python.d.plugin/openldap/openldap.chart.py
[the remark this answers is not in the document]
ekartsonakis: ok

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: Sorry I don't realy get you. I already added a new variable inside `openldap.conf` where I think we store all plugin "settings" variables. I set it there with value `1`. If I create a new `DEFAULT_TIMEOUT` globaly how will I use it with configuration.get. "configuration" method is loaded inside service class. Anyway I added a default value as you suggested.

Ferroin:
So, the more I think about it, the less it makes sense to me to have an auto-detection job.  Why should we provide defaults for the username/password when they're going to have to be manually configured by the user regardless?

Assuming we don't provide a default job, we shouldn't have config defaults for the username and password, and there should be a fast-path to return False from `check()` in the event that the username is undefined.

Also, I'm not 100% certain if this supports anonymous binding to query stats, or even if we should support that, but it should be documented whether it's supported or not.

collectors/python.d.plugin/openldap/openldap.conf
```diff
@@ -0,0 +1,72 @@
...
+#
+# job_name:
+#     name: myname            # the JOB's name as it will appear at the
+#                             # dashboard (by default is the job_name)
+#                             # JOBs sharing a name are mutually exclusive
+#     update_every: 1         # the JOB's data collection frequency
+#     priority: 60000         # the JOB's order on the dashboard
+#     retries: 60             # the JOB's number of restoration attempts
+#     autodetection_retry: 0  # the JOB's re-check interval in seconds
+#
```
Ferroin: There should be documentation here about what extra parameters this module takes for configuration.  See https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/ceph/ceph.conf for an example of what this should look like.

collectors/python.d.plugin/openldap/openldap.chart.py
l2isbad: i meant this


```python
DEFAULT_TIMEOUT = 1

ORDER = [
...
...
...

self.timeout = configuration.get('timeout', DEFAULT_TIMEOUT)
```

Ferroin:
@l2isbad Are we certain that won't try an anonymous bind though?  Because if it does, then we'll still generate network traffic when we arguably shouldn't.

l2isbad:
To be honest i am not sure, my `openldap` experience ~ yours.
i see that `simple_bind` allows `user` and `pass` to be `None` and i guess this is a valid option, i may be wrong, ofc!

@earthgecko since you are the only `openldap` user here can you clarify that question? 

> connection without credentials

valid or not?

Ferroin:
Connecting without credentials is technically valid dependent on the server configuration.  As a general rule, a properly secured server blocks it except for specific sub-trees, or just completely blocks it.

I don't know if it's something we want to support or not (I would lean towards no, because exposing performance data to anyone who asks for it is generally a bad idea).  If we don't, we obviously should just bail in `check()` before calling `simple_bind()` so that we're certain we're not generating any network traffic if we're not configured.  If we do, then we need to decide whether we want to try it for the case of having no configuration.

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+
+        self.server = configuration.get('server')
+        self.port = configuration.get('port')
```
l2isbad: please add `DEFAULT_SERVER` and `DEFAULT_PORT` (localhost and 389)

l2isbad:
> Connecting without credentials is technically valid dependent on the server configuration.

I think if it's a valid option we should support  it 

> we're not generating any network traffic if we're not configured.

default `server` is `localhost`

collectors/python.d.plugin/openldap/openldap.conf
```diff
@@ -0,0 +1,72 @@
...
+# ----------------------------------------------------------------------
+# AUTO-DETECTION JOBS
+
+username : "cn=monitor,dc=supperb,dc=eu"
+password : "yourpass"
```
l2isbad: please comment out `username` and `password`

Ferroin:
>> we're not generating any network traffic if we're not configured.
>
> default server is localhost

But trying to connect over the loop-back interface still comes with a nontrivial amount of overhead compared to us checking if we even want to.

l2isbad:
We have a lot of modules which auto detection jobs has no credentials by default.

ekartsonakis:
In my opinion this plugin should be manualy enable for the following reasons:

1. Not so many possibilities to allow unbinded (anonymous) requests for openldap stats.
2. Monitoring statistics for openldap is something that an admin have to enable manualy. In other words they are not exposed out of the box.
3. An extra python module is needed that has to be installed.

Btw, netdata is supposed to be (at least as I understand) as a service that exposes metrics running in localhost (primarily). Of course we give the option to change the server and set a remote one, but in that case that means that probably the user is not the admin of the LDAP server so there will be no need to gather this kind of metrics or there will be no access at that level.

Technicaly you can allow anonymous access to stats but I believe that no one does it in that way.

I already include in the readme [this link that describe the stats activation steps](https://www.openldap.org/doc/admin24/monitoringslapd.html)

Ferroin:
The first point (allowing anonymous access to stats is uncommon) is really the only one that is an issue here.  The module automatically disables itself if the required openldap Python module is not present, and we have lots of stuff that runs by default but requires setup of the monitored service to work correctly (see for example the nginx and apache modules).

collectors/python.d.plugin/openldap/openldap.conf
l2isbad: and this and we are done :tada:

collectors/python.d.plugin/openldap/openldap.chart.py
l2isbad: please do

l2isbad:
> Of course we give the option to change the server and set a remote one

I see no reason why not to do so. `python-ldap` allows it, why shouldn't we

Ferroin:
@cakrit It looks like there's some issue with the netlify checks.

ekartsonakis:
> I see no reason why not to do so. python-ldap allows it, why shouldn't we

I didn't write that we shouldn't.

collectors/python.d.plugin/python.d.conf
```diff
@@ -72,6 +72,7 @@ logind: no
 nginx_log: no
 # nsd: yes
 # ntpd: yes
+openldap: no
```
l2isbad: please change to `# openldap: yes`

The module will attempt to connect only on servers with `python-ldap` package installed.

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.conf
ekartsonakis: I think it it now ok.

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done.

collectors/python.d.plugin/openldap/openldap.conf
```diff
@@ -0,0 +1,74 @@
...
+# ----------------------------------------------------------------------
+# OPENLDAP EXTRA PARAMETERS
+
+# Set here your LDAP connection settings
+
+username : "cn=admin,dc=example,dc=com"   # The bind user with right to access monitor statistics
+password : "yourpass"                     # The password for the binded user
+#server   : 'localhost'                    # The listening address of the LDAP server
+#port     : 389                            # The listening port of the LDAP server
+#timeout  : 1                              # Seconds to timeout if no connection exists
```
l2isbad: please comment out all extra parameters (`username` and `password` too)

l2isbad:
> I didn't write that we shouldn't.

Ha, you are absolutely right, sorry

collectors/python.d.plugin/openldap/openldap.conf
ekartsonakis: ok. So finaly wilI I change the code to allow anonymous?

cakrit:
> @cakrit It looks like there's some issue with the netlify checks.

Don't worry about it. Not sure why it's failing, but I'll debug and fix it after the merge.

collectors/python.d.plugin/python.d.conf
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.conf
l2isbad: If you want to deny anonymous connection you should add

```python
if not (self.user and self.password):
    self.error('anonymous connections are not allowed, please specify user and password')
    return False
```

right after `if not HAS_LDAP: ...`  check

I am in doubt about that. Since it's technically allowed and `ldap` package allows it i think we should allow it too.

@Ferroin you decide!

collectors/python.d.plugin/python.d.conf
```diff
@@ -72,7 +72,7 @@ logind: no
 nginx_log: no
 # nsd: yes
 # ntpd: yes
-openldap: no
+openldap: yes
```
l2isbad: `openldap: yes` and `# openldap: yes` is two different things for `python.d.plugin`, please add `#`

collectors/python.d.plugin/openldap/openldap.conf
earthgecko: @l2isbad sorry I think you have the wrong user regarding your earlier question

> @earthgecko since you are the only openldap user here can you clarify that question?

I am unsure how you tagged me as the only openldap user here, honestly that is a surprise to me too, I did not even know I was using openldap :)  Sorry I cannot clarify anything here ;)

l2isbad: @earthgecko ops, i missed that somehow :sweat_smile:

collectors/python.d.plugin/python.d.conf
ekartsonakis: 👍

collectors/python.d.plugin/openldap/openldap.conf
ekartsonakis: done

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,202 @@
...
+                result_type, result_data = self.conn.result(num, 1)
+            except ldap.LDAPError as error:
+                self.error(error)
+                self.alive = False
+                return None
+
+            try:
+                if result_type == 101:
+                    val = int(result_data[0][1].values()[0][0])
+            except ValueError:
```
l2isbad: let's log the error if any, i think it will be good for debugin purposes

use `self.debug`

l2isbad:
@ekartsonakis you know what? You did it :tada: Thanks for contributing to `netdata`!

collectors/python.d.plugin/openldap/openldap.conf
Ferroin: As a general rule I would say no, but we really aren't in the business of telling people how they should secure their systems, so I think supporting it is probably fine.

collectors/python.d.plugin/openldap/openldap.chart.py
Ferroin: If we're catching a `ValueError` for any reason, we should be logging it, because it means something is misbehaving pretty seriously.  Same goes for pretty much anything except module specific errors and some specific expected cases of `OSError`.

So, yes, please make sure this gets logged if it happens.

l2isbad: This indexing worries me. I feel like we need `IndexError` here too. I am sure i have mentioned it already

Ferroin: I can't tell for certain without getting into the details of the OpenLDAP python module, but it's probably a good idea to catch it here too, provided of course that things are properly logged.

collectors/python.d.plugin/openldap/openldap.conf
ekartsonakis: Finally I aggree with @Ferroin 😀  But now maybe I will need one more day because I am out of time now 😞.

l2isbad: You need to add one more think and we are good - https://github.com/netdata/netdata/pull/4513#discussion_r233155374
!

collectors/python.d.plugin/openldap/openldap.chart.py
Ferroin: Expanding on that a bit, syntax errors are actually a sub-class of `Exception` in Python, so any typos in the code inside a try/except block that catches `Exception` will mask those.

Other things that are also exceptions that should almost never be caught except explicitly include a handful of standard exit conditions (Ctrl-C and `sys.exit()`, though those are only an issue for a bare `except` clause), a number of cases where the interpreter's internal state is not guaranteed safe (`MemoryError` and `SystemError`), and a wide variety of exceptions that can indicate logic errors in the code.

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -139,8 +141,8 @@ def __init__(self, configuration=None, name=None):
 
         self.server = configuration.get('server', DEFAULT_SERVER)
         self.port = configuration.get('port', DEFAULT_PORT)
-        self.username = configuration.get('username')
-        self.password = configuration.get('password')
+        self.username = configuration.get('username', DEFAULT_USERNAME)
+        self.password = configuration.get('password', DEFAULT_PASSWORD)
```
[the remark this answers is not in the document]
ekartsonakis: To be honest, I tried with `None` in many variations but I was too tired  to troubleshoot it when I tested. I can work more on it this afternoon:
```
2018-11-14 08:33:09: python.d DEBUG: plugin: main: module status: 'openldap' => [OK] (jobs: 1)
2018-11-14 08:33:09: python.d ERROR: openldap: openldap: check() unhandled exception: must be string, not None
2018-11-14 08:33:09: python.d ERROR: openldap: openldap: check() => [FAILED]
2018-11-14 08:33:09: python.d INFO: plugin: main: FINISHED
bash-4.2$
```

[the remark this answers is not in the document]
l2isbad: hm, please add trace flag and run plugin in debug mode again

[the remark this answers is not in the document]
ekartsonakis: Ok after some test, seems that simple bind method doesn't like 'None' values. But since we allow anonymous binds, simple bind os not necessary. So check out my last commit, and let me know if I need to change something further. It is tested and working.

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -181,20 +183,21 @@ def get_data(self):
...
+            except Exception as error:
+                self.error("Empty result. Check bind username/password. Message: ",error)
                 self.alive = False
                 return None
 
             try:
                 if result_type == 101:
                     val = int(result_data[0][1].values()[0][0])
-            except ValueError:
+            except ( ValueError, IndexError) as error:
```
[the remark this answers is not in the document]
ekartsonakis: fixed

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,204 @@
...
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+
+        self.server = configuration.get('server', DEFAULT_SERVER)
+        self.port = configuration.get('port', DEFAULT_PORT)
+        self.username = configuration.get('username', None)
+        self.password = configuration.get('password', None)
```
l2isbad: `None` is default, please remove it.

```python
d.get("missing_key") == d.get("missing_key", None)
```

collectors/python.d.plugin/openldap/openldap.chart.py
```diff
@@ -0,0 +1,204 @@
...
+            self.conn.unbind()
+            self.conn = None
+            self.alive = False
+
+    def connect(self):
+        try:
+            self.conn = ldap.initialize('ldap://%s:%s' % (self.server, self.port))
+            self.conn.set_option(ldap.OPT_NETWORK_TIMEOUT, self.timeout)
+            if self.username and self.password:
+                self.conn.simple_bind(self.username, self.password)
```
l2isbad: if `self.conn.simple_bind` doesn't work with `None, None` please add `if not (self.username and self.password)` to `check()`.

l2isbad: > Ok after some test, seems that simple bind method doesn't like 'None' values.

Can you show a traceback? (`python.d.plugin openldap debug trace`)

ekartsonakis: here it is:
```
2018-11-14 09:35:34: python.d INFO: plugin: main: Using python 2
2018-11-14 09:35:34: python.d ERROR: openldap: openldap: check() unhandled exception: must be string, not None
2018-11-14 09:35:34: python.d ERROR: openldap: openldap: Traceback (most recent call last):
  File "/usr/libexec/netdata/plugins.d/python.d.plugin", line 277, in check_job
    check_ok = bool(job.check())
  File "/usr/libexec/netdata/python.d/openldap.chart.py", line 183, in check
    return self.connect() and self.get_data()
  File "/usr/libexec/netdata/python.d/openldap.chart.py", line 165, in connect
    self.conn.simple_bind(self.username, self.password)
  File "/usr/lib64/python2.7/site-packages/ldap/ldapobject.py", line 201, in simple_bind
    return self._ldap_call(self._l.simple_bind,who,cred,RequestControlTuples(serverctrls),RequestControlTuples(clientctrls))
  File "/usr/lib64/python2.7/site-packages/ldap/ldapobject.py", line 99, in _ldap_call
    result = func(*args,**kwargs)
TypeError: must be string, not None

2018-11-14 09:35:34: python.d ERROR: openldap: openldap: check() => [FAILED]
2018-11-14 09:35:34: python.d ERROR: openldap: openldap: Traceback (most recent call last):
  File "/usr/libexec/netdata/plugins.d/python.d.plugin", line 19, in <module>
    from time import monotonic as time
ImportError: cannot import name monotonic

2018-11-14 09:35:34: python.d INFO: plugin: main: FINISHED
```

ekartsonakis: > if self.conn.simple_bind doesn't work with None, None please add if not (self.username and self.password) to check().

But we don't even use `conn.simple_bind` in case username and password are not provided.

collectors/python.d.plugin/openldap/openldap.chart.py
ekartsonakis: done 👍

collectors/python.d.plugin/openldap/openldap.chart.py
l2isbad: Wait, we don't need `conn.simple_bind` for anonymous connection?

l2isbad: >  num = self.conn.search(dn, ldap.SCOPE_BASE, 'objectClass=*', [attr, ])
result_type, result_data = self.conn.result(num, 1)

this ^^ works without `simple_bind` ?

ekartsonakis: If I add this check, in case there is an anonymous bind the check will never allow the connection to be established.  Anonymous bind doesn't even need a simple_bind method to be called.

That's why we use simple_bind ONLY if username and password are provided and are NOT None:

```
if self.username and self.password:
                self.conn.simple_bind(self.username, self.password)
```

Is there something wrong in my thinking?

l2isbad: > Anonymous bind doesn't even need a simple_bind method to be called.

I was not aware of this. If so, then yes, your code is OK.

l2isbad:
@ekartsonakis perfect, enjoyed working on this with you :+1:

ekartsonakis:
The joy is all mine! Thank you so much for everything you taught me!!  Let me know if you need any beginner's hands in this or another project!

Issue #4487, opened 2018-10-25
netdata allocates 170MB memory after startup (without the database)


74cmonty:
##### Question summary

I have configured a local netdata registry server.
Everything works as expected.
I was expecting that this server would allocate small amount of memory.
Therefore I put it on a container (LXC) with 200MB.

After starting up this container the memory allocation shows +170MB.
In my opinion this is too much.
In order to reduce the memory allocation I set paramter `history = 60`, but this was no help.

###### OS / Environment

Alpine Linux 3.8

###### Expected results

I have seen in another LXC that netdata registry server allocates only 50MB.
This is a size I would expect.

ktsaou:
The memory allocated by the registry is tracked with these charts:

![image](https://user-images.githubusercontent.com/2662304/47536078-0a6ff100-d8c7-11e8-8a9f-e67b46823b1d.png)

As you can see, the global netdata registry has allocated 350MB, but it is huge: 1M persons and 500k machines.

Could you please your versions of these charts?

ktsaou:
Please also post this chart:

![image](https://user-images.githubusercontent.com/2662304/47536501-0cd34a80-d8c9-11e8-9d28-92e155f38e05.png)

74cmonty:
Hi,
I have taken the screenshots below directly from registry node.
![2018-10-26_11-55-49](https://user-images.githubusercontent.com/12523591/47559630-4a11f980-d916-11e8-9f2a-8e507d796a49.png)
![2018-10-26_11-54-05](https://user-images.githubusercontent.com/12523591/47559650-54cc8e80-d916-11e8-84d5-59745c51a0e8.png)

ktsaou:
ok, so it is not the registry.

Could you please post the legend at the bottom of the menu? Like this:

![image](https://user-images.githubusercontent.com/2662304/47560098-fa383000-d91f-11e8-877e-b74bfbf020b6.png)

74cmonty:
Sure.
![2018-10-26_12-10-17](https://user-images.githubusercontent.com/12523591/47560351-40899100-d918-11e8-8676-f709cb882ce2.png)

ktsaou:
ok, it is not the db either.

Are you streaming metrics from other servers to it?
Are you pushing statsd metrics to it?

74cmonty:
Nope.
Actually I don't know how to configure netdata to stream or push data.

This is my config:
````
# global netdata configuration

[global]
        history = 60
[web]
[...]

Issue #4518, opened 2018-10-29
Disk full (inodes) due to netdata


alekc:
##### Bug report summary
Honestly not sure if it's really a bug, but still maybe it's worth checking out. 

On one of my machines I ran out of space due to hard disk filling up with unrotated log file, once that file has been deleted I noticed that there still an issue with inodes. 

```
df -i
Filesystem      Inodes  IUsed   IFree IUse% Mounted on
/dev/sda1       940240 930220   10020   99% /
udev            634681    321  634360    1% /dev
tmpfs           636892    525  636367    1% /run
tmpfs           636892      1  636891    1% /dev/shm
tmpfs           636892      4  636888    1% /run/lock
tmpfs           636892     13  636879    1% /sys/fs/cgroup
/dev/sdb1      1310720 159172 1151548   13% /var/www
tmpfs           636892      4  636888    1% /run/user/999
tmpfs           636892      4  636888    1% /run/user/0
```
Netdata uses 687k inodes 
```
root@lamp-dev-2:/var/cache# find . -printf "%h\n" | cut -d/ -f-2 | sort | uniq -c | sort -rn
 687884 ./netdata
   1342 ./apt
     52 ./man
     22 ./fontconfig
     11 ./dictionaries-common
     10 .
      8 ./pbuilder
      5 ./oracle-jdk8-installer
      5 ./debconf
      1 ./ldconfig
root@lamp-dev-2:/var/cache# 
```
Going deeper
```
  10506 ./system.interrupts
  10493 ./cpu.cpu1_interrupts
  10468 ./cpu.cpu0_interrupts
   6693 ./apps.cpu
   6654 ./apps.mem
   5503 ./apps.vmem
   4300 ./apps.threads
   4255 ./apps.processes
   4217 ./apps.cpu_user
   4210 ./apps.swap
   4210 ./apps.cpu_system
   4204 ./apps.pwrites
   4181 ./apps.minor_faults
   4175 ./apps.preads
   4175 ./apps.major_faults
   4163 ./apps.lwrites
   4162 ./apps.sockets
   4162 ./apps.pipes
   4162 ./apps.files
   4150 ./netdata.plugin_proc_modules
   4148 ./apps.lreads
   3919 ./ipv4.icmpmsg
   3840 ./cgroup_temp_web_1.throttle_serviced_ops
   3763 ./groups.swap
   3763 ./groups.pwrites
   3763 ./groups.preads
   3763 ./groups.minor_faults
[...]

PR #4519, opened 2018-10-29
Fix interrupts

ktsaou:
##### Summary

Netdata was enabling per core interrupts that are always zero.
This PR eliminates them. Also for softirqs.

Related to #4487 	

##### Additional Information

```paste below

```


Ferroin:
May also be related to #4518.

ktsaou:
> May also be related to #4518.

This definitely seems related.

Issue #4285, opened 2018-09-25
Add SCSI device support to smartd_log.chart.py


ScrumpyJack:
man smartd says:
 [For ATA devices] "At each check cycle attributes are logged as a line of semicolon  separated  triplets  of  the  form "attribute-ID;attribute-norm-value;attribute-raw-value;". 
For SCSI devices error counters and temperature [are] recorded in the form "counter-name;counter-value;"

The plugin relies on the attribute-ID, but the smartd log files are only populated with attribute-ID for ATA devices (weirdly) 

Would it be possible to add support for SCSI disk smartd log files into smartd_log.chart.py?

l2isbad:
> For SCSI devices error counters and temperature [are] recorded in the form "counter-name;counter-value;"

please post here some lines from SCSI smartd log file

ScrumpyJack:
uploaded the csv output for a scsi drive here

https://b.ilet.to/uoluC

l2isbad:
Does all those counters has corresponding names/ids?

Which counters should we chart?

l2isbad:
@Ferroin any ideas?

Ferroin:
From a cursory look at the attributes listed, it seems to me that we should be tracking the following by default:

* `temperature`
* `*-total-unc-errors`: Appears to count uncorrected errors, which should be reported.
* `*-total-err-corrected`: Appears to count corrected errors, which should be reported.

The issue here is that the only one which maps directly to an ATA SMART attribute is temperature.  There's not any ID because SCSI sends the names directly as part of the protocol, instead of ATA, which sends the ID but not the name (and you have to map the ID to the name in software).  The reported values are functionally raw values, but they're in natural units for the counter type (degrees celsius for the temperature, individual errors for the error counters, etc).

l2isbad:
@ScrumpyJack 
what do you think?

ScrumpyJack:
looks good.
something clever could be done with the ratio
*-gb-processed/*-corr-algorithm-invocations
to get predictive failure, but knowing when that ratio is significant is another story.

l2isbad:
Hm, i don't know how to organize dashboard menu..

current version:
![screenshot_20180929_113456](https://user-images.githubusercontent.com/22274335/46240010-fb874300-c3db-11e8-84c4-7e4921c5afd4.png)

do not distinguish between ata and scsi and just add this as menu items?
```python
'read-total-err-corrected',
'read-total-unc-errors',
'write-total-err-corrected',
[...]

PR #4523, opened 2018-10-30
smartd_log refactor plus SCSI support

l2isbad:
##### Summary

attempt to Fix: #4285

- scsi support
- change grouping/processing attributes (https://github.com/netdata/netdata/issues/4285#issuecomment-425654443)

##### Component Name

:snake: `smartd_log` :snake: 

##### Additional Information
todo:
- [x] attribute value preprocessing (like https://github.com/netdata/netdata/issues/4285#issuecomment-428556760)
- [x] readme update
- [ ] alarms
- [x] hadle log rotation

problems:
- netdata doesn't suppport deletion of dimension deletion in runtime

### DIFF
--- a/collectors/python.d.plugin/smartd_log/README.md
+++ b/collectors/python.d.plugin/smartd_log/README.md
@@ -23,52 +23,57 @@ It produces following charts for SCSI devices:
 
 
 For ATA devices:
+1. **Read Error Rate**
 
-1. **Soft Read Error Rate**
+2. **Seek Error Rate**
 
-2. **SATA Interface Downshift**
+3. **Soft Read Error Rate**
 
-3. **UDMA CRC Error Count**
+4. **Write Error Rate**
 
-4. **Throughput Performance**
+5. **SATA Interface Downshift**
 
-5. **Seek Time Performance**
+6. **UDMA CRC Error Count**
 
-6. **Start/Stop Count**
+7. **Throughput Performance**
 
-7. **Power-On Hours Count**
+8. **Seek Time Performance**
 
-8. **Power Cycle Count**
+9. **Start/Stop Count**
 
-9. **Unexpected Power Loss**
+10. **Power-On Hours Count**
 
-10. **Spin-Up Time**
+11. **Power Cycle Count**
 
-11. **Spin-up Retries**
+12. **Unexpected Power Loss**
 
-12. **Calibration Retries**
+13. **Spin-Up Time**
 
-13. **Temperature**
+14. **Spin-up Retries**
 
-14. **Reallocated Sectors Count**
+15. **Calibration Retries**
 
-15. **Reserved Block Count**
+16. **Temperature**
 
-16. **Program Fail Count**
+17. **Reallocated Sectors Count**
 
-17. **Erase Fail Count**
+18. **Reserved Block Count**
 
-18. **Wear Leveller Worst Case Erase Count**
+19. **Program Fail Count**
 
-19. **Unused Reserved NAND Blocks**
+20. **Erase Fail Count**
 
-20. **Reallocation Event Count**
+21. **Wear Leveller Worst Case Erase Count**
 
-21. **Current Pending Sector Count**
+22. **Unused Reserved NAND Blocks**
 
-22. **Offline Uncorrectable Sector Count**
+23. **Reallocation Event Count**
 
-23. **Percent Lifetime Used**
+24. **Current Pending Sector Count**
+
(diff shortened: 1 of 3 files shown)


Ferroin:
Overall LGTM, though I haven't had a chance yet to look in-depth at it.

@l2isbad I just posted a Python snippet in https://github.com/netdata/netdata/issues/4285#issuecomment-434668314 that should be a reasonable approximation of the logic used by smartctl to decode the raw temperature values.  Not sure if you want to use it or not, but it's there if you want it.

l2isbad:
> temperature, minimum, maximum, and over-temperature count

Why do we need all this? What should we do with minimum and maximum? how to chart it?

Ferroin:
I don't think they need to be charted, but the data is there, so I see no reason not to expose it in some way.  If we don't want that data at all though, then we don't need to parse the raw values either.

l2isbad:
It's not `want/don't want`. I don't know how to use it, `min/max` is not important data for me, but i may be wrong.

My proposal is to use `194` normalized for temperature. We can rewrite it later if the need arises. I will add `NOTE` and your code snippet as comments.

Ferroin:
Sounds good to me.  I can see uses for the data, but can't think of any way that it makes sense to expose it through Netdata (chart variables come to mind, but that only makes sense if we do attributes as dimensions or one attribute per chart).

l2isbad:
@Ferroin It seems that some attrs have to be decoded depending on disk type/vendor. [Example](https://github.com/netdata/netdata/issues/4285#issuecomment-434531577) `Power On Hours`. 

So if you have the opportunity to test this PR in your env please do

Ferroin:
@l2isbad I plan to do so, but it may be this weekend before I get the chance to do it properly.  I don't know how interesting the results may be though, I don't have any particularly exotic disks.

Ferroin:
Error rates are more useful than total error counts, both for creating alarms, and for actually seeing device health.  Put simply, on a healthy disk, they should not change almost ever.  On a disk that's starting to fail, they will be consistently high.  In contrast, for total counts, a non-zero value just tells you an error happened at some point, not that they are continuing to happen. 

Given this, I think incremental is correct here, provided that the charts are annotated in such a way that it's obvious that they're displaying the error rate, not the total error count.

l2isbad:
> On a disk that's starting to fail, they will be consistently high. 

If the behaviour is 
> On a disk that's starting to fail, they will be consistently high

 then yes, `incremental` is OK.

Ferroin:
> If the behaviour is
>
>> On a disk that's starting to fail, they will be consistently high
>
> then yes, incremental is OK.

Put simply, if the disk failure isn't indicated by high error rates, then it won't realistically be indicated by high error counts either.

l2isbad:
@ScrumpyJack 
please test, this version supports SCSI devices

l2isbad:
ok, i am waiting for some feedback from you @ScrumpyJack

@Ferroin please review when you have time

Ferroin:
LGTM so far, though I still haven't had a chance to test it (it looks like I might get a chance this evening, but if not I will get it tested this weekend).  Are the two remaining items in the 'todo' list planned for this PR, or is the intent to work on them longer term?

l2isbad:
btw @Ferroin you have time please look at 
https://github.com/netdata/netdata/issues/4285#issuecomment-425654443
one more time

I added only attrs with `family`. We might miss something.

Ferroin:
> About the alarms. I hope for your help. Feel free to add them (you can add commits to this PR)

I'll try and look into getting some written up.  It's going to be interesting figuring out what makes sense for some of them though.  There's a lot of people who run unhealthy disks until they die instead of replacing them early, and it would be good not to annoy them, but if the thresholds are too high, the alarms just won't be useful.

l2isbad:
>  There's a lot of people who run unhealthy disks until they die

Haha, it's me

Ferroin:
Regarding attributes:, ATA attributes 1, 7, and 206 might be worth providing graphs for.  While their raw values are vendor-specific in format, their normalized values are reasonably consistent across vendors, they start at 100, and decrease towards 0 as the error rates go up.

l2isbad:
>  ATA attributes 1, 7, and 206

Please add missing values to the table and i will add right after that

Ferroin:
I've updated the table.

l2isbad:
@Ferroin We can add alarms in a separate PR. Please review/approve.

Ferroin:
@l2isbad Sorry about the delayed review, I meant to review this last week but got sidetracked by other things.

PR #4537, opened 2018-10-31
Minor refactoring

TChatzigiannakis:
- Extracted repeated pattern (netdata exit checks) into a common macro in various plugins.
- Added diagnostic message to a corner-case.

### DIFF
--- a/daemon/common.h
+++ b/daemon/common.h
@@ -26,6 +26,11 @@
 
 #define config_generate(buffer, only_changed) appconfig_generate(&netdata_config, buffer, only_changed)
 
+// ----------------------------------------------------------------------------
+// shortcuts for common patterns
+#define yield() if(unlikely(netdata_exit)) break
+
+
 
 // ----------------------------------------------------------------------------
 // netdata include files
--- a/daemon/common.h
+++ b/daemon/common.h
@@ -28,7 +28,10 @@
 
 // ----------------------------------------------------------------------------
 // shortcuts for common patterns
-#define yield() if(unlikely(netdata_exit)) break
+#define yield() if(unlikely(netdata_exit))  \
+                    break;                  \
+                else                        \
+                    ((void)0)                          
 
 
 
--- a/daemon/common.h
+++ b/daemon/common.h
@@ -28,10 +28,10 @@
 
 // ----------------------------------------------------------------------------
 // shortcuts for common patterns
-#define yield() if(unlikely(netdata_exit))  \
-                    break;                  \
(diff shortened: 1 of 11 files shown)


daemon/common.h
```diff
@@ -26,6 +26,11 @@
 
 #define config_generate(buffer, only_changed) appconfig_generate(&netdata_config, buffer, only_changed)
 
+// ----------------------------------------------------------------------------
+// shortcuts for common patterns
+#define yield() if(unlikely(netdata_exit)) break
```
ktsaou: I think this will introduce a nasty issue.

Example:

```c
if(x == 0)
  yeld();
else
  something_else();
```

When will `something_else()` run?

I know this is not used like that today, but then all contributors will have to remember that `yield()` is a special case, and cannot be used like all other functions.

ktsaou: The generally acceptable way for doing things like this, is:

```c
#define yield() do { if(unlikely(netdata_exit)) break; } while(0)
```

The above does not have the problem, but `break`, breaks that `while()` loop, not the indented one.

Another way would be to define it like that:

```c
#define IF_WE_HAVE_TO_EXIT_BREAK_THE_LOOP if(unlikely(netdata_exit)) break
```

This would be better, because people would understand this is a special macro.
Generally however, I don't think it will add much value compared to the existing.

gmosx:
I think `yield` is a bad name, it reminds me of coroutines.

daemon/common.h
TChatzigiannakis: > When will `something_else()` run?

You're right. This can be amended by making this macro a full `if`-`else` block, considering that the `else` belongs to the closest `if`.

gmosx: > I think this will introduce a nasty issue.
> ...
> When will `something_else()` run?

This is one important reason to *not* skip curly braces for one-liners. Additionally, the braces make the code much easier to understand.

ktsaou:
@TChatzigiannakis thank you!

However, I don't want to merge this.

The key reasons are:

1. You use `break` in a macro you mask as a function. This significantly lowers readability.

2. When we debug code, we frequently search for the return points of functions or the break points of loops. The name you have chosen has the word `exit` in it, which indicates an abnormal exit, not a return, neither a loop break.

So, although your PR is trivial, I think it significantly reduces the readability of the code. I will be confused myself when I will try to debug an issue in these functions.

So, since this PR does not actually add a feature or something important for netdata, I think the best option is not to merge it.

Thank you for your understanding!

Issue #4533, opened 2018-10-31
incremental chart algorithm doesn't handle counter wrap properly


Ehekatl:
##### Bug report summary

I found this bug while using node.d snmp plugin, if I use a Counter32 value, the maximum length of the 32bit counter will be full in very short interval, and it reset to 0 after that, with current implementation this will produce 0 value on the graph. See:

![screen shot 2018-10-31 at 3 23 07 pm](https://user-images.githubusercontent.com/7362589/47784273-0b2dda80-dd40-11e8-939b-2bf47ca5c048.png)

##### Component Name

netdata core

##### Steps To Reproduce

any non-monolithically increased data on a chart with incremental algorithm should able to reproduce this

##### Expected behavior

I believe the correct implementation of incremental algorithm should drop the value if it's not monolithically increase (current value < previous value) and re-count on next data point.

ktsaou:
I have applied a fix with PR #4538 
Can you test it?

ktsaou:
Are you sure they are not right?
The chart is not based at 0.
It runs 700 - 900 Mbps

What is `update_every` for this? I will add a unittest to make sure netdata does this correct.

ktsaou:
You can stop netdata and run the `node.d.plugin 1 debug snmp` by hand and post here 10-20 iterations. I'll use that data to add a unit test, and I'll also verify netdata does it right.

Ehekatl:
@ktsaou ah...sorry for the confusion, I take a close look and found the data and graph on netdata slave is good, what I really want to present is this graph:

![screen shot 2018-11-01 at 7 07 38 pm](https://user-images.githubusercontent.com/7362589/47848940-eeaaa480-de0a-11e8-9a7e-a20c26f678b9.png)

It's from our grafana (by netdata slave -> netdata stream master -> prometheus), you can see the curve it's really wrong, so I take the screenshot from the netdata slave, without really check the axe value. 

I think the problem here is because I only upgrade  netdata slave but the query calculation was happen on netdata stream master...

Let me upgrade and see

ktsaou:
merged the fix.
Thank you for reporting this!

PR #4538, opened 2018-11-01
incremental overflows should not show zeros values

ktsaou:
##### Summary

Incremental values in netdata presented a zero point when overflown.
This caused issues in a few cases, like the one presented in #4533 .

The problem is that netdata does not know the size of the counter (8bit, 16bit, 32bit, 64bit).

This PR detects the size of the overflown counter, by maintaining the max value shown for the counter. Based on the max value seen in the past, it assumes the counter will either be 8, 16, 32, or 64 bit.

---

This PR also uses the last unused bit of the storage number format, to support x100 multipliers. This makes netdata support numbers above 64 bit.

Unit tests have been added to verify the above work properly.

##### Component Name

core/database

##### Additional Information

```paste below

```

### DIFF
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -239,6 +239,7 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
     rd->last_calculated_value = 0;
     rd->collected_value = 0;
     rd->last_collected_value = 0;
+    rd->collected_value_max = 0;
     rd->collected_volume = 0;
     rd->stored_volume = 0;
     rd->last_stored_value = 0;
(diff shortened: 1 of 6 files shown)


database/rrdset.c
```diff
@@ -1381,13 +1381,29 @@ void rrdset_done(RRDSET *st) {
                     if(!(rrddim_flag_check(rd, RRDDIM_FLAG_DONT_DETECT_RESETS_OR_OVERFLOWS)))
                         storage_flags = SN_EXISTS_RESET;
 
-                    rd->last_collected_value = rd->collected_value;
-                }
+                    uint64_t last = (uint64_t)rd->last_collected_value;
+                    uint64_t new = (uint64_t)rd->collected_value;
+                    uint64_t max = (new > last) ? new : last;
```
vlvkobal: It will be better to store the maximum value in the database. The probability of changing a counter type on the fly is much less than the probability of sudden spikes.

ktsaou: You are right.
Once we do this, the only way to change it would be to remove the dimension from disk.

ktsaou: ok, did it per session. So, every time netdata starts, this is reset.

ktsaou:
@vlvkobal 	please review again...

database/rrddim.c
```diff
@@ -380,6 +381,9 @@ inline collected_number rrddim_set_by_pointer(RRDSET *st, RRDDIM *rd, collected_
 
     rd->collections_counter++;
 
+    collected_number v = (value >= 0) ? value : -value;
+    if(unlikely(v > rd->collected_value_max)) rd->collected_value_max = v;
```
vlvkobal: Do we need to update the maximum value for every data collection? I'd do it only for overflows.

ktsaou: The expensive part here is the condition, not the update, since the update will only happen when the value is the max.

To check for overflow, the condition would be more complex at that point. It would require to take the `abs()` of the `last_collected_value` too and then execute a very similar condition.

Also, keeping in the structure the `collected_value_overflow_max` is somewhat niche. The simple max could have more uses in the future. This is why I chose to keep it on all metrics, of all types (not just `incremental` ones).

PR #4546, opened 2018-11-02
add tor python module

l2isbad:
##### Summary

Fix: #2908

##### Component Name

:snake:  tor module :snake: 

Module connects to tor control port to collect traffic statistics.

**Requirements:**
* `tor` program
* `stem` python package

It produces only one chart:

1. **Traffic**
 * read
 * write

##### Additional Information

@FedericoCeratto i used part of your code, are you ok with that?

@candrews it seems you are interested in this module, please test

### DIFF
--- a/collectors/python.d.plugin/tor/README.md
+++ b/collectors/python.d.plugin/tor/README.md
@@ -1,0 +1,48 @@
+# tor
+
+Module connects to tor control port to collect traffic statistics.
+
+**Requirements:**
+* `tor` program
+* `stem` python package
+
+It produces only one chart:
+
+1. **Traffic**
+ * read
+ * write
+
+### configuration
+
+Needs `control_port` and `password`
+
+Here is an example for local server:
+
+```yaml
+update_every : 1
+priority     : 60000
+
+local_tcp:
+ name: 'local'
+ control_port: 9051
+ password: 'password'
+
+local_socket:
+ name: 'local'
+ control_port: '/var/run/tor/control'
+ password: 'password'
+```
+
+### prerequisite
+
+Add to `/etc/tor/torrc`:
+
+```
+ControlPort 9051
+```
+
+For more options please read the manual.
+
+Without configuration, module attempts to connect to `127.0.0.1:9051`.
+
+---
--- a/collectors/python.d.plugin/tor/README.md
+++ b/collectors/python.d.plugin/tor/README.md
@@ -14,7 +14,7 @@ It produces only one chart:
 
 ### configuration
 
-Needs `control_port` and `password`
+Needs only `control_port`
 
 Here is an example for local server:
 
@@ -25,12 +25,10 @@ priority     : 60000
 local_tcp:
  name: 'local'
  control_port: 9051
- password: 'password'
 
 local_socket:
  name: 'local'
  control_port: '/var/run/tor/control'
- password: 'password'
 ```
 
 ### prerequisite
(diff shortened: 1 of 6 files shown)


collectors/python.d.plugin/tor/tor.chart.py
```diff
@@ -0,0 +1,108 @@
...
+            ['write', 'write', 'incremental', 1, -1024],
+        ]
+    }
+}
+
+
+class Service(SimpleService):
+    """Provide netdata service for Tor"""
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
```
Ferroin: Personally, I would prefer to see `super()` used here instead of explicitly calling out the parent class.  All of the framework services are new-style classes on Python 2.7 (they inherit from `SimpleService`, which inherits from `object`), so we can completely eliminate the explicit reference to the parent class with `super()`

l2isbad: ok

collectors/python.d.plugin/tor/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+It produces only one chart:
+
+1. **Traffic**
+ * read
+ * write
+
+### configuration
+
+Needs `control_port` and `password`
```
candrews: `password` isn't required; tor doesn't require a password be used to access the control port.

collectors/python.d.plugin/tor/tor.conf
```diff
@@ -0,0 +1,81 @@
...
+#
+#     control_port: 'port'    # tor control port
+#     password: 'password'    # tor control password
+#
+# ----------------------------------------------------------------------
+# AUTO-DETECTION JOBS
+# only one of them will run (they have the same name)
+#
+# local_tcp:
+#  name: 'local'
```
candrews: I think these 2 auto-detection jobs should not be commented out

collectors/python.d.plugin/tor/tor.conf
```diff
@@ -0,0 +1,81 @@
...
+#  control_port: 9051
+#  password: 'password'
+#
+# local_socket:
+#  name: 'local'
+#  control_port: '/var/run/tor/control'
+#  password: 'password'
```
candrews: Default should be not to not use a password

collectors/python.d.plugin/tor/tor.conf
```diff
@@ -0,0 +1,81 @@
...
+#     password: 'password'    # tor control password
+#
+# ----------------------------------------------------------------------
+# AUTO-DETECTION JOBS
+# only one of them will run (they have the same name)
+#
+# local_tcp:
+#  name: 'local'
+#  control_port: 9051
+#  password: 'password'
```
candrews: Default should be not to not use a password

candrews:
Seems to be working, `/usr/libexec/netdata/plugins.d/python.d.plugin 1 debug trace tor` shows data from tor.

collectors/python.d.plugin/tor/tor.conf
l2isbad: Hm, i think we need not socket job.

just tested

```
socket dir needs to exist, and to be accessible only by the user account that is running Tor
```
so there is no way `netdata` user can read it

l2isbad: About the socket.

just tested and got

```
socket dir needs to exist, and to be accessible only by the user account that is running Tor
```

so there is no way `netdata` user can read it

l2isbad:
@Ferroin please review/approve

Issue #4541, opened 2018-11-01
move `/proc` and `/sys` python modules to `proc` plugin


ktsaou:
##### Feature idea summary

`python.d.plugin` should not be used for modules that collect info from `/sys` and `/proc`. These should be handled by `/proc` plugin.

The following modules can be moved:

- [x] https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/cpufreq
- [x] https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/cpuidle
- [x] https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/mdstat
- [x] https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/linux_power_supply

it is very important to move the first 2 (cpufreq and cpuidle), since these are commonly found in most installations. These will help python do less work.

Ferroin:
Regarding the cpufreq module, if you tag me for review on it, I'd be more than happy to help with testing on some non-x86 systems.  I've got a couple of ARM, PPC, and SPARC systems I can check it against (all of which are working correctly with the current python.d code).

Regarding the Linux PSU module, please make a point to tag me for review on that one so I don't miss it.  I'm actively using the python.d module on a number of systems, so I have a significant interest in making sure it keeps working.  I can also answer any questions there might be about the API it's using.

l2isbad:
I am planning to implement #4010
Should i do it? Or it will be added to C version?

ktsaou:
> I am planning to implement #4010
> Should i do it? Or it will be added to C version?

@vlvkobal 	can you do these?

vlvkobal:
Yes, I can. I Assigned myself to #4010.

vlvkobal:
Done

PR #4562, opened 2018-11-05
Move cpufreq python module to proc plugin

vlvkobal:
##### Summary
`cpufreq` python module reads data from files:
`/sys/devices/system/cpu/cpu*/cpufreq/scaling_cur_freq`
`/sys/devices//system/cpu/cpu*/cpufreq/stats/time_in_state`

Partly implements #4541

##### Component Name
`cpufreq` python module
`proc` plugin

##### Additional Information
`scaling_cur_freq` was implemented a year ago, but was disabled by default. `time_in_state` provides more accurate statistics and should be chosen for data acquisition if available.

### DIFF
--- a/collectors/proc.plugin/proc_stat.c
+++ b/collectors/proc.plugin/proc_stat.c
@@ -8,10 +8,20 @@ struct per_core_single_number_file {
     unsigned char found:1;
     const char *filename;
     int fd;
-    procfile *ff; //TODO: Separate structure for time_in_state?
     collected_number value;
-    //TODO: last_ticks for every frequency. Separate structure for time_in_state?
     RRDDIM *rd;
+};
+
+struct last_ticks {
+    collected_number frequency; // TODO: Do we need to check frequency?
+    collected_number ticks;
+};
+
+struct per_core_time_in_state_file {
+    const char *filename;
+    procfile *ff;
+    size_t last_ticks_len;
+    struct last_ticks *last_ticks;
 };
 
 #define CORE_THROTTLE_COUNT_INDEX    0
@@ -35,6 +45,8 @@ struct cpu_chart {
     RRDDIM *rd_guest_nice;
 
     struct per_core_single_number_file files[PER_CORE_FILES];
+
+    struct per_core_time_in_state_file time_in_state_files[PER_CORE_FILES]; // TODO: Make it a separate structure?
 };
 
 static int keep_per_core_fds_open = CONFIG_BOOLEAN_YES;
@@ -89,7 +101,6 @@ static int read_per_core_files(struct cpu_chart *all_cpu_charts, size_t len, siz
         f->found = 1;
 
         f->value = str2ll(buf, NULL);
-        // info("read '%s', parsed as " COLLECTED_NUMBER_FORMAT, buf, f->value);
         if(likely(f->value != 0))
             files_nonzero++;
     }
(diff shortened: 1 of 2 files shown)


Ferroin:
Cursory testing seems to indicate this works as intended.  I'll spin up some of the alternate architecture systems I've got to test it more thoroughly over the weekend.

A couple of comments that are relatively independent of testing though:

* It would be nice to have a log message indicating which mechanism is being used.  This is of course mostly of use for debugging, but it's still generally useful to know.
* I did not explicitly check this because I don't have a system configured such that I can test it, but it doesn't look to me like this will fall back to the regular files if the `time_in_state` files suddenly disappear.  That ideally should be handled, because the kernel component that provides that interface can be built as a module and unloaded without warning.
* The fact that we can now use the CPUFreq stats interface to get more accurate data should probably be documented outside of the code itself.

ktsaou:
@vlvkobal 	is this ready for review? I see wip in the title, so I understand you have more to add. Right?

vlvkobal:
@ktsaou It's ready. The changes I might make are only some messages/documentation.

Ferroin:
> Like in #4104, the question is what to do with cpufreq python module? It would be good to delete python code and use cpu.cpufreq chart id in proc plugin.

I would suggest the following approach:

1. Merge this, with it disabled by default.  Make a prominent note in the documentation that this will be used by default starting not in the next release (1.12), but in the release _after_ that.
2. Merge something so that this can tell if the Python module is enabled or not, which should ideally be reusable for other cases like this.
3. After the next release (1.12) hits, merge a change to the configuration so that this runs by default, and the Python plugin does not.
4. After at least one release of this being the default, remove the python.d module.  By this time, we _should_ have things working correctly so that we can uninstall old files.

l2isbad:
disabling

https://github.com/netdata/netdata/blob/87735fa406c86adb1e72ae0dd992c4b5c8279ba7/collectors/python.d.plugin/python.d.plugin.in#L59

ktsaou:
I would prefer the @l2isbad 	way, so disable the old ones immediately and make this the primary. This will allow users to get the benefit immediately, given that we have tested it properly.

@Ferroin 	the procedure you suggest is super and I totally agree for conflicting configuration changes, or for changes that change the default behavior of netdata (like making the netdata daemon not fork by default). When we can provide an automated migration to new features or even configuration changes, I think should prefer the immediate way. I have used this approach many times in the past. Check this for example that migrates old netdata.conf directives to new ones : https://github.com/netdata/netdata/blob/0f8b440c39d61a64e6c700199f7e22153f22b416/daemon/main.c#L369-L419

So, @vlvkobal 	please add the obsolete python modules to the variable suggested by @l2isbad 	and let's merge this.

ktsaou:
> add the obsolete python modules to the variable suggested by @l2isbad and let's merge this.

of course, except if someone sees a reason not to do it.

Ferroin:
@ktsaou I had actually forgotten that the python.d.plugin code had a way to blacklist modules, had I remembered that, I would have actually suggested the exact same thing that @l2isbad did.

collectors/proc.plugin/proc_stat.c
```diff
@@ -557,7 +694,7 @@ int do_proc_stat(int update_every, usec_t dt) {
                 else
                     rrdset_next(st_scaling_cur_freq);
 
-                chart_per_core_files(&all_cpu_charts[1], all_cpu_charts_size - 1, SCALING_CUR_FREQ_INDEX, st_scaling_cur_freq, 1, 1000, RRD_ALGORITHM_ABSOLUTE);
+                chart_per_core_files(&all_cpu_charts[1], all_cpu_charts_size - 1, CPU_FREQ_INDEX, st_scaling_cur_freq, 1, 1000, RRD_ALGORITHM_ABSOLUTE);
```
ktsaou: netdata has a nice feature in for calculating things like this:

```
RRD_ALGORITHM_PCENT_OVER_DIFF_TOTAL
```

Using this algorithm, you can pass the incremental values collected as-is (not need to find the delta), and netdata will calculate the incremental sum of all dimensions and then the percentage of each dimension. We use this for all CPU charts that report ticks.

ktsaou:
@vlvkobal 	good job!
I have added a few suggestions to optimize it further.

collectors/proc.plugin/proc_stat.c
```diff
@@ -33,6 +47,8 @@ struct cpu_chart {
     RRDDIM *rd_guest_nice;
 
     struct per_core_single_number_file files[PER_CORE_FILES];
+
+    struct per_core_time_in_state_file time_in_state_files[PER_CORE_FILES];
```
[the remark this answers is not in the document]
vlvkobal: Thank you! I wouldn't notice it!

collectors/proc.plugin/proc_stat.c
vlvkobal: We use ticks only for calculating the average frequency in the case of `time_in_state` is available. So, the algorithm is the correct one.

PR #4568, opened 2018-11-06
Read all data from the socket when the data is larger than 4096 bytes.

j0nnyw:
##### Summary

The current check does not handle the scenario when attempting to read more than 4096 bytes from a socket.

##### Component Name

SocketService.py

##### Additional Information

```paste below

```

### DIFF
--- a/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService.py
+++ b/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService.py
@@ -217,7 +217,7 @@ class SocketService(SimpleService):
 
             self.debug('received data')
             data += buf.decode('utf-8', 'ignore') if not raw else buf
-            if self._check_raw_data(data):
+            if len(buf) < 4096:
                 break
 
         self.debug('final response: {0}'.format(data))


l2isbad:
Hi @j0nnyw 

`_check_raw_data` is needed, it's overwritten in different modules, like

https://github.com/netdata/netdata/blob/0840d52e7147902d7dfd417ca33a415c3da9a5c4/collectors/python.d.plugin/squid/squid.chart.py#L94-L107


But in general i agree, default `_check_raw_data` should be

```python
    @staticmethod
    def _check_raw_data(data):
        """
        Check if all data has been gathered from socket
        :param data: str
        :return: boolean
        """
        return not bool(data)
```

instead of

https://github.com/netdata/netdata/blob/0840d52e7147902d7dfd417ca33a415c3da9a5c4/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService.py#L250-L257

l2isbad:
We have 2 choices:
1. read 4096 bytes and break

if module needs to read more or some specific break case it should implement `_check_raw_data`

2. read until buf is b''

if module needs to read more or some specific break case it should implement `_check_raw_data`


@Ferroin what do you think. Current behaviour is 1, should we change it to 2?

l2isbad:
@Ferroin anothey way is to update documentation for the `SocketService` :smile: (about the need to implement '_check_raw_data'  if 4096 is not enough)

Ferroin:
Well yes, that would work too.  I don't think changing the API like this is likely to break anything in-tree, and it might actually reduce the amount of coded needed in some modules.

j0nnyw:
@l2isbad @Ferroin - I have implemented my own `_check_raw_data` method in my plugin as per your suggestion and it is working correctly with data larger than 4096 bytes. A documentation update is probably all that is required in this case. I will close this pull request.

Thanks!

ktsaou:
I am a bit confused. The current implementation of `SocketService` does not allow reading more than 4096 bytes of data?

l2isbad:
Yes, current behaviour is to read first 4096 bytes. (it hasn't been changed since @paulfantom implemented it in that way)

If necessary, It can be overwritten by any module.

ktsaou:
> If necessary, It can be overwritten by any module.

What can be overwritten?

l2isbad:
Read break condition. Current _read break_ is read first 4096 bytes.

ktsaou:
So, each module can define the size of the single read?

l2isbad:
No, you can't directly define number of bytes. 

Module implements `_check_raw_data` interface if it's needed (overwrite default one). 

Examples:

squid:
https://github.com/netdata/netdata/blob/4c57ca9d3af87361eb4577cae8134607611417b7/collectors/python.d.plugin/squid/squid.chart.py#L94-L107

PR #4618, opened 2018-11-12
Bad text in example config

LordMike:
##### Summary
Fixes a bad example config text.


l2isbad:
@LordMike thanks for contributing!

PR #4653, opened 2018-11-13
properly parse network interface names with colon on them

### LINKED ISSUE #4536: inbound packets dropped inbound
All,

So something interesting I came across.  I was able to solve all of my dropped udp packets by increasing the buffers.  This alert still continued to come back on multiple hosts, sometimes not even while under any load.

Doing some digging it appears that while ifconfig will show dropped packets, looking at the interfaces via ethtool -S interface | grep drop I see no drops at all on the interface.  I went ahead and turned on the firewall/ufw on one of the problematic hosts, and since them ifconfig is not showing any inbound packets being dropped.

If you are not running ufw is the best practice here to just this disable this alarm?  

Thanks!

ktsaou:
##### Summary

fixes #4536

This PR fixes 2 bugs:

1. Network interface names may have a colon (`:`) in them. Netdata was incorrectly parsing these network interface names (`:` was assumed to be a separator).

   The effect of this was twofold:

   - the collected values was totally wrong (all values were shifted).
   - when interfaces sharing the same prefix, the same chart was updated multiple times per iteration, resulting in totally crazy values.

2. Families that had `:` in them, resulted in DOM element IDs with `:` in them. This crashed jquery.

This PR fixed them both.

##### Additional Information

```paste below

```

### DIFF
--- a/collectors/proc.plugin/proc_net_dev.c
+++ b/collectors/proc.plugin/proc_net_dev.c
@@ -440,11 +440,14 @@ int do_proc_net_dev(int update_every, usec_t dt) {
     static procfile *ff = NULL;
     static int enable_new_interfaces = -1;
     static int do_bandwidth = -1, do_packets = -1, do_errors = -1, do_drops = -1, do_fifo = -1, do_compressed = -1, do_events = -1;
-    static char *path_to_sys_devices_virtual_net = NULL, *path_to_sys_class_net_speed = NULL;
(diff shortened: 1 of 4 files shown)


ktsaou:
@gmosx 	you have placed the `NETDATA.name2id()` function in `compatibility.js` which I think is not something about compatibility. We generally need this to properly convert strings to DOM element IDs.

gmosx:
> @gmosx you have placed the `NETDATA.name2id()` function in `compatibility.js` which I think is not something about compatibility. We generally need this to properly convert strings to DOM element IDs.

@ktsaou OK, I will move it to `utils.js`.

PR #4657, opened 2018-11-14
Fix changelog path, add all README.md files to Debian package doc

runejuhl:
##### Summary
Recent changes to changelog and documentation structure have caused Debian packaging to break. This PR fixes the path to changelog and adds all `README.md` files to the Debian package.


ktsaou:
@runejuhl thank you!
@paulfantom 	merge as you see fit.

PR #4673, opened 2018-11-15
PR-2976 replacement

jimcooley:
##### Summary
This replaces #2976 

##### Component Name
PagerDuty integration

##### Additional Information
Uses curl instead of pd-send to interact with the PagerDuty api.


jimcooley:
This should have been closed in favor of #4694, which has been merged.

PR #4697, opened 2018-11-20
Fix typo and py2 compatibility issue.

Ferroin:
##### Summary
PR #4667 accidentally introduced a py2 compatibility issue and a typo, both of which break the sensors module.  This fixes both.

Fixes: #4692 

##### Component Name
collectors/python.d.plugin

### DIFF
--- a/collectors/python.d.plugin/python_modules/third_party/lm_sensors.py
+++ b/collectors/python.d.plugin/python_modules/third_party/lm_sensors.py
@@ -35,39 +35,39 @@ class ErrorNoEntry(SensorsError):
     pass
 
 
-class ErrorAccessRead(SensorsError, PermissionError):
+class ErrorAccessRead(SensorsError, OSError):
     pass
 
 
-class ErrorKernel(SensorError, OSError):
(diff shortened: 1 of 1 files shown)


Ferroin:
@l2isbad If you end up accepting this, please just go ahead and merge it.  I'm not going to be available to do so most of the day today, and would rather not wait until tomorrow to get it merged.

l2isbad:
https://github.com/netdata/netdata/blob/0f7f3d811dbfed4d991eafc9994cf06b777d0452/collectors/python.d.plugin/python_modules/third_party/lm_sensors.py#L38-L47

Why we need `OSError`, `ZeroDivisionError` there?

Ferroin:
@l2isbad We don't, but others might.  I added it in the upstream PR because these errors are technically special cases of specific Python errors.  The `ErrorAccess*` cases are permissions errors, which are a type of `OSError` (py3 has `PermissionsError` as a subclass of `OSError`), the `ErrorDivZero` is obviously a case of a `ZeroDivisionError`, etc.

PR #4703, opened 2018-11-21
Add missing quote to tc-qos-helper.sh.in

drwtsn32x:
Missing trailing quote in this file was causing QoS graphs to not show up.

Before the change, the following errors were added to /var/log/netdata/error.log once every second:

```
/usr/libexec/netdata/plugins.d/tc-qos-helper.sh: line 310: unexpected EOF while looking for matching `"'
/usr/libexec/netdata/plugins.d/tc-qos-helper.sh: line 316: syntax error: unexpected end of file
```

After change no more errors and QoS graphs reappeared.


cakrit:
Thank you for the catch and the fix @drwtsn32x :)

PR #4714, opened 2018-11-22
Overzealous escape for #3253

gmosx:
I am not sure this is really needed. But just to be on the safe side.
Also, there is the chance that I totally misunderstood the issue.


ktsaou:
This is will fix this specific issue, but it is not fixing the root cause of it.
Let me check the C source a bit.

gmosx:
> Let me check the C source a bit.

OK.

gmosx:
Closed in favor of #4716

PR #4759, opened 2018-11-27
bugfix: query engine resampling duration

ktsaou:
##### Summary

The query engine has a feature to resample the data to change the units. For example to turn "per second" metrics to "per minute" or "per hour".

This is done by passing `&gtime=SECONDS` to the `/api/v1/data` method.

This feature is not (yet) used in standard dashboards of netdata. it was created as part of the `d3pie` chart (which is also not used yet on standard dashboards), so that, for example the section head of the web log charts could show pie charts with per minute analysis of the data. It is currently available to be used in custom dashboards.

When we refactored the query engine, there was a bug: the time reported was incorrectly stating the ending time of the grouping, not the starting one. This made `d3pie` believe the data returned by netdata had a duration of 1 second, instead of the resampling time requested. The data were returned correctly, only the legend of chart was wrong.

This PR fixes also 2 issues related to resampling:

1. If the duration requested was smaller that the re-sampling rate (i.e. requested a time-frame of 50 seconds, with "per minute" re-sampling), the query engine was returning "per 50 seconds" re-sampling. Now, it extends the time-frame to accommodate at least one full re-sampling duration.

2. If the duration requested was not aligned to the re-sampling rate (i.e. requested a time-frame of 100 seconds, with "per minute" re-sampling), the query engine was returning only 1 point, thus ignoring the oldest 40 seconds. This had the problem, that on hover on another chart, this chart was presenting an error (`no data`) on the re-sampled chart. This PR extends the time-frame to accommodate another point, if the remaining duration is above 1/10th of the re-sampling duration requested. So the `no data` effect now appear only in cases 90% of the data should not be displayed.

##### Additional Information

```paste below

```

### DIFF
--- a/web/api/queries/query.c
+++ b/web/api/queries/query.c
@@ -492,7 +492,7 @@ static inline void do_dimension(
     r->min = min;
     r->max = max;
     r->before = max_date;
-    r->after = min_date;
+    r->after = min_date - (r->group - 1) * r->st->update_every;
     rrdr_done(r, rrdr_line);
 
     #ifdef NETDATA_INTERNAL_CHECKS
@@ -539,7 +539,7 @@ static void rrd2rrdr_log_request_response_metdata(RRDR *r
          , resampling_group
 
          // after
-         , (size_t)r->after - (group - 1) * r->st->update_every
+         , (size_t)r->after
          , (size_t)after_wanted
          , (size_t)after_requested
          , (size_t)rrdset_first_entry_t(r->st)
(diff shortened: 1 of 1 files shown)


web/api/queries/query.c
```diff
@@ -666,26 +666,41 @@ RRDR *rrd2rrdr(
             info("INTERNAL CHECK: %s: requested gtime %ld secs, is greater than the desired duration %ld secs", st->id, resampling_time_requested, duration);
             #endif
 
-            group = available_points; // use all the points
+            after_requested = before_requested - resampling_time_requested;
+            duration = before_requested - after_requested;
```
vlvkobal: Is it better for readability than `duration = resampling_time_requested`?

ktsaou: This line appears 3 times. I tried to have it exactly the same in all 3 occasions. For readability.

PR #4765, opened 2018-11-27
Improve support for slack recipients

### LINKED ISSUE #3722: Send alerts via Slack to a single user (direct message)?
I'm trying to send netdata alerts to just myself on my Slack instance, and it doesn't seem possible. The `DEFAULT_RECIPIENT_SLACK` variable in `/etc/netdata/health_alarm_notify.conf` seems to only accept a channel. When I try using my username, I get 404's from the test script:

```
# SENDING TEST WARNING ALARM TO ROLE: sysadmin
2018-05-12 18:16:10: alarm-notify.sh: ERROR: failed to send slack notification for: server test.chart.test_alarm is WARNING to 'kevlar1818', with HTTP error code 404.
# FAILED

# SENDING TEST CRITICAL ALARM TO ROLE: sysadmin
2018-05-12 18:16:10: alarm-notify.sh: ERROR: failed to send slack notification for: server test.chart.test_alarm is CRITICAL to 'kevlar1818', with HTTP error code 404.
# FAILED

# SENDING TEST CLEAR ALARM TO ROLE: sysadmin
2018-05-12 18:16:11: alarm-notify.sh: ERROR: failed to send slack notification for: server test.chart.test_alarm is CLEAR to 'kevlar1818', with HTTP error code 404.
# FAILED
```

It works fine if I use a channel (either public or private).

### LINKED ISSUE #4755: Change in Incomming Webhooks Slack API breaks alerts
##### Bug report summary
Slack ignores "channel" and "username" parameters.
The API has been changed so that now it is required to specify the channel the messages will be posted to during the creation of a webhook 

>The majority of your legacy code for sending messages using incoming Webhooks should continue to work within a Slack app without much modification; the only thing you can no longer do is customize the destination channel and author identity at runtime.

 [source](https://api.slack.com/custom-integrations/incoming-webhooks)

I guess that it also makes #3722 obsolete.

##### Component Name
slack.alert

##### Steps To Reproduce
Create new incomming webhook by following this [guide](https://api.slack.com/incoming-webhooks)
Set DEFAULT_RECIPIENT_SLACK parameter in health_alarm_notify.conf.
Try to send an alert to that channel by using ./alarm-test.sh script.

##### Expected behavior
Alert being send to chosen channel

cakrit:
Fixes #3722 
Fixes #4755 

##### Summary

- Removed the default addition of '#' to the "channel", so it can support both channels and users
- Add the '#' if it's not already there in the recipient, to specify a channel (for backwards compatibility). 
- If the recipient is just a '#', netdata will not be sending the channel at all. This means that users only need to configure the channel or the user on the Slack webhook.

##### Component Name
health/notifications

### DIFF
--- a/health/notifications/alarm-notify.sh.in
+++ b/health/notifications/alarm-notify.sh.in
@@ -1476,9 +1476,21 @@ send_slack() {
 
     for channel in ${channels}
     do
+        # Default entry in the recipient is without a hash in front (backwards-compatible). Accept specification of channel or user.
+        if [ "${channel::1}" != "#" ] && [ "${channel::1}" != "@" ] ; then channel="#$channel"; fi
+        
+        # If channel is equal to "#" then do not send the channel attribute at all. Slack also defines channels and users in webhooks.
+        if [ "${channel}" = "#" ] ; then 
+            ch=""
+            chstr="without specifying a channel"
+        else
+            ch="\"channel\": \"${channel}\","
+            chstr="to '${channel}'"
+        fi
+        
         payload="$(cat <<EOF
         {
-            "channel": "#${channel}",
+            $ch
             "username": "netdata on ${host}",
             "icon_url": "${images_base_url}/images/banner-icon-144x144.png",
             "text": "${host} ${status_message}, \`${chart}\` (_${family}_), *${alarm}*",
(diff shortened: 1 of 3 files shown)


cakrit:
The only thing I haven't tested it with is with different role recipients, but they are using exactly the same code to send the notification, so it shouldn't have any issues.

Issue #4010, opened 2018-07-26
mdstat RAID0 support


weidercs:
Not sure if it has to do with my first array being md1 as opposed to md0 but here's the debug output:

```
netdata@media1:/etc/netdata/python.d$ /usr/libexec/netdata/plugins.d/python.d.plugin 1 debug trace mdstat
2018-07-25 21:57:37: python.d INFO: plugin: main: Using python 2
2018-07-25 21:57:37: python.d DEBUG: plugin: main: module load source: 'mdstat' => [OK]
2018-07-25 21:57:37: python.d DEBUG: plugin: main: module load config: 'mdstat' => [OK]
2018-07-25 21:57:37: python.d DEBUG: plugin: main: job initialization: 'mdstat mdstat' => ['OK']
2018-07-25 21:57:37: python.d DEBUG: plugin: main: module status: 'mdstat' => [OK] (jobs: 1)
2018-07-25 21:57:37: python.d ERROR: mdstat: mdstat: Failed to read data from /proc/mdstat or there is no active arrays
2018-07-25 21:57:37: python.d INFO: mdstat: mdstat: check() => [FAILED]
2018-07-25 21:57:37: python.d INFO: plugin: main: FINISHED
```
```
netdata@media1:/etc/netdata/python.d$ cat /proc/mdstat
Personalities : [raid0] [linear] [multipath] [raid1] [raid6] [raid5] [raid4] [raid10] 
md1 : active raid0 sdd[1] sdb[0]
      15627790336 blocks super 1.2 512k chunks
      
unused devices: <none>
```

l2isbad:
Hi @weidercs 

```
Personalities : [raid1] [raid6] [raid5] [raid4]
md_d0 : active raid5 sde1[0] sdf1[4] sdb1[5] sdd1[2] sdc1[1]
      1250241792 blocks super 1.2 level 5, 64k chunk, algorithm 2 [5/5] [UUUUU]
      bitmap: 0/10 pages [0KB], 16384KB chunk

unused devices: <none>
```

I see no `[5/5]` part in your output

> [n/m] means that ideally the array would have n devices however, currently, m devices are in use. Obviously when m >= n then things are good.

weidercs:
Looks like Raid-0 doesn't print that information, is it necessary for netdata?

> Depending on what type of array is defined, slightly different information is available through /proc/mdstat.

```
md1 : active raid0 sdc1[1] sdb1[0]
      35840768 blocks 64k chunks
```

[...]

PR #4768, opened 2018-11-27
Move mdstat python module to proc plugin

vlvkobal:
##### Summary
cpuidle python module reads data from files:
`/proc/mdstat`
`/sys/block/*/md/mismatch_cnt`

Partly implements #4541

##### Component Name
`mdstat` python module
`proc` plugin

##### Additional Information
C module parses `mdstat` using `procfile()` capabilities, instead of using regular expressions.


Ferroin:
Relevant: #4010

cakrit:
Please also update README.md when the module becomes obsolete. Same as with the BASH obsolete modules.

ilyam8:
@vlvkobal ping me whe this is ready, i will test it

ktsaou:
Please rebase this. It broke with the merge of cpuidle.
Is this ready for review?

ktsaou:
@vlvkobal 	just read it. Nice work!

I see only 1 issue:

If we add and remove raid arrays all the time, the plugin will not mark the obsolete ones as `OBSOLETE` in netdata, so netdata will never cleanup. This is a somewhat rare use case, it was probably not supported by the python plugin either, but I think it is relatively easy to be supported here.

So, I suggest to refactor it, like this:

- The structure is mainly needed just for keeping pointers to charts and dimensions. All other data are read from the data source every time. So, I would create a linked list with the pointers, a name and a hash. A function called `find_or_allocate_mdstat_data(const char *name)` could be used to search through the list or create a new entry. This can be done with hashes, to be very fast.

- Each of the nodes in the linked list could have a member called `updated`. This should be set to 1 when `find_or_allocate_mdstat_data()` returns it.

- At the end of each iteration, a loop through the list should identify the nodes that have `updated == 0`, and mark the charts as `OBSOLETE` in netdata (and possibly free that nodes). It should also set `updated = 0` on all nodes, to be ready for the next iteration.

I think the above setup will also simplify it a bit.

What do you think?

PR #4796, opened 2018-11-30
[wip] internal clock of netdata in microseconds

ktsaou:
##### Summary

**THIS IS WORK IN PROGRESS. IT DOES NOT WORK YET.**

This PR switches the internal clocks of netdata to microseconds. All timestamps, all durations, everything netdata does internally is now switched to microseconds, including data collection, backends, streaming, health monitoring, statsd.

Before this PR, netdata was using microseconds resolution only for interpolating collected metrics before storing them in the database. Apart that, everything else had second resolution.

After this PR, netdata can operate at the sub-second level for all its functions.

Currently this PR just changed the datatypes and the names of all related variables (so that it can be easier to spot left-overs, ie. all variables ending in `_usec` have to be compared and used in operations with other variables also in `_usec`).

To complete the task, several APIs need to also change to support microseconds (streaming, external plugins API, the javascript code for the UI, etc).

> I have intentionally removed a function to make travis fail - just to make sure this will not be merged accidentally.


gmosx:
What is the status on this?

cakrit:
There are no plans to do this in the short term. Closing and we can revisit when we agree to progress.

Issue #4085, opened 2018-08-28
Postqueue from lxc container


WoozyMasta:
When I try to get a postfix message queue from an lxc container, I get an error

```
python.d ERROR: postfix: local: _get_data() failed. Command: ['/usr/bin/lxc', 'exec', 'mail', '--', '/usr/sbin/postqueue', '-p']. Error: 'NoneType' object has no attribute '__getitem__'
```

To get the queue, I changed the command in the config `/etc/netdata/python.d/postfix.conf`
```
local:
  command: 'lxc exec mail -- /usr/sbin/postqueue -p'
```
How can properly parse the output of commands from the container?

ilyam8:
Hey @WoozyMasta 

Please run `python.d.plugin` in debug mode

```
opt/netdata/usr/libexec/netdata/plugins.d/python.d.plugin postfix debug trace
```

ilyam8:
edited

orgoj:
Try: lxc-attach -n mail -- postqueue -p

PR #4799, opened 2018-11-30
Add debug instructions for python modules

cakrit:
Used instruction from @ilyam8 in #4085 to add a (How to debug python modules) section.

### DIFF
--- a/collectors/python.d.plugin/README.md
+++ b/collectors/python.d.plugin/README.md
@@ -63,7 +63,14 @@ other_job:
 
 `update_every`, `retries`, and `priority` are always optional.
 
----
+## How to debug a python module
+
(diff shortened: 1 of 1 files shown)


Ferroin:
This should mention that for accurate results, the command has to be run as the same user which Netdata normally runs as (usually `netdata` or `root`).

ilyam8:
Ye, we have it in the [wiki](https://docs.netdata.cloud/docs/add-more-charts-to-netdata/)
```
# become user netdata
sudo su -s /bin/bash netdata

# execute the plugin in debug mode, for a specific module.
# example for the python plugin, mysql module:
/usr/libexec/netdata/plugins.d/python.d.plugin 1 debug trace mysql
```

cakrit:
Updated it. Add-more-charts will need a lot of work, not touching it yet.

PR #4904, opened 2018-12-03
Fix integer out of range error on Postgres 11

arenddeboer:
##### Summary
Fix integer out of range error on Postgres 11

##### Component Name
collectors/python.d.plugin/postgres/postgres.chart.py

##### Additional Information
On PostgreSQL 11 (haven't tested on other versions) an integer out of range error is thrown due to casting to INT:
```
2018-12-03 03:22:25: python.d ERROR: postgres: socket: update() unhandled exception: integer out of range
```
Casting to BIGINT solves this error.
The PostgreSQL Docs do not list wal_block_size as bigint, so I do not fully understand why this is required.
https://www.postgresql.org/docs/11/functions-info.html

Netdata: 1.11.1-68-g0fad9bf5
OS: 	Ubuntu 18.10
PostgreSQL: PostgreSQL 11.1 (Ubuntu 11.1-1.pgdg18.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0, 64-bit

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -261,7 +261,7 @@ WHERE application_name IS NOT NULL;
     'REPSLOT_FILES': """
 WITH wal_size AS (
   SELECT
-    current_setting('wal_block_size')::INT * setting::INT AS val
+    current_setting('wal_block_size')::BIGINT * setting::BIGINT AS val
   FROM pg_settings
   WHERE name = 'wal_segment_size'
   )


arenddeboer:
On further testing, it is sufficient to do:
```current_setting('wal_block_size')::INT * setting::BIGINT AS val```

**Comment by ilyam8** [comment]
@anayrat 
please review

**Comment by anayrat** [comment]
Thanks but it is is little bit more complicated. The unit of wal_segment_size has changed in 11 : 
In PG 10:
```
name            | wal_segment_size
setting         | 2048
unit            | 8kB
```
In PG 11:
```
name            | wal_segment_size
setting         | 16777216
unit            | B
```

**Comment by anayrat** [comment]
So this part of the query is different before and after PG 11 : 
Before PG 11 : 
```
 WITH wal_size AS (
  SELECT
    current_setting('wal_block_size')::BIGINT * setting::BIGINT AS val
  FROM pg_settings
  WHERE name = 'wal_segment_size'
  ) 
```

After PG11:
```
 WITH wal_size AS (
  SELECT
     setting AS val
  FROM pg_settings
  WHERE name = 'wal_segment_size'
  ) 
```

I think we should find a way to manage queries across versions. I like how it is managed in check_pgactivity : https://github.com/OPMDG/check_pgactivity/blob/master/check_pgactivity#L7394

The script detect server's version and use the right query.

ilyam8:
> The script detect server's version and use the right query.

Yeah, we should implement it too.

@arenddeboer as you can see this is more complicated

arenddeboer:
Thanks for looking into it.
Can't promise any work on this, but might pick this issue up at a later time.

ilyam8:
i'll try to refactor `postgres` module this weekend

PR #4906, opened 2018-12-03
Don't use IE11 incompatible for-const #4710

gmosx:
IE11 seems to have problems with the `for (const...` construct. This commit removes all occurrences of  this construct.


ktsaou:
approved. Should you use `var` then? Not specifying anything is better?

gmosx:
I am not sure, `for (const` was the correct construct.

gmosx:
@ktsaou, added the `var` keyword, pls review again.

PR #4912, opened 2018-12-04
Introduced IEC-compliant unit abbreviations #4711

gmosx:
This PR introduced additional, IEC-compliant unit-conversions to dashboard.js


ktsaou:
I think this is not right.
We should not change `1024` to `1000`.
We should only change `KB` to `KiB`, `MB` to `MiB`, `GB` to `GiB`, etc.

But for this to work, we also have to change how data collection plugins report metrics to netdata. @gmosx 	this has do be done by @ilyam8 	and @vlvkobal 	.

cakrit:
I'll set up a call to discuss what needs to be done.

gmosx:
@cakrit I reverted the coefficients for the older abbreviations, can you check pls?

ktsaou:
If this is not WIP, remove it from the title to merge it.

PR #4916, opened 2018-12-04
RabbitMQ chart for message rates should be "line"

dex4er:
##### Summary

I changed a types of chars for RabbitMQ.

Chart for message rates should be "line". It is because for correct system the "ack", "deliver" and "publish" should be similar and it is much easier to notice it with "line" type of chart.

Usually for memory and disk space we use "area" type so it is changed too.

##### Component Name

collector rabbitmq


ilyam8:
@dex4er thank you for contributing :+1:

Issue #2477, opened 2017-07-18
FQDN in alert sending


m-emelchenkov:
Hi!

it's nice to have FQDN (hostname -f) in alerts (which delivered via alarm-notify.sh script), because now I get, for example: "www needs attention - ram in use". So I do not know what "www" means because it is only just a part of a full domain name which is a hostname for the server.

ktsaou:
Hi,

the hostname used in alerts is the one defined in `netdata.conf` at `[global].hostname` (so that slave hosts will give the right hostname). It defaults to whatever `gethostname()` returns.

But you are right. We should probably use also `getdomainname()` and append the two to fix the real full hostname.

We could also use `hostname -f` at the little comment at the end of emails to uniquely identify the host sending the alarm:


![image](https://user-images.githubusercontent.com/2662304/28332351-2d6a708c-6bfd-11e7-9dfc-10960a3357fb.png)

ktsaou:
To implement the first (add domain name), we probably need to fix #2335 first.

m-emelchenkov:
I use Telegram bot as an alarm delivery method. Thanks for reply and for info about [global].hostname—yes, I already know this option, and I use it to name some servers more human-readable, just want to have FQDN as a default for those servers which hostname.domainname is already have good meaning so do not configure it manually. I agree, UI should be fixed in first place.

ktsaou:
@Ferroin 	we could modify `alarm-notify.sh` to send the FQDN of the host sending the alarm. It could be an option actually.

cakrit:
I think that we shouldn't replace the existing functionality of the host that is configured in netdata.conf, just add a line with the FQDN and the IP on each alarm, regardless of how netdata calls alarm-notify.sh. It's a very easy addition. Agreed?

Ferroin:
FQDN should be easy, though I agree that it should be optional, especially since figuring out that there isn't one can take a long time on a system that doesn't have one.

IP is trickier, because it's not unusual for systems to have 2 or more IP's.

Ferroin:
Alternatively, why not just make it so the user can choose to either display the short hostname (like we currently do) or the FQDN?  That would amount to at most 4-5 lines of changes in `alarm-notify.sh`.

cakrit:
Fine. Let's say this then:
- Default is exactly as is. 
- We add an option to health_alarm_notify.conf to use FQDN instead.
[...]

Issue #809, opened 2016-08-21
health monitoring enhancements


ktsaou:
- [x] basic rendering of alarms on the dashboard.
- [x] allow alarms to repeat notifications #2956 **top priority**
- [x] Silencing alarms/maintenance periods #2673, #3187 
- [ ] render all alarm data on the dashboard and allow users to view health configuration per chart, including the available variables. #4941 
- [ ] central monitoring: allow alarms to be propagated to a central netdata server, which will send all notifications. The central monitoring server should also be able to detect when netdata servers are restarted, or netdata servers are unreachable. #1527 
- [ ] allow propagating the alarms to third party NMS systems. Currently, this can only be done by setting the `exec` line on alarms and templates. However, only one such line is allowed. We could have a solution that allows multiple scripts to handle the same alarm. 
- [ ] implement an alarm flood protection system. #1663
- [ ] allow variables in the `family` index in the form `context.variable`. This is useful in cases where an alarm template needs to refer to another alarm that overlaps with a dimension, so it is not directly exposed in the `family` index. #4154 
- [ ] allow charts to have tags and alarm templates to examine these tags. This would allow disk space alarm templates to be applied on disks with a specific mount point, or fstype, etc. #(https://github.com/firehol/netdata/issues/1231#issuecomment-267164281) -> #4986, #3813 
- [x] Rich notifications #2351, #946 
- [ ] More integrations #2720, #3774, #3997, #3788
- [x] Make alarm-notify.sh modular #3963 
- [ ] Better alarms and alarm handling #4727, #4313, #2477, #1663
- [ ] USE and RED methods checklist #4313

davidak:
one should be able to acknowledge warning/critical states, so it don't send alerts. the acknowledgement get's deleted if it recovers.

nice would be an option to acknowledge the current value (97% disk usage is OK, but alert again for >= 98%)!

optionally comment acknowledgement
optionally send alert/mail for adding an acknowledgement (with comment)
optionally set expire date/time or timespan like `30m` or `1h 30m` or `2d`

this is how it looks in Icinga Web 1:
<img width="590" alt="bildschirmfoto 2016-09-01 um 20 49 14" src="https://cloud.githubusercontent.com/assets/91113/18180256/538c456a-7086-11e6-80c3-f7233ed4fc56.png">

ktsaou:
this is nice, but it is to be done after roles, authentication, etc.

rasmusdencker:
[...]

PR #4943, opened 2018-12-07
Add support for providing FQDN in alarm notifications.

Ferroin:
##### Summary
This adds an option to alarm-notify.sh to make it use the system's FQDN instead of it's simple hostname when sending alarm notifications.  This can be enabled by adding `use_fqdn="YES"` to `health_alarm_notify.conf`.

This does not work correctly for alarms being sent by a master node on behalf of a slave system, and includes an explicit check so that it falls back to just sending the simple hostname in such cases.

This commit also cleans up misuse of the `${this_host}` variable, which is supposed to just be a temporary variable.

##### Component Name
health

##### Additional Information
Minimally tested, it runs correctly on my systems and does as advertised.

Relevant to:#809

Fixes: #2477

### DIFF
--- a/health/notifications/alarm-notify.sh.in
+++ b/health/notifications/alarm-notify.sh.in
@@ -198,7 +198,7 @@ if [ ${1} = "unittest" ] ; then
     old_status="${5}"          # the previous status: REMOVED, UNINITIALIZED, UNDEFINED, CLEAR, WARNING, CRITICAL
 else
     roles="${1}"               # the roles that should be notified for this event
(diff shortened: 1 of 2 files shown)


Ferroin:
@cakrit The fact that the check fails when `hostname` is set in `netdata.conf` was intentional, though I forgot to document it.  The intent was that we wouldn't override whatever the user set for `hostname`, only the default host name that was auto-detected by Netdata, which _should_ match the output of `hostname -s`.

cakrit:
From my tests it works well, so I'm merging.

PR #4960, opened 2018-12-09
Move power supply python module to proc plugin

vlvkobal:
##### Summary
Linux power supply python module reads data from directories:
`/sys/class/power_supply/*`

Partly implements #4541

##### Component Name
`linux_power_supply` python module
`proc` plugin


Ferroin:
Adding myself to the review list as I wrote the original module, actively use it, and have a vested interest in making sure this works.

I probably won't have a chance to actually properly test this until later this week though.

vlvkobal:
@Ferroin It isn't ready yet. I'll let you know when It's finished.

Ferroin:
@vlvkobal OK.  Let me kn ow if you've got any questions about the Python code or the API's it's using.

vlvkobal:
@Ferroin It's ready for testing.

ktsaou:
I didn't go through it in very detail.
It seems good, but please make sure it is tested with valgrind before merging.

vlvkobal:
> it seems you need a better version of read_single_file_number() that will also support keeping open the file

I thought about it, but it seemed to me that in this particular case it would have too many parameters.

vlvkobal:
@Ferroin I'm merging it, but I still need your feedback.

Ferroin:
@vlvkobal Sorry about taking so long to get back with you.  I didn't get a chance until about an hour ago to actually try and test this.

Overall looks good, the only issue I see is that the documentation still says that the module provides `min` and `empty` metrics that always read back as zero if those attributes aren't present but corresponding `max` and `full` metrics are present.

The original python module provided that functionality so that people could write alarms that look at charge, energy, or voltage as a percentage of the device's expected operating range for that parameter.  This isn't something I would consider critical functionality (though it would be nice to have), but the documentation should be updated to reflect the fact that this functionality isn't present if it's not going to be added.

PR #4970, opened 2018-12-11
fix month 'Dec' being detected as IPv6 address in ovpn python.d plugin

vpnable:
##### Summary
When getting the number of users and total amount of transferred bytes in the openvpn python.d plugin the TLS_REGEX matches more than it should and detects Date and Time values as valid users, e.g. Tue **Dec 11 17**:28:04 2018

##### Component Name
ovpn_status_log.chart.py


ilyam8:
@vpnable thanks!

PR #4978, opened 2018-12-12
Fix to #4968, custom recipients were not working properly

### LINKED ISSUE #4968: custom notification method does not work
##### Bug report summary

I have a report saying the after #3960 the `custom` notification method does not work any more.

The `custom` method allows users to provide a bash function that is called for every notification, and which is configured via the same means (roles, recipients, filter by criticality, etc).

cakrit:
##### Summary
Fixes #4968 
The custom recipient variable substitution wasn't working properly

##### Component Name
health notifications


Ferroin:
I"m going to wait on a response from @epoppis about whether this works or not before accepting it, as my own testing obviously didn't find the issue.

PR #4944, opened 2018-12-07
postgres fix: detect servers version and use the right query

ilyam8:
##### Summary
Fixes: #4910
Fixes: #4904

##### Component Name
python.d module `postgres`

##### Additional Information

add support for use specific query depending on server version

@anayrat 
@arenddeboer 

if you have free time please test, the issue should be fixed

@anayrat please look at all queries, I could break something :sweat_smile:

PR #4988, opened 2018-12-14
postgres timeouts

ilyam8:
##### Summary
added to postgres module

- `connect_timeout` (default is 2 sec)
- `statement_timeout` (default is 5000ms)

##### Component Name

`netdata/collectors/python.d.plugin/postgres`

##### Additional Information

Based on 
https://github.com/netdata/netdata/pull/4944#issuecomment-446754970
https://github.com/netdata/netdata/pull/4944#issuecomment-447258227

@anayrat 

`connect_timeout` works nice

But i have problems with `statement_timeout`. 

I tried to emulate network/other delay/issues with iptables - `iptables -A INPUT -p tcp --dport 5432 -j DROP`. Module hangs forever, `statement_timeout` doesn't help.

But

```
>>> import os
>>> os.environ['PGOPTIONS'] = '-c statement_timeout=1000'
>>> import psycopg2
>>> cnn = psycopg2.connect("dbname=test")
>>> cur = cnn.cursor()
>>> cur.execute("select pg_sleep(2000)")
psycopg2.extensions.QueryCanceledError: canceling statement due to statement timeout
```

it works in such cases^^

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -17,6 +17,12 @@ except ImportError:
 from bases.FrameworkServices.SimpleService import SimpleService
 
 
+DEFAULT_PORT = 5432
+DEFAULT_USER = 'postgres'
+DEFAULT_CONNECT_TIMEOUT = 2       # seconds
+DEFAULT_STATEMENT_TIMEOUT = 2000  # ms
+
+
 WAL = 'WAL'
 ARCHIVE = 'ARCHIVE'
 BACKENDS = 'BACKENDS'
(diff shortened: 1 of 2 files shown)


anayrat:
> I tried to emulate network/other delay/issues with iptables - iptables -A INPUT -p tcp --dport 5432 -j DROP. Module hangs forever, statement_timeout doesn't help.

It is normal, in this case it is a kind of network timeout, not a statement timeout. For network timeout it is managed on server side : https://www.postgresql.org/docs/current/runtime-config-connection.html (look tcp_keepalive*)

Statement timeout is useful when your query lasts too long. So pg_sleep is the right test ;)

ilyam8:
Thanks for the explanation!

Issue #4981, opened 2018-12-13
Netdata doesn't properly lookup docker container name when running in ECS with task level cpu/memory limits enabled


climbbike:
So we are using Netdata in a hybrid environment with Docker. It works great including monitoring docker containers, however, we found that when running Docker on AWS ECS, if ECS_ENABLE_TASK_CPU_MEM_LIMIT is enabled (sets mem/cpu limits per task), the docker container name is not pulled correctly.

From what I can tell cgroups are normally named like this for Docker:

cat /proc/self/cgroup
9:perf_event:/docker/81509de4623e2fb6ac9a799df48d03c7a0787796cfd92fd984b80e027c4bb221
8:memory:/docker/81509de4623e2fb6ac9a799df48d03c7a0787796cfd92fd984b80e027c4bb221
7:hugetlb:/docker/81509de4623e2fb6ac9a799df48d03c7a0787796cfd92fd984b80e027c4bb221
6:freezer:/docker/81509de4623e2fb6ac9a799df48d03c7a0787796cfd92fd984b80e027c4bb221

However in ECS if you are limiting cpu/memory per task, it changes the names to be like this:

cat /proc/self/cgroup
9:perf_event:/ecs/a52246f2-daa7-47b4-9e84-0043331e1da0/38dcd24c1a7ac4b0401a1ed29ef79b0c030631df33322db9b092687c0bfd0a38
8:memory:/ecs/a52246f2-daa7-47b4-9e84-0043331e1da0/38dcd24c1a7ac4b0401a1ed29ef79b0c030631df33322db9b092687c0bfd0a38
7:hugetlb:/ecs/a52246f2-daa7-47b4-9e84-0043331e1da0/38dcd24c1a7ac4b0401a1ed29ef79b0c030631df33322db9b092687c0bfd0a38

The relevant code seems to be here:
https://github.com/netdata/netdata/blob/master/collectors/cgroups.plugin/cgroup-name.sh.in

specifically with:

`    if [[ "${CGROUP}" =~ ^.*docker[-_/\.][a-fA-F0-9]+[-_\.]?.*$ ]]
        then
        # docker containers
`
Since the pattern is not matched, the id is not parsed and the name is not properly looked up.

From what I can tell, the only way to work around this is to disable mem/cpu limits per task in which case the naming goes back to the standard Docker naming but we would like to continue to use mem/cpu limits on ECS.

Thanks for any help or suggestions.

vlvkobal:
[...]

PR #4991, opened 2018-12-14
Update cgroup-name.sh.in

n0coast:
##### Summary
Not a perfect fix, but does get the container name if using ECS into netdata graphs. See #4981 
##### Component Name
collectors/cgropus.plugin/cgroup-name.sh.in
##### Additional Information
I'm not familiar enough with the way this gets called so it's tough to test, but this changes the random string of numbers currently being pulled when using ECS to the actual container name.


n0coast:
Before updated cgroup-name.sh:
image

After updated cgroup-name.sh:
image

ktsaou:
Thank you!

PR #4993, opened 2018-12-15
Integrate patches from freeipmi and set paramters

### LINKED ISSUE #3977: FreeIPMI Plugin cant graph the wattage
With ipmitool I can see for example the Power Meter and even how much power the different power supplies use. 
For example (with ipmitool sdr)
`Power Meter      | 326 Watts         | ok`
`Power Supply 1   | 170 Watts         | ok`
`Power Supply 2   | 135 Watts         | ok`

On netdata i can only see graphs for 
`IPMI Sensors State `
`IPMI Events `
`System Celcius Temperatures read by IPMI`

Is it possible to add a graph of the wattages?

Preisschild:
Fixes #3977


ilyam8:
gj @Preisschild :+1:

ktsaou:
nice! very nice!

Issue #2422, opened 2017-07-06
on frontend, if JavaScript is disabled, there's no graceful degradation


magicgoose:
if JavaScript is disabled, there's no graceful degradation, everything breaks down beyond recognition.

Tested on `http://london.my-netdata.io/default.html`

This is how it looks like:
![screen shot 2017-07-06 at 07 27 17-fullpage](https://user-images.githubusercontent.com/1442206/27895655-d6a44dca-621c-11e7-8a27-61debb030f8a.png)

ktsaou:
You are right. I'll try to fix that.

gmosx:
But what would be a 'graceful degradation' ? And why would someone want to use Netdata GUI without JavaScript enabled?

Ferroin:
> But what would be a 'graceful degradation' ?

Displaying a message that you need JavaScript enabled to view this site, just like used to be the standard for sites that required it.  Bonus points if it's done in a way that works correctly with stuff that blocks script execution without disabling JavaScript, because browser extensions that do that are starting to become popular in the tech-savvy community.

> And why would someone want to use Netdata GUI without JavaScript enabled?

They probably wouldn't, and I'm not sure we should support it.

gmosx:
OK, I guess displaying a message is reasonable. Regarding proper handling of the script blockers, I will do some research.

Ferroin:
@gmosx I can't say I've done any research, but the few pages I've seen that do handle them correctly have the message in the actual HTML outside of the usual `<noscript>` tags so that it gets rendered, and then delete that first-thing in their scripts.

PR #4999, opened 2018-12-17
Show a warning if JavaScript is disabled #2422

gmosx:
Implements #2422

### DIFF
--- a/web/gui/index.html
+++ b/web/gui/index.html
@@ -55,10 +55,16 @@
 </head>
 
 <body data-spy="scroll" data-target="#sidebar" data-offset="100">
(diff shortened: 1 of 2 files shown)


gmosx:
Please review/test extensively. This is a potential dangerous change (i.e. it will block Netdata Agent Web/GUI from starting).

gmosx:
The font is the font that is already used by Netdata web/gui to display the loading message.

mfundul:
You can get the Edge version by clicking at the three dots, choosing settings and scrolling to the bottom of General Settings where it says "About this app". You can get Windows 10 version by running the command the winver.exe tool.

Issue #5068, opened 2018-12-28
How can I create alarms on statsd synthetic charts.


ritzzbv:
##### Question summary
I have a system in which I am generating statsd counters and using synthetic charts I am showing multiple statsd counters in single chart.
How can I create alarms on these metrics shown in synthetic charts.
##### OS / Environment
Ubuntu.
##### Component Name
Statsd
##### Expected results

Sample statsd config is below:
`[app]
  name = ABC_Stats
  metrics = ABC_Stats*
  private charts = no

[GoWorker]
  title = Goworker related
  family = Goworker
  context = ABC_Stats.context
  units = entries/min
  type = area
  dimension =  ABC_Stats_Goworker.RESPONSES 'Responses' last 1 1
  dimension =  ABC_Stats_Goworker.SERVER1_RUN 'SERVER1' last 1 1
  dimension =  ABC_Stats_Goworker.SERVER2_RUN 'SERVER2' last 1 1`

Now I want to generate alarms on the above three metrics mainly, ABC_Stats_Goworker.RESPONSES, ABC_Stats_Goworker.SERVER1_RUN, ABC_Stats_Goworker.SERVER2_RUN

Thanks in advance.

cakrit:
You just need to add a new .conf file under your /etc/netdata/health.d folder with the alarm definition. Have you gone through [the documentation](https://docs.netdata.cloud/health/)? 

A very helpful API call to show the [variables](https://docs.netdata.cloud/health/#variables) available to the alarm you will define is https://registry.my-netdata.io/api/v1/alarm_variables?chart=GoWorker (not sure if it's case-sensitive, you can try it).

ritzzbv:
@cakrit thank you for your quick reply. I have been through docs but was getting confused w.r.t synthetic charts. Will Try to do more research now.
Thanks

cakrit:
Of course the URL I sent to get the variables availabe for your alarm is wrong, it should be http://yourhost:yourport/api/v1/alarm_variables?chart=...

ritzzbv:
Fully solved my issue by going through documentation

PR #5069, opened 2018-12-28
Add how to add new alarm

cakrit:
Related to #5068 

##### Summary

Add quick guide on how to add a new alarm.


ktsaou:
I think an empty alarm is also needed, so that people can copy and paste it.

PR #5076, opened 2018-12-29
Fix update instructions URL in frontend

jorisvervuurt:
##### Summary

In the web GUI, I noticed that the update instructions URL in the update modal pointed to a non-existing page.

This commit fixes the URL, so that it points to the correct document.

##### Component Name

Web GUI


cakrit:
Please get this approved, it's because installer was moved under packaging.

PR #5080, opened 2018-12-30
Dashboard grammar change

Xalaxis:
##### Summary

Changes the message for Total Swap I/O to something that reads a little better to me, also adds a missing parentheses.

##### Component Name

Default dashboard

### DIFF
--- a/web/gui/dashboard_info.js
+++ b/web/gui/dashboard_info.js
@@ -647,7 +647,7 @@ netdataDashboard.context = {
     },
 
     'system.swapio': {
-        info: 'Total Swap I/O. (netdata measures both <code>in</code> and <code>out</code>. If either of them are not shown in the chart, it is because they are zero - you can change the page settings to always render all the available dimensions on all charts).'
+        info: 'Total Swap I/O. (netdata measures both <code>in</code> and <code>out</code>. If either of them is not shown in the chart, it is because they are zero - you can change the page settings to always render all the available dimensions on all charts).'
     },
 
     'system.pgfaults': {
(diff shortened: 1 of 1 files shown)


Xalaxis:
This is probably area/web

Xalaxis:
@gmosx You are correct, I've removed the unneeded parenthesis.

Xalaxis:
@cakrit How does this look?  The changes now are:
"If either of them **are** not shown in the chart, it is because **they** are zero" rather than "If either of them **is** not shown in the chart, it is because **it is** zero"

Xalaxis:
I think you're right.  I've swapped it back again.  So I guess that just leaves the change of "it is because it is zero" to "it is because they are zero".  That seems alright?

cakrit:
No, I'm sorry mate, it was grammatically correct as it was. What's confusing is the use of "it is". The first "it is" in that sentence refers to **the fact** of either of the metrics being zero. We could have said "if two elephants appear to me to be pink, it is because I've consumed too much LSD". So it has nothing to do with the "either of them".

The second "it is" is the one that is influenced by the "either", so it also has to be "is" instead of "are". My sociology teacher in sophomore year in college made a big fuss over starting sentences (or parts of sentences) with "it" or "this", precisely because they cause confusion. To prevent the confusion, I suggested doing away with the "it" altogether.

> If either in or out is not be shown in the chart, the reason is that the metric is zero. 

Note that here we still have an "is" that follows the "either". What's implied is "the metric". A more verbose version is then:

> If either of the metrics 'in' or 'our' is not shown in the chart, the reason is that the metric is zero. 

P.S. I enjoyed this conversation, even though it has nothing to do with netdata :)

Xalaxis:
On the contrary, this is clearly the highest priority pull request netdata has ever received 😉 .  I've added your suggestion.

PR #5096, opened 2019-01-01
Add variables to alarm-notify.sh 

### LINKED ISSUE #946: alarm notifications should state a count of active alarms per state
It would be nice each notification email to have a count of all raised alarms (warning, critical).

### LINKED ISSUE #2351: False positive alerts and alert tuning
I am having lots of problems with false positive alerts (especially disk backlog), and while tweaking the tests to have different alert triggers helps, it doesn't resolve the problem.  It would be very convenient if there was a way to have the alert include something like:

This alert was triggered because the value ### exceeded the level ###.  If this is a false positive, consider updating the alert to allow these values as non-warning or critical: 

<test alert conditions that would have not considered the values to be an alert>

So, the alert gives instructions on how to adjust the test to not be an alert.  Am I making sense?

cakrit:
##### Summary
Fixes #946 
Fixes #2351 

##### Additional Information

Pass number of warning and critical alarms, evaluated expression and expression variable values. Use the new variables in the email notifications.
Doesn't quite do what was requested in #2351, but the values are available for use in any notification method we choose.

### DIFF
--- a/health/health.c
+++ b/health/health.c
@@ -153,7 +153,29 @@ static inline void health_alarm_execute(RRDHOST *host, ALARM_ENTRY *ae) {
     const char *exec      = (ae->exec)      ? ae->exec      : host->health_default_exec;
     const char *recipient = (ae->recipient) ? ae->recipient : host->health_default_recipient;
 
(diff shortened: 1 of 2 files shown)


Ferroin:
Overall LGTM.  I'll try and get this tested later today.

PR #5105, opened 2019-01-02
postgres : fix WAL query

anayrat:
##### Summary

Since previous refactor (d0de3356f3d83d273b2a6a19dfd5920b54a6f391) we forgot to remove extra bracket (needed to escape single bracket in original query). Due to this, the query returned 0.

This PR fix the query.

##### Component Name

Postgres

cc @ilyam8


ilyam8:
@anayrat 
thanks!

PR #5116, opened 2019-01-04
Clarify backend modes of operation

### LINKED ISSUE #5104: Anomalous (big) values on graphite/carbon
##### Bug report summary

Sending data to a graphite/(carbon) backend I got anomalous (very big) values

##### OS / Environment

Debian Stretch on both monitored server and carbon/graphite/grafana server; netadata installed from source, updated with ./netdata-updater.sh, version v1.12.0-rc1-34-gd5fb2be8 (but I got the same with 1.11). graphite-carbon installed from official debian packages (0.9.15-1), grafana 5.4.2 from debian packages on grafana repository.

##### Component Name

backends

##### Steps To Reproduce

Not sure if it's a bug or a misconfiguration on my side,  I tryed to follow the docs but I'm not so deep in to carbon, grafana and netdata.

This my netdata.conf configuration, the only changes parts from installed default are the following:

```
[registry]
registry to announce = http://remote.host.istance:19999
[backend]
        enabled = yes
        type = graphite
        destination = tcp:remote.carbon.instance:2003
        data source = as collected
        update every = 60
```

this is the result in grafana using carbon as datasource for the cpu load1 and load5 average:

grafanacpuload

but also looking at the data I got:

```
~# whisper-info /var/lib/graphite/whisper/netdata/haynes/system/load/load1.wsp 
maxRetention: 86400
xFilesFactor: 0.5
aggregationMethod: average
fileSize: 17308

Archive 0
retention: 86400
secondsPerPoint: 60
points: 1440
size: 17280
offset: 28
``` 

and:

```
~# whisper-fetch /var/lib/graphite/whisper/netdata/haynes/system/load/load1.wsp
 ...
1546447500	880.000000
1546447560	800.000000
1546447620	580.000000
1546447680	1320.000000
1546447740	1250.000000
1546447800	1110.000000
1546447860	640.000000
1546447920	350.000000
1546447980	330.000000
1546448040	None

```
so getting thousand for  a server with normal load:

##### Expected behavior

I'd expected to have values similar to the one showed by netdata:

netdata

Regards
Simone

cakrit:
##### Summary

Fixes #5104

##### Component Name

Docs

### DIFF
--- a/backends/README.md
+++ b/backends/README.md
@@ -36,18 +36,20 @@ X seconds (though, it can send them per second if you need it to).
 
 3. Netdata can filter metrics (at the chart level), to send only a subset of the collected metrics.
 
-4. Three modes of operation (for all backends):
+4. Netdata supports three modes of operation for all backends. **If you can't decide which mode to use, 
+   we suggest `as collected` for Prometheus and `average` for all other backends**
 
    - `as collected`: the latest collected value is sent to the backend. This means that if netdata
      is configured to send data to the backend every 10 seconds, only 1 out of 10 values will appear
-     at the backend server. The values are sent exactly as collected, before any multipliers or
-     dividers applied and before any interpolation. This mode emulates other data collectors,
-     such as `collectd` or `telegraf`.
+     at the backend server. The values are sent exactly as collected, i.e. raw, unprocessed values, 
+     before any multipliers or dividers applied and without any interpolation/calculation of average 
+     values etc. As a result, the values may not match what you would see via the Netdata UI. This mode 
+     emulates other data collectors, such as `collectd` or `telegraf` and is the suggested method for a 
+     Prometheus backend.
 
    - `average`: the average of the interpolated values shown on the netdata graphs is sent to the
      backend. So, if netdata is configured to send data to the backend server every 10 seconds,
-     the average of the 10 values shown on the netdata charts will be used. **If you can't decide
-      which mode to use, use `average`.**
+     the average of the 10 values shown on the netdata charts will be used. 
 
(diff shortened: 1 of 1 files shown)


backends/README.md
```diff
@@ -36,18 +36,20 @@ X seconds (though, it can send them per second if you need it to).
 
 3. Netdata can filter metrics (at the chart level), to send only a subset of the collected metrics.
 
-4. Three modes of operation (for all backends):
+4. Netdata supports three modes of operation for all backends. **If you can't decide which mode to use,
```
ktsaou: I think we should suggest this:

- `as-collected` sends to backends the metrics as they are collected, in the units they are collected. So, counters are sent as counters and gauges are sent as gauges, much like all data collectors do. For example, to calculate CPU utilization in this format, you need to know how to convert kernel ticks to percentage.

- `average` sends to backends normalized metrics from the netdata database. In this mode, all metrics are sent as gauges, in the units netdata uses. This abstracts data collection and simplifies visualization, but you will not be able to copy and paste queries from other sources to convert units. For example, CPU utilization percentage is calculated by netdata, so netdata will convert ticks to percentage and sent the average percentage to the backend.

Time-series databases suggest to collect the raw values (`as-collected`).

If you plan to invest on building your monitoring around a time-series database, and you already know (or you will invest in learning) how to convert units and normalize the metrics at Grafana or other visualization tools, we suggest to use `as-collected`.

If, on the other hand, you just need long term archiving of netdata metrics, and you plan to mainly work with netdata, we suggest to use `average`. It decouples visualization from data collection, so it will generally be a lot simpler.

Issue #5133, opened 2019-01-08
python go_expvar: reuse same expvar key in different charts


liepumartins:
I have a go program with several incrementing expvars, I would like to chart both incremental and absolute values. Goal is to get and show total number of operations and operations per minute and operations per hour.

However, if I try to add different chart using the same expvar, this happens `INFO: go_expvar: myapp: Duplicate expvar key counters.downloaded: skipping line.` 
All the other ids are unique.

```
...
- id: "counters_down"
      options:
        name: counters
        title: "downloaded"
        units: items
        family: counters
        context: expvar.foo.counters
        chart_type: line
      lines:
        - {expvar_key: 'counters.downloaded_orders', expvar_type: int, id: counters_downloaded, name: 'Items'}
- id: "counters_down_rate"
      options:
        name: counters
        title: "downloaded per second"
        units: items
        family: counters
        context: expvar.foo.counters
        chart_type: line
      lines:
        - {expvar_key: 'counters.downloaded_orders', expvar_type: int, id: counters_downloaded_rate, name: 'Items/s', algorithm: incremental}
...
```

ilyam8:
Hi @liepumartins 

The problem is 

https://github.com/netdata/netdata/blob/36ae17db06d434375ab6a5104dbd04506c2e22cb/collectors/python.d.plugin/go_expvar/go_expvar.chart.py#L160-L162

I think it is safe to remove this check.

Please remove it and see if the issue is fixed.

ilyam8:
@liepumartins fix from my prev post won't work

give me some time

ilyam8:
i read `go_expvar` code and it seems that https://github.com/netdata/netdata/issues/5133#issuecomment-452343559 will do

so please test it

liepumartins:
This will not help, because next line is `self.expvars[ev_key] = (ev_type, line_id)`
So the last line definition using particular ev_key will overwrite all previous.

I tried and it happens as I described - instead of first chart, now I get last one.

It appears that by design this is impossible because, each key is bound to exactly one line https://github.com/netdata/netdata/blob/36ae17db06d434375ab6a5104dbd04506c2e22cb/collectors/python.d.plugin/go_expvar/go_expvar.chart.py#L164

ilyam8:
Yes, you are right. I am working on it :+1:

PR #5141, opened 2019-01-09
go_expavar fix: don't check for duplicate expvars

ilyam8:
##### Summary

Fixes: #5133

##### Component Name
https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/go_expvar
##### Additional Information

It's ok to have dimension with same id in different charts, this check is not needed.

@liepumartins please test


ilyam8:
Nice @liepumartins 
thx for reporting!

Issue #2401, opened 2017-06-28
collect CPU, memory and disk limits from cgroups and provide alarms using thresholds as percentages


sn-x:
In file src/sys_fs_cgroup.c, lines  2089, 2120

We are currently displaying all available cores on the host.
Would it be more appropriate if we only count the cores assigned to the guest?

ktsaou:
I guess you mean this part: https://github.com/ktsaou/netdata/blob/fbd2e582b7708b2bee58432c793b9a7b100a9bc8/src/sys_fs_cgroup.c#L2115-L2148

Well, if only certain cores are assigned to the guest, then only these cores should have non-zero values. The way it is now, it wastes some memory in this case (for the unused cores), but the code is a lot simpler and faster. For the end user there should not be any difference. The dashboard hides by default all dimensions that are just zero.

sn-x:
Hi! Thank you for the reply.

You are correct, that is the graph I was talking about and also the graph above it.
![cgroup cpu graphs](https://user-images.githubusercontent.com/15047082/27673707-6a6a5052-5ca2-11e7-8f0e-28732b839bb8.png)

A side question - is it currently possible to create cpu alerts on cgroups? If my guest has 4 cores (400%) and usage is above 320% (80%)? How would this be accomplished?

ktsaou:
Something like this:

```
 alarm: high_cpu_on_my_vm
    on: cgroup_lxc_XXXXX.cpu
lookup: average -10s unaligned
 every: 10s
  warn: $this > 300
  crit: $this > 380
  info: my container uses too much cpu
    to: sysadmin
```


The `lookup` line calculates the last 10 seconds average of the sum of all dimensions of the chart and assigns the result to `$this`. The rest I guess are self explained.

sn-x:
ktsao, you are doing the gods works. I am very impressed with netdata. I value your work and all the effort you have put in! One feature that I like very much are alerts with percentile thresholds. This pretty much eliminates user error (my error).

Consider this:
 - if we know that a namespace has 8G of memory available, we can reuse the percentile thold - 80%
 - if we know that a namespace has 4 cores available, we can reuse the percentile thold - 80%
Even if we increase/decrease allocated resources, tholds can stay the same, which would be a very big plus.

Like I said, I appreciate everything you have done! Maybe this can be treated as a proposal for improvement.

ktsaou:
You are right. This is good.
Let's see how we can find this info...

ktsaou:
[...]

PR #5172, opened 2019-01-14
Add cgroup cpu and memory limits and alarms

vlvkobal:
##### Summary
Add limit variables to cgroup charts. Provide alarms for limited resources. Add charts for resource usage within cgroup limits.
- [x] CPU
- [x] memory

Partly implements #2401

##### Component Name
`cgroups` plugin

##### Additional Information
Disk limits can't be implemented on the existing charts. If we decide to implement them, it will be done in another PR.

### DIFF
--- a/collectors/cgroups.plugin/README.md
+++ b/collectors/cgroups.plugin/README.md
@@ -89,6 +89,10 @@ The whole point for the additional pattern list, is to limit the number of times
 
 The above pattern list is matched against the path of the cgroup. For matched cgroups, netdata calls the script [cgroup-name.sh](cgroup-name.sh.in) to get its name. This script queries `docker`, or applies heuristics to find give a name for the cgroup.
 
+### alarms
+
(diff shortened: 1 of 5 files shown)


vlvkobal:
I don't use any data from Docker. Look at  [CFS Bandwidth Control](https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt). There are two files to control CPU limits for cgroups: `cpu.cfs_period_us` and `cpu.cfs_quota_us`.

What is "dial"? Is this a type of chart?

vlvkobal:
I tested it with `--cpus=` option. It works. Actually, docker changes the same cgroup files: `cpu.cfs_period_us` and `cpu.cfs_quota_us` when `--cpus=` is set.

cakrit:
Just tested the last changes, it looks good to me.

ktsaou:
Please make sure it is tested under valgrind

Issue #4563, opened 2018-11-05
How to setup the tc plugin 


ryanmatthieu:
##### Question summary
Hello ,

I am using netdata on a project for school about Quality of Service. The tool is working great but I have an issue on how to setup the tc plugin (https://github.com/netdata/netdata/tree/master/collectors/tc.plugin)

I tried to run the **tc-qos-helper.sh.in** script but it kept showing me this message : 

```
2018-11-04 22:59:56: tc-qos-helper.sh.in: WARNING: FireQoS is not installed on this system. Use FireQoS to apply traffic QoS and expose the class names to netdata. Check https://github.com/netdata/netdata/wiki/You-should-install-QoS-on-all-your-servers
2018-11-04 22:59:56: tc-qos-helper.sh.in: WARNING: Cannot find file '@libconfigdir_POST@/tc-qos-helper.conf'.
2018-11-04 22:59:56: tc-qos-helper.sh.in: WARNING: Cannot find file '@configdir_POST@/tc-qos-helper.conf'.
/home/server/tc-qos-helper.sh.in: line 175: /home/server/loopsleepms.sh.inc: No such file or directory
```
And no charts have been added to the netdata web interface .
I am not using fireQoS but the native tc tools , am I doing something wrong ?

How should I set it up so it shows charts with _**tc.world_in**_ and _**tc.world_out**_ as shown on the demo of the plugin ?

Sorry if I misread or misunderstood anything and many thanks in advance for your help .

##### OS / Environment
lubuntu-16.04.3-desktop-amd64
Using a VM on Virtual Box 

##### Component Name
tc.plugin

##### Expected results
Show charts for tc metrics

ktsaou:
Hi, there is not much to configure for `tc` plugin.
Just install netdata and start it up.

If you have QoS running (no matter if it is configured by FireQoS or not), netdata should be able to pick it up and show it.

To configure class names, you will have to edit `/etc/iproute2/tc_cls`. With FireQoS this is not needed (the plugin picks up the names set by FireQoS).

ktsaou:
You can create `etc/iproute2/tc_cls`. This is just to pick up the names.

Can you post the output of:

```sh
/usr/libexec/netdata/plugins.d/tc-qos-helper.sh
```

Leave it running for 2-3 seconds and stop it with Control-C.

ryanmatthieu:
This was the output : 
```
[...]

PR #5221, opened 2019-01-19
Improvements to QoS (tc) documentation

cakrit:
##### Summary
Add configuration method provided by @laneshields at #4563 
Wording improvements.
Move additional example chart further up

##### Component Name
Docs

### DIFF
--- a/collectors/tc.plugin/README.md
+++ b/collectors/tc.plugin/README.md
@@ -163,68 +163,11 @@ And this is what you are going to get:
 
 ![image](https://cloud.githubusercontent.com/assets/2662304/14436322/c91d90a4-0024-11e6-9fb1-57cdef1580df.png)
 
-## QoS Configuration example with tc
+## QoS Configuration with tc
 
-_Example provided in [github issue #4563](https://github.com/netdata/netdata/issues/4563#issuecomment-455711973)_
+First, setup the tc rules in rc.local using commands to assign different DSCP markings to different classids. You can see one such example in [github issue #4563](https://github.com/netdata/netdata/issues/4563#issuecomment-455711973). 
 
-First, setup the tc rules in rc.local using the following commands to assign different DSCP markings to different classids In the particular example, `netem` is used on these interfaces to inject impairments, which is why the root `qdisc` is created that way:
-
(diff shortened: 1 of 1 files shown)


collectors/tc.plugin/README.md
```diff
@@ -166,16 +163,89 @@ And this is what you are going to get:
...
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0x28 0xff classid 2:10
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0x40 0xff classid 2:16
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0x48 0xff classid 2:18
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0x60 0xff classid 2:24
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0x68 0xff classid 2:26
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0x80 0xff classid 2:32
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0x88 0xff classid 2:34
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0xa0 0xff classid 2:40
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0xb8 0xff classid 2:46
+tc filter add dev eth6 protocol ip parent 2: prio 1 u32 match ip dsfield 0xc0 0xff classid 2:48
```
ktsaou: I don't like this part. We are giving a specific QoS ruleset to users, about some `eth5` and `eth6` network interfaces, with static rates and bursts. Why do we need this?

I have written FireQoS to avoid exactly that.

cakrit: That's why it's the easy way and the hard way. Someone already added such instructions in the issue I mentioned in the documentation (#4563). I could show just ~5 lines instead of the whole thing. What do you think?

cakrit: I modified it, please check again.

Issue #4028, opened 2018-08-02
dashboard_info.js underscore in json keys not working 


XFabian:
Hi
 I created a custom_dasboard_info.js file and included it in the netdata conf and this works.
But when I include my custom plugin called web_perf like this:
``
'web_perf': {
     title: undefined,
     icon: '<i class="fas fa-eye"></i>',
     info:'Blablabla my info here' 
}
``
This does not change anything. The chart.py and conf file are also called web_perf.
When I change the name chart.py and conf file to webPerf and change the key name to webPerf it starts working.
I glanced at some JSON posts and did not find an entry that underscores are not allowed in JSON keys.

XFabian:
Ok I just saw the web_log entry so I don't think it is the underscore in the key name.
But the name web_perf is still not working.

cakrit:
Closing due to inactivity. @XFabian if you're able to provide more info, please do so and I'll reopen it.

vlvkobal:
Underscore for keys in `netdataDashboard.menu` is treated as a special character. Menu info is working for me only if I don't use special symbols in keys. If I understand it correctly, the `web_log` entry works because it has submenus defined in `netdataDashboard.submenu`.

cakrit:
We found the culprit, it's the logic of 'enrichChartData', at https://github.com/netdata/netdata/blob/master/web/gui/main.js#L1161

The code itself refers to #807, which will be a breaking change left for v2.0. 
@gmosx : As we're building a completely new UI, I expect that we'll face the same issue there, unless if you've found a better way of handling the various complex ways charts are named currently.

gmosx:
In the new code I will support whatever is implemented in the scope of #807. For the moment I am more concentrated in the layout engine, not the actual charts.

cakrit:
Let's keep this as feature request so that we don't forget to test this in the new code.

ilyam8:
Closing this feature request. If you are still interested in it, consider creating a topic on [our community forum](https://community.netdata.cloud/c/feature-requests/agent-fr/21).

PR #5233, opened 2019-01-21
Mail log plugin for python.d

apardyl:
##### Summary
Implement a new python.d plugin for analyzing mail server logs.

##### Component Name
collectors/python.d.plugin/mail_log

##### Additional Information
Currently MTA monitoring in Netdata is limited to plotting queue size for postfix. By parsing the system mail log (`/var/log/mail.log`) we can monitor a larger number of parameters:
* received email count by status (accepted, soft-rejected, rejected, greylisted) and response code (https://www.iana.org/assignments/smtp-enhanced-status-codes/smtp-enhanced-status-codes.xhtml),
* email delivery status (sent, deferred, rejected) and response code,
* number of incoming connections.

screenshot from 2019-02-19 02-09-28


Ferroin:
I think this is a great addition to Netdata, it provides information that's extremely useful for people running mail servers, and it looks like it does so pretty efficiently.

collectors/python.d.plugin/mail_log/mail_log.chart.py
```diff
@@ -0,0 +1,207 @@
...
+    # outgoing
+    outgoing_codes = {
+        "2.0.0": 0
+    }
+    sent = 0
+    deferred = 0
+    bounced = 0
+
+
+class MailParser:
```
Ferroin: You need to have this class inherit from `object`, otherwise things may not work correctly on Python 2.x (fixing this will also get rid of the new alert reported by LGTM).

ilyam8:
> and it looks like it does so pretty efficiently.

i see it uses 2-3 regex searches per line.

apardyl:
Sorry for the delay (I have been very busy through last 3 weeks), I finished work on this modules - it should work with any postfix server with or without plugable spam filters. I chose not to implement exim4 support at this moment as I have no exim logs for testing.

One thing I still cannot get to work - the chart icon and info in UI - I've added an entry to `web/gui/dashboard_info.js` , but for an unknown to me reason it is ignored. Other than that this PR is ready.

ilyam8:
@apardyl 

i see 2 problems:
  - module is hard to read, i think we could simplify it
  - 2-3 regex search per line, this is room for optimization, if it is possible we should avoid it

To begin with please post here some mail logs, i would like to understand parse logic and how it could be optimized

apardyl:
@ilyam8 
> To begin with please post here some mail logs, i would like to understand parse logic and how it could be optimized

Example logs from a production server with comments (names and addresses redacted):
https://gist.github.com/apardyl/a300dd3ef1e490c26d6754524c4faa1d

> * 2-3 regex search per line, this is room for optimization, if it is possible we should avoid it

There at most 2 regexes per line:
* the first `POSTFIX_RE` matches postfix log lines (the syslog mail.log file usually contains logs from other applications, e.g. mailbox server or spam filter) and extracts postfix module name and the very next following word (enough to determine the type of the event and get the message id if applicable)
* the second one filters out informations relevant for given module (`QUEUE_RE`, `REJECT_RE`, `STATUS_RE`)

This can be handled with a single complex regex but for readability I decided to split it per line type (performance-wise the difference is negligible, especially compared to the fact, that a single mail delivery operation often requires several LDAP and DNS lookups and antivirus + antispam scans).

apardyl:
I've found the cause of the dashboard_info issue - this plugin has an underscore in its name  #4028. I can either add its name here https://github.com/netdata/netdata/blob/master/web/gui/main.js#L1374 or rename it to `maillog` or something else without special characters.

ilyam8:
@apardyl 

This charts
https://github.com/apardyl/netdata/blob/b76acf73084c956524430d11379cf9ceb00b85b3/collectors/python.d.plugin/mail_log/mail_log.chart.py#L13-L58

Are they postfix specific or not?

https://github.com/apardyl/netdata/blob/b76acf73084c956524430d11379cf9ceb00b85b3/collectors/python.d.plugin/mail_log/mail_log.chart.py#L169-L174

And how about to avoid 
https://github.com/apardyl/netdata/blob/b76acf73084c956524430d11379cf9ceb00b85b3/collectors/python.d.plugin/mail_log/mail_log.chart.py#L183

It is really hard to read/understand such code.

Can we make parser only responsible for parse? Input - lines, output - stats. I see you add updating chart logic to the base `MailParser` class.

ilyam8:
btw check if the module works with py2

apardyl:
@ilyam8 sorry for not replying again :)

Those charts you asked about apply to every mail server (MTA) - basically they display incoming and outgoing mail trafic by status and ESMTP status code.

I refactored the parts of code you highlighted and moved parser choice to config (my previous idea was to run parsers for all server types (only postfix for now) every time). I hope you find it more readable now.

I also checked this module under python2.7 (`python2 /usr/libexec/netdata/plugins.d/python.d.plugin mail_log debug'), everything seems to be working correctly after fixing one python3-style `super()`.

apardyl:
Hey @ilyam8, could you take a look at this PR?

ilyam8:
@apardyl sorry for the delay, somehow i missed your request :cry: will do this week

ilyam8:
@apardyl 

Main question

Are these statistics and charts suitable for other system mail logs?

https://github.com/netdata/netdata/blob/a41d1380d1a50679bef902399b37d5aba88ff384/collectors/python.d.plugin/mail_log/mail_log.chart.py#L61-L77

https://github.com/netdata/netdata/blob/a41d1380d1a50679bef902399b37d5aba88ff384/collectors/python.d.plugin/mail_log/mail_log.chart.py#L13-L58

If they (or at least part of them) are specific to postfix (which is very likely i think) we need to rename the collector to `postfix_mail_log`.

Ferroin:
@ilyam8 At least in theory, they're pretty generic.  The exact method of collection, however, is entirely specific to Postfix.  It might be possible to adapt it to work with other mail systems, but I don't know how easy that may be.

ilyam8:
> the exact method of collection, however, is entirely specific to Postfix. 

This is ok, i was asking about charts/metrics. Is metrics set is the same for other mail systems?

https://github.com/netdata/netdata/blob/a41d1380d1a50679bef902399b37d5aba88ff384/collectors/python.d.plugin/mail_log/mail_log.chart.py#L61-L77

Ferroin:
Ah, OK.  It should in theory be possible to get mostly equivalent metrics (maybe not the same exact names, but functionally the same numbers) out of most other full MTA platforms, I'm just not certain how one might go about doing so (I've never really used any full-fledged MTA other than Postfix).

apardyl:
Those metrics are standard for any ESMTP-compliant MTA (any production-ready MTA server), except for greylisting, which is common spam filtering technique implemented by many spam filters.

Code and status standard doc: https://www.iana.org/assignments/smtp-enhanced-status-codes/smtp-enhanced-status-codes.xhtml

collectors/python.d.plugin/mail_log/mail_log.chart.py
```diff
@@ -0,0 +1,234 @@
...
+    temp_rejected = 0
+    perm_rejected = 0
+    greylisted = 0
+    # outgoing
+    outgoing_codes = {
+        "2.0.0": 0
+    }
+    sent = 0
+    deferred = 0
+    bounced = 0
```
ilyam8: These should be instance variables, not class

collectors/python.d.plugin/mail_log/mail_log.chart.py
```diff
@@ -0,0 +1,234 @@
...
+    """Abstract MailParser class, mail log parsers must implement the _parse_line method"""
+
+    def __init__(self, log_service):
+        self.service = log_service
+
+    def _parse_line(self, line):
+        raise NotImplementedError
+
+    def parse_line(self, line):
+        self._parse_line(line)
```
ilyam8: i think you can add `parse_lines`

collectors/python.d.plugin/mail_log/mail_log.chart.py
```diff
@@ -0,0 +1,234 @@
...
+    sent = 0
+    deferred = 0
+    bounced = 0
+
+
+class MailParser(object):
+    """Abstract MailParser class, mail log parsers must implement the _parse_line method"""
+
+    def __init__(self, log_service):
+        self.service = log_service
```
ilyam8: This class should have `MailStatistics` since

> Those metrics are standard for any ESMTP-compliant MTA (any production-ready MTA server)

collectors/python.d.plugin/mail_log/mail_log.chart.py
```diff
@@ -0,0 +1,234 @@
...
+    QUEUE_RE = re.compile(r'from=<.*\(queue active\)')
+    REJECT_RE = re.compile(r'reject: [^:]+: (\d{3} )?(\d+\.\d+\.\d+)')
+    STATUS_RE = re.compile(r'dsn=(\d+\.\d+\.\d+).*status=(\w+)')
+
+    def __init__(self, log_service):
+        super(PostfixParser, self).__init__(log_service)
+        self.greylist_status = log_service.configuration.get("greylist_status", "Try again later")
+        # Since postfix does not log messages passing milters (only rejected ones) we store last 50 ids of messages
+        # received by the `cleanup` process and count them as accepted when queued for delivery for the first time.
+        self.queue_ids = collections.OrderedDict()
```
ilyam8: Please dont pass to parser class Service class, if the parser need some config params pass only them.I see it only need `greylist_status`

I think this class should be only resposnible for parsing and populating statistics, i suggest to remove other logic from it. It will help to keep things simple.

collectors/python.d.plugin/mail_log/mail_log.chart.py
```diff
@@ -0,0 +1,234 @@
...
+        2. "log_path" must must exist and be readable by netdata user
+        """
+        if not self.log_path:
+            self.error('log path is not specified')
+            return False
+
+        if not (self._find_recent_log_file() and os.access(self.log_path, os.R_OK)):
+            self.error('{log_file} not readable or does not exist'.format(log_file=self.log_path))
+            return False
+        return True
```
ilyam8: We need to check if parser is not `None` in `check`. I suggest to move parser init in `check` and log a meaningful error if the type is wrong.

And we need to ensure if the log file actually contains logs of chosen type (postfix, etc.). Not sure how to achieve this. Any ideas?

Ferroin: Checking for the presence of matching log lines will give an insanely high false negative rate for whether this should run or not.  Just because someone is running Postfix (or whatever other MTA) does not mean that there will be lines in the mail log from that MTA at any arbitrary point in time (think log rotation).

This is a case that I think we should just rely on the behavior of charts that have never had non-zero values being hidden by default, or just not enable the module by default and let users enable it if they want it.

collectors/python.d.plugin/mail_log/mail_log.chart.py
```diff
@@ -0,0 +1,234 @@
...
+        except (ValueError, AttributeError) as ex:
+            self.error(ex)
+            return None
+
+    def increment_incoming_code(self, code):
+        """Increments number of incoming messages per status code, adds a new line to chart if required"""
+        if code not in self.stats.incoming_codes:
+            self.charts["incoming_codes"].add_dimension(['mail_incoming_codes_' + code, code, 'incremental', 1, 1])
+            self.stats.incoming_codes[code] = 0
+        self.stats.incoming_codes[code] += 1
```
ilyam8: This should be `update_incoming_codes_chart` and `update_outgoing_codes_chart` methods and this should be handles by `Service`, not parser.

ilyam8:
In general i suggest to use less python magic, i would like to avoid passing `Service` class to `Parser` and manipulate `Service` attrs/methods from there.

collectors/python.d.plugin/mail_log/mail_log.chart.py
ilyam8: If i understand it right we cant have default parser here. if there is no way to determine MTA we should run the collector, so we cant have it enabled by default. 

> just rely on the behavior of charts that have never had non-zero values being hidden by default

Hm, in this case the module it will silently run forever.

Ferroin: The issue I'm worried about is a case where Netdata starts up right after the log files got rotated.  Unless the system deals with lots of mail all the time, it's not unreasonable to expect that there might be no lines whatsoever from the MTA in the log file for quite some time, in which case the module wouldn't start up even though it should be running.

Perhaps disable the module by default, and maybe add a special warning that gets logged if it finds no lines from an MTA in some number of collection cycles (probably 3600 or more)?

ilyam8: Oh, i didnt know, so all MTAs write to same file? :see_no_evil: 	

> if it finds no lines from an MTA in some number of collection cycles (probably 3600 or more)?

Well, if the module failes to get data python.d will increase it update every up to 10min gradually. And it will never kill the job (current implementation/strategy - if the job passes check never kill it even if it failed to collect data over 9k times). 

https://github.com/netdata/netdata/blob/adb7026b14fea7bbebee93c8958d92ee9cf1efaf/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SimpleService.py#L52-L55

This is bad because the module will have to read a lot of lines after 10 minutes delay :sweat_smile:

ilyam8: My main concern that it will be confusing/wrong to have zero postfix mail log metrics if user has no postfix installed (current implementation).

Ferroin: > Oh, i didnt know, so all MTAs write to same file? 🙈

As a general rule, yes, all MTA's (as well as all mail filtering tools and pretty much anything else that doesn't run as a normal user and handles mail) write logs via the syslog `mail` facility, which in standard setups goes to a single file (typically `/var/log/mail.log`).  That's the reason that there's a need to match on the lines being from Postfix itself.

Realistically though, you will only have one MTA running on a system, plus a handful of mail filters.  In general for such a case, you won't git many (if any) log lines when mail is not being processed, so short of this running when the local MTA is something it doesn't support reading log entries from, you shouldn't see much if any slowdown when it fails to collect data.

> My main concern that it will be confusing/wrong to have zero postfix mail log metrics if user has no postfix installed (current implementation).

Agreed, so it sounds like default disabled is probably the best option here.

PR #5254, opened 2019-01-23
Additions for repeat notifications feature

npapak:
##### Summary
These are some changes for repeat notification feature which described in Issue #3963 . 
Now you can set a filed in netdata.conf file under health block and set a global notification repeat interval in seconds.

E.g.:
```
[health]
...
repeat notification alarms every seconds = 10
```

The default value is 0 and the notifications  raised one time.

In addition, you can add a field in any configuration file under health.d directory which describes the interval in seconds of notification repetition for this alarm only.

E.g.:
```
template: disk_space_usage
...
repeat: 10
```

##### Component Name
Notification repeat.


cakrit:
I'll do more tests once the changes I've requested are done.

npapak:
I made the requested changes and fixes from @cakrit . 
Furthermore, I change the configuration representation for global setting and for custom setting.
Now the configurations must have the following format:

```
[health]
...
repeat notification alarms every = 10s
```

The default value is 0 and the notifications raised one time.
Now you can set the interval in seconds(s), minutes(m), hours(h), days(d), weeks(w), months(M) and years(Y).

For the custom alarm configuration you have to add on the config file the entry `repeat: time_interval>`.
The interval should have the same format as in global configuration (seconds(s), minutes(m), hours(h), days(d), weeks(w), months(M) and years(Y)).

E.g.:
```
template: disk_space_usage
...
repeat: 10s
```

In addition, you can see the repeat interval and the number of the alarm notifications of each alarm in `http://your.netdata:19999/api/v1/alarms?all` .

libnetdata/config/appconfig.c
```diff
@@ -280,6 +280,54 @@ long long appconfig_get_number(struct config *root, const char *section, const c
     return strtoll(s, NULL, 0);
 }
 
+int appconfig_get_duration(struct config *root, const char *section, const char *name, char * value)
```
cakrit: Let's not have the exact same code as health_parse_duration copied. Please have a common function that can be called both from here and from health_config.c

cakrit: Put `parse_duration` in `appconfig.c` as well, at least that way we have things in two files instead of three.

cakrit: Thought I replied but I can't see it here, better put `parse_duration` in `appconfig.c` as well. At least we only use two files this way.

PR #5257, opened 2019-01-24
Increase Chinese support

Fhaohaizi:
I added “dashboard_inf_zh.js” for Increase Chinese support.Chinese support for major indicators has been added and will continue to improve in the future.

### DIFF
--- a/netdata-installer-zh.sh
+++ b/netdata-installer-zh.sh
@@ -2,25 +2,8 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # shellcheck disable=SC1090,SC1091,SC1117,SC2002,SC2034,SC2044,SC2046,SC2086,SC2129,SC2162,SC2166,SC2181
 
-
-export PATH="${PATH}:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
-uniquepath() {
-	local path=""
-	while read; do
-		if [[ ! ${path} =~ (^|:)"${REPLY}"(:|$) ]]; then
-			[ ! -z "${path}" ] && path="${path}:"
-			path="${path}${REPLY}"
-		fi
-	done < <(echo "${PATH}" | tr ":" "\n")
-
-	[ ! -z "${path}" ] && [[ ${PATH} =~ /bin ]] && [[ ${PATH} =~ /sbin ]] && export PATH="${path}"
-}
-uniquepath
-
 netdata_source_dir="$(pwd)"
 
 rm ${netdata_source_dir}/web/gui/dashboard_info.js
-echo "${netdata_source_dir}/web/gui/dashboard_info.js被删除了"
 cp ${netdata_source_dir}/web/gui/dashboard_info_zh.js ${netdata_source_dir}/web/gui/dashboard_info.js
-echo "${netdata_source_dir}/web/gui/dashboard_info.js被复制了"
 ./netdata-installer.sh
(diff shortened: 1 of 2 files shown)


Fhaohaizi:
> Codacy Here is an overview of what got changed by this pull request:
> 
> ```diff
> Issues
> ======
> - Added 29
>            
> ```
> 
> See the complete overview on [Codacy](https://app.codacy.com/app/netdata/netdata/pullRequest?prid=2942133&bid=10889326)

I just translated some descriptive text into Chinese.Some of the questions you have raised, I don't know how to modify them. However, during my testing, the netdata software can be installed normally.I just want to replace the text in the data description to make Chinese users better use netdata.

Ferroin:
I'm not sure the approach taken here with the installation is the best option for a couple of reasons:

* Undoing what it does requires a clean re-install.
* It doesn't look to me like the switch to Chinese will persist through updates.
* It won't work with any kind of proper packaging (you'll need a separate package for each language).
* It forces a specific static language.
* It assumes that the translated copy will be kept in-sync with the original English version.

I think a much better approach would be to handle this completely in the client-side dashboard code.  That way, we can:

* Avoid any need to mess with file names pre-install (which solves the first three issues).
* Actually pay attention to the user's preferred language list as reported by the browser to select a default language for the dashboard.
* Provide a way in the dashboard itself to switch the display language.
* Fill in any parts that are missing from the translated version with the original English text without having to make the whole translated file contain all the English text that isn't yet translated (and optionally display a notice that some parts of the page aren't translated if the translation doesn't cover everything in the original English version).

----

As far as the Codacy stuff, I'm not sure what to say about that, I _think_ it's just getting confused here, but I'm not sure.

Fhaohaizi:
> I'm not sure the approach taken here with the installation is the best option for a couple of reasons:
> 
> * Undoing what it does requires a clean re-install.
> * It doesn't look to me like the switch to Chinese will persist through updates.
> * It won't work with any kind of proper packaging (you'll need a separate package for each language).
> * It forces a specific static language.
> * It assumes that the translated copy will be kept in-sync with the original English version.
> 
> I think a much better approach would be to handle this completely in the client-side dashboard code. That way, we can:
> 
> * Avoid any need to mess with file names pre-install (which solves the first three issues).
> * Actually pay attention to the user's preferred language list as reported by the browser to select a default language for the dashboard.
> * Provide a way in the dashboard itself to switch the display language.
> * Fill in any parts that are missing from the translated version with the original English text without having to make the whole translated file contain all the English text that isn't yet translated (and optionally display a notice that some parts of the page aren't translated if the translation doesn't cover everything in the original English version).
> 
> As far as the Codacy stuff, I'm not sure what to say about that, I _think_ it's just getting confused here, but I'm not sure.

thanks。

> I'm not sure the approach taken here with the installation is the best option for a couple of reasons:
> 
> * Undoing what it does requires a clean re-install.
> * It doesn't look to me like the switch to Chinese will persist through updates.
> * It won't work with any kind of proper packaging (you'll need a separate package for each language).
> * It forces a specific static language.
> * It assumes that the translated copy will be kept in-sync with the original English version.
> 
> I think a much better approach would be to handle this completely in the client-side dashboard code. That way, we can:
> 
> * Avoid any need to mess with file names pre-install (which solves the first three issues).
> * Actually pay attention to the user's preferred language list as reported by the browser to select a default language for the dashboard.
> * Provide a way in the dashboard itself to switch the display language.
> * Fill in any parts that are missing from the translated version with the original English text without having to make the whole translated file contain all the English text that isn't yet translated (and optionally display a notice that some parts of the page aren't translated if the translation doesn't cover everything in the original English version).
> 
> As far as the Codacy stuff, I'm not sure what to say about that, I _think_ it's just getting confused here, but I'm not sure.

I thought about this simple.thanks，your，replay。

Fhaohaizi:
> I disagree with creating one more installer file just so we have another language support. This is creating more harm than there is benefit from it.



> I disagree with creating one more installer file just so we have another language support. This is creating more harm than there is benefit from it.

limited ability.I thought about this simple.thank your replay。

Fhaohaizi:
水平有限，妄自提交。希望netdata早些支持多语言。

PR #5282, opened 2019-01-28
Database engine

### LINKED ISSUE #4687: new database format design
##### Feature idea summary

netdata internal database v2.
Specs to follow.

### LINKED ISSUE #5303: Prototype for netdata internal database v2.0
##### Feature idea summary

netdata internal database v2.0
The first step towards #4687

##### Expected behavior

New database storage engine that is based on data files that contain metrics and journal files that contain metadata. No more mmap, it will be based on asynchronous I/O and on-disk compression. It will optimize the I/O path to be viable even on rotating media. This will also allow the database memory footprint to be smaller than the database size on disk.

mfundul:
##### Summary
netdata internal database v2.0 Fixes #5303. Work In Progress.
##### Component Name
[/database](https://github.com/netdata/netdata/tree/master/database)
##### Additional Information
Implements the new database format design.
- [X] on-disk meta-data protocol
- [X] mock page-cache
- [X] database asynchronous engine
- [X] data file implementation
- [x] mock database skeleton utilizing full path from API to disk
- [x] journal file implementation
- [x] scalable page-cache implementation
- [x] metrics index
- [x] space management
- [ ] metrics map
- [ ] metrics map log file implementation

Plan for Minimum Viable Product:
1. finalize DB API and modify netdata to support pages for data collection and queries
2. implement scalable page cache, metrics index and space management
3. map metrics to UUIDs deterministically by a string ID->UUID predictable transformation
4. keep netdata as is, only collected metrics are stored and retrieved in the DB
5. no upgrade necessary, previously stored metrics are invalidated the same way as when there is a granularity change or when data corruption is detected in netdata
6. keep both the old and the new database in the code, keeping the old database as default


mfundul:
How to test:
```
diff --git a/daemon/main.c b/daemon/main.c
index e93e97ca..01c28f2f 100644
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -688,13 +688,18 @@ void send_statistics( const char *action, const char *action_result, const char
     }
     freez(command_to_run);
 }
-
+extern uint8_t global_compress_alg;
+void rrdengine_main(void);
 int main(int argc, char **argv) {
     int i;
     int config_loaded = 0;
     int dont_fork = 0;
     size_t default_stacksize;
 
+if (argc > 1)
+    global_compress_alg = 1;
+rrdengine_main();
+return 0;
     // set the name for logging
     program_name = "netdata";
```

paulfantom:
For testing add installation of libuv dev library to:
- https://github.com/netdata/netdata/blob/master/.travis.yml#L19 (enables test stage to complete)
- https://github.com/netdata/netdata/blob/master/.travis.yml#L42 (enables 1/5 of build stage to run)
- https://github.com/netdata/netdata/blob/master/.travis.yml#L57 (enables packaging stage to run)
- https://github.com/netdata/netdata/blob/master/.travis.yml#L69 (enables nightly stage to run)
This will allow to run basic tests and the first travis stage. Build stage will fail in 4/5 as this will need to have docker images modified. I would suggest to add this dependency to a docker images when this PR gets close to being merged.

To add dependency to docker images you will need to modify files in a different repository. So this modification will affect all PRs in this repo. You will need to modify all files in this directory: https://github.com/netdata/helper-images/tree/master/oses


CC: @paulkatsoulakis

mfundul:
@paulkatsoulakis see [here](https://github.com/netdata/netdata/blob/f7de69830bf7da577d5e99d4ff39f038cbfce573/packaging/installer/README.md#install-netdata-on-linux-manually) for how to update the required packages of the CI images with the new dependencies of the DB engine.

Everybody else, the above is what you have to do to before installing this PR in your systems to test it.

To enable the new DB engine you need to change to the new `memory mode` in `/etc/netdata/netdata.conf` in this way:
```
[global]
        memory mode = dbengine
        # page cache size = 32
        # dbengine disk space = 256
```

`page cache size` and `dbengine disk space` are the maximum values in MiB for memory and disk space respectively.

packaging/installer/README.md
```diff
@@ -234,6 +234,15 @@ package|description
 
 *Netdata will greatly benefit if you have the above packages installed, but it will still work without them.*
 
+package|description
+:-----:|-----------
+`libuv`|multi-platform support library with a focus on asynchronous I/O
+`liblz4`|Extremely Fast Compression algorithm
```
paulkatsoulakis: The package has a different name in centos, probably needs to be clear here how to find it per distro

mfundul: These names are the official package names from the developers. The more distro specific names are in configure.ac in the error messages.

paulkatsoulakis:
Agreed with @cakrit we will be merging this regardless codacy failing.
Issues related to packaging concerns are merged in, testing has completed and team is confident we are ready to merge it.

To avoid problems in the middle of the night, i will merge this early tomorrow morning and then trigger nightlies to produce the new binaries for verification

paulkatsoulakis:
As agreed with the team, i force merged this PR. I will now trigger the nightly builds to get a release that includes the DB, so that we can verify all artifacts working fine with and without the database

idef1x:
Could it be that database change changed behavior to backends? I use graphite and average the netdata data, which it will send every 60 seconds. When my netdata automatically updated via cron at 6.25am, my graphs started to change as seen in the picture. In the meantime I played and looked around to finally reinstall netdata from commit 0d5fa83e224afa6d530ed6a5e6e6fe35a8af955e (fixing segmentation fault). The graphs looks back to normal (to me at least) again.
load

mfundul:
> 
> 
> Could it be that database change changed behavior to backends? I use graphite and average the netdata data, which it will send every 60 seconds. When my netdata automatically updated via cron on the 16th at 6.25am, my graphs started to change as seen in the picture. In the meantime I played and looked around to finally reinstall netdata from commit [0d5fa83](https://github.com/netdata/netdata/commit/0d5fa83e224afa6d530ed6a5e6e6fe35a8af955e) (fixing segmentation fault). The graphs looks back to normal (to me at least) again.
> load

Did you explicitly set `memory mode = dbengine` to change the database engine? The new database is not on by default.

idef1x:
No...it's on default setting (# memory mode = save) and it happened right after the netdata-updater cron job. So looking through the commits between the previous nightly update (the 15th) and the one on the 16th,  I thought it might something to do with the database change..(the others didn't seem relevant to me, but I am not a programmer)..

PR #5286, opened 2019-01-28
Add back the symlink netdata-latest.gz.run

cakrit:
static build 64 breaks without it. Hasn't built properly since Dec.


cakrit:
> Pipeline is working without it and artifacts are uploading to GCS

The static64 build has been failing for the last month. 

`https://github.com/netdata/binary-packages/blob/master/build-latest.sh` expects to find `"${SRC}/netdata-latest.gz.run")` : 

```
# build the static netdata
run ./packaging/makeself/build-x86_64-static.sh

# read the latest file generated
latest=$(run readlink "${SRC}/netdata-latest.gz.run")
```

PR #5326, opened 2019-02-04
Remove deprecated API call

Aisbergg:
##### Summary
The `mysql` plugin is not working anymore using the latest version 1.4.x of MySQLdb library. The reason is the call to a deprecated API that was recently removed (see: https://github.com/PyMySQL/PyMySQL/pull/763). 

##### Component Name
`mysql` python.d plugin

##### Additional Information
There using `with self.__connection as cursor:` does not work anymore. Commiting to the database is now solely handled by the `connection` object. When all queries were executed the cursor can simply be closed.


ilyam8:
@Aisbergg thank you!

PR #5327, opened 2019-02-04
Allow user to override the default behavior for read-only mounts

### LINKED ISSUE #5071: Netdata w/ Docker Container not show Disk space utilization for mounts
##### Bug report summary
Using netdata inside the official docker container (netdata/netdata) a partition mounted under `/mnt/<name>` is not recognized by netdata. It only found the disk (sdb in this case) in _disks_ section. **Disk space utilization** info for the given mount are not dispalyed.     
##### OS / Environment
Ubuntu Server 18.04.1 LTS
##### Steps To Reproduce
Command: 
`cat /host/proc/self/mountinfo | grep /mnt/<name>` inside the container 
`cat /proc/self/mountinfo | grep /mnt/<name>` on the host
##### Expected behavior
The outputs of the above commands are different: inside the container nothing appear, on the host the mount appear. The docker-compose file is correctly configured with `/proc:/host/proc:ro` directive and the container user is set with `grep docker /etc/group | cut -d ':' -f 3`.

### LINKED ISSUE #5231: netdata docker container not showing disk space utilization for lvm mounts
netdata in a docker container is executed with the command provided on the `README.md` and result shown below -
image
None of the dm-X disks are showing disk utilization.

adding `  -v /:/host:ro` actually shows disk utilizatio for all mounts except root. Removing `ro` fixes that aswell but I feel kind uncomfortable mounting it all as `rw`
image

Is this a known issue?
Running on ubuntu 14.04

vlvkobal:
##### Summary
When netdata runs in a container, it is desirable to have a possibility of displaying statistics for read-only mounts. By default charts for these mounts are disabled. This PR allows to enable them in `netdata.conf`
```
[plugin:proc:diskspace:/host]
	space usage = yes
	inodes usage = yes
```

Fixes #5071
Fixes #5231

##### Component Name
`diskspace` plugin

### DIFF
--- a/collectors/diskspace.plugin/README.md
+++ b/collectors/diskspace.plugin/README.md
@@ -2,6 +2,33 @@
 
 This plugin monitors the disk space usage of mounted disks, under Linux.
 
+Two charts are available for every mount:
+ - Disk Space Usage
+ - Disk Files (inodes) Usage
+
+## configuration
+
+Simple patterns can be used to exclude mounts from showed statistics based on path or filesystem. By default read-only mounts are not displayed. To display them `yes` should be set for a chart instead of `auto`.
+
+```
+[plugin:proc:diskspace]
+    # remove charts of unmounted disks = yes
+    # update every = 1
+    # check for new mount points every = 15
+    # exclude space metrics on paths = /proc/* /sys/* /var/run/user/* /run/user/* /snap/* /var/lib/docker/*
+    # exclude space metrics on filesystems = *gvfs *gluster* *s3fs *ipfs *davfs2 *httpfs *sshfs *gdfs *moosefs fusectl
+    # space usage for all disks = auto
+    # inodes usage for all disks = auto
+```
+
+Charts can be enabled/disabled for every mount separately:
+
+```
+[plugin:proc:diskspace:/]
+    # space usage = auto
+    # inodes usage = auto
+```
+
 > for disks performance monitoring, see the `proc` plugin, [here](../proc.plugin/#monitoring-disks)
 
 
--- a/collectors/diskspace.plugin/plugin_diskspace.c
+++ b/collectors/diskspace.plugin/plugin_diskspace.c
@@ -190,7 +190,7 @@ static inline void do_disk_space_stats(struct mountinfo *mi, int update_every) {
     if(unlikely(m->do_space == CONFIG_BOOLEAN_NO && m->do_inodes == CONFIG_BOOLEAN_NO))
         return;
 
-    if(unlikely(mi->flags & MOUNTINFO_READONLY && !m->collected))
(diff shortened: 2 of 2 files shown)


cakrit:
- Why don't you add a different config parameter? The logic is a bit strange like this. 
- You mentioned in the call adding some documentation. Can you add some content to the README file which is essentially empty? At least on how it works, the charts it produces and the config.

vlvkobal:
> * Why don't you add a different config parameter? The logic is a bit strange like this.

What config parameter? By default, we have `auto` instead of `yes`. The logic is quite straight: if `chart = yes` netdata should show it.

> * You mentioned in the call adding some documentation. Can you add some content to the README file which is essentially empty? At least on how it works, the charts it produces and the config.

I mentioned my [reply](https://github.com/netdata/netdata/issues/5071#issuecomment-460442980) in #5071. Added requested content to the README.

collectors/diskspace.plugin/README.md
```diff
@@ -2,6 +2,33 @@
 
 This plugin monitors the disk space usage of mounted disks, under Linux.
 
+Two charts are available for every mount:
+ - Disk Space Usage
+ - Disk Files (inodes) Usage
+
+## configuration
+
+Simple patterns can be used to exclude mounts from showed statistics based on path or filesystem. By default read-only mounts are not displayed. To display them `yes` should be set for a chart instead of `auto`.
```
cakrit: This doesn't where I need to set the yes. You probably want to mention it after the first quoted fragment.

collectors/diskspace.plugin/plugin_diskspace.c
```diff
@@ -190,7 +190,7 @@ static inline void do_disk_space_stats(struct mountinfo *mi, int update_every) {
     if(unlikely(m->do_space == CONFIG_BOOLEAN_NO && m->do_inodes == CONFIG_BOOLEAN_NO))
         return;
 
-    if(unlikely(mi->flags & MOUNTINFO_READONLY && !m->collected))
+    if(unlikely(mi->flags & MOUNTINFO_READONLY && !m->collected && m->do_space != CONFIG_BOOLEAN_YES && m->do_inodes != CONFIG_BOOLEAN_YES))
```
cakrit: The space and inodes settings can be set independently and are checked independently further down. So shouldn't this be `&& (m->collected && m->do_space != CONFIG_BOOLEAN_YES || m->do_inodes != CONFIG_BOOLEAN_YES)` ?

collectors/diskspace.plugin/README.md
vlvkobal: You can set it for `space usage for all disks` and/or `inodes usage for all disks` as well.

PR #5330, opened 2019-02-05
Remove codacy warnings

cakrit:
##### Summary
Declare id and chart once. 

Resolves warnings:
'id' used outside of binding context. (block-scoped-var) 
'id' is already defined. (no-redeclare)


cakrit:
> I don't like the change, declarations should be as specific as possible.
> You could use `let` to have proper scoping (and hopefully silence the warning).

Used `let` and `const`, it fixes it.

PR #5350, opened 2019-02-07
cups.plugin: Support older versions

simonnagl:
##### Summary

The usage of cupsGetIntegerOption did block users using older cups
versions than 2.2.4. After redistribution this method all cups versions
since 1.7 can be used. httpConnect2 is now the most recently added
method in cups.h.

##### Component Name

`cups.plugin`

##### Additional Information

Addition to #5324

### DIFF
--- a/collectors/cups.plugin/cups_plugin.c
+++ b/collectors/cups.plugin/cups_plugin.c
@@ -136,6 +136,37 @@ void parse_command_line(int argc, char **argv) {
     }
 }
 
+/*
+ * 'cupsGetIntegerOption()' - Get an integer option value.
+ *
+ * INT_MIN is returned when the option does not exist, is not an integer, or
+ * exceeds the range of values for the "int" type.
+ *
+ * @since CUPS 2.2.4/macOS 10.13@
+ */
+
+int					/* O - Option value or @code INT_MIN@ */
+getIntegerOption(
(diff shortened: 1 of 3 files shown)


collectors/cups.plugin/cups_plugin.c
```diff
@@ -136,6 +136,37 @@ void parse_command_line(int argc, char **argv) {
...
+ * 'cupsGetIntegerOption()' - Get an integer option value.
+ *
+ * INT_MIN is returned when the option does not exist, is not an integer, or
+ * exceeds the range of values for the "int" type.
+ *
+ * @since CUPS 2.2.4/macOS 10.13@
+ */
+
+int					/* O - Option value or @code INT_MIN@ */
+getIntegerOption(
```
simonnagl: @ktsaou @vlvkobal @cakrit This is copied from the cups repository. cups is licensed under Apache 2.0. Do I need to add some redistribution note somewhere?

ktsaou: I think not. It is just a small wrapper anyone could write. Not much of an algorithm to consider it intellectual property.

mfundul: From https://www.gnu.org/licenses/license-list.html#apache2 we can see Apache 2.0 is compatible with GPL 3 and can be merged.

PR #5351, opened 2019-02-08
fix docs: web_log time_multiplier

Wing924:
##### Summary

web_log's time_multiplier expect milliseconds

##### Component Name

python/web_log


ilyam8:
apache time in microseconds

>  %D | The time taken to serve the request, in microseconds

nginx in seconds

charts has 1000 divisor

the idea is convert whatever time we have to microseconds.

Wing924:
@ilyam8 Ah! It's my mistake.

Issue #5346, opened 2019-02-07
Alert alarm flapping


infnada:
##### Bug report summary

Alerts alarms on memory available are going really crazy (flapping) on 2 of the hosts

All alarms are configured by default

![image](https://user-images.githubusercontent.com/6633771/52409413-b97eb400-2ad5-11e9-8c0e-4b8a81efb300.png)

![image](https://user-images.githubusercontent.com/6633771/52409443-c8fdfd00-2ad5-11e9-9132-e61d1d1f468d.png)

And more, and more, and more alerts

This alarm conditions make no sense (i dont know how to read this...):
![image](https://user-images.githubusercontent.com/6633771/52410111-aa006a80-2ad7-11e9-8810-23e7bf518a59.png)

![image](https://user-images.githubusercontent.com/6633771/52410063-8a694200-2ad7-11e9-99b3-bed0979c9665.png)

##### OS / Environment
Docker Swarm Container (Docker version 18.09.1, build 4c52b90) over Centos (CentOS Linux release 7.6.1810 (Core))

##### Netdata version (ouput of `netdata -V`)
netdata v1.12.0-5-g8e818c28

##### Component Name
Web -> Alarms -> mem.available

##### Steps To Reproduce
Dont know, just have a lot of RAM used

##### Expected behavior

Alarm should not flap like this when there is 87,3% used ram. Or... as you can see in the images the alarm at some point is WARNING when mem.available = 7,21% and CLEAR when mem.available = 7,22% (it makes no sense).

Logs from one of netdata container:

`2019-02-07 11:42:20: 98: 31 '[10.255.0.108]:43568' 'DATA' (sent/all = 706/1470 bytes -52%, prep/sent/total = 0.10/0.21/0.31 ms) 200 '/api/v1/alarms?active&_=1549526067323'

2019-02-07 11:42:20: 98: 31 '[10.255.0.108]:43568' 'DATA' (sent/all = 583/1819 bytes -68%, prep/sent/total = 0.23/0.30/0.53 ms) 200 '/api/v1/alarm_log?after=1549531648&_=1549526067324'

2019-02-07 11:42:31: 98: 31 '[10.255.0.108]:43568' 'DATA' (sent/all = 963/2921 bytes -67%, prep/sent/total = 0.15/0.20/0.34 ms) 200 '/api/v1/alarms?active&_=1549526067325'

[...]

PR #5353, opened 2019-02-08
fix warning condition for mem.available

cakrit:
##### Summary
Fixes: #5346

##### Component Name
health

##### Additional Information
Warning limit for mem.available was wrong, causing flapping alarms.


ilyam8:
https://github.com/netdata/netdata/blob/9fbf340bb4bae5460fa38f2d34c25a42489f5720/health/health.d/ram.conf#L60

cakrit:
> same for FreeBSD (line 60)

Awesome, thanks! Fixed that one too.

PR #5361, opened 2019-02-10
Split nfacct plugin into separate process

### LINKED ISSUE #3749: Split NFACCT handling to a dedicated helper program.
This has been mentioned before in a couple of issues, but I didn't see one open specifically for this, so I'm opening this so that it's more publicly visible.

Netdata actually has support for tracking counters for Linux firewall rules, but it only works if you run Netdata as root (or possibly with the CAP_NET_ADMIN capability, I'm not sure about this though).

Ideally, this needs to be split out to a SUID helper program or just made into a binary module (like apps.plugin or the fping plugin) so that Netdata can be run as a normal user and still collect this data.

Relevant code is currently in [src/plugin_nfacct.c](https://github.com/firehol/netdata/blob/master/src/plugin_nfacct.c) and [src/plugin_nfacct.h](https://github.com/firehol/netdata/blob/master/src/plugin_nfacct.h), though the conntrack stuff in there too can just be kept as-is (it works as a non-root user).

vlvkobal:
##### Summary
As far as Netfilter statistics can be obtained with root access rights only, the `nfacct` plugin was separated from netdata daemon so that it isn't needed anymore to run the daemon as root for collecting Netfilter stats.

Fixes #3749

##### Component Name
`nfacct` plugin


cakrit:
Just one note, can you add nfacct to the monitored applications [here](https://github.com/netdata/netdata/blob/master/collectors/apps.plugin/apps_groups.conf#L83)?

cakrit:
Pressing red button so we can have it merged for a bit before @vlvkobal goes on leave.

PR #5365, opened 2019-02-11
sslcheck module: (remote) SSL certificate expiry time check

p-thurner:
##### Summary
Hi Netdata community!

So I think we need to graph the days until ssl certificates expire. I have forked the portcheck plugin to do that.

##### Component Name
collector/sslcheck

##### Additional Information
After lots of trying around, reading the doc and some similar issues I can just not get this plugin to graph anything.
I have done this on a fresh netdata setup with `curl | bash` setup of netdata on a Debian 9 server:
- created `/etc/netdata/python.d/sslcheck.conf`
- `git clone https://github.com/blunix/netdata-plugin-sslcheck /usr/src/netdata/collectors/python.d.plugin/sslcheck/` - thats the same code there as in this PR
- `systemctl restart netdata.service`

But it neither shows errors or any mention of the sslcheck plugin in the logs, nor does it show anything in the WebUI. I must be missing something.. Do you guys have any idea?

Thanks a lot in advance!

### DIFF
--- a/collectors/python.d.plugin/sslcheck/README.md
+++ b/collectors/python.d.plugin/sslcheck/README.md
@@ -13,9 +13,10 @@ update_every : 60
 
 example_org:
   host: 'example.org'
-  days_until_expiration_warning: 5
+  days_until_expiration_warning: 15
 
 my_site_org:
   host: 'my-site.org'
-  days_until_expiration_warning: 5
+  days_until_expiration_warning: 10
+  days_until_expiration_critical: 1
 ```
--- a/collectors/python.d.plugin/sslcheck/README.md
+++ b/collectors/python.d.plugin/sslcheck/README.md
@@ -13,7 +13,6 @@ update_every : 60
 
 example_org:
   host: 'example.org'
-  days_until_expiration_warning: 15
 
 my_site_org:
   host: 'my-site.org'
--- a/collectors/python.d.plugin/sslcheck/README.md
+++ b/collectors/python.d.plugin/sslcheck/README.md
@@ -1,9 +1,26 @@
-### configuration
+# SSL certificate expiry check
+
+Checks the time until a remote SSL certificate expires in days.
+
+## Requirements
+
+None
+
+## Example configuration
 
 ```yaml
-master:
-  host: www.my-netdata.io
-  port: 443
-  timeout: 10
+job_name:
+  name: myname                 # [optional] the JOB's name as it will appear at the
+                               # dashboard (by default is the job_name)
+  host: 'my-netdata.io'        # [required] the remote host address in either IPv4, IPv6 or as DNS name
+  port: 443                    # [required] the port number to check. Specify an integer, not service name
+  daysuntilexpiration: 5       # [required] days before the status is critical?
+  timeout: 10                  # [optional] the socket timeout when connecting
+  update_every: 120            # [optional] the JOB's data collection frequency. As the chart is in days
+                               # remaining, a high value should be chosen
+  priority: 60000              # [optional] the JOB's order on the dashboard
+  penalty: yes                 # [optional] the JOB's penalty
 ```
+
+
 
--- a/collectors/python.d.plugin/sslcheck/README.md
+++ b/collectors/python.d.plugin/sslcheck/README.md
@@ -1,0 +1,9 @@
+### configuration
+
+```yaml
+master:
+  host: www.my-netdata.io
+  port: 443
+  timeout: 10
+```
+
--- a/collectors/python.d.plugin/sslcheck/README.md
+++ b/collectors/python.d.plugin/sslcheck/README.md
@@ -1,26 +1,21 @@
 # SSL certificate expiry check
 
-Checks the time until a remote SSL certificate expires in days.
+Checks the time until a remote SSL certificate expires.
 
 ## Requirements
 
 None
 
-## Example configuration
+### configuration
 
 ```yaml
-job_name:
-  name: myname                 # [optional] the JOB's name as it will appear at the
-                               # dashboard (by default is the job_name)
-  host: 'my-netdata.io'        # [required] the remote host address in either IPv4, IPv6 or as DNS name
-  port: 443                    # [required] the port number to check. Specify an integer, not service name
-  daysuntilexpiration: 5       # [required] days before the status is critical?
-  timeout: 10                  # [optional] the socket timeout when connecting
-  update_every: 120            # [optional] the JOB's data collection frequency. As the chart is in days
-                               # remaining, a high value should be chosen
-  priority: 60000              # [optional] the JOB's order on the dashboard
-  penalty: yes                 # [optional] the JOB's penalty
+update_every : 60
+
+example_org:
+  host: 'example.org'
+  days_until_expiration_warning: 5
+
+my_site_org:
+  host: 'my-site.org'
+  days_until_expiration_warning: 5
 ```
-
-
-
--- a/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
+++ b/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
@@ -56,7 +56,7 @@ class Service(SimpleService):
         self.debug('run check : host {host}:{port}, update every {update}s, timeout {timeout}s'.format(
             host=self.host, port=self.port, update=self.update_every, timeout=self.timeout))
 
-        return True
+        return bool(self.get_data())
 
     def get_data(self):
         conn = create_ssl_conn(self.host, self.timeout)
--- a/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
+++ b/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
@@ -26,7 +26,11 @@ CHARTS = {
             ],
             'lines': [
                 ['time'],
-            ]
+            ],
+            'variables': [
+                ['days_until_expiration_warning'],
+            ],
+
         },
 }
 
@@ -42,6 +46,7 @@ class Service(SimpleService):
         self.host = configuration.get('host')
         self.port = configuration.get('port', 443)
         self.timeout = configuration.get('timeout', 3)
+        self.days_warn = configuration.get('days_until_expiration_warning', 5)
 
     def check(self):
         if not self.host:
@@ -73,7 +78,8 @@ class Service(SimpleService):
             return None
 
         return {
-            'time': calc_cert_expiration_time(peer_cert).seconds
+            'time': cert_expiration_seconds(peer_cert),
+            'days_until_expiration_warning': self.days_warn,
         }
 
 
@@ -88,8 +94,9 @@ def create_ssl_conn(hostname, timeout):
     return conn
 
 
-def calc_cert_expiration_time(cert):
+def cert_expiration_seconds(cert):
     expiration_date = datetime.datetime.strptime(cert['notAfter'], SSL_DATE_FMT)
     current_date = datetime.datetime.utcnow()
+    delta = expiration_date - current_date
 
-    return expiration_date - current_date
+    return ((delta.days * 86400 + delta.seconds) * 10 ** 6 + delta.microseconds) / 10 ** 6
--- a/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
+++ b/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # Description: simple ssl expiration check netdata python.d module
-# Original Author: ccremer (github.com/ccremer)
+# Original Author: Peter Thurner (github.com/p-thurner)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import datetime
--- a/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
+++ b/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
@@ -10,6 +10,9 @@ import ssl
 from bases.FrameworkServices.SimpleService import SimpleService
 
 
+update_every = 60
+
+
 ORDER = [
     'time_until_expiration',
 ]
--- a/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
+++ b/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
@@ -21,7 +21,7 @@ DAYS_UNTIL_EXPIRATION = 'whatisthis'
 ORDER = ['daysuntilexpiration']
 
 CHARTS = {
-    'daysvalid': {
+    'daysuntilexpiration': {
         'options': [None, 'Days until expiration', 'days', 'daysuntilexpiration', 'sslcheck.daysuntilexpiration', 'line'],
         'lines': [
             [DAYS_UNTIL_EXPIRATION, 'days until expiration', 'absolute', 100, 1000]
@@ -65,7 +65,7 @@ class Service(SimpleService):
         :return: dict
         """
         data = dict()
-        data[DAYS_UNTIL_EXPIRATION] = self.ssl_valid_time_remaining(self.host)
+        data[DAYS_UNTIL_EXPIRATION] = self.ssl_valid_time_remaining(self.host, self.port)
 
         return data
 
--- a/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
+++ b/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+# Description: simple ssl expiration check netdata python.d module
+# Original Author: ccremer (github.com/ccremer)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import ssl
+import socket
+
+import datetime
+
+try:
+    from time import monotonic as time
+except ImportError:
+    from time import time
+
+from bases.FrameworkServices.SimpleService import SimpleService
+
+
+DAYS_UNTIL_EXPIRATION = 'whatisthis'
+
+ORDER = ['daysuntilexpiration']
+
+CHARTS = {
+    'daysvalid': {
+        'options': [None, 'Days until expiration', 'days', 'daysuntilexpiration', 'sslcheck.daysuntilexpiration', 'line'],
+        'lines': [
+            [DAYS_UNTIL_EXPIRATION, 'days until expiration', 'absolute', 100, 1000]
+        ]
+    }
+}
+
+
+# Not deriving from SocketService, too much is different
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.host = self.configuration.get('host')
+        self.port = self.configuration.get('port')
+        self.timeout = self.configuration.get('timeout', 3)
+
+    def check(self):
+        """
+        Parse configuration, check if configuration is available, and dynamically create chart lines data
+        :return: boolean
+        """
+        if self.host is None or self.port is None:
+            self.error('Host or port missing')
+            return False
+        if not isinstance(self.port, int):
+            self.error('"port" is not an integer. Specify a numerical value, not service name.')
+            return False
+
+        self.debug('Enabled sslcheck: {host}:{port}, update every {update}s, timeout: {timeout}s'.format(
+            host=self.host, port=self.port, update=self.update_every, timeout=self.timeout
+        ))
+        # We will accept any (valid-ish) configuration, even if initial connection fails (a service
+        # might be down from the beginning)
+        return True
+
+    def _get_data(self):
+        """
+        Get number of days until the certificate for the configured hostname expires
+        :return: dict
+        """
+        data = dict()
+        data[DAYS_UNTIL_EXPIRATION] = self.ssl_valid_time_remaining(self.host)
+
+        return data
+
+    def ssl_expiry_datetime(self, hostname, port):
+        ssl_date_fmt = r'%b %d %H:%M:%S %Y %Z'
+
+        context = ssl.create_default_context()
+        conn = context.wrap_socket(
+            socket.socket(socket.AF_INET),
+            server_hostname=hostname,
+        )
+
+        conn.settimeout(self.timeout)
+
+        conn.connect((hostname, port))
+        ssl_info = conn.getpeercert()
+        # parse the string from the certificate into a Python datetime object
+        return datetime.datetime.strptime(ssl_info['notAfter'], ssl_date_fmt)
+
+    def ssl_valid_time_remaining(self, hostname, port):
+        """Get the number of days left in a cert's lifetime."""
+        expires = self.ssl_expiry_datetime(hostname, port)
+        return expires - datetime.datetime.utcnow()
+
+    # def ssl_expires_in(self, hostname, buffer_days=14):
+    #    """
+    #    Check if `hostname` SSL cert expires within `buffer_days`.
+    #    :return: Bool
+    #    """
+    #    remaining = self.ssl_valid_time_remaining(hostname)
+
+    #    # if the cert expires in less than two weeks or is already expired, return True
+    #    if remaining < datetime.timedelta(days=0):
+    #        # cert has already expired
+    #        return True
+    #    elif remaining < datetime.timedelta(days=buffer_days):
+    #        # expires sooner than the buffer
+    #        return True
+    #    else:
+    #        # everything is fine
+    #        return False
+
--- a/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
+++ b/collectors/python.d.plugin/sslcheck/sslcheck.chart.py
@@ -10,76 +10,86 @@ import ssl
 from bases.FrameworkServices.SimpleService import SimpleService
 
 
-DAYS_UNTIL_EXPIRATION = 'daysuntilexpiration'
-
-ORDER = ['daysuntilexpiration']
+ORDER = [
+    'time_until_expiration',
+]
 
 CHARTS = {
-    'daysuntilexpiration': {
-        'options': [None, 'days until certificate expiration', 'days', 'days until expiration', 'sslcheck.daysuntilexpiration', 'line'],
-        'lines': [
-            [DAYS_UNTIL_EXPIRATION, 'days until expiration', 'absolute', 10, 0]
-        ]
-    }
+        'time_until_expiration': {
+            'options': [
+                None,
+                'Time Until Certificate Expiration',
+                'seconds',
+                'certificate expiration',
+                'sslcheck.time_until_expiration',
+                'line',
+            ],
+            'lines': [
+                ['time'],
+            ]
+        },
 }
 
 
-# Not deriving from SocketService, too much is different
+SSL_DATE_FMT = r'%b %d %H:%M:%S %Y %Z'
+
+
 class Service(SimpleService):
     def __init__(self, configuration=None, name=None):
         SimpleService.__init__(self, configuration=configuration, name=name)
         self.order = ORDER
         self.definitions = CHARTS
-        self.host = self.configuration.get('host')
-        self.port = self.configuration.get('port')
-        self.timeout = self.configuration.get('timeout', 3)
+        self.host = configuration.get('host')
+        self.port = configuration.get('port', 443)
+        self.timeout = configuration.get('timeout', 3)
 
     def check(self):
-        """
-        Parse configuration, check if configuration is available, and dynamically create chart lines data
-        :return: boolean
-        """
-        if self.host is None or self.port is None:
-            self.error('Host or port missing')
-            return False
-        if not isinstance(self.port, int):
-            self.error('"port" is not an integer. Specify a numerical value, not service name.')
+        if not self.host:
+            self.error('host parameter is mandatory, but it is not set')
             return False
 
-        self.debug('Enabled sslcheck: {host}:{port}, update every {update}s, timeout: {timeout}s'.format(
-            host=self.host, port=self.port, update=self.update_every, timeout=self.timeout
-        ))
-        # We will accept any (valid-ish) configuration, even if initial connection fails (a service
-        # might be down from the beginning)
+        self.debug('run check : host {host}:{port}, update every {update}s, timeout {timeout}s'.format(
+            host=self.host, port=self.port, update=self.update_every, timeout=self.timeout))
+
         return True
 
-    def _get_data(self):
-        """
-        Get number of days until the certificate for the configured hostname expires
-        :return: dict
-        """
-        data = dict()
-        data[DAYS_UNTIL_EXPIRATION] = self.ssl_valid_time_remaining(self.host, self.port)
+    def get_data(self):
+        conn = create_ssl_conn(self.host, self.timeout)
 
-        return data
+        try:
+            conn.connect((self.host, self.port))
+        except Exception as error:
+            self.error("error on connection to {0}:{1} : {2}".format(self.host, self.port, error))
+            return None
 
-    def ssl_expiry_datetime(self, hostname, port):
-        ssl_date_fmt = r'%b %d %H:%M:%S %Y %Z'
+        peer_cert = conn.getpeercert()
+        conn.close()
 
-        context = ssl.create_default_context()
-        conn = context.wrap_socket(
-            socket.socket(socket.AF_INET),
-            server_hostname=hostname,
-        )
+        if peer_cert is None:
+            self.warning("no certificate was provided by {0}:{1}".format(self.host, self.port))
+            return None
+        elif not peer_cert:
+            self.warning("certificate was provided by {0}:{1}, but not validated".format(self.host, self.port))
+            return None
 
-        conn.settimeout(self.timeout)
+        return {
+            'time': calc_cert_expiration_time(peer_cert).seconds
+        }
 
-        conn.connect((hostname, port))
-        ssl_info = conn.getpeercert()
-        # parse the string from the certificate into a Python datetime object
-        return datetime.datetime.strptime(ssl_info['notAfter'], ssl_date_fmt)
 
-    def ssl_valid_time_remaining(self, hostname, port):
-        """Get the number of days left in a cert's lifetime."""
-        expires = self.ssl_expiry_datetime(hostname, port)
-        return (expires - datetime.datetime.utcnow()).days
+def create_ssl_conn(hostname, timeout):
+    context = ssl.create_default_context()
+    conn = context.wrap_socket(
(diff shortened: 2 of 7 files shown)


ilyam8:
@p-thurner i think we need something like this

https://github.com/influxdata/telegraf/tree/master/plugins/inputs/x509_cert

What do you think?

And since we have [go.d.plugin](https://github.com/netdata/go.d.plugin) we willing to add new modules  to it, if there is no limitations. So, the question, are you willing to rewrite it in go?

ilyam8:
@p-thurner gj btw, we definitely need this one :+1:

p-thurner:
@ilyam8 hm I'd rather keep it in python as I didn't learn golang yet.. Is it ok to keep this one in python?

I'd be super happy to get this working (invested some hours into just getting the graph into the webui with no luck) :(
If this works I also want to add:
- cronjobs (like mk-job from check-mk)
- apt security upgrades
- ssl setup rating (by asking the ssllabs API)

Then I have pretty much what my previous monitoring solution had.

So I would be very thankful for any help to get this working.. Will commit more plugins then!

ilyam8:
It is ok to have them in python, but i'd prefer to have new modules in go. It is very easy/fast to write it in go, because it is already written (telegraf) :smile: I can do it tomorrow.

Please describe how the other 3 modules works (how they fetches data - http req/file read). WIll see, maybe some of them would be ok to have in python.

ilyam8:
> I'd be super happy to get this working

in general - you `debug` mode, it helps a lot.

1. `sudo su -s /bin/bash netdata`
2. `./python.d.plugin debug trace <module_name>`

p-thurner:
@ilyam8, check-mk is another monitoring system. It has a shell script called `mk-job`, which is used to execute cronjobs like so:
```
5 0 * * * root mk-job nightly-backup /usr/local/bin/backup >/dev/null
```
This will create a simple file in `/var/lib/check_mk/agent/nightly-backup` with the following exemplary content:
```
user@host:~$ cat /var/lib/check_mk/agent/nightly-backup
start_time 1549921929
exit_code 0
real_time 0:01.00
user_time 0.00
system_time 0.00
reads 0
writes 0
max_res_kbytes 2176
avg_mem_kbytes 0
invol_context_switches 1
vol_context_switches 2
```
Here is the shell script: https://github.com/opinkerfi/check_mk/blob/master/agents/mk-job
It allows to monitor the exit status of the cronjob, how long the job was running and if it was executed at a specified time.. quite handy :)

Apt security upgrades is simple, just count the number of security upgrades.. With the `unattended-upgrades` configured on Debian and Ubuntu systems, it makes sense to count those packages and alert if the number is still above 0 after 24 hours (the package "randomly" installs security upgrades throughout the day so not all machines in the world pull the upgrades at the same time).

Now you may know this: https://www.ssllabs.com/ssltest/analyze.html?d=my%2dnetdata.io&s=104.28.3.248&latest (one of many I suppose)
It returns something like A+, A, B, C and so on depending on "how nicely" you setup your SSL on your webserver. I suppose A+ could be 0, A could be 1, B could be 2 and so on. This way one could graph it.
ssllabs of course gets angry when you hit their API once a minute.. They do however cache the result for a while.. I'm not sure how to get that into netdata but it would be pretty awesome to have it as unmaintained webservers get a worse rating after some time (weeks / month).

p-thurner:
btw if anyone can still help me debug my sslcheck plugin it would still be awesome :) I would still want to get it working to learn.

ilyam8:
@p-thurner 

> btw if anyone can still help me debug my sslcheck plugin it would still be awesome :) I would still want to get it working to learn.

`debug mode/error.log` (in thar order) are your friends

cakrit:
One note from my side, `update every` should default to something very large, perhaps once per hour. No point checking every second for something that makes sense on a daily timeframe.

Ferroin:
On the note of the SSL configuration thing, it might be worth looking at https://observatory.mozilla.org, they provide a clear quantifiable score (currently on the range of 0 to 135), and check some useful security aspects beyond just SSL configuration, including various security related HTTP headers.  They cache results for 5 minutes though, so that's probably the absolute minimum polling period (I think SSLLabs caches less aggressively).

For the APT security updates, on systems that have a recent version of APT, you can list all upgradeable packages with `apt list --upgradeable` (though it may require root).  The output format is one line per package, with each line looking something like this: `apt/unstable 1.8.0~rc3 amd64 [upgradable from: 1.8.0~beta1]`.  I'm not sure how to parse security updates out of that, but being able to just list all available updates is still useful.

ilyam8:
My thougths:

#### SSL certificate expiry time check 
 
Ok, lets do it in python. 

#### Monitoring cronjobs

According your descriptions is a super specific thing. It is not monitoring cron job actually, it is parsing files from produced by some scripts. Job couldbe executed once a day, once a week etc. There is nothing to chart.  

I think we don't need this module in core. We can make a link on the third-party modules page tho.

#### Monitoring how well you setup SSL on your webserver

> One more "problem" with that is that ssllabs itself takes a while to get the present the final result, which means the check plugin would run for several minutes the first time, then would take half a second (cached result) until the cache of ssllabs expires again.. So that one is a bit more complicated.
I suppose a simple shellscript / cronjob could query ssllabs and write the results to a textfile, where they could be picked up by a netdata plugin.

Same. 

tl;dr
 * SL certificate expiry time check  in python, module kind of ok, can be merged in core
 * others are too specific, i suggest to add them on third-party page and that is it.


@paulfantom @Ferroin @cakrit

Ferroin:
I would tend to agree with @ilyam8 on the SSL server configuration (I think it would be neat to have, you could get an alert when your server configuration falls out of BCP) and the cron jobs (also, would be kind of neat to have, but I don't consider it critical (if you have a proper email server set up, any sensible cron system will email you if there are errors)).

I do, however, think having a module that reports how many updates are pending on the system (not necessarily just security updates, and not necessarily just APT either) would be awesome, but I think that's probably something that should only be exposed as a counter, not a graph, and I also think the check frequency should be significantly longer for it by default than Netdata normally tracks (at least 30 minutes, possibly even longer).  It's not something that is likely to change with high frequency, so it just doesn't make sense to track historical data in most cases except for triggering alarms on changes or non-zero values.  If we do decide to do such a module, I can provide some help to get it working for Gentoo, as well as being able to help with testing on Debian and a couple of other distros.

p-thurner:
## Cronjobs
> if you have a proper email server set up, any sensible cron system will email you if there are errors
> Job couldbe executed once a day, once a week etc. There is nothing to chart.

I would like to politely disagree here. I run a bunch of servers with PHP websites. My customers use a lot of PHP written scripts / cron jobs. Sometimes these jobs behave weirdly, use way more ressources and take much longer than they usually do. Having thing monitoring for this is a very vital thing in my opinion.
Graphing the status file that `mk-job` outputs would not be to hard in my opinion - one could make a plugin that accepts a config file with multiple sections like so:
```
my_favorite_job:
  status_file: /var/foo/bar/my_favorite_job.txt
  every: 12h
```

All of this is of course not monitoring the _actual_ cronjob.. But it is close enough in my opinion to detect:
- cronjobs runs to long
- exits != 0
- needs to much ressources

## SSL "score"

The mozilla thing sounds very cool, especially the score.. Caching only 5 mins of course is a bit less.. SSLLabs caches the result a bit longer I think. I also just found this: https://github.com/ssllabs/ssllabs-scan which we maybe could use for more simplicity.

## Apt

I agree with you guys, regular and security upgrades can be a counter (I would not mind a graph either).
I previously used bloonix for monitoring, which does it like this (danger, perl :P)
https://github.com/bloonix/bloonix-plugins-linux/blob/master/plugins/check-linux-updates#L234-L244
Pretty much the following in a nutshell. Please note that I ran this as regular user (on my ubuntu 18.04 Laptop), not as root.
```
# Regular
user@host:~$ apt-get --force-yes --simulate --quiet --yes --allow-unauthenticated upgrade 2>/dev/null | grep "^Inst" | wc -l
119

# Security
user@host:~$ apt-get --force-yes --simulate --quiet --yes --allow-unauthenticated upgrade 2>/dev/null | egrep "^Inst|Security" | wc -l
```
Sadly the second one doesn't output anything for me as I don't have any server where not all security upgrades are installed x) But according to this thread it should work: https://askubuntu.com/questions/774805/how-to-get-a-list-of-all-pending-security-updates

I can only join the conversation with Debian / Ubuntu, I'm not very familiar with other distros.. I assume having several plugins for zypper / apt / rpm / whatnot wouldn't hurt and increase readability of the plugins.

p-thurner:
> One note from my side, `update every` should default to something very large, perhaps once per hour. No point checking every second for something that makes sense on a daily timeframe.

I will get back to working on the sslcheck this evening.

ilyam8:
>: Graphing the status file that `mk-job` outputs would not be to hard in my opinion (with the note that the graph wouldn't change very often of course..)

Having charts that updating once a day is confusing imo and kind of alien for netdata. I agree that we need to support collecting data without  actually charting it, but not we can't

And no, it is not that easy as you think. Your cronjob monigtoring script should read that file **right after** cron job is done and not every hour/6hour/etc.

Ferroin:
On the note of cronjobs, what you're talking about checking other than exit status can actually be done a couple of different ways without needing a new plugin for Netdata.  One option would be to use the existing `apps.plugin` together with some creative abuse of the process name (IOW, use _something_ to change the process name of the script to a unique, recognizable name, and then configure `apps.plugin` to look for it and treat it as it's own group).  The other is to put each cron job in it's own cgroup, and then just use Netdata's cgroup support to monitor them (this has the nice bonus that it can be configured to automatically kill the cron job if it uses too many resources).

For the upgrade related stuff, it looks like the command I had posted generically works correctly for regular users too, though it still doesn't appear to provide an easy way to identify security updates.  I'd kind of like to avoid depending on classic `apt-get` given that current stable versions of Ubuntu, Mint, and Debian have both already moved to the new `apt` command as the default, but that's probably just me.  On the Gentoo side,  I could probably have something working for just upgrades pretty quickly, but it would be dependent on an extra piece of software that isn't installed by default on Gentoo (though most sane people who use Gentoo have said software installed already).

p-thurner:
> both already moved to the new `apt` command as the default

From my Ubuntu 18.04 Laptop:
```
user@host:~$ apt update | grep fo
[...]
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
[...]
```

I think `apt-get` is still a valid choice for the future.

Ferroin:
> From my Ubuntu 18.04 Laptop:
> 
> ```
> user@host:~$ apt update | grep fo
> [...]
> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
> [...]
> ```
> I think `apt-get` is still a valid choice for the future.

Yeah, I'd forgotten that they still complain if you use it in a pipe.  I seriously doubt that the interface of the particular sub-command we would be using will change, but it probably is better to follow the official advice and use `apt-get` for scripting.

p-thurner:
Howdy fellas,

sorry didn't find time the last few days. So I did this on a fresh Debian 9 installation with no firewall configured:
```
# Setup the hostname in /etc/hosts, /etc/hostname and via `hostname` command to "monitoring.example.com"
# Update all packages to latest
apt update; apt -y upgrade; apt -y dist-upgrade; apt -y install unattended-upgrades curl git; dpkg-reconfigure unattended-upgrades
# Install netdata
bash <(curl -Ss https://my-netdata.io/kickstart.sh)

# Git clone the sslcheck plugin (same code as in this PR) and move files in place (as root user)
git clone https://github.com/blunix/netdata-plugin-sslcheck.git
cd netdata-plugin-sslcheck/
cp sslcheck.chart.py /usr/libexec/netdata/python.d/
editor /etc/netdata/python.d/sslcheck.conf (contents below)
my_sslcheck:
  host: my-netdata.io
  port: 443
  timeout: 10
  daysuntilexpiration: 5
daysuntilexpiration: 5


# Restart netdata
systemctl restart netdata.service
```

I then tried the debug command:
```
sudo su -s /bin/bash netdata
cd ~
netdata@monitoring:~$ /usr/libexec/netdata/plugins.d/python.d.plugin debug trace sslcheck
2019-02-18 14:04:58: python.d INFO: plugin: main: Using python 2
2019-02-18 14:04:58: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d.conf'
2019-02-18 14:04:58: python.d ERROR: plugin: main: cannot load '/etc/netdata/python.d.conf': [Errno 2] No such file or directory: '/etc/netdata/python.d.conf'. Will try stock version.
2019-02-18 14:04:58: python.d DEBUG: plugin: main: loading '/usr/lib/netdata/conf.d/python.d.conf'
2019-02-18 14:04:58: python.d DEBUG: plugin: main: module load source: 'sslcheck' => [OK]
2019-02-18 14:04:58: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d/sslcheck.conf'
2019-02-18 14:04:58: python.d DEBUG: plugin: main: job initialization: 'sslcheck my_sslcheck' => ['OK']
2019-02-18 14:04:58: python.d DEBUG: plugin: main: module status: 'sslcheck' => [OK] (jobs: 1)
2019-02-18 14:04:58: python.d DEBUG: sslcheck: my_sslcheck: Enabled sslcheck: my-netdata.io:443, update every 1s, timeout: 10s
2019-02-18 14:04:58: python.d INFO: sslcheck: my_sslcheck: check() => [OK]
2019-02-18 14:04:58: python.d DEBUG: sslcheck: my_sslcheck: create() => [NOT ADDED] chart 'daysuntilexpiration' not in definitions. Skipping it.
2019-02-18 14:04:58: python.d ERROR: sslcheck: my_sslcheck: create() => [FAILED] (charts: 0)
2019-02-18 14:04:58: python.d INFO: plugin: main: FINISHED
```

Looks good so far - not sure why the debug command wasn't outputting anything for me the last time.

Now that `[NOT ADDED] chart 'daysuntilexpiration' not in definitions` I don't fully get.. Any ideas?

ilyam8:
@p-thurner (not in definitions == not in CHARTS)

```
ORDER = ['daysuntilexpiration']

 CHARTS = {
    'daysvalid': {
```

p-thurner:
ah yes, thank you, also just got that very simple one ;)

```
CHARTS = {
    'daysuntilexpiration': {
```

Now I get:
```
netdata@monitoring:~$ /usr/libexec/netdata/plugins.d/python.d.plugin debug trace sslcheck
2019-02-18 14:16:35: python.d INFO: plugin: main: Using python 2
2019-02-18 14:16:35: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d.conf'
2019-02-18 14:16:35: python.d ERROR: plugin: main: cannot load '/etc/netdata/python.d.conf': [Errno 2] No such file or directory: '/etc/netdata/python.d.conf'. Will try stock version.
2019-02-18 14:16:35: python.d DEBUG: plugin: main: loading '/usr/lib/netdata/conf.d/python.d.conf'
2019-02-18 14:16:35: python.d DEBUG: plugin: main: module load source: 'sslcheck' => [OK]
2019-02-18 14:16:35: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d/sslcheck.conf'
2019-02-18 14:16:35: python.d DEBUG: plugin: main: job initialization: 'sslcheck my_sslcheck' => ['OK']
2019-02-18 14:16:35: python.d DEBUG: plugin: main: module status: 'sslcheck' => [OK] (jobs: 1)
2019-02-18 14:16:35: python.d DEBUG: sslcheck: my_sslcheck: Enabled sslcheck: my-netdata.io:443, update every 1s, timeout: 10s
2019-02-18 14:16:35: python.d INFO: sslcheck: my_sslcheck: check() => [OK]
CHART netdata.runtime_sslcheck_my_sslcheck '' 'Execution time for sslcheck_my_sslcheck' 'ms' 'python.d' netdata.pythond_runtime line 145000 1

2019-02-18 14:16:35: python.d DEBUG: sslcheck: my_sslcheck: create() => [OK] (charts: 1)
2019-02-18 14:16:35: python.d DEBUG: sslcheck: my_sslcheck: started, update frequency: 1
2019-02-18 14:16:36: python.d DEBUG: sslcheck: my_sslcheck: none of the charts has been updated
2019-02-18 14:16:36: python.d DEBUG: sslcheck: my_sslcheck: update => [FAILED] (elapsed time: -, failed retries in a row: 1)
2019-02-18 14:16:37: python.d DEBUG: sslcheck: my_sslcheck: none of the charts has been updated
2019-02-18 14:16:37: python.d DEBUG: sslcheck: my_sslcheck: update => [FAILED] (elapsed time: -, failed retries in a row: 2)
2019-02-18 14:16:38: python.d DEBUG: sslcheck: my_sslcheck: none of the charts has been updated
2019-02-18 14:16:38: python.d DEBUG: sslcheck: my_sslcheck: update => [FAILED] (elapsed time: -, failed retries in a row: 3)
```

ilyam8:
> 2019-02-18 14:16:36: python.d DEBUG: sslcheck: my_sslcheck: none of the charts has been updated

That usually means that returned by `get_data` dict contains no dimension ids (keys).

p-thurner:
Ah ok, I see, thank you @ilyam8. I'm now getting this output, which looks better:

```
netdata@monitoring:~$ /usr/libexec/netdata/plugins.d/python.d.plugin debug trace sslcheck
2019-02-18 14:44:59: python.d INFO: plugin: main: Using python 2
2019-02-18 14:44:59: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d.conf'
2019-02-18 14:44:59: python.d ERROR: plugin: main: cannot load '/etc/netdata/python.d.conf': [Errno 2] No such file or directory: '/etc/netdata/python.d.conf'. Will try stock version.
2019-02-18 14:44:59: python.d DEBUG: plugin: main: loading '/usr/lib/netdata/conf.d/python.d.conf'
2019-02-18 14:44:59: python.d DEBUG: plugin: main: module load source: 'sslcheck' => [OK]
2019-02-18 14:44:59: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d/sslcheck.conf'
2019-02-18 14:44:59: python.d DEBUG: plugin: main: job initialization: 'sslcheck my_sslcheck' => ['OK']
2019-02-18 14:44:59: python.d DEBUG: plugin: main: module status: 'sslcheck' => [OK] (jobs: 1)
2019-02-18 14:44:59: python.d DEBUG: sslcheck: my_sslcheck: Enabled sslcheck: my-netdata.io:443, update every 1s, timeout: 10s
2019-02-18 14:44:59: python.d INFO: sslcheck: my_sslcheck: check() => [OK]
CHART netdata.runtime_sslcheck_my_sslcheck '' 'Execution time for sslcheck_my_sslcheck' 'ms' 'python.d' netdata.pythond_runtime line 145000 1

2019-02-18 14:44:59: python.d DEBUG: sslcheck: my_sslcheck: create() => [OK] (charts: 1)
2019-02-18 14:44:59: python.d DEBUG: sslcheck: my_sslcheck: started, update frequency: 1
CHART sslcheck_my_sslcheck.daysuntilexpiration '' 'Days until expiration' 'days' 'daysuntilexpiration' 'sslcheck.daysuntilexpiration' line 60000 1 '' 'python.d.plugin' 'sslcheck'
DIMENSION 'daysuntilexpiration' 'days until expiration' absolute 100 1000 ''

BEGIN sslcheck_my_sslcheck.daysuntilexpiration 0
SET 'daysuntilexpiration' = 212
END

BEGIN netdata.runtime_sslcheck_my_sslcheck 0
SET run_time = 33
END

2019-02-18 14:44:59: python.d DEBUG: sslcheck: my_sslcheck: update => [OK] (elapsed time: 33, failed retries in a row: 0)
BEGIN sslcheck_my_sslcheck.daysuntilexpiration 1003913
SET 'daysuntilexpiration' = 212
END

BEGIN netdata.runtime_sslcheck_my_sslcheck 1003913
SET run_time = 13
END

2019-02-18 14:45:00: python.d DEBUG: sslcheck: my_sslcheck: update => [OK] (elapsed time: 13, failed retries in a row: 0)
BEGIN sslcheck_my_sslcheck.daysuntilexpiration 996083
SET 'daysuntilexpiration' = 212
END

BEGIN netdata.runtime_sslcheck_my_sslcheck 996083
SET run_time = 16
END

2019-02-18 14:45:01: python.d DEBUG: sslcheck: my_sslcheck: update => [OK] (elapsed time: 16, failed retries in a row: 0)
```

On my dev machine I added a print statement for the returned dict (named `data` in the code), looks like this:
```
{'daysuntilexpiration': 212}
```

Still no graph in the webui though..

p-thurner:
Here is the dict the portcheck plugin returns:
```
{'connect': 45.0, 'timeout': 0, 'success': 1, 'no_connection': 0}
```
I assume I also have to return something like `timeout` and `success`? Are there any pre-defined expected values (as in is there doc for this somewhere)?

ilyam8:
@p-thurner 
see `error.log` (grep <module_name> error.log)

ilyam8:
> Are there any pre-defined expected values (as in is there doc for this somewhere)?

No, return only what make sense.

p-thurner:
Error log looks ok:
```
root@monitoring:~# tail -n 200 /var/log/netdata/error.log | grep sslcheck
2019-02-18 15:13:32: python.d INFO: sslcheck: my_sslcheck: check() => [OK]
2019-02-18 15:13:32: netdata ERROR : PLUGINSD[python.d] : File /var/cache/netdata/netdata.runtime_sslcheck_my_sslcheck/main.db is too old. Clearing it.
2019-02-18 15:13:32: netdata INFO  : PLUGINSD[python.d] : File /var/cache/netdata/netdata.runtime_sslcheck_my_sslcheck/run_time.db is too old (last collected 1550502812 seconds ago, but the database is 3996 seconds). Clearing it.
```

p-thurner:
So there are some database files being created now:
```
root@monitoring:~# ls -lha /var/cache/netdata/netdata.runtime_sslcheck_my_sslcheck
total 40K
drwxrwx---   2 netdata netdata 4.0K Feb 18 15:16 .
drwxrwxr-x 184 netdata netdata  12K Feb 18 15:13 ..
-rw-rw----   1 netdata netdata 4.9K Feb 18 15:16 main.db
-rw-rw----   1 netdata netdata  16K Feb 18 15:16 run_time.db
```
Still no graph though :( Do I have to "setup" or "enable" the plugin somewhere else?

ilyam8:
> Error log looks ok:

Decrease uptate_every while depeloping :smile:

p-thurner:
@ilyam8 I set it to `1` with the same lines in error log:

```
root@monitoring:~# cat /etc/netdata/python.d/sslcheck.conf 
my_sslcheck:
  host: my-netdata.io
  port: 443
  timeout: 10
  daysuntilexpiration: 5
  update_every: 1

root@monitoring:~# grep sslcheck /var/log/netdata/error.log |tail
2019-02-18 17:16:17: python.d INFO: sslcheck: my_sslcheck: check() => [OK]
2019-02-18 17:16:17: netdata ERROR : PLUGINSD[python.d] : File /var/cache/netdata/netdata.runtime_sslcheck_my_sslcheck/main.db is too old. Clearing it.
2019-02-18 17:16:17: netdata INFO  : PLUGINSD[python.d] : File /var/cache/netdata/netdata.runtime_sslcheck_my_sslcheck/run_time.db is too old (last collected 1550510177 seconds ago, but the database is 3996 seconds). Clearing it.
```

ilyam8:
i meant decrease, because of [calc next calculation](https://github.com/netdata/netdata/blob/4cf62f7eba36ce4182c87545554f2b2e4b3dc5e3/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SimpleService.py#L41-L43).

Dunno, if it works in debug mode you should see it on the dashboard.

let met check it

ilyam8:
i created sslcheck.chart.py, added your code, restarted `netdata.service` and it works, i see `sslcheck` on the dashboard.

p-thurner:
lol cool ok thanks for the input. Will give it a try tomorrow.

p-thurner:
One last thing I don't get though.. In the WebUI, the graph seems to multiply by 0.1 for some reason. I don't quite get why. When I return a float instead of an int, I get the same result:

screenshot netdata

ilyam8:
```python
[DAYS_UNTIL_EXPIRATION, 'days until expiration', 'absolute', 100, 1000]
```

multiplier is 100, divisor is 1000

what is return value from `ssl_valid_time_remaining` ?

p-thurner:
> multiplier is 100, divisor is 1000
ah ok, thank you!

> what is return value from `ssl_valid_time_remaining` ?
It returns `{'daysuntilexpiration': 67}`

Works for me now. Any further changes recommended?

p-thurner:
what is a good update_time ?

ilyam8:
> Works for me now. Any further changes recommended?

i'll review the PR later in the evening.


> what is a good update_time ?

what is update_time?

p-thurner:
sorry, update_every

collectors/python.d.plugin/sslcheck/README.md
```diff
@@ -0,0 +1,9 @@
+### configuration
+
+```yaml
+master:
+  host: www.my-netdata.io
+  port: 443
+  timeout: 10
+```
+
```
ilyam8: please add more info to the README

Description, requirements if any, what charts it produces, defaults values etc, etc.

[example](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/mongodb)

collectors/python.d.plugin/sslcheck/sslcheck.chart.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
+# Description: simple ssl expiration check netdata python.d module
+# Original Author: ccremer (github.com/ccremer)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import ssl
+import socket
+
+import datetime
```
ilyam8: arrange imports in alphabetical order

collectors/python.d.plugin/sslcheck/sslcheck.chart.py
```diff
@@ -0,0 +1,93 @@
...
+
+#try:
+#    from time import monotonic as time
+#except ImportError:
+#    from time import time
```
ilyam8: remove this

collectors/python.d.plugin/sslcheck/sslcheck.chart.py
```diff
@@ -0,0 +1,93 @@
...
+
+DAYS_UNTIL_EXPIRATION = 'daysuntilexpiration'
+
+ORDER = ['daysuntilexpiration']
+
+CHARTS = {
+    'daysuntilexpiration': {
+        'options': [None, 'days until certificate expiration', 'days', 'days until expiration', 'sslcheck.daysuntilexpiration', 'line'],
+        'lines': [
+            [DAYS_UNTIL_EXPIRATION, 'days until expiration', 'absolute', 10, 0]
```
ilyam8: i suggest to remove `mul` and `div`, do mul/div in `get_data` if needed. And how about to return number of seconds? I think this will be better

collectors/python.d.plugin/sslcheck/sslcheck.chart.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+# Not deriving from SocketService, too much is different
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.host = self.configuration.get('host')
+        self.port = self.configuration.get('port')
```
ilyam8: i suggest to set 443 as default port

collectors/python.d.plugin/sslcheck/sslcheck.chart.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        Parse configuration, check if configuration is available, and dynamically create chart lines data
+        :return: boolean
+        """
+        if self.host is None or self.port is None:
+            self.error('Host or port missing')
+            return False
+        if not isinstance(self.port, int):
+            self.error('"port" is not an integer. Specify a numerical value, not service name.')
+            return False
```
ilyam8: remove that stuff, check only `host`.

```python
if not self.host:
   ...
```

collectors/python.d.plugin/sslcheck/sslcheck.chart.py
```diff
@@ -0,0 +1,93 @@
...
+
+        self.debug('Enabled sslcheck: {host}:{port}, update every {update}s, timeout: {timeout}s'.format(
+            host=self.host, port=self.port, update=self.update_every, timeout=self.timeout
+        ))
+        # We will accept any (valid-ish) configuration, even if initial connection fails (a service
+        # might be down from the beginning)
```
ilyam8: > even if initial connection fails

why? i suggest to return `False` if module failed to collect data

collectors/python.d.plugin/sslcheck/sslcheck.chart.py
```diff
@@ -0,0 +1,93 @@
...
+        ssl_info = conn.getpeercert()
+        # parse the string from the certificate into a Python datetime object
+        return datetime.datetime.strptime(ssl_info['notAfter'], ssl_date_fmt)
+
+    def ssl_valid_time_remaining(self, hostname, port):
+        """Get the number of days left in a cert's lifetime."""
+        expires = self.ssl_expiry_datetime(hostname, port)
+#        print ((expires - datetime.datetime.utcnow()).days * 1.0)
+#        return ((expires - datetime.datetime.utcnow()).days * 10.0)
+        return (expires - datetime.datetime.utcnow()).days
```
ilyam8: convert to seconds and change chart units to `seconds`

collectors/python.d.plugin/sslcheck/sslcheck.conf
```diff
@@ -0,0 +1,72 @@
...
+                               # dashboard (by default is the job_name)
+  host: 'my-netdata.io'        # [required] the remote host address in either IPv4, IPv6 or as DNS name
+  port: 443                    # [required] the port number to check. Specify an integer, not service name
+  daysuntilexpiration: 5       # [required] days before the status is critical?
+  timeout: 10                  # [optional] the socket timeout when connecting
+  update_every: 3600           # [optional] the JOB's data collection frequency. As the chart is in days
+                               # remaining, a high value should be chosen
+  priority: 60000              # [optional] the JOB's order on the dashboard
+  penalty: yes                 # [optional] the JOB's penalty
+
```
ilyam8: i suggest to copy config from another module, this config contains too much portcheck specific things

binaryplease:
SSL certificate expiry would be greatly appreaciated! When can we expect this to be merged?

ilyam8:
@p-thurner do you have time to continue working on the PR?

p-thurner:
@ilyam8 hi, yes, I will finish the plugin, please apologize for the delay! I will follow your instructions.

If you can live with around two weeks that would be great, I have a lot of work atm. I will definitly finish it and then tend to apt, cron and so on.

ilyam8:
> If you can live with around two weeks that would be great

@p-thurner [i see people are interested in this being merged](https://github.com/netdata/netdata/pull/5365#issuecomment-471320507), and most of the requested chages are cosmetic. If you don't mind, i could add them myself (i can commit directly to this PR i believe) and we will have the module in 1.13 (which is released in a couple of days). What do you think?

p-thurner:
@ilyam8 absolutely, please do and thanks for the help!

collectors/python.d.plugin/sslcheck/sslcheck.chart.py
```diff
@@ -0,0 +1,105 @@
...
+
+update_every = 60
+
+
+ORDER = [
+    'time_until_expiration',
+]
+
+CHARTS = {
+        'time_until_expiration': {
```
Ferroin: Indentation is off here, should be 4 spaces.

ilyam8:
Screenshot_20190312_195729

Alarm:
 - warning if `cert exp time < days_until_expiration_warning` (days_until_expiration_warning 5 by default)

health/health.d/sslcheck.conf
```diff
@@ -0,0 +1,9 @@
+
+template: sslcheck_days_until_expiration
+      on: sslcheck.time_until_expiration
+   calc:  $time
+   units: seconds
+   every: 60s
+    warn: $this < $days_until_expiration_warning*24*60*60
```
Ferroin: I'd suggest adding a critical level here too, ideally also configurable, and probably defaulting to one or zero days left until expiration.

ilyam8:
There is a bug in calculation update time

https://github.com/netdata/netdata/blob/6f0d13fbb98978cfca850d72a6b90d6db56e9a2a/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SimpleService.py#L41-L43

```python
In [1]: import time                                                                                               

In [2]: def c(i): 
      :     t = time.monotonic() 
      :     d = t - (t%i) + i 
      :     return t, d, d - t 
      :                                                                                                           

In [3]: c(60)                                                                                                     
Out[3]: (84688.099451557, 84720.0, 31.900548442994477)

In [4]: c(60)                                                                                                     
Out[4]: (84689.748439694, 84720.0, 30.25156030600192)

In [5]: c(60)                                                                                                     
Out[5]: (84693.18335048, 84720.0, 26.816649519998464)

In [6]: c(60)                                                                                                     
Out[6]: (84702.413057355, 84720.0, 17.58694264499354)

In [7]: c(60)                                                                                                     
Out[7]: (84720.166530715, 84780.0, 59.83346928500396)
```

That means that module first poll can be delayed up to `update_every`, which is bad if update_every value is high.

Will be fixed in a separate PR.

health/health.d/sslcheck.conf
ilyam8: I think we don't need both. I suggest to change warning to critical.
What do you think?

ilyam8: i have a second thought about it, if there is only one limit it should be warning, not critical.

But i have no strong opinion on that.

p-thurner: As someone using a _lot_ of letsencrypt, which breaks sometimes (the certbot tool setting weird symlinks and such) I would vote for these defaults:

30 days or less: warning
7 days or less: critical

Please also remember that people might use this for extended validation certs, which (with some less up to date providers) take "time" to renew, especially during holidays and alike.

Ferroin: An SSL certificate nearing it's expiration date is significantly less of an issue than one which is either expired or going to expire within the next 24 hours.  Given that, I think it makes sense to have both, though it might make sense to bump up the default warning threshold to something like 14 days (BCP according to most CA's is to renew at least a month early, so people being sane would not have issues with bogus warnings even then).

p-thurner: 14 days is not enough on around the Christmas holidays for an extended validation cert. I vote for being paranoid here.

p-thurner:
:thumbsup: and thanks a lot for the help @ilyam8 !! :)

Issue #5344, opened 2019-02-07
Allow running multiple freeipmi jobs from the same netdata


jimidak:
#####  Question/Issue:
There is no 'example' in ``` https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/example``` .  I see an example for  charts.py. Is that the boilerplate for a basic test module in Python?

##### OS / Environment 
 Ubuntu 14

##### Component Name
  python.d.plugin

##### Expected results
  A simple py module sample, with instructions. 

I would like to write a plugin in python, to collect remote IPMI results.

Thanks for the help.

ilyam8:
Hi @jimidak 

> How to write python module

https://docs.netdata.cloud/collectors/python.d.plugin/#how-to-write-a-new-module

> to collect remote IPMI results.

https://docs.netdata.cloud/collectors/freeipmi.plugin/#freeipmiplugin

jimidak:
Thanks so much! Let me take a crack at it, and reply back.

jimidak:
Would ``` netdata/collectors/python.d.plugin/example/example.chart.py ``` be a boilerplate for a module written in Python? Would a function/method under the Service class be where I do the "real work"?
I have read about the freeipmi plugin, and I will use that on hosts where I can install freeipmi and ipmimonitoring-dev, but I have some EXSi hosts and other appliances, where I can't do that, and so I wanted to write a module in Python, for remote collecting that IPMI.  I have a Py script that makes calls to remote hosts, and gathers IPMI data... I'm just unsure how to get that in the python.d.plugin dir as a module like the example.charts.py

ilyam8:
> I have a Py script that makes calls to remote hosts, and gathers IPMI data

could you show it please

jimidak:
right now it's just a simple py2 with user input, and a statically set vaule. I will make it py3 and non-interactive, and built it in a class, so I could import it, but this gives you the idea of what I'm currently doing.

```
#!/usr/bin/python
import os
cmd = 'ipmimonitoring -s 88 -h 10.117.6.190 -u root -p xxxx -D LAN_2_0'
value = int(raw_input("What is the ID you would like a status for? :" ))
if value == 88:
    print os.system(cmd)
```

```
$ ./ipmi.py
What is the ID you would like a status for? :88
ID | Name   | Type         | State    | Reading    | Units | Event
88 | Status | Power Supply | Nominal  | N/A        | N/A   | 'Presence detected'
0
```

[...]

PR #5384, opened 2019-02-13
Add driver-type option to the freeipmi plugin

cakrit:
##### Summary
Fixes: #5344 
Permit processing of a `driver-type`  directive, that `ipmimonitoring` uses to set the protocol version for outband (different host) or driver type for inband (same host) communications.

##### Component Name
freeipmi.plugin

##### Additional Information
See `man ipmimonitoring`. The parser functions are based on the code in `freeipmi-1.6.3/common/parsecommon` from https://www.gnu.org/software/freeipmi/download.html. The difference is that `/usr/include/ipmi_monitoring.h` defines different enums, so the return values changed and the `INTELCMD` driver type isn't supported.

### DIFF
--- a/collectors/freeipmi.plugin/README.md
+++ b/collectors/freeipmi.plugin/README.md
@@ -68,6 +68,12 @@ The plugin supports a few options. To see them, run:
   password PASS           connect to remote IPMI host
                           default: local IPMI processor
 
+  driver-type IPMIDRIVER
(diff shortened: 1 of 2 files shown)


jimidak:
That you @cakrit , I cloned your branch of Netdata, and ran the freeipmi plugin, with your new flag/option ```driver-type LAN_2_0```, and it worked!  Thanks for all the investigation and effort.

PR #5395, opened 2019-02-14
Add zero minimum in linux power supply module

vlvkobal:
##### Summary
Add a zero empty/min dimension when it is undefined but the corresponding full/max value exists.

Fixes [missing zero dimensions](https://github.com/netdata/netdata/pull/4960#issuecomment-447041594) in linux power supply module.

##### Component Name
`proc.plugin`

### DIFF
--- a/collectors/proc.plugin/sys_class_power_supply.c
+++ b/collectors/proc.plugin/sys_class_power_supply.c
@@ -19,6 +19,7 @@ struct ps_property_dim {
 
     RRDDIM *rd;
     unsigned long long value;
+    int always_zero;
 
     struct ps_property_dim *next;
 };
@@ -184,6 +185,7 @@ int do_sys_class_power_supply(int update_every, usec_t dt) {
                 for(pr_idx = 0; pr_idx < 3; pr_idx++) {
                     if(unlikely(do_property[pr_idx] != CONFIG_BOOLEAN_NO)) {
                         struct ps_property *pr = NULL;
+                        int min_value_found = 0, max_value_found = 0;
 
                         for(pd_idx = pr_idx * 5; pd_idx < pr_idx * 5 + 5; pd_idx++) {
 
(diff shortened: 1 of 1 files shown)


vlvkobal:
@Ferroin please test it.

collectors/proc.plugin/sys_class_power_supply.c
```diff
@@ -215,6 +222,16 @@ int do_sys_class_power_supply(int update_every, usec_t dt) {
                                 pr->property_dim_root = pd;
                             }
                         }
+
+                        // create a zero empty/min dimension
+                        if(unlikely(max_value_found && !min_value_found)) {
```
Ferroin: I'm not entirely sure that this should be marked as unlikely.  Most of the systems I've seen actually are missing at least one lower bound dimension (usually the `*_min_design` or `*_empty_design` dimensions).

I'm not entirely sure it _shouldn't_ be marked unlikely either though, given that it only really affects startup of the collector, and it shouldn't affect things much.

collectors/proc.plugin/sys_class_power_supply.c
```diff
@@ -258,36 +275,38 @@ int do_sys_class_power_supply(int update_every, usec_t dt) {
             for(pr = ps->property_root; pr && !read_error; pr = pr->next) {
                 struct ps_property_dim *pd;
                 for(pd = pr->property_dim_root; pd; pd = pd->next) {
-                    char buffer[30 + 1];
+                    if(likely(!pd->always_zero)) {
```
Ferroin: This, on the other hand, makes sense to be marked as likely, because most of the systems I have seen have significantly more correct (not-always-zero) dimensions.

Ferroin:
@vlvkobal I probably won't have time to test this today, I'll try to get it tested this weekend.  I've added a couple of comments to the code itself, but other than those, it looks generally correct to me.

ktsaou:
As we discussed this module has a few issues to improve maintainability.
@cakrit 	we can merge this, but it needs some cleanup.

PR #5396, opened 2019-02-15
Fix ceph.chart.py for Python3

GaetanF:
With Python3, Ceph plugin failed with error : 
```
2019-02-14 18:58:14: python.d ERROR: ceph: ceph: check() unhandled exception: the JSON object must be str, not 'bytes'
2019-02-14 18:58:14: python.d INFO: ceph: ceph: check() => [FAILED]
```

To fix them, we need to decode the bytes to string and fix the sort function of dict using keys.

### DIFF
--- a/collectors/python.d.plugin/ceph/ceph.chart.py
+++ b/collectors/python.d.plugin/ceph/ceph.chart.py
@@ -161,7 +161,7 @@ class Service(SimpleService):
         :return: None
         """
         # Pool lines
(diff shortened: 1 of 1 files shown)


ilyam8:
@GaetanF hi

have you tested it with both py2 and py3?

GaetanF:
Yes ! The same script works with Python 2 and Python 3

ilyam8:
nice, thanks!

PR #5403, opened 2019-02-16
Fix #5402: sensors plugin only plots sensors that have hard-coded limits

### LINKED ISSUE #5402: sensors plugin only plots sensors that have hard-coded limits
`sensors.chart.py` has hard-coded limits for some sensor types. When querying sensor values, the `get_data` method checks whether there is a limit for them and discards values that are out of range. However, due to an indentation mistake, it also discards values if there are no limits for its sensor type. This is clearly not intentional, since the script has extensive code parts for other sensor types as well.

This issue can easily be fixed by correcting the indentation of the misindented line, which reveals another problem: `get_data` scales values by 1000 but the `CHARTS` constant specifies a scale factor of 1000000 for some sensors, resulting in plotting 1/1000th of the actual values for affected sensors.

zivanfi:
`sensors.chart.py` has hard-coded limits for some sensor types. When querying sensor values, the `get_data` method checks whether there is a limit for them and discards values that are out of range. However, due to an indentation mistake, it also discards values if there are no limits for its sensor type. This is clearly not intentional, since the script has extensive code parts for other sensor types as well.

This issue can easily be fixed by correcting the indentation of the misindented line, which reveals another problem: `get_data` scales values by 1000 but the `CHARTS` constant specifies a scale factor of 1000000 for some sensors, resulting in plotting 1/1000th of the actual values for affected sensors.

Fixes #5402.

### DIFF
--- a/collectors/python.d.plugin/sensors/sensors.chart.py
+++ b/collectors/python.d.plugin/sensors/sensors.chart.py
@@ -41,7 +41,7 @@ CHARTS = {
     'power': {
         'options': [None, ' power', 'Watt', 'power', 'sensors.power', 'line'],
         'lines': [
-            [None, None, 'absolute', 1, 1000000]
+            [None, None, 'absolute', 1, 1000]
         ]
     },
     'fan': {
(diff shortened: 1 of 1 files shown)


ilyam8:
@zivanfi thanks!

ilyam8:
@zivanfi please sign the CLA

zivanfi:
Sorry, I was not aware that I have to sign a CLA when I opened the pull request. I have read the CLA since then and I found it rather vague. I would not feel comfortable signing it, therefore I'm closing my pull request.

PR #5431, opened 2019-02-19
fix #5430 - LogService._get_raw_data under python3 fails on undecodable data

### LINKED ISSUE #5430: Under python3 the LogService._get_raw_data fails on undecodable data
##### Bug report summary
Under python3 the LogService._get_raw_data fails with an exception when undecodable bytes are read from a log file (e.g. the file encoding is different from what `locale.getpreferredencoding()` specifies or some garbage data is logged by the monitored application). This blocks any LogService-based collector until a log rotation happens or Netdata is restarted (the method will retry reading the same line every time it is invoked).

##### OS / Environment
Arch Linux, python3.7

v1.12.0-46-gf1d69813 (built from master)

##### Component Name
collectors/python.d.plugin

##### Steps To Reproduce
As an example with fail2ban module:
```
head -c 200 < /dev/urandom >> /var/log/fail2ban.log 
```
This will result with having error.log flooded with
```
python.d ERROR: fail2ban: fail2ban: update() unhandled exception: 'utf-8' codec can't decode byte 0xbd in position 0: invalid start byte
```

##### Expected behavior
_get_raw_data ignores the encoding errors by skipping invalid bytes or the entire line.

apardyl:
##### Summary
Fixes: #5430
A quick fix for #5430 - set decoding error handling to `replace` instead of the default `None`/`strict` for python3

##### Component Name
collectors/python.d.plugin

##### Additional Information
in the reported bug: #5430

### DIFF
--- a/collectors/python.d.plugin/python_modules/bases/FrameworkServices/LogService.py
+++ b/collectors/python.d.plugin/python_modules/bases/FrameworkServices/LogService.py
@@ -18,6 +18,7 @@ class LogService(SimpleService):
         self.__glob_path = self.log_path
         self._last_position = 0
         self.__re_find = dict(current=0, run=0, maximum=60)
+        self.open_args = {'errors': 'replace'} if sys.version_info[0] > 2 else {}
 
     def _get_raw_data(self):
         """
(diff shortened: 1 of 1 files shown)


collectors/python.d.plugin/python_modules/bases/FrameworkServices/LogService.py
```diff
@@ -34,7 +35,8 @@ def _get_raw_data(self):
             elif size < self._last_position:
                 self._last_position = 0  # read from beginning if file has shrunk
 
-            with open(self.log_path) as fp:
+            open_args = {'errors': 'replace'} if sys.version_info[0] > 2 else {}
```
ilyam8: do it only once, not on every iteration

ilyam8:
edited OP post

collectors/python.d.plugin/python_modules/bases/FrameworkServices/LogService.py
apardyl: moved to init

PR #5435, opened 2019-02-19
apcupsd - Treat ONBATT the same as ONLINE

### LINKED ISSUE #5428: apcupsd chart not updating when on battery
UPS: SMT1500I connected via USB
OS: Fedora

Current issue: When the UPS goes onto battery, the netdata plugin seems to not gather any further data until the UPS comes back onto power. 

apcaccess status shows the correct values however, so the UPS is certainly communicating correctly with the server. 

capture1
capture2
capture3

cakrit:
##### Summary
Fixes #5428 

##### Component Name
apcupsd plugin

##### Additional Information
Status ONBATT should be equivalent to status ONLINE


cakrit:
Please hold reviews until the reporter tests the solution.

Issue #5467, opened 2019-02-22
go.d.plugin installation fails due to insufficient timeout


elfgoh:
##### Bug report summary
In my jenkins log, netdata fails to install due to a failed download of `go.d.plugin`

```
Fri Feb 22 12:40:36 +08 2019 [INFO] [/tmp/netdata-kickstart-HbDvAH/netdata-v1.12.1-1-gf2ca1ab9]# curl -L --connect-timeout 5 --retry 3 https://github.com/netdata/go.d.plugin/releases/download/v0.1.0/go.d.plugin-v0.1.0.linux-arm 
Fri Feb 22 12:40:37 +08 2019 [INFO]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Fri Feb 22 12:40:37 +08 2019 [INFO]                                  Dload  Upload   Total   Spent    Left  Speed
Fri Feb 22 12:40:42 +08 2019 [INFO] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
Fri Feb 22 12:40:42 +08 2019 [INFO] Warning: Transient problem: timeout Will retry in 1 seconds. 3 retries left.
Fri Feb 22 12:40:49 +08 2019 [INFO] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
[...]

PR #5476, opened 2019-02-22
Fix curl timeout

elfgoh:
##### Summary
Fixes [timeout issue for spotty network connections](https://github.com/netdata/netdata/issues/5467#issuecomment-466582869). Previous fix #5468 did not address curl commands in other parts  of netdata. This has lead to installation failures

This PR increases curl timeout to address those failures

##### Component Name
Installer


gmosx:
Good catch.

elfgoh:
This isnt exactly fixed even though it has been merged https://github.com/netdata/netdata/issues/5492

PR #5497, opened 2019-02-25
prevent localhost to be received from slaves; fixes #5488

### LINKED ISSUE #5488: Need more graceful handling of duplicate machine GUIDs
##### Bug report summary

We spent a lot of time trying to identify the root cause of some problems in a master slave setup. The root cause ended up being that one of the slaves was created by cloning the master, taking the same machine GUID with it. We should have at least a detection of such cases, so we can warn users about it.

##### Steps To Reproduce
Master and slave with the same GUID.
The master hostname configured in netdata.conf is overwritten by the slave's hostname and we have issues with the charts as well. 

##### Expected behavior
Report an error, or have a self-cleaning behavior (e.g. regenerate a machine GUID when a duplicate is reported).

ktsaou:
##### Summary

netdata already has a protection for duplicate streaming of the same host.
This is controlled with `multiple connections` in `stream.conf`.

The default is `allow`, which allows multiple connections. Users can set it to `deny` to prevent duplicate connections for the same host (new connections are rejected until the old connection disconnects).

However, this check did not work for `localhost`.

So, this PR prevents `localhost` to be received from remote netdata servers.

##### Component Name

streaming

##### Additional Information

fixes #5488

### DIFF
--- a/streaming/rrdpush.c
+++ b/streaming/rrdpush.c
@@ -897,6 +897,13 @@ static int rrdpush_receive(int fd
         return 1;
     }
 
+    if(host == localhost) {
+        close(fd);
+        log_stream_connection(client_ip, client_port, key, machine_guid, hostname, "DENIED - ATTEMPT TO RECEIVE METRICS FOR LOCALHOST MACHINE_GUID");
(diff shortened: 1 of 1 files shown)


cakrit:
I asked @paulkatsoulakis to test, we'll merge immediately after he responds

cakrit:
Fixing it in #5511 instead

Issue #5491, opened 2019-02-25
centos6: python.d.plugin zombie process after netdata restart


Wing924:
##### Bug report summary

Simetimes, zombie processes exist after restarting netdata.

```
$ ps faxu | grep netdata | grep -v grep
netdata   3931  0.2  4.2 139372 20112 ?        S    13:00   0:01 /usr/bin/python /opt/netdata/usr/libexec/netdata/plugins.d/python.d.plugin 1
netdata   4244  0.0  0.0      0     0 ?        Z    13:00   0:00  \_ [postqueue] <defunct>
netdata   4295  0.3  3.1  15924 14592 ?        Sl   13:00   0:01 /opt/netdata/bin/srv/netdata -P /opt/netdata/var/run/netdata/netdata.pid
netdata   4316  0.1  3.8 137428 18124 ?        S    13:00   0:00  \_ /usr/bin/python /opt/netdata/usr/libexec/netdata/plugins.d/python.d.plugin 1
netdata   4608  0.0  0.0      0     0 ?        Z    13:00   0:00  |   \_ [postqueue] <defunct>
netdata   4317  0.0  1.1 110840  5260 ?        Sl   13:00   0:00  \_ /opt/netdata/usr/libexec/netdata/plugins.d/go.d.plugin 1
root      4318  0.5  0.2   1456  1020 ?        S    13:00   0:02  \_ /opt/netdata/usr/libexec/netdata/plugins.d/apps.plugin 1
```

##### OS / Environment

OS: CentOS release 6.3 (Final)
kernel: 2.6.32-696.18.7.el6.x86_64

##### Netdata version (ouput of `netdata -V`)

netdata v1.12.1 (static binary)
(compile from source don't have this issue)

##### Component Name

daemon

##### Steps To Reproduce

service netdata restart
ps aux | grep netdata

##### Expected behavior

netdata children processes should exist when parent process exists.

ilyam8:
what i did:
 * installed centos6 (`2.6.32-754.10.1.el6.x86_64`)
 * installed netdata v1.12.1 static binary

Problem:

i see no SIGHUP for `python.d.plugin` process on `service netdata restart` (in strace output).

Python process dies only on SIGPIPE. If it doesn't write anything it will stay forever.

Solution:
 - add a _heartbeat_ func to python.d.plugin


Still it is not clear why there is no SIGHUP on netdata process restart.

cakrit:
[...]

PR #5503, opened 2019-02-25
python.d.plugin add heartbeat

ilyam8:
##### Summary

Fixes: #5491

##### Component Name

[`collectors/python.d.plugin`](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/python.d.plugin.in)

##### Additional Information

Add a simple print func, that prints '\n' every second to stdout. This is needed to exit on SIGPIPE.


ilyam8:
@cakrit please approve

Issue #4131, opened 2018-09-07
SpigotMC Chart won't show


alsoGAMER:
I don't know why but spigotmc chart won't show on netdata's dashboard..
Here screenshot of dashboard, configuration of spigotmc server and chart, also the log of the spigot and netdata server.

![screenshot_2018-09-07 get control of your linux servers simple effective awesome](https://user-images.githubusercontent.com/35269770/45239030-1f2b0200-b2e4-11e8-9281-5366a3286988.png)
[error.log](https://github.com/firehol/netdata/files/2362254/error.log)
[latest.log](https://github.com/firehol/netdata/files/2362255/latest.log)
[config file.zip](https://github.com/firehol/netdata/files/2362259/config.file.zip)

Komic:
`enable-query` needs to be set to `true` (that's a server.properties setting). Restart Spigot then netdata after that.

Ferroin:
Based on the config, everything looks correct to me.  The error log indicates that the module's check method failed, which means that it could not connect for some reason, but the usual reasons (mismatched password, port, or hostname, or `rcon-enable` being false), so I'm not quite sure exactly what is going on here.  I'm going to look into throwing together a version of the module with better debugging info so we can hopefully figure out exactly why this is failing.

In the meantime, feel free to try @Komic's suggestion, though I would be surprised if it works (we don't use the GameSpy4 interface for querying the server, so if it does fix things, that means that the meaning of that option has changed and the Minecraft documentation needs updated).

Komic:
Oh my bad! I thought 25575 was the query port for some reason, please ignore my previous comment. @alsoGAMER, did you make sure to edit `/etc/netdata/python.d/spigotmc.conf` properly?

Ferroin:
#4148 adds some extra logging to the spigotmc module that should hopefully help track down exactly what is wrong here (at minimum, it will tell us if it's a socket error or a protocol error).

@alsoGAMER If you want to test with the updated module without having to go through installing a test build of Netdata, you can fetch the module file with the changes from [here](https://raw.githubusercontent.com/Ferroin/netdata/spigotmc-extra-debugging/python.d/spigotmc.chart.py) and just drop it in `/usr/libexec/netdta/python.d` and restart netdata (and when you're done checking, you can revert it by just updating or reinstalling netdata however you normally would).

alsoGAMER:
And nothing...
[...]

PR #5507, opened 2019-02-25
Convert SpigotMC module to use regexes for parsing.

Ferroin:
##### Summary

This converts the SpigotMC python module to use regular expressions for parsing, which makes the parsing a bit more robust at the cost of some performance.

Additionally, this relaxes the parsing logic surrounding the handling of the user counts to just grab the first integer after the timestamp and treat it as the count of users.

##### Component Name

python.d.plugin / spigotmc.chart.py

##### Additional Information

Fixes: #4131

-----

I've verified the regexes against sample console output I had saved for testing though, as well as the console outputs posted in #4131, so while I haven't directly tested the code myself (I don't have a working SpigotMC installation that I can test against right now, and I unfortunately don't have the time to get one working right now either) it _should_ work, but it would be good to have someone who has a working SpigotMC installation test this.

### DIFF
--- a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
+++ b/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
@@ -107,9 +107,9 @@ class Service(SimpleService):
             raw = self.console.command('tps')
             match = _TPS_REGEX.match(raw)
             if match:
-                data['tps1'] = int(match.group(1)) * PRECISION
-                data['tps5'] = int(match.group(2)) * PRECISION
-                data['tps15'] = int(match.group(3)) * PRECISION
+                data['tps1'] = int(float(match.group(1)) * PRECISION)
+                data['tps5'] = int(float(match.group(2)) * PRECISION)
+                data['tps15'] = int(float(match.group(3)) * PRECISION)
             else:
                 self.error('Unable to process TPS values.')
         except mcrcon.MCRconException:
--- a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
+++ b/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
@@ -118,8 +118,6 @@ class Service(SimpleService):
             self.error('Connection is dead.')
             self.alive = False
             return None
-        except (TypeError, LookupError, IndexError):
-            self.error('Unable to process TPS values.')
         try:
             raw = self.console.command('list')
             match = _LIST_REGEX.match(raw)
@@ -133,6 +131,4 @@ class Service(SimpleService):
             self.error('Connection is dead.')
             self.alive = False
             return None
-        except (TypeError, LookupError):
-            self.error('Unable to process user counts.')
         return data
--- a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
+++ b/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
@@ -46,9 +46,7 @@ _TPS_REGEX = re.compile(
     re.X | re.A
 )
 _LIST_REGEX = re.compile(
-    r'^.*?'       # Message lead-in
-    r'(\d+)'      # Current user count.
-    r'.*?\d*.*$', # Rest of the line, which should include another number.
+    r'(\d+)', # Current user count.
     re.X | re.A
 )
 
@@ -120,10 +118,10 @@ class Service(SimpleService):
             return None
         try:
             raw = self.console.command('list')
-            match = _LIST_REGEX.match(raw)
+            match = _LIST_REGEX.search(raw)
             if not match:
                 raw = self.console.command('online')
-                match = _LIST_REGEX.match(raw)
+                match = _LIST_REGEX.search(raw)
             if match:
                 data['users'] = int(match.group(1))
             else:
--- a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
+++ b/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
@@ -39,17 +39,17 @@ CHARTS = {
     }
 }
 _TPS_REGEX = re.compile(
-    r'^.*: .*?'                 # Message lead-in
-    r'([0-9]{1,2}.[0-9]+), .*?' # 1-minute TPS value
-    r'([0-9]{1,2}.[0-9]+), .*?' # 5-minute TPS value
-    r'([0-9]{1,2}\.[0-9]+).*$', # 15-minute TPS value
-    re.X
+    r'^.*: .*?'           # Message lead-in
+    r'(\d{1,2}.\d+), .*?' # 1-minute TPS value
+    r'(\d{1,2}.\d+), .*?' # 5-minute TPS value
+    r'(\d{1,2}\.\d+).*$', # 15-minute TPS value
+    re.X | re.A
 )
 _LIST_REGEX = re.compile(
-    r'^.*?'          # Message lead-in
-    r'([0-9]*)'      # Current user count.
-    r'.*?[0-9]*.*$', # Rest of the line, which should include another number.
-    re.X
+    r'^.*?'       # Message lead-in
+    r'(\d+)'      # Current user count.
+    r'.*?\d*.*$', # Rest of the line, which should include another number.
+    re.X | re.A
 )
 
 
--- a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
+++ b/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
@@ -121,6 +121,9 @@ class Service(SimpleService):
         try:
             raw = self.console.command('list')
             match = _LIST_REGEX.match(raw)
+            if !match:
+                raw = self.console.command('online')
+                match = _LIST_REGEX.match(raw)
             if match:
                 data['users'] = int(match.group(1))
             else:
--- a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
+++ b/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
@@ -5,6 +5,7 @@
 
 import socket
 import platform
+import re
 
 from bases.FrameworkServices.SimpleService import SimpleService
 
@@ -49,6 +50,10 @@ class Service(SimpleService):
         self.password = self.configuration.get('password', '')
         self.console = mcrcon.MCRcon()
         self.alive = True
+        self._tps_regex = re.compile(
+            b'^.*: ([0-9]{1,2}.[0-9]{1,2}), .*?([0-9]{1,2}.[0-9]{1,2}), .*?([0-9]{1,2}.[0-9]{1,2}).*$'
+        )
+        self._list_regex = re.compile(b'^.*: .*?([0-9]*).*?[0-9]*.*$')
 
     def check(self):
         if platform.system() != 'Linux':
@@ -91,27 +96,28 @@ class Service(SimpleService):
         data = {}
         try:
             raw = self.console.command('tps')
-            # The above command returns a string that looks like this:
-            # '§6TPS from last 1m, 5m, 15m: §a19.99, §a19.99, §a19.99\n'
-            # The values we care about are the three numbers after the :
-            tmp = raw.split(':')[1].split(',')
-            data['tps1'] = float(tmp[0].lstrip(u' §a*')) * PRECISION
-            data['tps5'] = float(tmp[1].lstrip(u' §a*')) * PRECISION
-            data['tps15'] = float(tmp[2].lstrip(u' §a*').rstrip()) * PRECISION
+            match = self._list_regex.match(raw)
+            if match:
+                data['tps1'] = int(match.group(1)) * PRECISION
+                data['tps5'] = int(match.group(2)) * PRECISION
+                data['tps15'] = int(match.group(3)) * PRECISION
+            else:
+                self.error('Unable to process TPS values.')
         except mcrcon.MCRconException:
             self.error('Unable to fetch TPS values.')
         except socket.error:
             self.error('Connection is dead.')
             self.alive = False
             return None
-        except (TypeError, LookupError):
+        except (TypeError, LookupError, IndexError):
             self.error('Unable to process TPS values.')
         try:
             raw = self.console.command('list')
-            # The above command returns a string that looks like this:
-            # 'There are 0/20 players online:'
-            # We care about the first number here.
-            data['users'] = int(raw.split()[2].split('/')[0])
+            match = self._list_regex.match(raw)
+            if match:
+                data['users'] = int(match.group(1))
+            else:
+                self.error('Unable to process user counts.')
         except mcrcon.MCRconException:
             self.error('Unable to fetch user counts.')
         except socket.error:
--- a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
+++ b/collectors/python.d.plugin/spigotmc/spigotmc.chart.py
@@ -39,16 +39,16 @@ CHARTS = {
     }
 }
 _TPS_REGEX = re.compile(
-    '^.*: .*?'                     # Message lead-in
-    '([0-9]{1,2}.[0-9]{1,2}), .*?' # 1-minute TPS value
-    '([0-9]{1,2}.[0-9]{1,2}), .*?' # 5-minute TPS value
-    '([0-9]{1,2}.[0-9]{1,2}).*$',  # 15-minute TPS value
+    r'^.*: .*?'                 # Message lead-in
+    r'([0-9]{1,2}.[0-9]+), .*?' # 1-minute TPS value
+    r'([0-9]{1,2}.[0-9]+), .*?' # 5-minute TPS value
+    r'([0-9]{1,2}\.[0-9]+).*$', # 15-minute TPS value
     re.X
 )
 _LIST_REGEX = re.compile(
-    '^.*?'          # Message lead-in
-    '([0-9]*)'      # Current user count.
-    '.*?[0-9]*.*$', # Rest of the line, which should include another number.
+    r'^.*?'          # Message lead-in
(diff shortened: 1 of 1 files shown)


collectors/python.d.plugin/spigotmc/spigotmc.chart.py
```diff
@@ -49,6 +50,10 @@ def __init__(self, configuration=None, name=None):
         self.password = self.configuration.get('password', '')
         self.console = mcrcon.MCRcon()
         self.alive = True
+        self._tps_regex = re.compile(
+            b'^.*: ([0-9]{1,2}.[0-9]{1,2}), .*?([0-9]{1,2}.[0-9]{1,2}), .*?([0-9]{1,2}.[0-9]{1,2}).*$'
+        )
+        self._list_regex = re.compile(b'^.*: .*?([0-9]*).*?[0-9]*.*$')
```
ilyam8: i suggest to make those regexes global. 

And please use

```
(?P<name>...)
Similar to regular parentheses, but the substring matched by the group is accessible via the symbolic group name name. Group names must be valid Python identifiers, and each group name must be defined only once within a regular expression. A symbolic group is also a numbered group, just as if the group were not named.
```

this is much clear

ilyam8: or maybe something like this

https://github.com/netdata/netdata/blob/2ee2d94891e88d9c773f7633746a38fb341f2bac/collectors/python.d.plugin/smartd_log/smartd_log.chart.py#L63-L74

ilyam8: it is nice to have comments:

```python
re.compile( 
     '(\d+);'  # attribute 
     '(\d+);'  # normalized value 
     '(\d+)',  # raw value 
     re.X 
 )
```
vs
```python
re.compile('(\d+);(\d+);(\d+)')
```

Ferroin: Doh!  Should have though of both points.  I've been too far out of Python since the holidays...

collectors/python.d.plugin/spigotmc/spigotmc.chart.py
```diff
@@ -91,27 +105,28 @@ def _get_data(self):
...
+            else:
+                self.error('Unable to process TPS values.')
         except mcrcon.MCRconException:
             self.error('Unable to fetch TPS values.')
         except socket.error:
             self.error('Connection is dead.')
             self.alive = False
             return None
-        except (TypeError, LookupError):
+        except (TypeError, LookupError, IndexError):
```
ilyam8: if regex matches we always have 3 x `([0-9]{1,2}.[0-9]{1,2})`

do we need this?

> except (TypeError, LookupError, IndexError):

Ferroin: Probably not.

collectors/python.d.plugin/spigotmc/spigotmc.chart.py
```diff
@@ -91,33 +105,30 @@ def _get_data(self):
...
-            # The values we care about are the three numbers after the :
-            tmp = raw.split(':')[1].split(',')
-            data['tps1'] = float(tmp[0].lstrip(u' §a*')) * PRECISION
-            data['tps5'] = float(tmp[1].lstrip(u' §a*')) * PRECISION
-            data['tps15'] = float(tmp[2].lstrip(u' §a*').rstrip()) * PRECISION
+            match = _TPS_REGEX.match(raw)
+            if match:
+                data['tps1'] = int(match.group(1)) * PRECISION
+                data['tps5'] = int(match.group(2)) * PRECISION
+                data['tps15'] = int(match.group(3)) * PRECISION
```
ilyam8: should be

```
int(float(match.group(1)) * PRECISION)
```

ilyam8:
@Ferroin 

`([0-9]{1,2}.[0-9]{1,2})`

are you sure about `{1,2}` ?

collectors/python.d.plugin/spigotmc/spigotmc.chart.py
Ferroin: Again, too much time away from Python.  I've gotten used to JS having only one numeric type for everything...

Ferroin:
For the integer part, yes, because the first number isn't guaranteed to be zero padded, but can't realistically be more than two digits.

For the decimal part, not so much, but I don't have any better suggestions (unless of course using a `+` is significantly faster,  in which case it should probably be switched to that).

ilyam8:
Idea

```python
re.findall(
	r'([0-9]{1,2}\.[0-9]+)',
	'§6TPS from last 1m, 5m, 15m: §a19.99: , §a19.99, §a19.99\n'
)

```

Ferroin:
@ilyam8 The problem is that I'm not 100% certain that won't match other things in the string too (the garbage bytes appear to be just that, consistent garbage, so i can't be sure they don't change).  By just matching the whole line we also get a much nicer indication if the format ever changes (instead of randomly wrong values, it will refuse to parse).

collectors/python.d.plugin/spigotmc/spigotmc.chart.py
```diff
@@ -37,6 +38,19 @@
         ]
     }
 }
+_TPS_REGEX = re.compile(
+    '^.*: .*?'                     # Message lead-in
+    '([0-9]{1,2}.[0-9]{1,2}), .*?' # 1-minute TPS value
```
ilyam8: if format is `1.11` and not `01.11` should be `([1-9]{1,2}.[0-9]{1,2})`

and if you unsure about decimal part please do [0-9]+

collectors/python.d.plugin/spigotmc/spigotmc.chart.py
```diff
@@ -37,6 +38,19 @@
...
+    r'([0-9]{1,2}.[0-9]+), .*?' # 1-minute TPS value
+    r'([0-9]{1,2}.[0-9]+), .*?' # 5-minute TPS value
+    r'([0-9]{1,2}\.[0-9]+).*$', # 15-minute TPS value
+    re.X
+)
+_LIST_REGEX = re.compile(
+    r'^.*?'          # Message lead-in
+    r'([0-9]*)'      # Current user count.
+    r'.*?[0-9]*.*$', # Rest of the line, which should include another number.
+    re.X
```
ilyam8: i see from your example the response is

'There are 0/20 players online:'`

But regex looks too complicated for that string. Reasons?

Ferroin: Because I meant to add logic to try a different command if we get nothing from `list`, and forgot to do so.

See https://github.com/netdata/netdata/issues/4131#issuecomment-420553899 for the info on the command and what it's output format is.

Pretty much, with this regex, we can successfully match the original `list` command's output, or the output from the `online` command mentioned in that comment with the same regex, and in both cases we don't have to care as much about surrounding formatting.

collectors/python.d.plugin/spigotmc/spigotmc.chart.py
```diff
@@ -91,33 +105,33 @@ def _get_data(self):
...
-        except (TypeError, LookupError):
-            self.error('Unable to process TPS values.')
         try:
             raw = self.console.command('list')
-            # The above command returns a string that looks like this:
-            # 'There are 0/20 players online:'
-            # We care about the first number here.
-            data['users'] = int(raw.split()[2].split('/')[0])
+            match = _LIST_REGEX.match(raw)
+            if !match:
```
ilyam8: python has no `!`, use `not match`

collectors/python.d.plugin/spigotmc/spigotmc.chart.py
ilyam8:
```python
In [8]: _LIST_REGEX = re.compile( 
      :     r'^.*?'          # Message lead-in 
      :     r'([0-9]*)'      # Current user count. 
      :     r'.*?[0-9]*.*$', # Rest of the line, which should include another num
      : ber. 
      :     re.X)                                                                

In [9]: _LIST_REGEX.match('There are 0/20 players online:').groups()             
Out[9]: ('',)

```

collectors/python.d.plugin/spigotmc/spigotmc.chart.py
```diff
@@ -37,6 +38,19 @@
...
+    r'(\d{1,2}.\d+), .*?' # 5-minute TPS value
+    r'(\d{1,2}\.\d+).*$', # 15-minute TPS value
+    re.X | re.A
+)
+_LIST_REGEX = re.compile(
+    r'^.*?'       # Message lead-in
+    r'(\d+)'      # Current user count.
+    r'.*?\d*.*$', # Rest of the line, which should include another number.
+    re.X | re.A
+)
```
ilyam8: i suggest to make regex more specific

```python
In [11]: _LIST_REGEX.match('There are 0/20 players online:').groups()            
Out[11]: ('0',)

In [12]: _LIST_REGEX.match('4').groups()                                         
Out[12]: ('4',)

In [13]: _LIST_REGEX.match('always matches! 33').groups()                        
Out[13]: ('33',)
```

ilyam8: if i understand it right you don't want to match any row if it has a number

Ferroin: In this case, we're only pulling the first line of the output (because of the `^` and the lack of the `re.MULTILINE` switch), so we don't really need to worry beyond that line, which should look like either `There are 0/20 players online:` if the `list` command works, or `There are 0 players online on a maximum of 50.` if it doesn't and we have to use the `online` command.  If a better option can be found that can reliably match both independent of the words and formatting (because those are the most likely things to change), I'm all for it, but what's here does work to cover both cases.

ilyam8: if you want to match the first number you can use `(\d+)` (search) or, if you want to be more precise, `There are (\d+)` (match).

ilyam8: your regex does the same actually, it matches first group of decimals, can be simplified to `(\d+)` (search)

ilyam8:
@alsoGAMER if you have time please test it, you are the only `spigotmc` user ~in the world~ we know

Ferroin:
I'm still hoping to get some feedback one way or another from @alsoGAMER or someone else who is currently using the module on this.

Assuming it's OK with you @ilyam8, I plan to wait until next Monday, and if we still haven't heard anything by then, just merge it.

elfhack:
Just a headsup, I dropped this version into a running 1.12.0 in the official docker image, and it errors out with a 

2019-03-10 00:42:15: python.d ERROR: plugin: main: module load source: 'spigotmc' => [FAILED]
2019-03-10 00:42:15: python.d ERROR: plugin: main: load source error : 'module' object has no attribute 'A'

ilyam8:
@elfhack 	

i see

https://github.com/netdata/netdata/blob/d2a183e9c0008937b93f918d41d188b8f46caa3a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py#L114

and

https://github.com/netdata/netdata/blob/d2a183e9c0008937b93f918d41d188b8f46caa3a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py#L128

> I've removed it, and unfortunately it still errors out with "Unable to process"

TPS values or user counter?

___ 

plese add 

```python
self.info('RAW TPS :', raw)
```

right after

https://github.com/netdata/netdata/blob/d2a183e9c0008937b93f918d41d188b8f46caa3a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py#L105

```python
self.info('RAW LIST :', raw)
```

right after

https://github.com/netdata/netdata/blob/d2a183e9c0008937b93f918d41d188b8f46caa3a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py#L120

and 

```python
self.info('RAW ONLINE :', raw)
```

after

https://github.com/netdata/netdata/blob/d2a183e9c0008937b93f918d41d188b8f46caa3a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py#L123


and run `spigotmc module` in debug mode, lets see what's the problem

https://docs.netdata.cloud/collectors/python.d.plugin/#how-to-debug-a-python-module

elfhack:
Did as you asked, i think the problem is somewhere deeper.
```
2019-03-10 14:49:09: python.d INFO: plugin: main: Using python 2
2019-03-10 14:49:09: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d.conf'
2019-03-10 14:49:09: python.d ERROR: plugin: main: cannot load '/etc/netdata/python.d.conf': [Errno 2] No such file or directory: '/etc/netdata/python.d.conf'. Will try stock version.
2019-03-10 14:49:09: python.d DEBUG: plugin: main: loading '/usr/lib/netdata/conf.d/python.d.conf'
2019-03-10 14:49:09: python.d DEBUG: plugin: main: module load source: 'spigotmc' => [OK]
2019-03-10 14:49:09: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d/spigotmc.conf'
2019-03-10 14:49:09: python.d DEBUG: plugin: main: job initialization: 'spigotmc spigot' => ['OK']
2019-03-10 14:49:09: python.d DEBUG: plugin: main: module status: 'spigotmc' => [OK] (jobs: 1)
2019-03-10 14:49:09: python.d INFO: spigotmc: spigot: check() => [OK]
CHART netdata.runtime_spigotmc_spigot '' 'Execution time for spigotmc_spigot' 'ms' 'python.d' netdata.pythond_runtime line 145000 5

2019-03-10 14:49:09: python.d DEBUG: spigotmc: spigot: create() => [OK] (charts: 2)
2019-03-10 14:49:09: python.d DEBUG: spigotmc: spigot: started, update frequency: 5
2019-03-10 14:49:13: python.d INFO: spigotmc: spigot: RAW LIST : 
2019-03-10 14:49:13: python.d ERROR: spigotmc: spigot: Unable to process TPS values.
2019-03-10 14:49:13: python.d INFO: spigotmc: spigot: RAW List: 
2019-03-10 14:49:13: python.d INFO: spigotmc: spigot: RAW Online: 
2019-03-10 14:49:13: python.d ERROR: spigotmc: spigot: Unable to process user counts.
2019-03-10 14:49:13: python.d DEBUG: spigotmc: spigot: get_data() returned no data
2019-03-10 14:49:13: python.d DEBUG: spigotmc: spigot: update => [FAILED] (elapsed time: -, failed retries in a row: 1)
```

However I can see it connecting to rcon (do not mind the hour difference, alpine is missing localtime)
```
[13:49:09 INFO]: Rcon connection from: /172.18.0.2
```

ilyam8:
well, i see

all these commands

https://github.com/netdata/netdata/blob/d2a183e9c0008937b93f918d41d188b8f46caa3a/collectors/python.d.plugin/spigotmc/spigotmc.chart.py#L105

elfhack:
Yeah, but talking to rcon gives back what you'd expect
```
➜ docker exec spigot rcon-cli list
There are 0 of a max 20 players online: 

➜ docker exec spigot rcon-cli tps 
§6TPS from last 1m, 5m, 15m: §a20.0, §a19.99, §a19.99
```

ilyam8:
@elfhack 

Take a look

I install spigotmc

> docker run -d -p 25575:25575 -e EULA=TRUE  -e RCON_PASSWORD=testing --name mc itzg/minecraft-server

that is python script

```python
import sys
import mcrcon

m = mcrcon.MCRcon()
m.connect('127.0.0.1', 25575, sys.argv[1])
print(m.command('list'))
```

result

```bash
[lgz :~]python spigot.py wrong

[lgz :~]python spigot.py testing
There are 0 of a max 20 players online: 
[lgz :~]
```

There is no `authentication error` or something, just no values

elfhack:
@ilyam8 

Hmm, I use the exact same container, but I don't declare any RCON_PASSWORD. I'll try defining one in a couple of minutes

ilyam8:
> but I don't declare any RCON_PASSWORD

this is the problem

elfhack:
Yeah, just grepped through my server.properties. That was dumb

ilyam8:
ok, py2 compatibility issue fixed in #5593, will have it merged in a minute

elfhack:
Yep, once i deleted the debug lines (python complained about converting some characters into ascii), it started working perfectly. Someone should leave a info about this password thing :/

ilyam8:
@elfhack 	
could you try to install #5593 and check if it works?

ilyam8:
> Someone should leave some warning about this password thing :/

i am not sure, it may be specific to `itzg/minecraft-server`, haven't tested other. Anyway i added some log messages about it

> self.error("'{0}' command returned no value, make sure you set correct password".format(COMMAND_TPS))

elfhack:
@ilyam8 

Nope, can't use this version

```
2019-03-10 22:06:26: python.d ERROR: spigotmc: spigotmc: check() unhandled exception: 'ascii' codec can't encode character u'\xa7' in position 0: ordinal not in range(128)
2019-03-10 22:06:26: python.d ERROR: spigotmc: spigotmc: Traceback (most recent call last):
  File "/usr/libexec/netdata/plugins.d/python.d.plugin", line 319, in check_job
    check_ok = bool(job.check())
  File "/usr/libexec/netdata/python.d/spigotmc.chart.py", line 82, in check
    return self._get_data()
  File "/usr/libexec/netdata/python.d/spigotmc.chart.py", line 119, in _get_data
    self.debug("'{0}' command output : {1}".format(COMMAND_TPS, raw))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa7' in position 0: ordinal not in range(128)
```

ilyam8:
fixed in #5598

elfhack:
Yep, this one works from the get go

PR #5509, opened 2019-02-26
Implementation of repeating notifications feature

kamcpp:
##### Summary
We keep a separate list of repeating alarm entries next to the existing alarm entries. We keep only one alarm entry for each alarm and we keep it updated. We create the alarm entry lazily and we keep it until the host object is freed. We still use the same logging mechanism for the repeating alarm entries therefore we keep track of all changes made to them like non-repeating alarm entries.

##### Component Name
Health component

##### Additional Information
This PR implements the feature requested in #2956.


kamcpp:
This is a WIP PR therefore further iterations might be added without further notice. The implementation works in simple cases but it might need further modifications for corner cases. Your comments and notes are most welcome.

health/health.c
```diff
@@ -125,7 +128,7 @@ static inline void health_alarm_execute(RRDHOST *host, ALARM_ENTRY *ae) {
     // find the previous notification for the same alarm
     // which we have run the exec script
     // exception: alarms with HEALTH_ENTRY_FLAG_NO_CLEAR_NOTIFICATION set
-    if(likely(!(ae->flags & HEALTH_ENTRY_FLAG_NO_CLEAR_NOTIFICATION))) {
+    if(likely(!is_repeating && !(ae->flags & HEALTH_ENTRY_FLAG_NO_CLEAR_NOTIFICATION))) {
```
kamcpp: We need this only if we want to notify about transitions from/to CLEAR status.

health/health.c
```diff
@@ -264,14 +268,56 @@ static inline void health_alarm_log_process(RRDHOST *host) {
...
+                else if(likely(rc->status == RRDCALC_STATUS_CLEAR)) {
+                    if(ae->alarm_id == rc->id)
+                        break;
+                }
+            }
+            if(unlikely(/*(ae->old_status != RRDCALC_STATUS_CLEAR &&
+                         ae->new_status == RRDCALC_STATUS_CLEAR) ||
+                        (ae->old_status == RRDCALC_STATUS_CLEAR &&
+                         (ae->new_status == RRDCALC_STATUS_WARNING ||
+                          ae->new_status == RRDCALC_STATUS_CRITICAL)) ||*/
```
kamcpp: Do we need these transitions from/to CLEAR status?

health/health_log.c
```diff
@@ -358,54 +361,100 @@ inline void health_alarm_log(
...
     debug(D_HEALTH, "Health adding alarm log entry with id: %u", host->health_log.next_log_id);
 
-    ALARM_ENTRY *ae = callocz(1, sizeof(ALARM_ENTRY));
-    ae->name = strdupz(name);
-    ae->hash_name = simple_hash(ae->name);
-
-    if(chart) {
-        ae->chart = strdupz(chart);
-        ae->hash_chart = simple_hash(ae->chart);
+    bool_t create_new_alarm_entry = 1;
```
kamcpp: Use TRUE instead.

health/health_log.c
```diff
@@ -358,54 +361,100 @@ inline void health_alarm_log(
...
+    bool_t is_repeating = make_bool(warn_repeat_every > 0 || crit_repeat_every > 0);
+    REPEATING_ALARM_ENTRY *target_alarm_entity = NULL;
+    if(unlikely(is_repeating)) {
+        REPEATING_ALARM_ENTRY *it;
+        for(it = host->health_rep_alarm_entry_list; it && it->next; it = it->next)
+            if(it->alarm_entry->alarm_id == alarm_id) {
+                target_alarm_entity = it;
+                create_new_alarm_entry = 0;
```
kamcpp: Use FALSE instead.

health/health.c
```diff
@@ -264,14 +268,56 @@ static inline void health_alarm_log_process(RRDHOST *host) {
...
                 health_process_notifications(host, ae);
+            }
         }
     }
 
     // remember this for the next iteration
     host->health_last_processed_id = first_waiting;
 
+    // process repeating alarm entries
+    for(REPEATING_ALARM_ENTRY *it = host->health_rep_alarm_entry_list; it; it = it->next) {
```
kamcpp: C99: For initialization

health/health.c
```diff
@@ -264,14 +268,56 @@ static inline void health_alarm_log_process(RRDHOST *host) {
...
+        if(unlikely(ae->last_repeat == 0)) {
+            if(likely(now >= ae->delay_up_to_timestamp)) {
+                health_process_notifications(host, ae);
+                ae->last_repeat = now;
+                debug(D_HEALTH, "Alarm entry from alarm %u is executed as repeating for the first time.", ae->alarm_id);
+            }
+        }
+        else {
+            int repeat_every = 0;
+            for(RRDCALC *rc = host->alarms; rc ; rc = rc->next) {
```
kamcpp: C99: For initialization

cakrit:
I'll start testing and deeper code review in a bit. One thing I noticed is that it seems we can't override the global 'repeat every' settings for a specific alarm or template. That's something that will also be quite useful.

health/health_log.c
```diff
@@ -358,54 +361,100 @@ inline void health_alarm_log(
         const char *units,
         const char *info,
         int delay,
-        uint32_t flags
+        uint32_t flags,
+        uint32_t warn_repeat_every,
+        uint32_t crit_repeat_every
 ) {
     debug(D_HEALTH, "Health adding alarm log entry with id: %u", host->health_log.next_log_id);
```
cakrit: This debug needs to be moved now, since it depends on whether create_new_alarm_entry will be true or not.

health/health_log.c
```diff
@@ -358,54 +361,100 @@ inline void health_alarm_log(
...
+        if(unlikely(!target_alarm_entity)) {
+            // TODO: Remove these entries once host is going to be freed
+            REPEATING_ALARM_ENTRY *rae = callocz(1, sizeof(REPEATING_ALARM_ENTRY));
+            rae->alarm_entry = ae;
+            if(unlikely(!host->health_rep_alarm_entry_list)) {
+                host->health_rep_alarm_entry_list = rae;
+            }
+            else {
+                rae->next = host->health_rep_alarm_entry_list;
+                host->health_rep_alarm_entry_list = rae;
```
cakrit: Why not just keep these two lines? The If isn't necessary. Exactly the same thing as lines 440-441 should be enough, rae->next will get a null if we fall into the case of lines 449-450 anyway.

database/rrdhost.c
```diff
@@ -534,6 +537,13 @@ void rrdhost_free(RRDHOST *host) {
     while(host->alarms)
         rrdcalc_unlink_and_free(host, host->alarms);
 
+    while(host->health_rep_alarm_entry_list) {
```
cakrit: I would put this code inside health_alarm_log_free. It seems to me a bit cleaner that way...

health/health.h
```diff
@@ -127,7 +127,9 @@ extern void health_alarm_log(
         const char *units,
         const char *info,
         int delay,
-        uint32_t flags);
+        uint32_t flags,
+        uint32_t warn_repear_every,
```
cakrit: typo, should be `warn_repeat_every`

cakrit:
> I'll start testing and deeper code review in a bit. One thing I noticed is that it seems we can't override the global 'repeat every' settings for a specific alarm or template. That's something that will also be quite useful.

I found it in the code, it's the `repeat: warning X critical Y`.  
I'm pretty sure that duration accepts X in the form '1m 30s', which won't be possible like this.
I think it will simplify things greatly, if you do the exact same thing as in  netdata.conf and allow for two different params, repeat warning = X and repeat critical = Y, or use an additional delimiter other than space.

health/health.c
```diff
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
+#include <assert.h>
```
cakrit: We don't really use assert anywhere (the one in the freeipmi plugin is an exception). Can't you handle the cases with unlikely and a message + correction if possible?

kamcpp:
> I'll start testing and deeper code review in a bit. One thing I noticed is that it seems we can't override the global 'repeat every' settings for a specific alarm or template. That's something that will also be quite useful.

I'll do that. I will also the config in the definition of the alarm or template.

health/health_log.c
kamcpp: OK, I'll do that.

health/health_log.c
kamcpp: OK, I'll check that.

database/rrdhost.c
kamcpp: OK, I'll do that.

health/health.h
kamcpp: Thanks, I'll correct that.

kamcpp:
> > I'll start testing and deeper code review in a bit. One thing I noticed is that it seems we can't override the global 'repeat every' settings for a specific alarm or template. That's something that will also be quite useful.
> 
> I found it in the code, it's the `repeat: warning X critical Y`.
> I'm pretty sure that duration accepts X in the form '1m 30s', which won't be possible like this.
> I think it will simplify things greatly, if you do the exact same thing as in netdata.conf and allow for two different params, repeat warning = X and repeat critical = Y, or use an additional delimiter other than space.

You mean it would be better to have the config like that, am I right? If yes, I'll do that.

health/health.c
kamcpp: OK, I'll do that. Assertions can be handful since it stops the engine gracefully when a never-happening condition is met. By the way, I'll remove them.

kamcpp:
> In addition to the comments I added, it looks like there's one basic problem. This code doesn't actually save to or read from the log on disk ae->last_repeat. Together with another apparent bug that adds alarms to health_rep_alarm_entry_list that shouldn't be there because they are now clear, this has the effect of trying to trigger an alarm notification for every single configured alarm, regardless of the status, upon restart (called from health.c L286). The only reason we don't actually get a notification is the checks inside health_process_notifications.
> 
> If you look a bit more carefully to how entries are added to and removed from health_rep_alarm_entry_list and you actually store the last_repeat to disk, I think we will be very close to merging. We can go through the details over a call as well.

My solution is to always keep an alarm entry for a repeating alarm in a dedicated separate list no matter what the actual status of the alarm is. That's because of the fact that the status of the alarm can change at any time. We don't continue repeating the notifications when the alarm is cleared but as soon as the alarm goes warning or critical, we use the same alarm entry and continue repeating the notifications.

About saving the alarm entries in the log; we store the occurrence of notifications as a log but there is no need to load them back as repeating entries. Eventually, we would have alarm entries for repeating alarms when Netdata daemon is back online and we continue to send the notifications when the alarm goes warning or critical. In this sense, there is no need to store last_repeat field.

cakrit:
> Do you mean that it would be better to have the config like what you mentioned above, am I right? If yes, I'll do that.

Yes

> My solution is to always keep an alarm entry for a repeating alarm in a dedicated separate list no matter what the actual status of the alarm is. That's because of the fact that the status of the alarm can change at any time. We don't continue repeating the notifications when the alarm is cleared but as soon as the alarm goes warning or critical, we use the same alarm entry and continue repeating the notifications.
> 
> About saving the alarm entries in the log; we store the occurrence of notifications as a log but there is no need to load them back as repeating entries. Eventually, we would have alarm entries for repeating alarms when Netdata daemon is back online and we continue to send the notifications when the alarm goes warning or critical. In this sense, there is no need to store last_repeat field.

I will need to explain in more detail where the problems are, in a call. To reiterate the main issues: 
- The current code attempts to send notifications for cleared alarms.
- The repeat period will not be respected between restarts (e.g. I request repeat of 1 hour. Alarm 1 is sent and I restart immediately. I will then get a second alarm, even though the hour hasn't passed). Not a good user experience. 
- Since all alarm entries with a repeat are copied, we iterate through them every time, even if no alarm is raised. This is inefficient and doesn't offer a benefit. 

Not much needs to be changed, we'll go through it together.

health/health.c
kamcpp: I removed the asserts. There a bunch of TODOs which I address in this PR.

health/health.h
kamcpp: Done.

database/rrdhost.c
kamcpp: These codes are not longer part of the change.

health/health_log.c
kamcpp: These codes are not longer part of the change.

health/health_log.c
kamcpp: These codes are not longer part of the change.

health/health.c
kamcpp: These codes are not longer part of the change.

health/health.c
kamcpp: These codes are not longer part of the change.

health/health_log.c
```diff
@@ -110,6 +111,8 @@ inline void health_alarm_log_save(RRDHOST *host, ALARM_ENTRY *ae) {
                             , ae->old_status
                             , ae->delay
 
+                            , (uint64_t)ae->last_repeat
```
cakrit: put it at the end.

cakrit:
I verified that saving and loading the last time we sent a notification actually has no effect.
The alarm list hasn't even been created at that point (see rrdhost.c line 242 and beyond).
I reversed the calls, to make it like this:
```
    if(host->health_enabled) {
        rrdhost_wrlock(host);
        health_readdir(host, health_user_config_dir(), health_stock_config_dir(), NULL);
        rrdhost_unlock(host);

		health_alarm_log_load(host);
		health_alarm_log_open(host);
    }
```

But even then, the following never evaluates as true: `if (rc->id == alarm_id && rrdcalc_isrepeating(rc))`

So it looks like we need the change in rrdhost.c and that the check here needs fixing and testing.

health/health_log.c
```diff
@@ -174,6 +177,19 @@ inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char *filena
                 continue;
             }
 
+            time_t last_repeat = (time_t)strtoul(pointers[25], NULL, 16);
```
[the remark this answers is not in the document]
kamcpp: Ah, you mean the older health-log.db! true. I move it to the end.

health/health_log.c
kamcpp: Sure, I'll do.

kamcpp:
> I verified that saving and loading the last time we sent a notification actually has no effect.
> The alarm list hasn't even been created at that point (see rrdhost.c line 242 and beyond).
> I reversed the calls, to make it like this:
> 
> ```
>     if(host->health_enabled) {
>         rrdhost_wrlock(host);
>         health_readdir(host, health_user_config_dir(), health_stock_config_dir(), NULL);
>         rrdhost_unlock(host);
> 
> 		health_alarm_log_load(host);
> 		health_alarm_log_open(host);
>     }
So, that's why I don't see the load of entries. I tried multiple times but I couldn't load the entires. I will focus on this.

> But even then, the following never evaluates as true: `if (rc->id == alarm_id && rrdcalc_isrepeating(rc))`
Why never? If we have an alarm entry for a repeating alarm (from the previous run), the condition becomes true and the code updates the 'last_repeat' field of the found alarm. By the way, I'll check this again.

kamcpp:
> The alarm list hasn't even been created at that point (see rrdhost.c line 242 and beyond).

I got your point now. OK, I'll figure it out.

health/health_log.c
kamcpp: Done.

health/health_log.c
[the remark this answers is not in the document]
kamcpp: Done.

kamcpp:
I tested the "last_repeat" field functionality and it works.

cakrit:
As discussed after a meeting with @ktsaou, we'll need  additional changes before we can get the code  to a point where we can we can merge it. @kamcpp are you ok with me closing this PR and continuing the work myself on a different one? We'll reference this one and you by name of course!

kamcpp:
Yes @cakrit, I’m totally fine.

pschiffe:
Hey guys, what's the current status? Was there a new PR created? I think I'm not the only one who cannot wait until this feature is ready :-)

thiagoftsm:
Hi @pschiffe you are not the unique that cannot wait, stay sure about this! :) 
I would like to say that I got the code this week to finish the necessary adjusts and I created a new PR with the changes after to review the nice work done by @kamcpp. @kamcpp as you can see in the new PR that I created for this request(#6309 ), I am using the majority of your code, but I decided to do the following changes:

- I changed the function alarm_ispeating, now instead to do a loop trough all the alarms to verify whether the alarm is yet there, I am using binary tree to speed up the search.
- I saw that you used definitions like bool_t, FALSE and TRUE to create a new type, but these definitions in my vision are not necessaries, I am saying this because since c99 we have the stdbool.h that have the definitions  bool, false and true. You will see that I converted everything directly to integer, but your whole algorithm was kept.

Best Regards!

PR #5511, opened 2019-02-26
Prevent data corruption upon GUID duplication between master and slave netdata instances

paulkatsoulakis:
##### Summary
patch to mitigate data corruption during replication of metrics, when duplication of machine GUID occurs between master/slave instances of netdata.
Reference issue #5488

##### Component Name
netdata/netdata/streaming

##### Additional Information
None

### DIFF
--- a/streaming/rrdpush.c
+++ b/streaming/rrdpush.c
@@ -868,9 +868,7 @@ static int rrdpush_receive(int fd
     tags = appconfig_set_default(&stream_config, machine_guid, "host tags", (tags)?tags:"");
     if(tags && !*tags) tags = NULL;
 
-    if(!strcmp(machine_guid, "localhost"))
-        host = localhost;
-    else if (strcmp(machine_guid, localhost->machine_guid) == 0) {
+    if (strcmp(machine_guid, localhost->machine_guid) == 0) {
         log_stream_connection(client_ip, client_port, key, machine_guid, hostname, "DENIED - ATTEMPT TO RECEIVE METRICS FROM MACHINE_GUID IDENTICAL TO MASTER");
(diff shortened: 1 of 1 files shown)


cakrit:
In @ktsaou's PR I also see a `close(fd)` on that branch of the code. Isn't that needed?

paulkatsoulakis:
> In @ktsaou's PR I also see a `close(fd)` on that branch of the code. Isn't that needed?
Yikes, you 're right. The fd is passed through and if not used need to be released. I 'll update the PR

paulkatsoulakis:
Not quite sure what happened with LGTM after the second commit.
I can't access its logs for the details

cakrit:
Retrying analysis, it seems to have stopped without a message.

streaming/rrdpush.c
```diff
@@ -870,6 +870,12 @@ static int rrdpush_receive(int fd
 
     if(!strcmp(machine_guid, "localhost"))
         host = localhost;
```
ktsaou: This is also a case I don't recall.
If `host == localhost`, then should we accept the metrics?

I think I had implemented this case to support the possibility of remote data collection for parts of a system. But I am not sure this is used anywhere...

cakrit: I don't see any way for `machine_guid` to get the value `localhost`, other than a user explicitly putting `localhost` in `/var/lib/netdata/registry/netdata.public.unique.id`. Agree that it looks like a useless piece of code.

paulkatsoulakis: I was wondering myself what is that check over there. I assumed is some kind of "hack" to replay some kind of a use case when machine_guid string is manually set to "localhost".

If that is not the case, then it is very likely we can just kill it

cakrit: Yes, remove it please, I approve and you merge.

PR #5519, opened 2019-02-27
Show streamed servers even for users that are not signed in! #5482

### LINKED ISSUE #5482: "Other servers" menu doesn't show remote streaming servers while not logged into nedata.cloud
Remote host is accessible as normal on `/host/alarm/`, and shows the dashboard as normal. The charts update fine. However there's no link in the server switcher menu to switch between hosts, because the whole menu is replaced by a "Please sign in to netdata.cloud" message. Upon signing in, the host appears in the menu as normal.

```
$ netdata -V
netdata v1.12.1-7-g86ed3869
```

gmosx:
Fixes #5482 
The streamed servers don't depend on the new cloud registry so we can always show the entries in the my-netdata menu.

### DIFF
--- a/web/gui/main.js
+++ b/web/gui/main.js
@@ -550,6 +550,17 @@ function renderStreamedHosts(options) {
 }
 
 function renderMachines(machinesArray) {
+    if (!isSignedIn()) {
+        if (!NETDATA.registry.isRegistryEnabled()) {
+            return (
+                `<div class="info-item" style="white-space: nowrap">
+                    <span>Please <a href="#" onclick="signInDidClick(event); return false">sign in to netdata.cloud</a> to view your nodes!</span>
+                    <div></div>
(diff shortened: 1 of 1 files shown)


ktsaou:
Could you please post a screenshot?

ktsaou:
Ideally, "sign in" should be at the top of the menu, to make sure it is visible even when there are dozens of servers streamed to a netdata master.

gmosx:
> Ideally, "sign in" should be at the top of the menu, to make sure it is visible even when there are dozens of servers streamed to a netdata master.

It will be confusing, and also will complicate the code. Unless you would like to always show the netdata nodes before the streamed databases.

gmosx:
OK, changed like this:

image

RX14:
Thanks!

PR #5536, opened 2019-03-02
Minor fixes around plugin_directories initialization

paulkatsoulakis:
##### Summary
This is a small PR attempting to simplify and clarify the plugin_directories initialization process within daemon. No functional changes should be included within this change. Feel free to recommend further adjustments or even reject the PR if it does not comply with any policy.

The changes include:

1) Removal of a magic number, primarily because magic numbers are evil and we need to know why we treat the first entry differently (using 0 is counter intuitive)

2) pull the initialization logic of the plugin_directories variable within the pluginsd scope, only expose a method to trigger it, nothing else needed as it seems (removed a couple of useless globals too)

3) renamed **netdata_configured_plugins_dir** to **netdata_configured_primary_plugins_dir**, because we have multiple directories for the plugins and that parameter actually contains our primary (or stock? not sure its clear to me yet what is the conceptual classification of that first dir)

##### Component Name
netdata/daemon
netdata/collectors/plugins.d

##### Additional Information

### DIFF
--- a/collectors/cgroups.plugin/sys_fs_cgroup.c
+++ b/collectors/cgroups.plugin/sys_fs_cgroup.c
@@ -210,10 +210,10 @@ void read_cgroup_plugin_configuration() {
                     " * "
             ), NULL, SIMPLE_PATTERN_EXACT);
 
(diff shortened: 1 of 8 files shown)


paulkatsoulakis:
Thanks bot :P

PR #5537, opened 2019-03-02
Registry delete should allow deleting the host we are looking at

### LINKED ISSUE #5518: Unexpected host seen in list of hosts
On Wed, Feb 27, 2019 at 11:12 AM Clarence Lin <mr.lin.clarence@gmail.com> wrote:
Hi,

In my google oauth login I see http://ruhighload.com:19999/, but I can't delete it.
Why?
This not my site.

unnamed

##### Component Name
web

cakrit:
##### Summary
Fixes #5518 

##### Component Name
registry

##### Additional Information

Remove check on whether a server delete request comes from the same server we are looking at

### DIFF
--- a/registry/registry_internals.c
+++ b/registry/registry_internals.c
@@ -126,6 +126,11 @@ REGISTRY_PERSON_URL *registry_verify_request(char *person_guid, char *machine_gu
         info("Registry Request Verification: URL not found for person, person: '%s', machine '%s', url '%s'", person_guid, machine_guid, url);
         return NULL;
     }
+    //else if (pu->machine != m) {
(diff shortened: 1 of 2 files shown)


registry/registry_internals.c
```diff
@@ -126,6 +126,11 @@ REGISTRY_PERSON_URL *registry_verify_request(char *person_guid, char *machine_gu
         info("Registry Request Verification: URL not found for person, person: '%s', machine '%s', url '%s'", person_guid, machine_guid, url);
         return NULL;
     }
+    //else if (pu->machine != m) {
```
gmosx: What is this commented block?

cakrit: We were considering adding it with @ktsaou, but decided to leave it commented out for now. The registry code will need a major revisit at some point, it's just that now is not the time.

PR #5540, opened 2019-03-03
Don't send zeroes for empty groups

### LINKED ISSUE #5514: apps_plugin plots zeroes for empty groups
##### Bug report summary

`apps_plugin` plots zeroes for group even if there is no processes in that group.

related issue #5510 	

##### OS / Environment

any

netdata v1.12.1-16-g2ee2d948

##### Component Name

[collectors/apps.plugin](https://github.com/netdata/netdata/tree/master/collectors/apps.plugin)

##### Steps To Reproduce
 - add an entry to `apps_groups.conf`
 - run a process that maches that entry
 - stop the process
 - see charts

##### Expected behavior

`apps_plugin` shouldn't return any values for group if there is no processes in that group

### LINKED ISSUE #5547: Alert if process isn't alive
##### Bug report summary
I'm setting up alerts for important processes that I need to know if they're alive, I've added them to the apps_groups.conf file like this:
```
Go.Main:                     *nsMain.js*
Go.Voice:                    *nsVoice.js*
Go.Chat:                     *nsChat.js*
Go.TicketsExternalService:   *nsTicketsExternalService.js*
Go.TicketsMailService:       *nsTicketsMailService.js*
Go.TicketsPostgresListen:    *nsTicketsPostgresListen.js*
Go.TicketsSockService:       *nsTicketsSockService.js*
....
```
I made my alerts based on apache example in your docs ([apache_last_collected_secs](https://docs.netdata.cloud/health/#example-1)), like:
```
# ****** Go.Main ******
template: nsMain_last_collected_secs
      on: apps.cpu
  family: cpu
    calc: $now - $apps.cpu.Go.Main_last_collected_t
   every: 5s
   units: seconds ago
    warn: $this > (($status >= $WARNING)  ? ($update_every) : ( 5 * $update_every))
    crit: $this > (($status == $CRITICAL) ? ($update_every) : (60 * $update_every))
   delay: down 10m multiplier 1.2 max 1h
      to: sysadmin
```
I can see all the variables in the API (using Postman) and view the variable apps.cpu.Go.Main_last_collected_t, but when I stop the respective process, the variable continues with the same value (in seconds), even if I refresh many times, these variables do not change.

Is it possible that _apps.plugin_ doens't refresh all processes at same time? Or my config isn's right?
I have tried with more common process (like httpd and Postgres) and this worked as expected.

##### OS / Environment
`Centos 7 x64`
`netdata v1.12.2-4-nightly`

vlvkobal:
##### Summary
Don't send zeroes when all processes from a group have been exited.

Fixes #5514 
Fixes #5547

##### Component Name
`apps.plugin`

### DIFF
--- a/collectors/apps.plugin/apps_plugin.c
+++ b/collectors/apps.plugin/apps_plugin.c
@@ -3176,21 +3176,21 @@ static void send_collected_data_to_netdata(struct target *root, const char *type
 
     send_BEGIN(type, "cpu", dt);
     for (w = root; w ; w = w->next) {
-        if(unlikely(w->exposed))
+        if(unlikely(w->exposed && w->processes))
             send_SET(w->name, (kernel_uint_t)(w->utime * utime_fix_ratio) + (kernel_uint_t)(w->stime * stime_fix_ratio) + (kernel_uint_t)(w->gtime * gtime_fix_ratio) + (include_exited_childs?((kernel_uint_t)(w->cutime * cutime_fix_ratio) + (kernel_uint_t)(w->cstime * cstime_fix_ratio) + (kernel_uint_t)(w->cgtime * cgtime_fix_ratio)):0ULL));
     }
     send_END();
 
     send_BEGIN(type, "cpu_user", dt);
     for (w = root; w ; w = w->next) {
-        if(unlikely(w->exposed))
+        if(unlikely(w->exposed && w->processes))
             send_SET(w->name, (kernel_uint_t)(w->utime * utime_fix_ratio) + (include_exited_childs?((kernel_uint_t)(w->cutime * cutime_fix_ratio)):0ULL));
     }
     send_END();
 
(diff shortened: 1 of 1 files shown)


ilyam8:
i don't think we need obsolete, not sending zeroes is better.

vlvkobal:
I discussed the PR with @ilyam8. He thinks that obsoleting dimensions is the last resort we should use in case we have tons of dimensions. In `apps.plugin` there is a finite number of groups, so we can afford to keep data in the database. @cakrit, @ktsaou, what do you think?

ilyam8:
I see no reason to obsolete dimensions (mark as obsolete => delete from database), it makes no sense to me

cakrit:
Since it's a finite, well known set, I think @ilyam8 is correct.

vlvkobal:
I removed obsoleting dimensions. Thus, only sending zeroes is suppressed.

Issue #5470, opened 2019-02-22
[FIX] RocketChat notifications not working


tctovsli:
##### Bug report summary
After configuring RocketChat notification as the documentation describes, notifications doesn't come to RocketChat.

##### Solution
The culprit of the problem, was in /usr/libexec/netdata/plugins.d/alarm-notify.sh, where rocketchat wasn't defined in the list method_names (line 146) although the script has a method for it. After adding rocketchat to this list, notifications started working.
```# -----------------------------------------------------------------------------
# List of all the notification mechanisms we support.
# Used in a couple of places to write more compact code.

method_names="
email
pushover
pushbullet
telegram
slack
alerta
flock
discord
hipchat
twilio
messagebird
pd
fleep
syslog
custom
msteam
kavenegar
prowl
awssns
"
```

On a side-note, there was also no definitions of rocketchat-recipients in health_alarm_notify.conf like all the other channels. I defined these before I found the issue with method_names.
Example of what was *not* defined in the bottom of the file: `role_recipients_rocketchat[sitemgr]="${DEFAULT_RECIPIENT_ROCKETCHAT}"`

I could make a PR, but I'm new to this project and I found this the fastest way to report the 2 issues I found.

cakrit:
> I could make a PR, but I'm new to this project and I found this the fastest way to report the 2 issues I found.

No better way to learn! Do the PR, we welcome all contributions and you'll feel more comfortable for your second one. :)

cakrit:
Congrats on your first PR. 
This was closed automatically when the PR was merged. You could add the role_recipients declarations as well if you like, in a different PR.

PR #5545, opened 2019-03-04
Added RocketChat default recipient for roles

tctovsli:
##### Summary
Adds RocketChat default recipient for all roles, just like the other notification channels.
This fixes one of the issues described in #5470
##### Component Name
health


tctovsli:
I have updated the file now. I checked with the config I have running, and it's similar.

Issue #5525, opened 2019-02-28
descrease python.d.plugin memory usage


ilyam8:
##### Feature idea summary

`python.d.plugin` imports a lot of packages (yaml, urllib3 + additional packages if they are present on the system - pymongo, pymysql, etc.) during auto-detection process and they live in memory even if they aren't needed later (module is not running).

For example, `python.d.plugin` consumes 21.1 MiB on my system (Manjaro) with only `example` module running.

___

Since it is almost impossible to remove imported packages from memory, the idea is to have a separate process that will:
 - load configs (yaml package)
 - load module and perform job check (job specific packages)
 - report to main process which modules should be loaded (with at least on successful job)

##### Expected behavior

`python.d.plugin` use less memory

PR #5552, opened 2019-03-05
python.d.plugin: use separate process for initial module checking

ilyam8:
##### Summary

This PR adds (major) changes only to `python.d.plugin` file.

Fixes: #5525

`pyhton.d.plugin` imports a lot of additional packages during initial module initialization/job creating/checking and there is no way to unimport them, even if they arn't needed. It consumes relatively a lot of ram.

___
Memory utilization comparing before/after the PR (one job `example` module, py3.7.2):

> 21.1 => 8.8 MiB

screenshot_20190305_111837

##### Component Name
[`collectors/python.d.plugin`](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/python.d.plugin.in)

##### Additional Information

This PR adds separate process for initial module checking. 

Logic:
 - main process spawns checker process
 - checker process loads every module, loads module config, creates jobs and runs job.check() for every job, if check success it adds the job to the list.
 - checker process returns list of modules and jobs.
 - main process loads only active modules, etc.

### DIFF
--- a/collectors/python.d.plugin/powerdns/powerdns.chart.py
+++ b/collectors/python.d.plugin/powerdns/powerdns.chart.py
@@ -125,6 +125,7 @@ class Service(UrlService):
         UrlService.__init__(self, configuration=configuration, name=name)
         self.order = ORDER
         self.definitions = CHARTS
+        self.url = configuration.get('url', 'http://127.0.0.1:8081/api/v1/servers/localhost/statistics')
 
     def check(self):
         self._manager = self._build_manager()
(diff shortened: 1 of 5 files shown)


ilyam8:
ok, i tested it with on Manjaro (`py 3.7.2`) and on Centos6(`py 2.6.6`), and it appears to be working correctly

ilyam8:
Hey @Ferroin if you could find a time and test the PR a bit it would be very nice

ilyam8:
ok, i don't think we need to run a job if `job.check()` returns False in main. It can be error prone, instead we can treat all jobs in main as autodetection jobs, so if a job's check(), for some reason, succed in child process and failed in main we will retry check again and again

ilyam8:
@ktsaou if you are ok with the changes this is ready.

ilyam8:
@cakrit please approve

ilyam8:
ok, here we go, i am on a high alert :alarm_clock:

PR #5557, opened 2019-03-05
Delay renaming for docker containers

### LINKED ISSUE #5512: Docker containers resources missing unless a netdata full restart executed
Hello,

Not sure if i am missing something , but i find this frustrating when it comes to displaying docker container resources such as cpu,mem,disk and eth0, they won't show up in the GUI unless i restart netdata, we are talking about ephemeral containers

The most frustrating is that sometimes netdata will show the container ID and not its name until i restart netdata. 

Would you suggest running the script below in a cron, so that whenever i deploy new containers, netdata would pick up their names without requiring a full restart ?

/usr/libexec/netdata/plugins.d/cgroup-name.sh

Still not sure why the others container resources(disk,eth0...) do not show immediately for newly(ephemeral) created containers

BTW i am running netdata locally on the host.

##### OS / Environment
Centos 7.6
netdata v1.12.0-rc3-54-g43628db

##### Component Name
cgroup/container

vlvkobal:
##### Summary
When a docker container is starting there is a possibility that the `docker ps` command will not return the name of the container. An additional renaming attempt after a delay was introduced to deal with the docker lag.

Fixes #5512

##### Component Name
`cgroups.plugin`

### DIFF
--- a/collectors/cgroups.plugin/cgroup-name.sh.in
+++ b/collectors/cgroups.plugin/cgroup-name.sh.in
@@ -175,4 +175,10 @@ if [ -z "${NAME}" ]; then
 fi
 
 info "cgroup '${CGROUP}' is called '${NAME}'"
-echo "${NAME_NOT_FOUND}" "${NAME}"
+echo "${NAME}"
+
+if [ "${NAME_NOT_FOUND}" -eq 1 ]; then
+	exit 2
+else
+	exit 0
+fi
--- a/collectors/cgroups.plugin/cgroup-name.sh.in
+++ b/collectors/cgroups.plugin/cgroup-name.sh.in
@@ -73,6 +73,7 @@ function docker_get_name() {
 	fi
 	if [ -z "${NAME}" ]; then
 		warning "cannot find the name of docker container '${id}'"
+		NAME_NOT_FOUND=1
 		NAME="${id:0:12}"
 	else
 		info "docker container '${id}' is named '${NAME}'"
@@ -95,6 +96,7 @@ function docker_validate_id() {
 
 DOCKER_HOST="${DOCKER_HOST:=/var/run/docker.sock}"
 CGROUP="${1}"
+NAME_NOT_FOUND=0
 NAME=
 
 # -----------------------------------------------------------------------------
@@ -173,4 +175,4 @@ if [ -z "${NAME}" ]; then
 fi
 
 info "cgroup '${CGROUP}' is called '${NAME}'"
-echo "${NAME}"
+echo "${NAME_NOT_FOUND}" "${NAME}"
--- a/collectors/cgroups.plugin/sys_fs_cgroup.c
+++ b/collectors/cgroups.plugin/sys_fs_cgroup.c
@@ -376,6 +376,9 @@ struct cgroup {
     char available;      // found in the filesystem
     char enabled;        // enabled in the config
 
+    char renaming_delayed;
+    char renamed;
+
     char *id;
     uint32_t hash;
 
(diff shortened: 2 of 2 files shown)


ilyam8:
The delay fix seems to be working, i've tested it. I created 100 containers, all of them were renamed correctly

ktsaou:
After how much time is checked again?

vlvkobal:
After
```
[plugin:cgroups]
    # check for new cgroups every = 10
```

ktsaou:
ok, I reviewed the code.

Here are some comments:

1. The script `cgroup-name` now returns 2 values. The first is a Boolean. I think the return value of the script is more appropriate for this.

2. There are 2 new cgroup variables introduced: `needs_renaming` and `renaming_delayed`.

  - `renaming_delayed` is set to 1, but never to 0
  - I find the whole idea with these variables a bit complicated. I simpler approach would be to set `pending_renames` to `2` when the cgroup is to be renamed. Then within `cgroup_get_chart_name()` first check if `pending_renames > 0`, and decrement it. If the script returns `0`, set `pending_renames = 0`. This way, `cgroup_get_chart_name()` should be called only when `pending_renames > 0`.

What do you think?

vlvkobal:
Now script uses an exit code instead of the second parameter.

I replaced two variables by one variable `pending_renames`. For my taste, two variables reflected the real meaning of the code a little bit more accurately.

cakrit:
Merging myself because I need to work on the same code.

Thatdarkone:
Does this pull resolve the issue of missing interfaces and disks ?
[https://github.com/netdata/netdata/issues/5512#issuecomment-467952475](url)

ilyam8:
@Thatdarkone disks are added only if there are any non zero data (so, it will be added later)

Example
 - container with disk

```
→ cat  /sys/fs/cgroup/blkio/docker/af442bc2a5f272ac045275e99244f8f178d59e36f970d2da1a979b171c4eb532/*
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
500
cat: /sys/fs/cgroup/blkio/docker/af442bc2a5f272ac045275e99244f8f178d59e36f970d2da1a979b171c4eb532/blkio.reset_stats: Permission denied
8:0 Read 7643136
8:0 Write 0
8:0 Sync 0
8:0 Async 7643136
8:0 Total 7643136
253:0 Read 7643136
253:0 Write 0
253:0 Sync 0
253:0 Async 7643136
253:0 Total 7643136
Total 15286272
8:0 Read 65
8:0 Write 0
8:0 Sync 0
8:0 Async 65
8:0 Total 65
253:0 Read 65
253:0 Write 0
253:0 Sync 0
253:0 Async 65
253:0 Total 65
Total 130
500
0
cat: /sys/fs/cgroup/blkio/docker/af442bc2a5f272ac045275e99244f8f178d59e36f970d2da1a979b171c4eb532/cgroup.event_control: Permission denied
4499
4571
0
4499
4571
```

 - container w/o disk

```
→ cat  /sys/fs/cgroup/blkio/docker/1535b27bc0604ed8a4f0712793371b1d86797ee3aa555eadeb176c9c6f60305e/*
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
Total 0
500
cat: /sys/fs/cgroup/blkio/docker/1535b27bc0604ed8a4f0712793371b1d86797ee3aa555eadeb176c9c6f60305e/blkio.reset_stats: Permission denied
Total 0
Total 0
500
0
cat: /sys/fs/cgroup/blkio/docker/1535b27bc0604ed8a4f0712793371b1d86797ee3aa555eadeb176c9c6f60305e/cgroup.event_control: Permission denied
6125
6159
0
6125
6159
```

PR #5569, opened 2019-03-07
Make sign-in functionality more visible

gmosx:
An info banner is introduced.

### DIFF
--- a/web/gui/index.html
+++ b/web/gui/index.html
@@ -20,7 +20,7 @@
     <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
     <meta name="author" content="costa@tsaousis.gr">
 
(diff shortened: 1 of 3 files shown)


ktsaou:
I installed this. I can see the "Sign In" button having dark background, but I can't see the banner. Are there any conditions for it to appear?

gmosx:
> I installed this. I can see the "Sign In" button having dark background, but I can't see the banner. Are there any conditions for it to appear?

If you sso once it will not show it again.

Make sure you clear the `signInBannerClosed` variable in localStorage.

gmosx:
> I see that a user closes it once, it will never appear again. If that was the intention, it works.

That's the intension. Obviously, we don't want to annoy the user, only inform him.

gmosx:
@ktsaou, fix for 'white' theme:

image

Issue #5608, opened 2019-03-12
Support connect to mysql via SSL


woosley:
it seems currently netdata python mysql check does not support connecting to mysql instance via ssl. In our case, connect to mysql via ssl is enforce on cloud, is it possible to add this feature?

ilyam8:
i think yes

> `python-mysqlclient` connection

```
        :param dict ssl:                                                                            
            dictionary or mapping contains SSL connection parameters;                               
            see the MySQL documentation for more details                                            
            (mysql_ssl_set()).  If this is set, and the client does not                             
            support SSL, NotSupportedError will be raised.
```

[mysql_ssl_set()](https://dev.mysql.com/doc/refman/8.0/en/mysql-ssl-set.html)

PR #5610, opened 2019-03-12
mysql module add ssl connection support

ilyam8:
##### Summary

Fixes: #5608 

Add ssl connection support to MySQLService

> python-mysqlclient connection

```
        :param dict ssl:                                                                            
            dictionary or mapping contains SSL connection parameters;                               
            see the MySQL documentation for more details                                            
            (mysql_ssl_set()).  If this is set, and the client does not                             
            support SSL, NotSupportedError will be raised.
```

[SSL connection parameters:](https://dev.mysql.com/doc/refman/5.6/en/mysql-ssl-set.html)

 - **key**: The path name of the client private key file.
 - **cert**: The path name of the client public key certificate file.
 - **ca**: The path name of the Certificate Authority (CA) certificate file. This option, if used, must specify the same certificate used by the server.
 - **capath**: The path name of the directory that contains trusted SSL CA certificate files.
 - **cipher**: The list of permitted ciphers for SSL encryption.

##### Component Name

[`collectors/python.d.plugin/python_modules/bases/FrameworkServices/MySQLService`](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/python_modules/bases/FrameworkServices/MySQLService.py)

___

@woosley please test it

```yaml
tcp:
  name     : 'local'
  user     : 'user'
  pass     : 'pass'
  host     : 'localhost'
  port     : '3306'
  ssl:
    key:   'path/to/key'
    cert:  'path/to/cet'
    ca:    'path/to/ca'
```

### DIFF
--- a/collectors/python.d.plugin/mysql/README.md
+++ b/collectors/python.d.plugin/mysql/README.md
@@ -7,7 +7,7 @@ Module monitors one or more mysql servers
 
 It will produce following charts (if data is available):
 
-1. **Bandwidth** in kbps
+1. **Bandwidth** in kilobits/s
  * in
  * out
 
@@ -16,9 +16,15 @@ It will produce following charts (if data is available):
  * questions
  * slow queries
 
-3. **Operations** in operations/sec
- * opened tables
- * flush
+3. **Queries By Type** in queries/s
+ * select
+ * delete
+ * update
+ * insert
+ * cache hits
+ * replace
+
+4. **Handlerse** in handlers/s
  * commit
  * delete
  * prepare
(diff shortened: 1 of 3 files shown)


woosley:
I only tested the case with ca configured, it seems other options are optional, and it works with MySQL-python in my case, but not pymysql. Pymysql complains about certificate not matching, maybe because I used a wildcard certificate. 

python is 2.7.5 and os is centos7.5, all packages are installed via yum. 
I would say the patch is ready even with the issue I mentioned above

ilyam8:
>  but not pymysql. Pymysql complains about certificate not matching, maybe because I used a wildcard certificate.


i checked and `pymysql` has same syntax, but it may not support `capath` and `cipher`, it depends on version.

PR #5612, opened 2019-03-12
Fix a bug in cgroups renaming

### LINKED ISSUE #5606: Proxmox containers no longer being renamed
##### Bug report summary
Starting recently (discovered today) Proxmox containers no longer appear to be renamed.  They are just showing up as their raw lxc container "names".  

##### OS / Environment
Debian GNU/Linux 9.8 (stretch)
Proxmox 5.3-11

netdata v1.12.2-28-nightly (re-check with latest nightly on clean install on a test VM running same Proxmox release)
netdata v1.12.2-27-g2c55171 (discovered here while testing pull for #5477)

##### Component Name
cgroups

##### Steps To Reproduce
Install netdata using kickstart.sh

##### Expected behavior
In the past the Proxmox container cgroup would be renamed to using the hostname provided in /etc/pve for the specific container.

vlvkobal:
##### Summary
`chart_id` has a slightly different format than `id` and the cgroup renaming script depends on this format. So, `chart_id` has to be passed to the script as a parameter instead of `id`.

Fixes #5606

##### Component Name
cgroups.plugin

##### Additional Information
The bug was introduced in #5557. At least proxmox containers were renamed incorrectly.

### DIFF
--- a/collectors/cgroups.plugin/sys_fs_cgroup.c
+++ b/collectors/cgroups.plugin/sys_fs_cgroup.c
@@ -888,7 +888,7 @@ static inline void cgroup_get_chart_name(struct cgroup *cg) {
     pid_t cgroup_pid;
     char command[CGROUP_CHARTID_LINE_MAX + 1];
 
-    snprintfz(command, CGROUP_CHARTID_LINE_MAX, "exec %s '%s'", cgroups_rename_script, cg->id);
+    snprintfz(command, CGROUP_CHARTID_LINE_MAX, "exec %s '%s'", cgroups_rename_script, cg->chart_id);
 
     debug(D_CGROUP, "executing command \"%s\" for cgroup '%s'", command, cg->id);
     FILE *fp = mypopen(command, &cgroup_pid);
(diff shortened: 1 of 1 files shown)


collectors/cgroups.plugin/sys_fs_cgroup.c
```diff
@@ -908,12 +908,14 @@ static inline void cgroup_get_chart_name(struct cgroup *cg) {
                 if(likely(!name_error))
                     cg->pending_renames = 0;
 
-                freez(cg->chart_title);
-                cg->chart_title = cgroup_title_strdupz(s);
+                if(likely(cg->pending_renames < 2)) {
```
cakrit: What does this if achieve? I don't understand it.

cakrit: Essentially, the first time we call rename, right after setting pending_renames to 2, we through away the value if we got the error. The second time, we keep it. Got it.

vlvkobal: There was a problem with a docker container renaming.  `cg->chart_id = cgroup_chart_id_strdupz(s)` will change the name of the docker container on the first rename and renaming will fail on the second attempt.

PR #5618, opened 2019-03-12
Fix Postgres connect, incorrect port value

Jaxmetalmax:
##### Summary
Fix postgres connection, port value was incorrect, it was taking host value instead of port.

##### Component Name
[collectors/python.d.plugin/postgres](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/postgres)


ilyam8:
i tested using unix socket :sweat_smile:

ilyam8:
thx @Jaxmetalmax

Issue #5616, opened 2019-03-12
python.d.plugin next update calc bug


ilyam8:
There is a bug in calculation update time

https://github.com/netdata/netdata/blob/6f0d13fbb98978cfca850d72a6b90d6db56e9a2a/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SimpleService.py#L41-L43

```python
In [1]: import time                                                                                               

In [2]: def c(i): 
      :     t = time.monotonic() 
      :     d = t - (t%i) + i 
      :     return t, d, d - t 
      :                                                                                                           

In [3]: c(60)                                                                                                     
Out[3]: (84688.099451557, 84720.0, 31.900548442994477)

In [4]: c(60)                                                                                                     
Out[4]: (84689.748439694, 84720.0, 30.25156030600192)

In [5]: c(60)                                                                                                     
Out[5]: (84693.18335048, 84720.0, 26.816649519998464)

In [6]: c(60)                                                                                                     
Out[6]: (84702.413057355, 84720.0, 17.58694264499354)

In [7]: c(60)                                                                                                     
Out[7]: (84720.166530715, 84780.0, 59.83346928500396)
```

That means that module first poll can be delayed up to `update_every` seconds, which is bad if update_every value is high.

Will be fixed in a separate PR.

_Originally posted by @ilyam8 in https://github.com/netdata/netdata/pull/5365#issuecomment-472092881_

ilyam8:
This can't be fixed and the problem is not specific to `python.d.plugin`. 

Change global update every to 300, restart netdata.service and see the dashboard :smile:

![Screenshot_20190403_144236](https://user-images.githubusercontent.com/22274335/55476239-bf4ccd80-561e-11e9-8966-361eeb92ccfa.png)


However, from what i understand, this can't be easily fixed, because `now - (now % update_every) + update_every;` is mandatory.

And we have 2 options:
 - leave it as is (delay up to update_every)
 - use `now - (now % 1) + 1;` for first run, and `now - (now % update_every) + update_every;` for next runs, which can lead to 2 runs in a row w/o delay

i don't like both. 2nd is even worse.

__

i may be missing something, dunno :thinking:

PR #5620, opened 2019-03-12
[wip] python.d.plugin: calc_next fix

ilyam8:
##### Summary

Fix: #5616

Suggested `next_run` calculation method has bug - first run is delayed up to `update_every` seconds, which is a problem if `update_every` value is high. 

```java
       /*
        * find the time of the next loop
        * this makes sure we are always aligned
        * with the netdata daemon
        */
       next_run = now - (now % update_every) + update_every;
```

This PR fixes the bug for `python.d.plugin`. 
`now - (now % 1) + 1` is used for first run calculation regardless of the `update_every` value.

##### Component Name

[`collectors/python.d.plugin/python_modules/bases/FrameworkServices/SimpleService`](collectors/python.d.plugin/python_modules/bases/FrameworkServices)

### DIFF
--- a/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SimpleService.py
+++ b/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SimpleService.py
@@ -39,8 +39,9 @@ class RuntimeCounters:
         self.runs = 1
 
     def calc_next(self):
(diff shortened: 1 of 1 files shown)


ilyam8:
It seems that `next_run = now - (now % update_every) + update_every;` is mandatory to minimize interpolation and to not reduce the accuracy of the data.

Issue #5624, opened 2019-03-13
ssl_check memory leak


ilyam8:
##### Bug report summary

`ssl_check` module has memory leak

looks like that is the problem:

> memory leak in SSLSocket.getpeercert()

https://bugs.python.org/issue29738

![Screenshot_20190313_190814](https://user-images.githubusercontent.com/22274335/54295395-fd2c7800-45c3-11e9-9c6c-68973a4d64be.png)

^^ i use `update_every` 1 in my test

##### OS / Environment

debian9.6, Python 2.7.13 (default, Sep 26 2018, 18:42:22)

##### Netdata version (ouput of `netdata -V`)

netdata v1.12.2-40-g97699c5c

##### Component Name

[collectors/python.d.plugin/sslcheck](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/sslcheck)

##### Steps To Reproduce
 - run `ssl_check` module
 - wait
 - ~profit~ memory leak

https://bugs.python.org/issue25569

```python
import socket
import ssl
import sys


def main():
    ctx = ssl.create_default_context()
    s = socket.create_connection(('www.bing.com', 443))
    s = ctx.wrap_socket(s, server_hostname='www.bing.com')
    while True:
        s.getpeercert()

        sys.stderr.write('.')
        sys.stderr.flush()


if __name__ == '__main__':
    main()
```

##### Expected behavior

no memory leak

ilyam8:
@p-thurner

ilyam8:
`python` version will be removed, this module will be rewritten in `go`

p-thurner:
oh. pity :/

ilyam8:
yes :+1:

PR #5626, opened 2019-03-13
Remove ssl check module

ilyam8:
##### Summary

Fixes: #5624

> memory leak in SSLSocket.getpeercert()

https://bugs.python.org/issue29738

##### Component Name
[collectors/python.d.plugin/sslcheck](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/sslcheck)


ilyam8:
alarm was not removed on purpose

ilyam8:
debian9.6, Python 2.7.13 (default, Sep 26 2018, 18:42:22)

```python
import socket
import ssl
import sys
import gc



def main():
    v = 0
    ctx = ssl.create_default_context()
    s = socket.create_connection(('www.bing.com', 443))
    s = ctx.wrap_socket(s, server_hostname='www.bing.com')
    while True:
        v += 1
        s.getpeercert()
        if v%100==0:
            print('COLLECT')
            gc.collect()

        sys.stderr.write('.')
        sys.stderr.flush()


if __name__ == '__main__':
    main()
```

Screenshot_20190313_200734

Issue #5631, opened 2019-03-14
ssl check module


ilyam8:
##### Feature idea summary
Since python [`sslcheck`](https://github.com/netdata/netdata/pull/5365) implementation  [was removed](https://github.com/netdata/netdata/pull/5626) due to [a bug](https://github.com/netdata/netdata/issues/5624) we need to rewrite the module in `go`.

##### Expected behavior

`sslcheck` module in `go`

PR #5645, opened 2019-03-14
x509check alarm fix

ilyam8:
##### Summary

rename

> health/health.d/sslcheck.conf → health/health.d/x509check.conf

**Why**

sslcheck module (#5365) was removed(#5626) because of memory leak bug (#5624).

The module was rewritten in go (#5631, https://github.com/netdata/go.d.plugin/pull/166). New module name - `x509check`. 

This PR changes name of the alarm.

### DIFF
--- a/health/Makefile.am
+++ b/health/Makefile.am
@@ -68,7 +68,6 @@ dist_healthconfig_DATA = \
     health.d/retroshare.conf \
     health.d/softnet.conf \
     health.d/squid.conf \
-    health.d/sslcheck.conf \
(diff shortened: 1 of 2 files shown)


paulkatsoulakis:
Hi, is there an issue referring to this change?
Otherwise it would be nice to have some more details on the PR description around the nature of the change. Regardless it's simplicity, it would help for people without context understand the reasoning faster

ilyam8:
@paulkatsoulakis you are right, i added missing info.

PR #5656, opened 2019-03-16
netdata/build: A small house keeping around coverity scan (no functional  or process change)

paulkatsoulakis:
##### Summary
This is a small cleanup on the coverity scan scriptlet.
The idea is to separate the install process from the actual scanning to distinct and simplify the two mechanisms, so that they can be reusable separately.

1) Split coverity install on seperate scriptlet, call it through travis when needed
2) Remove shellcheck exclussions, they dont seem to be raised any more (i might be using wrong version of shellcheck, will see)
3) Fix authors,copyrights etc
4) Make sure all echo goes to stderr
5) When installing, use mktemp for the temporary directory and put everything there. Remove afterwards

##### Component Name
netdata/build

##### Additional Information
None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,7 @@ jobs:
     script: ./tests/installer/checksums.sh
   - name: coverity
     install: sudo apt-get install -y zlib1g-dev uuid-dev libipmimonitoring-dev libmnl-dev libnetfilter-acct-dev
-    script: ./coverity-scan.sh || echo "Coverity failed :("
-    env: INSTALL_COVERITY="true"
+    script: ./coverity-install.sh && ./coverity-scan.sh || echo "Coverity failed :("
     if: type = cron
 
   - stage: build
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ jobs:
   - name: coverity
     install: sudo apt-get install -y zlib1g-dev uuid-dev libipmimonitoring-dev libmnl-dev libnetfilter-acct-dev
     script: ./coverity-install.sh && ./coverity-scan.sh || echo "Coverity failed :("
-    if: type = cron
+#   if: type = cron
 
   - stage: build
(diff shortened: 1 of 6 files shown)


cakrit:
How was this tested? Did the travis build on this PR use that code? How can we verify it works?

paulkatsoulakis:
> How was this tested? Did the travis build on this PR use that code? How can we verify it works?

Syntax-wise is something that travis allows.
On top of that, i am setting up my forked netdata repo to work with travis end to end so that i can have a "beta" stack for the CI flows and i will verify it works there before merging

paulkatsoulakis:
Outcome will show up here: https://travis-ci.com/paulkatsoulakis/netdata/jobs/185822831
The specific activity runs on cron only, i will have to revert a temp commit after the execution finishes

paulkatsoulakis:
Oops, i was missing the coverity token from the env, i will re-run it

paulkatsoulakis:
Here's the successful coverity logic run:
https://travis-ci.com/paulkatsoulakis/netdata/jobs/185972167
In order to do the test, i had to remove the line that forces the execution to happen through cron.

FWIW I had to do more changes after all to make this run from my personal account.
They are required if we want to have a beta stack to work with. I also had a few bugs that i had to fix.

Note: for this change to be fully complete, we should parse the last curl result in coverity-scan.sh and check for any message other than "Build successfully submitted" and report error, otherwise we won't know if coverity worked or not as we always say success regardless the result of curl (I had an Access Denied while testing that wasn't detected).

@cakrit Should i push this change too or not worth doing much nit picking?

cakrit:
It's a good idea to have a log of whether the scan was initiated or not, so I'd say do it.

paulkatsoulakis:
> It's a good idea to have a log of whether the scan was initiated or not, so I'd say do it.

cool, done

Issue #5629, opened 2019-03-13
Isn't it a security risk not having SSL mode for postgres plugin?


DenLilleMand:
##### Question summary
Isn't it a security risk not having SSL mode for postgres plugin?
In the normal configuration you can specify a SSLMode and a key.

##### OS / Environment
Any

##### Component Name
Postgres
##### Expected results
To have the normal configuration options the driver supports

ilyam8:
https://www.postgresql.org/docs/9.1/libpq-ssl.html

PR #5661, opened 2019-03-18
postgres add ssl connection support

ilyam8:
##### Summary

Fixes: #5629

https://www.postgresql.org/docs/9.1/libpq-ssl.html

##### Component Name

[/collectors/python.d.plugin/postgres](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/postgres)

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.conf
+++ b/collectors/python.d.plugin/postgres/postgres.conf
@@ -71,7 +71,7 @@
 #     connect_timeout     : 2                  # in seconds, default is 2
 #     statement_timeout   : 2000               # in ms, default is 2000
 #
-# SSL connection parameters (https://www.postgresql.org/docs/9.1/libpq-ssl.html)
+# SSL connection parameters (https://www.postgresql.org/docs/current/libpq-ssl.html)
 #     sslmode             : mode               # one of [disable, allow, prefer, require, verify-ca, verify-full]
 #     sslrootcert         : path/to/rootcert   # the location of the root certificate file
 #     sslcrl              : path/to/crl        # the location of the CRL file
--- a/collectors/python.d.plugin/postgres/postgres.conf
+++ b/collectors/python.d.plugin/postgres/postgres.conf
@@ -68,8 +68,15 @@
 #     password            : 'example_pass'
 #     host                : 'localhost'
 #     port                : 5432
-#     connect_timeout     : 2     # in seconds, default is 2
-#     statement_timeout   : 2000  # in ms, default is 2000
+#     connect_timeout     : 2                  # in seconds, default is 2
+#     statement_timeout   : 2000               # in ms, default is 2000
+#
+# SSL connection parameters (https://www.postgresql.org/docs/9.1/libpq-ssl.html)
+#     sslmode             : mode               # one of [disable, allow, prefer, require, verify-ca, verify-full]
+#     sslrootcert         : path/to/rootcert   # the location of the root certificate file
+#     sslcrl              : path/to/crl        # the location of the CRL file
+#     sslcert             : path/to/cert       # the location of the client certificate file
+#     sslkey              : path/to/key        # the location of the client key file
 #
 # Additionally, the following options allow selective disabling of charts
 #
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -824,7 +824,7 @@ class Service(SimpleService):
                 conf.get('statement_timeout', DEFAULT_STATEMENT_TIMEOUT)),
         }
 
-        # https://www.postgresql.org/docs/9.1/libpq-ssl.html
+        # https://www.postgresql.org/docs/current/libpq-ssl.html
         ssl_params = {
             'sslmode': conf.get('sslmode'),
             'sslrootcert': conf.get('sslrootcert'),
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -23,27 +23,41 @@ DEFAULT_CONNECT_TIMEOUT = 2       # seconds
 DEFAULT_STATEMENT_TIMEOUT = 5000  # ms
 
 
-WAL = 'WAL'
-ARCHIVE = 'ARCHIVE'
-BACKENDS = 'BACKENDS'
-TABLE_STATS = 'TABLE_STATS'
-INDEX_STATS = 'INDEX_STATS'
-DATABASE = 'DATABASE'
-BGWRITER = 'BGWRITER'
-LOCKS = 'LOCKS'
-DATABASES = 'DATABASES'
-STANDBY = 'STANDBY'
-REPLICATION_SLOT = 'REPLICATION_SLOT'
-STANDBY_DELTA = 'STANDBY_DELTA'
-REPSLOT_FILES = 'REPSLOT_FILES'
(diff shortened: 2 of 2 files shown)


ilyam8:
@DenLilleMand could you test it?

@anayrat

DenLilleMand:
@ilyam8 yes, i will have a look

collectors/python.d.plugin/postgres/postgres.conf
```diff
@@ -68,8 +68,15 @@
 #     password            : 'example_pass'
 #     host                : 'localhost'
 #     port                : 5432
-#     connect_timeout     : 2     # in seconds, default is 2
-#     statement_timeout   : 2000  # in ms, default is 2000
+#     connect_timeout     : 2                  # in seconds, default is 2
+#     statement_timeout   : 2000               # in ms, default is 2000
+#
+# SSL connection parameters (https://www.postgresql.org/docs/9.1/libpq-ssl.html)
```
anayrat: 9.1 is deprecated, you should use "current" documentation : https://www.postgresql.org/docs/current/libpq-ssl.html

ilyam8: i see nothing has been changed since 9.1 for ssl, same modes, same params, so the only ivalid thing is link.

Thank you @anayrat

ilyam8: @anayrat btw if there is a chance you could test it would be very nice

DenLilleMand:
> @DenLilleMand could you test it?

@ilyam8 sorry for the delay, thanks for fixing this, i had to figure out how to checkout a PR from a fork first. Install it locally and actually debug my own setup of this postgres plugin.


First of all, it works for the options `sslrootcert=/path/to/file` and `sslmode=verify-full`. The way i know it works is that if i e.g. give it a invalid file to `sllrootcert` with `sslmode=verify-full` value it crashes. And i checked that the params passed to the connection string actually exist, with a debug stmt.

I think that the link in the code for documentation should also point to: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
so people can see the exact names of all of the possible params.
Looking at the list it is also clear that we dont support all of it, but i think these are the normal params you have included.

More code related my concern is that you just copy in the options, so someone believing that they're running SSL, but had a wrong configuration will be deluded. As an example if i add just the `sslrootcert`, but don't specify the `sslmode` it actually just passes as successful. Probably the same for the rest of the params. My suggestion is making it mandatory adding the `sslmode` option(if you have added any other SSL options).  Then they will get the needed crashing/errors on a ssl param failure like e.g. cannot find the file specified in `sslrootcert` or one of the other certificates/files.

ilyam8:
> Could consider a more generic way of solving it simply passing everything from the configuration in as keys/values

Please open a new issue, the request is nice :+1: 

> My suggestion is making it mandatory adding the sslmode option(if you have added any other SSL options).

Agree.

ilyam8:
@DenLilleMand i made suggested changed, please test it again

DenLilleMand:
> @DenLilleMand i made suggested changes, please test it again

You have a off by one 
```
        if CONN_PARAM_SSL_MODE not in ssl_params and len(ssl_params) > 1:
            raise ValueError("mandatory 'sslmode' param is missing, please set")
```
Should be
```
        if CONN_PARAM_SSL_MODE not in ssl_params and len(ssl_params) > 0:
            raise ValueError("mandatory 'sslmode' param is missing, please set")
```
When changing that it seems to work just fine :+1:

DenLilleMand:
And it is not important, maybe i just don't understand. But i have a bunch of sections socket,tcp,tcpipv4,tcpipv6:
```
socket:
    name     : 'xxxxxxxxxx'
    user     : 'xxxxxx'
    host     : 'xxxxx'
    database : 'xxxx'
    password : 'XXXXXXXXXXXXXXXXXXXX'
    #sslrootcert  : '/etc/netdata/python.d/xxxx.pem'
    sslmode  : 'verify-full'

tcp:
    name     : 'xxxxxxx'
    database : 'xxxx'
    user     : 'xxxx'
    host     : 'XXXXXXXXXXX'
    port     : 5432
    password : 'XXXXXXXXXXXXXX'
    sslmode  : 'verify-full'
    sslrootcert  : '/etc/netdata/python.d/xxxx.pem'
...continues with 3 more sections....
```
I think that this is the default postgres config file taken from `/etc/netdata/orig/python.d/postgres.conf` but it works only to leave
in the configuration like this:
```
name     : 'xxxxxxx'
database : 'xxxx'
user     : 'xxxx'
host     : 'XXXXXXXXXXX'
port     : 5432
password : 'XXXXXXXXXXXXXX'
sslmode  : 'verify-full'
```
i guess i am just wondering why that is. Of course that is not for this PR to fix

ilyam8:
> But i have a bunch of sections socket,tcp,tcpipv4,tcpipv6:

Well, those are different jobs. You can have any number of jobs (local db, remote db, etc.)

> JOBS (data collection sources)
> 
> The default JOBS share the same *name*. JOBS with the same name
> are mutually exclusive. Only one of them will be allowed running at
> any time. This allows autodetection to try several alternatives and
> pick the one that works.


> but it works only to leave
in the configuration like this:

Ye, it will work too. In that case config will be considered as single job config.

ilyam8:
> @DenLilleMand i made suggested changes, please test it again

i created [the issue](https://github.com/netdata/netdata/issues/5684)

DenLilleMand:
> > @DenLilleMand i made suggested changes, please test it again
> 
> i created [the issue](https://github.com/netdata/netdata/issues/5684)

Yeah i already tested it, it works :) thanks

DenLilleMand:
> > But i have a bunch of sections socket,tcp,tcpipv4,tcpipv6:
> 
> Well, those are different jobs. You can have any number of jobs (local db, remote db, etc.)
> 
> > JOBS (data collection sources)
> > The default JOBS share the same _name_. JOBS with the same name
> > are mutually exclusive. Only one of them will be allowed running at
> > any time. This allows autodetection to try several alternatives and
> > pick the one that works.
> 
> > but it works only to leave
> > in the configuration like this:
> 
> Ye, it will work too. In that case config will be considered as single job config.
> 
> code:
> 
> [netdata/collectors/python.d.plugin/python.d.plugin.in](https://github.com/netdata/netdata/blob/27f18ef651c6ccd2eb3ebf1938219fd674734b86/collectors/python.d.plugin/python.d.plugin.in#L370-L390)
> 
> Lines 370 to 390 in [27f18ef](/netdata/netdata/commit/27f18ef651c6ccd2eb3ebf1938219fd674734b86)
> 
>  def gather_jobs_configs(self): 
>      job_names = [v for v in self.config if isinstance(self.config[v], dict)] 
>   
>      if len(job_names) == 0: 
>          job_conf = JobConf(JOB_BASE_CONF) 
>          job_conf.set_defaults_from_module(self.source) 
>          job_conf.update(self.config) 
>          job_conf.set_job_name(self.name) 
>          job_conf.set_override_name(job_conf.pop('name')) 
>          return [job_conf] 
>   
>      configs = list() 
>      for job_name in job_names: 
>          raw_job_conf = self.config[job_name] 
>          job_conf = JobConf(JOB_BASE_CONF) 
>          job_conf.set_defaults_from_module(self.source) 
>          job_conf.set_defaults_from_config(self.config) 
>          job_conf.update(raw_job_conf) 
>          job_conf.set_job_name(job_name) 
>          job_conf.set_override_name(job_conf.pop('name')) 
>          configs.append(job_conf)

Oh okay, thanks for the thorough explanation!

PR #5662, opened 2019-03-18
Add SMS Server Tools 3 SMS notifications

### LINKED ISSUE #2720: notifications with smstools (direct connection to mobile phone)
Looking at implementing local SMS for alerts, mostly for critical. I found smstools3, which Ubuntu has a package for and seem pretty easy to do a bash script send messages.  
  
Any reason anybody else can think of not to use smstools3 or a better local SMS gateway?  
  
Worth making documentation and possibly a PR for the script?

cakrit:
##### Summary
Fixes #2720 

##### Component Name
health

##### Additional Information
Can't test it, we'll wait for user tests.


paulkatsoulakis:
> Can't test it, we'll wait for user tests.

What if we purchase a raspberry pi with a gsm module?
Might worth having the raspberry for netdata usage too

Issue #651, opened 2016-07-05
netdata package maintainers


ktsaou:
# This issue has been converted to a wiki page

For the latest info check it here: https://github.com/firehol/netdata/wiki/netdata-package-maintainers

---

I think it would be useful to prepare a wiki page with information about the maintainers of netdata for the Linux distributions, automation systems, containers, etc.

Let's see who is who:

---
## Official Linux Distributions

| Linux Distribution | Netdata Version | Maintainer | Related URL |
| :-: | :-: | :-: | :-- |
| Arch Linux | Release | @svenstaro | [netdata @ Arch Linux](https://www.archlinux.org/packages/community/x86_64/netdata/) |
| Arch Linux AUR | Git | @sanskritfritz | [netdata @ AUR](https://aur.archlinux.org/packages/netdata-git/) |
| Gentoo Linux | Release + Git | @candrews | [netdata @ gentoo](https://github.com/gentoo/gentoo/tree/master/net-analyzer/netdata) |
| Debian | Release | @lhw @FedericoCeratto | [netdata @ debian](https://anonscm.debian.org/git/collab-maint/netdata.git) |
| Slackware | Release | @willysr | [netdata @ slackbuilds](https://slackbuilds.org/repository/14.2/system/netdata/)
| Ubuntu |  |  |  |
| Red Hat / Fedora / Centos |  |  |  |
| SuSe / openSuSe |  |  |  |

---
## FreeBSD

System|Initial PR|Core Developer|Package Maintainer
|:-:|:-:|:-:|:-:|
FreeBSD|#1321|@vlvkobal|@mmokhi

---
## MacOS

System|URL|Core Developer|Package Maintainer
|:-:|:-:|:-:|:-:|
MacOS Homebrew Formula|[link](https://github.com/Homebrew/homebrew-core/blob/master/Formula/netdata.rb)|@vlvkobal|@rickard-von-essen

---
## Unofficial Linux Packages

| Linux Distribution | Netdata Version | Maintainer | Related URL |
| :-: | :-: | :-: | :-- |
| Ubuntu | Release | @gslin | [netdata @ gslin ppa](https://launchpad.net/~gslin/+archive/ubuntu/netdata) https://github.com/firehol/netdata/issues/69#issuecomment-217458543 |

---
## Embedded Linux

| Embedded Linux | Netdata Version | Maintainer | Related URL |
| :-: | :-: | :-: | :-- |
[...]

PR #5670, opened 2019-03-19
Fix for: postgres: socket: update() unhandled exception: integer out of range

WoLpH:
##### Summary
When running netdata on the PostgreSQL 11 install on a FreeBSD 11.2 system I got the error:

```
postgres: socket: update() unhandled exception: integer out of range
```

##### Component Name
PostgreSQL

##### Additional Information

The fix is rather simple. Calculate the result using a `BIGINT` instead of an `INT to avoid out-of-range issues.

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -430,7 +430,7 @@ GROUP BY
     V10: """
 WITH wal_size AS (
   SELECT
-    current_setting('wal_block_size')::INT * setting::INT AS val
+    current_setting('wal_block_size')::BIGINT * setting::INT AS val
   FROM pg_settings
   WHERE name = 'wal_segment_size'
   )


WoLpH:
The CLA system appears to be down unfortunately, resulting in an internal server error for: https://cla-assistant.io/auth/github/callback?code=xxx

So I hereby declare to agree with the CLA

ilyam8:
i see you are fixing v10 query

https://github.com/netdata/netdata/blob/02d823b815abe8e156b3a4b6ec7aa0ccf7df39a6/collectors/python.d.plugin/postgres/postgres.chart.py#L430-L433

Since you have `PostgreSQL 11` DEFAULT version of the query should be returned by query factory

https://github.com/netdata/netdata/blob/02d823b815abe8e156b3a4b6ec7aa0ccf7df39a6/collectors/python.d.plugin/postgres/postgres.chart.py#L557-L560

___

Please run `python.d.plugin` in debug mode
https://docs.netdata.cloud/collectors/python.d.plugin/#how-to-debug-a-python-module

ilyam8:
@WoLpH 	
btw what is your netdata version?

ilyam8:
i recall the `BIGINT` problem, but i believe it is fixed in latest @anayrat

anayrat:
Yes, it should be fixed. As @ilyam8  pointed out, the `DEFAULT` query should used. Maybe there is an issue in detecting postgres version?

WoLpH:
I'm using the latest netdata release that's available as a FreeBSD package, which is 1.11.0_rolling

As  far as I can see that's still the newest available: https://www.freshports.org/net-mgmt/netdata

I can try a manual install from source to see if that makes a difference

ilyam8:
> I'm using the latest netdata release that's available as a FreeBSD package, which is 1.11.0_rolling

1.11.0 is old, please update to the latest, i am sure the problem is fixed there

WoLpH:
I can confirm that indeed fixes it. Now the next question, how do we get a new release into freebsd...

WoLpH:
Thank you all for the help!

ilyam8:
> how do we get a new release into freebsd...

it is a good question,according #651 @mmokhi is the [maintainer](https://www.freshports.org/net-mgmt/netdata)

mmokhi:
> > how do we get a new release into freebsd...
> 
> it is a good question,according #651 @mmokhi is the [maintainer](https://www.freshports.org/net-mgmt/netdata)

Hi,
Sorry for my recent AFK, has been busy with the university very intense :slightly_smiling_face: 
The package has been updated to the latest release < https://svnweb.freebsd.org/changeset/ports/496470 >

Happy spring :tulip: :blush:

WoLpH:
Excellent! Thank you so much @mmokhi  :)

mmokhi:
You're welcome,
My pleasures that I'm a help :blush:

PR #5678, opened 2019-03-21
netdata/packaging/uninstaller: Update uninstall document

paulkatsoulakis:
##### Summary
Complementary PR for #5675, update doc
Parent issue #5290 

##### Component Name
netdata/packaging/uninstaller

##### Additional Information
None


paulkatsoulakis:
> * It's not in the current dir (not downloading anything)

yikes, sorry for missing to add the right path.

> * Don't assume they have it, the instruction to download should remain if they don't find it in `/usr/libexec...`

I thought about it and i figured it wouldn't be necessary to let the download option.
But you 're definitely right, i didn't consider existing installations that will need the right instructions too.

I will fix asap, sorry for that

Issue #5681, opened 2019-03-22
push kickstart.sh to https://my-netdata.io/kickstart.sh


MichaelRiss:
##### Bug report summary
`packaging/installer/kickstart.sh` needs to be pushed to `https://my-netdata.io/kickstart.sh` to fix #5584 for users updating with `bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait`

##### OS / Environment
Fedora 28 running plainly on a x86_64 machine

##### Netdata version (ouput of `netdata -V`)
netdata v1.12.2-46-nightly

##### Component Name
packaging/installer

##### Steps To Reproduce
`bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait` still suffers from #5584
##### Expected behavior
`bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait` should not suffer
 from #5584 anymore

paulkatsoulakis:
Hello @MichaelRiss, excellent catch! 
Sorry for the inconvenience, indeed we seem to have missed it! 
It will be included asap

cakrit:
This is something @ktsaou does, I notified him.
@paulkatsoulakis please add a check to the daily builds to verify that the kickstart.sh and kickstart-static64.sh md5sums are the same in the repo, packaging/installer/README.md and http://my-netdata.io
We have a travis test already for the first two, it runs even for PRs. The third one can't be tested for a PR though, my-netdata.io is updated manually as a second step. When we detect a discrepancy, see if travis can somehow notify @ktsaou that he needs to update the file.

paulkatsoulakis:
This is fixed now and we also increased our monitoring capability so that we mitigate this in the future in a timely manner.
Moving forward we will definitely look into automating the whole process to avoid this ever happening in the future.

@MichaelRiss let us know with a new issue if you face any issue again.

PR #5689, opened 2019-03-23
Integrity testing: Check published kickstart files integrity

paulkatsoulakis:
##### Summary
As part of issue #5681 we are automating the integrity testing of published kickstart files, so that we are notified on a timely manner.
This is just a preparatory step, we need to further adjust the pipeline so that it does not affect the overall build process when kickstart integrity testing fails.

##### Component Name
netdata/packaging/installer

##### Additional Information
None


.travis/validate_kickstart_integrity.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#
+# Mechanism to validate kickstart files integrity status
+#
+# Copyright: SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Author  : Pavlos Emm. Katsoulakis (paul@netdata.cloud)
+
+KICKSTART_URL="https://my-netdata.io/kickstart.sh"
+KICKSTART_MD5="e051d1baed4c69653b5d3e4588696466"
```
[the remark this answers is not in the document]
paulkatsoulakis: Well noted, merged and cleaned up

tests/installer/checksums.sh
```diff
@@ -1,19 +1,49 @@
...
-		echo "current checksum: $NEW_CHECKSUM"
-		exit 1
+	CHECKSUM_IN_README=$(grep "$file" $README_DOC | grep md5sum | cut -d '"' -f2)
+	KICKSTART_URL="https://my-netdata.io/$file"
+	KICKSTART_MD5="$(md5sum "packaging/installer/$file" | cut -d' ' -f1)"
+	CALCULATED_MD5="$(curl -Ss ${KICKSTART_URL} | md5sum | cut -d ' ' -f 1)"
+
+	# Conditionally run the website validation
+	if [ -z "${LOCAL_ONLY}" ]; then
+		echo "Validating ${KICKSTART_URL} against ${KICKSTART_MD5}.."
```
cakrit: should be `against local file`, or `against packaging/installer/$file`

tests/installer/checksums.sh
```diff
@@ -1,19 +1,49 @@
...
 
 for file in kickstart.sh kickstart-static64.sh; do
-	OLD_CHECKSUM=$(grep "$file" packaging/installer/README.md | grep md5sum | cut -d '"' -f2)
-	NEW_CHECKSUM="$(md5sum "packaging/installer/$file" | cut -d' ' -f1)"
-	if [ "$OLD_CHECKSUM" != "$NEW_CHECKSUM" ]; then
-		echo "Invalid checksum for $file in docs."
-		echo "checksum in docs: $OLD_CHECKSUM"
-		echo "current checksum: $NEW_CHECKSUM"
-		exit 1
+	CHECKSUM_IN_README=$(grep "$file" $README_DOC | grep md5sum | cut -d '"' -f2)
```
cakrit: I'd name this var README_MD5, for consistency.

tests/installer/checksums.sh
```diff
@@ -1,19 +1,49 @@
...
+	KICKSTART_MD5="$(md5sum "packaging/installer/$file" | cut -d' ' -f1)"
+	CALCULATED_MD5="$(curl -Ss ${KICKSTART_URL} | md5sum | cut -d ' ' -f 1)"
+
+	# Conditionally run the website validation
+	if [ -z "${LOCAL_ONLY}" ]; then
+		echo "Validating ${KICKSTART_URL} against ${KICKSTART_MD5}.."
+		if [ "$KICKSTART_MD5" == "$CALCULATED_MD5" ]; then
+			echo "${KICKSTART_URL} looks fine"
+		else
+			echo "${KICKSTART_URL} has changed, please update it"
```
cakrit: I'd say "${KICKSTART_URL} md5sum does not match local file. It needs to be updated"

tests/installer/checksums.sh
```diff
@@ -1,19 +1,49 @@
...
+			exit 1
```
cakrit: Exiting here is problematic. We don't want to stop the build process, we want to notify @ktsaou to update the file. 
exiting also means that we are not checking kickstart64, if kickstart has a problem. 

After the echo, I'd append an error to an initially empty param. At the end, if the param is not empty, send a slack notification to a new github channel. Then it won't be just ktsaou that gets notified.

tests/installer/checksums.sh
paulkatsoulakis: ack

tests/installer/checksums.sh
paulkatsoulakis: ack

tests/installer/checksums.sh
paulkatsoulakis: ack

tests/installer/checksums.sh
paulkatsoulakis: I am glad you raise it. Unfortunately travis does not natively support notifications on a stage basis.
I am looking into the webhooks option though, so that i can make failures here silent and just pop a notification on a slack channel for us to track.

This first approach was intended to cover us partially using Travis slack notifications, that is why i added it as a last step in the pipeline to avoid breaking other stuff.

I think webhooks will be much cooler approach, i 'll post a commit with the necessary adjustments

paulkatsoulakis:
> Looks fine, haven't tested it.

I have tested the script itself offline and i have also tested the pipeline by letting it run without the conditional "branch = master and type = cron". I will of course keep an eye to fix any possible unexpected problems

Issue #5693, opened 2019-03-24
web_log reports unmatched lines


hsegnitz:
web_log plugin reports unmatched lines.

This is my log:

```txt

::1 localhost - [24/Mar/2019:01:16:51 +0100] "GET /status?full&json HTTP/1.1" 200 1359 "-" "-"
2a01:4f8:173:e02::2 gitlab.home.segnitz.net - [24/Mar/2019:01:16:51 +0100] "POST /api/v4/jobs/request HTTP/1.1" 204 0 "-" "gitlab-ci-multi-runner 9.5.1 (9-5-stable; go1.8.3; linux/amd64)"


```

To be honest, I changed the log format:

I replaced the "-" between the IP(v6) address and the user name with the vhost, seen that with quite a few hosters...

hsegnitz:
...found a better solution ;)

PR #5694, opened 2019-03-24
add a description how to achieve vhost access logs with nginx in web_log

hsegnitz:
##### Summary
I found out by trial and error how to make nginx log vhosts in an additional file and then how to make netdata read that file and wanted to share that with future users.

##### Component Name

web_log.conf in python collector

##### Additional Information
A minor thing: At least nginx writes IPv6 addresses in the $host variable in square brackets, so I included them in the pattern.

fixes: #5693


ilyam8:
thanks!

PR #5703, opened 2019-03-25
Send custom docs

### LINKED ISSUE #5702: Feature: Netdata alarms as POST events to a configurable endpoint in health_alarm_notify.conf
##### Feature idea summary

Simple request: I would like the ability to give an endpoint such as (http||https://my-api-service.com/netdata/alarms that Netdata will send a POST json body to in the event that alarms are triggered, I prefer this to email route so I can build application logic around some of the alarms we see and it will just give a cleaner programmatic approach.  This means that alarms have to have some sort of standard JSON structure we can model POJO's(plain old java objects) or model any object in your language of choice to parse the alarm data and act on it.

This also requires documenting the structure of the alarm netdata will send, which I currently see no such json/yaml structure outlined anywhere.

The existing conf of netdata alarms reads like the system already can use underlying curl, so it sounds like we are close to being here already, we just need to allow custom integration of our own endpoints as opposed to these vendor app notification systems with a public data structure format of all netdata alerts. 

Lastly the log level for events worthy of having POST events sent should be configurable as well, for instance in prod I want CRITICAL events only to keep noise down and such. 

Edit - I am also aware of the api endpoints we can scrape alarm data off of too, but I rather write a listening app that waits on netdata pushes to send me details than polling a given endpoint every __ seconds.

##### Expected behavior

Alarm starts going off in netdata, it sends a json body post request to an endpoint every time it triggers, to which we can take action on from an application ingesting those POST requests.

cakrit:
##### Summary
Fixes #5702 

##### Component Name
docs

##### Additional Information
Also added 4 missing parameters to the `health_alarm_notify.conf` list of variables available to the custom sender.


cakrit:
Thanks for the review @Ferroin, rebuilding to fix HTML generation issue with the lists.

PR #5707, opened 2019-03-25
NEW: authdb option for MongoDB collector

andvgal:
##### Summary

MongoDB has users bound to particular database. There are cases when monitoring user is outside of the default "admin" database, but has all required roles granted.

This PR adds a simple "authdb" parameter to use for authentication. The parameter defaults to the original "admin" value. The rest is left as is.

##### Component Name

MongoDB python collector.

##### Additional Information

N/A

### DIFF
--- a/collectors/python.d.plugin/mongodb/mongodb.chart.py
+++ b/collectors/python.d.plugin/mongodb/mongodb.chart.py
@@ -424,6 +424,7 @@ class Service(SimpleService):
         SimpleService.__init__(self, configuration=configuration, name=name)
         self.order = ORDER[:]
         self.definitions = deepcopy(CHARTS)
+        self.authdb = self.configuration.get('authdb', 'admin')
         self.user = self.configuration.get('user')
         self.password = self.configuration.get('pass')
         self.host = self.configuration.get('host', '127.0.0.1')
@@ -707,7 +708,7 @@ class Service(SimpleService):
         try:
             connection = MongoClient(**conn_vars)
             if self.user and self.password:
-                connection.admin.authenticate(name=self.user, password=self.password)
+                connection[self.authdb].authenticate(name=self.user, password=self.password)
             # elif self.user:
             #     connection.admin.authenticate(name=self.user, mechanism='MONGODB-X509')
             server_status = connection.admin.command('serverStatus')
--- a/collectors/python.d.plugin/mongodb/mongodb.chart.py
+++ b/collectors/python.d.plugin/mongodb/mongodb.chart.py
@@ -708,7 +708,7 @@ class Service(SimpleService):
         try:
             connection = MongoClient(**conn_vars)
             if self.user and self.password:
-                connection[self.authdb].authenticate(name=self.user, password=self.password)
+                getattr(connection, self.authdb).authenticate(name=self.user, password=self.password)
             # elif self.user:
             #     connection.admin.authenticate(name=self.user, mechanism='MONGODB-X509')
             server_status = connection.admin.command('serverStatus')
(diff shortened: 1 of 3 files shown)


ilyam8:
@andvgal thanks!

Please add new param to the [module configuration file](https://github.com/netdata/netdata/blob/1ad58c3b1238ed1d2d4a835111a831dc00799a07/collectors/python.d.plugin/mongodb/mongodb.conf#L62-L70).

collectors/python.d.plugin/mongodb/mongodb.chart.py
```diff
@@ -707,7 +708,7 @@ def _create_connection(self):
         try:
             connection = MongoClient(**conn_vars)
             if self.user and self.password:
-                connection.admin.authenticate(name=self.user, password=self.password)
+                connection[self.authdb].authenticate(name=self.user, password=self.password)
```
ilyam8: i see requirements is `python-pymongo package v2.4+.`

it would be very nice if you check that `connection` in py2.4 has `__getitem__` method. It is very likely, but worth to check.

collectors/python.d.plugin/mongodb/mongodb.conf
```diff
@@ -80,3 +82,11 @@ local:
...
     host : '127.0.0.1'
     port : 27017
+
+authsample:
+    name : 'secure'
+    host : 'mongodb.example.com'
+    port : 27017
+    authdb : 'admin'
+    user : 'monitor'
+    password : 'supersecret'
```
ilyam8: i suggest to comment out the job, otherwise python.d.plugin will create it and try to connect to 'mongodb.example.com'

andvgal: Oops, didn't think about that.

collectors/python.d.plugin/mongodb/mongodb.chart.py
andvgal: Are you referring to PyMongo v2.4 or Python v2.4? The former has __getitem__ even in version 1.0, but is the latter really used anywhere? I recall even old CentOS has older version and I doubt netdata is used in such cases.

andvgal: I guess `getattr()` call instead of array subscript style which calls `__getitem__()` should fit, if Python below 2.4 support is in concern.

andvgal: Done

ilyam8: i am referring to pymongo2.4

I am talking about this change
`connection.admin` to `connection[self.authdb]`

ok, i see from the code
```python
    def __getattr__(self, name):
        """Get a database by name.

        Raises :class:`~pymongo.errors.InvalidName` if an invalid
        database name is used.

        :Parameters:
          - `name`: the name of the database to get
        """
        return database.Database(self, name)

    def __getitem__(self, name):
        """Get a database by name.

        Raises :class:`~pymongo.errors.InvalidName` if an invalid
        database name is used.

        :Parameters:
          - `name`: the name of the database to get
        """
        return self.__getattr__(name)
```

so `connection.admin` == `connection[self.authdb]`, i wonder if it is true for pymongo 2.4

ilyam8: ok, this code haven't been changed since 2.4

ilyam8:
@cakrit 	this one is ready

cakrit:
Many thanks @andvgal!

Issue #5705, opened 2019-03-25
netdata/packaging/installer: Finish up netdata installer refactoring


paulkatsoulakis:
##### Feature idea summary
This issue is a wrap-up task of PR #5505 
We will use it to track any further problems identified during testing and also make it the link between the old PR and the new that i will be producing with any further fixes we may come up with.

CC: @cakrit
 
##### Expected behavior
netdata-installer.sh runs as before, with the fixes included.

paulkatsoulakis:
Last time the previous PR was tested, the following problem was identified when run stable and improved versions of netdata in brand new containers:


Raw diff of logs between the two versions
```
$ diff install.err.paulfantom install.err.stable 
7c7
< 
---
> Press ENTER to build and install netdata to your system > 
131c131
< [/root/netdata52]# chown -R root:netdata /usr/libexec/netdata 
---
> [/root/netdata52]# chown -R root /usr/libexec/netdata 
158c158
< [/root/netdata52]# chown root:netdata /usr/libexec/netdata/plugins.d/cgroup-network-helper.sh 
---
> [/root/netdata52]# chown root /usr/libexec/netdata/plugins.d/cgroup-network-helper.sh 
177c177
< [/root/netdata52]# tar -xf /tmp/netdata-go-BmvTXc/config.tar.gz -C /usr/lib/netdata/conf.d/ 
---
> [/root/netdata52]# tar -xf /tmp/netdata-go-nXVIre/config.tar.gz -C /usr/lib/netdata/conf.d/ 
183c183
< [/root/netdata52]# mv /tmp/netdata-go-BmvTXc/go.d.plugin-v0.2.0.linux-amd64 /usr/libexec/netdata/plugins.d/go.d.plugin 
---
> [/root/netdata52]# mv /tmp/netdata-go-nXVIre/go.d.plugin-v0.2.0.linux-amd64 /usr/libexec/netdata/plugins.d/go.d.plugin 
191a192,197
> [/root/netdata52]# chmod a+rX /usr/libexec 
>  OK   
> 
> [/root/netdata52]# chmod a+rX /usr/share/netdata 
>  OK   
> 
211,213c217,218
< Ooops! it seems netdata is not started.
< [/root/netdata52]# /usr/bin/systemctl start netdata 
<  OK   
---
> OK. NetData Started!
> 
215c220
<  OK  netdata started! 
---
> -------------------------------------------------------------------------------
218c223
[...]

PR #5736, opened 2019-03-29
netdata/packaging/installer: netdata-installer.sh script (and subscripts) refactoring wrap-up

paulkatsoulakis:
##### Summary
PR from issue #5705 that is the wrap up of PR #5505 from Pawel.
It only includes some small changes on top of the ones that Pawel had already fixed from the feedback of #5505.

Please follow up the testing that took place on #5705 and give any further feedback on other issues you spot

##### Component Name
netdata/packaging/installer

##### Additional Information
None


paulkatsoulakis:
issue #5705 has detailed the steps of testing that took place prior to this PR.
I think we did quite extensive checks and in the end we did good because all the issues raised by the various logs were all for fixing.

Merging

Issue #5713, opened 2019-03-26
netdata/packaging/build: Build getting broken sporadically at times in py2-psycopg2 package


paulkatsoulakis:
##### Bug report summary
We have sporadic build failures at py2-psycopg2 . 

Here's the end of the log with the details of the error

```
(41/54) Replacing gdbm (1.13-r1 -> 1.13-r1)
(42/54) Replacing sqlite-libs (3.26.0-r3 -> 3.26.0-r3)
(43/54) Replacing python2 (2.7.15-r3 -> 2.7.15-r3)
(44/54) Replacing mariadb-connector-c (3.0.8-r0 -> 3.0.8-r0)
(45/54) Replacing py-mysqldb (1.2.5-r1 -> 1.2.5-r1)
(46/54) Replacing py-psycopg2 (2.7.5-r0 -> 2.7.5-r0)
(47/54) Replacing py-egenix-mx-base (3.2.9-r0 -> 3.2.9-r0)
(48/54) Replacing db (5.3.28-r1 -> 5.3.28-r1)
(49/54) Replacing libsasl (2.1.27-r1 -> 2.1.27-r1)
(50/54) Replacing libldap (2.4.47-r2 -> 2.4.47-r2)
(51/54) Replacing libpq (11.1-r0 -> 11.1-r0)
(52/54) Replacing py2-psycopg2 (2.7.5-r0 -> 2.7.5-r0)
[91mERROR: py2-psycopg2-2.7.5-r0: failed to rename usr/lib/python2.7/site-packages/.apk.e6a4129270b3b99b221d1dbc3289eb2ec9dd67d2f0113746 to usr/lib/python2.7/site-packages/psycopg2-2.7.5-py2.7.egg-info.
[0m(53/54) Replacing py-yaml (4.1-r0 -> 4.1-r0)
(54/54) Replacing py2-yaml (4.1-r0 -> 4.1-r0)
Executing busybox-1.29.3-r10.trigger
Executing ca-certificates-20190108-r0.trigger
1 error; 99 MiB in 60 packages
fetch http://mirrors.polyverse.io/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
fetch http://mirrors.polyverse.io/alpine/v3.9/community/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
[...]

PR #5751, opened 2019-03-31
Fix the Polyverse reinstall that caused APK broken state

archisgore:
##### Summary

This fix is for issue: https://github.com/netdata/netdata/issues/5713

For a reason still unknown, once `py2-psycopg2` package fails to reinstall due to an upstream APK 
 bug (which we are actively fixing), permanently putting APK into a broken state.

Initially, I had fixed this through a PR: https://github.com/netdata/netdata/commit/2925b27d2a2c2c218ecc88ca5fa6aaecb5422a34

But this had two problems:
1. The less scalable problem was that the fix was targetted to py2-psycopg2`
2. More so, it was discovered that `apk fix` itself only ever works intermittently (https://github.com/netdata/netdata/issues/5713)

This change updates the reinstall block with the latest recommended blob from Polyverse in general (running the install script, and then checking exit code before proceeding to update/reinstall as a separate `if ... fi` block.)

In addition, we have blacklisted `py2-psycopg2` in our repository, thus preventing it from reinstalling, and created a mechanism to ensure we can add more blacklisted packages as we find them in the medium run.

In the long run, we're going to fix APK itself to remedy the situation in general for all Alpine users (the issue also happens with non-polyverse involvement, we just ended up surfacing it due to the in-place reinstall case.)

##### Component Name

Docker Packaging


archisgore:
Build broken. Fixing...

archisgore:
This PR is ready to merge.

packaging/docker/Dockerfile
```diff
@@ -39,11 +39,13 @@ FROM netdata/base:${ARCH}
 
 # Conditional subscribiton to Polyverse's Polymorphic Linux repositories
 RUN if [ "$(uname -m)" == "x86_64" ]; then \
+        apk update && apk upgrade; \
```
paulkatsoulakis: Why `apk update` and `apk upgrade` now is before polyverse at line 42 instead of after?

archisgore: It should be both - before and after. We learned a lot during this testing/repro and using some best practices.

So let's say we want Polymorphic `curl`. If the base container is not updated, then this can happen:

`curl 1.0` -> `polymorphic curl 2.0`

So we're upgrading AND reinstalling all at once. This had a few risks and failure modes we took the opportunity to fix. Instead now what happens is:

`curl 1.0` -> `curl 2.0` (upgrade)
`curl 2.0` -> `polymorphic curl 2.0` (reinstall)

The first upgrade allows any triggers, upgrade scripts, config migrations, etc. to happen cleanly without Polyverse getting in the way.

archisgore: That's an excellent question @paulkatsoulakis. You're absolutely correct that the first upgrade is meant to ensure you're on the latest versions, before we install polymorphic versions.

On your question about switching packages - that is the one thing we guarantee and ensure will always work - our packages show up as drop-in replacements for any upstream packages. Versions, dependencies, etc. The switch is transparent and easy. In fact you see it now - when we removed py2-psycopg2 from our repo, nothing changed on the image, but the error went away.

paulkatsoulakis: cool thanks

PR #5779, opened 2019-04-02
Add preferred disk id pattern

### LINKED ISSUE #4957: support filtering disk naming by id
Following https://github.com/netdata/netdata/pull/3084 and https://github.com/netdata/netdata/issues/3081, I have set

```
[plugin:proc:/proc/diskstats]                                                   
  name disks by id = yes
```

This has the effect of setting my disk names to names which look like

`wwn-0x5000c5xxxxxxxxxx`

instead of (what I would prefer)

`ata-CT250MX500SSD1_xxxxxxxxxx` or `md-name-any:PV1`

Can a prefix filter be added to the `name disks by id` option to filter for the desired name format?

vlvkobal:
##### Summary
Closes #4957

##### Component Name
proc plugin, diskstats module

### DIFF
--- a/collectors/proc.plugin/proc_diskstats.c
+++ b/collectors/proc.plugin/proc_diskstats.c
@@ -418,7 +418,7 @@ failed:
     if(!found)
         result[0] = '\0';
     else if(!preferred)
-        strncpyz(result, first_result, result_size - 1);
+        strncpyz(result, first_result, result_size);
 
     freez(first_result);
 
--- a/collectors/proc.plugin/proc_diskstats.c
+++ b/collectors/proc.plugin/proc_diskstats.c
@@ -529,7 +529,7 @@ static void get_disk_config(struct disk *d) {
     }
 }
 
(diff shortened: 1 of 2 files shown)


ilyam8:
i tested functionality, it works correctly

collectors/proc.plugin/proc_diskstats.c
```diff
@@ -394,8 +396,12 @@ static inline int get_disk_name_from_path(const char *path, char *result, size_t
             //info("DEVICE-MAPPER ('%s', %lu:%lu): filename '%s' matches.", disk, major, minor, filename);
 
             snprintfz(result, result_size - 1, "%s%s%s", (prefix)?prefix:"", (prefix)?"_":"", de->d_name);
+
             found = 1;
-            break;
+
+            if(simple_pattern_matches(preferred_ids, result)) {
+                break;
```
ilyam8: hold on, what if simple pattern never matches, where is fallback?

PR #5778, opened 2019-04-02
netdata/packaging/ci: Add kickstart execution integrity tests

paulkatsoulakis:
##### Summary
There are times that we fall into the trap of breaking our installer at the kickstart side.
Add steps that validate kickstart at travis build to mitigate the risk of breaking the script.

Fixes #5769 

Note: This will be also included on the pipeline redesign and will probably extended to a more sophisticated test than just expect an exit 0 from the script

##### Component Name
netdata/packaging/kickstart
netdata/packaging/ci

##### Additional Information
None

PR #5780, opened 2019-04-02
netdata/packaging/ci: fix travis syntax

paulkatsoulakis:
##### Summary
Introduced from PR #5778 -- found after all why Travis didn't respond.
Wasn't really intuitive the way it failed, linter showed the syntax which means i need to make some adjustments on my process to avoid ending up in the same loop hole again.

##### Component Name
netdata/packaging/ci

##### Additional Information
None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,9 @@ jobs:
   - name: check checksums for kickstart files
     script: ./tests/installer/checksums.sh
     env: LOCAL_ONLY="true"
-  - name: Kickstart: Script execution integrity
+  - name: Kickstart Script execution integrity
     script: docker run -it -v "${PWD}:/code:rw" -w /code "netdata/os-test:centos7" packaging/installer/kickstart.sh --dont-wait --dont-start-it --install /tmp
-  - name: Kickstart-static64: Script execution integrity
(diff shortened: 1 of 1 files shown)


paulkatsoulakis:
Had to force merge this change, as part of the previous PR is it broke mission critical functionality

paulkatsoulakis:
I have added the following hook to my repo, to avoid any chance of pushing travis pipeline broken.
At least broken based on travis linter.

What happened was travis collapses when you push a broken yml file and it doesn't present anything to the user. The only place i could spot a problem was the linter on my laptop, thus the pre-hook below to protect myself

```
$ cat .git/hooks/pre-commit        
#!/bin/sh
#
# Prevent the developer from committing a broken travis pipeline config.
# Travis will not warn you if you break the config, it will collapse instead and freeze all operations
#
# Author: Pavlos Emm. Katsoulakis (paul@netdata.cloud)

git diff --quiet --cached .travis.yml
if [ $? -eq 1 ]; then
        travis lint .travis.yml | grep -v "unexpected key stages, dropping" | grep -v "unexpected key jobs, dropping" | grep '^\[x\]'
        exit 1
else
        echo "Travis pipeline was not modified, check is skipped"
fi
```

PR #5792, opened 2019-04-03
Support configurable syslog facilities

### LINKED ISSUE #5717: How can i set syslog facility for netdata daemon?
According to netdata doc, netdata daemon can send log to syslog, and it works fine. But how can i set the syslog facility? Which parameter name should i use? I tried to search 'facility' in github repo, no parameter found.

```
access log = syslog
error log = syslog
debug log = syslog
```

thiagoftsm:
##### Summary

Fixes #5717 

In the link https://github.com/netdata/netdata/issues/5717 there was a request for the system admin to select the facility that he wanna have in syslog. In this commit I am creating two functions and I am also changing two to solve the described issue. 

The function log_facility_id is the unique that is really necessary to fix the problem. When I created it I used all the facilities that are listed in the operate system headers, including the facility that is not available(mark) and it was commented due this or it is considered deprecated(security). This situation happens, because I am considering possible mistakes in the configure file. When someone chooses a facility that is not available, I am starting the syslog with the default that was present in the log.c previously.

The function log_facility_name was created to keep the same pattern that I found inside the file web/server/web_server.c . This function is not necessary to fix the problem, but it helps with possible future features.

The function syslog_init now has an argument, the argument is an integer that represents the facility that will be set.

Finally I had to append few lines of code inside the function open_log_file. I am considering that the config file will have a configuration variable "facility log" for the user to set the facility that is wished. I called three functions in the same line, instead to create variables to store the results, because the return of the called function is direct the input for the new function, so I am considering that there is not any necessity to extract results from the registers to put in memory addresses.

##### Component Name
I am changing the file libnetdata/log/log.c

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -355,6 +355,10 @@ void log_init(void) {
     snprintfz(filename, FILENAME_MAX, "%s/access.log", netdata_configured_log_dir);
     stdaccess_filename = config_get(CONFIG_SECTION_GLOBAL, "access log", filename);
 
+	char deffacility[8];
+	snprintfz(deffacility,7,"%s","daemon");
+	facility_log = config_get(CONFIG_SECTION_GLOBAL, "facility log",  deffacility);
+
     error_log_throttle_period = config_get_number(CONFIG_SECTION_GLOBAL, "errors flood protection period", error_log_throttle_period);
     error_log_errors_per_period = (unsigned long)config_get_number(CONFIG_SECTION_GLOBAL, "errors to trigger flood protection", (long long int)error_log_errors_per_period);
     error_log_errors_per_period_backup = error_log_errors_per_period;
--- a/libnetdata/log/log.c
+++ b/libnetdata/log/log.c
@@ -19,11 +19,300 @@ const char *stdaccess_filename = NULL;
 const char *stderr_filename = NULL;
 const char *stdout_filename = NULL;
 
-void syslog_init(void) {
+// ----------------------------------------------------------------------------
+// Log facility(https://tools.ietf.org/html/rfc5424)
+//
+// The facilities accepted in the Netdata are in according with the following
+// header files for their respective operate system:
+// 	sys/syslog.h (Linux )
+// 	sys/sys/syslog.h (FreeBSD)
+// 	bsd/sys/syslog.h (darwin-xnu)
+
+int log_facility_id(char *facility_name)
+{
+	if ( !(strcmp(facility_name,"auth") ) )
+	{
+		return LOG_AUTH;
+	}
+	else if  ( !(strcmp(facility_name,"authpriv") ) )
+	{
+		return LOG_AUTHPRIV;
+	}
+#ifdef __FreeBSD__
+	else if ( !(strcmp(facility_name,"console") ) )
+	{
+		return LOG_CONSOLE;
+	}
+#endif
+	else if ( !(strcmp(facility_name,"cron") ) )
+	{
+		return LOG_CRON;
+	}
+	else if ( !(strcmp(facility_name,"daemon") ) )
+	{
+		return LOG_DAEMON;
+	}
+	else if ( !(strcmp(facility_name,"ftp") ) )
+	{
+		return LOG_FTP;
+	}
+#ifdef __APPLE__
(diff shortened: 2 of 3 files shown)


thiagoftsm:
Initially I compiled with the traditional steps(./configure && make && make install ), so I could not identify the problem that happens when we use script. After to use the script, I had the same problem and I could fix using the following steps:

I changed the libnet/log/log.c adding the variable facility_log and I put an extern reference to it inside the file libnet/log/log.h . After this I moved few lines of code that were inside the function open_log_file to the daemon/main.c, so I am reading the option "facility log" there now.

I accepted all the suggestions given and I wrote the body of the function log_facility_id using hashes. I agree 100% that we could consider this "not a big deal", but when we remove the strcmp we avoid some call instructions and no less important, I could learn more about the source code. I decided to use simple_hash instead simple_uhash, because I arrived in the conclusion it is the correct case here, please can you confirm this?

Finally, to save some byte codes, I commented the function(log_facility_name) that is not been used anywhere.

daemon/main.c
```diff
@@ -355,6 +355,10 @@ void log_init(void) {
     snprintfz(filename, FILENAME_MAX, "%s/access.log", netdata_configured_log_dir);
     stdaccess_filename = config_get(CONFIG_SECTION_GLOBAL, "access log", filename);
 
+	char deffacility[8];
+	snprintfz(deffacility,7,"%s","daemon");
+	facility_log = config_get(CONFIG_SECTION_GLOBAL, "facility log",  deffacility);
```
cakrit: Why not make facility_log int and put the call to log_facility_id here as well? Then you only call that function once.

libnetdata/log/log.c
```diff
@@ -70,7 +487,9 @@ static FILE *open_log_file(int fd, FILE *fp, const char *filename, int *enabled_
     if(!strcmp(filename, "syslog")) {
         filename = "/dev/null";
         devnull = 1;
-        syslog_init();
+
+		syslog_init(log_facility_id(facility_log));
```
cakrit: Indentation

libnetdata/log/log.c
```diff
@@ -70,7 +487,9 @@ static FILE *open_log_file(int fd, FILE *fp, const char *filename, int *enabled_
...
+        //syslog_init();
```
cakrit: Just remove it.

daemon/main.c
thiagoftsm: I do not see any problem to use an integer, but it will be necessary to export log_facility_id. I did not use integer initially, because yesterday when we talked about the libnetdata/log/log.c you said me "Since we only have one config setting for the facility, there's no reason to go through the initialization and call log_facility_id again. Just use a static int,",  after read this I supposed that we had to keep this function static, so I decided to use a pointer to get the configuration and carry it to the log.

cakrit: Let me spell it out. 
- Function `open_log_file` is called at least three times, for access.log, error.log, debug.log and it's also called we reopen the log files (I don't remember exactly where, you can check the usages of this function). 
- Since we unconditionally initialize syslog each time filename="syslog",  we are also making three calls to `log_facility_id`. 
- However, there's a single input and a single output to that function, as we only have a single config parameter to control the facility. So there's no point in calling it again, with the exact same input. 

There are a few ways to prevent this:
- Put a static int ihavealreadydonethis either in open_log_file or in log_facility_id. The first time we enter it's set e.g. to -1, the second to something else. We check its value and based on that, we don't go through the checks again. There are many variations of this solution.
- Don't call `log_facility_id` from `open_log_file` at all. Do instead:
```
facility_log = log_facility_id(config_get(CONFIG_SECTION_GLOBAL, "facility log",  deffacility));
```
and do `syslog_init(facility_log)` in `open_log_file`. Of course then `facility_log` needs to be an int. 

The older comments were a bit vague, so I get why you didn't understand them.

thiagoftsm: Now I could understand the previous request. I found another solution and I decided use it. Inside the function syslog_init we already have an "static int i", so the function openlog is called only one time, so I returned the syslog_init for the previous format and I put the call of the function log_facility_id inside the openlog.

PR #5796, opened 2019-04-04
Add memory free and per process memory usage to nvidia_smi

huww98:
##### Summary

Add memory free
Add per process memory usage

##### Component Name

nvidia_smi

### DIFF
--- a/collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
+++ b/collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
@@ -364,24 +364,24 @@ class Service(SimpleService):
         for idx, root in enumerate(parsed.findall('gpu')):
             gpu = GPU(idx, root)
             data.update(gpu.data())
-            self.updateProcessesMemChart(gpu)
+            self.update_processes_mem_chart(gpu)
 
         return data or None
 
-    def updateProcessesMemChart(self, gpu):
+    def update_processes_mem_chart(self, gpu):
         ps = gpu.processes()
         if not ps:
             return
         chart = self.charts['gpu{0}_{1}'.format(gpu.num, PROCESSES_MEM)]
-        dim_ids = []
+        active_dim_ids = []
         for p in ps:
             dim_id = 'gpu{0}_process_mem_{1}'.format(gpu.num, p['pid'])
-            dim_ids.append(dim_id)
+            active_dim_ids.append(dim_id)
             if dim_id not in chart:
                 chart.add_dimension([dim_id, '{0} {1}'.format(p['pid'], p['process_name'])])
         for dim in chart:
-            if dim.id not in dim_ids:
-                chart.del_dimension(dim.id)
+            if dim.id not in active_dim_ids:
+                chart.del_dimension(dim.id, hide=False)
 
     def check(self):
         if not self.poller.has_smi():
--- a/collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
+++ b/collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
@@ -26,10 +26,11 @@ FAN_SPEED = 'fan_speed'
 GPU_UTIL = 'gpu_utilization'
 MEM_UTIL = 'mem_utilization'
 ENCODER_UTIL = 'encoder_utilization'
-MEM_ALLOCATED = 'mem_allocated'
+MEM_USAGE = 'mem_usage'
 TEMPERATURE = 'temperature'
 CLOCKS = 'clocks'
 POWER = 'power'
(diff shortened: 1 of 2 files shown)


collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
```diff
@@ -325,14 +350,24 @@ def get_data(self):
...
-            data.update(GPU(idx, root).data())
+            gpu = GPU(idx, root)
+            data.update(gpu.data())
+            ps = gpu.processes() or []
+
+            chart = self.charts['gpu{0}_{1}'.format(gpu.num, PROCESSES_MEM)]
+            for p in ps:
+                dim_name = 'gpu{0}_process_mem_{1}'.format(gpu.num, p['pid'])
+                if dim_name not in chart:
+                    chart.add_dimension([dim_name, '{0} {1}'.format(p['pid'], p['process_name'])])
```
ilyam8: Do we need this on every call?

ilyam8:
@huww98 

one more thing, please move poller.start to `get_data`

```python
if not self.poller.is_started:
    self.poller.start()
```

collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
huww98: I think so. Because processes can be created at any time. Every `get_data` may contains newly created processes

ilyam8: Hm, then we need to delete stale processes from the chart. We can add it later tho. Do you know how many processes can be created during lifetime? I am trying to understand is cleanup is critical or not.

huww98: Sorry for the delay. I'm using this to monitor GPUs running machine learning training processes. Typically, each GPU will run 1-4 processes, and each process will run for several hours. I'm going to add this function tomorrow.

ilyam8: use https://github.com/netdata/netdata/pull/5795

huww98:
`del_dimension` works fine. But it can be more appropriate to only obsolete but not hide these dimensions. Since it's not implemented yet, we can improve this in the future?

ilyam8:
@huww98 

> Since it's not implemented yet, we can improve this in the future?
> But it can be more appropriate to only obsolete but not hide these dimensions.

we can change 

https://github.com/netdata/netdata/blob/d9d41e6b9e23db9aa48481dd7efa4c317273c9fb/collectors/python.d.plugin/python_modules/bases/charts.py#L203

to

```python
def del_dimension(self, dimension_id, hide=True):

```

what do you think?

collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
```diff
@@ -320,22 +345,44 @@ def __init__(self, configuration=None, name=None):
...
         data = dict()
         for idx, root in enumerate(parsed.findall('gpu')):
-            data.update(GPU(idx, root).data())
+            gpu = GPU(idx, root)
+            data.update(gpu.data())
+            self.updateProcessesMemChart(gpu)
 
         return data or None
 
+    def updateProcessesMemChart(self, gpu):
```
ilyam8: use snake case please

collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
```diff
@@ -320,22 +345,44 @@ def __init__(self, configuration=None, name=None):
...
+        ps = gpu.processes()
+        if not ps:
+            return
+        chart = self.charts['gpu{0}_{1}'.format(gpu.num, PROCESSES_MEM)]
+        dim_ids = []
```
ilyam8: change name to active_dim_ids

ilyam8:
@huww98 
do these changes https://github.com/netdata/netdata/pull/5796#issuecomment-480167846

and use `hide=False`

huww98:
Works well. Your solution is good, I'm thinking this overly complicated.

2019-04-05 15-21-16 的屏幕截图

ilyam8:
>  I'm thinking this overly complicated.

what do you mean? what is complicated?

ilyam8:
@huww98 
i am gonne merge the PR, are you done?

huww98:
I mean i was thinking this overly complicated before. I was thinking that we need to mark old dimensions as obsolete then delete them after an hour. Now I realize it is not necessary. My English is not very good. Sorry for that.

> i am gonne merge the PR, are you done?

And yes, I'm done

ilyam8:
Your English is ok, man :+1:

PR #5799, opened 2019-04-04
[WIP] Check disks against their names

### LINKED ISSUE #5744: netdata/collectors/proc.plugin: wrong output for nvme disk stats, nvme disks not autodiscovered
##### Bug report summary

Netdata has issues with collecting stats from nvme drives on one of my servers. On a system with 8 nvme drives, only one drive gets discovered, and even for this drive, the values produced are wrong (read/write rates are steady 8TB/s, utilization is shown to be 1 billion percent, the backlog has value of 28000 minutes, etc.). Charts for all other disks on the system are okay.

EDIT: I have the same issue on a second server, although they are identical - same OS, same HW config, so nothing really surprising here.

See the attached gist for the list of drives and contents of `/proc/diskstats`:

https://gist.github.com/kralewitz/7ab7da4f3f2ad8163d598b7c0794858d

Steps I have tried:

1) compile netdata with debug enabled and flags `0x0000000000010000 (D_PROCFILE)`

After running in debug mode, `debug.log` has nothing interesting, but `error.log` is getting spammed with this:

```
2019-03-29 18:37:05: netdata INFO  : PLUGIN[proc] : INTERNAL CHECK: host '<snip>', chart 'disk_util.nvme3c33n1' is collected in the same interpolation point: short by 335031 microseconds
```
Repeated over and over again for all charts of the nvme disk in question.

2) Adding configuration for other nvme disks to `netdata.conf` by hand, e.g.:

`
[plugin:proc:/proc/diskstats:nvme1c33n1]
enable = yes
...
`
However, the dashboard still shows only charts for the one nvme disk (`nvme3c33n1`) that was autodiscovered before. There's no mention of the other disk i configured manually in `error.log`

Any suggestions on what to try next/how to debug this further?

##### OS / Environment

Debian 9.8

`Linux <snip> 4.19.0-0.bpo.2-amd64 #1 SMP Debian 4.19.16-1~bpo9+1 (2019-02-07) x86_64 GNU/Linux`

Both netdata v1.13.0, and the latest netdata v1.13.0-50-gb9b96490 compiled in debug mode

##### Component Name

netdata/collectors/proc.plugin

##### Steps To Reproduce

Not sure

##### Expected behavior

Netdata correctly discovers all nvme disks in use and produces meaningful charts for them :)

vlvkobal:
##### Summary
With modern Linux kernels using major and minor numbers is not mandatory for drivers anymore. Checks against the device name in addition to major and minor numbers were added for identifying disks correctly in the diskstats module.

Fixes #5744

##### Component Name
proc plugin

### DIFF
--- a/collectors/proc.plugin/proc_diskstats.c
+++ b/collectors/proc.plugin/proc_diskstats.c
@@ -17,6 +17,7 @@
 static struct disk {
     char *disk;             // the name of the disk (sda, sdb, etc, after being looked up)
     char *device;           // the device of the disk (before being looked up)
+    uint32_t hash;
     unsigned long major;
     unsigned long minor;
     int sector_size;
@@ -543,13 +544,14 @@ static struct disk *get_disk(unsigned long major, unsigned long minor, char *dis
     static struct mountinfo *disk_mountinfo_root = NULL;
 
     struct disk *d;
+    uint32_t hash = simple_hash(disk);
 
     // search for it in our RAM list.
     // this is sequential, but since we just walk through
     // and the number of disks / partitions in a system
     // should not be that many, it should be acceptable
     for(d = disk_root; d ; d = d->next)
-        if(unlikely(d->major == major && d->minor == minor))
+        if(unlikely(d->major == major && d->minor == minor && d->hash == hash && !strcmp(d->device, disk)))
             return d;
 
     // not found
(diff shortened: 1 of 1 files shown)


vlvkobal:
As far as incorrect major and minor numbers for nvme disks is a bug for a specific kernel (4.19) and we can do nothing sane about it, I'm closing the PR.

Issue #5821, opened 2019-04-08
macOS: python.d.plugin crashes on start


brianoconnorgaming:
##### Bug report summary
Ive installed netdata on a fresh copy of macOS twice with the same results. As soon as installation is complete and the system reboots to start the services Python quits unexpectedly. If I then try to reboot or to manually restart the services myself I get the same result. Attached is the full Python crash report.

##### OS / Environment
Model: MacPro5,1, BootROM MP51.0085.B00
Processors: 8 processors, Quad-Core Intel Xeon, 2.26 GHz, 64 GB, SMC 1.39f5
Memory: 64 GB 1066 MHz DDR3
Graphics: ATI Radeon HD 5870 1024 MB
OS: macOS Mojave 10.14.4

#### Netdata version (ouput of `netdata -V`)
1.13.0 I believe, netdata -V doesnt work on macOS

##### Steps To Reproduce
on startup of service or restart of service python crashes immediately every time. have gone as far as a fresh install of macOS and fresh install of Netdata
##### Expected behavior
services to run as expected 
[Python Crash Report as Related to NetData.zip](https://github.com/netdata/netdata/files/3056257/Python.Crash.Report.as.Related.to.NetData.zip)

ilyam8:
Found it

Similar problem
https://github.com/ansible/ansible/issues/32499

> This is apparently due to some new security changes made in High Sierra that are breaking lots of Python things that use fork(). Rumor has it that adding
export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES before your Ansible run should clear it up. The code that's causing issues is well below Ansible in the stack.


So exporting `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` will fix the problem

brianoconnorgaming:
Would you mind elaborating on that?

ilyam8:
@brianoconnorgaming added the fix

details about the problem in the issue i posted above

brianoconnorgaming:
Thanks but that doesn’t really help ME.  Not to say it’s not useful, I just don’t fully understand what the “fix” is. Thanks

ilyam8:
The fix will be in next release, it works, i tested it on `Mojave 10.14.4 (18E226)`

python.d.plugin: `/usr/local/Cellar/netdata/1.13.0/libexec/netdata/plugins.d/python.d.plugin`

And there is no bug for python3. So there several ways how to get it working right now:
 - install python3 and set python3 as first option in python.d.plugin
[...]

PR #5838, opened 2019-04-11
python.d.plugin: py2 fix crash on macos

ilyam8:
##### Summary

Fixes: #5821

similar issue: https://github.com/ansible/ansible/issues/32499

> This is apparently due to some new security changes made in High Sierra that are breaking lots of Python things that use fork(). Rumor has it that adding
export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES before your Ansible run should clear it up. The code that's causing issues is well below Ansible in the stack.

^^

multiprocessing broken on macOS
python.d.plugin uses multiprocessing package

Fix:
 - export `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` in case of MacOS

##### Component Name

[/collectors/python.d.plugin](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin)

### DIFF
--- a/collectors/python.d.plugin/python.d.plugin.in
+++ b/collectors/python.d.plugin/python.d.plugin.in
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
-'''':; exec "$(command -v python || command -v python3 || command -v python2 ||
+'''':;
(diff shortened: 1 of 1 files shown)


ilyam8:
py2 problems btw

PR #5839, opened 2019-04-11
Adjust functions inside daemon/main.c

thiagoftsm:
##### Summary
While I am working in the issues 4717 and 2553, I could observe that there are 4 functions inside the file daemon/main.c that are being exported, but they are not present in any header, and when I studied the body of these functions, I could not find any motive to export them to. So I am suggesting to change them to static.

##### Component Name
remove_option(int opt_index, int *argc, char **argv) : this function is responsible to adjust the values of argv, that is a variable used only in the main, so my suggestion is to do this function static. I know there is a TODO saying to remove it, but while it is not, I think it can be static.

void set_global_environment() : Considering that the environment variables must be present before the daemon runs, I do not see any possibility to call this function from other file.

void log_init(void): I have an argument analogous the previous, but here it is for global variables and not for environment variables.
In this function I changed the tabulation too, I saw that except the lines that I append before, all the tabulation was given with 4 spaces instead a tab, so to keep the pattern I changed to 4 spaces.

void web_server_threading_selection(void): Considering that when we start the web server it will need to have this behavior already defined, I think this function is static too.

void web_server_config_options(void): The argument to consider this function static is exactly the same of the previous.


thiagoftsm:
I discovered that we had to remove the previous commit when I did this, I already fixed this now!

mfundul:
My opinion is that this kind of small changes to the code will create some kind of git log pollution. What I mean is that somebody looking at git will struggle to find what he's looking for while trying to filter out small changes that don't really add much value.

Another objection I have is that some functions have been written by @ktsaou him having in mind that he would like to use them in the future, and this may lead to undoing some commits adding even more git log pollution.

Considering that we will have to refactor much of the code anyway, I prefer to postpone this kind of changes until we reach the point we have to touch the code for functional reasons, and then it will be the perfect opportunity to clean it up. Right now, I think it's a little premature.

Issue #5783, opened 2019-04-03
feature: LDAPS support for openldap-collector


tctovsli:
##### Feature idea summary
The current OpenLDAP collector doesn't work with LDAPS.
You are able to change what port to connect to, but that isn't enough.
If the feature is already there, it isn't documented in the current collector.

##### Expected behavior
Successfully collect data from OpenLDAP over TLS (LDAPS)

ilyam8:
i did a quick search on google, some related links: [link1](https://bneijt.nl/blog/post/connecting-to-ldaps-with-self-signed-cert-using-python/), [link2](https://stackoverflow.com/questions/7716562/pythonldapssl)

@tctovsli 

>  but that isn't enough

Do you know what changes we need?

I read through the link i posted above, i noticed:
 - `ldap.initialize('ldaps://ldap.example.com')`: `ldaps:` instead of `ldap:`
 - `l.simple_bind_s` instead of `l.simple_bind`
 - additional options `ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_NEVER)`
 - `l.unbind_s()` instead of `l.unbind()`

Maybe i missed something.

@ekartsonakis any chance you could help with this?

ekartsonakis:
Hi, Give me some time please to have a look on this. This feature must be implemented.

ekartsonakis:
Ok I opened a pull request. https://github.com/netdata/netdata/pull/5827

ilyam8:
@ekartsonakis 
the question, are you sure we don't need to use
 - `simple_bind_s` instead of `simple_bind`
 - `unbind_s` instead of `unbind`

in case of tls?

PR #5859, opened 2019-04-12
Openldap tls support rebased

ekartsonakis:
*this is a second pull request of [this](https://github.com/netdata/netdata/pull/5827), with the right author commits

Summary
Added support to make secure LDAP TLS connections. User can also choose to skip certificate check in case self signed ones are used.

Fixes: #5783


ilyam8:
thank you @ekartsonakis

PR #5865, opened 2019-04-15
Fix Raw_Read_Error_Rate and Seek_Error_Rate

matiasdelellis:
My Western Digital Red report raw values and netdata shows that I had 200 reading errors when it is incorrect..

ID# ATTRIBUTE_NAME          FLAG     VALUE WORST THRESH TYPE      UPDATED  WHEN_FAILED RAW_VALUE
  1 Raw_Read_Error_Rate     0x002f   200   200   051    Pre-fail  Always       -       0
  7 Seek_Error_Rate         0x002e   100   253   000    Old_age   Always       -       0

With this approach (the same as the issue #4908 for attr 194), if I had more than 200 errors, I would report 200, but since any value greater than 0 is important, I guess this is acceptable.

### DIFF
--- a/collectors/python.d.plugin/smartd_log/smartd_log.chart.py
+++ b/collectors/python.d.plugin/smartd_log/smartd_log.chart.py
@@ -440,6 +440,16 @@ class AtaNormalized(BaseAtaSmartAttribute):
         return self.normalized_value
 
 
+class Ata1(BaseAtaSmartAttribute):
+    def value(self):
+        return min(int(self.normalized_value), int(self.raw_value))
+
+
+class Ata7(BaseAtaSmartAttribute):
+    def value(self):
+        return min(int(self.normalized_value), int(self.raw_value))
+
+
 class Ata9(BaseAtaSmartAttribute):
     def value(self):
         value = int(self.raw_value)
@@ -475,6 +485,10 @@ class SCSIRaw(BaseSCSISmartAttribute):
 def ata_attribute_factory(value):
     name = value[0]
 
+    if name == ATTR1:
+        return Ata1(*value)
+    if name == ATTR7:
+        return Ata7(*value)
     if name == ATTR9:
         return Ata9(*value)
     elif name == ATTR190:
@@ -482,7 +496,6 @@ def ata_attribute_factory(value):
     elif name == ATTR194:
         return Ata194(*value)
     elif name in [
-        ATTR1,
         ATTR7,
         ATTR202,
         ATTR206,


matiasdelellis:
The ideal would be something more configurable, where to select how to interpret each value..

Something like this?
```
WDC_WD10EFRX_68FYTN0-WD_WCC4J7CCC9YN.ata.csv:
   raw_values: '1, 7'
```

ilyam8:
@Ferroin 

@matiasdelellis i think min/max for 1 and 7 won't work for all cases.

Attr1 and attr7 both are vendor specific.

`KINGSTON_SHFS37A120G-50026B7266003D93.ata.csv`: 1;120;0
`ST500DM002_1BD142-Z3TW9E8F.ata.csv`: `1;108;16381976`

i need normilezed for both.

https://github.com/netdata/netdata/issues/4285#issuecomment-426020233

> Attribute 7 (Seek Error Rate) has mostly useless raw values. However, the cooked value is useful as almost all vendors use it correctly. Pretty much, if the cooked value is less than 100, there is a non-zero number of seek errors, which is usually indicative of significant mechanical wear on the drive.

ilyam8:
something like this
https://github.com/netdata/netdata/pull/5865#issuecomment-483085186
could be useful, we need to think about it

ilyam8:
We even could have raw mode  - collect all data as is without processing and chart both raw and normalized values. I think it would be useful in some cases like collect and export to some place and process there. Maybe i am wrong.

What do you think? @matiasdelellis @Ferroin

ilyam8:
i looked at 1 and 7 attributes on my disks and @modem7 disks from https://github.com/netdata/netdata/issues/5919#issuecomment-487092700

I see raw value in smartctl output, but the value for one of my disks is weird.

Somehow `1;108;0;` is decoded to `8528`

| Tables        | Attr1 Log | Attr1 Smartctl|Attr7 Log  | Attr7 Smartctl  |
| ------------- |:-------------:| -----:|:-------------:| -----:|
| ST500DM002_1BD142-Z3TW9E8F     | 1;108;0; | 8528 |7;78;17417470315; | 17417470315 |

From https://en.wikipedia.org/wiki/S.M.A.R.T.

Attr 1
> The raw value has different structure for different vendors and is often not meaningful as a decimal number.

Attr 7
> The raw value has different structure for different vendors and is often not meaningful as a decimal number.

@matiasdelellis that is why we are using normalized value for this attribuites. At that is why units is `value` and not `errors`, so 200 doesn't mean you have 200 read errors (this is not clear from the chart i agree)

ilyam8:
I feel like we have some problems with `smartd_log`. We need to rethink it.

We have 3 types of values:
 - normalized
 - raw
 - decoded

Curently `smartd_log` uses all these types as metrics based on some algorithm.

https://github.com/netdata/netdata/blob/adb7026b14fea7bbebee93c8958d92ee9cf1efaf/collectors/python.d.plugin/smartd_log/smartd_log.chart.py#L433-L454

And it is not clear what value type is used and why. 

i am adding wontfix label for this PR because it fixes some cases and breaks other.

matiasdelellis:
Hi,
Sorry for the delay.

> And it is not clear what value type is used and why.

I agree in your comments, and still I don't understand smartd values correctly. This PR, I just try to solve a personal problem, working on a previous 'solution', but it is not the best option.

PR #5893, opened 2019-04-18
NEW: allow additional name argument passing to urllib3 request in UrlService

andvgal:
##### Summary

It solves monitoring of HTTP(S) endpoints which require POST data (allows passing "body"). This approach also adds more flexibility for other purposes.

##### Component Name

Python UrlService

##### Additional Information

N/A


ilyam8:
thanks @andvgal :+1:

PR #5894, opened 2019-04-18
NEW: Energi Core daemon monitoring, suits other Bitcoin forks

andvgal:
##### Summary

Adds monitoring of Energi Core Bitcoin-derived daemon monitoring. It is named after the contributing cryptocurrency team, but targets a wide range of projects.

As there is no strict RPC standardization, there is no 100% guarantee that it will work with other forks of Bitcoin. Although, most of them should be supported and can be used with this plugin.

##### Component Name

New Python collector module.

##### Additional Information

Smoke tested with `make dist`.

Depends on https://github.com/netdata/netdata/pull/5893
Ignore `UrlService` changed here.

### DIFF
--- a/collectors/python.d.plugin/energid/README.md
+++ b/collectors/python.d.plugin/energid/README.md
@@ -9,23 +9,28 @@ with many others forks including bitcoind itself.
 Introduces several new charts:
 
 1. **Blockchain Index**
-    - blocks
-    - headers
+ * blocks
+ * headers
+
 2. **Blockchain Difficulty**
-    - diff
+ * diff
+
 3. **MemPool** in MiB
-    - Max
-    - Usage
-    - TX Size
+ * Max
+ * Usage
+ * TX Size
+
 4. **Secure Memory* in KiB
-    - Total
-    - Locked
-    - Used
+ * Total
+ * Locked
+ * Used
+
 5. **Network**
-    - Connections
+ * Connections
+
 6. **UTXO** (Unspent Transaction Output)
-    - UTXO
-    - Xfers (related transactions)
+ * UTXO
+ * Xfers (related transactions)
 
 Configuration is needed, disabled by default.
 
--- a/collectors/python.d.plugin/energid/README.md
+++ b/collectors/python.d.plugin/energid/README.md
@@ -9,23 +9,23 @@ with many others forks including bitcoind itself.
 Introduces several new charts:
 
 1.  **Blockchain Index**
-    * blocks
-    * headers
+    *   blocks
+    *   headers
 2.  **Blockchain Difficulty**
-    * diff
+    *   diff
 3.  **MemPool** in MiB
-    * Max
-    * Usage
-    * TX Size
+    *   Max
+    *   Usage
+    *   TX Size
 4.  **Secure Memory** in KiB
-    * Total
-    * Locked
-    * Used
+    *   Total
+    *   Locked
+    *   Used
 5.  **Network**
-    * Connections
+    *   Connections
 6.  **UTXO** (Unspent Transaction Output)
-    * UTXO
-    * Xfers (related transactions)
+    *   UTXO
+    *   Xfers (related transactions)
 
 Configuration is needed in most cases of secure deployment to specify RPC credentials.
 However, Energi, Bitcoin and Dash daemons are checked on startup by default.
--- a/collectors/python.d.plugin/energid/README.md
+++ b/collectors/python.d.plugin/energid/README.md
@@ -1,0 +1,32 @@
+# energid
+
+A collector for [Energi Core](https://github.com/energicryptocurrency/energi) node instance monitoring.
+
+As Energi Core Gen 1 & 2 are based on the original Bitcoin code and
+supports very similar JSON RPC, there is quite high chance the module works
+with many others forks including bitcoind itself.
+
+Introduces several new charts:
+
+1. **Blockchain Index**
+    - blocks
+    - headers
+2. **Blockchain Difficulty**
+    - diff
+3. **MemPool** in MiB
+    - Max
+    - Usage
+    - TX Size
+4. **Secure Memory* in KiB
+    - Total
+    - Locked
+    - Used
+5. **Network**
+    - Connections
+6. **UTXO** (Unspent Transaction Output)
+    - UTXO
+    - Xfers (related transactions)
+
+Configuration is needed, disabled by default.
+
+[![analytics](https://www.google-analytics.com/collect?v=1&aip=1&t=pageview&_s=1&ds=github&dr=https%3A%2F%2Fgithub.com%2Fnetdata%2Fnetdata&dl=https%3A%2F%2Fmy-netdata.io%2Fgithub%2Fcollectors%2Fpython.d.plugin%2Fenergid%2FREADME&_u=MAC~&cid=5792dfd7-8dc4-476b-af31-da2fdb9f93d2&tid=UA-64295674-3)]()
--- a/collectors/python.d.plugin/energid/README.md
+++ b/collectors/python.d.plugin/energid/README.md
@@ -11,18 +11,23 @@ Introduces several new charts:
 1.  **Blockchain Index**
     * blocks
     * headers
+    
 2.  **Blockchain Difficulty**
     * diff
+    
 3.  **MemPool** in MiB
     * Max
     * Usage
     * TX Size
-4.  **Secure Memory* in KiB
+    
+4.  **Secure Memory** in KiB
     * Total
     * Locked
     * Used
+    
 5.  **Network**
     * Connections
+    
 6.  **UTXO** (Unspent Transaction Output)
     * UTXO
     * Xfers (related transactions)
--- a/collectors/python.d.plugin/energid/README.md
+++ b/collectors/python.d.plugin/energid/README.md
@@ -1,6 +1,7 @@
 # energid
 
-A collector for [Energi Core](https://github.com/energicryptocurrency/energi) node instance monitoring.
+A collector for [Energi Core](https://github.com/energicryptocurrency/energi)
+node instance monitoring.
 
 As Energi Core Gen 1 & 2 are based on the original Bitcoin code and
 supports very similar JSON RPC, there is quite high chance the module works
@@ -9,29 +10,35 @@ with many others forks including bitcoind itself.
 Introduces several new charts:
 
 1.  **Blockchain Index**
-    *   blocks
-    *   headers
-2.  **Blockchain Difficulty**
-    *   diff
-3.  **MemPool** in MiB
-    *   Max
-    *   Usage
-    *   TX Size
-4.  **Secure Memory** in KiB
-    *   Total
-    *   Locked
-    *   Used
-5.  **Network**
-    *   Connections
-6.  **UTXO** (Unspent Transaction Output)
-    *   UTXO
-    *   Xfers (related transactions)
+    * blocks
+    * headers
 
-Configuration is needed in most cases of secure deployment to specify RPC credentials.
-However, Energi, Bitcoin and Dash daemons are checked on startup by default.
+1.  **Blockchain Difficulty**
+    * diff
 
-It may be desired to increase retry count for production use due to possibly long
-daemon startup.
+1.  **MemPool** in MiB
+    * Max
+    * Usage
+    * TX Size
+
+1.  **Secure Memory** in KiB
+    * Total
+    * Locked
+    * Used
+
+1.  **Network**
+    * Connections
+
+1.  **UTXO** (Unspent Transaction Output)
+    * UTXO
+    * Xfers (related transactions)
+
+Configuration is needed in most cases of secure deployment to specify RPC
+credentials. However, Energi, Bitcoin and Dash daemons are checked on
+startup by default.
+
+It may be desired to increase retry count for production use due to possibly
+long daemon startup.
 
 ## Configuration
 
--- a/collectors/python.d.plugin/energid/README.md
+++ b/collectors/python.d.plugin/energid/README.md
@@ -8,29 +8,24 @@ with many others forks including bitcoind itself.
 
 Introduces several new charts:
 
-1. **Blockchain Index**
-  * blocks
-  * headers
-
-2. **Blockchain Difficulty**
-  * diff
-
-3. **MemPool** in MiB
-  * Max
-  * Usage
-  * TX Size
-
-4. **Secure Memory* in KiB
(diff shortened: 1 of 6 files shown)


collectors/python.d.plugin/energid/README.md
```diff
@@ -0,0 +1,37 @@
...
+ * Used
+
+5. **Network**
+ * Connections
+
+6. **UTXO** (Unspent Transaction Output)
+ * UTXO
+ * Xfers (related transactions)
+
+Configuration is needed, disabled by default.
```
Ferroin: There should probably be some example configuration listed here, even if it's just a trivial case.

andvgal:
@Ferroin It is part of energid.conf, but OK. It can be added there.

Maybe you know how to run Codacy Markdown checks locally then?

collectors/python.d.plugin/python.d.conf
```diff
@@ -41,6 +41,7 @@ chrony: no
 # dockerd: yes
 # dovecot: yes
 # elasticsearch: yes
+energi: no
```
Ferroin: If there are no jobs defined for the module (and the included config does not define any), it won't be run by default, even if it would be according to python.d.conf.  Given this, the standard for modules that require local configuration is to list them as enabled by default in python.d.conf (so just like the surrounding lines) so that users only have to add jobs to the configuration for the module itself to enable it.

andvgal: OK

collectors/python.d.plugin/energid/README.md
andvgal: Done

Ferroin:
Regarding the Codacy checks, I'm not sure of any way to run them locally, but in this case the only issue I see is the indentation of the nested list items (Markdown technically requires two or more spaces of indentation for each nesting level), so just fixing that should clear up the checks.  My suggestion would be to just use 4 spaces there instead of 2 like Codacy is suggesting, as that will make it consistent with the other documentation.

andvgal:
Well, it was 4 spaces originally, but Codacy did not like that either.

Tried to copy list style from other modules. Could it be they are just broken and not shown in PR checks...?

andvgal:
That's interesting. Without any changes to the Markdown list, Codacy accepted it from the third time,

Ferroin:
> Well, it was 4 spaces originally, but Codacy did not like that either.

> That's interesting. Without any changes to the Markdown list, Codacy accepted it from the third time,

Sounds like a bug on Codacy's side to me.

collectors/python.d.plugin/energid/energid.chart.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class Service(UrlService):
+    def __init__(self, configuration=None, name=None):
+        UrlService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.host = self.configuration.get('host', '127.0.0.1')
+        self.port = self.configuration.get('port', 9796)
+        self.url = 'http://{host}:{port}'.format(
```
ilyam8: hardcoded `http`

collectors/python.d.plugin/energid/energid.chart.py
```diff
@@ -0,0 +1,132 @@
...
+        'options': [None, 'Network', 'count', 'network', 'energi.network', 'line'],
+        'lines': [
+            ['network.connections', 'Connections', 'absolute'],
+        ],
+    },
+    'utxo': {
+        'options': [None, 'UTXO', 'count', 'UTXO', 'energi.utxo', 'line'],
+        'lines': [
+            ['utxo.count', 'UTXO', 'absolute'],
+            ['utxo.xfers', 'Xfers', 'absolute'],
```
ilyam8: Please use `_` instead of `.`. I recall this is needed for prometheus (exporting).

collectors/python.d.plugin/energid/energid.chart.py
```diff
@@ -0,0 +1,132 @@
...
+                'network.timeoffset' : r['timeoffset'],
+                'network.connections': r['connections'],
+            },
+            'gettxoutsetinfo': lambda r: {
+                'utxo.count' : r['txouts'],
+                'utxo.xfers' : r['transactions'],
+                'utxo.size' : r['disk_size'],
+                'utxo.amount' : r['total_amount'],
+            },
+        }
```
ilyam8: i think better to move these `self.methods` to global vars, we don't need them to be unique for every instance of Service.

collectors/python.d.plugin/energid/energid.chart.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _get_data(self):
+        batch = []
+
+        for i, method in enumerate(self.methods.keys()):
```
ilyam8: `keys()` call not needed

collectors/python.d.plugin/energid/energid.chart.py
```diff
@@ -0,0 +1,132 @@
...
+            batch.append({
+                'version': '1.1',
+                'id': i,
+                'method': method,
+                'params': [],
+            })
+
+        result = self._get_raw_data(body=json.dumps(batch))
+        result = json.loads(result.decode('utf-8'))
```
ilyam8: add `if not result` check after `self._get_raw_data` call

collectors/python.d.plugin/energid/energid.chart.py
ilyam8: i think `url` is enough, it is clear then have 3 different opts

```yaml
  url: `http://1.2.3.4:4321`
```

collectors/python.d.plugin/energid/energid.chart.py
andvgal: There is still dot left in the chart names. It's OK, right?

collectors/python.d.plugin/energid/energid.chart.py
```diff
@@ -0,0 +1,150 @@
...
+        UrlService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.host = self.configuration.get('host', '127.0.0.1')
+        self.port = self.configuration.get('port', 9796)
+        self.url = '{scheme}://{host}:{port}'.format(
+            scheme=self.configuration.get('scheme', 'http'),
+            host=self.host,
+            port=self.port,
+        )
```
ilyam8: Why do we need host, port and scheme? i suggest to use `url` (es is a bad example :smile: )

collectors/python.d.plugin/energid/energid.chart.py
ilyam8: i don't see any dots :thinking:

collectors/python.d.plugin/energid/energid.chart.py
```diff
@@ -0,0 +1,150 @@
...
+
+        if not result:
+            return None
+
+        result = json.loads(result.decode('utf-8'))
+        data = dict()
+
+        for i, (_, handler) in enumerate(METHODS.items()):
+            r = result[i]
+            data.update(handler(r['result']))
```
ilyam8: Is there any possibility to get `KeyError`? If not leave it as is

collectors/python.d.plugin/energid/energid.chart.py
andvgal: In general, only `port` configuration is needed. There is no URL involved in most other related configuration files of the daemon itself and dependent services.

ilyam8: well, so you mean host is always localhost. Still `url` will do. And it is 1 param instead of 3. I suggest to use it.

collectors/python.d.plugin/energid/energid.chart.py
andvgal: It's possible to get an error on startup, but it should be more generic and handled in `UrlService`. There is relative long startup period what requires more retries through.

collectors/python.d.plugin/energid/energid.chart.py
andvgal: In most cases, the host is '127.0.0.1' and not even localhost which may resolve to IPv6 ::1.

andvgal: The main point here, it is NOT HTTP service. It is RPC which is framed with bare simple HTTP.

collectors/python.d.plugin/energid/energid.conf
```diff
@@ -0,0 +1,78 @@
...
+# ----------------------------------------------------------------------
+# AUTO-DETECTION JOBS
+# only one of them will run (they have the same name)
+#
+# local:
+#     host: '127.0.0.1'
+#     port: 9796
+#     user: energi
+#     pass: energi
+#
```
ilyam8: How about to add those 2 jobs from readme as default jobs? If 9796 is default port for energi and 8332 default port for bitcoin it makes sense.

collectors/python.d.plugin/energid/energid.chart.py
andvgal: Technically, it's possible to get an error here on monitored service restart. Is it a problem for netdata?

collectors/python.d.plugin/energid/energid.conf
andvgal: Ehm, yes, it's possible without credentials. It may even work in some default deployments.

collectors/python.d.plugin/energid/energid.chart.py
andvgal: I'll add a check here, just in case.

ilyam8: No it is not a problem.

collectors/python.d.plugin/energid/energid.conf
andvgal: Added Energi, Bitcoin and Dash which are closely related by codebase and should behave similarly.

andvgal:
Well, Codacy markdown behaves strange...

andvgal:
Codacy is passed, but it shows "changes requested" with no obvious items left.

collectors/python.d.plugin/energid/energid.chart.py
andvgal: Done, needs to be marked as resolved.

collectors/python.d.plugin/energid/energid.chart.py
andvgal: Just need to be marked as resolved to by @ilyam8  to make the bot happy.

collectors/python.d.plugin/energid/energid.chart.py
```diff
@@ -0,0 +1,150 @@
...
+        self.header = {
+            'Content-Type': 'application/json',
+        }
+
+    def _get_data(self):
+        batch = []
+
+        for i, method in enumerate(METHODS):
+            batch.append({
+                'version': '1.1',
```
paulkatsoulakis: The version value 1.1 sounds like something that should exist in the context of the collector config, so that it can be set to something different if it ever changes without touching the implementation

collectors/python.d.plugin/python_modules/bases/FrameworkServices/UrlService.py
```diff
@@ -88,13 +88,13 @@ def _build_manager(self, **header_kw):
             self.error('build_manager() error:', str(error))
             return None
 
-    def _get_raw_data(self, url=None, manager=None):
+    def _get_raw_data(self, url=None, manager=None, **kwargs):
```
paulkatsoulakis: Could you elaborate a bit on the necessity of the variable arguments on the core library item?

ilyam8: This is needed to pass arbitrary named arguments to urrlib3.PoolMannager().request method

> Signature: m.request(method, url, fields=None, headers=None, **urlopen_kw)

https://github.com/netdata/netdata/blob/d242e02452ff309334b05a17d6e1061ea3caeac8/collectors/python.d.plugin/python_modules/bases/FrameworkServices/UrlService.py#L119-L127

And it it already added in #5893

ilyam8: i think @andvgal better to rebase and undo these commits otherwise it is confusing

collectors/python.d.plugin/energid/energid.chart.py
ilyam8: Oh, and it would be nice to have a link to the docs how to build body with all these versions and methods. Add it as comment please @andvgal

collectors/python.d.plugin/python_modules/bases/FrameworkServices/UrlService.py
andvgal: Done, rebased

collectors/python.d.plugin/energid/energid.chart.py
andvgal: A comment has been added. The version does not really matters. As the payload is created manually, I believe putting JSON-RPC version into configuration has no use at all and just bloats the code. However, it can be put into a module variable just to eat a bit more memory.

andvgal:
Yeah... It seems Codacy Markdown got conquered. Looks like, the root cause of tons of misleading errors was broken bold notation in one of the list items.

andvgal:
OK, it was absolutely not obvious and many other READMEs fail Markdown check.

I'll submit a separate PR with `remark-lint` installed.

andvgal:
OK...

1. Codacy integration has some flaws - it incorrectly reflects status in GitHub during processing.
2. The correct remark-lint config seems to be here: https://github.com/codacy/codacy-remark-lint/raw/master/.remarkrc.js
    - However, English text processing has to be disabled locally.

collectors/python.d.plugin/energid/energid.chart.py
paulkatsoulakis: I see. If the version does not matter, it would be cleaner if the other end has the version info as optional field and then we can simply omit the field completely. That would simplify things.

Otherwise we can stick with the variable i guess

ilyam8: >  I believe putting JSON-RPC version into configuration has no use at all and just bloats the code

i agree, we can add it later if needed, i think it is ok to have it hardcoded for now

andvgal: Any way, a constant with JSON-RPC version has been added on module level.

ilyam8:
thanks @andvgal !

PR #5898, opened 2019-04-19
NEW: local remark-lint checks and autofix support

andvgal:
##### Summary

After epic battle with Codacy checks in https://github.com/netdata/netdata/pull/5894 , I think local remark-lint check+autofix with proper configuration is essential.

There are lots of README.md files which fail Codacy check, but those are ignored in PRs. Once anyone wants to add changes,  tons of problems appear suddenly.

##### Component Name

Markdown READMEs.

##### Additional Information

The config as declared by Codacy Docker image is added. English text check are disabled for now as it's a different area of hell.

Unfortunately, `remark-stringify` is not exactly compatible with all other `remark` present options. So, auto-fix leads to a conflict with itself. So, it is left as a manual helper.


andvgal:
I dropped the Markdown changes as remark-lint seems to be a bit conflicting with itself. So, PR is only for the ability to run checks for now.

cakrit:
Wow mate, I ran it and got 15905 warnings! 
Let me understand this a bit, because I also hit a wall with the Codacy complaints in one PR and when I went on Codacy to see what's configured there, I didn't see any warnings. I actually installed remark-lint locally to understand the settings, but didn't get as far as you. 

With these additions, is Codacy supposed to read the two files and properly run the linter? 
Or is the idea to do a local installation using this `package.json` and do `npm run lint-md` ourselves?

Another question, since you took the time: In https://github.com/netdata/netdata/pull/5898#issuecomment-484881273 you were trying to get warnings fixed automatically? With so many warnings, I need to look into something that will fix most of our MDs programmatically, so I'd appreciate any hints.

andvgal:
@cakrit 
Well, when I was fighting with Codacy, I got misled many times by the fact that a temporary "green" status was appearing before the warnings were found. That's one thing to note in a "fight".

From what I understood, Codacy has some internal config, but I could grab something from their Docker image (https://github.com/codacy/codacy-remark-lint/) and left link to that. The config does not work AS-IS and requires disabling linguistic stuff as Codacy may have own vocabulary and/or a bit different production config in use.

So, the files are for local usage:
1. `npm install`
2. `npm run lint-md` or `npm run lint-md-path {some_path_here}`

An example auto-fix command is in "fix-md" script. The problem is that remark-cli is not well designed for multiple file processing and the `--output` key must be the last to fix files in place.

Actually, the auto-fix is "almost" good, but there are issues:

1. The main problem of multi-level lists and one space difference between lint itself and built-in formatting engine - this can be fixed by AWK, perhaps.
2. The Google Analytics tracker gets some `<>` garbage. It should be just `analytics`, but not `[analytics]()` - this can be fixed by a script as well.
3. Then, there is a set of common warnings which are not fixed by auto-format: 120-character line limit, missing fenced code block language, etc. - lots of manual work to do.

Fixed READMEs have almost all lines changed.

As Codacy checks only changed file, I believe this PR can be just a handy tool to check the changes locally and not to fight with Codacy through tons of commits in a PR like I've done.

cakrit:
@gmosx need your approval too...

cakrit:
I forgot about this one. Merging it and I'll see when I can get around to using it to fix the warnings. Thanks a lot @andvgal !!!

Issue #5541, opened 2019-03-03
Python modules not running on Ubuntu Server 18.04


Steve8291:
##### Bug report summary
No python modules will display charts on Ubuntu Server 18.04
Both Sensors and Redis modules appear to be working.
Output of `/usr/libexec/netdata/plugins.d/python.d.plugin 1 debug sensors`

> 2019-03-03 13:23:36: python.d INFO: plugin: main: Using python 2
> 2019-03-03 13:23:36: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d.conf'
> 2019-03-03 13:23:36: python.d ERROR: plugin: main: cannot load '/etc/netdata/python.d.conf': [Errno 2] No such file or directory: '/etc/netdata/python.d.conf'. Will try stock version.
> 2019-03-03 13:23:36: python.d DEBUG: plugin: main: loading '/usr/lib/netdata/conf.d/python.d.conf'
> 2019-03-03 13:23:36: python.d DEBUG: plugin: main: module load source: 'sensors' => [OK]
> 2019-03-03 13:23:36: python.d DEBUG: plugin: main: loading '/etc/netdata/python.d/sensors.conf'
> 2019-03-03 13:23:36: python.d ERROR: plugin: main: cannot load '/etc/netdata/python.d/sensors.conf' : [Errno 2] No such file or directory: '/etc/netdata/python.d/sensors.conf'. Will try stock version.
> 2019-03-03 13:23:36: python.d DEBUG: plugin: main: loading '/usr/lib/netdata/conf.d/python.d/sensors.conf'
> 2019-03-03 13:23:36: python.d DEBUG: plugin: main: job initialization: 'sensors sensors' => ['OK']
> 2019-03-03 13:23:36: python.d DEBUG: plugin: main: module status: 'sensors' => [OK] (jobs: 1)
> 2019-03-03 13:23:36: python.d INFO: sensors: sensors: check() => [OK]
> CHART netdata.runtime_sensors '' 'Execution time for sensors' 'ms' 'python.d' netdata.pythond_runtime line 145000 1
> DIMENSION run_time 'run time' absolute 1 1
> 
> 2019-03-03 13:23:36: python.d DEBUG: sensors: sensors: create() => [OK] (charts: 3)
> 2019-03-03 13:23:36: python.d DEBUG: sensors: sensors: started, update frequency: 1
[...]

PR #5911, opened 2019-04-24
SocketService: set socket operation timeout before connecting

ilyam8:
##### Summary
Fixes: #5541

This PR:
 - removes `setblocking` calls

```python
    def setblocking(self, v):
        if v:
            self.settimeout(None)
        else:
            self.settimeout(0.0)
```

 - sets socket timeout before call to `connect`

 - adds separate `connect`, `read` and `write` timeouts. Default is 2 seconds for all.

> socket.settimeout(value)
Set a timeout on blocking socket operations. The value argument can be a nonnegative float expressing seconds, or None. If a float is given, subsequent socket operations will raise a timeout exception if the timeout period value has elapsed before the operation has completed. Setting a timeout of None disables timeouts on socket operations. s.settimeout(0.0) is equivalent to s.setblocking(0); s.settimeout(None) is equivalent to s.setblocking(1).

##### Component Name
[`/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService`](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService.py)

### DIFF
--- a/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService.py
+++ b/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService.py
@@ -16,6 +16,9 @@ else:
 from bases.FrameworkServices.SimpleService import SimpleService
 
 
+DEFAULT_SOCKET_TIMEOUT = 3
+
+
 class SocketService(SimpleService):
     def __init__(self, configuration=None, name=None):
         self._sock = None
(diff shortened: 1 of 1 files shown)


ilyam8:
@Steve8291 

i think this PR is fixes your problem, please test it!

Steve8291:
@ilyam8 
Here is a new link to the paste if you still want to see it.
https://pastebin.com/hmR40rLz
I don't know why the other one didn't work.
I'm also not sure how to test a pull request. Do I just make the code changes directly or do I have to compile netdata on my machine or something like that? Sorry for my lack of knowledge here.
Looks like the PR was committed into "master" So can I just run an update to get it?

ilyam8:
> Looks like the PR was committed into "master" So can I just run an update to get it?

Yes. 

> connecting socket to "::1", port XXXX

Will hang only for 2 seconds there.

Steve8291:
@ilyam8 
Thanks for fixing this. All my charts are coming up much faster now after a restart or a reload of netdata. I'm currently running: v1.14.0-11-nightly that has the patch applied.

ilyam8:
@Steve8291 it seems you have configured `ip6tables` with `DROP` by default.

Steve8291:
@ilyam8 
I that a problem?
I do have all ipv6table chains set to DROP temporarily until I can get around to writing more specific rules. Is it the OUTPUT Chain that needs set to ACCEPT?

ilyam8:
> I that a problem?

Ye, i suggest to chage default policy to `REJECT` for `INPUT` chain.

PR #5914, opened 2019-04-25
Add AWS Kinesis backend

### LINKED ISSUE #5596: AWS Kinesis as backend collector
##### Question summary
I am wondering if there has been any work in pushing netdata streams to an AWS Kinesis data stream.
##### OS / Environment
AWS
##### Component Name
backend or streaming
##### Expected results
Sending JSON blobs to Kinesis data stream with appropriate authentication headers and metadata as data is collected.

vlvkobal:
##### Summary
Add AWS Kinesis Data Streams as a backend using AWS SDK for C++.

Closes #5596

##### Component Name
backends


paulkatsoulakis:
Interesting, make dist validator hit here

irvintim:
Vladimir,  Thank you for creating this.  JSON data is streaming to our Kinesis stream from our test instance, so that is working well.  I'll et it run for a few days, and will toss some additiaonl data at it to see how it scales.

I have a few suggestions and an issue.

ISSUE: 
1. Following the AWS C++ SDK installation instructions to build and install this library, it by default installs the shared library in /usr/local/lib64/.  Netdata can't find the library there out of the box, so I had to do the following before running netdata:
export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib64
Since this is the default location for this library, this should be added to the library path that netdata searches if kinesis is activated?

SUGGESTION:
1. I'd suggest changing "auth key id" to "aws_access_key_id" and "secure key" to "aws_secret_access_key" in the config (and perhaps throughout the code).  This is the standard nomenclature in the aws config files.
2. Along those lines, I'd suggest making the auth key and secret key optional, since the SDK will obtain those credentials via a standard method if they aren't provided on the client class constructor -- see https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/credentials.html


Thanks again!  This is super helpful.

Tim

PR #5920, opened 2019-04-26
Expand dashboard auto-scaling and convertible units

qingkunl:
##### Summary

Add a two more units that allow auto-scaling and conversion: nanoseconds and num.

As a result, e.g. 1000 nanoseconds -> 1 microseconds, 1000000 nanoseconds -> 1 milliseconds, 1000 num -> 1 num (K) etc.

##### Component Name

web/gui/dashboard

##### Additional Information

This PR just expands the existing auto-scaling units (e.g. packets/s, milliseconds, etc.) and follow the corresponding coding style.

### DIFF
--- a/web/gui/dashboard.js
+++ b/web/gui/dashboard.js
@@ -868,6 +868,13 @@ NETDATA.unitsConversion = {
             'TiB': 1024,
             'PiB': 1024 * 1024,
             'EiB': 1024 * 1024 * 1024
+        },
+        'num': {
+            'num': 1,
+            'num (K)': 1000,
+            'num (M)': 1000000,
+            'num (G)': 1000000000,
+            'num (T)': 1000000000000
         }
         /*
         'milliseconds': {
--- a/web/gui/dashboard.js
+++ b/web/gui/dashboard.js
@@ -974,6 +974,72 @@ NETDATA.unitsConversion = {
                         + NETDATA.zeropad(milliseconds);
                 }
             }
+        },
+        'nanoseconds': {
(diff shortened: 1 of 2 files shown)


qingkunl:
CLA signed

qingkunl:
@cakrit Thanks for the feedback. I've updated the OP. However, I cannot find a common chart for illustration. I've got the need for these two units in our monitoring scenario, which I may not be able to share. Please let me know what I can do.

cakrit:
It's ok, we can just mess with the python example collector's units and ranges to make it produce metrics that go from nanosec to sec. @jacekkolasa, I'm talking about `collectors/python.d.plugin/example/example.chart.py`. You could directly edit the installed file you have, probably under `/usr/libexec/netdata/python.d/example.chart.py` and restart netdata. By default it's disabled, so you'd also need to run `/etc/netdata/edit-config python.d.conf` and set `example` to `yes`, in order to see the chart it produces.

qingkunl:
Thanks @cakrit 

FYI, just rebased my forked repo to ToT netdata.

jacekkolasa:
I've tested it in all cases, looks good, is compatible with existing style. In situations when max is 1.1 ms, i think it would be better to switch to microseconds (so instead of 0.20, 0.40 see 200, 400). But that's just my personal preference, and if you as an end user find your formatting well, then i'd trust your choice. :)

Thanks @qingkunl!

@cakrit i think it's ready to merge

cakrit:
Thank you @qingkunl !!!

PR #5956, opened 2019-05-03
SSL implementation for Netdata

thiagoftsm:
##### Summary
I am working to bring HTTPS for Netdata, but I am splitting this job in steps and I am bringing here the first step. With this pull request I am creating two new files inside the libnetdata/socket that will be responsible to encrypt the socket communication. I did not put these files inside the webserver directly because the OpenSSL has one function to start the library for the whole project, so when we start it one time, everybody can use the context that will be created to encrypt what is necessary, not only the webserver. 

##### Component Name
libnetdata/socket/security.c - the file where will be stored all functions related with the encrypted socket.
libnetdata/socket/security.h - the header that is already included inside libnetdata/libnetdata.h
libnetdata/libnetdata.h - changes described in the previous file.
daemon/main.c - Start of the encryptation environment case the key and certified files were given.
##### Additional Information

For we start use the ssl two new web server configuration were created "ssl key" and "ssl certificate", in these options the users will specify the key and the certificate PEM.

In my initial tests I used the following openssl command to generate the self signed certified:

openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365


paulkatsoulakis:
Hi Thiago! Off to a great start, SSL is going to give us yet one more advantage!

Here's some first comments from my side:
1) Check Travis errors,`make dist` does not include the new files in the .tar.gz resuling in a broken source base. Seems that you have missed to update some of the Makefiles, i can have a look for you if you need assistance

2) You need to define which are the package dependencies that are introduced to support SSL, so that we include them on our base images for docker (second error in travis is docker build failing)

cheers!

thiagoftsm:
Hi Paul,

Sorry by the delay to answer you! 

1. This was a great fault that I did not notice, but I am adding it on my test list for the next P.R.
2. I will check this and prepare tomorrow.
3. There is not any problem to move to another directory, but I would suggest we continue using it inside libnetdata/socket. Do you agree?
4. 
a) In the daemon/main.c I am setting default values for the key and certificate, and when the main calls the function security_start_ssl() it uses stat to check the file, case it has error with the files it does not start the SSL library.
b) It will happen the same thing described before.

I tested both cases in different operate systems and the netdata continue running, tomorrow (Sunday) I will start more hard tests with HTTPS and I analyse this again.

Thank you very much by your contributions and tips.

daemon/main.c
```diff
@@ -344,7 +348,20 @@ static const char *verify_required_directory(const char *dir) {
...
 
-void log_init(void) {
+#ifdef ENABLE_HTTPS
+static void security_init(){
+    char filename[FILENAME_MAX + 1];
+    snprintfz(filename, FILENAME_MAX, "%s/key.pem",netdata_configured_user_config_dir);
+    security_key    = config_get(CONFIG_SECTION_WEB, "ssl key",  filename);
+
+    snprintfz(filename, FILENAME_MAX, "%s/cert.pem",netdata_configured_user_config_dir);
+    security_cert    = config_get(CONFIG_SECTION_WEB, "ssl certificate",  filename);
```
ilyam8: I see that `key.pem` and `cert.pem` names are hardcoded. Not a security specialist, but i think there are no mandatory rule for key to have a name `key` and for cert to have a name `cert`. Not a security specialist tho.

Neither dir and name shouldn;t be hardcoded imo.

 @Ferroin what do you think?

thiagoftsm: Due the fact we need to have default values, these are suggestion for the default key and certificate, this happens because we cannot let it as NULL.

ilyam8: oh, these are defaults, ok.

ilyam8: So by default we expecte both `cert.pem` and `key.pem` to be in the `/opt/netdata/etc/netdata` ? 

>  netdata_configured_user_config_dir  = config_get(CONFIG_SECTION_GLOBAL, "config directory",       netdata_configured_user_config_dir);

> # config directory = /opt/netdata/etc/netdata

thiagoftsm: Yes this is expected, I think it is the better place, another possibility would be the netdata home directory.

ilyam8: @Ferroin i am really interested in your opinion oh that

web/server/README.md
```diff
@@ -91,6 +91,28 @@ Netdata supports access lists in `netdata.conf`:
 
 - `allow management from` checks the IPs to allow API management calls. Management via the API is currently supported for [health](../api/health/#health-management-api)
 
+### Enabling TLS/SSL support
+
+Netdata since version 1.14 supports encrypted connections. To enable this option it is necessary to genereate a certificate and a key, case you do not have one it is possible to generate them  with the following command
```
ilyam8: since 1.15 i guess

thiagoftsm: You are correct, I already fixed here and I will update in the next merge.

daemon/main.c
paulkatsoulakis: why not an SSL directory within etc/netdata ?

thiagoftsm: No problem, we can do this. I think this will organize better the files.

thiagoftsm:
The travis does not have the openssl headers so it has the same problems commented before, I removed the WIP because all the initial tests give me the expected results, but it is necessary to say that I do not have the routine to work with Netdata ACL, so I probably did not test all the possible combination. I could not test the Unix socket too, I had problem with permission so I must adjust my personal image.

web/server/README.md
```diff
@@ -91,6 +91,28 @@ Netdata supports access lists in `netdata.conf`:
 
 - `allow management from` checks the IPs to allow API management calls. Management via the API is currently supported for [health](../api/health/#health-management-api)
```
cakrit: We need a long comment here, telling us how to configure the various ports for SSL in the various cases

paulkatsoulakis:
> The travis does not have the openssl headers so it has the same problems commented before, I removed the WIP because all the initial tests give me the expected results, but it is necessary to say that I do not have the routine to work with Netdata ACL, so I probably did not test all the possible combination. I could not test the Unix socket too, I had problem with permission so I must adjust my personal image.

Regarding Travis we had to update the supportive OS specific docker images that we have available, to contain all required packages. I am preparing the new images and i will rerun Travis myself

web/server/README.md
thiagoftsm: When I updated the documentation, it was necessary to move the SSL section from the previous position, because it was behind the ACL section, I moved it and I created the necessary documentation .

web/server/README.md
```diff
@@ -91,15 +115,57 @@ Netdata supports access lists in `netdata.conf`:
...
 
+## SSL access list
+
+When the certificate is defined as one of the configuration to use, the Netdata will enable the TLS and it will redirect all HTTP request to HTTPS, on the other hand, the stream will be text pure until the user define an ACL to it, of course it is also possible to change the server behavior changing the ACL. For both cases we will use the variable `bind to` to change the default behavior.
+
+To change the behavior of master or slave we will use the definition `^SSL` that can have two possible values
```
cakrit: The slave config doesn't belong there. These settings tell the server which ports to listen to and how. You can have a scenario with host A streaming to host B streaming to host C. You can't tell a host how to behave as a client, in the same place where we tell it how to behave as a server. It's much simpler actually for the slaves. You will need to modify what can be added to the `destination` of stream.conf for the slave. i.e. this config param:
```
    # Where is the receiving netdata?
    # A space separated list of:
    #
    #      [PROTOCOL:]HOST[%INTERFACE][:PORT]
    #
    #
    # PROTOCOL  = tcp, udp, or unix (only tcp and unix are supported by masters)
    # HOST      = an IPv4, IPv6 IP, or a hostname, or a unix domain socket path.
    #             IPv6 IPs should be given with brackets [ip:address]
    # INTERFACE = the network interface to use (only for IPv6)
    # PORT      = the port number or service name (/etc/services)
    #
    # This communication is not HTTP (it cannot be proxied by web proxies).
    destination =
```
So we add an option here to tell the slave to communicate with the master via ssl.

I'll be testing with these settings for now, but it needs to change.

thiagoftsm: All right, I understood. I am waiting you finish your tests, when it is done I will change what is necessary, including this documentation.

daemon/main.c
```diff
@@ -1080,6 +1097,12 @@ int main(int argc, char **argv) {
         log_init();
         error_log_limit_unlimited();
 
+        // --------------------------------------------------------------------
+        // get the certificate and start security
+#ifdef ENABLE_HTTPS
+        security_init();
```
cakrit: There are no return values and/or log messages to tell us if it succeeded or not.

web/server/README.md
```diff
@@ -57,6 +57,30 @@ The API requests are serviced as follows:
...
+### Enabling TLS support
+
+Netdata since version 1.16 has the feature to encrypt connections. To enable this option it is necessary to have a certificate and a key, case you do not have one yet, it is possible to generate them  with the following command
+
+$ openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -keyout key.pem -out cert.pem
+
+, it is important to say that the previous command will create a self signed certificate and the browsers will show warnings about it.
+
+To start using HTTPS with Netdata it is only necessary to write the path for your certificate (`ssl certificate`) and key(`ssl key`) inside the `web section`
+
```
cakrit: need to say somewhere here that the netdata user should be able to read both files (explain chown and chmod commands needed)

daemon/main.c
thiagoftsm: The security_init function is responsible to set the global variables and to start the openssl library, but before the version 1.1.0 all the functions except SSL_Library_init were void, and the SSL_Library_init always returns 1 according official documentation. For newest versions they began to deliver error, so I changed to have message.

web/server/README.md
thiagoftsm: I appended more a parameter for the destination, this way we split the configuration.

cakrit:
- Check the codacy errors
- Some bad links in the documentation (I'll provide a better rewrite after we're done with the testing though):
```
8:20:36 PM: -- ERROR: ./web/server/README.md - ../streaming is neither a file or a directory. Giving up!
8:20:36 PM: -- ERROR: ./streaming/README.md - ../server is neither a file or a directory. Giving up!
```
- Travis failing

cakrit:
On the Chrome console I get the following:

```
GET https://localhost:19999/api/v1/data?chart=system.ram&format=array&points=480&group=average&gtime=0&options=absolute%7Cpercentage%7Cjsonwrap%7Cnonzero&after=-480&dimensions=used%7Cbuffers%7Cactive%7Cwired&_=1558509953641 net::ERR_SSL_VERSION_INTERFERENCE
```

cakrit:
Connected via nginx proxy and unix socket. Got the following error: 
```
Secure Connection Failed

An error occurred during a connection to netdata.domain.tld:81. SSL received a record that exceeded the maximum permissible length. Error code: SSL_ERROR_RX_RECORD_TOO_LONG

    The page you are trying to view cannot be shown because the authenticity of the received data could not be verified.
    Please contact the website owners to inform them of this problem.
```
We obviously still do SSL with unix sockets as well, we shouldn't be doing it.

cakrit:
Issued an http request and it did redirect, but it took a very long time (30sec, i.e. there was a timeout). I saw the following:
```
2019-05-22 11:23:44: netdata ERROR : WEB_SERVER[static1] : (1092@/home/chri:web_client_send): Cannot send HTTP headers to web client. (errno 32, Broken pipe)
2019-05-22 11:23:44: netdata ERROR : WEB_SERVER[static1] : (1133@/home/chri:web_client_send): HTTP headers failed to be sent (I sent 386 bytes but the system sent -1 bytes). Closing web client.
```
After that, I actually did get the dashboard.
It's really strange because the first couple of times I did it, it redirected fine. Then I repeated the test another 5 times and got a timeout each time for the first https request after the redirect. At the same time, going directly to the https URL responded immediately.

web/server/README.md
cakrit: I will redo the README when we're done

cakrit:
I will just add comments on the code so we can resolve them one by one.

streaming/rrdpush.c
```diff
@@ -79,6 +79,18 @@ int rrdpush_init() {
         default_rrdpush_enabled = 0;
     }
 
+#ifdef ENABLE_HTTPS
```
cakrit: This is still not working quite as it should

> streaming sockets (not HTTP protocol)
Protocol parameter in bind to is ^SSL=force - Ignore and do the previous, *until netdata slaves can actually use SSL*. After we implement that, only accept encrypted connections.

Essentially, when I tell the master server SSL=force, it should deny the streaming connection. I did a `bind to = *^SSL=force` and it accepted an unencrypted connection.

In other words:
Default for tcp sockets except streaming = force
Default for streaming = optional
Override the default if the user provides ^SSL=....

streaming/stream.conf
```diff
@@ -17,7 +17,7 @@
     # Where is the receiving netdata?
```
cakrit: As discussed in the demo, we need a few more things here. 
Check section "Certificate Verification" in https://curl.haxx.se/docs/sslcerts.html, it covers all the cases. 
We will have the exact same thing when netdata acts as a client (i.e. in backends now and later when a netdata agent tries to connect to https://netdata.cloud).

streaming/stream.conf
```diff
@@ -26,6 +26,8 @@
     #             IPv6 IPs should be given with brackets [ip:address]
     # INTERFACE = the network interface to use (only for IPv6)
     # PORT      = the port number or service name (/etc/services)
+    # SSL       = when this word appear at the end of the destination string
+    #             the Netdata will do encrypt connection with the master.
```
cakrit: When I put the :SSL on my Fedora slave, I got this: 2019-05-22 17:31:05: netdata ERROR : STREAM_SENDER[fedora] : Failed to allocate SSL structure.
I haven't given it any ssl certificate paths and I shouldn't have to, correct?
I mean, only the server needs them, not the client

web/server/README.md
```diff
@@ -57,6 +57,35 @@ The API requests are serviced as follows:
 - `badges` gives access only to the badges API calls.
 - `management` gives access only to the management API calls.
 
+### Enabling TLS support
```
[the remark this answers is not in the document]
thiagoftsm: Thanks!

thiagoftsm:
- ERROR: ./web/server/README.md - ../streaming is neither a file or a directory. Giving up!
Here it was only a broke link, I already fixed, so you do not need to worry.

- net::ERR_SSL_VERSION_INTERFERENCE
I had to create a new rule on the server, this problems was happen, because we were allowing the TLS 1.3, right now we have a limitation to allow until TLS 1.2

- Error code: SSL_ERROR_RX_RECORD_TOO_LONG
When I tried to use the nginx with the 'normal ' configuration below, I had the same problems:
location /netdata/ {
proxy_redirect off;
            proxy_set_header Host $host;
            proxy_set_header X-Forwarded-Host $host;
            proxy_set_header X-Forwarded-Server $host;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_http_version 1.1;
            proxy_pass_request_headers on;
            proxy_set_header Connection "keep-alive";
            proxy_store off;
            proxy_pass http://localhost:19999;
}

 it was fixed after I do the following changes:

location /netdata/ {
           proxy_set_header X-Forwarded-Proto https; #I added this line
           proxy_pass https://localhost:19999; #I changed to HTTPS instead HTTP
}
ssl on;
ssl_certificate /etc/nginx/cert2048.pem;
ssl_certificate_key /etc/nginx/key2048.pem;

In my case, the key and certificate are the same I am using in the master.

ilyam8:
> ssl_certificate /etc/nginx/cert2048.pem;
> ssl_certificate_key /etc/nginx/key2048.pem;

> In my case, the key and certificate are the same I am using in the master.

just a note: they shouldn't be the same, nginx should have its own certs to terminate https connections

web/server/static/static-threaded.c
```diff
@@ -152,10 +152,67 @@ static void *web_server_add_callback(POLLINFO *pi, short int *events, void *data
     struct web_client *w = web_client_create_on_fd(pi->fd, pi->client_ip, pi->client_port, pi->port_acl);
     w->pollinfo_slot = pi->slot;
 
-    if(unlikely(pi->socktype == AF_UNIX))
+    if(unlikely(pi->socktype == AF_UNIX)) {
```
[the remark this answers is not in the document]
thiagoftsm: I could solve this bug using getpeername, but I did not like the solution that I gave case the function returns error.

streaming/stream.conf
thiagoftsm: You are correct, it is not necessary to have any certificate and key in the slave side, it was developer mistake, I copied and paste a code, but I had to change and I did not do! It was fixed.

thiagoftsm:
I fixed the majority of the problems yesterday, and the Unix socket is working fine! 
@ilyam8 you are correct, we must use different certificates and when I tested this way everything was fine, I used equal certificates initially, because it was the simplest thing, but it is not the correct road as you said.

thiagoftsm:
The SSL implementation is bringing some new possibilities of configuration for Netdata.
The first step to realize any test with Netdata is to create the SSL keys, we are using the following command:

$ openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -nodes -keyout key.pem -out cert.pem

After this is necessary to set on the master the following parameters:

[web]

When we do this configuration the server assumes that the SSL must be forced, except to stream. We can manipulate this behavior using the following parameter on bind to:

[web]
        bind to = *=dashboard|registry|badges|management|streaming|netdata.conf^SSL=optional unix:/tmp/netdata/netdata.sock

The newest parameter ^SSL accepts two parameter, "optional" and "force". The default value is "optional".

On the slave side we also have configuration to do. All the configuration will be inside stream.conf. The first option that has changes is "destination", now we can append ":SSL" to indicate that the stream is encrypted:

[stream]
        destination = 10.0.0.1:19999:SSL

Case the SSL is not specified, the stream will try to connect with plain and when it is present it will send encrypted information.

By default the server will always verify whether the certificate is valid, to avoid this check it is necessary to set

[stream]
        ssl invalid certificate = yes

Everything that we talked here cannot affect the unix socket, all the encryptation is done only with TCP.

What are the expected behaviors:

server(without "ssl key","ssl certificate" and ^SSL) + slave(without :SSL and "ssl invalid certificate" = yes) = Message in plain text for all channels
server(with "ssl key","ssl certificate" and without ^SSL) + slave(without :SSL and "ssl invalid certificate" = yes) = Message in plain text only for stream
server(with "ssl key","ssl certificate" and without ^SSL) + slave(with :SSL and "ssl invalid certificate" = no) = Message encrypted for all channels and without streaming when we have invalid server certificate
server(with "ssl key","ssl certificate" and without ^SSL) + slave(with :SSL and "ssl invalid certificate" = yes) = Message encrypted for all channels and streaming accepting invalid server certificate

server(with "ssl key","ssl certificate" and with ^SSL=force) + slave(without :SSL and "ssl invalid certificate" = no) = It does not accept stream and encrypt the other channels
server(with "ssl key","ssl certificate" and with ^SSL=force) + slave(with :SSL and "ssl invalid certificate" = no) = All channel encrypted, but stream won't connect with master that has invalid certificate
server(with "ssl key","ssl certificate" and with ^SSL=force) + slave(with :SSL and "ssl invalid certificate" = yes) = All channel encrypted and stream won't verificate certificate


server(with "ssl key","ssl certificate" and with ^SSL=optional) + slave(without :SSL and "ssl invalid certificate" = no) = It allows connection from HTTP and HTTPS, the stream works normally
server(with "ssl key","ssl certificate" and with ^SSL=optional) + slave(with :SSL and "ssl invalid certificate" = no) = It allows connection from HTTP and HTTPS, but stream won't connect with master that has invalid certificate
server(with "ssl key","ssl certificate" and with ^SSL=optional) + slave(with :SSL and "ssl invalid certificate" = yes) = It allows connection from HTTP and HTTPS and stream connect with master that has invalid certificate

The tests that are necessary yet:
- I did not test the newest configuration with old versions, but the configuration described adjust the Netdata to work like old versions. 
- I did not use UDP at all.
- All the browsers problems described here was present initially in my computers, but I am not seeing more them, it is necessary to check whether I am not working with cache.

ilyam8:
> ssl invalid certificate

i suggest to use `ssl skip certificate verification` instead. It is verbose but it describes exactly what is does.

ilyam8:
@thiagoftsm i did some simple tests

1. `ssl` dir is not created after installation, i think it should be created since we expect ssl stuff there by default.

2. with `force` option on the first visit to http page (`http://127.0.0.1:19999`) i got

sc

Is it as planned?

ilyam8:
btw all your `What are the expected behaviors:` should be in the readme imo (table!)

ilyam8:
And one more

if `cert.pem` or `key.pem` exist, but not readable we shouldn;t fall back to http.

web/server/README.md
```diff
@@ -57,6 +57,35 @@ The API requests are serviced as follows:
...
+
+## Certificate and key
+Netdata since version 1.16 has the feature to encrypt connections. To enable this option it is necessary to have a certificate and a key, case you do not have one yet, it is possible to generate them  with the following command
+
+$ openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -keyout key.pem -out cert.pem
```
ilyam8: As the `netdata` user i guess?
Does `netdata` user have enough permissions (write) to create certs in `ssl/` ?

ilyam8:
I haven't tested master/slave so far, `What are the expected behaviors` cases are hard to read.

web/server/README.md
thiagoftsm: The ideal case it is the netdata user, but case the user installed the software with other user either it is necessary to use it or to give read permission for all users.

thiagoftsm:
"if cert.pem or key.pem exist, but not readable we shouldn;t fall back to http." This is the expected behavior, because the master won't activate the SSL,  I will simulate this to confirm that this is happening.
Hi @ilyam8 I fixed good part of the things that you already asked me, I will verify the "force" on http now, this is the unique thing I did not fixed yet.

web/server/web_client.c
```diff
@@ -1303,7 +1373,16 @@ void web_client_process_request(struct web_client *w) {
...
             }
             break;
-
+#ifdef ENABLE_HTTPS
+        case HTTP_VALIDATION_REDIRECT:
+        {
+            buffer_flush(w->response.data);
+            w->response.data->contenttype = CT_TEXT_HTML;
+            buffer_strcat(w->response.data, "<!DOCTYPE html><!-- SPDX-License-Identifier: GPL-3.0-or-later --><html><body onload=\"window.location ='https://'+ window.location.hostname + ':' + window.location.port +  window.location.pathname\">Redirecting to safety connection, case your browser does not support redirection, please click <a onclick=\"window.location.href ='https://'+ window.location.hostname + ':' + window.location.port +  window.location.pathname\">here</a>.</body></html>");
+            w->response.code = 301;
```
cakrit: Issued an http request and it did redirect, but it took a very long time (30sec, i.e. there was a timeout). I saw the following:
```
2019-05-22 11:23:44: netdata ERROR : WEB_SERVER[static1] : (1092@/home/chri:web_client_send): Cannot send HTTP headers to web client. (errno 32, Broken pipe)
2019-05-22 11:23:44: netdata ERROR : WEB_SERVER[static1] : (1133@/home/chri:web_client_send): HTTP headers failed to be sent (I sent 386 bytes but the system sent -1 bytes). Closing web client.
```
After that, I actually did get the dashboard.
It's really strange because the first couple of times I did it, it redirected fine. Then I repeated the test another 5 times and got a timeout each time for the first https request after the redirect. At the same time, going directly to the https URL responded immediately. 

On 27-May I did more tests. First time, I got the netdata error you get when something fails (alert couldn't download something). Second time, got the 30s timeout and then got the redirect. Third time, redirect worked.

thiagoftsm: This problem has a direct relationship with the problem that Ilya reported early today, I already found the place in the code where we are having this problem, I am working to fix now. The problem is that I am doing the redirect, but I did not close the connection properly.

streaming/rrdpush.c
cakrit: It no longer accepts the unencrypted connection, but the log messages are misleading:
On the master, I don't get any messages. I should be seeing at least a connection refused or something.
On the slave, I see `STREAM fedora [send to 192.168.122.1:19999]: server is not replying properly (is it a netdata?).`

docs/Running-behind-nginx.md
```diff
@@ -117,6 +117,34 @@ Using the above, you access netdata on the backend servers, like this:
 - `http://nginx.server/netdata/server1/` to reach `backend-server1`
 - `http://nginx.server/netdata/server2/` to reach `backend-server2`
 
+### Using TLS communication
```
cakrit: Replace with the following:

### Using TLS communication

In case the Netdata web server has been [configured to use TLS](../web/server/#enabling-tls-support),
you must also encrypt the communication between Nginx and Netdata.

To enable encryption, first [enable SSL on nginx](http://nginx.org/en/docs/http/configuring_https_servers.html) and then put the following in the location section of the Nginx configuration: 
```
proxy_set_header X-Forwarded-Proto https;
proxy_pass https://localhost:19999;
```

If nginx is not configured as described here, you will probably receive the error `SSL_ERROR_RX_RECORD_TOO_LONG`.

netdata-installer.sh
```diff
@@ -443,6 +443,12 @@ if [ -d "${NETDATA_PREFIX}/etc/netdata" ]; then
 	done
 fi
 
+# default directory for certificates
+if [ ! -d "${NETDATA_PREFIX}/etc/netdata/ssl" ]; then
+	run mkdir "${NETDATA_PREFIX}/etc/netdata/ssl"
```
cakrit: align the tabs

docs/Running-behind-nginx.md
thiagoftsm: The next pull will have this fix, thanks!

netdata-installer.sh
thiagoftsm: It is done for next pull too, thanks!

streaming/rrdpush.c
thiagoftsm: It will also be in the next pull, thanks!

streaming/README.md
```diff
@@ -391,6 +391,43 @@ This is how such a solution will work:
 <img src="https://cloud.githubusercontent.com/assets/2662304/23627295/e3569adc-02b8-11e7-9d55-4014bf98c1b3.png"/>
 </p>
 
+### SSL on stream
```
cakrit: Move this section right before 'Viewing remote host dashboards' and replace it with the following:

### Securing the communication

Netdata does not activate TLS encryption by default. To encrypt the connection, you first need to [enable TLS support](../web/server/#enabling-tls-support) on the master. With encryption enabled on the receiving side, we need to instruct the slave to use SSL as well. On the slave's `stream.conf`, configure the destination as follows:

```
[stream]
    destination = host:port:SSL
```

The word SSL appended to the end of the destination tells the slave that the connection must be encrypted.

#### Certificate verification

When SSL is enabled on the slave, the default behavior will be do not connect with the master unless the server's certificate can be verified via the default chain. In case you want to avoid this check, add to the slave's `stream.conf` the following:

```
[stream]
```

#### Expected behaviors

With the introduction of SSL, the master-slave communication behaves as shown in the table below, depending on the following configurations:
- Master TLS (Yes/No): Whether the `[web]` section in `netdata.conf` has `ssl key` and `ssl certificate`.
- Master port SSL (-/force/optional): Depends on whether the `[web]` section `bind to` contains a `^SSL=force` or `^SSL=optional` directive on the port(s) used for streaming.
- Slave TLS (Yes/No): Whether the destination in the slave's `stream.conf` has `:SSL` at the end.
- Slave SSL Verification (yes/no): Value of the slave's `stream.conf` `ssl skip certificate verification` parameter (default is no).

 Master TLS enabled | Master port SSL | Slave TLS | Slave SSL Ver. | Behavior
:------:|:-----:|:-----:|:-----:|:--------
No | - | No | no | Legacy behavior. The master-slave stream is unencrypted.
Yes | force | No | no | The master rejects the slave connection.
Yes | -/optional | No | no | The master-slave stream is unencrypted (expected situation for legacy slaves and newer masters)
Yes | -/force/optional | Yes | no | The master-slave stream is encrypted, provided that the master has a valid SSL certificate. Otherwise, the slave refuses to connect.
Yes | -/force/optional | Yes | yes | The master-slave stream is encrypted.

cakrit: In this file, we need to also slightly modify the contents of the configuration section. Mention that they will need to read the `securing the communication` section for the additional options in the slave's stream.conf. i.e.

```
[stream]
    enabled = yes | no
    destination = IP:PORT[:SSL] ...
    api key = XXXXXXXXXXX
```

This is an overview of how these options can be combined:

target | memory<br/>mode | web<br/>mode | stream<br/>enabled | backend | alarms | dashboard
-------|:-----------:|:---:|:------:|:-------:|:---------:|:----:
headless collector|`none`|`none`|`yes`|only for `data source = as collected`|not possible|no
headless proxy|`none`|not `none`|`yes`|only for `data source = as collected`|not possible|no
proxy with db|not `none`|not `none`|`yes`|possible|possible|yes
central netdata|not `none`|not `none`|`no`|possible|possible|yes

For the options to encrypt the data stream between the slave and the master, refer to [securing the communication](#securing-the-communication)

streaming/stream.conf
cakrit: We can accept just the skip verification to merge this, but please create a new issue so we can add more options later. Put here a reference to that issue and we can resolve this.

web/server/README.md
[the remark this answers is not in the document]
cakrit: I'll work on this tomorrow, the other two READMEs will be ok with the changes I made.

web/server/README.md
cakrit: I will state that the files need to be readable by netdata in the update I'll make tomorrow.

streaming/README.md
thiagoftsm: it will be there in the next poll!

thiagoftsm:
@ilyam8 and @cakrit I fixed the problems that you described, we are having problem sometimes with the Google Chrome, because either it redirect incorrect the web page, due the fact the redirect right now is done with HTTP instead Header, and searching in the internet I discover that the Google suggests to use HTTP Header, or the it has a problem  with the Ciphers(error:00000001:lib(0):func(0):reason(1) ), this last is the most problematic, because we cannot restrict the CIPHERS so much, because this affects other browsers.  I changed the code to log more information about these errors on the server side. Tomorrow I will finish the redirect with Headers(301), this was not present in the webserver.

ilyam8:
On clean install i get

```bash
[/home/ilyam/src/test/netdata]# mkdir /opt/netdata/etc/netdata/ssl 
mkdir: cannot create directory '/opt/netdata/etc/netdata/ssl': No such file or directory
 FAILED   

[/home/ilyam/src/test/netdata]# chown -R : /opt/netdata/etc/netdata/ssl 
chown: cannot access '/opt/netdata/etc/netdata/ssl': No such file or directory
 FAILED   
```

cakrit:
> @ilyam8 and @cakrit I fixed the problems that you described, we are having problem sometimes with the Google Chrome, because either it redirect incorrect the web page, due the fact the redirect right now is done with HTTP instead Header, and searching in the internet I discover that the Google suggests to use HTTP Header, or the it has a problem with the Ciphers(error:00000001:lib(0):func(0):reason(1) ), this last is the most problematic, because we cannot restrict the CIPHERS so much, because this affects other browsers. I changed the code to log more information about these errors on the server side. Tomorrow I will finish the redirect with Headers(301), this was not present in the webserver.

Don't reply in comments here, but on review comments, so we can resolve them. @ilyam8, same for you, please add comments on the code itself somewhere (even in the documentation). We can't track resolution like this. Creating a new one for the `SSL Handshake error (error:00000001:lib(0):func(0):reason(1))`

web/server/web_client.c
```diff
@@ -143,7 +143,19 @@ void web_client_request_done(struct web_client *w) {
             debug(D_WEB_CLIENT, "%llu: Closing filecopy input file descriptor %d.", w->id, w->ifd);
```
cakrit: I'm always getting `SSL Handshake error (error:00000001:lib(0):func(0):reason(1))` now.

cakrit: Actually, I get the problem with ^SSL=force. When I don't set it, I get forbidden 3 times and then it works. Very strange...

Screenshot from 2019-05-28 12-12-46

netdata-installer.sh
```diff
@@ -443,6 +443,12 @@ if [ -d "${NETDATA_PREFIX}/etc/netdata" ]; then
 	done
 fi
 
+# default directory for certificates
+if [ ! -d "${NETDATA_PREFIX}/etc/netdata/ssl" ]; then
+    run mkdir "${NETDATA_PREFIX}/etc/netdata/ssl"
+    run chown -R "${ROOT_USER}:${NETDATA_GROUP}" "${NETDATA_PREFIX}/etc/netdata/ssl"
+fi
```
ilyam8: On clean install i get

```bash
[/home/ilyam/src/test/netdata]# mkdir /opt/netdata/etc/netdata/ssl 
mkdir: cannot create directory '/opt/netdata/etc/netdata/ssl': No such file or directory
 FAILED   

[/home/ilyam/src/test/netdata]# chown -R : /opt/netdata/etc/netdata/ssl 
chown: cannot access '/opt/netdata/etc/netdata/ssl': No such file or directory
 FAILED   
```

Steps to reproduce:
 - remove netdata
 - install netdata

`ssl` directory is not created during install

thiagoftsm: Thanks! I assumed that it exists before, so I used now "mkdir -p" to fix this. Thanks!

cakrit: Get rid of this section completely. Just add `"ssl"` just like `"custom-plugins.d"` was added in #6080 and #6089 (two different files, changing just one line containing the list of the subdirectories).

thiagoftsm: All right, I saw the lines and I already changed!

cakrit: Push the changes so we can resolve this plz.

streaming/rrdpush.c
cakrit: Comment here when it's ready for testing. I tried again and didn't see a difference.

thiagoftsm: Sorry, I changed here, but I did not execute a pull, because I am finishing the redirect, it will be ready after the Daily Standup.

thiagoftsm: I realized the commit with the reduction of the cipher numbers given by the Netdata and the installer directory, I am returning for the latest bug.

web/server/README.md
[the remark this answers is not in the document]
cakrit: Ok, try replacing the entire document with the contents of 
[README.txt](https://github.com/netdata/netdata/files/3227868/README.txt)

netdata-installer.sh
cakrit: I mentioned two files. `packaging/makeself/install-or-update.sh` also needs to be updated.

thiagoftsm: I updated both with the newest pull, and before this I also finish the redirect using Header instead a HTML page, but I kept the HTML page that redirects to simplify the life of a users that want to use the API to request something, with the page in hand they do not need to see the headers.

web/server/README.md
```diff
@@ -57,6 +57,59 @@ The API requests are serviced as follows:
...
+Inbound unix socket connections are unaffected, regardless of the SSL settings.  
+To enable SSL, provide the path to your certificate and private key in the `[web]` section of `netdata.conf`:
+
+```
+[web]
+	ssl key = /etc/netdata/ssl/key.pem
+	ssl certificate = /etc/netdata/ssl/cert.pem
+```
+
+Both files must be readable by the netdata user. For a master/slave connection, only the master needs these settings.
```
ilyam8: If any of those files is not readable/invalid the behavior is to fall back to http (even with `Force`). This should be documented imo.

cakrit: Good catch. here is the document with the update, please use this one @thiagoftsm 
[README.txt](https://github.com/netdata/netdata/files/3228172/README.txt)

thiagoftsm: When any file is not readable, the library won't load, so we will always have http.

web/server/web_client.c
thiagoftsm: This error is not happening more with this last pull, now I have one that is error:00000006:lib(0):func(0):reason(6), according with the information I read this is related with certificate that is not trust, and this is my case, because I am using self signed certificate.

web/server/README.md
thiagoftsm: I updated now!

web/server/web_client_cache.c
```diff
@@ -36,6 +48,12 @@ static void web_client_free(struct web_client *w) {
     buffer_free(w->response.data);
     freez(w->user_agent);
     freez(w);
+#ifdef ENABLE_HTTPS
+    if ( w->ssl.conn){
```
cakrit: Got a SEGFAULT here. I just had netdata use a unix socket and sent a request from nginx with SSL enabled on the server.

web/server/web_client.c
cakrit: I don't know what's happening, but it really doesn't work at all. I saw a redirect to https://127.0.1.1:19999 that didn't make sense and I keep getting forbidden, even from Firefox

web/server/web_client_cache.c
thiagoftsm: I tested on Slackware current and Debian 9.9 and I could not see the same problem, so I decided to read the official documentation of the library to understand what was happening. I saw that a possible change that I wrote last night could create this problem and I changed it, please, try again when you have a time. I will also continue testing on my side with different environments.

thiagoftsm:
I released more one update here,  while I was testing I checked that when I used "bind to = *^SSL=force" the result was not expected, I lost completely the access due the ACL that the system was receiving, so it was necessary to adjust the function that sets the ACL.

thiagoftsm:
Hi guys, I will give a last information about this last update that I am releasing. I tested during more than  6 hours with Debian 9.9, CentOS 7 and Slackware-Linux, this last was the unique that I tested using master and slave, the others I only tested the master. I got the following results:

1 - I verified that we had a bug with CentOS had a bug, to be more specific the OpenSSL is very old ,0.92. The ciphers that are used with the newest versions have different names when compared with 0.9x series, but  this OpenSSL version has a lot of functions deprecated, so I decided to remove completely the cipher definition for this specific version and I am allowing the browser negotiate with the server all options. The restriction that I am using with this version is only to avoid SSLv2 and SSLv3.
2 - We had more a bug with ourself, that the Netdata was redirecting every time that the request came from HTTP to the root of the HTTPS independt of the file requested, I used our internal variable that has the request to fix this, but this internal variable has been eliminated with the FIX of the URL parser, so this specific part of the code will be changed in the next days with another PR, but unitl there this solution can be kept.
3 - The unique "problem" remaining will appear, if and only if, the master or slave is running on a computer that haven't the hostname well configured, in other words, case the Netdata cannot resolve the hostname, or it is redirected to the loopback, when the Master/Slave sends the header with redirect, it will redirect for the computer that did the request. I think we can put this inside the manual too.

web/server/web_client_cache.c
cakrit: Same thing. 
```
NETDATA_SYSTEM_OS_NAME="Manjaro Linux"
NETDATA_SYSTEM_OS_ID=manjaro
NETDATA_SYSTEM_OS_ID_LIKE=arch
NETDATA_SYSTEM_OS_VERSION=18.0.4
NETDATA_SYSTEM_OS_VERSION_ID=unknown
NETDATA_SYSTEM_OS_DETECTION=Mixed
NETDATA_SYSTEM_KERNEL_VERSION=4.19.32-1-MANJARO
```

cakrit: It doesn't always happen. Where id did, it looked like poll_close_fd was called twice in the same thread. Will do a few more tests to see if I can find the cause.

thiagoftsm: During my tests yesterday, It never happened, I am doing a review in the parts of the code where I am managing the memory to check this specific problem;

cakrit: I sent you the valgrind output after a segfault

netdata-installer.sh
cakrit: The comment on the redirect shouldn't be here, but install-or-update.sh is updated.

web/server/web_client.c
cakrit: From thiago: 
I updated both with the newest pull, and before this I also finish the redirect using Header instead a HTML page, but I kept the HTML page that redirects to simplify the life of a users that want to use the API to request something, with the page in hand they do not need to see the headers.

web/server/web_client_cache.c
thiagoftsm: This was already fixed and the newest with stream looks like it is solved too!

web/server/web_client.c
```diff
@@ -724,7 +729,7 @@ const char *web_response_code_to_string(int code) {
 }
 
 static inline char *http_header_parse(struct web_client *w, char *s, int parse_useragent) {
-    static uint32_t hash_origin = 0, hash_connection = 0, hash_accept_encoding = 0, hash_donottrack = 0, hash_useragent = 0, hash_authorization = 0;
+    static uint32_t hash_origin = 0, hash_connection = 0, hash_accept_encoding = 0, hash_donottrack = 0, hash_useragent = 0, hash_authorization = 0, hash_host = 0;
```
cakrit: Fix Codacy issue (reduce scope of hash_accept_encoding)

web/server/web_client.c
```diff
@@ -847,6 +857,16 @@ static inline HTTP_VALIDATION http_request_validate(struct web_client *w) {
         w->mode = WEB_CLIENT_MODE_OPTIONS;
     }
     else if(!strncmp(s, "STREAM ", 7)) {
+#ifdef ENABLE_HTTPS
+        if ( (w->ssl.flags) && (netdata_use_ssl_on_stream & NETDATA_SSL_FORCE)){
+            w->header_parse_tries = 0;
+            w->header_parse_last_size = 0;
+            web_client_disable_wait_receive(w);
+            error("The server is configured to always use encrypt connection, please enable the SSL on slave.");
```
cakrit: Please add the slave's host name here, if you can.

libnetdata/socket/socket.c
```diff
@@ -301,14 +301,37 @@ void listen_sockets_close(LISTEN_SOCKETS *sockets) {
     sockets->failed = 0;
 }
 
+WEB_CLIENT_ACL socket_ssl_acl(char *ssl){
```
cakrit: valgrind: Uninitialized value & jump condition in line 988. I sent you the valgrind.xml

web/server/web_client.c
thiagoftsm: It was added with the last pull, thanks!

web/server/web_client.c
thiagoftsm: It was also fixed with the last pull.

libnetdata/socket/socket.c
thiagoftsm: I will talk more about it with you.

streaming/rrdpush.c
```diff
@@ -486,16 +535,51 @@ static int rrdpush_sender_thread_connect_to_master(RRDHOST *host, int default_po
               , host->program_name
               , host->program_version
     );
-
+    http[eol] = 0x00;
```
mfundul: Is this necessary? I think snprintfz() already does this.

web/server/web_server.c
```diff
@@ -137,6 +137,5 @@ void web_client_initialize_connection(struct web_client *w) {
     web_server_log_connection(w, "CONNECTED");
 
     web_client_cache_verify(0);
-}
-
 
+}
```
mfundul: I think we shouldn't change this file since we did not add any new code. It shouldn't be in the commit log cause it would be repository pollution.

streaming/rrdpush.c
thiagoftsm: This code was appended, because Codacy was saying that we were given a buffer without a 0x00 termination. As I said yesterday in our talk, we are having some information from Codacy that sometimes does not make sense, and this is a great example of them.

web/server/README.md
```diff
@@ -57,6 +57,72 @@ The API requests are serviced as follows:
...
+```
+[web]
+	ssl key = /etc/netdata/ssl/key.pem
+	ssl certificate = /etc/netdata/ssl/cert.pem
+```
+
+Both files must be readable by the netdata user. If any of the two files does not exist or is unreadable, Netdata falls back to HTTP. Case the files are not readable for Netdata, you can change the owner and permission for the files with the following commands:
+
+```
+$ chown netdata /etc/netdata/ssl/key.pem /etc/netdata/ssl/cert.pem
```
cakrit: no. `chown root:netdata` is enough. We just need to read this.

PR #5962, opened 2019-05-05
Update Node app_group

jonfairbanks:
##### Summary
Fixes the node app_group since they show like: `node /app/index.js` and not just `node`

##### Component Name
collectors/apps.plugin/apps_groups.conf

##### Additional Information
I've already signed the contributors file when adding similar in the past.


vlvkobal:
Thank you for the PR!

Issue #5964, opened 2019-05-05
phpfpm for multiple websites


Kofl:
Hello,

we had a look at the phpfpm plugin and it would be great for our need.
So far we have multiple phpfpm processes running, each per vhost for apache.

The configuration example only mentions one URL to configure:

```
local:
  url     : 'http://localhost/status'
```

but we would need to monitor one url per vhost. Any idea how to do that?

Thanks,
Thomas

OS: CentOS 7
php-fpm versions : 7.1, 7.2, 7.3

cakrit:
The config file in `/etc/netdata/orig/python.d/phpfm.conf` has much better info:

```
# ----------------------------------------------------------------------
# JOBS (data collection sources)
#
# The default JOBS share the same *name*. JOBS with the same name
# are mutually exclusive. Only one of them will be allowed running at
# any time. This allows autodetection to try several alternatives and
# pick the one that works.
#
# Any number of jobs is supported.
#
# All python.d.plugin JOBS (for all its modules) support a set of
# predefined parameters. These are:
#
# job_name:
#     name: myname            # the JOB's name as it will appear at the
#                             # dashboard (by default is the job_name)
#                             # JOBs sharing a name are mutually exclusive
#     update_every: 1         # the JOB's data collection frequency
#     priority: 60000         # the JOB's order on the dashboard
#     penalty: yes            # the JOB's penalty
#     autodetection_retry: 0  # the JOB's re-check interval in seconds
#
# Additionally to the above, PHP-FPM also supports the following:
#
#     url: 'URL'       # the URL to fetch nginx's status stats
#                      # Be sure and include ?full&status at the end of the url
#
# if the URL is password protected, the following are supported:
#
#     user: 'username'
#     pass: 'password'
#

# ----------------------------------------------------------------------
# AUTO-DETECTION JOBS
# only one of them will run (they have the same name)

localhost:
  name : 'local'
  url  : "http://localhost/status?full&json"

localipv4:
  name : 'local'
  url  : "http://127.0.0.1/status?full&json"

localipv6:
  name : 'local'
  url  : "http://[::1]/status?full&json"
```

Modifying the README...

ilyam8:
yo @Kofl you can define any number of jobs, details in the [module configuration file](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/phpfpm/phpfpm.conf)

Here is example for two instances:

```yaml
local:
  url     : 'http://127.0.0.1/status'

remote:
[...]

PR #5970, opened 2019-05-05
Add more configuration info to the PHP-FM collector readme

cakrit:
Update the README due to the confusion in issue #5964
Add more details on the PHP-FM module configuration.

### DIFF
--- a/collectors/python.d.plugin/phpfpm/README.md
+++ b/collectors/python.d.plugin/phpfpm/README.md
@@ -22,9 +22,7 @@ It produces following charts:
 
 ### configuration
 
-Needs only `url` to server's `status`
-
-Here is an example for local instance:
+Needs only `url` to server's `status`. The minimum configuration for a single job with a server in localhost is the following:
 
 ```yaml
 update_every : 3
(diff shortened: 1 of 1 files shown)


ilyam8:
i suggest:
 - update example config adding remote insance (like in [apache readme](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/apache#configuration))
 - add a link to collector configuration file (like [there](https://github.com/ilyam8/go.d.plugin/tree/master/modules/x509check#configuration))

Please do not copy the whole .conf to readme

ilyam8:
There is one more example
https://github.com/netdata/go.d.plugin/tree/master/modules/scaleio

i went further and moved chart to a separate file, i think readme should be small and clear and not wall of text.

cakrit:
> i went further and moved charts to a separate file, i think readme should be small and clear and not wall of text.

This changes the logic of one README.md per collector with effects on the generated html docs as well. If you've moved them, the info will be invisible in the docs site and we'll need to add yet another drop-down (not good, as we already have 4 levels to get to the modules). There's nothing wrong with a long README, though we could definitely format the list of charts in a much better way, so it doesn't take up all this space.  

I'll make changes on this one when I get some free time.

cakrit:
I'm closing this. We'll review with @ilyam8 the collector documentation overall.

PR #5971, opened 2019-05-05
Add "text-only" chart renderer

### LINKED ISSUE #5578: Custom dashboard - Just displaying a number
##### Question summary
When creating a custom dashboard (as documented at https://docs.netdata.cloud/web/gui/custom/#custom-dashboards), is there a way to just display a number, without a chart? As an example, I want to display "System Uptime" (system.uptime) as just a time span, without a graph. The way I imagine this working is that I could add a `div` or `span` with `data-chart-library="inline"` or something like that, and it'd just insert the raw value into the element.

Alternatively, I'm happy to write my own JavaScript to actually do the rendering on my page. After loading `dashboard.js`, is there a JS API I can hook in to get chart data? It could be something like this:
```
register({
  chart: 'system.uptime',
  handler: (data) => {
    // Run whatever code I want, using `data`
  },
})
```

Is such a function available in the JS SDK, or is it tightly coupled to the chart libraries themselves?

Daniel15:
##### Summary
Adds a "text-only" chart renderer. The idea is that you can do something like this:

```html
<div data-netdata="system.cpu" data-chart-library="textonly"></div>
```

And it'll just show the raw value within the element. Probably worth adding more options, but this is just a basic implementation to start with.

Demo: https://d.ls/netdata/textonly-test.htm

Fixes #5578 

##### Component Name
web/gui

### DIFF
--- a/web/gui/Makefile.am
+++ b/web/gui/Makefile.am
@@ -29,6 +29,7 @@ DASHBOARD_JS_FILES = \
     src/dashboard.js/charting/d3pie.js \
     src/dashboard.js/charting/d3.js \
     src/dashboard.js/charting/peity.js \
+    src/dashboard.js/charting/textonly.js \
     src/dashboard.js/charting.js \
     src/dashboard.js/chart-registry.js \
     src/dashboard.js/common.js \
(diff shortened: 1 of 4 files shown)


cakrit:
@Daniel15 can you also put a couple of examples in web/gui/dashboard.html so that others can see it in action? I will merge when you add that as well.

Daniel15:
Sorry I didn't get back to this @cakrit  - I can submit another PR adding some examples if you like.

cakrit:
That would be great mate, much appreciated!

PR #5974, opened 2019-05-06
Change HTTP version to HTTP method

SahAssar:
##### Summary

The actual screenshot shows methods, not versions, so change the text to reflect that.

The screenshot used:

 image 

##### Component Name

web_log docs


cakrit:
Thank you very much for the correction!

PR #6062, opened 2019-05-20
Prometheus remote write backend

### LINKED ISSUE #5619: Feature: add VictoriaMetrics backend for long-term archiving
##### Feature idea summary

Netdata supports archiving to [long-term storage](https://docs.netdata.cloud/backends/). It lacks support for [VictoriaMetrics](https://github.com/VictoriaMetrics/VictoriaMetrics/wiki/Single-server-VictoriaMetrics).

##### Expected behavior

It would be nice to have support for VictoriaMetrics backend in Netdata.
VictoriaMetrics may accept data points via the following APIs:
- Prometheus [remote write API](https://prometheus.io/docs/prometheus/latest/storage/#overview). The API allows sending data points in a simple [RemoteWrite Protocol Buffer](https://github.com/prometheus/prometheus/blob/master/prompb/remote.proto) over HTTP.
- InfluxDB line protocol.
- Graphite text protocol with tags.

As an additional bonus, support for writing data to Netdata backend via Prometheus remote write API automatically adds support for [many existing solutions supporting remote write API](https://prometheus.io/docs/operating/integrations/#remote-endpoints-and-storage).

vlvkobal:
##### Summary
Allow sending metrics to backends which support Prometheus Remote Write API.

Closes #5619

##### Component Name
backends


Makefile.am
```diff
@@ -512,6 +519,11 @@ if ENABLE_BACKEND_KINESIS
 else
     netdata_LINK = $(CCLD) $(CFLAGS) $(LDFLAGS) -o $@
 endif
+if ENABLE_BACKEND_PROMETHEUS_REMOTE_WRITE
+    netdata_LINK = $(CXXLD) $(CXXFLAGS) $(LDFLAGS) -o $@
+else
+    netdata_LINK = $(CCLD) $(CFLAGS) $(LDFLAGS) -o $@
```
paulkatsoulakis: all netdata_LINK should be part of a single if-else tree, but have you confirmed that is necessary to override the linker for this?

backends/backends.c
```diff
@@ -254,9 +254,9 @@ void *backends_main(void *ptr) {
     BUFFER *b = buffer_create(1), *response = buffer_create(1);
     int (*backend_request_formatter)(BUFFER *, const char *, RRDHOST *, const char *, RRDSET *, RRDDIM *, time_t, time_t, BACKEND_OPTIONS) = NULL;
     int (*backend_response_checker)(BUFFER *) = NULL;
+    int do_kinesis = 0, do_prometheus_remote_write = 0;
```
paulkatsoulakis: seems not necessary, why changed scope of do_kinesis variable here?

backends/backends.c
```diff
@@ -337,9 +337,8 @@ void *backends_main(void *ptr) {
             backend_request_formatter = format_dimension_stored_json_plaintext;
 
     }
-#if HAVE_KINESIS
     else if (!strcmp(type, "kinesis") || !strcmp(type, "kinesis:plaintext")) {
-
+#if HAVE_KINESIS
```
paulkatsoulakis: Same question with above actually, why kinesis stuff moving around?
Perhaps i should see the whole file to understand it?

backends/backends.c
```diff
@@ -354,9 +353,19 @@ void *backends_main(void *ptr) {
             backend_request_formatter = format_dimension_collected_json_plaintext;
         else
             backend_request_formatter = format_dimension_stored_json_plaintext;
+#else
+        error("AWS Kinesis support isn't compiled");
+#endif /* HAVE_KINESIS */
+    }
+    else if (!strcmp(type, "prometheus remote write")) {
+#if HAVE_PROTOBUF
```
paulkatsoulakis: HAVE_PROTOBUF indicates the presence of protobufs, this should allow setting ENABLE_BACKEND_PROMETHEUS_REMOTE_WRITE that we should use in turn throughout the implementation to enable prometheus related code.
Makes sense, what do you think? I am just thinking it might be confusing, but i am open to understand this approach too

backends/backends.c
```diff
@@ -354,9 +353,19 @@ void *backends_main(void *ptr) {
...
+        do_prometheus_remote_write = 1;
 
+        init_write_request();
+#else
+        error("Prometheus remote write support isn't compiled");
+#endif /* HAVE_PROTOBUF */
```
paulkatsoulakis: dito

backends/backends.c
```diff
@@ -451,6 +460,10 @@ void *backends_main(void *ptr) {
         size_t count_charts_total = 0;
         size_t count_dims_total = 0;
 
+#if HAVE_PROTOBUF
```
paulkatsoulakis: ditto

backends/backends.c
```diff
@@ -508,6 +521,7 @@ void *backends_main(void *ptr) {
             rrdhost_unlock(host);
         }
         rrd_unlock();
+#endif /* HAVE_PROTOBUF */
```
paulkatsoulakis: ditto

backends/backends.c
```diff
@@ -775,6 +789,12 @@ void *backends_main(void *ptr) {
     }
 #endif
 
+#if HAVE_PROTOBUF
```
paulkatsoulakis: ditto

backends/backends.h
```diff
@@ -53,4 +53,8 @@ extern int discard_response(BUFFER *b, const char *backend);
 #include "backends/aws_kinesis/aws_kinesis.h"
 #endif
 
+#if HAVE_PROTOBUF
```
paulkatsoulakis: ditto

backends/backends.c
vlvkobal: This one was needed to print an error message if a user set kinesis backend but its support hadn't been compiled.

backends/backends.c
vlvkobal: Throughout the C code, we usually use defines in form `HAVE_*`. Should we change this approach? The only `ENABLE_*` is `ENABLE_DBENGINE`.

Makefile.am
vlvkobal: When I was working on Kinesis backend, I couldn't link the daemon without the hack with `_LINK`. Though,  of course, it would be better to get rid of this override.

backends/backends.c
mfundul: ENABLE_DBENGINE originated from autoconf as advised [here](https://autotools.io/autoconf/arguments.html).

Makefile.am
```diff
@@ -500,14 +506,13 @@ NETDATA_COMMON_LIBS = \
...
 	$(NULL)
-# TODO: Find more graceful way to add libs for AWS Kinesis
 
 sbin_PROGRAMS += netdata
 netdata_SOURCES = $(NETDATA_FILES)
 netdata_LDADD = \
 	$(NETDATA_COMMON_LIBS) \
 	$(NULL)
-if ENABLE_BACKEND_KINESIS
+if ENABLE_CXX_LINKER
```
paulkatsoulakis: There is a ENABLE_BACKEND_KINESIS on Makefile.am.
should it be changed too ?

vlvkobal: No. We need ENABLE_CXX_LINKER for both Kinesis and Remote Write backends. ENABLE_BACKEND_KINESIS is used specifically for Kinesis dependencies.

paulkatsoulakis: ah yes of course, i rushed in, thanks!

PR #6064, opened 2019-05-21
Add perforce server process monitoring

akwan:
##### Summary

Perforce Helix has a variety of [Linux server processes](http://filehost.perforce.com/perforce/r18.2/bin.linux26x86_64/) named `p4d` (server daemon), `p4broker` (broker daemon), `p4p` (proxy daemon). Under production scenarios, only one of these would likely be running on a server or container.

##### Component Name

collectors/apps.plugin/apps_groups.conf


cakrit:
Thanks @akwan! :)

PR #6071, opened 2019-05-22
Create missing /etc/netdata/custom-plugins.d

Steve8291:
Not sure if this is the right place to put this but it would seem to be an easy fix for the error that shows up in the log files:
netdata ERROR : PLUGINSD : cannot open plugins directory '/etc/netdata/custom-plugins.d' (errno 2, No such file or directory)

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -403,6 +403,15 @@ progress "Compile netdata"
 run make -j$(find_processors) || exit 1
 
 # -----------------------------------------------------------------------------
+progress "Create custom-plugins.d"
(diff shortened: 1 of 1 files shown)


Steve8291:
Thanks @cakrit  
I just created a new pull request in what I think is the right file. 
/netdata-master/packaging/makeself
Maybe it also needs to be in /netdata-master/netdata-installer.sh as well?

cakrit:
wow, I didn't even notice we had this in makeself. nice catch! Yes, I approved that PR but either add the change to netdata-installer.sh to it as well, or do a third one please.

cakrit:
And I'm closing this one, just so we don't make a mistake.

PR #6089, opened 2019-05-22
Create missing /etc/netdata/custom-plugins.d

Steve8291:
Fix Error message in error.log:
netdata ERROR : PLUGINSD : cannot open plugins directory '/etc/netdata/custom-plugins.d' (errno 2, No such file or directory)


cakrit:
Thank you @Steve8291 !

PR #6112, opened 2019-05-24
Fix incorrect module name: energi

Steve8291:
I think this should be "energid" because that is the name of the module and currently setting it to "no" in the config file does not turn it off.


ilyam8:
@Steve8291 you are right, thanks!

PR #6113, opened 2019-05-24
Remove system_info copying

vlvkobal:
##### Summary
Fixes segmentation fault and missing system info introduced with #6096.

Fixes https://github.com/netdata/cloud/issues/779

##### Component Name
daemon
streaming
database

### DIFF
--- a/database/rrdhost.c
+++ b/database/rrdhost.c
@@ -812,6 +812,8 @@ restart_after_removal:
 // RRDHOST - set system info from environment variables
 
 int rrdhost_set_system_info_variable(RRDHOST *host, struct rrdhost_system_info *system_info, char *name, char *value) {
+    int res = 0;
+
     if(host) rrdhost_wrlock(host);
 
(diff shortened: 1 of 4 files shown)


database/rrdhost.c
```diff
@@ -811,82 +811,66 @@ void rrdhost_cleanup_obsolete_charts(RRDHOST *host) {
...
         system_info->container = strdupz(value);
     }
     else if(!strcmp(name, "NETDATA_SYSTEM_CONTAINER_DETECTION")){
+        freez(system_info->container_detection);
         system_info->container_detection = strdupz(value);
     }
-    else return 1;
+    else {
+        if(host) rrdhost_unlock(host);
+        return 1;
```
paulkatsoulakis: can i propose here to use a variable  "int RESULT=0" that you will set to 1 at this point.
Then you only do in the end the unlock and "return RESULT";

PR #6116, opened 2019-05-24
Active processes number alert

### LINKED ISSUE #2239: alarms to monitor the number of processes in a system
A quite common case is for a cron job to keep forking processes endlessly that do not end in time, so the number of processes grows forever.

It would be very nice to have an alarm for this case.

apardyl:
##### Summary
Resolves #2239

This PR adds an alarm on abnormally high number of active processes.

The thresholds are as follow (roughly 80% and 90% of max pid value):
* FreeBSD (max process number: 999999, hardcoded):
  * WARNING: 80000 (clear alarm on 75000)
  * CRITICAL: 90000 (clear alarm on 85000)
* Linux (default max process number: 32768, can be manually changed with sysctl, up to 2^22 on 64-bit kernels and 32768 on 32-bit):
  * WARNING: 26000 (clear alarm on 25000)
  * CRITICAL: 30000 (clear alarm on 28000)

The value is taken from `system.active_processes.active`, measured every 5s (should be enough to send a notification before the limit is reached unless someone launches an exponential fork-bomb). There is no alarm for Mac OS X as the macos.plugin does not seem to return `system.active_processes`. 

##### Component Name
health

##### Additional Information
Values from:
* https://www.freebsd.org/cgi/man.cgi?query=intro&apropos=0&sektion=2&manpath=FreeBSD+12.0-RELEASE&arch=default&format=html (Process ID)
* https://github.com/torvalds/linux/blob/master/include/linux/threads.h (PID_MAX_LIMIT)

### DIFF
--- a/health/Makefile.am
+++ b/health/Makefile.am
@@ -65,6 +65,7 @@ dist_healthconfig_DATA = \
     health.d/phpfpm.conf \
     health.d/portcheck.conf \
     health.d/postgres.conf \
+    health.d/processes.conf \
     health.d/qos.conf \
     health.d/ram.conf \
     health.d/redis.conf \
--- a/health/health.d/processes.conf
+++ b/health/health.d/processes.conf
@@ -1,0 +1,27 @@
+# you can disable an alarm notification by setting the 'to' line to: silent
+
+   alarm: active_processes_limit_freebsd
+      on: system.active_processes
+      os: freebsd
+   hosts: *
+    calc: $active
+   units: processes
+   every: 5s
+    warn: $this > (($status >= $WARNING)  ? (75000) : (80000))
+    crit: $this > (($status == $CRITICAL) ? (85000) : (90000))
+   delay: down 5m multiplier 1.5 max 1h
+    info: the number of active processes
+      to: sysadmin
+
+   alarm: active_processes_limit
+      on: system.active_processes
+      os: linux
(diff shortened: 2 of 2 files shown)


ilyam8:
> can be manually changed with sysctl

Since it is configurable we could read `/proc/sys/kernel/pid_max` and add the value to the `system.active_processes` chart as variable. If so alarm will calc %, i think it is better then hardcoded values.

What do you think @apardyl ?
cc @vlvkobal

apardyl:
@ilyam8 Yes, we can do that. We would probably also need to read `/proc/sys/kernel/threads-max` - `pid_max` is the max PID that can be assigned, `threads-max` is the actual limit of threads/processes (by default higher than pid_max, but can be set lower that it), at least if I understand http://man7.org/linux/man-pages/man5/proc.5.html correctly.
I'm not sure if this is needed - 32768 processes is quite a number, I don't think many users get close to it on a single (virtual)machine. I can implement this as well, I see it would require adding one or two (if we want to check `threads-max`) new files to monitor to the proc.plugin.

ilyam8:
@apardyl 
I think you are right, your PR is ok. As for the `pid_max`/`threads-max` i suggest to open a new issue as a feature request and implement it later.

cc @Ferroin

cakrit:
Thanks @apardyl !

PR #6121, opened 2019-05-25
netdata/packaging/installer: nits and fixes

### LINKED ISSUE #6094: netdata/packaging/installer: Fine tuning based on user feedback and findings from testing
##### Bug report summary
1) check for `tar`, `grep` and inform the user accordingly in various key places
2) When go is not possible to be installed for any reason, just inform the user he can disable-go and fail cleanly
3)long shot, but try to clear some script errors that break the output (even though they are not affecting execution), for example:
```
[root@4500509af5ff netdata]# ./install-required-packages.sh netdata
./install-required-packages.sh: line 990: pkg_grep: bad array subscript
./install-required-packages.sh: line 993: pkg_grep: bad array subscript
./install-required-packages.sh: line 980: validate_: command not found
Loading /etc/os-release ...
```
##### OS / Environment
Any, though some exist on specific distros

latest

##### Component Name
netdata/packaging/installer

##### Steps To Reproduce
Run the scripts on the various distros

##### Expected behavior
Installers should always produce meaningful instructions when unhandled situations arise or complete successfully netdata install.

paulkatsoulakis:
##### Summary
In this PR we attempt to further fine tune installation process. Fixes include visual adjustments, better support and cleaner handling (details below)

1) Go.d plugin is no longer mandatory.
   a) Make it a soft error when something goes wrong with download or install
   b) Inform the user about the option of disabling it the next time

2) Add better support for debian 8 and 9, start/stop process and startup script installation was failing

3) Add better handling on initial stop/start of netdata service, use the identified start/stop behaviour when doing the final service restart

4) More verbosity here and there. Also changed a couple of existing prints with clearer messages

5) In the FreeBSD case use onestart to avoid warnings about rc.conf, but at the same time warn during start script install about the missing step

6) Wherever you running commands use the run method to pretty-print and soft handle errors.

##### Component Name
netdata/packaging/installer

##### Additional Information
Fixes #6094


packaging/installer/functions.sh
```diff
@@ -303,7 +303,7 @@ install_non_systemd_init() {
 				run rc-update add netdata default &&
 				return 0
 
-		elif [ "${key}" = "debian-7" ] || [ "${key}" = "ubuntu-12.04" ] || [ "${key}" = "ubuntu-14.04" ]; then
+		elif [[ ${key} =~ (debian-[789]).* ]] || [ "${key}" = "ubuntu-12.04" ] || [ "${key}" = "ubuntu-14.04" ]; then
```
ktsaou: debian 8 and 9 are init based, not systemd?

paulkatsoulakis: hm, i fall in the dockerized version trap again? 
I did not find anything in the release notes of both versions either
Let me check again

packaging/installer/functions.sh
```diff
@@ -477,20 +480,29 @@ restart_netdata() {
...
-		service netdata stop
-		stop_all_netdata
-		service netdata restart && started=1
+		echo >&2
+		echo >&2 "Stopping all netdata threads"
+		run stop_all_netdata
+		local start_cmd="${NETDATA_START_CMD}"
+
+		if [ "$(uname 2> /dev/null)" == "FreeBSD" ]; then
+			start_cmd="${NETDATA_START_CMD/ start/ onestart}"
```
ktsaou: hm... I now see what you meant.
It is probably better to set 2 variables, at the point `NETDATA_START_CMF` is set:

```sh
NETDATA_START_CMD="the command to start netdata from the installer"
NETDATA_USER_START_CMD="the command a user should use to start netdata"
```

packaging/installer/functions.sh
paulkatsoulakis: What's wrong with me? i did not find it in the first read through
https://www.debian.org/releases/jessie/amd64/release-notes/ch-whats-new.en.html#systemd

I was also a bit misguided from a bug report on a devel version of jessie that fall into that branch. 
I 'll use a droplet to test again

packaging/installer/functions.sh
paulkatsoulakis: ack

paulkatsoulakis:
Spotted on potential issue when our kickstart running under dash.
The double equality sign is not working on dash, so i modified all findings to single equal which works on all other cases AND dash

netdata-installer.sh
```diff
@@ -44,15 +44,21 @@ else
...
+download_go() {
 	url="${1}"
 	dest="${2}"
+
 	if command -v curl >/dev/null 2>&1; then
-		run curl -sSL --connect-timeout 10 --retry 3 "${url}" >"${dest}" || fatal "Cannot download ${url}"
+		run curl -sSL --connect-timeout 10 --retry 3 "${url}" >"${dest}"
 	elif command -v wget >/dev/null 2>&1; then
-		run wget -T 15 -O - "${url}" >"${dest}" || fatal "Cannot download ${url}"
+		run wget --waitretry 3 -T 10 -O - "${url}" >"${dest}"
```
cakrit: There was a reason we did the timeout and retries for wget like this. Please put it back as it was, we had tested thoroughly this part of the code.

netdata-installer.sh
```diff
@@ -775,24 +781,34 @@ install_go() {
...
 
 		# Checksum validation
 		if ! (cd "${tmp}" && safe_sha256sum -c "sha256sums.txt"); then
+
+			echo >&2
+			echo >&2 "go.d plugin checksum validation failure."
+			echo >&2 "Either check the error or consider disabling it by issuing '--disable-go' in the installer"
+			echo >&2
+
 			run_failed "go.d.plugin package files checksum validation failed."
```
cakrit: Not sure if this is needed any more. Perhaps it should be combined with the above. Right now it shows:

```
[/home/christopher/CLionProjects/netdata]# curl -sSL --connect-timeout 10 --retry 3 https://github.com/netdata/go.d.plugin/releases/download/v0.6.0/config.tar.gz
curl: (6) Could not resolve host: github.com
FAILED

go.d.plugin-v0.6.0.linux-amd64: FAILED
config.tar.gz: FAILED
sha256sum: WARNING: 2 computed checksums did NOT match

go.d plugin checksum validation failure.
Either check the error or consider disabling it by issuing '--disable-go' in the installer

FAILED  go.d.plugin package files checksum validation failed.

[/home/christopher/CLionProjects/netdata]# rm -rf /usr/lib/netdata/conf.d/go.d
OK

[/home/christopher/CLionProjects/netdata]# rm -rf /usr/lib/netdata/conf.d/go.d.conf
OK

[/home/christopher/CLionProjects/netdata]# tar -xf /tmp/netdata-go-3c7sL3/config.tar.gz -C /usr/lib/netdata/conf.d/
tar: This does not look like a tar archive

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
FAILED
```

The second is also problematic

netdata-installer.sh
paulkatsoulakis: The idea was to align the way we retry and timeout on curl, on wget.
I am okay with any values, as long as we use the same pattern for either curl or wget so that we use consistent settings between the two

PR #6122, opened 2019-05-25
Fix error with apcupsd_status

Steve8291:
apcupsd_status fails to run.
Fixes the following errors in log:
/usr/libexec/netdata/plugins.d/../charts.d/apcupsd.chart.sh: line 50: apcupsd_status: command not found
/usr/libexec/netdata/plugins.d/../charts.d/apcupsd.chart.sh: line 51: [: !=: unary operator expected

### DIFF
--- a/collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
+++ b/collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
@@ -47,8 +47,8 @@ apcupsd_check() {
 			error "cannot get information for apcupsd server ${host} on ${apcupsd_sources[${host}]}."
 			failed=$((failed + 1))
 		else
-			apcupsd_status = "$(apcupsd_get ${apcupsd_sources[${host}]} | awk '/^STATUS.*/{ print $3 }')"
-			if [ ${apcupsd_status} != "ONLINE" ]  && [ ${apcupsd_status} != "ONBATT" ]; then
+			apcupsd_status=$(apcupsd_get ${apcupsd_sources[${host}]} | awk '/^STATUS.*/{ print $3 }')
+			if [ "$apcupsd_status" != "ONLINE" ]  && [ "$apcupsd_status" != "ONBATT" ]; then
 				error "APC UPS ${host} on ${apcupsd_sources[${host}]} is not online."
 				failed=$((failed + 1))
 			else


collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
```diff
@@ -47,8 +47,8 @@ apcupsd_check() {
 			error "cannot get information for apcupsd server ${host} on ${apcupsd_sources[${host}]}."
 			failed=$((failed + 1))
 		else
-			apcupsd_status = "$(apcupsd_get ${apcupsd_sources[${host}]} | awk '/^STATUS.*/{ print $3 }')"
-			if [ ${apcupsd_status} != "ONLINE" ]  && [ ${apcupsd_status} != "ONBATT" ]; then
+			apcupsd_status=$(apcupsd_get ${apcupsd_sources[${host}]} | awk '/^STATUS.*/{ print $3 }')
+			if [ "$apcupsd_status" != "ONLINE" ]  && [ "$apcupsd_status" != "ONBATT" ]; then
```
paulkatsoulakis: Great catch! Empty string would break that the way it was before.
Please re-instate the curly brackets though on the variables

paulkatsoulakis: Other than that, change looks fine. Thanks for reporting it !

Steve8291: I've made a new pull request here:
[https://github.com/netdata/netdata/pull/6126](url)
Could you tell me why the curly braces are preferred over quotes. I've always used quotes unless I wanted to expand into some other text?

paulkatsoulakis: The curly brackets and the quotes are two separate things.

Curly brackets in general is a good practice, to make sure variables are expanded correctly.  You may look for bash curly brackets usage to further detail the need.
 They are not always necessary obviously, but for consistency we use them on all variables (I do not recall also whether shellcheck linter complaints when they are missing).

Now quotes, are mission critical on the if-else condition because if the variable is either empty or has spaces, it will break the parsing of the line and the script will break.

I hope this clarifies

paulkatsoulakis:
Hi @Steve8291 I will be closing this.
From what i see this is included on #6126 Pull Request, so no need for separate PR

PR #6124, opened 2019-05-25
tmp file not created

Steve8291:
##### Summary
Trying to fix the following error:

> rm: cannot remove '/tmp/.netdata-charts.d-p2IviKryfN/run.8500': No such file or directory

I'm pretty sure what I have fixed was incorrect but the error still persists.
With my changes charts.d will create the dir /tmp/netdata-charts.d-XXXXXX
However, there is no "run.XXXX' file written into it.
Hopefully someone can look at this and see what else is wrong.

##### Component Name
charts.d

### DIFF
--- a/collectors/charts.d.plugin/charts.d.plugin.in
+++ b/collectors/charts.d.plugin/charts.d.plugin.in
@@ -17,7 +17,7 @@ export PATH="${PATH}:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin"
 
 PROGRAM_FILE="$0"
 PROGRAM_NAME="$(basename $0)"
-PROGRAM_NAME="${PROGRAM_NAME/.plugin/}"
+PROGRAM_NAME="${PROGRAM_NAME}/.plugin/"
 MODULE_NAME="main"
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


paulkatsoulakis:
Hi @Steve8291, thanks for your PR!  We love it when our users take the initiative to support us!

Looking at the PR i can see one fix that is certainly a typo BUG.
The change regarding temp folder creation name, does not seem to be a reason to raise the error you mention.

I 'll have a look at the workflow and come back with further comments.

Steve8291:
Thanks @paulkatsoulakis I'm still new to using github and I don't have the years of experience coding that most of you surely do. I'd like to try to help out where I can with this project because it's amazing!
I should have made that an "Issue" rather than a "Pull-request." Sorry about that. I knew it didn't fix the problem but was trying to track it down. 
If it helps:
Line 307 of charts.d.plugin.in is where `rm "${TMP_DIR}/run.${pid}"` in the function "run" fails to remove the file. It also looks like that same function is where the `"${TMP_DIR}/run.${pid}"` file is written to/created.

Steve8291:
I think I've figured this error out. It is not caused by the charts.d.plugin.in file. However, the `"${TMP_DIR}/run.${pid}"` file is still not being written but maybe this only occurs when there is an error?
The problem was being caused by `/usr/libexec/netdata/charts.d/apcupsd.chart.sh`
It was passing incorrect info to the `run()` function in charts.d.plugin
I'll submit a new pull request to be reviewed.
I no longer see the error in my logs and apcupsd is making charts. Hope I got it.

paulkatsoulakis:
@Steve8291 if that is okay with you, i will close this PR as it looks invalid overall.
The first change, is not a bug and the second is not attending your issue (you opened separate PR already).

To explain a bit, the command `PROGRAM_NAME="${PROGRAM_NAME/.plugin/}"` basically removes `.plugin` from PROGRAM_NAME and re-assigns it to PROGRAM_NAME.  This is on purpose as you can understand, because we want the name without the extension later

paulkatsoulakis:
Thanks again for reporting your problem.
If that is okay, you always need to report an issue to us with your problem and then if you have a concluded on a change that you wish to share with us, you can proceed with a pull request towards netdata/netdata.

Don't worry though, we are always here to help for netdata

PR #6126, opened 2019-05-26
Pass correct info to run funct.

Steve8291:
Changed to allow the correct parameters to be sent to the run function.
Also changed to set apcupsd_status variable correctly.

##### Summary
Trying to fix the following errors that show up in the error.log:
```

> rm: cannot remove '/tmp/.netdata-charts.d-vTJX1mmTxi/run.24956': No such file or directory
> /usr/libexec/netdata/plugins.d/../charts.d/apcupsd.chart.sh: line 50: apcupsd_status: command not found
> /usr/libexec/netdata/plugins.d/../charts.d/apcupsd.chart.sh: line 51: [: !=: unary operator expected

```
Also trying to fix the delay that seems to happen for me for apcupsd charts to appear on the web.
##### Component Name
collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh


collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
```diff
@@ -41,13 +41,13 @@ apcupsd_check() {
...
 	for host in "${!apcupsd_sources[@]}"; do
-		run apcupsd_get "${apcupsd_sources[${host}]}" >/dev/null
+		apcupsd_get "${apcupsd_sources[${host}]}" >/dev/null
 		# shellcheck disable=2181
 		if [ $? -ne 0 ]; then
 			error "cannot get information for apcupsd server ${host} on ${apcupsd_sources[${host}]}."
 			failed=$((failed + 1))
 		else
-			apcupsd_status = "$(apcupsd_get ${apcupsd_sources[${host}]} | awk '/^STATUS.*/{ print $3 }')"
+			apcupsd_status=$(apcupsd_get ${apcupsd_sources[${host}]} | awk '/^STATUS.*/{ print $3 }')
```
paulkatsoulakis: Why remove the quotes here?
I would instead add quotes to $apcupsd_status on the if-else

collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
```diff
@@ -41,13 +41,13 @@ apcupsd_check() {
 
 	local host working=0 failed=0
 	for host in "${!apcupsd_sources[@]}"; do
-		run apcupsd_get "${apcupsd_sources[${host}]}" >/dev/null
+		apcupsd_get "${apcupsd_sources[${host}]}" >/dev/null
```
paulkatsoulakis: you mention passing correct info to run, but here you remove it.
Could you elaborate a bit on the change? Thanks!

Steve8291: Line# 26 would cause run() to run twice. I thought that was the problem.
Line# 44 was passing `run apcupsd_get` back into `run`
I think this would essentially send the array apcupsd_sources through twice.
That seemed like a problem.

collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
Steve8291: I thought that setting a variable `apcupsd_status` to the output of a command required that there be no spaces between the equals sign. I've also not seen it done with quotes around the $(commands).

paulkatsoulakis: when assigning the variables we follow the code style like so. We add the quotes.
In the if-else on the other hand, is mandatory to avoid breaking the script in the peculiar case the variable is empty.

collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
paulkatsoulakis: Could you provide details on the PR description around the problem you faced and what are you trying to fix with this change?

Steve8291: Done. Let me know if you need me to provide anything else.

paulkatsoulakis: Ok now i see what's the case. Just go to file `charts.d.plugin.in` line 307 and do the following:

change this
```
307         rm "${TMP_DIR}/run.${pid}"
```
to this
```
307         rm -f "${TMP_DIR}/run.${pid}"
```

I am not 100% sure why this doesn't get created all the times even empty, but adding -f should silence this soft error

Steve8291: That would fix it. I was afraid to do something like that in case it was problematic that the file wasn't being created in the first place. I'll add a new commit with this change if you are good with it.

collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
Steve8291: Thanks for the explanation @paulkatsoulakis. So the proper fix is going to be:
```
apcupsd_status="$(apcupsd_get ${apcupsd_sources[${host}]} | awk '/^STATUS.*/{ print $3 }')"
if [ "${apcupsd_status}" != "ONLINE" ]  && [ "${apcupsd_status}" != "ONBATT" ]; then
```
Adding the quotes back in to line 50 and removing the spaces around the equals sign.
Adding quotes on the test variables on line 51
If that is OK I will make a new pull request with the changes.

paulkatsoulakis: no need to make new pull request, you can just update your branch and push

collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh
paulkatsoulakis: yes, please go ahead.

Steve8291:
@cakrit I committed back into this branch "patch-8" and then did a 
`git rebase --interactive patch-8`
I've never done a rebase before. Please let me know if I did it incorrectly. Looks like I caused some errors on the LGTM code analysis.

cakrit:
> @cakrit I committed back into this branch "patch-8" and then did a
> `git rebase --interactive patch-8`
> I've never done a rebase before. Please let me know if I did it incorrectly. Looks like I caused some errors on the LGTM code analysis.

It still shows conflicts. While having the branch checked out, I use `git pull --rebase upstream master`. It will show some errors and list the files. I edit each file, look for `>>>`, update it, save it and then do `git add filename` for each of them. Finally, you do `git rebase --continue` (the instructions are shown after the first command).

Issue #6125, opened 2019-05-26
web_log reports unmatched lines


Komic:
This seems to be the problematic line (should parse as a 502):

```txt

2a01:abc:abc:abcd:abcd:abcd:abc:abc - - [26/May/2019:03:50:04 +0200] "GET /:/websockets/notifications?X-Plex-Token=redacted HTTP/1.1" 502 166 261 0.002 0.000, 0.004 "-" "okhttp/2.6.0"

```

ilyam8:
Hi @Komic 	

> 0.000,

i think the problem is the comma. Do you really need it? What is you log format?

Komic:
My log_format directive is the one provided by web_log.conf, `log_format netdata '$remote_addr - $remote_user [$time_local] "$request" $status $body_bytes_sent $request_length $request_time $upstream_response_time "$http_referer" "$http_user_agent"';`.

Wing924:
upstream_response_time may contain "," if first upstream server return error and nginx request to second server.

ilyam8:
ok, i think removing traling spaces from

https://github.com/netdata/netdata/blob/9f532f4981bdc4455a3115e580fe4014ff928efb/collectors/python.d.plugin/web_log/web_log.chart.py#L663

and

https://github.com/netdata/netdata/blob/9f532f4981bdc4455a3115e580fe4014ff928efb/collectors/python.d.plugin/web_log/web_log.chart.py#L671

will fix the problem, @Komic could you test it?

Komic:
yup, that worked, thanks!

Wing924:
btw, since upstream_response_time may contain space, you need to wrap it with double quotation.

Wing924:
btw, since upstream_response_time may contain space, you need to wrap it with double quotation.

PR #6138, opened 2019-05-27
web_log: remove trailing space from nginx_ext2 pattern

ilyam8:
##### Summary
ssia 

Fixes: #6125

> upstream_response_time may contain "," if first upstream server return error and nginx request to second server.

##### Component Name

[/collectors/python.d.plugin/web_log](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/web_log)

### DIFF
--- a/collectors/python.d.plugin/web_log/web_log.chart.py
+++ b/collectors/python.d.plugin/web_log/web_log.chart.py
@@ -4,9 +4,8 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import bisect
-import re
 import os
-
+import re
 from collections import namedtuple, defaultdict
 from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


collectors/python.d.plugin/web_log/web_log.chart.py
```diff
@@ -660,7 +660,7 @@ def find_regex(self, last_line):
                                        r' (?P<bytes_sent>\d+)'
                                        r' (?P<resp_length>\d+)'
                                        r' (?P<resp_time>\d+\.\d+)'
-                                       r' (?P<resp_time_upstream>[\d.-]+) ')
+                                       r' (?P<resp_time_upstream>[\d.-]+)')
```
Wing924: `r' "?(?P<resp_time_upstream>[\d.-]+)(, [\d.]+)*"?')`
it should match `0.01`, `"0.01"`, `0.01, 0.02` and `"0.01, 0.02"`.

ilyam8: Since `resp_time_upstream` is the last parameter `"?(?P<resp_time_upstream>[\d.-]+)` is enough, it will match the same but faster

ilyam8: if we do not need sum all `resp_time_upstream` ofc

Wing924: Because `resp_time_upstream` should be wrapped with `"`, I think 
`r' "?(?P<resp_time_upstream>[\d.-]+)')` is better.

Issue #5851, opened 2019-04-12
go.d.plugin x509check module does not track invalid hosts


astorath:
##### Bug report summary
If go.d/x509check module can't connect to source for any reason (invalid certificate, connection issues, etc.) it does not output any metric.

##### OS / Environment
- CentOS 7
- Ubuntu 18.04

##### Netdata version (ouput of `netdata -V`)
- netdata v1.13.0
- netdata v1.14.0-rc0

##### Component Name
go.d/x509check

##### Steps To Reproduce
- enable x509check plugin with `edit-config go.d.conf`:
```
modules:
  x509check: yes
```
- add some job to monitor `edit-config go.d/x509check.conf`:
```
update_every: 1
autodetection_retry: 1

jobs:
  - name: wrong_host
    source: https://wrong.host.badssl.com:443
```

##### Expected behavior

IMHO the best option is to use empty value for such cases.

ilyam8:
Hi @astorath 

From the [x509check readme](https://github.com/netdata/go.d.plugin/blob/master/modules/x509check/README.md) i see it collects only `Time Until Certificate Expiration in seconds`

> If go.d/x509check module can't connect to source for any reason (invalid certificate, connection issues, etc.) it does not output any metric.

There is no metric if module failed to connect to the source.

___
There are 2 cases:
 - failed to connect on start
 - failed to connect during runtime

If i understand you right you are talking about 1st case and your suggestion is to run the job and create empty chart? Am i right?

What is your use case? Why do you need that?

astorath:
@ilyam8 
Hi, my case is the first one: if something is wrong with the cert, I don't see any metric at all, and can't use it in alerting. So yes, creating an empty metric is fine, I think.

ilyam8:
check [portcheck](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/portcheck) and [httpcheck](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/httpcheck) modules

I think you could use them.

x509check is responsible only for checking time until certificate expiration.

ilyam8:
Speaking about checks - i think we need to add `last_collected_secs` to the x509check module, like

https://github.com/netdata/netdata/blob/e45c89c088dba522f07e724d74b75983079b105e/health/health.d/apache.conf#L4-L13

ilyam8:
But i don't want to change current behavior of x509check collector - start the job if first data collection was not successful

cakrit:
[...]

PR #6139, opened 2019-05-27
x509check: add last_collected alarm

ilyam8:
##### Summary

ssia

Fixes: #5851

> Hi, my case is the first one: if something is wrong with the cert, I don't see any metric at all, and can't use it in alerting. So yes, creating an empty metric is fine, I think.

##### Component Name

[/health](https://github.com/netdata/netdata/tree/master/health)

### DIFF
--- a/health/health.d/x509check.conf
+++ b/health/health.d/x509check.conf
@@ -1,4 +1,18 @@
 
+# make sure x509check is running
+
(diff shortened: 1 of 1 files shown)


ilyam8:
@cakrit one more approve please :sweat_smile:

PR #6164, opened 2019-05-29
Fix mdstat disks chart

### LINKED ISSUE #6108: "mdstat.mdX_disks" chart's total is misleading
##### Bug report summary
Minor issue here, also likely simple to fix.

Basically, I find the chart "mdstat.mdX_disks" to be misleading: 

To illustrate the way I noticed the issue, in this example, I have a 3 disks RAID 5 array:
netdata raid inuse

"Total" shows 3, in use shows 2 (since one disk is reconstructing).
Problem is on the graph, by looking it fast, it seems like I have 5 disks while I have 3.
I might be too one sided here but it doesn't make sens to me to present it that way. What do you think?

The max value of the chart equals to "total + in use", whereas what would be expected is: The max value is the total.
A better representation could be: the max value is active disks, and negative value would be out of RAID or "not in use" disks. this could be achieved by just subtracting "inuse" value from "total".

##### OS / Environment
OpenMediaVault (Debian 9 x64)

v1.14.0-80-g132e8377

##### Component Name
mdstat.mdx_disks chart

##### Steps To Reproduce
Have any mdstat array with one disk or more, check the chart.
Notice that the chart's height is equal to "in use" disks + "total" disks.

##### Expected behavior
The "total" information is not that useful in the chart.
What I would do is a graph comparable to the network interface chart:
On top in green, you have active drives. On bottom in red, you have out of order drives.
This can be achieved with a simple subtraction.

PS:
BTW I was impressed/surprised/glad to see mdstat info in Netdata in the first place! Thanks again for the good work, keep up!

vlvkobal:
##### Summary
The total number of disks in mdstat.XX_disks chart was displayed incorrectly. As it was suggested, the total number of disks was changed to the number of disks marked "down". Now "inuse" and "down" disks are stacked correctly.

Fixes #6108

##### Component Name
mdstat module in proc plugin

### DIFF
--- a/collectors/proc.plugin/proc_mdstat.c
+++ b/collectors/proc.plugin/proc_mdstat.c
@@ -439,7 +439,7 @@ int do_proc_mdstat(int update_every, usec_t dt) {
 
                 if(unlikely(!raid->rd_inuse && !(raid->rd_inuse = rrddim_find(raid->st_disks, "inuse"))))
                     raid->rd_inuse = rrddim_add(raid->st_disks, "inuse", NULL, 1, 1, RRD_ALGORITHM_ABSOLUTE);
-                if(unlikely(!raid->rd_down && !(raid->rd_down = rrddim_find(raid->st_disks, "total"))))
+                if(unlikely(!raid->rd_down && !(raid->rd_down = rrddim_find(raid->st_disks, "down"))))
                     raid->rd_down = rrddim_add(raid->st_disks, "down", NULL, 1, 1, RRD_ALGORITHM_ABSOLUTE);
 
                 rrddim_set_by_pointer(raid->st_disks, raid->rd_inuse, raid->inuse_disks);
--- a/collectors/proc.plugin/proc_mdstat.c
+++ b/collectors/proc.plugin/proc_mdstat.c
@@ -13,7 +13,7 @@ struct raid {
     unsigned long long failed_disks;
 
     RRDSET *st_disks;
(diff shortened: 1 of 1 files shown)


collectors/proc.plugin/proc_mdstat.c
```diff
@@ -439,11 +439,11 @@ int do_proc_mdstat(int update_every, usec_t dt) {
 
                 if(unlikely(!raid->rd_inuse && !(raid->rd_inuse = rrddim_find(raid->st_disks, "inuse"))))
                     raid->rd_inuse = rrddim_add(raid->st_disks, "inuse", NULL, 1, 1, RRD_ALGORITHM_ABSOLUTE);
-                if(unlikely(!raid->rd_total && !(raid->rd_total = rrddim_find(raid->st_disks, "total"))))
-                    raid->rd_total = rrddim_add(raid->st_disks, "total", NULL, 1, 1, RRD_ALGORITHM_ABSOLUTE);
+                if(unlikely(!raid->rd_down && !(raid->rd_down = rrddim_find(raid->st_disks, "total"))))
```
ilyam8: > (raid->st_disks, "total")
down

vlvkobal: Thank you @ilyam8

ilyam8:
@vlvkobal has you tested it yourself?

vlvkobal:
Not yet. I'll do it before merging.

Issue #6149, opened 2019-05-28
varnish plugin not showing (varnish version 4)


vpapas:
Hello,

After upgrading to netdata v1.15.0-29-ge11a4e4 varnish metrics are not showing.

Here is the netadata error log:

> 2019-05-28 17:59:52: python.d INFO: plugin[checker] : varnish : checking
2019-05-28 17:59:52: python.d INFO: plugin[checker] : varnish : source successfully loaded
2019-05-28 17:59:52: python.d INFO: plugin[checker] : varnish : found config file '/usr/lib/netdata/conf.d/python.d/varnish.conf'
2019-05-28 17:59:52: python.d INFO: plugin[checker] : varnish : created 1 job(s) from the config
2019-05-28 17:59:52: python.d ERROR: varnish[varnish] : No output from 'varnishstat'. Is it running? Not enough privileges?
2019-05-28 17:59:52: python.d INFO: plugin[checker] : varnish[varnish] : check failed, skipping the job
2019-05-28 17:59:52: python.d INFO: plugin[checker] : varnish : all jobs failed, skipping module

The following command is working:

> su -s /bin/bash -c "varnishstat -1" netdata

Also, user netdata is in varnish group.

We are running varnish 4 (varnish-4.0.5-1.el7.x86_64).

ilyam8:
Hi @vpapas 	

https://github.com/netdata/netdata/blob/e11a4e4a5dc5231ffb2f58ff05f32156c052766a/collectors/python.d.plugin/varnish/varnish.chart.py#L198-L202

The output from 'varnishstat -1' is empty for some reason.

Try to run it in debug mod
https://docs.netdata.cloud/collectors/python.d.plugin/#how-to-debug-a-python-module

vpapas:
Hello,

Here is the output of the debug:

> /usr/libexec/netdata/plugins.d/python.d.plugin 1 debug varnish trace
2019-05-28 18:42:33: python.d INFO: plugin[main] : using python v2
2019-05-28 18:42:33: python.d INFO: plugin[main] : starting setup
2019-05-28 18:42:33: python.d INFO: plugin[main] : checking for config in ['/etc/netdata', '/usr/lib/netdata/conf.d']
2019-05-28 18:42:33: python.d INFO: plugin[main] : config found, loading config '/usr/lib/netdata/conf.d/python.d.conf'
2019-05-28 18:42:33: python.d INFO: plugin[main] : config successfully loaded
[...]

PR #6168, opened 2019-05-29
varnish v4 compatibility fix

ilyam8:
##### Summary
Fixes: #6149

`varnishstat` v4 doesnt have `-t` option

##### Component Name
[/collectors/python.d.plugin/varnish](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/varnish)


ilyam8:
ok, i tested it myself, it fixes the problem for varnish v4

PR #6189, opened 2019-06-02
Documentation: Correct example list for python.d SimpleService

kvisle:
MySQL is not using SimpleService, however - ceph is an example that does.


ilyam8:
@kvisle thx for contributing!

PR #6192, opened 2019-06-02
Add note to make smartd directory

Steve8291:
At least on Ubuntu if you do not create this dir manually, all the .csv files get written to /var/lib/smartmontools as the default location.

### DIFF
--- a/collectors/python.d.plugin/smartd_log/README.md
+++ b/collectors/python.d.plugin/smartd_log/README.md
@@ -85,7 +85,8 @@ For this you need to set `smartd_opts` (or `SMARTD_ARGS`, check _smartd.service_
 # dump smartd attrs info every 600 seconds
 smartd_opts="-A /var/log/smartd/ -i 600"
 ```
-
+You may need to create the smartd directory before smartd will write to it:  
+`mkdir -p /var/log/smartd`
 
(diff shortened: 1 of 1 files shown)


Steve8291:
Thanks @cakrit 
Couldn't find any specific info on it on the smartd man page. Just an explanation on the requirement for absolute paths.

cakrit:
> Thanks @cakrit
> Couldn't find any specific info on it on the smartd man page. Just an explanation on the requirement for absolute paths.

No problem, just the first addition then, something like "Otherwise, all the smartd .csv files may get written to `/var/lib/smartmontools` (default location)"

Steve8291:
I think I got it. If not let me know. Still learning my way around github.

cakrit:
No, it didn't make it here.  You need first to commit and then to do `git push origin patch-5`
You can verify in two ways:
- In the "Conversation" tab, you'd see your new commit appearing.
- In the "Files Changed" tab, you'd see the file with the new changes you made.

ilyam8:
> /var/lib/smartmontools as the default location.

i think we need to add a job with `/var/lib/smartmontools` path since it is default.

cakrit:
This change is already in the README. Thanks @Steve8291 !

Issue #6200, opened 2019-06-03
Update fails because of new system dependencies


churnd:
I have netdata installed on several KVM hosts that run Debian 8/9.  I ran the install via `bash <(curl -Ss https://my-netdata.io/kickstart.sh)` about a few months ago.  Over the weekend, they have all been failing to update due to missing system dependencies.  I decided to uninstall & reinstall on one of the hosts to see if it would install more packages that weren't already there.  Turns out, it did bring in several new packages:

```Loading /etc/os-release ...

/etc/os-release information:
NAME            : Debian GNU/Linux
VERSION         : 9 (stretch)
ID              : debian
ID_LIKE         :
VERSION_ID      : 9

We detected these:
Distribution    : debian
Version         : 9
Codename        : 9 (stretch)
Package Manager : install_apt_get
Packages Tree   : debian
Detection Method: /etc/os-release
Default Python v: 2

 > Checking if package 'autoconf-archive' is installed...
 > Checking if package 'zlib1g-dev' is installed...
 > Checking if package 'uuid-dev' is installed...
 > Checking if package 'libmnl-dev' is installed...
 > Checking if package 'libuv1-dev' is installed...
dpkg-query: no packages found matching libuv1-dev
 > Checking if package 'liblz4-dev' is installed...
dpkg-query: no packages found matching liblz4-dev
 > Checking if package 'libssl-dev' is installed...
dpkg-query: no packages found matching libssl-dev
 > Checking if package 'libjudy-dev' is installed...
dpkg-query: no packages found matching libjudy-dev

The following command will be run:

 >> IMPORTANT <<
    Please make sure your system is up to date
    by running:   apt-get update

apt-get install libjudy-dev
apt-get install liblz4-dev
apt-get install libssl-dev
apt-get install libuv1-dev


Press ENTER to run it >
```

What is the proper procedure for handling new system requirements when netdata needs them?  I'm assuming the auto-update feature doesn't do this automatically (which is good IMO).  The updates were all failing due to the missing libssl-dev:

https://gist.github.com/churnd/496e0556d5fc442f001416b2320a5070

cakrit:
The info on how to install the required packages is [here](https://docs.netdata.cloud/packaging/installer/#prepare-your-system). 

`kickstart.sh` also has the following installer options:
```
#  --dont-wait       do not prompt for user input
#  --non-interactive do not prompt for user input
```

[...]

Issue #6102, opened 2019-05-23
Introduce cross-distro release testing on our CI 


paulkatsoulakis:
##### Feature idea summary
* Prepare the necessary docker images for each distribution we support
* Implement basic tests in travis

##### Expected behavior
Have our netdata code run on all supported distributions

PR #6209, opened 2019-06-04
netdata/packaging/ci: Add lifecycle checks to bare operating system installations

paulkatsoulakis:
##### Summary
As part of #6102 we had to introduce more cross-distribution runs, to guarantee that we don't allow failed code reaching our users.
On top of that, through #6200 we realised that our current CI infrastructure did not provide enough visibility on possible issues coming from missing package dependencies.
Until now we used our own pre-compiled docker images to run our installers, which was hiding possible issues like the one on #6200.

Hopefully  with this PR we put an end to such issues and we get closer in the cross-distro builds

##### Component Name
netdata/packaging

##### Additional Information
References #6102

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,7 +141,7 @@ jobs:
   - stage: "Artifacts validation on bare OS, stable to current lifecycle checks"
 
     name: Run netdata lifecycle on Ubuntu 16.04 (xenial)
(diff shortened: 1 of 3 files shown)


paulkatsoulakis:
Merged after tests passed as agreed with @cakrit

PR #6212, opened 2019-06-04
Add code style guidance to CONTRIBUTING

cakrit:
##### Summary
Add basic code style guidelines

##### Component Name
docs

##### Additional Information
Based on internal discussion on 4-Jun-19


thiagoftsm:
I agree 100% with everything written in the file.

ilyam8:
> Do not unnecessarily use braces where a single statement will do.

> Thus, as the supply of new-lines on your screen is not a renewable resource (think 25-line terminal screens here), you have more empty lines to put comments on.

very weak point imo

PR #6220, opened 2019-06-05
Backend and SSL!

### LINKED ISSUE #1642: Support OpenTSDB back-end via HTTP(S) API
I'd love to use `netdata` as a replacement metric collector to `tcollector` in my production environment, shipping metrics to an OpenTSDB cluster, while also gaining the benefit of better real-time metrics for my users.

Today `netdata` only supports the OpenTSDB telnet API for submitting metrics, which offers no encryption, authentication, or verification. That's a non-starter for any production environment, especially when the archival metric cluster is on the other side of a network boundary.

Supporting the HTTP(S) OpenTSDB API would greatly increase `netdata` traction in the metrics community.

thiagoftsm:
##### Summary
Fixes #1642 

##### Component Name
backend directory.

##### Additional Information
After to do encryption of the web server and streams, we are beginning to work on the backend side, we started with Opentsdb


thiagoftsm:
To test communication between Netdata and OpenTSDB using HTTP(S), it will be necessary to install the following software:

- Apache HBASE : The database used to organize the data. This is a Java server, I used the binary given by Apache.
- Apache Zookeeper : Necessary to OpenTSDB. I used a Slackbuild script to build it.
- OpenTSDB: This I cloned from the github and I executed the following steps to compile using(  ./build.sh ).

To start  the tests it is necessary to do the following steps
- I started HBASE:
$ /usr/local/hbase/bin/start-hbase.sh
- I started Zookeper: 
# sh /etc/rc.d/rc.zookeeper start
- I created the database using: 
$ env COMPRESSION=NONE HBASE_HOME=/usr/local/hbase/ ./src/create_table.sh
- Finally I ran the OpenTSDB
./build/tsdb tsd --port=4242 --staticroot=build/staticroot --cachedir=/tmp/tsd --zkquorum=localhost:2181 --auto-metric 

An important thing, case we do not start OpenTSDB with "--auto-metric" it will fail to store the Netdata data. 
To set netdata.conf, please, read the documentation attached with this PR.

backends/backends.c
```diff
@@ -719,6 +781,21 @@ void *backends_main(void *ptr) {
...
+                    if(!opentsdb_ssl.flags) {
+                        written = SSL_write(opentsdb_ssl.conn,buffer_tostring(b), len);
+                    } else {
+                        written = send(sock, buffer_tostring(b), len, flags);
+                    }
+                } else {
+                    written = send(sock, buffer_tostring(b), len, flags);
+                }
+#else
+                written = send(sock, buffer_tostring(b), len, flags);
```
vlvkobal: 1. You can't move it here. It should be after the `ENABLE_PROMETHEUS_REMOTE_WRITE` block.
2. Please get rid of `written = send(sock, buffer_tostring(b), len, flags);` 3 times in a row

backends/opentsdb/README.md
```diff
@@ -0,0 +1,72 @@
+# OpenTSDB with HTTP
+
+Since the version 1.16 the Netdata has the feature to communicate with OpenTSDB using HTTP API, to enable this channel
+it is necessary to set the following options in your netdata.conf
+
```
vlvkobal: Since the version 1.16 the Netdata -> Since version 1.16 Netdata

HTTP API, to -> HTTP API. To

backends/opentsdb/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+[backend]
+    type = opentsdb:http
+    destination = localhost:4242
+```
+
+, in this example we are considering that OpenTSDB is running with its default port(4242).
```
vlvkobal: , in -> In

port(4242) -> port (4242)

backends/opentsdb/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## HTTPS
+
+Netdata also supports to send the metrics using SSL/TLS, this is the preferred option, because it is the safest, but OpenTDSB
```
vlvkobal: supports to send the metrics -> supports sending metrics

Too many commas. Split it to shorter sentences.

backends/opentsdb/opentsdb.c
```diff
@@ -87,4 +87,118 @@ int process_opentsdb_response(BUFFER *b) {
...
+    int length =  snprintfz(message
+                           , sizeof(message)
+                           ,"{"
+                            "\"metric\": \"%s.%s.%s\","
+                            "\"timestamp\": %llu,"
+                            "\"value\": "COLLECTED_NUMBER_FORMAT ","
+                            "\"tags\": {"
+                            "    \"host\": \"%s%s%s\""
+                            "  }"
+                            "}"
```
vlvkobal: Inconsistent indentation. Missing leading spaces for the third argument.

backends/opentsdb/opentsdb.c
```diff
@@ -87,4 +87,118 @@ int process_opentsdb_response(BUFFER *b) {
...
+        int length =  snprintfz(message
+                , sizeof(message)
+                ,"{"
+                 "\"metric\": \"%s.%s.%s\","
+                 "\"timestamp\": %llu,"
+                 "\"value\": "CALCULATED_NUMBER_FORMAT ","
+                 "\"tags\": {"
+                 "    \"host\": \"%s%s%s\""
+                 "   }"
+                 "}"
```
vlvkobal: Inconsistent indentation. Missing leading spaces for the third argument.

backends/opentsdb/opentsdb.c
```diff
@@ -87,4 +87,118 @@ int process_opentsdb_response(BUFFER *b) {
...
+                , dimension_name
+                , (unsigned long long)last_t
+                , value
+                , hostname
+                , (host->tags)?" ":""
+                , (host->tags)?host->tags:""
+        );
+
+        if(length > 0) {
+            opentsdb_build_message(b, message, hostname,length);
```
vlvkobal: Missing space after a comma

backends/opentsdb/opentsdb.c
```diff
@@ -87,4 +87,118 @@ int process_opentsdb_response(BUFFER *b) {
...
+        }
+
+        return 1;
+    }
+
+    return 0;
+}
```
vlvkobal: Missing newline

backends/backends.c
```diff
@@ -301,6 +301,10 @@ void *backends_main(void *ptr) {
 
     global_backend_options = backend_parse_data_source(source, global_backend_options);
 
+#ifdef ENABLE_HTTPS
+    struct netdata_ssl opentsdb_ssl = { NULL , NETDATA_SSL_START};
```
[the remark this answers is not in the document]
vlvkobal: What is the logic of placing the struct declaration here? It looks like a randomly chosen place for me.

backends/backends.c
```diff
@@ -334,7 +338,22 @@ void *backends_main(void *ptr) {
...
         else
             backend_request_formatter = format_dimension_stored_opentsdb_telnet;
+    }
+    else if(!strcmp(type, "opentsdb:http") || !strcmp(type, "opentsdb:https")) {
+
+        default_port = 4242;
+        backend_response_checker = process_opentsdb_response;
+
+#ifdef ENABLE_HTTPS
+        if (!strcmp(type+12,"ps")) {
```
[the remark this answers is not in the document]
vlvkobal: You've already compared the full strings. Why don't you use a variable instead of such cryptic string comparison?

backends/backends.c
```diff
@@ -668,7 +687,20 @@ void *backends_main(void *ptr) {
...
+                        if(!opentsdb_ssl.flags) {
+                            r = SSL_read(opentsdb_ssl.conn,&response->buffer[response->len], response->size - response->len);
+                        } else {
+                            r = recv(sock, &response->buffer[response->len], response->size - response->len, MSG_DONTWAIT);
+                        }
+                    } else {
+                        r = recv(sock, &response->buffer[response->len], response->size - response->len, MSG_DONTWAIT);
+                    }
+#else
                     ssize_t r = recv(sock, &response->buffer[response->len], response->size - response->len, MSG_DONTWAIT);
```
vlvkobal: 3 * `r = recv(sock, &response->buffer[response->len], response->size - response->len, MSG_DONTWAIT);`

vlvkobal: error: redeclaration of ‘r’ with no linkage

libnetdata/socket/security.c
```diff
@@ -127,7 +128,11 @@ void security_start_ssl(int type) {
         netdata_srv_ctx =  security_initialize_openssl_server();
     }
     else {
-        netdata_cli_ctx =  security_initialize_openssl_client();
+        if(type == 1) {
+            netdata_cli_ctx = security_initialize_openssl_client();
+        } else if (type == 2) {
```
vlvkobal: Magic number?

vlvkobal: `switch` would be easier to read than a bunch of `if...else`

thiagoftsm: It is not a magic number exactly, it was an order defined during  the development, firstly we developed the server(0), the next was the stream(1) and finally we began to work with backends, and the first backend is opentsdb(2);

For this specific case that we have only 3, I agree that tje switch can be used, but this is not always the case.

vlvkobal:
It is an old code, but wouldn't it be more clear if you rename `netdata_cli_ctx` -> `netdata_client_ctx`?

libnetdata/socket/security.c
paulkatsoulakis: This does look like a magic number indeed if i may say

thiagoftsm:
> It is an old code, but wouldn't it be more clear if you rename `netdata_cli_ctx` -> `netdata_client_ctx`?

I will create a new PR for this request, because it does not have direct relationship with the backend, this variable is a related to the stream.

libnetdata/socket/security.c
vlvkobal: I meant that it would be better to use macros instead of 0, 1, 2.

backends/README.md
```diff
@@ -76,7 +76,7 @@ of `netdata.conf` from your netdata):
 ```
 [backend]
     enabled = yes | no
-    type = graphite | opentsdb | json | prometheus_remote_write | kinesis
```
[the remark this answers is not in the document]
thiagoftsm: It was wrong edition, I did not work with prometheus during the whole process. I apologize by this!

backends/backends.c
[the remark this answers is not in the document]
thiagoftsm: I followed a conclusion that I arrived after to study the function backends_main(), the majority of the variables are declared before the main loop, so I decided to put this variable in this position. It could be nearest the main loop without any affect  to the code, so I can say that it is "almost" random.

backends/backends.c
```diff
@@ -325,7 +329,7 @@ void *backends_main(void *ptr) {
             backend_request_formatter = format_dimension_stored_graphite_plaintext;
 
     }
-    else if(!strcmp(type, "opentsdb") || !strcmp(type, "opentsdb:telnet")) {
+    else if(!strcmp(type, "opentsdb") || !strcmp(type, "opentsdb:telnet") ) {
```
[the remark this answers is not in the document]
ilyam8: How about to add move `get backend type` functionality to a separate func, use it, and switch on result?

backends/backends.c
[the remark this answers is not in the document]
thiagoftsm: It is necessary to check the necessity to start the ssl environment for this group. I agree that I was creating not necessary opcodes so I changed the code for we only test the last position of the pointer.

backends/backends.c
```diff
@@ -668,7 +687,20 @@ void *backends_main(void *ptr) {
                 while(sock != -1 && errno != EWOULDBLOCK) {
                     buffer_need_bytes(response, 4096);
 
+                    ssize_t r;
+#ifdef ENABLE_HTTPS
+                    if(opentsdb_ssl.conn) {
+                        if(!opentsdb_ssl.flags) {
+                            r = SSL_read(opentsdb_ssl.conn,&response->buffer[response->len], response->size - response->len);
```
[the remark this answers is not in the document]
thiagoftsm: it was fixed!

backends/opentsdb/README.md
thiagoftsm: Fixed!

backends/opentsdb/opentsdb.c
thiagoftsm: This is also missed in the telnet functions, do we need to adjust them too?

backends/backends.c
thiagoftsm: I created an PR previously for this error, and i do not have more it here!

backends/backends.c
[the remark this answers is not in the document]
thiagoftsm: It is an awesome idea Ilya, I will do this right now!

backends/backends.c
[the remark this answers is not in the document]
ilyam8: It will be an easy check after `get backend type` (or whatever name itll have) function will be implemented

backends/backends.c
```diff
@@ -246,6 +246,114 @@ static void backends_main_cleanup(void *ptr) {
...
+ * @param backend_request_formatter function called to format the msessage to the backend
+ * @param type the backend string selector.
+ *
+ * @return It returns 0 case there is not necessity to set additional variables, 1 case it is necessary to adjust prometheus
+ *          2 case it is necessary to adjust kinesis and -1 on error
+ */
+int backend_set_type(int *default_port
+                     ,int (**backend_response_checker)(BUFFER *)
+                     ,int (**backend_request_formatter)(BUFFER *, const char *, RRDHOST *, const char *, RRDSET *, RRDDIM *, time_t, time_t, BACKEND_OPTIONS)
+                     ,const char *type) {
```
ilyam8: Hey @thiagoftsm 	

This is nice, but i think it would be better to move `type` parsing logic outside the func, it will simplify things

something like this

```c
enum backend_type {
    UNKNOWN,
    TYPE1,
    TYPE2,
};

enum backend_type parse_backend_type(const char *type);

int is_backend_type_https(enum backend_type);
/*
use it^^ instead magic thing `*(type+13) == 's'`
*/
```

After it you have simple `case ...switch` inside your `backend_set` function. 

Will this work? What do you think?

thiagoftsm: Hi @ilyam8 , I agree this will work and it also looks like a good idea, but let me explain firstly the motive I wrote this way. As you can see I moved inside the function the common algorithm, but I was forced to use an integer return, because we have prometheus and kinesis doing more than the common. Case we move type out of the function, we would test the results to have the value compatible with would not avoid the treatment of not common variables in another place, no less important in the way that I wrote I do not need to test again with an "if/else if" or switch the result to define the common variables. If I understood correct, the current algorithm tests only two times to define the not common for the "normal" case, while the enum suggestion could force us to test 1 more time to let the code more readble.

ilyam8: I can only repeat - do not optimize things until there is a reason for it. If you are in doubt do benchmark and see if it worth. Otherwise prefer readable over unreadable.

i am sure - case on enum is super mega fast and it costs nothing.

> but I was forced to use an integer return, because we have prometheus and kinesis doing more than the common

That means that something is wrong there and the function a candidate for refactoring.

___
That is how i think:
 - i see function name 'do_this'
 - i read function and it `does this, does that and that`

> backend_set_type

The function is not responsible for parsing, it only sets. Parse loginc should be outside. Or name it `backend_parse_type_and_set`. Looks bad, right? :smile:

backends/backends.c
```diff
@@ -247,102 +247,187 @@ static void backends_main_cleanup(void *ptr) {
...
+    *default_port = 4242;
+    *backend_response_checker = process_opentsdb_response;
 
 #ifdef ENABLE_HTTPS
-        if ( *(type+13) == 's') {
-            security_start_ssl(2);
-        }
+    if ( *(type+13) == 's') {
+        security_start_ssl(2);
+    }
```
ilyam8: i think `const char *type is not needed in all `backend_set_`, i see you need it only (so far) in `backend_set_opentsdb_telnet_variables.

How about to add `BACKEND_TYPE_OPENTSDB_HTTPS` type? it will solv the problem

backends/backends.c
```diff
@@ -246,6 +246,199 @@ static void backends_main_cleanup(void *ptr) {
...
+ * Set the variables necessaries to work with this specific backend.
+ *
+ * @param default_port  the default port of the backend
+ * @param backend_response_checker function called to check the result.
+ * @param backend_request_formatter function called to format the msessage to the backend
+ * @param type the backend string selector.
+ */
+void backend_set_kinesis_variables(int *default_port
+        ,int (**backend_response_checker)(BUFFER *)
+        ,int (**backend_request_formatter)(BUFFER *, const char *, RRDHOST *, const char *, RRDSET *, RRDDIM *, time_t, time_t, BACKEND_OPTIONS)
```
ilyam8: please use `typedef`, it is much clear then `int (**backend_request_formatter)(BUFFER *, const char *, RRDHOST *, const char *, RRDSET *, RRDDIM *, time_t, time_t, BACKEND_OPTIONS)` in every func

backends/backends.c
thiagoftsm: I think it is not necessary, when you did this question you gave me an idea, give me few minutes and I will upload a new version!

backends/backends.c
thiagoftsm: The newest version will also bring the typedef!

backends/backends.c
```diff
@@ -425,32 +510,43 @@ void *backends_main(void *ptr) {
...
+
+    void (*backend_fp[]) (int *,int (**brc)(BUFFER *),int (**bqf)(BUFFER *, const char *, RRDHOST *, const char *, RRDSET *, RRDDIM *, time_t, time_t, BACKEND_OPTIONS),const char *) = {
+            backend_set_graphite_variables
+            ,backend_set_opentsdb_telnet_variables
+            ,backend_set_opentsdb_http_variables
+            ,backend_set_json_variables
+            ,backend_set_prometheus_variables
+            ,backend_set_kinesis_variables
+    };
+    backend_fp[work_type](&default_port,&backend_response_checker,&backend_request_formatter,type);
```
ilyam8: This is very error prone to rely on index and order. Just make it a function - a function that returns appropriate `backend_set_` function based on `type`

backends/backends.c
ilyam8: ok, but please don;t use `char *type` in set funcs, it is wrong, only `backend_select` should use it

paulkatsoulakis: I am not sure if it is clear what does const mean when declared on input parameters of a method.  It does not imply visibility, it rather implies value can't be changed.
Check [here](https://en.wikipedia.org/wiki/Const_(computer_programming)#Parameters_and_variables) and you 'll see:
```A const automatic variable (non-static local variable) means that single assignment is happening, though a different value may be used each time, such as const int x_squared = x*x. A const parameter in pass-by-reference means that the referenced value is not modified – it is part of the contract – while a const parameter in pass-by-value (or the pointer itself, in pass-by-reference) does not add anything to the interface (as the value has been copied), but indicates that internally, the function does not modify the parameter (it is a single assignment). For this reason, some favor using const in parameters only for pass-by-reference, where it changes the contract, but not for pass-by-value, where it exposes the implementation.```

File: backends/backends.c (hunk 19 above)
Comment (thiagoftsm) [reply backends/backends.c]: I am sorry, I did not understand the  "error prone" cited, this is a function pointer and I am calling a function based in the index that I received from the function type. We won't have error, because when the software reaches this point, it will have only valid values to call the function. This is not much different of what we have in the file netdata/web/api/web_api_v1.c, the unique difference is the fact that I did not the struct api_commands to call, I called it directly.

File: backends/backends.c (hunk 17 above)
Comment (paulkatsoulakis) [reply backends/backends.c]: At this specific line of code, i would only indicate the magic numbers that don't help code readability.

File: backends/backends.c (hunk 19 above)
Comment (ilyam8) [reply backends/backends.c]: Correct me if i am wrong - if you use `enum` you should **never** rely on `enum` value and on `enum` order.
 
That is what you are doing here.

If you change the order or value in the `enum` your code doesn't work. If you change the order in your array of funcs your code doesnt work again.

Comment (thiagoftsm) [reply backends/backends.c]: I could understand better now your point now, so let me explain better what I did. I am using the enum as a kind of variable that limit the possible value that is accepted, where the first value means error and the other values are valid index, so there is not any problem to use the value instead the name, because they are valid index. I understand that in object orientation languages this is not normal. 
I am bringing with this new fix a possible solution that helps to avoid developer errors, because the developer will need to change more places now, but as you can see I am also using index number instead names. Case we do not agree with this solution, we will be forced to use names instead number, but this will create a function, that is not necessary, but it will be more readable.

Comment (ilyam8) [reply backends/backends.c]: > but this will create a function, that is not necessary

i think it is, but i am not insist on it, dont create it ok, but please change the implementation - use `if/else` or (better) `switch` to set the setter func instead of creating array and using enum as index.

Comment (ilyam8) [reply backends/backends.c]: i think connecting array order with enam actual values is not good, lets ask @mfundul @vlvkobal. Maybe it is false alarm from me

Comment (thiagoftsm) [reply backends/backends.c]: No it was not, I understood your point, because developers can write wrong things and your point of view helped me to write a structure that at least force the user to think more one time before to invert anything.
Firstly I will explain the computer side about "function is not necessary", when I said this, I was considering that I already have everything to call the function, so to create another step is not necessary, on the other hand, when we see the human side, I agree that it would be more clear to have a switch with all the options. No less important,  I agree with you let us bring @mfundul and @vlvkobal to explain their opinion, everybody will grow with this and the software will be optimized, with more resources and of course it will be better!

File: backends/backends.c (hunk 9 above)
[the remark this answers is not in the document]
Comment (vlvkobal) [reply backends/backends.c]: I'd move it [here](https://github.com/netdata/netdata/blob/f2d5c1b7ee36574e3c8fd4cea3d86ab325e28792/backends/backends.c#L458).

File: backends/backends.c [hunk 20]
```diff
@@ -246,6 +246,207 @@ static void backends_main_cleanup(void *ptr) {
...
+ *
+ * @param default_port  the default port of the backend
+ * @param brc function called to check the result.
+ * @param brf function called to format the msessage to the backend
+ * @param type the backend string selector.
+ */
+void backend_set_kinesis_variables(int *default_port
+        ,backend_response_checker_t brc
+        ,backend_request_formatter_t brf
+        ) {
```
vlvkobal: Please do better formatting for functions. Something like [this](https://github.com/netdata/netdata/blob/master/database/engine/rrdengine.c#L133).

backends/opentsdb/opentsdb.c
vlvkobal: Yes.

libnetdata/socket/security.c
vlvkobal: I think the following code is easier to read:
```c
void security_start_ssl(int type) {
    switch (type) {
        case NETDATA_SSL_CONTEXT_SERVER:
            {
                struct stat statbuf;

                if (stat(security_key, &statbuf) || stat(security_cert, &statbuf)) {
                    info("To use encryption it is necessary to set \"ssl certificate\" and \"ssl key\" in [web] !\n");
                    return;
                }
            }
            netdata_srv_ctx =  security_initialize_openssl_server();
            break;
        case NETDATA_SSL_CONTEXT_STREAMING:
            netdata_cli_ctx = security_initialize_openssl_client();
            break;
        case NETDATA_SSL_CONTEXT_OPENTSDB:
            netdata_opentsdb_ctx = security_initialize_openssl_client();
            break;
    }
}
```

backends/backends.c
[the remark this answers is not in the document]
thiagoftsm: All right, it will be in this line in the next commit.

backends/backends.c
thiagoftsm: It was also changed for the next commit.

backends/opentsdb/opentsdb.c
thiagoftsm: I adjusted. This was an error that happened, because I copied and paste the function that has this missing space. It is also fixed for the next.

libnetdata/socket/security.c
thiagoftsm: After the talk on Friday night, I was convinced that I did not need to avoid more the "switch", definitively the compilers improved a lot since the first time I executed a reverse engineer and  I compared both "if ... else if " with "swtich".

thiagoftsm:
> > It is an old code, but wouldn't it be more clear if you rename `netdata_cli_ctx` -> `netdata_client_ctx`?
> 
> I will create a new PR for this request, because it does not have direct relationship with the backend, this variable is a related to the stream.

PR #6336 was created to improve the name

vlvkobal:
```
backends/backends.c: In function 'backend_set_kinesis_variables':
backends/backends.c:270:6: error: 'backend_response_checker' undeclared (first use in this function); did you mean 'backend_response_checker_t'?
  270 |     *backend_response_checker = process_json_response;
      |      ^~~~~~~~~~~~~~~~~~~~~~~~
      |      backend_response_checker_t
backends/backends.c:270:6: note: each undeclared identifier is reported only once for each function it appears in
backends/backends.c:272:10: error: 'backend_request_formatter' undeclared (first use in this function); did you mean 'backend_request_formatter_t'?
  272 |         *backend_request_formatter = format_dimension_collected_json_plaintext;
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
      |          backend_request_formatter_t
backends/backends.c:260:63: warning: unused parameter 'brc' [-Wunused-parameter]
  260 |                                    backend_response_checker_t brc,
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
backends/backends.c:261:64: warning: unused parameter 'brf' [-Wunused-parameter]
  261 |                                    backend_request_formatter_t brf)
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
backends/backends.c: At top level:
backends/backends.c:432:25: warning: missing braces around initializer [-Wmissing-braces]
  432 | } backend_commands[]  = {
      |                         ^
  433 |         backend_set_graphite_variables, 0
      |         {
  434 |         ,backend_set_opentsdb_telnet_variables, 1
      |         }{
  435 |         ,backend_set_opentsdb_http_variables, 2
      |         }{
  436 |         ,backend_set_json_variables, 3
      |         }{
  437 |         ,backend_set_prometheus_variables, 4
      |         }{
  438 |         ,backend_set_kinesis_variables, 5
      |         }{
  439 | };
      | }
backends/backends.c: In function 'backends_main':
backends/backends.c:531:5: warning: enumeration value 'BACKEND_TYPE_UNKNOWN' not handled in switch [-Wswitch]
  531 |     switch (work_type) {
      |     ^~~~~~
backends/backends.c:531:5: warning: enumeration value 'BACKEND_TYPE_GRAPHITE' not handled in switch [-Wswitch]
backends/backends.c:531:5: warning: enumeration value 'BACKEND_TYPE_OPENTSDB_TELNET' not handled in switch [-Wswitch]
backends/backends.c:531:5: warning: enumeration value 'BACKEND_TYPE_JSON' not handled in switch [-Wswitch]
make[2]: *** [Makefile:2937: backends/backends.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/home/vlad/test/netdata'
make[1]: *** [Makefile:3059: all-recursive] Error 1
make[1]: Leaving directory '/home/vlad/test/netdata'
make: *** [Makefile:1716: all] Error 2
 FAILED

```

vlvkobal:
~14 requests are still unresolved.

backends/opentsdb/opentsdb.c
thiagoftsm: For we send the JSON compact and readable, I began to use two tabs indentation, as you will see in the in the next commit.

backends/opentsdb/opentsdb.c
thiagoftsm: For we send the JSON compact and readable, I began to use two tabs indentation, as you will see in the in the next commit.

backends/opentsdb/opentsdb.c
thiagoftsm: This was already fixed.

backends/opentsdb/opentsdb.c
thiagoftsm: I saw that graphite also has this problem, so probably we will need to review the other backends.

backends/backends.c
thiagoftsm: I fixed it now!

thiagoftsm:
> ```
> backends/backends.c: In function 'backend_set_kinesis_variables':
> backends/backends.c:270:6: error: 'backend_response_checker' undeclared (first use in this function); did you mean 'backend_response_checker_t'?
>   270 |     *backend_response_checker = process_json_response;
>       |      ^~~~~~~~~~~~~~~~~~~~~~~~
>       |      backend_response_checker_t
> backends/backends.c:270:6: note: each undeclared identifier is reported only once for each function it appears in
> backends/backends.c:272:10: error: 'backend_request_formatter' undeclared (first use in this function); did you mean 'backend_request_formatter_t'?
>   272 |         *backend_request_formatter = format_dimension_collected_json_plaintext;
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~
>       |          backend_request_formatter_t
> backends/backends.c:260:63: warning: unused parameter 'brc' [-Wunused-parameter]
>   260 |                                    backend_response_checker_t brc,
>       |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
> backends/backends.c:261:64: warning: unused parameter 'brf' [-Wunused-parameter]
>   261 |                                    backend_request_formatter_t brf)
>       |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
> backends/backends.c: At top level:
> backends/backends.c:432:25: warning: missing braces around initializer [-Wmissing-braces]
>   432 | } backend_commands[]  = {
>       |                         ^
>   433 |         backend_set_graphite_variables, 0
>       |         {
>   434 |         ,backend_set_opentsdb_telnet_variables, 1
>       |         }{
>   435 |         ,backend_set_opentsdb_http_variables, 2
>       |         }{
>   436 |         ,backend_set_json_variables, 3
>       |         }{
>   437 |         ,backend_set_prometheus_variables, 4
>       |         }{
>   438 |         ,backend_set_kinesis_variables, 5
>       |         }{
>   439 | };
>       | }
> backends/backends.c: In function 'backends_main':
> backends/backends.c:531:5: warning: enumeration value 'BACKEND_TYPE_UNKNOWN' not handled in switch [-Wswitch]
>   531 |     switch (work_type) {
>       |     ^~~~~~
> backends/backends.c:531:5: warning: enumeration value 'BACKEND_TYPE_GRAPHITE' not handled in switch [-Wswitch]
> backends/backends.c:531:5: warning: enumeration value 'BACKEND_TYPE_OPENTSDB_TELNET' not handled in switch [-Wswitch]
> backends/backends.c:531:5: warning: enumeration value 'BACKEND_TYPE_JSON' not handled in switch [-Wswitch]
> make[2]: *** [Makefile:2937: backends/backends.o] Error 1
> make[2]: *** Waiting for unfinished jobs....
> make[2]: Leaving directory '/home/vlad/test/netdata'
> make[1]: *** [Makefile:3059: all-recursive] Error 1
> make[1]: Leaving directory '/home/vlad/test/netdata'
> make: *** [Makefile:1716: all] Error 2
>  FAILED
> ```

I fixed the problem.
And of course, I had forgotten this, but thanks to compile and share, when we use "if..else if" we do not need to populate code with options that do not have nothing to process! ;)

thiagoftsm:
> ~14 requests are still unresolved.

I reviewed option by option and I could not find 14 unresolved,  but I resolved the newest and the others that I missed,. I followed the democracy and I am using switch instead function pointer to call the functions, but I am continue considering the function pointer the best option!

backends/backends.c
```diff
@@ -313,74 +505,63 @@ void *backends_main(void *ptr) {
...
-        backend_response_checker = process_opentsdb_response;
-
-        if(BACKEND_OPTIONS_DATA_SOURCE(global_backend_options) == BACKEND_SOURCE_DATA_AS_COLLECTED)
-            backend_request_formatter = format_dimension_collected_opentsdb_telnet;
-        else
-            backend_request_formatter = format_dimension_stored_opentsdb_telnet;
-
+    BACKEND_TYPE work_type = backends_select_type(type);
+    if (work_type == BACKEND_TYPE_UNKNOWN) {
+        goto cleanup;
```
ilyam8: i think you need to log error here (and include `type` into it)

thiagoftsm: Ok, I will prepare this and when Travis and LGTM finish their job, I will upload the code again with possible fix in the results of the two cited tools too.

backends/backends.h
```diff
@@ -15,6 +15,22 @@ typedef enum backend_options {
...
 } BACKEND_OPTIONS;
 
+typedef enum backend_types {
+    BACKEND_TYPE_UNKNOWN = -1,
+    BACKEND_TYPE_GRAPHITE = 0,
+    BACKEND_TYPE_OPENTSDB_TELNET = 1,
+    BACKEND_TYPE_OPENTSDB_HTTP = 2,
+    BACKEND_TYPE_JSON = 3,
+    BACKEND_TYPE_PROMETEUS = 4,
+    BACKEND_TYPE_KINESIS = 5
```
ilyam8: Actual values are not used anymore, i suggest to remove -1, 0 stuff.

And one more thing about it, i think you need to add protocl suffix to all types, otherwise it is not clear why ony `opentsdb` has `http/telnet`

backends/backends.c
ilyam8: >  when Travis and LGTM finish their job

it will take a while :smile: 	i would not wait :smile:

thiagoftsm: :D 
Only to complement, I saw that I already had logged inside the function "backend_select_type", but I am moving from there to the position that you specified.

ilyam8: oh, sorry, you are right. But yes, log it right after `backends_select_type` is better then inside the func.

btw i would suggest ot rename

`backends_select_type` -> `backend_select_type`

backends/backends.h
thiagoftsm: All right I already changed here, so I am removing the values and allowing the compiler to give the value to them.

About the protocol suffix, I confess I did not study all the other backends, so I will rename opentsdb from:

BACKEND_TYPE_OPENTSDB_TELNET,
BACKEND_TYPE_OPENTSDB_HTTP,

to

BACKEND_TYPE_OPENTSDB_USING_TELNET,
BACKEND_TYPE_OPENTSDB_USING_HTTP,

I think this will help to let the things more clear.

backends/backends.c
ilyam8: i edited the comment, see it on github

thiagoftsm: I accepted the suggestion and I rename, but this created a problem. As you can see in this file, there are functions not created by me, that begins with "backends_", so we have to create a common pattern for the whole file. Let us wait other opinions for I change the functions and we have only one pattern.

backends/backends.h
ilyam8: I think better to add comments in this particular case to make it more clear. But i like the direction in general - to make something clear :smile: 

so 
> TYPE_OPENTSDB_HTTP

lgtm, it is clear from the context what is `HTTP` there.

ilyam8: ok, you added `USING` thing, the problem is not resolved - it is completely unclear why only types have `USING_PROTOCOL`

thiagoftsm: All right, I will comment all the options to have the things more clear.
I already commit everything we talked, except this! :)

backends/backends.c
ilyam8: well, i see `backend_set_opentsdb_http_variables`... :smile: 	

Ok, if there is a pattern, then use it, i was not aware

backends/opentsdb/opentsdb.c
vlvkobal: Cool! JSON is more readable now, indeed 👍. But what I meant was the parameter alignment:
```
                           , sizeof(message)
                           , "{"
                             "  \"metric\": \"%s.%s.%s\","
                             "  \"timestamp\": %llu,"
                             "  \"value\": "COLLECTED_NUMBER_FORMAT ","
                             "  \"tags\": {"
                             "    \"host\": \"%s%s%s\""
                             "  }"
                             "}"
                           , prefix
```
Look at the parameters `"..."`. I shifted them to the right by one space.

backends/opentsdb/opentsdb.c
vlvkobal: Look at [line 148](https://github.com/netdata/netdata/pull/6220/files#diff-dce11c7d7d759221988833b869b7d6e9R148).

backends/opentsdb/opentsdb.c
vlvkobal: Just put one `Enter` to the end of the file.

backends/opentsdb/opentsdb.c
thiagoftsm: All right, now I understood completely what you were saying! I copied you adjusts for the next pull!

backends/opentsdb/opentsdb.c
thiagoftsm: I forgotten this code is present in two different places in this file. I fixed it now.

PR #6222, opened 2019-06-05
netdata/daemon: Eliminate a couple of warnings, plus tabs removal

paulkatsoulakis:
##### Summary
This PR fixes style issues on a file (stray tabs while the whole file is indented with spaces).
It also attends two warnings seen during compilation, due to the fact some part of the code being hidden within ifdefs. 

##### Component Name
netdata/daemon

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -742,13 +742,13 @@ void send_statistics( const char *action, const char *action_result, const char 
         if (likely(access(optout_file, R_OK) != 0)) {
             as_script = mallocz(sizeof(char) * (strlen(netdata_configured_primary_plugins_dir) + strlen("anonymous-statistics.sh") + 2));
             sprintf(as_script, "%s/%s", netdata_configured_primary_plugins_dir, "anonymous-statistics.sh");
-                        if (unlikely(access(as_script, R_OK) != 0)) {
-                                netdata_anonymous_statistics_enabled=0;
-                                info("Anonymous statistics script %s not found.",as_script);
-                                freez(as_script);
-                         } else {
-                                netdata_anonymous_statistics_enabled=1;
-                         }
+            if (unlikely(access(as_script, R_OK) != 0)) {
+               netdata_anonymous_statistics_enabled=0;
+               info("Anonymous statistics script %s not found.",as_script);
+               freez(as_script);
+            } else {
+               netdata_anonymous_statistics_enabled=1;
+            }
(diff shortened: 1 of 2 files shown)


daemon/main.c
```diff
@@ -742,20 +742,20 @@ void send_statistics( const char *action, const char *action_result, const char
...
             sprintf(as_script, "%s/%s", netdata_configured_primary_plugins_dir, "anonymous-statistics.sh");
-			if (unlikely(access(as_script, R_OK) != 0)) {
-				netdata_anonymous_statistics_enabled=0;
-				info("Anonymous statistics script %s not found.",as_script);
-				freez(as_script);
-			} else {
-				netdata_anonymous_statistics_enabled=1;
-			}
-		} else {
+                        if (unlikely(access(as_script, R_OK) != 0)) {
```
ilyam8: why additional indentation for that if/else?

thiagoftsm: I think I know the problem, because I had something like this, the Clion uses spaces to do tabulation instead tab, so this can happen because the writer were using tabs or the computer uses 8 spaces instead 4 to tabulation. I had problem with the second.

paulkatsoulakis: its not additional indent. There were tabs and switched them to spaces to match properly

paulkatsoulakis: Now i saw what ilya meant, i just pushed the fix.
In the first place the code had extra indent already.
I thought he meant that my change introduced extra indent!

fixed, thanks ilya!

daemon/main.c
```diff
@@ -899,12 +899,9 @@ int main(int argc, char **argv) {
                             return 0;
                         }
                         else if(strncmp(optarg, createdataset_string, strlen(createdataset_string)) == 0) {
-                            unsigned history_seconds;
-
                             optarg += strlen(createdataset_string);
-                            history_seconds = (unsigned )strtoull(optarg, NULL, 0);
-
 #ifdef ENABLE_DBENGINE
+                            unsigned history_seconds = (unsigned )strtoull(optarg, NULL, 0);
```
ilyam8: minor: extra space

> (unsigned )

mfundul: wait, how did this line get in the #ifdef? what's going on here?

mfundul: and why did you move the decleration together with the definition, this is not ANSI C, which it was before.

PR #6234, opened 2019-06-06
Remove CNCF logo and TOC presentation reference

dankohn:
Your use of the CNCF logo and mention of the TOC presentation create the
false impression that Netdata is a CNCF-hosted project. I would appreciate
if you accept this PR to clarify that it is not.


cakrit:
Wiki updated as well.

PR #6237, opened 2019-06-06
dns_query_time py module: saving dns request in 'r', checking response for answer, recording '-…

n0coast:
##### Summary
Currently dns_query_time.py will record the time delta for DNS queries that return no values in the 'answer', this commit checks for a null value in 'answer' and records a -50 if true. Open to suggestions on other ways to handle non-answer responses, but this works for my current use case.

##### Component Name
collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py 

##### Additional Information
Image of DNS Response Time graph with code change in place recording -50 for intermittent requests returning no answer:
DNS Response Time

### DIFF
--- a/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
+++ b/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
@@ -85,7 +85,7 @@ def dns_request(server_list, timeout, domains):
 
         try:
             resp = dns.query.udp(request, ns, timeout=t)
-            if not (resp.rcode == dns.rcode.NOERROR or resp.answer):
+            if (resp.rcode == dns.rcode.NOERROR and not resp.answer):
                 query_time = -100
             else:
                 query_time = resp.time * 1000
--- a/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
+++ b/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
@@ -85,10 +85,10 @@ def dns_request(server_list, timeout, domains):
 
         try:
             resp = dns.query.udp(request, ns, timeout=t)
-            if (resp.rcode == dns.rcode.NOERROR and not resp.answer):
-                query_time = -100
-            else:
+            if (resp.rcode() == dns.rcode.NOERROR and resp.answer):
                 query_time = resp.time * 1000
+            else:
+                query_time = -100
         except dns.exception.Timeout:
             query_time = -100
         finally:
--- a/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
+++ b/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
@@ -8,11 +8,6 @@ from socket import getaddrinfo, gaierror
 from threading import Thread
 
 try:
-    from time import monotonic as time
-except ImportError:
-    from time import time
-
-try:
     import dns.message
     import dns.query
     import dns.name
@@ -89,16 +84,15 @@ def dns_request(server_list, timeout, domains):
         request = dns.message.make_query(domain, dns.rdatatype.A)
 
         try:
-            dns_start = time()
-            r = dns.query.udp(request, ns, timeout=t)
-            dns_end = time()
-            if not r.answer:
-                q.put({'_'.join(['ns', ns.replace('.', '_')]): -50})
+            resp = dns.query.udp(request, ns, timeout=t)
+            if not (resp.rcode == dns.rcode.NOERROR or resp.answer):
+                query_time = -100
             else:
-                query_time = round((dns_end - dns_start) * 1000)
-                q.put({'_'.join(['ns', ns.replace('.', '_')]): query_time})
+                query_time = resp.time * 1000
         except dns.exception.Timeout:
-            q.put({'_'.join(['ns', ns.replace('.', '_')]): -100})
+            query_time = -100
+        finally:
+            q.put({'_'.join(['ns', ns.replace('.', '_')]): query_time})
 
     for server in server_list:
         th = Thread(target=dns_req, args=(server, timeout, que))
@@ -149,6 +143,8 @@ def create_charts(aggregate, server_list):
                         '_'.join(['ns', ns.replace('.', '_')]),
                         ns,
                         'absolute',
+                        1,
+                        100,
                     ]
                 ]
             }
--- a/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
+++ b/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
@@ -143,8 +143,6 @@ def create_charts(aggregate, server_list):
                         '_'.join(['ns', ns.replace('.', '_')]),
                         ns,
                         'absolute',
-                        1,
-                        100,
                     ]
                 ]
             }
--- a/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
+++ b/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
@@ -90,10 +90,13 @@ def dns_request(server_list, timeout, domains):
 
         try:
             dns_start = time()
-            dns.query.udp(request, ns, timeout=t)
+            r = dns.query.udp(request, ns, timeout=t)
             dns_end = time()
-            query_time = round((dns_end - dns_start) * 1000)
-            q.put({'_'.join(['ns', ns.replace('.', '_')]): query_time})
+            if not r.answer:
+                q.put({'_'.join(['ns', ns.replace('.', '_')]): -50})
+            else:
+                query_time = round((dns_end - dns_start) * 1000)
+                q.put({'_'.join(['ns', ns.replace('.', '_')]): query_time})
         except dns.exception.Timeout:
             q.put({'_'.join(['ns', ns.replace('.', '_')]): -100})
 


ilyam8:
Hey @n0coast 	
First of all thanks for contributing.

btw all these '-100', '-50' is really wrong, it was a bad choice to use negative number to show errors..

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
```diff
@@ -90,10 +90,13 @@ def dns_req(ns, t, q):
 
         try:
             dns_start = time()
-            dns.query.udp(request, ns, timeout=t)
+            r = dns.query.udp(request, ns, timeout=t)
             dns_end = time()
-            query_time = round((dns_end - dns_start) * 1000)
-            q.put({'_'.join(['ns', ns.replace('.', '_')]): query_time})
+            if not r.answer:
```
ilyam8: i suggest to check `rcode` 
http://www.dnspython.org/docs/1.14.0/dns.rcode-module.html
```python
resp = dns.query.udp(request, ns, timeout=t)
if not (resp and resp.rcode == dns.rcode.NOERROR):
    put
    return
```

ilyam8:
@n0coast 

`-100` means that request was not successful (timeout, wrong, etc.) So i suggest to use `-100` for all cases. If we need something to indicate a reason of failure then we need a separate chart for it.

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
n0coast: > i suggest also to check `rcode`
> http://www.dnspython.org/docs/1.14.0/dns.rcode-module.html
> 
> ```python
> resp = dns.query.udp(request, ns, timeout=t)
> if not (resp and resp.rcode == dns.rcode.NOERROR):
>     put
>     return
> ```

This doesn't seem to work, in my case resp will exist but not resp.answer:
```NO RESPONSE: id 63190
opcode QUERY
rcode NOERROR
flags QR RD RA
;QUESTION
sub.domain.com. IN A
;ANSWER
;AUTHORITY
domain.com. 445 IN SOA a1.verisigndns.com. mdnshelp.verisign.com. 1533674417 28800 7200 1209600 86400
;ADDITIONAL```
note rcode is NOERROR but no data exists in answer section, it's not a common failure scenario I don't think, but it's the one I'm battling at the moment.

> btw all these '-100', '-50' are really wrong, it was a bad choice to use negative number to show errors..
Would it be possible to return 'None' instead of -100? It seems to break the code trying it with debug mode.

Comment (ilyam8) [reply collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py]: oh,you are right, i reviewed at night :see_no_evil: 	

```python
if not (resp.rcode == dns.rcode.NOERROR and resp.answer):
```

btw this is source code

http://www.dnspython.org/docs/1.15.0/dns.query-pysrc.html#udp

i see response has `time` attribute, lets use it!

```python
250      finally: 
251          if begin_time is None: 
252              response_time = 0 
253          else: 
254              response_time = time.time() - begin_time 
255          s.close() 
256      r = dns.message.from_wire(wire, keyring=q.keyring, request_mac=q.mac, 
257                                one_rr_per_rrset=one_rr_per_rrset) 
258      r.time = response_time 
```

ilyam8: > Would it be possible to return a null value instead of -100?

It is very easy to do, the problem that it breaks alarms, which is not acceptable.

n0coast: the response_time value is in seconds, multiplying by 1000 to get ms and rounding gives many responses of `0` - removing rounding still produces an integer, is zero an acceptable response?

ilyam8: I guess it is < 0.5 ms that is why it is rounded to 0.

to make it more precise multiply `query_time` by 100000 in dns_request
and

change 
https://github.com/netdata/netdata/blob/90108a57d15f6714669daa61ba8e7603d7da1218/collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py#L146-L148

to 
```python
 '_'.join(['ns', ns.replace('.', '_')]), 
 ns, 
 'absolute', 
 1,
 100,
```

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
```diff
@@ -89,13 +84,15 @@ def dns_req(ns, t, q):
         request = dns.message.make_query(domain, dns.rdatatype.A)
 
         try:
-            dns_start = time()
-            dns.query.udp(request, ns, timeout=t)
-            dns_end = time()
-            query_time = round((dns_end - dns_start) * 1000)
-            q.put({'_'.join(['ns', ns.replace('.', '_')]): query_time})
+            resp = dns.query.udp(request, ns, timeout=t)
+            if not (resp.rcode == dns.rcode.NOERROR or resp.answer):
```
ilyam8: `or` ? i recall you said it was `resp.rcode == dns.rcode.NOERROR` and `not resp.answer` in your case?

if so we should use `and`

n0coast: Good catch!

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
```diff
@@ -85,7 +85,7 @@ def dns_req(ns, t, q):
 
         try:
             resp = dns.query.udp(request, ns, timeout=t)
-            if not (resp.rcode == dns.rcode.NOERROR or resp.answer):
+            if (resp.rcode == dns.rcode.NOERROR and not resp.answer):
```
ilyam8: wait wait

`resp.rcode == dns.rcode.NOERROR` is OK

lets do

```python
if resp.rcode == dns.rcode.NOERROR and resp.answer: 
    ok ...
else:
    not ok...
```

ilyam8: we good only if `resp.rcode == dns.rcode.NOERROR` and `resp.answer`, right?

n0coast: got it, pushing another commit now

n0coast: hmm, doesn't seem to work that way

simple code snipet:
```
import dns.query

try:
    qname = dns.name.from_text('www.google.com')
    q = dns.message.make_query(qname, dns.rdatatype.A)
    r = dns.query.udp(q, '8.8.8.8')
    resp.rcode == dns.rcode.NOERROR
except dns.exception.Timeout:
    print('failure')
except:
    print('fallback')
```

ilyam8:
```python
    def dns_req(ns, t, q):
        domain = dns.name.from_text(choice(domains))
        request = dns.message.make_query(domain, dns.rdatatype.A)
        query_time = -100

        try:
            resp = dns.query.udp(request, ns, timeout=t)
            if resp.rcode == dns.rcode.NOERROR and resp.answer:
                query_time = resp.time * 100000
        except dns.exception.DNSException:
            pass
        finally:
            name = '_'.join(['ns', ns.replace('.', '_')]
            q.put({name: query_time})
```

n0coast: multiplying by 10000 gives _huge_ numbers, since it's ms to sec 1000 should be correct and looks good on the graphs I've been using to validate. resp.rcode needs parenthesis - both should be in the commit being validated now.

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
```diff
@@ -89,13 +84,15 @@ def dns_req(ns, t, q):
         request = dns.message.make_query(domain, dns.rdatatype.A)
 
         try:
-            dns_start = time()
-            dns.query.udp(request, ns, timeout=t)
-            dns_end = time()
-            query_time = round((dns_end - dns_start) * 1000)
-            q.put({'_'.join(['ns', ns.replace('.', '_')]): query_time})
+            resp = dns.query.udp(request, ns, timeout=t)
+            if (resp.rcode() == dns.rcode.NOERROR and resp.answer):
```
ilyam8: wait, `rcode` is `int` according docs

http://www.dnspython.org/docs/1.14.0/dns.message.Message-class.html

how is it `rcode()` ?

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
ilyam8: > multiplying by 10000 gives huge numbers

multiply by `100000` (not 10000) in the code and add divisor 100 to chart

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
ilyam8: wait,  it is method, i am wrong

n0coast: I dunno, but if you try it you'll see:

```import dns.query

qname = dns.name.from_text('www.google.com')
q = dns.message.make_query(qname, dns.rdatatype.A)
resp = dns.query.udp(q, '8.8.8.8')
print(resp.rcode == dns.rcode.NOERROR)
print(resp.rcode() == dns.rcode.NOERROR)
```

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
```diff
@@ -89,13 +84,15 @@ def dns_req(ns, t, q):
...
+                query_time = resp.time * 1000
```
ilyam8: please change to `100000`

resp.time in seconds, we convert it to ms 

* 1000

since we add `100` as divisor in chart, we need to multiply by 1000 * 100 (100000)

you will see x100 numbers in debug, but they will be ok on the dashboard

collectors/python.d.plugin/dns_query_time/dns_query_time.chart.py
ilyam8: Sorry for the mess. My bad.

n0coast: not a problem - thanks for working through and offering advice!

ilyam8:
lgtm @n0coast :+1: 	again, sorry for the little mess :smile:

Issue #4717, opened 2018-11-22
web server URL parsing ignores escaped characters


ktsaou:
There is a bug at the internal web server of netdata. It parses the URLs like this:

1. a request comes

2. it checks it received the whole of it (it searches for `\r\n\r\n` - i.e. an empty line as defined in the spec)

3. BUG 1: it decodes the whole URL. This is a bug since any `/` or `?` URL encoded will be decoded before actually parsing the URL. To work properly it should decode the URL part by part (i.e. first split it in its parts and then decode each part of it). **The way it is now, URL encoding cannot be used to encode characters that are URL control characters (`/`, `?`, `&`, `%`).** Even when URL encoded, these character will still play an important role in URL parsing (although they are escaped).

4. BUG 2: it does not separate URL `path` from URL `query string`. It treats the whole URL as a continuous string, separated by `/?&`, which of course leads to bugs like the one we face here.

So, the right flow should be like this:

1. receive the request

2. make sure it received the whole of it

3. split the path from the query string

4. URL decode the path to find the API command or the filename to be served

5. split the query string in its parts (separated with `&`)

6. for each part of the query string, split it at the `=` sign to find the `name` and `value` and decode both of them.

I think the whole HTTP parser needs to be rewritten.

Related to #4716 #4714 #3253

cakrit:
Isn't this closed by #4716 ? What is left to do?

ktsaou:
I applied a fix to make parsing a bit more strict, but the whole HTTP parser is faulty. It needs to be rewritten. Not high priority, but this will eventually lead to issues as netdata matures.

PR #6247, opened 2019-06-07
Url parser refactoring

thiagoftsm:
##### Summary
After some problems caused by two bugs present in the URL parser, we decided to do a complete review on it. No less important, it has been commented to simplify further change on it.

Related: #4717

##### Component Name
All the URL parser


ilyam8:
Why do not use some good library as url parser?
https://github.com/uriparser/uriparser/tree/master/src

The library is tested and looks very nice. You don't need to re implement url parsing functionality.

thiagoftsm:
Good point Ilya, the Netdata had its own library defined for its necessities. In our case, for example, we do not need to store all the Headers sent by the client, few them are enough for us to deliver the response. Another good point it is the fact that we do not need to allocate new memory for almost nothing in the parser process, so this increase the speed to process other things. This is something that goes in the same direction of the motive we have our own database. Of course, we can study other libraries to improve our library.

ilyam8:
> This is something that goes in the same direction of the motive we have our own database.

Ok, let me explain my point. I tend to disagree. Database is another case.

> so this increase the speed to process other things.

There is nothing to optimize, we don't have problem with url parsing speed. So, if there is no problen, better to prefer simplicity and readability over optimizing.

URL parsing has no connection with any part of netdata, it is standalone thing imo, i mean we don't need to do it in specific way because, for example, part A needs only B and C.

If you will do it that way you won't need to come back and rewrite implementation. So the interface will be stable which is important.

> for example, we do not need to store all the Headers sent by the client

URL has no headers.

thiagoftsm:
Hi Ilya,

Everytime that you say for the kernel something, it will need to expend time with the process instead other one, so this affects directly the database too. We are rewriting the  URL parser due a request done in the issue #4717 , there you can have a better vision of the limitations that we have right now.

The URL parsing has connection with other parts, because after the parsing happens, it can read a file from filesystem or to connect with the database. No less important, we are talking about threads competing among them by the same resource.

You are 100% correct, URL does not have headers, but we are working with client request, I think Costa and I probably did not use the correct terms, because we are in the reality rewriting the whole parsing process and in this process it is necessary to keep few Headers sent by the browser. I will give an example based in a job that I did, when a client send a HTTP  request for a Netdata that only accepts SSL connection, we get not only the URL that was sent for us, that is the the first line of the request, but we get the "Host:" Header for we send a 301 message with the link to redirect the new request.

ilyam8:
well #4717 is about URL parser.

i see Costa mentioned there

> I think the whole HTTP parser needs to be rewritten.

i think it is another issue, if you are doing it, then the PR title ir wrong, it is not URL parser PR.

>  it will need to expend time with the process instead other one, so this affects directly the database too.

I don't see how it affects database. URL parsing affects database?

Anyway this is theory, you don't have any tests/benchmarks how it affetcs speed. I dont think that URL parsing affects the speed and we shoud optimize it.

___

Dont get me wrong, that is just my point i want to share/discuss.

@vlvkobal @mfundul 	what is your opinion on this?

thiagoftsm:
It will never affect direct, because after the parse I will only call the DB to get data, so I am not slowing down it due code changes. The parser affects indirect, because the time that it needs to parser the request the other threads can be stopped, everything depends of the processor schedule. I will give a code for you that you help you to understand this.

libnetdata/inlined.h
```diff
@@ -31,6 +31,29 @@ static inline uint32_t simple_hash(const char *name) {
...
+ * Simple N hash
+ *
+ * Calculate the hash for name that has the length len.
+ *
+ * @param name the input to the hash
+ * @param len the length of the input
+ *
+ * @return It returns the hash calculated.
+ */
+static inline uint32_t simple_nhash(const char *name,size_t len) {
```
mfundul: Code style, you should add a space so that it is: `const char *name, size_t len`

libnetdata/inlined.h
```diff
@@ -31,6 +31,29 @@ static inline uint32_t simple_hash(const char *name) {
...
+    unsigned char *s = (unsigned char *) name;
+    size_t i;
+    uint32_t hval = 0x811c9dc5;
+    i = 0;
+    do {
```
mfundul: Code style, you're wasting a lot of lines:
```
    i = 0;
    do {
```
 and do while is not readable here.
It should be `for (i=0 ; i < len ; ++i) {`

libnetdata/inlined.h
```diff
@@ -42,6 +65,21 @@ static inline uint32_t simple_uhash(const char *name) {
     return hval;
 }
 
+static inline uint32_t simple_nuhash(const char *name,size_t len) {
+    unsigned char *s = (unsigned char *) name;
+    size_t i;
+    uint32_t hval = 0x811c9dc5, c;
+
+    i = 0;
+    do {
```
mfundul: Code style: same as above, you should use `for()`.

web/api/badges/web_buffer_svg.c
```diff
@@ -912,46 +924,54 @@ int web_client_api_request_v1_badge(RRDHOST *host, struct web_client *w, char *u
...
-        else if(!strcmp(name, "precision")) precision_str = value;
-        else if(!strcmp(name, "scale")) scale_str = value;
-        else if(!strcmp(name, "alarm")) alarm = value;
+    uint32_t i = 0;
+    uint32_t end = w->total_params;
+    char save[WEB_FIELDS_MAX];
+    char *value;
+    size_t lvalue;
+    if(end) {
+        do {
```
mfundul: Code style:
Again for() is more readable, you waste those lines:
```
     uint32_t i = 0;
...
     if(end) {
```
and you can remove the whole identation of `if (end)` since
```
for (i = 0 ; i < end ; ++i) {
```
already covers the `if (end)` case.
```

File: web/api/badges/web_buffer_svg.c [hunk 5]
```diff
@@ -1137,6 +1157,14 @@ int web_client_api_request_v1_badge(RRDHOST *host, struct web_client *w, char *u
     }
 
     cleanup:
+    if(end) {
```
mfundul: Code style, same as above, you get can get rid of `if (end)` and `i =0;` with a for loop.

web/api/exporters/allmetrics.c
```diff
@@ -17,60 +17,83 @@ struct prometheus_output_options {
...
-                    else
-                        prometheus_output_options &= ~prometheus_output_flags_root[i].flag;
-
-                    break;
+    uint32_t end = w->total_params;
+    uint32_t i;
+    char sname[WEB_FIELDS_MAX];
+    char svalue[WEB_FIELDS_MAX];
+    if (end) {
+        i = 0;
```
mfundul: Code style: this `i = 0;` is less readable than a for loop `for (i = 0 ; i < len ; ++i) {`

web/api/health/health_cmdapi.c
```diff
@@ -73,75 +70,86 @@ int web_client_api_request_v1_mgmt_health(RRDHOST *host, struct web_client *w, c
...
-                                (hash == hash_alarm && !strcasecmp(key, HEALTH_ALARM_KEY)) ||
-                                (hash == hash_template && !strcasecmp(key, HEALTH_TEMPLATE_KEY)) ||
-                                (hash == hash_chart && !strcasecmp(key, HEALTH_CHART_KEY)) ||
-                                (hash == hash_context && !strcasecmp(key, HEALTH_CONTEXT_KEY)) ||
-                                (hash == hash_host && !strcasecmp(key, HEALTH_HOST_KEY)) ||
-                                (hash == hash_families && !strcasecmp(key, HEALTH_FAMILIES_KEY))
-                                ) {
-                            silencer = create_silencer();
+            uint32_t end = w->total_params;
+            if (end) {
```
mfundul: Code style:
Same as above, you can replace:
```
             if (end) {
                uint32_t i = 0;
                do {
```
with `for (i = 0 ; i < end ; ++i) {`

web/api/web_api_v1.c
```diff
@@ -194,15 +194,30 @@ inline uint32_t web_client_api_request_v1_data_google_format(char *name) {
...
+    (void)url;
     int all = 0;
 
-    while(url) {
-        char *value = mystrsep(&url, "&");
-        if (!value || !*value) continue;
+    uint32_t end = w->total_params;
+    if(end) {
+        uint32_t  i = 0;
+        do {
```
mfundul: Code style:
```
    if(end) {
        uint32_t  i = 0;
        do {
```
same as above, you can replace those lines with a for loop.

web/api/web_api_v1.c
```diff
@@ -212,18 +227,36 @@ inline int web_client_api_request_v1_alarms(RRDHOST *host, struct web_client *w,
...
-        char *value = mystrsep(&url, "&");
-        if (!value || !*value) continue;
-
-        char *name = mystrsep(&value, "=");
-        if(!name || !*name) continue;
-        if(!value || !*value) continue;
-
-        if(!strcmp(name, "after")) after = (uint32_t)strtoul(value, NULL, 0);
+    uint32_t end = w->total_params;
+    if(end) {
```
mfundul: Cody style: You can replace those lines:
```
     if(end) {
        uint32_t  i = 0;
        do {
...
         } while (++i < end);
     }
```
with
```
for (i = 0 ; i < end ; ++i) {
...
}
```
it's more readable.
```

File: web/api/web_api_v1.c [hunk 10]
```diff
@@ -232,28 +265,45 @@ inline int web_client_api_request_v1_alarm_log(RRDHOST *host, struct web_client
...
-        //  goto cleanup;
-        //}
+    uint32_t  i = 0;
+    uint32_t end = w->total_params;
+    //In normal situation this vector is not necessary,
+    //but it is safety to have it here.
+    char save[WEB_FIELDS_MAX];
+    char *value;
+    size_t vlength;
+    if(end) {
```
mfundul: Code style: same as above, you can use a for loop here.

web/api/web_api_v1.c
```diff
@@ -273,16 +323,47 @@ inline int web_client_api_request_single_chart(RRDHOST *host, struct web_client
     w->response.data->contenttype = CT_APPLICATION_JSON;
     st->last_accessed_time = now_realtime_sec();
     callback(st, w->response.data);
-    return 200;
+    ret =  200;
 
     cleanup:
+    if(end) {
```
mfundul: Code style: same as above, you can use a for loop here.

web/api/web_api_v1.c
```diff
@@ -333,75 +436,103 @@ inline int web_client_api_request_v1_data(RRDHOST *host, struct web_client *w, c
...
-                    format = web_client_api_request_v1_data_google_format(google_out);
+    uint32_t end = w->total_params;
+    //Save the value present in the options different of tqx
+    char save[WEB_FIELDS_MAX];
+    //Save the pointer where ';' and ':' are in tqx to restore
+    char *tqxname[WEB_FIELDS_MAX];
+    uint32_t tqxtotal=0;
+    char *value ;
+    size_t lvalue;
+    if(end) {
```
mfundul: Code style:
You can replace:
```
     if(end) { 
         uint32_t i = 0;
         do {
...
         } while (++i < end); 
     } 
```
with
```
for (uint32_t i = 0 ; i < end ; ++i) {
...
}
```
```

File: web/api/web_api_v1.c [hunk 13]
```diff
@@ -488,10 +618,54 @@ inline int web_client_api_request_v1_data(RRDHOST *host, struct web_client *w, c
         buffer_strcat(w->response.data, ");");
 
     cleanup:
+    if(end) {
+        uint32_t i;
+        if (tqxtotal) {
```
mfundul: Code style:
you can replace:
```
         if (tqxtotal) { 
             i = 0; 
             do { 
                 *tqxname[i] = ':'; 
             } while (++i < tqxtotal); 
         } 
```
with
```
             for (i = 0 ; i < tqxtotal ; ++i) { 
                 *tqxname[i] = ':'; 
             }
```

web/api/web_api_v1.c
```diff
@@ -488,10 +618,54 @@ inline int web_client_api_request_v1_data(RRDHOST *host, struct web_client *w, c
...
+            i = 0;
+            do {
+                *tqxname[i] = ':';
+            } while (++i < tqxtotal);
+        }
+
+        i = 0;
```
mfundul: Code style: same as above, you can use a for loop.

web/api/web_api_v1.c
```diff
@@ -488,10 +618,54 @@ inline int web_client_api_request_v1_data(RRDHOST *host, struct web_client *w, c
...
+ * Restore the original values for the registry variables.
+ *
+ * @param names the structure where the names are linked
+ * @param sname the value saved for names
+ * @param values the structure where the values are linked
+ * @param svalue the value saved for values
+ * @param end the number of elements to check
+ */
+static void web_client_api_restore_registry_variable(struct web_fields *names,char *sname,struct web_fields *values,char *svalue,uint32_t end) {
+    if(!end) {
```
mfundul: Code style:
you can replace:
```
     if(!end) { 
         return; 
    }

    uint32_t i = 0;
    do {
...
     } while(++i < end); 
```
with:
```
    for (uint32_t i = 0 ; i < end ; ++i) {
...
     } 
```

web/api/web_api_v1.c
```diff
@@ -551,26 +736,40 @@ inline int web_client_api_request_v1_registry(RRDHOST *host, struct web_client *
     int redirects = 0;
 */
 
-    while(url) {
-        char *value = mystrsep(&url, "&");
-        if (!value || !*value) continue;
+    uint32_t i = 0;
+    uint32_t end = w->total_params;
+    if (!end) {
```
mfundul: Code style:
you can replace:
```
     if (!end) { 
         goto nothing; 
    }

    char sname[WEB_FIELDS_MAX];
    char vname[WEB_FIELDS_MAX];
    do { 
...
    } while (++i < end );
nothing: 
```
with:
```
    char sname[WEB_FIELDS_MAX];
    char vname[WEB_FIELDS_MAX];
     for (i = 0 ; i < end ; ++i)
...
    }
```

libnetdata/inlined.h
```diff
@@ -42,6 +65,21 @@ static inline uint32_t simple_uhash(const char *name) {
     return hval;
 }
 
+static inline uint32_t simple_nuhash(const char *name,size_t len) {
```
mfundul: Code style, you should insert a space after a comma (`name, size_t len`), there are a lot of places in the code you could do that, I'm not going to write a comment for each because I will pollute the review log but you should do a cleanup.

libnetdata/url/url.c
```diff
@@ -79,3 +79,128 @@ char *url_decode_r(char *to, char *url, size_t size) {
...
+            if ( begin == end )
+            {
+                break;
+            }
+
+            if (*begin == '\n')
+            {
+                counter++;
+            }
+        } else if (*begin == '\n') {
```
mfundul: Logic:
If I have the string "\n\nAB" I think the we will have `counter == 2`, is this possible?

libnetdata/url/url.c
```diff
@@ -79,3 +79,128 @@ char *url_decode_r(char *to, char *url, size_t size) {
...
+inline HTTP_VALIDATION url_is_request_complete(char *begin,char *end,size_t length) {
+    //Message cannot be complete when first and last address are the same
+    if ( begin == end) {
+        return HTTP_VALIDATION_INCOMPLETE;
+    }
+
+    //We always set as begin the first character of the message, but we wanna check
+    //the last characters, so I am adjusting it case it is possible.
+    if ( length > 3  ) {
+        begin = end - 4;
```
mfundul: Logic:
Can we get out of bounds here? If I have `length == 4` doesn't this mean that `begin` is `begin[0]`, `end` is `begin[3]` and `end - 4` is `begin[-1]`?

libnetdata/url/url.c
```diff
@@ -79,3 +79,128 @@ char *url_decode_r(char *to, char *url, size_t size) {
...
+ *
+ * @return It returns the number of variables processed
+ */
+int url_parse_query_string(struct web_fields *names,struct web_fields *values,char *moveme,char *divisor) {
+    uint32_t i = 0;
+    uint32_t max = WEB_FIELDS_MAX;
+
+    do {
+        if ( i == max) {
+            error("We are exceeding the maximum number of elements possible(%u) in this query string(%s)",max,moveme);
```
mfundul: Logic:
Is it ok that this will cause a `return ++i` and it will return `WEB_FIELDS_MAX + 1`?

web/server/web_client.c
```diff
@@ -812,109 +849,348 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
...
-    char *s = (char *)buffer_tostring(w->response.data), *encoded_url = NULL;
+static inline HTTP_VALIDATION web_client_is_complete(char *begin,char *end,size_t length){
+    if ( begin == end){
+        return HTTP_VALIDATION_INCOMPLETE;
+    }
 
-    size_t last_pos = w->header_parse_last_size;
-    if(last_pos > 4) last_pos -= 4; // allow searching for \r\n\r\n
-    else last_pos = 0;
+    if ( length > 3  ){
```
mfundul: Logic: same as above, does this get out of bounds?

web/server/web_client.c
```diff
@@ -812,109 +849,348 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
...
-                info("Disabling slow client after %zu attempts to read the request (%zu bytes received)", w->header_parse_tries, buffer_strlen(w->response.data));
-                w->header_parse_tries = 0;
-                w->header_parse_last_size = 0;
-                web_client_disable_wait_receive(w);
-                return HTTP_VALIDATION_NOT_SUPPORTED;
+            if (*begin == '\n')
+            {
+                counter++;
             }
+        } else if (*begin == '\n') {
```
mfundul: Logic: same as above, what if we have strings like `"\n\nAB"`?

web/server/web_client.c
```diff
@@ -812,109 +849,348 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
...
+ * @param enddir a pointer to the position with the last valid character
+ * @param endcmd the end of the last character.
+ */
+static inline void web_client_set_directory(struct web_client *w,char *begin,char *enddir,char *endcmd) {
+    if (enddir) {
+        w->directory.body = begin;
+        w->directory.length = enddir - begin;
+
+        if (!strncmp(w->directory.body,"api",w->directory.length)) {
+            begin = enddir + 1;
```
mfundul: Logic: can this get out of bounds? Is `enddir + 1` always smaller than `endcmd`?

web/server/web_client.c
```diff
@@ -812,109 +849,348 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
...
+ */
+static inline void web_client_set_without_query_string(struct web_client *w) {
+    w->query_string.body = NULL;
+    w->query_string.length = 0;
+
+    char *test = w->path.body+1;
+    if (!strncmp(test,"api/v1/",7) ) {
+        test += 7;
+        char *endofcommand=strchr(test,' ');
+        w->command.length=(endofcommand)?(size_t)endofcommand-(size_t)test:strlen(test);
```
mfundul: Logic: you want to do pointer arithmetic here so `(size_t)endofcommand-(size_t)test` should be `(size_t)(endofcommand - test)`.

web/server/web_client.c
```diff
@@ -812,109 +849,348 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
...
+
+    //Is there a separator?
+    char *moveme = strchr(w->path.body,'?');
+    char *enddir;
+    if (moveme) { //I have a query string
+        w->path.length = moveme - w->path.body;
+        w->query_string.body = moveme;
+        w->query_string.length = w->decoded_length - w->path.length;
+
+        enddir = strchr(w->path.body+1,'/');
```
mfundul: Logic: can `strchr()` return NULL here? Can `w->path.body+1` be out of bounds?

web/server/web_client.c
```diff
@@ -812,109 +849,348 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
...
+        char *begin = w->path.body+1;
+        web_client_set_directory(w,begin,enddir,moveme);
+        if (w->query_string.body) {
+            enddir = strchr(moveme,'=');
```
mfundul: Logic: can `strchr()` return NULL here?

mfundul: I see NULL is handled in `web_client_parse_request()`

web/server/web_client.c
```diff
@@ -812,109 +849,348 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
...
+        //The fact that I do not have query string, does not mean I do not have directories,
+        //so it is necessary to continue working
+        enddir = strchr(w->path.body+1,'/');
+        w->directory.body = w->path.body + 1;
+        if(enddir) {
+            w->directory.length = (size_t)(enddir - w->directory.body);
+            enddir++;
+
+            w->version.body = enddir;
+            enddir = strchr(++enddir,'/');
```
mfundul: Logic: at this point we've already done `++enddir` twice, can we get out of bounds?

web/server/web_client.c
```diff
@@ -812,109 +849,348 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
...
+ */
+static inline void web_client_parse_headers(struct web_client *w,char *s) {
+    while(*s) {
+        // find a line feed
+        while(*s && *s++ != '\r');
+
+        // did we reach the end?
+        if(unlikely(!*s)) break;
+
+        if (*s == '\n') {
```
mfundul: Logic: what if there is no `'\n'`, is this a problem?

libnetdata/inlined.h
thiagoftsm: All right, I am fixing this with the last commit!

libnetdata/inlined.h
thiagoftsm: I confess I am working my brain to understand the concept of readable, but I am changing because I know the final result will be the same.

web/api/badges/web_buffer_svg.c
thiagoftsm: I agree that for covers the end, but I do not see any necessity to use it, we will need to talk about this readable when you return!

web/api/web_api_v1.c
thiagoftsm: I am sorry, I will keep the declaration outside the parenthesis!

libnetdata/url/url.c
thiagoftsm: you are correct and this will be discarded later in the parser, because we do not have a valid request. The syntax of this function was based in the h2o/h2o web server parser request. 
Case we would follow the RFC like it was written, we would need to accept only '\r\n\r\n' at the end, but before the HTTP 1.1 some clients sent '\n\n' to finish a message, so it is a kind of compatibility. Now we have the script (request.sh) to test the URLs inside tests/urls, case you comment the line relative to (*begin == \n) you will see that wget can fail.

libnetdata/url/url.c
thiagoftsm: Begin cannot be negative, because this is mermory allocated by the server. So the pointer will be far from the first positions of the memory, because they are used by the processor when the computer starts, but I agree it is safety to have length >= 4!

libnetdata/url/url.c
thiagoftsm: All the places that the constant WEB_FIELDS_MAX is used, it is to define the limit value, and with this we never reach it when we are working with the vector in a loop, but I agree that when we have a vector is more logic to avoid reach that value, so I am fixing this.

web/server/web_client.c
thiagoftsm: This is worse. While we were developing @ilyam8 gave us a suggestion to move part of the code to the libnetdata, the suggestion was accept and this function was copied to there, but after the tests I forgot to remove it from here. Thanks by the review!

web/server/web_client.c
thiagoftsm: Yes it is. The "enddir" is a pointer for the PATH in the first line of the request, and the request structure in the first line will always have structure like this:

METHOD PATH HTTP/1.1\r\n\r\n

So we will have at least 13 bytes after the path.

web/server/web_client.c
thiagoftsm: Both cases will give the correct result, but I agree that it is more readable this way (Am I learning correctly the concept of readable?)

web/server/web_client.c
thiagoftsm: In the theory this would not happen, because we have before this a search by "?" that is the start of the query string, and for we have the interrogation we need to have before, ate least something like "/TEXT?", but you are correct, because in the practice somebody can send a bad request having only "?".

web/server/web_client.c
thiagoftsm: It is the same previous explanation about the first line of the request, so we will have at least more 13 bytes in front.

web/server/web_client.c
thiagoftsm: This function I am carrying from the previous parser, it is not a problem this part of the code, because this is the header parser(The headers are from the second line until the last line of the request), and we already saw in the previous functions of the algorithm that we will have at least two '\n' at the end of the request.

web/server/web_client.c
mfundul: This can be worse cause some compilers givevsize_t be 32 bit when the pointers are 64 bit and if you're unlucky you get bogus numbers. The correct thing is to do pointer arithmetic before an integer cast.

libnetdata/url/url.c
vlvkobal: If we are checking "\r\n\r\n" at the end, why can't we simply use `strncmp()`? Why do you need `begin`, `end`, and 'length', if only two of them are more than enough?

ilyam8:
Ok, that is what i did:

**Setup**:
 - locally compiled and installed `url_parser` branch netdata
 - 2 slaves (master latest)

**Tests**:
 - executed [`request.sh`](https://github.com/thiagoftsm/netdata/blob/URL_parser_review/tests/urls/request.sh) script
 - did some browsing


**Bugs**:
i noticed the bug with diplaying alarms in the gui.

Some of the [`pihole` alarms](https://github.com/netdata/netdata/blob/master/health/health.d/pihole.conf) has bad synax, see the `delay` string

https://github.com/netdata/netdata/blob/b27f0c65342cee0463eaf62d7e53112aaa1fdbb0/health/health.d/pihole.conf#L43-L53

i used wrong keyword - `delay` instead of `down/up`.

**gui**:
 - on master i see **all** the alarms (option with wrong keyword was not applied)
 - on this branch i see only couple of alarms

Screenshot_20190625_172513


**http api**:
 - master and branch returns **all** alarms. No problems there.

vlvkobal:
`http://localhost:19999/api/v1/alarms?all&_=1561479810533` doesn't work. There are at least two issues.
- `web_client_api_request_v1_alarms()` checks the parameter value instead of the parameter name.
- `url_parse_query_string()` treats the part of the query string up to the '=' symbol as a first parameter and ignores '&'.

thiagoftsm:
Considering all the comments that I received, I decided to change the direction. When I read again the issue #4717 I saw that I did not understand properly what was the suggestion there, instead to rewrite I was recreating and this brought a lot of problems. With this last commit, I am doing exactly what was asked there. I am reordering the parser and the path is separated of the query string. Until now the unique thing missed is the separation between fields in the query string.

vlvkobal:
A request like the following one
`http://arch-esxi:19999/api/v1/(*@&$!$%%5E)!$*%&)!$*%%5E*!%5E%!%5E$%!%5E%(!*%5E*%5E%(*@&$%5E%(!%5E#*&!^#$*&!^%)@($%^)!*&^(!*&^#$&#$)!$%^)!$*%&)#$!^#*$^!(*#^#)!%^!)$*%&!(*&$!^#$*&^!*#^$!*^)%(!*&$%)(!&#$!^*#&$^!*^%)!$%)!(&#$!^#*&^$`
makes a web server thread consume all the CPU time available until the netdata restart.

One request:
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
19421 netdata   20   0  287932  71984  11256 S  99.7   1.8   1:16.77 netdata
```

Three requests:
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
19421 netdata   20   0  288292  72336  11256 S 299.7   1.8   3:58.67 netdata
```

The master branch works well. It gives the answer: `Unsupported v1 API command: (*@&amp;$!$ E)!$* !$* E*!^ 5E$ 5E *^*^ @&amp;$^ ^`

mfundul:
The backtrace for the busy loop is:
```
#0  0x00005555555e4d49 in web_client_split_path_query (w=w@entry=0x55555623c970, s=s@entry=0x555556215924 "//api/v1/(*@&$!$%%5E)!$*%&)!$*%%5E*!%5E%!%5E$%!%5E%(!*%5E*%5E%(*@&$%5E%(!%5E")
    at web/server/web_client.c:916
916	            if (!strncmp(test,"3f",2) || !strncmp(test,"3F",2)) {
(gdb) bt
#0  0x00005555555e4d49 in web_client_split_path_query (w=w@entry=0x55555623c970, s=s@entry=0x555556215924 "//api/v1/(*@&$!$%%5E)!$*%&)!$*%%5E*!%5E%!%5E$%!%5E%(!*%5E*%5E%(*@&$%5E%(!%5E")
    at web/server/web_client.c:916
#1  0x00005555555e53d0 in http_request_validate (w=w@entry=0x55555623c970) at web/server/web_client.c:1037
#2  0x00005555555e5a00 in web_client_process_request (w=w@entry=0x55555623c970) at web/server/web_client.c:1483
#3  0x00005555555e84a8 in web_server_rcv_callback (pi=0x555555bb7390, events=0x555555bb7234) at web/server/static/static-threaded.c:255
#4  0x000055555558c3aa in poll_events_process (now=7834, revents=1, pf=0x555555bb7230, pi=0x555555bb7390, p=0x7fffeeb7c730) at libnetdata/socket/socket.c:1318
#5  poll_events (sockets=<optimized out>, add_callback=add_callback@entry=0x5555555e8add <web_server_add_callback>, del_callback=del_callback@entry=0x5555555e89e3 <web_server_del_callback>, 
    rcv_callback=rcv_callback@entry=0x5555555e8448 <web_server_rcv_callback>, snd_callback=snd_callback@entry=0x5555555e836d <web_server_snd_callback>, 
    tmr_callback=tmr_callback@entry=0x5555555e80b5 <web_server_tmr_callback>, access_list=0x555555a8bea0, data=0x0, tcp_request_timeout_seconds=60, tcp_idle_timeout_seconds=60, 
    timer_milliseconds=1000, timer_data=0x555555b81af0, max_tcp_sockets=64) at libnetdata/socket/socket.c:1592
#6  0x00005555555e8ffc in socket_listen_main_static_threaded_worker (ptr=0x555555b81af0) at web/server/static/static-threaded.c:389
#7  0x000055555558ef56 in thread_start (ptr=0x555555b9a6c0) at libnetdata/threads/threads.c:126
#8  0x00007ffff66106db in start_thread (arg=0x7fffeeb7d700) at pthread_create.c:463
#9  0x00007ffff633988f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

web/server/web_client.c
```diff
@@ -891,29 +857,149 @@ static inline HTTP_VALIDATION http_request_validate(struct web_client *w) {
...
+    //I am not using the strcasestr, because it is fastest to check %3f and compare
+    //the next character.
+    //We executed some tests with "encodeURI(uri);" described in https://www.w3schools.com/jsref/jsref_encodeuri.asp
+    //on July 1st, 2019, that show us that URLs won't have '?','=' and '&' encoded, but we decided to move in front
+    //with the next part, because users can develop their own encoded that won't follow this rule.
+    char *moveme = s;
+    ptr = strchr(moveme, '%');
+    while (moveme) {
+        if(ptr) {
+            char *test = (ptr+1);
```
mfundul: `ptr` never changes in this `while (moveme)` loop and as a result `moveme = ptr` will never cause a break out of this loop.

libnetdata/url/url.c
```diff
@@ -79,3 +79,151 @@ char *url_decode_r(char *to, char *url, size_t size) {
...
+        //Message cannot be complete when first and last address are the same
+        return 0;
+    }
+
+    //This math to verify  the last is valid, because we are discarding the POST
+    if (length >= 4) {
+        begin = end - 4;
+    }
+
+    return ((strstr(begin,"\r\n\r\n")))?1:0;
```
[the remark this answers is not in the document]
mfundul: Cause there may be more characters in between, `strstr` will look try to match the string if it's further down the buffer.

libnetdata/url/url.c
mfundul: `if (length >= 4) { ` must change to `if (length > 4) {`. Also, this function is called in two places, and the first place that invokes it may have already done `last_pos -= 4;` which should probably be removed from there.

thiagoftsm: I copied and paste from the old algorithm, and I appended wrong the part of the text, I am fixing for the next commit.

libnetdata/url/url.c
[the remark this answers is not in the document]
thiagoftsm: Complementing @mfundul comments, we always work with the supposition that we will have the characters "\r\n\r\n" as the end of the request for the majority of the URLs request, except case the request is a POST, in this case it is a divisor. This is a code that was already in the URL parser.

libnetdata/url/url.h
```diff
@@ -25,4 +25,24 @@ extern char *url_decode(char *str);
...
 
+#define WEB_FIELDS_MAX 400
+extern int url_map_query_string(char **out, char *url);
+extern void url_parse_query_string(char *output, size_t max, char **map, int total);
+
+// http_request_validate()
+// returns:
+// = 0 : all good, process the request
+// > 0 : request is not supported
+// < 0 : request is incomplete - wait for more data
```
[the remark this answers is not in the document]
thiagoftsm: This is a comment that is already present in the file https://github.com/netdata/netdata/blob/master/web/server/web_client.c , so I only kept it.

libnetdata/url/url.h
```diff
@@ -25,4 +25,24 @@ extern char *url_decode(char *str);
...
+
+typedef enum {
+    HTTP_VALIDATION_OK,
+    HTTP_VALIDATION_NOT_SUPPORTED,
+    HTTP_VALIDATION_INCOMPLETE,
+    HTTP_VALIDATION_REDIRECT
+} HTTP_VALIDATION;
+
```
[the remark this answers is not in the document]
thiagoftsm: When I created the previous URL parser revision, I was using this enum in the url.c, this was the motive I kept it here, but I agree that now it is better it stays inside web_client.h
I also removed the comments that I described in the previous.

web/server/web_client.c
thiagoftsm: It must be inside the while, I changed it! Thanks!

libnetdata/url/url.h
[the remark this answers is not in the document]
thiagoftsm: I reviewed the spaces and I changed the indentation of the shell script from tab to space. I also saw that the shell script was generating a big sample of data (> 10 GB), so I reduced the number of tests with it.

thiagoftsm:
> The master branch works well. It gives the answer: `Unsupported v1 API command: (*@&amp;$!$ E)!$* !$* E*!^ 5E$ 5E *^*^ @&amp;$^ ^`

The problem was a correct variable in a wrong place, @mfundul found the error and I fixed it. Thanks @mfundul !

web/server/web_client.c
thiagoftsm: @vlvkobal this was the problem of the netdata consumes 100%.

mfundul:
> 
> 
> > The master branch works well. It gives the answer: `Unsupported v1 API command: (*@&amp;$!$ E)!$* !$* E*!^ 5E$ 5E *^*^ @&amp;$^ ^`
> 
> The problem was a correct variable in a wrong place, @mfundul found the error and I fixed it. Thanks @mfundul !

@vlvkobal is the one who discovered the problem [here](https://github.com/netdata/netdata/pull/6247#issuecomment-513190995). You should include this test in your scripts.

thiagoftsm:
> > > The master branch works well. It gives the answer: `Unsupported v1 API command: (*@&amp;$!$ E)!$* !$* E*!^ 5E$ 5E *^*^ @&amp;$^ ^`
> > 
> > 
> > The problem was a correct variable in a wrong place, @mfundul found the error and I fixed it. Thanks @mfundul !
> 
> @vlvkobal is the one who discovered the problem [here](https://github.com/netdata/netdata/pull/6247#issuecomment-513190995). You should include this test in your scripts.

I agree 100% , I am working in the script right now, using the shellcheck on it, when I upload the newest version of the script it will be there.

thiagoftsm:
@mfundul and @vlvkobal the newest script is bringing the test that you did and showed a bug in the previous implementation, the new script is also 100% according all tests done by shellcheck, no less important it does not have the final extension ".sh", I adjusted the Makefile to follow the same pattern present in the other scripts.

vlvkobal:
```
In file included from libnetdata/url/../libnetdata.h:7,
                 from libnetdata/url/url.c:3:
libnetdata/url/url.c: In function 'url_decode_r':
libnetdata/url/url.c:243:29: warning: pointer targets in passing argument 1 of 'utf8_check' differ in signedness [-Wpointer-sign]
  243 |     if(unlikely( utf8_check(to) )) //NULL means sucess here
      |                             ^~
      |                             |
      |                             char *
./config.h:356:41: note: in definition of macro 'unlikely'
  356 | #define unlikely(x) __builtin_expect(!!(x), 0)
libnetdata/url/url.c:158:42: note: expected 'unsigned char *' but argument is of type 'char *'
  158 | unsigned char *utf8_check(unsigned char *s)
      |                           ~~~~~~~~~~~~~~~^
```

thiagoftsm:
> ```
> In file included from libnetdata/url/../libnetdata.h:7,
>                  from libnetdata/url/url.c:3:
> libnetdata/url/url.c: In function 'url_decode_r':
> libnetdata/url/url.c:243:29: warning: pointer targets in passing argument 1 of 'utf8_check' differ in signedness [-Wpointer-sign]
>   243 |     if(unlikely( utf8_check(to) )) //NULL means sucess here
>       |                             ^~
>       |                             |
>       |                             char *
> ./config.h:356:41: note: in definition of macro 'unlikely'
>   356 | #define unlikely(x) __builtin_expect(!!(x), 0)
>       |                                         ^
> libnetdata/url/url.c:158:42: note: expected 'unsigned char *' but argument is of type 'char *'
>   158 | unsigned char *utf8_check(unsigned char *s)
>       |                           ~~~~~~~~~~~~~~~^
> ```

This was brought from other PR, but I fixed it now creating a cast!

vlvkobal:
You've forgotten to include `tests/urls/request.sh` to `.gitignore`.

thiagoftsm:
> You've forgotten to include `tests/urls/request.sh` to `.gitignore`.

I changed it, I really did not remember the existence of this file. Thank you very much!

PR #6264, opened 2019-06-11
Update README.md

mfundul:
##### Summary
Fix documentation to display correct openSUSE dependency.


ilyam8:
Since you are on the Readme

I noticed

> # CentOS / Red Hat Enterprise Linux
yum install autoconf automake curl gcc git libmnl-devel libuuid-devel openssl-devel libuv-devel lz4-devel Judy-devel lm_sensors make MySQL-python nc pkgconfig python python-psycopg2 PyYAML zlib-devel

This is the only distro that have `python` stuff on the list, i suggest to remove them:
 - MySQL-python
 - python
 - python-psycopg2
 - PyYAML


`lm_sensors` is not needed too i think

ilyam8:
@paulkatsoulakis have a look at those lists btw, i guess they could include some other unneeded packages.

we are missing `Arch` on that list btw

mfundul:
Also need to add about OKay repository for CentOS 6 for libuv1 (https://okay.com.mx/blog-news/rpm-repositories-for-centos-6-and-7.html) and for Arch arch/aur for libjudy.

database/engine/README.md
```diff
@@ -106,4 +106,11 @@ There are explicit memory requirements **per** DB engine **instance**, meaning *
 An important observation is that RAM usage depends on both the `page cache size` and the 
 `dbengine disk space` options. 
 
+## File descriptor requirements
+
+The Database Engine may keep a **significant** amount of files open per instance (e.g. per streaming
+slave or master server). When configuring your system you should make sure there are at least 60
+file descriptors available per `dbengine` instance.
+
```
cakrit: Give them some more info here. Each host streaming data to a master is considered a separate dbengine instance, you need to increase the limit in the systemd service (if not running systemd then where do I do it), mention also that the other things that use fds are the web server and the collector plugins and provide an example sizing (e.g. from stackscale).

packaging/installer/README.md
```diff
@@ -202,11 +202,11 @@ dnf install zlib-devel libuuid-devel libuv-devel lz4-devel Judy-devel openssl-de
 yum install autoconf automake curl gcc git libmnl-devel libuuid-devel openssl-devel libuv-devel lz4-devel Judy-devel lm_sensors make MySQL-python nc pkgconfig python python-psycopg2 PyYAML zlib-devel
 
 # openSUSE
-zypper install zlib-devel libuuid-devel libuv-devel liblz4-devel judy-devel openssl-devel libmnl-devel gcc make git autoconf autoconf-archive autogen automake pkgconfig curl findutils
+zypper install zlib-devel libuuid-devel libuv-devel liblz4-devel judy-devel libopenssl-devel libmnl-devel gcc make git autoconf autoconf-archive autogen automake pkgconfig curl findutils
 
 ```
 
-Please note that for RHEL/CentOS you might need [EPEL](http://www.tecmint.com/how-to-enable-epel-repository-for-rhel-centos-6-5/).
+Please note that for RHEL/CentOS you might need [EPEL](http://www.tecmint.com/how-to-enable-epel-repository-for-rhel-centos-6-5/). Especially for using `memory mode = dbengine` EPEL is strictly required. In addition, RHEL/CentOS version 6 also needs [OKay](https://okay.com.mx/blog-news/rpm-repositories-for-centos-6-and-7.html) for package libuv version 1. Arch linux needs arch/aur for package Judy.
```
cakrit: Move the information inside the system-specific list:
```
Alpine Linux and its derivatives (you have to install bash yourself, before using the installer)
Arch Linux and its derivatives
Gentoo Linux and its derivatives
Debian Linux and its derivatives (including Ubuntu, Mint)
Fedora and its derivatives (including Red Hat Enterprise Linux, CentOS, Amazon Machine Image)
SuSe Linux and its derivatives (including openSuSe)
SLE12 Must have your system registered with Suse Customer Center or have the DVD. See #1162
```

We want everyone to have dbengine enabled, so it's not a SHOULD, but a MUST.

database/engine/README.md
```diff
@@ -112,5 +112,34 @@ The Database Engine may keep a **significant** amount of files open per instance
 slave or master server). When configuring your system you should make sure there are at least 60
 file descriptors available per `dbengine` instance.
 
+Netdata allocates 50% of available file descriptors to its web server sockets, 25% to its statsd sockets
+and the rest is available to the daemon and the dbengine. This means that only 25% of the allocated file
+descriptors are accessible by the dbengine. You should take that into account when configuring your service
+or global file descriptor limits.
+
+If for example one wants to allocate 65536 file descriptors to the netdata service on a systemd system
+she needs to edit the netdata systemd file `/etc/systemd/system/multi-user.target.wants/netdata.service`
```
ilyam8: she? :woman:

mfundul: We have to respect our female users as well :D

ilyam8: `Did You Just Assume My Gender?` meme.jpg :girl:

cakrit: The correct grammar is "If one ... then one ...". We don't use this form though, we just say "you". You really don't want get into gender-specific pronouns, especially nowadays.

database/engine/README.md
```diff
@@ -106,4 +106,41 @@ There are explicit memory requirements **per** DB engine **instance**, meaning *
...
+file descriptors available per `dbengine` instance.
+
+Netdata allocates 25% of the available file descriptors to its Database Engine instances. This means that only 25%
+of the file descriptors that are available to the Netdata service are accessible by dbengine instances.
+You should take that into account when configuring your service
+or system-wide file descriptor limits. You can roughly estimate that the netdata service needs 2048 file
+descriptors for every 10 streaming slave hosts when streaming is configured to use `memory mode = dbengine`.
+
+If for example one wants to allocate 65536 file descriptors to the netdata service on a systemd system
+one needs to edit the netdata systemd file `/etc/systemd/system/multi-user.target.wants/netdata.service`
```
paulkatsoulakis: As per the netdata-installer.sh, method install_netdata_service() from functions.sh library, we don't put netdata.service under `multi-user.target.wants`


```374                         SYSTEMD_DIRECTORY=""
375 
376                         if [ -d "/lib/systemd/system" ]; then
377                                 SYSTEMD_DIRECTORY="/lib/systemd/system"
378                         elif [ -d "/usr/lib/systemd/system" ]; then
379                                 SYSTEMD_DIRECTORY="/usr/lib/systemd/system"
380                         fi
381 
382                         if [ "${SYSTEMD_DIRECTORY}x" != "x" ]; then
383                                 echo >&2 "Installing systemd service..."
384                                 run cp system/netdata.service "${SYSTEMD_DIRECTORY}/netdata.service" &&
385                                         run systemctl daemon-reload &&
386                                         run systemctl enable netdata &&
387                                         return 0
388                         else
389                                 echo >&2 "no systemd directory; cannot install netdata.service"
390                         fi
```

mfundul: I used `netdata-installer.sh` though and `netdata.service` was put under `multi-user.target.wants`. Please check the host netdata-deb9-test-mfundul.netdata.rocks. There's something we're missing.

mfundul: It's possible that systemctl creates `/etc/systemd/system/multi-user.target.wants/netdata.service` after installing the service.

mfundul: Here it is:
```
lrwxrwxrwx 1 root root 35 Jun 18 09:32 /etc/systemd/system/multi-user.target.wants/netdata.service -> /lib/systemd/system/netdata.service
```
This is probably what `systemctl enable` creates.

Issue #2413, opened 2017-07-04
riak and riak-cs support


promethe42:
Hi there,

fantastic work on netdata!
I would like to know if there is anything in the roadmap regarding riak and riak-cs support.
They are both supported by datadog:

http://docs.datadoghq.com/integrations/riak/
http://docs.datadoghq.com/integrations/riakcs/

and I would love to be able to track those in netdata.
We would be happy to help and make a PR if anyone can point us in the right direction.

Thanks!

l2isbad:
hello @promethe42 

> We would be happy to help and make a PR if anyone can point us in the right direction.

from riak docs

> Riak provides data related to current operating status, which includes statistics in the form of counters and histograms. These statistics are made available through the HTTP API via the /stats endpoint.

https://github.com/DataDog/integrations-core/tree/master/riak

promethe42:
thx @l2isbad 

I guess what we need is here:

https://github.com/DataDog/integrations-core/blob/master/riak/check.py

and the relevant netdata doc is here:

https://github.com/firehol/netdata/wiki/How-to-write-new-module

I could mimic the [RabbitMQ plugin](https://github.com/firehol/netdata/blob/master/python.d/rabbitmq.chart.py) since it also uses an HTTP API to build the charts.

Correct?

l2isbad:
Yeah. RabbitMQ plugin uses [threading](https://github.com/firehol/netdata/blob/master/python.d/rabbitmq.chart.py#L148-L155) to get data from [different API calls](https://github.com/firehol/netdata/blob/master/python.d/rabbitmq.chart.py#L131-L132). You don't need this stuff.

promethe42:
thx @l2isbad 

Another HTTP based plugin to recommend as a source then?

l2isbad:
[tomcat](https://github.com/firehol/netdata/blob/master/python.d/tomcat.chart.py)

In general you only need to create  custom `_get_data method` (if there is no reason to overwrite UrlService check method).

```python
def _get_data(self):
    raw = self._get_raw_data()    # returns raw data or None
    if not raw:
        return None  # 
    code to parse the data...
```

promethe42:
Nice!
And should I do an external plugin or a PR?

l2isbad:
PR with riak python module (riak.chart.py and riak.conf) for python.d.plugin

You can start from
```python
# -*- coding: utf-8 -*-
# Description: riak netdata python.d module
from base import UrlService
from json import loads

[...]

PR #6286, opened 2019-06-14
Add a riak plugin

jchristgit:
##### Summary

Fixes: #2413

Add a plugin for monitoring Riak KV instances.

##### Component Name

collectors/python.d.plugin/riak

##### Additional Information

As discussed in #2413, this adds a plugin for monitoring Riak KV instances. The plugin uses Riak's /stats endpoint to determine statistics about it.

Healthchecks are also included, but I did not get them to display in the Alarms page on the web dashboard, which leads me to think that they might not work as expected. I dropped the configuration file into `/etc/netdata/health.d/riak.conf` but it seems there's something going wrong there. Would be happy for any pointers.

As far as the [Pull Request Checklist](https://docs.netdata.cloud/collectors/python.d.plugin/#pull-request-checklist-for-python-plugins) goes, it seems some paths over there are changed, so I hope I found the correct configuration files :smile:


collectors/python.d.plugin/python.d.conf
```diff
@@ -88,6 +88,7 @@ nginx_log: no
 # redis: yes
 # rethinkdbs: yes
 # retroshare: yes
+# riak: yes
```
ilyam8: i think we need to name it `riakkv` or `riakkv`

collectors/python.d.plugin/riak/riak.chart.py
```diff
@@ -0,0 +1,528 @@
...
+        self.keys = list(KEYS)
+        for k in ["mean", "median", "95", "99", "100"]:
+            for m in STAT_KEYS:
+                self.keys.append(m + "_" + k)
+        for k in ["min", "max", "mean", "median", "95", "99", "999"]:
+            for m in SEARCH_LATENCY_KEYS:
+                self.keys.append(m + "_" + k)
+        for k in ["min", "max", "mean", "median", "total"]:
+            for m in VNODEQ_KEYS:
+                self.keys.append(m + "_" + k)
```
ilyam8: I suggest to move keys building from init. For example, you could make it a simple function

```python

def keys():
    ...

....
....

self.keys = keys()

```

collectors/python.d.plugin/riak/riak.chart.py
```diff
@@ -0,0 +1,528 @@
...
+        """
+        Format data received from http request
+        :return: dict
+        """
+        raw = self._get_raw_data()
+        if not raw:
+            return None
+
+        data = dict()
+        stats = loads(raw)
```
ilyam8: use `try, except` for loads `(TypeError, ValueError)`

ilyam8:
@jchristgit very nice :+1: 	And please sign the CLA

collectors/python.d.plugin/riak/riak.chart.py
```diff
@@ -0,0 +1,528 @@
...
+# https://docs.riak.com/riak/kv/latest/using/reference/statistics-monitoring/index.html
+
+from json import loads
+
+from bases.FrameworkServices.UrlService import UrlService
+
+# default module values (can be overridden per job in `config`)
+update_every = 2
+priority = 60000
+retries = 60
```
ilyam8: `priority = 60000` and `retries = 60` are defaults, remove them

> update_every = 2

please add a comment why default update every is 2, and add it to the README to

jchristgit:
@ilyam8 Thanks for the feedback! I'll apply and test your suggestions on Monday to make sure I don't coincidentally break something :D

A question regarding the CLA - from my understanding the CLA suggests that I wrote all of the code myself, but the original base implementation was written by you (see https://github.com/netdata/netdata/issues/2413#issuecomment-312923018). Should I add a comment in the code for attribution or how does that work?

jchristgit:
After some more thinking, could we get rid of the manual `if x in keys` when loading the response altogether? It would remove a lot of code that we currently have for figuring out which keys to parse in the response (and a lot of them are unused currently already)

ilyam8:
> Should I add a comment in the code for attribution or how does that work?

Nah


> After some more thinking, could we get rid of the manual if x in keys when loading the response altogether? I

yes, just return `json.loads` result

collectors/python.d.plugin/python.d.conf
jchristgit: Renamed in d7f04df1db2cb0967b8482fa0a4ecf6a50eb1795.

collectors/python.d.plugin/riak/riak.chart.py
jchristgit: Custom keys building has been removed altogether in 7469f4d2d13abd74c1b5c9a0ba1d9eebc43f7496 per my comment below.

collectors/python.d.plugin/riak/riak.chart.py
jchristgit: Wrapped in `try` / `except` in fa4e72e18b237a1333f8c2aad2e66762eb47940d. Is there anything else I should do in case of error besides `return None`?

collectors/python.d.plugin/riak/riak.chart.py
jchristgit: Documented in ba3df872d7357a877a26b25b669bcd8e04add5c6 with defaults removed.

collectors/python.d.plugin/python.d.conf
ilyam8: riak_kv

web/gui/dashboard.js
```diff
@@ -4194,23 +4194,6 @@ NETDATA.peityChartCreate = function (state, data) {
...
-
-    // Round based on number of decimal places to show
-    var precision = Math.pow(10, decimalPlaces);
-    var value = Math.round(data.result[0] * precision) / precision;
-
-    state.element.textContent = prefix + value + suffix;
-    return true;
-}
-
-NETDATA.textOnlyUpdate = NETDATA.textOnlyCreate;
```
ilyam8: why these changes?

collectors/python.d.plugin/python.d.conf
jchristgit: Missed that! Fixed in fe218783995f039712fc4feca8b446a331c64762.

web/gui/dashboard.js
jchristgit: Honestly, no clue. Reset to master in 03ba0d113bcc72f6b0e3f3e56625a6ee1a6501e8.

web/gui/dashboard_info.js
```diff
@@ -267,6 +267,12 @@ netdataDashboard.menu = {
         info: 'Performance metrics for <b>RetroShare</b>. RetroShare is open source software for encrypted filesharing, serverless email, instant messaging, online chat, and BBS, based on a friend-to-friend network built on GNU Privacy Guard (GPG).'
     },
 
+    'riak': {
```
ilyam8: `riak_kv`

@vlvkobal 	please remind me  - do we have any problems with `_` underscore in names ? i recall you faced some issues

web/gui/dashboard_info.js
```diff
@@ -267,6 +267,12 @@ netdataDashboard.menu = {
...
+        title: 'Riak',
```
ilyam8: Riak KV

collectors/python.d.plugin/riak/riak.chart.py
ilyam8: no :+1:

collectors/python.d.plugin/riak_kv/riak_kv.chart.py
```diff
@@ -0,0 +1,314 @@
...
+        Format data received from http request
+        :return: dict
+        """
+        raw = self._get_raw_data()
+        if not raw:
+            return None
+
+        try:
+            return loads(raw)
+        except (TypeError, ValueError):
```
ilyam8: oh, i forgot, my bad

please log the error if any

web/gui/dashboard_info.js
jchristgit: @ilyam8 Doesn't this apply to the metrics though? The metrics are namespaced under `riak.` currently (e.g. `riak.kv`, `riak.dt` and so on). I tested this with `riak_kv` earlier but I recall the information here wasn't applied anymore

collectors/python.d.plugin/riak_kv/riak_kv.chart.py
jchristgit: Added in 49e8aee8d9ed99808a77c44b7bfa0beaf358dd9e.

web/gui/dashboard_info.js
ilyam8: > I tested this with riak_kv earlier but I recall the information here wasn't applied anymore

That is why i summon @vlvkobal 	:smiling_imp:

web/gui/dashboard_info.js
ilyam8: > title

jchristgit: Changed in 71baea7939079c01b75878ac094681a5c093e16d.

web/gui/dashboard_info.js
ilyam8: Thanks. So we need to remove `_` from the name. How about to name it simply `riakkv` @jchristgit?

jchristgit: Yup, renamed in bfcaf640dbfd7ac7afba8bcf98f26bed110803b5.

jchristgit:
@ilyam8 Any idea about the healthchecks not displaying in the alarms page on the web dashboard?

ilyam8:
@jchristgit 
it seems you need to add them manually to the
https://github.com/netdata/netdata/blob/master/health/Makefile.am

ilyam8:
@jchristgit i have no `riak kv`, so i cant properly test the collector. Please make sure it works on py2/py3, it survives riak restart and there is no problems with alarms (valid calculations, etc).

ilyam8:
Very nice work @jchristgit 	:+1: LGTM

jchristgit:
@ilyam8 Thanks for the feedback.

As for the healthchecks, it seems that the Makefile entry was also missing for proper installation - I dropped the health configuration file as-is into `/etc/netdata/health.d/riakkv.conf` on the machine I'm testing with. However, it seems like that's not being processed in the first place, which is my issue. I will check the other points you mentioned.

ilyam8:
>  I dropped the health configuration file as-is into /etc/netdata/health.d/riakkv.conf

make sure the file is readable by `netdata` user

___

and it is not needed, installer will add the alarm, just intall your branch

jchristgit:
Tested surviving Riak restarts and Python 2 compatability, both work.

The configuration file is readable for netdata - no error in the logfile either (disabled log flood protection to make sure I'm not missing something).

Maybe it'd be better to leave out the health configuration setup from this PR until I have it figured out? I can open up a separate PR once I get it to work.

ilyam8:
>  I just figured this out - I need to reference the context, not the context + the metric. I'll update the plugin accordingly

this is true for templates, yes

jchristgit:
I've simplified two metric names, fixed the existing health checks and added two more health checks. My colleague does not have any other ideas for health checks or metrics to collect either, so this would be done from my side.

ilyam8:
i installed the PR 

> [ilyam-pc plugins.d]# ./python.d.plugin riakkv
2019-06-18 15:51:31: python.d INFO: plugin[main] : using python v3
2019-06-18 15:51:31: python.d ERROR: plugin[main] : unknown modules : ['riakkv']
DISABLE

please add the plugin to the
https://github.com/netdata/netdata/blob/83d6968b19101accc4f424aa8a2a49487a8d8b25/collectors/python.d.plugin/Makefile.am

jchristgit:
Added to the Makefile. Also fixed an issue related to looking up the current value on latency healthchecks instead of the three-minute average.

ilyam8:
i installed the branch, lgtm

PR #6291, opened 2019-06-15
doc: remove single/multi-threaded web server configuration

nodiscc:
##### Summary

Support for single/multi-threaded server configuration was removed in https://github.com/netdata/netdata/pull/4969 (only the static-threaded server was kept), but https://docs.netdata.cloud/docs/performance/#6-use-the-single-threaded-web-server still recommends configuring it as single-threaded.

This removes the obsolete paragraph.

##### Component Name

Documentation


cakrit:
Excellent, thanks!

PR #6303, opened 2019-06-18
Handle file descriptors running out

### LINKED ISSUE #6265: netdata/daemon: Service down with multiple "too many open files" occurring on DB engine
##### Bug report summary
As discussed we need to investigate the root cause of netdata crash on one of our demo servers due to maximum open files limit.

- Netdata should not crash when max file descriptors are reached.

- Netdata should allocate a FD budget for dbengine and inform the user of the typical requirements. The user should be able to configure the service FD budget depending on her needs such as the amount of streaming slaves connected to this master.

- Netdata should not allow a new streaming slave to connect to this master if the FD budget has run out.

- In the unlikely event of descriptors running out dbengine will continue writing on the same data and journal files until the problem is resolved. The error condition will be recorded with specific statistics and a new alarm will be generated.

CC: @cakrit @ktsaou 
##### OS / Environment
Any

Latest

##### Component Name
netdata/database

##### Steps To Reproduce
Run netdata on a system with very low max file connections setting to get a reproduction

##### Expected behavior
Netdata should not crash when max open file limits are reached

mfundul:
##### Summary
Fixes #6265
##### Component Name
database/engine
##### Additional Information
Fixes cases where netdata dbengine would abort when running out of file descriptors. When dbengine runs out of descriptors:
- during instance start-up, all state is rolled back and the rrd host fails to initialize
- during normal operation new file creation is postponed and retried indefinitely
- FS error statistics are gathered per dbengine instance and globally
- a dbengine FS error alarm is generated based on dbengine global FS error statistics


mfundul:
Still need to:
- create alarm
- handle file descriptor budget?

mfundul:
We can merge this as is since it's already an improvement, but we should discuss how and if we want to handle budgeting of file descriptors. In its current form, dbengine can handle any and all errors of "too many open files" without crashing netdata.

paulkatsoulakis:
We should push this branch to stackscale octopussy before we merge

database/engine/datafile.c
```diff
@@ -205,15 +236,21 @@ static int load_data_file(struct rrdengine_datafile *datafile)
...
+    error = ret;
+    ret = uv_fs_close(NULL, &req, file, NULL);
+    if (ret < 0) {
+        error("uv_fs_close(%s): %s", path, uv_strerror(ret));
+        ++ctx->stats.fs_errors;
+        rrd_stat_atomic_add(&global_fs_errors, 1);
+    }
     uv_fs_req_cleanup(&req);
-    return ret;
+    return error;
```
paulkatsoulakis: The variable error i think is useless.
Just use ret as it was, unless i miss something from the rest of the context, but probably is just a leftover

database/engine/journalfile.c
```diff
@@ -13,7 +13,7 @@ static void flush_transaction_buffer_cb(uv_fs_t* req)
 
     uv_fs_req_cleanup(req);
     free(io_descr->buf);
```
paulkatsoulakis: why not freez here too? I assume combined with using the respective mallocz too (haven't checked the declaration)

database/engine/journalfile.c
```diff
@@ -162,19 +189,21 @@ int create_journal_file(struct rrdengine_journalfile *journalfile, struct rrdeng
...
-    }
-    if (req.result < 0) {
-        fatal("uv_fs_write: %s", uv_strerror((int)req.result));
+        assert(req.result < 0);
+        error("uv_fs_write: %s", uv_strerror(ret));
+        ++ctx->stats.io_errors;
+        rrd_stat_atomic_add(&global_io_errors, 1);
     }
     uv_fs_req_cleanup(&req);
     free(superblock);
```
paulkatsoulakis: freez/mallocz?

database/engine/journalfile.c
```diff
@@ -433,9 +464,15 @@ int load_journal_file(struct rrdengine_instance *ctx, struct rrdengine_journalfi
...
+    error = ret;
+    ret = uv_fs_close(NULL, &req, file, NULL);
+    if (ret < 0) {
+        error("uv_fs_close(%s): %s", path, uv_strerror(ret));
+        ++ctx->stats.fs_errors;
+        rrd_stat_atomic_add(&global_fs_errors, 1);
+    }
     uv_fs_req_cleanup(&req);
-    return ret;
+    return error;
```
paulkatsoulakis: same like before, we don't really need "error" variable here.

database/engine/datafile.c
mfundul: No this is a bugfix, look carefully.

database/engine/journalfile.c
mfundul: This is a bugfix, look carefully.

database/engine/journalfile.c
mfundul: there is no mallocz here.

database/engine/journalfile.c
paulkatsoulakis: mallocz not used on purpose because ?

mfundul: did you look at the code?

paulkatsoulakis: My question is:
Why use malloc/free in the first place and not use "mallocz/freez" everywhere consistently?

mfundul: I don't use malloc, please look at the code before doing the code review.

paulkatsoulakis: Oh i see, so we use `posix_memalign()` for performance, got it

database/engine/journalfile.c
paulkatsoulakis: OMG that's why i didn't see it, i had to track down the gotos.

PR #6311, opened 2019-06-20
fix elasticsearch plugin

Wing924:
##### Summary

* Fix cluster_health_nodes should be area not stack because `data_nodes` is a part of `nodes`.
* Fix cluster_stats_nodes not show correct metrics.

sample raw data:
```json
"nodes": {
  "count": {
    "total": 10,
    "data": 7,
    "coordinating_only": 0,
    "master": 3,
    "ingest": 7
  }
}
```

##### Component Name
Python Collector
Elasticsearch module


ilyam8:
Hey @Wing924 thanks you!

Issue #6204, opened 2019-06-04
Pihole stats modules 


modem7:
##### Feature idea summary
Pi-Hole monitoring

##### Expected behavior
As netdata works excellently on a pi, may I suggest some form of pihole stats monitoring? 

Eg. 
DNS
DHCP
Queries blocked 
Total queries 
Percent blocked
And maybe (unsure if there's any point in this one as the value rarely changes significantly) number of domains blocked. 

Pi-Hole now is available with an api, so it we should be able to get the information from it much like we can with a remote nginx instance.

ilyam8:
Hi @modem7 	

> Pi-Hole now is available with an api, so it we should be able to get the information from it much like we can with a remote nginx instance.

Could you provide a link to the api readme with all endpoints etc.? i see it has API v2 and v3.

i found this app
https://github.com/Qrome/Pi-hole-Monitor

it uses `/admin/api.php?summary`.

modem7:
This is what I can find:

https://discourse.pi-hole.net/t/pi-hole-api/1863
https://github.com/pi-hole/api
https://pi-hole.net/2018/10/15/announcing-our-restful-api-contributions-welcome/

api.php?summary shows the following:
![Capture1](https://user-images.githubusercontent.com/4349962/58895538-0f621200-86ec-11e9-81fb-377d4d6d12c7.JPG)
![Capture2](https://user-images.githubusercontent.com/4349962/58895539-0f621200-86ec-11e9-9537-811e9c894664.JPG)

Hope that helps some?

ilyam8:
@modem7 

> api.php?summary

do you mean `/admin/api.php?summary` ?

i am not clear, with `api key` you have completely different response for same requests? or you get some info in addition?

What is `API key`? http basic auth?

modem7:
Within PiHole there is an API key they call "WebPassword", which is similar in format to 7547b7cf6fb7947a7cf5bc3850b3614cf943c9867cb324543fe93739494857cd.

On here: https://discourse.pi-hole.net/t/pi-hole-api/1863 - you can see what values you are able to retrieve with and without the API key.

So https://pihole.myiphost.com/admin/api.php?**topClients**&auth=webpassword would allow you to view the top clients as you have authorisation, so it seems that we'd be able to get two types of values:

1. Unauthorized, so we'd be able to get it working out of the box
2. Authorized, needs additional config to enter the web password/api key.

Unsure if that made **any** sense!

ilyam8:
[...]

PR #6320, opened 2019-06-20
pihole collector: add alarms

ilyam8:
##### Summary
Alarm for [`pihole`](https://github.com/netdata/go.d.plugin/tree/master/modules/pihole) collector.

Arlarms to monitor:
 - Pi-hole availability
 - blocked dns queries precentage
 - blocklist last update time
 - gravity.list file existence
 - Pi-hole blocking functionality status (on/off)

Related: #6204

##### Component Name
[health/](https://github.com/netdata/netdata/tree/master/health)

##### Additional Information

Test:
 - i installed the branch
 - checked the dashboard alarm section

Alarms are there.

### DIFF
--- a/health/health.d/pihole.conf
+++ b/health/health.d/pihole.conf
@@ -1,10 +1,67 @@
- template: pihole_blocked_queries
-      on: pihole.dns_queries_percentage
+
+ # Make sure Pi-hole is responding.
+
+template: pihole_last_collected_secs
+      on: pihole.dns_queries_total
+    calc: $now - $last_collected_t
+   units: seconds ago
    every: 10s
-   units: %
-    calc: $blocked
-    warn: $this > ( ($status >= $WARNING ) ? ( 50 ) : ( 75 ) )
-    crit: $this > ( ($status >= $CRITICAL) ? ( 75 ) : ( 100 ) )
-   delay: down 5m
-    info: dhcp-range utilization above threshold!
-      to: sysadmin
+    warn: $this > (($status >= $WARNING)  ? ($update_every) : ( 5 * $update_every))
+    crit: $this > (($status == $CRITICAL) ? ($update_every) : (60 * $update_every))
+   delay: down 5m multiplier 1.5 max 1h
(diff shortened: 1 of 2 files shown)


ilyam8:
@modem7 i think we are wrong on `blocklist last update` alarm. Yes, the cron job runs every week, but it does not guarantee the list is updated on every run.

So we need to understand if there is an update schedule or no. I hope devs will give some info about it. 

Am i right?

modem7:
Unsure myself, but I would have assume that the counter is reset whenever the cronjob runs, but without knowing a bit more of the code, I couldn't be 100% on that!

ilyam8:
> Unsure myself, but I would have assumed that the counter is reset whenever the cronjob runs, but without knowing a bit more of the code, I couldn't be 100% on that!

could you run the job by hand and see if the counter resets?

modem7:
> 
> 
> > Unsure myself, but I would have assumed that the counter is reset whenever the cronjob runs, but without knowing a bit more of the code, I couldn't be 100% on that!
> 
> could you run the job by hand and see if the counter is reset?

OK, so I ran pihole -g which I can only assume is the same command the cronjob runs, and the timer has been reset to 0 as hoped!

ilyam8:
@modem7

I just realized that I could do it myself :see_no_evil: 	

> pihole -g which I can only assume is the same command the cronjob runs

Same, you are right.

health/health.d/pihole.conf
```diff
@@ -0,0 +1,67 @@
...
+
+  # Blocked DNS queries.
+
+ template: pihole_blocked_queries
+       on: pihole.dns_queries_percentage
+    every: 10s
+    units: %
+     calc: $blocked
+     warn: $this > ( ($status >= $WARNING ) ? ( 50 ) : ( 75 ) )
+     crit: $this > ( ($status >= $CRITICAL) ? ( 75 ) : ( 85 ) )
```
paulkatsoulakis: Shouldn't critical be all the way to 99?
I may miss how exactly the rules are parsed

PR #6345, opened 2019-06-25
Tomcat status invalid XML fix

### LINKED ISSUE #6343: Tomcat collector break on invalid XML caused by single quotes in Memory Pool names (code solution inside)
##### Bug report summary
The Tomcat plugin breaks when collecting data from a Tomcat 8 server running with OpenJDK : The XML returned by the `/manager/status?XML=true` page contains single quotes attributes in the `<jvm>` tags, and the Memory Pool names contains single quotes (ie. `CodeHeap 'non-nmethods'`). Thus producing invalid XML, and provoking the error `(...) is not a vaild XML page. Please add "?XML=true" to tomcat status page` (and there's a minor typo on "vaild", just sayin').

Example: 
```xml
    <jvm>
        (...)
        <memorypool name='CodeHeap 'non-nmethods'' type='Non-heap memory' usageInit='2555904' usageCommitted='3670016' usageMax='7593984' usageUsed='1349376'/>
        <memorypool name='CodeHeap 'non-profiled nmethods'' type='Non-heap memory' usageInit='2555904' usageCommitted='24379392' usageMax='122032128' usageUsed='24303360'/>
        <memorypool name='CodeHeap 'profiled nmethods'' type='Non-heap memory' usageInit='2555904' usageCommitted='35848192' usageMax='122032128' usageUsed='33184384'/>
        <memorypool name='Compressed Class Space' type='Non-heap memory' usageInit='0' usageCommitted='7127040' usageMax='1073741824' usageUsed='6459400'/>
        (...)
    </jvm>
```

##### OS / Environment
Debian GNU/Linux 9 (stretch)
Apache Tomcat/8.5.14
OpenJDK 11.0.3+7-Debian-1bpo92

v1.15.0-114-nightly

##### Component Name
python.d.plugin/tomcat.chart.py

##### Steps To Reproduce
Configure tomcat plugin with a Tomcat 8 server running with OpenJDK 11.

##### Expected behavior
The tomcat plugin should chart the data returned by the status page.

#### Tested solution
I added the following regular expression replace to the XML `raw_data` returned in `tomcat.chart.py` . It removes all redundant single quotes in the attributes values.

```python
        if raw_data:
            raw_data = re.sub(r"='([^']+)'([^']+)''", r"='\g<1>\g<2>'", raw_data)
```
Another possible solution would be to replace all the single quote attributes by double quote ones.

Danamir:
##### Summary
Quick fix to replace single quotes ending content in single quote attributes in Tomcat status XML response, as it would produce an error when parsing the XML.
This is a tentative fix, a more robust approach would be to replace the single quote attributes by double quotes, but it can't be done via a simple regular expression.
Fixes #6343

##### Component Name
`collectors/python.d.plugin/tomcat/tomcat.chart.py`

### DIFF
--- a/collectors/python.d.plugin/tomcat/tomcat.chart.py
+++ b/collectors/python.d.plugin/tomcat/tomcat.chart.py
@@ -105,8 +105,11 @@ class Service(UrlService):
         data = None
         raw_data = self._get_raw_data()
         if raw_data:
-            # Fix attributes ending with single quote content
-            raw_data = re.sub(r"='([^']+)'([^']+)''", r"='\g<1>\g<2>'", raw_data)
+            # Fix XML attributes ending with single quote content, ie. <memorypool name='CodeHeap 'non-nmethods'' ... />
+            # Cf. Issue https://github.com/netdata/netdata/issues/6343
+            pattern = re.compile(r"='([^']+)'([^']+)''")
+            raw_data = pattern.sub(r"='\g<1>\g<2>'", raw_data)
+
             try:
                 xml = ET.fromstring(raw_data)
             except ET.ParseError:
--- a/collectors/python.d.plugin/tomcat/tomcat.chart.py
+++ b/collectors/python.d.plugin/tomcat/tomcat.chart.py
@@ -14,7 +14,6 @@ MiB = 1 << 20
 # Regex fix for Tomcat single quote XML attributes
 # affecting Tomcat < 8.5.24 & 9.0.2 running with Java > 9
 # cf. https://bz.apache.org/bugzilla/show_bug.cgi?id=61603
-fix_tomcat_single_quote = True
 single_quote_regex = re.compile(r"='([^']+)'([^']+)''")
 
 ORDER = [
@@ -122,12 +121,12 @@ class Service(UrlService):
         if not raw_data:
             return False
 
-        if fix_tomcat_single_quote and single_quote_regex.search(raw_data):
+        if single_quote_regex.search(raw_data):
             self.warning('Tomcat status page is returning invalid single quote XML, please consider upgrading '
                          'your Tomcat installation. See https://bz.apache.org/bugzilla/show_bug.cgi?id=61603')
             self.parse = self.xml_single_quote_fix_parse
 
-        return True
+        return self.parse(raw_data)
 
     def _get_data(self):
         """
--- a/collectors/python.d.plugin/tomcat/tomcat.chart.py
+++ b/collectors/python.d.plugin/tomcat/tomcat.chart.py
@@ -95,22 +95,6 @@ CHARTS = {
 }
 
 
-class XMLParser:
-    @staticmethod
-    def parse(data):
-        try:
-            return ET.fromstring(data)
-        except ET.ParseError:
-            return None
-
-
-class XMLSingleQuoteFixParser(XMLParser):
-    @staticmethod
-    def parse(data):
-        data = single_quote_regex.sub(r"='\g<1>\g<2>'", data)
-        return XMLParser.parse(data)
-
-
 class Service(UrlService):
     def __init__(self, configuration=None, name=None):
         UrlService.__init__(self, configuration=configuration, name=name)
@@ -118,6 +102,32 @@ class Service(UrlService):
         self.definitions = CHARTS
         self.url = self.configuration.get('url', 'http://127.0.0.1:8080/manager/status?XML=true')
         self.connector_name = self.configuration.get('connector_name', None)
+        self.parse = self.xml_parse
+
+    def xml_parse(self, data):
+        try:
+            return ET.fromstring(data)
+        except ET.ParseError:
+            self.debug('%s is not a valid XML page. Please add "?XML=true" to tomcat status page.' % self.url)
+            return None
+
+    def xml_single_quote_fix_parse(self, data):
+        data = single_quote_regex.sub(r"='\g<1>\g<2>'", data)
+        return self.xml_parse(data)
+
+    def check(self):
+        self._manager = self._build_manager()
+
+        raw_data = self._get_raw_data()
+        if not raw_data:
+            return False
+
+        if fix_tomcat_single_quote and single_quote_regex.search(raw_data):
+            self.warning('Tomcat status page is returning invalid single quote XML, please consider upgrading '
+                         'your Tomcat installation. See https://bz.apache.org/bugzilla/show_bug.cgi?id=61603')
+            self.parse = self.xml_single_quote_fix_parse
+
+        return True
 
     def _get_data(self):
         """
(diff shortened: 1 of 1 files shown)


collectors/python.d.plugin/tomcat/tomcat.chart.py
```diff
@@ -104,10 +105,12 @@ def _get_data(self):
         data = None
         raw_data = self._get_raw_data()
         if raw_data:
+            # Fix attributes ending with single quote content
+            raw_data = re.sub(r"='([^']+)'([^']+)''", r"='\g<1>\g<2>'", raw_data)
```
ilyam8: please compile the regex

ilyam8: and add some comments why we need it (comments + link the the issue would be nice)

collectors/python.d.plugin/tomcat/tomcat.chart.py
```diff
@@ -105,8 +105,11 @@ def _get_data(self):
         data = None
         raw_data = self._get_raw_data()
         if raw_data:
-            # Fix attributes ending with single quote content
-            raw_data = re.sub(r"='([^']+)'([^']+)''", r"='\g<1>\g<2>'", raw_data)
+            # Fix XML attributes ending with single quote content, ie. <memorypool name='CodeHeap 'non-nmethods'' ... />
+            # Cf. Issue https://github.com/netdata/netdata/issues/6343
+            pattern = re.compile(r"='([^']+)'([^']+)''")
```
ilyam8: move compile to the global area, we dont want compile it on every iteration (on every get_data call)

Danamir: Oh yeah I see, my bad.

collectors/python.d.plugin/tomcat/tomcat.chart.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, configuration=None, name=None):
         self.definitions = CHARTS
         self.url = self.configuration.get('url', 'http://127.0.0.1:8080/manager/status?XML=true')
         self.connector_name = self.configuration.get('connector_name', None)
+        self.single_quote_regex = re.compile(r"='([^']+)'([^']+)''")
```
ilyam8: Python `re` package is thread safe, make it global var and just use in `get_data`. There is no reason to have copy of regex for in every instance

Danamir: Well you've made me realize a whole lot of my other projects code is not optimized !

collectors/python.d.plugin/tomcat/tomcat.chart.py
```diff
@@ -104,11 +127,18 @@ def _get_data(self):
         data = None
         raw_data = self._get_raw_data()
         if raw_data:
-            try:
-                xml = ET.fromstring(raw_data)
-            except ET.ParseError:
-                self.debug('%s is not a vaild XML page. Please add "?XML=true" to tomcat status page.' % self.url)
+            if fix_tomcat_single_quote and single_quote_regex.search(raw_data):
+                self.warning('Tomcat status page is returning invalid single quote XML, please consider upgrading '
+                             'your Tomcat installation. See https://bz.apache.org/bugzilla/show_bug.cgi?id=61603')
```
ilyam8: i think we need to add `check` method to `Service` and select appropiate parser there.
`check` method executed only once.

you need 

```python
def check(self):
    self._manager = self._build_manager()  # this is needed

    if not raw_data:
        return None

     <parser select>
     <assigne parser to self.parse>

     <if selected return true, otherwise false>
```


and i think we don'nt need parser as Classes, simpe funcs will do

Danamir: Something like that ? (cf. next commit)

I'm not sure I understood the `<if selected return true, otherwise false>` part. I always return True as long as the `raw_data` check is ok ; otherwise it does not work.

collectors/python.d.plugin/tomcat/tomcat.chart.py
```diff
@@ -95,6 +102,32 @@ def __init__(self, configuration=None, name=None):
...
+        return self.xml_parse(data)
+
+    def check(self):
+        self._manager = self._build_manager()
+
+        raw_data = self._get_raw_data()
+        if not raw_data:
+            return False
+
+        if fix_tomcat_single_quote and single_quote_regex.search(raw_data):
```
ilyam8: i dont get why we need `fix_tomcat_single_quote`, since it is always `True`
`single_quote_regex.search(raw_data)` should be enough

and change `return True` to `return bool(self.parse(raw_data))`

Danamir: It was a precaution to be able to disable the fix quickly in case it brought problems, maybe with a configuration option. But I can remove it if it's better.

ilyam8: Yes, remove it

Danamir: Ok.

If I put `return self.parse(raw_data)` in `check` method, I still have `xml = self.parse(raw_data)` in the `_get_data` method ; is it a problem if parse is called twice ? Otherwise I could put the result in a `self.xml` variable, but it seems strange to have a content affectation in a check method.

ilyam8: it is ok, `check` is called only once before start.

flow is
 - call `check()`, if check returns 'False` drop the job, otherwise run job in thread
 - every `update_every` call `job.get_data()`

ilyam8: that is why i suggest to `return self.parse(raw_data)` - to make sure we can parse the data  before the start. And i suggest to use `bool` to explicitly show that function returns boolean and not the result of `self.parse` (techically bool is not needed, yes)

ilyam8:
ok, @Danamir looks good, i will do some tests later :+1:

Danamir:
Alright ! 👍

ilyam8:
ok, i tested the PR, lgtm

collectors/python.d.plugin/tomcat/tomcat.chart.py
```diff
@@ -95,6 +101,32 @@ def __init__(self, configuration=None, name=None):
...
+        raw_data = self._get_raw_data()
+        if not raw_data:
+            return False
+
+        if single_quote_regex.search(raw_data):
+            self.warning('Tomcat status page is returning invalid single quote XML, please consider upgrading '
+                         'your Tomcat installation. See https://bz.apache.org/bugzilla/show_bug.cgi?id=61603')
+            self.parse = self.xml_single_quote_fix_parse
+
+        return bool(self.parse(raw_data))
```
ilyam8: one minor thing

> /opt/netdata/usr/libexec/netdata/python.d/tomcat.chart.py:129: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  return bool(self.parse(raw_data))

please change it to `return self.parse(raw_data) is not None`

Danamir: Done !

ilyam8:
thanks for contributing @Danamir

PR #6369, opened 2019-07-01
netdata/packaging: Netdata binary packages generation - spec file refinement, support for nightlies (RPM)

### LINKED ISSUE #5969: [Binary releases] Optimise netdata.spec file
##### Feature idea summary
Optimise netdata.spec file, sync parts of the `netdata-installer.sh` steps to the spec file.
In order to be compliant with the way a SPEC file is structured, we need to decompose netdata-installer.sh artifacts and sync the existing SPEC sections with all the necessary activities that are required for the install to create a consistent netdata installation.

References #5963

paulkatsoulakis:
##### Summary
This PR incorporates improvements in the following areas for binary packaging:

1) Refine netdata.spec.in to generate identical netdata setup, just like our latest `netdata-installer.sh` implementation.
    This change was focusing on multiple areas of interest:
     1) Package dependencies, cross-distribution
     2) Configuration management
     3) Install/uninstall concerns

2) Add support for generating packages over the latest stable release AND the latest nightly code
   We had to modify our pipeline, so that we can either generate packages for our latest stable release or the latest commit available.  To support this we had to introduce two repositories, namely `netdata/netdata` for the stable channel and `netdata/netdata-edge` for the nightlies.

   If you want to build the latest stable release, you just follow what we already have documented.  To trigger a nightlies build, you need to append the keyword `[Build latest]` along with the desired package group you wish to build.

In order for you to test the generated packages, you will need to install my development repository, named `paulkatsoulakis/netdata-edge` and also the `epel-release` if you testing on CentOS.
To do so, you will need to execute the following on your RPM based distribution and it will auto-setup the repo for you
```
curl -s https://e65b0da7d95db9492c1efd2fd06782375a11b6cf46221d49:@packagecloud.io/install/repositories/paulkatsoulakis/netdata-edge/script.rpm.sh | sudo bash
```

Then to install netdata all you need to do is install package `netdata` with the respective package manager for the distro you are testing on.

**Important note**: At this stage, the RPM we will provide will not have support for Kinesis and Xenstat plugins. Those require slightly extra effort that i could not sort out right now, so i will be adding more backlog items to mitigate them at a next iteration

##### With this PR, the following operating systems will receive RPM
* Redhat Enterprise (https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux and https://access.redhat.com/support/policy/updates/errata)
   * Versions 6.* 
   * Versions 7.*

* Fedora (https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle). 
   * Fedora 28
   * Fedora 29
   * Fedora 30

* CentOS (https://en.wikipedia.org/wiki/CentOS#CentOS_releases and https://wiki.centos.org/FAQ/General#head-fe8a0be91ee3e7dea812e8694491e1dde5b75e6d)
   * Centos 6.*
   * Centos 7.*

* OpenSuSE (https://en.opensuse.org/Lifetime)
   * Open SuSE Leap 15.0
   * Open SuSE Leap 15.1

##### Component Name
netdata/packaging
netdata/packaging/ci

##### Additional Information
Closes #5969


thiagoftsm:
Everything worked fine during the installation process, but the post install is missing something, because the netdata was not started automatically.
Congratulations!

paulkatsoulakis:
> I think the python code needs to have "try .. catch" to treat the errors, the way it is right now, you are not having the proper treatment of errors and this code tries to get information from the filesystem that cannot be there due some reason.

Actually, this is intentional.
The steps in the script are very 1-2-3 and there is no real benefit for exceptional handling upon errors.
Any exception will be directly raised on the upper layer (Travis hooks in our case) and they report all the stack trace details on the logs and the process stops.  Then we have Travis on failure triggers to initiate notifications or anything further we might want to do with the failure

thiagoftsm:
> Actually, this is intentional.

Now that you explained, it was clear for me. When I saw the code I only see a possibility of error.

thiagoftsm:
Now it is working correctly! The installation process was 100% successful! To be more explicit I could install and after the installation process it was started.
When I tried to remove, it also kept the netdata.conf that I changed.

Issue #6371, opened 2019-07-02
[BUG] CRITICAL: main[main] SIGPIPE received. Terminating...


weijing24:
### Bug report summary
I have tried everything I can do, but I can't find what cause netdata ternimate.
here is my netdata.log
```
2019-07-01 19:11:02: go.d WARNING: httpcheck[XXX] Get https://XXX/ping: read tcp 172.22.0.2:46298->XXX:443: read: connection reset by peer
2019-07-01 19:11:02: go.d WARNING: httpcheck[XXX] Get https://XXX/ping: read tcp 172.22.0.2:34484->XXX:443: read: connection reset by peer
2019-07-01 19:11:02: go.d WARNING: httpcheck[XXX] Get https://XXX/ping: read tcp 172.22.0.2:35448->XXX:443: read: connection reset by peer
2019-07-01 19:11:02: go.d WARNING: httpcheck[XXX] Get https://XXX/ping: read tcp 172.22.0.2:60538->XXX:443: read: connection reset by peer
2019-07-01 19:11:02: go.d CRITICAL: main[main] SIGPIPE received. Terminating...
2019-07-01 19:11:02: netdata ERROR : PLUGINSD[go.d] : read failed
2019-07-01 19:11:02: netdata ERROR : PLUGINSD[go.d] :'/usr/libexec/netdata/plugins.d/go.d.plugin' (pid 310) disconnected after 549928 successful data collections (ENDs). (errno 22, Invalid argument)
2019-07-01 19:11:02: netdata ERROR : PLUGINSD[go.d] : child pid 310 exited with code 1.
2019-07-01 19:11:02: netdata ERROR : PLUGINSD[go.d] :'/usr/libexec/netdata/plugins.d/go.d.plugin' (pid 310) exited with error code 1, but has given useful output in the past (924156 times). Waiting a bit before starting it again. (errno 22, Invalid argument)
2019-07-01 19:11:46: alarm-notify.sh: INFO: sent slack notification for: XXX httpcheck_XXX.request_status.httpcheck_last_collected_secs is WARNING to '#alarming-bot' 
```
There are many alarms about last_collected_secs, I just pick one example.
here is my /var/docker/netdata/netdata.conf
```
[global]
        history = 7200
        update every = 10
[plugins]
        proc = no
        diskspace = no
        cgroups = no
        tc = no
        idlejitter = no
        enable running new plugins = yes
        check for new plugins every = 60
        python.d = no
        charts.d = no
        node.d = no
[...]

PR #6373, opened 2019-07-02
Add more debug messages for pluginsd pipe errors

vlvkobal:
##### Summary
To debug the root cause of the #6371 issue we need to know more about an error on pipe close.

##### Component Name
pluginsd orchestrator

### DIFF
--- a/collectors/plugins.d/plugins_d.c
+++ b/collectors/plugins.d/plugins_d.c
@@ -154,7 +154,12 @@ inline size_t pluginsd_process(RRDHOST *host, struct plugind *cd, FILE *fp, int 
 
         char *r = fgets(line, PLUGINSD_LINE_MAX, fp);
         if(unlikely(!r)) {
-            error("read failed");
+            if(feof(fp))
+                error("read failed: end of file");
+            else if(ferror(fp))
+                error("read failed: input error");
+            else
+                error("read failed: unknown error");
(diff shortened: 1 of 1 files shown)


collectors/plugins.d/plugins_d.c
```diff
@@ -154,7 +154,12 @@ inline size_t pluginsd_process(RRDHOST *host, struct plugind *cd, FILE *fp, int
 
         char *r = fgets(line, PLUGINSD_LINE_MAX, fp);
         if(unlikely(!r)) {
-            error("read failed");
+            if(feof(fp))
```
paulkatsoulakis: How about some of those to get more fine grained error report for any error
```
NAME
     perror, strerror, strerror_r, sys_errlist, sys_nerr -- system error messages

LIBRARY
     Standard C Library (libc, -lc)

SYNOPSIS
     #include <stdio.h>

     void
     perror(const char *s);

     extern const char * const sys_errlist[];
     extern const int sys_nerr;

     #include <string.h>

     char *
     strerror(int errnum);

     int
     strerror_r(int errnum, char *strerrbuf, size_t buflen);
```

man perror should give you the details

mfundul: netdata `error` already does `strerror()` inside.

Issue #6378, opened 2019-07-02
Issues reported by cppcheck


ac000:
##### Bug report summary

Running the cppcheck.sh script under the current commit c56e086ba341 ("Easily disable alarms, by persisting the silencers configuration (#6360)) produces the following *issues*

[/home/andrew/programming/netdata/collectors/cgroups.plugin/cgroup-network.c:233]: (warning) Assignment of function parameter has no effect outside the function. Did you forget dereferencing it?

[/home/andrew/programming/netdata/tests/profile/benchmark-line-parsing.c:648]: (warning) Unnecessary comparison of static strings.

[/home/andrew/programming/netdata/web/server/web_client.c:214]: (portability) Non reentrant function 'getpwuid' called. For threadsafe applications it is recommended to use the reentrant replacement function 'getpwuid_r'.
[/home/andrew/programming/netdata/web/server/web_client.c:222]: (portability) Non reentrant function 'getpwnam' called. For threadsafe applications it is recommended to use the reentrant replacement function 'getpwnam_r'.
[/home/andrew/programming/netdata/web/server/web_client.c:245]: (portability) Non reentrant function 'getgrgid' called. For threadsafe applications it is recommended to use the reentrant replacement function 'getgrgid_r'.
[/home/andrew/programming/netdata/web/server/web_client.c:253]: (portability) Non reentrant function 'getgrnam' called. For threadsafe applications it is recommended to use the reentrant replacement function 'getgrnam_r'.
160/160 files checked 100% done

[/home/andrew/programming/netdata/tests/profile/benchmark-line-parsing.c:536] -> [/home/andrew/programming/netdata/tests/profile/benchmark-line-parsing.c:506]: (warning) Null pointer dereference: base
[*]: (information) Unmatched suppression: nullPointerRedundantCheck

##### OS / Environment

Fedora 30 (x86-64)

##### Netdata version (ouput of `netdata -V`)

v1.15.0-136-gc56e086b

##### Component Name

General

##### Steps To Reproduce

./cppcheck.sh

##### Expected behavior

Generally no problems reported

ac000:
So the last warning (Null pointer dereference) seems to be a false positive [0] with this telling trac ticket[1].

[0]: https://sourceforge.net/p/cppcheck/discussion/general/thread/8561473a/?limit=25
[1]: https://trac.cppcheck.net/ticket/7146

thiagoftsm:
Hi @ac000, firstly thanks by your feedback!

[...]

PR #6386, opened 2019-07-03
Cppcheck fixes

ac000:
This comprises a couple of commits

collector: cgroups: Fix a cppcheck warning
tests/profile: Remove somewhat redundant code

The first removes a superfluous line from a function. This should not be controversial.

The second commit also removes some code, that *I* believe is somewhat redundant. See the commit message for the details and also https://github.com/netdata/netdata/issues/6378

Yes, this one may be more controversial...

Cheers,
Andrew

### DIFF
--- a/collectors/cgroups.plugin/cgroup-network.c
+++ b/collectors/cgroups.plugin/cgroup-network.c
@@ -230,7 +230,6 @@ static struct ns {
 };
 
 int switch_namespace(const char *prefix, pid_t pid) {
-    if(!prefix) prefix = "";
 
 #ifdef HAVE_SETNS
 
--- a/tests/profile/benchmark-line-parsing.c
+++ b/tests/profile/benchmark-line-parsing.c
@@ -643,11 +643,6 @@ void main(void)
     total_active_file_hash = simple_hash("total_active_file");
     total_unevictable_hash = simple_hash("total_unevictable");
 
-    // cache functions
-    (void)simple_hash2("hello world");
-    (void)strcmp("1", "2");
-    (void)strtoull("123", NULL, 0);
-
   unsigned long i, c1 = 0, c2 = 0, c3 = 0, c4 = 0, c5 = 0, c6 = 0, c7;
   unsigned long max = 1000000;
 


ac000:
ok to test

tests/profile/benchmark-line-parsing.c
```diff
@@ -643,11 +643,6 @@ void main(void)
     total_active_file_hash = simple_hash("total_active_file");
     total_unevictable_hash = simple_hash("total_unevictable");
 
-    // cache functions
```
mfundul: It's possible that those lines made a difference in small platforms such as raspberry pi. @ktsaou should know why this was added, we should discuss about it before removing them.

collectors/cgroups.plugin/cgroup-network.c
```diff
@@ -230,7 +230,6 @@ static struct ns {
 };
 
 int switch_namespace(const char *prefix, pid_t pid) {
-    if(!prefix) prefix = "";
```
cakrit: The fix is wrong. The line should be moved inside the `ifdef`

mfundul: It's not wrong cause the same check is being done again inside the `ifdef` by `proc_pid_fd()`. It's redundant code.

tests/profile/benchmark-line-parsing.c
ac000: Sure. However at least the strcmp() and strtoull() calls are made in test1() which is actually called twice with the timing of the second one taken, so they would at least seem redundant in either case.

collectors/cgroups.plugin/cgroup-network.c
ac000: Right. Perhaps I wasn't clear enough in my commit message?

tests/profile/benchmark-line-parsing.c
mfundul: Yeah, this `test1()` was probably coded to run twice at a different date than the cache functions code, so it defeats the purpose in that case.

collectors/cgroups.plugin/cgroup-network.c
cakrit: Nope, just too hasty from my part.

tests/profile/benchmark-line-parsing.c
cakrit: @mfundul just drop a note to @ktsaou to ask him about this, the code does look very strange there.

ilyam8: at least it should be commented better

ktsaou: I think these are left overs.

The current code now tests `max` iterations. In older versions of it, it used `TSC` CPU commands to test a single iteration. When it was testing just one iteration, it was important the functions to be in the processor caches. Otherwise, loading the function would impact the result significantly.

So, I think we can now remove them.

PR #6387, opened 2019-07-03
Daemon fix double kills of collection threads on shutdown

### LINKED ISSUE #6190: charts.d kills process twice
##### Bug report summary
The following show up in the error.log every time netdata is stopped:
`systemctl stop netdata`
or
`systemctl restart netdata`
or
`reboot`

> 2019-06-01 07:48:55: netdata ERROR : PLUGINSD[charts.d] : read failed
2019-06-01 07:48:55: netdata ERROR : PLUGINSD[charts.d] : '/usr/libexec/netdata/plugins.d/charts.d.plugin' (pid 27374) disconnected after 50 successful data collections (ENDs).
2019-06-01 07:48:56: netdata INFO  : PLUGINSD : stopping plugin thread: plugin:charts.d
2019-06-01 07:48:56: netdata INFO  : PLUGINSD[charts.d] : data collection thread exiting
2019-06-01 07:48:56: netdata INFO  : PLUGINSD[charts.d] : killing child process pid 27374
2019-06-01 07:48:56: netdata ERROR : PLUGINSD[charts.d] : Request to kill pid 27374, but it is not running. (errno 3, No such process)
2019-06-01 07:48:56: netdata INFO  : PLUGINSD[charts.d] : thread with task id 27366 finished

This doesn't really affect netdata running but it does fill up the logs with errors that make it harder to troubleshoot what is really going wrong with a plugin.

It seems like when a shutdown command is given:
/collectors/plugins.d/plugins_d.c keeps trying to read and produces the "read failed" error.
/collectors/plugins.d/plugins_d.c sends the "disconnected after xx successful data collections (ENDs)" error.
This is produced whenever any plugin (charts.d, python.d, etc) exits so I'm guessing it is intended.
collectors/plugins.d/plugins_d.c then attempts to stop the process by calling "killpid" from /daemon/main.c
Somewhere the process has already been killed so /daemon/main.c sends the error "(errno 3, No such process)"

##### OS / Environment
Ubuntu 18.04 headless server

netdata v1.15.0-42-nightly

##### Component Name
/collectors/charts.d.plugin/charts.d.plugin.in

##### Steps To Reproduce
Disable all modules except apcupsd or sensors in charts.d.conf (sensors disabled in python.d.conf)
Note: I have tested this with only apcupsd or sensors modules enabled. Both individually produce the same errors on exit. So I'm assuming the errors do not have to do with the modules but with the charts.d plugin
`systemctl stop netdata.service`
`truncate -s 0 /var/log/netdata/error.log`
`systemctl start netdata.service`
No errors will show on start.
`systemctl stop netdata.service`
All the above errors show.

##### Expected behavior
Stop or restart without causing exit errors.

emmrk:
##### Summary
Disables collection threads killed with `SIGTERM` to stop the cleanup function from attempting to kill them again.

##### Component Name
daemon

##### Additional Information
Fixes #6190

### DIFF
--- a/collectors/plugins.d/plugins_d.c
+++ b/collectors/plugins.d/plugins_d.c
@@ -554,7 +554,7 @@ void *pluginsd_worker_thread(void *arg) {
             }
 
             if(likely(cd->serial_failures <= SERIAL_FAILURES_THRESHOLD)) {
-                infoerr("'%s' (pid %d) does not generate useful output but it reports success (exits with 0). %s.",
+                info("'%s' (pid %d) does not generate useful output but it reports success (exits with 0). %s.",
                     cd->fullfilename, cd->pid,
                     cd->enabled ?
                         "Waiting a bit before starting it again." :
--- a/collectors/plugins.d/plugins_d.c
+++ b/collectors/plugins.d/plugins_d.c
@@ -556,6 +556,7 @@ void *pluginsd_worker_thread(void *arg) {
                 // we have collected something
 
                 if(likely(cd->serial_failures <= 10)) {
+                    if (code == -1) cd->enabled = 0;
                     error("'%s' (pid %d) exited with error code %d, but has given useful output in the past (%zu times). %s", cd->fullfilename, cd->pid, code, cd->successful_collections, cd->enabled?"Waiting a bit before starting it again.":"Will not start it again - it is disabled.");
                     sleep((unsigned int) (cd->update_every * 10));
                 }
--- a/collectors/plugins.d/plugins_d.c
+++ b/collectors/plugins.d/plugins_d.c
@@ -522,6 +522,69 @@ static void pluginsd_worker_thread_cleanup(void *arg) {
 }
 
 #define SERIAL_FAILURES_THRESHOLD 10
+static void pluginsd_worker_thread_handle_success(struct plugind *cd) {
+    if (likely(cd->successful_collections)) {
+        sleep((unsigned int) cd->update_every);
+        return;
+    }
+
+    if(likely(cd->serial_failures <= SERIAL_FAILURES_THRESHOLD)) {
+        info("'%s' (pid %d) does not generate useful output but it reports success (exits with 0). %s.",
+            cd->fullfilename, cd->pid,
+            cd->enabled ?
+                "Waiting a bit before starting it again." :
+                "Will not start it again - it is now disabled.");
+        sleep((unsigned int) (cd->update_every * 10));
+        return;
+    }
+
+    if (cd->serial_failures > SERIAL_FAILURES_THRESHOLD) {
+        error("'%s' (pid %d) does not generate useful output, although it reports success (exits with 0)."
+            "We have tried to collect something %zu times - unsuccessfully. Disabling it.",
+            cd->fullfilename, cd->pid, cd->serial_failures);
+        cd->enabled = 0;
+        return;
+    }
+
+    return;
+}
+
+static void pluginsd_worker_thread_handle_error(struct plugind *cd, int code) {
+    if (code == -1) {
+        info("'%s' (pid %d) was killed with SIGTERM. Disabling it.", cd->fullfilename, cd->pid);
+        cd->enabled = 0;
+        return;
+    }
+
+    if (!cd->successful_collections) {
+        error("'%s' (pid %d) exited with error code %d and haven't collected any data. Disabling it.",
+            cd->fullfilename, cd->pid, code);
+        cd->enabled = 0;
+        return;
+    }
+
+    if (cd->serial_failures <= SERIAL_FAILURES_THRESHOLD) {
+        error("'%s' (pid %d) exited with error code %d, but has given useful output in the past (%zu times). %s",
+            cd->fullfilename, cd->pid, code, cd->successful_collections,
+            cd->enabled ?
+                "Waiting a bit before starting it again." :
+                "Will not start it again - it is disabled.");
+        sleep((unsigned int) (cd->update_every * 10));
+        return;
+    }
+
+    if (cd->serial_failures > SERIAL_FAILURES_THRESHOLD) {
+        error("'%s' (pid %d) exited with error code %d, but has given useful output in the past (%zu times)."
+            "We tried to restart it %zu times, but it failed to generate data. Disabling it.",
+            cd->fullfilename, cd->pid, code, cd->successful_collections, cd->serial_failures);
+        cd->enabled = 0;
+        return;
+    }
+
+    return;
+}
+#undef SERIAL_FAILURES_THRESHOLD
+
 void *pluginsd_worker_thread(void *arg) {
     netdata_thread_cleanup_push(pluginsd_worker_thread_cleanup, arg);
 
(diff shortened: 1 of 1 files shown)


collectors/plugins.d/plugins_d.c
```diff
@@ -544,53 +545,77 @@ void *pluginsd_worker_thread(void *arg) {
...
+        if (likely(code == 0)) {
+            // the plugin reports success
+
+            if (likely(cd->successful_collections)) {
+                sleep((unsigned int) cd->update_every);
+                goto exit_code_analyzed;
+            }
+
+            if(likely(cd->serial_failures <= SERIAL_FAILURES_THRESHOLD)) {
+                infoerr("'%s' (pid %d) does not generate useful output but it reports success (exits with 0). %s.",
```
cakrit: I haven't seen infoerr used anywhere in the code, didn't even know it was defined. Why did you choose to use that one instead of info?

emmrk: Well, I've seen it with the other logging-related macros and either got distracted by its "INFO" label, or that was just a runaway autocomplete. Will change to  `info()`? I don't think this particular message should be an `error()` - after all, it's just a retry.

```c
#define info(args...)    info_int(__FILE__, __FUNCTION__, __LINE__, ##args)
#define infoerr(args...) error_int("INFO", __FILE__, __FUNCTION__, __LINE__, ##args)
#define error(args...)   error_int("ERROR", __FILE__, __FUNCTION__, __LINE__, ##args)
#define fatal(args...)   fatal_int(__FILE__, __FUNCTION__, __LINE__, ##args)
```

emmrk: Changed in to `info()`

vlvkobal:
Why did you reverse success and failure cases? Why don't you use `else` instead of `if(code != 0)`?

vlvkobal:
What is the reason for the extensive usage of `goto exit_code_analyzed;`?

emmrk:
First of all, a disclaimer: this is what I, personally, think is better. If the project's style is agreed upon and should be maintained, then I'll discard this commit as non-essential (as it says in the commit message).

The reason is purely stylistic: this way the code reads more like English (as in, a natural language), and is therefore easier to understand. Instead of figuring out nested `if - else`'s that sometimes take their first branches on `true`s and sometimes on `false`s, I would argue that a more straightforward approach makes these various cases more comprehensible.

For me it reads, for example: "the error code is 0, and there are successful collections - everything is fine; if there aren't, and the collector hadn't failed too much, retry; if it had - disable it.". Now compare that with the previous version.

`goto`s are just a way to organize this without the nesting. 

I tried to rewrite this code for humans to read, but it still should be just as performant: the most often (if I understand the module's way of operation correctly) case of "everything is fine" is denoted as `likely` and positioned closer to the entry point; all the `goto`s and other `if`s (and the absence of an `else` that you are concerned about) are very straightforward for the compiler to optimize. I took a look at the assembly, it seems adequate.

So, to recap, I strongly believe this is just better to read, and there are no other side-effects. There also aren't any other reasons for this refactoring.

What's you opinion, @vlvkobal?

ilyam8:
@emmrk @vlvkobal wouldnt be better to split `pluginsd_worker_thread` into several funcs? And get rid of both netsted if/else and goto's?

ilyam8:
> There also aren't any other reasons for this refactoring.

i see no `refactoring` parh in the OP, maybe it should be there with motives. I agree that nested if/else is (always) hard to read/follow. 

If there is a willing to make it readable (which is very nice), then it would be nice to have it w/o goto's. My 2 cents.

mfundul:
I am in favour of using `goto` in C code when it makes things clearer, keeps code shorter, cuts down nesting as shown in [dup_mm()](https://lxr.missinglinkelectronics.com/linux/kernel/fork.c#L1309). In this example the coder saves a ton of duplicated cleanup code and nesting. `goto` instructions are typically used as an exception handling substitute in C.

In this PR I don't find the new code to be more readable than the old code. The new code is 63 lines long when it used to be 42 lines, and the following comments were removed for no good reason:
```
 // nothing collected - disable it

 // we have collected something

 // we have collected nothing so far
```
Adding the comments back would make the new code 69 lines versus 42 lines of the old code.

Moreover, I find this from the old code:
```
 if(code != 0) {
...
 else {
```
more readable than this from the new code:
```
 if (likely(code == 0)) {
...
 if(code != 0) {
```

Ideally I'd keep the best of the 2 approaches, something like this:
```
 if (likely(code == 0)) {
...
else {
```

ilyam8:
Trying to fix nested if/else is a good intention imo, thumbs up for it :+1:

emmrk:
Thank you all for the review!

I took most of your advice: success and error handling are now separate functions, no more `goto`s... but haven't returned the comments. Do we really want them anymore when the conditions are so simple?

Also, the build now fails because
```
The command "sudo apt-get install -y libcap2-bin zlib1g-dev uuid-dev fakeroot libipmimonitoring-dev libmnl-dev libnetfilter-acct-dev gnupg python-pip" failed and exited with 100 during .
```

ilyam8:
> Do we really want them anymore when the conditions are so simple?

i would say no. I see no reason to have them after refactoring.

collectors/plugins.d/plugins_d.c
```diff
@@ -544,47 +608,12 @@ void *pluginsd_worker_thread(void *arg) {
         // get the return code
         int code = mypclose(fp, cd->pid);
```
ilyam8: i suggest to remove comment and rename `code` to `ret_code`. Adding context via variable name >> using comment

collectors/plugins.d/plugins_d.c
```diff
@@ -521,6 +521,70 @@ static void pluginsd_worker_thread_cleanup(void *arg) {
...
+            cd->fullfilename, cd->pid, code, cd->successful_collections, cd->serial_failures);
+        cd->enabled = 0;
+        return;
+    }
+
+    return;
+}
+#undef SERIAL_FAILURES_THRESHOLD
+
 void *pluginsd_worker_thread(void *arg) {
```
ilyam8: Why the func returns `void *` ?

mfundul: It's due to it being a pthread function, see [pthread_create](http://man7.org/linux/man-pages/man3/pthread_create.3.html).

ilyam8: Ok, thanks!

collectors/plugins.d/plugins_d.c
ilyam8: A whatever name it should be. Gist is

> Adding context via variable name >> using comment

collectors/plugins.d/plugins_d.c
```diff
@@ -521,6 +521,70 @@ static void pluginsd_worker_thread_cleanup(void *arg) {
...
+        error("'%s' (pid %d) exited with error code %d and haven't collected any data. Disabling it.",
+            cd->fullfilename, cd->pid, code);
+        cd->enabled = 0;
+        return;
+    }
+
+    if (cd->serial_failures <= SERIAL_FAILURES_THRESHOLD) {
+        error("'%s' (pid %d) exited with error code %d, but has given useful output in the past (%zu times). %s",
+            cd->fullfilename, cd->pid, code, cd->successful_collections,
+            cd->enabled ?
```
ilyam8: Hm, i think the whole handle plugin exit/restart handling logic is a little bit lame.
For example why do we need to `sleep((unsigned int) (cd->update_every * 10)` if `!cd->enabled` ?

should we even pass `cd` to handle func if it is disabled after `pluginsd_process`? :thinking: 	

i suggest to dig into it and see if there is any logic problems. Could you do it @emmrk?

ilyam8: ^^ this should be another PR tho

collectors/plugins.d/plugins_d.c
emmrk: Right, done!

collectors/plugins.d/plugins_d.c
emmrk: Well yes, there's at least this one unnecessary `sleep`... I've made it into an issue (https://github.com/netdata/netdata/issues/6402) so that it's not lost (is that a good idea?). I'll probably look into it later

emmrk:
@cakrit @vlvkobal are you OK with this PR? Are there any more issues you'd like me to address here?

PR #6391, opened 2019-07-04
Center the chart on timeframe when an alarm is raised

### LINKED ISSUE #5810: Links to alarms in notifications should always center the chart on the timeframe
##### Bug report summary

Most of our alarm notifications contain a link that you can click to go directly to the chart at the time of the alarm. Unfortunately, this only works as long as the alarm is active. 

##### Expected behavior

Be able to click on older/cleared alarms.
Bonus feature: Make the alarm log clickable, so I can visit older timeframes from there as well.

thiagoftsm:
##### Summary
When an alarm happens, we were sending notification for our users, but the notifications were missing an important information that allows to centralize the Chart to demonstrate to our users.
##### Component Name
UI
##### Additional Information
Closes #5810


web/gui/goto-host-from-alarm.html
```diff
@@ -92,6 +93,9 @@
 }
 
 function netdataURL(url) {
+    if(urlOptions.alarm_when != 0) {
+        urlOptions.alarm_when -= 150;
```
cakrit: Why -150? What does that do?

thiagoftsm: In normal situation our charts has 5 minutes(300 seconds), so here I was doing a test where I was trying to centralize the document in the 300 second that I have in my screen, but of course this is not good, because case some user try to see an alarm that happens in the first second he won't have data centralized. 
When I wrote this part of the code, I did not have enough knowledge of our java script files, now that I got the necessary knowledge, I will remove this line for the next commit, but it was useful in the beginning for I understand the whole process.

thiagoftsm:
Hi @cakrit , I changed the necessary javascript classes for we have the centralization of the alarms. I also changed the html files for when we click in the alarm log table it opens the chart for that log.

For both cases I am using the same behavior for the user does not need learn different approaches for the same new feature. I executed the following steps to test it:

1 - I received an alarm either in email or Slack, after this I copied the link.
2 - I opened another tab in the google Chrome or Firefox and I paste the link there, after this I used enter and it brings the charts centralizing the alarm on it.

When I was testing the table, I did not open new tab, I executed the following steps:

1 - I opened the log and I clicked in different rows 
2 - After this it opens a new tab and it goes directly to  the chart centralizing the alarm.

For both tests you will see that I am freezing the new tab/window that is open, the person will always have the chart freeze in the specific range to analyze all the parameters available. 

I removed the WIP, but I cannot consider this closed yet, because I have few doubts:

1 - I am centralizing the data with 60s on each side, is this a good value? Will we allow the user to change it?
2 - I did not find in the documentation where I can describe the user options in the web GUI. Which document are we using for this?
3 - Is this exactly like you imagined before? Or am I missing something?

Finally, and and no less important, in the console we have a link that sends information to https://netdata.cloud/alarms/redirect?agentID=MY_ID&chart=system.ram&alarm_id=1562612089

that is completely different structure of the alarm links used when the IP address is an internal network address, so I think I need to talk with somebody in the cloud team to finish this. Who is the person?

Best regards!

web/gui/src/dashboard.js/main.js
```diff
@@ -3160,6 +3160,19 @@ let chartState = function (element) {
         this.data_url += "&gtime=" + this.gtime;
         this.data_url += "&options=" + this.chartURLOptions();
 
+        if(NETDATA.options.redirect_freeze) {
```
cakrit: I didn't go into the details of the code to understand what's supposed to happen, but it didn't work for me. I click on the slack notification and I don't get centered on the correct time, or get the freezing.

health/notifications/web/README.md
```diff
@@ -1,8 +1,23 @@
...
 The netdata dashboard shows HTML notifications, when it is open.
 
 Such web notifications look like this:
 ![image](https://cloud.githubusercontent.com/assets/2662304/18407279/82bac6a6-7714-11e6-847e-c2e84eeacbfb.png)
 
+## Alarms
+
+The netdata GUI has a link for the alarms raised in the system on top menu, when the user access it, (s)he will have the following options:
+
+- Active : The active alarms of the Netdata. This option does not include the alarms that were not raised.
```
mfundul: `The active alarms of the Netdata` you should remove the `the `.

thiagoftsm: Thanks, I corrected the grammar.

web/gui/goto-host-from-alarm.html
```diff
@@ -101,6 +105,8 @@
         + ';alarm_unique_id=' + urlOptions.alarm_unique_id.toString()
         + ';alarm_id=' + urlOptions.alarm_id.toString()
         + ';alarm_event_id=' + urlOptions.alarm_event_id.toString()
+        + ';after=' + alarm_time.toString()
+        + ';alarm_freeze=' + urlOptions.alarm_freeze.toString()
```
jacekkolasa: so, to summarize slack discussion, i think we don't need to have special case for `alarm_freeze`. Dashboard is handling `before` and `after` properties and this should be enough - the freeze problem was probably caused because you omitted `before` property. To have more clean code, i think it's best to calculate the -/+ 2 minutes directly in goto-host-from-alarm.html

thiagoftsm: I could not work with the before correctly, I remember that every time I tried to change it I did not have success, this was the one of the motives I created the alarm_freeze.

Thanks by your tips.

jacekkolasa:
@cakrit what happens if the alarm happened 30 seconds ago, do you still want to center that point? When we'll do that, the right half of the chart will be almost empty
Currently Agent Dashboard forbids going to before/after perdiods in the future, so it corrects the before/after params in the url, for example:
alarm at 12:00, now it's 12:00:05
you have redirect to after: 11:58:00, before: 12:02:00
dashboard makes a correction, to: after: 11:58:05, before: 12:02:05

is this acceptable? Or you want me/us to prevent correction of after/before in main.js (in case of alarm)?

cakrit:
No need to change the automatic correction, it makes sense not to show the future. It's impossible when panning too, so leave it as is. You'll eventually add markers on the charts to indicate the exact moment an alarm was raised/cleared/changed severity anyway, so it will eventually be very obvious when the change happened.

jacekkolasa:
@thiagoftsm i think it would be good if we don't remove the hash params, so user can pass the url to someone else and it will show similar result. To make this happen, i think it's good to just redirect user to new url with new `after`/`before` params, and reload the page. It will be easy to adapt it to react-router in the future. This change is simple enough so we don't have any issues with localStorage, etc. Can you check how it will work for you?

I'm proposing some changes. I was working from `master` branch, sorry, it was easier for me this way :) I managed to get proper results when:

i've added this in goto-host-from-alarm.html at the end of `netdataURL(url)` function
```
        + ';alarm_when=' + urlOptions.alarm_when.toString()`

git diff goto-host-from-alarm.html
diff --git a/web/gui/goto-host-from-alarm.html b/web/gui/goto-host-from-alarm.html
index ec53df08..a5a6910e 100644
--- a/web/gui/goto-host-from-alarm.html
+++ b/web/gui/goto-host-from-alarm.html
@@ -69,6 +69,7 @@ var urlOptions = {
     alarm_unique_id: 0,
     alarm_id: 0,
     alarm_event_id: 0,
+    alarm_when: 0,
     hasProperty: function(property) {
         return typeof this[property] !== 'undefined';
     }
@@ -101,6 +102,7 @@ function netdataURL(url) {
         + ';alarm_unique_id=' + urlOptions.alarm_unique_id.toString()
         + ';alarm_id=' + urlOptions.alarm_id.toString()
         + ';alarm_event_id=' + urlOptions.alarm_event_id.toString()
+        + ';alarm_when=' + urlOptions.alarm_when.toString()
         ;
 }



```

and did the following changes in main.js:
```
git diff --cached
diff --git a/web/gui/main.js b/web/gui/main.js
index 71af6b59..d70e74a0 100644
--- a/web/gui/main.js
+++ b/web/gui/main.js
@@ -71,6 +71,7 @@ var urlOptions = {
     alarm_unique_id: 0,
     alarm_id: 0,
     alarm_event_id: 0,
+    alarm_when: null,
 
     hasProperty: function (property) {
         // console.log('checking property ' + property + ' of type ' + typeof(this[property]));
@@ -139,7 +140,7 @@ var urlOptions = {
             }
         }
 
-        var numeric = ['after', 'before', 'highlight_after', 'highlight_before'];
+        var numeric = ['after', 'before', 'highlight_after', 'highlight_before', 'alarm_when'];
         len = numeric.length;
         while (len--) {
             if (typeof urlOptions[numeric[len]] === 'string') {
@@ -153,6 +154,24 @@ var urlOptions = {
             }
         }
 
+        if (urlOptions.alarm_when) {
+            // if alarm_when exists, create after/before params
+            // -/+ 2 minutes from the alarm, and reload the page
+            const alarmTime = new Date(urlOptions.alarm_when * 1000).valueOf();
+            const timeMarginMs = 2 * 60 * 1000; // 2 mins
+
+            const after = alarmTime - timeMarginMs;
+            const before = alarmTime + timeMarginMs;
+
+            const newHash = document.location.hash.replace(
+              /;alarm_when=[0-9]*/i,
+              ";after=" + after + ";before=" + before,
+            )
+
+            history.replaceState(null, '', newHash);
+            location.reload();
+        }
+
         if (urlOptions.server !== null && urlOptions.server !== '') {
             netdataServerStatic = document.location.origin.toString() + document.location.pathname.toString();
             netdataServer = urlOptions.server;
```

Please let me know what you think if it's clear for you

jacekkolasa:
also this little change for css would be nice (display pointer cursor on clickable rows)
```
#alarms_log_table tbody tr {
    cursor: pointer;
}
```

you can add it in main.css, for example on line 586

thiagoftsm:
Hi @jacekkolasa ,

I tested all the changes that you suggested and I bring it to the final code, because I saw your code is more clean and it has the advantage of do not create new variables, on the other hand as we already talked about, I lost completely the redirect from the cloud as it was reported on ( https://github.com/netdata/cloud-backend/issues/40 ). I remember that I always created the link using a variable change inside the alarm-notify.sh .

I will use this message to ask a doubt for you, for we centralize the active alarms using the new codes, do I only need to set a new link with `after` and `before` and reload the site when the alarm Modal is closed? I had a previous code in my mind that won't work with these changes you brought to us.

Thank you very much by the help with the Java script code.

thiagoftsm:
Hi @jacekkolasa ,

Let me explain the situation that we are right now and what we are missing in this PR.
We are working with 3 scenarios in this PR:

-  Netdata sends notification for our users for the channels that they configure, this notification has a link with the variable set `alarm_when`, this link can be a local link or a netdata.cloud link. For both the Netdata is redirecting fine and we have the charts centralized, thanks the fix made by cloud and the code that you gave us.
- Another possibility to centralize the alarms is going to `Alarms` and after this `Log` on the client front end, this is also working fine.
- Now we are missing only one place, please see the next image

err16

When I click in the first link in the option `Active`  it redirects without to freeze the chart, in the best case scenario, we would redirect and freeze the chart from this link, and case the user gives a reload everything would work automatic again. Another possibility is we click and open a new window with the alarm frozen, right now we do this for the other two options.

Please, any doubt, let me know.

cakrit:
Also by clicking on an alarm log entry:
Screenshot from 2019-09-16 15-09-57

thiagoftsm:
> Also by clicking on an alarm log entry:
This is already working @cakrit . I created a link to it an @jacekkolasa gave us the CSS that changes the mouse behavior when goes over it.

jacekkolasa:
@thiagoftsm i've pushed one change, if `alarm.status` is `"WARNING"` or `"CRITICAL"`, then the chart is centered around `alarm.last_status_change` value. Do you think that property is ok? Or maybe we should calculate lookup start time?

thiagoftsm:
> @thiagoftsm i've pushed one change, if `alarm.status` is `"WARNING"` or `"CRITICAL"`, then the chart is centered around `alarm.last_status_change` value. Do you think that property is ok? Or maybe we should calculate lookup start time?

I understood that the variable last_status_change is reporting the exact time that the alarm was changed to `WARNING` or `CRITICAL`, so it looks like OK initially for me, I will do a round of tests now to confirm.

thiagoftsm:
Hi @cakrit ,

I removed the [WIP] from the chart name, because I saw that the last changes made by @jacekkolasa fixed the latest problem with this PR, please, can you test in your side to confirm that everything is correct?

@jacekkolasa thank you very much by all contribution given here.

PR #6392, opened 2019-07-04
Make use of GCC's __attribute__((unused))

ac000:
This pull request comprises eight commits

This updated pull request obeys the coding style and converts tabs to spaces.

Updated again for coding style consistency with surrounding code.

configure.ac: Add support for GCC's __attribute__((unused))
collectors/statsd.plugin: Mark a function argument as __maybe_unused
collectors/apps.plugin: Mark a function argument as __maybe_unused
libnetdata/locks/locks: Mark function arguments as __maybe_unused
libnetdata/socket/security: Mark a function argument as __maybe_unused
collectors/cgroups.plugin: Mark a function argument as __maybe_unused
database/rrddim: Mark function arguments as __maybe_unused
database/rrdvar: Mark a function argument as __maybe_unused

I like compiling with -Wall and -Wextra to catch various classes of potential problems.

While -Wall doesn't actually introduce any compiler warnings, -Wextra (along with -Wall) produces quite a few of "warning: unused parameter ‘XXXX’ [-Wunused-parameter]"

Now we could just use -Wno-unused-parameter but this can catch legitimate bugs. So instead, we can make use of GCC's __attribute__((unused)), now, that's quite a mouthful, so we create aliases __always_unused and __maybe_unused for it instead. These are set in configure.ac and for non GCC compilers they are just set empty.

See the first commit message for more details.

The rest of the commits make use of the new defines.

I'd like to see us using -Wall & -Wextra by default, if this pull request is accepted I'll perhaps try for that next.

Updated again just to expand on the commit message introducing the new macros to note that they can be used on objects other than function parameters and to also note that clang has support for this attribute also.

Cheers,
Andrew

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -437,6 +437,14 @@ else
     AC_DEFINE_UNQUOTED([unlikely(x)], [(x)], [gcc branch optimization])
 fi
 
+if test "${GCC}" = "yes"; then
+    AC_DEFINE([__always_unused], [__attribute__((unused))], [gcc unused attribute])
+    AC_DEFINE([__maybe_unused], [__attribute__((unused))], [gcc unused attribute])
+else
+    AC_DEFINE([__always_unused], [], [dummy unused attribute])
+    AC_DEFINE([__maybe_unused], [], [dummy unused attribute])
+fi
+
 if test "${enable_pedantic}" = "yes"; then
     enable_strict="yes"
     CFLAGS="${CFLAGS} -pedantic -Wall -Wextra -Wno-long-long"
--- a/collectors/apps.plugin/apps_plugin.c
+++ b/collectors/apps.plugin/apps_plugin.c
@@ -2006,7 +2006,7 @@ static inline int debug_print_process_and_parents(struct pid_stat *p, usec_t tim
     return indent + 1;
 }
 
-static inline void debug_print_process_tree(struct pid_stat *p, char *msg) {
+static inline void debug_print_process_tree(struct pid_stat *p, char *msg __maybe_unused) {
     debug_log("%s: process %s (%d, %s) with parents:", msg, p->comm, p->pid, p->updated?"running":"exited");
     debug_print_process_and_parents(p, p->stat_collected_usec);
 }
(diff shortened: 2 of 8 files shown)


ac000:
Hmm, OK musl libc uses __unused itself in various places... let's rename __unused to __always_unused

vlvkobal:
Hm. I always use `sudo CFLAGS="-O1 -ggdb -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1" ./netdata-installer.sh --enable-plugin-nfacct --disable-go --disable-lto --dont-wait`. I don't see any warnings you are talking about. How did you get them?

mfundul:
If you remove `-DNETDATA_INTERNAL_CHECKS=1` you get a bunch of warnings due to arguments that are used only for internal checks.

configure.ac
```diff
@@ -437,6 +437,14 @@ else
     AC_DEFINE_UNQUOTED([unlikely(x)], [(x)], [gcc branch optimization])
 fi
 
+if test "${GCC}" = "yes"; then
+    AC_DEFINE([__always_unused], [__attribute__((unused))], [gcc unused attribute])
```
ilyam8: > Hmm, OK musl libc uses __unused itself in various places... let's rename __unused to __always_unused

If there is a collision problem i suggest to use `netdata` prefix, like `__netdata_used`.

`__always_unused` doesnt look goot to me, because `__unused` already means always. And if we have `__always_unused`,  then `__unused` means something else.

ac000: musl uses __unused variables for structure padding (others generally use __pad or somesuch), so you would never generally see or use __unused from it and they are certainly not the same thing here.

But if you're not keen on the naming, here are some alternatives

__a_unused / __m_unused
UNUSED / MAYBE_UNUSED

other suggestions welcome.

ac000: @mfundul I will convert the tabs to spaces (the blasphemy! ;), I need to sort out a vim config for netdata...)

ilyam8: > __a_unused / __m_unused

Well, i think if you can make something descriptive then make it. `a` and `m` is not descriptive at all. So `always/maybe` > `a/m`.

To be honest i dont have strong opinion on that, always/maybe is ok.

ac000:
@vlvkobal I certainly think we should be compiling with some of those flags by default.

ac000:
I believe all review comments/requests have been addressed.

PR #6397, opened 2019-07-04
Update apps_groups.conf for time group

mbarper:
For Ubuntu 16.04 encountered this:
* The output for the 'ps -e' and '/proc/PID/stat' is "systemd-timesyn".
* The output for the substring mode '/proc/PID/cmdline' is "systemd-timesyn" as well

In order to check this group correctly systemd-timesyncd service has to be searched as substring with value "systemd-timesyn*"

Update apps_groups.conf

### DIFF
--- a/collectors/apps.plugin/apps_groups.conf
+++ b/collectors/apps.plugin/apps_groups.conf
@@ -189,7 +189,7 @@ print: cups* lpd lpq
 # -----------------------------------------------------------------------------
 # time servers and clients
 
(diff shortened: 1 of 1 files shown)


ilyam8:
@mbarper please sigh the CLA

mbarper:
> @mbarper please sigh the CLA

Done!

vlvkobal:
Good catch, thank you @mbarper!

PR #6399, opened 2019-07-05
Ignore /dev and /run space/inode usage

### LINKED ISSUE #6361: Netdata not ignoring /dev and /run by default
##### Bug report summary

Netdata is sending me alerts about disk space and inodes usage from 

* /run/lock
* /run
* /dev

##### OS / Environment

Debian 9, x64

netdata v1.15.0-129-nightly

##### Component Name

diskspace.plugin

##### Steps To Reproduce

This happened on a fresh Debian 9 (launched from AWS Debian's official AMIs) and with netdata installed from the kickstart script.

##### Expected behavior

Netdata's diskspace plugin should ignore /dev and /run.

vlvkobal:
##### Summary
There were useless alerts for `/dev` and `/run` space/inode usage on a fresh Debian 9 installation. We are disabling these checks by default.

Fixes #6361

##### Component Name
health

### DIFF
--- a/health/health.d/disks.conf
+++ b/health/health.d/disks.conf
@@ -13,7 +13,7 @@ template: disk_space_usage
       on: disk.space
       os: linux freebsd
    hosts: *
-families: *
+families: !/dev !/dev/* !/run !/run/* *
     calc: $used * 100 / ($avail + $used)
    units: %
    every: 1m
@@ -27,7 +27,7 @@ template: disk_inode_usage
       on: disk.inodes
       os: linux freebsd
    hosts: *
-families: *
+families: !/dev !/dev/* !/run !/run/* *
     calc: $used * 100 / ($avail + $used)
    units: %
    every: 1m


thiagoftsm:
I am letting it running during few hours to confirm that everything is correct after this fix.

ilyam8:
@vlvkobal 	have you tested it?

hey @alexandernst if you have time could you test the PR?

thiagoftsm:
I ran the following test here:

1 - I verified that in my debian VM I had 488 MB reserved to it, but it was using 0MB.
2 - Due the fact we wanna test no messages for disk space, I had to run the following steps:
# dd if=/dev/zero of=test count=476000 bs=1024
# cp test /dev

This last step filled 96% of the /dev without raise an alarm. After this I decided to move to the extreme.
3 - I ran the following commands:
# dd if=/dev/zero of=test count=500000 bs=1024
# cp test /dev

These last commands try to move a file bigger than /dev and I received a alarm for "out of disk space time". 

Considering these 2 tests I think this PR is doing what is hoped from. The udev is small and it won't use a lot of space when correctly set. Before to merge this, we need to check with @alexandernst whether he is not receiving other alarms related with disk space. 

I have to do two questions for @alexandernst :

1 - Are these tests that I realized compared with your real world problem?
2 - And case it is possible, please, can you say us how much free space you have allocated for your /dev?

Depending of the answers I will approve this, because this specific PR can hidden a message that can give us an alarm for a hidden worm working in the computer, for example.

Finally, it worked as it is hoped.

Ferroin:
@thiagoftsm While this is silencing an alert that could, in theory, be tripped by malware, it takes some pretty creative behavior to turn a full `/dev` or `/run` into an actual issue once the system has finished startup.  Both rarely, if ever, need new files added after startup, and the small percentage of software that would actually be impacted by them being full breaks pretty noisily even without the alarm from Netdata.

alexandernst:
@thiagoftsm 

Answering to your questions in order:

1. My real work problem is just a completely new Debian AMI launched in a t3.nano
2. Here is a df:

```
root@ip-172-17-20-168:/home/admin# df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            225M     0  225M   0% /dev
tmpfs            48M  8.6M   39M  19% /run
/dev/nvme0n1p1  7.9G  5.1G  2.4G  69% /
tmpfs           236M   64K  236M   1% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           236M     0  236M   0% /sys/fs/cgroup
tmpfs            48M     0   48M   0% /run/user/1000
```

thiagoftsm:
Hi @alexandernst and @Ferroin ,

Thank you by the answers and explanations given.
At least in my tests everything worked as it hoped, so I am approving the PR and waiting for other tests from other people.

PR #6400, opened 2019-07-06
Snapshot uniqueId fix

### LINKED ISSUE #6384: Cannot see charts in an imported snapshot
##### Bug report summary
If a Netdata snapshot is loaded while the `Which dimensions to show?` dashboard option is set to `All`, an error is displayed instead of a chart:
```
system.cpu: cannot get data from snapshot for key: "system.cpu,dygraph,null,ms%7Cflip%7Cjsonwrap"
```

The error in the console:
```
system.cpu (gauge 8380238e-6656-b481-cb4c-2a1735ba4574): snapshot does not include data for key "system.cpu,gauge,null,absolute%7Cjsonwrap"
```

The corresponding key in the snapshot file: `system.cpu,dygraph,iowait,ms%7Cflip%7Cjsonwrap%7Cnonzero`

##### OS / Environment
Arch Linux

v1.15.0-140-gbbb33ab4

##### Component Name
web gui

##### Steps To Reproduce
Export/save a snapshot, set the `Which dimensions to show?` dashboard option to `All`, import/load the saved snapshot.

##### Expected behavior
Charts from an imported snapshot should be displayed irrespective of dashboard options set.

jacekkolasa:
Fixes #6384 

##### Summary
When generating uniqueId for the snapshot, `NETDATA.options.current.eliminate_zero_dimensions` was taken into account and in some scenarios the keys were not matching (for example when exporting the snapshot `eliminate_zero_dimensions` were always `true`). I've removed the presence of `nonzero` key, so there's no conflict now.

Also, from what I understand `All` / `Non Zero` option in "Which dimensions to show?" option should not result in different representation in charts.

### DIFF
--- a/web/gui/dashboard.js
+++ b/web/gui/dashboard.js
@@ -8118,10 +8118,10 @@ let chartState = function (element) {
     };
 
     this.chartDataUniqueID = function () {
-        return this.id + ',' + this.library_name + ',' + this.dimensions + ',' + this.chartURLOptions();
+        return this.id + ',' + this.library_name + ',' + this.dimensions + ',' + this.chartURLOptions(true);
     };
 
-    this.chartURLOptions = function () {
+    this.chartURLOptions = function (isForUniqueId) {
         let ret = '';
 
         if (this.override_options !== null) {
@@ -8136,7 +8136,7 @@ let chartState = function (element) {
 
         ret += '%7C' + 'jsonwrap';
 
-        if (NETDATA.options.current.eliminate_zero_dimensions) {
+        if (!isForUniqueId && NETDATA.options.current.eliminate_zero_dimensions) {
             ret += '%7C' + 'nonzero';
         }
 
--- a/web/gui/dashboard.js
+++ b/web/gui/dashboard.js
@@ -8136,7 +8136,9 @@ let chartState = function (element) {
 
         ret += '%7C' + 'jsonwrap';
 
-        if (!isForUniqueId && NETDATA.options.current.eliminate_zero_dimensions) {
+        // always add `nonzero` when it's used to create a chartDataUniqueID
+        // we cannot just remove `nonzero` because of backwards compatibility with old snapshots
+        if (isForUniqueId || NETDATA.options.current.eliminate_zero_dimensions) {
             ret += '%7C' + 'nonzero';
         }
 
--- a/web/gui/src/dashboard.js/main.js
+++ b/web/gui/src/dashboard.js/main.js
@@ -3071,10 +3071,10 @@ let chartState = function (element) {
     };
 
     this.chartDataUniqueID = function () {
-        return this.id + ',' + this.library_name + ',' + this.dimensions + ',' + this.chartURLOptions();
+        return this.id + ',' + this.library_name + ',' + this.dimensions + ',' + this.chartURLOptions(true);
     };
 
-    this.chartURLOptions = function () {
+    this.chartURLOptions = function (isForUniqueId) {
         let ret = '';
 
         if (this.override_options !== null) {
@@ -3089,7 +3089,7 @@ let chartState = function (element) {
 
         ret += '%7C' + 'jsonwrap';
 
-        if (NETDATA.options.current.eliminate_zero_dimensions) {
+        if (!isForUniqueId && NETDATA.options.current.eliminate_zero_dimensions) {
             ret += '%7C' + 'nonzero';
         }
 
--- a/web/gui/src/dashboard.js/main.js
+++ b/web/gui/src/dashboard.js/main.js
@@ -3089,7 +3089,9 @@ let chartState = function (element) {
 
         ret += '%7C' + 'jsonwrap';
 
-        if (!isForUniqueId && NETDATA.options.current.eliminate_zero_dimensions) {
+        // always add `nonzero` when it's used to create a chartDataUniqueID
+        // we cannot just remove `nonzero` because of backwards compatibility with old snapshots
+        if (isForUniqueId || NETDATA.options.current.eliminate_zero_dimensions) {
             ret += '%7C' + 'nonzero';
         }
 


jacekkolasa:
@cakrit we are calling that method in `chartURL()` - to optionally add nonzero param to /api/v1/data call [here](https://github.com/jacekkolasa/netdata/blob/194bb26d1774c600448375f4ae66428108ca49e6/web/gui/src/dashboard.js/main.js#L3161), but I haven't checked yet if `nonzero` param is used by the BE, I thought that it's used to shrink the output payload

thiagoftsm:
Hi @jacekkolasa, I am one of the Netdata developers, you said something about the web server, so I decided to come here to help case it is necessary. I will give you two information that can help your work:

1 - The option 'nonzero' is used in the backend to set a flag to search data inside our internal database.
2 - Another information to help you, the variable 'eliminate_zero_dimensions' will always be 'true' by default when the user opens the GUI of Netdata.

Best Regards!

vlvkobal:
image

vlvkobal:
The PR doesn't work for me. @jacekkolasa, have you tested it?

thiagoftsm:
Hi @vlvkobal , probably the "&&" inside the chartURLOptions is avoiding to send the 'nonzero'. 
@jacekkolasa  I am working with the java script files in the last days, and I had to begin using the "private window" on firefox to avoid cache, please, check whether the cache is not creating wrong impression of your results.

jacekkolasa:
@thiagoftsm thank you for the tips! I'll explain the bug in more details: Normally `chartURLOptions()` is used to help generating an url, but it was also used to encode keys for the snapshot. The problem is that when exporting the snapshot, some default settings are used for `eliminate_zero_dimensions` [at this moment](https://github.com/jacekkolasa/netdata/blob/dd73f3e0cd0d9a1797f0c82de0fae5ccf97e9464/web/gui/main.js#L3661) (and perhaps to other global variables as well), and as you've pointed out, that flag is always `true` initially, so all keys have `nonzero` flag set up and that doesn't work when having `All` option turned on.

@vlvkobal thanks for checking that PR, thanks to you i've realised that i haven't made the fix backwards compatible (old snapshots will not work). Can you please let me know if you've used the old snapshot, or you generated it with my changes?

Anyway, i'll post a fix for backwards compatibility

jacekkolasa:
@cakrit @thiagoftsm @vlvkobal i've pushed the fix, what do you think? I was trying different solutions but this seems to be the most concise, even though it looks a little hacky :/

thiagoftsm:
Hi @jacekkolasa , our GUI code is really not simple, before last Thursday I did not had the necessity to touch it, but since them I am working with it every day not only to do what I need, but to understand it.

Before I give you a final opinion, I have to do a question to understand better what you are doing, what is the issue number that created this work for you? I would like to see it first, to confirm that everything was reached with your commit. I am writing this due a personal experience, but I won't let you without a specific answer, when I saw that "&&" in your first commit, I had in my mind that this would break something, this is the motive I explained you what was happening in the back end with the "nonzero". I could not test your first commit, but as you could see with @vlvkobal tests, what I though really happened.  The netdata GUI in normally accessed by browser, but it receives information from different sources and it also send requests to different source, so we need to take care a lot when we use conditions like that "&&", please, read these last words as a tip.  Now about the last change that you, it did looks like very good, the  or("||" ) will avoid the error given by @vlvkobal and possible others, because now the function chartURLOptions is delivering the results when "eliminate_zero_dimensions" is also false.

More details I only can give when I see what was asked to generate this PR, please, add it in the first description that you gave.

Best regards!

jacekkolasa:
@thiagoftsm thanks for the throughout feedback. I agree that the GUI code is quite complicated and can sometimes produce unexpected results, due to its imperative nature. The issue number is #6384. Please ask any questions you have, if you want i can write bigger description of the problem and about the research i've done.

In short, what my solution does now is:
- produce always 'nonzero' suffix in id/url (a key for each graph) when creating a snapshot - this is working as it was before (it was set [here](https://github.com/jacekkolasa/netdata/blob/dd73f3e0cd0d9a1797f0c82de0fae5ccf97e9464/web/gui/main.js#L3591))
- produce always 'nonzero' suffix in id/url when reading a snapshot - before my change it was adding `nonzero` suffix only when the setting was currently selected by the user - and it was a source of this bug
- produce always 'nonzero' suffix when making api call for the data (based on user selection - from global NETDATA object), like it was before my change

One doubt which i have is - when user is in 'All' mode and is reading a snapshot created in 'Nonzero' mode, should we switch the mode at that time? I assumed that we shouldn't.

Thanks again for checking that!

thiagoftsm:
Hi @jacekkolasa, now I understood 100% what was the problem and what you were trying to fix, it is not necessary to write more nothing for us, but next time remember to associate the issue number with your PR, because this helps us to create the scene where the problem is happening to check your code.

@cakrit in my vision everything is OK now,  let us wait @vlvkobal to confirm this with his tests. Case everything is fine, congratulation @jacekkolasa!

@cakrit, please, the GUI talks directly with the web server that I am one of the responsible, can you or me to associate my user to these files to check when there is a PR that touches them? After to expend some time with the JS files, I think I can help any person that will try to communicate with the server and case necessary I can touch the JS codes too.

thiagoftsm:
Hi @jacekkolasa, I received one information from our coworkers and I am giving it to you to avoid problems in your side too. According to him we cannot touch the file dashboard.js, because it is created during the compilation time, so we only need to change the files present  web/gui/src/dashboard.js/, case we change it directly it can create compilation errors.

Best Regards!

jacekkolasa:
@thiagoftsm thanks, i will remember that. FYI dashboard.js file which i've pushed to this PR is auto-generated. Please let me know if you think it should be removed from the PR (I assumed we have to commit it because it was already up to date in git history)

thiagoftsm:
Hi @jacekkolasa,  considering that LGTM and Travis did not raise any alarm, you do not need to change the PR, no less important @vlvkobal already approved, so everything is working properly! Now that is official that everything is working, congratulations again!

PR #6414, opened 2019-07-09
SSL vs. TLS consistency and clarification in documentation

joelhans:
##### Summary

With TLS landing in the stable v1.16.0 release I wanted to make sure that we're being clear and as consistent as possible about the fact that we're using TLS 1.2 as the encryption protocol, not any of the obsolete SSL protocols.

The fact that "SSL" is colloquially used to refer to any encrypted web connection—it's kind of like "Kleenex" being used for any brand of tissue—complicates this a bit. In certain places I try to ease the transition with `TLS/SSL`, but in my opinion we should be consistent in using `TLS` whenever possible, as that is technically correct.

##### Component Name

docs

##### Additional Information

I've also included some grammatical/stylistic fixes to the areas relevant to TLS.

In an ideal world, any references in configuration files are changed from `ssl` to `tls`, but that's probably a huge project without any real value for the user. For example:

```
[stream]
    destination = host:port:SSL #old
    destination = host:port:TLS #new
```

### DIFF
--- a/docs/Running-behind-nginx.md
+++ b/docs/Running-behind-nginx.md
@@ -119,10 +119,9 @@ Using the above, you access Netdata on the backend servers, like this:
 
 ### Using TLS communication
 
-In case the Netdata web server has been [configured to use TLS](../web/server/#enabling-tls-support),
-you must also encrypt the communication between Nginx and Netdata.
+In case Netdata's web server has been [configured to use TLS](../web/server/#enabling-tls-support), you must also encrypt the communication between Nginx and Netdata.
 
-To enable encryption, first [enable SSL on nginx](http://nginx.org/en/docs/http/configuring_https_servers.html) and then put the following in the location section of the Nginx configuration:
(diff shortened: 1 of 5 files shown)


docs/Running-behind-nginx.md
```diff
@@ -133,15 +132,15 @@ If nginx is not configured as described here, you will probably receive the erro
...
+Create an authentication file to enable basic authentication via Nginx. Do not use authentication without having first [enabled TLS](#using-tls-communication)!
+
+If you don't have an authentication file, you can use the following command:
 
 ```
 printf "yourusername:$(openssl passwd -apr1)" > /etc/nginx/passwords
 ```
 
-And enable the authentication inside your server directive:
+And thenenable the authentication inside your server directive:
```
vlvkobal: A missed whitespace `then enable`

joelhans: Got it. Thanks!

cakrit:
I'll approve when @vlvkobal and @thiagoftsm approve.

joelhans:
@thiagoftsm: Fixed that sentence you requested!

PR #6419, opened 2019-07-10
Add global configuration option for zero metrics

### LINKED ISSUE #6315: Global option for enabling charts with zero metrics
##### Feature idea summary
By default, Netdata will enable monitoring metrics for disks, memory, and network only when they are not zero. These charts can be enabled permanently using the corresponding configuration options but we already have over 200 such options. It is desirable to have the possibility to change the default behaviour globally.

Preliminary discussion on the matter: #5845

##### Expected behavior
It is possible to enable all charts with zero values using a global configuration option.

vlvkobal:
##### Summary
If all metrics are zero in a chart, it is not displayed by default. A global configuration option will be added to allow users to see charts with all zero metrics if desired. 

Closes #6315

##### Component Name
proc plugin
diskspace plugin
cgroups plugin
freebsd plugin
macos plugin
tc plugin

### DIFF
--- a/collectors/proc.plugin/proc_meminfo.c
+++ b/collectors/proc.plugin/proc_meminfo.c
@@ -219,7 +219,8 @@ int do_proc_meminfo(int update_every, usec_t dt) {
 
     unsigned long long SwapUsed = SwapTotal - SwapFree;
 
-    if(do_swap == CONFIG_BOOLEAN_YES || SwapTotal || SwapUsed || SwapFree) {
+    if(do_swap == CONFIG_BOOLEAN_YES || (do_swap == CONFIG_BOOLEAN_AUTO &&
+                                         (SwapTotal || SwapUsed || SwapFree || netdata_zero_metrics_enabled == CONFIG_BOOLEAN_YES))) {
         do_swap = CONFIG_BOOLEAN_YES;
 
         static RRDSET *st_system_swap = NULL;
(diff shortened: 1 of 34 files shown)


collectors/proc.plugin/proc_meminfo.c
```diff
@@ -219,7 +219,8 @@ int do_proc_meminfo(int update_every, usec_t dt) {
 
     unsigned long long SwapUsed = SwapTotal - SwapFree;
 
-    if(do_swap == CONFIG_BOOLEAN_YES || SwapTotal || SwapUsed || SwapFree) {
+    if(do_swap == CONFIG_BOOLEAN_YES || (do_swap == CONFIG_BOOLEAN_AUTO &&
+                                         (SwapTotal || SwapUsed || SwapFree || netdata_zero_metrics_enabled == CONFIG_BOOLEAN_YES))) {
         do_swap = CONFIG_BOOLEAN_YES;
```
ilyam8: Could you format it better? It looks ugly and it is hard to read, it is even longer then 120 characters. Do you like it yourself?

Screenshot_20190710_114014

ilyam8: This is true for all conditions. I believe there is always a possibility to make them readable.

vlvkobal: A graceful solution needs a great deal of refactoring. So I won't do it for now.

ilyam8:
ok, i tested the PR installing it from source. Works as expected.

Issue #6412, opened 2019-07-09
Double free or corruption (again)


cakrit:
```
 latest pull from git 

admin-testakis-01 [netdata] # gcc --version
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-36)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

admin-testakis-01 [netdata] # /usr/sbin/netdata -P /var/run/netdata/netdata.pid -D
*** Error in `/usr/sbin/netdata': double free or corruption (!prev): 0x0000000001772430 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81489)[0x7f8cede7d489]
/usr/sbin/netdata[0x40fb6a]
/usr/sbin/netdata[0x456aa4]
/usr/sbin/netdata[0x4ae09a]
/usr/sbin/netdata[0x425a98]
/lib64/libpthread.so.0(+0x7dd5)[0x7f8cee1d0dd5]
/lib64/libc.so.6(clone+0x6d)[0x7f8cedef9ead]
======= Memory map: ========
00400000-004e0000 r-xp 00000000 fd:00 25560466                           /usr/sbin/netdata
006df000-006e0000 r--p 000df000 fd:00 25560466                           /usr/sbin/netdata
006e0000-006f3000 rw-p 000e0000 fd:00 25560466                           /usr/sbin/netdata
006f3000-006fb000 rw-p 00000000 00:00 0
0142f000-0178a000 rw-p 00000000 00:00 0                                  [heap]
7f8ce1577000-7f8ce1579000 rw-p 00000000 00:00 0
7f8ce1579000-7f8ce158e000 r-xp 00000000 fd:00 25204759                   /usr/lib64/libgcc_s-4.8.5-20150702.so.1
7f8ce158e000-7f8ce178d000 ---p 00015000 fd:00 25204759                   /usr/lib64/libgcc_s-4.8.5-20150702.so.1
7f8ce178d000-7f8ce178e000 r--p 00014000 fd:00 25204759                   /usr/lib64/libgcc_s-4.8.5-20150702.so.1
7f8ce178e000-7f8ce178f000 rw-p 00015000 fd:00 25204759                   /usr/lib64/libgcc_s-4.8.5-20150702.so.1
[...]

PR #6276, opened 2019-06-13
ZRAM info collector module (proc.plugin)

RaZeR-RBI:
##### Summary
Implemented ZRAM collector module (#2578) in ```proc.plugin```.

The module gets the ZRAM device list by reading /proc/devices, obtaining an device ID from it, then enumerating the devices in /dev filtering them by corresponding major device number it got from previous step.

It takes the data from ```/sys/block/{name}/mm_stat```.

##### Component Name
netdata/collectors

##### Additional Information
Here is a screenshot (a bit zoomed out, but you should get the idea):
[Click me!](https://i.imgur.com/HCzrwWS.png)

My C is a little bit rusty, I hope I haven't missed something important. Feedback and criticism is very welcome :wink:

Fixes: #2578

PR #6424, opened 2019-07-10
(re-open) ZRAM info collector module (proc.plugin)

RaZeR-RBI:
Re-opening #6276 (initial PR with ZRAM module that was merged) in response to #6412 (double free and corruption).

I haven't reproduced the bug on my machines yet (my guess is that it may be tied to Linux distribution), but I've added a pair of additional checks that (possibly) could crash the app.

Any info about the environment where the bug was encountered is very appreciated. Sorry for inconvenience :flushed:


collectors/proc.plugin/sys_block_zram.c
```diff
@@ -0,0 +1,288 @@
...
+    procfile *ff = NULL;
+    ZRAM_DEVICE device;
+
+    if (unlikely(!dir))
+        return 0;
+    while ((de = readdir(dir)))
+    {
+        snprintfz(filename, FILENAME_MAX, "/dev/%s", de->d_name);
+        if (unlikely(stat(filename, &st) != 0))
+        {
```
ilyam8: Hey, it would be nice to not add another code style

https://docs.netdata.cloud/contributing/#code-style

https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces

vlvkobal:
@RaZeR-RBI I see the WIP label is still attached. Are you going to make changes to the PR?

RaZeR-RBI:
> 
> 
> @RaZeR-RBI I see the WIP label is still attached. Are you going to make changes to the PR?

I've attached the WIP label in order to prevent merging due to reasons explained in the first post. If everything is ok (the bug is fixed) or if the label is misleading we can remove it :wink:

PR #6425, opened 2019-07-10
Styling improvements to documentation

joelhans:
##### Summary

I've been trying to improve the look and feel of the docs here and there in all my different WIP branches and decided it was finally time to put them all together.

- Move away from the `highlight` plugin and to the `pymdownx.highlight` plugin, as the latter is better supported and doesn't do weird things with inline code blocks.
- Add some padding to the bottom of each page.
- Change the colors to grey and green to better match our new site (it's not perfect but it's way closer than the blue).
- Add a `docs/README.md` file to direct people toward the docs site.
- A few CSS tweaks to make things look nicer.

##### Component Name

docs


joelhans:
This should be good to go with Costa's approval today.

PR #6426, opened 2019-07-10
Utf8 Badge Fix And URL Parser International Support (initial)

### LINKED ISSUE #3117: badge do not suppport other language
I used the label option with an non-english character, the badge did not appear the character.

underhood:
##### Summary
Fixes  #3117

Additionally it adds support for UTF-8 in URL parser (as it should).
Label sizes now are updated by browser with JavaScript (although guess is still calculated by verdana11_widths with minor improvements)

##### Component Name
API/Badges, LibNetData/URL

##### Additional Information
It was found that not only verdana11_widths need to be updated but the url parser replaces international characters with spaces (one space per each byte of multibyte character).
Therefore I update both to support international chars.

##### Screenshots Before/After
Before "v1.16.0":
OLDTAG_v1_16_0

After Firefox:
Latest_Firefox

After Chromium:
Latest_Chromium


underhood:
### Better compare
#### Old
old
#### New
new

libnetdata/url/url.c
```diff
@@ -52,17 +52,73 @@ char *url_decode(char *str) {
...
+        to_read--;
+    }
+
+    return byte_length;
+}
+
 char *url_decode_r(char *to, char *url, size_t size) {
     char *s = url,           // source
          *d = to,            // destination
          *e = &to[size - 1]; // destination end
```
ilyam8: Why not name them `src`, `dst` (or even `source`, `destination`), etc. In that case you dont need those comments.

libnetdata/url/url.c
```diff
@@ -52,17 +52,73 @@ char *url_decode(char *str) {
...
+    //10XX XXXX is not valid character -> check length == 1
+    if(length > 4 || length == 1)
+        return -1;
+    //0XXX XXXX means 1 byte
+    return (length) ? length : 1;
+}
+
+//decode % encoded UTF-8 characters and copy them to *d
+//return count of bytes written to *d
+char url_decode_multibyte_utf8(char *s, char *d, char* d_end) {
```
ilyam8: > return count of bytes written to *d

> return byte_length;

it looks like it returns `byte_length`

`s` => `src`, `d` => `dst`, return `bytes_written` - after it you don't need comments.

I think you don't need comments just to explain naiming.

Maybe i am missing something and this is not `c` way. What do you think?

libnetdata/url/url.c
underhood: Hi,

They were called s, d and e (END) in original code. So I kind of tried to imitate original naming.

ilyam8: Opsie. It is from original code, nvm then :see_no_evil: (in my opinion it makes no sense to name vars poorly and use comments to *decode* names).

underhood: Yep agree with you. But as I mentioned I just kind of looked at code around (new to netdata) and tried to keep same 'style' although I will check what are actual official coding style rules (if any) when I get few free minutes tomorrow :)

libnetdata/url/url.c
ilyam8: ok, since this reflects surrounded code i have no question. But i strongly believe this is not part of the style we accepted.

thiagoftsm:
Hi @underhood ,

Your PR is working as expected, I used Latin alphabet and Cyrillic alphabet and for both I could get the result correctly, so congratulation.

Now let me explain a little more about the format for you adjust it and we approve the PR. For example, in the declaration of the function  url_decode_multibyte_utf8, you are putting the pointer signal in two different positions:

 url_decode_multibyte_utf8(char *s, char *d, char* d_end) 

At least in my vision, the "char* d_end" is something that I hope to see in Java code, in C code we generally use the other, so you are having two patterns in the same function declaration.

Another thing that I saw, in the function verdana_width, in the line 152 you have a variable "double t" that will never be used, because you created another in the line 162. It is completely possible that with the compilers that we have nowadays, this missing variable will be ignored and it won't try to move the value for the specific registers, but case we compile this file applying "-Werror -Wfatal-errors -Wall" it won't compile, because the warning will become an error.

Please, do these two adjusts and I will approve it!

web/api/badges/web_buffer_svg.c
```diff
@@ -139,135 +139,7 @@ static double verdana11_widths[256] = {
...
-    [247] = 0.0,
-    [248] = 0.0,
-    [249] = 0.0,
-    [250] = 0.0,
-    [251] = 0.0,
-    [252] = 0.0,
-    [253] = 0.0,
-    [254] = 0.0,
-    [255] = 0.0
+    [127] = 0.0
```
cakrit: It may be irrelevant to the change here, but there is a test that behaves differently now. I passed label=%82 (extended ASCII). I used to get spaces/empty chars, no I get 82. not sure why this happens though, please investigate.

web/api/badges/web_buffer_svg.c
```diff
@@ -862,19 +745,45 @@ void buffer_svg(BUFFER *wb, const char *label, calculated_number value, const ch
                 "<stop offset=\"1\" stop-opacity=\".1\"/>"
             "</linearGradient>"
             "<mask id=\"round\">"
-                "<rect width=\"%0.2f\" height=\"%0.2f\" rx=\"%0.2f\" fill=\"#fff\"/>"
+                "<rect class=\"bdge-ttl-width\" width=\"%0.2f\" height=\"%0.2f\" rx=\"%0.2f\" fill=\"#fff\"/>"
```
cakrit: Very cool that you added the classes, but what's the point without a css? Can I even use them? Maybe there's something I don't understand here...

web/api/badges/web_buffer_svg.c
```diff
@@ -277,15 +149,26 @@ static inline double verdana11_width(char *s) {
...
-        if(t == 0.0)
-            s++;
+        double t;
+        if(*s & 0x80) {
+            //if UTF8 multibyte char found
+            //count it as one character and guess it's width equal 1em
+            //as lable width will be updated with JavaScript this is not so important
+            while(*s & 0x80)
+                s++;
+            w += verdana11_widths['m'] + VERDANA_KERNING;
```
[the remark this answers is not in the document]
underhood: Idea is you can now have UTF8 chars which can be up to 4 Bytes long - therefore your table will be so so huge if you want to include all characters in UTF8 space. So now we just guess size and for characters we do not understand just put 1em. (Although you are right 1m should be 12pt not size of 'm' plus kerning. I will fix that asap.
Size of label is now updated by Browser as I think it should be. You at the C code size do not really know the size very reliably currently. Let browser do that stuff which it is good at - figuring out sizes, spacing, layout and rendering pixels :).

web/api/badges/web_buffer_svg.c
```diff
@@ -277,15 +149,26 @@ static inline double verdana11_width(char *s) {
...
 
     while(*s) {
-        double t = verdana11_widths[(unsigned char)*s];
-        if(t == 0.0)
-            s++;
+        double t;
+        if(*s & 0x80) {
+            //if UTF8 multibyte char found
+            //count it as one character and guess it's width equal 1em
+            //as lable width will be updated with JavaScript this is not so important
```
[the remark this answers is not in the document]
underhood: Ok I will make commit with minor - spelling fix

web/api/badges/web_buffer_svg.c
```diff
@@ -277,15 +149,26 @@ static inline double verdana11_width(char *s) {
     char *d = s;
 
     while(*s) {
-        double t = verdana11_widths[(unsigned char)*s];
-        if(t == 0.0)
-            s++;
+        double t;
```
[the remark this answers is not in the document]
underhood: Yep as your colleague mentioned that is useless var. Good Catch... Thanks

web/api/badges/web_buffer_svg.c
```diff
@@ -838,7 +721,7 @@ void buffer_svg(BUFFER *wb, const char *label, calculated_number value, const ch
     value_width = verdana11_width(value_string) + (BADGE_HORIZONTAL_PADDING * 2);
     total_width = label_width + value_width;
 
-    escape_xmlz(label_escaped, label_buffer, LABEL_STRING_SIZE);
+    escape_xmlz(label_escaped, label, LABEL_STRING_SIZE);
```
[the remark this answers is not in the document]
underhood: currently verdana11_width is just used to guess the size and should IMHO not remove any characters

web/api/badges/web_buffer_svg.c
underhood: Well i removed it because we know only sizes of ASCII characters. Everything above 0x80 is UTF char and for it we guess size of 1em now (which will be updated by browser with actual rendered text size). Therefore I saw no use of values above 0x80 as we will never access them anyway

web/api/badges/web_buffer_svg.c
underhood: reason is javascript code below which actually updates widths according to real rendered text size in browser. So we guess size with verdana11_width but browser in the end updates the values to real measured ones. This will also then work if some user doesn't have veranda font (fallback font will be used) etc.etc.

underhood:
Hi @thiagoftsm 

not sure I completely understand you in first point

> url_decode_multibyte_utf8(char *s, char *d, char *d_end)

The meaning here is:

- s - source
- d - destination
- d_end - last_byte of destination

Alternative may be source, destination and destination size.
Do you object to this or naming of variables?

For second point "double t" yep that slipped my attention I will remove it.

web/api/badges/web_buffer_svg.c
cakrit: Cool! I missed the JS code.

web/api/badges/web_buffer_svg.c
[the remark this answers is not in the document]
cakrit: You missed the question. I don't have a problem with not removing characters, but you made a decision and left behind code that became useless after that decision. The str copy and everything related to `char* d` inside the function should be removed, since there's no reason any more to update the string.

web/api/badges/web_buffer_svg.c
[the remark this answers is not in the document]
cakrit: With the JS change you made, can't we get rid of the sizing thing altogether, or will it show on the browser?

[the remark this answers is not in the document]
underhood: Well JS will overwrite the sizing if it is active (gets executed). If somebody has Javascript disabled then the values calculated by C code will be used.
Of course that can be improved too with enough time. There are ways to calculate it without hardcoded char size array. But I would make that a separate pull request in future.
Additionally I do not like also the way how HTML/CSS/JS (in this case just CSS/JS) is part of C code and I would move on to develop "Templating engine" of sorts.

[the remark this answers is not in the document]
cakrit: Got it. Just waiting for the 11-12pt fix so we can resolve this.

web/api/badges/web_buffer_svg.c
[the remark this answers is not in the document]
underhood: OK thanks for clarification. I see your point and you are right it can be "cleaned up".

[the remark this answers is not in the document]
underhood: updated in 80860f2
although as mentioned i suggest reworking this function to be able to properly handle unicode chars in future
i can eventually do this

web/api/badges/web_buffer_svg.c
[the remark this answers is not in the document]
underhood: updated in 80860f2
although as mentioned i suggest reworking this function to be able to properly handle unicode chars in future
i can eventually do this in future

ilyam8:
@underhood If you are still working on it please add `wip` to the PR title

underhood:
> @underhood If you are still working on it please add `wip` to the PR title

hi, almost done... sry for delay working on it only on lunch breaks and evenings

web/api/badges/web_buffer_svg.c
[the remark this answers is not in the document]
underhood: feb5711 and 35ea7ea solve this

ilyam8:
> hi, almost done... sry for delay working on it only on lunch breaks and evenings

no worries, mate

`wip` is needed to show that the work in progress and we dont need to review.

underhood:
> > hi, almost done... sry for delay working on it only on lunch breaks and evenings
> 
> no worries, mate
> 
> `wip` is needed to show that the work in progress and we dont need to review.

OK done for now with latest push... in case no one has new requests or see new problems

libnetdata/url/url.c
```diff
@@ -52,15 +52,78 @@ char *url_decode(char *str) {
...
+    return 0;
+}
+
+//this (utf8 string related) should be moved in separate file in future
+char url_utf8_get_byte_length(char c) {
+    if(!IS_UTF8_BYTE(c))
+        return 1;
+
+    char length = 0;
+    while(likely(c & 0x80)) {
```
ilyam8: i did ctrl-f `& 0x80` and found several places where you can use `IS_UTF8_BYTE`

underhood: Well for example in fucntion url_utf8_get_byte_length it is on purpose
because there i do not check is it UTF8 char but in cycle i take MSB bit out and count it.... so the condition is also c & 0x80 but meaning why is different

ilyam8: Nice, thanks for the explanation!

underhood:
Yep strange. Compiles fine here. If someone has idea come with it. Otherwise I will investigate on Saturday (tomorrow my schedule is packed)

libnetdata/string/utf8.h
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#ifndef NETDATA_STRING_UTF8_H
+#define NETDATA_STRING_UTF8_H 1
+
+#define IS_UTF8_BYTE(x) (x & 0x80)
+#define IS_UTF8_STARTBYTE(x) (IS_UTF8_BYTE(x)&&(x & 0x40))
+
+#endif /* NETDATA_STRING_UTF8_H */
```
thiagoftsm: Please, append a new line after this line.

libnetdata/libnetdata.h
```diff
@@ -312,5 +312,6 @@ extern char *netdata_configured_host_prefix;
 #include "url/url.h"
 #include "json/json.h"
 #include "health/health.h"
+#include "string/utf8.h"
```
thiagoftsm: We also use cmake to do compilation of netdata, this change is not enough to bring a new header.
Tip: Search where a header file that is inside libnetdata is declared in the Netdata project.

libnetdata/url/url.c
```diff
@@ -52,15 +52,78 @@ char *url_decode(char *str) {
...
     char *s = url,           // source
          *d = to,            // destination
          *e = &to[size - 1]; // destination end
 
     while(*s && d < e) {
         if(unlikely(*s == '%')) {
-            if(likely(s[1] && s[2])) {
-                char t = from_hex(s[1]) << 4 | from_hex(s[2]);
+            char t = url_percent_escape_decode(s);
+            if(t & 0x80) { //UTF-8 multibyte character
```
thiagoftsm: Why are you not using IS_UTF8_BYTE here?

libnetdata/url/url.c
```diff
@@ -52,15 +52,70 @@ char *url_decode(char *str) {
...
+    //10XX XXXX is not valid character -> check length == 1
+    if(length > 4 || length == 1)
+        return -1;
+    //0XXX XXXX means 1 byte
+    return (length) ? length : 1;
+}
+
+//decode % encoded UTF-8 characters and copy them to *d
+//return count of bytes written to *d
+char url_decode_multibyte_utf8(char *s, char *d, char* d_end) {
```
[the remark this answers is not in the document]
underhood: Aaah ok understand... You mean coding style `type* name` vs. `type *name`

underhood:
Put WIP smaller obvious fixes I will do asap but to %c2 issue I will not be able to get until weekend

libnetdata/libnetdata.h
underhood: Yep clear. Thanks for pointer. Thought it will be something like that. Will fix when at pc again. Went away after the push.

libnetdata/url/url.c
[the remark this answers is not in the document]
thiagoftsm: You got now! :)

web/api/badges/web_buffer_svg.c
```diff
@@ -139,153 +139,35 @@ static double verdana11_widths[256] = {
...
-    [246] = 0.0,
-    [247] = 0.0,
-    [248] = 0.0,
-    [249] = 0.0,
-    [250] = 0.0,
-    [251] = 0.0,
-    [252] = 0.0,
-    [253] = 0.0,
-    [254] = 0.0,
-    [255] = 0.0
```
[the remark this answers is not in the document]
underhood: Hi there are 2 parts to this at least so it seems to me at first glance:

1. URL parser %c2 issues -> will check that and come back to you. Today very busy with other things.
I know some parts of URL parser clearly needs improving. I do not like the old code which kind of tries to replace unknowns with spaces... IMHO this is dangerous and we need to have a way to "fail parsing" and return HTTP code 400 to client as soon as we see something suspicious. But as i mentioned I do not want to put everything into one huge pull request.
2. verdana11_width -> here I am confused maybe what you mean
With newest changes verdana11_width it doesn't modify the string and just returns width of it (best guess basis, as later it is updated by javascript)
```` C
if(likely(!(*s & 0x80))){ // Byte 1XXX XXXX is not valid in UTF8
     double t = verdana11_widths[(unsigned char)*s];
````
Here we check if the value is < 128 and only then we access this array (bound check). Your complaint is that we access array of size 128 with char which is -128 to 127 (0-255)? But wouldn't following be the same case?:
```` C
#define SAMPLE_ARRY_SIZE 10
int sample_arry[SAMPLE_ARRY_SIZE];
for (int i = 0; i < SAMPLE_ARRY_SIZE; i++){
   sample_arry[i] = whatever;
} 
````
You are accessing array of size 10 with integer which can be huge. That is however normal thing as long as you bound check before (which in C is responsibility of developer). Same you do with new verdana11_width code (there is bound check `*s & 0x80`).
Or is your point that you want to make that bound check more obvious with `#define`?

File: libnetdata/url/url.c (hunk 12 above)
Comment (underhood) [reply libnetdata/url/url.c]: yep

File: libnetdata/libnetdata.h (hunk 11 above)
Comment (underhood) [reply libnetdata/libnetdata.h]: Done in #6d101b4

File: web/api/badges/web_buffer_svg.c (hunk 14 above)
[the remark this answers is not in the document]
Comment (thiagoftsm) [reply web/api/badges/web_buffer_svg.c]: About 1, it is very good idea to concentrate in the problem instead the whole url parser, it was already rewritten and it will enter in test soon. We also know about answer problems that we have with it, and in the correct time it will be fixed. Thanks to remember us this. When you finish to solve the problem with %c2 let us know!
About 2 and the (unsigned char ) cast that follows,  I did some math converting the hex 0x80(128) to the binary and I agree with you, it will never go nothing bigger and equal than 128 due this conditional.
I won't mark this talk as resolved, because you cited the bug '%c2' here, but the rest is ok.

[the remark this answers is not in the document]
Comment (underhood) [reply web/api/badges/web_buffer_svg.c]: Hi, so I am able to get to it now. Credits to my wife for caring for the small one.

### %82
So when I sent http://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label=%82
I get:


Reason: %82 is invalid character in UTF-8 therefore we should reject it.
Good way to handle it: return 400 to client and do not attempt to parse further for security reasons. Maybe create issue for that and fix it as next step
**How is it handled now:**
%82 is recognized as invalid and % is skipped and "82" is then taken as normal string (not prepended by 82)

Original code from TAG v1.16.0:
````C
                char t = from_hex(s[1]) << 4 | from_hex(s[2]);
                // avoid HTTP header injection
                *d++ = (char)((isprint(t))? t : ' ');
                s += 2;
````
will skip also those 2 chars which I also can do. However as I mentioned in new code:
````C
                else {
                    //TODO!: Should stop parsing and send HTTP error code 400 back to client instead
                    //      same goes for other errors - if we see something suspicious abort and send 400
                    *d++ = ' ';
````

### %c2
I cannot reproduce the error you mention. With latest version I committed it behaves completely same as %82:
c2
Here the slight difference being that %c2 (1100 0010b) is first byte of multibyte character but there is no second byte... So at least at my PC with latest commit applied it correctly recognizes that this is not valid UTF8 character and rejects it by ignoring % and handling chars behind it as ASCII same as in case of %82.

Therefore IMHO same applies as before: best would be to return HTTP code 400 and immediately stop parsing (as URL is clearly malformed - maybe somebody is trying funky stuff etc.). Current solution is just ignore % char if there is something wrong.

>Unfortunately i can't reproduce the:
>XML Parsing Error: not well-formed
>Location: http://192.168.0.17:19999/api/v1/badge.svg?chart=system.cpu&label=%c2
>Line Number 1, Column 819:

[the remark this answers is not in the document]
Comment (underhood) [reply web/api/badges/web_buffer_svg.c]: Hi, so i took a look at the code and it's not that hard to do 400 error return instead. So in latest commit I implemented it. Now if not valid utf8 found we return 400 BAD REQUEST

**Comment by underhood** [comment]
Hi guys,

so I made it actually return HTTP STATUS 400 in case there is anything wrong in URL (as it should). Original code tried to skip strange chars but still used the input which was dangerous. Therefore this is improvement in any case.

**Comment by cakrit** [comment]
@underhood please sign the CLA again, we made a change to it.

File: CMakeLists.txt [hunk 15]
```diff
@@ -320,6 +320,7 @@ set(LIBNETDATA_FILES
         libnetdata/json/jsmn.h
         libnetdata/health/health.c
         libnetdata/health/health.h
+        libnetdata/string/uth8.h
```
cakrit: Spelling error here, you mean utf8.h

cakrit:
Thanks for the fix. Can you please also:
- Do a rebase so that we get rid of the Netlify errors? 
- Sign the CLA again?

underhood:
Yep working on it. CLA assistant is glitchy. First it was showing me empty license (with multiple refreshes) now it's stuck on loading after I pressed sign button.
glitch

underhood:
Done

thiagoftsm:
Hi @underhood ,

I  was testing your PR now, and I saw that for small utf8 your code is doing a good job, so I decided to go to the limits ( 5 and 6 bytes), and I got the same error I was having with %c2. Please check this for us!

Netdata4

underhood:
Add overlong UTF8 checking that was originally planned to add with next pull request.
Function that check this is under good license 
http://www.cl.cam.ac.uk/~mgk25/short-license.html

thiagoftsm:
> > I was testing your PR now, and I saw that for short utf8 your code is doing a good job, so I decided to go to the limits ( 5 and 6 bytes), and I got the same error I was having with %c2. Please check this for us!
> > Add overlong UTF8 checking that was originally planned to add with next pull request.
> 
> Well UTF8 chars longer than 4 bytes (not valid) are correctly rejected. Your failing case is reporting 4 byte char with 4 bytes given and %c0 in middle. This then is rejected by browser.
> 
> Added overlong UTF8 checking that was originally planned to add with next pull request. Therefore this and possible other issues are covered as well.
> Function that check this is under good license
> http://www.cl.cam.ac.uk/~mgk25/short-license.html

Hi @underhood, of course it is very important to have in our minds whether the input is valid or not, but returning for the output showed in the image, our current parser does not show any error for the wrong string given as argument. I know our parser is not complete yet, this is the motive that you are working on it, but there are some behaviors that it is interesting to keep. In my vision, at least, I think it is better to hidden the wrong characters instead to send the message "URL not valid. I don't understand you...".

underhood:
Well in case you have malformed Utf8 as input I believe returning 400 is correct way to go. If you just type special chars into the address bar browser will correctly fork url.  If you have something else it is probably some hacking attempt (or testing). I think it is quite dangerous trying to fix it. If you have Utf8 char that requires 4 bytes but there are 3... How many bytes do you skip? 3or4? What if 4th byte is beginning of next Utf8 char? What if 4th byte is start of url encoded ASCII char? I think trying to continue parsing url when we know it is not ok is inherently unsafe and sketchy (user will have some funky stuff in the badge)

cakrit:
> Well in case you have malformed Utf8 as input I believe returning 400 is correct way to go. If you just type special chars into the address bar browser will correctly form url. If you have something else it is probably some hacking attempt (or testing). I think it is quite dangerous trying to fix it. If you have Utf8 char that requires 4 bytes but there are 3... How many bytes do you skip? 3or4? What if 4th byte is beginning of next Utf8 char? What if 4th byte is start of url encoded ASCII char? I think trying to continue parsing url when we know it is not ok is inherently unsafe and sketchy (user will have some funky stuff in the badge)

This makes sense. Let's approve @thiagoftsm

PR #6428, opened 2019-07-10
New 'homepage' for documentation site

joelhans:
##### Summary

The main `README.md` file from the GitHub repo is too complex a starting point for the documentation site. If they're reading the documentation they're probably already using Netdata and don't need to be marketed to. So, I created a new "homepage" for the documentation site that's simpler and guides people to the most popular pages.

Use the Netlify preview to see how this looks and functions.

##### Component Name

docs

##### Additional Information

The `buildhtml.sh` script still spits out errors, one for each link in the `DOCUMENTATION.md` file, and I'm not exactly sure why. Everything builds fine.

```
Fixing links
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
sed: can't read docs/generator/doc/./DOCUMENTATION.md: No such file or directory
Calling mkdocs
```


DOCUMENTATION.md
```diff
@@ -0,0 +1,53 @@
...
+
+  </div>
+  <div class="nav-configuration">
+    <a class="nav-button" href="docs/configuration-guide/">Configuration guide</a>
+    <p>Take your configuration options from the <em>getting started guide</em> to the next level. Increase metrics retention, modify how charts are displayed, disable collectors, and modify alarms.</p>
+  </div>
+
+</div>
+
+4. **Advanced users**: For those who already understand how to access a Netdata dashboard and perform basic configuration, feel free to see what's behind any of these other doors.
```
cakrit: This 4 is a leftover now.

cakrit:
Barring the use of Javascript in the generated site, it looks to me like we should be replacing the README.md with the contents of DOCUMENTATION.md. This edit thing just won't work with the movement we did. 
I also tried renaming DOCUMENTATION.md to index.md and removing the replacements. The result was that DOCUMENTATION.md did become the home page, but we got a none link under the about, leading to a 404.

Both index.md and readme.md have special meaning in mkdocs, they work as index.html, with little we can do to modify that behavior...

joelhans:
Hmm. I hadn't noticed the edit icon issue. I have an idea:

- Create a new document at `docs/what-is-netdata.md` and copy-paste relevant material from the main README into it.
- Edit the build script so that once building is complete, overwrite `index.html` with the content of `DOCUMENTATION.html`.

This will give us the new homepage, and the edit buttons will go to the appropriate `.md` files in the repo. The issue created now is that we have two places to maintain content about what Netdata is (`README.md` and `docs/what-is-netdata.md`), but I'm okay with doing that maintenance work as needed.

I'll try this out and report back with my results.

joelhans:
Okay @cakrit, I think I've worked around the caveats you mentioned.

I added a new page at `docs/what-is-netdata.md` to contain that big-picture content that's currently in `README.md`. Since is mostly marketing, it's a bit more hidden now than it used to be.

Added L92-94 in `buildhtml.sh` to replace the HTML page built from `README.md` with the HTML page built from `DOCUMENTATION.md`. By doing it after the HTML is generated, the edit links and sidebar work as expected.

```
`# Replace index.html with DOCUMENTATION/index.html
echo "Replacing index.html with DOCUMENTATION/index.html"
cat ${GENERATOR_DIR}/${SITE_DIR}/DOCUMENTATION/index.html > ${GENERATOR_DIR}/${SITE_DIR}/index.html
```

PR #6435, opened 2019-07-11
Update docs health monitoring and health management api

jghaanstra:
##### Summary
Update docs health monitoring and health management api to include missing information about time scheduling of notifications and permission on using the mamagement api.

##### Component Name
Docs health monitoring and health management api

##### Additional Information
Is seems my code editor also removed some obsolete spaces, hence the many file changes even though I just added a couple of lines of text.

### DIFF
--- a/web/api/health/README.md
+++ b/web/api/health/README.md
@@ -65,7 +65,7 @@ You can access the API via GET requests, by adding the bearer token to an `Autho
 curl "http://myserver/api/v1/manage/health?cmd=RESET" -H "X-Auth-Token: Mytoken"
 ```
 
-By default access to the health management API is only allowed from `localhost`. Accessing the API from anything else will return a 403 error with the message `You are not allowed to access this resource.`. You can change permissions by editing the `allow management from` variable in netdata.conf within the [web] section. See [web server access lists](../../web/server/#access-lists) for more information.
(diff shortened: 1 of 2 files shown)


web/api/health/README.md
```diff
@@ -59,73 +59,75 @@ The API is available by default, but it is protected by an `api authorization to
...
-You can access the API via GET requests, by adding the bearer token to an `Authorization` http header, like this: 
+You can access the API via GET requests, by adding the bearer token to an `Authorization` http header, like this:
 
 ```
-curl "http://myserver/api/v1/manage/health?cmd=RESET" -H "X-Auth-Token: Mytoken" 
+curl "http://myserver/api/v1/manage/health?cmd=RESET" -H "X-Auth-Token: Mytoken"
 ```
 
-The command `RESET` just returns netdata to the default operation, with all health checks and notifications enabled. 
+By default access to the health management API is only allowed from `localhost`. Accessing the API from anything else will return a 403 error with the message `You are not allowed to access this resource.`. You can change permissions by editing the `allow management from` variable in netdata.conf within the [web] section. See [web server access lists](../../web/server/#access-lists) for more information.
```
joelhans: The link in this paragraph is causing a build error. My guess is that you don't need the `web/` part of the URL—try `[web server access lists](../../server/#access-lists)`—but it's hard to tell looking at it from here. Can you investigate? Thank you!

```
12:17:14 PM: -- ERROR: ./web/api/health/README.md - ./web/api/health/../../web/server/README.md does not exist
```

jghaanstra: I think you are right. Let's see how I can fix this, I'm not very experienced with Github.

PR #6440, opened 2019-07-12
Update Netdata via nginx

prhomhyse:
##### Summary

This guide works fine but it isn't really beginner friendly. 
So I explained why the user would perform certain actions and what some steps do. 

I also tested the configuration files on actual servers with both Nginx and Netdata installed. 

##### Component Name

docs


docs/Running-behind-nginx.md
```diff
@@ -1,8 +1,42 @@
...
+
+test Nginx configuration with `nginx -t` 
+
+restart Nginx to effect the change with `/etc/init.d/nginx restart` or `service nginx restart`.
+
+## Ways to pass Netdata via Nginx: ##
+
+### As a virtual host ###
+
+With this method instead of `123.45.67.89:19999`, the Netdata dashboard can be accessed via a human-readable URL such as `netdata.example.com` used in the configuration below.
```
ilyam8: There are special addresses for documentation and examples
https://tools.ietf.org/html/rfc5737

> The use of designated address ranges for
   documentation and examples reduces the likelihood of conflicts and
   confusion arising from the use of addresses assigned for some other
   purpose.

>    The blocks 192.0.2.0/24 (TEST-NET-1), 198.51.100.0/24 (TEST-NET-2),
   and 203.0.113.0/24 (TEST-NET-3) are provided for use in
   documentation.

Maybe we should use them instead of random public addresses.

> inetnum:        123.32.0.0 - 123.47.255.255
netname:        SAMSUNGSDS
descr:          SamsungSDS Inc.

prhomhyse: @ilyam8 I agree, I have edited the IP address to 192.0.2.0 reserved for documentation.

prhomhyse:
@cakrit I have made the requested changes. 
The Netlify build failed, I can't view the logs to figure out where in the docs is causing the failure.

cakrit:
Netlify has issues with the logs. I'll run docs/buildhtml.sh locally and let you know...

cakrit:
`-- ERROR: ./docs/netdata-security.md - #running-netdata-behind-nginx header not found in file ./docs/Running-behind-nginx.md`

cakrit:
The command is `docs/generator/buildhtml.sh` mkdocs will fail unless you have it installed, but you should be able to get to that part without ERROR messages from broken links.

cakrit:
In general, not easy to change to headings, as it's the only thing used to hyperlink from one doc to the other.

joelhans:
@prhomhyse, here's the relevant part of the writeup that I've been working on for building the docs locally. It'll be useful for you moving forward. I need to get around to submitting the PR so everyone can use it... maybe later today!

---

To build the documentation, you need `python`/`pip`, `mkdocs`, and `mkdocs-material` installed on your machine.

Follow the [Python installation instructions](https://www.python.org/downloads/) for your machine.

Use `pip`, which was installed alongside Python, to install `mkdocs` and `mkdocs-material`.

``` bash
$ pip install mkdocs mkdocs-material
```

??? note "Troubleshooting"
    If you're having trouble with the installation of Python, `mkdocs`, or `mkdocs-material`, try looking into the `mkdocs` [installation instructions](https://squidfunk.github.io/mkdocs-material/getting-started/#installation).

When that is finished, navigate to the root directory of the Netdata repository and run the documentation generator script.

``` bash
$ sh docs/generator/buildhtml.sh
```

This process will take some time. Once finished, the built documentation site will be located at `docs/generator/build/`.


### Run a local web server to test documentation

The best way to view the documentation site you just built is to run a simple web server from the `docs/generator/build/` directory. So, navigate there and run a Python-based web server:

```
$ cd docs/generator/build/
$ python3 -m http.server 20000
```

Feel free to replace the port number you want this web server to listen on (port `20000` in this case [only one higher than the agent!]).

Open your web browser and navigate to `http://localhost:20000`. If you replaced the port earlier, change it here as well. You can now navigate through the documentation as you would on the live site!

docs/Running-behind-nginx.md
```diff
@@ -1,8 +1,42 @@
-# Netdata via nginx
+# Running Netdata Behind Nginx #
```
cakrit: Why did you add these hash tags after the heading? It's not done anywhere else and doesn't cause issues if we don't have them. I DO see the link in Netlify (left navbar) having a # now, which is bad. Not sure what else may be affected...

docs/Running-behind-nginx.md
```diff
@@ -1,8 +1,42 @@
...
-### As a virtual host
+[Nginx](https://nginx.org/en/) is an HTTP and reverse proxy server, a mail proxy server, and a generic TCP/UDP proxy server used to host websites and applications of all sizes. 
+
+The software is known for its low impact on memory resources, high scalability, and its modular, event-driven architecture which can offer secure, predictable performance.
+
+## Why Nginx ##
+
+Nginx is used and useful in cases when you do not want to access Netdata via the default `IP ADDRESS:19999`.
+
+With Nginx, the data is accessed via human-readable domains such as `example.com` or `sub.example.com`.
```
cakrit: I don't need nginx for that, I just add a DNS entry. I do need nginx if I want to access different instances of netdata by using 'host/netdata/server1', 'host/netdata/server2'

cakrit: Also, in versions prior to 1.16.0, using a proxy was the only available way to encrypt the communication (https).

prhomhyse:
@cakrit Thank you, I have corrected the link. The Netlify tests are passed.

I did not have mkdocs installed on my machine, else I would have seen the error.

@joelhans I am already setting up mkdocs. It is useful.

docs/Running-behind-nginx.md
prhomhyse: I added them for consistency. 
I am used to adding the hash tags after headings from other guides I've written with markdown.

Removing them now so as to not cause issues.

docs/Running-behind-nginx.md
prhomhyse: Sure, you do need Nginx if the DNS entry is to work. 
A lot of cases, it is easier to remember the domains than IPs. 

More than to just add a DNS entry, you need to configure Nginx to serve the domain.

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
...
+
+You can edit the Nginx configuration file with Nano, Vim or any other Text editors you are comfortable with.
+
+After making changes to the configuration files:
+
+test Nginx configuration with `nginx -t` 
+
+restart Nginx to effect the change with `/etc/init.d/nginx restart` or `service nginx restart`.
+
+## Ways to pass Netdata via Nginx:
```
joelhans: No need for the colon (`:`) at the end of this header.

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
...
+
+Configuration options in Nginx are known as directives. Directives are organized into groups known as blocks or contexts. The two terms can be used interchangeably.
+
+Depending on your installation source, you’ll find an example configuration file at `/etc/nginx/conf.d/default.conf` or `etc/nginx/sites-enabled/default`, in some cases you may have to manually create the `sites-available` and `sites-enabled` directories. 
+
+You can edit the Nginx configuration file with Nano, Vim or any other Text editors you are comfortable with.
+
+After making changes to the configuration files:
+
+test Nginx configuration with `nginx -t`
```
joelhans: Could you put these two instructions into an ordered list? Might look cleaner that way.

```suggestion
- Nginx configuration with `nginx -t`
- Restart Nginx to effect the change with `/etc/init.d/nginx restart` or `service nginx restart`
```

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
...
+
+### Nginx configuration file
+
+All Nginx configurations can be found in the `/etc/nginx/` directory. The main configuration file is `/etc/nginx/nginx.conf`. Website or app-specific configurations can be found in the `/etc/nginx/site-available/` directory.
+
+Configuration options in Nginx are known as directives. Directives are organized into groups known as blocks or contexts. The two terms can be used interchangeably.
+
+Depending on your installation source, you’ll find an example configuration file at `/etc/nginx/conf.d/default.conf` or `etc/nginx/sites-enabled/default`, in some cases you may have to manually create the `sites-available` and `sites-enabled` directories. 
+
+You can edit the Nginx configuration file with Nano, Vim or any other Text editors you are comfortable with.
```
joelhans: No need to capitalize "Text":

```suggestion
You can edit the Nginx configuration file with Nano, Vim or any other text editors you are comfortable with.
```

docs/Running-behind-nginx.md
prhomhyse: Noted. I have removed it

docs/Running-behind-nginx.md
prhomhyse: Putting the two instructions in an unordered list does make it cleaner. 
I did not remove **test** because it explains why we run `nginx -t`.

docs/Running-behind-nginx.md
prhomhyse: I have corrected that. 
Thanks

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
-# Netdata via nginx
+# Running Netdata Behind Nginx
```
cakrit: Did we decide to start using [Capitalization instead of Sentence Case](https://danielmiessler.com/blog/a-list-of-different-case-types/)?
If so, then we need to be consistent everywhere. We currently have only a couple of titles using Capitalization, which I see as bugs.

docs/Running-behind-nginx.md
cakrit: No, I mean that I can just have a DNS entry pointing to the IP of my netdata! Put the following in your `/etc/hosts` and go to http://netdata.mydomain.com:19999. You will see netdata there.
```
127.0.1.1	netdata.mydomain.com
```

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
...
+
+Nginx is used and useful in cases when you do not want to access Netdata via the default `IP ADDRESS:19999`.
+
+With Nginx, the data is accessed via human-readable domains such as `example.com` or `sub.example.com`. 
+
+By default, Nginx is fast and lightweight out of the box.
+
+Nginx can be configured to secure your Netdata insights by the use of authentication to access the dashboard. 
+
+### Nginx configuration file
```
cakrit: This doesn't belong under Why Nginx.

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
...
+
+You can edit the Nginx configuration file with Nano, Vim or any other text editors you are comfortable with.
+
+After making changes to the configuration files:
+
+- test Nginx configuration with `nginx -t` 
+
+- restart Nginx to effect the change with `/etc/init.d/nginx restart` or `service nginx restart`.
+
+## Ways to pass Netdata via Nginx
```
cakrit: Ways to access Netdata via Nginx

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
...
+
+Configuration options in Nginx are known as directives. Directives are organized into groups known as blocks or contexts. The two terms can be used interchangeably.
+
+Depending on your installation source, you’ll find an example configuration file at `/etc/nginx/conf.d/default.conf` or `etc/nginx/sites-enabled/default`, in some cases you may have to manually create the `sites-available` and `sites-enabled` directories. 
+
+You can edit the Nginx configuration file with Nano, Vim or any other text editors you are comfortable with.
+
+After making changes to the configuration files:
+
+- test Nginx configuration with `nginx -t`
```
cakrit: I believe we use Sentece Case for such lists as well.

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
...
+
+Depending on your installation source, you’ll find an example configuration file at `/etc/nginx/conf.d/default.conf` or `etc/nginx/sites-enabled/default`, in some cases you may have to manually create the `sites-available` and `sites-enabled` directories. 
+
+You can edit the Nginx configuration file with Nano, Vim or any other text editors you are comfortable with.
+
+After making changes to the configuration files:
+
+- test Nginx configuration with `nginx -t` 
+
+- restart Nginx to effect the change with `/etc/init.d/nginx restart` or `service nginx restart`.
```
cakrit: This one has a dot at the end, the previous item on the list doesn't. We need to be consistent.

docs/Running-behind-nginx.md
```diff
@@ -1,9 +1,43 @@
...
+
+## Ways to pass Netdata via Nginx
 
 ### As a virtual host
 
+With this method instead of `192.0.2.0:19999`, the Netdata dashboard can be accessed via a human-readable URL such as `netdata.example.com` used in the configuration below.
```
cakrit: What does "instead of SOME RANDOM IP" mean? 
Human-readable URL is not the point here, as explained above.

docs/Running-behind-nginx.md
cakrit: Also, you didn't add the two valid reasons I mentioned for using a reverse proxy:
- Password-protect access to netdata, until distributed authentication is implemented via the netdata cloud Sign In mechanism.
- A proxy was necessary to encrypt the communication to netdata, until v1.16.0, which provided TLS (https) support.

docs/Running-behind-nginx.md
```diff
@@ -114,8 +152,8 @@ Of course you can add as many backend servers as you like.
 
 Using the above, you access Netdata on the backend servers, like this:
 
-- `http://nginx.server/netdata/server1/` to reach `backend-server1`
-- `http://nginx.server/netdata/server2/` to reach `backend-server2`
+- `http://netdata.example.com/netdata/server1/` to reach `backend-server1`
+- `http://netdata.example.com/netdata/server2/` to reach `backend-server2`
 
 ### Using TLS communication
```
cakrit: Try to think of something nicer for this, that is also using the verb consistently with the other titles below. e.g. 'Limit direct access to Netdata' and 'Encrypt the communication between Nginx and Netdata', or both turned to 'Limiting' and 'Using'.
Perhaps you'll leave it as just 'Use TLS communication', though I think we put the wrong title here. We have a link to this header further down in the document as well, the build process will warn you if you leave something out.

docs/Running-behind-nginx.md
prhomhyse: For consistency, I agree with sentence case. 
I have changed made the change.

docs/Running-behind-nginx.md
prhomhyse: Yes, adding that to /etc/hosts also works. 

I have removed the sentence, and paraphrased it under ["Enable authentication"](https://docs.netdata.cloud/docs/running-behind-nginx/#enable-authentication)

thiagoftsm:
An additional information about nginx, @cakrit called my attention this morning for the next group of sentences:

"In case Netdata’s web server has been configured to use TLS, you must also encrypt the communication between Nginx and Netdata.

To enable encryption, first enable TLS/SSL on Nginx and then put the following in the location section of your Nginx configuration:"

We realized that we were missing something, because in some tests made today I could verify that the user is not forced to encrypt the communication between Nginx and Netdata, Nginx will do this alone, so I will give a suggestion here to fix these paragraphs:

"In case Netdata’s web server has been configured to use TLS, it is necessary to specify inside the Nginx configuration that the final destination is using TLS. To do this, please, append the following parameters in your nginx.conf

proxy_set_header X-Forwarded-Proto https;
proxy_pass https://localhost:19999;

Optionally it is also possible to enable TLS/SSL on Nginx, this way the user will have encrypt not only the communication between Nginx and Netdata, but also between the user and Nginx."

Best Regards!

docs/Running-behind-nginx.md
prhomhyse: I have removed the previous statement 

> Nginx is used and useful in cases when you do not want to access Netdata via the default `IP 
ADDRESS:19999`.

>With Nginx, the data is accessed via human-readable domains such as `example.com` or >`sub.example.com`

and added the valid reasons.

docs/Running-behind-nginx.md
prhomhyse: I earlier wrote `123.45.67.89:19999` and @ilyam8 pointed out that I should use IPs reserved for documentation purposes. 
I have replaced it with `IP ADDRESS:19999`. 

The point of the sentence is to tell a user/reader why they would want to access Netdata via Nginx as a virtual host.

Although adding `127.0.1.1	netdata.mydomain.com` to `/etc/hosts` still works fine, we are addressing how it's done with Nginx.

To prevent repetition, I have rewritten the previous sentence in **Why Nginx**.

prhomhyse: For self-explanatory purposes, I have further changed it from just `IP ADDRESS:19999` to `SERVER_IP_ADDRESS:19999`.

prhomhyse:
@cakrit I have made the changes.

Thanks! @thiagoftsm

prhomhyse:
I build the local documentation with MkDocs and found out docs/generator/doc was not ignored

docs/Running-behind-nginx.md
cakrit: Actually, I was referring to the heading level of 'Nginx configuration file'. We have it as a child of 'Why Nginx', which isn't correct. You probably want to make it the same level as 'Why Nginx'.

cakrit:
Two things still missing, note the left over comments.
To build, just use docs/generator/buildhtml.sh. If you want to user mkdocs directly, you'll need to tell it to use the correct mkdocs.yaml. I haven't tried it on its own for a while, send me the command you use and I'll try to replicate.

I did make some changes to the build process a few days ago, you'd also need to ensure your branch is updated (git rebase). The netlify checks in the PR are correct.

cakrit:
You accidentally brought into this PR multiple, unrelated commits. 
It will be messy to clean up now. I suggest starting a new PR from a clean code base. I've found the following to be extremely helpful: 
https://medium.com/@ruthmpardee/git-fork-workflow-using-rebase-587a144be470

prhomhyse:
Arrrrgh, my error. 
I will close this and start a new PR.

Thank you

PR #6455, opened 2019-07-14
Fix nodes menu sizing (responsive)

### LINKED ISSUE #5812: my-netdata menu dynamic sizing
We need to fix the following:

Screenshot from 2019-04-05 19-03-07

builat:
##### Summary
Fixes #5812 
This PR affects top menu dropdown with hostname.

##### Component Name
web

##### Additional Information
Solution based on different layouts for different display sizes.


ilyam8:
@builat please sign the CLA

builat:
@ilyam8 done.

VLegakis:
Hello @builat ,
Thank you for this. Seems good in most cases. In some others though the previous implementation seemed a bit better (following screenshots on the improvement of this PR + previous case)
We could improve it a bit further I suppose if for the rest it looks good:

With this PR:
Screenshot 2019-07-15 at 14 17 30
Previous implementation:
Screenshot 2019-07-15 at 14 17 30
With this PR:
Screenshot 2019-07-15 at 14 14 44
Previous implementation
Screenshot 2019-07-15 at 14 15 44

VLegakis:
Hello @builat,
It looks very nice. 
As I have posted above, the only improvement we have found would be in the last screen you posted (smallest screen size) it would be nice to show the node's name (instead of only the menu arrow) since this is the way it was working before (see my previous attachments too) and there is a lot of space for the node name to show there.
It's not something very important. Let me know and depending on your availability we can proceed.
Many thanks again,
V.

builat:
@VLegakis Ok. I will implemet this part the way you say in few days. Your point is more reasonable then mine.

@cakrit Yes, you right. As far as I discovered in my code research, I can confirm that there are at list one (may be more) **.png** file not used in current production layout.

builat:
@cakrit probably it is a better idea to create separate issue for removing unused content from build.

VLegakis:
Great! Thanks for the improvements @builat! 
Approved

PR #6460, opened 2019-07-15
json function could create overflow

thiagoftsm:
##### Summary
When the netdata uses its internal json library instead json-c, it is possible to have an overflow, because the sprintf was creating a source that could be bigger than the destination, to avoid this problem it is necessary to use snprintf.
##### Component Name
json library
##### Additional Information
Alert reported by Markos in an internal chat.


libnetdata/json/json.c
```diff
@@ -319,8 +319,8 @@ size_t json_walk_array(char *js, jsmntok_t *t, size_t nest, size_t start, JSON_E
             info("JSON: JSON walk_array ignoring element with name:%s fullname:%s",e->name, e->fullname);
             continue;
         }
-        sprintf(ne.name, "%s[%lu]", e->name, i);
-        sprintf(ne.fullname, "%s[%lu]", e->fullname, i);
+        snprintf(ne.name, JSON_NAME_LEN, "%s[%lu]", e->name, i);
+        snprintf(ne.fullname, JSON_FULLNAME_LEN, "%s[%lu]", e->fullname, i);
```
mfundul: You should be using snprintfz that netdata typically uses since it's safer.

thiagoftsm: Ok!  In this last commit I changed snprintf to snprintfz.

PR #6466, opened 2019-07-15
update Nginx guide with changes

prhomhyse:
##### Summary
This guide works fine but it isn't really beginner friendly.
So I explained why the user would perform certain actions and what some steps do.

I also tested the configuration files on actual servers with both Nginx and Netdata installed.

##### Component Name
docs

##### Additional Information

This is a clean PR from a previously closed one.
It has the requested changes and updates

### DIFF
--- a/docs/Running-behind-nginx.md
+++ b/docs/Running-behind-nginx.md
@@ -1,8 +1,42 @@
-# Netdata via nginx
+# Running Netdata behind Nginx
 
-To pass Netdata via a nginx, use this:
+## Intro
+
+[Nginx](https://nginx.org/en/) is an HTTP and reverse proxy server, a mail proxy server, and a generic TCP/UDP proxy server used to host websites and applications of all sizes. 
+
+The software is known for its low impact on memory resources, high scalability, and its modular, event-driven architecture which can offer secure, predictable performance.
+
+## Why Nginx
+
+- By default, Nginx is fast and lightweight out of the box.
+
+- Nginx is used and useful in cases when you want to access different instances of Netdata from a single server.
+
+- Password-protect access to Netdata, until distributed authentication is implemented via the Netdata cloud Sign In mechanism.
+
+- A proxy was necessary to encrypt the communication to netdata, until v1.16.0, which provided TLS (HTTPS) support.
+
+## Nginx configuration file
+
+All Nginx configurations can be found in the `/etc/nginx/` directory. The main configuration file is `/etc/nginx/nginx.conf`. Website or app-specific configurations can be found in the `/etc/nginx/site-available/` directory.
+
+Configuration options in Nginx are known as directives. Directives are organized into groups known as blocks or contexts. The two terms can be used interchangeably.
+
+Depending on your installation source, you’ll find an example configuration file at `/etc/nginx/conf.d/default.conf` or `etc/nginx/sites-enabled/default`, in some cases you may have to manually create the `sites-available` and `sites-enabled` directories. 
+
+You can edit the Nginx configuration file with Nano, Vim or any other text editors you are comfortable with.
+
+After making changes to the configuration files:
+
+- Test Nginx configuration with `nginx -t`.
+
+- Restart Nginx to effect the change with `/etc/init.d/nginx restart` or `service nginx restart`.
+
+## Ways to access Netdata via Nginx
 
 ### As a virtual host
+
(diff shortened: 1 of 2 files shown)


docs/Running-behind-nginx.md
```diff
@@ -114,25 +152,25 @@ Of course you can add as many backend servers as you like.
...
 
-If nginx is not configured as described here, you will probably receive the error `SSL_ERROR_RX_RECORD_TOO_LONG`.
+Optionally it is also possible to [enable TLS/SSL on Nginx](http://nginx.org/en/docs/http/configuring_https_servers.html), this way the user will encrypt not only the communication between Nginx and Netdata but also between the user and Nginx.
+
+If Nginx is not configured as described here, you will probably receive the error `SSL_ERROR_RX_RECORD_TOO_LONG`.
 
 ### Enable authentication
 
-Create an authentication file to enable basic authentication via Nginx. Do not use authentication without having first [enabled TLS](#using-tls-communication)!
+Create an authentication file to enable basic authentication via Nginx, this secures your Netdata dashboard. Do not use authentication without having first [enabled TLS](#using-tls-communication)!
```
cakrit: Please remove the "Do not use authentication without..." completely. 
Perhaps Thiago's comment in the last PR was confusing, but this statement is completely wrong and we didn't catch it.

docs/Running-behind-nginx.md
```diff
@@ -114,25 +152,25 @@ Of course you can add as many backend servers as you like.
 
 Using the above, you access Netdata on the backend servers, like this:
 
-- `http://nginx.server/netdata/server1/` to reach `backend-server1`
-- `http://nginx.server/netdata/server2/` to reach `backend-server2`
+- `http://netdata.example.com/netdata/server1/` to reach `backend-server1`
+- `http://netdata.example.com/netdata/server2/` to reach `backend-server2`
 
 ### Using TLS communication
```
cakrit: Chang to: Encrypt the communication between nginx and netdata.
Thiago had comments on how this should be written, I'll let him review the new statements.

thiagoftsm: Hi @cakrit the encrypt communication was changed according my initial suggestion

docs/Running-behind-nginx.md
thiagoftsm: I agree with @cakrit , let me try to explain better. When we have the Netdata set with TLS, it is a very good idea to use it, this is the motive we adjusted the text between the lines 160-167.
When we "enable authentication" in the Nginx, it does not have direct relationship with the "using-tls-communicaton", it is completely possible to have the TLS without to use Nginx to authenticate and the opposite is also true, so please remove "Do not use authentication without having first [enabled TLS](#using-tls-communication)!"

prhomhyse: It was a bit confusing to me as well.
I have effected the change

docs/Running-behind-nginx.md
prhomhyse: 'Encrypt the communication between Nginx and Netdata' looks better than just 'Using TLS communication'

@thiagoftsm you suggested its content.

@cakrit I have updated the header title.

docs/Running-behind-nginx.md
thiagoftsm: Let me explain better the situation. When we activate the TLS on Netdata, this means that connections direct with Netdata will use encrypt channel, but the user is free to allow connections between him and Nginx without encryptation. In the same way, it is possible the user thinks that it is not necessary to encrypt the connection between Nginx and netdata,  because he created an authentication in his Nginx, so for I person access the Netdata, it will be necessary to authenticate firstly in Nginx. Finally, an user that wanna have everything encrypted, can activate TLS on Netdata and also set Auhtentication in Nginx with TLS. All the scenarios are possible, our goal is only to say for the user that they are possible, but there is not any obligation to activate two channels together or three, the user is free to decide.

docs/Running-behind-nginx.md
thiagoftsm: Your sentence let the things more explicit, but there are different ways of encrypt a channel, TLS is one of them, considering that we are already using the TLS in the description, I do not see any problem in change the title.

docs/Running-behind-nginx.md
prhomhyse: I understand now.

The goal is for the user to have freedom of choice from different possibilities.

thiagoftsm: Exactly!

PR #6472, opened 2019-07-16
Get user and group names from files 

### LINKED ISSUE #5585: Cant see user name when run netdata in docker
##### Question summary
I run netdata in docker, the user names in the "per user statistics" under `Users` tag are not shown, instead it shows the user id (except for `root` `netdate` `xfs`). I want see the user name in the legand.

It looks like the problem is about docker not netdata, but i search it on google and fail to solve it.

below is my docker-composer file:
```
version: "3"
services:
  netdata:
    image: netdata/netdata
    ports:
      - 19999:19999
    cap_add:
      - SYS_PTRACE
    environment:
      - PGID=999
    security_opt:
      - apparmor:unconfined
    volumes:
      - /proc:/host/proc:ro
      - /sys:/host/sys:ro
      - /var/run/docker.sock:/var/run/docker.sock:ro
    restart: always
```

##### OS / Environment
ubuntu 18.04
docker 18.09.3

vlvkobal:
##### Summary
When Netdata runs inside a container, it can't display correct names for "User" and "User Groups" sections. That is because Netdata uses `getpwuid()` and `getgrgid()` functions, which get local user and group names, while we want to monitor processes outside the container.

We will read `/etc/passwd` and `/etc/group` files and fall back to `getpwuid()` and `getgrgid()` only if the files can't be read. To read the host files from the container the `/etc` folder should be mounted with the `-v /etc:/host/etc:ro` option.

Fixes #5585

##### Component Name
apps plugin

### DIFF
--- a/packaging/docker/README.md
+++ b/packaging/docker/README.md
@@ -24,7 +24,8 @@ This is good for an internal network or to quickly analyse a host.
 ```bash
 docker run -d --name=netdata \
   -p 19999:19999 \
-  -v /etc:/host/etc:ro \
+  -v /etc/passwd:/host/etc/passwd:ro \
+  -v /etc/group:/host/etc/group:ro \
   -v /proc:/host/proc:ro \
   -v /sys:/host/sys:ro \
   -v /var/run/docker.sock:/var/run/docker.sock:ro \
@@ -48,11 +49,14 @@ services:
     security_opt:
       - apparmor:unconfined
     volumes:
-      - /etc:/host/etc:ro
+      - /etc/passwd:/host/etc/passwd:ro
+      - /etc/group:/host/etc/group:ro
       - /proc:/host/proc:ro
       - /sys:/host/sys:ro
       - /var/run/docker.sock:/var/run/docker.sock:ro
 ```
+
+If you don't want to use the apps.plugin functionality, you can remove the mounts of `/etc/passwd` and `/etc/group` (they are used to get proper user and group names for the monitored host) to get slightly better security.
 
 ### Docker container names resolution
 
@@ -134,7 +138,8 @@ services:
     security_opt:
       - apparmor:unconfined
     volumes:
-      - /etc:/host/etc:ro
+      - /etc/passwd:/host/etc/passwd:ro
+      - /etc/group:/host/etc/group:ro
       - /proc:/host/proc:ro
       - /sys:/host/sys:ro
       - /var/run/docker.sock:/var/run/docker.sock:ro
--- a/packaging/docker/README.md
+++ b/packaging/docker/README.md
@@ -24,6 +24,7 @@ This is good for an internal network or to quickly analyse a host.
 ```bash
 docker run -d --name=netdata \
   -p 19999:19999 \
+  -v /etc:/host/etc:ro \
   -v /proc:/host/proc:ro \
   -v /sys:/host/sys:ro \
   -v /var/run/docker.sock:/var/run/docker.sock:ro \
@@ -47,6 +48,7 @@ services:
     security_opt:
       - apparmor:unconfined
     volumes:
+      - /etc:/host/etc:ro
       - /proc:/host/proc:ro
       - /sys:/host/sys:ro
       - /var/run/docker.sock:/var/run/docker.sock:ro
@@ -132,6 +134,7 @@ services:
     security_opt:
       - apparmor:unconfined
     volumes:
+      - /etc:/host/etc:ro
       - /proc:/host/proc:ro
       - /sys:/host/sys:ro
       - /var/run/docker.sock:/var/run/docker.sock:ro
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ To try Netdata in a docker container, run this:
 ```
 docker run -d --name=netdata \
   -p 19999:19999 \
+  -v /etc:/host/etc:ro \
   -v /proc:/host/proc:ro \
   -v /sys:/host/sys:ro \
   -v /var/run/docker.sock:/var/run/docker.sock:ro \
--- a/README.md
+++ b/README.md
@@ -98,7 +98,8 @@ To try Netdata in a docker container, run this:
 ```
 docker run -d --name=netdata \
   -p 19999:19999 \
-  -v /etc:/host/etc:ro \
+  -v /etc/passwd:/host/etc/passwd:ro \
+  -v /etc/group:/host/etc/group:ro \
   -v /proc:/host/proc:ro \
   -v /sys:/host/sys:ro \
   -v /var/run/docker.sock:/var/run/docker.sock:ro \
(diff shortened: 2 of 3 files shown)


mfundul:
Can it be a problem if the system uses NIS or LDAP instead of `/etc/passwd`? `getpwuid` apparently supports all those password databases.

Also, will this be a problem for users who want to monitor their container instead of their host?

vlvkobal:
That's why I left a fallback to `getpwuid()` and `getgrgid()`. If somebody wants to monitor his container, he should just remove `-v` flags from the `docker run` command.

mfundul:
Is there a possibility that the user will get wrong info by successfully reading `/etc/passwd` instead of doing `getpwuid()`?

vlvkobal:
There is such a possibility in case the system is not using these files, but they are still in place.

ilyam8:
i dont see any info was it tested or not and how, lack of WIP means it was, but i think it should be explicitly mentioned

vlvkobal:
Yes. It was tested in both: a host system and a docker container.

paulkatsoulakis:
What about security concerns?
Are we asking the users to let the container read the passwd file always by default?

cakrit:
@Ferroin can we have your view on this? Is it a good idea?

mfundul:
When users install netdata in a container to monitor the host instead of the container itself, it is only natural that this container will have almost full access to host resources.

I fail to see where netdata in this container is less secure than installing netdata directly at the host.

Ferroin:
General thoughts on this:

* Barring serious bugs in Docker or the kernel components it leverages, this cannot by definition be objectively less secure than running Netdata directly on the host.  Most users, however, will expect it to be significantly more secure, so there's a risk of this resulting in a false sense of security.
* If possible, mounting only the `/etc/passwd` and `/etc/group` files instead of all of `/etc` would be preferred (if, at some point in the future, Netdata needs a lot more files out of `/etc`, then maybe it might make sense to mount all of `/etc`).  I'm not sure if Docker can do that or not though.
* This obviously won't work with authentication systems that don't use `/etc/passwd` or `/etc/group`.  The common case for this (LDAP) should probably be handled somehow, but I don't think this PR is the place to do so as it will likely need changes to both the Docker image itself, and Netdata.

mfundul:
Moreover, as far as I know, reading `/etc/passwd` is not an issue since this file is usually readable by all users and it doesn't contain the encrypted passwords anymore, since people store them in `/etc/shadow`.

What I suspect is that using LDAP is going to be an issue since `/etc/passwd` may be inaccurate.

vlvkobal:
Well, the way we have it now is not more accurate in that case.

mfundul:
> 
> 
> Well, the way we have it now is not more accurate in that case.

I agree that this PR is strictly better than what we have for netdata in a container. Moreover, it doesn't regress when netdata runs in the host since it falls back to `getpwuid()` in that case.

Ferroin:
> What I suspect is that using LDAP is going to be an issue since /etc/passwd may be inaccurate.

The best solution for that is probably going to be teaching Netdata to just speak to the LDAP server and query it directly.  If done right, it wouldn't even need anything to be passed through from the host system.

But, of course, that's a likely to be a lot of work and include adding yet another (ideally optional) dependency, hence my comment that this PR is probably not the best place to worry about it.

Ferroin:
Oh, also, forgot to mention, I am in favor of adding this (it is in almost always strictly better than what we have right now), but I do think it should include something in the docs explaining that if you don't want to use the apps.plugin functionality, you can remove the mount of `/etc` to get slightly better security.

cakrit:
docker run accepts single files in -v

https://github.com/moby/moby/issues/10616#issuecomment-92106473

Let's just suggest that user's mount the files we need

paulkatsoulakis:
So in order to close this PR lets do:
1) Mount only the needed files, instead of the full directory.
2) Update the documentation in apps.plugin to instruct what the user may add to the config when running on docker, so that apps.plugin can show the names accurately

packaging/docker/README.md
```diff
@@ -48,12 +49,15 @@ services:
...
     volumes:
-      - /etc:/host/etc:ro
+      - /etc/passwd:/host/etc/passwd:ro
+      - /etc/group:/host/etc/group:ro
       - /proc:/host/proc:ro
       - /sys:/host/sys:ro
       - /var/run/docker.sock:/var/run/docker.sock:ro
 ```
 
+If you don't want to use the apps.plugin functionality, you can remove the mounts of `/etc/passwd` and `/etc/group` (they are used to get proper user and group names for the monitored host) to get slightly better security.
```
ilyam8: Do we really need that `slightly` word?

ilyam8:
My 2 cents.

> you can remove the mounts of `/etc/passwd` and `/etc/group` (they are used to get proper user and group names for the monitored host) to get slightly better security

>  slightly better security

> better security

If you aware of any security issues with moutning `/etc/passwd` and `/etc/group` please explain instead of `slightly better security`.
If no i suggest to remove security part completely. I am not aware of any btw, those files are word readable on any system i know. (not a security specialist).

Ferroin:
> If you aware of any security issues with moutning /etc/passwd and /etc/group please explain them instead of slightly better security.
> If no i suggest to remove security part completely. I am not aware of any btw, those files are word readable on any system i know. (not a security specialist).

It's an information disclosure risk.  Just because something is world readable by local users doesn't mean you want the whole world to be able to read it.

More concretely, the `/etc/passwd` file is extremely useful for planning attacks on a system.  It tells you who can actually log in, what shell everybody uses, some limited information about the directory structure (from the home directory entries), some limited information about what type of software is used by the system (stupid example, but if you see a user called `netdata`, you can be reasonably sure the system uses Netdata or did so at one point), and helps narrow down what type of system it is (certain distributions use specific UID to username mappings (especially the base system users (stuff like `bin` or `daemon`) differ between some Linux distros).  The `/etc/group` file is not as useful to an attacker, but can still help with planning (oh, this user can log in, has a shell I can use, and is in a bunch of administrative groups, they should be a priority target for SSH brute-forcing).  Both can also be used to aid in victim selection for social engineering attacks.

Now, the risk here is relatively low (hence the 'slightly') because Netdata's server component is reasonably secure, but it's still a risk.

ilyam8:
>  Just because something is world readable by local users doesn't mean you want the whole world to be able to read it.

Yes, i completely agree. 

But how is not mounting those 2 files gives you a better security? Someone hacked netdata container and got the opportunity to read the files to plan the attack?

Ferroin:
> But how is not mounting those 2 files gives you a better security? Someone hacked netdata container and got the opportunity to read the files to plan the attack?

No, someone hacked the Netdata container and then used the files to plan an attack on the host system that's running the container.  You have to remember we're exposing host data into a container here.

ilyam8:
prom node exporter docker install example
https://github.com/prometheus/node_exporter#using-docker

PR #6474, opened 2019-07-16
Add a statement about permissions for the diskspace plugin

### LINKED ISSUE #6401: No LVM disk space usage on CentOS
# Question summary
the lvm disk space usage isn't monitoring.
##### OS / Environment
CentOS 7.3

I did reviewed issue #3696 but couldn't understand anything

vlvkobal:
##### Summary
Make clear that Netdata needs search permission for a mount point in order to monitor it. 

Fixes #6401

##### Component Name
diskspace plugin

### DIFF
--- a/collectors/diskspace.plugin/README.md
+++ b/collectors/diskspace.plugin/README.md
@@ -1,11 +1,6 @@
 # diskspace.plugin
 
-This plugin monitors the disk space usage of mounted disks, under Linux. The plugin requires Netdata to have execute/search permissions on the mount point itself, as well as each component of the absolute path to the mount point. You can create a separate group to control access to mount points, include `netdata` user to the group, and set the execute/search permissions using
-
-```sh
-chown :volumeadm /mnt/volume
-chmod g+x /mnt/volume
-```
+This plugin monitors the disk space usage of mounted disks, under Linux. The plugin requires Netdata to have execute/search permissions on the mount point itself, as well as each component of the absolute path to the mount point.
 
 Two charts are available for every mount:
  - Disk Space Usage
--- a/collectors/diskspace.plugin/README.md
+++ b/collectors/diskspace.plugin/README.md
@@ -1,6 +1,6 @@
 # diskspace.plugin
 
-This plugin monitors the disk space usage of mounted disks, under Linux.
+This plugin monitors the disk space usage of mounted disks, under Linux. To be able to get the statistics, Netdata needs to have the search permission on the mount point.
 
 Two charts are available for every mount:
  - Disk Space Usage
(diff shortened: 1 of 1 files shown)


collectors/diskspace.plugin/README.md
```diff
@@ -1,6 +1,6 @@
 # diskspace.plugin
 
-This plugin monitors the disk space usage of mounted disks, under Linux.
+This plugin monitors the disk space usage of mounted disks, under Linux. To be able to get the statistics, Netdata needs to have the search permission on the mount point.
```
Ferroin: I'd suggest rewording this new sentence to:

`This plugin requires Netdata to have execute permissions on the mount point itself, as well as each component of the absolute path to the mount point.'

This explicitly spells things out in a way that a user who doesn't have a bunch of UNIX background should understand and be able to translate into an actual fix, whereas 'search permission' is something most people are going to have to look up (even I had to think for a minute to realize that it was talking about the execute bit being set in the directory permissions).  Also, I think it's important to point out that the permission is needed for each component of the path to the mount point, not just the mount point itself.

mfundul: Why not both, as in "have search permissions (set the `execute` bit, see [man chmod](https://linux.die.net/man/1/chmod))"

Ferroin: That works too, though I do think it's still worth mentioning that those permissions are needed on each path component.  It's something that someone who's experienced with UNIX-like systems will already know, but I regularly see it catch people by surprise who do not have lots of experience.

mfundul: Sure, I wasn't suggesting to remove the "each component" part.

collectors/diskspace.plugin/README.md
```diff
@@ -1,6 +1,11 @@
 # diskspace.plugin
 
-This plugin monitors the disk space usage of mounted disks, under Linux.
+This plugin monitors the disk space usage of mounted disks, under Linux. The plugin requires Netdata to have execute/search permissions on the mount point itself, as well as each component of the absolute path to the mount point. You can create a separate group to control access to mount points, include `netdata` user to the group, and set the execute/search permissions using
+
+```sh
+chown :volumeadm /mnt/volume
+chmod g+x /mnt/volume
+```
```
Ferroin: I'm going to be entirely honest, I do not like having the example.

For most 'normal' users, it's kind of useless as the collector will just work for them since the default permissions for a mount point in most cases will include world search permissions.  And, if somebody who doesn't know better just blindly follows it for all the mount points on their system, it will break their system.

For those users who are likely to be affected by this issue, the ideal solution is too system specific to post an example (for example, if the filesystem type of the mount supports POSIX ACLs, then adding a specific ACL to give the Netdata user search permissions is almost always going to be a better solution than the given example), and this particular example is likely to cause any of a wide variety of subtle issues if it even works to begin with (and it won't on a well secured system using a MAC system like SELinux, which is the most likely setup to cause this issue in the first place).

vlvkobal: I agree with @Ferroin. @cakrit - your turn :).

cakrit: As if I will ever disagree with @Ferroin ... Respect.

PR #6512, opened 2019-07-22
mongodb collector: take password from both pass and password field

xificurC:
##### Summary
The docs say to use "password", the code worked with "pass". To keep backwards compatibility both ways I suggest to look for both keys, with preference for "pass" since that is the current behavior.

##### Component Name
python.d mongodb

##### Additional Information
Right now using user+password logged an error message "no users authenticated", which wasn't really helpful. Working on the error handling there might be helpful :)

### DIFF
--- a/collectors/python.d.plugin/mongodb/mongodb.chart.py
+++ b/collectors/python.d.plugin/mongodb/mongodb.chart.py
@@ -426,7 +426,7 @@ class Service(SimpleService):
         self.definitions = deepcopy(CHARTS)
         self.authdb = self.configuration.get('authdb', 'admin')
         self.user = self.configuration.get('user')
-        self.password = self.configuration.get('pass')
+        self.password = self.configuration.get('pass', self.configuration.get('password'))
         self.host = self.configuration.get('host', '127.0.0.1')
         self.port = self.configuration.get('port', 27017)
         self.timeout = self.configuration.get('timeout', 100)


ilyam8:
Hey @xificurC 	

As you said it is an error in the module configuration file. I suggest to fix this error. It should be `pass`.

https://github.com/netdata/netdata/blob/aaf31295c35f790ebb3880b3f94cd787213ab98b/collectors/python.d.plugin/mongodb/mongodb.conf#L69-L72

I see it is `pass` it [module reamde](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/mongodb#configuration) too.

xificurC:
The choice is yours, I'll just say I disagree. You provided configuration examples that others might be already using (with the bug). People might have copied the file or snippet from the readme for later use. The change you propose is effectively breaking userspace. By allowing both inputs you aren't breaking anyone and since the change is so small I don't see any benefits of your solution.

Of course, this is my subjective opinion and in the end the choice is yours

ilyam8:
> The choice is yours, I'll just say I disagree. You provided configuration examples that others might be already using (with the bug). 

Then it doesnt work, so there is no `backwards compatibility`

ilyam8:
My point - if you find the error in the configuration, then fix the error in the configuration.

xificurC:
As I said, the choice is yours and I respect it. It's not exactly an earth-shattering bug, so any fix is good. I'm closing this PR since you decided for a different fix.

ilyam8:
There is no bug, it is a minor error in the config, just a typo, was added in #5707

ilyam8:
Anyway thx @xificurC, i will change the `password` to `pass` in the config

xificurC:
It doesn't matter what you need to fix, be it documentation or configuration or code. The plugin didn't work as advertised and I literally had to debug this to find the cause.

Anyway, thank you for your quick response @ilyam8 and for this great tool!

Issue #6515, opened 2019-07-22
migrate dashboard.js to React


jacekkolasa:
We are starting to migrate dashboard.js to React/Redux/Typescript stack. We want to achieve full backwards compatibility for all standard use-cases, and also maintain support for undocumented global settings, which are stored in `window.NETDATA` global object.

This is a sketch description of steps needed, but the list can be (slightly) changed during the work.

Here's how the final result should look like:

- dashboard.js file should include everything that is needed for it to work, as it was before. It needs to do the following:
  - Do all the work which is needed for backwards-compatibility reasons, that is mainly creating a global `NETDATA` object
  - create react root DOM node
  - load the react app. Normally the bundled react .js file should be placed at the end of the `body`, but `dashboard.js` needs to be placed before the `body` (to allow global `window.NETDATA` overrides) so we need to make sure the react app is ran at the right moment.
- updating to new `dashboard` should be instant. In case of bugs, we'll leave old dashboard.js in the repository under different name (`dashboard-old.js`). There will be a flag which could be toggled if for some reasons (i.e. regression bugs) user wants to use the old dashboard.
- we plan to use create-react-app (CRA) as a starting point. If it will be not enough, then we'll be working directly with `webpack`
- dashboard.js will be compiled at the build time by node.js, which should be available during agent compilation time
- available (exposed) configuration options should be listed and groomed by the team - leaving too many options is harder to maintain, so we need to find a sweet spot
- development task should be created, which should take any .html file into the dev environment and run CRA in development mode, with hot-reload
- any global state should be transformed into local/redux state, for better visibility
- any global configs should put into separate utility, so it's easy to see all the dependencies and possible configuration combinations
- react DOM node will actually not be used - but React app will use our "directives" (`<div data-netdata="unique.id"></div>`) and render the DOM inside as React Portals. This way users can still build static `html` which our `dashboard` will later enhance using multiple portals with react/redux ecosystem

To do a smooth refractor, I'm trying to do it step by step - having the app working with legacy `dashboard.js` and replacing small parts one after another. To do this, i plan to replace old-school callbacks by dispatching redux actions. Also, to make the refractor easier it's possible that some parts/components will not be structured as a fresh well-thought app. I think it's better to transform everything to React Components making sure it works, and later eventually do a subsequent refractor, because refractoring React App is much easier than refractoring legacy code to React.

# Tech choices

- Redux - for state management
- Redux-saga - we decided to use this to manage side-effects and asynchronous flow in Redux
[...]

PR #6516, opened 2019-07-22
[WiP] React dashboard

jacekkolasa:
Dashboard.js React migration
Issue: #6515 

The progress of work is stored in the [issue](#6515 )


web/gui/dashboard/src/App.tsx
```diff
@@ -0,0 +1,30 @@
...
+import Ps from 'perfect-scrollbar';
+
+import './types/global'
+import './dashboard'
+import logo from './logo.svg'
+import './App.css'
+
+window.Ps = Ps
+
+const App: React.FC = () => { // eslint-disable-line arrow-body-style
```
Azein: Are React.FC / React.SFC types still relevant?

jacekkolasa: @Azein are they not? `FC` is just `FunctionComponent`, but you can use it without specifying Props, so they default to `{}`

jacekkolasa: btw. this is create-react-app boilerplate

jacekkolasa:
@VLegakis I understand your concerns, the reason why there isn't much Typescript files is because it's just the beginning of the work - and to start the work i needed first to write some non-TS files (`dashboard-react.js`), which will make sure that existing .html files will not break. Please let me know if you have any questions

Ferroin:
Please avoid back-merging the master branch into open PR's (rebase the PR branch whenever possible, GitHub will update things correctly when you force push the rebased branch), it pulls in a bunch of people as reviewers who aren't necessary, confuses the CLA check, makes the list of changes excessive (which complicates review, you've got dozens of files changed now that have nothing to do with rewriting the dashboard), and can significantly complicate actually merging the PR.

thiagoftsm:
Hi @jacekkolasa,

To complement what @Ferroin said, check that you have various files related to MongoDB that does not have direct relationship with your PR, no less important, your PR has 153 files to review.

jacekkolasa:
@Ferroin cc: @thiagoftsm apologies, that was a mistake, normally i always do rebase, i'll try to fix this with force push

jacekkolasa:
@Ferroin it should be good now, please let me know if there will be any more issues, thanks again for quick reaction

web/gui/dashboard-react.js
```diff
@@ -0,0 +1,250 @@
...
+  fake_chart_rendering: false,    // when set to true, the dashboard will download data but will not render the charts
+
+  passive_events: null,           // true if the browser supports passive events
+
+  // the current profile
+  // we may have many...
+  current: {
+    units: 'auto',              // can be 'auto' or 'original'
+    temperature: 'celsius',     // can be 'celsius' or 'fahrenheit'
+    seconds_as_time: true,      // show seconds as DDd:HH:MM:SS ?
```
thiagoftsm: I do not know whether this variable name is reflecting correctly what it is proposed, the time is by I.S. defined as seconds, so "seconds_as_time" it is something hoped.

web/gui/dashboard/src/dashboard.js
```diff
@@ -0,0 +1,9951 @@
...
+    seconds -= (minutes * 60);
+
+    let strings = [];
+
+    if (days > 1) {
+        strings.push(days.toString() + options.space + options.days);
+    } else if (days === 1) {
+        strings.push(days.toString() + options.space + options.day);
+    }
+
```
thiagoftsm: I will write a doubt here that will be the same for the next blocks of "if" and "else if", you are working with values for "hours", "minutes" and "second" when they are bigger than 1 and equal 1, but these three variables also accepts the value 0. Why are you not working with zeros?

web/gui/dashboard/src/domains/chart/utils/seconds4human.ts
```diff
@@ -0,0 +1,80 @@
...
+  seconds -= (minutes * 60)
+
+  const strings = []
+
+  if (days > 1) {
+    strings.push(days.toString() + options.space + options.days)
+  } else if (days === 1) {
+    strings.push(days.toString() + options.space + options.day)
+  }
+
```
thiagoftsm: The same doubt that I wrote in other file can be applied here.

web/gui/dashboard-react.js
jacekkolasa: @thiagoftsm i agree it's confusing, but i think i need to explain few things :) This file (dashboard-react) is a copy-paste of some constants from the old dashboard. There are a lot of things that can be done with this, but I cannot do everything at once. So my idea is to first migrate the logic, chart creation, state handling to the new app. I'm refactoring about 60-80% of the app code, so when i have a choice to just do copy/paste (with minimal tweaks), then i'm doing it so the refractor will finish sooner. What do you think of that approach? Of course i can keep a note of your comment, this is definitely something we want to fix

web/gui/dashboard/src/dashboard.js
jacekkolasa: the zeros are not taken into account because when there are 0 days/hours/minutes/seconds we simply don't append anything to the string. for example `seconds4human(60 * 60) ==="1 hour"`

web/gui/dashboard/src/domains/chart/utils/seconds4human.ts
jacekkolasa: same as in the comment above. btw. FYI this file is the migrated version and `web/gui/dashboard/src/dashboard.js` is a file which will be removed completely after the refractor, it's just there so the app can run with both old and new code

web/gui/dashboard-react.js
ilyam8: dunno,`seconds_as_time` sounds not great, not terrible for me :smile: 	
meaning is `show seconds as time or not`

ilyam8: @thiagoftsm check the dashboard `Settings->Localte->Convert seconds to time?`

thiagoftsm: Your last comment makes more sense for me, I think it was my physicist side, instead my developer time that wrote that comment. :smile:

web/gui/dashboard/src/dashboard.js
thiagoftsm: I saw that we were not appending nothing to string, this was something that let me worried, because we will have in different times something like: "2019-08-07 00:00:00" or "2019-08-07 01:00:00" and so on.

About the day is not necessary to worry, the day 0 is far from independent of the most common 0 reference that we will get to test.

web/gui/dashboard/src/domains/chart/components/chart-container.tsx
```diff
@@ -0,0 +1,202 @@
...
+  portalNode,
+}: Props) => {
+  const chartDetails = useSelector((state: AppStateT) => selectChartDetails(
+    state, { id: chartUuid },
+  ))
+
+  const [shouldFetch, setShouldFetch] = useState<boolean>(true)
+  useInterval(() => {
+    setShouldFetch(true)
+  }, 2000)
```
Azein: 2000 could be a named constant / config member somewhere for easier control

web/gui/dashboard/src/domains/chart/components/chart-container.tsx
```diff
@@ -0,0 +1,202 @@
...
+    && (!isGlobalPanAndZoomMaster || !!globalPanAndZoom.shouldForceTimeRange)
+  const fetchDataParams = useSelector((state: AppStateT) => selectChartFetchDataParams(
+    state, { id: chartUuid },
+  ))
+
+  // don't send new requests too often (throttle)
+  // corresponds to force_update_at in old dashboard
+  // + 50 is because normal loop only happened there once per 100ms anyway..
+  const globalPanAndZoomThrottled = useThrottle(globalPanAndZoom,
+    window.NETDATA.options.current.pan_and_zoom_delay + 50)
```
Azein: I suppose you have the same concern, but still - I think that the less magic numbers the better.

web/gui/dashboard/src/domains/chart/components/chart-container.tsx
```diff
@@ -0,0 +1,202 @@
...
+  // todo take width via hook/HOC and put this into useMemo
+  const chartWidth = portalNode.getBoundingClientRect().width
+    - (hasLegend ? 140 : 0)
+
+  /**
+   * fetch data
+   */
+  const chartData = useSelector((state: AppStateT) => selectChartData(state, { id: chartUuid }))
+  const dispatch = useDispatch()
+  useEffect(() => {
```
Azein: Seems like a pretty big effect to fully comprehend. I'd be afraid to change something here =) Can it be somehow decomposed even more?

web/gui/dashboard/src/domains/chart/components/portals.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react"
+import { createPortal } from "react-dom"
+
+import { getAttributes } from "../utils/transformDataAttributes"
+import { ChartContainer } from "./chart-container"
+
+const nodesArray = Array.from(document.querySelectorAll("[data-netdata-react]"))
```
Azein: So, the node collection never changes, right?

web/gui/dashboard/src/domains/chart/reducer.ts
```diff
@@ -0,0 +1,43 @@
...
+import { fetchDataAction, updateChartDetailsAction } from "./actions"
+import { ChartState } from "./chart-types"
+
+export type StateT = {
+  [chartId: string]: ChartState
+}
+
+const initialState = {
+}
+export const initialSingleState = {
```
Azein: what is initialSingleState?

web/gui/dashboard/src/domains/chart/sagas.ts
```diff
@@ -0,0 +1,45 @@
...
+type FetchDataSaga = { payload: FetchDataPayload }
+function* fetchDataSaga({ payload }: FetchDataSaga) {
+  const {
+    // props for api
+    chart, format, points, group, gtime, options, after, before,
+    // props for the store
+    fetchDataParams, id,
+  } = payload
+  let response
+  try {
```
Azein: Maybe the try-catch thing could be moved to separate module, just not to distract from reading effects handling, as it's an independent implementation detail?

web/gui/dashboard/src/domains/chart/utils/get-portal-node-styles.ts
```diff
@@ -0,0 +1,34 @@
+import { Attributes } from "./transformDataAttributes"
+import { ChartLibraryConfig } from "./chartLibrariesSettings"
+
+type GetPortalNodeStyles = (attributes: Attributes, chartSettings: ChartLibraryConfig) => {
+  height: string | undefined,
+  width: string | undefined,
+  minWidth: string | undefined
+}
+export const getPortalNodeStyles: GetPortalNodeStyles = (attributes, chartSettings) => {
+  let width
```
Azein: It may seem as a lot of boilerplate work for a very tiny profit, yet it may be useful to abstract the let/if   procedures into immutable getters, with R.cond or something like that.

web/gui/dashboard/src/domains/global/selectors.ts
```diff
@@ -0,0 +1,40 @@
...
+import { createSelector } from "reselect"
+
+import { AppStateT } from "store/app-state"
+
+import { GetKeyArguments, getKeyForCommonColorsState } from "./reducer"
+import { storeKey } from "./constants"
+
+export const createSelectAssignedColors = (args: GetKeyArguments) => (state: AppStateT) => {
+  const keyName = getKeyForCommonColorsState(args)
+  const substate = state[storeKey].commonColorsKeys[keyName]
```
Azein: For variable-dependant selection and &&-based protection just the R.path could be used, I think.

web/gui/dashboard/src/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  <Provider store={store}>
+    <App />
+  </Provider>,
+  document.getElementById("root"),
+)
+
+// If you want your app to work offline and load faster, you can change
+// unregister() to register() below. Note this comes with some pitfalls.
+// Learn more about service workers: https://bit.ly/CRA-PWA
+serviceWorker.unregister()
```
Azein: Is the CRA service worker needed for some purpose? Otherwise we can safely remove it.

web/gui/dashboard/src/utils/createRequestAction.ts
```diff
@@ -0,0 +1,24 @@
+import { createAction } from "redux-act"
+
+// slightly simplified version of the creator used in the cloud
+// we will unify it when some typing issues will be fixed (cloud version didn't warn on bad payload)
```
Azein: I believe, now it has been improved in Cloud, so you might want to take a look. Also there are some convenient features like error-handling saga.

web/gui/dashboard/src/domains/chart/components/chart-container.tsx
jacekkolasa: thx for reminder, forgot to place `todo` here, the interval will be set based on settings from BE

web/gui/dashboard/src/domains/chart/components/chart-container.tsx
jacekkolasa: completely agree, those numbers need to be stored in different place, we just need to think if we want to expose them to the user like `window.NETDATA.options`

web/gui/dashboard/src/domains/chart/components/chart-container.tsx
jacekkolasa: thanks for input, I'll see what i can do

web/gui/dashboard/src/domains/chart/components/portals.tsx
jacekkolasa: important point. technically this will be only used in Custom Dashboards (Agent and Cloud should use react components directly) , but even there we should add an option to restart on DOM change

web/gui/dashboard/src/domains/chart/reducer.ts
jacekkolasa: bad naming, i guess :) so if this is the state in that domain:
```
export type StateT = {
  [chartId: string]: ChartState
}
```
then `initialSingleState` is initial for `ChartState` type. It is exported so even if the state is empty, selectors will return `initialSingleState`

web/gui/dashboard/src/domains/chart/utils/get-portal-node-styles.ts
jacekkolasa: good point. i was leaving code similar to old dashboard, because i thought it will be easy to compare and eventually find bugs (for example here: https://github.com/netdata/netdata/blob/6f3007537f1e4214bd5635d26f17c3d98e78afa6/web/gui/dashboard.js#L6506-L6522) but now i think most people reading new code won't even know the history of it, so it's better to write clean from the beginning

jacekkolasa:
@Azein thx for your comments. I agree that still a lot of codebase is hard to read, that can happen when:
- code is similar to old dashboard code, and is a tech debt. Not a high priority, it can be even helpful to have the code working like this during refractor, so we can faster compare old/new codebase. Example: https://github.com/netdata/netdata/blob/e02afe31e6218a31a03d20af4383b9587b76ddde/web/gui/dashboard/src/domains/chart/utils/formatters.ts - this file could be drastically improved (simplified) but refractor is IMO low priority, at least it's isolated chunk. If during refractor there will be bugs in that part, they will be easy to catch just by comparing to old codebase.
- optimizations - unfortunately we need to be very careful with that. There's a lot of `useCallback`, `useMemo`, and the way `dygraphChart` updates is complicated. I think that after refractor we will be able to better locate bottlenecks and not optimize where it's not needed
- there are of course places where code is not readable because i didn't knew a better way :) - it will be helpful if we go through codebase so you can share your observations

.eslintrc
```diff
@@ -57,7 +53,6 @@ rules:
   no-case-declarations: 2
   no-div-regex: 2
   no-else-return: 0
-  no-empty-label: 2
```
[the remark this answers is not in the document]
jacekkolasa: @VLegakis this is old .eslintrc, used globally for the whole repo. i didn't touched that, just removed few problematic rules. linter settings are overwritten for TS project in `/web/gui/dashboard/.eslintrc` file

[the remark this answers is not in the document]
VLegakis: Ok. Thanks! I missed that.

ilyam8:
@jacekkolasa do we still need this PR?

jacekkolasa:
no, thanks Ilya for reminder

PR #6520, opened 2019-07-23
netdata/packaging: Adopt netdata-updater to run properly for static64 installations.

### LINKED ISSUE #4122: No netdata-updater.sh in cron.daily after installing with kickstart-static64.sh
Hi,
I have successfully installed netdata on different Linux (SLES 12, Debian 9) with _kickstart-static64.sh_.

However, there's no file _netdata-updater.sh_ to update this installation.

Question:
How should I update netdata when installing with _kickstart-static64.sh_?

paulkatsoulakis:
##### Summary
Since we already ship netdata-updater.sh with our static64 installer, all that is left is to make sure that the updater can handle the proper update of a static installation with the latest version.

This change introduces the necessary artifacts that gets us there.
In simple words, the updater becomes aware of the installation type (static or not) and then with minor adjustments we just execute a different type of install process.

#### Component Name
netdata/packaging/installer

##### Additional Information
Fixes #4122

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -199,6 +199,7 @@ NETDATA_PREFIX=
 LIBS_ARE_HERE=0
 NETDATA_CONFIGURE_OPTIONS="${NETDATA_CONFIGURE_OPTIONS-}"
 RELEASE_CHANNEL="nightly"
+IS_NETDATA_STATIC_BINARY="no"
 while [ -n "${1}" ]; do
 	case "${1}" in
 		"--zlib-is-really-here") LIBS_ARE_HERE=1;;
@@ -207,6 +208,7 @@ while [ -n "${1}" ]; do
 		"--dont-wait") DONOTWAIT=1;;
 		"--auto-update"|"-u") AUTOUPDATE=1;;
 		"--stable-channel") RELEASE_CHANNEL="stable";;
+		"--static") IS_NETDATA_STATIC_BINARY="yes";;
 		"--enable-plugin-freeipmi")  NETDATA_CONFIGURE_OPTIONS="${NETDATA_CONFIGURE_OPTIONS//--enable-plugin-freeipmi/} --enable-plugin-freeipmi";;
 		"--disable-plugin-freeipmi") NETDATA_CONFIGURE_OPTIONS="${NETDATA_CONFIGURE_OPTIONS//--disable-plugin-freeipmi/} --disable-plugin-freeipmi";;
 		"--enable-plugin-nfacct") NETDATA_CONFIGURE_OPTIONS="${NETDATA_CONFIGURE_OPTIONS//--enable-plugin-nfacct/} --enable-plugin-nfacct";;
@@ -1073,6 +1075,7 @@ INSTALL_UID="${UID}"
 NETDATA_GROUP="${NETDATA_GROUP}"
 REINSTALL_COMMAND="${REINSTALL_COMMAND}"
 RELEASE_CHANNEL="${RELEASE_CHANNEL}"
+IS_NETDATA_STATIC_BINARY="${IS_NETDATA_STATIC_BINARY}"
 # This value is meant to be populated by autoupdater (if enabled)
 NETDATA_TARBALL_CHECKSUM="new_installation"
 EOF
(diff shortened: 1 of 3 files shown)


thiagoftsm:
Hi @paulkatsoulakis ,

I tested your PR today and I notice things that was different that I hoped, let me describe the tests for you check:

 - I installed a VM from zero to be sure that I would not have Netdata installed.
- After this I installed Netdata using "./netdata-installer.sh --static" and I saw that it did not create a static binary at /usr/sbin/netdata , it was created a normal binary linked with the libraries. I executed this test, because I saw that you changed the "netdata-installer.sh".
- The next step I used "kickstart-static64.sh" and it was installed correctly inside /opt/netdata, but it initialized the netdata that was present at /usr/sbin/ instead the netdata that were installed by kickstart-static64.sh at /opt/netdata/bin/srv/netdata . Considering that we know where netdata is installed by the kickstart-static64.sh, I think it needed to call the file using the whole path. This is not critical, but it was something not expected.
- Finally I used netdata-updater.sh that was inside /opt/netdata/usr/libexec, the script got the tar.gz file and it compiled it at /tmp, after this it updated everything in the directory /opt/netdata/, but when I decided to check the file, I saw that instead to have a static binary /opt/netdata/bin/srv/netdata, it became a normal binary linked. 

I cannot say that it was not successful, because at the end I am having the Netdata installed, but when I began to do the tests, I did not hope that after an update I would have a binary linked instead another static.

paulkatsoulakis:
> Hi @paulkatsoulakis ,
> 
> I tested your PR today and I notice things that was different that I hoped, let me describe the tests for you check:
> 
> * I installed a VM from zero to be sure that I would not have Netdata installed.
> * After this I installed Netdata using "./netdata-installer.sh --static" and I saw that it did not create a static binary at /usr/sbin/netdata , it was created a normal binary linked with the libraries. I executed this test, because I saw that you changed the "netdata-installer.sh".
> * The next step I used "kickstart-static64.sh" and it was installed correctly inside /opt/netdata, but it initialized the netdata that was present at /usr/sbin/ instead the netdata that were installed by kickstart-static64.sh at /opt/netdata/bin/srv/netdata . Considering that we know where netdata is installed by the kickstart-static64.sh, I think it needed to call the file using the whole path. This is not critical, but it was something not expected.

In principal we do not allow multiple installations of netdata, thus we can't support much in such peculiar situation.

In this case what happened is the installer attempted to first start netdata using command `netdata` which of course works because on the standard paths there is a netdata binary. So it did not proceed to attempt anything else that would result in finding the right binary.

I think in the peculiar case the user installs both netdata, we should remain with the one in sbin, so we did right to start that one.

> * Finally I used netdata-updater.sh that was inside /opt/netdata/usr/libexec, the script got the tar.gz file and it compiled it at /tmp, after this it updated everything in the directory /opt/netdata/, but when I decided to check the file, I saw that instead to have a static binary /opt/netdata/bin/srv/netdata, it became a normal binary linked.
> 
If you run kickstart-static64.sh, it brings you what is already deployed on our google storage.
In order to test the static installation you need to first compile the static binary and run it directly.
My apologies for misguiding you on this one.  To prepare the static binary you need to run `./packaging/makeself/build-x86_64-static.sh` and then execute the generated .run that will give you the right installer version.

> I cannot say that it was not successful, because at the end I am having the Netdata installed, but when I began to do the tests, I did not hope that after an update I would have a binary linked instead another static.

thiagoftsm:
Hi @paulkatsoulakis ,

I had success in al the process of installation and update.  
This time the option "--static" was not available to call from netdata-installer.sh as we talked yesterday. 
A last doubt, I ran 7 times since yesterday the compilation and update to test this PR, and now I have in my /tmp directory various directories with the name netdata-kickstart-* and netdata-updater-* , this is happening because Slackware does not clean by default the /tmp directory after a reboot. Is it not a good idea to remove these directories at the end of an update or installer process?

paulkatsoulakis:
> Hi @paulkatsoulakis ,
> 
> I had success in al the process of installation and update.
> This time the option "--static" was not available to call from netdata-installer.sh as we talked yesterday.
> A last doubt, I ran 7 times since yesterday the compilation and update to test this PR, and now I have in my /tmp directory various directories with the name netdata-kickstart-* and netdata-updater-* , this is happening because Slackware does not clean by default the /tmp directory after a reboot. Is it not a good idea to remove these directories at the end of an update or installer process?

Great catch, you 're absolutely right, fixed!

PR #6524, opened 2019-07-23
Add MongoDB backend

### LINKED ISSUE #4096: Archive metrics to mongoDB
How to get mongoDB backend for netdata?

I tried with the following setup and was unable to see any data stream/tables etc into mongoDB

```
[backend]
        enabled = yes
        type = json
        destination = localhost:27017
        data source = average
        prefix = netdata
        update every = 10
        buffer on failures = 10
        timeout ms = 20000
        send charts matching = *
        send names instead of ids = yes
```

vlvkobal:
##### Summary
Support archiving metrics to the MongoDB database.

Closes #4096

##### Component Name
backends


thiagoftsm:
Hi @vlvkobal ,

I saw that you are starting the development, but I decided to test to see what was happening until now, and I could not compile the code on  Slackware current and I think I will have the same problem on Debian stable and Centos 7 due the same motive, in all these distribution we do not have the file "<mongoc/mongoc.h>", instead we have "libmongoc-1.0/mongoc.h". For Debian you can check more details in this link ( https://packages.debian.org/search?searchon=contents&keywords=mongoc.h&mode=path&suite=stable&arch=any ) and for Centos ( https://centos.pkgs.org/7/epel-x86_64/mongo-c-driver-devel-1.3.6-1.el7.x86_64.rpm.html )

vlvkobal:
Does `/usr/include/libmongoc-1.0/mongoc/mongoc.h` exist on Slackware?

cakrit:
I wouldn't expect a library like this to exist everywhere. Perhaps consider redistributing it and adding an entry to https://github.com/netdata/netdata/blob/master/REDISTRIBUTED.md?

thiagoftsm:
> Does `/usr/include/libmongoc-1.0/mongoc/mongoc.h` exist on Slackware?

No, on Slackware we do not have /usr/include/libmongoc-1.0/mongoc.h like Centos 7, but Slackware is not a distribution used everywhere, on the other hand Centos can be a problem.

thiagoftsm:
Hi @vlvkobal ,

I tested now and I saw that the problem was fixed, I could compile the netdata with your last commit.

backends/backends.c
```diff
@@ -547,6 +589,18 @@ void *backends_main(void *ptr) {
...
+        case BACKEND_TYPE_MONGODB: {
+#if HAVE_MONGOC
+            if(!mongodb_init(mongodb_uri, mongodb_database, mongodb_collection))
+                do_mongodb = 1;
+            else
+                error("BACKEND: cannot initialize MongoDB backend");
+#else
+            error("BACKEND: MongoDB support isn't compiled");
+#endif // HAVE_MONGOC
+            backend_set_mongodb_variables(&default_port,&backend_response_checker,&backend_request_formatter);
```
thiagoftsm: You missed spaces in the  backend_set_mongodb_variables.

backends/backends.c
```diff
@@ -833,12 +897,53 @@ void *backends_main(void *ptr) {
...
+                const char *first_char = buffer_tostring(b);
+
+                debug(D_BACKEND, "BACKEND: mongodb_insert(): uri = %s, database = %s, collection = %s, \
+                      buffer = %zu", mongodb_uri, mongodb_database, mongodb_collection, buffer_len);
+
+                if(!mongodb_insert(first_char)) {
+                    sent += buffer_len;
+                    chart_transmission_successes++;
+                    chart_receptions++;
+                }
```
thiagoftsm: I personally like the else bellow '}', but this is a different pattern from the rest of the file.

backends/mongodb/mongodb.c
```diff
@@ -0,0 +1,207 @@
...
+                      "\"chart_context\": \"%s\","
+                      "\"chart_type\":\"%s\","
+                      "\"units\": \"%s\","
+
+                      "\"id\":\"%s\","
+                      "\"name\":\"%s\","
+                      "\"value\":" COLLECTED_NUMBER_FORMAT ","
+
+                      "\"timestamp\": %llu}",
+            prefix,
```
thiagoftsm: The second argument of the function buffer_sprint has a different alignment of the others variables.

backends/mongodb/mongodb.c
```diff
@@ -0,0 +1,207 @@
...
+                          "\"chart_context\": \"%s\","
+                          "\"chart_type\":\"%s\","
+                          "\"units\": \"%s\","
+
+                          "\"id\":\"%s\","
+                          "\"name\":\"%s\","
+                          "\"value\":" CALCULATED_NUMBER_FORMAT ","
+
+                          "\"timestamp\": %llu}",
+                prefix,
```
thiagoftsm: The same thing about alignment happens here, at least in github they are having different alignments .

backends/backends.c
vlvkobal: There are a lot of places in the file where else is below '}'.

backends/backends.c
```diff
@@ -300,6 +300,35 @@ void backend_set_prometheus_variables(int *default_port,
...
 
+/**
+ * Set MongoDB variables
+ *
+ * Set the variables necessaries to work with this specific backend.
+ *
+ * @param default_port  the default port of the backend
+ * @param brc function called to check the result.
+ * @param brf function called to format the msessage to the backend
+ * @param type the backend string selector.
```
thiagoftsm: I think you forgot to remove the comment for "type" or you did not create this parameter yet.

backends/backends.c
```diff
@@ -547,6 +589,18 @@ void *backends_main(void *ptr) {
...
+        case BACKEND_TYPE_MONGODB: {
+#if HAVE_MONGOC
+            if(!mongodb_init(mongodb_uri, mongodb_database, mongodb_collection))
+                do_mongodb = 1;
+            else
+                error("BACKEND: cannot initialize MongoDB backend");
+#else
+            error("BACKEND: MongoDB support isn't compiled");
+#endif // HAVE_MONGOC
+            backend_set_mongodb_variables(&default_port, &backend_response_checker, &backend_request_formatter);
```
thiagoftsm: Considering what we have inside this function, in my opinion you need to call it inside the if of the line 594, because  we only need to set the Mongodb variables case we have it. 
The code that you wrote is not wrong, because it will not set the variables, but I do not think that the compiler will avoid to move these 3 variables to a register and call the function, so we will save some opcodes. No less important, after to see where this function is called, I think we can put the whole function "backend_set_mongodb_variables" inside an "#if HAVE_MONGOC", because I cannot see it been useful for other backends.

thiagoftsm:
Hi @vlvkobal ,

We will need to take an important decision about the MongoDB versions that we will support. When I compiled with an old version of mongo-c-driver (1.6.1) I got the following errors:

backends/mongodb/mongodb.c: In function 'mongodb_init':
backends/mongodb/mongodb.c:16:11: warning: implicit declaration of function 'mongoc_uri_new_with_error'; did you mean 'mongoc_gridfs_file_error'? [-Wimplicit-function-declaration]
   16 |     uri = mongoc_uri_new_with_error (uri_string, &error);
      |           ^~~~~~~~~~~~~~~~~~~~~~~~~
      |           mongoc_gridfs_file_error
backends/mongodb/mongodb.c:16:9: warning: assignment to 'mongoc_uri_t *' {aka 'struct _mongoc_uri_t *'} from 'int' makes pointer from integer without a cast [-Wint-conversion]
   16 |     uri = mongoc_uri_new_with_error (uri_string, &error);
      |         ^
backends/mongodb/mongodb.c:22:41: error: 'MONGOC_URI_SOCKETTIMEOUTMS' undeclared (first use in this function); did you mean 'MONGOC_DEFAULT_SOCKETTIMEOUTMS'?
   22 |     mongoc_uri_set_option_as_int32(uri, MONGOC_URI_SOCKETTIMEOUTMS, 9000); // TODO: use variable for timeout
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                         MONGOC_DEFAULT_SOCKETTIMEOUTMS
backends/mongodb/mongodb.c:22:41: note: each undeclared identifier is reported only once for each function it appears in
  CCLD     cups.plugin
backends/mongodb/mongodb.c: In function 'mongodb_insert':
backends/mongodb/mongodb.c:64:14: warning: implicit declaration of function 'mongoc_collection_insert_one'; did you mean 'mongoc_collection_insert_bulk'? [-Wimplicit-function-declaration]
   64 |         if (!mongoc_collection_insert_one(collection, insert, NULL, NULL, &error)) {
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |              mongoc_collection_insert_bulk

After to update the lib to the version 1.14.0 everything compiled without any problem.
We will need to decide between to create codes with the old api or to document that we will only support mongo-c-driver from the version that they released these new functions. For the first case, it will be necessary to have "#if" to select specific codes between versions.

backends/mongodb/mongodb.c
```diff
@@ -78,3 +80,93 @@ void mongodb_cleanup() {
...
+        }
+        else if(name[0] == 'c' && !strcmp(name, "collection")) {
+            collection = strdupz(value);
+        }
+    }
+
+    fclose(fp);
+
+    if(unlikely(!collection || !*collection)) {
+        error("BACKEND: collection name is a mandatory MongoDB parameter but it is not configured");
```
thiagoftsm: I know this is more @joelhans stuff, but only to confirm, is it not necessary to put a "," before "but"?

backends/mongodb/mongodb.c
```diff
@@ -0,0 +1,186 @@
...
+       return 1;
+    }
+
+    free_bson(insert, n_documents);
+
+    return 0;
+}
+
+void mongodb_cleanup() {
+    mongoc_collection_destroy (mongodb_collection);
```
thiagoftsm: For the majority of the functions in this PR, you are using the format:

function_name(argument1, argument2);

but, for mongoc functions it looks like you are using a new pattern:

mongoc_function (argument);

It is not for all, as you can see in the 81 free_bson does not have a space between function name and argument, but in this function and others bson you are adding a space, that is not wrong, but it is not the pattern.

backends/mongodb/mongodb.c
```diff
@@ -0,0 +1,186 @@
...
+
+    mongoc_init();
+
+    uri = mongoc_uri_new_with_error(uri_string, &error);
+    if(unlikely(!uri)) {
+       error("BACKEND: failed to parse URI: %s. Error message: %s", uri_string, error.message);
+       return 1;
+    }
+
+    mongoc_uri_set_option_as_int32(uri, MONGOC_URI_SOCKETTIMEOUTMS, 9000); // TODO: use variable for timeout
```
thiagoftsm: This is a bool function, is it not a good idea at least to show an info message case it returns a false value?

backends/mongodb/mongodb.c
```diff
@@ -0,0 +1,186 @@
...
+
+    mongodb_client = mongoc_client_new_from_uri(uri);
+    if(unlikely(!mongodb_client)) {
+       return 1;
+    }
+
+    mongoc_client_set_appname(mongodb_client, "netdata");
```
thiagoftsm: The same question that I wrote in the line 24 can be applied here, because it is also bool.

backends/mongodb/mongodb.c
```diff
@@ -0,0 +1,186 @@
...
+
+    mongodb_collection = mongoc_client_get_collection(mongodb_client, database_string, collection_string);
```
thiagoftsm: It is necessary to have a test like line 27 here to check the return of the pointer. I think here an info is not enough, we need to write an error  message, because it can means operate system without resources.

backends/mongodb/mongodb.c
```diff
@@ -0,0 +1,186 @@
...
+    if(unlikely(!uri)) {
+       error("BACKEND: failed to parse URI: %s. Error message: %s", uri_string, error.message);
+       return 1;
+    }
+
+    mongoc_uri_set_option_as_int32(uri, MONGOC_URI_SOCKETTIMEOUTMS, 9000); // TODO: use variable for timeout
+
+    mongodb_client = mongoc_client_new_from_uri(uri);
+    if(unlikely(!mongodb_client)) {
+       return 1;
```
thiagoftsm: Please, add an error message here due the same motive explained at 31.

backends/backends.c
vlvkobal: Done

backends/mongodb/mongodb.c
vlvkobal: It can't return an error with this macros as a parameter, but I've added a condition with an error message for consistency.

backends/mongodb/mongodb.c
vlvkobal: `mongodb_collection` is checked in `mongoc_collection_insert_many()`.

backends/mongodb/mongodb.c
thiagoftsm: All right, OpenSSL has something like this too. Thanks!

thiagoftsm:
I decided to do more one round of tests with the options listed in the link http://mongoc.org/libmongoc/current/mongoc_uri_t.html#ssl-options that @vlvkobal gave me. For this I used the URI:

mongodb://netdata:netdata@localhost:27017/?authSource=netdata&ssl=true&allowinvalidcertificates=true&sslclientcertificatekeyfile=/etc/netdata/mongodb.pem

And I received the error:

ilyam8:
> mongodb://netdata:netdata@localhost:27017/?authSource=netdata&ssl=true&allowinvalidcertificates=true&sslclientcertificatekeyfile=/etc/netdata/mongodb.pem

http://mongoc.org/libmongoc/current/mongoc_ssl_opt_t.html#server-certificate-verification

vlvkobal:
I ran a mongodb container:
```
docker run --name mongo-ssl -p 27017:27017 -v /etc/mongodb.pem:/etc/mongodb.pem mongo --sslMode requireSSL --sslPEMKeyFile /etc/mongodb.pem
```
Configured the mongodb backend:
```
uri = mongodb://localhost:27017/?ssl=true&sslclientcertificatekeyfile=/etc/mongodb.pem&sslallowinvalidcertificates=true
database = netdata
collection = backend
~

```

Inside the container:
```
root@41f9e64e10d5:/# mongo --ssl --sslPEMKeyFile /etc/mongodb.pem --sslAllowInvalidCertificates
...
> show databases
admin    0.000GB
config   0.000GB
local    0.000GB
netdata  0.000GB
> use netdata
switched to db netdata
> db.backend.count()
12864
> db.backend.find()
{ "_id" : ObjectId("5d42ac7414491b6cbe205662"), "prefix" : "netdata", "hostname" : "arch-esxi", "chart_id" : "netdata.plugin_tc_time", "chart_name" : "netdata.plugin_tc_time", "chart_family" : "tc.helper", "chart_context" : "netdata.plugin_tc_time", "chart_type" : "netdata", "units" : "milliseconds/run", "id" : "run_time", "name" : "run time", "value" : 11, "timestamp" : 1564650609 }
...
```
Thus, it works with SSL enabled.

When SSL is disabled on the netdata side (`uri = mongodb://localhost:27017`), I get
```
2019-08-01 12:18:52: netdata ERROR : BACKENDS : BACKEND: No suitable servers found (`serverSelectionTryOnce` set): [Server closed connection. calling ismaster on 'localhost:27017']
2019-08-01 12:18:52: netdata ERROR : BACKENDS : BACKEND: failed to write data to database backend 'mongodb://localhost:27017'. Willing to write 747728 bytes, wrote 0 bytes.
```

backends/backends.c
```diff
@@ -450,7 +475,18 @@ void *backends_main(void *ptr) {
 
 #if ENABLE_PROMETHEUS_REMOTE_WRITE
     int do_prometheus_remote_write = 0;
-    BUFFER *http_request_header = buffer_create(1);
+    BUFFER *http_request_header = NULL;
```
paulkatsoulakis: This also looks like a prometheus related change, could you clarify why changed to null?

backends/backends.c
```diff
@@ -547,6 +584,30 @@ void *backends_main(void *ptr) {
...
+            if(likely(!mongodb_init(mongodb_uri, mongodb_database, mongodb_collection, mongodb_socket_timeout))) {
+                backend_set_mongodb_variables(&default_port, &backend_response_checker, &backend_request_formatter);
+                do_mongodb = 1;
+            }
+            else {
+                error("BACKEND: cannot initialize MongoDB backend");
+                goto cleanup;
+            }
+#else
+            error("BACKEND: MongoDB support isn't compiled");
```
paulkatsoulakis: error or just warning/info?

backends/backends.c
```diff
@@ -658,7 +719,8 @@ void *backends_main(void *ptr) {
         size_t count_dims_total = 0;
 
 #if ENABLE_PROMETHEUS_REMOTE_WRITE
-        clear_write_request();
+        if(do_prometheus_remote_write)
```
paulkatsoulakis: isn't a bit too much to have both a condition and the if-def that kills the code if prometheus not available?

paulkatsoulakis:
Also on a general comment, can mongodb be enabled without being configured without causing too much noise? We 're gonna have to make it always enabled for .RPM and .DEB packages

backends/backends.c
vlvkobal: No, because ENABLE_PROMETHEUS_REMOTE_WRITE can be defined when the remote write backend is not enabled.

backends/backends.c
vlvkobal: It's an error because it will be displayed when the user set
```
[backend]
	enabled = yes
	type = mongodb
```

backends/backends.c
```diff
@@ -524,6 +560,7 @@ void *backends_main(void *ptr) {
 #if ENABLE_PROMETHEUS_REMOTE_WRITE
             do_prometheus_remote_write = 1;
 
+            http_request_header = buffer_create(1);
```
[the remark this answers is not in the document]
vlvkobal: It was intentionally moved so that the `http_request_header` is initialized only when the remote write backend is enabled.

backends/backends.c
vlvkobal: The same reason, as in the previous comment.

vlvkobal:
> can mongodb be enabled without being configured without causing too much noise?

Do you mean
> can mongodb be enabled without causing too much noise?

The noise is usual. There is no more of it than for any other backend.

backends/backends.c
paulkatsoulakis: I am not following, do_prometheus_remote_write=1 is set once at some point, with only obvious condition ENABLE_PROMETHEUS_REMOTE_WRITE to be set

cosmix:
I'd like to review this please before it's approved.

thiagoftsm:
> I'd like to review this please before it's approved.

I added you @cosmix  as a reviewer, we could not identify errors with Mongo 4, but I had problems with Mongo 3.4.x .

configure.ac
```diff
@@ -980,7 +1013,7 @@ if test "${enable_lto}" != "no"; then
 fi
 if test "${have_lto}" = "yes"; then
     oCFLAGS="${CFLAGS}"
-    CFLAGS="${CFLAGS} -flto ${OPTIONAL_MATH_CFLAGS} ${OPTIONAL_NFACCT_CFLAGS} ${OPTIONAL_ZLIB_CFLAGS} ${OPTIONAL_UUID_CFLAGS} ${OPTIONAL_LIBCAP_CFLAGS} ${OPTIONAL_IPMIMONITORING_CFLAGS} ${OPTIONAL_CUPS_CFLAGS} ${OPTIONAL_XENSTAT_FLAGS} ${OPTIONAL_KINESIS_CFLAGS} ${OPTIONAL_PROMETHEUS_REMOTE_WRITE_CFLAGS}"
```
paulkatsoulakis: Nice, i was wondering why we add all these flags with LTO enabled only.
Have you confirmed that this wasn't intentional for some reason ?

backends/backends.c
vlvkobal: `ENABLE_PROMETHEUS_REMOTE_WRITE` is defined by autoconf if all required libraries are present in the system. `do_prometheus_remote_write` is set to 1 if `ENABLE_PROMETHEUS_REMOTE_WRITE` is defined AND the user enabled the prometheus remote write backend in the configuration.

configure.ac
vlvkobal: No, I didn't ask anyone. I couldn't find the reason for this to be intentional. I think it was done by mistake.

thiagoftsm:
Updating my previous comments, I retest now the PR with Debian 9.9, MongoDB 3.2.11 and libmongo-c 1.14.0 and everything worked fine, so the problem with the version 3.4.1 was probably with the machine that I tested.

thiagoftsm:
I did all the testes again with mongodb 3.4.2 and and libmongoc-1.14.0, and everything was ok, so I had a problem with my initial installation. 
So I can confirm that I had success using MongoDB 3.2, 3.4 and 4.0 .

backends/mongodb/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Configuration
+
+To enable data sending to the mongodb backend set the following options in `netdata.conf`:
+```
+[backend]
+    enabled = yes
+    type = mongodb
+```
+
+In the netdata configuration directory run `./edit-config mongodb.conf` and set [MongoDB URI](https://docs.mongodb.com/manual/reference/connection-string/), database name, and collection name:
```
joelhans: Please capitalize Netdata here!

```suggestion
In the Netdata configuration directory run `./edit-config mongodb.conf` and set [MongoDB URI](https://docs.mongodb.com/manual/reference/connection-string/), database name, and collection name:
```

backends/mongodb/README.md
```diff
@@ -0,0 +1,31 @@
+# MongoDB backend
+
+## Prerequisites
+
+To use MongoDB as a backend `libmongoc` 1.7.0 or higher should be [installed](http://mongoc.org/libmongoc/current/installing.html) first. Next, Netdata should be re-installed from the source. The installer will detect that the required libraries are now available.
```
joelhans: Add a comma after "backend":

```suggestion
To use MongoDB as a backend, `libmongoc` 1.7.0 or higher should be [installed](http://mongoc.org/libmongoc/current/installing.html) first. Next, Netdata should be re-installed from the source. The installer will detect that the required libraries are now available.
```

backends/mongodb/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Configuration
+
+To enable data sending to the mongodb backend set the following options in `netdata.conf`:
```
joelhans: I think you should capitalize MongoDB here, as you're not referencing a process/user/etc.

```suggestion
To enable data sending to the MongoDB backend set the following options in `netdata.conf`:
```

vlvkobal:
I'm not changing CODEOWNERS. It's a rebase because of #6626. Actually, it's a one-line change.

thiagoftsm:
> I'm not changing CODEOWNERS. It's a rebase because of #6626. Actually, it's a one-line change.

all right, thanks by the explanation.

backends/mongodb/mongodb.c
```diff
@@ -0,0 +1,193 @@
...
+
+    mongoc_init();
+
+    uri = mongoc_uri_new_with_error(uri_string, &error);
+    if(unlikely(!uri)) {
+        error("BACKEND: failed to parse URI: %s. Error message: %s", uri_string, error.message);
+        return 1;
+    }
+
+    socket_timeout = mongoc_uri_get_option_as_int32(uri, MONGOC_URI_SOCKETTIMEOUTMS, socket_timeout);
```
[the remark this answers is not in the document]
vlvkobal: Yes, it is a fallback value. I think we don't need to rename the variable because 99,99% of the time it will be the actual value of the socket timeout.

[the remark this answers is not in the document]
cosmix: The problem is that this is not reflected in the naming (both variable and configuration) and when someone sets the timeout value to some value, they will expect it to be applied. When it isn't, they will — very justifiably — be upset. 

So, in my opinion the variable should, at the very least, be clear that it's a fallback value. But there's a more fundamental 'design' issue here: typically, passed parameters are _overriding_ pre-existing values, not used as fallbacks. 

Even though it's a relatively minor point from a technical standpoint, it is quite important in terms of user experience: the implementation, as it is today, can be confusing, both to developers and end-users.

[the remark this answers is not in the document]
vlvkobal: > The problem is that this is not reflected in the naming (both variable and configuration) 

Hm... I don't understand. What "configuration" are you talking about? The variable which is passed to `mongoc_uri_get_option_as_int32()` is internal and does not represent any configuration option.

> when someone sets the timeout value to some value, they will expect it to be applied

I will be applied. If the user sets the socket timeout using MongoDB URI, it overrides the internal variable.

> But there's a more fundamental 'design' issue here: typically, passed parameters are overriding pre-existing values, not used as fallbacks.

It's the design of the `libmongoc` library. We also have the same design for the `config_get` and `config_set` functions in Netdata.

> the implementation, as it is today, can be confusing, both to developers and end-users

I see that the variable name can be confusing to developers. I'll change its name. But how it could be confusing to end-users?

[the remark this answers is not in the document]
cosmix: > It's the design of the libmongoc library. We also have the same design for the config_get and config_set functions in Netdata.

Not at all. `libmongoc` doesn't require you to "get and set" a value in the way it's done. What you're doing is _correct_, and your thinking is absolutely valid (i.e. allow the user to set a timeout in the connection URI, then, failing that, use a sane default), but — without documentation — the implementation can confuse people as it is.

Plus we might have wanted to expose the timeout in the configuration file at some point, instead of calculating using the `global_backend_update_every`.

Anyway, I think renaming the variable works for me as this is turning into more of a theoretical discussion about design and conventions than a CR :)

PR #6538, opened 2019-07-24
[ci skip] minor/vanity - add self.name() to contrib.md

underhood:


cakrit:
@paulkatsoulakis don't we automatically generate this? If not, put an issue on your backlog to do it...

PR #6539, opened 2019-07-24
minor - code readability HTTP CODES as defines && clear warnings

underhood:
##### Summary
Just a minor code cleanup for better readability of code that I didn't want to include in UTF8 URL parser PR due to IMHO being unrelated

##### Component Name
libnetdata/server

##### Additional Info
No behavior change. Just code readability - something that annoyed me a bit when I was going trough code. *(not all people keep all the codes in memory (besides 200/400/404))*
Additionally one compiler warning fixed if compiling with `-Wpointer-sign`

### DIFF
--- a/web/server/web_client.c
+++ b/web/server/web_client.c
@@ -16,8 +16,8 @@ inline int web_client_permission_denied(struct web_client *w) {
     w->response.data->contenttype = CT_TEXT_PLAIN;
     buffer_flush(w->response.data);
     buffer_strcat(w->response.data, "You are not allowed to access this resource.");
-    w->response.code = 403;
-    return 403;
+    w->response.code = HTTP_RESP_FORBIDDEN;
+    return HTTP_RESP_FORBIDDEN;
 }
 
 static inline int web_client_crock_socket(struct web_client *w) {
@@ -337,7 +337,7 @@ static inline int access_to_file_is_not_permitted(struct web_client *w, const ch
     w->response.data->contenttype = CT_TEXT_HTML;
     buffer_strcat(w->response.data, "Access to file is not permitted: ");
     buffer_strcat_htmlescape(w->response.data, filename);
-    return 403;
+    return HTTP_RESP_FORBIDDEN;
 }
 
 int mysendfile(struct web_client *w, char *filename) {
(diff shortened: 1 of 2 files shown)


thiagoftsm:
Hi @underhood ,

I liked what the PR you brought, without any doubt, the code will be more readable with these changes, but before we approve, I will ask you to change one thing that we missed in the previous PR, because I had a problem in my compiler and I could not detect it. As you can see below

/home/thiago/Netdata/netdata/libnetdata/url/url.c: In function ‘url_decode_r’:
/home/thiago/Netdata/netdata/libnetdata/url/url.c:211:29: warning: pointer targets in passing argument 1 of ‘utf8_check’ differ in signedness [-Wpointer-sign]
  211 |     if(unlikely( utf8_check(to) )) //NULL means sucess here
      |                             ^~
      |                             |
      |                             char *
/home/thiago/Netdata/netdata/./config.h:356:41: note: in definition of macro ‘unlikely’
  356 | #define unlikely(x) __builtin_expect(!!(x), 0)
/home/thiago/Netdata/netdata/libnetdata/url/url.c:126:42: note: expected ‘unsigned char *’ but argument is of type ‘char *’
  126 | unsigned char *utf8_check(unsigned char *s)
      |                           ~~~~~~~~~~~~~~~^

It is necessary to cast the variable when we call utf8_check, but for these new code that you are bringing  there is nothing to change.

Thanks a lot!

libnetdata/url/url.c
```diff
@@ -208,7 +208,7 @@ char *url_decode_r(char *to, char *url, size_t size) {
 
     *d = '\0';
 
-    if(unlikely( utf8_check(to) )) //NULL means sucess here
+    if(unlikely( utf8_check((unsigned char *)to) )) //NULL means sucess here
```
ilyam8: > minor - code readability HTTP CODES as defines 
> No behavior change. Just code readability

Why do not add it as a separate PR? Or change the title/OP.

underhood: did it on request of @thiagoftsm

ilyam8: I know, but now the PR changes dont reflect title/OP.

underhood: OK, so do you want me to kick that commit out and make it separate PR *(cleaner, or let it be part of Url parser refactoring #6247)* or just change a title here

ilyam8: I think better to split not related things. It feels natural for me. However title/op should always reflect changes.

This was a comment, so just keep this in mind :+1:

underhood:
> 
> 
> Damn, we just got a conflict due to a different PR getting merged. Can you please rebase?

Yep, the URL rework PR also fixed the comp warning in url.c. Will do when i get home

underhood:
I kicked out that url.c change

PR #6549, opened 2019-07-26
Add configurable default locations for trusted CA certificates

### LINKED ISSUE #6529: Valid let's encrypt certifcate considerated as invalid by streaming client
##### Bug report summary

A valid let's encrypt certificate is considered as invalid for reason 20 X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY by streaming client even if openssl _sclient return 0 in verify

##### OS / Environment

ubuntu 18.04

netdata v1.16.0-70-nightly

##### Component Name

streaming

##### Steps To Reproduce

- Create a let's encrypt certificate with proper hostname
- Configure the master with full (chain+cert) file as certificate and key as ssl key
- Configure a slave to use SSL for streaming with tcp:hostname:port:SSL without skipping certificate validation
- watch the stream fail in error.log with SSL RFC4158 check:  We have a invalid certificate, the tests result with 20 and message error:00000000:lib(0):func(0):reason(0)

Whereas openssl s_client works fine
```
echo Q | openssl s_client -connect <REDACTED_HOST>:PORT | grep -i verify
depth=2 O = Digital Signature Trust Co., CN = DST Root CA X3
verify return:1
depth=1 C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
verify return:1
depth=0 CN = <REDACTED HOST>
verify return:1
DONE
    Verify return code: 0 (ok)
```

##### Expected behavior

[send to tcp:host:port]: established communication - ready to send metrics...

thiagoftsm:
##### Summary
Self-signed certificates and certificates created with letsencrypt can be rejected by OpenSSL, due safety reason or to be signed by a not trusted Certification entity. This PR brings two new options for the stream.conf that allows the slave to connect with the master using these two kind of certificates without report an error.
##### Component Name
stream.
##### Additional Information
Fixes #6529 
To test this PR, it is necessary the following steps:

1 - To create a self-signed certificate or a trusted certificate with letsencrypt.
2 - To add the certificate to the trusted certificate list(The README.md teaches this steps)
3 - To enable options of stream.conf according README.md.


streaming/README.md
```diff
@@ -236,6 +236,44 @@ When TLS/SSL is enabled on the slave, the default behavior will be to not connec
     ssl skip certificate verification = yes
 ```
 
+#### Trusted Certificate
```
cakrit: This needs some grammar fixes @joelhans but, more importantly, it's not the master certificate that we need to bring to the slave, since it gets it from the master already. 

What we do here  is specify the locations of Certificate Authority (CA) certificates, necessary to verify the master certificate. In other words, we configure the slave so that it can validate the certificate chain. 

The parameters in the config file should be called 
SSL_CTX_load_verify_locations CApath
SSL_CTX_load_verify_locations CAfile

If we do that, we don't need to explain much about what goes in those two parameters, just add a link to 
 https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_load_verify_locations.html

thiagoftsm: I changed the variable names in the stream.conf and I brought to the documentation the latest version of the OpenSSL link instead the previous that you gave us.
With the changes in the text that I executed, I hope I could remove the ideat that the option allows we use only the Netdata master certificate, the correct is to say that it is possible to add the certificate to the Trusted list and instead to search in the whole list of Trusted certificates, we can work directly with the netdata master.
In the last PR I also changed the plugin_d.c, because when the Netdata is compiled without SSL it was giving a warning. Considering that this was happening due the previous PR that we merged to fix the other SSL problems, I changed the variable here, because it has relationship with the stream and SSL.

streaming/rrdpush.c
```diff
@@ -90,13 +95,20 @@ int rrdpush_init() {
             }
         }
     }
+
     char *invalid_certificate = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "ssl skip certificate verification", "no");
     if ( !strcmp(invalid_certificate,"yes")){
         if (netdata_validate_server == NETDATA_SSL_VALID_CERTIFICATE){
-            info("The Netdata is configured to accept invalid certificate.");
+            info("The Netdata is configured to accept invalid SSL certificate.");
```
cakrit: The Netdata -> Netdata

streaming/rrdpush.c
```diff
@@ -90,13 +95,20 @@ int rrdpush_init() {
...
             netdata_validate_server = NETDATA_SSL_INVALID_CERTIFICATE;
         }
     }
+
+    netdata_ssl_cert_directory = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "CApath", "/etc/ssl/certs/");
+    netdata_ssl_master_cert = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "CAfile", "/etc/ssl/certs/");
+    if(!strcmp(netdata_ssl_cert_directory,netdata_ssl_master_cert)) {
+        info("Netdata does not have a SSL master certificate, so it will use the default OpenSSL configuration to validade certificates!");
```
cakrit: validade -> validate.

streaming/rrdpush.c
```diff
@@ -90,13 +95,20 @@ int rrdpush_init() {
...
     if ( !strcmp(invalid_certificate,"yes")){
         if (netdata_validate_server == NETDATA_SSL_VALID_CERTIFICATE){
-            info("The Netdata is configured to accept invalid certificate.");
+            info("The Netdata is configured to accept invalid SSL certificate.");
             netdata_validate_server = NETDATA_SSL_INVALID_CERTIFICATE;
         }
     }
+
+    netdata_ssl_cert_directory = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "CApath", "/etc/ssl/certs/");
+    netdata_ssl_master_cert = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "CAfile", "/etc/ssl/certs/");
```
cakrit: But this default value is not a file, it's a directory!

streaming/rrdpush.c
```diff
@@ -90,13 +95,20 @@ int rrdpush_init() {
...
+    if(!strcmp(netdata_ssl_cert_directory,netdata_ssl_master_cert)) {
```
cakrit: This isn't correct. Either of the two settings could actually be null.

streaming/rrdpush.c
thiagoftsm: I changed the word and I moved the line for security.c where it makes more sense now.

streaming/rrdpush.c
thiagoftsm: The idea here was to verify whether the user gave some different value, this was the motive I set them equal, but I understood your point of view and I changed the default CAfile argument.

streaming/rrdpush.c
thiagoftsm: It was also fixed and the description for this is not different from the previous that I gave.

thiagoftsm:
@joelhans thank you very much by the revision in the grammar and style!

joelhans:
:+1: on the documentation part of it now.

streaming/rrdpush.c
ilyam8: Dir as default for CAFile?

> /etc/ssl/certs/

Since those parameters related to CA i think it is nice to have `ca` in the name.

ilyam8: `CApath` and `CAfile` have same meaning, at least for me. `path` suffix doesnt mean dirtory. If we have `CAFile` then it feels natural to use `CADir`.

paulkatsoulakis: > `CApath` and `CAfile` have same meaning, at least for me. `path` suffix doesnt mean dirtory. If we have `CAfile` then it feels natural to use `CAdir`.

I lost you a little bit here, those two variables mean exactly what they say.
CAPath -> The directory where multiple certificates on the right format may exist.
CAFile -> A single CA Certificate file path

ilyam8: i am only saying that `CADir` looks better for me then `CAPath` since we are using `CAFile`. I admit that `CAPath` will do too.

thiagoftsm: > Dir as default for CAFile?
> 
> > /etc/ssl/certs/
> 

Hi @ilyam8 , this was the first commit in this PR, it is not my this way since yesterday last commit. I used the fact that on Unices everything is a file

> Since those parameters related to CA i think it is nice to have `ca` in the name.
> 
> > netdata_ssl_cert_directory
> > netdata_ssl_master_cert

I liked your idea, it will be more readable.

>>i am only saying that CADir looks better for me then CAPath since we are using CAFile (it is obvious that both of them are paths). I admit that CAPath will do too.

I agree completely, I also prefer Directory instead Path, but case we do this we will lose the connection with the OpenSSL documentation. We renamed it to CApath and CAfile for our users to have more sources of information.

@paulkatsoulakis due the fact it was not commit, I executed Ilya's suggestion in  this PR yet. This PR also brings fixes for the 1.16.1, is it not better to merge it?

libnetdata/socket/security.c
```diff
@@ -283,7 +297,48 @@ int security_test_certificate(SSL *ssl) {
...
+ *
+ * @param ctx the context where the path will be set.
+ * @param file the file with Netdata master certificate.
+ * @param path the directory where the certificates are stored.
+ *
+ * @return It returns 0 on success and -1 otherwise.
+ */
+int security_location_for_context(SSL_CTX *ctx, char *file, char *path) {
+    struct stat statbuf;
+    if (stat(file, &statbuf)) {
```
paulkatsoulakis: Correct if i miss some concept, but this means we always need a certificate file available in `*file`, in order to validate with the cert directory.  Is this intentional?

paulkatsoulakis: Maybe `*file` is the certificate we validate ?

PR #6555, opened 2019-07-27
netdata/packaging: Move tarball checksum information into lib dir of netdata

### LINKED ISSUE #6550: /etc/netdata/.environment gets updated constantly with auto-updates
##### Question summary
I am using `etckeeper` in a POC the server, and I have set that to manual commits only.

It is somewhat annoying that important values, such as `RELEASE_CHANNEL="nightly"` get to get mixed with `NETDATA_TARBALL_CHECKSUM` which gets updated "frequently enough"

`etckeeper` with manual commits basically means that e.g. `apt-get` won't work with a dirty `/etc`

##### OS / Environment

Ubuntu 18.04

##### Expected results

Move `NETDATA_TARBALL_CHECKSUM` out of `/etc`.
I am not sure of its significance, but in a quick assesment, it doesn't feel like a "serious configuration option"

paulkatsoulakis:
##### Summary
1) Introduce NETDATA_LIB_DIR within environment file, we are going to need to refer to it
2) move NETDATA_TARBALL_CHECKSUM to /netdata.tarball.checksum
3) Make sure you support transitional period, which means load and remove the variable from the old location then move it to the new one

##### Component Name
netdata/packaging/installer

##### Additional Information

Fixes #6550

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -985,8 +985,13 @@ fi
 
 # -----------------------------------------------------------------------------
 progress "Copy uninstaller"
-sed "s|ENVIRONMENT_FILE=\"/etc/netdata/.environment\"|ENVIRONMENT_FILE=\"${NETDATA_PREFIX}/etc/netdata/.environment\"|" packaging/installer/netdata-uninstaller.sh > ${NETDATA_PREFIX}/usr/libexec/netdata-uninstaller.sh
-chmod 750 ${NETDATA_PREFIX}/usr/libexec/netdata-uninstaller.sh
+if [ -f "${NETDATA_PREFIX}"/usr/libexec/netdata-uninstaller.sh ]; then
+	echo >&2 "Removing uninstaller from old location"
+	rm -f "${NETDATA_PREFIX}"/usr/libexec/netdata-uninstaller.sh;
+fi
+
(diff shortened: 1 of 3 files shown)


thiagoftsm:
Hi @paulkatsoulakis ,

Considering that you are changing the installer here, I will do a question for you. I was working here and I saw  that the scripts "netdata-uninstaller.sh" and "netdata-updater.sh" were installed in my directory /usr/libexec/ instead /usr/libexec/netdata , Why do not we put both in the last directory?

paulkatsoulakis:
> Hi @paulkatsoulakis ,
> 
> I tested your PR now, and I think we need to change the netdata-installer.sh, another possibility is the fact that I did not understand the idea behind the the checksum in the installer. After I executed the netdata-installer.sh, we have the file netdata.tarball.checksum in the correct directory, but instead to have a checksum we have "new_installation", this string is not a checksum.
> Now about the update, at the end of the process I had the checksum stored in the same file that I had only "new_installation".
> My suggestion for you, let us have two lines in the file netdata.tarball.checksum, the first line saying that we have a new installation or an update executed and the second line will be the checksum of them.
> 
> Best regards!

new installation always gets that keyword "new_installation" as checksum.

paulkatsoulakis:
> Hi @paulkatsoulakis ,
> 
> Considering that you are changing the installer here, I will do a question for you. I was working here and I saw that the scripts "netdata-uninstaller.sh" and "netdata-updater.sh" were installed in my directory /usr/libexec/ instead /usr/libexec/netdata , Why do not we put both in the last directory?

Good point, i guess to avoid deleting itself but that's not a problem i guess.
I 'll move it and test it.

Issue #4934, opened 2018-12-06
netdata server usage - stream statistics


gdelafond:
Hello,

On each netdata there is an information about its memory usage like : 
` netdata on XXXX, collects every second 1 801 metrics, presented as 260 charts and monitored by 0 alarms, using 297 MB of memory for 11 hours, 55 mins and 8 secs of real-time history.`

I do not find how to know the number of received stream and the total ram used by these streams. Is this information available somewhere ?

Regards,

cakrit:
Well, it's not precisely what you're asking, but check [here](https://docs.netdata.cloud/database/). If by 'stream' you are referring to a hierarchical setup with a master node receiving metrics from slave nodes, you can see [here](https://docs.netdata.cloud/streaming/#options-for-the-receiving-node) a section in the configuration where you can change the history (entries stored) for each slave. 

Finally, you may want to also look at the general suggestions on [Performance](https://docs.netdata.cloud/docs/performance/), as disabling plugins you don't need also reduces the required memory.

gdelafond:
@cakrit yes i'm referring to a hierarchical setup. I know there are information in the documentation, but here I have hundreds of server streaming to the master node. Every day this number is different, it is not very easy to know easily the node number and the RAM used by all the stream on the master node.

Maybe a graph in the "Netdata monitring/netdata" group could be added to monitor the number of stream received and the RAM used by these stream.

cakrit:
This would be an interesting feature, for sure. The title is appropriate, "stream statistics" as we probably don't just want to show the RAM used. Adding it to the backlog.

ilyam8:
Closing this feature request. We will re-evaluate the request internally.

PR #6558, opened 2019-07-28
Attempt to provide stream statistics as per #4934 request

stelfrag:
##### Summary
Provide a new section under netdata / monitoring for stream statistics. The new feature
is activated by adding a [global] section in the streams.conf and setting
###### activate stream statistics = yes

The default setting is "no"

##### Component Name
Affected files

- collectors/plugins.d/plugins_d.c 
- collectors/plugins.d/plugins_d.h 
- streaming/rrdpush.c


collectors/plugins.d/plugins_d.c
```diff
@@ -246,6 +277,115 @@ inline size_t pluginsd_process(RRDHOST *host, struct plugind *cd, FILE *fp, int
...
+                , "NetData Incoming Streams"
+                , "Count"
+                , "netdata"
+                , "stats"
+                , 130004
+                , localhost->rrd_update_every
+                , RRDSET_TYPE_LINE
+            );
+
+            if (likely(__stream_mem_st))
```
thiagoftsm: Hi @stelfrag,

Please, add an else in this group for you write an error log case there was a problem to allocate the variables. I saw that you are checking whether these variables are not null in the rest of the code, but it is also good to know when something wrong happens.

stelfrag: Thank you @thiagoftsm for pointing this out!! I will add it

thiagoftsm: Nice! I will  mark as resolved, because you solved as expected! :)

streaming/rrdpush.c
```diff
@@ -76,6 +77,7 @@ int rrdpush_init() {
     default_rrdpush_api_key     = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "api key", "");
     default_rrdpush_send_charts_matching      = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "send charts matching", "*");
     rrdhost_free_orphan_time    = config_get_number(CONFIG_SECTION_GLOBAL, "cleanup orphan hosts after seconds", rrdhost_free_orphan_time);
+    stream_stats_enabled        = (unsigned int)appconfig_get_boolean(&stream_config, CONFIG_SECTION_GLOBAL, "activate stream statistics", 0);
```
thiagoftsm: Hi @stelfrag

The first time that I review your code I did not test it, because I saw we could miss some important information from the Operate System, after you fix that problem, today I began to test your PR.  In the first test that I executed, I decided to print the variable stream_stats_enabled inside the log files on Netdata master, and it was always set to zero independent of the position that I put  the parameter "activate stream statistics"  inside my stream.conf file, so I came here and I saw that you are mixing concepts. The "stream_config" variable is associated with "CONFIG_SECTION_STREAM" and here your are putting it with the CONFIG_SECTION_GLOBAL that is a section of the netdata.conf, please, fix this for I continue with the tests.

stelfrag: Hi @thiagoftsm 

Thank you for your feedback. I am assuming you put a **[global]** section in the
*stream.conf*  (eg. /etc/netdata/stream.conf) file with a parameter 

```
[global]
    activate stream statistics = yes
```

The code block below is intended to look for the configuration parameter in the stream.conf under a *[global]* section

`stream_stats_enabled        = (unsigned int)appconfig_get_boolean(&stream_config, CONFIG_SECTION_GLOBAL, "activate stream statistics", 0);`

I will try to reproduce the behavior you are experiencing and analyze the logic of the
*appconfig_get_boolean* function to make sure it is behaving as I expected it to do
so I can provide solution

thank you !!

thiagoftsm: Now I understood what you are trying to do better, so I will ask more two things to you. As you can see we have inside the directory streaming the file README.md, this is a file where we write our documentation, please, explain inside this file the new section in our stream.conf. No less important, inside the same directory(streaming) we also have a stream.conf file that is copied for our users when they use the command /etc/netdata/edit-config, add the newest section inside the file, I would prefer to insert on top, and comment it there.

Best Regards!

stelfrag: Thanks for the feedback @thiagoftsm . I will update README.md and stream.conf

cakrit:
Where am I supposed to see these metrics? I enabled it, have a slave, but can't find them:
Screenshot from 2019-08-09 11-07-22

Also, I'd set the default to "yes".

stelfrag:
Once activated and netdata is restarted there will be a  a "streams" section under Netdata monitoring (refresh may be required)

ilyam8:
I installed the PR and i dont see 

> "streams" section under Netdata monitoring

I restarted netdata service on master several times, still no stream metrics.

stelfrag:
> I installed the PR and i dont see
> 
> > "streams" section under Netdata monitoring
> 
> 
> i did:
> 
> * intalled PR
> * added `activate stream statistics = yes` to `[global] section`
> * restarted netdata service on master
> 
> My setup: master-1 slave. I see slave in the `your other netdata servers` section.

Hello @ilyam8 please confirm that you created a [global] section in your *stream.conf* configuration file and added the `activate stream statistics = yes` there

ilyam8:
@stelfrag 	it is been a while... To be honest i dont remember, will test the PR again today later

ilyam8:
ok, i wont be testing since @cakrit confirms it is working, i assume i had same problems:

> I saw the charts immediately, I had made the mistake of putting the option in the netdata.conf [global] option before. 

Stream statistics looks like an important thing, why dont enable it by default?

collectors/plugins.d/plugins_d.c
```diff
@@ -377,6 +549,39 @@ inline size_t pluginsd_process(RRDHOST *host, struct plugind *cd, FILE *fp, int
             if(unlikely(rrdset_flag_check(st, RRDSET_FLAG_DEBUG)))
                 debug(D_PLUGINSD, "requested an END on chart %s", st->id);
 
+            if (likely(stream_stats_enabled) && (now_realtime_sec() - stream_updated > 0)) {
```
cosmix: Same here: I would probably use a an array literal of a `static struct` to define these and iterate over it, instead of repeating the `if` statements here.

streaming/stream.conf
```diff
@@ -218,3 +218,19 @@
     #proxy destination = IP:PORT IP:PORT ...
     #proxy api key = API_KEY
     #proxy send charts matching = *
+
+#
+# Glogal section for stream configuration on the master server
```
cosmix: Typo: Global

collectors/plugins.d/plugins_d.c
```diff
@@ -246,6 +277,129 @@ inline size_t pluginsd_process(RRDHOST *host, struct plugind *cd, FILE *fp, int
...
+                , "NetData Incoming Streams"
+                , "Count"
+                , "netdata"
+                , "stats"
+                , 130004
+                , localhost->rrd_update_every
+                , RRDSET_TYPE_LINE
+            );
+
+            if (likely(__stream_mem_st))
```
cosmix: I would use an array literal of `static struct`s here and iterate over it instead of having several repeating if statements. It’s cleaner, less prone to error, more maintainable and the performance impact is minimal.

stelfrag:
> I see what's happening now. The request to put this setting in a `[global]` section of `stream.conf` doesn't seem to make much sense. I only enabled it on the master and I saw the statistics immediately. So if we are to having a setting at all, it belongs inside the master's configuration in `stream.conf` (i.e. under what we is labelled in a comment as `2. ON MASTER NETDATA - THE ONE THAT WILL BE RECEIVING METRICS`, not in a new, `[global]` section.
> 
> I saw the charts immediately, I had made the mistake of putting the option in the `netdata.conf` `[global]` option before. I'd still like this to change though.

The reason for the new section is that the option is  global. There is no section on the master Netdata other than [API_KEY] and [MACHINE_GUID] and it would be misleading to add the option there e.g under a specific [API_KEY] and have it globally enabled.

In that regard, it would make more sense to put the option in the main Netdata configuration file (but that would split the streams configuration in two files if this feature is is to be expanded. 

This would also allow one to disable the stream statistics per [API_KEY] in an intuitive way i.e. 

- Enable globally
- Disable on demand per [API_KEY]

One way to enhance the stream statistics and might as well require new options in  the [API_KEY] section would be to add group statistics, so several key metrics of the "same" group would be presented together in graphs

Example: monitor 100 slaves for CPU usage averages (e.g. user, system, idle) in a single graph by 
setting a "stream group code" configuration in the API KEY section

Of course automatic grouping can also occur e.g by OS version, distribution etc

> A couple more things:
> 
> * Check the tabs in the `stream.conf` file, some comments were off. You'll move them again now anyway, due to the change above.
> * The README should probably say that by default in v1.17 and beyond the master stores and displays the stream statistics, you can disable it by setting the option to 'no'
> * There's a conflict in the README, you'll need to do a rebase there.

Ok making a note on those

collectors/plugins.d/plugins_d.c
stelfrag: Hi @cosmix 

You are right. It would make more sense to have an iteration and it will be the way to go if this is to be further expanded -- it can easily have half a dozen of more charts to process and it would be a nightmare to review and maintain. 

In addiction to that it would almost certainly go beyond static structures and require additional book keeping (memory allocation) -- image just these 5 charts grouped by some sort of OS / Kernel version or CPU arch when having a couple of hundred slaves 

All in all  yes !  this would would be rewritten

cakrit:
@stelfrag I resolved the conflicts, but please do a commit with the additional fixes requested here, so we can merge.

streaming/rrdpush.c
```diff
@@ -78,6 +79,9 @@ int rrdpush_init() {
     default_rrdpush_api_key     = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "api key", "");
     default_rrdpush_send_charts_matching      = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "send charts matching", "*");
     rrdhost_free_orphan_time    = config_get_number(CONFIG_SECTION_GLOBAL, "cleanup orphan hosts after seconds", rrdhost_free_orphan_time);
+    stream_stats_enabled        = (unsigned int)appconfig_get_boolean(&stream_config, CONFIG_SECTION_GLOBAL, "activate stream statistics", 1);
```
vlvkobal: `1` -> `CONFIG_BOOLEAN_YES`

streaming/stream.conf
```diff
@@ -218,3 +218,19 @@
...
+#
+
+[global]
+				# Display a new "streams" sections under netdata monitoring
+    # that will provide several stream statistics. For now this
+    # includes 4 new charts: 
+				# 1. Number of incoming streams (slaves)
+    # 2. Number of processors (Active processors reported by the slaves)
+    # 3. Number of total slave metrics (dimensions)
+    # 4. Number of total activate slave processes (system.active_processes reported by the slaves)
```
vlvkobal: `activate -> active`

vlvkobal:
When slave is (re)connecting, there are a lot of "heavy" calls (`strlen`, `rrdset_find`, `rrddim_find`) for every chart (hundreds or even thousands for every slave), dimension (thousands for every slave), and variable (tens of thousands for every slave).

Also, on every update for every chart `strcmp` is called.

Can't it be a serious performance impact for "fat" masters with many slaves? Or even for "thin" masters?

streaming/rrdpush.c
stelfrag: Will change to CONFIG_BOOLEAN_YES

stelfrag:
> When a slave is (re)connecting, there are a lot of "heavy" calls (`strlen`, `rrdset_find`, `rrddim_find`) for every chart (hundreds or even thousands for every slave), dimension (thousands for every slave), and variable (tens of thousands for every slave).
> 
> Also, on every update for every chart `strcmp` is called.
> 
> Can't it be a serious performance impact for "fat" masters with many slaves? Or even for "thin" masters?

Hi @vlvkobal 

It's true you get a lot of strlen via the **get_set_memory_usage** calls. That's in order to be able to more accurately calculate the memory usage. Assuming a maximum value for those (even larger that what would actually be allocated -- say 256 bytes for all those values) could be "close enough" and good enough calculation.

The rrdset_find  and rrddim_find calls cannot be avoided right now. This is because the sets and dimensions survive a slave re-connection (and rightly so because it would be more expensive to 
de-allocate the memory each time a slave restarts or there is network disconnection for a second)

Things would be easier if rrdset_create and rrddim_add would pass back information if the set and dim respectively already existed, therefore there is no actual memory allocation. 

This would require some rewrite (another version of rrddim_add and rrdset_create that would either return a struct or fill a struct with additional information)

Unless there is a variable in the set or dim that I can use and safely assume this is a freshly allocated 
structure

stelfrag:
> @stelfrag I resolved the conflicts, but please do a commit with the additional fixes requested here, so we can merge.

We agreed on a **[master_streaming]** section that will go in the stream.conf file on section 2, just above the [API_KEY] section. 

This will: 

- Keep the stream related configuration in one file (on the master and slaves)
- Avoid the "global" keyword that will confuse the community
- The prefix **master_** will further emphasize this will have only effect on the master node

streaming/README.md
```diff
@@ -108,6 +108,27 @@ A new file is introduced: [stream.conf](stream.conf) (to edit it on your system
 `/etc/netdata/edit-config stream.conf`). This file holds streaming configuration for both the
 sending and the receiving Netdata.
 
+You can add a [global] section on the master server's [stream.conf](stream.conf) and activate
+additional slave statistics as follows.
+
+```
+[global]
+				activate stream statistics = yes
```
joelhans: Too much indentation here. Looks funky when rendered—Markdown respects all that whitespace.

```suggestion
    activate stream statistics = yes
```
```

File: streaming/README.md [hunk 9]
```diff
@@ -108,6 +108,27 @@ A new file is introduced: [stream.conf](stream.conf) (to edit it on your system
...
+```
+
+This will add a new streams section under netdata monitoring and provide five new charts with slave
```
joelhans: netdata -> Netdata

streaming/README.md
```diff
@@ -108,6 +108,27 @@ A new file is introduced: [stream.conf](stream.conf) (to edit it on your system
...
+statistics
```
joelhans: Put a colon `:` at the end here.

streaming/README.md
```diff
@@ -108,6 +108,27 @@ A new file is introduced: [stream.conf](stream.conf) (to edit it on your system
...
+
+A "memory" chart that attempts to measure the amount of RAM allocated for the slaves in the master server (charts and dimensions)
+
+A "streams" chart that reports the number currently incoming streams (slaves)
+
+A "processors" chart that reports the number of active processors of all slaves by incercepting the active processors HOST variable
+
+A "metrics" chart to count all the dimensions provided by the slaves
+
+A "processes" chart for active processes reported by the slaves by intercepting the "system.active_processes" dimension
```
joelhans: I think this might be better served as an unordered list.

streaming/README.md
```diff
@@ -108,6 +108,27 @@ A new file is introduced: [stream.conf](stream.conf) (to edit it on your system
...
+```
+
+This will add a new streams section under netdata monitoring and provide five new charts with slave
+statistics
+
+A "memory" chart that attempts to measure the amount of RAM allocated for the slaves in the master server (charts and dimensions)
+
+A "streams" chart that reports the number currently incoming streams (slaves)
+
+A "processors" chart that reports the number of active processors of all slaves by incercepting the active processors HOST variable
```
joelhans:
```suggestion
A "processors" chart that reports the number of active processors of all slaves by intercepting the active processors' `HOST` variable
```

streaming/README.md
joelhans: I think it makes sense to put dimensions in an inline code block.

```suggestion
A "processes" chart for active processes reported by the slaves by intercepting the `system.active_processes` dimension
```

PR #6561, opened 2019-07-29
Handle disconnected sockets in unbound collector.

### LINKED ISSUE #6434: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected FreeBSD
Hi,

I have set-up netdata (v1.15.0) to monitor my local unbound server on FreeBSD 12 through its control socket.

My error logs are flooding every second (poll rate) with the following: 
2019-07-11 19:42:51: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected
2019-07-11 19:42:52: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected
2019-07-11 19:42:53: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected
2019-07-11 19:42:54: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected
2019-07-11 19:42:55: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected
2019-07-11 19:42:56: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected
2019-07-11 19:42:57: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected
2019-07-11 19:42:58: python.d ERROR: unbound[local] : [Errno 57] Socket is not connected

I have in unbound.conf (python.d plugin):
```
local:
 extended: true
 socket: /var/run/unbound/unbound-control.sock
```

I have in unbound.conf:
```
remote-control:
        control-enable: yes
        control-interface: /var/run/unbound/unbound-control.sock
```

**The unbound charts are working fine, although logs are obviously flooding.**

Ferroin:
#### Summary
This adds an explicit check for the case of a socket that's already
disconnected and skips logging an error message.  The conditionn
technically is an error, but it's one that we can recover from trivially
by just doing nothing in this case (we were trying to disconnect the
scoket anyway, so if it's already disconnected, we don't need to change
anything).

This uses Python's `errno` module so that we can detect this situation
in a system-agnostic manner.

##### Component Name
collectors/python.d

##### Additional Information

Fixes #6434

-----

@ilyam8 I'm of the opinion that this check should probably just go into the `SocketService` base class instead of being an as-needed thing in specific modules.  I honestly can't think of any case where a socket being disconnected as reported in #6434 would be an actual error in the disconnect method (it just means the socket is already in the state we want it to be in).

@Duffyx Could you check and see if this change stops the error logs getting flooded with the message you were seeing?  As mentioned in the issue itself, I wasn't able to reproduce this locally, so I'm going to need some help to make sure it works.

### DIFF
--- a/collectors/python.d.plugin/unbound/unbound.chart.py
+++ b/collectors/python.d.plugin/unbound/unbound.chart.py
@@ -5,6 +5,7 @@
 
 import os
 import sys
+import errno
 
 from copy import deepcopy
 
@@ -263,6 +264,21 @@ class Service(SocketService):
                 self.perthread = False
             self.request = tmp
         return result
+
+    def _disconnect(self):
+        # Custom _disconnect method to work around the issue reported in #6434
+        if self._sock is not None:
+            try:
+                self.debug('closing socket')
+                self._sock.shutdown(2)  # 0 - read, 1 - write, 2 - all
+                self._sock.close()
+            except Exception as error:
+                if ('errno' in error) and (error.errno == errno.ENOTCONN):
+                    pass
+                else:
+                    self.error(error)
+                self.error(error)
+            self._sock = None
 
     @staticmethod
     def _check_raw_data(data):
--- a/collectors/python.d.plugin/unbound/unbound.chart.py
+++ b/collectors/python.d.plugin/unbound/unbound.chart.py
@@ -5,7 +5,6 @@
 
 import os
 import sys
-import errno
 
 from copy import deepcopy
 
@@ -264,18 +263,6 @@ class Service(SocketService):
                 self.perthread = False
             self.request = tmp
         return result
-
-    def _disconnect(self):
-        # Custom _disconnect method to work around the issue reported in #6434
-        if self._sock is not None:
-            try:
-                self.debug('closing socket')
-                self._sock.shutdown(2)  # 0 - read, 1 - write, 2 - all
-                self._sock.close()
-            except Exception as error:
-                if not (hasattr(error, 'errno') and error.errno == errno.ENOTCONN):
-                    self.error(error)
-            self._sock = None
 
     @staticmethod
     def _check_raw_data(data):
--- a/collectors/python.d.plugin/unbound/unbound.chart.py
+++ b/collectors/python.d.plugin/unbound/unbound.chart.py
@@ -255,7 +255,7 @@ class Service(SocketService):
             raw = self._get_raw_data()
             if raw is None:
                 result = False
-                self.warning('Recieved no data from socket.')
+                self.warning('Received no data from socket.')
             else:
                 for line in raw.splitlines():
                     if line.startswith('threads'):
@@ -278,10 +278,13 @@ class Service(SocketService):
         raw = self._get_raw_data()
         data = dict()
         tmp = dict()
-        for line in raw.splitlines():
-            stat = line.split('=')
-            tmp[stat[0]] = stat[1]
-        for item in self.statmap:
-            if item in tmp:
-                data[self.statmap[item][0]] = float(tmp[item]) * self.statmap[item][1]
+        if raw is not None:
+            for line in raw.splitlines():
+                stat = line.split('=')
+                tmp[stat[0]] = stat[1]
+            for item in self.statmap:
+                if item in tmp:
+                    data[self.statmap[item][0]] = float(tmp[item]) * self.statmap[item][1]
+        else:
+            self.warning('Received no data from socket.')
         return data
--- a/collectors/python.d.plugin/unbound/unbound.chart.py
+++ b/collectors/python.d.plugin/unbound/unbound.chart.py
@@ -253,15 +253,19 @@ class Service(SocketService):
             else:
                 self.request = b'UBCT1 status\n'
             raw = self._get_raw_data()
-            for line in raw.splitlines():
-                if line.startswith('threads'):
-                    self.threads = int(line.split()[1])
-                    self._generate_perthread_charts()
-                    break
-            if self.threads is None:
-                self.info('Unable to auto-detect thread counts, disabling per-thread stats.')
-                self.perthread = False
-            self.request = tmp
+            if raw is None:
+                result = False
+                self.warning('Recieved no data from socket.')
+            else:
+                for line in raw.splitlines():
+                    if line.startswith('threads'):
+                        self.threads = int(line.split()[1])
+                        self._generate_perthread_charts()
+                        break
+                if self.threads is None:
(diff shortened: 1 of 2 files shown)


collectors/python.d.plugin/unbound/unbound.chart.py
```diff
@@ -264,6 +265,21 @@ def check(self):
...
+            try:
+                self.debug('closing socket')
+                self._sock.shutdown(2)  # 0 - read, 1 - write, 2 - all
+                self._sock.close()
+            except Exception as error:
+                if ('errno' in error) and (error.errno == errno.ENOTCONN):
+                    pass
+                else:
+                    self.error(error)
+                self.error(error)
```
ilyam8: we dont need last `self.error(error)

Duffyx:
Hmm, this seems to break functionality as the charts don't work anymore. The error:

2019-07-29 19:04:21: python.d ERROR: unbound[local] : update() unhandled exception: argument of type 'OSError' is not iterable
2019-07-29 19:04:24: python.d ERROR: unbound[local] : unix socket "/var/run/unbound/unbound-control.sock": error sending request: [Errno 32] Broken pipe
2019-07-29 19:04:24: python.d ERROR: unbound[local] : update() unhandled exception: argument of type 'OSError' is not iterable
2019-07-29 19:04:27: python.d ERROR: unbound[local] : unix socket "/var/run/unbound/unbound-control.sock": error sending request: [Errno 32] Broken pipe
2019-07-29 19:04:27: python.d ERROR: unbound[local] : update() unhandled exception: argument of type 'OSError' is not iterable
2019-07-29 19:04:33: python.d ERROR: unbound[local] : unix socket "/var/run/unbound/unbound-control.sock": error sending request: [Errno 32] Broken pipe
2019-07-29 19:04:33: python.d ERROR: unbound[local] : update() unhandled exception: argument of type 'OSError' is not iterable

Ferroin:
@Duffyx Yeah, looking back at the code now, that's my fault (I've apparently gotten too used to JavaScript's object model...).

Fixed that particular issue now.

Duffyx:
:-) Thats fine. Just run time tested this on my machine. Works great. No more log flooding, graphs are working. Thanks!

ilyam8:
> update() unhandled exception: 'NoneType' object has no attribute 'splitlines'

Nice catch @Duffyx, @Ferroin please fix the issue in this PR too.

> I'm of the opinion that this check should probably just go into the SocketService base class instead of being an as-needed thing in specific modules.

lgtm

Ferroin:
>> I'm of the opinion that this check should probably just go into the SocketService base class instead of being an as-needed thing in specific modules.
>>
> lgtm

@ilyam8 would you prefer I just update this PR for that, or should I close this one and open a new one?

ilyam8:
Update this one is ok

Ferroin:
Updated to fix the issue that @Duffyx noticed, as well as moving the check to the `SocketService` class so that all modules can benefit from it.

@Duffyx that last error line you saw should be replaced with this update to be a warning instead.  I don't think there's anything we can do about the first two error lines, but as long as the graphs are working it's probably not an issue (especially since it's only happening on startup).

ilyam8:
@Ferroin 	

https://github.com/netdata/netdata/blob/fc3ca0d3c1f0971bfb0b1822abef7b1d2ccaafd5/collectors/python.d.plugin/unbound/unbound.chart.py#L277-L281

ilyam8:
@Duffyx please test the fix one more time

Duffyx:
2019-07-30 20:41:29: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:29: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:30: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:30: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:31: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:31: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:32: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:32: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:33: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:33: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:34: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:34: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:35: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:35: python.d WARNING: unbound[local] : Received second no data from socket.

Error message: Received second no data from socket. The "second" I added to debug from which statement it was coming. But it is thus coming from _get_data function.

Ferroin:
@Duffyx Huh, that's odd that it would be flooding the logs.  The condition it's looking for is what should have been triggering that single group of error messages on startup that you mentioned above.  Can you confirm for certain that the info shown in the charts actually makes sense (maybe do something that would look up a lot of DNS records via Unbound and confirm that the charts show the requests)?

@ilyam8 Any ideas why this would now be spamming the log when the same condition it's looking for only happened once before?

Duffyx:
@Ferroin So I just did some DNS benchmarks and charts seem to react just fine on the load.

ilyam8:
> 2019-07-30 20:41:34: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:34: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:35: python.d WARNING: unbound[local] : Received second no data from socket.
2019-07-30 20:41:35: python.d WARNING: unbound[local] : Received second no data from socket.

I see 2 log messages every second. @Duffyx You are saying that module works fine, you see the charts on the dashboartd and there are no gaps and at the same time there is warning msg every seconds and that msg comes from `_get_data`?

Ferroin:
@ilyam8 Nice catch that it's happening more than once a second, I hadn't noticed that.  That makes me think that my initial gut feeling about this was right, and Python is being stupid and associating the else clause with the wrong if statement (there's an if statement inside a loop on the immediately previous line).

Duffyx:
Second, hold your horses. My patches might be not 100% right after you stated that its python thats stupid. I just noticed my identation wasnt quite right.

Duffyx:
Oh yeah baby, it's the chosen one! :-) I'm not seeing any more errors at all related to the unbound plugin after applying this fix. Sorry for that last noise, thats on me!

Issue #4817, opened 2018-12-01
Gearmand monitoring


ilyam8:
We are willing to support monitoring ***[Gearmand](http://gearman.org/)*** with netdata.

To get started, we need a user that will act as a [sponsor](https://docs.netdata.cloud/contributing/#sponsor-a-part-of-netdata) during the implementation.
He/she will assist us with the following:

  - Specify the metrics required and how they should be presented in charts.
  - Specify the alarms that would make sense for each metric.
  - When the implementation passes QA, test the implementation in production.
  - Use the charts and alarms in their day to day work and provide additional feedback after the collector is delivered.

We would also appreciate getting votes :+1: for this issue from the community, to understand the interest in the particular collector.

agronick:
I have made a chart plugin. I think the stacked chart works well for this:

![image](https://user-images.githubusercontent.com/2042303/61063448-f8d76680-a3cd-11e9-9d2b-a050ade75bbd.png)

I will clean it up and post to Github.

agronick:
Here it is https://github.com/agronick/gearman-netdata/tree/master

ilyam8:
Implementation looks a little bit hacky. Are you willing to make a PR btw?

About the gearmand thing, i checked [`datadog` implementation](https://github.com/DataDog/integrations-core/tree/master/gearmand)

They collect summary and per stats queued/idle/active workers statistics. There is no summary stats in your implementation.

I see there can be a lot of tasks. For some reason datadog has hardcoded limitation - [200 tasks](https://github.com/DataDog/integrations-core/blob/7a0b2a7d52593deb95cd927d43dee9da082a8ad5/gearmand/datadog_checks/gearmand/gearmand.py#L18) to collect data from.


```python
            # Display a warning in the info page
            self.warning(
                (
                    "Too many tasks to fetch. "
                    "You must choose the tasks you are interested in by editing the gearmand.yaml configuration file "
                    "or get in touch with Datadog support"
                )
            )
```

:smile:

ilyam8:
The funny thing that 200 is hardcoded and there is no way to change via config file (maybe i am missing something, dunno)

https://github.com/DataDog/integrations-core/blob/7a0b2a7d52593deb95cd927d43dee9da082a8ad5/gearmand/datadog_checks/gearmand/gearmand.py#L71-L80

agronick:
I can make a PR. So you want a summary graph and then the ability for users to set the jobs they care about in the config? Should it default to all jobs if none is set? 

[...]

PR #6567, opened 2019-07-30
Gearman plugin for Netdata

agronick:
# Summary
This adds a Gearman chart plugin discussed in #4817.
##### Component Name
python.d/gearman
health.d/gearman
##### Additional Information
I removed idle from the total graph because a worker could be counted more than once if it was registered to more than one job.

### DIFF
--- a/collectors/python.d.plugin/gearman/gearman.chart.py
+++ b/collectors/python.d.plugin/gearman/gearman.chart.py
@@ -93,7 +93,7 @@ class Service(SocketService):
     def _get_worker_data(self):
         """
         Split the data returned from Gearman into a list of lists
-        :return: list
+        :return: generator
         """
 
         try:
@@ -127,16 +127,31 @@ class Service(SocketService):
         }
 
     def _add_chart(self, job_name):
-        self.active_jobs.add(job_name)
+        """
+        Adds a new job chart
+        :param job_name: The name of the job to add
+        :type job_name: string
+        :return: None
+        """
+
         job_key = 'job_{0}'.format(job_name)
         template = job_chart_template(job_name)
         new_chart = self.charts.add_chart([job_key] + template['options'])
         for dimension in template['lines']:
             new_chart.add_dimension(dimension)
 
+        self.active_jobs.add(job_name)
+
     def _remove_chart(self, job_name):
-        self.active_jobs.remove(job_name)
+        """
+        Removes a job chart
+        :param job_name: The name of the job to remove
+        :type job_name: string
+        :return: None
+        """
+
         job_key = 'job_{0}'.format(job_name)
         self.charts[job_key].obsolete()
+        self.active_jobs.remove(job_name)
 
 
--- a/collectors/python.d.plugin/gearman/gearman.chart.py
+++ b/collectors/python.d.plugin/gearman/gearman.chart.py
@@ -6,6 +6,7 @@
 
 from bases.FrameworkServices.SocketService import SocketService
 from copy import deepcopy
+import collections
 
 priority = 60500
 
@@ -112,15 +113,14 @@ class Service(SocketService):
         try:
             raw = self._get_raw_data()
         except (ValueError, AttributeError):
-            return
+            return []
 
         if raw is None:
             self.debug("Gearman returned no data")
-            return
+            return []
 
         job_lines = raw.splitlines()[:-1]
-        job_lines = [job.split() for job in job_lines]
-        job_lines.sort(key=lambda line: line[0])
+        job_lines = [job.split() for job in sorted(job_lines)]
 
         for line in job_lines:
             line[1:] = map(int, line[1:])
@@ -129,14 +129,7 @@ class Service(SocketService):
 
     def process_jobs(self, active_jobs):
 
-        # There is no way to get an accurate total idle count because
-        # a worker can be assigned to more than one job.
-        output = {
-            'total_pending': 0,
-            'total_idle': 0,
-            'total_running': 0,
-        }
-
+        output = collections.defaultdict(int)
         found_jobs = set()
 
         for parsed_job in active_jobs:
--- a/collectors/python.d.plugin/gearman/gearman.chart.py
+++ b/collectors/python.d.plugin/gearman/gearman.chart.py
@@ -85,11 +85,12 @@ class Service(SocketService):
 
         try:
             active_jobs = self.get_active_jobs()
-            found_jobs, job_data = self.process_jobs(active_jobs)
-            self.remove_stale_jobs(found_jobs)
-            return job_data
         except GearmanReadException:
             return None
+
+        found_jobs, job_data = self.process_jobs(active_jobs)
+        self.remove_stale_jobs(found_jobs)
+        return job_data
 
     def get_active_jobs(self):
         active_jobs = []
--- a/collectors/python.d.plugin/gearman/gearman.chart.py
+++ b/collectors/python.d.plugin/gearman/gearman.chart.py
@@ -13,10 +13,10 @@ priority = 60500
 
 CHARTS = {
     'total_workers': {
-        'options': [None, 'Total Workers', 'Workers', 'Total Workers', 'gearman.workers', 'stacked'],
+        'options': ['total', 'Total Workers', 'Workers', 'Total Workers', 'gearman.total', 'line'],
         'lines': [
             ['total_queued', 'Queued', 'absolute'],
-            ['total_active', 'Active', 'absolute'],
+            ['total_running', 'Running', 'absolute'],
         ]
     },
 }
@@ -24,11 +24,11 @@ CHARTS = {
 
 def job_chart_template(job_name):
     return {
-        'options': [None, job_name, 'Workers', 'Jobs', 'gearman.{0}'.format(job_name), 'stacked'],
+        'options': [job_name, job_name, 'Workers', 'Jobs', 'gearman.job', 'stacked'],
         'lines': [
             ['{0}_queued'.format(job_name), 'Queued', 'absolute'],
             ['{0}_idle'.format(job_name), 'Idle', 'absolute'],
-            ['{0}_active'.format(job_name), 'Active', 'absolute'],
+            ['{0}_running'.format(job_name), 'Running', 'absolute'],
         ]
     }
 
@@ -60,7 +60,8 @@ class Service(SocketService):
         # a worker can be assigned to more than one job.
         output = {
             'total_queued': 0,
-            'total_active': 0,
+            'total_idle': 0,
+            'total_running': 0,
         }
 
         found_jobs = set()
@@ -80,7 +81,7 @@ class Service(SocketService):
                 output.update(job_data)
                 found_jobs.add(job_name)
 
-                for sum_value in ('queued', 'active'):
+                for sum_value in ('queued', 'running', 'idle'):
                     output['total_{0}'.format(sum_value)] += job_data['{0}_{1}'.format(job_name, sum_value)]
 
         for to_remove in self.active_jobs - found_jobs:
@@ -122,7 +123,7 @@ class Service(SocketService):
         return {
             '{0}_queued'.format(job[0]): pending,
             '{0}_idle'.format(job[0]): idle,
-            '{0}_active'.format(job[0]): running,
+            '{0}_running'.format(job[0]): running,
         }
 
     def _add_chart(self, job_name):
--- a/collectors/python.d.plugin/gearman/gearman.chart.py
+++ b/collectors/python.d.plugin/gearman/gearman.chart.py
@@ -6,14 +6,11 @@
 
 from bases.FrameworkServices.SocketService import SocketService
 from copy import deepcopy
-import collections
-
-priority = 60500
 
 
 CHARTS = {
     'total_workers': {
-        'options': ['total', 'Total Workers', 'Workers', 'Total Workers', 'gearman.total', 'line'],
+        'options': [None, 'Total Jobs', 'Jobs', 'Total Jobs', 'gearman.total_jobs', 'line'],
         'lines': [
             ['total_pending', 'Pending', 'absolute'],
             ['total_running', 'Running', 'absolute'],
@@ -24,7 +21,7 @@ CHARTS = {
 
 def job_chart_template(job_name):
     return {
-        'options': [job_name, job_name, 'Workers', 'Jobs', 'gearman.job', 'stacked'],
+        'options': [None, job_name, 'Jobs', 'Activity by Job', 'gearman.single_job', 'stacked'],
         'lines': [
             ['{0}_pending'.format(job_name), 'Pending', 'absolute'],
             ['{0}_idle'.format(job_name), 'Idle', 'absolute'],
@@ -58,6 +55,11 @@ def parse_worker_data(job):
         'metrics': job_metrics,
     }
 
+
+class GearmanReadException(BaseException):
+    pass
+
+
 class Service(SocketService):
     def __init__(self, configuration=None, name=None):
         super(Service, self).__init__(configuration=configuration, name=name)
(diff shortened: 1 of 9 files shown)


ilyam8:
@agronick check https://github.com/netdata/netdata/pull/6567#issuecomment-516577844

agronick:
I signed it.

thiagoftsm:
Hi @agronick ,

Codacy is giving an error in your code, it is saying the following:

"
Method could be a function

def _build_job(self, job):"

You can see more details in the link https://app.codacy.com/app/netdata/netdata/pullRequest?prid=3934476&bid=13662897

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
+# Description: dovecot netdata python.d module
+# Author: Kyle Agronick (agronick)
+# SPDX-License-Identifier: GPL-3.0+
+
+# Gearman Netdata Plugin
+
+from bases.FrameworkServices.SocketService import SocketService
+from copy import deepcopy
+
+update_every = 3
```
ilyam8: Why `3`? :thinking:

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
...
+priority = 60500
+
+
+CHARTS = {
+    'total_workers': {
+        'options': ['total', 'Total Workers', 'Workers', 'Total Workers', 'gearman.total', 'line'],
```
ilyam8: > gearman.total

keep in mind that context used as a metric name when exporting to prometheus for example, so i think we need to change the name

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
...
+            ['total_queued', 'Queued', 'absolute'],
+            ['total_running', 'Running', 'absolute'],
+        ]
+    },
+}
+
+
+def job_chart_template(job_name):
+    return {
+        'options': [job_name, job_name, 'Workers', 'Jobs', 'gearman.job', 'stacked'],
```
ilyam8: same for context

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
...
+
+        found_jobs = set()
+
+        for job in self._get_worker_data():
+            job_name = job[0]
+
+            # Gearman does not clean up old jobs
+            # We only care about jobs that have
+            # some relevant data
+            if any(job[1:]):
```
ilyam8: please do 

```python
if not any(job[1:]):
    continue

...
...
...

```

> jobs that have some relevant data

i suggest to create a var, because it is not clear what is `job[1:]`

```python
job_name = job[0]
job_raw_data = job[1:]
```

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
...
+
+                job_data = self._build_job(job)
+                output.update(job_data)
+                found_jobs.add(job_name)
+
+                for sum_value in ('queued', 'running', 'idle'):
+                    output['total_{0}'.format(sum_value)] += job_data['{0}_{1}'.format(job_name, sum_value)]
+
+        for to_remove in self.active_jobs - found_jobs:
+            self._remove_chart(to_remove)
```
ilyam8: i think it is better to split the logic

1. get data
2. parse data (add job stats, summary stats)
3. update charts (add(new jobs), remove(stale jobs))
4. update active jobs

what do you think?

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
...
+
+        if found_jobs:
+            return output
+
+    def _get_worker_data(self):
```
ilyam8: Please dont make method protected. This is true for all methods. `_get_data` is a bad example.

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
...
+        except (ValueError, AttributeError):
+            return
+
+        if raw is None:
+            self.debug("Gearman returned no data")
+            return
+
+        for line in sorted([job.split() for job in raw.splitlines()][:-1], key=lambda x: x[0]):
+            line[1:] = map(int, line[1:])
+            yield line
```
ilyam8: i suggest do not use generator, just return None or list.

> sorted([job.split() for job in raw.splitlines()][:-1], key=lambda x: x[0])

doesnt look cool, please humanize it. For example i would suggest to make a function with a good name and move details there.

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
...
+        :return: None
+        """
+
+        job_key = 'job_{0}'.format(job_name)
+        template = job_chart_template(job_name)
+        new_chart = self.charts.add_chart([job_key] + template['options'])
+        for dimension in template['lines']:
+            new_chart.add_dimension(dimension)
+
+        self.active_jobs.add(job_name)
```
ilyam8: i dont think you need to add something to `active_jobs` in  `add_chart`. According the name you only need to add a chart there.

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,157 @@
...
+    def _remove_chart(self, job_name):
+        """
+        Removes a job chart
+        :param job_name: The name of the job to remove
+        :type job_name: string
+        :return: None
+        """
+
+        job_key = 'job_{0}'.format(job_name)
+        self.charts[job_key].obsolete()
```
ilyam8: delete the chart from the charts after obsolete, i dont think we need to keep it in the charts

ilyam8:
Hey @agronick. Are you going to finish the PR?

collectors/python.d.plugin/gearman/gearman.chart.py
agronick: Or this.

collectors/python.d.plugin/gearman/gearman.chart.py
agronick: I'll removed it. Its arbitrary. It can be anything.

agronick:
Sorry, I was looking for updates in my feed. Must of missed them when they came through. Will make the changes today.

collectors/python.d.plugin/gearman/gearman.chart.py
agronick: I will try to break it out some more without looping more than twice.

collectors/python.d.plugin/gearman/gearman.chart.py
ilyam8: let me give you an example

https://github.com/netdata/netdata/blob/2fbc25e1b08959e49e9ece4e816f2ba41a863f84/collectors/python.d.plugin/sensors/sensors.chart.py#L23-L28

context is `sensors.temperature`, as you can see it makes sense, `gearman.total` is not something clear, total what?

Back to `sensors.temperature`. Keep in mind that netdata is an agent and people use it to collect metrics and export them to some db. Netdata export metrics.

```
netdata_sensors_temperature_Celsius_average{chart="sensors.coretemp_isa_0000_temperature",family="temperature",dimension="Core 0"} 36.2000000 1565853021000
netdata_sensors_temperature_Celsius_average{chart="sensors.coretemp_isa_0000_temperature",family="temperature",dimension="Core 1"} 36.8000000 1565853021000
netdata_sensors_temperature_Celsius_average{chart="sensors.coretemp_isa_0000_temperature",family="temperature",dimension="Core 2"} 35.8000000 1565853021000
netdata_sensors_temperature_Celsius_average{chart="sensors.coretemp_isa_0000_temperature",family="temperature",dimension="Core 3"} 36.2000000 1565853021000
netdata_sensors_temperature_Celsius_average{chart="sensors.coretemp_isa_0000_temperature",family="temperature",dimension="Core 4"} 36.0000000 1565853021000
netdata_sensors_temperature_Celsius_average{chart="sensors.coretemp_isa_0000_temperature",family="temperature",dimension="Core 5"} 35.1000000 1565853021000
```

The core element in metrics name is context - `sensors_temperature`. So this is crucial to have good context, because the context is metric name actually.

ilyam8: ok, in your case metric name will be

`netdata_gearman_total_Workers`

I [see](https://prometheus.io/docs/practices/naming/#metric-names) the valid order is
 - agent prefix
 - app prefix
 - metric name
 - metric unit

In your case `total` is metric name, that is why i am complaining :sweat_smile:

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,218 @@
+# Description: dovecot netdata python.d module
+# Author: Kyle Agronick (agronick)
+# SPDX-License-Identifier: GPL-3.0+
+
+# Gearman Netdata Plugin
+
+from bases.FrameworkServices.SocketService import SocketService
+from copy import deepcopy
+
+priority = 60500
```
ilyam8: If there is no reason to not use default priority i suggest to remove it.

ilyam8: The priority thing is uses for order on the dashboard. If all collectors use default priority then alphabetical order will be used.

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,218 @@
...
+        found_jobs, job_data = self.process_jobs(active_jobs)
+        self.remove_stale_jobs(found_jobs)
+
+        if job_data:
+            return job_data
+
+    def get_active_jobs(self):
+        active_jobs = []
+
+        for job in self.get_worker_data():
```
ilyam8: `get_worker_data` can return `None`, in that case `for job in self.get_worker_data(): wil raise an Exception

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,218 @@
...
+        Format data received from socket
+        :return: dict
+        """
+
+        active_jobs = self.get_active_jobs()
+        found_jobs, job_data = self.process_jobs(active_jobs)
+        self.remove_stale_jobs(found_jobs)
+
+        if job_data:
+            return job_data
```
ilyam8: i see it is always `job_data`, at least we always have

```
        output = {
            'total_pending': 0,
            'total_idle': 0,
            'total_running': 0,
        }
```

collectors/python.d.plugin/gearman/gearman.chart.py
agronick: Ah, yeah that only works if its an iterator.

collectors/python.d.plugin/gearman/gearman.chart.py
agronick: You're right. Should have caught that one.

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,211 @@
...
+        try:
+            raw = self._get_raw_data()
+        except (ValueError, AttributeError):
+            return []
+
+        if raw is None:
+            self.debug("Gearman returned no data")
+            return []
+
+        job_lines = raw.splitlines()[:-1]
```
ilyam8: please add a comment with an example output from `_get_raw_data()`

collectors/python.d.plugin/gearman/gearman.chart.py
ilyam8: use `return job_data or None`, it is more clear then implicit return

ilyam8: btw i think that 
```
        output = {
            'total_pending': 0,
            'total_idle': 0,
            'total_running': 0,
        }
```

is ok, otherwise there will be no metrics if there is no jobs

ilyam8:
@agronick 
about total counters, we should return `0` if there is no jobs, we should return `None` if there is an error in `get_worker_data`

collectors/python.d.plugin/gearman/gearman.chart.py
agronick: Don't you want no metrics with no jobs?

collectors/python.d.plugin/gearman/gearman.chart.py
agronick: I was seeing the different Gearman servers in different order depending on the name. Is this preferred or is there a different way to suppress that?

collectors/python.d.plugin/gearman/gearman.chart.py
ilyam8: No no, no job - zero total metrics, no metrics only in case of error

collectors/python.d.plugin/gearman/gearman.chart.py
ilyam8: > different Gearman

Not sure i understood you correctly, this is the order on the dashboard. 

(almost) all python modules have default priority - default == netdata will use alphabetical.

```
 - system
 ... 
 - apache job(s)
 ...
 - gearmand job(s)
 ...
 - sensors job(s)
```

collectors/python.d.plugin/gearman/gearman.chart.py
```diff
@@ -0,0 +1,228 @@
...
+        self.order = ['total_workers']
+
+    def _get_data(self):
+        """
+        Format data received from socket
+        :return: dict
+        """
+
+        try:
+            active_jobs = self.get_active_jobs()
```
ilyam8: `GearmanReadException` can be raised only in `get_active_jobs`, please do not add other func into `try/except` block

ilyam8:
@agronick please verify that code works with both py2 and py3 and it survives service restart (stop gearmand service; wait; start gearmand service or something like that)

agronick:
I've tested it with Python 2.7.5 and 3.6. Both ran without issues. I've been running it for weeks connected to 3 different servers. It has handled restarts as well as jobs being added and removed without issue.

ilyam8:
> your alarm and conf file are not being installed in the correct place

I dont confirm it, i installed the PR and all files are in place.

> test@debian10:/opt/netdata/usr/libexec/netdata/plugins.d$ locate gearman.conf
/home/test/src/gearman/collectors/python.d.plugin/gearman/gearman.conf
/home/test/src/gearman/health/health.d/gearman.conf
/opt/netdata/usr/lib/netdata/conf.d/health.d/gearman.conf
/opt/netdata/usr/lib/netdata/conf.d/python.d/gearman.conf

thiagoftsm:
> > your alarm and conf file are not being installed in the correct place
> 
> I dont confirm it, i installed the PR and all files are in place.
> 
> > test@debian10:/opt/netdata/usr/libexec/netdata/plugins.d$ locate gearman.conf
> > /home/test/src/gearman/collectors/python.d.plugin/gearman/gearman.conf
> > /home/test/src/gearman/health/health.d/gearman.conf
> > /opt/netdata/usr/lib/netdata/conf.d/health.d/gearman.conf
> > /opt/netdata/usr/lib/netdata/conf.d/python.d/gearman.conf

I will test in another VM to confirm.

thiagoftsm:
Hi @agronick ,

Confirming what @ilyam8 said, I found a problem with the `netdata-installer.sh`, your PR is installing the files correctly.

Best regards,

agronick:
Ok, cool. Sounds good.

ilyam8:
@cakrit @cosmix @joelhans please review, the PR cant be merged w/o your approvement

ilyam8:
@agronick good job 👍 thanks!

PR #6578, opened 2019-07-31
netdata/systemd: Add slack notification mechanism on netdata.service

paulkatsoulakis:
##### Summary
To get ourselves promptly notified on failing services, we added the support for slack notifications when netdata service dies and runs with systemd.

To set this mechanism up, all you need to do is do a `sudo systemctl edit netdata` and add the following line:
```
[Service]
Environment=NETDATA_SYSTEMD_SLACK_NOTIFY_URL=[THE_SLACK_INCOMING_WEBHOOK_URL_HERE]
```

##### Component Name
netdata/installer

##### Additional Information
none


ilyam8:
Do we need some timeouts for curl in `slack.sh`? I noticed there is no timeouts. It can tike forever in some cases.

ktsaou:
Guys, we are a monitoring system. What is this PR?

Configure netdata at london3 to monitor all the netdata demo sites.
Repeat that process for atlanta to do the same.

cosmix:
@ktsaou it is good practice — for fault tolerance — not to use the same system to monitor itself. I asked Paul to use systemd to monitor netdata and send an alert if it fails, following recent failures that might have been systemic.

paulkatsoulakis:
If i may say, i kind of agree with @cosmix, circular dependencies are naughty things.
On the other hand i do hear the perspective from Costa, but we always have a netdata that may not be seen by another netdata when it goes down.

I think @ktsaou and @cosmix can discuss internally and let me know what would our strategy be in this case.

paulkatsoulakis:
After discussing this issue internally, I will be cancelling this PR and proceed with a new more netdata-oriented approach

PR #6589, opened 2019-08-02
Fix configure.ac allowing deprecated LZ4 library call

### LINKED ISSUE #6580: netdata-installer.sh exited with error
##### Bug report summary
netdata-installer.sh exited with error 
cciV04OR.ltrans10.o:(.text+0x3ca3): undefined reference to `LZ4_compress_default'

##### OS / Environment
Linux 3.16.0-9-amd64 #1 SMP Debian 3.16.68-2 (2019-06-17) x86_64 GNU/Linux

Daily build using bash <(curl -Ss https://my-netdata.io/kickstart.sh)
Aug 01 2019

mfundul:
Fix configure.ac allowing deprecated LZ4 library call.  Fixes #6580

Previously it was possible for older versions of LZ4 libraries to succesfully pass the ./configure phase of the build process even though they shouldn't.

By requiring the most recent API call of LZ4 that is being used by netdata, which is LZ4_compress_default(), we eliminate the risk of a false positive LZ4 flag in the ./configure phase of the build.

As a result, any system with a deprecated LZ4 library will build netdata without LZ4 support (and subsequently without dbengine support).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -283,7 +283,7 @@ OPTIONAL_UV_LIBS="${UV_LIBS}"
 
 AC_CHECK_LIB(
     [lz4],
-    [LZ4_decompress_safe],
+    [LZ4_compress_default],
     [LZ4_LIBS="-llz4"]
 )
 


thiagoftsm:
I tested  the PR on Debian 8 and I verified that the Netdata 8 was not linked with the Netdata after this change.

PR #6595, opened 2019-08-05
HTTP response message

### LINKED ISSUE #6559: HTTP errors inside web api more readable.
##### Feature idea summary
We had a PR that created pre processors/defintions for the response code, but this was not applied to all the code yet, it is necessary to move this pattern for the whole web server code, because we are working with two different patterns now.
##### Expected behavior
When the developers read the code, they must find only one pattern for HTTP response.

thiagoftsm:
##### Summary
When we merged an PR in the last sprint, Netdata has a new pattern for the file web_client.c, every time it returned an HTTP error, it did not have explicit the error number, the codes was changed for the error name, but this was not carry for all the Netdata web files. This PR fixes this creating again one unique pattern to give the error.
##### Component Name
Web
##### Additional Information
Fixes #6559

### DIFF
--- a/web/api/health/health_cmdapi.c
+++ b/web/api/health/health_cmdapi.c
@@ -116,7 +116,7 @@ void health_silencers2file(BUFFER *wb) {
  * @return It returns 200 on success and another code otherwise.
  */
 int web_client_api_request_v1_mgmt_health(RRDHOST *host, struct web_client *w, char *url) {
-    int ret = 400;
+    int ret;
     (void) host;
 
     BUFFER *wb = w->response.data;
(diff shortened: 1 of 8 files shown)


web/api/health/health_cmdapi.c
```diff
@@ -116,7 +116,7 @@ void health_silencers2file(BUFFER *wb) {
  * @return It returns 200 on success and another code otherwise.
  */
 int web_client_api_request_v1_mgmt_health(RRDHOST *host, struct web_client *w, char *url) {
-    int ret = 400;
+    int ret;
```
mfundul: This used to be `int ret = 400;` why did you remove the assignment?

thiagoftsm: I removed it, because it is never returned, as you can see in the whole function, it is set `ret = 403` on lines 134 and 139 and we also set `ret = 200` at line 193, so I believe this function was changed during the time and this `ret = 400` is there due the first versions, because the latest version is overwritten the values before the function ends.

PR #6606, opened 2019-08-06
netdata/installer: do not enable netdata, if it was previously disabled

### LINKED ISSUE #6601: netdata re-enables itself after being disabled
##### Bug report summary

netdata re-enables itself and starts on next reboot after being disabled and stopped

##### OS / Environment

ubuntu 18.04 lts

netdata v1.16.1

##### Steps To Reproduce

1. install netdata with `bash <(curl -Ss https://my-netdata.io/kickstart.sh)`
2. stop and disable netdata `systemctl disable netdata` and `systemctl stop netdata`
3. confirm netdata is dead and disabled `systemctl status netdata`
4. change the checksum in /var/lib/netdata/netdata.tarball.checksum to force an update
5. run /usr/libexec/netdata/netdata-updater.sh
6. check netdata is still dead and disabled `systemctl status netdata`

If netdata-updater finds a new version, it runs `./netdata-installer.sh --auto-update  --dont-wait --dont-start-it` 

The new version is installed and it's not started, however `systemctl status netdata` show's it's now enabled, and netdata will start on the next reboot.

##### Expected behavior

If I stop and disable netdata, it stays stopped and disabled.

ps. I find netdata very useful and thank you to the creator(s) and the community for their work.\

paulkatsoulakis:
##### Summary
Fixes #6601 
Basically we do not want to enable netdata, if it was previously disabled

##### Component Name
netdata/installer

##### Additional Information
None

### DIFF
--- a/packaging/installer/functions.sh
+++ b/packaging/installer/functions.sh
@@ -380,10 +380,17 @@ install_netdata_service() {
 			fi
 
 			if [ "${SYSTEMD_DIRECTORY}x" != "x" ]; then
+				ENABLE_NETDATA_IF_PREVIOUSLY_ENABLED="run systemctl enable netdata"
+				IS_NETDATA_ENABLED=`systemctl is-enabled netdata 2> /dev/null || echo "Netdata not there"`
+				if [ "${IS_NETDATA_ENABLED}" == "disabled" ]; then
+					echo >&2 "Netdata was there and disabled, make sure we don't re-enable it ourselves"
+					ENABLE_NETDATA_IF_PREVIOUSLY_ENABLED="echo"
+				fi
+
 				echo >&2 "Installing systemd service..."
 				run cp system/netdata.service "${SYSTEMD_DIRECTORY}/netdata.service" &&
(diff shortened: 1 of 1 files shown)


packaging/installer/functions.sh
```diff
@@ -380,10 +380,17 @@ install_netdata_service() {
 			fi
 
 			if [ "${SYSTEMD_DIRECTORY}x" != "x" ]; then
+				ENABLE_NETDATA_IF_PREVIOUSLY_ENABLED="run systemctl enable netdata"
+				IS_NETDATA_ENABLED=`systemctl is-enabled netdata 2> /dev/null || echo "Netdata not there"`
+				if [ "${IS_NETDATA_ENABLED}" == "disabled" ]; then
+					echo >&2 "Netdata was there and disabled, make sure we don't re-enable it ourselves"
+					ENABLE_NETDATA_IF_PREVIOUSLY_ENABLED="echo"
```
cosmix: I'd use `true` here, instead of `echo`.

PR #6607, opened 2019-08-06
update package version requirements for LZ4 and libuv

mfundul:
update package version requirements for LZ4 and libuv


mfundul:
> LGTM for the most part, but let's start enforcing some standard capitalization in this section, which we'll eventually bring to the other tables, too. Just capitalizing the first word in the description, and nothing else.
Agreed, the descriptions were taken from the package maintainers so there was no consistent style.

PR #6552, opened 2019-07-26
netdata/packaging: Documentation on distribution support matrix and functionality availability

paulkatsoulakis:
#### Summary
With this PR we introduce the required documentation that will give us sufficient visibility on key aspects like:
1) Distributions we officially support and commit that netdata must properly running there
2) Details on the various core and peripheral functionalities and how those are available through the different distribution channels available

This is a initial idea of the document, we expect to further modify and adapt based on internal and external (user) feedback.

As a side fix, we also added two more distributions to be built on our pipeline that was missed: Ubuntu/Xenial and Ubuntu/Trusty.  There were a set of minor fixes too in order to fully support execution of those two builds

##### Component Name
netdata/packaging/documentation

##### Additional Information
Fixes #6036 
Fixes #6300 
Fixes #6610

PR #6609, opened 2019-08-06
BuildRequires protobuf-devel >= 3

stormi:
The version of protobuf-devel that is required at build time must be >=3
because version 2 can't handle the proto files created for version 3.

##### Summary
The spec file doesn't build with protobuf-devel 2. Enforcing version >= 3 as a BuildRequires will make it more obvious.

### DIFF
--- a/netdata.spec.in
+++ b/netdata.spec.in
@@ -193,7 +193,7 @@ Requires: cups
 
 # Prometheus remote write dependencies
 BuildRequires: snappy-devel
-BuildRequires: protobuf-devel
+BuildRequires: protobuf-devel >= 3
 %if 0%{?suse_version}
 BuildRequires: libprotobuf-c-devel
 %else
@@ -207,7 +207,7 @@ Requires: libprotobuf15
 %else
 Requires: snappy
 Requires: protobuf-c
-Requires: protobuf
+Requires: protobuf >= 3
 %endif
 # end - prometheus remote write dependencies
 


stormi:
Looks like it breaks the build for CentOS 6 and 7... But I had to upgrade protobuf to a version that can handle protocol version 3, in order to be able to build on CentOS 7. This is what brought me here. 

So maybe there's a way to build for CentOS 7 with protobuf 2 after all?

paulkatsoulakis:
Hello! This has been included on #6552 but the PR was late to be merged.
Apologies for the late response on this.
I will close this and kindly follow up resolution after merge of 6552

stormi:
Looking at that other PR, I don't see any changes to the spec file. So the build will fail earlier but still fail despite the installation of all build dependencies described by `BuildRequires`. Shouldn't my change be accepted in addition to that other PR?

paulkatsoulakis:
> Looking at that other PR, I don't see any changes to the spec file. So the build will fail earlier but still fail despite the installation of all build dependencies described by `BuildRequires`. Shouldn't my change be accepted in addition to that other PR?

We are striving to keep things as simple as possible.
We are building multiple distributions using the same spec file to reduce complexity, at least at this stage. This PR in order to pass, it must allow us to successfully build RPM and also be able to install the generated RPM on the following distributions 

RHEL 6,7,8
CentOS 6,7,8
Open SuSE 15.0, 15.1
Open SuSE Tumbleweed
Fedora 28,29,30
SLES 11, 12, 15

The other thing is also that our dependency graph needs to maintain some consistency with our .DEB package support. So things are getting a bit more complicated with this addition.

The PR i mentioned, only mitigates the build failure.
we still face the risk of losing functionality on specific distributions due to the lack of the appropriate package availability.

We plan to re-iterate separately on how to ensure our RPM packages contain all of our capabilities on all distributions on an efficient and reliable way

stormi:
Ok, I think I understand: current mitigation allows the build to pass by properly detecting that protobuf 3 is not available at configure stage, at the cost of reduced functionality. I was not sure if protobuf was a strong requirement or not, and it looks like it isn't. 

Closing. Thanks!

paulkatsoulakis:
Not to mention that of course any insights on this are most welcome

paulkatsoulakis:
> Ok, I think I understand: current mitigation allows the build to pass by properly detecting that protobuf 3 is not available at configure stage, at the cost of reduced functionality. I was not sure if protobuf was a strong requirement or not, and it looks like it isn't.
> 
> Closing. Thanks!

Well, that is the thing.  We really want all functionality available for our users, but some things are not hard requirements.

Protobuf package specifically is needed by [prometheus remote write](https://docs.netdata.cloud/packaging/distributions/#prometheus-remote-write) backend, which is not really needed for all of our users and the ones who really want it could settle with a distribution that can support it.

We do strive to find solution for all supported distributions, but this is a long standing tough battle.
We do rely on our users feedback a lot, so any insights are welcome

Issue #6054, opened 2019-05-20
Add alarm variables to the response of chart and data


cakrit:
`api/v1/info/alarm_variables?chart=XX` returns the variables available for creation of alarms.

We have a request to add them to both `/chart` and `/data` call responses as well.

_Originally posted by @ilyam8 in https://github.com/netdata/netdata/issues/5990#issuecomment-491498124_

alpes214:
@ilyam8  @cakrit 
The /alarm_variables?chart=XX request has chart_variables field in its response. 
- same field with the same name and values should be added to the response for /chart?chart=XX request.
- the /data request has many different response formats, some of them like csv cannot be extended easily. Therefor chart_variables should be added to response when /data request is called with options=jsonwrap parameter only: /data?chart=XX&options=jsonwrap

I doubt about /data request.
@ilyam8 is my understanding correct?

ilyam8:
> same field with the same name and values should be added to the response for /chart?chart=XX request.

Not sure.

I have

```python
CHARTS = {
    'random': {
        'options': [None, 'A random number', 'random number', 'random', 'random', 'line'],
        'lines': [
            ['random1']
        ],
        'variables': [
            ['var_1', 999],
            ['var_2', 999],
            ['var_3', 999]
        ]
    }
}
```

There you can see 3 chart variables.

I did `/alarm_variables?chart=example.random`. My `chart_variables`:

```json
{
    "var_1": 999,
    "var_3": 999,
    "var_2": 999,
    "random2_raw": 38,
    "red": null,
    "update_every": 1,
    "random3_raw": 49,
    "random1_raw": 31,
    "random1_last_collected_t": 1565088357.00000,
    "random2_last_collected_t": 1565088357.00000,
    "last_collected_t": 1565088357.00000,
    "random1": 0,
    "random2": 0,
    "random3": 0,
    "green": null,
    "random3_last_collected_t": 1565088357.00000,
    "collected_total_raw": 118
}
```

i think we need only `var_1`, `var_2` and `var_3` from there.

ilyam8:
Maybe it is not even possible to distinguish between user defined chart vars and internally created, it looks like it is our case :sweat_smile: If so then yes

> same field with the same name and values should be added to the response for /chart?chart=XX request.

ilyam8:
> Maybe it is not even possible to distinguish between user defined chart vars and internally created

It is possible

[...]

PR #6615, opened 2019-08-08
Add alarm variables to the response of chart and data (#6054)

alpes214:
##### Summary
Implements feature #6054 

Now requests like 

http://localhost:19999/api/v1/chart?chart=example.random
http://localhost:19999/api/v1/data?chart=example.random&options=jsonwrap&options=showcustomvars

- return chart variables in their responses. Chart variables include only those with options set to RRDVAR_OPTION_CUSTOM_CHART_VAR
- for /api/v1/data requests chart variables are returned when parameter options=jsonwrap and options=showcustomvars

##### Component Name
[/database](https://github.com/netdata/netdata/tree/master/database/)
[/web/api/formatters](https://github.com/netdata/netdata/tree/master/web/api/formatters)

##### Additional Information
To test functionality the collectors/python.d.plugin/example chart has been configured to have the following variables:

```python
CHARTS = {
    'random': {
        'options': [None, 'A random number', 'random number', 'random', 'random', 'line'],
        'lines': [
            ['random1']
        ],
        'variables': [
            ['var_1', 999],
            ['var_2', 999],
            ['var_3', 999],
            ['var_4', 'abc']
        ]
    }
}
```
After that responses for requests 
```bash
curl http://localhost:19999/api/v1/data?chart=example.random&options=jsonwrap&options=showcustomvars
curl http://localhost:19999/api/v1/chart?chart=example.random
```
compared with response from request 

```bash
curl http://localhost:19999/api/v1/alarm_variables?chart=example.random
```

Expected: /api/v1/chart and /api/v1/data requests should contain same custom chart variables with same values as in /api/v1/alarm_variables

### DIFF
--- a/database/rrdvar.c
+++ b/database/rrdvar.c
@@ -248,18 +248,19 @@ int health_variable_lookup(const char *variable, uint32_t hash, RRDCALC *rc, cal
 struct variable2json_helper {
     BUFFER *buf;
     size_t counter;
+    RRDVAR_OPTIONS options;
 };
 
-static int single_custom_variable2json(void *entry, void *data) {
+static int single_variable2json(void *entry, void *data) {
     struct variable2json_helper *helper = (struct variable2json_helper *)data;
     RRDVAR *rv = (RRDVAR *)entry;
     calculated_number value = rrdvar2number(rv);
 
-    if(rv->options & RRDVAR_OPTION_CUSTOM_CHART_VAR) {
+    if (helper->options == RRDVAR_OPTION_DEFAULT || rv->options & helper->options) {
         if(unlikely(isnan(value) || isinf(value)))
-            buffer_sprintf(helper->buf, "%s\n\t\t\t\t\"%s\": null", helper->counter?",":"", rv->name);
+            buffer_sprintf(helper->buf, "%s\n\t\t\"%s\": null", helper->counter?",":"", rv->name);
         else
-            buffer_sprintf(helper->buf, "%s\n\t\t\t\t\"%s\": %0.5" LONG_DOUBLE_MODIFIER, helper->counter?",":"", rv->name, (LONG_DOUBLE)value);
+            buffer_sprintf(helper->buf, "%s\n\t\t\"%s\": %0.5" LONG_DOUBLE_MODIFIER, helper->counter?",":"", rv->name, (LONG_DOUBLE)value);
 
         helper->counter++;
     }
@@ -267,29 +268,15 @@ static int single_custom_variable2json(void *entry, void *data) {
     return 0;
 }
 
-static int single_variable2json(void *entry, void *data) {
-    struct variable2json_helper *helper = (struct variable2json_helper *)data;
-    RRDVAR *rv = (RRDVAR *)entry;
-    calculated_number value = rrdvar2number(rv);
-
-    if(unlikely(isnan(value) || isinf(value)))
-        buffer_sprintf(helper->buf, "%s\n\t\t\"%s\": null", helper->counter?",":"", rv->name);
-    else
-        buffer_sprintf(helper->buf, "%s\n\t\t\"%s\": %0.5" LONG_DOUBLE_MODIFIER, helper->counter?",":"", rv->name, (LONG_DOUBLE)value);
-
-    helper->counter++;
-
-    return 0;
-}
-
 void health_api_v1_chart_custom_variables2json(RRDSET *st, BUFFER *buf) {
     struct variable2json_helper helper = {
             .buf = buf,
-            .counter = 0
+            .counter = 0,
+            .options = RRDVAR_OPTION_CUSTOM_CHART_VAR
     };
 
     buffer_sprintf(buf, "{");
-    avl_traverse_lock(&st->rrdvar_root_index, single_custom_variable2json, (void *)&helper);
+    avl_traverse_lock(&st->rrdvar_root_index, single_variable2json, (void *)&helper);
     buffer_strcat(buf, "\n\t\t\t}");
 }
 
@@ -298,7 +285,8 @@ void health_api_v1_chart_variables2json(RRDSET *st, BUFFER *buf) {
 
     struct variable2json_helper helper = {
             .buf = buf,
-            .counter = 0
+            .counter = 0,
+            .options = RRDVAR_OPTION_DEFAULT
     };
 
     buffer_sprintf(buf, "{\n\t\"chart\": \"%s\",\n\t\"chart_name\": \"%s\",\n\t\"chart_context\": \"%s\",\n\t\"chart_variables\": {", st->id, st->name, st->context);
--- a/database/rrdvar.c
+++ b/database/rrdvar.c
@@ -250,6 +250,23 @@ struct variable2json_helper {
     size_t counter;
 };
 
+static int single_custom_variable2json(void *entry, void *data) {
+    struct variable2json_helper *helper = (struct variable2json_helper *)data;
+    RRDVAR *rv = (RRDVAR *)entry;
+    calculated_number value = rrdvar2number(rv);
+
+    if(rv->options & RRDVAR_OPTION_CUSTOM_CHART_VAR) {
+        if(unlikely(isnan(value) || isinf(value)))
+            buffer_sprintf(helper->buf, "%s\n\t\t\t\t\"%s\": null", helper->counter?",":"", rv->name);
+        else
+            buffer_sprintf(helper->buf, "%s\n\t\t\t\t\"%s\": %0.5" LONG_DOUBLE_MODIFIER, helper->counter?",":"", rv->name, (LONG_DOUBLE)value);
+    }
+
+    helper->counter++;
+
+    return 0;
+}
+
 static int single_variable2json(void *entry, void *data) {
     struct variable2json_helper *helper = (struct variable2json_helper *)data;
     RRDVAR *rv = (RRDVAR *)entry;
@@ -263,6 +280,17 @@ static int single_variable2json(void *entry, void *data) {
     helper->counter++;
 
     return 0;
+}
+
+void health_api_v1_chart_custom_variables2json(RRDSET *st, BUFFER *buf) {
+    struct variable2json_helper helper = {
+            .buf = buf,
+            .counter = 0
+    };
+
+    buffer_sprintf(buf, "{");
+    avl_traverse_lock(&st->rrdvar_root_index, single_custom_variable2json, (void *)&helper);
+    buffer_strcat(buf, "\n\t\t\t}");
 }
 
 void health_api_v1_chart_variables2json(RRDSET *st, BUFFER *buf) {
--- a/database/rrdvar.c
+++ b/database/rrdvar.c
@@ -260,9 +260,9 @@ static int single_custom_variable2json(void *entry, void *data) {
             buffer_sprintf(helper->buf, "%s\n\t\t\t\t\"%s\": null", helper->counter?",":"", rv->name);
         else
             buffer_sprintf(helper->buf, "%s\n\t\t\t\t\"%s\": %0.5" LONG_DOUBLE_MODIFIER, helper->counter?",":"", rv->name, (LONG_DOUBLE)value);
-    }
 
-    helper->counter++;
+        helper->counter++;
+    }
 
     return 0;
 }
--- a/health/health.h
+++ b/health/health.h
@@ -62,6 +62,7 @@ extern void health_alarms2json(RRDHOST *host, BUFFER *wb, int all);
 extern void health_alarm_log2json(RRDHOST *host, BUFFER *wb, uint32_t after);
 
 void health_api_v1_chart_variables2json(RRDSET *st, BUFFER *buf);
+void health_api_v1_chart_custom_variables2json(RRDSET *st, BUFFER *buf);
 
 extern int health_alarm_log_open(RRDHOST *host);
 extern void health_alarm_log_close(RRDHOST *host);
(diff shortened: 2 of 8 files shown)


alpes214:
Hi @thiagoftsm ,

Sorry, this is my fault. I did not check with /api/v1/charts. The fix is in the netdata repository. Please have a look. 

I am not sure how to resolve conversation, may be because I do not have rights to do so.

Regards!

ilyam8:
i changed example.chart.py charts to

```python
CHARTS = {
    'random': {
        'options': [None, 'A random number', 'random number', 'random', 'random', 'line'],
        'lines': [
            ['random1']
        ],
        'variables': [
            ['var_1', 999],
            ['var_2', 999],
            ['var_3', 'var_3'],
        ]
    }
}
```

> http://localhost:19999/api/v1/chart?chart=example.random

chart variables

```json
{
    "var_1": 999,
    "var_3": 0,
    "var_2": 999
}
```

> http://localhost:19999/api/v1/data?chart=example.random&options=jsonwrap

chart variables

```json
{
    "var_1": 999,
    "var_3": 0,
    "var_2": 999
}
```

alpes214:
@ilyam8 

I switched to master branch (original implementation), put your changes into example.chart.py

```python
        'variables': [
            ['var_1', 999],
            ['var_2', 999],
            ['var_3', 'var_3'],
        ]
```

and checked 
http://s89830:19999/api/v1/alarm_variables?chart=example.random

the response contains this 

```python
	"chart_variables": {
		"var_1": 999.00000,
		"var_3": 0.00000,
		"var_2": 999.00000,
	}
```
The var_3 value in /api/v1/chart and /api/v1/data responses is the same as in the /api/v1/alarm_variables from master branch. I though it is expected that /api/v1/chart and /api/v1/data return same chart_variables and values as /api/v1/alarm_variables. Please let me know if this assumption is not correct.

ilyam8:
I dont get your question. I didnt want to say that there is some problem https://github.com/netdata/netdata/pull/6615#issuecomment-519801792

Both repsonses are ok.

alpes214:
> 
> 
> I dont get your question. I didnt want to say that there is some problem [#6615 (comment)](https://github.com/netdata/netdata/pull/6615#issuecomment-519801792)
> 
> Both repsonses are ok.

@ilyam8 
The situation when string value is converted to 0 is not usual and I was not sure that this is correct behavior, even for original /api/v1/alarm_variables. But if both responses are ok, then thank you for confirmation.

ilyam8:
> The situation when string value is converted to 0 is not usual and I was not sure that this is correct behavior

https://docs.netdata.cloud/collectors/plugins.d/#variable

> The value is floating point (netdata used long double).

The "correct behaviour" is to fix `python.d.plugin`, it shouldnt send to netdata variables with wrong values.

ilyam8:
> The situation when string value is converted to 0 is not usual and I was not sure that this is correct behavior

I dont think this is correct too. If variable has no (correct) value we shouldnt set it to zero. Maybe there is some logic why this is needed i am not aware of.

ilyam8:
Documentation of the API to swagger should be updated

ilyam8:
@alpes214 

https://github.com/netdata/netdata/pull/6516#issuecomment-518660755

alpes214:
> @alpes214
> 
> [#6516 (comment)](https://github.com/netdata/netdata/pull/6516#issuecomment-518660755)

thank you @ilyam8 

should I create a new pull request?
or this one can be merged into master later?

database/rrdvar.c
```diff
@@ -250,6 +250,23 @@ struct variable2json_helper {
     size_t counter;
 };
 
+static int single_custom_variable2json(void *entry, void *data) {
```
cakrit: Contains copy of code. Big no-no. Please refactor to at least something similar to the following:

```c
static int single_variable2json_worker(void *entry, void *data, int custom) {
    ...
}

static int single_custom_variable2json(void *entry, void *data) {
 return single_variable2json_worker(entry, data, 1);
}

static int single_variable2json(void *entry, void *data) {
    return single_variable2json_worker(entry, data, 0);
}
```

alpes214: @cakrit 
I used a different approach for refactoring. Please let me know if it is not ok.

ktsaou:
I haven't read the code, but adding this to /API/v1/data may have a great impact to all chart refreshes. They will be slower and use more bandwidth.

It would be better to add an option to enable it for this method.

thiagoftsm:
Hi @alpes214 ,

To complement of @ktsaou  said, I was testing your code, and before I change the python.d example, I decided to run without any change to check the results, when I ran the following command:

$ curl -k https://localhost:19999/api/v1/chart?chart=example.random

I got the following answer:

 {
                        "id": "example.random",
                        "name": "example.random",
                        "type": "example",
                        "family": "random",
                        "context": "random",
                        "title": "A random number (example.random)",
                        "priority": 90000,
                        "plugin": "python.d.plugin",
                        "module": "example",
                        "enabled": true,
                        "units": "random number",
                        "data_url": "/api/v1/data?chart=example.random",
                        "chart_type": "line",
                        "duration": 253209,
                        "first_entry": 1565560508,
                        "last_entry": 1565813716,
                        "update_every": 1,
                        "dimensions": {
                                "random1": { "name": "random1" },
                                "random2": { "name": "random2" },
                                "random3": { "name": "random3" }
                        },
                        "chart_variables": {
                        },
                        "green": null,
                        "red": null,
                        "alarms": {
                                "example_alarm2": {
                                        "id": 1565815352,
                                        "status": "UNINITIALIZED",
                                        "units": "random number",
                                        "update_every": 2
                                },
                                "example_alarm1": {
                                        "id": 1565815351,
                                        "status": "UNINITIALIZED",
                                        "units": "random number",
                                        "update_every": 2
                                }
                        }
                }

Something fine. But, when I ran the second test bellow without to redirect to a file:

 curl  -k "https://localhost:19999/api/v1/data?chart=example.random&options=jsonwrap"

I received as the last line of the output the following message:

`curl: (18) transfer closed with 6307188 bytes remaining to read`

when I saw the message I decided to redirect for a file to see how many bytes it was trying to get, and I received this as output:

$ curl -o aa -k "https://localhost:19999/api/v1/data?chart=example.random&options=jsonwrap"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 9600k  100 9600k    0     0  40.4M      0 --:--:-- --:--:-- --:--:-- 40.4M

Without any doubt, your PR is working, but it can do download of a big sample of data, please, create an option that is default false and when an user move it to enable, I think it could be good to create an initial limit for the sample of data, unless the user really specify  that he wanna get everything available. 

Best regards!

ilyam8:
> curl: (18) transfer closed with 6307188 bytes remaining to read

@thiagoftsm i think this is not related to the PR.

The only thing PR does - it adds `chart_variables` section to the answer

> http://127.0.0.1:19999/api/v1/data?chart=example.random&options=jsonwram&options=jsonwrap

```json
{
    "var1": 1,
    "var2": 2
}
```

^^^

i realized that adding chart vars to the `data` call is not a right thing.

Reason - they are not `constants`, they are changed over time (we can't guarantee they not). In that case it doesnt make sense only the current value in the data call.

alpes214:
@thiagoftsm, @ilyam8 
> > curl: (18) transfer closed with 6307188 bytes remaining to read
> 
> @thiagoftsm i think this is not related to the PR.
> 
> The only thing PR does - it adds `chart_variables` section to the answer
> 
> > http://127.0.0.1:19999/api/v1/data?chart=example.random&options=jsonwram&options=jsonwrap
> 
> ```json
> {
>     "var1": 1,
>     "var2": 2
> }
> ```

Small addition... /api/v1/data request has parameter `after` which is used to specify how much data should be returned in response. The default value is -600 seconds. Minus means 600 seconds from now. The request with default value -600 executed:

```bash
val@s89830:~$ time curl -o aa -k "http://registry.my-netdata.io/api/v1/data?chart=example.random&options=jsonwrap&after=-600"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 31241    0 31241    0     0   374k      0 --:--:-- --:--:-- --:--:--  376k

real    0m0.092s
user    0m0.008s
sys     0m0.004s
```

The request executed with -200000 seconds:

```bash
val@s89830:~$ time curl -o aa -k "http://registry.my-netdata.io/api/v1/data?chart=example.random&options=jsonwrap&after=-200000"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 9958k    0 9958k    0     0  7911k      0 --:--:--  0:00:01 --:--:-- 7916k

real    0m1.271s
user    0m0.028s
sys     0m0.032s
```
The difference in execution time and response size if obvious. In both cases response contains chart variables. 

Chart variables are returned when options=jsonwrap only. There is another option `milliseconds` with response format very close to jsonwrap, but without chart variables. Here is a call example with `milliseconds` option for comparison:

```bash
val@s89830:~$ time curl -o aa -k "http://registry.my-netdata.io/api/v1/data?chart=example.random&options=milliseconds&after=-200000"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 10.2M    0 10.2M    0     0  9929k      0 --:--:--  0:00:01 --:--:-- 9938k

real    0m1.075s
user    0m0.028s
sys     0m0.036s
```

It looks like: 1) `after` and `options` parameters may be used to tune response size and time; 2) the chart variables do not influence significantly on response. There could be an issue if user will create a huge number of variables (megabytes), but I am not sure that this is possible in working/production system.

> 
> ^^^
> 
> i realized that adding chart vars to the `data` call is not a right thing.
> 
> Reason - they are not `constants`, they are changed over time (we can't guarantee they not). In that case it doesnt make sense only the current value in the data call.

@ilyam8 does it mean, that I should remove chart variables implementation for /api/v1/data request?

ilyam8:
@alpes214 no, i am just thinking out loud. As i said we need to understand - are custom chart vars part of chart data, if so then they should be there, if no then not. I am sure they are (that is why i crossed out that part in my prev comment). 

And i was wrong, they are `constant` according [docs](https://docs.netdata.cloud/collectors/plugins.d/#variable)

> VARIABLE defines a variable that can be used in alarms. This is to used for setting constants (like the max connections a server may accept).

So all is ok :+1:

cakrit:
Just an optional param like showcustomvars (no/yes) to both requests, with default = no and let's finish this. Indeed the data request response time is related mainly to the number of data points requested, so that's not a good performance tests. @ktsaou's point regarding the charts requests is valid, there will be an effect, with little benefit to the users (not really sure who needs to use just custom vars and for which purpose and very few charts have them, but we need to iterate for every chart anyway).

ktsaou:
The crucial call is the data one. This is executed for every single chart refresh.

To measure the impact of something you should stress test netdata with and without this code. There are many tools that do this. You should use the same API option in both tests, as the most expensive operation is the  generation of the text output of the data points.

We need the data call to be the fastest possible.

alpes214:
> 
> 
> Just an optional param like showcustomvars (no/yes) to both requests, with default = no and let's finish this. Indeed the data request response time is related mainly to the number of data points requested, so that's not a good performance tests. @ktsaou's point regarding the charts requests is valid, there will be an effect, with little benefit to the users (not really sure who needs to use just custom vars and for which purpose and very few charts have them, but we need to iterate for every chart anyway).

I have implemented a bit different approach, i.e. I did not add additional parameter `showcustomvars` but added new option with the same name. The URL to request data with custom variables will look like this 
```bash
curl "http://localhost:19999/api/v1/data?chart=example.random&after=-2&options=jsonwrap&options=showcustomvars"
```
I believe this way we will not overload request with additional parameters. Please le me know if this is not ok.

Regarding stress test of the /api/v1/data request. This task requires thorough planning and will take some time. It would be good to create a separate issue for this task if possible.

cakrit:
FYI for stress tests, there's a k6 config in tests/k6/data.js 
We built it for dbengine tests though, so populating so much data for the example chart is not part of the script. With the addition of this option it won't be required.

alpes214:
> 
> 
> Hold on, for some reason the UI doesn't draw the charts any more.
> I thought initially it was something wrong with my installation, but the master branch has the same issue. I'm doing some more tests and will come back. It doesn't seem related to these changes, I do get the data request responses properly...

there is a problem with my change. I am going to commit a fix.

alpes214:
> 
> 
> Invalid JSON returned.
> 
> https://localhost:19999/api/v1/data?chart=system.cpu&format=json&points=493&group=average&gtime=0&options=ms%7Cflip%7Cjsonwrap%7Cnonzero&after=-20
> 
> ```
>    "format": "json,
>    "result": {
> ```
I think it is fixed.

cakrit:
Did you close this by mistake? Are you able to reopen or do you want me to do it?

alpes214:
> 
> 
> Did you close this by mistake? Are you able to reopen or do you want me to do it?

yes, please reopen it.

I tried to resolve requested change (conversation) but closed this PR. It looks like I do not have permissions on netdata repository to resolve converstations.

ilyam8:
Anyway conversation should be closed by the one who started it.

alpes214:
> There's a stray double quote when we do show custom vars now:
> 
> https://localhost:19999/api/v1/data?chart=system.idlejitter&format=json&points=493&group=average&gtime=0&options=ms%7Cflip%7Cjsonwrap%7Cnonzero%7Cshowcustomvars&after=-20
> 
> ```
> ...
>    "chart_variables": {
> 			}",
>    "result": {
> ```

tests/stress.sh helped to generate multiple /api/v1/data requests and to test this PR for different formats and options. @cakrit, I think the issue is fixed. Please have a look.

Issue #6616, opened 2019-08-08
MySQL plugin not working after last update


badcrc:
##### Bug report summary

After this night's update I the mysql plugin don't work, seems to be related to this PR: https://github.com/netdata/netdata/commit/a22b7e9a8112481aac83343473eb87f9d678ecd9

The user I have configured for mysql monitoring don't have permissions for doing a `show slave` command because the server doesn't have a mysql replication (it's a percona cluster with galera replication.

##### OS / Environment

CentOS 7 / MySQL 5.7.25-28-57-log Percona XtraDB Cluster

##### Netdata version (ouput of `netdata -V`)

netdata v1.16.1-10-nightly

##### Component Name

mysql plugin

##### Steps To Reproduce

- Configure a user for monitoring with no  SUPER or REPLICATION CLIENT permissions
- Test mysql plugin: `/usr/libexec/netdata/plugins.d/python.d.plugin debug 1 mysql`

```
2019-08-08 09:17:17: python.d ERROR: mysql[tcp] : Removed query: user_statistics[SHOW USER_STATISTICS;]. Error: error
2019-08-08 09:17:17: python.d ERROR: mysql[tcp] : Removed query: slave_status[SHOW SLAVE STATUS;]. Error: error
2019-08-08 09:17:17: python.d ERROR: mysql[tcp] : _get_data() failed. Error: 'slave_status'
```

##### Expected behavior

- Continue to monitor mysql as before

badcrc:
As a workaround I have modified `mysql.chart.py` and added a try/except for the `slave_status` part:

```
        try:
            if 'slave_status' in raw_data:
                status = self.get_slave_status(raw_data['slave_status'])
                data.update(status)
            else:
                self.queries.pop('slave_status')
        except:
            pass
```

ilyam8:
@badcrc please test #6617

badcrc:
That works:

```
2019-08-08 10:04:59: python.d INFO: plugin[main] : setup complete, 1 active module(s) : '['mysql']'
2019-08-08 10:04:59: python.d INFO: plugin[main] : mysql : created 1 job(s)
2019-08-08 10:04:59: python.d INFO: plugin[main] : mysql[tcp] : init successful
2019-08-08 10:04:59: python.d ERROR: mysql[tcp] : Removed query: user_statistics[SHOW USER_STATISTICS;]. Error: error
[...]

PR #6597, opened 2019-08-05
mysql collector: slave_status charts per replication channel

ilyam8:
##### Summary

Fixes: #6353

Ref: https://dev.mysql.com/doc/refman/8.0/en/replication-channels.html

> To provide compatibility with previous versions, the MySQL server automatically creates on startup a default channel whose name is the empty string (""). This channel is always present; it cannot be created or destroyed by the user. If no other channels (having nonempty names) have been created, replication statements act on the default channel only, so that all replication statements from older slaves function as expected

> SHOW SLAVE STATUS reports the status for all channels, except the group_replication_applier channel.

___ 

As we can see in https://github.com/netdata/netdata/issues/6353#issuecomment-507095714

`SHOW SLAVE STATUS` returns several rows, one row per replication channel.

Before this PR:
 - slave status charts only for the first channel.

After this PR:
 - slave status charts for every channel.

##### Component Name
[/collectors/python.d.plugin/mysql](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/mysql)

##### Additional Information

Tests:
 - i tested PR with 1 replication channel, works ok
 - i didnt test the PR with n replication channels (@vpapas will do i hope)

@vpapas could you test the PR?

___ 

> SHOW SLAVE STATUS reports the status for all channels, except the group_replication_applier channel.

From what i understand the only way to monitor replication metrics if instance in replication group is to execute `SHOW SLAVE STATUS CHANNEL <NAME>` for every channel. `SHOW SLAVE STATUS` returns nothing in that case.

PR #6617, opened 2019-08-08
mysql: do not remove `slave_status` from queries in module

ilyam8:
##### Summary

Fixes: #6616

bug was added in #6597

MySQLService removes query itself in  case of error

https://github.com/netdata/netdata/blob/66d6498ed4fa433fdd397da5ef1470312e539bd6/collectors/python.d.plugin/python_modules/bases/FrameworkServices/MySQLService.py#L144-L151

##### Component Name

[/collectors/python.d.plugin/mysql](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/mysql)


ilyam8:
Ok, the problem was with user w/o replication priviliges, now it is fixed, i tested both cases
 - user w/ replication privileges
 - user w/o replication privileges

PR #6621, opened 2019-08-08
 Change "netdata" to "Netdata" in all docs

### LINKED ISSUE #4725: Change "netdata" to "Netdata" in all docs
The official name of project netdata/netdata is "Netdata". All references to "netdata" that aren't about the executable, the directory structure, github links etc. will need to be replaced with "Netdata". This can't safely be done with simple replacement, as the lower case version appears several thousand times in the various places. 

We'll be doing it gradually and any help would be much appreciated.

joelhans:
##### Summary

Fixes #4725. Or, at least gets as close to fixing as I can manage with my sanity intact for now. I'm sure I missed some of them, but I caught nearly 800 instances of `netdata` across the repo, so it's a good start.

If someone wants to test this, or look for more I might have missed, I used VSCode to do three separate searches of the entire repo (well, only `.md` files in the repo) using regex:

`\snetdata\s`
`netdata\s`
`\snetdata`

I then did the same search through each individual file that showed up in the search results and then replaced the strings as it made sense, avoiding anywhere the string `netdata` was used as a command, a process, a user, in a URL, or referring to a configuration file, like `netdata.conf`.

I didn't bother searching for `netdata` on its own, without the regex for whitespace, as it would be very rare for us to capitalize `netdata` if it's not a standalone word in a sentence. Plus, there's 5000+ results for that, so parsing it would take forever.

##### Component Name

docs

##### Additional Information

I did not change the changelog or the historical changelog. I don't feel OK changing the names of the issues/PRs our community had submitted, but perhaps we can change the text


ilyam8:
> i'm sure I missed some of them

i tried to find some misses, but failed. gj @joelhans 	:+1:

joelhans:
@cakrit Conflict should be resolved now.

PR #6624, opened 2019-08-09
.travis.yml: Remove 'sudo: true' as it is now deprecated

knatsakis:
##### Summary
Remove 'sudo: true' from .travis.yml because it has been deprecated

##### Component Name
.travis.yml

##### Additional Information
According to the following blog post, the 'sudo: true/false' setting in .travis.yml has been deprecated:
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Some more information can be found here as well:
https://changelog.travis-ci.com/linux-builds-run-on-vms-by-default-77106

Also a test build with all stages and jobs can be found here:
https://travis-ci.org/knatsakis/netdata/builds/569643721 (There are some failures, but they are not related to the sudo change as far as I can tell).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 dist: xenial
-sudo: true
 language: c
 services:
   - docker


paulkatsoulakis:
It's interesting that the job terminated with `The job exceeded the maximum time limit for jobs, and has been terminated.`, it's like we got slower somehow ? Are you sure this is not affected by this option? Could you attempt to rebuild and see how it goes?

knatsakis:
Hi @paulkatsoulakis,

I ran the same test build with 'sudo: true' here:
https://travis-ci.org/knatsakis/netdata/builds/569696679

You will see that, it times out again, during compilation, after 50 minutes at the same jobs ("Build & Publish docker images). The 50 minute timeout is documented here: https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts

So the behavior is the same, both with 'sudo: true' and without.
I believe it is due to the fact that travis-ci.org jobs are slower than travis-ci.com. Unfortunately, I don't have a way to test with travis-ci.com.

PR #6639, opened 2019-08-12
Bugreport 931843

sunflowerbofh:
##### Summary

Fixing https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=931843

##### Component Name

apache2


joelhans:
Thanks for the contribution! It looks like your PR includes some commits from other contributors. Do you think you could clean that up so we only have to worry about your commits? That may be part of what's causing Travis to show errors.

sunflowerbofh:
not possible to sync commits

PR #6652, opened 2019-08-13
Prometheus version update

### LINKED ISSUE #6630: Prometheus documentation version update
##### Feature idea summary
Our documentation now is saying for we do download of the Prometheus 2.3.2, but today when I got the Prometheus to test, I verified that the current release is 2.11.1, so instead to do download of a specific version, I think it would be better to download of the stable version.

The link identified with problem was https://docs.netdata.cloud/backends/prometheus/ , this is not the unique link talking about Prometheus, we also have https://docs.netdata.cloud/backends/walkthrough/#prometheus talking about,  in this last we do download of 1.7.1, but we have a better text that does not allow miss understanding "Prometheus provides a tarball of their latest stable versions here: https://prometheus.io/download/. "

##### Expected behavior
Do not create a wrong vision for users that Netdata works with a specific version of Prometheus.

prhomhyse:
##### Summary
Fixes #6630 

This PR updates the version of Prometheus in the documentation and also fixes some markdown lint warnings.

##### Component Name
doc


joelhans:
I wonder if we can use one of these one-liners to always download the latest version of the package that begins with `prometheus` and ends with `linux-amd64.tar.gz`:

https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8

5 minutes of experimenting led nowhere for me, but if you can figure it out, that'll help make the article version-agnostic.

prhomhyse:
@cakrit 
I have removed the mention of versions in both files
Thanks for the tip @joelhans 
I have figured it out and added it to both files.

And as expected your other PR conflicted and thankfully, it was just one of the files 😅

PR #6655, opened 2019-08-13
.travis.yml: packaging/docker/check_login.sh can cause the build to fail

knatsakis:
##### Summary
Change .travis.yml so that the script packaging/docker/check_login.sh can cause the build to fail promptly (Publish for release & Nightly release stages).

##### Component Name
.travis.yml

##### Additional Information
I believe that packaging/docker/check_login.sh was meant to cancel the release build if the credentials are wrong. But currently the build keeps running even if check_login.sh fails.

With the current .travis.yml, this is in accordance with the [documented](https://docs.travis-ci.com/user/job-lifecycle#customizing-the-build-phase) behaviour:

> When one of the build commands returns a non-zero exit code, the Travis CI build **runs the subsequent commands as well**, and accumulates the build result.

This PR makes the check_login.sh failure to fail the build promptly.

You can find before and after test builds here:
https://travis-ci.org/knatsakis/netdata/jobs/571436083#L455
https://travis-ci.org/knatsakis/netdata/jobs/571436087#L406
and here:
https://travis-ci.org/knatsakis/netdata/jobs/571508994
https://travis-ci.org/knatsakis/netdata/jobs/571508998

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -505,10 +505,11 @@ jobs:
           - echo "Last commit:" && git log -1
           - echo "GIT Describe:" && git describe
           - echo "packaging/version:" && cat packaging/version
-          - packaging/docker/check_login.sh
-          - echo "Switching to latest master branch, to pick up tagging if any" && git checkout master && git pull
-          - packaging/docker/build.sh
-          - packaging/docker/publish.sh
+          - >-
+            packaging/docker/check_login.sh
+            && echo "Switching to latest master branch, to pick up tagging if any" && git checkout master && git pull
+            && packaging/docker/build.sh
+            && packaging/docker/publish.sh
         after_failure: post_message "TRAVIS_MESSAGE" "<!here> Docker image publishing failed"
 
       name: Build & Publish docker image for i386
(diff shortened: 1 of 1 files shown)


paulkatsoulakis:
Actually, it was not meant to cancel but rather provide extra information prior to the build process to start

knatsakis:
Hi @paulkatsoulakis,

I guess, I was mislead by the comment in the file:
> This is a credential checker script, to help get early input on docker credentials status
> If these are wrong, then build/publish has no point running

In any case, is there actually any point for the packaging/docker/build.sh and packaging/docker/publish.sh scripts to run if the credential check fails?

paulkatsoulakis:
> Hi @paulkatsoulakis,
> 
> I guess, I was mislead by the comment in the file:
> 
> > This is a credential checker script, to help get early input on docker credentials status
> > If these are wrong, then build/publish has no point running
> 
> In any case, is there actually any point for the packaging/docker/build.sh and packaging/docker/publish.sh scripts to run if the credential check fails?

It is a good question.
Credentials failing will only block publish step.
I think it will be good to always have the build result available.

On the new CI infrastructure we would certainly have distinct steps of build and publishing.

paulkatsoulakis:
Hi @knatsakis, coming back on my previous comment i just realized that i had already put the build-only step at `Docker container build process`, so in this case indeed there is no point of running the rest of the steps if the login is not working.

Kindly resolve conflicts and we can merge this

knatsakis:
Hi @paulkatsoulakis, I have resolved the merge conflicts.

PR #6659, opened 2019-08-14
Make MongoDB backend multithreaded

### LINKED ISSUE #6599: Make MongoDB backend multithreaded
##### Feature idea summary
The `mongoc` library blocks execution of the metric archiving process until all buffered data has been sent and a response has been received from the database. Blocking operations should be run in separate threads to allow Netdata buffering more than one batch of metrics at a time.

##### Expected behavior
Netdata sends metrics to a MongoDB database in a nonblocking way.

vlvkobal:
##### Summary
- Detach blocking operations into separate threads.
- Clean up dynamic objects using cleanup handlers on Netdata exit.

Closes #6599

##### Component Name
backends

### DIFF
--- a/backends/backends.c
+++ b/backends/backends.c
@@ -267,7 +267,7 @@ static struct objects_to_clean {
 #ifdef ENABLE_HTTPS
     struct netdata_ssl *opentsdb_ssl;
 #endif
-} objects_to_clean;
+};
 
 static void backends_main_cleanup(void *objects_to_clean) {
     struct objects_to_clean *objects = (struct objects_to_clean *)objects_to_clean;
@@ -543,6 +543,8 @@ BACKEND_TYPE backend_select_type(const char *type) {
  * @return It always return NULL.
  */
 void *backends_main(void *ptr) {
+    struct objects_to_clean objects_to_clean;
+
     netdata_thread_cleanup_push(backends_main_cleanup, (void *)&objects_to_clean);
 
     int default_port = 0;
@@ -1328,7 +1330,9 @@ void *backends_main(void *ptr) {
         // ------------------------------------------------------------------------
         // update the monitoring charts
 
+#if HAVE_MONGOC
 update_charts:
+#endif
         if(likely(chart_ops->counter_done)) rrdset_next(chart_ops);
         rrddim_set(chart_ops, "read",         chart_receptions);
         rrddim_set(chart_ops, "write",        chart_transmission_successes);
--- a/backends/backends.c
+++ b/backends/backends.c
@@ -517,16 +517,16 @@ BACKEND_TYPE backend_select_type(const char *type) {
     else if(!strcmp(type, "opentsdb:http") || !strcmp(type, "opentsdb:https")) {
         return BACKEND_TYPE_OPENTSDB_USING_HTTP;
     }
-    else if (!strcmp(type, "json") || !strcmp(type, "json:plaintext")) {
+    else if(!strcmp(type, "json") || !strcmp(type, "json:plaintext")) {
         return BACKEND_TYPE_JSON;
     }
-    else if (!strcmp(type, "prometheus_remote_write")) {
+    else if(!strcmp(type, "prometheus_remote_write")) {
         return  BACKEND_TYPE_PROMETEUS;
     }
-    else if (!strcmp(type, "kinesis") || !strcmp(type, "kinesis:plaintext")) {
+    else if(!strcmp(type, "kinesis") || !strcmp(type, "kinesis:plaintext")) {
         return BACKEND_TYPE_KINESIS;
     }
-    else if (!strcmp(type, "mongodb") || !strcmp(type, "mongodb:plaintext")) {
+    else if(!strcmp(type, "mongodb") || !strcmp(type, "mongodb:plaintext")) {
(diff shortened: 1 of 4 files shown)


backends/backends.c
```diff
@@ -460,37 +543,64 @@ BACKEND_TYPE backend_select_type(const char *type) {
...
+
     // set the default socket timeout in ms
-    int32_t mongodb_default_socket_timeout = (int32_t)(global_backend_update_every >= 2)?(global_backend_update_every * MSEC_PER_SEC - 500):1000;
+    int32_t mongodb_default_socket_timeout = (int32_t)(global_backend_update_every < 30)?(global_backend_update_every * 10 * MSEC_PER_SEC):300 * MSEC_PER_SEC;
 
 #endif
 
 #ifdef ENABLE_HTTPS
-            struct netdata_ssl opentsdb_ssl = {NULL , NETDATA_SSL_START};
+    struct netdata_ssl opentsdb_ssl = {NULL , NETDATA_SSL_START};
```
thiagoftsm: thanks to correct this!

backends/backends.c
```diff
@@ -900,43 +1043,77 @@ void *backends_main(void *ptr) {
 
 #if HAVE_MONGOC
         if(do_mongodb) {
+            for(int i = 0; i < MONGODB_THREADS_NUMBER; i++) {
```
thiagoftsm: Some old C compilers  does not accept declaration of variables inside a `for`, case my memory is correct, this was only brought to ANSI C with C 99, please, put the declaration of the variable out side the loop.

backends/backends.c
```diff
@@ -900,43 +1043,77 @@ void *backends_main(void *ptr) {
...
+                netdata_mutex_lock(&mongodb_threads[i].mutex);
+                if(mongodb_threads[i].finished) {
+                    if(unlikely(mongodb_threads[i].error)) {
+                        // oops! we couldn't send (all or some of the) data
+                        error("BACKEND: failed to write data to database backend '%s'. Willing to write %zu bytes, wrote %zu bytes.",
+                              mongodb_uri, mongodb_threads[i].n_bytes, 0UL);
```
thiagoftsm: The `error` is a definition  for the function error_int present inside the file libnetdata/log/log.c , inside this function we have a call for the vfprintf to format the message that will be redirect for our error log, case you see any source of vfprintf you will be that every time that it finds a `%`, it will go to a switch to identify the function that will be called to convert that specific value to a string, considering that the last argument will be awlays the constant `0`, I will give a suggestion that you write the zero direct in the message to avoid this whole process to convert 0 to '0x30' that is its ascii value.

backends/backends.c
```diff
@@ -900,43 +1043,77 @@ void *backends_main(void *ptr) {
...
+                netdata_mutex_unlock(&mongodb_threads[i].mutex);
+            }
+
             size_t buffer_len = buffer_strlen(b);
-            size_t sent = 0;
 
-            while(sent < buffer_len) {
-                const char *first_char = buffer_tostring(b);
+            if(unlikely(mongodb_thread_index == MONGODB_THREAD_INDEX_UNDEFINED)) {
+                error("BACKEND: failed to write data to database backend '%s'. All available threads are busy. Willing to write %zu bytes, wrote %zu bytes.",
```
thiagoftsm: The same thing wrote in the line 1052 is applied here.

backends/backends.c
```diff
@@ -1193,44 +1373,7 @@ void *backends_main(void *ptr) {
...
-        if(opentsdb_ssl.conn) {
-            SSL_free(opentsdb_ssl.conn);
-        }
-    }
-#endif
-
     netdata_thread_cleanup_pop(1);
+
     return NULL;
 }
```
thiagoftsm: Please, add a blank line at the end of this file.

backends/backends.c
```diff
@@ -595,6 +705,13 @@ void *backends_main(void *ptr) {
                 goto cleanup;
             }
 
+            mongodb_threads = callocz(MONGODB_THREADS_NUMBER, sizeof(struct mongodb_thread));
+
+            for(int i = 0; i < MONGODB_THREADS_NUMBER; i++) {
```
thiagoftsm: The same problem that I described on line 1046 is happening here.

backends/backends.c
vlvkobal: At least 7 of C99 features are already in use in the Netdata code. This particular feature has not been used yet, so I'll fix it. Let us discuss the matter on the meetup.

backends/backends.c
vlvkobal: Hm... It is already in place.

backends/backends.c
thiagoftsm: All right, my worry is with people using old Linux version that won't have this option enabled by default.

backends/backends.c
thiagoftsm: I did not do download of the code yesterday, but at least here in Github we do not have an additional line.

vlvkobal: If there is no newline at the end of a file, Github shows a [little red symbol](https://octicons.github.com/icon/no-newline/).

thiagoftsm: I did not know this, thanks!

Issue #6542, opened 2019-07-25
- Open an agent dashboard, without being signed-in


VLegakis:
- Open an agent dashboard, without being signed-in
- Click on Nodes view icon
- Sign-in screen opens
- Request to sign-in

Expected:
After successful sign-in, the user must be redirected in the Nodes view screen (based on his selection)

Actual:
After successful sign-in, the user is redirected back in the Agent Dashboard.

_Originally posted by @VLegakis in https://github.com/netdata/cloud/issues/908#issuecomment-514994358_

PR #6663, opened 2019-08-14
changed naming for redirect_uri param

jacekkolasa:
A change requested by the cloud team, since they're using snake_case for query string params and want to use more generic `uri` (instead of `url`)

This param was introduced recently to fix this issue: https://github.com/netdata/netdata/issues/6542
This PR needs also corresponding change in the [Cloud](https://github.com/netdata/cloud/pull/1237), and if only 1 change will be applied, there will be no breaking error because of this, but the minor issue described in #6542 will be happening again, until both Agent and Cloud will be updated.

redirectUrl > redirect_uri in query params for cloud sign-in call
redirectUrl > redirectUri in response object from iframe's window.parent.postMessage

### DIFF
--- a/web/gui/main.js
+++ b/web/gui/main.js
@@ -4897,8 +4897,8 @@ function handleSignInMessage(e) {
     cloudToken = e.data.token;
 
     netdataRegistryCallback(registryAgents);
-    if (e.data.redirectUri) {
-        window.location.replace(e.data.redirectUri);
+    if (e.data.redirectURI) {
+        window.location.replace(e.data.redirectURI);
     }
 }
 
--- a/web/gui/main.js
+++ b/web/gui/main.js
@@ -708,7 +708,7 @@ function openAuthenticatedUrl(url) {
     if (isSignedIn()) {
         window.open(url);
     } else {
-        window.open(`${NETDATA.registry.cloudBaseURL}/account/sign-in-agent?id=${NETDATA.registry.machine_guid}&name=${encodeURIComponent(NETDATA.registry.hostname)}&origin=${encodeURIComponent(window.location.origin + "/")}&redirectUrl=${encodeURIComponent(window.location.origin + "/" + url)}`);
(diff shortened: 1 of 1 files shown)


web/gui/main.js
```diff
@@ -4897,8 +4897,8 @@ function handleSignInMessage(e) {
     cloudToken = e.data.token;
 
     netdataRegistryCallback(registryAgents);
-    if (e.data.redirectUrl) {
-        window.location.replace(e.data.redirectUrl);
+    if (e.data.redirectUri) {
```
nicolasparada: Could you name the js var `redirectURI`?
From cloud we do that.
Ex: `agentID`, `callbackURL`, `redirectURI`.

jacekkolasa:
@cakrit i've wrote "minor issue" because `no redirect after login - when accessing protected route as an unauthenticated user` is an issue that i was experiencing a lot few years ago, on major websites. Also because the last fix is present only for 2 weeks and I assumed that mostly users that upgrade the core frequently will experience that issue.

But, having something broken is totally unnecessary. I've added the backwards compatibility to the Cloud. I've wrongly estimated our priorities and how big compromises we can make. I'll be more aware of our guidelines next time. Thanks for your attention on this!

So, to be 100% backwards compatible, i think we can merge this PR after the change on [the Cloud side](https://github.com/netdata/cloud/pull/1237) will be deployed.

Different naming in `&redirect_uri=` and `e.data.redirectURI` is caused by convention used [here](https://github.com/netdata/cloud/pull/1237/files#diff-540c313b4f67fa6fb60b2b3a25edd1f9R21-R31). I've tested this

web/gui/main.js
jacekkolasa: absolutely! done

nicolasparada:
We can detect if the old parameter name is still being used by checking in the logs:
The URL path should contain "/account/sign-in-agent" and "redirectUrl=".

jacekkolasa:
the related change is deployed to cloud production, so it's safe to proceed with merging this PR

PR #6673, opened 2019-08-15
.travis.yml: Improve template and commit_message predicates readability

knatsakis:
##### Summary
Improve readability of commit_message predicates and package (DEB/RPM) building templates

##### Component Name
.travis.yml

##### Additional Information
You can find test builds, here:
https://travis-ci.org/knatsakis/netdata/builds/572431794
and here:
https://travis-ci.org/knatsakis/netdata/builds/572433452

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ stages:
 
   - name: Publish for release
     # We don't run on release candidates
-    if: branch = master AND type != pull_request AND type != cron AND commit_message =~ /(\[netdata release candidate\]|\[netdata major release\]|\[netdata minor release\]|\[netdata patch release\])/ AND tag !~ /(-rc)/
+    if: branch = master AND type != pull_request AND type != cron AND commit_message =~ /\[netdata (release candidate|(major|minor|patch) release)\]/ AND tag !~ /(-rc)/
 
     # Build DEB packages under special conditions
   - name: Package ubuntu/* and debian/*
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -60,7 +60,7 @@ stages:
     if: branch = master AND type != pull_request AND type != cron
 
   - name: Publish for release
-    if: branch = master AND type != pull_request AND type != cron AND commit_message =~ /(\[netdata release candidate\]|\[netdata major release\]|\[netdata minor release\]|\[netdata patch release\])/
+    if: branch = master AND type != pull_request AND type != cron AND commit_message =~ /\[netdata (release candidate|(major|minor|patch) release)\]/
```
paulkatsoulakis: Have you confirmed this pattern correctly translates all possible combinations?
It looks correct, is just good to have confirmed results.

You can cancel the process once it reaches to the point that proves it got the right keyword

knatsakis:
Please find the test builds and the respective commits below:

For '[netdata release candidate]':
https://travis-ci.org/knatsakis/netdata/builds/572916735
https://github.com/knatsakis/netdata/commit/ccd1495ded69956c5a71ad516d58e7df2a8c9403

For '[netdata major release]':
https://travis-ci.org/knatsakis/netdata/builds/572917160
https://github.com/knatsakis/netdata/commit/5330b50f171bdf60dbf661ae6db8a314b389ab7c

For '[netdata minor release]':
https://travis-ci.org/knatsakis/netdata/builds/572916960
https://github.com/knatsakis/netdata/commit/9743dbf79cd5d2d6b3be551c19b32b81c275eef7

For '[netdata patch release]':
https://travis-ci.org/knatsakis/netdata/builds/572917414
https://github.com/knatsakis/netdata/commit/729ee57b42be206e6a229cb9669f39c07aaa6962

For '[netdata no release]':
https://travis-ci.org/knatsakis/netdata/builds/572920874
https://github.com/knatsakis/netdata/commit/e7ae0c1f3752d5330e1b1d4e93819bf684936fcb

You will notice that all builds above have a 'Publish for release' stage except the last one.

.travis.yml
knatsakis: Please find the test builds and the respective commits below:

For '[netdata release candidate]':
https://travis-ci.org/knatsakis/netdata/builds/572916735
https://github.com/knatsakis/netdata/commit/ccd1495ded69956c5a71ad516d58e7df2a8c9403

For '[netdata major release]':
https://travis-ci.org/knatsakis/netdata/builds/572917160
https://github.com/knatsakis/netdata/commit/5330b50f171bdf60dbf661ae6db8a314b389ab7c

For '[netdata minor release]':
https://travis-ci.org/knatsakis/netdata/builds/572916960
https://github.com/knatsakis/netdata/commit/9743dbf79cd5d2d6b3be551c19b32b81c275eef7

For '[netdata patch release]':
https://travis-ci.org/knatsakis/netdata/builds/572917414
https://github.com/knatsakis/netdata/commit/729ee57b42be206e6a229cb9669f39c07aaa6962

For '[netdata no release]':
https://travis-ci.org/knatsakis/netdata/builds/572920874
https://github.com/knatsakis/netdata/commit/e7ae0c1f3752d5330e1b1d4e93819bf684936fcb

You will notice that all builds above have a 'Publish for release' stage except the last one.

paulkatsoulakis: Thanks for running the tests @knatsakis, as you can see from these results, you allow both nightly and major/minor release to run on the same build. See for example: https://travis-ci.org/knatsakis/netdata/builds/572917160

This is not right. Nightlies run only from cron, while release runs based on the commit.

I also notice packages are triggered, which is probably a bad idea because we extract the architecture from the triggering commit that is absent in this case

Thanks!

knatsakis: Please note that the test builds where ran with a commit slightly different than the commit in the PR. You can see the differences here:
https://github.com/knatsakis/netdata/compare/knatsakis-travis-ci-template-readability...knatsakis:knatsakis-travis-ci-template-readability-test

You will see that I have short-circuited all 'if's. This is something that I do in order to be able to test all the stages of .travis.yml. Also I do it in order to remove the 'if: branch = master', because otherwise I would not be able to test from a branch.

Also, this why I put two builds in my original pull request comment. One is without these modifications (but doesn't test much) and the other is with the modifications.

If you can think of a better way to test, I am all ears :)

So, this is why you saw the discrepancies that you mention.
Could you please review this again under the light of this new info?

Thanks.

paulkatsoulakis: I see, it would be cleaner if you just modify `branch = master` to `branch = my-branch` and demonstrate also that all conditionals are working as expected.
Of course i am not saying they are wrong, but when we test we need to be sure we test with the final production setup and the conditionals are crucial part of the production set up of travis.yml

paulkatsoulakis:
Oh @knatsakis kindly solve conflicts and i 'll get to merge it, thanks !

knatsakis:
Hi @paulkatsoulakis, I have resolved the merge conflicts.

knatsakis:
> LGTM -
> 
> FWIW The consolidation on the pattern matching is good stuff. The rest of the changes i haven't seen what is the value added, it would be nice to have some more explanations that would verify the requirement of doing them, like for example compliance with a standard or with a specific syntax.
> 
> Other than that, looks okay

Well, I have no justification for the changes in this pull request, other than "improved readability", which I mentioned in my first comment. Of course "improved readability" can be very subjective.

Actually, I was hoping that you would see value in the rest of the changes, just like you did for the pattern matching stuff.

Personally, I like the removal of the:
`if: true == false`
lines. This is made possible by the use of the _template private variable.

.travis.yml
knatsakis: It is true that `branch = my-branch` would be cleaner, but I think that testing the conditionals is not necessary, unless the pull request changes the conditionals.

This pull request *does* change the conditionals, but only the last part. Also, in order to test some jobs such as "Nightly release", I have to change the conditionals anyway in order to also remove the `type = cron` predicate and again invalidate the test that conditionals work as expected.

The reason I am removing the `branch = master` completely is that in order to save time and effort, I have a single commit:
https://github.com/knatsakis/netdata/commit/10aa8757c160b6e1b14f33aa5079c8a3b6335f40
that I cherry pick in every branch that I want to test. This commit does some necessary stuff in order to test all the travis jobs, e.g: disables slack notifications that otherwise cause a failed build.

I would like to find a better way to test travis.yml from a fork with a better technique. Any ideas?

Finally, since https://github.com/netdata/netdata only has a master branch, is there any point in having the `branch = master` predicate at all?

PR #6677, opened 2019-08-16
Add clang-format. Update Contribution guidelines.

cosmix:
##### Summary

Despite already having very specific contribution guidelines, new code in the netdata C agent is still not observing the code style agreed upon. This PR adds a `.clang-format` file that can be used by clang to automatically format the linux kernel style (with our accepted modifications).


.clang-format
```diff
@@ -0,0 +1,118 @@
...
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles: false
+SpacesInContainerLiterals: false
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard: Cpp03
+TabWidth: 4
+UseTab: Always
```
cakrit: We use spaces, not tabs.

cakrit:
Is this correct? Doesn't look very good...

```
-          RRDSET *st                // the chart
-        , RRDDIM *rd                // the dimension
-        , time_t after              // the start timestamp
-        , time_t before             // the end timestamp
-        , BACKEND_OPTIONS backend_options // BACKEND_SOURCE_* bitmap
-        , time_t *first_timestamp   // the first point of the database used in this response
-        , time_t *last_timestamp    // the timestamp that should be reported to backend
+    RRDSET *st // the chart
+    ,
+    RRDDIM *rd // the dimension
+    ,
+    time_t after // the start timestamp
+    ,
+    time_t before // the end timestamp
+    ,
+    BACKEND_OPTIONS backend_options // BACKEND_SOURCE_* bitmap
+    ,
+    time_t *first_timestamp // the first point of the database used in this response
+    ,
+    time_t *last_timestamp // the timestamp that should be reported to backend
```

mfundul:
There is a lot of @ktsaou code that has been painstakingly formatted by hand and is bound to look awful after the linter.

cosmix:
@mfundul we'll have to check that and see how we deal with it on a case by case basis. i.e. I see the point, it's perfectly valid, but we need to standardize!

thiagoftsm:
I will bring here something that I told for @vlvkobal in his PR on Wednesday and it is realted to this issue, I know that C99 is officially the C version that Netdata uses, but I was not comfortable when I read his code and there was a:

```for( int i =0 ; i < MAX_MONGODB_THREADS ; i++ ) ```

I agree that we cannot discard features that was brought to newest versions of C, but  this changed introduced in the language forces me to think I am reading an Object Orientation code, I do not remember exactly the whole kernel team wrote as the C style used there, I remember they are very conservative, but case we define that codes like I gave example will be accept, it will be good to keep an unique pattern for the whole Netdata code.

File: .clang-format [hunk 2]
```diff
@@ -0,0 +1,118 @@
...
+#
+# For more information, see:
+#
+#   Documentation/process/clang-format.rst
+#   https://clang.llvm.org/docs/ClangFormat.html
+#   https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+#
+---
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
```
cosmix: change this to: `AlwaysBreak`

.clang-format
```diff
@@ -0,0 +1,118 @@
...
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: false
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterClass: false
+  AfterControlStatement: false
+  AfterEnum: false
+  AfterFunction: false
```
cosmix: restore value to `true`.

.clang-format
```diff
@@ -0,0 +1,118 @@
...
+DisableFormat: false
+ExperimentalAutoDetectBinPacking: false
+#FixNamespaceComments: false # Unknown to clang-format-4.0
+
+#IncludeBlocks: Preserve # Unknown to clang-format-5.0
+IncludeCategories:
+  - Regex: '.*'
+    Priority: 1
+IncludeIsMainRegex: '(Test)?$'
+IndentCaseLabels: false
```
cosmix: Change to `true`.

PR #6691, opened 2019-08-19
Tabulatin on health/health.c

thiagoftsm:
##### Summary
As @mfundul detected doing a review, we had the file health/health.c with two patterns of tabulation, this PR fixes this problem.
##### Component Name
Health

### DIFF
--- a/health/health.c
+++ b/health/health.c
@@ -531,7 +531,6 @@ int update_disabled_silenced(RRDHOST *host, RRDCALC *rc) {
              (rc->rrdcalc_flags & RRDCALC_FLAG_SILENCED)?"true":"false"
         );
     }
-
     if(rc->rrdcalc_flags & RRDCALC_FLAG_DISABLED)
         return 1;
     else
--- a/health/health.c
+++ b/health/health.c
@@ -155,7 +155,6 @@ void health_reload_host(RRDHOST *host) {
  * Reload the host configuration for all hosts.
  */
 void health_reload(void) {
-
     rrd_rdlock();
 
     RRDHOST *host;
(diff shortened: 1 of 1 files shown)


health/health.c
```diff
@@ -509,32 +515,30 @@ SILENCE_TYPE check_silenced(RRDCALC *rc, char* host, SILENCERS *silencers) {
...
+             host->hostname,
+             rc->name,
+             (rrdcalc_flags_old & RRDCALC_FLAG_DISABLED)?"true":"false",
+             (rc->rrdcalc_flags & RRDCALC_FLAG_DISABLED)?"true":"false",
+             (rrdcalc_flags_old & RRDCALC_FLAG_SILENCED)?"true":"false",
+             (rc->rrdcalc_flags & RRDCALC_FLAG_SILENCED)?"true":"false"
+        );
+    }
+
+    return (rc->rrdcalc_flags & RRDCALC_FLAG_DISABLED)?1:0;
```
cakrit: I would leave this line as it was, since we're supposedly only doing tab changes in this PR. It may help align the rest of the lines as well, because the line numbers change due to this. I don't like the diff that github shows (probably just because of this line), it's misleading.

thiagoftsm: You are correct, I return it for the previous code changing only the tabulation.

health/health.c
```diff
@@ -155,7 +155,6 @@ void health_reload_host(RRDHOST *host) {
  * Reload the host configuration for all hosts.
  */
 void health_reload(void) {
-
     rrd_rdlock();
```
cakrit: Add that empty line back

health/health.c
```diff
@@ -318,6 +317,13 @@ static inline void health_process_notifications(RRDHOST *host, ALARM_ENTRY *ae)
     health_alarm_execute(host, ae);
 }
 
+/**
```
cakrit: Move this comment to another PR

thiagoftsm: I removed, I will create a specific for comments on alarm code.

health/health.c
thiagoftsm: I returned it with another commit, but it will be necessary a PR to remove it, because this function has a pattern with this blank like that is not present in other functions of the file.

thiagoftsm:
> There are two more like the other one, that still cause misalignment and difficulty to compare.
> Basically, see what else need to be moved to another PR, so that all lines align perfectly and we just see tab changes.

There were more than 2 places, so I changed the function call closing and blank line missed.

cakrit:
Still see at the end some line number differences.

thiagoftsm:
> Still see at the end some line number differences.

I got the file from master using the link ( https://raw.githubusercontent.com/netdata/netdata/master/health/health.c ) and I removed the tabulation again from scratch, I saw that github continues giving difference, but this time I was not using a file that was changed before, it is the master branch changed.

thiagoftsm:
> Even `diff` gets confused. I used an online tool and this is fine.

I understand you, I was confused with the fact that after I do what was described in the previous message, I continue having problems with lines.

PR #6693, opened 2019-08-20
netdata/installer: Add support for offline installs using kickstart or kickstart-static64

### LINKED ISSUE #6684: Support for a fully offline installation using pre-compiled binary
##### Feature idea summary
In the current set of installation methods available for netdata as per the documentation, the pre-compiled binary installation method script downloads the binaries straight from Github, and then installs it on the machine. In some enterprises, we do not have the luxury to do that since we might not have internet access. There should also be some script which we can use and also supply the binary ourselves, which will make the installation fully offline.

##### Expected behavior
Something like this should make sense:

```
/path/to/kickstart-static64.sh /path/to/pre-built-binary
```
This should skip downloading the binary from the internet and just install the binary from the location provided. There should be no connection to the internet to install netdata.

paulkatsoulakis:
##### Summary
We introduce the option `--local-files` to allow our users to manually define the source and the checksum file for the kickstart to use in the installation process.
This option is useful for cases that users do not want to run kickstart with an internet connection

Note: we still require internet though for the dependency management

##### Component Name
netdata/installer/kickstart
netdata/installer/kickstart-static64

##### Additional Information
Fixes #6684

### DIFF
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -530,6 +530,12 @@ download the required files manually.
 ```bash
 cd /tmp
 
+# Since you won't be having an active internet connection on the destination, you will need to manually get the kickstart script itself
+# Download kickstart.sh or kickstart-static64.sh, depending the installation you have chosen to run
+curl -s https://my-netdata.io/kickstart.sh > kickstart.sh
+or
+curl -s https://my-netdata.io/kickstart-static64.sh > kickstart-static64.sh
+
 # Netdata tarball
 curl -s https://api.github.com/repos/netdata/netdata/releases/latest | grep "browser_download_url.*tar.gz" | cut -d '"' -f 4 | wget -qi -
 
@@ -559,10 +565,10 @@ the location and names of the files you just downloaded.
 
 ```bash
 # kickstart.sh
-bash &lt;(curl -Ss https://my-netdata.io/kickstart.sh) --local-files /tmp/netdata.tar.gz /tmp/checksums.txt /tmp/go.d.binary.tar.gz /tmp/go.d.config.tar.gz
+bash kickstart.sh --local-files /tmp/netdata.tar.gz /tmp/checksums.txt /tmp/go.d.binary.tar.gz /tmp/go.d.config.tar.gz
 
 # kickstart-static64.sh
-bash &lt;(curl -Ss https://my-netdata.io/kickstart-static64.sh) --local-files /tmp/netdata.tar.gz /tmp/checksums.txt
+bash kickstart-static64.sh --local-files /tmp/netdata.tar.gz /tmp/checksums.txt
 ```
 
 Now that you're finished with your offline installation, you can move on to our
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -514,7 +514,7 @@ a few extra steps to make it work.
 
 By default, the `kickstart.sh` and `kickstart-static64.sh` download Netdata
 assets, like the precompiled binary and a few dependencies, using the system's
-internet connect, but you can also supply these files from the local filesystem.
+internet connection, but you can also supply these files from the local filesystem.
 
 First, download the required files. If you're using `kickstart.sh`, you need the
 Netdata tarball, the checksums, the go.d plugin binary, and the go.d plugin
@@ -545,14 +545,14 @@ curl -s https://api.github.com/repos/netdata/go.d.plugin/releases/latest | grep 
 ```
 
 Move these files to the `/tmp` directory on the offline system in whichever way
-your organization allows.
+your defined policy allows (if any).
 
 Now you can run either the `kickstart.sh` or `kickstart-static64.sh` scripts
 using the `--local-tarball-override` option. This option requires you to specify
 the location and names of the files you just downloaded. 
 
 !!! note When using `--local-tarball-override`, the `kickstart.sh` or
-    `kickstart-64.sh` scripts won't download any Netdata assets from the
+    `kickstart-static64.sh` scripts won't download any Netdata assets from the
     internet. But, you may still need a connection to install dependencies using
     your system's package manager. The scripts will warn you if your system
     doesn't have all the dependencies.
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -48,7 +48,7 @@ To learn more about the pros and cons of using *nightly* vs. *stable* releases, 
 Verify the integrity of the script with this:
 
 ```bash
-[ "b6d16c171ccad073b86327246151d875" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
+[ "6e54dd18a482dc8271b4ec245b986d3f" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
 ```
 
 _It should print `OK, VALID` if the script is the one we ship._
@@ -108,7 +108,7 @@ This script installs Netdata at `/opt/netdata`.
 Verify the integrity of the script with this:
 
 ```bash
-[ "4415e8c13e529a795abb953a9be14ad5" = "$(curl -Ss https://my-netdata.io/kickstart-static64.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
+[ "97a7696d59f8c91d0db8f9bca2e8c594" = "$(curl -Ss https://my-netdata.io/kickstart-static64.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
 ```
 
 *It should print `OK, VALID` if the script is the one we ship.*
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -48,7 +48,7 @@ To learn more about the pros and cons of using *nightly* vs. *stable* releases, 
 Verify the integrity of the script with this:
 
 ```bash
-[ "dc844b0f55d86afcb6c9405c76330371" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
+[ "735e9966a4cf0187863e06a5282b34a7" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
 ```
 
 _It should print `OK, VALID` if the script is the one we ship._
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -109,7 +109,7 @@ This script installs Netdata at `/opt/netdata`.
 Verify the integrity of the script with this:
 
 ```bash
-[ "97a7696d59f8c91d0db8f9bca2e8c594" = "$(curl -Ss https://my-netdata.io/kickstart-static64.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
+[ "8a9f864ddcb70b7dc01d4e5bf4550a81" = "$(curl -Ss https://my-netdata.io/kickstart-static64.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
 ```
 
 *It should print `OK, VALID` if the script is the one we ship.*
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -109,7 +109,7 @@ This script installs Netdata at `/opt/netdata`.
 Verify the integrity of the script with this:
 
 ```bash
-[ "722e0164ccdf98316b3ce5f563ffc2bf" = "$(curl -Ss https://my-netdata.io/kickstart-static64.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
+[ "5c432e78d5f40403e8b7c00b1a19a7ca" = "$(curl -Ss https://my-netdata.io/kickstart-static64.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
 ```
 
 *It should print `OK, VALID` if the script is the one we ship.*
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -67,13 +67,14 @@ The `kickstart.sh` script passes all its parameters to `netdata-installer.sh`, s
 -   `--dont-start-it`: Prevent the installer from starting Netdata automatically.
 -   `--stable-channel`: Automatically update only on the release of new major versions.
 -   `--no-updates`: Prevent automatic updates of any kind.
+-   `--local-tarball-override`: Useful for offline installations. Pass two file paths, one of the tarball and one of the checksum file to force kickstart run the process using those files.
 
 Example using all the above parameters:
 
 ```bash
-$ bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait --dont-start-it --no-updates --stable-channel
+$ bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait --dont-start-it --no-updates --stable-channel --local-tarball-override /tmp/my-selfdownloaded-tarball.tar.gz /tmp/checksums.txt
 ```
-
+Note: `--stable-channel` and `--local-tarball-override` overlap, if you use the tarball override the stable channel option is not effective
 </details>
 
 Once Netdata is installed, see [Getting Started](../../docs/GettingStarted.md).
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -48,7 +48,7 @@ To learn more about the pros and cons of using *nightly* vs. *stable* releases, 
 Verify the integrity of the script with this:
 
 ```bash
-[ "dc439c58bfc41de23b14b86d71c79a31" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
+[ "dc844b0f55d86afcb6c9405c76330371" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
 ```
 
 _It should print `OK, VALID` if the script is the one we ship._
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -48,7 +48,7 @@ To learn more about the pros and cons of using *nightly* vs. *stable* releases, 
 Verify the integrity of the script with this:
 
 ```bash
-[ "6e54dd18a482dc8271b4ec245b986d3f" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
+[ "574456abacd54a625b44275fda43c507" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
 ```
 
 _It should print `OK, VALID` if the script is the one we ship._
@@ -67,12 +67,12 @@ The `kickstart.sh` script passes all its parameters to `netdata-installer.sh`, s
 -   `--dont-start-it`: Prevent the installer from starting Netdata automatically.
 -   `--stable-channel`: Automatically update only on the release of new major versions.
 -   `--no-updates`: Prevent automatic updates of any kind.
--   `--local-tarball-override`: Useful for offline installations. Pass two file paths, one of the tarball and one of the checksum file to force kickstart run the process using those files.
+-   `--local-tarball-override`: Useful for offline installations. Pass four file paths: the netdata tarball, the checksum file, the go.d plugin tarball and the go.d plugin config tarball to force kickstart run the process using those files.
 
 Example using all the above parameters:
 
 ```bash
-$ bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait --dont-start-it --no-updates --stable-channel --local-tarball-override /tmp/my-selfdownloaded-tarball.tar.gz /tmp/checksums.txt
+$ bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait --dont-start-it --no-updates --stable-channel --local-tarball-override /tmp/my-selfdownloaded-tarball.tar.gz /tmp/checksums.txt /tmp/manually.downloaded.go.d.binary.tar.gz /tmp/manually.downloaded.go.d.config.tar.gz
 ```
 Note: `--stable-channel` and `--local-tarball-override` overlap, if you use the tarball override the stable channel option is not effective
 </details>
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -120,12 +120,14 @@ The `kickstart-static64.sh` script passes all its parameters to `netdata-install
 -   `--dont-start-it`: Prevent the installer from starting Netdata automatically.
 -   `--stable-channel`: Automatically update only on the release of new major versions.
 -   `--no-updates`: Prevent automatic updates of any kind.
+-   `--local-tarball-override`: Useful for offline installations. Pass two file paths, one of the tarball and one of the checksum file to force kickstart run the process using those files.
 
 Example using all the above parameters:
 
 ```sh
-$ bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh) --dont-wait --dont-start-it --no-updates --stable-channel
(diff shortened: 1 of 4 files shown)


paulkatsoulakis:
> * The info in the README is hidden. Offline installation deserves at least a short paragraph with the exact steps.
ack

> * The info was only added to kickstart.sh, static64 needs it too.
oops, right

> * Not sure what you mean about dependency management, but we need to at least explain it.
I meant the install required packages script that we call, to install needed packages for netdata to compile and install. this mechanism applies for kickstart.sh

> * I don't get what will be done with the go.d dependency. Should they add flags to not include it? Can we also have an option to read the package locally if they manually download it?
yikes! Good point, i 'll rework that part

paulkatsoulakis:
@joelhans could you propose placement for  
```
The info in the README is hidden. Offline installation deserves at least a short paragraph with the exact steps.
```

packaging/installer/README.md
```diff
@@ -67,13 +67,14 @@ The `kickstart.sh` script passes all its parameters to `netdata-installer.sh`, s
 -   `--dont-start-it`: Prevent the installer from starting Netdata automatically.
 -   `--stable-channel`: Automatically update only on the release of new major versions.
 -   `--no-updates`: Prevent automatic updates of any kind.
+-   `--local-tarball-override`: Useful for offline installations. Pass four file paths: the netdata tarball, the checksum file, the go.d plugin tarball and the go.d plugin config tarball to force kickstart run the process using those files.
```
joelhans: A few tweaks here:

```suggestion
-   `--local-tarball-override`: Used for offline installations. Pass four file paths: the Netdata tarball, the checksum file, the go.d plugin tarball, and the go.d plugin config tarball, to force kickstart run the process using those files.
```

packaging/installer/README.md
```diff
@@ -67,13 +67,14 @@ The `kickstart.sh` script passes all its parameters to `netdata-installer.sh`, s
...
 
 Example using all the above parameters:
 
 ```bash
-$ bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait --dont-start-it --no-updates --stable-channel
+$ bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait --dont-start-it --no-updates --stable-channel --local-tarball-override /tmp/my-selfdownloaded-tarball.tar.gz /tmp/checksums.txt /tmp/manually.downloaded.go.d.binary.tar.gz /tmp/manually.downloaded.go.d.config.tar.gz
 ```
-
+Note: `--stable-channel` and `--local-tarball-override` overlap, if you use the tarball override the stable channel option is not effective
```
joelhans: I'm not sure I understand what "is not effective" means. Does `--local-tarball-override` default to the nightly updates?

packaging/installer/README.md
```diff
@@ -119,12 +120,14 @@ The `kickstart-static64.sh` script passes all its parameters to `netdata-install
 -   `--dont-start-it`: Prevent the installer from starting Netdata automatically.
 -   `--stable-channel`: Automatically update only on the release of new major versions.
 -   `--no-updates`: Prevent automatic updates of any kind.
+-   `--local-tarball-override`: Useful for offline installations. Pass two file paths, one of the tarball and one of the checksum file to force kickstart run the process using those files.
```
joelhans:
```suggestion
-   `--local-tarball-override`: Used for offline installations. Pass two file paths, one for the tarball and one fir the checksum file, to force kickstart run the process using those files.
```

packaging/installer/README.md
```diff
@@ -119,12 +120,14 @@ The `kickstart-static64.sh` script passes all its parameters to `netdata-install
...
 
 Example using all the above parameters:
 
 ```sh
-$ bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh) --dont-wait --dont-start-it --no-updates --stable-channel
+$ bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh) --dont-wait --dont-start-it --no-updates --stable-channel --local-tarball-override /tmp/my-selfdownloaded-tarball.tar.gz /tmp/checksums.txt
 ```
+Note: `--stable-channel` and `--local-tarball-override` overlap, if you use the tarball override the stable channel option is not effective
```
joelhans: Same comment here as the one above. We should clarify which channel they'll be on when they use `--local-tarball-override`.

packaging/installer/README.md
paulkatsoulakis: --stable-channel is only meaningful when we plan to download something.
When using the override, the user provides the tarballs so the flag doesn't make any sense

joelhans:
Here's what I put together for a section on offline installations. This method works for `kickstart.sh`, but I got a failure for `kickstart-static64.sh`, which I'll include in another comment. Not sure if it's my issue or an issue in the script.

We can put this new section above **Nightly vs. stable releases**.

```
## Offline installations

You can install Netdata on systems without internet access, but you need to take a few extra steps to make it work.

By default, the `kickstart.sh` and `kickstart-static64.sh` download Netdata assets, like the precompiled binary and a few dependencies, using the system's internet connect, but you can also supply these files from the local filesystem.

First, download the required files. If you're using `kickstart.sh`, you need the Netdata tarball, the checksums, the go.d plugin binary, and the go.d plugin configuration. If you're using `kickstart-static64.sh`, you need only the Netdata tarball and checksums.

Download the files you need to a system of yours that's connected to the internet. You can use the commands below, or visit the [latest Netdata release page](https://github.com/netdata/netdata/releases/latest) and [latest go.d plugin release page](https://github.com/netdata/go.d.plugin/releases) to download the required files manually.

```bash
## Offline installations

You can install Netdata on systems without internet access, but you need to take a few extra steps to make it work.

By default, the `kickstart.sh` and `kickstart-static64.sh` download Netdata assets, like the precompiled binary and a few dependencies, using the system's internet connect, but you can also supply these files from the local filesystem.

First, download the required files. If you're using `kickstart.sh`, you need the Netdata tarball, the checksums, the go.d plugin binary, and the go.d plugin configuration. If you're using `kickstart-static64.sh`, you need only the Netdata tarball and checksums.

Download the files you need to a system of yours that's connected to the internet. You can use the commands below, or visit the [latest Netdata release page](https://github.com/netdata/netdata/releases/latest) and [latest go.d plugin release page](https://github.com/netdata/go.d.plugin/releases) to download the required files manually.

```bash
cd /tmp

# Netdata tarball
curl -s https://api.github.com/repos/netdata/netdata/releases/latest | grep "browser_download_url.*tar.gz" | cut -d '"' -f 4 | wget -qi -

# Netdata checksums
curl -s https://api.github.com/repos/netdata/netdata/releases/latest | grep "browser_download_url.*txt" | cut -d '"' -f 4 | wget -qi -

# go.d plugin (for linux-amd64 systems)
# For binaries for other OS types and architectures, visit the go.d releases 
# page: https://github.com/netdata/go.d.plugin/releases/latest
curl -s https://api.github.com/repos/netdata/go.d.plugin/releases/latest | grep "browser_download_url.*linux-amd64.tar.gz" | cut -d '"' -f 4 | wget -qi -

# go.d configuration 
curl -s https://api.github.com/repos/netdata/go.d.plugin/releases/latest | grep "browser_download_url.*config.tar.gz" | cut -d '"' -f 4 | wget -qi -
```

Move these files to the `/tmp` directory on the offline system in whichever way your organization allows.

Now you can run either the `kickstart.sh` or `kickstart-static64.sh` scripts using the `--local-tarball-override` option. This option requires you to specify the location and names of the files you just downloaded. 

!!! note
    When using `--local-tarball-override`, the `kickstart.sh` or `kickstart-64.sh` scripts won't download any Netdata assets from the internet. But, you may still need a connection to install dependencies using your system's package manager. The scripts will warn you if your system doesn't have all the dependencies.

```bash
# kickstart.sh
bash <(curl -Ss https://my-netdata.io/kickstart.sh) --local-tarball-override /tmp/netdata.tar.gz /tmp/checksums.txt /tmp/go.d.binary.tar.gz /tmp/go.d.config.tar.gz

# kickstart-static64.sh
bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh) --local-tarball-override /tmp/netdata.tar.gz /tmp/checksums.txt
```

Now that you're finished with your offline installation, you can move on to our [getting started guide](../../docs/GettingStarted.md)!
```

**Comment by joelhans** [comment]
And now, the output from trying to run `kickstart-static64.sh`:

```
sh kickstart-static64.sh --local-tarball-override ~/tmp/netdata-v1.16.0.tar.gz ~/tmp/sha256sums.txt
kickstart-static64.sh: line 213: shift: shift count out of range
 --- Installation sources were given as input, running installation using "/home/joel/tmp/netdata-v1.16.0.tar.gz" --- 
[/tmp/netdata-kickstart-wlHHnl]$ mv /home/joel/tmp/netdata-v1.16.0.tar.gz /tmp/netdata-kickstart-wlHHnl/netdata-latest.gz.run 
 OK   

[/tmp/netdata-kickstart-wlHHnl]$ mv /home/joel/tmp/sha256sums.txt /tmp/netdata-kickstart-wlHHnl/sha256sum.txt 
 OK   

 ABORTED  Static binary checksum validation failed. Stopping netdata installation and leaving binary in /tmp/netdata-kickstart-wlHHnl 
```

**Comment by paulkatsoulakis** [comment]
> kickstart-static64.sh: line 213: shift: shift count out of range

woops! That slipped through, let me check, thanks Joel!

Btw KUDOS on the instructions, the commands put together are nit and clean as so the instructions, thanks !

**Comment by paulkatsoulakis** [comment]
BTW on the given commands, you can add `$(uname -s | tr '[:upper:]' '[:lower:]')` to get the OS and `$(uname -m)` to get the ARCH and then instruct the user that he can open a github issue if his desired OS-ARCH not available on releases.

I 'll make the adjustments and wait for your comments/approval

**Comment by thiagoftsm** [comment]
I began to test the PR, the first test that I did was to use kickstart.sh without any argument to check that it was compiling ok, and in this test I did not have any problem. I compiled on RHEL 7.7 .
But, the next test that I did showed a message that I was not hopping to see :

VirtualBox_redhat7_21_08_2019_11_28_35

I imagined that I would only see the error message, because  I did not give the correct input, but it tried to remove something before show the message and it raises an error on line 114. In my vision for this specific feature, it is necessary validate the arguments before to try to do anything.

File: packaging/installer/kickstart.sh [hunk 5]
```diff
@@ -293,6 +299,35 @@ while [ -n "${1}" ]; do
 	elif [ "${1}" = "--stable-channel" ]; then
 		RELEASE_CHANNEL="stable"
 		shift 1
+	elif [ "${1}" == "--local-tarball-override" ]; then
+		shift 1
+		if [ -z "${1}" ]; then
+			fatal "Missing netdata: Option --local-tarball-override requires extra information. The desired tarball for netdata, the checksumi, the go.d plugin tarball and the go.d plugin config tarball, in this particular order"
```
ilyam8: it is hard to understand the order when it is one line with `,` and `and`
Maybe numbered list?

thiagoftsm: The error on Debian 9.9 is different, this time it did not showed the error that I had on RHEL 7, but it did not stop the installation process, it began to compile with only one argument given:

VirtualBox_Debian9_21_08_2019_11_52_53

joelhans:
Tested the new OS/ARCH command and it works fine for me. Definitely a cleaner solution! Thanks for that. The docs side of this LGTM at this point.

paulkatsoulakis:
> I began to test the PR, the first test that I did was to use kickstart.sh without any argument to check that it was compiling ok, and in this test I did not have any problem. I compiled on RHEL 7.7 .
> But, the next test that I did showed a message that I was not hopping to see :
> 
> VirtualBox_redhat7_21_08_2019_11_28_35
> 
> I imagined that I would only see the error message, because I did not give the correct input, but it tried to remove something before show the message and it raises an error on line 114. In my vision for this specific feature, it is necessary validate the arguments before to try to do anything.

I notice the rm -f error over there.
It is because of ${sudo} being surounded by quotes. I will attend this, it is not directly related to this PR.

The second error you see @thiagoftsm, works as intended. You did not provide all required arguments. Kickstart.sh requires 4 arguments after the tarball override while kickstart-static64 requires two arguments.  Check the docs for details.

packaging/installer/kickstart.sh
paulkatsoulakis: wow, this is interesting @thiagoftsm, you need to provide me with the BASH versions from both cases

@ilyam8 adding comma separated list of items does not provide some value. space or comma, they are still an ordered list of items that the user must provide in the right order.  If we add `,` for separation, it will just make us do more processing to get what we want and validate

thiagoftsm: On RHEL 7.7 I am using bash version bash-4.2.46-33.el7 .
On Debian 9.9 I am using bash version bash-4.4-5.
Let me know case you need more information.

thiagoftsm:
> The second error you see @thiagoftsm, works as intended. You did not provide all required arguments. Kickstart.sh requires 4 arguments after the tarball override while kickstart-static64 requires two arguments. Check the docs for details.

I understood that it is necessary to have the 4 arguments, the second message is very clear, I only appended the image to demonstrate the problem, because I did not hope to see the first.
Until now I did not test what is expected by the script, I am digging your code and testing the possible errors in different distributions.

packaging/installer/kickstart.sh
paulkatsoulakis: ah, spotted the error for that one. My apologies, it should be single equality rather than double

paulkatsoulakis:
> > The second error you see @thiagoftsm, works as intended. You did not provide all required arguments. Kickstart.sh requires 4 arguments after the tarball override while kickstart-static64 requires two arguments. Check the docs for details.
> 
> I understood that it is necessary to have the 4 arguments, the second message is very clear, I only appended the image to demonstrate the problem, because I did not hope to see the first.
> Until now I did not test what is expected by the script, I am digging your code and testing the possible errors in different distributions.

Kindly test again. The first error should be attended now.
Also the other error on debian, was a miss on my side, it should have been single equal sign not double

thiagoftsm:
> Kindly test again. The first error should be attended now.
> Also the other error on debian, was a miss on my side, it should have been single equal sign not double

The error that we were talking before is not more present in the RHEL 7.7, so I decided to test another thing, I copied part of the line 128 from the README.md to execute, and when I ran the following command looks like that it is in an infinite loop:

VirtualBox_redhat7_22_08_2019_14_53_37

Case I run only `./kickstart-static64.sh` I can install Netdata wihout any problem.

paulkatsoulakis:
> ore present in the RHEL 7.7, so I decided to test another thing, I copied part of the line 128 from the README.md to execute, and when I ran the following command looks like that it is in an infinite loop:

yikes! `kickstart.sh` and `kickstart-static64.sh` have totally different way of using shift and when i fixed another issue, i basically raised that infinite loop, so silly!

You also help me spot another issue though, there is no --no-updates option for static64 kickstart.
I removed it from the documentation.

I 'll see how to mention on the readme what are the steps to manually enable automatic updating by setting up netdata updater on cron

packaging/installer/README.md
```diff
@@ -501,6 +504,67 @@ The Netdata team maintains two releases of the Netdata agent: **nightly** and **
...
 -   Retain more control over the Netdata version you use
 
+## Offline installations
+
+You can install Netdata on systems without internet access, but you need to take
+a few extra steps to make it work.
+
+By default, the `kickstart.sh` and `kickstart-static64.sh` download Netdata
+assets, like the precompiled binary and a few dependencies, using the system's
+internet connect, but you can also supply these files from the local filesystem.
```
cosmix: connect → connection

packaging/installer/README.md
```diff
@@ -501,6 +504,67 @@ The Netdata team maintains two releases of the Netdata agent: **nightly** and **
...
+# go.d plugin 
+# For binaries for OS types and architectures not listed on [go.d releases](https://github.com/netdata/go.d.plugin/releases/latest), kindly open a github issue and we will do our best to serve your request
+export OS=$(uname -s | tr '[:upper:]' '[:lower:]') ARCH=$(uname -m | sed -e 's/i386/386/g' -e 's/i686/386/g' -e 's/x86_64/amd64/g' -e 's/aarch64/arm64/g' -e 's/armv64/arm64/g' -e 's/armv6l/arm/g' -e 's/armv7l/arm/g' -e 's/armv5tel/arm/g') && curl -s https://api.github.com/repos/netdata/go.d.plugin/releases/latest | grep "browser_download_url.*${OS}-${ARCH}.tar.gz" | cut -d '"' -f 4 | wget -qi -
+
+# go.d configuration 
+curl -s https://api.github.com/repos/netdata/go.d.plugin/releases/latest | grep "browser_download_url.*config.tar.gz" | cut -d '"' -f 4 | wget -qi -
+```
+
+Move these files to the `/tmp` directory on the offline system in whichever way
+your organization allows.
```
cosmix: > your organization allows.

Refers to enterprise users, which is a bit presumptive. I would rephrase this.

packaging/installer/README.md
```diff
@@ -501,6 +504,67 @@ The Netdata team maintains two releases of the Netdata agent: **nightly** and **
...
+
+Now you can run either the `kickstart.sh` or `kickstart-static64.sh` scripts
+using the `--local-tarball-override` option. This option requires you to specify
+the location and names of the files you just downloaded. 
+
+!!! note When using `--local-tarball-override`, the `kickstart.sh` or
+    `kickstart-64.sh` scripts won't download any Netdata assets from the
```
cosmix: Did you mean `kickstart-static64.sh`?

packaging/installer/kickstart-static64.sh
```diff
@@ -1,7 +1,15 @@
...
 # SPDX-License-Identifier: GPL-3.0-or-later
 # shellcheck disable=SC1117,SC2039,SC2059,SC2086
-
+#
+#  Options to run
+#  --dont-wait              do not wait for input
+#  --non-interactive        do not wait for input
+#  --dont-start-it          do not start netdata after install
+#  --stable-channel         Use the stable release channel, rather than the nightly to fetch sources
+#  --local-tarball-override Use a manually provided tarball for the installation
```
cosmix: The option name is too long for no good reason. `--local-files` is good enough in my opinion.

packaging/installer/kickstart.sh
```diff
@@ -10,9 +10,10 @@
 # bash <(curl -Ss https://my-netdata.io/kickstart.sh) all
 #
 # Other options:
-#  --dont-wait       do not prompt for user input
-#  --non-interactive do not prompt for user input
-#  --no-updates      do not install script for daily updates
+#  --dont-wait                do not prompt for user input
+#  --non-interactive          do not prompt for user input
+#  --no-updates               do not install script for daily updates
+#  --local-tarball-override   set the full path of the desired tarball to run install with
```
cosmix: Again, I would use a shorter option name, probably the same one used in the static installer.

packaging/installer/kickstart.sh
```diff
@@ -311,10 +346,17 @@ dependencies
...
 
-set_tarball_urls "${RELEASE_CHANNEL}"
+if [ -z "${NETDATA_LOCAL_TARBALL_OVERRIDE}" ]; then
+	set_tarball_urls "${RELEASE_CHANNEL}"
+
+	download "${NETDATA_TARBALL_CHECKSUM_URL}" "${TMPDIR}/sha256sum.txt"
+	download "${NETDATA_TARBALL_URL}" "${TMPDIR}/netdata-latest.tar.gz"
+else
+	progress "Installation sources were given as input, running installation using \"${NETDATA_LOCAL_TARBALL_OVERRIDE}\""
+	run mv "${NETDATA_LOCAL_TARBALL_OVERRIDE_CHECKSUM}" "${TMPDIR}/sha256sum.txt"
```
cosmix: Maybe `cp` instead of `mv`ing and `rm` if the installation succeeds. This way, if there is a problem with the install, the user won't have to re-copy the files to `${TMPDIR}`. Typically files in `${TMPDIR}` will be cleared after sometime anyway (when depends on operating system/linux distro etc.) 

What do you think?

paulkatsoulakis: this is indeed a great catch.
as agreed on slack i 'll copy the files, but not remove them in the end since these are what the user downloaded and he may want to keep them around

packaging/installer/README.md
```diff
@@ -545,21 +545,21 @@ Move these files to the `/tmp` directory on the offline system in whichever way
...
+!!! note When using `--local-files`, the `kickstart.sh` or
     `kickstart-static64.sh` scripts won't download any Netdata assets from the
     internet. But, you may still need a connection to install dependencies using
     your system's package manager. The scripts will warn you if your system
     doesn't have all the dependencies.
 
 ```bash
 # kickstart.sh
-bash &lt;(curl -Ss https://my-netdata.io/kickstart.sh) --local-tarball-override /tmp/netdata.tar.gz /tmp/checksums.txt /tmp/go.d.binary.tar.gz /tmp/go.d.config.tar.gz
+bash &lt;(curl -Ss https://my-netdata.io/kickstart.sh) --local-files /tmp/netdata.tar.gz /tmp/checksums.txt /tmp/go.d.binary.tar.gz /tmp/go.d.config.tar.gz
```
ramotswa: Forgive my ignorance; but presumably this won't work if we're working on a disconnected environment? Shouldn't we be referencing the bash file in packaging/installer/ rather than the one online?

paulkatsoulakis: well, we give the users the shortcut to directly execute the above command, but in the offline mode they will just download and execute it manually.

It may make sense to add a note though

paulkatsoulakis: Actually, you just pinpointed something without realising it.
Kickstart has the dependency script that is running directly from the web, that will also need to be present for the process to work offline

paulkatsoulakis: coming back on the instructions, I don't know if we should be adjusting them to make them completely guided. To me it kind of makes sense that the user can simply download and run the kickstart if he can't directly execute the proposed command, otherwise we make our instructions too verbose imho

I think i would need commends from @joelhans on this one

ramotswa: Sure, as you deem fit. I think I’m just coming from a lazy perspective (although technical enough to figure out I can’t run those instructions). I would hope to be able to just copy and paste the instructions in the offline install section. In order for me to do this offline install I would love to:
1. Checkout / download master branch
2. Execute one bash file that will download all of my temporary files I need (possibly into an ‘offline install’ folder rather than /tmp) as well as copying any kickstart bash scripts to that folder. 
3. Copy that folder to the non internet machine.
4. Execute one bash script to install netdata using the pre-downloaded files

Might be too much for now, and I can def try and do this myself and contribute, but that would be my best case scenario. :)

paulkatsoulakis:
I will be merging the changes to get on time for the release.
@joelhans by all means please to propose changes on the documentation if you see fit as per the previous comments

packaging/installer/README.md
paulkatsoulakis: That is a good point actually, it just increases the scope of the change.
You are basically asking for a preparation script, which is a layer on top of the offline install of kickstart.

This script will gather all required elements for an offline install basically.
The `offline-install.sh` script would execute the following steps with the `--prepare` switch:

1) Download all required netdata items that kickstart now needs in a folder, including the kickstart.sh script
2) Download all needed binary packages into a folder from the respective repo (yum, apt, etc)

Then using the `--install` switch, it would:
1) Install offline packages to the system
2) run the kickstart with the right flags.


I think i should create a new issue for this, so that we can prioritise and do it.
It needs some thinking to cut the right corners

thiagoftsm:
Hi @paulkatsoulakis ,

After our talks I decided to test direct the new feature and I could detect that there is a real problem with `bash 4.2.46`, as you can see it does not the arguments in list:

VirtualBox_redhat7_02_09_2019_07_46_01

I am using `RHEL 7.7` yet.

paulkatsoulakis:
> Hi @paulkatsoulakis ,
> 
> After our talks I decided to test direct the new feature and I could detect that there is a real problem with `bash 4.2.46`, as you can see it does not the arguments in list:
> 
> VirtualBox_redhat7_02_09_2019_07_46_01
> 
> I am using `RHEL 7.7` yet.

Hi @thiagoftsm !
The `kickstart-static64.sh` does not take multiple arguments, that is why you get this error.
I will make an adjustment to provide a more informed error when the user provides more arguments than needed

cakrit:
I understand what you're saying, but the last line is still wrong:
```
bash < (curl -Ss https://my-netdata.io/kickstart.sh) --local-files /tmp/netdata.tar.gz /tmp/checksums.txt /tmp/go.d.binary.tar.gz /tmp/go.d.config.tar.gz
```

- How am I going to run this on a system with no internet connectivity?
- Where are the `kickstart-static64.sh` commands?

Moveover, will `kickstart.sh` actually not try to download dependencies when `--local-files` is provided? I didn't check the code, just asking about this one.

paulkatsoulakis:
> I understand what you're saying, but the last line is still wrong:
> 
> ```
> bash < (curl -Ss https://my-netdata.io/kickstart.sh) --local-files /tmp/netdata.tar.gz /tmp/checksums.txt /tmp/go.d.binary.tar.gz /tmp/go.d.config.tar.gz
> ```

Good point Chris, on the "Offline install" section we do not clarify that the script itself should be downloaded also.  I 'll go through the documentation one more time.

> 
> * How am I going to run this on a system with no internet connectivity?
> * Where are the `kickstart-static64.sh` commands?

On each section we have added explanations for the --local-files option and we have also the offline install section where we explain what is needed for each


> 
> Moveover, will `kickstart.sh` actually not try to download dependencies when `--local-files` is provided? I didn't check the code, just asking about this one.

Yes, this is another thing that we need to take care, that is why we introduced #6748 that will provide a true offline install toolkit.  This PR only provides the capability of running the kickstart without downloading any netdata assets. package management is not taken care of.

It is common that an isolated system, may still have controlled accessibility on a repository to fetch any packages that is why th einitial approach is like that.  We introduce #6748 to provide support for the fully isolated locked down environments

PR #6710, opened 2019-08-22
netdata/installer: fix static64 installer always overwriting configuration

### LINKED ISSUE #6692: Static distribution overwrites edited config files, ignores /etc/netdata
##### Bug report summary

When running the static version installer (`bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh)`), which is the recommended way to update, the installer overwrites my edited `netdata.conf` file with its empty one.

I've tried moving the config files I've edited to `/etc/netdata`, but the binary version appears to ignore this folder.

##### OS / Environment

```
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 9.9 (stretch)
Release:        9.9
Codename:       stretch
```

netdata v1.16.1-62-gd261783c

##### Component Name

Binary installer

##### Steps To Reproduce

```bash
# Install netdata
bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh)

# Get complete netdata.conf file and make changes
cd /opt/netdata/etc/netdata
curl http://localhost:19999/netdata.conf > netdata.conf
vi netdata.conf  # change something...

# Run installer again (i.e. update netdata)
bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh)

# Config file has been overwritten:
cat netdata.conf
# netdata can generate its own config which is available at 'http://<netdata_ip>/netdata.conf'
# You can download it with command like: 'wget -O /opt/netdata/etc/netdata/netdata.conf http://localhost:19999/netdata.conf'
```

##### Expected behavior

The installer should not simply overwrite configuration files the user has edited.

It should offer to either back up the existing config file, place the new one beside it (`netdata.conf.dist` or the like), or also read config from /etc/netdata and leave those files entirely alone.

paulkatsoulakis:
##### Summary
As per the #6692 issue, it has been identified that a BUG exists on the static installation regarding the preservation of existing configuration files when the installer runs for the second time.

After investigating the scenarios, it seems that the static installer was always re-creating the folder structure, thus resulting into always having a clean installation of netdata regardless of any customisations a user may have done in the config folder.

With this PR we attempt to mitigate the issue by touching two places:
1) Initial entry point script of makeself, to trigger the back-up of the configuration folder if it already exists

2) Adjust the install or update scriptlet and make it reinstate the existing folder and also trigger stock config files clean up (if the files weren't modified)

##### Component Name
netdata/installer

##### Additional Information
Fixes #6692

### DIFF
--- a/packaging/makeself/install-or-update.sh
+++ b/packaging/makeself/install-or-update.sh
@@ -10,6 +10,13 @@ umask 002
 renice 19 $$ >/dev/null 2>/dev/null
 
 # -----------------------------------------------------------------------------
+if [ -d /opt/netdata/etc/netdata.old ]; then
+	progress "Found old etc/netdata directory, reinstating this"
+	mv -f /opt/netdata/etc/netdata /opt/netdata/etc/netdata.new
+	mv -f /opt/netdata/etc/netdata.old /opt/netdata/etc/netdata
+	progress "Trigger stock config clean up"
+	rm -f /opt/netdata/etc/netdata/.installer-cleanup-of-stock-configs-done
+fi
 
 STARTIT=1
 
@@ -215,11 +222,6 @@ fi
 
 
 # -----------------------------------------------------------------------------
-if [ -d /opt/netdata/etc/netdata.old ]; then
-	progress "Found old etc/netdata directory, reinstating this"
-	mv -f /opt/netdata/etc/netdata /opt/netdata/etc/netdata.new
-	mv -f /opt/netdata/etc/netdata.old /opt/netdata/etc/netdata
-fi
 
 if [ ${STARTIT} -eq 0 ]; then
     create_netdata_conf "/opt/netdata/etc/netdata/netdata.conf"
--- a/packaging/makeself/install-or-update.sh
+++ b/packaging/makeself/install-or-update.sh
@@ -215,6 +215,11 @@ fi
 
 
 # -----------------------------------------------------------------------------
+if [ -d /opt/netdata/etc/netdata.old ]; then
+	progress "Found old etc/netdata directory, reinstating this"
+	mv -f /opt/netdata/etc/netdata /opt/netdata/etc/netdata.new
+	mv -f /opt/netdata/etc/netdata.old /opt/netdata/etc/netdata
+fi
 
 if [ ${STARTIT} -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


paulkatsoulakis:
> Centos 7.6.1810
> bash 4.2.46-31

My mistake @thiagoftsm i did not explain how to test the changes.
You need to do the following:
1) Run the command `./packaging/makeself/build-x86_64-static.sh`
2) Once it is done, execute the .run created inside `netdata/artifacts` folder that is created.

thiagoftsm:
Hi @paulkatsoulakis ,

When I tried to run the command that you said, the script said me something about a specific directory that I never created before, so my suggestion is we use the current directory that I am running instead one that looks like it is created for very specific machines or building process.

VirtualBox_Centos_27_08_2019_07_16_00

paulkatsoulakis:
> Hi @paulkatsoulakis ,
> 
> When I tried to run the command that you said, the script said me something about a specific directory that I never created before, so my suggestion is we use the current directory that I am running instead one that looks like it is created for very specific machines or building process.
> 
> VirtualBox_Centos_27_08_2019_07_16_00

My mistake that i did not clarify that, you need to run the command from netdata repository root directory, that is why i provided the execution like that

paulkatsoulakis:
So the error proved to be selinux issue.
Related stackoverflow that helped pin it: https://stackoverflow.com/questions/24288616/permission-denied-on-accessing-host-directory-in-docker

thiagoftsm:
@paulkatsoulakis the tip that you gave me to run `setenforce 0` solved the problem with doker and I could run the tests.

When I installed the first time I got a clean netdata.conf as expected, after this I copied the netdata.conf from `system/netdata.conf` to /opt/netdata/etc/netdata/ . Finally I run the installer again and it kept the netdata.conf that I had copied, to confirm that everything was ok, I decided to run one more time, and this time it created a clean netdata.conf, so the problem persists for more than one installation process.

paulkatsoulakis:
The repeating execution of the .run installer, was producing a false behaviour because of the last commit that fixed the bug.

I tried now the following scenarios that are working fine:
1) multiple runs maintain the modified configs in etc
2) any unmodified stock config on etc folder is cleaned up
3) netdata starts successfully after completion

Issue #2781, opened 2017-09-22
configure external commands (python, node.js, etc) in netdata.conf


magao:
The command necessary to run a plugin may not be on the path, or you may want to override the default (e.g. use `python3` in preference to `python`, or possibly `pypy`).

I would suggest another option for each plugin:

```
[plugin:python.d]
	# command =
```

which gets passed to the plugin somehow. Don't really want to change the command line, so maybe set an environment variable. If non-empty, the plugin should try to use it before the default commands. For example, for `python.d`:

```
#!/usr/bin/env bash
'''':; exec "$(command -v "${NETDATA_PLUGIN_PYTHOND_COMMAND}" || command -v python || command -v python3 || command -v python2 || echo "ERROR python IS NOT AVAILABLE IN THIS SYSTEM")" "$0" "$@" # '''
```

l2isbad:
https://github.com/firehol/netdata/issues/1731#issuecomment-277767601

magao:
Whilst that method works, it's pretty cumbersome, especially if you're changing it fairly commonly (e.g. testing against various python versions). Using a dedicated environment variable I can just do:

`NETDATA_PLUGIN_PYTHOND_COMMAND=python3 /opt/NetData/libexec/netdata/plugins.d/python.d.plugin myplugin.chart.py debug`

to test against python3.

l2isbad:
> if you're changing it fairly commonly (e.g. testing against various python versions)

Well, is such cases better will be (not supported by current python.d.plugin but it's easy to fix)

`python2.7 python.d.plugin debug 1 myplugin.chart.py`
`python3.4 python.d.plugin debug 1 myplugin.chart.py`

ktsaou:
We could add a few `netdata.conf` for `python`, `node.js`, `bash`, `curl`, `fping`, `sendmail`, etc and set a few environment variables for all the plugins to use the netdata.conf configured ones.

I think we will eventually need this functionality, especially if we move certain features of the external scripts into netdata.

I'll leave this as enhancement.

ktsaou:
@cakrit 	another one for controlling external plugin modules from `netdata.conf`.
The idea of having a `netdata.conf` called `[environment]` to allow configuring external plugins env variables is also nice (today, such variables are scattered in netdata.conf).

cakrit:
Leaving it on me until we have someone to work on this.

ilyam8:
[...]

PR #6736, opened 2019-08-29
Configure calling of python.d from netdata.conf

### LINKED ISSUE #1731: python.d.plugin crashes with RHEL 5
Because RHEL  uses phython 2.4, python.d.plugin does not work.

It is possible to install python26 in RHEL 5, but it is not trivial to make it default. 

So, to workaround the problem I installed **python26**, **python26-requests** and **python26-mysqldb.** 

And changed the second line of **python.d.plugin** 

```bash
'''':; exec "$(command -v python || command -v python3 || command -v python2 || echo "ERROR python IS NOT AVAILABLE IN THIS SYSTEM")" "$0" "$@" # '''
```
to 
```bash
'''':; exec "$(command -v python26 || command -v python3 || command -v python2 || echo "ERROR python IS NOT AVAILABLE IN THIS SYSTEM")" "$0" "$@" # '''
```

Just leaving this here as it could be useful to someone else

amoss:
Fixes #1731 (without links or path changes).
Fixes python part of #2781.

##### Summary
Allows configuration of the python.d launch from netdata.conf

##### Component Name
area/daemon

##### Additional Information
Adds a new switch to the python.d plugin script to allow an explicit choice of the python interpreter binary.
Adds a logging handler to python.d to allow log to be written to disk. This is disabled by default and enabled with a command-line option to preserve original behaviour.


amoss:
It looks as if the Travis error exists against your main branch already - I don't think it is caused by these commits as it looks like a problem setting the container up for the test.

cakrit:
Hi @amoss , @ilyam8 will review it shortly. Can you please check the Codacy warnings?

amoss:
I can see that there is 1 warning from Codacy but it does not show me what it is in the Codacy interface. The logs do not show up either, it just says that it cannot analyse it.

When I put the PR in there was a different error about a build failing that was already in a failing state, but that one has disappeared.

Does it show you what the warning message is?

ilyam8:
Hi @amoss, thx for contributing!

> Adds a new switch to the python.d plugin script to allow an explicit choice of the python interpreter binary.

Please add to the OP how it is suppossed to work. Where and what syntax?

From the code i see that it:
 - must be 2nd argument in cmd
 - must be 1st argument `command options` in `netdata.conf`
 - must starts with `-p`

> ../python.d.plugin 1 -ppython3 example

Looks a little bit lame to me. Maybe use some env_var would be better? what do you think?


> Adds a logging handler to python.d to allow log to be written to disk.

Currently netdata is responsible for writing logs to disk/other destination, i think this functionality should be there. Otherwise we need to implement it in every external collector.

____

And i did some (a log) changes to python.d in https://github.com/netdata/netdata/pull/6661 and it will be merged soon, so you will need to do a rebase after it is merged.

amoss:
> Looks a little bit lame to me. Maybe use some env_var would be better? what do you think?

There is already an environment variable to control the path, but not the explicit choice of binary. I think it is easier to write the binary into the configuration file (where it is visible and easy to update) instead of changing the environment of the netdata user. Updates in the login files of daemon users are harder to keep track of when checking the state of a server.

> Currently netdata is responsible for writing logs to disk/other destination, i think this functionality should be there. Otherwise we need to implement it in every external collector.

Currently it seems that the logs from python.d just get dropped instead of being passed back to netdata. Putting the functionality in netdata would mean:

1. Using the python logging daemon to format the logs and sending them back to netadata over a pipe.
2. Implementing some code in netdata that read in the logs so that they can be written out to disk.

It seems a little redundant to do it this way (basically two logging daemons), what would be the benefit?

> And i did some (a log) changes to python.d in #6661 and it will be merged soon, so you will need to do a rebase after it is merged.

I'll take a look at the changes and do a rebase.

ilyam8:
> Currently it seems that the logs from python.d just get dropped instead of being passed back to netdata. 

What do you mean? They are in `error.log`.

amoss:
> What do you mean? They are in error.log.

I'll double check that and get back to you - but I don't see any of the python.d logging in error.log on either of the netdata installations that I've tried.

ilyam8:
Current flow is - python.d writes logs to stderr, netdata writes them to disk. This is true for all external collectors.

amoss:
This is not what I've seen during testing, for example the web_log plugin does not like tab characters in its config file. When I saw this problem it was not written into the error.log at all I had to follow the debugging advice of runnning python.d separately under the netdata user to see the error message. But again - I'll double check this on a live installation and get back to you with the result.

amoss:
Ok, I went back and checked. All of the python.d logging is being written into error.log as you said. I must have had some weird configuration problem when I tested it.

In that case the logging to the separate python.d file is completely redundant and I will rip it back out. This just leaves the command line passing. 

The `-p` is to indicate that it is an explicit path, it lands second because the first parameter is positional and everything after that is being passed into the python argv untouched. I can change it easyily to change it so that it can appear in any order in the optional args so that e.g. `../python.d.plugin 1 trace debug -ppython3` would work or to add a long-form version e.g. `./python.d.plugin 1 --path=python3` if you want?

ilyam8:
@amoss 	

> optional args so that e.g ../python.d.plugin 1 trace debug -ppython3 would work 
> add a long-form version e.g. ./python.d.plugin 1 --path=python3 if you want

i think 1st one is better.

PR #6743, opened 2019-08-30
netdata/packaging: fix coverity problem in travis (2 issues)

paulkatsoulakis:
##### Summary
1) Add slack notification when things fail, we dont want to silently fail it
2) Change the tarball version, seems we now downloading a different version

##### Component Name
netdata/ci

##### Additional Information
None


paulkatsoulakis:
Overriding approvals, this is a simple CI change i 'll be closely monitoring

Issue #6734, opened 2019-08-28
sidebar-info update - DB engine


jacekkolasa:
##### Feature idea summary
We should remove `Netdata is using ** of memory on **for ** minutes and ** seconds of real-time history.` from the info block on the right-side sidebar. That info is no more valid.

Also, we shouldn't show the tip to use DB engine, if user is already using it.

related to https://github.com/netdata/netdata/issues/6587

jacekkolasa:
@joelhans cc: @cakrit 
We needed to remove `Netdata is using ** of memory on **for ** minutes and ** seconds of real-time history.` sentence because that is not true (the memory and time is wrong), but perhaps we shouldn't remove the `host` information, what do you think? If you agree, could you please provide the changed copy of:

> Every second, Netdata collects # metrics, presents them in # charts and monitors them with # alarms. Netdata is using # MB of memory on HOSTNAME for # hour, # minutes, and # seconds of real-time history.
without the last sentence but with hostname info?

joelhans:
Perhaps it's as simple as this?

> Every second, Netdata collects 2,087 metrics on **HOSTNAME**, presents them in 296 charts, and monitors them with 80 alarms.

While you're mucking with this section, could you add a comma between **charts** and **and**? So, it would read: `...presents them in # charts, and monitors them...`

I'm a big fan of the Oxford comma...

jacekkolasa:
I love the additional coma, I would never come up with that, I'm glad I've asked you to look at it :) done

joelhans:
Well, it was my error to begin with, so this was a good opportunity to fix it!

PR #6744, opened 2019-08-30
[WiP] sidebar-info update - DB engine

jacekkolasa:
solves #6734 

##### Summary
User should not see the DB tip on the sidebar, when he/she already configured db engine. (on the host level, for now)
Additionally, `Netdata is using ** of memory on **for ** minutes and ** seconds of real-time history.` sentence should be removed

##### Component Name
Web


web/gui/main.js
```diff
@@ -1786,21 +1786,26 @@ function renderPage(menus, data) {
...
-      data.alarms_count.toLocaleString() + '</b> alarms. Netdata is using ' +
-      Math.round(data.rrd_memory_bytes / 1024 / 1024).toLocaleString() + ' MB of memory on <b>' +
-      data.hostname.toString() + '</b> for ' +
-      NETDATA.seconds4human(data.update_every * data.history, {
-        minute: 'minute', minutes: 'minutes', second: 'second', seconds: 'seconds', space: '&nbsp;',
-      }) +
-      ' of real-time history.<br />&nbsp;<br />' + 'Get more history by ' +
-      '<a href="https://docs.netdata.cloud/docs/configuration-guide/#increase-the-metrics-retention-period" target=_blank>configuring Netdata\'s <b>history</b></a> or using the <a href="https://docs.netdata.cloud/database/engine/" target=_blank>DB engine.</a>' +
-      '<br/>&nbsp;<br/><b>netdata</b><br/>' + data.version.toString() + '</small></li>';
+      'Netdata collects <b>' + data.dimensions_count.toLocaleString() + '</b> metrics on ' +
```
cosmix: Use of the `b` tag is discouraged in the HTML5 spec. Please replace with `strong` or `em` if you want to emphasize.

thiagoftsm: Considering @cosmix last comment, would not be better to use CSS to do this?

jacekkolasa: @thiagoftsm you're absolutely right, however in our case there's no separate styles for this, just bootstrap.css adds
```
b,
strong {
  font-weight: bold;
}
```
Done, but i've only pushed to the modified fragment. There are 20 more cases in main.js, and even if i'm 99% sure nothing bad would happen, i would prefer to change it on refractor phase (also for rebase reasons)

jacekkolasa: tested in chrome, safari, ff

thiagoftsm: Thank you by the explanation!

Issue #6741, opened 2019-08-29
[/proc/stat cpuidle] Avoid isolated CPUs


Saruspete:
##### Bug report summary
When configured on a typical "real-time" setup, the CPU Idle module (now integrated in proc.plugins (PR #4635) ) freeze the netdata daemon.

On Linux, there is a [boot-param](https://www.kernel.org/doc/Documentation/admin-guide/kernel-parameters.txt) "isolcpus" which removes given CPUs to the default scheduling affinity. With nohz_full opt, it also disables all scheduling tick on selected CPUs.
So, no process will be able to run on the listed cpus, unless explicitely granted with `/usr/bin/taskset` or syscall `sched_setaffinity()` (which is called by `pthread_setaffinity_np()` ).

Usually, we isolate these CPUs to dedicate them to a specific latency critical process, which is running in a spinning loop (infinite non-blocking poll) + realtime scheduling policy.

Since PR #4635  cpuidle plugin has been merged inside proc.plugins, which is also part of core netdata daemon. When cpuidle [schedules one of its thread](https://github.com/netdata/netdata/blame/v1.16.0/collectors/proc.plugin/proc_stat.c#L993-L1003) to an isolated cpu, the thread hangs (never scheduled) and netdata can't be stopped / killed (until the blocked thread gets a higher priority than the spinning process).

This hang is normal and expected as the Kernel Scheduler has the explicit order (through realtime policy) to never de-schedule the realtime process (which is spinning in an infinite polling loop). 
That's why we isolate the CPU from the standard scheduling policy through `isolcpus` bootparam.

It's not a bug in netdata per-se, but rather a lack of environment awareness. 

##### OS / Environment
Linux / Redhat 6/7 , Fedora 30, (or any version).

You might need the following sysctl to be effective:
```
kernel.sched_rt_runtime_us = 1000000
```

This sysctl is usually to 950000, which deschedule any realtime process, to let other non-prioritized process to run nonetheless.

##### Netdata version (ouput of `netdata -V`)

> netdata v1.16.0

##### Component Name
proc.plugins

##### Steps To Reproduce

To reproduce on a standard desktop (to ack bug):
1. spawn an infinite loop on CPU3 (or any other):  `taskset -c 3 chrt -r 1 bash -c 'while :; do :; done'`
2. Now every process that will go on CPU3 will hang, stuck in "running" state. Try: `taskset -c 3 bash`

here's a typical Realtime setup, on a linux machine (isolating only cpu3) :
[...]

PR #6745, opened 2019-08-30
[libnetdata/thread] Set thread name from tag

Saruspete:
When calling netdata_thread_create, you can provide a tag to identify
the thread in the error message
This patch uses this tag to rename the thread if supported by the
pthread distribution to allow better identification of the thread.

This have helped identify the bug #6741

Sample output:

```
ps -L -o tid,pid,comm,args -p $(pgrep netdata)
  TID   PID COMMAND         COMMAND
 5123  5123 netdata         ./netdata -D
 5133  5123 PLUGIN[proc]    ./netdata -D
 5134  5123 netdata         ./netdata -D
 5135  5123 PLUGIN[cgroups] ./netdata -D
 5136  5123 netdata         ./netdata -D
 5137  5123 STATSD          ./netdata -D
 5138  5123 BACKENDS        ./netdata -D
 5139  5123 netdata         ./netdata -D
 5140  5123 PLUGINSD        ./netdata -D
 5141  5123 netdata         ./netdata -D
 5142  5123 HEALTH          ./netdata -D
 5143  5123 netdata         ./netdata -D
 5145  5123 netdata         ./netdata -D
 5146  5123 PLUGINSD        ./netdata -D
 5147  5123 PLUGINSD[apps]  ./netdata -D
 5148  5123 netdata         ./netdata -D
 5150  5123 netdata         ./netdata -D
 5151  5123 STATSD          ./netdata -D
```

### DIFF
--- a/libnetdata/threads/threads.c
+++ b/libnetdata/threads/threads.c
@@ -112,27 +112,25 @@ static void thread_cleanup(void *ptr) {
 static void thread_set_name(NETDATA_THREAD *nt) {
 
     if (nt->tag) {
-        char *threadname = strdupz(nt->tag);
         int ret = 0;
 
-        if (threadname) {
-            // Name is limited to 16 chars
-            threadname[15] = 0;
+        // Name is limited to 16 chars
+        char threadname[16];
+        strncpyz(threadname, nt->tag, 15);
+
 #if defined(__FreeBSD__)
-            pthread_set_name_np(pthread_self(), threadname);
+        pthread_set_name_np(pthread_self(), threadname);
 #elif defined(__APPLE__)
-            ret = pthread_setname_np(threadname);
+        ret = pthread_setname_np(threadname);
 #else
-            ret = pthread_setname_np(pthread_self(), threadname);
+        ret = pthread_setname_np(pthread_self(), threadname);
 #endif
 
-            if (ret != 0)
-                error("cannot set pthread name of %d to %s. ErrCode: %d", gettid(), threadname, ret);
-            else
-                info("set name of thread %d to %s", gettid(), threadname);
+        if (ret != 0)
+            error("cannot set pthread name of %d to %s. ErrCode: %d", gettid(), threadname, ret);
+        else
+            info("set name of thread %d to %s", gettid(), threadname);
 
-            free(threadname);
-        }
     }
 }
 
--- a/libnetdata/threads/threads.c
+++ b/libnetdata/threads/threads.c
@@ -109,7 +109,7 @@ static void thread_cleanup(void *ptr) {
     netdata_thread = NULL;
 }
 
-static void *thread_set_name(NETDATA_THREAD *nt) {
+static void thread_set_name(NETDATA_THREAD *nt) {
 
     if (nt->tag) {
         char *threadname = strdupz(nt->tag);
--- a/libnetdata/threads/threads.c
+++ b/libnetdata/threads/threads.c
@@ -119,7 +119,7 @@ static void thread_set_name(NETDATA_THREAD *nt) {
             // Name is limited to 16 chars
             threadname[15] = 0;
 #if defined(__FreeBSD__)
-            ret = pthread_set_name_np(pthread_self(), threadname);
+            pthread_set_name_np(pthread_self(), threadname);
 #elif defined(__APPLE__)
             ret = pthread_setname_np(threadname);
 #else
--- a/libnetdata/threads/threads.c
+++ b/libnetdata/threads/threads.c
@@ -109,6 +109,33 @@ static void thread_cleanup(void *ptr) {
     netdata_thread = NULL;
 }
 
+static void *thread_set_name(NETDATA_THREAD *nt) {
+
+    if (nt->tag) {
+        char *threadname = strdupz(nt->tag);
+        int ret = 0;
+
+        if (threadname) {
+            // Name is limited to 16 chars
+            threadname[15] = 0;
+#if defined(__FreeBSD__)
+            ret = pthread_set_name_np(pthread_self(), threadname);
+#elif defined(__APPLE__)
+            ret = pthread_setname_np(threadname);
+#else
+            ret = pthread_setname_np(pthread_self(), threadname);
+#endif
+
+            if (ret != 0)
+                error("cannot set pthread name of %d to %s. ErrCode: %d", gettid(), threadname, ret);
+            else
+                info("set name of thread %d to %s", gettid(), threadname);
+
+            free(threadname);
+        }
+    }
+}
+
 static void *thread_start(void *ptr) {
     netdata_thread = (NETDATA_THREAD *)ptr;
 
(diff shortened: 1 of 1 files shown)


mfundul:
According to [this](https://stackoverflow.com/a/7989973) there are similar solutions for linux, BSD and Mac OS. However, since OS X needs the thread itself to call such a function you should move your code to run inside the newly created thread context itself. This way, you can add the case for OS X and that should cover all kernels supported by netdata.

Saruspete:
Yes, that was the post I based myself onto. 

I willingly didn't handle the MacOS case, as it requires to duplicate the same code to put it in all calling start functions: 
- socket_listen_main_static_threaded_worker,
- st->start_routine (proc_main cgroups_main, cpuidlejitter_main check_main statsd_main macos_main diskspace_main freebsd_main pluginsd_main health_main)
- statsd_collector_thread
- pluginsd_worker_thread
- rrdpush_sender_thread
- rrdpush_receiver_thread

I'm not feeling confident enough to change so many places, just to handle one corner case, in a corner os for server.
Also, this kind of code would need to be re-added by every new thread.

mfundul:
> 
> 
> Yes, that was the post I based myself onto.
> 
> I willingly didn't handle the MacOS case, as it requires to duplicate the same code to put it in all calling start functions:
> 
>     * socket_listen_main_static_threaded_worker,
> 
>     * st->start_routine (proc_main cgroups_main, cpuidlejitter_main check_main statsd_main macos_main diskspace_main freebsd_main pluginsd_main health_main)
> 
>     * statsd_collector_thread
> 
>     * pluginsd_worker_thread
> 
>     * rrdpush_sender_thread
> 
>     * rrdpush_receiver_thread
> 
> 
> I'm not feeling confident enough to change so many places, just to handle one corner case, in a corner os for server.
> Also, this kind of code would need to be re-added by every new thread.

I believe adding it in `static void *thread_start(void *ptr)` covers all those cases.

Saruspete:
Indeed, misread the flow... sorry.

Reused the same listing as in  libnetdata.h: 
```
#if defined(__FreeBSD__)
#include <pthread_np.h>
#define NETDATA_OS_TYPE "freebsd"
#elif defined(__APPLE__)
#define NETDATA_OS_TYPE "macos"
#else
#define NETDATA_OS_TYPE "linux"
#endif /* __FreeBSD__, __APPLE__*/
```

Reread the doc and fixed the limit of 15 chars. Now all threads have a name (or truncated to fit in 15 chars)

libnetdata/threads/threads.c
```diff
@@ -121,6 +121,26 @@ static void *thread_start(void *ptr) {
...
+            if (pthread_setname_np(threadname) != 0)
+                error("cannot set pthread name of %d to %s", gettid(), threadname);
+#else
+            if (pthread_setname_np(pthread_self(), threadname) != 0)
+                error("cannot set pthread name of %d to %s", gettid(), threadname);
+#endif
+            info("set name of thread %d to %s", gettid(), threadname);
+            free(threadname);
+        }
+    }
```
ilyam8: I think it would be nice to move set tag stuff to a separate func. @mfundul

mfundul: It would be nice I agree, I did not specifically ask it cause it's only one place for now, but I also thought of it.

Saruspete: Created dedicated static func `thread_set_name` + added return code in log message if error.

mfundul:
According to [this](https://www.freebsd.org/cgi/man.cgi?query=pthread_set_name_np&sektion=3) you would also need the header file for FreeBSD:
```
     #include <pthread_np.h>
```

Saruspete:
Travis errors unrelated to commit.

`pthread_np.h` is already included for FreeBSD in `libnetdata.h` line 42, itself included by `thread.c` line 3

Saruspete:
can someone request travis rebuild ?

libnetdata/threads/threads.c
```diff
@@ -109,6 +109,33 @@ static void thread_cleanup(void *ptr) {
     netdata_thread = NULL;
 }
 
+static void thread_set_name(NETDATA_THREAD *nt) {
+
+    if (nt->tag) {
+        char *threadname = strdupz(nt->tag);
```
mfundul: The name should point directly to the string instead of copying it, i.e.
```
         char *threadname = nt->tag; 
```
The tag itself has already been `strdupz`ed in `netdata_thread_create()` and in any case this function calls `free(threadname);` which would corrupt `threadname` if it was necessary to duplicate it.

So you should also remove:
```
             free(threadname); 
```

libnetdata/threads/threads.c
```diff
@@ -109,6 +109,33 @@ static void thread_cleanup(void *ptr) {
...
+        int ret = 0;
+
+        if (threadname) {
+            // Name is limited to 16 chars
+            threadname[15] = 0;
+#if defined(__FreeBSD__)
+            ret = pthread_set_name_np(pthread_self(), threadname);
```
mfundul: The `pthread_set_name_np()` function has a `void` return type so you should remove `ret = ` because it breaks compilation in FreeBSD.

Saruspete: Done. Seems pthread on apple is returning int too.

It's a shame travis doesn't provide freebsd env for testing :(

mfundul: At the moment I'll have to manually test freebsd and ask somebody else to do the same for OS X.

libnetdata/threads/threads.c
mfundul: Can you explain your rationale here because you resolved this conversation but I still don't understand why. What I'm asking for is to remove the `strdupz()` and `free()` calls from `thread_set_name()`.

Saruspete: My gh page didnt' get updated when I woke it from sleep, sorry.
Because of the limit of 15chars for `pthread_setname_np` I have to force a null line 120: `threadname[15] = 0;`, else the setname call fails with `ERANGE`

As it's a reference to a netdata_thread, I don't want to change the tag value (just for a limitation from the pthread).
Hence, I have to copy it, split it to 15chars, and free it.

mfundul: Ok I see. In that case I would prefer we not mess with heap memory any more than necessary (too much `malloc()` fragmentation), and you would also need to use `freez()` instead of `free()` for consistency reasons.

Instead I propose you use the stack and one of netdata's internal functions that take care of the trailing `'\0'`, that would be [strncpyz](https://github.com/netdata/netdata/blob/0fca36fd596e49395141561bc42c168a240f51b8/libnetdata/inlined.h#L238).

How about doing this:
```
char threadname[16];
strncpyz(threadname, tag, 15);
```
And also remove  `free()` and `threadname[15] = 0`?

Saruspete: Much better indeed !

mfundul:
This only needs OS X testing before I approve and merge it. @ilyam8 @cosmix ?

cosmix:
@mfundul well, thread information is not readily exposed in process status tools on macOS, so it's largely irrelevant on that OS.

PR #6747, opened 2019-08-30
make coverity-scan.sh usable by hand

ktsaou:
`./coverity-scan.sh` should be easily usable by hand.

Prior to this PR, developers should remember to export certain variables (`REPOSITORY` and `COVERITY_SCAN_SUBMIT_MAIL`). They were given to option to configure `COVERITY_SCAN_TOKEN` and configure the path of the `cov-build` by configuring separate files for each of them. So, a developer was not able to just configure its environment and then just run the script every time it is needed.

This PR allows devs to create the file `.coverity-scan.conf` with the following information:

```sh
# the repository to report to coverity - devs can set here their own fork
REPOSITORY="netdata/netdata"

# the email of the developer, as given to coverity
COVERITY_SCAN_SUBMIT_MAIL="you@example.com"

# the token given by coverity to the developer
COVERITY_SCAN_TOKEN="TOKEN taken from Coverity site"

# the absolute path of the cov-build - optional
COVERITY_BUILD_PATH="/opt/cov-analysis-linux64-2019.03/bin/cov-build"

# when set, the script will print on screen the curl command that submits the build to coverity
# this includes the token, so the default is not to print it.
COVERITY_SUBMIT_DEBUG=1
```

The script is also updated to configure netdata with the following:

```

	--enable-https \
	--enable-jsonc \
	--enable-plugin-nfacct \
	--enable-plugin-freeipmi \
	--enable-plugin-cups \
	--enable-backend-prometheus-remote-write \

```

Ideally, the script should build all the features of netdata, including:

```

	--enable-plugin-xenstat \
	--enable-backend-kinesis \
	--enable-backend-mongodb \

```

@paulkatsoulakis I don't know if we should enable these on travis.

Also:

- `coverity-install.sh` is not needed any more. Run `./coverity-scan.sh install` to install coverity build tools. @paulkatsoulakis please update travis and then simply remove `coverity-install.sh` from the repo (the update version of this script just calls `./coverity-scan.sh install`.
- the installation will fail if the downloaded version is not the expected one
- the scan will fail if the expected version of coverity is not installed
- the installer `functions.sh` is used to simplify the script and beautify the output

### DIFF
--- a/coverity-scan.sh
+++ b/coverity-scan.sh
@@ -26,6 +26,12 @@
 # COVERITY_SUBMIT_DEBUG=1
 #
 # All these variables can also be exported before running this script.
+#
+# If the first parameter of this script is "install",
+# coverity build tools will be downloaded and installed in /opt/coverity
+
+# the version of coverity to use
+COVERITY_BUILD_VERSION="cov-analysis-linux64-2019.03"
 
 source packaging/installer/functions.sh || exit 1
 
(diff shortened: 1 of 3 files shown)


coverity-scan.sh
```diff
@@ -1,62 +1,167 @@
...
+  covbuild="$(which cov-build 2>/dev/null || command -v cov-build 2>/dev/null)"
+  if [ -z "$covbuild" ]; then
+    fatal "Failed to install coverity."
+  fi
+
+  progress "Coverity scan tools are installed."
+  return 0
+}
+
+if [ "${1}" = "install" ]
```
paulkatsoulakis: Would it make sense to use a `--with-install` option instead that will just run the install together with the run?

paulkatsoulakis:
> 	--enable-plugin-xenstat \
> 	--enable-backend-kinesis \
> 	--enable-backend-mongodb \

Yep, i need to add a few different combinations of configure options to be tested rather than a single one that we have now.  I 'll get to it as soon as we merge this

coverity-scan.sh
```diff
@@ -1,62 +1,167 @@
...
+
+source packaging/installer/functions.sh || exit 1
+
+cpus=$(find_processors)
 [ -z "${cpus}" ] && cpus=1
 
+if [ -f ".coverity-scan.conf" ]
+then
+	source ".coverity-scan.conf" || exit 1
+fi
```
paulkatsoulakis: I would probably write it as: `[ -f ".coverity-scan.conf" ] && source ".coverity-scan.conf"`, but that may be just matter of personal preference.

Also instead of the `exit 1` logic, is better to just set `set -e` to enable strict error checking and just use the or option for command we don't want them to stop the process if they fail. what do you think?

paulkatsoulakis:
BTW we can just merge this and let me take over the rest of the work, but from the commit flow i understand you still working on it

paulkatsoulakis:
Taking over as agreed, will come with follow up PR to wrap this up

PR #6668, opened 2019-08-15
Update README.md for pfSense

swordfish6975:
Just installed it this is what I had to do.

PR #6768, opened 2019-09-03
(Documentation) fix pfsense instructions and links

Fohdeesha:
The repo links for the pfsense install section were broken again, so I updated them. I also cleaned up some of the grammar and instructions - the note about editing the package URLs to substitute either i386 or a different freebsd version were removed, as pfsense does not support i386, and only supports freebsd 11.  I did keep (and expand) the note about potentially needing to edit the URL when a new version of netdata is released, as we will probably forget to edit this document URL every time a new version of netdata is added to freebsd's repo.

I realize the latest netdata release is 1.17, however the 1.15 URL in this PR is the latest available from the freebsd repo. It also now requires the package libuv, so that instruction was added as well.


Fohdeesha:
Just noticed there's another PR that implements part of what I did as well: https://github.com/netdata/netdata/pull/6668

I added his bind instruction, as well as added the suggestions made there by @joelhans - so hopefully this PR alone should be enough to fix the pfsense section

packaging/installer/README.md
```diff
@@ -342,28 +342,31 @@ cd netdata
 
 ##### pfSense
 
-To install Netdata on pfSense run the following commands (within a shell or under Diagnostics/Command Prompt within the pfSense web interface).
+To install Netdata on pfSense, run the following commands (within a shell or under the **Diagnostics/Command** prompt within the pfSense web interface).
 
-Change platform (i386/amd64, etc) and FreeBSD versions (10/11, etc) according to your environment and change Netdata version (1.10.0 in example) according to latest version present within the FreeSBD repository:-
-
-Note first three packages are downloaded from the pfSense repository for maintaining compatibility with pfSense, Netdata is downloaded from the FreeBSD repository.
+Note the first four packages are downloaded from the pfSense repository for maintaining compatibility with pfSense, Netdata and Python are downloaded from the FreeBSD repository.
```
cosmix: → Note that the first four packages […]

packaging/installer/README.md
```diff
@@ -342,28 +342,31 @@ cd netdata
...
-pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.13.0.txz
+pkg install libuv
+pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/python36-3.6.9.txz
+pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.15.0.txz
 ```
+**Note:** If you receive a ` Not Found` error during the last two commands above, you will either need to manually look in the [repo folder](http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/) for the latest available package and use its URL instead, or you can try manually changing the netdata version in the URL to the latest version.  
+
+You must edit `/usr/local/etc/netdata/netdata.conf` and change `bind to = 127.0.0.1` to `bind to = 0.0.0.0`.
+
+Now to start Netdata manually, run `service netdata onestart`
```
cosmix: Consider revising: I would suggest either `Now, to start Netdata manually […]` or `To start Netdata manually […]`.

packaging/installer/README.md
Fohdeesha: fixed

packaging/installer/README.md
Fohdeesha: fixed

PR #6781, opened 2019-09-04
Changes to launching the python.d plugin aggregator.

### LINKED ISSUE #1731: python.d.plugin crashes with RHEL 5

amoss:
Passing python.d arguments from netdata.conf

Path to the specific python interpreter can be passed on the command-line.
Reused existing configuration options, changed the plugin script to match. To
specify a python binary use the -p switch. The existing pythond options can
be passed in, for example:

[plugin:python.d]
        command options = -p/usr/local/bin/python2.7 trace debug

##### Summary
Allows configuration of the python.d launch from netdata.conf

##### Component Name
area/daemon

##### Additional Information
@ilyam8 --> Replaces PR#6736 (it auto-closed when I rebased the master and I cannot reopen it).
Fixes #1731 (without links or path changes).
Fixes python part of #2781.

I've stripped out the logging changes and rebased the fork against the current tip. It should not interact with your changes in #6661 any longer.

The argument with the -p switch is filtered out from the arguments passed to the python interpreter so it is now position independent.


ilyam8:
lgtm 👍

cakrit:
I resolved the conflict. @ilyam8 plz check and merge if all good.

ilyam8:
thank you @amoss

PR #6794, opened 2019-09-05
update grep to be more specific

n0coast:
when grepping for openfiles the netdata example is not exact enough and will return results from too many processes, just adding `.pid` to the grep _should_ pick up the correct process.

### DIFF
--- a/docs/high-performance-netdata.md
+++ b/docs/high-performance-netdata.md
@@ -126,7 +126,7 @@ You can check limits with following commands:
 
 ```sh
 cat /proc/$(ps aux | grep "nginx: master process" | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
-cat /proc/$(ps aux | grep "netdata.pid" | head -n1 | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
+cat /proc/$(ps aux | grep "\/netdata " | head -n1 | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
 ```
 
 View of the files:
--- a/docs/high-performance-netdata.md
+++ b/docs/high-performance-netdata.md
@@ -126,7 +126,7 @@ You can check limits with following commands:
 
 ```sh
 cat /proc/$(ps aux | grep "nginx: master process" | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
-cat /proc/$(ps aux | grep "netdata" | head -n1 | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
+cat /proc/$(ps aux | grep "netdata.pid" | head -n1 | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
 ```
 
 View of the files:
--- a/docs/high-performance-netdata.md
+++ b/docs/high-performance-netdata.md
@@ -126,7 +126,7 @@ You can check limits with following commands:
 
 ```sh
 cat /proc/$(ps aux | grep "nginx: master process" | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
-cat /proc/$(ps aux | grep "\/netdata " | head -n1 | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
+cat /proc/$(ps aux | grep "\/[n]etdata " | head -n1 | grep -v grep | awk '{print $2}')/limits | grep "Max open files"
 ```
 
 View of the files:


thiagoftsm:
Hi @n0coast ,

I tested your suggestion now, and I verified that it won't work depending of the netdata.conf variables as you can see in the image 

err

but, I agree with you that we need probably to be more specific.

Best regards!

ilyam8:
@n0coast 	

> when grepping for openfiles the netdata example is not exact enough and will return results from too many processes

I think `head -n1` in the pipe to solve the issue.

 

Do you have different output from

> cat /proc/$(ps aux | grep "netdata" | head -n1 | grep -v grep | awk '{print $2}')/limits | grep "Max open files"

and 

> cat /proc/$(ps aux | grep "netdata.pid" | head -n1 | grep -v grep | awk '{print $2}')/limits | grep "Max open files"

?

n0coast:
"| head -n1" won't help since you can't guarantee other processes with the string `netdata` in them weren't started before the actual netdata binary, a better option would be to grep for "netdata.*-D", which I believe is the default option and should be present on most systems.

Here is an example of my netdata master output for "ps aux | grep netdata
```
$ ps aux | grep netdata
root      3453  0.0  0.0 105452  4328 ?        Ss   Sep06   0:00 /sbin/dhclient -q -cf /etc/dhcp/dhclient-eth1.conf -lf /var/lib/dhclient/dhclient--eth1.lease -pf /var/run/dhclient-eth1.pid -H netdata-master-ip-n-n-n-n eth1
root      3579  0.0  0.0 105452  3988 ?        Ss   Sep06   0:01 /sbin/dhclient -6 -nw -lf /var/lib/dhclient/dhclient6--eth1.lease -pf /var/run/dhclient6-eth1.pid eth1 -H netdata-master-ip-n-n-n-n
netdata   4516  212 45.1 33390636 29512316 ?   Ssl  Sep06 30217:50 /usr/sbin/netdata -P /var/run/netdata/netdata.pid -D -W set global process scheduling policy keep -W set global OOM score keep
netdata   4563  0.0  0.0 140412 19448 ?        Sl   Sep06   3:58 /usr/bin/python /usr/libexec/netdata/plugins.d/python.d.plugin 1
root     10366  0.8  0.0  22840  4768 ?        S    13:00   0:42 /usr/libexec/netdata/plugins.d/apps.plugin 1
ec2-user 14381  0.0  0.0 119468   932 pts/0    S+   14:19   0:00 grep --color=auto netdata
```

`ps aux | grep -E netdata.*-D|head -n1` will select (on my systems) the correct process everytime.

ilyam8:
i think 

> grep "\/netdata "

will also work

n0coast:
Yes, that appears to work as well, though it feels a bit less precise as other processes (like netdata plugins) also match, with `head -n1` it shouldn't be a problem.

Whichever way you feel is best.

ilyam8:
it matches only main process on my install

```
test@debian10:~$ ps aux | grep "\/[n]etdata "
netdata  20145  0.3  3.7 186520 38164 ?        Ssl  07:09   0:59 /opt/netdata/usr/sbin/netdata -P /opt/netdata/var/run/netdata/netdata.pid -D -W set global process scheduling policy keep -W set global OOM score keep
```

vs

```
test@debian10:~$ ps aux | grep netdata
netdata  20145  0.3  3.7 186520 38164 ?        Ssl  07:09   0:59 /opt/netdata/usr/sbin/netdata -P /opt/netdata/var/run/netdata/netdata.pid -D -W set global process scheduling policy keep -W set global OOM score keep
netdata  20180  0.0  2.0 124292 20380 ?        Sl   07:09   0:03 /opt/netdata/usr/libexec/netdata/plugins.d/go.d.plugin 1
netdata  20184  0.0  2.5  39296 26160 ?        Sl   07:09   0:05 /usr/bin/python /opt/netdata/usr/libexec/netdata/plugins.d/python.d.plugin 1
netdata  21585  0.2  0.4  14144  4852 ?        S    11:09   0:03 /opt/netdata/usr/libexec/netdata/plugins.d/apps.plugin 1
netdata  21586  0.0  0.3   3868  3132 ?        S    11:09   0:00 bash /opt/netdata/usr/libexec/netdata/plugins.d/tc-qos-helper.sh 1
test     21707  0.0  0.0   6076   884 pts/0    S+   11:31   0:00 grep netdata
```

n0coast:
You are correct - I missed the " " at the end when testing.

ilyam8:
@n0coast 	please update the PR

ilyam8:
@n0coast 	

it is minor thing but still

> "\/[n]etdata "

is better because it doesnt catch `grep ...`

PR #6796, opened 2019-09-05
Allow hostnames in Access Control Lists

### LINKED ISSUE #6438: Securing Netdata via Netdata.conf
##### Question summary

I've been unsuccessful in trying to lock down the rather excellent Netdata running on a cloud VPS from a US based host via the Netdata.conf allow options. I'd greatly appreciate some assistance or perhaps improved documentation.

I want it to remain accessible on my VPS localhost if necessary but need to lock down web access to myself. The Google sign in doesn't seem to provide this since Netdata is immediately visible via the ipaddress:19999 url.

If I add my IP address to the allow connections from line in Netdata.conf, it lets me view Netdata, but my home IP is a dynamic one provided by my UK fibre broadband provider and changes occasionally or I reboot or update my router.

My router auto updates a Netgear Dynamic DNS hostname with my current IP address, so I tried using this domain in the allow connections field, but it doesn't work even though the hostname resolves to my IP address correctly. 

This works okay:

`allow connections from = localhost 86.xxx.xxx.999`

This doesn't work at all, access denied!

`allow connections from = localhost example.mynetgear.com`

Thank you.

##### OS / Environment

`root@vpsxxxxx [/etc/netdata]# pt-summary
Percona Toolkit System Summary Report
    Date | 2019-07-11 20:18:26 UTC (local TZ: BST +0100)
    Hostname | vpsxxxx.inmotionhosting.com
      Uptime | 17:49,  1 user,  load average: 0.10, 0.13, 0.18
    Platform | Linux
     Release | CentOS Linux release 7.6.1810 (Core) 
      Kernel | 3.10.0-042stab138.1
Architecture | CPU = 64-bit, OS = 64-bit
   Threading | NPTL 2.17
     SELinux | Disabled
 Virtualized | OpenVZ/Virtuozzo
Processor
  Processors | physical = 2, cores = 24, virtual = 48, hyperthreading = yes
      Speeds | 48x2299.744
      Models | 48xIntel(R) Xeon(R) CPU E5-2670 v3 @ 2.30GHz
      Caches | 48x30720 KB
Memory 
       Total | 3.0G
        Free | 683.1M
        Used | physical = 1.9G, swap allocated = 3.0G, swap used = 232.0k, virtual = 1.9G
      Shared | 137.4M
     Buffers | 473.3M
      Caches | 874.8M
       Dirty | 280 kB
     UsedRSS | 1.6G
  Swappiness | 1
 DirtyPolicy | 20, 10
 DirtyStatus | 0, 0
`

##### Component Name

netdata.conf

##### Expected results

Netdata should hopefully allow connections from a correctly resolving domain as it does with an IP address.

amoss:
##### Summary
The Access Control List (ACL) configuration parameters can now use hostnames with simple patterns. Incoming connections are resolved using reverse DNS to obtain the hostname. Where a hostname is resolved, forward DNS resolution is performed to check the IP address is really associated with the hostname. If the checks pass then the patterns supplied are checked. Any patterns supplied for numeric ip addresses are also checked.

##### Component Name
daemon

##### Additional Information
Fixes #6438

### DIFF
--- a/libnetdata/socket/socket.c
+++ b/libnetdata/socket/socket.c
@@ -1005,13 +1005,15 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
 
     int nfd = accept4(fd, (struct sockaddr *)&sadr, &addrlen, flags);
     if (likely(nfd >= 0)) {
-        if (getnameinfo((struct sockaddr *)&sadr, addrlen, client_ip, (socklen_t)ipsize, client_port, (socklen_t)portsize, NI_NUMERICHOST | NI_NUMERICSERV) != 0) {
+        if (getnameinfo((struct sockaddr *)&sadr, addrlen, client_ip, (socklen_t)ipsize, 
+                        client_port, (socklen_t)portsize, NI_NUMERICHOST | NI_NUMERICSERV) != 0) {
             error("LISTENER: cannot getnameinfo() on received client connection.");
             strncpyz(client_ip, "UNKNOWN", ipsize - 1);
             strncpyz(client_port, "UNKNOWN", portsize - 1);
         }
-        char client_host[NI_MAXHOST+1];
-        if(getnameinfo((struct sockaddr *)&sadr, addrlen, client_host, (socklen_t)ipsize, NULL, 0, NI_NAMEREQD | NI_NUMERICSERV) != 0)
+        char client_host[NI_MAXHOST];
+        if(getnameinfo((struct sockaddr *)&sadr, addrlen, client_host, (socklen_t)ipsize, 
+                       NULL, 0, NI_NAMEREQD | NI_NUMERICSERV) != 0)
             client_host[0] = 0;
         else {
             struct addrinfo *addr_infos = NULL;
@@ -1023,17 +1025,20 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
                 struct addrinfo *scan = addr_infos;
                 int   validated = 0;
                 while(scan) {
-                    char address[INET6_ADDRSTRLEN+1];
+                    char address[INET6_ADDRSTRLEN];
                     address[0] = 0;
                     switch(scan->ai_addr->sa_family) {
                         case AF_INET:
-                            inet_ntop(AF_INET, &((struct sockaddr_in*)(scan->ai_addr))->sin_addr, address, INET6_ADDRSTRLEN+1);
+                            inet_ntop(AF_INET, &((struct sockaddr_in*)(scan->ai_addr))->sin_addr, 
+                                      address, INET6_ADDRSTRLEN);
                             break;
                         case AF_INET6:
-                            inet_ntop(AF_INET6, &((struct sockaddr_in6*)(scan->ai_addr))->sin6_addr, address, INET6_ADDRSTRLEN+1);
+                            inet_ntop(AF_INET6, &((struct sockaddr_in6*)(scan->ai_addr))->sin6_addr, 
+                                      address, INET6_ADDRSTRLEN);
                             break;
                     }
-                    debug(D_LISTENER, "Incoming ip %s reverse-resolved onto %s, validating against forward-resolution %s", client_ip, client_host, address);
+                    debug(D_LISTENER, "Incoming ip %s rev-resolved onto %s, validating against forward-resolution %s",
+                          client_ip, client_host, address);
                     if(!strcmp(client_ip, address)) {
                         validated = 1;
                         break;
@@ -1041,7 +1046,7 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
                     scan = scan->ai_next;
                 }
                 if(!validated) {
-                    error("LISTENER: Cannot validate '%s' as address of '%s', not listed in DNS", client_ip, client_host);
+                    error("LISTENER: Cannot validate '%s' as ip of '%s', not listed in DNS", client_ip, client_host);
                     client_host[0] = 0;
                 }
             }
@@ -1090,9 +1095,11 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
             }
 
             // Allow patterns to match against either the resolved hostname or the numeric ip address.
-            if(unlikely(!simple_pattern_matches(access_list, client_ip) && (client_host[0]==0 || !simple_pattern_matches(access_list, client_host)))) {
+            if(unlikely(!simple_pattern_matches(access_list, client_ip) && 
+                        (client_host[0]==0 || !simple_pattern_matches(access_list, client_host)))) {
                 errno = 0;
-                debug(D_LISTENER, "Permission denied for client '%s' (%s), port '%s'", client_ip, client_host, client_port);
+                debug(D_LISTENER, "Permission denied for client '%s' (%s), port '%s'", 
+                      client_ip, client_host, client_port);
                 error("DENIED ACCESS to client '%s'", client_ip);
                 close(nfd);
                 nfd = -1;
--- a/libnetdata/socket/socket.c
+++ b/libnetdata/socket/socket.c
@@ -1005,13 +1005,48 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
 
     int nfd = accept4(fd, (struct sockaddr *)&sadr, &addrlen, flags);
     if (likely(nfd >= 0)) {
-        char client_host[NI_MAXHOST+1];
-        if(getnameinfo((struct sockaddr *)&sadr, addrlen, client_host, (socklen_t)ipsize, NULL, 0, NI_NAMEREQD|NI_NUMERICSERV) != 0)
-            client_host[0] = 0;
-        if (getnameinfo((struct sockaddr *)&sadr, addrlen, client_ip, (socklen_t)ipsize, client_port, (socklen_t)portsize, NI_NUMERICHOST|NI_NUMERICSERV) != 0) {
+        if (getnameinfo((struct sockaddr *)&sadr, addrlen, client_ip, (socklen_t)ipsize, client_port, (socklen_t)portsize, NI_NUMERICHOST | NI_NUMERICSERV) != 0) {
             error("LISTENER: cannot getnameinfo() on received client connection.");
             strncpyz(client_ip, "UNKNOWN", ipsize - 1);
             strncpyz(client_port, "UNKNOWN", portsize - 1);
+        }
+        char client_host[NI_MAXHOST+1];
+        if(getnameinfo((struct sockaddr *)&sadr, addrlen, client_host, (socklen_t)ipsize, NULL, 0, NI_NAMEREQD | NI_NUMERICSERV) != 0)
+            client_host[0] = 0;
+        else {
+            struct addrinfo *addr_infos = NULL;
+            if(getaddrinfo(client_host, NULL, NULL, &addr_infos)!=0) {
+                error("LISTENER: cannot validate hostname '%s' by resolving it",client_host);
+                client_host[0] = 0;
+            }
+            else {
+                struct addrinfo *scan = addr_infos;
+                int   validated = 0;
+                while(scan) {
+                    char address[INET6_ADDRSTRLEN+1];
+                    address[0] = 0;
+                    switch(scan->ai_addr->sa_family) {
+                        case AF_INET:
+                            inet_ntop(AF_INET, &((struct sockaddr_in*)(scan->ai_addr))->sin_addr, address, INET6_ADDRSTRLEN+1);
+                            break;
+                        case AF_INET6:
+                            inet_ntop(AF_INET6, &((struct sockaddr_in6*)(scan->ai_addr))->sin6_addr, address, INET6_ADDRSTRLEN+1);
+                            break;
+                    }
+                    debug(D_LISTENER, "Incoming ip %s reverse-resolved onto %s, validating against forward-resolution %s", client_ip, client_host, address);
+                    if(!strcmp(client_ip, address)) {
+                        validated = 1;
+                        break;
+                    }
+                    scan = scan->ai_next;
+                }
+                if(!validated) {
+                    error("LISTENER: Cannot validate '%s' as address of '%s', not listed in DNS", client_ip, client_host);
+                    client_host[0] = 0;
+                }
+            }
+            if(addr_infos!=NULL)
+                freeaddrinfo(addr_infos);
         }
 
 #ifdef __FreeBSD__
@@ -1057,7 +1092,7 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
             // Allow patterns to match against either the resolved hostname or the numeric ip address.
             if(unlikely(!simple_pattern_matches(access_list, client_ip) && (client_host[0]==0 || !simple_pattern_matches(access_list, client_host)))) {
                 errno = 0;
-                debug(D_LISTENER, "Permission denied for client '%s', port '%s'", client_ip, client_port);
+                debug(D_LISTENER, "Permission denied for client '%s' (%s), port '%s'", client_ip, client_host, client_port);
                 error("DENIED ACCESS to client '%s'", client_ip);
                 close(nfd);
                 nfd = -1;
--- a/libnetdata/socket/socket.c
+++ b/libnetdata/socket/socket.c
@@ -1018,20 +1018,20 @@ extern int connection_allowed(int fd, char *client_ip, char *client_host, size_t
         struct sockaddr_storage sadr;
         socklen_t addrlen = sizeof(sadr);
         int err = getpeername(fd, (struct sockaddr*)&sadr, &addrlen);
-        if (err!=0 ||
+        if (err != 0 ||
             (err = getnameinfo((struct sockaddr *)&sadr, addrlen, client_host, (socklen_t)hostsize,
                               NULL, 0, NI_NAMEREQD)) != 0) {
             error("Incoming connection on '%s' does not match a numeric pattern, "
                   "and host could not be resolved (err=%s)", client_ip, gai_strerror(err));
-            if (hostsize>=8)
+            if (hostsize >= 8)
                 strcpy(client_host,"UNKNOWN");
             return 0;
         }
         struct addrinfo *addr_infos = NULL;
-        if (getaddrinfo(client_host, NULL, NULL, &addr_infos)!=0) {
+        if (getaddrinfo(client_host, NULL, NULL, &addr_infos) !=0 ) {
             error("LISTENER: cannot validate hostname '%s' from '%s' by resolving it",
                   client_host, client_ip);
-            if (hostsize>=8)
+            if (hostsize >= 8)
                 strcpy(client_host,"UNKNOWN");
             return 0;
         }
@@ -1058,7 +1058,7 @@ extern int connection_allowed(int fd, char *client_ip, char *client_host, size_t
         }
         if (!validated) {
             error("LISTENER: Cannot validate '%s' as ip of '%s', not listed in DNS", client_ip, client_host);
-            if (hostsize>=8)
+            if (hostsize >= 8)
                 strcpy(client_host,"UNKNOWN");
         }
         if (addr_infos!=NULL)
--- a/libnetdata/socket/socket.c
+++ b/libnetdata/socket/socket.c
@@ -1114,8 +1114,7 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
         }
         if(!connection_allowed(&sadr, addrlen, client_ip, ipsize, access_list)) {
             errno = 0;
-            debug(D_LISTENER, "Permission denied for client '%s', port '%s'", client_ip, client_port);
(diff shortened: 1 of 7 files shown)


libnetdata/socket/socket.c
```diff
@@ -1010,6 +1010,44 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
             strncpyz(client_ip, "UNKNOWN", ipsize - 1);
             strncpyz(client_port, "UNKNOWN", portsize - 1);
         }
+        char client_host[NI_MAXHOST+1];
+        if(getnameinfo((struct sockaddr *)&sadr, addrlen, client_host, (socklen_t)ipsize, NULL, 0, NI_NAMEREQD | NI_NUMERICSERV) != 0)
```
cosmix: I don't see why you're calling `getnameinfo()` again, given that we're calling it a few lines up. Can you refactor the code so that we only have one call to it, or is there another reason for which we need two calls?

libnetdata/socket/socket.c
```diff
@@ -1010,6 +1010,44 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
...
+            client_host[0] = 0;
+        else {
```
cosmix: We're in the process of establishing a code style convention for C code. Please consult CONTRIBUTING.md for our style guide.

libnetdata/socket/socket.c
```diff
@@ -1010,6 +1010,44 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
...
+            struct addrinfo *addr_infos = NULL;
+            if(getaddrinfo(client_host, NULL, NULL, &addr_infos)!=0) {
+                error("LISTENER: cannot validate hostname '%s' by resolving it",client_host);
+                client_host[0] = 0;
+            }
+            else {
+                struct addrinfo *scan = addr_infos;
+                int   validated = 0;
+                while(scan) {
+                    char address[INET6_ADDRSTRLEN+1];
```
cosmix: I don't think you need the `+1` here.

libnetdata/socket/socket.c
```diff
@@ -1051,9 +1089,10 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
                 client_ip[ipsize - 1] = '\0';
             }
 
-            if(unlikely(!simple_pattern_matches(access_list, client_ip))) {
+            // Allow patterns to match against either the resolved hostname or the numeric ip address.
+            if(unlikely(!simple_pattern_matches(access_list, client_ip) && (client_host[0]==0 || !simple_pattern_matches(access_list, client_host)))) {
```
cosmix: For a publicly accessible (port-scannable) host, I'm not sure the `unlikely()` macro is of much value here. I guess it really depends on the host etc.

libnetdata/socket/socket.c
amoss: If we call it once then we don't know if the response was an ip-address or a hostname. In the case that it is a hostname we need the ip-address as well: it propagates outside of the function and it means that numeric patterns in the config will still work.

cosmix: The way it's structured now, that's true. Hence my question, if you could think of a way to refactor the code to achieve the proper flow with one call. :)

libnetdata/socket/socket.c
amoss: It seems to match the style guide (four space indent, not using braces around a single line body). Could you tell me how it is in the wrong style?

libnetdata/socket/socket.c
```diff
@@ -1010,6 +1010,44 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
             strncpyz(client_ip, "UNKNOWN", ipsize - 1);
             strncpyz(client_port, "UNKNOWN", portsize - 1);
         }
+        char client_host[NI_MAXHOST+1];
```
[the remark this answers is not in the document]
amoss: The example in the manpage seems to indicate that, but I was following the previous definition of client_ip on line 1398. Would you like me to change just client_host or both client_host and client_ip?

libnetdata/socket/socket.c
mfundul: If the flow is to get both an IP address and a hostname then it is not possible with a single call to `getnameinfo()`.

libnetdata/socket/socket.c
amoss: Correct, it includes the null-terminator. I'll remove the +1s.

libnetdata/socket/socket.c
amoss: The case that I'm thinking about is when somebody puts something like: "allow connections from = 1.2.3.4 host.network" because they want to allow based on either the numeric address or the resolved DNS. That's what I would do as the reverse-dns might require a PTR record in someone else's server (e.g. an ISP) and the numeric entry acts as fall-back. If I refactor it to use a single call then it won't be possible to use a fallback.

libnetdata/socket/socket.c
cosmix: Well, for starters the 120 char line limit is broken in several places (which is something you could argue about the rest of the file too, but we only started observing the style guide last week, and we'll have to deal with the monstrous reformatting PR at some point soon)

amoss: Oh yikes that is 160 characters long, I'll fix that...

libnetdata/socket/socket.c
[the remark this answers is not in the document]
amoss: Checked the defines and the max length of an ip-address is only 45 characters vs 1025 so I've changed both declarations.

libnetdata/socket/socket.c
amoss: I agree - saving the branch misprediction cost is insignificant compared to logging the error.

libnetdata/socket/socket.c
amoss: Refactored the code to remove the second call from code-paths where possible, only called twice now when needed to perform both pattern matches. Also deferred the DNS lookup until after the pattern-match.

libnetdata/socket/socket.c
amoss: I think it now matches the kernel style guide + exceptions.

amoss:
@cosmix I think that all of your comments have been addressed, please take a look at the new commits.

When you are happy with the code I can squash the commits into a single.

libnetdata/socket/socket.c
```diff
@@ -1044,21 +1112,13 @@ int accept_socket(int fd, int flags, char *client_ip, size_t ipsize, char *clien
...
-                debug(D_LISTENER, "Permission denied for client '%s', port '%s'", client_ip, client_port);
-                error("DENIED ACCESS to client '%s'", client_ip);
-                close(nfd);
-                nfd = -1;
-                errno = EPERM;
-            }
+        if(!connection_allowed(&sadr, addrlen, client_ip, ipsize, access_list)) {
+            errno = 0;
+            debug(D_LISTENER, "Permission denied for client '%s', port '%s'", client_ip, client_port);
+            error("DENIED ACCESS to client '%s'", client_ip);
```
cosmix: Since you're already logging an error, you can lose the debug log and just add the port to the error message, if you feel it's needed.

libnetdata/socket/socket.c
cosmix: That's great, thank you! Here are my thoughts: `getnameinfo()` _without the flag_ returns hostname or, failing to resolve that, the IP. So you could call it _only once_, get whatever it returns (if anything), check whether it matches an IP or a hostname. If it's a hostname, you can do the forward lookup and get the IP. If it's an IP then you have your IP. One call, some more logic on your end.

Let me know if I'm missing something here @mfundul @amoss.

amoss: This would break the existing functionality in the case that a numeric pattern is being used and reverse DNS is setup on the connecting IP. The single call would resolve e.g. 4.3.2.1.in-address.arpa. while the user may have specified 1.2.3.4 as their pattern. This is outside of user control if they do not own the ip block.

libnetdata/socket/socket.c
amoss: Fixed.

libnetdata/socket/socket.c
cosmix: That doesn't seem right. If a reverse DNS is setup on the connecting IP, it will resolve to whatever the PTR is, i.e. an FQDN. Then you'd validate that FQDN by doing a forward lookup and get the IP. Then you'd compare both the IP you got and the FQDN with your list. This, however, as per the `man` page is not wise, as someone could create a malicious PTR to spoof their origin. So we need the two calls (or, at the very least a call to `getaddrinfo()` and one to `getnameinfo()`, as per the man page.)

amoss: Reverse DNS does not always have a PTR to a FQDN. Most ISPs will set it on addresses inside their block. So, for example, my home connection is very similar to ua-85-228-174-160.bbcust.telenor.se and this is what I see when I run "w" on a server that I'm logged into it. If I was using the existing functionality then I would have set:

`allow connections from = 85.228.174.160`

This pattern would not match the hostname returned by the single `getnameinfo()` when NI_NUMERICHOST is not set. We cannot perform an automatic mapping ourselves to determine this because there is no standard for the syntax of these psuedo hostnames (i.e. each ISP will format them slightly differently).

So to support both numeric IP addresses and hostnames (in a secure way to avoid the spoofing of PTRs) we end up needing all three calls -> two to `getaddrinfo() `to handle both numeric IPs and hostnames, and the final call to `getnameinfo()` if we match the hostname against a pattern.

cosmix: But `ua-85-228-174-160.bbcust.telenor.se` *is* an FQDN. And a forward lookup would point to 85.228.174.160. Adding both would match the `allow connections from` with the IP only. The only thing necessitating the multiple calls is the spoofing imho. But, then again, I've been working for many hours straight so I may be missing something here. :)

amoss: True ;) I meant that it is not a unique (or intentional) hostname for my ip address.

If the user adds both into the `allow connections from` config then it would also work with a single `getaddrinfo()` call. But... there will be existing configurations that do not have both (because it was not needed previously) and these would suddenly break.

Also, the machine that I am connecting to netdata from may not have reverse DNS when I write the config, and it would work fine. But a change to a third party's DNS configuration can then break my access - outside of my control. Leaving both calls in the code avoids all of this and ensures that it can work with either numeric IPs or hostnames.

cosmix: Maybe I didn't do such a good job explaining myself, but I don't think any existing configurations (with IP only) would break, as long as the PTR records had matched A records, (as per the RFC.) The only case in which anything would break would be that of an unmatched PTR record (i.e. missing the A record) and/or spoofed PTR records. But this discussion is getting too academic, as we both agree we need the extra calls to address the spoofing anyway! Thanks Andrew :)

amoss:
What did you have in the `allow connections from` field?

cosmix:
> What did you have in the `allow connections from` field?

```allow connections from = localhost <hostname>
allow dashboard from = localhost <hostname>
```

amoss:
Was it a connection from 127.0.0.1 that failed?

cosmix:
> Was it a connection from 127.0.0.1 that failed?

No, it was a connection from the external host IP that fails.

amoss:
I hadn't looked at the `allow dashboard from` settings. I tested it by curling the "host:19999" part from a host with a FQDN. I'll need to change the setup of the test and get back to you.

amoss:
I've setup a new test environment, firstly I tried:
```
[web]
allow connections from = localhost ua-xxx-xxx-xxx-xxx.telenor.se
```

This worked as expected, I can only connect from the host with the FQDN or via a nginx proxy running on the host. If I add the other line that you tried so that I have:

```
[web]
allow connections from = localhost ua-xxx-xxx-xxx-xxx.telenor.se
allow dashboard from = localhost ua-xxx-xxx-xxx-xxx.telenor.se
```

Then I get an error message that says "You are not allowed to access this resource.".

Does this match what you see when you test it?

cosmix:
> Then I get an error message that says "You are not allowed to access this resource.".
> 
> Does this match what you see when you test it?

Yes.

amoss:
Well then the fix is simple, remove the extra dashboard line :)

In that particular case the check is redundant as the `allow connections from` config prevents the dashboard from being opened on any machine that does not match the pattern. I assume that you are thinking about a more sophisticated answer though, to add some functionality that was not requested in the original issue. What kind of scope do you have in mind?
One useful case would be a config like the following:
```
[web]
    allow connections from = 130.100.*.*
    allow dashboard from = projector*
```
i.e. use the dashboard config to restrict the range specified in the connections config, and the hardest case there is if the original match in the accept_socket() was numeric then handle the DNS checks later in web_client_update_acl_matches(). It is not obvious how to do that with looking around the code some more, I'll need to check how the client_ip members in the POLLJOB and the web_client related to each other and how they are used from the rest of the code. My first estimate would be to add a client_host field in the same style and use "UNKNOWN" or "UNRESOLVED" to handle the different cases and call the dns checks later if necessary. Is this roughly what you have in mind?

cosmix:
Yes, it does work without it. But: That the check is redundant does not mean it should result in failure. On the contrary, it should function as it would had the `dashboard` line been commented out: it should allow whoever matches both lines to connect. What is happening here is bad from a user experience perspective. A valid host cannot connect, despite being listed in both the general `allow connections from` and the more specific `dashboard` lines. Doesn't get more esoteric than this. :)

So that's not a fix.

Now, for the other case you mention — yes, that's what I had in mind and I agree that would be beyond the scope of the original request.

So, while I admit that the "let's restrict the hosts further to what's already specified in the general allow connections from line" case is something else entirely, I don't think having _the same_ hosts on dashboard and the general line should result in denied access to valid hosts, be they specified via IP or hostname (btw the with an IP it still works, even if both lines are active)

amoss:
It will be a while before I check in some code to handle each of the host configuration options that are visible in ewb_client_update_acl_matches() (as I need to go through most of the web server code to see where web_client structures are used and how they relate to the POLLINFO structure in the socket code). In the meantime I found this buffer overflow that you may want to merge separately.

cosmix:
Thanks @amoss that's great. Could you please send the buffer overflow fix in another PR so that we can deal with it separately from this one?

amoss:
New PR#6817 with just the one commit.

cakrit:
Gents, the check isn't redundant. "allow connections from" determines if I should permit a connection at all, even before checking what the request is about. Unless a host appears there, it doesn't matter what you put in the other allow from, it just won't connect.

Adding the hostname also to "allow dashboards from"  is one way to let that hostname access the dashboard. Another is to have * in the "allow dashboards from", so whatever is allowed from "allow connections from" will be allowed to access the dashboard. 

I haven't read the code, but my understanding is that the change was made only on the first level, which means it will need to be fixed.

cosmix:
@cakrit I think that much was clear to Andrew. :)

amoss:
I can confirm that much was clear to me :)

I looked at the other relevant cases and I'm going to tie it together so that each of the configuration options:
```
allow dashboard from
allow registry from
allow badges from
allow management from
allow streaming from
allow netdata.conf from
```
Will work in the same way: each one allows a refinement of the set of allowed connections using a mixture of ip addresses and hostnames.

libnetdata/socket/socket.c
```diff
@@ -995,21 +995,89 @@ int accept4(int sock, struct sockaddr *addr, socklen_t *addrlen, int flags) {
 }
 #endif
 
+/*
+ * ---------------------------------------------------------------------------------------------------------------------
+ * connection_allowed() - if there is an access list then check the connection matches a pattern.
```
thiagoftsm: Hi @amoss ,

Few days ago we agreed in Netdata to use `Doxygen` format to comment the functions, please, update this comment for their format.

amoss: I was asked earlier in the thread to ignore the existing style and follow your style guide. It currently says that comments should be in the linux kernel style (which is how the comment is styled).

thiagoftsm:
Hi @amoss ,

Thank you by the fixes you did here.
We already merged the other PR that you created after few tests, but here in this PR we have the old version of the code, and as you can see in the Travis output (https://travis-ci.com/netdata/netdata/jobs/234755586) it is not compiling due the same warning I reported in the previous PR, my suggestion is for you to do a rebase here to eliminate these errors in the check.

amoss:
@thiagoftsm I've rewritten the history to remove the buffer overflow changes from the commits and rebased it on the latest netdata/master.

amoss:
@thiagoftsm 

I've changed the variable name and checks that it recompiles but I cannot build the code. There seems to be a problem in the netdata repo (not my fork) that the master branch won't rebuild?

`make[2]: don't know how to make coverity-install.sh. Stop`

thiagoftsm:
Hi @amoss ,

I verified and I could compile without any problem your PR, I also saw that everything continue working with IPs, tomorrow I will run a last test this time using names.

> @thiagoftsm
> 
> I've changed the variable name and checks that it recompiles but I cannot build the code. There seems to be a problem in the netdata repo (not my fork) that the master branch won't rebuild?
> 

About the next error, I remember there was a PR removing the file `coverity-install.sh` weeks ago and you can confirm this going to https://github.com/netdata/netdata/, so I do not think there is a problem with our master, but I agree that sometimes Travis gives false positives.

> `make[2]: don't know how to make coverity-install.sh. Stop`

Best regards!

amoss:
Why are you removing localhost from the system and renaming it? It is not clear why you think that this should work.

cakrit:
I believe @thiagoftsm meant that he modified his laptop's `/etc/hosts`. This is precisely the type of requests we WANT to block Thiago. Otherwise, I could put your hostname on my machine and get access!

thiagoftsm:
Hi @cakrit and @amoss ,

I think I was not clear enough, I am bringing a new image that will help to explain my point:

err18

Firstly As you can see I changed my `/etc/hosts` and using the command ping I can obtain equal answers from `localhost` and `hades`, but when I configure my netdata only using aliases that are present in the /etc/hosts I cannot access more information, I tested with `allow connection from`  and we have the same behavior. I did not configure a DNS server to verify that aliaes are a problem when we use a DNS server.

To finish my point about the necessity we accept aliases names inside `/etc/hosts` as valid names, see my Apache configuration:

err20

And a successful access to it: 

err19

Please, let me know case I am not clear yet with this subject. 

Best regards!

amoss:
Hi @thiagoftsm,
I do not think that your test sequence is a valid test of the functionality.
You are adding hades as a local alias for localhost (and then it connects in the case that localhost is in the allowed hosts).
You are then removing localhost as an alias of 127.0.0.1 and trying to connect. I don't think that you can do this.
Also, changing your local hosts file is not updating the DNS name of the host as required in the original issue, the security clarifications of how it should work, or in the long discussion between myself and @cosmix. As @cakrit points out above, what you are trying to test is exactly the spoofing that the patch does not allow.
Cheers,
Andrw

thiagoftsm:
Thank you @amoss and @cakrit , I understand the whole point, and considering that the Apache 2.4.41 accepts Alias and the official documentation of Amazon ( https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html ) also says that they accept in the AWS service, I will give a suggestion, please change our documentation explaining that we won't accept aliases in our configuration.
I am moving in front the tests now after this talk.

cakrit:
@thiagoftsm, I'm not sure what you mean, to tell you the truth, because you're not actually doing anything with DNS A records. I tried myself to play with `/etc/hosts` just on my own laptop to access the UI there, but it's really hard when you try to emulate from localhost to localhost. One thing that can be tested easily without proper DNS zone configuration and should definitely be allowed is the following:

- Install netdata with this branch on Host1.
- Allow connections only from Host2
- Add to the Host1 `/etc/hosts` file a line with `IP2 Host2`. 
- Bring up the Host1 UI (http://Host1:19999/) from Host2 (but be absolutely certain than when Host2 accesses Host1, the IP that the request comes from will be IP2).

thiagoftsm:
Hi @cakrit,

I executed another round of tests, but this time I used another computer to access Netdata as you have suggested, please, see the next image

err21

as you can see, here I could access from the local computer, and I created an enter for another computer `ceres` that has the IP `192.168.0.17` and it is listed in the `allow connections from`. But, when I executed an SSH in the computer and I tried to access the same host, I could not recceive the same kind of information, because Netdata is blocking it. Considering that Netdata is not creating logs for this kind of rejection, I decided to bring a tcpdump to display the data from the other host

err22

As you can see the tcpdump is resolving the IP to the computer name. 
Finally I changed the `allow connections from` to accept connections from `ceres.localdomain` that is the hostname, and I executed the same tests

err24

and as it was hoped, everything worked fine, because I did not use aliases. 
This is the motive I am saying that we need to change our documentation to say that we will accept to use names, if and only if, they are not aliases. In my vision, we cannot only bring the possibility to use names, it is necessary to explicit the names that we accept, and possibly the motive we are only accepting this. 

I have other things to do for Netdata, but I will try to study bind again to do tests using a names server instead /etc/hosts to confirm that CNAMEs will also be blocked.

amoss:
Hi @thiagoftsm,

Both of your requested changes have been made.

Cheers,
Andrew

Issue #13, opened 2015-09-29
/proc/slabinfo


ktsaou:
Support:

`/proc/slabinfo`

mfundul:
Implemented by  #6800.

PR #6800, opened 2019-09-06
Collector slabinfo

Saruspete:
#### Summary

Provide new collector parsing `/proc/slabinfo` to provide details on kernel slab structures.

Asked by issue #13 (very happy for the oldest issue in backlog) 

##### Component Name

collectors/slabinfo.plugin

##### Additional Information

This slabinfo details allows to have clues on actions done on your system.
In the following screenshot, you can clearly see a `find` done on a ext4 filesystem (the number of `ext4_inode_cache` & `dentry` are rising fast), and a few seconds later, an admin issued a `echo 3 > /proc/sys/vm/drop_cached` as their count dropped.

netdata_slabinfo

### DIFF
--- a/collectors/slabinfo.plugin/README.md
+++ b/collectors/slabinfo.plugin/README.md
@@ -7,12 +7,9 @@ Each internal structure (process, file descriptor, inode...) is stored within a 
 
 ## configuring Netdata for slabinfo
 
-There is currently no configuration needed, aside to have the `slabinfo.plugin` plugin being able to read 
-`/proc/slabinfo`. As this procfile is only readable by root, (mode `0400` in the kernel source), we need to grant the 
-slabinfo.plugin this access, either by:
--  with Setuid root: `chown root: slabinfo.plugin && chmod u+s slabinfo.plugin`
--  with capability dac_override: `setcap cap_dac_override+ep slabinfo.plugin`
+There is currently no configuration needed.
 
+As `/proc/slabinfo` is only readable by root, this plugin needs to be setuid root.
 
 ## For what use
 
--- a/collectors/slabinfo.plugin/README.md
+++ b/collectors/slabinfo.plugin/README.md
@@ -7,9 +7,9 @@ Each internal structure (process, file descriptor, inode...) is stored within a 
 
 ## configuring Netdata for slabinfo
 
-There is currently no configuration needed, aside to have the `slabinfo.plugin` plugin being able to read `/proc/slabinfo`.
-
-This can be done either by:
+There is currently no configuration needed, aside to have the `slabinfo.plugin` plugin being able to read 
+`/proc/slabinfo`. As this procfile is only readable by root, (mode `0400` in the kernel source), we need to grant the 
+slabinfo.plugin this access, either by:
 -  with Setuid root: `chown root: slabinfo.plugin && chmod u+s slabinfo.plugin`
 -  with capability dac_override: `setcap cap_dac_override+ep slabinfo.plugin`
 
--- a/collectors/slabinfo.plugin/README.md
+++ b/collectors/slabinfo.plugin/README.md
@@ -1,0 +1,25 @@
+# slabinfo.plugin
+
+SLAB is a cache mechanism used by the Kernel to avoid fragmentation.
+
+Each internal structure (process, file descriptor, inode...) is stored within a SLAB.
+
+
+## configuring Netdata for slabinfo
+
+There is currently no configuration needed, aside to have the `slabinfo.plugin` plugin being able to read `/proc/slabinfo`.
+
+This can be done either by:
+-  with Setuid root: `chown root: slabinfo.plugin && chmod u+s slabinfo.plugin`
+-  with capability dac_override: `setcap cap_dac_override+ep slabinfo.plugin`
+
+
+## For what use
+
+This slabinfo details allows to have clues on actions done on your system.
+In the following screenshot, you can clearly see a `find` done on a ext4 filesystem (the number of `ext4_inode_cache` & `dentry` are rising fast), and a few seconds later, an admin issued a `echo 3 > /proc/sys/vm/drop_cached` as their count dropped.
+
+![netdata_slabinfo](https://user-images.githubusercontent.com/9157986/64433811-7f06e500-d0bf-11e9-8e1e-087497e61033.png)
+
+
+[![analytics](https://www.google-analytics.com/collect?v=1&aip=1&t=pageview&_s=1&ds=github&dr=https%3A%2F%2Fgithub.com%2Fnetdata%2Fnetdata&dl=https%3A%2F%2Fmy-netdata.io%2Fgithub%2Fcollectors%2Fslabinfo.plugin%2FREADME&_u=MAC~&cid=5792dfd7-8dc4-476b-af31-da2fdb9f93d2&tid=UA-64295674-3)](<>)
--- a/collectors/slabinfo.plugin/README.md
+++ b/collectors/slabinfo.plugin/README.md
@@ -9,7 +9,7 @@ Each internal structure (process, file descriptor, inode...) is stored within a 
 
 There is currently no configuration needed.
 
-As `/proc/slabinfo` is only readable by root, this plugin needs to be setuid root.
+As `/proc/slabinfo` is only readable by root, this plugin is setuid root.
 
 ## For what use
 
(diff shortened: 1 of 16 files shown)


vlvkobal:
I think you should set permissions for the plugin in the https://github.com/netdata/netdata/blob/master/netdata-installer.sh#L718-#L768 section and mention in the documentation the fact that the plugin has the special privileges and its cause.

Saruspete:
Added link in doc generator & more detail on why we need it setuid (or setcap).

collectors/slabinfo.plugin/README.md
```diff
@@ -0,0 +1,25 @@
...
+Each internal structure (process, file descriptor, inode...) is stored within a SLAB.
+
+
+## configuring Netdata for slabinfo
+
+There is currently no configuration needed, aside to have the `slabinfo.plugin` plugin being able to read 
+`/proc/slabinfo`. As this procfile is only readable by root, (mode `0400` in the kernel source), we need to grant the 
+slabinfo.plugin this access, either by:
+-  with Setuid root: `chown root: slabinfo.plugin && chmod u+s slabinfo.plugin`
+-  with capability dac_override: `setcap cap_dac_override+ep slabinfo.plugin`
```
vlvkobal: You've already granted access to the `slabinfo` file at installation time. I don't think users need to know about other possible ways of configuring it. You can use a simple sentence, something like
> Keep in mind that libxenstat requires root access, so the plugin is setuid to root.

Saruspete: Is that better ? 

> There is currently no configuration needed.
>
> As `/proc/slabinfo` is only readable by root, this plugin needs to be setuid root.

vlvkobal: LGTM

vlvkobal:
Please add the plugin to
- `collectors/README.md`
- `collectors/apps.plugin/apps_groups.conf`
- `collectors/plugins.d/README.md`
- `docs/Add-more-charts-to-netdata.md`

You should set `SLABINFO_PLUGIN_FILES` in `CMakeLists.txt`. Please make sure that `make` after `cmake` work correctly.

collectors/slabinfo.plugin/slabinfo.c
```diff
@@ -0,0 +1,389 @@
...
+#define SLAB_PAGE_SIZE PAGE_SIZE
+
+// ----------------------------------------------------------------------------
+
+// callback required by fatal()
+void netdata_cleanup_and_exit(int ret) {
+	exit(ret);
+}
+
+void send_statistics( const char *action, const char *action_result, const char *action_data) {
```
vlvkobal: An extra space.

collectors/slabinfo.plugin/slabinfo.c
```diff
@@ -0,0 +1,389 @@
...
+
+
+
+unsigned int do_slab_stats(int update_every) {
+
+	static unsigned int loops = 0;
+	struct slabinfo *sactive = NULL, *s = NULL;
+
+	// Main processing loop
+	while ( running ) {
```
vlvkobal: Extra spaces.

Saruspete:
tested cmake, seems ok now.

collectors/slabinfo.plugin/slabinfo.c
Saruspete: Fixed. Is there any linter that could check this automatically ?

Saruspete:
For the alarms, I don't think so. Or at least not for the behavior I saw.
Slabinfo is an insight on the kernel memory management, and while it provide a useful way of checking what behavior is consuming kernel resources (file-descriptor (filp), processes (task_struct) fs (dentry, {fuse,rpc,fat,mqueue,ext4,sock,shmem,proc}_inode_cache), network (conntrack, iptables...)) I don't see any generic way of knowing if allocations are an issue or totally relevant for the kind of job done by the server. 

I did use the debug a lot, but removed it as I didn't know how to impelement it or what value to choose in log.h.
do you have any hint on this, so I could put them back ?

Saruspete:
Or for the debug, do you mean I didn't use it correctly ?

collectors/slabinfo.plugin/slabinfo.c
vlvkobal: #6677 is on its way to being merged.

vlvkobal:
I meant that usually, we use the `debug` command-line option for an external plugin and simply print debug messages to `stderr`:
```
        else if(strcmp("debug", argv[i]) == 0) {
            debug = 1;
            continue;
        }
```

```
                    " Available command line options:\n"
                    "\n"
                    "  SECONDS                 data collection frequency\n"
                    "                          minimum: %d\n"
                    "\n"
                    "  debug                   enable verbose output\n"
                    "                          default: disabled\n"
```

```
if(debug) fprintf(stderr, "freeipmi.plugin: hostname set to '%s'\n", hostname);
```
It allows debugging the plugin as a separate unit.

vlvkobal:
One last thing - we don't use tabs in C code. Please convert them to spaces (4 spaces per tab).

Saruspete:
Fixed. 
fwiw I'm really not fond of using space (a visual indent) to reflect a logical indent, but unity is more important :)

vlvkobal:
You've introduced compile warnings with the commit for the debug functions:
```
collectors/slabinfo.plugin/slabinfo.c: In function 'get_slabstruct':
collectors/slabinfo.plugin/slabinfo.c:22:21: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'int' [-Wformat=]
   22 |     fprintf(stderr, "slabinfo.plugin DEBUG (%04lu@%-10.10s:%-15.15s)::", __LINE__, __FILE__, __FUNCTION__); \
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
collectors/slabinfo.plugin/slabinfo.c:22:21: note: in definition of macro 'slabdebug'
   22 |     fprintf(stderr, "slabinfo.plugin DEBUG (%04lu@%-10.10s:%-15.15s)::", __LINE__, __FILE__, __FUNCTION__); \
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
collectors/slabinfo.plugin/slabinfo.c:22:49: note: format string is defined here
   22 |     fprintf(stderr, "slabinfo.plugin DEBUG (%04lu@%-10.10s:%-15.15s)::", __LINE__, __FILE__, __FUNCTION__); \
      |                                             ~~~~^
      |                                                 |
      |                                                 long unsigned int
      |                                             %04u
```

Saruspete:
Well, from the error message, I would say cadacy is having a bad time :
```
%lu in format string (no. 1) requires 'unsigned long' but the argument type is 'size_t {aka unsigned long}'.
```

Is that a case of "error success" ?

mfundul:
> 
> 
> Well, from the error message, I would say cadacy is having a bad time :
> 
> ```
> %lu in format string (no. 1) requires 'unsigned long' but the argument type is 'size_t {aka unsigned long}'.
> ```
> 
> He wants `unsigned long`, but the type is `unsigned long` Is that a case of "error success" ?

You're not going to like what I am about to say, but you have to cast it to `(unsigned long)` to make the compiler happy, and we need the project to not have any warnings. The explanation is something like that for the compiler it has been defined as a different type even though it's a `typedef` of identical properties.

Saruspete:
And on travis, the error is not related to the build : 
```
tests/updater_checks.sh: line 80: bats: command not found

The command "docker run -it -v "${PWD}:/netdata:rw" -w /netdata "centos:7" tests/updater_checks.sh" exited with 127.
```

vlvkobal:
> And on travis, the error is not related to the build

It doesn't mean you should ignore the Coverity issues ;)

Saruspete:
Cast done. will retrigger travis to.

Saruspete:
Force pushed to resolve conflict in `netdata.spec.in` due to cups plugin added where slabinfo.plugin was.

What's the usual wait-time for merge ? Does it need to be part of a sprint or milestone ? (asking for PR #6843 too ^^)

vlvkobal:
It looks like you force pushed the old version. Several last commits have disappeared.

Saruspete:
Thankfully you have an eagle eye !
The more mistakes I make, the more I believe all of you are trustworthy to keep netdata clean and stable !

cakrit:
> What's the usual wait-time for merge ? Does it need to be part of a sprint or milestone ? (asking for PR #6843 too ^^)

It only depends on how long it takes us to do a proper review. PRs for C code addition/changes are always trickier, but we generally try to prioritize bug fixes and PR reviews over other work.

Saruspete:
Ok makes sense :-) thanks for the info !

PR #6801, opened 2019-09-06
NPM Packages version update

prhomhyse:
##### Summary
Netdata documentation uses [remark-cli](https://www.npmjs.com/package/remark-cli). 
This PR upgrades the versions of packages in bid to reduce `npm audit` security warnings when you run `npm install`. 
##### Component Name
Docs
##### Additional Information
`npm audit` currently outputs below after the manual report table.
```node
6 vulnerabilities (1 low, 4 high, 1 critical) in 2233 scanned packages
  6 vulnerabilities require manual review. See the full report for details.
```
The following dependencies of the devDependencies are reported 

`minimatch` a dependency of `gaze`
`growl` a dependency of `jasmine-growl-reporter`
`lodash` a dependency of `gaze`

I am not sure other parts of Netdata that depend on the packages, but it's worth looking at if we can remove or replace the vulnerable packages.


prhomhyse:
@jacekkolasa 

Some files that make up the documentation reside in folders other than /docs that's why it is `.` instead of `./docs/**`. 

>Remark works the same way after installing new packages. I have to specify the files (./docs/** instead of .) but before updating node_modules I had the same experience.

Please clarify me a little bit on the above, remark does not work for you when the script is set to `.`?

jacekkolasa:
@prhomhyse yes, when i do `npm run lint-md` i got :
`TypeError: Cannot read property 'path' of undefined` from `node_modules/unified-engine/lib/finder.js:266:37`

Btw. perhaps we should also include the lockfile, could you commit your working package-lock.json?

Maybe that's because i'm on macOS? I've tried both in node v8 and v10

prhomhyse:
I use Mac OS as well, let me commit my package-lock.json real quick to see if it fixes that error.

prhomhyse:
@jacekkolasa 

My Node version is 12.10.0

I can still test with v8 and v10 if the error persist after the package-lock.json file.

jacekkolasa:
that worked, interesting, thanks! I've ran that on node v10 and saw 8096 warnings :) looks like it's good to keep the lockfile

jacekkolasa:
@cakrit or @VLegakis, we'll need approval once again

prhomhyse:
I tested remark on Node 8 and NPM v3.5.2 with this current commit. 

8097 errors, it works fine.

PR #6812, opened 2019-09-10
netdata/ci: Use specific user upon commit to avoid CLA issues

paulkatsoulakis:
##### Summary
Use the right author upon commits

##### Component Name
netdata/ci

##### Additional Information
None


paulkatsoulakis:
reopened on #6813

PR #6817, opened 2019-09-11
Buffer overflow

amoss:
##### Summary

The host field in the web_client is to store the value of the Host HTTP header,
but it is an arbitrary size and there are no length checks. I could not see an
easy way to exploit it but this checks it will not overflow the buffer.

##### Component Name
daemon / web_server

##### Additional Information
Splitting out from PR#6796.

### DIFF
--- a/web/server/web_client.c
+++ b/web/server/web_client.c
@@ -791,7 +791,7 @@ static inline char *http_header_parse(struct web_client *w, char *s, int parse_u
         w->auth_bearer_token = strdupz(v);
     }
     else if(hash == hash_host && !strcasecmp(s, "Host")){
-        strncpyz(w->host, v, (ve-v<sizeof(w->host)-1?ve-v:sizeof(w->host)-1));
+        strncpyz(w->host, v, ((size_t)(ve - v) < sizeof(w->host)-1 ? (size_t)(ve - v) : sizeof(w->host)-1));
     }
 #ifdef NETDATA_WITH_ZLIB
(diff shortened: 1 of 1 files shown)


amoss:
There are no warnings when I compile it on my system (clang + FreeBSD). This means that I cannot verify if they are removed from another system. I think that you should insert casts into the statement and check that they remove the warnings that you see.

I can't predict the type of the pointer arithmetic on the left (as it depends on the memory model) but the sizeof is guaranteed to be size_t so I would recommend trying:
        strncpyz(w->server_host, v, ((size_t)(ve-v)<sizeof(w->server_host)-1?(size_t)(ve-v):sizeof(w->server_host)-1));

thiagoftsm:
Hi @amoss,

I understood everything that you said about this to be critical, and I agree 100%, but considering that your PR had warnings, it could not be merged. I did not want create another PR with the code that you wrote in the comments, because what you did was correct, and case I created PR  I would fix the problem not you.

I am doing a deep test now with your PR to verify to verify that we are not missing nothing, when I finish, I will approve it.

Thank you very much @amoss by your fixes, we appreciate a lot not only your help, but the help from the whole community. I hope with these last paragraphs we  could clarify the whole situaton and I apologize by any misunderstood case there was one.

mfundul:
> I did not want create another PR with the code that you wrote in the comments, because what you did was correct, and case I created PR I would fix the problem not you.

@thiagoftsm I think he meant you commiting directly in his fork in this PR.

thiagoftsm:
Thanks @mfundul for the explanation.
I already merged the PR.
Thanks again @amoss !

PR #6843, opened 2019-09-13
[collector/proc.plugin] Add /proc/pagetypeinfo parser

### LINKED ISSUE #6802: Monitor /proc/pagetypeinfo for memory fragmentation
##### Feature idea summary
The Linux Kernel splits its memory space in Zones (eg, for x86_64):
- DMA : @ 0 to 16MB, for legacy reasons
- DMA32 : @ 16MB to 4GB, for 32bits hardware
- Normal: @4GB to ..., the standard addressing.

Each of these zones is split in pages of 2^10 (1MB for 4KB pagesize) by the **buddyallocator**.
To limit internal fragmentation, it will try to take the smallest page that satisfy the request.

eg: an application requests 100K. The allocator will search for a page of order 7 (128K). If it can't find one, it will try to search for a higher order (8) and split it in 2 half (buddies). If it still can't, it'll do the same twice with an order 9 page, ... up to the max order.

When a page is released, the allocator will try to merge it with its buddy to form a higher page.

If all pages are low-level pages, it often denotes memory fragmentation. Most of the time, this is due to the kernel cache that uses unmovable pages. You can clean the most consumed (inode & dentries) by issuing a "echo 3 >/proc/sys/vm/drop_caches".

The 

A "clean" (ie non-fragmented) machine will have high order pages (8, 9, 10) : 
```
odin [00:13:12][0][~] cat /proc/pagetypeinfo 
Page block order: 9
Pages per block:  512

Free pages count per migrate type at order       0      1      2      3      4      5      6      7      8      9     10 
Node    0, zone      DMA, type    Unmovable      0      0      0      0      2      1      1      0      1      0      0 
Node    0, zone      DMA, type      Movable      0      0      0      0      0      0      0      0      0      1      3 
Node    0, zone      DMA, type  Reclaimable      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone      DMA, type   HighAtomic      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone      DMA, type          CMA      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone      DMA, type      Isolate      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type    Unmovable      1      0      0      0      0      0      1      1      1      1      0 
Node    0, zone    DMA32, type      Movable      2      1      2      0      1      3      2      1      2      1    593 
Node    0, zone    DMA32, type  Reclaimable      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type   HighAtomic      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type          CMA      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type      Isolate      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone   Normal, type    Unmovable    870    530    391    157    103     41      9      2      1      0      0 
Node    0, zone   Normal, type      Movable   5886   9235   5728   4072   1561    324    115     41     12      4  13018 
Node    0, zone   Normal, type  Reclaimable      3      4      8     11      2      3      1      1      1      0      0 
Node    0, zone   Normal, type   HighAtomic      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone   Normal, type          CMA      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone   Normal, type      Isolate      0      0      0      0      0      0      0      0      0      0      0 
```

while a more fragmented server will have mostly low-order pages:
```
Page block order: 9
Pages per block:  512

Free pages count per migrate type at order       0      1      2      3      4      5      6      7      8      9     10 
Node    0, zone      DMA, type    Unmovable      0      1      1      0      2      1      1      0      1      0      0 
Node    0, zone      DMA, type  Reclaimable      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone      DMA, type      Movable      0      0      0      0      0      0      0      0      0      1      3 
Node    0, zone      DMA, type      Reserve      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone      DMA, type          CMA      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone      DMA, type      Isolate      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type    Unmovable    159      6      2      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type  Reclaimable      9   8271   6716      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type      Movable    589   8078   3128      9      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type      Reserve      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type          CMA      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone    DMA32, type      Isolate      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone   Normal, type    Unmovable   1373   1465   1173      2      0      0      0      0      0      0      0 
Node    0, zone   Normal, type  Reclaimable     14      5     13      0      0      0      0      0      0      0      0 
Node    0, zone   Normal, type      Movable  16256  80265 156907    529     67      0      0      0      0      0      0 
Node    0, zone   Normal, type      Reserve      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone   Normal, type          CMA      0      0      0      0      0      0      0      0      0      0      0 
Node    0, zone   Normal, type      Isolate      0      0      0      0      0      0      0      0      0      0      0 
```

If you want more details, you can see my [memory management presentation](https://www.slideshare.net/Saruspete/memory-management-112860641)

##### Expected behavior

What I'm foreseeing is :
- 1 graph per Zone (DMA, DMA32, Normal) & Usual type (Movable / Unmovable). While DMA is almost useless, it could still be useful for 32bits (if supported). Other types (Reclaimable, HighAtomic, CMA, Isolate) are rarely used and would add dimensions for nothing.
- each graph will have the 10 orders as dimension, named and multiplied by their size (eg, dim0 = 4KB, dim10=1024KB).
- Another stack graph as the distribution of the total memory of the system, between **used** (without detail on the page size, that would need parsing of /proc/*/smaps), and the different pages types.

Would a new file like `collectors/proc.plugin/proc_pagetypeinfo.c` be fine ?

Saruspete:
### Provides parsing of /proc/pagetypeinfo to provide fragmentation of free memory pages.

The buddyallocator is the system memory allocator: the whole memory space is split
 in **physical pages** (whose size is determined by the CPU Architecture & MMU). These physical pages are grouped into "page groups" of 2^10 (order 10) physical pages. (also numa node & zones).

When the kernel or an application requests some memory, the buddyallocator provides a page that matches closest the request. 
If an application requests 100KB, the buddy allocator will try to find a page of 128KB (order 7: 2^7).
If there is no such order available, it will find a page frm a superior order, and split it in 2 * 128K page. If no super page is available, it will try to reclaim memory from buffer / cache (reclaimable).

In x86, this translates to:  
- each physical page is 4KB,
- minimal size (order 0) of a pagegroup = 4KB
- maximal size (order 10) or a pagegroup = 4MB
- Zone DMA = 24bits: 0 => 16MB
- Zone DMA32 = 32bits: 16MB => 4GB
- Zone Normal = 4GB and up

pagetypeinfo show statistics of free memory available from the buddy allocator.
The sum of all buddyallocator pages must be equal the "Free Memory" from `/proc/meminfo`

##### Component Name

This addition is done in `collectors/proc.plugin`.

##### Additional Information
Closes #6802 

Ideas about configuration items and default values are welcome.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -261,6 +261,7 @@ PROC_PLUGIN_FILES = \
 	collectors/proc.plugin/proc_softirqs.c \
 	collectors/proc.plugin/proc_loadavg.c \
 	collectors/proc.plugin/proc_meminfo.c \
+	collectors/proc.plugin/proc_pagetypeinfo.c \
 	collectors/proc.plugin/proc_net_dev.c \
 	collectors/proc.plugin/proc_net_ip_vs_stats.c \
 	collectors/proc.plugin/proc_net_netstat.c \
--- a/collectors/all.h
+++ b/collectors/all.h
@@ -35,7 +35,6 @@
 #define NETDATA_CHART_PRIO_SYSTEM_PGPGIO               151
 #define NETDATA_CHART_PRIO_SYSTEM_RAM                  200
 #define NETDATA_CHART_PRIO_SYSTEM_SWAP                 201
-#define NETDATA_CHART_PRIO_SYSTEM_MEMFRAG              220
 #define NETDATA_CHART_PRIO_SYSTEM_SWAPIO               250
 #define NETDATA_CHART_PRIO_SYSTEM_NET                  500
 #define NETDATA_CHART_PRIO_SYSTEM_IPV4                 500 // freebsd only
--- a/collectors/all.h
+++ b/collectors/all.h
@@ -35,6 +35,7 @@
 #define NETDATA_CHART_PRIO_SYSTEM_PGPGIO               151
 #define NETDATA_CHART_PRIO_SYSTEM_RAM                  200
 #define NETDATA_CHART_PRIO_SYSTEM_SWAP                 201
+#define NETDATA_CHART_PRIO_SYSTEM_MEMFRAG              220
 #define NETDATA_CHART_PRIO_SYSTEM_SWAPIO               250
 #define NETDATA_CHART_PRIO_SYSTEM_NET                  500
 #define NETDATA_CHART_PRIO_SYSTEM_IPV4                 500 // freebsd only
@@ -90,6 +91,7 @@
 #define NETDATA_CHART_PRIO_MEM_KSM_RATIOS             1302
 #define NETDATA_CHART_PRIO_MEM_NUMA                   1400
 #define NETDATA_CHART_PRIO_MEM_NUMA_NODES             1410
+#define NETDATA_CHART_PRIO_MEM_PAGEFRAG               1450
 #define NETDATA_CHART_PRIO_MEM_HW                     1500
 #define NETDATA_CHART_PRIO_MEM_HW_ECC_CE              1550
 #define NETDATA_CHART_PRIO_MEM_HW_ECC_UE              1560
--- a/collectors/proc.plugin/plugin_proc.c
+++ b/collectors/proc.plugin/plugin_proc.c
@@ -32,7 +32,7 @@ static struct proc_module {
         { .name = "/sys/block/zram", .dim = "zram", .func = do_sys_block_zram },
         { .name = "/sys/devices/system/edac/mc", .dim = "ecc", .func = do_proc_sys_devices_system_edac_mc },
         { .name = "/sys/devices/system/node", .dim = "numa", .func = do_proc_sys_devices_system_node },
-        { .name = "/proc/pagetypeinfo", .dim = "meminfo", .func = do_proc_pagetypeinfo },
+        { .name = "/proc/pagetypeinfo", .dim = "pagetypeinfo", .func = do_proc_pagetypeinfo },
 
         // network metrics
         { .name = "/proc/net/dev", .dim = "netdev", .func = do_proc_net_dev },
--- a/collectors/proc.plugin/plugin_proc.c
+++ b/collectors/proc.plugin/plugin_proc.c
@@ -32,6 +32,7 @@ static struct proc_module {
         { .name = "/sys/block/zram", .dim = "zram", .func = do_sys_block_zram },
         { .name = "/sys/devices/system/edac/mc", .dim = "ecc", .func = do_proc_sys_devices_system_edac_mc },
         { .name = "/sys/devices/system/node", .dim = "numa", .func = do_proc_sys_devices_system_node },
+        { .name = "/proc/pagetypeinfo", .dim = "meminfo", .func = do_proc_pagetypeinfo },
 
         // network metrics
         { .name = "/proc/net/dev", .dim = "netdev", .func = do_proc_net_dev },
--- a/collectors/proc.plugin/proc_pagetypeinfo.c
+++ b/collectors/proc.plugin/proc_pagetypeinfo.c
@@ -292,7 +292,6 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
             systemorders[o].size += pagelines[p].free_pages_size[o];
         }
 
-        assert(p < pagelines_cnt);
         p++;
     }
 
--- a/collectors/proc.plugin/proc_pagetypeinfo.c
+++ b/collectors/proc.plugin/proc_pagetypeinfo.c
@@ -35,7 +35,9 @@ struct pageline {
 	int node;
 	char *zone;
 	char *type;
+	int line;
 	uint64_t free_pages[MAX_PAGETYPE_ORDER];
+	RRDDIM  *rd[MAX_PAGETYPE_ORDER];
 };
 
 struct nodezone {
@@ -61,13 +63,15 @@ static inline uint64_t pageline_total_size(struct pageline *p) {
 static inline uint64_t pageline_total_count(struct pageline *p) {
 	uint64_t sum = 0, o;
 	for (o=0; o<MAX_PAGETYPE_ORDER; o++)
-		sum += p->free_pages[0];
+		sum += p->free_pages[o];
 	return sum;
 }
 
 // Check if a line of /proc/pagetypeinfo is valid to use
+// Free block lines starts by "Node" && 4th col is "type"
 #define pagetypeinfo_line_valid(ff, l) (strncmp(procfile_lineword(ff, l, 0), "Node", 4) == 0 && strncmp(procfile_lineword(ff, l, 4), "type", 4) == 0)
 
+// Dimension name from the order
 #define dim_name(s, o) (snprintfz(s, 16,"%luKB (%lu)", (1 << o) * PAGE_SIZE / 1024, o))
 
 int do_proc_pagetypeinfo(int update_every, usec_t dt) {
@@ -119,12 +123,29 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 		cnt_zones = 0;
 		pagelines_cnt = 0;
 
+		// Pass 1: how many lines would be valid
+		for (l = 4; l < lines; l++) {
+			if (!pagetypeinfo_line_valid(ff, l))
+				continue;
+
+			pagelines_cnt++;
+		}
+
+		// Init pagelines from scanned lines
+		if (!pagelines) {
+			pagelines = callocz(pagelines_cnt, sizeof(struct pageline));
+			if (!pagelines) {
+				error("PLUGIN: PROC_PAGETYPEINFO: Cannot allocate %lu pagelines of %lu B", pagelines_cnt, sizeof(struct pageline));
+				return 1;
+			}
+		}
+
+		// Pass 2: Scan the file again, with details
+		p = 0;
 		for (l=4; l < lines; l++) {
 
-			// Free block lines starts by "Node" && 4th col is "type"
-			if (!pagetypeinfo_line_valid(ff, l)) {
+			if (!pagetypeinfo_line_valid(ff, l))
 				continue;
-			}
 
 			size_t nodenum = strtoul(procfile_lineword(ff, l, 1), NULL, 10);
 			char *zonename = procfile_lineword(ff, l, 3);
@@ -148,18 +169,17 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 					cnt_pagetypes = l - 4;
 			}
 
-			pagelines_cnt++;
+			// Populate the line
+			pgl = &pagelines[p];
 
-		}
+			pgl->line = l;
+			pgl->node = nodenum;
+			pgl->type = typename;
+			pgl->zone = zonename;
+			for (o = 0; o < MAX_PAGETYPE_ORDER; o++)
+				pgl->free_pages[o] = str2uint64_t(procfile_lineword(ff, l, o+6));
 
-		// Init pagelines
-		if (!pagelines) {
-			pagelines = callocz(pagelines_cnt, sizeof(struct pageline));
-			if (!pagelines) {
-				error("PLUGIN: PROC_PAGETYPEINFO: Cannot allocate %lu B for pageline", pagelines_cnt * sizeof(struct pageline));
-				return 1;
-			}
-
+			p++;
 		}
 
 		// Init the RRD graphs
@@ -172,7 +192,7 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 			, "pagetype"
 			, NULL
 			, "System orders available"
-			, "KB"
+			, "B"
 			, PLUGIN_PROC_NAME
 			, PLUGIN_PROC_MODULE_PAGETYPEINFO_NAME
 			, NETDATA_CHART_PRIO_SYSTEM_MEMFRAG
@@ -194,17 +214,23 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 		// TODO
 
 		// Per-Numa Node & Zone & Type (full detail). Only if sum(line) > 0
-		st_nodezonetype = callocz(cnt_zones, sizeof(RRDSET*));
+		st_nodezonetype = callocz(cnt_zones, sizeof(RRDSET));
 		for (p = 0; p < pagelines_cnt; p++) {
 			pgl = &pagelines[p];
 
 			// Skip empty pagelines
-			if (!pgl || pageline_total_count(pgl) == 0)
+			// TODO: add configuration for override
+			if (!pgl || pageline_total_count(pgl) == 0) {
+				error("Skipping zero total for page %lu", p);
 				continue;
+			}
 
+			char id[4+2+1+MAX_ZONETYPE_NAME+1+MAX_PAGETYPE_NAME+1];
+			snprintfz(id, 4+2+1+MAX_ZONETYPE_NAME+1+MAX_PAGETYPE_NAME, "node%d_%s_%s", pgl->node, pgl->zone, pgl->type);
 
-			char id[MAX_ZONETYPE_NAME+1];
-			snprintfz(id, MAX_ZONETYPE_NAME, "node%d_%s_%s", pgl->node, pgl->zone, pgl->type);
+			char name[MAX_ZONETYPE_NAME + MAX_PAGETYPE_NAME +1];
+			snprintfz(name, MAX_ZONETYPE_NAME + MAX_PAGETYPE_NAME, "Node %d %s %s",
+				pgl->node, pgl->zone, pgl->type);
 
 			st_nodezonetype[p] = rrdset_create_localhost(
 					"mem"
@@ -212,23 +238,22 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 					, NULL
 					, "pagetype"
 					, NULL
-					, "Page Size Distribution"
-					, "pages size"
+					, name
+					, "B"
 					, PLUGIN_PROC_NAME
 					, PLUGIN_PROC_MODULE_PAGETYPEINFO_NAME
-					, NETDATA_CHART_PRIO_MEM_PAGEFRAG
+					, NETDATA_CHART_PRIO_MEM_PAGEFRAG + p
 					, update_every
 					, RRDSET_TYPE_STACKED
 			);
-
 			for (o = 0; o < MAX_PAGETYPE_ORDER; o++) {
 				char id[3+1];
 				snprintfz(id, 3, "%lu", o);
 				char name[20+1];
 				dim_name(name, o);
 
-				RRDDIM *rd = rrddim_add(st_nodezonetype[p], id, name, PAGE_SIZE, 1, RRD_ALGORITHM_ABSOLUTE);
-				rrddim_set_by_pointer(st_order, rd, pgl->free_pages[o]);
+				pgl->rd[o] = rrddim_add(st_nodezonetype[p], id, name, PAGE_SIZE, 1, RRD_ALGORITHM_ABSOLUTE);
+				rrddim_set_by_pointer(st_order, pgl->rd[o], pgl->free_pages[o]);
 			}
 		}
 	}
@@ -249,7 +274,6 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 			continue;
 
 		int words = procfile_linewords(ff, l);
-//error("Line has %d words", words);
 
 		if (words < 6+cnt_pageorders) {
 			error("Unable to read line %lu, only %d words found instead of %d", l, words, 6 + cnt_pageorders);
@@ -257,12 +281,10 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 		}
 
 		for (o = 0; o < MAX_PAGETYPE_ORDER; o++) {
-//error("Updating order %lu", o);
 			// Reset counter
 			if (p == 0)
 				systemorders[o].count = 0;
 
-//error("data update for pageline=%lu order=%lu, value=%lu", p, o, pagelines[p].free_pages[o]);
 			// Update orders of the current line
 			pagelines[p].free_pages[o] = str2uint64_t(procfile_lineword(ff, l, o+6));
 
@@ -286,6 +308,19 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 
 
 	// Per Node-Zone-Type
+	for (p = 0; p < pagelines_cnt; p++) {
+		// Skip empty graphs
+		if (!st_nodezonetype[p]) {
+			error("Skipping line %lu", p);
+			continue;
+		}
+
+		rrdset_next(st_nodezonetype[p]);
+		for (o = 0; o < MAX_PAGETYPE_ORDER; o++)
+			rrddim_set_by_pointer(st_nodezonetype[p], pagelines[p].rd[o], pagelines[p].free_pages[o]);
+
+		rrdset_done(st_nodezonetype[p]);
+	}
 
 	return 0;
 }
--- a/collectors/proc.plugin/proc_pagetypeinfo.c
+++ b/collectors/proc.plugin/proc_pagetypeinfo.c
@@ -181,7 +181,7 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
             pgl->node = nodenum;
             pgl->type = typename;
             pgl->zone = zonename;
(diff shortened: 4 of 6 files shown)


Saruspete:
Configuration flags to be added :
- enable the collector (default = yes)
- enable the details (1 graph per zone & type) (default = auto: show only non-empty lines)
- enable the summary (whole system, all zones summed). (default = yes)

Saruspete:
Could someone retrigger the travis build ? Again, internal travis failure.
Why couldn't travis use smarter features, like zfs snapshots for the builds instead of recreating a new one each time ?

```
E: Package 'fakeroot' has no installation candidate
E: Unable to locate package libipmimonitoring-dev
E: Unable to locate package libmnl-dev
E: Unable to locate package libnetfilter-acct-dev
E: Unable to locate package python-pip
The command "sudo apt-get install -y libcap2-bin zlib1g-dev uuid-dev fakeroot libipmimonitoring-dev libmnl-dev libnetfilter-acct-dev gnupg python-pip" failed and exited with 100 during .
```

cakrit:
I triggered it again, but Travis has been testing our sanity lately. No guarantees it will work, will do one final run if it still fails. We can still merge if it's approved and Travis doesn't come back to its senses.

cakrit:
It worked, Travis is ok with this PR.

Saruspete:
I guess you already have your own tickets to work on, but would someone be able to look on it ? 
i'm pretty sure some edgecase-fixes / obvious enhancements would be needed.

many thanks ! :)

cakrit:
@vlvkobal will review this one too. You're stretching him a bit, but you know he does quality reviews! ;)

Saruspete:
> @vlvkobal will review this one too. You're stretching him a bit, but you know he does quality reviews! ;)

Indeed he does ! It was just to have an idea about the usual times. so no stress and no rush :-)

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+        }
+        rrdset_done(st_order);
+    }
+
+    // Per Node-Zone-Type
+    if (do_detail) {
+        for (p = 0; p < pagelines_cnt; p++) {
+            // Skip empty graphs
+            if (!st_nodezonetype[p]) {
+                error("Skipping line %lu", p);
```
vlvkobal: A lot of redundant messages in `error.log`.

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+                snprintfz(dimid, 3, "%lu", o);
+                char dimname[20+1];
+                dim_name(dimname, o);
+
+                pgl->rd[o] = rrddim_add(st_nodezonetype[p], dimid, dimname, PAGE_SIZE, 1, RRD_ALGORITHM_ABSOLUTE);
+            }
+        }
+    }
+
+    if(unlikely(!cnt_nodes)) {
```
vlvkobal: Check it before chart initialization and swap `cnt_nodes` with `pagelines_cnt`.

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+
+        // Per-Numa Node & Zone & Type (full detail). Only if sum(line) > 0
+        st_nodezonetype = callocz(cnt_zones, sizeof(RRDSET));
+        for (p = 0; p < pagelines_cnt; p++) {
+            pgl = &pagelines[p];
+
+            // Skip invalid, refused or empty pagelines if not explicitely requested
+            if (!pgl
+                || do_detail == CONFIG_BOOLEAN_NO
+                || (do_detail == CONFIG_BOOLEAN_AUTO && pageline_total_count(pgl) == 0))
```
vlvkobal: Can't pages be moved to other types at runtime?

collectors/all.h
```diff
@@ -35,6 +35,7 @@
 #define NETDATA_CHART_PRIO_SYSTEM_PGPGIO               151
 #define NETDATA_CHART_PRIO_SYSTEM_RAM                  200
 #define NETDATA_CHART_PRIO_SYSTEM_SWAP                 201
+#define NETDATA_CHART_PRIO_SYSTEM_MEMFRAG              220
```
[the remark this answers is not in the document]
Saruspete: Removed and used MEM_PAGEFRAG + offset

Makefile.am
```diff
@@ -257,6 +257,7 @@ PROC_PLUGIN_FILES = \
 	collectors/proc.plugin/proc_softirqs.c \
 	collectors/proc.plugin/proc_loadavg.c \
 	collectors/proc.plugin/proc_meminfo.c \
+	collectors/proc.plugin/proc_pagetypeinfo.c \
```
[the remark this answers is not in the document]
Saruspete: I always forgot about cmake ! (and many other things)

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include "plugin_proc.h"
+
+// For PAGE_SIZE
```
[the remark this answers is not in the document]
Saruspete: Used the same solution.

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+    static RRDSET **st_nodezonetype = NULL;
+
+    // Local temp variables
+    size_t l, o, p;
+    struct pageline *pgl = NULL;
+
+    // --------------------------------------------------------------------
+    // Startup: Open /proc/pagetypeinfo
+    if(unlikely(!ff)) {
+        ff = procfile_open(PLUGIN_PROC_MODULE_PAGETYPEINFO_NAME, " \t,", PROCFILE_FLAG_DEFAULT);
```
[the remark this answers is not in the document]
Saruspete: Fixed

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+            for (o = 0; o < MAX_PAGETYPE_ORDER; o++)
+                pgl->free_pages_size[o] = str2uint64_t(procfile_lineword(ff, l, o+6)) * 1 << o;
+
+            p++;
+        }
+
+        // Init the RRD graphs
+
+        // Per-Order: sum of all node, zone, type Grouped by order
+        if (do_global == CONFIG_BOOLEAN_YES) {
```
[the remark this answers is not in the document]
Saruspete: Better indeed

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+                char dimname[20+1];
+                dim_name(dimname, o);
+
+                pgl->rd[o] = rrddim_add(st_nodezonetype[p], dimid, dimname, PAGE_SIZE, 1, RRD_ALGORITHM_ABSOLUTE);
+            }
+        }
+    }
+
+    if(unlikely(!cnt_nodes)) {
+        error("PLUGIN: PROC_PAGETPEINFO: Cannot find the number of CPUs in %s", PLUGIN_PROC_MODULE_PAGETYPEINFO_NAME);
```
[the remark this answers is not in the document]
Saruspete: CPU was referring as physical CPU (that is Numa node). that was misleading indeed.

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+            );
+            for (o = 0; o < MAX_PAGETYPE_ORDER; o++) {
+                char id[3+1];
+                snprintfz(id, 3, "%lu", o);
+
+                char name[20+1];
+                dim_name(name, o);
+
+                systemorders[o].rd = rrddim_add(st_order, id, name, PAGE_SIZE, 1, RRD_ALGORITHM_ABSOLUTE);
+                rrddim_set_by_pointer(st_order, systemorders[o].rd, systemorders[o].size);
```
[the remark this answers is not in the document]
Saruspete: Thought I needed to somehow init it when declared. The value attribution is made later indeed.

collectors/proc.plugin/proc_pagetypeinfo.c
Saruspete: indeed, not needed here.

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+            }
+        }
+
+
+        // Per-Numa Node & Zone & Type (full detail). Only if sum(line) > 0
+        st_nodezonetype = callocz(cnt_zones, sizeof(RRDSET));
```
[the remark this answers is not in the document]
Saruspete: `pagelines_cnt` is the size of `struct pagelines`, cnt_zones is the number of zones (so lines) with non-empty values.
Both the naming and this allocation are misleading, i'll fix this

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,344 @@
...
+        do_detail = config_get_boolean_ondemand(CONFIG_SECTION_PLUGIN_PROC_PAGETYPEINFO, "enable detail per-type", CONFIG_BOOLEAN_AUTO);
+        filter_types = simple_pattern_create(
+                config_get(CONFIG_SECTION_PLUGIN_PROC_PAGETYPEINFO, "hide charts id matching", "")
+                , NULL
+                , SIMPLE_PATTERN_SUFFIX
+        );
+
+
+        // 4th line is the "Free pages count...". Just substract the 8 words.
+        cnt_pageorders = procfile_linewords(ff, 3) - 9;
```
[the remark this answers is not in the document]
Saruspete: The 4th line is always the list of orders (that is, constant 
[mm/vmstat.c :: pagetypeinfo_show](https://github.com/torvalds/linux/blob/master/mm/vmstat.c#L1502-L1505) + [mm/vmstat.c :: pagetypeinfo_showfree](https://github.com/torvalds/linux/blob/master/mm/vmstat.c#L1402-L1404) 
Should always be 11, but can be overridden [include/linux/mmzone.h](https://github.com/torvalds/linux/blob/master/include/linux/mmzone.h#L25-L30)

collectors/proc.plugin/proc_pagetypeinfo.c
Saruspete: They may, but I'm not sure how to make it "dynamic". Most of the time, only 3 types over the 6 are filled, and putting them all would generate 3 not-shown-types * (NUMA-Nodes + 2 DMA/DMA32 charts, each with 10 dimensions, and all at 0.

So if config `enable detail per-type` is forced to `yes`, it'll show them all, as user explicitely wants them

else, the default `auto` will only show non-empty lines at discovery time. i'm no sure i should rescan often, nor we should create ephemeral graph, as I had issues with the cleanup of and processing of ephemeral cgroups

vlvkobal:
```
In file included from collectors/proc.plugin/../../daemon/../libnetdata/libnetdata.h:7,
                 from collectors/proc.plugin/../../daemon/common.h:6,
                 from collectors/proc.plugin/plugin_proc.h:6,
                 from collectors/proc.plugin/proc_pagetypeinfo.c:3:
collectors/proc.plugin/proc_pagetypeinfo.c: In function 'do_proc_pagetypeinfo':
collectors/proc.plugin/proc_pagetypeinfo.c:109:31: warning: comparison of integer expressions of different signedness: 'size_t' {aka 'long unsigned int'} and 'int' [-Wsign-compare]
  109 |     if(unlikely(pagelines_cnt == -1)) {
      |                               ^~
./config.h:359:41: note: in definition of macro 'unlikely'
  359 | #define unlikely(x) __builtin_expect(!!(x), 0)
collectors/proc.plugin/proc_pagetypeinfo.c:58:51: warning: format '%lu' expects argument of type 'long unsigned int', but argument 4 has type 'long int' [-Wformat=]
   58 | #define dim_name(s, o, pagesize) (snprintfz(s, 16,"%luKB (%lu)", (1 << o) * pagesize / 1024, o))
      |                                                   ^~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                      |
      |                                                                                      long int
collectors/proc.plugin/proc_pagetypeinfo.c:203:17: note: in expansion of macro 'dim_name'
  203 |                 dim_name(name, o, pagesize);
      |                 ^~~~~~~~
collectors/proc.plugin/proc_pagetypeinfo.c:58:54: note: format string is defined here
   58 | #define dim_name(s, o, pagesize) (snprintfz(s, 16,"%luKB (%lu)", (1 << o) * pagesize / 1024, o))
      |                                                    ~~^
      |                                                      |
      |                                                      long unsigned int
      |                                                    %lu
collectors/proc.plugin/proc_pagetypeinfo.c:58:51: warning: format '%lu' expects argument of type 'long unsigned int', but argument 4 has type 'long int' [-Wformat=]
   58 | #define dim_name(s, o, pagesize) (snprintfz(s, 16,"%luKB (%lu)", (1 << o) * pagesize / 1024, o))
      |                                                   ^~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                      |
      |                                                                                      long int
collectors/proc.plugin/proc_pagetypeinfo.c:252:17: note: in expansion of macro 'dim_name'
  252 |                 dim_name(dimname, o, pagesize);
      |                 ^~~~~~~~
collectors/proc.plugin/proc_pagetypeinfo.c:58:54: note: format string is defined here
   58 | #define dim_name(s, o, pagesize) (snprintfz(s, 16,"%luKB (%lu)", (1 << o) * pagesize / 1024, o))
      |                                                    ~~^
      |                                                      |
      |                                                      long unsigned int
      |                                                    %lu
```

collectors/proc.plugin/proc_pagetypeinfo.c
vlvkobal: You've changed the variable, but you also shouldn't try to initialize charts if you can't get data from the file. Move this check and the error message to L181.

vlvkobal: Between these lines
```
        // Init the RRD graphs

        // Per-Order: sum of all node, zone, type Grouped by order
```

collectors/proc.plugin/proc_pagetypeinfo.c
vlvkobal: You can initialize charts and dimensions right before setting the dimensions like in
https://github.com/netdata/netdata/blob/master/collectors/proc.plugin/proc_meminfo.c#L222-L257

You can also make them fully dynamic using `rrdset_is_obsolete()` and `rrddim_is_obsolete()` but it doesn't make sense if data changes are not generally expected.

vlvkobal: Please pay attention to `netdata_zero_metrics_enabled == CONFIG_BOOLEAN_YES`. It is a global option to enable `auto` charts permanently.

collectors/proc.plugin/proc_pagetypeinfo.c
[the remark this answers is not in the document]
vlvkobal: What if the user set a wrong file in the configuration and the `cnt_pageorders` becomes negative?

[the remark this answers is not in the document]
Saruspete: It wouldn't have an effect in current code, but it showed me the parsing loops were using the constant `MAX_PAGETYPE_ORDER` insteada of `cnt_pageorders`. 
So I added more strict checks for value, and tried to make it parse a buggy pagetypeinfo file.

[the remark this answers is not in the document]
Saruspete: When unrolling this, seems that if we call `callocz_int` with 0 object, it will `malloc` only 1 `size_t`, then try to return `(void *)&n[1];`, which generates a segfault.

collectors/proc.plugin/proc_pagetypeinfo.c
Saruspete: Added more checks for values and cleanup. If we can't get the data from file (pageline_cnt == 0) we return with error.
If we didn't had any valid line for `pagetypeinfo_line_valid` return too with error.

collectors/proc.plugin/proc_pagetypeinfo.c
Saruspete: I've added the configuration check for `netdata_zero_metrics_enabled`.
I'm still checking for fully dynamic rrd.

collectors/proc.plugin/proc_pagetypeinfo.c
vlvkobal: Hm. I don't understand what's the difference between `p` and `pagelines_cnt`?

And why do you need
```
    if(unlikely(!pagelines_cnt)) {
        error("PLUGIN: PROC_PAGETPEINFO: Cannot find the number of NUMA Nodes in %s", PLUGIN_PROC_MODULE_PAGETYPEINFO_NAME);
        return 1;
    }
```
after charts initialization?

Saruspete: I've made the validity check sooner, against `pagelines_cnt` directly and removed this dead code. It was for when I planned to do per-numa node summaries, but it isn't useful anymore.

collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -0,0 +1,333 @@
...
+        if (!pagelines) {
+            pagelines = callocz(pagelines_cnt, sizeof(struct pageline));
+            if (!pagelines) {
+                error("PLUGIN: PROC_PAGETYPEINFO: Cannot allocate %lu pagelines of %lu B", pagelines_cnt, sizeof(struct pageline));
+                return 1;
+            }
+        }
+
+        // Pass 2: Scan the file again, with details
+        p = 0;
```
vlvkobal: It looks like you don't use it anymore.

Saruspete: I do, for `pgl = &pagelines[p];` as I can't use "l" which is the line (starts at 4 & have gaps due to `pagetypeinfo_line_valid`).

collectors/proc.plugin/proc_pagetypeinfo.c
vlvkobal: Do you want to make charts more dynamic here or in a separate PR?

Saruspete: On a separate PR would be better, depending on user feedback.

collectors/proc.plugin/plugin_proc.c
```diff
@@ -32,6 +32,7 @@ static struct proc_module {
         { .name = "/sys/block/zram", .dim = "zram", .func = do_sys_block_zram },
         { .name = "/sys/devices/system/edac/mc", .dim = "ecc", .func = do_proc_sys_devices_system_edac_mc },
         { .name = "/sys/devices/system/node", .dim = "numa", .func = do_proc_sys_devices_system_node },
+        { .name = "/proc/pagetypeinfo", .dim = "meminfo", .func = do_proc_pagetypeinfo },
```
cosmix: Why is `.dim` set to `meminfo` here?

Saruspete: Very good question... most probably a copy/paste error.

Saruspete:
Netlify failed on an unrelated documentation error
Travis failed on package download on last step.
> Error: Error downloading packages:
  Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f30&arch=x86_64

cakrit:
Please do a rebase, so that netlify can succeed as well.

Saruspete:
Should be better now.

Saruspete:
Should be better now. hoping I won't have to do it again, i'm still not used to it and always end doing forcepush...

Saruspete:
Following the issue on slabinfo generating too much metrics on small systems, here this the number of metrics generated is relative to the number of numa nodes (not CPU). This number of added metrics is to compare to the high number of per-cpu metrics (interrupts, softirq, cpu detail...). it's not that high, but still significant :
- **Standard laptop** 1CPU: (2+1) Zones * 3 migrate type * 10 orders = 90 metrics
- **RPi Like** 1 CPU: (2+1) Zones * 4 migrate type * 10 orders = 120 metrics
- **Standard Server** 2 NUMA Nodes : (2+2) Zones * 3 migrate type * 10 orders = 120 metrics
- **Dual Xeon Skylake/CascadeLake SNC** (4 NUMA Nodes): (2+4) * 3 migrate type * 10 orders = 180 metrics
- **Dual AMD Ryzen Server** (8 NUMA Nodes): (2+8) Zones * 3 migrate type * 10 metrics = 300 metrics

Does it seem ok for you ?

Saruspete:
For the dimensions, its defined by kernel constant `MAX_ORDER`, set to 11 in `include/linux/mmzone.h`. 

There is an override config `[CONFIG_FORCE_MAX_ZONEORDER](https://cateee.net/lkddb/web-lkddb/FORCE_MAX_ZONEORDER.html)` but only available for very exotic arch (xtensa, sparc, nios2, csky, c6x, arc, s390,metag). 
Some arch have conditions that cannot be met (ia64: `true && !true`, mips: `page16 && page32 && page64`) or exceedingly hardware related: (arm: `SOC_AM33xx && (SA1111 || ARCH_EFM32)`

And I never saw a custom kernel config where someone changed this value.

vlvkobal:
Thank you @Saruspete!

ilyam8:
@vlvkobal @Saruspete this collector no longer works after https://github.com/torvalds/linux/commit/abaed0112c1db08be15a784a2c5c8a8b3063cdd3

Saruspete:
Seeing the commit dates back to 2019, I would have a hard time asking to revert it (as from their rule "we do no break user-space"). While I would have liked a way to revert this from bootparam, a simple chmod is easier than adding code just for that permission.

So for the people interested in this collector (seems Facebook/Meta is, as well as many other low-latency oriented companies), we can propose them to execute a chgrp/chmod of this proc-file during startup.

And that way, we could ask the same for slabinfo, and merge the slabinfo collector into proc.plugin too.

what do you think ?

ilyam8:
> as from their rule "we do no break user-space"

Fun fact: yesterday I watched [this video](https://www.youtube.com/watch?v=Nn-SGblUhi4) that is highly related 😄

> And that way, we could ask the same for slabinfo, and merge the slabinfo collector into proc.plugin too: that would be more integrated, and less scary than an exec with setuid/setcap rights.

@vlvkobal @Ferroin what do you think? My opinion is an external collector is better.

Saruspete:
> Fun fact: yesterday I watched [this video](https://www.youtube.com/watch?v=Nn-SGblUhi4) that is highly related smile

Nice video. I wouldn't have expected anything else from LP & LT.

> @vlvkobal @Ferroin what do you think? My opinion is an external collector is better.

External collector has pros and cons.
- Pro: Allows to avoid integrating another optional third-party library (eg, freeipmi, cups, netlink)
- Pro: Allows to grant extended capabilities to restricted binary/code (eg perf, apps)
- Con: increase complexity (new binary, build rules, packaging files...)
- Con: cannot be integrated with netdata.conf (eg, slabinfo has no configuration)

So, should this collector have needed a third party library, or required anything else than a chmod, I would also have been in favor of an external module.
However, the solution here being a chmod to be added either at system startup, or within the service definition, I'd rather keep it within proc.plugin. And also merge the slabinfo module in proc.plugin to be able to have configuation within netdata.conf, especially as its disabled by default.

PR #6864, opened 2019-09-17
test change

emiquelito:

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,10 @@ Netdata is not hosted by the CNCF but is the 3rd most starred open-source projec
 
 ---
 
+
 People get **addicted to Netdata**.<br/>
 Once you use it on your systems, **there is no going back**! *You have been warned...*
+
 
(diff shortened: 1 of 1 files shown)


cakrit:
I saw it. The page shown after it was built/deployed and then tore down is a bit confusing, I wasn't sure if I needed to restart it. Since no action is required and clicking on the link apparently triggers the build/deployment again, a big message to inform users of that might be useful. However, I believe we're done here. If I notice anything else, I'll contact support. Thank you for your help!

emiquelito:
@cakrit thanks for the feedback. You can actually add  a TTL to these checks so they stay up a bit longer. You can use the "auto_deploy_ttl" in the Squash YAML file for this, I believe a 30m TTL should suffice here.

https://www.squash.io/docs/global-yaml-fields/#auto-deploy-ttl

That being said we will explore some ideas to make that process of automatic shutting down a bit less confusing, maybe by adding a notification in the UI as you suggested.

Thanks

PR #6871, opened 2019-09-17
Prevent legend highlighting

jacekkolasa:
### Summary
Prevent highlighting of legend, when shift+clicking

##### Component Name
dashboard.js

### DIFF
--- a/web/gui/console.html
+++ b/web/gui/console.html
@@ -66,7 +66,7 @@
         var netdataTheme = "white"
         var netdataNoBootstrap = true
     </script>
-    <script type="text/javascript" src="dashboard.js?v20190902-0"></script>
+    <script type="text/javascript" src="dashboard.js?v20190917-0"></script>
 </body>
 
 </html>
--- a/web/gui/dash-example.html
+++ b/web/gui/dash-example.html
@@ -412,8 +412,7 @@ class PickNSort {
 		constructor (base_url, link_base_url) {
 			this.base_url = base_url; // URL of netdata host, with port
 			this.link_base_url = link_base_url || base_url; // Reverse proxy URL (Optional)
-			this.link_base_url;
-			this.netdata_info;
+			// this.netdata_info;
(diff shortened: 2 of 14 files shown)


jacekkolasa:
@cakrit i've done it because the highlighting doesn't look good when you toggle dimensions with shift/ctrl. But on second thought, I think that the css fix haven't been done previously because now the user cannot select the dimension name to copy it to clipboard. Do you agree that we shouldn't block selecting when user wants to copy the name?

I can fix this problem in other way - by clearing the selection when user shift+clicks the legend. This has one drawback though - the selection is visible for part of the second. I'll push the commit with that other solution

cakrit:
Wow, I didn't even know it was a thing. Now that I saw what you're talking about:
- Clicking on the first dimension while holding Shift and then clicking on the 4th dimension should select dimensions 1,2,3,4. 
- Clicking on the first dimension while holding Ctrl and then clicking on the 4th dimension should select dimensions 1,4.
 
Without this change, the first case does two wrong things:
 - Selects 1,4 instead of 1-4
 - Highlights the legend

I don't get why Shift and Click highlights things on any page and I'm not sure what's the ideal solution here, but I get why you did what you did and IMHO it's better than what we have today. Perhaps our UX gents (@ivorjvr @oyabun ) have some ideas

jacekkolasa:
Right now (in current UX) there is no difference between shift and ctrl. Moreover, on MacOS only shift+clicking works. Maybe if you use Ctrl on Windows/Linux, user doesn't see the weird highlighting of text.

Of course by "Selecting" I was only meaning highlighting of text, browser native behaviour.

In the future (i recommend after dashboard refractor) we can Toggle range of dimensions via shift+click and fix Ctrl+click for macOS. That should be easy change

cakrit:
Ok, cancel this PR then, put the LGTM fixes on another PR and discuss with the UX guys what we could/should do in the future. You can tell I'm not a MacOS user :)

PR #6875, opened 2019-09-17
fix LGTM warnings

jacekkolasa:
##### Summary
make LGTM happier

### DIFF
--- a/web/gui/dash-example.html
+++ b/web/gui/dash-example.html
@@ -412,7 +412,6 @@ class PickNSort {
 		constructor (base_url, link_base_url) {
 			this.base_url = base_url; // URL of netdata host, with port
 			this.link_base_url = link_base_url || base_url; // Reverse proxy URL (Optional)
-			// this.netdata_info;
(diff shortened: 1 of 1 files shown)


web/gui/dash-example.html
```diff
@@ -412,8 +412,7 @@ <h1>Alarms</h1>
 		constructor (base_url, link_base_url) {
 			this.base_url = base_url; // URL of netdata host, with port
 			this.link_base_url = link_base_url || base_url; // Reverse proxy URL (Optional)
-			this.link_base_url;
-			this.netdata_info;
+			// this.netdata_info;
```
cosmix: You shouldn't leave commented code in the codebase. Why didn't you remove the line?

jacekkolasa: I thought that the contributor left those 2 lines just to indicate that those properties are later used in this class

jacekkolasa:
```suggestion
```

PR #6880, opened 2019-09-18
Remove Dollar sign from Bash code in documentation and fix remark-lint warnings

prhomhyse:
##### Summary

In the [one-line installation](https://docs.netdata.cloud/packaging/installer/#one-line-installation), click "copy to clipboard" button. 

Pasting it on any terminal (or anywhere else) to begin the installation of Netdata:

Below is what gets pasted. 
```sh
$ bash <(curl -Ss https://my-netdata.io/kickstart.sh)
```
This is the output:

```
$: command not found
```

Installation code, of course, works perfectly without the `$` being copied. 

This works as expected:

```
bash <(curl -Ss https://my-netdata.io/kickstart.sh)
```

Another way to tackle this would be to prevent the starting `$` in bash or shell code from getting copied on MkDocs when that button is clicked. 

However, there are already other bash/shell codes that do not have the dollar signs starting them. 

I resolved some remark-lint warnings in the HAProxy and Uninstall docs. 

##### Component Name
Docs

### DIFF
--- a/collectors/charts.d.plugin/apache/README.md
+++ b/collectors/charts.d.plugin/apache/README.md
@@ -26,7 +26,7 @@ It has been tested with apache 2.2 and apache 2.4. The latter also provides conn
 Apache 2.2 response:
 
 ```sh
-$ curl "http://127.0.0.1/server-status?auto"
+curl "http://127.0.0.1/server-status?auto"
 Total Accesses: 80057
 Total kBytes: 223017
 CPULoad: .018287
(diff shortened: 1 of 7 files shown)


cakrit:
Please update the title and the description in the OP, as you've added many other changes here as well. For the future, it's best if you focus each PR on a specific thing.

prhomhyse:
@cakrit noted. 
I'd focus each PR on specific changes. 
I just wanted to clean up the remark-lint warnings on the pages as I found them.

cakrit:
@joelhans please review, haven't looked at the generated html on all these files.

Issue #3925, opened 2018-07-03
Badges: Specify Label & Value sizes and Alignment


ccremer:
Hi, it's me ;)

I just wanted to know whether it's possible to set a fixed label and value width for the netdata badges. As I'm currently developing a custom dashboard where i use badges in a widget. However, visually they do not look misaligned.

Would it be possible to specify the desired label and value length, so that ordering multiple badges vertically would make more sense? Take the screenshot below for example. I would like to set the red line as the alignment reference. Expand the left side so that it ranges from left to red line (grey arrows), while the right (green) part takes the rest of the width.
![badges](https://user-images.githubusercontent.com/12159026/42202987-b4c9f0a8-7e9d-11e8-843a-5060ef7dcc5b.png)

Ideally, the label (& value) can be aligned left, center or right.

underhood:
@cakrit I will take care of this

ilyam8:
fixed in #6903

PR #6903, opened 2019-09-21
#3925 implementation

underhood:
# Summary

#3925
Implements parameters to add fixed width to badges as requested in #3925 

##### Component Name
web

##### Additional Information
If fixed width set the requester is responsible to actually making sure that the width is enough for data displayed.
How to use is documented in relevant docu file

### DIFF
--- a/web/api/badges/web_buffer_svg.c
+++ b/web/api/badges/web_buffer_svg.c
@@ -743,20 +743,40 @@ void buffer_svg(BUFFER *wb, const char *label, calculated_number value, const ch
                 "<rect class=\"bdge-ttl-width\" width=\"%0.2f\" height=\"%0.2f\" rx=\"%0.2f\" fill=\"#fff\"/>"
             "</mask>"
             "<g mask=\"url(#round)\">"
-                "<rect class=\"bdge-rect-lbl\" width=\"%0.2f\" height=\"%0.2f\" fill=\"%s\"/>"
-                "<rect class=\"bdge-rect-val\" x=\"%0.2f\" width=\"%0.2f\" height=\"%0.2f\" fill=\"%s\"/>"
+                "<rect class=\"bdge-rect-lbl\" width=\"%0.2f\" height=\"%0.2f\" fill=\"%s\"/>",
+        total_width, height,
+        total_width, height, round_corner,
+        label_width, height, label_color_escaped); //<rect class="bdge-rect-lbl"
+
+    if(fixed_width_lbl > 0 && fixed_width_val > 0) {
+        buffer_sprintf(wb,
+                "<clipPath id=\"lbl-rect\">"
+                    "<rect class=\"bdge-rect-lbl\" width=\"%0.2f\" height=\"%0.2f\"/>"
+                "</clipPath>",
+        label_width, height); //<clipPath id="lbl-rect"> <rect class="bdge-rect-lbl"
+    }
+
+    buffer_sprintf(wb,
+                "<rect class=\"bdge-rect-val\" x=\"%0.2f\" width=\"%0.2f\" height=\"%0.2f\" fill=\"%s\"/>",
+        label_width, value_width, height, value_color_escaped);
+    
+    if(fixed_width_lbl > 0 && fixed_width_val > 0) {
+        buffer_sprintf(wb,
+                "<clipPath id=\"val-rect\">"
+                    "<rect class=\"bdge-rect-val\" x=\"%0.2f\" width=\"%0.2f\" height=\"%0.2f\"/>"
+                "</clipPath>",
+        label_width, value_width, height);
+    }
+
+    buffer_sprintf(wb,
                 "<rect class=\"bdge-ttl-width\" width=\"%0.2f\" height=\"%0.2f\" fill=\"url(#smooth)\"/>"
             "</g>"
             "<g fill=\"#fff\" text-anchor=\"middle\" font-family=\"DejaVu Sans,Verdana,Geneva,sans-serif\" font-size=\"%0.2f\">"
-                "<text class=\"bdge-lbl-lbl\" x=\"%0.2f\" y=\"%0.0f\" fill=\"#010101\" fill-opacity=\".3\">%s</text>"
-                "<text class=\"bdge-lbl-lbl\" x=\"%0.2f\" y=\"%0.0f\">%s</text>"
-                "<text class=\"bdge-lbl-val\" x=\"%0.2f\" y=\"%0.0f\" fill=\"#010101\" fill-opacity=\".3\">%s</text>"
-                "<text class=\"bdge-lbl-val\" x=\"%0.2f\" y=\"%0.0f\">%s</text>"
+                "<text class=\"bdge-lbl-lbl\" x=\"%0.2f\" y=\"%0.0f\" fill=\"#010101\" fill-opacity=\".3\" clip-path=\"url(#lbl-rect)\">%s</text>"
+                "<text class=\"bdge-lbl-lbl\" x=\"%0.2f\" y=\"%0.0f\" clip-path=\"url(#lbl-rect)\">%s</text>"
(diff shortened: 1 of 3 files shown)


underhood:
sample query to enable and test fixed withd label:
`http://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label=rbo%F0%9F%98%81%F0%9F%98%B3%C5%A1%C5%A1%C5%A1%C5%A1ot&fixed_width_lbl=150&fixed_width_val=200`

web/api/badges/web_buffer_svg.c
```diff
@@ -688,15 +688,15 @@ static inline void calc_colorz(const char *color, char *final, size_t len, calcu
...
+void buffer_svg(BUFFER *wb, const char *label, calculated_number value, const char *units, const char *label_color, const char *value_color, int precision, int scale, uint32_t options, int fixed_width_lbl, int fixed_width_val) {
     char    value_color_buffer[COLOR_STRING_SIZE + 1]
             , value_string[VALUE_STRING_SIZE + 1]
             , label_escaped[LABEL_STRING_SIZE + 1]
             , value_escaped[VALUE_STRING_SIZE + 1]
             , label_color_escaped[COLOR_STRING_SIZE + 1]
             , value_color_escaped[COLOR_STRING_SIZE + 1];
 
-    double label_width, value_width, total_width, height = 20.0, font_size = 11.0, text_offset = 5.8, round_corner = 3.0;
+    double label_width = fixed_width_lbl, value_width = fixed_width_val, total_width, height = 20.0, font_size = 11.0, text_offset = 5.8, round_corner = 3.0;
```
thiagoftsm: Hi @underhood ,

I know that we are talking here about numbers(`double` and `int`), but it is possible some compilers will generate warnings that in this code we are giving directly `int` variables to `double` variables, please, can you add some cast here?

web/api/badges/web_buffer_svg.c
```diff
@@ -866,7 +885,7 @@ int web_client_api_request_v1_badge(RRDHOST *host, struct web_client *w, char *u
     if(!st) st = rrdset_find_byname(host, chart);
     if(!st) {
         buffer_no_cacheable(w->response.data);
-        buffer_svg(w->response.data, "chart not found", NAN, "", NULL, NULL, -1, scale, 0);
+        buffer_svg(w->response.data, "chart not found", NAN, "", NULL, NULL, -1, scale, 0, -1, -1);
```
thiagoftsm: Why are you not using the variables `fixed_width_lbl` and `fixed_width_lbl_str` here instead `-1, -1`?

thiagoftsm: I was studying the issue that generated it, and I think I understood.

underhood: thinking behind is if there is any error default to self scaling to be actually able to see the error (as the caller will give space expecting the good output). But maybe i can make it fixed in all cases.

thiagoftsm:
> Resolving package dependencies...
> Problem: curl-7.65.3-1.2.x86_64 requires libcurl4 = 7.65.3, but this requirement cannot be provided
>   not installable providers: libcurl4-7.65.3-1.2.i586[repo-oss]
>                    libcurl4-7.65.3-1.2.x86_64[repo-oss]
> ```

Do not worry with these kind of message, sometimes these false positive happen, but as you can see after your last commit it was not more present.

underhood:
Hi,

please read the docs i also created for the feature.
It is fixed width. Therefore we do not in this case check if the place is big enough. Docs state that in such case requester is responsible to provide enough space.
Of course I can do if the fixed_width given is not enough I increase the valuse to minimum needed,
_**BUT** wouldn't that defeat the purpose_ of **fixed_width_label**.

thiagoftsm:
@underhood I understand your arguments, I only found strange we accept that the label can cross the value side, I will wait other opinions  here to check that this is acceptable.
I am also adding another member of the core here to express his opinion, @ilyam8 , because I know his eye is very good to see if this change is good for user eyes.
I am not saying that your PR is wrong, I am only believing that to the user eyes this also can be a problem.

ilyam8:
@thiagoftsm 

The request is:
 - Specify Label & Value sizes and Alignment

This PR adds fixed width label and value. And user is responsible to set "correct"  width. Looks good to me.

> to increase width if it is smaller than needed.

I agree with @underhood that this is wrong.

But at the same time i agree with @thiagoftsm  that `text len > width` cases should be handled. I remember that @cakrit suggested a nice approach somewhere (slack). But this can be a separate PR.

@underhood what about alignment? 

@ccremer it would be nice to have some inputs from you

underhood:
with latest changes oversize label will get clipped as suggested by @cakrit as opposed to overflowing to the right side of label.
https://underhoodbadge-width-zj0xi.squash.io/api/v1/badge.svg?chart=system.cpu&label=rbo😁😳ššššot&fixed_width_lbl=50&fixed_width_val=200
Capture

thiagoftsm:
I verified that we had a considerable improvement from the last time I tested your PR, and I liked a lot, I will only give more one suggestion for we improve a little more. I executed two different tests, the first using utf-8, I sent

curl -k -o cmp.svg "https://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label=rbo%F0%9F%98%81%F0%9F%98%B3%C5%A1%C5%A1%C5%A1%C5%A1ot&fixed_width_lbl=150&fixed_width_val=200"

cmp

After this I cut the label width from 150 to 15
curl -k -o cmp2.svg "https://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label=rbo%F0%9F%98%81%F0%9F%98%B3%C5%A1%C5%A1%C5%A1%C5%A1ot&fixed_width_lbl=15&fixed_width_val=200"

cmp2

Finally I executed tests using only ASCII text:
curl -k -o cmp4.svg "https://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label=abcdefghijklmnopqrstuvxz&fixed_width_lbl=150&fixed_width_val=200"

cmp4

And again I cut from 150 to 15:

curl -k -o cmp5.svg "https://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label=abcdefghijklmnopqrstuvxz&fixed_width_lbl=15&fixed_width_val=200"

cmp5

As you can see in both cases we did not cut the labels from the beginning, could you adjust your code for we cut from the start?

underhood:
@thiagoftsm Hi currently all texts are center aligned all the time. I prefer not to add exception to this behavior in same cases for the sake of consistency.
In future I plan to rework this a bit -> hint ugly inline svg and javasctipt code inside `c` code
then i can implement also alignment parameters

thiagoftsm:
I understood your point, and I think we can keep the behavior to keep the texts centralized, but in my vision, we could do some kind of math to first select the part of the text that will be print and after this to cut the text, with this we could cut the text using the initial part of it.
Another thing, considering that we agreed to cut the text, I will ask you to update the documentation giving this additional information.
Thanks!

joelhans:
Thanks for the fixes. Docs LGTM.

ccremer:
Hello guys.
I am the initial requester of this feature.
It's very nice that this gets added to netdata, however, meanwhile I do not need this functionality for a dashboard anymore, now Grafana is the more flexible approach for my use case.

But the use case would have been to be able to stack multiple badges vertically, and each would be the same width. Whether the labels and values are centrally, right or left aligned, is up to the user to decide what looks best, IMO.

thiagoftsm:
Hi @ccremer ,

In my vision we can do what you wrote by steps, we first need to finish this, that fixes part of the problem and after this we can work with the possibility to bring more information in an unique request.
For sure I would like to read the opinion of PR author about his next steps in the direction to bring what you wrote.

Best regards!

cakrit:
@thiagoftsm If it works and doesn't cause issues, please approve. The issue with the centering isn't that important.

PR #6911, opened 2019-09-23
Detect deadlock in dbengine page cache

### LINKED ISSUE #6910: DBengine hanging without logging
##### Bug report summary

We have a server collecting about 16,000 dimensions (mainly due to per user cgroups)

The `page cache size` had been left at the default of 32Mb.

Netdata was repeatedly hanging and a junior admin who was looking into it ended up spending a couple of days chasing down lots of red herrings.

After we finally took some stack backtraces we could see nearly all the threads were stuck in [`sleep_usec` in `pg_cache_reserve_pages`](https://github.com/netdata/netdata/blob/v1.17.1/database/engine/pagecache.c#L252)

This was able to pinpoint the issue as not enough page cache.

Increasing `page cache size` fixed the hanging.

There was no clue in the error.log that the `page cache size` was too small.

##### OS / Environment

CloudLinux 7.6 running cPanel/WHM 11.78
208 user accounts

netdata v1.17.1

( installed from [PackageCloud EL7 Repo](https://packagecloud.io/netdata/netdata/el/7/x86_64) )

##### Component Name

DBEngine Logging

##### Steps To Reproduce

Configure netdata to collect too many dimensions for the configured `page_cache_size`

##### Expected behavior

Some kind of warning that there are contention issues with the `page cache`.

Maybe once the failures has reached the failure ceiling then a warning would be appropriate.

mfundul:
##### Summary
Fixes #6910
##### Component Name
database/engine
##### Additional Information
Detect deadlock in dbengine page cache when there are too many metrics and print error message.


cakrit:
I know that this is what the user asked, but it looks like with this we'll leave netdata hanging and log how many error messages until someone realizes what happened and reconfigures it? 

I believe we need a warning and then a critical alarm when we are getting close to the page cache limit. I'd also change the behavior to stop handling new metrics, so that we remain within the limit, always remaining in a critical alarm state. 

No problem merging this if it won't fill the log with errors, but it definitely doesn't fix the problem, regardless of the specificity of the user request.

KrisShannon:
TBH,  When I said logging I just meant some kind of visible warning.

An alarm should be fine,  especially if it also stops handling new metrics so the web UI remains responsive

mfundul:
This will not fill the log but the deadlock will remain as it is currently.

mfundul:
With the latest commit it should now drop metrics so as to resolve the deadlock condition and raise a corresponding alarm. Netdata remains responsive.

mfundul:
> 
> 
> Hi @mfundul ,
> 
> I began to test your PR now, and it was reporting a log of messages inside my error.log, before to bring the images that will show what I am saying, let me describe the tests that I did:
> 
>     * I compiled Netdata using: `CFLAGS="-O1 -ggdb -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1" ./netdata-installer.sh --disable-go --disable-lto --dont-wait`
> 
>     * I changed the file `/usr/libexec/netdata/python.d/example.chart.py` to have 16000 dimensions instead 4, I did this changing the line: `for i in range(1, 16000):`
> 
>     * my netdata.conf had the following parameters:
> 
> 
> ```
> memory mode = dbengine
> page cache size = 32
> dbengine disk space = 256
> ```
> 
> I ran Netdata during only 6 minutes with these configuration and I had more than 200000 messages:
> err33

I've changed it to only print once per dbengine instance.

Issue #6849, opened 2019-09-15
netdata/packaging: Build from source for Judy-devel and libuv for Centos7 x86


paulkatsoulakis:
##### Bug report summary
Based on Travis build: https://travis-ci.com/netdata/netdata/jobs/235311638
We don't seem to have an official repository i know of that provides Judy-devel and libuv.

We have to build from source for those two particular packages, for x86 Centos-7

##### OS / Environment
CentOS 7, 32 bit (x86)

##### Netdata version (ouput of `netdata -V`)
Latest

##### Component Name
netdata/packaging

##### Steps To Reproduce
Just trigger the 32 bit RPM package build, it will fail on CentOS 7
Sample: https://travis-ci.com/netdata/netdata/jobs/235311638

##### Expected behavior
No failures

paulkatsoulakis:
We already have a way to build judy from source four our docker images, i will be using this method for the OSes we do not have a binary package available.

paulkatsoulakis:
Got a script for libuv to compile under 32 bit centos 7 and got a test failing 

you try to build latest release from their repository for libuv and fails a test ```FAIL: test/run-tests
======================================================
1 of 1 test failed
Please report to https://github.com/libuv/libuv/issues
======================================================
make[1]: *** [Makefile:5209: check-TESTS] Error 1
make[1]: Leaving directory '/opt/libuv-1.32.0'
make: *** [Makefile:5469: check-am] Error 2
``` ```not ok 48 - fs_copyfile
# exit code 6
# Output from process `fs_copyfile`:
# Assertion failed in test/test-fs-copyfile.c on line 202: req.result == UV_EACCES
```

paulkatsoulakis:
```[root@4a8d3561eb92 test]# cat /etc/os-release 
NAME=Fedora
VERSION="29 (Container Image)"
ID=fedora
VERSION_ID=29
VERSION_CODENAME=""
PLATFORM_ID="platform:f29"
PRETTY_NAME="Fedora 29 (Container Image)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:29"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f29/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
[...]

PR #6914, opened 2019-09-23
netdata/ci: nits and fixes around package release workflow

paulkatsoulakis:
##### Summary
As per the latest changes we have been doing recently, we raised a couple of sneaky bugs on travis workflow evaluation. Below is the list of fixes in this PR

1) We removed i386 from CentOS 7 due to #6849 as requested by @cakrit. 
    This made us unintentionally break the parsing of .travis.yml from Travis CI.  The reason is that travis expects the first job within a stage not to be part of the sequence of jobs (meaning the first "name" key in the stage, should not be added with a dash). The result of this was not catastrophic though, but it was messing up presentation on Travis, since the packaging job was showing up as part of the previous stage which was wrong.  To mitigate this we simply changed the order and put Centos-6 first, since that one is enabled for all possible scenarios (i386, amd64)

2) We renamed "Packaging for release", which was wrong and retarded and was a leftover from what i took over in March. We now call it "Support activities on main branch", since we run on every commit on master.

3) Within this "Support activities on main branch", we now run first the labeler that runs unconditionally and is always present (to avoid breaking presentation again). We also made the changelog generation and tagging of release conditional, to avoid showing it up if it was not going to do anything.  To do this, we basically brought all the checks the script does internally, on the surface.  
  We kind of duplicating conditions like this, but it is kind of inevitable because the scripts inside take more decisions based on the commit message.  For now we can survive like this, because we sanitise the presentation of travis steps and we can do another iteration to perhaps deprecate `.travis/tagger.sh` and simplify `.travis/generate_changelog_and_tag_release.sh` completely 

### **UPDATE**:
With commit https://github.com/netdata/netdata/pull/6914/commits/69897bb35c829a4ec736473cacbb1968481b33d8 we basically clean up tagger and changelog generation logic for the release.
The commit has all the details, i will avoid explaining here again.  In very high level, we basically brought the tagging logic and conditions up on travis.yml and then we did a couple of renames of files to be slightly more consistent.

##### Component Name
netdata/ci

##### Additional Information
N/A


paulkatsoulakis:
Do not merge this yet, i want to attempt to clean up more

paulkatsoulakis:
I am not sure what happened to Codady, i just went i and clicked "analyse again" to see what happens. Probably multiple force pushes made it crap out? Will see.

PR #6916, opened 2019-09-23
Suggest using /run or /var/run for the unix socket

### LINKED ISSUE #3347: opensuse - bind to unix socket and nginx
I tried setting up the netdata - nginx combo over unix socket and ran into `502 Bad Gateway`. Following the instructions at [1], I used

```
# cat /opt/netdata/etc/netdata/netdata.conf
...
[web]
        bind to = unix:/tmp/netdata.sock
# cat /etc/nginx/conf.d/netdata.conf
upstream backend {
    server unix:/tmp/netdata.sock;
    keepalive 64;
}
...
```

This did not work for me on opensuse Tumbleweed. The actual error from `/var/log/nginx/error.log`:

```
2018/01/26 15:28:01 [crit] 32428#32428: *26 connect() to unix:/tmp/netdata.sock failed (2: No such file or directory) while connecting to upstream, client: 127.0.0.1, server: example.com, request: "GET /status?full&json HTTP/1.1", upstream: "http://unix:/tmp/netdata.sock:/status?full&json", host: "127.0.0.1"
```

After some looking around, I found an answer to the problem at [2]. The fix was to put the socket inside `/run/netdata` like `bind to = unix:/run/netdata/netdata.sock`.

Maybe this should be put into the wiki.

[1] https://github.com/firehol/netdata/wiki/Running-behind-nginx#limit-direct-access-to-netdata
[2] https://serverfault.com/questions/463993/nginx-unix-domain-socket-error/464025#464025

cakrit:
##### Summary
Fixes #3347

##### Component Name
docs

##### Additional Information
Don't use /tmp for unix sockets

### DIFF
--- a/web/server/README.md
+++ b/web/server/README.md
@@ -33,7 +33,7 @@ The ports to bind are controlled via `[web].bind to`, like this:
 ```
 [web]
    default port = 19999
-   bind to = 127.0.0.1=dashboard^SSL=optional 10.1.1.1:19998=management|netdata.conf hostname:19997=badges [::]:19996=streaming^SSL=force localhost:19995=registry *:http=dashboard unix:/tmp/netdata.sock
+   bind to = 127.0.0.1=dashboard^SSL=optional 10.1.1.1:19998=management|netdata.conf hostname:19997=badges [::]:19996=streaming^SSL=force localhost:19995=registry *:http=dashboard unix:/run/netdata/netdata.sock
 ```
 
 Using the above, Netdata will bind to:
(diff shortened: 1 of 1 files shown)


CtrlAltDel64:
I recommend this be changed here as well:
https://docs.netdata.cloud/docs/running-behind-nginx/#limit-direct-access-to-netdata

cakrit:
@CtrlAltDel64 mind doing a PR for it?

web/server/README.md
```diff
@@ -33,7 +33,7 @@ The ports to bind are controlled via `[web].bind to`, like this:
 ```
 [web]
    default port = 19999
-   bind to = 127.0.0.1=dashboard^SSL=optional 10.1.1.1:19998=management|netdata.conf hostname:19997=badges [::]:19996=streaming^SSL=force localhost:19995=registry *:http=dashboard unix:/tmp/netdata.sock
+   bind to = 127.0.0.1=dashboard^SSL=optional 10.1.1.1:19998=management|netdata.conf hostname:19997=badges [::]:19996=streaming^SSL=force localhost:19995=registry *:http=dashboard unix:/run/netdata/netdata.sock
```
thiagoftsm: Please, let us use as suggestion `/var/run`, because `/run` is present only on Linux as you can see in this screenshot from FreeBSD:

VirtualBox_FreeBSD_27_10_2019_19_37_34

PR #6926, opened 2019-09-23
netdata.spec.in: Do not build CUPS plugin subpackage on CentOS 6 and CentOS 7

knatsakis:
##### Summary
Do not build CUPS plugin subpackage on CentOS 6 and CentOS 7

##### Component Name
RPM packaging

##### Additional Information
According to #5350 the CUPS plugin needs a cups library newer than 1.7, but CentOS 6 and CentOS 7 include versions 1.4.2 and 1.6.3 respectively. This causes the library to not be detected by ./configure and the CUPS plugin to not being build. This, in turn causes the packaging jobs for CentOS 6 and CentOS 7 to fail.

This pull request disables the CUPS plugin for CentOS 6 and CentOS 7. A test build can be found here:
https://travis-ci.org/knatsakis/netdata/builds/588618391

### DIFF
--- a/netdata.spec.in
+++ b/netdata.spec.in
@@ -192,7 +192,9 @@ Requires: freeipmi
 # end - freeipmi plugin dependencies
 
 # CUPS plugin dependencies
+%if 0%{?centos_ver} != 6 && 0%{?centos_ver} != 7
 BuildRequires: cups-devel
+%endif
 # end - cups plugin dependencies
 
 # Prometheus remote write dependencies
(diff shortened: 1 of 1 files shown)


paulkatsoulakis:
Great catch, just keep in mind that this will be a problem for all installation channels.
Our kickstart is relying on the respective packaging system of the OS to get the dependencies in.

I will be updating the documentation with a PR, just mentioning here the comment for the record.
I will also refer this PR on the documentation fix, thanks !

ScrumpyJack:
The requirement for cups-devel 1.7 also causes builds on Oracle Linux 6 and 7 to fail

ilyam8:
@ScrumpyJack if there is a bug in latest please open an issue

PR #6931, opened 2019-09-24
minor - check for curl to not get wrong error message

underhood:
##### Summary
Very small change -> Check for `curl` presence in system.
Before if `curl` is not installed you will get error message (and start investigating what is wrong with netdata):
```bash
using netdata server at: http://127.0.0.1:19999
Cannot download charts from server: http://127.0.0.1:19999
```
even if netdata running and responding well.
##### Component Name
tests/stress.sh

### DIFF
--- a/tests/stress.sh
+++ b/tests/stress.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+if ! hash curl 2>/dev/null
+then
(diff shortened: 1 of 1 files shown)


underhood:
I can look at it when i have some time again. This PR was meant to only give more normal error message for the next guy _(so he doesn't loose 10 minutes wondering what is going on like I did)_.
Maybe better to make that next PR anyway.

cakrit:
Ok, merging it as is then, it's an improvement after all.

PR #6936, opened 2019-09-24
[RFC] Buffer.c per thread mempool

underhood:
### Summary
**_Bigger change there may be some rough edges at this point_** 

I will test and verify more however as much as I can see and according to tests I ran it seems to be working stable.

During beginning of my investigation of Netdata and how it works I noticed in logs lot of time we allocate and free same size of BUFFER in cycle.

As I investigated further I decided mempooling might be an advantage. Therefore small mempool is created for each thread when buffer_create is called first time. When buffer is free and not too big _(`BUFFER_MEMPOOL_MAXSIZE 1024*16`)_ instead of calling free we keep the memory for reuse next time someone requests buffer _(of same or smaller size)_. Therefore it is possible to get bigger buffer than necessary when calling buffer_create but that is not a problem. There is limit on biggest BUFFER that can be mempooled and this is done to protect memory usage from going crazy (as I saw buffers occasionally growing to almost 1MB - we will not mempool such buffers).

I have created small benchmark in `tests/profile/benchmark-buffer.c` that will stress test the buffer creation and destruction on 8 threads in parallel.

To run the bench:
1. `autoreconf -ivf`
1.  compile netdata without mempool `configure --disable-bufmempool` and `make`
1.  `cd tests/profile && rm -f ./benchmark-buffer && make benchmark-buffer && time ./benchmark-buffer` (the rm -f ./benchmark-buffer is needed as otherwise makfile thinks nothing to compile as this makefile doesn't track change in libnetdata).
1.  compile netdata with mempool `configure` and `make`
1.  `cd tests/profile && rm -f ./benchmark-buffer && make benchmark-buffer && time ./benchmark-buffer` (the rm -f ./benchmark-buffer is needed as otherwise makfile thinks nothing to compile as this makefile doesn't track change in libnetdata).

Results from my development virtual machine (limited to 3 cores, Ubuntu AMD64) are:
**mempooling disabled (running in 3 core VM in 8 threads)**
```
Benchmarking netdata Buffer.c
Mempool is: DISABLED
Starting Test 1
Starting Test 2
Starting Test 3
ALL SUCCESSFUL.

real    0m28,419s
user    1m24,314s
sys     0m0,337s
```
**NEW LOGIC mempool enabled (running in 3 core VM in 8 threads)**
```
Benchmarking netdata Buffer.c
Mempool is: ACTIVE
Starting Test 1
Starting Test 2
Starting Test 3
ALL SUCCESSFUL.

real    0m16,326s
user    0m48,306s
sys     0m0,201s
```

Therefore it seems that mempooling gives performance advantage.

Additionally I noticed for huge buffers we are increasing step by step (1024 bytes) for single buffer in some cases ~20 times. Therefore for big buffers growth algorithm has been tweaked to not need so often to increase and copy memory.

Run the benchmark again with my VM increased from 3->6 core limit (my PC is 8 core AMD construction machine with 32 gigs of RAM, older but should be enough) with following results:
**NEW LOGIC: mempool ACTIVE (running in 6 core VM in 8 threads)**
```
Benchmarking netdata Buffer.c
Mempool is: ACTIVE
Starting Test 1
Starting Test 2
Starting Test 3
ALL SUCCESSFUL.

real	0m10,250s
user	0m53,058s
sys	0m0,105s
```
**mempool DISABLED (running in 6 core VM in 8 threads)**
```
Benchmarking netdata Buffer.c
Mempool is: DISABLED
Starting Test 1
Starting Test 2
Starting Test 3
ALL SUCCESSFUL.

real	0m21,127s
user	1m49,435s
sys	0m0,163s
```

**Directly on metal (no VM 8core and 8threads) old:**
```
real    0m12,797s
user    1m29,782s
sys     0m0,188s
```
**new**
```
real    0m5,543s
user    0m40,398s
sys     0m0,138s
```

All numbers with DEBUGGING disabled of course. If you enable debug logs the buffer_mempool is logging a lot more than older implemetation. Didn't test performance in DEBUG mode though.

Additional advantage of this PR is reduced memory fragmentation.

##### Component Name
libnetdata/buffer/buffer.c

### Libuv migration
All pthreads functions used have direct equivalents in libuv. Therefore this change should not affect migration.


libnetdata/buffer/buffer.c
```diff
@@ -29,6 +29,14 @@ static inline void _buffer_overflow_check(BUFFER *b, const char *file, const cha
     }
 }
 
+//to know how we are configured at runtime
+char* buffer_mempool_status() {
```
thiagoftsm: We are trying to define a new pattern for Netdata code, this patterns says for we use `*` near the function name instead the type.

cakrit: We said we'll first do a PR that applies the new code style to all existing code and the ask everyone writing C code to utilize the `.clang-format` we have in the repo. I say let's first concentrate on whether there's anything wrong with what -based on the benchmarks- appears to be an excellent improvement.

underhood: agreed, that is a point of this PR at this stage. I have to go over it again myself as at this point there might be skeletons in closet. I should have more time to work on it during weekend.

underhood:
Additionally results with just test 2 running 3 times and running at 32 threads:
```
Benchmarking netdata Buffer.c
Mempool is: DISABLED
Starting Test 1
Starting Test 2
Starting Test 3
ALL SUCCESSFUL.

real    0m26,570s
user    3m26,997s
sys     0m0,476s

Benchmarking netdata Buffer.c
Mempool is: DISABLED
Starting Test 1
Starting Test 2
Starting Test 3
ALL SUCCESSFUL.

real    0m26,074s
user    3m22,579s
sys     0m0,437s
```
vs,

```
Benchmarking netdata Buffer.c
Mempool is: ACTIVE
Starting Test 1
Starting Test 2
Starting Test 3
ALL SUCCESSFUL.

real    0m15,705s
user    2m1,720s
sys     0m0,319s

Benchmarking netdata Buffer.c
Mempool is: ACTIVE
Starting Test 1
Starting Test 2
Starting Test 3
ALL SUCCESSFUL.

real    0m15,690s
user    2m1,894s
sys     0m0,243s
```

underhood:
With latest commit the Buffer Mempool statistics can be seen directly in netdata itself in `netdata monitoring/buffer mempool`. For this Netdata has to be compiled with `BUFFER_MEMPOOL_STATS`:
``` C
#define BUFFER_USE_MEMPOOL 1
#define BUFFER_USE_MEMPOOL_PER_THREAD 1

#define BUFFER_MEMPOOL_STATS 1
```
Additional stats are incoming. Here is the graph after clean start of netdata and many page refreshes and some playing around where it can be seen it seems to avoid lot of malloc/free cycles (reduction in heap fragmentation and improvement in performace under high parallel loads)
Screenshot from 2019-10-05 18-27-13

underhood:
closed due to no interest and not maintained for years

PR #6948, opened 2019-09-25
netdata/packaging: Make spec file more consistent with version dependencies, plus some documentation nits

paulkatsoulakis:
##### Summary
Sice we have identified that we missed a critical CUPS version dependency, we need to update our spec file so that we guarantee we always depend on the right versions of CUPS.
This way we ensure we won't bump into build problems in the future.

Also, take note of the CUPs version required in the distributions table

##### Component Name
netdata/packaging
netdata/docs

##### Additional Information
N/A

### DIFF
--- a/netdata.spec.in
+++ b/netdata.spec.in
@@ -193,7 +193,7 @@ Requires: freeipmi
 
 # CUPS plugin dependencies
 %if 0%{?centos_ver} != 6 && 0%{?centos_ver} != 7
-BuildRequires: cups-devel
+BuildRequires: cups-devel >= 1.7
 %endif
 # end - cups plugin dependencies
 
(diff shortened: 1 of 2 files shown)


paulkatsoulakis:
@cakrit, in order to denote the versions CUPs is not available, we probably need to introduce a distributions table under each capability and tick the availability of the functionality.

Does that make sense or you think it would be too much info?
It may become overwhelming, given we want to cover a great range of distributions

cakrit:
> @cakrit, in order to denote the versions CUPs is not available, we probably need to introduce a distributions table under each capability and tick the availability of the functionality.
> 
> Does that make sense or you think it would be too much info?
> It may become overwhelming, given we want to cover a great range of distributions

We can start by adding what we know in the relevant paragraphs of the distributions md. All of this information, including supported installation methods will eventually become a proper app eventually, giving users the specifics for their distro only. So the info will be captured in a format that's easy for an app to process (yaml or json), instead of markdown.

Issue #6934, opened 2019-09-24
Galera cluster status monitoring


ekexcello:
##### Feature idea summary
It would be good to see WSREP cluster status and size, not just replication flow that is already present in Netdata.
For example, following parameters are missing:
`

SHOW STATUS LIKE 'wsrep_cluster_weight'
Variable_name: wsrep_cluster_weight
        Value: 1
SHOW STATUS LIKE 'wsrep_cluster_size'
Variable_name: wsrep_cluster_size
        Value: 1
SHOW STATUS LIKE 'wsrep_cluster_status'
Variable_name: wsrep_cluster_status
        Value: Primary
SHOW STATUS LIKE wsrep_local_state
Variable_name: wsrep_local_state
        Value: 4
SHOW STATUS LIKE 'wsrep_local_state_comment'
Variable_name: wsrep_local_state_comment
        Value: Synced
SHOW STATUS LIKE 'wsrep_open_transactions'
Variable_name: wsrep_open_transactions
        Value: 18
SHOW STATUS LIKE 'wsrep_connected'
Variable_name: wsrep_connected
        Value: ON
SHOW STATUS LIKE 'wsrep_local_index'
Variable_name: wsrep_local_index
        Value: 0
SHOW STATUS LIKE 'wsrep_ready'
Variable_name: wsrep_ready
        Value: ON
SHOW STATUS LIKE 'wsrep_thread_count'
Variable_name: wsrep_thread_count
        Value: 98
`
Most of these parameters are strings, so they should be displayed like status table in Netdata interface.

These parameters can be used for status monitoring, for example, when **wsrep_cluster_status** is changed, an alert should be sent to 'dba' or 'admin' contact.
Also, when node leaves the cluster or joines it (thus, **wsrep_cluster_size** is changed) an alert should be sent, too, for example, if node left the cluster, this event is critical (while joining can be just notification).

As from what we can see now, Netdata already extracts information from same variable group, so adding these parameters should also be possible, shouldn't it?

ilyam8:
Hi @ekexcello 

Can we get all the info in one request?

ilyam8:
Oh, i think all those values are in `SHOW GLOBAL STATUS;` response, right?

ekexcello:
Hello Ilya,
Yes, all these variables are present in `SHOW GLOBAL STATUS;` or even in `SHOW STATUS LIKE 'wsrep_%';` output.

ilyam8:
[...]

PR #6962, opened 2019-09-27
mysql: collect galera cluster metrics

ilyam8:
##### Summary

Fixes: #6934

Info sources:
 - https://mariadb.com/kb/en/library/galera-cluster-status-variables
 - https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html

This PR adds:
   - [X] cluster charts 
   - [x] new charts description (dashboard_info)
   - [x] alarms

New collected metrics:
 - `wsrep_cluster_weight`
 - `wsrep_cluster_size`
 - `wsrep_cluster_status`
 - `wsrep_local_state`
 - `wsrep_open_transactions`
 - `wsrep_connected`
 - `wsrep_ready`
 - `wsrep_thread_count`

##### Component Name

[`/collectors/python.d.plugin/mysql`](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/mysql)

##### Additional Information

cc: @ekexcello

### DIFF
--- a/collectors/python.d.plugin/mysql/mysql.chart.py
+++ b/collectors/python.d.plugin/mysql/mysql.chart.py
@@ -744,7 +744,7 @@ class WSRepDataConverter:
     def convert(self, key, value):
         if key == 'wsrep_connected':
             return self.convert_connected(value)
-        if key == 'wsrep_ready':
+        elif key == 'wsrep_ready':
             return self.convert_ready(value)
         elif key == 'wsrep_cluster_status':
             return self.convert_cluster_status(value)
--- a/collectors/python.d.plugin/mysql/mysql.chart.py
+++ b/collectors/python.d.plugin/mysql/mysql.chart.py
@@ -628,7 +628,7 @@ CHARTS = {
     'galera_cluster_size': {
         'options': [None, 'Number of Nodes in the Cluster', 'num', 'galera', 'mysql.galera_cluster_size', 'line'],
         'lines': [
-            ['wsrep_cluster_size', 'size', 'absolute'],
+            ['wsrep_cluster_size', 'nodes', 'absolute'],
         ]
     },
     'galera_cluster_weight': {
--- a/collectors/python.d.plugin/mysql/mysql.chart.py
+++ b/collectors/python.d.plugin/mysql/mysql.chart.py
@@ -117,6 +117,14 @@ GLOBAL_STATS = [
     'Connection_errors_peer_address',
     'Connection_errors_select',
     'Connection_errors_tcpwrap',
+    'Com_delete',
+    'Com_insert',
+    'Com_select',
+    'Com_update',
+    'Com_replace'
+]
+
+GALERA_STATS = [
     'wsrep_local_recv_queue',
     'wsrep_local_send_queue',
     'wsrep_received',
@@ -126,11 +134,15 @@ GLOBAL_STATS = [
     'wsrep_local_bf_aborts',
     'wsrep_local_cert_failures',
     'wsrep_flow_control_paused_ns',
-    'Com_delete',
-    'Com_insert',
-    'Com_select',
-    'Com_update',
-    'Com_replace'
+    'wsrep_cluster_weight',
+    'wsrep_cluster_size',
+    'wsrep_cluster_status',
+    'wsrep_local_state_comment',
+    'wsrep_open_transactions',
+    'wsrep_connected',
+    'wsrep_local_index',
+    'wsrep_ready',
+    'wsrep_thread_count'
 ]
 
 
@@ -216,7 +228,15 @@ ORDER = [
     'galera_queue',
     'galera_conflicts',
     'galera_flow_control',
-    'userstats_cpu'
+    'galera_cluster_status',
+    'galera_cluster_state',
+    'galera_cluster_size',
+    'galera_cluster_weight',
+    'galera_connected',
+    'galera_ready',
+    'galera_open_transactions',
+    'galera_thread_count',
+    'userstats_cpu',
 ]
 
 CHARTS = {
@@ -594,6 +614,59 @@ CHARTS = {
             ['wsrep_flow_control_paused_ns', 'paused', 'incremental', 1, 1000000],
         ]
     },
+    'galera_cluster_status': {
+        'options': [None, 'Cluster Component Status', 'status', 'galera', 'mysql.galera_cluster_status', 'line'],
+        'lines': [
+            ['wsrep_cluster_status', 'status', 'absolute'],
+        ]
+    },
+    'galera_cluster_state': {
+        'options': [None, 'Cluster Component State', 'state', 'galera', 'mysql.galera_cluster_state', 'line'],
+        'lines': [
+            ['wsrep_local_state_comment', 'state', 'absolute'],
+        ]
+    },
+    'galera_cluster_size': {
+        'options': [None, 'Number of Nodes in the Cluster', 'num', 'galera', 'mysql.galera_cluster_size', 'line'],
+        'lines': [
+            ['wsrep_cluster_size', 'size', 'absolute'],
+        ]
+    },
+    'galera_cluster_weight': {
+        'options': [None, 'The Total Weight of the Current Members in the Cluster', 'weight', 'galera',
+                    'mysql.galera_cluster_weight', 'line'],
+        'lines': [
+            ['wsrep_cluster_weight', 'weight', 'absolute'],
+        ]
+    },
+    'galera_connected': {
+        'options': [None, 'Whether the Node is Connected to the Cluster', 'boolean', 'galera',
+                    'mysql.galera_connected', 'line'],
+        'lines': [
+            ['wsrep_connected', 'connected', 'absolute'],
+        ]
+    },
+    'galera_ready': {
+        'options': [None, 'Whether the Node is Ready to Accept Queries', 'boolean', 'galera',
+                    'mysql.galera_ready', 'line'],
+        'lines': [
+            ['wsrep_ready', 'ready', 'absolute'],
+        ]
+    },
+    'galera_open_transactions': {
+        'options': [None, 'The Number of Locally Running Registered Inside the WSRep Provider Transactions', 'num',
+                    'galera', 'mysql.galera_open_transactions', 'line'],
+        'lines': [
+            ['wsrep_open_transactions', 'open transactions', 'absolute'],
+        ]
+    },
+    'galera_thread_count': {
+        'options': [None, 'Total Number of WSRep (applier/rollbacker) Threads', 'num', 'galera',
+                    'mysql.galera_thread_count', 'line'],
+        'lines': [
+            ['wsrep_thread_count', 'threads', 'absolute'],
+        ]
+    },
     'userstats_cpu': {
         'options': [None, 'Users CPU time', 'percentage', 'userstats', 'mysql.userstats_cpu', 'stacked'],
         'lines': []
@@ -663,6 +736,65 @@ def userstats_chart_template(name):
 DEFAULT_REPL_CHANNEL = ''
 
 
+# Write Set REPlication
+# https://galeracluster.com/library/documentation/galera-status-variables.html
+# https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html
+class WSRepDataConverter:
+    unknown_value = -1
+
+    def convert(self, key, value):
+        if key == 'wsrep_connected':
+            return self.convert_connected(value)
+        if key == 'wsrep_ready':
+            return self.convert_ready(value)
+        elif key == 'wsrep_cluster_status':
+            return self.convert_cluster_status(value)
+        elif key == 'wsrep_local_state_comment':
+            return self.convert_local_state_comment(value)
+        return value
+
+    def convert_connected(self, value):
+        # https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html#wsrep_connected
+        if value == 'OFF':
+            return 0
+        if value == 'ON':
+            return 1
+        return self.unknown_value
+
+    def convert_ready(self, value):
+        # https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html#wsrep_ready
+        if value == 'OFF':
+            return 0
+        if value == 'ON':
+            return 1
+        return self.unknown_value
+
+    def convert_local_state_comment(self, value):
+        # https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html#wsrep_local_state_comment
+        if value == 'Joining':
+            return 1
+        elif value == 'Donor' or value == 'Desynced':
+            return 2
+        elif value == 'Joined':
+            return 3
+        elif value == 'Synced':
+            return 4
+        return self.unknown_value
+
+    def convert_cluster_status(self, value):
+        # https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html#wsrep_cluster_status
+        if value == 'Primary':
+            return 1
+        elif value == 'Non-Primary':
+            return 2
+        elif value == 'Disconnected':
+            return 3
+        return self.unknown_value
+
+
+wsrep_converter = WSRepDataConverter()
+
+
 class Service(MySQLService):
     def __init__(self, configuration=None, name=None):
         MySQLService.__init__(self, configuration=configuration, name=name)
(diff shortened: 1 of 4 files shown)


ilyam8:
@ekexcello

i added all metrics[ you requested](https://github.com/netdata/netdata/issues/6934).

Could you test the PR? If i need to add something else feel free to suggest. If all is ok lets discuss alarms.

ekexcello:
Thank you, Ilya, 
Sorry for the delay, it seems that you found all the units and statuses yourself, so I'll just test this on the cluster (it will take a couple of days to see at least basic things) and will update you.
--

ilyam8:
@ekexcello 

see

> https://mariadb.com/kb/en/library/galera-cluster-status-variables

It is very likely there is something useful we miss

ekexcello:
Hello Ilya,

I'm testing monitoring now, everything seems to be working, the only thing is that 
Cluster Component Status (mysql_local_MariaDB_Galera.galera_cluster_status)
and
Cluster Component State (mysql_local_MariaDB_Galera.galera_cluster_state)
seems to have descriptions of '-1' and '1' states mixed.
For example, now cluster reports Component State as '-1' that is 'unknown' according to description, while in SQL CLI I see wsrep_local_state_comment:"Donor/Desynced".
Same for Component Status reporting '-1 == unknow' while cluster was in non-Primary state.
However, in perspective of cluster status monitoring, this is minor issue.

As for missing variables... well, it might be reasonable for someone to track wsrep_desync_count and %_oooe / %_oool variables, but at the moment probably I'd like not to request this - this is extra while cluster status is really required.
Thank you very much for adding these features!

ilyam8:
@ekexcello 

> non-Primary

I used values from this article

> https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html#wsrep_cluster_status

 - `Primary`
 - `Non-Primary`
 - `Disconnected`

`-1` means module failed to convert text value to int.

I fixed issues with state and status, please update and test it again

ilyam8:
About alarms:

> Also, when node leaves the cluster or joines it (thus, wsrep_cluster_size is changed) an alert should be sent, too, for example, if node left the cluster, this event is critical (while joining can be just notification).

> if node left the cluster, this event is critical

Added

> while joining can be just notification

There is no notification levels, only `warn` and `crit`. Since adding node is not `warn` or `crit` level event i didnt add alarm on node adding.

ekexcello:
Hello Ilya,

Status descriptions are now correct, checked this. But unfortunately, the alert you cretaed returned incorrect data and false positive:

mysql galera cluster size max 2m : 2 size
mysql galera cluster size : 0 size

I have had to change template to following:

```
# galera cluster size

template: mysql_galera_cluster_size_max_2m
      on: mysql.galera_cluster_size
  lookup: max -2m average
   units: size
   every: 10s
    info: max cluster size 2 minute
      to: dba

template: mysql_galera_cluster_size
      on: mysql.galera_cluster_size
  lookup: average -20s
   units: size
   every: 10s
    warn: $this > $mysql_galera_cluster_size_max_2m
    crit: $this < $mysql_galera_cluster_size_max_2m
   delay: up 20s down 5m multiplier 1.5 max 1h
    info: cluster size has changed
      to: dba

```

a) changed "calc" to "lookup". If cluster size changes, average for time periods of different length won't match and alert will be sent. However, this is not completely correct as in the message I received when one node went down, I've got 'mysql galera cluster size = 1.65 size' that looks strange considering that 'size' is measured in instances. This is better then nothing, but it would be much better to have more reliable cluster size change detection;
b) added delay of 20 seconds to exclude flapping. In fact, it shoudl send a message on second measurement with sizes not equal to previous;
c) added 'warn: $this > $mysql_galera_cluster_size_max_2m' warning. If cluster size decreases, this is a serious issue. If extra node joins the cluster, this can put one node into 'doner' state or add potential instability if number of nodes in the cluster is even, so this should be for sure recorded, but again, this is not critical as cluster remains functional when extra node joins.

ilyam8:
@ekexcello thanks!

I applied almost of your changes, please test it again

ilyam8:
current alarm is

```
template: mysql_galera_cluster_size_max_2m
      on: mysql.galera_cluster_size
  lookup: max -2m absolute
   units: size
   every: 10s
    info: max cluster size 2 minute
      to: dba

template: mysql_galera_cluster_size
      on: mysql.galera_cluster_size
    calc: $nodes
   units: size
   every: 10s
    warn: $this > $mysql_galera_cluster_size_max_2m
    crit: $this < $mysql_galera_cluster_size_max_2m
   delay: up 20s down 5m multiplier 1.5 max 1h
    info: cluster size has changed
      to: dba
```

collectors/python.d.plugin/mysql/mysql.chart.py
```diff
@@ -663,6 +734,65 @@ def userstats_chart_template(name):
...
+
+    def convert_ready(self, value):
+        # https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html#wsrep_ready
+        if value == 'OFF':
+            return 0
+        if value == 'ON':
+            return 1
+        return self.unknown_value
+
+    def convert_local_state_comment(self, value):
```
thiagoftsm: In this function you are testing values in the range [0,3], but the official documentation uses value between [1,4], considering that I do not have a galera cluster, I would like to confirm that this is not creating problems in possible reports, can you confirm this?

collectors/python.d.plugin/mysql/mysql.chart.py
```diff
@@ -663,6 +734,65 @@ def userstats_chart_template(name):
...
+            return 2
+        elif value == 'Synced':
+            return 3
+        return self.unknown_value
+
+    def convert_cluster_status(self, value):
+        # https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html#wsrep_cluster_status
+        if value == 'Primary':
+            return 0
+        elif value == 'Non-Primary' or value == 'non-Primary':
```
thiagoftsm: The official documentation does not say nothing about `non-Primary`, why are we not testing `non-primary` too?

ekexcello:
Thank you, Ilya, now it works.
The only thing that would be convenient is to change units to 'nodes' if possible:
```
template: mysql_galera_cluster_size_max_2m
      on: mysql.galera_cluster_size
  lookup: max -2m absolute
   units: nodes
   every: 10s
    info: max cluster size 2 minute
      to: dba

template: mysql_galera_cluster_size
      on: mysql.galera_cluster_size
    calc: $nodes
   units: nodes
   every: 10s
    warn: $this > $mysql_galera_cluster_size_max_2m
    crit: $this < $mysql_galera_cluster_size_max_2m
   delay: up 20s down 5m multiplier 1.5 max 1h
    info: cluster size has changed
      to: dba
```
Anyway, the alert now works as desired.

collectors/python.d.plugin/mysql/mysql.chart.py
ilyam8: > https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html#wsrep_local_state_comment

I wouldnt say it is official documentaion. There is just a table with numbers. 

Probably we need to use

> wsrep_local_state
> Internal Galera Cluster FSM state number.

Without decoding text to int.


https://github.com/codership/galera/blob/f2164433df3830e02adf319d1b0c4a8524a7e0bc/galera/src/replicator_smm_stats.cpp#L11-L24

I was googling for `galera source code` and found this

https://github.com/codership/galera/blob/f2164433df3830e02adf319d1b0c4a8524a7e0bc/galera/src/replicator_smm_stats.cpp#L11-L24

```cpp
static const char* state_str[GALERA_STAGE_MAX] =
{
    "Initialized",
    "Joining",
    "Joining: preparing for State Transfer",
    "Joining: requested State Transfer",
    "Joining: receiving State Transfer",
    "Joined",
    "Synced",
    "Donor/Desynced",
    "Joining: State Transfer request failed",
    "Joining: State Transfer failed",
    "Destroyed"
};
```

thiagoftsm: Thanks!

collectors/python.d.plugin/mysql/mysql.chart.py
ilyam8: i think it depends on implementation/version/something else

see
https://github.com/netdata/netdata/pull/6962#issuecomment-536513199

ilyam8: > Same for Component Status reporting '-1 == unknow' while cluster was in non-Primary state.

thiagoftsm: I understood now, I thought we were missing, because in the link present in the function comments, I could not find the value `non-Primary`.

collectors/python.d.plugin/mysql/mysql.chart.py
ilyam8: Probably you are right

> Internal number and the corresponding human-readable comment of the node’s state. Possible values are:

> Internal number

So it is `[1-4]`

thiagoftsm: I am glad I could help you! :)

ilyam8:
we found (thx for @vlvkobal )

https://github.com/codership/wsrep-API/blob/eab2d5d5a31672c0b7d116ef1629ff18392fd7d0/wsrep_api.h#L256-L264

```cpp
typedef enum wsrep_member_status {
    WSREP_MEMBER_UNDEFINED, //!< undefined state
    WSREP_MEMBER_JOINER,    //!< incomplete state, requested state transfer
    WSREP_MEMBER_DONOR,     //!< complete state, donates state transfer
    WSREP_MEMBER_JOINED,    //!< complete state
    WSREP_MEMBER_SYNCED,    //!< complete state, synchronized with group
    WSREP_MEMBER_ERROR,     //!< this and above is provider-specific error code
    WSREP_MEMBER_MAX
} wsrep_member_status_t;
```

@ekexcello

i think it is better  to collect

> wsrep_local_state
> Description: Internal Galera Cluster FSM state number.


instead of

> wsrep_local_state_comment¶
> Description: Human-readable explanation of the state.


In that way we dont need to convert anything.


___

Please test the PR again

ekexcello:
Hello @ilyam8 and @thiagoftsm 

I'm testing new code... It seems that conversion from numeric (this time) to human-readable status has inaccuracies this time.
For example, now cluster has following state:

```
| wsrep_local_state          | 2                                    |
| wsrep_local_state_comment  | Donor/Desynced                       |
| wsrep_cluster_status       | Primary                              |

```
while according to graph legend, `0: joining, 1: donor or desynced, 2: joined, 3: synced`

BTW, during testing, another state was found:
```
| wsrep_local_state          | 0                                    |
| wsrep_local_state_comment  | Initialized                          |
```

This was found on donor node that has initialized cluster and became donor for another node.

It seems that Galera release we use (galera-26.4.2 with MariaDB 10.4.7) has more state comments then numerical states, so `0` can be both _joining_ and _Initialized_. Unfortunately, node that is just started and joining the cluster can't respond to any queries until it becomes synced one, so state comment on joining node can't be checked directly.

As for Primary/non-Primary state, this one works correctly.
However, it came clear that alert for node entering non-primary or any other then 'Primary' state will be required. Can I ask you for this additional status alarm?

ekexcello:
Small update.
Statuses seems to be shifted, on SQL CLI status is

```
| wsrep_local_state          | 4                                    |
| wsrep_local_state_comment  | Synced                               |
```
and chart comments are present for statuses '-1 - 3' only for Cluster Component State chart

ilyam8:
@ekexcello 

ok, i will use

https://github.com/codership/galera/blob/f2164433df3830e02adf319d1b0c4a8524a7e0bc/galera/src/replicator_smm_stats.cpp#L11-L24

```cpp
static const char* state_str[GALERA_STAGE_MAX] =
{
    "Initialized",
    "Joining",
    "Joining: preparing for State Transfer",
    "Joining: requested State Transfer",
    "Joining: receiving State Transfer",
    "Joined",
    "Synced",
    "Donor/Desynced",
    "Joining: State Transfer request failed",
    "Joining: State Transfer failed",
    "Destroyed"
};
```

I see that array is used for `locale_state_comment`

https://github.com/codership/galera/blob/f2164433df3830e02adf319d1b0c4a8524a7e0bc/galera/src/replicator_smm_stats.cpp#L45-L73

```cpp
static const char* state2stats_str(galera::ReplicatorSMM::State    state,
                                   galera::ReplicatorSMM::SstState sst_state)
{
    using galera::ReplicatorSMM;


    switch (state)
    {
    case galera::ReplicatorSMM::S_DESTROYED :
        return state_str[10];
    case galera::ReplicatorSMM::S_CLOSED :
    case galera::ReplicatorSMM::S_CLOSING:
    case galera::ReplicatorSMM::S_CONNECTED:
    {
        if (sst_state == ReplicatorSMM::SST_REQ_FAILED)  return state_str[8];
        else if (sst_state == ReplicatorSMM::SST_FAILED) return state_str[9];
        else                                             return state_str[0];
    }
    case galera::ReplicatorSMM::S_JOINING:
    {
        if (sst_state == ReplicatorSMM::SST_WAIT) return state_str[4];
        else                                      return state_str[1];
    }
    case galera::ReplicatorSMM::S_JOINED : return state_str[5];
    case galera::ReplicatorSMM::S_SYNCED : return state_str[6];
    case galera::ReplicatorSMM::S_DONOR  : return state_str[7];
    }


    gu_throw_fatal << "invalid state " << state;
}
```

ekexcello:
Sorry for misleading you, it seems that I was running old binary. After re-building and restarting Netdata, statuses seems to be correct:

0: undefined, 1: joining, 2: donor/desynced, 3: joined, 4: synced.

Also, I tried to define alarms for cluster_status and cluster_state (that is in fact node state), looks like:

```
template: mysql_galera_cluster_status
      on: mysql.galera_cluster_status
  lookup: average -10s
   every: 10s
    crit: $this != 0
   delay: up 30s down 5m
    info: node and cluster status (-1: unknown, 0: primary/quorum present, 1: non-primary/quorum lost, 2: disconnected).
      to: dba

template: mysql_galera_cluster_state
      on: mysql.galera_cluster_state
  lookup: average -10s
   every: 10s
    warn: $this < 4
    crit: $this < 2
   delay: up 30s down 5m
    info: node state (0: undefined, 1: joining, 2: donor/desynced, 3: joined, 4: synced)
      to: dba

```

ilyam8:
@ekexcello `disconnected` is default state in standalone install.

I suggest to merge this PR as it is. You will play with different alarms and we will add them later. What do you think?

ilyam8:
standalone install has no `wsrep_local_state_comment` so it is ok to add node state alarm

ekexcello:
@ilyam8 
I agree that it will be good to merge as is, as for alarms - I'll probably will be happy with current set. Anyway, if metric is collected, alarms can be defined if needed. Thank you once again for this!

ilyam8:
@ekexcello 

i cant add cluster state status alarm because of https://github.com/netdata/netdata/issues/6980

We need to check if `mysql_galera_cluster_state` alarm exists and only if it exists we check `galera_status != 0` (otherwise the alarm will be false positive for standalone installs)

ekexcello:
No problem, @ilyam8, as I mentioned, for the metrics that are collected, it's not a big deal to create custom alarm definition, so I can still use the alarm for Galera installation I have as custom one. This should not block merging PRs, I hope.
Thank you anyway!

ilyam8:
@cakrit @cosmix @joelhans please

web/gui/dashboard_info.js
```diff
@@ -1260,6 +1260,39 @@ netdataDashboard.context = {
...
+
+    'mysql.galera_connected': {
+        info: '<code>0</code> means that the node has not yet connected to any of the cluster components. ' +
+            'This may be due to misconfiguration.'
+    },
+
+    'mysql.open_transactions': {
+        info: 'The number of locally running transactions which have been registered inside the wsrep provider. ' +
+            'This means transactions which have made operations which have caused write set population to happen. ' +
+            'Transactions which are read only are not counted.'
```
joelhans: These sentences are a little confusing. I've tried to make the sentences active instead of passive, but I'm not sure if it's enough. Maybe there's a doc you can point me to so I can better understand how to explain these?

```
        info: 'The number of locally running transactions registered by the wsrep (write-set replication) provider. ' +
            'This number includes any transactions with operations that trigger write-set population. ' +
            'Netdata's MySQL module does not count read-only transactions.'
```

joelhans: Hooray for the new multi-line review comments!

ilyam8: > These sentences are a little confusing.

THis is ctrl-c ctrl-v from

> https://galeracluster.com/library/documentation/galera-status-variables.html

See `wsrep_open_transactions` var description

joelhans: Well, with that in mind, we can just leave the text as it is. I feel like the risk I edit the sentences and make them wrong is greater than having a weird sentence or two in the dashboard.

Issue #6541, opened 2019-07-25
elasticsearch plugin should support indices metrics


Wing924:
##### Feature idea summary

Elasticsearch should optionally export indices metrics, which can be get from `/_cat/indices`.

output samples:
```
curl 'localhost:9200/_cat/indices?format=json&pretty'
[
  {
    "pri.store.size": "650b",
    "health": "yellow",
    "status": "open",
    "index": "twitter",
    "pri": "5",
    "rep": "1",
    "docs.count": "10",
    "docs.deleted": "3",
    "store.size": "650b"
  }
]
```

should export:
```
# gauge
elasticsearch_index_docs_count{dimension="twitter"} 10
# gauge
elasticsearch_index_store_size{dimension="twitter"} 650
# gauge
elasticsearch_index_replica{dimension="twitter"} 5
# gauge red=0, yellow=1, green=2
elasticsearch_index_health{dimension="twitter"} 1
```

##### Expected behavior

have elasticsearch index chart
* don't need 1 chart per index
* a chart with index docs count
* a chart with index store size
* a chart with index replica
* a chart with index health

Maybe it's better to implement it by Go.

PR #6965, opened 2019-09-27
elasticsearch: collect metrics from _cat/indices

ilyam8:
##### Summary

Fixes: #6541

This PR adds per index metrics:
 - docs count
 - store size
 - number of replicas
 - health status

Modules gathers these metrics from `_cat/indices` endpoint. Per index metrics is optional and disabled by default.

##### Component Name

[`/collectors/python.d.plugin/elasticsearch`](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/elasticsearch/elasticsearch.chart.py)

### DIFF
--- a/collectors/python.d.plugin/elasticsearch/README.md
+++ b/collectors/python.d.plugin/elasticsearch/README.md
@@ -1,6 +1,6 @@
 # elasticsearch
 
-This module monitors Elasticsearch performance and health metrics.
+This module monitors `Elasticsearch` performance and health metrics.
 
 It produces:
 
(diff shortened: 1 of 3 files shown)


ilyam8:
@Wing924 could you test the PR?

> elasticsearch_index_replica{dimension="twitter"} 5
# gauge red=0, yellow=1, green=2

I did
 - green: 0
 - yellow: 1
 - red: 2


Reason - netdata hides zeros by default, it would be wrong to have red status hidden.

Wing924:
@ilyam8 
スクリーンショット 2019-09-30 22 03 50
This module become very slow after apply this change.
I think query every seconds is too hard for Elasticsearch.

ilyam8:
@Wing924 `elasticsearch` module does queries every 5 sec by default.

collectors/python.d.plugin/elasticsearch/README.md
```diff
@@ -1,6 +1,6 @@
 # elasticsearch
 
-This module monitors Elasticsearch performance and health metrics.
+This module monitors `Elasticsearch` performance and health metrics.
```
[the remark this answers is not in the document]
ilyam8: yes, good idea

collectors/python.d.plugin/elasticsearch/elasticsearch.chart.py
```diff
@@ -550,88 +616,131 @@ def _get_data(self):
...
+        if not parsed:
             return queue.put(dict())
 
-        data = self.json_reply(raw_data)
+        data = fetch_data(raw_data=parsed, metrics=HEALTH_STATS)
+        dummy = {
+            'status_green': 0,
+            'status_red': 0,
+            'status_yellow': 0,
+            'status_foo1': 0,
```
Wing924: debug code?

ilyam8: kind of 😄

ilyam8:
> docs count unit is wrong?

Looks ok, try ti disable `Scale Units` option (Settings->Locale)

PR #6970, opened 2019-09-28
Fix coverity erro (CID 349552) double lock

### LINKED ISSUE #6969: Deadlock defect introduced in database
##### Bug report summary

A new defect was detected by Coverity Scan:
```
** CID 349552:  Program hangs  (ORDER_REVERSAL)
/database/rrddim.c: 404 in rrddim_add_custom()


*** CID 349552:  Program hangs  (ORDER_REVERSAL)
/database/rrddim.c: 404 in rrddim_add_custom()
398         }
399     
400         if(unlikely(rrddim_index_add(st, rd) != rd))
401             error("RRDDIM: INTERNAL ERROR: attempt to index duplicate dimension '%s' on chart '%s'", rd->id, st->id);
402     
403         if(host->alarms_with_foreach || host->alarms_template_with_foreach) {
>>>     CID 349552:  Program hangs  (ORDER_REVERSAL)
>>>     Calling "__netdata_rwlock_wrlock" acquires lock "rrdhost.rrdhost_rwlock" while holding lock "rrdset.rrdset_rwlock" (count: 1 / 14).
404             rrdhost_wrlock(host);
405             rrdcalc_link_to_rrddim(rd, st, host);
406     
407             rrdhost_unlock(host);
408         }
409         rrdset_unlock(st);
```
The offending commit is:
```
commit e3471fa5727bcf286dd3b52ec0cdecd8fdf7067e
Author: thiagoftsm <thiagoftsm@gmail.com>
Date:   Fri Sep 27 12:24:54 2019 +0000

    Create a template for all dimensions (#6560)
```
introduced by #6560 
##### Component Name
database

thiagoftsm:
##### Summary
Fixes #6969 

We had a deadlock inside the database after to merge template to dimension, this commit fixes this error.
##### Component Name
Health

### DIFF
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -400,7 +400,6 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
     if(unlikely(rrddim_index_add(st, rd) != rd))
         error("RRDDIM: INTERNAL ERROR: attempt to index duplicate dimension '%s' on chart '%s'", rd->id, st->id);
 
-    rrdset_unlock(st);
     if(host->alarms_with_foreach || host->alarms_template_with_foreach) {
         int count = 0;
         int hostlocked;
@@ -430,6 +429,7 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
             error("Cannot lock host to create an alarm for the dimension.");
         }
     }
+    rrdset_unlock(st);
 
     return(rd);
 }
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -403,30 +403,26 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
     if(host->alarms_with_foreach || host->alarms_template_with_foreach) {
         int count = 0;
         int hostlocked;
-        trylockagain:
-        hostlocked = pthread_rwlock_trywrlock(&host->rrdhost_rwlock);
-        if(!hostlocked)
+        for (count = 0 ; count < 5 ; count++)
         {
-            rrdcalc_link_to_rrddim(rd, st, host);
-            rrdhost_unlock(host);
-        }
-        else if (hostlocked == EBUSY)
-        {
-            if (count <= 5)
+            hostlocked = netdata_rwlock_trywrlock(&host->rrdhost_rwlock);
+            if(!hostlocked)
             {
-                usleep(200000);
-                count++;
-                goto trylockagain;
+                rrdcalc_link_to_rrddim(rd, st, host);
+                rrdhost_unlock(host);
+                break;
             }
-            else
+            else if (hostlocked != EBUSY)
             {
-                error("Failed to create an alarm for dimension %s of chart %s 5 times. Skipping alarm."
-                      , rd->name, st->name);
+                error("Cannot lock host to create an alarm for the dimension.");
             }
+            usleep(200000);
         }
-        else
+
+        if (count == 5)
         {
-            error("Cannot lock host to create an alarm for the dimension.");
+            error("Failed to create an alarm for dimension %s of chart %s 5 times. Skipping alarm."
+            , rd->name, st->name);
         }
     }
     rrdset_unlock(st);
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -187,9 +187,11 @@ void rrdcalc_link_to_rrddim(RRDDIM *rd, RRDSET *st, RRDHOST *host) {
 
 RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collected_number multiplier, collected_number divisor, RRD_ALGORITHM algorithm, RRD_MEMORY_MODE memory_mode) {
     RRDHOST *host = st->rrdhost;
+    /*
     if(host->alarms_with_foreach || host->alarms_template_with_foreach) {
         rrdhost_wrlock(host);
     }
+     */
     rrdset_wrlock(st);
 
     rrdset_flag_set(st, RRDSET_FLAG_SYNC_CLOCK);
@@ -405,9 +407,11 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
 
     rrdset_unlock(st);
     if(host->alarms_with_foreach || host->alarms_template_with_foreach) {
+        rrdhost_wrlock(host);
         rrdcalc_link_to_rrddim(rd, st, host);
         rrdhost_unlock(host);
     }
+
     return(rd);
 }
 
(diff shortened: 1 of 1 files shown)


mfundul:
I don't think this is enough, since you are holding only the host lock and you allow the dimension to be deleted during iteration.

thiagoftsm:
I push another commit after to read your newest text in the ISSUE related to this PR.

mfundul:
The only question remaining to be answered is whether a caller of `rrddim_add_custom()` already hold the host lock, which would result in a deadlock. If that is not the case, then this should suffice.

database/rrddim.c
```diff
@@ -204,7 +208,6 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
         return rd;
```
mfundul: You need to handle error cases such as this one. You could add something like:
```
if (host_locked)
    rrdhost_unlock(host);
```

thiagoftsm:
> The only question remaining to be answered is whether a caller of `rrddim_add_custom()` already holds the host lock, which would result in a deadlock. If that is not the case, then this should suffice.

Without to use function `rrdhost_lock()` I have the following error message:

``
`2019-09-28 16:48:09: netdata ERROR : PLUGIN[proc] : RW_LOCK: failed to obtain write lock (code 35)
```

and after this Netdata exit with failure, so it is necessary to call this function.

File: database/rrddim.c (hunk 1 above)
Comment (thiagoftsm) [reply database/rrddim.c]: I was verifying Netdata code here, the function `rrdhost_wrlock` and `rrdhost_unlock` are definitions to other internal functions, `netdata_rwlock_wrlock_debug` and `netdata_rwlock_unlock_debug`, these other two functions call pthread functions to try lock and try unlock, these functions already have error messages inside it.
I tried to use pthread_rwlock_trywrlock to check errors and I received a lot  `RW_LOCK: failed to obtain write lock (code 35)` and the alarms where not created, os I returned for the current code we have here, because it is working fine.

**Comment by thiagoftsm** [comment]
I tested all options that you wrote in the original issue and here, and I returned for the first solution that you gave us, because it is the unique that does not give us ` PLUGINSD[apps] : RW_LOCK: failed to obtain write lock (code 35)` .
For sure we do not need to lock the chart, because the last part of the code inside the function only create alarms, so we are only reading chart names, the necessity to lock host is related to the alarms associated to it.

**Comment by mfundul** [comment]
> 
> 
> I tested all options that you wrote in the original issue and here, and I returned for the first solution that you gave us, because it is the unique that does not give us ` PLUGINSD[apps] : RW_LOCK: failed to obtain write lock (code 35)` .
> For sure we do not need to lock the chart, because the last part of the code inside the function only create alarms, so we are only reading chart names, the necessity to lock host is related to the alarms associated to it.

This will not work because the code is referencing the chart and dimension by pointers `st` and `rd` respectively from:
```
 rrdcalc_link_to_rrddim(rd, st, host);
```
and if for example a thread is deleting a dimension (`rd`) then netdata will crash.

**Comment by mfundul** [comment]
That is a theoretical problem of course, since it's hard to think of a practical case for this, since the dimension has just been created. But still, it's a potential bug, and if there's no chance for it now, there may be in the future.

**Comment by thiagoftsm** [comment]
> That is a theoretical problem of course, since it's hard to think of a practical case for this, since the dimension has just been created. But still, it's a potential bug, and if there's no chance for it now, there may be in the future.

You are correct, I thought the situation again and I found a way to avoid the error reported and do not raise errors with our current code in this last commit, please, verify this  solution now, I think we can accept 5 tries to lock the host to create the alarms, mainly considering that this did not happen in my tests and probably during the review, because nobody reported errors related to this. But, I have a doubt related to the time I am waiting for to try lock again.

File: database/rrddim.c [hunk 2]
```diff
@@ -400,13 +400,37 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
...
-        rrdcalc_link_to_rrddim(rd, st, host);
-
-        rrdhost_unlock(host);
+        int count = 0;
+        int hostlocked;
+        trylockagain:
+        hostlocked = pthread_rwlock_trywrlock(&host->rrdhost_rwlock);
+        if(!hostlocked)
+        {
+            rrdcalc_link_to_rrddim(rd, st, host);
```
mfundul: Here you are reusing the `rd` dimension pointer without holding the lock for chart `st`. If `rd` is deleted in the meantime it would crash. Moreover, even if you retook the chart lock, still you would be reusing the `rd` pointer which could have been freed during the period when you did not actually hold the chart lock.

We need to think this through.

thiagoftsm: We are creating ```rd``` in this function, so we cannot delete it at this point. What we can do is to move back `rrdset_unlock` this will avoid we delete the chart. Considering that the write lock for chart is different of write lock for host, I do not think the problem was there. Right now I really believe that the error reported from Coverity was related I try to lock something that was possibly locked. I moved the function back and I did not notice any error reported.

mfundul: Well, since you moved the lock again `rd` is safe again but you still keep the wrong lock order only now you're try-locking instead of locking. So, in the previous cases of deadlock, now you're delaying and failing, which is probably why coverity does not write this down. But still, there's probably cases where we delay and fail during try-locking. Is that a problem?

thiagoftsm: I could not detect any error while Netdata was trying to lock to create the alarm, but I had when I tried to do some changes in the previous commits. 
About your question, I thought about it and I decided to try 5 times and finally raise an error. What can happen is we do not have an alarm created, but I used Netdata simple pattern to create 11 dimensions in a chart and I did not have any error when I tried to lock, so I think this is not a problem, at least it was not a problem with the charts I tried.
After to read your message I decided to repeat a test that I did in your PR, I changed the example.python to have 160 dimensions and I created an alarm using `foreach` to create an alarm for each dimensions and again I did not have any error.

mfundul: Make sure you fix the grammar of the error message:
```
After to try creating an alarm for dimension %s on chart %s 5 times without success, I am moving in front
```
to something like:
```
Failed to create an alarm for dimension %s of chart %s 5 times. Skipping alarm.
```

thiagoftsm: Thank you, I changed the previous message for your suggestion, I liked it so I decided to keep it .

mfundul: You should use `netdata_rwlock_trywrlock()` instead of using pthread locks directly. Also, please reduce the number of lines according to our latest code style specifications and try to use a `for()` loop instead of the `goto` if possible here.

thiagoftsm: I executed the changes asked and Netdata continues working without raise errors and with all alarms created.

ilyam8: > Also, please reduce the number of lines according to our latest code style specifications

@thiagoftsm clion can autoformat code, just use it, see `code` section

thiagoftsm: I will check this, thanks @ilyam8!

mfundul: I'm waiting for the code style change to approve this.

thiagoftsm: The last commit fixes the format. Thanks!

PR #6977, opened 2019-10-01
Make dbengine the default memory mode

### LINKED ISSUE #6767: Make dbengine the default memory mode
##### Feature idea summary
Make sure `memory mode = dbengine` is the default. For that to be the case we need to finish:
- Stress testing
- Instrumentation & Characterization
- Benchmarking

Tasks:
- [x] Combine and parallelize dbengine validation unittest with dataset generator into a multithreaded dbengine benchmark that validates data integrity while constantly generating new data.
- [ ] Run k6 load generator and get API performance numbers.
- [ ] Combine k6 load testing with dbengine dataset generation to get API query performance numbers with heavy data collection interference.
- [x] List the OSs where we know it is supported OOB
    - See whether `kickstart.sh` enables `dbengine` on OSs without manually installing repositories below:
    - [x] Ubuntu 18.04
    - [ ] CentOS 6
    - [ ] CentOS 7
    - [ ] Debian 8
    - [x] Debian 9
    - [x] Debian 10
    - [ ] Manjaro 18
    - [x] Fedora 30
    - [x] openSUSE 15
    - [ ] FreeBSD 12 (not supported by `kickstart.sh`)
- [x] Make the default memory mode `dbengine` when the support is compiled in the netdata binary

mfundul:
##### Summary
Make dbengine the default memory mode.
Combine and parallelize dbengine validation unittest with dataset generator into a multithreaded dbengine benchmark that validates data integrity while constantly generating new data.
##### Component Name
database/engine
##### Additional Information
Fixes #6767

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -307,6 +307,7 @@ int help(int exitcode) {
             "  -W debug_flags=N         Set runtime tracing to debug.log.\n\n"
             "  -W unittest              Run internal unittests and exit.\n\n"
             "  -W createdataset=N       Create a DB engine dataset of N seconds and exit.\n\n"
+            "  -W stresstest=N          Run a DB engine stress test for N seconds and exit.\n\n"
             "  -W set section option value\n"
             "                           set netdata.conf option from the command line.\n\n"
             "  -W simple-pattern pattern string\n"
@@ -887,6 +888,7 @@ int main(int argc, char **argv) {
                         char* stacksize_string = "stacksize=";
                         char* debug_flags_string = "debug_flags=";
                         char* createdataset_string = "createdataset=";
+                        char* stresstest_string = "stresstest=";
 
                         if(strcmp(optarg, "unittest") == 0) {
                             if(unit_test_buffer()) return 1;
@@ -910,6 +912,14 @@ int main(int argc, char **argv) {
 #ifdef ENABLE_DBENGINE
                             unsigned history_seconds = (unsigned )strtoull(optarg, NULL, 0);
                             generate_dbengine_dataset(history_seconds);
+#endif
+                            return 0;
+                        }
+                        else if(strncmp(optarg, stresstest_string, strlen(stresstest_string)) == 0) {
+                            optarg += strlen(stresstest_string);
+#ifdef ENABLE_DBENGINE
+                            unsigned test_duration_sec = (unsigned )strtoull(optarg, NULL, 0);
+                            dbengine_stress_test(test_duration_sec);
 #endif
                             return 0;
                         }
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -306,8 +306,13 @@ int help(int exitcode) {
             "  -W stacksize=N           Set the stacksize (in bytes).\n\n"
             "  -W debug_flags=N         Set runtime tracing to debug.log.\n\n"
             "  -W unittest              Run internal unittests and exit.\n\n"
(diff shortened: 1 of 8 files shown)


daemon/main.c
```diff
@@ -905,14 +912,33 @@ int main(int argc, char **argv) {
                             fprintf(stderr, "\n\nALL TESTS PASSED\n\n");
                             return 0;
                         }
+#ifdef ENABLE_DBENGINE
                         else if(strncmp(optarg, createdataset_string, strlen(createdataset_string)) == 0) {
                             optarg += strlen(createdataset_string);
-#ifdef ENABLE_DBENGINE
-                            unsigned history_seconds = (unsigned )strtoull(optarg, NULL, 0);
+                            unsigned history_seconds = strtoul(optarg, NULL, 0);
```
thiagoftsm: `unsigned` type for the majority of the compilers is `unsigned integer`, but the function ``` strtoul``` returns `unsigned long' that is bigger than `unsigned` for 64 bits host, so I think it is better to keep the cast here.

mfundul: Done.

daemon/unit_test.c
```diff
@@ -2021,25 +2033,259 @@ void generate_dbengine_dataset(unsigned history_seconds)
...
+    uint8_t same;
+    time_t time_now, time_retrieved;
+    collected_number generatedv;
+    calculated_number value, expected;
+    storage_number n;
+    struct rrddim_query_handle handle;
+
+    do {
+        // pick a chart and dimension
+        i = random() % DSET_CHARTS;
```
thiagoftsm: ```random()``` returns `long int`, to avoid warnings, I think it is good to cast here.

daemon/unit_test.c
```diff
@@ -2021,25 +2033,259 @@ void generate_dbengine_dataset(unsigned history_seconds)
...
+        }
+
+        rd->state->query_ops.init(rd, &handle, time_after, time_before);
+        ++thread_info->queries_nr;
+        for (time_now = time_after ; time_now <= time_before ; time_now += update_every) {
+            generatedv = generate_dbengine_chart_value(i, j, time_now);
+            expected = unpack_storage_number(pack_storage_number((calculated_number) generatedv, SN_EXISTS));
+
+            if (unlikely(rd->state->query_ops.is_finished(&handle))) {
+                fprintf(stderr, "    DB-engine stresstest %s/%s: at %lu secs, expecting value "
```
thiagoftsm: why are you not using ```error()``` function here?

daemon/unit_test.c
mfundul: I haven't seen any warnings from the compiler or static analysis. Until we see an actual warning I think we shouldn't pollute the code with unnecessary casts.

daemon/unit_test.c
mfundul: The binary is running as standalone, there is no service running or logging being set-up. It's running like this `./netdata -W stresstest=10`

thiagoftsm: Thanks.

daemon/unit_test.c
thiagoftsm: I will compile this PR in a 32 bits computer to confirm that the warning is not present.

thiagoftsm: I tested in 32 bits computer and I also compiled with clang, there was not any warning for both compilation.

PR #6984, opened 2019-10-01
Netdata not returning correct value for unknow variables

### LINKED ISSUE #6980: health: nonexistent variable in expression always evaluates to zero
##### Bug report summary

SSIA

[from the wiki](https://docs.netdata.cloud/health/#expressions)

> Expressions can have variables. Variables start with $. Check below for more information.

> There are two special values you can use:
>   - nan, for example $this != nan will check if the variable this is available. A variable can be nan if the database lookup failed. All calculations (i.e. addition, multiplication, etc) with a nan result in a nan.

##### OS / Environment

debian10

netdata v1.17.1-127-g2cb7c2fd

##### Component Name

health

##### Steps To Reproduce

 - use nonexistent variable in health expression

##### Expected behavior

nonexistent variable in expression always evaluates to `nan`

thiagoftsm:
##### Summary
Fixes #6980 

Netdata was returning a wrong value according our documentation when the alarms had a unknown variable name, this PR fixes this existent problem.
##### Component Name
Health


ilyam8:
I tested this PR, @thiagoftsm there are still some problems.

alarm (`$mysql_galera_cluster_state` doesnt exist, i.e == `nan`)

```
template: mysql_galera_cluster_status
      on: mysql.galera_cluster_status
    calc: $wsrep_cluster_status
   every: 10s
    crit: $mysql_galera_cluster_state != nan AND $this != 0
   delay: up 30s down 5m multiplier 1.5 max 1h
    info: node and cluster status (-1: unknown, 0: primary/quorum present, 1: non-primary/quorum lost, 2: disconnected)
      to: dba
```

See
Screenshot 2019-10-02 at 11 27 56



The problem is gray value, it should be green.  If i use existing var in `crit` it works correctly.

thiagoftsm:
I will work on this today, I already solved the problem to have `zero` when the variable is not set, I will work in the raise alarm now when it is possible. In my initial tests I used a `calc` stream that always failed. Thanks!

stelfrag:
@thiagoftsm 
using undefined variables, correctly returns **nan**  and can now trigger an alarm. 

I used a simple config to raise a warning for cpu temp alarm (with very low temp setting of 23) :
```
warn: $fake = nan and $this > (($status >= $WARNING) ? (72) : (23))
```
before the patch the alarm would not be triggered

After the patch the alarm is raised successfully because $fake is set to **nan**

thiagoftsm:
Thank you @stelfrag !

ilyam8:
i did same tests as in https://github.com/netdata/netdata/pull/6984#issuecomment-537392008 and now it works correctly

PR #6992, opened 2019-10-04
Remove warning from Coverity

thiagoftsm:
##### Summary
Coverity reported us that on CID 348642 that we have an unnecessary check on our code, this PR is moving up the code inside ```if``` to remove the issue reported.
##### Component Name
Database
Health

### DIFF
--- a/libnetdata/eval/eval.c
+++ b/libnetdata/eval/eval.c
@@ -1077,7 +1077,6 @@ int expression_evaluate(EVAL_EXPRESSION *expression) {
         // although there is an unknown variable
         // the expression was evaluated successfully
         expression->error = EVAL_ERROR_OK;
-        expression->result = NAN;
     }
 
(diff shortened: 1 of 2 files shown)


libnetdata/eval/eval.c
```diff
@@ -1077,6 +1077,7 @@ int expression_evaluate(EVAL_EXPRESSION *expression) {
         // although there is an unknown variable
         // the expression was evaluated successfully
         expression->error = EVAL_ERROR_OK;
+        expression->result = NAN;
```
ilyam8: This change is from #6984

I think it shouldnt be there

thiagoftsm: Yesterday I found another place to put this result, but I did not have time to test the impact of it in the current results, I will do this during the weekend.
Thanks!

PR #6993, opened 2019-10-04
Tutorials to support v1.18 features

joelhans:
##### Summary

I've written tutorials on HDFS/Zookeeper monitoring and the new dimension templates to support marketing around the upcoming v1.18 release. Would definitely appreciate feedback on these, especially @ilyam8 for the HDFS/Zookeeper tutorial, and @thiagoftsm for the dimension templates one.

I also tweaked a few other docs to clarify that dbengine is now the default database.

The goal is to have these fixed, approved, and merged by 10/10.

##### Component Name

docs

### DIFF
--- a/docs/tutorials/dimension-templates.md
+++ b/docs/tutorials/dimension-templates.md
@@ -1,0 +1,159 @@
+# Use dimension templates to mass-create alarms
+
+Your ability to monitor the health of your systems and applications relies on your ability to create and maintain
+the best set of alarms for your particular needs.
+
+In v1.18 of Netdata, we introduced **dimension templates** for alarms, which simplifies the process of writing [alarm
+entities](../../health/README.md#entities-in-the-health-files) for charts with many dimensions.
+
+Dimension templates can condense many individual entities into one—no more copy-pasting one entity and changing the
+`alarm`/`template` and `lookup` lines for each dimension you'd like to monitor.
+
+They are, however, an advanced health monitoring feature. For more basic instructions on creating your first alarm,
+check out our [health monitioring documentation](../../health/), which also includes
+[examples](../../health/README.md#examples).
+
+> For more information about dimension templates, including examples of before/after examples, see our [v1.18
+> announcement post](https://blog.netdata.cloud/posts/release-1.18/).
+
+## The fundamentals of `foreach`
+
+Our dimension templates updates brings a new `foreach` parameter to the existing [`lookup`
+line](../../health/README.md#alarm-line-lookup). This is where the magic happens.
+
+You use the `foreach` parameter to specify which dimensions you want to monitor with this single alarm. You can separate
+them with a comma (`,`) or a pipe (`|`), or you can use a wildcard (`*`) to say that you want an alarm for every
+dimension the chart has to offer.
+
+The `foreach` parameter _has_ to be the last parameter in your `lookup` line, and if you have both `of` and `foreach` in
+the same `lookup` line, Netdata will ignore the `of` parameter and use `foreach` instead.
+
+Let's get into some examples so you can see how the new parameter works.
+
+> ⚠️ The following entities are examples to showcase the functionality and syntax of dimension templates. They are not
+> meant to be run as-is in production systems.
+
+## Condensing entities with `foreach`
+
+Let's say you want to monitor the `system`, `user`, and `nice` dimensions in your system's overall CPU utilization.
+Before dimension templates, you would need the following three entities:
+
+```yaml
+ alarm: cpu_system
+    on: system.cpu
+lookup: average -3s percentage of system
+ every: 10s
+  warn: $this > 50
+  crit: $this > 80
+
+ alarm: cpu_user
+    on: system.cpu
+lookup: average -3s percentage of user
+ every: 10s
+  warn: $this > 50
+  crit: $this > 80
+
+ alarm: cpu_nice
+    on: system.cpu
+lookup: average -3s percentage of nice
+ every: 10s
+  warn: $this > 50
+  crit: $this > 80
+```
+
+With dimension templates, you can condense these into a single alarm. Take note of the `alarm` and `lookup` lines.
+
+```yaml
+ alarm: cpu_template
+    on: system.cpu
+lookup: average -3s percentage foreach system,user,nice
+ every: 10s
+  warn: $this > 50
+  crit: $this > 80
+```
+
+The `alarm` line specifies the naming scheme Netdata will use. You can use whatever naming scheme you'd like, with `.`
+and `_` being the only allowed symbols.
+
+The `lookup` line has changed from `of` to `foreach`, and we're now passing three dimensions.
+
+In this example, Netdata will create three alarms with the names `cpu_template_system`, `cpu_template_user`, and
+`cpu_template_nice`. Every 10 seconds, each alarm will use the same database query to calculate the average CPU usage
+for the `system`, `user`, and `nice` dimensions over the last 3 seconds and send out alarms if necessary.
+
+Let's look at some other examples of how `foreach` works so you can best apply it in your configurations.
+
+### Using the wildcard in `foreach`
+
+In the last example, we used `foreach system,user,nice` to create three distinct alarms using dimension templates. But
+what if you want to quickly create alarms for _all_ the dimensions of a given chart? 
+
+Use a wildcard `*`!
+
+Instead of monitoring system CPU usage, let's monitor per-application CPU usage using the `apps.cpu` chart. The wildcard
+tells Netdata to create a separate alarm for _every_ process on your system:
+
+```yaml
+ alarm: app_cpu
+    on: apps.cpu
+lookup: average -3s percentage foreach *
+ every: 10s
+  warn: $this > 50
+  crit: $this > 80
+```
+
+This entity will now create alarms for every dimension in the `apps.cpu` chart. Given that most `apps.cpu` charts have
+10 or more dimensions, using the wildcard ensures you catch every CPU-hogging process.
+
+## Using `foreach` with alarm templates
+
+Dimension templates also work with [alarm templates](../../health/README.md#entities-in-the-health-files). Alarm
+templates help you create alarms for all the charts with a given context—for example, all the cores of your system's
+CPU.
+
+By combining the two, you can create dozens of individual alarms with a single template entity. Here's how you would
+create alarms for the `system`, `user`, and `nice` dimensions for every chart in the `cpu.cpu` context—or, in other
+words, every CPU core.
+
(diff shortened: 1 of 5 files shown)


docs/tutorials/monitor-hadoop-cluster.md
```diff
@@ -0,0 +1,179 @@
...
+
+At the bottom of the file, you will see two example jobs, both of which are commented out:
+
+```yaml
+# [ JOBS ]
+#jobs:
+#  - name: master
+#    url: http://127.0.0.1:9870/jmx
+#
+#  - name: slave
```
ilyam8: lets change master/slave to namenode/datanode

docs/tutorials/monitor-hadoop-cluster.md
```diff
@@ -0,0 +1,179 @@
...
+The result for a simple HDFS setup, running entirely on `localhost` and without certificate authentication, might look
+like this:
+
+```yaml
+# [ JOBS ]
+jobs:
+  - name: master
+    url: http://127.0.0.1:9870/jmx
+
+  - name: slave
```
ilyam8: master/slave

docs/tutorials/monitor-hadoop-cluster.md
```diff
@@ -0,0 +1,179 @@
...
+
+Finally, restart Netdata.
+
+```sh
+sudo service restart netdata
+```
+
+Upon restart, Netdata should recognize your HDFS/Zookeeper servers, enable the HDFS and Zookeeper modules, and begin
+showing real-time metrics for both in your Netdata dashboard. 🎉
+
```
ilyam8: we could mention default alarms for hdfs/zookeeper

 - [hdfs alarms](https://github.com/netdata/netdata/blob/master/health/health.d/hdfs.conf)
 - [zookeeper](https://github.com/netdata/netdata/blob/master/health/health.d/zookeeper.conf)

docs/tutorials/dimension-templates.md
```diff
@@ -0,0 +1,168 @@
...
+![Three new alarms created from the dimension template](https://user-images.githubusercontent.com/1153921/66218994-29523800-e67f-11e9-9bcb-9bca23e2c554.png)
+
+Let's look at some other examples of how `foreach` works so you can best apply it in your configurations.
+
+### Using the wildcard in `foreach`
+
+In the last example, we used `foreach system,user,nice` to create three distinct alarms using dimension templates. But
+what if you want to quickly create alarms for _all_ the dimensions of a given chart? 
+
+Use a wildcard `*`!
```
ilyam8: i suggest to add a link to simple pattern syntax

> https://docs.netdata.cloud/libnetdata/simple_pattern/

ilyam8:
Apart from hdfs/zookeepr we have 2 more new collectors:
 - [gearman](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/gearman)
 - [vcsa](https://github.com/netdata/go.d.plugin/tree/master/modules/vcsa)

just saying, i dont know if we need to mention it

docs/tutorials/dimension-templates.md
```diff
@@ -0,0 +1,168 @@
...
+> For more information about dimension templates, including examples of before/after examples, see our [v1.18
+> announcement post](https://blog.netdata.cloud/posts/release-1.18/).
+
+## The fundamentals of `foreach`
+
+Our dimension templates update creates a new `foreach` parameter to the existing [`lookup`
+line](../../health/README.md#alarm-line-lookup). This is where the magic happens.
+
+You use the `foreach` parameter to specify which dimensions you want to monitor with this single alarm. You can separate
+them with a comma (`,`) or a pipe (`|`), or you can use a wildcard (`*`) to say that you want an alarm for every
```
thiagoftsm: Instead to say that the person can use "wildcard", I would suggest we say that the user can use "Netdata simple pattern" and link to our documentation.

docs/tutorials/dimension-templates.md
```diff
@@ -0,0 +1,168 @@
...
+for the `system`, `user`, and `nice` dimensions over the last 3 seconds and send out alarms if necessary.
+
+You can find these three alarms active by clicking on the **Alarms** button in the top navigation, and then clicking on
+the **All** tab and scrolling to the **system - cpu** collapsible section.
+
+![Three new alarms created from the dimension template](https://user-images.githubusercontent.com/1153921/66218994-29523800-e67f-11e9-9bcb-9bca23e2c554.png)
+
+Let's look at some other examples of how `foreach` works so you can best apply it in your configurations.
+
+### Using the wildcard in `foreach`
```
thiagoftsm: Please, also change here "wildcard" for "simple pattern"

docs/tutorials/longer-metrics-storage.md
```diff
@@ -7,30 +7,27 @@ Many people think Netdata can only store about an hour's worth of real-time metr
...
 
 ```conf
 [global]
     memory mode = dbengine
 ```
 
-Next, restart Netdata. On Linux systems, we recommend running `sudo service netdata restart`. You're now using the
-database engine!
+If `memory mode` is set to anything but `dbengine`, change it and restart Netdata. On Linux systems, we recommend
+running `sudo service netdata restart`. You're now using the database engine!
```
thiagoftsm: The command "service" is not available in all Linuxes, it only exists on distribution that has ```systemd``` , is it not better to add an information after the service instruction for these distributions, something like "case you have systemd on your Linux or use the common command to restart Netdata on your distribution."?

docs/tutorials/longer-metrics-storage.md
```diff
@@ -69,15 +67,16 @@ aren't ready to make the move.
 
 ## Using the round-robin database
 
-By default, Netdata uses a round-robin database to store 1 hour of per-second metrics. Here's the default setting for
-`history` in the `netdata.conf` file that comes pre-installed with Netdata.
+In previous versions, Netdata used a round-robin database to store 1 hour of per-second metrics. To see if you're still
```
thiagoftsm: When we install netdata using netdata-static64.sh it creates a ``` nedata.conf``` with the following data:

```
# netdata can generate its own config which is available at 'http://<netdata_ip>/netdata.conf'
# You can download it with command like: 'wget -O /usr/local/netdata/etc/netdata/netdata.conf http://localhost:19999/netdata.conf'
```

A suggetion, instead to say for the person see the netdata.conf, it is safest to run the following command:

```
grep "memory mode" /var/log/netdata/error.log  | grep "INFO  : MAIN"
```

that will bring an output like this:

```
2019-10-03 13:30:29: netdata INFO  : MAIN : Host 'ceres' (at registry as 'hades') with guid 'YYYY' initialized, os 'linux', timezone 'UTC', tags '', program_name 'netdata', program_version 'v1.17.1-164-ged371b15', update every 1, memory mode ram, history entries 3996, streaming disabled (to '' with api key ''), health enabled, cache_dir '/var/cache/netdata', varlib_dir '/var/lib/netdata', health_log '/var/lib/netdata/health/health-log.db', alarms default handler '/usr/libexec/netdata/plugins.d/alarm-notify.sh', alarms default recipient 'root'

File: docs/tutorials/longer-metrics-storage.md [hunk 9]
```diff
@@ -56,9 +53,10 @@ size` and `dbengine disk space`.
 metrics.
 
 Based on our testing, these default settings will retain about two day's worth of metrics when Netdata collects 2,000
```
mfundul: With the new slab collectors that have been added to the agent I believe it's closer to 4,000 metrics now.

joelhans:
> Apart from hdfs/zookeepr we have 2 more new collectors:
> 
>     * [gearman](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/gearman)
> 
>     * [vcsa](https://github.com/netdata/go.d.plugin/tree/master/modules/vcsa)
> 
> 
> just saying, i dont know if we need to mention it

We'll mention this in our release notes. Thanks for the reminder though!

joelhans:
@ilyam8 @thiagoftsm Addressed (almost) all your comments and requests—thanks!

docs/tutorials/longer-metrics-storage.md
joelhans: Will the default settings will still store two days worth of metrics @ 4,000 metrics, or does this need to change to something like:

> Based on our testing, these default settings will retain about one day's worth of metrics when Netdata collects 4,000 metrics every second.

docs/tutorials/longer-metrics-storage.md
joelhans: I'm a little confused about how this is simpler than asking them to open up `netdata.conf`. Does the command offer better information, or are you thinking it's easier to pop that command into the terminal and look at the output?

docs/tutorials/longer-metrics-storage.md
mfundul: It will be one day now.

ilyam8:
@joelhans 

i have a question about lookup time in all examples. 

> -3s

It is totally ok for examples, but it is not so good in general. It will catch all cpi spikes (and it has no `delay` param).

See our stock cpu alarms, we dont use 3s there

> https://github.com/netdata/netdata/blob/master/health/health.d/cpu.conf

docs/tutorials/longer-metrics-storage.md
ilyam8: > grep "memory mode" /var/log/netdata/error.log  | grep "INFO  : MAIN"


there is no guarantee the path is `/var/log/...`

thiagoftsm: I used this, because it is the default option when we do not use `--install` and I found error.log generic.

ilyam8: ok, same goes for "memory mode", dont forget logs are rotated

thiagoftsm: I was thinking here, I believe the safest solution will be:

```
curl -o netdata_chart_list.json -k https://localhost:19999/api/v1/charts?all ; grep memory_mode netdata_chart_list.json
```

The expected result is

report

docs/tutorials/longer-metrics-storage.md
joelhans: Changed. Thank you!

joelhans:
@ilyam8 Good catch. They're some really basic alarms that Thiago wrote up as examples and I simplified. I have a note in the doc that the examples aren't meant to be used in real systems, but you're right that I can probably just change them to `-10m` and `every 1m` without any consequences.

joelhans:
@mfundul Can I get an approval on this if you're happy with the new text around the DB engine metrics?

PR #6994, opened 2019-10-04
Checklinks fix

### LINKED ISSUE #6982: Documentation build script breaks some relative links
##### Bug report summary

I noticed recently that many of the links between two files in the `docs/` folder are broken, even though in the source Markdown file they are valid relative links. Here's an example from `docs/getting-started.md`:

```
We have a tutorial that walks you through both options: [**Changing how long Netdata stores
metrics**](tutorials/longer-metrics-storage.md).
```

Because there is a file in `docs/tutorials/longer-metrics-storage.md`, and we are already in the `docs/` folder, this link should be valid and shouldn't need to be changed by the build script.

I found a few lines in the [checklinks.sh script](https://github.com/netdata/netdata/blob/master/docs/generator/checklinks.sh), which add extra parent "steps" to the links, and thus break these kinds of relative links.

https://github.com/netdata/netdata/blob/master/docs/generator/checklinks.sh#L120
https://github.com/netdata/netdata/blob/master/docs/generator/checklinks.sh#L133
https://github.com/netdata/netdata/blob/master/docs/generator/checklinks.sh#L148

I did find a workaround for this issue (#6983), where I link to `../docs/tutorials/longer-metrics-storage.md`, but I worry we won't remember to implement this workaround all the time. Given that neither our remark settings or the build script can identify these as broken, a permanent fix is the better solution.

##### Component Name

docs

cakrit:
##### Summary
Fixes #6982 

- Fix a bug that adds unecessary `../` before some links
- Better indentation to make debug messages easier to see.

##### Component Name
docs

##### Additional Information

@joelhans  found a bug in `checklinks.sh`.
> In the getting started guide (since the PR was merged in) there's a link to tutorials/longer-metrics-storage.md. That's a valid relative link, but then the script uses sed to add an extra ../ that breaks the link once the build completes. It links to tutorials/longer-metrics-storage/ instead of docs/tutorials/longer-metrics-storage/.
There's other places with broken links to/from files in the netdata-cloud/ and contributing/ folders, too.
I found two lines on checklinks.sh that add the ../: if [ "$fname" != "README.md" ] ; then s="../$s"; fi. They're on 148 and 167. When I comment those out the build finishes and the links work again.
Those lines are probably there for a reason, but I can't find that commenting them out introduces different broken links. At the time there weren't any folders nested inside of docs/, which seems to be where the trouble lies.

### DIFF
--- a/docs/generator/checklinks.sh
+++ b/docs/generator/checklinks.sh
@@ -31,16 +31,16 @@ printhelp () {
 
 fix () {
 	if [ "$EXECUTE" -eq 0 ] ; then
-		echo "-- SHOULD EXECUTE: $1"
+		echo "  - SHOULD EXECUTE: $1"
 	else
-		dbg "-- EXECUTING: $1"
+		dbg "  - EXECUTING: $1"
 		eval "$1"
 	fi
 }
 
 testURL () {
 	if [ "$TESTURLS" -eq 0 ] ; then return 0 ; fi
-	dbg "-- Testing URL $1"
+	dbg "  - Testing URL $1"
 	curl -sS "$1" > /dev/null
 	if [ $? -gt 0 ] ; then
 		return 1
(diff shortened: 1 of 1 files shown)


cakrit:
@joelhans will test this and compare the results with the other linkchecker. 
We can't get rid of this one yet, due to the way netlify builds the site (converting `foo/xyz.md` to `foo/xys/index.html`)

cakrit:
Waiting for @joelhans ' review to merge.

PR #7012, opened 2019-10-07
Documenting the structure of the data responses.

### LINKED ISSUE #7005: Responses to API calls not properly documented in swagger.yaml
All `200` responses should be properly documented. e.g. we currently show the `chart` response in the `data` call, which is incorrect and misleading.

amoss:
##### Summary
Fixes #7005 and some small miscellaneous problems and reviewing the source and documentation.
Fix the initial typo so that the response schema for /api/v1/data? is used instead of the /api/v1/chart? scheme.
Improve the structure of the "result" property in the object to show the fields.
Introduce polymorphic structure so that the different data formats are described.

##### Component Name
area/docs

##### Additional Information
The responses of the `api/v1/data` endpoint are now documented in swagger to match the behaviour in the code.
Tested request / response pairs against the demo system in the swagger editor.
The `400` response was missing from `api/v1/chart`.
Changed the description of the `api/v1/allmetrics` endpoint (but not the schema of the response).
Added missing fields to the `api/v1/chart_summary` endpoint.
Some quoted strings in the generated `.yaml` have been unquoted by the editor, the quotes seem to be optional.
Made the description strings consistent with full-stops and capitalisation.

### DIFF
--- a/web/api/netdata-swagger.json
+++ b/web/api/netdata-swagger.json
@@ -1262,7 +1262,7 @@
       ]
     },
     "data_flat": {
-      "description": "Data response in csv / tsv / tsv-excel / ssv / ssv-comma formats",
+      "description": "Data response in csv / tsv / tsv-excel / ssv / ssv-comma / markdown / html formats",
       "allOf": [
         {
           "$ref": "#/definitions/data"
@@ -1276,8 +1276,46 @@
         }
       ]
     },
+    "data_array": {
+      "description": "Data response in array format",
+      "allOf": [
+        {
+          "$ref": "#/definitions/data"
+        },
+        {
+          "properties": {
+            "result": {
+              "type": "array",
+              "items": {
+                "type": "number"
+              }
+            }
+          }
+        }
+      ]
+    },
+    "data_csvjsonarray": {
+      "description": "Data response in csvjsonarray format",
+      "allOf": [
+        {
+          "$ref": "#/definitions/data"
+        },
+        {
+          "properties": {
+            "result": {
+              "description": "The first inner array contains strings showing the labels of each column, each subsequent array contains the values for each point in time",
+              "type": "array",
+              "items": {
+                "type": "array",
+                "items": {}
+              }
+            }
+          }
+        }
+      ]
+    },
     "data_datatable": {
-      "description": "Data response in datatable format",
+      "description": "Data response in datatable / datasource formats (suitable for Google Charts)",
       "allOf": [
         {
           "$ref": "#/definitions/data"
@@ -1292,11 +1330,21 @@
                   "items": {
                     "type": "object",
                     "properties": {
-                      "id": {},
-                      "label": {},
-                      "pattern": {},
-                      "type": {},
-                      "p": {}
+                      "id": {
+                        "description": "Always empty - for future use"
+                      },
+                      "label": {
+                        "description": "The dimension returned from the chart"
+                      },
+                      "pattern": {
+                        "description": "Always empty - for future use"
+                      },
+                      "type": {
+                        "description": "The type of data in the column / chart-dimension"
+                      },
+                      "p": {
+                        "description": "Contains any annotations for the column"
+                      }
                     },
                     "required": [
                       "id",
@@ -1308,7 +1356,21 @@
                 },
                 "rows": {
                   "type": "array",
-                  "items": {}
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "c": {
+                        "type": "array",
+                        "items": {
+                          "properties": {
+                            "v": {
+                              "description": "Each value in the row is represented by a c value with five v fields: data, null, null, 0, the value. This format is fixed by the Google Charts API."
+                            }
+                          }
+                        }
+                      }
+                    }
+                  }
                 }
               }
             }
--- a/web/api/netdata-swagger.json
+++ b/web/api/netdata-swagger.json
@@ -1061,7 +1061,7 @@
         },
         "red": {
           "type": "number",
-          "description": "Chart health red trheshold."
+          "description": "Chart health red threshold."
         }
       }
     },
--- a/web/api/netdata-swagger.json
+++ b/web/api/netdata-swagger.json
@@ -71,6 +71,9 @@
               "$ref": "#/definitions/chart"
             }
           },
+          "400": {
+            "description": "No chart id was supplied in the request"
+          },
           "404": {
             "description": "No chart with the given id is found."
           }
@@ -914,6 +917,15 @@
           "type": "string",
           "description": "netdata version of the server."
         },
+        "release_channel": {
+          "type": "string",
+          "description": "The release channel of the build on the server",
+          "example": "nightly"
+        },
+        "timezone": {
+          "type": "string",
+          "description": "The current timezone on the server"
+        },
         "os": {
           "type": "string",
           "description": "The netdata server host operating system.",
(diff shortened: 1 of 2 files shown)


web/api/netdata-swagger.json
```diff
@@ -1224,12 +1242,158 @@
...
+          "$ref": "#/definitions/data"
+        },
+        {
+          "properties": {
+            "result": {
+              "description": "The first inner array contains strings showing the labels of each column, each subsequent array contains the values for each point in time",
+              "type": "array",
+              "items": {
+                "type": "array",
+                "items": {}
```
thiagoftsm: Is this a missing pair of `{}`? For all the previous you wrote something in front of every key.

web/api/netdata-swagger.json
```diff
@@ -1224,12 +1242,158 @@
...
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "c": {
+                        "type": "array",
+                        "items": {
+                          "properties": {
+                            "v": {
+                              "description": "Each value in the row is represented by a c value with five v fields: data, null, null, 0, the value. This format is fixed by the Google Charts API."
```
thiagoftsm: I understood the fact that you are talking about the keys when you wrote "represented by a c value", but is it not better to give some emphasis for these variables to give the notion that is not only a consonant?

web/api/netdata-swagger.yaml
```diff
@@ -35,137 +34,219 @@ paths:
...
+          description: No chart with the given id is found.
   /alarm_variables:
     get:
-      summary: 'List variables available to configure alarms for a chart'
-      description: 'Returns the basic information of a chart and all the variables that can be used in alarm and template health configurations for the particular chart or family'
+      summary: List variables available to configure alarms for a chart
+      description: >-
+        Returns the basic information of a chart and all the variables that can
+        be used in alarm and template health configurations for the particular
+        chart or family
```
thiagoftsm: You are missing a `.` at the end.

web/api/netdata-swagger.yaml
```diff
@@ -35,137 +34,219 @@ paths:
...
-          format: 'as returned by /charts'
+          format: as returned by /charts
       responses:
         '200':
-            description: 'A javascript object with information about the chart and the available variables'
-            schema:
-              $ref: '#/definitions/alarm_variables'
+          description: >-
+            A javascript object with information about the chart and the
+            available variables
```
thiagoftsm: You are missing `.` at the end, I know the original had the same problem, but I saw that there are descriptions with dots and others without, and I arrived in the conclusion that the best would be everything with dot at the end.

web/api/netdata-swagger.yaml
```diff
@@ -35,137 +34,219 @@ paths:
...
           type: string
-          format: 'as returned by /charts'
+          format: as returned by /charts
           allowEmptyValue: false
           default: system.cpu
         - name: dimension
           in: query
-          description: 'zero, one or more dimension ids or names, as returned by the /chart call, separated with comma or pipe. Netdata simple patterns are supported.'
+          description: >-
+            zero, one or more dimension ids or names, as returned by the /chart
```
thiagoftsm: Please, write `Z` in upper case in the word `zero`

web/api/netdata-swagger.yaml
```diff
@@ -225,182 +319,266 @@ paths:
...
           required: false
           type: string
-          format: 'any text'
+          format: any text
           allowEmptyValue: true
         - name: value_color
           in: query
-          description: 'a color to be used for the background of the label. You can set multiple using a pipe with a condition each, like this: color<value|color>value|color:null The following operators are supported: >, <, >=, <=, =, :null (to check if no value exists).'
+          description: >-
+            a color to be used for the background of the label. You can set
```
thiagoftsm: Please, uppercase for `a` in the beginning.

web/api/netdata-swagger.yaml
```diff
@@ -225,182 +319,266 @@ paths:
...
           required: false
           type: string
-          format: 'any text'
+          format: any text
           allowEmptyValue: true
         - name: multiply
           in: query
-          description: 'multiply the value with this number for rendering it at the image (integer value required)'
+          description: >-
+            multiply the value with this number for rendering it at the image
```
thiagoftsm: Please, uppercase for `m` in Multiply

web/api/netdata-swagger.json
amoss: I think this is here because the "items" is mandatory next to an array, but the contents of the array are untyped. The type changes between `string` and `number` after the first iteration. Maybe it looks strange because I squashed together the empty {} as a value?

web/api/netdata-swagger.json
amoss: Good point, updated.

web/api/netdata-swagger.yaml
amoss: I saw the same thing when I was writing the strings. In the end I went with dot-less to try to be consistent. I'll do a grep / count over the json and see which is more common.

amoss:
```
grep description web/api/netdata-swagger.json | wc -l
     183
grep description web/api/netdata-swagger.json | grep '\."' | wc -l
     105
```
Yes, you are correct - the dot at the end is grammatically correct and it is the more common choice. I'll put dots on all the descriptions.

amoss: Fixed.

web/api/netdata-swagger.yaml
amoss: Fixed.

web/api/netdata-swagger.yaml
amoss: Fixed.

web/api/netdata-swagger.yaml
amoss: Fixed.

web/api/netdata-swagger.yaml
amoss: See comments below.

thiagoftsm:
> I think this is here because the "items" is mandatory next to an array, but the contents of the array are untyped. The type changes between string and number after the first iteration. Maybe it looks strange because I squashed together the empty {} as a value?

Exactly this broke the previous pattern of to have a description in front a key, so I decided to ask you that this was not a missing information and it would be the correct information. Thank you!

amoss:
> There are 5 description fields with quotes (in web/api/netdata-swagger.yaml)
> Line 253 description: 'For JSONP responses, the callback function name.'
> Line 315 description: 'Zero, one or more dimension ids, as returned by the /chart call.'
> Line 584 description: 'If passed, all enabled alarms are returned.'
> Line 1038 description: 'The API version this conforms to, currently 1.'
> Line 1142 description: 'The result requested, in the format requested.'

The quotes do not change the semantic meaning of the `.yaml`. The value of the keys is a string anyway, the presence of the quotes simply forces the type. We didn't put it in, the swagger editor generated that file from the `.json` directly. I've checked the `.json` for those entries against the fields that do not have explicit quotes and they are the same. I this [this](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html) is the reference for the syntax and it indicates that the quotes are optional.

web/api/netdata-swagger.json
```diff
@@ -281,9 +284,9 @@
         ],
         "responses": {
           "200": {
-            "description": "The call was successful. The response should include the data.",
+            "description": "The call was successful. The response includes the data in the format requested. Swagger2.0 does not process the discriminator field to show polymorphism. The response will be one of the sub-types of  the data-schema according to the chosen format, e.g. json -> data_json.",
```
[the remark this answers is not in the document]
amoss: Eagle-eyed! Well spotted.

Issue #6721, opened 2019-08-23
AM2320 digital i2c temp sensor support


tommybuck:
##### Feature idea summary
Add support for the i2c AM2320 digital temp sensor on a raspberry pi.
https://www.adafruit.com/product/3721
##### Expected behavior
Show the current tempersture and humidity readings as two graphs under the sensors section of netdata.

ilyam8:
Hm, cant check the link

> Error 1020 Ray ID: 50afac21dd07e00f • 2019-08-23 20:00:23 UTC
Access denied
What happened?
This website is using a security service to protect itself from online attacks.

___

> i2c AM2320 digital temp sensor on a raspberry pi.

Do you know how to get metrics?

tommybuck:
See attached datasheet.

ilyam8:
nothing is attached

tommybuck:
[adafruit-am2320-temperature-humidity-i2c-sensor.pdf](https://github.com/netdata/netdata/files/3555741/adafruit-am2320-temperature-humidity-i2c-sensor.pdf)

ilyam8:
It is not clear from the pdf how to collect data 🤔

tommybuck:
If you read the full code example it prints out the current values from the
sensor.

import time
import board
import busio
import adafruit_am2320
# create the I2C shared bus
i2c = busio.I2C(board.SCL, board.SDA)
am = adafruit_am2320.AM2320(i2c)
while True:
print("Temperature: ", am.temperature)
print("Humidity: ", am.relative_humidity)
time.sleep(2)

ilyam8:
oh, nice, it looks it is easy to add a simple python module

@tommybuck 	try this:
 - copy code to `am2320.chart.py`
 - mode `am2320.chart.py` to python.d dir
 - restart netdata service
 - check the dashboard

```python
try:
    import board
    import busio
    import adafruit_am2320
    HAS_STUFF = True
except ImportError:
    HAS_STUFF = False


from bases.FrameworkServices.SimpleService import SimpleService


ORDER = [
    'temperature',
    'humidity',
]

CHARTS = {
    'temperature': {
        'options': [None, 'Temperature', 'celsius', 'temperature', 'am2320.temperature', 'line'],
        'lines': [
            ['temperature']
        ]
    },
    'humidity': {
        'options': [None, 'Relative Humidity', 'percentage', 'humidity', 'am2320.humidity', 'line'],
        'lines': [
            ['humidity']
        ]
[...]

PR #7024, opened 2019-10-08
add support for am2320 sensor

tommybuck:
##### Summary
fixes: #6721

Add support for am2320 I2C temperature and humidity sensor
##### Component Name
AM2320
##### Additional Information
see the readme file for installation tips.


collectors/python.d.plugin/am2320/am320.chart.py
```diff
@@ -0,0 +1,60 @@
...
+        except Exception as error:
+            self.error("error on creating I2C shared bus : {0}".format(error))
+            return False
+
+        return True
+
+    def get_data(self):
+        return {
+            'temperature': self.am.temperature,
+            'humidity': self.am.relative_humidity,
```
Ferroin: Is ti at all possible for access of either of these attributes to throw an exception?  If so, they need to be wrapped in a proper try/except clause, and if not it should probably be documented that they don't in a comment above the code in question.

ilyam8:
> file:///Users/ilyam/Downloads/adafruit-am2320-temperature-humidity-i2c-sensor.pdf

> If you want to take another reading, wait 2 seconds before requesting temperature or relative_humidity
again. Data in the sensor is only updated every 2 seconds!

It doesnt mean we shouldnt collect data every seconds but we need to mention it in the readme.

collectors/python.d.plugin/am2320/am320.chart.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def check(self):
+        if not HAS_STUFF:
+            self.error("following libraries are needed to get module working: board, busio, adafruit_am2320")
+            return False
+
+        try:
+            i2c = busio.I2C(board.SCL, board.SDA)
+            self.am = adafruit_am2320.AM2320(i2c)
+        except Exception as error:
```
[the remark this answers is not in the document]
ilyam8: > https://github.com/adafruit/Adafruit_CircuitPython_AM2320/blob/26c73be5d524517e3f019716d952004a8f4c4dcb/adafruit_am2320.py#L87-L123


> adafruit_am2320.AM2320(i2c)

__init__:
 - `ValueError`

temperature/humidity:
 - `OSError`
 - `RuntimeError`

Since they dont use library specifif exceptions probably it is ok to use `Exception` here. 

What do you think @Ferroin ?

collectors/python.d.plugin/am2320/am320.chart.py
```diff
@@ -0,0 +1,60 @@
...
+class Service(SimpleService):
+    def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.order = ORDER
+        self.definitions = CHARTS
+        self.am = None
+
+    def check(self):
+        if not HAS_STUFF:
+            self.error("following libraries are needed to get module working: board, busio, adafruit_am2320")
```
Ferroin: This should ideally reference the package that the modules are part of instead of the specific modules, especially since `board` and `busio` are really generic and don't turn up a specific package when searching [PyPI](https://pypi.org/).

collectors/python.d.plugin/am2320/am320.chart.py
ilyam8: > Is it at all possible for access of either of these attributes to throw an exception?

 - `OSError`
 - `RuntimeError`

> https://github.com/adafruit/Adafruit_CircuitPython_AM2320/blob/26c73be5d524517e3f019716d952004a8f4c4dcb/adafruit_am2320.py#L98-L137

Ferroin:
Aside from the code comments I made:

* The name of the README file should be `README.md`. That's why the build is breaking right now (the Makefile assumes that specific name when building things).
* The README should have proper sections to it using regular markdown syntax.  Take a look at how the README.md files for the other Python plugins are structured.
* You need a proper header on the Python code.  At minimum, it needs to have a comment at the top of the file with the correct SPDX license tag.  Check the first few lines of any of the other plugins for what this should look like.

Let me know if you need some help figuring out the last two points.

collectors/python.d.plugin/am2320/am320.chart.py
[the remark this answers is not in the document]
Ferroin: I still think we should check those specific errors instead of a blanket `Exception` case.  I'll open an issue in the upstream repo to hopefully get the case of them raising a `RuntimeError` instead of using  a proper local exception class fixed.

collectors/python.d.plugin/am2320/am320.chart.py
ilyam8: yes, according docs all user to do is

> sudo pip3 install adafruit-circuitpython-am2320

so call it `AM2320 library`

collectors/python.d.plugin/am2320/am320.chart.py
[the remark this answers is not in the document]
ilyam8: @tommybuck

use `ValueError` here

collectors/python.d.plugin/am2320/am320.chart.py
ilyam8: @tommybuck

```python
try:
    return ...
except (OSError, RuntimeError) as error:
    self.error(error)
    return None
```

collectors/python.d.plugin/am2320/am320.chart.py
Ferroin: I'd go a bit further and refer to the full package name.  It's not likely that there will be any confusion over the naming, but it's best to just be specific when writing error messages.

collectors/python.d.plugin/am2320/am320.chart.py
Ferroin: @tommybuck Wrapping the function body itself.  So the whole function would be:

```python
def get_data(self):
    try:
        return {
            'temperature': self.am.temperature,
            'humidity': self.am.relative_humidity,
        }
    except (OSError, RuntimeError) as error:
        self.error(error)
        return None
```

collectors/python.d.plugin/am2320/am320.chart.py
Ferroin: The full PyPI package name is adafruit-circuitpython-am2320, so the error message should read something like either:

`Could not find the AM2320 library.`

or:

`Could not find the adafruit-circuitpython-am2320 package.`

ilyam8: HAS_STUFF -> HAS_AM2320

following libraries are needed -> adafruit-circuitpython-am2320 library is needed

Ferroin:
> > Aside from the code comments I made:
> > ```
> > * The name of the README file should be `README.md`. That's why the build is breaking right now (the Makefile assumes that specific name when building things).
> > 
> > * The README should have proper sections to it using regular markdown syntax.  Take a look at how the README.md files for the other Python plugins are structured.
> > 
> > * You need a proper header on the Python code.  At minimum, it needs to have a comment at the top of the file with the correct SPDX license tag.  Check the first few lines of any of the other plugins for what this should look like.
> > ```
> > 
> > 
> > Let me know if you need some help figuring out the last two points.
> 
> yes please.

Full header for the Python file should be something along the lines of:

```python
# -*- coding: utf8 -*-
# Description: AM2320 netdata module
# Author: Tom Buck (tommybuck)
```

The first line signals some older versions of Python that the file is UTF-8 encoded instead of whatever the local default is.

The second and third help identify the file itself (for the author line, you don't have to put your full name if you don't want to, you could do just your GitHub account name instead).

The fourth is a licensing tag.  It's essentially equivalent to the boilerplate license text that has been the norm for years for FOSS projects to put at the top of their files.  This is the only one that's strictly required (all code _has_ to have a license declared).

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,37 @@
+# AM2320
+This module will display a graph of the temperture and humity from a AM2320 sensor
+
+**Requirements:**
+ - Adafruit Circuit Python AM2320 library
+ - Adafruit AM2320 I2C sensor
+ - Python 3
```
Ferroin: The module needs to work with Python 2 as well as Python 3 if at all possible.  Your code looks like it should work find, but I'm not sure about the Adafruit libraries.

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,37 @@
...
+ - Raspberry Pi board
```
Ferroin: Unless I'm mistaken, this should work with any system that can run Linux which supports the Adafruit Circuit Python module for the AM2320 sensor, correct?

If so, it would probably be better to remove this line from the requirements and update the configuration section to clarify that the instructions for for a Raspberry Pi.

Ferroin:
You appear to have a typo in the name of the Python file.  Pretty sure it should be `am2320.chart.py` not `am320.chart.py`.

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,37 @@
+# AM2320
+This module will display a graph of the temperture and humity from a AM2320 sensor
```
joelhans: Put a period at the end.

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Hardware install:
+Connect the am2320 to the Raspbery Pi I2C pins
+
+Raspberry Pi 3B/4 Pins:
+
+Board 3.3V (pin 1) to sensor VIN (pin 1)
+Board SDA (pin 3) to sensor SDA (pin 2)
+Board GND (pin 6) to sensor GND (pin 3)
+Board SCL (pin 5) to sensor SCL (pin 4)
```
joelhans: Maybe this is better as an unordered list?

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,37 @@
...
+Software install:
+sudo pip3 install adafruit-circuitpython-am2320
+sudo usermod -G I2C netdata
+edit /etc/netdata/netdata.conf
+find [plugin:python.d]
+add  command options = -ppython3
+Save file
+move am2320.chart.py to /usr/libexec/netdata/python.d 
+restart netdata service
+check the dashboard
```
joelhans: This would probably work better as an unordered list as well. Also, I'd like to see commands and references to files in code blocks.

collectors/python.d.plugin/am2320/am2320.chart.py
```diff
@@ -0,0 +1,65 @@
...
+    import board
+    import busio
+    import adafruit_am2320
+    HAS_AM2320 = True
+except ImportError:
+    HAS_AM2320 = False
+
+
+#from bases.FrameworkServices.SimpleService import SimpleService
+from base import SimpleService
```
ilyam8: use

> from bases.FrameworkServices.SimpleService import SimpleService

collectors/python.d.plugin/am2320/am2320.chart.py
```diff
@@ -0,0 +1,65 @@
...
+        return True
+
+    def get_data(self):
+        try:
+            return {
+            'temperature': self.am.temperature,
+            'humidity': self.am.relative_humidity,
+           }
+
+        except (OSError, RuntimeError) as error:
```
ilyam8: note: we need to understand if we need to re init `adafruit_am2320.AM2320` in case of `OSError, RuntimeError`

ilyam8: probably not, can be fixed later if this is wrong

tommybuck:
Ok so how do we do that?

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,44 @@
...
+You may also need to add two I2C pullup resistors if your board does not already have them. The Raspberry Pi does have internal pullup resistors but it doesnt hurt to add them anyway. You can use 2.2K - 10K but we will just use 10K. The resistors go from VDD to SCL and SDA each.
+
+Software install:
+- `sudo pip3 install adafruit-circuitpython-am2320`
+- `sudo usermod -G I2C netdata`
+- edit `/etc/netdata/netdata.conf`
+- find `[plugin:python.d]`
+- add  `command options = -ppython3`
+- Save the file.
+- move `am2320.chart.py` to `/usr/libexec/netdata/python.d`
```
ilyam8: you dont need to move it, it will be installed to the dir after the PR is merged

tommybuck:
Can this be merged now?

collectors/python.d.plugin/am2320/am2320.chart.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def check(self):
+        if not HAS_AM2320:
+            self.error("Could not find the adafruit-circuitpython-am2320 package.")
+            return False
+
+        try:
+            i2c = busio.I2C(board.SCL, board.SDA)
+            self.am = adafruit_am2320.AM2320(i2c)
+        except Exception as error:
```
ilyam8: please change it to `ValueError`

collectors/python.d.plugin/am2320/am2320.chart.py
```diff
@@ -0,0 +1,64 @@
+try:
```
ilyam8: https://github.com/netdata/netdata/blob/380529c7cf217f62ffcc7347a93596c12d7fe90e/collectors/python.d.plugin/apache/apache.chart.py#L1-L5

please add the header

ilyam8:
@tommybuck 

What about alarms? Does it make sense to add and alarm for high/low temperature? 

I suggest to add at least `avalability` alarm

see

https://github.com/netdata/netdata/blob/master/health/health.d/apache.conf


```
template: am2320_last_collected_secs
      on: am2320.temperature
    calc: $now - $last_collected_t
   units: seconds ago
   every: 10s
    warn: $this > (($status >= $WARNING)  ? ($update_every) : ( 5 * $update_every))
    crit: $this > (($status == $CRITICAL) ? ($update_every) : (60 * $update_every))
      to: webmaster
```

you need to add it to the makefile 

https://github.com/netdata/netdata/blob/380529c7cf217f62ffcc7347a93596c12d7fe90e/health/Makefile.am#L23-L26

Ferroin:
> What about alarms? Does it make sense to add and alarm for high/low temperature?

This is an external sensor, not something integrated into a specific system.  Exact values for such alarms would be 100% dependent on what the user is using the sensor for.

I do agree though that an availability alarm makes sense.

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,43 @@
+# _*_ coding: utf-8 _*_
+# Description: AM2320 netdata module
+# Author: tommybuck
+# SPDX-License-Identifier: GPL-3.0-or-Later
```
joelhans: I believe this header is supposed to go in the `am2320.chart.py` file.

ilyam8: nice catch! 😄

collectors/python.d.plugin/am2320/am2320.chart.py
ilyam8: move the header from the README.md

collectors/python.d.plugin/am2320/am2320.chart.py
tommybuck: done

collectors/python.d.plugin/am2320/am2320.chart.py
tommybuck: done

collectors/python.d.plugin/am2320/README.md
tommybuck: done

health/health.d/am2320.conf
```diff
@@ -0,0 +1,12 @@
+# mae sure am2320 is sending stats
```
ilyam8: make

tommybuck: :man_facepalming:

tommybuck: oops. done.

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,38 @@
...
+- Board 3.3V (pin 1) to sensor VIN (pin 1)
+- Board SDA (pin 3) to sensor SDA (pin 2)
+- Board GND (pin 6) to sensor GND (pin 3)
+- Board SCL (pin 5) to sensor SCL (pin 4)
+
+You may also need to add two I2C pullup resistors if your board does not already have them. The Raspberry Pi does have internal pullup resistors but it doesnt hurt to add them anyway. You can use 2.2K - 10K but we will just use 10K. The resistors go from VDD to SCL and SDA each.
+
+Software install:
+- `sudo pip3 install adafruit-circuitpython-am2320`
+- `sudo usermod -G I2C netdata`
```
ilyam8: this can be done during install

https://github.com/netdata/netdata/blob/cd097aad6f29e41340c275448e1e3398f0bb8396/packaging/makeself/install-or-update.sh#L87

collectors/python.d.plugin/am2320/README.md
```diff
@@ -0,0 +1,38 @@
...
+- edit `/etc/netdata/netdata.conf`
+- find `[plugin:python.d]`
+- add  `command options = -ppython3`
+- Save the file.
```
ilyam8: make it lowercase

tommybuck: Make what lowercase?

ilyam8: > Save

tommybuck: done

collectors/python.d.plugin/am2320/README.md
tommybuck: done. Also removed the instruction from the README.md file.

ilyam8:
@joelhans please take a look. The PR overall lgtm.

ilyam8:
@tommybuck good job 👍

health/Makefile.am
```diff
@@ -22,6 +22,7 @@ dist_userhealthconfig_DATA = \
 healthconfigdir=$(libconfigdir)/health.d
 dist_healthconfig_DATA = \
     health.d/adaptec_raid.conf \
+	health.d/am2320.conf \
```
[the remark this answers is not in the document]
tommybuck: done.

thiagoftsm:
Thank you very much @tommybuck !

PR #7027, opened 2019-10-08
netdata/installer: add missing flags on installer

### LINKED ISSUE #7025: ./netdata-installer.sh --disable-https
I would like to disable https, but I can't use netdata-installer.sh with option --disable-https

paulkatsoulakis:
##### Summary
As per our #7025 report, we missed to add https and dbengine disable options to the list of available options for our `netdata-installer.sh`

This PR mitigates that issue

**Note**: Haven't tested it, @knatsakis please help with verification and approval of PR

##### Component Name
netdata/installer

##### Additional Information
Fixes #7025

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -158,6 +158,8 @@ USAGE: ${PROGRAM} [options]
   --disable-go               Disable installation of go.d.plugin.
   --enable-plugin-freeipmi   Enable the FreeIPMI plugin. Default: enable it when libipmimonitoring is available.
   --disable-plugin-freeipmi
+  --disable-https            Explicitly disable TLS support
+  --disable-dbengine         Explicitly disable DB engine support
   --enable-plugin-nfacct     Enable nfacct plugin. Default: enable it when libmnl and libnetfilter_acct are available.
   --disable-plugin-nfacct
   --enable-plugin-xenstat    Enable the xenstat plugin. Default: enable it when libxenstat and libyajl are available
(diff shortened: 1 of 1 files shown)


paulkatsoulakis:
To test this you basically need to run `netdata-installer.sh` with the mentioned flags and make sure that the configure output shows that given functionality is disabled (set to no)

cosmix:
@paulkatsoulakis this cannot be merged as is, based on the comment above by @thiagoftsm.

paulkatsoulakis:
> @paulkatsoulakis this cannot be merged as is, based on the comment above by @thiagoftsm.

Hi @cosmix, this change is only focusing on allowing the user to pass those two flags from the installer down configure command. The problem with the flag, sounds like an orthogonal issue related to the recent changes of netdata's default DB engine mode

thiagoftsm:
Hi @paulfantom,

you are right, we changed things in our side that this PR is not considering according the results, this is the motive I gave some screenshots  for you understand better. Now netdata has `dbengine` as default memory mode and when it does not have conditions to use it, netdata uses `save`, but somehow this PR decided to set for `none`. It is expected that when the user disables dbengine he will have netdata using by default `save`.

knatsakis:
Hi @thiagoftsm,

I have tested this PR with the latest master, and I got `memory mode = save` when I ran the installer with `--disable-dbengine`.

Could you please test again with the latest master?

Thanks

PR #7037, opened 2019-10-09
fixed - cgroup-network-helper doesn't work on Proxmox 6

vakartel:
##### Summary
add a check for 'emulator' subdirectory
add full cgroup-path with 'cgroup.procs' filename to 'c' variable

##### Component Name
 collectors/cgroups.plugin/cgroup-network-helper.sh


ilyam8:
@vakartel thx!

Issue #7030, opened 2019-10-08
pfSense cannot currently install netdata version 1.17.1


lordcantide:
##### Bug report summary
netdata version 1.17.1 service will not start after upgrade from 1.15
running command "/usr/local/sbin/netdata" generates the following error:
-Shared object "libjson-c.so.4" not found, required by "netdata"

The file appears related to the json-c library dependency and is apparently supposed to be optional for netdata. The current stable release of pfSense distributes with json-c version 0.13, but patch 0.13.1 changelog references a version bump which appears related to this issue. It is unclear when pfSense will distribute a package update given that the relevant update was ported to FreeBSD 5/Sept/2018.

##### OS / Environment
pfSense 2.4.4-RELEASE-p3

##### Netdata version (ouput of `netdata -V`)
Does not run. See error in summary.

##### Steps To Reproduce
1) Remove netdata version 1.15 and old dependencies.
2) Reinstall using instructions documented here - https://docs.netdata.cloud/packaging/installer/#pfsense
3) Edit "bind to" in netdata.conf
4) Attempt to "service netdata onestart" without errors.
5) Check "service netdata onestatus" to verify that service is not running.
6) Run "/usr/local/sbin/netdata" to confirm error in summary.

##### Expected behavior
netdata service will start.

thiagoftsm:
Hi @lordcantide ,

Thank you for your report, I already could see that our documentation is missing important information about the installation for 1.17.1:

![VirtualBox_pfsense_09_10_2019_13_36_10](https://user-images.githubusercontent.com/49162938/66486445-1fbf3a80-ea9a-11e9-8a74-7dc17491c052.png)

It is also possible that our image has not been compiled correctly, we are studying the case and we will release the fix ASAP.

Best regards

thiagoftsm:
Confirming what I said before, we have problem in the documentation and our machine that generates the image are not compatible with 1.17.1, another thing, the source of the package is not kept by Netdata it is kept by Netdata team (pkg.freebsd.org), I will talk with our team to regenerate the packages and I will contact the FreeBSD team for we fix this.
Finally, here is the last output after I detect problem with packages:

![VirtualBox_pfsense_09_10_2019_13_59_13](https://user-images.githubusercontent.com/49162938/66488525-9578d580-ea9d-11e9-9e2a-608abd7db6fe.png)

diizzyy:
This is not a bug as pfsense doesn't use FreeBSD's package repo by default
[...]

PR #7041, opened 2019-10-09
Fix BSD/pfSense documentation

thiagoftsm:
##### Summary
Fixes part of #7030 

It was verified that our instructions to install Netdata on FreeBSD and pfSense was missing some packages, this PR brings the packages that were missed, but it does not solve yet the main problem that is the missing link. I gave a hot fix that worked for our user, but we need to compile Netdata with the libraries present in pfSense and deliver a package specific for it, because FreeBSD and pfSense are not having more the same library at least for json-c.

To have a complete fix would be necessary to execute the three steps:

```
pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/Judy-1.0.5_2.txz
ln -s /usr/local/lib/libjson-c.so /usr/local/lib/libjson-c.so.4
pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.17.1.txz
```

but, I did not convince myself that we need to say for our user to create a link, so I am adding Judy and keeping yet the version 1.15.0. 
##### Component Name
Documentation


packaging/installer/README.md
```diff
@@ -347,13 +347,14 @@ cd netdata
...
+Note that the first four packages are downloaded from the pfSense repository for maintaining compatibility with pfSense, Netdata, Judy and Python are downloaded from the FreeBSD repository.
 
 ```sh
 pkg install pkgconf
 pkg install bash
 pkg install e2fsprogs-libuuid
 pkg install libuv
+pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/Judy-1.0.5_2.txz
 pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/python36-3.6.9.txz
 pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.15.0.txz
```
amoss: Should the version number here be updated to 17.1?

packaging/installer/README.md
```diff
@@ -333,7 +333,7 @@ This is how to install the latest Netdata version from sources on FreeBSD:
 
 ```sh
 # install required packages
-pkg install bash e2fsprogs-libuuid git curl autoconf automake pkgconf pidof
+pkg install bash e2fsprogs-libuuid git curl autoconf automake pkgconf pidof Judy liblz4 libuv json-c
```
amoss: I don't have clean system to test this on, but have you got openssl-devel installed as well?

packaging/installer/README.md
thiagoftsm: Case we bring 1.17.1 we will need to force our users to create a manual link, because the current package is linked with a difference version and Netdata does not start, but I agree that this would be the correct. Should we ask user to create the link for the library?

packaging/installer/README.md
thiagoftsm: Everytime that I installed FreeBSD I had the openssl installed, this is the motive I did not add it in the list, did you already see a FreeBSD without it?

diizzyy: OpenSSL is in base so there's no need to link it against ports unless you need bleeding edge for some reason

packaging/installer/README.md
diizzyy: These are moving targets so expect breakage quite frequently if you're going to hardlink packages. There's a quartely branch but using the provided package framework is the best way to avoid surprises.

packaging/installer/README.md
amoss: The version in base doesn't include pkg-config info and we try to autodetect the package at install time. I think it works fine in the regular installer, it is only the CMake routines that give a problem at the moment, but I thought I would check.

amoss: The installer should find the libcrypto.so from base, but there is no pkg-config to work against in that case. I'll install pfSense tomorrow morning and check the initial state and the fix.

thiagoftsm: On BSD I agree 100%, but the problem with pfSense is a little more complicated, it looks like that pfSense is not keeping the same soft link used on FreeBSD and we are installing package from FreeBSD repository, this is the motive our user reported the issue, I think the best way to solve this is either to compile the packages with the correct libraries or to link static with json-c, considering that we began to have this problem between both Operate System.

amoss: When I test this on a fresh 2.4.4-RELEASE both json-c and liblz4 are installed by default. Is this a packaging problem on the `-p3` release candidate that disappears if you do an upgrade to the released version?

packaging/installer/README.md
vlvkobal:
```
# pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.15.0.txz
pkg: http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.15.0.txz: Not Found
```

The only package available in the `latest` repository is `netdata-1.17.1.txz`

thiagoftsm: A very good point Vladimir with this last I am convinced that it is necessary to add the `ln` command that I was avoiding until now, but I think we need to fix this for our users only install packages.

packaging/installer/README.md
thiagoftsm: I did not test on -p3, but this list of packages is for FreeBSD, this PR fixes both, because I saw neither FreeBSD nor pfSense specification would allow the users to have dbengine activated.

packaging/installer/README.md
amoss: After adding the Judy package the following works:
```
pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.17.1.txz
```
Please update the PR with the change to the version number.

thiagoftsm: Nice, I already updated with the last commit.

packaging/installer/README.md
```diff
@@ -356,7 +356,8 @@ pkg install e2fsprogs-libuuid
 pkg install libuv
 pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/Judy-1.0.5_2.txz
 pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/python36-3.6.9.txz
-pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.15.0.txz
+ln -s /usr/local/lib/libjson-c.so /usr/local/lib/libjson-c.so.4
```
amoss: This is not what I see when I test it on 2.4.4-RELEASE.
After installing the json-c package I have `/usr/local/lib/libjson-c.so -> libjson-c.so.3.0.1`, but I also have the problem that `service onestart netdata` does not start the system and I cannot find any errors. Still investigating...

amoss: Resolved with @thiagoftsm over slack - instructions are correct and testing has been replicated.

PR #7051, opened 2019-10-10
kickstart: pass options to installer

### LINKED ISSUE #7040: --stable-channel option not working
##### Bug report summary
When using the --stable-channel option to only update upon major releases, the installation does not seem to honor that option.  The generated .environment file contains:
RELEASE_CHANNEL="nightly"

##### OS / Environment
cat /etc/os-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.6 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.6"
PRETTY_NAME="Red Hat Enterprise Linux"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.6:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.6
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.6"

netdata -V
netdata v1.17.1

##### Component Name
packaging/installer

##### Steps To Reproduce
Using either of the installation commands does not honor --stable-channel:
bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait --stable-channel
bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh) --dont-wait --stable-channel

##### Expected behavior
Expecting the generated .environment file to contain:
RELEASE_CHANNEL="stable"

so that when the cron.daily job runs, it will only update upon major release.

oxplot:
- pass --stable-channel option
- append --dont-wait option to list of passthrough options

##### Summary

Passes through kickstarter options to the installer

##### Component Name

Packaging - Installer

##### Additional Information

Fixes #7040

### DIFF
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -194,6 +194,7 @@ while [ -n "${1}" ]; do
 		shift 1
 	elif [ "${1}" = "--stable-channel" ]; then
 		RELEASE_CHANNEL="stable"
+		inner_opts="${inner_opts} ${1}"
 		shift 1
 	elif [ "${1}" = "--local-files" ]; then
 		shift 1
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -194,7 +194,6 @@ while [ -n "${1}" ]; do
 		shift 1
 	elif [ "${1}" = "--stable-channel" ]; then
 		RELEASE_CHANNEL="stable"
-		inner_opts="${inner_opts} ${1}"
 		shift 1
 	elif [ "${1}" = "--local-files" ]; then
 		shift 1
--- a/packaging/installer/kickstart.sh
+++ b/packaging/installer/kickstart.sh
@@ -307,6 +307,7 @@ while [ -n "${1}" ]; do
 		shift 1
 	elif [ "${1}" = "--stable-channel" ]; then
 		RELEASE_CHANNEL="stable"
+		NETDATA_INSTALLER_OPTIONS="$NETDATA_INSTALLER_OPTIONS --stable-channel"
 		shift 1
 	elif [ "${1}" = "--local-files" ]; then
 		shift 1
@@ -351,7 +352,7 @@ done
 
 if [ "${INTERACTIVE}" = "0" ]; then
 	PACKAGES_INSTALLER_OPTIONS="--dont-wait --non-interactive ${PACKAGES_INSTALLER_OPTIONS}"
-	NETDATA_INSTALLER_OPTIONS="--dont-wait"
+	NETDATA_INSTALLER_OPTIONS="$NETDATA_INSTALLER_OPTIONS --dont-wait"
 fi
 
 TMPDIR=$(create_tmp_directory)
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -48,7 +48,7 @@ To learn more about the pros and cons of using *nightly* vs. *stable* releases, 
 Verify the integrity of the script with this:
 
 ```bash
-[ "735e9966a4cf0187863e06a5282b34a7" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
+[ "0ae8dd3c4c9b976c4342c9fc09d9afae" = "$(curl -Ss https://my-netdata.io/kickstart.sh | md5sum | cut -d ' ' -f 1)" ] && echo "OK, VALID" || echo "FAILED, INVALID"
 ```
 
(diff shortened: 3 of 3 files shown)


paulkatsoulakis:
we also need the checksum info to be updated on packaging/installer/readme.md when modifying kickstarts

knatsakis:
This looks like a good fix for kickstart.sh. But the case with kickstart-static64.sh is more complicated.

kickstart.sh does generate `.environment` but this is not the case for kickstart-static64.sh.

In the second case the `.environment` file is include in netdata-$version.gz.run. So,
`.environment` has to be fixed in netdata-$version.gz.run in the first place or kickstart-static64.sh must sed it

ilyam8:
@oxplot please sign the CLA

packaging/installer/kickstart-static64.sh
```diff
@@ -194,6 +194,7 @@ while [ -n "${1}" ]; do
 		shift 1
 	elif [ "${1}" = "--stable-channel" ]; then
 		RELEASE_CHANNEL="stable"
+		inner_opts="${inner_opts} ${1}"
```
knatsakis: Hi @oxplot, are you sure that the inner script accepts the "--stable-channel" option?

paulkatsoulakis:
> Also @paulkatsoulakis please ensure that @knatsakis knows how to update the two scripts, once the PR is merged.

We 've already discussed the details with @knatsakis earlier today, but feel free to ping me should you need more help or documentation

packaging/installer/kickstart.sh
```diff
@@ -307,6 +307,7 @@ while [ -n "${1}" ]; do
 		shift 1
 	elif [ "${1}" = "--stable-channel" ]; then
 		RELEASE_CHANNEL="stable"
+		NETDATA_INSTALLER_OPTIONS="$NETDATA_INSTALLER_OPTIONS --stable-channel"
```
knatsakis: Could you make this so that it doesn't add an extra space when `$NETDATA_INSTALLER_OPTIONS` is empty?

packaging/installer/kickstart.sh
```diff
@@ -351,7 +352,7 @@ done
 
 if [ "${INTERACTIVE}" = "0" ]; then
 	PACKAGES_INSTALLER_OPTIONS="--dont-wait --non-interactive ${PACKAGES_INSTALLER_OPTIONS}"
-	NETDATA_INSTALLER_OPTIONS="--dont-wait"
+	NETDATA_INSTALLER_OPTIONS="$NETDATA_INSTALLER_OPTIONS --dont-wait"
```
knatsakis: The same here: could you make this so that it doesn't add an extra space when $NETDATA_INSTALLER_OPTIONS is empty?

packaging/installer/kickstart.sh
oxplot: This is harmless in bash and spaces are collapsed. I don't think it justifies the extra fluff. Also, https://github.com/netdata/netdata/blob/master/packaging/installer/kickstart-static64.sh#L193 seems to do the same.

packaging/installer/kickstart.sh
oxplot: See my other comment

packaging/installer/kickstart-static64.sh
oxplot: oh I see — I wrongly assumed based on https://github.com/netdata/netdata/pull/7051#pullrequestreview-300044516 that both scripts behave the same.

I suggest then to only fix kickstart and leave kickstart-static64 for another PR, given how trivial this fix is (so it doesn't drag on).

oxplot:
are we happy to leave changes to static64 to another PR?

packaging/installer/kickstart-static64.sh
paulkatsoulakis: If i may say, @knatsakis if we have decided that through netdata-updater.sh for the static install, we want to automatically set the release channel variable in .environment to stable, when the user uses the stable flag so that updater is automatically pulling stable releases, then it makes sense to add the option. But to make it complete we will need to add handling on the other side too (makeself installer).

If we agree that we do not want to automatically force the stable release channel on the static install, as it was the initial decision with @cakrit when we fixed the updater to work for the static install, then indeed no need to add it to inner opts

knatsakis:
> are we happy to leave changes to static64 to another PR?

Yes, please leave changes to static64 to another PR.
Kindly undo those changes so that I can merge this PR.

PR #7055, opened 2019-10-10
netdata/ci: fix package generation flow, during release

paulkatsoulakis:
##### Summary
Because tasks within a stage run in parallel, this messes up our workflow for the release.
We rely on the set of commits in place to take decisions, so packaging must be triggered once draft release is complete.

To achieve that, we remove the separate step and make package generation a job within the draft release generation

##### Component Name
netdata/ci

##### Additional Information
Fix for the release process


knatsakis:
As per @paulkatsoulakis request, I created a new PR (#7105) for fixing #7063 

Closing this.

PR #7056, opened 2019-10-10
Disable slabinfo plugin by default

### LINKED ISSUE #7053: slabinfo: disable by default
##### Feature idea summary

`slabinfo` plugin enabled by default, it collects huge amount of metrics. Number of collected metrics increased on ~50% on my vm with this plugin enabled (~1100->1700).

I may be wrong but these metrics are not so useful for most of the people. And the representation is kind of lame - several charts with 100+ dimensions. 

So it doesnt add much value but adds a lot of pressure on network usage. Just imagine how much network  usage increased  in master-slave env.

##### Expected behavior

Disable `slabinfo` plugin until some filters/limiters implemented to reduce the amount of collected metrics.

vlvkobal:
##### Summary
There are a lot of metrics in slabinfo charts. Hundreds or even thousands (depends on host). Thus, we are disabling the slabinfo charts by default.

Fixes #7053

##### Component Name
slabinfo plugin

### DIFF
--- a/collectors/slabinfo.plugin/README.md
+++ b/collectors/slabinfo.plugin/README.md
@@ -7,7 +7,9 @@ Each internal structure (process, file descriptor, inode...) is stored within a 
 
 ## configuring Netdata for slabinfo
 
-There is currently no configuration needed.
+The plugin is disabled by default because it collects and displays a huge amount of metrics.
+
+There is currently no configuration needed for the plugin itself.
 
 As `/proc/slabinfo` is only readable by root, this plugin is setuid root.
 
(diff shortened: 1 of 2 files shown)


collectors/slabinfo.plugin/README.md
```diff
@@ -7,7 +7,9 @@ Each internal structure (process, file descriptor, inode...) is stored within a
 
 ## configuring Netdata for slabinfo
 
-There is currently no configuration needed.
+The plugin is disabled by default because it collects and displays a huge amount of metrics.
```
ilyam8: How about to mention how to enable it?

vlvkobal: Fixed

Issue #6033, opened 2019-05-16
Defunct timeout processes from charts.d plugin in Docker


utdrmac:
##### Bug report summary
Unable to take actions on system due to inability to fork new processes due to netdata creation of `timeout` defunct processes

##### OS / Environment
Docker, 1.18
Ubuntu 18.04

##### Netdata version (ouput of `netdata -V`)
```
# netdata -V
netdata v1.14.0-11-ge91b9ca
```

##### Component Name
netdata core / charts.d / nut plugin?

##### Steps To Reproduce
Run netdata in docker.
Disable 'apps' plugin.
Enable charts.d plugin for nut, configure nut plugin

```
# cat netdata.conf
[plugins]
  apps = no
```

##### Expected behavior
For netdata to not exhaust kernel of process slots

```
-- The initial discovery of a problem on the system
# docker start grafana
runtime/cgo: pthread_create failed: No space left on device
runtime/cgo: pthread_create failed: No space left on device
SIGABRT: abort
PC=0x7f6c7d022e97 m=3 sigcode=18446744073709551610
...

-- Plus received this alarm from another cron job running
/bin/sh: 1: Cannot fork

-- No space? That's odd.
# df -h /
Filesystem                         Size  Used Avail Use% Mounted on
/dev/mapper/ubuntu--vg-ubuntu--lv  5.9G  3.7G  2.0G  66% /

-- No, not a space issue.
-- What is currently running?
# ps -Af
...
201        300 30437  0 00:11 ?        00:00:00 [timeout] <defunct>
201        301 30437  0 May15 ?        00:00:00 [timeout] <defunct>
201        302 30437  0 02:28 ?        00:00:00 [timeout] <defunct>
201        303 30437  0 08:26 ?        00:00:00 [timeout] <defunct>
201        304 30437  0 04:26 ?        00:00:00 [timeout] <defunct>
201        305 30437  0 12:39 ?        00:00:00 [timeout] <defunct>
...(repeated thousands of times)

-- WTH is that?
# ps -Af | grep 30437
201      30437 30412  1 May15 ?        00:14:16 /usr/sbin/netdata -u netdata -D -s /host -p 19999

-- What? Why is netdata spawning so many timeouts and why are they defuncting?

# ps -Af | grep timeout | wc -l
32127

-- Woah! netdata has spawned 32,000 timeouts! Kill it!

# kill -9 30437
# ps -Af | grep timeout
[...]

PR #7059, opened 2019-10-10
feat(reaper): Add process reaper support

stevenh:
##### Summary
Add a child process reaper to the main netdata app if running as init (pid = 1).

This prevents zombie processes when a child is re-parented to netdata when its running in a container.

Also:
* Few style cleanups to match surrounding code.

Fixes: #6033

##### Component Name
netdata binary

##### Additional Information
This re-purposes old commented out code in `popen.c`, which already implemented part of the required process tracking.

Without this on a standard netdata docker install we saw at least one zombie `timeout` process straight after the container was started.

### DIFF
--- a/daemon/signals.c
+++ b/daemon/signals.c
@@ -132,12 +132,12 @@ static void reap_child(pid_t pid) {
     siginfo_t i;
 
     errno = 0;
-    info("SIGNAL: Reaping pid: %d...", pid);
+    debug("SIGNAL: Reaping pid: %d...", pid);
     if (waitid(P_PID, (id_t)pid, &i, WEXITED|WNOHANG) == -1) {
         if (errno != ECHILD)
             error("SIGNAL: Failed to wait for: %d", pid);
         else
-            info("SIGNAL: Already reaped: %d", pid);
+            debug("SIGNAL: Already reaped: %d", pid);
         return;
     } else if (i.si_pid == 0) {
         // Process didn't exit, this shouldn't happen.
@@ -146,25 +146,25 @@ static void reap_child(pid_t pid) {
 
     switch (i.si_code) {
     case CLD_EXITED:
-        info("SIGNAL: Child %d exited: %d", pid, i.si_status);
+        debug("SIGNAL: Child %d exited: %d", pid, i.si_status);
         break;
     case CLD_KILLED:
-        info("SIGNAL: Child %d killed by signal: %d", pid, i.si_status);
+        debug("SIGNAL: Child %d killed by signal: %d", pid, i.si_status);
         break;
     case CLD_DUMPED:
-        info("SIGNAL: Child %d dumped core by signal: %d", pid, i.si_status);
+        debug("SIGNAL: Child %d dumped core by signal: %d", pid, i.si_status);
         break;
     case CLD_STOPPED:
-        info("SIGNAL: Child %d stopped by signal: %d", pid, i.si_status);
+        debug("SIGNAL: Child %d stopped by signal: %d", pid, i.si_status);
         break;
     case CLD_TRAPPED:
-        info("SIGNAL: Child %d trapped by signal: %d", pid, i.si_status);
+        debug("SIGNAL: Child %d trapped by signal: %d", pid, i.si_status);
         break;
     case CLD_CONTINUED:
-        info("SIGNAL: Child %d continued by signal: %d", pid, i.si_status);
+        debug("SIGNAL: Child %d continued by signal: %d", pid, i.si_status);
         break;
     default:
-        error("SIGNAL: Child %d gave us a SIGCHLD with code %d and status %d.", pid, i.si_code, i.si_status);
+        debug("SIGNAL: Child %d gave us a SIGCHLD with code %d and status %d.", pid, i.si_code, i.si_status);
     }
 }
 
@@ -250,7 +250,7 @@ void signals_handle(void) {
                                 fatal("SIGNAL: Received %s. netdata now exits.", name);
 
                             case NETDATA_SIGNAL_CHILD:
-                                info("SIGNAL: Received %s. Reaping...", name);
+                                debug("SIGNAL: Received %s. Reaping...", name);
                                 reap_children();
                                 break;
 
--- a/daemon/signals.c
+++ b/daemon/signals.c
@@ -132,12 +132,12 @@ static void reap_child(pid_t pid) {
     siginfo_t i;
 
     errno = 0;
-    debug("SIGNAL: Reaping pid: %d...", pid);
+    debug(D_CHILDS, "SIGNAL: Reaping pid: %d...", pid);
     if (waitid(P_PID, (id_t)pid, &i, WEXITED|WNOHANG) == -1) {
         if (errno != ECHILD)
             error("SIGNAL: Failed to wait for: %d", pid);
         else
-            debug("SIGNAL: Already reaped: %d", pid);
+            debug(D_CHILDS, "SIGNAL: Already reaped: %d", pid);
         return;
     } else if (i.si_pid == 0) {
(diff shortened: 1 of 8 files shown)


libnetdata/popen/popen.c
```diff
@@ -128,6 +149,60 @@ static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, c
...
+void myp_free(void) {
+    struct mypopen *mp, *next;
+
+    if (myp_tracking == 0)
+        return;
+
+    netdata_mutex_lock(&myp_lock);
+    for (mp = mypopen_root; mp != NULL; mp = next) {
+        next = mp->next;
+        free(mp);
```
mfundul: since we will change `malloc()` to `mallocz()` this should become `freez()`

daemon/signals.c
```diff
@@ -157,6 +229,11 @@ void signals_handle(void) {
                             case NETDATA_SIGNAL_FATAL:
                                 fatal("SIGNAL: Received %s. netdata now exits.", name);
 
+                            case NETDATA_SIGNAL_CHILD:
+                                if (reaper_enabled == 1)
+                                    reap_child();
```
mfundul: I think that at this point, there may have been multiple `SIGCHLD` signals before `signals_waiting[i].count = 0;` was set.

I think a single `reap_child()` call is not enough and that we need a loop. What do you think?

libnetdata/popen/popen.c
```diff
@@ -2,46 +2,66 @@
...
+    netdata_mutex_lock(&myp_lock);
+    for (mp = mypopen_root; mp != NULL; mp = mp->next) {
+        if (mp->pid == pid) {
+            if (mp->next != NULL)
+                mp->next->prev = mp->prev;
+            if (mp->prev != NULL)
+                mp->prev->next = mp->next;
+            if (mypopen_root == mp)
+                mypopen_root = mp->next;
+            free(mp);
```
mfundul: since we will change `malloc()` to `mallocz()` this should become `freez()`

libnetdata/popen/popen.c
```diff
@@ -2,46 +2,66 @@
...
+            break;
+        }
     }
+
+    if (mp != NULL)
```
mfundul: Shouldn't that be `if (mp == NULL)` ?

daemon/signals.c
stevenh: Yes indeed, didn't spot this wasn't the actual signal handler.

I would prefer to actual rework the signal handlers to be inline, otherwise you've going to have to deal with races. This would should also be more performant thoughts?

ilyam8:
@stevenh i would suggest to not force-push w/o reason, it is not clear what is fixed and we have to review from scratch.

daemon/signals.c
mfundul: I'm afraid that is not going to work because in the signal handlers themselves you can do almost nothing, and netdata already does **a lot** of heavy lifting in `signals_handle()`. You cannot even use mutexes such as those from `myp_reap()` in `reap_child()`. See [here](http://man7.org/linux/man-pages/man7/signal-safety.7.html) for reference.

`signals_handle()` is being called by the netdata daemon master thread, that is possibly the one with `pid == 1` in the container case, and it's devoted to exclusively handle signals after the initialization. Due to the limitations of signal safety the scope of a perfect solution would be too big in relation to the motivation behind this pull request.

I would prefer you figure out  a way to catch the signals without making a lot of changes.

libnetdata/popen/popen.c
```diff
@@ -2,46 +2,66 @@
...
-    if(!mp) {
-        fatal("Cannot allocate %zu bytes", sizeof(struct mypopen))
+static void myp_add(pid_t pid) {
+    struct mypopen *mp;
+
+    if (myp_tracking == 0)
+        return;
+
+    mp = mallocz(sizeof(struct mypopen));
+    if (mp == NULL) {
```
mfundul: Forgot to mention, `mallocz()` already does the `fatal()` inside so you can remove this `if` statement.

daemon/signals.c
stevenh: Hmm actually looks like this was a conscious decision, to avoid long running signal handlers. I will take a different approach.

stevenh: Yep already spotted that when investigating the options, but thanks for the heads up.

stevenh:
Thanks for the feedback so far guys, always good to get to know the nuances of project, no force pushes. 👍 

I've addressed all the feedback, fixing a few other niggles along the way, as well as eliminating some robustness issues which became apparent when stress testing the for multiple children exiting; i.e. the data race when setting pid.

daemon/signals.c
```diff
@@ -100,6 +122,76 @@ void signals_reset(void) {
...
+            if (errno != ECHILD) // This shouldn't happen with WNOHANG but does.
+                error("SIGNAL: Failed to wait");
+            return;
+        } else if (i.si_pid == 0) {
+            // No child exited.
+            return;
+        } else if (myp_reap(i.si_pid) == 0) {
+            // myp managed, sleep for a short time to avoid busy wait while
+            // this is handled by myp.
+            usleep(10000);
```
mfundul: I really hate this but I don't see a way around it without rewriting the daemon.

mfundul: A potential problem here is that all signal handling is effectively blocked until somebody calls `mypclose()`, such as health reload or exiting.

daemon/signals.c
```diff
@@ -100,6 +122,76 @@ void signals_reset(void) {
...
+        break;
+    case CLD_TRAPPED:
+        info("SIGNAL: Child %d trapped by signal: %d", pid, i.si_status);
+        break;
+    case CLD_CONTINUED:
+        info("SIGNAL: Child %d continued by signal: %d", pid, i.si_status);
+        break;
+    default:
+        error("SIGNAL: Child %d gave us a SIGCHLD with code %d and status %d.", pid, i.si_code, i.si_status);
+    }
```
mfundul: I'm not sure we want to print to the netdata log all reaped processes of a container that happen to be re-parented to the netdata process. I feel like this should be a silent operation.

daemon/signals.c
```diff
@@ -100,6 +122,76 @@ void signals_reset(void) {
...
+    if (reaper_enabled == 1)
+        myp_free();
+}
+
+// reap_child reaps the child identified by pid.
+static void reap_child(pid_t pid) {
+    siginfo_t i;
+
+    errno = 0;
+    info("SIGNAL: Reaping pid: %d...", pid);
```
mfundul: Same as below, I think we should remove the prints.

stevenh: How about switching to debug, as they may be useful if there's a bad plugin or something?

daemon/signals.c
stevenh: Indeed wasn't a fan either, the only thing I thought of was to have this handled in a separate thread; seems like this could be a future enhancement?

Shame its not in `go` as that would just be a few lines of code, not so simple in `C` though ;-)

mfundul: Even in `go` I don't think it's guaranteed that `waitid(P_ALL, (id_t)0, &i, WEXITED|WNOHANG|WNOWAIT)` will return any other pid than the one being handled by `myp` which would not allow the loop to iterate.

daemon/signals.c
mfundul: Sure, you can switch to debug.

daemon/signals.c
stevenh: I actually meant its much easier to throw up a `goroutine` to handle each type of signal, not that it solves the wait issue.

cakrit:
We appreciate the work @reaper!!!

ilyam8:
@stevenh the reaper 😄

Issue #7039, opened 2019-10-09
Kickstart-static64.sh doesn't install netdata-updater cron


rdf6:
##### Bug report summary
When installing the static binary, it does not install the netdata-updater into /etc/cron.daily.

##### OS / Environment
cat /etc/os-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.6 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.6"
PRETTY_NAME="Red Hat Enterprise Linux"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.6:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.6
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.6"

##### Netdata version (ouput of `netdata -V`)
netdata -V
netdata v1.17.1

##### Component Name
packaging/installer

##### Steps To Reproduce
bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh) --dont-wait --stable-channel
...
ls -al /etc/cron.daily/netdata-updater
ls: cannot access /etc/cron.daily/netdata-updater: No such file or directory

##### Expected behavior
Should install the netdata-updater into /etc/cron.daily, just like it does when using kickstart.sh.  I do see that the following file does exist, but isn't being copied to cron.daily:

/opt/netdata/usr/libexec/netdata/netdata-updater.sh

cakrit:
As per [the documentation](https://docs.netdata.cloud/packaging/installer/update/#you-downloaded-a-binary-package), this isn't actually a bug, but a very good feature request.

knatsakis:
The documentation could use some improvement. [Here](https://docs.netdata.cloud/packaging/installer/#one-line-installation), it mentions:

> To install Netdata from a binary package on any Linux distro and any kernel version on Intel/AMD 64bit systems, and keep it up to date with our nightly releases automatically, run the following:
>
> `bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh)`

rdf6:
[...]

PR #7060, opened 2019-10-11
Partial fix for #7039

knatsakis:
##### Summary
Partial fix for #7039

- Moved updater installation code from netdata-installer.sh to packaging/installer/functions.sh
- packaging/makeself/install-or-update.sh uses above code to install
netdata updater

##### Component Name
packaging/makeself/install-or-update.sh

##### Additional Information
Another PR will be needed, in order to update kickstart-static64.sh and enable the netdata installer by default

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -1050,63 +1050,13 @@ END
 echo >&2 "Uninstall script copied to: ${TPUT_RED}${TPUT_BOLD}${NETDATA_PREFIX}/usr/libexec/netdata/netdata-uninstaller.sh${TPUT_RESET}"
 echo >&2
 
+# -----------------------------------------------------------------------------
 progress "Install netdata updater tool"
 
-if [ -f "${NETDATA_PREFIX}"/usr/libexec/netdata-updater.sh ]; then
-	echo >&2 "Removing updater from previous location"
-	rm -f "${NETDATA_PREFIX}"/usr/libexec/netdata-updater.sh
-fi
+install_or_remove_netdata_updater || run_failed "Cannot install netdata updater tool."
 
-if [ -f "${INSTALLER_DIR}/packaging/installer/netdata-updater.sh" ]; then
-	sed "s|THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT|${NETDATA_USER_CONFIG_DIR}/.environment|" "${INSTALLER_DIR}/packaging/installer/netdata-updater.sh" > "${NETDATA_PREFIX}/usr/libexec/netdata/netdata-updater.sh" || exit 1
-else
-	sed "s|THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT|${NETDATA_USER_CONFIG_DIR}/.environment|" "${NETDATA_SOURCE_DIR}/packaging/installer/netdata-updater.sh" > "${NETDATA_PREFIX}/usr/libexec/netdata/netdata-updater.sh" || exit 1
-fi
 
-chmod 0755 ${NETDATA_PREFIX}/usr/libexec/netdata/netdata-updater.sh
-echo >&2 "Update script is located at ${TPUT_GREEN}${TPUT_BOLD}${NETDATA_PREFIX}/usr/libexec/netdata/netdata-updater.sh${TPUT_RESET}"
-echo >&2
-
-# Figure out the cron directory for the distro
-crondir=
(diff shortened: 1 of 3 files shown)


packaging/installer/functions.sh
```diff
@@ -727,3 +727,64 @@ safe_sha256sum() {
...
+	if [ -f "${NETDATA_PREFIX}"/usr/libexec/netdata-updater.sh ]; then
+		echo >&2 "Removing updater from previous location"
+		rm -f "${NETDATA_PREFIX}"/usr/libexec/netdata-updater.sh
+	fi
+
+	if [ "${INSTALLER_DIR}" ] && [ -f "${INSTALLER_DIR}/packaging/installer/netdata-updater.sh" ]; then
+		cat "${INSTALLER_DIR}/packaging/installer/netdata-updater.sh" > "${NETDATA_PREFIX}/usr/libexec/netdata/netdata-updater.sh" || exit 1
+	fi
+
+	if [ "${NETDATA_SOURCE_DIR}" ] && [ -f "${NETDATA_SOURCE_DIR}/packaging/installer/netdata-updater.sh" ]; then
```
cakrit: kickstart-static64.sh doesn't define `INSTALLER_DIR` and `NETDATA_SOURCE_DIR`. Has this been tested?

knatsakis: @chris, This is the reason I changed the logic within `install_or_remove_netdata_updater` compared to what it was in `netdata-installer.sh`

Before it was
```
if [ -f "${INSTALLER_DIR}/packaging/installer/netdata-updater.sh" ]; then
```

Now it is
```
if [ "${INSTALLER_DIR}" ] && [ -f "${INSTALLER_DIR}/packaging/installer/netdata-updater.sh" ]; then
```

I have also tested this by running the installation on a clean VM

packaging/installer/functions.sh
```diff
@@ -727,3 +727,64 @@ safe_sha256sum() {
 		fatal "I could not find a suitable checksum binary to use"
 	fi
 }
+
+install_or_remove_netdata_updater() {
```
cosmix: I'm not fond of these kinds of functions that 'decide' on their own what to do. I may want to change the logic. I may want to do something non-standard. If you want to have this kind of functionality you should — at the very least — create functions that `remove` and `install` and then invoke them from your `do_this_or_not_this_function()`. This way, you can easily compartmentalize the actual functionality from whatever logic you're using to manage your updater install.

packaging/installer/functions.sh
cakrit: Help me understand. Since both variables are empty, how do we end up with a file in `${NETDATA_PREFIX}/usr/libexec/netdata/netdata-updater.sh` when calling this from `kickstart-static64.sh`?

knatsakis: When using kickstart.sh the netdata-updater.sh script is not installed by make install, so this function has to copy it.

But when using kickstart-static64.sh the full .tar.gz is extracted under /opt/netdata (which is NETDATA_PREFIX in this case), and the netdata-updater.sh script is already under NETDATA_PREFIX.

packaging/installer/functions.sh
paulkatsoulakis: +1 on this

netdata-installer.sh
```diff
@@ -1050,63 +1050,20 @@ END
...
 echo >&2
 
-progress "Install netdata updater tool"
-
-if [ -f "${NETDATA_PREFIX}"/usr/libexec/netdata-updater.sh ]; then
-	echo >&2 "Removing updater from previous location"
-	rm -f "${NETDATA_PREFIX}"/usr/libexec/netdata-updater.sh
-fi
+# -----------------------------------------------------------------------------
+progress "Install (but not enable) netdata updater tool"
```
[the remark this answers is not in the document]
knatsakis: Resolved in #7200

packaging/installer/functions.sh
```diff
@@ -727,3 +727,96 @@ safe_sha256sum() {
...
+get_crondir() {
+	crondir=
+	[ -d "/etc/periodic/daily" ] && crondir="/etc/periodic/daily"
+	[ -d "/etc/cron.daily" ] && crondir="/etc/cron.daily"
+
+	echo "${crondir}"
+}
+
+check_crondir_permissions() {
+	if [ -z "${1}" ]; then
```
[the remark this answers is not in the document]
knatsakis: Resolved in #7200

PR #7078, opened 2019-10-13
fix(freeipmi): Update frequency config check

stevenh:
##### Summary
Fix the update frequency check for freeipmi which was warning that 5 was too frequent an it was setting it to 5.

##### Component Name
collectors/freeipmi


ilyam8:
thx @stevenh

PR #7090, opened 2019-10-14
Clang warnings

### LINKED ISSUE #7089: Clang issues
##### Bug report summary
While I was working with FreeBSD, I noticed that there are various warning messages, good part of them are related to the parameter `-L/usr/local/lib`, but there were few warnings related to codes that I am showing in the next images:

Warning0

Warning1

##### OS / Environment
FreeBSD 12.0-RELEASE-p10

netdata v1.18.0-21-g909fce2e
##### Component Name
Netdata installer
##### Steps To Reproduce
Run the command `netdata-installer.sh` on FreeBSD
##### Expected behavior
Do not have more these warnings.

thiagoftsm:
##### Summary
Fixes #7089

This PR brings the necessaries fixes during the compilation time with clang,
##### Component Name
Netdata installer.

### DIFF
--- a/daemon/global_statistics.c
+++ b/daemon/global_statistics.c
@@ -156,7 +156,7 @@ static inline void global_statistics_copy(struct global_statistics *gs, uint8_t 
 
     if(options & GLOBAL_STATS_RESET_WEB_USEC_MAX) {
         uint64_t n = 0;
-        __atomic_compare_exchange((uint64_t *)&global_statistics.web_usec_max, (uint64_t *)&gs->web_usec_max, &n, 1, __ATOMIC_SEQ_CST,
+        __atomic_compare_exchange(&global_statistics.web_usec_max, &gs->web_usec_max, &n, 1, __ATOMIC_SEQ_CST,
                                   __ATOMIC_SEQ_CST);
     }
 #else
--- a/daemon/global_statistics.c
+++ b/daemon/global_statistics.c
@@ -156,7 +156,7 @@ static inline void global_statistics_copy(struct global_statistics *gs, uint8_t 
 
     if(options & GLOBAL_STATS_RESET_WEB_USEC_MAX) {
         uint64_t n = 0;
-        __atomic_compare_exchange(&global_statistics.web_usec_max, &gs->web_usec_max, &n, 1, __ATOMIC_SEQ_CST,
+        __atomic_compare_exchange((uint64_t *)&global_statistics.web_usec_max, (uint64_t *)&gs->web_usec_max, &n, 1, __ATOMIC_SEQ_CST,
                                   __ATOMIC_SEQ_CST);
     }
 #else
--- a/libnetdata/health/health.c
+++ b/libnetdata/health/health.c
@@ -112,7 +112,7 @@ int health_silencers_json_read_callback(JSON_ENTRY *e)
         case JSON_OBJECT:
 #ifndef ENABLE_JSONC
             e->callback_function = health_silencers_json_read_callback;
-            if(e->name && strcmp(e->name,"")) {
+            if(strcmp(e->name,"")) {
                 // init silencer
                 debug(D_HEALTH, "JSON: Got object with a name, initializing new silencer for %s",e->name);
 #endif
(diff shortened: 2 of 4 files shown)


thiagoftsm:
> I agree with your reasoning about the unnecessary checks: the double `(( ... )))` is redundant, and the `->name` and `->fullname` fields cannot be null as they are fields inside the structure.
> 
> There seem to be some issues in the code that is already there.
> 
> Why is `ne` being initialized in L293-L298 but then overwritten by the `memcpy()` in L304? If this initialization is redundant then it should be removed.
> 
> The docstring above the procedure says that `e` is the output structure, but it is being read from. Could you check this please?

This file is an old contribution from our users, I removed the initial set, because it is useless as you said and I updated the comments.

Thank you very much to point these necessary changes.

amoss:
I pulled out our nightly on FreeBSD12 and checked the build. I also see this warning in addition to the ones that you are fixing, is this on your system as well?
```
daemon/global_statistics.c:159:68: warning: passing 'volatile uint64_t *' (aka 'volatile unsigned long *') to parameter of type 'uint64_t *'
      (aka 'unsigned long *') discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        __atomic_compare_exchange(&global_statistics.web_usec_max, &gs->web_usec_max, &n, 1, __ATOMIC_SEQ_CST,
```

thiagoftsm:
> I think that you've fixed the warnings and improved the code, but there may be a longer term issue here for us to think about. Reviewing the build-system I don't think that json-c is a soft-dependency any more. It seems to have become a hard dependency over time (i.e. that fix for the pfSense install). We should decide if it is a hard- or soft- requirement for the install and if it is then we should drop all of the code that is `#ifndef JSONC` in this file.
> 
> I've approved the review because I don't think that we should make further changes now - but we should think about it after the build system moves to CMake.

You are correct, we need to define an unique JSON library, high now we are using use either json-c or jsmn  and our internal codes, it would be good to define only one.

thiagoftsm:
> I pulled out our nightly on FreeBSD12 and checked the build. I also see this warning in addition to the ones that you are fixing, is this on your system as well?
> 
> ```
> daemon/global_statistics.c:159:68: warning: passing 'volatile uint64_t *' (aka 'volatile unsigned long *') to parameter of type 'uint64_t *'
>       (aka 'unsigned long *') discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
>         __atomic_compare_exchange(&global_statistics.web_usec_max, &gs->web_usec_max, &n, 1, __ATOMIC_SEQ_CST,
> ```

I missed this one, I brought now the solution.
Thank you very much!

amoss:
I've had a long look at it but I'm not sure it is a good idea. As far as I understand `__atomic_compare_exchange` is a macro that after configuration will land on either the clang intrinsic or the system-defined version. The first argument to each of these possibilities is a `volatile uint64 *` so I don't understand why we see the warning or why we fix it by casting away the `volatile` label.

I would hesitate to change something this subtle without understanding it much better than I currently do. I don't think that the code matches the intent. The block in the `#else` case is easy to understand: copy the structure, if the option is set then clear the counter. But in the atomic case why are we comparing to `gs->web_usec_max` before it is initialised? Would it not make more sense to do an atomic swap without a comparison, swapping with the zero value in n and then putting the value retrieved into `gs` so that the count is preserved?

I think that this review need more than one pair of eyes. Please get another review from somebody who knows this code first, and then ping me again so that we can discuss it.

thiagoftsm:
I understand what you said @amoss , it is good to remember that is a warning given only from clang, we never had this with gcc.

amoss:
Yes that is true, and I wonder if the cast will create a warning message on gcc if the type signature is different.

thiagoftsm:
I compiled with GCC and I did not have any problem, case you agree, I can remove this specific warning and we can treat it in a separate PR. what is your opinion considering that you found it?

thiagoftsm:
I returned for the previous situation where I did not bring fix for this warning.

PR #7102, opened 2019-10-15
Fix segmentation fault in FreeBSD when statsd is disabled

### LINKED ISSUE #7064: Disabling statsd causes a segfault in FreeBSD 12.
##### Bug report summary
Disabling statsd causes a segfault in FreeBSD 12.
##### OS / Environment
FreeBSD 12
netdata v1.18.0
##### Component Name
statsd
##### Steps To Reproduce
edit `netdata.conf` and set:
```
[statsd]
	enabled = no
```
and run `./netdata -D`
##### Expected behavior
not crash

vlvkobal:
##### Summary
Netdata was crashing on FreeBSD when statsd was disabled due to the 
peculiarities of `pthread_cleanup_push` behavior in FreeBSD.

Fixes #7064

##### Component Name
statsd plugin

### DIFF
--- a/collectors/statsd.plugin/statsd.c
+++ b/collectors/statsd.plugin/statsd.c
@@ -2222,7 +2222,7 @@ void *statsd_main(void *ptr) {
     // ----------------------------------------------------------------------------------------------------------------
     // statsd setup
 
-    if(!statsd.enabled) return NULL;
+    if(!statsd.enabled) goto cleanup;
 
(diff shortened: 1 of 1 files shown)


stevenh:
Looks good to me `pthread_cleanup_push` and `pthread_cleanup_pop` which underpin the `netdata` versions should always be matching and in the same lexical scope, so the original return NULL was a clear violation.

Issue #6749, opened 2019-08-31
netdata/ci: Wrap up coverity script refinement, make it usable, extend CI pipeline compile checks


paulkatsoulakis:
##### Bug report summary
Currently coverity scans were run exclusively from Travis, resulting on the absence of proper developer toolkit for coverity scanning

As part of this issue we need to:
- [x] wrap up #6747 with a new PR to attend comments and inputs found in #6747

~Update CI pipeline to execute multiple versions of the installer with different configure flags~

- [ ] Make coverity scan run with most (if not all) of the configure flags available for maximum coverage as needed

##### OS / Environment
Any

##### Netdata version (ouput of `netdata -V`)
N/A

##### Component Name
netdata/ci

##### Expected behavior
Coverity should be usable by developers, the CI pipeline that runs coverity should cover all functional cases and once this is resolved C developers should have coverity on their daily agenda

paulkatsoulakis:
Branch `fix-coverity-toolkit` on my fork has the work for this, i 'll be working on it today.
I had to deal with Athens municipality this morning and that got me delayed

paulkatsoulakis:
xenstat plugin proved to be more complex that i thought.
We don't seem to have worked with this on distros other than NCP-ng where the initial development took place.

From the tests i did on Ubuntu 18.04 and Cent OS 7 distributions (containerised), i could not get netdata to compile successfully with any of the available packages.

The error i couldn't get through with is:
```
make[2]: Leaving directory `/netdata/web'
make[2]: Entering directory `/netdata'
  CCLD     xenstat.plugin
/tmp/ccm39i63.ltrans0.ltrans.o: In function `vbd_metrics_collect.isra.2':
/netdata/collectors/xenstat.plugin/xenstat_plugin.c:336: undefined reference to `xenstat_vbd_error'
collect2: error: ld returned 1 exit status
make[2]: *** [xenstat.plugin] Error 1
make[2]: Leaving directory `/netdata'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/netdata'
make: *** [all] Error 2
[root@80ea85095f33 netdata]# history
```

After discussing with @vlvkobal, it seems that we need to work with the following repository to get things running for netdata, rather than using an existing binary package. At least i wasn't able to spot which binary package would cover us.

https://github.com/xen-project/xen

paulkatsoulakis:
[...]

PR #7103, opened 2019-10-15
xenstat.plugin: check xenstat_vbd_error presence

arkamar:
##### Summary

The `xenstat_vbd_error` function is going to be part of the **xen 4.13**
release. The workaround here is to check if the function is already
present in configuration phase and set `vbd_m->error` always to `0` if
`xenstat_vbd_error` does not exist.

##### Component Name

collectors/xenstat.plugin

##### Additional Information

This PR is related to #6749


vlvkobal:
I confirm, `xenstat_vbd_error` was introduced in https://github.com/xen-project/xen/commit/91c3e3dc91d614d3385d69d03e2dbefbe8abced4

The PR looks good to me. I don't have `xenstat` installed to test it though.

arkamar:
Thanks!

Issue #7111, opened 2019-10-16
python: Unbound Collector unhandled exception


mkoubar:
python.d WARNING: plugin[main] : unbound[unbound] : unhandled exception on init : AttributeError("'NoneType' object has no attribute 'get'",), skipping the job

ilyam8:
hi @mkg20001 

please do and share the output


> ./python.d.plugin debug trace unbound

mkoubar:
root@develser02:~# /usr/libexec/netdata/plugins.d/python.d.plugin unbound debug trace 1
2019-10-16 13:13:08: python.d INFO: plugin[main] : using python v2
2019-10-16 13:13:08: python.d DEBUG: plugin[main] : looking for 'python.d.conf' in ['/etc/netdata', '/usr/lib/netdata/conf.d']
2019-10-16 13:13:08: python.d DEBUG: plugin[main] : loading '/usr/lib/netdata/conf.d/python.d.conf'
2019-10-16 13:13:08: python.d DEBUG: plugin[main] : '/usr/lib/netdata/conf.d/python.d.conf' is loaded
2019-10-16 13:13:08: python.d DEBUG: plugin[main] : looking for 'pythond-jobs-statuses.json' in /var/lib/netdata
2019-10-16 13:13:08: python.d DEBUG: plugin[main] : loading '/var/lib/netdata/pythond-jobs-statuses.json'
2019-10-16 13:13:08: python.d DEBUG: plugin[main] : '/var/lib/netdata/pythond-jobs-statuses.json' is loaded
2019-10-16 13:13:09: python.d DEBUG: plugin[main] : [unbound] looking for 'unbound.conf' in ['/etc/netdata/python.d', '/usr/lib/netdata/conf.d/python.d']
2019-10-16 13:13:09: python.d DEBUG: plugin[main] : [unbound] loading '/usr/lib/netdata/conf.d/python.d/unbound.conf'
2019-10-16 13:13:09: python.d DEBUG: plugin[main] : [unbound] '/usr/lib/netdata/conf.d/python.d/unbound.conf' is loaded
2019-10-16 13:13:09: python.d INFO: plugin[main] : [unbound] built 1 job(s) configs
2019-10-16 13:13:09: python.d DEBUG: unbound[unbound] : No unix socket specified. Trying TCP/IP socket.
2019-10-16 13:13:09: python.d DEBUG: unbound[unbound] : No host specified. Using: "localhost"
[...]

PR #7112, opened 2019-10-16
unbound: fix init

ilyam8:
##### Summary

fixes: #7111

This PR:
 - fixes `ssl.PROTOCOL_TLS` import (added in py 2.7.13, it deprecates `ssl.PROTOCOL_SSLv23`)
 - fixes unhandled `IOError`, `OSError` exceptions in `SocketService._connect2socket` method during `ssl.wrap_socket` (it happens if ssl key or cert is not exists/readable)
 - adds two checks to the unbound module check method: check if ssl key is readable, check if ssl cert is readable
 - moves load unbound config logic to a sep method

##### Component Name

[/collectors/python.d.plugin/unbound/unbound](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/unbound)

### DIFF
--- a/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService.py
+++ b/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SocketService.py
@@ -80,6 +80,9 @@ class SocketService(SimpleService):
         if self.tls:
             try:
                 self.debug('Encapsulating socket with TLS')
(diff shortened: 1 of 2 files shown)


ilyam8:
@mkoubar please test the PR

ilyam8:
the bug is fixed
> https://github.com/netdata/netdata/issues/7111#issuecomment-543040530

PR #7121, opened 2019-10-17
Convert recursion timings to miliseconds.

### LINKED ISSUE #7120: unbound.chart.py Recursion Timings
@ilyam8 
dns-time

Ferroin:
##### Summary
This avoids them being mangled to the point of being useless when the
locale settings are configured to convert seconds to times.

These were originally implemented as seconds because that's how Unound
itself reports them.

##### Component Name

collectors/python.d/unbound

##### Additional Information

Fixes #7120 

Shows no issues in testing, though the switch-over doesn't drop old data or convert it properly (that's probably a bug elsewhere though, and I don't feel it should block this PR).

### DIFF
--- a/collectors/python.d.plugin/unbound/unbound.chart.py
+++ b/collectors/python.d.plugin/unbound/unbound.chart.py
@@ -32,10 +32,10 @@ CHARTS = {
         ]
     },
     'recursion': {
-        'options': [None, 'Recursion Timings', 'seconds', 'Unbound', 'unbound.recursion', 'line'],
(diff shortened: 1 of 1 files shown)


collectors/python.d.plugin/unbound/unbound.chart.py
```diff
@@ -32,10 +32,10 @@
         ]
     },
     'recursion': {
-        'options': [None, 'Recursion Timings', 'seconds', 'Unbound', 'unbound.recursion', 'line'],
+        'options': [None, 'Recursion Timings', 'miliseconds', 'Unbound', 'unbound.recursion', 'line'],
```
ilyam8: a typo

> milliseconds

PR #6989, opened 2019-10-04
mysql: add cluster_status alarm

ilyam8:
##### Summary

galera cluster status alarm suggested by @ekexcello in https://github.com/netdata/netdata/pull/6962#issuecomment-537002612

Added as a separate PR because the alarm is false positive until #6980 is fixed.

Should be merged after #6984

##### Component Name

[health/health.d/mysql.conf](https://github.com/netdata/netdata/blob/master/health/health.d/mysql.conf)

PR #7139, opened 2019-10-20
detect if the disk cannot keep up with data collection

### LINKED ISSUE #7043: Generate alarms if the disk cannot keep up with data collection
##### Feature idea summary
If the disk cannot keep up with metric collection this condition should be detectable and alarms should be generated. This is for `memory mode = dbengine`.

mfundul:
##### Summary
Fixes #7043 
##### Component Name
database/engine


health/health.d/dbengine.conf
```diff
@@ -23,4 +23,16 @@ lookup: sum -10m unaligned of I/O errors
...
+    on: netdata.dbengine_global_errors
+    os: linux freebsd macos
+ hosts: *
+lookup: sum -10m unaligned of flushing errors
+ units: errors
+ every: 3s
+  crit: $this > 0
+ delay: down 1h multiplier 1.5 max 3h
+  info: number of times dbengine failed to flush in time the last 10 minutes, metric data will not be stored in the database, please reduce disk load or use a faster disk
     to: sysadmin
```
thiagoftsm: Please add a new line at the end of this file.

configs.signatures
```diff
@@ -381,7 +381,7 @@ declare -A configs_signatures=(
   ['7deb236ec68a512b9bdd18e6a51d76f7']='python.d/mysql.conf'
   ['7e5fc1644aa7a54f9dbb1bd102521b09']='health.d/memcached.conf'
   ['7f13631183fbdf79c21c8e5a171e9b34']='health.d/zfs.conf'
-  ['e48b89d4a97b96acf9a88970ab858c3b']='health.d/dbengine.conf'
+  ['f9ea512f96fc3971e1314d790e9a96ea']='health.d/dbengine.conf'
```
ilyam8: We shouldnt change the signatures by hand.

mfundul: Is there a way to do it automatically? :D

ilyam8: Well. If i remember it correctly this file is not used at all. We always rewrite stock config, we never rewrite user config. These signatures were used before stock/user configs separation.

See #6989. I updated `mysql.conf` there and i changed no singnatures.

ilyam8: @knatsakis correct me i i am wrong

health/health.d/dbengine.conf
```diff
@@ -23,4 +23,16 @@ lookup: sum -10m unaligned of I/O errors
...
+ alarm: 10min_dbengine_global_flushing_errors
+    on: netdata.dbengine_global_errors
+    os: linux freebsd macos
+ hosts: *
+lookup: sum -10m unaligned of flushing errors
+ units: errors
+ every: 3s
+  crit: $this > 0
+ delay: down 1h multiplier 1.5 max 3h
+  info: number of times dbengine failed to flush in time the last 10 minutes, metric data will not be stored in the database, please reduce disk load or use a faster disk
```
Ferroin: Phrasing suggestion for this:

'number of times in the last 10 minutes that the dbengine failed to completely flush data to disk'

Flows a bit better from the perspective of a native AmE speaker and more concretely explains the exact conditions that resulted in the alarm.

Ferroin:
> I retest the PR with the last update, I could not raise the alarm again, but everything is working fine with it.

I ended up testing this one myself this time.  I was able to get the alarm to trigger by running in a VM with the storage directory being mounted over NFS with severe bandwidth limiting and intentionally using highly inefficient mount options for NFS itself.  It seems to me that this is only likely to trigger when you've got particularly slow storage and insane numbers of metrics (well into the hundreds of thousands, possibly more), though it's definitely good to have it.

PR #7149, opened 2019-10-22
Implement the main flow for the Exporting Engine

### LINKED ISSUE #6956: Implement the main flow for the Exporting Engine
Implement the high level flow for exporting engine based on design described in #6769.

vlvkobal:
##### Summary
Closes #6956

##### Component Name
exporting engine (former name "backends")


exporting/init_connectors.c
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Initialize connectors
+ *
+ * @param engine an engine data structure.
+ * @return Returns 0 on success, 1 on failure.
+ */
+int init_connectors(struct engine *engine)
+{
+    for (struct connector *connector = engine->connector_root; connector; connector = connector->next) {
```
thiagoftsm: I am sorry, I do not remember, did we agree that we would allow variable declaration inside `for` loop?

thiagoftsm: Next file for example, you are creating the variable outside.

exporting/init_connectors.c
```diff
@@ -0,0 +1,41 @@
...
+        switch (connector->config.type) {
+            case BACKEND_TYPE_GRAPHITE:
+                if (init_graphite_connector(connector) != 0)
+                    return 1;
+                break;
+            default:
+                error("EXPORTING: unknown exporting connector type");
+                return 1;
+        }
+        for (struct instance *instance = connector->instance_root; instance; instance = instance->next) {
```
thiagoftsm: The same doubt about I have for this region.

exporting/process_data.c
```diff
@@ -0,0 +1,270 @@
...
+ * Substitute '_' for any special character except '.'.
+ *
+ * @param dst where to copy name to.
+ * @param src where to copy name from.
+ * @param max_len the maximum size of copied name.
+ * @return Returns the size of the copied name.
+ */
+size_t exporting_name_copy(char *dst, const char *src, size_t max_len)
+{
+    size_t n;
```
thiagoftsm: This is the unique place of the file that you have a variable created outside the loop.

exporting/tests/test_exporting_engine.c
```diff
@@ -0,0 +1,401 @@
...
+    connector->next = NULL;
+    connector->start_batch_formatting = __mock_start_batch_formatting;
+    connector->start_host_formatting = __mock_start_host_formatting;
+    connector->start_chart_formatting = __mock_start_chart_formatting;
+    connector->metric_formatting = __mock_metric_formatting;
+    connector->end_chart_formatting = __mock_end_chart_formatting;
+    connector->end_host_formatting = __mock_end_host_formatting;
+    connector->end_batch_formatting = __mock_end_batch_formatting;
+
+    connector->instance_root = (struct instance *)malloc(sizeof(struct instance));
```
thiagoftsm: Is it not safe to use `callocz` here?

exporting/tests/test_exporting_engine.c
```diff
@@ -0,0 +1,401 @@
...
+static void test_prepare_buffers(void **state)
+{
+    (void)state;
+
+    struct engine *engine = (struct engine *)malloc(sizeof(struct engine));
+    memset(engine, 0xD1, sizeof(struct engine));
+    engine->after = 1;
+    engine->before = 2;
+
+    engine->connector_root = (struct connector *)malloc(sizeof(struct connector));
```
thiagoftsm: please, use `mallocz` here.

exporting/init_connectors.c
vlvkobal: Yes, we agreed that we will allow all c99 features in our code on the last meetup.

.clang-format
```diff
@@ -58,7 +59,7 @@ CommentPragmas: '^ IWYU pragma:'
 ConstructorInitializerAllOnOneLineOrOnePerLine: false
 ConstructorInitializerIndentWidth: 4
 ContinuationIndentWidth: 4
-Cpp11BracedListStyle: false
+Cpp11BracedListStyle: true
```
ilyam8: Shouldnt it be in a separate PR?

exporting/process_data.c
vlvkobal: Thanks, it's just a copy-paste of the old code. I'll fix it.

exporting/tests/test_exporting_engine.c
vlvkobal: It's safe. I'll review/rearrange tests later.

exporting/tests/test_exporting_engine.c
vlvkobal: We don't need `<anything>z()` functions in tests.

.clang-format
vlvkobal: Yes. It should. I'll do a separate PR when I'm content with automatic formatting.

exporting/init_connectors.c
thiagoftsm: Thank you!

exporting/tests/test_exporting_engine.c
thiagoftsm: All right!

libnetdata/config/appconfig.c
```diff
@@ -42,6 +44,78 @@ struct section {
...
+}
+
+int is_valid_connector(char *type){
+    if (unlikely(!type))
+        return 0;
+
+    if(!strcmp(type, "graphite") || !strcmp(type, "graphite:plaintext")) {
+            return 1;
+        }
+        else if(!strcmp(type, "opentsdb") || !strcmp(type, "opentsdb:telnet")) {
```
thiagoftsm: All `else  if` statement are not correct aligned.

stelfrag: I will fix it -- compiler didn't complain

.clang-format
amoss: When I try it there is still a problem with long argument lists, they indent far to the left while the kernel style guide pushes them far to the right (so that the indent lines up with the first argument), e.g:
```
void __wrap_finished_web_request_statistics(uint64_t dt, uint64_t bytes_received, uint64_t bytes_sent,
                                            uint64_t content_size, uint64_t compressed_content_size) {
}
```
... gets mangled into ...
```
void __wrap_finished_web_request_statistics(
    uint64_t dt, uint64_t bytes_received, uint64_t bytes_sent, uint64_t content_size, uint64_t compressed_content_size)
{

```

amoss: For linux kernel style (that matches the wording in the style guide and the random files that I checked):
```
AlignAfterOpenBracket: true
BinPackArguments: true
BinPackParameters: true
AllowAllParametersOfDeclarationOnNextLine: false
```

vlvkobal: I don't have a strong opinion on `AlignAfterOpenBracket` and `AllowAllParametersOfDeclarationOnNextLine`. But I strongly dislike `BinPackArguments: true` and `BinPackParameters: true`. Code becomes unreadable when there are multiple arguments in a row if there are multiple rows.

mfundul: Why are we changing clang-format here? Didn't we vote in the previous meet-up for the various options and agree on them?

amoss: The reasons that it keeps coming up are:
* The docs do not match the decision made at the meet-up so the current coding style is an opaque mess known only to the people that discussed it in person.
* None of the code-base matches the coding-style so it is a problem on every commit.
* We all need to internalize the coding-style so that it is natural - this is hard when it is not documented.

It is true that it does not relate to the work on the Exporting Engine and I think that @vlvkobal is going to create a separate issue to handle this. After a long discussion on slack, I agree with @vlvkobal about the binpacking and the rest of the conversation can wait until there is an appropriate issue. I have some suggestions on how to fix the docs, but they can wait for the new issue.

mfundul: `clang-format` is the only official decision that has been made regarding code-style.
- The docs are wrong, they should be fixed.
- code-base is wrong, will be fixed by a massive code-style commit as has already been decided

amoss: Perhaps we should continue this discussion in an RFC issue about the proposed changes to the `.clang-format` ?

mfundul: Fine by me but the changes to clang-format should be reverted.

vlvkobal: Yes. It's not the right place to discuss the format. Let's do it in #7271.

exporting/process_data.c
vlvkobal: `n` is used outside the loop. So the code won't be changed.

ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/netdata/netdata/pullRequest?prid=4361001&bid=14813674)

vlvkobal:
The exporting engine should be merged to the master branch as an experimental feature.

exporting/read_config.c
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * @return It returns the connector id.
+ */
+BACKEND_TYPE exporting_select_type(const char *type)
+{
+    if (!strcmp(type, "graphite") || !strcmp(type, "graphite:plaintext")) {
+        return BACKEND_TYPE_GRAPHITE;
+    } else if (!strcmp(type, "opentsdb") || !strcmp(type, "opentsdb:telnet")) {
+        return BACKEND_TYPE_OPENTSDB_USING_TELNET;
+    } else if (!strcmp(type, "opentsdb:http") || !strcmp(type, "opentsdb:https")) {
```
thiagoftsm: Hi @vlvkobal ,

I verified that you are keeping the option for `opentsd:https`, but I could not see any call for OpenSSL library in the new code, are you planning to do this late? O r do you prefer I bring this when it is merged?

vlvkobal: I'm going to do that later.

thiagoftsm: All right, thanks!

thiagoftsm:
Should we add the file `exporting.conf` as a valid option for the script `edit-config`?

thiagoftsm:
I notice that we are not giving any warning for our users when two exporters are exactly the same and I think this could help our users. To detect this I used the following steps:

1 - I set inside netdata.conf:
```
enabled = yes
type = json
destination = localhost:5448
```
2 - And I used this `exporting.conf`:
```
[exporting:global]
    enabled = yes

[json:testing_1]
    enabled = yes
    destination = localhost:5448

[json:testing_2]
    enabled = yes
    destination = localhost:5449
```

My servers for this specific test were two `nc` so I received a lot of messages like this:

```
netdata ERROR : BACKENDS : BACKEND: failed to write data to database backend 'localhost:5448'. Willing to write 211333 bytes, wrote 155292 bytes. Will re-connect. (errno 35, Resource temporarily unavailable)
```

Normal servers  will accept more than one connection, this means that the final result would be data duplication. Another possibility is instead to say for our users that there are two equal connections and move in front with them, we should disable one of the duplication and move in front with only one, I think this is the best option between the two that I am bringing here.

vlvkobal:
> Should we add the file exporting.conf as a valid option for the script edit-config?

Yes, but not yet. The format hasn't been finalized yet and we haven't decided on transition from `netdata.conf` to `exporting.conf` so I prefer to have just a brief readme for now.

Issue #7147, opened 2019-10-21
Python plugin for Memcachier


joseliber:
Hello everyone. I'm using the Memcached python plugin as a base to write a [Memcachier](https://www.memcachier.com/documentation/analytics-api) plugin.
To retrieve the data from Memcachier, I have to make two requests to this URL: https://analytics.memcachier.com/api/v1

First request will return a dict (actually, json parsed as dict) of `cache_id`.
With this ID, I can make the second request to:
https://analytics.memcachier.com/api/v1/{id}/stats

The result will have this structure:
```
{
  "a.b.c.totallyaserver.com:12345": {
    "limit_maxbytes": 0,
    "time": 0,
    "auth_cmds": 0,
    "auth_errors": 0,
    "bytes_read": 0,
    "bytes_written": 0,
    "cas_badval": 0,
    "cas_hits": 0,
    "cas_misses": 0,
    "cmd_delete": 0,
    "cmd_flush": 0,
    "cmd_get": 0,
    "cmd_set": 0,
    "cmd_touch": 0,
    "curr_connections": 0,
    "decr_hits": 0,
    "decr_misses": 0,
    "delete_hits": 0,
    "delete_misses": 0,
    "evictions": 0,
    "expired": 0,
    "get_hits": 0,
    "get_misses": 0,
    "incr_hits": 0,
    "incr_misses": 0,
    "total_connections": 0,
    "touch_hits": 0,
    "touch_misses": 0,
    "bytes": 0,
    "curr_items": 0,
    "total_items": 0
  }
}
```

Supposing I only have one cache_id and one _server_ (a.b.c.totallyaserver.com:12345 in the example above), I can just get the nested dict and send this to netdata as the response.

My question is, what would be the best way to make this work with multiple cache_id and/or multiple server.

The easiest way would be putting this information in the config file and have netdata retrive it from there. Then it will be `server*ids` jobs (each with one id, one server) for each run. The counterpoint for this is that the user would have to perform extra steps on the config file before having a functional plugin.

Please let me know your thoughts on this.

Regards,
José.

ilyam8:
Hi. According [docs](https://www.memcachier.com/documentation/analytics-api)

> curl "https://analytics.memcachier.com/api/v1/login"
  --user CRED_USERNAME:CRED_PASSWORD

return 

> JSON response with memcachier ID

```json
{
    "cache_id": 1561
}
```

I guess it could contains several IDs, right?

joseliber:
[...]

PR #7185, opened 2019-10-24
add support for memcachier

joseliber:
##### Summary
Add support for Memcachier service stats via [analytics API](https://www.memcachier.com/documentation/analytics-api)

Fixes: #7147

Work based on Memcached collector.

##### Component Name
collectors/python.d/memcachier

##### Additional Information
Collector is disabled by default since it needs to be configured with user/pass
before activation.

From personal experience, I've restricted the cache_id and cache_server to only 1 result per job. I've searched through the Memcachier documentation and did tests with multiple caches per user to confirm this behavior. This is explained in the README as well.

I've purposefully left out the following metrics:
    "auth_cmds": 0,
    "auth_errors": 0,

Since auth_cmds will increase with each stats request made by netdata (1/s currently).

### DIFF
--- a/collectors/python.d.plugin/memcachier/README.md
+++ b/collectors/python.d.plugin/memcachier/README.md
@@ -1,0 +1,24 @@
+# memcachier
+
+Memcachier monitoring module. Data is retrieved using the [analytics API](https://www.memcachier.com/documentation/analytics-api).
+Based on the [Memcached plugin](../memcached).
+
+This plugin expects to receive only one cache_id and only one cache_server per job instance.
+
+If you're having issues, running the plugin in [debug mode](../#how-to-debug-a-python-module) will show the results from each request.
+
+## Configuration
+
+It is necessary to provide a valid user/pass parameter in the configuration file before using this plugin.
+Due to this restriction, the plugin is deactivated by default.
+
+Sample:
+
+```
+instance1:
+  name: 'testing'
+  user: '012DEF'
+  pass: '32D1G175000000000000000000000000'
+```
+
+[![analytics](https://www.google-analytics.com/collect?v=1&aip=1&t=pageview&_s=1&ds=github&dr=https%3A%2F%2Fgithub.com%2Fnetdata%2Fnetdata&dl=https%3A%2F%2Fmy-netdata.io%2Fgithub%2Fcollectors%2Fpython.d.plugin%2Fmemcachier%2FREADME&_u=MAC~&cid=5792dfd7-8dc4-476b-af31-da2fdb9f93d2&tid=UA-64295674-3)](<>)
--- a/collectors/python.d.plugin/memcachier/memcachier.chart.py
+++ b/collectors/python.d.plugin/memcachier/memcachier.chart.py
@@ -1,0 +1,204 @@
+# -*- coding: utf-8 -*-
+# Description: memcachier netdata python.d module
+# Author: José Canal (joseliber)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import json
+
+from bases.FrameworkServices.UrlService import UrlService
+
+
+ORDER = [
+    'cache',
+    'net',
+    'connections',
+    'items',
+    'evicted',
+    'expired',
+    'get',
+    'get_rate',
+    'set_rate',
+    'flush_rate',
+    'delete_rate',
+    'cas',
+    'delete',
+    'increment',
+    'decrement',
+    'touch',
+    'touch_rate',
+]
+
+CHARTS = {
+    'cache': {
+        'options': [None, 'Cache Size', 'MiB', 'cache', 'memcachier.cache', 'stacked'],
+        'lines': [
+            ['avail', 'available', 'absolute', 1, 1 << 20],
+            ['used', 'used', 'absolute', 1, 1 << 20]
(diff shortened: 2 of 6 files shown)


collectors/python.d.plugin/memcachier/memcachier.chart.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # we need to guarantee that only one cache_id is passed forward
+        data = json.loads(data)
+        if len(data) == 1 and type(data['cache_id']) == int:
+            self.url = 'https://analytics.memcachier.com/api/v1/{id}/stats'.format(id=data['cache_id'])
+            return True
+
+        self.error('This plugin only works with one cache_id per job.')
+        self.debug('Request result: {r}'.format(r=data))
+        return False
```
ilyam8: lets make happy path last

```python
    if smth1:
        return False

    if smth2:
        return False

    return True
```

collectors/python.d.plugin/memcachier/memcachier.chart.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        self._manager = self._build_manager()
+        data = self._get_raw_data()
+
+        if data is None:
+            return False
+
+        # we need to guarantee that only one cache_id is passed forward
+        data = json.loads(data)
+        if len(data) == 1 and type(data['cache_id']) == int:
```
ilyam8: I am still not clear. Could we have several ids in the response or not? It is not clear from their docs.

How about to ask on [github](https://github.com/memcachier/)?

collectors/python.d.plugin/memcachier/memcachier.chart.py
```diff
@@ -0,0 +1,204 @@
...
+    'flush_rate': {
+        'options': [None, 'Rate', 'requests/s', 'flush ops', 'memcachier.flush_rate', 'line'],
+        'lines': [
+            ['cmd_flush', 'rate', 'incremental']
+        ]
+    },
+    'delete_rate': {
+        'options': [None, 'Rate', 'requests/s', 'delete ops', 'memcachier.delete_rate', 'line'],
+        'lines': [
+            ['cmd_delete', 'rate', 'incremental']
```
ilyam8: Do we really need a separate chart per cmd operation? Why not put them in one chart?

collectors/python.d.plugin/memcachier/README.md
```diff
@@ -0,0 +1,24 @@
+# memcachier
+
+Memcachier monitoring module. Data is retrieved using the [analytics API](https://www.memcachier.com/documentation/analytics-api).
+Based on the [Memcached plugin](../memcached).
+
+This plugin expects to receive only one cache_id and only one cache_server per job instance.
```
joelhans:
```suggestion
This plugin expects to receive only one `cache_id` and only one `cache_server` per job instance.
```

ilyam8:
Hi @joseliber, are you going to finish the PR?

joseliber:
Hi @ilyam8, yes.
I've got new information about the Analytics API so I'm changing part of the code to better handle the API's responses.
Reference issue:
https://github.com/memcachier/docs/issues/32

Regards,
José.

joseliber:
Hi @ilyam8, I have done all suggestions above, but I have one question about the charts in the case where we have more than one `server_id`. Example response:

```
{
  "a.server.com:11211": {
     metrics...
  },
  "b.server.com:11211": {
     metrics...
  }
}
```

Supposing I have two servers with data, I'm testing with something like this:

```python
...
metrics = json.loads(response)

for i in range(len(metrics)):
    data = list(metrics.values())[i]
...
return data
```

Obviously this just overwrites `data` with the values from the last server, so it is not a good solution.

What would be a good way to present info from all servers?

image

1 - Concatenate the `server_id` with the chart id (e.g. cache a.server.com). So this will create multiple charts per server for the same metric.

2 - Add the `server_id` to the chart options (not sure if possible) or something similar to this. The idea would be to have one cache chart with two graphs, one for each server:
  - Cache Size a.server.com (memcachier_job.cache)
  - Cache Size b.server.com (memcachier_job.cache)

3 - Embed the server information in the dimensions. We would have something like:
  - available (a.server.com)
  - available (b.server.com)
  - used (a.server.com)
  - used (b.server.com)


Taking into account that the server names are not short, I think that number 3 wouldn't be a good idea.

Also, I'll probably need to build the CHARTS dynamically, are there any instructions on how to do this?

Please let me know your thoughts.

Regards,
José.

ilyam8:
@joseliber you need to add prefix to the:
 - chart_id
 - dim_id

And you need
 - create charts template, example:

> https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/beanstalk/beanstalk.chart.py#L117-L168

 - add some attribute to track new servers, example:

```python
class Service(SimpleService):
        ...
        self.seen_servers = set()
```

 - add create charts from template for new added servers, example:

https://github.com/netdata/netdata/blob/8db9b928e9d3050bd2af6d7554d98645201fdd69/collectors/python.d.plugin/beanstalk/beanstalk.chart.py#L217-L226


That is pretty much it, i think you got the gist.

ilyam8:
hi @joseliber any news?

PR #7187, opened 2019-10-24
Fix sizeof inside callocz

### LINKED ISSUE #7160: Coverity  CID 350131:  Incorrect expression
##### Bug report summary
From coverity scan:
```
CID 350131:  Incorrect expression  (SIZEOF_MISMATCH)
/collectors/proc.plugin/proc_pagetypeinfo.c: 225 in do_proc_pagetypeinfo()
219                     systemorders[o].rd = rrddim_add(st_order, id, name, pagesize, 1, RRD_ALGORITHM_ABSOLUTE);
220                 }
221             }
222     
223     
224             // Per-Numa Node & Zone & Type (full detail). Only if sum(line) > 0
>>>     CID 350131:  Incorrect expression  (SIZEOF_MISMATCH)
>>>     Passing argument "4960UL /* sizeof (RRDSET) */" to function "callocz" and then casting the return value to "RRDSET **" is suspicious.
225             st_nodezonetype = callocz(pagelines_cnt, sizeof(RRDSET));
226             for (p = 0; p < pagelines_cnt; p++) {
227                 pgl = &pagelines[p];
228     
229                 // Skip invalid, refused or empty pagelines if not explicitely requested
230                 if (!pgl
```
##### OS / Environment
Linux
##### Component Name
Collector
##### Expected behavior
Remove the report from Coverity

thiagoftsm:
##### Summary
Fixes #7160 

Coverity reported us that the callocz was allocating to a pointer to pointer a simple vector, this was using more memory than necessary and thanks this we did not have any bug reported, this PR fixes this problem
##### Component Name
Collectors

### DIFF
--- a/collectors/proc.plugin/proc_pagetypeinfo.c
+++ b/collectors/proc.plugin/proc_pagetypeinfo.c
@@ -222,7 +222,7 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 
 
         // Per-Numa Node & Zone & Type (full detail). Only if sum(line) > 0
-        st_nodezonetype = callocz(pagelines_cnt, sizeof(RRDSET));
+        st_nodezonetype = callocz(pagelines_cnt, sizeof(RRDSET *));
         for (p = 0; p < pagelines_cnt; p++) {
             pgl = &pagelines[p];
 


collectors/proc.plugin/proc_pagetypeinfo.c
```diff
@@ -222,7 +222,7 @@ int do_proc_pagetypeinfo(int update_every, usec_t dt) {
 
 
         // Per-Numa Node & Zone & Type (full detail). Only if sum(line) > 0
-        st_nodezonetype = callocz(pagelines_cnt, sizeof(RRDSET));
+        st_nodezonetype = callocz(pagelines_cnt, sizeof(RRDSET *));
```
mfundul: I prefer:
`st_nodezonetype = callocz(pagelines_cnt, sizeof(*st_nodezonetype));`
cause this way you don't have to go lookup the definition. Also, if the definition changes you don't have to remember changing the allocation as well.

thiagoftsm: I understand your point, but case the variable type is changed, a lot of other things will be changed in this code.

PR #7206, opened 2019-10-28
Suggest using /var/run/netdata for the unix socket

CtrlAltDel64:
##### Summary

Additional fix to #3342
(in another location in docs)

##### Component Name

docs

##### Additional Information

Don't use /tmp for unix sockets.

### DIFF
--- a/docs/Running-behind-nginx.md
+++ b/docs/Running-behind-nginx.md
@@ -205,7 +205,7 @@ You can also use a unix domain socket. This will also provide a faster route bet
 
 ```
 [web]
-    bind to = unix:/tmp/netdata.sock
+    bind to = unix:/run/netdata/netdata.sock
 ```
 
 *note: Netdata v1.8+ support unix domain sockets*
@@ -214,7 +214,7 @@ At the Nginx side, use something like this to use the same unix domain socket:
 
 ```conf
 upstream backend {
-    server unix:/tmp/netdata.sock;
+    server unix:/run/netdata/netdata.sock;
     keepalive 64;
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/Running-behind-nginx.md
```diff
@@ -205,7 +205,7 @@ You can also use a unix domain socket. This will also provide a faster route bet
 
 ```
 [web]
-    bind to = unix:/tmp/netdata.sock
+    bind to = unix:/run/netdata/netdata.sock
```
thiagoftsm: Hi @CtrlAltDel64 ,

Please, change the suggestion from `/run` to `/var/run`, because `/run` is a Linux directory only and Netdata also runs on FreeBSD. I am appending a screenshot from FreeBSD to demonstrate this:

BSD_directory_structure

Best regards!

docs/Running-behind-nginx.md
```diff
@@ -214,7 +214,7 @@ At the Nginx side, use something like this to use the same unix domain socket:
 
 ```conf
 upstream backend {
-    server unix:/tmp/netdata.sock;
+    server unix:/run/netdata/netdata.sock;
```
thiagoftsm: Please, do the same change here.

docs/Running-behind-nginx.md
Ferroin: Note also that on most modern Linux distributions `/var/run` is the same as `/run` (more specifically, `/var/run` is a symlink pointing to `/run`), so recommending `/var/run` does not really break anything on Linux (it's actually a bit more portable, because some Linux systems do not actually provide `/run`, but pretty much all of them provide `/var/run`).

CtrlAltDel64: Change has been added. Thank you for the explanation.

thiagoftsm:
Thank you @CtrlAltDel64 for your contribution!

PR #7209, opened 2019-10-28
proc.plugin: add pressure stall information

### LINKED ISSUE #5901: PSI (Pressure Stall Information)
##### Feature idea summary
Facebook implemented PSI (Pressure Stall Information) for Linux 4.20 and higher. PSI presents high quality information about the pressure the IO, memory and CPU is under. More information here:
* https://facebookmicrosites.github.io/psi/
##### Expected behavior
I'd be nice if the Pressure Stall Information could be displayed as a line graph under their respective categories.

hexchain:
##### Summary

Add [PSI](https://www.kernel.org/doc/html/latest/accounting/psi.html) monitoring to proc.plugin. Fixes #5901.

##### Component Name
collectors/proc.plugin

##### Additional Information
This is WIP because:
- [x] Explanation text on the dashboard is currently missing
- [ ] CGroup2 PSI monitoring is currently missing

Things I'd like to make sure before continuing:
- [x] Currently, it resides under "System", and has 5 graphs. Does it deserve its own section?
- [x] Is there a better way to arrange them?
- [ ] The "total" fields are not graphed. Do we need them?
- [x] Do we need a switch for each graph (like in proc_loadavg)?
- [x] The `procfile_lines` returns 2 for `/proc/pressure/cpu`, which should only have 1 line, for example:
    ```
    % python                   
    Python 3.7.4 (default, Oct  4 2019, 06:57:26) 
    [GCC 9.2.0] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> open('/proc/pressure/cpu', 'rb').read()
    b'some avg10=3.22 avg60=4.15 avg300=4.07 total=11747246754\n'
    ```
    Is this intended or is it a bug? Currently, I'm using `lines > 2` to workaround this behavior. What would be a better way to handle this?

<details><summary>Screenshot</summary>

image

</details>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -265,34 +265,34 @@ PROC_PLUGIN_FILES = \
     collectors/proc.plugin/proc_pagetypeinfo.c \
     collectors/proc.plugin/proc_pressure.c \
     collectors/proc.plugin/proc_pressure.h \
-	collectors/proc.plugin/proc_net_dev.c \
-	collectors/proc.plugin/proc_net_ip_vs_stats.c \
-	collectors/proc.plugin/proc_net_netstat.c \
-	collectors/proc.plugin/proc_net_rpc_nfs.c \
-	collectors/proc.plugin/proc_net_rpc_nfsd.c \
-	collectors/proc.plugin/proc_net_snmp.c \
-	collectors/proc.plugin/proc_net_snmp6.c \
-	collectors/proc.plugin/proc_net_sctp_snmp.c \
-	collectors/proc.plugin/proc_net_sockstat.c \
-	collectors/proc.plugin/proc_net_sockstat6.c \
-	collectors/proc.plugin/proc_net_softnet_stat.c \
-	collectors/proc.plugin/proc_net_stat_conntrack.c \
-	collectors/proc.plugin/proc_net_stat_synproxy.c \
-	collectors/proc.plugin/proc_self_mountinfo.c \
-	collectors/proc.plugin/proc_self_mountinfo.h \
-	collectors/proc.plugin/zfs_common.c \
-	collectors/proc.plugin/zfs_common.h \
-	collectors/proc.plugin/proc_spl_kstat_zfs.c \
-	collectors/proc.plugin/proc_stat.c \
-	collectors/proc.plugin/proc_sys_kernel_random_entropy_avail.c \
-	collectors/proc.plugin/proc_vmstat.c \
-	collectors/proc.plugin/proc_uptime.c \
-	collectors/proc.plugin/sys_kernel_mm_ksm.c \
-	collectors/proc.plugin/sys_block_zram.c \
-	collectors/proc.plugin/sys_devices_system_edac_mc.c \
-	collectors/proc.plugin/sys_devices_system_node.c \
-	collectors/proc.plugin/sys_fs_btrfs.c \
-	collectors/proc.plugin/sys_class_power_supply.c \
+    collectors/proc.plugin/proc_net_dev.c \
+    collectors/proc.plugin/proc_net_ip_vs_stats.c \
+    collectors/proc.plugin/proc_net_netstat.c \
+    collectors/proc.plugin/proc_net_rpc_nfs.c \
+    collectors/proc.plugin/proc_net_rpc_nfsd.c \
+    collectors/proc.plugin/proc_net_snmp.c \
+    collectors/proc.plugin/proc_net_snmp6.c \
+    collectors/proc.plugin/proc_net_sctp_snmp.c \
+    collectors/proc.plugin/proc_net_sockstat.c \
+    collectors/proc.plugin/proc_net_sockstat6.c \
+    collectors/proc.plugin/proc_net_softnet_stat.c \
+    collectors/proc.plugin/proc_net_stat_conntrack.c \
+    collectors/proc.plugin/proc_net_stat_synproxy.c \
+    collectors/proc.plugin/proc_self_mountinfo.c \
+    collectors/proc.plugin/proc_self_mountinfo.h \
+    collectors/proc.plugin/zfs_common.c \
+    collectors/proc.plugin/zfs_common.h \
+    collectors/proc.plugin/proc_spl_kstat_zfs.c \
+    collectors/proc.plugin/proc_stat.c \
+    collectors/proc.plugin/proc_sys_kernel_random_entropy_avail.c \
+    collectors/proc.plugin/proc_vmstat.c \
+    collectors/proc.plugin/proc_uptime.c \
+    collectors/proc.plugin/sys_kernel_mm_ksm.c \
+    collectors/proc.plugin/sys_block_zram.c \
+    collectors/proc.plugin/sys_devices_system_edac_mc.c \
+    collectors/proc.plugin/sys_devices_system_node.c \
+    collectors/proc.plugin/sys_fs_btrfs.c \
+    collectors/proc.plugin/sys_class_power_supply.c \
     $(NULL)
 
 TC_PLUGIN_FILES = \
--- a/Makefile.am
+++ b/Makefile.am
@@ -264,6 +264,7 @@ PROC_PLUGIN_FILES = \
     collectors/proc.plugin/proc_meminfo.c \
     collectors/proc.plugin/proc_pagetypeinfo.c \
     collectors/proc.plugin/proc_pressure.c \
+    collectors/proc.plugin/proc_pressure.h \
 	collectors/proc.plugin/proc_net_dev.c \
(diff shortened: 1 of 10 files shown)


vlvkobal:
> Currently, it resides under "System", and has 5 graphs. Does it deserve its own section?

Yes

> Is there a better way to arrange them?

I don't know if it's a better way, but you can put the cpu and memory charts under the relevant first-level menu sections. Personally, I'm OK with the way it is.

> The "total" fields are not graphed. Do we need them?

The total absolute stall time is a different metric from what the ratios show. I think we need separate charts for the totals.

> Do we need a switch for each graph (like in proc_loadavg)?

Do you mean enable/disable configuration options? Yes, we need them.

> The procfile_lines returns 2 for /proc/pressure/cpu, which should only have 1 line... Is this intended or is it a bug?

I think it is a bug, but I suggest to stick with the workaround for now.

hexchain:
Updated to add a "pressure" section (no dashboard changes yet) and a configuration toggle for every graph.

ilyam8:
@hexchain please avoid force-pushes

> https://github.com/netdata/netdata/pull/7059#issuecomment-541880555

hexchain:
@ilyam8 Sure, sorry for that.

web/gui/dashboard_info.js
```diff
@@ -17,6 +17,17 @@ netdataDashboard.menu = {
         info: 'Overview of the key system metrics.'
     },
 
+    'pressure': {
+        title: 'Pressure',
+        icon: '<i class="fas fa-fire"></i>',
+        info: '<a href="https://www.kernel.org/doc/html/latest/accounting/psi.html">Pressure Stall Information</a> ' +
+            'identifies and quantifies the disruptions caused by resource contentions.<br>' +
```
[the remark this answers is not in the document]
hexchain: The word `identifies` should not be uppercase I think?

hexchain:
Sorry for the force push this time - it was a rebase.

hexchain:
Could someone please take a look?

ilyam8:
@vlvkobal @stelfrag the info this plugin gathers seems very good for me, please plan the review

vlvkobal:
@hexchain, what about total absolute stall time? Are you going to implement these charts?

@Ferroin, @ilyam8 what do you think about the "Pressure" section in the high-level menu right below "System Overview" (current implementation)? Or should it be a section inside "System Overview"?

ilyam8:
I tend to `a section inside "System Overview"?`

Makefile.am
```diff
@@ -263,34 +263,36 @@ PROC_PLUGIN_FILES = \
...
-    collectors/proc.plugin/proc_uptime.c \
-    collectors/proc.plugin/sys_kernel_mm_ksm.c \
-    collectors/proc.plugin/sys_block_zram.c \
-    collectors/proc.plugin/sys_devices_system_edac_mc.c \
-    collectors/proc.plugin/sys_devices_system_node.c \
-    collectors/proc.plugin/sys_fs_btrfs.c \
-    collectors/proc.plugin/sys_class_power_supply.c \
+    collectors/proc.plugin/proc_pressure.c \
+    collectors/proc.plugin/proc_pressure.h \
+	collectors/proc.plugin/proc_net_dev.c \
```
vlvkobal: Please fix indentations. Tabs should be used only for recipes. Look at #7252.

collectors/cgroups.plugin/sys_fs_cgroup.c
```diff
@@ -479,6 +501,8 @@ struct cgroup {
     RRDSET *st_queued_ops;
     RRDSET *st_merged_ops;
 
+    // pressure charts are inside `struct pressure' due to complexity
```
vlvkobal: Do you really need this comment? Moreover, it's in the wrong place.

Makefile.am
hexchain: I think that's some merging mistake. Will fix.

collectors/cgroups.plugin/sys_fs_cgroup.c
hexchain: I put it there because I see RRD*s for all other charts are put directly into `struct cgroup`. It is mostly a note in case people may not know where to look for the pressure RRD*s and get confused. I'm surely willing to remove it if it's useless.

vlvkobal: I think you can remove it.

hexchain:
As for the placement, maybe it's better to put the charts for cpu, io, and memory under "System Overview" > cpu, disk, and mem respectively?

vlvkobal:
No-no, I meant the placement of the comment itself.

ilyam8:
@vlvkobal 
https://github.com/netdata/netdata/pull/7209#issuecomment-558997285
👆

vlvkobal:
Ah... You are right @ilyam8 :)

ilyam8:
@hexchain 

This is good too, but the bad side of it that it will be kind of hidden from a user. I would suggest to put the stuff under `psi` section (or some other name).

hexchain:
> @hexchain, what about total absolute stall time? Are you going to implement these charts?

I can implement this but that would be another 5 more charts for the whole system and every cgroup. And I would probably make them using `RRD_ALGORITHM_INCREMENTAL`, which also makes me wonder how different it could be from the percentage values we already have.

hexchain:
@ilyam8 
> This is good too, but the bad side of it that it will be kind of hidden from a user. I would suggest to put the stuff under `psi` section (or some other name).

Under "System Overview"?

vlvkobal:
Let's wait for @Ferroin's opinion before making a final decision.

Ferroin:
I would tend to agree with @ilyam8 on this one. Conceptually, PSI is supposed to be a better replacement for the load average metrics, so I'd suggest at least putting it in the System Overview section like the load average is.

In an ideal situation, I'd expect to see the CPU PSI info directly below the main CPU chart (where Load Average is right now), the IO PSI info directly below the main Disk IO charts (though that might be a bit misleading, as the IO tracking is all IO, not just disk IO), and the Memory PSI info directly below the main RAM and Swap charts.

ilyam8:
Ok, i read the article
> https://lwn.net/Articles/759658/

And i agree with @Ferroin `ideal situation`

vlvkobal:
> I can implement this but that would be another 5 more charts

I don't see any problems with five more charts since they are useful. If you wish, you can implement them in a different PR.

hexchain:
> > I can implement this but that would be another 5 more charts
> 
> I don't see any problems with five more charts since they are useful. If you wish, you can implement them in a different PR.

Fine, I'll defer that to another PR.

Ferroin:
> > > I can implement this but that would be another 5 more charts
> > 
> > 
> > I don't see any problems with five more charts since they are useful. If you wish, you can implement them in a different PR.
> 
> Fine, I'll defer that to another PR.

Just a quick thought, but might it make sense to have the total absolute time charts be optional?  Some people may not be interested in them and if there are a lot of cgroups on the system they may want to reduce Netdata's resource usage (RAM and disk space) by turning them off without losing the other charts.

Now that I think of it, it may make sense to do similarly for each type of chart (CPU/Memory/IO).

hexchain:
> Now that I think of it, it may make sense to do similarly for each type of chart (CPU/Memory/IO).

Currently, all charts are optional and can be turned off individually.

vlvkobal:
@hexchain please finish with the menu sections so we could merge the PR.

hexchain:
> In an ideal situation, I'd expect to see the CPU PSI info directly below the main CPU chart (where Load Average is right now), the IO PSI info directly below the main Disk IO charts (though that might be a bit misleading, as the IO tracking is all IO, not just disk IO), and the Memory PSI info directly below the main RAM and Swap charts.

The charts are now arranged in this way, but I'm kinda unsure about the description texts.

Ferroin:
> > In an ideal situation, I'd expect to see the CPU PSI info directly below the main CPU chart (where Load Average is right now), the IO PSI info directly below the main Disk IO charts (though that might be a bit misleading, as the IO tracking is all IO, not just disk IO), and the Memory PSI info directly below the main RAM and Swap charts.
> 
> The charts are now arranged in this way, but I'm kinda unsure about the description texts.

I think what you've got right now is good actually, it's concise but still conveys the most important information and includes a link to more details, though I'd suggest replacing 'on a given resource' with a reference to the particular resource (CPU/Memory/IO).

hexchain:
@Ferroin 

>I'd suggest replacing 'on a given resource' with a reference to the particular resource (CPU/Memory/IO).

Done.

ilyam8:
@hexchain thanks! good work 👍

Issue #7165, opened 2019-10-23
Fully test the new URL parsing code.


amoss:
#### Summary

- [ ] Build performance tests (wrk) to check behaviour under load.
- [ ] See if we can put the set of fuzzing urls into wrk as a payload generator.

#### Expected result

The test results should show that the new URL parsing code is properly integrated, and that we can demonstrate good coverage of the API automatically.

amoss:
The swagger definitions got upgraded to OpenAPI3.0 to fix a bug. We were planning to improve the handling of the response formats when this was done - it will be necessary for the fuzzer to decode the responses in this piece of work anyway. The missing points are moving into this issue:

In OpenAPI3.0 there is a new "oneOf" operator that can be used to indicate the sub-types. Converting the API requires:

* Changing the fields in the header for the new syntax.
* Working out how to use `oneOf` in the definitions.
* Checking for differences between Swagger2.0 and OpenAPI3.0

ilyam8:
Enhance the testability of the Netdata internal web-server

Issue #7164, opened 2019-10-23
Rewrite the URL parsing code.


amoss:
#### Summary

Careful analysis of the code in the picohttpparser has shown that it does not solve the problem that we have, and it solves a different problem that we do not have.

What we need:

* We need to process a complete http request, detecting partial requests and avoid potential for slow TCP attacks.
* Check if the request uses a GET or our custom STREAM and dispatch appropriately.
* Handle any differences in file-transfers vs API calls (e.g. setting socket options).
* Break the path down into components, and use to handle the dispatch into the API.
* Break down the rest of the URL, extracting the parameter key/value pairs quickly, and producing a form that we can use in the API.
* Extract the values of white-listed headers that we care about (avoiding buffer overflow attacks).

What the picohttpparser code does:

* Builds a dictionary of all the headers in the request for arbitrary processing by the h2o web-server.
* Treats the URL as an opaque string without any further processing.
* Extract an arbitrary method but does no processing on it.

Proposed work:

* Implement the list above ("what we need") directly in the existing code.
* Use the unit-tests to make sure it is bullet-proof.
* Use the fuzzer to check that it accesses the API correctly.
* Put gcov / lcov into the unit-test build to make sure that we've tested it properly.

Notes about complexity:

* The largest change in the implementation style is that we should not rewrite the URL string in order to call `simple_hash` as this creates shared state that is pushed down a deep call-chain.
* Add `simple_hash_n` to `libnetdata`: a variation that does not search for a null-terminator but instead hashes `n` characters.
* Build a fixed size auxiliary structure that identifies the location of *known* parameter values in the string in a single-pass.
* Avoid partitioning the string with null-terminators to separate into tokens.

#### Expected result

Given the wide estimates for planning, and the team perception of risk the result will be demonstrable code on a fork. Testing and review before it is merged into main will be handled in a separate issue.

amoss:
Sprint milestone label should have been removed when the other issues in the epic were re-split after the planning meeting - doing it now to clear up before tomorrow.

amoss:
Task will not be finished in this sprint - we had a change of priorities and this is being delayed until after the Winter Release.

amoss:
Not stale, just resting.

ilyam8:
Closing this feature request. We will re-evaluate the request internally.

Issue #7029, opened 2019-10-08
Document the data-model of the allmetrics endpoint.


amoss:
### Feature idea summary
In issue #7005 we fixed the swagger documentation for the `api/v1/data` endpoint to show the structure of the responses for each of the possible formats. Currently the `api/v1/allmetrics` endpoint has no documentation for the responses.

##### Expected behavior
Update the swagger documentation to show the data model for the different responses to the `api/v1/allmetrics` endpoint. This involves the json formatted response, the shell reponse and the promethesis_write reponse.

The fuzzer will report that positive / negative tests against the endpoint pass.

amoss:
The format of the responses is not defined in the swagger. It needs to be written so that the fuzzer can match the server responses against it for testing.

ilyam8:
Closing this feature request. We will re-evaluate the request internally.

Issue #6968, opened 2019-09-28
Improve API documentation for `api/v1/alarms`


cakrit:
##### Feature idea summary
The response model for the `alarms` call isn't documented in our Swagger documentation. We need to add descriptions for each returned value. See the `data` response for an example.

##### Expected behavior
Explain every variable returned in the JSON. The fuzzer will report tests pass for negative and positive tests against the endpoint.

thiagoftsm:
While I was reviewing the PR 7012 I notice that it will be necessary to do updates in the file `web/api/netdata-swagger.yaml` that are related to the alarm, so I am bringing the comments that I wrote there to here for it be fixed when the new PR that has relationship with alarm is created.

- Variable `latest_alarm_log_unique_id:` : The alarm log is not "only" int32 it is unsigned int 32, please see database/rrdcalc.cfunctionrrdcalc_get_unique_id`.
- Variable `now`: Now it is not an int32, now is the current timestamp that is time_t, I think it is more safe to say it is unsigned int 64 bits, for more details, please see health/health_json.c function health_alarms2json and libnetdata/clocks/clocks.c function now_realtime_sec()
- Variable `id` : The id is unsigned long, so it will be int32 if and only if you are working on 32 bits computer.
- Variable `last_status_change:` : please, this variable has the same pattern described for the other time_t variables, can you change it?
- Variable `last_updated` : please, this variable has the same pattern described for the other time_t variables, can you change it?
- Variable `next_update` : please, this variable has the same pattern described for the other time_t variables, can you change it?
- Variable `delay_multiplier` : Delay multiplier is a float value not an int32, please, change it.
- Variable `delay_up_to_timestamp` : please, this variable has the same pattern described for the other time_t variables, can you change it?
- Variable `db_after` : This is unsigned long, please, can you change it?
- Variable `db_before` : This is unsigned long, please, can you change it? All the corrections asked here can be confirmed in the file health/health_json.c in the function health_rrdcalc2json_nolock

For alarm_log_entry the following changes are necessary:
- Variable `unique_id`: This is unsigned int32, please, can you change it?
- Variable `alarm_id `: This is unsigned int32, please, can you change it?
- Variable `alarm_event_id` : This is unsigned int32, please, can you change it?
- Variable `exec_run` : This is unsigned int32, please, can you change it?
- Variable `when`: This is time, so it follows the rules for time_t and it is defined as unsigned long, please, can you change it?
[...]

PR #7210, opened 2019-10-28
Building a fuzzer against the API (issue #7163)

### LINKED ISSUE #7163: Fuzz-testing the current API.
#### Summary

- [x] Fuzz-test the API (generate from our swagger definition ).
- [x] Customize the fuzzer to check database interactions, response formats (matching swagger defs) and return codes of requests.

We have identified fuzzers that can work with our swagger definition. These are implemented in python, need to check if there is a pure-c version that will introduce fewer dependencies.

Work needed: Modify the fuzzer so that the generated URLs are more relevant to our API.

#### Expected result

A fuzzing tool that can be called from the command-line that generates URLs / expected responses from our latest swagger definition.

amoss:
##### Summary
Fixes #7163 
Writing a fuzzer for the web API. Validates responses against the data-model defined in swagger. Because this performs model-checking against the swagger (in addition to using it for generating the docs) - upgrading the swagger model to be machine-checkable in addition to human-readable.

##### Component Name
area/daemon
area/web

##### Additional Information
Currently generates a small set of negative / positive test-cases and validates the responses as fully as is currently possible. The documentation-related issues within the same epic (#6968 and #7029) are areas of the swagger that needs to be implemented to allow full validation. The incorrect response codes that are found during the fuzzing process will be fixed during the replacement of the url parsing code in #7164. A deeper scanning depth for a larger fuzzing set will be implemented during #7165 .

### DIFF
--- a/tests/api/fuzzer.py
+++ b/tests/api/fuzzer.py
@@ -191,13 +191,13 @@ class GetPath(object):
                         target[name].values.add(v)
                 if req and len(target[name].values) == 0:
                     print(f"FAIL: No default values in swagger for required parameter {name} in {self.url}")
-            for code, schema in spec['responses'].items():
-                if code[0] == "2" and 'schema' in schema:
-                    self.success = schema['schema']
-                elif code[0] == "2":
-                    L.error(f"2xx response with no schema in {self.url}")
-                else:
-                    self.failures[code] = schema
+        for code, schema in spec['responses'].items():
+            if code[0] == "2" and 'schema' in schema:
+                self.success = schema['schema']
+            elif code[0] == "2":
+                L.error(f"2xx response with no schema in {self.url}")
+            else:
+                self.failures[code] = schema
 
     def generate_success(self, host):
         args = "&".join([f"{p.name}={some(p.values)}" for p in self.req_params.values()])
--- a/tests/api/fuzzer.py
+++ b/tests/api/fuzzer.py
@@ -29,6 +29,16 @@ def pretty_json(tree, depth=0, max_depth=None):
         else:
             print(f"{indent}{k}: {type(v)}")
 
+
+def build_url(host_maybe_scheme, base_path):
+    if '//' not in host_maybe_scheme:
+        host_maybe_scheme = '//' + host_maybe_scheme
+    url_tuple = urllib.parse.urlparse(host_maybe_scheme)
+    if base_path[0] == '/':
+        base_path = base_path[1:]
+    return url_tuple.netloc, posixpath.join(url_tuple.path, base_path)
+
+
 #######################################################################################################################
 # Data-model and processing
 
@@ -58,11 +68,15 @@ def does_response_fit_schema(schema_path, schema, resp):
        rapidly as we attempt to cover the full semantics of languages of trees defined in swagger. Instead
        we have some special cases that describe the parts of the semantics that we've used to describe the
        netdata API.
+
+       If we hit an error (in the schema) that prevents further checks then we return early, otherwise we
+       try to collect as many errors as possible.
     '''
+    success = True
     if "type" not in schema:
         L.error(f"Cannot progress past {schema_path} -> no type specified in dictionary")
         pretty_json(schema)
-        return
+        return False
     if schema["type"] == "object":
         if isinstance(resp, dict) and "properties" in schema and isinstance(schema["properties"], dict):
             L.debug(f"Validate properties against dictionary at {schema_path}")
@@ -71,63 +85,71 @@ def does_response_fit_schema(schema_path, schema, resp):
                 if v.get("required", False) and k not in resp:
                     L.error(f"Missing {k} in response at {schema_path}")
                     pretty_json(resp)
-                    return
+                    return False
                 if k in resp:
-                    does_response_fit_schema(posixpath.join(schema_path, k), v, resp[k])
-            #pretty_json(schema,max_depth=1)
-            #pretty_json(resp,max_depth=1)
+                    if not does_response_fit_schema(posixpath.join(schema_path, k), v, resp[k]):
+                        success = False
         elif isinstance(resp, dict) and "additionalProperties" in schema \
                 and isinstance(schema["additionalProperties"], dict):
             kv_schema = schema["additionalProperties"]
             L.debug(f"Validate additionalProperties against every value in dictionary at {schema_path}")
             if "type" in kv_schema and kv_schema["type"] == "object":
                 for k, v in resp.items():
-                    does_response_fit_schema(posixpath.join(schema_path, k), kv_schema, v)
-                    #L.debug(f"Validate {k}/{repr(v)} against {kv_schema}")
+                    if not does_response_fit_schema(posixpath.join(schema_path, k), kv_schema, v):
+                        success = False
             else:
-                L.error("Don't understand what the additionalProperties means?")
+                L.error("Don't understand what the additionalProperties means (it has no type?)")
+                return False
         else:
             L.error(f"Can't understand schema at {schema_path}")
             pretty_json(schema, max_depth=1)
+            return False
     elif schema["type"] == "string":
         if isinstance(resp, str):
             L.debug(f"{repr(resp)} matches {repr(schema)} at {schema_path}")
-            return
+            return True
         L.error(f"{repr(resp)} does not match schema {repr(schema)} at {schema_path}")
+        return False
     elif schema["type"] == "boolean":
         if isinstance(resp, bool):
             L.debug(f"{repr(resp)} matches {repr(schema)} at {schema_path}")
-            return
+            return True
         L.error(f"{repr(resp)} does not match schema {repr(schema)} at {schema_path}")
+        return False
     elif schema["type"] == "number":
         if 'nullable' in schema and resp is None:
-            L.debug(f"{repr(resp)} matches {repr(schema)} at {schema_path} (because nullable!)")
-            return
+            L.debug(f"{repr(resp)} matches {repr(schema)} at {schema_path} (because nullable)")
+            return True
         if isinstance(resp, int) or isinstance(resp, float):
             L.debug(f"{repr(resp)} matches {repr(schema)} at {schema_path}")
-            return
+            return True
         L.error(f"{repr(resp)} does not match schema {repr(schema)} at {schema_path}")
+        return False
     elif schema["type"] == "integer":
         if 'nullable' in schema and resp is None:
-            L.debug(f"{repr(resp)} matches {repr(schema)} at {schema_path} (because nullable!)")
-            return
+            L.debug(f"{repr(resp)} matches {repr(schema)} at {schema_path} (because nullable)")
+            return True
         if isinstance(resp, int):
             L.debug(f"{repr(resp)} matches {repr(schema)} at {schema_path}")
-            return
+            return True
         L.error(f"{repr(resp)} does not match schema {repr(schema)} at {schema_path}")
+        return False
     elif schema["type"] == "array":
         if "items" not in schema:
             L.error(f"Schema for array at {schema_path} does not specify items!")
-            return
+            return False
         item_schema = schema["items"]
         if not isinstance(resp,list):
             L.error(f"Server did not return a list for {schema_path} (typed as array in schema)")
-            return
+            return False
         for i, item in enumerate(resp):
-            does_response_fit_schema(posixpath.join(schema_path, str(i)), item_schema, item)
+            if not does_response_fit_schema(posixpath.join(schema_path, str(i)), item_schema, item):
+                success = False
     else:
-        L.error(f"What to do with the schema? {type(resp)} at {schema_path}")
+        L.error(f"Invalid swagger type {schema["type"]} for {type(resp)} at {schema_path}")
         pretty_json(schema, max_depth=1)
+        return False
+    return success
 
 
 class GetPath(object):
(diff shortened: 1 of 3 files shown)


tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+       language and we can validate a given server response by checking if the language accepts a particular
+       server response. This is basically a parser, but instead of strings we are operating on languages
+       of trees.
+
+       This could probably be extended to arbitrary swagger definitions - but the amount of work increases
+       rapidly as we attempt to cover the full semantics of languages of trees defined in swagger. Instead
+       we have some special cases that describe the parts of the semantics that we've used to describe the
+       netdata API.
+    '''
+    if "type" in schema  and  schema["type"]=="object":
```
cosmix: I would split the `type` key presence check from the value check so as not to repeat it every time which would result in another nested if.

tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+        if isinstance(resp,dict)  and  "properties" in schema  and  isinstance(schema["properties"],dict):
```
cosmix: Again, I would split the instance check here.

tests/api/fuzzer.py
thiagoftsm: Please add a space between variables, values and comparison.

tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+            resp_json = json.loads(resp.text)
+        except json.decoder.JSONDecodeError as e:
+            L.error("Non-json response - how to validate?")
+            return
+        if resp.status_code==200:
+            if self.success is not None:
+                does_response_fit_schema(posixpath.join(self.url,"200"), self.success, resp_json)
+            else:
+                L.error("Missing schema?")
+            #print(json.loads(resp.text))
```
cosmix: Remove commented-out code please.

tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+                if code=="200" and 'schema' in schema:
+                    self.success = schema['schema']
+                elif code=="200":
+                    L.error(f"200 reponse with no schema in {self.url}")
+                else:
+                    self.failures[code] = schema
+
+    def generate_success(self, host):
+        args = "&".join([f"{p.name}={some(p.values)}" for p in self.req_params.values() ])
+        base_url = urllib.parse.urljoin(host,self.url)
```
thiagoftsm: I will write something here that  is valid for the whole code until now and probably for the next methods, sometimes you are calling external methods, I suppose these methods can raise exceptions, why are you not using `try ... except` to catch them?

tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+def resolve_refs(spec, spec_root=None):
+    '''Find all "$ref" keys in the swagger spec and inline their target schemas.
+
+       As with all inliners this will break if a definition recursively links to itself, but this should not
+       happen in swagger as embedding a structure inside itself would produce a record of infinite size.'''
+    if spec_root is None:
+        spec_root = spec
+    newspec = {}
+    for k,v in spec.items():
+        if k=="$ref":
```
thiagoftsm: please, add space between variable, value and comparison here too.

tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+
+def resolve_refs(spec, spec_root=None):
+    '''Find all "$ref" keys in the swagger spec and inline their target schemas.
+
+       As with all inliners this will break if a definition recursively links to itself, but this should not
+       happen in swagger as embedding a structure inside itself would produce a record of infinite size.'''
+    if spec_root is None:
+        spec_root = spec
+    newspec = {}
+    for k,v in spec.items():
```
thiagoftsm: please, add space between variable, value and comparison here too.

tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+import urllib.parse
+
+
+#######################################################################################################################
+### Utilities
+
+def some(s):
+    return random.choice(sorted(s))
+
+def pretty_json(tree, depth=0, max_depth=None):
```
cosmix: Why not just use `json.dumps()`?

tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+                        target[name].values.add(v)
+                if req and len(target[name].values)==0:
+                    print(f"FAIL: No default values in swagger for required parameter {name} in {self.url}")
+            for code,schema in spec['responses'].items():
+                if code=="200" and 'schema' in schema:
+                    self.success = schema['schema']
+                elif code=="200":
+                    L.error(f"200 reponse with no schema in {self.url}")
+                else:
+                    self.failures[code] = schema
```
cosmix: Success codes span the whole 2xx series. It’s not just `200` you should be considering here.

tests/api/fuzzer.py
```diff
@@ -0,0 +1,319 @@
...
+import sys
+import posixpath
+import urllib.parse
+
+
+#######################################################################################################################
+### Utilities
+
+def some(s):
+    return random.choice(sorted(s))
```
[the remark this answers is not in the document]
amoss: Casting to a list is non-deterministic and will result in a different order. If the `--reseed` option has not been passed the order should be repeatable, the sort is just a way of picking an arbitrary deterministic order.

tests/api/fuzzer.py
amoss: Yes, there is lots of duplication in this method. I'm still trying to work out what the set of common operations on the schema looks like and then I will clean it up.

tests/api/fuzzer.py
amoss: It will all be gone by the time I remove the [WIP] marker (I'm still uncommenting it during dev for testing).

tests/api/fuzzer.py
amoss: It's really just for readability - the inlined version of the swagger spec is several pages long and that method just splits elements onto lines and indents the tree to show the shape.

tests/api/fuzzer.py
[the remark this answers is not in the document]
cosmix: So? Why is a deterministic order needed here tho? Why should it be repeatable? The whole point of a fuzz test is non-determinism. A non-deterministic order is just `more_randomness` in the returned element, as far as I'm concerned and that's a good thing. Maybe I'm missing something tho.

tests/api/fuzzer.py
amoss: This will be replaced with some code to pull any success codes that are defined out of the swagger. This hardcoded value is just a hack to get started.

tests/api/fuzzer.py
amoss: Do we have a style guide for python? I'm a fan of horizontal alignment to indicate structure in the expression so if we have not set one yet I have some ideas.

amoss:
> In addition to the comments, `pep8`-ify your code please.

Do we do have any local differences other than `pep8 --max-line-length=120` ?

cosmix:
> Do we do have any local differences other than `pep8 --max-line-length=120` ?

No.

amoss:
Tasks left to be completed:
* Handle the array cases for `alarm_log?` when validating the schema.
* Push missing schemas into appropriate documentation issues.
* Extend depth with a heuristic: process `charts?` response first to extract valid chart ids and then fuzz the `chart?chart=` and `data?format=json&chart=` paths more deeply.
* Remove the dead code / re-organise / clean up.

tests/api/fuzzer.py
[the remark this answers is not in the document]
amoss: The final version of the fuzzer after #7165 is intended to operate in several modes (or depths). The smaller sets are designed to be fast validations of netdata rather than extensive explorations of the parameter space. For these smaller tests we can put them straight into the build pipeline and repeatability will mean that we can rely on them - i.e. changing some code and retriggering the tests should not reveal a different set of problems to the one that an engineer is currently working on. For the larger tests we want to get lucky, so maximizing the randomness seems appropriate. 

At this point the tool is not really a fuzzer -> they aim for a large state space to overcome the single bit of information per test ("did the SUT  crash when we fuzzed it?"). Now that the swagger validation is up and running this is actually something more similar to a model-checker in that we retrieve a large amount of information about the SUT and what it did during the test sequence, i.e. it didn't crash but could we parse the output from the set of API calls to check that they are all valid. For this part to be useful it seems that repeatability is more desirable.

After #7165 it will glue both of these modes into the stress-tester. I'm not entirely sure what kind of tool that is, but if we leave it running against a build for a few hours (e.g. have a bot pull the nightly and run it on a schedule) then it should produce a high degree of confidence that the nightly is behaving as we expect.

tests/api/fuzzer.py
amoss: I will look at the urlparse and the requests calls. Catching the exceptions will tidy up the output if they are converted into L.error / L.info calls for the logging,

thiagoftsm:
I began to test the PR now, and when I decided to test it on a computer with self signed certificate I got this output:

err

this was the motive I said that it would be necessary to use `try .. exception` inside the functions, the output error is not clear.
Please, add an option for we test self-signed certificate servers.

amoss:
> this was the motive I said that it would be necessary to use `try .. exception` inside the functions, the output error is not clear.

I don't understand how you saw that output because I did wrap the call in a `try...except block` and it suppresses that message when I try it.

> Please, add an option for we test self-signed certificate servers.

I didn't think of this. Please try the new version with `--tls-no-verify` and let me know if it works for you (I do not have a configuration here to test it).

thiagoftsm:
Hi @amoss ,

The idea behind python `try .. except`  is exactly to handle the error and avoid the registers to be changed, we have something like this when we are working with C. My suggestion was in the direction to give a better error message for our users.
Your solution for self-signed certificate worked great, as you can see in the next image, congratulations!

output

Now returning for the previous output, as you can see the python script is saying:

`FAIL 2xx response with no schema in api/v1/badge.svg`

but, when I run the command displayed in the next image, I got 200 OK:

output1

I am attaching both files for you see the output:

[messages.zip](https://github.com/netdata/netdata/files/3790622/messages.zip)

at least for me it looks like that the `FAIL` is not according the output that we are receiving from netdata, please, can you verify this for us?

thiagoftsm:
I forgot to say, please check that when we run on a local netdata it is finishing with an error.

amoss:
> `FAIL 2xx response with no schema in api/v1/badge.svg`
> ...
> at least for me it looks like that the `FAIL` is not according the output that we are receiving from netdata, please, can you verify this for us?

The fuzzer is working correctly in this case. The problem is that the swagger spec is incorrect / incomplete. There is a missing definition of the schema for the response that comes back in these cases (it is `allmetrics`` in your screenshot but it is the same issue). 

The `FAIL` is indicating a mismatch between the swagger and the behaviour of netdata. In this case we will fix it by expanding the swagger definition in #7029. In the case of `badge.svg?` the response type is defined as as json in the swagger, but the server puts an svg directly in the response body,

amoss:
> I forgot to say, please check that when we run on a local netdata it is finishing with an error.

I didn't understand what you meant here, could you explain please?

thiagoftsm:
> I didn't understand what you meant here, could you explain please?

In the first image of the previous report, it looks like that the script was expecting to have a json output and it closed after this showing errors, but we only have json output when we have a valid chart, when it is not valid, Netdata always deliver a plain text:

output2

As you can see the script stopped to run when it received an unexpected output, but the output is correct considering what Netdata does nowadays, so you will need to verify firstly a `200` output to parse late. When we have a chart that does not exist, Netdata returns `404`.

Best regards!

thiagoftsm:
Hi @amoss ,

I ran the exactly command that you asked me, I am also giving you the python version that I am using to help you

err

This test I ran on Slackware Current, I will test on other Linux as well.

amoss:
The path is relative to the base of our repo - are you in the right directory for it to work?

ilyam8:
@thiagoftsm just ctr-c ctrl-v from the terminal. Or at least choose an area you want to show and the whole desktop.

thiagoftsm:
I went to `tests/api` and I ran the command, I imagined that it was appending the URL to host before to do the request, now I ran with the local `netdata-swagger.json` and I got the following:

```
ash-5.0$ python3 fuzzer.py --url file://home/thiago/Netdata/netdata/web/api/netdata-swagger.json --host mechani.se/netdata
  File "fuzzer.py", line 351, in <module>
    spec = json.loads(get_the_spec(args.url))
  File "fuzzer.py", line 265, in get_the_spec
    with open(url[7:]) as f:
FileNotFoundError: [Errno 2] No such file or directory: 'home/thiago/Netdata/netdata/web/api/netdata-swagger.json'
```

As you can see I have the same error.

amoss:
That looks like a relative path, from your original directory you could use:
```
python3 fuzzer.py --url file://../../web/api/netdata-swagger.json --host mechani.se/netdata
```
Or if you want to use an absolute path you can put an extra slash after the protocol:
```
python3 fuzzer.py --url file:///home/thiao/Netdata/netdata/web/api/netdata-swagger.json --host mechani.se/netdata
```
The `netdata-swagger.json` file must be there - it is standard in the repo.

thiagoftsm:
> That looks like a relative path, from your original directory you could use:
> 
> ```
> python3 fuzzer.py --url file://../../web/api/netdata-swagger.json --host mechani.se/netdata
> ```

Now everything is working fine, I will start a computer exclusively to test this today, because it needs time.

tests/api/fuzzer.py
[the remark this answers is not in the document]
cosmix: > At this point the tool is not really a fuzzer 

This. ;)

tests/api/fuzzer.py
cosmix: `dumps()` is prettifying input.

PR #7216, opened 2019-10-29
support rhel by not installing the cups plugin when v1.7 is not satisfiable

### LINKED ISSUE #7215: netdata/packaging: rhel7 can not provide cups-devel >= 1.7
##### Bug report summary
When building an RPM with the provided netdata.spec the conditional which installs the cups plugin only checks for centos version 6 and 7 which returns the following error on rhel7:

```cups-devel >= 1.7 is needed by netdata-v1.18.1-1.el7.x86_64```

This is not a satisfiable dependency on rhel7, it is on rhel8 though.  I'll be submitting a PR shortly to add an additional case for this conditional.

##### OS / Environment
Redhat 7
v1.18.1
##### Component Name
packaging `netdata.spec.in`
##### Steps To Reproduce
`rpmbuild -ba ~/rpmbuild/SPECS/netdata.spec`
##### Expected behavior
A successful build

kevenwyld:
##### Summary
Fixes #7215 
##### Component Name
packaging: netdata.conf.in
##### Additional Information
I've tested this on rhel7 by building the RPM with all other satisfiable deps which works fine.  Installing the resulting RPM can create a running instance.  It looks like you don't have tests for this but maybe I missed them?  I'm happy to add whatever someone more experienced with this project can suggest.


Saruspete:
Hello,

@kevenwyld I'm having trouble with this change. Our current issue is `%{centos_ver}` var does not exist in RHEL, and its counterpart is indeed `%{rhel}`
However, your check will disable Cups plugin for every non-RHEL distribution:

The test `%if 0%{?centos_ver} != 6 && 0%{?centos_ver} != 7 && 0%{?rhel} >= 8 ` will apply accordingly:
- RHEL7 : `0 != 6 && 0 != 7 && 7 >= 8` => True, True, False: don't support cups
- RHEL8 : `0 != 6 && 0 != 7 && 8 >= 8` => True, True, True: support cups
- CentOS8: `8 != 6 && 8 != 7 && 0 >= 8` => True, True, False: don't support cups
- Non-rhel/centos: `0 != 6 && 0 != 7 && 0 >=8` => True, True, False: don't support cups.

So either: 
- check if the `rhel` var is defined, and set the `centos_ver` with it, so we don't have 2 sets of vars to check.
- copy the same tests for rhel as for centos_ver var: `%if 0%{?centos_ver} != 6 && 0%{?centos_ver} != 7 && 0%{?rhel} != 6 && 0%{?rhel} != 7`

Thanks,

prologic:
@kevenwyld Let's apply @Saruspete 's suggestion:

> check if the rhel var is defined, and set the centos_ver with it, so we don't have 2 sets of vars to check.

kevenwyld:
Sorry for the silence I have been away on vacation.  I will get to this next week.  Thank you for the suggestions I am a bit new to this and missed the logic error.

kevenwyld:
Well that didn't work as I expected, or at all.  I'll do more testing locally before I push to here again.

kevenwyld:
Ok, as requested I'm setting `centos_ver` to `%rhel` when `%rhel` is defined.  I've tested on centos 6/7 and rhel7.  Some tests appear to be failing now but I don't think they are related to my change.

Issue #7049, opened 2019-10-10
[Spike] investigate random high values from #4962 and #5297


mfundul:
##### Bug report summary
Spike investigate random high values from #4962 and #5297 and how we can improve on this.

mfundul:
I propose we disallow the use of low precision counters for writing collectors since it is impossible to differentiate between a counter reset and a counter overflow.

Ferroin:
@mfundul I think it makes more sense to avoid them whenever possible, document why, and clearly document what collectors actually use them (as well as the implications of those collectors using low-precision counters).  There are simply going to be some cases where we can't avoid them, and that way we'll let the users decide if having the data is worth it or not.

mfundul:
> 
> 
> @mfundul I think it makes more sense to avoid them whenever possible, document why, and clearly document what collectors actually use them (as well as the implications of those collectors using low-precision counters). There are simply going to be some cases where we can't avoid them, and that way we'll let the users decide if having the data is worth it or not.

The problem is that the counter overflow code is creating the effect of #4962 , which is actually a regression for netdata caused by PR #4538 . This counter overflow code is conflicting with counter resets.

Ferroin:
> The problem is that the counter overflow code is creating the effect of #4962 , which is actually a regression for netdata caused by PR #4538 . This counter overflow code is conflicting with counter resets.

I understand that.  My point was that there are numerous use cases where users just won't care about that, and we should account for those use cases.  Using high-precision counters internally will just shift where the problem is (we'll start showing restarts for low-bit width counters instead of properly tracking overflows).  We could probably do even better by having Netdata watch the rate of change and factor that into whether it thinks we've reset or overflowed, or by teaching Netdata about the bit-width of the counters it's collection (that way it knows for certain where the overflow point is, and thus can better guess if we overflowed).  However, we really can't completely solve this unless we figure out a way to know for certain when whatever we're monitoring restarts, which is not realistically possible.  So, we're at the mercy of whatever we're collecting regardless, and should just properly acknowledge that fact.

mfundul:
> 
[...]

PR #7220, opened 2019-10-29
[WIP] Fix counter reset detection

### LINKED ISSUE #4962: Randon high value in incremental chart after service restart
##### Bug report summary
I wriite python.d plugin for journalbeat, it just parse json and send data to netdata. After update to 1.11 it has strage behavior.
##### OS / Environment
Fedora 29 netdata 1.11git
##### Component Name
Custom journalbeat python.d plugin
https://github.com/pontostroy/netdata/blob/master/collectors/python.d.plugin/journalbeat/journalbeat.chart.py
##### Steps To Reproduce
Start journalbeat,
start netdata, 
restart journalbead
incremental values rise to random
/usr/libexec/netdata/plugins.d/python.d.plugin journalbeat debug trace 
shows good values 0,5,7 but chart show this
netdata

##### Expected behavior
Like on 1.10, after journalbeat restart chart  must show real numbers

### LINKED ISSUE #5297: Bandwidth counter overflow on master when slave rebooted
##### Bug report summary
When a Netdata slave is rebooted, the IP bandwidth counter displays correctly on the slave's dashboard. However, when the slave is viewed from the master's dashboard, there's a huge spike (3.8 Gbps) just after the reboot. This is on a system with a gigabit interface. There are other systems where a reboot causes an even higher spike like 2 million terabits/s.

**Master**
screenshot from 2019-01-30 14-47-26
**Slave**
screenshot from 2019-01-30 14-47-36

##### OS / Environment
Debian Stretch

netdata 1.11.1+dfsg-2

##### Component name
proc_net plugin

##### Steps To Reproduce
Set up netdata in a master-slave configuration and reboot the slave. Watch the dashboard provided by the master and see the bandwidth spike up after the reboot has completed.

##### Expected behavior
Netdata should handle the interface reset correctly and not display unrealistic values.

### LINKED ISSUE #7217: Reduce false positive notifications on boot
### Intro
I found no forum/user group so I resolved to posting a question here

#### Summary
When a system is rebooted many notifications are sent

### Detailed description
We have 10+ virtual machines running on a server 3 each with a netdata instance. 
When the Physical hardware (host) is rebooted it calls a shutdown and then startup of each vm.
When a vm is restarting we get many notifications (first critical then recovered) such as (seeming any number of the below inconsistently)

disk_backlog.sda
disk_util.sda
disk_inodes._dev_shm
disk_space._dev_shm
system.swapio
ipv4.tcphandshake
ipv4.udperrors

Is no one else affected or annoyed with this functionality, is there some sort of misconfiguration?
Is there a way to prevent notifications during shutdown/startup of the system?

I do apologise if this is a duplicate issue, I did try check for a similar issue bot found nothing relevant.

Since this is running on multiple hosts I will focus on the most common:
netdata v1.18.1
os: Arch linux 
kernel: 5.7.3

mfundul:
##### Summary
Fixes #4962
Fixes #5297
Fixes #7217

##### Component Name
daemon
database
##### Additional Information

This is the first step to fix some of the problems discussed in #7049 .

It is theoretically impossible to always be 100% sure about things such as hardware counters that overflow. As a result it is theoretically impossible to both solve the counter reset and the counter overflow prroblem.

This implementation attempts to solve both issues by mostly erring on the safe side as if it was a counter reset.

8-bit and 16-bit or any arbitrary precision counters are not supported as it would be too much effort.

We attempt to identify most of the wrap-arounds without creating too many false negatives for counter resets.

### DIFF
--- a/database/rrdset.c
+++ b/database/rrdset.c
@@ -1463,10 +1463,10 @@ void rrdset_done(RRDSET *st) {
                     uint64_t max = (uint64_t)rd->collected_value_max;
                     uint64_t cap = 0;
 
-                         if(max > 0x00000000FFFFFFFFULL) cap = 0xFFFFFFFFFFFFFFFFULL;
-                    else if(max > 0x000000000000FFFFULL) cap = 0x00000000FFFFFFFFULL;
-                    else if(max > 0x00000000000000FFULL) cap = 0x000000000000FFFFULL;
-                    else                                 cap = 0x00000000000000FFULL;
+                    if (max > 0x00000000FFFFFFFFULL)
+                        cap = 0xFFFFFFFFFFFFFFFFULL;
+                    else
+                        cap = 0x00000000FFFFFFFFULL;
 
                     uint64_t delta = cap - last + new;
 
--- a/database/rrdset.c
+++ b/database/rrdset.c
@@ -1469,11 +1469,22 @@ void rrdset_done(RRDSET *st) {
                         cap = 0x00000000FFFFFFFFULL;
 
                     uint64_t delta = cap - last + new;
+                    uint64_t max_acceptable_rate = (cap / 100) * MAX_INCREMENTAL_PERCENT_RATE;
 
-                    rd->calculated_value +=
-                            (calculated_number) delta
-                            * (calculated_number) rd->multiplier
-                            / (calculated_number) rd->divisor;
+                    // If the delta is less than the maximum acceptable rate and the previous value was near the cap
+                    // then this is an overflow. There can be false positives such that a reset is detected as an
+                    // overflow.
+                    // TODO: remember recent history of rates and compare with current rate to reduce this chance.
+                    if (delta < max_acceptable_rate) {
+                        rd->calculated_value +=
+                                (calculated_number) delta
+                                * (calculated_number) rd->multiplier
+                                / (calculated_number) rd->divisor;
+                    } else {
+                        // This is a reset. Any overflow with a rate greater than MAX_INCREMENTAL_PERCENT_RATE will also
+                        // be detected as a reset instead.
+                        rd->calculated_value += (calculated_number)0;
+                    }
                 }
                 else {
                     rd->calculated_value +=
--- a/daemon/unit_test.c
+++ b/daemon/unit_test.c
@@ -628,95 +628,9 @@ struct test test5 = {
 };
 
 // --------------------------------------------------------------------------------------------------------------------
-// test5b - 16 bit overflows
+// test5b - 64 bit overflows
 
 struct feed_values test5b_feed[] = {
-        { 0,       0x000000000000FFFFULL / 3 * 0 },
-        { 1000000, 0x000000000000FFFFULL / 3 * 1 },
-        { 1000000, 0x000000000000FFFFULL / 3 * 2 },
-        { 1000000, 0x000000000000FFFFULL / 3 * 0 },
-        { 1000000, 0x000000000000FFFFULL / 3 * 1 },
-        { 1000000, 0x000000000000FFFFULL / 3 * 2 },
-        { 1000000, 0x000000000000FFFFULL / 3 * 0 },
-        { 1000000, 0x000000000000FFFFULL / 3 * 1 },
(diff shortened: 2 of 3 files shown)


database/rrdset.c
```diff
@@ -1463,17 +1463,28 @@ void rrdset_done(RRDSET *st) {
...
                     uint64_t cap = 0;
 
-                         if(max > 0x00000000FFFFFFFFULL) cap = 0xFFFFFFFFFFFFFFFFULL;
-                    else if(max > 0x000000000000FFFFULL) cap = 0x00000000FFFFFFFFULL;
-                    else if(max > 0x00000000000000FFULL) cap = 0x000000000000FFFFULL;
-                    else                                 cap = 0x00000000000000FFULL;
+                    if (max > 0x00000000FFFFFFFFULL)
+                        cap = 0xFFFFFFFFFFFFFFFFULL;
+                    else
+                        cap = 0x00000000FFFFFFFFULL;
```
Ferroin: While I agree that 8-bit, 16-bit, and arbitrary precision counters should just be ignored by this code, I do think we should handle cases of signed 32 and 64 bit counters (wrap points would be 0x000000007FFFFFFFULL and 0xFFFFFFFFFFFFFFFULL respectively). There is, unfortunately, some software (and hardware) out there that makes the poor choice of using signed types for counters, and the likelihood of either transition being a false positive is pretty low once we factor in the below logic.

mfundul:
Actually, netdata only supports up to 64-bit signed values, so the existing code was wrong. See this:
```
typedef long long collected_number;
```
in `libnetdata/storage_number/storage_number.h`.

I will remove support for 64-bit unsigned values.

mfundul:
Or figure out a way to fix unsigned 64-bit counters, but it's not obvious.

Ferroin:
> Or figure out a way to fix unsigned 64-bit counters, but it's not obvious.

I think we'd need a new counter type, because there are some things we track that actually do need to be able to have negative values (though usually they're within a signed 16-bit, or even signed 8-bit range).

Ferroin:
> As soon as a signed counter wraps around we have a problem because it will begin to produce valid (as far as netdata is concerned) negative values. It might be simpler at this point to fall back to the old code that only handles unsigned counters if we are to improve the existing code-base without making a significant effort to fix things the proper way.

Just realized I was misinterpreting things, there should be no _counters_ we track that need to be negative, it's gauges that may need to be negative (assuming all plugins are reporting things sanely).

Assuming we can confirm that, it becomes trivial to reliably identify overflow of signed 32-bit counters, because it's a distinctly different state from reset (IOW, if the counter goes negative when it wasn't close to zero to begin with, it overflowed). No matter what we do though, we can't properly handle overflow of whatever the largest type we have is, which I didn't think of until now either (because we can't _store_ the overflow-corrected value).

mfundul:
I added some unit tests and made sure that two's complement for overflowing counters handles signed and unsigned cases. We should be covered even if a physical device overflows into negative values, the delta should always be positive. In the end only the bit width will matter for the incremental algorithm.

database/rrdset.c
```diff
@@ -1463,17 +1465,29 @@ void rrdset_done(RRDSET *st) {
                     uint64_t max = (uint64_t)rd->collected_value_max;
                     uint64_t cap = 0;
 
-                         if(max > 0x00000000FFFFFFFFULL) cap = 0xFFFFFFFFFFFFFFFFULL;
-                    else if(max > 0x000000000000FFFFULL) cap = 0x00000000FFFFFFFFULL;
-                    else if(max > 0x00000000000000FFULL) cap = 0x000000000000FFFFULL;
-                    else                                 cap = 0x00000000000000FFULL;
+                    // Signed values are handled by exploiting two's complement which will produce positive deltas
+                    if (max > 0x00000000FFFFFFFFULL)
```
cosmix: Why not use a `const` (or, dog forbid a `#define`) for these to avoid messing up your `F`s?

mfundul: The lazy answer is that I modified the existing code which was this way, but I agree that constants are better, **especially** if they values are reused.

Issue #7222, opened 2019-10-29
Apache module not working with letsencrypt cert Error: [SSL: CERTIFICATE_VERIFY_FAILED]


coloomb:
##### Bug report summary
Apache module not working with letsencrypt certificate

##### OS / Environment
Red Hat Enterprise Linux Server release 7.7 (Maipo)

##### Netdata version (ouput of `netdata -V`)
netdata v1.18.1-26-nightly

##### Component Name
Apache module

##### Steps To Reproduce
1. setup apache vhost on port 443, use letsencrypt ssl cert
2. setup forward from 80 to 443 for all traffic
3. run apache module:

```
bash-4.2$ /usr/libexec/netdata/plugins.d/python.d.plugin debug 1 apache
2019-10-29 13:57:40: python.d INFO: plugin[main] : using python v2
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : looking for 'python.d.conf' in ['/etc/netdata', '/usr/lib/netdata/conf.d']
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : loading '/usr/lib/netdata/conf.d/python.d.conf'
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : '/usr/lib/netdata/conf.d/python.d.conf' is loaded
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : looking for 'pythond-jobs-statuses.json' in /var/lib/netdata
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : loading '/var/lib/netdata/pythond-jobs-statuses.json'
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : '/var/lib/netdata/pythond-jobs-statuses.json' is loaded
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : [apache] looking for 'apache.conf' in ['/etc/netdata/python.d', '/usr/lib/netdata/conf.d/python.d']
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : [apache] loading '/usr/lib/netdata/conf.d/python.d/apache.conf'
2019-10-29 13:57:40: python.d DEBUG: plugin[main] : [apache] '/usr/lib/netdata/conf.d/python.d/apache.conf' is loaded
2019-10-29 13:57:40: python.d INFO: plugin[main] : [apache] built 3 job(s) configs
2019-10-29 13:57:40: python.d DEBUG: apache[localhost] : urllib3 version: 1.10.2
[...]

PR #7223, opened 2019-10-29
UrlService: allow to skip tls_verify for http scheme

ilyam8:
##### Summary

Fixes: #7222

Do not check scheme applying `tls_verify` option.

##### Component Name

[/collectors/python.d.plugin/python_modules/bases/FrameworkServices/UrlService](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/python_modules/bases/FrameworkServices/UrlService.py)


coloomb:
I will test it tomorrow if that's OK. Many thanks!

ilyam8:
> The alternative to the function (not tls_ca_file) or (is_https and not tls_verify) or not tls_verify 

`skip_tls_verify` is true by default for https and false for `http`. So `...or not tls_verify` wont work. Should be disabled for both schemes but that is how it was implemented from the beginning.

Issue #7218, opened 2019-10-29
rabbitmq[local] : update() unhandled exception: invalid literal for int() with base 10: 'disk_free_monitoring_disabled'


Helios-23:
i get the following error over and over but cannot track down what is causing this. the rabbitmq module seems to be reporting correctly.

python.d ERROR: rabbitmq[local] : update() unhandled exception: invalid literal for int() with base 10: 'disk_free_monitoring_disabled'

ilyam8:
Hi @Helios-23 

Please run rabbitmq module in debug mode and share the output with error

> https://docs.netdata.cloud/collectors/python.d.plugin/#how-to-debug-a-python-module

ilyam8:
And please update to the latest netdata

ilyam8:
ok, i think `rabbitmq` returns `disk_free_monitoring_disabled` indead of number on  `api/nodes` call.

rel

> https://stackoverflow.com/questions/43460638/can-rabbitmq-disk-free-monitoring-be-enabled?rq=1
> https://github.com/NYTimes/collectd-rabbitmq/issues/57

ilyam8:
@Helios-23 i think the bug is fixes in #7226, could you test it?

Helios-23:
hello sorry for the late response, ive updated my registry and server with rabbitmq to: v1.19.0-48-nightly and still getting this error:

2019-12-06 07:50:45: python.d ERROR: rabbitmq[local] : update() unhandled exception: invalid literal for int() with base 10: 'disk_free_monitoring_disabled'

ilyam8:
Hi, could [you do](https://github.com/nytimes/collectd-rabbitmq/issues/57#issuecomment-295823721)

> curl localhost:15672/api/nodes/<rabbit node>/ -u <rabbit user> | jq .

Helios-23:
heres what i get:

curl localhost:15672/api/nodes/mynode/ -u myuser | jq .
Enter host password for user 'myuser ':
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   193  100   193    0     0  11506      0 --:--:-- --:--:-- --:--:-- 12062
parse error: Invalid numeric literal at line 1, column 23

ilyam8:
Current code handles it

please do as `netdata` user

> ./python.d.plugin debug trace rabbitmq

Helios-23:
[...]

PR #7226, opened 2019-10-29
rabbitmq: handle "disk_free": "disk_free_monitoring_disabled"

ilyam8:
##### Summary
Fixes: #7218

`/api/nodes` endpoint returns `disk_free_monitoring_disabled` as a value for `disk_free` and `disk_free_limit` if rabbitmq failes to get this data.

rel: 
> https://stackoverflow.com/questions/43460638/can-rabbitmq-disk-free-monitoring-be-enabled?rq=1

##### Component Name

[/collectors/python.d.plugin/rabbitmq](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/rabbitmq)

### DIFF
--- a/collectors/python.d.plugin/rabbitmq/rabbitmq.chart.py
+++ b/collectors/python.d.plugin/rabbitmq/rabbitmq.chart.py
@@ -231,6 +231,7 @@ class Service(UrlService):
 
         data = loads(raw)
         stats = fetch_data(raw_data=data, metrics=NODE_STATS)
+        handle_disabled_disk_monitoring(stats)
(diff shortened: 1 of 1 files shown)


ilyam8:
Yes, we could be more defensive. Imo we dont need to, because currently we aware of only one problem which is explicitly handled.

ilyam8:
The problem is

https://github.com/rabbitmq/rabbitmq-management-agent/blob/0f8c6a0f0ca2c7f5b08f8158a3b0059d814b0d23/src/rabbit_mgmt_external_stats.erl#L167-L171

Issue #7211, opened 2019-10-28
Wrong Postgres database size chart after deleting a databases


jeff-alves:
##### Bug report summary
We have a machine with several databases in Postgres, one of them was very large, about 40GB, and was migrated to a separate machine.
After the migration, the database was droped from the old machine, the 40 GB was freed from the disk ![Screenshot_20191028_172352](https://user-images.githubusercontent.com/7071467/67718296-417f5380-f9ae-11e9-8a23-5cc344e3e723.png)
, but in the Postgres database size chart it still shows that the deleted database contains about 20 GB.
![Screenshot_20191028_172709](https://user-images.githubusercontent.com/7071467/67718382-6a074d80-f9ae-11e9-96f7-322bd61c8e1c.png)

I already made the query "select pg_database_size ('rm');" and it returns that the database does not exist.
##### OS / Environment
Ubuntu 18.04
PostgreSQL 9.5.17 on x86_64-pc-linux-gnu (Ubuntu 9.5.17-1.pgdg18.04+1), compiled by gcc (Ubuntu 7.4.0-1ubuntu1~18.04) 7.4.0, 64-bit
##### Netdata version (ouput of `netdata -V`)
netdata v1.18.1-26-nightly
##### Component Name
postgres_local.database_size
##### Steps To Reproduce
Have a lot of PostgreSQL databases, and drop one of them....
##### Expected behavior
Droped database size goes to 0 at the chart

ilyam8:
Hi @jeff-alves 

https://github.com/netdata/netdata/blob/fdeac75f9c2e83799c7a44a8a8647d5a561ec96e/collectors/python.d.plugin/postgres/postgres.chart.py#L906-L921

could you add

```python
self.data = dict()
```

right before `try` and see if it fixes the bug ?

jeff-alves:
> could you add
> 
> ```python
> self.data = dict()
> ```
> 
> right before `try` and see if it fixes the bug ?

It works! The dropped database has gone from the chart....
I don't know if it's expected, but it's ok.
Maybe it should stay while the full log rotate, showing  the size history log and the new size (0 bytes), and gone only after all his history is 0 bytes..
(sorry, google translate...)

PR #7228, opened 2019-10-30
postgres: do not return cached data

ilyam8:
### Summary

Fixes: #7211

Postgres module doesnt reset cached data, so if database is removed it keeps returning its stats.

After this PR module resets `self.data` dict on every iteration. 

##### Component Name

[/collectors/python.d.plugin/postgres](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/postgres)

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -904,6 +904,7 @@ class Service(SimpleService):
         if not self.alive and not self.reconnect():
             return None
 
+        self.data = dict()
         try:
             cursor = self.conn.cursor(cursor_factory=DictCursor)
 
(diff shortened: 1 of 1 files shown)


ilyam8:
@anayrat correct me if i m wrong - we dont need to cache any data, we do queries on every iteration and return reslt, right?

anayrat:
> @anayrat correct me if i m wrong - we dont need to cache any data, we do queries on every iteration and return reslt, right?

Yes, we don't have to cache any data ;)

amoss:
I wonder why the cache was in there. Have you checked the performance impact of removing it?

ilyam8:
> I wonder why the cache was in there. Have you checked the performance impact of removing it?


it is not the cache actually. There is [no need to store old data](https://github.com/netdata/netdata/pull/7228#issuecomment-547781055). Module does queries on every iteration and returns result.

PR #7238, opened 2019-11-01
.travis.yml: Prevent nightly jobs from timing out (again)

knatsakis:
#### Summary
Prevent nightlies from timing out (again). Using a different approach this time because the first attempt was not good enough.

##### Component Name
.travis.yml

##### Additional Information
[Test build output](https://travis-ci.org/knatsakis/netdata/jobs/605792730)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ matrix:
   allow_failures:
     - env: ALLOW_SOFT_FAILURE_HERE=true
 
-
+before_install:
+  - source .travis/utils.sh
 
 # Install dependencies for all, once
 #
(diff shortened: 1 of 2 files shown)


cosmix:
What was the problem with `travis_wait`?

knatsakis:
The problem was that 20 minutes were not enough after all for building the arm64 docker image.

I decided to spend the time needed in order to fix this in a more proper way. We should be ok until we reach the 60 minute time limit.

cosmix:
> The problem was that 20 minutes were not enough after all for building the arm64 docker image.
> 
> I decided to spend the time needed in order to fix this in a more proper way. We should be ok until we reach the 60 minute time limit.

If so, why not `travis_wait xx` where `xx` > `20`?

knatsakis:
I suppose I could have used travis_wait 60 (maximum time limit for the whole job) and be done with it.

If I recall correctly though, travis_wait was causing wrong timestamps (with #7239) because it buffers the output until the commands exits.

PR #7240, opened 2019-11-01
Minor docker related cleanups

knatsakis:
##### Summary
Minor docker related cleanups

##### Component Name
.travis.yml

##### Additional Information
[Test build](https://travis-ci.org/knatsakis/netdata/builds/605992367)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 dist: bionic
 language: c
-services:
-  - docker
 
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


knatsakis:
Hi @fracasula, 
I will try to explain:

- Regarding 
```
services:
￼   - docker
```
The docker service is installed and enabled by default in the `bionic` travis image that we use, so there is no need to explicitly define it.

- Regarding
```
sudo apt install -y --only-upgrade docker-ce
```
There is no need to waste time and add uncertainty by trying to upgrade the docker-ce package, because the latest version is already provided by the `bionic` travis image.

- Regarding
```
docker info
```
It is not needed, as travis prints by default all the relevant docker information (client/daemon version, etc.) for every job during startup.

PR #7246, opened 2019-11-04
netdata/docs: netdata installer documentation minor nit

paulkatsoulakis:
##### Summary
Spotted it while looking for that particular one-liner, it does not affect actual process but is a matter of consistency in naming

It was introduced in 545aa9745, i 'm surprised we didn't notice it earlier

##### Component Name
netdata/docs
netdata/installer

##### Additional Information
None

### DIFF
--- a/packaging/installer/README.md
+++ b/packaging/installer/README.md
@@ -200,13 +200,13 @@ Try our experimental automatic requirements installer (no need to be root). This
 Install the packages for having a **basic Netdata installation** (system monitoring and many applications, without  `mysql` / `mariadb`, `postgres`, `named`, hardware sensors and `SNMP`):
 
 ```sh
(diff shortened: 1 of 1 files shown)


joelhans:
Just so I'm clear, this PR is just changing the name of the file that `curl` is piping to, and then the script to be run with `bash`?

The existing commands work, but this is just clearer in terms of consistent filenames, yes?

thiagoftsm:
> Just so I'm clear, this PR is just changing the name of the file that `curl` is piping to, and then the script to be run with `bash`?

The text means "only" this, but the content of the  files `install-required-packages.sh`  and  `kickstart.sh` and what we expcted from them are completely different.
 
> The existing commands work, but this is just clearer in terms of consistent filenames, yes?

Yes, this consistent is important, because of the meaning of each script, as I said before.

joelhans:
Thanks for the clarification, @thiagoftsm. I just wanted to be sure I understood what the change did before approving.

Issue #6998, opened 2019-10-05
1m tcp syn queue cookies and alarms


dwreski:
##### Question summary
Hi, this question has been asked numerous times before, but there is no clear answer.

I am continually receiving alerts similar to this on a moderately busy Fedora 29 webserver with 128GB of RAM on a gigabit link.

1m tcp syn queue cookies = 1 cookies the number of times the TCP SYN queue of the kernel was full and sent SYN cookies, during the last minute

##### OS / Environment
uname -r
5.2.11-100.fc29.x86_64

rpm -q netdata
netdata-1.17.1-1.fc29.x86_64

netstat -s | grep -i listen
    15 SYNs to LISTEN sockets dropped

nstat -a | grep -i listen
TcpExtListenDrops               15                 0.0

##### Expected results
Attached is my sysctl.conf. Perhaps there is a kernel parameter I should be adjusting to fix this? This is also happening on multiple systems with similar configurations.

I've read #4281, #3234 and others and see no final solution. I'm not sure what other information I can provide to help troubleshoot this.

[sysctl.conf.log](https://github.com/netdata/netdata/files/3693622/sysctl.conf.log)

scambra:
I have this alarm many times every day. No full drop or overflow, but cookies counter is really high:

nstat -za|grep -i -e tcpreq -e listen
TcpExtListenOverflows           0                  0.0
TcpExtListenDrops               35                 0.0
TcpExtTCPReqQFullDoCookies      43807              0.0
TcpExtTCPReqQFullDrop           0                  0.0
TcpExtTCPFastOpenListenOverflow 0                  0.0

ilyam8:
The alarm is

https://github.com/netdata/netdata/blob/7b3f17d32425a2671cb733c4053b49cf6aa96b1c/health/health.d/tcp_listen.conf#L70-L82

___

[If syn cookies is enabled](https://blog.cloudflare.com/syn-packet-handling-in-the-wild/) (which is default):

> When a SYN cookie is being sent out:
 > - TcpExtTCPReqQFullDoCookies / LINUX_MIB_TCPREQQFULLDOCOOKIES is incremnted.
 > - TcpExtSyncookiesSent / LINUX_MIB_SYNCOOKIESSENT is incremented.
 > - Linux used to increment TcpExtListenDrops [but doesn't from kernel 4.7.](https://github.com/torvalds/linux/commit/9caad864151e525929d323de96cad382da49c3b2)


[...]

PR #7250, opened 2019-11-04
Update SYN cookie alarm to be less aggressive.

Ferroin:
Based on discussion from #6998

This should make the alarms much less noisy in cases where SYN cookies aren't being used with a particularly high frequency.

### DIFF
--- a/health/health.d/tcp_listen.conf
+++ b/health/health.d/tcp_listen.conf
@@ -71,12 +71,12 @@
       on: ip.tcp_syn_queue
       os: linux
    hosts: *
-  lookup: sum -60s unaligned absolute of TCPReqQFullDoCookies
+  lookup: average -60s unaligned absolute of TCPReqQFullDoCookies
    units: cookies
    every: 10s
-    warn: $this > 0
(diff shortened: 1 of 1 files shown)


health/health.d/tcp_listen.conf
```diff
@@ -71,12 +71,12 @@
...
-  lookup: sum -60s unaligned absolute of TCPReqQFullDoCookies
+  lookup: average -60s unaligned absolute of TCPReqQFullDoCookies
    units: cookies
    every: 10s
-    warn: $this > 0
-    crit: $this > (($status == $CRITICAL) ? (0) : (60))
-   delay: up 0 down 5m multiplier 1.5 max 1h
+    warn: $this > 1
+    crit: $this > (($status == $CRITICAL) ? (0) : (5))
+   delay: up 5 down 5m multiplier 1.5 max 1h
```
ilyam8: i think delay should be lookup time * n, so 10 or 20

and please update

https://github.com/netdata/netdata/blob/cdf698d71702b8934d1f0e1aaf6588b225de0fd8/health/health.d/tcp_listen.conf#L57-L68

PR #7256, opened 2019-11-05
web_log: treat 401 Unauthorized requests as successful

amichelic:
The web_log chart module currently treats requests with a 401 status code as bad requests. However, the 401 answers just say that the incoming request wasn't authorized properly and are often followed by a request with the correct authorization.

We're using Kerberos based authentication on our internal pages, and for this authentication method, _every_ request needs to be sent in two steps:
* Client sends an unauthenticated request
* Server replies 401 with WWW-Authenticate:Negoatiate <token>
* Client sends request again with Authorization: Negotiate <much-more-data>
* Server replies 200

The token sent in the first request will be different for every request, thus it is required to send two requests - and there will be (if all authentication tries are successful) around 50% requests with a 401 code.

To show more correct data in netdata (and avoid false alarms), I'd suggest to include the 401 code in the successful_requests bin (instead of bad_requests).

### DIFF
--- a/collectors/python.d.plugin/web_log/web_log.chart.py
+++ b/collectors/python.d.plugin/web_log/web_log.chart.py
@@ -867,7 +867,7 @@ class Web:
         :return:
         """
         code_class = code[0]
-        if code_class == '2' or code == '304' or code_class == '1':
+        if code_class == '2' or code == '304' or code_class == '1' or code == '401':
             self.data['successful_requests'] += 1
         elif code_class == '3':
             self.data['redirects'] += 1
--- a/web/gui/dashboard_info.js
+++ b/web/gui/dashboard_info.js
@@ -1822,7 +1822,7 @@ netdataDashboard.context = {
     // web_log
 
     'web_log.response_statuses': {
-        info: 'Web server responses by type. <code>success</code> includes <b>1xx</b>, <b>2xx</b> and <b>304</b>, <code>error</code> includes <b>5xx</b>, <code>redirect</code> includes <b>3xx</b> except <b>304</b>, <code>bad</code> includes <b>4xx</b>, <code>other</code> are all the other responses.',
+        info: 'Web server responses by type. <code>success</code> includes <b>1xx</b>, <b>2xx</b>, <b>304</b> and <b>401</b>, <code>error</code> includes <b>5xx</b>, <code>redirect</code> includes <b>3xx</b> except <b>304</b>, <code>bad</code> includes <b>4xx</b> except <b>401</b>, <code>other</code> are all the other responses.',
         mainheads: [
             function (os, id) {
                 void(os);


ilyam8:
@amichelic 

I personally find that grouping confusing. It uses terms from [rfc7231](https://www.rfc-editor.org/rfc/rfc7231.txt)

```
   o  1xx (Informational): The request was received, continuing process

   o  2xx (Successful): The request was successfully received,
      understood, and accepted
   o  3xx (Redirection): Further action needs to be taken in order to
      complete the request
   o  4xx (Client Error): The request contains bad syntax or cannot be
      fulfilled
   o  5xx (Server Error): The server failed to fulfill an apparently
      valid request
``` 

but doesn't include redirect 304 into `redirects` and treats 1xx as `Successful` which is not correct.

401 is successful for one setup (Kerberos based authentication) and not for other. Probably this should be configurable.

What do you think?

edit: or we need to use different terms (do not use informational, successful, etc.)

ilyam8:
@amichelic but in general i agree `401` should be treated as Successful.

ilyam8:
Please update this line

https://github.com/netdata/netdata/blob/cdf698d71702b8934d1f0e1aaf6588b225de0fd8/web/gui/dashboard_info.js#L1825

ilyam8:
> https://www.restapitutorial.com/httpstatuscodes.html

> The request requires user authentication. The response MUST include a WWW-Authenticate header field (section 14.47) containing a challenge applicable to the requested resource. The client MAY repeat the request with a suitable Authorization header field (section 14.8). If the request already included Authorization credentials, then the 401 response indicates that authorization has been refused for those credentials.

Indeed, 401 doesnt look like `bad` request for me.

ilyam8:
@Ferroin @amoss what do you think guys

cosmix:
The whole 4xx series is about _client errors_, not _server errors_. In that respect, it’s not a _bad_ request, but on the other hand, it is not _successful_ either. It’s all about the definition.

stelfrag:
The 401 is a **successful return status** that indicates you "Failed to authenticate yourself". So it's a valid change for me.

ilyam8:
@cosmix it is custom grouping made by @ktsaou. As [i said i find it confusing](https://github.com/netdata/netdata/pull/7256#issuecomment-549871731).

cosmix:
It depends on the definition and how we want to define its semantics. A high number of `401`s may indicate an IAM service failure; or a penetration attempt on a frontline service by a botnet. 

Technically speaking, yes, the request was successfully handled by the server. That, however, it doesn’t mean much. I think the whole ‘bad/good requests’ is wrong as it really depends on the services, the topology, etc, for alarming based on ‘bad’ to make any sense to people.

ilyam8:
I 100% agree, i would better remove the chart 😄 Or make it configurable with no defaults.

amoss:
I would agree that the 401s should be clustered with successful responses. My reasoning would be that it does not indicate an (unsuccessful) end to the sequence, but rather an expected intermediate stage in a successful sequence. In other words - it's a normal response as part of a successful access.

PR #7257, opened 2019-11-05
[WIP] Upgraded swagger docs from Dolphin tool.

### LINKED ISSUE #7028: Convert the swagger documentation to OpenAPI3.0
##### Feature idea summary
Currently the web API is documented in Swagger 2.0. The schema has only partial support for polymorphism; object definitions can inherit fields from super-types, but there is no way to indicate the sub-types of an object schema. We use this in documenting the different types of responses to calls in the API.

Update: a recent merge has broken the current swagger docs in the tree. The update uses features that are not available in swagger 2.0 so it can only be fixed in one of two ways:

1. Revert the changes, which will break the new fuzzer tool.
2. Upgrade the spec to OpenAPI 3.0 - if it does not create other problems.

##### Expected behavior
This will fix the errors shown in the documentation when the swagger editor is opened.

amoss:
##### Summary
Fixes #7028 
Changed the handling of the nullable elements and $refs to fix the current errors in the spec.

##### Component Name
area/web
area/doc

##### Additional Information

This is a proposed fix for 7028. This will correct the structural errors shown in the latest version of the swagger docs, but it performs an upgrade to OpenAPI3.0.

### DIFF
--- a/web/api/netdata-swagger.json
+++ b/web/api/netdata-swagger.json
@@ -1,19 +1,10 @@
 {
-  "swagger": "2.0",
+  "openapi": "3.0.0",
   "info": {
     "title": "NetData API",
     "description": "Real-time performance and health monitoring.",
     "version": "1.11.1_rolling"
   },
-  "host": "registry.my-netdata.io",
-  "schemes": [
-    "https",
-    "http"
-  ],
-  "basePath": "/api/v1",
-  "produces": [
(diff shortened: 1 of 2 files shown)


amoss:
@cakrit and @joelhans Please can you take a look at these files and tell me if they break anything that we use to generate the documentation?

PR #7266, opened 2019-11-07
use uint64_t instead of unsigned long long int

deepnum:
use <stdint.h> standard header file

### DIFF
--- a/registry/registry_internals.h
+++ b/registry/registry_internals.h
@@ -12,7 +12,7 @@
 
 // ----------------------------------------------------------------------------
 // COMMON structures
-#include <stdint.h>
+
 struct registry {
     int enabled;
 
--- a/registry/registry_internals.h
+++ b/registry/registry_internals.h
@@ -12,28 +12,28 @@
 
 // ----------------------------------------------------------------------------
 // COMMON structures
-
+#include <inttypes.h>
 struct registry {
     int enabled;
 
     // entries counters / statistics
-    unsigned long long persons_count;
-    unsigned long long machines_count;
-    unsigned long long usages_count;
-    unsigned long long urls_count;
-    unsigned long long persons_urls_count;
-    unsigned long long machines_urls_count;
-    unsigned long long log_count;
+    uint64_t persons_count;
+    uint64_t machines_count;
+    uint64_t usages_count;
+    uint64_t urls_count;
+    uint64_t persons_urls_count;
+    uint64_t machines_urls_count;
+    uint64_t log_count;
 
     // memory counters / statistics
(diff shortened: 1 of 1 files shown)


registry/registry_internals.h
```diff
@@ -12,28 +12,28 @@
 
 // ----------------------------------------------------------------------------
 // COMMON structures
-
+#include <stdint.h>
```
mfundul: Please remove this line. It is included from `libnetdata.h` that already has:
```
#if defined(HAVE_INTTYPES_H)
#include <inttypes.h>
#elif defined(HAVE_STDINT_H)
#include <stdint.h>
#endif
```

mfundul:
What is the motivation for this PR?

deepnum:
https://matt.sh/howto-c

deepnum:
why libnetdata.h has those lines

mfundul:
> 
> 
> https://matt.sh/howto-c

This is plain wrong. There are plenty of use cases for types such as `char`, `long` etc, not to mention there are also a ton of libraries using those types that we actively use. Moreover netdata has to support various platforms that may or may not have `stdint.h`.

In your case it is probably better to use `uint64_t` since this looks like legacy code right there.

mfundul:
> 
> 
> why libnetdata.h has those lines

Those lines are included everywhere and with the current structure you don't have to explicitly include any system library in other places. Those lines also cover various different platforms, that may or may not have the corresponding header file.

registry/registry_internals.h
deepnum: Ok

deepnum:
>Moreover netdata has to support various platforms that may or may not have stdint.h.

@mfundul example?

Ferroin:
> > Moreover netdata has to support various platforms that may or may not have stdint.h.
> 
> @mfundul example?

I'm rather curious myself.  The only possibility I can think of is Windows, but MSVC provides stdint.h on any of the versions we'd potentially be using (I mean, we're not seriously going to be building for XP are we?). Glibc, uClibc, Musl, Bionic, all three major BSD libc's, as well as Solaris, macOS, and pretty much every other UNIX platform that's been updated in the last decade has and provides it, so unless we're thinking of building Netdata in a standalone C environment (that is, like OS kernel code which has to provide everything itself), I'm not seeing any platforms we're considering supporting that don't have stdint.h.

mfundul:
My phrasing was unfortunate, I meant prefering `inttypes.h` over `stdint.h` or not, as the code shows. Brief description in [wikipedia](https://en.wikipedia.org/wiki/C_data_types#inttypes.h).

PR #7271, opened 2019-11-07
Fine tune clang-format

vlvkobal:
##### Summary
During the last meetup, we vote for a new `.clang-format` configuration. As far as we didn't try to see how actually code is looking for every parameter, there were some misunderstandings.

I propose some subtle changes to the format. The most important are `BinPackArguments` and `BinPackParameters`. It's very difficult to read code when there are multiple rows with multiple parameters in a row like in the following example:
```
buffer_sprintf(
    b, "%s.%s.%s.%s%s%s " COLLECTED_NUMBER_FORMAT " %llu\n", prefix, hostname, chart_name, dimension_name,
    (host->tags) ? ";" : "", (host->tags) ? host->tags : "", rd->last_collected_value,
    (unsigned long long)rd->last_collected_time.tv_sec);
```

Start of the discussion https://github.com/netdata/netdata/pull/7149#discussion_r341551942

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -16,7 +16,7 @@ AlignConsecutiveDeclarations: false
 #AlignEscapedNewlines: Left # Unknown to clang-format-4.0
 AlignOperands: true
 AlignTrailingComments: true
-AllowAllParametersOfDeclarationOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortBlocksOnASingleLine: false
 AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: None
@@ -26,8 +26,8 @@ AlwaysBreakAfterDefinitionReturnType: None
 AlwaysBreakAfterReturnType: None
 AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: false
-BinPackArguments: true
-BinPackParameters: true
+BinPackArguments: false
+BinPackParameters: false
 BraceWrapping:
   AfterClass: false
   AfterControlStatement: false
@@ -58,7 +58,7 @@ CommentPragmas: '^ IWYU pragma:'
 ConstructorInitializerAllOnOneLineOrOnePerLine: false
 ConstructorInitializerIndentWidth: 4
 ContinuationIndentWidth: 4
-Cpp11BracedListStyle: false
+Cpp11BracedListStyle: true
 DerivePointerAlignment: false
 DisableFormat: false
 ExperimentalAutoDetectBinPacking: false


thiagoftsm (dismissed):

ilyam8:
@vlvkobal please add before/after the PR example

vlvkobal:
__Before__
```
buffer_sprintf(
    b, "%s.%s.%s.%s%s%s " CALCULATED_NUMBER_FORMAT " %llu\n", prefix, hostname, chart_name, dimension_name,
    (host->tags) ? ";" : "", (host->tags) ? host->tags : "", value, (unsigned long long)last_t);
```
__After__
```
buffer_sprintf(
    b,
    "%s.%s.%s.%s%s%s " CALCULATED_NUMBER_FORMAT " %llu\n",
    prefix,
    hostname,
    chart_name,
    dimension_name,
    (host->tags) ? ";" : "",
    (host->tags) ? host->tags : "",
    value,
    (unsigned long long)last_t);```

**Comment by vlvkobal** [comment]
Cpp11BracedListStyle

__Before__
```
const struct CMUnitTest tests[] = { cmocka_unit_test(test_exporting_engine),
                                    cmocka_unit_test(test_read_exporting_config) };
```
__After__
```
const struct CMUnitTest tests[] = {cmocka_unit_test(test_exporting_engine),
                                   cmocka_unit_test(test_read_exporting_config)};
```

**Comment by vlvkobal** [comment]
`AllowAllParametersOfDeclarationOnNextLine`

__Before__
```
void function(
    parameter, parameter, parameter, parameter, parameter, parameter, parameter, parameter, parameter, parameter);
```
__After__
```
void function(
    parameter,
    parameter,
    parameter,
    parameter,
    parameter,
    parameter,
    parameter,
    parameter,
    parameter,
    parameter);
```

**Comment by amoss** [comment]
Here is a representative sample of our code base. First, the original in all its unique and varied styles:
```
struct config netdata_config = {
        .sections = NULL,
        .mutex = NETDATA_MUTEX_INITIALIZER,
        .index = {
                .avl_tree = {
                        .root = NULL,
                        .compar = appconfig_section_compare
                },
                .rwlock = AVL_LOCK_INITIALIZER
        }
};


struct option_def option_definitions[] = {
    // opt description                                    arg name       default value
    { 'c', "Configuration file to load.",                 "filename",    CONFIG_DIR "/" CONFIG_FILENAME},
    { 'D', "Do not fork. Run in the foreground.",         NULL,          "run in the background"},
    { 'd', "Fork. Run in the background.",                NULL,          "run in the background"},
    { 'h', "Display this help message.",                  NULL,          NULL},
    { 'P', "File to save a pid while running.",           "filename",    "do not save pid to a file"},
    { 'i', "The IP address to listen to.",                "IP",          "all IP addresses IPv4 and IPv6"},
    { 'p', "API/Web port to use.",                        "port",        "19999"},
    { 's', "Prefix for /proc and /sys (for containers).", "path",        "no prefix"},
    { 't', "The internal clock of netdata.",              "seconds",     "1"},
    { 'u', "Run as user.",                                "username",    "netdata"},
    { 'v', "Print netdata version and exit.",             NULL,          NULL},
    { 'V', "Print netdata version and exit.",             NULL,          NULL},
    { 'W', "See Advanced options below.",                 "options",     NULL},
};

static inline int check_host_and_call(RRDHOST *host, struct web_client *w, char *url, int (*func)(RRDHOST *, struct web_client *, char *)) {
    // Output options description.
    for( i = 0; i < num_opts; i++ ) {
        fprintf(stream, "  -%c %-*s  %s", option_definitions[i].val, max_len_arg, option_definitions[i].arg_name ? option_definitions[i].arg_name : "", option_definitions[i].description);
        if(option_definitions[i].default_value) {
            fprintf(stream, "\n   %c %-*s  Default: %s\n", ' ', max_len_arg, "", option_definitions[i].default_value);
        } else {
            fprintf(stream, "\n");
        }
        fprintf(stream, "\n");
    }

    const char *mode;
    switch(w->mode) {
        case WEB_CLIENT_MODE_FILECOPY:
            mode = "FILECOPY";
            break;

        case WEB_CLIENT_MODE_OPTIONS:
            mode = "OPTIONS";
            break;

        case WEB_CLIENT_MODE_STREAM:
            mode = "STREAM";
            break;

        case WEB_CLIENT_MODE_NORMAL:
            mode = "DATA";
            break;

        default:
            mode = "UNKNOWN";
            break;
    }

    // access log
    log_access("%llu: %d '[%s]:%s' '%s' (sent/all = %zu/%zu bytes %0.0f%%, prep/sent/total = %0.2f/%0.2f/%0.2f ms) %d '%s'",
               w->id
               , gettid()
               , w->client_ip
               , w->client_port
               , mode
               , sent
               , size
               , -((size > 0) ? ((size - sent) / (double) size * 100.0) : 0.0)
               , dt_usec(&w->tv_ready, &w->tv_in) / 1000.0
               , dt_usec(&tv, &w->tv_ready) / 1000.0
               , dt_usec(&tv, &w->tv_in) / 1000.0
               , w->response.code
               , strip_control_characters(w->last_url)
    );
}
```

Here it is again in our current `.clang-format`:
```

struct config netdata_config = { .sections = NULL,
                                 .mutex = NETDATA_MUTEX_INITIALIZER,
                                 .index = { .avl_tree = { .root = NULL, .compar = appconfig_section_compare },
                                            .rwlock = AVL_LOCK_INITIALIZER } };

struct option_def option_definitions[] = {
    // opt description                                    arg name       default value
    { 'c', "Configuration file to load.", "filename", CONFIG_DIR "/" CONFIG_FILENAME },
    { 'D', "Do not fork. Run in the foreground.", NULL, "run in the background" },
    { 'd', "Fork. Run in the background.", NULL, "run in the background" },
    { 'h', "Display this help message.", NULL, NULL },
    { 'P', "File to save a pid while running.", "filename", "do not save pid to a file" },
    { 'i', "The IP address to listen to.", "IP", "all IP addresses IPv4 and IPv6" },
    { 'p', "API/Web port to use.", "port", "19999" },
    { 's', "Prefix for /proc and /sys (for containers).", "path", "no prefix" },
    { 't', "The internal clock of netdata.", "seconds", "1" },
    { 'u', "Run as user.", "username", "netdata" },
    { 'v', "Print netdata version and exit.", NULL, NULL },
    { 'V', "Print netdata version and exit.", NULL, NULL },
    { 'W', "See Advanced options below.", "options", NULL },
};

static inline int
check_host_and_call(RRDHOST *host, struct web_client *w, char *url, int (*func)(RRDHOST *, struct web_client *, char *))
{
    // Output options description.
    for (i = 0; i < num_opts; i++) {
        fprintf(
            stream, "  -%c %-*s  %s", option_definitions[i].val, max_len_arg,
            option_definitions[i].arg_name ? option_definitions[i].arg_name : "", option_definitions[i].description);
        if (option_definitions[i].default_value) {
            fprintf(stream, "\n   %c %-*s  Default: %s\n", ' ', max_len_arg, "", option_definitions[i].default_value);
        } else {
            fprintf(stream, "\n");
        }
        fprintf(stream, "\n");
    }

    const char *mode;
    switch (w->mode) {
        case WEB_CLIENT_MODE_FILECOPY:
            mode = "FILECOPY";
            break;

        case WEB_CLIENT_MODE_OPTIONS:
            mode = "OPTIONS";
            break;

        case WEB_CLIENT_MODE_STREAM:
            mode = "STREAM";
            break;

        case WEB_CLIENT_MODE_NORMAL:
            mode = "DATA";
            break;

        default:
            mode = "UNKNOWN";
            break;
    }

    // access log
    log_access(
        "%llu: %d '[%s]:%s' '%s' (sent/all = %zu/%zu bytes %0.0f%%, prep/sent/total = %0.2f/%0.2f/%0.2f ms) %d '%s'",
        w->id, gettid(), w->client_ip, w->client_port, mode, sent, size,
        -((size > 0) ? ((size - sent) / (double)size * 100.0) : 0.0), dt_usec(&w->tv_ready, &w->tv_in) / 1000.0,
        dt_usec(&tv, &w->tv_ready) / 1000.0, dt_usec(&tv, &w->tv_in) / 1000.0, w->response.code,
        strip_control_characters(w->last_url));
}
```

Here it is again with @vlvkobal's proposed changes:
```

struct config netdata_config = {.sections = NULL,
                                .mutex = NETDATA_MUTEX_INITIALIZER,
                                .index = {.avl_tree = {.root = NULL, .compar = appconfig_section_compare},
                                          .rwlock = AVL_LOCK_INITIALIZER}};

struct option_def option_definitions[] = {
    // opt description                                    arg name       default value
    {'c', "Configuration file to load.", "filename", CONFIG_DIR "/" CONFIG_FILENAME},
    {'D', "Do not fork. Run in the foreground.", NULL, "run in the background"},
    {'d', "Fork. Run in the background.", NULL, "run in the background"},
    {'h', "Display this help message.", NULL, NULL},
    {'P', "File to save a pid while running.", "filename", "do not save pid to a file"},
    {'i', "The IP address to listen to.", "IP", "all IP addresses IPv4 and IPv6"},
    {'p', "API/Web port to use.", "port", "19999"},
    {'s', "Prefix for /proc and /sys (for containers).", "path", "no prefix"},
    {'t', "The internal clock of netdata.", "seconds", "1"},
    {'u', "Run as user.", "username", "netdata"},
    {'v', "Print netdata version and exit.", NULL, NULL},
    {'V', "Print netdata version and exit.", NULL, NULL},
    {'W', "See Advanced options below.", "options", NULL},
};

static inline int
check_host_and_call(RRDHOST *host, struct web_client *w, char *url, int (*func)(RRDHOST *, struct web_client *, char *))
{
    // Output options description.
    for (i = 0; i < num_opts; i++) {
        fprintf(
            stream,
            "  -%c %-*s  %s",
            option_definitions[i].val,
            max_len_arg,
            option_definitions[i].arg_name ? option_definitions[i].arg_name : "",
            option_definitions[i].description);
        if (option_definitions[i].default_value) {
            fprintf(stream, "\n   %c %-*s  Default: %s\n", ' ', max_len_arg, "", option_definitions[i].default_value);
        } else {
            fprintf(stream, "\n");
        }
        fprintf(stream, "\n");
    }

    const char *mode;
    switch (w->mode) {
        case WEB_CLIENT_MODE_FILECOPY:
            mode = "FILECOPY";
            break;

        case WEB_CLIENT_MODE_OPTIONS:
            mode = "OPTIONS";
            break;

        case WEB_CLIENT_MODE_STREAM:
            mode = "STREAM";
            break;

        case WEB_CLIENT_MODE_NORMAL:
            mode = "DATA";
            break;

        default:
            mode = "UNKNOWN";
            break;
    }

    // access log
    log_access(
        "%llu: %d '[%s]:%s' '%s' (sent/all = %zu/%zu bytes %0.0f%%, prep/sent/total = %0.2f/%0.2f/%0.2f ms) %d '%s'",
        w->id,
        gettid(),
        w->client_ip,
        w->client_port,
        mode,
        sent,
        size,
        -((size > 0) ? ((size - sent) / (double)size * 100.0) : 0.0),
        dt_usec(&w->tv_ready, &w->tv_in) / 1000.0,
        dt_usec(&tv, &w->tv_ready) / 1000.0,
        dt_usec(&tv, &w->tv_in) / 1000.0,
        w->response.code,
        strip_control_characters(w->last_url));
}

```

**Comment by amoss** [comment]
Although I hated the binpacking changes at first I am starting to like them as they penalize long argument lists to functions. I would argue (weakly) that we should also make the following changes to `.clang-format`:
```
AlignConsecutiveAssignments: true                                                                                       
AlignConsecutiveDeclarations: true 
```
Because it is not the 1970s and we can afford the horizontal space to make structure clearer. But it's a personal preference and I don't care strongly. This is the effect that it has:
```
struct config netdata_config = {.sections = NULL,
                                .mutex    = NETDATA_MUTEX_INITIALIZER,
                                .index    = {.avl_tree = {.root = NULL, .compar = appconfig_section_compare},
                                          .rwlock   = AVL_LOCK_INITIALIZER}};

struct option_def option_definitions[] = {
    // opt description                                    arg name       default value
    {'c', "Configuration file to load.", "filename", CONFIG_DIR "/" CONFIG_FILENAME},
    {'D', "Do not fork. Run in the foreground.", NULL, "run in the background"},
    {'d', "Fork. Run in the background.", NULL, "run in the background"},
    {'h', "Display this help message.", NULL, NULL},
    {'P', "File to save a pid while running.", "filename", "do not save pid to a file"},
    {'i', "The IP address to listen to.", "IP", "all IP addresses IPv4 and IPv6"},
    {'p', "API/Web port to use.", "port", "19999"},
    {'s', "Prefix for /proc and /sys (for containers).", "path", "no prefix"},
    {'t', "The internal clock of netdata.", "seconds", "1"},
    {'u', "Run as user.", "username", "netdata"},
    {'v', "Print netdata version and exit.", NULL, NULL},
    {'V', "Print netdata version and exit.", NULL, NULL},
    {'W', "See Advanced options below.", "options", NULL},
};

static inline int
check_host_and_call(RRDHOST *host, struct web_client *w, char *url, int (*func)(RRDHOST *, struct web_client *, char *))
{
    // Output options description.
    for (i = 0; i < num_opts; i++) {
        fprintf(
            stream,
            "  -%c %-*s  %s",
            option_definitions[i].val,
            max_len_arg,
            option_definitions[i].arg_name ? option_definitions[i].arg_name : "",
            option_definitions[i].description);
        if (option_definitions[i].default_value) {
            fprintf(stream, "\n   %c %-*s  Default: %s\n", ' ', max_len_arg, "", option_definitions[i].default_value);
        } else {
            fprintf(stream, "\n");
        }
        fprintf(stream, "\n");
    }

    const char *mode;
    switch (w->mode) {
        case WEB_CLIENT_MODE_FILECOPY:
            mode = "FILECOPY";
            break;

        case WEB_CLIENT_MODE_OPTIONS:
            mode = "OPTIONS";
            break;

        case WEB_CLIENT_MODE_STREAM:
            mode = "STREAM";
            break;

        case WEB_CLIENT_MODE_NORMAL:
            mode = "DATA";
            break;

        default:
            mode = "UNKNOWN";
            break;
    }

    // access log
    log_access(
        "%llu: %d '[%s]:%s' '%s' (sent/all = %zu/%zu bytes %0.0f%%, prep/sent/total = %0.2f/%0.2f/%0.2f ms) %d '%s'",
        w->id,
        gettid(),
        w->client_ip,
        w->client_port,
        mode,
        sent,
        size,
        -((size > 0) ? ((size - sent) / (double)size * 100.0) : 0.0),
        dt_usec(&w->tv_ready, &w->tv_in) / 1000.0,
        dt_usec(&tv, &w->tv_ready) / 1000.0,
        dt_usec(&tv, &w->tv_in) / 1000.0,
        w->response.code,
        strip_control_characters(w->last_url));
}

```

**Comment by amoss** [comment]
What I would argue very strongly though is that we must update the documentation to match whatever the actual style is, with a representative sample of code to show it. As things stand, having the contributors guide say one thing and the `.clang-format` imply another is just a waste of time. 

I am happy to update the docs when people agree about the actual style.

**Comment by mfundul** [comment]
With our current `.clang-format` in master branch this line:


**BEFORE**
```
    buffer_sprintf(b, "%s.%s.%s.%s%s%s " COLLECTED_NUMBER_FORMAT " %llu\n", prefix, hostname, chart_name, dimension_name, (host->tags) ? ";" : "", (host->tags) ? host->tags : "", rd->last_collected_value, (unsigned long long)rd->last_collected_time.tv_sec);
```
becomes:


**DEFAULT NETDATA FORMATTING**
```
    buffer_sprintf(
        b, "%s.%s.%s.%s%s%s " COLLECTED_NUMBER_FORMAT " %llu\n", prefix, hostname, chart_name, dimension_name,
        (host->tags) ? ";" : "", (host->tags) ? host->tags : "", rd->last_collected_value,
        (unsigned long long)rd->last_collected_time.tv_sec);
```
which I agree with @vlvkobal I don't really like.

Instead of changing `BinPack` configuration which changes everything we have agreed on, instead I propose setting `AlignAfterOpenBracket: Align`:


**ALIGNAFTEROPENBRACKET: ALIGN**
```
    buffer_sprintf(b, "%s.%s.%s.%s%s%s " COLLECTED_NUMBER_FORMAT " %llu\n", prefix, hostname, chart_name,
                   dimension_name, (host->tags) ? ";" : "", (host->tags) ? host->tags : "", rd->last_collected_value,
                   (unsigned long long)rd->last_collected_time.tv_sec);
```

**Comment by vlvkobal** [comment]
> AlignConsecutiveAssignments: true                                                                                       
AlignConsecutiveDeclarations: true

I like it.

@mfundul - you are genius! I'm really happy for you! But I'm not so clever :(. It takes me a lot of time to read and understand
```
    buffer_sprintf(
        b, "%s.%s.%s.%s%s%s " COLLECTED_NUMBER_FORMAT " %llu\n", prefix, hostname, chart_name, dimension_name,
        (host->tags) ? ";" : "", (host->tags) ? host->tags : "", rd->last_collected_value,
        (unsigned long long)rd->last_collected_time.tv_sec);
```
or
```
    buffer_sprintf(b, "%s.%s.%s.%s%s%s " COLLECTED_NUMBER_FORMAT " %llu\n", prefix, hostname, chart_name,
                   dimension_name, (host->tags) ? ";" : "", (host->tags) ? host->tags : "", rd->last_collected_value,
                   (unsigned long long)rd->last_collected_time.tv_sec);
```

**Comment by amoss** [comment]
> Instead of changing `BinPack` configuration which changes everything we have agreed on, instead I propose setting `AlignAfterOpenBracket: Align`:

This was the change that I wanted after reading @vlvkobal's comments in the exporter issue... but I've had time to think about it and now I think it is a good change. Varying number of parameters packed into lines because of a split are cramped and difficult to read. After looking the BinPack false output a few times:

* It wastes a horrible amount of vertical space.
* It makes it very easy to see the number of parameters, and their values.

I know that the internal answer to the the first point is `it's not the 1970s its the 2020s` :) but, the problem with wasting vertical space is that it decreases code density and puts related lines of code further apart on the screen. This makes it harder to follow values as they flow through a function.

The more that I look at the output the more that I think this is a benefit -> it gives us an incentive to reduce the number of parameters in a call.

I propose doing both -> disable binpack and align on opening brackets.

**Comment by vlvkobal** [comment]
> the problem with wasting vertical space is that it decreases code density and puts related lines of code further apart on the screen. This makes it harder to follow values as they flow through a function.

Shouldn't we adopt the best practices and make functions shorter?

**Comment by mfundul** [comment]
> 
> 
> > the problem with wasting vertical space is that it decreases code density and puts related lines of code further apart on the screen. This makes it harder to follow values as they flow through a function.
> 
> Shouldn't we adopt the best practices and make functions shorter?

I thought you were proposing making the functions longer by adding new lines after every argument.

**Comment by amoss** [comment]
> Shouldn't we adopt the best practices and make functions shorter?

That is not a best practice. I can dig out a detailed reference if you would like, but the general summary is that if you cut code into pieces that are too small then the organizational problem simply shifts up one level of abstraction. In particular, functions that have a long series of sequential (dependent) processing steps should not be artificially cut into smaller pieces than one "logical" operation.

**Comment by vlvkobal** [comment]
> I can dig out a detailed reference if you would like

I'd like to read it.

**Comment by vlvkobal** [comment]
Oh. I forgot to mention. I don't understand, how saving 10 lines for parameters could help reading a function 100 lines long?

**Comment by mfundul** [comment]
> 
> 
> Oh. I forgot to mention. I don't understand, how saving 10 lines for parameters could help reading a function 100 lines long?

Look at this function:
https://elixir.bootlin.com/linux/latest/source/fs/nfs/nfs42proc.c#L220

It has 7 function calls inside it that span 2 lines and is a very readable function body with nice flow.

If those function calls were transformed so that each argument took one line then this would be 13 additional lines that would waste space and reduce readability.

**Comment by stelfrag** [comment]
> > Shouldn't we adopt the best practices and make functions shorter?
> 
> That is not a best practice. I can dig out a detailed reference if you would like, but the general summary is that if you cut code into pieces that are too small then the organizational problem simply shifts up one level of abstraction. In particular, functions that have a long series of sequential (dependent) processing steps should not be artificially cut into smaller pieces than one "logical" operation.

Good point.. !  I would rather read a few more lines of code (if needed) to avoid following a function call chain of logic (brain stack overload)

**Comment by ilyam8** [comment]
> Shouldn't we adopt the best practices and make functions shorter?

I think you mean make function[ responsible for one thing](https://www.kernel.org/doc/html/v4.17/process/coding-style.html#functions) which makes them shorter. 

If function does several things (but it is logically one) then these things should be on the same level of abstraction.

>   I would rather read a few more lines of code (if needed) to avoid following a function call chain of logic

Then you dont need to follow functions chain, because you dont need details of implementation.

**Comment by vlvkobal** [comment]
> 13 additional lines that would waste space

I don't think it's a "waste". We are not extremely limited in space in 2020.

**Comment by mfundul** [comment]
> 
> 
> > 13 additional lines that would waste space
> 
> I don't think it's a "waste". We are not extremely limited in space in 2020. Forget VT50, VT52, and VT100 😃 .

It's about readability and flow. This function call is perfectly readable in this context as just another step of a process:
```
	status = nfs4_call_sync(server->client, server, &msg,
				&args->seq_args, &res->seq_res, 0);
```
but it doesn't help at all being like that:
```
	status = nfs4_call_sync(server->client,
                                server,
                                &msg,
				&args->seq_args,
                                &res->seq_res,
                                0);
```

This looks like much more code than it should be in this context.

**Comment by vlvkobal** [comment]
Readability is not about the number of lines. It's about structure!

> This function call is perfectly readable in this context as just another step of a process

Then we should write every function call in one line.

As for
```
	status = nfs4_call_sync(server->client,
                                server,
                                &msg,
				&args->seq_args,
                                &res->seq_res,
                                0);
```
it's absolutely incorrect example and it's a mess. How is it connected to the style we are talking about? Look at
```
status = nfs4_call_sync(server->client,
                        server,
                        &msg,
                        &args->seq_args,
                        &res->seq_res,
                        0);
```

**Comment by amoss** [comment]
> It's about readability and flow. This function call is perfectly readable in this context as just another step of a process:
> 
> ```
> 	status = nfs4_call_sync(server->client, server, &msg,
> 				&args->seq_args, &res->seq_res, 0);
> ```

No, but come on. It's not *perfectly* readable otherwise it would be this:
```
status = nfs4_call_sync(server->client,  server,        &msg,                                                           
                        &args->seq_args, &res->seq_res, 0);  
```
But I don't think that clang-format can do that with any amount of tweaking.

I wasn't in the meeting where this was discussed the first time, but I'm having a sense of deja vu. So was the outcome last time that there are strong options on code style in the team that are exactly opposite and so we have a "compromise" that is about 50% in the middle?

**Comment by amoss** [comment]
It looks as if none of the strong opinions on code-style will shift in the near future - so let's skip that whole discussion and do something productive. Why do we not have a linting report for C code in the Travis reports?

I would like to see something like this (feel free to improve):
```
#!/bin/bash                                                                                                             
                                                                                                                        
find . -type f -name '*.[ch]' | while read filename; do                                                                 
    echo -ne "$filename: "                                                                                              
    diff <(clang-format -style=file $filename) $filename | wc -l                                                        
done  
```

This would not block merging PRs with un-linted code (as most of our code would fail). But it does give us some idea how how close each PR is to the target coding style. Over time it can become the maintainer's problem to get those numbers down to zero in their AoR.

mfundul:
Ok, my updated proposal summary.
- Don't touch `BinPackArguments` and `BinPackParameters`.
- Change `AlignAfterOpenBracket` to `Align` so that packing does not suck.
- Introduce `ExperimentalAutoDetectBinPacking: true` so that the programmer has full flexibility to decide whether to pack or not, depending on the context.

cosmix:
I prefer the original `Cpp11BracedListStyle` setting myself. I'm ok (and rather prefer) the other changes proposed by @vlvkobal.

vlvkobal:
> we will need to vote again, for each parameter separately

It looks like the sticking point is BinPack*. For other parameters we can easily find a consensus.

vkalintiris (dismissed):

thiagoftsm (dismissed):

thiagoftsm (approved):

vkalintiris (approved):

underhood (approved):

stelfrag (approved):

MrZammler (approved):

PR #7289, opened 2019-11-08
apps.plugin: add process group for git-related processes

nodiscc:
##### Summary

Add git/software forge related processes to [apps_groups.conf](https://github.com/netdata/netdata/blob/master/collectors/apps.plugin/apps_groups.conf):
 - Gitea self-hosted Git service https://gitea.io/en-us/
 - Gitlab continuous integration runner https://docs.gitlab.com/runner/

##### Component Name

[/collectors/apps.plugin](https://github.com/netdata/netdata/tree/master/collectors/apps.plugin)

##### Additional Information

Default configuration change. Group called 'git' for lack of a better name.

### DIFF
--- a/collectors/apps.plugin/apps_groups.conf
+++ b/collectors/apps.plugin/apps_groups.conf
@@ -305,3 +305,5 @@ node: node*
 factorio: factorio
 
 p4: p4*
+
+git: gitea gitlab-runner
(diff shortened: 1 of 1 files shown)


vlvkobal:
I think `git` is a bit misleading. `git-services` would be a better name for the group.

ilyam8:
i see `git` is in he `build` group

https://github.com/netdata/netdata/blob/5aa60c67a8dbfaced66d04f9e36798b8a7a7782b/collectors/apps.plugin/apps_groups.conf#L210

I guess it should be moved in the new `git-services` group

vlvkobal:
`git` a tool, not a service.

vlvkobal:
Thank you @nodiscc

PR #7297, opened 2019-11-10
Function to validate destination and better error message

### LINKED ISSUE #4591: no error for url style stream destination
##### Bug report summary

If you specify the stream destination host like `tcp://examle.com` netdata tells you that the connect failed, instead of an more appropriate `syntax error`.
##### OS / Environment

freebsd 11.2p4
##### Component Name

stream.conf
##### Steps To Reproduce

Set stream.conf like below

```yaml
[stream]
    enabled = yes
    destination = tcp://monitoring.fago.me
```

##### Expected behavior

Some kind of better error message, that tells you what the problem is, instead of sending one down the wrong rabbit hole (I was sure it was a network problem until i attached gdb)

### LINKED ISSUE #7833: Validate destination
##### Feature idea summary
The issue #4591 brought an important point, Netdata was not testing the destinations before to connect with it, it would be good to give better messages every time we try to connect with an external server.
##### Expected behavior
Raise messages when destinations are not valid

thiagoftsm:
##### Summary
Fixes #4591 
Fixes #7833 
When the user was specifying hosts to connect like `tcp://192.168.30.5:19999`, netdata was given the following error:

```
2019-11-11 00:23:58: netdata ERROR : STREAM_SENDER[poseidon] : Definition 'tcp:' does not specify a host. (errno 2, No such file or directory)
```

This was happening, because the function `connect_to_one_of` present inside `libnetdata/socket/socket.c` is considering `/` as a string limit, to give better messages for our user, case netdata receives `//` as argument a new message must be given to help users to fix their problems.
##### Component Name
Libnetdata socket
##### Additional Information


ilyam8:
Correct format is

https://github.com/netdata/netdata/blob/675383b26af3a339dc966c685d3e380cc39538f0/streaming/stream.conf#L20

not 

> [PROTOCOL:]IP[%INTERFACE][:PORT]

stelfrag:
As it is now it reports: 

```
2019-11-11 13:24:58: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
```
However, it will keep on trying
```
2019-11-11 13:36:13: netdata INFO  : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: connecting...
2019-11-11 13:36:13: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
2019-11-11 13:36:13: netdata ERROR : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: failed to connect
2019-11-11 13:36:18: netdata INFO  : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: connecting...
2019-11-11 13:36:32: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
2019-11-11 13:36:32: netdata ERROR : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: failed to connect
2019-11-11 13:36:37: netdata INFO  : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: connecting...
2019-11-11 13:36:37: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
2019-11-11 13:36:37: netdata ERROR : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: failed to connect
2019-11-11 13:36:42: netdata INFO  : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: connecting...
2019-11-11 13:36:42: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
2019-11-11 13:36:42: netdata ERROR : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: failed to connect
```

thiagoftsm:
> Correct format is
> 
> https://github.com/netdata/netdata/blob/675383b26af3a339dc966c685d3e380cc39538f0/streaming/stream.conf#L20
> 
> not
> 
> > [PROTOCOL:]IP[%INTERFACE][:PORT]

Thank you, I used the comment inside the code, but I agree it is better to use the documentation example.

thiagoftsm:
> The % in the string needs to escaped.
> "destiny" should be "destination".
> The condition `(e+1)` should not be in the test.

Thank you! You are correct, last night I considered that we could have NULL, but it does not make sense as you wrote.

thiagoftsm:
> As it is now it reports:
> 
> ```
> 2019-11-11 13:24:58: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
> ```
> 
> However, it will keep on trying
> 
> ```
> 2019-11-11 13:36:13: netdata INFO  : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: connecting...
> 2019-11-11 13:36:13: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
> 2019-11-11 13:36:13: netdata ERROR : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: failed to connect
> 2019-11-11 13:36:18: netdata INFO  : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: connecting...
> 2019-11-11 13:36:32: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
> 2019-11-11 13:36:32: netdata ERROR : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: failed to connect
> 2019-11-11 13:36:37: netdata INFO  : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: connecting...
> 2019-11-11 13:36:37: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
> 2019-11-11 13:36:37: netdata ERROR : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: failed to connect
> 2019-11-11 13:36:42: netdata INFO  : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: connecting...
> 2019-11-11 13:36:42: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
> 2019-11-11 13:36:42: netdata ERROR : STREAM_SENDER[homeroot] : STREAM homeroot [send to tcp://192.168.0.114:19999]: failed to connect
> ```

It was good you bring this comment, because I think we can create another PR to solve this. Netdata has a counter to verify the number of times it tries to reconnect, but it does not have an algorithm to stop after few errors. The original idea is "We are having problem now, but the master can return and I must to send". Should we bring an interval to try to reconnect with the master after a failure?

ilyam8:
> Should we bring an interval to try to reconnect with the master after a failure?

if there is a syntax error you dont need to connect/reconnect

amoss:
This is the bit that is confusing for a user (rather than a developer):
```
2019-11-11 13:36:13: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified tcp://192.168.0.114:19999 does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
```

Because that raw string is not being split and reassembled we are feeding to the user whatever they put in. If they misunderstood then they won't understand the problem. We need to split it for processing so if we want to be clearer then we can present it like this:
```
2019-11-11 13:36:13: netdata ERROR : STREAM_SENDER[homeroot] : A destiny specified PROTOCOL="tcp" IP="//192.168.0.114" PORT="19999" does not have the correct format [PROTOCOL:]IP[%INTERFACE][:PORT]
```
This would make the problem clearer (although with the correct format as @ilyam8 pointed out further above in the thread).

stelfrag:
> It was good you bring this comment, because I think we can create another PR to solve this. Netdata has a counter to verify the number of times it tries to reconnect, but it does not have an algorithm to stop after few errors. The original idea is "We are having problem now, but the master can return and I must to send". Should we bring an interval to try to reconnect with the master after a failure?

Well I think this is a different issue (and probably another PR to fix it as you said)

But this error, because of the incorrect syntax, will be fatal no matter how many retries we attempt.

thiagoftsm:
@ilyam8 and @stelfrag I agree with both of you and netdata will not connect more when there is a `/` on  the string.
I am keeping the message inside the socket library, because the function is called from other parts of the code and the error could happen there.
@amoss I understood your point of view and case we parse, we could do the things more clean for the users, but the function where we are adding messages does not force the people to specify the protocol and port, so my suggestion is to keep the message as it is, for the user compare his current input with the format expected. I am open to continue talking about this.

amoss:
We don't need to add any new parsing code -  in order to attempt the connection we need to have separated the string into those parts. Printing the original (unsplit) string does not allow the user to understand the problem.

thiagoftsm:
> We don't need to add any new parsing code - in order to attempt the connection we need to have separated the string into those parts. Printing the original (unsplit) string does not allow the user to understand the problem.

We do this inside the function `connect_to_this`, if and only if, we do not have a '/' in the connection, case we have we send only `tcp:` to the function, so we would need to create a parser common for both functions to avoid dupplication.

amoss:
I think that the error message and the fix are in the wrong place. 

The splitting of the string to determine which characters are protocol, and which are host, happens inside `connect_to_this()`. This is the point at which the extra `//` in the string will cause a problem - because it will end up in the value of `host` when passed to `connect_to_this_ip46()`. This is the point to check that it is valid and does not start with characters that are illegal in a hostname.

Reading the code of `connect_to_this()` is very worrying. Why is it trying to find a '[' character in the host string?

stelfrag:
The brackets for IPv6 

```
   # HOST      = an IPv4, IPv6 IP, or a hostname, or a unix domain socket path.
   #             IPv6 IPs should be given with brackets [ip:address]
```
ahh its a fun chase

amoss:
It certainly is :) I don't understand what that means. An IPv6 address is typically specified in hex with `:`-separators. What does that description mean, that we should wrap them in `[` ... `]` ? (it's the "ip:address" part that is confusing me).

thiagoftsm:
> It certainly is :) I don't understand what that means. An IPv6 address is typically specified in hex with `:`-separators. What does that description mean, that we should wrap them in `[` ... `]` ? (it's the "ip:address" part that is confusing me).

The original code expect different kind of inputs and we also support different kind of separators, this is the motive I said we would have to create a common parser for different functions case we really decide that we need to split the input to give more information for our users.

thiagoftsm:
More one thing, please, remember that these functions are used in other parts of the code, for example, we use them to parse `bind to` option.

stelfrag:
Risking to break existing code (calls from other locations), I would also consider expanding the check to allow `tcp:` and `tcp://` and update the docs.

---

Having said that, since this is causing a lot of discussion already, we may defer the "quick fix" and analyze it a bit more.  

After all, a "fatal" error such as this should give enough information to the caller not to try to reconnect -- it won't work.

The user can fix the configuration for now and we can analyze this further

thiagoftsm:
> Risking to break existing code (calls from other locations), I would also consider expanding the check to allow `tcp:` and `tcp://` and update the docs.
> 
> Having said that, since this is causing a lot of discussion already, we may defer the "quick fix" and analyze it a bit more.
> 
> After all, a "fatal" error such as this should give enough information to the caller not to try to reconnect -- it won't work.
> 
> The user can fix the configuration for now and we can analyze this further

I agree with everything written here! :)

amoss:
I wondered about handling `tcp://` directly as well. The `//` is not legal syntax for any kind of host that we specify. But let's not kick the can down the road indefinitely:

* `connect_to_this` is called from exactly one place in the entire repo (`connect_to_one_of`).
* `connect_to_one_of` is called from exactly two places in the entire repo, the one that we are investigating and `./backends/backends.c:997`.

So the only places that a change could have an impact is here, and in the backend config (or the new exporters config that is replacing it). Tolerating 'tcp://' is backward-compatible -> there are no legal configurations that it could impact.

So a better fix all-around is to remove the new warning message and change the `connect_to_this` code to this:
```
    if(strncmp(host, "tcp:", 4) == 0) {
        host += 4;
        protocol = IPPROTO_TCP;
        socktype = SOCK_STREAM;
        if(host[0]=='/' && host[1]=='/')
           host += 2;
    }
```

Then update the string in the docs to match and fix the comment so that it matches the documentation and the functionality in the code.

Thoughts?

thiagoftsm:
> 
> Thoughts?

I think you missed my initial comment, when netdata calls `connect_to_one_of` it is passing only `tcp:` , so it is not possible to do exactly what you are saying, please, try to run the configuration that raises error in another branch and you will see the motive the error messages was placed there. ;)

amoss:
When you say initial comment, I guess you mean this from the middle of the discussion:

> We do this inside the function connect_to_this, if and only if, we do not have a '/' in the connection, case we have we send only tcp: to the function, so we would need to create a parser common for both functions to avoid dupplication.

So I've checkout your pull request, set up a test condition and walked through it all several times. Sure, it is true that "tcp:" arrives at `connect_to_this`. But this is not what we should be fixing. I then set up two tests on the original code. In the first case I use
```
    destination = mechani.se:19999
```
which produces this:
```
2019-11-11 18:39:50: netdata DEBUG : STREAM_SENDER[loki] : (0759@libnetdata:connect_to_this): Original def='mechani.se:19999'
2019-11-11 18:39:50: netdata DEBUG : STREAM_SENDER[loki] : (0813@libnetdata:connect_to_this): Attempting connection to host = 'mechani.se', service = '19999', interface = '', protocol = 6 (tcp = 6, udp = 17)
2019-11-11 18:39:50: netdata DEBUG : STREAM_SENDER[loki] : (0661@libnetdata:connect_to_this): Address info: host = '185.217.0.188', service = '19999', ai_flags = 0x00, ai_family = 2 (PF_INET = 2, PF_INET6 = 10), ai_socktype = 1 (SOCK_STREAM = 1, SOCK_DGRAM = 2), ai_protocol = 6 (IPPROTO_TCP = 6, IPPROTO_UDP = 17), ai_addrlen = 16 (sockaddr_in = 16, sockaddr_in6 = 28)
2019-11-11 18:39:50: netdata DEBUG : STREAM_SENDER[loki] : (0673@libnetdata:connect_to_this): ai_addr = sin_family: 2 (AF_INET = 2, AF_INET6 = 10), sin_addr: '185.217.0.188', sin_port: '19999'
2019-11-11 18:39:50: netdata DEBUG : STREAM_SENDER[loki] : (0737@libnetdata:connect_to_this): Connected to '185.217.0.188' on port '19999'.
```
and in the second case I used:
```
    destination = tcp:mechani.se:19999
```
which produces this:
```
2019-11-11 18:47:30: netdata DEBUG : STREAM_SENDER[loki] : (0759@libnetdata:connect_to_this): Original def='tcp:mechani.se:19999'
2019-11-11 18:47:30: netdata DEBUG : STREAM_SENDER[loki] : (0813@libnetdata:connect_to_this): Attempting connection to host = 'mechani.se', service = '19999', interface = '', protocol = 6 (tcp = 6, udp = 17)
2019-11-11 18:47:30: netdata DEBUG : STREAM_SENDER[loki] : (0661@libnetdata:connect_to_this): Address info: host = '185.217.0.188', service = '19999', ai_flags = 0x00, ai_family = 2 (PF_INET = 2, PF_INET6 = 10), ai_socktype = 1 (SOCK_STREAM = 1, SOCK_DGRAM = 2), ai_protocol = 6 (IPPROTO_TCP = 6, IPPROTO_UDP = 17), ai_addrlen = 16 (sockaddr_in = 16, sockaddr_in6 = 28)
2019-11-11 18:47:30: netdata DEBUG : STREAM_SENDER[loki] : (0673@libnetdata:connect_to_this): ai_addr = sin_family: 2 (AF_INET = 2, AF_INET6 = 10), sin_addr: '185.217.0.188', sin_port: '19999'
2019-11-11 18:47:30: netdata DEBUG : STREAM_SENDER[loki] : (0737@libnetdata:connect_to_this): Connected to '185.217.0.188' on port '19999'.
```
In both of these correct cases the `definition` argument contains enough information for `connect_to_this` to make the connection. Our target should be that in this case:
```
    destination = tcp://mechani.se:19999
```
we see the same - that enough information arrives at the `connect_to_this` call in order to make the connection. Then the two-line change that I proposed will fix the user's problem and make the product easier to use. When I look into why it only receives "tcp:" I took a deeper look at the code in `connect_to_one_of` and I really don't understand why it is there at all. If I make the following change (i.e. remove everything that looks like it should not be there):
```
 /*       const char *e = s;

        // skip path, moving both s(tart) and e(nd)
        if(*e == '/')
            while(!isspace(*e) && *e != ',') s = ++e;

        // skip separators, moving both s(tart) and e(nd)
        while(isspace(*e) || *e == ',') s = ++e;

        // move e(nd) to the first separator
        while(*e && !isspace(*e) && *e != ',' && *e != '/') e++;

        // is there anything?
        if(!*s || s == e) break;

        if (e && e[0] == '/' && e[1] == '/') {
            error("A destination specified %s does not have the correct format [PROTOCOL:]HOST[%%INTERFACE][:PORT]"
                  , s);
            sock = -1;
            break;
        }

        char buf[e - s + 1];
        strncpyz(buf, s, e - s);*/
        if(reconnects_counter) *reconnects_counter += 1;
        sock = connect_to_this(destination, default_port, timeout);
        if(sock != -1) {
            if(connected_to && connected_to_size) {
                strncpy(connected_to, destination, connected_to_size);
                connected_to[connected_to_size - 1] = '\0';
            }
            break;
        }
        //s = e;
```
Then I see the following in the logs:
```
2019-11-11 19:01:45: netdata DEBUG : STREAM_SENDER[loki] : (0759@libnetdata:connect_to_this): Original def='tcp://mechani.se:19999'
2019-11-11 19:01:45: netdata DEBUG : STREAM_SENDER[loki] : (0813@libnetdata:connect_to_this): Attempting connection to host = '//mechani.se', service = '19999', interface = '', protocol = 6 (tcp = 6, udp = 17)
```
Which then looks like it can be fixed properly, and indeed, the two-line fix that I proposed earlier will now work.

There are always two ways to fix something:
1. Make it more complex to hide the problem.
2. Make it simpler to remove the problem.

It's possible that there is some complexity that I've missed in the code that I've removed, in which case please explain it to me - I'm still not up to speed on everything that we do. But otherwise I think this is a clean fix to the problem and we should try to do (2) where possible, instead of (1).

Does this make sense?

thiagoftsm:
Your suggestion is not considering one thing that is very important for this function. Case you open the file `stream.conf` you will have the following comments:

```
    # Where is the receiving netdata?
    # A space separated list of:
    #
    #      [PROTOCOL:]HOST[%INTERFACE][:PORT][:SSL]
    #
```

As you can see the function ` connect_to_one_of`  will try to connect to one of the hosted that are listed in front `destination`, so the changes that you are suggesting will bring other  possible errors case I have destinations like these:

```
destination = tcp:192.168.0.14:19999:SSL tcp:192.168.0.14:19999 unix:192.168.0.12
destination = tcp:192.168.0.14:19999:SSL,tcp:192.168.0.14:19999,unix:192.168.0.12
destination = tcp:192.168.0.14:19999:SSL/tcp:192.168.0.14:19999/unix:192.168.0.12
```

Pay attention for the fact that all 3 options are valid for stream and backend.

We know this function is not considering to have `//` and I added a message here due this, like @stelfrag says we are probably changing our focus. Case you are thinking "Why did you not change backends.c and avoid to have messages sent to backends like you did with stream?", well they do not have the same structure so I changed the code that I know better.

Please, let me know case you have more doubts.

stelfrag:
I agree with simple-fixing things if at all possible (and maintain compatibility) and adding new feature (extending valid syntax ) at the same time. 

In fact it's much better if code is removed to archive that :-)

But his turns to a rewrite (not a bad thing!) deserves an issue by itself, urgency status and 1 pt :-)

thiagoftsm:
> There are always two ways to fix something:
>     1. Make it more complex to hide the problem.

We never want something like this.

>     2. Make it simpler to remove the problem.

We wanna do this without to break compatibility.

> It's possible that there is some complexity that I've missed in the code that I've removed, in which case please explain it to me - I'm still not up to speed on everything that we do. But otherwise I think this is a clean fix to the problem and we should try to do (2) where possible, instead of (1).

I wrote an answer before explaining the complexity missed. ;)

> Does this make sense?

Everything that you wrote also make sense and I am sure that the answer I gave before this will help you to understand.

amoss:
Wow...

Ok, so we have a "feature" which is not mentioned at all in https://docs.netdata.cloud/streaming/, but it is "hinted" at in the comments in `stream.conf`:
```
    # A space separated list of:
    #
    #      [PROTOCOL:]HOST[%INTERFACE][:PORT][:SSL]
    #
```

And we have a piece of code that does something else entirely (adds commas and slashes as separators). Are the extra separators really a "feature" or are they a "bug"?

If we really want to preserve existing behaviour (which doesn't seem important for an undocumented feature) then adding another layer of checks and errors is still a bad fix for the problem.

The existing error message should correctly report that it cannot connect to a host called "tcp:".

amoss:
I've wiped out my existing review as you fixed those specific problems, but I don't think that the proposed fix of adding another layer of complexity should be done. I'll approve it if you fix the error message as I suggested above. If not I'm sure two other people will approve it.

We are deep into diminishing returns here in terms of the effort to expend on closing a year-old bug. I would suggest that cleaning it up (removing some of the extras like the undocumented separators, and fixing the existing error message) is an improvement and leave it there.

thiagoftsm:
> Ok, so we have a "feature" which is not mentioned at all in https://docs.netdata.cloud/streaming/, but it is "hinted" at in the comments in `stream.conf`:

Yes, we have this, I believe we keep it to keep backward compatibility.

> ```
>     # A space separated list of:
>     #
>     #      [PROTOCOL:]HOST[%INTERFACE][:PORT][:SSL]
>     #
>     # If many are given, the first available will get the metrics.
> ```
> 
> And we have a piece of code that does something else entirely (adds commas and slashes as separators). Are the extra separators really a "feature" or are they a "bug"?

I think it is the same answer that I commented few lines above.

> If we really want to preserve existing behaviour (which doesn't seem important for an undocumented feature) then adding another layer of checks and errors is still a bad fix for the problem.
> 
> The existing error message should correctly report that it cannot connect to a host called "tcp:".

We cannot to connect with `tcp://`, I will add this. ;)

thiagoftsm:
Hi @amoss ,

I thought a bit and you convinced me to do not touch socket.c, instead this I wrote the error message that gives more information for our users inside backend.c . :D

backends/backends.c
```diff
@@ -1124,7 +1124,11 @@ void *backends_main(void *ptr) {
                 }
             }
             else {
-                error("BACKEND: failed to update database backend '%s'", destination);
+                if (!strstr(destination,"//")) {
+                    error("BACKEND: failed to update database backend '%s'", destination);
+                } else {
+                    error("BACKEND: failed to connect to destination '%s', please remove '//'.", destination);
```
ilyam8: it is likely i missing context, but

> `failed to update database backend`

This looks like runtime error, is it? (i think `database` word is not needed btw).


> `failed to connect to destination '%s', please remove '//'.`

This looks like initialization error, is it?

thiagoftsm: The first is a run time error as you described. I kept the word `database`, because it was present in the original message.
And you are also correct about the second, because case the wrong string is present, it won't be able to connect.

ilyam8: Why it even try to connect if `dst` is wrong? My question is how come runtime and init errors are live in the same place? I dont ask to fix/do something, just noticed strange thing imo.

thiagoftsm: The goal of this PR is to give messages to clarify what is happening, I agree that to connect again is an error, I did not create any other PR to fix this, because we already had defined that we will have the new `exporter engine` that will be responsible to obsolete this code.
When the user created #4591, there was an error description only for `stream`, but while I was working, I discovered that the error could also happen with `backend`, this is the motive I extended the PR.

ilyam8: Well, to be honest your fix is too specific (just check for `//`)

> If you specify the stream destination host like tcp://examle.com netdata tells you that the connect failed, instead of an more appropriate syntax error

The fix should be:
 - introduce `validate_destination` function and validate destination. Report syntax error if syntax is invalid.

ilyam8: @amoss what do you think? The problem in `https://github.com/netdata/netdata/issues/4591` is lack of appropriate error message in case of invalid `destination` parameter. So we need to validate it first and log error if it is invalid. 

Do you think current fix is ok?

amoss: Well... there are two competing concerns here:
1. We have spent far too much time on this issue already, in comparison to the value that it delivers to users and relative to the other work that we have to do.
2. I explained the nice way to fix it back in November.

There are now two possibilities:
1. Fix it quickly
2. Fix it in the cleanest way possible (by removing the use of the alternative separators and documenting the change).

@thiagoftsm / @ilyam8 Which of the two possibilities do you prefer?

ilyam8: imo this PR doesnt fix https://github.com/netdata/netdata/issues/4591 and i would suggest to not merge it as it is

https://github.com/netdata/netdata/issues/4591 is low prio, can be fixed later

thiagoftsm: @ilyam8  from #4591 we have the following request:

```
Some kind of better error message, that tells you what the problem is, instead of sending one down the wrong rabbit hole (I was sure it was a network problem until i attached gdb)
```

What am I missing here?
@amoss I think that now we can do the quick fix  and in the future when it is necessary to do a big change inside the daemon (This will happen when we change pthreads to libuv ).

ilyam8: Ok, i will try to make it clear.

Current logic: 
 - do not validate destination
 - try to connect
 - if connection failed log `failed to update database backend '%s' destination`

Problem in https://github.com/netdata/netdata/issues/4591 - **lack of syntax validation** of `destination` parameter. Therefore netdata will try to connect to anything.

___

Example: `destination = ?!?!??!?!`

```
2020-01-20 09:51:06: netdata INFO  : BACKENDS : enabled backend for host 'localhost'
2020-01-20 09:51:06: netdata ERROR : BACKENDS : BACKEND: failed to update database backend '?!?!??!?!'
```

> BACKEND: failed to update database backend '?!?!??!?!'

As you see your `check for // in the end` doesnt do.

___

Problem: lack of syntax validation of `destination` parameter.

Soultion: add syntax validation of `destination` parameter.

Logic should be:

 - ~~do not~~ validate destination
 - if destination is invalid, log appropriate message and exit
 - try to connect
 - if connection failed log `failed to update database backend '%s' destination`

amoss: I don't understand why removing the extra separators and fixing it cleaning requires a major change to the daemon that should be delayed until pthreads are converted to libuv - could you explain why this is?

thiagoftsm:
@ilyam8 and @amoss I convert the original issue request for a generic request, I also changed the place that the error messages were raised for a more generic place that allow us to show all arguments given.
@ilyam8 Netdata will raise the error message different times until it reaches the maximum specified in the configuration, the default is 10.

ilyam8:
@thiagoftsm pleae update the PR title to reflect changes in the PR

thiagoftsm:
> @thiagoftsm pleae update the PR title to reflect changes in the PR

Done, thank you for the suggestion.

libnetdata/socket/socket.c
```diff
@@ -105,14 +105,78 @@ char *strdup_client_description(int family, const char *protocol, const char *ip
...
+ *
+ * We are considering valid hostname, addresses that follow the specifications of `hostname` manual
+ * http://man7.org/linux/man-pages/man7/hostname.7.html
+ *
+ * @param host the host that will be used to connect
+ *
+ * @return 1 case it is a valid hostname and 0 otherwise.
+ */
+int is_valid_tcp_hostname(char *host) {
+    int ret = 1;
```
ilyam8: Should this be `1` by default? What if `!host || !*host`? Does this function expect host to not be empty?

thiagoftsm: By steps, this function is called from the line 885 and there is already a test for NULL in the line 880, so it does not need to test for Null. I decided do not move the code at 880 inside this function, because the error messages for each case is different.
About the default value to be, it was a good catch. It is not necessary ,so I removed it with the last commit. Thank you!

ilyam8: > By steps, this function is called from the line 885 and there is already a test for NULL in the line 880, so it does not need to test for Null. 

I saw it, but this is **dependence on the order excution** which is not safe.  If function expects `host` to not be null it should be in the function description i guess.

thiagoftsm: Now I got your point of view, you are correct, somebody could invert the order and brings a new but for Netdata. I moved the code inside the new function and I did all necessaries adjusts for this.
Thank you again!

libnetdata/socket/socket.c
```diff
@@ -808,9 +879,17 @@ int connect_to_this(const char *definition, int default_port, struct timeval *ti
 
     debug(D_CONNECT_TO, "Attempting connection to host = '%s', service = '%s', interface = '%s', protocol = %d (tcp = %d, udp = %d)", host, service, interface, protocol, IPPROTO_TCP, IPPROTO_UDP);
 
-    if(!*host) {
-        error("Definition '%s' does not specify a host.", definition);
-        return -1;
+    int test = test_tcp_hostname(host);
```
amoss: The `host` still points to the beginning of the string, not the part that contains the hostname.

I tested it with:
```
    destination = tcp://mechani.se:19999
```
And the error that I get in the logs is:
```
2020-01-24 09:33:32: netdata ERROR : STREAM_SENDER[loki] : Definition 'tcp:' does not specify a host.
```

backends/backends.c
cosmix: I agree with @ilyam8 and @amoss here. First of all, I also don't see why we need to wait for the libuv transition to fix the input validation. And I think @ilyam8 succinctly presented what the logic should be here.

libnetdata/socket/socket.c
thiagoftsm: The last commit brought the message that has been parsed instead of the message that was parsed.

backends/backends.c
thiagoftsm: Thanks a lot to express your opinion here.
I changed the code for we connect only when the destination address is valid. When it is not valid we will never connect and we will write only one time the error inside `error.log`.
About to try to reconnect when we fail one time, I am keeping the rule that is already present.

backends/backends.c
```diff
@@ -997,7 +998,7 @@ void *backends_main(void *ptr) {
                 // usec_t start_ut = now_monotonic_usec();
                 size_t reconnects = 0;
 
-                sock = connect_to_one_of(destination, default_port, &timeout, &reconnects, NULL, 0);
+                sock = connect_to_one_of(destination, &do_not_reconnect, default_port, &timeout, &reconnects, NULL, 0);
```
stelfrag: Are you actually using `do_not_reconnect` to not try reconnection? Or am I missing something?

stelfrag: Never mind. I see it's in the low lever connect (the different variable name fooled me for a second) :smile:

libnetdata/socket/socket.c
```diff
@@ -846,18 +927,28 @@ int connect_to_one_of(const char *destination, int default_port, struct timeval
...
+        else
+            connect = 1;
+
+        if(connect) {
+            char buf[e - s + 1];
+            strncpyz(buf, s, e - s);
+            if(reconnects_counter) *reconnects_counter += 1;
+            sock = connect_to_this(buf, s, default_port, timeout);
+            if(sock == -2)
+                *do_not_connect |=  (1U<<i);
```
stelfrag: When I specify a wrong host (as detected by `test_tcp_hostname`) it returns -1 and the `do_not_connect` is not set
example would be to try to stream to `192.168.0._1` 

```
2020-02-13 13:49:59: netdata ERROR : STREAM_SENDER[debian1] : One of the values specified is not valid host = '192.168.122._1', service = '19999', interface = '', protocol = 6
2020-02-13 13:49:59: netdata ERROR : STREAM_SENDER[debian1] : STREAM debian1 [send to 192.168.122._1:19999]: failed to connect
```

stelfrag:
When I setup a slave and specify

```
destination = tcp://192.168.122.1:19999
```

I get 
```
2020-02-13 14:06:14: netdata INFO  : STREAM_SENDER[debian1] : STREAM debian1 [send to tcp://192.168.122.1:19999]: connecting...
2020-02-13 14:06:14: netdata ERROR : STREAM_SENDER[debian1] : The first definition 'tcp://192.168.122.1:19999' of the list does not specify a host.
2020-02-13 14:06:14: netdata ERROR : STREAM_SENDER[debian1] : One of the values specified is not valid host = '��', service = '19999', interface = '', protocol = 6
2020-02-13 14:06:14: netdata ERROR : STREAM_SENDER[debian1] : STREAM debian1 [send to tcp://192.168.122.1:19999]: failed to connect
2020-02-13 14:06:19: netdata INFO  : STREAM_SENDER[debian1] : STREAM debian1 [send to tcp://192.168.122.1:19999]: connecting...
2020-02-13 14:06:19: netdata ERROR : STREAM_SENDER[debian1] : One of the values specified is not valid host = '��', service = '19999', interface = '', protocol = 6
2020-02-13 14:06:19: netdata ERROR : STREAM_SENDER[debian1] : STREAM debian1 [send to tcp://192.168.122.1:19999]: failed to connect
```

libnetdata/socket/socket.c
thiagoftsm: Probably the name is not good, but this is a variable used to identify the address that is invalid and we won't connect again. This is used as a flag, where each one of the 32 positions identify one of the 32 possible address in the list that tried to reconnect. Every time that the address is invalid(return -2 ) I enable the bit and I do not try to reconnect again. Should we rename the variable?

backends/backends.c
thiagoftsm: Probably the name is not good, but this is a variable used to identify the address that is invalid and we won't connect again. This is used as a flag, where each one of the 32 positions identify one of the 32 possible address in the list that tried to reconnect. Every time that the address is invalid(return -2 ) I enable the bit and I do not try to reconnect again. Should we rename the variable?

PR #7308, opened 2019-11-12
CMocka tests for Issue 7274

### LINKED ISSUE #7274: Complete CMocka unit-testing for request processing in the web-server
#### Summary
Issue #7229 concluded with a PoC that we mock enough of the pieces of netdata to perform testing inside `web_client.c` at a very low-level. This complements the high-level tester the fuzzer performs after issue #7163. For the low-level unit-tests we must ensure:

- [x] Extend the testdriver to generate partial requests.
- [x] The processing handles partial reception of requests correctly (i.e. `web_client->flags` is correctly set on partial / complete requests.
- [ ] The information required for further processing is correctly extracted: url, method, `cookie1`, `cookie2`, `origin` and `user_agent`.
- [x] Compliance with RFC2616 *and* RFC7230 (including the notes on implementation). This implies that we must process `\n` even if they are missing `\r` in the string, otherwise we expose ourselves to naked newlines in URLs.
- [x] Check correct decoding of escaped characters within URLs.
- [x] Take a dataset of exploit URLs with weird escaping and null-behaviour and put the generic patterns that cover them into the unit-tests (@amoss has one collected from an `nginx` instance already).
- [x] Correct decoding of the URL w.r.t. to path components, and that the mocked despatch point for the API is called correctly.

There is an additional issue that we cannot unit-test yet - the correct parameter decoding and passing to the API. This cannot be unit-tested yet as it is embedded throughout the API processing code and needs to be replaced with a single decoding point first. This will happen in #7164 .

The behaviour of STREAM requests will not be in the scope of this issue - we will mock the entry point to the stream processing and *only* check that the HTTP method is correctly detected.

#### Expected result
Complete set of unit-tests for the http request processing.

amoss:
##### Summary
Fixes #7274 

##### Component Name
web/api/tests

##### Additional Information
Still experimenting with gluing CMocka onto a large set of tests.


web/server/web_client.c
```diff
@@ -1266,7 +1266,7 @@ static inline void web_client_send_http_header(struct web_client *w) {
                 while((bytes = SSL_write(w->ssl.conn, buffer_tostring(w->response.header_output), buffer_strlen(w->response.header_output))) < 0) {
                     count++;
                     if(count > 100 || (errno != EAGAIN && errno != EWOULDBLOCK)) {
-                        error("Cannot send HTTP headers to web client.");
+                        error("Cannot send HTTPS headers to web client.");
```
thiagoftsm: Hi @amoss ,

I will give a suggestion for you, OpenSSL can give you more details about the error as you can see in the documentation (https://netdata-cloud.slack.com/), please, call SSL_get_error() together with ERR_error_string(), for an example you can see https://github.com/netdata/netdata/blob/master/backends/backends.c#L1022 .

amoss: This is a good point but I don't want to complicate this issue by changing the code being tested as well. In the next sprint I'll be rewriting the URL parsing and then I can take a look at the error codes.

thiagoftsm: All right!

web/api/tests/web_api.c
```diff
@@ -126,43 +211,234 @@ static void build_request(struct web_buffer *wb, const char *url, bool use_cr, s
...
 {
     buffer_free(w->response.data);
     buffer_free(w->response.header);
     buffer_free(w->response.header_output);
     free(w);
-    free(localhost);
 }
 
+// ---------------------------------- Parameterized test-families -----------------------------------------------------
+// There is no way to pass a parameter block into the setup fixture, we would have to patch CMocka and maintain it
```
vlvkobal: [cmocka_unit_test_prestate_setup_teardown(f, setup, teardown, state)](https://api.cmocka.org/group__cmocka__exec.html#ga07ea44503d0f6c84c460b4a878cf31d5)

vlvkobal:
How do you run the tests? I get
```
PASS: libnetdata/tests/str2ld_testdriver
PASS: libnetdata/storage_number/tests/storage_number_testdriver
./test-driver: line 107: 19972 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: web/api/tests/web_api_testdriver
FAIL: web/api/tests/valid_urls_testdriver
============================================================================
Testsuite summary for netdata v1.16.1-677-g14fae887
============================================================================
# TOTAL: 4
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  2
# XPASS: 0
# ERROR: 0
============================================================================
See ./test-suite.log
============================================================================
```

vlvkobal:
You forgot to add the tests to `CMakeLists.txt`.

web/api/tests/web_api.c
amoss: The state is passed into the tests if nothing is returned from the setup, it is not passed into the setup. (L3441) in https://git.cryptomilk.org/projects/cmocka.git/tree/src/cmocka.c

You are correct about what the docs say, but it did not do that when I tested it and the source of CMocka seems to disagree with their documentation.

amoss:
> How do you run the tests? I get
> 
> ```
> PASS: libnetdata/tests/str2ld_testdriver
> PASS: libnetdata/storage_number/tests/storage_number_testdriver
> ./test-driver: line 107: 19972 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
> FAIL: web/api/tests/web_api_testdriver
> FAIL: web/api/tests/valid_urls_testdriver
> ============================================================================
> Testsuite summary for netdata v1.16.1-677-g14fae887
> ============================================================================
> # TOTAL: 4
> # PASS:  2
> # SKIP:  0
> # XFAIL: 0
> # FAIL:  2
> # XPASS: 0
> # ERROR: 0
> ============================================================================
> See ./test-suite.log
> ============================================================================
> ```

That looks correct - both test-drivers currently have failing cases in the URL parser. I am running them either with `make check` or directly via e.g. `web/api/tests/web_api_testdriver`. If you look in `web/api/tests/web_api_testdriver.log` then you can see the individual cases that fail (and the same in `web/api/tests/valid_urls_testdriver.log`. I have not thought of a good way to filter this.

amoss:
> You forgot to add the tests to `CMakeLists.txt`.

I've added them in. You may get some (minor) conflicts when you merge the exporting engine unit tests.

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,771 @@
...
+    localhost = NULL;
+}
+
+static void absolute_url(void **state)
+{
+    (void)state;
+
+    if (localhost != NULL)
+        free(localhost);
+    localhost = malloc(sizeof(RRDHOST));
```
thiagoftsm: At netdata we create localhost only when it starts, here you are creating for the second time inside this file, cannot this give the wrong vision about what netdata does?

thiagoftsm: Another difference, we allocate it with callocz instead mallocz.

amoss: I think unit testing code operates under different constraints to the main application code. This code is not intended to be executed at "run-time" for the application, only at "build-time". The reasoning behind flushing out the `localhost` allocation is to guarantee that we don't have state leakage between tests. We could think of each individual test as an application lifetime of its own, so we do things multiple times within a test-driver that we would only do once within the application itself.

The lack of `_z` calls is to prevent code in the system being tested being called as part of the testing environment itself. In this case I think the l`ocalhost` is unused but we need a dummy parameter for some of the procedures being called.

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,771 @@
...
+
+    web_client_process_request(w);
+
+    assert_string_equal(w->decoded_query_string, "?blah");
+    destroy_web_client(w);
+    free(localhost);
+    localhost = NULL;
+}
+
+/* Any number of blank lines before the request should be ignored according to the RFC */
```
thiagoftsm: Netdata follows the section 5 of this RFC https://tools.ietf.org/html/rfc2616, you are using another here that probably updated it, please, can you bring the RFC number in this comment?

thiagoftsm: We are using section 5.1:

```
   The Request-Line begins with a method token, followed by the
   Request-URI and the protocol version, and ending with CRLF. The
   elements are separated by SP characters. No CR or LF is allowed
   except in the final CRLF sequence.

        Request-Line   = Method SP Request-URI SP HTTP-Version CRLF
```

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,771 @@
...
+    char debug[160];
+    repr(debug, sizeof(debug), w->response.data->buffer, w->response.data->len);
+    printf("->%s\n", debug);
+
+    char expected_url_repr[160];
+    repr(expected_url_repr, sizeof(expected_url_repr), "inf\no\t", 6);
+
+    web_client_process_request(w);
+
+    assert_int_equal(w->response.code, HTTP_RESP_BAD_REQUEST);
```
thiagoftsm: Apache returns 404 in situations like this, you are expecting a 400, why do you want something different from Apache?

web/api/tests/valid_urls.c
amoss: Finally! I was starting to think that I had imagined reading this during the sprint :)

It's in RFC2616, but it is in an earlier section (4.1):
```
In the interest of robustness, servers SHOULD ignore any empty
   line(s) received where a Request-Line is expected. In other words, if
   the server is reading the protocol stream at the beginning of a
   message and receives a CRLF first, it should ignore the CRLF.
```

web/api/tests/valid_urls.c
amoss: It comes from section 5.1 in RFC2616 (emphasis added):
```
The Request-Line begins with a method token, followed by the
   Request-URI and the protocol version, and ending with CRLF. The
   elements are separated by SP characters. **No CR or LF is allowed
   except in the final CRLF sequence.**
```
Apache is wrong :)

thiagoftsm: Interesting :O

thiagoftsm:
Hi @amoss ,

I got the same error that @vlvkobal  reported when I ran `make check`:

```
PASS: libnetdata/tests/str2ld_testdriver
PASS: libnetdata/storage_number/tests/storage_number_testdriver
./test-driver: line 107:  2901 Segmentation fault      "$@" > $log_file 2>&1
FAIL: web/api/tests/web_api_testdriver
FAIL: web/api/tests/valid_urls_testdriver
============================================================================
Testsuite summary for netdata v1.18.1-177-g66c4f04b
============================================================================
# TOTAL: 4
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  2
# XPASS: 0
# ERROR: 0
============================================================================
See ./test-suite.log
============================================================================
```

After this I decided to run the software directly (./web/api/tests/web_api_testdriver) and I got these following errors:

```
[  FAILED  ] /api/v1/info@7,2/368
[  FAILED  ] /api/v1/info@7,3/368
[  FAILED  ] /api/v1/info@7,368/368

 66 FAILED TEST(S)
Segmentation fault
```

Finally I ran the software with `valgrind` to see more detail about the last error and I got this:

```

 66 FAILED TEST(S)
==2968== Invalid write of size 8
==2968==    at 0x405678: sprintf (stdio2.h:36)
==2968==    by 0x405678: __wrap_debug_int (web_api.c:126)
==2968==    by 0x40ECD9: buffer_free (buffer.c:378)
==2968==    by 0x405114: destroy_web_client (web_api.c:234)
==2968==    by 0x405D05: api_info_launcher (web_api.c:378)
==2968==    by 0x405D05: main (web_api.c:440)
==2968==  Address 0x445ffe is in a rw- anonymous segment
==2968== 
==2968== 
==2968== Process terminating with default action of signal 11 (SIGSEGV)
==2968==  Access not within mapped region at address 0x446000
==2968==    at 0x405678: sprintf (stdio2.h:36)
==2968==    by 0x405678: __wrap_debug_int (web_api.c:126)
==2968==    by 0x40ECD9: buffer_free (buffer.c:378)
==2968==    by 0x405114: destroy_web_client (web_api.c:234)
==2968==    by 0x405D05: api_info_launcher (web_api.c:378)
==2968==    by 0x405D05: main (web_api.c:440)
==2968==  If you believe this happened as a result of a stack
==2968==  overflow in your program's main thread (unlikely but
==2968==  possible), you can try to increase the size of the
==2968==  main thread stack using the --main-stacksize= flag.
==2968==  The main thread stack size used in this run was 8388608.
==2968== 
==2968== HEAP SUMMARY:
==2968==     in use at exit: 237,936,106 bytes in 26,618 blocks
==2968==   total heap usage: 31,218 allocs, 4,600 frees, 249,068,420 bytes allocated
==2968== 
==2968== LEAK SUMMARY:
==2968==    definitely lost: 0 bytes in 0 blocks
==2968==    indirectly lost: 0 bytes in 0 blocks
==2968==      possibly lost: 0 bytes in 0 blocks
==2968==    still reachable: 237,936,106 bytes in 26,618 blocks
==2968==         suppressed: 0 bytes in 0 blocks
==2968== Rerun with --leak-check=full to see details of leaked memory
==2968== 
==2968== Use --track-origins=yes to see where uninitialised values come from
==2968== For lists of detected and suppressed errors, rerun with: -s
==2968== ERROR SUMMARY: 7653 errors from 30 contexts (suppressed: 0 from 0)
Segmentation fault
```

All these tests that reported these errors were executed on Slackware current, kernel 4.19.84, gcc 9.2.0, cmocka 1.1.15 and valgrind 3.15

web/api/tests/valid_urls.c
thiagoftsm: Thank you for the explanation!

web/api/tests/valid_urls.c
thiagoftsm: To complement your original text I found this RFC https://tools.ietf.org/html/rfc7231#section-6.5.1 that describes the errors.

thiagoftsm:
After to see the errors on Linux, I decided to run the same tests on FreeBSD 12.0-p12, this time I got the following output:

```
`libnetdata/tests/str2ld_testdriver' is up to date.
`libnetdata/storage_number/tests/storage_number_testdriver' is up to date.
`web/api/tests/web_api_testdriver' is up to date.
`web/api/tests/valid_urls_testdriver' is up to date.
make  check-TESTS
PASS: libnetdata/tests/str2ld_testdriver
PASS: libnetdata/storage_number/tests/storage_number_testdriver
FAIL: web/api/tests/web_api_testdriver
FAIL: web/api/tests/valid_urls_testdriver
============================================================================
Testsuite summary for netdata v1.18.1-177-g66c4f04b
============================================================================
# TOTAL: 4
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  2
# XPASS: 0
# ERROR: 0
============================================================================
See ./test-suite.log
============================================================================
*** Error code 1

Stop.
make[4]: stopped in /usr/home/thiago/Netdata/netdata
*** Error code 1

Stop.
make[3]: stopped in /usr/home/thiago/Netdata/netdata
*** Error code 1

Stop.
make[2]: stopped in /usr/home/thiago/Netdata/netdata
*** Error code 1

Stop.
make[1]: stopped in /usr/home/thiago/Netdata/netdata
*** Error code 1

Stop.
```

Please, pay attention for the fact that when compiled with `clang` on FreeBSD the `segmentation fault` did not happen.

When I run again the direct test (./web/api/tests/web_api_testdriver) I also got the expected results, please fix the code for `gcc`, your code has problems when compiled with it and due the fact this is the main compiler on Linux world I cannot approve this PR.

I verified the code comparing with RFCs and everything is correct as you said.

Congratulations!

amoss:
Hi @thiagoftsm,

Thanks for taking a deeper look into the failing cases. There are two different kinds of problem here:
* Test cases that currently fail.
* Test cases that crash.

I'm expecting there to be test cases that currently fail (as we have not yet reworked the `web_client.c` to fix the problems) but there should not be any test cases that crash. The code was developed on linux with gcc so I'm not sure yet what is suppressing the problem on my system. Great problem report - I'm going to dive into the stack trace first to find out which part of the test cleanup is causing the problem.

> When I run again the direct test (./web/api/tests/web_api_testdriver) I also got the expected results, please fix the code for gcc, your code has problems when compiled with it and due the fact this is the main compiler on Linux world I cannot approve this PR.

Absolutely - I'll fix the crash and run it through valgrind.

web/api/tests/web_api.c
```diff
@@ -126,43 +214,239 @@ static void build_request(struct web_buffer *wb, const char *url, bool use_cr, s
...
+    }
+
+    web_client_process_request(def->instance);
+
+    if (def->prefix_len == def->full_len)
+        assert_int_equal(def->instance->flags & WEB_CLIENT_FLAG_WAIT_RECEIVE, 0);
+    else
+        assert_int_equal(def->instance->flags & WEB_CLIENT_FLAG_WAIT_RECEIVE, WEB_CLIENT_FLAG_WAIT_RECEIVE);
+    assert_int_equal(def->instance->mode, WEB_CLIENT_MODE_NORMAL);
+    printf("decoded: %s\n", def->instance->decoded_query_string);
```
vlvkobal: In every test it prints an empty string. It just clutters the output. Do you really need it?

web/api/tests/web_api.c
```diff
@@ -18,72 +54,124 @@ RRDHOST *__wrap_rrdhost_find_by_hostname(const char *hostname, uint32_t hash)
...
 
 int __wrap_web_client_api_request_v1(RRDHOST *host, struct web_client *w, char *url)
 {
-    printf("api requests: %s\n", url);
+    char url_repr[160];
+    repr(url_repr, sizeof(url_repr), url, strlen(url));
+    printf("web_client_api_request_v1(url=\"%s\")\n", url_repr);
+    check_expected_ptr(host);
+    check_expected_ptr(w);
+    check_expected_ptr(url_repr);
```
vlvkobal:
```
[  ERROR   ] --- No entries for symbol url_repr.
web/api/tests/web_api.c:80: error: Could not get value to check parameter url_repr of function __wrap_web_client_api_request_v1
web/api/tests/web_api.c:288: note: Previously declared parameter value was declared here

```
You should use `check_expected_ptr(url)`.

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,777 @@
...
+}
+
+int __wrap_web_client_api_request_v1(RRDHOST *host, struct web_client *w, char *url)
+{
+    char url_repr[160];
+    repr(url_repr, sizeof(url_repr), url, strlen(url));
+    printf("web_client_api_request_v1(url=\"%s\")\n", url_repr);
+    check_expected_ptr(host);
+    check_expected_ptr(w);
+    check_expected_ptr(url_repr);
```
vlvkobal: `check_expected_ptr(url)`

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,777 @@
...
+    char debug[4096];
+    repr(debug, sizeof(debug), w->response.data->buffer, w->response.data->len);
+    printf("-> \"%s\"\n", debug);
+
+    //char expected_url_repr[4096];
+    //repr(expected_url_repr, sizeof(expected_url_repr), def->url_out_repr, strlen(def->url_out_repr));
+
+    expect_value(__wrap_web_client_api_request_v1, host, localhost);
+    expect_value(__wrap_web_client_api_request_v1, w, w);
+    expect_string(__wrap_web_client_api_request_v1, url_repr, "info");
```
vlvkobal: url_repr -> url

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,777 @@
...
```
vlvkobal: url_repr -> url

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,777 @@
...
```
vlvkobal: url_repr -> url

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,777 @@
...
+    repr(expected_url_repr, sizeof(expected_url_repr), "inf\no\t", 6);
+
+    web_client_process_request(w);
+
+    assert_int_equal(w->response.code, HTTP_RESP_BAD_REQUEST);
+
+    destroy_web_client(w);
+    free(localhost);
+}
+static void newline_in_url(void **state)
```
vlvkobal: You should add `expext_` macros for `__wrap_web_client_api_request_v1` in `newline_in_url`, `cr_in_url`, and `bad_version`.

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,777 @@
...
+    buffer_strcat(w->response.data, "GET / HTTP/1.1\r\n\r\n");
+
+    char debug[4096];
+    repr(debug, sizeof(debug), w->response.data->buffer, w->response.data->len);
+    printf("-> \"%s\"\n", debug);
+
+    //char expected_url_repr[4096];
+    //repr(expected_url_repr, sizeof(expected_url_repr), def->url_out_repr, strlen(def->url_out_repr));
+
+    expect_string(__wrap_mysendfile, filename, "/");
```
vlvkobal: An excessive check. Leads to "__wrap_mysendfile: '%s' parameter still has values that haven't been checked."

web/api/tests/valid_urls.c
```diff
@@ -0,0 +1,777 @@
...
+    buffer_strcat(w->response.data, "GET // HTTP/1.1\r\n\r\n");
+
+    char debug[4096];
+    repr(debug, sizeof(debug), w->response.data->buffer, w->response.data->len);
+    printf("-> \"%s\"\n", debug);
+
+    //char expected_url_repr[4096];
+    //repr(expected_url_repr, sizeof(expected_url_repr), def->url_out_repr, strlen(def->url_out_repr));
+
+    expect_string(__wrap_mysendfile, filename, "//");
```
vlvkobal: An excessive check. Leads to "__wrap_mysendfile: '%s' parameter still has values that haven't been checked."

web/api/tests/web_api.c
amoss: I've fixed it the other way around, by changing the `expect_string` to use url_repr. The reasoning here is that it makes it easier to check for nulls and high-ascii characters in the interface while it is still changing. Later on it might more sense to flip both to `url`.

web/api/tests/valid_urls.c
amoss: > Generally, you shouldn't have the "No entries for symbol" errors in the CMocka output.

This is not correct. Each of those cases is a case that fails because the tested code does the wrong thing. It should not call the API despatch point because it should not consider the URL to be valid (in each case it is malformed). The unit testing code will work correct (by asserting the return code) when the tested code is fixed.

I thought it would be cleaner to have the opposite of an `expect_` in the test to indicate that we should not hit the mock during the test but I could not see a way to do that.

web/api/tests/valid_urls.c
amoss: See comment above - in these cases the tested code should not call `web_client_api_request_v1`.

web/api/tests/valid_urls.c
amoss: I think the check is correct - the code being tested is calling the wrong result. It should treat everything outside of the `/api/v1` subtree as a file-request, so in this case it should call `mysendfile` with the argument supplied.

web/api/tests/valid_urls.c
amoss: I think the check is correct - the code being tested is calling the wrong result. It should treat everything outside of the /api/v1 subtree as a file-request, so in this case it should call mysendfile with the argument supplied.

web/api/tests/valid_urls.c
vlvkobal: I don't understand it. `__wrap_web_client_api_request_v1, url_repr` doesn't have the `url_repr` parameter. How your `expect_` could help in identifying wrong calls? It fails because of the parameter name mismatch.

amoss:
@vlvkobal I think the new version fixes the cases that you pointed out (apart from the ones that I've explained above are valid failures in the code under test). The logs were broken. It is possible that you think the new output is more "cluttered" -> I'm open to suggestions on cleaning it up.

web/api/tests/web_api.c
```diff
@@ -126,43 +220,238 @@ static void build_request(struct web_buffer *wb, const char *url, bool use_cr, s
...
+};
+
 static void api_info(void **state)
 {
+    (void)state;
+    struct test_def *def = (struct test_def *)shared_test_state;
+    shared_test_state = def->next;
+
+    if (def->prev != NULL && !def->prev->completed && strlen(log_buffer) > 0) {
+        printf("Log of failing case %s:\n", def->prev->name);
```
vlvkobal: It is printed in the next test. Is it possible to make it appear in the failed test itself?

web/api/tests/web_api.c
vlvkobal: Now it's much better. Passed tests are clean.

web/api/tests/web_api.c
vlvkobal: You can't check `expect_string` ahainst `url_repr` because `__wrap_web_client_api_request_v1` doesn't have such parameter.

thiagoftsm:
I ran again the tests and this time we are not having more segmentation faults, but I notice the number of failed tests reduced from 60 to 58 : 

```
[  FAILED  ] /api/v1/info@7,2/368
[  FAILED  ] /api/v1/info@7,3/368
[  FAILED  ] /api/v1/info@7,368/368

 58 FAILED TEST(S)
``` 

I am considering that this reduction is happening, because bugs were fixed @amoss ,please, can you confirm this?

web/api/tests/web_api.c
amoss: The control-flow in CMocka does not allow that. When the test fails control does not come back to the unit test.

web/api/tests/valid_urls.c
amoss: The `expect_` macros are not limited to parameters, they can be applied to arbitrary values in the procedure being tested. In this case there is an exact 1:1 mapping from `url` to `url_repr`, but the `repr()` output is easier to work with in cases where the strings contain null characters.

amoss:
> I am considering that this reduction is happening, because bugs were fixed @amoss ,please, can you confirm this?

I think that it is an error. The eight cases that have started passing are full requests where there are no `\r` in the buffer. I know from the code that these cases cannot be validated correctly so something has broken in the testing harness. Well spotted. I'll try and work out what it is and come back with a fix.

PR #7312, opened 2019-11-13
Updating the Travis pipeline (issue 7189)

### LINKED ISSUE #7189: Travis builds are not showing the same warnings as developer systems.
### Bug report summary

The PR for #7116 (7132) introduced a warning into the build on linux that was not visible in the Travis build logs (or on the dev system where the PR was written). Travis needs to enable more warnings, and the docs for debugging should include the same build configuration that we use for checking PRs.

##### OS / Environment
Linux (any, but verified on debian 10).

netdata v1.18.1-23-g01aaa909

##### Component Name
area/tests
area/dcs

##### Steps To Reproduce
CFLAGS="-O1 -ggdb -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1" ./netdata-installer.sh --disable-lto --dont-wait

produces the following in the build output:
```
libnetdata/simple_pattern/simple_pattern.c:276:22: warning: format '%d' expects argument of type 'int', but argument 8 has type 'unsigned int' [-Wformat=]
     debug(debug_type,"dump_pattern(%p) child=%p next=%p mode=%d match=%s", root, root->child, root->next, root->mode,
```

##### Expected behavior
Travis builds should include the options for warnings from above. 
The Travis build output for a PR should produce this warning.
The nightly build should be free from warnings (change the `%d` to a `%u` to test the change).
The docs for debugging netdata should be consistent with the Travis environment (check that we document these CFLAGS).

amoss:
##### Summary
Fixes #7189 

The current build doesn't match our conventions during development on the team. We have a different build configuration in Travis to the one that we should use to keep the code warning free. There is no indication on how the code matches our coding style.

* Added a linting pass. This is non-blocking but will output a measure of how much each .c and .h file deviates from our current `.clang-format`.

* Changed the standard built to include all of the warning that we are using in dev.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,6 +104,10 @@ jobs:
       name: Run shellchecking on BASH
       script: shellcheck --format=gcc $(find . -name '*.sh.in' -not -iwholename '*.git*')
 
+      # Check if any of the C code deviates from our coding style
+    - name: Check coding style
+      script: find . -type f -name '*.[ch]' | while read filename; do echo -ne "$filename: "; diff <(clang-format -style=file $filename) $filename | wc -l; done
+
       # This falls under same stage defined earlier
     - name: Run checksum checks on kickstart files
       script: ./tests/installer/checksums.sh
@@ -120,7 +124,7 @@ jobs:
 
       name: Standard netdata build
       script: fakeroot ./netdata-installer.sh --install $HOME --dont-wait --dont-start-it --enable-plugin-nfacct --enable-plugin-freeipmi --disable-lto
-      env: CFLAGS='-O1 -DNETDATA_INTERNAL_CHECKS=1 -DNETDATA_VERIFY_LOCKS=1'
+      env: CFLAGS="-O1 -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1"
       after_failure: post_message "TRAVIS_MESSAGE" "<!here> standard netdata build is failing (Still dont know which one, will improve soon)"
 
     - name: Docker container build process (alpine installation)
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,7 +124,7 @@ jobs:
 
       name: Standard netdata build
       script: fakeroot ./netdata-installer.sh --install $HOME --dont-wait --dont-start-it --enable-plugin-nfacct --enable-plugin-freeipmi --disable-lto
-      env: CFLAGS="-O1 -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1"
+      env: CFLAGS='-O1 -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1'
       after_failure: post_message "TRAVIS_MESSAGE" "<!here> standard netdata build is failing (Still dont know which one, will improve soon)"
 
     - name: Docker container build process (alpine installation)
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,7 +106,7 @@ jobs:
 
       # Check if any of the C code deviates from our coding style
     - name: Check coding style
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -120,7 +124,7 @@ jobs:
 
       name: Standard netdata build
       script: fakeroot ./netdata-installer.sh --install $HOME --dont-wait --dont-start-it --enable-plugin-nfacct --enable-plugin-freeipmi --disable-lto
-      env: CFLAGS='-O1 -DNETDATA_INTERNAL_CHECKS=1 -DNETDATA_VERIFY_LOCKS=1'
+      env: CFLAGS="-O1 -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1"
```
knatsakis: Hi @amoss,

Please change the CFLAGS for the unit tests as well

knatsakis: Also, please keep the single quotes, unless double quotes are needed for interpolation

amoss: We don't use those - they are not unit tests, they are a really bad idea. I think we should remove that stage completely until we are ready to switch on the CMocka unit tests - @cosmix?

Will change to single quotes.

cosmix: I agree @amoss. I think, however, that we should keep whatever is there until we can replace it. In other words, as soon as we have our initial CMocka suite for e.g. the exporting engine, replace it please. cc @vlvkobal @knatsakis

amoss: So I'll leave it in, but ignore the CFLAGS changes in that stage as it will not have any effect. I'm ready to push the docs changes to match but there is no Travis output yet. Can we check if I have horribly broken it.

.travis/clangformat.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
knatsakis: @amoss, could you please use `#!/usr/bin/env bash` here?

Thanks.

amoss: Will do.

daemon/main.c
```diff
@@ -921,8 +921,8 @@ int main(int argc, char **argv) {
                     {
                         char* stacksize_string = "stacksize=";
                         char* debug_flags_string = "debug_flags=";
-                        char* createdataset_string = "createdataset=";
-                        char* stresstest_string = "stresstest=";
+                        //char* createdataset_string = "createdataset="; Unused?
+                        //char* stresstest_string = "stresstest=";  Unused?
```
mfundul: How are those unused? They are used in line 945.

.travis.yml
mfundul: Which is not a unit test? In any case, `-Wformat-signedness` does not work with old compilers and can break with many distributions.

daemon/main.c
amoss: It seems that our internal build instructions don't include the dbengine... I'll update the Travis pipeline again and fix this differently.

amoss: I cannot reply underneath your other comment (github ui?)

> Which is not a unit test? In any case, -Wformat-signedness does not work with old compilers and can break with many distributions.

This is not changing how we package the software for distribution so it doesn't matter. This is how we check that incoming PRs compile on the known build environment inside the Travis pipeline.

amoss: @knatsakis Please could you look at https://travis-ci.com/netdata/netdata/jobs/256057119 ?

Because we do not have libJudy installed the "standard build" on Travis is not our standard build. The installer disables the dbengine when Judy is not available, so we are not checking that our standard configuration builds.

Also it seems that CMocka is missing from that environment as well - so we could not enable the new unit tests.

How do we fix this?

knatsakis: @amoss can you please create a new issue describing which dependencies you want me to add?

amoss: I can, or if possible I could fix it here. Is it the install block in L19-L37 of the `.travis.yml` file? I can push a commit that adds the missing packages from https://docs.netdata.cloud/packaging/installer/ ?

I'm looking at the "This is how to do it by hand" section and the later " Netdata DB engine can be enabled when these are installed" sections.

amoss:
Everything mentioned in the reviews has been fixed. Please can @mfundul and @knatsakis re-check and change their review status when they are happy.

Normally two reviews are sufficient, but in this case I would appreciate it if @cosmix and @vlvkobal could also have a look at the latest state and make sure that we have not missed anything.

The configuration changes work - the pre-requisites for the dbengine are now installed on the VM, and as you can see in the build logs we are now checking what we ship as a default + the extra warnings for a dev environment. In the unit-testing stage this means that the dbengine unit tests are also being run now (successfully).

PR #7318, opened 2019-11-14
- Retrieve current affinity of the process and make sure not to

### LINKED ISSUE #6741: [/proc/stat cpuidle] Avoid isolated CPUs

stelfrag:
start an idle thread (wake_cpu_thread) if the core is not  in
  the allowed cpu set.

##### Summary
Fixes #6741
Prevent isolated CPUs from being used by the idle thread 

This is done by reading the current affinity of the process and making sure
that the excluded cores do not spawn an idle thread (`wake_cpu_thread`)

### DIFF
--- a/collectors/proc.plugin/proc_stat.c
+++ b/collectors/proc.plugin/proc_stat.c
@@ -993,6 +993,13 @@ int do_proc_stat(int update_every, usec_t dt) {
         for(core = 0; core < schedstat_cores_found; core++) {
             if(unlikely(!(cpuidle_charts[core].active_time - cpuidle_charts[core].last_active_time))) {
                 pthread_t thread;
+                cpu_set_t global_cpu_set;
+
+                sched_getaffinity(getpid(), sizeof(cpu_set_t), &global_cpu_set);
+
+                if (unlikely(!CPU_ISSET(core, &global_cpu_set))) {
+                    continue;
+                }
 
                 if(unlikely(pthread_create(&thread, NULL, wake_cpu_thread, (void *)&core)))
                     error("Cannot create wake_cpu_thread");
--- a/collectors/proc.plugin/proc_stat.c
+++ b/collectors/proc.plugin/proc_stat.c
@@ -995,11 +995,13 @@ int do_proc_stat(int update_every, usec_t dt) {
                 pthread_t thread;
                 cpu_set_t global_cpu_set;
 
-                sched_getaffinity(getpid(), sizeof(cpu_set_t), &global_cpu_set);
-
-                if (unlikely(!CPU_ISSET(core, &global_cpu_set))) {
-                    continue;
+                if (likely(!pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &global_cpu_set))) {
+                    if (unlikely(!CPU_ISSET(core, &global_cpu_set))) {
+                        continue;
+                    }
                 }
+                else
+                    error("Cannot read current process affinity");
 
                 if(unlikely(pthread_create(&thread, NULL, wake_cpu_thread, (void *)&core)))
                     error("Cannot create wake_cpu_thread");


collectors/proc.plugin/proc_stat.c
```diff
@@ -993,6 +993,13 @@ int do_proc_stat(int update_every, usec_t dt) {
         for(core = 0; core < schedstat_cores_found; core++) {
             if(unlikely(!(cpuidle_charts[core].active_time - cpuidle_charts[core].last_active_time))) {
                 pthread_t thread;
+                cpu_set_t global_cpu_set;
+
+                sched_getaffinity(getpid(), sizeof(cpu_set_t), &global_cpu_set);
```
mfundul: You should check the returned value of `sched_getaffinity` before testing the `global_cpu_set` variable.

mfundul: Also it would be better to use `pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &global_cpu_set)` to be consistent with the `pthread_setaffinity_np()` call.

stelfrag: Ok I will change that!

Issue #7086, opened 2019-10-14
Agent Claiming - Core


gpapamathND:
As an Admin of Netdata Nodes, I would like to have the option to claim nodes in workspaces (and warRooms), as a proof of ownership and trust of the Netdata Nodes (by the Admins), as well as applying extra security.

**The main idea**
**`Claiming of Nodes`**: is a Node onboarding process in a workspace (and warRooms), that will be achieved by using a common invitation mechanism per workspace. The invitation mechanism, includes creation of a public claiming-token by the Admins of the Workspace. A Node, can subscribe to the invite (i.e. is allowed to be included in the Workspace & added automatically in WarRooms) with two ways:

1.  `Manually` - this is what most people with a few systems will do:
- Via the config of the Node, by setting manually the invitation token
- Via the installer of Node, by defining manually the invitation token, during the installation

2. `Automatically` - this is what most people with hundreds or even thousands of systems will do:
- Via a command line tool on the Node.

Implementation

This issue will serve as a description of the implementation work that is required to deliver claiming in the agent. In order to implement the design as agreed we have a skeleton `claim.sh` written in bash (this generates the keys and accesses the cloud API).

* Put `claim.sh.in` into the tree and add autotools to install it into the target file-system.
* Provide a `#define` with the path that can be used to call `exec()` from the agent.
* Patch up any paths in the `claim.sh` at install time, target file layout is: 
```
etc/netdata/claim.d/private.pem
etc/netdata/claim.d/public.pem
etc/netdata/claim.d/cloud_fullchain.pem
etc/netdata/claim.d/token
etc/netdata/claim.d/rooms
etc/netdata/claim.d/is_claimed
```
* Sort out file permissions.
* Provide support for three use-cases:
1. Script is called directly in the shell, then cli-tool is used to reload agent state.
2. Agent is run with command-line arguments (including token), `exec()` on the script.
3. Put the token in the file, run the agent with `-claim` and `exec()` the script.
* Resolve the hole in the design: do we delete the token after (3) if the claim is unsuccessful (may requrie some knowledege of target use under Ansible).
* Put return codes into the script (0 for success, non-0 for error).
* Uninstaller steps to clean up.

cakrit:
To understand what information needs to be sent from the agent side to the cloud during the claiming process, look at https://github.com/netdata/cloud-backend/issues/71

cakrit:
The todos are:
[...]

Issue #7052, opened 2019-10-10
[Spike] agent claiming


mfundul:

cosmix:
What is the definition of done here? @mfundul 

Suggestions from the planning:

1. A design document for the claiming on the agent side.
2. An estimation of the work needed to complete claiming on the agent side.

mfundul:
There are some problems in implementing https://github.com/netdata/netdata/issues/7086 .

There is the problem of authenticating the netdata agent identification. MACHINE_GUID is not enough because:
- MACHINE_GUID is not a secret, it can be visible in the network or known to other users. Thus it is not secure. Malicious netdata agents can copy the MACHINE_GUID and connect to netdata cloud hosts without authorization. Moreover, the original netdata agent will be locked out of netdata cloud which is a Denial-Of-Service attack.
- MACHINE_GUID can be easily cloned when creating VMs or containers by mistake, thus it is not unique.
- basing MACHINE_GUID on hardware IDs does not fix the uniqueness issue because the IDs themselves are often virtualized by the hypervisor as part of the virtual machine firmware. It is entirely dependent on the hypervisor type and configuration whether those will be unique or not.

Netdata cloud being a Certificate Authority can work around the issue of agent identification in the following way:
- Agents without a Certificate will create one before claiming but after installation.
- Netdata cloud remembers the public key for a given MACHINE_GUID. Any subsequent claiming requests **or** connection attempts for this MACHINE_GUID will be denied. The agent shall be advised to change its MACHINE_GUID.

However, there is still the possibility of cloning an agent after claiming, and as a result also cloning its Certificate. In this case, when netdata cloud detects a second connection attempt the agent is advised to create a new Certificate.

The agent also needs to get the invitation token in a secure and practical way. One way to do that is by using a new netdata configuration file with suitable permissions. This file will not be globally visible.

However, we also need a command line tool to interact with the agent at runtime, with some form of Inter-Process-Communication. I have rejected the following possibilities:

- POSIX pipes: they are unidirectional and as a result not practical to return a status. Moreover, they are multi-process unsafe and not portable to Windows.
- network sockets: there is no easy and obvious security model with networking even though its low performance is not likely to be an issue since we don't intend to use commands in rapid succession.
- message queues: we do not want asynchronous messages like a mailbox, we need real-time communication with the agent when it's online.
- shared memory: it's way too complicated to implement some sort of IPC queue, that is both multi-process safe and secure, from scratch.

[...]

PR #7325, opened 2019-11-17
Implement netdata command server and cli tool

### LINKED ISSUE #7269: Build cli tool for netdata
##### Feature idea summary
Build cli tool for netdata. We need a command server in the netdata agent and a cli utility so that we can communicate with the agent from the command line. Signals are insufficient to do that because they are limited in number, they do not carry any information and they are not cross-platform. This is described in #7052 and will help implement #7086.

mfundul:
##### Summary
Fixes #7269
##### Component Name
daemon
cli
##### Additional Information
Implement netdata command server and cli tool

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -628,6 +628,8 @@ set(DAEMON_FILES
         daemon/main.h
         daemon/signals.c
         daemon/signals.h
+        daemon/commands.c
+        daemon/commands.h
         daemon/unit_test.c
         daemon/unit_test.h
         )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -650,6 +650,12 @@ set(NETDATA_FILES
         ${WEB_PLUGIN_FILES}
         )
 
+set(NETDATACLI_FILES
+        daemon/commands.h
+        cli/cli.c
+        cli/cli.h
+        )
+
 include_directories(AFTER .)
 
 add_definitions(
@@ -784,6 +790,15 @@ IF(XENSTAT_LIBRARIES)
 ELSE()
     SET(ENABLE_PLUGIN_XENSTAT False)
 ENDIF()
+
+
+# -----------------------------------------------------------------------------
+# netdatacli
+
+add_executable(netdatacli config.h ${NETDATACLI_FILES})
+target_link_libraries (netdatacli libnetdata ${NETDATA_COMMON_LIBRARIES})
+target_include_directories(netdatacli PUBLIC ${NETDATA_COMMON_INCLUDE_DIRS})
+target_compile_options(netdatacli PUBLIC ${NETDATA_COMMON_CFLAGS})
 
 
 # -----------------------------------------------------------------------------
--- a/cli/cli.c
+++ b/cli/cli.c
@@ -1,0 +1,139 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include "cli.h"
+#include "../libnetdata/required_dummies.h"
+
+static uv_pipe_t client_pipe;
+uv_write_t write_req;
+uv_shutdown_t shutdown_req;
+
+static char command_string[MAX_COMMAND_LENGTH];
+static unsigned command_string_size;
+
+struct command_context {
+    uv_work_t work;
+    uv_stream_t *client;
+    cmd_t idx;
+    char *args;
+    char *message;
+    cmd_status_t status;
+};
+
+static void pipe_read_cb(uv_stream_t *client, ssize_t nread, const uv_buf_t *buf)
+{
+    if (0 == nread) {
+        fprintf(stderr, "%s: Zero bytes read by command pipe.\n", __func__);
+    } else if (UV_EOF == nread) {
+//      fprintf(stderr, "EOF found in command pipe.\n");
+        fprintf(stdout, "%s\n", command_string);
+    } else if (nread < 0) {
+        fprintf(stderr, "%s: %s\n", __func__, uv_strerror(nread));
+    }
+
+    if (nread < 0) { /* stop stream due to EOF or error */
+        (void)uv_read_stop((uv_stream_t *)client);
+    } else if (nread) {
+        size_t to_copy;
+
+        to_copy = MIN(buf->len, MAX_COMMAND_LENGTH - 1 - command_string_size);
+        strncpyz(command_string + command_string_size, buf->base, to_copy);
+    }
+    if (buf && buf->len) {
+        free(buf->base);
+    }
+}
+
+static void alloc_cb(uv_handle_t *handle, size_t suggested_size, uv_buf_t *buf)
+{
+    (void)handle;
+
+    buf->base = malloc(suggested_size);
+    buf->len = suggested_size;
+}
+
+static void shutdown_cb(uv_shutdown_t* req, int status)
+{
+    int ret;
+
+    (void)req;
+    (void)status;
+
+    /* Reset buffer to receive reply */
+    command_string_size = 0;
+    command_string[0] = '\0';
+
+    ret = uv_read_start((uv_stream_t *)&client_pipe, alloc_cb, pipe_read_cb);
+    if (ret) {
+        fprintf(stderr, "uv_read_start(): %s\n", uv_strerror(ret));
+        uv_close((uv_handle_t *)&client_pipe, NULL);
+        return;
+    }
+
+}
+
+static void pipe_write_cb(uv_write_t* req, int status)
+{
+    int ret;
+
+    (void)req;
+    (void)status;
+
+    ret = uv_shutdown(&shutdown_req, (uv_stream_t *)&client_pipe, shutdown_cb);
+    if (ret) {
+        fprintf(stderr, "uv_shutdown(): %s\n", uv_strerror(ret));
+        uv_close((uv_handle_t *)&client_pipe, NULL);
+        return;
+    }
+}
+
+static void connect_cb(uv_connect_t* req, int status)
+{
+    int ret;
+    uv_buf_t write_buf;
+    char *s;
+
+    (void)req;
+    if (status) {
+        fprintf(stderr, "uv_pipe_connect(): %s\n", uv_strerror(status));
+        fprintf(stderr, "Make sure the netdata service is running.\n");
+        exit(1);
+    }
+    command_string_size = 0;
+    command_string[0] = '\0';
+    s = fgets(command_string, MAX_COMMAND_LENGTH, stdin);
+    (void)s; /* We don't need input to communicate with the server */
+    command_string_size = strlen(command_string);
+
+    write_req.data = &client_pipe;
+    write_buf.base = command_string;
+    write_buf.len = command_string_size;
+    ret = uv_write(&write_req, (uv_stream_t *)&client_pipe, &write_buf, 1, pipe_write_cb);
+    if (ret) {
+        fprintf(stderr, "uv_write(): %s\n", uv_strerror(ret));
+    }
+//  fprintf(stderr, "COMMAND: Sending command: \"%s\"\n", command_string);
+}
+
+int main(int argc, char **argv)
+{
+    int ret;
+    static uv_loop_t* loop;
+    uv_connect_t req;
+
+    (void)argc;
+    (void)argv;
+    loop = uv_default_loop();
+
+    ret = uv_pipe_init(loop, &client_pipe, 1);
+    if (ret) {
+        fprintf(stderr, "uv_pipe_init(): %s\n", uv_strerror(ret));
+        return 1;
+    }
+    uv_pipe_connect(&req, &client_pipe, PIPENAME, connect_cb);
+
+    uv_run(loop, UV_RUN_DEFAULT);
+
+    uv_close((uv_handle_t *)&client_pipe, NULL);
+
+    return 0;
+}
--- a/cli/cli.c
+++ b/cli/cli.c
@@ -4,11 +4,16 @@
 #include "../libnetdata/required_dummies.h"
 
 static uv_pipe_t client_pipe;
-uv_write_t write_req;
-uv_shutdown_t shutdown_req;
+static uv_write_t write_req;
+static uv_shutdown_t shutdown_req;
 
 static char command_string[MAX_COMMAND_LENGTH];
 static unsigned command_string_size;
+
+static char response_string[MAX_COMMAND_LENGTH];
+static unsigned response_string_size;
+
+static int exit_status;
 
 struct command_context {
     uv_work_t work;
@@ -19,13 +24,51 @@ struct command_context {
     cmd_status_t status;
 };
 
+static void parse_command_reply(void)
+{
+    FILE *stream = NULL;
+    char *pos;
+    int syntax_error = 0;
+
+    for (pos = response_string ;
+         pos < response_string + response_string_size  && !syntax_error ;
+         ++pos) {
+        /* Skip white-space characters */
+        for ( ; isspace(*pos) && ('\0' != *pos); ++pos) {;}
+
+        if ('\0' == *pos)
+            continue;
+
+        switch (*pos) {
+        case CMD_PREFIX_EXIT_CODE:
+            exit_status = atoi(++pos);
+            break;
+        case CMD_PREFIX_INFO:
+            stream = stdout;
+            break;
+        case CMD_PREFIX_ERROR:
+            stream = stderr;
+            break;
+        default:
+            syntax_error = 1;
+            fprintf(stderr, "Syntax error, failed to parse command response.\n");
+            break;
+        }
+        if (stream) {
+            fprintf(stream, "%s\n", ++pos);
+            pos += strlen(pos);
+            stream = NULL;
+        }
+    }
+}
+
 static void pipe_read_cb(uv_stream_t *client, ssize_t nread, const uv_buf_t *buf)
 {
     if (0 == nread) {
         fprintf(stderr, "%s: Zero bytes read by command pipe.\n", __func__);
     } else if (UV_EOF == nread) {
 //      fprintf(stderr, "EOF found in command pipe.\n");
-        fprintf(stdout, "%s\n", command_string);
+        parse_command_reply();
     } else if (nread < 0) {
         fprintf(stderr, "%s: %s\n", __func__, uv_strerror(nread));
     }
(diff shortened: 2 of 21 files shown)


daemon/commands.h
```diff
@@ -0,0 +1,62 @@
...
+#define MAX_COMMAND_LENGTH 4096
+
+typedef enum cmd {
+    CMD_HELP = 0,
+    CMD_RELOAD_HEALTH,
+    CMD_SAVE_DATABASE,
+    CMD_LOG_ROTATE,
+    CMD_EXIT,
+    CMD_FATAL,
+    CMD_TOTAL_COMMANDS
```
amoss: Why are you putting this inside the enum and then adding BUGON later with a compile-time expression? It looks like it would be easier to just` #define (sizeof(command_info_array) / sizeof(command_info_t))`. Adding something that can go wrong, and then a `sanity_check` to stop it is not as simple as just not letting it go wrong.

daemon/commands.c
```diff
@@ -0,0 +1,539 @@
...
+    }
+    info("COMMAND: Sending reply: \"%s\"", reply_string);
+}
+
+cmd_status_t execute_command(cmd_t idx, char *args, char **message)
+{
+    cmd_status_t status;
+    cmd_type_t type = command_info_array[idx].type;
+
+    cmd_lock_by_type[type](idx);
```
amoss: Having mutexes on each command makes sense, as it stops a collision with the signal handling code in the daemon. This doesn't make any sense to me. Why do we want to stop two idempotent commands from running at the same time, or two high-priority commands.. ?
The one case it would make sense for if the command type is exclusive, it does not make sense for the the types.

daemon/commands.c
```diff
@@ -0,0 +1,539 @@
...
+    }
+    ret = uv_accept(server, (uv_stream_t *)client);
+    if (ret) {
+        error("uv_accept(): %s", uv_strerror(ret));
+        uv_close((uv_handle_t *)client, NULL);
+        freez(client);
+        return;
+    }
+
+    ++clients;
```
amoss: I didn't think that we supported multiple client connections - either this should trigger an error or we should separate state for multiple connections.

daemon/commands.c
```diff
@@ -0,0 +1,539 @@
...
+    }
+}
+
+static void pipe_read_cb(uv_stream_t *client, ssize_t nread, const uv_buf_t *buf)
+{
+    if (0 == nread) {
+        info("%s: Zero bytes read by command pipe.", __func__);
+    } else if (UV_EOF == nread) {
+        info("EOF found in command pipe.");
+        parse_commands(client);
```
amoss: If we only handle commands when the pipe closes then this will work like a batch interface instead of a CLI. Is this the intention?

daemon/commands.c
```diff
@@ -0,0 +1,539 @@
...
+{
+    (void)index;
+}
+
+static void pipe_write_cb(uv_write_t* req, int status)
+{
+    (void)status;
+    uv_pipe_t *client = req->data;
+
+    uv_close((uv_handle_t *)client, NULL);
```
amoss: This implies that the command server allows exactly one command per connection - but the positioning of the call-backs in the sequence, the mutexes and the dynamic allocation of memory inside `parse_commands` suggest otherwise.
Basic design questions:
* How many active connections are allowed?
* How many commands are allowed over each connection?

If the answers are "one" and "one" then the code is too complex and needs to be simplified. If the answers are different then it's unclear why the parsing is being triggered at the end of the pipe read and the connection closed after execution of a command.

daemon/commands.c
```diff
@@ -0,0 +1,539 @@
...
+{
+    char *message = NULL, *pos;
+    cmd_t i;
+    cmd_status_t status;
+    struct command_context *cmd_ctx;
+
+    status = CMD_STATUS_FAILURE;
+
+    for (i = 0 ; i < CMD_TOTAL_COMMANDS ; ++i) {
+        pos = strstr(command_string, command_info_array[i].cmd_str);
```
amoss: Why is this a substring?

daemon/commands.c
```diff
@@ -0,0 +1,539 @@
...
+        if (pos) {
+            cmd_ctx = mallocz(sizeof(*cmd_ctx));
+            cmd_ctx->work.data = cmd_ctx;
+            cmd_ctx->client = client;
+            cmd_ctx->idx = i;
+            cmd_ctx->args = pos + strlen(command_info_array[i].cmd_str);
+            cmd_ctx->message = NULL;
+
```
amoss: Is this where we would lock mutexes for the actual command and despatch into the daemon?

CMakeLists.txt
```diff
@@ -626,6 +626,8 @@ set(DAEMON_FILES
         daemon/main.h
         daemon/signals.c
         daemon/signals.h
+        daemon/commands.c
```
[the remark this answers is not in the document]
mfundul: This part is still [WIP]. We haven't solved the libuv dependency yet.

daemon/commands.h
mfundul: This enum indexes various data structures, it's not just a part of `command_info_array[]` (e.g. `execute_command(CMD_SAVE_DATABASE, NULL, NULL);` in `signals.c`). Hence it exists.

As a result, there is nothing stopping a developer from modifying the definition of `command_info_array[]` (e.g. add a line). The `sanity_check()` is the only way to catch the bug at compile time and preserve consistency.

daemon/commands.h
```diff
@@ -0,0 +1,62 @@
...
+} cmd_status_t;
+
+#define CMD_STATUS_SUCCESS_STR "SUCCESS"
+#define CMD_STATUS_FAILURE_STR "FAILURE"
+#define CMD_STATUS_BUSY_STR "BUSY"
+
+typedef enum cmd_type {
+    CMD_TYPE_EXCLUSIVE = 0, // No other command can run at the same time
+    CMD_TYPE_ORTHOGONAL,    // Other commands are allowed to run concurrently but calls to this command are serialized
+    CMD_TYPE_IDEMPOTENT,    // Any call to any command is allowed at the same time as its execution
```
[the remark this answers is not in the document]
mfundul: There is no implication about thread-safety. If the command is not re-entrant but idempotent then it's the command's responsibility to protect any critical sections. The command server will allow concurrent execution.

daemon/main.c
```diff
@@ -1285,6 +1285,11 @@ int main(int argc, char **argv) {
         else debug(D_SYSTEM, "Not starting thread %s.", st->name);
     }
 
+    // ------------------------------------------------------------------------
+    // Initialize netdata agent command serving from cli and signals
+
+    commands_init();
```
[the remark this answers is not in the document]
mfundul: I will have to add error messages depending on the return value.

cli/cli.c
```diff
@@ -0,0 +1,139 @@
...
+struct command_context {
+    uv_work_t work;
+    uv_stream_t *client;
+    cmd_t idx;
+    char *args;
+    char *message;
+    cmd_status_t status;
+};
+
+static void pipe_read_cb(uv_stream_t *client, ssize_t nread, const uv_buf_t *buf)
```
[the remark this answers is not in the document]
mfundul: They are handled fine. The shutdown will only affect the send side.

cli/cli.c
```diff
@@ -0,0 +1,139 @@
...
+    char *s;
+
+    (void)req;
+    if (status) {
+        fprintf(stderr, "uv_pipe_connect(): %s\n", uv_strerror(status));
+        fprintf(stderr, "Make sure the netdata service is running.\n");
+        exit(1);
+    }
+    command_string_size = 0;
+    command_string[0] = '\0';
```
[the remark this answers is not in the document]
mfundul: Because `fgets()` may fail.

cli/cli.c
```diff
@@ -0,0 +1,139 @@
...
+    s = fgets(command_string, MAX_COMMAND_LENGTH, stdin);
+    (void)s; /* We don't need input to communicate with the server */
+    command_string_size = strlen(command_string);
+
+    write_req.data = &client_pipe;
+    write_buf.base = command_string;
+    write_buf.len = command_string_size;
+    ret = uv_write(&write_req, (uv_stream_t *)&client_pipe, &write_buf, 1, pipe_write_cb);
```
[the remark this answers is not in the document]
mfundul: The `cli` will only handle one line of text. You call it again to run a different command.

daemon/commands.c
mfundul: Active connections can be many.

The parsing is triggered at the end because each command sends an EOF. The connection is closed only after the status is sent back to `cli`, which will be after the command execution.

daemon/commands.c
mfundul: It's a reader-writer lock. The commands are not stopped because they are readers. The only writer is a command of exclusive type . The mutexes exist only for orthogonal commands and they are command specific.

daemon/commands.c
mfundul: It works for now but we probably have to adjust this when we add more commands.

daemon/commands.c
mfundul: Any mutexes will be called after thread spawning so as to not block the event loop.

daemon/commands.c
mfundul: Each invocation of `cli` issue exactly one command.

daemon/commands.c
mfundul: We have separate state indeed.

daemon/commands.h
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#ifndef NETDATA_COMMANDS_H
+#define NETDATA_COMMANDS_H 1
+
+#ifdef _WIN32
+# define PIPENAME "\\\\?\\pipe\\netdata-cli"
```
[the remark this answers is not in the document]
mfundul: No.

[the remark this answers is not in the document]
mfundul: To clarify, netdata does not work with Windows, the porting will take many months, I am just trying to be proactive here to minimize the porting effort.

amoss:
There is no discussion of how many connections and how many commands-per-connection in the design that you proposed in #7269 / #7052.

* Why many connections, but only one command per connection? (what is the use-case for this)
* Why is there an exit command if it exits after one command by design?
* Why are you planning to spawn threads for commands? (again, what is the use-case for multiple commands in parallel)

mfundul:
> 
> 
> There is no discussion of how many connections and how many commands-per-connection in the design that you proposed in #7269 / #7052.
> 
>     * Why many connections, but only one command per connection? (what is the use-case for this)
> 

The "connection" is an implementation detail that has to do with the API of libuv (and its implementation underneath, UNIX sockets vs Windows named pipes). Let's just call it many commands.

The use-case is many outstanding commands, as in, many invocations of `netdatacli` so as to facilitate long running commands that are not mutually exclusive. Each invocation of `netdatacli` sends exactly one command. I would reverse the question and ask why do we need a `netdatacli` invocation to issue more than one commands?

>     * Why is there an exit command if it exits after one command by design?
> 

The `exit` command gracefully shuts down the netdata agent. All the commands have to do with `netdata` and not with `netdatacli` itself. Even `help` is not served by `netdatacli` but has to ask `netdata` to fetch a list of commands.

>     * Why are you planning to spawn threads for commands? (again, what is the use-case for multiple commands in parallel)

The idea is that there will be many commands that may run in parallel. One command could take minutes to complete. The event loop of the command server can not block or it will stop serving any other commands that the user may attempt to invoke. Hence, each command is dispatched to the libuv workqueue and the command server event loop remains responsive.

One example let's say is, a cron job issued a database dump and immediately afterwards netdata cloud initiated an agent claiming procedure.

thiagoftsm:
A few suggestions:

1 - After to run the `help` command the agent cannot exit, I think this is the unique command that can never exit when we are inside the terminal.
2 - We need to have an option to specify the destination for `netdatacli`, it will be very interesting from an unique computer to send commands for all netdata inside a network.
3 - This is probably some wrong expectation from my side, but I would like to confirm, when I ran the the command `log rotate`, I was expecting to have more files inside my `/var/log/netdata`, but I continue with exactly the same files, for me this is an error, can you explain us what is expected from this command output?
4 - A suggestion, when we run `python` or `psql` cli commands, when we join the terminal the first thing that we receive is the version of the system that we are connecting, please, add a kind of welcome message that gives information about Netdata.

Finally, you are creating a nice tool, we were needing something like this! :)

daemon/commands.h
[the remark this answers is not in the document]
Ferroin: 'IDEMPOTENT" is not the right term here then, though I'm not sure what would be (`FULLY_CONCURRENT` maybe?).

Something being 'idempotent'  means that you can call it again with the exact same set of arguments without it causing any change in state, provided there has been no other change in state since the last time it was called.  While some of the commands may fit that description (I'd actually hope that as many as possible are, as it makes working with the API much saner), that really says nothing about the concurrency requirements (except that it's easier to make them idempotent the less concurrent they are).

Ferroin:
> The exit command gracefully shuts down the netdata agent. All the commands have to do with netdata and not with netdatacli itself. Even help is not served by netdatacli but has to ask netdata to fetch a list of commands.

Could we name that command something other than `exit` so that it's less ambiguous what exactly it does?  Usually, when you have a protocol like this, `exit` as a command means you're closing the connection.

Same goes for the 'log rotate' command, the name implies it's rotating the logs itself, but what it's really doing (based on reading the code) is reopening the log files.

Also, I don't know that we really _need_ the `fatal` command unless we want it for debugging.  Most users aren't likely to ever use it (unless there's something I'm missing about `fatal()` being more reliable than a graceful exit, but at that point why not just send SIGABRT (which exists on Windows as well)).

amoss:
> The "connection" is an implementation detail that has to do with the API of libuv (and its 
> implementation underneath, UNIX sockets vs Windows named pipes). Let's just call it many commands.

It's not really an implementation detail, we can refer to it as "one invocation of the tool" and the question remains the same - why are we doing this way? Is this the best way to supply the functionality - what are the use-cases driving this?

> The use-case is many outstanding commands, as in, many invocations of netdatacli so as to 
> facilitate long running commands that are not mutually exclusive. Each invocation of netdatacli 
> sends exactly one command. I would reverse the question and ask why do we need a 
> netdatacli invocation to issue more than one commands?

Why are we not sending the commands as a sequence from a single invocation of the tool? This is conceptually simpler - there is no spawning of threads to handle things that may be concurrent - and the resulting code is simpler. We've made a decision not to do this way - so what use-case are we supporting by performing concurrent injection of commands rather than sequencing them and waiting for each to terminate?

amoss:
> The idea is that there will be many commands that may run in parallel. One command could 
> take minutes to complete. The event loop of the command server can not block or it will stop 
> serving any other commands that the user may attempt to invoke. Hence, each command is 
> dispatched to the libuv workqueue and the command server event loop remains responsive.

> One example let's say is, a cron job issued a database dump and immediately afterwards 
> netdata cloud initiated an agent claiming procedure.

So the use-case is that we have multiple CLI-invocations connecting to the agent in parallel for long-running jobs. Yet we have a design where they are all multiplexed onto the same pipe., and we use the EOF on the pipe to trigger the command processing. How are we distinguishing the command/response for each invocation?

Ferroin:
> Why are we not sending the commands as a sequence from a single invocation of the tool? This is conceptually simpler - there is no spawning of threads to handle things that may be concurrent - and the resulting code is simpler. We've made a decision not to do this way - so what use-case are we supporting by performing concurrent injection of commands rather than sequencing them and waiting for each to terminate?

Even if we were to allow command sequences like that, we still need to support concurrent execution, because it's well within reason that users might execute more than one command at once (at least, once we have a wider variety of commands).

amoss:
> Even if we were to allow command sequences like that, we still need to support concurrent 
> execution, because it's well within reason that users might execute more than one command at 
> once (at least, once we have a wider variety of commands).

Yes this is true. I'm exploring whether we are spawning threads at the most appropriate place and whether the design around the use of a pipe is the most appropriate for the target use-cases. The example @mfundul gave above (about the database dump during an agent claiming process) is a very interesting case for considering how we should do this.

daemon/commands.h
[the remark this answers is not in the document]
mfundul: It's true what you're saying, the command server does not guarantee that commands of type `CMD_TYPE_IDEMPOTENT` will stop producing side-effects after multiple calls.

mfundul:
About commands such as `log rotate`, this PR just duplicated the existing commands from the agent (as triggered by signals) to the command server. The semantics are identical, If there are complaints about that we have to fix them in the signal handling code as well.

About the interactivity of `netdatacli`, I did not put any effort into that, to be honest. The only reason we have `netdatacli` instead of `nc` or `socat` is because we don't want to create additional dependencies in netdata and for Windows support (`netdatacli` uses libuv which is portable). The idea is that `netdatacli` is mostly going to be used by scripts to perform tasks (e.g. `echo "claim agent token MACHINE_GUID" | netdatacli`). This is going to be too much effort to be like `python`, `psql`, `targetcli` etc. The original intent was to assist with #7086.

We can definitely change some names that are unfortunate. `exit` and `IDEMPOTENT` will have to be something more fitting.

About the server pipe and `EOF`, no message goes over it directly. It works pretty much like a network listener, multiple new client streams are created from that server pipe and this is how we have context per thread and shutting the stream down does not affect the command server.

thiagoftsm:
I understood the goal of the software, when I gave the message suggestion it was because when  we run `netdatacli` we do not have any information on it, let me change the example to clarify my suggestion, we could bring some messages like `/etc/motd` when somebody runs the software, they would know for example that they can call `help` to understand what they have inside the software.

stelfrag:
Good work Marko !

The way this works (the client side) is just reading from the pipe and display on screen (just "string" feedback from the server about the command execution) and it works great (does exactly what it needs)

To get extended feedback from the server a quick hack would be to prefix each output from the server with an indicator for example:

- O <message>
- E <message>
- P <progress %> 
- X <int> 

So the netdatacli will stdout the "O" prefixed messages, stderr the E ones, and exit with the X code
The P could just make it spin to indicate progress (0 to 100)

CMakeLists.txt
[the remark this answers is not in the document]
amoss: This was resolved in the meeting today.

daemon/main.c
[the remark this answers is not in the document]
amoss: Made it void.

cli/cli.c
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    uv_pipe_connect(&req, &client_pipe, PIPENAME, connect_cb);
+
+    uv_run(loop, UV_RUN_DEFAULT);
+
+    uv_close((uv_handle_t *)&client_pipe, NULL);
+
+    return exit_status;
+}
```
thiagoftsm: Missing new line here.

amoss: This is not good tonight - please no syntax. This must be closed.

PR #7359, opened 2019-11-22
Tutorials to support v1.19 release

joelhans:
##### Summary

I apologize in advance for the roughness of these drafts. It was difficult to get everything set up properly so I could actually test the installation/configuration steps... especially Unbound. I haven't even had time to proofread these and find/fix typos. So, read at your own risk.

Another point that @ilyam8 brought up is the importance of the custom log parsing in the web log collector. I worry that the existing tutorial is already long enough, and it doesn't yet include anything to help users set up the custom parsing. So, perhaps we need a third tutorial that explains that process in-depth. I'm still working on understanding how the parsing works myself.

These need to be merged before the release, so they're of high priority.

##### Component Name

docs


docs/tutorials/unbound-metrics.md
```diff
@@ -0,0 +1,140 @@
...
+Next, make your `unbound.conf`, `unbound_control.key`, and `unbound_control.pem` files readable by Netdata.
+
+```bash
+sudo setfacl -m user:netdata:r unbound.conf
+sudo setfacl -m user:netdata:r unbound_control.key
+sudo setfacl -m user:netdata:r unbound_control.pem
+```
+
+Finally, ensure that your `unbound.conf` file is indented using spaces, not tabs, as our collector can't parse a
+configuration file with tabs. There are many ways to accomplish this, but here is one option:
```
ilyam8: our collector can parse it

https://github.com/netdata/go.d.plugin/blob/e198258d6d50c46e8299c5e961ac7c0b89f6ad94/modules/unbound/init.go#L174-L177

ilyam8: configs in testdata are intended with tabs

https://github.com/netdata/go.d.plugin/tree/master/modules/unbound/testdata

And we have tests to read/parse them

docs/tutorials/unbound-metrics.md
```diff
@@ -0,0 +1,140 @@
...
+can't have the same name.
+
+The collector supports both cumulative and non-cumulative modes. Visit Unbound's [statistics configuration
+documentation](https://www.nlnetlabs.nl/documentation/unbound/howto-statistics/) for details on enabling cumulative
+mode, if you're interested.
+
+> It is important you know whether your Unbound server is running in cumulative or non-cumulative mode, as a conflict
+> between modes will create incorrect charts.
+
+At this point, you're ready to edit the `unbound.conf` file according to your needs.
```
ilyam8: stock unbound.conf contains 1 predefined job

See
https://github.com/netdata/go.d.plugin/blob/e198258d6d50c46e8299c5e961ac7c0b89f6ad94/config/go.d/unbound.conf#L127-L136

If the user did

```
sudo setfacl -m user:netdata:r unbound.conf
sudo setfacl -m user:netdata:r unbound_control.key
sudo setfacl -m user:netdata:r unbound_control.pem
```

netbound collector will work after restart of netdata service

docs/tutorials/unbound-metrics.md
```diff
@@ -0,0 +1,140 @@
...
+```bash
+sed -i $'s/\t/    /g' *.conf
+```
+
+Finally, take note whether you're using Unbound in _cumulative_ or _non-cumulative_ mode. This will become relevant when
+configuring the collector.
+
+## Configure the Unbound collector module
+
+To use the Go version of the Unbound collector, you need to explicitly enable it. Open your `go.d.conf` configuration
```
ilyam8: This is not true for unbound. It will be enabled by default. I see no reason to disable it. @Ferroin?

docs/tutorials/unbound-metrics.md
ilyam8: If it is about configuring jobs in general, then nvm 😄

docs/tutorials/unbound-metrics.md
Ferroin: We need to keep in mind that some people will still be using the Python version of the collector, which does need the config to use spaces. Some distros that provide Netdata packages (Gentoo for one) do not yet support the Go plugin.

docs/tutorials/unbound-metrics.md
Ferroin: If the Go version of the collector can find the Unbound config file reliably (provided permissions are right) and parse it reliably, then yeah, I see no need to disable it by default.

joelhans:
@ilyam8 or @Ferroin — Can either of you take a screenshot of some Unbound charts that actually have metrics? I have it running on a VPS, but the charts are rather... boring.

Also, I've pushed fixes to address each of your comments.

ilyam8:
I will try to do it tomorrow (i will need a script to load unbound server).

ilyam8:
> brought up is the importance of the custom log parsing

It seems i said it wrong, i meant custom log format feature is important. There is no such thing as custom log parsing at the moment. It is still nginx/apache logs. But format can be any.

The feature you are referencing to is

> https://github.com/netdata/go.d.plugin/tree/master/modules/weblog#custom-fields-feature

Very simple.

docs/tutorials/collect-unbound-metrics.md
```diff
@@ -0,0 +1,134 @@
...
+    tls_cert: /etc/unbound/unbound_control.pem
+    tls_key: /etc/unbound/unbound_control.key
+```
+
+To see all the available options, see the default [unbound.conf
+file](https://github.com/netdata/go.d.plugin/blob/master/config/go.d/unbound.conf).
+
+## Tweak Unbound collector alarms
+
+**COMING SOON: Ilya is adding/updating alarms and I will finish this section then.**
```
ilyam8: No alarms today. We decided to add them later. I created the issue btw

https://github.com/NLnetLabs/unbound/issues/123

joelhans: Got it. We can update this tutorial after they're up.

PR #7395, opened 2019-12-02
network interface speed, duplex, operstate #5989

### LINKED ISSUE #5989: network interface speed, duplex, operstate
##### Feature idea summary

raise an alarm in case the network interface speed changes to a lower speed or is not equally to the fastest possible speed

##### Expected behavior

raise an alarm in case the network interface speed changes to a lower speed
raise an alarm in case the network interface speed is not equally to the fastest possible speed

stelfrag:
Fixes #5989
##### Summary
- Read and process /sys/class/net/XXXX/duplex and /sys/class/net/XXXX/operstate
    
    - Create custom variables duplex_state and operstate under the bandwidth chart
    
    - Map states to numeric values
      Duplex state (variable duplex_state)
    
        0 = unknown
        1 = half duplex
        2 = full duplex
    
      Operstate status map (variable operstate)
    
        0 = unknown
        1 = notpresent
        2 = down
        3 = lowerlayerdown
        4 = testing
        5 = dormant
        6 = up


ilyam8:
> duplex_state

i think `duplex` will do. We used to call it speed, duplex, not duplex state, speed state.

stelfrag:
> > duplex_state
> 
> i think `duplex` will do. We used to call it speed, duplex, not duplex state, speed state.

I will change that

stelfrag:
@thiagoftsm Can you please check if the internal variables as defined are the way they should be?

thiagoftsm:
> @thiagoftsm Can you please check if the internal variables as defined are the way they should be?

They are very very good :) , unfortunately, I do not have conditions to test now, because I am working on FreeBSD, but I will test this PR today yet.

collectors/proc.plugin/proc_net_dev.c
```diff
@@ -5,6 +5,19 @@
 #define PLUGIN_PROC_MODULE_NETDEV_NAME "/proc/net/dev"
 #define CONFIG_SECTION_PLUGIN_PROC_NETDEV "plugin:" PLUGIN_PROC_CONFIG_NAME ":" PLUGIN_PROC_MODULE_NETDEV_NAME
 
+// As defined in https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-net
+const char *operstate_names[] = { "unknown", "notpresent", "down", "lowerlayerdown", "testing", "dormant", "up" };
+
+static inline int get_operstate(char *operstate)
+{
+    for (int i = 0; i < (int) (sizeof(operstate_names) / sizeof(char *)); i++) {
```
Ferroin: This line appears to be causing issues in Travis: https://travis-ci.com/netdata/netdata/jobs/262680560#L4684

Based on the error message, the only change that should be needed is moving the declaration of `i` out of the for loop.

thiagoftsm: @stelfrag another possibility would be to add -std=c99 when we are compiling collectors, I remember that we decided to use c99 in our codes, but it looks like that we did not adjust the compiler options for this.

ilyam8: @vlvkobal ^^ 😄

vlvkobal: C99 support was already added to the master branch a few days ago.

Ferroin: A rebase seems to be in order then prior to merging. Given the lack of conflicts, it should just be a trivial one too.

vlvkobal: A merge of the master branch will be enough.

stelfrag:
Rebase and push a var definition outside a for loop to avoid warning

PR #7416, opened 2019-12-03
Added ability to get pod name from cgroup with kubectl in bare metal deployment

WoozyMasta:
##### Summary
In those cases when kubernetes and netdata are manually installed on bare metal, the`cgroup-name.sh` script cannot get the names of the containers due to empty variables:
* `$KUBERNETES_SERVICE_HOST`
* `$KUBERNETES_PORT_443_TCP_PORT`

Also missing file `/var/run/secrets/kubernetes.io/serviceaccount/token`

My edits add an alternative opportunity to get the name of the container if no variables are set and `kubelet` is running on the system and `kubectl` binary is available.

By default, `/etc/kubernetes/admin.conf` is used to access kubectl, if an error occurs during execution, an entry will be added to the log with the warning level, but if you specify the path in the variable `$KUBE_CONFIG`, then it will be used your custom config.

For example, you can use systemd daemon:
```
[Service]
Environment=KUBE_CONFIG=/etc/netdata/kubernetes.conf
```

##### Component Name
`collectors/cgroups.plugin/cgroup-name.sh.in`

##### Additional Information

before/after

### DIFF
--- a/collectors/cgroups.plugin/cgroup-name.sh.in
+++ b/collectors/cgroups.plugin/cgroup-name.sh.in
@@ -96,11 +96,14 @@ function k8s_get_name() {
 			cut -d' ' -f1
 			)"
 		elif ps -C kubelet >/dev/null 2>&1 && command -v kubectl >/dev/null 2>&1; then
-			if kubectl get pod --all-namespaces >/dev/null 2>&1; then
-				NAME="$(kubectl get pod --all-namespaces --output='json' | \
+			if [[ -z ${KUBE_CONFIG+x} ]]; then
+				KUBE_CONFIG="/etc/kubernetes/admin.conf"
+			fi
+			if kubectl --kubeconfig=$KUBE_CONFIG get pod --all-namespaces >/dev/null 2>&1; then
+				NAME="$(kubectl --kubeconfig=$KUBE_CONFIG get pod --all-namespaces --output='json' | \
 				jq -r '.items[] | select(.metadata.uid == "'$id'") | .metadata.name')"
 			else
-				warning "kubectl cannot get pod list, check for configuration file in $HOME/.kube/config"
+				warning "kubectl cannot get pod list, check for configuration file in $KUBE_CONFIG, or set this path to env \$KUBE_CONFIG"
 			fi
 		fi
 	else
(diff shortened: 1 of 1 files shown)


cakrit:
This looks awesome, though I didn't run it to test. Can you please fix the `shellcheck` error from the `"'$id'"` syntax? (SC2086, Double quote to prevent globbing and splitting)

WoozyMasta:
I turned off the shellcheck SC2086 for this alert, because the bash expansion of variable inside single quotes in a command, and single quotes are needed there.

cakrit:
Oh, a couple of more things:
- If you rebase this branch, the netlify errors will go way
- the cla is still complaining, probably because one of your commits wasn't done with the proper author. it would be great if you could `amend` the commit to pass to it the proper author, i.e. the one matching your github user. 

None of the two are really required, but they will resolve the unsuccessful checks.

amoss:
I don't have a k8s install to test this - @stelfrag do you already have an installation to test this on? If not I'll set one up towards the end of the sprint and test/review this.

cakrit:
The [Netlify errors are a fluke](https://community.netlify.com/t/sudden-failed-build-due-to-python-dependency/10081), merging.

jiuchen1986:
Hi, I wonder why following changes are brought in:

image

With this change, netdata on a node, which is one of kubernetes nodes, without any access to kubernetes cluster information will not show cgroups/containers ran by kubelet. Why did this change to couple netdata crgoups monitoring so tightly with kubernetes. Sometimes we just want to monitor containers no matter they are created by docker cli or kubelet.

ilyam8:
Well, good point. If i understand it right a cggroup is disabled if `cgroup-name` bash script returns `3`, doesnt sound good.

https://github.com/netdata/netdata/blob/885caf08dd41a770489f91db6d226260122e0b94/collectors/cgroups.plugin/sys_fs_cgroup.c#L1171-L1174

https://github.com/netdata/netdata/blob/885caf08dd41a770489f91db6d226260122e0b94/collectors/cgroups.plugin/cgroup-name.sh.in#L114-L117

ilyam8:
@jiuchen1986 please open an issue

jiuchen1986:
> @jiuchen1986 please open an issue

Open an issue #9287

Issue #7432, opened 2019-12-04
Clarify meaning of asterisk and question marks on Netdata distribution support matrix


ncmans:
The two tables in question:
* https://docs.netdata.cloud/packaging/distributions/#amd64-architecture
* https://docs.netdata.cloud/packaging/distributions/#x86-architecture

1. What is the meaning of question mark against a distro? Either define it or replace them with ✗ and N/A which are clearly understood.

2. What is the meaning of the asterisk against CentOS and RHEL distros? A footnote at the bottom of the table linked to the asterisk is useful.

thiagoftsm:
> What is the meaning of question mark against a distro? Either define it or replace them with ✗ and N/A which are clearly understood.

RHEL has subversion,  so the meaning is we are supporting all 7.X versions, the same thing happens with CentOS, that has the latest 7.X is `CentOS Linux release 7.7.1908`

ncmans:
Oh! I'm so used to seeing `X` that the asterisk completely threw me off.

ncmans:
> > What is the meaning of question mark against a distro? Either define it or replace them with ✗ and N/A which are clearly understood.
> 
> RHEL has subversion, so the meaning is we are supporting all 7.X versions, the same thing happens with CentOS, that has the latest 7.X is `CentOS Linux release 7.7.1908`

I assume you're talking about the asterisk (e.g. `6.*`) and not the question mark.

thiagoftsm:
Yes, the question mark is something that I cannot understand either, it looks like that this was not set yet, but it could be.

PR #7434, opened 2019-12-04
Replace ? with ✗ on distro support matrix

ncmans:
Question marks are used on tables in https://docs.netdata.cloud/packaging/distributions with no definite meaning.

Fixes https://github.com/netdata/netdata/issues/7432


ncmans:
> This replaces all question marks with Xs. In certain columns, the question mark does make sense, it means that we haven't tested it yet. In CI: Smoke testing and CI: Testing it doesn't make sense, either we have it in there, or not.

@cakrit Where does it make sense?

cakrit:
> @cakrit Where does it make sense?

Where we haven't tested the installation method. e.g. for RHEL 8.x, we haven't tested kickstart.sh (or apparently haven't tested much of anything), so we don't know if it works or not => We can't put a check or an X, as X means that it was tested and doesn't work.

cosmix:
I think @cakrit has a point here, but I also find that the original description by @ncmans is valid:

> with no definite meaning.

is something we should improve. This does not mean replacing the question marks with ✗, but it does mean clarifying the semantics. A small legend perhaps?

ncmans:
Closing as I don't have time ATM to look at this — please re-open if need to

PR #7439, opened 2019-12-05
Add LABEL_SOURCE enum to label struct

cakrit:
So we can tell where the labels came from.

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -1274,8 +1274,8 @@ int main(int argc, char **argv) {
        Leave the code commented out to support easy testing until #7410 is complete (and then delete it).
 
     netdata_rwlock_wrlock(&localhost->labels_rwlock);
-    struct label *l = create_label("HostLabel1","Höst Låbel");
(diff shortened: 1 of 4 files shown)


cakrit:
@thiagoftsm some comments somehow appeared in this PR, but perhaps you had added them on another one. I had done a screw up with the branches, remotes and commits in the various repos, this one is clean, so please add a review here.

amoss:
> I remembered another important thing now, please, add uint32_t variables for we create hashes for the values of these two pointers.

Not within this PR - let's keep things simple. I'll address your comments on the original push in a separate PR.

amoss:
> @thiagoftsm some comments somehow appeared in this PR, but perhaps you had added them on another one. I had done a screw up with the branches, remotes and commits in the various repos, this one is clean, so please add a review here.

I think something has gone a little wrong - @thiagoftsm please see Slack. These changes don't need to be made in @cakrit's PR - although it seems possible that you commented on an earlier commit and the github UI is showing them in the wrong place.

Issue #7385, opened 2019-11-29
LGTM alerts & blog post on GitHub security lab


cakrit:
We want to do a blog post on https://securitylab.github.com/research. We've discussed with GitHub and agreed that a good way to appear there is the following: We currently have 7 alerts on https://lgtm.com/projects/g/netdata/netdata/?mode=list that we would need to handle anyway.

We move to resolve them, writing in detail the steps that we followed. If there are any false positives that we mark as irrelevant in the process, we explain why they were false positives.

So we do what we would normally do with these alarms anyway, but we make it a point of recording the investigation and actions, so that we can end up with a draft post that @jlbriston will deliver to the guys at GitHub, for review, corrections, publication.

Let's please try to prioritize this, as security alerts are top priority, even if they turn out to be false positives.

jlbriston:
@cakrit fyi I made an issue earlier on Marketing here also, I'm fine with keeping 1 for tech and 1 for marketing for this - https://github.com/netdata/marketing/issues/35

vlvkobal:
@jacekkolasa five issues left and all of them are related to JavaScript.

cakrit:
All alerts are gone now, which is awesome. If detailed comments haven't been sent already to @jlbriston , please comment here on the exact process you followed to address each one of the alerts.

jacekkolasa:
@jlbriston 
Well, the fixes are made weren't unfortunately "spectacular bug hunting", but i'll try:

One major error which LGTM was pointing out was [js/client-side-unvalidated-url-redirection](https://help.semmle.com/wiki/display/JS/Client-side+URL+redirect). In part of the app we are redirecting to new url, based on the message from our iFrame. The callback looks very similar to callback from DOM event - that's why LGTM was signalising error. In this case, it was false-positive - we are in control of that iFrame, and of that callback. But the error made us more aware of potential problems like this in the future.
That is a vey nice feature of LGTM, something that linters working locally (like `eslint`) will not detect.

The rest of javascript alerts were just recommendations on code style. What was good about it was that it "magically" guesses our settings based on surrounding code, and there was no need to explicitly state our code-style settings.

jlbriston:
@jacekkolasa thank you for this! Can you explain more on what potential errors could have occurred if there wasn't a fix? 

For the javascript alerts, could there have been any potential issues if the code style wasn't changed? Did you accept their suggestion and update the code style?

jacekkolasa:
(responded in the meeting)

PR #7441, opened 2019-12-05
address lgtm alerts

jacekkolasa:
##### Summary

Fixes javascript lgtm warning from #7385

1 error silenced: https://lgtm.com/rules/1979240340/
because it's a false-positive - LGTM thinks that redirect is caused by user input (sees the `event` property), but actually it only comes from iframe.

The rest is just lint (semicolons added) and unused code removal

##### Component Name

web

### DIFF
--- a/.remarkrc.js
+++ b/.remarkrc.js
@@ -1,8 +1,5 @@
 // Source: https://github.com/codacy/codacy-remark-lint/raw/master/.remarkrc.js
 
-const fs = require("fs");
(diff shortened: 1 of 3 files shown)


jacekkolasa:
lgtm doesn't show some fixes but perhaps that's a bug in lgtm, i think it will be marked fixed when we merge to master. I was looking in LGTM page at those changed files and i don't see there any alerts anymore

Issue #7410, opened 2019-12-03
Load labels from a config file


amoss:
- [x] Add functionality to the config file routines to reload a single section.
- [x] Add a `[host labels]` section to `netdata.conf`.
- [x] Add functionality to build a list of key/value pairs in a fresh list (using the struct from #7408) but not store it in the RRDHOST.
- [x] A separate procedure that acquires the locks, replaces the RRDHOST list of labels, releases the locks and then cleans up the old list.
- [x] Label keys must be `[a-zA-Z0-9_.-]+`.
- [x] Label values can be UTF-8 strings.
- [x] Must skip over `from environment` and `from kubernetes pods` as key values.
- [x] Load the labels from the config during agent startup.

stelfrag:
Good job! that describes the building blocks we need to work on, then we just need
to assemble them together and verify that we have the "business" logic in place.

Comments:

- What is the `from environment` directive -- you mean if there `NETDATA_LABEL_<name> =<value>` in the environment?
- `from kubernetes` should make a note, somehow, to let the cgroup-name.sh script that it needs to fetch / parse and store the kubernetes labels (by calling the new function we will build)
   - Determine if the check for kubernetes can be done as soon as possible (The #7416 arrives just in time to further tune this) 

To sum it  up. We assume that in the labels section we have:

1. User defined labels
1. from kubernetes directive (or other container)
1. More user defined labels what may override things (especially from part 2)

We will:
- Apply the labels ASAP from section 1
- Wait for container labels to get fetched  / parsed and added
  - Need to check what it would mean to do this synchronously
    - When the config file is loaded, not everything is initialized
    - If it can be done, what impact will that have to the agent startup time
  - If not synchronously, need to examine what will happen during the time that the agent config is laoded but container labels are still missing (because we wait for step 2 to complete before we move on to step 3)
    - Metrics and alarms for example will be in place and may start using labels before we complete steps 2 and 3
- Apply labels from 3

amoss:
The `from environment =` directive was to support labels in a docker container. We didn't specifiy in the technical refinement meeting what those variables would be, or how that interface would work so the comment about skipping over the directive is really a placeholder for a future task. This could be `NETDATA_LABEL_` as you describe if this matches how we anticipate the docker interface would work.

There would be no container labels in this task at all. Your summary about the startup time and the need to delay everything until the labels are populated is entirely correct - but it would happen after #7411 during the #7412 integration task.

PR #7451, opened 2019-12-06
Labels from netdata.conf

thiagoftsm:
##### Summary
Implements #7410 

This PR brings the capability to work with labels inside netdata.conf, it also will bring functions necessaries for other parts of label epic.
##### Component Name
Database
Daemon

### DIFF
--- a/database/rrd.h
+++ b/database/rrd.h
@@ -164,6 +164,7 @@ struct label {
 };
 
 struct label *create_label(char *key, char *value, LABEL_SOURCE label_source);
+extern int is_valid_label(char *key);
 
 // ----------------------------------------------------------------------------
 // RRD DIMENSION - this is a metric
--- a/database/rrd.h
+++ b/database/rrd.h
@@ -164,7 +164,6 @@ struct label {
 };
 
 struct label *create_label(char *key, char *value, LABEL_SOURCE label_source);
-extern int is_valid_label(char *key);
 
 // ----------------------------------------------------------------------------
 // RRD DIMENSION - this is a metric
--- a/database/rrd.h
+++ b/database/rrd.h
@@ -150,10 +150,10 @@ typedef enum rrddim_flags {
 
 typedef enum label_source {
     LABEL_SOURCE_AUTO             = 0,
-    LABEL_SOURCE_NETDATA_CONF     = (1 << 0),
-    LABEL_SOURCE_DOCKER           = (1 << 1),
-    LABEL_SOURCE_ENVIRONMENT      = (1 << 2),
-    LABEL_SOURCE_KUBERNETES       = (1 << 3)
+    LABEL_SOURCE_NETDATA_CONF     = 1,
+    LABEL_SOURCE_DOCKER           = 2,
+    LABEL_SOURCE_ENVIRONMENT      = 3,
+    LABEL_SOURCE_KUBERNETES       = 4
 } LABEL_SOURCE;
 
 struct label {
@@ -163,7 +163,10 @@ struct label {
     struct label *next;
 };
 
+char *translate_label_source(LABEL_SOURCE l);
 struct label *create_label(char *key, char *value, LABEL_SOURCE label_source);
+struct label *add_label_to_list(struct label *l, char *key, char *value, LABEL_SOURCE label_source);
+void reload_host_labels();
 
 // ----------------------------------------------------------------------------
 // RRD DIMENSION - this is a metric
--- a/database/rrdhost.c
+++ b/database/rrdhost.c
@@ -707,6 +707,16 @@ void rrdhost_save_charts(RRDHOST *host) {
     rrdhost_unlock(host);
 }
 
+int is_valid_label(char *key) {
+    while(key) {
+        if(!isascii(*key) || !isdigit(*key) || !isalpha(*key) || *key != '.' || *key != '_' || *key != '-')
+            return 0;
+
+        key++
+    }
+    return 1;
+}
+
 struct label *load_auto_labels()
 {
     return NULL;
@@ -714,7 +724,9 @@ struct label *load_auto_labels()
 
 struct label *load_config_labels()
 {
-    return NULL;
+    struct label *ret = NULL;
+//add_label_to_list
+    return ret;
 }
 
 struct label *load_kubernetes_labels()
--- a/database/rrdhost.c
+++ b/database/rrdhost.c
@@ -707,7 +707,7 @@ void rrdhost_save_charts(RRDHOST *host) {
     rrdhost_unlock(host);
 }
 
-static int is_valid_label(char *key) {
+static int is_valid_label_key(char *key) {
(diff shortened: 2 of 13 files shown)


database/rrd.h
```diff
@@ -164,6 +164,7 @@ struct label {
 };
 
 struct label *create_label(char *key, char *value, LABEL_SOURCE label_source);
+extern int is_valid_label(char *key);
```
cakrit: we'll only use this in rrdhost.c, I don't think you need an extern here

database/rrdhost.c
```diff
@@ -707,14 +707,27 @@ void rrdhost_save_charts(RRDHOST *host) {
     rrdhost_unlock(host);
 }
 
+int is_valid_label(char *key) {
```
cakrit: It would be nice to get this function in the branch soon, so I can call it too.

database/rrdhost.c
```diff
@@ -707,14 +707,27 @@ void rrdhost_save_charts(RRDHOST *host) {
...
+    while(key) {
+        if(!isascii(*key) || !isdigit(*key) || !isalpha(*key) || *key != '.' || *key != '_' || *key != '-')
```
ilyam8: >  Label keys must be [a-zA-Z0-9_.-]+.

no ascii there

database/rrd.h
thiagoftsm: All right, I was in doubt about the possibility to call it out, thank you! I removed it from rrd.h and I used static in the file that it is declared.

database/rrdhost.c
thiagoftsm: I changed everything requested, for me we can merge this PR and I create another to finalize #7410, I was also blocked before you merge a PR today.

database/rrdhost.c
thiagoftsm: All these values are ascii according the function definition `checks whether c is a 7-bit unsigned char value that fits into the ASCII character set.`, but I agree that it gives the possibility to have other values, so I removed the function.

Thanks!

health/health.c
```diff
@@ -139,6 +139,9 @@ void health_reload_host(RRDHOST *host) {
             t->flags |= HEALTH_ENTRY_FLAG_UPDATED;
     }
 
+    //Reload the section "host labels" from netdata.conf, because it can be new labels that will indicate alarms.
+    config_reload_section(NULL, "host labels");
```
stelfrag: Should probably discuss this; to remove spaces from the section

thiagoftsm: It is now the unique section with space, I agree that this is a break of pattern and should be discussed.

database/rrdhost.c
```diff
@@ -707,14 +707,31 @@ void rrdhost_save_charts(RRDHOST *host) {
     rrdhost_unlock(host);
 }
 
+static int is_valid_label(char *key) {
+    while(key) {
+        if(!isdigit(*key) || !isalpha(*key) || *key != '.' || *key != '_' || *key != '-')
```
ilyam8: Function name is `is_valid_label`, do we need to check label value. Or valid key == valid label. Or this is `is_valid_label_key`?

thiagoftsm: No it is only a check for the key, I think it is better to allow the users define the labels that they prefer, for keys would be more complicated, this is the motive we are defining this function.

thiagoftsm: I accepted your suggestion and I create a new commit to rename the function.

ilyam8: I see label value has req

#7410

> Label values can be UTF-8 strings.

ilyam8: ok, fixed 😄

database/rrdhost.c
```diff
@@ -707,14 +707,31 @@ void rrdhost_save_charts(RRDHOST *host) {
     rrdhost_unlock(host);
 }
 
+static int is_valid_label_key(char *key) {
+    while(key) {
+        if(!isdigit(*key) || !isalpha(*key) || *key != '.' || *key != '_' || *key != '-')
```
ilyam8: the condition is wrong, have you tested it?

ilyam8: > while(key) = > while (*key != '\0')

ilyam8: And check for `NULL` is missing

libnetdata/config/appconfig.c
```diff
@@ -472,6 +477,23 @@ int appconfig_load(struct config *root, char *filename, int overwrite_used)
...
+                        error("INTERNAL ERROR: Cannot remove '%s' from  section '%s', it was not inserted before.",
+                               cv2->name, co->name);
+
+                    free(cv2);
+                    cv2 = save;
+                }
+                co->values = NULL;
+                config_section_unlock(co);
+            }
+
```
amoss: Github UI won't let me attach comment to L505-L516:

This just needs one part from the code you wrote in the separate function: if we are reloading a single section, skip overwriting any fields outside that section. i.e.

```
[general]
    ... ignore any changes don't write anything....
[host labels]
   ... write all the stuff in here...
[plugins]
   ... ignore everything, don't update any changes...
```

health/health.c
```diff
@@ -140,7 +140,7 @@ void health_reload_host(RRDHOST *host) {
     }
 
     //Reload the section "host labels" from netdata.conf, because it can be new labels that will indicate alarms.
-    config_reload_section(NULL, CONFIG_SECTION_HOST_LABEL);
+    config_load(NULL, 1, CONFIG_SECTION_HOST_LABEL);
```
amoss: I don't think this call should be here?

thiagoftsm: Case we reload all alarms, they will need to have all the labels updated before we apply them. I decided to set it here to avoid two calls to `netdatacli`, the first would be to reload the labels and after this we would need another to call the alarm reload. I agree  that your comment is more atomic, but should we keep this to simplify users life?

database/rrdhost.c
amoss: It should be:
```
if(!(*key == '.' || *key == '_' || *key == '-' || isdigit(*key) || isalpha(*key) ))
```
Please test it.

health/health.c
amoss: Updating the label config is an explicit operation by the user, it should not be a side-effect of reloading the alarms - it makes it harder for the user to understand why something has happened.

libnetdata/config/appconfig.c
thiagoftsm: You are right, I forgot to bring this from the function I removed, I brought it now. Thank you!

health/health.c
thiagoftsm: All right, I removed it from health in the last commit.

database/rrdhost.c
thiagoftsm: I changed for your version, it is more readable, and I also adjust the loop end.

PR #7458, opened 2019-12-06
Labels fill list from netdata configuration

thiagoftsm:
##### Summary
Finish #7410 

This PR brings the last bullets that were missed o the issue 7410.
##### Component Name
Daemon
Database

### DIFF
--- a/database/rrdhost.c
+++ b/database/rrdhost.c
@@ -708,6 +708,14 @@ void rrdhost_save_charts(RRDHOST *host) {
 }
 
 static int is_valid_label_key(char *key) {
+    //Prometheus exporter
+    if(!strcmp(key, "chart") || !strcmp(key, "family")  || !strcmp(key, "dimension"))
+        return 0;
+
+    //Internal dimension
+    if (*key == '_')
+        return 0;
+
     while(*key) {
         if(!(isdigit(*key) || isalpha(*key) || *key == '.' || *key == '_' || *key == '-'))
             return 0;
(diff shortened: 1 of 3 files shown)


database/rrdhost.c
```diff
@@ -712,10 +712,16 @@ static int is_valid_label_key(char *key) {
...
+    //Netdata internal
     if (*key == '_')
         return 0;
 
+    key++;
+    //Prometheus internal
+    if (*key == '_')
+        return 0;
+
+    key++;
```
ilyam8: i think there is no reason to check for `__` if `_` is already wrong (return 0). And i think you dont need `key++` after the check.

This is perfect case to use unit tests btw

thiagoftsm: I removed with the last commit, you right, I had a test that did not reach the loop when there was one unique `_` , so the second is not necessary.
I agree this is a perfect place for unique test.

database/rrdhost.c
```diff
@@ -712,12 +712,7 @@ static int is_valid_label_key(char *key) {
...
 
-    //Netdata internal
-    if (*key == '_')
-        return 0;
-
-    key++;
-    //Prometheus internal
+    //Netdata and Prometheus  internal
     if (*key == '_')
         return 0;
```
ilyam8: `key++;` after is not needed too (befre `while` loop)

thiagoftsm:
> The new labels are not loaded from the config file when the "reload-labels" command is used from the clitool.

I missed a call of a function, I removed the call from health_reload, but I did not put inside `load_config_labels()`, now it is ok.

Thank you!

PR #7462, opened 2019-12-07
Labels to alarms

### LINKED ISSUE #7369: As a user, I want to be able to use custom host labels in alarm configurations, so that I can group a collection of resources that are logically related
- [x] Add new parameter to alarms and alarm templates that allows a user to apply the alarm only to charts of hosts with the specific labels. The parameter should accept [netdata simple patterns](https://docs.netdata.cloud/libnetdata/simple_pattern/) 
- [ ] See labels as available variables in `/api/v1/alarm_variables`

thiagoftsm:
##### Summary
Closes #7369 

This PR brings the labels feature to alarms.
##### Component Name
Health
##### Additional Information
While I was developing, I used the following alarm

```
 alarm: dev_dim_template_user
    on: system.cpu
    os: linux
lookup: sum -3s at 0 every 3 percentage of user
 units: %
 every: 1s
  warn: $this > 1
  crit: $this > 4
 label: label1 = 123 
```

and I changed `netdata.conf` to 

```
[host labels]
        label1 = 123
        label2 =456
        label3= 789
```

I executed the following tests:

1 - I run netdata to confirm I had alarms.
2 - change `label1 = 123 ` to invalid value `label1 = 123456`, I cannot have alarms
3 - Use only the label name  inside `label:`, for example, `label1` I will have alarms again.
4 - Use simple patterns with `label:`, for example `label*` instead `label1`
5 - Change the previous lookup line for one with `foreach`:
`lookup: sum -3s at 0 every 3 percentage foreach system,user,nice`
I must have alarms again
6 - Change `alarm:` to `template:` I also need to have alarms.

### DIFF
--- a/database/rrdcalc.c
+++ b/database/rrdcalc.c
@@ -605,10 +605,9 @@ void rrdcalc_unlink_and_free(RRDHOST *host, RRDCALC *rc) {
     rrdcalc_free(rc);
 }
 
-void rrdcalc_labels_unlink_alarm_from_host(RRDHOST *host) {
-    netdata_rwlock_rdlock(&host->labels_rwlock);
+static void rrdcalc_labels_unlink_alarm_loop(RRDHOST *host, RRDCALC *alarms) {
     RRDCALC *rc, *clean = NULL;
-    for (rc = host->alarms; rc; rc = rc->next) {
+    for (rc = alarms; rc; rc = rc->next) {
         if (!rc->labels) {
             continue;
         }
@@ -651,6 +650,15 @@ void rrdcalc_labels_unlink_alarm_from_host(RRDHOST *host) {
               clean->labels);
         rrdcalc_unlink_and_free(host, clean);
     }
+
+
+}
+
+void rrdcalc_labels_unlink_alarm_from_host(RRDHOST *host) {
+    netdata_rwlock_rdlock(&host->labels_rwlock);
+
+    rrdcalc_labels_unlink_alarm_loop(host, host->alarms);
+    rrdcalc_labels_unlink_alarm_loop(host, host->alarms_with_foreach);
 
     netdata_rwlock_unlock(&host->labels_rwlock);
 }
--- a/database/rrdcalc.c
+++ b/database/rrdcalc.c
@@ -615,7 +615,7 @@ void rrdcalc_labels_unlink() {
 
         if (host->labels) {
             rrdhost_wrlock(host);
-            netdata_rwlock_wrlock(&host->labels_rwlock);
+            netdata_rwlock_rdlock(&host->labels_rwlock);
 
             RRDCALC *rc, *clean = NULL;
             for (rc = host->alarms; rc; rc = rc->next) {
--- a/database/rrdcalc.c
+++ b/database/rrdcalc.c
@@ -559,6 +559,8 @@ void rrdcalc_free(RRDCALC *rc) {
     freez(rc->units);
     freez(rc->info);
     simple_pattern_free(rc->spdim);
+    freez(rc->labels);
+    simple_pattern_free(rc->splabels);
     freez(rc);
 }
 
--- a/database/rrdcalc.c
+++ b/database/rrdcalc.c
@@ -605,6 +605,56 @@ void rrdcalc_unlink_and_free(RRDHOST *host, RRDCALC *rc) {
     rrdcalc_free(rc);
 }
 
+void rrdcalc_labels_unlink() {
+    rrd_rdlock();
+
+    RRDHOST *host;
+    rrdhost_foreach_read(host) {
+        if (unlikely(!host->health_enabled))
+            continue;
+
+        if (host->labels) {
+            rrdhost_wrlock(host);
+            netdata_rwlock_wrlock(&host->labels_rwlock);
+
+            RRDCALC *rc, *clean = NULL;
+            for (rc = host->alarms; rc; rc = rc->next) {
+                if (clean) {
+                    error("Health configuration for alarm '%s' cannot be applied, because the host %s does not have the label(s) '%s'",
+                          clean->name,
+                          host->hostname,
+                          clean->labels);
+                    rrdcalc_unlink_and_free(host, clean);
+                    clean = NULL;
+                }
+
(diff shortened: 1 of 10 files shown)


database/rrdcalctemplate.c
```diff
@@ -4,6 +4,32 @@
 #include "rrd.h"
 
 // ----------------------------------------------------------------------------
+
+static int rrdcalctemplate_has_label(RRDCALCTEMPLATE *rt,  RRDHOST *host) {
+    errno = 0;
+    struct label *move = host->labels;
+    if(move && rt->labels) {
+        netdata_rwlock_wrlock(&host->labels_rwlock);
```
amoss: This should be the reader lock.

database/rrdcalc.c
```diff
@@ -603,6 +605,56 @@ void rrdcalc_unlink_and_free(RRDHOST *host, RRDCALC *rc) {
...
+    rrd_rdlock();
+
+    RRDHOST *host;
+    rrdhost_foreach_read(host) {
+        if (unlikely(!host->health_enabled))
+            continue;
+
+        if (host->labels) {
+            rrdhost_wrlock(host);
+            netdata_rwlock_wrlock(&host->labels_rwlock);
```
amoss: Why is this the writer lock?

health/health.c
```diff
@@ -577,6 +577,8 @@ void *health_main(void *ptr) {
     time_t now                = now_realtime_sec();
     time_t hibernation_delay  = config_get_number(CONFIG_SECTION_HEALTH, "postpone alarms during hibernation for seconds", 60);
 
+    rrdcalc_labels_unlink();
```
amoss: If I've read the `rrdcalc_labels_unlink` code correctly then it scans the current labels to decide which alarms to filter out. This can change if the user reloads the labels while the agent is running.

database/rrdcalctemplate.c
thiagoftsm: I agree 100% with this and the necessary change to the next lock, I will write an answer here that is also applied for all other comments that you did, this is a [WIP] yet, so sometimes I only copy and paste the code to test whether the main goal is reached, but before to remove the [WIP] I do a revision of everything.
Thank you very much to save me time, I already brought all the changes that you requested, you save me time! :)

health/health.c
thiagoftsm: In the previous commits I had not arrive in the reload yet, but you are right, we must call it inside health_reload(). I brought this in the last commit.

thiagoftsm:
Reviewers, please, remember that to test this PR it will be necessary to test the following alarms:

- Normal alarm
- Template alarm
- Normal alarm with `foreach`
- Template alarm with `foreach`

Best regards!

thiagoftsm:
> I can't test this - as an end-user I need to understand what config file I change and how I change it to filter alarms and alarm templates by a host label. Please add this to the PR so that I can test it before we merge it. I did hunt back through the issues to find the discussion about how this will look but I cannot find it - and it needs to be here to be testable.

I wrote the missing documentation and I brought it in the last commit.

health/REFERENCE.md
```diff
@@ -370,6 +371,22 @@ increasing. Eventually, the comparison will find the averages in the two time-fr
...
+
+```yaml
+label: _is_master = false
+```
+ 
+This option accepts [Netdata simple patterns](../libnetdata/simple_pattern/), so, case you wanna apply an alarm to either
+a master or a slave, you can write:
+
+```yaml
+label: _is_master = *
```
ilyam8: how about AND and OR ?

ilyam8: several labels, like

(_is_master = false AND _is_slave = false) OR (_is_master = true AND _is_slave = true)

thiagoftsm: It is completely possible, we are using simple pattern, so the users are able to do any valid combination inside label.

thiagoftsm: I  am waiting the final definition about this internal label to change the documentation.

ilyam8: Not it is not possible. I mean AND/OR between labels.

ilyam8: What is separator btw? How do i separate one label from other?

That is example, again, i want to apply alarm to:
 - master and salve servers
OR
 - standalone servers

i need something like this:

`label: (_is_master = "true" AND _is_slave = "true") OR (_is_master = "false" AND _is_slave = "false")`

thiagoftsm: All right, I set inside my netdata.conf the following labels:

```
[host labels]
        label1 = 123
        label2 =456
        label3= 789
        label4 = 10
```

And I tested with success the following lines:

```
label: label1=123 OR label25=456
label: label1=123 AND label2=456
```

About `_is_slave`, it is not set in any place, we only have `_is_master` merged until now, probably, this is the problem in your side.

ilyam8: No, it seems i wasnt clear. I havent tested. I read your documentation. It wasnt clear that OR/AND syntax is working.

So:
 - you can add any number of `label:`? i didnt see it in the docs
 - if i use simple pattern (expression with spaces) why symbols should i use to make it explicit? `"`?

```
label: label1="a* b* !c* *" AND label2="a* b* !c* *"
```

what is logical operator between `label:`? OR?

ilyam8: @thiagoftsm so it is possible to use `AND/OR` and `()` to group things inside `label:`?

If so this should be in the docs.

thiagoftsm: I got now, I will extend the documentation given better explanation, thanks!

thiagoftsm: I extended the documentation as you requested @ilyam8 , let me know whether I am clear now on it.

ilyam8: > if i use simple pattern (expression with spaces) what symbols should i use to make it explicit? "?

> label: label1="a* b* !c* *" AND label2="a* b* !c* *"

like that?

thiagoftsm:
> The only thing missing is an explicit list of tests, let's do this for the PR that merges the branch back into master today.

I updated the initial comment with the list of tests I did.

PR #7485, opened 2019-12-09
Create internal netdata labels

### LINKED ISSUE #7400: Enabler: Internal netdata labels
- [ ] The following will be default internal labels (not exported to third party systems, unless explicitly requested by users):

|Description|api/v1/info|Label Name|
|---|---|---|
|OS name|os_name|_os_name|
|OS version| os_version|_os_version|
|Kernel version|kernel_version|_kernel_version|
|Architecture|architecture|_architecture|
|Virtualization|virtualization|_virtualization|
|Container Tech|container|_container|

- [ ] For streaming, we should also have default labels specifying if a host is a master  (has metrics streamed to it) and/or a slave (streams metrics to another host).
    - _is_master=[true/false]
    - _is_slave=[true/false] AND/OR _streams_to=[master host, as defined in stream.conf]

vlvkobal:
##### Summary
Closes #7400

##### Component Name
daemon


thiagoftsm:
A suggestion before I approve it, everything works as expected, but you did not remove the comments on top of the function `load_auto_labels`, please, can you remove them?

database/rrdhost.c
```diff
@@ -401,6 +401,33 @@ RRDHOST *rrdhost_find_or_create(
...
+        if (!is_master_label_updated) {
+            struct label *prev_label = NULL;
+            struct label *label = localhost->labels;
+
+            uint32_t is_master_hash = simple_hash("_is_master");
+
+            netdata_rwlock_wrlock(&localhost->labels_rwlock);
+            while (label) {
+                if (label->key_hash == is_master_hash && !strcmp(label->key, "_is_master")) {
+                    struct label *new_label = create_label("_is_master", "true", LABEL_SOURCE_AUTO);
```
amoss: There is no need to create a new node in the list - just overwrite the value.

database/rrdhost.c
```diff
@@ -401,6 +401,33 @@ RRDHOST *rrdhost_find_or_create(
                 , system_info
                 , 0
         );
+
```
amoss: This procedure is only called from one place (ignoring the unit-test entry points) - inside `rrdpush_receive()`, which is only called from `rrdpush_receiver_thread()`.

The way it is implemented in this PR, receiving a STREAM will cause the flag to flip to true. But if the stream is closed then the flag will not flip back to false. There are two ways for how to implement this:
1. Statically decide if we are a master based on the loaded configuration (then we don't need this code).
2. Dynamically decide if we are master because there is an incoming stream (then we need something more complex than this).

Option (2) is complicated in the presence of multiple incoming streams. @cakrit how do you want this implemented: fully dynamic view of whether we are receiving metrics, or a static view of if we are expecting metrics?

cakrit: We only discussed Option 1 for both slave and master. Making it dynamic may keep flipping the label, which is not very useful.

database/rrdhost.c
vlvkobal: To overwrite the value I need to change your code to
```
struct label *create_label(char *key, char *value)
{
    struct label *label = callocz(1, sizeof(struct label));
    if (label != NULL) {
        label->key = strdupz(key);
        label->value = strdupz(value);
    }
    return label;
}
```

database/rrdhost.c
vlvkobal: I didn't find an easy way to get information from the configuration directly. That's why I put the code into `rrdhost_find_or_create()`.

amoss: Then the solution is simplest - we delete this piece of code and put something in the startup sequence that determines from the config whether the node is a master or not.

database/rrdhost.c
amoss: No you do not - you just update the structure.

database/rrdhost.c
vlvkobal: What if the config is wrong and Netdata never receives data from slaves? Will we treat the node as a master?

database/rrdhost.c
vlvkobal: Ok. "false" has more characters than "true". But it's a dirty hack to suppose that the code will never be changed. What if at some point somebody decides that words should be, for example, "yes" and "no"?

database/rrdhost.c
ilyam8: It seems it is `is_configured_as_master` label

amoss: Yes - if the configuration is wrong and the node never receives data from slaves then we will still consider it to be a master. 

This is significantly easier than the alternative, which (as @cakrit points out above) will lead to a poor user experience as labels will flip states and so they will not be able to use them to filter nodes, which is the desired purpose of this feature.

To do the alternative properly we would need a counter of how many slaves are connected, rather than just a simple flag and it would be be a much more complex implementation for something that would look worse when the user tries to access the functionality by filtering the nodes.

Think of it as `is_configured_as_master` as @ilyam8 put it - but keep the specified name `_is_master`.

database/rrdhost.c
amoss: That is a fair point - ignore my comment about updating the structure directly.

daemon/get-kubernetes-labels.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
thiagoftsm: I think you need a rebase, this is a file that was already merged.

libnetdata/config/appconfig.h
```diff
@@ -98,6 +98,36 @@
 #define CONFIG_MAX_NAME 1024
 #define CONFIG_MAX_VALUE 2048
```
thiagoftsm: These are also changed that were merged days ago.

daemon/get-kubernetes-labels.sh
vlvkobal: I already deleted the file. Look at the last commits.

libnetdata/config/appconfig.h
vlvkobal: The same. Conflicts were resolved.

thiagoftsm: Github continued showing me the files, I pressed "ctrl+shift+R" and now everything is ok.

streaming/rrdpush.c
```diff
@@ -151,6 +151,25 @@ static inline int should_send_chart_matching(RRDSET *st) {
...
 
+int configured_as_master() {
+    struct section *section = NULL;
+    int is_master = 0;
+
+    appconfig_wrlock(&stream_config);
+    for (section = stream_config.sections; section; section = section->next) {
+        char buf[GUID_LEN + 1];
+
+        if (regenerate_guid(section->name, buf) != -1 &&
```
amoss: It would be clearer to call `uuid_parse()` directly.

vlvkobal: I've changed it.

database/rrdhost.c
vlvkobal: It has been rewritten. Now the configuration file is used to check if a node was configured as a master.

PR #7489, opened 2019-12-09
The step-by-step Netdata tutorial

joelhans:
##### Summary

Here's the PR for my massive beginner-centric tutorial on configuring and using Netdata. This also includes some fresh styling for the homepage to accommodate the new material!

Not sure how the review/approval process will work on this one. There is _quite a lot_ of writing here.

Thanks much to @jlbriston and @thiagoftsm for their hard work in reviewing the initial versions!

##### Component Name

docs


docs/step-by-step/step-00.md
```diff
@@ -0,0 +1,94 @@
...
+completely new to Netdata, or have never tried health monitoring/performance troubleshooting systems before, this
+tutorial is perfect for you.
+
+If you have monitoring experience, or would rather get straight into configuring Netdata to your needs, you can jump
+straight into code and configurations with our [getting started guide](../getting-started.md).
+
+> This tutorial contains instructions for Netdata installed on a Linux system. Many of the instructions will work on
+> other supported operating systems, like FreeBSD and MacOS, but we can't make any guarantees.
+
+## Before we get started
```
andrewm4894: It maybe could be useful somewhere in this section to let users know how and where to reach out to ask questions or raise issues if they run into them. I'm almost thinking in terms of newbie type questions people might be too afraid to raise as issues in the repo. Not really sure how we want them to engage but just thinking somewhere around here could be a good chance to communicate to them the ways they can ask questions, get help, and or otherwise get involved in the community.

docs/step-by-step/step-01.md
```diff
@@ -0,0 +1,139 @@
...
+-   Human-readable units
+-   Metrics that are structured into charts, families, and contexts
+-   Automatically generated dashboards
+-   Charts designed for visual anomaly detection
+-   Hundreds of pre-configured alarms
+
+By standardizing your monitoring infrastructure, Netdata tries to make at least one part of your administrative tasks
+easy!
+
+## How Netdata has complementary systems, not competitors
```
andrewm4894: I wonder if somewhere in here or above it could be nice to link to this post: https://blog.netdata.cloud/posts/redefining-monitoring-netdata/

docs/step-by-step/step-01.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Per-second data collection
+
+Our first principle is per-second data collection for all metrics.
+
+That matters because you can't monitor a 2-second service-level agreement (SLA) with 10-second metrics. You can't detect
+quick anomalies if your metrics don't show them.
+
+How do we solve this? By decentralizing monitoring. Each node collects metrics, triggers alarms, and builds dashboards
+locally. That means you can scale Netdata to any size of infrastructure and keep the power of per-second metrics.
```
Ferroin: I would add to this that we try to focus on making the collection process as efficient as possible. The distributed nature of netdata is a major factor in making per-second resolution possible for large networks, but it's really the honestly amazing efficiency of data collection in the core that makes it possible on a per-system basis.

docs/step-by-step/step-02.md
```diff
@@ -0,0 +1,204 @@
...
+
+![Animated GIF of using the menus and
+submenus](https://user-images.githubusercontent.com/1153921/63464031-3ee30980-c413-11e9-886a-44594f60e0a9.gif)
+
+Netdata shows all its charts on a single page, so you can also scroll up and down using the mouse wheel, your
+touchscreen/touchpad, or the scrollbar.
+
+Both menus and the items displayed beneath them, called **submenus**, are populated automatically by Netdata based on
+what it's collecting. If you run Netdata on many different systems using different OS types or versions, the
+menus and submenus may look a little different for each one.
```
Ferroin: Might make sense to add a comment here mentioning that if you go to the dashboard right after starting Netdata, it may not show all the menus/submenus yet (some collectors take a while to initialize and send in data for the first time, and charts for them won't show up on the dashboard until the first data for them comes in).

Not exactly critical or anything, but I've seen a fair number of people get caught off-guard by this behavior.

docs/step-by-step/step-02.md
```diff
@@ -0,0 +1,204 @@
...
+charts](https://user-images.githubusercontent.com/1153921/65066637-9785c380-d939-11e9-8e26-6933ce78c172.gif)
+
+### Show and hide dimensions
+
+Each dimension can be hidden by clicking on it. Hiding dimensions simplifies the chart and can help you better discover
+exactly which aspect of your system is behaving strangely.
+
+### Resize charts
+
+Additionally, resize charts by clicking-and-dragging the icon on the bottom-right corner of any chart. To restore the
```
Ferroin: Something appears to have been cut off here.

docs/step-by-step/step-02.md
ilyam8: And ofc amazing which dimension to show `Not Zero` default feature!

docs/step-by-step/step-05.md
```diff
@@ -0,0 +1,334 @@
...
+
+We'll put the next two lines—`warn` and `crit`—together. In these lines, you declare at which percentage you want to
+trigger a warning or critical alarm. Notice the variable `$this`, which is the value calculated by the `lookup` line.
+These lines will trigger a warning if that average RAM useag average goes above 80%, and a critical alert if it's above
+90%.
+
+```yaml
+  warn: $this > 80
+  crit: $this > 90
+```
```
Ferroin: I would suggest adding something here at least mentioning the need for hysteresis to avoid alarm flapping. It's used by a vast majority of the stock alarms and the symptoms of not having it can be confusing to people who don't already have the background in mathematics or signal processing to know about it.

docs/step-by-step/step-06.md
```diff
@@ -0,0 +1,113 @@
...
+[Python](../../collectors/python.d.plugin/), although some use [Bash](../../collectors/charts.d.plugin/) (`charts.d`) or
+[Node.js](../../collectors/node.d.plugin/) (`node.d`).
+
+## Enable and disable plugins
+
+You don't need to explicitly enable plugins to auto-detect properly configured sources, but it's useful to know how to
+enable or disable them.
+
+One reason you might want to _disable_ plugins is to improve Netdata's performance on low-resource systems, like
+ephemeral nodes or edge devices.
```
Ferroin: Maybe mention that disabling orchestrator plugins in particular can save significant amounts of resources?

docs/step-by-step/step-07.md
```diff
@@ -0,0 +1,108 @@
...
+modal](https://user-images.githubusercontent.com/12263278/64901503-ee696f80-d691-11e9-9678-8d0e2a162402.gif)
+
+Select the Netdata snapshot file to import. Once the file is loaded, the dashboard will update with critical information
+about the snapshot and the system from which it was taken. Click **import** to render it.
+
+Your Netdata dashboard will load data contained in the snapshot into charts. Because the snapshot only covers a certain
+period, it won't update with new metrics.
+
+An imported snapshot is also temporary. If you reload your browser tab, Netdata will remove the snapshot data and
+restore your real-time dashboard for your machine.
```
Ferroin: I think we should probably explicitly call out here that loading a snapshot from an untrusted source is risky. It's not immediately obvious to that this might be the case to people who don't understand in detail how snapshots wor

docs/step-by-step/step-10.md
```diff
@@ -0,0 +1,213 @@
...
+You're almost through! At this point, you should be pretty familiar with now Netdata works and how to configure it to
+your liking.
+
+In this step of the tutorial, we're going to add a proxy in front of Netdata. We're doing this for both improved
+performance and security, so we highly recommend following these steps. Doubly so if you installed Netdata on a
+publicly-accessible remote server.
+
+> ❗ If you installed Netdata on the machine you're currently using (aka on `localhost`), and have been accessing Netdata
+> at `http://localhost:19999`, you can skip this step of the tutorial, as you can't set up a proxy for a service that's
+> running locally.
```
Ferroin: This is a bit disingenuous, as it absolutely is possible to set up a proxy on the local system, it just makes essentially no sense unless you're also exposing Netdata externally.

I would suggest changing it from 'you can't set up a proxy for a service that's running locally' to 'there is no benefit to setting up a proxy for a service that's running locally' (though that's not entirely true either, as you can make Netdata use a UNIX socket and then use a proxy to password protect the dashboard even for local users).

docs/step-by-step/step-02.md
ilyam8: Was sarcastic actually. I find it very confusing 😄

Ferroin: That's kind of my point though. Both the behavior of hiding zero dimensions by default and the delays in populating charts right after startup are points of confusion for users, and explicitly mentioning those behaviors in the tutorial makes it less likely that users will be confused by them.

ilyam8: > Both the behavior of hiding zero dimensions by default and the delays in populating charts

There is one more about it. Some internal collectors ignore zero value metrics (they are not collected)

https://docs.netdata.cloud/collectors/cgroups.plugin/#charts-with-zero-metrics

😄

docs/step-by-step/step-06.md
joelhans: Done.

docs/step-by-step/step-01.md
joelhans: Great point. Added some new text on this.

docs/step-by-step/step-01.md
joelhans: Good idea! Done.

docs/step-by-step/step-00.md
joelhans: I wish we had a more user-friendly platform for people to ask questions, but GitHub is it for now. I am willing to sacrifice my own email account for people who have questions, though...

docs/step-by-step/step-07.md
joelhans: Could you elaborate a bit? Or are you talking about general security best practices, like not plugging in USB drives you find on the sidewalk?

docs/step-by-step/step-05.md
joelhans: I'll research hysteresis a bit more and maybe just add a small note about it. The last time I looked at the `crit: $this > (($status == $CRITICAL) ? (85) : (95))` I couldn't wrap my head around it in less than 2 minutes and so I moved on, but you're right that users should at least be introduced to the idea so they don't think "my" way is the best way to write entities forever.

Ferroin: > I'll research hysteresis a bit more and maybe just add a small note about it. The last time I looked at the `crit: $this > (($status == $CRITICAL) ? (85) : (95))` I couldn't wrap my head around it in less than 2 minutes and so I moved on, but you're right that users should at least be introduced to the idea so they don't think "my" way is the best way to write entities forever.

Let me know if you need any help. I can't guarantee that I can make it perfectly clear, but a different perspective may help with understanding.

docs/step-by-step/step-07.md
Ferroin: Last I knew, snapshots included some JS code that gets executed on load, and we kind of just trust that the code in question is not malicious.

Assuming I'm correct about my understanding of snapshots, that could get all the way to arbitrary code execution if the user has a sufficiently insecure browser.

Even aside from that though, we're parsing a very large amount of untrusted data when we load a snapshot, and I'm not sure that the relevant code has been properly audited.

docs/step-by-step/step-05.md
ilyam8: > crit: $this > (($status == $CRITICAL) ? (85) : (95)) I couldn't wrap my head around it in less than 2 minutes and so I moved on

Haha completely understand, this is opposite of user friendly 😄

Ferroin: > > crit: $this > (($status == $CRITICAL) ? (85) : (95)) I couldn't wrap my head around it in less than 2 minutes and so I moved on
> 
> Haha completely understand, this is opposite of user friendly

I mean, people who have a lot of background in C-like languages and a basic understanding of shell script/Perl shouldn't have any issues with the logic.

The problem is that other than the choice of symbology (and yes, I would much prefer a more verbose symbology), there really isn't a 'better' way to express hysteresis, you kind of need to show both trip points to make things unambiguous, and there's only so many ways you can do that without making the config parsing significantly more complex.

joelhans: I never realized there was already an explanation of hysteresis here: https://docs.netdata.cloud/health/reference/#special-use-of-the-conditional-operator. That's quite clear and helpful. So, I'll add a shorter version of this and a link to this section for them to learn more.

docs/step-by-step/step-02.md
joelhans: Got it.

docs/step-by-step/step-07.md
joelhans: Got it. Added a blockquote to address this without getting too specific.

docs/step-by-step/step-05.md
joelhans: ... and done!

PR #7490, opened 2019-12-09
[WIP] Agent cloud link 7413

### LINKED ISSUE #7413: Initiate the Agent-Cloud-Link
When the agent is in a claimed state:

* Periodic timer to check the status of the ACL.
* If the agent is claimed and the link is not operational set up the wss + MQTT transport layer.

This will require:
* Including the libmosquito library as a dependency, sync with the SRE team.
* Add a heartbeat.
* Add error-checking.
* Tuning the connection parameters (timeouts) to be added to config.

Demo against a standard broker (VerneMq?)

stelfrag:
Fixes #7413 
##### Summary

[WIP] Agent cloud link

- Setup main mqtt thread to connect to a broker using V5 of the MQTT protocol (TBD)

- Send alarms to topic "netdata/alarm"

- Add error checks to handle connection failures

- Add params in new section [mqtt] for
  Broker, port
  Maximum concurrent sent  / recev messages

- Dummy function to check claiming status

- Generic mqtt_send command to publish message to a base topic , sub topic
  It will end up in the form base_topic/sub_topic

- Add host/port in the connection failure error message

### DIFF
--- a/daemon/mqtt.c
+++ b/daemon/mqtt.c
@@ -1,0 +1,188 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include "mqtt.h"
+#include "common.h"
+
+struct mosquitto *mosq = NULL;
+
+
+int am_i_claimed()
+{
+    return 1;
+}
+
+// Thread cleanup
+static void mqtt_main_cleanup(void *ptr) {
+    struct netdata_static_thread *static_thread = (struct netdata_static_thread *)ptr;
+    static_thread->enabled = NETDATA_MAIN_THREAD_EXITING;
+
+    info("cleaning up...");
+
+    static_thread->enabled = NETDATA_MAIN_THREAD_EXITED;
+}
+
+
+/**
+ * MQTT Main
+ *
+ * This thread will simply call the internal libmosquitto loop to handle
+ * pending requests - both inbound and outbound
+ *
+ * @param ptr is a pointer to the netdata_static_thread structure.
+ *
+ * @return It always returns NULL
+ */
+void *mqtt_main(void *ptr) {
+
+    netdata_thread_cleanup_push(mqtt_main_cleanup, ptr);
+
+    time_t now                = now_realtime_sec();
+    time_t hibernation_delay  = config_get_number(CONFIG_SECTION_HEALTH, "postpone alarms during hibernation for seconds", 60);
+
+    //unsigned int loop = 0;
+    while(!netdata_exit) {
+        int rc;
+        //loop++;
+        //debug(D_HEALTH, "Health monitoring iteration no %u started", loop);
+
+        if (am_i_claimed() == 0) {
+            usleep(60000 * 1000);
+            continue;
+        }
+
+        if (unlikely(mosq == NULL)) {
+            info("Initializing MQTT connection");
+            mqtt_init();
+            continue;
+        }
+
+        // Call the loop to handle inbound and outbound messages  // Timeout after 60 seconds
+        rc = mosquitto_loop(mosq, 60 * 1000, 1);
+        if (unlikely(rc == MOSQ_ERR_INVAL)) {
+            error("Invalid MQTT connection; dying?");
+            break;
+        }
+
+        if (unlikely(rc !=MOSQ_ERR_SUCCESS )) {
+            error("MQTT loop error code %d (%s)", rc, mosquitto_strerror(rc));
+        }
+
+        //usleep(1000 * 1000);
+
+    } // forever
+    mqtt_shutdown();
+
+    netdata_thread_cleanup_pop(1);
+    return NULL;
+}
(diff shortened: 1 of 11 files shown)


daemon/mqtt.c
```diff
@@ -0,0 +1,401 @@
...
+
+    info("Detected libmosquitto library version %d, %d.%d.%d",libmosq_version, libmosq_major, libmosq_minor, libmosq_revision);
+
+    while(!netdata_exit) {
+        int rc;
+
+        // TODO: This may change when we have enough info from the claiming itself to avoid wasting 60 seconds
+        // TODO: Handle the unclaim command as well -- we may need to shutdown the connection
+        if (am_i_claimed() == 0) {
+            usleep(60000 * 1000);
```
thiagoftsm: Hi @stelfrag ,

Please change the call from `usleep` to `nanosleep`, because this function was declared obsolete according official documentation http://man7.org/linux/man-pages/man3/usleep.3.html.

daemon/mqtt.c
```diff
@@ -0,0 +1,401 @@
...
+    if (unlikely(rc != MOSQ_ERR_SUCCESS))
+        error("Failed to tune the thread model for libmoquitto (%s)",mosquitto_strerror(rc));
+
+    // attempt V5 protocol for now, no doc on the return status for now
+    rc = mosquitto_int_option(mosq, MQTT_PROTOCOL_V5, 0);
+    if (unlikely(rc != MOSQ_ERR_SUCCESS))
+        error("MQTT protocol specification rc = %d (%s)", rc, mosquitto_strerror(rc));
+
+
+    // These added for the V5 -- may need to be removed if we go to 3.x
```
thiagoftsm: Instead to remove, would not be better to comment the code?

daemon/mqtt.c
```diff
@@ -0,0 +1,401 @@
...
+    //mosquitto_tls_set(mosq, "/etc/netdata/mqtt/ca.crt", NULL, "/etc/netdata/mqtt/server.crt", "/etc/netdata/mqtt/server.key", NULL);
+
+    rc = mosquitto_connect_async(mosq, mqtt_broker_hostname, mqtt_broker_port, NETDATA_MQTT_PING_INTERVAL);
+    if (unlikely(rc != MOSQ_ERR_SUCCESS))
+        error("Connect %s MQTT status = %d (%s)", netdata_configured_hostname, rc, mosquitto_strerror(rc));
+    else
+        info("Establishing MQTT link to %s", netdata_configured_hostname);
+
+    // TODO: This may need to be elsewhere -- also check return code
+    mqtt_subscribe("netdata/command");
```
thiagoftsm: Only to confirm, because I did not study the library libmosquitto yet, I verified that you were reporting errors in different parts of this function, but you never returned from errors, so  I understood that the library sets default values every time that you moved in front, is my conclusion correct?

daemon/mqtt.c
stelfrag: I will do !! thank you @thiagoftsm

daemon/mqtt.c
stelfrag: We will most likely go with version 5 to take advantage of  new features. It can be set to be an option and select the appropriate version as needed in the future

daemon/mqtt.c
stelfrag: Several functions will fail becaus of incorrect arguments, others when the actual connection is attempted (e.g. you never get a connection established callback for example)

For the subscribe command more error handling will be added (also it may be moved around) depending on the agent cloud link protocol

daemon/mqtt.c
amoss: The protocol design is still changing but my current estimate is that we will need v5 to do the authentication properly - we can talk about this in the groom.

daemon/mqtt.c
```diff
@@ -0,0 +1,401 @@
...
+        final_topic = topic;
+    }
+
+    rc = mosquitto_pub_topic_check(final_topic);
+    if (unlikely(rc != MOSQ_ERR_SUCCESS))
+        return rc;
+
+    int msg_len = strlen(message);
+
+    // TODO: handle encoding validation -- the message should be UFT8 encoded by the sender
```
amoss: I think we can skip this - the protocol just defines a payload as an application specific stream of bytes.

stelfrag: yes, may not be needed..  just put a note there in case we need it

daemon/mqtt.c
stelfrag: Changed to sleep_usec as the rest of the code base in the agent

thiagoftsm: Thank you! :)

stelfrag:
Closing this -- will post a new issue to investigate how we will go about doing the MQTT over wss upgrade

PR #7493, opened 2019-12-10
Add support for host labels to API responses

### LINKED ISSUE #7096: Enabler: All API calls support labels, where applicable

vlvkobal:
##### Summary
Host labels have been added to
- [x] `api/v1/info`
- [x] `api/v1/charts`
- [x] `api/v1/chart`
- [x] `api/v1/data`
- [x] `api/v1/alarm_variables`
- [x] `api/v1/alarms`

responses in the following format:
```
"labels": {
	"app": "netdata",
	"controller-revision-hash": "64966d4b74",
	"release": "netdata",
	"role": "slave"
},
```
Documentation of the Netdata API for Swagger Editor has been updated.
All responses have been checked in a JSON validator.

Partly implements #7096

##### Component Name
web


web/api/formatters/json_wrapper.c
```diff
@@ -26,16 +26,29 @@ void rrdr_json_wrapper_begin(RRDR *r, BUFFER *wb, uint32_t format, RRDR_OPTIONS
...
                        "   %sid%s: %s%s%s,\n"
                        "   %sname%s: %s%s%s,\n"
-                       "   %sview_update_every%s: %d,\n"
+                   , kq, kq
+                   , kq, kq, sq, r->st->id, sq
+                   , kq, kq, sq, r->st->name, sq);
+
+    RRDHOST *host = r->st->rrdhost;
+    buffer_sprintf(wb, "   %slabels%s: {\n", kq, kq);
+    int count = 0;
```
thiagoftsm: A suggestion, I verified that the variable `count` is incremented during the loop and its main goal is to avoid to print a break line when we are in the first address of the loop, I think you can discard this variable and save the first address `host->labels`, with this you can compare the address values and print the break line only when the addresses are different.

web/api/formatters/json_wrapper.c
```diff
@@ -32,14 +32,20 @@ void rrdr_json_wrapper_begin(RRDR *r, BUFFER *wb, uint32_t format, RRDR_OPTIONS
...
     buffer_sprintf(wb, "   %slabels%s: {\n", kq, kq);
+
     int count = 0;
-    netdata_rwlock_rdlock(&host->labels_rwlock);;
+    netdata_rwlock_rdlock(&host->labels_rwlock);
     for (struct label *label = host->labels; label; label = label->next) {
         if(count > 0) buffer_strcat(wb, ",\n");
-        buffer_sprintf(wb, "      %s%s%s: %s%s%s", kq, label->key, kq, sq, label->value, sq);
+
+        char value[CONFIG_MAX_VALUE + 1];
```
amoss: If this is the target of the expansion it does not include space for the extra escaping characters.

web/api/web_api_v1.c
```diff
@@ -774,11 +774,15 @@ inline void host_labels2json(RRDHOST *host, BUFFER *wb, size_t indentation) {
...
     int count = 0;
-    netdata_rwlock_rdlock(&host->labels_rwlock);;
+    netdata_rwlock_rdlock(&host->labels_rwlock);
     for (struct label *label = host->labels; label; label = label->next) {
         if(count > 0) buffer_strcat(wb, ",\n");
         buffer_strcat(wb, tabs);
-        buffer_sprintf(wb, "\"%s\": \"%s\"", label->key, label->value);
+
+        char value[CONFIG_MAX_VALUE + 1];
+        escape_json_string(value, label->value, CONFIG_MAX_VALUE);
```
amoss: Maximum length config value with multiple characters to be escaped will cause a buffer-overflow at this point.

libnetdata/inlined.h
```diff
@@ -246,6 +246,20 @@ static inline char *strncpyz(char *dst, const char *src, size_t n) {
     return p;
 }
 
+static inline void escape_json_string(char *dst, char *src, size_t len) {
```
amoss: The result of this is being written into a resizing buffer, why are we not using it instead of a fixed-length temporary array?

web/api/formatters/json_wrapper.c
vlvkobal: It stops copying data when maximum length is reached.

web/api/web_api_v1.c
vlvkobal: Look at the previous comment.

web/api/formatters/json_wrapper.c
amoss: So it will output a truncated label - which is not correct either.

libnetdata/inlined.h
vlvkobal: For label creation, you tried to avoid memory allocations sacrificing readability. Why do we need to use resizing buffer here, when these operations will take place much more frequently than label creation. Are we afraid of losing some characters in insanely long label values in API responses?

amoss: At this point in the code we are building a response in a resizing buffer already so there is no sacrifice in performance.

Double the size of the buffer to handle the worst-case.

vlvkobal: Done

PR #7509, opened 2019-12-12
Close kubernetes

### LINKED ISSUE #7411: Script to pull kubernetes labels from pod API
* Extract the part of `cgroup-names.sh` that accesses the pod API.
* Extract the checks that guard execution only if inside a pod.
* Extend to extract the user-defined labels from the data returned from the API.
* Make this work in the initial environment of the agent (i.e. `exec()` the script before threads are spawned in the daemon).
* Check performance of this approach and that it does not slow down agent startup.
* Ensure this also works if called from a thread in the agent (to support reloading changes).
* Create a linked list using the struct from #7408 but do not attempt to link it into RRDHOST.

Some information about Kubernetes labels is available at https://www.replex.io/blog/9-best-practices-and-examples-for-working-with-kubernetes-labels.
There is some similar recent work in: https://github.com/netdata/netdata/pull/7416

cakrit:
##### Summary
Fixes #7411 
- Add check for script exit code and handle it appropriately (logging, freeing)
- Fix length check for label value 
- Add log message when a label is ignored due to invalid name
- Remove test return values for the script
- Convert "Attempting to" msg from info to debug

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,6 +113,7 @@ system/netdata-freebsd
 system/edit-config
 
 daemon/anonymous-statistics.sh
+daemon/get-kubernetes-labels.sh
 
 health/notifications/alarm-notify.sh
 collectors/cgroups.plugin/cgroup-name.sh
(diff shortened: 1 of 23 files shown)


cakrit:
wrong branch

PR #7524, opened 2019-12-13
Add Fedora 31 CI integrations.

Ferroin:
##### Summary

Fedora 31 has been out for roughly 2.5 months (release date was 2019-10-29). It's about time we add package generation and publishing for it to CI.

##### Component Name

area/packaging

##### Additional Information

Sample build results: https://travis-ci.org/Ferroin/netdata/jobs/624623473?utm_medium=notification&utm_source=email

I've not rigorously tested that the generated RPM works perfectly, but it appears to install correctly, and it looks like it runs, collects data, and serves the Web UI correctly. However, it would be good to have som additional testing of this before merging. Note that I've only tested the amd64 package (I don't have an arm64 system I can test against).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -421,6 +421,14 @@ jobs:
         - PACKAGE_TYPE="rpm" REPO_TOOL="yum"
         - ALLOW_SOFT_FAILURE_HERE=true
 
+    - name: "Build & Publish RPM package for Fedora 31"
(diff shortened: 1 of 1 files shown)


Ferroin:
This now also adds Fedora 31 lifecycle checks to CI. These were previously disabled by 6e68ff6f8 because Fedora 31 was not yet released at the time.

PR #7525, opened 2019-12-13
Agent claiming

### LINKED ISSUE #7086: Agent Claiming - Core

mfundul:
##### Summary
Fixes #7086
##### Component Name
daemon
packaging
##### Additional Information
Add the option to claim nodes in workspaces.


joelhans:
One thing that's always bugged me a bit—not Markos' fault by any means and not limited to this PR—is how we seem to use **agent** and **node** interchangeably throughout Netdata as a whole. I'd love to bring about some consistency in this doc and others, but I've never had clarity on which word is more "correct," or if they actually do mean different things.

Maybe...
agent = Netdata daemon?
node = the system on which the Netdata daemon runs?

PR #7548, opened 2019-12-16
Alarm Log labels

### LINKED ISSUE #7555: Complete the alarms support for labels.
There were some tasks left over from #7096 that needed to be completed after a dependency was resolved. In order to add the labels to the alarm_log there are several parts that were not discussed. 

* The labels need to be added to ALARM_ENTRY. 
* The health-log.db format needs to be expanded. 
* Write/parse labels to/from the file.
* Don't store labels in each alarm entry in the alarm log, but put all of them in a new type of line (e.g. L l1=v1 l2=v2 l3=v4)

thiagoftsm:
##### Summary
Fixes #7555 and https://github.com/netdata/product/issues/144 

##### Component Name
Health
##### Additional Information
Explaining the road selected to bring #7555 and updated on https://github.com/netdata/product/issues/144.

- The labels need to be added to ALARM_ENTRY and the line must start with `L`.

To test this feature, it is only necessary to create an alarm with label.


health/health_log.c
```diff
@@ -132,6 +134,7 @@ inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char *filena
 
     netdata_rwlock_rdlock(&host->health_log.alarm_log_rwlock);
 
+    netdata_rwlock_rdlock(&host->labels_rwlock);
```
amoss: Why do you hold the labels lock around this loop? I can't find any access to host->labels inside the locked section.

amoss:
For testing I created an alarm like this:
```
../dev-install/netdata/etc/netdata/edit-config health.d/cpu.conf
```
The test alarm contains this:
```
alarm: testing                                                                                                          
    on: system.cpu                                                                                                      
lookup: sum -3s at 0 every 3 percentage of user                                                                         
 units: %                                                                                                               
 every: 1s                                                                                                              
  warn: $this > 10                                                                                                      
  crit: $this > 20                                                                                                      
  label: _is_master=false
```
I've checked that it appears in the `health-log.db`, e.g. :
```
U       loki    5dfa0cb9        5dfa096e        0000000b        5dfa0d0c        5dfa0c66        5dfa3464        00000000        00000000        10000002        00000000        5dfa3464        testing system.cpu      cpu                     1@/home/amoss/netdata/dev-install/netdata/etc/netdata/health.d/cpu.conf %               0       0       0       0       nan     nan     0000000000000000        _is_master=false
```
When I check the alarm log through the API I can see that the label has been inserted into the json correctly, e.g. `url http://localhost:19999/api/v1/alarm_log? | less` contains entries like this:
```
        {
                "hostname": "loki",
                "unique_id": 1576668768,
                "alarm_id": 1576667502,
                "alarm_event_id": 24,
                "name": "testing",
                "chart": "system.cpu",
                "family": "cpu",
                "processed": true,
                "updated": true,
                "exec_run": 1576679652,
                "exec_failed": false,
                "exec": "/home/amoss/netdata/dev-install/netdata/usr/libexec/netdata/plugins.d/alarm-notify.sh",
                "recipient": "root",
                "exec_code": 0,
                "source": "1@/home/amoss/netdata/dev-install/netdata/etc/netdata/health.d/cpu.conf",
                "units": "%",
                "when": 1576679652,
                "duration": 203,
                "non_clear_duration": 0,
                "status": "WARNING",
                "old_status": "CLEAR",
                "delay": 0,
                "delay_up_to_timestamp": 1576679652,
                "updated_by_id": 1576668769,
                "updates_id": 1576668767,
                "value_string": "10.5%",
                "old_value_string": "6.73%",
                "last_repeat": "0",
                "silenced": "false",
                "label": "_is_master=false",
                "info":null,
                "value":10.5307415,
                "old_value":6.7319796
        },
```

But I see a problem in some of the entries that are not the alarm that I'm checking, in some cases alarms without a label filter contain: 
```
                "label": "",

```
and in some cases it is this:
```
                "label": "null",
```
Why the difference?

health/health_log.c
thiagoftsm: When I was developing, I was thinking in the possibility to do not load the logs case the labels were not present, but I changed my mind late and I forgot to remove this lock.

thiagoftsm:
> Why the difference?

You already had logs that were created without labels and Netdata assigned NULL to strcuture, because it was allocated with `callocz`.

amoss:
> > Why the difference?
> 
> You already had logs that were created without labels and Netdata assigned NULL to strcuture, because it was allocated with `callocz`.

So the behavior after this patch is to use `"label" : ""` and the `"label" : "null"` is just historical. This would make sense from the position in the log, although I can't verify the timestamps.

vlvkobal:
My configuration:
```
[host labels]
	my_label2 = begin
	label1 = value1
	label2 = value2
	label3 = val_ue3
```
```
alarm: my_alarm1
on: system.cpu
lookup: average -5s of user
every: 2s
crit: $this > 50

alarm: my_alarm2
on: system.cpu
lookup: average -5s of user
every: 2s
crit: $this > 50
label: my_label2 = begin

alarm: my_foreach_alarm1
on: system.cpu
lookup: average -5s foreach user,system
every: 2s
crit: $this > 50

alarm: my_foreach_alarm2
on: system.cpu
lookup: average -5s foreach user,system
every: 2s
crit: $this > 50
label: my_label2 = begin

template: my_template1
on: cpu.cpu
lookup: average -5s of user
every: 2s
crit: $this > 50

template: my_template2
on: cpu.cpu
lookup: average -5s of user
every: 2s
crit: $this > 50
label: my_label2 = begin

template: my_foreach_template1
on: cpu.cpu
lookup: average -5s foreach user,system
every: 2s
crit: $this > 50

template: my_foreach_template2
on: cpu.cpu
lookup: average -5s foreach user,system
every: 2s
crit: $this > 50
label: my_label2 = begin
```
There is only one alarm in the alarm log with `"label": "my_label2=begin"` - `"name": "my_alarm2"`. For everything else there is `"label": "null"` in the log.

```
"name": "my_foreach_alarm2_user"
"label": "null"

"name": "my_template2"
"label": "null"

"name": "my_foreach_template2_user"
"label": "null"
```

thiagoftsm:
> There is only one alarm in the alarm log with `"label": "my_label2=begin"` - `"name": "my_alarm2"`. For everything else there is `"label": "null"` in the log.

I verified now and it looks like that you find a problem inside Netdata with your tests that affects this PR, but it is not a bug in this PR. When I used your `netdata.conf` configuration together this alarm configuration:

```
 alarm: dev_dim_template_system
    on: system.cpu
    os: linux
lookup: sum -3s at 0 every 3 percentage of system
 units: %
 every: 1s
  warn: $this > 1
  crit: $this > 4
label : label1 = value1

 alarm: dev_dim_template_user
    on: system.cpu
    os: linux
lookup: sum -3s at 0 every 3 percentage of user
 units: %
 every: 1s
  warn: $this > 1
  crit: $this > 4
label : label2 = value2

 alarm: dev_dim_template_nice
    on: system.cpu
    os: linux
lookup: sum -3s at 0 every 3 percentage of nice
 units: %
 every: 1s
  warn: $this > 1
  crit: $this > 4
label : label3 = val_ue3
```

I could have different alarms raised and stored inside the database.

You used only `my_label2` in your alarms, so we expect to have only them on logs, but when I changed your configuration setting different labels, I also only could have `my_label2` on my alarms. Considering that your alarms apply to only one chart and they have the same dimension target, it looks like that Netdata stores the first alarm and ignore others.

amoss:
This seems to have slipped as the milestone is set to the older sprint. What was the outcome on this? Our last set of reviews were accepts but it was never merged. I recall that there was a discussion in a review session about whether this was the right approach, or if we should strip the labels from the alarm_log completely but I cannot find any documentation on what we decided. What was the decision, and does it mean that we are merging or cancelling this piece of work?

vlvkobal:
This PR is incomplete. After the last changes, it saves labels to the `health-log.db` file, but doesn't parse the labels and doesn't add them to the ALARM_ENTRY.

BTW, the description in the OP doesn't match the #7555.

amoss:
I thought that there was a problem with writing the labels to each line of the `health-log.db` file and that the resolution was to strip the labels, and insert a new kind of line indicating the state of the labels when the host starts?

thiagoftsm:
> This seems to have slipped as the milestone is set to the older sprint. What was the outcome on this? Our last set of reviews were accepts but it was never merged. I recall that there was a discussion in a review session about whether this was the right approach, or if we should strip the labels from the alarm_log completely but I cannot find any documentation on what we decided. What was the decision, and does it mean that we are merging or cancelling this piece of work?

The outcome was to add all labels to alarm log creating a new identification inside alarm logs that starts with `L`.
The definitions were set in this link https://github.com/netdata/product/issues/144.

thiagoftsm:
> This PR is incomplete. After the last changes, it saves labels to the `health-log.db` file, but doesn't parse the labels and doesn't add them to the ALARM_ENTRY.
> 
> BTW, the description in the OP doesn't match the #7555.

It does not match #7555, because we changed what we wanted after the meeting on December 20th, 2019. I copied the link with the newest definitions in the previous answer.

thiagoftsm:
> Definition of the task changed after the demo, please update to the new description.

I was not in the Demo today, please, could you give me the new description?

amoss:
@thiagoftsm: I think the confusion is due to me - when I said the demo I meant the meeting on Dec 20th. I checked the definitions in https://github.com/netdata/product/issues/144 and this matches what I remember. I'll review the code.

health/health_log.c
```diff
@@ -67,6 +67,35 @@ inline void health_log_rotate(RRDHOST *host) {
...
+        write[wb->len-1] = '\0';
+
+        if (unlikely(fprintf(host->health_log_fp, "L\t%s"
+                , write
+        ) < 0))
+            error("HEALTH [%s]: failed to save alarm log entry to '%s'. Health data may be lost in case of abnormal restart.",
+                  host->hostname, host->health_log_filename);
+        else {
+            host->health_log_entries_written++;
+        }
```
amoss: Please add `buffer_free(wb);`. The alternative is to make it static / single-initialization, but this is a very infrequent operation and we can probably save the memory instead.

Issue #7486, opened 2019-12-09
Send labels from slave to master


cakrit:
The streaming protocol needs to be extended, so that slaves can send host labels to their master.

stelfrag:
Adding a command to the protocol will require both master + slave to be at the same version (or at least, for the master, the minimum version that will support the new command). 

This is because any command that  the master does not understand, terminates the streaming connection. 

This will happen  if there are cases where slave agents may be updated and masters are not, 

Just a thought ..

If one of the existing commands is used (e.g. VARIABLE), it could allow the "labels" enabled slave to send labels to a master as special variables e.g. LABEL_ prefixed (of course the master would not know they are labels), until the master is also updated to an appropriate version and handle those LABEL_ variables

thiagoftsm:
I agree with you, I think it will be a safe choice avoid to send the labels a second time case the streaming is closed and we will send a message for the user error log requesting an update on master side.

amoss:
This should not be added to the sprint milestone without an estimate of the work and agreement from the team.

thiagoftsm:
From product channel:

Streaming: We add a protocol version to the STREAM get request. The server responds with support for that version, if it has been upgraded. Then we know we can send the host labels

amoss:
Merged but not closed?

PR #7549, opened 2019-12-16
Stream with labels

thiagoftsm:
##### Summary
Implements #7486 

##### Component Name
Stream
##### Additional Information
During the development I used this branch and I also used other branch without these changes, and the following tests were done:

- Master and slave using `stream_with_label` . For this case we must have labels on master for master and slave.
- Master with `stream_with_label` and slave with another branch. In this case only master has labels.
- Master with another branch and slave with `stream_with_label`. Another case that is only possible to have labels on master.

To confirm that labels were delivered from slave to master, I always used `host/HOSTNAME/api/v1/info` and I also saw `error.log` to confirm that the expected messages were there.


database/rrdhost.c
```diff
@@ -889,6 +889,21 @@ struct label *create_label(char *key, char *value, LABEL_SOURCE label_source)
     return result;
 }
 
+void clean_label_list(RRDHOST *host, struct label *new_labels)
```
amoss: This isn't a cleaning operation -> a better name would be `replace_label_list`.

But, as I'm about to explain above, I don't understand why you've factored this out of the reload function to call from another file.

collectors/plugins.d/plugins_d.c
```diff
@@ -616,6 +620,39 @@ inline size_t pluginsd_process(RRDHOST *host, struct plugind *cd, FILE *fp, int
...
+        else if(likely(hash == OVERWRITE_HASH && !strcmp(s, PLUGINSD_KEYWORD_OVERWRITE))) {
+            debug(D_PLUGINSD, "requested a OVERWITE a variable");
+            if(!host->labels) {
+                host->labels = new_labels;
+            } else {
+                clean_label_list(host, new_labels);
+            }
+
+            new_labels = NULL;
+        }
```
amoss: I'll tag comments on this line, but really they apply to the whole of this block.

I don't understand what you are doing here, or why - there is no description in the issue of what you are trying to achieve. It looks like you are adding some kind of feature that will overwrite the entire set of host labels and I don't understand which requirement this would come from. I think we should probably discuss whether or not this is a requirement of host labels and their support in streaming.

In addition - it is not a good idea to replace the `host->labels` without holding the write lock. Even if the value is null it could be overwritten by a CLI generated reload operation?

database/rrdhost.c
thiagoftsm: When we are receiving the stream from another host, Netdata will update the `RRDHOST` for that connection instead the `localhost`, this will force us to remove the previous labels that were set when the connection started and we will load a new, this is the motive I began to export this function.

collectors/plugins.d/plugins_d.c
thiagoftsm: By steps, every time that the stream sends a new group of labels, we will update that labels on master, but we are not overwritten the master label, it is the slave database instance that is changed.

We had two options to do this, but in any case the protocol would need to be updated. Case I decided to use the existent keywords I would need to add conditions inside the previous parser to identify a label or a chart value, or I could create two new commands that would be responsible exclusively for labels. I opted for the second option, because I thought it would be better to have different commands for charts and labels, because the data types are different.

amoss:
@underhood This already has two approving reviews from the team, it does not need a third from me :) Please go ahead and merge.

thiagoftsm:
Thank you guys! :)

PR #7554, opened 2019-12-17
Send host labels via exporting connectors 

### LINKED ISSUE #7406: Send host labels via the exporting engine.
"As a user, I want to attach custom labels to a host and see those labels on every metric of that host that appears in my archiving data stream, so that I can better search/filter/group through the metrics I aggregate in my TSDB of choice"

- Add labels to all target destinations 
- Add option to exporting engine on whether to include internal labels ('_XXX').
- Only use the new exporting engine (no back-porting into backends code)

vlvkobal:
##### Summary
Add support for host labels to the following exporting connectors
- [x] JSON
- [x] Graphite
- [x] OpenTSDB

Closes #7406

##### Component Name
exporting engine


ilyam8:
> Send host labels via exporting connectors

Are you going to enable it by default?

vlvkobal:
> Are you going to enable it by default?

Only labels configured by a user are enabled by default.

exporting/graphite/graphite.c
```diff
@@ -51,6 +51,74 @@ int init_graphite_instance(struct instance *instance)
...
+ * @param instance an instance data structure.
+ * @param host a data collecting host.
+ * @return Always returns 0.
+ */
+int format_host_labels_graphite_plaintext(struct instance *instance, RRDHOST *host)
+{
+    if (!instance->labels)
+        instance->labels = buffer_create(1024);
+
+    if (!instance->labels)
```
thiagoftsm: A suggestion, instead to have this condition here, you could move it for line 89 and put `{ }` for the if, because this condition only makes sense after you try to allocate again, of course, this suggestion can be ignored case this function is called only one time.

exporting/json/json.c
```diff
@@ -51,6 +51,52 @@ int init_json_instance(struct instance *instance)
...
+ * @param instance an instance data structure.
+ * @param host a data collecting host.
+ * @return Always returns 0.
+ */
+int format_host_labels_json_plaintext(struct instance *instance, RRDHOST *host)
+{
+    if (!instance->labels)
+        instance->labels = buffer_create(1024);
+
+    if (!instance->labels)
```
thiagoftsm: The same suggestion for graphite, can be applied here.

exporting/graphite/graphite.c
vlvkobal: You are right. If there is a memory allocating error `callocz()` and `mallocz()` call `fatal()` so there is no need to check the result.

exporting/exporting_engine.h
```diff
@@ -13,44 +13,62 @@
...
+    EXPORTING_SOURCE_DATA_SUM               = (1 << 2),
 
-    EXPORTING_SOURCE_DATA_AS_COLLECTED = (1 << 0),
-    EXPORTING_SOURCE_DATA_AVERAGE      = (1 << 1),
-    EXPORTING_SOURCE_DATA_SUM          = (1 << 2),
+    EXPORTING_OPTION_SEND_CONFIGURED_LABELS = (1 << 3),
+    EXPORTING_OPTION_SEND_AUTOMATIC_LABELS  = (1 << 4),
 
-    EXPORTING_OPTION_SEND_NAMES        = (1 << 16)
+    EXPORTING_OPTION_SEND_NAMES             = (1 << 16)
```
thiagoftsm: Did you jump from 8 (1 << 3) to 32768 ( 1<< 16) because SEND_NAMES will be always last? Or did you do a math on his mind and there is a wrong jump here?

exporting/json/json.c
```diff
@@ -51,6 +51,44 @@ int init_json_instance(struct instance *instance)
...
+    buffer_strcat(instance->labels, "\"labels\":{");
+
+    int count = 0;
+    netdata_rwlock_rdlock(&host->labels_rwlock);
+    for (struct label *label = host->labels; label; label = label->next) {
+        if (!should_send_label(instance, label))
+            continue;
+
+        char value[CONFIG_MAX_VALUE * 2 + 1];
+        sanitize_json_string(value, label->value, CONFIG_MAX_VALUE);
```
thiagoftsm: `value` variable has two times the size given as argument, should we not increase the argument value?

exporting/opentsdb/opentsdb.c
```diff
@@ -83,6 +83,58 @@ int init_opentsdb_http_instance(struct instance *instance)
...
+ */
+
+void sanitize_opentsdb_label_value(char *dst, char *src, size_t len)
+{
+    while (*src != '\0' && len) {
+        if (isalpha(*src) || isdigit(*src) || *src == '-' || *src == '_' || *src == '.' || *src == '/')
+            *dst++ = *src;
+        else
+            *dst++ = '_';
+        src++;
```
thiagoftsm: Do not we accept UTF8 values for OpenTSDB values?

exporting/graphite/graphite.c
```diff
@@ -51,6 +51,60 @@ int init_graphite_instance(struct instance *instance)
...
+ */
+
+void sanitize_graphite_label_value(char *dst, char *src, size_t len)
+{
+    while (*src != '\0' && len) {
+        if (isspace(*src) || *src == ';' || *src == '~')
+            *dst++ = '_';
+        else
+            *dst++ = *src;
+        src++;
```
thiagoftsm: Do not we accept UTF-8 values for graphite?

exporting/opentsdb/opentsdb.c
```diff
@@ -192,6 +246,42 @@ static inline void opentsdb_build_message(BUFFER *buffer, char *message, const c
...
+    if (unlikely(!sending_labels_configured(instance)))
+        return 0;
+
+    netdata_rwlock_rdlock(&host->labels_rwlock);
+    for (struct label *label = host->labels; label; label = label->next) {
+        if (!should_send_label(instance, label))
+            continue;
+
+        char escaped_value[CONFIG_MAX_VALUE * 2 + 1];
+        sanitize_json_string(escaped_value, label->value, CONFIG_MAX_VALUE);
```
thiagoftsm: The variable `escaped_value` has two times the size given as argument, should we not have a bigger argument length?

exporting/exporting_engine.h
vlvkobal: I just kept the original values unchanged
https://github.com/netdata/netdata/blob/master/backends/backends.h#L8-L16

exporting/json/json.c
vlvkobal: No. The `value` variable should be twice as large as the `label->value` in case every symbol in the string is escaped.

exporting/tests/test_exporting_engine.c
```diff
@@ -678,6 +680,103 @@ static void test_simple_connector_worker(void **state)
...
+
+    assert_string_equal(dst, "check____string");
+}
+
+static void test_sanitize_opentsdb_label_value(void **state)
+{
+    (void)state;
+
+    char *src = "check \t\\\" #&$? -_./ string";
+    char dst[26 + 1];
```
thiagoftsm: The string `*src` is 29 bytes long

thiagoftsm: do you really want to copy  less bytes from src?

exporting/exporting_engine.h
thiagoftsm: All right, thanks!

exporting/json/json.c
thiagoftsm: Ok, I got it.

exporting/graphite/graphite.c
vlvkobal: Yes, we accept. We don't touch any characters except spaces, `;`, and `~`.

exporting/opentsdb/opentsdb.c
vlvkobal: Already answered.

exporting/tests/test_exporting_engine.c
vlvkobal: After sanitizing it becomes a 26-character string.

exporting/graphite/graphite.c
thiagoftsm: When we merge the PR that you already reviwed, I think you do not need to go every address, instead, you need to use the function that we already have inside Netdata to measure the size of the character to advance.

exporting/tests/test_exporting_engine.c
vlvkobal: It's a test. I'm not copying, I'm checking the sanitizing function. I don't need more bytes in `dst`.

thiagoftsm: All right, I only wanted to be sure about what we are testing here, thanks!

exporting/opentsdb/opentsdb.c
vlvkobal: Fixed

thiagoftsm: Thanks! :)

exporting/graphite/graphite.c
vlvkobal: It was discussed and a decision was made to keep checks for every character because otherwise there is a risk to access memory outside allocated boundaries.

Ehekatl:
hope this can get merge soon, can't wait to try exporting with multiple targets

vlvkobal:
This PR just adds host labels to the output of exporting connectors. The exporting engine itself is already in the master branch and you can send metrics to multiple targets. Please read https://github.com/netdata/netdata/blob/master/exporting/README.md.

Ehekatl:
@vlvkobal ya, but host labels are enssentail in our use case

vlvkobal:
You can set `host tags` in the `backend` section of `netdata.conf`.

vlvkobal:
I forgot to mention that, unfortunately, you can't use different exporters when you use `host tags` as far as the format for tags is not the same for different TSDBs.

Ehekatl:
that's ok, we have multiple backend with the same type
we want double write for HA

underhood:
After the commit dfa1541 everything working correctly.

PR #7559, opened 2019-12-17
Skip unit testing during CI when it's not needed.

Ferroin:
##### Summary

Our current unit testing takes almost 20 minutes on average during the CI process, which is almost twice as long as any other step, and it's only going to get longer as we get better test coverage.

This commit slightly alters how we handle unit testing during CI so that only those unit tests which are actually needed get run.

This is achieved by using the `$TRAVIS_COMMIT_RANGE` environment variable provided by Travis to determine which commits we're actually testing, and then using that information to query Git for what files actually changed.

As of right now, the only unit testing we're doing is in the dbengine code, so we check to see if any of 
those files or the C unit testing files are modified by the PR and only runs the unit tests for those if they were modified. The checks are a bit more wide-ranging than they probably need to be so that we make sure to avoid any false negatives.

For the cases where we don't actually need to run unit tests, this provides a net speedup of about 14-16 minutes (it would be a bit better, but we need a full deep clone of the repo for this to work reliably, so the clone step is taking a bit longer than it was previously).

##### Component Name

area/ci

##### Additional Information

Internally this follows these rules to determine whether to run unit tests or not:

1. If Travis supplied no commit information (which means that this is the first push of a new branch), run all the unit tests.
2. If we can't identify the first or last commit of the range, run all the tests. This happens when a branch is force pushed causing a history rewrite.
3. Otherwise, query `git` to produce a list of all files changed by the range of commits we're testing. If any files changed that are either unit tested or handle the running of the unit tests, run the appropriate unit tests.
4. Otherwise, skip the unit tests.

Sample build for rule 2 above: https://travis-ci.org/Ferroin/netdata/builds/626224019
Sample build for rule 3 above: https://travis-ci.org/Ferroin/netdata/builds/626233338
Sample build for rule 4 above: https://travis-ci.org/Ferroin/netdata/builds/626206692

This produces a net speedup of builds that don't actually need unit testing of roughly 15-16 minutes.

Long-term, I'm planning on updating rules 1 and 2 above to just look at the entirety of the branch (unless it's the master branch) to determine whether to run tests instead of just always running them under those conditions.

### DIFF
--- a/.travis/run-unit-tests.sh
+++ b/.travis/run-unit-tests.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+#
+# Unit-testing script
+#
+# This script does the following:
+#   1. Check whether any files were modified that would necessitate unit testing (using the `TRAVIS_COMMIT_RANGE` environment variable).
+#   2. If there are no changed files that require unit testing, exit successfully.
+#   3. Otherwise, run all the unit tests.
+#
+# We do things this way because our unit testing takes a rather long
+# time (average 18-19 minutes as of the original creation of this script),
+# so skipping it when we don't actually need it can significantly speed
+# up the CI process.
+#
+# Copyright: SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Author: Austin S. Hemmelgarn <austin@netdata.cloud>
+#
+# shellcheck disable=SC2230
+
+install_netdata() {
+    echo "Installing Netdata"
+    fakeroot ./netdata-installer.sh --install $HOME --dont-wait --dont-start-it --enable-plugin-nfacct --enable-plugin-freeipmi --disable-lto
+}
+
+c_unit_tests() {
+    echo "Running C code unit tests"
+    $HOME/netdata/usr/sbin/netdata -W unittest
+}
+
+run_c_unit_tests=
+
+if [ -z ${TRAVIS_COMMIT_RANGE} ] ; then
+    # Travis gave us no commit range, so just run all the unit tests.
+    # Per the docs, this is the case when a new branch is pushed for the first time.
+    run_c_unit_tests=1
+else
+    COMMIT1="$(echo ${TRAVIS_COMMIT_RANGE} | cut -f 1 -d '.')"
+    COMMIT2="$(echo ${TRAVIS_COMMIT_RANGE} | cut -f 4 -d '.')"
+
+    if [ "$(git cat-file -t ${COMMIT1} 2>/dev/null)" = commit -a "$(git cat-file -t ${COMMIT2} 2>/dev/null)" = commit ] ; then
+        changed_paths="$(git diff --numstat ${COMMIT1}..${COMMIT2} --  | cut -f 3)"
+
+        # Check for changes that would require the C code to be re-tested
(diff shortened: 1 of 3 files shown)


.travis/run-unit-tests.sh
```diff
@@ -0,0 +1,67 @@
...
+# shellcheck disable=SC2230
+
+install_netdata() {
+    echo "Installing Netdata"
+    fakeroot ./netdata-installer.sh --install $HOME --dont-wait --dont-start-it --enable-plugin-nfacct --enable-plugin-freeipmi --disable-lto
+}
+
+c_unit_tests() {
+    echo "Running C code unit tests"
+    $HOME/netdata/usr/sbin/netdata -W unittest
```
thiagoftsm: Please, add here a call for `make check`, because it calls other unit tests we have.

Ferroin: That would be the CMocka stuff, right?

If so, that's on the list of stuff to add, it's just not included in this PR because we aren't actually doing it on master right now anyway, and my goal here was just to handle what we were already doing.

thiagoftsm: Yes, you are correct.
Thank you for the explanation!

Ferroin:
As a note to anyone looking at this PR, I'd like to leave it open at least for a couple of days so that anybody it might affect has a chance to comment. Assuming I've got the required approvals and nobody objects, I'll plan to merge it Thursday.

.travis/run-unit-tests.sh
```diff
@@ -0,0 +1,67 @@
...
+if [ -z ${TRAVIS_COMMIT_RANGE} ] ; then
+    # Travis gave us no commit range, so just run all the unit tests.
+    # Per the docs, this is the case when a new branch is pushed for the first time.
+    run_c_unit_tests=1
+else
+    COMMIT1="$(echo ${TRAVIS_COMMIT_RANGE} | cut -f 1 -d '.')"
+    COMMIT2="$(echo ${TRAVIS_COMMIT_RANGE} | cut -f 4 -d '.')"
+
+    if [ "$(git cat-file -t ${COMMIT1} 2>/dev/null)" = commit -a "$(git cat-file -t ${COMMIT2} 2>/dev/null)" = commit ] ; then
+        changed_paths="$(git diff --numstat ${COMMIT1}..${COMMIT2} --  | cut -f 3)"
```
ncmans: You can use `--name-only` flag instead of `--numstat` and `cut` combo.

Ferroin: Good point, had completely forgotten about `--name-only`.

Ferroin:
Marking this WIP. I slightly misunderstood how `TRAVIS_COMMIT_RANGE` works and as a result the code here won't properly handle PR's that are being incrementally updated (that is, the user is pushing new commits on top of the PR branch to update the PR).

I'll update things here once I've got that fixed (probably some time later this week).

Ferroin:
This is ready for review again.

The following two changes have been made:

* When handling PR's, it now looks at the whole commit history of the PR relative to the target branch. This ensures that PR's which need unit testing will always have it run.
* We now list the full set of commits we're looking at in the build log using `git log --format=oneline --abbrev-commit`. This will make it easier to spot any issues with the detection logic since we will know exactly what commits the script is examining.

PR #7560, opened 2019-12-17
Revert "Fix race condition in dbengine (#7533)"

amoss:
We are removing this fix for further internal testing, it will be returning after we iron out
some bugs.

This reverts commit 53ab093d84919c743450199a31bca9a13412e451.


thiagoftsm:
When I run this PR with `-D` and I press `^-c` I am receiving the following messages:

```
/usr/libexec/netdata/plugins.d/tc-qos-helper.sh: line 233: echo: write error: Broken pipe
/usr/libexec/netdata/plugins.d/tc-qos-helper.sh: line 243: echo: write error: Broken pipe
/usr/libexec/netdata/plugins.d/tc-qos-helper.sh: line 292: echo: write error: Broken pipe
```

Issue #7568, opened 2019-12-18
api calls 403 error httpd CentOS 6.10


kkoomen:
##### Bug report summary

I use httpd on CentOS 6.10 and the calls to get all the data do work for several minutes, but eventually it stops working (the data is not updating anymore) and I'm getting 403 errors for the api calls. See screenshot:

![Screenshot 2019-12-18 at 10 20 35](https://user-images.githubusercontent.com/10693490/71051339-fe0dbe00-2182-11ea-8e7d-84bd415b1921.png)

The above errors occur because the response returns a server 403.

![Screenshot 2019-12-18 at 11 42 36](https://user-images.githubusercontent.com/10693490/71054061-85f7c600-218b-11ea-8fc7-1bcd300b87b1.png)

##### OS / Environment

CentOS release 6.10 (Final)

##### Netdata version (ouput of `netdata -V`)

netdata v1.19.0-129-nightly

##### Steps To Reproduce

1) Setup netdata

2) Add the following configuration in `/etc/netdata/netdata.conf`

```
[global]
    access log = none
[web]
    bind to = localhost
    enable gzip compression = no
```

3) Add the following apache directives to `/etc/httpd/conf/httpd.conf` (or in my case: a virtualhost-specific config file located at `/etc/httpd/conf/sites-available/my-website.conf.vhost`):

```
RewriteEngine On
ProxyRequests Off
ProxyPreserveHost On

ProxyPass "/" "http://localhost:19999/" connectiontimeout=5 timeout=30 keepalive=on
ProxyPassReverse "/" "http://localhost:19999/"
```

##### Expected behavior

It should be able to do api calls without getting a 403.

kkoomen:
I am seeing errors in my httpd log that say this (replaced my production url with `my-website.com`):

```
[Wed Dec 18 12:09:32 2019] [info] Initial (No.1) HTTPS request received for child 10 (server my-website.com:443)
[Wed Dec 18 12:09:32 2019] [error] [client 67.212.238.232] client denied by server configuration: proxy:http://localhost:19999/api/v1/alarms, referer: https://my-website.com/
[Wed Dec 18 12:09:32 2019] [debug] mod_proxy_http.c(56): proxy: HTTP: canonicalising URL //localhost:19999/error/403.html
[...]

PR #7578, opened 2019-12-18
Docs: Add notice about mod_evasive to Apache proxy guide

joelhans:
##### Summary

Issue #7568 resurfaced an issue that Netdata's dashboard has with Apache's `mod_evasive` module, which protects from DDoS attacks. Because accessing the dashboard creates a high volume of requests, the module's default config gets the user blocked pretty quickly.

This PR adds text to the Apache proxy guide to give notice about this module and tell users to edit their `mod_evasive.conf` file to mitigate the issue.

##### Component Name

docs

### DIFF
--- a/docs/Running-behind-apache.md
+++ b/docs/Running-behind-apache.md
@@ -227,6 +227,31 @@ If you want to enable CSP within your Apache, you should consider some special r
 
 Note: Changes are applied by reloading or restarting Apache.
 
+## Using Netdata with Apache's `mod_evasive` module
+
+The `mod_evasive` Apache module helps system administrators protect their web server from brute force and distributed
+denial of service attack (DDoS) attacks.
+
+Because Netdata sends a request to the web server for every chart update, it's normal to create 20-30 requests per
+second, per client. If you're using `mod_evasive` on your Apache web server, this volume of requests will trigger the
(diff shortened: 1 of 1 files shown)


joelhans:
@kkoomen Feel free to leave feedback too. I can't add you as a reviewer, but perhaps you can create a review yourself. I thought about adding something about configuring just one virtualhost instead of the entire module, but I'm worried it gets too complex at that point. If you have a simple config/explanation that works, we can maybe add that too.

docs/Running-behind-apache.md
```diff
@@ -227,6 +227,35 @@ If you want to enable CSP within your Apache, you should consider some special r
...
+exceed that threshold, and `mod_evasive` will add your IP address to a blocklist.
+
+Our users have found success by setting `DOSPageCount` to `30`. Try this, and raise the value if you continue to see 403
+errors while accessing the dashboard.
+
+```conf
+DOSPageCount 30
+```
+
+Restart Apache with `sudo service apache2 restart`, or the appropriate method to restart services on your system, to
```
cosmix: I believe there’s no need to restart. They can just `reload`

kkoomen: @cosmix Just tested it out. A restart is required.

kkoomen:
@joelhans I've added additional documentation and fixed 2 typos you made as well. See #7582.

PR #7584, opened 2019-12-19
Set standard name to non-libnetdata threads (libuv, pthread)

Saruspete:
##### Summary
Using the same logic as libnetdata' `thead_set_name()`, we should set the name of the other threads.

Implementing `uv_thread_set_name` within libnetdata for libuv threads

Came from issue #7531 / [comment #566285792](https://github.com/netdata/netdata/issues/7531#issuecomment-566285792) 

##### Component Name
- libnetdata/threads/
- exporting
- database/engine
- daemon/command
- collector/proc.plugin/proc_stat

##### Additional Information

Generated output: 

```
 1886  1886 17848 netdata         ./netdata -D
 1886  1896 17848 DBENGINE        ./netdata -D
 1886  1897 17848 PLUGIN[proc]    ./netdata -D
 1886  1898 17848 PLUGIN[diskspac ./netdata -D
 1886  1899 17848 PLUGIN[cgroups] ./netdata -D
 1886  1900 17848 PLUGIN[tc]      ./netdata -D
 1886  1901 17848 PLUGIN[idlejitt ./netdata -D
 1886  1902 17848 STATSD          ./netdata -D
 1886  1905 17848 WEB_SERVER[stat ./netdata -D
 1886  1906 17848 PLUGINSD        ./netdata -D
 1886  1907 17848 HEALTH          ./netdata -D
 1886  1908 17848 DAEMON_COMMAND  ./netdata -D
 1886  1911 17848 PLUGINSD[slabin ./netdata -D
 1886  1912 17848 PLUGINSD[python ./netdata -D
 1886  1913 17848 PLUGINSD[pcm]   ./netdata -D
 1886  1914 17848 WEB_SERVER[stat ./netdata -D
 1886  1916 17848 WEB_SERVER[stat ./netdata -D
 1886  1917 17848 PLUGINSD[apps]  ./netdata -D
 1886  1919 17848 WEB_SERVER[stat ./netdata -D
 1886  1920 17848 WEB_SERVER[stat ./netdata -D
 1886  1921 17848 PLUGINSD[go.d]  ./netdata -D
 1886  1930 17848 WEB_SERVER[stat ./netdata -D
 1886  1933 17848 STATSD_COLLECTO ./netdata -D
 1886  2405 17848 DBENGINE        ./netdata -D
 1886  2406 17848 DBENGINE        ./netdata -D
 1886  2407 17848 DBENGINE        ./netdata -D
 1886  2409 17848 DBENGINE        ./netdata -D
```

### DIFF
--- a/exporting/init_connectors.c
+++ b/exporting/init_connectors.c
@@ -65,9 +65,9 @@ int init_connectors(struct engine *engine)
 
             // dispatch the instance worker thread
             uv_thread_create(&instance->thread, connector->worker, instance);
-            char threadname[32];
-            snprintfz(threadname, 31, "EXPORTING-%zu", instance->index);
-            uv_thread_set_name(instance->thread, threadname);
+            char threadname[NETDATA_THREAD_NAME_MAX+1];
+            snprintfz(threadname, NETDATA_THREAD_NAME_MAX, "EXPORTING-%zu", instance->index);
+            uv_thread_set_name_np(instance->thread, threadname);
         }
     }
 
--- a/exporting/init_connectors.c
+++ b/exporting/init_connectors.c
@@ -65,6 +65,9 @@ int init_connectors(struct engine *engine)
 
             // dispatch the instance worker thread
             uv_thread_create(&instance->thread, connector->worker, instance);
+            char threadname[32];
+            snprintfz(threadname, 31, "EXPORTING-%zu", instance->index);
+            uv_thread_set_name(instance->thread, threadname);
         }
     }
 
--- a/libnetdata/threads/threads.c
+++ b/libnetdata/threads/threads.c
@@ -109,14 +109,13 @@ static void thread_cleanup(void *ptr) {
     netdata_thread = NULL;
 }
 
-static void thread_set_name(NETDATA_THREAD *nt) {
+static void thread_set_name_np(NETDATA_THREAD *nt) {
 
     if (nt->tag) {
         int ret = 0;
 
-        // Name is limited to 16 chars
-        char threadname[16];
-        strncpyz(threadname, nt->tag, 15);
+        char threadname[NETDATA_THREAD_NAME_MAX+1];
+        strncpyz(threadname, nt->tag, NETDATA_THREAD_NAME_MAX);
 
 #if defined(__FreeBSD__)
         pthread_set_name_np(pthread_self(), threadname);
@@ -134,15 +133,14 @@ static void thread_set_name(NETDATA_THREAD *nt) {
     }
 }
 
-void uv_thread_set_name(uv_thread_t ut, const char* name) {
-    int ret=0;
+void uv_thread_set_name_np(uv_thread_t ut, const char* name) {
+    int ret = 0;
 
-    // Name is limited to 16 chars
-    char threadname[16];
-    strncpyz(threadname, name, 15);
+    char threadname[NETDATA_THREAD_NAME_MAX+1];
+    strncpyz(threadname, name, NETDATA_THREAD_NAME_MAX);
 
 #if defined(__FreeBSD__)
-    ret = pthread_set_name_np(ut, threadname);
+    pthread_set_name_np(ut, threadname);
 #elif defined(__APPLE__)
     // Apple can only set its own name
 #else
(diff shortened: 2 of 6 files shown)


Saruspete:
@mfundul for DBENGINE, I didn't find a way to have a more specific name than "DBENGINE". Do you know how (if) these 4 additionnal threads could be identified (if its worth the work) ?

libnetdata/threads/threads.c
```diff
@@ -134,6 +134,25 @@ static void thread_set_name(NETDATA_THREAD *nt) {
     }
 }
 
+void uv_thread_set_name(uv_thread_t ut, const char* name) {
+    int ret=0;
+
+    // Name is limited to 16 chars
+    char threadname[16];
```
mfundul: It's a good idea to use a `#define` here since we also have the function `thread_set_name()` above.

libnetdata/threads/threads.c
```diff
@@ -134,6 +134,25 @@ static void thread_set_name(NETDATA_THREAD *nt) {
...
+    strncpyz(threadname, name, 15);
+
+#if defined(__FreeBSD__)
+    ret = pthread_set_name_np(ut, threadname);
```
mfundul: This function returns `void`.

libnetdata/threads/threads.c
```diff
@@ -134,6 +134,25 @@ static void thread_set_name(NETDATA_THREAD *nt) {
     }
 }
 
+void uv_thread_set_name(uv_thread_t ut, const char* name) {
```
mfundul: I would like the name to be something like `uv_thread_set_name_np` since it's as non-portable as it gets all the way down.

exporting/init_connectors.c
```diff
@@ -65,6 +65,9 @@ int init_connectors(struct engine *engine)
 
             // dispatch the instance worker thread
             uv_thread_create(&instance->thread, connector->worker, instance);
+            char threadname[32];
+            snprintfz(threadname, 31, "EXPORTING-%zu", instance->index);
```
mfundul: Why 32 characters? another good reason for the name length `#define`.

mfundul:
> 
> 
> @mfundul for DBENGINE, I didn't find a way to have a more specific name than "DBENGINE". Do you know how (if) these 4 additionnal threads could be identified (if its worth the work) ?

I don't think it's worth it for now. We'll see if we can fix it in the future If need be.

Saruspete:
> I don't think it's worth it for now. We'll see if we can fix it in the future If need be.

Ok, it should be enough for now indeed.
Thanks.

libnetdata/threads/threads.c
Saruspete: ok, should I also rename `thread_set_name` to `thread_set_name_np` ?

mfundul: If you like. But hacking libuv this way is much more serious.

libnetdata/threads/threads.c
Saruspete: Done.
`#define NETDATA_THREAD_NAME_MAX 15`

libnetdata/threads/threads.c
Saruspete: Fixed. Miss in my revert.

exporting/init_connectors.c
Saruspete: Done in `threads.h`:
`#define NETDATA_THREAD_NAME_MAX 15`

libnetdata/threads/threads.c
thiagoftsm: Guys, are you sure we need to keep the prefix `uv_` that is used for libuv? This could be a problem case someday libuv decides to bring a function with the same name.

mfundul: This is a hack that may break at any point if a certain libuv version changes the implementation. It's a libuv-specific hack and I think if libuv decides to implement this we should have an `#ifdef` or something with the 2 versions. So, in a way, I think uv has to be part of the name.

thiagoftsm: All right, I will test the PR now.

Saruspete:
@mfundul can you check & validate your change-request to get the merge unblocked please ?

PR #7594, opened 2019-12-20
Restrict quotes in label values

### LINKED ISSUE #7593: Quotes inside label values
##### Feature idea summary
Netdata host labels were restricting values for labels, but it was accepting any kind of value for label values. It is necessary to avoid quotes in label values, because quotes are used to have values with spaces.
##### Expected behavior
Every time that a label has quotes in its value, it will be rejected.

thiagoftsm:
##### Summary
Fixes #7593 

This PR brings restriction to label values, I am also using it to move our utf8 function from url.c to the string library, because now this function is common for more code.

##### Component Name
labels
libnetdata
##### Additional Information
TO test this PR, I defined the following values inside my `[host labels]` section:

```
[host labels]
  label1 = "double"7
  label2 = 'single'
  label3 = Rejeição'
```


docs/configuration-guide.md
```diff
@@ -143,7 +143,8 @@ restrictions are applied for label names:
 - Names cannot start with `_`, but it can be present in other parts of the name.
 - Names only accept alphabet letters, numbers, dots, and dashes.
 
-The policy for values are more flexible, but you can not use exclamation (`!`), whitespaces (` `)  and asterisk (`*`), 
+The policy for values are more flexible, but you can not use exclamation (`!`), whitespaces (` `), single quotes (`'`),
+double quotes (`"`) and asterisk (`*`),
```
joelhans:
```suggestion
double quotes (`"`), and asterisks (`*`), 
```

thiagoftsm: Done thanks!

database/rrdhost.c
```diff
@@ -709,6 +709,19 @@ void rrdhost_save_charts(RRDHOST *host) {
     rrdhost_unlock(host);
 }
 
+static int is_valid_label_value(char *key) {
```
vlvkobal: Why do you use a `key` variable if it is a value?

libnetdata/string/utf8.c
```diff
@@ -0,0 +1,29 @@
...
+ * @param c is the utf8 character
+ *  *
+ * @return It reurns the length of the specific character.
+ */
+char url_utf8_get_byte_length(char c) {
+    if(!IS_UTF8_BYTE(c))
+        return 1;
+
+    char length = 0;
+    while(likely(c & 0x80)) {
```
vlvkobal: Why don't you use `IS_UTF8_BYTE()` here?

database/rrdhost.c
thiagoftsm: I copied and paste from a function, because I had to do only small changes, but I agree that to rename the pointer will let the code more readable.

thiagoftsm: I copied and paste from a function, because I had to do only small changes, but I agree that to rename the pointer will let the code more readable.

libnetdata/string/utf8.c
thiagoftsm: This function was an old contribution that we received, I only moved it from `libnetdata/url/url.c` to here, because it became a common code for other files. I agree with your suggestion and I changed it. Thanks!

ilyam8:
@thiagoftsm why do we call label name a `key`? 🤔

libnetdata/string/utf8.c
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Get byte length
+ *
+ * Measure the utf8 length
+ *
+ * @param c is the utf8 character
+ *  *
+ * @return It reurns the length of the specific character.
+ */
+char url_utf8_get_byte_length(char c) {
```
ilyam8: should we keep `url` prefix?

database/rrdhost.c
```diff
@@ -798,7 +810,7 @@ struct label *load_config_labels()
         struct config_option *cv;
         for(cv = co->values; cv ; cv = cv->next) {
             char *name = cv->name;
-            if(is_valid_label_key(name) && strcmp(name, "from environment") && strcmp(name, "from kubernetes pods") ) {
+            if(is_valid_label_key(name) && is_valid_label_value(cv->value) && strcmp(name, "from environment") && strcmp(name, "from kubernetes pods") ) {
```
ilyam8: shouldnt this be a function? like `is_valid_label_config_option`?

thiagoftsm:
> @thiagoftsm why do we call label name a `key`? thinking

We cannot have two equal names, so they are working like primary key.

libnetdata/string/utf8.c
thiagoftsm: I am keeping the name in this PR, because it is used inside url library and I did not want to changes files that does not have direct relationship with labels, but I agree that it is necessary to have another PR renaming it.

ilyam8:
> so they are working like primary key.

Label has name and value, right? There is no connection with `primary key` imo.

> We cannot have two equal names

Is there the check in the code?

database/rrdhost.c
thiagoftsm: I kept 

> > so they are working like primary key.
> 
> Label has name and value, right? There is no connection with `primary key` imo.

Yes, our documentation says this, the interpretation from my side happens based in the `libnetdata/config/appconfig.c`, I will rename to follow the documentation.

> > We cannot have two equal names
> 
> Is there the check in the code?

Yes we have, when we load configuration using `appconfig` every time that the labels are duplicated, Netdata overwrites the value, so we cannot have two labels with the same name when the label comes from `netdata.conf`, about the other sources(Kubernet and Docker), I cannot say anything.

thiagoftsm:
> > so they are working like primary key.
> 
> Label has name and value, right? There is no connection with `primary key` imo.

Yes, our documentation says this, the interpretation from my side happens based in the `libnetdata/config/appconfig.c`, I will rename to follow the documentation.

> > We cannot have two equal names
> 
> Is there the check in the code?

Yes we have, when we load configuration using `appconfig` every time that the labels are duplicated, Netdata overwrites the value, so we cannot have two labels with the same name when the label comes from `netdata.conf`, about the other sources(Kubernet and Docker), I cannot say anything.

database/rrdhost.c
thiagoftsm: It could be, but we do not have a defined pattern for this, different collectors have `if` with various tests.
I liked your suggestion so I converted it to a `static inline` function.

database/rrdhost.c
```diff
@@ -810,7 +814,7 @@ struct label *load_config_labels()
         struct config_option *cv;
         for(cv = co->values; cv ; cv = cv->next) {
             char *name = cv->name;
-            if(is_valid_label_name(name) && is_valid_label_value(cv->value) && strcmp(name, "from environment") && strcmp(name, "from kubernetes pods") ) {
+            if( is_valid_label_config_option(name, cv->value)) {
```
ilyam8: i think it is ok to pas `cv` to the func

vlvkobal:
After renaming `key` -> `name` there is inconsistency with all the existing code for host labels:
https://github.com/netdata/netdata/blob/master/database/rrd.h#L159-L160

ilyam8:
> After renaming key -> name there is inconsistency with all the existing code for host labels:

worth to rename

> is_valid_label_key(name)
> _key(name)

thiagoftsm:
> worth to rename

I renamed the variable inside the funciton, but I kept the name `_key_`, because this was defined inside the main structure.

ilyam8:
Yes, i know. I think it should be renamed everywhere, because it is label name and not label key.

thiagoftsm:
> Yes, i know. I think it should be renamed everywhere, because it is label name and not label key.

Personally I do not have objection with this, but this is a big change on `host labels`, we can discuss this with everybody and create a new issue specific for this request,

amoss:
Labels are key/values: the identifier is unique and acts as a key to identify the label and access the value. Why are we talking about changing this to name/value? It does not seem to improve the code.

ilyam8:
> Labels are key/values

i was thinking that they are name/values. If label has no name and it is key, then  ,indeed, it should stay as it is.

thiagoftsm:
> I'm confused after reading this PR. The requirement was to reject configuration entries that contain quotes so that they are illegal values for host labels. Maybe I've missed something but the configuration file is not treated as utf-8 is it? I have not seen any scanning code that decodes multi-byte UTF-8 code-points for example.

Firstly I apologize the confusion created. During the first commits of the PR, I remember I was working with the possibility of `netdata.conf` to be an UTF-8 file, but during the development and few talks with other core people, I arrived in the conclusion that it was not necessary to keep this consideration to discard the quotes.

> So then the code in `is_valid_label_value` seems to be all that we need to check for an error. Why are we moving / renaming UTF-8 functions from the url code and making a new string module inside libnetdata?

You are right, we only need the function cited. The previous rename and moving are from the first commits, but this move is also necessary. I can remove these changes from this PR and create another for we have this function available for other parts of the code, it does not make sense we keep an UTF-8 function restricted to URL. @amoss  what is your opinion about this suggestion?

amoss:
I would like to get this merged so that we can complete the host labels support. Please do as you have proposed and strip out the changes to the UTF-8 code so that this PR only does the restriction of the quotes in the label values.

In general it is true that the UTF-8 code would be useful in places other than the URL code. However we don't have any need for it yet, and premature generalization can be be almost as bad as premature optimization. What do you think about holding off on changes to the UTF-8 processing until we have a need to make them?

thiagoftsm:
> I would like to get this merged so that we can complete the host labels support. Please do as you have proposed and strip out the changes to the UTF-8 code so that this PR only does the restriction of the quotes in the label values.

All right, I brought the changes with the last two commits.

> In general it is true that the UTF-8 code would be useful in places other than the URL code. However we don't have any need for it yet, and premature generalization can be be almost as bad as premature optimization. What do you think about holding off on changes to the UTF-8 processing until we have a need to make them?

I saved all necessary changes here with me, when we begin to bring Windows I will create a PR with the codes.

database/rrdhost.c
```diff
@@ -710,6 +710,18 @@ void rrdhost_save_charts(RRDHOST *host) {
     rrdhost_unlock(host);
 }
 
+static int is_valid_label_value(char *value) {
+    while(*value) {
+        if(*value == '"' || *value == '\'') {
```
amoss: This does not match the docs - should exclude `*` and `!` as well.

docs/configuration-guide.md
```diff
@@ -143,7 +143,8 @@ restrictions are applied for label names:
 - Names cannot start with `_`, but it can be present in other parts of the name.
 - Names only accept alphabet letters, numbers, dots, and dashes.
 
-The policy for values are more flexible, but you can not use exclamation (`!`), whitespaces (` `)  and asterisk (`*`), 
+The policy for values are more flexible, but you can not use exclamation (`!`), whitespaces (` `), single quotes (`'`),
```
amoss: "The policy for values is more flexible" and
"exclamation" -> "exclamation marks"

docs/configuration-guide.md
```diff
@@ -143,7 +143,8 @@ restrictions are applied for label names:
...
+double quotes (`"`), and asterisk (`*`),
```
amoss: ", or asterisks"

database/rrdhost.c
thiagoftsm: Thank you, I changed the function!

docs/configuration-guide.md
thiagoftsm: It was changed in the last commit!

docs/configuration-guide.md
thiagoftsm: Thank you very much!

Issue #7585, opened 2019-12-19
Improve the reliability of the shutdown of Netdata in the install and update scripts.


Ferroin:
Currently, the install and upgrade scripts scripts just try to politely shut down any running Netdata instance and don't fall back to taking more aggressive action when that fails. This causes any issues with the shutdown process to be more problematic for many users.

We need to rework how we're handling the shutdown of the agent in the install/upgrade scripts so that it's as reliable as realistically possible.

Related issues: #7248

Issue #7320, opened 2019-11-14
netdata fails to restart after nightly update


mismacku:
##### Kudos
As always, huge thanks for all of the work you put into netdata!  
  
##### Bug report summary
It looks like netdata fails to die during the daily update. I'm running netdata on ~75 servers and this seems to happen only on my low powered Dell R330 servers. (specs below).  Netdata is running directly on the hosts, not in containers or VM's. The problem doesn't occur on every update, but does occur several times per week on ~6 servers. Simply restarting with `systemctl restart netdata` resolves the issue.  
  
netdata-installer.log:  
```[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# tar xf /tmp/netdata-go-YwCYLW/go.d.plugin-v0.10.0.linux-amd64.tar.gz  ... OK
[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# mv go.d.plugin-v0.10.0.linux-amd64 /usr/libexec/netdata/plugins.d/go.d.plugin  ... OK
[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# chown root:netdata /usr/libexec/netdata/plugins.d/go.d.plugin  ... OK
[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# chmod 0750 /usr/libexec/netdata/plugins.d/go.d.plugin  ... OK
[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# cp system/netdata.service /lib/systemd/system/netdata.service  ... OK
[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# systemctl daemon-reload  ... OK
[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# systemctl enable netdata  ... OK
[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# stop_all_netdata  ... [root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# kill 15448  ... OK
[root /tmp/netdata-updater-ontssx/netdata-v1.18.1-28-g9a9bee64]# kill 15448  ... FAILED with exit code 1
OK
[...]

Issue #7248, opened 2019-11-04
Uninstaller does not kill the daemon that is running on FreeBSD


amoss:
##### Bug report summary
The uninstaller script should kill an active daemon process if it is running. In both privileged and non-privileged installs the uninstall script fails to kill the daemon.

##### OS / Environment
FreeBSD bamse 12.0-RELEASE-p6 FreeBSD 12.0-RELEASE-p6 GENERIC  amd64

##### Netdata version (ouput of `netdata -V`)
netdata v1.18.1-47-gcb906c08

##### Component Name
Installer

##### Steps To Reproduce
Install netdata.
Run the uninstaller until it says that netdata has been stopped.
Switch to another shell and run `ps -A` to verify that it is still running.

##### Expected behavior
The netdata process should be terminated.

After some testing with @knatsakis - there is no `/proc` file-system by default in FreeBSD. The uninstaller is being very careful about terminating processes. After the pid-file is read the `/proc/<pid>/stat` is checked to ensure it really is a netdata instance that is running. On FreeBSD we have no pid-file created at run-time, and the proc filesystem is not accessible so no process is killed.

A possible solution is to add a fall-back that detects this situation and asks for user confirmation to run `killall netdata`.

Saruspete:
IIRC, "kill -0" is supported among multiple OS to check if a PID is still existing: https://lists.freebsd.org/pipermail/freebsd-hackers/2017-March/050671.html

Ferroin:
> IIRC, "kill -0" is supported among multiple OS to check if a PID is still existing: https://lists.freebsd.org/pipermail/freebsd-hackers/2017-March/050671.html

While true, I'm not sure that's especially useful here. The issue is that we try to verify that a given PID is netdata before we try to kill it, but that's failing on FreeBSD due to the combined lack of `/proc` and a PID file. The problem should be trivial to solve, because the logic for updates/upgrades works correctly on FreeBSD and terminates Netdata, we're just using different code.

Regardless, I plan to deal with this together with #7585.

PR #7595, opened 2019-12-20
Update handling of shutdown of the Netdata agent on update and uninstall.

Ferroin:
##### Summary

This improves the logic in the update and uninstall scripts used for kickstart installations so that they more reliably stop the Netdata agent process. This will help mitigate the impact of bugs in the shutdown code for the agent itself.

##### Component Name

area/packaging

##### Additional Information

Fixes: #7585 
Fixes: #7248 
Fixes: #7320

---

The new logic is intended to work like so:

1. If we can find `netdatacil` in `$PATH` and there appears to be an agent running, use that to try and shutdown the agent, then wait for 20 seconds before proceeding.
2. For each live Netdata PID, spend up to 60 seconds trying to stop it with SIGTERM, and then up to another 60 seconds trying with SIGKILL if SIGTERM fails.

This can technically still leave a live agent process running, though it is very unlikely. I plan to update the installer and updater code in a future PR to log an error and not start Netdata if they see what appears to be a live agent still running.

---

As of right now this appears to work based on running the following tests on Linux (in Docker containers):

* Make a clean install of an older version of Netdata (specifically a nightly from 2019-12-19), then update using this code, followed by uninstalling using this code.
* Same as above, but with `netdatacli` manually removed from `$PATH`.
* Same as the above, but with GDB used to manipulate the signal mask so that the agent was ignoring SIGTERM.

---

I've now run equivalent tests to the above on FreeBSD as well as on more Linux distributions, and am reasonably confident that this works correctly.

### DIFF
--- a/packaging/installer/functions.sh
+++ b/packaging/installer/functions.sh
@@ -424,7 +424,7 @@ install_netdata_service() {
 pidisnetdata() {
 	if [ -d /proc/self ]; then
 		[ -z "$1" -o ! -f "/proc/$1/stat" ] && return 1
-		[ "$(cat "/proc/$1/stat" | cut -d '(' -f 2 | cut -d ')' -f 1)" = "netdata" ] && return 0
+		[ "$(cut -d '(' -f 2 "/proc/$1/stat" | cut -d ')' -f 1)" = "netdata" ] && return 0
 		return 1
 	fi
 	return 0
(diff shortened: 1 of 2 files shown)


ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/netdata/netdata/pullRequest?prid=4715579&bid=15747602)

prologic:
I haven't run through your testing steps; but the code looks good.

> Same as the above, but with GDB used to manipulate the signal mask so that the agent was ignoring SIGTERM.

This is what I am worried about; that there _might_ be some corner cases where the agent is either ignoring signals or waiting on other pids that do not terminate.

PR #7600, opened 2019-12-21
Adjust alarm labels

### LINKED ISSUE #7599: Rename alarm `label`
##### Feature idea summary
Netdata will have other group of labels in a near future, it will be good to have different `label` lines, to avoid mix the labels. And also it is necessary to do:

- "label" should be called "host labels"
-   we should use the tokenizer from plugins.d that respects quotes and returns array of pointers: https://github.com/netdata/netdata/blob/master/collectors/plugins.d/plugins_d.c#L37

To bring this last we will need to change libnetdaa.
##### Expected behavior
To have `host labels` instead `label` on alarms.

thiagoftsm:
##### Summary
Fixes #7599 

This PR brings the necessary changes for Netdata to have `host labels` instead `label` parsed on alarms.
##### Component Name
Health
##### Additional Information
During the development, I created few labels inside my netdata.conf in the new setion `[host labels]`, after this I ran `cp tests/template_dimension/system_cpu.conf.unique_alarm /etc/netdata/health.d/system_cpu.conf` and I inserted the labels for each alarm, the labels were defined following these steps:

- All alarms with the same label defined using `host labels:`
- Different labels for each alarm, here I also used valid labels.
- Change the labels to invalid values inside the alarm, but I kept one with correct values.

### DIFF
--- a/health/REFERENCE.md
+++ b/health/REFERENCE.md
@@ -394,21 +394,21 @@ By defining labels inside of `netdata.conf`, you can now apply labels to alarms.
 line to any alarms you'd like to apply to hosts that have the label `room = server`.
 
 ```yaml
-host lables: room = server
+host labels: room = server
 ```
 
 You can also combine labels when applying them to alarms. For example, if you want to raise a specific alarm only for hosts 
 inside a room that were installed at a specific time, you can write the following label line:
 
 ```yaml
-host lables: room = workstation AND installed = 201705
+host labels: room = workstation AND installed = 201705
 ```
 
 The `label` is a space-separated list that accepts simple patterns. For example, you can create an alarm 
 that will be applied to all hosts installed in the last decade with the following line:
 
 ```yaml
-host lables: installed = 201*
+host labels: installed = 201*
 ```
 
 See our [simple patterns docs](../libnetdata/simple_pattern/) for more examples.
--- a/health/REFERENCE.md
+++ b/health/REFERENCE.md
@@ -68,7 +68,7 @@ Netdata parses the following lines. Beneath the table is an in-depth explanation
 | [`delay`](#alarm-line-delay)                        | no              | Optional hysteresis settings to prevent floods of notifications.                      |
 | [`repeat`](#alarm-line-repeat)                      | no              | The interval for sending notifications when an alarm is in WARNING or CRITICAL mode.  |
 | [`option`](#alarm-line-option)                      | no              | Add an option to not clear alarms.                                                    |
-| [`host labels`](#alarm-line-host-label)                   | no              | List of labels present on a host.                                                     |
+| [`host labels`](#alarm-line-host-label)             | no              | List of labels present on a host.                                                     |
 
 The `alarm` or `template` line must be the first line of any entity.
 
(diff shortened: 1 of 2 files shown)


health/REFERENCE.md
```diff
@@ -394,21 +394,21 @@ By defining labels inside of `netdata.conf`, you can now apply labels to alarms.
 line to any alarms you'd like to apply to hosts that have the label `room = server`.
 
 ```yaml
-label: room = server
+host lables: room = server
```
vlvkobal: lables -> labels

health/REFERENCE.md
```diff
@@ -394,21 +394,21 @@ By defining labels inside of `netdata.conf`, you can now apply labels to alarms.
...
 ```
 
 You can also combine labels when applying them to alarms. For example, if you want to raise a specific alarm only for hosts 
 inside a room that were installed at a specific time, you can write the following label line:
 
 ```yaml
-label: room = workstation AND installed = 201705
+host lables: room = workstation AND installed = 201705
```
vlvkobal: Simple patterns don't support the `AND` operator.

lables -> labels.

health/REFERENCE.md
```diff
@@ -394,21 +394,21 @@ By defining labels inside of `netdata.conf`, you can now apply labels to alarms.
 line to any alarms you'd like to apply to hosts that have the label `room = server`.
 
 ```yaml
-label: room = server
+host lables: room = server
 ```
 
 You can also combine labels when applying them to alarms. For example, if you want to raise a specific alarm only for hosts 
 inside a room that were installed at a specific time, you can write the following label line:
```
vlvkobal: You can't do that with a simple pattern.

label -> host labels

health/REFERENCE.md
```diff
@@ -394,21 +394,21 @@ By defining labels inside of `netdata.conf`, you can now apply labels to alarms.
...
 
 ```yaml
-label: room = workstation AND installed = 201705
+host lables: room = workstation AND installed = 201705
 ```
 
 The `label` is a space-separated list that accepts simple patterns. For example, you can create an alarm
```
vlvkobal: label -> host labels

health/REFERENCE.md
```diff
@@ -394,21 +394,21 @@ By defining labels inside of `netdata.conf`, you can now apply labels to alarms.
...
-label: room = workstation AND installed = 201705
+host lables: room = workstation AND installed = 201705
 ```
 
 The `label` is a space-separated list that accepts simple patterns. For example, you can create an alarm 
 that will be applied to all hosts installed in the last decade with the following line:
 
 ```yaml
-label: installed = 201*
+host lables: installed = 201*
```
vlvkobal: lables -> labels

health/REFERENCE.md
thiagoftsm: Thank you!

health/REFERENCE.md
thiagoftsm: I used `:% s/label:/host lables:/`, so the error was replicated, thank you very much!

health/REFERENCE.md
vlvkobal: You forgot about simple patterns.

thiagoftsm: I am doing references to simple pattern on line 411, they idea behind these changes on our manual was to increase complexity until the last example, this is the motive it was not cited before. I also wanted to avoid repetition of ideas.

health/REFERENCE.md
thiagoftsm: I changed it with the latest commit, thanks!

ilyam8:
> Did you remove the ability to use AND

i assumed all logic (AND, OR, etc) never worked, `host labels` is just a simple pattern.

thiagoftsm:
> > Did you remove the ability to use AND
> 
> i assumed all logic (AND, OR, etc) never worked, `host labels` is just a simple pattern.

For we bring this we will need to change the whole logic behind `simple pattern', I think this will be another PR.

PR #7628, opened 2019-12-26
Remove EOL distros from CI jobs.

Ferroin:
##### Summary

This removes Fedora 28 (which went EOL 2019-05-28) and Ubuntu 18.10 (which went EOL 2019-07-18) from the lifecycle checks and packaging process. Both have been EOL for more than 3 months, are not supported by the upstream developers, and are also not supported by their respective communities either.

##### Component Name

area/ci
area/packaging

##### Additional Information

This is intended as a proactive change to mitigate against future failed builds that we will likely not care about at that point. It also helps speed up both nightlies and PR checks.


knatsakis:
@Ferroin, I think that also the [docs](https://docs.netdata.cloud/packaging/distributions) should be slightly updated regarding Fedora 28

Ferroin:
> @Ferroin maybe you could remove Fedora 28 from [here](https://github.com/netdata/helper-images/blob/master/oses/Dockerfile.fedora28) as well?

netdata/helper-images#26

Issue #7627, opened 2019-12-26
Update issue and PR labeling to use GitHub Actions


Ferroin:
This is a long-term planned change to decouple of labeling process from our CI process, which will make a number of aspsects of managing both somewhat easier.

Assuming we don't want to write custom actions to handle this, [this](https://github.com/actions/labeler) seems to be our best option.

PR #7630, opened 2019-12-26
Switch PR labeling to use GitHub Actions.

Ferroin:
##### Summary

This updates the PR labeling functionality to use GitHub Actions instead of being a recurring job in Travis.

Aside from the obvious change of needing to check somewhere other than Travis to verify this is working, this results in the following changes to how automatic labeling works for PRs:

* The automatic labeling now gets run whenever a PR is opened, reopened, updated, or marked ready for review instead of running whenver a commit is pushed to `master` or the nightlies get run. This should improve responsiveness for users.
* Labels will not be automatically removed anymore. This is a limitation of the action being used, but is one I feel is actually a good thing. Specific sub-sections of existing files may be relevant to areas other than those that the file itself is part of, and the auto-removal of labels with the current solution makes it hard to express this properly.
* A couple of labels listed in the existing labeler script have been updated to more correctly reflect the current state of affairs (especially the packaging and tests labels).
* The selection of labels which may be automatically applied has been expanded to include every `area/*` label that has a corresponding set of files in the repo.

##### Component Name

area/ci

##### Additional Information

Fixes: #7627
Partially mitigates the issue solved by: #7543 

Minimally tested with equivalent code in a throwaway repository.


Ferroin:
I've added the core team for review on this as I'd like their input to make sure the label mappings in `.github/labeler.yml` are correct.

Please don't merge until this gets a review from them.

.github/labeler.yml
```diff
@@ -0,0 +1,91 @@
...
+area/external:
+  - collectors/charts.d.plugin/*
+  - collectors/charts.d.plugin/**/*
+  - collectors/node.d.plugin/*
+  - collectors/node.d.plugin/**/*
+
+area/health:
+  - health/*
+  - health/**/*
+
```
thiagoftsm: I think we are missing `area/libnetdata` , because sometimes it is necessary to append functions inside it.

Ferroin: This PR isn't to change what labels we have, just how they get applied to PR's (there actually is a GitHub Action available for managing what labels a repo has as part of the repo itself, but that's a bit out of scope for this PR IMO).

However, I do think that it would be a good idea to add such a label, as nothing else we have currently covers that part of the code. If someone on the core team wants to add it, I'll be happy to update the PR to include a labeling rule for it.

thiagoftsm: All right, I confess I do not see any motive for we do not have labels for `libnetdata`. It is expected that we will have big changes on it next year when we begin to bring support to Microsoft.

Ferroin:
@cosmix @cakrit Could one of you two double check that the label mappings in `.travis/labeler.yml` are correct in this PR? The format is YAML, with keys corresponding to labels and values being a list of regexes for paths to apply those labels for.

ilyam8:
PR name

> Switch PR labeling to use GitHub Actions

Comment from [deleted file](https://github.com/netdata/netdata/blob/master/.travis/labeler.sh)

> This is a simple script which should apply labels to unlabelled issues from last 3 days.

> should apply labels to unlabelled issues

I guess the title is wrong @Ferroin ?

Ferroin:
> PR name
> 
> > Switch PR labeling to use GitHub Actions
> 
> Comment from [deleted file](https://github.com/netdata/netdata/blob/master/.travis/labeler.sh)
> 
> > This is a simple script which should apply labels to unlabelled issues from last 3 days.
> 
> > should apply labels to unlabelled issues
> 
> I guess the title is wrong @Ferroin ? This PRs switches both PR and issues labeling?

The comments are wrong, the script being removed hasn't done labeling of issues since we started using Issue Templates, the comments never got updated for some reason.

Ferroin:
Merging this. If the labels are wrong, we can just fix it later.

Issue #7332, opened 2019-11-18
Fix permissions in accordance with #6619 /usr/libexec/netdata/plugins.d


knatsakis:
##### Bug report summary
Fix permissions in accordance with #6619. This issue is for the /usr/libexec/netdata/plugins.d directories and files

Issue #7301, opened 2019-11-11
Fix permissions in accordance with #6619 /usr/sbin/netdata	


knatsakis:
##### Bug report summary
Fix permissions in accordance with #6619. This issue is for /usr/sbin/netdata

Issue #7300, opened 2019-11-11
Fix permissions in accordance with #6619 /usr/libexec/netdata


knatsakis:
##### Bug report summary
Fix permissions in accordance with #6619. This issue is for the /usr/libexec/netdata directories and files except files in /usr/libexec/netdata/plugins.d

Issue #7299, opened 2019-11-11
Fix permissions in accordance with #6619 /usr/lib64/netdata


knatsakis:
##### Bug report summary
Fix permissions in accordance with #6619. This issue is for the /usr/lib64/netdata directories and files

PR #7632, opened 2019-12-26
Fix install permissions

Ferroin:
##### Summary

This updates the handling of permissions in the installer script and the RPM spec file so that theya re consistent with each other and with a clean install done with `make install`.

This fixes most of the problems pointed out in #6619 other than not using capabilities consistently in places we could use them, which is being tracked by a separate issue (#7631).

##### Component Name

area/packaging

##### Additional Information

Fixes: #6619
Fixes: #7298 
Fixes: #7299 
Fixes: #7300 
Fixes: #7301 
Fixes: #7332

This has been minimally tested in Docker containers for a handful of distributions. I will remove the [WIP] when I've finished more extensive testing, but it is functionally ready for review right now.

A sizable percentage of this involves just trusting `make install` to get the permissions right, which it seems to in all cases.


Ferroin:
After further testing, this appears to be working correctly, removing the WIP.

PR #7636, opened 2019-12-27
Update the distribution support matrix to represent reality.

Ferroin:
##### Summary

This makes the following changes:

* Remove Debian 7: It's been EOL for more than 1.5 years now, anybody using it should not be expecting support anywhere, so it should fall under 'All other linux'.
* Update 'All other Linux': to show that we aren't doing CI testing at all for anything we don't have listed.
* Update CentOS 6 and RHEL 6 to reflect the lack of CI/CD support: We aren't doing anything with these systems in CI/CD, so show that correctly instead of being 'unknown'.
* Mark CI testing on 32-bit x86 as not happening: We're not doing _any_ CI testing on 32-bit x86, but the matrix claims we're doing CI smoke testing.

##### Component Name

area/docs


Ferroin:
Rebased to resolve merge conflict.

joelhans:
How might this relate, if at all, to #7432/#7434?

It seems like both PRs are working at, more or less, the same underlying issue, which is the complexity of the matrix and whether something is unknown or negative. I may be wrong though!

Ferroin:
> How might this relate, if at all, to #7432/#7434?
> 
> It seems like both PRs are working at, more or less, the same underlying issue, which is the complexity of the matrix and whether something is unknown or negative. I may be wrong though!

I view this one as mostly orthogonal to both of those PR's. The primary goal here is to accurately reflect that:

* We currently provide essentially zero support for Debian 7 since it's been completely EOL for more than a year and a half.
* We no longer build packages for RHEL 6 or CentOS 6
* We do no CI related stuff for any Linux distros we don't have listed.
* We currently do no 32-bit testing in CI (I plan to try and get this changed, but for now the docs should reflect reality).
* We do not build 32-bit RPM packages for any distro (nor should we, as 32-bit is functionally dead when dealing with RPM distros).

joelhans:
> I view this one as mostly orthogonal to both of those PR's. The primary goal here is to accurately reflect that:

Thanks for the info! Just wanted to double-check that there isn't a bunch of conflicting ideas/edits going into this one already-complex document.

I have a few ideas about how to fix some of the issues around the distribution matrix, but they're also outside of the scope of what you're doing here. Maybe we can connect in Athens about it!

Ferroin:
> I have a few ideas about how to fix some of the issues around the distribution matrix, but they're also outside of the scope of what you're doing here. Maybe we can connect in Athens about it

I'd love to chat about this in Athens, I've also got a few thoughts on how we could make it both better for users and possibly also a bit easier to maintain.

PR #7638, opened 2019-12-27
minor: docu/css - remove anoying underlines

underhood:
##### Component Name
Documentation remove underlines from under the badges. (https://docs.netdata.cloud/web/api/badges/)

##### Additional Information
Before/Current:
before

After/This PR:
After


underhood:
Became obsolete with redesign of docu site with #7753

Issue #7644, opened 2019-12-30
cannot set libuv thread name to XXXX. Err: 13 (error 13, Permission denied)


startryst:
##### Bug report summary
```
2019-12-31 03:22:54: netdata INFO  : MAIN : All threads finished.
2019-12-31 03:22:54: netdata INFO  : MAIN : EXIT: freeing database memory...
2019-12-31 03:22:54: netdata INFO  : MAIN : Freeing all memory for host 'alysz2'...
2019-12-31 03:22:54: netdata INFO  : MAIN : Shutting down RRD engine event loop.
2019-12-31 03:22:54: netdata INFO  : MAIN : Shutting down RRD engine event loop complete.
2019-12-31 03:22:54: netdata INFO  : MAIN : Freed 81511648 bytes of memory from page cache.
2019-12-31 03:22:54: netdata INFO  : MAIN : SYSTEM_INFO: free 0x1d18660
2019-12-31 03:22:54: netdata INFO  : MAIN : EXIT: removing netdata PID file '/var/run/netdata/netdata.pid'...
2019-12-31 03:22:54: netdata ERROR : MAIN : EXIT: cannot unlink pidfile '/var/run/netdata/netdata.pid'. (errno 13, Permission denied)
2019-12-31 03:22:54: netdata INFO  : MAIN : EXIT: all done - netdata is now exiting - bye bye...
```

```
2019-12-31 03:22:56: netdata INFO  : MAIN : Loading journal file "/var/cache/netdata/dbengine/journalfile-1-0000000002.njf".
2019-12-31 03:22:56: netdata INFO  : MAIN : Journal file "/var/cache/netdata/dbengine/journalfile-1-0000000002.njf" loaded (size:2502656).
2019-12-31 03:22:56: netdata ERROR : MAIN : cannot set libuv thread name to DBENGINE. Err: 13 (errno 13, Permission denied)
```

```
2019-12-31 03:22:56: netdata ERROR : MAIN : cannot set libuv thread name to DAEMON_COMMAND. Err: 13 (errno 13, Permission denied)
2019-12-31 03:22:56: netdata INFO  : MAIN : netdata initialization completed. Enjoy real-time performance monitoring!
2019-12-31 03:22:56: netdata INFO  : MAIN : /usr/libexec/netdata/plugins.d/anonymous-statistics.sh 'START' '-' '-'
```

##### OS / Environment
CentOS release 6.9 (Final)

##### Netdata version (ouput of `netdata -V`)
[...]

PR #7653, opened 2020-01-02
[libnetdata/threads] Change log level on error

Saruspete:
##### Summary
Reduce log for `uv_thread_set_name_np` from `error` to `info`.

Thread renaming is not a requirement, but ease bug reporting. On bug #7644 for an unknown reason, libuv could not rename its threads on CentOS6. So the logs showed an error log that was reported.

##### Component Name
- libnetdata/threads/thread.c


ilyam8:
thanks @Saruspete

PR #7640, opened 2019-12-30
Properly check all shell scripts.

Ferroin:
##### Summary

This updates the shell script checking process to check all of our shell scripts, instead of just the ones with `.sh.in` file extensions.

##### Component Name

area/ci

##### Additional Information

Fixes: #7100 

Instead of trying to do convoluted things with the `find` command's `-name` option, I opted to just switch to using the `-regex` option to do the file matching. This is a bit slower, but a whole lot more concise when it comes to identifying what files we need to check.

This results in a lot of 'new' warnings, which I also intend to fix with this PR. I've not looked too deep into them yet though, and will push new commits onto the PR as I do so. I'll mark this ready for review once I've got all the 'new' warnings fixed.

PR #7662, opened 2020-01-02
Fix buildyaml.sh script so that docs generation works correctly.

Ferroin:
##### Summary

Based on https://github.com/netdata/netdata/pull/7640#discussion_r362519469.

##### Component Name

area/docs

### DIFF
--- a/docs/generator/buildyaml.sh
+++ b/docs/generator/buildyaml.sh
@@ -37,7 +37,7 @@ navpart() {
 	if [ -z "$maxdepth" ]; then maxdepth=1; fi
 	if [[ -n $excludefirstlevel ]]; then mindepth=2; else mindepth=1; fi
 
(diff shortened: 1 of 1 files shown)


docs/generator/buildyaml.sh
```diff
@@ -37,7 +37,7 @@ navpart() {
 	if [ -z "$maxdepth" ]; then maxdepth=1; fi
 	if [[ -n $excludefirstlevel ]]; then mindepth=2; else mindepth=1; fi
 
-	for f in $(find $dir -mindepth $mindepth -maxdepth $maxdepth -name "${file}.md" -printf '%h\0%d\0%p\n' | sort -t '\0' -n | awk -F '\0' '{print $3}'); do
+	for f in $(find $dir -mindepth $mindepth -maxdepth $maxdepth -name "${file}.md" -printf '%h|%d|%p\n' | sort -t '|' -n | awk -F '|' '{print $3}'); do
```
prologic: I _think_ I know why @cosmix added these `\0`(s) -- It has to do with looping constructs; my earlier comment on using the "while read" form should be used here; But I'll approve this change anyway since AFAIK `for x in y; do` in Sh can't operate on inputs separated by null characters.

PR #7663, opened 2020-01-02
error exit when rrdhost localhost init fails #7504

### LINKED ISSUE #7504: Segfault if `localhost` cannot be initialized in `rrd_init()`
##### Bug report summary
On daemon startup, `rrd_init()` calls `rrdhost_create()` to populate the global `localhost` variable. If this call to `rrdhost_create()` fails, `localhost` is set to `NULL`. Other parts of the code [don't expect `localhost` to be `NULL`][1], leading to segmentation faults.

`rrdhost_create()` can for example fail if the database files cannot be written and the DB engine cannot be initialized. (initial error: *netdata ERROR : MAIN : Failed to open file "/var/cache/netdata/dbengine/datafile-1-0000000034.ndf". (errno 13, Permission denied)*).  I haven't checked if there are other error cases where `rrdhost_create()` might fail for different reasons.

Maybe it would be reasonable to simply print an error message and exit if basic initialization like that of `localhost` fails.

##### OS / Environment
`Linux version 4.19.0-6-amd64 (debian-kernel@lists.debian.org) (gcc version 8.3.0 (Debian 8.3.0-6)) #1 SMP Debian 4.19.67-2+deb10u2 (2019-11-11)`

netdata v1.19.0

##### Component Name
daemon

##### Steps To Reproduce
```
chmod a-w /var/cache/netdata/dbengine/*
netdata -D
```

##### Expected behavior
Program exit with return code != 0.

 [1]: https://github.com/netdata/netdata/blob/92b58d556e2e3a0bc1e42a7c2d59dc855c9a1eee/backends/backends.c#L509

underhood:
##### Summary
Fixes #7504, When rrd_init fails for localhost exit with error code

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -941,8 +941,10 @@ int main(int argc, char **argv) {
                             default_rrd_update_every = 1;
                             default_rrd_memory_mode = RRD_MEMORY_MODE_RAM;
                             default_health_enabled = 0;
-                            if(!rrd_init("unittest", NULL))
-                                fatal("rrd_init failed for unittest");
+                            if(!rrd_init("unittest", NULL)) {
+                                fprintf(stderr, "rrd_init failed for unittest\n");
+                                return 1;
+                            }
                             default_rrdpush_enabled = 0;
                             if(run_all_mockup_tests()) return 1;
                             if(unit_test_storage()) return 1;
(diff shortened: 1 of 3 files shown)


daemon/main.c
```diff
@@ -941,7 +941,8 @@ int main(int argc, char **argv) {
                             default_rrd_update_every = 1;
                             default_rrd_memory_mode = RRD_MEMORY_MODE_RAM;
                             default_health_enabled = 0;
-                            rrd_init("unittest", NULL);
+                            if(!rrd_init("unittest", NULL))
+                                fatal("rrd_init failed for unittest");
```
mfundul: You should use `fprintf(stderr, ` here since the log service hasn't been initialized yet.

PR #7688, opened 2020-01-08
Adds a Dockerfile.docs for more easily and reproducibly building/rebuilding docs

prologic:
This is part of a larger body of work related to CI/CD in general where we have decided to use [Docker](https://www.docker.com) to facilitate easier packaging, builds and reproducibility. This PR introduces a new `Dockerfile` specifically for building our documentation.

The intention is to publish this image to our [NetData Docker Hub Org](https://hub.docker.com/netdata) so other maintainers/contributors can also more easily and quickly get going on contributing to our docs.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -39,6 +39,7 @@ web/gui/ @jacekkolasa @cosmix
 # Ownership by filetype (overwrites ownership by directory)
 *.md @cosmix @joelhans
 *.am @cosmix @Ferroin @knatsakis @ncmans @prologic
+Dockerfile.* @Ferroin @knatsakis @ncmans @prologic @cosmix
 
 # Ownership of specific files
 .gitignore @cosmix @Ferroin @knatsakis @ncmans @prologic
@@ -49,6 +50,7 @@ web/gui/ @jacekkolasa @cosmix
 .csslintrc @cosmix @Ferroin @knatsakis @ncmans @prologic
 .codeclimate.yml @cosmix @Ferroin @knatsakis @ncmans @prologic
 .codacy.yml @cosmix @Ferroin @knatsakis @ncmans @prologic
+Dockerfile @Ferroin @knatsakis @ncmans @prologic @cosmix
 netdata.spec.in @cosmix @Ferroin @knatsakis @ncmans @prologic
 netdata-installer.sh @cosmix @Ferroin @knatsakis @ncmans @prologic
 netlify.toml @cosmix
--- a/Dockerfile.docs
+++ b/Dockerfile.docs
@@ -1,0 +1,26 @@
+ARG PYVER=3.8
+
+FROM python:${PYVER}-alpine
+
+WORKDIR /netdata
+
+# These scripts use Bash(ism) so install Bash
+# TODO: Maybe rewrite the link checker in something more sane
+RUN apk add --no-cache -U bash
+
+# The scripts also clone the netdata/netdata repo using git
+# TODO: Maybe also optionally support bind-mounted sources
+RUN apk add --no-cache -U git
+
+# The scripts also use GNU find options
+RUN apk add --no-cache -U findutils
+
+# Copy and Install build dependencies first to cache them so we don't have to
+# do this every single time we want to rebuild the docs. The cache is busted
+# when/if the SHA of the requirements.txt is changed.
+COPY docs/generator /netdata/docs/generator
+RUN pip install -r /netdata/docs/generator/requirements.txt
+
+COPY . .
+
+CMD ["/netdata/docs/generator/buildhtml.sh"]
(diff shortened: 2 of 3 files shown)


prologic:
This has been tested with @joelhans in-person along with verified to be working from Linux (@Ferroin) and macOS (@jacekkolasa @prologic). This greatly simplifies the process for rebuilding the docs for maintainers/contributors, but we also preserve the existing instructions.

Dockerfile.docs
```diff
@@ -0,0 +1,25 @@
+# Using Python 2.7 here because mkdocs>=1.0.1 is broken on Python 3.3, 3.4, 3.7 and 3.8 :/
```
cosmix: What about 3.6? If it works on 3.6, please use that instead of 2.7.

.github/CODEOWNERS
```diff
@@ -39,6 +39,7 @@ web/gui/ @jacekkolasa @cosmix
 # Ownership by filetype (overwrites ownership by directory)
 *.md @cosmix @joelhans
 *.am @cosmix @Ferroin @knatsakis @ncmans @prologic
+Dockerfile.* @Ferroin @knatsakis @ncmans @prologic
```
cosmix: Please add me as code owner wherever you made changes.

Dockerfile.docs
prologic: Sorry I tried this too! I've amended the comment to reflect this :) Only 2.7 works :/ I would love to help the upstream project fix the bugs I found but not now; maybe later :)

.github/CODEOWNERS
prologic: Done

Dockerfile.docs
cosmix: Is this reflected anywhere on their docs? All I saw on their site is adding support for 3.7 in 1.0. I don’t think it’s worth contributing to mkdocs at this time. We have been discussing moving away from it for the past several months. So 2.7 is ok for now, but not a good long term decision, as Python 2.x is EOL. Let’s sync with @joelhans to determine a good long-term course of action here.

prologic: I agree! FWIW I _think_ `mkdocs`' docs (_hmmm_) is a bit misleading somewhat and/or not as well maintained as I'd like :) It was quite painful to get this working on anything but 2.7 which is a shame. If we stick with static site generators (_which I think we should_) I can highly recommend [Hugo](https://gohugo.io/) here; but I do not know how well it will fit specifically in our use case for how we've llaid out our docs and presentation.

Dockerfile.docs
```diff
@@ -0,0 +1,25 @@
+# NOTE: `mkdocs` only seems to work on either Python 2.7 or 3.6 (despite upstream docs claiming support for other 3.x versions)
+FROM python:3.6-alpine
```
prologic: @cosmix I'm sorry; I was looking at some netlify errors today and noticed we are inf act using Python 3.6 -- So I re-tested and it works fine. I must have thought I tried 3.6 and ran into bugs too but I am mistaken. mkdocs only seems to work on 2.7 and 3.6 (despite their website and pypi claminig other versions). I will fix/ammend this comment / FROM line :)

prologic: And... Done.

cosmix: Hold on. They claim comparability with 3.7 too. Can you check it please?

prologic: I can and will comment with the results.

prologic: I re-tested a bunch of different Python versions and here are the resutls:

```bash
prologic@Jamess-iMac
Tue Jan 14 07:28:07
~/NetData/netdata
 (mkdocs_docker_image) 0
$ ./test.sh
Building netdata/docs:py3.3 ... ERR
Running netdata/docs:py3.3 ... ERR
Building netdata/docs:py3.4 ... ERR
Running netdata/docs:py3.4 ... ERR
Building netdata/docs:py3.5 ... OK
Running netdata/docs:py3.5 ... OK
Building netdata/docs:py3.6 ... OK
Running netdata/docs:py3.6 ... OK
Building netdata/docs:py3.7 ... OK
Running netdata/docs:py3.7 ... OK
Building netdata/docs:py3.8 ... OK
Running netdata/docs:py3.8 ... OK
```

I can't explain some bugs / Tracebacks I was seeing at the meetup; but *meh* I'll amend with 3.8 and move on :)

prologic: Script if you want to verify/confirm:

```bash
#!/bin/sh

for PYVER in 3.3 3.4 3.5 3.6 3.7 3.8; do
  printf "Building netdata/docs:py%s ... " "$PYVER"
  if docker build -q --build-arg PYVER="$PYVER" -f Dockerfile.docs -t netdata/docs:py"$PYVER" . > "${PYVER}".build.log 2>&1; then
    printf "OK\n"
  else
    printf "ERR\n"
  fi

  printf "Running netdata/docs:py%s ... " "$PYVER"
  if docker run -i -t --rm -v "$PWD":/netdata netdata/docs:py"$PYVER" > "${PYVER}".log 2>&1; then
    printf "OK\n"
  else
    printf "ERR\n"
  fi
done
```

prologic:
Added a [netdata/docs](https://hub.docker.com/r/netdata/docs) Docker Hub build trigger for this.

PR #7694, opened 2020-01-09
Fixes #7680 Remote write

### LINKED ISSUE #7680: [remote write] raw value were used in data source average/sum
##### Bug report summary

Prometheus remote write always use raw value regardless the data source type.

```
  add_metric(name, chart, family, dimension, hostname, rd->last_collected_value, timeval_msec(&rd->last_collected_time));
```

Ubuntu 18.04

v1.19.0

##### Component Name

Backend

##### Steps To Reproduce

Use `average` in backend with remote write, get the same value as of `raw` data source

##### Expected behavior

the calculated value should be used instead of `rd->last_collected_value`

Ehekatl:
##### Summary

Fixes #7680 

use calculated value instead of raw value

##### Component Name

Backend remote write


thiagoftsm:
Thank you @Ehekatl !

PR #7695, opened 2020-01-09
Enhance prometheus remote write

### LINKED ISSUE #7676: backend http request need proper Host header
##### Bug report summary

When using prom remote write, it use hostname as backend Host header

```
                    buffer_sprintf(http_request_header,
                                    "POST %s HTTP/1.1\r\n"
                                    "Host: %s\r\n"
                                    "Accept: */*\r\n"
                                    "Content-Length: %zu\r\n"
                                    "Content-Type: application/x-www-form-urlencoded\r\n\r\n",
                                    remote_write_path,
                                    hostname,
                                    data_size
                    );
```

it should use destination's hostname for proper routing the requests

Ubuntu 18.04

v1.19.0

##### Component Name
netdata backend

##### Steps To Reproduce

use prom remote write with domain as destination

```
[backend]
  enabled = yes
  type = prometheus_remote_write
  data source = as collected
  destination = DOMAIN
```

##### Expected behavior

Backend follow HTTP standard use destination's domain in Host header

### LINKED ISSUE #7678: Support host variables in prometheus remote write
##### Feature idea summary

Currently remote write will not send host variables

##### Expected behavior

It behave the same as in `/api/v1/allmetrics`

Ehekatl:
##### Summary

- Fixes #7676  support specify destination hostname in remote backend
- Fixes #7678 Support send host/chart variables in remote backend
- Support filter host/chart variables with `send charts matching`

##### Component Name

Backend

### DIFF
--- a/backends/backends.c
+++ b/backends/backends.c
@@ -507,6 +507,7 @@ void *backends_main(void *ptr) {
     const char *destination     = config_get(CONFIG_SECTION_BACKEND, "destination", "localhost");
     global_backend_prefix       = config_get(CONFIG_SECTION_BACKEND, "prefix", "netdata");
     const char *hostname        = config_get(CONFIG_SECTION_BACKEND, "hostname", localhost->hostname);
+    const char *remote_hostname = config_get(CONFIG_SECTION_BACKEND, "remote hostname", "localhost");
     global_backend_update_every = (int)config_get_number(CONFIG_SECTION_BACKEND, "update every", global_backend_update_every);
     int buffer_on_failures      = (int)config_get_number(CONFIG_SECTION_BACKEND, "buffer on failures", 10);
     long timeoutms              = config_get_number(CONFIG_SECTION_BACKEND, "timeout ms", global_backend_update_every * 2 * 1000);
@@ -1069,7 +1070,7 @@ void *backends_main(void *ptr) {
                                     "Content-Length: %zu\r\n"
                                     "Content-Type: application/x-www-form-urlencoded\r\n\r\n",
                                     remote_write_path,
-                                    hostname,
+                                    remote_hostname,
                                     data_size
                     );
 
--- a/backends/backends.c
+++ b/backends/backends.c
@@ -218,6 +218,13 @@ inline int backends_can_send_rrdset(BACKEND_OPTIONS backend_options, RRDSET *st)
     return 1;
 }
 
+inline int backends_can_send_rrdvar(BACKEND_OPTIONS backend_options, RRDVAR *rv) {
+    if (!simple_pattern_matches(charts_pattern, rv->name)) {
+        return 0;
+    }
+    return 1;
+}
+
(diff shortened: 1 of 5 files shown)


ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=4788751&bid=15945864)

backends/backends.c
```diff
@@ -507,6 +514,7 @@ void *backends_main(void *ptr) {
     const char *destination     = config_get(CONFIG_SECTION_BACKEND, "destination", "localhost");
     global_backend_prefix       = config_get(CONFIG_SECTION_BACKEND, "prefix", "netdata");
     const char *hostname        = config_get(CONFIG_SECTION_BACKEND, "hostname", localhost->hostname);
+    const char *remote_hostname = config_get(CONFIG_SECTION_BACKEND, "remote hostname", "localhost");
```
vlvkobal: The `remote hostname` option isn't needed. You should use the `destination` instead.

backends/backends.c
```diff
@@ -1069,7 +1077,7 @@ void *backends_main(void *ptr) {
                                     "Content-Length: %zu\r\n"
                                     "Content-Type: application/x-www-form-urlencoded\r\n\r\n",
                                     remote_write_path,
-                                    hostname,
+                                    remote_hostname,
```
vlvkobal: Please use the `destination`. It conforms to https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html.

vlvkobal:
As for the `simple_pattern_matches`, I think it's too much for a huge amount of variables we have in the database. It's also misleading, because `send charts matching` means _charts_ not _charts and variables_. We already have the `variables` option in the `allmetrics` API call, and I think we don't need an additional filter. For the remote write backend, you can add a similar option in the `backend` section, let's name it `expose variables`.

Ehekatl:
@vlvkobal Ya this pr is very hacky.
 the filter is ugly because I can't set flag on variables.

Let me explain what's this for
I'm very happy with all other variables, except the nic_speed one (which was added by myself😭
That was a convenient trick to calculate bandwidth alarm.

However, variables don't have dimension, so each interface result in a different variable name.
In the end we have massive labels polluting our tsdb.

I believe a better way to support this would be extend variable to have flag and dimension.

vlvkobal:
> I believe a better way to support this would be extend variable to have flag and dimension.

We are going already the [other way](https://github.com/netdata/netdata/issues/7730). I believe _Static system information_ and _labels_ could solve the problem with the `nic_speed`.

A the moment I suggest simplifying the PR - removing the `simple_pattern_matches` and adding the `expose variables` for the remote write.

ilyam8:
@Ehekatl Hi. Do you plan to finish the PR?

vlvkobal:
I'm closing the PR - it is outdated. We moved to the exporting engine and the backends subsystem will be deleted soon. The only thing that could be implemented in the exporting engine is variables in remote write so #7678 will still be open.

PR #7702, opened 2020-01-09
Parse host tags

### LINKED ISSUE #7700: Parse host tags
#### Feature idea summary
In the new exporters, we will no longer support "host tags". Users will have to define them as custom host labels in netdata.conf. We need to update the documentation to this effect. But we will need to parse existing "host tags" from the netdata.conf [backend] section and consider them custom host labels (need to check the format that each backend uses for tags/labels, so we know how to process them).

##### Expected behavior
Host tags are parsed in accordance with the configured backend format and stored as labels.

vlvkobal:
##### Summary
Fixes #7700 
Host tags in the `backend` section should be parsed in accordance with the configured backend format and stored as host labels.

##### Component Name
daemon


database/rrdhost.c
```diff
@@ -812,6 +813,136 @@ struct label *load_config_labels()
     return l;
 }
 
+static inline void strip_last_symbol(char *str, char symbol)
+{
+    char *end = str;
+
+    while (*end && *end != symbol)
+        end++;
+    if (*end && *end == symbol)
```
thiagoftsm: I think you do not need to compare again the `while` condition here, it is enough to check whether `*end == symbol`, because case it is not, this already means that `*end` is `NULL`.

vlvkobal: Removed an unnecessary check.

database/rrdhost.c
```diff
@@ -664,6 +664,7 @@ void rrdhost_free(RRDHOST *host) {
     // free it
 
     freez((void *)host->tags);
+    free_host_labels(host->labels);
```
amoss: The lock for the `host->labels` has not been acquired so this is unsafe.

mfundul: I'm not sure it is unsafe at this point. In theory the host data structure is unreachable here so that's why it can be freed as a whole as observed in the subsequent line:
`    freez(host);`

amoss: Which is the point that it becomes unreachable - and how does it interact with the lock around the labels list?

I think we might have to change the sections that acquire the labels lock so that they acquire the host lock first.

amoss: I had a more detailed look at the code and it is definitely a problem, there are different ways to fix it - please discuss which one would work best:
* Remove the separate lock for the labels completely and acquire the host lock to access the field.
* Acquire the host lock around every section that acquires the labels lock.
* Acquire the labels lock in `rdhost_free` before acquiring the host lock and then free it after the structure is unlinked.

ktsaou:


```diff

Clones added
============
- exporting/json/json.c  1
- exporting/graphite/graphite.c  1
         

Clones removed
==============
+ exporting/opentsdb/opentsdb.c  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=4789740&bid=15948278)

ktsaou:


```diff

Clones added
============
- exporting/graphite/graphite.c  1
- exporting/json/json.c  1
         

Clones removed
==============
+ exporting/opentsdb/opentsdb.c  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=4789740&bid=15948278)

database/rrdhost.c
vlvkobal: After a discussion with @mfundul we came to the conclusion that the problem is more general - there are a lot of places in Netdata, where host locks are not acquired but they should. I added host locks for host labels, but we need to check other possible places that miss host locks as well (it should be planned as a separate task).

vlvkobal:
I tested the PR with different types of backends (graphite, json, opentsdb, and default). For example, for the following configuration:
```
[backend]
        enabled = yes
        type = graphite
        # destination = localhost:1234
        host tags = host_tag1=host_value1;host_tag2=host_value2
        #host tags = host_tag1=host_value1 host_tag2=host_value2
        #host tags = "host_tag1" : "host_value1" , "host_tag2" : "host_value2"
        #host tags = host_tag1="host_value1",host_tag2="host_value2"
        #host tags = { "ho\"st_tag1" : "host_value1", "host_tag2" : "ho\"st_value2" }
        #host tags = [ "host_value1", "ho\"st_value2" ]
        #host tags = "host_json"
        #host tags = host_json
```
I get the following response for `api/v1/info` API call:
```
	"host_labels": {
		"host_tag1": "host_value1",
		"host_tag2": "host_value2",
		"_os_name": "Arch Linux",
		...
```

vlvkobal:
The host locking part was tested with
- `netdatacli reload-labels`
- `api/v1/info` API call
- streaming enabled on the master
- streaming enabled on the slave
- exporting connectors: graphite, json, opentsdb
- labels are written to the health log without crashes

It's easy to check in the code that the host is locked when `rrdcalc_labels_unlink_alarm_from_host()` and `rrdcalctemplate_is_there_label_restriction()` are called.

Everywhere when the `labels_rwlock` is acquired the host lock is also acquired or checked with `rrdhost_rdlock()` (Netdata crashes on the check when it is built with `-DNETDATA_INTERNAL_CHECKS=1 -DNETDATA_VERIFY_LOCKS=1`).

database/rrdcalc.c
```diff
@@ -662,6 +662,7 @@ static void rrdcalc_labels_unlink_alarm_loop(RRDHOST *host, RRDCALC *alarms) {
 }
 
 void rrdcalc_labels_unlink_alarm_from_host(RRDHOST *host) {
+    rrdhost_check_rdlock(host);
```
amoss: I don't understand why this line has been added, please walk me through why it is here. As far as I understand the `rdhost_check_rdlock()` call, it checks that a read-lock is currently held on the `rwlock` and it aborts Netdata if this is not true (through the `fatal()` call). So we are using this as a run-time assertion that there is a read-lock held on the `rwlock`. 
Looking at the call-site for this procedure in health.c:157 there is a `write-lock` acquired on the host immediately above the call. This does not seem to match the assumption that it is read-locked?

vlvkobal: We need at least a read lock and we don't mind to have a write lock. The `rdhost_check_rdlock()` call uses `pthread_rwlock_trywrlock()` underneath, so it basically means "any lock" is already acquired.

amoss: Yes I agree about where the code goes. When I looked up the manpage I was working from this description:
> The pthread_rwlock_wrlock() function shall apply a write lock to the read-write lock referenced by rwlock. The calling thread acquires the write lock if no other thread (reader or writer) holds the read-write lock rwlock. Otherwise, the thread shall block until it can acquire the lock. The calling thread may deadlock if at the time the call is made it holds the read-write lock (whether a read or write lock).

So, are we relying on `EDEADLK` to indicate that the calling thread aleady holds the lock? I could not see anywhere in the code that distinguished this case from `EBUSY`. What have I missed?

thiagoftsm:
When I use `JSON` as backend, I have the expected result when I do a request for `/api/v1/info/:

```
host_tag | "a=123 b=567"
```

but, when I tested with `OpenTSDB`  I am receiving

```
a=123 b=567 | "a=123 b=567 
```

and finally when I tested for `Prometheus` I am receiving:

```
a | "123 b=567"
```

I was expecting to have the same result that I had with json.
I used the following configurations during the tests:

```
[backend]
     host tags = a=123 b=567
     enabled = yes
     type = json
     #type = opentsdb:http 
     #type = opentsdb:htts
     #type = prometheus
```

vlvkobal:
You should use a string with an appropriate format for the `host tags` option for every type of backend.

vlvkobal:
For example, if `type = json`, you can use `host tags = { "a" : "123", "b" : "567" }` or `host tags "123"` or `host tags = 123`

thiagoftsm:
> For example, if `type = json`, you can use `host tags = { "a" : "123", "b" : "567" }` or `host tags = [ "123", "567" ]` or `host tags = "123"` or `host tags = 123`

Ok, I got it thanks!
But, considering that users can have wrong typo, should we ignore the specific label when it is not in the expected format instead to create a "wrong" label?

vlvkobal:
We can't figure out the format from the string itself. Anyway, if the user had wrong string in `host tags` he had a wrong output to the external database and data could not be received by the database. We do the parsing only for the backward compatibility.

database/rrdcalc.c
vlvkobal: We can see the number of the error in the debug message:
```
int ret = __netdata_rwlock_trywrlock(rwlock);

debug(D_LOCKS, "RW_LOCK: netdata_rwlock_trywrlock(0x%p) = %d in %llu usec, from %lu@%s, %s()", rwlock, ret, ... );
```

vlvkobal: Don't forget, that `rdhost_check_rdlock()` is for debugging only.

PR #7703, opened 2020-01-09
Enable the same name for alarm and dimension

### LINKED ISSUE #7670: apps plugin not in-sync between alarm and chart dashboard
##### Bug report summary
Before last night upgrade, everything was fine, but after that, as you see below, the alarm dashboard show nothing returns in terms of how many processes:
image

but in the dashboard, it shows the number of process running in real time:
image

the alarm configuration is pretty simple:
```
   alarm: dnsmasq
      on: apps.processes
    calc: $dnsmasq
   every: 10s
    crit: $this != 1
   units: processes
   delay: up 1m down 5m multiplier 1.5 max 1h
      to: sysadmin
```

for apps_groups.conf, just add a few lines at the top of the file:

```
sshd: sshd
dnsmasq: dnsmasq
autossh: autossh

```
CentOS Linux release 7.3.1611 (Core) 

netdata v1.19.0-239-nightly

### LINKED ISSUE #7834: Split indexes to enable different names on alarms
##### Bug report summary
In the issue #7670 the user described a bug that was present on Netdata, but after an investigation it was identified that there was an ancient bug affecting other areas on Netdata that needs to be fixed.

##### OS / Environment

/etc/os-release:ID=slackware

##### Netdata version
netdata v1.19.0-379-g129c57e2
 

##### Component Name
Health
##### Steps To Reproduce
Set dimension and alarm with the same name.
##### Expected behavior
Enable users to give any name for alarms.

thiagoftsm:
##### Summary
Fixes #7670 
Fixes #7834 

Netdata was using one index for `RRDVAR`s  in two different places of the code, this did not allow us to have variables with the same name, but our documentation did not define that this is not possible. This PR brings another index to fix this ancient bug.
##### Component Name
Health
##### Additional Information
I used the following alarm to test :

```
 alarm: user
      on: system.cpu
    calc: $user
   every: 10s
    crit: $this != 1
   units: processes
   delay: up 1m down 5m multiplier 1.5 max 1h
      to: sysadmin
```

I also use normal alarms to confirm that nothing was break.


database/rrd.h
```diff
@@ -483,6 +483,7 @@ struct rrdset {
     calculated_number red;                          // red threshold for this chart
 
     avl_tree_lock rrdvar_root_index;                // RRDVAR index for this chart
+    avl_tree_lock rrdvar_alarm_name_index;          // RRDVAR index to store the alarm names assoacited with this chart
```
stelfrag: Typo assoacited

mfundul: This is a persistent data structure and you cannot change it like that. This change destroys the signature and the compatibility of the chart for memory modes of `map` and `save`. As a result, all the database data will promptly be deleted, even if the user has data going back 1 year. You need to figure out a way around this, by exploiting the padding:
```
    size_t unused[5];
```

thiagoftsm: Thank you!

thiagoftsm: A very good catch, thank you!

amoss: Can we document this please? Preferably with a giant warning comment above the structure that explains that it is persistent and that changing it has implications on the database file storage.

thiagoftsm: To complement @amoss comments, I think the whole talk that we had during the meet up about `RRDHOST` and `RRDSET` should go to a documentation.

thiagoftsm: @mfundul I agree that padding could solve the problem with compatibility, but this would also bring some "ugly" code for the structure, so I decided to move the index completely outside the `RRDSET` structure.

database/rrdcalc.c
```diff
@@ -570,7 +577,10 @@ void rrdcalc_unlink_and_free(RRDHOST *host, RRDCALC *rc) {
     debug(D_HEALTH, "Health removing alarm '%s.%s' of host '%s'", rc->chart?rc->chart:"NOCHART", rc->name, host->hostname);
 
     // unlink it from RRDSET
-    if(rc->rrdset) rrdsetcalc_unlink(rc);
+    struct rrdcalc_rrdset_alarm search;
+    search.st = rc->rrdset;
+    struct rrdcalc_rrdset_alarm *rra = (struct rrdcalc_rrdset_alarm *)avl_search_lock(&host->alarms_idx_health_name, (avl *)&search);
```
mfundul: You should move those lines inside `rrdsetcalc_unlink()` and not change the function prototype. You can get the `rrdset` from inside the function, in the first line where it does `RRDSET *st = rc->rrdset;`.

database/rrdset.c
```diff
@@ -331,7 +331,10 @@ void rrdset_free(RRDSET *st) {
     freez(st->exporting_flags);
 
     while(st->variables)  rrdsetvar_free(st->variables);
-    while(st->alarms)     rrdsetcalc_unlink(st->alarms);
+    struct rrdcalc_rrdset_alarm search;
+    search.st = st;
+    struct rrdcalc_rrdset_alarm *rra = (struct rrdcalc_rrdset_alarm *)avl_search_lock(&host->alarms_idx_health_name, (avl *)&search);
```
mfundul: You should move those lines inside `rrdsetcalc_unlink()` and not change the function prototype. You can get the `rrdset` from inside the function, in the first line where it does `RRDSET *st = rc->rrdset;`.

mfundul:
I would say there is a more generic problem remaining, other than what's described in the bug report, the alarm name can be the same with various combinations of the concatenation of `chart id`/`chart name`/`family name`/`context`.

In practice, a user can name an alarm similarly to a context/family/chart name and have collisions, that might even lead to memory corruption.

database/rrdcalc.c
mfundul: Are you doing this for performance reasons? To save 2 lookups?

database/rrdcalc.h
```diff
@@ -139,14 +139,23 @@ struct rrdcalc {
     struct rrdcalc *next;
 };
 
+struct rrdcalc_rrdset_alarm {
+    avl avl1;
+
+    RRDSET *st;
+    avl_tree_lock rrdvar_alarm_name_index;
+
+    struct rrdcalc_rrdset_alarm *next;
```
mfundul: I don't see `next` being used anywhere in the code, other than iterating for `alarm_index_link()` and `alarm_index_unlink_and_free()` but I don't see how they're doing something with it.

database/rrd.h
```diff
@@ -738,6 +738,9 @@ struct rrdhost {
     RRDCALCTEMPLATE *templates;
     RRDCALCTEMPLATE *alarms_template_with_foreach;
 
+    //Index used with health
+    avl_tree_lock alarms_idx_health_name;
+    struct rrdcalc_rrdset_alarm *alarm_set;
```
mfundul: I don't see `alarm_set` being used anywhere in the code, other than iterating for `alarm_index_link()` and `alarm_index_unlink_and_free()` but I don't see how they're doing something with it.

ilyam8:
> In practice, a user can name an alarm similarly to a context/family/chart name and have collisions, that might even lead to memory corruption.

Do we need to fix specific case

> Enable the same name for alarm and dimension

then?

mfundul:
I'm wondering if the best solution is to just prefix every alarm name with "_alarm" somehow and be done with it.

database/rrdcalc.c
thiagoftsm: Yes, it was due performance.

database/rrdcalc.h
thiagoftsm: Clion is not showing usage, but you can find the usage inside rrdhost in the following functions:

- `alarm_index_link`
- `alarm_index_unlink_and_free`

ilyam8:
> I'm wondering if the best solution is to just prefix every alarm name with "_alarm" somehow and be done with it.

I agree, but if we do this we break using alarm calc value as variable in another alarm

https://github.com/netdata/netdata/blob/49f6950820f083956ddc2c9327eeae4e9ef0d5ed/health/health.d/mysql.conf#L105-L122

database/rrdcalc.h
mfundul: But they're not doing something with those functions.

database/rrd.h
thiagoftsm: This is also used inside `rrdhost` in the same functions that I described before.

database/rrdcalc.h
mfundul: `avl_remove_lock` does all the work anyway.

database/rrd.h
mfundul: But `alarm_set` doesn't do anything. `avl_remove_lock()` does all the work.

database/rrdcalc.h
thiagoftsm: `avl_remove_lock` removes the link inside the tree, the functions were created to free the structure that was allocated before.

database/rrd.h
thiagoftsm: I fixed it.

database/rrdcalc.h
thiagoftsm: Thank you for the explanation, it was completely removed with the last commit.

database/rrd.h
thiagoftsm: It was removed with the last commit.

thiagoftsm:
> > In practice, a user can name an alarm similarly to a context/family/chart name and have collisions, that might even lead to memory corruption.
> 
> Do we need to fix specific case
> 
> > Enable the same name for alarm and dimension
> 
> then?
> 
> Or it should be complex solution
> 
> #7670 is not that important and high priority

The issue is related with `name` and `dimensions`, my suggestion is we fix this with the PR and we create another issue with the others.  I can work during this sprint for the other cases when we finalize this.

ilyam8:
> The issue is related with name and dimensions, my suggestion is we fix this with the PR and we create another issue with the others. 

Why do we need partial fix? That is my point. There is no hurry with #7670. We dont need to fix it asap.

thiagoftsm:
> > The issue is related with name and dimensions, my suggestion is we fix this with the PR and we create another issue with the others.
> 
> Why do we need partial fix? That is my point. There is no hurry with #7670. We dont need to fix it asap.

All right, I am working with 7297 now and I will check the other cases when I finish it.

ilyam8:
We identified bigger problem investigating #7670. I suggest to open a new issue with proper problem description and close  #7670 if favor of it (and do not merge the fix). @mfundul what do you think?

mfundul:
> 
> 
> We identified bigger problem investigating #7670. I suggest to open a new issue with proper problem description and close #7670 if favor of it (and do not merge the fix). @mfundul what do you think?

Fine by me.

stelfrag:
> We identified bigger problem investigating #7670. I suggest to open a new issue with proper problem description and close #7670 if favor of it (and do not merge the fix). @mfundul what do you think?

I think it may be one of those cases where the documentation is updated to indicate the strange behavior if variables clash and ,  as it is low priority, handle it later

ktsaou:


```diff

Clones added
============
- database/rrdvar.c  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=4790012&bid=15949268)

thiagoftsm:
> I would say there is a more generic problem remaining, other than what's described in the bug report, the alarm name can be the same with various combinations of the concatenation of `chart id`/`chart name`/`family name`/`context`.
> 
> In practice, a user can name an alarm similarly to a context/family/chart name and have collisions, that might even lead to memory corruption.

I created the other two indexes that we were missing to complete split alarms from charts. I need to call the attention for the fact that in normal situation the unique case that we will return a value different of NAN is when the dimension is also used as name(It is also possible to return NAN here, but the dimension must be NAN), because the other variables are not calculated at all.

database/rrdcalc.c
```diff
@@ -183,16 +195,24 @@ inline void rrdsetcalc_unlink(RRDCALC *rc) {
...
+    struct rrdcalc_rrdset_alarm *rra = (struct rrdcalc_rrdset_alarm *)avl_search_lock(&host->alarms_idx_health_name, (avl *)search);
+    if(rra) {
+        rrdvar_free(host, &rra->rrdvar_alarm_index, rc->local);
+    }
     rc->local = NULL;
 
-    rrdvar_free(host, &st->rrdfamily->rrdvar_root_index, rc->family);
+    rra = (struct rrdcalc_rrdset_alarm *)avl_search_lock(&host->alarms_idx_health_family, (avl *)search);
+    if(rra) {
+        rrdvar_free(host, &rra->rrdvar_alarm_index, rc->local);
```
mfundul: Shouldn't this be:
```
         rrdvar_free(host, &rra->rrdvar_alarm_index, rc->family); 
```
?

database/rrdcalc.c
```diff
@@ -183,16 +195,24 @@ inline void rrdsetcalc_unlink(RRDCALC *rc) {
...
+    }
     rc->family = NULL;
 
-    rrdvar_free(host, &host->rrdvar_root_index, rc->hostid);
+    rra = (struct rrdcalc_rrdset_alarm *)avl_search_lock(&host->alarms_idx_health_hostid, (avl *)search);
+    if(rra) {
+        rrdvar_free(host, &rra->rrdvar_alarm_index, rc->local);
```
mfundul: Shouldn't this be:
```
         rrdvar_free(host, &rra->rrdvar_alarm_index, rc->hostid); 
```
?

database/rrdvar.c
```diff
@@ -227,17 +241,19 @@ int health_variable_lookup(const char *variable, uint32_t hash, RRDCALC *rc, cal
...
 
-    rv = rrdvar_index_find(&host->rrdvar_root_index, variable, hash);
-    if(rv) {
-        *result = rrdvar2number(rv);
+    ret = health_variable_lookup_test(result, &search, variable, hash, &host->alarms_idx_health_family);
+    if(ret)
+        return 1;
+
+    ret = health_variable_lookup_test(result, &search, variable, hash, &host->alarms_idx_health_hostid);
+    if(ret)
```
mfundul: What about all the variables that are in `st->rrdfamily->rrdvar_root_index` and `host->rrdvar_root_index`? Are they lost forever? For example, dimension variables, see here:
https://github.com/netdata/netdata/blob/cb7f19ba2dc40d24ec9cd68c08a7d20acd891a37/database/rrddimvar.c#L134-L151

database/rrdcalc.c
thiagoftsm: You are right, I copied and pasted, but I forgot to change. Thank you!

database/rrdcalc.c
thiagoftsm: Yes, thank you!

database/rrdvar.c
thiagoftsm: Of course no, you are correct, I fix the problem now.

Thank you!

database/rrdcalc.c
```diff
@@ -66,15 +66,27 @@ static void rrdsetcalc_link(RRDSET *st, RRDCALC *rc) {
...
+    struct rrdcalc_rrdset_alarm search;
+    search.st = st;
+    struct rrdcalc_rrdset_alarm *rra = (struct rrdcalc_rrdset_alarm *)avl_search_lock(&host->alarms_idx_health_name, (avl *)&search);
+    if (rra) {
+        rc->local  = rrdvar_create_and_index("local",  &rra->rrdvar_alarm_index, rc->name, RRDVAR_TYPE_CALCULATED, RRDVAR_OPTION_RRDCALC_LOCAL_VAR, &rc->value);
+    }
+
+    rra = (struct rrdcalc_rrdset_alarm *)avl_search_lock(&host->alarms_idx_health_family, (avl *)&search);
+    if (rra) {
+        rc->family = rrdvar_create_and_index("family", &rra->rrdvar_alarm_index, rc->name, RRDVAR_TYPE_CALCULATED, RRDVAR_OPTION_RRDCALC_FAMILY_VAR, &rc->value);
```
mfundul: Due to `search.st` you cannot find alarm names from a different chart that belongs to the same family anymore.

database/rrdcalc.c
```diff
@@ -66,15 +66,27 @@ static void rrdsetcalc_link(RRDSET *st, RRDCALC *rc) {
...
     snprintfz(fullname, RRDVAR_MAX_LENGTH, "%s.%s", st->id, rc->name);
-    rc->hostid   = rrdvar_create_and_index("host", &host->rrdvar_root_index, fullname, RRDVAR_TYPE_CALCULATED, RRDVAR_OPTION_RRDCALC_HOST_CHARTID_VAR, &rc->value);
+    rra = (struct rrdcalc_rrdset_alarm *)avl_search_lock(&host->alarms_idx_health_hostid, (avl *)&search);
+    if (rra) {
+        rc->hostid   = rrdvar_create_and_index("host", &rra->rrdvar_alarm_index, fullname, RRDVAR_TYPE_CALCULATED, RRDVAR_OPTION_RRDCALC_HOST_CHARTID_VAR, &rc->value);
 
-    snprintfz(fullname, RRDVAR_MAX_LENGTH, "%s.%s", st->name, rc->name);
-    rc->hostname = rrdvar_create_and_index("host", &host->rrdvar_root_index, fullname, RRDVAR_TYPE_CALCULATED, RRDVAR_OPTION_RRDCALC_HOST_CHARTNAME_VAR, &rc->value);
+        snprintfz(fullname, RRDVAR_MAX_LENGTH, "%s.%s", st->name, rc->name);
+        rc->hostname = rrdvar_create_and_index("host", &rra->rrdvar_alarm_index, fullname, RRDVAR_TYPE_CALCULATED, RRDVAR_OPTION_RRDCALC_HOST_CHARTNAME_VAR, &rc->value);
```
mfundul: Due to `search.st` you cannot find alarm names from a different chart that belongs to the same host anymore.

cosmix:
Guys what’s going on with this PR?

mfundul:
> 
> 
> Guys what’s going on with this PR?

It was deprioritized by @cakrit

cosmix:
@cakrit is this PR still relevant?

thiagoftsm:
> @cakrit is this PR still relevant?

It was deprioritized, so I wil close it.

Issue #7198, opened 2019-10-25
kickstart-static64.sh does not write a proper REINSTALL_COMMAND to .environment


knatsakis:
##### Bug report summary
kickstart-static64.sh does not write a proper REINSTALL_COMMAND to /opt/netdata/etc/netdata/.environment

##### OS / Environment
kickstart-static64.sh installations

##### Netdata version (ouput of `netdata -V`)
v1.18.1

##### Component Name
kickstart-static64.sh

##### Steps To Reproduce
Install using kickstart-static64.sh. The REINSTALL_COMMAND in /opt/netdata/etc/netdata/.environment is always `./netdata-installer.sh --install /opt`

##### Expected behavior
The REINSTALL_COMMAND should contain all the options that where passed to the initial installer (esp. --auto-updates)

Kramerican:
As a user who is quite confused as to what is happening with auto-updates, I just have a question: Does this explain why, seemingly, netdata removes the /etc/cron.daily/netdata-updater cron job on its own accord?

I ask as I thought this had been solved, re-installed netdata and doublechecked that the updater script was in the cron folder.

Now when I check I see I am several versions behind, and the cronjob has disappeared.

knatsakis:
@Kramerican, I am sorry that this has caused confusion.

I think this issue, along with #7197 explains the behavior that you are describing. Auto-update does not work with kickstart-static64.sh currently. (Only with kickstart.sh)

We are working to resolve this issue and you can monitor progress here and in #7197.

Thanks a lot for your feedback!

Issue #7197, opened 2019-10-25
netdata-updater.sh script disables itself on kickstart-static64.sh installations


knatsakis:
##### Bug report summary
The netdata-updater.sh script disables automatic cron updates (disables itself), on kickstart-static64.sh installations

##### OS / Environment
kickstart-static64.sh installations.

##### Netdata version (ouput of `netdata -V`)
v1.18.1

##### Component Name
netdata-updater.sh

##### Steps To Reproduce
Install netdata using kickstart-static64.sh. Run the netdata-updater.sh script (/etc/cron.daily/netdata-updater). The netdata-updater script is removed from /etc/cron.daily

##### Expected behavior
The netdata-updater.sh script should remain under /etc/cron.daily. (It must use the REINSTALL_COMMAND variable from .environment)

knatsakis:
This is closely related to #7039

knatsakis:
Blocked by #7198

knatsakis:
Consider reverting commit 6277b861d9948ec10ab40b09cd0c8e9e5201780b when this is fixed

knatsakis:
@Ferroin @prologic we should not forget to revert 6277b861d9948ec10ab40b09cd0c8e9e5201780b when this is fixed.

Ferroin:
Based on investigation, this is a result of the `netdata-updater.sh` script not supporting installs with the `kickstart-static64.sh` script.

I'm going to investigate getting this working correctly, but it will likely take a while.

Kramerican:
I am now seeing this exact same behavior resurface. My Netdata installations auto-updated for a while, and now they are stuck on v1.19.0-460-gc92dd8ee

checking /etc/cron.daily reveals that the netdata updater job has been removed from there, once again ...

ikarlo:
> I am now seeing this exact same behavior resurface. My Netdata installations auto-updated for a while, and now they are stuck on v1.19.0-460-gc92dd8ee
> 
> checking /etc/cron.daily reveals that the netdata updater job has been removed from there, once again ...

Also on v1.20 in my installations

prologic:
[...]

PR #7704, opened 2020-01-09
Make auto-updates work on kickstart-static64 installs.

Ferroin:
##### Summary

This PR makes auto-updating work correctly for installations done using the `kickstart-static64.sh` installer.

It consists of two logical parts:

1. Update handling of the `.environment` file to store only the options to be used for reinstallation instead of the full command. We currently only have two possible 'commands', and they are trivially differentiated based on whether the install is with `kickstart.sh` or `kickstart-static64.sh`, so we don't really need to store the command that was used.
2. Actually enable the currently dead code in `netdata-updater.sh` that handles the case of static installs, and adds some extra support code to make it work correctly.

##### Component Name

area/packaging

##### Additional Information

Fixes: #7197 
Fixes: #7198 

Minimally tested so far (passes CI lifecycle checks without issue), will mark ready for review once more rigorous testing is finished.

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -93,16 +93,12 @@ printf 'LDFLAGS="%s" ' "${LDFLAGS}" >>netdata-installer.log
 printf "%q " "${PROGRAM}" "${@}" >>netdata-installer.log
 printf "\\n" >>netdata-installer.log
 
-REINSTALL_COMMAND="$(
-	printf "%q " "${PROGRAM}" "${@}"
+REINSTALL_OPTIONS="$(
+	printf "%s" "${*}"
 	printf "\\n"
 )"
 # remove options that shown not be inherited by netdata-updater.sh
-REINSTALL_COMMAND="${REINSTALL_COMMAND// --dont-wait/}"
-REINSTALL_COMMAND="${REINSTALL_COMMAND// --dont-start-it/}"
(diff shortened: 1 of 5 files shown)


Ferroin:
Now tested in Docker on a number of distros with local builds of the static distribution tarballs.

netdata-installer.sh
```diff
@@ -93,16 +93,12 @@ printf 'LDFLAGS="%s" ' "${LDFLAGS}" >>netdata-installer.log
...
+	printf "%s" "${*}"
 	printf "\\n"
 )"
 # remove options that shown not be inherited by netdata-updater.sh
-REINSTALL_COMMAND="${REINSTALL_COMMAND// --dont-wait/}"
-REINSTALL_COMMAND="${REINSTALL_COMMAND// --dont-start-it/}"
-if [ "${REINSTALL_COMMAND:0:1}" != "." ] && [ "${REINSTALL_COMMAND:0:1}" != "/" ] && [ -f "./${PROGRAM}" ]; then
-	REINSTALL_COMMAND="./${REINSTALL_COMMAND}"
-fi
+REINSTALL_OPTIONS="$(echo "${REINSTALL_OPTIONS}" | sed 's/--dont-wait//g' - | sed 's/--dont-start-it//g' -)"
```
knatsakis: The `-` is not needed as an argument to sed

```suggestion
REINSTALL_OPTIONS="$(echo "${REINSTALL_OPTIONS}" | sed 's/--dont-wait//g' | sed 's/--dont-start-it//g')"
```

PR #7705, opened 2020-01-09
litespeed: add support for different .rtreport format

lucasRolff:
##### Summary

Add support for different .rtreport format

Some .rtreport files do not contain `MAXCONN` and `MAXSSL_CONN` variables, but simply list the in use connections.

This PR adds support for this by looking for lines starting with `PLAINCONN:` as well.

##### Component Name

python.d litespeed

##### Additional Information

I'm ops @ LiteSpeed Technologies


ilyam8:
@lucasRolff thanks!

PR #7713, opened 2020-01-13
minor - update README.md for ACLK

underhood:
adds INSTRUCTIONS for modified mosquitto:
had to search for it so next guys can find it here right away

### DIFF
--- a/aclk/README.md
+++ b/aclk/README.md
@@ -1,8 +1,11 @@
 This is the agent cloud link (ACLK) information file.
 
+### For the `mosquitto` lib:
 The installer does not yet have steps to pull and build the local library.
 cd project_root
+
+``` text
 git clone ssh://git@github.com/netdata/mosquitto mosquitto/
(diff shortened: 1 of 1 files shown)


aclk/README.md
```diff
@@ -1 +1,8 @@
-This is the agent cloud link (ACLK) information file
\ No newline at end of file
+This is the agent cloud link (ACLK) information file.
+
+The installer does not yet have steps to pull and build the local library.
+cd project_root
+git clone ssh://git@github.com/netdata/mosquitto mosquitto/
+(cd mosquitto/lib && make)      # Ignore the cpp error
```
ilyam8: please use

```
text
```

see current

> https://github.com/netdata/netdata/blob/underhood-patch-1/aclk/README.md

doesnt look good

underhood: Hi this is only temporary for internal use. Not user facing documentation.

ilyam8: > internal use

if internal use implies reading... 😄

PR #7725, opened 2020-01-14
Cleanup packaging/makeself/build-x86_64-static.sh to use /bin/sh and remove use of sudo

prologic:
##### Summary

This is just a little PR to cleanup and remove the use of `sudo` of the script used to build/rebuild the self-extracting tarballs we use in the `kickstart-static64.sh` installer.

Couple of things here:

- Use POSIX Sh here `#!/bin/sh` as we're not really using any Bash(isms) anyway and its a simple script
- Remove the use of `sudo`; mostly used in `sudo docker ...` which is not necessary as typically you would have read/write permissions to the Docker daemon UNIX socket on your system (_if you don't; you should!_)
  - One less thing run as root

##### Component Name

Packaging

##### Additional Information

On first run (_wasn't able to repro again_) `tar` complains it cannot find `bash-4.18.4/ChangeLog` despite it actually being available on the file system when this script is executed.

### DIFF
--- a/packaging/makeself/build-x86_64-static.sh
+++ b/packaging/makeself/build-x86_64-static.sh
@@ -1,4 +1,5 @@
-#!/usr/bin/env bash
+#!/bin/sh
+
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 . $(dirname "$0")/../installer/functions.sh || exit 1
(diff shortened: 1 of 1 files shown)


prologic:
> On first run (wasn't able to repro again) tar complains it cannot find bash-4.18.4/ChangeLog despite it actually being available on the file system when this script is executed.

This is pretty consistent on my dev environment (_macOS + Docker Desktop + Brew user-land_); I suspect its an issue on my system and maybe related to macOS filesystems.

prologic:
I'm so very sorry. This broke our Nightlies and Static builds 😢

cc @ilyam8 ^^^ I believe this is the root cause. Putting up a fix now,.

prologic:
#7971 fixes what I broke here :/

PR #7735, opened 2020-01-15
Adds Docker based build system for Binary Packages, CI/CD, Smoke Testing and Development.

prologic:
##### Summary

This is still work-in-progress; but utilises the work @Ferroin did in netdata/helper-images#31 to put together a fully functional (_only `debian:buster` tested at this point_) `Dockerfile` that does:

- Builds the Debian package
- Tests the Debian package can be installed in a clean environment
- Runs some basic smoke tests

This ensures a few things:

- Our packages actually work and their dependencies are correct
- The NetData agent actually runs
- Some basic functionality is not broken

##### Component Name

Packaging / CI

##### Additional Information

To test this; firstly go checkout netdata/helper-iamges#31 If you use [hub](https://github.com/github/hub) you can just do this:

```bash
$ git co https://github.com/netdata/helper-images/pull/31/files
$ docker build -t netdata/builder:debian_buster -f package-builders/Dockerfile.debian_buster .
```

Then you can test this PR with:

```bash
$ docker build --build-arg DISTRO=debian --build-arg DISTRO_VERSION=buster -t netdata/netdata:debian_buster .
$ docker run --rm netdata/netdata:debian_buster
```


ktsaou:


```diff

Issues
======
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=4822465&bid=16030468)

ktsaou:


```diff

Issues
======
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=4822465&bid=16030468)

.dockerfiles/install.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+install_deb() {
+	apt-get update
+	apt-get install -y --no-install-recommends \
+		curl=7.64.0-4 \
+		netcat=1.10-41.1 \
+		jq=1.5+dfsg-2+b1
```
Ferroin: We should ideally be installing these after Netdata, to ensure they don't interfere with it pulling in the correct dependencies.

Additionally, pinning specific versions is likely to break the install process. Additionally, everything we appear to be doing in the test scripts with these has been stable behavior for many years, so it's not like we need to worry about it changing (I'd actually think we're more likely to hit issues because of pinned versions not being available than due to the interfaces we're using here changing).

.dockerfiles/install.sh
```diff
@@ -0,0 +1,22 @@
...
+	apt-get install -y "/artifacts/netdata_${VERSION}_${ARCH}.deb"
+	apt-get clean
+	rm -rf /var/lib/apt/lists/*
```
Ferroin: While this is generally good practice, I personally see little value in doing it. The images being used here are going to be deleted in very short order in a significant majority of cases, so the few dozen MB of extra space usage quite simply shouldn't matter enough to warrant spending time on reducing it.

.dockerfiles/test.sh
```diff
@@ -0,0 +1,27 @@
...
+		i=$((i + 1))
+		if [ "$i" -gt "$timeout" ]; then
+			printf "Timed out!\n"
+			return 1
+		fi
+	done
+	printf "OK\n"
+}
+
+netdata -D >netdata.log 2>&1 &
```
Ferroin: Depending on the distro, Netdata may actually already have been started. We should account for that here.

Dockerfile
```diff
@@ -0,0 +1,37 @@
+ARG ARCH=amd64
+ARG DISTRO=debian
+ARG DISTRO_VERSION=buster
+ARG VERSION=0
+
+FROM netdata/builder:${DISTRO}_${DISTRO_VERSION} AS build
```
Ferroin: We should probably remove the `_` here and update the Debian images to use that format (and a version number instead of a codename). Debian is really the only anomalous one here, all the others are already using `${DISTRO}${DISTRO_VERSION}` for the tag, and Debian is only the way it is right now because I was trying to match up with the tags we were using for the os-test repository.

prologic:
This now works for Fedora too (_tested on `fedora:30`_):

```sh
$ docker build -f package-builders/Dockerfile.fedora30 -t netdata/builder:fedora_30 .
$ docker build --build-arg DISTRO=fedora --build-arg DISTRO_VERSION=30 -t netdata/netdata:fedora_30 .
$ docker run -i -t --rm netdata/netdata:fedora_30
```

.dockerfiles/install.sh
prologic: > We should ideally be installing these after Netdata, to ensure they don't interfere with it pulling in the correct dependencies.

Agreed! I'll change this.

prologic: > Additionally, pinning specific versions is likely to break the install process. Additionally, everything we appear to be doing in the test scripts with these has been stable behavior for many years, so it's not like we need to worry about it changing (I'd actually think we're more likely to hit issues because of pinned versions not being available than due to the interfaces we're using here changing).

Okay I buy it :) But how do I make Codeacy shut up about this ? :)

.dockerfiles/install.sh
prologic: yes yes I agree; I'll remove these lines for now.

My thinking here long-term is we can in theory also replace the official Docker image that folks use to run NetData with _any_ of these. i.e: Our CI/CD pipelines can also publish these resulting images; we may not of course decide to publish all of them, just a few (like 1 alpine, 1 debian and 1 rhel) -- Let users decide which flavour of the Docker image they want to run up in their environments -- many of the upstream official images do this too :)

.dockerfiles/test.sh
prologic: Hmm actually I'm not certain about this myself. In many cases the Docker image for many distros doesn't actually start the typical "init" process. In the case of SystemD this means even though the package installed the unit file, it won't be started.

This is okay because we're not really interested in testing SystemD here -- We should of course (_separately_) ensure we properly test our SystemD (_or other startup mechanisms_) elsewhere.

Dockerfile
prologic: I don't actually mind which way we do this. It will make no difference. Just bare in mind I'd like to get this whole flow working for "multiarch" too (_somehow_)

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+ENV DISTRO=$DISTRO
+ENV DISTRO_VERSION=$DISTRO_VERSION
+ENV VERSION=$VERSION
+
+COPY .dockerfiles/install.sh /install.sh
+COPY .dockerfiles/test.sh /test.sh
+
+COPY --from=build /netdata/artifacts /artifacts
+
+RUN /install.sh || exit 1
```
knatsakis: Is `|| exit 1` needed? Is it not true that if a RUN command fails then the build fails?

.dockerfiles/install.sh
Ferroin: > Okay I buy it :) But how do I make Codeacy shut up about this ? :)

No idea, I generally just ignore it when it's making recommendations that are situationally stupid like this.

.dockerfiles/test.sh
Ferroin: Yeah, systemd won't start it, but I'm not sure that 'conventional' init systems won't. Debian's old-style init.d stuff might in this case (though I don't think any of that is being used in the Debian Docker images.

Dockerfile
Ferroin: I'd rather have the numeric version number for the Debian images like we do for everything else. I'll update the tags on Docker Hub so that the new images get generated this way.

.dockerfiles/install.sh
prologic: Okay 🤣

Dockerfile
prologic: 👍

.dockerfiles/install.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/sh
+
+install_debian_like() {
+	# This is needed to ensure package installs don't prompt for any user input.
+	ENV DEBIAN_FRONTEND=noninteractive
```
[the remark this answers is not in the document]
prologic: Fixed :)

.dockerfiles/test.sh
```diff
@@ -0,0 +1,27 @@
...
+	name="${3}"
+	timeout="${4:-30}"
+
+	printf "Waiting for %s on %s:%s ... " "${name}" "${host}" "${port}"
+
+	i=0
+	while ! nc -z "${host}" "${port}"; do
+		sleep 1
+		i=$((i + 1))
+		if [ "$i" -gt "$timeout" ]; then
```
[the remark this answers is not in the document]
prologic: True -- In Bash :)

> In POSIX sh, ++ is undefined. [SC2039]

.dockerignore
```diff
@@ -0,0 +1 @@
+.gitignore
```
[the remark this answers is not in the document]
prologic: This is symlinked currently. It looks a bit wierd in Github.

```
lrwxr-xr-x  1 prologic staff     10 Jan 18 08:49 .dockerignore -> .gitignore
```

I'll have to copy `.gitignore` to .`dockerignore` and duplicate it if you want me to ignore more things in the Docker context sent to the daemon. But `Dockerfile` and `.dockerignore` are small so 🤷‍♂️

Dockerfile
prologic: Yes it does. Fixed :)

.dockerignore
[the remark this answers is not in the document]
knatsakis: No worries, It was just a suggestion. I hadn't realized it was a symlink.

Dockerfile
Ferroin: Unresolving this as it's not been updated here.

Also, the image name is currently `netdata/package-builders` for the package builder images. `netdata/builder` is something different (I think it's used for the conainerized build tests). The tag names there are currently `${DISTRO_NAME}${DISTRO_VERSION}`, where the version is the standard numeric representation of the release version for the distro (8/910 for debian, 29/30/31 for fedora, 16.04/18.04/etc for Ubuntu, 15.0/15.1 for SuSE, etc).

Ferroin: Quick change to resolve the above conversation:

```suggestion
FROM netdata/package-builders:${DISTRO}${DISTRO_VERSION} AS build
```

Dockerfile
```diff
@@ -0,0 +1,42 @@
+ARG ARCH=amd64
+ARG DISTRO=debian
+ARG DISTRO_VERSION=buster
```
Ferroin: Secondary change needed as a result of changed Docker Hub tags mentioned in the below conversation:

```suggestion
ARG DISTRO_VERSION=10
```

Ferroin:
Also, almost fogot, don't forget to handle the aliasing for the OpenSUSE images here as mentioned in the description of https://github.com/netdata/helper-images/pull/40.

prologic:
> However, we might want to move the Dockerfile under `packaging` and name it something like `Dockerfile.packager` to avoid any confusion regarding it's relation to our official Docker images (which is built using `packaging/docker/Dockerfile`).

To be honest; I'd really would like us to reverse this entirely.

- Move `packaging/docker/Dockerfile` to `/Dockerfile`. That's there it _should_ belong so its obvious to our users from the root of the repo that a) Yes we are "Dockerized" and b) Yes you can just `docker build .` this as usual (_if that doesn't work for some reason we should fix that!_)
- Move _this_ `Dockerfile` to `packaging/Dockerfile.packager`

Sound good?

prologic:
> Also, almost fogot, don't forget to handle the aliasing for the OpenSUSE images here as mentioned in the description of [netdata/helper-images#40](https://github.com/netdata/helper-images/pull/40).

There is no CI sround this at the moment; and I won't be hooking any up in _this_ PR until its merged.

This `Dockerfile` _really_ needs to be invokved with a little bit of shell to work correctly -- the `--build-arg` / etc ... -- But otherwise yes I won't forget in the next PR :)

prologic:
> Looks good other than the explicitly suggested changes. I've done testing using those and everything seems to work correctly as far as I can tell.
> 
> However, we might want to move the Dockerfile under `packaging` and name it something like `Dockerfile.packager` to avoid any confusion regarding it's relation to our official Docker images (which is built using `packaging/docker/Dockerfile`).

I've incorporated your suggested changes as well as restructured this a bit. But Github still shows you as "Requesting Changes" 🤦‍♂️-- Also I can't re-request your review 🤦‍♂️

Mind taking a quick look at the recent changes? 😃

.dockerfiles/test.sh
```diff
@@ -0,0 +1,26 @@
...
+	port="${2}"
+	name="${3}"
+	timeout="${4:-30}"
+
+	printf "Waiting for %s on %s:%s ... " "${name}" "${host}" "${port}"
+
+	i=0
+	while ! nc -z "${host}" "${port}"; do
+		sleep 1
+		if [ "$((++i))" -gt "$timeout" ]; then
```
knatsakis: I thought that `$((++i))` was not working with `/bin/sh`. Maybe switch to bash?

knatsakis: Also, maybe we could use something like https://github.com/vishnubob/wait-for-it (it has a Debian/Ubuntu package) or https://hub.docker.com/r/waisbrot/wait (It's a docker image) to avoid re-inventing the wheel? Just a suggestion.

ilyam8:
> Also I can't re-request your review

I wonder why 🤔 This `Re-request review` thing doenst work?

Screenshot 2020-01-24 at 15 12 32

packaging/scripts/install.sh
```diff
@@ -0,0 +1,56 @@
...
+
+install_fedora_like() {
+  # Using a glob pattern here because I can't reliably determine what the
+  # resulting package name will be (TODO: There must be a better way!)
+
+  # Install NetData
+  dnf install -y /artifacts/netdata-"${VERSION}"-*.rpm
+
+  # Install testing tools
+  dnf install -y curl nc jq
```
Ferroin: Just noticed this, but this will fail on CentOS 7 (it still uses `yum`).

prologic:
> I wonder why 🤔 This `Re-request review` thing doenst work?

For whatever reason that wasn't showing up as an option after I did the following sequence of things:

- Committed @Ferroin 's suggestions in the Github Review UI
- Pulled those changes in locally via `git pull --rebase`
- Made some more local commits
- And re-pushed all changes.

This seems to have confused Github somehow 🤣

.dockerfiles/test.sh
prologic: > I thought that $((++i)) was not working with /bin/sh. Maybe switch to bash?

Yeah it's a Bash thing :)

prologic: I want to use `/bin/sh` mostly in our CI/CD because I don't really care for pulling in Bash and all its dependencies :) Some ~13MB on Alpine for example

prologic: > Also, maybe we could use something like https://github.com/vishnubob/wait-for-it (it has a Debian/Ubuntu package) or https://hub.docker.com/r/waisbrot/wait (It's a docker image) to avoid re-inventing the wheel? Just a suggestion.

I _actually_ completely agree; I borrowed this code from elsewhere anyway. But for now this will work. It might get iterated out for sure :)

packaging/scripts/install.sh
prologic: Fixed

prologic: Just writing a quick `./packaging/build_package_install_test.sh` script as a quick starting point for kickstarting the new CI/CD for all of this.

I got sick and tired of typing :)

prologic: Will mark this as resolved once I have the script working and CentOS 7 tested.

Issue #7221, opened 2019-10-29
Netdata - All streammed metrics in one display


duhzon:
Hello there, im trying to setup a networking in netdata with main node and other 14 pc. So far i got all pc running  netdata separately  being acessible by any ssh with any subnet  localhost:19999. I know my next step is stream, i got a tutorial here [stream]. My goal is to make netdata show in one display every pc side by side every cpu usage(or other metric) in a single screen,for example. I heard that its possible using java script or python but i dont have any clue where to get information about this, already googled, searched in here, etc. I really have to set node and stream for that? There is a tutorial for that? Its more a question than a issue.
Thanks.

##### OS / Environment
Im using Fedora 26 (Server Edition)
My netdata is in version Netdata v1.18.1

thiagoftsm:
Hi @duhzon ,

The stream was created to centralize the information in only one place, we also have Netdata cloud that was made for you have the hosts in only one screen without the necessity to create your own dashboard. 

You can find our documentation for stream in this link https://docs.netdata.cloud/streaming/ .
About the cloud, you can sign in on the cloud using the link `Sign in` and after to do the login, you will have in your left side the name of the computer and case you click on it, you will have the nodes available, this is the same place where everything slave connected will appear.

Any additional doubt, please, let us know.

ilyam8:
@duhzon 

Other simple option is 

> https://docs.netdata.cloud/web/gui/custom/#dash-multi-host-dashboard

duhzon:
Thank you! i was looking into netdata and notice that was a "node" button. I got all node linked by that way but i cant link node0 (the main node), because node option on localhost:19999 dont show up for node 0 (I already tried uptade netdata, all machines its running with newest version). 
But my question is further than that, after i got all it "nodded up" i have to got netdata showing main metrics of all nodes in a single dashboard monitor. I heard about modified dashboard but i really cant find a good tutorial.

ilyam8:
@tnyeanderson if you have time please ^^

tnyeanderson:
@duhzon These things can definitely be confusing, and I would love to improve the docs on this as I admit they are not as good as they should be. I just haven't had the time... I'll take a few minutes to help here, though:

[...]

Issue #6913, opened 2019-09-23
dash.html


eglyn:
Hello, 

I am interested by the multi-Host Dashboard (dash.html), but I can not display multiple dashboards on the same page ...
I follow the documentation witch says: 

> Change the following line in dash.html to reflect your URLs. The second URL is used if you access your netdata dashboard from a reverse proxy. The reverse proxy URL is optional, if it is not set then both will use the netdata host URL.

> var dash = new Dash('http://localhost:19999');

and add the dashboard.js to the head section and I see on my dash.html page my netdata server, but how add the others machines ?

documentation says: 
>dash-example.html is an all-in-one page that automatically fetches graphs from all your hosts. Just add your graphs and charts (or use the defaults) one time using the dash-* syntax and it will be automatically replicated for all of your hosts; showing alarms and graphs for all your hosts on one page!

what "using the dash-* syntax" means ?

thx in advance :)

cakrit:
@tnyeanderson wanna help with this and perhaps improve the docs and comments in the file?

tnyeanderson:
> I see on my dash.html page my netdata server, but how add the others machines ?

I may need to know a little bit more about your setup. `dash.html` works like this: If you have multiple "slave" instances that stream to a master netdata host, then once you set the URL to your master as described, it will fetch the list of the streamed "slave" hosts from the master. Then for *each* of these, it will generate the graphs and charts indicated by the `dash-*` syntax.

Are you hosting `dash.html` on your master? Or on another system?

> and add the dashboard.js to the head section

`dash.html` adds this path automatically when your master URL is set correctly, so this may be part of the problem. Don't add it twice! All that is required of you to get started is change the URL as shown, dash takes care of the rest :)

> what "using the dash-* syntax" means ?

If you look in the `dash.html` file, you will see elements with class names like `dash-graph` and `dash-chart`. These are what is replicated for each streamed host. The file comes preloaded with graphs/charts that are commonly used, but they can be changed to present the graphs that you need. More information regarding the `dash-*` syntax is available [here](https://github.com/netdata/netdata/tree/master/web/gui/custom), just beneath the part about changing the URL (starts with "If you want to change the graphs or styling to fit your needs"). It isn't very clear that this is talking about the dash syntax, so I will work on improving that documentation. Thank you for pointing that out!

[...]

Issue #7671, opened 2020-01-05
Community custom-made dashboard html´s


rdnsx:
Hi, 

maybe i´m just missing something, i thought would be obvious... but is there a place/pool were users can provide and/or get custom-made dashboard html´s made by the communitiy?

Cheers
rdnsx

ilyam8:
Hi, no, there is no such place unfortunately 😢

joelhans:
As Ilya said, there's currently no place for this, but I really like the idea.

One option is to create a new page in the documentation, almost like a gallery, with screenshots and links to the custom dashboard's code. I could link to that page from a few places throughout the documentation.

Or, I could use a Notion board, like this: https://www.notion.so/joelnetdata/Custom-dashboards-gallery-ccb2d839331b4dea925048a380bccdd6

Another option would be some tool/service that lets users submit their custom dashboards automatically. I'm not sure if such a tool exists, but I'd be happy to experiment if someone has an idea.

rdnsx:
Thanks for your joining. I like the notion board idea and links in the documentation. 
For me it's like:

1. I wanne see a preview, name and short description of each dashboard

2. Users must be able to upload the preview picture and html file themselfes

3. External links to download the file from are not allowed, to ensure downloadlinks are working at any time.

4. Maybe a rating possibility to make sure uploaded Dashboards are working.

joelhans:
2 is going to be the toughest to implement, but I'll continue looking into how we could implement this.

rdnsx:
Perfect. You know... For me there would be no bigger pain than finding the perfect template and the external link doesnt work, because the provider of the file cahnged whatever...

amoss:
This seems like a really good idea. There is something similar for Graphviz at https://graphviz.gitlab.io/gallery/ and it helps their community see what is possible and get ideas on how to implement things. This would definitely be great for Netdata.

tnyeanderson:
Check out `web/gui/dash-example.html`. See the README in `web/gui/custom` for instructions.

That's at least a start :) some discussion on this in #7757

cakrit:
I suggest we extend what we have as follows:
- The existing directory `web/gui/custom` gets subdirectories, one for each user-contributed custom dashboard. 
-  `web/gui/custom/README.md` lists those directories, with a short explanation of what each one contains, perhaps an image with a screenshot as mentioned above and a link to the README of the particular directory.
[...]

PR #7757, opened 2020-01-16
Small updates to dash.html

tnyeanderson:
##### Summary

* Improves documentation based on #7221 #6913
* Clarifies that `dash.html` is provided more as a courtesy or an experiment than a core part of netdata. Suggest netdata cloud for users that require more functionality
* Styling improvements! Sticky hostnames so you can see where you are no matter how far you've scrolled!
* Some code cleanup (formatting and whitespace, mostly)

##### Component Name

* dash.html

##### Additional Information

The update process is manual. Back up your current dash.html file, then copy the example file again and recreate any of the changes you had made when setting up dash.html the first time.

##### A Question

One of the things I like about dash.html is that it has only one component: the file itself. All the CSS, JS, and HTML are in the same file. This makes it portable as it simply queries the netdata API, independent of the file location.

However this approach does come with disadvantages: namely the update process.

If different components were stored in different files (perhaps in a dedicated directory), then user configuration could be stored independently and CSS/JS would update transparently to the user upon a netdata upgrade.

However, this does make it seem more "integrated" into the core of netdata, and users might come to expect the netdata team to support it. This was not my original intent, I simply wanted others to have access to my experiment. It is still not my intent, but I wanted to float the idea to the team and get your thoughts.

I figure that netdata wants to promote their cloud platform, which I also encourage. I understand that these changes could mean additional responsibility and possibly less traffic to the cloud platform, so it may not be best from a business perspective.

If the netdata team wants to make this change, I am happy to do so and submit another PR for it, I just want to get your input moving forward with full information.


tnyeanderson:
This is ready for review, sorry for the extra commits!

jacekkolasa:
@tnyeanderson thanks for you notes! You're right, it would make sense to keep things separated, to allow smooth upgrades. To avoid concerns about being it "officially supported" part, we could put it into some specific folder (`community`?). @cakrit any thoughts from the product perspective?

We'll quite soon merge the new Dashboard (which resides now in separate repository https://github.com/netdata/dashboard) and I would personally wait until we finish the migration.

Anyway, thanks for the maintenance work! :) Sticky is a good idea. Can i just ask if you change the indents back to spacebars? We have agreed on a [common linting rules](https://github.com/netdata/eslint-config-netdata/blob/master/.eslintrc.js), but i haven't put them yet to `netdata` repository (i plan to do that after switching to new dashboard)

tnyeanderson:
Okay the tabs should be (4) spaces now, sorry about that! I like the idea of a "community" folder. Other options could be:

* A separate "community" repository (I like this idea as it allows for these kinds of projects to be easier to maintain in the future)
* A separate repo for this project (don't like this idea as much, if at all)

At the end of the day I want to do whatever the netdata team thinks is best. Either way, this should be ready to merge!

joelhans:
I think we'll end up with a community dashboard repo that builds a very simple gallery site using something like Hugo. Users could then submit PRs with their custom dash HTML/CSS/JS, plus a screenshot and any other instructions (like you have here), to have them added to the gallery.

It's something I'm already exploring because of #7671, but it's admittedly a lower priority for me at the moment. If anyone has ideas, I'm open to anything.

web/gui/custom/README.md
```diff
@@ -74,37 +78,39 @@ First, rename the file so it doesn't get overwritten. For instance, with a webro
 cp /usr/share/netdata/web/dash-example.html /usr/share/netdata/web/dash.html
 ```
 
-Change the following line in `dash.html` to reflect your URLs. The second URL is used if you access your netdata dashboard from a reverse proxy. The reverse proxy URL is optional, if it is not set then both will use the netdata host URL.
+Find and change the following line in `dash.html` to reflect your netdata URLs. The second URL is only used if you access your netdata dashboard through a reverse proxy. The reverse proxy URL is optional; if it is not set then both will use the netdata host URL.
```
joelhans: Netdata should always be capitalized.

```suggestion
Find and change the following line in `dash.html` to reflect your Netdata URLs. The second URL is only used if you access your Netdata dashboard through a reverse proxy. The reverse proxy URL is optional; if it is not set then both will use the Netdata host URL.
```

web/gui/custom/README.md
```diff
@@ -65,7 +65,11 @@ header:
 
 ## Dash (Multi-Host Dashboard)
 
-`dash-example.html` is an all-in-one page that automatically fetches graphs from all your hosts. Just add your graphs and charts (or use the defaults) one time using the `dash-*` syntax and it will be automatically replicated for all of your hosts; showing alarms and graphs for all your hosts on **one page!**
+`dash-example.html` is an all-in-one page that automatically fetches graphs from all your hosts. Just add your graphs and charts (or use the defaults) one time using the `dash-*` syntax, and your selections will be automatically replicated for all of your hosts; showing alarms and graphs for all your hosts on **one page!**
+
+__**Dash will only work if you have implemented netdata streaming using `stream.conf`**__
+
+`dash-example.html` was created as an experiment to demonstrate the capabilities of netdata in a multi-host environment. If you desire more features, submit a pull request or check out netdata cloud!
```
joelhans:
```suggestion
`dash-example.html` was created as an experiment to demonstrate the capabilities of netdata in a multi-host environment. If you desire more features, submit a pull request or check out Netdata Cloud!
```

tnyeanderson:
@joelhans I think that's a great idea. If someone wants to initialize a `community` repo here I will take care of the leg work getting dash.html in there (with better structure, and a more unique name) and set up a general README and CONTRIBUTE to get it kick-started.

joelhans:
As you've seen in the other thread, @tnyeanderson, we may end up keeping these files within the core repo for now. But, I think we can move ahead with these changes, figure out the right location for custom dashboards, and then figure out the best way to move your dash code to that finalized location.

tnyeanderson:
Sounds good to me, let me know if/how I can help!

PR #7765, opened 2020-01-16
Add an environment variable check to Travis configuration to allow disabling nightlies.

Ferroin:
##### Summary

This allows us to control whether nightly builds run based on the presence or absence of an environment variable, instead of requiring us to remove and then re-add a scheduled build job. The primary advantage to using this approach is that nightlies will run at the same time each day, instead of moving to whenever they happen to be re-enabled.

The variable in question is named `RUN_NIGHTLY`, and must be defined with a value of 'yes' for nightly builds to run.

When this is merged, I will open a PR to update the release process documentation to reflect this change.

##### Component Name

area/ci

##### Additional Information

When an environment variable is not present, Travis will expand it to an empty string just like POSIX sh does, thus ensuring the check will fail if the environment variable is not defined.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,16 +47,16 @@ stages:
   - name: Build process
   - name: Artifacts validation
   - name: Artifacts validation on bare OS, stable to current lifecycle checks
-    if: branch = master AND (type = pull_request OR type = cron)
+    if: branch = master AND (type = pull_request OR (type = cron AND env(RUN_NIGHTLY) = yes))
 
     # Nightly operations
   - name: Nightly operations
-    if: branch = master AND type = cron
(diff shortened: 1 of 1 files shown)


Ferroin:
As a quick note, I've preemptively added the required environment variable to the Travis settings for this repository so that nightlies will run as normal when this gets merged.

knatsakis:
@Ferroin could you also update https://github.com/netdata/docs/blob/master/infrastructure/Releasing%20Netdata%20Agent.md regarding how to enable/disable the nightlies?

Ferroin:
> @Ferroin could you also update https://github.com/netdata/docs/blob/master/infrastructure/Releasing%20Netdata%20Agent.md regarding how to enable/disable the nightlies?

I was planning on doing so right after merging this.

PR #7768, opened 2020-01-16
Limit PR labeler runs to the main repo.

Ferroin:
##### Summary

This limits runs of the PR labeler to the main repo only, which should cut down on noise people get from GitHub.

##### Component Name

area/ci

##### Additional Information

Tested on personal fork that this prevents the labeler from actually trying to do anything, as well as verifying that it does indeed match repos correctly.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -9,6 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: paulfantom/periodic-labeler@master
+        if: github.repository == 'netdata/netdata'
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: paulfantom/periodic-labeler@master
+        if: github.repository == 'netdata/netdata'
```
prologic: Ahh lovely :)

prologic: So Github Actions YAML is Turning Complete? 🤣

PR #7770, opened 2020-01-16
Issue 7488 docker labels

### LINKED ISSUE #7488: RFC: "Host" labels in docker containers
In #7400, we are taking some of the environment parameters auto-detected by netdata and converting into "host" labels. However, in a docker container, looking at the container we're running in for these parameters may well be the wrong thing to do.

Users run netdata in a docker container on their system, in order to monitor **the host system**, not the container itself. Our whole internal definition of `host` is misleading here. Yes, the container itself can be considered a `host` by its own right, but what's the point? If we were to switch our focus and acknowledge that the container's only purpose is to monitor the host machine that runs docker, the netdata docker instance itself should be no more than just another cgroup on that host. This "point of view shift" is already there for everything that we get from the host's `var` and `sys`. So when I look at the dashboard of a netdata running on docker, I may be seeing on the URL whatever was specified for that container, but the key charts themselves correctly show me what's happening on the host machine. We can't directly access packaged applications on that host machine, so auto-discovery for a mysql server would not work, but we at least provide the info users need the most.

Getting back to the environment parameters mentioned earlier, netdata currently populates that information by executing [system-info.sh](https://github.com/netdata/netdata/blob/master/daemon/system-info.sh). So it's just looking at the info of the container itself. While useful for statistical and possibly later for self-diagnosing purposes, this information shouldn't be used to label a netdata's charts, when running as a docker container. Instead, we should try to find a way to get the host's "system info", if at all possible. 

Further down the line, we should attempt to retrieve [additional labels from the docker metadata](https://docs.docker.com/config/labels-custom-metadata/), if a user has provided netdata access to the docker socket. Alternative ways may be possible in specific use cases, such as with [Amazon ECS container metadata](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-metadata.html). We'll need some assistance identifying such alternatives. However, we need to be clear that those labels should only be attached to the monitored cgroups (including the netdata container) and not to the host itself. 

We will have an internal discussion on these matters and would like to also request assistance from the community.

amoss:
##### Summary
Fixes #7488 

##### Component Name
daemon
web

##### Additional Information
Still to do before completion:
* Put in the virtualization variable via docker (run and compose)
* Check that I have not broken the GA.


daemon/main.c
```diff
@@ -731,6 +731,11 @@ static int load_netdata_conf(char *filename, char overwrite_used) {
     return ret;
 }
 
+// coverity[ +tainted_string_sanitize_content : arg-0 ]
+inline void coverity_remove_taint(char *s)
+{
```
thiagoftsm: Please add inside this function `(void)s;` to remove the warning

```
daemon/main.c: In function 'coverity_remove_taint':
daemon/main.c:735:41: warning: unused parameter 's' [-Wunused-parameter]
  735 | inline void coverity_remove_taint(char *s)
```

when Netdata is compiled with the following command line:

```
CFLAGS="-O1 -ggdb -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1" ./netdata-installer.sh --disable-go --disable-lto --dont-wait --dont-start-it
```

daemon/system-info.sh
```diff
@@ -8,111 +8,182 @@ KERNEL_VERSION="$(uname -r)"
...
-# detect the operating system
+# detect the virtualization
 
-OS_DETECTION="unknown"
-NAME="unknown"
-VERSION="unknown"
-VERSION_ID="unknown"
-ID="unknown"
-ID_LIKE="unknown"
+if [ -z ${VIRTUALIZATION} ]; then
```
thiagoftsm: Please add double quote around "${VIRTUALIZATION}" to remove the following warning from `shellcheck` and, of course, to avoid problems with variable content:

```
In daemon/system-info.sh line 13:
if [ -z ${VIRTUALIZATION} ]; then
        ^---------------^ SC2086: Double quote to prevent globbing and word splitting.
```

daemon/system-info.sh
```diff
@@ -8,111 +8,182 @@ KERNEL_VERSION="$(uname -r)"
...
+            CONTAINER="$(systemd-detect-virt -c)"
+            CONT_DETECTION="systemd-detect-virt"
+    else
+            if grep -q "^flags.*hypervisor" /proc/cpuinfo 2>/dev/null; then
+                    VIRTUALIZATION="hypervisor"
+                    VIRT_DETECTION="/proc/cpuinfo"
+            elif [ -n "$(command -v dmidecode)" ]; then
+                    # Virtualization detection from https://unix.stackexchange.com/questions/89714/easy-way-to-determine-virtualization-technology
+                    # This only works as root
+                    if dmidecode -s system-product-name 2>/dev/null | grep -q "VMware\|Virtual\|KVM\|Bochs"; then
```
thiagoftsm: Please, take a look in the first answer of this link `https://unix.stackexchange.com/questions/89714/easy-way-to-determine-virtualization-technology`, I think we can add more expected results for `grep`.

daemon/system-info.sh
```diff
@@ -8,111 +8,182 @@ KERNEL_VERSION="$(uname -r)"
...
+CONTAINER_NAME="unknown"
+CONTAINER_VERSION="unknown"
+CONTAINER_VERSION_ID="unknown"
+CONTAINER_ID="unknown"
+CONTAINER_ID_LIKE="unknown"
+
+if [ "${KERNEL_NAME}" = "Darwin" ]; then
+        # Mac OS
+        OIFS="$IFS"
+        IFS=$'\n'
```
thiagoftsm: Another `shellcheck` warning for this:

```
In daemon/system-info.sh line 92:
        IFS=$'\n'
            ^---^ SC2039: In POSIX sh, $'..' is undefined.
```

daemon/system-info.sh
```diff
@@ -8,111 +8,182 @@ KERNEL_VERSION="$(uname -r)"
...
+        set $(sw_vers) > /dev/null
```
thiagoftsm:
```
In daemon/system-info.sh line 93:
        set $(sw_vers) > /dev/null
            ^--------^ SC2046: Quote this to prevent word splitting.
```

daemon/system-info.sh
```diff
@@ -8,111 +8,182 @@ KERNEL_VERSION="$(uname -r)"
...
+        NAME=$(echo $1 | tr "\n\t" '  ' | sed -e 's/ProductName:[ ]*//' -e 's/[ ]*$//')
```
thiagoftsm:
```
In daemon/system-info.sh line 94:
        NAME=$(echo $1 | tr "\n\t" '  ' | sed -e 's/ProductName:[ ]*//' -e 's/[ ]*$//')
                    ^-- SC2086: Double quote to prevent globbing and word splitting.
```

daemon/system-info.sh
```diff
@@ -8,111 +8,182 @@ KERNEL_VERSION="$(uname -r)"
...
+        VERSION=$(echo $2 | tr "\n\t" '  ' | sed -e 's/ProductVersion:[ ]*//' -e 's/[ ]*$//')
```
thiagoftsm:
```
In daemon/system-info.sh line 95:
        VERSION=$(echo $2 | tr "\n\t" '  ' | sed -e 's/ProductVersion:[ ]*//' -e 's/[ ]*$//')
                       ^-- SC2086: Double quote to prevent globbing and word splitting.
```

daemon/system-info.sh
```diff
@@ -8,111 +8,182 @@ KERNEL_VERSION="$(uname -r)"
...
+        ID="mac"
+        ID_LIKE="mac"
```
thiagoftsm:
``` 
In daemon/system-info.sh line 97:
        ID_LIKE="mac"
        ^-----^ SC2034: ID_LIKE appears unused. Verify use (or export if used externally).
```

daemon/main.c
amoss: Done.

daemon/system-info.sh
amoss: That's where I took the code from :) The other answers require parsing the output of different sources. I've added it to the description of the follow-up issue that @Ferroin is going to work on as he knows about which parts will need extra protection (such as the cases where `dmidecode` can hang a VM). I added the comment into the script as a marker.

daemon/system-info.sh
amoss: The code around that makes no sense at all. I'll take a look at what it should do on a mac.

daemon/system-info.sh
amoss: Fixed.

daemon/system-info.sh
amoss: Rewrote this block to get rid of the weird stuff.

daemon/system-info.sh
amoss: Cleaned up.

daemon/system-info.sh
amoss: Cleaned up.

daemon/system-info.sh
amoss: Fixed.

daemon/system-info.sh
thiagoftsm: All right :)

streaming/rrdpush.c
```diff
@@ -581,12 +581,12 @@ static int rrdpush_sender_thread_connect_to_master(RRDHOST *host, int default_po
...
-              , (host->system_info->os_id_like) ? host->system_info->os_id_like : ""
-              , (host->system_info->os_version) ? host->system_info->os_version : ""
-              , (host->system_info->os_version_id) ? host->system_info->os_version_id : ""
-              , (host->system_info->os_detection) ? host->system_info->os_detection : ""
+              , (host->system_info->host_os_name) ? host->system_info->host_os_name : ""
+              , (host->system_info->host_os_id) ? host->system_info->host_os_id : ""
+              , (host->system_info->host_os_id_like) ? host->system_info->host_os_id_like : ""
+              , (host->system_info->host_os_version) ? host->system_info->host_os_version : ""
+              , (host->system_info->host_os_version_id) ? host->system_info->host_os_version_id : ""
+              , (host->system_info->host_os_detection) ? host->system_info->host_os_detection : ""
```
thiagoftsm: Hi @amoss ,

After you review the code and I verified that everything was fixed, I began to test the PR, and I detected a problem with the stream. When I executed a request to `https://localhost:19999/host/darkstar/api/v1/info` on the master, I verified that the system identified correctly that I am running on Virtual Box

err0

on the other hand, I lost completely the OS description

err1

but when I call the `system-info.sh` direct from the terminal I can see the os information

err3

To confirm that this is a problem with this PR, I recompiled Netdata again on both sides(master and slave) using another branch and I could verify that the slave is streaming is working correctly.

amoss: This looks quite serious - I'll setup some tests and check what has happened.

amoss: I've had a look and in order to fix it I need to change the streaming protocol. The current streaming protocol is too closely connected to the internal structure of the `system_info` structure, and this structure has changed so that the field names communicated in the initial `STREAM` request no longer exist.

The `STREAM` request needs to communicate both the host and container versions of these fields, so that we can see the correct info in API requests and also in the GA results taht we deliver. What I propose is:

https://github.com/netdata/netdata/blob/6ba71e1704ef9fcd9335e1e4e1ad844b00a082fd/streaming/rrdpush.c#L558-L571'

* Change these to the HOST_ and CONTAINER_ versions.

https://github.com/netdata/netdata/blob/6ba71e1704ef9fcd9335e1e4e1ad844b00a082fd/streaming/rrdpush.c#L1398-L1401

* Add a check here that if we are failing on `SYSTEM_` create an error that the slave must be upgraded to the same version as the master for compatibility.

To some extent our hand are tied because we are communicating the names of internal field values in a protocol. I think it is reasonable to force an upgrade to the same version where necessary.

It's your AoR in the code - what would you like us to do?

thiagoftsm: We can do the changes that you are suggesting, but we also need to keep in our mind backward compatibility. 
Case we are saying that is necessary to change the variable names, this will be a major change in the stream protocol, so it will be necessary to do something like I did on #7549 , as you can see there, we will need to do the following changes at `streaming/rrdpush.c`:

- Create a new funny message (This is the most important step);
- Change the protocol version (line 28 );
- Add a new if at line 650. Probably it is  a good time to move the lines `651 - 662` for a function
- We also need to change `rrdpush_receive` to accept a third kind of message, probably lines 1151-1158 are good candidate to become another function.
- Finally we need to do a last change inside the function `rrdpush_receiver_thread_spawn`, among the lines 1395 - 1402

Now we need to change the function `rrdhost_set_system_info_variable` inside `database/rrdhost.c`. The changes are

-  We need more one argument for the function that will identify the protocol version
- Depending of the version, we will use specific variable name to parse.

Of course this also brings the necessity to change `database/rrd.h`

Best regards!

amoss: Created a new issue #7804 to track the work in updating the protocol.

thiagoftsm:
The outputs now are almost there, but I think it is necessary to do more a change, please, take a look firstly in the local machine output:

local

and now the slave output that runs on Virtual Box

remote

I would expect that they would have the same fields somehow on the slave side some fields are removed. What is more strange to me is the fact that we eliminated the container variables. I do not think this is an error, but I think they should be equal.

amoss:
@thiagoftsm This is correct - I left some notes in the code about where to put the new fields after we can upgrade the protocol version.

PR #7777, opened 2020-01-17
Drop dirty dbengine pages if disk cannot keep up

### LINKED ISSUE #7268: Collectors block if dbengine page cache is full of dirty buffers
##### Bug report summary
The netdata collectors block if dbengine page cache is full of dirty buffers, and as a result metrics stop being gathered until the problem is resolved. This can happen if there is a huge number of metrics being gathered, or the disk is too slow or too busy, or if the disk has gone bad.

netdata v1.18.1
##### Component Name
database/engine
##### Expected behavior
New metrics should never stop being collected, old dirty buffers should be dropped in favour of continuous data collection.

### LINKED ISSUE #7468: Repeated dbengine global flushing errors
##### Bug report summary
No matter how I setup dbengine, I still get `10min dbengine global flushing errors` alerts during periods of moderate to high server usage. I have tried putting the dbengine files on an HDD, SSD, and even a RAMDISK without being able to solve the issue. I have increased the page cache size repeatedly without effect. 

##### OS / Environment
Docker

##### Netdata Version
`netdata v1.19.0-53-g7953472f`

##### Component Name
dbengine

##### Steps To Reproduce
Current settings:
```
[global]
    page cache size = 128
    dbengine disk space = 512
````
##### Expected behavior
No errors. Netdata should be able to handle moderate server utilization without being unable to write to disk.

### PR DESCRIPTION
**Pull request by mfundul**
##### Summary
Fixes #7268 
Fixes #7468 
##### Component Name
database/engine
##### Additional Information
Introduce dirty page pressure handling in the dbengine page cache that invalidates pages when the disk cannot keep up with the flushing speed. Make sure progress is made by managing real threads instead of relying on libuv workers since the workers will be exhausted by internal I/O queuing.

### CODE REVIEW

File: database/engine/rrdengine.c [hunk 1]
```diff
@@ -248,6 +249,109 @@ static void do_commit_transaction(struct rrdengine_worker_config* wc, uint8_t ty
...
+    Word_t Index;
+    unsigned nr_committed_pages;
+
+    do {
+        uv_rwlock_wrlock(&pg_cache->committed_page_index.lock);
+        for (Index = 0,
+             PValue = JudyLFirst(pg_cache->committed_page_index.JudyL_array, &Index, PJE0),
+             descr = unlikely(NULL == PValue) ? NULL : *PValue;
+
+             descr != NULL;
```
thiagoftsm: Only to confirm, did you insert the previous test to avoid compilation warning?

database/engine/rrdengine.c
```diff
@@ -503,15 +608,24 @@ static void after_delete_old_data(uv_work_t *req, int status)
...
+
+    /* interrupt event loop */
+    uv_stop(wc->loop);
 }
 
-static void delete_old_data(uv_work_t *req)
+static void delete_old_data(void *arg)
 {
-    struct rrdengine_instance *ctx = req->data;
+    struct rrdengine_instance *ctx = arg;
```
thiagoftsm: With gcc 9.2 I did not have warnings or errors to compile the code, but should we not add a explicit cast to `struct rrdengine_instance *` here to avoid warnings with old compilers?

database/engine/rrdengine.c
thiagoftsm: Case the first answer is `yes`, why are you using this comparison instead `(void)desc` that is the most common code used to avoid this warning in Netdata code?

database/engine/rrdengine.c
mfundul: It's one of the differences between C and C++, there should be no warning here.

https://stackoverflow.com/questions/1736833/void-pointers-difference-between-c-and-c

database/engine/rrdengine.c
mfundul: This is the test condition of the `for` loop, it's the middle part of it.

database/engine/rrdengine.c
thiagoftsm: All right.

amoss:
I've started reviewing the code - but this will take a few days. Last time this code was changed we saw some subtle bugs so I need to read around it more this time. Conveniently I'm examining the dbengine code for the roadmap task anyway so there is some overlap...

caller-graph:
image

call-graph:
image

database/engine/rrdengineapi.c
```diff
@@ -725,9 +738,11 @@ void rrdeng_get_35_statistics(struct rrdengine_instance *ctx, unsigned long long
     array[30] = (uint64_t)global_io_errors;
     array[31] = (uint64_t)global_fs_errors;
     array[32] = (uint64_t)rrdeng_reserved_file_descriptors;
-    array[33] = (uint64_t)ctx->stats.flushing_errors;
-    array[34] = (uint64_t)global_flushing_errors;
-    assert(RRDENG_NR_STATS == 35);
+    array[33] = (uint64_t)ctx->stats.pg_cache_over_half_dirty_events;
```
amoss: There is a comment about that changing the indices can break user code. Does changing the use of index 33 from the number of flushing errors to the number of over_half_dirty events affect any user code?

mfundul: All corresponding statistics have been adjusted. The hazard of memory corruption is about the number of the actual statistics. Changing individual statistics only changes the semantics.

ilyam8:
this #7777 is lucky one we need to merge it asap 😄

database/engine/rrdengineapi.c
amoss: Ah ha. I see that you have changed the number of the actual statistics as well...

PR #7778, opened 2020-01-17
Fix libuv IPC pipe cleanup problem

### LINKED ISSUE #7734: Port in use after uninstall from Ubuntu
##### Bug report summary
After installing Netdata from the Ubuntu repository and then uninstalling it and installing it using the bash script, the netdata service always fails to start with the following error in the logs:
```
2020-01-14 21:32:03: netdata ERROR : MAIN : uv_pipe_bind(): address already in use (errno 98, Address already in use)
netdata: daemon/commands.c:543: command_thread: Assertion `0 == uv_loop_close(loop)' failed.
```

Even though the output of `sudo lsof -i -P -n | grep 19999` is blank, and so is `ps fax | grep 19999`. The output of `netstat -nat | grep LISTEN | grep 19999` is also blank.

I have uninstalled the Ubuntu package using the —purge flag to no effect and I’ve uninstalled and reinstalled using your bash script twice with the same error. The install does not have any failures except when starting the service (due to the above issue).

I also have adjusted the `netdata.conf` to contain the following:
```
[web]
  default port = 30002
```
As just a random port I know is open, and even then the process still fails to start with systemctl.

Note that netdata did work properly with the Ubuntu package, it’s only when trying to switch over did this issue arise.

Linux discus-spark1 4.15.0-72-generic #81-Ubuntu SMP Tue Nov 26 12:20:02 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 18.04.3 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="18.04.3 LTS (Bionic Beaver)"
/etc/os-release:ID=ubuntu
/etc/os-release:PRETTY_NAME="Ubuntu 18.04.3 LTS"

netdata v1.19.0-314-nightly

mfundul:
Fixes #7734 

When netdata fails to delete any pre-existing unix domain sockets due to permission or other reasons it should gracefully disable the command server. Moreover, the uninstaller should help clean up given sufficient permission.

### DIFF
--- a/daemon/commands.c
+++ b/daemon/commands.c
@@ -540,6 +540,7 @@ error_after_pipe_bind:
 error_after_pipe_init:
     uv_close((uv_handle_t *)&async, NULL);
 error_after_async_init:
+    uv_run(loop, UV_RUN_DEFAULT);
     assert(0 == uv_loop_close(loop));
 error_after_loop_init:
     freez(loop);
@@ -587,7 +588,7 @@ void commands_init(void)
     return;
 
 after_error:
-    error("Failed to initialize command server.");
+    error("Failed to initialize command server. The netdata cli tool will be unable to send commands.");
 }
 
 void commands_exit(void)
--- a/daemon/commands.c
+++ b/daemon/commands.c
@@ -526,7 +526,7 @@ static void command_thread(void *arg)
     info("Shutting down command event loop.");
     uv_close((uv_handle_t *)&async, NULL);
     uv_close((uv_handle_t*)&server_pipe, NULL);
-    uv_run(loop, UV_RUN_DEFAULT);
+    uv_run(loop, UV_RUN_DEFAULT); /* flush all libuv handles */
 
     info("Shutting down command loop complete.");
     assert(0 == uv_loop_close(loop));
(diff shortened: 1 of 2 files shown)


daemon/commands.c
```diff
@@ -540,6 +540,7 @@ static void command_thread(void *arg)
 error_after_pipe_init:
     uv_close((uv_handle_t *)&async, NULL);
 error_after_async_init:
+    uv_run(loop, UV_RUN_DEFAULT);
```
amoss: Have you put this in to force the close callback to occur before the loop close?

mfundul: Exactly.

amoss: It makes sense in the context of this bug fix, but it will be quite difficult for A Random Person to see that when they read the code without the context. In particular, any libuv code that exists now will get used as reference when people are porting other sub-systems such as the web server. Could you add a short comment just to say why it is there?

mfundul: Done.

PR #7780, opened 2020-01-17
Fix a Coverity issue

vlvkobal:
##### Summary
Fix an issue with an unchecked return value.

##### Component Name
exporting engine

### DIFF
--- a/exporting/init_connectors.c
+++ b/exporting/init_connectors.c
@@ -64,7 +64,8 @@ int init_connectors(struct engine *engine)
             }
 
             // dispatch the instance worker thread
-            uv_thread_create(&instance->thread, connector->worker, instance);
+            if (uv_thread_create(&instance->thread, connector->worker, instance))
+                return 1;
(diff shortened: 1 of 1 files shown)


exporting/init_connectors.c
```diff
@@ -64,7 +64,8 @@ int init_connectors(struct engine *engine)
             }
 
             // dispatch the instance worker thread
-            uv_thread_create(&instance->thread, connector->worker, instance);
+            if (uv_thread_create(&instance->thread, connector->worker, instance))
```
thiagoftsm: A suggestion, since the version 1.4.1(http://docs.libuv.org/en/v1.x/threading.html#c.uv_thread_create)  `uv_thread_create` returns UV_E* errors, so we could use `uv_strerror`  to give more details about the error that happened when we were trying to create a thread, could you add a call for `uv_strerror` for we store the information inside `error.log`?

vlvkobal: Done.

thiagoftsm: Thank you! :)

PR #7789, opened 2020-01-18
Fixes shellcheck in CI for PRs

prologic:
##### Summary

Fixes running `shellcheck` in PR(s) in CI confirmed in #7488 and #7787 and found by @thiagoftsm 

This _may_ not be 100% correct here; but the rationale is:

- Check all `*.sh` and `*.sh.in` shell scripts.
- Check all `*.sh` shell scripts against after the build process.

The scripts in `*.sh.in` are templated; so we check them twice.

##### Component Name

area/ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,8 +98,10 @@ jobs:
       # Do code quality, syntax checking and other pre-build activities
     - stage: Code quality, linting, syntax, code style
 
-      name: Run shellchecking on BASH
-      script: shellcheck --format=gcc $(find . -name '*.sh.in' -not -iwholename '*.git*')
+      name: Run shellchecking on Shell Scripts
+      script:
+        - shellcheck --format=gcc $(find . -name '*.sh.in' -not -iwholename '*.git*')
+        - shellcheck --format=gcc $(find . -name '*.sh' -not -iwholename '*.git*')
 
       # Check if any of the C code deviates from our coding style
     - name: Check coding style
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -149,6 +151,10 @@ jobs:
       after_failure: post_message "TRAVIS_MESSAGE" "'make dist' failed"
 
 
+    - name: Re-check Shell Scriopts (after build) with shellcheck
```
thiagoftsm: There is a typo problem here `Scriopts`, that does not affect the main goal of the PR, so I will approve it, and case you change this, I will approve again.

.travis.yml
```diff
@@ -149,6 +151,10 @@ jobs:
...
+      script:
+        - shellcheck --format=gcc $(find . -name '*.sh' -not -iwholename '*.git*')
```
Ferroin: Travis doesn't share state between stages, so this will just do the same thing as line 104.

.travis.yml
prologic: Fixed 🤣

prologic:
> As things stand right now, this breaks the build, significantly. That's the reason that #7640 was so damn big, I was trying to enable shellcheck for everything and had to fix everything so that it didn't break CI for everyone.

Ahh but of course!

We can't do this; not like this. I'll come up with a different solution for @thiagoftsm and the rest of the team that doesn't horribly break the world.

.travis.yml
prologic: That wasn't the case being covered here; it was the case where template shell scripts go from `*.sh.in` to the end result `*.sh` and are re-checked in case the templating broke them in some way :)

Ferroin: That still doesn't work though, because each individual job operates on a clean, completely fresh clone of the source repo. So to do this checking, you'd have to run the build in this job and then run the shellcheck command.

prologic: Old; took a different route :)

Issue #5983, opened 2019-05-09
[Binary releases] Redesign versioning scheme on netdata


paulkatsoulakis:
##### Feature idea summary
During our implementation of packaging infrastructure with netdata, we bumped into a problem with versioning in our repository. Debian packaging tools do not accept the letter "v" as part of the version, it rather expects a very specific structure on the folder and tarball filenames.

We handle this at the packaging process for now, but as part of this story we want to eliminate v letter, so that our codebase is more compliant with our packaging infrastructure

##### Expected behavior
Versioning no longer includes 'v' anywhere, without breaking any existing implementation or process

paulkatsoulakis:
Things to check:
* What happens to the update process from dashboard (we have a check to notify the user for new versions)
* What happens between packaging version and kickstart install

Saruspete:
Hello,

i'm having a bad time trying to update from 1.13 to v1.15 : I've created the RPM from the provided tools (contrib/rhel/build-netdata-rpm.sh) and the generated RPM indeed have a "v" prefix. This messes up the upgrade, and yum finds the 1.13 is more recent than v1.15

May I help you to push this issue forward ?

Thanks,

cakrit:
For any change, ensure that we respect the fact that older installations will still have the "v" in front of them. The following are the ones I could find/think of:

The UI version check is done here:
https://github.com/netdata/netdata/blob/master/web/gui/main.js#L2870
https://github.com/netdata/netdata/blob/master/web/gui/main.js#L2871
These need a check if the string starts with "v", to get rid of it. Now they assume it does.

The daemon gets the version from either `git describe` or `packaging/version`, not quite sure:
https://github.com/netdata/netdata/blob/master/configure.ac#L8
https://github.com/netdata/netdata/blob/master/configure.ac#L9
Regardless, the CI process that produces the tag can just remove the "v" and it should work.

https://github.com/netdata/helmchart/blob/master/Chart.yaml#L5
Easy fix, we just change it when we get a new tag for 1.16.0, without the "v"

The updater doesn't care about the format, it does a sha256sum check.

https://github.com/netdata/netdata/blob/master/daemon/anonymous-statistics.sh.in#L26
Needs to change if the X.Y.Z-XXX format changes to X.Y.Z.XXX. The leading v is removed anyway with the `tr`

May want to reverse #5440, but not necessary.

paulkatsoulakis:
> Hello,
> 
[...]

PR #7794, opened 2020-01-19
Fixes Source0 URL in RPM spec

### LINKED ISSUE #7141: Packaging: tag Source0 in spec file seems to have wrong URL
In https://github.com/netdata/netdata/blob/master/netdata.spec.in#L90 the Source0 field does not evaluate to an existing URL. 

Take netdata 1.18.1 for example:
```
Name: netdata
Version: 1.18.1
Source0: https://github.com/netdata/%{name}/releases/download/%{version}/%{name}-%{version}.tar.gz
```
The URL evaluates to:
https://github.com/netdata/netdata/releases/download/1.18.1/netdata-1.18.1.tar.gz which does not exist.

I suggest to change Source0 to:
```
Source0: https://github.com/netdata/netdata/archive/v%{version}/%{name}-%{version}.tar.gz
```

Example result: https://github.com/netdata/netdata/archive/v1.18.1/netdata-1.18.1.tar.gz

prologic:
##### Summary

Fixes the `Source0` URL in the `netdata.spec.in` RPM spec file.

Adds the missing `v` prefix to the version strings.

Fixes #7141 

##### Component Name

area/packaging

### DIFF
--- a/netdata.spec.in
+++ b/netdata.spec.in
@@ -87,7 +87,7 @@ Version:	%{version}
 Release:	1%{?dist}
 License:	GPLv3+
 Group:		Applications/System
-Source0:	https://github.com/netdata/%{name}/releases/download/%{version}/%{name}-%{version}.tar.gz
+Source0:	https://github.com/netdata/%{name}/releases/download/v%{version}/%{name}-v%{version}.tar.gz
 URL:		http://my-netdata.io
 
 # #####################################################################


Ferroin:
Have we investigated why the current version of the spec file works in our CI package builds yet? I'm hesitant to approve this unless we can conclusively determine that it won't break CI packaging.

knatsakis:
> Have we investigated why the current version of the spec file works in our CI package builds yet? I'm hesitant to approve this unless we can conclusively determine that it won't break CI packaging.

I thought that this field is informational only, but I could be mistaken.

prologic:
> > Have we investigated why the current version of the spec file works in our CI package builds yet? I'm hesitant to approve this unless we can conclusively determine that it won't break CI packaging.

No we haven't and we need to; becuase we're changing the way we build our packages and I can't for the life of me figure out how we go from a `netdata.spec.in` to `netdata.spec` and who or what is responsible for swapping out `%{name}` and `%{version}` for their proper values :/

> 
> I thought that this field is informational only, but I could be mistaken.

I _thought_ so too but 🤷‍♂️Its possible to be used be tools/scripts I guess?

Ferroin:
> > > Have we investigated why the current version of the spec file works in our CI package builds yet? I'm hesitant to approve this unless we can conclusively determine that it won't break CI packaging.
> 
> No we haven't and we need to; becuase we're changing the way we build our packages and I can't for the life of me figure out how we go from a `netdata.spec.in` to `netdata.spec` and who or what is responsible for swapping out `%{name}` and `%{version}` for their proper values :/

`netdata.spec.in` to `netdata.spec` is currently GNU autotools templating. There's only one value that gets modified (`@PACKAGE_VERSION@`, right near the top), but that will be handled with `sed` in the new build system.

All the `%{foo}` stuff is variable expansion within the RPM file itself, the build environment provides all (or at least most) of these itself, and we define the other few ourselves (the `%global` and `%define` lines near the top of the file).
> 
> > I thought that this field is informational only, but I could be mistaken.
> 
> I _thought_ so too but Its possible to be used be tools/scripts I guess?

Unless I'm mistaken, if the sources can't be found under the rpmbuild directory, the build process will try to fetch them from this URL.

This doesn't matter in the new packaging flow that we're working on right now, as we bypass the source unpacking stuff to just use the local copy of the sources directly (`rpmbuild` expects to get a tarball which it then unpacks, but we already have the sources locally so that's just a waste of time).

prologic:
> https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-a-docker-container-action

Okay fair enough. I guess we're switching the templating from `autotools` to `sed` for simplicity?

prologic:
> All the `%{foo}` stuff is variable expansion within the RPM file itself, the build environment provides all (or at least most) of these itself, and we define the other few ourselves (the `%global` and `%define` lines near the top of the file).

In _some_ unrelated testing the other day I actually found this wasn't working as expected. Lines with `%{name}` and `%{version}` were just getting removed silently from the resulting `netdata.spec` file.

Saruspete:
Greetings,

I tried to generate the package using my usual scripts that removes the "v" prefix from the version, and this time the rpmbuild failed due to trying to search for 

In a previous issue, I asked to remove the "v" prefix, as it messes up pretty much all version comparison tests. So I searched how the script generates the version. Let's put it here (maybe also in the documentation)

1) `configure.ac` sets `VERSION_STRING` from `git describe --tags` to find the latest tag in the current branch. That leads to something like `v1.20.0`
2) `configure.ac` sets its variable `PACKAGE_RPM_VERSION` to `VERSION_STRING`. This `PACKAGE_RPM_VERSION` is used in multiple `.in` scripts.
3) `configure.ac` calls `AC_INIT with `VERSION_STRING`, which will automagically define `PACKAGE_VERSION` [See autoconf doc](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Initializing-configure.html)
4) `netdata.spec.in` uses the variable `PACKAGE_VERSION`
5) `build-netdata-rpm.sh` extracts the version with `grep PACKAGE_VERSION config.h`

So, the `v` prefix is already contained in the "version" variable, and should not be added within the URLs. I got this double `vv` issue with the following sequence:

```
git clone https://github.com/netdata/netdata.git netdata.testspec
cd netdata.testspec
git reset --hard v1.20.0
autoreconf -ivf
./configure
rpmbuild -ba ./netdata.spec
```

ping @Ferroin  @prologic  @knatsakis

Ferroin:
> Greetings,
> 
> I tried to generate the package using my usual scripts that removes the "v" prefix from the version, and this time the rpmbuild failed due to trying to search for
> 
> In a previous issue, I asked to remove the "v" prefix, as it messes up pretty much all version comparison tests. So I searched how the script generates the version. Let's put it here (maybe also in the documentation)
> 
> 1. `configure.ac` sets `VERSION_STRING` from `git describe --tags` to find the latest tag in the current branch. That leads to something like `v1.20.0`
> 2. `configure.ac` sets its variable `PACKAGE_RPM_VERSION` to `VERSION_STRING`. This `PACKAGE_RPM_VERSION` is used in multiple `.in` scripts.
> 3. `configure.ac` calls `AC_INIT with `VERSION_STRING`, which will automagically define `PACKAGE_VERSION` [See autoconf doc](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Initializing-configure.html)
> 4. `netdata.spec.in` uses the variable `PACKAGE_VERSION`
> 5. `build-netdata-rpm.sh` extracts the version with `grep PACKAGE_VERSION config.h`
> 
> So, the `v` prefix is already contained in the "version" variable, and should not be added within the URLs. I got this double `vv` issue with the following sequence:
> 
> ```
> git clone https://github.com/netdata/netdata.git netdata.testspec
> cd netdata.testspec
> git reset --hard v1.20.0
> autoreconf -ivf
> ./configure
> rpmbuild -ba ./netdata.spec
> ```
> 
> From the output of the build log, this instance misses the `v` as it's referenced `netdata-1.19.441-1.el7.x86_64`, with first occurence line 1725
> 
> ```
> Version set to 1.19.441 from v1.19.0.441.latest
> ```
> 
> That line seems to be produced in [.travis/package_management/common.py](https://github.com/netdata/netdata/blame/279b583a6f8bbbb96ba2b55f37bc1ffe6f9fd81b/.travis/package_management/common.py#L24-L26) where the [previous version](https://github.com/netdata/netdata/blame/cbb0528d32576d146920cfdbc3964644a39c5d11/.travis/package_management/common.py) didn't remove the `v`
> 
> ping @Ferroin @prologic @knatsakis

Ah, that's what I was failing to remember. I could have sworn there was some reason we had things the way they were, but for the life of me I couldn't find any.

@Saruspete I'll bring this up directly in the SRE meeting today.

Saruspete:
Thanks. The issue I was thinking about was #5983

prologic:
@Ferroin I don't understand #7141 anymore; did we somehow change the way `%{version}` is computed since `v1.18.1`? It would appear that the previous `Source0` template variable is now correct. All tags have always had the `v` prefix; so the only thing that might have changed since `v1.18.x` to now is how we compute the version string when building the RPM packages?

Ferroin:
@prologic I'm honestly not sure.

Did we every verify that #7141 was actually a problem outside of what the OP there was trying to do before we went to fix this? I don't recall seeing anything about that in the issue itself, and I'm starting to suspect that the issue was something with how the OP was trying to build the RPM packages, not our own code. Usually RPM versions do not include a `v` at the beginning, and the OP of #7141 may have not realized that our code breaks this convention and requires the `v` at the beginning of the version string.

Honestly though, unless @Saruspete has an issue with it, I'd like to fix _our_ RPM build code to let us keep this PR, since it brings us inline with standard practices for version handling in RPM files.

Saruspete:
The issue here is with this PR, you create a split between the version in git tag (v1.20.0) and the version somewhere else. That is, you remove the "v" in your [travis/package_management/common.py](https://github.com/netdata/netdata/blame/279b583a6f8bbbb96ba2b55f37bc1ffe6f9fd81b/.travis/package_management/common.py#L24-L26) to add it back on the RPM.

Either you remove the v in the git tag and remove it everywhere, either you keep it all the way.
Currently, I can't build the RPM with the provided scripts. It only works on your own travis chain, but not out of the box (it tries to use "vv1.20.0" : One v added by this PR + one v from the git tag.

Saruspete:
Also, [rpm packaging guide](https://rpm-packaging-guide.github.io/) does not advice for this kind of discrepancy between url / name / version: 

```
Source0:        https://example.com/%{name}/release/%{name}-%{version}.tar.gz
```

Ferroin:
OK, so we need to update our scripts to not strip the `v` and then revert this. I'll open an issue for this shortly.

Ferroin:
And FWIW, stuff like this is part of why I _hate_ RPM. DEB package builds don't mess with this crap, they just assume you're running from the local source directory and sidestep all of this.

prologic:
@Saruspete I hope the PR that was just merged fixes your scripts? Please let us know ASAP if it does not!

Saruspete:
Hello,

Confirmed, the official build script in contrib works again, and so do my scripts.
Tested with following sequence: (note the `tag -a` for git describe to pick this annotated tag)
```
git clone https://github.com/netdata/netdata.git netdata.testspec
cd netdata.testspec
git tag -a "v1.20.1" -m test
./contrib/rhel/build-netdata-rpm.sh
```

Many thanks for the fast reactivity (again :) ) !

prologic:
Awesome *phew* :)

Issue #7656, opened 2020-01-02
bug template suggestions


ilyam8:
i see our bug template

```
<!---
When creating a bug report please:
- Verify first that your issue is not already reported on GitHub
- Test if the latest release and master branch are affected too.
- Provide a clear and concise description of what the bug is in "Bug report 
  summary" section.
- Try to provide as much information about your environment (OS distribution,
  running in container, etc.) as possible to allow us reproduce this bug faster.
- Write which component is affected. We group our components the same way our
  code is structured so basically:
  component name = dir in top level directory of repository
- Describe how you found this bug and how we can reproduce it. Preferable with
  a minimal test-case scenario. You can paste gist.github.com links for larger
  files
- Provide a clear and concise description of what you expected to happen.
-->

##### Bug report summary

##### OS / Environment (output of `uname -a; command -v sw_vers && sw_vers || grep -Hv "^#" /etc/*release`)

##### Netdata version (ouput of `netdata -V`)

##### Component Name

##### Steps To Reproduce

##### Expected behavior
```

I assume most of the people dont read text between `<!---  -->` (me included 😅 	):
 - a lot of text in one place
 - it looks like a recommendation
 - not clear which bullet applies for which section

I think we need to add relevant description to every `####` section, like

```
##### section name
<!-- description-->

##### section name
<!-- description-->

...
```

_Originally posted by @ilyam8 in https://github.com/netdata/netdata/pull/7550#issuecomment-570153740_

thiagoftsm:
A good suggestion, in your vision should we also add or remove sections that we have? I confess I am happy with the current sections.

ilyam8:
I dont want to remove/change sections, my point is that it is better to have

```
##### section name
<!-- section description-->

##### section name
<!-- section description-->
```

then

```
<!-- global description-->

##### section name

##### section name
```

thiagoftsm:
I understood your suggestion, I only added comments about change the `sections` for we change everything, case it was necessary, with one unique PR.

PR #7805, opened 2020-01-20
bug_report improvements

ilyam8:
##### Summary

Fixes: #7656

##### Component Name

`.github/ISSUE_TEMPLATE`

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,7 +4,7 @@ about: Create a bug report to help us improve
 labels: bug, needs triage
 ---
 
-<!---
+<!--
 When creating a bug report please:
 - Verify first that your issue is not already reported on GitHub.
 - Test if the latest release and master branch are affected too.
@@ -13,7 +13,7 @@ When creating a bug report please:
 ##### Bug report summary
 <!-- Provide a clear and concise description of what the bug. -->
 
-##### OS / Environment ()
+##### OS / Environment
 <!--
 Provide as much information about your environment (OS distribution, running in container, etc.)
 as possible to allow us reproduce this bug faster.
@@ -22,7 +22,7 @@ To get this information execute:
 - uname -a; grep -Hv "^#" /etc/*release  # linux/bsd
 - uname -a; sw_vers                      # macOS
 
-Place config in the code section.  
+Place output in the code section.  
  -->
 ```
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -21,7 +21,7 @@ as possible to allow us reproduce this bug faster.
 To get this information execute:
 - uname -a; grep -Hv "^#" /etc/*release  # linux
 - uname -a; uname -K                     # bsd
-- sw_vers                                # macOS
+- uname -a; sw_vers                      # macOS
 
 Place output in the code section.  
  -->
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,19 +11,19 @@ When creating a bug report please:
 -->
 
 ##### Bug report summary
-<!-- Provide a clear and concise description of what the bug. -->
+<!-- Provide a clear and concise description of the bug you're experiencing. -->
 
 ##### OS / Environment
 <!--
-Provide as much information about your environment (OS distribution, running in container, etc.)
+Provide as much information about your environment (which operating system and distribution you're using, if Netdata is running in a container, etc.)
 as possible to allow us reproduce this bug faster.
 
-To get this information execute:
-- uname -a; grep -Hv "^#" /etc/*release  # linux
-- uname -a; uname -K                     # bsd
+To get this information, execute the following commands based on your operating system:
+- uname -a; grep -Hv "^#" /etc/*release  # Linux
+- uname -a; uname -K                     # BSD
 - uname -a; sw_vers                      # macOS
 
-Place output in the code section.  
+Place the output from the command in the code section below.  
  -->
 ```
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,31 +4,52 @@ about: Create a bug report to help us improve
...
+##### OS / Environment
+<!--
+Provide as much information about your environment (OS distribution, running in container, etc.)
+as possible to allow us reproduce this bug faster.
 
-##### Netdata version (ouput of `netdata -V`)
+To get this information execute:
+- uname -a; grep -Hv "^#" /etc/*release  # linux/bsd
+- uname -a; sw_vers                      # macOS
+
```
thiagoftsm: The previous 3 lines should be on `OS /Environment`

thiagoftsm: Sorry, I only saw now that you are removing Netdata version.

ilyam8: i havent remove anything, i moved stuff to appropriate section.

> The previous 3 lines should be on OS /Environment

Those limes under `OS /Environment` section.

ilyam8:
@thiagoftsm 
https://raw.githubusercontent.com/ilyam8/netdata/bug_template_improvements/.github/ISSUE_TEMPLATE/bug_report.md

ilyam8:
btw

> uname -a; grep -Hv "^#" /etc/*release  # linux/bsd

is not correct, it recall it doesnt work on (at least all) bsd systems, should be improved.

amoss:
> btw
> 
> > uname -a; grep -Hv "^#" /etc/*release  # linux/bsd
> 
> is not correct, i recall it doesnt work on (at least all) bsd systems, should be improved.

This has just been updated in `daemon/system-info.sh`, there is an example in there on how to find this for FreeBSD.

ilyam8:
https://github.com/netdata/netdata/blob/49f6950820f083956ddc2c9327eeae4e9ef0d5ed/daemon/system-info.sh#L95-L101

This?

so just `uname -a` ?

amoss:
Adding the `uname -K` gives the explicit kernel version, but it is the other part that does not work on all BSDs isn't it? There is no `/etc/release` on FreeBSD for example.

What about something like:
```
sw_vers 2>/dev/null || grep -Hv "^#" /etc/*release || (uname -a; uname -K)
```

ilyam8:
> sw_vers 2>/dev/null || grep -Hv "^#" /etc/*release 2>/dev/nul || (uname -a; uname -K)


i dont see reason why we need to have one line (which is big already and will grow in the future 100%) if we can have 3:

- linux
- bsd
- macos

It is even easier for me - clear what is what.

joelhans:
I'll try to review this soon. Does the SRE team know about this yet? I feel like one of them said something at the meetup about optimizing this very template. Maybe drop in their channel to see if anyone wants to participate.

Ferroin:
I second @prologic's assessment, but will hold off officially reviewing as this should also be approved by @joelhans.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,31 +4,53 @@ about: Create a bug report to help us improve
...
+Provide as much information about your environment (OS distribution, running in container, etc.)
+as possible to allow us reproduce this bug faster.
 
-##### Netdata version (ouput of `netdata -V`)
+To get this information execute:
+- uname -a; grep -Hv "^#" /etc/*release  # linux
+- uname -a; uname -K                     # bsd
+- uname -a; sw_vers                      # macOS
+
+Place output in the code section.
```
joelhans:
```suggestion
Place the output from the command in the code section below.
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,31 +4,53 @@ about: Create a bug report to help us improve
...
+ -->
+```
+
+```
+
+##### Netdata version
+<!--
+Provide output of netdata -V.
+ 
+If netdata is running execute: $(ps aux | grep -E -o "[a-zA-Z/]+netdata ") -V
```
joelhans:
```suggestion
If Netdata is running, execute: $(ps aux | grep -E -o "[a-zA-Z/]+netdata ") -V
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,31 +4,53 @@ about: Create a bug report to help us improve
...
+ -->
+ 
 
 ##### Component Name
+<!--
+Write which component is affected. We group our components the same way our code is structured so basically: 
+component name = dir in top level directory of repository.
+-->
```
joelhans: Let's try something like this:

Let us know which component is affected by the bug. Our code is structured according to its component, so the component name is the same as the top level directory of the repository. For example, a bug in the dashboard would be under the web component.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,31 +4,53 @@ about: Create a bug report to help us improve
...
 
 ##### Steps To Reproduce
+<!--
+Describe how you found this bug and how we can reproduce it. Preferable with a minimal test-case scenario.
```
joelhans:
```suggestion
Describe how you found this bug and how we can reproduce it, preferably with a minimal test-case scenario.
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,31 +4,53 @@ about: Create a bug report to help us improve
...
+You can paste gist.github.com links for larger files.
```
joelhans:
```suggestion
If you'd like to attach larger files, use gist.github.com and paste in links.
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,31 +4,53 @@ about: Create a bug report to help us improve
...
+##### OS / Environment
+<!--
+Provide as much information about your environment (OS distribution, running in container, etc.)
+as possible to allow us reproduce this bug faster.
 
-##### Netdata version (ouput of `netdata -V`)
+To get this information execute:
+- uname -a; grep -Hv "^#" /etc/*release  # linux
+- uname -a; uname -K                     # bsd
+- uname -a; sw_vers                      # macOS
```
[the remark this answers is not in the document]
Ferroin: I _think_ that should work on all modern BSD variants, but we technically only support FreeBSD right now.

PR #7809, opened 2020-01-21
add possibility to change badge text font color

### LINKED ISSUE #7791: badge font color
Hi 
How to change badge font color?

underhood:
##### Summary
Fixes #7791 - request of user to be able to specify color of the text in badge

__Additionally__
Fixes problem where the custom color given to `label_color` was ignored using both `#RRGGBB` and `RRGGBB` syntax (in contrast with specification in documentation which allows this).
Unifies all `/color/` values with same syntax. 2 options:
- one of predefined colors
- custom color specified as hex in `RGB` or `RRGGBB` format *(without preceeding `#` char)*

##### Component Name
agent `api/v1/badge.svg` and related documetation

##### Additional Information
color can be specified with HEX value as in HTML without `#` character in the beggining. Either in `RGB` or `RRGGBB` format

### DIFF
--- a/web/api/badges/README.md
+++ b/web/api/badges/README.md
@@ -192,6 +192,14 @@ These are options dedicated to badges:
 
     The supported operators are `<`, `>`, `<=`, `>=`, `=` (or `:`) and `!=` (or `<>`).
 
+-   `text_color_lbl=RGB` or `text_color_lbl=RRGGBB`
+
+    Specifies hexadecimal color *(like in HTML but without `#` prepended)* for the font of left/label side of the badge. If not given or given with empty value default color will be used.
+
+-   `text_color_val=RGB` or `text_color_val=RRGGBB`
+
+    Specifies hexadecimal color *(like in HTML but without `#` prepended)* for the font of right/value side of the badge. If not given or given with empty value default color will be used.
+
 -   `precision=NUMBER`
 
     The number of decimal digits of the value. By default Netdata will add:
--- a/web/api/badges/README.md
+++ b/web/api/badges/README.md
@@ -180,7 +180,21 @@ These are options dedicated to badges:
 
 -   `label_color=COLOR`
 
-    The color of the label (the left part). You can use any HTML color, include `#NNN` and `#NNNNNN`. The following colors are defined in Netdata (and you can use them by name): `green`, `brightgreen`, `yellow`, `yellowgreen`, `orange`, `red`, `blue`, `grey`, `gray`, `lightgrey`, `lightgray`. These are taken from <https://github.com/badges/shields> so they are compatible with standard badges.
+    The color of the label (the left part). You can use any HTML color in `RGB` or `RRGGBB` hex notation (without `#` character in beginning). Additionally you can use one of the following predefined colors (and you can use them by their name):
+    
+    - `green`
+    - `brightgreen`
+    - `yellow`
+    - `yellowgreen`
+    - `orange`
+    - `red`
+    - `blue`
+    - `grey`
+    - `gray`
+    - `lightgrey`
+    - `lightgray`
+    
+    These are taken from <https://github.com/badges/shields> so they are compatible with standard badges.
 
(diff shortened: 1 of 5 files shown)


thiagoftsm:
Hi @underhood ,

I think your PR is almost there, I tested now it with different combinations of RGB and everything worked as expected, and when I gave the color name as argument it gave me an error, because it was expecting only RGB values. This is 100% correct according the code and the documentation that you are bringing, but there is a problem in my point of view, the background variables `label_color` and `value_color` accepts RGB values and texts, no less important when we give a strange value for these variables, they use their default value. Please, could you transport this behavior for the new variables you brought? 

Thank you!

underhood:
Hi, agreed that we should make it same for all \color\ parameters. However during my testing  the parameter `label_color` doesn't behave as described by documentation says when using custom colors. Additionally sometimes it produces black label when value is wrong instead of falling back to default color.
So i think there is now 2 points:
1. unifing the thing - agreed
2. fixing also older parameters like `label_color` to behave as specification says

thiagoftsm:
>     1. new parameters in this PR are working as documentented

I verified this in my tests, this is really perfect! :)

>     2. older parameter `label_color` (in master) doesn't behave as per documentation. To be specific:
>        
>        * `http://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label_color=ffffff` gives me black label
>        * `http://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label_color=#ffffff` falls back to gray label

For this case, my suggestion is we create another PR that will fix the documentation and we follow this pattern to bring the new feature.

>     3. unifying all the `~color~` parameters to behave same - which is your point and I agree

Nice, I will review again when they are behaving equally.

Thanks!

underhood:
now in all parameters specifing color we can use same syntax interchangeably. `predefined_color_name` or `RGB` or `RRGGBB`. When wrong value given we should revert to default color instead of returning error.

`http://127.0.0.1:19999/api/v1/badge.svg?chart=system.cpu&label_color=orange&value_color=grey:null|green%3C9|a00%3C10&text_color_lbl=0f0&text_color_val=orange`

stelfrag:
I think it would be helpful to to see a small table that lists the keys that can be used to alter the colors for the label (background and text) and similarly for the value 

Something like

 *area* | Background | Text
------- | --------| ----
Label  | label_color | text_color_lbl
Value  | value_color | text_color_val

underhood:
thanks, I will wait on @joelhans  opinion regarding @stelfrag suggestion for documentation.

stelfrag:
I wanted to quick check the color change so I added floating number (see **9.8** below) which doesn't work because probably you are parsing it as integer

It can be fixed or documented 

```
https://127.0.0.1:19999//api/v1/badge.svg?chart=system.cpu&before=-100&after=100&value_color=red:null|green%3C=9.8|yellow%3E5&text_color_val=green
```

underhood:
Well the logic behind parsing the conditions in `value_color` is not touched by this PR. Therefore it works same as before. I can add to documentation that it currently accepts only integer. But if we want to change it to accept floats as well it should be done in separate PR so this one doesn't turn into complete rework of badges.

stelfrag:
I agree. Note in docs just to avoid any questions and revisit in the future

underhood:
ok I added this (pre-existing) detail into docu

underhood:
> I think it would be helpful to to see a small table that lists the keys that can be used to alter the colors for the label (background and text) and similarly for the value
> 
> Something like
> _area_ 	Background 	Text
> Label 	label_color 	text_color_lbl
> Value 	value_color 	text_color_val
> 
> @joelhans what do you think ?

@stelfrag I added the table into the docu as per your suggestion

underhood:
CI reboot

underhood:
@joelhans if you have 5mins can you take a look on docu in this PR?

joelhans:
@underhood On it. I have a recommendation for formatting but I need a few minutes to make sure I get the syntax right before I send it over.

underhood:
thank you @joelhans, i taken over your improvements to documentation :)

Issue #7798, opened 2020-01-20
Use system service manager to try and shut down Netdata during installs, updates, and uninstalls.


mfundul:
Confirmed, that when I manually run `netdata-updater.sh`, netdata is stopped with 6/ABRT.

I assume this is due to the following code triggering the signal which is treated as failure by systemd:

https://github.com/netdata/netdata/blob/c7e13846a87054d1e6e412d8edbc25b9dcbc684f/packaging/installer/functions.sh#L494-L506

Would it be possible to maybe check if the system is managed with systemd, and if so shut down the service with systemd?

Or just in some way that is not treated as failure by systemd?

_Originally posted by @johnramsden in https://github.com/netdata/netdata/issues/6999#issuecomment-576144967_

Ferroin:
There are two issues here:

1. We should be trying to use whatever service management mechanism we detected during install to stop the agent before we try anything else to stop it.
2. _Something_ is triggering a SIGABRT here, which should not at all be the case since we never seem to be doing anything that explicitly triggers it. I suspect an assertion is failing somewhere in one of our dependencies and calling `abort()` explicitly. This is why systemd sees this is a failure state (because an abort _is_ an abnormal termination (see [man 3 abort](http://man7.org/linux/man-pages/man3/abort.3.html)).

Both of these need fixed, but they really are orthogonal to each other. I can take care of the first (shouldn't take me too long, but won't be until tomorrow since it's a holiday today in the US), but I don't have the expertise to track down what's causing the second.

mfundul:
@johnramsden since this is reproducible for you can you give us the text output of running this script? There may be some very useful print in there that will help solve this.

PR #7814, opened 2020-01-22
Attempt to use system service manager to shut down Netdata.

Ferroin:
##### Summary

This updates the installer and uninstaller to attempt to use the system service management interface to shut down netdata when running as root.

This will allow us to properly shut down Netdata from the perspective of the system service manager, thus avoiding cases where it gets restarted immediately.

If this step succeeds, we wait 20 seconds before checking for a successful shutdown. If the shutdown was unsuccessful or this step failed, we continue with the same logic we already had, first attempting to use `netdatacli` and then falling back to direct usage of signals for shutdown.

##### Component Name

area/packaging

##### Additional Information

Minimally tested to verify that it doesn't break the build and that the correct code gets run on systemd, OpenRC, and classic init.d based systems. I do not have a system immediately available to test FreeBSD handling on, and cannot test macOS handling.

Fixes: #7798

### DIFF
--- a/packaging/installer/functions.sh
+++ b/packaging/installer/functions.sh
@@ -467,7 +467,22 @@ netdata_pids() {
 }
 
 stop_all_netdata() {
-    local p
+    local p uname
+
+    if [ "${UID}" -eq 0 ] ; then
+        uname="$(uname 2>/dev/null)"
+        if issystemd; then
+            systemctl stop netdata
+        elif [ "${uname}" = "Darwin" ]; then
+            launchctl stop netdata
+        elif [ "${uname}" = "FreeBSD" ]; then
+            # This may return failure if Netdata is not enabled.
+            /etc/rc.d/netdata stop || true
+        else
+            service netdata stop
+        fi
+        sleep 20
+    fi
 
     if [ -n "$(netdata_pids)" ] || [ -n "$(builtin type -P netdatacli)" ]; then
         netdatacli shutdown-agent
--- a/packaging/installer/functions.sh
+++ b/packaging/installer/functions.sh
@@ -144,7 +144,7 @@ run_failed() {
 }
 
 ESCAPED_PRINT_METHOD=
-if prnitf "%q " test >/dev/null 2>&1; then
+if printf "%q " test >/dev/null 2>&1; then
 	ESCAPED_PRINT_METHOD="printfq"
 fi
 escaped_print() {
--- a/packaging/installer/functions.sh
+++ b/packaging/installer/functions.sh
@@ -521,19 +521,19 @@ stop_all_netdata() {
         # Any of these may fail, but we need to not bail if they do.
         if issystemd; then
             if systemctl stop netdata ; then
-                sleep 20
+                sleep 5
             fi
         elif [ "${uname}" = "Darwin" ]; then
             if launchctl stop netdata ; then
-                sleep 20
+                sleep 5
             fi
         elif [ "${uname}" = "FreeBSD" ]; then
             if /etc/rc.d/netdata stop ; then
-                sleep 20
+                sleep 5
             fi
         else
             if service netdata stop ; then
-                sleep 20
+                sleep 5
             fi
         fi
     fi
--- a/packaging/installer/functions.sh
+++ b/packaging/installer/functions.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 # SPDX-License-Identifier: GPL-3.0-or-later
-# shellcheck disable=SC1091,SC1117,SC2002,SC2004,SC2034,SC2046,SC2059,SC2086,SC2129,SC2148,SC2154,SC2155,SC2162,SC2166,SC2181,SC2193
 
 # make sure we have a UID
 [ -z "${UID}" ] && UID="$(id -u)"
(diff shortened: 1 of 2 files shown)


Ferroin:
I've now tested on FreeBSD as well, and it looks like things are working correctly there too.

packaging/installer/functions.sh
```diff
@@ -10,30 +9,15 @@
 
 setup_terminal() {
     TPUT_RESET=""
-    TPUT_BLACK=""
```
knatsakis: I think these definitions should be kept for completeness, even if unused and shellcheck complains about them.

knatsakis: I think these definitions should be kept for completeness, even if unused and shellcheck complains about them.

packaging/installer/functions.sh
```diff
@@ -42,30 +26,15 @@ setup_terminal() {
         if [ $(($(tput colors 2>/dev/null))) -ge 8 ]; then
             # Enable colors
             TPUT_RESET="$(tput sgr 0)"
-            TPUT_BLACK="$(tput setaf 0)"
```
knatsakis: The same for these definitions as well. I think we should keep them for completeness

packaging/installer/functions.sh
```diff
@@ -161,21 +131,22 @@ find_processors() {
...
 
 run_failed() {
-    printf >&2 "${TPUT_BGRED}${TPUT_WHITE}${TPUT_BOLD} FAILED ${TPUT_RESET} ${*} \n\n"
+    printf >&2 "%s FAILED %s %s \n\n" "${TPUT_BGRED}${TPUT_WHITE}${TPUT_BOLD}" "${TPUT_RESET}" "${*}"
 }
 
 ESCAPED_PRINT_METHOD=
-printf "%q " test >/dev/null 2>&1
-[ $? -eq 0 ] && ESCAPED_PRINT_METHOD="printfq"
+if prnitf "%q " test >/dev/null 2>&1; then
```
knatsakis: Typo?

```suggestion
if printf "%q " test >/dev/null 2>&1; then
```

packaging/installer/netdata-uninstaller.sh
```diff
@@ -97,16 +119,17 @@ setup_terminal() {
...
 
 run_failed() {
-	printf >&2 "${TPUT_BGRED}${TPUT_WHITE}${TPUT_BOLD} FAILED ${TPUT_RESET} ${*} \n\n"
+	printf >&2 "%s FAILED %s %s \n\n" "${TPUT_BGRED}${TPUT_WHITE}${TPUT_BOLD}" "${TPUT_RESET}" "${*}"
 }
 
 ESCAPED_PRINT_METHOD=
-printf "%q " test >/dev/null 2>&1
-[ $? -eq 0 ] && ESCAPED_PRINT_METHOD="printfq"
+if prnitf "%q " test >/dev/null 2>&1; then
```
knatsakis: Typo?

```suggestion
if printf "%q " test >/dev/null 2>&1; then
```

prologic:
> I do not have a system immediately available to test FreeBSD handling on, and cannot test macOS handling.

I can easily test on FreeBSD here for you.

But do we even have `netdata.plist` service here for macOS? I haven't check... macOS uses [LaunchD](https://en.wikipedia.org/wiki/Launchd) which [Homebrew](https://brew.sh) has a [wrapper](https://github.com/Homebrew/homebrew-services) for (_which we should be baking into our Formulae if we don't already have one or do_)...

packaging/installer/functions.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 # SPDX-License-Identifier: GPL-3.0-or-later
-# shellcheck disable=SC1091,SC1117,SC2002,SC2004,SC2034,SC2046,SC2059,SC2086,SC2129,SC2148,SC2154,SC2155,SC2162,SC2166,SC2181,SC2193
```
prologic: Thanks! 🤣Let's actively remove these from user-facing shell scripts as we do other things. 👍

packaging/installer/functions.sh
prologic: Yeah I kind of agree.

packaging/installer/netdata-uninstaller.sh
```diff
@@ -65,6 +64,29 @@ if [[ $EUID -ne 0 ]]; then
 	fi
 fi
 
+# -----------------------------------------------------------------------------
+# portable service command
+
+service_cmd="$(command -v service 2>/dev/null)"
+rcservice_cmd="$(command -v rc-service 2>/dev/null)"
+systemctl_cmd="$(command -v systemctl 2>/dev/null)"
+service() {
```
prologic:
```suggestion
portable_service() {
```

mfundul:
What is the reasoning behind waiting for 20 seconds after a successful shutdown?

In the cases of systemd and netdatacli the shutdown command is practically synchronous and the agent should have shutdown by the time those commands have returned.

In the case of sending a SIGNAL, since this is asynchronous, you have to wait arbitrarily long because you don't know how much time it's going to take to flush the database.

Ferroin:
> What is the reasoning behind waiting for 20 seconds after a successful shutdown?
> 
> In the cases of systemd and netdatacli the shutdown command is practically synchronous and the agent should have shutdown by the time those commands have returned.
> 
> In the case of sending a SIGNAL, since this is asynchronous, you have to wait arbitrarily long because you don't know how much time it's going to take to flush the database.

Based on personal experience, no, `systemctl stop foo` is not reliably synchronous.  The other shutdown methods in the first pass aren't reliably synchrnous either for that matter, though we can probably cut back on the delay to maybe 5 seconds.

packaging/installer/netdata-uninstaller.sh
Ferroin: This was copied directly from `functions.sh`. I intentionally did not change the name to make it trivial to check that the code is the same.

prologic:
> Based on personal experience, no, `systemctl stop foo` is not reliably synchronous. The other shutdown methods in the first pass aren't reliably synchrnous either for that matter, though we can probably cut back on the delay to maybe 5 seconds.

It _would_ be synchronous if SystemD called [`waitpid`](https://linux.die.net/man/2/waitpid) -- But _maybe_ SystemD doesn't do this correctly or at all? (_I dunno I haven't checked the source code..._)

Ferroin:
> > Based on personal experience, no, `systemctl stop foo` is not reliably synchronous. The other shutdown methods in the first pass aren't reliably synchrnous either for that matter, though we can probably cut back on the delay to maybe 5 seconds.
> 
> It _would_ be synchronous if SystemD called [`waitpid`](https://linux.die.net/man/2/waitpid) -- But _maybe_ SystemD doesn't do this correctly or at all? (_I dunno I haven't checked the source code..._)

It may do so in the current versions and be properly synchronous, but I've had issues in the past with it not being synchronous.

Issue #7731, opened 2020-01-14
Improve the system-info meta-data output.


amoss:
##### Feature idea summary
Update the output of the `system-info.sh` script so that it includes all of the following information. 

##### Expected behavior
Each of the following data is visible in the info response and message.

Already existing, but being changed in #7488.
* Operating system
* OS version, architecture (x86, etc.)
* Kernel name, version
* Container and Virtualization technology

New, to be added as described in the discussion below:
* Number of cores
* Nominal frequency (not the current frequency)
* Amount of RAM
* Sum of available capacity on disks

The main requirement on the new information is that we use a staggered approach to allow graceful degradation of the information returned rather than failure. This should be robust across the range of architectures and systems (bare-metal, VMs and containers) discussed below. An exact definition of the variables exported by the script will appear when #7488 is complete.

Ferroin:
We've got OS, arch, OS version, kernel name, and kernel version detection about as good as we can make it in the existing `system-info.sh` script.

For the others:

### Number of CPU cores

Trivial solution is to use `nproc --all`, provided in GNU coreutils. Assuming that fails but `/sys` exists and we can list directories in it (the normal case unless some LSM is in use), the following will do the same:

`find /sys/devices/system/cpu -mindepth 1 -maxdepth 1 -type d -name 'cpu*' | grep -Ev 'idle|freq' | wc -l`

Otherwise, we have to count entries in `/proc/cpuinfo` like so:

`grep -c ^processor /proc/cpuinfo`

### 'Nominal' Frequency

Assuming this is the 'design' frequency of the CPU, we actually need `/sys` to obtain this info, as well as cpufreq support in the kernel. The 'design' frequency for a given logical CPU core can be found in `/sys/devices/system/cpu/cpu<N>/cpufreq/base_frequency`, where `<N>` is the core number. The max frequency (which may be higher than the design frequency) can be found in `/sys/devices/system/cpu/cpu<N>/cpufreq/scaling_max_frequency`. Both report values in hertz without any scaling.

Note that we cannot rely on DMI info for this for a few specific reasons:

* `dmidecode` will actually crash or hang some virtual machines. It needs to be specially handled by the VM's firmware itself for it to work.
* DMI data is only available reliably on x86 systems.
* The data exposed by DMI might not be accurate. For example, on my laptop it reports 8.3 GHz as the max CPU frequency, which is blatantly wrong (2.8GHz design, 3.8GHz boost).

[...]

PR #7815, opened 2020-01-22
Improve the system-info.sh script to report CPU and RAM meta-data.

Ferroin:
##### Summary

This adds extra info to the output of the `system-info.sh` script about the CPU and RAM.

It adds the following keys:

* `NETDATA_LOGICAL_CPU_COUNT`: This reports a count of the number of logical CPU cores in the system as an integer. This may be inaccurate due to kernel command-line options, or hardware peculiarities. In particular, if we end up using `dmidecode` we get the total physical number of logical cores instead of what the kernel may be using (which might be less due to either some CPU's being hidden or booting without SMP support).
* `NETDATA_CPU_MODEL`: This reports the CPU model as a string. This usually includes the product family, and may additionally include any combination of an exact model name and number, the vendor name, an exact core count, and the intended operating frequency.
* `NETDATA_CPU_VENDOR`: This reports the CPU vendor as a string. This is required because the model information may not include the vendor, and thus may be overly vague.
* `NETDATA_CPU_FREQ`: This reports a best estimate of the design frequency of the CPU (that is, the frequency it is intended to run at). This may be significantly higher or lower than the actual value due to any number of factors (including, but not limited to, overclocking, underclocking, lack of actual information, and inaccurate information being reported on some CPUs). The value will include a unit abbreviation, with the normal cases being either hertz, or megahertz.
* `NETDATA_CPU_DETECTION`: Indicates how we gathered the CPU data, either 'none' if we were unable to gather any, or a space separated list of methods (see below for the possible values).
* `NETDATA_TOTAL_RAM`: This reports the total usable RAM as seen by the OS as a number with an associated unit (usually bytes or kilobytes). This will usually be a bit less than the 'advertised' RAM total for the system, typically by less than 1%.
* `NETDATA_RAM_DETECTION`: Indicates how we gathered the CPU data, either 'none' if we were unable to gather any. See below for possible values

Data is collected using some combination of the following detection methods, in preferred order of usage:

* `lscpu`: Used for all the CPU information. This leverages the `lscpu` command to collect data from multiple sources. Depends on the `lscpu` command being installed.
* `dmidecode`: Used for all the CPU information. This leverages the DMI information found on many commodity systems. Depends on hardware/firmware support as well as having the `dmidecode` command installed.
* `nproc`: Used to get a count of logical CPU cores. This uses the `nproc` command from GNU coreutils. Requires either GNU coreutils, or a compatible `nproc` command.
* `sysctl` Used to get CPU core count and model name as well as the total RAM. Uses the BSD `sysctl` command to fetch data directly from the kernel. Only used on BSD.
* `uname`: Used to fetch CPU model and vendor. Uses switches specific to the GNU version of `uname`.
* `sysfs`: Used to fetch CPU core counts and frequency info. Queries `/sys` directly on Linux to retrieve the information. Frequency information from `/sys` may override any obtained from` lscpu` or `dmidecode` on some systems.
* `procfs`: Used to fetch all CPU and RAM information. Queries `/proc/cpuinfo` and `/proc/meminfo` directly to retrieve the information.

##### Component Name

area/

##### Additional Information

This is a partial implementation of the items requested in #7731, covering all the CPU and RAM info on Linux and FreeBSD. It does not currently handle macOS since I cannot find conclusive documentation on how to get this info from a shell script there, and does not include the disk information as that did not appear to be conclusively decided upon in the issue.

Tested minimally on a selection of VM's and physical hardware.


thiagoftsm:
Hi @Ferroin ,

I tested now on my laptop and I saw that probably it is necessary to adjust the parser, please, take a look in the script output:

```
NETDATA_CONTAINER_OS_NAME=none
NETDATA_CONTAINER_OS_ID=none
NETDATA_CONTAINER_OS_ID_LIKE=none
NETDATA_CONTAINER_OS_VERSION=none
NETDATA_CONTAINER_OS_VERSION_ID=none
NETDATA_CONTAINER_OS_DETECTION=none
NETDATA_HOST_OS_NAME=Slackware
NETDATA_HOST_OS_ID=slackware
NETDATA_HOST_OS_ID_LIKE=unknown
NETDATA_HOST_OS_VERSION=14.2
NETDATA_HOST_OS_VERSION_ID=14.2
NETDATA_HOST_OS_DETECTION=/etc/os-release
NETDATA_SYSTEM_KERNEL_VERSION=5.4.13
NETDATA_SYSTEM_VIRTUALIZATION=unknown
NETDATA_SYSTEM_VIRT_DETECTION=none
NETDATA_SYSTEM_CONTAINER=unknown
NETDATA_SYSTEM_CONTAINER_DETECTION=none
NETDATA_CPU_LOGICAL_CPU_COUNT=4
0-3
NETDATA_CPU_VENDOR="GenuineIntel"
NETDATA_CPU_MODEL="Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz"
NETDATA_CPU_FREQ="2700000 Hz"
NETDATA_CPU_DETECTION="lscpu sysfs"
NETDATA_TOTAL_RAM="8064228 kB"
NETDATA_RAM_DETECTION=procfs
``` 
and now part of my `lscpu` output:

```
Address sizes:                   39 bits physical, 48 bits virtual
CPU(s):                          4
On-line CPU(s) list:             0-3
Thread(s) per core:              2
Core(s) per socket:              2
Socket(s):                       1
NUMA node(s):                    1
```

as you can see the script is also using the value from `On-line CPU(s) list` as result, but I think that we need only the total for the variable `NETDATA_CPU_LOGICAL_CPU_COUNT`.

Of course I can be wrong, please, can you check this on your side?

Ferroin:
@thiagoftsm  No, you're right, I somehow missed that during testing. I'll push a fix shortly.

thiagoftsm:
> @thiagoftsm No, you're right, I somehow missed that during testing. I'll push a fix shortly.

Thanks!

Ferroin:
> > @thiagoftsm No, you're right, I somehow missed that during testing. I'll push a fix shortly.
> 
> Thanks!

Correction, it isn't behaving like that on any of my test systems or my laptop.

Can you post the output of `lscpu | grep -F "CPU(s):"` on your system?

thiagoftsm:
> lscpu | grep -F "CPU(s):"

Yes, this is the output:

```
bash-5.0# lscpu | grep -F "CPU(s):"
CPU(s):                          4
NUMA node0 CPU(s):               0-3
bash-5.0# lscpu --version
lscpu from util-linux 2.34
bash-5.0# grep --version
grep (GNU grep) 3.4
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and others; see
<https://git.sv.gnu.org/cgit/grep.git/tree/AUTHORS>.
```

I am also adding the `lscpu` and `grep` versions.

thiagoftsm:
Considering that 100% of my computers are Slackware, I decided to run on a cloud computer to confirm that it is not something specific of the Slackware, this time I ran on a CentOS 7:

```
[fractal@malta2183 ~]$ lscpu | grep -F "CPU(s):"
CPU(s):                4
NUMA node0 CPU(s):     0-3
[fractal@malta2183 ~]$ lscpu --version
lscpu from util-linux 2.23.2
[fractal@malta2183 ~]$ grep --version
grep (GNU grep) 2.20
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and others, see <http://git.sv.gnu.org/cgit/grep.git/tree/AUTHORS>.
[fractal@malta2183 ~]$ uname -a
Linux XYZ 3.10.0-1062.9.1.el7.x86_64 #1 SMP Fri Dec 6 15:49:49 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

And I have the same output format

Ferroin:
OK, this is strange, I'm not seeing behavior like that anywhere on my end. Can you try without the `-F` option to `grep`?

thiagoftsm:
> OK, this is strange, I'm not seeing behavior like that anywhere on my end. Can you try without the `-F` option to `grep`?

Of course, this is the result:

```
bash-5.0# lscpu | grep "CPU(s):"
CPU(s):                          4
NUMA node0 CPU(s):               0-3
```

thiagoftsm:
Please, test on your side the next comand:

```
 lscpu |grep "^CPU(s)" | cut -f 2 -d ':' | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//'
```

to add the "^" was enough in my side.

daemon/system-info.sh
```diff
@@ -189,4 +286,11 @@ echo "NETDATA_SYSTEM_VIRTUALIZATION=${VIRTUALIZATION}"
 echo "NETDATA_SYSTEM_VIRT_DETECTION=${VIRT_DETECTION}"
 echo "NETDATA_SYSTEM_CONTAINER=${CONTAINER}"
 echo "NETDATA_SYSTEM_CONTAINER_DETECTION=${CONT_DETECTION}"
+echo "NETDATA_CPU_LOGICAL_CPU_COUNT=${LCPU_COUNT}"
+echo "NETDATA_CPU_VENDOR=\"${CPU_VENDOR}\""
```
amoss: Please strip the extra internal quotes so that it is consistent with the other cases. When Netdata parses the output of the script it takes everything between the `=` and the newline as a string, we don't break on spaces.

Ferroin: Ah, OK, I thought we were parsing this as shell variables. I'll update that shortly.

prologic:
> Please, test on your side the next comand:
> 
> ```
>  lscpu |grep "^CPU(s)" | cut -f 2 -d ':' | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//'
> ```
> 
> to add the "^" was enough in my side.

This `lscpu` at least doesn't exist on macOS; probably BSD too; haven't checked yet.

amoss:
The virtualization detection in the final case is because there is no access to `systemd-detect-virt` or `dmidecode` on FreeBSD.

The difference in units may cause a problem:
> NETDATA_TOTAL_RAM="32930224 kB"
vs
> NETDATA_TOTAL_RAM="2107158528 B"

@cakrit How is this field being interpreted by the far end?

prologic:
> The difference in units may cause a problem:

Good catch! I didn't notice this! We _should_ probably just report bytes.

Ferroin:
> > Please, test on your side the next comand:
> > ```
> >  lscpu |grep "^CPU(s)" | cut -f 2 -d ':' | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//'
> > ```
> > 
> > 
> > to add the "^" was enough in my side.
> 
> This `lscpu` at least doesn't exist on macOS; probably BSD too; haven't checked yet.

Yes, it's Linux only.

---



> The virtualization detection in the final case is because there is no access to `systemd-detect-virt` or `dmidecode` on FreeBSD.

Some of the CPU related stuff is worse on FreeBSD for the same reason.

---

> > The difference in units may cause a problem:
> 
> Good catch! I didn't notice this! We _should_ probably just report bytes.

I'll add a commit to just drop everything to base units (this affects CPU frequency as well).

Ferroin:
> Tried to run this on debian 10 bare-metal and the following happens:
> 
> > daemon/system-info.sh: 286: daemon/system-info.sh: Illegal number: 32887316 kB

Fixed now, I had forgotten to update how we're extracting the RAM values to remove the units.

thiagoftsm:
@Ferroin ,

I apologize my wrong review, I messed up with branches. Now that I could run the current, I verified that I could not have any RAM information, as you can see in the previous comment.

amoss:
> @Ferroin ,
> 
> I apologize my wrong review, I messed up with branches. Now that I could run the current, I verified that I could not have any RAM information, as you can see in the previous comment.

I'm seeing the same thing, it is also reported as 0 on my machine.

Ferroin:
> > @Ferroin ,
> > I apologize my wrong review, I messed up with branches. Now that I could run the current, I verified that I could not have any RAM information, as you can see in the previous comment.
> 
> I'm seeing the same thing, it is also reported as 0 on my machine.

On it. I completely forgot that `/proc/meminfo` is intended to be human readable, and therefore does some alignment tricks that don't play nice with `cut`.

ilyam8:
> Tested minimally on a selection of VM's and physical hardware.

i wonder how we can test all these 

> grep -F MemTotal /proc/meminfo | cut -f 2 -d ':' | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//' | cut -f 1 -d ' '

If somebody wants to change something how to know he doesnt break some cases.

I have no idea how to test bash functions, but it looks we need it because this script is important.

Ferroin:
> > Tested minimally on a selection of VM's and physical hardware.
> 
> i wonder how we can test all these

We by definition can't test all of them. What I did test here covered all our standard supported distributions in VM's and docker containers, and Gentoo and Debian on a couple of different bare metal systems.

The issues here are all going to be related to cross-arch stuff and VM versus container versus bare metal.  All the stuff we're poking at for this is either rock solid kernel APIs or stuff provided by the hardware itself. so distro doesn't matter much, but exact combination of kernel and hardware/hypervisor/container matters quite a lot, which makes it rather challenging to test in a properly automated manner.
> 
> > grep -F MemTotal /proc/meminfo | cut -f 2 -d ':' | sed -e 's/^[[:space:]]_//' -e 's/[[:space:]]_$//' | cut -f 1 -d ' '
> 
> If somebody wants to change something how to know he doesnt break some cases.
> 
> I have no idea how to test bash functions, but it looks we need it because this script is important.

That's why review is important.

I should probably go through and make that sed command a function though, as it's used a lot and it's non-obvious what it does (it's supposed to strip leading and trailing whitespace, though I'm pretty sure it only works for a single line).

prologic:
> We by definition can't test all of them. What I did test here covered all our standard supported distributions in VM's and docker containers, and Gentoo and Debian on a couple of different bare metal systems.

I _do_ believe there is a way we can test things like this; but at this stage I'd rather hold of on implementing anything just yet until we've had a change to work out the general pattern of things we should be testing for, against what systems, etc. I  have some ideas around this I'd like to do a small spike for to investigate feasibility.

stelfrag:
Raspberry PI 3 -- 

If it matters, note the CPU FREQ comes with a comma. In Andrew's test above it has a dot

```
NETDATA_CONTAINER_OS_NAME=none
NETDATA_CONTAINER_OS_ID=none
NETDATA_CONTAINER_OS_ID_LIKE=none
NETDATA_CONTAINER_OS_VERSION=none
NETDATA_CONTAINER_OS_VERSION_ID=none
NETDATA_CONTAINER_OS_DETECTION=none
NETDATA_HOST_OS_NAME=Raspbian GNU/Linux
NETDATA_HOST_OS_ID=raspbian
NETDATA_HOST_OS_ID_LIKE=debian
NETDATA_HOST_OS_VERSION=9 (stretch)
NETDATA_HOST_OS_VERSION_ID=9
NETDATA_HOST_OS_DETECTION=/etc/os-release
NETDATA_SYSTEM_KERNEL_VERSION=4.19.66-v7+
NETDATA_SYSTEM_ARCHITECTURE=armv7l
NETDATA_SYSTEM_VIRTUALIZATION=none
NETDATA_SYSTEM_VIRT_DETECTION=systemd-detect-virt
NETDATA_SYSTEM_CONTAINER=unknown
NETDATA_SYSTEM_CONTAINER_DETECTION=none
NETDATA_CPU_LOGICAL_CPU_COUNT=4
NETDATA_CPU_VENDOR=
NETDATA_CPU_MODEL=ARMv7 Processor rev 4 (v7l)
NETDATA_CPU_FREQ=1200,0000
NETDATA_CPU_DETECTION=lscpu
NETDATA_TOTAL_RAM=1016795136
NETDATA_RAM_DETECTION=procfs
```

stelfrag:
Ok it's the LOCALE settings 

> Raspberry PI 3 --
> 
> If it matters, note the CPU FREQ comes with a comma. In Andrew's test above it has a dot
> 
> ```
> NETDATA_CONTAINER_OS_NAME=none
> NETDATA_CONTAINER_OS_ID=none
> NETDATA_CONTAINER_OS_ID_LIKE=none
> NETDATA_CONTAINER_OS_VERSION=none
> NETDATA_CONTAINER_OS_VERSION_ID=none
> NETDATA_CONTAINER_OS_DETECTION=none
> NETDATA_HOST_OS_NAME=Raspbian GNU/Linux
> NETDATA_HOST_OS_ID=raspbian
> NETDATA_HOST_OS_ID_LIKE=debian
> NETDATA_HOST_OS_VERSION=9 (stretch)
> NETDATA_HOST_OS_VERSION_ID=9
> NETDATA_HOST_OS_DETECTION=/etc/os-release
> NETDATA_SYSTEM_KERNEL_NAME=Linux
> NETDATA_SYSTEM_KERNEL_VERSION=4.19.66-v7+
> NETDATA_SYSTEM_ARCHITECTURE=armv7l
> NETDATA_SYSTEM_VIRTUALIZATION=none
> NETDATA_SYSTEM_VIRT_DETECTION=systemd-detect-virt
> NETDATA_SYSTEM_CONTAINER=unknown
> NETDATA_SYSTEM_CONTAINER_DETECTION=none
> NETDATA_CPU_LOGICAL_CPU_COUNT=4
> NETDATA_CPU_VENDOR=
> NETDATA_CPU_MODEL=ARMv7 Processor rev 4 (v7l)
> NETDATA_CPU_FREQ=1200,0000
> NETDATA_CPU_DETECTION=lscpu
> NETDATA_TOTAL_RAM=1016795136
> NETDATA_RAM_DETECTION=procfs
> ```

Ferroin:
> Ok it's the LOCALE settings
> 
> > Raspberry PI 3 --
> > If it matters, note the CPU FREQ comes with a comma. In Andrew's test above it has a dot
> > ```
> > NETDATA_CONTAINER_OS_NAME=none
> > NETDATA_CONTAINER_OS_ID=none
> > NETDATA_CONTAINER_OS_ID_LIKE=none
> > NETDATA_CONTAINER_OS_VERSION=none
> > NETDATA_CONTAINER_OS_VERSION_ID=none
> > NETDATA_CONTAINER_OS_DETECTION=none
> > NETDATA_HOST_OS_NAME=Raspbian GNU/Linux
> > NETDATA_HOST_OS_ID=raspbian
> > NETDATA_HOST_OS_ID_LIKE=debian
> > NETDATA_HOST_OS_VERSION=9 (stretch)
> > NETDATA_HOST_OS_VERSION_ID=9
> > NETDATA_HOST_OS_DETECTION=/etc/os-release
> > NETDATA_SYSTEM_KERNEL_NAME=Linux
> > NETDATA_SYSTEM_KERNEL_VERSION=4.19.66-v7+
> > NETDATA_SYSTEM_ARCHITECTURE=armv7l
> > NETDATA_SYSTEM_VIRTUALIZATION=none
> > NETDATA_SYSTEM_VIRT_DETECTION=systemd-detect-virt
> > NETDATA_SYSTEM_CONTAINER=unknown
> > NETDATA_SYSTEM_CONTAINER_DETECTION=none
> > NETDATA_CPU_LOGICAL_CPU_COUNT=4
> > NETDATA_CPU_VENDOR=
> > NETDATA_CPU_MODEL=ARMv7 Processor rev 4 (v7l)
> > NETDATA_CPU_FREQ=1200,0000
> > NETDATA_CPU_DETECTION=lscpu
> > NETDATA_TOTAL_RAM=1016795136
> > NETDATA_RAM_DETECTION=procfs
> > ```

We should still be stripping out any non-digit characters so it can just be parsed as a number without having to worry about locale. I'll push a commit to take care of that later this morning.

ilyam8:
@Ferroin I meant to test parsing part. You have some input and you expect certain output.

Input is either from `/proc/sys/etc` or from command `lscpu/dmidecode/etc`. Dont know if it makes sense.

Ferroin:
> @Ferroin I meant to test parsing part. You have some input and you expect certain output.
> 
> Input is either from `/proc/sys/etc` or from command `lscpu/dmidecode/etc`. Dont know if it makes sense.

Ah, OK, that makes sense then. I'm really not sure what the most sensible way to go about setting that up is though.

amoss:
If we just strip characters then it will create a problem, for example:
```
NETDATA_CPU_FREQ=1200,0000
```
will become the following, which we cannot use:
```
NETDATA_CPU_FREQ=12000000
```
Instead it would make sense to strip the tailing part using something like `grep -o '`[0-9]*'`

Ferroin:
> If we just strip characters then it will create a problem, for example:
> 
> ```
> NETDATA_CPU_FREQ=1200,0000
> ```
> 
> will become the following, which we cannot use:
> 
> ```
> NETDATA_CPU_FREQ=12000000
> ```
> 
> Instead it would make sense to strip the trailing part using something like `grep -o '^[0-9]*'`

It's a Raspberry Pi, so 12000000 is actually the correct value here, not 1200, since the script is supposed to be reporting in hertz, not megahertz.

The math involved here should _never_ be producing a decimal point, so either the math is wrong, or that's a digit grouping mark (at least US standard uses `,` for digit grouping and `.` for decimal point) as I first thought (and whatever locale @stelfrag is using is inconsistent with what I would have expected for a Greek locale).

@stelfrag What exact locale are you using, and do you happen to know what the digit grouping rules for it are?

Ferroin:
Thinking about this further, I think the correct approach (after I double check the math) is to use `LC_NUMERIC=C` for the whole script, so that we have no digit grouping.

Ferroin:
Nope, just looked again, the math is off somehow, because the Pi should be reporting 1200000, we've got an extra zero somehow (it's 1.2GHz, not 12GHz).

@stellfrag Could you post the output of `lscpu` from that Raspberry Pi? I don't have one immediately available to check on.

Ferroin:
No, NVM, it's the `lscpu` parsing that's wrong, I forgot to mark it as MHz and truncate properly there.

stelfrag:
It is MHz

```
pi@homeroot:~ $ lscpu 
Architecture:          armv7l
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    1
Core(s) per socket:    4
Socket(s):             1
Model:                 4
Model name:            ARMv7 Processor rev 4 (v7l)
CPU max MHz:           1200,0000
CPU min MHz:           600,0000
BogoMIPS:              38.40
Flags:                 half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32
```

The LC_NUMERIC=el_GR.UTF-8 but when set to C it produces

```
Architecture:          armv7l
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    1
Core(s) per socket:    4
Socket(s):             1
Model:                 4
Model name:            ARMv7 Processor rev 4 (v7l)
CPU max MHz:           1200.0000
CPU min MHz:           600.0000
BogoMIPS:              38.40
Flags:                 half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32
```

Ferroin:
OK, that also explains why the digit grouping was odd, it wasn't digit grouping at all but a decimal point that didn't get properly handled by the shell.

Ferroin:
OK, just pushed a commit that should fix things on the Pi, as well as any other system we use `lscpu` on.

stelfrag:
> OK, just pushed a commit that should fix things on the Pi, as well as any other system we use `lscpu` on.

Perfect with any locale 
```
: 
NETDATA_CPU_MODEL=ARMv7 Processor rev 4 (v7l)
NETDATA_CPU_FREQ=1200000000
NETDATA_CPU_DETECTION=lscpu
NETDATA_TOTAL_RAM=1016795136
NETDATA_RAM_DETECTION=procfs
```

PR #7817, opened 2020-01-22
Don't remove groups/users in Debian postrm

### LINKED ISSUE #6933: Uninstalling Netdata should not remove `netdata` user and group.
##### Bug report summary
The uninstall process removes the netdata user and group if they exist on the system.

##### OS / Environment
All

Current master branch, behavior has existed for quite some time.

##### Component Name
`kickstart.sh`, Debian packages, possibly other packages (haven't checked any of the RPM's)

##### Steps To Reproduce
Uninstall an existing install of Netdata.

##### Expected behavior
The `netdata` user and group are not removed by the uninstall process.

---

Fedora's packaging guide has some good info on why this is bad practice [here](https://fedoraproject.org/wiki/Packaging:UsersAndGroups#Allocation_Strategies).  Quoting the relevant part:

> **Do not remove users or groups**
We never remove users or groups created by packages. There's no sane way to check if files owned by those users/groups are left behind (and even if there would, what would we do with them?) and leaving those behind with ownerships pointing to now nonexistent users/groups may result in security issues when a semantically unrelated user/group is created later and reuses the UID/GID. Also, in some setups deleting the user/group might not be possible or/nor desirable (eg. when using a shared, remote user/group database). Cleanup of unused users/groups is left to the system administrators to take care of if they so desire.

Adding to this, you have no way to know that the user or group you're removing is the same one which you added, and orphaned files are likely to cause confusion for users in addition to the potential security issues.

We try to deal with the orphaned files issue, but we can't be sure we got everything, so it's still a potential risk, and we quite simply can't deal with any of the other issues properly.

prologic:
Fixes #6933

##### Summary

As per discussion in team meeting. We should not remove users/groups in Debian postrm.

##### Component Name

area/packaging

### DIFF
--- a/contrib/debian/netdata.postrm
+++ b/contrib/debian/netdata.postrm
@@ -1,60 +1,43 @@
-#! /bin/sh
+#!/bin/sh
 
 set -e
 
 dpkg-maintscript-helper dir_to_symlink \
-	/var/lib/netdata/www/.well-known /usr/share/netdata/www/.well-known 1.18.1~ netdata -- "$@"
+  /var/lib/netdata/www/.well-known /usr/share/netdata/www/.well-known 1.18.1~ netdata -- "$@"
 dpkg-maintscript-helper dir_to_symlink \
-	/var/lib/netdata/www/css /usr/share/netdata/www/css 1.18.1~ netdata -- "$@"
+  /var/lib/netdata/www/css /usr/share/netdata/www/css 1.18.1~ netdata -- "$@"
 dpkg-maintscript-helper dir_to_symlink \
-	/var/lib/netdata/www/fonts /usr/share/netdata/www/fonts 1.18.1~ netdata -- "$@"
+  /var/lib/netdata/www/fonts /usr/share/netdata/www/fonts 1.18.1~ netdata -- "$@"
 dpkg-maintscript-helper dir_to_symlink \
-	/var/lib/netdata/www/images /usr/share/netdata/www/images 1.18.1~ netdata -- "$@"
+  /var/lib/netdata/www/images /usr/share/netdata/www/images 1.18.1~ netdata -- "$@"
 dpkg-maintscript-helper dir_to_symlink \
-	/var/lib/netdata/www/lib /usr/share/netdata/www/lib 1.18.1~ netdata -- "$@"
+  /var/lib/netdata/www/lib /usr/share/netdata/www/lib 1.18.1~ netdata -- "$@"
 dpkg-maintscript-helper dir_to_symlink \
-	/var/lib/netdata/www/static /usr/share/netdata/www/static 1.18.1~ netdata -- "$@"
+  /var/lib/netdata/www/static /usr/share/netdata/www/static 1.18.1~ netdata -- "$@"
 
 case "$1" in
-	remove)
-		;;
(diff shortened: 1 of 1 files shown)


contrib/debian/netdata.postrm
```diff
@@ -16,45 +16,31 @@ dpkg-maintscript-helper dir_to_symlink \
 	/var/lib/netdata/www/static /usr/share/netdata/www/static 1.18.1~ netdata -- "$@"
 
 case "$1" in
-	remove)
-		;;
+remove) ;;
 
+\
```
knatsakis: This confuses me. Why is it needed?
```suggestion
```

contrib/debian/netdata.postrm
```diff
@@ -16,45 +16,31 @@ dpkg-maintscript-helper dir_to_symlink \
...
 	purge)
```
knatsakis: Shouldn't the indentation be similar to `remove)` and `*)` ?
```suggestion
purge)
```

contrib/debian/netdata.postrm
prologic: Oh. actually the auto-formatter is doing this shfmt Out-of-my-control unless I go to the effort of tweaking the default settings for shfmt or something. I haven't looked into it tbh.

prologic: Oh. actually the auto-formatter is doing this shfmt Out-of-my-control unless I go to the effort of tweaking the default settings for shfmt or something. I haven't looked into it tbh.

contrib/debian/netdata.postrm
prologic: See above -- Same thing.

prologic:
```sh
$ shfmt --help
usage: shfmt [flags] [path ...]

If no arguments are given, standard input will be used. If a given path
is a directory, it will be recursively searched for shell files - both
by filename extension and by shebang.

  -version  show version and exit

  -l        list files whose formatting differs from shfmt's
  -w        write result to file instead of stdout
  -d        error with a diff when the formatting differs
  -s        simplify the code
  -mn       minify the code to reduce its size (implies -s)

Parser options:

  -ln str   language variant to parse (bash/posix/mksh, default "bash")
  -p        shorthand for -ln=posix

Printer options:

  -i uint   indent: 0 for tabs (default), >0 for number of spaces
  -bn       binary ops like && and | may start a line
  -ci       switch cases will be indented
  -sr       redirect operators will be followed by a space
  -kp       keep column alignment paddings

Utilities:

  -f        recursively find all shell files and print the paths
  -tojson   print syntax tree to stdout as a typed JSON
```

prologic: I _think_ I've fixed this behaviour. For users of Vim and [vim-autoformat](https://github.com/chiel92/vim-autoformat) put this in your `$HOME/.vimrc`:

```
" auto-format
au BufWrite * :Autoformat
let g:autoformat_autoindent = 0
let g:autoformat_retab = 0
let g:autoformat_remove_trailing_spaces = 0

let g:formatdef_custom_shfmt = '"shfmt -i 2 -ci -sr"'
let g:formatters_sh = ['custom_shfmt']
```

PR #7829, opened 2020-01-23
Remove all refernces to .keep files

### LINKED ISSUE #7292: Race condition during installation of .keep files to the same directory
##### Bug report summary
Installlation phase sometimes fails, when trying to install `.keep` files into filesystem with symlinks instead of read directories, when at least two symlink point at the same directory.

##### OS / Environment
Linux, [Buildroot](https://buildroot.org/)
https://patchwork.ozlabs.org/patch/1186612/

v1.18.1

##### Component Name
autotools installation phase

##### Steps To Reproduce
This bug is reproducible only when multiple paths to which `.keep` files are installed resolve to single directory. In case of default [Buildroot](https://buildroot.org/) configuration this is true for `/var/log/netdata` and `/var/cache/netdata` paths. `/var/log` and `/var/cache` are symlinks to `/tmp`. During installation phase two `install` processes are spawned to put `.keep` into `/tmp/netdata`. This is when the race starts. Installation procedure fails when this happens:
1) `install(1)` check whether `/tmp/netdata/.keep` is already there and unlinks it in that case 
2) `install(1)` creates file at `/tmp/netdata/.keep`
3) `install(2)` checks whether `/tmp/netdata/.keep` is there and unlinks it
4) `install(1)` tries to change permission to `/tmp/netdata/.keep`, but it fails because there is no such file
5) `install(2)` creates a `/tmp/netdata/.keep`
6) `install(2)` changes permission to `/tmp/netdata/.keep`

At the end the directory structure is okay, but installation stops, because 4) step has failed. 
##### Expected behavior
Installation phase exists with code 0, which means success.

prologic:
##### Summary

Fixes #7292 

##### Component Name

area/packaging

### DIFF
--- a/collectors/charts.d.plugin/Makefile.am
+++ b/collectors/charts.d.plugin/Makefile.am
@@ -31,7 +31,6 @@ dist_charts_DATA = \
 
 userchartsconfigdir=$(configdir)/charts.d
 dist_userchartsconfig_DATA = \
-    .keep \
     $(NULL)
 
 # Explicitly install directories to avoid permission issues due to umask
--- a/Makefile.am
+++ b/Makefile.am
@@ -115,10 +115,6 @@ AM_CFLAGS = \
     $(NULL)
 
 sbin_PROGRAMS =
-dist_cache_DATA = packaging/installer/.keep
-dist_varlib_DATA = packaging/installer/.keep
-dist_registry_DATA = packaging/installer/.keep
-dist_log_DATA = packaging/installer/.keep
 plugins_PROGRAMS =
 
 LIBNETDATA_FILES = \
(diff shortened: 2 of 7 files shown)


collectors/charts.d.plugin/Makefile.am
```diff
@@ -31,7 +31,6 @@ dist_charts_DATA = \
 
 userchartsconfigdir=$(configdir)/charts.d
 dist_userchartsconfig_DATA = \
-    .keep \
     $(NULL)
```
Ferroin: I'm not really an expert on automake, but do we need this variable at all anymore if it ends up being empty? Same for all the other changes like this in other files.

prologic: You're probably right; but as neither of us are experts in `autotools` or C for that matter I'll just leave this alone for now :)

PR #7839, opened 2020-01-23
Auto Assign PRs to their Authors

prologic:
##### Summary

Fixes a minor annoyance with the ZenHub UI and Github PRs where by default the Assignee of the PR is "unassigned". IHMO its better that this is the Author of the PR so its more clearly visible in "card form" :)

##### Component Name

area/ci

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -1,0 +1,1 @@
+addAssignees: author
--- a/.github/workflows/autoassign.yml
+++ b/.github/workflows/autoassign.yml
@@ -1,0 +1,10 @@
+name: 'Auto Assign'
+on: pull_request
+
+jobs:
+  auto-assign:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: kentaro-m/auto-assign-action@v1.0.1
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/autoassign.yml
+++ b/.github/workflows/autoassign.yml
@@ -1,10 +1,16 @@
-name: 'Auto Assign'
-on: pull_request
+---
+# Handles Assigning PR(s) to theit Author.
+name: Pull Request Auto Assign
+
+on:
+  schedule:
+    - cron: '*/5 * * * *'
 
 jobs:
   auto-assign:
     runs-on: ubuntu-latest
     steps:
       - uses: kentaro-m/auto-assign-action@v1.0.1
+        if: github.repository == 'netdata/netdata'
         with:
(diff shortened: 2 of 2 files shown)


prologic:
> Skips the process to add reviewers/assignees since PR type is draft

So I'll have to publish this PR to actually test it :)

thiagoftsm:
Hi @prologic ,

I remember that we had agreed months ago that the PR is not assigned for the author, because he is not the reviewer of it, but I can be wrong, so my suggestion is we assign the product team as reviewer for they give their opinion about this.

prologic:
> I remember that we had agreed months ago that the PR is not assigned for the author, because he is not the reviewer of it, but I can be wrong, so my suggestion is we assign the product team as reviewer for they give their opinion about this.

The reviewers is untouched; it's just the "Assignee" which shows in in ZenHub as I found out :)

Happy to drop this PR though if that's not what we want.

prologic:
For example (_see attached_)

Screen Shot 2020-01-24 at 07 45 15

ncmans:
This is one of the first things I've done every where I've worked. Never quite figured why there is an assignee for a PR to begin with (my only suspect is that PR derives from Issue and hence carries with it all its attributes).

thiagoftsm:
I understood what you are doing here, but I assigned to them the PR for they measure this, because it has an important impact in their work! ;)

prologic:
> I understood what you are doing here, but I assigned to them the PR for they measure this, because it has an important impact in their work! ;)

In order not to break your workflow; this action would have to leave a PR you've already assigned alone right? If 'unassigned" you would be okay with it being set to the PR author?

ilyam8:
> Never quite figured why there is an assignee for a PR to begin with

I think there is no strict definition who is assignee so it is up to the team project to decide who is assignee - pr opener or a person responsible for the specific area (who will merge/close the pr - this more when it is the PR from a contributor).

ilyam8:
So we need to define how we use the assignee field and use it this way. Lack of icon on zenhub sounds like not enough strong point for me but i am personally not against.

prologic:
> So we need to define how we use the assignee field and use it this way. Lack of icon on zenhub sounds like not enough strong point for me but i am personally not against.

I agree completely. And I won't want to sound petty here; but PR(s) I see in "Review/QA" column without a gravatar I recognise visually tend to want to make me go check if its a PR by a Contributor that we've over looked. -- That alone was the premise for exploring this in the first place and hence this PR -- I feel strongly about being responsive to both our users and contributors alike :)

prologic:
As per discussion with @cakrit I'll bring this up in our next retro.

PR #7851, opened 2020-01-24
Stream with version

### LINKED ISSUE #7804: Change the streaming protocol to allow forward-compatability.
##### Description
During the testing of the PR for issue #7488 we discovered that the new division of `system_info` into `HOST_` and `CONTAINER_` labels is incompatible with the design of the streaming protocol. The handshake for a stream connection names the structure fields as parameters in the initial request.

To change this we need to introduce a new version of the streaming protocol that transfers the two sets of labels. Unfortunately we have discovered that the changes in issue #7549 that were meant to allow forward-compatibility in the streaming protocol are not forward-compatible with further changes.

The basic problem is that the protocol message must be one of two exact strings to avoid the connection being aborted and there is no way to put a version number in the strings.

To move forward we need to redesign the streaming protocol handshake so that it is forward-compatible (and also backwards compatible with the recent changes).

#### Summary

Originally the startup message "Hit me baby, one more time..." from master to slave was used as a protocol discriminator. After #7549 it is acting as a version discriminator and hosts will disconnect when they do not recognize the message.

The protocol will be updated to perform agreement of highest compatible version during handshake. For simplicity we will switch to integer protocol versions, using the following scheme:

```
0 - original
1 - host labels
2 - host labels + docker labels
3 - host labels + docker labels + chart labels
```

The advantage of integer numbering over a bitmask scheme is that we can depend on grouping of features after negotiation, which leads to simpler code. We need to handle the three schemes in a backwards compatible way:

```
1. old slave                      -> no "NETDATA_PROTOCOL_VERSION"
2. slave with the patch #7549     -> NETDATA_PROTOCOL_VERSION=1.1
3. new slave                      -> ver=2&...
```

```
A. old master                -> "Hit me baby, one more time..."
B. master with the patch     -> "... with host labels"
C. new master               -> negotiate according to psuedo-code

master_version = 2;
if ('ver=n' was in the url  ) {
  "Hit me with your rhythm stick: %d" % min(n,master_version)
  use_version = n;
} else {
  "Hit me baby, one more time..."
   version = 0;
}
```

Falling back to the lowest version fixes a current incompatibility with systems from before patch #7549. On the client side we can map the response string back to the negotiated version.

thiagoftsm:
##### Summary
Fixes #7804 

We verified a necessity to bring an integer value that will be used to define the stream version, considering all changes that will come with labels and other subjects, this PR brings this variable to the handshake.
##### Component Name
stream
##### Additional Information
To test this PR it was used:

- This current PR
- A PR with the previous version of the stream protocol
- Netdata 1.14 to be sure that we had the first version(The `Hit me baby` version).


amoss:
I don't think the functionality of this PR matches the spec in the issue, based on what I've read in the code. It could be my misunderstanding, but I don't see the following:

* The negotiation of the lowest numbered protocol.
* The removal of the bitwise flags to allow the negotiation.

When testing it is important to add another case to the list that you are checking it against - a future version. We are implementing a forward-compatible protocol, we need to check it against a higher numbered version (in both directions).

stelfrag:
I have compile error with the command

`sudo CFLAGS="-O0 -ggdb -Wall -Wextra" ./netdata-installer.sh  --disable-go --disable-lto --dont-wait --dont-start-i
`

```
/usr/bin/ld: /tmp/netdata.vl2dqB.ltrans0.ltrans.o: in function `get_system_info':
/home/stelios/git/netdata/daemon/main.c:768: undefined reference to `coverity_remove_taint'
/usr/bin/ld: /home/stelios/git/netdata/daemon/main.c:769: undefined reference to `coverity_remove_taint'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:3720: netdata] Error 1
make[2]: Leaving directory '/home/stelios/git/netdata'
make[1]: *** [Makefile:4768: all-recursive] Error 1
make[1]: Leaving directory '/home/stelios/git/netdata'
```

stelfrag:
> I have compile error with the command
> 
> `sudo CFLAGS="-O0 -ggdb -Wall -Wextra" ./netdata-installer.sh --disable-go --disable-lto --dont-wait --dont-start-i `
> 
> ```
> /usr/bin/ld: /tmp/netdata.vl2dqB.ltrans0.ltrans.o: in function `get_system_info':
> /home/stelios/git/netdata/daemon/main.c:768: undefined reference to `coverity_remove_taint'
> /usr/bin/ld: /home/stelios/git/netdata/daemon/main.c:769: undefined reference to `coverity_remove_taint'
> collect2: error: ld returned 1 exit status
> make[2]: *** [Makefile:3720: netdata] Error 1
> make[2]: Leaving directory '/home/stelios/git/netdata'
> make[1]: *** [Makefile:4768: all-recursive] Error 1
> make[1]: Leaving directory '/home/stelios/git/netdata'
> ```

But of course it compiles with any non **-O0**  flag..  *argh*

amoss:
Ah, that sounds like something I have broken. Why does '-O0' cause it to fail?

stelfrag:
> Ah, that sounds like something I have broken. Why does '-O0' cause it to fail?

I have no clue. Its that same LTO thing I guess.  Even `--enable-lto` doesn't fix it.

thiagoftsm:
> I don't think the functionality of this PR matches the spec in the issue, based on what I've read in the code. It could be my misunderstanding, but I don't see the following:
> 
>     * The negotiation of the lowest numbered protocol.
> 
>     * The removal of the bitwise flags to allow the negotiation.
> 
> 
> When testing it is important to add another case to the list that you are checking it against - a future version. We are implementing a forward-compatible protocol, we need to check it against a higher numbered version (in both directions).

You are correct, I apologize, because I really miss this specific change, now with the last commit everything should be correct.

thiagoftsm:
> But of course it compiles with any non **-O0** flag.. _argh_

I always compile with **-O1** and I do not have problem, I will investigate this during the weekend.

streaming/rrdpush.c
```diff
@@ -1174,10 +1174,10 @@ static int rrdpush_receive(int fd
 
     info("STREAM %s [receive from [%s]:%s]: initializing communication...", host->hostname, client_ip, client_port);
     char initial_response[HTTP_HEADER_SIZE];
-    if (stream_flags & STREAM_HAS_INTEGER_VERSION) {
+    if (stream_version > 2) {
         info("STREAM %s [receive from [%s]:%s]: Netdata is using the newest stream protocol.", host->hostname, client_ip, client_port);
-        sprintf(initial_response, "%s%u", START_STREAMING_PROMPT_VN, STREAMING_PROTOCOL_CURRENT_VERSION);
-    } else if (stream_flags & LABEL_FLAG_UPDATE_STREAM) {
+        sprintf(initial_response, "%s%u", START_STREAMING_PROMPT_VN, stream_version);
+    } else if (stream_version > 0) {
```
ilyam8: Are `old`, `second` and `newest`  good names? Why do not use version numbers.

And `stream_version > 0` condition looks suspicious, i guess it can be only `1`. Am i right?

Where could i read about stream versioning? Is there a document or an discussion in some issue at least?

thiagoftsm: > Are `old`, `second` and `newest` good names? Why do not use version numbers.

I removed the words `second` and `newest` and I am using version now.

> And `stream_version > 0` condition looks suspicious, i guess it can be only `1`. Am i right?

We are in the third version already, unfortunately due the previous versions do not have the version explicit, we "lost" the count. You were right, I also had a mistake, but I fixed with another commit.

> Where could i read about stream versioning? Is there a document or a discussion in some issue at least?

This was not documented in the past, but you can see more details about where we are in this link https://github.com/netdata/netdata/issues/7804#issue-552508664 .

ilyam8: Ok, i see, so it is `0-3`.

streaming/rrdpush.c
```diff
@@ -641,26 +648,41 @@ static int rrdpush_sender_thread_connect_to_master(RRDHOST *host, int default_po
...
-        host->labels_flag &= ~LABEL_FLAG_STOP_STREAM;
+    http[received] = '\0';
+    int answer = -1;
+    char *version_start = strchr(http, '=');
+    if(version_start) {
+        version_start++;
+        answer = memcmp(http, START_STREAMING_PROMPT_VN, (size_t)(version_start - http));
+        if(!answer) {
+            rrdpush_set_flags_to_newest_stream(host);
+            host->stream_version = (uint32_t)strtol(version_start, NULL, 10);
```
stelfrag: In this case the slave doesn't report the version that was negotiated

streaming/rrdpush.c
```diff
@@ -641,26 +648,41 @@ static int rrdpush_sender_thread_connect_to_master(RRDHOST *host, int default_po
...
+        }
     } else {
-        answer = strncmp(http, START_STREAMING_PROMPT, strlen(START_STREAMING_PROMPT));
+        answer = memcmp(http, START_STREAMING_PROMPT_V2, strlen(START_STREAMING_PROMPT_V2));
         if(!answer) {
-            host->labels_flag |= LABEL_FLAG_STOP_STREAM;
-            host->labels_flag &= ~LABEL_FLAG_UPDATE_STREAM;
-            info("STREAM %s [send to %s]: is using an old Netdata.", host->hostname,  connected_to);
+            rrdpush_set_flags_to_newest_stream(host);
```
stelfrag: There are identical messages that the slave uses an old Netdata but the check has been changed to START_STREAMING_V2

database/rrd.h
```diff
@@ -46,6 +46,8 @@ extern time_t rrdset_free_obsolete_time;
 #define RRDSET_MAGIC        "NETDATA RRD SET FILE V019"
 #define RRDDIMENSION_MAGIC  "NETDATA RRD DIMENSION FILE V019"
 
+#define STREAMING_PROTOCOL_CURRENT_VERSION (unsigned int)2
```
stelfrag: I am being picky, I know :slightly_smiling_face: 
this is unsigned int and in the data structure defined as uint32_t

streaming/rrdpush.c
thiagoftsm: Hi @stelfrag ,

This part of the code  has relationship with the `master` response. As you can see in the line `664`, I am bringing a termination for the response received, because the unique possibility for we have equals in the answer is the master to work with the version bigger or equal `2`. This is the motive that I start testing the presence of equal before to move for the oldest versions. 
Right now the majority of Netdata installed are the oldest, but in a near future this will invert, so I decided to put the test to new before any other test.

best regards!

streaming/rrdpush.c
thiagoftsm: The beginning is 100% equal, and this was done to create "message compatibility" (It is between quotes, because I think I generalized the concept ).
The system won't fail when do these kind of tests, because we are using `memcmp` with the length we want to compare.
Finally, we have this order, because I hope, that in a short period of time, nobody will use the oldest Netdata stream protocol.

streaming/rrdpush.c
stelfrag: Yes that's fine but in that case the slave doesn't report with what version it connected. 
It would be nice to have an indication in line 695

```    
info("STREAM %s [send to %s]: established communication - ready to send metrics...", host->hostname, connected_to);
```
That will say with which version it connected. Now with identical versions in master and slave, I get info on the master (for example`Netdata is using the newest stream protocol`) but not in the slave

database/rrd.h
thiagoftsm: According the definition of `bits/types.h` `uint32_t` is a ` unsigned int`, but I agree that is more readable to set as `uint32_t` here, so I changed in the last commit! :) 
You are not being "picky", you are doing the correct, but I confess sometimes I am not so rigorous, because I am thinking more in the ASM code instead of the code to be readable .

streaming/rrdpush.c
thiagoftsm: I loved your idea, I brought it with the last commit. :)
Thank you very much!

streaming/rrdpush.c
amoss: But we are trying to negotiate the highest version that both ends understand?

thiagoftsm: For while yes, but when we begin to move from protocol version 2 in front, we will use the algorithm specified in the issue 7804.

stelfrag:
Part 1 of the checks to see if the master responds OK to version from a slave

#### Running agent with latest version and having slave connect:

- With old netdata
   - we get the old protocol negotiation **Hit me baby, push them over...**
- With current version
   - we get the host labels protocol negotiation **Hit me baby, push them over and bring the host labels...**
- With current version (passing ver=2 -- attempt to use current protocol version)
   - we get the host labels protocol negotiation **Hit me baby, push them over and bring the host labels...**
      - This is because the NETDATA_PROTOCOL_VERSION is present and ver is ignored
- With current version (passing ver=6 -- fake higher version) and skipping NETDATA_PROTOCOL_VERSION
   - we get the highest version the agent can support **Hit me baby, push them over with the version=2**
      - This is because current version is 2

#### Running agent with fake version *4* and having slave connect:
- With old netdata
   - we get the old protocol negotiation **Hit me baby, push them over...**
- With current version
   - we get the host labels protocol negotiation **Hit me baby, push them over and bring the host labels...**
- With current version (passing ver=2 -- attempt to use current protocol version)
   - we get the host labels protocol negotiation **Hit me baby, push them over and bring the host labels...**
      - This is because the NETDATA_PROTOCOL_VERSION is present and ver is ignored
- With current version (passing ver=6 -- fake higher version) and skipping NETDATA_PROTOCOL_VERSION
   - we get the highest version the agent can support **Hit me baby, push them over with the version=4**
      - This is because the master has fake version 4

thiagoftsm:
@stelfrag and @vlvkobal I brought a commit that should fix the problems reported, please, take a look when you have a time.

Thank you!

vlvkobal:
I tested it in all possible version combinations (including fake future versions). Handshake is working and versions are correct.

stelfrag:
I started a master and slave with the latest patches and in the logfile I get that both servers use version 2, however in the logfile in the master I get: 
```
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1d5239d4-0582-11ea-bb8e-7f6153c1117c', which is not used.
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'debian1', which is not used.
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'debian1', which is not used.
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'ba261d98-0578-11ea-b7ed-5254005bbf93', which is not used.
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1', which is not used.
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'linux', which is not used.
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'Europe/Athens', which is not used.
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '2', which is not used.
2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1.1', which is not used.
```
The incoming slave connection pushes:
```
STREAM key=1d5239d4-0582-11ea-bb8e-7f6153c1117c&hostname=debian1&registry_hostname=debian1&machine_guid=ba261d98-0578-11ea-b7ed-5254005bbf93&update_every=1&os=linux&timezone=Europe/Athens&tags=&ver=2&NETDATA_SYSTEM_OS_NAME=Debian GNU/Linux&NETDATA_SYSTEM_OS_ID=debian&NETDATA_SYSTEM_OS_ID_LIKE=unknown&NETDATA_SYSTEM_OS_VERSION=10 (buster)&NETDATA_SYSTEM_OS_VERSION_ID=10&NETDATA_SYSTEM_OS_DETECTION=/etc/os-release&NETDATA_SYSTEM_KERNEL_NAME=Linux&NETDATA_SYSTEM_KERNEL_VERSION=4.19.0-6-amd64&NETDATA_SYSTEM_ARCHITECTURE=x86_64&NETDATA_SYSTEM_VIRTUALIZATION=kvm&NETDATA_SYSTEM_VIRT_DETECTION=systemd-detect-virt&NETDATA_SYSTEM_CONTAINER=unknown&NETDATA_SYSTEM_CONTAINER_DETECTION=none&NETDATA_PROTOCOL_VERSION=1.1 HTTP/1.1
User-Agent: netdata/v1.19.0-425-g9dce6e85
Accept: */*
```

streaming/rrdpush.c
```diff
@@ -1396,31 +1427,34 @@ int rrdpush_receiver_thread_spawn(RRDHOST *host, struct web_client *w, char *url
...
+            name = "NETDATA_HOST_OS_VERSION_ID";
+        else if (!strcmp(name, "NETDATA_SYSTEM_OS_DETECTION"))
+            name = "NETDATA_HOST_OS_DETECTION";
+        else if(!strcmp(name, "NETDATA_PROTOCOL_VERSION") && stream_version == UINT_MAX) {
+            stream_version = 1;
+        }
+
+        if (unlikely(rrdhost_set_system_info_variable(system_info, name, value)) && stream_version != 1) {
+            info("STREAM [receive from [%s]:%s]: request has parameter '%s' = '%s', which is not used.",
                      w->client_ip, w->client_port, key, value);
```
stelfrag: This produces a lot of info messages if the version is not 1 (so as we are currently in version 2) this will always trigger log entries. Also you probably wanted to log `name, value` instead of `key, value`

stelfrag:
Quick summary as I am testing this in detail: 

- The stream version starts `undefined` (UINT_MAX) and if nothing changes (no `ver` parameter or  `NETDATA_PROTOCOL_VERSION` found) it will default to the oldest/original version which is 0. :+1: 

- If a `ver` keyword is sent by the slave the version is set to be the minimum of whatever the slave wants and the master currently supports :+1: 

- If no `ver` is found (ie. ver is `undefined`, and a `NETDATA_PROTOCOL_VERSION` key is found, it switches to version 1 :+1:

streaming/rrdpush.c
```diff
@@ -28,6 +28,7 @@
 #define STREAMING_PROTOCOL_VERSION "1.1"
```
mfundul: What's the relationship of this with `STREAMING_PROTOCOL_CURRENT_VERSION`?

stelfrag: I think this is the "in between" protocol version that simply sets the NETDATA_PROTOCOL_VERSION and its out there now -- it is ignored if a `ver` is present during the stream establishment

thiagoftsm:
> I started a master and slave with the latest patches and in the logfile I get that both servers use version 2, however in the logfile in the master I get:
> 
> ```
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1d5239d4-0582-11ea-bb8e-7f6153c1117c', which is not used.
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'debian1', which is not used.
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'debian1', which is not used.
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'ba261d98-0578-11ea-b7ed-5254005bbf93', which is not used.
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1', which is not used.
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'linux', which is not used.
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'Europe/Athens', which is not used.
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '2', which is not used.
> 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1.1', which is not used.
> ```

> ```

Last Friday @vlvkobal and I was discussing about reorganization of the code and we missed this, I am returning the code with the last commit.

streaming/rrdpush.c
thiagoftsm: I fixed this with the last commit.

streaming/rrdpush.c
thiagoftsm: The original idea was to have conditions to use versions like the kernel version, but after a talk with Andrew we moved it to an integer version, so as @stelfrag said this is between stream versions.

stelfrag:
> > I started a master and slave with the latest patches and in the logfile I get that both servers use version 2, however in the logfile in the master I get:
> > ```
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1d5239d4-0582-11ea-bb8e-7f6153c1117c', which is not used.
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'debian1', which is not used.
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'debian1', which is not used.
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'ba261d98-0578-11ea-b7ed-5254005bbf93', which is not used.
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1', which is not used.
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'linux', which is not used.
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = 'Europe/Athens', which is not used.
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '2', which is not used.
> > 2020-02-03 12:06:14: netdata INFO  : WEB_SERVER[static5] : STREAM [receive from [192.168.122.129]:48308]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1.1', which is not used.
> > ```
> 
> > 
> 
> Last Friday @vlvkobal and I was discussing about reorganization of the code and we missed this, I am returning the code with the last commit.

I am still getting one message

```
2020-02-03 16:41:40: netdata INFO  : WEB_SERVER[static3] : STREAM [receive from [192.168.122.129]:49600]: request has parameter '1d5239d4-0582-11ea-bb8e-7f6153c1117c' = '1.1', which is not used.
```
Is it intentional to log the key or the name of the parameter?

thiagoftsm:
> Is it intentional to log the key or the name of the parameter?

No this was bug, thank you!
I fixed with the last commit.

stelfrag:
> > Is it intentional to log the key or the name of the parameter?
> 
> No this was bug, thank you!
> I fixed with the last commit.

Ok, I will recheck when you push the change

streaming/rrdpush.c
```diff
@@ -1396,31 +1427,36 @@ int rrdpush_receiver_thread_spawn(RRDHOST *host, struct web_client *w, char *url
...
+                name = "NETDATA_HOST_OS_VERSION_ID";
+            else if (!strcmp(name, "NETDATA_SYSTEM_OS_DETECTION"))
+                name = "NETDATA_HOST_OS_DETECTION";
+            else if(!strcmp(name, "NETDATA_PROTOCOL_VERSION") && stream_version == UINT_MAX) {
+                stream_version = 1;
+            }
+
+            if (unlikely(rrdhost_set_system_info_variable(system_info, name, value))) {
+                info("STREAM [receive from [%s]:%s]: request has parameter '%s' = '%s', which is not used.",
                      w->client_ip, w->client_port, key, value);
```
stelfrag: @thiagoftsm  I still see that you put in the logfile `w->client_ip, w->client_port, key, value`. Did you mean to write the actual parameter (`name`) that is not used?

thiagoftsm: Hi @stelfrag , I did not created this message, this is in Netdata since a long time ago, so I decided to keep it while I was organizing the code.
But, I liked your idea, I have only one doubt. Should we not create a separate PR for this?

stelfrag: Ok then I don't have a strong case about fixing it now. It will report keys when a new version of the streaming protocol is out and the slave is updated before the master :+1:

streaming/rrdpush.c
```diff
@@ -643,26 +650,44 @@ static int rrdpush_sender_thread_connect_to_master(RRDHOST *host, int default_po
...
+    int answer = -1;
+    char *version_start = strchr(http, '=');
+    uint32_t version;
+    if(version_start) {
+        version_start++;
+        version = (uint32_t)strtol(version_start, NULL, 10);
+        answer = memcmp(http, START_STREAMING_PROMPT_VN, (size_t)(version_start - http));
+        if(!answer) {
+            rrdpush_set_flags_to_newest_stream(host);
+            host->stream_version = (uint32_t)strtol(version_start, NULL, 10);
```
stelfrag: You already compute the stream_version in line 672. Maybe use that?

streaming/rrdpush.c
ilyam8: @stelfrag you mean it will log api key?

stelfrag: > @stelfrag you mean it will log api key?

Yes with the api key -- I edited the comment -- thanks @ilyam8 !!

ilyam8: Keep in mind that netdata error log is word readable, if api key is private it shouldnt be logged

thiagoftsm: A very good point @ilyam8 , you convinced me to use name now.

streaming/rrdpush.c
thiagoftsm: You are right, thank you!
The last commit removes the unnecessary call.

streaming/rrdpush.c
thiagoftsm: I changed the message using `name` instead the `api key`.
Thank you guys!

amoss: If there are any connection errors on the stream then the api key is logged in the error message (in rrdpush.c e.g. L1462), if this is a problem then please raise an issue to fix it. I suspect the correct approach is that the error log should not be world readable.

thiagoftsm: @amoss , this will be subject for another PR. The sre team was working days ago with installation permissions, probably we need to talk with them about it.

amoss: That sounds good - let's have a chat within the team about the pros/cons of making the error log not world writable first. If the team thinks that it is a good idea then we can raise an issue for the SRE team to look at the file permissions.

ilyam8: Dunno, anyway private info in error.log in very questionable idea. Changing perms doesnt help. We pretty often ask user to share logs and they do it private. It means that we shouldnt log any private things there. At least for now when we have one error.log for everything.

Issue #7124, opened 2019-10-17
unbound queries data-desired-units


mkoubar:
@ilyam8 can we add data-desired-units feature to unbound section so that number of queries will not that huge because after running for 1-day numbers will be as below
![queries](https://user-images.githubusercontent.com/16787802/67045976-03756a80-f138-11e9-8b07-0092b1c1d3eb.png)

ilyam8:
@Ferroin 

I think we need to make dimensions of this chart `incremental` and change units to `queries/s`. I think total number of queries is not that useful. What do you think?

ilyam8:
ok, Process Queries chart is wrong, we need so split it at least in to 3 charts

```
# Total Queries
total.num.queries=1249

# Cache
total.num.cachehits=235
total.num.cachemiss=1014

# Queries By Type
total.num.queries_ip_ratelimited=0
total.num.prefetch=3
total.num.zero_ttl=0
total.num.recursivereplies=1014
```

The question is - sum of queries by type is less them total number of queries? What are we missing?

ilyam8:
Ok, it seems we need to do some research about unbound metrics. Current implementation has some problems. We should collect more metrics and group them differently.

Metrics w/ description:
 - https://docs.datadoghq.com/integrations/unbound/#metrics


@Ferroin dont hurry with PR, lets discuss first.

mkoubar:
FYI : @ilyam8 @Ferroin when statistics-cumulative: yes extended-statistics: yes are activated in unbound.conf 
the result of unbound-control  stats_noreset will give all these results below:
thread0.num.queries=169493
thread0.num.queries_ip_ratelimited=0
thread0.num.cachehits=152164
thread0.num.cachemiss=17329
thread0.num.prefetch=2633
thread0.num.zero_ttl=0
thread0.num.recursivereplies=17322
thread0.requestlist.avg=1.87261
thread0.requestlist.max=19
thread0.requestlist.overwritten=0
thread0.requestlist.exceeded=0
thread0.requestlist.current.all=4
thread0.requestlist.current.user=4
thread0.recursion.time.avg=0.098119
thread0.recursion.time.median=0.0489934
thread0.tcpusage=0
thread1.num.queries=172811
thread1.num.queries_ip_ratelimited=0
thread1.num.cachehits=155926
thread1.num.cachemiss=16885
thread1.num.prefetch=2540
thread1.num.zero_ttl=0
thread1.num.recursivereplies=16883
[...]

PR #7853, opened 2020-01-24
/collectors/python.d: remove unbound module

ilyam8:
### Summary

This PR removes [python.d unbound module](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/unbound).

It was deprecated in https://github.com/netdata/netdata/pull/7365 because of https://github.com/netdata/netdata/issues/7124

There is no sense to have old python code in the repo, we remove it but it will remain in the git history and in our ❤️ forever!

##### Component Name

[`/collectors/python.d.plugin`](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin)

##### Additional Information

Python module has been deprecated since [25 nov 2019](https://github.com/netdata/netdata/commit/20aa91bd108d6e9f0e945b4e575429d2e7c2c081), `python.d.plugin` skips the module during module loading. It is just dead code atm.

https://github.com/netdata/netdata/blob/db265d6f63182cd5ee58bf6f76cb50a5f41dd263/collectors/python.d.plugin/python.d.plugin.in#L129-L139

https://github.com/netdata/netdata/blob/db265d6f63182cd5ee58bf6f76cb50a5f41dd263/collectors/python.d.plugin/python.d.plugin.in#L143


Ferroin:
While I agree with the general sentiment, I'm largely of the opinion that we should be keeping the Python collectors around at least until Go becomes the default in place of Python (and also until we have the Gentoo stuff working properly and including the Go plugin).

ilyam8:
> until Go becomes the default in place of Python

It is already. Python version is:

- buggy, no reason to use it
- deprecated, it is impossible to run it, it is hardcoded in the python.d.plugin

ilyam8:
Since Nov 25 2019

> https://github.com/netdata/netdata/pull/7365/commits/20aa91bd108d6e9f0e945b4e575429d2e7c2c081

Ferroin:
> > until Go becomes the default in place of Python
> 
> It is already. Python version is:
> 
> * buggy, no reason to use it
> * deprecated, it is impossible to run it, it is hardcoded in the python.d.plugin

So, IOW, we're automatically running the Go plugin instead of the Python one on new installs? Because that's what I mean by Go being the default, not just using it preferentially for this collector, because right now anybody who hasn't explicitly switched to using Go will be losing functionality if we remove this (myself included, there's still no automatic support for the Go plugin on Gentoo).

ilyam8:
> So, IOW, we're automatically running the Go plugin instead of the Python one on new installs?

Yes. As i said python version is deprecated since 25 Nov 2019. We did as an exception for unbound because it was the fix - python has a lot of problems and the fix was deprecate it and rewrite in go.

ilyam8:
> on Gentoo

it seems same on FreeBSD, see

> https://github.com/netdata/go.d.plugin/issues/319#issuecomment-573642303

Ferroin:
Keep in mind also that on both Gentoo and FreeBSD, it's rather easy to revert 20aa91b with a local patch on install, and it's actually far easier than dealing with handling the Go plugin outside of package management.

On the Gentoo side of things, once I finally get some time to properly test it, it should be pretty easy to get an ebuild working for it, at which point I"m not worried on a personallevel about merging this (though it would be good to wait until we get the overlay officially listed).

On the FreeBSD side I'm not sure what the best approach is, but I think it will be to do a separate port for the Go plugin (just like how I'm doing a separate ebuild on Gentoo) so that versioning is tracked properly, but that's beyond my expertise.

ilyam8:
Ok, let me make it clear - python version is broken, i wouldnt want people to use it. **It's been deprecated for 2 month already** - it just doesnt work. You can:

- install netdata from source (it will download go.d.plugin)
- download it manually
- rename your py `unbound` to get it working again

Ferroin:
> Ok, let me make it clear - python version is broken, i wouldnt want people to use it. **It's been deprecated for 2 month already** - it just doesnt work.

Not to be facetious, but the functional Unbound charts on the Netdata dashboards of my systems indicate otherwise. It doesn't work in a lot of cases and does have bugs, but it still works fine for some people. Yes, we should discourage people from using it (and marking it obsolete is a great way to do so), but removing the functionality when we know the new option doesn't work yet for some people without putting a significant burden on them to use it is not particularly user friendly.

Note that I'm not saying we should do anything like actually reverting 20aa91b on master, just that I feel we should just leave this code alone until we get the Go plugin working on Gentoo (which we'll hopefully be supporting officially in the near future) and FreeBSD (which we already officially support).

>  You can:
> 
> * install netdata from source (it will download go.d.plugin)

We're actively trying to move away from telling people to do this.

> * download it manually

Same as above, and it's also a pain to manage because you have to manually deal with keeping it up to date.

> * rename your py `unbound` to get it working again

That's only a solution until this PR is merged.

I wasn't kidding though about it being easy to patch Netdata on install on Gentoo, there's specific functionality in Portage for applying local patches to the sources during the install process before the build gets run, and this one is a one-line patch.

ilyam8:
> but removing the functionality when we know the new option doesn't work yet for some people without putting a significant burden on them to use it is not particularly user friendly.

Who are those people you are talking about? 

There is no option, py version doesn work on old and new installs.

> but it still works fine for some people

We can discuss it forever, as is said it was an exception for unbound, becaue to fix it i (or you) would need to rewrtie it from scratch. In my opinion it doesnt work at all.

ilyam8:
@Ferroin 

Current official flow for **a new** Gentoo user (it doesnt affect old install, it doesnt remove python version) who is interesed in `Unbound` monitoring:

- install netdata via Gentoo pkg manager
- go to netdata documentation
- see that python version is deprecated
- download `go.d.plugin` (**once** you dont need to keep it up to date)
- monitor `Unbound`

i've fixed a lot of bug in go version, it collects more metrics, does correct grouping, it has alarms, it has tests, etc.

If you think that there are people that cant for some reason follow official way and they need python version then please:

- fix python version (use go version as an example)
- add a note to the python readme how to activate python version (rename)

If you do it i am ok with `wontfix` this PR. Think if it worth it.

For now i see no reason to not merge it.

joelhans:
I have no stakes in the conversation above, just one question—is it necessary to completely remove mention of Unbound from the main README and `what-is-netdata.md`? Couldn't you just change the link to point to the Go collector?

ilyam8:
> is it necessary to completely remove mention of Unbound from the main README and what-is-netdata.md

it is necessary to remove mention of python version since i am removing it.

> Couldn't you just change the link to point to the Go collector?

I could, but i noticed that `README.md` and `Add-more-charts-to-netdata.md` are outdated and we need to fix them - there is no poing to add only unbound. I will do it in another PR.

ilyam8:
@Ferroin waiting for you answer https://github.com/netdata/netdata/pull/7853#issuecomment-578264088

Issue #150, opened 2016-04-03
install script shebang portability


bakman2:
I am trying to install this on a synology, the bash paths are not in default directories.
Could you change the scripts to using more portable shebangs ie.
```#!/usr/bin/env bash```

ktsaou:
I don't know if this the most portable way.
@alonbl @philwhineray any ideas?

alonbl:
As this is not a standard shell, but bash, for these that are bash specific I would be flexible.
The other method of doing that would be to generate a stub /usr/libexec/xxx/run which will be symlinked by all our scripts /usr/libexec/xxx/firehol->run, this will in turn execute @BASH@ $0.sh "$@", the advantage is that the main script files can be written without being a template.

bakman2:
I figured that there is no quick/dirty way to install it on Synology, as there is no zlib-dev available. 
In the end, it was easier to install chroot and install from there (i have updated the wiki with this info)

bakman2:
I'll close this one.

ktsaou:
> I figured that there is no quick/dirty way to install it on Synology, as there is no zlib-dev available. 
> In the end, it was easier to install chroot and install from there (i have updated the wiki with this info)

If the zlib dependency is a problem, I think the zlib folks have a way of adding it to embeddable applications (without re-distributing their entire source package). If you need this, open a github issue and I'll give it a try at some point.

PR #7862, opened 2020-01-27
Update `api/v1/info `

### LINKED ISSUE #7751: Add the new meta-data from system-info.sh to the web API and ACLK.
# Feature idea summary

The `system-info.sh` script will be updated in #7731 to add the following new information

*    Number of cores
*    Nominal frequency (not the current frequency)
*    Amount of RAM
*    Sum of available capacity on disks

The format of these new environment variables will be defined. The data needs to be read into `main.c` where the script is called, stored in an appropriate place and then included in the `/api/v1/info` response as new fields, and in the `info` metadata message on the ACLK. For this task only the web API handler needs to be updated - the ACLK functionality calls that handler.

##### Expected behavior

We should be able to see the new information in `/api/v1/info` queries on a range of systems that includes different processor architectures (ARM, AMD, Intel) and different types of netdata installation (bare-metal, VM, container).

thiagoftsm:
##### Summary
Fixes #7751 

This PR brings to the API the new variables requested
##### Component Name
Web
##### Additional Information
To confirm that they are present, run in the terminal the command `daemon/system-info.sh` and do a request for the address `http://localhost:19999/api/v1/info`, the variables `NETDATA_CPU_LOGICAL_CPU_COUNT`, `NETDATA_CPU_FREQ` and NETDATA_TOTAL_RAM` must be in the HTTP response with the same values, but with different names.


web/api/web_api_v1.c
```diff
@@ -814,6 +814,9 @@ inline int web_client_api_request_v1_info(RRDHOST *host, struct web_client *w, c
     buffer_sprintf(wb, "\t\"os_version\": \"%s\",\n", (host->system_info->host_os_version) ? host->system_info->host_os_version : "");
     buffer_sprintf(wb, "\t\"os_version_id\": \"%s\",\n", (host->system_info->host_os_version_id) ? host->system_info->host_os_version_id : "");
     buffer_sprintf(wb, "\t\"os_detection\": \"%s\",\n", (host->system_info->host_os_detection) ? host->system_info->host_os_detection : "");
+    buffer_sprintf(wb, "\t\"cores_total\": \"%s\",\n", (host->system_info->host_cores) ? host->system_info->host_cores : "");
+    buffer_sprintf(wb, "\t\"cpu_freq\": \"%s\",\n", (host->system_info->host_cpu_freq) ? host->system_info->host_cpu_freq : "");
+    buffer_sprintf(wb, "\t\"ram_total\": \"%s\",\n", (host->system_info->host_ram_total) ? host->system_info->host_ram_total : "");
```
stelfrag: Is there a specific reason that we expose (and store cores, ram and cpu freq) as strings? 
In general everything is exposed as strings (except the alarms object)

stelfrag:
It exposes the extra information as seen below
```
	"cores_total": "12",
	"cpu_freq": "2600000",
	"ram_total": "16624017408",
	"container_os_name": "none",
	"container_os_id": "none",
	"container_os_id_like": "none",
	"container_os_version": "none",
	"container_os_version_id": "none",

	"kernel_name": "Linux",
	"kernel_version": "5.3.0-26-generic",
	"architecture": "x86_64",
	"virtualization": "none",
	"virt_detection": "systemd-detect-virt",
	"container": "unknown",
	"container_detection": "none",
	"host_labels": {
		"_is_master": "true",
		"_virt_detection": "systemd-detect-virt",
		"_container_detection": "none",
		"_container": "unknown",
		"_virtualization": "none",
		"_architecture": "x86_64",
		"_kernel_version": "5.3.0-26-generic",
		"_os_version": "19.10 (Eoan Ermine)",
		"_os_name": "Ubuntu"
	},
```
What about exposing additional information to host_labels?

web/api/web_api_v1.c
thiagoftsm: Since I began to work with Netdata, this was the pattern, I believe this happens, because the protocol used to send the response is a text protocol by definition.

thiagoftsm:
> What about exposing additional information to host_labels?

By steps, when I grow up I wanna have a computer with this number of cores.
I did not added it as labels according the current host labels definition, this would be possible, but we would need to change some concepts, before to move in front.

web/api/web_api_v1.c
Ferroin: Pretty certain we're using JSON, which absolutely supports integer and floating-point values.

thiagoftsm: You are correct @Ferroin , we are using JSON that has these supports, an important question for both, before I bring possible changes. Case we decide to do this on Netdata side, this will bring more processing to the web server every time that it generates the response, should we do this?

Ferroin: I just remembered why we would want these to be strings. If the script that gives this data to the core can't figure out values for these, it returns 'unknown' as a string, so we need to support passing that on.

thiagoftsm: All right, I will keep it as text for while.

amoss:
> What about exposing additional information to host_labels?

This is a good point, it is not obvious what the most useful way would be. I've asked in product issue #150 to find out.

stelfrag:
> > What about exposing additional information to host_labels?
> 
> This is a good point, it is not obvious what the most useful way would be. I've asked in product issue #150 to find out.

@thiagoftsm if changes are requested due to host labels, I will review again

thiagoftsm:
> > > What about exposing additional information to host_labels?
> > 
> > 
> > This is a good point, it is not obvious what the most useful way would be. I've asked in product issue #150 to find out.
> 
> @thiagoftsm if changes are requested due to host labels, I will review again

Thank you ! :)

thiagoftsm:
> @thiagoftsm if changes are requested due to host labels, I will review again

I had to rename the variables and add  the disk size, but there was not changes related to the labels yet.

database/rrdhost.c
```diff
@@ -569,6 +569,10 @@ void rrdhost_system_info_free(struct rrdhost_system_info *system_info) {
         freez(system_info->host_os_version);
         freez(system_info->host_os_version_id);
         freez(system_info->host_os_detection);
+        freez(system_info->host_cores);
+        freez(system_info->host_cpu_freq);
+        freez(system_info->host_ram_total);
+        freez(system_info->host_disk_space);
```
mfundul: I don't like how `rrdhost_set_system_info_variable()` always frees whatever is in every `system_info` field, relying on the undocumented fact that if something hasn't been set yet it should be NULL.

database/rrdhost.c
```diff
@@ -1181,6 +1201,22 @@ int rrdhost_set_system_info_variable(struct rrdhost_system_info *system_info, ch
         freez(system_info->kernel_name);
         system_info->kernel_name = strdupz(value);
     }
+    else if(!strcmp(name, "NETDATA_SYSTEM_CPU_LOGICAL_CPU_COUNT")){
```
mfundul: I don't like how rrdhost_set_system_info_variable() always frees whatever is in every system_info field, relying on the undocumented fact that if something hasn't been set yet it should be NULL.

amoss: Why is this a problem? What do you think will go wrong?

mfundul: Somebody could call `rrdhost_set_system_info_variable()` without having zeroed the structure.

Also, `freez()` is not documented to handle NULL arguments.

mfundul: One solution is to document both functions. E.g.

```
// system_info fields must be allocated or NULL
rrdhost_set_system_info_variable(struct rrdhost_system_info *system_info...)

// If ptr is NULL, no operation is performed.
freez()

```

thiagoftsm: I understood your point of view @mfundul , and I will add an `if` before to free the information.
We did not change this before, because this was the original code and we only brought the necessary variables.

thiagoftsm: Something analogous what I am bringing is also done inside health

mfundul: Wait, you don't need an `if`, what do you mean exactly?

thiagoftsm: I commit now the changes, please, take a look.
This pattern that I brought here is used inside `health/health_config.c`, but it was not present in this file.

mfundul: I saw the change, this doesn't solve anything:
1. If the `system_info` structure is NOT zeroed out you will corrupt the memory anyway
2. If the `system_info` structure IS zeroed out then you gain nothing because `freez()` was already safe because it calls `free()` inside that is safe according to documentation.

thiagoftsm: 1 . System_info is zeroed inside `stream/rrdpush.c` and `daemon/main.c`, there is not way Netdata will arrive in this point with any value.
2 . Sure, but we have two different "patterns" on our code, one used in health, that is possible to be reloaded during runtime, and another here. I added the `if` for we have only one pattern in our code. I agree that it was not necessary to add `ifs` now, but it can be in the future.

mfundul: 1. That's the point, we may add calls in the future.
2. The `if`s don't add anything, cause if it's NULL `freez()` works anyway. If it's random garbage, then the `freez()` will corrupt the memory equally with or without an `if`.

thiagoftsm: I agreed with `1.` and `2.`, so I am keeping the `if` for case we start calling this function during runtime, and I also added the suggested documentation.

thiagoftsm:
> Unexpected environment variable

Good catch, thank you very much!
The last commit removes the warnings.

ilyam8:
> Unexpected environment variable

this fix doesnt related to the PR, does it?

thiagoftsm:
> > Unexpected environment variable
> 
> this fix doesnt related to the PR, does it?

The last commit fixes the side effect. It has relationship, because the original PR brought the variables that we needed, but the script also has few new variables that won't be in the response, so I am removing from the error.log the variables that are not listed in the documentation.

ilyam8:
i see you added 5 `else if` in the last commit but i have 9 `Unexpected environment variable`.

thiagoftsm:
> i see you added 5 `else if` in the last commit but i have 9 `Unexpected environment variable`.

Please, give us the list. The `5 if` are related with the variable that won't be inside the response, but I also added other variables that are part of the response, so all the 9 should be declared in the `if .. else if` groups.

database/rrdhost.c
amoss: Please remove the ifs again - they are redundant as there is a check inside `freez()`. @mfundul While your point is technically true it is not helpful in this case. The whole of the Netdata codebase operates on the assumption that `callocz` is used to allocate data-structures specifically so that they are zero'd out as their initial state. There is no code-path in which the contents of that structure are uninitialized. Whilst you may disagree with this policy as it exists globally within the entire code-base - this PR is not the place to address it.

ilyam8:
The list https://github.com/netdata/netdata/pull/7862#issuecomment-583453370

Probably there is a other way to add all of them - find all available vars and add missing

amoss:
> Please, give us the list.

Replying as there will not be a notification of the edit - @ilyam8 has put it in his comment above. They are the list of variables that @Ferroin that added to the script. We do not seem to be using all of them.

database/rrdhost.c
thiagoftsm: Are you sure ?

https://github.com/netdata/netdata/blob/4dc217a48a0b337e1f0baa13faa93a12c575a500/libnetdata/libnetdata.c#L171

:D

thiagoftsm:
> ```shell
> NETDATA_CPU_LOGICAL_CPU_COUNT
> ```

I verified now, all the variables without exception are in the code now, so when we merge these errors will disappear.

database/rrdhost.c
stelfrag: Ah, geez... yes I had checked that and saw there is a check -- BUT failed to see that works only if NETDATA_LOG_ALLOCATIONS is set -- thanks @thiagoftsm :smile:

PR #7864, opened 2020-01-27
Brings pattern to code

### LINKED ISSUE #7863: API info has two patterns in the code.
##### Bug report summary
After we merge the PRs that brought containers to `web/api`, it would be good to have only one pattern for the code.

##### OS / Environment
```
/etc/os-release:ID=slackware
```

##### Component Name
Web
##### Steps To Reproduce

1. Read the code

##### Expected behavior
To have the same number of elements in the response, independent of what is running on the host.

thiagoftsm:
##### Summary
Fixes #7863 

Brings back one unique pattern to the code and to response. I am considering that is better to have always the same number of variables as the response, and the previous code was not allowing us to have this.
##### Component Name
web

### DIFF
--- a/web/api/web_api_v1.c
+++ b/web/api/web_api_v1.c
@@ -815,18 +815,12 @@ inline int web_client_api_request_v1_info(RRDHOST *host, struct web_client *w, c
     buffer_sprintf(wb, "\t\"os_version_id\": \"%s\",\n", (host->system_info->host_os_version_id) ? host->system_info->host_os_version_id : "");
     buffer_sprintf(wb, "\t\"os_detection\": \"%s\",\n", (host->system_info->host_os_detection) ? host->system_info->host_os_detection : "");
 
-    if (host->system_info->container_os_name)
-        buffer_sprintf(wb, "\t\"container_os_name\": \"%s\",\n", host->system_info->container_os_name);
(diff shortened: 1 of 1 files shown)


underhood:
please update desc. this changes also that variables will be in json with empty values instead of just skippping them

thiagoftsm:
> please update desc. this changes also that variables will be in json with empty values instead of just skippping them

I updated it, thank you!

amoss:
These were intended to be optional fields - i.e. if we are not in a container then they would not be added to the response.

thiagoftsm:
> These were intended to be optional fields - i.e. if we are not in a container then they would not be added to the response.

All right, so I will close the PR. Thank you for the explanation.
@underhood I apologize for the time you expended here.

underhood:
no problem it was very small PR so it was just few minutes anyway

amoss:
> All right, so I will close the PR. Thank you for the explanation.

That's fine - I can explain in advance as well :)

There are two things here which might be slightly wrong, but are not worth spending the time to fix immediately:
1. Although these fields were intended to be optional, there was a later change that filled in default values for the GA reporting. This may have caused them to stop being optional.
2. The existing pattern in the source code violates the coding style (line width) and needs to be reformatted at some point.

PR #7866, opened 2020-01-27
Add disk size detection to system-info.sh.

### LINKED ISSUE #7731: Improve the system-info meta-data output.

Ferroin:
##### Summary

This adds really basic detection of total disk size to the system-info.sh script. This detection is _wildly_ inaccurate in many cases, but it's not realistically possible to make it much better than what's added here without making the script exponentially more complicated (there are just way too many edge cases and special conditions to worry about).

This adds the following keys to the output of the script:

* NETDATA_TOTAL_DISK_SIZE: This indicates the total detected disk size   in bytes.
* NETDATA_DISK_DETECTION: This indicates what detection method was used   for determining the total disk size. It may return either `df` (which works almost everywhere but is wildly inaccurate in a lot of common cases) or `sysfs` (which is   linux specific, but is 100% accurate for a majority of cases).

On most platforms, this parses the output of the `df` command, limiting only to filesystems that are used on fixed disks, which provides reasonably accurate results in most trvial cases, but kind of falls apart the moment people are doing anything remotely complicated in terms of storage (like using a logical volume manager or under-comitting their disk space).

On Linux, it will preferentially try to parse the info out of `/sys/block`, filtering on device major numbers that are actually used for fixed disks and excluding devices that are indicated to be removable. This provides a very accurate result if the system does not use removable media as primary storage, but requires that the user who runs the script can read the contents of `/sys/block`.

##### Component Name

area/daemon

##### Additional Information

Tested on a variety of Linux systems as well as FreeBSD. This should work on macOS as well, but I can't test it there due to lack of a usable system for testing.

Fixes #7731


stelfrag:
```
NETDATA_CONTAINER_OS_NAME=none
NETDATA_CONTAINER_OS_ID=none
NETDATA_CONTAINER_OS_ID_LIKE=none
NETDATA_CONTAINER_OS_VERSION=none
NETDATA_CONTAINER_OS_VERSION_ID=none
NETDATA_CONTAINER_OS_DETECTION=none
NETDATA_HOST_OS_NAME=Ubuntu
NETDATA_HOST_OS_ID=ubuntu
NETDATA_HOST_OS_ID_LIKE=debian
NETDATA_HOST_OS_VERSION=18.04.3 LTS (Bionic Beaver)
NETDATA_HOST_OS_VERSION_ID=18.04
NETDATA_HOST_OS_DETECTION=/etc/os-release
NETDATA_SYSTEM_KERNEL_VERSION=4.15.0-74-generic
NETDATA_SYSTEM_VIRTUALIZATION=kvm
NETDATA_SYSTEM_VIRT_DETECTION=systemd-detect-virt
NETDATA_SYSTEM_CONTAINER=unknown
NETDATA_SYSTEM_CONTAINER_DETECTION=none
NETDATA_CPU_LOGICAL_CPU_COUNT=2
NETDATA_CPU_VENDOR=GenuineIntel
NETDATA_CPU_MODEL=Intel Core Processor (Skylake, IBRS)
NETDATA_CPU_FREQ=0
NETDATA_CPU_DETECTION=lscpu
NETDATA_TOTAL_RAM=1987891200
NETDATA_RAM_DETECTION=procfs
NETDATA_TOTAL_DISK_SIZE=0
NETDATA_DISK_DETECTION=sysfs
```

Output of `df -T` in this case is
```
Filesystem     Type     1K-blocks    Used Available Use% Mounted on
udev           devtmpfs   1987804       0   1987804   0% /dev
tmpfs          tmpfs       217172    1124    216048   1% /run
/dev/vda2      ext4      18509520 6584272  11070292  38% /
tmpfs          tmpfs       970648       0    970648   0% /dev/shm
tmpfs          tmpfs         5120       0      5120   0% /run/lock
tmpfs          tmpfs       970648       0    970648   0% /sys/fs/cgroup
/dev/loop0     squashfs    125312  125312         0 100% /snap/docker/418
/dev/loop3     squashfs     17408   17408         0 100% /snap/aws-cli/151
/dev/loop2     squashfs     91264   91264         0 100% /snap/core/8268
/dev/loop4     squashfs      8832    8832         0 100% /snap/canonical-livepatch/90
/dev/loop1     squashfs     91264   91264         0 100% /snap/core/8213
/dev/loop5     squashfs    123264  123264         0 100% /snap/docker/423
tmpfs          tmpfs       194128       0    194128   0% /run/user/1000
```

Ferroin:
> ```
> NETDATA_CONTAINER_OS_NAME=none
> NETDATA_CONTAINER_OS_ID=none
> NETDATA_CONTAINER_OS_ID_LIKE=none
> NETDATA_CONTAINER_OS_VERSION=none
> NETDATA_CONTAINER_OS_VERSION_ID=none
> NETDATA_CONTAINER_OS_DETECTION=none
> NETDATA_HOST_OS_NAME=Ubuntu
> NETDATA_HOST_OS_ID=ubuntu
> NETDATA_HOST_OS_ID_LIKE=debian
> NETDATA_HOST_OS_VERSION=18.04.3 LTS (Bionic Beaver)
> NETDATA_HOST_OS_VERSION_ID=18.04
> NETDATA_HOST_OS_DETECTION=/etc/os-release
> NETDATA_SYSTEM_KERNEL_NAME=Linux
> NETDATA_SYSTEM_KERNEL_VERSION=4.15.0-74-generic
> NETDATA_SYSTEM_ARCHITECTURE=x86_64
> NETDATA_SYSTEM_VIRTUALIZATION=kvm
> NETDATA_SYSTEM_VIRT_DETECTION=systemd-detect-virt
> NETDATA_SYSTEM_CONTAINER=unknown
> NETDATA_SYSTEM_CONTAINER_DETECTION=none
> NETDATA_CPU_LOGICAL_CPU_COUNT=2
> NETDATA_CPU_VENDOR=GenuineIntel
> NETDATA_CPU_MODEL=Intel Core Processor (Skylake, IBRS)
> NETDATA_CPU_FREQ=0
> NETDATA_CPU_DETECTION=lscpu
> NETDATA_TOTAL_RAM=1987891200
> NETDATA_RAM_DETECTION=procfs
> NETDATA_TOTAL_DISK_SIZE=0
> NETDATA_DISK_DETECTION=sysfs
> ```
> 
> Output of `df -T` in this case is
> 
> ```
> Filesystem     Type     1K-blocks    Used Available Use% Mounted on
> udev           devtmpfs   1987804       0   1987804   0% /dev
> tmpfs          tmpfs       217172    1124    216048   1% /run
> /dev/vda2      ext4      18509520 6584272  11070292  38% /
> tmpfs          tmpfs       970648       0    970648   0% /dev/shm
> tmpfs          tmpfs         5120       0      5120   0% /run/lock
> tmpfs          tmpfs       970648       0    970648   0% /sys/fs/cgroup
> /dev/loop0     squashfs    125312  125312         0 100% /snap/docker/418
> /dev/loop3     squashfs     17408   17408         0 100% /snap/aws-cli/151
> /dev/loop2     squashfs     91264   91264         0 100% /snap/core/8268
> /dev/loop4     squashfs      8832    8832         0 100% /snap/canonical-livepatch/90
> /dev/loop1     squashfs     91264   91264         0 100% /snap/core/8213
> /dev/loop5     squashfs    123264  123264         0 100% /snap/docker/423
> tmpfs          tmpfs       194128       0    194128   0% /run/user/1000
> ```

Hmm, looks like VirtIO disk devices aren't being handled correctly by the logic being used for `/sys`.  This gets a bit complicated, because the device numbers for the VirtIO block device driver are dynamically assigned. I'll look into adding some logic to handle that, since we really should be using `/sys` for this if at all possible.

Ferroin:
> Hmm, looks like VirtIO disk devices aren't being handled correctly by the logic being used for /sys. This gets a bit complicated, because the device numbers for the VirtIO block device driver are dynamically assigned. I'll look into adding some logic to handle that, since we really should be using /sys for this if at all possible.

Correction, it's in the local range, and appears to have a stable block major number, I'll have a fix shortly.

Ferroin:
@stelfrag Most recent commit should fix things for you.

Ferroin:
OK, should actually work now. I need to send a patch to LKML to correct the documentation, because everything from 240-254 is listed as 'LOCAL/RESERVED' for block device majors, not dynamic allocation (which is what's obviously being used for).

daemon/system-info.sh
```diff
@@ -323,6 +318,16 @@ elif [ "${KERNEL_NAME}" = FreeBSD ] ; then
         DISK_SIZE="$((total * 1024))"
 else
         if [ -d /sys/block ] ; then
+                # List of device majors we actually count towards total disk space.
+                # The meanings of these can be found in `Documentation/admin-guide/devices.txt` in the Linux sources.
+                # The ':' surrounding each number are important for matching.
+                dev_major_whitelist=':3:8:9:21:22:28:31:33:34:44:45:47:48:49:50:51:52:53:54:55:56:57:65:66:67:68:69:70:71:72:73:74:75:76:77:78:79:88:89:90:91:93:94:96:98:101:104:105:106:107:108:109:110:111:112:114:116:128:129:130:131:132:134:135:136:137:138:139:140:141:142:143:153:160:161:179:180:202:256:257:'
+
+                if [ "${VIRTUALIZATION}" != "unknown" ] ; then
```
mfundul: I don't think it's about CPU virtualization. Most virtual block device drivers would fall in this range, such as device mapper devices, which also typically use 253. The device mapper covers a huge range of configurations.

Ferroin: We don't want DM devices though, we want the physical disks for the system. The point of looking at `/sys` is so we can match only on actual local physical devices that aren't removable storage and, which will, outside of this special case for virtualized systems, give us a 100% accurate listing of the real total disk space on the system (barring insanity from things like a host-protected area on ATA devices).

mfundul: Multipath I/O is one example, where the SCSI layer shows 8 identical devices but they are merged into 1 device-mapper multipath device, and the usable disk space is indeed 1/8th of what SCSI will show. It doesn't even have to be iSCSI devices, they can be attached via fibre channel which is effectively local storage.

thiagoftsm:
I verified that for hard disks exclusively using `sysfs` everything is going fine:

```
bash-5.0# ls -l /sys/block/
total 0
lrwxrwxrwx 1 root root 0 Jan 27 15:03 loop0 -> ../devices/virtual/block/loop0
lrwxrwxrwx 1 root root 0 Jan 27 15:03 loop1 -> ../devices/virtual/block/loop1
lrwxrwxrwx 1 root root 0 Jan 27 15:03 loop2 -> ../devices/virtual/block/loop2
lrwxrwxrwx 1 root root 0 Jan 27 15:03 loop3 -> ../devices/virtual/block/loop3
lrwxrwxrwx 1 root root 0 Jan 27 15:03 loop4 -> ../devices/virtual/block/loop4
lrwxrwxrwx 1 root root 0 Jan 27 15:03 loop5 -> ../devices/virtual/block/loop5
lrwxrwxrwx 1 root root 0 Jan 27 15:03 loop6 -> ../devices/virtual/block/loop6
lrwxrwxrwx 1 root root 0 Jan 27 15:03 loop7 -> ../devices/virtual/block/loop7
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram0 -> ../devices/virtual/block/ram0
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram1 -> ../devices/virtual/block/ram1
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram10 -> ../devices/virtual/block/ram10
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram11 -> ../devices/virtual/block/ram11
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram12 -> ../devices/virtual/block/ram12
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram13 -> ../devices/virtual/block/ram13
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram14 -> ../devices/virtual/block/ram14
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram15 -> ../devices/virtual/block/ram15
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram2 -> ../devices/virtual/block/ram2
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram3 -> ../devices/virtual/block/ram3
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram4 -> ../devices/virtual/block/ram4
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram5 -> ../devices/virtual/block/ram5
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram6 -> ../devices/virtual/block/ram6
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram7 -> ../devices/virtual/block/ram7
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram8 -> ../devices/virtual/block/ram8
lrwxrwxrwx 1 root root 0 Jan 27 15:03 ram9 -> ../devices/virtual/block/ram9
lrwxrwxrwx 1 root root 0 Jan 27 15:03 sda -> ../devices/pci0000:00/0000:00:17.0/ata1/host0/target0:0:0/0:0:0:0/block/sda
lrwxrwxrwx 1 root root 0 Jan 27 15:03 sdb -> ../devices/pci0000:00/0000:00:17.0/ata3/host2/target2:0:0/2:0:0:0/block/sdb
bash-5.0# cat /sys/block/sda/size 
1953525168
bash-5.0# cat /sys/block/sdb/size 
234455040
bash-5.0# ./daemon/system-info.sh 
NETDATA_CONTAINER_OS_NAME=none
NETDATA_CONTAINER_OS_ID=none
NETDATA_CONTAINER_OS_ID_LIKE=none
NETDATA_CONTAINER_OS_VERSION=none
NETDATA_CONTAINER_OS_VERSION_ID=none
NETDATA_CONTAINER_OS_DETECTION=none
NETDATA_HOST_OS_NAME=Slackware
NETDATA_HOST_OS_ID=slackware
NETDATA_HOST_OS_ID_LIKE=unknown
NETDATA_HOST_OS_VERSION=14.2
NETDATA_HOST_OS_VERSION_ID=14.2
NETDATA_HOST_OS_DETECTION=/etc/os-release
NETDATA_SYSTEM_KERNEL_VERSION=5.4.14
NETDATA_SYSTEM_VIRTUALIZATION=unknown
NETDATA_SYSTEM_VIRT_DETECTION=none
NETDATA_SYSTEM_CONTAINER=unknown
NETDATA_SYSTEM_CONTAINER_DETECTION=none
NETDATA_CPU_LOGICAL_CPU_COUNT=4
NETDATA_CPU_VENDOR=GenuineIntel
NETDATA_CPU_MODEL=Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
NETDATA_CPU_FREQ=2700000
NETDATA_CPU_DETECTION=lscpu sysfs
NETDATA_TOTAL_RAM=8257773568
NETDATA_RAM_DETECTION=procfs
NETDATA_TOTAL_DISK_SIZE=2187980208
NETDATA_DISK_DETECTION=sysfs
bash-5.0# bc -l
bc 1.07.1
Copyright 1991-1994, 1997, 1998, 2000, 2004, 2006, 2008, 2012-2017 Free Software Foundation, Inc.
This is free software with ABSOLUTELY NO WARRANTY.
For details type `warranty'. 
234455040 + 1953525168
2187980208
```

I will wait to others to check before I approve it.

stelfrag:
```
NETDATA_CONTAINER_OS_NAME=none
NETDATA_CONTAINER_OS_ID=none
NETDATA_CONTAINER_OS_ID_LIKE=none
NETDATA_CONTAINER_OS_VERSION=none
NETDATA_CONTAINER_OS_VERSION_ID=none
NETDATA_CONTAINER_OS_DETECTION=none
NETDATA_HOST_OS_NAME=FreeBSD
NETDATA_HOST_OS_ID=FreeBSD
NETDATA_HOST_OS_ID_LIKE=FreeBSD
NETDATA_HOST_OS_VERSION=12.1-RELEASE-p1
NETDATA_HOST_OS_VERSION_ID=unknown
NETDATA_HOST_OS_DETECTION=uname
NETDATA_SYSTEM_KERNEL_NAME=FreeBSD
NETDATA_SYSTEM_KERNEL_VERSION=1201000
NETDATA_SYSTEM_ARCHITECTURE=amd64
NETDATA_SYSTEM_VIRTUALIZATION=unknown
NETDATA_SYSTEM_VIRT_DETECTION=none
NETDATA_SYSTEM_CONTAINER=unknown
NETDATA_SYSTEM_CONTAINER_DETECTION=none
NETDATA_CPU_LOGICAL_CPU_COUNT=2
NETDATA_CPU_VENDOR=unknown
NETDATA_CPU_MODEL=Intel Core Processor (Skylake, IBRS)
NETDATA_CPU_FREQ=unknown
NETDATA_CPU_DETECTION=sysctl
NETDATA_TOTAL_RAM=2107064320
NETDATA_RAM_DETECTION=sysctl
NETDATA_TOTAL_DISK_SIZE=7890923520
NETDATA_DISK_DETECTION=df
```

df output is 

```
[stelios@freebsd ~]$ df 
Filesystem    1K-blocks    Used   Avail Capacity  Mounted on
/dev/vtbd0s1a   7705980 2830460 4259044    40%    /
devfs                 1       1       0   100%    /dev
```

This will actually report bytes

daemon/system-info.sh
Ferroin: OK, let's approach my reasoning for this from a different direction:

We quite simply cannot make this perfect, there is literally no way to do so.

What's present here does a good enough job in a vast majority of cases and provides better accuracy for the common cases than for the uncommon ones.

More specifically:

* `df` is inaccurate (significantly so in some cases) in a number of very common configurations because it only looks at mounted devices. It reports low on a vast majority of 'trivial' storage configurations by anywhere from a few MB to double-digit percentages of total disk space, and high by arbitrary numbers in a number of non-trivial but reasonably common configurations, and that's all with it explicitly ignoring network and in-memory filesystems. It also reports inaccurately by some measures when dealing with network disks (it's inconsistent whether any arbitrary person would consider an iSCSI/NBD/ATAoE device to be a part of the system or not, and often depends on the exact circumstances).
* Checking `/sys` is extremely (read 'perfectly') accurate in a significant percentage of cases that `df` gives bogus results, because it looks at block devices directly. As it's implemented here, it correctly handles almost all standard cases except multipath, stuff that's thinly-provisioned at the hardware level, and VirtIO block devices (notably, it does handle VirtIO SCSI devices correctly, and that's what most people _should_ be using with QEMU/KVM), as well as covering a number of less common cases that `df` gets wrong (like thinly-provisioned logical volumes).

I would be more than happy to improve upon this logic, but there are two caveats:

* We need this functionality now for the ACLK stuff. While this is not perfect, it's good enough for the testing and pre-release stuff we're currently doing there, and there's no reason it can't be further improved later, once it's not potentially blocking development.
* Complexity of this code will increase _very_ quickly as we handle more and more specific use cases. Just off the top of my head, we're looking at at least a dozen more lines _just_ to properly handle the multipath case because we need to filter on DM table type (we _can't_ include all DM devices there, we'll get ridiculously high estimates if we do (potentially more than twice the actual disk space, and that's if they're not using thin provisioning, caching, or other complicated stacked layouts)). I'm arguably not even handling the VirtIO block case 'correctly' here because I'm not excluding other device types from the list.

If so desired, I can redo the sysfs handling to be far more robust, but that's probably going to take multiple days if we want to account for proper testing.

Ferroin:
> ```
> NETDATA_CONTAINER_OS_NAME=none
> NETDATA_CONTAINER_OS_ID=none
> NETDATA_CONTAINER_OS_ID_LIKE=none
> NETDATA_CONTAINER_OS_VERSION=none
> NETDATA_CONTAINER_OS_VERSION_ID=none
> NETDATA_CONTAINER_OS_DETECTION=none
> NETDATA_HOST_OS_NAME=FreeBSD
> NETDATA_HOST_OS_ID=FreeBSD
> NETDATA_HOST_OS_ID_LIKE=FreeBSD
> NETDATA_HOST_OS_VERSION=12.1-RELEASE-p1
> NETDATA_HOST_OS_VERSION_ID=unknown
> NETDATA_HOST_OS_DETECTION=uname
> NETDATA_SYSTEM_KERNEL_NAME=FreeBSD
> NETDATA_SYSTEM_KERNEL_VERSION=1201000
> NETDATA_SYSTEM_ARCHITECTURE=amd64
> NETDATA_SYSTEM_VIRTUALIZATION=unknown
> NETDATA_SYSTEM_VIRT_DETECTION=none
> NETDATA_SYSTEM_CONTAINER=unknown
> NETDATA_SYSTEM_CONTAINER_DETECTION=none
> NETDATA_CPU_LOGICAL_CPU_COUNT=2
> NETDATA_CPU_VENDOR=unknown
> NETDATA_CPU_MODEL=Intel Core Processor (Skylake, IBRS)
> NETDATA_CPU_FREQ=unknown
> NETDATA_CPU_DETECTION=sysctl
> NETDATA_TOTAL_RAM=2107064320
> NETDATA_RAM_DETECTION=sysctl
> NETDATA_TOTAL_DISK_SIZE=7890923520
> NETDATA_DISK_DETECTION=df
> ```
> 
> df output is
> 
> ```
> [stelios@freebsd ~]$ df 
> Filesystem    1K-blocks    Used   Avail Capacity  Mounted on
> /dev/vtbd0s1a   7705980 2830460 4259044    40%    /
> devfs                 1       1       0   100%    /dev
> ```
> 
> This will actually report bytes

That's intentional. Just like for all the other numeric values, the script intentionally reports in base units (in this case bytes) so that scaling can be done as appropriate by whatever other higher-level would be presenting this data.

In this case it's just forcing `df` to report in a specific unit size (in this case 1024-byte blocks) and then multiplying the result by the unit scaling factor (in this case 1024) to get the returned value. This is nessecary because other than GNU `df`, most implementations do not have a switch that causes them to report values directly as bytes. The net error here is usually going to be less than -0.1%, and often less than -0.001%.

daemon/system-info.sh
```diff
@@ -286,6 +286,70 @@ elif [ -r /proc/meminfo ] ; then
...
+        if (lsvfs | grep -q apfs) ; then
+            types="${types},apfs"
+        fi
+
+        if (lsvfs | grep -q ufs) ; then
+            types="${types},ufs"
+        fi
+
+        DISK_DETECTION="df"
+        total="$(df -t ${types} -k | tail -n +1 | awk '{s+=$2} END {print s}')"
```
prologic: At least on my macOS systems; this fails with:

```sh
$ df -t hfs,ufs
df: no file systems processed
```

prologic: I _suspect_ it _might_ have to do with the fact that I'm using GNU coreutils:

```sh
$ which df
/usr/local/opt/coreutils/libexec/gnubin/df
```

We _should_ account for this too as rare as this _may_ be most devs / sysadmins / etc would setup their macOS environments with Brew and override some of the default BSD utilities.

prologic: But then again the `/bin/df` utility on macOS that comes installed with macOS / Darwin fails differently again:

```sh
$ /bin/df -t hfs,ufs -k | tail -n +1 | awk '{s+=$2} END {print s}'
df: -k: No such file or directory
```

prologic: In fact are we sure about the usage the `-k` option?

See for example (_with GNU coreutils_):

```sh
prologic@Jamess-iMac
Tue Jan 28 13:27:20
~
 0 0 0
$ df -t nfs -k | tail -n +1 | awk '{s+=$2} END {print s}'
df: no file systems processed


prologic@Jamess-iMac
Tue Jan 28 13:27:22
~
 1 0 0
$ df -t ufs -k | tail -n +1 | awk '{s+=$2} END {print s}'
df: no file systems processed


prologic@Jamess-iMac
Tue Jan 28 13:27:26
~
 1 0 0
$ df -t apfs -k | tail -n +1 | awk '{s+=$2} END {print s}'
489620265

prologic@Jamess-iMac
Tue Jan 28 13:27:30
~
 0 0 0
$ df -t apfs,ufs,nfs -k | tail -n +1 | awk '{s+=$2} END {print s}'
df: no file systems processed
```

Ferroin: It's the `-t` option in the case of GNU coreutils that's causing problems.

Personally, I think the correct approach here is to just always use `/bin/df` on macOS and figure out why it's having issues with `-k`.

I suspect this is a case of it being _super_ picky about option ordering (all the online copies of manpages for the macOS `df` list `-t` last in the options, which I had not noticed before).

amoss:
There is some duplicate counting on MacOS - do we have any way around this?
```
NETDATA_CONTAINER_OS_NAME=none
NETDATA_CONTAINER_OS_ID=none
NETDATA_CONTAINER_OS_ID_LIKE=none
NETDATA_CONTAINER_OS_VERSION=none
NETDATA_CONTAINER_OS_VERSION_ID=none
NETDATA_CONTAINER_OS_DETECTION=none
NETDATA_HOST_OS_NAME=mac
NETDATA_HOST_OS_ID=Mac OS X
NETDATA_HOST_OS_ID_LIKE=mac
NETDATA_HOST_OS_VERSION=10.15.1
NETDATA_HOST_OS_VERSION_ID=unknown
NETDATA_HOST_OS_DETECTION=sw_vers
NETDATA_SYSTEM_KERNEL_NAME=Darwin
NETDATA_SYSTEM_KERNEL_VERSION=19.0.0
NETDATA_SYSTEM_VIRTUALIZATION=unknown
NETDATA_SYSTEM_VIRT_DETECTION=none
NETDATA_SYSTEM_CONTAINER=unknown
NETDATA_SYSTEM_CONTAINER_DETECTION=none
NETDATA_CPU_LOGICAL_CPU_COUNT=unknown
NETDATA_CPU_VENDOR=unknown
NETDATA_CPU_MODEL=unknown
NETDATA_CPU_FREQ=unknown
NETDATA_CPU_DETECTION=none
NETDATA_TOTAL_RAM=unknown
NETDATA_RAM_DETECTION=none
NETDATA_TOTAL_DISK_SIZE=1999853748224
NETDATA_DISK_DETECTION=df

┌[333|28Jan1507|6|0]-[amoss@wintermute]-[/Users/amoss/netdata/localfork.sysinfo]
└>df -h
Filesystem      Size   Used  Avail Capacity iused      ifree %iused  Mounted on
/dev/disk1s1   466Gi   10Gi  394Gi     3%  483409 4881969471    0%   /
devfs          194Ki  194Ki    0Bi   100%     678          0  100%   /dev
/dev/disk1s2   466Gi   57Gi  394Gi    13%  437729 4882015151    0%   /System/Volumes/Data
/dev/disk1s5   466Gi  4.0Gi  394Gi     2%       4 4882452876    0%   /private/var/vm
map auto_home    0Bi    0Bi    0Bi   100%       0          0  100%   /System/Volumes/Data/home
/dev/disk1s4   466Gi  507Mi  394Gi     1%      17 4882452863    0%   /Volumes/Recovery
```

amoss:
Went back to check debian and it seems that it has skipped the drive that `/home` is mounted from?
```
NETDATA_CONTAINER_OS_NAME=none
NETDATA_CONTAINER_OS_ID=none
NETDATA_CONTAINER_OS_ID_LIKE=none
NETDATA_CONTAINER_OS_VERSION=none
NETDATA_CONTAINER_OS_VERSION_ID=none
NETDATA_CONTAINER_OS_DETECTION=none
NETDATA_HOST_OS_NAME=Debian GNU/Linux
NETDATA_HOST_OS_ID=debian
NETDATA_HOST_OS_ID_LIKE=unknown
NETDATA_HOST_OS_VERSION=10 (buster)
NETDATA_HOST_OS_VERSION_ID=10
NETDATA_HOST_OS_DETECTION=/etc/os-release
NETDATA_SYSTEM_KERNEL_VERSION=4.19.0-6-amd64
NETDATA_SYSTEM_VIRTUALIZATION=none
NETDATA_SYSTEM_VIRT_DETECTION=systemd-detect-virt
NETDATA_SYSTEM_CONTAINER=unknown
NETDATA_SYSTEM_CONTAINER_DETECTION=none
NETDATA_CPU_LOGICAL_CPU_COUNT=8
NETDATA_CPU_VENDOR=GenuineIntel
NETDATA_CPU_MODEL=Intel(R) Core(TM) i7-4790S CPU @ 3.20GHz
NETDATA_CPU_FREQ=4000000000
NETDATA_CPU_DETECTION=lscpu
NETDATA_TOTAL_RAM=33676611584
NETDATA_RAM_DETECTION=procfs
NETDATA_TOTAL_DISK_SIZE=2422387296
NETDATA_DISK_DETECTION=sysfs

┌[4039|28Jan1512|6|0]-[amoss@loki]-[/home/amoss/netdata/fork.reviews]
└>df -h
Filesystem                Size  Used Avail Use% Mounted on
udev                       16G     0   16G   0% /dev
tmpfs                     3.2G  347M  2.8G  11% /run
/dev/sdb2                  28G   20G  6.2G  77% /
tmpfs                      16G  717M   15G   5% /dev/shm
tmpfs                     5.0M  4.0K  5.0M   1% /run/lock
tmpfs                      16G     0   16G   0% /sys/fs/cgroup
/dev/sdb1                 511M  5.1M  506M   1% /boot/efi
/dev/sdb5                 829G   13G  774G   2% /home
tmpfs                     3.2G   20K  3.2G   1% /run/user/118
tmpfs                     3.2G  116K  3.2G   1% /run/user/1000
\\192.168.1.4\Alexandria  5.5T  4.9T  590G  90% /nas

```

Ferroin:
> There is some duplicate counting on MacOS - do we have any way around this?

Unfortunately not really. We're entirely dependent on macOS on the `df` output, and handling duplicates when dealing with APFS isn't exactly easy. We have the same issue when using `df` on Linux when BTRFS is being used and there are multiple subvolumes mounted from the same volume.

> Went back to check debian and it seems that it has skipped the drive that `/home` is mounted from?

Is `/dev/sdb` marked as removable media on that system (value of `/sys/block/sdb/removable`)? If so, that would be why.

Ferroin:
Quick clarification for everyone, the current code is known to report inaccurately in the following cases:

* APFS on macOS: Counts volumes once for each place they are mounted. There is no 'correct' fix for this AFAICT, but a close approximation is not too hard (it requires a lot more filtering on the `df` output though). Estimated to be equivalent to 5-6 LoC extra to handle this case, but the result will introduce a chance of reporting _low_ values in some cases.
* BTRFS on Linux when using `df` for disk detection: Counts volumes once for each mounted subvolume. This does have a correct fix (use `btrfs subvolume list` to construct a list of subvolume relationships and filter based on that)) that should get us near perfect accuracy in most cases, but it requires EUID=0 and/or special setup from the user and is rather complicated. Estimated at approximately 20 LoC to fix.
* Multipath on Linux when using `sysfs` for disk detection: Counts each drive once for each path it's accessible through. This has a potential correct fix (look at DM tables to determine what DM targets are multipath volumes, count those, and then exclude their components from the computation), but it's rather complicated. Estimated to need more than 20 LoC to fix.
* LVM on Linux when using `df` for disk detection: All kinds of complicated cases here where the actual disk space may be higher or much lower than what's reported through `df`. The trivial cases to explain are thinly provisioned LV's (which will result in much higher reported size than actual) and cached LV's (which will result in a lower reported size than actual). The fix here is to ignore `df` and use data from `/sys/block`, which is the entire reason we have detection through `/sys/block` on Linux.
* Non-trivial GEOM setups on FreeBSD: Same issues as for LVM on Linux, but there's not any good solution since parsing the output of `geom` is complicated (it's designed to be consumed by humans, not computers). There may be some alternative I'm missing though.
* Thinly-provisioned storage hardware: Same issues as for thinly-provisioned logical volumes. Thankfully, this stuff is pretty rare, because there is literally no way for us to detect and account for it in most cases.
* Hardware RAID: We'll report the RAID volume sizes, not the total disk size. We can work around this in some cases on Linux by using `/sys/block`, but not all, and there's no good way to mitigate it on any other platform.
* VM's with LVM on Linux when using `sysfs` for disk detection: Because of needed handling for the VirtIO Block driver, this will usually report high values. This is fixable by looking specifically for VirtIO block device major number in `/proc/devices`. Estimated 6 LoC to fix, but this should be a non-issue for most people except developers, as the BCP for production deployment of QEMU/KVM VM's is to use VirtIO SCSI instead of VirtIO Block for the storage stack.
* Linux systems where primary storage is removable media using `sysfs` for disk detection: Effectively, we'll ignore the removable media. This behavior is intentional so that we report fixed storage capacity for the system and ignore things like flash drives that happen to be plugged in. There's not really any good way to fix this honestly, at least not without counting transiently connected storage devices towards the total disk capacity for the system.

----

As a secondary note, I do intend to fix all of these issues that I can, but I'm of the opinion that those fixes should be done as secondary PR's since they will require more complicated testing to verify and @thiagoftsm is waiting on this to be merged so that he can get the stuff he's doing done.

thiagoftsm:
Hi @Ferroin ,

Please, do not worry! Take the necessary time to  briing everything necessary. I will do an unique request, keep the output variables with the same name, when this PR is merged, I will change my PR.

Thank you!

amoss:
> Is `/dev/sdb` marked as removable media on that system (value of `/sys/block/sdb/removable`)? If so, that would be why.

I think it is not as the value is 0?

Ferroin:
> > Is `/dev/sdb` marked as removable media on that system (value of `/sys/block/sdb/removable`)? If so, that would be why.
> 
> I think it is not as the value is 0?

OK, that should be getting counted then. What are the contents of `/sys/block/sdb/dev` and `/sys/block/sdb/size` on that system? Also, is there a `/dev/sda` that isn't getting used for anything?

daemon/system-info.sh
prologic: Looks like you fixed it (_I can't follow the comment trail anymore on this PR_)

```sh
NETDATA_TOTAL_DISK_SIZE=752057774080
NETDATA_DISK_DETECTION=df
```

amoss:
> As a secondary note, I do intend to fix all of these issues that I can, but I'm of the opinion that those fixes should be done as secondary PR's since they will require more complicated testing to verify and @thiagoftsm is waiting on this to be merged so that he can get the stuff he's doing done.

This is entirely correct and the end of the sprint is upon us. LGTM - the individual problematic cases can be handled in a follow-up issue. These should be independent from @thiagoftsm's work on the API, although please be aware that he will change the script variable names to make them consistent by adding `SYSTEM_`

Ferroin:
Merging as requested. Another PR improving the logic here to better handle some of the cases mentioned in https://github.com/netdata/netdata/pull/7866#issuecomment-579293389 will be opened in the near future.

Issue #7869, opened 2020-01-27
Fails to build with `-fno-common` / GCC 10


candrews:
##### Bug report summary
gcc-10 and above flipped a default from -fcommon to -fno-common: https://gcc.gnu.org/PR85678

netdata currently fails to compile when given the CFLAG `-fno-common`:
```
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libnetdata/libnetdata.o (symbol from plugin): in function `strdupz':
(.text+0x0): multiple definition of `silencers'; collectors/apps.plugin/apps_plugin.o (symbol from plugin):(.text+0x0): first defined here
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libnetdata/locks/locks.o (symbol from plugin): in function `netdata_thread_disable_cancelability':
(.text+0x0): multiple definition of `silencers'; collectors/apps.plugin/apps_plugin.o (symbol from plugin):(.text+0x0): first defined here
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libnetdata/log/log.o (symbol from plugin): in function `facility_log':
(.text+0x0): multiple definition of `silencers'; collectors/apps.plugin/apps_plugin.o (symbol from plugin):(.text+0x0): first defined here
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libnetdata/popen/popen.o (symbol from plugin): in function `myp_init':
(.text+0x0): multiple definition of `silencers'; collectors/apps.plugin/apps_plugin.o (symbol from plugin):(.text+0x0): first defined here
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libnetdata/procfile/procfile.o (symbol from plugin): in function `procfile_adaptive_initial_allocation':
(.text+0x0): multiple definition of `silencers'; collectors/apps.plugin/apps_plugin.o (symbol from plugin):(.text+0x0): first defined here
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libnetdata/os.o (symbol from plugin): in function `get_system_cpus':
[...]

PR #7870, opened 2020-01-27
.travis.yml: Add -fno-common to CFLAGS

candrews:
-fno-common will be set by default in GCC 10, see https://gcc.gnu.org/PR85678

Therefore, netdata should test with that configuration to be ready for this change.

See https://github.com/netdata/netdata/issues/7869

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,7 +121,7 @@ jobs:
 
       name: Standard netdata build
       script: fakeroot ./netdata-installer.sh --install $HOME --dont-wait --dont-start-it --enable-plugin-nfacct --enable-plugin-freeipmi --disable-lto
-      env: CFLAGS='-O1 -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1'
+      env: CFLAGS='-O1 -Wall -Wextra -Wformat-signedness -fstack-protector-all -fno-common   -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1'
       after_failure: post_message "TRAVIS_MESSAGE" "<!here> standard netdata build is failing (Still dont know which one, will improve soon)"
 
     - name: Docker container build process (alpine installation)


candrews:
I completely agree - the changes to fix this issue should be made and this PR merged after that to ensure no such future issues appear.

amoss:
It seems that the underlying problem is that some global variables are defined in header files that are included into multiple translation units. The cases that I see that cause errors with `-fno-common` are:
```
./libnetdata/health/health.h:47:SILENCERS *silencers;
./web/api/web_api_v1.h:32:char *api_secret;
./daemon/common.h:87:int netdata_ready;
./health/health.h:55:char *silencers_filename;
```

None of these variables should be declared in header-files. I will create an issue to move them into the appropriate place and mark this issue as blocked until it is completed.

thiagoftsm:
@amoss , please, assing the issue to me and I will create a PR today yet for this. Thanks!

mfundul:
> 
> 
> It seems that the underlying problem is that some global variables are defined in header files that are included into multiple translation units. The cases that I see that cause errors with `-fno-common` are:
> 
> ```
> ./libnetdata/health/health.h:47:SILENCERS *silencers;
> ./web/api/web_api_v1.h:32:char *api_secret;
> ./daemon/common.h:87:int netdata_ready;
> ./health/health.h:55:char *silencers_filename;
> ```
> 
> None of these variables should be declared in header-files. I will create an issue to move them into the appropriate place and mark this issue as blocked until it is completed.

They need not be moved, the only thing needed is an `extern` at the beginning of each offending line.

amoss:
> They need not be moved, the only thing needed is an `extern` at the beginning of each offending line.

The variables are not declared in any .c translation units, only in the header files. Adding `extern` to the header will result in them not being defined anywhere.

amoss:
> The variables are not declared in any .c translation units, only in the header files. Adding `extern` to the header will result in them not being defined anywhere.

I had to go back and check this, @mfundul is right. @thiagoftsm please fix it the way that @mfundul suggests as it will be quicker and cleaner.

amoss:
> I had to go back and check this, @mfundul is right. @thiagoftsm please fix it the way that @mfundul suggests as it will be quicker and cleaner.

After a longer slack conversation we are back on the same page, let me summerize:

* Currently those variables have no definitions in `.c` files, only declarations in `.h` files.
* Adding an `extern` to the header file will work on gcc.
* It is not guaranteed to work on every compiler/system.
* Best practice is to put `extern` before the declaration in the header, and use a single definition in a .c file.

Please follow best-practice and do both:
* Keep the existing declarations and add `extern`.
* Choose the most appropriate translation unit (`.c` file) for each variable and add a single definition there.

thiagoftsm:
#7877 brings the fixes for this.

amoss:
@prologic - this should be ready to go, please ping me if you need anything else.

candrews:
> this should be ready to go, please ping me if you need anything else.

Indeed, all tests are now passing :tada:

prologic:
Thanks @amoss @thiagoftsm and @candrews for the PR :)

Issue #7826, opened 2020-01-22
Native Binary Packages


prologic:
##### Feature idea summary

This is part of a larger body of work to support and actively maintain native binary packages for all supported OS / Distributions we support today starting with Debian, Ubuntu, Fedora, CentOS, RHEL and openSUSE on the amd64 / x86_64 architecture and move the default installation method to "install from Native Binary Packages".

We have limited support for native binary packages for _some_ Distributions but not all. Our `kickstart.sh` script also continues to build form source. Part of the bigger problem is the way this work was originally done using LXC which turned out to be more work/effort and troublesome that it was worth.

The new plan is to mirror what we have today using Docker-based build and test flows and continue to publish to [packagecloud](https://packagecloud.io/netdata/netdata)

Existing work already done include:
- Docker-based builder images for the above distributions in [netdata/helper-images](https://github.com/netdata/helper-images)
- `Dockerfile` for a full build/install/test cycle #7735

##### Expected behavior

The default installation method for most systems is "Native Binary Packages" with a fallback to "Build from Source".

mickae1:
hi,I would like to know if there is a script to generate a deb file ?

prologic:
There is, but have you checked out [repo](https://packagecloud.io/netdata/netdata)?

mickae1:
> There is, but have you checked out [repo](https://packagecloud.io/netdata/netdata)?

I tried, but after adding the repo to my source list, 

Apt-get update 
Apt-get install netdata
Answer : no package exist with this name ..

Where is the script to create .deb package?

prologic:
The regular Debian control files are on `contrib/debian` you can build the DEB packages there using the normal tooling. Looks like you're trying to get Netdata on an ARM system with Debian packages? (_which AFAIK we don't support at this time_).

mickae1:
> The regular Debian control files are on `contrib/debian` you can build the DEB packages there using the normal tooling. Looks like you're trying to get Netdata on an ARM system with Debian packages? (_which AFAIK we don't support at this time_).

can you explain me how to proceed ?

prologic:
@mickae1 Are you able to file a separate bug report for your inability to use our prebuilt Debian packages we provide? Those _should_ work.

mickae1:
@prologic  it's done #10076

Ferroin:
All parts finished.

Issue #7825, opened 2020-01-22
Migrate kickstart.sh to use Native Binary Packages by default


prologic:

Ferroin:
Done with the release of the new kickstart script.

PR #7890, opened 2020-01-29
Adds docs for how to build/install NetData on CentOS 8.x

### LINKED ISSUE #6967: Kickstart.sh doesn't work for CentOS 8 & RHEL 8
##### Bug report summary
Running the `kickstart.sh` on CentOS 8 fails on 'Installation of required packages'.

```
/etc/os-release information:
NAME            : CentOS Linux
VERSION         : 8 (Core)
ID              : centos
ID_LIKE         : rhel fedora
VERSION_ID      : 8

We detected these:
Distribution    : centos
Version         : 8
Codename        : 8 (Core)
Package Manager : install_yum
Packages Tree   : centos
Detection Method: /etc/os-release
Default Python v: 2

This system is receiving updates from Red Hat Satellite or Spacewalk server.
Last metadata expiration check: 2:42:06 ago on Fri Sep 27 14:58:16 2019.
No match for argument: Judy-devel
No match for argument: MySQL-python
No match for argument: autoconf-archive
No match for argument: autogen
No match for argument: libmnl-devel
No match for argument: libuv-devel
No match for argument: python
There are following alternatives for "python": python2, python36
No match for argument: python-psycopg2



We are very sorry!

Installation of required packages failed.

What to do now:

1. Make sure your system is updated.
Most of the times, updating your system will resolve the issue.


2. If the error message is about a specific package, try removing
that package from the command and run it again.
Depending on the broken package, you may be able to continue.


3. Let us know. We may be able to help.
Open a github issue with the above log, at:


https://github.com/firehol/netdata/issues                                                                                             
```
##### OS / Environment
CentOS 8
1.10.0 (I am limited to this version due to company policies. Version should be irrelevant to this issue) 
##### Component Name
Installation scripts
##### Steps To Reproduce
`bash kickstart.sh all --non-interactive`
##### Expected behavior
Installation should succeed.

prologic:
##### Summary

Partially fixes #6967

##### Component Name

area/docs

##### Additional Information

### DIFF
--- a/packaging/installer/methods/manual.md
+++ b/packaging/installer/methods/manual.md
@@ -117,7 +117,7 @@ Netdata DB engine can be enabled when these are installed (they are optional):
 
 ### CentOS / RHEL 8.x
 
-For CentOS / RHEL 8.x a lot of development packages have moved out into thwi
+For CentOS / RHEL 8.x a lot of development packages have moved out into their
 own separate repositories. Some other dependeicies are either missing completely
 or have to be sourced by 3rd-parties.
 
--- a/packaging/installer/methods/manual.md
+++ b/packaging/installer/methods/manual.md
@@ -31,6 +31,7 @@ should be installed on your system to build and run Netdata. It supports most ma
         [EPEL](http://www.tecmint.com/how-to-enable-epel-repository-for-rhel-centos-6-5/).
         In addition, RHEL/CentOS version 6 also need
         [OKay](https://okay.com.mx/blog-news/rpm-repositories-for-centos-6-and-7.html) for package libuv version 1.
+    -   CentOS 8 / RHEL 8 requires a bit of extra work. See the dedicated section below.
 
 -   **SuSe** Linux and its derivatives (including **openSuSe**)
 
@@ -113,6 +114,40 @@ Netdata DB engine can be enabled when these are installed (they are optional):
 | `openssl`| Cryptography and SSL/TLS toolkit|
 
 *Netdata will greatly benefit if you have the above packages installed, but it will still work without them.*
+
+### CentOS / RHEL 8.x
+
+For CentOS / RHEL 8.x a lot of development packages have moved out into thwi
+own separate repositories. Some other dependeicies are either missing completely
+or have to be sourced by 3rd-parties.
+
+CentOS 8.x:
+
+- Enable the PowerTools repo
+- Enable the EPEL repo
+- Enable the Extra repo from [extra.getpagespeed.com](https://extras.getpagespeed.com/release-el8-latest.rpm)
+
+And install the minimum required dependencies:
+
+```sh
+# Enable config-manager
+yum install -y 'dnf-command(config-manager)'
+
+# Enable PowerTools
+yum config-manager --set-enabled PowerTools
+
+# Enable EPEL
+yum install -y epel-release
+
+# Install Repo for libuv-devl (NEW)
+yum install -y https://extras.getpagespeed.com/release-el8-latest.rpm
+
+# Install Devel Packages
+yum install autoconf automake curl gcc git libmnl-devel libuuid-devel openssl-devel libuv-devel lz4-devel make nc pkgconfig python3 zlib-devel
+
+# Install Judy-Devel directly
+yum install -y http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm
+```
 
 ---
 


prologic:
Hmmm I don't like it. I just don't think the way this is laid out in the Preview is all that good now. cc @joelhans if you have any good idea for how to structure "special" documentation for CentOS / RHEL 8.x

joelhans:
@prologic How often is this kind of special case going to be required, for any OS we support? Will this affect CentOS / RHEL for _every_ version moving forward?

I thought about creating separate pages for every OS we support, which would list our recommended installation methods, and could contain information like this. However, I worry about maintaining these pages, especially since we sometimes don't recommend the same installation method even between two versions of the same distro.

Ferroin:
> How often is this kind of special case going to be required, for any OS we support? Will this affect CentOS / RHEL for _every_ version moving forward?

This is likely to be the only special case, but it's almost certainly going to affect every future version of RHEL, and possibly all future versions of CentOS (it remains to be seen how blindly the CentOS devs want to follow RHEL's rather developer-hostile choice on this).

prologic:
> @prologic How often is this kind of special case going to be required, for any OS we support? Will this affect CentOS / RHEL for _every_ version moving forward?

Until we have Native Binary Package #7826 and Migrate the default `kickstart.sh` behaviour to use said packages #7825 we're stuck with having to provide a) good documentation and b) handling these special cases.

I don't expect this sort of special case to occur very frequently; no. But....
 
> I thought about creating separate pages for every OS we support, which would list our recommended installation methods, and could contain information like this. However, I worry about maintaining these pages, especially since we sometimes don't recommend the same installation method even between two versions of the same distro.

I do feel that we _should_ do this :) -- I'd personally love it if we could restructure the "installation :: Other Methods" section of our docs with nice logos for our OS / Distro(s) and dedicated sections for each where that makes sense.

Even better if we can "import" certain pieces of docs that are common in some cases that would be better maintainability wise.

This is the only PR that kind of ruins this section a bit though so I'm not to worried.

prologic:
> (it remains to be seen how blindly the CentOS devs want to follow RHEL's rather developer-hostile choice on this).

Actually... CentOS is just a repacked RHEL -- So don't really have a choice in the matter really. They did have to rename the repo to "PowerTools" however since RHEL has "Code Builder" trademarked or something I'm told. Anyway you can find the discussion on the #centos and #rhel IRC channels on FreeNode :)

rodeengel:
I'm not sure if this is the right spot for this but the CentOS8 steps did not work for me. I was able to get the following to work:

# Enable config-manager
yum install -y 'dnf-command(config-manager)'

# Enable PowerTools
yum config-manager --set-enabled PowerTools

# Enable EPEL
yum install -y epel-release

# Install Repo for libuv-devl (NEW)
yum config-manager --disable epel
yum install -y https://extras.getpagespeed.com/release-el8-latest.rpm
yum config-manager --enable epel

# Install Devel Packages
dnf install http://downloads.redhat.com/redhat/rhel/rhel-8-beta/appstream/x86_64/Packages/libuv-devel-1.23.1-1.el8.x86_64.rpm
dnf install openssl-devel
yum install autoconf automake curl gcc git libmnl-devel libuuid-devel lz4-devel make nc pkgconfig python3 zlib-devel

# Install Judy-Devel directly
yum install -y http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm

prologic:
> I'm not sure if this is the right spot for this but the CentOS8 steps did not work for me. I was able to get the following to work:

Thanks for spotting this! This is perfectly the right place :D Early feedback is always appreciated as I can't possibly get everything right on the first go!

What didn't work for you specifically? (_I fixed the fomratting of your comment; hope that's okay!_)

packaging/installer/methods/manual.md
```diff
@@ -114,6 +115,40 @@ Netdata DB engine can be enabled when these are installed (they are optional):
 
 *Netdata will greatly benefit if you have the above packages installed, but it will still work without them.*
 
+### CentOS / RHEL 8.x
+
+For CentOS / RHEL 8.x a lot of development packages have moved out into thwi
```
ncmans:
```suggestion
For CentOS / RHEL 8.x a lot of development packages have moved out into their
```

rodeengel:
Thank you for the response!

I was having an issue getting both libuv-devel and openssl-devel to install with the rest of the yum install line for devel packages.

prologic:
> Thank you for the response!
> 
> I was having an issue getting both libuv-devel and openssl-devel to install with the rest of the yum install line for devel packages.

Are you able to share what problems you were having? Note that the PR #7918 adds support for CentOS 8.x in our `kickstart.sh` and `install-required-packages.sh` which kickstart calls out to. You are welcome to try that script locally with a Git checkout or see if what you were doing is/was different from that that PR is doing?

Curious to know if/what problems you were having in case I missed something :)

joelhans:
> I do feel that we _should_ do this :) -- I'd personally love it if we could restructure the "installation :: Other Methods" section of our docs with nice logos for our OS / Distro(s) and dedicated sections for each where that makes sense.

Let's talk about this more then! If you all think it's maintainable, then we should go ahead and do it. We're also getting a FE dev to help with marketing efforts over the next few weeks, and one of the projects will be some kind of installation "wizard," where the user selects their OS+version and we give them the recommended installation method. These two projects might go hand-in-hand.

prologic:
> Let's talk about this more then! If you all think it's maintainable, then we should go ahead and do it. We're also getting a FE dev to help with marketing efforts over the next few weeks, and one of the projects will be some kind of installation "wizard," where the user selects their OS+version and we give them the recommended installation method. These two projects might go hand-in-hand.

Setup some time on a calendar if you wish and we can video chat about this :) Or Slack or whatever :)

rodeengel:
> 
> 
> > Thank you for the response!
> > I was having an issue getting both libuv-devel and openssl-devel to install with the rest of the yum install line for devel packages.
> 
> Are you able to share what problems you were having? Note that the PR #7918 adds support for CentOS 8.x in our `kickstart.sh` and `install-required-packages.sh` which kickstart calls out to. You are welcome to try that script locally with a Git checkout or see if what you were doing is/was different from that that PR is doing?
> 
> Curious to know if/what problems you were having in case I missed something :)

Thanks for the followup Prologic. I did a fresh install this morning using the steps listed on:
https://github.com/netdata/netdata/blob/64f3b696aa72eed41831d738c4accaffd175767a/packaging/installer/methods/manual.md 

These worked for me and the install is up and running. I'm not sure why they were not working the other day.

PR #7912, opened 2020-01-29
Add doc with post-install instructions for GCP

### LINKED ISSUE #7854: Update documentation with post-install instructions for GCP users
Given #7786 and other feedback from users on GCP, we need to document how users can quickly fix some common configuration issues, particularly around firewalls that prevent users from accessing the Netdata dashboard.

This could also be extended with details about other platforms, but for the sake of this issue we'll document GCP and I'll research if we have documented solutions for other cloud providers in the issues already.

I'm thinking right now that this work is blocked by #7841, which will add a folder at `packaging/installer/methods/`. Once merged, I could put a `cloud-providers.md` file into that folder with these instructions, and link to this page from the main installation page as well.

I also need some insight into the GCP interface to help direct users to where they need to go to add a firewall rule. @pree-t1, do you think you could help with that? A few screenshots would be enough, or just written instructions: Click on **Settings**, then **Firewall rules**, etc... If not, someone around here should have a GCP account...

Working firewall rule for GCP:

```
Name: <name>
Type: Ingress
Targets: <name-tag>
Filters: 0.0.0.0/0
Protocols/ports: 19999
Action: allow
Priority: 1000

Also, add the above target tag to the instance's network tags. I was able to get it working with these changes.
```

@prologic FYI!

joelhans:
##### Summary

Fixes #7854

I added a new page at `packaging/installer/methods/cloud-providers.md` with some basic installation and post-installation instructions. For now, I've only included the firewall issue, and the solution for GCP users, as I couldn't find anything in our issues that offers definitive solutions. Also, I don't have accounts for these platforms, so I can't test them all out myself. But we can definitely expand this document with other common issues + their fixes as users report them.

##### Component Name

docs


joelhans:
> Preview LGTM. We should add appropriate sections for AWS and Azure too.
> 
> I _think_ this is good enough once we add the other two, any other more complex cloud provider configurations to do with security groups, virtual private networks, etc _should_ probably be beyond the scope of our docs.

Managed to create some free accounts on Azure/AWS to test things out, added those instructions.

Ferroin:
This is possibly just me, but I feel we should add links to the docs on running Netdata proxied through a web server here. Realistically, that's a much more secure and reliable option than directly exposing Netdata to the internet and may not even need any changes in the VPC configuration to work.

joelhans:
> This is possibly just me, but I feel we should add links to the docs on running Netdata proxied through a web server here. Realistically, that's a much more secure and reliable option than directly exposing Netdata to the internet and may not even need any changes in the VPC configuration to work.

I could definitely do that. 

> This is possibly just me, but I feel we should add links to the docs on running Netdata proxied through a web server here. Realistically, that's a much more secure and reliable option than directly exposing Netdata to the internet and may not even need any changes in the VPC configuration to work.

Done!

PR #7914, opened 2020-01-30
Adds new simpler (Alpine based) Dockerfile for quick dev and testing

### LINKED ISSUE #7899: Docker files for dev / testing.
#### Summary

One of the problems that we face in reviewing PRs is that there are many complex configurations for Netdata that are difficult to setup and use for testing. The current approach is to use a set of VMs that contain various OS configurations, Netdata installs and all of the other tools that we need for a full monitoring stack. The problem with this is that it is not robust, easy to maintain and does not allow repeatable testing during development.

We can overcome these issues with a custom version of the Dockerfiles that are used for packaging Netdata for different target environments. The main differences between our use-cases and packaging are:

* Must support incremental builds for dev - not always a clean install.
* Must support debugging / valgrind and other tools.
* Must support injection of partial configuration files for a predictable setup.
* Must be usable in docker-compose to create multi-node configurations.

### Expected outcome

Support the following configurations for dev/testing:

* Simple single Netdata installation (i.e. the basic one-node case).
* Master/slave configurations with multiple Netdata nodes.
* ACLK testing configuration with custom fork, VerneMQ etc.
* Monitoring stack with Prometheus / Grafana and other configs that we need to test exporting on.

prologic:
##### Summary

Fixes #7899

Usage:

```sh
$ docker build -t netdata . 
$ docker run --rm -p 19999:19999/tcp netdata
```

Or simply:

```sh
$ docker build -t netdata . && docker run --rm -p 19999:19999/tcp netdata
```

Passing installer options:

```sh
$ docker build --build-arg INSTALLER_ARGS="--disable-dbengine" -t netdata .
```

Timeing / Performance:

With Docker's caching and being rather careful here (see: [Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)):

- Nothing Cached: ~1m42s to build/run
- Dependencies Cached: ~42s to build/run
- Everything Cached: ~3s to run (_nothing rebuilt_)

##### Component Name

area/ci
area/packaging

##### Additional Information

Yes this is a little bit of duplication here; but this is a vastly more
simplified `Dockerfile` at the top-level to make the entire dev/test
cycle much faster and easier for ourselves, maintainers, contributors
and really anyone.

**Note:** This is only really useful for quick dev/test cycles and does
          nothing to validate that the builds work on GCC or GLIBC or any
          other OS / Distro for that matter (_since this is using ALpine_)
          but IHMO that's okay because we have other CI/CD scripts/flows
          that cover that when a PR goes up.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,69 @@
+# TODO: Create a netdata/package-builder:alpine9
+#FROM netdata/package-buidler:alpine AS build
+FROM alpine:3.9 AS build
+
+# Install Dependencies
+RUN apk add --no-cache -U alpine-sdk bash curl libuv-dev zlib-dev \
+	                  util-linux-dev libmnl-dev gcc make git autoconf \
+                          automake pkgconfig python logrotate
+
+# Pass optional ./netdata-installer.sh args with --build-arg INSTALLER_ARGS=...
+ARG INSTALLER_ARGS=""
+
+# Copy Sources
+# Can also bind-mount sources with:
+# $ docker run -v $PWD:/netdata
+
+WORKDIR /netdata
+COPY . .
+
+# Build
+RUN ./netdata-installer.sh --dont-wait --dont-start-it --disable-go "${INSTALLER_ARGS}"
+
+FROM alpine:3.9 AS runtime
+
+# Install runtime dependeices
+RUN apk --no-cache -U add curl bash libuv zlib util-linux libmnl python
+
+# Create netdata user/group
+RUN addgroup -S netdata && \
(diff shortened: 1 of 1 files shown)


prologic:
> @prologic
> 
> 1. How do you intend to keep this in sync with the official Docker image?

I don't. However the "devel" package duplicate will be elimiated in both cases once the mentioned TODO is addressed.

> 2. What specific use case/limitation are you addressing?

As per the summary. Quick dev/test cycles.

There are a few extra things that our "production image" does that you don't need in a quick dev/test cycle nor want to wait for.

prologic:
@ncmans Thanks for making me triple check :)

PR #7925, opened 2020-01-30
Cleanup of macOS installation docs

joelhans:
##### Summary

The existing macOS installation page is barebones and poorly-formatted—this update makes it more readable and encourages users to install Netdata with Homebrew. I also wanted to include details about where the config/stock config directory ends up, since that's caught me off-guard more than once.

I think the plan is to eventually have our own Homebrew repo with a formulae we supply ourselves, but that's TBD. Maybe @prologic or @Ferroin have some more details on that.

##### Component Name

docs

### DIFF
--- a/packaging/installer/methods/macos.md
+++ b/packaging/installer/methods/macos.md
@@ -5,40 +5,39 @@ can use any of Netdata's [external plugins](../../../collectors/plugins.d/README
 have installed on your macOS system. You could also use a macOS system as the master node in a [streaming
 configuration](../../../streaming/README.md).
 
-We support two methods of installing Netdata on macOS, although both involve [Homebrew](https://brew.sh/). Install that
-first, then proceed to either install Netdata via a Homebrew package, or directly from source.
+We recommend installing Netdata with the community-created and -maintained [**Homebrew
+package**](#install-netdata-with-the-homebrew-package). 
 
-The Homebrew package will be easier to install, although it only updates with [major
-releases](../README.md#nightly-vs-stable-releases), not nightly updates. 
-
--   [Install Homebrew on macOS](#install-homebrew-on-macos)
 -   [Install Netdata via the Homebrew package](#install-netdata-with-the-homebrew-package)
 -   [Install Netdata from source](#install-netdata-from-source)
 
-## Install Homebrew on macOS
+## Install Netdata with the Homebrew package
 
-The first step, for either method, is to install Homebrew on your macOS system. Use their installation script:
+If you don't have [Homebrew](https://brew.sh/) installed already, begin with their installation script:
 
 ```bash
 /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 ```
 
-Now that you have Homebrew installed, you can move on to either of the two following installation methods.
-
-## Install Netdata with the Homebrew package
-
-Homebrew's community maintains a package that helps you install Netdata and its dependencies in one step:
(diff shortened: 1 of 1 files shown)


packaging/installer/methods/macos.md
```diff
@@ -1,42 +1,74 @@
...
+If you want to use the [database engine](../../../database/engine/README.md) to store your metrics, you need to download
+and install the [Judy library](https://sourceforge.net/projects/judy/) before proceeding compiling Netdata.
 
-# install required packages
-brew install ossp-uuid autoconf automake pkg-config
+Next, download Netdata from our GitHub repository:
 
-# download Netdata
+```bash
 git clone https://github.com/netdata/netdata.git --depth=100
```
Ferroin: This line can be updated to use `--depth=1` instead of `--depth=100`. An install doesn't need the last 100 commits, only the most recent one.

Alternatively, if we think we need some history for some reason, just drop the `--depth` option altogether.

ilyam8: i recall we made it 100 for a reason, it is related with google analytics somehow. We had some problems with it with 1.

joelhans: Fascinating! I'll do some digging to see what I can find about this.

joelhans: Here's the PR where it changed: https://github.com/netdata/netdata/pull/5119. I guess it had to do with moving the doc from `installer/README.md` to `packaging/installer/README.md`? I'm not very sure, to be honest.

My gut says keep it as-is for now.

ilyam8: Yes, it was versioining problem, there were no version with depth=1 or something

Ferroin: If this is actually related to analytics beyond ensuring versions are sane, I'll be amazed.

Digging a bit further, it looks like we need the history for `git describe` (which we use during the build) to work correctly. In this case, this is a classic example of an arbitrary magic number that doesn't actually fix anything, and should just be removed, because the case this is supposed to prevent from happening (no recent tags in the local history) can still happen (especially considering that users may set `depth` to whatever the hell they want anyway).

joelhans: Okay, I'll follow your lead on this. If you think it should be cut, I'll cut it!

Ferroin: Posting one final comment here so that there's a proper record of my reasoning on this so that people don't have to go digging to find it in the future:

When building out of a git clone of the source, we use `git describe` to determine the version string. `git describe` looks for the most recent tag in the history of the current branch to construct the first part of the version, and then uses the commit hash as the second part if the current commit isn't the one with the discovered tag. In the event that there is no tag (as is usually the case when doing a shallow clone with a short history), it just uses the commit hash as the version, which is not what we want because it breaks assumptions made by many tools that try to parse version numbers.

The original change that bumped this to 100 was a prophylactic to mitigate this, but it's not really a correct fix because it's perfectly reasonable that the most recent tag won't be in the last 100 commits (in fact, I'm pretty sure it isn't as of the current master branch as I'm writing this), so we can still have the same type of failure here.

Realistically, a shallow clone is just an efficiency boost for end users and nothing else, but even a deep clone of `master` isn't particularly bad (we haven't even hit 100MB yet for the clone process), so there's not a huge benefit to doing a shallow clone unless you're on a super slow connection or have to pay ridiculous premiums for data usage.

Given all of this, there's not much point to having a `--depth` option here. Most users won't care about it, and the few that do should be savvy enough to use `--depth` appropriately.

prologic:
It works. Tested it on my Mac.

PR #7938, opened 2020-01-31
Fixes error/warnings found by shellcheck for ./packaging/installer/netdata-updater.sh

prologic:
See #7195

**Depend on:** #7937 (_review that first_)

##### Summary

As per title

##### Component Name

area/packaging

##### Additional Information

PR #7939, opened 2020-01-31
Adds support for only performing updates if there is a newer version

### LINKED ISSUE #7195: netdata-updater.sh always updates during first run after installation
##### Bug report summary
The netdata-updater.sh always performs a reinstallation during the first run on a clean installation

##### OS / Environment
All

v1.18.1

##### Component Name
netdata-updater.sh, kickstart.sh kickstart-static64.sh

##### Steps To Reproduce
Perform a clean netdata installation using kickstart.sh. Run the netdata-updater.sh script (/etc/cron.daily/netdata-updater.sh). A reinstallation/update will be performed regardless of whether there is a newer version available or not.

##### Expected behavior
An update should only occur if there is a newer version available.

### LINKED ISSUE #7196: Remove support for local tarballs from netdata-updater.sh
##### Feature idea summary
Remove support for local tarballs from netdata-updater.sh

The netdata-updater.sh does not need to support local tarballs. If the user needs to upgrade their netdata installation on a system without internet access, they can reinstall using kickstart.sh or kickstart-static64.sh.

The netdata-updater.sh should only be used for non-interactive, automatic updates on systems with internet access.

Removing said support will make this script simpler and more robust.

prologic:
- Fixes #7195
- Fixes #7196
- **Depends on:** #7938 -> #7937 (_review those first_)

##### Summary

As per title

##### Component Name

area/packaging

##### Additional Information

I'm not sure how robust this will be long-term as this relies on our versioning
scheme to remain consistent and always of the form: `vX.Y.Z-Build-Sha`.

##### Test Plan

- [x] Test with a build from master
- [x] Test with a build from a previous release


packaging/installer/netdata-updater.sh
```diff
@@ -170,50 +201,46 @@ export NETDATA_LIB_DIR="${NETDATA_LIB_DIR:-${NETDATA_PREFIX}/var/lib/netdata}"
...
+  download "${NETDATA_TARBALL_CHECKSUM_URL}" "${TMPDIR}/sha256sum.txt"
+  download "${NETDATA_TARBALL_URL}" "${TMPDIR}/netdata-latest.gz.run"
+  if ! grep netdata-latest.gz.run "${TMPDIR}/sha256sum.txt" | safe_sha256sum -c - > /dev/null 2>&1; then
+    fatal "Static binary checksum validation failed. Stopping netdata installation and leaving binary in ${TMPDIR}"
+  fi
+
+  # Do not pass any options other than the accept, for now
+  sh "${TMPDIR}/netdata-latest.gz.run" --accept "${REINSTALL_OPTIONS}"
+
+  if ! rm -r "${TMPDIR}"; then
```
knatsakis: I think the author's intention here was to not delete the $TMPDIR if `sh netdata-latest.gz.run` returned an error.

I am not sure if you changed this behaviour deliberately.

prologic: Let me see...

prologic: Yes you are right; this was a mistake. Fixed in #7938

prologic:
> Also, there is 1 shellcheck warning

Hmm I don't see it `shellcheck` turns up empty.

prologic:
> > Also, there is 1 shellcheck warning
> 
> Hmm I don't see it `shellcheck` turns up empty.

[Found it](https://github.com/netdata/netdata/pull/7939/checks?check_run_id=422252988) and fixed. My local `shellcheck` wasn't showing it :/

packaging/installer/netdata-updater.sh
```diff
@@ -73,14 +73,45 @@ download() {
   fi
 }
 
-set_tarball_urls() {
-  local extension="tar.gz"
+function parse_version() {
+  r="${1}"
+  if echo "${r}" | grep -q '^v.*'; then
```
ncmans: nit: you could just remove the starting v uncondtionally (ie `sed 's/^v//'`) instead of this if block

Issue #7799, opened 2020-01-20
go.d.plugin: do not download on every update


ilyam8:
##### Feature idea summary

We download `go.d.plugin` and its modules configuration on every update. It is a little bit annoying.

##### Expected behavior

Download `go.d.plugin` and its modules configuration only if:
- there is no `go.d.plugin` in the `plugins.d/`
- existent `go.d.plugin` version is not latest

___

There is `-v` option to get current version. Possible output:

- build w/o `ldflags`

```
go.d.plugin, version: unknown
```

- build w `ldflags` (using `make`)

```
go.d.plugin, version: v0.14.1
go.d.plugin, version: v0.14.1-dirty
go.d.plugin, version: v0.14.1-1-g4c5f98c
go.d.plugin, version: v0.14.1-1-g4c5f98c-dirty
```

- very old version has not `-v`, it prints to `stderr`

```
unknown flag `v'
```
___

Actual version can be found in the [packaging/go.d.version](https://github.com/netdata/netdata/blob/master/packaging/go.d.version)

We need:
- check for `go.d.plugin` in `plugins.d/`, if it is not there - download
- execute `go.d.plugin -v`, extract version, if version < extracted or `go.d.plugin -v` returns error or `unknown` - download

@Ferroin we had a talk, you suggestes to assign you. If you have time i suggest to fix it if it is relatively easy/fast fix, because it ~~saves some puppies~~ gives a value for the user (download the whole thing on every update - every day - is not good).

PR #7946, opened 2020-02-01
install go.d.plugin only if there is new version

ilyam8:
##### Summary

Fixes: #7799

##### Component Name

netdata-installer

##### Additional Information

Tests:

download latest if:

- [x] no go.d.plugin
- [x] go.d.plugin actual version is less then latest
- [x] error during checking/parsing versions

skip downloading if:
- [x] NETDATA_DISABLE_GO param is set
- [x] go.d.plugin actual version is equal to the latest
- [x] go.d.plugin actual version is greater then the latest

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -887,9 +887,9 @@ should_install_go() {
 
   govercomp "$latest_version" "$actual_version"
   case $? in
-  1) return 0 ;; # >
-  3) return 0 ;; # error
-  *) return 1 ;; # =,<
+  0) return 1 ;; # =
+  2) return 1 ;; # <
+  *) return 0 ;; # >, error
   esac
 }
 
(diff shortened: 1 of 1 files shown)


ilyam8:
@Ferroin i assume it is not enough to fix it in the `netdata-installer.sh`, i see `install_go` function exists in the 4 different files:

> https://github.com/netdata/netdata/search?q=install_go&unscoped_q=install_go

Could you help me?

The fix seems to be working, at least for installer (which i use prettry often)

ilyam8:
i understand that i need to copy paste the fix to the
- [/contrib/debian/install_go.sh](https://github.com/netdata/netdata/blob/cf6e42273c64fdfdf7a11254d9983b3cff87849c/contrib/debian/install_go.sh)
- [netdata.spec.in](https://github.com/netdata/netdata/blob/1477784048319e84127e466add6eb4a092c235e1/netdata.spec.in)

but i dont know how to test the changes there

ilyam8:
@Ferroin if those 2 files are used for building i guess i dont need to modify them, if so the PR is ready

Ferroin:
> @Ferroin if those 2 files are used for building i guess i dont need to modify them, if so the PR is ready

You're correct, the spec file and stuff under `contrib/debian/` are only used for package builds.

Issue #7803, opened 2020-01-20
xenstat plugin doesn't build against Xen 4.13


stormi:
##### Bug report summary
The xenstat plugin, that compiled well against xen 4.11.1, fails to build against xen 4.13.

Here are the error messages:
```
collectors/xenstat.plugin/xenstat_plugin.c: In function 'xenstat_collect':
collectors/xenstat.plugin/xenstat_plugin.c:418:5: warning: implicit declaration of function 'xenstat_node_freeable_mb' [-Wimplicit-function-declaration]
     node_metrics.freeable_mb = xenstat_node_freeable_mb(node);
     ^
collectors/xenstat.plugin/xenstat_plugin.c:460:9: error: unknown type name 'xenstat_tmem'
         xenstat_tmem *tmem = xenstat_domain_tmem(domain);
         ^
collectors/xenstat.plugin/xenstat_plugin.c:460:9: warning: implicit declaration of function 'xenstat_domain_tmem' [-Wimplicit-function-declaration]
collectors/xenstat.plugin/xenstat_plugin.c:460:30: warning: initialization makes pointer from integer without a cast [enabled by default]
         xenstat_tmem *tmem = xenstat_domain_tmem(domain);
                              ^
collectors/xenstat.plugin/xenstat_plugin.c:461:9: warning: implicit declaration of function 'xenstat_tmem_curr_eph_pages' [-Wimplicit-function-declaration]
         d->tmem.curr_eph_pages = xenstat_tmem_curr_eph_pages(tmem);
         ^
collectors/xenstat.plugin/xenstat_plugin.c:462:9: warning: implicit declaration of function 'xenstat_tmem_succ_eph_gets' [-Wimplicit-function-declaration]
         d->tmem.succ_eph_gets = xenstat_tmem_succ_eph_gets(tmem);
         ^
collectors/xenstat.plugin/xenstat_plugin.c:463:9: warning: implicit declaration of function 'xenstat_tmem_succ_pers_puts' [-Wimplicit-function-declaration]
         d->tmem.succ_pers_puts = xenstat_tmem_succ_pers_puts(tmem);
         ^
collectors/xenstat.plugin/xenstat_plugin.c:464:9: warning: implicit declaration of function 'xenstat_tmem_succ_pers_gets' [-Wimplicit-function-declaration]
         d->tmem.succ_pers_gets = xenstat_tmem_succ_pers_gets(tmem);
         ^
  CC       daemon/common.o
[...]

PR #7951, opened 2020-02-03
Removed tmem stat collection from xenstat plugin for Xen 4.13

rushikeshjadhav:
##### Summary
The compilation of netdata fails #7803  This is because of removal of tmem functionality from Xen 4.13 and thus from libxenstat.

##### Component Name
collectors/xenstat.plugin

### DIFF
--- a/collectors/xenstat.plugin/xenstat_plugin.c
+++ b/collectors/xenstat.plugin/xenstat_plugin.c
@@ -83,16 +83,6 @@ struct vcpu_metrics {
     struct vcpu_metrics *next;
 };
 
-struct tmem_metrics {
-    unsigned long long curr_eph_pages;
-    unsigned long long succ_eph_gets;
-    unsigned long long succ_pers_puts;
-    unsigned long long succ_pers_gets;
-
-    int pages_chart_generated;
-    int operation_chart_generated;
-};
-
 struct vbd_metrics {
     unsigned int id;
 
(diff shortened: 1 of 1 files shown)


ilyam8:
hi @rushikeshjadhav, thx!

please sign the cla

ilyam8:
Hi @stormi could you test the fix?

stormi:
> Hi @stormi could you test the fix?

Well, @rushikeshjadhav and I both work for the XCP-ng project and he made the pull request because we need that Xen 4.13 support in netdata soon and have no idea when the maintainers will fix it.

We don't know the netdata codebase well so all I can say is that the PR fixes the build. Now, would there be changes that need to be made to other parts of netdata?

ilyam8:
@vlvkobal please take a look

vlvkobal:
Don't we need to keep the tmem stuff supported in earlier versions of the libxenstat?

Ferroin:
> Don't we need to keep the tmem stuff supported in earlier versions of the libxenstat?

Possibly. None of the frontend stuff ever really got past the point of being somewhat experimental, and even if it did it was _super_ niche (the only useful thing I know of that you could do with it was having self-ballooning PV guests, but only if they were running Linux, and usually only if it was a custom build of the kernel).

vlvkobal:
@rushikeshjadhav, @stormi what are your thoughts on keeping the `tmem` stats for older versions of the `libxenstat`? Could it make sense?

stormi:
In general it makes sense to me not to remove features that were working previously, but as @Ferroin said, tmem was "*super* niche" (someone else confirmed that to me) so I think we could simply remove it in this case and only add it back if just one person misses it, which I doubt.

vlvkobal:
Tested it on XCP-ng 8.0. It works.

PR #7964, opened 2020-02-03
Indicate FreeIPMI supported in Docker image

lassebm:
##### Summary
As per recent PRs, FreeIPMI is now supported in the Docker image.


joelhans:
@Ferroin @prologic Adding you both as reviewers here, as this is probably related to work you've been doing. The file itself looks good to me but I'd like confirmation from the SRE team that this is the correct change!

joelhans:
Thank you, @lassebm!

PR #7971, opened 2020-02-04
Fixes static builds and nightlies

prologic:
##### Summary

This was the cause for our nightlies being broken for so long :/

#7725 broke this :/ So sorry :/

##### Component Name

area/packaging

##### Additional Information

🤦‍♂️

##### Test Plan

**Beofre:**

```sh
prologic@Jamess-iMac
Tue Feb 04 17:12:28
~/NetData/netdata
 (master) 0
$ dki -t -v /var/run/docker.sock:/var/run/docker.sock -v $PWD:/netdata -w /netdata debian:buster /bin/bash
root@9d864f5524d8:/netdata# ./packaging/makeself/build-x86_64-static.sh
./packaging/makeself/build-x86_64-static.sh: 327: ./packaging/makeself/../installer/functions.sh: Syntax error: "(" unexpected (expecting "then")
```

**After:**

```sh
prologic@Jamess-iMac
Tue Feb 04 17:16:44
~/NetData/netdata
 (master) 0
$ dki -t -v /var/run/docker.sock:/var/run/docker.sock -v $PWD:/netdata -w /netdata debian:buster /bin/bash
root@f6bb12d81e86:/netdata# ./packaging/makeself/build-x86_64-static.sh
[/netdata]# docker run -v /netdata:/usr/src/netdata.git:rw alpine:3.7 /bin/sh /usr/src/netdata.git/packaging/makeself/install-alpine-packages.sh \n./packaging/makeself/../installer/functions.sh: line 227: docker: command not found
root@f6bb12d81e86:/netdata# exit
```

### DIFF
--- a/packaging/installer/install-required-packages.sh
+++ b/packaging/installer/install-required-packages.sh
@@ -511,12 +511,11 @@ require_cmd() {
 
   local wanted found
   for wanted in "${@}"; do
-    if command -v "${wanted}" 2> /dev/null; then
+    if command -v "${wanted}" > /dev/null 2>&1; then
       found="$(command -v "$wanted" 2> /dev/null)"
     fi
     [ -n "${found}" ] && [ -x "${found}" ] && return 0
   done
-
   return 1
 }
 
--- a/packaging/makeself/build-x86_64-static.sh
+++ b/packaging/makeself/build-x86_64-static.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 # SPDX-License-Identifier: GPL-3.0-or-later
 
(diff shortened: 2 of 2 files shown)


prologic:
Fortunately this was the only instance of this I broke:

```sh
$ git log -v -p --author='1290234+prologic@users.noreply.github.com' | grep -E '#!\/bin\/sh'
```

prologic:
I really _should_ explain what happened here... I have this rather strong opinion that you should write POSIX Sh wherever possible for simple stuff and once that gets too complicated you should resort to using a different language /tool. I did this in a few place early on when I first joined NetData but I didn't realise at the time how much we really relied on Bash and Bash(isms) / features :/

Basically the problem here is you can't go from running a shell script in `#!/bin/sh` to `#!/bin/ bash` **and** source in another script with `source /path/to/functions.sh` or `. /path/to/functions.sh` and expect this to work. The parent shell (_the current process / shell interpreter_) will horribly b0rk.

:/

Lesson(s):

- Understand more deeply before changing something
- Improve the visibility of our CI/CD pipelines
- Decouple our CI/CD pipelines and get off Travis

prologic:
Finally also fixed and found the bug causing the Lifecycle tests to fail on Debian:

It will still fail on Travis CI however because it downloads from master -- the master verison of `require_cmd()` is broken.

ilyam8:
this PR pass 34 checks, safe to approve even without reading it! 😄

PR #7984, opened 2020-02-05
Fix permissions issues caused by 986bc2052.

Ferroin:
##### Summary

For some reason, without `sudo` here, we get permissions issues in CI,so re-add it.

##### Component Name

area/packaging

##### Additional Information

I'm honestly not even sure _why_ we're changing file ownership here. Looking at the git history gives no real insights.

This is, however, what's causing nightlies to fail.


Ferroin:
I'm waiting on the build to go Green, and also hoping for review by an SRE.

prologic:
Oh 🤦‍♂️

PR #7992, opened 2020-02-06
/docs/generator: remove monitoring suffix from collectors readme header

ilyam8:
##### Summary

ssia

Fixes: netdata/marketing#79

##### Component Name

`/docs`


ilyam8:
There is some hardcoded stuff in the buildyaml.sh, i think i need to be more specific.

cakrit:
It looks like this is removing the word "Monitoring" from the README completely. We want to remove it only from the links. So it's a step that would need to be done later, not sure how. Will check.

ilyam8:
> We want to remove it only from the links. 

It makes sense, have no idea how to do it. Ok, this pr is definitely `wontfix` then.

Issue #7878, opened 2020-01-28
outdated docs: stale info about supported collectors


ilyam8:
##### Bug report summary

[Removing](https://github.com/netdata/netdata/pull/7853) `unbound` python module from the docs i noticed that following md files are outdated:

-   [/README.md](https://github.com/netdata/netdata/blob/master/README.md)
-   [/docs/what-is-netdata.md](https://github.com/netdata/netdata/blob/master/docs/what-is-netdata.md)
-   [/docs/Add-more-charts-to-netdata.md](https://github.com/netdata/netdata/blob/master/docs/Add-more-charts-to-netdata.md)

@joelhans 

1. `/README.md` and `/docs/what-is-netdata.md` look identical
2. `/docs/Add-more-charts-to-netdata.md` name != content

All 3 dcs overlap - have ~ same list of collectors.

I think it would be nice to have it in one place, not it N places.

 What do you think?

##### Component Name

`docs`

##### Expected behavior

collectors info in docs are up to date

___

### Internal Plugins

|plugin|lang|O/S|description|
|:----:|:--:|:-:|:----------|
|[cgroups.plugin](https://github.com/netdata/netdata/tree/master/collectors/cgroups.plugin)|`c`|linux|collects resource usage of Containers, libvirt VMs and systemd services, on Linux systems.|
|[checks.plugin](https://github.com/netdata/netdata/tree/master/collectors/checks.plugin)|`c`|any|a debugging plugin.|
|[diskspace.plugin](https://github.com/netdata/netdata/tree/master/collectors/diskspace.plugin)|`c`|linux|collects disk space usage metrics on Linux mount points.|
|[freebsd.plugin](https://github.com/netdata/netdata/tree/master/collectors/freebsd.plugin)|`c`|freebsd|collects resource usage and performance data on FreeBSD systems.|
|[idlejitter.plugin](https://github.com/netdata/netdata/tree/master/collectors/idlejitter.plugin)|`c`|any|measures CPU latency and jitter on all operating systems.|
|[macos.plugin](https://github.com/netdata/netdata/tree/master/collectors/macos.plugin)|`c`|macos|collects resource usage and performance data on MacOS systems.|
|[proc.plugin](https://github.com/netdata/netdata/tree/master/collectors/proc.plugin)|`c`|linux|collects resource usage and performance data on Linux systems.|
[...]

PR #7995, opened 2020-02-06
Docs: Change sed to allow parentheses in heading links

joelhans:
##### Summary

Similar to #7431, I found that when you putting parentheses in headings in the documentation you can't link to them using an anchor link. I struggled with this when trying to work on collectors docs with @ilyam8 and #7878.

For example, this pair of anchor link and header will fail during build.

```
[this is an anchor link to the heading below](#lets-try-this-test)

## Let's try this (test)
```

This change will cut parentheses from the string when building the anchor tag so it all works OK.

##### Component Name

docs

### DIFF
--- a/docs/generator/checklinks.sh
+++ b/docs/generator/checklinks.sh
@@ -55,7 +55,7 @@ testinternal () {
 	ilnk=${3}
 	header=${ilnk//-/}
 	dbg "   - Searching for \"$header\" in $ifile"
-	tr -d '[],_.:? `'< "$ifile" | sed -e 's/-//g' -e "s/'//g" | grep -i "^\\#*$header\$" >/dev/null
+	tr -d '[],_.:? `'< "$ifile" | sed -e 's/-//g' -e "s/'//g" -e "s/(//g" -e "s/)//g" | grep -i "^\\#*$header\$" >/dev/null
 	if [ $? -eq 0 ] ; then
 		dbg "   - $ilnk found in $ifile"
 		return 0


docs/generator/checklinks.sh
```diff
@@ -55,7 +55,7 @@ testinternal () {
 	ilnk=${3}
 	header=${ilnk//-/}
 	dbg "   - Searching for \"$header\" in $ifile"
-	tr -d '[],_.:? `'< "$ifile" | sed -e 's/-//g' -e "s/'//g" | grep -i "^\\#*$header\$" >/dev/null
+	tr -d '[],_.:? `'< "$ifile" | sed -e 's/-//g' -e "s/'//g" -e "s/(//g" -e "s/)//g" | grep -i "^\\#*$header\$" >/dev/null
```
ilyam8: You could use `[]`, it means `any of`

> sed -e "s/[-'\)\(]//g"

ilyam8: Not sure if it looks better 😄 Just saying.

ilyam8: > tr -d '[],_.:? `'< "$ifile" | sed ...

Should we use only `sed` or `tr` there? i see we use both to achieve same goal - remove set of characters.

cakrit: Yes, I think I was having trouble putting a single quote inside a single quote and dash meant range. No need to be sticklers about it, it will all go away with the learn site.

PR #8004, opened 2020-02-07
Update database servers group in apps_groups.conf

DefauIt:
- added group for column oriented databases - `columndb`
- added `clickhouse-server` to group `columndb`
- added influxd to `timedb` group


ilyam8:
thx @DefauIt

PR #8011, opened 2020-02-08
Fix missing variables on stream

### LINKED ISSUE #8010: Missing information on stream
##### Bug report summary
We are missing all container variables when the stream is started from slave to master, it is necessary to bring this new information to the protocol.

##### OS / Environment
```
All environments
```

##### Netdata version
netdata v1.19.0-463-g9cf6f461
 

##### Component Name
Stream

##### Steps To Reproduce
1. Set master Netdata and start it
2. Configure stream.conf and starts the stream
3. Access `/api/v1/info` on slave
4. Access `/host/hostname/api/v1/info` on master

The information on 3 and 4 will be different.

##### Expected behavior
To transfer container information from slave to master

thiagoftsm:
##### Summary
Fixes #8010 

This PR fixes the problem described in the issue.
##### Component Name
Stream
##### Additional Information
To confirm that everything is running correct, it is necessary to set master and slave with this branch. After this access the following URLs:

 - http://slave:19999/api/v1/info
-  https://master:19999/host/darkstar/api/v1/info

The container information must be the same in both.


stelfrag:
I put comments in the issue instead of here :slightly_frowning_face:  I put a link below for reference:

https://github.com/netdata/netdata/issues/8010#issuecomment-583843007
https://github.com/netdata/netdata/issues/8010#issuecomment-583844487

thiagoftsm:
> I put comments in the issue instead of here slightly_frowning_face I put a link below for reference:

I brought the 'container_` variables, but I missed the other 4, but I fixed this with the last commit, please, test again in your side.

Thank you for the initial tests!

PR #8017, opened 2020-02-10
Fix: Documentation states in one place default is dbengine and in another save is default

underhood:
##### Summary
Documentation states in one place default is dbengine and in another save is default which might confuse user.

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -50,8 +50,8 @@ You can select the memory mode by editing `netdata.conf` and setting:
 
 ```conf
 [global]
-    # ram, save (the default, save on exit, load on start), map (swap like)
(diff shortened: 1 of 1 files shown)


underhood:
docu_info

database/README.md
```diff
@@ -50,8 +50,8 @@ You can select the memory mode by editing `netdata.conf` and setting:
 
 ```conf
 [global]
-    # ram, save (the default, save on exit, load on start), map (swap like)
-    memory mode = save
+    # dbengine (default), ram, save (the default if dbengine not availible), map (swap like), none, alloc
```
ilyam8: a typo

> availible

PR #8025, opened 2020-02-10
Add support for libmosquitto to netdata-installer.sh for ACLK support.

Ferroin:
##### Summary
Implements/Closes #7906 

This adds basic support for fetching, building, and preparing for usage our custom fork of libmosquitto. 

This largely uses the same control-flow logic as the handling of the Go plugin does, namely:

* If we fail to download the sources due to a lack of `curl` or `wget`, we fail the build and complain to the user about it.
* If the checksum verification for the source archive fails, we fail the build and tell the user about it.
* If we fail for any other reason, we continue.

The whole sequence can be disabled by specifying `--disable-libmosquitto` on install.

It supports local copies of the source archive just like the Go plugin does. if `NETDATA_LOCAL_TARBALL_OVERRIDE_MOSQUITTO` is defined, we'll use that as a path to a local copy of the source archive instead of downloading it.

Failure to unpack or build the sources, as well as failure to copy the generated files to the right place for the Netdata build, will be logged on both the console and in the installer log file.

This also generalizes the code used by the go.d plugin to download tarballs so that it's more readily reusable.

##### Component Name

area/packaging

##### Additional Information

I've done basic testing on Linux/FreeBSD for this by running the modified script and verifying that the required files end up under `mosquitto/` in the Netdata sources. This is currently untested on macOS as I do not have a system to test on (which should finally be remedied some time within the next few weeks).

The failure cases for this are unfortunately not well tested, as it's actually kind of hard to force the unpacking, building, or copying to fail.

PR #8030, opened 2020-02-10
Check if ACLK can be built

### LINKED ISSUE #7740: ACLK - Packaging / installation of the agent with libmosquitto statically linked
In order to deploy the results of #7739 into our production code, all available methods of building / installing Netdata must be updated to:

* Pull in the latest version of the forked library (netdata/mosquitto).
* The fork needs to be updated to fix the cahnges that allow the static build of the library and any changes made during the spike.
* The autotools build needs to be updated.

# Expected result

* Instructions for pulling the fork / building the library.
* Necessary changes to autotools / CMake.

Anything else that is required to support the SRE team in #7906.

underhood:
##### Summary

Closes #7740, to be used together with PR #8025 .

Check if ACLK can be built:
- is static lib build of forked mosquitto present?
- is libwebsockets present (standard library)?

By default `autodetect`. If `--enable-aclk` and ACLK cannot be build error during build. If `--disable-aclk` don't build even if possible.

Currently this switch is purposefully undocumented and hidden behind ACLK=yes enviroment variable.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -594,7 +594,7 @@ netdata_SOURCES = $(NETDATA_FILES)
 
 if ENABLE_ACLK
 netdata_LDADD = \
-    mosquitto/lib/libmosquitto.a \
+    externaldeps/mosquitto/lib/libmosquitto.a \
     $(NETDATA_COMMON_LIBS) \
     $(NULL)
 else
--- a/Makefile.am
+++ b/Makefile.am
@@ -594,7 +594,7 @@ netdata_SOURCES = $(NETDATA_FILES)
 
 if ENABLE_ACLK
 netdata_LDADD = \
-    externaldeps/mosquitto/lib/libmosquitto.a \
+    externaldeps/mosquitto/libmosquitto.a \
     $(NETDATA_COMMON_LIBS) \
     $(NULL)
 else
--- a/aclk/mqtt.h
+++ b/aclk/mqtt.h
@@ -4,7 +4,7 @@
 #define NETDATA_MQTT_H
 
 #ifdef ENABLE_ACLK
-#include "externaldeps/mosquitto/lib/mosquitto.h"
+#include "externaldeps/mosquitto/mosquitto.h"
 #endif
 
 void _show_mqtt_info();
--- a/aclk/mqtt.h
+++ b/aclk/mqtt.h
@@ -4,7 +4,7 @@
 #define NETDATA_MQTT_H
 
 #ifdef ENABLE_ACLK
-#include "mosquitto/lib/mosquitto.h"
+#include "externaldeps/mosquitto/lib/mosquitto.h"
 #endif
 
 void _show_mqtt_info();
--- a/configure.ac
+++ b/configure.ac
@@ -432,7 +432,7 @@ AM_CONDITIONAL([ENABLE_HTTPS], [test "${enable_https}" = "yes"])
 #currenlty env var ACLK must be set to 'yes' to even consider building ACLK
 if test "${ACLK}" = "yes"; then
     AC_MSG_CHECKING([if libmosquitto static lib is present])
-    if test -f "externaldeps/mosquitto/lib/libmosquitto.a"; then
+    if test -f "externaldeps/mosquitto/libmosquitto.a"; then
         HAVE_libmosquitto_a="yes"
     else 
         HAVE_libmosquitto_a="no"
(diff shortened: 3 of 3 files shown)


amoss:
Tested it on linux, it detects the file is there but does not seem to update the `config.h` or Makefile appropriately. In the case that the file is not there it works fine. I think the check should be more specific, for other libraries we check the presence of a known procedure in the library instead of just checking the file.

Test:
* Ran PR 8025 first and saved the mosquitto library (these two PRs do not match up on the location of the library yet).
```
git clean -dxf
autoreconf -ivf
ACLK=yes ./configure
grep mosq Makefile
#netdata_DEPENDENCIES = mosquitto/lib/libmosquitto.a \
#netdata_LDADD = mosquitto/lib/libmosquitto.a \
```
```
grep HAVE_libmosq config.h
```
I would expect a commented line that indicates the HAVE_ flag is not set but it is not produced.
Snippet from config.log:
```
configure:8278: checking if libmosquitto static lib is present                                                          
configure:8285: result: no                                                                                              
configure:8288: checking if netdata ACLK can be enabled                                                                 
configure:8295: result: no                                                                                              
configure:8301: checking if netdata ACLK should be enabled                                                              
configure:8309: result: no
```
* Copy the mosquitto lib into the expected path
```
autoreconf -ivf 
ACLK=yes ./configure
grep mosq Makefile
#netdata_DEPENDENCIES = mosquitto/lib/libmosquitto.a \
#netdata_LDADD = mosquitto/lib/libmosquitto.a \
```
I see the same result:
```
grep mosq Makefile
#netdata_DEPENDENCIES = mosquitto/lib/libmosquitto.a \
#netdata_LDADD = mosquitto/lib/libmosquitto.a \
```
Relevant snippet of config.log:
```
configure:8278: checking if libmosquitto static lib is present                                                          
configure:8285: result: yes                                                                                             
configure:8288: checking if netdata ACLK can be enabled                                                                 
configure:8295: result: no                                                                                              
configure:8301: checking if netdata ACLK should be enabled                                                              
configure:8309: result: no
```

Should the result of 8288 not be yes? Again the config.h contains no HAVE_ flags.

underhood:
Hi there seems to be misunderstanding. Can you post full output of config?

Only flag that is set in config.h is ACLK_ENABLED the HAVE_ will not go
there by design.

Additionaly the check can build ACLK checks 2 things:
1. libmosquitto.a is present
2. Libwebsockets is present - I don't see the result of this check because
you copy pasted only part of output.

Anyway I will still recheck in morning for mistakes although I did test it
already.

underhood:
reset CI (random failure on SUSE build)

Makefile.am
```diff
@@ -594,7 +594,7 @@ netdata_SOURCES = $(NETDATA_FILES)
 
 if ENABLE_ACLK
 netdata_LDADD = \
-    mosquitto/lib/libmosquitto.a \
+    externaldeps/mosquitto/lib/libmosquitto.a \
```
amoss: This doesn't match the installer (@Ferroin please confirm). It should be `externaldeps/mosquitto/libmosquitto.a`.

Makefile.am
```diff
@@ -594,7 +594,7 @@ netdata_SOURCES = $(NETDATA_FILES)
 
 if ENABLE_ACLK
 netdata_LDADD = \
-    externaldeps/mosquitto/lib/libmosquitto.a \
+    externaldeps/mosquitto/libmosquitto.a \
```
Ferroin: The current version of #8025 actually does use `externaldeps/mosquitto/lib` for these.

aclk/mqtt.h
```diff
@@ -4,7 +4,7 @@
 #define NETDATA_MQTT_H
 
 #ifdef ENABLE_ACLK
-#include "externaldeps/mosquitto/lib/mosquitto.h"
+#include "externaldeps/mosquitto/mosquitto.h"
```
Ferroin: Same here.

configure.ac
```diff
@@ -432,7 +432,7 @@ AM_CONDITIONAL([ENABLE_HTTPS], [test "${enable_https}" = "yes"])
 #currenlty env var ACLK must be set to 'yes' to even consider building ACLK
 if test "${ACLK}" = "yes"; then
     AC_MSG_CHECKING([if libmosquitto static lib is present])
-    if test -f "externaldeps/mosquitto/lib/libmosquitto.a"; then
+    if test -f "externaldeps/mosquitto/libmosquitto.a"; then
```
Ferroin: And here as well.

Makefile.am
underhood: are you sure? i just run it and it doesn't create lib folder.

Ferroin: Ah, no, you're right. I forgot what I had changed when updating it to use `externaldeps`.

PR #8037, opened 2020-02-11
Agent cloud link 2

stelfrag:
##### Summary
* Support for popcorn-ing at connect
* Add delay function for reconnect (not integrated yet / awaiting for wss merge and callback)
  - Uses exponential delay + 0-999 ms jitter (with a maximum of 1024 seconds delay)
* MQTT build in exponential backoff delay to 1024 seconds
* Integrate chart detection (add and delete)
* Integrate alarm reload detection (disables chart / alarm updates while reloading)
* Integrate alarm trigger (health_log)
* Detect  collector addition / deletion. Triggers on_connect metadata
* JSON payload parse / tuning
* Encoding added
* Add debugging messages


database/rrdset.c
```diff
@@ -424,6 +424,9 @@ void rrdset_delete(RRDSET *st) {
     }
 
     recursively_delete_dir(st->cache_dir, "left-over chart");
+#ifdef ENABLE_ACLK
+    aclk_del_collector(st->rrdhost->hostname, st->plugin_name, st->module_name);
```
underhood: Just a general comment. I would maybe add a generic `del_collector_notif` function def and then here call all of those functions that registered.
The reasoning behind this is that RRDHOT/RRDSET/RRDDIM are already interdependent with all code making it quite a spaghetti.
That way RRDSET doesn't have to know about existence of some `ACLK` but will just call all registered callbacks of same function type.

stelfrag: It would be nice to have callbacks indeed. Luckily in this case we only try to delete a collector when a chart is deleted and the (host, plugin_name, module_name) combination ends up having 0 charts registered

ktsaou:


```diff

Clones added
============
- aclk/agent_cloud_link.c  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=4982881&bid=16455460)

aclk/agent_cloud_link.c
```diff
@@ -5,20 +5,30 @@
 
 // Read from the config file -- new section [agent_cloud_link]
 // Defaults are supplied
-int aclk_recv_maximum = 0; // default 20
-int aclk_send_maximum = 0; // default 20
 
 int aclk_port = 0;          // default 1883
```
underhood: We should change this to `#DEFINE` and remove the comment `// default 1883` because then the comments are not maintained (we are already using 9002 as default because of wss).
The same goes for aclk_hostname.

aclk/agent_cloud_link.c
```diff
@@ -133,6 +141,41 @@ struct aclk_query_queue {
...
+/*
+ * After a connection failure -- delay in milliseconds
+ * When a connection is established, the delay function
+ * should be called with
+ *
+ * mode 0 to reset the delay
+ * mode 1 to sleep for the calculated amount of time [0 .. ACLK_MAX_BACKOFF_DELAY * 1000] ms
+ *
+ */
+unsigned long int aclk_delay(int mode)
```
underhood: rename to aclk_reconnect_delay? would help with code readability

aclk/agent_cloud_link.c
```diff
@@ -327,59 +403,295 @@ struct aclk_query *aclk_queue_pop()
...
-    if (unlikely(action == PUBLICH_TOPIC_FREE)) {
-        if (likely(topic)) {
-            freez(topic);
-            topic = NULL;
-        }
+    if (unlikely(!global_base_topic)) {
+        char tmp_topic[ACLK_MAX_TOPIC + 1], *tmp;
+
+        snprintf(tmp_topic, ACLK_MAX_TOPIC, ACLK_TOPIC_STRUCTURE, is_agent_claimed());
+        tmp = strchr(tmp_topic, '\n');
```
underhood: I think this is too little too late. We should check it after loading and maybe in separate function validate_claiming_id? Additionally we should check for special characters like #/ etc.etc.

underhood: therefore the function that is loading it should either check it by itself or call `validate_claiming_id` which then should check it more carefully (not just look for endline) and set agent state to unclaimed if anything suspicious found

stelfrag: True this needs to go all the way back to the claiming code. I will open a new issue to handle it in the claim.c. Once we reach this point the agent is assumed claimed which should not happen of the id is anyway invalid

underhood:
> 0-999 ms jitter (with a maximum of 1024 seconds delay)

wouldn't it be better to implement jitter as percentage of the actual delay value?

aclk/agent_cloud_link.h
```diff
@@ -6,20 +6,21 @@
 #include "../daemon/common.h"
 #include "mqtt.h"
 
+#define ACLK_VERSION 1
+#define ACLK_THREAD_NAME "ACLKQ"
```
underhood: as the main thread is called `ACLK_Main`. I propose this one should be called `ACLK_Query`

stelfrag: Sure .. good idea

aclk/agent_cloud_link.c
stelfrag: Anything that can help to read code better -- sure!

database/rrdset.c
underhood: yes this comment applies to multiple similar places where now RRDXYZ is dependent on ACLK... but maybe idea to improve in future. I will not consider this to be a blocker for merge of this PR

aclk/agent_cloud_link.c
stelfrag: Good idea. I will add define for this `ACLK_DEFAULT_PORT` and also add `ACLK_DEFAULT_HOST`

aclk/agent_cloud_link.c
amoss: Well spotted @underhood - yes there is a new requirement on the claiming id that it cannot use any character that would be invalid as a component in an MQTT topic. We need to validate when it is loaded in - but we should also reject it with an error at the point of running the the claim script.

database/rrddim.c
```diff
@@ -522,7 +522,7 @@ int rrddim_unhide(RRDSET *st, const char *id) {
 
     rrddim_flag_clear(rd, RRDDIM_FLAG_HIDDEN);
 #ifdef ENABLE_ACLK
-    aclk_update_chart(st->rrdhost, st->id);
+    aclk_update_chart(st->rrdhost, st->id, ACLK_CMD_CHART);
```
underhood: same applies here i would prefer not to add ACLK as dependency for RRDDIM (there is many places like that). But as I said before i can accept this PR in this way (just prefer it to be done differently)

database/rrdset.c
amoss: Please don't do this - while it would make the code more generic we have no use-cases that warrant it. If a use-case is added in the future and there multiple notification to send out then it would make sense to add the callback that you are describing - but currently it adds complexity with no benefit.

underhood:
@stelfrag (not integrated yet / awaiting for wss merge and callback) -> should we create issue right away or please let me know when you need this + some detail. If we just leave it as an comment on this PR i worry it might get lost.

database/rrddim.c
stelfrag: Do we don't have callbacks unfortunately, which I would register and have it automatically fire. 
I also hated adding an extra param to the aclk_update_chart (tried with del_chart) but codacy started complaining about duplicate code.

amoss: Why is `st->rrdhost` being passed to the aclk? (Considering the discussion about leaking pointers)

stelfrag:
> > 0-999 ms jitter (with a maximum of 1024 seconds delay)
> 
> wouldn't it be better to implement jitter as percentage of the actual delay value?

Yes it could be done like that but I just followed some recommendations about IoT devices from various sources eg. [Implementing exponential backoff](https://cloud.google.com/iot/docs/how-tos/exponential-backoff).

database/rrddim.c
amoss: I think the check against `localhost` and extraction of the hostname should happen in `rrddim.c` instead of in the aclk code.

mfundul: In any case though, this pattern is everywhere in the code. Meaning that you get `st->rrdhost` or `rd->rrdset` or `rd->rrdset->rrdhost` all the time.

At this point, it should be safe to assume that the pointer of a parent structure is not going away as long as you keep the the child structure alive.

Nobody can guarantee that the parent structure is not being modified though.

aclk/agent_cloud_link.c
```diff
@@ -133,6 +141,41 @@ struct aclk_query_queue {
...
+    }
+
+    delay = (1 << fail);
+
+    if (delay >= ACLK_MAX_BACKOFF_DELAY) {
+        delay = ACLK_MAX_BACKOFF_DELAY * 1000;
+    }
+    else {
+        fail++;
+        delay = (delay * 1000) + (random() % 1000);
```
amoss: Why are you adding jitter to the back-off times? I've seen this in services that being started at the same time (e.g. Lets Encrypt renewals being triggered in crontabs) but in this case I don't understand what the common start time would be.

database/rrddim.c
stelfrag: I have tried both (passing hostname and *host). At this point there is no benefit either way and it needs to be revisited. In short I wanted a single line outside of non aclk code (eventually we may need to handle slaves as well if we reach a decision that slaves are automatically claimed)

stelfrag:
> @stelfrag (not integrated yet / awaiting for wss merge and callback) -> should we create issue right away or please let me know when you need this + some detail. If we just leave it as an comment on this PR i worry it might get lost.

Making a note of that

database/rrddim.c
amoss: Ok, well it does not make the situation any worse than it already is.

amoss: Using callbacks here does not remove the dependency between the codes - it moves it from being a compile-time dependency to a run-time dependency which has both benefits and drawbacks. The drawback is that it removes visibility of what the dependency is unless you go and find all the points in the code that add callbacks. In particular it breaks the ability of the editor or doxygen to give useful information.

aclk/agent_cloud_link.c
stelfrag: If the link to our cloud infrastructure drops when we have 10000 agents connected, they will all retry in 0, 1, 2, 4 .. etc seconds. So we will get 10000 attempts at the same time. The jitter will smooth that out. I put a link in a previous comment -- recommendation for IoT devices. 

Whether it will help or not -- I can't say :smile:

amoss: That is a really good point - but is one second of jitter enough to help in that scenario? This is something that should have been groomed with @cosmix and the cloud team.

aclk/agent_cloud_link.c
```diff
@@ -327,59 +403,295 @@ struct aclk_query *aclk_queue_pop()
...
+    hostname_hash = simple_hash(hostname);
+
+    // Note that the first entry is "dummy"
+    tmp_collector  = collector_list->next;
+    prev_collector = collector_list;
+    while (tmp_collector) {
+        if (plugin_hash == tmp_collector->plugin_hash &&
+            module_hash == tmp_collector->module_hash &&
+            hostname_hash == tmp_collector->hostname_hash &&
+            (!strcmp(hostname, tmp_collector->hostname)) &&
```
underhood: why compare hostname by hash and then also full string compare?

aclk/agent_cloud_link.c
mfundul: @stelfrag followed these guidelines https://cloud.google.com/iot/docs/how-tos/exponential-backoff
```
random_number_milliseconds is a random number of milliseconds less than or equal to 1000
```

stelfrag: Good point. If there is capacity planning already in place i.e. that we know we will handle X number of requests / second and we have a good estimation of agents claimed. The latter I think is unknown. It can be discussed of course and adjust the value as needed (either statically,  via agent configuration or cloud-to-agent setting)

aclk/agent_cloud_link.c
mfundul: That's the original code from the netdata agent that Kosta wrote. The idea is that if the hashes are different you did a very fast comparison.

underhood: ok thanks I get it, it's faster in case when this condition is evaluated where most of the time those strings are different. Why not use the same logic then for plugin_hash and module_hash?

underhood: ok thanks, clear

PR #8047, opened 2020-02-11
Add required build dep for ACLK dependencies to Dockerfile.

Ferroin:
##### Summary

This ensures that we have OpenSSL development files when building the Docker image, so that we can build the libmosquitto component of the ACLK functionality.

##### Component Name

area/packaging

##### Additional Information

This is literally the only change needed now for libmosquitto to work in the Docker images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,8 @@ FROM alpine:3.9 AS build
 
 # Install Dependencies
 RUN apk add --no-cache -U alpine-sdk bash curl libuv-dev zlib-dev \
-	                  util-linux-dev libmnl-dev gcc make git autoconf \
-                          automake pkgconfig python logrotate
+                          util-linux-dev libmnl-dev gcc make git autoconf \
+                          automake pkgconfig python logrotate openssl-dev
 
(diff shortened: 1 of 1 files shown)


prologic:
Don't we want to be modifying `./packaging/docker/`?

Ferroin:
> Don't we want to be modifying `./packaging/docker/`?

Uh, yeah, this is a good point, since we're not currently using this Dockerfile for our own Docker builds...

It's actually an update to the helper images though that's needed, as all that our packaging stuff does is pull one of our custom builder images. I'll post a PR for that shortly.

Ferroin:
> > Don't we want to be modifying `./packaging/docker/`?
> 
> Uh, yeah, this is a good point, since we're not currently using this Dockerfile for our own Docker builds...
> 
> It's actually an update to the helper images though that's needed, as all that our packaging stuff does is pull one of our custom builder images. I'll post a PR for that shortly.

Correction, we don't actually appear to need any changes there, it's already pulling in the OpenSSL development stuff we need.

prologic:
Even better 🤣

Issue #8056, opened 2020-02-12
ANN: NetData v1.19.0-432 broke the Automatic Updater


prologic:
# If you are on NetData v1.19.0-432 read on...

#7939 broke the NetData Automatic Updater (_usually run via Cron_).

If you are on this version; you have two quick-fixes:

1. Amend the script with this patch:

```diff
diff --git a/packaging/installer/netdata-updater.sh b/packaging/installer/netdata-updater.sh
index fc85039d..0f01ada8 100755
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -144,7 +144,7 @@ update() {
   info "Current Version: ${current_version}"
   info "Latest Version: ${latest_version}"

-  if [ "${latest_version}" -gt 0 ] && [ "${current_version}" -gt 0 ] && [ "${current_version}" -ge "${current_version}" ]; then
+  if [ "${latest_version}" -gt 0 ] && [ "${current_version}" -gt 0 ] && [ "${current_version}" -ge "${latest_version}" ]; then
     info "Newest version ${current_version} <= ${latest_version} is already installed"
   elif [ -n "${NETDATA_TARBALL_CHECKSUM}" ] && grep "${NETDATA_TARBALL_CHECKSUM}" sha256sum.txt >&3 2>&3; then
     info "Newest version is already installed"
```

This is usually located in `/usr/libexec/netdata/netdata-updater.sh` under a normal installation. You can easily find it by using `find`:

```sh
/netdata # find / -type f -name 'netdata-updater.sh'
/usr/libexec/netdata/netdata-updater.sh
```

2. Re-running the Installer which will also update your installation past `v1.19.0-432`

We apologise for any inconvenience caused.

----

Related Issues/Reports: #8027 #8029 #5053

amoss:
Why is the test the opposite way around to the info message? 
* The test is current >= latest
* The info is current <= latest

prologic:
> Why is the test the opposite way around to the info message?
> 
> * The test is current >= latest
> * The info is current <= latest

Fixed with #8078 8078

PR #8057, opened 2020-02-12
Added hack to sha256sums.txt to force users of NetData on v1.19.0-483 with a broken updater to update to latest

prologic:
See #8056

##### What I did

Commented out a bunch of irrelevant things in `./.travis/create_artifacts.sh`
so I could make sure this change will work and is correct. Then ran inside
a Docker container:

```sh
/netdata # ./.travis/create_artifacts.sh
.
.
.
--- Copy artifacts to separate directory ---
checksums:
70294f6daf017811448658fd415286773a8fead2372f39c67521fdb4a617ff61  ./latest-version.txt
b63ef24a57d013e63da383f9e89c6aedc2ad1e400048fdafb91da7d708279169  ./netdata-latest.tar.gz
```

Observed that the `netdata-vx.y.z` lines are removed.

The result is the broken logic in the NetData Updater should not be able to
find a valid version and there fail the condition and fall through.

Even though it will still error with `[: : integer expression expected`
the next condition will run as demonstrated by this simple script:

````sh
#!/bin/sh

latest_version=
current_version=
#current_version="001019000432"

if [ "${latest_version}" -gt 0 ] && [ "${current_version}" -gt 0 ] && [ "${current_version}" -ge "${current_version}" ]; then
  echo "Newest version ${current_version} <= ${latest_version} is already installed"
elif [ "${current_version}" -gt 0 ]; then
  echo "Current version: ${current_version}"
else
  echo "No version"
fi
```

### DIFF
--- a/.travis/create_artifacts.sh
+++ b/.travis/create_artifacts.sh
@@ -62,5 +62,10 @@ cd artifacts
 ln -s "${BASENAME}.tar.gz" netdata-latest.tar.gz
 ln -s "${BASENAME}.gz.run" netdata-latest.gz.run
 sha256sum -b ./* >"sha256sums.txt"
+
+# TODO: Remove this after 1 successfuly nightly or until there are no v1.19.0-432 instances left
+# XXX: See: https://github.com/netdata/netdata/issues/8056
+sed -i -e '/netdata-v/d' sha256sums.txt
+
 echo "checksums:"
 cat sha256sums.txt


Ferroin:
The question becomes how long we need to keep this for.

Correct me if I'm wrong, but this will also block updates on the fixed systems until it gets reverted, and may actually cause them to downgrade instead?

prologic:
> The question becomes how long we need to keep this for.
> 
> Correct me if I'm wrong, but this will also block updates on the fixed systems until it gets reverted, and may actually cause them to downgrade instead?

We only need this to be in place for 1 Nightly build.

I _was_ trying to find stats on how many agents we have running `v1.19.0-432` but was not successful in finding this. @cakrit Can you help with this?

prologic:
Merging for the impending Nightly in ~1h50m and will test that it forcibly upgrades broken `v1.19.0-432` installs.

prologic:
Reverting this to to Nightly build failures:

- https://travis-ci.com/netdata/netdata/jobs/286659329
- https://travis-ci.com/netdata/netdata/jobs/286659330

I don't have time/bandwidth to fully understand why; so will have to revisit a fix that forces users out of `v1.19.0-432`

PR #8058, opened 2020-02-12
collectors: apps.plugin: apps_groups: create 'dns' group.

k0ste:
I found that `unbound` placed in `email` group, but actually is a validating, recursive, caching DNS resolver. Also I was add another name server daemons in new `dns` group:
- [unbound](https://nlnetlabs.nl/projects/unbound/about/) 
- [nsd](https://www.nlnetlabs.nl/nsd/)
- [powerdns](https://www.powerdns.com)
- [knot](https://www.knot-dns.cz)
- [gdnsd](https://gdnsd.org/)
- [dnsmasq](http://www.thekelleys.org.uk/dnsmasq/doc.html)
- [yadifa](https://www.yadifa.eu/)
- [systemd-resolved](https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html)

### DIFF
--- a/collectors/apps.plugin/apps_groups.conf
+++ b/collectors/apps.plugin/apps_groups.conf
@@ -115,7 +115,7 @@ columndb: clickhouse-server*
 # -----------------------------------------------------------------------------
 # email servers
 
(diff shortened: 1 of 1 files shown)


ilyam8:
`Dnsmasq` is

> Small caching DNS proxy and DHCP/TFTP server

but i think it is ok to put it to the `dns` group 🤔

ilyam8:
thx @k0ste

PR #7894, opened 2020-01-29
ACLK agent 1

stelfrag:
##### Summary
This brings initial support for the ACLK (agent cloud link) using the libmosquitto library. The code
however is disabled, see below

##### Additional Information
- The code is disabled by default (the ACLK envirionment variable must be set to **yes**)  so it should not affect the agent operation at all.

- A new aclk folder is added with the code to support the ACLK layer functions and the libmosquitto integration

- Function calls to the ACLK layer in rrddim.c and rrdset.c are protected by a  ACLK_ENABLE check

PR #7988, opened 2020-02-06
initial MQTT over Secure Websockets support for ACLK

underhood:
#### Summary
Initial implementation of MQTT over secure websockets using custom fork of mosquitto in `netdata/mosquitto`. All currently under feature flag until ready for use.

This code allows MQTT over websockets so we can improve the whole ACLK code as whole in next PR and therefore it is not meant as finished at this point.

First #7894 has been merged this is the next step.

Closes #7880

##### Component Name
`daemon/aclk`

##### Additional Information
All is hidden under feature flag `ENABLE_ACLK` now.
To build first clone netdata/mosquitto into mosquitto subdir of netdata sourcecode.
Then run `make` in `mosquitto/lib` no need to build anything else.
Set env var `export ACLK=yes` and run `autoreconf -ivf`...

Before testing don't forget to "fake claim" the node by `echo -n "any_id" > etc/netdata/claim.d/claimed_id`

PR #8063, opened 2020-02-12
[WIP] ACLK with WSS support

### LINKED ISSUE #7880: Integrate MQTT/WSS code from SPIKE #7739 with @stelios ACLK work
The SPIKE #7880 (code in https://github.com/underhood/netdata/tree/mqtt_lws) result was successful but need to be merged into the code in a deployable way.

* The PR from @stelfrag's work on the application-layer needs to be merged first.
* The PoC code from the spike needs to be integrated in Netdata on top of this code.
* Any changes to netdata/mosquitto that result during #7740 need to be updated.
* This task needs to be coordinated with the actual deployment of the library into production.

## Expected result

When the PR for this issue is merged we should have fully functional transport later for the ACLK that wraps MQTT inside wss and allows connection to VerneMQ. The actual establishment of this link should remain gated behind a feature-flag until we are happy that the ACLK details are stable and we are ready to switch it on.

stelfrag:
##### Summary
- Initial implementation of MQTT over secure websockets using custom fork of mosquitto in netdata/mosquitto. All currently under feature flag until ready for use.

- This code allows MQTT over websockets so we can improve the whole ACLK code as whole in next PR and therefore it is not meant as finished at this point.

- Support for popcorn-ing at connect
- Add delay function for reconnect (not integrated yet for wss)
  - Uses fibonnaci exponential delay + 0-999 ms jitter (with a maximum of 1024 seconds delay)
- MQTT build in exponential backoff delay to 1024 seconds
- Integrate chart detection
- Integrate alarm reload detection
- Integrate alarm trigger (health_log)
- Detect collector addition / deletion. Triggers on_connect metadata
- JSON payload parse / tuning

- First #7894 has been merged this is the next step.

- Closes #7880

##### Component Name
daemon/aclk

Additional Information
All is hidden under feature flag ENABLE_ACLK now.

Before testing don't forget to "fake claim" the node by echo -n "any_id" > etc/netdata/claim.d/claimed_id


aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+struct lws_wss_packet_buffer {
+	unsigned char* data;
+	size_t data_size;
+	struct lws_wss_packet_buffer *next;
+};
+
+static inline struct lws_wss_packet_buffer *lws_wss_packet_buffer_new(void* data, size_t size)
+{
+	struct lws_wss_packet_buffer *new = callocz(1, sizeof(struct lws_wss_packet_buffer));
+	if(data) {
```
thiagoftsm: The function `callocz` calls the `fatal` function and finishes netdata when it cannot allocate, so in my vision this check for the return it is not necessary, please, reevaluate this check.

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+static inline void lws_wss_packet_buffer_free(struct lws_wss_packet_buffer *item)
+{
+	freez(item->data);
+	freez(item);
+}
+
+static inline void _aclk_lws_wss_read_buffer_clear(struct lws_ring *ringbuffer)
+{
+	size_t elems = lws_ring_get_count_waiting_elements(ringbuffer, NULL);
+	lws_ring_consume(ringbuffer, NULL, NULL, elems);
```
thiagoftsm: This function returns a size_t value, should we not add a test for return here?

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+	size_t data_size;
+	struct lws_wss_packet_buffer *next;
+};
+
+static inline struct lws_wss_packet_buffer *lws_wss_packet_buffer_new(void* data, size_t size)
+{
+	struct lws_wss_packet_buffer *new = callocz(1, sizeof(struct lws_wss_packet_buffer));
+	if(data) {
+		new->data = mallocz(LWS_PRE+size);
+    	memcpy(new->data+LWS_PRE, data, size);
```
thiagoftsm: There is a problem with tabulation here.

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+
+	if(!callbacks || !target_hostname)
+		return NULL;
+
+	inst = callocz(1, sizeof(struct aclk_lws_wss_engine_instance));
+
+	inst->host = target_hostname;
+	inst->port = target_port;
+
+	memset(&info, 0, sizeof(struct lws_context_creation_info));
```
thiagoftsm: The `callocz` already resets everything to zero, why are we having a call to `memset` here?

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+    info.options = LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT;
+	info.port = CONTEXT_PORT_NO_LISTEN;
+    info.protocols = protocols;
```
thiagoftsm: Please, fix the tabulation here.

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+	info.user = inst;
+	
+	inst->lws_context = lws_create_context(&info);
```
thiagoftsm: Please, add a condition here to verify whether the context was allocated correctly.

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+failure_cleanup:
+	lws_context_destroy(inst->lws_context);
+	free(inst);
+	return NULL;
+}
+
+void aclk_lws_wss_client_destroy(struct aclk_lws_wss_engine_instance* inst) {
+	lws_context_destroy(inst->lws_context);
+	inst->lws_context = NULL;
+	inst->lws_wsi = NULL;
```
thiagoftsm: You set a `NULL` for `lws_wsi`, but there was a call to destroy only `lws_context`, should we not call a function before to set it to `NULL`?

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+
+	aclk_lws_wss_clear_io_buffers(inst);
+
+#ifdef ACLK_LWS_MOSQUITTO_IO_CALLS_MULTITHREADED
```
thiagoftsm: We do not have this `ifdef` at line `217` when the mutex is used are we missing the `ifdef` there? Or are we adding a wrong `fidef` here?

thiagoftsm: We have more situations like this in the code.

aclk/aclk_lws_wss_client.c
underhood: AFAIK it should close all. I think this from following documentation of the function:

lws_context_destroy(inst->lws_context);
context | Websocket context  This function closes any active connections and then frees the  context.  After calling this, any further use of the context is  undefined.

aclk/aclk_lws_wss_client.c
underhood: the check is for data and callocz is for new

aclk/aclk_lws_wss_client.c
underhood: `aclk_lws_mutex_lock` and `_unlock` if defined like this:
```C
#ifdef ACLK_LWS_MOSQUITTO_IO_CALLS_MULTITHREADED
	#define aclk_lws_mutex_init(x) netdata_mutex_init(x)
	#define aclk_lws_mutex_lock(x) netdata_mutex_lock(x)
	#define aclk_lws_mutex_unlock(x) netdata_mutex_unlock(x)
#else
	#define aclk_lws_mutex_init(x)
	#define aclk_lws_mutex_lock(x)
	#define aclk_lws_mutex_unlock(x)
#endif
```

so in other case ifdef is not necessary

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+#define UNUSED(x) (void)(x)
+
+static int aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reason, void *user, void *in, size_t len);
+
+struct aclk_lws_wss_perconnect_data {
+    int todo;
+};
+
+//TODO config??
+int aclk_lws_wss_use_ssl = 1;
```
amoss: I don't think we have a use-case that does not use ssl.

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+	i.context = inst->lws_context;
+	i.port = inst->port;
+	i.address = inst->host;
+	i.path = "/mqtt";
+	i.host = inst->host;
+	i.protocol = "mqtt";
+	if(aclk_lws_wss_use_ssl)
+//TODO!!!!!! REMOVE ME 
+#define ACLK_SSL_ALLOW_SELF_SIGNED 1
+#ifdef ACLK_SSL_ALLOW_SELF_SIGNED
```
amoss: We could replace this with `#ifdef MAINTAINER_MODE` and it will still be useful for internal testing. I notice that maintainer mode doesn't actually do anything in the `configure.ac` though so you would need to add a def inside the test block.

aclk/aclk_lws_wss_client.c
underhood: how does the callocz clear info?

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+	}
+	tail->next = item;
+}
+
+static inline struct lws_wss_packet_buffer *lws_wss_packet_buffer_pop(struct lws_wss_packet_buffer **list)
+{
+	struct lws_wss_packet_buffer *ret = *list;
+	if(ret != NULL)
+		*list = ret->next;
+	//ret->next = NULL;
```
amoss: Why is this line commented - if it is unnecessary then please remove. If unlinking from the list is important then please uncomment.

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+    	memcpy(new->data+LWS_PRE, data, size);
+		new->data_size = size;
+	}
+	return new;
+}
+
+static inline void lws_wss_packet_buffer_push(struct lws_wss_packet_buffer **list, struct lws_wss_packet_buffer *item)
+{
+	struct lws_wss_packet_buffer *tail = *list;
+	if(!*list) {
```
amoss: What about the case that list==NULL?

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+	struct lws_wss_packet_buffer *new = callocz(1, sizeof(struct lws_wss_packet_buffer));
+	if(data) {
+		new->data = mallocz(LWS_PRE+size);
+    	memcpy(new->data+LWS_PRE, data, size);
+		new->data_size = size;
+	}
+	return new;
+}
+
+static inline void lws_wss_packet_buffer_push(struct lws_wss_packet_buffer **list, struct lws_wss_packet_buffer *item)
```
amoss: Is this is an append rather than a push?

aclk/aclk_lws_wss_client.c
underhood: responsibility of caller in this case

aclk/aclk_lws_wss_client.c
underhood: Currently not needed but I left it there as a reminder if I want to use it in different way in future.

aclk/aclk_lws_wss_client.c
underhood: i removed it in PR #7988 as that is where it belongs

aclk/aclk_lws_wss_client.c
underhood: i removed it in PR #7988 as that is where it belongs

aclk/aclk_lws_wss_client.c
underhood: Also info for reviewer -> this has not been forgotten it's just not needed

aclk/aclk_lws_wss_client.c
underhood: thank you fixed in PR #7988 where this code belong

aclk/aclk_lws_wss_client.c
underhood: Fixed in PR #7988 where this code belong. After it merged it will be fixed here as well

aclk/aclk_lws_wss_client.c
underhood: yes do you want me to rename to append or push_back? Ideal of course is if i redo this to avoid dynamic memory allocations. So it might get removed completely in MQTT/WSS refinement.

aclk/aclk_lws_wss_client.c
underhood: fixed in #7988

aclk/aclk_lws_wss_client.c
underhood: fixed in #7988

aclk/aclk_lws_wss_client.c
thiagoftsm: Thank you!

aclk/aclk_lws_wss_client.c
thiagoftsm: I am sorry for this, thank you!

aclk/aclk_lws_wss_client.c
thiagoftsm: `callocz` does a call to `calloc` ant test returns, you can find it inside `libnetdata/libnetdata.c`. To clean it we have the `freez` that is a call for `free`.

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
+#include "aclk_lws_wss_client.h"
+
+#include "libnetdata/libnetdata.h"
+
+#define UNUSED(x) (void)(x)
```
thiagoftsm: Please, move this definition to a global file, my suggestion `libnetdata/libnetdata.h`, because it is very useful!
Thank you to bring it and congratulations! :)

aclk/aclk_lws_wss_client.c
```diff
@@ -0,0 +1,340 @@
...
+		retval = -1;
+		inst->upstream_reconnect_request = 0;
+	}
+
+	switch (reason) {
+	case LWS_CALLBACK_CLIENT_WRITEABLE:
+		aclk_lws_mutex_lock(&inst->write_buf_mutex);
+		data = lws_wss_packet_buffer_pop(&inst->write_buffer_head);
+		if(likely(data)) {
+			lws_write(wsi, data->data + LWS_PRE, data->data_size, LWS_WRITE_BINARY);
```
thiagoftsm: This function returns `-1` on error, should we not have a work on the result to avoid clean the buffer before to be sure the data was sent?

aclk/agent_cloud_link.c
```diff
@@ -129,6 +161,31 @@ struct aclk_query_queue {
...
+        return 0;
+    }
+
+    if (fail == 0) {
+        srandom(time(NULL));
+        fail++;
+        return 0;
+    }
+
+    unsigned long int delay = (unsigned long int)((powl(ACLK_DELAY_SEED, fail)) * 1000 + (random() % 1618));
```
thiagoftsm: Is there any possibility of `fail` one day to become a number different of [0,1] where this interval only have integer numbers? The actual code says no, but it is possible to be changed in the future. Case the answer is `no`, please, remove completely the call to `powl`, because this function will only return either `ACLK_DELAY_SEED` or `1`, but these values can be received with a simple condition, instead to use a function that has a high cost to process(Probably it has a Taylor's series truncation or an algorithm that optimizes, but cannot be faster than an `if`).

aclk/agent_cloud_link.c
```diff
@@ -169,20 +233,26 @@ struct aclk_query *aclk_query_find_position(time_t time_to_run)
 }
 
 // Need to have a lock before calling this
-struct aclk_query *aclk_query_find(char *topic, char *data, char *msg_id, char *query)
+struct aclk_query *aclk_query_find(char *topic, char *data, char *msg_id, char *query, ACLK_CMD cmd, struct aclk_query **last_query)
 {
-    struct aclk_query *tmp_query;
+    struct aclk_query *tmp_query, *prev_query;
+    (void) cmd;
```
thiagoftsm: Why are you not using `UNUSED` for cmd?

aclk/agent_cloud_link.c
```diff
@@ -192,31 +262,53 @@ struct aclk_query *aclk_query_find(char *topic, char *data, char *msg_id, char *
...
+        return 0;
+    }
+
     run_after = now_realtime_sec() + run_after;
 
     QUERY_LOCK;
-    tmp_query = aclk_query_find(topic, data, msg_id, query);
+    struct aclk_query *last_query = NULL;
+
+    //last_query = NULL;
```
thiagoftsm: Considering that `last_query` was set to NULL on 282, I think we can remove 284.

aclk/aclk_lws_wss_client.c
amoss: We should not have commented out code in production - you should either put it back in or remove it.

aclk/aclk_lws_wss_client.c
underhood: resolved in #7988

aclk/aclk_lws_wss_client.c
underhood: renamed to `lws_wss_packet_budder_append` in #7988

aclk/aclk_lws_wss_client.c
underhood: as per docu LWS will buffer reminder of data if it cannot be send in one piece and suppress the `LWS_CALLBACK_CLIENT_WRITEABLE` until the message sending is finished. I will add handling for this anyway in the refinement issue just to be sure.

aclk/aclk_lws_wss_client.c
underhood: I will remove it... i considered this comment to be useful in future. We have other commented out code merged already though.

aclk/aclk_lws_wss_client.c
underhood: I think we have some misunderstanding. I see what callocz do. But this code use `callocz` to create `inst` (on heap). But `memset` is used to clear `info` (on stack). Or you mean something else and I am confused (possibility of course).

thiagoftsm: You are correct, I did not see the difference between `info` and  `inst` while I was reading your code, I appoligize.

aclk/agent_cloud_link.c
stelfrag: This is a bug -- there is a fail++ missing.. Will be fixed and be available in PR 8037

aclk/agent_cloud_link.c
stelfrag: Will do that

aclk/agent_cloud_link.h
```diff
@@ -28,9 +32,10 @@
 
 #define ACLK_MAX_TOPIC  255
 
-#define ACLK_RECONNECT_DELAY 1          // reconnect delay -- with backoff stragegy fow now
-#define ACLK_MAX_RECONNECT_DELAY 120
-#define ACLK_VERSION "1"
+#define ACLK_RECONNECT_DELAY 1 // reconnect delay -- with backoff stragegy fow now
+#define ACLK_MAX_RECONNECT_DELAY ACLK_MAX_BACKOFF_DELAY / 1000
```
amoss: This should be `(ACLK_MAX_BACKOFF_DELAY / 1000)` to guard against misuse.

aclk/agent_cloud_link.h
```diff
@@ -12,14 +12,18 @@
...
 
 #define ACLK_MSG_TYPE_CHART "chart"
+#define ACLK_CHART_TOPIC "chart"
 #define ACLK_ALARMS_TOPIC "alarms"
 #define ACLK_METADATA_TOPIC "meta"
 #define ACLK_COMMAND_TOPIC "cmd"
 #define ACLK_TOPIC_STRUCTURE "/agent/%s"
 
-#define ACLK_STARTUP_WAIT 30             // Seconds to wait before establishing initialization process
+#define ACLK_DELAY_SEED 1.61803        // Exponential fibonnaci backoff
```
amoss: It says in the spec that we will use powers of two.

aclk/agent_cloud_link.c
```diff
@@ -129,6 +161,31 @@ struct aclk_query_queue {
...
 } aclk_queue = { .aclk_query_head = NULL, .aclk_query_tail = NULL, .count = 0 };
 
+/*
+ * After a connection failure -- delay in milliseconds
+ * When a connection is established, the delay function
+ * should be called with mode 0 to reset the fail count
+ */
+unsigned long int aclk_delay(int mode)
+{
+    static int fail = 0;
```
amoss: If fail is intended as a counter then I don't see where timeout failure causes it to be incremented?

aclk/agent_cloud_link.c
amoss: This seems to be very over-complicated - will it be significantly better than `1<<fail` ?

aclk/agent_cloud_link.c
```diff
@@ -129,6 +161,31 @@ struct aclk_query_queue {
...
+
+    if (fail == 0) {
+        srandom(time(NULL));
+        fail++;
+        return 0;
+    }
+
+    unsigned long int delay = (unsigned long int)((powl(ACLK_DELAY_SEED, fail)) * 1000 + (random() % 1618));
+
+    return MIN(delay, ACLK_MAX_BACKOFF_DELAY);
```
amoss: It would be better to make the failure count saturate so that it sticks at the highest value in the sequence rather than cropping to another value.

aclk/agent_cloud_link.c
thiagoftsm: All right, thank you for the explanation!  :)

aclk/agent_cloud_link.c
stelfrag: Will refine

aclk/agent_cloud_link.h
stelfrag: Will change to power of 2 to much the doc

aclk/agent_cloud_link.c
stelfrag: 1 << fail will work (throwing away fibonacci) :+1:  :smile:

thiagoftsm: But, multiply by 2 is completely different of to use the `powl`. 
This is a complete different algorithm for me, I did not enter in the details about the algorithm, because I imagined that there was a motive to use `powl`, why are changing from the fastest function possible to multiply by 2?

amoss: How is `powl()` faster than bit-shifting for powers of two?

stelfrag: It will change to power of two. The delay (for a maximum of 1024) will produce values like

```
0.        0ms
1.     2756ms
2.     4417ms
3.     8346ms
4.    16144ms
5.    32771ms
6.    64401ms
7.   128789ms
8.   256981ms
9.   512283ms
10.  1024ms
```

stelfrag: The fibonacci would be
```
0.        0ms
1.     1680ms
2.     2681ms
3.     4960ms
4.     6979ms
5.    11719ms
6.    17963ms
7.    29921ms
8.    46427ms
9.    76081ms
10.   122972ms
11.   198152ms
12.   321754ms
13.   520540ms
14.   842604ms
15.  1024000ms
```

thiagoftsm: > How is `powl()` faster than bit-shifting for powers of two?

It is impossible to be faster.

thiagoftsm: I only suppose that there was a good motive to select Fibonacci instead power of two, but after these answers, it looks like that what really matter is to have a constant to do math.
Finally the functions 2^x and 1.618^x will deliver different outputs, this was the motive I asked more information, because the input is the same, the output never will be. ;)
I do not have any objection to change, I only wanted to understand the motive it was selected 1.618 .

Issue #8064, opened 2020-02-12
Dirty build of Netdata


vlvkobal:
##### Bug report summary
Since #8025 after I've built Netdata I get
```
[vlad@arch-esxi netdata]$ git status
warning: could not open directory 'netdata-mosquitto-9MAOO5/': Permission denied
```
and
```
[vlad@arch-esxi netdata]$ sudo git status
On branch master
Your branch is ahead of 'origin/master' by 12 commits.
  (use "git push" to publish your local commits)

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        externaldeps/
        netdata-mosquitto-9MAOO5/
```

##### Component Name
packaging

##### Steps To Reproduce
run the `netdata-installer.sh` script.

##### Expected behavior
`git status` provides a clean output.

ilyam8:
btw, i see mosquitto dirs are not deleted, i have a new dir after very install

```console
test@DEB10-1:~/netdata$ sudo du -h -d 1 *mosquitto*
8.6M	netdata-mosquitto-4Zf5x5/mosquitto-v.1.6.8_Netdata-2
11M	netdata-mosquitto-4Zf5x5
8.1M	netdata-mosquitto-Uc6seR/mosquitto-v.1.6.8_Netdata-2
10M	netdata-mosquitto-Uc6seR
8.6M	netdata-mosquitto-uYgZLO/mosquitto-v.1.6.8_Netdata-2
11M	netdata-mosquitto-uYgZLO
8.6M	netdata-mosquitto-Ww6XlJ/mosquitto-v.1.6.8_Netdata-2
11M	netdata-mosquitto-Ww6XlJ
`

PR #8066, opened 2020-02-12
Make temporary directories relative to $TEMPDIR.

Ferroin:
##### Summary

This corrects the invocations of `mktemp` so that they produce temporary directories in `$TEMPDIR` instead of the current directory. We need to use `-t` for this for compatibility with busybox `mktemp`.

##### Component Name

area/packaging

##### Additional Information

Fixes: #8064

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -192,3 +192,6 @@ docs/generator/mkdocs.yml
 
 #CLion files
 netdata.cbp
+
+# External dependencies
+externaldeps/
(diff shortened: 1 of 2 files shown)


ilyam8:
what about `externaldeps` dir? is it expected to have it not deleted after install?

Ferroin:
> what about `externaldeps` dir? is it expected to have it not deleted after install?

It should be in `.gitignore`, so it shouldn't matter. It also doesn't grow bigger on each install either.

Issue #7771, opened 2020-01-17
Shared libraries for eBPF collectors


thiagoftsm:
##### Feature idea summary
Netdata is bringing eBPF collectors to monitor parts of the Linux Kernel, but we do not want to bring big dependencies for our users. To solve this problem we decided to create a shared library that will be dynamic loaded from the collectors. This shared library will be downloaded during the installation process.

###### Environment

To build our shared library we must have  the following environment:
- A kernel compiled with the following options:
```
CONFIG_KPROBES=y
CONFIG_KPROBES_ON_FTRACE=y
CONFIG_HAVE_KPROBES=y
CONFIG_HAVE_KPROBES_ON_FTRACE=y
CONFIG_KPROBE_EVENTS=y
```
- Libelf headers
- llvm/clang , because GCC cannot compile eBPF codes.
- Kernel headers
- Libbpf ( https://github.com/libbpf/libbpf )

About the last option for the environment, it is necessary to give important information. We are using the latest version of eBPF when the kernel is newest than 4.14.X, but for version before 1.15, we changed the libbpf to also give support for these kernels, so we will have 2 libbpf versions been compiled depending of the kernel.

In theory we do not need to compile the codes for all kernel versions, I have worked only with `long term` kernels, and when a distribution decides to work with a different version, I always used the immediate `long term` before the version running on the distribution, unless there is a big change for that specific version, that will set another kernel to compile.

###### Distribution

About the Linux distribution, I am developing using Slackware current, but I also could create without big effort the environment on Fedora and Ubuntu. 

###### Test
The whole structure to build the shared library also has codes used to test the eBPF programs. To run these tests, it is necessary to mount the following file systems:

```
mount -t debugfs nodev /sys/kernel/debug
mount -t tracefs nodev /sys/kernel/tracing
```

##### Expected behavior
We expect to have conditions to change `netdata-installer.sh` and other scripts to do download of the shared libraries and eBPF programs that will be used with Netdata.

prologic:
Removing the "Needs Triage" label as I'm fairly confident I understand what needs to be done here. @Ferroin I'd like to get on to this next after we've mostly happy (_doesn't have to be perfect_) with the binary packaging / multiarch stuff we're currently doing.

Ferroin:
[...]

PR #7979, opened 2020-02-05
eBPF process plugin

thiagoftsm:
##### Summary
Implements #7884 

##### Component Name
Collector

##### Additional Information
This collector is disabled by default, to enable it is necessary to add the following line inside yor netdata.conf:

```
[plugins]
    ebpf_process = yes
```

This collector needs additional files that will be built in our CI/CD, but while we are finishing this, I am bringing files that were built in the repository for you test.

[4_14_x.zip](https://github.com/netdata/netdata/files/4196633/4_14_x.zip)

[4_19_x.zip](https://github.com/netdata/netdata/files/4196437/4_19_x.zip)

[5_x.zip](https://github.com/netdata/netdata/files/4196439/5_x.zip)

The file 5_x.zip has the libraries and eBPF program necessaries to compile on all 5.x versions, but I did not test on 5.6-rc.
4_19_x.zip allows tests on 4.15, 4.16, 4.17, 4.18, 4.19 and 4.20.
Finally 4_14_x.zip allow tests on 4.11, 4.12, 4.13 and 4.14

Before you start the tests with the collector it is necessary to install manually the files inside each zip file and it will also be necessary to create a symbolic link for the shared library, everything can be done using the following commands:

- unzip 5_x.zip
- cp usr/lib64/libbpf_kernel.so /usr/lib64/libbpf_kernel.so
- ln -s  /usr/lib64/libbpf_kernel.so /usr/lib64/libbpf_kernel.so.0
- cp *ebpf* /usr/libexec/netdata/plugins.d/

An important thing about the 2nd and 3rd commands in the previous list, case your Linux distribution does not have `/usr/lib64/` in the library path list, you need to move `libbpf_kernel.so` for a directory where the libraries are, or to change `LD_LIBRARY_PATH` to add the path for the library. There are Linux distributions that will require that you run `ldconfig` after the copy.

Finally there are points that I think we need to agree, and I would like to hear your opinion before:

1 - What charts should have alarms? What are the values for them?
2 - Are the texts explaining the charts good enough for you understand what it means? 
3 - Netdata has collectors that finishes themselves after 14400 seconds, this collector does not close normally, it has steps to clean everything before it closes totally, and thanks this the parent will close faster than the child that cleans everything. Can we keep it running forever? Or it is also a good idea to restart it?

PR #8075, opened 2020-02-12
Integrates the eBPF Kernel Collector with the NetData Installer

prologic:
##### Summary

In order to close out netdata/kernel-collector#11 we are going to update the
NetData Installer to:

- Determine if the User's Kernel is compatible.
- Figure out which compatible version of the `netdata_ebpf_*` package to Download and Install from the `netdata/kernel-collector` Github Releases.
- Install it.

This is the PR commits will be added to and you are welcome to add comments here if you have feedback on the approach/opinions/etc.

This will also get #7771 -- We will defer some of the other work in the EPIC
to a later time (_they are still however important to get done_).

##### Component Name

- area/ci
- area/packaging

##### Test Plan

- [x] Publish a new Release of `netdata/kernel-collector`
- [x] Install a fresh NetData Agent with options to include/install the eBPF collector.
- [x] Ensure we are collecting eBPF metrics.

##### Additional Information


netdata-installer.sh
```diff
@@ -1022,6 +1024,72 @@ install_go() {
 
 install_go
 
+should_install_ebpf() {
+  if [ "${NETDATA_ENABLE_EBPF:=0}" -ne 1 ]; then
```
thiagoftsm: Should we also add a condition to verify the kernel version here? Case the kernel is previous `4.11`, we won't have conditions, independent of the users say they want to use.
I saw that you added a condition in the next function to this, but the name of this function says me that we could bring the test from `install_ebpf` to here.
This is only a suggestion.

prologic: Sorry yes you are right; it was late at night and I _actually_ meant to have the check here :) I will move it!

netdata-installer.sh
```diff
@@ -1022,6 +1027,75 @@ install_go() {
...
+    run_failed "ebpf not enabled. --enable-ebpf to enable"
+    return 1
+  fi
+
+  # Check Kernel Config
+  if ! get "https://raw.githubusercontent.com/netdata/kernel-collector/master/tools/check-kernel-config.sh" | bash -; then
+    run_failed "Kernel unsupported or missing required config"
+    return 0
```
ilyam8: it looks like should return `1` here

prologic: Thank you! Yes you are correct.

prologic:
**Ready for Full Review!**

prologic:
2/3 of the Test Plan complete.

prologic:
Expected Installation:

```sh
 --- eBPF Kernel Collector (opt-in) ---
 Using eBPF Kernel Package built against Linux 4.19
 --- Installing eBPF plugin ---
 Using eBPF Kernel Package built against Linux 4.19
 Detected GLIBC
 Getting latest eBPF Package URL for netdata_ebpf-4_19-glibc.tar.xz ...
 Downloading eBPF Package https://github.com/netdata/kernel-collector/releases/download/v0.0.1/netdata_ebpf-4_19-glibc.tar.xz ...
 Extracting netdata_ebpf-4_19-glibc.tar.xz ...
 Finding suitable lib directory ...
[/home/prologic/netdata]# cp -a -v /tmp/netdata-ebpf-wt9gaP/usr/lib64/libbpf_kernel.so /usr/local/lib \n'/tmp/netdata-ebpf-wt9gaP/usr/lib64/libbpf_kernel.so' -> '/usr/local/lib/libbpf_kernel.so'
 OK

[/home/prologic/netdata]# cp -a -v /tmp/netdata-ebpf-wt9gaP/libnetdata_ebpf.so /usr/libexec/netdata/plugins.d \n'/tmp/netdata-ebpf-wt9gaP/libnetdata_ebpf.so' -> '/usr/libexec/netdata/plugins.d/libnetdata_ebpf.so'
 OK

[/home/prologic/netdata]# cp -a -v /tmp/netdata-ebpf-wt9gaP/dnetdata_ebpf_process.o /usr/libexec/netdata/plugins.d \n'/tmp/netdata-ebpf-wt9gaP/dnetdata_ebpf_process.o' -> '/usr/libexec/netdata/plugins.d/dnetdata_ebpf_process.o'
 OK

[/home/prologic/netdata]# cp -a -v /tmp/netdata-ebpf-wt9gaP/pnetdata_ebpf_process.o /usr/libexec/netdata/plugins.d \n'/tmp/netdata-ebpf-wt9gaP/pnetdata_ebpf_process.o' -> '/usr/libexec/netdata/plugins.d/pnetdata_ebpf_process.o'
 OK

[/home/prologic/netdata]# cp -a -v /tmp/netdata-ebpf-wt9gaP/rnetdata_ebpf_process.o /usr/libexec/netdata/plugins.d \n'/tmp/netdata-ebpf-wt9gaP/rnetdata_ebpf_process.o' -> '/usr/libexec/netdata/plugins.d/rnetdata_ebpf_process.o'
 OK

[/home/prologic/netdata]# ln -v -f -s /usr/local/lib/libbpf_kernel.so libbpf_kernel.so.0 \n'libbpf_kernel.so.0' -> '/usr/local/lib/libbpf_kernel.so'
 OK

[/home/prologic/netdata]# ldconfig \n OK

ePBF installation all done!
```

prologic:
Screen Shot 2020-02-14 at 13 13 45

prologic:
And it all works 😀

run

stelfrag:
Installing on fedora 31 (vm) 
```
 --- Installing eBPF plugin --- 
 Using eBPF Kernel Package built against Linux 5.4
 Detected GLIBC
 Getting latest eBPF Package URL for netdata_ebpf-5_4-glibc.tar.xz ...
 Downloading eBPF Package https://github.com/netdata/kernel-collector/releases/download/v0.0.1/netdata_ebpf-5_4-glibc.tar.xz ...
 Extracting netdata_ebpf-5_4-glibc.tar.xz ...
 Finding suitable lib directory ...
 FAILED  Could not find a suitable lib directory 
```

mfundul:
I see no problems in Ubuntu 18.04, that is without having any ebpf collectors. Maybe it's missing from this PR? In any case no problems for the agent in Ubuntu 18.04.

netdata-installer.sh
```diff
@@ -254,7 +255,11 @@ while [ -n "${1}" ]; do
     "--disable-x86-sse") NETDATA_CONFIGURE_OPTIONS="${NETDATA_CONFIGURE_OPTIONS//--disable-x86-sse/} --disable-x86-sse" ;;
     "--disable-telemetry") NETDATA_DISABLE_TELEMETRY=1 ;;
     "--disable-go") NETDATA_DISABLE_GO=1 ;;
-    "--disable-cloud") NETDATA_DISABLE_LIBMOSQUITTO=1 ; NETDATA_CONFIGURE_OPTIONS="${NETDATA_CONFIGURE_OPTIONS//--disable-aclk/} --disable-aclk" ;;
+    "--enable-ebpf") NETDATA_ENABLE_EBPF=1 ;;
```
ilyam8: If i remember it right we will have several ebpf plugins. Should this flag be more specific? Or the plan is either to download all ebpf stuff (several plugins) or nothing?

mfundul:
CentOS 7 is:
```
 --- eBPF Kernel Collector (opt-in) --- 
 Using eBPF Kernel Package built against Linux 4.19
 Downloading check-kernel-config.sh ...
[/root/netdata]# chmod +x /tmp/netdata-ebpf-uAxCg5/check-kernel-config.sh \n OK   

[/root/netdata]# /tmp/netdata-ebpf-uAxCg5/check-kernel-config.sh \nYour kernel appears to be older than 4.11. This may still work in some cases, but probably won't.
 Missing Kernel Config: KPROBE_EVENTS
 FAILED   

 FAILED  Kernel unsupported or missing required config 

```

vlvkobal:
It works on the latest Arch Linux with the `linux 5.5.3.arch1-1` kernel correctly. #7979 shows eBPF charts after installing the libraries using #8075.

amoss:
It installs fine on debian buster, both in the default and pulling the 4.19 libraries when it is enabled.

netdata-installer.sh
```diff
@@ -1022,6 +1027,171 @@ install_go() {
...
+    echo >&2 " Detected GLIBC"
+    libc="glibc"
+  elif ldd --version 2>&1 | grep -q -i 'gnu libc'; then
+    echo >&2 " Detected GLIBC"
+    libc="glibc"
+  elif ldd --version 2>&1 | grep -q -i musl; then
+    echo >&2 " Detected musl"
+    libc="musl"
+  else
+    echo >&2 " ERROR: Cannot detect a valid libc on your system!"
```
Ferroin: I think this should probably be 'a supported libc' not 'a valid libc'. It's Linux, there's pretty much always going to be _some_ libc there, we just don't support anything other than musl and glibc 9and technically eglibc by virtue of supporting glibc).

thiagoftsm:
I notice now that when we are using  `netdata-installer.sh` to install on distributions with `4.19.X` the collector cannot run, because it is returning `errno 22`, on the other hand, when I do download of the file `netdata_ebpf-4_19-glibc.tar.xz` and do the installation manually, everything is working fine, so I suspect the installer is not installing the correct files on the environment.

netdata-installer.sh
```diff
@@ -1022,6 +1027,175 @@ install_go() {
...
+    return 1
+  fi
+
+  echo >&2 " Extracting ${PACKAGE_TARBALL} ..."
+  tar -xf "${tmp}/${PACKAGE_TARBALL}" -C "${tmp}"
+
+  echo >&2 " Finding suitable lib directory ..."
+  libdir=
+  libdir="$(ldconfig -v 2> /dev/null | grep ':$' | sed -e 's/://' | sort -r | grep 'usr' | head -n 1)"
+  if [ -z "${libdir}" ]; then
```
thiagoftsm: @prologic I added these lines to fix problems with Fedora.

netdata-installer.sh
```diff
@@ -1022,6 +1027,174 @@ install_go() {
...
+  fi
+
+  if [ -z "${libdir}" ]; then
+    run_failed "Could not find a suitable lib directory"
+    return 1
+  fi
+
+  run cp -a -v "${tmp}/usr/lib64/libbpf_kernel.so" "${libdir}"
+  run cp -a -v "${tmp}/libnetdata_ebpf.so" "${NETDATA_PREFIX}/usr/libexec/netdata/plugins.d"
+  run cp -a -v "${tmp}"/pnetdata_ebpf_process.o "${NETDATA_PREFIX}/usr/libexec/netdata/plugins.d"
```
thiagoftsm: @prologic I removed one line here, because we defined that the developer mode will come in future versions

netdata-installer.sh
stelfrag: Fedora 31 now builds fine

thiagoftsm: Thank you to confirm! :)

netdata-installer.sh
```diff
@@ -1022,6 +1027,174 @@ install_go() {
...
+  kver="${1}"
+
+  # XXX: Logic taken from Slack discussion in #ebpf
+  # everything that has a version <= 4.14 can use the code built to 4.14
+  # Continue the logic, everything that has a version <= 4.19 and >= 4.15 can use the code built to 4.19
+  # Finally,  everybody that is using 5.X can use what is compiled with 5.4
+
+  kpkg=
+
+  if [ "${kver}" -gt 0 ] && [ "${kver}" -ge 005000000 ]; then
```
thiagoftsm: Why are we checking here when a kernel version is greater than 0?

netdata-installer.sh
```diff
@@ -1022,6 +1027,174 @@ install_go() {
...
+    echo >&2 " Using eBPF Kernel Package built against Linux 5.4"
+    kpkg="5_4"
+  elif [ "${kver}" -gt 0 ] && [ "${kver}" -ge 004015000 ] && [ "${kver}" -le 004020017 ]; then
```
thiagoftsm: I fixed the comparison here, the previous was not allowing the next `elif` to be compared.

thiagoftsm: I could not understand the motive we also comparing a kernel greater than 0, I think we can test only the limits [min, max]

netdata-installer.sh
prologic: Nice

netdata-installer.sh
prologic: Ahh I see :)

netdata-installer.sh
prologic: Its just a guard

netdata-installer.sh
prologic: You are probably right :)

netdata-installer.sh
prologic: Not entirely sure to be honest. I think @thiagoftsm has more knowledge around this. My gut instinct tells me we ship all ebpf programs.

netdata-installer.sh
prologic: Fine :) Bust some Liniux(es) don't ship a libc a all 🤣

netdata-installer.sh
prologic: Awesome :)

netdata-installer.sh
prologic: No problems!

netdata-installer.sh
prologic: Just a guard; but I think we can remove it.

netdata-installer.sh
prologic: Cool!

netdata-installer.sh
thiagoftsm: You are right @ilyam8 , and I am using  on 7979 `NETDATA_ENABLE_EBPF' because the libraries to enable one is the exact the same to enable others.

netdata-installer.sh
```diff
@@ -1022,6 +1027,174 @@ install_go() {
...
+    echo >&2 " Using eBPF Kernel Package built against Linux 4.14"
+    kpkg="4_14"
+  else
+    echo >&2 " ERROR: Cannot detect a supported kernel on your system!"
+    return 1
+  fi
+
+  echo "${kpkg}"
+  return 0
+}
```
Ferroin: The more I think a bout it, the more I think we should probably be installing the code for all the kernel variants we support and selecting at runtime which to load. It's not unusual for people to upgrade kernels on systems, and the current arrangement will result in a kenrel upgrade breaking the eBPF collectors until a user re-installs Netdata.

prologic: As stated in #ebpf and #core-team and our SRE Sync today; I agree. We have a few different approaches we can take here:

- Install all variants and put the right files in place depending on the current Kernel and libc.
- Teach the agent how to load the right variants itself (_C level changes?_)
- Install all variants and put the right files in place depending on the current Kernel and libc and teach the agent how to call out to a shell script that "redoes this" if/when the Kernel changes.
- A variant of the previous but download on demand rather than install all variants.

I'm in favour of the last two personally and I'm indifferent on "install all variants" or "download on demand". I can see both approaches working equally as well and @thiagoftsm and I have already talked about the possibility of "building on demand" too remotely.

Lets discuss further after the `v1.20.0` release in #ebpf and/or a quick meeting and draft up a feature request / issue.

thiagoftsm:
Thank you everybody for the reviews, I am already working in the problems reported to create a bug fix for them!

netdata-installer.sh
thiagoftsm: I prefer "A variant of the previous but download on demand rather than install all variants.", considering that you liked it, let us move in front with it. :)

prologic: Sounds fine to me. @thiagoftsm Do you mind creating an issue around this? Its super unclear ot me when / how we'll prioritise it; but we have a list of future enhancements we want to make too including full e2e VM testing so we're more confident what we're shipping will work for our users :) -- I'd bucket this issue/feature along with that too somewhat.

netdata-installer.sh
```diff
@@ -1022,6 +1027,174 @@ install_go() {
...
+
+should_install_ebpf() {
+  if [ "${NETDATA_ENABLE_EBPF:=0}" -ne 1 ]; then
+    run_failed "ebpf not enabled. --enable-ebpf to enable"
```
ilyam8: should this be failed?

It looks like an error during install which is not

```
>  --- eBPF Kernel Collector (opt-in) ---
 FAILED  ebpf not enabled. --enable-ebpf to enable
```

prologic: Is there a better function to use here? I couldn't find one; so I'll probably just change this to `echo >&2 "..."`

prologic: https://github.com/netdata/netdata/compare/master...prologic:ebpf_installer_tweaks?expand=1

netdata-installer.sh
thiagoftsm: All right, I will create in few minutes!

prologic: Link it here when done and mark this thread as "resolved" :)

prologic: Created => https://github.com/netdata/netdata/issues/8118

thiagoftsm: Thank you!

PR #8076, opened 2020-02-13
Revert "Added hack to sha256sums.txt to force users of NetData on v1.19.0-483 with a broken updater to update to latest"

prologic:
Reverts netdata/netdata#8057


prologic:
See [this comment](https://github.com/netdata/netdata/pull/8057#issuecomment-585542828)

PR #8100, opened 2020-02-15
Remove Fedora 29 from CI and packaging.

Ferroin:
##### Summary

Official upstream support for Fedora 29 ended on 2019-11-26, almost 3 months ago.

##### Component Name

area/ci
area/packaging

##### Additional Information

This is intended to be merged after the upcoming release of 1.20 so that we have one last full release on Fedora 29.


prologic:
I've seen you put up a few of these PR(s)... And I'll repeat something I said in our most recent meetup; I _feel_ that we should not disable/remove support for older Systems/Distro(s) _until_ it becomes far too much effort/work to support. For this specific case support only ended a few months ago; I _feel_ that we could still continue to support, test and ship this for the foreseeable future; I mean until it gets to a point where things have changed/diverged so much that the effort is too high to support.

Thoughts?

Ferroin:
I would much prefer to be proactive about this myself, because it lets us be predictable in terms of what we support, as well as letting us focus more on supporting a wider variety of platforms (because we don't have to pay attention to old platforms as much).

Upstream is not supporting these versions _AT ALL_. These aren't cases of the distro maintainers saying "we aren't providing ny more updates or security fixes for this", it's a case of "we're not providing _any_ updates for this at all, and if you come to us asking for help we'll tell you to upgrade and then probably ignore you if you complain about it".

In this particular case, upstream doesn't even support upgrades from this version to the latest anymore.

Note that I'm not saying here that we should come out and say "Hey, we don't support this particular distribution at this version at all.", I just feel that we need to not be wasting resources on stuff that few if any people will actually be using.  For pretty much all the platforms we build packages for except CentOS, outside of a shot window just after a new release, most users are on either a LTS release or the most recent release. By simply covering things we can support that aren't EOL, we're already serving about 95% or more of the users who wouldn't be using Docker anyway.

In all honesty, I'd like to just limit our support to distros/platforms that are still supported by upstream, and keep supporting things until our next release at least a month after they go EOL upstream.

prologic:
I agree with you; but it had to be said. Let it be then for the record that this is the stance we're taking.

As such I _believe_ we should document this in our docs repo so that we're all clear and agree. New engineers that join our team (_at some point_) will have similar if not the same thoughts and questions and we can easily point them to our docs.

Are you okay to add a PR to docs to cover our decision on supporting OS/Distro/Systems for the Agent? What we support, when we stop supporting, etc.

Ferroin:
@prologic Yes, I'll be more than happy to do one. Once that's official, I'll also look at doing a PR to update the user documentation so that it's publicly documented that we take this approach.

PR #8125, opened 2020-02-18
Update eBPF docs with better install/enable instructions

joelhans:
##### Summary

After discussing the eBPF installation and enabling process with Thiago, I decided to go into the document to clarify the process of getting this collector working. We didn't have any details about the required `--enable-ebpf` option, which would have probably stumped a lot of our ambitious users.

I'm not sure if this qualifies as a bug fix, but if not, it should be merged immediately after the release goes out so users will be able to find the correct information right away.

##### Component Name

docs
collectors/

I will look at the Netlify preview once it finishes building.

### DIFF
--- a/collectors/ebpf_process.plugin/README.md
+++ b/collectors/ebpf_process.plugin/README.md
@@ -59,8 +59,7 @@ commands. You can also configure your system's `/etc/fstab` configuration to mou
 
 eBPF collection is only enabled if you install Netdata with the `--enable-ebpf` option. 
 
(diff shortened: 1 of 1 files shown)


thiagoftsm:
Thank you @joelhans , I think it is more clear for the users the steps to install and enable the plugin.

joelhans:
There's a link error with the Chinese translation that I'm hoping to fix with netdata/localization#27. That should fix the Netlify build errors.

Issue #8077, opened 2020-02-13
Install local version of libwebsockets v3 if unavailable.


amoss:
##### Feature idea summary

If the system installing Netdata does not have libwebsockets v3+ then download and build it.
This is complicated by the library detection routines being in autotools which is called after the installer.

##### Expected behavior

* The netdata-installer should try to download the libwebsockets code, put it into a /tmp build directory and build a static version of the library. 
* This should then be copied into externaldeps/libwebsockets along with any header files.
* The configure script should check if libwebsockets v3 is in the base system, by checking if it can resolve one of the changed names as a symbol.
* If it cannot then it add the static version of the library to the build target if-and-only-if:
  * the static build of libmosquitto succeeded
  * the --disable-cloud switch was not passed to the installer (so it did not set the appropriate --disable flags for configure).
* If the build of libwebsockets fails then we splash an error message to the console (as discussed on product/282) and disable the ACLK.

Investigate whether there wide-enough support for rpath (across the linux distros, the BSDs and MacOS) to allow an alternative, in which case:

* Build v3 of libwebsockets as shared library.
* Install into an appropriate path under the installation prefix,  e.g. some variation on /usr/lib, /usr/libexec or /var/lib as appropriate?
* Set the rpath in the build.
* Change the handling of libmosquitto to use the same approach.

##### Necessary test-cases

* Feature flag disabled.
* Systems with libwebsockets v2 + libmosquitto available.
* Systems with libwebsockets v3 + libmosquitto available.
* Systems with libwebsockets v2 + libmosquitto unavailable.
* Systems with libwebsockets v3 + libmosquitto unavailable.
[Simulate the last two by artificially breaking the build].

Ferroin:
Those first two steps should be happening _after_ we figure out if we need libsebsockets or not, not before. We're wasting resources otherwise, and may be costing our users money as a result. The check should be trivial on most systems, as we can just use `pkg-config --modversion libwebsockets` to get an exact version number.

Ferroin:
I'm not entirely sure using RPATH is worth the effort for this. It's at least partly dependent on both the compiler and the libc being used even on Linux, and using dynamic loading for libraries we only link into one binary doesn't really get us anything. The whole point of dynamic linking is to save disk space and reduce the runtime memory footprint of binaries by dedpuplicating code that is being used by multiple binaries, which doesn't matter for us if we're only linking into one binary (or even only two or three).

[...]

PR #8144, opened 2020-02-20
Added libwebsockets bundling code to netdata-installer.sh

Ferroin:
##### Summary

This adds basic code to prepare and bundle libwebsockets 3 on systems which do not have it.

It also adds cmake and libwebsockets to the build-time dependencies pulled in by the `install-required-packages.sh` script.

##### Component Name

area/packaging

Verified that the required packages are installed correctly on Linux platforms we support, and that we correctly build and bundle libwebsockets if we need it.

##### Additional Information

Fixes: #8077


Ferroin:
To anyone who goes to review this, the lifecycle checks are failing due to an internal git error that I can't reproduce locally. Based on my own testing though, the code does indeed appear to be working correctly.

underhood:
Hi in my case it fails at checksum verify. Seems like file `./packaging/websockets.checksums` is missing?

underhood:
after i created the checksum file with the checksum of the library manually the verification suceeds but then I get another error:
```bash
[/home/timo/projects/netdata]$ tar -xf /tmp/netdata-libwebsockets-g4eN7O/v3.2.2.tar.gz -C /tmp/netdata-libwebsockets-g4eN7O \n OK   

CMake Error: The source directory "/tmp/netdata-libwebsockets-g4eN7O/websockets-v3.2.2" does not exist.
Specify --help for usage, or press the help button on the CMake GUI.
make: *** /tmp/netdata-libwebsockets-g4eN7O/websockets-v3.2.2: No such file or directory.  Stop.
 FAILED  Failed to build libwebsockets. The install process will continue, but you may not be able to connect this node to Netdata Cloud. 
```
I think the folder is wrong because this is what i see when i do ls:
``` bash
 timo@timowss  /tmp/netdata-libwebsockets-g4eN7O  ls -la
total 9644
drwx------.  3 timo timo     100 feb 21 18:29 .
drwxrwxrwt. 35 root root     880 feb 21 18:30 ..
drwxrwxr-x. 15 timo timo     540 jan 17 13:38 libwebsockets-3.2.2
-rw-rw-r--.  1 timo timo      80 feb 21 18:29 sha256sums.txt
-rw-rw-r--.  1 timo timo 9867911 feb 21 18:29 v3.2.2.tar.gz
```

netdata-installer.sh
```diff
@@ -469,32 +465,75 @@ bundle_libmosquitto() {
...
+  LIBWEBSOCKETS_PACKAGE_BASENAME="${LIBWEBSOCKETS_PACKAGE_VERSION}.tar.gz"
+
+  if fetch_and_verify "websockets" \
+                      "https://github.com/warmcat/libwebsockets/archive/${LIBWEBSOCKETS_PACKAGE_BASENAME}" \
+                      "${LIBWEBSOCKETS_PACKAGE_BASENAME}" \
+                      "${tmp}" \
+                      "${NETDATA_LOCAL_TARBALL_OVERRIDE_LIBWEBSOCKETS}"
+  then
+    if run tar -xf "${tmp}/${LIBWEBSOCKETS_PACKAGE_BASENAME}" -C "${tmp}" && \
+       build_libwebsockets "${tmp}/websockets-${LIBWEBSOCKETS_PACKAGE_VERSION}" && \
```
underhood: in my system it unpacks to `${tmp}/libwebsockets-` not websockets.

underhood: On my system it unpacks to `${tmp}/libwebsockets-3.2.2` not `${tmp}/libwebsockets-v3.2.2`

underhood:
after fixing all of the above the cmake complains i have to add cd to:
```
build_libwebsockets() {
  cd "${1}"
  cmake "${1}"
  make -C "${1}"
}
```
then it builds

netdata-installer.sh
```diff
@@ -259,6 +259,7 @@ while [ -n "${1}" ]; do
     "--enable-ebpf") NETDATA_ENABLE_EBPF=1 ;;
     "--disable-cloud")
       NETDATA_DISABLE_LIBMOSQUITTO=1
+      NETDATA_SKIP_LIBWEBSOCKETS=1
```
prologic: Nit: Stick to `NEDTAT_DISABLE_XXX`? for consistency?

prologic:
@Ferroin I thought we had decided to patch the `Makefile`(s) for `libmosquitto` to avoid having to pull in `cmake`? Did we change our minds?

netdata-installer.sh
Ferroin: It's not disabling anything though, just skipping trying to bundle libwebsockets.

Ferroin:
> @Ferroin I thought we had decided to patch the `Makefile`(s) for `libmosquitto` to avoid having to pull in `cmake`? Did we change our minds?

`cmake` is a requirement for libwebsockets, so we either need it as a dependency now, _or_ we fork libwebsockets and maintain a probably ridiculously large set of modification so that it doesn't need `cmake`. I'm favoring the first option since it's far less work for everyone and it lets us handle libmosquitto better in some cases.

prologic:
Fair enough; I just missed the discussion :) All good!

prologic:
Don't forget to fix the conflicts ;)

underhood:
please take a look at @shortpatti change requests on https://github.com/netdata/netdata/pull/8157 as the code originates here. I will then reaproove and rebase my PR

underhood:
Hi There is new requirement that was discovered today and we have to make sure LWS_WITH_SOCKS5 is enabled and we build LWS with SOCKS5 support.

Ferroin:
> Hi There is new requirement that was discovered today and we have to make sure LWS_WITH_SOCKS5 is enabled and we build LWS with SOCKS5 support.
> 
> Is that hard to do?

I'm not sure. I'll look into it shortly.

prologic:
Actually I just got reminded to look at the CI error. Looking at this... Please hold off merging for the time being...

prologic:
Failing CentOS 8.x checks all good 👍 See #8200 and #8201

amoss:
This fails on Debian 10.

```
Scanning dependencies of target websockets_shared
[  1%] Building C object CMakeFiles/websockets_shared.dir/lib/core/alloc.c.o
/tmp/netdata-libwebsockets-UJZLXV/libwebsockets-3.2.2/lib/core/alloc.c: In function '_realloc':
/tmp/netdata-libwebsockets-UJZLXV/libwebsockets-3.2.2/lib/core/alloc.c:84:46: error: unused parameter 'reason' [-Werror=unused-parameter]
 _realloc(void *ptr, size_t size, const char *reason)
                                  ~~~~~~~~~~~~^~~~~~
cc1: all warnings being treated as errors
make[2]: *** [CMakeFiles/websockets_shared.dir/build.make:63: CMakeFiles/websockets_shared.dir/lib/core/alloc.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:105: CMakeFiles/websockets_shared.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

amoss:
It seems strange that we are adding packages to the dependencies in the manual instructions that we are not using. Why are we telling people to install libwebsockets packages that will not be used as we then build a local version statically?

amoss:
Further investigation shows that the Debian 10 build in CI/CD worked but failed locally. The reason is that this has not been tested in a realistic dev environment. When building locally we do something like this:
```
export ACLK=yes; CFLAGS="-O1 -ggdb -Wall -Wextra -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1" ./netdata-installer.sh --disable-lto
```
The CFLAGS that are supposed to control the netdata build are being passed to the externaldeps where they cause libwebsockets to fail to build. This will break the build for all the devs on the core-team. Please fix.

Ferroin:
> It seems strange that we are adding packages to the dependencies in the manual instructions that we are not using. Why are we telling people to install libwebsockets packages that will not be used as we then build a local version statically?

That part predates the decision to just bundle LWS. I'll remove it in a PR that I'll put up shortly that will also handle scrubbing CFLAGS for the LWS build.

Issue #7840, opened 2020-01-23
ACLK complete support for proxies (HTTP/SOCKS5) (Claiming/Auth/Link)


cakrit:
##### Feature idea summary
Outbound connections from agents to the cloud may need to go through a transparent proxy. As a user, I want to be able to provide the proxy information to netdata, so that my agent can work correctly with the cloud, through a transparent proxy. 

##### Expected behavior

- The `https_proxy` definition consists of a proxy ip/hostname, port, and optionally username and password (see for example the `curl` `-x` option). 
- The `https_proxy` definition must be provided by the user during claiming (either via one or more parameters to the claiming command, or via additional configuration settings). 
- The `https_proxy` definition must be persisted, to be read by netdata for the ACLK establishment.

### Notes
1. Check also the WSS forward proxy case
2. To provide a way for users to check if the ACLK works when setting up their proxy
3. In case the proxy is not accessible after its configuration, we need to provide hints to guide the users (eg check static routing, iptables or firewall policies)

underhood:
We should support read also environment variables `http_proxy` and `https_proxy` not just our config
`http_proxy=http://USERNAME:PASSWORD@SERVER:PORT/`

cakrit:
> We should support read also environment variables `http_proxy` and `https_proxy` not just our config
> `http_proxy=http://USERNAME:PASSWORD@SERVER:PORT/`

Nice! Though we ONLY use https of course, but sure, let's follow any conventions that exist already.

underhood:
I would say it should be in following order:
```
1. commandline flags (to be able to override environment variable)
2. env variable
3. netdata config
```
not 100% sure about 2/3 order though

stelfrag:
> I would say it should be in following order:
> 
> ```
> 1. commandline flags (to be able to override environment variable, i might want to use different proxy than rest of the system for different purposes)
> 2. env variable
> 3. netdata config
> ```
> 
> not 100% sure about 2/3 order though

I agree with the order. 

ENV affects the whole system, so it should be 2 and netdata.config overrides it

ilyam8:
if 

> netdata.config overrides it

which sounds correct, then netdata.config should be 2, right?

stelfrag:
Yes @ilyam8 you are correct. :slightly_smiling_face:

underhood:
Currently the proxy support depends completely on the libwebsockets library and its support for proxies.

Quick Googling shows
[...]

PR #8146, opened 2020-02-20
Add support for SOCKS5 proxy to the ACLK.

### LINKED ISSUE #8089: Implement SOCKS5 support for the ACLK
#### Feature idea summary

ACLK connection to VerneMQ should be able to be proxied through SOCKS5 proxy.
Netdata should accept the standard environmental variable for setting the proxy. This should be overwritten by a setting from the config file if found.

* Define the config setting.
* Testing will be via `ssh -D`.

Testing configuration will be:
* Fedora on the host where ssh -D is running.
* A VM which is running the target sshd and VerneMQ - where VerneMQ is not accessible directly from the host.
* The target endpoint will be "localhost:9002" via the SOCKS5 proxy.

##### Expected behavior

- [x] update netdata code to pass appropriate settings to LWS for socks proxy
- [x] commandline flag and config file setting should override the env var
- [x] test LWS implementation of Socks proxy works well for us (pass at least 100k msgs trough proxy to check it is stable as well

underhood:
##### Summary
Closes #8089 

We read config value `agent_cloud_link` value `proxy`. Which can have following values:
- `none` - don't use proxy
- `env` - use environment variable`socks_proxy` if can otherwise disabled
- `socks5[h]://[user:pass@]host:port` - use this configured proxy regardless of any environment variables

##### Component Name
ACLK

VerneMQ running in VM (listening on port 9002 on VM, all ports to VM blocked except ssh), socks proxy created by `ssh user@VM -D 8888`, `netdata -D -x socks5://127.0.0.1:8888` and set to connect to verneMQ `127.0.0.1:9002` (this IP is interpreted from proxy server side).

##### Additional Information
`export ACLK=yes` must be set before running netdata installer. If testing with selfsigned certificate on VernMQ then `-DACLK_SSL_ALLOW_SELF_SIGNED` must be set.

### DIFF
--- a/aclk/aclk_common.c
+++ b/aclk/aclk_common.c
@@ -1,0 +1,35 @@
+#include "aclk_common.h"
+
+struct {
+    ACLK_PROXY_TYPE type;
+    const char *url_str;
+} supported_proxy_types[] = {
+    { .type = PROXY_TYPE_SOCKS5,   .url_str = "socks5" ACLK_PROXY_PROTO_ADDR_SEPARATOR  },
+    { .type = PROXY_TYPE_SOCKS5,   .url_str = "socks5h" ACLK_PROXY_PROTO_ADDR_SEPARATOR },
+    { .type = PROXY_TYPE_UNKNOWN,  .url_str = NULL                                      },
+};
+
+static inline ACLK_PROXY_TYPE aclk_find_proxy(const char *string)
+{
+    int i = 0;
+    while( supported_proxy_types[i].url_str ) {
+        if(!strncmp(supported_proxy_types[i].url_str, string, strlen(supported_proxy_types[i].url_str)))
+            return supported_proxy_types[i].type;
+        i++;
+    }
+    return PROXY_TYPE_UNKNOWN;
+}
+
+ACLK_PROXY_TYPE aclk_verify_proxy(const char *string)
+{
+    if(!string)
+        return PROXY_TYPE_UNKNOWN;
+
+    while(*string == 0x20)
+        string++;
+
+    if(!*string)
+        return PROXY_TYPE_UNKNOWN;
+
+    return aclk_find_proxy(string);
+}
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -1,6 +1,8 @@
 #include "aclk_lws_wss_client.h"
 
 #include "libnetdata/libnetdata.h"
+#include "../daemon/common.h"
+#include "aclk_common.h"
 
 static int aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reason, void *user, void *in, size_t len);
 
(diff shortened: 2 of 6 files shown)


underhood:
reset squash

aclk/agent_cloud_link.c
```diff
@@ -1128,8 +1135,15 @@ int aclk_init(ACLK_INIT_ACTION action)
     aclk_hostname = config_get(CONFIG_SECTION_ACLK, "agent cloud link hostname", ACLK_DEFAULT_HOST);
     aclk_port = config_get_number(CONFIG_SECTION_ACLK, "agent cloud link port", ACLK_DEFAULT_PORT);
 
+    aclk_proxy = config_get(CONFIG_SECTION_ACLK, ACLK_PROXY_CONFIG_VAR_NAME, "");
```
amoss: There are very few places in the config that use empty strings as defaults, the common pattern is to use a well-defined constant to turn it off such as "no" or "none" or "disabled".

aclk/mqtt.c
```diff
@@ -42,7 +42,7 @@ void _link_shutdown()
     return;
 }
 
-int _link_lib_init(char *aclk_hostname, int aclk_port, void (*on_connect)(void *), void (*on_disconnect)(void *))
+int _link_lib_init(char *aclk_hostname, int aclk_port, const char *aclk_proxy, void (*on_connect)(void *), void (*on_disconnect)(void *))
```
amoss: Many of the changes in this PR are to read the proxy string in the top layer and pass it down through the calls into the lower layer connection. Why not just read the connection in the lower layer? The common pattern to do this in the netdata code is a static variable inside the function that would use it that is initialised on first access.

daemon/main.c
```diff
@@ -263,20 +263,23 @@ void cancel_main_threads() {
...
+    { 'P', "File to save a pid while running.",               "filename",    "do not save pid to a file"},
+    { 'i', "The IP address to listen to.",                    "IP",          "all IP addresses IPv4 and IPv6"},
+    { 'p', "API/Web port to use.",                            "port",        "19999"},
+    { 's', "Prefix for /proc and /sys (for containers).",     "path",        "no prefix"},
+    { 't', "The internal clock of netdata.",                  "seconds",     "1"},
+    { 'u', "Run as user.",                                    "username",    "netdata"},
+    { 'v', "Print netdata version and exit.",                 NULL,          NULL},
+    { 'V', "Print netdata version and exit.",                 NULL,          NULL},
+#ifdef ENABLE_ACLK
+    { 'x', "Socks5 or http proxy for outbound connections",   "proxy",       NULL},
```
amoss: I don't recall any discussion deciding that we should add this flag. There was definitely some discussion in the estimation meeting but I don't think that we reached a clear decision. What is the use-case for this flag? Why can those use-cases not be achieved using the configuration file? Why are we adding a top-level flag instead of something under `-W`?

aclk/mqtt.c
amoss: Rechecking my comments: "read the connection" should be "read the configuration".

daemon/main.c
underhood: https://github.com/netdata/netdata/issues/7840

aclk/aclk_common.c
```diff
@@ -0,0 +1,44 @@
+#include "aclk_common.h"
```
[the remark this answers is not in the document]
underhood: used by main as well as ACLK

underhood:
of course I have to change this PR as the code changed under it during the weekend. I will do that on Monday.

daemon/main.c
amoss: It’s not in the definition of the issue. You’ve mentioned in a comment that you want to add it, but where has it been agreed?

aclk/aclk_common.c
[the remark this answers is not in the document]
amoss: Is this file necessary without the addition of the command-line switch?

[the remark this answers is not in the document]
underhood: No. I separated it due to main and also later those functions might be used by claiming and authentication.

aclk/mqtt.c
underhood: reason for this is already preexisting code:
```C
aclk_hostname = config_get(CONFIG_SECTION_ACLK, "agent cloud link hostname", ACLK_DEFAULT_HOST);
aclk_port = config_get_number(CONFIG_SECTION_ACLK, "agent cloud link port", ACLK_DEFAULT_PORT);
```

therefore i put also proxy reading in the same place for consistency (which as a result means you have to pass this parameters down the chain):
```C
aclk_proxy = config_get(CONFIG_SECTION_ACLK, ACLK_PROXY_CONFIG_VAR_NAME, "");
```

amoss: I don't think that having all of the `config_` calls about a particular section in the same function is a benefit - to check all the access points to a particular section a recursive grep is needed anyway as we cannot guarantee this in general. Passing the parameters down across the layers adds unnecessary complexity. I think it is better to read it in the lower layer.

stelfrag: If we want to have the config exposed in the api (even if the aclk is not active) we need to do the calls to the config parameters twice. 

Now if a config option is added to disable the cloud, we need to read that option first and make a decision if we want to hide the rest of the ACLK related configuration 
So  if (example)

```
agent cloud link functionality = no
```

we won't read the rest of the parameters or set defaults

I agree passing the parameters two or three layers down is not needed (the aclk layer needs simplification in that regard as well)

underhood: also if it is at this level (upper) it will be common for possible HTTPS fallbacks etc.

amoss: @stelfrag that raises an interesting question - I don't think we have config options that override other config options yet. Is that possible given the current way the configuration file is loaded?

amoss: @underhood it would still be less complex to read the config option in the (not yet designed) HTTPS component and use it there.

daemon/main.c
```diff
@@ -263,20 +263,23 @@ void cancel_main_threads() {
...
-    { 'P', "File to save a pid while running.",           "filename",    "do not save pid to a file"},
-    { 'i', "The IP address to listen to.",                "IP",          "all IP addresses IPv4 and IPv6"},
-    { 'p', "API/Web port to use.",                        "port",        "19999"},
-    { 's', "Prefix for /proc and /sys (for containers).", "path",        "no prefix"},
-    { 't', "The internal clock of netdata.",              "seconds",     "1"},
-    { 'u', "Run as user.",                                "username",    "netdata"},
-    { 'v', "Print netdata version and exit.",             NULL,          NULL},
-    { 'V', "Print netdata version and exit.",             NULL,          NULL},
-    { 'W', "See Advanced options below.",                 "options",     NULL},
+    // opt description                                        arg name       default value
```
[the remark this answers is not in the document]
underhood: reverted in rebase

libnetdata/config/appconfig.h
```diff
@@ -82,18 +82,19 @@
...
-#define CONFIG_SECTION_PLUGINS   "plugins"
-#define CONFIG_SECTION_CLOUD     "cloud"
-#define CONFIG_SECTION_REGISTRY  "registry"
-#define CONFIG_SECTION_HEALTH    "health"
-#define CONFIG_SECTION_BACKEND   "backend"
-#define CONFIG_SECTION_STREAM    "stream"
-#define CONFIG_SECTION_EXPORTING "exporting:global"
-#define CONFIG_SECTION_HOST_LABEL   "host labels"
-#define EXPORTING_CONF           "exporting.conf"
+#define CONFIG_SECTION_GLOBAL      "global"
```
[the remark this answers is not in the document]
underhood: reverted in rebase

aclk/aclk_common.c
```diff
@@ -0,0 +1,44 @@
...
+
+struct {
+    ACLK_PROXY_TYPE type;
+    const char *url_str;
+} supported_proxy_types[] = {
+    { .type = PROXY_TYPE_SOCKS5,   .url_str = ACLK_PROXY_PROTO_SOCKS5  },
```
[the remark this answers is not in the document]
underhood: removed single use defines ACLK_PROXY_PROTO_SOCKS5 and ACLK_PROXY_PROTO_SOCKS5h

daemon/main.c
underhood: removed the flag as per slack discussion

amoss:
Further information about the build error, this was a clean install. To replicate:
```
build_external/bin/clean-install.sh arch current
```
Inside the arch linux box:
```
pacman -S elfutils
readelf -a /lib/libwebsockets.so
```
There are no symbols that relate to socks at all.

underhood:
looks like SOCKS5 support is optional and has to be compiled in with ccmake option. LWS documentation conveniently doesn't state that. Will have to update the installer script then and merge this after the installer script builds the static LWS.

underhood:
```
  agent cloud link hostname = localhost
   agent cloud link port = 9002
```

can it be localhost is resolving to IPv6 and VerneMQ listens only on IPv4?

underhood:
Another thing to check is configuration of sshd:
AllowTcpForwarding yes

amoss:
Have you tested this on the target configuration: the real testing environment?

amoss:
> Another thing to check is configuration of sshd:
> AllowTcpForwarding
> GatewayPorts

Both of these are set to true in the configuration. But regardless of the configuration it should either connect, or fail with a clear reason. Reporting connection refused when the port is open and accessible does not seem to be clear, or allow us to diagnose the problem.

underhood:
Well we are depending entirely on the LWS implementation of SCOKS5. I will look into if we can get some better diag info.

underhood:
What is the OS you are running Netdata on? Debian? I might retest if somehow agent on debian doesn't work.

amoss:
It's Arch Linux in a container on MacOS - as I documented above. Please focus on the real testing environment - I sent you a longer message on slack.

aclk/aclk_lws_wss_client.c
```diff
@@ -296,6 +297,11 @@ int aclk_lws_wss_connect(char *host, int port)
         lws_set_proxy(vhost, ":");
         break;
     case PROXY_TYPE_SOCKS5:
+        log = mallocz(strlen(proxy));
```
amoss: Use `strdupz()` so there is enough space for the null.

PR #8150, opened 2020-02-21
docs: Added PR title guidelines to the contribution guidelines, to make CHANGELOG more meaningful

cakrit:
##### Summary

PR titles are usually too cryptic and hide the problem they solve in technical jargon. 
We need to agree on guidelines that will make the changelog easier to read & understand. This will also greatly facilitate writing the release notes.

##### Component Name

docs

##### Next steps

As @prologic suggested, if we agree on such a format, we can have CI run regexp checks on the titles, to request improvements.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -181,6 +181,16 @@ However, it's not always that simple. Our [PR approval process](#pr-approval-pro
 
 One thing you will need to do only for your first pull request in Netdata is to accept the CLA. Until you do, the automated check for the CLA acceptance will be showing as failed. 
 
+#### PR title guidelines
+
+All verbs in PR titles must be in the past tense. E.g. "Added" instead of "Adds", "Fixed" instead of "Fixes".
+
+Bug fixes must have the title "Fixed [description of the problem as a user would perceive it], by [explanation of the correction]". 
+
+Improvements must have the title "[Area or module] : [Added/Refactored/Improved/Corrected/Started to/Stopped/Modified] [x], [in order to/so that] [benefit]". The second part may be left out, if the benefit is obvious from the improvement "x".  
+
+New collectors must have the title "[New collector] : [Name of collector]".
+
 ### PR approval process
 
 Each PR automatically [requires a review](https://help.github.com/articles/about-required-reviews-for-pull-requests/) from the code owners specified in `.github/CODEOWNERS`. Depending on the files contained in your PR, several people may be need to approve it.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -181,6 +181,16 @@ However, it's not always that simple. Our [PR approval process](#pr-approval-pro
 
 One thing you will need to do only for your first pull request in Netdata is to accept the CLA. Until you do, the automated check for the CLA acceptance will be showing as failed. 
 
+#### PR title guidelines
+
+All verbs in PR titles must be in the past tense. E.g. "Added" instead of "Adds", "Fixed" instead of "Fixes".
```
prologic: I _guess_ for some reason or another (_unsure where I got the habit from_) I've always put up PRs with verbs in the present tense because it reads better that way when opening/reviewing a PR. I'm a bit indifferent and okay with this but it'll take some mental retraining to accomplish. :)

prologic:
Do we want to put some CI that tees of PR creation events to try to enforce these rules?

joelhans:
> Do we want to put some CI that tees of PR creation events to try to enforce these rules?

As much as sanely possible, that would be cool. It's one thing to enforce these internally, but another to get outside contributors to adhere to it. I bet most contributors won't read these guidelines.

It looks like we _can_ re-title_ other users' PRs if they don't adhere to the rules. Maybe we first ask that they re-title it, and if they don't, we do it ourselves and leave a comment explaining the change we made and why we made it.

CONTRIBUTING.md
```diff
@@ -181,6 +181,16 @@ However, it's not always that simple. Our [PR approval process](#pr-approval-pro
...
+
+Bug fixes must have the title "Fixed [description of the problem as a user would perceive it], by [explanation of the correction]". 
+
+Improvements must have the title "[Area or module] : [Added/Refactored/Improved/Corrected/Started to/Stopped/Modified] [x], [in order to/so that] [benefit]". The second part may be left out, if the benefit is obvious from the improvement "x".
```
Ferroin: I think 'Optimized' should probably be added here, ans we might want to preferentially use that instead of 'Refactored' for cases that really are performance optimizations. It should make the release notes read much better.

CONTRIBUTING.md
cakrit: Agreed, but avoids having to modify EVERY SINGLE VERB when writing release notes.

cakrit:
@ferroin you can add the "Removed" verb on another PR, I merged this one so people don't have to reapprove.

PR #8159, opened 2020-02-22
backup group updated, added lsyncd

DefauIt:


ilyam8:
thx for the contribution @DefauIt 👋

PR #8163, opened 2020-02-22
Fixed reconnect issues on the ACLK.

amoss:
##### Summary
Fix the reconnection issues on the ACLK.

##### Component Name
ACLK

To reproduce the failure case - use the docker-compose.

```
build_external/bin/clean-install arch current
docker-compose -f build_external/projects/aclk-testing/vernemq-compose.yml up
docker-compose -f build_external/projects/aclk-testing/agent-compose.yml up --build
...
docker-compose -f build_external/projects/aclk-testing/vernemq-compose.yml down
...
docker-compose -f build_external/projects/aclk-testing/vernemq-compose.yml up
```

To recover the output more effectively at each state you can do this:

```
docker-compose -f build_external/projects/aclk-testing/agent-compose.yml logs | grep ACLK
```
Each up/down operation is effectively switching machine on/off in the virtual network.

Tried cycling both the broker and agent in different periods to cover all combinations of disconnect and restart timing.

##### Additional Information

Not all of the callbacks generated by LWS have their return codes checked, so there are callback
kinds that cannot ask for the WSI to be destroyed. Connection failures at different phases in
the establishment phase trigger different callback from different points. If the connection fails
early enough then no LWS_CALLBACK_WSI_DESTROY is generated and so we did not trigger our own
callback to inform the upper layer.

When the connection is torn down via callback that does not lead to the WSI_DESTROY the WSI is
internally deallocated after the callback - i.e. the pointer stops being valid at the end of
the callback. We handle this by clearing the pointer and crossing our fingers that there is no
memory leak as LWS gives us no visibility into this and if we call lws_close_free_wsi() twice
on the same pointer it can cause a segmentation fault.

This was tested in docker-compose by taking he broker/agent machines up/down in each combination.
It has no been tested in valgrind yet - as I have to update the docker-compose configuration
to handle valgrind. If you can run it through valgrind during review then please do so.

Ultimately this fix is problematic and we may have to fork libwebsockets to fix this poor aspect
of their API design.

### DIFF
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -191,7 +191,37 @@ static inline int received_data_to_ringbuff(struct lws_ring *buffer, void* data,
 	}
 	return 1;
 }
-
+   
+static const char *aclk_lws_callback_name(enum lws_callback_reasons reason)
+{
+	switch(reason)
+	{
+		case LWS_CALLBACK_CLIENT_WRITEABLE:
+			return "LWS_CALLBACK_CLIENT_WRITEABLE";
+		case LWS_CALLBACK_CLIENT_RECEIVE:
+			return "LWS_CALLBACK_CLIENT_RECEIVE";
+		case LWS_CALLBACK_PROTOCOL_INIT:
+			return "LWS_CALLBACK_PROTOCOL_INIT";
+		case LWS_CALLBACK_SERVER_NEW_CLIENT_INSTANTIATED:
+			return "LWS_CALLBACK_SERVER_NEW_CLIENT_INSTANTIATED";
+		case LWS_CALLBACK_USER:
+			return "LWS_CALLBACK_USER";
+		case LWS_CALLBACK_CLIENT_CONNECTION_ERROR:
+			return "LWS_CALLBACK_CLIENT_CONNECTION_ERROR";
+		case LWS_CALLBACK_CLIENT_CLOSED:
+			return "LWS_CALLBACK_CLIENT_CLOSED";
+		case LWS_CALLBACK_WS_PEER_INITIATED_CLOSE:
+			return "LWS_CALLBACK_WS_PEER_INITIATED_CLOSE";
+		case LWS_CALLBACK_WSI_DESTROY:
+			return "LWS_CALLBACK_WSI_DESTROY";
+		case LWS_CALLBACK_CLIENT_ESTABLISHED:
+			return "LWS_CALLBACK_CLIENT_ESTABLISHED";
+		default:
+			// Not using an internal buffer here for thread-safety with unknown calling context.
+			error("Unknown LWS callback %u", reason);
+			return "unknown";
+	}
+}
 static int
 aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reason,
 			void *user, void *in, size_t len)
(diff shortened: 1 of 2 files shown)


aclk/aclk_lws_wss_client.c
```diff
@@ -206,6 +236,7 @@ aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reason,
 		return -1;
 	}
 
+	// Callback servicing is forced when we are closed from above.
 	if( inst->upstream_reconnect_request ) {
 		error("Closing lws connectino due to libmosquitto error.");
```
cosmix:
```suggestion
		error("Closing lws connection due to libmosquitto error.");
```

aclk/aclk_lws_wss_client.c
```diff
@@ -214,76 +245,86 @@ aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reason,
...
+			if(!received_data_to_ringbuff(inst->read_ringbuffer, in, len))
+				retval = 1;
+			aclk_lws_mutex_unlock(&inst->read_buf_mutex);
+
+			if(likely(inst->callbacks.data_rcvd_callback))
+				// to future myself -> do not call this while read lock is active as it will eventually
+				// want to acquire same lock later in aclk_lws_wss_client_read() function
+				inst->callbacks.data_rcvd_callback();
+			else
+				inst->data_to_read = 1; //to inform logic above there is reason to call mosquitto_loop_read
```
cosmix: Haven’t looked at the mosquito layer, but for safety shouldn’t this be part of the critical section?

aclk/aclk_lws_wss_client.c
```diff
@@ -214,76 +245,86 @@ aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reason,
...
-						LWS_CALLBACK_USER, ACLK_LWS_WSS_RECONNECT_TIMEOUT);
-			inst->reconnect_timeout_running = 1;
-		}
-		/* FALLTHRU */
-#endif
-		//no break here on purpose we want to continue with LWS_CALLBACK_WSI_DESTROY
+		inst->lws_wsi = NULL; // inside libwebsockets lws_close_free_wsi is called after callback
+        if (inst->callbacks.connection_closed)
+            inst->callbacks.connection_closed();
+		return -1;			  // the callback response is ignored, hope the above remains true
```
cosmix: Maybe define a constant instead of returning magic number `-1` directly?

aclk/aclk_lws_wss_client.c
```diff
@@ -214,76 +245,86 @@ aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reason,
...
             inst->callbacks.connection_closed();
         break;
 	case LWS_CALLBACK_CLIENT_ESTABLISHED:
 		inst->websocket_connection_up = 1;
 		if(inst->callbacks.connection_established_callback)
 			inst->callbacks.connection_established_callback();
 		break;
+
 	default:
+		error("Unexecpted callback from libwebsockets %s",aclk_lws_callback_name(reason));
```
cosmix:
```suggestion
		error("Unexpected callback from libwebsockets %s",aclk_lws_callback_name(reason));
```

PR #8187, opened 2020-02-24
Cleaning up the ACLK part 2

### LINKED ISSUE #8034: ACLK improvements
#### Summary

- Additional configuration options for parameter tuning
  - Retrieve parameters early enough so that defaults are set and show up in the runtime configuration
  - Eg. Max reconnection timeout, ACLK disable option via a config option
- Ensure that the ACLK configuration (default values) appear in the netdata.conf
  - The values are now fetched only when an agent is claimed and there is an attempt to establish the ACLK
- Integration of the re-connection delay on ACLK failure
  - When the wss connection drops the aclk layer needs to:
    - Wait `aclk_reconnect_delay()` seconds
    - Try to re-establish the link

##### Additional suggestions, improvements, fixes based on the PR #8139 comments
- Remove defines that are used only once
- Rewrite the parts of the code that use pthread functions directly
  - Should use libuv
- Move macros to header files
- Remove the code that checks if we need mqtt over wss
  - The protocol will support MQTT over secure websockets
- Check instances of structures that are publicly exposed when there is no need and make static
  - Example: The lws_engine_instance
- Check instances of functions that are publicly exposed when there is no need and make static
  - Example: aclk_query_find
- Clean up code not in use and the associated supporting structure fields
  - Example: defer deletion of queries
- Remove the on_connect / on_disconnect call backs.
  - The code should be accessing the necessary connect and disconnect functions directly
- Remove all code related to MQTT that implies not use of wss
  - Currently there is code in place that uses libmosquitto to offer MQTT either plaintext or over SSL, Since we introduced libwebsockets, there that code path is not used at runtime. 
  - Note: in the future we may decide to use that code path if libmosquitto  supports libwebsockets.
- Specs compliance regarding the filtering on duplicate messages arriving at the same time
  - Currently the code will filter out those messages arriving at the same second with identical fields
     - Topic, payload, msg-id
- Handle cases where the agent receives invalid JSON requests
  - This should be rare possibly an oversight/part of a breaking change that we may have introduced in the protocol. Still we must have a way to identify and fix in coordination with the cloud team

amoss:
#### Summary
Closes #8034 
Second phase of the cleaning up. Addressing all outstanding comments about code quality and robustness.

##### Component Name
ACLK.

* Tested the compilation works with the feature-flag disabled.
* Tested the ACLK still functions in the docker-compose environment.
* Test the reconnection behaviour and on-connect payload delivery in the `build_external` docker-compose.

##### Additional Information

* There are currently known bugs - we will be tracking bugs during the smoke testing so that they  can be cleaned up next week.
* On a reconnect `mqtt_init` is called again and causes a memory-leak.
* The LWS event-loop returns instantly in the time after initialization, but before connection. This floods the library with calls and breaks something. A 100ms delay has been added as a work-around but must be fixed properly before we do load-testing.
* All queries are ignored during pop-corning, but commands from the cloud are in the same queue for processing so the agent would ignore requests during this period.
* Any query resets the pop-corning timer (including encapsulated queries) so pop-corning can go on indefinitely.
* The implementation of the time-out / check is wrong and it will always sleep 10s longer than necessary.


aclk/aclk_lws_wss_client.c
```diff
@@ -5,378 +5,375 @@
...
-		"aclk-wss",
-		aclk_lws_wss_callback,
-		sizeof(struct aclk_lws_wss_perconnect_data),
-		0, 0, 0, 0
-	},
-	{ NULL, NULL, 0, 0, 0, 0, 0 }
-};
+static const struct lws_protocols protocols[] = { { "aclk-wss", aclk_lws_wss_callback,
+                                                    sizeof(struct aclk_lws_wss_perconnect_data), 0, 0, 0, 0 },
+                                                  { NULL, NULL, 0, 0, 0, 0, 0 } };
```
underhood: seems in this case it is less readable?

amoss: I can't remember the change but I can guess what happened. The `clang-format` doesn't have a setting for vertical alignment of elements inside structures. It bin-packs them instead so if you had this arranged in a more readable way then the linter removed it. I prefer the vertical alignment personally, but the team "agrees" on the `clang-format` settings...

aclk/aclk_lws_wss_client.c
```diff
@@ -5,378 +5,376 @@
...
-	info.user = inst;
-	
-	inst->lws_context = lws_create_context(&info);
-	if(!inst->lws_context)
-		goto failure_cleanup_2;
+    memset(&info, 0, sizeof(struct lws_context_creation_info));
+    info.options = LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT;
+    info.port = CONTEXT_PORT_NO_LISTEN;
+    info.protocols = protocols;
+    info.user = engine_instance;
```
underhood: is this necessary if we remove
`struct aclk_lws_wss_engine_instance *inst = lws_context_user(lws_get_context(wsi));`

amoss: That's true.

underhood:
adds warning:
```
aclk/aclk_lws_wss_client.c: In function ‘aclk_lws_wss_callback’:
aclk/aclk_lws_wss_client.c:290:49: warning: passing argument 1 of ‘aclk_lws_wss_connect’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  290 |             aclk_lws_wss_connect(engine_instance->host, engine_instance->port); // Makes the outgoing connection
      |                                  ~~~~~~~~~~~~~~~^~~~~~
aclk/aclk_lws_wss_client.c:163:32: note: expected ‘char *’ but argument is of type ‘const char *’
  163 | int aclk_lws_wss_connect(char *host, int port)
      |                          ~~~~~~^~~~
```

aclk/aclk_lws_wss_client.c
vlvkobal: I wouldn't say that the team "agrees" :).

Issue #7724, opened 2020-01-14
Verify/Fix usage of CI check clang-format


prologic:
#### Bug report summary

In reviewing netdata/netdata#7640 I noticed that the `clang-format` CI lint checks that are done against the codebase against `*.[ch]` files just print line counts of differences from actual code committed vs. accept formatting standards.

Example:

```bash
$ .travis/clangformat.sh | head -n 5
./database/rrddim.c: 475
./database/rrdhost.c: 1031
./database/rrdfamily.c: 73
./database/rrdcalc.h: 153
./database/rrd.h: 895
```

This currently produces ~279 lines of output I'm not sure anyone really looks at (_because its far too much output_):

```bash
$ .travis/clangformat.sh | wc -l
297
```

This issue captures this and is a discussion for how to:

- [ ] What do we want to do here?
- [ ] Should we fail CI checks?
- [ ] Should we en-mass cleanup the codebase?
- [ ] Or should we simply remove this check?
 
##### OS / Environment (output of `uname -a; command -v sw_vers && sw_vers || grep -Hv "^#" /etc/*release`)

N/A

##### Netdata version (ouput of `netdata -V`)

N/A

##### Component Name

CI

##### Steps To Reproduce

```bash
.travis/clangformat.sh
```

##### Expected behavior

This check should either:

- Fail CI

or:

- Be removed (_if not useful_)

Ferroin:
This ultimately needs to be up to the C devs and possibly @cosmix and/or @cakrit. I'm personally of the opinion that it would be good to enforce the coding style, but I'm also not the one who's writing the code this would impact.

amoss:
If we remove the check then it should be replaced with some measure of how much the code-base deviates from the coding-style as defined by the clang-format linter. The long-term goal is for the entire code-base to match the agreed upon coding style in a way that is mechanically checkable. The measure used is coarse but realistic. It is not being read at the moment because enforcing the code style guide is not currently a priority - but using it will become more important in the future.

prologic:
I raised this issue only to encourage discussion of what we'd like to do here and as a placeholder for us to fix the check. Once we've decided what we'd like to do one way or the either we can easily put up a PR that either enforces the check, improves it, etc. But as @Ferroin rightfully says, this isn'y really up to either of us on the SRE team (_we've decided in team discussions that code we write and maintain will just be auto-formatted, so we would probably learn towards deviations from the agreed upon tool used would fail CI_)

cakrit:
[...]

PR #8188, opened 2020-02-24
Remvoed the use of clang-format that does not actually block PRs or surface anything to developers

prologic:
See #7724


Dockerfile.dev
```diff
@@ -0,0 +1,19 @@
...
+
+
+# Copy Sources
+WORKDIR /netdata
+COPY . .
+
+# Build
+RUN ./netdata-installer.sh -u --dont-wait --dont-start-it --disable-go
+
+CMD ["/bin/sh"]
```
Ferroin: Looks like this got included by mistake somehow.

.clang-format
```diff
@@ -1,118 +0,0 @@
-# SPDX-License-Identifier: GPL-2.0
```
Ferroin: Are we certain that nobody's actually using this? I'd hate to remove it had have to add it back again immediately because someone was using it.

amoss: We are all using this - why is it being removed?

amoss:
Had a chat with @prologic on slack and we're going to skip this for now.

prologic:
Misunderstanding on my part! #8196

Issue #8200, opened 2020-02-25
Investigate GetPageSpeed repo usage


prologic:
Currently our CI fails with:

```sh
2020-02-25T17:08:44.2077932Z  > Checking for getpagespeed-extras ...
2020-02-25T17:08:44.2078113Z yum repolist 
2020-02-25T17:08:44.7508651Z CentOS-8 - AppStream                             13 kB/s | 4.3 kB     00:00    
2020-02-25T17:08:45.2993145Z CentOS-8 - Base                                 7.2 kB/s | 3.8 kB     00:00    
2020-02-25T17:08:45.5869910Z CentOS-8 - Extras                               5.8 kB/s | 1.5 kB     00:00    
2020-02-25T17:08:47.0547172Z CentOS-8 - PowerTools                           1.4 MB/s | 2.0 MB     00:01    
2020-02-25T17:08:47.7745390Z Extra Packages for Enterprise Linux 8 - x86_64   68 kB/s |  20 kB     00:00    
2020-02-25T17:08:48.5889350Z Running in non-interactive mode, assuming yes (y)
2020-02-25T17:08:48.5889489Z  > Would have promptedfor PowerTools not found, shall I install it? ...
2020-02-25T17:08:48.5892679Z yum -y install https://extras.getpagespeed.com/release-el8-latest.rpm 
2020-02-25T17:08:48.9095002Z Last metadata expiration check: 0:00:01 ago on Tue Feb 25 17:08:47 2020.
2020-02-25T17:08:49.3693354Z release-el8-latest.rpm                           62 kB/s |  13 kB     00:00    
2020-02-25T17:08:49.5319448Z Dependencies resolved.
2020-02-25T17:08:49.5329996Z ================================================================================
2020-02-25T17:08:49.5330118Z  Package                         Arch       Version      Repository        Size
2020-02-25T17:08:49.5330217Z ================================================================================
2020-02-25T17:08:49.5330307Z Installing:
2020-02-25T17:08:49.5330973Z  getpagespeed-extras-release     noarch     10-11        @commandline      13 k
2020-02-25T17:08:49.5331043Z 
2020-02-25T17:08:49.5331123Z Transaction Summary
2020-02-25T17:08:49.5331191Z ================================================================================
2020-02-25T17:08:49.5331451Z Install  1 Package
2020-02-25T17:08:49.5331513Z 
2020-02-25T17:08:49.5335104Z Total size: 13 k
[...]

PR #8201, opened 2020-02-25
Disable the CentOS 8.x installer CI

prologic:
See #8200


log
```diff
@@ -0,0 +1,267 @@
+Loading /etc/os-release ...
```
knatsakis: Is this committed by mistake?

prologic:
#8202 supersedes this @knatsakis sorry :)

PR #8203, opened 2020-02-26
Retry transient failures from our lifecycle tests on Travis CI

prologic:


.travis.yml
```diff
@@ -170,7 +170,7 @@ jobs:
       after_failure: post_message "TRAVIS_MESSAGE" "Build/Install failed on ubuntu 18.04"
 
     - name: Run netdata lifecycle, on ubuntu 18.04 (Containerized)
-      script: docker run -it -v "${PWD}:/netdata:rw" -w /netdata "netdata/os-test:ubuntu1804" bats --tap tests/lifecycle.bats
+      script: for i in "$(seq 0 4)" do; docker run -it -v "${PWD}:/netdata:rw" -w /netdata "netdata/os-test:ubuntu1804" bats --tap tests/lifecycle.bats && break; sleep 15; done
```
Ferroin: Can we add a post-run delimiter so that it's easy to spot the point after a run fails where the new one begins? Doesn't need to be fancy, just something unique enough to search for.

prologic: Done

prologic:
CI should have caught:

```sh
$ yamllint .travis.yml
.
.
.
173:51    error    syntax error: mapping values are not allowed here (syntax)
.
.
.

```

😳 #8295 to fix this

PR #8212, opened 2020-02-26
New  react dashboard.js to Netdata

### LINKED ISSUE #7940: add feature flag for new dashboard release
##### Feature idea summary
We want to release new Dashboard but still have an option to fallback to old version, in case of bugs or bad experience.

##### Expected behavior
- we should add a new flag (feature flag) to Netdata, called `use new dashboard`
- flag should be OFF by default, at least right now, until we test it internally
- when that flag is ON, requests to `index` (`/`) in web server should point to `index-new.html` instead of `index.html`

When we remove 2 versions, we'll then switch to more friendly names

thiagoftsm:
##### Summary
Fixes #7940 

This PR brings the possibility  of user to select between two dashboards to see Netdata charts.
##### Component Name
Web
To test this PR it is necessary to add in the section `[web]` a new variable called `use react dashboard`, when it is set to `yes` it uses the new dashboard, and when it has the value `no` it will use the old dashboard.

During my tests, I had to press `CTRL+SHIFT+R` for Firefox ignore the cache and give me the correct file. Another possibility it is to remove completely the cache, using `private window`, for example.


Ferroin:

web/server/web_client.c
```diff
@@ -1490,7 +1490,11 @@ static inline int web_client_process_url(RRDHOST *host, struct web_client *w, ch
     strncpyz(filename, w->last_url, FILENAME_MAX);
     tok = mystrsep(&url, "?");
     buffer_flush(w->response.data);
-    return mysendfile(w, (tok && *tok)?tok:"/");
+    char *file_selection = (tok && *tok)?tok:"/";
+    if(web_client_default_dashboard == WEB_SERVER_REACT_DASHBOARD && !strcmp(file_selection, "/"))
```
mfundul: Does this comparison with "/" work for all cases?

web/server/static/static-threaded.c
```diff
@@ -482,6 +496,11 @@ void *socket_listen_main_static_threaded(void *ptr) {
 
             static_workers_private_data = callocz((size_t)static_threaded_workers_count, sizeof(struct web_server_static_threaded_worker));
 
+            if (is_there_react_dashboard()) {
```
mfundul: Is there a way to have non-static-threaded web server? In theory we should allow dynamic but I can't find this codepath for some reason.

web/server/web_client.c
thiagoftsm: Yes, I could not see any problems with my tests.

web/server/static/static-threaded.c
thiagoftsm: Considering what I saw in parts commented on the web server, I think it was in the past, but it was moved to be static, so it is possible to have it.

jacekkolasa:
@thiagoftsm i've added a short note on how to enable the new Dashboard. Do you think it's ok?

thiagoftsm:
> @thiagoftsm i've added a short note on how to enable the new Dashboard. Do you think it's ok?

I think it is enough, but I prefer to ask our specialists, @joelhans and @shortpatti , plese,  can you review @jacekkolasa changes in our documentation that describes how to enable the new dashboard? Thank you!

web/gui/README.md
```diff
@@ -157,3 +157,8 @@ For information on creating custom dashboards from scratch, see the [custom
 dashboards](custom/) or [Atlassian Confluence dashboards](confluence/) guides.
 
 [![analytics](https://www.google-analytics.com/collect?v=1&aip=1&t=pageview&_s=1&ds=github&dr=https%3A%2F%2Fgithub.com%2Fnetdata%2Fnetdata&dl=https%3A%2F%2Fmy-netdata.io%2Fgithub%2Fweb%2Fgui%2FREADME&_u=MAC~&cid=5792dfd7-8dc4-476b-af31-da2fdb9f93d2&tid=UA-64295674-3)]()
+
+## New React Dashboard
```
joelhans: We should move this much higher in the document. Let's move it to above **Navigating the standard dashboard**

web/gui/README.md
```diff
@@ -157,3 +157,8 @@ For information on creating custom dashboards from scratch, see the [custom
...
+
+We've been working on refactoring the Dashboard code, to make adding new features and fixing bugs easier. Currently it doesn't include any noticeable changes in UI. The new Dashboard is added to the codebase, but it's not enabled by default. To enable it, please add `use react dashboard = yes` in `[Web]` section in your `netdata.conf`.
```
joelhans:
```suggestion
We've been working on refactoring the standard dashboard code to simplify the process of adding new features and fixing bugs. 

The new dashboard is added to the codebase, but isn't enabled by default. To enable it, add `use react dashboard = yes` in `[web]` section in your `netdata.conf`. Currently, the refactored dashboard doesn't include any noticeable changes in UI.
```

web/gui/README.md
jacekkolasa: done, thanks for tip!

web/server/static/static-threaded.c
cakrit: This function doesn't even belong here, but I don't care right now. Create a new issue to discuss how to properly implement this please. We still need it merged immediately!

cosmix: This is unacceptable guys. This function shouldn't be here in the first place. Choosing the right HTML file to serve should not be done in C. The web server should be agnostic.

web/server/static/static-threaded.c
```diff
@@ -454,6 +455,19 @@ static void socket_listen_main_static_threaded_cleanup(void *ptr) {
...
+
+    FILE *fs = fopen(webfilename, "r");
+    if(fs) {
+        fclose(fs);
+        return 1;
+    }
+
+    return 0;
+}
+
```
cosmix: This should be removed and replaced with the respective functionality either at the installer level, or some script. There's no way we're going to have a C function that's coupled to the version or implementation of the dashboard.

web/server/web_client.c
cosmix: No mention of React or any specific version of the dashboard (or other javascript code) should be in the C code.

web/server/web_server.h
```diff
@@ -47,6 +47,10 @@ extern int web_client_timeout;
 extern int web_client_first_request_timeout;
 extern long web_client_streaming_rate_t;
 
+#define WEB_SERVER_OLD_DASHBOARD 0
+#define WEB_SERVER_REACT_DASHBOARD 1
```
cosmix: Nope. Remove those please.

netdata-installer.sh
```diff
@@ -790,6 +794,46 @@ cat << OPTIONSEOF
...
+
+copy_react_dashboard() {
+  run mv "${1}/index.html" "${1}/index-new.html"
+  run cp -r "${1}"/* "${NETDATA_WEB_DIR}"
+}
+
+install_react_dashboard() {
+  if [ -n "${NETDATA_DISABLE_REACT_DASHBOARD}" ] ; then
+    return 0
+  fi
```
ilyam8: is there a way to compare latest/actual and download it only if the dashboard not installed/has newer version?

thiagoftsm: The indexes have different checksum, so we can use it.

thiagoftsm:
I am closing the PR, because we took a new direction with #8353 .

PR #8214, opened 2020-02-26
Remove the confusion around the multiple Dockerfile(s) we have

prologic:
##### Summary

- `Dockerfile` -> `Dockerfile.test
- `Dockerfile` -> `./packaging/docker/Dockerfile` **Our current production Docker Image**

##### Component Name

- area/packaging

##### Test Plan

1. Ensured `docker build` works at the top-level with the now symlinked `Dockerfile`.
2. Ensured `docker build` works for the `Dockerfile.test` and it can be run correctly.

##### Additional Information

**NOTE:** The `Dockerfile.test` was put toegether quickly as a quick/fast way
to build the agent and run it for dev/testing purposes only. It _should_ not be
used in production at this time.


amoss:
I don’t understand the context for this change - what problem are you fixing?

Ferroin:
> I don’t understand the context for this change - what problem are you fixing?

It's aligning with the convention that a file called `Dockerfile` in the root of a source tree is what gets used to build the actual _production_ Docker images for that software. We should be making it as easy as possible for people new to the project to figure out what we're doing when it comes to stuff like this.

amoss:
If `Dockerfile.test` is for use in dev/testing - why not use the Dockerfiles / scripts in `build_external`? In particular - it would be nice to keep:
* Incremental builds
* Multiple base images
* Something that we can easily insert into `docker-compose` for more complex scenarios.

Ferroin:

prologic:
> If `Dockerfile.test` is for use in dev/testing - why not use the Dockerfiles / scripts in `build_external`? In particular - it would be nice to keep:
> 
> * Incremental builds
> * Multiple base images
> * Something that we can easily insert into `docker-compose` for more complex scenarios.

Those will have to remain for now, they're a lot more complex and even contain various `docker-compose` setups.

prologic:
> @cosmix We need the deep clone so we can skip them if nothing that we actually run unit tests for has changed.

Or put another way, I don't know how to simply symlink a `Dockerfile.test` to something in `./build-external/`

cosmix:
@prologic you could then specify a shallow clone with a set `depth`. Of course, this raises the question "What should the `depth` parameter be?", whose answer complicates things a bit, but it would be much faster if we could do that instead.

prologic:
> @prologic you could then specify a shallow clone with a set `depth`. Of course, this raises the question "What should the `depth` parameter be?", whose answer complicates things a bit, but it would be much faster if we could do that instead.

Right now I "don't know" what this is or should be; but we can/will certainly re-evaluate it for sure! A full clone is pointless and waste of time here which is why the default is a shallow clone of depth 1 :) (_as it should be_)

prologic:
Going to merge this despite:

> 28 successful, 2 expected, and 5 neutral checks

Assuming they will never report back or something else is wrong; but unrelated to this change.

PR #8219, opened 2020-02-27
packaging/installer: fix bundle_libwebsockets

ilyam8:
##### Summary

Installer always bundles `libwebsockets`

There is a wrong check in the `bundle_libwebsockets` function.

##### Component Name

`packaging/installer`

I did 

> sudo ./netdata-installer.sh --install /opt --disable-cloud

Installer doesnt try to bundle libwebsockets


underhood:
this should also pass --disable-aclk flag down to configure

Issue #8151, opened 2020-02-21
ACLK Claiming Script has to be updated


underhood:
##### Feature idea summary
Claiming has been changed. Script `netdata-claim,sh` has to be updated and tested.

##### Expected behavior
claiming script works with test endpoint provided by cloud team

underhood:
Closed by #8220

hmoragrega:
The script does not parse correctly the new error format response

PR #8220, opened 2020-02-27
Fix claiming script to reflect API changes on netdata.cloud

cosmix:
This PR fixes the netdata.cloud claiming script.


underhood:
Closes  #8151

PR #8232, opened 2020-02-27
Update function pointer and calls on collector

### LINKED ISSUE #8231: Hash table from collectors
##### Feature idea summary
The first ebpf collectors had static size, but this is not good, considering that some variables monitored are flexible  and can be adjusted using a file inside `/proc` or `systemctl`.

This PR won't change any eBPF program, because they are coded with  default values, but we are also changing the shared library used to load the eBPF programs that will load the hash tables with new values collected on user ring.

The hash tables used with the eBPF collectors has direct relationship with the feature we are monitoring, but in normal situation we are using per CPU hash to speed up the kernel rin and we group them in the user ring.

After to bring the solution  on kernel-collector with the PR https://github.com/netdata/kernel-collector/pull/83 it is necessary to adjust the collector to use it.
##### Expected behavior
To have conditions to set hash sizes from collectors.

thiagoftsm:
##### Summary
Fixes #8231 

This PR brings compatibility to eBPF collectors with the changes we are bringing to the shared library in the PR https://github.com/netdata/kernel-collector/pull/83.
##### Component Name
Collector
1 - It is necessary to do a download of the collector according your kernel and libc on https://github.com/netdata/kernel-collector/actions/runs/46397830
2 - It is necessary to do the manual installation of the files.
3 - Compile and run the collector.

##### Additional Information
This PR cannot be merged before https://github.com/netdata/kernel-collector/pull/83, else it won't work.
When we merge 83, we will also need to bring a new release.

### DIFF
--- a/collectors/ebpf_process.plugin/ebpf_process.c
+++ b/collectors/ebpf_process.plugin/ebpf_process.c
@@ -37,7 +37,7 @@ void netdata_cleanup_and_exit(int ret) {
 // ----------------------------------------------------------------------
 //Netdata eBPF library
 void *libnetdata = NULL;
-int (*load_bpf_file)(char *, int) = NULL;
+int (*load_bpf_file)(char *, int, int *) = NULL;
 int (*set_bpf_perf_event)(int, int);
 int (*perf_event_unmap)(struct perf_event_mmap_page *, size_t);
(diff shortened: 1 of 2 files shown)


Ferroin:

Issue #8247, opened 2020-03-01
ceph plugin error unhandled exception: 'osd_perf_infos'


olobarius:
##### Bug report summary

current netdata ceph plugin function _get_osd_perf incompatible with array returned from recent ceph 14.2 osd perf mon_command.
Current osd perf returns osd_perf_infos under osdstats.

Howto fix: (add ['osdstats'] at the end of the function)
So, in /usr/libexec/netdata/python.d/ceph.chart.py change

```python
    def _get_osd_perf(self):
        """
        Get ceph osd performance
        :return: ceph osd perf --format json
        """
        return json.loads(self.cluster.mon_command(json.dumps({
            'prefix': 'osd perf',
            'format': 'json'
        }), '')[1].decode('utf-8'))
```

TO

```python
    def _get_osd_perf(self):
        """
        Get ceph osd performance
        :return: ceph osd perf --format json
        """
        return json.loads(self.cluster.mon_command(json.dumps({
            'prefix': 'osd perf',
            'format': 'json'
        }), '')[1].decode('utf-8'))['osdstats']
```

Then it will work again.

##### OS / Environment
I am running netdata 1.20 on proxmox 6.1-5 with ceph 14.2.6 Nautilus.

Linux sipxen 5.3.18-2-pve #1 SMP PVE 5.3.18-2 (Sat, 15 Feb 2020 15:11:52 +0100) x86_64 GNU/Linux
/etc/os-release:PRETTY_NAME="Debian GNU/Linux 10 (buster)"
/etc/os-release:NAME="Debian GNU/Linux"
/etc/os-release:VERSION_ID="10"
/etc/os-release:VERSION="10 (buster)"
/etc/os-release:VERSION_CODENAME=buster
/etc/os-release:ID=debian
/etc/os-release:HOME_URL="https://www.debian.org/"
/etc/os-release:SUPPORT_URL="https://www.debian.org/support"
/etc/os-release:BUG_REPORT_URL="https://bugs.debian.org/"

##### Netdata version
netdata v1.20.0-50-nightly

##### Component Name
ceph plugin

##### Steps To Reproduce

```cmd
./edit-config python.d/ceph.conf
/usr/libexec/netdata/plugins.d/python.d.plugin 1 debug trace ceph

Error:
2020-03-01 11:20:43: python.d ERROR: ceph[ceph] : update() unhandled exception: 'osd_perf_infos'
[...]

PR #8248, opened 2020-03-01
python.d/ceph: fix get_osd_perf() for 14.2+

ilyam8:
##### Summary

Fixes: #8247

Ceph module uses 'osd_perf_infos' data from `osd perf` response. Its been moved under `osdstats` since 14.2.

cc @lets00

##### Component Name
[`collectors/python.d.plugin/ceph`](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/ceph)

I did no tests, solution is pretty simple and proposed by @olobarius in #8247

Tested by @paulmezz

___

This PR backward compatible with older versions.

Before 14.2 we got `osd_perf_infos` directly from `ceph osd perf --format json` response. In current version it is in `osdstats` section of the response.

### DIFF
--- a/collectors/python.d.plugin/ceph/ceph.chart.py
+++ b/collectors/python.d.plugin/ceph/ceph.chart.py
@@ -327,10 +327,15 @@ class Service(SimpleService):
         Get ceph osd performance
         :return: ceph osd perf --format json
         """
-        return json.loads(self.cluster.mon_command(json.dumps({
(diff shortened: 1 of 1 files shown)


ilyam8:
@olobarius

Was `osdstats` in the `osd perf` before? If it was current fix is not good, it wont work for versions before 14.2

ilyam8:
@olobarius it would be nice if you test the PR

PR #8254, opened 2020-03-02
Added support for building libmosquitto on FreeBSD/macOS.

Ferroin:
##### Summary

This takes advantage of the fact that cmake is now a required build dependency to allow building libmosquitto on FreeBSD and macOS systems.

##### Component Name

area/packaging

Verified that the installer appears to work correctly on FreeBSD 11, FreeBSD 12, and macOS 10.15.

##### Additional Information

This will bring ACLK dependency support for FreeBSD and macOS installs using `kickstart.sh` on par with Linux. If we want people to actually have ACLK support on these platforms though, we need to get the Homebrew formula and the FreeBSD port updated to bundle libmosquitto and depend on LWS 3.x, as those are the installation methods most users are using on them.

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -458,11 +458,19 @@ trap build_error EXIT
 # -----------------------------------------------------------------------------
 
 build_libmosquitto() {
-  if [ "$(uname -s)" = Linux ] ; then
+  if [ "$(uname -s)" = Linux ]; then
     run env CFLAGS= CXXFLAGS= LDFLAGS= make -C "${1}/lib"
   else
     pushd ${1} > /dev/null || return 1
-    run env CFLAGS= CXXFLAGS= LDFLAGS= cmake -D WITH_STATIC_LIBRARIES:boolean=YES .
+    if [ "$(uname)" = "Darwin" ] && [ -d /usr/local/opt/openssl ]; then
+      run env CFLAGS= CXXFLAGS= LDFLAGS= cmake \
+        -D OPENSSL_ROOT_DIR=/usr/local/opt/openssl \
+        -D OPENSSL_LIBRARIES=/usr/local/opt/openssl/lib \
+        -D WITH_STATIC_LIBRARIES:boolean=YES \
+        .
+    else
+      run env CFLAGS= CXXFLAGS= LDFLAGS= cmake -D WITH_STATIC_LIBRARIES:boolean=YES .
+    fi
     run env CFLAGS= CXXFLAGS= LDFLAGS= make -C lib
     run mv lib/libmosquitto_static.a lib/libmosquitto.a
     popd || return 1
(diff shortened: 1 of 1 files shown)


cosmix:
@Ferroin This is not working for me on macOS 10.15.3. Both `libmosquitto` and `libwebsockets` fail to be configured/built. What's worse is that even though the installer clearly indicates failure, it doesn't properly determine its state and considers the first lib (mosquitto) built and proceeds with the second, failing miserably.

Both libs fail to configure as cmake cannot seem to be able to find OpenSSL (I've got OpenSSL installed via homebrew)

Ferroin:
> @Ferroin This is not working for me on macOS 10.15.3. Both `libmosquitto` and `libwebsockets` fail to be configured/built. What's worse is that even though the installer clearly indicates failure, it doesn't properly determine its state and considers the first lib (mosquitto) built and proceeds with the second, failing miserably.
> 
> Both libs fail to configure as cmake cannot seem to be able to find OpenSSL (I've got OpenSSL installed via homebrew)

Can you try building inside a homebrew shell (`brew sh`)?

I'm very quickly comming to the conclusion that we need to just get on top of managing the Homebrew formula ourselves and drop support for the auto-build approach in the install scripts on macOS.

cosmix:
@Ferroin yes of course it does, but that's beside the point: a user wouldn't know (and shouldn't be expected) to do that! The installer should be taking care of this, which I think is probably more trouble than it's worth — I probably agree with you re: managing the Homebrew formula ourselves.

Ferroin:
> @Ferroin yes of course it does, but that's beside the point: a user wouldn't know (and shouldn't be expected) to do that! The installer should be taking care of this, which I think is probably more trouble than it's worth — I probably agree with you re: managing the Homebrew formula ourselves.

Agreed, though if it's working correctly in `brew sh`, I _should_ be able to get the rest of Netdata to build and install correctly from there too.

Long term of course it's best if we just handle the homebrew formula ourselves.

prologic:
- Tested on FreeBSD 12.1
- Tested current `master` on FreeBSD 12.1 (_weird that the compilation works without this PR?!_)
- Tested on Linux (_Alpine_)

prologic:
- Tested and fixed on macOS

Ferroin:
@amoss @underhood Could one of the tw of you take a look at this possibly? This should be the last thing needed to get working ACLK support on FreeBSD/macOS.

underhood:
I do not have any mac available but I will try to take a look later today or during the weekend.

amoss:
This builds both libraries on MacOS - but this branch is very out of date and I cannot see what else works. Please rebase on master so that I can see how this interacts with the other recent changes to the same code.

amoss:
I don't think that @prologic can be a reviewer when he is also an author.

prologic:
> I don't think that @prologic can be a reviewer when he is also an author.

I contributed, not authored 🤣

amoss:
I don’t think that you should give a review approval when you wrote part of the code - the point of reviewing to get a fresh pair of eyes to look at the code.

Ferroin:
@amoss He only contributed to one commit here, it still make ssense for him to review the ones he didn't contribute to. This is unfortunately one area where GitHub could significantly improve their review handling...

amoss:
Perhaps I am struggling with the GitHub UI - let's check. It says that there are two commits:
* One from @Ferroin that changes 7-9 lines.
* One from @prologic that changes 30-40 lines.

Is this an accurate representation because it makes it look like @prologic wrote most of the changes?

Ferroin:
> Perhaps I am struggling with the GitHub UI - let's check. It says that there are two commits:
> 
> * One from @Ferroin that changes 7-9 lines.
> * One from @prologic that changes 30-40 lines.
> 
> Is this an accurate representation because it makes it look like @prologic wrote most of the changes?

Yes, it's accurate. The changes from @prologic are to get the `netdata-installer.sh` script working correctly on macOS. They're needed to test the PR properly there without having to run the installer from `brew sh`. The commit from me is the one that the PR title refers to. I still see value to having him approve that commit, and the reverse as well (though I can't actually approve his since I'm the PR author, and you can't review your own PR's).

amoss:
It makes sense that he approves it to indicate that he is happy with the other commit, but not that he is one of the two reviewers necessary to merge. We should have two "non-authors" read the PR as reviewers.

amoss:
If you rebase and force push I can be the first of those two reviewers...

Ferroin:
> If you rebase and force push I can be the first of those two reviewers...

Done.

prologic:
> It makes sense that he approves it to indicate that he is happy with the other commit, but not that he is one of the two reviewers necessary to merge. We should have two "non-authors" read the PR as reviewers.

I guess you don't want me to re-review this then. That's fine. But I also cannot remove myself as a reviewer on this PR (_the Github UI will not let me :/_). So I will just unsubscribe myself.

PR #8301, opened 2020-03-04
XXX: Test #8291 for blocking PRs with CI checks for YAML

prologic:
See #8291


prologic:
> reviewdog / yamllint (pull_request) Failing after 20s

Good. so that will prevent us from putting up PRs that horribly break YAML files we use for CI/CD in future. Thank you @Ferroin 🙏

PR #8305, opened 2020-03-04
Revert "Fixed Source0 URL in RPM spec (#7794)"

### LINKED ISSUE #8283: RPM versioning is broken.
In short, as of #7794, RPM package builds don't work outside of our own CI workflows.

This comes down to the fact that we end up with a version in the RPM spec file after templating that does not match up with the tag in `git`.

Fixing this requires:

1. Removing the code stripping the `v` from the version number in the tag in our CI workflow for building RPM packages.
2. Reverting #7794.

prologic:
This reverts commit 1477784048319e84127e466add6eb4a092c235e1.

Fixes #8283

##### Summary

As discussed internally we believe we changed something between `v1.18.x` and
`v1.20.x` where we no longer strip the `v` prefix for what evetually becomes
`%{version}) in the `netdata.spec.in`.

- [ ] Prove this

##### Component Name

- area/packaging

##### Test Plan

Evidence

##### Additional Information

See `TBD` above.

### DIFF
--- a/netdata.spec.in
+++ b/netdata.spec.in
@@ -87,7 +87,7 @@ Version:	%{version}
 Release:	1%{?dist}
 License:	GPLv3+
 Group:		Applications/System
-Source0:	https://github.com/netdata/%{name}/releases/download/v%{version}/%{name}-v%{version}.tar.gz
+Source0:	https://github.com/netdata/%{name}/releases/download/%{version}/%{name}-%{version}.tar.gz
 URL:		http://my-netdata.io
 
 # #####################################################################


prologic:
Ultimately; when we build the RPM packages `version` is set as:

```sh
$ sift PACKAGE_VERSION
.
.
.
netdata.spec.in:3:%global version @PACKAGE_VERSION@
```

Which itself is derived from the following script:

```sh
$ sift PACKAGE_VERSION

.

.
contrib/rhel/build-netdata-rpm.sh:16:version=$(grep PACKAGE_VERSION < config.h | cut -d '"' -f 2)
```

Which results (_after a `autoreconf && ./configure`_) of:

```sh
$ version=$(grep PACKAGE_VERSION < config.h | cut -d '"' -f 2); echo "${version}"
v1.20.0-86-ge51ea95e
```

And... `config.h` is auto-generated by the configure script apparently (cc @Ferroin)

PR #8310, opened 2020-03-05
Updated PR labeling to utilize new area/build label.

Ferroin:
##### Summary

This updates the configuration used for labeling PR's to utilize the newly added `area/build` label for PR's that change the build system.

##### Component Name

area/ci

Verified configuration against what files are actually part of the build system and ensured that the file parses correctly.

##### Additional Information

I've added @netdata/core for review on this one so that someone there can help verify that the split in the config as to what's `area/build` and what's `area/packaging` is correct.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -19,6 +19,8 @@ area/backends:
 area/build:
   - build/*
   - build/**/*
+  - build_external/*
+  - build_external/**/*
   - CMakeLists.txt
   - configure.ac
   - Makefile.am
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -16,6 +16,14 @@ area/backends:
   - exporting/*
   - exporting/**/*
 
+area/build:
+  - build/*
```
amoss: Please add everything under `build_external`

Ferroin: Ah, yes, I forgot about that.

Issue #8279, opened 2020-03-04
Fix outstanding issues on the ACLK implementation


amoss:
#### Summary

This is a list of outstanding implementation issues for tracking.

amoss:
Things that have not been converted to issues yet (depends on whether they are fixed as part of ongoing work or handled separately):
* Memory leaks, e.g. https://github.com/netdata/netdata/pull/8167#issuecomment-590064204
* The uninitialised variables inside the JSON parser that valgrind is complaining about.
* On a reconnect `mqtt_init` is called again and causes a memory-leak.
* The LWS event-loop returns instantly in the time after initialization, but before connection. This floods the library with calls and breaks something. A 100ms delay has been added as a work-around but must be fixed properly before we do load-testing.
* All queries are ignored during pop-corning, but commands from the cloud are in the same queue for processing so the agent would ignore requests during this period.
* Any query resets the pop-corning timer (including encapsulated queries) so pop-corning can go on indefinitely.
* The implementation of the time-out / check is wrong and it will always sleep 10s longer than necessary.

amoss:
There is nothing left in this epic - I reassigned the documentation issue to the cloud blockers epic.

PR #8317, opened 2020-03-05
Implement the ACLK Challenge-Response Authentication

### LINKED ISSUE #7738: ACLK - Complete the access to the authentication endpoint
#### Feature summary

- Retrieve the private key that was stored to the disk during claiming.
- The cloud endpoints have a real certificate that be validated from the system-level trust roots.
- First request is to the new cloud endpoint for the challenge, receiving the challenge in a JSON structure.
- Extract the challenge string and decrypt it using the RSA private key.
- Second request delivers the  challenge to the second endpoint and receives user/password in a JSON structure.
- The username and password are set in the aclk thread.
- Need to establish how to test that the correct username/password are passed to the broker.

#### Expected output

- We need the endpoints from the cloud team.
- Test against either VerneMQ or Mosquitto to establish the correct credentials are passed.

amoss:
##### Summary
Closes #7738 

##### Component Name
ACLK

Extensive testing has been performed during the smoke-test sessions, and then again locally. The review should not try to test this PR - setting up the test-configuration is not a good investment of time when it was done in smoke-testing, this will be tested again extensively on the real test system.

##### Additional Information
Please perform the highest-quality review that you can within a limited amount of time (2-3 hours?). There is a list of known issues in the ACLK that this PR does not fix (#8279). Because functional correctness is being handled via separate testing please focus on code quality / safety.

### DIFF
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -349,6 +349,8 @@ static const char *aclk_lws_callback_name(enum lws_callback_reasons reason)
             return "LWS_CALLBACK_WSI_DESTROY";
         case LWS_CALLBACK_CLIENT_ESTABLISHED:
             return "LWS_CALLBACK_CLIENT_ESTABLISHED";
+        case LWS_CALLBACK_OPENSSL_PERFORM_SERVER_CERT_VERIFICATION:
+            return "LWS_CALLBACK_OPENSSL_PERFORM_SERVER_CERT_VERIFICATION";
         default:
             // Not using an internal buffer here for thread-safety with unknown calling context.
             error("Unknown LWS callback %u", reason);
@@ -403,6 +405,7 @@ static int aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reas
         case LWS_CALLBACK_OPENSSL_LOAD_EXTRA_CLIENT_VERIFY_CERTS:
         case LWS_CALLBACK_GET_THREAD_ID: // ?
         case LWS_CALLBACK_EVENT_WAIT_CANCELLED:
+        case LWS_CALLBACK_OPENSSL_PERFORM_SERVER_CERT_VERIFICATION:
             // Expected and safe to ignore.
             debug(D_ACLK, "Ignoring expected callback from LWS: %s", aclk_lws_callback_name(reason));
             return retval;
@@ -444,7 +447,7 @@ static int aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reas
             break;
 
         default:
-            error("Unexecpted callback from libwebsockets %s", aclk_lws_callback_name(reason));
+            error("Unexpected callback from libwebsockets %s", aclk_lws_callback_name(reason));
             break;
     }
     return retval; //0-OK, other connection should be closed!
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -309,6 +309,7 @@ int aclk_lws_wss_connect(char *host, int port)
 
 #ifdef ACLK_SSL_ALLOW_SELF_SIGNED
     i.ssl_connection = LCCSCF_USE_SSL | LCCSCF_ALLOW_SELFSIGNED | LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK;
+    info("Disabling SSL certificate checks");
 #else
     i.ssl_connection = LCCSCF_USE_SSL;
 #endif
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -214,9 +214,6 @@ static inline int check_socks_enviroment(const char **proxy) {
     if(!tmp)
         return 1;
 
-    if(strnlen(tmp, ACLK_PROXY_MAXLEN) == ACLK_PROXY_MAXLEN)
-        return 1;
-
     if(aclk_verify_proxy(tmp) == PROXY_TYPE_SOCKS5) {
         *proxy = tmp;
         return 0;
@@ -238,9 +235,6 @@ static const char *aclk_lws_wss_get_proxy_setting(ACLK_PROXY_TYPE *type) {
             *type = PROXY_TYPE_SOCKS5;
         return proxy;
     }
-
-    if(strnlen(proxy, ACLK_PROXY_MAXLEN) == ACLK_PROXY_MAXLEN)
-        return proxy;
 
     *type = aclk_verify_proxy(proxy);
     if(*type == PROXY_TYPE_UNKNOWN) {
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -906,7 +906,7 @@ char *send_https_request(char *host, char *port, char *url, BUFFER *b, char *pay
         url, host);
     if (payload != NULL)
         buffer_strcat(b, payload);      // TODO Content-length ?
-    int sock = connect_to_this_ip46(IPPROTO_TCP, SOCK_STREAM, host, 0, "443", &timeout);
+    int sock = connect_to_this_ip46(IPPROTO_TCP, SOCK_STREAM, host, 0, port, &timeout);
 
     if (unlikely(sock == -1)) {
         error("Handshake failed");
@@ -1015,7 +1015,7 @@ void aclk_get_challenge()
     char url[1024];
     sprintf(url, "/api/v1/auth/node/%s/challenge", agent_id);
     info("Retrieving challenge from cloud: %s", url);
-    send_https_request("traefik", "443", url, b, NULL);
+    send_https_request("localhost", "8443", url, b, NULL);
     info("Challenge response from cloud: %s", b->buffer);
     // {"challenge":"BfsCcoS16WfrX+t0sP3sEE1p9PnSEIYqXuSSzpqQ/H+du5TZFM8bFHvsdWDvqrW2vnanBUNmeZdjxAAu8cDuIxbGCVc8WiyPeTE4WiLeZnycVHi6B81vW38Lh/KrgJdtfewlh5e434ey4onp9UBdCJy9sjrSQZR6yEj0rB4ilvKjuyV2gJOysx6EVU5VBIfOphf/QBIiYroPmUL5WM0E6Re1g6P0au+Tb1N08kwbmOnY7VWk3/cqVvf0S9iV80Yrt69nqWXMl65cu9y9L4XZ4b7fi82Z7nwRIJYyHse8LAgUzraFGz3Z84Po3dnOaouvSQhY52AuwHpfojet+knXSg=="}
     struct dictionary_singleton challenge = { .key = "challenge", .result = NULL };
@@ -1062,7 +1062,7 @@ void aclk_get_challenge()
     info("Password phase: %s",response_json);
     // TODO - host
     sprintf(url, "/api/v1/auth/node/%s/password", agent_id);
-    send_https_request("traefik", "443", url, b, response_json);
+    send_https_request("localhost", "443", url, b, response_json);
     info("Password response from cloud: %s", b->buffer);
 
     struct dictionary_singleton password = { .key = "password", .result = NULL };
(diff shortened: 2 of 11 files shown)


aclk/aclk_lws_wss_client.c
```diff
@@ -315,7 +319,7 @@ static int aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reas
             break;
 
         default:
-            error("Unexecpted callback from libwebsockets %s", aclk_lws_callback_name(reason));
+            error("Unexpected callback from libwebsockets %s", aclk_lws_callback_name(reason));
```
underhood: I am not sure we should log every callback even as LWS library grows callbacks can be added that we are simply not interested in. But up to you

amoss: That is a fair point. My reasoning is as follows: we need to be sure about every callback that is generated by an external dependency. In a sense this part of the code is there to enforce white-listing of those events, i.e. one of us has sat and read the docs and made an explicit effort to put it onto the "don't care" list. If the library changes in the future and new callbacks are generated then somebody needs to perform that explicit step and this logging will force that to happen.

aclk/agent_cloud_link.c
```diff
@@ -131,6 +133,57 @@ int cloud_to_agent_parse(JSON_ENTRY *e)
...
 
 
+static RSA *aclk_private_key = NULL;
+static int create_private_key()
+{
+    char filename[FILENAME_MAX + 1];    struct stat statbuf;
+    snprintfz(filename, FILENAME_MAX, "%s/claim.d/private.pem", netdata_configured_user_config_dir);
+
+    if (lstat(filename, &statbuf) != 0) {
+        error("Claimed agent cannot establish ACLK - private key not found '%s' failed reason=\"%s\".", filename, strerror(errno));
```
mfundul: The `error()` netdata function already prints `strerror(errno)` so it's going to print this twice.

aclk/agent_cloud_link.c
```diff
@@ -131,6 +133,57 @@ int cloud_to_agent_parse(JSON_ENTRY *e)
...
+    aclk_private_key = PEM_read_bio_RSAPrivateKey(key_bio, NULL, NULL, NULL);
+    BIO_free(key_bio);
+    if (aclk_private_key!=NULL)
+    {
+        freez(private_key);
+        return 0;
+    }
+    char * err = mallocz(130);;
+    ERR_load_crypto_strings();
+    ERR_error_string(ERR_get_error(), err);
```
mfundul: From the manpage "buf must be at least 256 bytes long.", this can cause memory corruption. Maybe use `ERR_error_string_n()` instead or remove the magic number 130?

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+        buffer_strcat(b, payload);
+    debug(D_ACLK, "Sending HTTPS req (%zu bytes): '%s'", b->len, buffer_tostring(b));
+    int sock = connect_to_this_ip46(IPPROTO_TCP, SOCK_STREAM, host, 0, port, &timeout);
+
+    if (unlikely(sock == -1)) {
+        error("Handshake failed");
+        return 1;
+    }
+
+    SSL_CTX *ctx = security_initialize_openssl_client();
```
mfundul: check for NULL?

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+    // Certificate chain: not updating the stores - do we need private CA roots?
+    // Calls to SSL_CTX_load_verify_locations would go here.
+    SSL *ssl = SSL_new(ctx);
```
mfundul: Check for NULL.

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+    SSL_set_fd(ssl, sock);
```
mfundul: Returns 0 on failure.

aclk/agent_cloud_link.c
```diff
@@ -131,6 +133,57 @@ int cloud_to_agent_parse(JSON_ENTRY *e)
...
+        return 1;
+    }
+    if (unlikely(statbuf.st_size == 0)) {
+        info("Claimed agent cannot establish ACLK - private key '%s' is empty.", filename);
+        return 1;
+    }
+
+    FILE *f = fopen(filename, "rt");
+    if (unlikely(f == NULL)) {
+        error("Claimed agent cannot establish ACLK - unable to open private key '%s' failed reason=\"%s\".", filename, strerror(errno));
```
mfundul: The error() netdata function already prints strerror(errno) so it's going to print this twice.

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+    if (err <= 0)
+    {
+        error("SSL_write() failed with err=%d", err);
+        SSL_free(ssl);
+        close(sock);
+        SSL_CTX_free(ctx);
+        return 1;
+    }
+    buffer_flush(b);
+    int bytes_read = SSL_read(ssl, b->buffer, b->size);
```
mfundul: On error `SSL-read()` can set `bytes_read` to a negative value.

aclk/agent_cloud_link.c
amoss: Yikes - I copied that from `security.c` so we have the same problem on the TLS in the web server.

amoss: Will fix.

aclk/agent_cloud_link.c
amoss: Will check.

aclk/agent_cloud_link.c
amoss: Will fix.

aclk/agent_cloud_link.c
amoss: Will fix.

aclk/agent_cloud_link.c
amoss: Will fix.

amoss: This also came from the TLS code - please check how this is handled for incoming https streams from slaves.

aclk/agent_cloud_link.c
amoss: Will fix.

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+}
+
+// Base-64 decoder.
+// Note: This is non-validating, invalid input will be decoded without an error.
+//       Challenges are packed into json strings so we don't skip newlines.
+//       Size errors (i.e. invalid input size or insufficient output space) are caught.
+size_t base64_decode(unsigned char *input, size_t input_size, unsigned char *output, size_t output_size)
+{
+    static char lookup[256];
+    static int first_time=1;
```
mfundul: I guess it's only being called by one thread?

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+        lookup['+'] = 62;
+        lookup['/'] = 63;
+    }
+    if ((input_size & 3) != 0)
+    {
+        error("Can't decode base-64 input length %zu", input_size);
+        return 0;
+    }
+    size_t unpadded_size = (input_size/4) * 3;
+    if ( (input_size/4) * 3 > output_size )
```
mfundul: You can use `unpadded_size` here.

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+    {
+        error("Output buffer size %zu is too small to decode %zu into", output_size, input_size);
+        return 0;
+    }
+    // Don't check padding within full quantums
+    for (size_t i = 0 ; i < input_size-4 ; i+=4 )
+    {
+        uint32_t value = (lookup[input[0]] << 18) + (lookup[input[1]] << 12) + (lookup[input[2]] << 6) + lookup[input[3]];
```
mfundul: Shouldn't those things be `input[i]`, `input[i+1]`, `input[i+2]` or something?

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+    }
+}
+
+size_t base64_encode(unsigned char *input, size_t input_size, char *output, size_t output_size)
+{
+    uint32_t value;
+    static char lookup[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                           "abcdefghijklmnopqrstuvwxyz"
+                           "0123456789+/";
+    if ((input_size/3+3)*4 >= output_size)
```
mfundul: I think this should either be:
```
    if ((input_size/3+1)*4 >= output_size)
```
or
```
    if (((input_size+3)/3)*4 >= output_size)
```

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+
+
+
+int private_decrypt(unsigned char * enc_data, int data_len, unsigned char *decrypted)
+{
+    int  result = RSA_private_decrypt( data_len, enc_data, decrypted, aclk_private_key, RSA_PKCS1_OAEP_PADDING);
+    if (result == -1) {
+        char * err = malloc(130);
+        ERR_load_crypto_strings();
+        ERR_error_string(ERR_get_error(), err);
```
mfundul: This is the same problem as above. Moreover, `ERR_load_crypto_strings()` should only be called once per programme I think. Also, it's deprecated in OpenSSL 1.1.0

aclk/agent_cloud_link.c
```diff
@@ -893,6 +946,402 @@ static void aclk_main_cleanup(void *ptr)
...
+    }
+
+    if (password.result == NULL ) {
+        error("Could not retrieve password from auth response");
+        return;
+    }
+    if (aclk_password != NULL )
+        freez(aclk_password);
+    aclk_username = strdupz(agent_id);
+    aclk_password = strdupz(password.result);
```
mfundul: Wasn't `password.result` already `strudupz`ed?

aclk/agent_cloud_link.c
amoss: The initialization is idempotent but this is an interesting comment about the common idiom that we use all over the code.

aclk/agent_cloud_link.c
amoss: Fixed.

aclk/agent_cloud_link.c
amoss: Fixed.

aclk/agent_cloud_link.c
amoss: I've misremember the code in `security.c` - it implements the check in the way that you suggest. I'll update this to do the same.

aclk/agent_cloud_link.c
amoss: Ah ha! But the TLS code handled it better. I've changed the logic in the function into a more standard set of labels/gotos to make sure the cleanup is handled correctly (and found an extra case that was not closing the socket properly).

aclk/agent_cloud_link.c
amoss: Ah yes, that was why I calculated it :) Fixed.

aclk/agent_cloud_link.c
amoss: Well spotted. Fixed.

aclk/agent_cloud_link.c
amoss: All true, I've removed the calls as we call it from elsewhere in the case that `OPENSSL_API_COMPAT<0x10100000L`.

aclk/agent_cloud_link.c
amoss: Fixed.

Issue #8316, opened 2020-03-05
Switch `install-required-packages.sh` to use Python 3 by default.


Ferroin:
Python 2 was officially EOL as of 2020-01-01. The final bugfix release is comming out in April, but will have no 'new' fixes other than those finished before 2020-01-01.

All major distributions we officially support have at least _some_ version of Python 3 in their most recent release, and some will actually be removing Python 2 support in their next release (at least Debian 11 and Ubuntu 20.04 will not have Python 2).

Given this and the fact that only one distribution that we officially support does not have Python 3.x (CentOS 7, which only includes Python 2.7), it is simpler from the perspective of the installer to switch from using Python 2 as the default version to Python 3.

Note that this is _not_ a proposal to remove Python 2.7 support. We should consider doing that eventually as well, as it will become more and more difficult to test on Python 2, but that is still a ways off for the time being.

The relevant changes required to do this constitute approximately 6 lines of code in `install-required-packages.sh` wth near zero maintenance overhead. On the other hand, special-casing those distributions that do not provide Python 2 will require significantly more changes and incur a persistent maintenance overhead going forwards as more and more distributions drop support for Python 2.

Issue #7364, opened 2019-11-25
Install netdata with Python3


AbdelmalekIhdene:
##### Question summary
I run the installation command `bash <(curl -Ss https://my-netdata.io/kickstart.sh)` and it prompts me to install python2.7 and successfully installs. However, I was wondering if it was possible to install netdata with python3 instead of python2.7.

##### OS / Environment
Ubuntu 18.04 LTS

ncmans:
Thanks @AbdelmalekIhdene for the report.

netdata can be installed with python 3 as well. However, the default python package on Ubuntu 18.04 LTS is python 2 and I've confirmed that it does indeed try to install `python` package even if `python3` is already installed.

I do think that netdata should pick any compatible python already installed and use that.

Ferroin:
We can check whether `python` or `python3` are installed using the following one-liner on any dpkg based system (including Ubuntu, Debian, and Devuan):

    dpkg --list-selections | grep -E -q 'python3?\s*install'

That will return an exit status of 0 if one of the packages with the exact names `python` or `python3` is installed, or 1 if neither is installed (and it doesn't need root privileges, but that's not really relevant here).

I'm of the opinion that we should probably prefer Python 3.x over Python 2.x on install if neither is installed, as it's a bit more efficient, though it's unlikely that we'll come across a system that doesn't have one of the two installed (every dpkg-based distro I know of has one or the other as part of the default install).

AbdelmalekIhdene:
Does netdata check for the existence of python2 or just for the existence of python? A quick fix for now, I believe, would be `alias python=python3`.

ncmans:
> Does netdata check for the existence of python2 or just for the existence of python? A quick fix for now, I believe, would be `alias python=python3`.

Installer checks if a certain package is installed by querying the package manager, and so aliasing python will have no effect in this case.

IceCodeNew:
```
update-alternatives --install /usr/bin/python python /usr/bin/python2 1
update-alternatives --install /usr/bin/python python /usr/bin/python3 2
```

I used to have the same problem and I find out these commands could solve this problem for good.

PR #8318, opened 2020-03-05
Switched to using Python 3 by default instead of Python 2.

Ferroin:
##### Summary

Most distros have Python 3, and some are already looking at dropping Python 2 in upcomming releases. All of our code works correctly on Python 3, so just use it by default for platforms that have it.

##### Component Name

area/packaging

Verified that package installation still works correctly on all platforms we officially support.

##### Additional Information

Fixes: #8316 
Fixes: #7364


prologic:
Does this also address #7364?

PR #8321, opened 2020-03-05
Migrate coverity-scan to Github Actions

### LINKED ISSUE #6749: netdata/ci: Wrap up coverity script refinement, make it usable, extend CI pipeline compile checks

prologic:
##### Summary

Fixes #6749 

##### Component Name

- area/ci

##### Test Plan

- Test in my fork with temporary Github secrets added to my fork.

##### Additional Information

I need the following secrets / enviornment variables currently in Travis:

- `COVERITY_SCAN_SUBMIT_MAIL`
- `COVERITY_SCAN_TOKEN`

If someone can get those to me (_preferably @knatsakis_) and added to our SRE secrets repo that would be awesome :)

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -6,17 +6,16 @@ on:
     - cron: '0 1 * * *'
 jobs:
   coverity:
+    if: github.repository == 'netdata/netdata'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-        if: github.repository == 'netdata/netdata'
       - name: Prepare environment
-        if: github.repository == 'netdata/netdata'
         run: |
-          bash <(curl -sS https://raw.githubusercontent.com/netdata/netdata/master/packaging/installer/install-required-packages.sh) --dont-wait --non-interactive netdata
-          sudo apt-get install -y libjson-c-dev libipmimonitoring-dev libcups2-dev libsnappy-dev libprotobuf-dev libprotoc-dev libssl-dev protobuf-compiler
+          ./packaging/installer/install-required-packages.sh --dont-wait --non-interactive netdata
+          sudo apt-get install -y libjson-c-dev libipmimonitoring-dev libcups2-dev libsnappy-dev \
+                                  libprotobuf-dev libprotoc-dev libssl-dev protobuf-compiler
       - name: Run coverity-scan
-        if: github.repository == 'netdata/netdata'
         run: |
           ./coverity-scan.sh --with-install
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,22 @@
+---
+# Runs coverity-scan.sh every 24h on `master`
+name: Coverity Scan
+on:
+  schedule:
+    - cron: '0 1 * * *'
+jobs:
+  coverity:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        if: github.repository == 'netdata/netdata'
+      - name: Prepare environment
+        if: github.repository == 'netdata/netdata'
+        run: |
+          bash <(curl -sS https://raw.githubusercontent.com/netdata/netdata/master/packaging/installer/install-required-packages.sh) --dont-wait --non-interactive netdata
+          sudo apt-get install -y libjson-c-dev libipmimonitoring-dev libcups2-dev libsnappy-dev libprotobuf-dev libprotoc-dev libssl-dev protobuf-compiler
+      - name: Run coverity-scan
+        if: github.repository == 'netdata/netdata'
+        run: |
+          ./coverity-scan.sh --with-install
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -17,5 +17,8 @@ jobs:
           sudo apt-get install -y libjson-c-dev libipmimonitoring-dev libcups2-dev libsnappy-dev \
                                   libprotobuf-dev libprotoc-dev libssl-dev protobuf-compiler
       - name: Run coverity-scan
+        env:
+          COVERITY_SCAN_TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
+          COVERITY_SCAN_SUBMIT_MAIL: ${{ secrets.COVERITY_SCAN_SUBMIT_MAIL }}
         run: |
           ./coverity-scan.sh --with-install
--- a/.travis.yml
+++ b/.travis.yml
@@ -485,15 +485,6 @@ jobs:
     # This is the nightly pre-execution step (Jobs, preparatory steps for nightly, etc)
     - stage: Nightly operations
 
-      name: Run coverity scan
-      before_script:
-        - bash <(curl -sS https://raw.githubusercontent.com/netdata/netdata/master/packaging/installer/install-required-packages.sh) --dont-wait --non-interactive netdata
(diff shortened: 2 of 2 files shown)


.github/workflows/coverity.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  schedule:
+    - cron: '0 1 * * *'
+jobs:
+  coverity:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        if: github.repository == 'netdata/netdata'
```
Ferroin: We can actually put this on the job itself instead of on each individual stage. I'm fairly certain that that will result in the job not even getting queued instead of it running and skipping each step individually.

prologic: Good point :)

prologic: Done

knatsakis:
> If someone can get those to me (_preferably @knatsakis_) and added to our SRE secrets repo that would be awesome :)

But, they are already there :-)

prologic:
> But, they are already there :-)

They are? Okay 😲 Clearly I don't `gopass sync` every often :) Lemme check!

prologic:
> Since a VM is going to be used for running this, I think the same VM should be used for all static analyzers. @Ferroin @prologic

I don't understand @cosmix :) We don't actually reall have any control over this per se unless we go to the effort of having our own self-hosted runners for _everything_ -- Which we plan to anyway for ARM* and BSD*

prologic:
> > But, they are already there :-)
> 
> They are? Okay 😲 Clearly I don't `gopass sync` every often :) Lemme check!

```sh
$ gopass ls
gopass
├── github.com
│   └── netdatabot
└── misc
    └── travis_netdata_netdata
```

Hmmm 🤔 @knatsakis I need those env vars from Travis; Pretty sure I can't copy/paste them out? :)

knatsakis:
> > > But, they are already there :-)
> > 
> > 
> > They are? Okay astonished Clearly I don't `gopass sync` every often :) Lemme check!
> 
> ```shell
> $ gopass ls
> gopass
> ├── github.com
> │   └── netdatabot
> └── misc
>     └── travis_netdata_netdata
> ```
> 
> Hmmm  @knatsakis I need those env vars from Travis; Pretty sure I can't copy/paste them out? :)

Will DM you about this.

thiagoftsm:
Continue the talk we had at Slack, I do not see any problem to change the email address that is receiving the messages, it is a good idea to set an email that all the core team can receive the information, on the other hand, I do not think we can remove the current email, so we need to find an email that has the actual email + core team email. @cakrit do we have an email with these people?

prologic:
Need review/approvals on this now with the secrets properly inserted into Github secrets and referenced in the workflow.

prologic:
The contents of the secret can be updated _any_ time.

cosmix:
> I don't understand @cosmix :) We don't actually reall have any control over this per se unless we go to the effort of having our own self-hosted runners for everything -- Which we plan to anyway for ARM* and BSD*

Maybe I wasn’t clear. You don’t need your own runner for this. What I’m saying is that when you specify a virtual environment on gha you essentially specify a VM. And ideally we

cosmix:
Closed by mistake.

prologic:
> Maybe I wasn’t clear. You shouldn’t need your own runner for this, I’ll take another look at the docs. What I’m saying is that when you specify a virtual environment on gha you essentially specify a VM. And ideally we should be using the same VM to run all static analyzers, if possible. I think the tricky part would be that of the presentation of the results only, as you can trivially just add another step, one for each analyzer.

Yes I see what you're saying... The only way to do this is to ensure all static analysis runs in the same job. However this has its own pros/cons; where this makes sense we will do this of course, but where we want parallel jobs we will split out into separate job(s). It also helps for clarity in the Github UI on PR(s) too if CI/CD "things" are split out in nice meaningful ways.

I guess we run the risk of Microsoft/Github getting a bit upset with us for utilising Azure as much as we do but then we are limited by the default limits anyway so I think we're okay :)

prologic:
Also I should mention just for completeness, it gets a little tricky when you want to "share" things across jobs such as what happens in matrix builds like in our kernel-collector since each matrix strategy results in a new job. Artefacts have to be shared across jobs via Github's Artefact store and essentially uploaded from one job and downloaded from another. :D

Ferroin:
> > I don't understand @cosmix :) We don't actually reall have any control over this per se unless we go to the effort of having our own self-hosted runners for everything -- Which we plan to anyway for ARM* and BSD*
> 
> Maybe I wasn’t clear. You shouldn’t need your own runner for this, I’ll take another look at the docs. What I’m saying is that when you specify a virtual environment on gha you essentially specify a VM. And ideally we should be using the same VM to run all static analyzers, if possible. I think the tricky part would be that of the presentation of the results only, as you can trivially just add another step, one for each analyzer.

@cosmix Can you please elaborate on why you feel that we need to use the same VM for all the static analyzers?

Doing that won't really gain us anything other than cutting down on how many things we might have queued at a time for GHA, and even that's not a huge issue here because any of the ones we're running are either infrequent (like Coverity), fast, or both, so they're not likely to cause any significant delays/

If you just want them to bail on the first failure and cancel the others, then that's a non-issue as that's the default behaviour for parallel jobs in the same workflow in GitHub Actions (it can be changed to behave like Travis instead, but you have to go out of your way to do it).

cosmix:
@Ferroin efficiency. The VM would be reused by all analyzers, working on the same RO dataset (the source code). It's faster and much less resource intensive. 

As I wrote, the main issue I'm seeing is communicating/rendering the results, not bailing out when one fails (the are independent from each other, i.e. I'd like the output of all of them). Either way, although I'd much prefer if we could just reuse an existing VM, it does go against the norms here so let's just ignore this and move on.

prologic:
> If you just want them to bail on the first failure and cancel the others, then that's a non-issue as that's the default behaviour for parallel jobs in the same workflow in GitHub Actions (it can be changed to behave like Travis instead, but you have to go out of your way to do it).

👍 unless we turn it off for some __good__ reason :)

prologic:
> @Ferroin efficiency. The VM would be reused by all analyzers, working on the same RO dataset (the source code). It's faster and much less resource intensive.

Sadly while this is true we can't really do this and we just loose out on the efficiency gains from parallelism. I'm just as surprised as anyone but the Microsoft Azure cloud is really insanely fast at spinning up VMs and their core infrastructure is actually really performant!

prologic:
> As I wrote, the main issue I'm seeing is communicating/rendering the results, not bailing out when one fails (the are independent from each other, i.e. I'd like the output of all of them). Either way, although I'd much prefer if we could just reuse an existing VM, it does go against the norms here so let's just ignore this and move on.

As I've mentioned before we __will__ do this where it makes sense; but to your own point I _think_ (_and I could be wrong here-) it actually makes things more clear and easier to find / spot check if we split things out in a logical way. For example (_and I should document this_) we are following a convention of:

- <workflow> / <job> (<step>)

So all the Installer CI checks follow the pattern of:

- Installer / Install (debian:10 ...)

For example.

Same for the recently merged consistency checks.

Hope this makes sense ;)

PR #8323, opened 2020-03-06
New kernels for netdata installer

### LINKED ISSUE #8322: Update kernels on netdata-installer
##### Feature idea summary
It is necessary to bring the new kernels present on `netdata/kernel-collector` to `netdata-installer`, it is also necessary to create a new release with these kernels.

##### Expected behavior
To install eBPF collector on 13 different kernel versions.

### LINKED ISSUE #8621: Redhat 7.8 and CentOS 7 ( eBPF support)
##### Feature idea summary
Red Hat released Red Hat 7.8 recently, it is necessary to verify whether we are supporting this new release, considering that the kernel name changed.

We also need to bring support for CentOS 7 as described on #8322.
##### Expected behavior
To run our eBPF collectors on Red Hat 7.8

thiagoftsm:
##### Summary
Fixes #8621

This PR brings to installer the latest kernels present in kernel-collector repository.
This is the last time that this function will be updated like it is, we need to create another algorithm for it.
##### Component Name
Installer
We need to create a new release to test this PR properly.

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -1255,46 +1255,72 @@ get_compatible_kernel_for_ebpf() {
 
   kpkg=
 
-  if [ "${rhver}" -ge 0080000000  ] ; then
+  # Kernel variables
+  rhver8="0080000000"
+  rhver7="0070061810"
+  kernel_5_0_0="005000000"
+  kernel_4_20_17="004020017"
+  kernel_4_20_0="004020000"
+  kernel_4_19_108="004019108"
+  kernel_4_19_105="004019105"
+  kernel_4_19_104="004019104"
+  kernel_4_19_102="004019102"
+  kernel_4_19_101="004019101"
+  kernel_4_19_98="004019098"
+  kernel_4_19_97="004019097"
+  kernel_4_19_6="004019006"
+  kernel_4_18_20="004018020"
+  kernel_4_18_17="004018017"
+  kernel_4_18_16="004018016"
+  kernel_4_18_0="004018000"
+  kernel_4_16_18="004016018"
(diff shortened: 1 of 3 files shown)


Ferroin:
@thiagoftsm If you rebase this onto the current master, it should fix the Travis CI checks.

prologic:
@thiagoftsm I _could_ be wrong here but the think the eBpf package selection logic here is wrong now; since we should be picking `glubc_centos7` or `glibc_centos8`

But maybe we can fix what we publish instead? In which case the package would be called 3.10.0-xxx` something?

Screen Shot 2020-03-18 at 08 06 25

thiagoftsm:
> Can we remove all magic numbers and have something like `VARIABLE`, `VARIABLE - 1` so that the code is more readable and maintainable?

I will do the changes on C code, and I will talk with @prologic about the shell script code.

netdata-installer.sh
```diff
@@ -1115,6 +1115,19 @@ function get_kernel_version() {
   printf "%03d%03d%03d" "${p[0]}" "${p[1]}" "${p[2]}"
 }
 
+function get_rh_version() {
+  if [ ! -f /etc/redhat-release ] ; then
+     echo "000000000"
```
prologic:
```suggestion
     printf "000000000"
```

thiagoftsm: Thank you I changed the PR changing echo to printf.

Ferroin:
Closing, as this has been superseded by and will be merged as part of #8665.

thiagoftsm:
Thank you @Ferroin !

PR #8329, opened 2020-03-06
Migrate Travis based checks to Github Actions

prologic:
##### Summary

As per title

##### Component Name

- area/ci

##### Test Plan

- Test in my fork

##### Additional Information


.github/workflows/checks.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Git clone repository
+        uses: actions/checkout@v2
+      - name: Run checksum checks on kickstart files
+        env:
+          LOCAL_ONLY: "true"
+        run: |
+          ./tests/installer/checksums.sh
+      - name: Web Dashboard pre-generated file consistency checks (dashboard.js)
+        run: |
+          cp web/gui/dashboard.js /tmp/dashboard.js && ./build/build.sh && diff /tmp/dashboard.js web/gui/dashboard.js
```
Ferroin: It may be possible to simplify this and make it more reliable. Take a look at the actual contents of `build/build.sh`, most of what this actually needs is just `autoreconf -ivf && ./configure && make dist`.

prologic: Done [see logs](https://github.com/netdata/netdata/pull/8329/checks?check_run_id=489296788)

.github/workflows/checks.yml
```diff
@@ -0,0 +1,29 @@
...
+          LOCAL_ONLY: "true"
+        run: |
+          ./tests/installer/checksums.sh
+      - name: Web Dashboard pre-generated file consistency checks (dashboard.js)
+        run: |
+          cp web/gui/dashboard.js /tmp/dashboard.js
+          autoreconf -ivf
+          ./configure --enable-maintainer-mode
+          make dist
+          diff -sNrdu /tmp/dashboard.js web/gui/dashboard.js
```
Ferroin: I had not thought of this originally, but these should probably be separate jobs, not separate steps of a single job. They're orthogonal to each other, and they run in parallel in Travis right now (jobs in Travis run in parallel, it's stages that are serialized).

prologic: Yup okay I _thought_ about doing this but I was in split minds about it.

prologic: Done

PR #8334, opened 2020-03-06
Fixed dependency names for Arch Linux.

Ferroin:
##### Summary

This adds specific overrides for a couple of packages on Arch Linux that we were trying to install using the wrong names.

This also switches the lifecycle tests to use the `install-required-packages.sh` script from the branch being tested instead of always pulling from `master`, which will help prevent issues like this from happening again in the future.

##### Component Name

area/packaging

Verified the script on a local Arch VM.

##### Additional Information

This fixes the recurring lifecycle failures for Arch in CI. I'm not sure why they were working before.

### DIFF
--- a/packaging/installer/install-required-packages.sh
+++ b/packaging/installer/install-required-packages.sh
@@ -948,6 +948,7 @@ declare -A pkg_lz4=(
   ['suse']="liblz4-devel"
   ['gentoo']="app-arch/lz4"
   ['clearlinux']="devpkg-lz4"
+  ['arch']="lz4"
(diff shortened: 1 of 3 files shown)


Ferroin:
Travis failures are expected, because the ifecycle checks are braindead (they _always_ end up using the `install-required-packages.sh` from the `master` branch, because they use `kickstart.sh` during the install).

Ferroin:
Merging this, as it fixes urgent CI issues and the only other person qualified to review it won't likely be able to do so for another 50 hours or so.

PR #8347, opened 2020-03-09
Update paragraph on install-required-packages

prologic:
##### Summary

This tool is very much __NOT__ experimental, so updating it to reflect current reality.

##### Component Name

- area/docs

##### Test Plan

- Netlify

##### Additional Information

### DIFF
--- a/packaging/installer/methods/manual.md
+++ b/packaging/installer/methods/manual.md
@@ -12,9 +12,9 @@ To install the latest git version of Netdata, please follow these 2 steps:
 
 ## Prepare your system
 
-Try our experimental automatic requirements installer (no need to be root). This will try to find the packages that
-should be installed on your system to build and run Netdata. It supports most major Linux distributions released after
-2010:
+Use our automatic requirements installer (_no need to be `root`_). This will try to find the packages that
+should be installed on your system to build and run Netdata. It supports a large variety of major Linux distributions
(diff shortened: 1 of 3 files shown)


prologic:
> 2:02:10 PM:    - ERROR: ./packaging/installer/methods/manual.md - https://github.com/netdata/netdata/blob/master/packaging/installer/install-required-packages.sh is an absolute link to a Netdata file. Please convert to relative.

Hmmm 🤔 @joelhans Is this a new check? I'm open to ideas or just remove this bit. I originally only wanted to remove the wording "experimental".

joelhans:
@prologic This isn't a new check, no. The easiest fix that works for me locally is to just link to the raw version of the file: `https://raw.githubusercontent.com/netdata/netdata/master/packaging/installer/install-required-packages.sh`. That way it doesn't get tripped up in the build script. 

I would say do that fix, or don't link it at all. There's another solution that involves linking to a specific blob, but that means if you update the file the link will reference an old version, which is not good.

packaging/installer/methods/manual.md
```diff
@@ -12,9 +12,9 @@ To install the latest git version of Netdata, please follow these 2 steps:
 
 ## Prepare your system
 
-Try our experimental automatic requirements installer (no need to be root). This will try to find the packages that
-should be installed on your system to build and run Netdata. It supports most major Linux distributions released after
-2010:
+Use our automatic requirements installer (_no need to be `root`_). This will try to find the packages that
```
shortpatti:
```suggestion
Use our automatic requirements installer (_no need to be `root`_), which attempts to find the packages that
```

prologic:
@shortpatti Oh I _think_ we both tried to fix the build error at the same time? :) Thanks! If this works please review/approve so we can merge this

PR #8354, opened 2020-03-10
Bulk add frontmatter to all documentation

joelhans:
##### Summary

This is an automated, bulk addition of frontmatter to all our documentation (`.md`) files. I used a script to create the `title` field via the `# blah blah` heading, and the `custom_edit_url` field via the pathname.

By adding basic frontmatter to all our documentation, we support two efforts:

1. Improve SEO by explicitly declaring page titles
2. Enable the new Docusaurus-powered documentation site with better page titles and `Edit this page on Github` links that work properly

I had wanted to do this in a more nuanced fashion, but we need to move quickly so that this isn't a blocker for both the above efforts. The marketing team has already discussed how we'll improve this existing frontmatter with good descriptions and keywords.

The frontmatter is commented out so that it doesn't appear when viewing the file in GitHub itself. The documentation build process strips these comments out before generating the HTML.

One implication of this effort is that the docs team will need to enforce similar frontmatter on every new document.

##### Component Name

docs
*

I have tested this PR by generating the `mkdocs` site locally, and it works fine. I also tested with our in-progress Docusaurus-powered documentation site, and everything appears to work normally. Finally, I'll test the Netlify deploy preview when it's live.

I will also ask Ian to check this PR against his ingest/sanitize script so we are aware if it creates any issues before this is merged.

### DIFF
--- a/.travis/README.md
+++ b/.travis/README.md
@@ -1,3 +1,10 @@
+<!--
+---
+title: "Description of CI build configuration"
+custom_edit_url: https://github.com/netdata/netdata/edit/master/.travis/README.md
+---
+-->
+
 # Description of CI build configuration
 
 ## Variables needed by travis
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -1,3 +1,10 @@
+<!--
+---
+title: "Breaking Changes"
+custom_edit_url: https://github.com/netdata/netdata/edit/master/BREAKING_CHANGES.md
+---
+-->
+
 # Breaking Changes
 
 -   remove deprecated bash modules (`apache`, `cpu_apps`, `cpufreq`, `exim`, `hddtemp`, `load_average`, `mem_apps`, `mysql`, `nginx`, `phpfpm`, `postfix`, `squid`, `tomcat`) [[#7962](https://github.com/netdata/netdata/pull/7962)]
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,3 +1,10 @@
+<!--
+---
+title: "The build system"
+custom_edit_url: https://github.com/netdata/netdata/edit/master/BUILD.md
+---
+-->
+
 # The build system
 
 We are currently migrating from `autotools` to `CMake` as a build-system. This document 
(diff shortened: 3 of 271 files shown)


prologic:
> @joelhans any plans on how to keep the frontmatter in sync with the page paths and titles? If not, I'd recommend some sort of CI check. ( @Ferroin @prologic )

@joelhans @ncmans is right! I'm not familiar with the tool you used specifically but it if is idempotent (_meaning re-running it just updates the comments and the respective k/v pairs_) then we can automate this entirely in CI so documentation maintainers and contributors need only worry about the `# ...` title itself in the Markdown documents.

You can either help me get this in for you or I can help you create an appropriate Github Actions workflow / either way :)

prologic:
Also I can't review diffs like this so if you amend your "Test Plan" with:

##### Test Plan

```sh
$ path/to/tool/you/ran [with any options]
```

I can verify the same output and approve :)

joelhans:
@ncmans No plans for syncing other than manual intervention.

@prologic I was hoping to keep my shameful Bash script under wraps, but here goes:

```
#!/bin/bash

for f in `find . -type d \( -path docs/generator -o -path ./.github -o -path node_modules -o -name HISTORICAL_CHANGELOG.md -o -path ./CHANGELOG.md -o -path .travis \) -prune -o -name "*.md"`; do
  TITLE=`sed -n '1{/^# .*$/p};q' $f | cut -d '#' -f 2 | sed 's/^ *//g'`
  DOCPATH=${f#*/}

  if [ ! -z "$TITLE" ]
  then
    FM="<!--\n---\ntitle: \"$TITLE\"\ncustom_edit_url: https://github.com/netdata/netdata/edit/master/$DOCPATH\n---\n-->\n"
    awk -i inplace -v x="$FM" 'NR==1{print x} 1' $f
  else
    echo "No title for: $f"
  fi
done
```

It is most definitely not idempotent, although maybe it could be! And could be written a million times better. I'm pretty sure `find` just ignores all those excludes anyway.

prologic:
WOW! Nice pit of shell there :) (_not shameful at all!_)

Yeah okay approved! As its not idempotent its a bit harder to automate. But we can discuss this in a new issue? We can do all sorts of magical things with a Markdown parser and a bit of shell :)

PR #8355, opened 2020-03-10
Remove possible erroneous blank line causing Travis to fail possibly and default to magical behaviour

prologic:


prologic:
Merging with 1 approval as __BREAK FIX__

PR #8365, opened 2020-03-10
Format of commit messages

amoss:
##### Summary
Changing the structure of commit mesages.

##### Component Name
n/a

n/a

##### Additional Information
n/a

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -193,7 +193,7 @@ New collectors must have the title "[New collector] : [Name of collector]".
 
 #### Commit messages when PRs are merged
 
-When a PR is squashed and merged into master the title of the commit message (first line) must be the PR title
+When a PR gets squashed and merged into master, the title of the commit message (first line) must be the PR title
 followed by the PR number.
 
 The body of the commit message should be a short description of the work, preferably taken from the connected issue.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -191,6 +191,13 @@ Improvements must have the title "[Area or module] : [Added/Refactored/Improved/
 
 New collectors must have the title "[New collector] : [Name of collector]".
 
+#### Commit messages when PRs are merged
+
+When a PR is squashed and merged into master the title of the commit message (first line) must be the PR title
+followed by the PR number.
+
+The body of the commit message should be a short description of the work, preferably taken from the connected issue.
+
 ### PR approval process
 
 Each PR automatically [requires a review](https://help.github.com/articles/about-required-reviews-for-pull-requests/) from the code owners specified in `.github/CODEOWNERS`. Depending on the files contained in your PR, several people may be need to approve it.


cakrit:
I would say "preferably taken from the connected issue, or the PR OP". Otherwise it LGTM

CONTRIBUTING.md
```diff
@@ -191,6 +191,13 @@ Improvements must have the title "[Area or module] : [Added/Refactored/Improved/
 
 New collectors must have the title "[New collector] : [Name of collector]".
 
+#### Commit messages when PRs are merged
+
+When a PR is squashed and merged into master the title of the commit message (first line) must be the PR title
```
shortpatti:
```suggestion
When a PR gets squashed and merged into master, the title of the commit message (first line) must be the PR title
```

amoss: Why "gets" instead of "is" - it does not seem to be grammatically correct? (I'm curious as it could be an American / British difference that I have not seen before).

mfundul:
The body should also include any useful intermediate commit messages should a developer choose to include them. It cannot be every single silly commit message though.

prologic:
Correct me if I'm wrong; but this is the default behaviour of "Squash and Merge" right?

CONTRIBUTING.md
```diff
@@ -191,6 +191,13 @@ Improvements must have the title "[Area or module] : [Added/Refactored/Improved/
 
 New collectors must have the title "[New collector] : [Name of collector]".
 
+#### Commit messages when PRs are merged
+
+When a PR gets squashed and merged into master, the title of the commit message (first line) must be the PR title
+followed by the PR number.
```
ilyam8: > followed by the PR number

Why the PR number is needed?

ilyam8: > this is the default behaviour of "Squash and Merge" 

oh, right

amoss: So it is easy to find the discussion about the work.

ilyam8: i think it is impossible to exclude PR number if you squash the PR 🤔

ilyam8: 🙈 possible ofc, PR num appended by defult when we squash, just checked 🙈

prologic: 😀

prologic: I _literally_ think we're just documenting what is default behaviour here and expectations :)

prologic:
@shortpatti You need to re-approve if you're okay with the changes :)

amoss:
> Correct me if I'm wrong; but this is the default behaviour of "Squash and Merge" right?

For the title it may be although there are recent commits that do not match the PR title. For the body the default behaviour is a bullet list of every commit in the PR which we should avoid.

prologic:
> For the title it may be although there are recent commits that do not match the PR title. For the body the default behaviour is a bullet list of every commit in the PR which we should avoid.

What would you propose instead?

amoss:
> What would you propose instead?

This:

> The body of the commit message should be a short description of the work, preferably taken from the connected issue.

PR #8381, opened 2020-03-11
Added deferred error message handling to the installer.

Ferroin:
##### Summary

This adds some code to the Netdata installer to allow deferring non-fatal errors so that their messages get printed at the end of the install process. It does this by constructing the list of errors as they occur and then printing them all at the end if the list is non-empty.

With these changes, failures to download, validate, or install libmosquitto, LWS, or the go.d plugin will cause the errors to be reported when the install script otherwise finishes successfully, making it much easier for users to notice that there was a non-fatal error.

##### Component Name

area/packaging

Verified locally by artificially inducing the errors in question that this is now deferring messages for.

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -41,6 +41,24 @@ fi
 cd "${NETDATA_SOURCE_DIR}" || exit 1
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ilyam8:
i tested defer functions, work great

Issue #8378, opened 2020-03-11
Remove the feature flag for ACLK.


amoss:
### Summary

* Confirm with @cosmix 
* Check installer state
* Check code state
* Add soft feature flag in config file.
* Fix product 535

PR #8382, opened 2020-03-11
Added switch to installer to fail if Netdata Cloud functionality can't be built.

Ferroin:
##### Summary

This adds a switch called `--require-cloud` to the installer script. When specified, it causes the script to explicitly fail if it can't build Netdata with Netdata Cloud support. It is mutually exclusive with the `--disable-cloud` switch, with only the first one specified being honored.

This is mostly intended to ensure that CI can test the build/install both with and without the cloud, but it is likely to be of interest to users who want to ensure they can use their systems with netdata Cloud.

This PR also updates the build testing CI to explicitly check builds both with and without Netdata Cloud built.

##### Component Name

area/packaging
area/ci

Verified behavior by running the script locally and manually inducing failures. 

##### Additional Information

This is blocked pending completion of #8378, as the configure scripts don't correctly handle the case of being passed `--enable-aclk` when the feature flag is not set.

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -193,7 +193,8 @@ USAGE: ${PROGRAM} [options]
                              This results in more frequent updates.
   --disable-go               Disable installation of go.d.plugin.
   --enable-ebpf              Enable eBPF Kernel plugin (Default: disabled, feature preview)
-  --disable-cloud            Disable all cloud functionality.
+  --disable-cloud            Disable all Netdata Cloud functionality.
+  --require-cloud            Fail the install if it can't build Netdata Cloud support.
   --enable-plugin-freeipmi   Enable the FreeIPMI plugin. Default: enable it when libipmimonitoring is available.
   --disable-plugin-freeipmi
   --disable-https            Explicitly disable TLS support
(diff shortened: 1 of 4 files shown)


netdata-installer.sh
```diff
@@ -280,8 +281,20 @@ while [ -n "${1}" ]; do
     "--disable-go") NETDATA_DISABLE_GO=1 ;;
     "--enable-ebpf") NETDATA_ENABLE_EBPF=1 ;;
     "--disable-cloud")
-      NETDATA_DISABLE_CLOUD=1
-      NETDATA_CONFIGURE_OPTIONS="${NETDATA_CONFIGURE_OPTIONS//--disable-cloud/} --disable-cloud"
+      if [ -n "${NETDATA_REQUIRE_CLOUD}" ] ; then
+        echo "Cloud explicitly enabled, ignoring --disable-cloud."
+      else
+        NETDATA_DISABLE_CLOUD=1
+        NETDATA_CONFIGURE_OPTIONS="${NETDATA_CONFIGURE_OPTIONS//--disable-aclk/} --disable-aclk"
```
amoss: You've changed the name of the switch - it should be "--disable-cloud" !!

Ferroin:
Superseded by #8568

PR #8383, opened 2020-03-11
General cleanup of GitHub Actions workflows.

Ferroin:
##### Summary

* Remove the installer workflow, it's duplicating work being done by the build workflow.
* Update the list of distros the build workflow runs against to include Debian 8 and Alpine Edge.
* Update the name of the build workflow to properly reflect that it's testing the installer.
* Update the build workflow to also check a plain build without using the installer.
* Split the `install-required-packages.sh` testing to it's own step, using a local docker image to carry the state over. This makes it easier to be certain what failed, and speeds up the other steps a bit.
* Use `netdata-all` for the `install-required-packages.sh` invocation instead of `all`. This skips installing things that we only care about for demo sites, which users are not realistically going to install.
* Re-enable CentOS 8 build checks (they were only failing because of using `all` with `install-required-packages.sh`).
* Re-enable OpenSUSE build checks (they also appear to only have been failing because of how we were invoking `install-required-packages.sh`).
* Remove checks for Fedora 29, OpenSUSE Leap 15.0, and Ubuntu 19.04. These are all EOL upstream, and we're not testing them in Travis right now either.

##### Component Name

area/ci

Workflwos run correctly and as expected in this PR.


.github/workflows/build.yml
```diff
@@ -7,10 +7,11 @@ on:
   pull_request:
 jobs:
   build:
-    name: Build
+    name: Build & Install
```
ilyam8: should you rename the file to reflect changes?

ilyam8: it is `build.yml`

Ferroin: I'm not 100% certain yet. I was kind of planning ot see what @prologic has to say about that.

prologic: Rename it :) Thats the convention I'm using.

prologic: Can we also:

```sh
mkdir .github/dockerfiles
mv .github/Dockerfile.build_test .github/dockerfiles/
```

I'd prefer it if we keep logical directory structures for our own sanity ;)

Ferroin:
Force pushed a rebase to kick Netlify so that it hopefully reports status correctly this time...

PR #8387, opened 2020-03-12
Fix coverity scan

prologic:
##### Summary

As per title

##### Component Name

- area/ci

##### Test Plan

- This PR itself

I originally got a CI failure for this, so the workflow works but it
was previously not tested orrectly, hence the addition of `pull_request:`
to the workflow to ensure it runs inside a pull request that touches either
the workflow or the script itself.

##### Additional Information

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: '0 1 * * *'
   pull_request:
+    paths:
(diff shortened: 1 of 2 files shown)


prologic:
I _realize_ why this was hard to test now; because we use a forking model. Redoing this PR a different way.

prologic:
Superseded by #8388 so I could properly test it.

PR #8388, opened 2020-03-12
Fix coverity scan

prologic:
##### Summary

As per title

##### Component Name

- area/ci

##### Test Plan

- This PR itself

I originally got a CI failure for this, so the workflow works but it
was previously not tested orrectly, hence the addition of `pull_request:`
to the workflow to ensure it runs inside a pull request that touches either
the workflow or the script itself.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -16,10 +16,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Prepare environment
+        env:
+          DEBIAN_FRONTEND: 'noninteractive'
         run: |
-          ./packaging/installer/install-required-packages.sh --dont-wait --non-interactive netdata
-          sudo apt-get install -y libjson-c-dev libipmimonitoring-dev libcups2-dev libsnappy-dev \
-                                  libprotobuf-dev libprotoc-dev libssl-dev protobuf-compiler
+          ./packaging/installer/install-required-packages.sh \
+            --dont-wait --non-interactive netdata
+          sudo apt-get install -y libjson-c-dev libipmimonitoring-dev \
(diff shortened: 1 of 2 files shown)


prologic:
>  Coverity Scan / coverity (pull_request) Successful in 3m

✅

.github/workflows/coverity.yml
```diff
@@ -12,12 +16,18 @@ jobs:
...
         run: |
-          ./packaging/installer/install-required-packages.sh --dont-wait --non-interactive netdata
-          sudo apt-get install -y libjson-c-dev libipmimonitoring-dev libcups2-dev libsnappy-dev \
-                                  libprotobuf-dev libprotoc-dev libssl-dev protobuf-compiler
+          ./packaging/installer/install-required-packages.sh \
+            --dont-wait --non-interactive netdata
+          sudo apt-get install -y libjson-c-dev libipmimonitoring-dev \
+                                  libcups2-dev libsnappy-dev libprotobuf-dev \
+                                  libprotoc-dev libssl-dev protobuf-compiler \
+                                  libnetfilter-acct-dev
```
Ferroin: `netdata-all` should be used so that we don't need an extra install step. If it doesn't pull in these packages, that's a bug and should be handled as such.

prologic: Whilst I agree in general, these however are "dev" dependencies and probably optional where normally a user would not need to install them. However as pointed out in our sync/meeting just now we probably are legitimately missing some things in our `install-required-pacakges.sh` script :/ We should fix that; but let's file a separate issue and do it there so we can merge this PR :)

prologic: Please link/ref the issue here once you've created it :)

Issue #8263, opened 2020-03-02
Fix known issues with claiming


amoss:
#### Summary

As we work through testing we are collecting issues with claiming together so that we can track them before the release.

#### Known issues

* Timeout handling. Before I set the proxy I run the claiming command (the command line stalled for many seconds, probably reached the 3*5min timeout limit) but the script reported “Node was successfully claimed.” which was not the case.
* `hostname` is not available on some systems but `$HOSTNAME` is set.
* Needs a debug flag to see the contents of `tmpin` and `tmpout`.
* The integer expression problem if the return code is blank

mfundul:
https://github.com/netdata/netdata/blob/fd343ac0c262cd6a1535b67e7894d06760e10259/claim/netdata-claim.sh.in#L219

This line of code should make sure that `HTTP_STATUS_CODE` has valid contents before testing against `204`.

PR #8392, opened 2020-03-12
Adds support for SOCKS5 proxy in claiming with curl

### LINKED ISSUE #8090: ACLK Claiming should support Socks5 proxy
##### Feature idea summary
ACLK Claiming should support Socks5 proxy

##### Expected behavior
- [ ] Clarify with Cloud/SRE about testing endpoint for Claiming
- [ ] Update and test claiming script to support Socks5 proxy (pass parameters from config file to override env_var)

underhood:
##### Summary
some things in claiming script have to be fixed as well for this to be 100% (see #8263) - therefore this PR works and brings SOCKS5 proxy but there are error reporting issues that will be fixes in #8263.

If agent configured with SOCKS5 proxy `[agent_cloud_link]` key `proxy` it should pass that setting to curl when claiming script is called from within netdata agent with `netdata -D -Wclaim`.

If agent configured with `proxy = none` claiming script should not use proxy even if environment variables are set.

Fixes #8090 

##### Component Name
ACLK/claiming

##### Test Plan
Make sure claiming script can be found by $PATH (e.g. if you test with NETDATA installed in custom dir)!

Test with local cloud test environment hidden behind socks5 proxy. *(example: run agent on VM that doesn't have direct access to `local cloud environment`)*

In config file of Netdata set options:

```
[agent_cloud_link]
    proxy = socks5://X.X.X.X:YYYY
and for other test
    proxy = none
```

When curl not present in system and wget present. Show error when trying to use SOCKS with wget (as wget doesn't have support).

### DIFF
--- a/aclk/aclk_common.c
+++ b/aclk/aclk_common.c
@@ -105,3 +105,45 @@ const char *aclk_lws_wss_get_proxy_setting(ACLK_PROXY_TYPE *type) {
 
     return proxy;
 }
+
+int aclk_decode_base_url(char *url, char **aclk_hostname, char **aclk_port)
+{
+int pos = 0;
+    if (!strncmp("https://", url, 8))
+    {
+        pos = 8;
+    }
+    else if (!strncmp("http://", url, 7))
+    {
+        error("Cannot connect ACLK over %s -> unencrypted link is not supported", url);
+        return 1;
+    }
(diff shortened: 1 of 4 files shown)


amoss:
Merged into PR 8406 for issue #8263.

PR #8398, opened 2020-03-13
ACLK: Improved the agent "pop-corning" phase

### LINKED ISSUE #8350: Fix the pop-corning implementation in the ACLK.
#### Summary

* Before the beginning of the sprint clarify with product what should happen to queries / commands on the ACLK before the popcorning process has finished.
* All queries are ignored (dropped) during pop-corning, but commands from the cloud are in the same queue for processing so the agent would ignore requests during this period.
* Any query resets the pop-corning timer (including encapsulated queries) so pop-corning can go on indefinitely.
* The implementation of the time-out / check is wrong and it will always sleep 10s longer than necessary.

stelfrag:
Fixes #8350 
##### Summary
- The agent stability status during pop-corning is now checked every second. 
- The agents switches to a stable state after ACLK_STABLE_TIMEOUT+1 seconds (currently 10+1)
- Commands from the cloud during pop-corning are ignored

##### Test Plan
- Run the agent in local dev environment
  - Notice a new message similar to 
    `Waiting for agent collectors to initialize. Last activity was 7 seconds ago` while the agent is waiting for the collectors to initialize

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -898,7 +898,6 @@ void *aclk_query_main_thread(void *ptr)
         time_t checkpoint;
 
         checkpoint = now_realtime_sec() - last_init_sequence;
-        info("Waiting for agent collectors to initialize last activity %ld seconds ago" , checkpoint);
         if (checkpoint > ACLK_STABLE_TIMEOUT) {
             agent_state = AGENT_STABLE;
             info("AGENT stable, last collector initialization activity was %ld seconds ago", checkpoint);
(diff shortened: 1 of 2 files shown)


stelfrag:
> Every call to aclk_queue_query will reset the popcorn timer. But not all calls are new collectors or charts. The specific cases that prevent those calls when pop-corning is happening are a bit opaque and will cause the code to be fragile when it is updated in the future. Please move the reset of the popcorning timer to the specific cases that should trigger it before they call aclk_queue_query (e.g. aclk_del_collector, aclk_add_collector and the chart case). Checking the cases some of the topics look a bit strange "_chart" ?

Moved check to the calling functions and changed some checks from agent_state  != AGENT_STABLE to == AGENT_INITIALIZING for better readability

PR #8407, opened 2020-03-16
Improve the impact of health code on netdata scalability

### LINKED ISSUE #8251: Netdata takes too long to give the metrics to the Prometheus server
##### Bug report summary
We have a Prometheus server that collects `average` and `as-collected` metrics every 90 seconds with the timeout of 75 seconds from our master netdata server which is monitoring 250 slaves. 
The problem is most of the time the scrape duration from the master netdata is bellow 20 seconds but sometimes the scrape duration reaches the timeout (75s) hence we lose the metrics on our grafana dashboards.
I tried to download the metrics manually via `wget` in the Prometheus server and here is the results:
image
As you can see, the metrics are downloaded in 1.5s but it took about 50 seconds to establish the connection to the master netdata.
And here is the network graph in the central netdata (the only process which may cause the network output is either we opening the web UI and the Prometheus server reading the metrics and the 777 Mb/s out peak is when I run the `wget` command):
image
 
##### OS / Environment

```
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 18.04.1 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="18.04.1 LTS (Bionic Beaver)"
/etc/os-release:ID=ubuntu
/etc/os-release:PRETTY_NAME="Ubuntu 18.04.1 LTS"
```

##### Netdata version

 netdata v1.20.0

### LINKED ISSUE #8368: Improve the impact of health code on netdata scalability
##### Feature idea summary
There is very high CPU utilization (almost 100% even in a 32-thread bare-metal server) when executing alarms for every slave at the same time, which lasts for many seconds. Mainly this code:

```
Thread 316 (Thread 0x7fbdb5459700 (LWP 41247)):
#0  0x00007fbdbc9b596d in read () from /lib64/libc.so.6
#1  0x00007fbdbc941c74 in __GI__IO_file_underflow () from /lib64/libc.so.6
#2  0x00007fbdbc942e42 in __GI__IO_default_uflow () from /lib64/libc.so.6
#3  0x00007fbdbc935bf4 in __GI__IO_getline_info () from /lib64/libc.so.6
#4  0x00007fbdbc9349f6 in fgets () from /lib64/libc.so.6
#5  0x000000000044869b in health_alarm_execute (ae=0x149b13790, host=0x8dfa2010) at health/health.c:331
#6  health_process_notifications (ae=0x149b13790, host=0x8dfa2010) at health/health.c:350
#7  health_alarm_log_process (host=0x8dfa2010) at health/health.c:370
#8  health_main (ptr=0x701048 <static_threads+616>) at health/health.c:957
#9  0x0000000000432b9f in thread_start (ptr=0x1353cd0) at libnetdata/threads/threads.c:170
#10 0x00007fbdbcc9be65 in start_thread () from /lib64/libpthread.so.0
#11 0x00007fbdbc9c488d in clone () from /lib64/libc.so.6
```

We can improve the impact of health code on netdata scalability by taking alarm execution and notification outside the critical region of the health main thread.

mfundul:
##### Summary
Fixes #8368
Fixes #8251

Improve netdata scalability by reducing the impact of alarm execution and notification to critical sections in the main health thread.

##### Component Name
health
daemon
##### Test Plan
Set-up a bare metal server with 500 lxc containers as netdata slaves connected to the master in the host. Used dbengine as memory mode for both netdata.conf and stream.conf.

Make sure the spawn server process shuts down correctly.
Make sure the spawn server process does not fail to start.

Run: `/etc/netdata/edit-config health.d/cpu.conf` and set contents such as:
```
template: 1min_cpu_usage
      on: system.cpu
      os: linux
   hosts: *
  lookup: average -1m unaligned of user,system,softirq,irq,guest
   units: %
   every: 30s
    warn: $this > 1
    crit: $this > 2
   delay: down 15s multiplier 1.5 max 30s
    info: average cpu utilization for the last 1 minute (excluding iowait, nice and steal)
      to: sysadmin
```
This modifies the stock alarm `cpu.conf` to be triggered easily, e.g. set the warning/critical thresholds at 1%. Verify the notification is sent by checking `sendmail` worked or looking at `error.log` for the output of `alarm-notify.sh`.

##### Additional Information
The implementation forks a new spawn server process that is the same netdata binary with a special command line argument.

### DIFF
--- a/libnetdata/popen/popen.c
+++ b/libnetdata/popen/popen.c
@@ -86,7 +86,7 @@ static void myp_del(pid_t pid) {
  * Returns -1 on failure, 0 on success. When FLAG_CREATE_PIPE is set, on success set the FILE *fp pointer.
  */
 static inline int custom_popene(const char *command, volatile pid_t *pidptr, char **env, uint8_t flags, FILE **fpp) {
-    FILE *fp;
+    FILE *fp = NULL;
     int ret = 0; // success by default
     int pipefd[2], error;
     pid_t pid;
--- a/libnetdata/popen/popen.c
+++ b/libnetdata/popen/popen.c
@@ -83,10 +83,11 @@ static void myp_del(pid_t pid) {
 #define FLAG_CLOSE_FD       2 // Close all file descriptors other than STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO
 
 /*
- * Returns NULL on failure. When FLAG_CREATE_PIPE is set returns the FILE * pointer
+ * Returns -1 on failure, 0 on success. When FLAG_CREATE_PIPE is set, on success set the FILE *fp pointer.
  */
-static inline unsigned long custom_popene(const char *command, volatile pid_t *pidptr, char **env, uint8_t flags) {
-    FILE *fp = (FILE *)1; // set to 1 to differentiate from NULL when not setting FLAG_CREATE_PIPE
+static inline int custom_popene(const char *command, volatile pid_t *pidptr, char **env, uint8_t flags, FILE **fpp) {
+    FILE *fp;
+    int ret = 0; // success by default
     int pipefd[2], error;
     pid_t pid;
     char *const spawn_argv[] = {
@@ -100,7 +101,7 @@ static inline unsigned long custom_popene(const char *command, volatile pid_t *p
 
     if (flags & FLAG_CREATE_PIPE) {
         if (pipe(pipefd) == -1)
-            return (unsigned long)NULL;
+            return -1;
         if ((fp = fdopen(pipefd[PIPE_READ], "r")) == NULL) {
             goto error_after_pipe;
         }
@@ -158,11 +159,13 @@ static inline unsigned long custom_popene(const char *command, volatile pid_t *p
         error("Failed to spawn command: '%s' from parent pid %d.", command, getpid());
         if (flags & FLAG_CREATE_PIPE) {
             fclose(fp);
-            fp = NULL;
         }
+        ret = -1;
     }
     if (flags & FLAG_CREATE_PIPE) {
         close(pipefd[PIPE_WRITE]);
+        if (0 == ret) // on success set FILE * pointer
+            *fpp = fp;
     }
 
     if (!error) {
@@ -173,7 +176,7 @@ static inline unsigned long custom_popene(const char *command, volatile pid_t *p
     if (posix_spawn_file_actions_destroy(&fa))
         error("posix_spawn_file_actions_destroy");
 
-    return (unsigned long)fp;
+    return ret;
 
 error_after_posix_spawn_file_actions_init:
     if (posix_spawn_file_actions_destroy(&fa))
@@ -187,7 +190,7 @@ error_after_pipe:
 
         close(pipefd[PIPE_WRITE]);
     }
-    return (unsigned long)NULL;
+    return -1;
 }
 
 // See man environ
@@ -248,17 +251,20 @@ int myp_reap(pid_t pid) {
 }
 
 FILE *mypopen(const char *command, volatile pid_t *pidptr) {
-    return (FILE *)custom_popene(command, pidptr, environ, FLAG_CREATE_PIPE | FLAG_CLOSE_FD);
+    FILE *fp = NULL;
+    (void)custom_popene(command, pidptr, environ, FLAG_CREATE_PIPE | FLAG_CLOSE_FD, &fp);
+    return fp;
 }
 
 FILE *mypopene(const char *command, volatile pid_t *pidptr, char **env) {
-    return (FILE *)custom_popene(command, pidptr, env, FLAG_CREATE_PIPE | FLAG_CLOSE_FD);
+    FILE *fp = NULL;
+    (void)custom_popene(command, pidptr, env, FLAG_CREATE_PIPE | FLAG_CLOSE_FD, &fp);
+    return fp;
 }
 
 // returns 0 on success, -1 on failure
 int netdata_spawn(const char *command, volatile pid_t *pidptr) {
-    unsigned long ret = custom_popene(command, pidptr, environ, 0);
-    return (ret != (unsigned long)NULL) ? 0 : -1;
+    return custom_popene(command, pidptr, environ, 0, NULL);
 }
 
 int custom_pclose(FILE *fp, pid_t pid) {
--- a/libnetdata/popen/popen.c
+++ b/libnetdata/popen/popen.c
@@ -82,8 +82,11 @@ static void myp_del(pid_t pid) {
 #define FLAG_CREATE_PIPE    1 // Create a pipe like popen() when set, otherwise set stdout to /dev/null
 #define FLAG_CLOSE_FD       2 // Close all file descriptors other than STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO
 
-static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, char **env, uint8_t flags) {
-    FILE *fp = (FILE *)1; // set to 1 to differentiate from NULL when running NO_PIPE_NULL_STDOUT
+/*
+ * Returns NULL on failure. When FLAG_CREATE_PIPE is set returns the FILE * pointer
+ */
+static inline unsigned long custom_popene(const char *command, volatile pid_t *pidptr, char **env, uint8_t flags) {
+    FILE *fp = (FILE *)1; // set to 1 to differentiate from NULL when not setting FLAG_CREATE_PIPE
     int pipefd[2], error;
     pid_t pid;
     char *const spawn_argv[] = {
@@ -97,7 +100,7 @@ static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, c
 
     if (flags & FLAG_CREATE_PIPE) {
         if (pipe(pipefd) == -1)
-            return NULL;
+            return (unsigned long)NULL;
         if ((fp = fdopen(pipefd[PIPE_READ], "r")) == NULL) {
             goto error_after_pipe;
         }
@@ -170,7 +173,7 @@ static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, c
     if (posix_spawn_file_actions_destroy(&fa))
         error("posix_spawn_file_actions_destroy");
 
-    return fp;
+    return (unsigned long)fp;
 
 error_after_posix_spawn_file_actions_init:
     if (posix_spawn_file_actions_destroy(&fa))
@@ -184,7 +187,7 @@ error_after_pipe:
 
         close(pipefd[PIPE_WRITE]);
     }
-    return NULL;
+    return (unsigned long)NULL;
 }
 
 // See man environ
@@ -245,17 +248,17 @@ int myp_reap(pid_t pid) {
 }
 
 FILE *mypopen(const char *command, volatile pid_t *pidptr) {
-    return custom_popene(command, pidptr, environ, FLAG_CREATE_PIPE | FLAG_CLOSE_FD);
+    return (FILE *)custom_popene(command, pidptr, environ, FLAG_CREATE_PIPE | FLAG_CLOSE_FD);
 }
 
 FILE *mypopene(const char *command, volatile pid_t *pidptr, char **env) {
-    return custom_popene(command, pidptr, env, FLAG_CREATE_PIPE | FLAG_CLOSE_FD);
+    return (FILE *)custom_popene(command, pidptr, env, FLAG_CREATE_PIPE | FLAG_CLOSE_FD);
 }
 
 // returns 0 on success, -1 on failure
 int netdata_spawn(const char *command, volatile pid_t *pidptr) {
-    FILE *fp = custom_popene(command, pidptr, environ, 0);
-    return (fp != NULL) ? 0 : -1;
+    unsigned long ret = custom_popene(command, pidptr, environ, 0);
+    return (ret != (unsigned long)NULL) ? 0 : -1;
 }
 
 int custom_pclose(FILE *fp, pid_t pid) {
--- a/libnetdata/popen/popen.c
+++ b/libnetdata/popen/popen.c
@@ -103,11 +103,13 @@ static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, c
         }
     }
 
-    // Mark all files to be closed by the exec() stage of posix_spawn()
-    int i;
-    for (i = (int) (sysconf(_SC_OPEN_MAX) - 1); i >= 0; i--) {
-        if (i != STDIN_FILENO && i != STDERR_FILENO && !(NO_PIPE_NULL_STDOUT == use_pipe && i == STDOUT_FILENO))
-            (void) fcntl(i, F_SETFD, FD_CLOEXEC);
+    if (USE_PIPE == use_pipe) { //TODO: remove me
+        // Mark all files to be closed by the exec() stage of posix_spawn()
+        int i;
+        for (i = (int) (sysconf(_SC_OPEN_MAX) - 1); i >= 0; i--) {
+            if (i != STDIN_FILENO && i != STDERR_FILENO && !(NO_PIPE_NULL_STDOUT == use_pipe && i == STDOUT_FILENO))
+                (void) fcntl(i, F_SETFD, FD_CLOEXEC);
+        }
     }
 
     if (!posix_spawn_file_actions_init(&fa)) {
@@ -118,7 +120,7 @@ static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, c
                 goto error_after_posix_spawn_file_actions_init;
             }
         } else { // NO_PIPE_NULL_STDOUT == use_pipe
-            if (posix_spawn_file_actions_addopen(&fa, STDOUT_FILENO, "/dev/null", 0, O_WRONLY)) {
+            if (posix_spawn_file_actions_addopen(&fa, STDOUT_FILENO, "/dev/null", O_WRONLY, 0)) {
                 error("posix_spawn_file_actions_addopen() failed");
                 // this is not a fatal error
             }
--- a/libnetdata/popen/popen.c
+++ b/libnetdata/popen/popen.c
@@ -78,8 +78,12 @@ static void myp_del(pid_t pid) {
 #define PIPE_READ 0
 #define PIPE_WRITE 1
 
-static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, char **env) {
-    FILE *fp;
+/* use_pipe flag definitions */
+#define NO_PIPE_NULL_STDOUT 0 // Don't create a pipe, set stdout to /dev/null
+#define USE_PIPE 1  // Create a pipe like popen()
+
+static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, char **env, uint8_t use_pipe) {
+    FILE *fp = (FILE *)1; // set to 1 to differentiate from NULL when running NO_PIPE_NULL_STDOUT
     int pipefd[2], error;
     pid_t pid;
     char *const spawn_argv[] = {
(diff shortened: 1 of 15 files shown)


Ferroin:
@mfundul I think you've got a typo in your issue description. This doesn't do anything at all with respect to #8383 except possibly unintentionally close it when merged.

mfundul:
> 
> 
> @mfundul I think you've got a typo in your issue description. This doesn't do anything at all with respect to #8383 except possibly unintentionally close it when merged.

Whoops!

ktsaou:


```diff

Clones added
============
- spawn/spawn_client.c  1
- spawn/spawn_server.c  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5186965&bid=17017988)

amoss:
The test plan is not detailed enough, Please provide examples of the alarm configuration that you are using and which conditions you are checking to ensure that alarm notifications are still delivered.

mfundul:
> 
> 
> The test plan is not detailed enough, Please provide examples of the alarm configuration that you are using and which conditions you are checking to ensure that alarm notifications are still delivered.

Added how to test alarms in the description.

amoss:
My first concerns are high-level issues before I get into the details of the code.
* The second netdata process that shows up in the process list looks ugly. What portable options do we have to change the process name?
* Why is the flag called `--special-spawn-server`?
* Why did you choose a separate process for handling forks instead of using `posix_spawn`?

mfundul:
> 
> 
> My first concerns are high-level issues before I get into the details of the code.
> 
>     * The second netdata process that shows up in the process list looks ugly. What portable options do we have to change the process name?

Nothing portable. The alternative is to create another binary, which I did consider for a long time but opted for reusing the daemon process.

> 
>     * Why is the flag called `--special-spawn-server`?
> 

The name could be anything, if you could propose something better that'd be great. This is supposed to be the netdata spawn server process.

>     * Why did you choose a separate process for handling forks instead of using `posix_spawn`?

I use `posix_spawn()` in the `netdata_spawn()` call. `posix_spawn()` is similar to `fork()/exec()` and creates a separate process. It is more performant in various OSes such as linux, exactly because it's more limited than fork(), so the OS can optimize the process cloning procedure

amoss:
> Nothing portable. The alternative is to create another binary, which I did consider for a long time but opted for reusing the daemon process.

What about a symlink to the same binary and checking `argv[0]` instead of a flag - does that still work across systems?

mfundul:
> 
> 
> > Nothing portable. The alternative is to create another binary, which I did consider for a long time but opted for reusing the daemon process.
> 
> What about a symlink to the same binary and checking `argv[0]` instead of a flag - does that still work across systems?

This should be the equivalent of having a different binary. We still have to do all the work in packaging.

amoss:
> I use posix_spawn() in the netdata_spawn() call. posix_spawn() is similar to fork()/exec() and creates a separate process. It is more performant in various OSes such as linux, exactly because it's more limited than fork(), so the OS can optimize the process cloning procedure

Guide me through it to check that I have understood.
* The performance problem that we are fixing is that in a large scale server the cost of the `fork` is prohibitive because of the time taken to rewrite the MMU to duplicate the process with the large memory space.
* `posix_spawn` does not duplicate the memory space for the new process as it creates an 'empty' process and does the `exec` internally.

What were the performance implications of calling `posix_spawn` directly (without a spawn server) from the large main process?

mfundul:
> 
> 
> > I use posix_spawn() in the netdata_spawn() call. posix_spawn() is similar to fork()/exec() and creates a separate process. It is more performant in various OSes such as linux, exactly because it's more limited than fork(), so the OS can optimize the process cloning procedure
> 
> Guide me through it to check that I have understood.
> 
>     * The performance problem that we are fixing is that in a large scale server the cost of the `fork` is prohibitive because of the time taken to rewrite the MMU to duplicate the process with the large memory space.
> 
>     * `posix_spawn` does not duplicate the memory space for the new process as it creates an 'empty' process and does the `exec` internally.
> 
> 
> What were the performance implications of calling `posix_spawn` directly (without a spawn server) from the large main process?

We have been using `posix_spawn()` directly from the main process for many months now. The performance implications have been overhead of over 100ms per spawn and thousands of netdata threads being disrupted and causing 100% CPU spikes at every spawn.

amoss:
> We have been using posix_spawn() directly from the main process for many months now. The performance implications have been overhead of over 100ms per spawn and thousands of netdata threads being disrupted and causing 100% CPU spikes at every spawn.

* What is the performance overhead of `posix_spawn` in the small spawn server?
* What are the variables that affect how it scales?

mfundul:
> 
> 
> > We have been using posix_spawn() directly from the main process for many months now. The performance implications have been overhead of over 100ms per spawn and thousands of netdata threads being disrupted and causing 100% CPU spikes at every spawn.
> 
>     * What is the performance overhead of `posix_spawn` in the small spawn server?

The small spawn server can spawn/execute `echo`/wait for it to finish with a rate of 3000 commands per second.

> 
>     * What are the variables that affect how it scales?

The size of the process and the number of file descriptors.

amoss:
It's interesting that `posix_spawn` is behaving that way as it should not have a performance penalty that scales up with the process size. Perhaps the slow fall-back path is being triggered. Did you consider using `clone` and setting the arguments so that the memory and fd-table were not duplicated?

mfundul:
> 
> 
> It's interesting that `posix_spawn` is behaving that way as it should not have a performance penalty that scales up with the process size. Perhaps the slow fall-back path is being triggered. Did you consider using `clone` and setting the arguments so that the memory and fd-table were not duplicated?

glibC is using `clone()` in Linux kernels. The memory is not actually duplicated, I've run tests comparing `fork()` to `posix_spawn()` and only `fork()` will trigger out of memory errors.

One performance problem stems from the fact that we don't keep track of file-descriptors in the netdata daemon. Had we proactively marked every file descriptor as close-on-exec then we would save up to 50ms.

The biggest performance problem though is the huge netdata process, thousands of threads and many page table entries. It can take north of 200ms to spawn in extreme scenarios, where the file descriptor overhead remains relatively steady.

Finally, `posix_spawn()` is the portable way for the UNIX systems, which is all we support at the moment. For Windows we will have to do something different. For linux, it's mostly going to be using the optimal flags for clone as shown here:
https://github.com/bminor/glibc/blob/c4d4419433f1b2c9c4d54ee6da2b0d2a30e3fda8/sysdeps/unix/sysv/linux/spawni.c#L382

amoss:
The flags for `spawnix` do not appear to be optimal. We have three variables that have an impact on the cost of the process creation:
* Number of file-descriptors
* Size of the memory space (if shared)
* Number of threads (if the new process is in the same thread-group ?)

In the slow case (overloaded server, using fork or the code-path above) all three seem to be having an impact on the cost of the process creation (3ms vs 200ms). What I'm still not clear on is what is the cost of performing the process creation from the main netdata process in the optimal case:
* Not sharing file-descriptors (i.e. `CLONE_FILES` is not set?).
* No sharing memory (i.e. `CLONE_VM` is not set?).
* Ensuring `CLONE_VFORK` is not set (so that the execution time of the spawned process is not part of the cost).
* Ensuring only a single thread is in the created process (`CLONE_THREAD` is false?).

Ideally this is the case that `posix_spawn` should provide but the code shows that it does not. What do we know about the process creation overhead in this ideal case from the main netdata process?

mfundul:
> 
> 
> The flags for `spawnix` do not appear to be optimal. We have three variables that have an impact on the cost of the process creation:
> 
>     * Number of file-descriptors
> 
>     * Size of the memory space (if shared)
> 
>     * Number of threads (if the new process is in the same thread-group ?)
> 
> 
> In the slow case (overloaded server, using fork or the code-path above) all three seem to be having an impact on the cost of the process creation (3ms vs 200ms). What I'm still not clear on is what is the cost of performing the process creation from the main netdata process in the optimal case:
> 
>     * Not sharing file-descriptors (i.e. `CLONE_FILES` is not set?).
> 

`CLONE_FILES` not being set is optimal. Sharing file descriptors is faster than cloning them.

>     * No sharing memory (i.e. `CLONE_VM` is not set?).
> 

No setting `CLONE_VM` is like `fork()`, it will trigger out of memory errors because it will copy the virtual memory space.

>     * Ensuring `CLONE_VFORK` is not set (so that the execution time of the spawned process is not part of the cost).

`CLONE_VFORK` is the way the OS compensates for using `CLONE_VM`. `vfork()` is the faster implementation of `fork()` and you really need `CLONE_VFORK` with `CLONE_VM` to not corrupt the process memory.

> 
>     * Ensuring only a single thread is in the created process (`CLONE_THREAD` is false?).
> 

`CLONE_THREAD` creates a thread instead of a process, so it's the equivalent of `pthread_create()` and it's not what we should be using here.

> 
> Ideally this is the case that `posix_spawn` should provide but the code shows that it does not. What do we know about the process creation overhead in this ideal case from the main netdata process?

So my conclusion is that for all the above flags, the only way for `posix_spawn()` to work correctly is how it's currently implemented by glibc. So I wouldn't say there exists a more ideal case for netdata.

stelfrag:
Having a second process `netdata` _may_ break scripts that check if netdata is running

amoss:
> Not setting CLONE_VM is like fork(), it will trigger out of memory errors because it will copy the virtual memory space.

Why is this? My understanding that that we are forking so that we can exec the notification script. There should not be any access to memory in the process address space.

amoss:
> CLONE_VFORK is the way the OS compensates for using CLONE_VM. vfork() is the faster implementation of fork() and you really need CLONE_VFORK with CLONE_VM to not corrupt the process memory.

But `CLONE_VFORK` is designed to implement a vfork - i.e. the parent will wait for the child to exit before continuing. This makes the execution time of the child part of the overall cost for the fork.

mfundul:
> 
> 
> Having a second process `netdata` _may_ break scripts that check if netdata is running

This shouldn't be the case because there are already a lot of `netdata` processes in many systems due to containers. We use the `.pid` files, systemd and `netdatacli` to locate instances in the current context.

mfundul:
> 
> 
> > Not setting CLONE_VM is like fork(), it will trigger out of memory errors because it will copy the virtual memory space.
> 
> Why is this? My understanding that that we are forking so that we can exec the notification script. There should not be any access to memory in the process address space.

This is why `posix_spawn()` sets `CLONE_VM`. Marking this memory as "shared" avoids the copying and `exec` eventually replaces the memory.

mfundul:
> 
> 
> > CLONE_VFORK is the way the OS compensates for using CLONE_VM. vfork() is the faster implementation of fork() and you really need CLONE_VFORK with CLONE_VM to not corrupt the process memory.
> 
> But `CLONE_VFORK` is designed to implement a vfork - i.e. the parent will wait for the child to exit before continuing. This makes the execution time of the child part of the overall cost for the fork.

This is true, but it is the trade-off of not copying the parent address space. This way the child can not alter the parent's memory and the parent cannot alter the child's memory (which is shared). This problem hasn't really been solved in the OSes.

amoss:
Ok, then I've finished going through the high-level architectural view. I can't see an easier way to solve the problem than you have used. Next part of the review will be slower as I need to go through the `custom_popene` code and check each of the steps / calls.

amoss:
> > Having a second process `netdata` _may_ break scripts that check if netdata is running
> 
> This shouldn't be the case because there are already a lot of `netdata` processes in many systems due to containers. We use the `.pid` files, systemd and `netdatacli` to locate instances in the current context.

What about the code in the installer that checks for and kills any currently running netdata process?

mfundul:
> 
> 
> > > Having a second process `netdata` _may_ break scripts that check if netdata is running
> > 
> > 
> > This shouldn't be the case because there are already a lot of `netdata` processes in many systems due to containers. We use the `.pid` files, systemd and `netdatacli` to locate instances in the current context.
> 
> What about the code in the installer that checks for and kills any currently running netdata process?

It should be what I described in general, I think the code pattern is the following:
https://github.com/netdata/netdata/blob/7cdb949d299b841d213aa682e6428a67380d12fd/packaging/installer/netdata-uninstaller.sh#L387

if `netdatacli` fails.

amoss:
Why does the uninstaller assume that `netdatacli` is in the path? We don't need to do that as we know where it goes at install time and there are many install configurations where the netdata binaries don't need to be in the path (containers for one, but also user installs under a prefix).

mfundul:
> 
> 
> Why does the uninstaller assume that `netdatacli` is in the path? We don't need to do that as we know where it goes at install time and there are many install configurations where the netdata binaries don't need to be in the path (containers for one, but also user installs under a prefix).

You should probably discuss this with @Ferroin , I don't think he's monitoring this discussion in the PR.

libnetdata/popen/popen.c
```diff
@@ -91,23 +95,36 @@ static inline FILE *custom_popene(const char *command, volatile pid_t *pidptr, c
...
+                goto error_after_posix_spawn_file_actions_init;
+            }
+        } else {
+            // set stdout to /dev/null
+            if (posix_spawn_file_actions_addopen(&fa, STDOUT_FILENO, "/dev/null", O_WRONLY, 0)) {
+                error("posix_spawn_file_actions_addopen() failed");
+                // this is not a fatal error
+            }
         }
     } else {
```
amoss, line 129: We don't need an if-then-else here when the check bails to an exit, would be cleaner to do:
```
if(posix_spawn_file_actions_init ... ) {
  error(...)
  goto ....
}
// Continue other cases
```
I think it makes it easier to read, not an essential change.

mfundul: Agreed. I kept the existing structure here, didn't want to refactor the existing code too much.

amoss:
I've finished reviewing the code in `custom_popene`, the ways that it interacts with our signals handler and each of the call sites that also get there via `mypopen()` and `mypopene()`. I can't find any problems. I see that we still have a loop that closes fds up to _SC_MAX_OPEN rather than using the posix_spawn_action to close the file-descriptors but the code inside glibc shows that this would take up memory per-fd so I can see why you left it alone.

Next up is the server.c and the client.c code to check the high-level flow through the code and then I'm done.

I'm very curious about something though - why does our reaper not appear to do any actual reaping? Have I missed something when I'm reading the code. For child processes it looks like we always sleep/check until they exit and there is no actual point at which we use the myp list of child processes to reap anything?

mfundul:
> 
> 
> I've finished reviewing the code in `custom_popene`, the ways that it interacts with our signals handler and each of the call sites that also get there via `mypopen()` and `mypopene()`. I can't find any problems. I see that we still have a loop that closes fds up to _SC_MAX_OPEN rather than using the posix_spawn_action to close the file-descriptors but the code inside glibc shows that this would take up memory per-fd so I can see why you left it alone.

This code is not being used by the spawn server, because it has high CPU overhead (`FLAG_CLOSE_FD` is not set). Instead, the spawn server itself closes `_SC_OPEN_MAX` descriptors once, and then it doesn't have to do it again because it doesn't create any additional file descriptors.

> 
> Next up is the server.c and the client.c code to check the high-level flow through the code and then I'm done.
> 
> I'm very curious about something though - why does our reaper not appear to do any actual reaping? Have I missed something when I'm reading the code. For child processes it looks like we always sleep/check until they exit and there is no actual point at which we use the myp list of child processes to reap anything?

The reaper is only reaping processes that are not explicitly spawned by netdata. So, if a netdata spawn forks another child process, it's going to be reaped.

amoss:
Why is `exit()` being called at the end of `spawn_client` ?

mfundul:
> 
> 
> Why is `exit()` being called at the end of `spawn_client` ?

That's a very good question. Pretty sure it should be a `return` instead.

mfundul:
This will require a new round of testing...

amoss:
Today I'm going through the business logic in the spawn client and server. Some initial questions / comments:

* Why is `copy_to_prot_buffer` defined as a static inline function in a header?
* Why does it access a global buffer shared between multiple translation units?
* Why is the command queue stored in a tree instead of a list?

mfundul:
> 
> 
> Today I'm going through the business logic in the spawn client and server. Some initial questions / comments:
> 
>     * Why is `copy_to_prot_buffer` defined as a static inline function in a header?
> 

Inlining functions in C has this standard practice, it's the safest way.
- Some analysis: https://blogs.oracle.com/d/inline-functions-in-c
- Some examples: https://elixir.bootlin.com/linux/v5.6/source/include/linux/fs.h#L519

>     * Why does it access a global buffer shared between multiple translation units?
> 

This buffer and the corresponding logic are common so the function and the buffer are deduplicated. This is not simply a matter of translation units anymore, because we actually fork into a completely different process. So, in practice, every process uses a completely different storage space for the global buffer (client buffer in netdata daemon, server buffer in spawn server).

>     * Why is the command queue stored in a tree instead of a list?

Since the commands that are being executed are completed out of order (e.g. one command takes 10 times more time than another to complete) the tree guarantees fast time for looking up any arbitrary command context.

amoss:
I wasn't expecting an answer on a Greek holiday, in fact I was coming back to answer some of these myself :)

I've finished tracing through the interactions with the health code so I've seen the serial number tracking and the purpose of the tree. Overall multiplexing the different queues (commands to be despatched, commands in flight, completed commands) by the flags works and I can't think of a better way to handle the `health_alarm_wait_for_execution()` when it is keyed by the serial number. Keeping the mutex/cond-var pair inside the cmd_info avoid a number of potential race hazards.

My question about the `static inline` wasn't really about how to define static inlined functions - why should this one be statically inlined when we are relying on LTO for everything else?

mfundul:
> 
> 
> I wasn't expecting an answer on a Greek holiday, in fact I was coming back to answer some of these myself :)
> 
> I've finished tracing through the interactions with the health code so I've seen the serial number tracking and the purpose of the tree. Overall multiplexing the different queues (commands to be despatched, commands in flight, completed commands) by the flags works and I can't think of a better way to handle the `health_alarm_wait_for_execution()` when it is keyed by the serial number. Keeping the mutex/cond-var pair inside the cmd_info avoid a number of potential race hazards.
> 
> My question about the `static inline` wasn't really about how to define static inlined functions - why should this one be statically inlined when we are relying on LTO for everything else?

Well, first of all not all builds in our CI/CD use LTO. My reason for using `static inline` has always been dead simple:
- If you want to inline a function for performance, the safest way is to use `static inline`

I don't give any thought to this, it's automatic. Should I inline? Then just use `static inline`. Apparently netdata also uses a lot of `static inline`:
https://github.com/netdata/netdata/blob/0fba85e2c20add69546cefbf37bb2033d2d1e052/libnetdata/inlined.h

amoss:
Sure, but I don't think that any of the other inlined functions then deliberately use a global variable on the implicit assumption that it will be a different buffer inside a different process. So my thought process would be is the performance difference important enough to warrant the break in encapsulation and the introduced gotcha in ongoing maintenance?

Both the protocol buffer and the spawn_cmd_queue should be static. In general any data that is not being deliberately shared between translation units should be static. One is easy to change and the other can't be because of this inlined procedure.

mfundul:
> 
> 
> Sure, but I don't think that any of the other inlined functions then deliberately use a global variable on the implicit assumption that it will be a different buffer inside a different process. So my thought process would be is the performance difference important enough to warrant the break in encapsulation and the introduced gotcha in ongoing maintenance?
> 
> Both the protocol buffer and the spawn_cmd_queue should be static. In general any data that is not being deliberately shared between translation units should be static. One is easy to change and the other can't be because of this inlined procedure.

I can always pass the arguments by reference to make it more maintainable, I just found it convenient to keep the function source code as small as possible.

amoss:
Passing the arguments by reference and making the originals static would solve the encapsulation problem.

I've finished reviewing all the code - no problems that I can find. I've moved on to doing some exploratory testing now. I've setup a synthetic chart that will trigger alarms so that I can flood the alarm notifications from a single agent.

amoss:
Testing results are a mixed bag. I used the synthetic chart generator in `aclk/tests` to load the system without hitting a cpu bottleneck. To replicate this use this fake charts.d:

```
#!/usr/bin/env bash

echo "CHART alarmtest.blah '' 'Generate alarm notifications' 'units' alarmtest alarmtest lines 1 1"
for i in {1..2000}
do
   echo "DIMENSION test$i '' percentage-of-absolute 1 1"
done

while true; do
    echo "BEGIN alarmtest.blah 1000000"
    for i in {1..2000}
    do
        echo "SET test$i = 0"
    done
    echo "END"
    sleep 1
    echo "BEGIN alarmtest.blah 1000000"
    for i in {1..2000}
    do
        echo "SET test$i = 99"
    done
    echo "END"
    sleep 1
done
```

By overwriting the contents of `aclk/tests/fake-charts.d.plugin` and then running `aclk/tests/install-fake-charts.d.sh` to put it into the installation prefix. Once this is running a synthetic chart with 2000 dimensions is created where each dimension just toggles between 0 and 99 every second.

The alarm config is:
```
alarm: flood
on: alarmtest.blah
lookup: max -1s foreach test*
units: %
every: 1s
warn: $this > 50
crit: $this > 90
info: trying to flood alarms for a test case
to: sysadmin
```
Put this in `etc/health.d/blah.conf` inside the installation prefix. I've checked it with 2000, 5000 and 10000 alarms. What happens is that two threads hit 100% cpu while the rest of the system stays idle. Netdata will run for a few seconds before the dashboard grinds to a halt and the agent gets stuck in a state where it cannot exit gracefully and needs a `kill -9`. This happens while the overall cpu load is only 25% and the rest of the system is behaving normally.

I would say that this work is a partial success - it fixes a major performance issue that we had, but overall the agent was not designed for this kind of load so the bottleneck has simply shifted to another part of the system. My initial impression is that we should merge this patch, and the health AoR should look at the two overloaded threads and do an architectural review of the entire component. This work can be folded into a feature that is coming up in the agent later this year that concerns the health component. By merging the spawn server we free @mfundul up to finish looking eBPF and then concentrate on the database improvements that we need. The health AoR is larger than it used to be so there should be enough capacity to do a proper architectural review and some large scale fixes.

@cosmix - what are your thoughts on this?

PR #8412, opened 2020-03-16
Adds missing files to ACLK CMake build

underhood:
##### Summary
aclk_common.c/aclk_common.h missing from CMAKE list of files for ACLK.

Noticed when reviewing #8411

##### Test Plan
ACLK builds with CMAKE


underhood:
CI stuck on CLA sign status (didn't  find other way to restart it)

PR #8426, opened 2020-03-18
Fixed JSON parsing

### LINKED ISSUE #8349: Fix the JSON parsing issues on the ACLK.
#### Summary

We need to fix problems that were observed during the smoke-test. 
* The uninitialised variables inside the JSON parser that valgrind is complaining about.
* The behavior depends on which json parser is in use. If the JSONC external dependency is available then it works, it seems to break on the internal version.

Currently there are somehwere up to three different json parsers in use in the project in different places. This is a problem that we need to resolve but we cannot do in this sprint because of time. 

This will be fixed now, and then after March 25 we will clean this up properly in #7701.

stelfrag:
Fixes #8349
##### Summary
When the internal JSON parser was used the callback function supplied was not called in some cases so the calling function would not get the complete parsed results.

##### Component Name
ACLK

##### Test Plan
Before the PR is applied
- Start a claimed agent without the optional json-c library
- When attempting to establish the ACLK, notice in the logfile error messages 
  `Could not retrieve challenge from auth response:.....`

After the PR is applied
- Start a claimed agent without the optional json-c library
- Challenge response completes successfully
- ACLK is established

### DIFF
--- a/aclk/aclk_lws_https_client.c
+++ b/aclk/aclk_lws_https_client.c
@@ -35,7 +35,7 @@ static int simple_https_client_callback(struct lws *wsi, enum lws_callback_reaso
             return -1;
         }
         ptr = perconn_data->data;
-        n = perconn_data->data_size;
+        n = perconn_data->data_size - 1;
         if (lws_http_client_read(wsi, &ptr, &n) < 0)
             return -1;
         ptr[n] = '\0';
--- a/aclk/aclk_lws_https_client.c
+++ b/aclk/aclk_lws_https_client.c
@@ -38,6 +38,7 @@ static int simple_https_client_callback(struct lws *wsi, enum lws_callback_reaso
         n = perconn_data->data_size;
         if (lws_http_client_read(wsi, &ptr, &n) < 0)
             return -1;
+        ptr[n] = '\0';
         return 0;
     case LWS_CALLBACK_WSI_DESTROY:
         debug(D_ACLK, "LWS_CALLBACK_WSI_DESTROY");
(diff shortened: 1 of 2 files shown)


aclk/aclk_lws_https_client.c
```diff
@@ -38,6 +38,7 @@ static int simple_https_client_callback(struct lws *wsi, enum lws_callback_reaso
         n = perconn_data->data_size;
         if (lws_http_client_read(wsi, &ptr, &n) < 0)
             return -1;
+        ptr[n] = '\0';
```
amoss: This writes off the end of the allocated space, change L38 to  `n = perconn_data->data_size-1`;

stelfrag: The n is overwritten by `lws_http_client_read` to the actual bytes read. That's what I want null terminated.

amoss:
Is there any impact on the code in `health.c` L116 and L131?

aclk/aclk_lws_https_client.c
amoss: Yes, I understand that. But the contract with `lws_http_client_read` is that it can read up to `n` bytes, so if it fills the buffer provided then `n` will be unchanged and the null will be written off the end of the allocated space.

stelfrag: You are right I will give it one less byte

stelfrag: Pushed the change

stelfrag:
> Is there any impact on the code in `health.c` L116 and L131?

I am investigating this both with and without json-c.

stelfrag:
> > Is there any impact on the code in `health.c` L116 and L131?
> 
> I am investigating this both with and without json-c.

It does not have any impact on the silencer code that is using it. 

Having a callback triggered by a `JSON_ENTRY`s callback function shouldn't require to set it again in the structure. However this fixes things as follows:

- In the case of the internal json parser the JSON_ENTRY has a NULL callback (before this PR) so L116, L131 fix that
- In the case of the json-c parser the JSON_ENTRY has an undefined value (due to json.c L433) so L116, L131 fix that as well.

PR #8429, opened 2020-03-18
Report Why ACLK build failed

### LINKED ISSUE #8050: During the agent installation, if the ACLK fails to be built, show an error message to the user
#### Summary

Ensure that the requirement in product#282 are met.

- [x] If ACLK error reporting not set yet by #8051 define it with this issue
- [x] If ACLK build fails make it prominent to the user so he knows about it. It should not be just small line easily overlooked in log. @amoss will speak to @jacekkolasa about this separately.
- [x] Netdata should log on startup it is build without ACLK
- [x] Report failure to Cloud to same endpoint #8051 
- [x] Respect DO_NOT_TRACK environment variable

(Some of this will be covered by PR 8025 but there are fresh requests at the bottom of the discussion).

underhood:
##### Summary
Report reason why ACLK build failed. Respect opt out.

Fixes #8050

##### Test Plan
Simulate ACLK build failure (e.g. by removing libomosq.a from `externaldeps` folder and rerun configure without running installer).

You should see events in Google Analytics.

Keep in mind changes merged by Andrew meanwhile (you must pass --with-cloud to configure instead of ACLK=yes). If cloud is not enabled by purpose do not report anything.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -562,6 +562,7 @@ if test "$enable_cloud" = "yes"; then
         HAVE_libmosquitto_a="yes"
     else
         HAVE_libmosquitto_a="no"
+        AC_DEFINE([ACLK_NO_LIBMOSQ], [1], [Libmosquitto.a was not found during build.])
     fi
     AC_MSG_RESULT([${HAVE_libmosquitto_a}])
 
@@ -571,6 +572,7 @@ if test "$enable_cloud" = "yes"; then
         HAVE_libwebsockets_a="yes"
     else
         HAVE_libwebsockets_a="no"
+        AC_DEFINE([ACLK_NO_LWS], [1], [libwebsockets.a was not found during build.])
     fi
     AC_MSG_RESULT([${HAVE_libwebsockets_a}])
 
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -1342,6 +1342,7 @@ int main(int argc, char **argv) {
     // ------------------------------------------------------------------------
     // Report ACLK build failure
 #ifndef ENABLE_ACLK
+    error("This agent doesn't have ACLK.");
     char filename[FILENAME_MAX + 1];
     snprintfz(filename, FILENAME_MAX, "%s/.aclk_report_sent", netdata_configured_varlib_dir);
     if( netdata_anonymous_statistics_enabled > 0 && access( filename, F_OK ) ) { // -1 -> not initialized
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -1338,6 +1338,15 @@ int main(int argc, char **argv) {
     netdata_ready = 1;
 
     send_statistics("START", "-",  "-");
+#ifndef ENABLE_ACLK
+    send_statistics("ACLK_DISABLED", "-", "-");
+#ifdef ACLK_NO_LWS
+    send_statistics("BUILD_FAIL_LWS", "-", "-");
+#endif
+#ifdef ACLK_NO_LIBMOSQ
+    send_statistics("BUILD_FAIL_MOSQ", "-", "-");
+#endif
+#endif
 
     // ------------------------------------------------------------------------
     // unblock signals
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -1345,7 +1345,7 @@ int main(int argc, char **argv) {
     error("This agent doesn't have ACLK.");
     char filename[FILENAME_MAX + 1];
     snprintfz(filename, FILENAME_MAX, "%s/.aclk_report_sent", netdata_configured_varlib_dir);
-    if( netdata_anonymous_statistics_enabled > 0 && access( filename, F_OK ) ) { // -1 -> not initialized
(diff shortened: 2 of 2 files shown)


amoss:
The parts of the configure that you are updating have changed recently, please rebase and rework around the new structure.

stelfrag:
- Removed `cmake`
- Executed with `--cloud-testing`

Result at the end of compilation process
```
The following non-fatal errors were encountered during the installation process:

* Failed to build libwebsockets. The install process will continue, but you may not be able to connect this node to Netdata Cloud.
```

however the agent was build with ACLK enabled because I had libwebsockets.a available from previous run.

I am guessing the `may not be able`  in that message leaves some hope to work?

underhood:
Well this PR only makes the message more visible on request of product. Doesn't change anything else incl. wording.

The installer leaving the files there from previous build is good question and maybe should be discussed with @Ferroin. (We have option always rebuild from scratch or installer detecting there is one built already). Both are hovewer out of scope of this issue.

underhood:
Hi, that message is from installer in master. Not related to this PR at all. We can ask @Ferroin if we want better handling in installer.

amoss:
It's fine - it requires the user to explicitly reach that state and the message reads "may".

stelfrag:
I am seeing that a file is created `aclk_report_sent` so that the report is not send twice. 

I can't find anything in the specs that specifies that. Did I miss something?

underhood:
Well the point is to know how many ACLK builds failed.

daemon/main.c
```diff
@@ -1339,6 +1339,27 @@ int main(int argc, char **argv) {
 
     send_statistics("START", "-",  "-");
 
+    // ------------------------------------------------------------------------
+    // Report ACLK build failure
+#ifndef ENABLE_ACLK
+    error("This agent doesn't have ACLK.");
+    char filename[FILENAME_MAX + 1];
+    snprintfz(filename, FILENAME_MAX, "%s/.aclk_report_sent", netdata_configured_varlib_dir);
```
stelfrag: I think we need to remove the file and send everytime (just as we do with START/ EXIT)

The file is never removed (as far as I can see) even if there is a successful build.

If we have a successful build followed by a failed one (say a month later) we want to know that.

underhood: Hmmm, well in such case we might want to get product involved as the task seems to be different.

amoss: No this does not seem to be correct. We want to know where installation failed because the dependency was unavailable. The case where it worked, but the user is recompiling it in the same source tree and it then fails does not seem to be relevant to the ask. What they care about is that if somebody tries to install it and the ACLK can't be enabled then we find out about it.

amoss: I understand the ask as being per-installation rather than per-compile.

configure.ac
```diff
@@ -562,6 +562,7 @@ if test "$enable_cloud" = "yes"; then
         HAVE_libmosquitto_a="yes"
     else
         HAVE_libmosquitto_a="no"
+        AC_DEFINE([ACLK_NO_LIBMOSQ], [1], [Libmosquitto.a was not found during build.])
```
amoss: Why not just put the HAVE_ flags straight into the config.h?

daemon/main.c
underhood: then we have to delete it from installer

amoss: I don't think that it needs to be deleted at all - if the failure has been reported then there is no need to report it again from the same machine.

configure.ac
underhood: can but don't see any advantage

configure.ac
```diff
@@ -606,6 +608,8 @@ if test "$enable_cloud" = "yes"; then
     fi
 
     AC_MSG_RESULT([${enable_aclk}])
+else
```
amoss: Why are you overwriting enable to no if it is not yes? This will interact very badly with the upcoming changes to the feature flag where the default will be a third option that must be preserved,

underhood: There was problem that @Ferroin identified if he passed flag --enable-aclk and ACLK environment variable was not set it would still pick up that flag and define #define ENABLE_ACLK

It might be obsolete now....

amoss: We've taken the env variable out completely and it is important that you do not overwrite the setting as it will persist.

underhood:
rebased on master and skipped the commit that was adding the override

daemon/main.c
```diff
@@ -1339,6 +1339,27 @@ int main(int argc, char **argv) {
...
+        send_statistics("ACLK_DISABLED", "-", "-");
+#ifdef ACLK_NO_LWS
+        send_statistics("BUILD_FAIL_LWS", "-", "-");
+#endif
+#ifdef ACLK_NO_LIBMOSQ
+        send_statistics("BUILD_FAIL_MOSQ", "-", "-");
+#endif
+        int fd = open(filename, O_WRONLY | O_CREAT | O_TRUNC, 444);
+        if (fd == -1)
+            fatal("Cannot create file '%s'. Please fix this.", filename);
```
amoss: This is not an appropriate way to handle an unexpected situation. This could break the agent on many configurations. Please remove the `fatal`.

PR #8430, opened 2020-03-18
Add the new cloud info in the info endpoint

amoss:
##### Summary
Added the fields to the `/api/v1/info` endpoint that the dashboard needs to display the status of the cloud support.

Changed the way the flags propagate through the installer and configure script.

**The ACLK environmental variable is no longer used**. It has been removed in preparation for flipping the feature flag. Supply `--cloud-testing` to the installer to enable everything. This is an internal only option as access to the cloud platform requires iauthentication.

##### Component Name
ACLK
web/api

##### Test Plan

* Tested with a claimed / unclaimed agent.
* Tested with the cloud disabled / enabled.
* Checked the availability flag tracks the dependency state.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -161,14 +161,27 @@ AC_ARG_ENABLE(
     ,
     [enable_jsonc="detect"]
 )
-if test "${ACLK}" = "yes"; then
+
+# -----------------------------------------------------------------------------
+# Check if cloud is enabled and if the functionality is available
+
 AC_ARG_ENABLE(
-    [aclk],
-    ,
-    [aclk_required="${enableval}"],
-    [aclk_required="detect"]
+    [cloud],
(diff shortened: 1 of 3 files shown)


amoss:
> I'm not sure I understand the details of this PR.
> So, (correct me if I'm wrong) to me this PR seems like implying the ACLK availability from the configuration or installation settings. Shouldn't the meaning of this be:
> Show `available` if the the link is actually established/up and running and `unavailable` if the link has connection issues, firewall banned etc? This is also why we'll show a popup notification when these states change right?
> Else, it looks ok I think. Let me know and if I'm wrong we can proceed.

This is the intended functionality. The description is on the linked issue. The idea is that if the dependencies failed to build then we want it to be obvious to the user that the cloud is not available. If the user has explicitly disabled the cloud then we want to strip any cloud UI from the dashboard.

PR #8454, opened 2020-03-22
bind to should be in [web] section and update netdata.service.v235.in too

amishmm:
##### Summary
1. In netdata.conf file, bind to is mentioned under [global] section but it should be under [web] section as per documentation here, https://docs.netdata.cloud/web/server/

2. Recently netdata.service.in was updated but same changes were not made to netdata.service.v235.in file. This PR does that.

### DIFF
--- a/system/netdata.conf
+++ b/system/netdata.conf
@@ -16,9 +16,6 @@
     # the default database size - 1 hour
     history = 3600
 
-    # by default do not expose the netdata port
-    bind to = localhost
-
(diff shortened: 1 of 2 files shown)


amishmm:
Hello, PTAL. @cosmix @ktsaou Thank you

prologic:
@Ferroin Let's get this approved ASAP and merged. This sucked that the default config we were shipping was wrong and we both missed this :/

PR #8456, opened 2020-03-22
Report ACLK Connection Failure

### LINKED ISSUE #8051: Report cases where the Agent Cloud link is blocked by firewall
#### Discussion for the planning meeting on Wednesday

* This needs to be groomed before we can estimate the complexity.
* If this is needed by March 25 then this causes a problem in how it can be placed in the upcoming sprint.

#### Summary

Ensure that the requirements of product#243 are met.

- [x] Define endpoint for ACLK connection failures with SRE team if not done by #8050 already. Endpoint will be GAnalytics
- [x] Implement reporting to the endpoint of cases like:
    - SSL update failed
    - WebSockets update failed
- [x] do not send anything if `DO_NOT_TRACK` env var is set

underhood:
##### Summary
If ACLK fails to connect report it in case telemetry is enabled (e.g. respect DO_NOT_TRACK of course)

We send up to last 10 events of LWS to the GA to be able to analyse at what point connection failed.

Fixes #8051 

##### Component Name
ACLK
##### Test Plan
Fail connection in different ways. Look at the reports in Google Analytics

Example of event data when connection failed on certificate verification: "30,1,58,19,29"

### DIFF
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -383,7 +383,7 @@ static int aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reas
     static int lws_shutting_down = 0;
     int i;
 
-    for (i = ACLK_LWS_CALLBACK_HISTORY; i > 0; i--)
+    for (i = ACLK_LWS_CALLBACK_HISTORY - 1; i > 0; i--)
         engine_instance->lws_callback_history[i] = engine_instance->lws_callback_history[i - 1];
     engine_instance->lws_callback_history[0] = (int)reason;
 
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -264,11 +264,15 @@ int aclk_lws_wss_connect(char *host, int port)
 {
     struct lws_client_connect_info i;
     struct lws_vhost *vhost;
+    int n;
 
     if (!engine_instance) {
         return aclk_lws_wss_client_init(host, port);
         // PROTOCOL_INIT callback will call again.
     }
+
+    for (n = 0; n < ACLK_LWS_CALLBACK_HISTORY; n++)
+        engine_instance->lws_callback_history[n] = 0;
 
     if (engine_instance->lws_wsi) {
         error("Already Connected. Only one connection supported at a time.");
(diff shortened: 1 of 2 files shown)


aclk/aclk_lws_wss_client.c
```diff
@@ -427,6 +461,8 @@ static int aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reas
             return -1;                       // the callback response is ignored, hope the above remains true
         case LWS_CALLBACK_WSI_DESTROY:
             aclk_lws_wss_clear_io_buffers(engine_instance);
+            if (!engine_instance->websocket_connection_up)
```
stelfrag: This is not always called. When I messed the password (after the challenge) the connection was not established.

Why does it expect to have this `websocket_connection_up == 0` when the code that sets this to 0 is L467

underhood: We want report to be sent only if WSI_DESTROY is called when there was not successful connection yet.
Idea is to report when ACLK link could not be established.

underhood: Point of this is to have an idea in how many cases ACLK cannot connect (e.g. by blocked firewall).

underhood: we are diagnosing if connection to the cloud was possible (not blocked by firewall) as soon as connection is established we know that is OK and we don't have to report anything.
It doesn't try to diagnose the MQTT failures that might happen after.

amoss: Why not use the state of `aclk_connected`? This is the high-level flag that the link was established correctly. Your logic that any WSI_DESTROY before success is a form of failure would apply better to this flag.

stelfrag:
It does report on connection failures
Example:
```
Could not connect MQTT over WSS server "dev.private.netdata.cloud:443". LwsReason:"HS: ws upgrade response not 101
```

will report 

`'ACLK_CONN_FAIL' 'FAIL' '30,45,1,24,58,58,58,19,29'`

underhood:
Yes i decided to log callbacks instead of handpicking simulated situations (my original plan). This way we should have reports for all cases and we can look at the callbacks to analyse (this is what we do anyway when we debug)

aclk/aclk_lws_wss_client.c
underhood: that is too late... we do not care about MQTT here. If connection TLS upgrade and then upgrade to Websockets succeeds then connection succeeded (proxy or firewall didn't block us)

amoss: Yes we do - that is the whole point.

aclk/aclk_lws_wss_client.c
```diff
@@ -330,12 +334,42 @@ static const char *aclk_lws_callback_name(enum lws_callback_reasons reason)
...
 static int aclk_lws_wss_callback(struct lws *wsi, enum lws_callback_reasons reason, void *user, void *in, size_t len)
 {
     UNUSED(user);
     struct lws_wss_packet_buffer *data;
     int retval = 0;
     static int lws_shutting_down = 0;
+    int i;
+
+    for (i = ACLK_LWS_CALLBACK_HISTORY; i > 0; i--)
+        engine_instance->lws_callback_history[i] = engine_instance->lws_callback_history[i - 1];
```
amoss: There is an off-by-one error, loop should start on `ACLK_LWS_CALLBACK_HISTORY-1`

underhood:
pushed - resolve merge conflict

PR #8462, opened 2020-03-23
update netdata.conf for debian build

and0x000:
##### Summary

`netdata.conf` in directory `system` was updated but debian build uses old version. The old version is potentially insecure, as it binds to all interfaces by default, thus exposing netdata on public interfaces.

##### Component Name

packaging

### DIFF
--- a/contrib/debian/netdata.conf
+++ b/contrib/debian/netdata.conf
@@ -1,16 +1,28 @@
-# NetData Configuration
-
-# The current full configuration can be retrieved from the running
-# server at the URL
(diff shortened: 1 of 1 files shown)


prologic:
Looks like this got superseded by #8463

and0x000:
> 
> 
> Looks like this got superseded by #8463

Yes, had wrong mail address in my commit, therefore the CLAassistant would not let me sign the CLA.

prologic:
No worries :)

Issue #8405, opened 2020-03-15
Netdata debian package fails to start after fresh install on buster


toadjaune:
##### Bug report summary

On debian Buster, when installing netdata through the stable packages, it fails to start with `Cannot cd to directory '/var/cache/netdata'`.
The edge repository is not affected.

##### OS / Environment
```
Linux buster 4.19.0-8-amd64 #1 SMP Debian 4.19.98-1 (2020-01-26) x86_64 GNU/Linux
/etc/os-release:PRETTY_NAME="Debian GNU/Linux 10 (buster)"
/etc/os-release:NAME="Debian GNU/Linux"
/etc/os-release:VERSION_ID="10"
/etc/os-release:VERSION="10 (buster)"
/etc/os-release:VERSION_CODENAME=buster
/etc/os-release:ID=debian
/etc/os-release:HOME_URL="https://www.debian.org/"
/etc/os-release:SUPPORT_URL="https://www.debian.org/support"
/etc/os-release:BUG_REPORT_URL="https://bugs.debian.org/"
```
This is a clean system created with vagrant/libvirt to reproduce this issue, nothing else has been installed on it.

##### Netdata version
```
netdata v1.20.0
```
##### Component Name
packaging (I assume)

##### Steps To Reproduce
I used the following Vagrantfile to create the debian system, but any up-to-date debian buster system should encounter the issue :
```
Vagrant.configure(2) do |config|
  config.vm.define "buster" do |subconfig|
    subconfig.vm.box = "debian/buster64"
    subconfig.vm.hostname = "buster"
  end
end
```
Reproduction steps : 
```
root@buster:~# apt install -y curl
[…]
root@buster:~# curl -s https://packagecloud.io/install/repositories/netdata/netdata/script.deb.sh | sudo bash
[…]
root@buster:~# apt install -y netdata
[…]
Setting up netdata (1.20.0) ...
dpkg-statoverride: warning: --update given but /var/cache/netdata does not exist
dpkg-statoverride: warning: deprecated --force option; use --force-all instead
dpkg-statoverride: warning: --update given but /var/lib/netdata/registry does not exist
Created symlink /etc/systemd/system/multi-user.target.wants/netdata.service → /lib/systemd/system/netdata.service.
Processing triggers for systemd (241-7~deb10u3) ...
Processing triggers for libc-bin (2.28-10) ...
W: Operation was interrupted before it could finish
root@buster:~# systemctl status netdata.service
[...]

PR #8463, opened 2020-03-23
update netdata.conf for debian build

and0x000:
##### Summary

netdata.conf in directory system was updated but debian build uses old version. The old version is potentially insecure, as it binds to all interfaces by default, thus exposing netdata on public interfaces.

##### Component Name

packaging


prologic:
Hi @and0x000 I finally have a PR #8468 up for #8405 -- Would you be willing to amend your PR here to reflect the changes I'm making there oto and use the same pattern? I'd rather avoid having multiple copies of files, it gets confusing which is which :)

and0x000:
> Hi @and0x000 I finally have a PR #8468 up for #8405 -- Would you be willing to amend your PR here to reflect the changes I'm making there oto and use the same pattern? I'd rather avoid having multiple copies of files, it gets confusing which is which :)

Yes, a single source of truth is more favorable, I totally agree with you.
I rebased onto your branch for #8468 and used the same schema to pull the netdata.conf file

prologic:
> If you want to include this into your PR right away, feel free to do so and discard my PR.

Thank you. I will! I'll cherry-pick your commits and preserve your authorship. Thank you! I'm still testing and have found further problems that I will fix before getting review and merging.

prologic:
cherry-picked into #8468

PR #8470, opened 2020-03-24
Make mosq and LWS lib build fails more prominent

underhood:
##### Summary
Minor -> if Libmosquitto or Libwebsockets fail to build make the warning at the end of installer more prominent (yellow bold).

On request of Product

##### Test Plan
force one of the libs build failure or add `false && \` after line 502 of build script (to force script to think it failed).

Result should look like this:
Screenshot from 2020-03-24 10-55-50

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -48,6 +48,10 @@ defer_error() {
   NETDATA_DEFERRED_ERRORS="${NETDATA_DEFERRED_ERRORS}\n* ${1}"
 }
 
+defer_error_highlighted() {
+  NETDATA_DEFERRED_ERRORS="${TPUT_YELLOW}${TPUT_BOLD}${NETDATA_DEFERRED_ERRORS}\n* ${1}${TPUT_RESET}"
+}
+
 print_deferred_errors() {
   if [ -n "${NETDATA_DEFERRED_ERRORS}" ] ; then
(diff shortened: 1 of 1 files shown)


Ferroin:
This might just be me, but I actually think we should probably just highlight all the deferred errors, as well as the bit of header text that gets printed if we encountered any.

Regardless of what error messages got deferred, they're important enough that we wanted them at the end of the install output, so just universally calling more attention to them makes more sense to me than only calling attention to some.

@netdata/product Could we get an opinion on this? Right now we're deferring error messages for the Netdata Cloud dependencies, and for the go.d plugin (lack of which means significantly reduced functionality in Netdata).

underhood:
@dim08 you requested this. Can you give opinion on the coment from @Ferroin ?

dim08:
Agree with Ferroin. Please don't do any changes right now. I will create an epic on UX changes for  Q2.2020 and group all relevant issues under it.

underhood:
Ok then i am merging

Ferroin:
@dim08 I fail to see how it makes sense to put the changes I suggested off. They actually would have made the overall code changes here simpler, easier to review, and reduced code churn (when the do get implemented, that newly added function is getting removed).

Unless @underhood and I both misunderstood your meaning.

PR #8472, opened 2020-03-24
Added Docker build arg to pass extra options to installer.

Ferroin:
##### Summary

This adds a build argument to the main Dockerfile to allow passing extra options to the netdata-installer script during the build process.

The value specified in the `EXTRA_INSTALL_OPTS` argument wil be added verbatm to the netdata-installer.sh script. The arguments `--dont-start-it` and `--dont-wait` are still implicitly passed to the script as those are required to get a correctly functioning Docker image.

##### Component Name

area/packaging

##### Test Plan

Regular builds verified on PR checks, locally tested with arbitrary argument combinations to verify that the bild works correctly.

##### Additional Information

This originated as a way to build Netdata Cloud support in the Docker images while it's still being tested, but is intended to be kept moving forwards as an easy way to create custom Docker images and simplify testing optional features from Docker.

To actually use this, use a command like the following from the root of the source tree:

    docker build . --build-arg EXTRA_INSTALL_OPTS="--foo"

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -12,6 +12,10 @@ ARG RELEASE_CHANNEL=nightly
 
 ENV JUDY_VER 1.0.5
 
+ARG EXTRA_INSTALL_OPTS
+
+ENV EXTRA_INSTALL_OPTS=$EXTRA_INSTALL_OPTS
+
 # Copy source
 COPY . /opt/netdata.git
 WORKDIR /opt/netdata.git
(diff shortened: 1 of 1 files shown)


amoss:
Why an ARG and an ENV - I've not seen both used together like that before - what does it do?

Ferroin:
> Why an ARG and an ENV - I've not seen both used together like that before - what does it do?

ARG is only accessible during the build process and can be set on the CLI at build time.

ENV is accessible al the way through to runtime (it will be in the default environment you get when you run the container), but AFAIK can't be set on the CLI at build time.

The main reason for having both here is so that we can see what extra options were passed to the installer at build time.

Ferroin:
@knatsakis I'd like to get an approval from you on this one before merging, as you're the primary user for this right now.

Issue #8471, opened 2020-03-24
Using python modules in new plugins


marcomusso:
##### Question summary

I managed to write my own python plugin for a custom check and it works fine in debug/trace mode (as `netdata`) according to the docs (snippet):

```
2020-03-24 13:01:56: python.d DEBUG: logcollector_keepalive_message_count[logcollector_keepalive_message_count] :
started, update frequency: 30
CHART logcollector_keepalive_message_count.logcollector_keepalive_message_count '' 'Keepalive message count' 'mess
ages' 'count' 'count' line 20000 30 '' 'python.d.plugin' 'logcollector_keepalive_message_count'
DIMENSION 'count' 'count' absolute 1 1 ' '

BEGIN logcollector_keepalive_message_count.logcollector_keepalive_message_count 0
SET 'count' = 3
END

BEGIN netdata.runtime_logcollector_keepalive_message_count 0
SET run_time = 34
END
```

When running from the agent I get, in `error.log`:

```
2020-03-24 12:02:23: python.d WARNING: plugin[main] : [logcollector_keepalive_message_count] error on loading source : ImportError('cannot import name Elasticsearch',), skipping it
```

This is of course because I'm using the Elasticsearch python module which it seems not being recognised when running from the agent (but it's installed system-wide).

##### OS / Environment

CentOS 7.4.1708
netdata 1.20.0-1 installed as rpm

##### Component Name

python.d scrip

##### Expected results

I was hoping that the debug/trace was enough to guarantee the script to work when executed fro the agent.

Ferroin:
There's a known issue with the handling of the Python module search paths being different when running as a plugin from those when just manually running Python. I don't remember if there's an issue open about it or not, but this is likely what you're running into.

marcomusso:
But is there a documentation on how to add python modules or shall I consider it a missing feature or a bug?

Ferroin:
Basic documentation for the plugins can be found at https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/README.md#how-to-write-a-new-module. That doesn't include info on the above-mentioned issue with the module search paths because it's not a very well understood problem (we know it's there, just not why it's happenng or how to fix it). Behaviorally, the issue is that the Python collectors can't reliably see modules installed through `pip`, even if they're installed as the Netdata user, but (usually) has no issues finding ones installed through the system package manager.

marcomusso:
[...]

PR #8474, opened 2020-03-24
python.d.plugin: add prefix to the module name during loading source file

ilyam8:
##### Summary

Fixes: #8471

##### Component Name

`collectors/python.d.plugin`

##### Test Plan

There is no cross pythond module importing, the fix doesnt affect any module/pythond itself.

However i installed the branch and tested it. I tested it debug mode for both py2 and 3 - no problems.

### DIFF
--- a/collectors/python.d.plugin/python.d.plugin.in
+++ b/collectors/python.d.plugin/python.d.plugin.in
@@ -42,12 +42,10 @@ except ImportError:
 
 PY_VERSION = sys.version_info[:2]  # (major=3, minor=7, micro=3, releaselevel='final', serial=0)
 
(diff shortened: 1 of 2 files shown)


ilyam8:
@stelfrag @vlvkobal @mfundul one more time please, i forgot to remove the prefix in SimpleService, it changes module name in logs

```sh
2020-03-25 13:19:03: python.d DEBUG: pythond_apache[localipv6] : urllib3 version: 1.25.8
```

=>

```sh
2020-03-25 13:19:03: python.d DEBUG: apache[localipv6] : urllib3 version: 1.25.8
```

PR #8475, opened 2020-03-24
Installer creates claim.d but is run as root, patch to correct ownership

amoss:
#### Summary

Fixes problems with claiming agents in the local dev env.

##### Component Name
packaging
claiming

##### Test Plan

Claiming an agent in the local dev env.

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -89,6 +89,8 @@ RUN \
         /var/cache/netdata \
         /var/lib/netdata \
         /var/log/netdata && \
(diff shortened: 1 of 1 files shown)


packaging/docker/Dockerfile
```diff
@@ -89,6 +89,8 @@ RUN \
         /var/cache/netdata \
         /var/lib/netdata \
         /var/log/netdata && \
```
nicolasparada:
```suggestion
        /var/log/netdata
```

PR #8489, opened 2020-03-25
Relaxes SSL checks for testing

underhood:


underhood:
@Ferroin @prologic can you help with CI here?

PR #8499, opened 2020-03-26
Enhanced ACLK header payload to include timestamp-offset-usec

### LINKED ISSUE #8503: Enhance the ACLK JSON header with a new timestamp-offset-usec field
#### Summary

Add a new field to the ACLK JSON header `timestamp-offset-usec` that will provide microseconds offset to the `timestamp` in the range 0 ... 999999

Example:

Existing header

```
{ 
    "type": "status-change",
    "msg-id": "ef4b559e-6f01-4e00-95ff-adeff620d559",
    "timestamp": 1585240780,
    "version": 1,
    "payload": ...
}
```

new header 

```
{
    "type": "status-change",
    "msg-id": "ef4b559e-6f01-4e00-95ff-adeff620d559",
    "timestamp": 1585240780,
    "timestamp-offset-usec": 629572,
    "version": 1,
    "payload": ...
}
```

stelfrag:
Fixes #8503
#### Summary
Added a new field in the ACLK header payload to include `timestamp-offset-usec`, that provides microseconds offset to the timestamp

##### Component Name
ACLK

##### Test Plan
- Agent with ACLK enabled
- All messages pushed to the cloud will have the header with the new field

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -954,9 +954,11 @@ static void aclk_main_cleanup(void *ptr)
         // Wakeup thread to cleanup
         QUERY_THREAD_WAKEUP;
         // Send a graceful disconnect message
-        time_t time_created = now_realtime_sec();
-        usec_t time_created_offset_usec = (now_realtime_usec() % USEC_PER_SEC);
         char *msg_id = create_uuid();
+
+        usec_t time_created_offset_usec = now_realtime_usec();
+        time_t time_created = time_created_offset_usec / USEC_PER_SEC;
+        time_created_offset_usec = time_created_offset_usec % USEC_PER_SEC;
 
         snprintfz(
             payload, 511,
@@ -1515,8 +1517,6 @@ void aclk_shutdown()
 inline void aclk_create_header(BUFFER *dest, char *type, char *msg_id)
 {
     uuid_t uuid;
-    time_t time_created;
-    usec_t time_created_usec_offset;
     char uuid_str[36 + 1];
 
     if (unlikely(!msg_id)) {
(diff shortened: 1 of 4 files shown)


aclk/agent_cloud_link.c
```diff
@@ -955,16 +955,18 @@ static void aclk_main_cleanup(void *ptr)
         QUERY_THREAD_WAKEUP;
         // Send a graceful disconnect message
         time_t time_created = now_realtime_sec();
+        usec_t time_created_offset_usec = (now_realtime_usec() % USEC_PER_SEC);
```
amoss: This will be non-monotonic if the second rolls over between the two calls.

stelfrag: Will fetch usecs only and do the math

aclk/agent_cloud_link.c
```diff
@@ -954,17 +954,21 @@ static void aclk_main_cleanup(void *ptr)
         // Wakeup thread to cleanup
         QUERY_THREAD_WAKEUP;
         // Send a graceful disconnect message
-        time_t time_created = now_realtime_sec();
         char *msg_id = create_uuid();
 
+        usec_t time_created_offset_usec = now_realtime_usec();
+        time_t time_created = time_created_offset_usec / USEC_PER_SEC;
+        time_created_offset_usec = time_created_offset_usec % USEC_PER_SEC;
```
fracasula: Does using a function instead of repeating these 3 lines of code three times have a performance impact that is not negligible? I don't have any C expertise and I know the agent's code is higly optimized so I'm asking merely out of curiousity.

stelfrag: Good point @fracasula  but these 3 lines are used repeatedly only when building a header (with each message that goes to the cloud).

The other two occasions are on MQTT connect to define a LWT and when detecting a graceful shutdown to send a `graceful` message instead of `unexpected` (those two cases are rarely used)

PR #8501, opened 2020-03-26
Cleans up cloud config files [agent_cloud_link] -> [cloud]

underhood:
##### Summary
Removes config section `[agent_cloud_link]` and move it's keys under `[cloud]` so that they are all in a common place.

##### Component Name

Agent/ACLK

##### Additional Information
@andrew I remember we wanted to put it all under `[cloud]` if not true just close this PR.


amoss:
Looks correct - check with @joelhans that we haven't broken anything in the docs.

underhood:
I will update his PR with documentation as soon as this gets merged.

PR #8505, opened 2020-03-26
Extend TLS Support

### LINKED ISSUE #8497: TLS v1.3 support
Additionally apache keeps good logs about which IP tries to access what resources and allows for http authentication
Also netdata only supports up to TLS 1.2 (which can no longer be considered secure)

_Originally posted by @Pingger in https://github.com/netdata/netdata/issues/7838#issuecomment-577901929_

thiagoftsm:
##### Summary
Fixes #8497 

This PR changes configurations on master Netdata to support newest TLS versions.
##### Component Name
Web
Stream
##### Test Plan
The test plan is majority focus in the stream communication, because we are assuming that the tester will use the browser to access the `master` to see information related to his slaves.

The following tests were done with this PR:

###### First test 
This test is the basis for the next tests considering that we will only change few values after this.

1 - Create key and certificate using openssl. It is possible to use self signed certificate.
2 - Set at least following variables inside `netdata.conf` on your master:
```
[web]
```
3 - Start your master
4 - Set the following variables inside your `stream.conf` on your slave:
```
[stream]
    destination = 192.168.0.12:19999:SSL
```
4 - Start your slave

With the previous options Netdata will start streaming using TLS 1.3, the next tests

##### Second Test
Now we only need to change configuration on `master` side, please, remember to stop your `slave` before you stop your master and only start your `slave` after to start your `master`.

The second test will rename options and bring a new option to set tls version: 

```
[web]
	tls key = /etc/netdata/ssl/key.pem
	tls certificate = /etc/netdata/ssl/cert.pem
        tls version = 1.3
```
1 - Stop Slave
2 - Stop master
3 - Start master
4 - Start slave

##### Third test

This test will change the `tls version` on master side only, on slave this is set during the handshake process.

```
[web]
   tls version = 1.2
```

1 - Stop Slave
2 - Stop master
3 - Start master
4 - Start slave

##### Fourth test
On this final test we will set the `tls version` and `tls ciphers`:

```
[web]
   tls version = 1.3
   tls ciphers = TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256
```

1 - Stop Slave
2 - Stop master
3 - Start master
4 - Start slave


libnetdata/socket/security.c
```diff
@@ -7,6 +7,14 @@ SSL_CTX *netdata_client_ctx=NULL;
 SSL_CTX *netdata_srv_ctx=NULL;
 const char *security_key=NULL;
 const char *security_cert=NULL;
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+const char *tls_version=NULL;
+const char *tls_ciphers=NULL;
+const char *ciphers[] = {
+        "ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES128-SHA",
+        "ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!MD5:!PSK:!aECDH:!EDH-DSS-DES-CBC3-SHA:!EDH-RSA-DES-CBC3-SHA:!KRB5-DES-CBC3-SHA",
+        "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256" };
```
Ferroin: Can you provide some info on how this list was decided on? I'm not saying it's a bad choice (though I have not yet audited it), but we _really_ should be transparent about why we're choosing defaults when it comes to potential security issues

ilyam8: it was exactly the reason i assigned you @Ferroin for exactly this question 😄

Ferroin: I think we may need to compute this dynamically actually based on whether or not the user has selected a specific protocol. If they have, we should restrict to only ciphers supported by that protocol (this is kind of important, since about half the known secure ciphers are TLS 1.3 only).

thiagoftsm: This was decided using the following command:

```
/usr/bin/openssl ciphers -s -v
```

I got the output of the command and I separated according the TLS version.

ilyam8: i feel it is not reliable way 🤔

Ferroin: @ilyam8 is correct here, that output changes based on the exact OpenSSL version, and you end up with all ciphers enabled, which is bad from a security perspective. I'll look into putting together a list either today or tomorrow based on protocol to use for defaults.

Also, I didn't see anywhere in the code,b ut are we telling OpenSSL to prefer the server's cipher list here? We ideally should be, as most clients use insecure defaults for compatibility reasons.

thiagoftsm: I will bring a new commit using https://www.openssl.org/docs/man1.1.1/man3/SSL_get_ciphers.html  in few minutes.

underhood:
Regarding the effect on ACLK (@amoss you were asking):
1. by reading code it should not have any
2. I tested just now agent still connects to test cloud environment

libnetdata/socket/security.c
thiagoftsm: The last commit removes cipher lists, now we are working with the OpenSSL default unless the user specifies ciphers with new options we are bringing.

Ferroin: OK, with that I'm going to mark this resolved, as that's consistent with 'normal' behavior of most apps that use OpenSSL.

Ferroin:
@thiagoftsm  I'm going to make a point to test this tomorrow just to be sure it works as expected, but it looks good to me.

libnetdata/socket/security.c
```diff
@@ -49,32 +53,59 @@ void security_openssl_library()
...
+    if (!strcmp(lversion, "1.1"))
+        return TLS1_1_VERSION;
+    else if (!strcmp(lversion, "1.2"))
+        return TLS1_2_VERSION;
+#if OPENSSL_VERSION_NUMBER >= 0x1010101bL
+    else if (!strcmp(lversion, "1.3"))
+        return TLS1_3_VERSION;
+#endif
+
+    return TLS_MAX_VERSION;
```
ilyam8: Is `TLSv1` a thing. I see i took it into account adding SSL protocl parsing in weblog

https://github.com/netdata/go.d.plugin/blob/1e6132f9572373005ab13ecdba9ef56a6bb6ca8b/modules/weblog/logline.go#L566-L575

i dont remember details tho 😄

thiagoftsm: Considering that you added, I will also bring. For sure it is not the safest option, but people can use it yet.

ilyam8: I have a second thought, it makes sense for weblog because it has to parse logs, but i think it doesnt make sense for Netdata to have it as an option.

Sorry for the noise 😅

thiagoftsm: he last commit brought the missing TLS versions, but SSL is too dangerous, so I will keep it out. I updated the documentation to give better understanding about what means the new option `tls version`.

thiagoftsm: > Sorry for the noise sweat_smile

Do not worry, I brought all TLS  options, because I think it is good to keep compatibility with old versions, but we also need to consider the security with old versions, so I think we do not need to extend so much to the past!

Thank you for your contribution! :)

libnetdata/socket/security.h
```diff
@@ -33,6 +33,10 @@ extern SSL_CTX *netdata_client_ctx;
 extern SSL_CTX *netdata_srv_ctx;
 extern const char *security_key;
 extern const char *security_cert;
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
```
ilyam8: It seems `0x10100000L` is a magic number, i suggest to fix it

web/server/README.md
```diff
@@ -96,6 +96,24 @@ openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -keyout key.pem -out
 openssl speed rsa2048 rsa4096
 ```
 
+### Select TLS version
+
+Since version 1.21 Netdata accepts `tls key` and `tls certificate` as replacement of the options `ssl key` and `ssl certificate`, 
+ but you do not need to change your `netdata.conf` case you are using old `ssl` variables, because old options are kept due compatibility.
```
joelhans:
```suggestion
Beginning with version 1.21, Netdata accepts `tls key` and `tls certificate` as replacement of the settings `ssl key` and `ssl certificate`. 
You do not need to change your `netdata.conf` if you are using the previous `ssl` settings, because these are kept for backwards compatibility.
```

web/server/README.md
```diff
@@ -96,6 +96,24 @@ openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -keyout key.pem -out
...
+
+Netdata also brings two new options related to TLS, now you can specify the TLS version and the ciphers that you want to use:
```
joelhans:
```suggestion
You can also specify the TLS version and the ciphers that you want to use:
```

web/server/README.md
```diff
@@ -96,6 +96,24 @@ openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -keyout key.pem -out
...
+
+```conf
+[web]
+    tls version = 1.3
+    tls ciphers = TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256
+```
+
+case you do not specify these options, Netdata will use the highest options available on your system.
```
joelhans:
```suggestion
If you do not specify these options, Netdata will use the highest options available on your system.
```

web/server/README.md
```diff
@@ -96,6 +96,24 @@ openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -keyout key.pem -out
 openssl speed rsa2048 rsa4096
 ```
 
+### Select TLS version
+
+Since version 1.21 Netdata accepts `tls key` and `tls certificate` as replacement of the options `ssl key` and `ssl certificate`,
```
joelhans:
```suggestion
Beginning with version 1.21, Netdata accepts `tls key` and `tls certificate` as replacement of the options `ssl key` and `ssl certificate`, 
```

web/server/README.md
```diff
@@ -96,6 +96,24 @@ openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -keyout key.pem -out
...
+```conf
+[web]
+    tls version = 1.3
+    tls ciphers = TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256
+```
+
+case you do not specify these options, Netdata will use the highest options available on your system.
+
+An important note about the `tls version`, Netdata accepts all the TLS version as arguments (`1`, `1.1`, `1.2` and `1.3`), but
+ only the last option is considered safe.
```
joelhans:
```suggestion
> While Netdata accepts all TLS versions as arguments (`1`, `1.1`, `1.2` and `1.3`), we recommend you use `1.3` for the most secure encryption.
```

libnetdata/socket/security.h
thiagoftsm: This is the way OpenSSL defines their version, this is define inside `/usr/include/openssl/opensslv.h`. What kind of fix are you talking about? Do we need to create a name definition for this version?

ilyam8: yes, i meant const or something

ilyam8: i see you are using it in 8 places

web/server/README.md
thiagoftsm: I used all suggestions that you gave me, thank you!

web/server/README.md
thiagoftsm: I will mark this and the next as resolved, for the motive I already given.

libnetdata/socket/security.h
thiagoftsm: I changed for a text.

web/server/README.md
```diff
@@ -98,21 +98,21 @@ openssl speed rsa2048 rsa4096
...
+You can also specify the TLS version and the ciphers that you want to use:
 
 ```conf
 [web]
     tls version = 1.3
     tls ciphers = TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256
 ```
 
-case you do not specify these options, Netdata will use the highest options available on your system.
+If you do not specify these options, Netdata will use the highest options available on your system.
```
Ferroin: For at least the ciphers, this isn't true. Netdata will use the default cipher list based on the selected protocol, which may not be the most secure possible.

web/server/README.md
```diff
@@ -98,21 +98,21 @@ openssl speed rsa2048 rsa4096
...
 
-An important note about the `tls version`, Netdata accepts all the TLS version as arguments (`1`, `1.1`, `1.2` and `1.3`), but
- only the last option is considered safe.
+While Netdata accepts all the TLS version as arguments (`1`, `1.1`, `1.2` and `1.3`), we recommend you use `1.3` for the most secure encryption.
```
Ferroin: Two things WRT this that may need to be considered in the code:

* We should probably accept 1.0 and alias it to 1 for the protocol version, as that's what most other software does.
* It would be a very good idea to support a list of protocol versions instead of enforcing a specific one. Only allowing a specific protocol is potentially problematic from the perspective of client-side support (that is, it may not be possible to support all the clients a user wants to support if they are only able to select a single protocol).

web/server/README.md
thiagoftsm: Case we use the highest protocol version, 1.3 right now, we will discard a lot of ciphers that are not considered safe.

web/server/README.md
thiagoftsm: For OpenSSL we have only SSLv3 and TLSx.y and considering that the option is named `tls version` I think this is not a problem, case OpenSSL uses another protocol, we will have to deal with this. 
About the alias, I think it is a good point, I will bring it in few minutes. Thank you!

Ferroin: There are enough differences though between TLS 1.2 and 1.3 that clients that only support up to 1.2 can't connect to a server that only supports 1.3. That's kind of the reason that there's a version distinction there.

And, unfortunately, there are still a lot of systems out there with clients that only support 1.2, or even just 1.1, because anything newer required a new version of OpenSSL with some rather problematic ABI compatibility issues.

web/server/README.md
Ferroin: That doesn't mean we automatically use the most secure ciphers though, which is what the statement above implies.. Take a look at the standard list for 1.2 for example, almost half of them are considered to at best be weak and at worst completely broken.

I'd suggest the following alternative:
```suggestion
If you do not specify these options, Netdata will use the highest available protocol version on your system and the default cipher list for that protocol provided by your TLS implementation.
```

web/server/README.md
thiagoftsm: I understand this, among the tests that I did, I forced the `slave` to use `1.2` and the server with `1.3` and as expected it could not work. 

I did not add any additional information for the issue about the relationship between tls versions, because openssl always try to get the highest value between client and server, on the other hand, I was forcing to use an unique version of the protocol. I will let this flexible between the minimum and maximum available for our users do not have problems.

web/server/README.md
thiagoftsm: I accepted the suggestion, I am finishing the tests here, and I will commit everything in few minutes. Thank you!

thiagoftsm: Done!

daemon/main.c
```diff
@@ -400,11 +400,26 @@ static const char *verify_required_directory(const char *dir) {
 #ifdef ENABLE_HTTPS
 static void security_init(){
     char filename[FILENAME_MAX + 1];
-    snprintfz(filename, FILENAME_MAX, "%s/ssl/key.pem",netdata_configured_user_config_dir);
-    security_key    = config_get(CONFIG_SECTION_WEB, "ssl key",  filename);
+    char common[8];
```
mfundul: We do we need this array? Can't we use the string literal `"none"`?

libnetdata/socket/security.h
mfundul: But I see here you're still using `0x10100000L` instead of `OPENSSL_VERSION_WITH_TLS_13`

daemon/main.c
```diff
@@ -400,11 +400,26 @@ static const char *verify_required_directory(const char *dir) {
...
+
+    security_cert    = config_get(CONFIG_SECTION_WEB, "ssl certificate",  common);
+    if (!strcmp(security_cert, common)) {
+        snprintfz(filename, FILENAME_MAX, "%s/ssl/cert.pem",netdata_configured_user_config_dir);
+        security_cert    = config_get(CONFIG_SECTION_WEB, "tls certificate",  filename);
+    }
 
-    snprintfz(filename, FILENAME_MAX, "%s/ssl/cert.pem",netdata_configured_user_config_dir);
-    security_cert    = config_get(CONFIG_SECTION_WEB, "ssl certificate",  filename);
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
```
mfundul: replace `0x10100000L` with `OPENSSL_VERSION_WITH_TLS_13`

daemon/main.c
thiagoftsm: I used this to be sure that compiles would not allocate memory, instead they would use the `stack`. I confess I am few years without to do reverse engineer, so I am not sure that this "problem" was already solved, but I will remove it as you suggested. I agree that for a small text this is not so necessary.

thiagoftsm: The last commit removed the array.

daemon/main.c
thiagoftsm: Thank you, I missed this number!

daemon/main.c
```diff
@@ -400,11 +400,24 @@ static const char *verify_required_directory(const char *dir) {
 #ifdef ENABLE_HTTPS
 static void security_init(){
     char filename[FILENAME_MAX + 1];
-    snprintfz(filename, FILENAME_MAX, "%s/ssl/key.pem",netdata_configured_user_config_dir);
-    security_key    = config_get(CONFIG_SECTION_WEB, "ssl key",  filename);
+    security_key    = config_get(CONFIG_SECTION_WEB, "ssl key",  "none");
+    if (!strcmp(security_key, "none")) {
```
mfundul: Why does `"ssl key"` take priority over `"tls key"`?

daemon/main.c
```diff
@@ -400,11 +400,24 @@ static const char *verify_required_directory(const char *dir) {
...
+        snprintfz(filename, FILENAME_MAX, "%s/ssl/key.pem",netdata_configured_user_config_dir);
+        security_key    = config_get(CONFIG_SECTION_WEB, "tls key",  filename);
+    }
+
+    security_cert    = config_get(CONFIG_SECTION_WEB, "ssl certificate",  "none");
+    if (!strcmp(security_cert, "none")) {
```
mfundul: Why does `"ssl certificate"` take priority over `"tls certificate"`?

streaming/rrdpush.c
```diff
@@ -109,6 +109,9 @@ int rrdpush_init() {
     }
 
     char *invalid_certificate = appconfig_get(&stream_config, CONFIG_SECTION_STREAM, "ssl skip certificate verification", "no");
+    if (!strcmp(invalid_certificate, "no"))
```
mfundul: Why does `"ssl skip certificate verification"` take priority over `"tls skip certificate verification"`?

daemon/main.c
thiagoftsm: `ssl key` was the start, but software like `MongoDB` are beginning to use `TLS` prefix instead `SSL` prefix, so I arrived in the conclusion that we need to keep the `SSL` prefix to keep compatibility, but it is also a good idea to bring the possibility to use `TLS` prefix, because other projects are doing this.

daemon/main.c
thiagoftsm: The same explanation given before. This is during the initialization, so we will test only one time.

streaming/rrdpush.c
thiagoftsm: The same argument used for the two previous. This is also used only during the initialization.

thiagoftsm:
Thank you everybody!

PR #8509, opened 2020-03-27
Remove stackscale demo link and clean up page

joelhans:
##### Summary

In investigating our demo sites @prologic found one that's not readily accessible, so this PR removes it from the documentation.


prologic:
🙏

PR #8521, opened 2020-03-28
github/workflow: add ACLK to the labeler config

ilyam8:
##### Summary

I noticed there are no labels in the https://github.com/netdata/netdata/pull/8499

This PR adds `aclk/` to the labeler config. 

##### Component Name

`.github/workflow`

##### Test Plan

No tests are needed

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -10,6 +10,10 @@
 #
 # Please keep the labels sorted and deduplicated.
 
+ACLK:
+  - aclk/*
+  - aclk/**/*
+
 area/backends:
   - backends/*
   - backends/**/*


ilyam8:
@prologic i think we dont need [document-start](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.document_start) rule to be on, are you ok with it?

prologic:
> @prologic i think we dont need [document-start](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.document_start) rule to be on, are you ok with it?
> 
> edit: i disabled the rule in #8522

I'm okay with this but I think you should edit the top-level config instead?

Issue #7701, opened 2020-01-09
JSON parser/generator as hard dependency (probably json-c)


underhood:
#### Feature idea summary
JSON being one of the most used I/O formats for the Netdata agent we should not reinvent the wheel and use `json-c` *(or other similar library)* to both parse/generate JSON to/from same data structures.

`json-c` *(https://github.com/json-c/json-c)* was chosen as it is already used in health as soft dependency but it is up to the discussion and better library might be chosen *(if it fits all the requirements, at least tries to adhere to JSON standards and has better performance)*.

Last necessity to parse JSON arrised in #7700 and it was discussed it makes no sense to handcraft YAJPG&trade; (Yet Another Json Parser and Generator).

Start using this parser for all new code that needs to handle JSON.

In case we will decide bringing in `json-c` as hard dependency is an issue we might look into something like cJSON *(https://github.com/DaveGamble/cJSON)* that we can bring into the codebase directly.

##### Expected behaviour
*Do not write own JSON parsing code in multiple places
*Do not hand generate JSON by `buffer_sprintf` in multiple places
*we should be able to parse JSON into same data structures  that are used to generate JSON string

amoss:
Some initial points:

* cJSON is not thread safe.
* The data-model for cJSON is one structure per JSON which implies heavy use of memory allocation / deallocation every time we parse/serialize JSON data.

The `buffer_sprintf` code that outputs JSON is heavily used and we need to consider the performance of anything that we replace it with. One advantage of the current code is that it does not touch the memory allocator during output of JSON

amoss:
This has been done via a different issue.

PR #8534, opened 2020-03-30
Fixed unicode decode in internal JSON parser

### LINKED ISSUE #8480: Add support to the internal JSON library to handle \uXXXX character decoding
#### Summary

Review the internal JSON library and provide support for \uXXXX decoding as this is now
not supported.

Example: 

Incoming payload from the ACLK 
```
{"type":"http","version":1,"callback-topic":"/svc/agent-data-ctrl/d1d3072c-b21c-43f1-baf2-aedb23373fda","msg-id":"b8a3e6fd-c084-4643-9409-6bb64525baa4","payload":"/api/v1/data?chart=system.cpu\u0026_=1585139042316\u0026format=array\u0026points=480\u0026group=average\u0026gtime=0\u0026options=absolute%7Cjsonwrap%7Cnonzero\u0026after=-480"}
```

should be JSON parsed to:

```
{"type":"http","version":1,"callback-topic":"/svc/agent-data-ctrl/d1d3072c-b21c-43f1-baf2-aedb23373fda","msg-id":"b8a3e6fd-c084-4643-9409-6bb64525baa4","payload":"/api/v1/data?chart=system.cpu&_=1585139042316&format=array&points=480&group=average&gtime=0&options=absolute%7Cjsonwrap%7Cnonzero&after=-480"}
```

Handle the `\u0026` to `&` map

stelfrag:
Fixes #8480 
##### Summary
The internal JSON parser identified \uxxxx type symbols but no decoding was done. This affected the way incoming ACLK queries were handled as they contain `\u0026` which should be decoded to `&` (to form valid URLs)
##### Component Name
ACLK

###### To replicate the problem before PR is applied
- Make sure `libjson-c-dev` dependency is not installed in your system
- Compile an ACLK enabled and claimed agent with `-DNETDATA_INTERNAL_CHECKS`
- Activate the `D_ACLK` flag (`debug flags = 0x0000000200000000`) in `netdata.conf`
- Start the agent and go to view charts in the dev cloud or local environment
- Monitor the `debug.log` for incoming cloud requests
- Notice an incoming payload of the form
  `"payload":"/api/v1/data?chart=system.load\u0026_=1585549971725 ... ` followed by an added query in the form
  `/api/v1/data?chart=system.load\u0026_=1585549971725 ...`
  This will fail as the webapi will not handle the `\u0026` properly.

####### To test the fix
- Compile the PR with the proposed fix
- Restart the agent (ACLK enabled and claimed as above)  and go to view charts in the dev cloud or local environment
- Notice an incoming payload of the form
  `"payload":"/api/v1/data?chart=system.load\u0026_=1585549971725 ... ` followed by an added query in the form
  `/api/v1/data?chart=system.load&_=1585549971725 ...`
  which is now correctly handled by the webapi

### DIFF
--- a/libnetdata/json/jsmn.c
+++ b/libnetdata/json/jsmn.c
@@ -3,6 +3,29 @@
 #include "../libnetdata.h"
 #include "jsmn.h"
 
+static inline int output_characters(char *wstr, unsigned long unicode_char)
+{
+    if (unicode_char < 0x80) {
+        *wstr = (char)((unicode_char >> 0 & 0x7F) | 0x00);
+        return 1;
+    } else if (unicode_char < 0x0800) {
+        *wstr = (char)((unicode_char >> 6 & 0x1F) | 0xc0);
+        *(wstr + 1) = (char)((unicode_char >> 0 & 0x3F) | 0x80);
+        return 2;
+    } else if (unicode_char < 0x010000) {
+        *wstr = (char)((unicode_char >> 12 & 0x0F) | 0xE0);
+        *(wstr + 1) = (char)((unicode_char >> 6 & 0x3F) | 0x80);
+        *(wstr + 2) = (char)((unicode_char >> 0 & 0x3F) | 0x80);
+        return 3;
+    } else {
+        *wstr = (char)(unicode_char >> 18 & 0x07 | 0xF0);
+        *(wstr + 1) = (char)((unicode_char >> 12 & 0x3F) | 0x80);
+        *(wstr + 2) = (char)((unicode_char >> 6 & 0x3F) | 0x80);
+        *(wstr + 3) = (char)((unicode_char >> 0 & 0x3F) | 0x80);
+        return 4;
+    }
+}
+
 /**
  * Alloc token
  *
@@ -115,7 +138,7 @@ static jsmnerr_t jsmn_parse_primitive(jsmn_parser *parser, const char *js,
  *
  * @return It returns 0 on success and another integer otherwise
  */
-static jsmnerr_t jsmn_parse_string(jsmn_parser *parser, const char *js,
+static jsmnerr_t jsmn_parse_string(jsmn_parser *parser, char *js,
                                    size_t len, jsmntok_t *tokens, size_t num_tokens) {
     jsmntok_t *token;
 
@@ -146,11 +169,29 @@ static jsmnerr_t jsmn_parse_string(jsmn_parser *parser, const char *js,
 
         /* Backslash: Quoted symbol expected */
         if (c == '\\') {
+            unsigned long unicode_char = 0;
             parser->pos++;
             switch (js[parser->pos]) {
                 /* Allowed escaped symbols */
-                case '\"': case '/' : case '\\' : case 'b' :
-                case 'f' : case 'r' : case 'n'  : case 't' :
+                case '\"':
+                case '/':
+                case '\\':
+                    js[parser->pos-1] = js[parser->pos];
+                    break;
+                case 'b':
+                    js[parser->pos-1] = '\b';
+                    break;
+                case 'f':
+                    js[parser->pos-1]  = '\f';
+                    break;
+                case 'r':
+                    js[parser->pos-1]  = '\r';
+                    break;
+                case 'n':
+                    js[parser->pos-1]  = '\n';
+                    break;
+                case 't':
+                    js[parser->pos-1]  = '\t';
                     break;
                     /* Allows escaped symbol \uXXXX */
                 case 'u':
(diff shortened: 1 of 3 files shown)


amoss:
One of the points above is wrong - I though I took it out of the review comment but the box is small and awkward: the token allocation doesn't maintain separate strings so ignore the third bullet.

stelfrag:
> I can see quite a few problems.
> 
> * If the string contains a short escape (e.g. `\t`) then it does not generate an error but it is not decoded.
> * If the escape is a two-byte character (e.g. `\u062a`) then only the low-order byte is written out corrupting the string.
>   * It is unclear what the effect of overwriting the input string will be on code that calls the parser, but it seems unnecessary when the code allocates a separate token to hold the substring.
> * The code seems to be written for potentially non-null-terminated inputs as it uses a `len` parameter but shrinking the string will change the length without notifying the caller.
> 
> Given the problems noted and the uncertainty in where this is called from it should not change the input the string, please rewrite it to decode into the token.

I will check how to avoid the string mutation. The part of the code that I added will move closer to the callback function, decode the string in a freshly allocated memory area, do the callback and claenup. 

This is a parser that does not do decoding whatsoever so it detects \uxxxx, \t, ]n \b etc and just passes them through.

amoss:
The string that is produced by the parser for the callback needs to be decoded - i.e. a `\t` needs to be mapped into a single byte for exactly the same reasons as decoding the `\uxxxx` substrings.

The interface around the parsing calls in json.c look well encapsulated so the memory management should fit in naturally (i.e. the strings only need to be available during the callbacks, not after the parsing completes).

underhood:
Seems to conform to spec with one thing I am not sure of:
```
2.5.  Strings

   The representation of strings is similar to conventions used in the C
   family of programming languages.  A string begins and ends with
   quotation marks.  All Unicode characters may be placed within the
   quotation marks except for the characters that must be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).

   Any character may be escaped.  If the character is in the Basic
   Multilingual Plane (U+0000 through U+FFFF), then it may be
   represented as a six-character sequence: a reverse solidus, followed
   by the lowercase letter u, followed by four hexadecimal digits that
   encode the character's code point.  The hexadecimal letters A though
   F can be upper or lowercase.  So, for example, a string containing
   only a single reverse solidus character may be represented as
   "\u005C".

   Alternatively, there are two-character sequence escape
   representations of some popular characters.  So, for example, a
   string containing only a single reverse solidus character may be
   represented more compactly as "\\".

   To escape an extended character that is not in the Basic Multilingual
   Plane, the character is represented as a twelve-character sequence,
   encoding the UTF-16 surrogate pair.  So, f2.5.  Strings

   The representation of strings is similar to conventions used in the C
   family of programming languages.  A string begins and ends with
   quotation marks.  All Unicode characters may be placed within the
   quotation marks except for the characters that must be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).

   Any character may be escaped.  If the character is in the Basic
   Multilingual Plane (U+0000 through U+FFFF), then it may be
   represented as a six-character sequence: a reverse solidus, followed
   by the lowercase letter u, followed by four hexadecimal digits that
   encode the character's code point.  The hexadecimal letters A though
   F can be upper or lowercase.  So, for example, a string containing
   only a single reverse solidus character may be represented as
   "\u005C".

   Alternatively, there are two-character sequence escape
   representations of some popular characters.  So, for example, a
   string containing only a single reverse solidus character may be
   represented more compactly as "\\".

   To escape an extended character that is not in the Basic Multilingual
   Plane, the character is represented as a twelve-character sequence,
   encoding the UTF-16 surrogate pair.  So, for example, a string
   containing only the G clef character (U+1D11E) may be represented as
   "\uD834\uDD1E".
```

the 12 char sequences ` So, for example, a string
   containing only the G clef character (U+1D11E) may be represented as
   "\uD834\uDD1E".`

cosmix:
@amoss can you please review this?

underhood:
@stelfrag  I think it might be useful to rebase on the master to pickup all the newest changes to installer, configure, ...

amoss:
I don't think that we need to support surrogate pairs. My reading of the json spec is that we must support utf-8 decoding, while utf-16 and utf-32 are optional. The 12-character example quoted above is a double-decoding step to remove the utf-16 surrogates. If we simply don't support utf-16 and output utf-8 then I think we can ignore them. @cosmix what do you think?

stelfrag:
> @stelfrag I think it might be useful to rebase on the master to pickup all the newest changes to installer, configure, ...

Rebased / pushed

underhood:
Well, there were issues with UTF escaping detected in internal parser, it comes down to -> do we want to fix the case that we have with cloud or do we want to fix UTF handling to be conforming to the specification.

The first option might be acceptable with issue #7701 in mind _(although I am not sure at this point if we decided we want to do that)_

PR #8544, opened 2020-03-30
Update pfsense.md

electropup42:
Note about Python package when installing on pfSense 2.4.5

##### Summary
Added note about installation on pfSense 2.4.5 no longer requiring python 3.6 package since 3.7 is installed already.

##### Component Name
pfSense install guide

##### Test Plan
Test on another pfSense install

### DIFF
--- a/packaging/installer/methods/pfsense.md
+++ b/packaging/installer/methods/pfsense.md
@@ -26,7 +26,8 @@ pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.17.1.txz
 
 **Note:** If you receive a `Not Found` error during the last two commands above, you will either need to manually look
 in the [repo folder](http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/) for the latest available package and use its
-URL instead, or you can try manually changing the netdata version in the URL to the latest version.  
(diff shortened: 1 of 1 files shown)


packaging/installer/methods/pfsense.md
```diff
@@ -26,7 +26,8 @@ pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.17.1.txz
 
 **Note:** If you receive a `Not Found` error during the last two commands above, you will either need to manually look
 in the [repo folder](http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/) for the latest available package and use its
-URL instead, or you can try manually changing the netdata version in the URL to the latest version.  
+URL instead, or you can try manually changing the netdata version in the URL to the latest version.
+**Note 2:** On pfSense 2.4.5, Python version 3.7 may be installed by the system, in which case you should drop the python package from the install list.
```
joelhans: A slight rephrasing, and I've also added a new line between the two notes so that they render as separate paragraphs.

I also added some people from the packaging team to double-check these instructions as well.

```suggestion
URL instead, or you can try manually changing the Netdata version in the URL to the latest version.

**Note:** On pfSense 2.4.5, Python version 3.7 may be installed by the system, in which case you should should not install Python from the FreeBSD repository as instructed above.
```

PR #8545, opened 2020-03-30
Switching over to soft feature flag

### LINKED ISSUE #8378: Remove the feature flag for ACLK.

amoss:
##### Summary
fixes #8378 

##### Component Name
cloud

##### Test Plan

* Default behavior is to allow the cloud (the --cloud-testing flag is gone).
* The --disable-cloud switch in the installer will disable everything.
* The default config option is `[global] netdata cloud = coming soon` -> this currently disables cloud functionality.
* The feature is enabled with `[global] netdata cloud = enabled` -> this still requires a local copy of the infrastructure or auth for the testing platform.
* If the feature is enabled at compile-time then it can be switched off with `[global] netdata cloud = disabled`.

In cases where the cloud is enabled, check:
* The flag in the `/api/v1/info` endpoint.
* That the ACLK is successfully established.

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -575,6 +575,21 @@ static void get_netdata_configured_variables() {
     get_system_HZ();
     get_system_cpus();
     get_system_pid_max();
+
+    // --------------------------------------------------------------------
+    // Check if the cloud is enabled
+#ifdef DISABLE_CLOUD
+    netdata_cloud_setting = 0;
+#else
+    char *cloud = config_get(CONFIG_SECTION_GLOBAL, "netdata cloud", "coming soon");
+    if (!strcmp(cloud, "coming soon")) {
+        netdata_cloud_setting = 0;          // Note: this flips to 1 after the release
+    } else if (!strcmp(cloud, "enable")) {
+        netdata_cloud_setting = 1;
+    } else if (!strcmp(cloud, "disable")) {
+        netdata_cloud_setting = 0;
+    }
+#endif
 }
 
 static void get_system_timezone(void) {
--- a/Makefile.am
+++ b/Makefile.am
@@ -469,14 +469,18 @@ BACKENDS_PLUGIN_FILES = \
     backends/prometheus/backend_prometheus.h \
     $(NULL)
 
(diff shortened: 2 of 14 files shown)


daemon/main.c
```diff
@@ -575,6 +575,21 @@ static void get_netdata_configured_variables() {
     get_system_HZ();
     get_system_cpus();
     get_system_pid_max();
+
+    // --------------------------------------------------------------------
+    // Check if the cloud is enabled
+#ifdef DISABLE_CLOUD
+    netdata_cloud_setting = 0;
+#else
+    char *cloud = config_get(CONFIG_SECTION_GLOBAL, "netdata cloud", "coming soon");
```
underhood: why not just have `enable`/`disable` and swap the default at the release?

amoss: Because the old default will live in config files after the update.

stelfrag:
Testing:

#### Building the agent without any flag
  - ACLK related code is compiled (source is pulled)
  - `netdata cloud = coming soon` is enabled in the config
     - ACLK remains disabled
     ```
     "cloud-enabled": false,
     "cloud-available": true,
     "agent-claimed": true,
     "aclk-available": false
    ```
  - Setting `netdata cloud = enable` in the config and restarting agent
    - ACLK is established
    ```
    "cloud-enabled": true,
    "cloud-available": true,
    "agent-claimed": true,
    "aclk-available": true
       ```
  - Setting `netdata cloud = disable` and restarting agent
    - ACLK is disabled
      ``` 
      "cloud-enabled": false,
      "cloud-available": true,
      "agent-claimed": true,
      "aclk-available": false
      ```

#### Building the agent with --disabled-cloud
Nothing is compiled, no source pull, the flags reported in that case 
```
	"cloud-enabled": false
	"cloud-available": false
	"agent-claimed": true  <- (because agent is claimed)
	"aclk-available": false
```

Note that the `netdata cloud` configuration in the `[global]` section accepts **enable** or **disable**
(in the test scenario it is enabled, disabled)

PR #8551, opened 2020-03-31
github/workflow: remove duplicate key (`line-length`) from the yamlli…

ilyam8:
##### Summary

Was doing the change in #8542 via github ui (it is wrong i know!), didnt notice the option is already in there.

##### Component Name

`ci`

##### Test Plan

No tests needed


ilyam8:
We run yamllint on diff, there is filtered yamllint errors (master)

https://github.com/netdata/netdata/runs/548475990?check_suite_focus=true

Issue #8558, opened 2020-03-31
tmp fills up with netdata-libwebsockets-XXXXX


lucasRolff:
##### Bug report summary

Over time it seems that on certain environments /tmp ends up taking a lot of space due to folders from netdata-libwebsockets-XXXX being left behind, every nightly run it grows bigger.

Running the updater manually ( `/usr/libexec/netdata/netdata-updater.sh`), I see a few issues:

```
 --- Prepare libwebsockets ---
v3.2.2.tar.gz: OK
[/root/netdata-updater-DyiENE/netdata-v1.20.0-278-gb8cef882]# tar -xf /tmp/netdata-libwebsockets-ryvUKl/v3.2.2.tar.gz -C /tmp/netdata-libwebsockets-ryvUKl
 OK

[/tmp/netdata-libwebsockets-ryvUKl/libwebsockets-3.2.2]# env CFLAGS= CXXFLAGS= LDFLAGS= cmake -D LWS_WITH_SOCKS5:bool=ON .
env: cmake: No such file or directory
 FAILED

[/tmp/netdata-libwebsockets-ryvUKl/libwebsockets-3.2.2]# env CFLAGS= CXXFLAGS= LDFLAGS= make
make: *** No targets specified and no makefile found.  Stop.
 FAILED

[/root/netdata-updater-DyiENE/netdata-v1.20.0-278-gb8cef882]# mkdir -p /root/netdata-updater-DyiENE/netdata-v1.20.0-278-gb8cef882/externaldeps/libwebsockets
 OK

[/root/netdata-updater-DyiENE/netdata-v1.20.0-278-gb8cef882]# cp /tmp/netdata-libwebsockets-ryvUKl/libwebsockets-3.2.2/lib/libwebsockets.a /root/netdata-updater-DyiENE/netdata-v1.20.0-278-gb8cef882/externaldeps/libwebsockets/libwebsockets.a
cp: cannot stat '/tmp/netdata-libwebsockets-ryvUKl/libwebsockets-3.2.2/lib/libwebsockets.a': No such file or directory
 FAILED

 FAILED  Failed to build libwebsockets. The install process will continue, but you may not be able to connect this node to Netdata Cloud.
```

As a result the 31MB of temporary files are left behind:

```
# du -sh /tmp/netdata-libwebsockets-ryvUKl
31M	/tmp/netdata-libwebsockets-ryvUKl
```

[...]

PR #8559, opened 2020-03-31
Added logic to bail early on LWS build if cmake is not present.

Ferroin:
##### Summary

This adds some basic logic to the installer to check if `cmake` is present or not and just skip attempting to build lbwebsockets.

##### Component Name

area/packaging

##### Test Plan

Verified locally by checking in Docker containers that it behaves correctly.

##### Additional Information

Fixes: #8558

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -536,6 +536,12 @@ bundle_libwebsockets() {
     return 0
   fi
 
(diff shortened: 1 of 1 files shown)


ilyam8:
The changes looks pretty trivial, approving w/o testing, reading the code is enough imo.

PR #8565, opened 2020-04-01
Set CFLAGS appropriately so all warnings are treated as errors during CI buidls

prologic:
##### Summary

SSIA

##### Component Name

- area/ci

##### Test Plan

- CI

##### Additional Information

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -42,7 +42,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CFLAGS: >
-        -O1 -ggdb -Wall -Wextra -Wformat-signedness
+        -O1 -ggdb -Wall -Werror -Wextra -Wformat-signedness
         -fstack-protector-all
         -DNETDATA_INTERNAL_CHECKS=1
         -D_FORTIFY_SOURCE=2
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -17,6 +17,7 @@ jobs:
           LOCAL_ONLY: "true"
         run: |
           ./tests/installer/checksums.sh
+
   dashboard-checks:
     name: Dashboard
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


vlvkobal:
I've tried to test it with #8567. It should trigger the following warning
```
claim/claim.c:7:14: warning: 'test_ci_checks' defined but not used [-Wunused-variable]
    7 | static char *test_ci_checks = "test";
      |              ^~~~~~~~~~~~~~
```
But the output of `run_install_with_dist_file.sh` is [clean](https://github.com/netdata/netdata/pull/8567/checks?check_run_id=551517225).

vlvkobal:
To treat warnings as errors we should also add `-Werror`, but if we do that, the `./configure` step can't be completed:
```
checking for pthread-config... no
configure: error: Cannot initialize pthread environment
 FAILED
```

prologic:
> To treat warnings as errors we should also add `-Werror`, but if we do that, the `./configure` step can't be completed:

Ahh I will add that now.

> ```
> checking for pthread-config... no
> configure: error: Cannot initialize pthread environment
>  FAILED
> ```

How do we address this?

mfundul:
We can't treat warnings as errors, there is at least one case where the warnings cannot be avoided:
https://github.com/netdata/netdata/issues/2504

vlvkobal:
I'm not sure it is possible to address the `configure` error. `config.log`:
```
| /* end confdefs.h.  */
| #include <pthread.h>
|                         static void routine(void *a) { a = 0; }
|                         static void *start_routine(void *a) { return a; }
| int
| main ()
| {
| pthread_t th; pthread_attr_t attr;
|                         pthread_create(&th, 0, start_routine, 0);
|                         pthread_join(th, 0);
|                         pthread_attr_init(&attr);
|                         pthread_cleanup_push(routine, 0);
|                         pthread_cleanup_pop(0) /* ; */
|   ;
|   return 0;
| }
configure:7341: result: no
configure:7252: checking for pthread-config
configure:7280: result: no
configure:7528: error: Cannot initialize pthread environment
```

prologic:
Closing as per comments about; this will not work currently as `./configure` bombs out and there are serious warnings we cannot work around (_allegedly_).

PR #8568, opened 2020-04-01
Fix regressions in cloud functionality (build, CI, claiming)

### LINKED ISSUE #8561: ACLK build - return flag to 3state
##### Bug report summary

During recent last-minute changes on configure script the 3-state behavior of Cloud flag was removed by accident. Reintroduce to be able to implement #8382

underhood:
##### Summary
Multiple bugs have been detected (while I was checking #8561) after recent changes in `configure.ac`. Fixes #8561 and 2 other bugs which do not have issues created.

List of found problems and their fixes:
- __*CRITICAL*__ (broken in master) the new detection, if mosquitto can be built, is not really working in `master` due to multiple problems *(syntax C `int main (int argc, char **argv)) {`, syntax configure, missing include, missing SSL link)*. This causes ACLK link to always fail - example OUTPUT of current master Fedora 31 CI build:
```
checking if libcap should be used... no
checking if libmosquitto static lib is present (and builds)... 
checking if libwebsockets static lib is present... yes
checking if netdata agent-cloud-link can be enabled... no
checking if netdata agent-cloud-link should/will be enabled... no
```
- add missing openssl check as it must be present for ACLK (this was missing since beginning mea culpa)
-  the 3 state behavior was lost when the flag was changed from ACLK to `--disable-cloud`. 

Behavior of configure flags:
    - `no flag` given to configure, try and build ACLK but if cannot just warn and build without
    - `--enable-cloud`, build ACLK if cannot ERROR - *build should FAIL in this case*
    - `--disable-cloud`, do not build ACLK (even if could)
##### Additionally (commits from Ferroin)
This PR cherry-picks Ferroin changes for installer and CI:
>This adds a switch called --require-cloud to the installer script. When specified, it causes the script to explicitly fail if it can't build Netdata with Netdata Cloud support. It is mutually exclusive with the --disable-cloud switch, with only the first one specified being honored.
>
>This is mostly intended to ensure that CI can test the build/install both with and without the cloud, but it is likely to be of interest to users who want to ensure they can use their systems with netdata Cloud.
>
>This PR also updates the build testing CI to explicitly check builds both with and without Netdata Cloud built.

##### Additionally (commits from Amoss)
Workaround for unrelated (to this PR) claiming issue that goes in as this is the last PR before code freeze.

##### Component Name
agent build

##### Test Plan
test all flags and without flag to configure script, test with purposefully missing `mosuiqtto.a/lws.a` in `externaldeps`, test with missing `openssl-dev` package on system.

Check config.h has correct flags in each case.

### DIFF
--- a/.github/workflows/build-and-install.yml
+++ b/.github/workflows/build-and-install.yml
@@ -18,7 +18,6 @@ jobs:
           - 'archlinux:latest'
           - 'centos:8'
           - 'centos:7'
-          - 'centos:6'
           - 'debian:bullseye'
           - 'debian:buster'
           - 'debian:stretch'
--- a/.github/workflows/build-and-install.yml
+++ b/.github/workflows/build-and-install.yml
@@ -74,13 +74,13 @@ jobs:
           PRE: ${{ matrix.pre }}
         run: |
           echo $PRE > ./prep-cmd.sh
-          docker build . -f .github/dockerfiles/Dockerfile.build_test -t build_test --build-arg BASE=${{ matrix.distro }}
+          docker build . -f .github/dockerfiles/Dockerfile.build_test -t test --build-arg BASE=${{ matrix.distro }}
       - name: Regular build on ${{ matrix.distro }}
         run: |
-          docker run --rm -w /netdata build_test /bin/sh -c 'autoreconf -ivf && ./configure && make -j2'
+          docker run -w /netdata test /bin/sh -c 'autoreconf -ivf && ./configure && make -j2'
       - name: netdata-installer on ${{ matrix.distro }}
         run: |
-          docker run --rm -w /netdata build_test /bin/sh -c './netdata-installer.sh --dont-wait --dont-start-it --disable-cloud'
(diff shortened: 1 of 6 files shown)


amoss:
I've tested:
* The installer default
* The installer with --disable-cloud

Running the configure manually to simulate @Ferroin's changes:
* With --enable-cloud=yes and the dep available
* With --enable-cloud=yes and removing the mosquito.h file
* With --enable-cloud=detect both available and with the missing header.
* With --enable-cloud=no in both cases.

These all work - we should merge in @Ferroin's changes as well so that we can see the test cases in CI/CD.

Ferroin:
> I've tested:
> 
> * The installer default
> * The installer with --disable-cloud
> 
> Running the configure manually to simulate @Ferroin's changes:
> 
> * With --enable-cloud=yes and the dep available
> * With --enable-cloud=yes and removing the mosquito.h file
> * With --enable-cloud=detect both available and with the missing header.
> * With --enable-cloud=no in both cases.
> 
> These all work - we should merge in @Ferroin's changes as well so that we can see the test cases in CI/CD.

If you want to pull in the changes from #8382 here to get the CI checks forthis, that's fine by me.

PR #8579, opened 2020-04-02
Coverity scan

amoss:
##### Summary
Fix the outstanding Coverity defects before the release.

##### Component Name
various

##### Test Plan

* Single-stepped through a debugger to check the loading of `claim_id` and the private key during the ACLK establishment for a claimed agent.

### DIFF
--- a/aclk/aclk_common.c
+++ b/aclk/aclk_common.c
@@ -42,7 +42,7 @@ ACLK_PROXY_TYPE aclk_verify_proxy(const char *string)
     if (!string)
         return PROXY_TYPE_UNKNOWN;
 
-    while (*string == 0x20)
+    while (*string == 0x20 && *string!=0)   // Help coverity (compiler will remove)
         string++;
 
     if (!*string)
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -167,6 +167,7 @@ static int create_private_key()
     fclose(f);
     if (bytes_read == 0) {
         info("Claimed agent cannot establish ACLK - private key '%s' is empty.", filename);
+        freez(private_key);
         return 1;
     }
 
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -155,13 +155,7 @@ static int create_private_key()
         error("Claimed agent cannot establish ACLK - unable to load private key '%s' failed.", filename);
         return 1;
     }
-
     debug(D_ACLK, "Claimed agent loaded private key len=%ld bytes", bytes_read);
-    if (bytes_read == 0) {
-        info("Claimed agent cannot establish ACLK - private key '%s' is empty.", filename);
-        freez(private_key);
-        return 1;
-    }
 
     BIO *key_bio = BIO_new_mem_buf(private_key, -1);
     if (key_bio==NULL) {
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -153,10 +153,6 @@ static int create_private_key()
         error("Claimed agent cannot establish ACLK - private key not found '%s' failed.", filename);
         return 1;
     }
-    if (unlikely(statbuf.st_size == 0)) {
-        info("Claimed agent cannot establish ACLK - private key '%s' is empty.", filename);
-        return 1;
-    }
 
     FILE *f = fopen(filename, "rt");
     if (unlikely(f == NULL)) {
@@ -169,6 +165,10 @@ static int create_private_key()
     private_key[bytes_read] = 0;
     debug(D_ACLK, "Claimed agent loaded private key len=%zu bytes", bytes_read);
     fclose(f);
+    if (bytes_read == 0) {
+        info("Claimed agent cannot establish ACLK - private key '%s' is empty.", filename);
+        return 1;
+    }
 
     BIO *key_bio = BIO_new_mem_buf(private_key, -1);
     if (key_bio==NULL) {
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -146,25 +146,17 @@ int cloud_to_agent_parse(JSON_ENTRY *e)
 static RSA *aclk_private_key = NULL;
 static int create_private_key()
 {
-    char filename[FILENAME_MAX + 1];    struct stat statbuf;
+    char filename[FILENAME_MAX + 1];
     snprintfz(filename, FILENAME_MAX, "%s/claim.d/private.pem", netdata_configured_user_config_dir);
 
-    if (lstat(filename, &statbuf) != 0) {
-        error("Claimed agent cannot establish ACLK - private key not found '%s' failed.", filename);
+    long bytes_read;
+    char *private_key = read_by_filename(filename, &bytes_read);
+    if (!private_key) {
+        error("Claimed agent cannot establish ACLK - unable to load private key '%s' failed.", filename);
         return 1;
     }
 
-    FILE *f = fopen(filename, "rt");
-    if (unlikely(f == NULL)) {
-        error("Claimed agent cannot establish ACLK - unable to open private key '%s'.", filename);
-        return 1;
-    }
-
-    char *private_key = callocz(1, statbuf.st_size + 1);
-    size_t bytes_read = fread(private_key, 1, statbuf.st_size, f);
-    private_key[bytes_read] = 0;
-    debug(D_ACLK, "Claimed agent loaded private key len=%zu bytes", bytes_read);
-    fclose(f);
+    debug(D_ACLK, "Claimed agent loaded private key len=%ld bytes", bytes_read);
     if (bytes_read == 0) {
         info("Claimed agent cannot establish ACLK - private key '%s' is empty.", filename);
         freez(private_key);
--- a/claim/claim.c
+++ b/claim/claim.c
@@ -135,6 +135,4 @@ void load_claiming_state(void)
         return;
     }
     info("File '%s' was found. Setting state to AGENT_CLAIMED.", filename);
-
-    snprintfz(filename, FILENAME_MAX, "%s/claim.d/private.pem", netdata_configured_user_config_dir);
 }
--- a/claim/claim.c
+++ b/claim/claim.c
@@ -110,27 +110,18 @@ void load_claiming_state(void)
     }
 
     char filename[FILENAME_MAX + 1];
-    struct stat statbuf;
-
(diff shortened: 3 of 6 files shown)


thiagoftsm:
What are the Coverity numbers you are fixing?

aclk/agent_cloud_link.c
```diff
@@ -169,6 +165,10 @@ static int create_private_key()
     private_key[bytes_read] = 0;
     debug(D_ACLK, "Claimed agent loaded private key len=%zu bytes", bytes_read);
     fclose(f);
+    if (bytes_read == 0) {
```
mfundul, line 168: Shouldn't this be `bytes_read != statbuf.st_size` or something?

aclk/agent_cloud_link.c
```diff
@@ -153,10 +153,6 @@ static int create_private_key()
         error("Claimed agent cannot establish ACLK - private key not found '%s' failed.", filename);
         return 1;
     }
-    if (unlikely(statbuf.st_size == 0)) {
```
mfundul, line 156: I think this check should stay, maybe with something like `statbuf.st_size < MIN_SIZE_SOMETHING`

aclk/aclk_common.c
```diff
@@ -42,7 +42,7 @@ ACLK_PROXY_TYPE aclk_verify_proxy(const char *string)
     if (!string)
         return PROXY_TYPE_UNKNOWN;
 
-    while (*string == 0x20)
+    while (*string == 0x20 && *string!=0)   // Help coverity (compiler will remove)
```
vlvkobal, line 45: I'd better mark this as a false positive.

aclk/agent_cloud_link.c
stelfrag: > Shouldn't this be `bytes_read != statbuf.st_size` or something?

It detects (probably) that you are trying to use `statbuf.st_size` after `lstat` succeeds but there is no guarantee that once you hit `fopen` it will be the same file (in size and all).

aclk/aclk_common.c
amoss: I considered that, but the code may change in the future and making Coverity understand that the input is sanitized is more robust.

aclk/agent_cloud_link.c
amoss: Why?

amoss:
> What are the Coverity numbers you are fixing?

They are listed in the commits.

aclk/agent_cloud_link.c
amoss: No - the check is to detect that the file was empty not that the size changed.

aclk/aclk_common.c
vlvkobal: Maybe. But it looks ugly :).

claim/claim.c
```diff
@@ -141,8 +129,12 @@ void load_claiming_state(void)
...
     claimed_id[bytes_read] = 0;
-    info("File '%s' was found. Setting state to AGENT_CLAIMED.", filename);
     fclose(f);
+    if (bytes_read == 0) {
+        info("File '%s' has no contents. Setting state to AGENT_UNCLAIMED.", filename);
+        return;
+    }
+    info("File '%s' was found. Setting state to AGENT_CLAIMED.", filename);
 
     snprintfz(filename, FILENAME_MAX, "%s/claim.d/private.pem", netdata_configured_user_config_dir);
```
stelfrag, line 139: This must be a long forgotten line?

amoss: That is a bit weird, I'll remove that.

aclk/agent_cloud_link.c
mfundul: Because a 1 byte file is not valid, is it?

aclk/agent_cloud_link.c
mfundul: If you have a 64 byte file and you read 4 bytes then it is an error, no?

aclk/agent_cloud_link.c
amoss: The validity check is loading the key into openssl. There is no definite value for MIN_SIZE_SOMETHING.

aclk/agent_cloud_link.c
amoss: What if the file changed between the time of the lstat and the time of the read (which is the whole point of the TOCTOU check in Coverity).

mfundul: The problem is the call `fread(private_key, 1, statbuf.st_size, f);` which has not been fixed.

If `statbuf.st_size` is smaller than the real length then you don't read the whole value.

Moreover, if `statbuf.st_size` is exactly right, which is more likely, but you fail to read the whole file (because `bytes_read` was cut short) then it's also invalid.

You need to set the correct length for the `fread()` call and check `if (bytes_read != length)`. The "or something" comment is probably solved by doing something like:
https://stackoverflow.com/questions/238603/how-can-i-get-a-files-size-in-c

mfundul: This is where we've done this before: https://github.com/netdata/netdata/blob/8fe7485a600f60be9ac20af923a61d516cb0335e/health/health.c#L52

amoss: But you are trying to check the validity of the file by considering if the number of bytes read in is the expected length. This does not tell us that the file or its contents are valid. Attempting to load the key into openssl is what tells us that the contents were valid. The piece of code that was moved was simply to check that we do not try to read a zero-length file. It was not necessary, but was there as an early bail for a condition that is clearly wrong.

Fixing the TOCTOU has been done. The fix that you are trying to put into the code is not valid, or necessary. The reason that I moved the check rather than eliminate it completely is because I cannot tell whether calling the openssl function with a zero-length string is problematic. Calling it with invalid data will cause a failure to initialise the key.

mfundul: Calling `fread()` with the length `statbuf.st_size` isn't TOCTOU?

amoss: I don't think so. There are two scenarios:
1. An attacker appends something to the file. It is not read by the call. No damage results.
2. An attacker trunctates the file. The call returns the number of bytes read, our memory is not corrupted and the key fails to validate when we call openssl.

Are you thinking of another case?

mfundul: The problem is not security, the problem is that it's a bug. I explained above what can go wrong. The user impact would be that an operation would fail when it wouldn't need to. If coverity is happy with this change then you can ignore this. In the health code this is how @thiagoftsm fixed the coverity defect though and I believe this is the correct approach.

amoss: Can you explain a scenario in which you think the bug will be triggered and what will occur?

mfundul: The race condition is rare to begin with, coverity worries about `statbuf.st_size` being different than the file length during `fopen()`.

It's mostly going to be exactly the same 99,999% of the cases. Especially since I don't believe we're generating variable sizes for keys, although I'm not sure about key size differences.

If somehow the file-system does not atomically update this file, which I consider improbable due to its small size, the scenario is the following:

- real length < `statbuf.st_size`:
  - this is safe because `fopen()` will stop at the end of file.
- real length > `statbuf.st_size`:
  - this will cause the code to read only a part of the key and it's the bug.

Moreover, we have to account for errors in `fopen()` returns fewer bytes than what it asked, but this may be unrelated to the coverity report.

amoss: If `create_private_key()` indicates a failure then it gets called again, so in the improbable situation that the file (which is 1679 bytes long) is not atomically updated during the claim and it changes size between the two calls... we simply call it again and it is validated correctly. In fact looking at the main loop in the aclk thread I see that we call it one more time after it has succeeded, I would speculate that nobody knows why, possibly for luck? 

The security problem is a potentially a real problem which is why Coverity checks it as it gives an attacker a way to affect a code path inside the code, potentially causing a buffer overwrite. That is not a problem the way that we've written the code so in other places in Netdata the same pattern has been marked as a false positive (if you recall we had a long conversation about the portability of `fstat` last time this happened).

So although technically the piece of code can go wrong locally, it is within a larger context that will stop it being a problem. So I am happy that it will not cause a problem for users. There is a long philosophical discussion that we are not going to have about whether or not it is still a bug. Yes, in general the fseek/ftell is a better way to read the file although there potential for it to go wrong. On Linux it should be fine but not every platform returns the number of bytes from ftell (I can't remember which one doesn't so I don't know if we support it or not).

So if we've finished establishing whether or not the code is broken, can we proceed with reviewing the Coverity fixes as I would like us to get the release out on time...

mfundul: `ftell()` should be standard C, if it's indeed problematic then the existing health code has an issue. If coverity is happy with the code then I'll leave it at that. I didn't request a code change in any case.

amoss: I'll look up the `ftell` thing then and see if my memory is correct in case we have a problem.

libnetdata/libnetdata.c
```diff
@@ -1453,3 +1453,40 @@ void recursive_config_double_dir_load(const char *user_path, const char *stock_p
...
+        fclose(f);
+        return NULL;
+    }
+    long size = ftell(f);
+    if (size <= 0 || fseek(f, 0, SEEK_END) < 0) {
+        fclose(f);
+        return NULL;
+    }
+    char *contents = callocz(size + 1, 1);
+    if (!contents) {
```
mfundul, line 1474: `callocz()` never fails.

amoss: So you like to live dangerously Mr Bond.

claim/claim.c
```diff
@@ -110,27 +110,18 @@ void load_claiming_state(void)
...
-    claimed_id = callocz(1, statbuf.st_size + 1);
-    size_t bytes_read = fread(claimed_id, 1, statbuf.st_size, f);
-    claimed_id[bytes_read] = 0;
-    fclose(f);
     if (bytes_read == 0) {
+        freez(claimed_id);
+        claimed_id = NULL;
         info("File '%s' has no contents. Setting state to AGENT_UNCLAIMED.", filename);
         return;
     }
```
mfundul: `bytes_read` can no longer be 0. `read_by_filename()` has already returned `NULL` before that.

Moreover, it is impossible for `read_by_filename()` to return a valid `claimed_id` pointer while at the same type setting `&bytes_read` to 0.

Finally, `bytes_read` is not initialized and `read_by_filename()` is not guaranteed to always set the pointer.

I think we can remove the whole `if(){` statement.

PR #8586, opened 2020-04-02
Fix Coverity Defect CID-349684

### LINKED ISSUE #8585: Coverity 349684
##### Bug report summary
Accoding Coverity we have a situation where a SILENCER can create a memory leak in the following code present in the file `libnetdata/health/health.c`:

```
Line 142:
CID 349684 (#1 of 1): Resource leak (RESOURCE_LEAK)5. leaked_storage: Variable test going out of scope leaks the storage it points to.
```

##### OS / Environment
```
Linux hades 5.4.29 #1 SMP Thu Apr 2 11:02:57 UTC 2020 x86_64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz GenuineIntel GNU/Linux
/etc/os-release:ID=slackware
```

##### Netdata version
 netdata v1.20.0-302-g71290d9c

##### Component Name
Health

##### Expected behavior
It expected to eliminate the possibility of memory leak with health.

thiagoftsm:
##### Summary
Fixes #8585 

This PR brings a fix for the coverity problem, now we are checking whether the silencer was added when the function is called, case it was not added, we add the pointer to list, on the other hand we release the pointer
##### Component Name
Health
##### Test Plan
To create the JSON files parsed from Netdata run the following command:

```bash
$ netdata
$ curl -k "https://localhost:19999/api/v1/manage/health?cmd=DISABLE ALL" -H "X-Auth-Token: TOKEN"
$ killall netdata
$ curl -k "https://localhost:19999/api/v1/manage/health?cmd=RESET" -H "X-Auth-Token: TOKEN"
$ curl -k "https://localhost:19999/api/v1/manage/health?cmd=SILENCE&context=load" -H "X-Auth-Token: TOKEN"
$ killall netdata
$ curl -k "https://localhost:19999/api/v1/manage/health?cmd=RESET" -H "X-Auth-Token: TOKEN"
```

After all `killall` verify whether your `error.log` demonstrates that Netdata was finalized with success.
##### Additional Information
The `libjson-c` is not expected to create errors for this case, but you can also test with it. My suggestion is to use an environment that will force the utilization of the other Json library present with Netdata.

### DIFF
--- a/libnetdata/health/health.c
+++ b/libnetdata/health/health.c
@@ -100,44 +100,6 @@ SILENCER *health_silencers_addparam(SILENCER *silencer, char *key, char *value) 
 }
 
 /**
- * Is SILENCER linked?
- *
- * Check whether the SILENCER is linked
- *
- * @param test is the silencer pointer
- *
- * @return It returns 1 case it is linked and 0 otherwise
- */
-int helth_is_silencer_linked(SILENCER *test) {
-    SILENCER *silencer;
-    for(silencer = silencers->silencers; silencer ; silencer = silencer->next) {
-        if (test == silencer)
-            return 1;
-    }
-
-    return 0;
-}
-
-/**
- * Free Silencer pointers
- *
- * This function cleans the silencer pointers and the SILENCER itself.
- */
-void health_free_sielencer_pointers(SILENCER *t) {
-    simple_pattern_free(t->alarms_pattern);
-    simple_pattern_free(t->charts_pattern);
-    simple_pattern_free(t->contexts_pattern);
-    simple_pattern_free(t->hosts_pattern);
-    simple_pattern_free(t->families_pattern);
-    freez(t->alarms);
-    freez(t->charts);
-    freez(t->contexts);
-    freez(t->hosts);
-    freez(t->families);
-    freez(t);
-}
-
-/**
  * JSON Read Callback
  *
  * Callback called by netdata to create the silencer.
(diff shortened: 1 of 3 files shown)


libnetdata/health/health.c
```diff
@@ -99,6 +99,44 @@ SILENCER *health_silencers_addparam(SILENCER *silencer, char *key, char *value)
...
+ *
+ * Check whether the SILENCER is linked
+ *
+ * @param test is the silencer pointer
+ *
+ * @return It returns 1 case it is linked and 0 otherwise
+ */
+int helth_is_silencer_linked(SILENCER *test) {
+    SILENCER *silencer;
+    for(silencer = silencers->silencers; silencer ; silencer = silencer->next) {
```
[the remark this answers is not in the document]
thiagoftsm: This code we are fixing is not the normal road used to create a SILENCER. We are creating this option to protect the code case in the future by accident somebody calls this function to create a new SILENCER.
We do not create a SILENCER in normal situation, this is done only when the user uses the API and for this case, the road to create a SILENCER is another optimized, but it does not have any locker too.

libnetdata/health/health.c
```diff
@@ -138,8 +176,13 @@ int health_silencers_json_read_callback(JSON_ENTRY *e)
...
             } else {
                 debug(D_HEALTH, "JSON: Adding %s=%s", e->name, e->data.string);
-                SILENCER *test = health_silencers_addparam(e->callback_data, e->name, e->data.string);
-                (void)test;
+                SILENCER *silencer = health_silencers_addparam(e->callback_data, e->name, e->data.string);
+                if(silencer) {
+                   if (!helth_is_silencer_linked(silencer))
+                       health_silencers_add(silencer);
+                   else
+                       health_free_sielencer_pointers(silencer);
```
vlvkobal, line 184: Why do you free pointers here? The result of the condition is inconsistent in different cases.

stelfrag: This makes things worse I think. The parser will have created a silencer and stored in `e->callback_data` (L121-L123). and  added to the linked list. 

So adding a param (L179) may override values in the silencer structure (memory leak), then it will see that it is linked and remove it (but will remain linked)

thiagoftsm: Thank you @stelfrag to explain!

PR #8591, opened 2020-04-03
fix resource leak (Coverity CID 349684)

mfundul:
##### Summary
Fixes Coverity CID 349684 

Resource leak (RESOURCE_LEAK) leaked_storage: Variable `test` going out of scope leaks the storage it points to.

```
3. alloc_fn: Storage is returned from allocation function health_silencers_addparam. [show details]
4. var_assign: Assigning: test = storage returned from health_silencers_addparam(e->callback_data, e->name, e->data.string).
141                SILENCER *test = health_silencers_addparam(e->callback_data, e->name, e->data.string);
142                (void)test;
CID 349684 (#1 of 1): Resource leak (RESOURCE_LEAK)
5. leaked_storage: Variable test going out of scope leaks the storage it points to.
143            }
144            break;
```

##### Component Name
health

### DIFF
--- a/libnetdata/health/health.c
+++ b/libnetdata/health/health.c
@@ -138,8 +138,7 @@ int health_silencers_json_read_callback(JSON_ENTRY *e)
                 else if (!strcmp(e->data.string,"DISABLE")) silencers->stype = STYPE_DISABLE_ALARMS;
             } else {
                 debug(D_HEALTH, "JSON: Adding %s=%s", e->name, e->data.string);
-                SILENCER *test = health_silencers_addparam(e->callback_data, e->name, e->data.string);
-                (void)test;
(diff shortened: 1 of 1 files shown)


thiagoftsm:
What will happen case there was an allocation in this function and a valid point was returned without to be linked? 
Did you test this solution on Coverity? I think it is creating another warning there.

mfundul:
> 
> 
> What will happen case there was an allocation in this function and a valid point was returned without to be linked?
> Did you test this solution on Coverity? I think it is creating another warning there.

You are right, it creates a different warning:
`Failing to save or free storage allocated by "health_silencers_addparam(e->callback_data, e->name, e->data.string)" leaks it.`

PR #8595, opened 2020-04-03
Correctly fixed RPM package builds on Fedora.

Ferroin:
##### Summary

Fedora builds it's packages using `-fPIC`, but we were not building LWS and mosquitto with this flag, causing the final link to fail.  This modifies the RPM file to ensure they get built with this flag independent of the environment, which will work everywhere unless someone tries to build the package with PIC disabled (which no distribution currently does).

##### Component Name

area/packaging

##### Test Plan

Verified locally using Docker package builds. Verified in Travis at https://travis-ci.org/github/Ferroin/netdata/builds/670595759

##### Additional Information

This differs from the original attempts at fixing this in that it modifies the environment the scripts we use to bundle LWS and mosquitto run in directly, instead of relying on multiple layers to pass through the required environment variables.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -329,7 +329,6 @@ jobs:
       <<: *RPM_TEMPLATE
       if: commit_message =~ /\[Package (amd64|arm64) RPM( Fedora)?\]/
       env:
-        - CFLAGS="-fPIC"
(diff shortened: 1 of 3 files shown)


Ferroin:
Merging as this is an important fix, has been approved by the primary person qualified to review it, and there's an imminent bugfix build incoming for the nightly channel.

PR #8606, opened 2020-04-05
packaging: fix errors during install-requred-packages

ilyam8:
### Summary

My system:

```cmd
[ilyam@ilyam-pc ~]$ cat /etc/os-release
NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
PRETTY_NAME="Manjaro Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.manjaro.org/"
SUPPORT_URL="https://www.manjaro.org/"
BUG_REPORT_URL="https://bugs.manjaro.org/"
LOGO=manjarolinux
```

There are two errors during execution of `install-requred-packages.sh`.

- `Unknown distribution ID: manjaro`
- `validate_tree_arch: command not found`

First one is minor and is likely expected but it makes me thingking that something is going wrong.

This PR fixed both errors.

##### Component Name

`packaging`

##### Test Plan

I did no tests, i think ci should be enough.

### DIFF
--- a/packaging/installer/install-required-packages.sh
+++ b/packaging/installer/install-required-packages.sh
@@ -192,7 +192,7 @@ get_os_release() {
   eval "$(grep -E "^(NAME|ID|ID_LIKE|VERSION|VERSION_ID)=" "${os_release_file}")"
   for x in "${ID}" ${ID_LIKE}; do
     case "${x,,}" in
-      alpine | arch | centos | debian | fedora | gentoo | sabayon | rhel | ubuntu | suse | opensuse-leap | sles | clear-linux-os)
+      alpine | arch | centos | clear-linux-os | debian | fedora | gentoo | manjaro | opensuse-leap | rhel | sabayon | sles | suse | ubuntu)
         distribution="${x}"
         version="${VERSION_ID}"
         codename="${VERSION}"
--- a/packaging/installer/install-required-packages.sh
+++ b/packaging/installer/install-required-packages.sh
@@ -1047,7 +1047,9 @@ declare -A pkg_zip=(
 )
 
 validate_package_trees() {
-  validate_tree_${tree}
+  if type -t validate_tree_${tree} > /dev/null; then
+    validate_tree_${tree}
+  fi
 }
 
 validate_installed_package() {


ilyam8:
Probably it is expected

https://github.com/ilyam8/netdata/blob/e1cec5a69db211df14031544614f97c2f60ca5b4/packaging/installer/install-required-packages.sh#L72

Feel free to `wontfix` && close the PR @Ferroin @prologic 

I did it because i saw `Unknow distribution ID` message.

ilyam8:
I see we have only `validate_tree_centos`

https://github.com/netdata/netdata/blob/a851701bb86967b0bb1c7cbcd135b7eace578898/packaging/installer/install-required-packages.sh#L1318-L1319

It is not clear why we have

https://github.com/netdata/netdata/blob/a851701bb86967b0bb1c7cbcd135b7eace578898/packaging/installer/install-required-packages.sh#L1049-L1051

and use it

prologic:
> I did no tests, i think ci should be enough.

Ci is only enough if you actually modify the appropriate workflow and add Manjaro to it :)

prologic:
@illes `validate_trees()` is in case there are other special systems we have to cater for and do "extra" things to modify or alter their package trees. So far this is only CentOS; but "future proof" :)

ilyam8:
> Ci is only enough if you actually modify the appropriate workflow and add Manjaro to it :)


I am not sure this was needed, but take a look

we set distribution in `autodetect_distribution`. There is no `manjaro`

https://github.com/netdata/netdata/blob/a851701bb86967b0bb1c7cbcd135b7eace578898/packaging/installer/install-required-packages.sh#L1850-L1854

Then we execute `detect_package_manager_from_distribution`

https://github.com/netdata/netdata/blob/a851701bb86967b0bb1c7cbcd135b7eace578898/packaging/installer/install-required-packages.sh#L1856-L1859

And there we expect `manjaro`

https://github.com/netdata/netdata/blob/a851701bb86967b0bb1c7cbcd135b7eace578898/packaging/installer/install-required-packages.sh#L341-L344

prologic:
Fine if you don't want/need it tested regularly in CI that's okay :)

I _assume_ [this](https://hub.docker.com/r/jonathonf/manjaro/) is the correct upstream Docker image to use if we change our minds later?

ilyam8:
> Ci is only enough if you actually modify the appropriate workflow and add Manjaro to it :)

manjaro is arch actually. Main difference is delayed updates (1-2 weeks).

I was fixing cosmetic error to be honest 😄

ilyam8:
> I assume this is the correct upstream Docker image

i guess yes, that jonathonf guy is from the manjaro team

ilyam8:
i am still not sure in https://github.com/netdata/netdata/pull/8606/commits/e1cec5a69db211df14031544614f97c2f60ca5b4

if you @prologic or @Ferroin think we dont need it i remove it

prologic:
Its fine. Get another approval from @Ferroin and merge it :)

ilyam8:
> Current approach does not work, as 'manjaro' isn't a recognized distribution name to match on later for selecting the package manager.


https://github.com/netdata/netdata/blob/a851701bb86967b0bb1c7cbcd135b7eace578898/packaging/installer/install-required-packages.sh#L341-L344

Ferroin:
> > Current approach does not work, as 'manjaro' isn't a recognized distribution name to match on later for selecting the package manager.
> 
> https://github.com/netdata/netdata/blob/a851701bb86967b0bb1c7cbcd135b7eace578898/packaging/installer/install-required-packages.sh#L341-L344

Sorry, missed that somehow.

ilyam8:
> as 'manjaro' isn't a recognized distribution name to match on later for selecting the package manager.

if it was the case, ci would fail, right? i am asking because i did not tests, i rely on ci checks

Ferroin:
> > as 'manjaro' isn't a recognized distribution name to match on later for selecting the package manager.
> 
> if it was the case, ci would fail, right? i am asking because i did no tests, i rely on ci checks

Manjaro isn't in our CI checks at all. There appear to be no up-to-date Docker images for it at the moment AFAICT. Long term we should probably tr and get it added, but It's mostly just old Arch with a fancy installer and some slightly different default behavior, so it's not exactly high priority.

ilyam8:
> Manjaro isn't in our CI checks at all.

i mean install-required-packages job would fail if we had no `manjaro` here?

https://github.com/netdata/netdata/blob/a851701bb86967b0bb1c7cbcd135b7eace578898/packaging/installer/install-required-packages.sh#L341-L344

PR #8611, opened 2020-04-06
Improved ACLK memory management and shutdown sequence

### LINKED ISSUE #8483: ACLK code cleanup
#### Summary
- Improve main thread cleanup during shutdown
  - There is an issue on how the `netdata_exit` variable is used to terminate the processing loop
- Trace memory leaks in the main ACLK code
   - Include MQTT library and LWS

stelfrag:
Fixes #8483 

##### Summary
- Improve the ACLK main thread shutdown sequence. 
  - The thread's cancelability is disabled and the shutdown is detected by the `netdata_exit` variable.
     - The netdata core would immediately cancel/kill the ACLK main thread during agent shutdown so the code outside the while loop would never be executed.
- The ACLK query thread empties the query queue and frees resources on shutdown
- Fixed memory leak (payload type parameter)
- Improved error / info message on disconnection from cloud
  - When the connection is dropped due to the agent shutting down, it logs an information message
  - When the connection actually fails, it logs an error message
- Trying to queue identical update commands to the query engine (i.e. an identical update is already in the queue) is now considered a success

##### Component Name
ACLK

##### Test Plan
- Compile agent with ACLK enabled (must be claimed)
- To test fixed memory leaks
  - After starting the agent as described below, switch to the cloud and observe charts to make sure the ACLK is in use (i.e. the agent will accept and process cloud queries)
  - Use valgrind e.g: 
     * `valgrind --undef-value-errors=no --leak-check=full -s <netdata executable>`
     * Allow it to run for sometime, stop the server and observe the valgrind report
- To test that the error `ACLK failed to queue chart_update command` is gone
  - Start the agent with as many plugins enabled as possible
    - The rapid initialization of the plugins, creating charts and adding dimensions within the same second would have triggered the `ACLK failed to queue chart_update command` in the error.log

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -1283,7 +1283,6 @@ void *aclk_main(void *ptr)
         info("Killing ACLK thread -> cloud functionality has been disabled");
         return NULL;
     }
-    netdata_thread_cleanup_push(aclk_main_cleanup, ptr);
 
     info("Waiting for netdata to be ready");
     while (!netdata_ready) {
@@ -1387,7 +1386,7 @@ exited:
     if (aclk_private_key != NULL)
         RSA_free(aclk_private_key);
 
-    netdata_thread_cleanup_pop(1);
+    aclk_main_cleanup(ptr);
     return NULL;
 }
 
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -187,7 +187,7 @@ biofailed:
  * should be called with
  *
  * mode 0 to reset the delay
- * mode 1 to sleep for the calculated amount of time [0 .. ACLK_MAX_BACKOFF_DELAY * 1000] ms
+ * mode 1 to calculate sleep time [0 .. ACLK_MAX_BACKOFF_DELAY * 1000] ms
  *
  */
 unsigned long int aclk_reconnect_delay(int mode)
@@ -209,8 +209,6 @@ unsigned long int aclk_reconnect_delay(int mode)
         fail++;
         delay = (delay * 1000) + (random() % 1000);
     }
-
-    //    sleep_usec(USEC_PER_MS * delay);
 
     return delay;
 }
@@ -308,7 +306,7 @@ int aclk_queue_query(char *topic, char *data, char *msg_id, char *query, int run
         if (tmp_query->run_after == run_after) {
             QUERY_UNLOCK;
             QUERY_THREAD_WAKEUP;
-            return 1;
+            return 0;
         }
 
         if (last_query)
@@ -865,18 +863,22 @@ int aclk_process_queries()
 static void aclk_query_thread_cleanup(void *ptr)
 {
     struct netdata_static_thread *static_thread = (struct netdata_static_thread *)ptr;
-    static_thread->enabled = NETDATA_MAIN_THREAD_EXITING;
 
     info("cleaning up...");
-
-    COLLECTOR_LOCK;
 
     _reset_collector_list();
     freez(collector_list);
 
-    COLLECTOR_UNLOCK;
+    // Clean memory for pending queries if any
+    struct aclk_query *this_query;
 
-    static_thread->enabled = NETDATA_MAIN_THREAD_EXITED;
+    do {
+        this_query = aclk_queue_pop();
+        aclk_query_free(this_query);
+    } while (this_query);
+
+    freez(static_thread->thread);
+    freez(static_thread);
 }
 
 /**
@@ -973,7 +975,6 @@ static void aclk_main_cleanup(void *ptr)
         }
     }
 
-    info("Disconnected");
 
     static_thread->enabled = NETDATA_MAIN_THREAD_EXITED;
 }
@@ -1322,6 +1323,8 @@ void *aclk_main(void *ptr)
 
     usec_t reconnect_expiry = 0; // In usecs
 
+    netdata_thread_disable_cancelability();
+
     while (!netdata_exit) {
         static int first_init = 0;
         size_t write_q, write_q_bytes, read_q;
@@ -1375,7 +1378,8 @@ void *aclk_main(void *ptr)
         }
     } // forever
 exited:
-    aclk_shutdown();
+    // Wakeup query thread to cleanup
+    QUERY_THREAD_WAKEUP;
 
     freez(aclk_username);
     freez(aclk_password);
@@ -1867,6 +1871,12 @@ int aclk_handle_cloud_request(char *payload)
             freez(cloud_to_agent.callback_topic);
 
         return 1;
+    }
+
+    // Checked to be "http", not needed anymore
+    if (likely(cloud_to_agent.type_id)) {
+        freez(cloud_to_agent.type_id);
+        cloud_to_agent.type_id = NULL;
     }
 
     if (unlikely(aclk_submit_request(&cloud_to_agent)))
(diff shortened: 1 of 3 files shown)


underhood:
@stelfrag can you rebase on master as there seems to be conflict

amoss:
In `aclk_main` should we push the cleanup if `!netdata_cloud_setting` ?

amoss:
I see this in valgrind, can we determine if it comes from the ACLK code or somewhere else in Netdata?

```
==24264== Thread 1:
==24264== 1,584 (176 direct, 1,408 indirect) bytes in 1 blocks are definitely lost in loss record 11,258 of 11,440
==24264==    at 0x483577F: malloc (vg_replace_malloc.c:299)
==24264==    by 0x4D90558: CRYPTO_zalloc (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4DBF4FE: RSA_new_method (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4DBDB64: ??? (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4CB53CC: ??? (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4CB2D6A: ??? (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4CB394C: ASN1_item_ex_d2i (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4CB39CA: ASN1_item_d2i (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4DBC26D: ??? (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4CAEFEA: d2i_PrivateKey (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4DA4A3A: PEM_read_bio_PrivateKey (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)
==24264==    by 0x4DA0DCA: PEM_read_bio_RSAPrivateKey (in /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1)

```

amoss:
There seems to be a mixture of:
```
RSA_free(aclk_private_key);
```
and
```
freez(private_key);
```
I think the freez call is wrong (L172 in agent_cloud_link.c) and should be `RSA_free`.

stelfrag:
> In `aclk_main` should we push the cleanup if `!netdata_cloud_setting` ?

Good point -- no need to push cleanup before that check

stelfrag:
> I think the freez call is wrong (L172 in agent_cloud_link.c) and should be `RSA_free`.

No that looks. It comes from the `read_by_filename` (uses callocz on that)

amoss:
In fact looking more closely at `create_private_key` we have multiple problems:

* It should not be inside that if statement.
* There should be an RSA_free inside.

e.g.

```
    BIO_free(key_bio);
    freez(private_key);
    if (aclk_private_key!=NULL)
    {
        RSA_free(aclk_private_key);
        return 0;
    }
```

underhood:
OK, tested this with cloud (after reinstalling `json-c`) and reviewed. Appart from Valgrind thing (which is probably the freeing of the key) that Andrew mention already I do not notice any obvious issues

underhood:
```
==71908== 1,584 (176 direct, 1,408 indirect) bytes in 1 blocks are definitely lost in loss record 11,357 of 11,570
==71908==    at 0x483980B: malloc (vg_replace_malloc.c:309)
==71908==    by 0x4B545CD: CRYPTO_zalloc (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4B86533: RSA_new_method (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4B83BD4: ??? (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4A6D984: ??? (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4A6B30A: ??? (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4A6BEB1: ASN1_item_ex_d2i (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4A6BF3E: ASN1_item_d2i (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4B82241: ??? (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4A6749B: d2i_PrivateKey (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4B6948A: PEM_read_bio_PrivateKey (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4B6541E: PEM_read_bio_RSAPrivateKey (in /usr/lib64/libcrypto.so.1.1.1d)
==71908==    by 0x4BC067: create_private_key (agent_cloud_link.c:168)
==71908==    by 0x4BF4DC: aclk_main (agent_cloud_link.c:1313)
==71908==    by 0x43C9B1: thread_start (threads.c:170)
==71908==    by 0x4F054E1: start_thread (in /usr/lib64/libpthread-2.30.so)
==71908==    by 0x501F6A2: clone (in /usr/lib64/libc-2.30.so)
```

stelfrag:
> OK, tested this with cloud (after reinstalling `json-c`) and reviewed. Appart from Valgrind thing (which is probably the freeing of the key) that Andrew mention already I do not notice any obvious issues

That issue if caused by `create_private_key()` called twice at about  L1318, L1327 and allocating `aclk_private_key` twice

amoss:
The call after the loop (that does not check the return value) can be removed. Perhaps we should check if `aclk_private_key!=NULL` inside `create_private_key` and just return success in that case?

All the tests look good though, I'm about to approve this.

stelfrag:
> The call after the loop (that does not check the return value) can be removed. Perhaps we should check if `aclk_private_key!=NULL` inside `create_private_key` and just return success in that case?
> 
> All the tests look good though, I'm about to approve this.

Yes I removed that call. Valgrind error went away

amoss:
Question:

We disable thread cancelability before the main loop, but we have a cleanup handler registered and it is removed as the last step in the thread main.

* Why do we have a cleanup if it is not used?
* Is there anything in that cleanup that we should be doing?

underhood:
Valgrind looks good now
```
==89546== LEAK SUMMARY:
==89546==    definitely lost: 0 bytes in 0 blocks
==89546==    indirectly lost: 0 bytes in 0 blocks
==89546==      possibly lost: 0 bytes in 0 blocks
==89546==    still reachable: 4,918,562 bytes in 42,172 blocks
==89546==         suppressed: 104 bytes in 1 blocks
```

stelfrag:
> Question:
> 
> We disable thread cancelability before the main loop, but we have a cleanup handler registered and it is removed as the last step in the thread main.
> 
> * Why do we have a cleanup if it is not used?
> * Is there anything in that cleanup that we should be doing?

The `netdata_thread_cleanup_pop` will execute the cleanup

amoss:
> The `netdata_thread_cleanup_pop` will execute the cleanup

Yes :) But why.... If it is only being called in that way why are we using a cleanup handler? i.e. I know it is called in the normal flow, but it is never called in an abnormal flow.

mfundul:
Guys ideally we shouldn't add anything posix thread spefic to facilitate the move to libuv threads. Thread cancelability is not supported in libuv.

amoss:
True, but without disabling cancel-ability the code after the main loop does not get called. The threads are killed before the netdata_exit flag is set.

amoss:
@stelfrag spotted this when he was cleaning up. I was going to ask you about https://github.com/netdata/netdata/commit/28d5018f3

Issue #7173, opened 2019-10-23
Implement thread dispatching and interthread communication in Exporting engine


vlvkobal:
- [x] Dispatch threads
- [x] Recheck all locks
- [x] Check all error cases, rewrite the code in order to respect locks

PR #8635, opened 2020-04-07
Show internal stats for the exporting engine

### LINKED ISSUE #7171: Collect and display internal statistics for Exporting engine
- Respect locks.
- Is it better to print stats from the main thread or from every connector worker?

vlvkobal:
##### Summary
Internal stats for the exporting engine are collected and displayed.

Closes #7171

##### Component Name
exporting engine

##### Test Plan
Check internal statistics in "Netdta Monitoring" menu for every connector and for the main exporting engine thread
image

### DIFF
--- a/exporting/send_internal_metrics.c
+++ b/exporting/send_internal_metrics.c
@@ -4,6 +4,10 @@
 
 /**
  * Create a chart for the main exporting thread CPU usage
+ *
+ * @param st_rusage the thead CPU usage chart
+ * @param rd_user a dimension for user CPU usage
+ * @param rd_system a dimension for system CPU usage
  */
 void create_main_rusage_chart(RRDSET **st_rusage, RRDDIM **rd_user, RRDDIM **rd_system)
 {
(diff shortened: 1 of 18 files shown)


exporting/send_internal_metrics.c
```diff
@@ -3,16 +3,170 @@
...
+ * Send the main exporting thread CPU usage
+ *
+ * @param st_rusage a thead CPU usage chart
+ * @param rd_user a dimension for user CPU usage
+ * @param rd_system a dimension for system CPU usage
+ */
+void send_main_rusage(RRDSET *st_rusage, RRDDIM *rd_user, RRDDIM *rd_system)
+{
+    struct rusage thread;
+    getrusage(RUSAGE_THREAD, &thread);
```
thiagoftsm, line 35: This function returns an integer, should we not test and report possible errors?

vlvkobal: I did it intentionally - it's the scope of the next planned issue #7173.

vlvkobal: Only two errors are possible in general, but not in this case (as well as in the other 31 places in the Netdata code). So, it doesn't make much sense to do an additional check.

exporting/send_internal_metrics.c
```diff
@@ -3,16 +3,170 @@
...
+
+    // ------------------------------------------------------------------------
+
+    struct rusage thread;
+    getrusage(RUSAGE_THREAD, &thread);
+
+    if (likely(stats->st_rusage->counter_done))
+        rrdset_next(stats->st_rusage);
+
+    rrddim_set_by_pointer(stats->st_rusage, stats->rd_user,   thread.ru_utime.tv_sec * 1000000ULL + thread.ru_utime.tv_usec);
```
thiagoftsm, line 168: Here you are multiplying the seconds by `1000000ULL` and you sum the result to microseconds, please, can you confirm that you did not want multiply the microseconds by 10 seconds?

exporting/send_internal_metrics.c
thiagoftsm: All right right, thank you!

exporting/send_internal_metrics.c
vlvkobal: No. The result in microseconds and is divided by 1000 to get microseconds in the chart.

ilyam8:
@vlvkobal since we have exporting stats how about to add some alarms?

vlvkobal:
It's not a critical thing. I'll add such alarms to our wish list.

PR #8643, opened 2020-04-08
Update freebsd.md

thenktor:
More detailed description how to install from packages collection

### DIFF
--- a/packaging/installer/methods/freebsd.md
+++ b/packaging/installer/methods/freebsd.md
@@ -7,8 +7,17 @@ custom_edit_url: https://github.com/netdata/netdata/edit/master/packaging/instal
 
 # Install Netdata on FreeBSD
 
-You can install Netdata from either the `ports` or `packages` collections.
+## Install from ports/packages
+You can install Netdata from either the `ports` or `packages` collections. To install from packages:
+```sh
+# pkg install netdata
+```
+You also need to enable the netdata service in `/etc/rc.conf` (add `netdata_enable="YES"`) and start the service:
+```sh
+# service netdata start
+```
 
+## Install latest version
 This is how to install the latest Netdata version from source on FreeBSD:
 
 ```sh


joelhans:
Adding a few more reviewers who can review this for technical accuracy.

packaging/installer/methods/freebsd.md
```diff
@@ -7,8 +7,17 @@ custom_edit_url: https://github.com/netdata/netdata/edit/master/packaging/instal
...
+You can install Netdata from either the `ports` or `packages` collections. To install from packages:
+```sh
+# pkg install netdata
+```
+You also need to enable the netdata service in `/etc/rc.conf` (add `netdata_enable="YES"`) and start the service:
+```sh
+# service netdata start
+```
 
+## Install latest version
```
joelhans, line 20: I think it's more correct to differentiate between ports/packages and source.

```suggestion
## Install from source

```

packaging/installer/methods/freebsd.md
```diff
@@ -7,8 +7,17 @@ custom_edit_url: https://github.com/netdata/netdata/edit/master/packaging/instal
...
-You can install Netdata from either the `ports` or `packages` collections.
+## Install from ports/packages
+You can install Netdata from either the `ports` or `packages` collections. To install from packages:
+```sh
+# pkg install netdata
+```
+You also need to enable the netdata service in `/etc/rc.conf` (add `netdata_enable="YES"`) and start the service:
+```sh
+# service netdata start
+```
```
joelhans, line 18: Just adding some newlines between text and codeblocks for readability.

```suggestion
## Install from ports/packages

You can install Netdata from either the `ports` or `packages` collections. To install from packages:

```sh
# pkg install netdata
```

You also need to enable the netdata service in `/etc/rc.conf` (add `netdata_enable="YES"`) and start the service:

```sh
# service netdata start
```
```

File: packaging/installer/methods/freebsd.md (hunk 1 above)
Comment (thenktor) [reply packaging/installer/methods/freebsd.md:20]: Actually installing from 'ports' also is installing from source. So probably 'latest source version' may be correct.

Comment (thenktor) [reply packaging/installer/methods/freebsd.md:20]: See: https://www.freshports.org/net-mgmt/netdata/
Installing from ports should work like this: `cd /usr/ports/net-mgmt/netdata/ && make install clean`
But I did not test and I'm not an experienced FreeBSD user.

Comment (thiagoftsm) [reply packaging/installer/methods/freebsd.md:20]: I will give additional information to complement this conversation, before to write this comment, I ran the following command:

```
portsnap fetch extract
```

When FreeBSD finished the extraction process, I verified that right now we have `Netdata 1.20` inside `/usr/ports/net-mgmt/netdata`, while we released Netdata 1.21 last Monday.

thiagoftsm: > See: https://www.freshports.org/net-mgmt/netdata/
> Installing from ports should work like this: `cd /usr/ports/net-mgmt/netdata/ && make install clean`
> But I did not test and I'm not an experienced FreeBSD user.

These are the steps to compile, you are right.

prologic: This is probably because we currently don't maintain the FreeBSD ports tree or packages. We _shoudl_ and we'll get there.

Issue #6470, opened 2019-07-16
Variables exposed as prometheus metrics grow stale over time


jabbrwcky:
##### Bug report summary

When scraping metrics with prometheus including variables, the variables are exported with a stale timestamp causing the metrics to be dropped by prometheus, if the Netdata process has been running for some while, e.g.:

```
# COMMENT netdata "[host]" to prometheus "[IP]", source "as collected", last seen 403962 seconds ago

netdata_info{instance="[host]",application="netdata",version="1.10.1_rolling"} 1 1563257864635

# COMMENT global host and chart variables
netdata_active_processors 4.0000000 1562759185000
```

The variables report the timestamp netdata was last restarted (first scrape possibly?), others report the current time.

Once the static data has aged enough, the variable values will start failing in the scrape.

##### OS / Environment

 ubuntu 14.04

##### Netdata version (ouput of `netdata -V`)

netdata 1.10.1_rolling

##### Component Name

prometheus export

##### Steps To Reproduce

On a host where netdata has been running for some time scrape metrics using an prometheus server also scraping the constant variables:

`http:/host:19999/api/v1/allmetrics\?format=prometheus\&source=raw\&variables=yes`

##### Expected behavior

The exporter should return static metrics values with the current timestamp, not the time the metrics were recorded.

ilyam8:
> https://docs.netdata.cloud/collectors/plugins.d/#variable

> These variables can be set and updated at any point.


> > The variables report the timestamp netdata was last restarted (first scrape possibly?), others report the current time.

According wiki it is valid to set `var` only on start and dont update the value during runtime (doesnt make sense if the value is not changed).

cakrit:
@jabbrwcky please upgrade to the latest version first. A lot of time has passed since 1.10. 
I'm not sure if there are any changes to this behavior since then, but it's always a good idea to have your netdata updated.

ilyam8:
wow, i didnt notice the version :see_no_evil:

jabbrwcky:
I'll have to ask our operations guys if they can upgrade it in general - I will try installing a more recent version manually on one of the hosts and see if the issue is still there

jabbrwcky:
I just tried `netdata v1.16.0-43-nightly` on one of the hosts and the timestamps of variables and changing metrics differ (not by much as I just started netdata), so I expect the error to occur also with the latest version when the process has been running long enough.

PR #8649, opened 2020-04-08
Ceph: Added OSD size collection

elelayan:
##### Summary

Added OSD size (=capacity) collection to the ceph plugin.
This will usually not vary:  an osd has a defined capacity based on the
underlying disk.
So it might not make sense to collect it, contrary to the osd_usage, which is the used space and varies as data is written to ceph.
But not collecting the osd_size prevents from computing used ratio (osd_usage / osd_size) or available space (osd_size - osd_usage), which are useful metrics.

##### Component Name

Ceph python plugin

##### Test Plan

I use the modified plugin on a production cluster:
 - PROXMOX Virtual Environment 6.1-8
 - Ceph version 14.2.8

I verified that the additional data was collected.

##### Additional Information

The 'kb' field has the same format as the 'kb_used' field in `ceph osd df --json` output.

Sample  `ceph osd df --json` output:

    {
      "nodes": [
        {
          "id": 12,
          "device_class": "ssd",
          "name": "osd.12",
          "type": "osd",
          "type_id": 0,
          "crush_weight": 1.8189849853515625,
          "depth": 2,
          "pool_weights": {},
          "reweight": 1,
          "kb": 1953513472,
          "kb_used": 723914120,
          "kb_used_data": 722465792,
          "kb_used_omap": 16,
          "kb_used_meta": 1448303,
          "kb_avail": 1229599352,
          "utilization": 37.05703238682349,
          "var": 0.6282679826388614,
          "pgs": 10,
          "status": "up"
        },
    
        (...)
    
      "stray": [],
      "summary": {
        "total_kb": 42477871104,
        "total_kb_used": 25054656432,
        "total_kb_used_data": 24961224336,
        "total_kb_used_omap": 13949325,
        "total_kb_used_meta": 79482162,
        "total_kb_avail": 17423214672,
        "average_utilization": 58.98284396281971,
        "min_var": 0.6262459694797988,
        "max_var": 1.2928955206659904,
        "dev": 13.779317402301619
      }

### DIFF
--- a/collectors/python.d.plugin/ceph/ceph.chart.py
+++ b/collectors/python.d.plugin/ceph/ceph.chart.py
@@ -31,6 +31,7 @@ ORDER = [
     'pool_read_operations',
     'pool_write_operations',
     'osd_usage',
+    'osd_size',
     'osd_apply_latency',
     'osd_commit_latency'
 ]
@@ -99,6 +100,10 @@ CHARTS = {
     },
     'osd_usage': {
         'options': [None, 'Ceph OSDs', 'KiB', 'osd', 'ceph.osd_usage', 'line'],
+        'lines': []
+    },
+    'osd_size': {
+        'options': [None, 'Ceph OSDs size', 'KiB', 'osd', 'ceph.osd_size', 'line'],
         'lines': []
     },
     'osd_apply_latency': {
@@ -189,6 +194,9 @@ class Service(SimpleService):
             self.definitions['osd_usage']['lines'].append([osd['name'],
                                                            osd['name'],
                                                            'absolute'])
+            self.definitions['osd_size']['lines'].append([osd['name'],
+                                                           osd['name'],
+                                                           'absolute'])
             self.definitions['osd_apply_latency']['lines'].append(['apply_latency_{0}'.format(osd['name']),
                                                                    osd['name'],
                                                                    'absolute'])
@@ -217,6 +225,7 @@ class Service(SimpleService):
                 data.update(self._get_pool_rw(pool_io))
             for osd in osd_df['nodes']:
                 data.update(self._get_osd_usage(osd))
+                data.update(self._get_osd_size(osd))
             for osd_apply_commit in osd_perf_infos:
                 data.update(self._get_osd_latency(osd_apply_commit))
             return data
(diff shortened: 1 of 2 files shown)


ilyam8:
> But not collecting the osd_size prevents from computing used ratio (osd_usage / osd_size) or available space (osd_size - osd_usage), which are useful metrics.

> So it might not make sense to collect it

We can add them as variables to the `osd_usage` chart, what do you think?

elelayan:
> > But not collecting the osd_size prevents from computing used ratio (osd_usage / osd_size) or available space (osd_size - osd_usage), which are useful metrics.
> 
> > So it might not make sense to collect it
> 
> We can add them as [variables](https://docs.netdata.cloud/collectors/plugins.d/#variable) to the `osd_usage` chart, what do you think?

Thanks for the suggestion. I'm not familiar with VARIABLES in netdata:
 - would that be multiple VARIABLE 'osd_size.1' ... 'osd_size.n'?
 - how would a 'osd_size.1' VARIABLE relate to the 'osd_usage.1' value?
 - could you point me to a similar graph?

ilyam8:
I think having it as a separate chart is acceptable too. I recall we have some problems with vars during exporting data.

As for variables:

```python
...

self.definitions['pool_usage']['lines'].append([pool['name'], pool['name'], 'absolute'])
self.definitions['pool_usage']['variables'].append([pool['name'] + '_kb')
...
        return {osd['name'] + '_kb': float(osd['kb'])}
```

elelayan:
Thanks @ilyam8, I'll experiment a bit more and try variables.

My PR was broken because osd_usage and osd_size had the same key (osd.x) so they overwrote each other, so I've pushed a fix.

elelayan:
I've tried adding size variables to the osd_usage chart.

They would be useable in alarms (visible in `/api/v1/alarm_variables?chart=ceph_storage.osd_usage`).

But they seem unusable from prometheus :
 - I couldn't find them in prometheus query or grafana, even if they appear in the scrape output
 - in the scrape output they have no dimension so it wouldn't be easy to correlate them with usage, contrary to lines, where I could use the query (`netdata_ceph_osd_usage_KiB_average - ignoring(chart) netdata_ceph_osd_size_KiB_average`)
 - they would go stale after some time according to #6470

(output from `/api/v1/allmetrics?format=prometheus_all_hosts&variables=yes`)
variable:
```
netdata_ceph_storage_osd_usage_size_osd_2_kb{instance="server1"} 3906994176.0000000 1587569329000
```
vs line:
```
netdata_ceph_osd_usage_KiB_average{chart="ceph_storage.osd_usage",family="osd",dimension="osd.2",instance="server1"} 2614347000.0000000 1587570190000
```

So I would rather go with the size chart as is in my current PR.
@ilyam8 what do you think?
Thanks,

ilyam8:
> But they seem unusable from prometheus :

Yes, you are right.

ilyam8:
> So I would rather go with the size chart as is in my current PR.

👍

elelayan:
Cool, thanks :-)

ilyam8:
@elelayan we are about to release new version, this PR will be merged after it

elelayan:
Thanks!

PR #8655, opened 2020-04-08
Update apps_groups.conf

thenktor:
Added Squeezebox server (Logitech Media Server) and Squeezebox clients: https://en.wikipedia.org/wiki/Logitech_Media_Server
https://en.wikipedia.org/wiki/Squeezelite


ilyam8:
thx @thenktor

PR #8658, opened 2020-04-09
Update synology.md

thenktor:
Make this more readable with sub-headlines. Also add a note that /etc/rc.local may not exist.

### DIFF
--- a/packaging/installer/methods/synology.md
+++ b/packaging/installer/methods/synology.md
@@ -14,7 +14,7 @@ issue](https://github.com/SynoCommunity/spksrc/issues/2758), still open as of 20
 Chroot package is not suitable for DSM versions greater than version 5 and may corrupt system libraries and render the
 NAS unable to boot.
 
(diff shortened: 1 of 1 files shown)


packaging/installer/methods/synology.md
```diff
@@ -14,9 +14,11 @@ issue](https://github.com/SynoCommunity/spksrc/issues/2758), still open as of 20
 Chroot package is not suitable for DSM versions greater than version 5 and may corrupt system libraries and render the
 NAS unable to boot.
 
-The good news is that the 64-bit static installer works fine if your NAS is one that uses the amd64 architecture. It
+The good news is that the [64-bit static installer](https://docs.netdata.cloud/packaging/installer/methods/kickstart-64/) works fine if your NAS is one that uses the amd64 architecture. It
```
joelhans, line 17: Documentation links should be relative, although I'm surprised to see this works just fine! Try this instead:

```suggestion
The good news is that the [64-bit static installer](kickstart-64.md) works fine if your NAS is one that uses the amd64 architecture. It
```

PR #8666, opened 2020-04-09
Microsoft Teams

luckman212:
##### Summary
- Change `msteam` -> `msteams` (correct naming)
- Fix nonworking adaptive card designer URL
- Removed `slack` reference in Teams stanza

This is a minor change, since the official name of the product is **[Microsoft Teams](https://products.office.com/en-US/microsoft-teams/group-chat-software)** I thought the config should reflect that, otherwise it almost looks like a typo.

##### Component Name
- Alarms / config

### DIFF
--- a/health/notifications/alarm-notify.sh.in
+++ b/health/notifications/alarm-notify.sh.in
@@ -163,7 +163,7 @@ pd
 fleep
 syslog
 custom
-msteam
+msteams
 kavenegar
 prowl
 irc
@@ -295,7 +295,7 @@ done
 SLACK_WEBHOOK_URL=
 
 # Microsoft Team configs
-MSTEAM_WEBHOOK_URL=
+MSTEAMS_WEBHOOK_URL=
 
 # rocketchat configs
 ROCKETCHAT_WEBHOOK_URL=
@@ -541,7 +541,7 @@ if [ "${SEND_PUSHOVER}" = "YES" ] ||
 	[ "${SEND_PROWL}" = "YES" ] ||
 	[ "${SEND_HANGOUTS}" = "YES" ] ||
 	[ "${SEND_CUSTOM}" = "YES" ] ||
-	[ "${SEND_MSTEAM}" = "YES" ] ||
+	[ "${SEND_MSTEAMS}" = "YES" ] ||
 	[ "${SEND_DYNATRACE}" = "YES" ]; then
 	# if we need curl, check for the curl command
 	if [ -z "${curl}" ]; then
@@ -553,7 +553,7 @@ if [ "${SEND_PUSHOVER}" = "YES" ] ||
 		SEND_PUSHBULLET="NO"
 		SEND_TELEGRAM="NO"
 		SEND_SLACK="NO"
-		SEND_MSTEAM="NO"
+		SEND_MSTEAMS="NO"
 		SEND_ROCKETCHAT="NO"
 		SEND_ALERTA="NO"
 		SEND_PD="NO"
@@ -695,8 +695,8 @@ for method in "${SEND_EMAIL}" \
 	"${SEND_AWSSNS}" \
 	"${SEND_SYSLOG}" \
 	"${SEND_SMS}" \
-	"${SEND_MSTEAM}" \
-    "${SEND_DYNATRACE}"; do
+	"${SEND_MSTEAMS}" \
+	"${SEND_DYNATRACE}"; do
 	
 	if [ "${method}" == "YES" ]; then
 		proceed=1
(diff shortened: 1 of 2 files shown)


ilyam8:
@luckman212 i think this PR breaks compatibility. If someone uses this method if will stop working after the PR is merged.

thiagoftsm:
@ilyam8 is right, my suggestion is you keep the old name for the global configuartion, but you also give condition to use the correct names.
Inside the script, you can do the necessary fixes for wrong names.

cosmix:
I would suggest the following: add the new function name (`send_msteams()`), but also keep the old one around for compatibility (For a few more releases at least). The old function should just call the new one.

luckman212:
> i think this PR breaks compatibility. If someone uses this method if will stop working after the PR is merged.

I'm not too familiar, but I ran some grep-style searches with [ag](https://github.com/ggreer/the_silver_searcher) over the entire repo and didn't find any other references to these functions. How are they called externally? Duplicating the functions sounds messy.

thiagoftsm:
Hi @luckman212,

Right now Netdata is calling this script using `posix_spawn` to create new process.  We firstly mount a string with the script as first value, the next values are the information about the alarm raised.

I agree that function duplication is not good, but considering that we are talking about Microsoft, I will give suggestion, let us do like Microsoft always did, instead to duplicate the original function, you can create a new function with the correct name, and inside this function you call the old function.

Best regards!

luckman212:
> @thiagoftsm Right now Netdata is calling this script using `posix_spawn` to create new process

I understand that, but I think all the logic of which alert targets to fire is contained within the Bash script itself, right? The `posix_spawn` is passing some descriptive parameters about the alarm but not calling `send_msteam()` directly, right? So changing the function name should not break anything?

ilyam8:
Problem is changing names in the config files.

thiagoftsm:
> > @thiagoftsm Right now Netdata is calling this script using `posix_spawn` to create new process
> 
> I understand that, but I think all the logic of which alert targets to fire is contained within the Bash script itself, right? The `posix_spawn` is passing some descriptive parameters about the alarm but not calling `send_msteam()` directly, right? So changing the function name should not break anything?

You are right, the problem is to remove variable names like `MSTEAM_WEBHOOK_URL` and  `SEND_MSTEAM`.

luckman212:
> @thiagoftsm: You are right, the problem is to remove variable names like `MSTEAM_WEBHOOK_URL` and `SEND_MSTEAM`.

> @ilyam8: Problem is changing names in the config files.

I looked at the original PR that added the MSTeams integration (#3903) and I also searched the repo codebase for any other references to things like `MSTEAM_WEBHOOK_URL`. I'm sorry if I'm being thick here but I still don't understand how changing these variable names could cause a problem.

Ferroin:
> > @thiagoftsm: You are right, the problem is to remove variable names like `MSTEAM_WEBHOOK_URL` and `SEND_MSTEAM`.
> 
> > @ilyam8: Problem is changing names in the config files.
> 
> I looked at the original PR that added the MSTeams integration (#3903) and I also searched the repo codebase for any other references to things like `MSTEAM_WEBHOOK_URL`. I'm sorry if I'm being thick here but I still don't understand how changing these variable names could cause a problem.

Changing the names of existing variables used for configuration will break existing configurations. If they should be changed, you have to keep support for the old names as well so that existing configs continue to work.

thiagoftsm:
Hi @luckman212 ,

We are trying to avoid that users already using the old variable names need to change their `health_alarm_notify.conf` to continue using the Microsoft Teams.

ilyam8:
@luckman212 i think following does:
 - keep all changes you did
 - make mapping old=>new right after load config func

https://github.com/luckman212/netdata/blob/cd5f5b0dcb2849815a32ab7de51ea0a7a8abd990/health/notifications/alarm-notify.sh.in#L384-L398

ilyam8:
i have something like this in mind


```sh
msteams_migration() {
  SEND_MSTEAMS=${SEND_MSTEAM:-$SEND_MSTEAMS}
  ...
}

msteams_migration
```

luckman212:
@ilyam8 ok, this `msteams_migration()` function goes into `health/notifications/alarm-notify.sh.in` ?

ilyam8:
@luckman212 yes. The idea it to make old configuration working with new code.

tomcbe:
Are there plans to merge this once the pending issues have been resolved? I'm asking because I started working on adding the relevant documentation for MS Teams and I would then use the new variable names and add a remark, that the old ones still work for backward compatibility.

ilyam8:
@tomcbe there are some issues that need to be resolved

see https://github.com/netdata/netdata/pull/8666#issuecomment-612104328

https://github.com/netdata/netdata/pull/8666#issuecomment-612107633

+ _This branch has conflicts that must be resolved_

---

@tomcbe if you are willing you can finish the PR if @luckman212 has no time at the moment

ilyam8:
> Are there plans to merge this once the pending issues have been resolved?

Yes.

tomcbe:
> @tomcbe if you are willing you can finish the PR if @luckman212 has no time at the moment

Ok, should I open a separate PR to finish it?

ilyam8:
> Ok, should I open a separate PR to finish it?

Yes.

thiagoftsm:
Hi @luckman212 ,

Please, could you fix the conflicts on this PR for we move forward with reviews, tests and merge it?

Best regards!

ilyam8:
Closed in favor of #9905

PR #8669, opened 2020-04-09
Added statusimage chart library

### LINKED ISSUE #8265: Display states on custom dashboard
I use a custom collector that monitors a binary value (either 0 or 1) an I would like to display this state in understandable form (e.g. "Device: Enabled" / "Device: Disabled") on a custom dashboard.

As far as I understood badges and alarms might help to make this happen but I'm not sure how.
Is an alarm a way / the best way to achieve this? And how could I create an alarm that translates the collected value (0/1) to a predefined text ("Enabled"/"Disabled") to show it e.g. in a badge on the dashboard?

Thanks a lot!

lab-w:
##### Summary
This PR adds a chart library displaying an image depending on the data. If the data equals zero the image given by statusimage-false-image is displayed, otherwise the image given by statusimage-true-image.
Fixes #8265 by displaying a state (value equals zero or not) via an image.

##### Component Name
Chart libraries

##### Test Plan
Can be tested by adding these section to a custom dashboard:

    <div data-netdata="<data to analyze>"
                data-chart-library="statusimage"
                data-update-every="1"
                data-statusimage-true-image="true.png"
                data-statusimage-false-image="false.png"
                data-after="-300"
                data-points="300"
    ></div>

Two images (true.png and false.png) needs to be available for testing.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -130,5 +130,6 @@ This is the list of contributors that have signed this agreement:
 |@akwan|Alan Kwan||
 |@underhood|Timotej Šiškovič||
 |@stevenh|Steven Hartland|steven.hartland@multiplay.co.uk|
+|@lab-w|Norbert Werner|norbert.werner@lab-w.org|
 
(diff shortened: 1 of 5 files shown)


ilyam8:
Thanks for the contribution, @lab-w, and apologies for the **delayed** response.

A dashboard-related PR should be opened in https://github.com/netdata/dashboard.

PR #8671, opened 2020-04-10
Consistent k8s cgroup naming

### LINKED ISSUE #8044: Apply consistent k8s container naming when running inside and outside the cluster
> Hi @angelbarrera92 👋 Any chance you could check latest version? Probably it is fixed.

The user has problems monitoring k8s containers from **outside** the k8s cluster. 
I believe I found the solution, but not sure of one thing:

When running from outside the cluster, the command
`curl -sSk -H "Authorization: Bearer $KUBE_TOKEN"  "https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_PORT_443_TCP_PORT/api/v1/pods"`

needs to be replaced with 
`kubectl get pods --all-namespaces -o json`

The rest of the pipes remain the same. 

So... `k8s_get_name` needs to accept an argument, telling it which of the two commands to run (inside/outside the cluster). 

An additional check needs to be added, to execute `k8s_get_name outside` and the current call needs to be made  `k8s_get_name inside` (don't get stuck on naming, this is just to clarify).

The question is, what is that additional check? Is it that the name is not found (i.e. we add it to the very end of all other attempts?). That needs replication. You just need to access one of our test k8s clusters, to see what happens. I'm giving both @ilyam8 and @vlvkobal access to that, so you can test it out.

_Originally posted by @cakrit in https://github.com/netdata/netdata/issues/3369#issuecomment-584701422_

cakrit:
Fixes #8044

##### Summary

Ensures we get consistent kubernetes cgroup container names, regardless of whether netdata runs from within a kubernetes pod, or directly on the host machine.

##### Test Plan

@ilyam8 to test on our k8s test cluster and on a minikube environment.

##### Additional Information

Fixes a bug that was originally fixed in #6885 and re-introduced in #7416.
Fixes the seemingly non-functional selector in #7416

### DIFF
--- a/collectors/cgroups.plugin/cgroup-name.sh.in
+++ b/collectors/cgroups.plugin/cgroup-name.sh.in
@@ -102,7 +102,7 @@ function k8s_get_name() {
 			if kubectl --kubeconfig=$KUBE_CONFIG get pod --all-namespaces >/dev/null 2>&1; then
 				#shellcheck disable=SC2086
 				NAME="$(kubectl --kubeconfig=$KUBE_CONFIG get pod --all-namespaces --output='json' | 
-				jq -r '.items[] | "k8s_\(.metadata.namespace)_\(.metadata.name)_\(.metadata.uid)_" + (.status.containerStatuses[]? | "\(.name) \(.containerID)") | 
+				jq -r '.items[] | "k8s_\(.metadata.namespace)_\(.metadata.name)_\(.metadata.uid)_" + (.status.containerStatuses[]? | "\(.name) \(.containerID)")' | 
 				grep "$id" | 
 				cut -d' ' -f1 
 				)"
--- a/collectors/cgroups.plugin/cgroup-name.sh.in
+++ b/collectors/cgroups.plugin/cgroup-name.sh.in
@@ -101,10 +101,10 @@ function k8s_get_name() {
 			fi
 			if kubectl --kubeconfig=$KUBE_CONFIG get pod --all-namespaces >/dev/null 2>&1; then
 				#shellcheck disable=SC2086
-				NAME="$(kubectl --kubeconfig=$KUBE_CONFIG get pod --all-namespaces --output='json' | \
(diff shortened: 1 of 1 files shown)


cakrit:
@ilyam8 may close this and create a better solution, perhaps calling a function, instead of copying/pasting code

majodev:
Hi,

I've just updated the latest stable netdata `v1.22.1` in our GKE cluster (we are running netdata inside a privileged docker container within kubernetes). Had to make the following updates to get cgroup-name.sh to properly display pod + all container names:

ilyam8:
hi @majodev 

> inside a privileged docker container

> line 105

It is being used when netdata is installed on the host system, if it is inside a container this part is executed

https://github.com/netdata/netdata/blob/ec7be759eb60642c7cf0e161c8db63b8a3375176/collectors/cgroups.plugin/cgroup-name.sh.in#L90-L97


am i missing something?

majodev:
@ilyam8 it's not about if's actually used, https://github.com/netdata/netdata/pull/8671/files#diff-8cdc1f0cf955be0a3ece91904cdad2e7R104-R108 just has a syntax error and misses a `'` before the `|`.

from

```bash
jq -r '.items[] | "k8s_\(.metadata.namespace)_\(.metadata.name)_\(.metadata.uid)_" + (.status.containerStatuses[]? | "\(.name) \(.containerID)") | 
```

to

```bash
jq -r '.items[] | "k8s_\(.metadata.namespace)_\(.metadata.name)_\(.metadata.uid)_" + (.status.containerStatuses[]? | "\(.name) \(.containerID)")' | 
```

cakrit:
I just added that single quote, that's what you get when you use the GitHub UI to edit long lines :(

Can we please get approvals here so we can merge it?

prologic:
Please rebase this PR to resolve the CI failures.

cakrit:
Rebasing gives several conflicts (bad). Opened #9303 instead

PR #8683, opened 2020-04-11
libnetdata/log: fatal add missing space

ilyam8:
##### Summary

ssia

```cmd
2020-04-11 21:56:13: netdata INFO  : MAIN : Cannot open the file /var/lib/netdata/health.silencers.json, so Netdata will work with the default health configuration.
2020-04-11 21:56:13: netdata FATAL : MAIN :Cannot create directory '/var/lib/netdata/registry'. # : Invalid argument
```

##### Component Name

`libnetdata/log`

##### Test Plan

Not needed


mfundul:
Does this affect the parsing of analytics we have from `fatal()`?

ilyam8:
You mean GA? I think it is unlikely. But i dont know for sure.

ilyam8:
Fixed in https://github.com/netdata/netdata/pull/9382

PR #8686, opened 2020-04-12
Update pfSense doc and add warning for apcupsd users

cryptoluks:
##### Summary
* update to latest netdata package versions in pfSense installation documentation and add missing dependencies

* add warning when using apcupsd to prevent memory leak and high cpu usage

##### Additional Information

If netdata is started before `apcupsd` (happens when starting apcupsd automatically through checkbox in GUI and netdata through shellcmd mechanism) a `cat` process will eat up CPU and RAM. The `cat` process seems to be triggered by the default activated `apcupsd` netdata charts plugin (writes to `/tmp/.netdata-charts.d-foo/run.foo`). When disabling the netdata apcupsd plugin, the cat process is not spawned at boot.

But without disabling the netdata apcupsd plugin it can be prevented by ensuring apcupsd is started before netdata with for example the following as a shellcmd command:

`service apcupsd onestart; service netdata onestart`

### DIFF
--- a/packaging/installer/methods/pfsense.md
+++ b/packaging/installer/methods/pfsense.md
@@ -36,7 +36,7 @@ URL instead, or you can try manually changing the netdata version in the URL to 
 
 You must edit `/usr/local/etc/netdata/netdata.conf` and change `bind to = 127.0.0.1` to `bind to = 0.0.0.0`.
 
-To start Netdata manually, run `service netdata onestart`
+To start Netdata manually, run `service netdata onestart`.
 
 **Warning:** If you are using the plugin `apcupsd`, you need to make sure that apcupsd is up before starting netdata. Otherwise a infinitely running `cat` process triggerd by the default activated netdata apcuspd charts plugin will eat up CPU and RAM (`/tmp/.netdata-charts.d-*/run-*`). This also applies for `OPNsense`.
 
--- a/packaging/installer/methods/pfsense.md
+++ b/packaging/installer/methods/pfsense.md
@@ -38,7 +38,7 @@ You must edit `/usr/local/etc/netdata/netdata.conf` and change `bind to = 127.0.
 
 To start Netdata manually, run `service netdata onestart`.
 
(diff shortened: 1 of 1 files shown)


ilyam8:
> If netdata is started before apcupsd (happens when starting apcupsd automatically through checkbox in GUI and netdata through shellcmd mechanism) a cat process will eat up CPU and RAM.

Hi @cryptoluks it sound like the proble, i suggest to create an issue.

We need either fix bash module or rewrite it in other language.

ilyam8:
@cryptoluks i see it is pretty simple to rewrite it in go, all we need is to send  `"\x00\x06status"`.

https://github.com/flyte/apcaccess/blob/develop/apcaccess/status.py

packaging/installer/methods/pfsense.md
```diff
@@ -30,7 +36,9 @@ URL instead, or you can try manually changing the netdata version in the URL to
 
 You must edit `/usr/local/etc/netdata/netdata.conf` and change `bind to = 127.0.0.1` to `bind to = 0.0.0.0`.
 
-To start Netdata manually, run `service netdata onestart`  
+To start Netdata manually, run `service netdata onestart`
```
joelhans, line 39:
```suggestion
To start Netdata manually, run `service netdata onestart`.
```

packaging/installer/methods/pfsense.md
```diff
@@ -30,7 +36,9 @@ URL instead, or you can try manually changing the netdata version in the URL to
...
+
+**Warning:** If you are using the plugin `apcupsd`, you need to make sure that apcupsd is up before starting netdata. Otherwise a infinitely running `cat` process triggerd by the default activated netdata apcuspd charts plugin will eat up CPU and RAM (`/tmp/.netdata-charts.d-*/run-*`). This also applies for `OPNsense`.
```
joelhans, line 41:
```suggestion
**Warning:** If you are using the `apcupsd` collector, you need to make sure that apcupsd is up before starting Netdata. Otherwise a infinitely running `cat` process triggered by the default activated apcuspd charts plugin will eat up CPU and RAM (`/tmp/.netdata-charts.d-*/run-*`). This also applies to `OPNsense`.
```

PR #8699, opened 2020-04-13
Fixed Arch Linux Ci checks.

Ferroin:
##### Summary

This explicitly pulls in libffi during CI checks on Arch Linux.

The `make` command depends on it (becaus it uses GNU Guile, which depends on libffi) but does not pull it in as a dependency. This isn't likely to affect user systems in most cases as they will almost certainly have Python installed, which also depends on libffi and pulls it in correctly, but our CI systems don't install anything unnecessary, so they don't have it.

##### Component Name

area/ci

##### Test Plan

CI checks for Arch Linux pass on this PR.

##### Additional Information

This is a workaround for an upstream bug in Arch.

### DIFF
--- a/.github/workflows/build-and-install.yml
+++ b/.github/workflows/build-and-install.yml
@@ -42,7 +42,7 @@ jobs:
             pre: 'apk add -U bash'
 
           - distro: 'archlinux:latest'
-            pre: 'pacman --noconfirm -Sy grep'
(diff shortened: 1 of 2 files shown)


prologic:
> This is a workaround for an upstream bug in Arch.

Please reference the upstream issue either here or in a newly created issue to track. I find it a bit odd that GNU Make should depend on Guiles.

PR #8715, opened 2020-04-14
Fixed the build matrix in the build & install checks.

Ferroin:
##### Summary

This fixes two issues with the build matrix in the GitHub Actions workflow for verifying the build and install process:

* Debian 9 is missing the `apt-get update` command as part of the preparation of the test image (though this was for some odd reason working).
* There were still references to Ubuntu 19.04 even though we removed tests for it a while back, causing the tests to still run (and just recently start failing).

##### Component Name

area/ci

##### Test Plan

The relevant checks pass on this PR.


Ferroin:
Whoever makes the second approval can just merge this provided there are no comments objecting to it.

PR #8725, opened 2020-04-15
Fix crash when shutdown with ACLK disabled

lassebm:
### Summary

This PR fixes a crash (segfault, `pthread_cancel()` on non-existent thread) when Netdata is shutdown with ACLK disabled, currently the default.

##### Component Name

ACLK

##### Test Plan

Verify no crash is observed when Netdata is shutdown. Perform verification in official Docker container (Alpine Linux/musl).


amoss:
Nice fix. There are a couple of more cases in the same function. I'll create a PR against the branch in your fork and then we can fix them all at the same time.

PR #8740, opened 2020-04-16
Added missing path

Jiab77:
##### Summary

In the command line given to edit the file `charts.d.conf`, the path to the script `edit-config` is missing so if you type it or paste it that way it won't work. So I think the easiest way to fix it would be to add the path in the command line given in the documentation.

##### Component Name

Documentation

##### Test Plan

Run the command as given in the documentation and with the added path to see the difference.

##### Additional Information

The other solution would be add to the path `/etc/netdata` to the global `PATH` variable.

### DIFF
--- a/docs/netdata-for-IoT.md
+++ b/docs/netdata-for-IoT.md
@@ -56,7 +56,14 @@ supported by `lm-sensors`.
 Netdata also has a bash version of the sensors plugin that can read RPi temperatures. It is disabled by default to avoid
 the conflicts with the python version.
 
-To enable it, run `sudo /etc/netdata/edit-config charts.d.conf` and uncomment this line:
+To enable it, run:
+
+```bash
+cd /etc/netdata
+sudo ./edit-config charts.d.conf
+```
+
+and uncomment this line:
 
 ```sh
 sensors=force
--- a/docs/netdata-for-IoT.md
+++ b/docs/netdata-for-IoT.md
@@ -56,7 +56,7 @@ supported by `lm-sensors`.
 Netdata also has a bash version of the sensors plugin that can read RPi temperatures. It is disabled by default to avoid
 the conflicts with the python version.
 
-To enable it, run `sudo edit-config charts.d.conf` and uncomment this line:
+To enable it, run `sudo /etc/netdata/edit-config charts.d.conf` and uncomment this line:
 
 ```sh
 sensors=force
--- a/docs/netdata-for-IoT.md
+++ b/docs/netdata-for-IoT.md
@@ -59,7 +59,7 @@ the conflicts with the python version.
 To enable it, run:
 
 ```bash
-cd /etc/netdata
+cd /etc/netdata # Replace this path with your Netdata config directory
 sudo ./edit-config charts.d.conf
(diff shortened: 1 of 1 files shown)


Jiab77:
Is there any issues with my pull request? Did I missed something?

ilyam8:
@Jiab77 the path is depends on `--install <PATH>` option. We use `sudo ./edit-config go.d/vsphere.conf` everywhere, example

> https://github.com/netdata/go.d.plugin/tree/master/modules/vsphere#configuration

We need the part on the IOT page with

```sh
cd /etc/netdata # Replace this path with your Netdata config directory
sudo ./edit-config go.d/vsphere.conf
```

Jiab77:
@ilyam8 I agree with you but I never use the `--install <PATH>` option and in this case it never work and I'm forced to prefix the path to `./edit-conf` and use it that way `sudo /etc/netdata/edit-conf <CONFFILE>`.

Maybe this can be caused by my way of installing Netdata. Here is the command line I'm using everywhere to setup Netdata:

```bash
bash <(curl -Ss https://my-netdata.io/kickstart.sh) all --dont-wait
```

> We need the part on the IOT page with
> ```bash
> cd /etc/netdata # Replace this path with your Netdata config directory
> sudo ./edit-config <CONFFILE>
> ```

Yes in this case if you can do the same of the [vsphere](https://github.com/netdata/go.d.plugin/tree/master/modules/vsphere#configuration) page to the [IOT](https://docs.netdata.cloud/docs/netdata-for-iot/) page, it should fix the issue.

ilyam8:
> Yes in this case if you can do the same of the vsphere  page to the IOT page, it should fix the issue.

Please do then.

Jiab77:
@ilyam8 How so? Just by changing my pull request?

ilyam8:
Yeah, add the changes in another commit.

Jiab77:
@ilyam8 I just did, let me know if this one is okay for you.

Jiab77:
@ilyam8 Sorry, the latest change should be the good one.

Jiab77:
Thanks a lot! @ilyam8

ilyam8:
@Jiab77 what is wrong with status feature btw? 😄

Jiab77:
@ilyam8 I just find it useless... It's GitHub not FB :laughing:

Jiab77:
@joelhans You're welcome!

It's always a pleasure for me to improve this project :+1: 

I haven't checked the code of `edit-config` but does it needs to be placed in the netdata config folder or it could be placed somewhere else like `/usr/bin` or `/usr/local/bin` or even `/bin` or `/sbin`.

Normally these paths are already added in the main `PATH` environment variable and in this case you could call `edit-config` from anywhere without any issues.

Jiab77:
btw I was just thinking that the `charts.d.conf` does not contains any lines that needs to be un-commented. I've installed netdata on 5 IoT devices with the `bash` method and I always had to add the line `sensors=force` instead of removing the `#` character.

should I open another issue for that?

ilyam8:
yeah, we need to update `charts.d.conf`

PR #8743, opened 2020-04-17
Ebpf index size

### LINKED ISSUE #8742: Index length
##### Feature idea summary
In a overload Operate System the eBPF collector needs to reset the index before the end of the day, it is necessary to increase the variable size to avoid this problem.

##### Expected behavior
Brings to the collector conditions to read the variables changed on `kernel-collector`.

thiagoftsm:
##### Summary
Fixes #8742 

This PR fixes the problem reported on #8742 and https://github.com/netdata/kernel-collector/issues/133.
##### Component Name
Collector

##### Test Plan
1 - Create a shell script that has an infinite loop to write and remove information from disk
2 - Keep the collector running.

##### Additional Information
This is an improvement from the previous version, it is not a bug fix.

### DIFF
--- a/collectors/ebpf_process.plugin/ebpf_process.c
+++ b/collectors/ebpf_process.plugin/ebpf_process.c
@@ -512,7 +512,7 @@ void *process_publisher(void *ptr)
 
 static void move_from_kernel2user_global() {
     uint64_t idx;
-    uint64_t res[NETDATA_GLOBAL_VECTOR];
+    netdata_idx_t res[NETDATA_GLOBAL_VECTOR];
 
     netdata_idx_t *val = hash_values;
     for (idx = 0; idx < NETDATA_GLOBAL_VECTOR; idx++) {
--- a/collectors/ebpf_process.plugin/ebpf_process.c
+++ b/collectors/ebpf_process.plugin/ebpf_process.c
@@ -529,26 +529,26 @@ static void move_from_kernel2user_global() {
         }
     }
 
-    aggregated_data[0].call = res[0]; //open
-    aggregated_data[1].call = res[14]; //close
-    aggregated_data[2].call = res[8]; //unlink
-    aggregated_data[3].call = res[5] + res[21]; //read + readv
-    aggregated_data[4].call = res[2] + res[18]; //write + writev
-    aggregated_data[5].call = res[10]; //exit
-    aggregated_data[6].call = res[11]; //release
-    aggregated_data[7].call = res[12]; //fork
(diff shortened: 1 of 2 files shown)


collectors/ebpf_process.plugin/ebpf_process.c
```diff
@@ -529,26 +529,26 @@ static void move_from_kernel2user_global() {
...
-    aggregated_data[0].call = res[0]; //open
-    aggregated_data[1].call = res[14]; //close
-    aggregated_data[2].call = res[8]; //unlink
-    aggregated_data[3].call = res[5] + res[21]; //read + readv
-    aggregated_data[4].call = res[2] + res[18]; //write + writev
-    aggregated_data[5].call = res[10]; //exit
-    aggregated_data[6].call = res[11]; //release
-    aggregated_data[7].call = res[12]; //fork
-    aggregated_data[8].call = res[16]; //thread
+    aggregated_data[0].call = res[NETDATA_KEY_CALLS_DO_SYS_OPEN]; //open
```
underhood, line 532: I think we can kick out those comments now as we use Defines not magic numbers anymore.

collectors/ebpf_process.plugin/ebpf_process.c
```diff
@@ -511,13 +511,13 @@ void *process_publisher(void *ptr)
 }
 
 static void move_from_kernel2user_global() {
-    uint32_t idx;
-    uint32_t res[NETDATA_GLOBAL_VECTOR];
+    uint64_t idx;
+    uint64_t res[NETDATA_GLOBAL_VECTOR];
```
underhood, line 515: makes sense to typedef it in common hdr?

thiagoftsm: I am not sure whether I understood your question, I arrived in the conclusion that you are talking about `NETDATA_GLOBAL_VECTOR` definition. This is also a static value inside the `eBPF program` this is the motive I used the definition here and inside the eBPF code.  Case we bring more kprobes one day, it will need to change both sides.

collectors/ebpf_process.plugin/ebpf_process.c
thiagoftsm: I removed the comments, I liked your suggestion. Thank you!

collectors/ebpf_process.plugin/ebpf_process.c
thiagoftsm: After we talk on Slack I understood better what you were proposing and I brought two commits to convert the code in more readable version.

PR #8747, opened 2020-04-17
packaging/docker/{build,publish}.sh: Simplify scripts. Support only single ARCH

knatsakis:
##### Summary
Simplify the docker packaging build and publish shell scripts. Support building and packaging only one ARCH at a time.

### DIFF
--- a/packaging/docker/build.sh
+++ b/packaging/docker/build.sh
@@ -62,11 +62,11 @@ docker run --rm --privileged multiarch/qemu-user-static:register --reset
 # Build images using multi-arch Dockerfile.
 TAG="${REPOSITORY,,}:${VERSION}-${ARCH}"
 echo "Building tag ${TAG}.."
-eval docker build --no-cache \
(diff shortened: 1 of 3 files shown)


packaging/docker/build.sh
```diff
@@ -51,22 +53,20 @@ if [ ! -z $CWD ] || [ ! "${TOP_LEVEL}" == "netdata" ]; then
...
-     eval docker build --no-cache \
-          --build-arg ARCH="${ARCH}" \
-          --build-arg RELEASE_CHANNEL="${RELEASE_CHANNEL}" \
-          --tag "${TAG}" \
-          --file packaging/docker/Dockerfile ./
-     echo "..Done!"
-done
+TAG="${REPOSITORY,,}:${VERSION}-${ARCH}"
+echo "Building tag ${TAG}.."
+eval docker build --no-cache \
```
prologic, line 65: Why do we need an `eval` here?

knatsakis: We don't

PR #8750, opened 2020-04-17
Additional cases for the thread exit fix

amoss:
##### Summary
Follow-up to 8725.

##### Component Name
aclk

##### Additional Information
There were some additional thread exits in the ACLK that had the same problem as fixed in the previous PR. This adds fixes to those cases.

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -1277,11 +1277,11 @@ static void aclk_try_to_connect(char *hostname, char *port, int port_num)
  */
 void *aclk_main(void *ptr)
 {
+    struct netdata_static_thread *static_thread = (struct netdata_static_thread *)ptr;
     struct netdata_static_thread *query_thread;
 
     if (!netdata_cloud_setting) {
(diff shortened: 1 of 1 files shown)


aclk/agent_cloud_link.c
```diff
@@ -1391,6 +1392,7 @@ void *aclk_main(void *ptr)
         RSA_free(aclk_private_key);
 
     aclk_main_cleanup(ptr);
+    static_thread->enabled = NETDATA_MAIN_THREAD_EXITED;
```
stelfrag, line 1395: The `aclk_main_cleanup` will set this before returning (`NETDATA_MAIN_THREAD_EXITING` and then `NETDATA_MAIN_THREAD_EXITED`) so it can be skipped

underhood: I would probably remove it from `aclk_main_cleanup` actually, and put `goto cleanup` for repeating 2 lines in code:
```
static_thread->enabled = NETDATA_MAIN_THREAD_EXITED;
    return NULL;
```

amoss: True, will remove.

PR #8752, opened 2020-04-17
Use a prefix for the old dashboard.

### LINKED ISSUE #8662: Update React dashboard handling to not require a switching script.
Based on recent discussions, we're ditching the `netdata-switch-dashboard.sh` script in favor of just installing the old dashboard prefixed under the new one.

This will require changes to the installer code to set up the prefixing correctly, and will probably need changes to the old dashboard code as well.

Ferroin:
##### Summary

This switches the code for handling of the new React dashboard to use a prefix for accessing the old dashboard instead of relying on a switching script to handle selecting the desired dashboard implementation.

It also adds proper support to the packaging code to bundle the React dashboard.

##### Component Name

area/web
area/packaging

##### Test Plan

I've done limited local testing to verify that the build process still works correctly for everything.

I still need to verify that the old dashboard works correctly when moved under a prefix like this.

##### Additional Information

Fixes #8662

This has the following limitations:

* If the download of the React dashboard fails, we won't fall back to the regular dashboard.
* External packages (Gentoo, Homebrew, FreeBSD Ports, etc) will have a broken dashboard until they update to use the React dashboard.
* Under some circumstances, we unfortunately have to remove existing custom dashboards during updates to installs that were done using the kickstart script.

Actually avoiding these issues would require about 5-10 times as much code, still have issues, and have a lot more potential for bugs because it would be much more complicated.


thiagoftsm:
After to install this PR, I observed that I only had inside `/usr/share/netdata/` the directory `web`, is this expected? 

An additional information, I think there is something broken, because when I ran the script I am getting the following outputs:

```
bash-5.0# sh /usr/libexec/netdata/netdata-switch-dashboard.sh react
Switching to react dashboard...
cat: /usr/share/netdata/web-classic/.files: No such file or directory
cat: /usr/share/netdata/web-react/.files: No such file or directory
cp: cannot stat '/usr/share/netdata/web-react/.': No such file or directory

bash-5.0# sh /usr/libexec/netdata/netdata-switch-dashboard.sh classic
Switching to classic dashboard...
cat: /usr/share/netdata/web-classic/.files: No such file or directory
cat: /usr/share/netdata/web-react/.files: No such file or directory
cp: cannot stat '/usr/share/netdata/web-classic/.': No such file or directory
```

prologic:
I don' t know what's going on here, and I haven't checked this out locally to test (yet). is this ready @Ferroin or is @thiagoftsm 's report a problem? I can't for the life of my spot where the bug might be in the scripts in this PR.

thiagoftsm:
> I don' t know what's going on here, and I haven't checked this out locally to test (yet). is this ready @Ferroin or is @thiagoftsm 's report a problem? I can't for the life of my spot where the bug might be in the scripts in this PR.

It was decided that we need to change the way we are selecting between dashboards, but the current changes removed completely the `classic` dashboard, during my tests I only had the react dashboard.

prologic:
> It was decided that we need to change the way we are selecting between dashboards, but the current changes removed completely the `classic` dashboard, during my tests I only had the react dashboard.

Yeah this is the part I don't understand where the bug is. I looked really hard :)

prologic:
@thiagoftsm I checked out this PR  locally and observed the behaviour here is correct.; there should only be `web` in `/usr/share/netdata` now after this PR :)

thiagoftsm:
I think I misunderstood, because I could not select more between `react` and `old` dashboard.

Ferroin:
With the most recent set of changes, none of the limitations mentioned in the original description should be present here.

I'm pretty sure that I got all the references updated properly for `old/index.html`, but I may have missed some. If @jacekkolasa or someone else who knows this code better than i do could check, that would be greatly appreciated.

jacekkolasa:
@Ferroin unfortunately few things are not working in this case:
1) if you're in subroute, for example `localhost/netdata`, then `localhost/netdata/old` will not work, `localhost/netdata/old/` will be needed. (that's because browser doesn't know `old` is a directory and `src=../dashboard.js` points to directly to `localhost`
2) streamed agents will need some work on front-end side (and in old main.js) - that's because links to `slave hosts` are now created like: `http://localhost/netdata/old/host/netdata1/#`, and it should be: `http://localhost/netdata/host/netdata1/old/#`... fixing this would also need to hardcode that `old` suffix directly to main.js which doesn't look very clean...

We'll sync tomorrow on that

thiagoftsm:
When I click on `export snapshot` I am receiving this error:

err11

After to click in OK, it opens the window, and when I tried to click on `Export`, I receive another error:

err12

yesbox:
When I open the alarm panel I get several errors, like "Cannot load required JS library: lib/clipboard-polyfill-be05dad.js" and several others.

I can tell the browser is trying to access http://myserver/lib/clipboard-polyfill-be05dad.js, but the URL to netdata is http://myserver/netdata/, because it's running behind an nginx subfolder configuration according to this instruction and had been doing so without previous issues for years.

https://github.com/netdata/netdata/blob/master/docs/Running-behind-nginx.md#as-a-subfolder-to-an-existing-virtual-host

I believe this commit broke virtual server configurations and this is the same issue @thiagoftsm reported.

ilyam8:
@yesbox if there is an issue in the latest i suggest to open an issue

jacekkolasa:
@yesbox i think it's a problem with `net::ERR_HTTP2_PROTOCOL_ERROR` or `net::ERR_CONNECTION_REFUSED`, if you go to network tab you should see that calls to those files return 200, but they actually are errors :( It's better visible on Chrome, Firefox shows just 200 OK but there is clearly some server error.

Let's continue in that issue
https://github.com/netdata/netdata/issues/9154#issuecomment-634251671

Issue #8486, opened 2020-03-25
Claiming improvements


cakrit:
-[ ] The [command-line parameters](https://github.com/netdata/netdata/tree/master/claim#netdata-agent-command-line) should be the suggested way to do claiming in docker. Add the info to claim/README.md.

@rduffield and @manos-saratsis we need a decision on the following. Both will work, the second is just a bit more awkward.
- [ ]  Option 1: At the end of the claiming script, and based on a command line param, we try to run the command  `netdatacli reload-claiming-state`. By default it is executed, if netdata calls it with a certain param, it is not executed.
  - If it fails, we just say that "the netdata agent will be connected to the cloud when it is started again", or something to that effect.
  - If it succeeds, we say "the netdata agent is connecting to the cloud". It would be awesome if in the future the reload-claiming-state actually waited for the establishment of the ACLK and returned any errors, so that the script could inform the user, but that can wait.
- [ ] Option 2: We modify the command we show on the cloud UI and add `;netdatacli reload-claiming-state` to the end of it (making an even longer command).

manos-saratsis:
@cakrit option1 is great

PR #8755, opened 2020-04-17
Docs: Add Docker instructions to claiming

joelhans:
##### Summary

Fixes netdata/marketing#223

I've been getting requests to add instructions on how to claim a Docker-ized Agent, and this PR adds my best effort to document this process.

There seems to be a preference to tell users how to pass daemon command line options (see #8486), but I haven't been able to figure out how to successfully pass those. Using `docker exec -it` plus the claiming script straight from Cloud seems to work just fine, on the other hand, and seems much easier for users. Using `docker run` means they have to stop their container first.

If someone has successfully claimed a Docker node using a command like `docker run -it netdata --entrypoint '/usr/sbin/netdata -D -W "claim url=...."'`, and believes that is indeed the preferred method of claiming in this case, I'm happy to add it, but I have not been able to get that to work.

##### Component Name

claim

### DIFF
--- a/claim/README.md
+++ b/claim/README.md
@@ -61,7 +61,56 @@ Hit **Enter**. The script should return `Agent was successfully claimed.`.
 
 If the claiming script returns errors, see the [troubleshooting information](#troubleshooting).
 
-### Claiming through a proxy
+### Claim an Agent running in Docker
+
+First, stop the container, replacing `netdata` with the name of your Agent container.
+
+```bash
+docker stop netdata
+```
+
+Then, start the container again with the `docker run` command, passing 
+
+```bash
+docker run -it netdata/netdata /usr/sbin/netdata -D -W "claim -token=BSH_u2gRGweLVttLOL4k00zhgedGWXdBObc_X1-7G9iBKJJUFFVq0MWwCg-gFIbm_d8_eMKmoY-HXsDVj0vh2qQZZ6gJ6T9jc60nU9QVO2fgPPfwNh-p3TeemCdPMtEQk6ZiKps -rooms=a380abef-2244-40e7-a193-ff880c4604ed -url=https://staging.netdata.cloud"
+```
+
+
+
+docker run -it netdata/netdata --entry-point '/usr/sbin/netdata -D -W "claim -token=BSH_u2gRGweLVttLOL4k00zhgedGWXdBObc_X1-7G9iBKJJUFFVq0MWwCg-gFIbm_d8_eMKmoY-HXsDVj0vh2qQZZ6gJ6T9jc60nU9QVO2fgPPfwNh-p3TeemCdPMtEQk6ZiKps -rooms=a380abef-2244-40e7-a193-ff880c4604ed -url=https://staging.netdata.cloud"'
+
+
+
+
+
+
+
+
+
+
+
+
+
+```bash
+docker run -d -it --name=netdata \
+  -p 20000:19999 \
+  -v /etc/passwd:/host/etc/passwd:ro \
+  -v /etc/group:/host/etc/group:ro \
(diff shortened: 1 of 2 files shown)


amoss:
You can claim a node in a container like this:
```
docker run -it netdata /usr/sbin/netdata -D -W set global "netdata cloud" enable -W set cloud "cloud base url" "https://staging.netdata.cloud" -W "claim -token=... -rooms=... -url=https://staging.netdata.cloud"
```

joelhans:
Thanks, @amoss! The syntax of that command finally makes some sense to me now.

I'm guessing that after 1.22 goes out they won't have to do the `set global...` part.

prologic:
> You can claim a node in a container like this:
> 
> ```
> docker run -it netdata /usr/sbin/netdata -D -W set global "netdata cloud" enable -W set cloud "cloud base url" "https://staging.netdata.cloud" -W "claim -token=... -rooms=... -url=https://staging.netdata.cloud"
> ```

Don't you want `docker exec` here? `run` will run a new container, claim that new instance then shut down?

amoss:
No, this is not running the claiming script separately from the agent. If you do that then it can be exec’d directly as you say. This starts the agent, tells it to run the claiming script and then continue in a claimed state. This would be used on ephemeral nodes.

cakrit:
When this is finalized, please state also what we'll do for k8s and especially our helm chart.

joelhans:
@amoss @cakrit I added some more text about ephemeral nodes and made two sections for the different claiming processes.

joelhans:
@cakrit I created a new issue to document the k8s side: #8779

I have no k8s experience, so I'll need help on creating some docs, either from you or someone who knows k8s.

claim/README.md
```diff
@@ -54,14 +54,58 @@ With `sudo`:
...
+The claiming process works with Agents running inside of Docker containers. You can use `docker exec` to run the
+claiming script on containers already running, or append the claiming script to `docker run` to create a new container
+and immediately claim it.
+
+#### Running Agent containers
+
+Claim a _running Agent container_ by appending the script offered by Cloud to a `docker exec ...` command:
+
+```bash
+docker exec -it netdata-claim.sh -token=TOKEN -rooms=ROOM1,ROOM2 -url=https://app.netdata.cloud
```
prologic, line 74: This is missing either a "Container ID" or "Container Name" (_if Docker was run with `docker run --name ...`_) as the first argument.  See:

```#!sh
$ docker exec --help

Usage:	docker exec [OPTIONS] CONTAINER COMMAND [ARG...]
```

claim/README.md
```diff
@@ -54,14 +54,58 @@ With `sudo`:
...
+  --cap-add SYS_PTRACE \
+  --security-opt apparmor=unconfined \
+  netdata/netdata \
+  /usr/sbin/netdata -D -W set global "netdata cloud" enable -W set cloud "cloud base url" "https://app.netdata.cloud" -W "claim -token=TOKEN -rooms=ROOM1,ROOM2 -url=https://app.netdata.cloud"
+```
+
+The container runs in detached mode, so you won't see any output. If the node does not appear in your Space, you can run the following to find any error output and use that to guide your [troubleshooting](#troubleshooting):
+
+```bash
+docker logs netdata 2>&1 | grep -E 'ACLK|claim|cloud'
```
prologic, line 105: You should add `--line-buffered` here to `grep` because the log stream from `docker logs` is line buffered.

claim/README.md
joelhans: Good catch. I'd missed that. Added a `netdata` name to the command and referenced it above.

claim/README.md
joelhans: Done.

PR #8756, opened 2020-04-17
Redirect traffic from docs.netdata.cloud -> learn.netdata.cloud

joelhans:
##### Summary

This PR adds 301 redirects to point pages on `docs.netdata.cloud` to `learn.netdata.cloud`. 301 redirects will ensure any deep links pointing to the old documentation site will redirect without interruption, and hopefully without any SEO de-value.

This PR does _not_ redirect the translations. I want to maintain them as-is for now, as we don't yet have a plan on the Learn roadmap to replace them.

My only hesitation to deploy this change is that I would prefer if the Learn site's endpoint was `netdata.cloud/learn`. @prologic has a ticket (netdata/internal#24) for this, but he's a busy guy and I'm not sure about the timeframe for getting CF running and figuring out the routing. If we _can_ get that endpoint enabled, I'll just change all the redirects before we merge this and begin redirecting traffic.

#### Caveats/bugs

- By making this change, we lose deploy previews. I'm not sure how heavily engineers use them. I generally do not.

##### Test Plan

I tested every single link. Yes.

You can verify it yourself by opening the Netlify deploy. That should take you directly to `learn.netdata.cloud`. You can then try any deep link, like [the installation guide](https://deploy-preview-8756--netdata.netlify.app/packaging/installer/) to get redirected.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -12,12 +12,12 @@
   command = "./buildhtml.sh"
 
 [[redirects]]
-  from = /docs/*
-  to   = https://learn.netdata.cloud/docs/agent/:splat
+  from = "/docs/*"
+  to   = "https://learn.netdata.cloud/docs/agent/:splat"
 
 [[redirects]]
-  from = /*
-  to   = https://learn.netdata.cloud/docs/agent/:splat
+  from = "/*"
+  to   = "https://learn.netdata.cloud/docs/agent/:splat"
 
 [[redirects]]
   from = "/docs/GettingStarted/"
--- a/netlify.toml
+++ b/netlify.toml
@@ -12,6 +12,11 @@
   command = "./buildhtml.sh"
 
 [[redirects]]
+  from  = "/"
+  to    = "https://learn.netdata.cloud/docs/agent/"
+  force = true
+
+[[redirects]]
   from  = "/docs/*"
   to    = "https://learn.netdata.cloud/docs/agent/:splat"
   force = true
--- a/netlify.toml
+++ b/netlify.toml
@@ -17,14 +17,39 @@
   force = true
 
 [[redirects]]
-  from  = "/*"
+  from  = "/aclk/*"
+  to    = "https://learn.netdata.cloud/docs/agent/aclk/:splat"
+  force = true
+
+[[redirects]]
+  from  = "/build_external/*"
+  to    = "https://learn.netdata.cloud/docs/agent/build_external/:splat"
+  force = true
+
+[[redirects]]
+  from  = "/claim/*"
+  to    = "https://learn.netdata.cloud/docs/agent/claim/:splat"
+  force = true
+
+[[redirects]]
+  from  = "/cli/*"
+  to    = "https://learn.netdata.cloud/docs/agent/:splat"
+  force = true
+
+[[redirects]]
+  from  = "/collectors/*"
+  to    = "https://learn.netdata.cloud/docs/agent/:splat"
+  force = true
+
+[[redirects]]
+  from  = "/packaging/*"
   to    = "https://learn.netdata.cloud/docs/agent/:splat"
   force = true
 
 [[redirects]]
   from = "/docs/GettingStarted/"
-  to = "/docs/getting-started"
+  to = "https://learn.netdata.cloud/docs/agent/getting-started"
 
 [[redirects]]
   from = "/docs/tutorials/dimension-templates/"
-  to = "/health/tutorials/dimension-templates/"
+  to = "https://learn.netdata.cloud/docs/agent/health/tutorials/dimension-templates/"
--- a/netlify.toml
+++ b/netlify.toml
@@ -12,6 +12,14 @@
   command = "./buildhtml.sh"
 
 [[redirects]]
+  from = /docs/*
+  to   = https://learn.netdata.cloud/docs/agent/:splat
+
+[[redirects]]
+  from = /*
+  to   = https://learn.netdata.cloud/docs/agent/:splat
+
+[[redirects]]
   from = "/docs/GettingStarted/"
   to = "/docs/getting-started"
 
--- a/netlify.toml
+++ b/netlify.toml
@@ -22,6 +22,11 @@
   force = true
 
 [[redirects]]
+  from  = "/backends/*"
+  to    = "https://learn.netdata.cloud/docs/agent/backends/:splat"
+  force = true
+
+[[redirects]]
   from  = "/build_external/*"
   to    = "https://learn.netdata.cloud/docs/agent/build_external/:splat"
   force = true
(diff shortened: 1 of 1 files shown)


aabatangle:
I'm fine with the 301 (permanent) redirect strategy. I'm also OK with the learn subdomain for now. Does it make sense to maintain translations versus relying on browser translations? What is the strategy for this long-term?

joelhans:
> I'm fine with the 301 (permanent) redirect strategy. I'm also OK with the learn subdomain for now. Does it make sense to maintain translations versus relying on browser translations? What is the strategy for this long-term?

I totally forgot about browser translations! :open_mouth:

The original plan was to utilize CrowdIn to provide both ML- and community-translated localization. The problem is that Docusaurus doesn't yet support localization, which means we either hack up our own solution for now or just wait until their team implements the feature again (it's on their roadmap).

Maybe, if we axe the translations in this change, @zack-shoylev can engage the community to gauge how badly it wants "official" localization in the form of something like CrowdIn or the like, and we use that to prioritize the effort to put behind implementing it.

aabatangle:
@joelhans Makes sense. Unless we support official language packs for the product, not sure it makes sense for the docs. We used [Pootle](https://en.wikipedia.org/wiki/Pootle) for community-contributed language packs for the product at Untangle and relied on browser translation for site/docs.

zack-shoylev:
@joelhans What is missing here (I didn't have time to look too much): https://docusaurus.io/docs/en/translation, can we provide a crowdin project for it?

prologic:
> My only hesitation to deploy this change is that I would prefer if the Learn site's endpoint was `netdata.cloud/learn`. @prologic has a ticket ([netdata/internal#24](https://github.com/netdata/internal/issues/24)) for this, but he's a busy guy and I'm not sure about the timeframe for getting CF running and figuring out the routing. If we _can_ get that endpoint enabled, I'll just change all the redirects before we merge this and begin redirecting traffic.

Still working on getting access to our CloudFlare account. One thing I should note here is taht by redirecting learn.netdata.cloud to netdata.cloud/learn it won't be transparent to users unless we put netdata.cloud/learn behind a L7 load balancer. Are we okay with this?

prologic:
Also question; Since these are 301 Permanent Redirect(s) -- Does this mean we plan to deprecate and eventually remove the `docs.` sub-domain ?

cakrit:
@prologic 
- I see you're an admin in cloudflare now
- I don't see docs.netdata.cloud going away, unless there's 0 traffic to it for months. We have links to resources there all over the place.

@joelhans @zack-shoylev 
- We agreed with @aabatangle that we can forget about translations for the time being. Some people will be disappointed, but not too many, Google translate is good enough in many cases.

@aabatangle 
- Localization of the application itself isn't necessarily tied to documentation of basic things in other languages. We gave the community one way to translate documentation, we're consciously taking that away and do need to say something about it, as a company (at least in the https://github.com/netdata/localization project).

cakrit:
One thing unrelated to the other comments, I'm not sure how to test that this works/will work for all URLs.

prologic:
> * I see you're an admin in cloudflare now

Yes thanks to Costa :)

> * I don't see docs.netdata.cloud going away, unless there's 0 traffic to it for months. We have links to resources there all over the place.

Okay. that's fine... No problems :) But then I don't understand why we're placing 301 Moved Permanently on these resources? Wouldn't it be better to transparently proxy them with a L7 proxy so its transparent to the user and _who knwos_ maybw e change our minds in future? :)

joelhans:
> @joelhans What is missing here (I didn't have time to look too much): https://docusaurus.io/docs/en/translation, can we provide a crowdin project for it?

@zack-shoylev We're using the [alpha v2](https://v2.docusaurus.io/) of Docusaurus, which doesn't support translations yet.

joelhans:
> Okay. that's fine... No problems :) But then I don't understand why we're placing 301 Moved Permanently on these resources? Wouldn't it be better to transparently proxy them with a L7 proxy so its transparent to the user and _who knwos_ maybw e change our minds in future? :)

Well, one reason is that I have no idea what an L7 proxy is!

Another reason is that the URLs are not 1:1, so I'm using these 301s with some "splats" to make sure the paths are correct. For example:

```
docs.netdata.cloud/docs/getting-started/
learn.netdata.cloud/docs/agent/getting-started/
```

I'm not sure if this is possible with an L7 proxy.

joelhans:
> I don't see docs.netdata.cloud going away, unless there's 0 traffic to it for months. We have links to resources there all over the place.

These 301 redirects will mean no traffic arrives on `docs.netdata.cloud`, but it will properly redirect any outside links to the corresponding page on `learn.netdata.cloud`.

> One thing unrelated to the other comments, I'm not sure how to test that this works/will work for all URLs.

I have manually tested every page present in the sidebar. We can also use Ahrefs to monitor both sites for any 404s or other errors caused by incorrect redirects and fix them accordingly.

Ferroin:
> > Okay. that's fine... No problems :) But then I don't understand why we're placing 301 Moved Permanently on these resources? Wouldn't it be better to transparently proxy them with a L7 proxy so its transparent to the user and _who knwos_ maybw e change our minds in future? :)
> 
> Well, one reason is that I have no idea what an L7 proxy is!
> 
> Another reason is that the URLs are not 1:1, so I'm using these 301s with some "splats" to make sure the paths are correct. For example:
> 
> ```
> docs.netdata.cloud/docs/getting-started/
> learn.netdata.cloud/docs/agent/getting-started/
> ```
> 
> I'm not sure if this is possible with an L7 proxy.

'L7' in this case is short for layer 7.  In effect, an application layer proxy. Nginx is a good example, and URL rewrites like this are absolutely possible and part of why you might choose to use a layer 7 proxy in the first place.

_However_, unless we're looking at reusing `docs.netdata.cloud` for some other purpose, I actually agree that using HTTP redirects here makes sense, with two caveats:

* It should probably be a 308, not a 301. I don't _think_ this actually matters, but 301 allows the user agent to change the request method, while 308 does not (so sumitting forms doesn't work right with 301's).
* We should probably keep `docs.netdata.cloud` around as a redirect, because the _name_ fits with one of the very common patterns used by a lot of projects for documentation sites (I'd actually question why we're not just moving  things in-place on this domain instead of creating a new one for this same reason).

joelhans:
> However, unless we're looking at reusing docs.netdata.cloud for some other purpose, I actually agree that using HTTP redirects here makes sense, with two caveats:

We have no plan to reuse docs.netdata.cloud for any other purpose. The reason we're moving away from docs.netdata.cloud is that the experience I want to build is far more than just documentation. I don't want a subdomain to limit the type of work I can publish.

joelhans:
@prologic @Ferroin I've added you two as reviewers so you can help me manage this from an SRE side. I don't want to make changes that make your lives harder, but some kind of redirection/L7/proxy really needs to happen this week, preferably by Wednesday. I'd like some time to test and track any 404s that might pop up because of this.

prologic:
> I'm not sure if this is possible with an L7 proxy.

It actually is :) And btw a L7 Proxy is just a fancy way of saying "Reverse Proxy". We can do both proxy + rewrite if necessary. But I see your point now about the differing paths.

joelhans:
Closing, as @prologic handled this in CloudFlare instead.

prologic:
I don't think you need this PR anymore :)

PR #8760, opened 2020-04-18
TTL headers

amoss:
##### Summary
Add the http headers (including TTL info) to the ACLK responses.
Fixes a pre-existing bug that non-ascii characters were encoded as `\0000` rather than `\u0000`.

##### Component Name
aclk

##### Test Plan
Claim a node and connect to staging.


aclk/agent_cloud_link.c
```diff
@@ -723,13 +723,58 @@ void aclk_del_collector(const char *hostname, const char *plugin_name, const cha
     }
 
     _free_collector(tmp_collector);
+
+}
+/*
+ * Take a buffer, encode it and rewrite it
```
gobwas, line 729: Looks like it doesn't rewrite it anymore?

amoss: It rewrites it into a temporary buffer.

aclk/agent_cloud_link.c
```diff
@@ -745,26 +791,34 @@ int aclk_execute_query(struct aclk_query *this_query)
...
+        char *encoded_header = aclk_encode_response(w->response.header_output->buffer, w->response.header_output->len, 1);
 
         buffer_sprintf(
-            local_buffer, "{\n\"code\": %d,\n\"body\": \"%s\"\n}", rc, encoded_response);
+            local_buffer, "{\n\"code\": %d,\n\"body\": \"%s\",\n\"headers\": \"%s\"\n}", 
+            w->response.code, encoded_response, encoded_header);
 
         buffer_sprintf(local_buffer, "\n}");
 
+        info("Response:%s", encoded_header);
```
stelfrag, line 811: Will we keep the `info` or use `debug(D_ACLK, ... )` ?

stelfrag:
Response header looks good expect from the `Date:` and `Expires:` 

- Claimed agent works fine in staging with the enhanced payload

@gobwas Is the cloud side decoding the `headers` or is it silently ignored? 

```
HTTP/1.1 200 OK\u000d\nConnection: close\u000d\nServer: NetData Embedded HTTP Server v1.21.1-40-g37a963f3\u000d\nAccess-Control-Allow-Origin: *\u000d\nAccess-Control-Allow-Credentials: true\u000d\nContent-Type: application/json; charset=utf-8\u000d\nDate: Thu, 01 Jan 1970 00:00:00 GMT\u000d\nCache-Control: no-cache, no-store, must-revalidate\u000d\nPragma: no-cache\u000d\nExpires: Thu, 01 Jan 1970 00:00:01 GMT\u000d\nContent-Length: 27892\u000d\n\u000d\n
```

cosmix:
@stelfrag the cloud side should be returning headers (currently we care about the `Expires` header) to the caller so that the intermediary reverse proxy can cache data.

aclk/agent_cloud_link.c
amoss: That was supposed to be deleted...

stelfrag:
> @stelfrag the cloud side should be returning headers (currently we care about the `Expires` header) to the caller so that the intermediary reverse proxy can cache data.

@cosmix -- apologies !! I had a confusing typo in my comment. Right now the agent will return an

`Expires: Thu, 01 Jan 1970 00:00:01 GMT`

header field. Will the cloud side return **that** to the caller?

amoss:
Is the clock on your machine wrong or is it always returning the start of the epoch?

stelfrag:
> Is the clock on your machine wrong or is it always returning the start of the epoch?

It is always start of epoch because the field is not set and ends up with zero in  `w->tv_ready.tv_sec`   


`web_client_build_http_header` L1126

```
if(unlikely(!w->response.data->expires)) {
        if(w->response.data->options & WB_CONTENT_NO_CACHEABLE)
            w->response.data->expires = w->tv_ready.tv_sec + localhost->rrd_update_every;
        else
            w->response.data->expires = w->tv_ready.tv_sec + 86400;
    }
```

amoss:
I've added the initialization of the timer field but I can't test it as I'm in the middle of another PR. Could you check that it produces the correct expiry header?

stelfrag:
> I've added the initialization of the timer field but I can't test it as I'm in the middle of another PR. Could you check that it produces the correct expiry header?

The `Expires` field is now filled correctly. 

There is also the `Date` field that is not filled (looks like a proper value of `time_t` is expected in `w->response.data->date`)

> "HTTP/1.1 200 OK\\u000d\\nConnection: close\\u000d\\nServer: NetData Embedded HTTP Server v1.21.1-42-g09ec475b\\u000d\\nAccess-Control-Allow-Origin: *\\u000d\\nAccess-Control-Allow-Credentials: true\\u000d\\nContent-Type: application/json; charset=utf-8\\u000d\\nDate: Thu, 01 Jan 1970 00:00:00 GMT\\u000d\\nCache-Control: no-cache, no-store, must-revalidate\\u000d\\nPragma: no-cache\\u000d\\nExpires: Tue, 21 Apr 2020 04:49:38 GMT\\u000d\\nContent-Length: 1557\\u000d\\n\\u000d\\n"

The `Expires` field, as calculated, will be at best one second in the future (the `WB_CONTENT_NO_CACHEABLE` option is set)

stelfrag:
The following seems to take care of the `Date` field as well

```
    now_realtime_timeval(&w->tv_ready);
    w->response.data->date = w->tv_ready.tv_sec;
```
I would consider moving this piece of code just before `web_client_build_http_header(w)` so that the request processing time doesn't affect the expiration

amoss:
That means that there is a bug in the `web_client_build_http_header` function: the relationship between the original date and the expiry time is wrong. The code that sets the expiry should use `update_every` as an offset from `w->response.data->date` not from `tv_ready`, which is just there to track how much time the query processing took for the log. I don't have time to fix this properly as retesting it would take longer than I have. I'll put in what you suggested.

aclk/agent_cloud_link.c
```diff
@@ -723,13 +723,59 @@ void aclk_del_collector(const char *hostname, const char *plugin_name, const cha
...
+ *
+ */
+
+static char *aclk_encode_response(char *src, size_t content_size, int keep_newlines)
+{
+    char *tmp_buffer = mallocz(content_size * 2);
+    char *dst = tmp_buffer;
+    while (content_size > 0) {
+        switch (*src) {
+            case '\n':
```
gobwas, line 739: Just for the symmetry I would add `case '\r':` here too. ;-)

Issue #8775, opened 2020-04-20
Add JSON-C dependency handling to installation and packaging code.


Ferroin:
Per the plan to switch to using JSON-C as a required dependency, we need to add support for it to the install and packaging code.

This will require two PR's, one in this repository, and one in netdata/helper-images.

PR #8776, opened 2020-04-20
Added JSON-C dependency handling to instlal and packaging.

Ferroin:
##### Summary

This adds support to the installer and the packaging code to properly pull in JSON-C so that we can use that instead of our buggy internal JSON parser.

##### Component Name

area/packaging

##### Test Plan

Tested locally and verified in CI.

##### Additional Information

This will require netdata/helper-images#74 to be merged before the CI can pass

Fixes: #8775

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -441,7 +441,12 @@ You may need to check these:
    If your system cannot find zlib, although it is installed
    run me with the option:  --libs-are-really-here
 
-3. You need basic build tools installed, like:
+3. The pacakge json-c-dev (or json-c-devel)has to be installed.
+
+   If your system cannot find json-c, although it is installed
+   run me with the option:  --libs-are-really-here
+
+4. You need basic build tools installed, like:
 
    gcc make autoconf automake pkg-config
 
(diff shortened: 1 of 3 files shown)


Ferroin:
The Docker related stuff is now merged.

underhood:
Should we update banner in `netdata-installer.sh`?
`build_error()` @424 to include info about this?

Ferroin:
@underhood probably.

thiagoftsm:
I used Debian 10.3 to test this PR now, I removed `libjson-c` before to compile, but the compilation was until the end, and it gave me the following output:

```
checking for json_object_get_type in -ljson... no
checking for dlopen in -ldl... yes
checking if netdata dbengine should be used... yes
checking if netdata https should be used... yes
checking if json-c should be used... no
```

Ferroin:
> I used Debian 10.3 to test this PR now, I removed `libjson-c` before to compile, but the compilation was until the end, and it gave me the following output:
> 
> ```
> checking for json_object_get_type in -ljson... no
> checking for dlopen in -ldl... yes
> checking if netdata dbengine should be used... yes
> checking if netdata https should be used... yes
> checking if json-c should be used... no
> ```

This PR only adds logic to the installer and packaging code to ensure that JSON-C is present. I'm intentionally avoiding touching the build system, as I don't want to break anything there and we try to use JSON-C if it's present.

PR #8777, opened 2020-04-20
Log malformed URLs instead of ignoring them silently

### LINKED ISSUE #8774: There is no log of a failure to understand requests from the cloud
If the URL in the payload for a cloud request is not understood then no error is log (the agent fails to process it silently). To see this connect an agent without json-c installed to the cloud, no charts are produced and no error is logged.

amoss:
##### Summary
Closes #8774 
Add a log message when the agent does not understand the cloud URL in a request.

##### Component Name
aclk

##### Test Plan
Run the agent without json-c installed (before we patch to make it a hard dep) and check the error is shown when the agent is connected to the cloud.

##### Additional Information
Problem will be hidden once json-c is a hard dep, must be tested before then.

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -124,6 +124,10 @@ int cloud_to_agent_parse(JSON_ENTRY *e)
                     data->payload = mallocz(len+1);
                     if (!url_decode_r(data->payload, e->data.string, len + 1))
                         strcpy(data->payload, e->data.string);
+                    else {
+                        data->payload[0] = 0;
+                        error("Malformed URL from cloud, cannot decode: %s",e->data.string);
+                    }
                 }
                 break;
             }


underhood:
For the test set in test plan it works as expexted:
```
2020-04-20 16:03:39: netdata INFO  : ACLK_Query : AGENT stable, last collector initialization activity was 11 seconds ago                                
2020-04-20 16:03:39: netdata INFO  : ACLK_Query : Sending MQTT len=201676 starts 09 7b 22                                                                2020-04-20 16:03:39: netdata INFO  : ACLK_Query : Sending MQTT len=1166346 starts 09 7b 22                                                               
2020-04-20 16:04:25: netdata ERROR : ACLK_Main : Malformed URL from cloud, cannot decode: /api/v1/data?chart=system.cpu\u0026_=1587391465141\u0026format=array\u0026points=172\u0026group=average\u0026gtime=0\u0026options=flip%7Cabs%7Cjsonwrap%7Cnonzero\u0026after=-300\u0026dimensions=guest_nice,guest,steal
,softirq,irq,user,system,nice,iowait                                                                                                                     2020-04-20 16:04:25: netdata ERROR : ACLK_Main : Malformed URL from cloud, cannot decode: /api/v1/data?chart=mem.available\u0026_=1587391465150\u0026form
at=array\u0026points=172\u0026group=average\u0026gtime=0\u0026options=flip%7Cabs%7Cjsonwrap%7Cnonzero\u0026after=-300\u0026dimensions=MemAvailable       2020-04-20 16:04:25: netdata ERROR : ACLK_Main : Malformed URL from cloud, cannot decode: /api/v1/data?chart=disk.dm-0\u0026_=1587391465166\u0026format=a
rray\u0026points=172\u0026group=average\u0026gtime=0\u0026options=flip%7Cabs%7Cjsonwrap%7Cnonzero\u0026after=-300\u0026dimensions=reads,writes           2020-04-20 16:04:25: netdata ERROR : ACLK_Main : Malformed URL from cloud, cannot decode: /api/v1/data?chart=disk.sr0\u0026_=1587391465214\u0026format=ar
ray\u0026points=172\u0026group=average\u0026gtime=0\u0026options=flip%7Cabs%7Cjsonwrap%7Cnonzero\u0026after=-300\u0026dimensions=reads,writes            2020-04-20 16:04:25: netdata ERROR : ACLK_Main : Malformed URL from cloud, cannot decode: /api/v1/data?chart=disk.sdb\u0026_=1587391465226\u0026format=ar
ray\u0026points=172\u0026group=average\u0026gtime=0\u0026options=flip%7Cabs%7Cjsonwrap%7Cnonzero\u0026after=-300\u0026dimensions=reads,writes            2020-04-20 16:04:25: netdata ERROR : ACLK_Main : Malformed URL from cloud, cannot decode: /api/v1/data?chart=disk.sda\u0026_=1587391465241\u0026format=ar
ray\u0026points=172\u0026group=average\u0026gtime=0\u0026options=flip%7Cabs%7Cjsonwrap%7Cnonzero\u0026after=-300\u0026dimensions=reads,writes            
```

aclk/agent_cloud_link.c
```diff
@@ -124,6 +124,10 @@ int cloud_to_agent_parse(JSON_ENTRY *e)
                     data->payload = mallocz(len+1);
                     if (!url_decode_r(data->payload, e->data.string, len + 1))
                         strcpy(data->payload, e->data.string);
+                    else {
```
stelfrag, line 127: The `else` is the success case so this will not work.

stelfrag:
Fails with correct URLs

```
Malformed URL from cloud, cannot decode: /api/v1/data?chart=disk.nvme0n1&_=1587392126750&format=array&points=172&group=average&gtime=0&options=flip%7Cabs%7Cjsonwrap%7Cnonzero&after=-300&dimensions=reads,writes
```
The `url_decode_r` cannot detect the \u0026 in the json payload so it will not raise an error.

underhood:
Yes 2nd test with json c installed also has the errors :/

aclk/agent_cloud_link.c
stelfrag: It copies the original string so that the final error will be generated by the webapi.

amoss: So is there an error being sent back over the link in the response?

stelfrag: There will the `code` set and the body will have the error message. It will send the same error code and text description that we get from typing the malformed url in `/api/v1/data?.....` 


Example

If the cloud requests a chart `system.lio` instead of `system.io` the following will be generated

(note this includes the TTL PR)


"\t{\"type\": \"http\",\n\t\"msg-id\": \"1fbde3c8-f9e4-4a0d-ad49-f0ae89e50884\",\n\t\"timestamp\": 1587396428,\n\t\"timestamp-offset-usec\": 346294,\n\t\"connect\": 1587396241,\n\t\"connect-offset-usec\": 135655,\n\t\"version\": 1,\n\t\"payload\": {\n\"code\": 404,\n\"body\": \"Chart is not found: system.lio\",\n\"headers\": \"HTTP/1.1 404 Not Found\\u000d\\nConnection: close\\u000d\\nServer: NetData Embedded HTTP Server v1.21.1-40-g37a963f3\\u000d\\nAccess-Control-Allow-Origin: *\\u000d\\nAccess-Control-Allow-Credentials: true\\u000d\\nContent-Type: text/plain; charset=utf-8\\u000d\\nDate: Thu, 01 Jan 1970 00:00:00 GMT\\u000d\\nCache-Control: no-cache, no-store, must-revalidate\\u000d\\nPragma: no-cache\\u000d\\nExpires: Thu, 01 Jan 1970 00:00:01 GMT\\u000d\\nContent-Length: 30\\u000d\\n\\u000d\\n\"\n}\n}"

amoss: I didn't release that, then I will close the PR without merging it.

amoss:
This is not necessary after a comment by @stelfrag

PR #8802, opened 2020-04-22
Removeed Polyverse Polymorphic Linux from Docker builds.

Ferroin:
##### Summary

This cuts the final Docker image sizes for 64-it x86 by approximately 140MB.

The functionality being removed here only ever gets used on 64-bit x86 Docker images.

The claimed benefit to the functionality being removed is improved resistance to exploits that rely on knowledge of the memory layout of certain structures. This is great in theory, but is of questionable benefit in practice for a couple of reasons:

* Because it's being done at build time and not startup, each instance of our containers started from a given image is the same. That consistency at least partially undermines the randomization being done. It was deemed in the original request that doing this unconditionally at runtime would be too much overhead.
* It doesn't actually protect Netdata itself, which is the primary attack surface for the container.
* While the class of bugs this protects against is indeed rather common, actually exploiting them generally requires access to vulnerable code in the target system. Our Docker image only runs Netdata, so being able to exploit these will usually mean that Netdata has already been compromised within the container, which is in turn also the only thing worth compromising in the container.

This still allows for runtime support similarly to how we previously supported it.

##### Component Name

area/packaging

##### Test Plan

Image builds verified in CI on this PR. Actual image sizes and continued correct fnctionality verified locally.

##### Additional Information

Originally added by #5137 based on proposal in #5034.

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -57,17 +57,6 @@ ARG ARCH
 # This image contains preinstalled dependecies
 FROM netdata/base:${ARCH}
 
-# Conditional subscription to Polyverse's Polymorphic Linux repositories
-RUN if [ "$(uname -m)" == "x86_64" ]; then \
-        apk update && apk upgrade; \
-        curl https://sh.polyverse.io | sh -s install gcxce5byVQbtRz0iwfGkozZwy support+netdata@polyverse.io; \
-        if [ $? -eq 0 ]; then \
-                apk update && \
-                apk upgrade --available --no-cache && \
-                sed -in 's/^#//g' /etc/apk/repositories; \
-        fi \
-    fi
-
 # Copy files over
 RUN mkdir -p /opt/src
 COPY --from=builder /app /
--- a/packaging/docker/README.md
+++ b/packaging/docker/README.md
@@ -27,9 +27,10 @@ documentation](https://docs.docker.com/engine/reference/builder/#understand-how-
 
 ### Package scrambling in runtime (x86_64 only)
 
-Our x86_64 Docker images use [Polymorphic Polyverse Linux package scrambling](https://polyverse.io/how-it-works/). For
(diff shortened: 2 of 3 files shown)


Ferroin:
> Yes I approve. I never got the point of why we did this in the first place.

AIUI, largely because someone from Polyverse asked us to. It's supposed to protect against certain types of exploits (similar to stuff in the kernel like SLAB freelist randomization). The problem is that _we_ don't have the same protection in Netdata itself, which is the only real attack target in the images.

> Please don't merge this until afger the `v1.22` release though.

Indeed.

Ferroin:
@cakrit, @ktsaou, @cosmix Your opinions on this would be appreciated.

The general upshot is that this is shrinking our 64-bit x86 Docker images (total space savings for the compressed images (what gets downloaded) is about 40MB, and for the uncompressed images (what gets saved on user systems) by 140MB) and speeding up our Docker image builds a bit by removing a build-time step that is of little practical benefit to most of our users.

cakrit:
Bringing it up for discussion at the C-level tomorrow.

Ferroin:
> Bringing it up for discussion at the C-level tomorrow.

If you need any more info or have questions about it, just let me know.

cakrit:
We have the green light on this, as long as we have the option to reintroduce it in the future. So I'd comment out the relevant code and leave a few comments in there.

Ferroin:
> We have the green light on this, as long as we have the option to reintroduce it in the future. So I'd comment out the relevant code and leave a few comments in there.

@cakrit The current approach taken here is to move the relevant code into the startup code for the Docker image and gate it behind the same environment variable we use for enabling this functionality at runtime, so that users who really want it can still use it. Is this approach sufficient or should we do something else?

cosmix:
It’s ok @Ferroin.

Ferroin:
> It’s ok @Ferroin.

Ok, in that case, merging this since it already has required approvals and has sat around for almost a month total.

PR #8813, opened 2020-04-23
Docs: Add instructions to persist metrics and restart policy

### LINKED ISSUE #8663: Updating netdata in docker (image update)
I've started netdata in following way:
```
docker run -d --name=netdata \
  -p 19999:19999 \
  -e PGID=995 \
  -e DO_NOT_TRACK=1 \
  -v /home/pi/docker_configs/netdata/etc/netdata:/etc/netdata \
  -v /etc/passwd:/host/etc/passwd:ro \
  -v /etc/group:/host/etc/group:ro \
  -v /proc:/host/proc:ro \
  -v /sys:/host/sys:ro \
  -v /etc/os-release:/host/etc/os-release:ro \
  -v /var/run/docker.sock:/var/run/docker.sock:ro \
  -v /etc/os-release:/host/etc/os-release:ro \
  --cap-add SYS_PTRACE \
  --security-opt apparmor=unconfined \
  --restart unless-stopped \
  netdata/netdata
```
And seems it's working fine. Now I wish to update image on latest one. So what I am doing?

1. stopped and removed netdata container
2. removed old image
3. pulled latest image and started newer container

It's also fine but I lose my metrics history because `/var/lib/netdata` isn't mounted and previous one was removed with image. So after I've decided to launch image with mounting `/var/lib/netdata`:
```
docker run -d --name=netdata \
  -p 19999:19999 \
  -e PGID=995 \
  -e DO_NOT_TRACK=1 \
  -v /home/pi/docker_configs/netdata/etc/netdata:/etc/netdata \
  -v /home/pi/docker_configs/netdata/var/lib/netdata:/var/lib/netdata \
  -v /etc/passwd:/host/etc/passwd:ro \
  -v /etc/group:/host/etc/group:ro \
  -v /proc:/host/proc:ro \
  -v /sys:/host/sys:ro \
  -v /etc/os-release:/host/etc/os-release:ro \
  -v /var/run/docker.sock:/var/run/docker.sock:ro \
  -v /etc/os-release:/host/etc/os-release:ro \
  --cap-add SYS_PTRACE \
  --security-opt apparmor=unconfined \
  --restart unless-stopped \
  netdata/netdata
```
But unfortunatelly this is not a finish, because `netdata` demon initialize `/var/lib/netdata` dir during `docker build` but not `docker run`. So you just can't mount `/var/lib/netdata` to container because it overwrites already full directory in container.
```
2020-04-09 13:59:40: netdata INFO  : MAIN : Cannot open the file /var/lib/netdata/health.silencers.json, so Netdata will work with the default health configuration.
2020-04-09 13:59:40: netdata FATAL : MAIN :Cannot create directory '/var/lib/netdata/registry'. # : Invalid argument

2020-04-09 13:59:40: netdata INFO  : MAIN : EXIT: netdata prepares to exit with code 1...
2020-04-09 13:59:40: netdata INFO  : MAIN : EXIT: cleaning up the database...
2020-04-09 13:59:40: netdata INFO  : MAIN : Cleaning up database [0 hosts(s)]...
2020-04-09 13:59:40: netdata INFO  : MAIN : EXIT: all done - netdata is now exiting - bye bye..
```

So how to bypass this problem? W/o this feature nobody can't update image w/o losing usefull data. Thanks.

### LINKED ISSUE #8800: Container Restart Policy
##### Feature idea summary
https://hub.docker.com/r/netdata/netdata described a Docker run commands but they don't include a restart policy. 

##### Expected behavior
I would be default include `--restart=unless-stopped` in those commands so that when a host machine is restarted the Container also startup again if it wasn't manually stopped. Alternatively I would suggest to at least mention the `--restart` flag as optional in the documentation with a note that host restart doesn't automatically start the Netdata container.

joelhans:
##### Summary

Fixes #8800
Fixes #8663

I've added two new instructions to the Docker docs:

1. Added a `--restart unless-stopped` policy to both the `docker run...` and Docker Compose methods of starting the container. (#8800)
2. Added a section containing instructions on how to create folders for the volumes, set permissions, and add the volumes to both `docker run...` and Docker Compose methods.

Now, I'm no Docker expert. These instructions worked for me, but if there's a way to optimize this process, or even have these named volumes (with correct permissions) set up automatically without having to `mkdir`/`chmod`/etc, I'm all ears.

##### Component Name

packaging/docker


ilyam8:
Named volumes are volumes managed by docker server.

https://github.com/netdata/netdata/issues/8663#issuecomment-612182386

To use them we need to add 

```cmd
  --mount source=netdatalib,target=/var/lib/netdata \
  --mount source=netdatacache,target=/var/cache/netdata \
```

to the docker run

or

```cmd
      - type: volume
        source: netdatalib
        target: /var/lib/netdata
      - type: volume
        source: netdatacache
        target: /var/cache/netdata
```

to the docker-compose

ilyam8:
This is not good and i think we shouldnt suggest it.

```
mkdir -p $(pwd)/netdata/var/lib/netdata/
mkdir -p $(pwd)/netdata/var/cache/netdata/
sudo chown :1024 -R netdata/
sudo chmod 775 -R netdata/
```

ilyam8:
And i think

https://github.com/netdata/netdata/pull/8813#issuecomment-618444252

this should be default and not in `if you want to persist` section.

ilyam8:
> https://docs.docker.com/develop/dev-best-practices/#where-and-how-to-persist-application-data#where-and-how-to-persist-application-data

> Avoid storing application data in your container’s writable layer using storage drivers. This increases the size of your container and is less efficient from an I/O perspective than using volumes or bind mounts.
Instead, store data using volumes.

That is what we are doing now ^^


> One case where it is appropriate to use bind mounts is during development, when you may want to mount your source directory or a binary you just built into your container. For production, use a volume instead, mounting it into the same location as you mounted a bind mount during development.

Bind mounts for development, named volumes for production


```cmd
mkdir -p $(pwd)/netdata/var/lib/netdata/
mkdir -p $(pwd)/netdata/var/cache/netdata/
sudo chown :1024 -R netdata/
sudo chmod 775 -R netdata/
```

^^ bind mounts

joelhans:
@ilyam8 Please take another look when you have time. I added the `--mount` options to the `docker run` command and added those `netdatalib`/`netdatacache` volumes to the Compose file. I definitely agree that this is much more streamlined.

Sorry for the back and forth. I'm just not that well-versed in Docker and had trouble following the issue to figure out the best solution.

ilyam8:
@joelhans have you tested it? it looks good to me but it is better to test the config anyway

ilyam8:
ive tested, all is ok 👍

joelhans:
@ilyam8 Just tested the `docker run` version and it works OK when stopping, removing, and updating the version of the image.

prologic:
I actually wondering why we are encouraging bind-mounts in `docker run --mount` but named volumes in `docker-compose`. See [this q&a](https://serverfault.com/questions/996785/docker-volumes-vs-mount-binds-what-are-the-use-cases)

ilyam8:
i see no bind mounts, `--mount` is just syntax (more clear and explicit then `-v`). 

> Originally, the -v or --volume flag was used for standalone containers and the --mount flag was used for swarm services. However, starting with Docker 17.06, you can also use --mount with standalone containers. In general, --mount is more explicit and verbose. The biggest difference is that the -v syntax combines all the options together in one field, while the --mount syntax separates them. Here is a comparison of the syntax for each flag.

> New users should try --mount syntax which is simpler than --volume syntax.

ilyam8:
```
  -v netdatalib:/var/lib/netdata \
  -v netdatacache:/var/cache/netdata \
```

does same, i am ok with both ways

joelhans:
Changed this to the `-v ...` syntax, if only for the sake of consistency. Tested it and it works fine.

prologic:
I am either confused or do not agree:

```#!sh
$ docker run --help | grep -E '(--mount|-v)'
      --mount mount                    Attach a filesystem mount to the container
  -v, --volume list                    Bind mount a volume
      --volume-driver string           Optional volume driver for the container
      --volumes-from list              Mount volumes from the specified container(s)
```

I _could_ be wrong but I don't think these flags are equivalent to each other?

prologic:
The explanation is also kind of confusing :/

ilyam8:
@joelhans if netlify errors are not related i suggest to merge the PR

joelhans:
@ilyam8 I thought we were in a code freeze. Maybe I'm wrong?

> The explanation is also kind of confusing :/

@prologic Which is confusing? My docs or the `docker run` output? If you have a concern with the PR I'd like to address it before I merge.

prologic:
> @prologic Which is confusing? My docs or the `docker run` output? If you have a concern with the PR I'd like to address it before I merge.

The Docker man pages :)

ilyam8:
> I thought we were in a code freeze

Yeah, i think you are right. @prologic Do we treat doc changes as _code changes_?

prologic:
> Yeah, i think you are right. @prologic Do we treat doc changes as _code changes_?

I personally consider "Doc Changes" as "bug fixes" as they "fix documentation.

joelhans:
Sorry for the re-requested reviews... had to rebase to fix a conflict.

PR #8824, opened 2020-04-25
Removed old function call in netdata-installer.sh

Ferroin:
##### Summary

When updating the dashboard handling code, I forgot to remove this call to a now nonexistent function.

##### Component Name

area/packaging

##### Test Plan

Passes CI.


ilyam8:
I searched in the source code, no other mention of `clean_existing_dashboard` https://github.com/netdata/netdata/search?q=clean_existing_dashboard&unscoped_q=clean_existing_dashboard

PR #8838, opened 2020-04-27
JSON-c as hard dep. for ACLK

underhood:
##### Summary

1. If `json-c-devel` (or equivalent pkg in other distros) available use it.
1. If not try and check if static build is available in `externaldeps` folder (by installer) and use it.
1. If no json-c is available don't build ACLK in `detect` (default mode - build ACLK if you can but build netdata always) and error out in configure if `--enable-aclk` flag given (telling us user specifically wants ACLK to be built)

##### Test Plan
Test with and without system JSON-c dev package. Test with configure flags `--enable-cloud`, `--disable-cloud`, and no flag (default autodetect mode). Test with renaming `libjson-c.a` and running configure (without installer who would build new lib) to simulate the case where the installer was not able to build static json-c.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -624,6 +624,10 @@ NETDATA_COMMON_LIBS = \
     $(OPTIONAL_EBPF_LIBS) \
     $(NULL)
 
+if LINK_STATIC_JSONC
+    NETDATA_COMMON_LIBS += externaldeps/jsonc/libjson-c.a
+endif
+
 NETDATACLI_FILES = \
     daemon/commands.h \
     $(LIBNETDATA_FILES) \
--- a/configure.ac
+++ b/configure.ac
@@ -436,8 +436,8 @@ AM_CONDITIONAL([ENABLE_HTTPS], [test "${enable_https}" = "yes"])
 # JSON-C
 
 if test "${enable_jsonc}" != "no" -a -z "${JSONC_LIBS}"; then
-	# Try and detect manual static build presence (from netdata-installer.h)
-	AC_MSG_CHECKING([if staticaly built json-c is present])
+	# Try and detect manual static build presence (from netdata-installer.sh)
+	AC_MSG_CHECKING([if statically built json-c is present])
 	if test -f "externaldeps/jsonc/libjson-c.a"; then
 		AC_DEFINE([LINK_STATIC_JSONC], [1], [static json-c should be used])
 		JSONC_LIBS="static"
--- a/configure.ac
+++ b/configure.ac
@@ -173,7 +173,6 @@ AC_ARG_ENABLE(
     [ enable_cloud="detect" ]
 )
 
-aclk_required="${enable_cloud}"
 if test "${enable_cloud}" = "no"; then
     AC_DEFINE([DISABLE_CLOUD], [1], [disable netdata cloud functionality])
 fi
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -435,6 +434,21 @@ AM_CONDITIONAL([ENABLE_HTTPS], [test "${enable_https}" = "yes"])
 
 # -----------------------------------------------------------------------------
 # JSON-C
+
+if test "${enable_jsonc}" != "no" -a -z "${JSONC_LIBS}"; then
+	# Try and detect manual static build presence (from netdata-installer.h)
```
amoss, line 439:
```suggestion
	# Try and detect manual static build presence (from netdata-installer.sh)
```

configure.ac
```diff
@@ -435,6 +434,21 @@ AM_CONDITIONAL([ENABLE_HTTPS], [test "${enable_https}" = "yes"])
...
+	AC_MSG_CHECKING([if staticaly built json-c is present])
```
amoss, line 440:
```suggestion
	AC_MSG_CHECKING([if statically built json-c is present])
```

configure.ac
```diff
@@ -435,6 +434,21 @@ AM_CONDITIONAL([ENABLE_HTTPS], [test "${enable_https}" = "yes"])
...
+	if test -f "externaldeps/jsonc/libjson-c.a"; then
+		AC_DEFINE([LINK_STATIC_JSONC], [1], [static json-c should be used])
```
amoss, line 442: I think that we should do a test compile to check that it works. The symbol check that is already there for the system installed library would be appropriate.

configure.ac
```diff
@@ -1218,7 +1239,8 @@ AC_SUBST([webdir])
 
 CFLAGS="${CFLAGS} ${OPTIONAL_MATH_CFLAGS} ${OPTIONAL_NFACCT_CFLAGS} ${OPTIONAL_ZLIB_CFLAGS} ${OPTIONAL_UUID_CFLAGS} \
     ${OPTIONAL_LIBCAP_CFLAGS} ${OPTIONAL_IPMIMONITORING_CFLAGS} ${OPTIONAL_CUPS_CFLAGS} ${OPTIONAL_XENSTAT_FLAGS} \
-    ${OPTIONAL_KINESIS_CFLAGS} ${OPTIONAL_PROMETHEUS_REMOTE_WRITE_CFLAGS} ${OPTIONAL_MONGOC_CFLAGS} ${LWS_CFLAGS}"
+    ${OPTIONAL_KINESIS_CFLAGS} ${OPTIONAL_PROMETHEUS_REMOTE_WRITE_CFLAGS} ${OPTIONAL_MONGOC_CFLAGS} ${LWS_CFLAGS} \
+	${OPTIONAL_JSONC_STATIC_CFLAGS}"
```
amoss, line 1243: Does this add a duplicate (looking at the change to Makefile.am) ?

underhood: nope this adds include dir. Change in Makefile adds static lib to be linked

Makefile.am
```diff
@@ -624,6 +624,10 @@ NETDATA_COMMON_LIBS = \
     $(OPTIONAL_EBPF_LIBS) \
     $(NULL)
 
+if LINK_STATIC_JSONC
```
amoss, line 627: This is technically correct but seems fragile (i.e. we should only end up with LINK_STATIC_JSONC set if OPTIONAL_JSONC_LIBS is empty). What about the following to make the relationship guaranteed?
```
If LINK_STATIC_JSONC
     NETDATA_COMMON_LIBS += externaldeps/jsonc/libjson-c.a
else
    NETDATA_COMMON_LIBS += $(OPTIONAL_JSONC_LIBS)
endif
```

PR #8840, opened 2020-04-27
Fix SSL renegotiation errors in old versions of openssl.

amoss:
##### Summary
Fix SSL renegotiation errors in old versions of openssl.

##### Component Name
aclk

##### Test Plan

* Setup nodefarm - connects variety of nodes to cloud staging.
* Use `vagrant suspend` to pause the machine long enough to become unreachable
* Use `vagrant resume` to un-pause the VM and force reconnection.

Run the agent through valgrind and check the shutdown sequence.

##### Additional Information

Not ready for review yet, need to rework the shutdown sequence and do some more testing.


underhood:
+1 on moving the context destroy out of the LWS statemachine

amoss:
This seems to be stable and is ready for review. I've tested the following configurations in the node farm:
* CentOS 7
* Debian 9
* Debian 10
* Fedora 30
* Ubuntu 16.04
* Ubuntu 18.04
I can't test the following because of a transient error that might clear up tomorrow (`dnf update` crashes).
* Fedora 31

In each configuration I've checked the following:
* Cause a timeout to drop the connection (e.g. `(cd box-debian-9; vagrant suspend; sleep 30; vagrant resume)`.
* Drop the connection from the network side (e.g. `ss -K dst 104.31/16`).

In some of the configurations I've checked the execution outside of `systemctl` with `valgrind` and `gdb` to check that the shutdown behaviour is clean, we have no memory leaks or initialised variables and the code paths do what I expect during the tear-down sequence.

@stelfrag / @underhood : Please proceed with review and testing, I'm happy this is stable and fixes the two problems that I know of (the reuse of the ssl socket and the inversion of the teardown sequence).

In parallel to your review I will continue to investigate other reports we have had (i.e. this connection already in use issue), expand the test set to cover more of our support matrix and maybe try to automate some really nasty tests. If I find any other definite problems I will fix them in a separate PR after this one is merged.

amoss:
Rebased to pick up the claiming changes and the installer changes for json-c.

amoss:
Rebased again to get the json-c changes to the build system.

PR #8851, opened 2020-04-29
Disable CI checks for openSUSE temporarily

prologic:
##### Summary

ssia

##### Component Name

- area/ci

##### Test Plan

This PR and #8840 (_onced it is rebased_).

##### Additional Information


.github/workflows/build-and-install.yml
```diff
@@ -25,9 +26,9 @@ jobs:
           - 'fedora:32'
           - 'fedora:31'
           - 'fedora:30'
-          - 'opensuse/leap:15.2'
-          - 'opensuse/leap:15.1'
-          - 'opensuse/tumbleweed:latest'
+            #- 'opensuse/leap:15.2'
+            #- 'opensuse/leap:15.1'
+            #- 'opensuse/tumbleweed:latest'
```
Ferroin, line 31: If we're just disabling fail-fast behavior, we can probably safely keep these enabled.

amoss:
Cherry-picked the commit directly into the affected PR.

.github/workflows/build-and-install.yml
prologic: See my previous comments about this in other PRs. I'd rather we remove openSUSE entirely for the time being.

PR #8836, opened 2020-04-27
Added installer code to bundle JSON-c if it's not present.

Ferroin:
##### Summary

Based on discussion earlier today about how to handle JSON-c.

If the cloud functionality isn't disabled and `pkg-config` can't find a locally installed copy of JSON-c, the installer will now download and build a static copy to be linked into Netdata.

This code can be forcibly enabled by passing `--build-json-c` to the installer.

##### Component Name

area/packaging

##### Test Plan

Manually verified that the code does what it's supposed to.

##### Additional Information

The structure under the `externaldeps` directory for this is as follows:

* Everything is put into `externaldeps/jsonc`.
* The static library is called `libjson-c.a` and is located in the root f the aforementioned directory.
* The header files are under `json-c`, just like they would be under `/usr/include` on most systems.

PR #8853, opened 2020-04-29
Added CI to verify JSON-C bundling code in installer.

Ferroin:
##### Summary

This adds CI code to verify the JSON-C bundling functionality introduced by #8838 and #8836.

##### Component Name

area/packaging

##### Test Plan

CI passes on this PR.

##### Additional Information

The new CI check explicitly requires the cloud functionality to build successfully when JSON-C is not provided by the system itself.

We skip this check on Arch Linux as JSON-C is part of the base install there (as a result of it being a dependenc of `cryptsetup`, which is part of the Arch Linux base install).

We skip this check on CentOS 7 as the version of `cmake` it provides is incompatible with the JSON-C build infrastructure. Installs on CentOS 7 when JSON-C is not installed through the package manager will succeed but will not have Netdata Cloud functionality.

### DIFF
--- a/.github/workflows/build-and-install.yml
+++ b/.github/workflows/build-and-install.yml
@@ -49,12 +49,9 @@ jobs:
 
           - distro: 'archlinux:latest'
             pre: 'pacman --noconfirm -Sy grep libffi'
-            rmjsonc: 'pacman -R --noconfirm json-c'
 
           - distro: 'centos:8'
             rmjsonc: 'dnf remove -y json-c-devel'
-          - distro: 'centos:7'
-            rmjsonc: 'yum remove -y json-c-devel'
 
           - distro: 'debian:bullseye'
             pre: 'apt-get update'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-install.yml
```diff
@@ -25,56 +26,92 @@ jobs:
           - 'fedora:32'
           - 'fedora:31'
           - 'fedora:30'
-            #- 'opensuse/leap:15.2'
-            #- 'opensuse/leap:15.1'
-            #- 'opensuse/tumbleweed:latest'
+          - 'opensuse/leap:15.2'
+          - 'opensuse/leap:15.1'
+          - 'opensuse/tumbleweed:latest'
```
prologic, line 31: Was this on purrpose?

Ferroin: Yes.

With `fail-fast: false` and the fact that I've turned off OpenSUSE as a required check, running these won't block PR's anymore. I'd suggest we reenable the Leap checks as they work fine and just leave Tumbleweed out of the required checks going forwards.

PR #8855, opened 2020-04-29
Add a Google Cloud Pub/Sub connector to the exporting engine

### LINKED ISSUE #8023: Implement Google Cloud Pub/Sub exporting connector
##### Feature idea summary
We need to analyze a huge amount of data from Netdata using big data instruments. We prefer to use Google Cloud services for the task. It's not convenient to use a TSDB and then convert the data to an appropriate format. So, we need a new exporting connector. 

One possibility we have is to use a columnar file format and push data to the Google Cloud Storage service. The other one looks more general/reasonable to implement - ingest the data to big data services using the Pub/Sub service. The latter also has more chances to be useful for Netdata users.

The problem is that Google Cloud SDK doesn't have C/C++ API which would be easy to integrate into Netdata. We have to investigate if we can use the gRPC service interface as an alternative to the Client API.

##### Tasks to complete:
- Initialize a connector instance.
- Implement dimension and batch formatters.
- Implement a worker.
- Add specific configuration options.
- Configure build (Autotools and CMake).
- Check integration with other protobuf-based libraries.
- Implement internal stats.
- Write documentation.
- Document functions.
- Manual testing.
- Write unit tests.

##### Expected behavior
Metrics can be exported from Netdata to Google Cloud data analytics services.

vlvkobal:
##### Summary
Implement a Google Cloud Pub/Sub exporting connector.

Closes #8023

##### Component Name
exporting engine

##### Test Plan
Create a Pub/Sub topic and a subscriber.
Get a file with Google Cloud credentials. Make it readable by Netdata. Configure the `credentials file` option.

`exporting.conf`
```
[pubsub:instance10]
enabled = yes
destination = pubsub.googleapis.com
credentials file = /etc/netdata/credentials.json
project id = my_project_id
topic id = test_exporting_connector
data source = as collected
update every = 10
send charts matching = system.processes
```
Pull messages received from Netdata at the Google Cloud [Pub/Sub console](https://console.cloud.google.com/cloudpubsub/topic/detail/test_exporting_connector).


ktsaou:


```diff

Clones added
============
- exporting/pubsub/pubsub_publish.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5442008&bid=17845621)

ktsaou:


```diff

Clones added
============
- exporting/pubsub/pubsub_publish.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5442008&bid=17845621)

ktsaou:


```diff

Clones added
============
- exporting/pubsub/pubsub_publish.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5442008&bid=17845621)

ktsaou:


```diff

Clones added
============
- exporting/pubsub/pubsub_publish.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5442008&bid=17845621)

ktsaou:


```diff

Clones added
============
- exporting/pubsub/pubsub_publish.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5442008&bid=17845621)

exporting/pubsub/pubsub.c
```diff
@@ -0,0 +1,155 @@
...
+    struct pubsub_specific_data *connector_specific_data = instance->connector_specific_data;
+
+    while (!netdata_exit) {
+        struct stats *stats = &instance->stats;
+        char error_message[ERROR_LINE_MAX + 1] = "";
+
+        uv_mutex_lock(&instance->mutex);
+        uv_cond_wait(&instance->cond_var, &instance->mutex);
+
+        // reset the monitoring chart counters
```
thiagoftsm, line 79: The reset is inside the `while` loop, so you are not accumulating the whole lifetime of the exporter, there are variables that will have value 1 case an error happen, but this can occur after hundred of iterations. Is this reset in the correct place?

exporting/pubsub/pubsub_publish.cc
```diff
@@ -0,0 +1,222 @@
...
+    const char *project_id, const char *topic_id)
+{
+    struct pubsub_specific_data *connector_specific_data = (struct pubsub_specific_data *)pubsub_specific_data_p;
+
+    try {
+        setenv("GOOGLE_APPLICATION_CREDENTIALS", credentials_file, 0);
+
+        std::shared_ptr<grpc::ChannelCredentials> credentials = grpc::GoogleDefaultCredentials();
+        if (credentials == nullptr) {
+            std::strncpy(error_message, "Can't load credentials", ERROR_LINE_MAX);
```
thiagoftsm, line 42: I will give a suggestion here for we avoid future errors from other developers. I verified that until now every time you called `pubsub_init`, you took care and you reset the array, but in the future case other developers call this function without to reset, we can have a problem. Considering that all the texts stored inside `error_message` are texts with well known size, please, insider a `NULL` character after each strncpy. Thank you!

exporting/pubsub/pubsub_publish.cc
```diff
@@ -0,0 +1,222 @@
...
+
+        grpc::CompletionQueue *cq = new grpc::CompletionQueue;
+        connector_specific_data->completion_queue = cq;
+
+        connector_specific_data->responses = new std::list<struct response>;
+
+        return 0;
+    } catch (std::exception const &ex) {
+        std::string em(std::string("Standard exception raised: ") + ex.what());
+        std::strncpy(error_message, em.c_str(), ERROR_LINE_MAX);
```
thiagoftsm, line 69: Considering that you accepted the suggestion, for this specific `strncpy`, we can use `em.length()` when the size is smaller than `ERROR_LINE_MAX` or `ERROR_LINE_MAX` to set the `NULL` character.

exporting/pubsub/pubsub_publish.cc
```diff
@@ -0,0 +1,222 @@
...
+
+    try {
+        google::pubsub::v1::PubsubMessage *message =
+            ((google::pubsub::v1::PublishRequest *)(connector_specific_data->request))->add_messages();
+        if (!message)
+            return 1;
+
+        message->set_data(data);
+    } catch (std::exception const &ex) {
+        return 1;
```
thiagoftsm, line 95: Should we not get what is the exception message like you did in the previous function?

exporting/pubsub/pubsub_publish.cc
```diff
@@ -0,0 +1,222 @@
...
+    try {
+        grpc::ClientContext *context = new grpc::ClientContext;
+
+        std::unique_ptr<grpc::ClientAsyncResponseReader<google::pubsub::v1::PublishResponse> > rpc(
+            ((google::pubsub::v1::Publisher::Stub *)(connector_specific_data->stub))
+                ->AsyncPublish(
+                    context, (*(google::pubsub::v1::PublishRequest *)(connector_specific_data->request)),
+                    ((grpc::CompletionQueue *)(connector_specific_data->completion_queue))));
+
+        struct response response;
```
thiagoftsm, line 123: Please, rename the function `response`, because it has the same name of the `type`, this can create a confusion.

exporting/pubsub/pubsub_publish.cc
```diff
@@ -0,0 +1,222 @@
...
+                              .AsyncNext(&got_tag, &ok, deadline);
+
+            if (next_status == grpc::CompletionQueue::GOT_EVENT) {
+                for (response = responses->begin(); response != responses->end(); ++response) {
+                    if ((void *)response->tag == got_tag)
+                        break;
+                }
+
+                if (response == responses->end()) {
+                    std::strncpy(error_message, "Cannot get Pub/Sub response", ERROR_LINE_MAX);
```
thiagoftsm, line 186: The same suggestion I gave in other function to insert a `NULL` character, I also give here, please, add a `NULL` character to avoid future errors.

exporting/read_config.c
```diff
@@ -376,15 +385,28 @@ struct engine *read_exporting_config()
...
             tmp_instance->config.connector_specific_config = connector_specific_config;
 
             connector_specific_config->stream_name = strdupz(exporter_get(instance_name, "stream name", "netdata"));
-
             connector_specific_config->auth_key_id = strdupz(exporter_get(instance_name, "aws_access_key_id", ""));
-
             connector_specific_config->secure_key = strdupz(exporter_get(instance_name, "aws_secret_access_key", ""));
         }
 
+        if (tmp_instance->config.type == EXPORTING_CONNECTOR_TYPE_PUBSUB) {
```
thiagoftsm, line 397: Are we  missing a `#ifndef ENABLE_EXPORTING_PUBSUB` before this `if`?  Case we are not, why do we have this instruction at line `316`?

exporting/tests/test_exporting_engine.h
```diff
@@ -145,6 +149,16 @@ void __wrap_kinesis_put_record(
     size_t data_len);
 int __wrap_kinesis_get_result(void *request_outcomes_p, char *error_message, size_t *sent_bytes, size_t *lost_bytes);
 
+int __wrap_pubsub_init(
```
thiagoftsm, line 152: Are we also missing `#if ENABLE_EXPORTING_PUBSUB` here?

exporting/pubsub/pubsub.c
vlvkobal: The counters are absolute, so the place is the correct one.

exporting/pubsub/pubsub_publish.cc
vlvkobal: I don't think it's confusing. The name of the type is `struct response`, not `response`.

exporting/tests/test_exporting_engine.h
vlvkobal: They are harmless declarations. We just don't use these functions in case pubsub is not enabled. There are no warnings when tests are built.

exporting/read_config.c
vlvkobal: At L316, we are checking if Pub/Sub support is compiled, and if it's not, then L397 will never be executed. No need for an additional `#ifdef`.

exporting/pubsub/pubsub_publish.cc
vlvkobal: The only thing I can think of that can go wrong here is memory allocation. It will be a global problem for the whole Netdata, not only for this particular operation. Isn't it enough to print "EXPORTING: Instance %s: Cannot add data to a message"?

exporting/pubsub/pubsub_publish.cc
vlvkobal: I don't see the problem with the explicit array initialization approach as far as `pubsub_init` and other functions from `pubsub_publish.cc` aren't supposed to be used anywhere else.
But, if we really want to be on the safe side, I suggest nullifying only the last byte of the array and don't bother with the string length.

exporting/pubsub/README.md
```diff
@@ -0,0 +1,34 @@
...
+[pubsub:my_instance]
+    enabled = yes
+    destination = pubsub.googleapis.com
+    credentials file = /etc/netdata/google_cloud_credentials.json
+    project id = my_project
+    topic id = my_topic
+```
+
+Set the `destination` option to a Pub/Sub service endpoint. `pubsub.googleapis.com` is the default one.
+You should provide access to the file with Google Cloud credentials for the `netdata` user.
```
joelhans, line 32: Is this saying that the `netdata` user should be able to read the `google_cloud_credential.json` file? Maybe there's another way to phrase that.

And where does `google_cloud_credential.json` come from? Do they download the file and put it in `/etc/netdata`, or create it on their own?

vlvkobal: Yes and yes.
As for the downloading, please look ad https://cloud.google.com/docs/authentication/getting-started. Should we add this link to the documentation?

exporting/pubsub/pubsub_publish.cc
thiagoftsm: To add NULL at the end is enough. I agree your code is not creating possible bugs, my concern is with future codes calling this function.

exporting/pubsub/pubsub_publish.cc
thiagoftsm: Yes, this would be enough. I think we need to report an error if exceptions happen.

exporting/pubsub/pubsub_publish.cc
thiagoftsm: All right, I agree with what you wrote, this is not a `typedef` call, but it was strange to my eyes when I read `response response`.

exporting/read_config.c
thiagoftsm: Case it is not supported, will we have `EXPORTING_CONNECTOR_TYPE_PUBSUB` defined in the default headers? This was the main motive I asked you.

exporting/pubsub/README.md
joelhans, line 32: We can start with something like this. I may not have the permissions right—feel free to correct me if I'm wrong!

```suggestion
Next, create the credentials JSON file by following Google Cloud's [authentication guide](https://cloud.google.com/docs/authentication/getting-started#creating_a_service_account). The user running the Agent (typically `netdata`) needs read access to `google_cloud_credentials.json`, which you can set with `chmod 644 google_cloud_credentials.json`.
```

exporting/pubsub/pubsub_publish.cc
vlvkobal: We do report the error. We just don't report the exception error message.

exporting/read_config.c
vlvkobal: Yes, we will have it defined.

exporting/pubsub/pubsub_publish.cc
vlvkobal: Done.

ktsaou:


```diff

Clones added
============
- exporting/pubsub/pubsub_publish.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5442008&bid=17845621)

exporting/pubsub/README.md
vlvkobal: Done.

ktsaou:


```diff

Clones added
============
- exporting/pubsub/pubsub_publish.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5442008&bid=17845621)

PR #8858, opened 2020-04-29
remove old-cloud related UI from old dashboard

jacekkolasa:
##### Summary
In 1.22, old dashboard will still be accessible via /old suffix. Since old cloud will not be available, i've removed:
- sign in button
- old Cloud CTA on bottom of the page
- Nodes View buttons: in top bar and dropdown menu

fixes https://github.com/netdata/product/issues/872

##### Component Name
web/gui

### DIFF
--- a/web/gui/main.js
+++ b/web/gui/main.js
@@ -792,11 +792,6 @@ function renderMyNetdataMenu(machinesArray) {
     if (!isSignedIn()) {
         html += (
             `<div class="agent-item">
-                <i class="fas fa-tv"></i>
-                <a onClick="openAuthenticatedUrl('console.html');" target="_blank">Nodes<sup class="beta"> beta</sup></a>
-                <div></div>
-            </div>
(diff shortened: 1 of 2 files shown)


jacekkolasa:
@cakrit You asked to remove Sign In buttons, but I removed also CTA on bottom of the page and `Nodes View (beta)` buttons. Please confirm if that's ok in case i cut too much.

web/gui/main.js
```diff
@@ -4807,11 +4802,7 @@ function signInDidClick(e) {
 }
 
 function shouldShowSignInBanner() {
```
cosmix, line 4804: Why not just remove the code where this function is called?

jacekkolasa: Simply because that file will most likely not be maintained :)
I could remove much more cloud-related code, one deletion after another, but it doesn't seem to be a priority so I just went with easiest and safest things to do.

PR #8815, opened 2020-04-23
Docs: Update with go-live claiming and ACLK information

joelhans:
##### Summary

**To be merged after the 1.22 code freeze is lifted.**

This PR follows up on #8818/#8850/#8843 to ensure the documentation aligns with the most up-to-date config options/settings.

I added the new settings/output to `aclk/README.md`, `claim/README.md`, and `daemon/README.md`.

##### Component Name

docs
claim/
daemon/
aclk/

PR #8859, opened 2020-04-29
Docs: Update with go-live claiming and ACLK information

joelhans:
##### Summary

Clean version of #8815, comparing my docs against the go-live branch rather than master so that it doesn't include all the other commits as it had in the other PR.

This PR follows up on #8818/#8850/#8843 to ensure the documentation aligns with the most up-to-date config options/settings. 

I added the new settings/output to `aclk/README.md`, `claim/README.md`, and `daemon/README.md`.

##### Component Name

claim/
daemon/
aclk/


aclk/README.md
```diff
@@ -13,36 +13,96 @@ The Agent-Cloud link (ACLK) is the mechanism responsible for connecting a Netdat
 [MQTT](https://en.wikipedia.org/wiki/MQTT) over secure websockets to first create, persist, encrypt the connection, and
 then enable the features found in Netdata Cloud. _No data is exchanged with Netdata Cloud until you claim a node._
 
-Read our [claiming documentation](/claim/README.md) for a guide for claiming a node using the ACLK and additional
-troubleshooting and reference information.
+For a guide for claiming a node using the ACLK, plus additional troubleshooting and reference information, read our
```
amoss, line 16:
```suggestion
For a guide to claiming a node using the ACLK, plus additional troubleshooting and reference information, read our
```

aclk/README.md
```diff
@@ -13,36 +13,96 @@ The Agent-Cloud link (ACLK) is the mechanism responsible for connecting a Netdat
...
+[claiming documentation](/claim/README.md) or the [get started with
+Cloud](https://learn.netdata.cloud/docs/cloud/get-started) guide.
 
 ## Enable and configure the ACLK
 
 The ACLK is enabled by default and automatically configured if the prerequisites installed correctly. You can see this
```
amoss, line 22: No file is created by default, this example does correspond to the default settings but they are stored internally.

aclk/README.md
```diff
@@ -13,36 +13,96 @@ The Agent-Cloud link (ACLK) is the mechanism responsible for connecting a Netdat
...
 
-2.  Change a runtime setting in your `netdata.conf` file. This setting only stops the Agent from attempting any
-    connection via the ACLK, but does not prevent the installer from downloading and compiling the ACLK's dependencies.
+### Disable at runtime
+
+You can change a runtime setting in your `var/lib/netdata/cloud.d/cloud.conf` file to disable the ACLK. This setting
+only stops the Agent from attempting any connection via the ACLK, but does not prevent the installer from downloading
+and compiling the ACLK's dependencies.
+
+Change the `enabled` setting to `no`:
```
amoss, line 55: First ensure the file is created, e.g. using @mfundul 's snippet from slack then set `enabled = no`.

aclk/README.md
```diff
@@ -13,36 +13,96 @@ The Agent-Cloud link (ACLK) is the mechanism responsible for connecting a Netdat
...
 [global]
-    netdata cloud = disable
+    enabled = no
+```
+
+If the file at `var/lib/netdata/cloud.d/cloud.conf` doesn't exist, you need to create it.
+
+```bash
+cd /var/lib/netdata/cloud.d
+cat > cloud.conf << EOF
```
amoss, line 66: This is missing the piece of code below, they cannot be split like this as the here-doc needs to be contiguous:
```bash
cd /var/lib/netdata/cloud.d
cat > cloud.conf << EOF
[global]
    enabled = no
    cloud base url = https://app.netdata.cloud
EOF
```

amoss: Maybe change the line to explain the change in prompt and leave it after the code?

> After you cut and paste the first two lines the prompt will change to `cat`, keep pasting in the lines and after the `EOF` your normal terminal prompt will come back. To get your prompt back the final line must contain only `EOF`, nothing else.

aclk/README.md
```diff
@@ -13,36 +13,96 @@ The Agent-Cloud link (ACLK) is the mechanism responsible for connecting a Netdat
...
+[global]
+    enabled = no
+    cloud base url = https://app.netdata.cloud
+EOF
 ```
 
+You also need to change the file's permissions. Use `grep "run as user" /etc/netdata/netdata.conf` to figure out which
+user your Agent runs as, and replace `netdata:netdata` as shown below if necessary:
+
+```bash
```
amoss, line 81: This will fail under their normal user account, they need to use sudo to do this.

amoss:
In the trouble-shooting section there are now more errors that can be reported if --require-cloud is set:

* "Could not find cmake, which is required to build JSON-C. The install process will continue, but Netdata Cloud support will be disabled."
* "Could not find cmake, which is required to build JSON-C. The install process will continue, but Netdata Cloud support will be disabled."
* "Failed to build JSON-C. Netdata Cloud support will be disabled."
* "Unable to fetch sources for JSON-C. Netdata Cloud support will be disabled."

stelfrag:
> In the trouble-shooting section there are now more errors that can be reported if --require-cloud is set:
> 
> * "Could not find cmake, which is required to build JSON-C. The install process will continue, but Netdata Cloud support will be disabled."
> * "Could not find cmake, which is required to build JSON-C. The install process will continue, but Netdata Cloud support will be disabled."
> * "Failed to build JSON-C. Netdata Cloud support will be disabled."
> * "Unable to fetch sources for JSON-C. Netdata Cloud support will be disabled."

The JSON-C part was copied twice. 

* "Could not find cmake, which is required to build libwebsockets. The install process will continue, but you may not be able to connect this node to Netdata Cloud."

aclk/README.md
joelhans: Got it.

joelhans:
@amoss Thanks for your comments. Pushed the changes, and we'll make sure we cover anything else in our meeting.

PR #8860, opened 2020-04-30
Try building on Clang on Travis too

nabijaczleweli:
##### Summary

Bionic has an ancient Clang, so try building with that. Maybe it'll work to expose #8854

Opening Draft, just to run the CI.

##### Component Name

CI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
 # Install dependencies for all, once
 #
 install:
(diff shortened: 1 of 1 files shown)


prologic:
@nabijaczleweli I actually have a local reproducer with a `Dockerfile` for the issue you and I are both working on at this minute :)

We actually have decided long ago to no longer use Travis and are actively migrating off of it.

If you give me a few mins I'll have a local reproducer posted on #8854 shortly :)

nabijaczleweli:
Oh okay, closing this in that case. I have to sleep now, but it's good to know it's in good hands!

PR #8866, opened 2020-04-30
Make cloud-enabled flag  be independent of ACLK

mfundul:
##### Summary
Make cloud-enabled flag in web/api/v1/info be independent of ACLK build success
##### Component Name
web
##### Test Plan
Have ACLK build fail and check dashboard info page


jacekkolasa:
works good, finally! On my unsupported macOS, i install normally and see `"cloud-enabled": true`. Later I also checked the `--disable-cloud` install option and afterwards `cloud-enabled` is `false`, as it should be

PR #8863, opened 2020-04-30
Fix builds using the Clang compiler

prologic:
##### Summary

ssia

Fixes #8854

##### Component Name

- area/build
- area/pacakging
- area/ci

##### Test Plan

- [x] This PR
  - [x] First run on first ocmmit will fail
  - [x] Second run on second commit that fixes the problem will pass

##### Additional Information

PR #8867, opened 2020-04-30
Added docmentation about workaround for clang build errors.

Ferroin:
##### Summary

This is based on discussion today during the standup about the handling of issue #8854. It ideally should be merged before the 1.22 release.

##### Component Name

area/docs

##### Test Plan

n/a

##### Additional Information

I'm not 100% certain that the place I'm putting this documentation is the best one, but I can't think of anywhere better myself. If anyone from the documentation/marketing team has suggestions for a better place to put it in the docs, I'll be happy to move it.

Long-term, the issue this is documenting a workaround for is planned to be fixed, but we've decided that it should not be put in the upcoming release as a fix would require modifying things in a way that would require rigorous re-testing o the cloud functionality in the agent, which we really don't have time for at the moment. This should be removed/reverted once #8863 is merged.

### DIFF
--- a/packaging/installer/methods/manual.md
+++ b/packaging/installer/methods/manual.md
@@ -206,6 +206,10 @@ You can edit this file to set options. One common option to tweak is `history`, 
 
 To apply the changes you made, you have to restart Netdata.
 
+### 'nonrepresentable section on output' errors
+
+Our current build process unfortunately has some issues when using certain configurations of the `clang` C compiler on Linux. If the installation fails with errors like `/bin/ld: externaldeps/libwebsockets/libwebsockets.a(context.c.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIC` and you are trying to build with `clang` on Linux, you will need to instead build using GCC to get a fully functional install of Netdata. In most cases, you can do this by running `CC=gcc ./netdata-installer.sh`.
+
 ## What's next?
 
 When you finish installing Netdata, be sure to visit our [step-by-step tutorial](../../../docs/step-by-step/step-00.md)
--- a/packaging/installer/methods/manual.md
+++ b/packaging/installer/methods/manual.md
@@ -208,7 +208,11 @@ To apply the changes you made, you have to restart Netdata.
 
 ### 'nonrepresentable section on output' errors
 
-Our current build process unfortunately has some issues when using certain configurations of the `clang` C compiler on Linux. If the installation fails with errors like `/bin/ld: externaldeps/libwebsockets/libwebsockets.a(context.c.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIC` and you are trying to build with `clang` on Linux, you will need to instead build using GCC to get a fully functional install of Netdata. In most cases, you can do this by running `CC=gcc ./netdata-installer.sh`.
+Our current build process unfortunately has some issues when using certain configurations of the `clang` C compiler on Linux.
+
+If the installation fails with errors like `/bin/ld: externaldeps/libwebsockets/libwebsockets.a(context.c.o): relocation R_X86_64_32 against '.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIC`, and you are trying to build with `clang` on Linux, you will need to build Netdata using GCC to get a fully functional install. 
+
+In most cases, you can do this by running `CC=gcc ./netdata-installer.sh`.
 
 ## What's next?
 


nabijaczleweli:
NB: this does not work around the updater, which will still happily fault the cron-daily job and send mails, daily, for what is a relatively simple bug.

prologic:
I'm not really sure I find this a good answer to be honest. We're effectively telling our users that _want_ to use Clang to use GCC instead. Why would we do that?

prologic:
:> This is based on discussion today during the standup about the handling of issue #8854.

There was no mishandling; just poor communication on my part.
@amoss and I had a good discussion and we're on the same page.

> It ideally should be merged before the 1.22 release.

Since when? I've parted the PR as we agreed we would __not__ be merging this until _after_ the upcoming `v1.22` release (_which IIRC is next week?_).

Ferroin:
> :> This is based on discussion today during the standup about the handling of issue #8854.
> 
> There was no mishandling; just poor communication on my part.
> @amoss and I had a good discussion and we're on the same page.
> 
> > It ideally should be merged before the 1.22 release.
> 
> Since when? I've parted the PR as we agreed we would **not** be merging this until _after_ the upcoming `v1.22` release (_which IIRC is next week?_).

The discussion in question was about how to handle the issue WRT the release itself. The decision was that we should document the known workaround for the install itself and then worry about fixing things after the release.

packaging/installer/methods/manual.md
```diff
@@ -206,6 +206,10 @@ You can edit this file to set options. One common option to tweak is `history`,
 
 To apply the changes you made, you have to restart Netdata.
 
+### 'nonrepresentable section on output' errors
+
+Our current build process unfortunately has some issues when using certain configurations of the `clang` C compiler on Linux. If the installation fails with errors like `/bin/ld: externaldeps/libwebsockets/libwebsockets.a(context.c.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIC` and you are trying to build with `clang` on Linux, you will need to instead build using GCC to get a fully functional install of Netdata. In most cases, you can do this by running `CC=gcc ./netdata-installer.sh`.
```
joelhans, line 211: I don't have any objections to placing this workaround here vs. anywhere else. This suggestion is mostly formatting—there's a backtick in the example output that blows up the inline code blocks, so I just changed that to a straight apostrophe. I also broke this up into some paragraphs to help distinguish between problem and solution.

```suggestion
Our current build process unfortunately has some issues when using certain configurations of the `clang` C compiler on Linux.

If the installation fails with errors like `/bin/ld: externaldeps/libwebsockets/libwebsockets.a(context.c.o): relocation R_X86_64_32 against '.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIC`, and you are trying to build with `clang` on Linux, you will need to build Netdata using GCC to get a fully functional install. 

In most cases, you can do this by running `CC=gcc ./netdata-installer.sh`.
```

prologic:
> The discussion in question was about how to handle the issue WRT the release itself. The decision was that we should document the known workaround for the install itself and then worry about fixing things after the release.

But I do not consider telling our users to use GCC a workaround for users wanting to use Clang. Sorry.

nabijaczleweli:
I'm (unsurprisingly) with @prologic on this — while it's true that most systems do contain GCC, I see no reason to forbid Clang-only (or, indeed, Clang-default) systems only because Clang exposes a build system bug that GCC "helpfully" masks. Especially since this is easily fixed (going by #8863, a one-line change) by passing exactly the flag the compiler suggests.

Ferroin:
OK, I think I need to reiterate here, _this is a stopgap_. We _will_ be fixing the issue this is documenting, we don't have time right now to do all the testing to be certain that the fix doesn't break anything else (and no, I don't think it will and would be amazed if it did (because we do _nothing_ that breaks PIC requirements), but it's not my call).

Note, also, that this isn't just "Clang on Linux doesn't work". It's an issue caused by the particular configuration of Clang that Debian uses and the peculiarities of the LWS build infrastructure. With the various sanitizers and other bells and whistles that Debian chooses to build Clang with turned off, it works just fine without issue, but I'm not sure how to do that from the command-line, and it would still only fix things until you try to update.

@prologic If you really want to go to the mat over this, take it up with @amoss and @cakrit, they're the ones you really need to convince, not me (well, more accurately if you can manage to convince them, the same argument will probably work on me too).

prologic:
Given recent comments I've reluctantly approved this PR. for the record I still believe the message we're brining across isn't the one I had hoped for, but willing to accept we will fix Clang/LLVM builds prompt.y (_there is a PR in-progerss afterall_).

PR #8870, opened 2020-05-01
dashboard v1.0.13

jacekkolasa:
##### Summary
fix dygraphs after toggling sidebar
rich aclk-related notifications, with fixed conditionals
terms and privacy moved to right sidebar

https://github.com/netdata/dashboard/pull/77
After toggling left sidebar, dygraph chart was showing empty space, without time domain.

https://github.com/netdata/dashboard/pull/78
There was following bug - when a call for /data was pending, and user scrolled away, then that call was terminated and the chart was not fetching data again (very serious bug, fortunately it was not present in any nightly)

https://github.com/netdata/dashboard/pull/79
there are specs described in:
https://github.com/netdata/product/issues/277

Specifically, the following scenario should be fulfilled:
> when cloud-enabled === true and cloud-available === false  we'll display "Installation error" on every browser session (you'll see that always when you'll go to Dashboard)
when agent-claimed === true and aclk-available === false we'll display "Connection Problem"

Also, "terms and conditions" should be visible in the right-hand sidebar.

##### Component Name
Dashboard


thiagoftsm:
@jacekkolasa I am not sure whether we are talking about the same bug, but after I do a login and I resize the windows, I see this:

err47

So I am thinking that the "resize" described in the link netdata/dashboard#77 is not work properly, please, can you confirm if this wrong resize is related to this PR?

I used Firefox 68.7.0esr with `Private Window`.

jacekkolasa:
Thanks a lot @thiagoftsm! I've recreated that issue. I think we should merge this PR still, and create additional ticket for the bug you described.

I was fixing that issue but it looks like i will need to make one hack to our dygraph-chart. Fortunately that problem now exists only in specific scenario:
- it happens on Firefox (at least i couldn't recreate it in Chrome)
- you need to be logged in, and resize from width >1200px to <1200px. Any additional resizing will fix dygraphs

We can consider if this is release blocker. Apart from that, i will work on it today.

amoss:
@jacekkolasa This has four approvals - please merge.

Issue #8731, opened 2020-04-15
Modular eBPF plugin


thiagoftsm:
##### Feature idea summary
The current eBPF plugin is monolithic and we need to change it for we load more eBPF programs. To reach this goal we will follow the same pattern already adopted with `proc.plugin`, in this work the following steps will be done:

- Everything related to the current eBPF program inside `ebpf.c` will be moved to another source file (ebpf_process.c). This implies to move the chart creation function, functions used to read data from kernel ring to user ring and process the data, functions that send messages to Netdata. 
- We will have the following changes in the configuration files: 
- `ebpf.conf` will be the global configurator, but we will create configurations for specific eBPF programs, these files will be created inside `ebpf.d`. 
   -   Create a new section inside that will be used to define the ebpf programs that will be loaded ( `[ebpf programs]` ).
    -  The configurations defined inside the global section of `ebpf.conf` can be overwritten for the specific configurations inside `ebpf.d`.
   - Example:
```
[global]
   load = entry
[ebpf programs]
   process = yes
   network viewer = yes
```

-  Like `proc.plugin` we will also have a global structure that will have the functions to be called in each new file added.
- The ebpf programs can be loaded either using `ebpf.conf` or trough command line arguments.
- Right now we have 3 threads inside the `eBPF plugin` ( "Publisher", "Collector", Log ). We will have only one thread per eBPF program and we group the "Publisher" and "Collector" will be merged inside an unique thread.

##### Expected behavior
Organize the collector to bring new eBPF programs.

underhood:
One thing I would like is to also organize/structure data a bit better in the plugin itself instead of one big array with all the values.

thiagoftsm:
> One thing I would like is to also organize/structure data a bit better in the plugin itself instead of one big array with all the values.

I agree, I began to work on this in the PR #8879 where I am creating the basis for this PR. You can see there an array of structures that will be passed as arguments for the threads, be free to give any suggestion.

Issue #8727, opened 2020-04-15
New charts for apps.plugin from ebpf.plugin collector.


thiagoftsm:
##### Feature idea summary
We will keep the current global charts to demonstrate our users how all the software are interacting with kernel and hardware. These charts are also important for us, 
because they are showing that Netdata is using eBPF.
 
We will create two new submenus inside Apps.plugin `ebpf.syscall` and `ebpf.net`, these submenus will be used to organize the next  11 charts

#### ebpf.syscall:
- Files opened - Number of calls to open files per period.
- Files closed - Number of calls to close files per period.
- Files deleted - Number of calls to vfs_delet to remove files per period.
- Number of vfs write calls - Number of calls to vfs_write per period.
- Number of vfs read calls - Number of calls to vfs_write per period.
- Number of bytes written through vfs_write - Number of bytes written trough vfs_write per period.
- Number of bytes read through vfs_read - Number of bytes read trough vfs_read per period.
- Number of processes created - Number of process created trough do_fork
- Number of threads created - Number of process created trough do_fork or __x86_64_sys_clone depending of the kernel version.
- Number of closed process - Number of times that a process called do_exit.
#### ebpf.net: 
- Bandwidth per process - Number of bytes transmited per seconds.

Netdata is also capable to demonstrate errors for syscall that returns values, so when users are running the ebpf collector as `return` mode, they will also have the following charts:

#### ebpf.syscall:
- Open error: Number of calls to open files that returned errors.
- Close error: Number of calls to close files that returned errors.
- VFS read error: Number of calls to read a file that returned error.    
- VFS write error: Number of calls to read a file that returned error.

For all these charts the dimensions will be the process group.
##### Expected behavior
When eBPF is completely integrated with apps.plugin, we will have more 15 chars given deeper details about process running.

Issue #7873, opened 2020-01-28
eBPF process and Apps.plugin integration


thiagoftsm:
##### Feature idea summary
We already have a global vision of what is happening with process, but it would be interesting to have the same vision per process group, so we need to do integration with eBPF plugin.

For while we are working with these possibilities to move in front with the integration:

1 - Shared Memory: Here we will create a common memory region between `apps.plugin` and `ebpf.plugin`. It will be necessary to use locks to update the values that will come from `ebpf.plugin`, here I am assuming that `apps.plugin` will be responsible to make all the charts, for more details read [this](https://github.com/netdata/netdata/issues/7873#issuecomment-618132093).

2 - Other IPCs: We can select another IPC that will allow `ebpf.plugin` to send data to `apps.plugin` that will generate the charts. [IPCs](https://github.com/netdata/netdata/issues/7873#issuecomment-619633734)

3 - Shared library: We could move outside `apps.plugin` the common code for both collectors to a specific shared library that will be used from both collectors.  For more details about the functions will need to be moved to the Shared library go to [this link](https://github.com/netdata/netdata/issues/7873#issuecomment-600957047)

##### Expected behavior
To define the road to generate new charts for the apps groups.

amoss:
@thiagoftsm Please explain the different options in this issue so that we can have an asynchronous discussion leading up to the grooming meeting.

thiagoftsm:
With the conclusion of #8112 and the results that we got on #7987 we need to define how we will do the integration between the collectors.
The eBPF program won't be able to have the same charts that we have at apps.plugins, but `apps.plugin` has a logic that we need for the `ebpf_program.plugin`, so we need to define the road to do this "integration".

thiagoftsm:
I will present few scenarios, firstly let us suppose that we want to have an eBPF plugins completely independent of apps.plugin, in this scenario we will have a hash table inside the collector where we will store the relationship between TGIDs and PIDs. We will also have two other hash tables inside the collector, one to store the PID statistic and other to store TGID statistic. We will have the following algorithm to update the tables:

1 - When the function is called we read the necessary values.
2 - After this we split the values for TGID and PID that are stored together, we can separate them using bitwise.
3 - We search inside the hash table that stores PID the previous statistic values for we update, case it is not present we create it. It is interesting when we create the PID information, we also create the other two rows inside the other hash tables.
4 - Finally we get the statistic values for TGID and we update/create it like the previous steps.

[...]

PR #8879, opened 2020-05-02
Ebpf options

### LINKED ISSUE #8716: Extend options accepted by ebpf_process.plugin
##### Feature idea summary
Right now eBPF plugin only accepts as argument the interval that it will send messages to Netdata. We need to extend the options that it accepts, considering the next steps with this collector, the following argument needed to be parsed for the collector:

- `-h` or `--help` or `help` : Show an information about the options available.
- `-v` or `--version` or `version`: The current version of the collector.
- `-g` or `--global` or `global`: Only create global charts.
- `-a` or `--all` or `all` : Show global charts and charts per process group.
- `-n` or `--net`or `net`: Show only charts related to network.
- `-p` or `--process` or `process`: Show only charts related to process.
- `-r` or `--return` or `return`: Change the collector to return mode. 

##### Expected behavior
Bring general options as arguments for the collector.

thiagoftsm:
##### Summary
Fixes #8716 

This PR brings the features described on #8716 and also brings the basis to develop #8731, #7873 and #8727.
##### Component Name
Collector
##### Test Plan
Compile this PR with the option `-DNETDATA_INTERNAL_CHECKS=1` and add the following options inside your `netdata.conf`:

```
[plugin:ebpf_process]
        # update every = 1
        command options = --global --all --net --process --return
        #command options = -a -r
        #command options = -a -n -p -r
        #command options = -g -n -p -r
```

When Netdata starts it is expected it will show messages related the option given.
There are two options that needs to be tested using terminal:

```bash
# ./ebpf_process.plugin --help
# ./ebpf_process.plugin --version
```

### DIFF
--- a/collectors/ebpf_process.plugin/ebpf_process.c
+++ b/collectors/ebpf_process.plugin/ebpf_process.c
@@ -874,9 +874,9 @@ static inline void ebpf_disable_apps() {
     }
 }
 
-static inline void ebpf_enable_specific_chart(struct  ebpf_module *em, int apps) {
+static inline void ebpf_enable_specific_chart(struct  ebpf_module *em, int disable_apps) {
     em->enabled = 1;
-    if (!apps) {
+    if (!disable_apps) {
         em->apps_charts = 1;
     }
     em->global_charts = 1;
@@ -889,14 +889,24 @@ static inline void ebpf_enable_all_charts(int apps) {
     }
 }
 
-static inline void ebpf_enable_chart(int enable, int apps) {
+static inline void ebpf_enable_chart(int enable, int disable_apps) {
     int i ;
     for (i = 0 ; ebpf_modules[i].thread_name ; i++ ) {
         if (i == enable) {
-            ebpf_enable_specific_chart(&ebpf_modules[i], apps);
+            ebpf_enable_specific_chart(&ebpf_modules[i], disable_apps);
             break;
         }
     }
+}
+
+static inline void ebpf_set_thread_mode(int mode) {
+    int i ;
+    for (i = 0 ; ebpf_modules[i].thread_name ; i++ ) {
+        ebpf_modules[i].mode = mode;
+    }
+}
+
+void ebpf_print_help() {
 }
 
 static void parse_args(int argc, char **argv)
@@ -914,6 +924,11 @@ static void parse_args(int argc, char **argv)
             }
         }
 
+        if (strcmp("help", w) == 0 || strcmp("-help", w) == 0 || strcmp("--help", w) == 0 || strcmp("-h", w) == 0 || strcmp("-H", w) == 0) {
+            ebpf_print_help();
+            exit(0);
+        }
+
         if (strcmp("version", w) == 0 || strcmp("-version", w) == 0 || strcmp("--version", w) == 0 || strcmp("-v", w) == 0 || strcmp("-V", w) == 0) {
             printf("ebpf.plugin %s\n", VERSION);
             exit(0);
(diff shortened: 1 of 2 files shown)


collectors/ebpf_process.plugin/ebpf_process.c
```diff
@@ -853,11 +867,164 @@ static int load_collector_file(char *path) {
...
+        char *w = argv[i];
+        if (!freq) {
+            int n = (int)str2l(w);
+            if(n > 0) {
+                freq = n;
+                continue;
+            }
+        }
 
+        if (strcmp("help", w) == 0 || strcmp("-help", w) == 0 || strcmp("--help", w) == 0 || strcmp("-h", w) == 0 || strcmp("-H", w) == 0) {
```
mfundul, line 960: Maybe do something like this?
https://github.com/netdata/netdata/blob/166958349599708ae34397a8407aa35982f8c81c/daemon/main.c#L912

thiagoftsm: I do not have any objection to use `getopt` or `getopt_long`, I used this algorithm, because I imagined we had a different pattern for plugins:

https://github.com/netdata/netdata/blob/166958349599708ae34397a8407aa35982f8c81c/collectors/apps.plugin/apps_plugin.c#L3855

https://github.com/netdata/netdata/blob/166958349599708ae34397a8407aa35982f8c81c/collectors/cups.plugin/cups_plugin.c#L117

mfundul: We can keep this as is for now, but we should keep in mind that the ebpf plugin will grow significantly and at some point we should refactor this anyway.

thiagoftsm: I convert to `getopt_long`, so we do not need to change in the future.

underhood:
The `--help` can be used only as root
```
aaaa@aaaa-VirtualBox:~/netdata$ ./ebpf_process.plugin --help
2020-05-06 08:42:57: ebpf_process.plugin ERROR : MAIN : [EBPF PROCESS] setrlimit(RLIMIT_MEMLOCK) (errno 1, Operation not permitted)
aaaa@aaaa-VirtualBox:~/netdata$ sudo ./ebpf_process.plugin --help

 Netdata ebpf.plugin v1.21.1-138-g4ce22b35
 Copyright (C) 2016-2020 Costa Tsaousis <costa@tsaousis.gr>
 Released under GNU General Public License v3 or later.
 All rights reserved.

 This program is a data collector plugin for netdata.

 Available command line options:

 SECONDS           set the data collection frequency.

 --help or -h      show this help.

 --version or -v   show software version.

 --global or -g    disable charts per application.

 --all or -a       Enable all chart groups (global and apps), unless -g is also given.

 --net or -n       Enable network viewer charts.

 --process or -p   Enable charts related to process run time.

 --return or -r    Run the collector in return mode.
```

collectors/ebpf_process.plugin/ebpf_process.c
```diff
@@ -853,11 +867,185 @@ static int load_collector_file(char *path) {
...
+            case 'p': {
+                enabled = 1;
+                ebpf_enable_chart(0, disable_apps);
+#ifdef NETDATA_INTERNAL_CHECKS
+                info("EBPF enabling \"PROCESS\" charts, because it was started with the option \"--process\" or \"-p\".");
+#endif
+                break;
+            }
+            case 'r': {
+                mode = 0;
```
underhood, line 1022: can we use enum here or at least define? For somebody who not have an overview of the code, it might not be obvious what mode = X means *(well it is clear here as there is info message with `return mode`)*.

underhood: No need to change it in this PR but for future i think it is something to improve

thiagoftsm: The last commit brought all changes you required @underhood. Thank you for the suggestions!

PR #8887, opened 2020-05-04
Go live

amoss:
##### Summary
The go-live feature branch to enable the cloud launch.

##### Component Name
Cloud

##### Test Plan
N/A


netdata-installer.sh
```diff
@@ -578,7 +581,7 @@ bundle_libwebsockets() {
     return 0
   fi
 
-  if [ -z "$(command -v cmake)" ] ; then
+  if [ -z "$(command -v cmake)" ]; then
```
prologic, line 584: Not sure where these lines came from but this is not the right way to test for the existence of a command on the system. We should isntead use:

```#!sh
if command -v "$command" >/dev/null 2>&1; then
  echo "$command exists"
fi
```

netdata-installer.sh
```diff
@@ -621,10 +624,77 @@ bundle_libwebsockets() {
...
+  run cp ${1}/*.h "${target_dir}/json-c" || return 1
+}
+
+bundle_jsonc() {
+  # If --build-json-c flag or not json-c on system, then bundle our own json-c
+  if [ -z "${NETDATA_BUILD_JSON_C}" ] && pkg-config json-c; then
+    return 0
+  fi
+
+  if [ -z "$(command -v cmake)" ]; then
```
prologic, line 651: Same here.

PR #8893, opened 2020-05-04
[WIP] Improved handling of page creation when storing metrics with older timestamps

### LINKED ISSUE #8384: Repeated dbengine global flushing errors for statsd
Hi,

I am having the same issue in v1.20 with the following setup : 

- 10 headless, memory = none collectors, streaming to a central colector
- Around 5k metrics / collector => ~50k metrics on the central collector
- Central collector is running in a lxc container (proxmox), storage is on a dedicated SSD (memory=dbengine)

- OS / Environment
> Linux grafana 4.15.18-12-pve #1 SMP PVE 4.15.18-35 (Wed, 13 Mar 2019 08:24:42 +0100) x86_64 GNU/Linux Debian 9.5 running in a LXC container (Proxmox) - 32GB of RAM (7GB used, the rest is used by the OS for caching), 8 CPU cores peak use 10%

- Netdata Version
>v1.20.0

- Component  Name
>dbengine

In error.log I get the messages : 
```
2020-03-11 10:04:09: netdata ERROR : MAIN : Failed to flush dirty buffers quickly enough in dbengine instance "/opt/netdata/var/cache/netdata/e295a868-5042-11e7-9bce-3863bb3e6c28/dbengine". M
etric data are being deleted, please reduce disk load or use a faster disk.
2020-03-11 10:04:09: netdata ERROR : MAIN : Failed to flush dirty buffers quickly enough in dbengine instance "/opt/netdata/var/cache/netdata/e295a868-5042-11e7-9bce-3863bb3e6c28/dbengine". M
etric data are being deleted, please reduce disk load or use a faster disk.
2020-03-11 10:04:09: netdata ERROR : MAIN : Failed to flush dirty buffers quickly enough in dbengine instance "/opt/netdata/var/cache/netdata/e295a868-5042-11e7-9bce-3863bb3e6c28/dbengine". M
etric data are being deleted, please reduce disk load or use a faster disk.
2020-03-11 10:04:09: netdata ERROR : MAIN : Failed to flush dirty buffers quickly enough in dbengine instance "/opt/netdata/var/cache/netdata/e295a868-5042-11e7-9bce-3863bb3e6c28/dbengine". M
etric data are being deleted, please reduce disk load or use a faster disk.
2020-03-11 10:04:09: netdata ERROR : MAIN : Failed to flush dirty buffers quickly enough in dbengine instance "/opt/netdata/var/cache/netdata/e295a868-5042-11e7-9bce-3863bb3e6c28/dbengine". M
etric data are being deleted, please reduce disk load or use a faster disk.
```

I also get alarms : 

image

Db engine related metrics : 
image

Storage metric for the volume used by the netdata central collector (the one having memory=dbengine), which seems to indicate the disk is not overloaded: 
image
image

Still on the central collector : 
image

```

[backend]
	enabled = yes
	type = graphite
	destination = tcp:127.0.0.1:2003
	data source = as-collected
	prefix = netdata
	update every = 30
	buffer on failures = 10000
	timeout ms = 20000


[global]
	update every = 1
	memory mode = dbengine
	page cache size = 192
	dbengine disk space = 2048
	errors flood protection period = 10
	errors to trigger flood protection = 100000


[statsd]
	decimal detail = 1
	update every (flushInterval) = 1
	max private charts allowed = 100000
	max private charts hard limit = 100001
	private charts memory mode = ram
	private charts history = 600
	private charts hidden = yes
	add dimension for number of events received = no
	gaps on gauges (deleteGauges) = yes
	listen backlog = 16384
	bind to = udp:*:8125 tcp:*:8125

```

_Originally posted by @jbfuzier in https://github.com/netdata/netdata/issues/7468#issuecomment-597529668_

stelfrag:
Fixes #8384 
##### Summary
Detect attempt to store metric in an older point in time and avoid extending the current page length if not necessary. 

Extending the page length on every metric store will eventually result in a need to allocate a new page. When a lot of metrics are received for chart(s) dimensions that have not been updated within the history retain period, the interpolation algorithm will attempt to store a lot of metrics eventually
creating a lot of unneeded pages.

##### Component Name
dbengine

##### Test Plan
- Setup a master - slave configuration
   - Stream metrics from slave to master periodically in such a way that there are gaps in the data collection
      - Data needs to be collected and streamed within the specified history retention defined in the master
   - Data interpolation will create extra pages

### DIFF
--- a/database/engine/rrdengineapi.c
+++ b/database/engine/rrdengineapi.c
@@ -144,6 +144,25 @@ void rrdeng_store_metric_next(RRDDIM *rd, usec_t point_in_time, storage_number n
     descr = handle->descr;
 
     if (descr) {
+
+        /*
+         * Decide if we try to store in the current page in descr->[start_time, end_time] and if so
+         * store the value directly without extending the page_length
+         */
+        if (unlikely(point_in_time < descr->start_time)) {
+            debug(D_RRDENGINE, "Metric not in the current page; discarding");
+            return;
+        }
+
+        if (unlikely(point_in_time >= descr->start_time && point_in_time <= descr->end_time)) {
+            uint64_t metric_index;
+            page = descr->pg_cache_descr->page;
+            metric_index = (uint64_t) (point_in_time - descr->start_time) / (USEC_PER_SEC * sizeof(number));
+            page[metric_index] = number;
+            debug(D_RRDENGINE, "Overwriting metric at index %"PRIu64, metric_index);
+            return;
+        }
+
         /* Make alignment decisions */
 
         if (descr->page_length == rd->rrdset->rrddim_page_alignment) {
--- a/database/engine/rrdengineapi.c
+++ b/database/engine/rrdengineapi.c
@@ -150,16 +150,26 @@ void rrdeng_store_metric_next(RRDDIM *rd, usec_t point_in_time, storage_number n
          * store the value directly without extending the page_length
          */
         if (unlikely(point_in_time < descr->start_time)) {
-            debug(D_RRDENGINE, "Metric not in the current page; discarding");
+            errno = 0;
+            error("Metric not in the current page; discarding");
             return;
         }
 
-        if (unlikely(point_in_time >= descr->start_time && point_in_time <= descr->end_time)) {
+        if (unlikely(point_in_time <= descr->end_time)) {
             uint64_t metric_index;
+            unsigned entries;
+
+            if (likely(descr->start_time != descr->end_time)) {
+                entries = descr->page_length / sizeof(storage_number);
+
+                metric_index = ((uint64_t)(point_in_time - descr->start_time)) * (entries - 1) /
+                               (descr->end_time - descr->start_time);
+            } else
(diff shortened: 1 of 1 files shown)


stelfrag:
Closing in favor of  #9738

mfundul:
Let's keep this open for now, this is very useful code and should be merged at some point.

amoss:
I think that we should not do this. It changes one of the (only?) fundamental constraints that the database relies on - that storage requests are on monotonically increasing timestamps. The effects of changing that are unpredictable and may cause problems in the future.

If we were to relax this constraint and allow storage of older data then it would need to be done in such a way that the (internal) view of pages was not exposed to the caller, as otherwise it is leaking dbengine internals into the rest of the agent. The caller should not need to know the position of the data on the page to know if the call will succeed or fail.

I'm about to add something similar to the replication branch that handles implicit flushing if the sequence of timestamps changes. Please take a look at that and then we can decide if it is a good idea to merge this change.

ktsaou:
@stelfrag this is very old. Either merge or close.

PR #8904, opened 2020-05-05
dashboard v1.0.14

jacekkolasa:
##### Summary
new Dashboard version

https://github.com/netdata/dashboard/pull/80
- fixes situation where when loading snapshot easypiechart was resulting in error

- better iframe-load-error detection for firefox (fixes unwanted lack of sign-in button)

This was previously broken when:
1) user signed in to cloud
2) cloud was inaccessible (for example by setting invalid cloud base url)
3) refreshing the page was showing nothing in place of sign in button (instead, a disabled button looking like loader should be shown)


underhood:
This is what i see now in FF (not sure if this is success or failure?):
Screenshot from 2020-05-05 11-24-17

FF private tab and chrome show sign-in button as usual

jacekkolasa:
Yes, @underhood , that is the design - if you were previously logged in, and after that you're offline, then the "Sign-in" and "+" buttons should be disabled.

PR #8913, opened 2020-05-06
Account for zfs.arc_size.min, and correct calc

### LINKED ISSUE #7871: "System RAM used" metric uses confusing denominator in formula
##### Bug report summary
Note: this is NOT #3381 (but involves the same metric) - this issue refers to the metric being confusing even outside of a container.

---

The formula for calculating the memory use is either wrong, or at the very least confusing.
`system.ram.ram_in_use` uses the following expression:
```bash
($used - $used_ram_to_ignore) * 100 / ($used - $used_ram_to_ignore + $cached + $free)
```
with `$used_ram_to_ignore` being the ZFS arc size, and documented as "memory that is reported as used, but [can be resized on demand]".

On my system the ZFS arc takes just under half of my RAM, which makes the ram_in_use alarm trigger eagerly.

I believe that `used_ram_to_ignore` should not be subtracted in the denominator, so `ram_in_use` represents the percentage of system memory that cannot be given to other programs.

(Also, I'm pretty sure that `used_ram_to_ignore` should be the difference between the current arc size and the min size of the arc, instead of the whole arc size.)

##### OS / Environment
Linux ceres 4.19.75 #1-NixOS SMP Sat Sep 21 05:17:15 UTC 2019 x86_64 GNU/Linux
NixOS 20.03pre194079.82fe35cd29b (Markhor)

##### Netdata version
netdata v1.17.1
 
##### Component Name
health

##### Steps To Reproduce
I discovered this bug when digging into why I was getting more alarms after updating netdata, and noticing that the calculated percentage on the "ram in use" widget was off.
Steps to reproduce:
1. Run netdata on a system with a large ZFS ARC in relation to normal memory.
2. Open netdata and find the "system - ram" "ram in use" alarm in the "All alarms" section.
3. Notice wonky reported values.

##### Expected behavior
I expected the calculated value to be closer to the amount of used RAM that cannot be automatically freed.

araemo:
##### Summary
ram_in_use was not correctly deducting the used_ram_to_ignore -
it deducted it from both the 'used' and 'unused' side of the equation,
which made a large ZFS ARC (like all of them) report high ram
utilization all the time, firing off the 90% alert despite 40% of the used
ram being reclaimable cache.

Fixes #7871

##### Component Name
health.d/ram.conf

##### Test Plan
Review the system - ram alarm on a system with ZFS which has been running long enough to fill up the ZFS arc.  Without this patch, you see the entire size of the arc reflected in this alarm as unavailable RAM.  With this patch, the ZFS arc (minus the minimum size) is treated as available cached memory when calculating this alert.

### DIFF
--- a/health/health.d/ram.conf
+++ b/health/health.d/ram.conf
@@ -5,7 +5,7 @@
       on: system.ram
       os: linux freebsd
    hosts: *
-    calc: ($zfs.arc_size.arcsz = nan)?(0):($zfs.arc_size.arcsz)
+    calc: ($zfs.arc_size.arcsz = nan)?(0):($zfs.arc_size.arcsz - $zfs.arc_size.min)
    every: 10s
     info: the amount of memory that is reported as used, but it is actually capable for resizing itself based on the system needs (eg. ZFS ARC)
 
@@ -14,7 +14,7 @@
       os: linux
    hosts: *
 #   calc: $used * 100 / ($used + $cached + $free)
-    calc: ($used - $used_ram_to_ignore) * 100 / ($used - $used_ram_to_ignore + $cached + $free)
+    calc: ($used - $used_ram_to_ignore) * 100 / ($used  + $cached + $free)
    units: %
    every: 10s
     warn: $this > (($status >= $WARNING)  ? (80) : (90))


thiagoftsm:
Hi @araemo ,

Firstly thank you for your contribution!
Please, could you sign the [CLA](https://github.com/netdata/netdata/pull/8913#issuecomment-624394339) for we have conditions to merge your PR.

best regards!

araemo:
Done.

vlvkobal:
@Ferroin please have a look.

PR #8927, opened 2020-05-06
Change the name of the connect message to update during an ongoing se…

amoss:
##### Summary
Change the name of the connect message during a session as agreed with @cosmix and @fracasula 

##### Component Name
ACLK

##### Test Plan
No testing - review only.

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -1684,7 +1684,7 @@ int aclk_send_info_metadata()
     // a fake on_connect message then use the real timestamp to indicate it is within the existing
     // session.
     if (aclk_metadata_submitted == ACLK_METADATA_SENT)
-        aclk_create_header(local_buffer, "connect", msg_id, 0, 0);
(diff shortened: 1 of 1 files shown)


aclk/agent_cloud_link.c
```diff
@@ -1684,7 +1684,7 @@ int aclk_send_info_metadata()
     // a fake on_connect message then use the real timestamp to indicate it is within the existing
     // session.
     if (aclk_metadata_submitted == ACLK_METADATA_SENT)
-        aclk_create_header(local_buffer, "connect", msg_id, 0, 0);
+        aclk_create_header(local_buffer, "update", msg_id, 0, 0);
```
fracasula, line 1687: This isn't going to change also the regular `connect` to `update`, right? Sorry didn't check the whole thing, I'm asking just in case :slightly_smiling_face:

amoss: No, the regular connect is the else branch. I asked @stelfrag to review as a check that I changed the right case.

PR #8956, opened 2020-05-11
update docs to Caddy v2

slhck:
##### Summary

Since Caddy v2 is released and v1 is no longer the default, update the docs and link.

##### Component Name

Docs


slhck:
This does not seem to work right, closing for now.

Issue #8990, opened 2020-05-12
Remove old docs generation scripts and CI checks


joelhans:
We need to remove all of the scripts, Netlify deploy previews, and CI checks related to building the old documentation site, which no longer exists. This is already interrupting the process of merging PRs, since the Netlify checks inevitably fail for one reason or another.

Here is my plan for removing this old logic:

1. **Remove scripts to build translations**. We're already having issues with the localization project causing build failures, and no one should be spending time maintaining these files. A broken link in the localization repo means nothing, since those files are not published anywhere.
2. **Clean up the `buildhtml.sh` script** so that it's as simple as possible. It doesn't need to run `mkdocs` if it's only checking for broken links.
3. **Convert the link-checking script into its own CI action** that can verify whether links are structured correctly in the documentation. This will ensure we don't merge PRs in netdata/netdata that contain broken links.
4. **Wind down localization project**. Add a notice to the README that it's not being used/maintained any more. Search the docs for any mentions of it, and remove.
5. **Remove all the old docs generation tooling**, minus the link-checking CI, in a fashion similar to what @prologic wanted to do with #8783.

Here's the CI currently exists on the Learn site:

1. **A daily job checking for broken internal links**. This check uses [broken-link-checker](https://www.npmjs.com/package/broken-link-checker) by running `npx broken-link-checker $WEBSITE_URL --ordered --recursive --exclude-external` to check all _internal_ links. If it finds any broken internal links, it creates an issue with the output so I know which link to fix.
2. **A weekly job checking for broken external links**. This never existed with the old docs site, so it's actually an enhancement over what we had. This runs `npx broken-link-checker $WEBSITE_URL -rof --exclude-internal --requests=2 --exclude "netdata.cloud/" --exclude "https://www.linkedin.com/company/my-netdata.io/" --exclude "https://github.com/" --exclude "my-netdata.io" --exclude "http://localhost" --exclude "firehol.org" --exclude "circleci.com" --exclude "redis.io" --exclude "telegram.org" --exclude "t.me"` to check all _external_ links, and again creates an issue if it finds any.
3. **A Markdown link checking GitHub Action on every PR**. This will catch any additional link errors that somehow escaped the initial checks in netdata/netdata.

And a plan for additional Learn CI:

1. **Automated ingest/deployment of changed docs**. I currently do the process manually, but we can automate it. It can run every day, every hour, or on manual trigger.
[...]

PR #8783, opened 2020-04-21
Remove old docs generation tooling

prologic:
##### Summary

ssia

##### Component Name

- area/docs

##### Test Plan

None -- Migration happened already.

##### Additional Information

PR #8991, opened 2020-05-12
Remove translation building/link-checking from docs generator

joelhans:
##### Summary

The first step in #8990. This PR removes the logic to build files from the localization project and check their links. These checks are not necessary any more and only cause erroneous failures on PRs.

##### Component Name

docs

##### Test Plan

Tested by running the `buildhtml.sh` script locally, as I used to do when testing docs changes. It builds fine as-is, and when I added an intentionally-broken link, it showed a warning and failed as expected.

And the Netlify checks do not end in failure.

### DIFF
--- a/docs/generator/buildhtml.sh
+++ b/docs/generator/buildhtml.sh
@@ -52,10 +52,10 @@ for f in "${EXCLUDE_LIST[@]}"; do
 	rm "${SRC_DIR}/$f"
 done
 
-echo "Fetching localization project"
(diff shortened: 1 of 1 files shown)


prologic:
> And the Netlify checks do not end in failure.

Oh so this PR fixes the Netlify preview? But but but it doesn't look the same as when I go to docs.netdata.cloud / learn.netdata.cloud so Hmmmm 🤔

joelhans:
Closing, as we'll move ahead with #8783 instead when the SRE team is ready.

PR #8992, opened 2020-05-12
Fix the latency issue on the ACLK and suppress the diagnostics

amoss:
##### Summary
Fix a latency issue on the ACLK.

The on-connect payloads were large enough to trigger a massive increase in latency on the link and prevent chart updates due to head-of-line blocking. The default window detection in libwebsockets was under-reporting the size of the available window in the network. Overwritten with some sensible values.

The large volume of ACLK per-message info-logging is not produced unless the agent is compiled with NETDATA_INTERNAL_CHECKS. The logging now includes latency measurements on the link.

##### Component Name
ACLK

##### Test Plan

Connected / disconnected multiple times to staging / production. Checked the table view and chart view of the agent.

### DIFF
--- a/aclk/mqtt.c
+++ b/aclk/mqtt.c
@@ -13,6 +13,10 @@ inline const char *_link_strerror(int rc)
     return mosquitto_strerror(rc);
 }
 
+#ifdef NETDATA_INTERNAL_CHECKS
+static struct timeval sendTimes[1024];
+#endif
+
 static struct mosquitto *mosq = NULL;
 
 
@@ -29,8 +33,14 @@ void publish_callback(struct mosquitto *mosq, void *obj, int rc)
     UNUSED(mosq);
     UNUSED(obj);
     UNUSED(rc);
-    info("Publish_callback: mid=%d", rc);
-    // TODO: link this with a msg_id so it can be traced
+#ifdef NETDATA_INTERNAL_CHECKS
+    struct timeval now, *orig;
+    now_realtime_timeval(&now);
+    orig = &sendTimes[ rc & 0x3ff ];
+    int64_t diff = (now.tv_sec - orig->tv_sec) * USEC_PER_SEC + (now.tv_usec - orig->tv_usec);
+
+    info("Publish_callback: mid=%d latency=%dms", rc, diff / 1000);
+#endif
     return;
 }
 
@@ -320,6 +330,7 @@ int _link_subscribe(char *topic, int qos)
 int _link_send_message(char *topic, unsigned char *message, int *mid)
 {
     int rc;
+    size_t write_q, write_q_bytes, read_q;
 
     rc = mosquitto_pub_topic_check(topic);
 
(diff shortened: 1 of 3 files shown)


aclk/mqtt.c
```diff
@@ -320,16 +330,27 @@ int _link_subscribe(char *topic, int qos)
...
         return rc;
 
     int msg_len = strlen((char*)message);
-    info("Sending MQTT len=%d starts %02x %02x %02x", msg_len, message[0], message[1], message[2]);
+    lws_wss_check_queues(&write_q, &write_q_bytes, &read_q);
     rc = mosquitto_publish(mosq, mid, topic, msg_len, message, ACLK_QOS, 0);
 
+#ifdef NETDATA_INTERNAL_CHECKS
+    char msgHead[64];
+    memset(msgHead,0,sizeof(msgHead));
```
thiagoftsm, line 346: I know it is not the highest priority for this PR, so we need to have another PR to put spaces between variables in this new block.

thiagoftsm:
@amoss  we are having various warnings when we compile it

```
In file included from aclk/../daemon/../libnetdata/libnetdata.h:313,
                 from aclk/../daemon/common.h:6,
                 from aclk/mqtt.c:4:
aclk/mqtt.c: In function 'publish_callback':
aclk/mqtt.c:42:10: warning: format '%d' expects argument of type 'int', but argument 6 has type 'int64_t' {aka 'long int'} [-Wformat=]
   42 |     info("Publish_callback: mid=%d latency=%dms", rc, diff / 1000);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~      ~~~~~~~~~~~
      |                                                            |
      |                                                            int64_t {aka long int}
aclk/../daemon/../libnetdata/log/log.h:86:71: note: in definition of macro 'info'
   86 | #define info(args...)    info_int(__FILE__, __FUNCTION__, __LINE__, ##args)
      |                                                                       ^~~~
aclk/mqtt.c:42:45: note: format string is defined here
   42 |     info("Publish_callback: mid=%d latency=%dms", rc, diff / 1000);
      |                                            ~^
      |                                             |
      |                                             int
      |                                            %ld
aclk/mqtt.c: In function '_link_send_message':
aclk/mqtt.c:347:21: warning: pointer targets in passing argument 2 of 'strncpy' differ in signedness [-Wpointer-sign]
  347 |     strncpy(msgHead,message,48);
      |                     ^~~~~~~
      |                     |
      |                     unsigned char *
In file included from /usr/include/features.h:450,
                 from /usr/include/malloc.h:22,
                 from aclk/../daemon/../libnetdata/libnetdata.h:34,
                 from aclk/../daemon/common.h:6,
                 from aclk/mqtt.c:4:
/usr/include/bits/string_fortified.h:103:1: note: expected 'const char * restrict' but argument is of type 'unsigned char *'
  103 | __NTH (strncpy (char *__restrict __dest, const char *__restrict __src,
```

aclk/mqtt.c
amoss: Already done it ;)

PR #8993, opened 2020-05-12
Allow ebpf_process.plugin to set rlimit

roedie:
##### Summary
Without this capability the plugin will not start and fail with:

ebpf_process.plugin ERROR : MAIN : [EBPF PROCESS] 
setrlimit(RLIMIT_MEMLOCK) (errno 1, Operation not
 permitted)

##### Component Name
- ebpf
-packaging

### DIFF
--- a/contrib/debian/netdata.postinst.in
+++ b/contrib/debian/netdata.postinst.in
@@ -54,6 +54,7 @@ case "$1" in
     chown -R root:netdata /usr/libexec/netdata/plugins.d
     chown -R root:netdata /var/lib/netdata/www
     setcap cap_dac_read_search,cap_sys_ptrace+ep /usr/libexec/netdata/plugins.d/apps.plugin
+    setcap cap_sys_resource+ep /usr/libexec/netdata/plugins.d/ebpf_process.plugin
 
     ;;
 esac


MrFreezeex:
AFAIK the netdata debian package doesn't install natively the ebpf plugin, so you should at least check for the existence of the file beforehand.

As this is required to be installed manually I don't know if it worth putting this in the package...

thiagoftsm:
Hi @MrFreezeex ,

Netdata 1.23 will bring ebpf_process.plugin as a default plugin.

Best regards!

thiagoftsm:
Firstly, thank you very much @roedie !

@prologic and @Ferroin the PR brings changes for our Debian packages, on Slackware I do not have any problem like this. I cannot see any problem to add this command line to generate the packages, please, can you take a look on this PR? Another doubt, should we do something analogous in the future for RPMs?

roedie:
Just te be sure, I did a default build with `dpkg-buildpackage -rfakeroot -us -uc` on the v1.22.0 tag  and the ebpf_process.plugin file was present in the deb package. So, there doesn't seem to be a need for the check.

roedie:
Just looking at this, this might also be fixed by PR #8996 when the suid bit is set. Although, I don't like setting the suid bit on binaries if they don't need it, but that might be a personal preference.

MrFreezeex:
> Just te be sure, I did a default build with `dpkg-buildpackage -rfakeroot -us -uc` on the v1.22.0 tag and the ebpf_process.plugin file was present in the deb package. So, there doesn't seem to be a need for the check.

Indeed, It was the case for 1.21.x, sorry for that...

> Just looking at this, this might also be fixed by PR #8996 when the suid bit is set. Although, I don't like setting the suid bit on binaries if they don't need it, but that might be a personal preference.

I totally agree with you, I would rather have your change than the suid bit... I will update my PR in consequence if this is merged.

roedie:
@MrFreezeex Now, I don't know if ebpf_process.plugin needs root access or not, so the suid fix might be totally valid in this case. I did not cross my mind the plugin might need root permissions.

thiagoftsm:
> @MrFreezeex Now, I don't know if ebpf_process.plugin needs root privileges or not, so the suid fix might be totally valid in this case. I did not cross my mind the plugin might need root permissions.

eBPF needs root privileges, because it calls `bpf()` syscall and it also needs to change  `/sys/kernel/debug/tracing/kprobe_events`, case we do this without root permission, the kernel returns access denied.

Best regards!

roedie:
In that case I'm closing this PR, please use #8996.

thiagoftsm:
Thank you one more time @roedie !

PR #8994, opened 2020-05-12
Remove mkdocs from docs generator

joelhans:
##### Summary

The second step in #8990, to be merged after #8991. This PR removes the logic to run/build the old docs site using `mkdocs` and simplifies the `buildhtml.sh` script so that it essentially only sets up a clean directory tree and runs the link-checking script against it.

##### Test Plan

Again, I ran `buildhtml.sh` locally and it works fine. When I add an intentionally-broken link, the script returns with an error.

### DIFF
--- a/docs/generator/buildhtml.sh
+++ b/docs/generator/buildhtml.sh
@@ -52,10 +52,10 @@ for f in "${EXCLUDE_LIST[@]}"; do
 	rm "${SRC_DIR}/$f"
 done
 
-echo "Fetching localization project"
-LOC_DIR=${GENERATOR_DIR}/localization
-rm -rf ${LOC_DIR}
-git clone https://github.com/netdata/localization.git ${LOC_DIR}
+# echo "Fetching localization project"
+# LOC_DIR=${GENERATOR_DIR}/localization
+# rm -rf ${LOC_DIR}
+# git clone https://github.com/netdata/localization.git ${LOC_DIR}
 
 echo "Preparing directories"
 MKDOCS_CONFIG_FILE="${GENERATOR_DIR}/mkdocs.yml"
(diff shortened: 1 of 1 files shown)


prologic:
> The second step in #8990, to be merged after #8991. This PR removes the logic to run/build the old docs site using mkdocs and simplifies the buildhtml.sh script so that it essentially only sets up a clean directory tree and runs the link-checking script against it.

I'm not really quite sure I understand this. a) The name is now just wrong `build.html.sh` just doesn't do what the name on the tin suggests :) and b) Didn't you just add a link checker workflow in the other PR? Why do we need this at all now? Just ask'n :)

cakrit:
> I'm not convinced we need this :) We can just run the `checklinks.sh` directly and we have a new workflow anyway.

That will require changes to the  `checklinks.sh`  script. It can be done, what Joel did is much easier. I'd merge it and improve it later or remove the test completely, if we find a better way forward.

joelhans:
@prologic You're right that if we merge #9003, then we don't need this PR at all. The easiest way to move ahead would be to merge #9003, followed by #8783.

This PR is an imperfect shortcut. #9003 and #8783 is the more "correct" way to go.

prologic:
> @prologic You're right that if we merge #9003, then we don't need this PR at all. The easiest way to move ahead would be to merge #9003, followed by #8783.
> 
> This PR is an imperfect shortcut. #9003 and #8783 is the more "correct" way to go.

Then let's do that! Close this PR and I will ensure we get the others approved/reviewed by me and merged!

prologic:
I'm slowly getting to them as they appear in my queues! (_sorry for the delay, swamped with large volume of notifs, oncall and hard focus time on eBPF_)

joelhans:
Closing, as we'll move ahead with #8783 instead when the SRE team is ready.

PR #9005, opened 2020-05-12
Fixes enable/start of netdata service in debian package

### LINKED ISSUE #8878: Netdata service is stopped after an upgrade with apt
##### Bug report summary
After doing an `apt upgrade netdata` the netdata service is no longer running.
We want to use unattended-upgrades to automatically update netdata on all of our servers, but at each update all of our netdata instances are stopped.

##### OS / Environment

Debian 10/Ubuntu 18.04

##### Netdata version
v1.21.1
 

##### Component Name
N/A

##### Steps To Reproduce

1. Install netdata using packagecloud repository https://packagecloud.io/netdata/netdata/install
2. Start netdata
2. Upgrade netdata using apt
3. Netdata is stopped using SIGTERM (just before the upgrade AFAIK)

##### Expected behavior
Netdata should be started after an upgrade (at least if the service was running before attempting the upgrade).
I don't have much experience with debian packaging so I don't really know what is the best option, but I saw that for instance nginx/apache are automatically started after an install/upgrade, perhaps a solution like that would be sufficient ?

If it's not suitable maybe netdata could restart itself (if the service is started) after the completion of the upgrade ?

MrFreezeex:
Signed-off-by: Arthur Outhenin-Chalandre <arthur@cri.epita.fr>

Fixes #8878 

##### Summary
Fixes start/enable of netdata because of missing `#DEBHELPER#` in debian scripts.
I also removed the templating for postinst, AFAIK It is unused as the templated script and the final script are identical.

##### Component Name
packaging

### DIFF
--- a/contrib/debian/netdata.postinst
+++ b/contrib/debian/netdata.postinst
@@ -67,4 +67,6 @@ case "$1" in
     ;;
 esac
 
+#DEBHELPER#
+
 exit 0
--- a/contrib/debian/netdata.postrm
+++ b/contrib/debian/netdata.postrm
@@ -48,4 +48,6 @@ case "$1" in
 
 esac
 
+#DEBHELPER#
+
 exit 0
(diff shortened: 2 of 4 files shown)


MrFreezeex:
I rebased this PR after the merge of #8996. It will also enable netdata instances installed through apt to auto-upgrade their instance.

prologic:
@Ferroin Please review.

Issue #8461, opened 2020-03-23
Overall cleanup of the exporting engine


vlvkobal:
About 30 minor things to implement or to think about. Draft list:

__To implement__
- [x] TLS support in the simple connector worker. OpenTSDB. ( PR #9068 )
- [x] Send anonymous stats for backends and exporting connectors ( PR #9125 )
- [x] Add an alarm to inform users that backends subsystem is deprecated ( PR #9075 )
- [x] Check return values for uv_mutex_init() and uv_cond_init() ( PR #9034 )
- [x] Check if splitting a buffer doesn’t cut '\n' from the output of the Kinesis connector. ( PR #9122 )
- [x] OpenTSDB - revise formatting/sending of HTTP headers ( Go to [description](https://github.com/netdata/netdata/issues/8461#issuecomment-629578886) )
- [x] Move nc-backend.sh (a simple backend database proxy)  ( PR #9030 )
- [x] Add error message: "stream name is a mandatory Kinesis parameter but it is not configured" ( PR #9048 )
- [x] Align update intervals for every instance based on database update interval, print a warning message ( PR #9131 )
- [x] Define alarms using templates ( PR #9075 )

__To create issues__
- [x] HTTP headers in the OpenTSDB connector
- [x] Test read_exporting_config, exporting_select_type
- [x] Check indexes for two instances in unit tests
- [x] Implement http endpoint in destination URL.
- [x] Delete backends after a transition period. host tags are used in rrd_init(). We should delete them after removing backends. (#7017)

__Done__
- [x] Rename fix_... in graphite connector. sanitize?
- [x] Deduplicate fix_graphite_label_key and fix_graphite_label_value
- [x] Write tests for label formatting (escape_json, fix_graphite_label_key, fix_graphite_label_value)
- [x] Implement buffer on failures for simple connection worker
- [x] Unit test for the instance initializer. Can I add JSON there?
- [x] Unit tests for instance initializer should be expanded after the connector structure was removed. I can also add NULLs. An example - test_aws_kinesis_init.

__Won't be implemented__
- [x] ~~Copy the old configuration to a new file~~
- [x] ~~Kinesis configuration can be read in `aws_kinesis.c` then specific_config structure could reside in `aws_kinesis.h`~~
- [x] ~~Default timeout, check the range of connection timeout for simple connector worker~~
- [x] ~~Initialize empty values for configuration options? The same for Prometheus remote write.~~
[...]

PR #9030, opened 2020-05-13
Move nc backend

thiagoftsm:
##### Summary
This PR moves a file and update documentation as specified at #8461.
##### Component Name
Exporting
##### Test Plan
There is not a proper test here, because we are only moving a file.  It is only necessary to verify if nothing is missed.

##### Additional Information
This PR does not brings everything listed on #8461 and it is not a blocker for any other action there.


backends/README.md
```diff
@@ -144,7 +144,7 @@ from your Netdata):
    When multiple servers are defined, Netdata will try the next one when the first one fails. This allows you to
    load-balance different servers: give your backend servers in different order on each Netdata.
 
-   Netdata also ships `nc-backend.sh`, a script that can be used as a fallback backend to save the
+   Netdata also ships `../exporting/nc-backend.sh`, a script that can be used as a fallback backend to save the
```
vlvkobal, line 147: I think we can't remove it from backends right now. Let's left backends part untouched. We will remove the hole subsystem after the transition period has ended.

thiagoftsm: Ok I copied the file back and I am keeping the duplication for while.

PR #9044, opened 2020-05-14
Regenerate topic base on connect

### LINKED ISSUE #9032: Node unreachable from Cloud after reclaiming
##### OS / Environment

OS:
root@media ~ # cat /etc/os-release
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"

Netdata Version:
root@media ~ # netdata -v
netdata v1.22.1-4-nightly

Installation Method:
bash <(curl -Ss https://my-netdata.io/kickstart.sh) --non-interactive >> $log 2>&1

Error log:
[error.log](https://github.com/netdata/netdata/files/4625006/error.log)
File attached.

##### Steps to reproduce

1. Remove from war room
2. Unclaim -id=$(uuidgen) method (see https://github.com/netdata/netdata/issues/8998 and https://github.com/netdata/netdata/issues/9002)
3. Node now unreachable

amoss:
##### Summary
Closes #9032 

If the agent is re-claimed at run-time then the new claim_id is not applied to the base-topic for the ACLK. This causes the broker to kick the agent for writing to the wrong topic.

##### Component Name
ACLK

##### Test Plan

* Start a claimed agent
* Remove the agent from its room
* Re-claim the agent (without stopping it), using -id=$(uuidgen)
* Re-connect the agent

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -24,6 +24,7 @@ static char *global_base_topic = NULL;
 static int aclk_connecting = 0;
 int aclk_connected = 0;             // Exposed in the web-api
 int aclk_force_reconnect = 0;       // Indication from lower layers
+int aclk_kill_link = 0;             // Tell the agent to tear down the link
 usec_t aclk_session_us = 0;         // Used by the mqtt layer
 time_t aclk_session_sec = 0;        // Used by the mqtt layer
 
@@ -990,6 +991,39 @@ void *aclk_query_main_thread(void *ptr)
     return NULL;
 }
 
+static void aclk_graceful_disconnect()
+{
+    size_t write_q, write_q_bytes, read_q;
+    time_t event_loop_timeout;
+
+    // Send a graceful disconnect message
+    BUFFER *b = buffer_create(512);
+    aclk_create_header(b, "disconnect", NULL, 0, 0);
+    buffer_strcat(b, ",\n\t\"payload\": \"graceful\"}\n");
+    aclk_send_message(ACLK_METADATA_TOPIC, (char*)buffer_tostring(b), NULL);
+    buffer_free(b);
+
+    event_loop_timeout = now_realtime_sec() + 5;
+    write_q = 1;
+    while (write_q && event_loop_timeout > now_realtime_sec()) {
+        _link_event_loop();
+        lws_wss_check_queues(&write_q, &write_q_bytes, &read_q);
+    }
+
+    aclk_shutting_down = 1;
+    _link_shutdown();
+    aclk_lws_wss_mqtt_layer_disconect_notif();
+
+    write_q = 1;
+    event_loop_timeout = now_realtime_sec() + 5;
+    while (write_q && event_loop_timeout > now_realtime_sec()) {
+        _link_event_loop();
+        lws_wss_check_queues(&write_q, &write_q_bytes, &read_q);
+    }
+    aclk_shutting_down = 0;
+}
+
+
 // Thread cleanup
 static void aclk_main_cleanup(void *ptr)
 {
@@ -999,35 +1033,9 @@ static void aclk_main_cleanup(void *ptr)
     info("cleaning up...");
 
     if (is_agent_claimed() && aclk_connected) {
-        size_t write_q, write_q_bytes, read_q;
-        time_t event_loop_timeout;
-
         // Wakeup thread to cleanup
         QUERY_THREAD_WAKEUP;
-        // Send a graceful disconnect message
-        BUFFER *b = buffer_create(512);
-        aclk_create_header(b, "disconnect", NULL, 0, 0);
-        buffer_strcat(b, ",\n\t\"payload\": \"graceful\"}\n");
-        aclk_send_message(ACLK_METADATA_TOPIC, (char*)buffer_tostring(b), NULL);
-        buffer_free(b);
-
-        event_loop_timeout = now_realtime_sec() + 5;
-        write_q = 1;
-        while (write_q && event_loop_timeout > now_realtime_sec()) {
-            _link_event_loop();
-            lws_wss_check_queues(&write_q, &write_q_bytes, &read_q);
-        }
-
-        aclk_shutting_down = 1;
-        _link_shutdown();
-        aclk_lws_wss_mqtt_layer_disconect_notif();
-
-        write_q = 1;
-        event_loop_timeout = now_realtime_sec() + 5;
-        while (write_q && event_loop_timeout > now_realtime_sec()) {
-            _link_event_loop();
-            lws_wss_check_queues(&write_q, &write_q_bytes, &read_q);
-        }
+        aclk_graceful_disconnect();
     }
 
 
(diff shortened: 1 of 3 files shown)


underhood:
Hmm testing it got me into infinite reconnection loop of some kind :/

amoss:
It looks like the problem is caused by changing the claiming data while the link is operational. We should probably tear-down the connection if the claiming state is reloaded.

underhood:
Just for the record *(as i have memory of gold fish)*, I did skip the `Remove the agent from its room` step in test plan when this happened.

claim/claim.c
```diff
@@ -109,12 +109,26 @@ void claim_agent(char *claiming_arguments)
...
+ * This only happens when the user has explicitly requested it:
+ *   - via the cli tool by reloading the claiming state
+ *   - after spawning the claim because of a command-line argument
+ * If this happens with the ACLK active under an old claim then we MUST KILL THE LINK
+ */
 void load_claiming_state(void)
 {
     if (claimed_id != NULL) {
         freez(claimed_id);
         claimed_id = NULL;
```
stelfrag, line 125: Very very small window that might cause a crash due to `char *agent_id = is_agent_claimed();` in `aclk_get_challenge` (L1242) if called in between the freez and the set to null.

amoss: This is true and I've been wondering about it recently. The risk is minor but it is possible that accessing the cli-tool could trigger it.

aclk/agent_cloud_link.c
```diff
@@ -1319,6 +1333,8 @@ static void aclk_try_to_connect(char *hostname, char *port, int port_num)
     if (unlikely(rc)) {
         error("Failed to initialize the agent cloud link library");
     }
+    create_publish_base_topic();
```
hmoragrega, line 1336: I think this still has the issue that the agent will fail the first attempt to reconnect after a claiming id change,
the base topic (re) creation has to happen before the connection attempt at `mqtt_attempt_connection`, otherwise it will attempt to connect to the old outbound topic and verne will deny it, the second try will succeed

stelfrag: I don't _see_ an issue with the base topic being created at that point (it can be moved a bit sure). 
The connection is down and it is trying to reconnect. The challenge / response always gets the fresh claim id (as long as the reload claiming has been done)

The base topic can be created after the connection attempt  (mqtt and lws upgrade) and only when the connection is marked UP it will attempt to subscribe to the `cmd` topic (with the newly created base topic)

hmoragrega: the error happens because on connect we try to set the lwt in `_link_set_lwt` to the `outbound/meta` topic, and the first time we're still using the old `global_base_topic` with the previous id

amoss: That's interesting. It worked well when I tested it because we don't use the topic for messages until later as @stelfrag mentioned. But I did not test the LWT and it will probably work the way that you have said. I will take another look.

amoss:
Rebased - new fix is a mutex around the `claim_id` to address @stelfrag 's concern, and I've moved the topic creation earlier to fix the problem with the LWT that @hmoragrega diagnosed.

amoss:
Fails in CI/CD - I need to check what is different to the build on my local system...

amoss:
CI/CD picked up a real problem - we had some existing code that would have failed the build when the cloud deps were not available. This has been in master for a while but not triggered the failure before. Fixed now.

web/api/web_api_v1.c
```diff
@@ -874,10 +874,13 @@ inline int web_client_api_request_v1_info_fill_buffer(RRDHOST *host, BUFFER *wb)
 #else
     buffer_strcat(wb, "\t\"cloud-available\": false,\n");
 #endif
-    if (is_agent_claimed() == NULL)
+    char *agent_id = is_agent_claimed();
+    if (agent_id == NULL)
         buffer_strcat(wb, "\t\"agent-claimed\": false,\n");
```
underhood, line 879: this is not the fault of this PR but I think we forgot to update the swagger docs with `agent-*` vars

aclk/agent_cloud_link.c
```diff
@@ -429,29 +433,29 @@ struct aclk_query *aclk_queue_pop()
...
-// it stored internally.
-// Need to check if additional logic should be added to make sure that there
-// is enough information to determine the base topic at init time
+// This is called during the connection, we delete any previous topic
+// in-case the user has changed the agent id and reclaimed.
 
 char *create_publish_base_topic()
 {
-    if (unlikely(!is_agent_claimed()))
+char *agent_id = is_agent_claimed();
```
underhood, line 441: missing tab?

aclk/agent_cloud_link.c
```diff
@@ -1577,13 +1607,13 @@ void aclk_disconnect()
     aclk_force_reconnect = 1;
 }
 
-void aclk_shutdown()
+/*void aclk_shutdown()
```
underhood, line 1610: remove dead code?

claim/claim.c
```diff
@@ -26,12 +26,19 @@ static char *claiming_errors[] = {
...
 static char *claimed_id = NULL;
 
-char *is_agent_claimed(void)
+/* Retrieve the claim id for the agent.
+ * Caller owns the string.
+*/
+char *is_agent_claimed()
 {
-    return claimed_id;
+char *result;
```
underhood, line 37: missing tab?

aclk/agent_cloud_link.c
```diff
@@ -992,6 +996,39 @@ void *aclk_query_main_thread(void *ptr)
...
+    event_loop_timeout = now_realtime_sec() + 5;
+    write_q = 1;
+    while (write_q && event_loop_timeout > now_realtime_sec()) {
+        _link_event_loop();
+        lws_wss_check_queues(&write_q, &write_q_bytes, &read_q);
+    }
+
+    aclk_shutting_down = 1;
+    _link_shutdown();
+    aclk_lws_wss_mqtt_layer_disconect_notif();
```
underhood, line 1020: shouldn't mosq_disconnect call this as well from `_link_shutdown`?

aclk/agent_cloud_link.c
```diff
@@ -992,6 +996,39 @@ void *aclk_query_main_thread(void *ptr)
...
+    // Send a graceful disconnect message
+    BUFFER *b = buffer_create(512);
+    aclk_create_header(b, "disconnect", NULL, 0, 0);
+    buffer_strcat(b, ",\n\t\"payload\": \"graceful\"}\n");
+    aclk_send_message(ACLK_METADATA_TOPIC, (char*)buffer_tostring(b), NULL);
+    buffer_free(b);
+
+    event_loop_timeout = now_realtime_sec() + 5;
+    write_q = 1;
+    while (write_q && event_loop_timeout > now_realtime_sec()) {
```
underhood, line 1013: this while is repeated few lines below - should we put it into func?

claim/claim.c
```diff
@@ -109,12 +116,35 @@ void claim_agent(char *claiming_arguments)
 #endif
 }
 
+#ifdef ENABLE_ACLK
+extern int aclk_connected, aclk_kill_link;
+extern void aclk_graceful_disconnect();
```
underhood, line 121: Is `aclk_graceful_disconnect` actually used here?

aclk/agent_cloud_link.c
amoss: No it is fine like this

aclk/agent_cloud_link.c
amoss: Possibly, but this is a forced shutdown and I think the motivation was to make sure it had occurred.

claim/claim.c
amoss: No, that was from the previous version

web/api/web_api_v1.c
amoss: In theory...

Issue #8679, opened 2020-04-11
netdata update script deleted the netdata.conf file


yasharne:
##### Question summary
I planned to update our master netdata servers to the latest stable version (1.21.0) using the script located in `/opt/netdata/usr/libexec/netdata/netdata-updater.sh`, after the upgrading process, I noticed all files inside `/opt/netdata/etc/netdata` directory including plugins and `stream.conf` are unchanged except `netdata.conf` file which says I can download the template using 127.0.0.1:19999/netdata.conf
Is it something expected in the process of updating?
##### OS / Environment
netdata v1.20.0
##### Expected results
keep `netdata.conf` like every other config file

prologic:
> Is it something expected in the process of updating?

I'm afarid this is a bit of a know problem at the moment (See: https://github.com/netdata/netdata/issues/8531) I don't yet have a fix for this; but my plan was to make it possible for a user (_you_) to be prompted of any changes and give you an opportunity to do things like:

- take upstream
- leave mine alone
- merge differences

A bit like what the `git` and `dpkg` do.

Issue #8531, opened 2020-03-29
Static distribution overwrites edited config files, with netdata v1.20.0


dmfreemon:
[I could not figure out how to re-open a issue/bug report, so...]

I'm seeing the same issue as:

https://github.com/netdata/netdata/issues/6692

except with netdata version 1.20.0.

Bug report summary (which is the same as #6692):

When running the static version installer (bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh)), which is the recommended way to update, the installer overwrites my edited netdata.conf file with its empty one.

My repo steps are exactly the same as #6692.

My version string is:
$ /opt/netdata/usr/bin/netdata -V
netdata v1.20.0-258-g182ea93b

I have confirmed this happens for (at least) these distros:

* CentOS Linux release 7.7.1908 (Core)
* Fedora 31

dmfreemon:
The automatic update script that runs via cron.daily also overwrites the netdata.conf file in the same way.

prologic:
I will get to this in the next day or two. Thanks for reporting its in my immediate backlog.

prologic:
Sorry for the delay (_other things got in the way_). I can confirm this is a bug and I can reproduce it. Here's what I know so far:

- The `kickstart_static64.sh` -- Specifically the `install_or_update.sh` script that is embedded inside the [Makeself](https://makeself.io/) archive _tries_ to copy old configs and creates a symlink called `orig`.
  - The problem with this is a) users don't seem to know about it b) it doesn't preserve the existing `netdata.conf` at all that I can find and c) its surprising behavior

I'm working on.a way to interactively preserve configuration changes on updates where you get the opportunity to do things like:

- take upstream version
- take your version
- merge and edit the result

I'm looking at what options exists that I can reuse (_depending on license_) and I know of:

- CRUX's `rejmerge` -- GPLv2
- Gentoo's `dispatch-conf` -- GPLv2
- _there are heaps of others embedded inside Linux distributions..._

Many of these tools however use the familiar `diff` and `patch` tools usually available on most systems including busybox-type systems like [uLinux](https://github.com/prologic/ulinux) where I'm testing this on...

I plan to have something testable today...

Ferroin:
Sorry for the (further) delay on this one. I should have a quick fix to avoid nuking the `netdata.conf` file some time later today, and if all goes well it should land in the nightlies within the next few days.

Ferroin:
[...]

PR #9046, opened 2020-05-14
Don't overwrite netdata.conf on update on static installs.

Ferroin:
##### Summary

The current code overwrites netdata.conf when updating a static install because we include a copy of the default empty netdata.conf in the package itself.

This removes that file from the generated package, but does not modify the install-time logic which attempts to create the file if it does not already exist (which appears to be working correctly.

##### Component Name

area/packaging

##### Test Plan

Locally built and verified that the static installer does not include `netdata.conf` in the archive anymore.

##### Additional Information

Fixes: #8531 
Fixes: #8679 

To all our users who were affected by this, sorry it took so long to fix it.

### DIFF
--- a/packaging/makeself/jobs/70-netdata-git.install.sh
+++ b/packaging/makeself/jobs/70-netdata-git.install.sh
@@ -22,6 +22,10 @@ run ./netdata-installer.sh --install "${NETDATA_INSTALL_PARENT}" \
     --dont-start-it \
     ${NULL}
 
+# Remove the netdata.conf file from the tree. It gets created at install
+# time instead.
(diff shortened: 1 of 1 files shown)


packaging/makeself/jobs/70-netdata-git.install.sh
```diff
@@ -22,6 +22,10 @@ run ./netdata-installer.sh --install "${NETDATA_INSTALL_PARENT}" \
     --dont-start-it \
     ${NULL}
 
+# Remove the netdata.conf file from the tree. It gets created at install
+# time instead.
```
prologic, line 26:
```suggestion
# Remove the netdata.conf file from the tree. It has hard-coded sensible defaults builtin.
```

PR #9048, opened 2020-05-14
Missing error aws

thiagoftsm:
##### Summary
This PR brings the missing error report  as specified at #8461.

##### Component Name
Exporting
##### Test Plan

1 - Compile the branch
2 - Configure `exporting.conf` without the necessary parameter:

```
[kinesis:my_instance2]
    enabled = yes
    destination = us-east-1
    #stream name = netdata
    # aws_access_key_id = 
    #aws_secret_access_key = 
    data source = as collected
    #data source = sum
    update every = 2
    send charts matching = system.processes
```

3 - Run Netdata
4 - Verify if the message was written inside error.log

##### Additional Information
This PR does not brings everything listed on #8461 and it is not a blocker for any other action there.

### DIFF
--- a/exporting/read_config.c
+++ b/exporting/read_config.c
@@ -390,7 +390,7 @@ struct engine *read_exporting_config()
             tmp_instance->config.connector_specific_config = connector_specific_config;
 
             connector_specific_config->stream_name = strdupz(exporter_get(instance_name, "stream name", ""));
-            if (!connector_specific_config->stream_name)
+            if (!strcmp(connector_specific_config->stream_name, ""))
                 error("stream name is a mandatory Kinesis parameter but it is not configured");
 
             connector_specific_config->auth_key_id = strdupz(exporter_get(instance_name, "aws_access_key_id", ""));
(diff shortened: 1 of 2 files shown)


exporting/read_config.c
```diff
@@ -389,7 +389,10 @@ struct engine *read_exporting_config()
 
             tmp_instance->config.connector_specific_config = connector_specific_config;
 
-            connector_specific_config->stream_name = strdupz(exporter_get(instance_name, "stream name", "netdata"));
+            connector_specific_config->stream_name = strdupz(exporter_get(instance_name, "stream name", ""));
+            if (!strcmp(connector_specific_config->stream_name, ""))
```
vlvkobal, line 393: I would check it in `init_aws_kinesis_instance()` and return 1 on error. Then the connector instance won't even start.

thiagoftsm: All right, I changed the PR, but we will need to take care of the buffer in another PR as we talked before.

PR #9058, opened 2020-05-15
claim: fix `New/ephemeral Agent containers` instruction

ilyam8:
##### Summary

ssia

we need to pass only args

https://github.com/netdata/netdata/blob/22b8f9498ad240055f7b63eb7d409dba60e4a5f4/packaging/docker/run.sh#L27

##### Component Name

docs

##### Test Plan

- [X] claim nd in docker container with the command, ensure it appears on the cloud


ilyam8:
during tests i realized that it is better to have token/rooms on a separate lines  (token/rooms ids are very long)

Issue #9051, opened 2020-05-14
Bundle missing in ClearLinux* OS: ERROR when running one line installation script


thoggs:
As I title, I noticed that, when trying to run the installation script using the --stable-channel flag, on Clear Linux * OS (Server), the script returns an error, claiming that lib libuuid-devel (package uuid is speaking) -dev).

**I solved the problem by installing:**
=> devpkg-util-linux bundle 

**ERROR:** netdata failed to build...
**OS:** Clear Linux* OS - Server
**Netdata v1.22.1**

Ferroin:
@thoggs Thank you for reporting this! I'll have a PR up with a fix shortly, and it should hopefully be in the nightly builds within the next few days.

thoggs:
> @thoggs Thank you for reporting this! I'll have a PR up with a fix shortly, and it should hopefully be in the nightly builds within the next few days.

Thank you very much for your attention.
I am also a developer, and I understand about Shell Script, maybe I can help the project to progress even more...

thoggs:
Hello,
Tested, problem solved...

![Captura de Tela 2020-05-26 às 01 15 35](https://user-images.githubusercontent.com/43221193/82861328-ac681d00-9ef2-11ea-91a3-52cb147c70c6.png)

PR #9060, opened 2020-05-15
Added required bundle for libuuid on ClearLinux.

Ferroin:
##### Summary

We somehow forgot to include the bundle for libuuid in our dependencies on ClearLinux, this fixes that.

##### Component Name

area/packaging.

##### Test Plan

Checked locally that the correct package is installed and verified the build of Netdata.

##### Additional Information

Fixes: #9051

### DIFF
--- a/.github/workflows/build-and-install.yml
+++ b/.github/workflows/build-and-install.yml
@@ -19,6 +19,7 @@ jobs:
           - 'archlinux:latest'
           - 'centos:8'
           - 'centos:7'
+          - 'clearlinux:latest'
           - 'debian:bullseye'
           - 'debian:buster'
           - 'debian:stretch'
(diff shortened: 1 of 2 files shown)


prologic:
> We somehow forgot to include the bundle for libuuid in our dependencies on ClearLinux, this fixes that.

I can only assume we have no CI check for Build / INstall on ClearLinux? Can we please add it?

Ferroin:
> > We somehow forgot to include the bundle for libuuid in our dependencies on ClearLinux, this fixes that.
> 
> I can only assume we have no CI check for Build / INstall on ClearLinux? Can we please add it?

Done.

It wasn't added originally because it was expected to be a very low user base compared to the other distros we check while theoretically taking a lot longer than the others to actually test (I should have just timed it back then instead of assuming that the almost 200MB of data that needs to be downloaded by the Docker image would be an issue).

PR #9074, opened 2020-05-17
Implement new incremental parser

### LINKED ISSUE #8180: Transform plugins.D to an incremental parser
##### Feature idea summary

In order for the Plugins.D parser to be usable by the [metadata log](https://github.com/netdata/docs/blob/master/architecture/agent/db/database-log-format-pros-cons.md), we need to trasform it; from a self-contained uninterruptible parser that consumes a whole serial stream and acts on it, to an incremental streaming parser that shares its own iterative state and allows for other contexts such as the metadata log code to process the results and initiate I/O to fill buffers for the parser to consume.

* `pluginsd_process()` is called from two places.
* The file pointer needs to be replaced with a buffer structure.
* Some kind of wrapper / shim needs to load the buffer in the different cases (pipes / streaming).
* Cleanly separate out the https cases.
* Allow the processing function to be called iteratively to parse the stream.

Examples:
```
    struct {
        .set = &act_on_set;
        .chart = &act_on_chart;
        .label = &act_on_label;
    } actions;
    buffer = pluginsd_stream_init(state, fp);
    for (i ; ; ++i) {
        pluginsd_parse_line(state, buffer, &actions);
        pluginsd_do_action(state, &actions);
    }
```
or:
```
    struct {
        .set = &act_on_set;
        .chart = &act_on_chart;
        .label = &act_on_label;
    } actions;
    pluginsd_stream_init(state, &actions);
    for (i ; ; ++i) {
        buffer = streaming_https(state, fp);
        count += pluginsd_parse_do_action(state, buffer);
        
        // pause parsing, do stuff, do I/O
    }
```

##### Test scenarios
* Local stream from a collector
* Remote stream over http from a slave
* Remote stream over https from a slave

stelfrag:
Fixes #8180
##### Summary
Implements new incremental parser to be used by internal plugins and slaves. 

Usage:
- A new parser is allocated with an appropriate `input` stream
  - An optional `read` and `eof` function can be supplied (defaults to the current implementation
    which expects a `FILE *`
- The pluginss.d language keywords are registered with callback functions
- Enter a loop to
  - Fetch next line from the input stream
  - Call the action function
     - Exit on error
  
##### Component Name
Streaming, Internal collectors

##### Test Plan
- Compile a master agent with the changes
  - Setup a slave that connects to the master without SSL
  - All internal plugins and slave connections should work as before
    - Notice in the logfile message like the following
```
netdata INFO  : PLUGINSD[go.d] : Registered 10 keywords for the parser
netdata INFO  : PLUGINSD[nfacct] : Registered 10 keywords for the parser
netdata INFO  : PLUGINSD[charts.d] : Registered 10 keywords for the parser
netdata INFO  : STREAM_RECEIVER[ubuntu1804,[192.168.122.103]:57304] : Registered 10 keywords for the pars
```
as plugins and incoming slaves initialize a parser and register keywords for processing

Forcing a plugin or a slave to stop will log a `PARSER ended` message as follows:
```
netdata INFO  : PLUGINSD[apps] : PARSER ended
netdata ERROR : PLUGINSD[apps] : '/home/stelios/netdata-inst//netdata/usr/libexec/netdata/plugins.d/apps.plugin' (pid 2982315) disconnected after 9380 successful data collections (ENDs).
netdata ERROR : PLUGINSD[apps] : child pid 2982315 killed by signal 15.
netdata INFO  : PLUGINSD[apps] : '/home/stelios/netdata-inst//netdata/usr/libexec/netdata/plugins.d/apps.plugin' (pid 2982315) was killed with SIGTERM. Disabling it.
netdata INFO  : PLUGINSD[apps] : thread with task id 2982299 finished
netdata INFO  : STREAM_RECEIVER[ubuntu1804,[192.168.122.103]:59308] : PARSER ended
netdata ERROR : STREAM_RECEIVER[ubuntu1804,[192.168.122.103]:59308] : STREAM ubuntu1804 [receive from [192.168.122.103]:59308]: disconnected (completed 4115 updates).
netdata INFO  : STREAM_RECEIVER[ubuntu1804,[192.168.122.103]:59308] : STREAM ubuntu1804 [receive from [192.168.122.103]:59308]: receive thread ended (task id 2982546)
```
- Stop the master and slave
  - Configure slave to connect to the master using SSL
    - Repeat the test as above


incremental_parser/incremental_parser.c
```diff
@@ -0,0 +1,338 @@
...
+        working_parser->data = tmp_parser_data->next;
+        freez(tmp_parser_data->line);
+        freez(tmp_parser_data);
+        return 0;
+    }
+
+    if (unlikely(fileno((FILE *)working_parser->input) == -1)) {
+        error("Stream is no longer valid");
+        return 1;
+    }
```
mfundul, line 199: This doesn't seem to allow working with `char *` lines.

stelfrag: Removed

mfundul:
Just noting we should be reusing the plugins.D language parsing code and parameterising the actions on language syntax.

amoss:
Yes, I will review it again when the structure is changed. While the work is ongoing - can we change the name?

"Incremental parser" makes sense now as it distinguishes it from the old parser, but the name will not make sense after the old parser is gone. Perhaps "plugins parser", "internal parser", "netdata parser" or "collectors parser" would make more sense in the long term?

amoss:
In addition to this (assuming that you fix the bad rebase) there is still a basic structural problem with the code. The parsing steps are still in the action callback, this means that they need to be duplicated for each version of the parser (streaming, log replay, collectors etc). 

As a concrete example:
```
PARSER_RC pluginsd_begin_action(void *user, char *id, usec_t microseconds)
{
    RRDSET *st = ((PARSER_USER_OBJECT *) user)->st;
    RRDHOST *host = ((PARSER_USER_OBJECT *) user)->host;

    if (unlikely(!id)) {
        error("requested a BEGIN without a chart id for host '%s'. Disabling it.", host->hostname);
        goto disable;
    }

    st = rrdset_find(host, id);
    if (unlikely(!st)) {
        error("requested a BEGIN on chart '%s', which does not exist on host '%s'. Disabling it.", id, host->hostname);
        goto disable;
    }

    if (likely(st->counter_done)) {
        if (likely(microseconds)) {
            if (((PARSER_USER_OBJECT *) user)->trust_durations)
                rrdset_next_usec_unfiltered(st, microseconds);
            else
                rrdset_next_usec(st, microseconds);
        } else
            rrdset_next(st);
    }
    ((PARSER_USER_OBJECT *) user)->st = st;
    return PARSER_RC_OK;

disable:
    ((PARSER_USER_OBJECT *)user)->enabled = 0;
    return PARSER_RC_ERROR;
}
```

Each of those validation checks is part of the parser, not part of the action being performed, so this part of the code needs to move into a common parser:
```

    if (unlikely(!id)) {
        error("requested a BEGIN without a chart id for host '%s'. Disabling it.", host->hostname);
        goto disable;
    }

    st = rrdset_find(host, id);
    if (unlikely(!st)) {
        error("requested a BEGIN on chart '%s', which does not exist on host '%s'. Disabling it.", id, host->hostname);
        goto disable;
    }
```

Leaving the actual action taken by this version of the parser as the callback:
```
    if (likely(st->counter_done)) {
        if (likely(microseconds)) {
            if (((PARSER_USER_OBJECT *) user)->trust_durations)
                rrdset_next_usec_unfiltered(st, microseconds);
            else
                rrdset_next_usec(st, microseconds);
        } else
            rrdset_next(st);
    }
```

Each of the actions has this problem and still need to be properly separate. We are not going to maintain three separate copies of the parsing code, we only want to change what the different call-sites do with the parsed command.

stelfrag:
> Did something go wrong with a rebase? This seems to merge the code for the random uuids and the metadata log into master - is this ready to go into production?

fixed that

stelfrag:
> Each of the actions has this problem and still need to be properly separate. We are not going to maintain three separate copies of the parsing code, we only want to change what the different call-sites do with the parsed command.

Ok that needs to change the calling params in actions  as needed

amoss:
New structure looks correct. Running some tests.

mfundul:
```
[ 97%] Linking CXX executable netdata
CMakeFiles/netdata.dir/collectors/plugins.d/plugins_d.c.o: In function `pluginsd_worker_thread':
/home/runner/work/netdata/netdata/collectors/plugins.d/plugins_d.c:346: undefined reference to `pluginsd_process'
CMakeFiles/netdata.dir/build.make:2171: recipe for target 'netdata' failed
CMakeFiles/netdata.dir/streaming/rrdpush.c.o: In function `rrdpush_receive':
/home/runner/work/netdata/netdata/streaming/rrdpush.c:1295: undefined reference to `pluginsd_process'
```
This is probably a new Cmocka problem.

mfundul:
50 slaves seem to be working ok.

Issue #8181, opened 2020-02-24
Change LABELS command in plugins.D language


mfundul:
##### Feature idea summary

Change the LABEL and OVERWRITE commands in plugins.D and streaming.

This will require another change in the streaming protocol version.

##### Analysis

Right now in the common language, one needs to send a list of `LABEL` commands followed by an
`OVERWRITE` to effectively set a set of host labels. For example:

```
LABEL "colour" = 1 "blue"
LABEL "colour" = 0 "red"
OVERWRITE labels
```

This implies a `HOST` and the way it is right now it can't be extended to support charts and dimensions without breaking compatibility.

Moreover, given that we define a metadata object as a set of labels, we don't need an "update" command in the language. What we should have is an object definition with some immutable labels. If someone wants to change the labels it should be a new object.

A simple way to support labels for arbitrary objects is for the `LABEL` to precede object definitions. This avoids the need for an `OVERWRITE` command. Moreover, we need to remove the "source" field of the existing `LABEL` command since by the time this command is streamed, or dumped into the log the priority resolution has already taken place and we know which key-value pair of the ones conflicting on the same key prevailed.

The new format will be:
```
LABEL key value
```

Precedes an object definition (`HOST`, `CHART` or `DIMENSION`, only hosts are supported initially). The number of labels attached to an object depend on the number of `LABEL` commands that preceded the object definition.

example:
```
LABEL "colour" "blue"
LABEL "length" "5"
HOST 00000000:0000:0000:0000:00000000001 mechani.se registry.netdata.cloud 5 Linux UTC 'tag1="value1"'
```

##### Steps

- Implement functions to parse the new syntax.
- Ensure backwards compatibility with the streaming protocol.
- Cover cases:
   1. Master has old protocol - Slave has new protocol
   2. Slave has old protocol - Master has new protocol
   3. Master and Slave have same `LABEL` protocol version

PR #9078, opened 2020-05-17
Add Metadata log files

### LINKED ISSUE #8183: Add metadata log context and log files
##### Enabler description

Add [metadata log](https://github.com/netdata/docs/blob/master/architecture/agent/db/database-log-format-pros-cons.md) threads, implement log file logic in append only manner.

No clean-up or log replay possible at this point. The needed data structures in the agent that would allow those operations don't exist yet.

Introduce the backwards compatible extensions to the `plugins.D` language, but only for flushing. No code will be added to the parsing code in `plugins_d.c`. This is the same situation as what streaming does with`rrdpush.c`. Parsing always goes to `plugins_d.c` but flushing is handled by the context of the producer.

- The metadata log files are next to the dbengine instance datafiles/journalfiles.
- The metadata log depends on the corresponding instance having been set-up beforehand.

- At all host/chart/dimension creation update points in the code write the plugins.D command lines to the metadata log files.

###### Acceptance criteria

* Check approval.
* A logfile that shows metadata operations in the new plugins.d protocol.

mfundul:
##### Summary
Fixes #8183 
##### Component Name
database
daemon
##### Test Plan

View the contents of the metadata log files as they are appended with a command such as:
```
tail -f /var/cache/netdata/dbengine/metadatalog-00001-00001.mlf | strings
```

##### Additional Information

Add metadata log threads, implement log file logic in append only manner.

No clean-up or log replay possible at this point. The needed data structures in the agent that would allow those operations don't exist yet.


database/engine/metadata_log/logfile.h
```diff
@@ -0,0 +1,77 @@
...
+extern int destroy_metadata_logfile(struct metadata_logfile *metalogfile);
+extern int create_metadata_logfile(struct metadata_logfile *metalogfile);
+extern int load_metadata_logfile(struct metalog_instance *ctx, struct metadata_logfile *logfile);
+extern void init_metadata_record_log(struct metalog_instance *ctx);
+extern int add_new_metadata_logfile(struct metalog_instance *ctx, unsigned tier, unsigned fileno);
+extern int init_metalog_files(struct metalog_instance *ctx);
+extern void finalize_metalog_files(struct metalog_instance *ctx);
+
+
+#endif /* NETDATA_LOGFILE_H */
```
stelfrag, line 77: Complains about `no newline` at the end of file :smile:

database/engine/metadata_log/metadatalogapi.h
```diff
@@ -0,0 +1,18 @@
...
+extern void metalog_commit_update_chart(RRDSET *st);
+extern void metalog_commit_delete_chart(RRDSET *st);
+extern void metalog_commit_update_dimension(RRDDIM *rd);
+extern void metalog_commit_delete_dimension(RRDDIM *rd);
+
+/* must call once before using anything */
+extern int metalog_init(struct rrdengine_instance *rrdeng_parent_ctx);
+extern int metalog_exit(struct metalog_instance *ctx);
+
+#endif /* NETDATA_METADATALOGAPI_H */
```
stelfrag, line 18: Newline here as well :exploding_head:

configure.ac
```diff
@@ -420,6 +420,9 @@ fi
 AC_MSG_RESULT([${enable_dbengine}])
 AM_CONDITIONAL([ENABLE_DBENGINE], [test "${enable_dbengine}" = "yes"])
 
+AC_DEFINE([ENABLE_METADATALOG], [1], [netdata metadata log])
```
amoss, line 423: Why do we have this? Won't it be switched on all the time...

mfundul: Yes, it's a leftover, it should go away.

mfundul: Wait, it's not there, it was the initial commit. You can look at the final version to make sure.

mfundul: Branch mess. Let me rebase first.

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -0,0 +1,290 @@
...
+    metalog_commit_record(ctx, buffer, METALOG_COMMIT_DELETION_RECORD);
+}
+
+void metalog_commit_update_host(RRDHOST *host)
+{
+    struct metalog_instance *ctx;
+    BUFFER *buffer;
+
+    /* Metadata are only available with dbengine */
+    if (!host->rrdeng_ctx)
```
amoss, line 33: The lock in L39 below guards against the read accesses to the host for the sprintf but not this or L36.

mfundul: This code has been taken from streaming. The host lock is primarily protects the list of charts. Moreover, it has been use to protect the list of labels when labels were added and in some cases to protect some writeable fields in the host structure.

However, most of those things are not 100% safe as we proved when we crashed the agent. The immutable values however are never protected in the current locking scheme. One thing that never changes is exactly what is in L36 which is `host->rrdeng_ctx->metalog_ctx`.

This is a way too big discussion to have but I would argue that protecting L36 and many more like it does not give any protection with the way things are currently.

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -0,0 +1,290 @@
...
+                   default_rrd_update_every,
+                   host->os,
+                   host->timezone,
+                   (host->tags) ? host->tags : "");
+
+    netdata_rwlock_rdlock(&host->labels_rwlock);
+    struct label *labels = host->labels;
+    while (labels) {
+        buffer_sprintf(buffer
+            , "LABEL \"%s\" = %d %s\n"
```
amoss, line 57: The `%s` should be quoted as `"%s"`.

amoss: I don't think that you need the `label_source`

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -0,0 +1,290 @@
...
+            , labels->key
+            , (int)labels->label_source
+            , labels->value);
+
+        labels = labels->next;
+    }
+    netdata_rwlock_unlock(&host->labels_rwlock);
+
+    buffer_strcat(buffer, "OVERWRITE labels\n");
```
amoss, line 66: As a note - I don't think we are required to match the streaming implementation of this.

database/engine/metadata_log/metadatalog.c
```diff
@@ -0,0 +1,342 @@
...
+    mlf_flush_records_buffer(wc);
+}
+
+#define MAX_CMD_BATCH_SIZE (256)
+
+void metalog_worker(void* arg)
+{
+    struct metalog_worker_config *wc = arg;
+    struct metalog_instance *ctx = wc->ctx;
+    uv_loop_t* loop;
```
amoss, line 229: Why the heap instead of the stack?

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -0,0 +1,290 @@
...
+ * Returns 0 on success, negative on error
+ */
+int metalog_init(struct rrdengine_instance *rrdeng_parent_ctx)
+{
+    struct metalog_instance *ctx;
+    int error;
+
+    rrdeng_parent_ctx->metalog_ctx = ctx = callocz(1, sizeof(*ctx));
+
+    ctx->max_disk_space = 32 * 1048576LLU;
```
amoss, line 240: This should be configurable

database/engine/metadata_log/metadatalogapi.c
mfundul: This is copy-paste from streaming for now. We will fix the `LABEL` command in https://github.com/netdata/netdata/issues/8181 as you're saying.

database/engine/metadata_log/metadatalogapi.c
mfundul: Yes, you are right, we will fix this in https://github.com/netdata/netdata/issues/8181

database/engine/metadata_log/metadatalogapi.c
mfundul: We will implement this during compaction.

database/engine/metadata_log/metadatalog.c
mfundul: This is just following the best practices of `test/` code in libuv repo. There is no real reason behind this.

database/engine/metadata_log/metadatalogprotocol.h
```diff
@@ -0,0 +1,50 @@
...
+#define METALOG_OTHER               (3) /* reserved */
+
+/*
+ * Metadata log record header
+ */
+struct rrdeng_metalog_record_header {
+    /* when set to STORE_PADDING jump to start of next block */
+    uint8_t type;
+
+    uint32_t reserved; /* reserved for future use */
```
amoss, line 38: This doesn't work well for either forward or backward compatibility, change it to a header_len field instead.

mfundul: I'm not keeping those yeah. We'll have to see a little bit later how I'll shape this, I'm not sure what I'm going to need yet.

database/engine/metadata_log/logfile.c
```diff
@@ -0,0 +1,482 @@
...
+    unsigned pos, size;
+    struct metadata_logfile *metalogfile;
+
+    if (unlikely(NULL == ctx->records_log.buf || 0 == ctx->records_log.buf_pos)) {
+        return;
+    }
+    /* care with outstanding records when switching metadata log files */
+    metalogfile = ctx->metadata_logfiles.last;
+
+    io_descr = mallocz(sizeof(*io_descr));
```
amoss, line 52: Does this leak? (I can't find the free)

amoss: Ah no - there it is in the callback :)

amoss:
This is inside the log:
```
HOST "7afcb008-99d6-11ea-bd72-0242ac110002" "fba9008562ca" "fba9008562ca" 1 "linux" "Etc/UTC" ""LABEL "_os_name" = 0 unknown
```

The extra quote looks wrong: `""LABEL`

mfundul:
> 
> 
> This is inside the log:
> 
> ```
> HOST "7afcb008-99d6-11ea-bd72-0242ac110002" "fba9008562ca" "fba9008562ca" 1 "linux" "Etc/UTC" ""LABEL "_os_name" = 0 unknown
> ```
> 
> The extra quote looks wrong: `""LABEL`

You're right. Just added a new line character.

PR #9079, opened 2020-05-18
Introduce a random sleep in the Netdata updater

prologic:
##### Summary

ssia

##### Component Name

- area/packaging

##### Test Plan

Tested the shell expression:

```#!sh
sleep $(((RANDOM % 10) + 1))s
```

##### Additional Information

This will hopefully aliviate the stampede like effect we're seeing in Netdata
Cloud Production where we see spikes of disconnections and reconenctions.
The timing lines up with the default Debian crontab entry for "daily" jobs
and thus adding a random sleep into the updater should help spread this out
a bit and reduce load on our system9s) in short intervals.

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -213,8 +213,9 @@ trap cleanup EXIT
 # Random sleep to aileviate stampede effect of Agents upgrading
 # and disconnecting/reconnecting at the same time (or near to).
 # But only we're not a controlling terminal (tty)
+# Randomly sleep between 1s and 60m
 if [ ! -t 1 ]; then
-  sleep $((RANDOM % (3600 - 1800 + 1) + 1800))s
+  sleep $(((RANDOM % 3600) + 1))s
 fi
 
 # Usually stored in /etc/netdata/.environment
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -212,7 +212,10 @@ trap cleanup EXIT
 
 # Random sleep to aileviate stampede effect of Agents upgrading
 # and disconnecting/reconnecting at the same time (or near to).
-sleep $(((RANDOM % 10) + 1))s
+# But only we're not a controlling terminal (tty)
+if [ ! -t 1 ]; then
+  sleep $((RANDOM % (3600 - 1800 + 1) + 1800))s
+fi
 
 # Usually stored in /etc/netdata/.environment
 : "${ENVIRONMENT_FILE:=THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT}"
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -185,8 +185,8 @@ update() {
       do_not_start="--dont-start-it"
     fi
 
-    if [ -n "${NETDATA_SELECTED_DASHBOARD}" ] ; then
-        env="NETDATA_SELECTED_DASHBOARD=${NETDATA_SELECTED_DASHBOARD}"
+    if [ -n "${NETDATA_SELECTED_DASHBOARD}" ]; then
+      env="NETDATA_SELECTED_DASHBOARD=${NETDATA_SELECTED_DASHBOARD}"
     fi
 
     info "Re-installing netdata..."
@@ -209,6 +209,10 @@ logfile=
 tmpdir=
 
 trap cleanup EXIT
+
+# Random sleep to aileviate stampede effect of Agents upgrading
+# and disconnecting/reconnecting at the same time (or near to).
+sleep $(((RANDOM % 10) + 1))s
 
 # Usually stored in /etc/netdata/.environment
 : "${ENVIRONMENT_FILE:=THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT}"
--- a/tests/updater_checks.bats
+++ b/tests/updater_checks.bats
@@ -56,6 +56,9 @@ setup() {
 	# Run the updater, with the override so that it uses the local repo we have at hand
 	# Try to run the installed, if any, otherwise just run the one from the repo
 	export NETDATA_LOCAL_TARBAL_OVERRIDE="${PWD}"
+	# Disable random sleep / splay for netdata-updater to avoid sampede effect
+	# of many agents (dis|re)connecting too quickly all at onace to Netdata Cloud
+	unset RANDOM; export RANDOM=0
 	/etc/cron.daily/netdata-updater || ./packaging/installer/netdata-updater.sh
 	! grep "new_installation" "${ENV}"
 }


cakrit:
The sleep command is too early in the script, there are a couple of possible failure conditions that would prevent installation further down. It should probably go before the actual running of the installation code, after the checksum validation. 

This is a bad hack, I don't like it at all. Have we really exhausted all other options and have to resort to this?

Ferroin:
> This is a bad hack, I don't like it at all. Have we really exhausted all other options and have to resort to this?

There is _zero_ functionality to provide this in any of the task schedulers we do or could use for our updater. The expectation is that if a job needs random jitter like this, it will handle it itself.. I'm not hugely fond of it myself (it should ideally be set up so it only runs when run non-interactively, and I'd go with a much longer range (more like 5-10 minutes, at least)), but it's better than nothing.

Aside from the stampede effect we're seeing in Netdata Cloud due to updates, it's also good practice to help avoid putting undue load on our update infrastructure.

Also, this is generally expected behavior for automated update systems. See for example FreeBSD's update tool (it refuses to run non-interactively without special handling, and the cron job mode sleeps a random number of seconds up to a full hour before actually doing it's job).

cakrit:
Ok, fine, just move it to after the sanity checks.

cakrit:
Fine, go ahead. I see we are trying to spread out the downloads themselves as well, so the sleep can't go much later than where it is...

packaging/installer/netdata-updater.sh
```diff
@@ -210,6 +210,10 @@ tmpdir=
 
 trap cleanup EXIT
 
+# Random sleep to aileviate stampede effect of Agents upgrading
+# and disconnecting/reconnecting at the same time (or near to).
+sleep $(((RANDOM % 10) + 1))s
+
```
Ferroin, line 216: On further thought, I'm very strongly of the opinion that this should:

1. Be used only when run non-interactively (no controlling terminal for the script is sufficient IMO).
2. Be _much_ longer on the maximum, probably at least 5 minutes, but more ideally closer to half an hour or an hour.

ilyam8: probably we need something like that

```
min = x
max = y
sleep = rand(max - min) + min
```

Ferroin: Provided it's a uniform distribution (which I believe is the case for the bash implementation of `$RANDOM`), we should just set the minimum at 0.

prologic: Unclear of the unformity of Bash's `$RANDOM` but the man page has this to say:

>        RANDOM Each time this parameter is referenced, a random integer between 0 and 32767 is generated.  The sequence of random numbers may  be
>              initialized  by  assigning  a  value  to  RANDOM.  If RANDOM is unset, it loses its special properties, even if it is subsequently
>              reset.

mfundul: Wait, so unsetting `$RANDOM` is a security concern? Like a shell could export `$RANDOM` to a subshell and kill the randomness?

Ferroin: @mfundul Yes. Most 'special' shell variables are only special until they're overridden by the environment. It's kind of like how some old languages let you use purely numeric identifiers.

prologic: Done

packaging/installer/netdata-updater.sh
```diff
@@ -210,6 +210,13 @@ tmpdir=
 
 trap cleanup EXIT
 
+# Random sleep to aileviate stampede effect of Agents upgrading
+# and disconnecting/reconnecting at the same time (or near to).
+# But only we're not a controlling terminal (tty)
+if [ ! -t 1 ]; then
+  sleep $((RANDOM % (3600 - 1800 + 1) + 1800))s
```
Ferroin, line 217: Unless I'm mistaken, this results in a range of 30-60 minutes for the delay. Is there some reason not to just make it 0-60 minutes instead if we're putting 60 minutes as the upper bound?

prologic: Its called "middle ground" I met you halfway. 😀

ilyam8: > Unless I'm mistaken, this results in a range of 30-60 minutes for the delay.

this is not clear from the code, i sugges to add it as a code ([min, max])

Ferroin: @prologic I think I might have been unclear then in what I was saying. I still think the lower limit should be 0 seconds, I just believe that we should spread across a much longer period of time than the 10s that your original code did.

If we really are just going to do a 30 minute range, then that should be 0-30 minutes, not 30-60.

prologic:
Sanity check before merging...

```#!sh
root@ubuntu2004:~# tail -f /var/log/syslog
May 26 00:45:01 ubuntu2004 CRON[81223]: (root) CMD (/usr/libexec/netdata/netdata-updater.sh)
root@ubuntu2004:~# ps aux | grep netdata-updater
root       81223  0.0  0.0   2608   544 ?        Ss   00:45   0:00 /bin/sh -c /usr/libexec/netdata/netdata-updater.sh
root       81224  0.0  0.3   7024  3412 ?        S    00:45   0:00 bash /usr/libexec/netdata/netdata-updater.sh
root       81227  0.0  0.0   6436   740 pts/0    S+   00:45   0:00 grep --color=auto netdata-updater
root@ubuntu2004:~# strace -p 81224
strace: Process 81224 attached
wait4(-1, ^Cstrace: Process 81224 detached
 <detached ...>
root@ubuntu2004:~# ps aux | grep sleep
root       81225  0.0  0.0   5476   592 ?        S    00:45   0:00 sleep 336s
root       81234  0.0  0.0   6432   724 pts/0    S+   00:46   0:00 grep --color=auto sleep
root@ubuntu2004:~# pstree -p 81223
sh(81223)───bash(81224)───sleep(81225)
```

Behaving as intended via CronD.

packaging/installer/netdata-updater.sh
```diff
@@ -185,8 +185,8 @@ update() {
       do_not_start="--dont-start-it"
     fi
 
-    if [ -n "${NETDATA_SELECTED_DASHBOARD}" ] ; then
-        env="NETDATA_SELECTED_DASHBOARD=${NETDATA_SELECTED_DASHBOARD}"
+    if [ -n "${NETDATA_SELECTED_DASHBOARD}" ]; then
+      env="NETDATA_SELECTED_DASHBOARD=${NETDATA_SELECTED_DASHBOARD}"
     fi
```
Ferroin, line 190: Not critical for this PR, but this block can actually be removed. It's left over from when we had the switcher script for the dashboard, I just forgot to remove it when that got removed.

prologic: Put up a PR to clean this up then :)

prologic:
Travis CI failures are becuase of:

```
00:00:49.878291  Running BATS file..
00:00:49.896820  1..3
00:02:22.894917  ok 1 install stable netdata using kickstart
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```

Unintended side-effect of introducing a random delay.

prologic:
This can be fixed by disabling Bash's `$RANDOM` special variable by doing:

```#!sh
unset RANDOM; export RANODM=0
```

This permeates to subshells too;

```#!sh
$ unset RANDOM; export RANDOM=0; ( echo $RANDOM ); /bin/sh -c 'echo $RANDOM'
0
14598
```

PR #9092, opened 2020-05-18
Global GUID lookup

### LINKED ISSUE #8900: Implement GUID map
##### Summary

Implement a global (across hosts) lookup structure that will 

- Map UUIDs (128 bit) to objects (hosts/charts/dimensions) as defined in the [design](https://github.com/netdata/docs/blob/master/architecture/agent/db/database-log-format-pros-cons.md).

- Map unique name identifiers (e.g. concatenated ids of host|chart|dimension) of objects to UUIDs (128 bit) Ref: [here](https://github.com/netdata/netdata/issues/8183)

Find suitable hash map library. Consider:
- Judy
- Sparsehash
- Android libcutils/hashmap
- khash
- AVL?

or something else.

We decided to use Judy + AVL to reuse existing libraries.
Risk: High memory and CPU overheads.

stelfrag:
Fixes #8900
##### Summary
Adds functions that will 
- Add a UUID (index) and string value
- Add a string value (index) and UUID
- Search by UUID or string object

##### Test Plan
- The functions are not used yet. Further details on the issues that will use of them.


Makefile.am
```diff
@@ -184,6 +184,13 @@ LIBNETDATA_FILES = \
     libnetdata/string/utf8.h \
     $(NULL)
 
+if ENABLE_DBENGINE
+    LIBNETDATA_FILES += \
+    libnetdata/global_uuid_map/global_uuid_map.c \
+    libnetdata/global_uuid_map/global_uuid_map.h \
```
mfundul, line 190: I believe those functions shouldn't be part of `/libnetdata`, cause no plugin that links against this code will ever use it. This code is exclusive to the core of the agent. Those functions are only called when dbengine code needs to be called as well.

One suggestion is under `database/engine/global_uuid_map`.

stelfrag: Good point-- I will move it there, since it makes sense only with DBENGINE active anyway

stelfrag: Moved

database/engine/global_uuid_map/global_uuid_map.c
```diff
@@ -0,0 +1,76 @@
...
+#ifdef NETDATA_INTERNAL_CHECKS
+    static uint32_t count = 0;
+#endif
+    if (unlikely(!object))
+        return 0;
+
+    Pvoid_t *PValue;
+
+    PValue = JudyHSIns(&JGUID_map, (void *) uuid, (Word_t) sizeof(uuid_t), PJE0);
+    if (*PValue)
```
amoss, line 19: This will crash if Judy returns an error condition. For example if `uuid==0` then it will trigger an internal error inside `JudyLIns` which returns `~0`. Always dereferencing seems dangerous - should this be `if (PValue!=PPJERR && *PValue)` ?

mfundul: Currently netdata always crashes when out of memory like in:
https://github.com/netdata/netdata/blob/98c7260d929a2f6ad2aabb4b27f0d77d49f6cae8/libnetdata/libnetdata.c#L86

We are consistently crashing in dbengine as well, but we're not logging it similarly.
If we are to be consistent we could add:
```
    if (PPJERR==PValue) fatal("JudyHSIns() cannot allocate memory.");
```

database/engine/global_uuid_map/global_uuid_map.c
```diff
@@ -0,0 +1,76 @@
...
+int guid_find(uuid_t uuid, char *object, size_t max_bytes)
+{
+    Pvoid_t *PValue;
+
+    PValue = JudyHSGet(JGUID_map, (void *) uuid, (Word_t) sizeof(uuid_t));
+    if (unlikely(!PValue))
+        return 1;
+
+    if (likely(object && max_bytes))
+        strncpyz(object, (char *) *PValue, max_bytes);
```
amoss, line 53: This may write `max_bytes+1` bytes

mfundul: Indeed. This should be `memcpy()`.

mfundul: or just use fewer bytes!

mfundul: `strncpyz()` is undocumented this way, maybe we should do something about this. It shouldn't be using more than max bytes, people forget this all the time.

database/engine/global_uuid_map/global_uuid_map.c
amoss: This would cover the out of memory case, but there are multiple conditions in Judy that trigger that error return. The other one that I saw in the review was parameter error. In this case we could guard against the parameter error explicitly, or guard against the error being returned from Judy. I think guarding against the error returning from Judy is overall more robust.

stelfrag: The `if (PPJERR==PValue) fatal ...` has been added.

database/engine/global_uuid_map/global_uuid_map.c
stelfrag: Adjusted to copy -1 bytes

Issue #8784, opened 2020-04-21
Install netdata from EPEL repository for CentOS 7


Nirjonadda:
We have installed netdata from Linux manually Installation guide.

Enabled the EPEL repo
Enabled the additional repo from okay.network

And installed the minimum required dependencies

`yum install autoconf automake curl gcc git libmnl-devel libuuid-devel openssl-devel libuv-devel lz4-devel Judy-devel make nc pkgconfig python zlib-devel cmake libwebsockets-devel`

Then Installed Netdata.

`cd /root/netdata`

`./netdata-installer.sh`

But my question is how can update or install netdata from EPEL repository for CentOS 7?

`yum install netdata`

Do you need uninstall netdata from previous installation or can be install without uninstall?

Ferroin:
You will need to uninstall the copy that you installed using `./netdata-installer.sh` if you want to use the EPEL package.

Nirjonadda:
One more question, what the different between RPM `./netdata-installer.sh` vs YUM EPEL package?

Nirjonadda:
OK we have installed EPEL package but `http://Server IP:19999/ `are not loading.

Unable to connect

Firefox can’t establish a connection to the server at IP:19999.

```
[root@na ~]# systemctl status netdata
● netdata.service - Real time performance monitoring
   Loaded: loaded (/usr/lib/systemd/system/netdata.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2020-04-22 01:05:48 +06; 6s ago
  Process: 15060 ExecStartPre=/bin/chown -R netdata:netdata /var/run/netdata (code=exited, status=0/SUCCESS)
  Process: 15057 ExecStartPre=/bin/mkdir -p /var/run/netdata (code=exited, status=0/SUCCESS)
  Process: 15054 ExecStartPre=/bin/chown -R netdata:netdata /var/cache/netdata (code=exited, status=0/SUCCESS)
  Process: 15052 ExecStartPre=/bin/mkdir -p /var/cache/netdata (code=exited, status=0/SUCCESS)
 Main PID: 15063 (netdata)
   CGroup: /system.slice/netdata.service
           ├─15063 /usr/sbin/netdata -P /var/run/netdata/netdata.pid -D
           ├─15156 /usr/bin/bash /usr/libexec/netdata/plugins.d/tc-qos-helper...
           ├─15161 /usr/libexec/netdata/plugins.d/apps.plugin 1
[...]

PR #9108, opened 2020-05-19
Package: obsoletes conflicting EPEL packages (#6879 #8784)

### LINKED ISSUE #6879: RPM fails to update, conflicts with ?itself?.  Packaging failure...
##### Bug report summary
On nightly "yum update", netdata failed to update via RPM. 
In short, many files seem to have "conflicted with themselves", including the directory /etc/netdata/.  Here's one error: 

> Transaction check error:
>   file /etc/netdata from install of **netdata**-1.17.1-1.el7.x86_64 conflicts with file from package **netdata-conf**-1.16.0-1.el7.noarch
> ... snip... 
Full transcript attached, 
[netdataFailedUpdate.txt](https://github.com/netdata/netdata/files/3616706/netdataFailedUpdate.txt)
(I suspect you've attempted to consolidate the existing "netdata-conf" & "netdata-data" RPMs into a single "netdata" RPM, without first uninstalling the other two?  Ya, that's gonna end badly.)

##### OS / Environment
Under the latest CentOS, 7.6.1810 in a VM, under oVirt (eg KVM); patched nightly.   We have a very limited installed base, just 6 machines for now: our Dev, Test, and Prod HAproxy servers.  All were installed via Ansible.   This is the first update we've ever installed; we only started using netdata Aug 1, 2019. 

##### Netdata version (output of `netdata -V`)
Current version: 1.16.0
Trying to update to 1.17.1-1 via RPM, on EPEL

##### Component Name
netdata's RPM packages

##### Steps To Reproduce
This problem occurred consistently on all 6 machines on which we had it installed.   I suggest, although I did not try: 

- install latest CentOS7 (c/w the EPEL repo) 
- yum install netdata-1.16.0  
- yum update -y

##### Expected behavior
"yum update -y" should update to the latest version, v1.17.1, without complaint.

##### Failed work-around:
I was initially not awake enough to realize that netdata consists of 3 separate RPM packages. I tried just removing the most obvious "yum remove netdata -y; yum update -y; yum install netdata", but I got the same errors - conflicts with netdata-data & netdata-conf (of course).

##### Work-around:
I tried to remove the entire thing, then re-install it again at the newest version:
- yum remove netdata* -y 
- yum install netdata -y

Extra credit: I ran our ansible playbook against this, to get the haproxy plugin re-activated... but I see you've also changed the structure of the directories! /etc/netdata/conf.d/ no longer exists under v1.17.1.   That's unexpected, seems undocumented.

More Extra credit: Under v1.16.0, I found that by setting: 
```
[plugins]
haproxy = yes
```
...into /etc/netdata/netdata.conf, then restarting the netdata service, it would auto-populate /etc/netdata/conf.d/ with a tree of other (config) files.   It no longer does this.  That original feature seemed to be undocumented.

Saruspete:
##### Summary
EPEL is splitting the main netdata package into 3:
- **netdata** that gets updated by our upstream package and is the core
- **netdata-data** that contains the dashboard (/usr/share/)
- **netdata-conf** that contains the configuratio (/etc/netata)

When users try to use netdata installer, it'll update only the "netdata" package. As this package also provides `/etc/netdata` and `/usr/share` files (currnetly belonging to netdata-conf and netdata-data) it'll fail during the post-transaction check.

Using "obsoletes" keyword, if official netdata package is installed, it'll also remove the 2 conflicting epel packages.

I didn't want to add the keyword "Provides", and rather break external packages that are dependant on EPEL versions.
That way, they can either:
* update their `Require` field to specify a file if they only need it (instead of a package)
* fail explicitely if they absolutely need the EPEL package (and potentially a patch in their tree) instead of a silent potential failure later.

Fixes #6879 #8784

##### Component Name
netdata.spec.in

##### Test Plan
- install epel:  `sudo yum install epel-release`
- install all 3 epel netdata packages: ` sudo yum install netdata netdata-{data,conf}`
- build netdata from source with script: `git clone https://github.com/netdata/netdata.git && netdata/contrib/rhel/build-netdata-rpm.sh`
- install netdata built rpm: `sudo yum install $HOME/rpmbuild/RPMS/netdata-v1*.rpm

This should specify the removal of **netdata-conf** (and/or **netdata-data**):
```
Installing:
 netdata                            x86_64                            v1.22.2-1.fc29                               @commandline                            9.8 M
     replacing  netdata-conf.noarch 1.18.1-1.fc29
```

### DIFF
--- a/netdata.spec.in
+++ b/netdata.spec.in
@@ -93,6 +93,10 @@ Group:		Applications/System
 Source0:	https://github.com/netdata/%{name}/releases/download/%{version}/%{name}-%{version}.tar.gz
 URL:		http://my-netdata.io
 
+# Remove conflicting EPEL packages
+Obsoletes:  %{name}-conf
+Obsoletes:  %{name}-data
+
 # #####################################################################
 # Core build/install/runtime dependencies
 # #####################################################################


Saruspete:
@prologic the error shown in the centos8 build has nothing to do with this change. 

```
Error: Unable to find a match: autoconf-archive json-c-devel libmnl-devel
```

can you try to restart the build ?

prologic:
You are right @Saruspete :) I've kicked off the CI build again, let's see.

With regards to the Test Plan; there's nothing wrong with it; I just want to see that it was done by either you or @Ferroin as I feel this is is a significant change and yes you are right we don't actually have good CI coverage here :/

Saruspete:
thanks @prologic seems the build is ok now.

For the test, the last block comes from my desktop (fedora 29, hence .fc29 in package name). I did a git-tag 1.22.2 with the test, and installed the package netdata-conf. It generated me the following output, and did remove the EPEL package:
```
Installing:
 netdata                            x86_64                            v1.22.2-1.fc29                               @commandline                            9.8 M
     replacing  netdata-conf.noarch 1.18.1-1.fc29
```

Regarding the keyword, the [official rpm doc](https://rpm.org/user_doc/dependencies.html) states the only difference is when using `rpm -i` instead of `rpm -U`:

> RPM removes all packages matching obsoletes of packages being installed. As it sees the obsoleting package as their updates. There has not to be a one to one relation ship between obsoleting and obsoleted packages. Note that rpm -i does not do updates and though treat Obsoletes: as Conflicts (since rpm-4.10). For most cases rpm -i should be avoided and rpm -U should be used unless this specific behavior is desired (e.g. for kernels).

> Updaters have a bit different view on Obsoletes: as they need to find out what packages to install as an update. As a result packages containing matching Obsoletes: are added as updates and replace the matching packages.

That shouldn't make any issue, as it's officially not recommended to use `rpm` directly as it does not fetch dependencies, and if someone is using the `-i` option, it would fail nonetheless if the package `netdata` from epel is already installed.

I've mailed the epel maintainer to see if he could place a `Depends: netdata =epel...` on the conf & data packages, so that if the main netdata package is removed / updated, this version-locked dependency would fail too, and either block (if they are not removed by our `Obsoletes` field) or be removed (as dependency will fail).

didier13150:
Hi,
I can't see where the problem is. You explicitly install netdata-conf and netdata-data packages, which are dependencies of netdata package and installed when you install the netdata package, but this packages are marked as dependencies to netdata package by rpm, and then automatically removed when netdata package is removed (just tested on fedora32, not epel7 with yum or epel8 with dnf).
If you explicitly install netdata-conf and netdata-data packages, they survive to netdata remove process.
netdata-conf and netdata-data cannot depend to netdata, which create circular dependencies. Reverse the dependencies is not a good solution.
I check ASAP if creating a meta package can solve all problems and I report you on this PR.
Cheers,
Didier FABERT (fedora and epel netdata maintainer)

Saruspete:
@didier13150 the 2 packages -conf & -data survival is the issue: they provide the same files that upstream netdata package provides, which conflicts and fails the install process. 
See the 2 bugs reported for full detail and output.

If the removal of the meta package removes also all related packages (like -core -data -conf), yeah that would work too.

Cheers

Issue #7172, opened 2019-10-23
Cleanup dynamic memory on exit in Exporting engine


vlvkobal:
- [x] A lot of allocated blocks to clean with a complex data structure.
- [x] Think about `netdata_thread_cleanup_pop()` in the main thread and implement the cleanup.
- [x] ~~Write unit tests.~~

It will involve going through every one of the 7 connectors and the engine code.

vlvkobal:
Done.

PR #9112, opened 2020-05-20
Dynamic memory cleanup for Pub/Sub exporting connector

vlvkobal:
##### Summary
Part of #7172

##### Component Name
exporting engine

##### Test Plan
Configure a Pub/Sub exporting connector instance. Run Netdata. Kill it with SIGTERM.

### DIFF
--- a/exporting/pubsub/pubsub.c
+++ b/exporting/pubsub/pubsub.c
@@ -57,6 +57,28 @@ int init_pubsub_instance(struct instance *instance)
 }
 
 /**
+ * Clean a PubSub connector instance
+ *
+ * @param instance an instance data structure.
+ */
+void pubsub_cleanup(struct instance *instance)
+{
+    info("EXPORTING: cleaning up instance %s ...", instance->config.name);
+
+    struct pubsub_specific_data *connector_specific_data =
+        (struct pubsub_specific_data *)instance->connector_specific_data;
+
+    buffer_free(instance->buffer);
+
+    freez(connector_specific_data);
+
+    info("EXPORTING: instance %s exited", instance->config.name);
+    instance->exited = 1;
+
+    return;
+}
+
+/**
  * Pub/Sub connector worker
  *
  * Runs in a separate thread for every instance.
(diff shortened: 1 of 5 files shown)


exporting/pubsub/pubsub.c
```diff
@@ -56,6 +56,35 @@ int init_pubsub_instance(struct instance *instance)
...
+ *
+ * @param instance an instance data structure.
+ */
+void clean_pubsub_instance(struct instance *instance)
+{
+    info("EXPORTING: cleaning up instance %s ...", instance->config.name);
+
+    struct pubsub_specific_data *connector_specific_data =
+        (struct pubsub_specific_data *)instance->connector_specific_data;
+    pubsub_cleanup(connector_specific_data);
```
thiagoftsm, line 70: Should we test `connector_specific_data` before to clean?

exporting/pubsub/pubsub.c
```diff
@@ -56,6 +56,35 @@ int init_pubsub_instance(struct instance *instance)
...
+    freez(connector_specific_data);
+
+    buffer_free(instance->buffer);
```
thiagoftsm, line 73: The same thing here, should we test `instance->buffer` before to clean?

exporting/pubsub/pubsub.c
```diff
@@ -56,6 +56,35 @@ int init_pubsub_instance(struct instance *instance)
...
+
+    struct pubsub_specific_config *connector_specific_config =
+        (struct pubsub_specific_config *)instance->config.connector_specific_config;
+    freez(connector_specific_config->credentials_file);
```
thiagoftsm, line 77: the sam question for all the connector_specific_config.

exporting/pubsub/pubsub.c
```diff
@@ -69,13 +98,16 @@ void pubsub_connector_worker(void *instance_p)
...
 
-    while (!netdata_exit) {
+    while (!instance->engine->exit) {
         struct stats *stats = &instance->stats;
         char error_message[ERROR_LINE_MAX + 1] = "";
 
         uv_mutex_lock(&instance->mutex);
         uv_cond_wait(&instance->cond_var, &instance->mutex);
 
+        if (unlikely(instance->engine->exit))
```
thiagoftsm, line 108: I think that when we are exiting here, we also need to unlock the mutex, I already changed this in other PR.

exporting/pubsub/pubsub.c
vlvkobal: `buffer_free` checks for NULL values.

exporting/pubsub/pubsub.c
vlvkobal: There are so many places where we don't check the pointer. It's pointless to check it on exit.

exporting/pubsub/pubsub.c
vlvkobal: The same answer 😃.

exporting/pubsub/pubsub.c
vlvkobal: Fixed.

exporting/pubsub/pubsub.c
thiagoftsm: I was thinking in the possibility that another contributor could forget to initialize them in the future and we were taking care to avoid clean what was never allocated.

thiagoftsm:
Please, remove the file `2`.

PR #9120, opened 2020-05-21
Generate global random GUIDs

### LINKED ISSUE #8901: Add randomized GUID
Add randomized GUIDs associated with all object creations for use with the [metadata log](https://github.com/netdata/netdata/issues/8183). 

- Call for each new `RRDHOST`, `RRDSET`, `RRDDIM`.
  - Search the GUID map
    - If it exists, reuse it
      - It will exist by the metadata log replay that populated the structure
    - If it doesn't exist, create and store in the structures and metadata log.
      - Use support in `libuuid` to generate random UUIDs.
- Avoid changing the structures by putting the uuids in the volatile handles (preserving memory mapped files for other modes).
- Handle the preexisting GUIDs already generated by the dbengine code (`rrdeng_store_metric_init`) so that they correctly map to the newly created randomized GUID.
  - This is to avoid migrating dbengine files after netdata update.
  - take the pseudo GUID and put it into the metadata log

stelfrag:
Fixes #8901
##### Summary
- Starts to generate random GUIDs for metrics
  - Stores new GUID in the global lookup table 
  - Keeps a copy in `dim->state->metric_uuid`
    - Exports it to the metadata log for the dimension instead of `dim->state->rrdeng_uuid`
- Starts to generate random GUIDs for charts
  - Stores in `set->chart_uuid`
  - Exports in metadata log (`metalog_commit_update_chart`)
- Generates GUID for host as well in `host->host_guid` 
  - This may not be needed since we have machine GUID 
     - Unless we plan to have a way not to lose data on machine GUID change

##### Test Plan
- Compile with `-D NETDATA_INTERNAL_CHECKS`
- GUIDs are reported in the `error.log`
  - Written in the metadata log

##### Additional Information
- Bulk load of GUIDs for the global map needs to happen on agent init / metadata loading
  - Metric init will detect and reuse the GUID from the global map
- The object format for building the global map is:

###### Metric
`machine_guid`/`chart->id`/`dim->id`

###### Chart
`machine_guid`/`chart->id

Example: 

```
Chart GUID [cc9bd5eb-357c-4af4-9b9b-8abcc5987ebf] on [d6b37186-74db-11ea-88b2-0bf5095b1f9e/services.throttle_io_write]
Metric GUID [79ab5619-066a-4a99-9a1e-a98f232eecc2] on [d6b37186-74db-11ea-88b2-0bf5095b1f9e/services.throttle_io_write/system.slice_irqbalance.service]
```


database/engine/global_uuid_map/global_uuid_map.c
```diff
@@ -27,11 +42,15 @@ int guid_store(uuid_t uuid, char *object)
     if (PPJERR == PValue)
         fatal("JudyHSIns() fatal error.");
     if (*PValue == NULL) {
-        uuid_t *value = mallocz(sizeof(uuid_t));
-        memcpy(value, &uuid, sizeof(uuid_t));
+        uuid_t *value = (uuid_t *) mallocz(sizeof(uuid_t));
+        uuid_copy(*value, uuid);
```
amoss, line 46: This is less clear than the old code because a `uuid_t` is not really a value but a pointer.

database/engine/global_uuid_map/global_uuid_map.c
```diff
@@ -4,17 +4,32 @@
...
+
+}
+
+void guid_bulk_load_unlock()
+{
+    uv_rwlock_wrunlock(&global_lock);
+}
+
+
+static inline int guid_store_internal(uuid_t uuid, char *object, int lock)
```
amoss, line 24: It would be good to match the existing style / structure in the dbegine. I think that the convention currently would be to have the internal call called `_nolock` and put the locking in the external wrapper.

database/engine/global_uuid_map/global_uuid_map.c
```diff
@@ -27,11 +42,15 @@ int guid_store(uuid_t uuid, char *object)
...
-        memcpy(value, &uuid, sizeof(uuid_t));
+        uuid_t *value = (uuid_t *) mallocz(sizeof(uuid_t));
+        uuid_copy(*value, uuid);
         *PValue = value;
     }
+    if (likely(lock))
+        uv_rwlock_wrunlock(&global_lock);
+
 #ifdef NETDATA_INTERNAL_CHECKS
+    static uint32_t count = 0;
```
amoss, line 53: The `info` should be a `debug`.

database/rrd.h
```diff
@@ -311,6 +311,7 @@ struct rrddim_query_handle {
 struct rrddim_volatile {
 #ifdef ENABLE_DBENGINE
     uuid_t *rrdeng_uuid;                 // database engine metric UUID
+    uuid_t *metric_uuid;                 // global UUID for this metric (unique_across hosts)
```
amoss, line 314: Why add a second GUID to dimensions - should we not use the same one?

database/rrd.h
```diff
@@ -458,8 +459,8 @@ struct rrdset {
 
     char *plugin_name;                              // the name of the plugin that generated this
     char *module_name;                              // the name of the plugin module that generated this
-
-    size_t unused[5];
+    uuid_t *chart_uuid;                             // Store the global GUID for this chart
```
amoss, line 462: It is too late to change this, but be aware that `size_t` is not guaranteed to be the pointer size across all architectures (and that `uintptr_t` is more appropriate for this use).

database/rrddim.c
```diff
@@ -350,15 +349,15 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
...
+        rd->state->collect_ops.init = rrdeng_store_metric_init;
         rd->state->collect_ops.store_metric = rrdeng_store_metric_next;
-        rd->state->collect_ops.finalize     = rrdeng_store_metric_finalize;
-        rd->state->query_ops.init           = rrdeng_load_metric_init;
-        rd->state->query_ops.next_metric    = rrdeng_load_metric_next;
-        rd->state->query_ops.is_finished    = rrdeng_load_metric_is_finished;
-        rd->state->query_ops.finalize       = rrdeng_load_metric_finalize;
-        rd->state->query_ops.latest_time    = rrdeng_metric_latest_time;
-        rd->state->query_ops.oldest_time    = rrdeng_metric_oldest_time;
+        rd->state->collect_ops.finalize = rrdeng_store_metric_finalize;
```
amoss, line 354: Lol. @mfundul's code has finally encountered the linter...

mfundul:
As discussed we should add some kind of object in the map instead of a composite string.
```
void foo(char *GUID, char *objectstr)
{
    char words[3];
    char *strobject = "MACHINEGUID\nTYPERRDSET\nRRDIMENSION";
    char *strobject2 = "MACHINEGUID\nTYPERRDSET";
    char *strobject3 = "MACHINEGUID";
    
    struct object {
        int type; // HOST, SET, DIM
        void *objectp; // RRDHOST, RRDSET, RRDDIM
    } object = malloc();
    unparse(objectstr, &type, words);

    object->type = type;
        switch(type) {
        case HOST:
            object->objectp = rrdhost_find(word[0]); 
            break;
        case SET:
            host = rrdhost_find(word[0]);
            object->objectp = rrdset_find(host, word[1]);
            break;
        case DIM:
            host = rrdhost_find(word[0]);
            st = rrdset_find(host, word[1]);
            object->objectp = rrddim_find(host, st, word[1]);
            break;
        default:
            assert(0);
            break;
    }
}
```

database/rrdset.c
```diff
@@ -769,6 +769,21 @@ RRDSET *rrdset_create_custom(
 
     rrdsetcalc_link_matching(st);
     rrdcalctemplate_link_matching(st);
+#ifdef ENABLE_DBENGINE
+    BUFFER *object = buffer_create(512);
+    buffer_sprintf(object, "%s/%s", host->machine_guid, st->id);
```
amoss, line 774: This seems a bit heavy - is it not cheaper to just `strlen(st->id)`, add the fixed length of the guid to that and `mallocz` the memory here?

database/engine/global_uuid_map/global_uuid_map.c
```diff
@@ -74,7 +117,45 @@ int find_guid_by_object(char *object, uuid_t *uuid)
...
+    if (init)
+        return;
+
+    init = 1;
+    info("Configuring locking mechanism for global GUID map");
+    assert(0 == uv_rwlock_init(&guid_lock));
+    assert(0 == uv_rwlock_init(&object_lock));
+    assert(0 == uv_rwlock_init(&global_lock));
+
+    int rc = guid_bulk_load("6fc56a64-05d7-47a7-bc82-7f3235d8cbda","d6b37186-74db-11ea-88b2-0bf5095b1f9e/cgroup_qemu_ubuntu18.04.cpu_per_core/cpu3");
```
amoss, line 153: What is this - test data?

database/rrddim.c
mfundul: I was just copying Kosta's style in the surrounding code, following the initial coding guidelines (before we decided to formalize a clang format).

database/rrdhost.c
```diff
@@ -259,6 +259,19 @@ RRDHOST *rrdhost_create(const char *hostname,
 
             return host;
         }
+        BUFFER *object = buffer_create(512);
+        buffer_sprintf(object, "%s", host->machine_guid);
+        host->host_uuid = callocz(1, sizeof(uuid_t));
+        if (unlikely(find_or_generate_guid((char *)buffer_tostring(object), host->host_uuid))) {
```
amoss, line 265: Why use the buffer if it only contains the `host->machine_guid` unchanged?

database/rrddim.c
amoss: I prefer the aligned equals as they make it easier to read, but they are a dead giveaway that the code has not been linted.

database/rrdset.c
mfundul: Yes it is cheaper, not by much though. In the common case `buffer_create()` will just be a `malloc()` and you won't have to explicitly call `strlen()`. It's probably not a huge difference.

In both cases the `sprintf()` will trigger `strlen()` calls anyways. I'm okay either way, although I prefer `char *` instead of `BUFFER *` optically.

amoss:
> As discussed we should add some kind of object in the map instead of a composite string.
> 
> ```
> void foo(char *GUID, char *objectstr)
> {
>     char words[3];
>     char *strobject = "MACHINEGUID\nTYPERRDSET\nRRDIMENSION";
>     char *strobject2 = "MACHINEGUID\nTYPERRDSET";
>     char *strobject3 = "MACHINEGUID";
>     
>     struct object {
>         int type; // HOST, SET, DIM
>         void *objectp; // RRDHOST, RRDSET, RRDDIM
>     } object = malloc();
>     unparse(objectstr, &type, words);
> 
>     object->type = type;
>         switch(type) {
>         case HOST:
>             object->objectp = rrdhost_find(word[0]); 
>             break;
>         case SET:
>             host = rrdhost_find(word[0]);
>             object->objectp = rrdset_find(host, word[1]);
>             break;
>         case DIM:
>             host = rrdhost_find(word[0]);
>             st = rrdset_find(host, word[1]);
>             object->objectp = rrddim_find(host, st, word[1]);
>             break;
>         default:
>             assert(0);
>             break;
>     }
> }
> ```

This looks like an attempt at RTTI. Before we go down that road - is there any point that we need to look up a GUID without knowing the type of object that we are looking for?

mfundul:
> 
> 
> > As discussed we should add some kind of object in the map instead of a composite string.
> > ```
> > void foo(char *GUID, char *objectstr)
> > {
> >     char words[3];
> >     char *strobject = "MACHINEGUID\nTYPERRDSET\nRRDIMENSION";
> >     char *strobject2 = "MACHINEGUID\nTYPERRDSET";
> >     char *strobject3 = "MACHINEGUID";
> >     
> >     struct object {
> >         int type; // HOST, SET, DIM
> >         void *objectp; // RRDHOST, RRDSET, RRDDIM
> >     } object = malloc();
> >     unparse(objectstr, &type, words);
> > 
> >     object->type = type;
> >         switch(type) {
> >         case HOST:
> >             object->objectp = rrdhost_find(word[0]); 
> >             break;
> >         case SET:
> >             host = rrdhost_find(word[0]);
> >             object->objectp = rrdset_find(host, word[1]);
> >             break;
> >         case DIM:
> >             host = rrdhost_find(word[0]);
> >             st = rrdset_find(host, word[1]);
> >             object->objectp = rrddim_find(host, st, word[1]);
> >             break;
> >         default:
> >             assert(0);
> >             break;
> >     }
> > }
> > ```
> 
> This looks like an attempt at RTTI. Before we go down that road - is there any point that we need to look up a GUID without knowing the type of object that we are looking for?

One way this will be useful is if we have an array of metadata log records that's essentially only GUIDs, e.g. during compaction.

database/engine/global_uuid_map/global_uuid_map.c
mfundul: yes

database/rrd.h
mfundul: Well that's been in netdata since forever and it's part of the on-disk signatures for `memory mode = save/map` :(

mfundul: Btw Judy has a very nice type `Word_t`.

amoss:
> > This looks like an attempt at RTTI. Before we go down that road - is there any point that we need to look up a GUID without knowing the type of object that we are looking for?
> 
> One way this will be useful is if we have an array of metadata log records that's essentially only GUIDs, e.g. during compaction.

In order to reach that point we must have already read and parsed the existing log structure. It seems that any processing that we need to do may have special cases according to the type of object, so it makes sense to maintain those types as a result of the loading the journal.

Adding an intermediate structure also adds a point of failure. Every piece of code that touches that structure needs to handle all three cases. Segregating the data instead is an approach that is correct by design.

I recommend avoiding this dynamic type information unless it is absolutely necessary - and that means identifying that we will have situations where both of the following are true:

* We need to know the type of the object being retrieved from the map.
* In a code-path where that information is not already available.

Otherwise we are adding redundancy and we need to ensure that we do not lose sync.

mfundul:
> 
> 
> > > This looks like an attempt at RTTI. Before we go down that road - is there any point that we need to look up a GUID without knowing the type of object that we are looking for?
> > 
> > 
> > One way this will be useful is if we have an array of metadata log records that's essentially only GUIDs, e.g. during compaction.
> 
> In order to reach that point we must have already read and parsed the existing log structure. It seems that any processing that we need to do may have special cases according to the type of object, so it makes sense to maintain those types as a result of the loading the journal.
> 
> Adding an intermediate structure also adds a point of failure. Every piece of code that touches that structure needs to handle all three cases. Segregating the data instead is an approach that is correct by design.
> 
> I recommend avoiding this dynamic type information unless it is absolutely necessary - and that means identifying that we will have situations where both of the following are true:
> 
>     * We need to know the type of the object being retrieved from the map.
> 
>     * In a code-path where that information is not already available.
> 
> 
> Otherwise we are adding redundancy and we need to ensure that we do not lose sync.

We're going to have a memory footprint and CPU usage problems if we're going to be parsing everything again during compaction. All those calculations can be avoided if the lookup structure points to memory, since it is populated only once during metadata log replay. But we can postpone such optimizations for after this increment is finished.

amoss:
> We're going to have a memory footprint and CPU usage problems if we're going to be parsing everything again during compaction. All those calculations can be avoided if the lookup structure points to memory, since it is populated only once during metadata log replay. But we can postpone such optimizations for after this increment is finished.

If we don't reparse the contents of the journal files during compaction then how do we know what it is in them?

mfundul:
> 
> 
> > We're going to have a memory footprint and CPU usage problems if we're going to be parsing everything again during compaction. All those calculations can be avoided if the lookup structure points to memory, since it is populated only once during metadata log replay. But we can postpone such optimizations for after this increment is finished.
> 
> If we don't reparse the contents of the journal files during compaction then how do we know what it is in them?

We can only parse the GUIDs, or if we feel like we can afford to use more memory we can keep records in memory (arrays of GUIDs per file).

amoss:
In order to know which parts of the log are relevant we need to check every record inside it. Having checked every record we then know the structure that the log describes, and the type of every object defined by a GUID. I don't see how we save memory by skipping that, but introducing an intermediate structure that we store at all times to define the type of the GUID. The downside that I see is that trying to optimize in this way introduces a point of failure by mixing the state of the structures in memory now and the state defined in the log.

mfundul:
We're not saving memory if we're storing the records in memory, we're just saving CPU time and I/O time (from not calling the replay log parser that needs to do all the lookups again).

Alternatively we can do minimal parsing by only getting the GUIDs but still suffer from the I/O overhead, but this way we don't use any additional memory and still save CPU time by not doing multiple lookups per record.

Additionally, I was thinking that it could be somewhat useful if our codebase was able to find an object pointer by its GUID for future proofing of the whole database remake we're planning.

Otherwise, we keep the lookup structures simple and redo all lookups like you suggested.

You can make the final call at this point so that we can move forward.

amoss:
I'm not going to make a final call on this one - you and @stelfrag are writing the code and have a better view of what is necessary than me. I'm trying to point out weaknesses in the way that you are going - but ultimately it is up to the both of you to weigh them, decide which set of factors is more relevant and make the decision. Autonomy and trust :)

I think a point to clarify is that we've both said "memory usage" but your last description suggests that we are talking about two different points in time. I mean memory usage throughout the lifetime of the agent, and I think that you mean memory usage during the compaction process.

mfundul:
> 
> 
> I'm not going to make a final call on this one - you and @stelfrag are writing the code and have a better view of what is necessary than me. I'm trying to point out weaknesses in the way that you are going - but ultimately it is up to the both of you to weigh them, decide which set of factors is more relevant and make the decision. Autonomy and trust :)
> 
> I think a point to clarify is that we've both said "memory usage" but your last description suggests that we are talking about two different points in time. I mean memory usage throughout the lifetime of the agent, and I think that you mean memory usage during the compaction process.

I'm leaving this on @stelfrag 's judgement. Stelios you can balance what you prefer with what's safer for us to hit our deadlines.

stelfrag:
The GUID for charts and metrics maps to a 33 and 49 byte value respectively. The value
eg for the metric which used to be a string concatenation of the 

`machine_guid/chart_id/dimension_id/ now maps to the binary representations of those as:

Index : GUID (in JGUID_map)

- byte 0 (GUID_TYPE)
- bytes 1-16 machine_guid
- bytes 17-32 chart_guid
- bytes 33-48 dimention_guid

This value also maps back to the GUID (JGUID_map_object)

amoss:
I think this addresses part of the conversation, but my concern was more direct:

Do we need to wrap each value in a struct to carry RTTI or do we know the type of each object that the GUID refers to from the calling context?

mfundul:
> 
> 
> I think this addresses part of the conversation, but my concern was more direct:
> 
> Do we need to wrap each value in a struct to carry RTTI or do we know the type of each object that the GUID refers to from the calling context?

The syntax of plugins.D , e.g. when replaying a metadata log, does not guarantee the type of the object. Infering it by postponing plugins.D parsing until the N subsequent commands can also be ambiguous so it's not safe either. Finally, it would make parsing needlessly complicated.

amoss:
Can you give a concrete example of the syntax where this is a problem?

mfundul:
> 
> 
> Can you give a concrete example of the syntax where this is a problem?

```
CONTEXT 10000000:1000:1000:1000:00000000001
```

Given the language definition the context could be:
1. another dimension
2. a chart
3. a host

In all three scenarios it may be valid to parse `DIMENSION` succesfully. If the GUID lookup is successful the only error conditions in this example are:
1. no error conditions (chart state is same as said dimension's parent). This is the same case as the current language implementation declaring multiple `DIMENSION`s sequentially.
2. no error conditions.
3. if the host is different than the current state's host thus resetting the current chart state. This is not applicable yet (multi-host DB).

Inferring the type would mean you also need to parse an additional command, e.g. the subsequent `DIMENSION` and then guessing that the `CONTEXT` needs to be referring to a chart. This also complicates the state machine.

amoss:
Two of these cases do not make sense to me based on my memory of the protocol.

1. Why would a dimension have another dimension as a parent? I cannot see what this would mean in terms of the data-model for the system, how it occurs in the current collector protocol or why we would add this?
2. This seems sensible
3. In a multi-host db why would we have a dimension without a chart. I do not understand why we need this.

mfundul:
> 
> 
> Two of these cases do not make sense to me based on my memory of the protocol.
> 
>     1. Why would a dimension have another dimension as a parent? I cannot see what this would mean in terms of the data-model for the system, how it occurs in the current collector protocol or why we would add this?
> 
>     2. This seems sensible
> 
>     3. In a multi-host db why would we have a dimension without a chart. I do not understand why we need this.

We can simply change the proposed language to `CONTEXT-CHART`, `CONTEXT-DIM`, `CONTEXT-HOST` and make it unambiguous.

My example above was modelled on the internal state of the `plugins.D` parser which uses 3 variables, `host`, `st` and `rd`.

amoss:
The `CONTEXT` is never generated by external code, only by the log generation so we should pick the option that simplifies the code. I like the idea of splitting the `CONTEXT` into separate options but I'm not convinced it is the simplest option.

Each type of object only has one type of parent.

So the result of a `CONTEXT` can be a `void*` and the action that consumes the parent can cast it to the correct type. Is there any danger in this approach or is it the simplest option?

mfundul:
> 
> 
> The `CONTEXT` is never generated by external code, only by the log generation so we should pick the option that simplifies the code. I like the idea of splitting the `CONTEXT` into separate options but I'm not convinced it is the simplest option.
> 
> Each type of object only has one type of parent.
> 
> So the result of a `CONTEXT` can be a `void*` and the action that consumes the parent can cast it to the correct type. Is there any danger in this approach or is it the simplest option?

Well the is no guarantee that there are no bugs, data corruption or other future users of the `CONTEXT` keyword such as streaming (which is also a part of this increment).

The requirement has always been that language parsing is reliable and does not crash.

What Stelios has done here is guarantee that internally (in memory) you will always get the correct type after a lookup. However, there are no `void *` pointers involved in this PR. What you get is a mapping from a GUID to a string identifier. As a result, the user of this code will have to do a manual lookup which will either be:
- `host = rrdhost_find(MACHINE_GUID)`
- `st = rrdset_find(host, chart-fullid)`
- `rd = rrddim_find(st, dimension-id)`

This way, again you are guaranteed there are no crashes or data races.

mfundul:
What I originally asked was object type and `void *` pointers but this is not what this PR does. Which is safer in a way.

amoss:
Ok, so using `CONTEXT-x` would still cause a crash on a (malicious) malformed logfile.

It looks as if the naming in the summary (and your comment) has shifted. In the code it looks as if the internal names are concatenations of GUIDS rather than ids. This is good - using ids would have caused us problems later when we handle labels properly.

It seems that the RTTI is necessary.

PR #9125, opened 2020-05-21
Send anonymous statistics from backends and exporting engine

vlvkobal:
##### Summary
In order to make the transition from backends to the new exporting engine smoother, we need to understand if Netdata users are ready for removing the backends subsystem completely.

Part of #8461

##### Component Name
exporting engine

##### Test Plan
Check `error.log`.

### DIFF
--- a/backends/backends.c
+++ b/backends/backends.c
@@ -697,6 +697,7 @@ void *backends_main(void *ptr) {
     // prepare the backend main loop
 
     info("BACKEND: configured ('%s' on '%s' sending '%s' data, every %d seconds, as host '%s', with prefix '%s')", type, destination, source, global_backend_update_every, hostname, global_backend_prefix);
+    send_statistics("BACKENDS_START", "OK", type);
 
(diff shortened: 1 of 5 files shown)


backends/backends.c
```diff
@@ -697,6 +697,7 @@ void *backends_main(void *ptr) {
     // prepare the backend main loop
 
     info("BACKEND: configured ('%s' on '%s' sending '%s' data, every %d seconds, as host '%s', with prefix '%s')", type, destination, source, global_backend_update_every, hostname, global_backend_prefix);
+    send_statistics("BACKENDS_START", "OK", type);
```
thiagoftsm, line 700: all the `info` and` and `debug` messages are in the singular ( `BACKEND:` ) why are you using plural here?

vlvkobal: I used it because the name of the subsystem is "backends". But, as far as a specific backend is starting here and taking into account what you said, I've changed the name.

PR #9128, opened 2020-05-21
Fix paths to trigger docs CI workflow

joelhans:
##### Summary

Docs workflow wasn't running on PRs that edit `.md` files, but this should fix it with the correct paths.

##### Component Name

area/ci
area/docs

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,10 +5,10 @@ on:
     branches:
       - master
     paths:
-      - '*.md'
+      - '**.md'
   pull_request:
     paths:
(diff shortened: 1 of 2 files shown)


joelhans:
I just verified this ran on the initial PR, which had an edited `.md` file. I've since removed it in the second push so that we don't inadvertently push some fake text to master.

```
=========================> MARKDOWN LINK CHECK <=========================


FILE: docs/getting-started.md
[✖] /fake/link/README.md → Status: 400 [Error: ENOENT: no such file or directory, access '/fake/link/README.md'] {

=========================================================================
```

PR #9129, opened 2020-05-21
Fix link in Prometheus doc

joelhans:
##### Summary

The Learn CI found this broken link, which was merged before we had good link-checking here. This PR fixes the link with the new format.

##### Component Name

area/exporting


joelhans:
Fixed in #9132 instead.

Issue #9093, opened 2020-05-19
Support eBPF plugin for Netdata static64


prologic:
##### Feature idea summary

Now that #8665 is merged and we have a fully working (_on by default_) eBPF plugin for Netdata which should start hitting our nightly builds; we alos need to bring support for eBPF to the Netdata static binaries we build and support through `kickstart-static64.sh`.

##### Expected behavior

Installing Netdata via `kicksart-static64.sh` also packages our eBPF plugin ans associated kernel-collector libraries/ebpf programs and the same functionality works in our staitcally linked Netdata agent binary.

prologic:
cc @thiagoftsm

PR #8865, opened 2020-04-30
Fix macOS builds building and linking against openssl

prologic:
##### Summary

ssia

##### Component Name

- area/packaging

##### Test Plan

<details>
<summary>
Before:
</summary>

```#!sh
$ brew  sh
brew ~/NetData/netdata$ ./netdata-installer.sh --dont-wait --dont-start-it --disable-go --require-cloud --install /tmp/netdata.
.
.
checking if cloud functionality should be enabled... yes
configure: WARNING: OpenSSL required for agent-cloud-link but not found. Try installing 'libssl-dev' or 'openssl-devel'.
checking if libmosquitto static lib is present (and builds)... no
checking if libwebsockets static lib is present... yes
checking if netdata agent-cloud-link can be enabled... no
configure: error: User required agent-cloud-link but it can't be built!
 FAILED
.
.
.
```
</details>

<details>
<summary>
After:
</summary>

```#!sh
$ brew  sh
brew ~/NetData/netdata$ ./netdata-installer.sh --dont-wait --dont-start-it --disable-go --require-cloud --install /tmp/netdata.
.
.
.
To start netdata run:

  /tmp/netdata/netdata/usr/sbin/netdata

Uninstall script copied to: /tmp/netdata/netdata/usr/libexec/netdata/netdata-uninstaller.sh

 --- Installing (but not enabling) the netdata updater tool ---
You need to run the installer as root for auto-updating via cron
 FAILED  Cannot cleanup old netdata updater tool.

Update script is located at /tmp/netdata/netdata/usr/libexec/netdata/netdata-updater.sh

 --- Check if we must enable/disable the netdata updater tool ---
You need to run the installer as root for auto-updating via cron
 FAILED  Cannot disable netdata updater tool

 --- Wrap up environment set up ---
Preparing .environment file
[/Users/prologic/NetData/netdata]$ chmod 0644 /tmp/netdata/netdata/etc/netdata/.environment
 OK

Setting netdata.tarball.checksum to 'new_installation'

 --- We are done! ---

  ^
  |.-.   .-.   .-.   .-.   .-.   .  netdata              .-.   .-.   .-.   .-
  |   '-'   '-'   '-'   '-'   '-'   is installed now!  -'   '-'   '-'   '-'
  +----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+--->

  enjoy real-time performance and health monitoring...

brew ~/NetData/netdata$
```
</details>

##### Additional Information

We need macOS CI :/

PR #9130, opened 2020-05-21
Add CI for our Static Netdata builds (which kickstart-static64 uses)

### LINKED ISSUE #9147: Add CI check for Static builds (kickstart-static64)

prologic:
##### Summary

ssia

- Fixes #9147

##### Component Name

- area/ci

##### Test Plan

- This PR.

##### Additional Information

In working on #9093 I discovered by acciden that something somewhere,
somehow had borkne our ability to build Netdata and statically link it.

~This means our recent kickstart-static64 builds are in very likely no longer
static and will start to cause problems for our users.~ They were in fact broken by #8865

~This PR is to prove that by putting up a CI check that fails if the following
shell expression also fails:~ This PR now ensures we do not break the static builds.

### DIFF
--- a/.github/scripts/build-static-x86_64.sh
+++ b/.github/scripts/build-static-x86_64.sh
@@ -1,0 +1,58 @@
+#!/bin/sh
+#
+# Builds the netdata-vX.Y.Z-xxxx.gz.run (static x86_64) artifact.
+
+set -e
+
+# shellcheck source=.github/scripts/functions.sh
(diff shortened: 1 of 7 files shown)


Ferroin:
OK, fix pushed, root cause was 5675f2f48a causing C/CXX/LD flags variables to not be passed through correctly.

The expression also appears to need to be fixed as well, as musl `ldd` is not giving the expected output for the resultant static binary.

PR #9134, opened 2020-05-22
Merge to make maint on the pfsense documentation easier going forward

wags007:
##### Summary
When trying to install on pfsense several if not most of the packages were coming back 404.  This will be a Maint problem going forward so I add added a script to produce the commands with the current versions.
##### Component Name
pfsense install guide.
##### Test Plan
Not sure what to put in for a test.

##### Additional Information
NA

### DIFF
--- a/packaging/installer/methods/pfsense.md
+++ b/packaging/installer/methods/pfsense.md
@@ -29,10 +29,7 @@ pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/py37-urllib3-1.25.7,1
 pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/py37-yaml-5.2.txz
 pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.20.0_3.txz
 ```
-
-**Note:** If you receive a `Not Found` error during the last two commands above, you will either need to manually look
-in the [repo folder](http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/) for the latest available package and use its
-URL instead, or you can try manually changing the Netdata version in the URL to the latest version.
+**Note:** If you receive a `Not Found` error running any of the above commands [this bash script](./pfsense_getLatestPackages.sh) will update all of the above packages to the current ones in the repo.  Bash,grep,awk and curl are the only requirements.  These should all be pre-installed on Linux, MacOS, and Windows WSL.
 
 **Note:** On pfSense 2.4.5, Python version 3.7 may be installed by the system, in which case you should should not install Python from the FreeBSD repository as instructed above.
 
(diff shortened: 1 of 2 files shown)


packaging/installer/methods/pfsense.md
```diff
@@ -29,10 +29,7 @@ pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/py37-urllib3-1.25.7,1
 pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/py37-yaml-5.2.txz
 pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.20.0_3.txz
 ```
-
-**Note:** If you receive a `Not Found` error during the last two commands above, you will either need to manually look
-in the [repo folder](http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/) for the latest available package and use its
-URL instead, or you can try manually changing the Netdata version in the URL to the latest version.
+**Note:** If you receive a `Not Found` error running any of the above commands [this bash script](./pfsense_getLatestPackages.sh) will update all of the above packages to the current ones in the repo.  Bash,grep,awk and curl are the only requirements.  These should all be pre-installed on Linux, MacOS, and Windows WSL.
```
joelhans, line 32:
```suggestion

> If you receive a `Not Found` error running any of the above commands [this bash script](./pfsense_getLatestPackages.sh) will output updated commands based on the most up-to-date packages. Bash, grep, awk, and curl are the only requirements. These should all be pre-installed on Linux, MacOS, and Windows WSL.
```

joelhans:
Thanks for this, @wags007! Could you sign our CLA so we know we can merge your work?

I'm not sure how closely the team adheres to the shell check, but it popped a few warnings that you might be able to fix.

prologic:
@wags007 Since you seems to be familiar with the subtleties of pfSense; Would you be willing to adapt our `./packaging/installer-install-required-packages.sh` to support pfSense? A new PR that does this would be great; these doco changes can still go ahead after @joelhans is happy.

packaging/installer/methods/pfsense.md
wags007: You are correct in your comments/corrections.  I think the wording is better now.

ilyam8:
That is actually a very nice idea to have such script. Updating documentation manually doesn't work at all (like #9499, #10714).

---

Unfrotunately, the script doesn't work

```cmd
0 ~ $ curl -s http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/
<html>
<head><title>403 Forbidden</title></head>
<body>
<center><h1>403 Forbidden</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

@wags007 are you willing to fix it?

joelhans:
@ilyam8 At this point, I would prefer to move this file, and others, out of the official docs and over to the community forum.

- pfSense
- FreeBSD
- Alpine
- FreeNAS
- Synology

Unless our packaging/SREs want to support them officially, we should really save ourselves the trouble of worrying about how to maintain these procedures that no one can actually test.

PR #9170, opened 2020-05-26
Update streaming protocol

### LINKED ISSUE #8921: Update streaming protocol
* Add a new version of the streaming protocol and version negotiation.
* Change the polling code in the sender / receiver to look for messages in the other direction.
* Add a simple state machine in the protocol start phase where the replication will occur

Note: this will happen on a feature-branch so that we don't leak versions with the new protocol version into the wild otherwise they will break when they connect to completed agents.

amoss:
##### Summary

Closes #8921 
This PR is mainly an enabler for upcoming patches on filling the chart gaps in the streaming protocol., but it also fixes some bugs in the streaming component:
* Incorrect handling of return codes
* Uses circular buffer for transmission to avoid starvation on high-latency link
* Remove the warnings from the time defines that were shown for every file on the mac.
* Restructured the sender.
* Made the locking around the sender buffer consistent.
* Fix a case where the stream version was not negotiated correctly.
* Adds bi-directional comms on the streaming and a place-holder command processor in the sender code for later.

##### Component Name
streaming

##### Test Plan

* Tested configurations of masters / slaves / proxies in docker containers.
* Tested bare-metal master-slave over WAN connection
* Checked the hi-throughput and a low-thoughput scenario
* Cable pulls, gdb pause/resume.
* Coverity scan
* Valgrind

##### Additional Information

Note for reviewers: no need to check the files under `build_external/`, they are only to set up testing scenarios in docker.

Another note: the concurrency has been improved but it is not quite right yet - I will change the thread cancellation / shutdown of the sender to make it more robust across the following PRs.


amoss:
Coverity has detected a bug that currently exists in master but cannot be triggered. Moving around the code in the `rrdpush` file has allowed the analyser to detect the problem. This will probably generate a Coverity report when this is merged, but the bug cannot be triggered in the agent currently and I will change the locking / thread spawning in the next PR to fix it.

build_external/scenarios/gaps_hi/master_guid
```diff
@@ -0,0 +1 @@
+00000000-0000-0000-0000-000000000000
```
thiagoftsm, line 1: Please, add a blank line here.

build_external/scenarios/gaps_hi/middle_guid
```diff
@@ -0,0 +1 @@
+11111111-1111-1111-1111-111111111111
```
thiagoftsm, line 1: Please, add a newline here.

build_external/scenarios/gaps_hi/slave_guid
```diff
@@ -0,0 +1 @@
+22222222-2222-2222-2222-222222222222
```
thiagoftsm, line 1: please add a new line here.

build_external/scenarios/gaps_lo/master_guid
```diff
@@ -0,0 +1 @@
+00000000-0000-0000-0000-000000000000
```
thiagoftsm, line 1: This is another file without a blank line.

build_external/scenarios/gaps_lo/middle_guid
```diff
@@ -0,0 +1 @@
+11111111-1111-1111-1111-111111111111
```
thiagoftsm, line 1: All the guid files need a new line at the end.

streaming/sender.c
```diff
@@ -0,0 +1,697 @@
...
+
+    host->rrdpush_sender_spawn = 0;
+
+    info("STREAM %s [send]: sending thread now exits.", host->hostname);
+
+    rrdhost_unlock(host);
+}
+
+void sender_init(struct sender_state *s, RRDHOST *parent) {
+    memset(s, 0, sizeof(*s));
```
thiagoftsm, line 553: A suggestion, instead to use `mallocz`, you could use `callocz` and remove this `memset`.

build_external/scenarios/gaps_hi/master_guid
amoss: Why should the GUID files have newlines? (They work fine without it, is there some specific reason)

streaming/sender.c
amoss: This is true, although as it stands this procedure guarantees the initialized state of the structure without any preconditions.

build_external/scenarios/gaps_hi/master_guid
thiagoftsm: The Github was informing the missing line, there is not a specific reason for this, probably the unique reason is to keep pattern with the other source code files.

thiagoftsm:
While I was executing tests there was a test that called my attention, I decided to test the slave using `ram` as `memory mode`. In this situation it is not possible to fill the gaps, but we are not giving any information for our users about this. Should we at least give an additional information for our users on situations like this?

thiagoftsm:
I used the following scenario and I detected a possible bug:

1 - I used bare metal to test
2 - The slave is 10 years old laptop with a wireless 100 megabits
3 - The master is newer, but it was also using wireless.
4 - I started the slave and I let it running few minutes before to start master
5 - I started the master

And I have these charts for the slave:

err67

I observed that in this scenario the first data that I have stored for the master happened at 1:40:24 while the first data for slave was stored 1:40:16, but the difference of time between the Slave start and the master start was 11 minutes.

Another problem visible on this image , it is the fact that the gaps are present.

amoss:
This looks like expected behaviour:
* We store data on the master using the master's clock to decide when the data was received.
* The gap-filling is not implemented yet.

This PR is an enabler that changes the structure of the sender:
* The buffer will no longer starve on a high-latency / high-bandwidth connection with bursty behaviour.

amoss:
@thiagoftsm I've finished the next PR in the sequence. It currently includes all the commits from this PR. It might be easier to close this PR and just continue testing the new one. It has these changes, plus some new (inactive) code to calculate gaps sizes and clock-slew. The important thing about the new one is that it fixes the bug where slaves can't reconnect. I'll add you as a reviewer on the new one so that you get a notification.

thiagoftsm:
@amoss I can approve it right now, I was running only a long test to confirm that I did not miss nothing during the review, do what you find better.

amoss:
If it is stable I'm happy to merge it then rebase the branch for the other PR, or just close it and use the other one.... @stelfrag still needs to review it, and I need @mfundul to take a look as I rewrote the changes he made for the spawn server. Maybe it is easier to close it in favour of https://github.com/netdata/netdata/pull/9214?

thiagoftsm:
It is very stable :) , I am 5 hours without any problems, I will continue this test in the next 3 hours to complete 8 hours test.
For me we can close and I do the necessary tests with the other PR.

amoss:
I've updated the comments on the new PR. They are becoming... wordy :). Closing in favour of https://github.com/netdata/netdata/pull/9214

Issue #8174, opened 2020-02-23
Docker Healthcheck support


ElleshaHackett:
**Describe the feature**
Being able to see a "health status" of the Netdata Docker container.

**Describe the solution you'd like**
I would like support for the Docker Healthcheck (that is also shown in Portainer.io and probably other Docker management software).

**Additional context**
The `Dockerfile` could contain something like this:

`HEALTHCHECK --interval=60s --timeout=10s --retries=3 CMD curl -sS http://localhost:19999 || exit 1`. 

For debugging and testing purposses you can use:

`docker inspect --format "{{json .State.Health}}" containername`

![image](https://user-images.githubusercontent.com/5116641/75119602-a315ad00-5684-11ea-820f-59508f47fb1d.png)

PR #9172, opened 2020-05-26
Added health check functionality to our Docker images.

Ferroin:
##### Summary

This PR adds a script to our Docker images to provide health checks for the Docker runtime to confirm that the container is operational. These checks are visible to external management tools that use Docker, and also allow for automated restarts of containers that get into a bad state.

The script itself queries the `/api/v1/info` endpoint and looks for a 200 status code on the returned response. It can be easily configured to use a different endpoint instead through a new environment variable (`NETDATA_HEALTHCHECK_TARGET`).

This PR also adds a `ping` command to `netdatacli` to allow querying if the daemon appears to be alive and responsive without hitting the REST API or relying on `netdatacli help` correctly returning if the daemon is up and failing if it's down. The health check can switched to utilize this instead of an API call by setting `NETDATA_HEALTHCHECK_TARGET=cli` for the Docker container. This alternative health check is intended for use by people who wish to impose strict ACL's on the Netdata API for security reasons.

##### Component Name

area/packaging
area/daemon

##### Test Plan

Verified locally that:

* The new `netdatacli` command works correctly.
* The default health check behavior works correctly.
* The `cli` health check variant works correctly.
* Specifying an arbitrary endpoint for the health check works correctly.

##### Additional Information

Fixes: #8174 

Long-term it would be nice if `netdatacli ping` could be extended to actually check the daemon state and verify it's not hung, but that's far beyond my abilities when it comes to C, and honestly a bit out of scope for this PR.


packaging/docker/README.md
```diff
@@ -98,6 +98,27 @@ volumes:
 
 Run `docker-compose up -d` in the same directory as the `docker-compose.yml` file to start the container.
 
+## Health Checks
+
+Our Docker image provides integrated support for health checks through the standard Docker interfaces.
+
+You can control how the helath checks run by using the environment variable `NETDATA_HEALTH_CHECK` as follows:
```
prologic, line 105:
```suggestion
You can control how the health checks run by using the environment variable `NETDATA_HEALTH_CHECK` as follows:
```

packaging/docker/Dockerfile
```diff
@@ -88,6 +89,7 @@ RUN \
         /var/lib/netdata \
         /var/log/netdata && \
     chown -R netdata:netdata /var/lib/netdata/cloud.d && \
+    chmod 0755 /health.sh && \
```
prologic, line 92: Please just chmod and commit the file with executable bits set.

```#!sh
chmod +x packaging/docker/health.sh
git add packaging/docker/health.sh
git ci -m "..."
```

packaging/docker/health.sh
```diff
@@ -0,0 +1,17 @@
...
+    # If users didn't request something else, query `/api/v1/info`.
+    NETDATA_HEALTH_CHECK="http://localhost:19999/api/v1/info"
+fi
+
+case "${NETDATA_HEALTH_CHECK}" in
+    cli)
+        netdatacli ping || exit 1
+        ;;
+    *)
+        curl -sS "${NETDATA_HEALTH_CHECK}" || exit 1
```
prologic, line 15: Please use `-q` as the first argument. And add `-L` in case of redirects.

It would be better to also rename `NETDATA_HEALTH_CHECK` to `NETDATA_HEALTH_CHECK_URL` to be more clear.

Ferroin: Not using `-q` was intentional here. Docker stores the health check output and you can inspect it (`docker inspect --format "{{json .State.Health }}" <container name> | jq`), which is useful for figuring out why it's failing in a case like this.

Agreed on the `-L` though.

Ferroin:
I assume @amoss has already commented on this, but this should not have been merged without approvals from @netdata/core regarding the C code.

amoss:
This is the PR that I was referring to on slack. I've read the code and there is no problem in this particular case - but everybody on the team needs to sticks to the same process. This is because it protects us in cases where we do not see the danger in a particular patch. The division that I would make is not between the agent team and the SRE team - it is not too helpful and you are trusted in the repo to the same extent that we are: - we are all on the same team. The important aspect is that a person from the relevant AoR needs to have approved the code before it is merged. This is something that I hold the members of the agent team to as well.

mfundul:
@Ferroin you know that `"help"` did the exact same thing but you just couldn't help yourself :D

Ferroin:
> @Ferroin you know that `"help"` did the exact same thing but you just couldn't help yourself :D

I seriously hope that we eventually get to the point that `netdata` does not need to be running for us to run `netdatacli help`.

mfundul:
> 
> 
> > @Ferroin you know that `"help"` did the exact same thing but you just couldn't help yourself :D
> 
> I seriously hope that we eventually get to the point that `netdata` does not need to be running for us to run `netdatacli help`.

We can do that easily, just that then we must be absolutely sure that `netdatacli` and `netdata` are always synchronized. Otherwise, it's just a print.

Ferroin:
> > > @Ferroin you know that `"help"` did the exact same thing but you just couldn't help yourself :D
> > 
> > 
> > I seriously hope that we eventually get to the point that `netdata` does not need to be running for us to run `netdatacli help`.
> 
> We can do that easily, just that then we must be absolutely sure that `netdatacli` and `netdata` are always synchronized. Otherwise, it's just a print.

Even if they're out of sync, it's still very likely to be more useful to the user than what currently gets spit out. We have _no_ documentation for this command other than it's help output and a docs page that doesn't appear to actually be linked into our docs, which itself just parrots the help output.

prologic:
@amoss Sorry about that :D Won't happen again! 🤞

PR #9165, opened 2020-05-26
Fix reliability of kickstart/kickstart-static64 with checksums sometimes failing

prologic:
##### Summary

ssia

- Fixes #8967

##### Component Name

- area/packaging

##### Test Plan

For end-users:

1. Run the following on your system of choice:

```#!sh
NETDATA_TARBALL_BASEURL=https://storage.googleapis.com/netdata-dev-nightlies bash <(curl -q -sSL https://storage.googleapis.com/netdata-dev-nightlies/kickstart.sh)
```

2. Ensure you are on the right version `netdata -v` and that things are
   working okay for you.
3. Confirm/Report that you don't see checksum failures inconsistently
   anymore.

Replace `kickstart.sh` with `kickstart-static64.sh` to test the x86_64
static Netdata agent or if you cannot build from source.

----

For Contributors/Maintainers:

1. Run: `git clean -d -f -x`
2. Run: `.github/scripts/build-artifacts.sh`

If you are on macOS it will fail the first time with:

```#!sh
$ .github/scripts/build-artifacts.sh
Preparing build ...                     [ OK ]
Building dist ...                       [ OK ]
Building static x86_64 ...              [ ERR ]
.
.
.
[/usr/src/netdata.git/packaging/makeself/tmp]# tar -zxpf bash-4.4.18.tar.gz
tar: bash-4.4.18/ChangeLog: Cannot utime: No such file or directory
tar: Exiting with failure status due to previous errors
Build failed.Dropping into a shell for debugging ...
sh: _kube_ps1_update_cache: command not found
sh: _direnv_hook: command not found
sh-3.2$
```

This is expected becuase of the very old way we do builds with
bind-mounted volumes of the source into a Docker container instead of a
proper `Dockerfile` with `COPY` instructions and optionally
bind-mounting sources iif necessary.

If this happens for you on macOS just re-run Step 2)

3. Copy the `kickstart.sh`  / `kickstart-static64.sh` from this
   branch/PR into `./artifacts`:

```#!sh
cp packaging/installer/kickstart* artifacts/
```

4. Serve up `./artifacts` using your favorite web server. I use
   [static](https://github.com/prologic/static):

```#!sh
static -r ./artifacts
```

5. Install Netdata via a custom `NETDATA_TARBALL_BASEURL`:

```#!sh
NETDATA_TARBALL_BASEURL=http://SERVER_IP:PORT bash <(curl -q -sSL http://SERVER_IP:PORT/kickstart.sh)
```

[...]

PR #9174, opened 2020-05-26
Really prevent overwriting netdata.conf on static installs.

Ferroin:
##### Summary

The previous fix was incompletely tested and didn't actually workdue to
using an incorrect variable name. This fixes that.

##### Component Name

area/packaging

##### Test Plan

Original testing only looked at intermediate stages, this one is now confirmed by testing the final output.

##### Additional Information

Fixes: #8531

### DIFF
--- a/packaging/makeself/jobs/70-netdata-git.install.sh
+++ b/packaging/makeself/jobs/70-netdata-git.install.sh
@@ -23,7 +23,7 @@ run ./netdata-installer.sh --install "${NETDATA_INSTALL_PARENT}" \
   "${NULL}"
 
 # Remove the netdata.conf file from the tree. It has hard-coded sensible defaults builtin.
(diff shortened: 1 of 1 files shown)


prologic:
If there are two ticks from @knatsakis it must be good :) Approved!

Ferroin:
@prologic You're right this does indeed not work on top of PR #9165, though I'm not sure why. However, testing just by itself, it does indeed appear to do as advertised and not overwrite `/opt/netdata/etc/netdata.conf`.

prologic:
> @prologic You're right this does indeed not work on top of PR #9165, though I'm not sure why. However, testing just by itself, it does indeed appear to do as advertised and not overwrite `/opt/netdata/etc/netdata.conf`.

Let's discuss in tonight sync or we can chat anytime you are available.

prologic:
Based on in-person discussion I'm approving this as I cannot see any other place where this wwould be misbehaving.

PR #9192, opened 2020-05-27
Add support for collectors updating their metadata

### LINKED ISSUE #8924: Add support for collectors updating their metadata
Add support for collectors updating their metadata.

When a collector begins it will update those structures (rrdset/rrddim) with more up-to-date information.

##### Cases

- `CHART name title family context charttype priority update_every options plugin module`
What cannot be updated for CHART is:
    - `type.id units`

- `DIMENSION name algorithm multiplier divisor options`
What cannot be updated for DIMENSION is:
   - `id`

- `HOST hostname registry_hostname update_every os timezone tags system_info`
What cannot be updated for HOST is:
   - `GUID`

##### Steps

- Atomically update the in-memory data structures.
- Write the updates to the metadata log as new records.

stelfrag:
Fixes #8924 
##### Summary
Expand the attributes of a chart / dimension / Host that can be dynamically updated

- Handle changes in dimensions (already there)
  - Queue a metadata log update by calling `metalog_commit_update_dimension`
- Perform checks on chart updates
  - Queue a metadata log update by calling  `metalog_commit_update_chart`

##### Test Plan
- TBA

### DIFF
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -211,6 +211,14 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
             rrddim_flag_clear(rd, RRDDIM_FLAG_ARCHIVED);
         }
 
+        int rc = rrddim_set_name(st, rd, name);
+        rc += rrddim_set_algorithm(st, rd, algorithm);
+        rc += rrddim_set_multiplier(st, rd, multiplier);
+        rc += rrddim_set_divisor(st, rd, divisor);
+        if (likely(!rc))
+            info("Dimension %s  (name, algorithm, multiplier and divisor) unchanged", rd->id);
+        else
+            rrdset_flag_set(st, RRDSET_FLAG_UPDATE_METADATA);
         rrdset_unlock(st);
         return rd;
     }
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -219,7 +219,6 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
         if(memory_mode == RRD_MEMORY_MODE_DBENGINE) {
 #ifdef ENABLE_DBENGINE
             if (unlikely(rc)) {
-                rrdset_flag_set(st, RRDSET_FLAG_UPDATE_METADATA);
                 metalog_commit_update_dimension(rd);
             }
 #endif
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -215,10 +215,15 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
         rc += rrddim_set_algorithm(st, rd, algorithm);
         rc += rrddim_set_multiplier(st, rd, multiplier);
         rc += rrddim_set_divisor(st, rd, divisor);
-        if (likely(!rc))
-            info("Dimension %s  (name, algorithm, multiplier and divisor) unchanged", rd->id);
-        else
-            rrdset_flag_set(st, RRDSET_FLAG_UPDATE_METADATA);
+        // DBENGINE available and activated?
+        if(memory_mode == RRD_MEMORY_MODE_DBENGINE) {
+#ifdef ENABLE_DBENGINE
+            if (unlikely(rc))
+                rrdset_flag_set(st, RRDSET_FLAG_UPDATE_METADATA);
+            else
+                info("Dimension %s (name, algorithm, multiplier and divisor) unchanged", rd->id);
+#endif
+        }
         rrdset_unlock(st);
         return rd;
     }
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -218,6 +218,7 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
         // DBENGINE available and activated?
 #ifdef ENABLE_DBENGINE
         if (likely(rd->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) && unlikely(rc)) {
+            debug(D_METADATALOG, "DIMENSION [%s] metadata updated", rd->id);
             metalog_commit_update_dimension(rd);
         }
 #endif
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -216,13 +216,11 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
         rc += rrddim_set_multiplier(st, rd, multiplier);
         rc += rrddim_set_divisor(st, rd, divisor);
         // DBENGINE available and activated?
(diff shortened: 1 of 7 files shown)


database/rrdhost.c
```diff
@@ -447,6 +456,11 @@ RRDHOST *rrdhost_find_or_create(
 
         // update host tags
         rrdhost_init_tags(host, tags);
+#ifdef ENABLE_DBENGINE
+        if (host->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
+            metalog_commit_update_host(host);
+        }
+#endif
```
mfundul, line 463: Can you add this in `rrdhost_create()` as well? We're missing a `HOST` command when streaming connects and it's a good opportunity to fix it here before we merge.

database/rrddim.c
```diff
@@ -202,15 +202,21 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
...
             rrddim_flag_clear(rd, RRDDIM_FLAG_ARCHIVED);
         }
-
+        // DBENGINE available and activated?
+#ifdef ENABLE_DBENGINE
+        if (likely(!is_archived && rd->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) && unlikely(rc)) {
+            debug(D_METADATALOG, "DIMENSION [%s] metadata updated", rd->id);
+            metalog_commit_update_dimension(rd);
+        }
+#endif
```
mfundul, line 219: Should we do ACLK updates as well?

database/rrdset.c
```diff
@@ -519,19 +519,114 @@ RRDSET *rrdset_create_custom(
...
+            }
+
+            rrdhost_unlock(host);
+
+#ifdef ENABLE_DBENGINE
+            if (is_archived == 0 && st->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE && mark_rebuild) {
+                debug(D_METADATALOG, "CHART [%s] metadata updated", st->id);
+                metalog_commit_update_chart(st);
+            }
+#endif
```
mfundul, line 626: Should we do ACLK updates as well?

database/rrdhost.c
stelfrag: Ok I will check it

This actually has an issue with streaming because it will do a `rrdhost_find_by_guid` and never actually try to update the info (via `rrdhost_find_or_create`) I am tweaking it to do so and make sure we will have all the info to write the metadata log entries (eg. labels)

database/rrdset.c
stelfrag: There are a few checks to be done for this to work properly. Since a plugin / module can also be updated, that may delete and add a new collector. 

ACLK keeps track of the active collectors by counting their chart and  decides if a chart update should go out or a full `on-connect` payload.

stelfrag: ACLK updates should not go out during metadata log replay

mfundul: `if (is_archived == 0)` means it's not during log replay though. Replay sets `is_archived = 1`.

stelfrag: I made a couple of changes to handle switch from archived to active (which should also update the collector list within ACLK)

mfundul:
You should really add some `#define`s or something with names, because all those `1`,`2`,`4`,`6`,`7`,`8` are very hard to read.

database/rrdset.c
```diff
@@ -525,108 +525,111 @@ RRDSET *rrdset_create_custom(
...
-            } else {
-                if (plugin != st->plugin_name) { // one is NULL?
-                    if (st->plugin_name)
-                        freez(st->plugin_name);
-                    st->plugin_name = plugin ? strdupz(plugin) : NULL;
-                    mark_rebuild = 1;
-                }
+        } else {
+            if (plugin != st->plugin_name) { // one is NULL?
+                if (st->plugin_name)
```
amoss, line 555: The `if` is redundant - if the old name is NULL it will not be freed.

database/rrdset.c
```diff
@@ -525,108 +525,111 @@ RRDSET *rrdset_create_custom(
...
-            } else {
-                if (module != st->module_name) {
-                    if (st->module_name)
-                        freez(st->module_name);
-                    st->module_name = module ? strdupz(module) : NULL;
-                    mark_rebuild = 1;
-                }
+        if (module && st->module_name) {
+            if (unlikely(strcmp(module, st->module_name))) {
+                old_module = st->module_name;
```
amoss, line 564: This is correct - the other cases have the redundant `if`.

database/rrdset.c
```diff
@@ -525,108 +525,111 @@ RRDSET *rrdset_create_custom(
...
+            if (unlikely((strcmp(st->context, new_context)))) {
+                old_context = st->context;
+                st->context = new_context;
+                st->hash_context = simple_hash(st->context);
+                mark_rebuild = 1;
+            } else
+                freez(new_context);
+        }
 
+        if (mark_rebuild) {
```
amoss, line 621: `if (mark_rebuild && (old_plugin || old_module ...))`

amoss: We don't want the pause if the collector sent the CHART again but nothing changed.

amoss:
> You should really add some `#define`s or something with names, because all those `1`,`2`,`4`,`6`,`7`,`8` are very hard to read.

Agree - make the `6` and `7` explicit `|` of different flags. They look like typos at the moment.

database/rrdset.c
```diff
@@ -519,18 +520,138 @@ RRDSET *rrdset_create_custom(
...
+        }
+
+        RRDSET_TYPE new_chart_type =
+            rrdset_type_id(config_get(st->config_section, "chart type", rrdset_type_name(chart_type)));
+        if (st->chart_type != new_chart_type) {
+            st->chart_type = new_chart_type;
+            mark_rebuild |= META_CHART_UPDATED;
+        }
+
+        char *new_family = family ? strdup(family): NULL;
```
mfundul, line 596: This is the fast path and we shouldn't `strdup()` unconditionally, right? Also we should use `strdupz()`.

database/rrdset.c
```diff
@@ -519,18 +520,138 @@ RRDSET *rrdset_create_custom(
...
+                old_family = st->family;
+                rrdfamily_free(host, st->rrdfamily);
+                st->family = new_family;
+                st->rrdfamily = rrdfamily_create(host, st->family);
+                mark_rebuild |= META_CHART_UPDATED;
+            } else
+                freez(new_family);
+        }
+
+        char *new_context = context ? strdup(context) : NULL;
```
mfundul, line 609: Same as above.

amoss: I think that it needs a `strdupz` here as the `json_fix_string` rewrites the input.

mfundul:
I don't know if it is worth it. We are duplicating memory usage for those 3 fields for every chart in the system, to support changing thing such as the family which will probably break netdata.

PR #9195, opened 2020-05-27
Initial version of metadata log parsing and replay.

### LINKED ISSUE #8903: Plugins.D language parsing in the metadata log and log replay
Open the metadata logs and integrate the incremental plugins.D parser to replay the log as defined [here](https://github.com/netdata/docs/blob/master/architecture/agent/db/database-log-format-pros-cons.md).

- load the metadata log file into memory.
- for each record in the log call the parser with the command.
- create the corresponding objects (rrdhost/rrdset/rrddim) and initialize them as inactive.
   - sane defaults for any collector state (e.g. last_collected)
   - export all the charts regardless of their state

Refactor rrdhost/rrdset/rrddim as needed to allow for partial initialization. There are going to be 2 states for all these objects:
1. inactive - the object has history but is not being collected
2. active - a collector has explicitly reinitialized the object

mfundul:
##### Summary
Open the metadata logs and integrate the incremental plugins.D parser to replay the log.
Fixes #8903
##### Component Name
daemon
database
aclk
health
collectors
exporting
web
##### Test Plan

The new endpoint to get archived charts is `/api/v1/archivedcharts`
To get the data of an archived chart the only way is via endpoint `/api/v1/data?chart=name`.


collectors/diskspace.plugin/plugin_diskspace.c
```diff
@@ -254,7 +254,7 @@ static inline void do_disk_space_stats(struct mountinfo *mi, int update_every) {
                                               netdata_zero_metrics_enabled == CONFIG_BOOLEAN_YES))) {
         if(unlikely(!m->st_space)) {
             m->do_space = CONFIG_BOOLEAN_YES;
-            m->st_space = rrdset_find_bytype_localhost("disk_space", disk);
+            m->st_space = rrdset_find_active_bytype_localhost("disk_space", disk);
```
stelfrag, line 257: There are other places where `rrdset_find_bytype_localhost` is used. Shouldn't be updated there as well?

mfundul: But of course!

database/engine/metadata_log/logfile.c
```diff
@@ -407,6 +597,15 @@ static int scan_metalog_files(struct metalog_instance *ctx)
         metadata_logfile_list_insert(ctx, metalogfile);
         ctx->disk_space += metalogfile->pos + metalogfile->pos;
```
stelfrag, line 598: why `metalogfile->pos` twice?

mfundul: Twice as fun!

collectors/diskspace.plugin/plugin_diskspace.c
mfundul: It's possible my brain deleted MacOS for known reasons.

stelfrag:
We probably need to exclude archived charts from updating the ACLK during metadata replay  in 
```
#ifdef ENABLE_ACLK
    if (netdata_cloud_setting) {
        aclk_add_collector(host->hostname, plugin, module);
        aclk_update_chart(host, st->id, ACLK_CMD_CHART);
    }
#endif
```
The collector will be recorded as active by counting the number of charts it has created which is not the case (the actual /api/v1/info) will not return the collector in the list as it will be filtered by `rrdset_is_available_for_viewers`

The `aclk_update_chart` will be a no-op because
- The agent is set to `INITIALIZING` 
- The `aclk_update_chart` has an additional check to only allow updates from localhost
   - Note that the `localhost` has not been initialized during metadata log replay so the `host != localhost` check will always be true

The `aclk_add_collector` should be called when a chart changes from archived to non-archived (collection has started)

The `aclk_del_collector` should be called when a chart changes to archived

web/api/web_api_v1.c
```diff
@@ -915,6 +924,7 @@ static struct api_command {
         { "data",            0, WEB_CLIENT_ACL_DASHBOARD, web_client_api_request_v1_data            },
         { "chart",           0, WEB_CLIENT_ACL_DASHBOARD, web_client_api_request_v1_chart           },
         { "charts",          0, WEB_CLIENT_ACL_DASHBOARD, web_client_api_request_v1_charts          },
+        { "archivedcharts",  0, WEB_CLIENT_ACL_DASHBOARD, web_client_api_request_v1_archivedcharts  },
```
amoss, line 927: What will this be used for, and how will it be used?

database/rrd.h
```diff
@@ -919,6 +928,14 @@ extern RRDSET *rrdset_find(RRDHOST *host, const char *id);
 
 extern RRDSET *rrdset_find_bytype(RRDHOST *host, const char *type, const char *id);
 #define rrdset_find_bytype_localhost(type, id) rrdset_find_bytype(localhost, type, id)
+/* This will not return charts that are archived */
```
amoss, line 931: The original `rrdset_find_bytype` in `rrset.c` should be removed instead of left as dead code.

I there a reason to move this to a `static inline` definition in a header file instead of an `inline` definition in the `.c` file like the original?

collectors/plugins.d/plugins_d.h
```diff
@@ -31,10 +31,10 @@
 #define PLUGINSD_KEYWORD_VARIABLE "VARIABLE"
 #define PLUGINSD_KEYWORD_LABEL "LABEL"
 #define PLUGINSD_KEYWORD_OVERWRITE "OVERWRITE"
-#define PLUGINSD_KEYWORD_CONTEXT "CONTEXT"
 #define PLUGINSD_KEYWORD_GUID "GUID"
-#define PLUGINSD_KEYWORD_HOST "HOST"
+#define PLUGINSD_KEYWORD_CONTEXT "CONTEXT"
```
amoss, line 35: This is a strange change as the list seems to have the defines, just in a different order. Keeping definitions in a sorted order prevents this kind of noise in the commit.

collectors/plugins.d/pluginsd_parser.c
```diff
@@ -560,6 +560,71 @@ PARSER_RC pluginsd_overwrite(char **words, void *user, PLUGINSD_ACTION  *plugins
...
+    if (unlikely(!uuid_str)) {
+        error("requested a GUID, without a uuid.");
+        return PARSER_RC_ERROR;
+    }
+    if (unlikely(strlen(uuid_str) != GUID_LEN || uuid_parse(uuid_str, uuid) == -1)) {
+        error("requested a GUID, without a valid uuid string.");
+        return PARSER_RC_ERROR;
+    }
+
+    info("Parsed uuid=%s", uuid_str);
```
amoss, line 577: Change to `debug` or remove.

collectors/plugins.d/pluginsd_parser.c
```diff
@@ -560,6 +560,71 @@ PARSER_RC pluginsd_overwrite(char **words, void *user, PLUGINSD_ACTION  *plugins
...
+    if (unlikely(!uuid_str)) {
+        error("requested a CONTEXT, without a uuid.");
+        return PARSER_RC_ERROR;
+    }
+    if (unlikely(strlen(uuid_str) != GUID_LEN || uuid_parse(uuid_str, uuid) == -1)) {
+        error("requested a CONTEXT, without a valid uuid string.");
+        return PARSER_RC_ERROR;
+    }
+
+    info("Parsed uuid=%s", uuid_str);
```
amoss, line 599: Change to `debug` or remove.

collectors/plugins.d/pluginsd_parser.c
```diff
@@ -560,6 +560,71 @@ PARSER_RC pluginsd_overwrite(char **words, void *user, PLUGINSD_ACTION  *plugins
...
+    if (unlikely(!uuid_str)) {
+        error("requested a TOMBSTONE, without a uuid.");
+        return PARSER_RC_ERROR;
+    }
+    if (unlikely(strlen(uuid_str) != GUID_LEN || uuid_parse(uuid_str, uuid) == -1)) {
+        error("requested a TOMBSTONE, without a valid uuid string.");
+        return PARSER_RC_ERROR;
+    }
+
+    info("Parsed uuid=%s", uuid_str);
```
amoss, line 621: Change to `debug` or remove.

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -221,16 +235,108 @@ void metalog_commit_delete_dimension(RRDDIM *rd)
...
+    return st;
+}
+
+RRDDIM *metalog_get_dimension_from_uuid(struct metalog_instance *ctx, uuid_t *metric_uuid)
+{
+    GUID_TYPE ret;
+    char dim_object[49], chart_object[33], id_str[1024], chart_fullid[RRD_ID_LENGTH_MAX + 1];
+    uuid_t *machine_guid, *chart_guid, *chart_char_guid, *dim_char_guid;
+
+    ret = find_object_by_guid(metric_uuid, dim_object, 49);
```
amoss, line 276: ret = find_object_by_guid(metric_uuid, dim_object, sizeof(dim_object));

collectors/plugins.d/pluginsd_parser.c
mfundul: Will do, but I need them in info for a little while to help with debugging. Same deal as the ACLK prints that remained `info()`. I will change this to `debug` as soon as I'm done testing compaction.

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -221,16 +235,108 @@ void metalog_commit_delete_dimension(RRDDIM *rd)
...
+    RRDHOST *host = ctx->rrdeng_ctx->host;
+    assert(!uuid_compare(host->host_uuid, *machine_guid));
+
+    chart_guid = (uuid_t *)(dim_object + 16);
+    dim_char_guid = (uuid_t *)(dim_object + 16 + 16);
+
+    ret = find_object_by_guid(dim_char_guid, id_str, 1024);
+    assert(GUID_TYPE_CHAR == ret);
+
+    ret = find_object_by_guid(chart_guid, chart_object, 33);
```
amoss, line 290: sizeof

mfundul: Yeah, we've got A LOT of magic numbers in this branch for some reason, I'm afraid this is not the only occurrence.

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -221,16 +235,108 @@ void metalog_commit_delete_dimension(RRDDIM *rd)
...
+        return NULL;
+
+    machine_guid = (uuid_t *)dim_object;
+    RRDHOST *host = ctx->rrdeng_ctx->host;
+    assert(!uuid_compare(host->host_uuid, *machine_guid));
+
+    chart_guid = (uuid_t *)(dim_object + 16);
+    dim_char_guid = (uuid_t *)(dim_object + 16 + 16);
+
+    ret = find_object_by_guid(dim_char_guid, id_str, 1024);
```
amoss, line 287: Use a sizeof.

The other two magic constants makes sense as multiples of 16 + terminators - where does this constant come from?

mfundul: sizeof(id_str)

web/api/web_api_v1.c
mfundul: The new endpoint to get archived charts is `/api/v1/archivedcharts`. You can query `127.0.0.1:19999/api/v1/archivedcharts` and get the list of charts that are archived. This is different than `/api/v1/charts` in that it only shows charts that no collector has touched yet. Those charts simply did not exist in the previous versions of netdata.

You can actually get the data of those charts by using the `/api/v1/chart?chart=name` endpoint.

Those 2 endpoints are the only way this increment can deliver value to the users.

amoss: I don't think that end users will use the API. If I restart netdata can I see charts via the dashboard if they are archvied? Will they be visible to the cloud?

database/engine/metadata_log/metadatalogapi.c
amoss: Why is id_str 1024 bytes?

web/api/web_api_v1.c
mfundul: The requirements stated that the only way to get the archived charts would be via a new API call. So they should not be visible in the dashboard. About ACLK I'm not sure what to do, it's related to @stelfrag 's comment. We should discuss this.

parser/parser.c
```diff
@@ -63,6 +63,9 @@ PARSER *parser_init(RRDHOST *host, void *user, void *input, PARSER_INPUT_TYPE fl
         rc += parser_add_keyword(parser, PLUGINSD_KEYWORD_END, pluginsd_end);
         rc += parser_add_keyword(parser, PLUGINSD_KEYWORD_BEGIN, pluginsd_begin);
         rc += parser_add_keyword(parser, "SET", pluginsd_set);
+        rc += parser_add_keyword(parser, PLUGINSD_KEYWORD_GUID, pluginsd_guid);
```
amoss, line 66: We don't need to expose these to collectors, you can call this in logfile.c:scan_metalog_file() after the call to `parser_init`.

database/engine/metadata_log/metadatalogapi.c
mfundul: It's supposedly a sensible maximum length for a dimension ID. Maybe it should be related to plugins.d LINE_MAX or something, netdata doesn't really set an upper limit here.

parser/parser.c
mfundul: That was my original plan but Kosta was explicit that we should have one plugins.D language, so this one is a keyword of the language. Ideally I'd have a more optimized language for the metadata log.

Currently, the action is NULL for the plugins so it will not be parsed.

stelfrag: This can be reused if we add GUID to the streaming protocol

stelfrag:
I get 
```
netdata INFO  : MAIN : Deleting data file "/home/stelios/netdata-inst//netdata/var/cache/netdata/dbengine/datafile-1-0000000154.ndf".
netdata: database/engine/metadata_log/metadatalogapi.c:319: metalog_delete_dimension_by_uuid: Assertion `rd' failed.

```
(I lowered the dbengine disk space to force the delete)

mfundul:
> 
> 
> I get
> 
> ```
> netdata INFO  : MAIN : Deleting data file "/home/stelios/netdata-inst//netdata/var/cache/netdata/dbengine/datafile-1-0000000154.ndf".
> netdata: database/engine/metadata_log/metadatalogapi.c:319: metalog_delete_dimension_by_uuid: Assertion `rd' failed.
> ```
> 
> (I lowered the dbengine disk space to force the delete)

Do you know if this is related to legacy UUIDs or if this is a random-only UUID database?

stelfrag:
> > I get
> > ```
> > netdata INFO  : MAIN : Deleting data file "/home/stelios/netdata-inst//netdata/var/cache/netdata/dbengine/datafile-1-0000000154.ndf".
> > netdata: database/engine/metadata_log/metadatalogapi.c:319: metalog_delete_dimension_by_uuid: Assertion `rd' failed.
> > ```
> > 
> > 
> > (I lowered the dbengine disk space to force the delete)
> 
> Do you know if this is related to legacy UUIDs or if this is a random-only UUID database?

This is an old file

`20373504 May 19 11:15 datafile-1-0000000154.ndf`

mfundul:
> 
> 
> > > I get
> > > ```
> > > netdata INFO  : MAIN : Deleting data file "/home/stelios/netdata-inst//netdata/var/cache/netdata/dbengine/datafile-1-0000000154.ndf".
> > > netdata: database/engine/metadata_log/metadatalogapi.c:319: metalog_delete_dimension_by_uuid: Assertion `rd' failed.
> > > ```
> > > 
> > > 
> > > (I lowered the dbengine disk space to force the delete)
> > 
> > 
> > Do you know if this is related to legacy UUIDs or if this is a random-only UUID database?
> 
> This is an old file
> 
> `20373504 May 19 11:15 datafile-1-0000000154.ndf`

Hmm, yes, I should remove the assertion completely. It's possible that there are old data that don't belong to any chart.

mfundul:
> 
> 
> I get
> 
> ```
> netdata INFO  : MAIN : Deleting data file "/home/stelios/netdata-inst//netdata/var/cache/netdata/dbengine/datafile-1-0000000154.ndf".
> netdata: database/engine/metadata_log/metadatalogapi.c:319: metalog_delete_dimension_by_uuid: Assertion `rd' failed.
> ```
> 
> (I lowered the dbengine disk space to force the delete)

I pushed the fix. Does it work for you now?

stelfrag:
> > I get
> > ```
> > netdata INFO  : MAIN : Deleting data file "/home/stelios/netdata-inst//netdata/var/cache/netdata/dbengine/datafile-1-0000000154.ndf".
> > netdata: database/engine/metadata_log/metadatalogapi.c:319: metalog_delete_dimension_by_uuid: Assertion `rd' failed.
> > ```
> > 
> > 
> > (I lowered the dbengine disk space to force the delete)
> 
> I pushed the fix. Does it work for you now?

yes that fixed it!!. Files deleted, TOMBSTONE commands generated. I will trace those now

database/rrd.h
mfundul: `rrdset_find_bytype()` is not dead code, it's being called by `rrdset_find_bytype_localhost()`, which is called by `rrdset_find_active_bytype_localhost()`. You could argue that `rrdset_find_bytype_localhost()` is dead code though.

When writing `rrdset_find_active_bytype_localhost()` I just wanted to make as few changes as possible, so I didn't refactor anything around it.

parser/parser.c
amoss: I don't think that Costa was explicit that it needs to go here and I do not think that we should expose it to the collectors as it adds no functionality, but adds a risk. Please move it as I have indicated.

database/engine/metadata_log/metadatalogapi.c
amoss: But in other places `RRD_ID_LENGTH_MAX `is used, and it is only 200.

amoss: There does not seem to be any length checking in `find_or_generate_guid` when the object is stored. This is a problem as it means there is no choice of constant length that will prevent a buffer overflow on retrieval.

parser/parser.c
mfundul: I will move it, but you do realize this is only possible because Stelios went above and beyond to implement pluggable language keywords.

database/engine/metadata_log/metadatalogapi.c
mfundul: `RRD_ID_LENGTH_MAX` is only good for charts unfortunately.

database/engine/metadata_log/logfile.c
```diff
@@ -273,18 +273,166 @@ static int check_metadata_logfile_superblock(uv_file file)
...
-    if (strncmp(superblock->magic_number, RRDENG_METALOG_MAGIC, RRDENG_MAGIC_SZ) ||
-        strncmp(superblock->version, RRDENG_METALOG_VER, RRDENG_VER_SZ)) {
+    if (strncmp(superblock->magic_number, RRDENG_METALOG_MAGIC, RRDENG_MAGIC_SZ)) {
         error("File has invalid superblock.");
         ret = UV_EINVAL;
     } else {
         ret = 0;
     }
-    error:
+    if (strncmp(superblock->version, RRDENG_METALOG_VER, RRDENG_VER_SZ)) {
```
amoss, line 282: This makes no sense as a way to store the version of the file structure. The versions are integers to allow simple comparison and to decide if the version on the disk is newer or older than the version in the code. Why is it being stored as a string inside a 16-byte array? This does not match the processing needs and simply adds a point of failure (the string is not an integer) that is not captured properly in this piece of code.

database/engine/metadata_log/logfile.c
```diff
@@ -273,18 +273,166 @@ static int check_metadata_logfile_superblock(uv_file file)
...
+    uv_file file;
+    uv_buf_t iov;
+    uv_fs_t req;
+    int ret;
+
+    ctx = metalogfile->ctx;
+    file = metalogfile->file;
+    iov = uv_buf_init(buf, len);
+    ret = uv_fs_read(NULL, &req, file, &iov, 1, offset, NULL);
+    if (unlikely(ret < 0 && ret != req.result)) {
```
amoss, line 327: Why are we testing for this?

mfundul: This may be an I/O error, permission error etc. Why shouldn't we test for this?

database/engine/metadata_log/logfile.c
```diff
@@ -273,18 +273,166 @@ static int check_metadata_logfile_superblock(uv_file file)
...
+                  metalogfile->fileno);
+            break;
+        }
+        if (metalogfile_read(metalogfile, buf, min_header_size, pos) < 0)
+            break;
+        header = (struct rrdeng_metalog_record_header *)buf;
+        if (METALOG_STORE_PADDING == header->type) {
+            info("%s: Skipping padding in metadata logfile %u-%u.", __func__, metalogfile->starting_fileno,
+                 metalogfile->fileno);
+            record_size = ALIGN_BYTES_FLOOR(pos + RRDENG_BLOCK_SIZE) - pos;
```
amoss, line 398: This doesn't look right. Why do you need the record bytes in the file to be aligned to steps of `RRDEND_BLOCK_SIZE`?

mfundul: This is not actively being used at the moment. If we ever utilize the `METALOG_STORE_PADDING` command it's going to align the record.

web/api/web_api_v1.c
amoss: I don't remember this point in the requirements - where is it (can I have a link?)

database/engine/metadata_log/logfile.c
mfundul: Ah, misread the code. This is indeed something else. If libuv has an internal error, such as out of memory, we do `fatal()`, otherwise we grab the I/O error, permission error etc and handle it.

parser/parser.c
amoss: I do. I've already told him that I love it three times and used the same technique to keep the new replication keywords local to streaming.

database/engine/metadata_log/logfile.c
amoss: Yes indeed - I wanted it to be an integer inside, not an ascii representation of an integer. Because it allows us to track cases like "all future versions" as `if (ver > CURRENT_VER)` and deal with them appropriately.

mfundul: You said you didn't care about ascii representation but it doesn't matter. I can do it binary and be done with it.

database/engine/metadata_log/logfile.c
amoss: The `ret<0` case is worth testing for (and handled in the next block), but the `ret>=req.result` seems a bit extreme (given the `POST` macro in the uv code).

database/engine/metadata_log/metadatalogapi.c
mfundul: Setting it to `PLUGINSD_LINE_MAX`. It's safe of overflows thanks to `max_bytes` in `find_object_by_guid(uuid_t *uuid, char *object, size_t max_bytes)`.

PR #8916, opened 2020-05-06
Fix support for Ubuntu 14.04 with Cloud support

prologic:
##### Summary

ssia

- Fixes #8905
- Fixes #7258

##### Component Name

- area/packaging
- area/ci

##### Test Plan

- [x] Test that it builds correctly (_without cloud_)
<details>
<summary>Details</summary>

```#!sh
prologic@Jamess-iMac
Wed May 06 15:13:20
~/NetData/netdata
 (ubuntu_1404_support) 0
$ cat Dockerfile.dev
FROM ubuntu:14.04 AS build

ARG DEPS=netdata

# Install all build dependencies
COPY packaging/installer/install-required-packages.sh /tmp/install-required-packages.sh
RUN /tmp/install-required-packages.sh --dont-wait --non-interactive ${DEPS}

# Copy Sources
WORKDIR /netdata
COPY . .

# Build
ARG BUILD=1
RUN [ -n "$BUILD" ] && ./netdata-installer.sh -u --dont-wait --dont-start-it --disable-go || echo "Skipping build ..."

CMD ["/usr/sbin/netdata", "-D"]
```

```#!sh
prologic@Jamess-iMac
Wed May 06 15:09:55
~/NetData/netdata
 (ubuntu_1404_support) 0
$ docker build -f Dockerfile.dev -t prologic/netdata:dev . && dki -t --rm -p 29999:19999 prologic/netdata:dev
Sending build context to Docker daemon  100.8MB
Step 1/9 : FROM ubuntu:14.04 AS build
 ---> 6e4f1fe62ff1
Step 2/9 : ARG DEPS=netdata
 ---> Using cache
 ---> 658bfc786452
Step 3/9 : COPY packaging/installer/install-required-packages.sh /tmp/install-required-packages.sh
 ---> Using cache
 ---> 0dd7c034b98e
Step 4/9 : RUN /tmp/install-required-packages.sh --dont-wait --non-interactive ${DEPS}
 ---> Using cache
 ---> 8b1bc49f4249
Step 5/9 : WORKDIR /netdata
 ---> Using cache
 ---> ac653e1a8a8b
Step 6/9 : COPY . .
 ---> e51203640a68
Step 7/9 : ARG BUILD=1
 ---> Running in ef6094123c69
Removing intermediate container ef6094123c69
 ---> a3835a474644
Step 8/9 : RUN [ -n "$BUILD" ] && ./netdata-installer.sh -u --dont-wait --dont-start-it --disable-go || echo "Skipping build ..."
[...]

Issue #8905, opened 2020-05-05
INFO: run netdata on ubuntu 14.04 LTS


ktsaou:
Ubuntu 14.04 LTS is EOL, but still if you have such a server, netdata can be run with this:

add a repo to get `libuv1-dev` (https://stackoverflow.com/a/49628607/4525767)

```sh
add-apt-repository ppa:acooks/libwebsockets6
```

export CFLAGS to disable the requirement of openssl host verification. This will lesser your security, but still the version of openssl installed on ubuntu 14.04 LTS cannot do better.

```sh
# as root, run:
export CFLAGS="-DACLK_SSL_ALLOW_SELF_SIGNED"
bash <(curl -Ss https://my-netdata.io/kickstart.sh)
```

With the above netdata is installed and it can communicate with netdata.cloud too.

(I haven't verified the kickstart works, because I used `netdata-installer.sh` from the `go-live` branch)

lassebm:
I've observed that while `CFLAGS` are persisted in `.environment` by `kickstart.sh`, it's ignored and overwritten by `netdata-updater.sh`, breaking the workaround during the nightly updates.

prologic:
> I've observed that while `CFLAGS` are persisted in `.environment` by `kickstart.sh`, it's ignored and overwritten by `netdata-updater.sh`, breaking the workaround during the nightly updates.

I have noticed this too. Btw what `CFLAGS` are you referring to here? I need to get back to this and get this working and working with Netdata cloud. The PR changes are not quite enough.

lassebm:
@prologic I need `-DACLK_SSL_ALLOW_SELF_SIGNED` for cloud connectivity on Ubuntu 14.04.

prologic:
> @prologic I need `-DACLK_SSL_ALLOW_SELF_SIGNED` for cloud connectivity on Ubuntu 14.04.

Perfect! Let me add that to this PR and do you mind helping me re-test this? I'm not sure I have an Ubuntu 14.04 VM but I'll go spin one up...

lassebm:
@prologic Sure, no problem.

zack-shoylev:
There are (related) CentOS, Ubuntu 14, Debian 7 issues:
https://github.com/netdata/netdata/issues/9152
https://github.com/netdata/netdata/issues/9153
@ktsaou told me this is probably caused by SSL problems with self-signed certs. On some systems autoconf may have to allow self-signed and also host verification will have to be disabled.

[...]

PR #8842, opened 2020-04-28
Add CI check for building against LibreSSL

prologic:
##### Summary

ssia

##### Component Name

- area/ci

##### Test Plan

- This PR and the new Checks
  - [ ] Should fail on master
  - [ ] Should pass on #8807

##### Additional Information

PR #9198, opened 2020-05-27
fix compilation for older systems

ktsaou:
##### Summary

This PR allows building netdata in operating systems:

1. that `clock_gettime()` is not part for libc, but `librt` or `libposix4`
2. that have an openssl version that does not support `X509_VERIFY_PARAM_set1_host()`

##### Component Name
netdata-installer.sh

##### Test Plan

On Centos 6 ACLK was disabled because of this:
```
checking if libmosquitto static lib is present (and builds)... no
```

The log was giving the error:

```
configure:8965: checking if libmosquitto static lib is present (and builds)
configure:8978: gcc -std=gnu99 -o conftest -O2    conftest.c externaldeps/mosquitto/libmosquitto.a -lcrypto -lssl >&5
externaldeps/mosquitto/libmosquitto.a(mosquitto.o): In function `mosquitto_lib_init':
mosquitto.c:(.text+0x5f): undefined reference to `clock_gettime'
externaldeps/mosquitto/libmosquitto.a(time_mosq.o): In function `mosquitto_time':
time_mosq.c:(.text+0x15): undefined reference to `clock_gettime'
collect2: ld returned 1 exit status
configure:8978: $? = 1
```

With this PR `-lrt` or whatever was needed to link with `clock_gettime()` is added to the above check, so the result is this:

```
checking if libmosquitto static lib is present (and builds)... yes
```

Because of an old OpenSSL library, we also see in the log:

```
checking for X509_VERIFY_PARAM_set1_host in -lcrypto... no
```

The following operating systems have been tested:

O/S|version|netdata compiles|with host validation?|ACLK working
----|:----:|:----:|:----:|:---:
centos|6.10|yes (requires EPEL repo for deps)|**NO**|yes
centos|7|yes|yes|yes
centos|8|yes|yes|yes
debian|7|no (cannot install libuv1)|-|-
debian|8|yes|**NO**|yes
debian|9|yes|yes|yes
debian|10|yes|yes|yes
ubuntu|12.04 LTS|no (cannot install libuv1)|-|-
ubuntu|14.04 LTS|yes (requires special repo for libuv1 #8905 #8916)|**NO**|yes
ubuntu|16.04 LTS|yes|yes|yes
ubuntu|18.04 LTS|yes|yes|yes
ubuntu|20.04 LTS|yes|yes|yes

So, this PR provides a working ACLK for centos 6, debian 8 and ubuntu 14.04 LTS.

### DIFF
--- a/aclk/aclk_lws_https_client.c
+++ b/aclk/aclk_lws_https_client.c
@@ -193,7 +193,8 @@ int aclk_send_https_request(char *method, char *host, char *port, char *url, cha
 #else
     i.ssl_connection = LCCSCF_USE_SSL;
 #endif
-#ifndef HAVE_X509_VERIFY_PARAM_set1_host
+#if defined(HAVE_X509_VERIFY_PARAM_set1_host) && HAVE_X509_VERIFY_PARAM_set1_host == 0
+#warning DISABLING SSL HOSTNAME VALIDATION BECAUSE IT IS NOT AVAILABLE ON THIS SYSTEM.
     i.ssl_connection |= LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK;
 #endif
 
--- a/aclk/aclk_lws_https_client.c
+++ b/aclk/aclk_lws_https_client.c
@@ -193,6 +193,9 @@ int aclk_send_https_request(char *method, char *host, char *port, char *url, cha
 #else
     i.ssl_connection = LCCSCF_USE_SSL;
 #endif
+#ifndef HAVE_X509_VERIFY_PARAM_set1_host
+    i.ssl_connection |= LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK;
+#endif
 
     i.port = atoi(port);
     i.address = host;
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -322,7 +322,8 @@ int aclk_lws_wss_connect(char *host, int port)
 #else
     i.ssl_connection = LCCSCF_USE_SSL;
 #endif
-#ifndef HAVE_X509_VERIFY_PARAM_set1_host
+#if defined(HAVE_X509_VERIFY_PARAM_set1_host) && HAVE_X509_VERIFY_PARAM_set1_host == 0
+#warning DISABLING SSL HOSTNAME VALIDATION BECAUSE IT IS NOT AVAILABLE ON THIS SYSTEM.
     i.ssl_connection |= LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK;
 #endif
     lws_client_connect_via_info(&i);
--- a/aclk/aclk_lws_wss_client.c
+++ b/aclk/aclk_lws_wss_client.c
@@ -322,6 +322,9 @@ int aclk_lws_wss_connect(char *host, int port)
 #else
     i.ssl_connection = LCCSCF_USE_SSL;
 #endif
+#ifndef HAVE_X509_VERIFY_PARAM_set1_host
+    i.ssl_connection |= LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK;
+#endif
     lws_client_connect_via_info(&i);
     return 0;
 }
--- a/configure.ac
+++ b/configure.ac
@@ -371,6 +371,12 @@ AC_CHECK_LIB(
     [SSL_LIBS="-lcrypto -lssl"]
 )
 
+AC_CHECK_LIB(
+    [crypto],
+    [X509_VERIFY_PARAM_set1_host],
+    [AC_DEFINE([HAVE_X509_VERIFY_PARAM_set1_host], [1], [X509_VERIFY_PARAM_set1_host])]
+)
+
 # -----------------------------------------------------------------------------
 # JSON-C library
 
@@ -586,7 +592,7 @@ if test "$enable_cloud" != "no"; then
     AC_MSG_CHECKING([if libmosquitto static lib is present (and builds)])
     if test -f "externaldeps/mosquitto/libmosquitto.a"; then
         LIBS_BKP="${LIBS}"
-        LIBS="externaldeps/mosquitto/libmosquitto.a ${OPTIONAL_SSL_LIBS}"
+        LIBS="externaldeps/mosquitto/libmosquitto.a ${OPTIONAL_SSL_LIBS} ${LIBS_BKP}"
         AC_LINK_IFELSE([AC_LANG_SOURCE([[#include "externaldeps/mosquitto/mosquitto.h"
                                          int main (int argc, char **argv) {
                                              int m,mm,r;
--- a/configure.ac
+++ b/configure.ac
@@ -374,8 +374,16 @@ AC_CHECK_LIB(
 AC_CHECK_LIB(
     [crypto],
     [X509_VERIFY_PARAM_set1_host],
-    [AC_DEFINE([HAVE_X509_VERIFY_PARAM_set1_host], [1], [X509_VERIFY_PARAM_set1_host])]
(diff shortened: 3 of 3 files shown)


aclk/aclk_lws_https_client.c
```diff
@@ -193,6 +193,9 @@ int aclk_send_https_request(char *method, char *host, char *port, char *url, cha
 #else
     i.ssl_connection = LCCSCF_USE_SSL;
 #endif
+#ifndef HAVE_X509_VERIFY_PARAM_set1_host
+    i.ssl_connection |= LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK;
```
amoss, line 197: This makes the ACLK insecure.

ktsaou:
@amoss any option that requires from the user to re-run the installer with another option is not acceptable. The installer should proceed and install netdata without user intervention.

I will make a change to inform the user about the relaxed security, at the end of the installation.

Could you please point the documentation that states that the link is confidential and secure? I will add a note there too.

amoss:
https://github.com/netdata/netdata/blob/master/aclk/README.md

(First paragraph)

amoss:
> @amoss any option that requires from the user to re-run the installer with another option is not acceptable.

Why is this unacceptable?

prologic:
There are two problems here

a) Any security conscious user or security teams at any company will either disallow the use of Netdata or disable the cloud functionality (_until we have an on-prom version_)

and

b) Very few users would actually pat attention to the messages printed unless we go to great lengths to suppress most of the build output and be a lot more terse.

Compromising on security should be the last thing we should consider IHMO.

ktsaou:
> Why is this unacceptable?

@amoss prior to this PR:

1. When there is an older openssl, compilation does not fail. Netdata is compiled, but ACLK cannot be established, although claiming works. This is broken behavior that leads to frustration.

2. The message you suggest has several issues: `The version of openssl on your system does not support a secure connection to the cloud. You can upgrade the library to fix this or DISABLE THE SECURITY ON THE LINK by adding CFLAGS="-DINSECURE_ACLK" to the installer.`

  - `upgrade the library`?
   Are you sure? Most probably such an upgrade will break the entire system. Nothing will work after that. This is a by far more dangerous advise, than running ACLK without hostname validation.

  - `DISABLE THE SECURITY ON THE LINK`?
   Only hostname validation is disabled. All the other security features are still there. The link is still encrypted and safe.

3. We disable hostname validation when the user is installing netdata on a system that DOES NOT SUPPORT HOSTNAME VALIDATION AT ALL! That operating system never supported this feature and will never support it. The user will not miss a feature his/her system supports.

So, what is the point of asking the user to re-run the installer with an additional option, if this is the only action he can take?

@prologic :

> Any security conscious user or security teams at any company will either disallow the use of Netdata or disable the cloud functionality (_until we have an on-prom version_)

Check again the table above. You say that a security conscious user or security team will disallow the use of netdata because on Centos 6, Debian 8 and Ubuntu 14.04 we automatically disable hostname validation that is not available in these systems, but they run production systems using Centos 6, debian 8 or ubuntu 14.04 LTS that do not support hostname validation at all. Hm... These are contradicting. Aren't they?

The whole point of this PR is to help people install netdata on older systems. THE USERS KNOW THEIR SYSTEMS ARE OLDER and we support the best security all the systems can do. Recent operating systems do not lose anything.

ktsaou:
updated the comment above. If you read it, please read it again.

amoss:
> updated the comment above. If you read it, please read it again.

I quite liked it the first time. I'll read it again...

ktsaou:
> I quite liked it the first time. I'll read it again...

This is the change:

Adding a special option for disabling hostname validation, may actually lead a lot of inexperienced users to include this option in their scripts for all O/Ses. So, it may have the opposite effect: users to run netdata without hostname validation, even when the system supports it.

amoss:
> > I quite liked it the first time. I'll read it again...
> 
> This is the change:
> 
> Adding a special option for disabling hostname validation, may actually lead a lot of inexperienced users to include this option in their scripts for all O/Ses. So, it may have the opposite effect: users to run netdata without hostname validation, even when the system supports it.

Yes - I agree this is quite a strong argument (and number (3) as well).

amoss:
@underhood Please be the second reviewer - please test it on Ubuntu 14.04 / CentOS 6 and something modern.

prologic:
I assume a valid option for users on older systems can also be to use our prebuilt static binary that is built against newer libraries?

underhood:
Should we at least add some warning to user? e.g. add the log "HOSTNAME CHECK DISABLED" to error log?
Also I would probably be safer with oposite logic -> if nothing is defined then use the hostname check... just in case

Ferroin:
> Also I would probably feel safer with oposite logic -> if nothing is defined then use the hostname check... just in case

On that note, has anybody tested this against LibreSSL? We quite literally _just_ got the build working correctly with it, I'd hate to have it broken again and require yet more fixes.

ktsaou:
I will swap `HAVE_` with `NO_`, but I don't know how to test libressl. Any directions?

amoss:
We had CI/CD checking this (https://github.com/netdata/netdata/pull/8842). @Ferroin / @prologic did it get disabled?

prologic:
> We had CI/CD checking this (#8842). @Ferroin / @prologic did it get disabled?

It got reverted for some reason I fail to.forget:

```
* f83b372d 2020-04-30 | Revert "Add CI check for building against LibreSSL (#8842)" [James Mills]
* 8081102f 2020-04-29 | Add CI check for building against LibreSSL (#8842) [James Mills]
```

Can re-add it back?

prologic:
https://github.com/netdata/netdata/pull/9216

amoss:
I don't know why it was reverted either, I've approved the PR to switch it back on. If you get an SRE to look at it then we can put it in and then @ktsaou can rebase his PR to get the libressl check shown here.

Ferroin:
> I don't know why it was reverted either, I've approved the PR to switch it back on. If you get an SRE to look at it then we can put it in and then @ktsaou can rebase his PR to get the libressl check shown here.

It's been fully approved and merged. @ktsaou if you rebase on latest master, it will pull in the CI check for LibreSSL.

Most of what I'm worried about here is whether or not the build itself works, because LibreSSL has a different API/ABI in some places. I suspect it will work fine with the most recent change (swapping to using host-checking by default instead of disabling it by default).

ktsaou:
rebased it

ktsaou:
worked! LibreSSL log:

```
checking for X509_VERIFY_PARAM_set1_host in -lcrypto... yes
```

underhood:
LGTM I will retest this just to be sure

underhood:
CentOS6:
configure:
```
checking for SHA256_Init in -lcrypto... yes
checking for X509_VERIFY_PARAM_set1_host in -lcrypto... no
configure: WARNING: DISABLING SSL HOSTNAME VALIDATION BECAUSE IT IS NOT AVAILABLE ON THIS SYSTEM.
```
config.h:
```
/* netdata ACLK */
#define ENABLE_ACLK 1

/* netdata dbengine usability */
/* #undef ENABLE_DBENGINE */

/* Pub/Sub API usability */
/* #undef ENABLE_EXPORTING_PUBSUB */

/* netdata HTTPS usability */
#define ENABLE_HTTPS 1
...
...
/* ssl host validation */
#define HAVE_X509_VERIFY_PARAM_set1_host 0
```
Connects to cloud after claiming.

zack-shoylev:
Just to confirm, in docs we should still recommend using the static build *first* for affected systems, correct? Further, in an ideal world, should the install script not switch to the static binary on these systems?

ktsaou:
> Just to confirm, in docs we should still recommend using the static build _first_ for affected systems, correct? Further, in an ideal world, should the install script not switch to the static binary on these systems?

yes to both

prologic:
> yes to both

+1

Issue #9043, opened 2020-05-14
Disabling cloud still performs requests to app.netdata.cloud


xPaw:
##### Bug report summary

I put `[global] enabled = false` in `/var/lib/netdata/cloud.d/cloud.conf`

Info api response:
```
	"cloud-enabled": false,
	"cloud-available": false,
	"agent-claimed": false,
	"aclk-available": false
```

##### OS / Environment

```
Debian 10
```

##### Netdata version

netdata v1.22.1-17-nightly

##### Component Name

Web/dashboard

##### Steps To Reproduce

1. Disable cloud
2. Open the dashboard and monitor requests in devtools to `sso/sign-in`

##### Expected behavior

When cloud is disabled, I expect netdata not to perform any external requests on the dashboard.

jacekkolasa:
@xPaw currently `cloud-enabled` setting is related more to Agent-Cloud connection (not Browser-Cloud), so we don't display Call To Actions to Cloud, notifications about insufficient dependencies or sign-in button. But we load the iframes, in case users sign in on different Agent, to have the Visited Nodes list. We don't store any information when you're not logged in, but if you want to prevent those calls, could you for example set `cloud base url = https://example.com`?

xPaw:
>  could you for example set cloud base url = https://example.com?

That does work, but feels like a hack, really.

The same issue happens with the registry, setting it to disabled still performs requests, and the only way to avoid that is to change the url.

jacekkolasa:
I agree, it's a hack, i'm giving this hint because we just don't have that feature yet... :( `"cloud-enabled": false` serves a little different purpose.

Another option would be setting `respect do not track policy = yes` in `/etc/netdata/netdata.conf` - this way no call to cloud is made, but you need to configure browser to add DNT flag. This is how it works in old Dashboard (accessible still with `/old` suffix). In new dashboard it doesn't work now because of a bug, but it will be fixed very soon (https://github.com/netdata/netdata/issues/9036)

xPaw:
The ideal solution I see is when cloud and registry are set to disabled, actually disable anything related to these features.

jacekkolasa:
I agree that we should allow disabling it completely, but for this we would need to add new flags. CC @cakrit @dim08

amoss:
[...]

PR #9213, opened 2020-05-28
dashboard v1.0.16

jacekkolasa:
- fixed navbar icons styling (alarm indicators above bell, fixed margins)
- slightly preciser way to resize dygraph after toggling left menu (fixed rare scenario when there was empty space in dygraph)
- leave door open for logout dropdown - no visible UI now but we'll be able to add dropdown in iframes without the need for dashboard update
- fix a rare issue with lost timeframe setting, which was happening when first panning was done on agent with high latency
- fix registry when Do Not Track is enabled
- support whitespaces in replicated-nodes urls
- add trailing slash in replicated-nodes links
- disable loop when checking for ACLK status (check just once, as in design)
- small performance improvement - less dependencies-loading cycles on init

includes:
https://github.com/netdata/dashboard/pull/81
https://github.com/netdata/dashboard/pull/82
https://github.com/netdata/dashboard/pull/84


underhood:
I don't know if you intend to fix also #9043 but my agent still makes requests to app.netdata.cloud even though:
```
	"cloud-enabled": false,
	"cloud-available": false,
	"agent-claimed": false,
	"aclk-available": false
```

jacekkolasa:
@underhood no, fix for that is not added, we wanted first to finish the first iteration of Custom Dashboards in the Cloud. cc @manos-saratsis

amoss:
It seems to be very broken on Debian 10. Panning a chart freezes the dashboard and then it becomes very unstable.

jacekkolasa:
i'm closing this PR. I haven't done any updates here because i spotted a place where there could be one bug and i didn't wanted it to be merged. I didn't had time to fix it since then, but I'll probably release a new version in next week.

Issue #8284, opened 2020-03-04
Add more robust checking of `make dist`.


Ferroin:
As of right now, our validation of `make dist` only reliably catches issues resulting from mandatory components in the install failing.

However, we now have a number of important optional components that we need to verify work correctly.

As a result, we need to update our checking to more robustly verify that installing from a tree prepared with `make dist` pulls everything in correctly.

prologic:
@Ferroin The "Checks / Dist" workflow is now migrated to GHA. Can you elaborate on what's needed here?

Ferroin:
It needs #8382 merged and that flag needs to be used for the `make dist` builds. However, #8382 is effectively blocked pending completion of #8378, which will resolve the build failures being seen in it.

prologic:
All yours then :)

PR #9218, opened 2020-05-29
Require cloud build to succeed in make dist checks.

Ferroin:
##### Summary

This updates our `make dist` check to only pass if building with Netdata Cloud functionality succeeds.

##### Component Name

area/ci

##### Test Plan

CI check passes on this PR.

##### Additional Information

Fixes: #8284

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -61,7 +61,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Prepare environment
         run: |
-          ./packaging/installer/install-required-packages.sh --dont-wait --non-interactive netdata-all
+          ./packaging/installer/install-required-packages.sh --dont-wait --non-interactive netdata
           sudo apt-get install -y libjson-c-dev libipmimonitoring-dev libcups2-dev libsnappy-dev \
                                   libprotobuf-dev libprotoc-dev libssl-dev protobuf-compiler \
                                   libnetfilter-acct-dev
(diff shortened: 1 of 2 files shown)


Ferroin:
> Should we refactor `.github/scripts/run_install_with_dist_file.sh` to accept arguments we can pass in from CI and have a few different things we can check? Not just assume that all checks will now ensure the Cloud functionality builds correctly, but other optional features as separate jobs?

I'd been thinking about this myself, but a lot more generically than just that.

Ferroin:
@prologic Could you possibly take a look into the failures here, I'm stumped. It looks like for some reason `install-required-packages.sh netdata` isn't pulling in `cmake`, but only in this particular GHA instance (if it weren't anywhere else, other checks would be failing and users would be complaining).

.github/workflows/checks.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Prepare environment
         run: |
-          ./packaging/installer/install-required-packages.sh --dont-wait --non-interactive netdata-all
+          ./packaging/installer/install-required-packages.sh --dont-wait --non-interactive netdata
```
prologic, line 64: I plan to remove `netdata` as an option to this tool.

prologic:
@Ferroin You need to see what the dist build failures are. The CentOS 8.x failures are a KP and I'm working on them. Please have a look :)

.github/workflows/checks.yml
Ferroin: Unless you're changing 'netdat-all' to be behave like the current 'netdata' option does, don't. We should not be unconditionally pulling in Node.js Python, and all of our Python dependencies on installs on user systems.

Ferroin:
@prologic Finally got around to updating this and getting it working correctly, if you could take another look at it that would be appreciated.

prologic:
Merged on behalf of @Ferroin (_while away on vacation_); as this directly impacts our CI and improves it.

PR #9226, opened 2020-05-30
[WIP] ebpf plugin linked with static library

### LINKED ISSUE #9225: Change ebpf.plugin to compile with static library
#### Feature idea summary
`ebpf.plugin` has only pointers to functions that are filled when it is loaded, but we do not need them when we link the binary statically, it is necessary to remove these references to compile Netdata statically.

We have different possibilities that can be used to bring the static library for the eBPF.plugin, but all them must remove the dynamic loading from the ebpf.plugin. The following options have been considered now:

- Link the libraries during compilation time, more details [here](https://github.com/netdata/netdata/issues/9225#issuecomment-647779656) .
- Build everything on kernel-collector repository, more details [here](https://github.com/netdata/netdata/issues/9225#issuecomment-647791250).
- Final decision [here](https://github.com/netdata/netdata/issues/9225#issuecomment-648741678)

##### Expected behavior
To have conditions to use `-static` package from `kernel-collector` repo.

thiagoftsm:
##### Summary
Fixes #9225 

 Bring the necessary changes to compile ebpf plugin with static library.
##### Component Name
ebpf collector
##### Test Plan

We are missing few changes to have conditions to create a Test Plan. @prologic, please, change the autoconf files and `netdata-installer.sh` to get the static library and link with `ebpf.plugin` every time the environment variable `STATIC=1` is set.

### DIFF
--- a/collectors/ebpf.plugin/ebpf_process.c
+++ b/collectors/ebpf.plugin/ebpf_process.c
@@ -23,15 +23,15 @@ static netdata_publish_syscall_t *process_publish_aggregated = NULL;
 
 static ebpf_functions_t process_functions;
 
+static int *map_fd = NULL;
+
 #ifndef STATIC
 /**
  * Pointers used when collector is dynamically linked
  */
 
 //Libbpf (It is necessary to have at least kernel 4.10)
-static int (*bpf_map_lookup_elem)(int, const void *, void *);
-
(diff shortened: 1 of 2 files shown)


thiagoftsm:
After an internal talk we decided to folow in another direction.

PR #9236, opened 2020-05-31
UrlService add support get headers

vsc55:
##### Summary
Support has been added to return the "headers" returned by the server and to obtain data such as cookis.

##### Component Name
collectors\python.d.plugin\python_modules\bases\FrameworkServices\UrlService.py

##### Test Plan
I am using it in a plugin that I have created to get statistics from Kerio Connect and it works without problem.
https://github.com/vsc55/netdata/blob/plugin-kerio-connect/collectors/python.d.plugin/kerio_connect/kerio_connect.chart.py

```
def _headers_get_cookies(self, headers):
    data_return = dict()
    if 'Set-Cookie' in headers:
        for item in headers['Set-Cookie'].split(";"):
            cookie = item.split("=", 1)
            data_return.update( {
                cookie[0] : "" if len(cookie) == 1 else cookie[1]
            } )
    return data_return

def _header_update(self):
    self.header = { 
        'content-type': 'application/json' 
    }
    if self._api_token:
        self.header.update( {
            'X-Token' : self._api_token
        } )
    if self._cookie_session:
        self.header.update( {
            "Cookie": "SESSION_CONNECT_WEBADMIN={0}".format(self._cookie_session)
        } )
    self._manager = self._build_manager()

def _login(self):
    # INFO: Get headers and save cookie
    headers, raw = self._get_raw_data_with_headers()
    cookies = self._headers_get_cookies(headers)
    self._cookie_session = "" if not 'SESSION_CONNECT_WEBADMIN' in cookies else cookies['SESSION_CONNECT_WEBADMIN']

def _logout(self):
    # INFO: Update header with cookie saved
   self._header_update()
   raw = self._get_raw_data()
```

##### Additional Information


ilyam8:
Hi @vsc55 ! Thx for contribution! I will look at it later.

A little hint, might be handy

If you need to iterate over a structure it is easier to do something like that

- define list of stats (`.` means it is a dict)

https://github.com/netdata/netdata/blob/ef3b10c6a1dc3b9fef01f3f384201cee922c580e/collectors/python.d.plugin/elasticsearch/elasticsearch.chart.py#L85-L97

- use one func to get all the data

https://github.com/netdata/netdata/blob/ef3b10c6a1dc3b9fef01f3f384201cee922c580e/collectors/python.d.plugin/elasticsearch/elasticsearch.chart.py#L763-L774

---

Profit - it is easy to observe what metrics are being collector having a list

vsc55:
Hi,
Any problem with the modification?

ilyam8:
I have no time to look into it, i have a feeling that we need something like `_get_response` instead of all those `_get_raw_data_with_*`, but i am not sure atm.

vsc55:
Ok, no problem.
I didn't want to modify the object any more to prevent another plugin from failing.

ilyam8:
Yes, it could be not as easy as it sound, i need to check it.

ilyam8:
@vsc55 i think we need move all code from `_get_raw_data_with_status` to `_do_request`

It should return response

https://github.com/netdata/netdata/blob/5a1d123ba762a118bcc1989d9f37022d016dcca0/collectors/python.d.plugin/python_modules/bases/FrameworkServices/UrlService.py#L148-L156

`_get_raw_data_with_status` should call it and do 

https://github.com/netdata/netdata/blob/5a1d123ba762a118bcc1989d9f37022d016dcca0/collectors/python.d.plugin/python_modules/bases/FrameworkServices/UrlService.py#L157-L159

And use can use `_do_request` in your code, it contains everything, it is beeter then keep adding `with_*` functions

vsc55:
Hi,
I have already updated the object, to see what you think.

What I was thinking about is adding support to be able to save and edit the query cookies.
What do you think about the idea?

ilyam8:
@vsc55 my suggestion was to not keep adding `_get_with_*` methods but add `_do_requsest` - it allows you to do whatever you need in your custom collector.

We dont need methods in UrlService which we dont use in our collectors. This _get_with_headers is needed only by you atm.

ilyam8:
@vsc55 looks good, we release v1.23.0 on Wed, this PR will be merged after the release

collectors/python.d.plugin/python_modules/bases/FrameworkServices/UrlService.py
```diff
@@ -120,22 +120,36 @@ def _get_raw_data(self, url=None, manager=None, **kwargs):
...
             self.error('Url: {url}. Error: {error}'.format(url=url or self.url, error=error))
             return None
 
-        if status == 200:
-            return data
+        if response.status == 200:
+            if isinstance(response.data, str):
+                return response.data
+            return response.data.decode(errors='ignore')
+
```
ilyam8, line 132: remove the empty line

ilyam8:
i checked the diff, installed the branch and tested it - no problems. 

@vsc55 thx!

PR #9243, opened 2020-06-01
Fix ebpf program load with musl

### LINKED ISSUE #9242: Musl cannot load ebpf programs
##### Bug report summary
According official [documentationt](https://wiki.musl-libc.org/functional-differences-from-glibc.html) musl does not work with the current flags we are passing to `dlopen`, it is necessary to change them.

##### OS / Environment
```
All 
```

##### Netdata version
netdata v1.22.1-182-g41c31b52

##### Component Name
ebpf collector

##### Steps To Reproduce

1. Install Netdata using script `kickstart_static64.sh`

##### Expected behavior
To be able to load ebpf programs with musl

thiagoftsm:
##### Summary
Fixes #9242
##### Component Name
ebpf collector

##### Test Plan

@prologic how can we test the `kickstart_static64` with this PR?


thiagoftsm:
This is not enough to use static link with musl, we changed the direction.

PR #9246, opened 2020-06-01
Add new exporting "home base" document

joelhans:
##### Summary

This PR creates a new "home base" for the exporting engine at `/docs/exporting/README.md` to let the original file at `/exporting/README.md` serve as a reference guide with the complete configuration options, and so on.

The goal here is to create a shorter document to help onboard users to exporting and/or get the migrated from backends quickly.

##### Component Name

area/exporting
area/docs

### DIFF
--- a/docs/export/README.md
+++ b/docs/export/README.md
@@ -83,7 +83,8 @@ For details on all the configuration options, see the [exporting reference](/exp
 
 > When migrating from backends, keep in kind that most settings for the exporting engine are identical to their backend
 > counterpart. You should be able to copy your backends settings into `exporting.conf` to complete your migration. The
-> only setting that won't work is `host tags`&mdash;use [`host labels`](/docs/tutorials/using-host-labels.md) instead.
+> only setting that won't work is `host tags`. Instead, use [host labels](/docs/tutorials/using-host-labels.md) to tag
+> your exported metrics.
 
 Restart your Agent to begin exporting to the destination of your choice. Because the Agent exports metrics as they're
 collected, you should start seeing data in your external database after only a few seconds.
@@ -92,6 +93,7 @@ collected, you should start seeing data in your external database after only a f
 
 -   [Exporting reference guide](/exporting/README.md)
 -   [Backends (deprecated)](/backends/README.md)
+-   [Use host labels to organize systems, metrics, and alarms](/docs/tutorials/using-host-labels.md)
 -   [Database engine](/database/engine/README.md)
 -   [Change how long Netdata stores metrics (long-term storage)](/docs/tutorials/longer-metrics-storage.md)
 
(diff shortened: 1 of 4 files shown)


docs/export/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+Next, edit and uncomment any other lines necessary to connect the exporting engine to your endpoint. You must change the
+`destination` setting in most situations.
+
+For details on all the configuration options, see the [exporting reference](/exporting/README.md#configuration).
+
+> When migrating from backends, keep in kind that most settings for the exporting engine are identical to their backend
+> counterpart. You should be able to copy your backends settings into `exporting.conf` to complete your migration. The
+> only setting that won't work is `host tags`&mdash;use [`host labels`](/docs/tutorials/using-host-labels.md) instead.
```
thiagoftsm, line 86: @joelhans I think there is a problem with the text "`host tags`&mdash;use", we do not have spaces between the values and I do not see a specific reason to use `&mdash;`.

docs/export/README.md
```diff
@@ -0,0 +1,100 @@
...
+Let's cover the process of enabling an exporting connector, using the Graphite connector as an example. These steps can
+be applied to other connectors as well.
+
+> If you are migrating from the deprecated backends system, this quickstart will also help you update your configuration
+> to the new format.
+
+Open the `exporting.conf` file with `edit-config`.
+
+```bash
+cd /etc/netdata   # Replace this path with your Netdata config directory, if different as found in the steps above
```
vlvkobal, line 31: "as found in the steps above" - which steps? It looks like this part should be removed.

docs/export/README.md
```diff
@@ -0,0 +1,100 @@
...
+    # send charts matching = *
+    # send hosts matching = localhost *
+```
+
+Next, edit and uncomment any other lines necessary to connect the exporting engine to your endpoint. You must change the
+`destination` setting in most situations.
+
+For details on all the configuration options, see the [exporting reference](/exporting/README.md#configuration).
+
+> When migrating from backends, keep in kind that most settings for the exporting engine are identical to their backend
```
vlvkobal, line 84: kind -> mind

docs/export/README.md
```diff
@@ -0,0 +1,100 @@
...
+> counterpart. You should be able to copy your backends settings into `exporting.conf` to complete your migration. The
+> only setting that won't work is `host tags`. Instead, use [host labels](/docs/tutorials/using-host-labels.md) to tag
```
vlvkobal, line 86: Should we mention here anything about the `type` option and section names, like in https://github.com/joelhans/netdata/blob/exporting/exporting/README.md#L24?

joelhans: Good call. I added a mention earlier in the document.

docs/export/README.md
joelhans: Good catch.

joelhans:
Thanks, gentlemen!

PR #9248, opened 2020-06-02
ipfs-plugin: change HTTP method to make it compatible with ipfs 0.5.0+

RubenKelevra:
##### Component Name

python.d plugin for IPFS

https://docs.ipfs.io/reference/http/api/#api-v0-stats-bw

##### Test Plan

Tested in the default installation under arch-linux

### DIFF
--- a/collectors/python.d.plugin/ipfs/ipfs.chart.py
+++ b/collectors/python.d.plugin/ipfs/ipfs.chart.py
@@ -63,6 +63,7 @@ class Service(UrlService):
         self.order = ORDER
         self.definitions = CHARTS
         self.baseurl = self.configuration.get('url', 'http://localhost:5001')
+        self.method = "POST"
         self.do_pinapi = self.configuration.get('pinapi')
         self.__storage_max = None
 


collectors/python.d.plugin/ipfs/ipfs.chart.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, configuration=None, name=None):
         self.order = ORDER
         self.definitions = CHARTS
         self.baseurl = self.configuration.get('url', 'http://localhost:5001')
+        self.method = "POST"
```
ilyam8, line 66: Well, it breaks compatibility with `ipfs` before `0.5.0`

It seems we need to check version first and then to set method. Is there an enpdoint th get version?

ilyam8: Jus realized. A quick fix is to add 2nd job to the module config

https://github.com/netdata/netdata/blob/dc3f425a7d2992a8ea3df2c15622e032a37bf512/collectors/python.d.plugin/ipfs/ipfs.conf#L74-L77

=>

```yaml
localhost:
  name   : 'local'
  url    : 'http://localhost:5001'
  pinapi : no

localhost_v5:
  name   : 'local'
  url    : 'http://localhost:5001'
  method : POST
  pinapi : no
```

ilyam8: where can i read about those issues? can you share the link?

RubenKelevra: @ilyam8 yeah sure, the improvements are mentioned in the [blog](https://blog.ipfs.io/2020-04-28-go-ipfs-0-5-0/). But there was a massive amount of fixes that are not fully mentioned.

Running 0.4.x is highly discouraged, because it actually slows down the whole network.

RubenKelevra: @ilyam8 here's the full changelog:

https://github.com/ipfs/go-ipfs/releases/tag/v0.5.0

ilyam8:
@dim08 latest version of ipfs is [v0.5.0](https://github.com/ipfs/go-ipfs/releases/tag/v0.5.0), i was released on April 28.

This PR makes ipfs collector compatible with the new version, but breaks for older versions.

@RubenKelevra states that new version has a lot of fixes and what is more important - performance fixes.

See our discussion

> https://github.com/netdata/netdata/pull/9248#discussion_r433640599

I am personally ok with merging this PR, it is still possible to get it working for users with old version of ipfs by adding `method: GET` configuration parameter to the configuration file.

Your call.

RubenKelevra:
@ilyam8 @dim08 

Any progress on that? :)

There's no stable version of go-ipfs. If a new version is released, the old version is EOL.

go-ipfs 0.6 is right around the corner (currently at RC-6) - so this plugin just works with a 2 version outdated daemon without manual intervention.

ilyam8:
@RubenKelevra we plan to release v1.23.0 in a few days, this PR wont be merged before it.

RubenKelevra:
Okay :)

Issue #9109, opened 2020-05-20
Netdata Cloud Agent on FreeBSD


manjotsc:
Can't add netdata to netdata cloud.

![Inked82404344-0f5b3f00-9a2f-11ea-8955-df549ec7ff5c_LI](https://user-images.githubusercontent.com/51430971/83343055-dbe8a200-a2c3-11ea-85f4-03332f42a692.jpg)

ilyam8:
@manjotsc hi, how did you install netdata?

manjotsc:
freebsd : pkg install http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/netdata-1.22.1.txz

ilyam8:
@manjotsc we dont maintain [netdata freebsd pkg](https://www.freebsd.org/cgi/ports.cgi?query=netdata&stype=all)

@mmokhi can you help?

manjotsc:
@mmokhi help needed

Thanks,

stevenh:
This is because of the libwebsockets and mosquitto, why are these internal deps and static?

manjotsc:
@stevenh where can I get those for FreeBSD.
Thanks

stevenh:
They are in the ports tree but they are old.

I have an updated port I'm working on which now compiles and links to shared libs but still doesn't work I get the following:
```
2020-05-30 01:04:06: netdata INFO  : ACLK_Main : Retrying to establish the ACLK connection in 128.722 seconds
2020-05-30 01:06:15: netdata INFO  : ACLK_Main : Attempting to establish the agent cloud link
2020-05-30 01:06:15: netdata INFO  : ACLK_Main : Retrieving challenge from cloud: app.netdata.cloud 443 /api/v1/auth/node/XXX/challenge
2020-05-30 01:06:15: netdata INFO  : ACLK_Main : aclk_send_https_request GET
2020-05-30 01:06:15: netdata ERROR : ACLK_Main : Challenge failed:  (errno 22, Invalid argument)
```
I suspect my patching of `mosquitto` may have an issue as `netdata` depends on a custom version :(

stevenh:
Ok found the issue if `libwebsockets` is compiled with `IPv6` you get this error, at least if the host itself doesn't have valid `IPv6` configuration.

stevenh:
@manjotsc the following 3 bug requests include the patches needed to add cloud support the ports tree version of netdata.

[...]

PR #9251, opened 2020-06-02
feat: Allow use of shared libraries

stevenh:
##### Summary
Allow the use of shared libraries for libwebsockets and mosquitto if available for cloud support.

This will make platform ports e.g. FreeBSD as external baked in versions can be eliminated.

I've based this off a patch created for FreeBSD's netdata port which uses external versions of these too libraries as the the build process doesn't support additional external downloads.

Fixes: #9109

##### Component Name
Build system

##### Test Plan
1. Build against external libs via FreeBSD ports.
2. Build against static libs via standard netdata build pipeline.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -669,8 +669,8 @@ netdata_SOURCES = $(NETDATA_FILES)
 
 if ENABLE_ACLK
 netdata_LDADD = \
-    externaldeps/mosquitto/libmosquitto.a \
-    externaldeps/libwebsockets/libwebsockets.a \
+    $(OPTIONAL_MOSQUITTO_LIBS) \
+    $(OPTIONAL_WEBSOCKETS_LIBS) \
     $(OPTIONAL_LIBCAP_LIBS) \
     $(NETDATA_COMMON_LIBS) \
     $(NULL)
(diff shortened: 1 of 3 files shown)


amoss:
Hello. When I look at the changes it seems to be dominated by removals of the static json-c option. I think that you might need to rebase on a newer tip of the tree.

Have you seen the approach that we are using in https://github.com/netdata/netdata/pull/9776?

The basic idea is that we have bundling options for all of our deps to avoid installation problems on systems without those libraries, but that we start adding explicit overrides so that package maintainers can switch off the bundling and force the use of another library. I think this would be a solution for the libraries here on FreeBSD.

stevenh:
The idea was to allow local libraries via `OPTIONAL_MOSQUITTO_LIBS` and `OPTIONAL_WEBSOCKETS_LIBS` for those respective libs, I got it working locally but needed more time to make it work here, which I unfortunately haven't had time for recently.

Will have have a look at the linked PR, but can't guarantee any time soon I'm afraid.

underhood:
@stevenh the latest master already has the ability to use libwebsockets as shared library.
Libmosquitto usage as shared library is not currently possible as we use modified version with custom patch

PR #9252, opened 2020-06-02
Metadata log compaction

### LINKED ISSUE #8902: Metadata log compaction
The [metadata log](https://github.com/netdata/docs/blob/master/architecture/agent/db/database-log-format-pros-cons.md) needs to be limited in disk space size. Implement a log compaction algorithm.

#### Analysis

We define a log record to be the set of consecutive `Plugins.D` lines in the metadata log file that define a "create" or a "delete" operation on an object (host/chart/dimension). The log records are many more than the actual objects in memory, e.g. 38 records for 12 objects.

##### Criteria

The criteria to trigger compaction can be:
- Exceeding some storage space quota of the journal files themselves.
- Exceeding some memory space quota of any journal file records that may be in the agent memory.
- Exceeding a threshold of the duplication factor. 1 would mean that each object is only defined once in the logs. 2 would mean that there are 2 records per object definition on average.

##### Algorithm

All "create" and "delete" operations that correspond to non-existing in-memory objects can safely be ignored. Moreover, any "delete" operations that correspond to objects that are in-memory should be ignored as well, although this would be an error condition since it would be impossible to recreate the same object (same GUID).

Now, when there is a "create" operation and the objects actually exist, one needs to create a new record for the newly created compacted part of the log. We dump the object from memory instead of copying the record from disk. This guarantees we have the latest version of the object after compaction. If we've already created an object record we don't have to dump it into the metadata log again.

Here are the rules for merging the log files:

1. When compaction begins, a new file is created with a name of `journal.N-N` where `journal.N` is the file with the oldest records or the current file being scanned.
2. Whenever the compaction process finishes scanning the records of `journal.X` the file `journal.N-X` is renamed to `journal.N-(X+1)`.
    - The first number (`N`)is the name this compacted file will eventually get.
    - The second number (`X+1`) is a pointer to the log file whose records are currently being compacted.
3. Any log file `journal.X` can be safely deleted if `journal.(X-c)-(X+d)` also exists, where `c >= 0` and `d > 0`, since all of its contents have already been compacted.
    - In other words the first number must be smaller or equal and the second number must be greater.
4. Whenever the `journal.X-Y` file hits the size limit or compaction ends, it is renamed to `journal.X`.
   - It should be impossible that the old `journal.X` hasn't been deleted yet because the compacted form should always be smaller (Edge case of same size is a no-op).
   - A new file is created following rule 1.
5. Compaction may stop at any point or when you reach the log file with the latest records.
   - The latest log file is actively being updated so it is not immutable like the other files.

Those rules should guarantee fault tolerance during compaction.

#### Steps

- When replaying metadata log during start-up create an internal list of active and inactive records.
- Maintain the record list per metadata log file when new metadata are generated.
- New records also need to be generated for the new metadata log files when the compaction operation is triggered.
- Periodically run the compaction algorithm in a single thread:
    - when the criteria are met (as described above in the Analysis section).
- Maintain a temporary lookup structure for log records during the compaction operation.
- Implement concurrency control to protect simultaneous access from the agent and the compaction process itself.
- During startup of dbengine, compaction cleanup process needs to delete any invalid files generated by compaction according to the algorithm.
- During startup of dbengine, compaction cleanup process needs to rename any valid temporary log files according to the algorithm.

mfundul:
##### Summary
Fixes #8902

The metadata log needs to be limited in disk space size. This PR implements a log compaction algorithm as described in #8902.

##### Component Name
database
daemon
##### Test Plan
You may start/stop the agent many times so as to accelerate the rate at which the metadata logs grow. Observe in `error.log` the ratio of `records` to `objects`:
```
netdata INFO  : MAIN : records=2057 objects=1661
```

When this ratio is over 1.5 the compaction process should trigger, which should decrease the ratio and replace the metadata log files with new ones.

You may also kill the agent during compaction to test fault-tolerance. During start-up, the agent should decide what to do with the compaction temporary files and the original metadata log files.

### DIFF
--- a/database/engine/metadata_log/compaction.c
+++ b/database/engine/metadata_log/compaction.c
@@ -1,0 +1,304 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+#define NETDATA_RRD_INTERNALS
+
+#include "metadatalog.h"
+
+void after_compact_old_records(struct metalog_worker_config* wc)
+{
+    struct metalog_instance *ctx = wc->ctx;
+    int error;
+
+    mlf_flush_records_buffer(wc, &ctx->compaction_state.records_log, &ctx->compaction_state.new_metadata_logfiles);
+    uv_run(wc->loop, UV_RUN_DEFAULT);
+
+    error = uv_thread_join(wc->now_compacting_files);
+    if (error) {
+        error("uv_thread_join(): %s", uv_strerror(error));
+    }
+    freez(wc->now_compacting_files);
+    /* unfreeze command processing */
+    wc->now_compacting_files = NULL;
+
+    wc->cleanup_thread_compacting_files = 0;
+
+    /* interrupt event loop */
+    uv_stop(wc->loop);
+
+    info("Finished metadata log compaction (id:%"PRIu32").", ctx->current_compaction_id);
+}
+
+static void metalog_flush_compaction_records(struct metalog_instance *ctx)
+{
+    struct metalog_cmd cmd;
+    struct completion compaction_completion;
+
+    init_completion(&compaction_completion);
+
+    cmd.opcode = METALOG_COMPACTION_FLUSH;
+    cmd.record_io_descr.completion = &compaction_completion;
+    metalog_enq_cmd(&ctx->worker_config, &cmd);
+
+    wait_for_completion(&compaction_completion);
+    destroy_completion(&compaction_completion);
+}
+
+/* The caller must have called metalog_flush_compaction_records() before to synchronize and quiesce the event loop. */
+static void compaction_test_quota(struct metalog_worker_config *wc)
+{
+    struct metalog_instance *ctx = wc->ctx;
+    struct logfile_compaction_state *compaction_state;
+    struct metadata_logfile *oldmetalogfile, *newmetalogfile;
+    unsigned current_size;
+    int ret;
+
+    compaction_state = &ctx->compaction_state;
+    newmetalogfile = compaction_state->new_metadata_logfiles.last;
+
+    oldmetalogfile = ctx->metadata_logfiles.first;
+
+    current_size = newmetalogfile->pos;
+    if (unlikely(current_size >= MAX_METALOGFILE_SIZE && newmetalogfile->starting_fileno < oldmetalogfile->fileno)) {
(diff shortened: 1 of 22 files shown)


stelfrag:
I notice that I have a file structure like this
```
525636  12:37 metadatalog-00000-00006.mlf
632730  12:38 metadatalog-00000-00007.mlf
```

`cat *.mlf | strings -n 5 | grep CHART  | wc` will give me **320** charts

I stop and restart the server and I end up with 
```
288099 12:38 metadatalog-00000-00007.mlf
  4096 12:38 metadatalog-00000-00008.mlf
```

`cat *.mlf | strings -n 5 | grep CHART  | wc` shows **20** charts

mfundul:
> 
> 
> I notice that I have a file structure like this
> 
> ```
> 525636  12:37 metadatalog-00000-00006.mlf
> 632730  12:38 metadatalog-00000-00007.mlf
> ```
> 
> `cat *.mlf | strings -n 5 | grep CHART | wc` will give me **320** charts
> 
> I stop and restart the server and I end up with
> 
> ```
> 288099 12:38 metadatalog-00000-00007.mlf
>   4096 12:38 metadatalog-00000-00008.mlf
> ```
> 
> `cat *.mlf | strings -n 5 | grep CHART | wc` shows **20** charts

Can you send me the error log of this run, or the mlf files, or both?

database/engine/metadata_log/compaction.c
```diff
@@ -0,0 +1,384 @@
...
+            break;
+        case GUID_TYPE_DIMENSION:
+            rd = metalog_get_dimension_from_uuid(ctx, uuid);
+            if (rd) {
+                if (ctx->current_compaction_id > rd->state->compaction_id) {
+                    rd->state->compaction_id = ctx->current_compaction_id;
+                    buffer = metalog_update_dimension_buffer(rd);
+                    metalog_commit_record(ctx, buffer, METALOG_COMMIT_CREATION_RECORD, uuid, 1);
+                } else {
+                    info("Chart has already been compacted, ignoring record.");
```
stelfrag, line 114: This should be dimension not chart

stelfrag:
The latest commit seems to fix the missing charts issue

PR #9255, opened 2020-06-02
Fixes documentation ambiguity leading into issue #8239

### LINKED ISSUE #8239: health:  simple pattern doesnt work well for `of` and `foreach`
##### Bug report summary

Alarm line `lookup` [documentation](https://docs.netdata.cloud/health/reference/#alarm-line-lookup)

> of DIMENSIONS is optional and has to be the last parameter. Dimensions have to be separated by , or |. The space characters found in dimensions will be kept as-is (a few dimensions have spaces in their names). This accepts Netdata simple patterns and the match-ids and match-names options affect the searches for dimensions.

Simple patterns [documentation](https://docs.netdata.cloud/libnetdata/simple_pattern/#netdata-simple-patterns)

> Simple patterns are a space separated list of words

___

Following two things are mutually exclusive:

- The space characters found in dimensions will be kept as-is (a few dimensions have spaces in their names).

- This accepts Netdata simple patterns

It is a common situation when **i want to lookup for not all dimensions but some of them** and i dont want to give a full list (why do we have simple pattern then?) or i dont know dim names, because they are dynamic.

For exampling i want to find sum of all dims except from `this` and `that`

```cmd
  lookup: sum -1m unaligned absolute match-names of !this !that *
```

^^ it is a simple pattern, but it doesnt work because of `The space characters found in dimensions will be kept as-is`.

##### OS / Environment

```
any
```

##### Netdata version

 
`netdata v1.20.0-57-g1ea0d8d0`

##### Component Name

`health`

##### Steps To Reproduce

1. use simple pattern with spaces in `of`
2. see it doesnt work

##### Expected behavior

Simple pattern works according its documentation for `of` and `foreach`.

underhood:
##### Summary

Fixes #8239

After studying the code, testing, and then verifying with colleagues the functionality works but the syntax is a bit different than the expected *(which confuses people and makes them think it is broken)*. The issue is therefore solved by small clarification in the documentation.

##### Component Name
Documentation

##### Additional Information

### DIFF
--- a/health/REFERENCE.md
+++ b/health/REFERENCE.md
@@ -200,8 +200,8 @@ Everything is the same with [badges](../web/api/badges/). In short:
 
 -   `of DIMENSIONS` is optional and has to be the last parameter. Dimensions have to be separated
      by `,` or `|`. The space characters found in dimensions will be kept as-is (a few dimensions
-     have spaces in their names). This accepts Netdata simple patterns *(with rules separated by 
-     `,` or `|` instead of space)* and the `match-ids` and `match-names` options affect the searches
+     have spaces in their names). This accepts Netdata simple patterns _(with `words` separated by
+     `,` or `|` instead of spaces)_ and the `match-ids` and `match-names` options affect the searches
      for dimensions.
 
 -   `foreach DIMENSIONS` is optional, will always be the last parameter, and uses the same `,`/`|`
--- a/health/REFERENCE.md
+++ b/health/REFERENCE.md
@@ -200,8 +200,9 @@ Everything is the same with [badges](../web/api/badges/). In short:
 
 -   `of DIMENSIONS` is optional and has to be the last parameter. Dimensions have to be separated
      by `,` or `|`. The space characters found in dimensions will be kept as-is (a few dimensions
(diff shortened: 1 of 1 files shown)


health/REFERENCE.md
```diff
@@ -200,8 +200,9 @@ Everything is the same with [badges](../web/api/badges/). In short:
 
 -   `of DIMENSIONS` is optional and has to be the last parameter. Dimensions have to be separated
      by `,` or `|`. The space characters found in dimensions will be kept as-is (a few dimensions
-     have spaces in their names). This accepts Netdata simple patterns and the `match-ids` and
-     `match-names` options affect the searches for dimensions.
+     have spaces in their names). This accepts Netdata simple patterns *(with rules separated by 
+     `,` or `|` instead of space)* and the `match-ids` and `match-names` options affect the searches
```
ilyam8, line 204: i think they are `words`, not `rules`

joelhans, line 204:
```suggestion
     have spaces in their names). This accepts Netdata simple patterns _(with rules separated by 
     `,` or `|` instead of a space)_ and the `match-ids` and `match-names` options affect the searches
```

PR #9257, opened 2020-06-03
Doc: Add missing slash

oneoneonepig:


ilyam8:
thx @oneoneonepig ✌️

joelhans:
@mfundul I remember when we were working on this document that you wanted to leave the initial slash out, because `cloud.d/` might not always be exactly in `/var/lib/netdata`. Correct?

I wonder if we either need to be more explicit about why the initial slash isn't there, or add the initial slash with a caveat that the user's `cloud.d/` directory may be in a different place altogether.

ilyam8:
@joelhans i think it depends on install path. Anyway `var/...` looks not correct to me.

PR #9261, opened 2020-06-03
Added GUID keyword to streaming protocol

### LINKED ISSUE #8918: Add GUIDs to streaming protocol
Add support for passing GUIDs through streaming. This way, an object of the slave should have the same GUID as an object of the master. 

Both the master and the slave must be using `memory mode = dbengine`.This will require a new streaming protocol version.

This will help with replication of metrics. Imagine having multiple netdata master servers for high availability. The slaves are streaming their metrics and pass on the GUIDs. This way, when a master becomes unavailable and a new one takes its place all metrics get to retain the same GUIDs and achieve metric history continuity and synchronization.

##### Analysis

```
GUID UUID

```

Precedes an object definition (`HOST`, `CHART` or `DIMENSION`)`. 

example:

```
GUID 10000000:1000:1000:1000:00000000001
CHART netdata.runtime_example '' 'Execution time' 'ms' 'python.d' netdata.pythond_runtime line 145000 1
```

##### Steps

- Implement functions to parse the new syntax.
- Ensure backwards compatibility with the streaming protocol.
- Cover cases:
   1. Master has old protocol - Slave has new protocol
   2. Slave has old protocol - Master has new protocol
   3. Master and Slave have same `GUID` protocol version
   4. Master is `dbengine` - Slave is not `dbengine`
       - Don't send
   5. Master is not `dbengine` - Slave is `dbengine`
       - Send, Ignore on receive
   6. Master and Slave are both `dbengine`
       - Send, use on receive

stelfrag:
Fixes #8918 
##### Summary
Adds the GUID keyword to the streaming protocol. It will be enabled only if DBENGINE
is compiled and in use. 

Required on the sender side to send GUID
- DBENGINE support compiled (`ENABLE_DBENGINE` flag)
- Minimum negotiated streaming protocol
- `rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE`

The same checks are performed on the receiver side

If DBENGINE support is not compiled the sender will not have code in place to send GUIDs. On the receiver side the GUID parsing code will be a no-op 

##### Test Plan
- Compile a master / slave with streaming protocol version 3
- Clear the metadata logs for the master and slave
- Start the master server
- Start the slave server

The GUIDs generated on the slave are streamed to the master and : 

- Used for slave's charts and dimensions
- Stored in the metadata logfile

Example 

##### Metadata on slave
```
GUID a42aa834-d591-4a52-aafd-9b633a3c074d
CHART "system.idlejitter" "" "CPU Idle Jitter" "microseconds lost/s" "idlejitter" "system.idlejitter" "area" 800 1 "   " "idlejitter.plugin" ""
CONTEXT a42aa834-d591-4a52-aafd-9b633a3c074d
GUID 635fd0a1-cc0d-02d3-4c50-6457e86c15cb
DIMENSION "min" "min" "absolute" 1 1 "  "
CONTEXT a42aa834-d591-4a52-aafd-9b633a3c074d
GUID 935b744c-b804-90a0-5756-2375f3d4bce3
DIMENSION "max" "max" "absolute" 1 1 "  "
CONTEXT a42aa834-d591-4a52-aafd-9b633a3c074d
GUID 07818b2c-8181-3fbd-b38b-d2d3de1d0034
DIMENSION "average" "average" "absolute" 1 1 "  "
```

##### Metadata on master
```
GUID a42aa834-d591-4a52-aafd-9b633a3c074d
CHART "system.idlejitter" "" "CPU Idle Jitter" "microseconds lost/s" "idlejitter" "system.idlejitter" "area" 800 1 "   " "idlejitter.plugin" ""
CONTEXT a42aa834-d591-4a52-aafd-9b633a3c074d
GUID 635fd0a1-cc0d-02d3-4c50-6457e86c15cb
DIMENSION "min" "min" "absolute" 1 1 "  "
CONTEXT a42aa834-d591-4a52-aafd-9b633a3c074d
GUID 935b744c-b804-90a0-5756-2375f3d4bce3
DIMENSION "max" "max" "absolute" 1 1 "  "
CONTEXT a42aa834-d591-4a52-aafd-9b633a3c074d
GUID 07818b2c-8181-3fbd-b38b-d2d3de1d0034
DIMENSION "average" "average" "absolute" 1 1 "  "

```


collectors/plugins.d/pluginsd_parser.c
```diff
@@ -561,11 +578,22 @@ PARSER_RC pluginsd_overwrite(char **words, void *user, PLUGINSD_ACTION  *plugins
...
     char *uuid_str = words[1];
     uuid_t uuid;
 
+    RRDHOST *host = ((PARSER_USER_OBJECT *) user)->host;
+
+    if (host->rrd_memory_mode != RRD_MEMORY_MODE_DBENGINE) {
+        return PARSER_RC_OK;
+    }
+
+    if (((PARSER_USER_OBJECT *) user)->cd->version < VERSION_GUID)
```
mfundul, line 594: I don't like how we have separate versions for streaming and the plugins.D language itself, but in practice we use the streaming version interchangeably in such a way that influences the language parsing itself.

At some point we should clearly define separate language version and streaming protocol versions.

collectors/plugins.d/pluginsd_parser.c
```diff
@@ -49,11 +49,16 @@ PARSER_RC pluginsd_chart_action(void *user, char *type, char *id, char *name, ch
...
+    uuid_t  *guid = &((PARSER_USER_OBJECT *)user)->guid;
 
-    st = rrdset_create(
-        host, type, id, name, family, context, title, units,
-        plugin, module, priority, update_every,
-        chart_type);
+    st = rrdset_create_custom(
+        host, type, id, name, family, context, title, units, plugin, module, priority, update_every, chart_type,
+        host->rrd_memory_mode, host->rrd_history_entries, 0,
+        uuid_is_null(*guid) ? NULL : guid);
```
mfundul, line 57: I know it's kind of implied by `callocz()` of the `PARSER_USER_OBJECT` but I think you should add an explicit initialization during parser startup such as `uuid_clear(user->guid)` in `streaming_parser()` definition to make it more obvious.

mfundul: Same would be needed for `pluginsd_process()`.

collectors/plugins.d/pluginsd_parser.c
```diff
@@ -49,11 +49,16 @@ PARSER_RC pluginsd_chart_action(void *user, char *type, char *id, char *name, ch
...
+
+#ifdef ENABLE_DBENGINE
+    uuid_clear(((PARSER_USER_OBJECT *) user)->guid);
```
mfundul, line 60: Since plugins.d is performance critical it may be worth it to not burn CPU/RAM cycles unconditionally at this point. Maybe add something like `if (uuid_was_null)`?

collectors/plugins.d/pluginsd_parser.c
```diff
@@ -124,7 +129,13 @@ PARSER_RC pluginsd_dimension_action(void *user, RRDSET *st, char *id, char *name
     UNUSED(user);
     UNUSED(algorithm);
 
-    RRDDIM *rd = rrddim_add(st, id, name, multiplier, divisor, algorithm_type);
+    uuid_t  *guid = &((PARSER_USER_OBJECT *)user)->guid;
+
+    RRDDIM *rd = rrddim_add_custom(
+        st, id, name, multiplier, divisor, algorithm_type, st->rrd_memory_mode, 0, uuid_is_null(*guid) ? NULL : guid);
+#ifdef ENABLE_DBENGINE
+    uuid_clear(((PARSER_USER_OBJECT *) user)->guid);
```
mfundul, line 137: Same as above, maybe make this conditional?

collectors/plugins.d/pluginsd_parser.c
stelfrag: I will add in streaming parser -- in pluginsd_process the GUID is not used

database/engine/metadata_log/logfile.c
```diff
@@ -552,7 +552,7 @@ static int scan_metalog_files(struct metalog_instance *ctx)
         .obsolete = 0,
         .started_t = INVALID_TIME,
         .next = NULL,
-        .version = 0,
+        .version = 9999,
```
mfundul, line 555: Shouldn't `.version = ` be also set in `receiver.c`?

collectors/plugins.d/pluginsd_parser.c
stelfrag: Good idea.. I will see if uuid_is_null is cheap

collectors/plugins.d/pluginsd_parser.c
stelfrag: Ah I see what you mean for the other keywords

collectors/plugins.d/pluginsd_parser.c
amoss: I don't see any need to use the streaming version for anything other than defining the version of the streaming protocol.

collectors/plugins.d/pluginsd_parser.c
```diff
@@ -51,13 +51,15 @@ PARSER_RC pluginsd_chart_action(void *user, char *type, char *id, char *name, ch
...
+    int is_guid_null = uuid_is_null(*guid);
     st = rrdset_create_custom(
         host, type, id, name, family, context, title, units, plugin, module, priority, update_every, chart_type,
         host->rrd_memory_mode, host->rrd_history_entries, 0,
-        uuid_is_null(*guid) ? NULL : guid);
+        is_guid_null ? NULL : guid);
 
 #ifdef ENABLE_DBENGINE
-    uuid_clear(((PARSER_USER_OBJECT *) user)->guid);
+    if (likely(!is_guid_null))
```
mfundul, line 61: In the case of collectors via plugins.D (and not streaming) the `likely` is going to hit the misprediction penalty 100% of the time. I think we should remove any hints in this check and let the CPU train itself depending on whether it is log replay, streaming or collectors.

database/engine/metadata_log/logfile.c
stelfrag: It is set right before calling the streaming parser

mfundul: Whoops, yes. I just saw `.version = 0` and didn't realize it's being overwritten.

mfundul:
Ok, I think there is a fundamental issue with this task which we haven't thought through.
Let's say we have a new master and an old slave:
1. The master is going to randomly generate GUIDs for the streamed data.
2. We upgrade the slave to also randomly generate GUIDs for its own data.
3. We enable the streaming protocol support for GUIDs.

Now we have different GUIDs streamed from the slave to what the master previously randomly generated. This will cause all preexisting data in the master to disappear.

mfundul:
> 
> 
> Why are doing part (1) - randomly generating a GUID for the streamed data? The receiver populates the metadata on the first connection from the info on the stream, and this will include the GUID that was chosen by the sender.

The sender can be of an older version.

mfundul:
> 
> 
> Why are doing part (1) - randomly generating a GUID for the streamed data? The receiver populates the metadata on the first connection from the info on the stream, and this will include the GUID that was chosen by the sender.

Misread that. So are you proposing that the master should not use metadata log whatsoever when the slave is of older version?

mfundul:
Btw, up until this point, the metadata log is always being used.

mfundul:
E.g. a master being streamed data by an old slave can still show archived dimensions and charts normally. This is the same use case as having a slave use a different memory mode.

The problematic situation is when the slave uses dbengine with old version.

mfundul:
We have a bug in the original metadata log generation code where the `HOST` keyword is missed if it's receiving the host instead of it being the localhost.

amoss:
So the sender is running an old build and does not assign GUIDs at all. It is being treated like any other collector - the GUID is being randomly assigned when the stream is received for the first time and stored in the metadata log.

We are then replacing this with a new sender that does pick GUIDs and transmit them over the connection. It is a new sender. I think it is fine to break the historical link with the old dataset. @cakrit / @ktsaou what do you think about this case?

PR #9214, opened 2020-05-28
Gap-detection and slew

amoss:
### Summary

This PR replaces PR9170 completely.
Closes #8921 
Closes #9063 
Fixes #8825

This PR adds (inactive) code to calculate clock-slew and gap-size when slaves reconnect to a master. This code will be activated in a later PR to fill the gaps in the charts.

The streaming component has been heavily restructured to fix several current bugs, and improve the concurrency protection.

* Incorrect handling of return codes
* Uses circular buffer for transmission to avoid starvation on high-latency link
* Remove the warnings from the time defines that were shown for every file on the mac.
* Restructured the sender and receiver, collected their state outside of the host into separate structures that can be locked.
* Made the locking around the sender buffer consistent.
* Fix a case where the stream version was not negotiated correctly.
* Adds bi-directional comms on the streaming and code to receive commands in the sender.
* Cleans up the SSL / raw-socket code and moves it out of the plugins.d parser
* Tracks liveness on the slave connections (seconds since last message received).
* On connection collisions it chooses the new/old socket depending on liveness.

##### Component Name
streaming

##### Test Plan

* Test high- and low- bandwidth settings.
* Used local containers and containers on a remote machine
* Running master on bare-metal

Test A:
1. Connect slaves (observe that no gap is detected on first connection)
2. Pause the containers and kill the socket using `ss -K dst 192.168.1.x`
3. Restart the containers 

Test B:
1. Connect slaves (observe that no gap is detected on first connection)
2. Pull the network cable
3. Reconnect

Test C:
1. Setup multiple containers with the same GUID and mix tests (1) and (2) to check the new collision detection code.
 
##### Additional Information

* We should coordinate about this on Monday, it probably needs to be merged before it blocks any work in the persistent metadata task, but we need to be reasonably sure that the new code is stable.

PR #9262, opened 2020-06-03
Fix frontmatter in circular_buffer README

joelhans:
##### Summary

There's a missing dash `-` in the frontmatter for the `libnetdata/circular_buffer/README.md` document introduced in #9214, which is causing build failures on Learn. This PR adds that dash.

I should have caught this in the initial PR. That's my mistake.

##### Component Name

libnetdata
docs

### DIFF
--- a/libnetdata/circular_buffer/README.md
+++ b/libnetdata/circular_buffer/README.md
@@ -1,5 +1,5 @@
 <!--
---
+---
 title: "circular_buffer"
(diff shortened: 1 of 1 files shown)


amoss:
> I should have caught this in the initial PR. That's my mistake.

@joelhans Sorry about that - I should have pinged you when the PR touched the documentation but I didn't realise - I'll do it next time.

PR #9263, opened 2020-06-03
Fixed handling of OpenSSL on CentOS/RHEL by bundling a static copy and selecting a configuration directory at install time.

Ferroin:
##### Summary

This adds code to bundle a static locally built copy of OpenSSL into our static builds instead of using the Alpine provided copy. It fixes two cases where our usage of OpenSSL currently fails:

* On CentOS, RHEL, OEL, Amazon Linux, and their derivatives, the OpenSSL configuration directory is `/etc/pki/tls` instead of the normal `/etc/ssl`. Any usage of TLS in our static builds currently fails there because it can't find that directory.
* TLS usage fails similarly on systems that do not have OpenSSL at all for the same reason.

To fix this, the newly bundled copy of OpenSSL is built to use `/opt/netdata/etc/ssl` as it's configuration directory. This directory is a symlink created at install time pointing to one of the following locations (in order of precedence):

* `/etc/pki/tls` (for CentOS, RHEL, OEL, AL, and similar).
* `/etc/ssl` (for sane distros that just use the default path).
* `/opt/netdata/share/ssl` (for systems that don't have OpenSSL, this contains a copy of the config and certificates from the build environment).

##### Component Name

area/packaging

##### Test Plan

- [x] Verify locally that it works on CentOS 8
- [x] Verify locally that it works on sane systems that use the default `/etc/ssl` path.
- [x] Verify locally that it works on a clean install without OpenSSL or CA certificates (Alpine and Debian Docker images fit this).
- [x] Verify that it passes CI.

##### Additional Information

TODO:

- [x] Handle the case of there being no `/etc/ssl` or `/etc/pki/tls` on the system.
- [x] Ensure all components (not just Netdata) use the bundled copy of OpenSSL.

NOTE:

This _DOES NOT_ modify anything about any type of build other than our static builds. Docker images, binary packages, and our regular installer are all unaffected.


packaging/makeself/install-or-update.sh
```diff
@@ -248,6 +248,24 @@ fi
...
+else
+  if [ -d /etc/pki/tls ] ; then
+    echo "Using /etc/pki/tls for TLS configuration and certificates"
+    ln -sf /etc/pki/tls /opt/netdata/etc/ssl
+  elif [ -d /etc/ssl ] ; then
+    echo "Using /etc/ssl for TLS configuration and certificates"
+    ln -sf /etc/ssl /opt/netdata/etc/ssl
+  else
+    echo "Using bundled TLS configuration and certificates"
+    ln -sf /opt/netdata/share/ssl /opt/netdata/etc/ssl
```
prologic, line 263: I am still not convinced we should be symlinking anything from the host system here. Someone explain to me again one more time with very real tangile use-cases why we'd care about the host's CA certs?

Ferroin: Standard SSL/TLS intercepting captive portals/transparent proxies on enterprise networks.

If we're running in such a network, connection to Netdata Cloud will fail if we don't have the local CA for the intercepting system. Preferentially using the system CA bundle means that admins can just go through the normal setup for such a scenario (namely, inject their own CA into the system trust store) and Netdata will just work.

prologic: Okay I actually buy that argument. Thanks!

prologic:
LGTM based on code review.

packaging/makeself/jobs/20-openssl.install.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# shellcheck source=packaging/makeself/functions.sh
+. "$(dirname "${0}")/../functions.sh" "${@}" || exit 1
+
+version="$(cat "$(dirname "${0}")/../openssl.version")"
+
+export LDFLAGS='-static'
+export PKG_CONFIG="pkg-config --static"
+
+run git clone git://git.openssl.org/openssl.git "${NETDATA_MAKESELF_PATH}/tmp/openssl"
```
underhood, line 12: Do this in one step and save some bandwith with `--single-branch`?
`git clone --branch <tag_name> <repo_url> --signle-branch`

PR #9268, opened 2020-06-04
Fix Coverity defects 359164, 359165 and 358989.

amoss:
##### Summary
Fix Coverity defects 359164, 359165 and 358989.

##### Component Name
cli
streaming

##### Test Plan

The thread deadlocks have not been noticed in the three years that they've been in the code. There is not a test-case, please read the code and analyze if they fix the problem.

##### Additional Information

CIDs 359164-5 were caused by `done_push` acquiring the host lock while holding the chart lock (to control the spawning of the sender thread). Replaced this with the new fine-grained mutex that guards the sender buffer instead.

### DIFF
--- a/streaming/rrdpush.c
+++ b/streaming/rrdpush.c
@@ -368,12 +368,10 @@ void rrdpush_send_labels(RRDHOST *host) {
 // rrdpush sender thread
 
 // Either the receiver lost the connection or the host is being destroyed.
-// Don't lock the sender buffer - doesn't affect consistency in either case.
-// TODO-GAPS During the host destruction sequence we should make sure the disconnect happens early enough to lock
-//           out collectors hitting the sender. Locking the mutex means there may be waiting threads when we free.
+// The sender mutex guards thread creation, any spurious data is wiped on reconnection.
 void rrdpush_sender_thread_stop(RRDHOST *host) {
-    rrdhost_wrlock(host);
 
+    netdata_mutex_lock(&host->sender->mutex);
     netdata_thread_t thr = 0;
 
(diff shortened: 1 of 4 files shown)


streaming/rrdpush.c
```diff
@@ -410,7 +410,7 @@ void log_stream_connection(const char *client_ip, const char *client_port, const
 
 
 static void rrdpush_sender_thread_spawn(RRDHOST *host) {
-    rrdhost_wrlock(host);
+    netdata_mutex_lock(&host->sender->mutex);
```
mfundul, line 413: You should do that for `rrdpush_sender_thread_stop()` as well.

mfundul:
Another thing that needs to change is `rrdpush_sender_thread_cleanup_callback()` in `sender.c`.

mfundul:
`rrdpush_sender_send_this_host_variable_now()` uses `host->rrdpush_sender_spawn` without protection.

streaming/rrdpush.c
amoss: Yeah.... I've changed the mutex for the sender mutex, but as my helpful comment shows there could be problems in doing that.

amoss:
> `rrdpush_sender_send_this_host_variable_now()` in `sender.c` uses `host->rrdpush_sender_spawn` without protection.

I think that one is fine - the buffer access inside the scope is synchronized and any spurious data in the buffer is wiped on connection (in case the flag is desynchronized).

amoss:
@mfundul I removed the others. The only access that is left inside `sender.c` is from the chart flush, which happens on the sender thread where no chart lock is held.

PR #9091, opened 2020-05-18
Add infiniband monitoring to collector proc.plugin

Saruspete:
##### Summary
Implement infiniband monitoring. Fixes #2620 

Infiniband is a fabric heavily used in HPC and supercomputers. It's designed for short range (local to  datacenter), provides RDMA, low latency communication and is handled directly by MPI libraries.

It often is coupled with "IPoIB" (IP over InfiniBand) which adds the IP connectivity over Infiniband messaging and is also seen as a simple network card. However, the MPI libraries doesn't use it, and talk directly to IB cards. 

From the drivers, there's no single file that provides details and summary over all IB interfaces, so we have to parse `/sys/class/infiniband/<verb>/ports/<portid>/counters/` path to get the counters provided by driver.
Each statistics are per-port.

This prevents to use the `procfile` helpers. As there is a lot of single statistics files (21 for `counters`, and more than 120 in `hw_counters` for OmniPath), I used macros to generate the repetitive elements over `read_single_number_file`

The `counters` are grouped in 3 categories (graphs): `bytes` `packets` and `errors`.

The `hw_counters` needs to be done if possible, as it seems very interface-specific. More feedback will be needed to have more output.

##### Component Name
collectors/proc.plugin

##### Test Plan

Emulates infiniband hw:

Change default path with netdata.conf to use a fake folder:
```
[plugin:proc:/sys/class/infiniband]
  dirname to monitor = /opt/netdata/sys/class/infiniband
```

Create a fake tree and populate it
```bash
typeset ibpath="/opt/netdata/sys/class/infiniband/test1_1/ports/99/counters/"
mkdir -p "$ibpath"
typeset -A counters
# Generate fake traffic
while sleep 1; do
  for counter in {port,multicast,unicast}_{rcv,xmit}_packets port_{rcv,xmit}_data port_{rcv_{,constraint_,remote_physical_,switch_relay_}errors,xmit_{discards,wait,constraint_errors}} link_{downed,error_recovery}  local_link_integrity_errors excessive_buffer_overrun_errors VL15_dropped symbol_error; do
    counters[$counter]=$((${counters[$counter]} + $RANDOM % 100))
    echo ${counters[$counter]} > "$ibpath/$counter"
  done
done
```

Create a state file
``` bash
echo "4: ACTIVE" > /opt/netdata/sys/class/infiniband/test1_1/state
```

That must create 3 graphs, under family "infiniband" / "test1_1-99"

PR #9279, opened 2020-06-05
Add zfsdstat plugin

roedie:
##### Summary
Add zfsdstat plugin to create per dataset charts for ZFS on Linux datasets

##### Component Name
collector

##### Test Plan
Added the plugin to my own install to see the results.

##### Additional Information
This is a simple bash plugin which detects all datasets within a system
running ZFS on Linux and creates charts for them.

These charts make it easier to identify very busy datasets.

The following charts are created:
- ZFS Dataset IOPS
- ZFS Dataset Bandwidth
- ZFS Dataset Unlinks

Screenshot 2020-06-05 at 13 31 06


ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5646040&bid=18489245)

ilyam8:
Hi @roedie 

> reads dataset statistics (IOPS Bandwidth, unlinks) from `/proc/spl/kstat/zfs/`

Not sure we want this in bash, all collectors that read from `/proc` are imlpemented in `c`.

roedie:
Hi @ilyam8 

I do understand that. But since I do not master C and I thought 'a plugin is better than no plugin' I created this pull.

I wouldn't enable it by default but leave the choice by the user.

roedie:
@ilyam8 If, for reasons I do understand, this plugin is not accepted, is there another place where netdata plugins are available? Something like https://exchange.icinga.com/.

I can imagine other people write plugins which might be useful but should not be in the netdata base distribution.

ilyam8:
> I thought 'a plugin is better than no plugin' I created this pull.

This is true, the problem side of that approach is - it is easy to add something, but it is hard to remove it later. If there are reasons not to add from the beginning i am personally wouldnt add it, but it is not my decision in that particular case.

We can do same as we did with some of py modules

> https://github.com/netdata/netdata/pull/4635

---

@vlvkobal @dim08 what do you think?

@Saruspete hey, perfaphs you would like to rewrite it in `c`?

---

>  is there another place where netdata plugins are available? 

> https://learn.netdata.cloud/docs/agent/collectors/collectors#third-party-plugins

Saruspete:
@ilyam8 thanks for the idea ! I'm a bit short on time atm, but it may be doable.

Couldn't it be integrated directly within the [current ZFS proc module](https://github.com/netdata/netdata/blob/master/collectors/proc.plugin/zfs_common.c) that takes its [entries from /proc/spl/kstat](https://github.com/netdata/netdata/blob/master/collectors/proc.plugin/proc_spl_kstat_zfs.c) too ? 

The only tricky part is from the dataset discovery from folders, but its seems similar to my current [PR for infiniband port discovery](https://github.com/netdata/netdata/pull/9091/files#diff-b016cb1ed3cfc83e793092f14536ddb1R207-R219).

ilyam8:
Sounds good to me, it makes sense to add this to the zfs proc module.

ktsaou:
@ilyam8 @vlvkobal  please update and merge or close.

vlvkobal:
I'm closing this because we are not planning to add or officially support new bash plugins. @roedie please make a PR to https://github.com/netdata/community.

Issue #9291, opened 2020-06-07
Segfault after upgrade to 1.22.1218


davidmehren:
##### Bug report summary
Netdata segfaults a few seconds after starting. 

##### OS / Environment
A Ubuntu 18.04 LXC container on Proxmox 6/Debian 10.
Kernel: `Linux legion 5.4.41-1-pve #1 SMP PVE 5.4.41-1 (Fri, 15 May 2020 15:06:08 +0200) x86_64 GNU/Linux`

##### Netdata version
The problems started after an apt upgrade to `1.22.1218`. I can also reproduce the problem with `v1.22.1-221-gff326cec` with and without valgrind.
 
##### Steps To Reproduce
1. Start Netdata: `netdata -D`
2. After a few seconds, netdata crashes with `Segmentation fault`.

I built the latest master with debug info and ran in valgrind:

```
valgrind $(which netdata) -D                                                                                                                                  ==19375== Memcheck, a memory error detector                                                                                                                                           ==19375== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.                                                                                                             
==19375== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info                                                                                                          
==19375== Command: /usr/sbin/netdata -D                                                                                                                                               
==19375==                                                                                                                                                                             
2020-06-07 13:18:06: netdata INFO  : MAIN : CONFIG: cannot load cloud config '/var/lib/netdata/cloud.d/cloud.conf'. Running with internal defaults.                                   
2020-06-07 13:18:07: netdata INFO  : MAIN : SIGNAL: Not enabling reaper                                                                                                               
==19375== Thread 2:                                                                                                                                                                   
==19375== Syscall param writev(vector[...]) points to uninitialised byte(s)                                                                                                           
==19375==    at 0x58C96E7: writev (writev.c:26)                                                                                                                                       ==19375==    by 0x637BA7A: ??? (in /usr/lib/x86_64-linux-gnu/libuv.so.1.0.0)                                                                                                          ==19375==    by 0x637CF1B: uv_write2 (in /usr/lib/x86_64-linux-gnu/libuv.so.1.0.0)                                                                                                    ==19375==    by 0x1BCFD3: spawn_process_cmd (spawn_client.c:160)                                                                                                                      
==19375==    by 0x1BCFD3: spawn_client (spawn_client.c:215)                                                                                                                           
[...]

PR #9297, opened 2020-06-08
Fix crash in #9291

amoss:
##### Summary
Fixes the crash part of #9291 - Did not account for the path where rrdhost_find_or_create fails and terminates the thread before setting rpt->host to a valid pointer.

##### Component Name
streaming

##### Test Plan
Not sure how to replicate the database failure, but the code change is simple enough to read.

### DIFF
--- a/streaming/receiver.c
+++ b/streaming/receiver.c
@@ -11,10 +11,12 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
         struct receiver_state *rpt = (struct receiver_state *) ptr;
 
         // Make sure that we detach this thread and don't kill a freshly arriving receiver
-        netdata_mutex_lock(&rpt->host->receiver_lock);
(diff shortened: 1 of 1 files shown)


ilyam8:
from #9291

> Alright, I think I got it: We have a host called "null" (german for "zero"). When I disable Netdata on it, the master does not crash, if I start it again, it crashes.

Did we test it? Is it really the name? 😄

amoss:
> from [#9291 (comment)](https://github.com/netdata/netdata/issues/9291#issuecomment-640209632)
> 
> > Alright, I think I got it: We have a host called "null" (german for "zero"). When I disable Netdata on it, the master does not crash, if I start it again, it crashes.
> 
> Did we test it? Is it really the name? 😄

This doesn’t fix the name part. @mfundul and @stelfrag need to check the dB code but this will prevent the crash if the find_or_create fails because of an error condition.

PR #9308, opened 2020-06-10
Correct the repo in the docs for CentOS 8.

Ferroin:
##### Summary

GetPageSpeed requires payment to actually use their repos, so we shouldn't be telling users to use them.

This updates the docs to point users at the repo we are already using in our installer for pulling in required dependencies.

##### Component Name

area/docs

##### Test Plan

n/a

##### Additional Information

Credit to @vlvkobal for noticing this.


dvershinin:
@Ferroin The `libmnl-devel` is in the PowerTools repo for CentOS 8.
The RHEL equivalent is Code Ready Builder repo.

image

PR #9310, opened 2020-06-10
Added dummy `--enable-ebpf` flag to avoid breaking updates.

Ferroin:
##### Summary

When we made eBPF default enabled, we removed the `--enable-ebpf` flag that had been used to enable it. This broke updates of existing installs that had explicitly enabled it, so re-add the flag as a functional no-op.

##### Component Name

area/packaging

##### Test Plan

Tested locally in Docker, passes CI.


prologic:
At some point we should remove it.

Ferroin:
> At some point we should remove it.

We kind of can't unless we want to add code to scrub the flags passed into the install script during the update process.

prologic:
> > At some point we should remove it.
> 
> We kind of can't unless we want to add code to scrub the flags passed into the install script during the update process.

That's what I implied. It would have even been a valid fix here rather than or in addition to adding a dummy flag.

PR #9318, opened 2020-06-11
Fix check for remote write header in unit tests

vlvkobal:
##### Summary
A Unit test was failing after HTTP header formatting was changed in #9302.

##### Component Name
exporting engine

### DIFF
--- a/exporting/tests/test_exporting_engine.c
+++ b/exporting/tests/test_exporting_engine.c
@@ -1194,9 +1194,10 @@ static void test_prometheus_remote_write_send_header(void **state)
         "POST /receive HTTP/1.1\r\n"
         "Host: test-host\r\n"
         "Accept: */*\r\n"
+        "X-Prometheus-Remote-Write-Version: 0.1.0\r\n"
         "Content-Length: 11\r\n"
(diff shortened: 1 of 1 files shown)


exporting/tests/test_exporting_engine.c
```diff
@@ -1194,9 +1194,10 @@ static void test_prometheus_remote_write_send_header(void **state)
         "POST /receive HTTP/1.1\r\n"
         "Host: test-host\r\n"
         "Accept: */*\r\n"
+        "X-Prometheus-Remote-Write-Version: 0.1.0\r\n"
         "Content-Length: 11\r\n"
         "Content-Type: application/x-www-form-urlencoded\r\n\r\n");
-    expect_value(__wrap_send, len, 125);
+    expect_value(__wrap_send, len, 167);
```
thiagoftsm, line 1200: Only to confirm, both the current string and the previous string do not have the correct length, the previous had a difference of 13 bytes and now we have difference of 15 values according `kate`:

image6
 
Is the expected?

vlvkobal: It's because you are counting in escape characters as well.

PR #9323, opened 2020-06-11
Change streaming terminology to parent-child in the code

amoss:
##### Summary

Change all uses of "master/slave" terminology to "parent-child".

##### Test Plan

No functional differences in the code.

##### Additional Information

There are many mentions of master that I have left in place:
* References to external systems (e.g. the MySQL collector and the MacOS firmware)
* The use of `webmaster`
* Anything in URLs (e.g. references to git branches)


database/rrdhost.c
```diff
@@ -838,7 +838,7 @@ struct label *load_auto_labels()
             add_label_to_list(label_list, "_virt_detection", localhost->system_info->virt_detection, LABEL_SOURCE_AUTO);
 
     label_list = add_label_to_list(
-        label_list, "_is_master", (localhost->next || configured_as_master()) ? "true" : "false", LABEL_SOURCE_AUTO);
+        label_list, "_is_parent", (localhost->next || configured_as_parent()) ? "true" : "false", LABEL_SOURCE_AUTO);
```
mfundul, line 841: What's the impact of this to compatibility?

amoss: It affects google-analytics but they are converted to indices before hitting the endpoint and nobody else uses it.

mfundul: I found a doc here: https://github.com/netdata/netdata/blob/master/docs/guides/using-host-labels.md#host-labels-in-streaming
so maybe users are already using this.

thiagoftsm: I agree with @mfundul , this looks like a compatibility break? The rest of the changes is right.

ilyam8: No worries, nobody will open an issue to return masters and slaves in 2020 😉 🤣

amoss: I've discussed it with @cakrit and agreed that the chances of people using it are minor and we are changing the terminology.

collectors/plugins.d/pluginsd_parser.c
```diff
@@ -306,7 +306,7 @@ PARSER_RC pluginsd_chart(char **words, void *user, PLUGINSD_ACTION  *plugins_act
 
     // parse the name, and make sure it does not include 'type.'
     if (unlikely(name && *name)) {
-        // when data are coming from slaves
+        // when data are streaming from child nodes
```
cosmix, line 309:
```suggestion
        // when data are streamed from child nodes
```

daemon/main.c
```diff
@@ -36,7 +36,7 @@ void netdata_cleanup_and_exit(int ret) {
         // exit cleanly
 
         // stop everything
-        info("EXIT: stopping master threads...");
+        info("EXIT: stopping main threads...");
```
cosmix, line 39: Typically there's one 'main' thread and this log message can be misinterpreted/confusing.

streaming/stream.conf
```diff
@@ -21,20 +21,20 @@
     #
     # If many are given, the first available will get the metrics.
     #
-    # PROTOCOL  = tcp, udp, or unix (only tcp and unix are supported by masters)
+    # PROTOCOL  = tcp, udp, or unix (only tcp and unix are supported by parents)
```
cosmix, line 24:
```suggestion
    # PROTOCOL  = tcp, udp, or unix (only tcp and unix are supported by parent nodes)
```

amoss:
> A few minor comments here, but there's more references to barbaric names elsewhere:
> 
>     1. streaming/stream.conf

Could you point to where - I've been through `stream.conf.in` and changed everything I could see.

amoss:
>     3\. collectors/proc.plugin/proc_diskstats.c

I've changed references to external entities - they have to be correct, and in those cases it is encoded in the path.

amoss:
> * web/gui/dashboard.js (cc @jacekkolasa)
> 
> * web/gui/src/dashboard.js/main.js (@jacekkolasa)

That can be a separate PR.

amoss:
> https://github.com/netdata/netdata/blob/parent-child/streaming/stream.conf#L144

Arghhhh! That one did not show up in a search - fixing...

mfundul:
It's hard to keep approving this, stop fixing things!

PR #9312, opened 2020-06-10
Change streaming terminology to parent/child in docs

joelhans:
##### Summary

This PR replaces the usage of master/slave with parent/child when referencing streaming configurations/setups in Netdata. This is actually a more precise way to describe the relationship between these nodes, and modernizes the language we use in general.

@amoss Is working on a parallel change to the actual code. His PR should be merged before this so that we don't point users toward configs that don't yet exist, such as host labels.

##### Component Name

area/docs

PR #9330, opened 2020-06-12
Change slave/master terminology to parent/child in dashboard

jacekkolasa:
related to:
[#9323 Change streaming terminology to parent-child in the code](https://github.com/netdata/netdata/pull/9323)
[#9312 Change streaming terminology to parent/child in docs](https://github.com/netdata/netdata/pull/9312)

### DIFF
--- a/web/gui/dashboard.js
+++ b/web/gui/dashboard.js
@@ -5825,7 +5825,7 @@ NETDATA.globalSelectionSync = {
     state: null,
     dontSyncBefore: 0,
     last_t: 0,
-    slaves: [],
+    children: [],
     timeoutId: undefined,
 
     globalReset: function () {
@@ -5833,7 +5833,7 @@ NETDATA.globalSelectionSync = {
         this.state = null;
         this.dontSyncBefore = 0;
         this.last_t = 0;
-        this.slaves = [];
+        this.children = [];
         this.timeoutId = undefined;
     },
 
@@ -5875,14 +5875,14 @@ NETDATA.globalSelectionSync = {
         this.state = state;
         this.last_t = 0;
 
-        // find all slaves
+        // find all children
         let targets = NETDATA.intersectionObserver.targets();
-        this.slaves = [];
+        this.children = [];
         let len = targets.length;
         while (len--) {
             let st = targets[len];
             if (this.state !== st && st.globalSelectionSyncIsEligible()) {
-                this.slaves.push(st);
+                this.children.push(st);
             }
         }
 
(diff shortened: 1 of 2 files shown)


thiagoftsm:
@jacekkolasa unless I am missing something, we do not need to change `web/gui/dashboard.js`, this is automatically build.

amoss:
I'm confused - when I go to look at what will be changed in the repo, `dashboard.js` is not a file it is a directory`

jacekkolasa:
@thiagoftsm We should not update this file manually, but we should commit the changes. Please note that if you do changes inside dashboard.js folder and run the build, you'll see a change in dashboard.js file in git.

@amoss there is both directory and folder with the same name

amoss:
> @amoss there is both directory and folder with the same name

A directory and a folder are the same thing. I see a file with this name in your PR, but in the repo it is a directory / folder. How does this work?

amoss:
Ah ok - I see what is happening. The file does not have `src/` in the path.

amoss:
This PR changes all the references to `slave`, but what about all the references to `master` ?

jacekkolasa:
Of course i wanted to say "file". Anyway, those files/folders will be removed soon, i need to make sure they are not used in any edge case scenarios. Normally, dashboard.js file should be replaced with the one from dashboard.git bundle - but perhaps those files can be used if downloading of dashboard.tar.gz fails

jacekkolasa:
oh, you're right, but do you think it's mandatory? There are much more cases of `master` word than `slave`. For example we use `master` branch. I'm asking as a non-native speaker

jacekkolasa:
Sorry for misleading title, i copy/pasted similar phrase to mark it as a related PR. Anyway, maybe i can do that change in other PR? Changes would need to have more testing and i would prefer to do it some other time

amoss:
> oh, you're right, but do you think it's mandatory? There are much more cases of `master` word than `slave`. For example we use `master` branch. I'm asking as a non-native speaker

My opinion on this is that we should remove references to master/slave relationships from within the code and the docs. This is the harmful case - it is an offensive metaphor that we should not use in the tech industry in general.

But the word `master` has other  meanings that are not related to slavery. It also means the original that other sources are derived from. So where we have used `master` to refer to git branches I left it alone. I also left alone references to `webmaster` as we use it to build email addresses.

Other people may have differing opinions on this but I would suggest that we strip any usage that uses master in the context of master/slave and/or controlling relationships. I'm not sure where this leaves the references in the .js code - it's up to you to decide what they mean and what we should do with them.

jacekkolasa:
@amoss i will gladly replace the `master` references to have a clear situation. We probably don't need to rush so this PR can wait until i'll add remaining changes directly here.

ilyam8:
These have to be changed on [the dashboard repo](https://github.com/netdata/dashboard).

PR #9338, opened 2020-06-15
Fix coverity errors CID 359774 and CID 359773

mfundul:
##### Summary
Fix New Defects reported by Coverity Scan for netdata/netdata

##### Component Name
database 

##### Additional Information
New defect(s) Reported-by: Coverity Scan
Showing 2 of 2 defect(s)

** CID 359774:  Memory - illegal accesses  (USE_AFTER_FREE)
/database/rrdset.c: 1901 in rrdset_done()

*** CID 359774:  Memory - illegal accesses  (USE_AFTER_FREE)
/database/rrdset.c: 1901 in rrdset_done()
1895     
1896     #ifdef ENABLE_DBENGINE
1897                         if (rd->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
1898                             rrddim_flag_set(rd, RRDDIM_FLAG_ARCHIVED);
1899                             rrddim_flag_clear(rd, RRDDIM_FLAG_OBSOLETE);
1900                             /* only a collector can mark a chart as obsolete, so we must remove the reference */
>>>     CID 359774:  Memory - illegal accesses  (USE_AFTER_FREE)
>>>     Dereferencing freed pointer "rd->state".
1901                             uint8_t can_delete_metric = rd->state->collect_ops.finalize(rd);
1902                             if (can_delete_metric) {
1903                                 /* This metric has no data and no references */
1904                                 metalog_commit_delete_dimension(rd);
1905                             } else {
1906                                 /* Do not delete this dimension */

** CID 359773:  Error handling issues  (CHECKED_RETURN)
/database/rrdhost.c: 248 in rrdhost_create()

*** CID 359773:  Error handling issues  (CHECKED_RETURN)
/database/rrdhost.c: 248 in rrdhost_create()
242                     error("Host '%s': cannot create directory '%s'", host->hostname, host->varlib_dir);
243            }
244     
245         }
246         if (host->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
247     #ifdef ENABLE_DBENGINE
>>>     CID 359773:  Error handling issues  (CHECKED_RETURN)
>>>     Calling "uuid_parse" without checking return value (as is done elsewhere 6 out of 7 times).
248             uuid_parse(host->machine_guid, host->host_uuid);
249             host->objects_nr = 1;
250             host->compaction_id = 0;
251             char dbenginepath[FILENAME_MAX + 1];
252             int ret;
253

### DIFF
--- a/database/rrdset.c
+++ b/database/rrdset.c
@@ -1904,6 +1904,8 @@ void rrdset_done(RRDSET *st) {
                             metalog_commit_delete_dimension(rd);
                         } else {
                             /* Do not delete this dimension */
+                            last = rd;
+                            rd = rd->next;
                             continue;
                         }
                     }
--- a/database/rrdhost.c
+++ b/database/rrdhost.c
@@ -245,7 +245,9 @@ RRDHOST *rrdhost_create(const char *hostname,
     }
     if (host->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
 #ifdef ENABLE_DBENGINE
-        uuid_parse(host->machine_guid, host->host_uuid);
+        if (unlikely(-1 == uuid_parse(host->machine_guid, host->host_uuid))) {
+            error("Host machine GUID is not valid.");
+        }
         host->objects_nr = 1;
         host->compaction_id = 0;
         char dbenginepath[FILENAME_MAX + 1];


database/rrdset.c
```diff
@@ -1904,6 +1904,8 @@ void rrdset_done(RRDSET *st) {
                             metalog_commit_delete_dimension(rd);
                         } else {
                             /* Do not delete this dimension */
+                            last = rd;
```
amoss, line 1907: Why are we checking if dimensions can be deleted at all in the case of the dbengine? The deletion would occur at the dbrotation, this whole scan of the dimensions seems to be redundant now in the case of the dbengine.

mfundul: It used to be that `OBSOLETE` dimensions were deleted. Now for `dbengine` this changes to a new state, that is `OBSOLETE` will be converted to `ARCHIVED` and the collector state is torn down. Rotation does indeed delete them in a different code path.

amoss: Yes, so do we need to scan the dimensions here to look for possible deletions when using dbengine?

mfundul: It's not deletions, it's state changes from `OBSOLETE` to `ARCHIVED`. The condition is the same but instead of deleting we archive them. Otherwise the dimensions would remain forever at the `OBSOLETE` state which would be wrong.

amoss: Ok, this makes sense. Very general question - why do we need `#ifdef ENABLE_DBENGINE` around blocks such as this?

mfundul: `dbengine` still isn't a required component for netdata so we need to support binaries that are built without it.

amoss: Yes, but why do we need to protect this block with a `#ifdef` ? The `if` could be enough.

mfundul: Some pieces of code breaks compilation such as `metalog_commit_delete_dimension()`.

PR #9311, opened 2020-06-10
Override linker and include paths for static builds.

Ferroin:
##### Summary

This is a continuation from #9263 that makes it actually work correctly.

##### Component Name

area/packaging

##### Test Plan

More thoroughly locally verified than last time.

##### Additional Information

Fixes: #9206

PR #9343, opened 2020-06-16
Revert "Override linker and include paths for static builds."

Ferroin:
Reverts netdata/netdata#9311


Ferroin:
I've already gotten verbal approval from @prologic on this earlier today during the SRE meeting, and have thus far been unable to find a quick solution that will make things work properly everywhere, so I'm merging this now to be certain it goes out in the nightly build tonight.

prologic:
Thanks @Ferroin

PR #9349, opened 2020-06-16
Replace assert calls

mfundul:
##### Summary
Replace all assert() calls with the new fatal_assert() for proper logging.

##### Component Name
daemon


amoss:
The ternary operator is typed as an expression, but the macro call looks like a statement. Why did you do this way instead of the `do { .. } while(0)` wrapper?

mfundul:
> 
> 
> The ternary operator is typed as an expression, but the macro call looks like a statement. Why did you do this way instead of the `do { .. } while(0)` wrapper?

I combined the [glibC](https://code.woboq.org/userspace/glibc/assert/assert.h.html#94) and [musl libC](https://git.musl-libc.org/cgit/musl/tree/include/assert.h#n8) implementations.

PR #9354, opened 2020-06-17
Improved error handling and recovery during compaction and metadata log replay

### LINKED ISSUE #9336: Netdata start fails with "database/engine/rrdengineapi.c:80: rrdeng_metric_init: Assertion `0' failed."
##### Bug report summary
I have an issue similar to #9113 but I am not sure if it's really the same problem. Netdata has stopped yesterday and won't start again:

```
[...]
2020-06-14 06:30:47: netdata INFO  : MAIN : Metadata log "/var/cache/netdata/dbengine/metadatalog-00000-00001.mlf" loaded (size:254011).
2020-06-14 06:30:47: netdata INFO  : MAIN : Loading metadata log "/var/cache/netdata/dbengine/metadatalog-00000-00002.mlf".
2020-06-14 06:30:47: netdata ERROR : MAIN : FAILED to generate GUID for GET
netdata: database/engine/rrdengineapi.c:80: rrdeng_metric_init: Assertion `0' failed.
EOF found in spawn pipe.
Shutting down spawn server event loop.
Shutting down spawn server loop complete.
```

I can remember that Netdata showed issues with not being able to connect to the cloud or something similar last week, but it worked without issues.

##### OS / Environment
```
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 18.04.4 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="18.04.4 LTS (Bionic Beaver)"
/etc/os-release:ID=ubuntu
/etc/os-release:PRETTY_NAME="Ubuntu 18.04.4 LTS"
```

##### Netdata version

netdata v1.22.1-276-nightly

(Which is really weird, because I selected the stable channel when I installed Netdata?!)

##### Steps To Reproduce

I am not sure, as I don't know how this issue is linked to my specific configuration or not.

i already ran the Netdata installer again, but the issue persists.

##### Expected behavior

Netdata should run normally.

stelfrag:
Fixes #9336
##### Summary
Add additional checks 

1. Recover from possible missing chart GUID compaction records that can result to incomplete metadata log replay information
    - When a dimension compaction request is received  a check will ensure that it's chart is also *force compacted* if needed
1. Metric GUID generation to detects and recovers from collisions (caused by inconsistent metadata log information)
   - This can happen when a corrupted metadata log file already exists due to case 1
1. Log unsuccessful GUID context switch (also due to corrupted metadata log file due to case 1)

##### Component Name
database

### DIFF
--- a/database/engine/metadata_log/metadatalogapi.c
+++ b/database/engine/metadata_log/metadatalogapi.c
@@ -315,7 +315,8 @@ RRDDIM *metalog_get_dimension_from_uuid(struct metalog_instance *ctx, uuid_t *me
     ret = find_object_by_guid(chart_char_guid, chart_fullid, RRD_ID_LENGTH_MAX + 1);
     fatal_assert(GUID_TYPE_CHAR == ret);
     RRDSET *st = rrdset_find(host, chart_fullid);
-    fatal_assert(st);
+    if (!st)
+        return NULL;
 
     RRDDIM *rd = rrddim_find(st, id_str);
 
--- a/database/engine/metadata_log/metadatalogapi.c
+++ b/database/engine/metadata_log/metadatalogapi.c
@@ -273,16 +273,19 @@ RRDSET *metalog_get_chart_from_uuid(struct metalog_instance *ctx, uuid_t *chart_
     uuid_t *machine_guid, *chart_char_guid;
 
     ret = find_object_by_guid(chart_uuid, chart_object, 33);
-    fatal_assert(GUID_TYPE_CHART == ret);
+    if (unlikely(GUID_TYPE_CHART != ret))
+        return NULL;
 
     machine_guid = (uuid_t *)chart_object;
     RRDHOST *host = ctx->rrdeng_ctx->host;
-    fatal_assert(!uuid_compare(host->host_uuid, *machine_guid));
+    if (unlikely(uuid_compare(host->host_uuid, *machine_guid)))
+        error("Metadata host machine GUID does not match the one assosiated with the chart");
 
     chart_char_guid = (uuid_t *)(chart_object + 16);
 
     ret = find_object_by_guid(chart_char_guid, chart_fullid, RRD_ID_LENGTH_MAX + 1);
-    fatal_assert(GUID_TYPE_CHAR == ret);
+    if (unlikely(GUID_TYPE_CHAR != ret))
+        return NULL;
     RRDSET *st = rrdset_find(host, chart_fullid);
 
     return st;
--- a/database/engine/metadata_log/metadatalogapi.c
+++ b/database/engine/metadata_log/metadatalogapi.c
@@ -278,8 +278,10 @@ RRDSET *metalog_get_chart_from_uuid(struct metalog_instance *ctx, uuid_t *chart_
 
     machine_guid = (uuid_t *)chart_object;
     RRDHOST *host = ctx->rrdeng_ctx->host;
-    if (unlikely(uuid_compare(host->host_uuid, *machine_guid)))
+    if (unlikely(uuid_compare(host->host_uuid, *machine_guid))) {
         error("Metadata host machine GUID does not match the one assosiated with the chart");
+        return NULL;
+    }
 
     chart_char_guid = (uuid_t *)(chart_object + 16);
 
@@ -303,20 +305,26 @@ RRDDIM *metalog_get_dimension_from_uuid(struct metalog_instance *ctx, uuid_t *me
 
     machine_guid = (uuid_t *)dim_object;
     RRDHOST *host = ctx->rrdeng_ctx->host;
-    fatal_assert(!uuid_compare(host->host_uuid, *machine_guid));
+    if (unlikely(uuid_compare(host->host_uuid, *machine_guid))) {
+        error("Metadata host machine GUID does not match the one assosiated with the dimension");
+        return NULL;
+    }
 
     chart_guid = (uuid_t *)(dim_object + 16);
     dim_char_guid = (uuid_t *)(dim_object + 16 + 16);
 
     ret = find_object_by_guid(dim_char_guid, id_str, sizeof(id_str));
-    fatal_assert(GUID_TYPE_CHAR == ret);
+    if (unlikely(GUID_TYPE_CHAR != ret))
+        return NULL;
 
     ret = find_object_by_guid(chart_guid, chart_object, sizeof(chart_object));
-    fatal_assert(GUID_TYPE_CHART == ret);
+    if (unlikely(GUID_TYPE_CHART != ret))
+        return NULL;
     chart_char_guid = (uuid_t *)(chart_object + 16);
 
     ret = find_object_by_guid(chart_char_guid, chart_fullid, RRD_ID_LENGTH_MAX + 1);
-    fatal_assert(GUID_TYPE_CHAR == ret);
+    if (unlikely(GUID_TYPE_CHAR != ret))
+        return NULL;
     RRDSET *st = rrdset_find(host, chart_fullid);
     if (!st)
         return NULL;
--- a/database/engine/metadata_log/metalogpluginsd.c
+++ b/database/engine/metadata_log/metalogpluginsd.c
@@ -105,15 +105,6 @@ PARSER_RC metalog_pluginsd_guid_action(void *user, uuid_t *uuid)
     return PARSER_RC_OK;
 }
 
-static inline void error_with_guid(uuid_t *uuid, char *reason)
-{
-    char  uuid_str[37];
-
-    uuid_unparse_lower(*uuid, uuid_str);
-    errno = 0;
-    error("%s (GUID = %s)", reason, uuid_str);
-}
-
 PARSER_RC metalog_pluginsd_context_action(void *user, uuid_t *uuid)
 {
     GUID_TYPE ret;
@@ -199,6 +190,8 @@ PARSER_RC metalog_pluginsd_tombstone_action(void *user, uuid_t *uuid)
                 rrdhost_wrlock(host);
                 rrdset_free(st);
                 rrdhost_unlock(host);
+            } else {
+                debug(D_METADATALOG, "Ignoring nonexistent chart metadata record.");
             }
             break;
         case GUID_TYPE_DIMENSION:
@@ -208,6 +201,9 @@ PARSER_RC metalog_pluginsd_tombstone_action(void *user, uuid_t *uuid)
                 rrdset_wrlock(st);
                 rrddim_free_custom(st, rd, 0);
                 rrdset_unlock(st);
+            }
+            else {
+                debug(D_METADATALOG, "Ignoring nonexistent dimension metadata record.");
             }
             break;
         case GUID_TYPE_HOST:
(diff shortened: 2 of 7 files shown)


database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -116,41 +116,45 @@ PARSER_RC metalog_pluginsd_context_action(void *user, uuid_t *uuid)
...
-
-            ret = find_object_by_guid(chart_char_guid, id_str, RRD_ID_LENGTH_MAX + 1);
-            fatal_assert(GUID_TYPE_CHAR == ret);
-            ((PARSER_USER_OBJECT *) user)->st = rrdset_find(host, id_str);
+        case GUID_TYPE_NOTFOUND:
+            uuid_unparse_lower(*uuid, object);
+            error("Failed to find valid context with GUID [%s]", object);
             break;
+        case GUID_TYPE_CHAR:
+            fatal_assert(0);
```
amoss, line 124: Remove this - log the error and continue.

database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -116,41 +116,45 @@ PARSER_RC metalog_pluginsd_context_action(void *user, uuid_t *uuid)
...
-            ret = find_object_by_guid(chart_char_guid, id_str, RRD_ID_LENGTH_MAX + 1);
-            fatal_assert(GUID_TYPE_CHAR == ret);
-            ((PARSER_USER_OBJECT *) user)->st = rrdset_find(host, id_str);
+            host = ctx->rrdeng_ctx->host;
+            switch (ret) {
+                case GUID_TYPE_CHART:
+                    chart_char_guid = (uuid_t *)(object + 16);
+
+                    ret = find_object_by_guid(chart_char_guid, id_str, RRD_ID_LENGTH_MAX + 1);
+                    fatal_assert(GUID_TYPE_CHAR == ret);
```
amoss, line 134: This is not appropriate - log the error and continue.

database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -116,41 +116,45 @@ PARSER_RC metalog_pluginsd_context_action(void *user, uuid_t *uuid)
...
+                    ((PARSER_USER_OBJECT *)user)->st = rrdset_find(host, id_str);
+                    break;
+                case GUID_TYPE_DIMENSION:
+                    chart_guid = (uuid_t *)(object + 16);
+
+                    ret = find_object_by_guid(chart_guid, chart_object, 33);
+                    fatal_assert(GUID_TYPE_CHART == ret);
```
amoss, line 141: Same here - no corruption in the log file should kill the agent. Ignore bad data, log errors and continue.

database/engine/rrdengineapi.c
```diff
@@ -74,10 +74,14 @@ void rrdeng_metric_init(RRDDIM *rd, uuid_t *dim_uuid)
...
-            freez(rd->state->metric_uuid);
-            rd->state->metric_uuid = NULL;
-            fatal_assert(0);
+            error("FAILED to reuse GUID for %s", rd->id);
+            if (unlikely(find_or_generate_guid(rd, rd->state->metric_uuid, GUID_TYPE_DIMENSION, 0))) {
+                errno = 0;
+                error("FAILED to generate GUID for %s", rd->id);
+                freez(rd->state->metric_uuid);
+                rd->state->metric_uuid = NULL;
+                fatal_assert(0);
```
amoss, line 83: Why kill the agent?

mfundul: This should be an impossible state. You can't have a dbengine dimension without a UUID, it has to fail here or it will fail later. The collectors expect to be able to use dimensions, there is no other failure mode here unfortunately.

database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -131,22 +139,33 @@ PARSER_RC metalog_pluginsd_context_action(void *user, uuid_t *uuid)
...
-                    ((PARSER_USER_OBJECT *)user)->st = rrdset_find(host, id_str);
+                    if (unlikely(GUID_TYPE_CHAR != ret)) {
+                        error_with_guid(uuid, "Failed to find valid chart name");
+                        ((PARSER_USER_OBJECT *)user)->st = NULL;
+                    }
+                    else
+                        ((PARSER_USER_OBJECT *)user)->st = rrdset_find(host, id_str);
                     break;
                 default:
-                    fatal_assert(0);
```
amoss: This should be logged as it indicates that the log is either corrupt or from a newer version than the agent.

database/engine/rrdengineapi.c
amoss: Just because it should be impossible is not a strong enough reason. If it was actually impossible then it would be valid to ignore it. If we have code to trap it thenit just seems to be unlikely, which is not the same. If something goes wrong - what is the least harmful way of dealing with it?

amoss: Why not mark the dimension as disabled / in error - and allow the agent to continue?

mfundul: When I detect state corruption I prefer to fail fast, before propagating the corruption to other components. This way it's easier to debug and you can do less damage to user data. That's my personal preference though.

amoss: In some systems that is desirable, especially if they are short-lived and interactive (e.g. command line tools). In others it can be quite harmful. We should aim for graceful degradation of service, and best-effort delivery where possible. In this specific case - what can cause the generation attempt to fail - and what does it tell us about the state of the system? i.e. does it make sense to continue, does it make sense to abort.

stelfrag: In this specific case  (i.e. after the collision has been detected) it will only fail if the newly generated GUID already exists in the map i.e. we have managed to generate a duplicate GUID.

In case of memory allocation errors with mallocz or JudyHSIns error,  it will never return because fatal() is called.

amoss: That seems insanely unlikely. Why not just repeat it if that happens instead of aborting the agent?

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -273,16 +273,19 @@ RRDSET *metalog_get_chart_from_uuid(struct metalog_instance *ctx, uuid_t *chart_
     uuid_t *machine_guid, *chart_char_guid;
 
     ret = find_object_by_guid(chart_uuid, chart_object, 33);
-    fatal_assert(GUID_TYPE_CHART == ret);
+    if (unlikely(GUID_TYPE_CHART != ret))
```
amoss, line 276: Log this

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -273,16 +273,19 @@ RRDSET *metalog_get_chart_from_uuid(struct metalog_instance *ctx, uuid_t *chart_
...
+        return NULL;
 
     machine_guid = (uuid_t *)chart_object;
     RRDHOST *host = ctx->rrdeng_ctx->host;
-    fatal_assert(!uuid_compare(host->host_uuid, *machine_guid));
+    if (unlikely(uuid_compare(host->host_uuid, *machine_guid)))
+        error("Metadata host machine GUID does not match the one assosiated with the chart");
```
amoss, line 282: Should this case continue instead of returning NULL?

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -273,16 +273,19 @@ RRDSET *metalog_get_chart_from_uuid(struct metalog_instance *ctx, uuid_t *chart_
...
 
     chart_char_guid = (uuid_t *)(chart_object + 16);
 
     ret = find_object_by_guid(chart_char_guid, chart_fullid, RRD_ID_LENGTH_MAX + 1);
-    fatal_assert(GUID_TYPE_CHAR == ret);
+    if (unlikely(GUID_TYPE_CHAR != ret))
```
amoss, line 287: Log this

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -315,7 +318,8 @@ RRDDIM *metalog_get_dimension_from_uuid(struct metalog_instance *ctx, uuid_t *me
     ret = find_object_by_guid(chart_char_guid, chart_fullid, RRD_ID_LENGTH_MAX + 1);
     fatal_assert(GUID_TYPE_CHAR == ret);
     RRDSET *st = rrdset_find(host, chart_fullid);
-    fatal_assert(st);
+    if (!st)
```
amoss, line 321: Log this

database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -116,41 +125,55 @@ PARSER_RC metalog_pluginsd_context_action(void *user, uuid_t *uuid)
...
-
-            ret = find_object_by_guid(chart_guid, chart_object, 33);
-            fatal_assert(GUID_TYPE_CHART == ret);
-            chart_char_guid = (uuid_t *)(chart_object + 16);
-
-            ret = find_object_by_guid(chart_char_guid, id_str, RRD_ID_LENGTH_MAX + 1);
-            fatal_assert(GUID_TYPE_CHAR == ret);
-            ((PARSER_USER_OBJECT *) user)->st = rrdset_find(host, id_str);
+            host = ctx->rrdeng_ctx->host;
+            switch (ret) {
```
amoss, line 138: Are you putting a `switch(ret)` inside a shared case in a `switch(ret)` to avoid repeating the `host = ` line above ??

database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -116,41 +125,55 @@ PARSER_RC metalog_pluginsd_context_action(void *user, uuid_t *uuid)
...
+                case GUID_TYPE_CHART:
+                    chart_char_guid = (uuid_t *)(object + 16);
+
+                    ret = find_object_by_guid(chart_char_guid, id_str, RRD_ID_LENGTH_MAX + 1);
+                    if (unlikely(GUID_TYPE_CHAR != ret))
+                        error_with_guid(uuid, "Failed to find valid chart name");
+                    else
+                        ((PARSER_USER_OBJECT *)user)->st = rrdset_find(host, id_str);
+                    break;
+                case GUID_TYPE_DIMENSION:
```
amoss, line 148: This seems wrong - if the GUID in the context command was a dimension then why treat it as a chart?

database/engine/metadata_log/metalogpluginsd.c
mfundul: This is a remnant from the persistent metadata feature branch. It used to have a lot of code about hosts but it got cleaned up and the monstrosity remained.

database/engine/metadata_log/metalogpluginsd.c
mfundul: We don't have a dimension state yet, in theory you could add dimension labels in this context and set `rd=dimension`, `st=dimension's chart`, `host=chart's host`.

Since the only state we have right now is `st`, we just set this.

database/engine/metadata_log/metadatalogapi.c
stelfrag: The caller will log this

PR #9355, opened 2020-06-17
adds support for multiple ACLK query processing threads

### LINKED ISSUE #9392: Multiple ACLK Query Threads
##### Feature idea summary

The ACLK Query thread should be multiple threads competing for work from the Queue. In the case of long queries, there are other threads that can pick up the work.

underhood:
##### Summary
- adds multiple (configurable) query threads working in parallel (competing for queries in queue)
- moves Query thread relateted funtions to separate files
- some cleanup on the go (removal of useless vars etc.)
- adds chart informing how many queries were processed by each of the threads (division of labor)
- adds chart showing how long it took us to gather data to satisfy cloud requests

Screenshot from 2020-06-23 19-39-09

Fixes #9392 

##### Component Name
ACLK

##### Test Plan
Test cloud still works.


underhood:
rebased on latest master after 5bc76f3b5bbba126c5bc651f4ada897c6ae05a94 merge

libnetdata/config/appconfig.c
```diff
@@ -340,6 +340,32 @@ long long appconfig_get_number(struct config *root, const char *section, const c
     return strtoll(s, NULL, 0);
 }
 
+long long appconfig_get_number_min_max(struct config *root, const char *section, const char *name, long long value, long long min, long long max)
```
stelfrag, line 343: It looks an overkill to add a new function min_max  just to check that we need at least 1 query thread up to a max defined

aclk/aclk_query.c
```diff
@@ -0,0 +1,581 @@
...
+
+#if ACLK_MAX_QUERY_THREADS > 10
+#error "You seem to have increased ACLK_MAX_QUERY_THREADS above 10. If you want to do that you need to update following thread name generation."
+#endif
+
+    char thread_name[strlen(ACLK_THREAD_NAME) + 3];
+    strcpy(thread_name, ACLK_THREAD_NAME);
+    thread_name[strlen(ACLK_THREAD_NAME)] = '_';
+    thread_name[strlen(ACLK_THREAD_NAME) + 2] = 0;
+
```
stelfrag, line 494: Why not just build it with sprintf ? 
If someone defines over 10 threads they will kill compilation and then they need to try to fix this.

aclk/aclk_common.h
```diff
@@ -17,6 +47,8 @@ const char *aclk_proxy_type_to_s(ACLK_PROXY_TYPE *type);
 #define ACLK_PROXY_ENV "env"
 #define ACLK_PROXY_CONFIG_VAR "proxy"
 
+#define ACLK_MAX_QUERY_THREADS 4
```
stelfrag, line 50: The query name setting code injects one character (thread number) to build a proper thread name. 

Any reason why not have a max default of 9 queries then?

libnetdata/config/appconfig.c
underhood: well it seems like common operation to have some config values where we want ensure min and max values

aclk/aclk_common.h
underhood: seem bit excessive but you are right it is just arbitrary limit. I can remove it if you think it is better

aclk/aclk_query.c
underhood: when I was doing this I was thinking 640kb (10 threads) is enough for everyone... this is the only reason... In fact I think the stats thread uses sprintf for the same purpose (when generating dimension names)

aclk/aclk_common.h
amoss: Please remove it. We try to avoid arbitrary limits where possible. In this case - if the node is a parent with 100s of children then we may want the ACLK request processing to scale up to 10s or 100s of threads.

libnetdata/config/appconfig.c
amoss: And yet it has never occurred in the hundreds of config options that we currently have. This is because we try to avoid placing arbitrary limits.

aclk/aclk_common.h
underhood: OK I will remove the max limitation and related code

stelfrag:
The more I look it, I think it could also be approached in the following way. Instead of having a lot of threads
being created on startup and having to loop and chart the statistics for them.

- Create one query thread on startup 
- On incoming message if all threads are busy (e.g. check available threads vs active threads)
  - Start a new thread if we are below the user defined thread count
     - When the maximum specified limit is reached we let the available threads handle the load
  - Queue the message

I think it will eventually dynamically adjust to the right number of threads for the particular agent workload

Handling spikes in workload that will result to many threads not needed anymore can also be handled by shutting down excess threads. 

A global variable `query_thread_shutdown_count` can be used to ask for threads to terminate if no longer needed.

- Decide how many threads to shutdown
  - When a thread services a request will check if `query_thread_shutdown_count > 1`, decrement and shutdown

I understand this will change the collected statistics from queries per thread to a different chart that just shows `queries threads active`

underhood:
@stelfrag hmmm that might be an improvement idea in the future *(starting threads on-demand with up to max thread count logic)*, but I do not think this is something to be changed with current PR - there is already another work being done (ACLK improvements) on top of this branch.

Additionally having sleeping threads doesn't really add much overhead.

I think this is valid improvement idea but not really a bug with this PR

underhood:
rebased on latest master (no other changes)

aclk/aclk_query.c
```diff
@@ -0,0 +1,574 @@
...
+pthread_cond_t query_cond_wait = PTHREAD_COND_INITIALIZER;
+pthread_mutex_t query_lock_wait = PTHREAD_MUTEX_INITIALIZER;
+#define QUERY_THREAD_LOCK pthread_mutex_lock(&query_lock_wait)
+#define QUERY_THREAD_UNLOCK pthread_mutex_unlock(&query_lock_wait)
+
+volatile int aclk_connected = 0;
+
+#ifndef __GNUC__
+#pragma region ACLK_QUEUE
```
mfundul, line 13: I think this will break most non-gcc compilers such as clang.

aclk/aclk_query.c
```diff
@@ -0,0 +1,574 @@
+#include "aclk_common.h"
+#include "aclk_query.h"
+#include "aclk_stats.h"
+
+pthread_cond_t query_cond_wait = PTHREAD_COND_INITIALIZER;
+pthread_mutex_t query_lock_wait = PTHREAD_MUTEX_INITIALIZER;
+#define QUERY_THREAD_LOCK pthread_mutex_lock(&query_lock_wait)
```
mfundul, line 7:
```suggestion
#define QUERY_THREAD_LOCK(X) pthread_mutex_lock(&query_lock_wait)
```

aclk/aclk_query.c
underhood: Clang has support as far as I know

aclk/aclk_query.c
underhood: ?

mfundul: It's just a suggestion. With this definition you can use the macro like a function in the code:
```
QUERY_THREAD_LOCK();
```
This reads like a function that does stuff.

aclk/aclk_query.c
underhood: 1. Microsoft Documentation on the feature:
https://docs.microsoft.com/en-us/cpp/preprocessor/region-endregion

1. LLVM change which enabled `pragma region` across the board:
https://reviews.llvm.org/D42248

1. GCC change request https://gcc.gnu.org/bugzilla//show_bug.cgi?id=85487

aclk/aclk_query.c
underhood: Oh OK I see you mean code readability. Well, I followed the pattern in ACLK that already existed which uses it without brackets. If we want to change it we have to do it for all the locks then in ACLK.

mfundul: I have not requested changes :)

aclk/aclk_query.c
mfundul: FreeBSD 12 does not complain so I guess we are ok.

aclk/aclk_query.c
```diff
@@ -0,0 +1,574 @@
...
+
+        this_query = aclk_queue.aclk_query_head;
+    }
+
+    if (likely(!this_query)) {
+        ACLK_QUEUE_UNLOCK;
+        return NULL;
+    }
+
+    if (!this_query->deleted && this_query->run_after > now_realtime_sec()) {
```
mfundul, line 108: Isn't `!this_query->deleted` always true after getting rid of the deleted entries? This could be an assertion like:
```
fatal_assert(!this_query->deleted);
```

underhood: hmm will take a look, I was not checking this as this is just moved from another file

mfundul: Since I'm not familiar with ACLK I'm looking at all the differences of this PR to try and make sense, so I don't know what to ignore!

aclk/aclk_query.c
```diff
@@ -0,0 +1,574 @@
...
+void aclk_query_threads_start(struct aclk_query_threads *query_threads)
+{
+    info("Starting %d query threads.", query_threads->count);
+
+    char thread_name[TASK_LEN_MAX];
+    query_threads->thread_list = callocz(query_threads->count, sizeof(struct aclk_query_thread *));
+    for (int i = 0; i < query_threads->count; i++) {
+        query_threads->thread_list[i] = callocz(1, sizeof(struct aclk_query_thread));
+        query_threads->thread_list[i]->thread = callocz(1, sizeof(netdata_thread_t));
+        query_threads->thread_list[i]->idx = i; //thread needs to know its index for statistics
```
mfundul, line 492: Do we need so many pointers? I'm having difficulty following all the memory allocations related to threads. Could this work if it was like the following?
```
    query_threads->thread_list = callocz(query_threads->count, sizeof(struct aclk_query_thread));
    for (int i = 0; i < query_threads->count; i++) {
        query_threads->thread_list[i].thread = callocz(1, sizeof(netdata_thread_t));
        query_threads->thread_list[i].idx = i; //thread needs to know its index for statistics 

        snprintf(thread_name, TASK_LEN_MAX, "%s_%d", ACLK_THREAD_NAME, i);
        netdata_thread_create(
            query_threads->thread_list[i].thread, thread_name, NETDATA_THREAD_OPTION_JOINABLE, aclk_query_main_thread,
            query_threads->thread_list[i]);
    } 
```
Moreover, could the `netdata_thread_t` be part of `struct aclk_query_thread` thus removing the extra allocation?
```
    query_threads->thread_list = callocz(query_threads->count, sizeof(struct aclk_query_thread));
    for (int i = 0; i < query_threads->count; i++) {
        query_threads->thread_list[i].idx = i; //thread needs to know its index for statistics 

        snprintf(thread_name, TASK_LEN_MAX, "%s_%d", ACLK_THREAD_NAME, i);
        netdata_thread_create(
            &query_threads->thread_list[i].thread, thread_name, NETDATA_THREAD_OPTION_JOINABLE, aclk_query_main_thread,
            query_threads->thread_list[i]);
    } 
```

aclk/agent_cloud_link.c
```diff
@@ -1400,18 +868,25 @@ void *aclk_main(void *ptr)
         }
     }
 
+    query_threads.count = appconfig_get_number(&cloud_config, CONFIG_SECTION_GLOBAL, "query thread count", 2);
```
mfundul, line 871: I feel that choosing 2 as the default number is too little. A lot of people have systems that could take advantage of more threads. Even raspberry pi has 4-cores these days.

underhood: Well currently even 1 thread is enough. This is the motive of default being 2. Until we see need for it (longer queries) I felt having 4 threads just sitting there is not necessary (even if they add little overhead)

mfundul: This should probably be in `netdata.conf` and not in the machine generated file in `/var/lib/cloud.d`.

mfundul: We should also have a line in a document somewhere that describes this so that users know how to configure this.

aclk/aclk_query.c
underhood: Thanks, you are right that is unnecessary indirection and only makes code harder to read. I simplified it.

aclk/agent_cloud_link.c
underhood: thanks, done

aclk/agent_cloud_link.c
```diff
@@ -1400,18 +868,25 @@ void *aclk_main(void *ptr)
         }
     }
 
-    aclk_stats_enabled = appconfig_get_boolean(&cloud_config, CONFIG_SECTION_GLOBAL, "statistics", CONFIG_BOOLEAN_YES);
+    query_threads.count = config_get_number(CONFIG_SECTION_CLOUD, "query thread count", 2);
+    if(query_threads.count < 1) {
+        error("You need at least one query thread. Overriding configured setting of \"%d\"", query_threads.count);
+        query_threads.count = 1;
```
stelfrag, line 874: The appconfig_set_number should change to config_set_number to go to netdata conf and the correct section

aclk/README.md
```diff
@@ -38,6 +38,16 @@ configuration uses two settings:
 If your Agent needs to use a proxy to access the internet, you must [set up a proxy for
 claiming](/claim/README.md#claim-through-a-proxy).
 
+Additionally in the `netdata.conf` section `[cloud]` you can configure following keys:
+```
```
stelfrag, line 42: Rephrase ?   you can configure the following keys:

mfundul:
We have an issue with statistics:
```
  CC       aclk/aclk_query.o
aclk/aclk_query.c: In function ‘aclk_query_main_thread’:
aclk/aclk_query.c:453:15: warning: ‘t’ may be used uninitialized in this function [-Wmaybe-uninitialized]
             if(aclk_metrics_per_sample.cloud_q_process_max < t)
               ^
aclk/aclk_query.c:387:12: note: ‘t’ was declared here
     usec_t t;
            ^

```

stelfrag:
I zoomed in to reduce the number of charts updated on my screen and I got

image

underhood:
> I zoomed in to reduce the number of charts updated on my screen and I got
> 
> image

the GUI seems in this case to downsample and query every 2s.

PR #9364, opened 2020-06-17
Add documentation for installing Netdata on k8s clusters

joelhans:
##### Summary

After wading my way through a crash-course in k8s over the last few days, I've realized we need some better information about how users can install Netdata on their k8s clusters. I'm no longer convinced simply linking to the Helm chart is enough, particularly when you add in things like service discovery.

I'm working on a second part to this—a guide that focuses on finding and interpreting all the charts this deployment creates (especially service discovery, which is awesome!), but I need a bit more time to put all the pieces together.

##### Component Name

area/docs
area/packaging

### DIFF
--- a/packaging/installer/methods/kubernetes.md
+++ b/packaging/installer/methods/kubernetes.md
@@ -36,8 +36,7 @@ Download the [Netdata Helm chart](https://github.com/netdata/helmchart) on the a
 `helm` binary installed.
 
 ```bash
-helm repo add netdata https://netdata.github.io/helmchart/
-helm repo update
+git clone https://github.com/netdata/helmchart.git netdata-helmchart
 ```
 
 > You may not need to configure the Helm chart to get a functioning service on your cluster, but you should read the
@@ -47,7 +46,7 @@ helm repo update
 Install the Helm chart to your cluster with `helm install`:
 
 ```bash
-helm install netdata netdata/netdata
+helm install netdata ./netdata-helmchart
 ```
 
 Run `kubectl get services` and `kubectl get pods` to confirm that your cluster now runs a `netdata` service, one
@@ -66,13 +65,13 @@ on your cluster's setup.
 To change a setting, use the `--set` or `--values` arguments along with `helm install`:
 
 ```bash
-helm install --set a.b.c=xyz netdata netdata/netdata
+helm install --set a.b.c=xyz netdata ./netdata-helmchart
 ```
 
 For example, to change the size of the persistent metrics volume, you would run the following:
 
 ```bash
-helm install --set parent.database.volumesize=4Gi netdata
+helm install --set parent.database.volumesize=4Gi ./netdata-helmchart
 ```
 
 ### Configure service discovery
@@ -91,7 +90,7 @@ configurations based on the service that pod runs, and begin monitoring them imm
 
 However, if you have changed some of these defaults, you'll need to open the `netdata-helmchart/sd-child.yml` file and
 edit the ports or image names that service discovery uses to find supported pods. You can then run `helm upgrade netdata
-netdata/netdata` to push changes to your cluster.
+./netdata-helmchart` to push changes to your cluster.
 
 ## Access the Netdata dashboard
 
@@ -125,7 +124,7 @@ If you update the Helm chart's configuration, either via `values.yml` or `sd-chi
 replacing `netdata` with the name of the release if you changed it upon installtion:
 
 ```bash
-helm upgrade netdata netdata/netdata
+helm upgrade netdata ./netdata-helmchart
 ```
 
 ## What's next?
--- a/packaging/installer/methods/kubernetes.md
+++ b/packaging/installer/methods/kubernetes.md
@@ -90,9 +90,24 @@ services](https://github.com/netdata/helmchart#service-discovery-and-supported-s
 If you haven't changed listening ports or other defaults, service discovery should find your pods, create the proper
 configurations based on the service that pod runs, and begin monitoring them immediately after depolyment.
 
-However, if you have changed some of these defaults, you'll need to open the `netdata-helmchart/sd-child.yml` file and
-edit the ports or image names that service discovery uses to find supported pods. You can then run `helm upgrade netdata
-./netdata-helmchart` to push changes to your cluster.
+However, if you have changed some of these defaults, you'll need to copy the `netdata-helmchart/sdconfig/child.yml`
+file, edit it, and pass the changed file to `helm install`. First, copy the file to a new location.
+
+```bash
+cp netdata-helmchart/sdconfig/child.yml .
+```
+
+Edit the new `child.yml` file according to your needs. See the [Helm chart
+configuration](https://github.com/netdata/helmchart#configuration) and the file itself for details. You can then run
+`helm install`/`helm upgrade` with the `--set-file` argument to use your configured `child.yml` file instead of the
+default.
+
+```bash
+helm install --set-file sd.child.configmap.from.value=../child.yml netdata ./netdata-helmchart
+helm upgrade --set-file sd.child.configmap.from.value=../child.yml netdata ./netdata-helmchart
+```
+
+Your configured service discovery is now pushed to your cluster.
 
 ## Access the Netdata dashboard
 
@@ -122,8 +137,8 @@ In the above example, access the dashboard by navigating to `http://203.0.113.0:
 
 ## Update/reinstall the Netdata Helm chart
 
-If you update the Helm chart's configuration, either via `values.yml` or `sd-child.yml`, you can run `helm upgrade`,
-replacing `netdata` with the name of the release if you changed it upon installtion:
+If you update the Helm chart's configuration, run `helm upgrade` to redeploy your Netdata service, replacing `netdata` 
+with the name of the release if you changed it upon installtion:
 
 ```bash
 helm upgrade netdata ./netdata-helmchart
--- a/packaging/installer/methods/kubernetes.md
+++ b/packaging/installer/methods/kubernetes.md
@@ -26,9 +26,11 @@ To install Netdata on a Kubernetes cluster, you need:
     administrative system.
 -   The [Helm package manager](https://helm.sh/) v3.0.0 or newer on the same administrative system.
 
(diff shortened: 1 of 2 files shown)


packaging/installer/methods/kubernetes.md
```diff
@@ -0,0 +1,134 @@
...
+-   The [Helm package manager](https://helm.sh/) on an administrative system.
+-   The [kubectl](https://kubernetes.io/docs/reference/kubectl/overview/) command line tool on the same system.
+
+The default configuration creates four pods in your cluster: three `child` pods and one `parent` pod. The `child` pods
+collect metrics and stream the information to the `parent` pod, which uses two persistent volumes to store metrics and
+alarms. The `parent` pod also handles alarm notifications and enables the Netdata dashboard using an ingress controller.
+
+## Install the Netdata Helm chart
+
+Download the [Netdata Helm chart](https://github.com/netdata/helmchart) on the administative system where you have the
```
ncmans, line 33: Preferred way to install charts is via repository and we have one: https://netdata.github.io/helmchart/

To install the repo:

```
helm repo add netdata https://netdata.github.io/helmchart/
helm repo update
```

To install the chart:

```
helm install netdata netdata/netdata
```

packaging/installer/methods/kubernetes.md
```diff
@@ -0,0 +1,134 @@
...
+dashboard](#access-the-netdata-dashboard) to confirm it's working as expected, or see the next section to [configure the
+Helm chart](#configure-the-netdata-helm-chart) to suit your cluster's particular setup.
+
+## Configure the Netdata Helm chart
+
+Read up on the various configuration options in the [Helm chart
+documentation](https://github.com/netdata/helmchart#configuration) to see if you need to change any of the options based
+on your cluster's setup.
+
+To change a setting, you can edit the `values.yml` file inside of the `netdata-helmchart` folder. Then install the chart
```
ncmans, line 64: Preferred way to set a custom setting for a chart is to pass those settings as `--set` args or `--values` arg to helm install:

```
helm install --set a.b.c=xyz netdata netdata/netdata
```

packaging/installer/methods/kubernetes.md
ncmans: Note: repo is a bit outdated, added an issue to bring it up to date: https://github.com/netdata/infra/issues/626

ncmans: Note: repo is a bit outdated, added an issue to bring it up to date

joelhans: Cool! I didn't know that was a thing. Updated this portion.

packaging/installer/methods/kubernetes.md
joelhans: What's the recommended way to edit the `sdconfig/slave.yml` file once that's enabled in the Helm chart? Are those values changeable via the same `--set ...` command?

packaging/installer/methods/kubernetes.md
ncmans: Sorry @joelhans , seems like I spoke too soon. The repo isn't official yet, so ignore my comments here for now.

joelhans: No worries! Reverted back to the old method.

packaging/installer/methods/kubernetes.md
```diff
@@ -0,0 +1,138 @@
...
+
+To install Netdata on a Kubernetes cluster, you need:
+
+-   A working cluster running Kubernetes v1.9 or newer.
+-   The [kubectl](https://kubernetes.io/docs/reference/kubectl/overview/) command line tool, within [one minor version
+    difference](https://kubernetes.io/docs/tasks/tools/install-kubectl/#before-you-begin) of your cluster, on an
+    administrative system.
+-   The [Helm package manager](https://helm.sh/) v3.0.0 or newer on the same administrative system.
+
+The default configuration creates four pods in your cluster: three `child` pods and one `parent` pod. The `child` pods
```
ilyam8, line 29: child is [daemonset](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/)

> A DaemonSet ensures that all (or some) Nodes run a copy of a Pod. As nodes are added to the cluster, Pods are added to them. As nodes are removed from the cluster, those Pods are garbage collected. Deleting a DaemonSet will clean up the Pods it created.

> three `child` pods

So it is _number of nodes_ children. 

Every node has netdata child - one netdata per node.

And only one node has 2 netdatas - parent + child.

cakrit:
LGTM, but the comments need to be addressed.
It would be great if on a next version we could extend this with with an example using terraform @knatsakis and @ncmans

packaging/installer/methods/kubernetes.md
joelhans: @ilyam8 Do you have any insights here? To change the service discovery config, do you edit `sdconfig/child.yml`, or do you use the `--set` command?

packaging/installer/methods/kubernetes.md
joelhans: Thanks for noticing this. I forgot to change this part from when I was talking about a specific 3-node cluster. Changed it up to be clearer that there's a child for every node, no matter how many nodes there are.

packaging/installer/methods/kubernetes.md
```diff
@@ -0,0 +1,140 @@
...
+
+Service discovery scans your cluster for pods exposed on certain ports and with certain image names. By default, it
+looks for its supported services on the ports they most commonly listen on, and using default image names. Service
+discovery currently supports [22 popular
+services](https://github.com/netdata/helmchart#service-discovery-and-supported-services).
+
+If you haven't changed listening ports or other defaults, service discovery should find your pods, create the proper
+configurations based on the service that pod runs, and begin monitoring them immediately after depolyment.
+
+However, if you have changed some of these defaults, you'll need to open the `netdata-helmchart/sd-child.yml` file and
```
ilyam8, line 93: netdata-helmchart/sdconfig/child.yml

> https://github.com/netdata/helmchart/blob/master/sdconfig/child.yml

packaging/installer/methods/kubernetes.md
```diff
@@ -0,0 +1,140 @@
...
+cockroachdb-public   ClusterIP      10.245.148.233   <none>         26257/TCP,8080/TCP   46h
+kubernetes           ClusterIP      10.245.0.1       <none>         443/TCP              47h
+netdata              LoadBalancer   10.245.160.131   203.0.113.0    19999:32231/TCP      74m
+```
+
+In the above example, access the dashboard by navigating to `http://203.0.113.0:19999`.
+
+## Update/reinstall the Netdata Helm chart
+
+If you update the Helm chart's configuration, either via `values.yml` or `sd-child.yml`, you can run `helm upgrade`,
```
ilyam8, line 125: /sdconfig/child.yml

packaging/installer/methods/kubernetes.md
ncmans: > What's the recommended way to edit the `sdconfig/slave.yml` file once that's enabled in the Helm chart? Are those values changeable via the same `--set ...` command?

One way we do it for cloud currently, is to use `--values sdconfig/slave.yml` and then override any other config further with `--set`.

ncmans:
> LGTM, but the comments need to be addressed.
> It would be great if on a next version we could extend this with with an example using terraform @knatsakis and @ncmans

Once we make the helm repo official (and keep it up-to-date with releases), the terraform example will be much cleaner (as will be the rest of the examples).

packaging/installer/methods/kubernetes.md
ncmans: > What's the recommended way to edit the `sdconfig/slave.yml` file once that's enabled in the Helm chart? Are those values changeable via the same `--set ...` command?

For this, the user can upload an updated `sdconfig/slave.yml` containing their changes, to a new configmap and reference that configmap via `sd.child.configmap.name` and `sd.child.configmap.key` configs.

ilyam8: I dont think it is possible to use `--set`, however there is `--set-file`

I created a PR

> https://github.com/netdata/helmchart/pull/109

When it is merged it will be possible to use `--set-file`


```
... --set-file sd.child.configmap.from.value=../child.yml
```

ilyam8: i think it is not correct to edit `/sdconfig/child.yml`, you use it as a reference, copy, edit and `--set-file` the new file.

packaging/installer/methods/kubernetes.md
joelhans: Thanks.

packaging/installer/methods/kubernetes.md
joelhans: Okay, I added some additional steps around editing a copy of `child.yml` and use `--set-file`, and we can merge this once netdata/helmchart#109 is merged and we know the process works as expected.

packaging/installer/methods/kubernetes.md
```diff
@@ -90,9 +90,24 @@ services](https://github.com/netdata/helmchart#service-discovery-and-supported-s
...
+```
+
+Edit the new `child.yml` file according to your needs. See the [Helm chart
+configuration](https://github.com/netdata/helmchart#configuration) and the file itself for details. You can then run
+`helm install`/`helm upgrade` with the `--set-file` argument to use your configured `child.yml` file instead of the
+default.
+
+```bash
+helm install --set-file sd.child.configmap.from.value=../child.yml netdata ./netdata-helmchart
+helm upgrade --set-file sd.child.configmap.from.value=../child.yml netdata ./netdata-helmchart
```
ilyam8, line 107: > `../child.yml`

this should be path to the file config, i used `../child.yml` a an example and to show that there is no requirement to have the file in the root dir.

ilyam8: something like this

```cmd
helm install --set-file sd.child.configmap.from.value=<FILE> netdata ./netdata-helmchart
```

---

```
--set-file stringArray  set values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)
```

joelhans: Change the instructions to copy `child.yml` outside the helmchart root directory, and tested it myself. Works great!

image

Changed their names to include `-testing-`.

PR #9374, opened 2020-06-18
Stop reading from /proc/sys/kernel/osrelease at trailing newline.

sjuxax:
##### Summary

`ebpf.plugin` and several parts of the `kernel-collector` sibling project either read `/proc/sys/kernel/osrelease` directly or use it as a fallback. This file ends with a newline, which is read into the `kernel_string` variable. `ebpf.plugin` then fails to find the eBPF libraries because it's looking for a filename that ends with a newline.

Since it never makes sense for `osrelease` to include a newline, update `ebpf.plugin` to stop reading by the newline. This fixes the eBPF plugin on my system (which is running a custom kernel) and allows it to load the `kernel-collector` components.

##### Component Name

`ebpf.plugin`

##### Test Plan

I haven't added any tests, but it'd be a good thing to add a test for. I assume a test for this doesn't already exist or it wouldn't have been a problem. 😊 

##### Additional Information

Similar commit for the `kernel-collectors` is on my local and may get around to making a PR for it some day.

### DIFF
--- a/libnetdata/ebpf/ebpf.c
+++ b/libnetdata/ebpf/ebpf.c
@@ -87,7 +87,7 @@ int get_kernel_version(char *out, int size) {
         return -1;
 
     move = patch;
-    while (*version) *move++ = *version++;
+    while (*version && *version != '\n') *move++ = *version++;
     *move = '\0';
 
     fd = snprintf(out, (size_t)size, "%s.%s.%s", major, minor, patch);


prologic:
@netdata/agent ☝️

sjuxax:
np, thanks for the reviews everyone. Was a little bummed to see this didn't get merged into 1.23, especially since eBPF is one of the release's major headlines. Is there something else I need to do before the PR can be merged? No big deal if this was patched through something else of course, would just be good to notate here for the record. Thanks!

underhood:
@sjuxax you seem to have 2 approving reviews already so probable reason why it didn't get merged is current code freeze (which is active around release)

thiagoftsm:
Hi @sjuxax ,

I am sorry, I forgot to give you an explanation about the code freeze, but as soon it finishes we will merge your PR.

best regards!

Issue #7587, opened 2019-12-19
Improve handling of the named pipe used by `netdatacli`.


Ferroin:
Currently, the named pipe that is used for sending commands to the agent process from the `netdatacli` command is hardcoded to be located at `/tmp/netdata-ipc` on UNIX systems. This is problematic for a few reasons:

* `/tmp` is world writable. This means that a user can theoretically cause issues by creating a file (or symlink) with the same name while Netdata is not running.
* `/tmp` is not consistently 'unified'. In some setups, each user has their own view of this directory, which 
* Per a strict inteprretation of the filesystem hierarchy standards, `/tmp` is not the correct place for the named pipe anyway, since it's a runtime file and not a temporary file.

To resolve these issues, I would propose the following two changes:

* The default location for this named pipe on UNIX systems should be changed to either `/run` (on Linux and other systems which provide `/run`) or `/var/run` (on systems which do not provide `/run`). This will solve the first and third issues above, but not the second one (which is also true for `/run`)
* The exact location should be made user configurable with an option in the config file. This will solve the second issue above.

mfundul:
I picked `/tmp/` since this is the default location for `mysql` and `postgreSQL` domain sockets as well. Moreover, I needed a directory that is guaranteed to not be globally viewable. E.g. if there are many containers running netdata inside, each container should have its own private socket to interface with its own private cli tool, which I hope is guaranteed with the choice of `/tmp/`.

Let's try not to break this assumption.

mfundul:
Then of course there is the question about the official docker installation of netdata. Are we supposed to separately install cli in the host filesystem and the rest in the container?

amoss:
It is true that the socket is in the wrong directory. I think that /run or /var/run matches the semantics better and we can create e.g. /var/run/netdata/ at install-time to  solve problem (1) above.

Our docker image does not bind-mount /run or /var/run so each container would still have its own private socket. I don’t see the problem with multiple containers that @mfundul describes above.

Problem (2) sounds like a benefit - this means that the netdata user would effectively have a private /tmp directory.

Number (3) is correct and it could cause a problem for our district packaging.

Ferroin:
> It is true that the socket is in the wrong directory. I think that /run or /var/run matches the semantics better and we can create e.g. /var/run/netdata/ at install-time to solve problem (1) above.

[...]

PR #9377, opened 2020-06-18
Write pid file (needed by logrotate)

### LINKED ISSUE #8898: wrong logrotate configuration in debian package at packagecloud.io
##### Bug report summary

If you install the debian package from packagecloud.io, the logs aren't reopened so netdata will still append to the old log file.

##### OS / Environment

```
debian 10 (buster) x86_64
```

##### Netdata version

 v1.21.1

```
$ LC_ALL=C dpkg -l netdata
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-======================================
ii  netdata        1.21.1       amd64        real-time charts for system monitoring

```

##### Steps To Reproduce

1. install the debian package from packagecloud.io
2. start it
3. wait one day and see that /var/log/netdata/error.log is empty and messages are appended to /var/log/netdata/error.log.1

##### Expected behavior

after the logs have been rotated the new messages should be appended to /var/log/netdata/error.log

##### Cause

the logrotate configuration supplied with the package (/etc/logrotate.d/netdata) uses 

``` /bin/kill -HUP `cat /var/run/netdata/netdata.pid 2>/dev/null` 2>/dev/null || true```

in the postrotate section, however the pid file doesn't exist since the systemd unit supplied doesn't use the -P parameter to start netdata (it is in the init.d file but that's not used with systemd).

```
$ systemctl status netdata
● netdata.service - Real time performance monitoring
   Loaded: loaded (/lib/systemd/system/netdata.service; enabled; vendor preset: enabled)
   Active: active (running) since Mon 2020-05-04 14:59:56 CEST; 18h ago
 Main PID: 14984 (netdata)
    Tasks: 35 (limit: 2358)
   Memory: 133.1M
   CGroup: /system.slice/netdata.service
           ├─14984 /usr/sbin/netdata -D
           ├─15077 /usr/libexec/netdata/plugins.d/go.d.plugin 1
           ├─15082 /usr/bin/python /usr/libexec/netdata/plugins.d/python.d.plugin 1
           ├─25886 /usr/libexec/netdata/plugins.d/nfacct.plugin 1
           ├─25887 /usr/libexec/netdata/plugins.d/apps.plugin 1
           └─31995 bash /usr/libexec/netdata/plugins.d/tc-qos-helper.sh 1
```
A possible solution would be to use `netdatacli reopen-logs` instead of  `kill -HUP` as suggested in issue #8244

The files are as distributed in the package:

```
$ dpkg -S /etc/logrotate.d/netdata /lib/systemd/system/netdata.service 
netdata: /etc/logrotate.d/netdata
netdata: /lib/systemd/system/netdata.service
$ sudo dpkg -V netdata
??5?????? c /etc/netdata/netdata.conf

```

olivluca:
This should fix issue #8898 at least until issue #7587 is fixed

Fixes #8898

### DIFF
--- a/contrib/debian/netdata.default
+++ b/contrib/debian/netdata.default
@@ -1,7 +1,3 @@
 # Extra arguments to pass to netdata
 #
-#EXTRA_OPTS=""
-#
-# Location of pidfile. Only used on SysV init based systems.
-#
-#PIDFILE="/var/run/netdata/netdata.pid"
+EXTRA_OPTS="-P /var/run/netdata/netdata.pid"


prologic (dismissed):

iigorkarpov (commented):
I've checked Debian 10, Debian 11, Ubuntu 18, Ubuntu 20, Ubuntu 22. No pid file created after uncommenting this line and restarting the service.
**_Correction. I experimented with PIDFILE, not EXTRA_OPTS. EXTRA_OPTS works._** I think this is wrong but still it could be used as a temporary workaround.

ilyam8:
@olivluca hi, can you rebase onto the master branch and resolve merge conflicts?

iigorkarpov:
However, this approach [is not recommende](https://www.freedesktop.org/software/systemd/man/systemd.service.html#PIDFile=)d for the systems with systemd:

`Note that PID files should be avoided in modern projects. Use Type=notify or Type=simple where possible, which does not require use of PID files to determine the main process of a service and avoids needless forking.`

iigorkarpov:
Please, instead of forcing pidfile creation use the following:

`/var/log/netdata/*.log {`
        `daily`
        `missingok`
        `rotate 14`
        `compress`
        `delaycompress`
        `notifempty`
        `sharedscripts`
        `copytruncate`
        `postrotate`
               ` #/bin/kill -HUP `cat /var/run/netdata/netdata.pid 2>/dev/null` 2>/dev/null || true`
        `endscript`
`}`

Ferroin:
@iigorkarpov It actually needs to be more complex than just that. We either need a separate logrotate config for systemd systems (I forget the exact command, but you can use systemctl to send signals to systemd services), or we would need to fix the issues with netdatacli and use _that_ instead (which works without a PID file).

ilyam8:
Guys, we run Netdata with "-P" when using from source/static method. Why can't we do the same for deb packages (until we get a proper solution with netdatacli in 2024).

iigorkarpov:
@ilyam8 because this is wrong for `systemd` based systems. The services under `systemd` management should not have PID files.
@Ferroin I agree. We discussed this yesterday. For the moment being adding `copytruncate` and removing `postrotate` section from `logrotate` configuration seems the lesser evil to me. This should work for non-`systemd` systems as well.
The only thing that worries me is the assumption that some older versions may not have this option. I have not been able to find out exactly when it was added.

ilyam8:
> **"-P" when using from source/static installation method**

^^ As you may guess **it includes systemd systems**.

> because this is wrong for systemd based systems.

I don't see how strictly interpreting the paragraph you found is better than having the issue unfixed for (soon will be) 2 years. Are there any other problems we can get apart from doing something "wrong" according to the PIDFile paramater description?

iigorkarpov:
@ilyam8 here we are persistently falling into the same sin that I have repeatedly mentioned.
Instead of behaving like a well-behaved application, we begin to reshape the user system to suit ourselves. This is unacceptable.

Ferroin:
@ilyam8 I have no particular issue with fixing things as this PR does _for now_, provided we fix them properly later (and I would argue the proper fix is to finally fix the handling of the named pipe for `netdatacli` and then switching to using that because it should work _everywhere_ instead of needing to modify how we’re handling services on some systems). The problem I have is that ‘later’ functionally never happens around here, which means it usually makes more sense to take the extra time to fix things properly in the first place because otherwise they will often never get fixed properly at all.

@iigorkarpov This would not, in fact, be us reshaping the user’s system to suit ourselves. It is us changing _OUR OWN APPLICATION_ to behave a bit better in certain circumstances, which is by definition expected of application developers. Systemd’s abject hatred of pidfiles (because of a whole slew of edge cases where they don’t work which should never crop up on a properly audited system in the first place) has nothing to do with how things outside of systemd are allowed to interact with a systemd service. The paragraph you’re quoting is, in fact, them doing exactly the thing you’re claiming this change would be doing.

iigorkarpov:
Let's look at it from the user's point of view. Recall that in this case it is a person with administrator privileges on this host. What does he see? A package that violates some convention that has become common in the last few years. And breaking it unnecessarily. Its creators are either too lazy or too unorganized or just technically illiterate. Either they _don't know how_ to do it, or they know, but they are arrogant enough to follow the conventions.
Either way, we look bad as a result.

Ferroin:
@iigorkarpov Yes, but not as bad as we do when we ship things _that don’t work at all_. Logrotate does not work properly right now on Debian systems. This PR fixes it without breaking anything else or impacting any other systems.

Yes, it’s not the ideal solution, but the ideal solution requires a significant amount of additional development work, during which time _logrotate will still be broken on Debian systems_.

iigorkarpov:
And why this is the best solution comparing to the small change to logrotate configuration?

Ferroin:
Because that affects all systems, not just the systems that we are seeing this particular issue on.

olivluca:
I see that now there are two versions of the systemd service and I don't know which one is used. 
[this one](https://github.com/netdata/netdata/blob/master/system/netdata.service.in already uses the -P parameter, so no patch is needed, while [this one](https://github.com/netdata/netdata/blob/master/system/netdata.service.v235.in) would need it in EXTRA_OPTS instead of in PIDFILE

Ferroin:
@olivluca Just fixing the one that needs fixed should be sufficient. I actually don’t remember the full list of which service file gets used where myself.

ilyam8 (approved):

Ferroin (approved):

PR #9384, opened 2020-06-19
Remove chart specific configuration from netdata.conf

### LINKED ISSUE #9320: Per chart settings on netdata.conf & streaming
##### Bug report summary
When we have dbengine with multiple child nodes sending to a parent, the web request for netdata.conf takes forever, because it includes the settings for each chart.

##### Steps To Reproduce
Visit http://host:19999/netdata.conf on any parent node with dbengine and tens of children

##### Expected behavior
netdata.conf returns in a timely manner.

We discussed with @ktsaou and he agrees to **remove the per-chart settings from netdata.conf**. They aren't useful anyway.

stelfrag:
Fixes #9320
##### Summary
Remove support for 'per chart' configuration in `netdata.conf` (except the `enabled` option)

Right now on every chart creation there is a lookup in the config file to determine if a section that matches the chart exists

- For parent nodes the section matches the chart name
- For child nodes the section is in the form guid/chart_name

##### Component Name
database
agent

##### Test Plan
- Fetch netdata.conf using ` http://host:19999/netdata.conf` to notice the absence of chart specific configuration.

### DIFF
--- a/database/rrd.c
+++ b/database/rrd.c
@@ -140,6 +140,7 @@ const char *rrdset_type_name(RRDSET_TYPE chart_type) {
 // RRD - cache directory
 
 char *rrdset_cache_dir(RRDHOST *host, const char *id, const char *config_section) {
+    UNUSED(config_section);
     char *ret = NULL;
 
     char b[FILENAME_MAX + 1];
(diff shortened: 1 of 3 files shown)


thiagoftsm (approved):
Everything was OK on my tests, I used `dbengine` and `RAM`, but for both tests I was using Virtual Machines.

PR #9396, opened 2020-06-22
Fix missing host variables on stream

### LINKED ISSUE #9375: Some agents not setting $active_processors alarm variable
##### Question summary
I am running a distributed setup with ~10 hosts streaming to a central server. Lately, I have been receiving a lot of "high system load" alerts. After some checking, I found that the `$active_processors` alarm variable is not being set on most of my systems, so the alarm is defaulting to 2 cpu cores (when most of my systems are 8+). When checking `https://my.central.server/host/specific.node/api/v1/alarm_variables?chart=system.load`, there is no `active_processors` variable under `host_variables` on the systems affected.

What's the best why I can debug this? I checked docs and I did not find a way to debug _internal_ plugins, like `proc.plugin`.
Thanks!

##### OS / Environment
Central host: Ubuntu 18
Most streaming nodes: Ubuntu 18

##### Component Name
`system.load`

thiagoftsm:
##### Summary
Fixes #9375 

We were not copying properly the data between buffer and with this the HOST variables were not properly copied to the parent.
##### Component Name
Stream
##### Test Plan
1 - Compile this PR mainly on children hosts.
2 - Start parent and after few seconds start the slave.
3 - Do the following request for the parent `https://localhost:19999/host/CHILD_NAME/api/v1/alarm_variables?chart=system.load` and check if the variable `active_processors` is present.
4 - Repeat the steps 1, 2 and 3 at least three times.

##### Additional Information

### DIFF
--- a/streaming/sender.c
+++ b/streaming/sender.c
@@ -70,8 +70,10 @@ static int rrdpush_sender_thread_custom_host_variables_callback(void *rrdvar_ptr
 }
 
 static void rrdpush_sender_thread_send_custom_host_variables(RRDHOST *host) {
+    sender_start(host->sender);
     int ret = rrdvar_callback_for_all_host_variables(host, rrdpush_sender_thread_custom_host_variables_callback, host);
     (void)ret;
+    sender_commit(host->sender);
 
     debug(D_STREAM, "RRDVAR sent %d VARIABLES", ret);
 }


streaming/sender.c
```diff
@@ -70,8 +70,10 @@ static int rrdpush_sender_thread_custom_host_variables_callback(void *rrdvar_ptr
 }
 
 static void rrdpush_sender_thread_send_custom_host_variables(RRDHOST *host) {
+    sender_start(host->sender);
```
amoss, line 73: Acquiring and releasing the lock twice (and also duplicating the copy from the build into the circular buffer) is less efficient. Was there a reason not to move the call into the locked section?

thiagoftsm: Yes,  I moved it there and the problem persists.

amoss: Are the callbacks deferred or do they happen during the call?

Issue #8198, opened 2020-02-25
Trying to update using a different method than the initial install should explicitly fail.


Ferroin:
When a user goes to update an installation of Netdata that was initially installed with `kickstart.sh` using `kickstart-static64.sh` or vice-versa, we currently end up with a technically incomplete replacement of the original installation unless the `--prefix=/opt/netdata` option was used for `kickstart.sh`.

We should never end up in that state.

Properly handling this in the installers scripts is actually kind of difficult, so I would suggest we simply explicitly do not support this, update the documentation accordingly, and then update the installer scripts to safely bail out and provide a useful error message if such a situation is detected.

Issue #8186, opened 2020-02-24
Regression: netdata working in wrong directories after update


Pingger:
##### Bug report summary

After updating to netdata 1.20 suddenly all configs are loaded and stored in /opt/netdata. here the /etc /opt /usr /var directories are now created and used instead of the normal ones that already exist.  
So the folder structure looks like this:

![image](https://user-images.githubusercontent.com/4761298/75183723-63fa6100-5743-11ea-8a76-03151e050305.png)

The installer even updated some files in the correct "/etc/netdata" folder, like the "edit-config" script
##### OS / Environment

```
Linux ************* #1 SMP Debian 4.19.98-1 (2020-01-26) x86_64 GNU/Linux
/etc/os-release:PRETTY_NAME="Debian GNU/Linux 10 (buster)"
/etc/os-release:NAME="Debian GNU/Linux"
/etc/os-release:VERSION_ID="10"
/etc/os-release:VERSION="10 (buster)"
/etc/os-release:VERSION_CODENAME=buster
/etc/os-release:ID=debian
/etc/os-release:HOME_URL="https://www.debian.org/"
/etc/os-release:SUPPORT_URL="https://www.debian.org/support"
/etc/os-release:BUG_REPORT_URL="https://bugs.debian.org/"

```

##### Component Name

area/packaging

##### Steps To Reproduce

1. run installer `bash <(curl -Ss https://my-netdata.io/kickstart-static64.sh)  --stable-channel --disable-telemetry`
2. have issues

##### Expected behavior

1. run installer,
2. configs are read from correct locations and not from some obscure undocumented locations!
2b. or be properly told, that there is a change in folder structure and what to be aware of when using configs of previous versions!

ilyam8:
```cmd
test@DEB10-1:/opt/netdata$ ls -l --group-directories-first
total 32
drwxrwxr-x 3 netdata netdata 4096 Feb 20 20:56 bin
drwxrwxr-x 4 netdata netdata 4096 Feb 24 15:06 etc
drwxr-xr-x 3 netdata netdata 4096 Feb 20 20:55 include
drwxr-xr-x 3 netdata netdata 4096 Feb 20 20:55 lib
drwxrwxr-x 6 netdata netdata 4096 Feb 20 20:55 share
drwxrwxr-x 2 netdata netdata 4096 Feb 20 20:56 system
drwxrwxr-x 5 netdata netdata 4096 Feb 24 15:06 usr
[...]

PR #9408, opened 2020-06-24
Add better checks for existing installs to the kickstart scripts.

Ferroin:
##### Summary

This adds more robust checks to our Kickstart install scripts to prevent users from accidentally installing Netdata over top of an existing install (which can lead to all kinds of problematic situations).

The logic for these checks works as follows:

* See if `netdata` is located somewhere in `$PATH` (using `command -v`), and if present assume there's an existing install.
* Otherwise, check if `/opt/netdata/bin/netdata` exists, if so assume there's an existing install, otherwise continue under the assumption there isn't.
* If there is an existing install, check if it's one done through our install scripts, if not stop the installation process.
* If it is one of our installs, look for and attempt to use the updater script to update the existing install (exiting with an error message if that fails).

This also adds a switch called '--allow-duplicate-install' which will cause the install to continue if it would normally refuse to continue due to an existing install.

##### Component Name

- area/packaging

##### Test Plan

Due to only impacting the `kickstart.sh` scripts themselves, this can be tested by just using the copies of the scripts form this PR to run through the 'normal' usage.

##### Additional Information

- Fixes: #8198
- Fixes: #8186 (_specifically fixes the last issue outlined there_).

### DIFF
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -4,12 +4,13 @@
 # shellcheck disable=SC1117,SC2039,SC2059,SC2086
 #
 #  Options to run
-#  --dont-wait              do not wait for input
-#  --non-interactive        do not wait for input
-#  --dont-start-it          do not start netdata after install
-#  --stable-channel         Use the stable release channel, rather than the nightly to fetch sources
-#  --disable-telemetry      Opt-out of anonymous telemetry program (DO_NOT_TRACK=1)
-#  --local-files            Use a manually provided tarball for the installation
+#  --dont-wait                do not wait for input
+#  --non-interactive          do not wait for input
+#  --dont-start-it            do not start netdata after install
+#  --stable-channel           Use the stable release channel, rather than the nightly to fetch sources
+#  --disable-telemetry        Opt-out of anonymous telemetry program (DO_NOT_TRACK=1)
+#  --local-files              Use a manually provided tarball for the installation
+#  --allow-duplicate-install  do not bail if we detect a duplicate install
 #
 # Environment options:
 #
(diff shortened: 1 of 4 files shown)


Ferroin:
Marking this ready for review as it appears to work correctly based on local testing.

packaging/installer/kickstart-static64.sh
```diff
@@ -179,6 +180,65 @@ sudo=""
 [ -z "${UID}" ] && UID="$(id -u)"
 [ "${UID}" -ne "0" ] && sudo="sudo"
 
+# ---------------------------------------------------------------------------------------------------------------------
+# look for an existing install and try to update that instead if it exists
+
+ndpath="$(command -v netdata)"
```
prologic, line 186: You should redirect stderr to `/dev/null`/ here.

packaging/installer/kickstart-static64.sh
```diff
@@ -179,6 +180,65 @@ sudo=""
...
+      else
+        progress "Attempting to update existing kickstart install instead of creating a new one"
+        if run ${sudo} "${ndprefix}/usr/libexec/netdata/netdata-updater.sh" ; then
+          progress "Updated existing kickstart install at ${ndpath}"
+          exit 0
+        else
+          fatal "Failed to update existing Netdata install"
+          exit 1
+        fi
+      fi
```
prologic, line 222: In both cases, these end up doing the same thing, except for the "messaging". Is there a reason we need to duplicate code here? Do we wan t to special case one over the other at some point?

Ferroin: It's a holdover from before I properly handled the Netdata install prefix. We can merge these two cases if desired.

Issue #9114, opened 2020-05-20
Inconsistent information on Netdata Cloud


thoggs:
Hello people,
I know that the Netdata Cloud project is recent, and with that it is a fact that we will have some small problems; still, for the progress of the project, which I like a lot and, I use it daily; I would like to report what is happening to me:

**NOTE:** Occurs with machines running Linux and FreeBSD; both clean installations, from scratch, with no other software running, ie only Netdata is running.

**Problem ->** My machine (a server running FreeBSD), is showing up with inconsistent information that doesn't match what my hardware is (on the Dashboard), as shown in the following photo:

**NOTE: It is not a Virtual Machine, it is a physical machine**

![Captura de Tela 2020-05-20 às 14 36 18](https://user-images.githubusercontent.com/43221193/82479446-c9db5800-9aa8-11ea-8b9e-c1ae374d9195.png)

**OS:** FreeBSD 12.1 RELEASE - Standard installation: no lib32
**Netdata Version:** 1.22.1-74-nightly

thiagoftsm:
Hello @thoggs ,

@Ferroin could be a problem with our `daemon/system-info.sh` script?
@thoggs , please, could you run `daemon/system-info.sh` and verify if it is identifying your computer information correctly?

Best regards!

thoggs:
> Hello @thoggs ,
> 
> @Ferroin could be a problem with our `daemon/system-info.sh` script?
> @thoggs , please, could you run `daemon/system-info.sh` and verify if it is identifying your computer information correctly?
> 
> Best regards!

Hello @thiagoftsm,

Here is the output from the script `system-info.sh`:

![Captura de Tela 2020-05-20 às 15 06 01](https://user-images.githubusercontent.com/43221193/82481461-b8477f80-9aab-11ea-92e7-1f0468a71611.png)

thiagoftsm:
Thank you very much @thoggs !
Is the script output matching your hardware?

thoggs:
So in reality yes, but in parts, for example:

**NETDATA_SYSTEM_TOTAL_DISK_SIZE=3097889267712** (this will give 3 terabyte right?)
In this case, no, because I don't have 3 terabytes of disk installed on this hardware.

thiagoftsm:
Thank you for the additional information!

> NETDATA_SYSTEM_TOTAL_DISK_SIZE=3097889267712 (this will give 3 terabyte right?)
[...]

PR #9425, opened 2020-06-25
Corrected virtualization detection in system-info.sh.

Ferroin:
##### Summary

This updates our virtualization detection to report 'none' when it can't find any virtualization instead of reporting 'unknown', providing better consistency in the Cloud display of node information.

##### Component Name

area/daemon

##### Test Plan

Script tested locally.

##### Additional Information

Fixes: #9114


thiagoftsm:
Except the VM that has `dmidecode` installed, all the other tests I did with VirtualBOX and LXC were fine, but on a Fedora Core VM, I could not run `dmidecode` as a normal user, no less important when I ran as root I got this:

VirtualBox_Fedora_25_06_2020_23_38_43

Should we change `Virtual` to `VirtualBox`?

PR #9428, opened 2020-06-26
Fixed duplicate alarm ids in health-log.db

stelfrag:
##### Summary
The metadata log replay creates charts with linked alarms that may not reuse alarm_ids already present in the
heath-log.db. This will result in entries with different alarm ids for the same chart / name e.g. when shown 
via `http://localhost:19999/api/v1/alarm_log`, used in ACLK etc

To make sure we reuse the alarm id originally assigned and recorded in the `health-log.db` and `health-log.db.old` files process those files before metadata log replay

Add check to recover from corrupted files (`health-log.db` and `health-log.db.old`) that may already have different alarm_id for the
same chart / name

1. For each recorded alarm_id
    - If it exists in another chart / name already loaded, generate a new one

##### Component Name
health
ACLK

### DIFF
--- a/health/health_log.c
+++ b/health/health_log.c
@@ -394,7 +394,7 @@ inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char *filena
     if(!host->health_max_alarm_id)  host->health_max_alarm_id  = (uint32_t)now_realtime_sec();
 
     host->health_log.next_log_id = host->health_max_unique_id + 1;
-    if (unlikely(host->health_log.next_alarm_id && host->health_log.next_alarm_id <= host->health_max_alarm_id))
+    if (unlikely(!host->health_log.next_alarm_id || host->health_log.next_alarm_id <= host->health_max_alarm_id))
         host->health_log.next_alarm_id = host->health_max_alarm_id + 1;
 
     debug(D_HEALTH, "HEALTH [%s]: loaded file '%s' with %zd new alarm entries, updated %zd alarms, errors %zd entries, duplicate %zd", host->hostname, filename, loaded, updated, errored, duplicate);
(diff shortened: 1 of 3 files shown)


health/health_log.c
```diff
@@ -286,6 +302,8 @@ inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char *filena
             //    error("HEALTH [%s]: line %zu of file '%s' provides an alarm for host '%s' but this is named '%s'.", host->hostname, line, filename, pointers[1], host->hostname);
 
             ae->unique_id               = unique_id;
+            if (!is_valid_alarm_id(host, pointers[14], pointers[13], alarm_id))
+                alarm_id = rrdcalc_get_unique_id(host, pointers[14], pointers[13], NULL);
```
mfundul, line 306: `rrdcalc_get_unique_id()` will eventually execute `host->health_log.next_alarm_id++`. Is this a problem?

mfundul: I mean, is this going to create collisions?

stelfrag: That should not be a problem by itself. There is code however that will force `host->health_log.next_alarm_id` host->health_log.next_alarm_id to the maximum found (L397) which would be bad in case one has been assigned. I will check that as well.

amoss: What about concurrency - which threads can end up calling `rrdcalc_get_unique_id()` and do we need to guard the counter with a mutex as @underhood suggested?

thiagoftsm: The function `health_alarm_log_read` calls `netdata_rwlock_rdlock(&host->health_log.alarm_log_rwlock);` before to load the file, at least for this function we should not have problems.

thiagoftsm:
@stelfrag I observed with the tests that I made, the `alarm_id` has not been used to create new alarms. I will wait for possible commits addressing @mfundul and @amoss concerns to retest before to approve.

PR #9434, opened 2020-06-27
Fix internal registry 

### LINKED ISSUE #9427: "registry to announce" not honoured in alarm-notify.sh
With the latest release (1.23.0), the url in the notification always points to https://app.netdata.cloud (and, btw it goes to a 404 page) even with "registry to announce" set.
I see in alarm-notify.sh that it overrides the local registry url if the file /var/lib/netdata/registry/netdata.public.unique.id exists.
I tried to remove that file but it is regenerated as soon as netdata restart.
In the previous version the notifications used the correct, internal, url.

thiagoftsm:
##### Summary
Fixes #9427 

The spawn server was initialized before the registry to be initialized, but environment variables set for the function `registry_init()` must be set before the spawn server to be called. This PR moves the registry initialization position to fix this bug. 

##### Component Name
Spawn Server
##### Test Plan
1 - Set registry server inside `netdata.conf`:

```
[registry]
        enabled = yes
        registry db directory = /var/lib/netdata/registry
        registry to announce = https://localhost:19999
        registry hostname = hades
```

2 - Create an alarm inside `/etc/netdata/health.d`, for example:

```
 alarm: dev_dim_template
    on: system.cpu
    os: linux
 units: %
 every: 1s
  warn: $this > 1
  crit: $this > 4
```

3 - Configure `health_alarm_notify.conf` to send the alarm for a channel.
4 - Start Netdata
5 - Verify if the link sent with alarm has the `registry to announce` in the URL `https://localhost:19999/goto-host-from-alarm.html?host=hades&chart=system.cpu&family=cpu&alarm=dev_dim_template_user&alarm_unique_id=1593269432&alarm_id=1593269248&alarm_event_id=7&alarm_when=1593269205`

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -1386,6 +1386,7 @@ int main(int argc, char **argv) {
     netdata_threads_init_after_fork((size_t)config_get_number(CONFIG_SECTION_GLOBAL, "pthread stack size", (long)default_stacksize));
 
     // fork the spawn server
+    registry_init();
     spawn_init();
     /*
      * Libuv uv_spawn() uses SIGCHLD internally:
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -1385,8 +1385,9 @@ int main(int argc, char **argv) {
 
     netdata_threads_init_after_fork((size_t)config_get_number(CONFIG_SECTION_GLOBAL, "pthread stack size", (long)default_stacksize));
 
-    // fork the spawn server
+    // initialyze internal registry
     registry_init();
+    // fork the spawn server
     spawn_init();
     /*
      * Libuv uv_spawn() uses SIGCHLD internally:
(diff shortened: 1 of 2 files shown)


daemon/main.c
```diff
@@ -1386,6 +1386,7 @@ int main(int argc, char **argv) {
     netdata_threads_init_after_fork((size_t)config_get_number(CONFIG_SECTION_GLOBAL, "pthread stack size", (long)default_stacksize));
 
     // fork the spawn server
+    registry_init();
```
mfundul, line 1389: Can you move this 1 line above so that the comment matches the spawn server?

thiagoftsm: Of course, I moved it and I added a proper comment.

daemon/main.c
```diff
@@ -1385,6 +1385,8 @@ int main(int argc, char **argv) {
 
     netdata_threads_init_after_fork((size_t)config_get_number(CONFIG_SECTION_GLOBAL, "pthread stack size", (long)default_stacksize));
 
+    // initialyze internal registry
```
mfundul, line 1388: typo: `initialyze` -> `initialize`

Issue #9423, opened 2020-06-25
DEB and RPM missing eBPF modules


thiagoftsm:
##### Bug report summary
As discussed on the issue #9221 the current packages we are delivering are not properly installing the eBPF programs necessary to run eBPF plugin.

##### OS / Environment
All distributions that has packages like DEB and RPM.

##### Netdata version
Netdata 1.23.0

##### Component Name
Packaging

##### Steps To Reproduce

1. Install Netdata using apt or yum

##### Expected behavior
Do not have problems like this:

```
2020-06-25 11:54:41: ebpf.plugin INFO  : EBPF SOCKET : Cannot load library /usr/libexec/netdata/plugins.d/libnetdata_ebpf.so.4.15.0-88-generic
ՠ for the current kernel.
2020-06-25 11:54:41: ebpf.plugin ERROR : EBPF SOCKET : Cannot load /usr/libexec/netdata/plugins.d/libnetdata_ebpf.so.4.16.18 default library.
2020-06-25 11:54:41: ebpf.plugin INFO  : EBPF SOCKET : thread with task id 7420 finished
2020-06-25 11:54:41: ebpf.plugin INFO  : EBPF PROCESS : Cannot load library /usr/libexec/netdata/plugins.d/libnetdata_ebpf.so.4.16.18 for the current kernel.
2020-06-25 11:54:41: ebpf.plugin ERROR : EBPF PROCESS : Cannot load /usr/libexec/netdata/plugins.d/libnetdata_ebpf.so.4.16.18 default library.
2020-06-25 11:54:42: ebpf.plugin INFO  : EBPF PROCESS : thread with task id 7419 finished
```

PR #9450, opened 2020-06-30
Properly include eBPF collector in binary packages.

Ferroin:
##### Summary

This adds the required logic to properly bundle the eBPF collector code in our binary DEB and RPM packages.

##### Component Name

area/packaging

##### Test Plan

Package builds locally verified using Docker

Packages verified by installing on clean VM instances and then confirming that the required files are where they're supposed to be.

##### Additional Information

This is intentionally based on master and not develop as it's assumed it will not be merged prior to the patch release that's occurring tomorrow.

Fixes: #9423

### DIFF
--- a/packaging/bundle-ebpf.sh
+++ b/packaging/bundle-ebpf.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+SRCDIR="${1}"
+PLUGINDIR="${2}"
+
+EBPF_VERSION="$(cat "${SRCDIR}/packaging/ebpf.version")"
+EBPF_TARBALL="netdata-kernel-collector-glibc-${EBPF_VERSION}.tar.xz"
+
+mkdir -p "${SRCDIR}/tmp/ebpf"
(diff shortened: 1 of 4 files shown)


packaging/bundle-ebpf.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+SRCDIR="${1}"
+PLUGINDIR="${2}"
+
+EBPF_VERSION="$(cat "${SRCDIR}/packaging/ebpf.version")"
+EBPF_TARBALL="netdata-kernel-collector-glibc-${EBPF_VERSION}.tar.xz"
+
+mkdir -p "${SRCDIR}/tmp/ebpf"
```
thiagoftsm, line 9: I have only one doubt for this package and also the dashboard,  would be better to use mktemp instead mkdir to create the directories?

Ferroin: Normally I would agree, but because these are running in a build environment that gets recreated each time we try to build the packages, it's simpler to just use fixed paths here.

prologic:
Has anyone else tested this? cc @netdata/agent

MrFreezeex:
I installed this patch on ubuntu 18 and the ebpf.plugin does start properly when executing it. I also have a "net" chart under "Applications". I think this comes from the ebpf.plugin, so It works at least on my system :)

PR #9429, opened 2020-06-26
Add guide for monitoring a k8s cluster with Netdata

joelhans:
##### Summary

Title is fairly self-explanatory. My idea is that k8s users first use the install instructions I'm hoping to add with #9364, then hop over to this guide to explore the various charts that they can now access once they've installed the Helm chart. I'm hoping it's helpful to explain what each of these sections does and what the capabilities are for each, as the UX around k8s isn't very refined right now.

This guide depends on and links to the file added in #9364, hence the broken link-check CI action. As long as we merge #9364 first and rebase this PR on top of that, the links should resolve OK.

##### Component Name

area/docs

PR #9466, opened 2020-07-02
Add guide for monitoring a k8s cluster with Netdata

joelhans:
##### Summary

Same content as #9429, but without the mangled Git history.

@ncmans & @knatsakis: I made one more commit after you both approved, to link from the install doc to the guide, and then proceeded to accidentally add a bunch of unrelated commits and couldn't figure out how to clean up the history.

> Title is fairly self-explanatory. My idea is that k8s users first use the install instructions I'm hoping to add with #9364, then hop over to this guide to explore the various charts that they can now access once they've installed the Helm chart. I'm hoping it's helpful to explain what each of these sections does and what the capabilities are for each, as the UX around k8s isn't very refined right now.

> This guide depends on and links to the file added in #9364, hence the broken link-check CI action. As long as we merge #9364 first and rebase this PR on top of that, the links should resolve OK.

##### Component Name

area/docs

### DIFF
--- a/docs/guides/monitor/kubernetes-k8s-netdata.md
+++ b/docs/guides/monitor/kubernetes-k8s-netdata.md
@@ -1,0 +1,275 @@
+<!--
+title: "Monitor a Kubernetes (k8s) cluster with Netdata"
+description: "Use Netdata's helmchart, service discovery plugin, and Kubelet/kube-proxy collectors for real-time visibility into your Kubernetes cluster."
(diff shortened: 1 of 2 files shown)


Ferroin:
Travis passed but never reported back. I've restarted the build there in the hopes that it will work this time. If it doesn't, then somebody who has admin privileges on the repo can just merge this without waiting for Travis.

PR #9474, opened 2020-07-04
cgroups.plugin: add and document support for reading container names from Podman

### LINKED ISSUE #7391: cgroups.plugin should parse for podman containers
Before [Podman](https://podman.io/) inevitably replaces Docker as the standard for containers, netdata's cgroups plugin should be able to parse container names. The `podman ps` command [behaves pretty much just like Docker's](https://www.mankier.com/1/podman-ps).
There might be other stuff in netdata about this compatibility I'm forgetting about.

Related to https://github.com/netdata/netdata/issues/4868

K900:
##### Summary
Fixes #7391 (somewhat). 

This is not as seamless as I'd like it to be, but Podman really doesn't want other people to look at your containers (and for good reason), so this is probably as much as we can do without rewriting the entire mechanism. I've also added a section to the documentation describing the issue and possible solutions for now.

On the implementation side, Podman's API both on the command line tool and on the Unix socket is intentionally identical to Docker's, so most of the Docker logic can be reused.

##### Component Name
cgroups.plugin

##### Test Plan

Tested on a server running Podman 2.0.1 and Netdata 1.23.1 with this patch applied, both with `docker-socket-proxy` and changed socket permissions.

### DIFF
--- a/collectors/cgroups.plugin/README.md
+++ b/collectors/cgroups.plugin/README.md
@@ -105,9 +105,9 @@ For this mapping Netdata provides 2 configuration options:
 
 The whole point for the additional pattern list, is to limit the number of times the script will be called. Without this pattern list, the script might be called thousands of times, depending on the number of cgroups available in the system.
 
-The above pattern list is matched against the path of the cgroup. For matched cgroups, Netdata calls the script [cgroup-name.sh](https://raw.githubusercontent.com/netdata/netdata/master/collectors/cgroups.plugin/cgroup-name.sh.in) to get its name. This script queries `docker`, `kubectl`, `podman` or applies heuristics to find give a name for the cgroup.
+The above pattern list is matched against the path of the cgroup. For matched cgroups, Netdata calls the script [cgroup-name.sh](https://raw.githubusercontent.com/netdata/netdata/master/collectors/cgroups.plugin/cgroup-name.sh.in) to get its name. This script queries `docker`, `kubectl`, `podman`, or applies heuristics to find give a name for the cgroup.
 
-#### a note on Podman container names
+#### Note on Podman container names
 
 Podman's security model is a lot more restrictive than Docker's, so Netdata will not be able to detect container names out of the box unless they were started by the same user as Netdata itself.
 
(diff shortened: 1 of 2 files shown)


prologic:
cc @joelhans

collectors/cgroups.plugin/README.md
```diff
@@ -105,7 +105,15 @@ For this mapping Netdata provides 2 configuration options:
 
 The whole point for the additional pattern list, is to limit the number of times the script will be called. Without this pattern list, the script might be called thousands of times, depending on the number of cgroups available in the system.
 
-The above pattern list is matched against the path of the cgroup. For matched cgroups, Netdata calls the script [cgroup-name.sh](https://raw.githubusercontent.com/netdata/netdata/master/collectors/cgroups.plugin/cgroup-name.sh.in) to get its name. This script queries `docker`, or applies heuristics to find give a name for the cgroup.
+The above pattern list is matched against the path of the cgroup. For matched cgroups, Netdata calls the script [cgroup-name.sh](https://raw.githubusercontent.com/netdata/netdata/master/collectors/cgroups.plugin/cgroup-name.sh.in) to get its name. This script queries `docker`, `kubectl`, `podman` or applies heuristics to find give a name for the cgroup.
```
joelhans, line 108:
```suggestion
The above pattern list is matched against the path of the cgroup. For matched cgroups, Netdata calls the script [cgroup-name.sh](https://raw.githubusercontent.com/netdata/netdata/master/collectors/cgroups.plugin/cgroup-name.sh.in) to get its name. This script queries `docker`, `kubectl`, `podman`, or applies heuristics to find give a name for the cgroup.
```

collectors/cgroups.plugin/README.md
```diff
@@ -105,7 +105,15 @@ For this mapping Netdata provides 2 configuration options:
...
+
+#### a note on Podman container names
```
joelhans, line 110:
```suggestion
#### Note on Podman container names
```

K900:
Applied both. No particular opinion on Oxford commas, and I was following the other headers in the file re: case.

PR #9489, opened 2020-07-06
Implemented the HOST command in metadata log replay

### LINKED ISSUE #9453: Support the HOST command in the metadata log replay
#### Summary

- Add support for the HOST command to the metadata log parser.
- Add the HOST GUID to the global GUID map
  - Compaction should work properly
- Add new archived flag for the HOST
  - Start a host in archived mode and reset when collection starts
  - Metadata log replay creates a host in archived mode
    - Charts are created in archived mode during metadata log replay
    - A collected chart will flag the chart and host as not archived

stelfrag:
Fixes #9453 
##### Summary
Register the HOST command in the metadata log parser and the corresponding action. Improve the detection of the
HOST GUID type (correctly registered in the global GUID map now) and make sure that during compaction the HOST
is written to the metadata log files before any CHART / DIMENSION

##### Component Name
database

##### Test Plan
- Start agent and wait for compaction to run
- Example the metadata log files using

`cat /home/stelios/netdata-inst/netdata//var/cache/netdata/dbengine/metadatalog-* | strings -n | head -n 30`

You should always observe the HOST command at the top e.g. (sample)

```
HOST "d6b37186-74db-11ea-88b2-0bf5095b1f9e" "futura" "futura" 1 "linux" "Europe/Athens" ""
LABEL "_os_name" = 0 Ubuntu
LABEL "_os_version" = 0 20.04 LTS (Focal Fossa)
LABEL "_kernel_version" = 0 5.4.0-40-generic
LABEL "_system_cores" = 0 12
LABEL "_system_cpu_freq" = 0 2600000
LABEL "_system_ram_total" = 0 16624041984
LABEL "_system_disk_space" = 0 1000204886016
LABEL "_architecture" = 0 x86_64
LABEL "_virtualization" = 0 none
LABEL "_container" = 0 unknown
LABEL "_container_detection" = 0 none
LABEL "_virt_detection" = 0 systemd-detect-virt
LABEL "_is_parent" = 0 true
OVERWRITE labels
CONTEXT d6b37186-74db-11ea-88b2-0bf5095b1f9e
GUID ab5de6e8-0303-40e4-8ade-cf6a5828586d
CHART "system.idlejitter" "" "CPU Idle Jitter" "microseconds lost/s" "idlejitter" "system.idlejitter" "area" 800 1 "   " "idlejitter.plugin" ""
GUID 0692372e-0ce9-f9a2-78ac-4208c9a74c4a
DIMENSION "min" "min" "absolute" 1 1 "  "
GUID 1bc106f9-8a10-376e-c7c2-492ba52367b8
DIMENSION "max" "max" "absolute" 1 1 "  "
GUID 566fd00e-a432-16d1-9090-b3b0b9255b1e
DIMENSION "average" "average" "absolute" 1 1 "  "
CONTEXT d6b37186-74db-11ea-88b2-0bf5095b1f9e
GUID 8130c6aa-15bf-4c15-9f91-a8b2f19e618a
CHART "netdata.statsd_metrics" "" "Metrics in the netdata statsd database" "metrics" "statsd" "netdata.statsd_metrics" "stacked" 132010 1 "   " "statsd.plugin" "stats"
GUID 83d940f6-4e96-ddfc-211f-92e728bc0eb5
DIMENSION "gauges" "gauges" "absolute" 1 1 "  "
```

### DIFF
--- a/database/engine/metadata_log/compaction.c
+++ b/database/engine/metadata_log/compaction.c
@@ -129,8 +129,9 @@ static void compact_record_by_uuid(struct metalog_instance *ctx, uuid_t *uuid)
             }
             break;
         case GUID_TYPE_HOST:
-            error_with_guid(uuid, "Compaction: Processing HOST record");
-            RRDHOST *this_host = metalog_get_host_from_uuid(ctx, uuid);
+            RRDHOST *this_host = ctx->rrdeng_ctx->host;
+            //TODO: will be enabled when multidb is activated
+            //RRDHOST *this_host = metalog_get_host_from_uuid(ctx, uuid);
             if (ctx->current_compaction_id > this_host->compaction_id) {
                 this_host->compaction_id = ctx->current_compaction_id;
                 buffer = metalog_update_host_buffer(this_host);
--- a/database/engine/metadata_log/compaction.c
+++ b/database/engine/metadata_log/compaction.c
@@ -118,12 +118,6 @@ static void compact_record_by_uuid(struct metalog_instance *ctx, uuid_t *uuid)
                     error("Forcing compaction of CHART %s", rd->rrdset->id);
                     compact_record_by_uuid(ctx, rd->rrdset->chart_uuid);
                 }
-
-                if (strcmp(rd->id, "iwlwifi_1-virtual-0_temp1") == 0) {
-                    error("SKIPPING %s on purpose (chart = %s)", rd->id, rd->rrdset->id);
-                   break;
-                }
-
                 if (ctx->current_compaction_id > rd->state->compaction_id) {
                     rd->state->compaction_id = ctx->current_compaction_id;
                     buffer = metalog_update_dimension_buffer(rd);
--- a/database/engine/metadata_log/compaction.c
+++ b/database/engine/metadata_log/compaction.c
@@ -79,7 +79,6 @@ static void compaction_test_quota(struct metalog_worker_config *wc)
 static void compact_record_by_uuid(struct metalog_instance *ctx, uuid_t *uuid)
 {
     GUID_TYPE ret;
-    RRDHOST *host = ctx->rrdeng_ctx->host;
     RRDSET *st;
     RRDDIM *rd;
     BUFFER *buffer;
@@ -92,6 +91,11 @@ static void compact_record_by_uuid(struct metalog_instance *ctx, uuid_t *uuid)
         case GUID_TYPE_CHART:
             st = metalog_get_chart_from_uuid(ctx, uuid);
             if (st) {
+                if (ctx->current_compaction_id > st->rrdhost->compaction_id) {
+                    error("Forcing compaction of HOST %s from CHART %s", st->rrdhost->hostname, st->id);
+                    compact_record_by_uuid(ctx, &st->rrdhost->host_uuid);
+                }
+
                 if (ctx->current_compaction_id > st->compaction_id) {
                     st->compaction_id = ctx->current_compaction_id;
                     buffer = metalog_update_chart_buffer(st, ctx->current_compaction_id);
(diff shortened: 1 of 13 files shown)


database/engine/metadata_log/compaction.c
```diff
@@ -106,12 +110,15 @@ static void compact_record_by_uuid(struct metalog_instance *ctx, uuid_t *uuid)
...
                 if (ctx->current_compaction_id > rd->rrdset->compaction_id) {
-                    error("Forcing compaction of chart %s", rd->rrdset->id);
-                    rd->rrdset->compaction_id = ctx->current_compaction_id;
-                    buffer = metalog_update_chart_buffer(rd->rrdset, ctx->current_compaction_id);
-                    metalog_commit_record(ctx, buffer, METALOG_COMMIT_CREATION_RECORD, rd->rrdset->chart_uuid, 1);
-                } else if (ctx->current_compaction_id > rd->state->compaction_id) {
+                    error("Forcing compaction of CHART %s", rd->rrdset->id);
+                    compact_record_by_uuid(ctx, rd->rrdset->chart_uuid);
+                }
+                if (ctx->current_compaction_id > rd->state->compaction_id) {
```
mfundul, line 121:
```suggestion
                else if (ctx->current_compaction_id > rd->state->compaction_id) {
```
This should be `else if` as before so that the dimension record is not dumped into the log twice, once from the chart dump and another one here. This is important so that the compacted log is optimal and we don't enter an infinite loop of compactions.

stelfrag: True no need to do the check in that case, I will change it.

It would not be a problem with duplicate dimension records though because the `rd->state->compaction_id` would have been updated

stelfrag: changed

database/engine/metadata_log/compaction.c
```diff
@@ -123,9 +129,11 @@ static void compact_record_by_uuid(struct metalog_instance *ctx, uuid_t *uuid)
             }
             break;
         case GUID_TYPE_HOST:
-            if (ctx->current_compaction_id > host->compaction_id) {
-                host->compaction_id = ctx->current_compaction_id;
-                buffer = metalog_update_host_buffer(host);
+            error_with_guid(uuid, "Compaction: Processing HOST record");
+            RRDHOST *this_host = metalog_get_host_from_uuid(ctx, uuid);
```
mfundul, line 133: You don't know here if `this_host` is `NULL` or even a different pointer than `host`. How do we resolve this? Maybe fallback to the old code for now, so that we can merge to master?

database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -4,6 +4,48 @@
...
+{
+    UNUSED(user);
+
+    if (strcmp(machine_guid, registry_get_this_machine_guid()) == 0)
+        return PARSER_RC_OK;
+
+    RRDHOST *host = rrdhost_find_by_guid(machine_guid, 0);
+    if (host)
+        return PARSER_RC_OK;
+
```
mfundul, line 19: We need code similar to `metalog_pluginsd_chart_action()` and `metalog_pluginsd_dimension_action()` like this:
```
    if (host) { /* It's a valid object */
        struct metalog_record record;
        struct metadata_logfile *metalogfile = state->metalogfile;

        uuid_copy(record.uuid, state->uuid);
        mlf_record_insert(metalogfile, &record);
        uuid_clear(state->uuid); /* Consume UUID */
    }

```

This is needed so that during log replay, the necessary `struct metalog_record` is created so that afterwards compaction won't miss it.

database/engine/metadata_log/compaction.c
stelfrag: Eventually it could be a different host (null would be an error).  I can skip the call and use `ctx->rrdeng_ctx->host` to fallback to the old code.

database/engine/metadata_log/metalogpluginsd.c
stelfrag: You are right. At the comment there is extra check during compaction so that if a chart compaction is attempted without the host being compacted -- it is forced -- I will add it

database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -4,6 +4,63 @@
...
+    void *user, char *machine_guid, char *hostname, char *registry_hostname, int update_every, char *os, char *timezone,
+    char *tags)
+{
+    struct metalog_pluginsd_state *state = ((PARSER_USER_OBJECT *)user)->private;
+
+    RRDHOST *host = rrdhost_find_by_guid(machine_guid, 0);
+    if (host)
+        goto write_replay;
+
+    if (strcmp(machine_guid, registry_get_this_machine_guid()) == 0) {
```
mfundul, line 17: The `registry_get_this_machine_guid()` function opens or creates a file under `var/lib/netdata`, and at this point I think we don't want to do that.

database/rrdhost.c
```diff
@@ -271,11 +276,15 @@ RRDHOST *rrdhost_create(const char *hostname,
         health_alarm_log_open(host);
     }
 
-    if (host->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
+    if (host && host->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
```
mfundul, line 279: Why do we change this? Just above there is `    if(host->health_enabled) {` which would crash if `host` was NULL.

stelfrag: No need I will remove it. The initialization moved up from a previous commit and it could have ended up being NULL

database/engine/metadata_log/metalogpluginsd.c
stelfrag: It has already been called at this point (before the metalog_init)

database/engine/metadata_log/metalogpluginsd.c
```diff
@@ -206,6 +263,33 @@ PARSER_RC metalog_pluginsd_tombstone_action(void *user, uuid_t *uuid)
...
+    char *timezone = words[6];
+    char *tags = words[7];
+
+    int update_every = 1;
+    if (likely(update_every_s && *update_every_s))
+        update_every = str2i(update_every_s);
+    if (unlikely(!update_every))
+        update_every = 1;
+
+    info("HOST PARSED: guid=%s, hostname=%s, reg_host=%s, update=%d, os=%s, timezone=%s, tags=%s",
```
mfundul, line 282: Since the language parsing is supposed to be common this function should be in `pluginsd_parser.c` like the other keywords of metadata log:
https://github.com/netdata/netdata/blob/42aa54eaf6a64348f23cc0c9ed0a071b79bc9cad/collectors/plugins.d/pluginsd_parser.c#L608

```suggestion
    debug(D_PLUGINSD, "HOST PARSED: guid=%s, hostname=%s, reg_host=%s, update=%d, os=%s, timezone=%s, tags=%s",
```

stelfrag:
Rebase / handle conflicts with #9520

PR #9504, opened 2020-07-08
Implemented default disk space size calculation for multihost db

### LINKED ISSUE #9452: Calculate dbegine storage size for the localhost multidb
#### Summary
Add a new `dbengine multihost disk space` parameter under the `[global]` section in `netdata.conf`

If defined it will be the maximum size the dbegine files of the locahost multidb will occupy. 

The first time an agent starts it will try to define a maximum size for the dbengine files to use as a default. The agent will try to detect legacy dbengine folders and calculates the aggregate disk space using 

`number of folders * dbengine disk space`  for disk space

where `dbengine disk space` is taken from he `[global]` section in `netdata.conf`. 

Note: that `dbengine disk space` defaults to a minimum of 64MB

The results of the calculation are stored under `var/lib/netdata/dbengine_multihost_size` in megabytes

---
#### Agent startup sequence after upgrading
1. `var/lib/netdata/dbengine_multihost_size` exists
    - Read the stored size and check for inconsistencies
      - Size cannot be less than `dbengine disk space`
    - Use the value as a default when reading the `dbengine multihost disk space` configuration option from `netdata.conf`
1. `var/lib/netdata/dbengine_multihost_size` does not exist
    - Run the calculation to produce the `var/lib/netdata/dbengine_multihost_size` file
    - If no legacy dbengine folders are found the recorded size will be the `dbengine disk space` default

stelfrag:
Fixes #9452
##### Summary
Calculates the initial default size for the multihost database and stores in under

`<installation_directory>/var/lib/netdata/dbengine_multihost_size`

The computed value is in MBytes

The calculation will happen only once (unless the file is deleted) and contains the default size for the multihost db
The user defined size is defined with the new `dbengine multihost disk space` parameter under the `[global]` section in `netdata.conf`

##### Component Name
database

##### Test Plan
- Start the agent 
- A file will be created `<installation_directory>/var/lib/netdata/dbengine_multihost_size`
  **Note: This file is created yet. The code that will actually create the file will be activated when multidb is enabled**
- If X are the number of dbengine / children folders under `<installation_directory>/var/cache/netdata` the calculated value in the file will be `(X + 1) * dbengine disk space`

##### Additional Information
Only the folders that have a valid UUID as a name under `<installation_directory>/var/cache/netdata` and contain a `dbengine`  folder are counted. eg

`/var/cache/netdata/e574ee06-6e83-11ea-98ac-b827eb6f70a8/dbengine`


database/engine/rrdengineapi.c
```diff
@@ -6,6 +6,8 @@ static struct rrdengine_instance default_global_ctx;
 
 int default_rrdeng_page_cache_mb = 32;
 int default_rrdeng_disk_quota_mb = 256;
+int computed_multidb_disk_quota_mb = 256;
```
mfundul, line 9: This global variable `computed_multidb_disk_quota_mb` is pretty much not used in a global scope.

It's initialized again inside `compute_multidb_diskspace()`, and it gets a value of `256` again inside `compute_multidb_diskspace()`, and its value could be returned by `compute_multidb_diskspace()` locally since it's used immediately afterwards.

daemon/main.c
```diff
@@ -543,6 +543,8 @@ static void get_netdata_configured_variables() {
     // ------------------------------------------------------------------------
     // get default Database Engine page cache size in MiB
 
+    compute_multidb_diskspace();
```
mfundul, line 546: This function needs to be called after `default_rrdeng_disk_quota_mb` has been set below in line 561.

database/engine/rrdenginelib.c
```diff
@@ -208,3 +208,83 @@ char *get_rrdeng_statistics(struct rrdengine_instance *ctx, char *str, size_t si
...
+//                fprintf(fp, "%d", rc * default_rrdeng_disk_quota_mb);
+//                info("Created file '%s' to store the computed value", multidb_disk_space_file);
+//                fclose(fp);
+//            } else
+//                error("Failed to store the default multidb disk quota size on '%s'", multidb_disk_space_file);
+        }
+    }
+
+    if (unlikely(computed_multidb_disk_quota_mb == - 1))
+        computed_multidb_disk_quota_mb = 256;
```
mfundul, line 289: If there is a valid localhost then we shouldn't override the `default_rrdeng_disk_quota_mb` value, which can be much bigger than 256.

database/engine/rrdengineapi.c
stelfrag: True !! `compute_multidb_diskspace` will now return the computed space

daemon/main.c
stelfrag: True will use any user supplied value for the calculations

database/engine/rrdenginelib.c
stelfrag: This will change to use the `default_rrdeng_disk_quota_mb`.  The `compute_multidb_diskspace` is now called after reading the configuration from the `netdata.conf` file.

database/engine/rrdengineapi.c
mfundul: You forgot to delete this from here and in `rrdengineapi.h`!

stelfrag: Removed the variable `computed_multidb_disk_quota_mb` 

In rrdengineapi.h I have the `default_multidb_disk_quota_mb` which is needed

mfundul: Yeap, I'm not seeing clearly I guess!

underhood:
@stelfrag LGTM but just to be picky - minor comment -> can you add those 2 missing endlines at the end of file?

stelfrag:
Added missing newlines

PR #9520, opened 2020-07-10
Remove health from archived metrics

mfundul:
##### Summary
Fixes #9457
##### Component Name
database
health
##### Test Plan
Have users that originally reported the issue verify it is fixed.

##### Additional Information
Disassociate health variables and alarms from archived charts and dimensions.

PR #9536, opened 2020-07-14
Fix transition from archived to active charts not generating alarms.

### LINKED ISSUE #9534: Most alarms are disappearing after restart and "phantom" disks
##### Bug report summary
I have noticed that only few alarms (from usually long list) are visible (in "All alarms" tab), so I tried to investigate what happens.

Deletion of dbengine data solved the issue... till netdata was restarted again - after restart I could see only fraction of all alarms:

image

That's all - subsequent restart didn't help, but deletion of data helps - once, until next restart.

Disabling all plugins didn't help too - the behavior was still the same.

Additional issue was related to deleted LVM devices - once deleted, they are still shown in netdata (even after restart and full browser refresh), and they even show some activity(!), though they are definitely do not exist anymore in the system - but probably this is unrelated and is worth dedicated ticket.

##### OS / Environment
```
Linux nuff 5.4.44-2-pve #1 SMP PVE 5.4.44-2
Debian 10 (Proxmox VE 6.2-9)
```

##### Netdata version
v1.23.1-77-g013c9da6

##### Steps To Reproduce
1. Remove all dbengine data files
2. Run netdata with this config:
```
[global]
        history = 90000
        memory mode = dbengine
        page cache size = 512
        dbengine disk space = 2048
        access log = none

[plugins]
        enable running new plugins = no
```
3. Check metrics and alarms
4. Restart
5. Check again

If the issue is not reproducible I could provide collected dbengine files (1st and 2nd run) and error logs (though there is nothing unusual).

Not sure if this is related to my hardware configuration or something else, but the observed problem is stable on at least two systems - mentioned above (real hardware) and on up-to-date Ubuntu 18.04 (QEMU guest) with completely fresh (first time) netdata install using kickstart script. Both systems are idling and have plenty of unused resources, so this is not RAM or disk space related.

Stable version (1.23.1) does not have this issue.

mfundul:
##### Summary
Fixes #9534
##### Component Name
health
database
##### Test Plan

- stop netdata
- delete dbengine data
- start netdata
- check all alarms
- restart netdata
- check all alarms

### DIFF
--- a/database/rrdset.c
+++ b/database/rrdset.c
@@ -534,9 +534,9 @@ RRDSET *rrdset_create_custom(
     char fullid[RRD_ID_LENGTH_MAX + 1];
     snprintfz(fullid, RRD_ID_LENGTH_MAX, "%s.%s", type, id);
 
+    int changed_from_archived_to_active = 0;
     RRDSET *st = rrdset_find_on_create(host, fullid);
     if (st) {
-        int changed_from_archived_to_active = 0;
         int mark_rebuild = 0;
         rrdset_flag_set(st, RRDSET_FLAG_SYNC_CLOCK);
         rrdset_flag_clear(st, RRDSET_FLAG_UPSTREAM_EXPOSED);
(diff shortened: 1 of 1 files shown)


ilyam8:
@mfundul do we need to include in the test plan test from https://github.com/netdata/netdata/pull/9520?

mfundul:
> 
> 
> @mfundul do we need to include in the test plan test from #9520?

The changes here only affect netdata after health has been reloaded so I would say no.

ilyam8:
I did same as described in https://github.com/netdata/netdata/issues/9534#issuecomment-658039510

The issue is fixed.

PR #9540, opened 2020-07-14
Fixed stored number accuracy

### LINKED ISSUE #6839: Netdata intermittently reporting full disk space usage for a few minutes
##### Bug report summary
About once every few days day I'll get an alert about the storage being full followed shortly after by it being reported correctly and closing the alert.

We have Netdata running on 40+ servers and this is the only one showing this behaviour. 
It's been doing doing this since September 4th 10AM AEST

##### OS / Environment
Debian Stretch on AWS EC2 T3

v1.17.0-49-nightly

Example Today:
chrome_2019-09-13_14-46-42

Earlier this week:
image

image

stelfrag:
##### Summary
Fixes #6839
Improves stored number accuracy

##### Component Name
agent

### DIFF
--- a/libnetdata/storage_number/storage_number.c
+++ b/libnetdata/storage_number/storage_number.c
@@ -61,6 +61,10 @@ storage_number pack_storage_number(calculated_number value, uint32_t flags) {
             m++;
         }
 
+        if (unlikely(n > (calculated_number) (0xffffff))) {
+            n /= 10;
+            m--;
+        }
         // the value was small enough and we multiplied it
         // so we add a divider to unpack it
         r += (0 << 30) + (m << 27); // the divider m
--- a/libnetdata/storage_number/storage_number.c
+++ b/libnetdata/storage_number/storage_number.c
@@ -61,7 +61,7 @@ storage_number pack_storage_number(calculated_number value, uint32_t flags) {
             m++;
         }
 
-        if (unlikely(n > (calculated_number) (0xffffff))) {
+        if (unlikely(n > (calculated_number) (0x00ffffff))) {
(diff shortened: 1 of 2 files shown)


amoss:
Fix works but it looks strange. Why not just change the constant to 0x199999 so that it preserves the intent but fixes the overflow?

mfundul:
> 
> 
> Fix works but it looks strange. Why not just change the constant to 0x199999 so that it preserves the intent but fixes the overflow?

That should work as well, unless there is some really messed up error error accumulation in some architectures that may break this. Maybe some non-x86 32-bit one, I don't know.

mfundul:
> 
> 
> Fix works but it looks strange. Why not just change the constant to 0x199999 so that it preserves the intent but fixes the overflow?

The other problem with changing the constant is that we break compatibility with what's already stored on disk.

amoss:
Well...` 0x199999 * 0x0a = FFFFFA` and we are operating within the lower 24-bits of a 32-bit word so I don't think that overflow can affect this. The semantics on disk are that we multiply/divide the stored value by 10 the given number of times - so the constant is only implied as a boundary condition and changing it will not affect the disk storage.

mfundul:
> 
> 
> Well...` 0x199999 * 0x0a = FFFFFA` and we are operating within the lower 24-bits of a 32-bit word so I don't think that overflow can affect this. The semantics on disk are that we multiply/divide the stored value by 10 the given number of times - so the constant is only implied as a boundary condition and changing it will not affect the disk storage.

Storage shouldn't be affected, you are right, just looked at the `unpack()` function again.

PR #9546, opened 2020-07-15
Fix SHA256 handling in eBPF bundling code.

Ferroin:
##### Summary

This makes it compatible both with busybox and older GNU userspaces.

##### Component Name

area/packaging

##### Test Plan

Verified locally by building in a CentOS 7 environment.

##### Additional Information

@prologic This will fix the currently broken CentOS 7 RPM builds, and should go in before the upcoming patch release.


prologic:
Merged so this fix goes into the patch release.

PR #9555, opened 2020-07-16
doc: netdata user is also a default

mrbarletta:
Both `root` and `netdata` users will try to connect using the socket connection and only `root` user was specified in the doc.
Just added `netdata` and some quick comment

##### Summary
Enhance MySQL plugin documentation
##### Component Name
Mysql
##### Test Plan
just some documentation

##### Additional Information
I added that by default, both users `root` / `netdata` will try to connect using the socket connection by default.

### DIFF
--- a/collectors/python.d.plugin/mysql/README.md
+++ b/collectors/python.d.plugin/mysql/README.md
@@ -331,7 +331,7 @@ This module will produce following charts (if data is available):
 2.  **Commands** in commands/s
 
     -   select
-    -   updater
+    -   update
     -   other
 
 ## Configuration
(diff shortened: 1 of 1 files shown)


collectors/python.d.plugin/mysql/README.md
```diff
@@ -331,7 +331,7 @@ This module will produce following charts (if data is available):
 2.  **Commands** in commands/s
 
     -   select
-    -   update
+    -   updater
```
ilyam8, line 334: It is not `updater`, it is `update`

https://github.com/netdata/go.d.plugin/blob/359f7334899f691a26701d57b0075471b8b98a80/modules/mysql/testdata/MariaDBv10.5.4-galera-%5Buser_statistics%5D.txt#L16-L18

mrbarletta: nice thanks!

Issue #9567, opened 2020-07-20
eBPF plugin doesnt start after #9234 has been merged


vlvkobal:
##### Bug report summary
eBPF plugin can't start after `netdata.service` configuration was changed in #9234
```
2020-07-20 16:02:52: netdata INFO  : PLUGINSD[ebpf] : connected to '/usr/libexec/netdata/plugins.d/ebpf.plugin' running on pid 133723
2020-07-20 16:02:52: ebpf.plugin ERROR : MAIN : Setrlimit(RLIMIT_MEMLOCK) (errno 1, Operation not permitted)
2020-07-20 16:02:52: netdata ERROR : PLUGINSD[ebpf] : read failed: end of file (errno 9, Bad file descriptor)
2020-07-20 16:02:52: netdata INFO  : PLUGINSD[ebpf] : PARSER ended
2020-07-20 16:02:52: netdata ERROR : PLUGINSD[ebpf] : '/usr/libexec/netdata/plugins.d/ebpf.plugin' (pid 133723) disconnected after 0 successful data collections (ENDs).
2020-07-20 16:02:52: netdata ERROR : PLUGINSD[ebpf] : child pid 133723 exited with code 4.
2020-07-20 16:02:52: netdata ERROR : PLUGINSD[ebpf] : '/usr/libexec/netdata/plugins.d/ebpf.plugin' (pid 133723) exited with error code 4 and haven't collected any data. Disabling it.
2020-07-20 16:02:52: netdata INFO  : PLUGINSD[ebpf] : thread with task id 133712 finished
```

##### OS / Environment
Arch Linux - 5.7.9

##### Netdata version
netdata v1.23.2-28-geb42d2ad4

##### Component Name
systemd configuration

##### Steps To Reproduce
Restart Netdata, check if ebpf.plugin is running.

##### Expected behavior
ebpf.plugin doesn't crash on systems with systemd.

thiagoftsm:
@vlvkobal only root can do operations with  `Setrlimit(RLIMIT_MEMLOCK)` , it looks like that permission was missed.

Ferroin:
> @vlvkobal only root can do operations with `Setrlimit(RLIMIT_MEMLOCK)` , it looks like that permission was missed.

After looking a bit further, I believe this is indeed the issue.

[...]

PR #9569, opened 2020-07-20
Added CAP_SYS_RESOURCE to capability bounding set.

Ferroin:
##### Summary

This is a mitigation for an issue caused by #9234, whereby the eBPF collector's attempts to call `setrlimit()` (in preparation for locking memory) fail on systemd-based systems.

I do not consider this a final fix (yet) because it's not clear that we actually _need_ to lock memory, and therefore we may not actually need this `setrlimit()` call. I suspect that locking memory allows for more accurate data collection, but is not actually required for the plugin to work, in which case it should be optional and not mandatory, but we should still merge this PR (so that people who want it don't have to change anything else). If, however, we truly gain no benefit from locking memory, then this change should be scrapped and we should just remove the calls to `setrlimit()` and `memlock()` instead.

##### Component Name

area/packaging

##### Test Plan

Without this change, the `ebpf.plugin` fails on systemd systems, with it it works correctly.

##### Additional Information

Mitigates (but does not necessarily fix): #9567

### DIFF
--- a/system/netdata.service.in
+++ b/system/netdata.service.in
@@ -51,6 +51,7 @@ CapabilityBoundingSet=CAP_FOWNER          # is required for freeipmi plugin
 CapabilityBoundingSet=CAP_SETPCAP         # is required for apps, perf and slabinfo plugins
 CapabilityBoundingSet=CAP_SYS_ADMIN       # is required for perf plugin
 CapabilityBoundingSet=CAP_SYS_PTRACE      # is required for apps plugin
+CapabilityBoundingSet=CAP_SYS_RESOURCE    # is required for ebpf plugin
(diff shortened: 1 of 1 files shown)


Ferroin:
I'd like to wait to get confirmation from @prologic, but if I haven't heard from him by 23:30 UTC I'll merge this so it goes out in the nightlies tonight.

@prologic if you see this before then and approve it, please just merge it.

prologic:
@Ferroin Please bring this up for discussion in tonight's SRE Sync. I don't have time to go through all this now; so let's discuss in our team standup.

prologic:
Approving and merging as a "nightly hotfix".

PR #9575, opened 2020-07-21
Fixes mempcpy->memcpy

### LINKED ISSUE #9563: Update 1.23.1 -> 1.23.2 Failed on FreeNAS
Hi,

Just tried to update my Netdata from 1.23.1 -> 1.23.2 and got the following error:

```
aclk/aclk_stats.o: In function `aclk_stats_main_thread':
aclk_stats.c:(.text+0x1b1): undefined reference to `mempcpy'
cc: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [Makefile:4327: netdata] Error 1
gmake[2]: Leaving directory '/root/netdata-v1.23.2'
gmake[1]: *** [Makefile:5460: all-recursive] Error 1
gmake[1]: Leaving directory '/root/netdata-v1.23.2'
gmake: *** [Makefile:2947: all] Error 2
 FAILED
```   

This is on my TrueNAS 12 installed in a jail (iocage).

```
FreeBSD netdata-custom 12.1-STABLE FreeBSD 12.1-STABLE 3ef17902a90(HEAD) TRUENAS  amd64
1201517
```
Also tried to download the tarball for 1.23.2 and tried to install that but exact same error.

underhood:
##### Summary

Due to typo, we call `mempcpy` instead of `memcpy`. Surprisingly function `mempcpy` actually exists (the reason why the compiler doesn't complain) in GNU but not everywhere. As a result, the compilation fails on FreeNAS.

`mempcpy` and `memcpy` do the same job and differ only in return value (which is not used anyway).

Fixes #9563

Credit to @mfundul and @abrbon for bringing attention to this.

### DIFF
--- a/aclk/aclk_stats.c
+++ b/aclk/aclk_stats.c
@@ -249,7 +249,7 @@ void *aclk_stats_main_thread(void *ptr)
         memcpy(&permanent, &aclk_metrics, sizeof(struct aclk_metrics));
         memset(&aclk_metrics_per_sample, 0, sizeof(struct aclk_metrics_per_sample));
 
-        mempcpy(aclk_queries_per_thread_sample, aclk_queries_per_thread, sizeof(uint32_t) * query_thread_count);
+        memcpy(aclk_queries_per_thread_sample, aclk_queries_per_thread, sizeof(uint32_t) * query_thread_count);
         memset(aclk_queries_per_thread, 0, sizeof(uint32_t) * query_thread_count);
         ACLK_STATS_UNLOCK;
 


abrbon:
> ##### Summary
> 
> Due to typo, we call `mempcpy` instead of `memcpy`. Surprisingly function `mempcpy` actually exists (the reason why the compiler doesn't complain) in GNU but not everywhere. As a result, the compilation fails on FreeNAS.
> 
> `mempcpy` and `memcpy` do the same job and differ only in return value (which is not used anyway).
> 
> Fixes #9563
> 
> Credit to @mfundul and @abrbon for bringing attention to this.
> ##### Component Name
> ##### Test Plan
> ##### Additional Information

Funny thing is that in my jail on TrueNAS 12.1 with the 12.1-RELEASE it seems to work fine.
So indeed the mempcpy does work in the latest 12.1 Release but not in 11.3 and below.

PR #9576, opened 2020-07-21
eBPF new charts documentation

thiagoftsm:
##### Summary
**More details next days**
##### Component Name
Documentation


collectors/ebpf.plugin/README.md
```diff
@@ -191,6 +191,61 @@ The eBPF collector enables and runs the following eBPF programs by default:
...
+[network viewer]
+    maximum dimensions = 500
+    ports = 1-1024 !145 !domain
+    hostnames = !example.com
+    ips = !127.0.0.1/8 10.0.0.0/8 172.16.0.0/12 192.168.0.0/16 fc00::/7
+```
+
+When you define a `ports` setting, Netdata will collect network metrics for that specific port. For example, if you
+write `ports = 19999`, Netdata will collect only connections for itself. The `ports`, `hostnames`, and `ips` settings
+all accept [simple patterns](/libnetdata/simple_pattern/README.md).
```
ilyam8, line 208: > The `ports`, `hostnames`, and `ips` settings
all accept [simple patterns]

Simple patterns logic: **return on the first match** (positive or negative match).

If i understand your example correctly it doesnt work like simple patterns.

> ports = 1-1024 !145 !domain

if we return on the first match we dont exclude `145`.

ilyam8: So, is order a thing for ips, ports, etc?

thiagoftsm: It is complicated to do 100% like `simple patterns` here, because users could define `ports = 1-1024 3128-5432 40000-65535`, so we have to test for all these acceptable ranges and we would exclude in the others. The same thing happens with IPS. 

About the order, we test the correct order when the values are given in a range, but it does not matter if users define something like this `192.168.0.0/16 172.20.0.0/12`.

ilyam8: i dont get it.

`10.0.0.0/8 !10.0.0.0-10.0.0.10`

Does 10.0.0.3 match?

thiagoftsm: > 10.0.0.0/8 !10.0.0.0-10.0.0.10

Does 10.0.0.3 match?

For this example `10.0.0.3` will be eliminated together all values in the range `10.0.0.0-10.0.0.10`, but we continue accepting all the remaining values.

ilyam8: ok, so it is

> match at least one positive "pattern" AND do not match any of negated "patterns"

the order is not relevant

collectors/ebpf.plugin/README.md
```diff
@@ -219,10 +219,40 @@ The following options are available:
...
 will be bundled into the `other` dimension. You can increase the number of shown dimensions by changing the `maximum
 dimensions` setting.
 
+The dimensions for the traffic charts are created using the destination IPs of the sockets by default. This can be
+changed setting `resolve hostname ips = yes` and restarting Netdata, after this Netdata will create dimensions using
+the `hostnames` every time that is possible to resolve IPs to their hostnames.    
+
+#### Default values
+
+When the file `/etc/netdata/ebpf.conf` is not created, `ebpf.plugin` will start with the following default values
```
mfundul, line 232:
```suggestion
When the file `/etc/netdata/ebpf.conf` is missing, `ebpf.plugin` will start with the following default values
```

collectors/ebpf.plugin/README.md
```diff
@@ -219,10 +219,40 @@ The following options are available:
     range of IPs, or use CIDR values. The default behavior is to only collect data for private IP addresess, but this
     can be changed with the `ips` setting.
     
-By default, Netdata displays up to 500 dimensions on network viewer charts. If there are more possible dimensions, they
+By default, Netdata displays up to 50 dimensions on network viewer charts. If there are more possible dimensions, they
```
mfundul, line 222: It needs to be clear what the criterion that qualifies which dimension will be displayed is. Is it the most recent one that displaces the rest? Is it the most frequently used ones? I guess in our case it would be the oldest dimensions are displayed and the newest are bundled into `other`?

collectors/ebpf.plugin/README.md
thiagoftsm: thank you, I fixed on the last commit.

collectors/ebpf.plugin/README.md
thiagoftsm: I changed the paragraph to explicit that we are showing on `1.24` the oldest one.

web/gui/dashboard_info.js
```diff
@@ -3196,6 +3196,32 @@ netdataDashboard.context = {
         info: 'Difference between the number of process created and the number of threads created per period(<code>process</code> dimension), it also shows the number of possible zombie process running on system.'
     },
 
+    'ebpf.outbound_bytes': {
+        title : 'Outbound bytes',
+        info: 'Bytes sent to specific destination using TCP or UDP ports. It shows the first 50 outbound connections collected for the plugin, if there are more connections they will be bundled into the `other` dimension.'
```
ilyam8, line 3201: > It shows the first 50 outbound connections collected

> yo can increase the number of shown dimensions dimensions by changing the `maximum dimensions` setting.

It seems you cant hardcode a number in the `info`

thiagoftsm: Good catch, I updated it now.

collectors/ebpf.plugin/README.md
```diff
@@ -219,9 +219,39 @@ The following options are available:
     range of IPs, or use CIDR values. The default behavior is to only collect data for private IP addresess, but this
     can be changed with the `ips` setting.
     
-By default, Netdata displays up to 500 dimensions on network viewer charts. If there are more possible dimensions, they
-will be bundled into the `other` dimension. You can increase the number of shown dimensions by changing the `maximum
-dimensions` setting.
+By default, Netdata `1.24.x` displays up to 49 older sockets (dimensions) on network viewer charts. If there are more
+ possible  dimensions, they will be bundled into the `other` dimension. You can increase the number of shown
```
mfundul, line 223:
```suggestion
By default, Netdata `1.24.x` displays up to 49 sockets (dimensions) on network viewer charts. If there are more
 dimensions, only the oldest ones will be shown individually and the rest will be bundled into the `other` dimension. You can increase the number of shown
```

thiagoftsm: Done!

web/gui/dashboard_info.js
```diff
@@ -3196,6 +3196,32 @@ netdataDashboard.context = {
...
+        info: 'Number of TCP packages retransmitted per period of time. It shows the oldest outbound connections collected for the plugin respecting <code>maximum dimensions</code> defined inside <code>/etc/netdata/ebpf.conf</code>, if there are more connections they will be bundled into the <b>>other</b> dimension.'
+    },
+
+    'ebpf.inbound_bytes': {
+        title : 'Inbound bytes',
+        info: 'Bytes received for a local service. It shows the oldest inbound connections collected for the plugin respecting <code>maximum dimensions</code> defined inside <code>/etc/netdata/ebpf.conf</code>, if there are more connections they will be bundled into the <b>other</b> dimension.'
+    },
+
+    'ebpf.inbound_packets': {
+        title : 'Inbound bytes',
```
aabatangle, line 3220: title: 'Inbound packets'

collectors/ebpf.plugin/README.md
```diff
@@ -219,9 +219,39 @@ The following options are available:
...
+#### Network traffic chart
+
+Netdata monitors network connections established for hosts, which are shown according the package final destination.
+
+You can visualize the most common destinations for `outbound connections` and measure the bandwidth utilized with 
+connections for local services. All these information are grouped using:
+
+-   `bytes`: Bytes sent and received for a destination.
+-   `packets`: Packets sent per period of time.
+-   `retransmission`: Number of TCP packages retransmitted. This is available only for `outbound connections`.
```
ilyam8, line 254: > Number of TCP packages

It is `calls` on the charts. calls != packets. If this is `packets` then charts should be fixed, if `calls` - docs.

ilyam8: > -   `packets`: Packets sent per period of time.

> per period of time.

It is per seconds in our case, if dimension algorithm is `incremental` then `netdata` calculates per second value.

I think `per period of time` is not needed at all and it is just `sent and received packets`

@joelhans

web/gui/dashboard_info.js
thiagoftsm: Thank you, I fixed.

collectors/ebpf.plugin/README.md
thiagoftsm: I fixed the documentation.

vlvkobal:
We forgot about that, but the eBPF plugin documentation should be in the "External plugins" section.

thiagoftsm:
We will bring another PR when we return working with Network viewer.

Issue #9077, opened 2020-05-17
FAILED not critical -> after installing using the official installation script


thoggs:
Hello,
After running the `netdata.io/kickstart.sh` installation script, at the end of the installation, you receive the following error:
**Note:** cronie is already installed
![Captura de Tela 2020-05-17 às 12 54 45](https://user-images.githubusercontent.com/43221193/82153620-f55b0a00-983e-11ea-8d4d-f8ac182ab102.png)

**Note 2:** installation made on a clean operating system, without any customization or any type of software that may interfere with the proper functioning of Netdata

**OS:** Clear Linux* OS v3110 (Server) - Kernel v5.6.13
**Netdata:** v1.22.1-48-nightly

prologic:
@thoggs Hi 👋-- To be clear your issue is that we don't install the cron job for the `netdata-updater` on your system? If so this is a bit of a KP and we lack support for cronie, busybox's cron and maybe others (_those are just the ones I'm  aware of_).

thoggs:
> Hi 👋-- To be clear your issue is that we don't install the cron job for the `netdata-updater` on your system? If so this is a bit of a KP and we lack support for cronie, busybox's cron and maybe others (_those are just the ones I'm aware of_).

I am happy for the work in progress; the Netdata development team is amazing.

prologic:
> I am happy for the work in progress; the Netdata development team is amazing.

Thank you ! We appreciate the positive feedback. I will make sure we support more Corn schedulers. At least the two I mention. I also want support for busybox's userland myself.

Ferroin:
This is potentially a bit more complicated than 'supportcronie and busybox crond' The issue isn't that we don't support them, it's that we don't support the approaches used by some distros that use them. _Most_ distros have `/etc/cron.daily` directories and provide default crontabs for whatever cron implementation you install through their PM that will execute the stuff in that directory on a daily basis. Some just don't, and some only do so for some cron implementations.

I can look at adding support for injecting a crontab directly on systems where we can do that and can't find `/etc/cron.daily`, and also at support for installing a systemd timer unit on systems that don't have cron but do have systemd., but there will always be some systems we can't support because we _really_ cannot go around modifying the system's master crontab or the root user's personal crontab.

prologic:
[...]

PR #9577, opened 2020-07-21
Added improved scheduler support for automatic updates.

Ferroin:
##### Summary

This adds support for using a drop-in crontab in `/etc/cron.d` or a systemd timer unit for scheduling auto-updates for Netdata as alternatives to our current support for linking the updater script in `/etc/cron.daily`. The install script will prefer the systemd timer on new installs if it is usable, and will fall back on new installs to the `/etc/cron.d` approach if neither the systemd timer nor `/etc/cron.daily` will work, and will convert existing installs to use the preferred method of scheduling for the system type.

The timer unit triggers a simple one-shot service which calls out to the updater script, and uses the standard `OnCalendar=daily` scheduling provided by systemd to trigger when to run, but is intentionally _not_ a persistent timer unit (that is, it won't run immediately at the first opportunity after a missed run, just at the next possible run). Note that due to limitations of the systemd timer arrangement, this will not email the user if the update fails.

The crontab method uses 02:57 local time as the start point (chosen arbitrarily to avoid matching up with what most distros use for daily jobs by default), and runs as root by default.

Together this will cover _most_ distributions scheduling setups. It notably does not cover usage of the BusyBox implementation of `crond`, but we cannot safely cover that as doing so would require manipulating existing files instead of just adding new ones (and likely does not cover ancient SVR4 and 4.3BSD era `crond` implementations either for the same reasons).

##### Component Name

area/packaging

##### Test Plan

Attempt installs on a variety of different system types to verify that the correct scheduler setup gets used for each type.

Manually tested on a couple of different systemd and `/etc/cron.daily` systems which all worked correctly. I don't have any test systems that do not have `/etc/cron.daily` but do have `/etc/cron.d` to verify on, but I've verified that selection by manually removing the `/etc/cron.daily` directories on a couple of test systems.

##### Additional Information

Fixes: #9077

### DIFF
--- a/build/subst.inc
+++ b/build/subst.inc
@@ -5,6 +5,7 @@
 		-e 's#[@]pluginsdir_POST@#$(pluginsdir)#g' \
 		-e 's#[@]configdir_POST@#$(configdir)#g' \
 		-e 's#[@]libconfigdir_POST@#$(libconfigdir)#g' \
+		-e 's#[@]pkglibexecdir_POST@#$(pkglibexecdir)#g' \
 		-e 's#[@]cachedir_POST@#$(cachedir)#g' \
 		-e 's#[@]registrydir_POST@#$(registrydir)#g' \
 		-e 's#[@]varlibdir_POST@#$(varlibdir)#g' \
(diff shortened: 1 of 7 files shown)


Ferroin:
Closing this as it will be split out as multiple PR's.

PR #9584, opened 2020-07-22
Add lock dir

### LINKED ISSUE #9347: Add directory for storing lock files + Env variable
##### Feature idea summary

Related: https://github.com/netdata/product/issues/1143

One and only one job with same `moduleName_jobName` should be in the running state. It is easy to achieve for a plugin and we do it.

The problem: we cannot guarantee that this condition is met globally - across all plugins. There is no single source of truth at the moment.

We decided to go with locking files approach:

- try to lock  `moduleName_jobName.lock` file
- success: start the job, do not release the lock while the job is up and running.
- fail: the job with the same name is being served by another plugin - drop it.

---

We need:
- a directory to store all those locks
- enviroment var that points to the dir
 
@thiagoftsm suggested to use `/var/lock/netdata`.

I am not sure about that, i see not all systems have it. I see it is a symlink to `/run/lock` on my manjaro and it seems other [distros moved to `/run/lock` too](https://wiki.debian.org/ReleaseGoals/RunDirectory#How_to_transition_from_.2Fvar.2Frun_to_.2Frun_and_.2Fvar.2Flock_to_.2Frun.2Flock.3F)

🤷‍♂️ we could use `/var/lib/netdata` ?

@Ferroin i am sure you know what the best way is!

vlvkobal:
##### Summary
In order to prevent python and go plugins from mutual conflicts, we are adding a directory for lock files.

Fixes #9347

##### Component Name
daemon

##### Test Plan
Run Netdata, check if `/var/lib/netdata/lock` is created and contains lock files only for running python plugins.

### DIFF
--- a/daemon/daemon.c
+++ b/daemon/daemon.c
@@ -63,11 +63,12 @@ void clean_directory(char *dirname)
     DIR *dir = opendir(dirname);
     if(!dir) return;
 
+    int dir_fd = dirfd(dir);
     struct dirent *de = NULL;
 
     while((de = readdir(dir)))
         if(de->d_type == DT_REG)
-            unlink(de->d_name);
+            unlinkat(dir_fd, de->d_name, 0);
 
     closedir(dir);
 }
--- a/daemon/daemon.c
+++ b/daemon/daemon.c
@@ -68,7 +68,8 @@ void clean_directory(char *dirname)
 
     while((de = readdir(dir)))
         if(de->d_type == DT_REG)
-            unlinkat(dir_fd, de->d_name, 0);
+            if (unlinkat(dir_fd, de->d_name, 0))
+                error("Cannot delete %s/%s", dirname, de->d_name);
 
     closedir(dir);
 }
--- a/daemon/daemon.c
+++ b/daemon/daemon.c
@@ -72,6 +72,7 @@ int become_user(const char *username, int pid_fd) {
 
     create_needed_dir(netdata_configured_cache_dir, uid, gid);
     create_needed_dir(netdata_configured_varlib_dir, uid, gid);
+    create_needed_dir(netdata_configured_lock_dir, uid, gid);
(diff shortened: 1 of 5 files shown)


daemon/daemon.c
```diff
@@ -58,6 +58,23 @@ void create_needed_dir(const char *dir, uid_t uid, gid_t gid)
...
+    struct dirent *de = NULL;
+
+    while((de = readdir(dir)))
+        if(de->d_type == DT_REG) {
+            char filename[FILENAME_MAX +1];
+            snprintfz(filename, FILENAME_MAX, "%s/%s", dirname, de->d_name);
+            unlink(filename);
+        }
+
+    closedir(dir);
```
ilyam8, line 75: Could we just delete the lock dir as a cleanup?

ilyam8: ok, i see, anyway we need to delete all content before the dir itself

daemon/daemon.c
```diff
@@ -58,6 +58,23 @@ void create_needed_dir(const char *dir, uid_t uid, gid_t gid)
...
+    DIR *dir = opendir(dirname);
+    if(!dir) return;
+
+    struct dirent *de = NULL;
+
+    while((de = readdir(dir)))
+        if(de->d_type == DT_REG) {
+            char filename[FILENAME_MAX +1];
+            snprintfz(filename, FILENAME_MAX, "%s/%s", dirname, de->d_name);
+            unlink(filename);
```
thiagoftsm, line 72: Please, test `unlink` return, and give an additional error message to the log.

vlvkobal: Done.

Ferroin:
Merging this as requested by @vlvkobal as Travis is causing problems again (build has passed, but we have two listed for some reason and one is stuck waiting for a response that will never come because there's only on actual build).

PR #9598, opened 2020-07-23
Added support for using `/etc/cron.d` for auto-updates.

Ferroin:
##### Summary

This adds support for scheduling auto-updates by placing a crontab in `/etc/cron.d` on systems that do not have either `/etc/cron.daily` or `/etc/periodic/daily`. The crontab fragment used for this runs the updater as root at 02:57 daily (but still uses the same random delay that our current auto-update approach uses), with the time chosen arbitrarily to be different from the times used by most major distributions for their daily scheduled tasks.

##### Component Name

area/packaging

##### Test Plan

Crontab verified to work correctly on a number of local test systems through manual testing. Detection logic in the installer tested on a couple of systems by explicitly removing `/etc/cron.daily` and `/etc/periodic/daily`.

##### Additional Information

Part of fixing #9077 
Partially supersedes #9577

### DIFF
--- a/build/subst.inc
+++ b/build/subst.inc
@@ -5,6 +5,7 @@
 		-e 's#[@]pluginsdir_POST@#$(pluginsdir)#g' \
 		-e 's#[@]configdir_POST@#$(configdir)#g' \
 		-e 's#[@]libconfigdir_POST@#$(libconfigdir)#g' \
+		-e 's#[@]pkglibexecdir_POST@#$(pkglibexecdir)#g' \
 		-e 's#[@]cachedir_POST@#$(cachedir)#g' \
 		-e 's#[@]registrydir_POST@#$(registrydir)#g' \
 		-e 's#[@]varlibdir_POST@#$(varlibdir)#g' \
--- a/packaging/installer/functions.sh
+++ b/packaging/installer/functions.sh
@@ -899,9 +899,7 @@ safe_sha256sum() {
 }
 
 _get_scheduler_type() {
-  if issystemd ; then
-    echo 'systemd'
-  elif _get_intervaldir > /dev/null ; then
+  if _get_intervaldir > /dev/null ; then
     echo 'interval'
   elif [ -d /etc/cron.d ] ; then
     echo 'crontab'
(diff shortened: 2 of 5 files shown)


Ferroin:
> This currently fails on Debian 9:
> 
> ```
>  --- Installing (but not enabling) the netdata updater tool ---
> Update script is located at /usr/libexec/netdata/netdata-updater.sh
> 
>  --- Check if we must enable/disable the netdata updater tool ---
> Unable to determine what type of auto-update scheduling to use, not enabling auto-updates.
> 
>  FAILED  Cannot enable netdata updater tool
> ```

Fixed. I had missed one of the bits of code for the systemd timer support, so the code was (correctly) detecting the preferred scheduler type as systemd, but didn't know how to actually set that up, thus returning the above error message.

prologic:
Can you please test on (_and show results_):

* Ubuntu 19.10
* Ubuntu 20.04
* Debian 9
* Debian 10
* CentOS 7
* CentOS 8

Thanks :)

The last thing we want to do is break the updater on folks!

Ferroin:
> Can you please test on (_and show results_):
> 
> * Ubuntu 19.10
> * Ubuntu 20.04
> * Debian 9
> * Debian 10
> * CentOS 7
> * CentOS 8
> 
> Thanks :)
> 
> The last thing we want to do is break the updater on folks!

The relevant output on all the above systems (and all the others I tested on without modifying the system itself) is (with the path correctly set based on the system type):

```
 --- Check if we must enable/disable the netdata updater tool --- 
Auto-updating has been enabled through cron, updater script linked to /etc/cron.daily/netdata-updater

If the update process fails and you have email notifications set up correctly for cron on this system, you should receive an email notification of the failure.
Successful updates will not send an email.
```

Basic testing with the environment manually manipulaed to force usage of `/etc/cron.d` spit out the correct message for that as well, and all the cases I checked that htis is supposed to support properly set up the update script where it was supposed to be.

prologic (approved):

Ferroin:
I'm holding off on merging this until after the upcoming release.

PR #9619, opened 2020-07-28
Remove PrivateMounts

Steve8291:
##### Summary
Since commit #9234 the following has been showing up in my journal logs:

>  /lib/systemd/system/netdata.service:63: Unknown lvalue 'PrivateMounts' in section 'Service'

Running `systemd-analyze verify netdata.service`

> /lib/systemd/system/netdata.service:63: Unknown lvalue 'PrivateMounts' in section 'Service'
> Attempted to remove disk file system, and we can't allow that.

It would appear that the `Attempted to remove disk file system, and we can't allow that.` message is due to a [bug](https://github.com/systemd/systemd/issues/8592) in version 237 of systemd, which is what my Ubuntu 18.04 is currently running. But that doesn't explain the "PrivateMounts" issue.

Looking at the [systemd man page](https://www.freedesktop.org/software/systemd/man/systemd.exec.html) for PrivateMounts it would seem that we do not need the `PrivateMounts=true` setting because the service file has `ProtectSystem=full` set:

> Other file system namespace unit settings — PrivateMounts=, PrivateTmp=, PrivateDevices=, ProtectSystem=, ProtectHome=, ReadOnlyPaths=, InaccessiblePaths=, ReadWritePaths=, … — also enable file system namespacing in a fashion equivalent to this option. Hence it is primarily useful to explicitly request this behaviour if none of the other settings are used.

##### Component Name
netdata/system/netdata.service

##### Test Plan
I have tested this on Ubuntu 18.04 running systemd version 237 and everything seems to be working fine.
Probably needs tested on other distros.

### DIFF
--- a/system/netdata.service.in
+++ b/system/netdata.service.in
@@ -60,7 +60,6 @@ ProtectHome=read-only
 # PrivateTmp break netdatacli functionality. See - https://github.com/netdata/netdata/issues/7587
 #PrivateTmp=true
 ProtectControlGroups=true
-PrivateMounts=true
 # We whitelist this because it's the standard location to listen on a UNIX socket.
 ReadWriteDirectories=/run/netdata
 


prologic:
I'm a bit confused here now... #9234 added this and this PR #9619 removes it. One of these two PRs is right 🤔

@Izorkin Can you chime in here since you wrote the PR #9234 ? Thanks!

Izorkin:
I checked `PrivateMounts = true;` only systemd v245 - https://github.com/NixOS/nixpkgs/pull/87867#discussion_r426155608
If `ProtectSystem = full`, the `PrivateMounts` option is enabled or not - I don't know.
Option `PrivateMounts` appeared in version 239. In version 237, the option is ignored and issues a warning.

Steve8291:
If the issue isn't a problem in newer versions of systemd I can close the pull request. Or if someone wants to tell me what to check to see if `ProtectSystem = full` really acomplishes the private namespacing without `PrivateMounts` I'd be happy to do so.

amoss:
The systemd docs say:
```
Other file system namespace unit settings — PrivateMounts=, PrivateTmp=, PrivateDevices=, ProtectSystem=, ProtectHome=, ReadOnlyPaths=, InaccessiblePaths=, ReadWritePaths=, … — also enable file system namespacing in a fashion equivalent to this option. Hence it is primarily useful to explicitly request this behaviour if none of the other settings are used.
```
So this implies that we do not need to set PrivateMounts if we are setting `ProtectSystem=full`.

prologic (approved):

prologic:
@Ferroin Please review and if happy approve and merge.

PR #9628, opened 2020-07-29
Added eBPF collector support to DEB and RPM packages.

Ferroin:
##### Summary

This PR adds working support for the eBPF plugin to our DEB and RPM packages. It adds code to prepare our local build of libbpf as well as properly bundling the actual eBPF programs into the final packages.

It also adds basic DEB and RPM build checks to our normal Travis workflow (done on Debian 10 and Fedora 31), as well as adding code to make the package build process actually show full output in the Travis log, and updating the configure script to explicitly show all the things we're checking for the eBPF collector build.

##### Component Name

area/packaging
area/ci

##### Test Plan

Verified both using local package builds and in Travis using the new DEB and RPM build checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -165,6 +165,46 @@ jobs:
       script: 'for i in $(seq 0 4); do printf "[XXX: Run #%s]\n" "$i";docker run -it -v "${PWD}:/netdata:rw" -w /netdata "fedora:31" tests/updater_checks.sh && break; done'
       after_failure: post_message "TRAVIS_MESSAGE" "Netdata updater process failed on bare Fedora 31"
 
+    - name: DEB package test
+      git:
+        depth: false
+      before_install:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -165,6 +165,46 @@ jobs:
       script: 'for i in $(seq 0 4); do printf "[XXX: Run #%s]\n" "$i";docker run -it -v "${PWD}:/netdata:rw" -w /netdata "fedora:31" tests/updater_checks.sh && break; done'
       after_failure: post_message "TRAVIS_MESSAGE" "Netdata updater process failed on bare Fedora 31"
 
+    - name: DEB package test
```
prologic, line 168: In fact just move all this to a GHA workflow 👍

Ferroin: I intentionally put it in Travis because that's the environment where we're building these packages, so we can be sure that we've got the exact same environment.

Ferroin:
Discussed with @prologic in the SRE sync today, merging this as he agreed with my reasoning about keeping the CI in Travis.

PR #9639, opened 2020-07-31
dashboard v1.0.26

jacekkolasa:
Most recent dashboard version, plus fixed package-lock file (to clear GH warning)


ilyam8:
@jacekkolasa scale units works strange - the dashboard scales units only when the values are being changed (looks like)

- 2nd chart has `Hz`

Screenshot 2020-07-31 at 17 41 30

- 3rd chart has `bytes`

Screenshot 2020-07-31 at 17 42 56

ilyam8:
There is definitly an issue, script to reproduce the bug
<details>
<summary>script click click</summary>

```python
from random import SystemRandom
 
priority = 90000
 
ORDER = [
    'random_Hz_dynamic',
    'random_Hz_static',
    'random_bytes_dynamic',
    'random_bytes_static',
]
 
CHARTS = {
    'random_Hz_dynamic': {
        'options': [None, 'A random number', 'Hz', 'Hz', 'Hz', 'line'],
        'lines': [
            ['random1']
        ]
    },
    'random_Hz_static': {
        'options': [None, 'A random number', 'Hz', 'Hz', 'Hz', 'line'],
        'lines': [
            ['random2']
        ]
    },
    'random_bytes_dynamic': {
        'options': [None, 'A random number', 'bytes', 'bytes', 'bytes', 'line'],
        'lines': [
            ['random3']
        ]
    },
    'random_bytes_static': {
        'options': [None, 'A random number', 'bytes', 'bytes', 'bytes', 'line'],
        'lines': [
            ['random4']
        ]
    }
}
 
 
class Service(SimpleService):
        self.order = ORDER
        self.definitions = CHARTS
        self.random = SystemRandom()
 
    @staticmethod
    def check():
        return True
 
    def get_data(self):
        data = dict()
 
        for i in range(1, 5):
            dimension_id = ''.join(['random', str(i)])
            if i % 2 == 0:
                data[dimension_id] = 1000 * 1000 * 1000 * self.random.randint(1, 100)
            else:
                data[dimension_id] = 1000 * 1000 * 1000
 
        return data
```


</details>

jacekkolasa:
@ilyam8 Indeed, there is an issue, good catch! It seems to be present for all units, not just the ones you added, and is present also in current Dashboard version, so there is no regression. There is a regression however from the /old version. It needs definitely to be fixed, but can we do it in separate PR? This Dashboard version has some other features/fixes which would be good to have.

PR #9644, opened 2020-08-03
installer: update `go.d.plugin` version to v0.20.0

ilyam8:
##### Summary

Added the Prometheus Generic Collector.

##### Component Name

`packaging/`

##### Test Plan

- [x] install this branch, ensure there is no errors and go.d.plugin version is 0.20.0


amoss:
The initial test-case was to install the `prometheus` package on Debian 10 (buster). This pulls in `node_exporter` by default. Netdata was working before the install. After the install the dashboard stops working. The browser inspector shows that the charts response does not validate properly, it is 23000 lines long and there seems to be a problem with the escaping of values.

Tracing through it with @ilyam8 reveals that the original data looks like this:
```
node_systemd_service_restart_total{state="systemd-fsck@dev-disk-by\\x2duuid-7D37\\x2dDD70.service"} 0
node_systemd_service_restart_total{state="systemd-fsck@dev-disk-by\\x2duuid-eb5d5d19\\x2db0f6\\x2d4607\\x2dbebf\\x2de5bd557619e7.service"} 0
```

The double-backslashes are causing us problems.

amoss:
@ilyam8 I think that we should fix the strings on entry to Netdata (from `go.d`) and also do the escaping in the web handler to avoid breaking the JSON. I changed your code fragment on Slack, could you transform the dimension names like this:
```
orig := "systemd-fsck@dev-disk-by\\x2duuid-7D37\\x2dDD70.service"
unescaped := html.UnescapeString(orig)
for ;orig!=unescaped; {
    orig = unescaped
    unescaped = html.UnescapeString(orig)
}
unescaped = unescaped.ReplaceAll(unescaped,"\\","")

fmt.Println(unescaped)
fmt.Println(escaped)
```
This will strip as many levels of encoding as possible from the string, and it is guaranteed to terminate as each replacement has fewer backslashes than the original. My golang is a bit rusty - if it quicker to compile a regex and use it to strip any remaining `\` characters then do it that way instead. The only other character that could break the JSON on output is a `"` - but it is not possible to pass one in a dimension name as the parser will process it.

I'll do the other side and escape any characters in the handler that would break the JSON syntax.

ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5952518&bid=19509803)

amoss:
@ilyam8 I've pushed the commit to fix the JSON output, it seems to work well and I can't find any problems. If @cakrit is happy to strip out the node_exporter charts then it could be done, but it does not seem to impact performance on my system. I would like to see the other unescaping code in action first though, before anything is removed.

@mfundul: I've now written some of the code and been a reviewer, could you review my C code in-case I've missed anything (it's one of our shared AoRs).

amoss:
I saw the Codacy linting warning - it is flat-out wrong. I assume it can't handle the loop macro correctly, I'll see if we can disable it.

ilyam8:
@prologic any ideas how to fix this?


```cmd
node_systemd_unit_tasks_current{name="systemd-fsck@dev-disk-by\\x2duuid-eb5d5d19\\x2db0f6\\x2d4607\\x2dbebf\\x2de5bd557619e7.service"} 0
```

=> after https://github.com/prometheus/prometheus/blob/fac7a4a0504404fa5d4c5abb8fcc9750bd5cbda7/pkg/textparse/promparse.go#L208-L236

```cmd
node_systemd_unit_tasks_max|name=systemd-fsck@dev-disk-by\x2duuid-eb5d5d19\x2db0f6\x2d4607\x2dbebf\x2de5bd557619e7.service
```

prologic:
@ilyam8 How do you have the Unicode equivalent of the ASCII `-` there? (`\x2d`) wtf/! Where is this coming from?

amoss:
Those escape sequences are not Unicode - both Unicode and JSON use \u. They look like c-format escape strings. I don’t think there is a standard library in golang that will handle them. Options here are probably:
1. ReplaceAll \x -> \u and then Unicode unescape 
2. Scan the string and manually decide those sequences

libnetdata/buffer/buffer.c
```diff
@@ -160,6 +160,31 @@ void buffer_strcat(BUFFER *wb, const char *txt)
     }
 }
 
+void buffer_strcat_jsonescape(BUFFER *wb, const char *txt)
```
mfundul, line 163: You need to put `void buffer_strcat_jsonescape(BUFFER *wb, const char *txt)` in a header file to avoid `implicit decleration` errors.

amoss: Ah. Did not see those. Will add...

ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5952518&bid=19509803)

ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5952518&bid=19509803)

ilyam8:
I removed `node_exporter` from the module stock config.

I dont know how to fix quickly that weird escaping seq in a string literal.

> ReplaceAll \x -> \u and then Unicode unescape

this wont do

> Scan the string and manually decide those sequences

It looks like this. We need to respect this spec https://golang.org/ref/spec#Rune_literals

- `\x` followed by exactly two hexadecimal digits.
- `\u` followed by exactly four hexadecimal digits.
- `\U` followed by exactly eight hexadecimal digits.
- `\` followed by exactly three octal digits.

PR #9646, opened 2020-08-03
installer: fix `govercomp`

ilyam8:
##### Summary

I noticed that installer downloads `v0.19.2` binary despite i have `v0.20.0` in the `/plugins.d`. This PR fixes it.

##### Component Name

`installer`

##### Additional Information

- [X] install this PR (master go version is `v0.19.2`), build go.d.plugin `v0.20.0`, put it in the `/plugins.d`, do reinstall, ensure `v0.19.2` is not downloaded

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -1211,7 +1211,7 @@ govercomp() {
   for ((i = 0; i < ${#ver1[@]}; i++)); do
     if [ "${ver1[i]}" -gt "${ver2[i]}" ]; then
       return 1
-    elif [ "${ver1[i]}" -gt "${ver2[i]}" ]; then
+    elif [ "${ver2[i]}" -gt "${ver1[i]}" ]; then
       return 2
     fi
   done


netdata-installer.sh
```diff
@@ -1211,7 +1211,7 @@ govercomp() {
   for ((i = 0; i < ${#ver1[@]}; i++)); do
     if [ "${ver1[i]}" -gt "${ver2[i]}" ]; then
       return 1
-    elif [ "${ver1[i]}" -gt "${ver2[i]}" ]; then
+    elif [ "${ver2[i]}" -gt "${ver1[i]}" ]; then
```
prologic, line 1214: I _could_ be wrong here but I"m not sure you can do this kind of comparison when it comes to versions as you end up effectively doing lexicographical comparisons and this may be problematic. It would be better if you parsed the version strings into numbers like I've done elsewhere in other parts of hte netdata installer and updater and compared actual integers. 🤔

ilyam8: it works

https://github.com/netdata/netdata/blob/bdda59ce541dc0be0092328ba74f01f057871989/netdata-installer.sh#L1198-L1220

the function expectes go.d.plugin version

https://github.com/netdata/netdata/blob/bdda59ce541dc0be0092328ba74f01f057871989/netdata-installer.sh#L1186-L1195

ilyam8: > you parsed the version strings into numbers

i do this - regex + split (IFS=.). We could use some other functions you mentioned, i am ok with it.

PR #9652, opened 2020-08-03
Fix installer to not install eBPF plugin/collector if installed as non-root (euid=0)

prologic:
##### Summary

ssia

##### Component Name

- packaging

##### Test Plan

Ran the installer as non-root:

```#!sh
The following non-fatal errors were encountered during the installation process:

* eBPF disabled in non-root (euid=0) installation

 --- We are done! ---
```

##### Additional Information

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -1367,6 +1367,12 @@ detect_libc() {
 }
 
 should_install_ebpf() {
+  if [ -n "$EUID" ] && [ $EUID -ne 0 ]; then
+    run_failed "Not running as root (euid=0), not installing eBPF..."
+    defer_error "eBPF disabled in non-root (euid=0) installation"
+    return 1
+  fi
+
   if [ "${NETDATA_DISABLE_EBPF:=0}" -eq 1 ]; then
     run_failed "eBPF explicitly disabled."
     defer_error "eBPF explicitly disabled."


prologic:
@thiagoftsm Can you test this on a normal install as root to make sure it all still works as expected? 👍

netdata-installer.sh
```diff
@@ -1367,6 +1367,12 @@ detect_libc() {
 }
 
 should_install_ebpf() {
+  if [ -n "$EUID" ] && [ $EUID -ne 0 ]; then
```
thiagoftsm, line 1370: I have only one doubt, I could observe that the PR is doing what is expected, it is not doing download of the files, but we are not avoiding the compilation of the ebpf.plugin. Considering this, I think we need to change the message 

```
Not running as root (euid=0), not installing eBPF...
```

, because they will have `ebpf.plugin` inside `/usr/libexec/netdata/plugins.d`.

thiagoftsm:
> @thiagoftsm Can you test this on a normal install as root to make sure it all still works as expected? +1

It works as expected, for me only the message needs to be changed.

netdata-installer.sh
prologic: Do we _want_ to avoid compliing the `ebpf.plugin` as well theN?

prologic:
> > @thiagoftsm Can you test this on a normal install as root to make sure it all still works as expected? +1
> 
> It works as expected, for me only the message needs to be changed.

See [my comment](https://github.com/netdata/netdata/pull/9652#discussion_r464781819)

netdata-installer.sh
thiagoftsm: To keep this exact message I think we should stop `ebpf.plugin` compilation, because the plugin will be compiled and it will always write an error when Netdata tries to load it. We could change the message for something like `Not running as root (euid=0), eBPF.plugin cannot be initialized ...`. If this happens, this PR can be merged.

amoss: Yes - we want to avoid the compilation if it is not being installed.

Ferroin: IMO, this should also translate to not building libbpf if we're not installing the eBPF plugin, so we'll also need a check there.

prologic: @thiagoftsm Where are we at with the above? Do we skip compiling this at all if we don't have to? I can't tell 😳

thiagoftsm: I think the best approach will be do not compile at all if it is not root.

PR #9661, opened 2020-08-05
Remove obsoleted libraries from install/uninstall scripts

vlvkobal:
##### Summary
We removed user-side shared libraries from kernel-collector repository https://github.com/netdata/kernel-collector/pull/172. We have to delete them from install/uninstall scripts as well.

##### Component Name
eBPF plugin, packaging

##### Test Plan
Run `netdata-installer.sh` and `netdata-uninstaller-sh`. Check if there are no errors related to `libbpf_kernel.so` or `libnetdata_ebpf.so`.

### DIFF
--- a/contrib/debian/rules
+++ b/contrib/debian/rules
@@ -96,9 +96,6 @@ override_dh_installdocs:
 		--target $(TOP)/usr/share/doc/netdata/ \
 		{} \;
 
-override_dh_shlibdeps:
-	dh_shlibdeps -X libnetdata_ebpf
-
 override_dh_fixperms:
 	dh_fixperms
 
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -1432,9 +1432,7 @@ install_ebpf() {
   # chown everything to root:netdata before we start copying out of our package
   run chown -R root:netdata "${tmp}"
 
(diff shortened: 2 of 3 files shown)


vlvkobal:
@prologic is it ok to remove `uninstall_ebpf()` from `netdata-uninstaller.sh`? What if somebody tries to uninstall an updated (so that old files are still in place) Netdata installation with a new script?

thiagoftsm:
I tested on FreeBSD and Slackware Linux, on Linux it works fine both installation and the uninstall process, on the other hand, 
on FreeBSD the installer tried to get the ebpf files and it did not extract it, because there was an error with hash, I think we 
should avoid do download of the epbf programs on FreeBSD.
I also can confirm that it did not try to look for the libraries more on FreeBSD

prologic:
> @prologic is it ok to remove `uninstall_ebpf()` from `netdata-uninstaller.sh`? What if somebody tries to uninstall an updated (so that old files are still in place) Netdata installation with a new script?

This is fine. The uninstaller needs some "rethink" anyway.

vlvkobal:
> on FreeBSD the installer tried to get the ebpf files

vlvkobal:
> on FreeBSD the installer tried to get the ebpf files

It's another issue. I'd fix that in another PR.

PR #9666, opened 2020-08-05
Fix artifact building for dist/static artifacts

prologic:
##### Summary

When you run `.github/scripts/build-artifacts.sh` you get a failure from building/testing on macOS where the build process complianing that the `openssl` directory already exists. I'm not sure why but this condition fixes that.

##### Component Name

- area/ci

##### Test Plan

Tested manually

##### Additional Information


prologic:
Someone beat me to dhit.

Issue #9664, opened 2020-08-05
Optimise size of Docker Image


prologic:
Recently the size of our Docker Image has increased from ~500MB to ~700MB.

- [Slack thread](https://netdata-cloud.slack.com/archives/CTGECGQMA/p1596497612040300)

Suspected cause: https://github.com/netdata/helper-images/pull/80

cc @amoss @ilyam8

prologic:
Tried to utilise [docker-slim](https://github.com/docker-slim/docker-slim) today but unfortunately its a little too clever and doesn't really work out very well. It strips too much and doesn't have context on what needs to remain.

Ferroin:
Closing this as it’s about as good as it can get without a significant redesign our how our Docker image works, which will be handled as a separate issue.

PR #9669, opened 2020-08-05
Fix netdata/netdata Docker image size

prologic:
##### Summary

Fixes the `netdata/netdata` Docker image size by ensuring that we
install Python dependencies in pre-built (_wheels_) form and only
install/build things in the `netdata/buidler` image and only include
runtim dependencies in the `netdata/base` runtime image.

Depends on:

- netdata/helper-images#84

Fixes:

- #9664 

##### Component Name

- area/packacing

##### Test Plan

```#!sh
$ dki -t --rm --entrypoint /bin/sh netdata/netdata:amd64
/ # python3
Python 3.8.5 (default, Jul 20 2020, 23:11:29)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pymongo
>>> import psycopg2
>>>
```

```#!sh
$ docker images netdata/netdata:amd64
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
netdata/netdata     amd64               37fe4941b674        7 minutes ago       314MB
```

##### Additional Information

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -60,6 +60,7 @@ FROM netdata/base:${ARCH}
 # Copy files over
 RUN mkdir -p /opt/src
 COPY --from=builder /app /
+COPY --from=builder /wheels /wheels
 COPY packaging/docker/health.sh /health.sh
 
 # Configure system
@@ -102,6 +103,9 @@ RUN \
     ln -sf /dev/stdout /var/log/netdata/access.log && \
     ln -sf /dev/stdout /var/log/netdata/debug.log && \
     ln -sf /dev/stderr /var/log/netdata/error.log
+
+# Install any Python wheels
+RUN pip install /wheels/*
 
 ENV NETDATA_PORT 19999
 EXPOSE $NETDATA_PORT


.travis.yml
```diff
@@ -262,7 +262,7 @@ jobs:
           - if [ -n "${BUILDER_NAME}" ]; then rm -rf /home/${BUILDER_NAME}/* && echo "Cleared /home/${BUILDER_NAME} directory" || echo "Failed to clean /home/${BUILDER_NAME} directory"; fi;
           - if [ -d "${PACKAGES_DIRECTORY}" ]; then rm -rf "${PACKAGES_DIRECTORY}"; fi;
 
-    - name: "Build & Publish DEB package for ubuntu/bionic"
+      name: "Build & Publish DEB package for ubuntu/bionic"
```
ilyam8, line 265: why this change?

https://github.com/prologic/netdata/blob/8c5cf82cbbb6061a6f7f0c0830d2c8276f212f6b/.travis.yml#L265-L295

prologic: This was a mistake. `git co -b ...` un existing commits. Fixed.

ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5967620&bid=19566390)

amoss:
What is the error at the end of the i386 Docker Build check?

ilyam8:
@amoss 

i think this PR blocked by https://github.com/netdata/helper-images/pull/84

prologic:
@Ferroin @ilyam8 Please review and approve while I rebase this!

ktsaou:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=5967620&bid=19566390)

prologic:
I'm going to merge this now so it gets into the next nightly. Okay with this @amoss ?

amoss:
> I'm going to merge this now so it gets into the next nightly. Okay with this @amoss ?

Sure, looks fine to me.

OdysLam:
Regarding the docker image, balena has managed to [package netdata](https://github.com/balena-io-examples/balena-netdata) into a docker image of 20mb, as it is intended to be used on IoT Edge devices.

Perhaps I could ask for some contribution upstream if we like this packaging.

Ferroin:
> Regarding the docker image, balena has managed to [package netdata](https://github.com/balena-io-examples/balena-netdata) into a docker image of 20mb, as it is intended to be used on IoT Edge devices.
> 
> Perhaps I could ask for some contribution upstream if we like this packaging.

They're also leveraging the Alpine package that we currently have no control or ownership of, which AFAICT does not include a number of features we want to be available in our official images.

There are definitely some things we can improve here, like figuring out why the stuff we copy from the builder stage to the final stage is more than 100MB (per `docker history`, that's the second largest part of our image, with the largest being all the deps we pull into the base image).

OdysLam:
> > Regarding the docker image, balena has managed to [package netdata](https://github.com/balena-io-examples/balena-netdata) into a docker image of 20mb, as it is intended to be used on IoT Edge devices.
> > Perhaps I could ask for some contribution upstream if we like this packaging.
> 
> They're also leveraging the Alpine package that we currently have no control or ownership of, which AFAICT does not include a number of features we want to be available in our official images.
> 
> There are definitely some things we can improve here, like figuring out why the stuff we copy from the builder stage to the final stage is more than 100MB (per `docker history`, that's the second largest part of our image, with the largest being all the deps we pull into the base image).

We don't publish the alpine package?  But yeah, for netdata to become more IoT friendly, we would need to address the issue of size, especially since we know that it can be done. For now, I just wanted to raise that particular repository to your attention, since I know we have a lot going on.  :)

Ferroin:
> We don't publish the alpine package? 

We aren't involved in any of the packages in the official distro repos. We publish our own DEB and RPM packages, but users have to add our repo to get those.

> But yeah, for netdata to become more IoT friendly, we would need to address the issue of size, especially since we know that it can be done.

Agreed. The solution to that IMO is to modularize the Docker images and set them up so they only include the bare minimum by default, but set it up so users can easily pull in deps for whatever features they want (for example, have them list desired features in an environment variable, and then have the image entrypoint look at that and pull in required deps). The problem is that actually doing that properly would require changes to how we handle a number of things in the core agent (namely, we would have to load exporting plugins and similar on-demand insteaf of linking them into the agent itself, and similar for some of the other things).

amoss:
What is a coarse breakdown of the 314-20=294MB in our image that is not in the Balena image?

Ferroin:
> What is a coarse breakdown of the 314-20=294MB in our image that is not in the Balena image?

I have not looked in-depth at the Alpine package (It's been mentioned before, but could we _please_ have the `netdata -V` output list what features are enabled? It would make this type of thing so much easier), but based on what I do know, they split the NodeJS and Python collectors to separate packages, don't include (or offer a way to get) the React dashboard or Go plugin, and based on dependencies does not appear to include almost any of the optional features except native TLS support for the web interface (no JSON-C, no dbengine, no Cloud support, no exporting engine, etc).

OdysLam:
@Ferroin, are we sure that the alpine packages is *that* weaker in functionality? I will be testing that image on a raspberry pi, will try both streaming, dbengine,etc.

So, in regards to this, how should I move forward so that it doesn't get buried but is considered to be tackled in the future?

Ferroin:
> @Ferroin, are we sure that the alpine packages is _that_ weaker in functionality? I will be testing that image on a raspberry pi, will try both streaming, dbengine,etc.

It's missing the dependencies that would be needed for any of the functionality I listed, so unless they did some crazy things with our build system to only load stuff when something actually tries to use it (not likely), they y definition cannot have any of that functionality.

> So, in regards to this, how should I move forward so that it doesn't get buried but is considered to be tackled in the future?

Handling the docker image as I mentioned above is something @prologic and I have discussed in passing a couple of times, and we'd kind of like to do it, but it's well beyond just a packaging issue. Actually modularizing things properly is going to need need some significant architectural changes in the agent and build system to be workable.

As far as handling of external packages, we really don't have the manpower right now to properly take over maintainership though I (and I suspect @prologic as well) would love to do so long-term so that we can ensure that everybody has access to all the features and we don't make users add custom repos just for fully featured for release builds.

PR #9685, opened 2020-08-07
Stop multi-host DB statistics from being counted multiple times.

mfundul:
##### Summary
Fix multi-host DB statistics.
##### Component Name
dbengine
##### Test Plan
Connect multiple hosts to multihost DB. Observe dbengine statistics not being magnified contrary to what it was before.

### DIFF
--- a/daemon/global_statistics.c
+++ b/daemon/global_statistics.c
@@ -537,12 +537,17 @@ void global_statistics_charts(void) {
     RRDHOST *host;
     unsigned long long stats_array[RRDENG_NR_STATS] = {0};
     unsigned long long local_stats_array[RRDENG_NR_STATS];
-    unsigned hosts_with_dbengine = 0, i;
+    unsigned dbengine_contexts = 0, counted_multihost_db = 0, i;
 
     rrd_rdlock();
     rrdhost_foreach_read(host) {
-        if (host->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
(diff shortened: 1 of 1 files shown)


daemon/global_statistics.c
```diff
@@ -537,12 +537,17 @@ void global_statistics_charts(void) {
...
-    unsigned hosts_with_dbengine = 0, i;
+    unsigned dbengine_contexts = 0, counted_multihost_db = 0, i;
 
     rrd_rdlock();
     rrdhost_foreach_read(host) {
-        if (host->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
-            ++hosts_with_dbengine;
+        if (host->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE && !rrdhost_flag_check(host, RRDHOST_FLAG_ARCHIVED)) {
+            if (&multidb_ctx == host->rrdeng_ctx) {
+                if (counted_multihost_db)
```
amoss, line 546: There is always exactly one multihost db - why check this in the loop instead of just skipping the multidb_ctx and adding one at the end?

mfundul: There are cases where we don't want that. E.g. localhost `memory mode = none` and the children are in legacy memory mode.

amoss: You have two conditions inside the loop. The outer one handles the case that you describe. The inner `if` is invariant and should be lifted out of the loop.

PR #9711, opened 2020-08-11
Set the default value of the home directory to the environment's HOME

### LINKED ISSUE #6865: Config AWS SNS Issue
##### Question summary
Netdata can't publish the message to SNS, but I try "aws sns ...", its success.
When try to trigger the alarm, error log show me as follow:
2019-09-17 07:20:17: netdata INFO  : MAIN : SIGNAL: Received SIGUSR2. Reloading HEALTH configuration...
  File "/var/lib/netdata/.local/bin/aws", line 19, in <module>
    import awscli.clidriver
ImportError: No module named awscli.clidriver
2019-09-17 07:20:18: alarm-notify.sh: ERROR: failed to send Amazon SNS notification for:
##### OS / Environment
# su netdata
sh-4.2$ echo $PATH
/var/lib/netdata/.local/bin:/root/.local/bin:/usr/local/sbin:/sbin:/bin:/usr/sbin:/usr/bin
sh-4.2$ aws --version
aws-cli/1.16.239 Python/2.7.5 Linux/3.10.0-957.1.3.el7.x86_64 botocore/1.12.229
sh-4.2$ aws sns publish --region us-east-2 --subject "test" --message "test" --target-arn myarn
{
    "MessageId": "fdc04413-4619-56c1-9507-11e9cb8f78f1"
}

cakrit:
##### Summary
Fixes #6865 (?)
For the default environment value of "HOME", use the user's "HOME", as defined in the environment.

##### Component Name
daemon

##### Test Plan
Test in various environments, including AWS SNS.

##### Additional Information
I don't know how to properly test this, I'll leave it up to @Ferroin

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -541,7 +541,7 @@ static void get_netdata_configured_variables() {
     netdata_configured_web_dir          = config_get(CONFIG_SECTION_GLOBAL, "web files directory",    netdata_configured_web_dir);
     netdata_configured_cache_dir        = config_get(CONFIG_SECTION_GLOBAL, "cache directory",        netdata_configured_cache_dir);
     netdata_configured_varlib_dir       = config_get(CONFIG_SECTION_GLOBAL, "lib directory",          netdata_configured_varlib_dir);
-    netdata_configured_home_dir         = config_get(CONFIG_SECTION_GLOBAL, "home directory",         netdata_configured_home_dir);
+    netdata_configured_home_dir         = config_get(CONFIG_SECTION_GLOBAL, "home directory",         getenv("HOME"));
 
     netdata_configured_lock_dir = initialize_lock_directory_path(netdata_configured_varlib_dir);
 
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -541,7 +541,8 @@ static void get_netdata_configured_variables() {
     netdata_configured_web_dir          = config_get(CONFIG_SECTION_GLOBAL, "web files directory",    netdata_configured_web_dir);
     netdata_configured_cache_dir        = config_get(CONFIG_SECTION_GLOBAL, "cache directory",        netdata_configured_cache_dir);
     netdata_configured_varlib_dir       = config_get(CONFIG_SECTION_GLOBAL, "lib directory",          netdata_configured_varlib_dir);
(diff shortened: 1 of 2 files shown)


cakrit:
Doesn't seem to be picking up the HOME env. var yet, checking

cakrit:
Nope, it does work, it correctly brings in my HOME environment

daemon/main.c
```diff
@@ -541,7 +541,7 @@ static void get_netdata_configured_variables() {
     netdata_configured_web_dir          = config_get(CONFIG_SECTION_GLOBAL, "web files directory",    netdata_configured_web_dir);
     netdata_configured_cache_dir        = config_get(CONFIG_SECTION_GLOBAL, "cache directory",        netdata_configured_cache_dir);
     netdata_configured_varlib_dir       = config_get(CONFIG_SECTION_GLOBAL, "lib directory",          netdata_configured_varlib_dir);
-    netdata_configured_home_dir         = config_get(CONFIG_SECTION_GLOBAL, "home directory",         netdata_configured_home_dir);
+    netdata_configured_home_dir         = config_get(CONFIG_SECTION_GLOBAL, "home directory",         getenv("HOME"));
```
mfundul, line 544: If `getenv("HOME")` is empty it should default to `netdata_configured_home_dir` which is `CACHE_DIR`.

cakrit: Sure, lemme do that.

amoss:
I don't think it needs major testing, the effect is quite small and predictable. @cakrit Can you point your PR at `develop` instead of `master`?

daemon/main.c
Ferroin: Is there a particular reason we're using the cache directory here instead of `/var/lib/netdata` like we configure as the actual home directory for the Netdata user?

cakrit:
@mfundul and @amoss I modifed `common.c` as well, to have VARLIB_DIR be the original default. See https://github.com/netdata/netdata/issues/6865#issuecomment-671893053

mfundul:
The code looks good to me.

Ferroin:
Between both honoring `$HOME` when it's already set and updating the default to match what we're settin for the use ron install, this should indeed fix #6865.

amoss:
Merging this as if we do a point release on the stable channel this should be included.

PR #9712, opened 2020-08-11
Fixed issue with missing alarms

### LINKED ISSUE #9706: Health stopped working
##### Bug report summary
While I was testing the PR #9698 I observed that initially I did not receive any alarm. After this I decided to clean up the `dbengine` files and starting my tests again, it worked during few hours, but after I change from a branch to another and recompile I stopped to receive any alarm.

In the first time that the problem happend, I observed that I had thousands of errors like this:

```
Ignoring metric with out of order timestamp
``` 

but, when the problem happens again, I did not receive more this error message.

##### OS / Environment
```
Linux hades 5.4.57 #1 SMP Sat Aug 8 12:10:00 UTC 2020 x86_64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz GenuineIntel GNU/Linux
/etc/os-release:ID=slackware
```

##### Netdata version
netdata v1.24.0-8-ga15e450e2
 

##### Component Name
Health
Dbengine

##### Steps To Reproduce
1. Start Netdata
2. Configure an alarm, for example:

```
 alarm: dev_dim_template
    on: system.cpu
    os: linux
 units: %
 every: 10s
  warn: $this > 1
  crit: $this > 4
```
3. Stop Netdata ( I did not test sending -USR2, all the problems happening when I was stopping Netdata, recompiling and starting it again)
4. Verify alarms and logs.

##### Expected behavior
To have alarms running, independent of the Netdata version or branch, on my case, I used the cited PR and the master branch, and I had problems with both.

stelfrag:
Fixes #9706
##### Summary
When the health subsystem is configured, archived charts are skipped to avoid unnecessary configuration that slows
down the metadata log replay. 

Health related configuration for those charts is performed when a chart switches from archived to active by having a collector gather metrics for any dimension of that chart.

The following init code (found in the health configuration initialization) did not run in that case

```
rrddim_foreach_read(rd, st) {
    rrdcalc_link_to_rrddim(rd, st, host);
}
```

##### Component Name
database
health

##### Test Plan
- Before the PR
- Configure an alarm as described in #9706 
- Restart the agent 
   - The alarm is not raised
- Perform health reload
- Alarm is now raised
  - Note that during health reload the init process is done properly because the chart is not archived anymore

- Apply the PR    
- Restart agent
- Alarm is now immediately raised

### DIFF
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -19,7 +19,7 @@ static inline void calc_link_to_rrddim(RRDDIM *rd)
             } else if (hostlocked != EBUSY) {
                 error("Cannot lock host to create an alarm for the dimension.");
             }
-            usleep(200000);
+            sleep_usec(USEC_PER_MS * 200);
         }
 
         if (count == 5) {
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -452,7 +452,6 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
     if(unlikely(rrddim_index_add(st, rd) != rd))
         error("RRDDIM: INTERNAL ERROR: attempt to index duplicate dimension '%s' on chart '%s'", rd->id, st->id);
 
-
     if (!is_archived)
         calc_link_to_rrddim(rd);
 
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -3,6 +3,32 @@
 #define NETDATA_RRD_INTERNALS
 #include "rrd.h"
 
+static inline void calc_link_to_rrddim(RRDDIM *rd)
+{
+    RRDHOST *host = rd->rrdset->rrdhost;
+    RRDSET  *st = rd->rrdset;
+    if (host->alarms_with_foreach || host->alarms_template_with_foreach) {
+        int count = 0;
+        int hostlocked;
+        for (count = 0; count < 5; count++) {
+            hostlocked = netdata_rwlock_trywrlock(&host->rrdhost_rwlock);
+            if (!hostlocked) {
+                rrdcalc_link_to_rrddim(rd, st, host);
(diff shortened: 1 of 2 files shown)


stelfrag:
@mfundul you are correct. Needs to move to dim creation

mfundul:
@stelfrag you should run coverity on your branch since you moved flagged code.

database/rrddim.c
```diff
@@ -3,6 +3,32 @@
...
+        for (count = 0; count < 5; count++) {
+            hostlocked = netdata_rwlock_trywrlock(&host->rrdhost_rwlock);
+            if (!hostlocked) {
+                rrdcalc_link_to_rrddim(rd, st, host);
+                rrdhost_unlock(host);
+                break;
+            } else if (hostlocked != EBUSY) {
+                error("Cannot lock host to create an alarm for the dimension.");
+            }
+            usleep(200000);
```
thiagoftsm, line 22: I will give a suggestion here, instead to use `usleep`, please, change to `sleep_usec` that is declared inside `libnetdata/clocks/clocks.c`.

stelfrag:
> @stelfrag you should run coverity on your branch since you moved flagged code.

yes it produces a warning about deadlock in that part of code (currently marked as false positive in netdata/netdata)

database/rrddim.c
stelfrag: Thanks @thiagoftsm -- I changed it

amoss:
> > @stelfrag you should run coverity on your branch since you moved flagged code.
> 
> yes it produces a warning about deadlock in that part of code (currently marked as false positive in netdata/netdata)

If the old flag had an owner then get them to recheck the moved code, otherwise it predates people owning flags - in which case: congratulations! Please verify it does not deadlock (with two confirmations) and put your name in the field.

PR #9729, opened 2020-08-12
Added code to release memory used by the global GUID map

### LINKED ISSUE #9728: Release the memory allocated for the global GUID map on agent shutdown
#### Summary

When the agent exits the memory allocated for the global GUID map is not released. Although that is not causing a issue
because the memory is reclaimed by the operating system anyway, using e.g `valgrind` to detect agent memory leaks pollutes the output with dozens of entries related to the global GUID map allocations.

e.g.

```
==2734329== 24 bytes in 1 blocks are possibly lost in loss record 1,121 of 2,895
==2734329==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2734329==    by 0x4CAD604: ??? (in /usr/lib/x86_64-linux-gnu/libJudy.so.1.0.3)
==2734329==    by 0x1CE9B3: guid_store_nolock (global_uuid_map.c:54)
==2734329==    by 0x1CF27A: find_or_generate_guid (global_uuid_map.c:243)
==2734329==    by 0x1A6C68: rrdhost_create (rrdhost.c:297)
==2734329==    by 0x1CC242: metalog_pluginsd_host_action (metalogpluginsd.c:68)
==2734329==    by 0x1958C4: metalog_pluginsd_host (pluginsd_parser.c:663)
==2734329==    by 0x1ECAE9: parser_action (parser.c:296)
==2734329==    by 0x1CA7F8: replay_record (logfile.c:411)
==2734329==    by 0x1CAED5: iterate_records (logfile.c:527)
==2734329==    by 0x1CB0B8: load_metadata_logfile (logfile.c:563)
==2734329==    by 0x1CB9A9: scan_metalog_files (logfile.c:696)
==2734329== 
==2734329== 32 bytes in 1 blocks are possibly lost in loss record 1,284 of 2,895
==2734329==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2734329==    by 0x4CAD604: ??? (in /usr/lib/x86_64-linux-gnu/libJudy.so.1.0.3)
==2734329==    by 0x1CEB0E: guid_store_nolock (global_uuid_map.c:83)
==2734329==    by 0x1CF27A: find_or_generate_guid (global_uuid_map.c:243)
==2734329==    by 0x1A6C68: rrdhost_create (rrdhost.c:297)
==2734329==    by 0x1A80CE: rrd_init (rrdhost.c:634)
==2734329==    by 0x129DC3: main (main.c:1448)
==2734329== 
==2734329== 32 bytes in 1 blocks are possibly lost in loss record 1,285 of 2,895
==2734329==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2734329==    by 0x4CAD604: ??? (in /usr/lib/x86_64-linux-gnu/libJudy.so.1.0.3)
==2734329==    by 0x1CEB0E: guid_store_nolock (global_uuid_map.c:83)
==2734329==    by 0x1CF27A: find_or_generate_guid (global_uuid_map.c:243)
==2734329==    by 0x1A6C68: rrdhost_create (rrdhost.c:297)
==2734329==    by 0x1CC242: metalog_pluginsd_host_action (metalogpluginsd.c:68)
==2734329==    by 0x1958C4: metalog_pluginsd_host (pluginsd_parser.c:663)
==2734329==    by 0x1ECAE9: parser_action (parser.c:296)
==2734329==    by 0x1CA7F8: replay_record (logfile.c:411)
==2734329==    by 0x1CAED5: iterate_records (logfile.c:527)
==2734329==    by 0x1CB0B8: load_metadata_logfile (logfile.c:563)
==2734329==    by 0x1CB9A9: scan_metalog_files (logfile.c:696)
```

stelfrag:
Fixes #9728
##### Summary
Cleanup the memory allocated for the global GUID map and release the associated Judy arrays.

##### Component Name
database

##### Test Plan
- Run with `valgrind` or address sanitizer before and after the cleanup code

##### Additional information
As this uses memory to keep track of the objects we need to release, we may want to enable this with a compile flag so that the code is activated only during debugging.

### DIFF
--- a/database/engine/global_uuid_map/global_uuid_map.c
+++ b/database/engine/global_uuid_map/global_uuid_map.c
@@ -10,12 +10,10 @@ static uv_rwlock_t global_lock;
 
 static void free_single_uuid(uuid_t *uuid)
 {
-    char uuid_str[37];
     Pvoid_t *PValue, *PValue1;
     char *existing_object;
     Word_t  size;
 
-    uuid_unparse_lower(*uuid, uuid_str);
     PValue = JudyHSGet(JGUID_map, (void *) uuid, (Word_t) sizeof(uuid_t));
     if (likely(PValue)) {
         existing_object = *PValue;
--- a/database/engine/global_uuid_map/global_uuid_map.c
+++ b/database/engine/global_uuid_map/global_uuid_map.c
@@ -8,6 +8,13 @@ static uv_rwlock_t guid_lock;
 static uv_rwlock_t object_lock;
 static uv_rwlock_t global_lock;
 
+
+void free_global_guid_map()
+{
+    JudyHSFreeArray(&JGUID_map, PJE0);
+    JudyHSFreeArray(&JGUID_object_map, PJE0);
+}
+
 static void free_single_uuid(uuid_t *uuid)
 {
     Pvoid_t *PValue, *PValue1;
@@ -36,11 +43,15 @@ void free_uuid(uuid_t *uuid)
     char object[49];
 
     ret = find_object_by_guid(uuid, object, sizeof(object));
-    if (GUID_TYPE_DIMENSION == ret)
+    if (GUID_TYPE_DIMENSION == ret) {
         free_single_uuid((uuid_t *)(object + 16 + 16));
+        info("FREE DIM");
+    }
 
-    if (GUID_TYPE_CHART == ret)
-        free_single_uuid((uuid_t *)(object + 16));
+    if (GUID_TYPE_CHART == ret) {
+            free_single_uuid((uuid_t *)(object + 16));
+            info("FREE CHART");
+    }
 
     free_single_uuid(uuid);
     return;
--- a/database/engine/global_uuid_map/global_uuid_map.c
+++ b/database/engine/global_uuid_map/global_uuid_map.c
@@ -43,15 +43,11 @@ void free_uuid(uuid_t *uuid)
     char object[49];
 
     ret = find_object_by_guid(uuid, object, sizeof(object));
-    if (GUID_TYPE_DIMENSION == ret) {
+    if (GUID_TYPE_DIMENSION == ret)
         free_single_uuid((uuid_t *)(object + 16 + 16));
-        info("FREE DIM");
-    }
 
-    if (GUID_TYPE_CHART == ret) {
+    if (GUID_TYPE_CHART == ret)
             free_single_uuid((uuid_t *)(object + 16));
-            info("FREE CHART");
-    }
 
     free_single_uuid(uuid);
     return;
--- a/database/engine/global_uuid_map/global_uuid_map.c
+++ b/database/engine/global_uuid_map/global_uuid_map.c
@@ -153,34 +153,6 @@ static inline int guid_store_nolock(uuid_t *uuid, void *object, GUID_TYPE object
 }
 
 
-inline int guid_store(uuid_t *uuid, char *object, GUID_TYPE object_type)
-{
-    uv_rwlock_wrlock(&global_lock);
-    int rc = guid_store_nolock(uuid, object, object_type);
-    uv_rwlock_wrunlock(&global_lock);
-    return rc;
-}
-
-/*
- * This can be used to bulk load entries into the global map
- *
- * A lock must be aquired since it will call guid_store_nolock
- * with a "no lock" parameter.
- *
- * Note: object memory must be allocated by caller and not released
- */
-int guid_bulk_load(char *uuid, char *object)
-{
-    uuid_t target_uuid;
-    if (likely(!uuid_parse(uuid, target_uuid))) {
-#ifdef NETDATA_INTERNAL_CHECKS
-        debug(D_GUIDLOG,"Mapping GUID [%s] on [%s]", uuid, object);
-#endif
-        return guid_store_nolock(&target_uuid, object, GUID_TYPE_CHAR);
-    }
-    return 1;
-}
-
 /*
  * Given a GUID, find if an object is stored
  *   - Optionally return the object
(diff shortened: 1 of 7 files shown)


database/rrddim.c
```diff
@@ -527,6 +527,10 @@ void rrddim_free_custom(RRDSET *st, RRDDIM *rd, int db_rotated)
             debug(D_RRD_CALLS, "Removing dimension '%s'.", rd->name);
             freez((void *)rd->id);
             freez(rd->cache_filename);
+#ifdef ENABLE_DBENGINE
+            freez(rd->state->metric_uuid);
```
amoss, line 531: Why is the metric_uuid being freed for ALLOC and NONE?

database/rrdset.c
```diff
@@ -388,6 +388,7 @@ void rrdset_free(RRDSET *st) {
         case RRD_MEMORY_MODE_ALLOC:
         case RRD_MEMORY_MODE_NONE:
         case RRD_MEMORY_MODE_DBENGINE:
+            freez(st->chart_uuid);
```
amoss, line 391: I can only see an allocation site for DBENGINE, why is this freed for ALLOC and NONE?

amoss:
Let me check that I've understood the big-picture behind what you're doing in this change.

* JudySL only stores strings and cannot contain NUL-bytes in the key.
* JudyHL allows keys with NUL-bytes but there is no interface to iterate over the keys, and no hook to free up the value when freeing the whole structure.
* The MAP_STORE is being used as an auxiliary store to track the pointers.
* The global_lock for the GUID-map is serializing access to the MAP_STORE so the code does not need to be thread-safe.
* The MAP_STORE is a monotonically growing list of what is stored in the Judy array.

If I've understood correctly then I'm concerned about what we are doing. We are removing the static analysis errors by freeing the values when the process terminates. But in order to do so we are:

* Using more memory.
* The size of the MAP_STORE will increase during execution of the agent so it will behave like a memory leak if the agent is running for a long duration.

database/rrddim.c
stelfrag: That's wrong I will fix it

database/rrdset.c
stelfrag: That's wrong I will fix it

stelfrag:
> Let me check that I've understood the big-picture behind what you're doing in this change.
> 
> * JudySL only stores strings and cannot contain NUL-bytes in the key.
> * JudyHL allows keys with NUL-bytes but there is no interface to iterate over the keys, and no hook to free up the value when freeing the whole structure.
> * The MAP_STORE is being used as an auxiliary store to track the pointers.
> * The global_lock for the GUID-map is serializing access to the MAP_STORE so the code does not need to be thread-safe.
> * The MAP_STORE is a monotonically growing list of what is stored in the Judy array.
> 
> If I've understood correctly then I'm concerned about what we are doing. We are removing the static analysis errors by freeing the values when the process terminates. But in order to do so we are:
> 
> * Using more memory.
> * The size of the MAP_STORE will increase during execution of the agent so it will behave like a memory leak if the agent is running for a long duration.

Yes we are using more memory to keep track what to release (which would be released anyway) as the size of the MAP_STORE will eventually max as the number of charts / dimensions stabilizes 

That's why I say we may want to enable this only during debugging (if tracking memory leaks) so that it keeps the valgrind's output cleaner.

Now I think about it, I guess it could work without storing extra memory by using the information stored in  `rd->state->metric_uuid`

amoss:
Can you tie it to the normal `_destroy` calls for the structures? (i.e. instead of holding an extra list attempt to free the uuid and remove it from the map when the structure is destroyed)

stelfrag:
> Can you tie it to the normal `_destroy` calls for the structures? (i.e. instead of holding an extra list attempt to free the uuid and remove it from the map when the structure is destroyed)

Yes I will try to do that

database/rrddim.c
```diff
@@ -527,6 +527,11 @@ void rrddim_free_custom(RRDSET *st, RRDDIM *rd, int db_rotated)
             debug(D_RRD_CALLS, "Removing dimension '%s'.", rd->name);
             freez((void *)rd->id);
             freez(rd->cache_filename);
+#ifdef ENABLE_DBENGINE
+            if (rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE)
+                freez(rd->state->metric_uuid);
```
amoss, line 532: At this point we know the `uuid` and the object address. A call int othe dbengine could delete the entry from the map and free the memory - then we don't need to track the global list.

amoss: (and the second call site obviously :) )

database/engine/global_uuid_map/global_uuid_map.c
```diff
@@ -8,6 +8,46 @@ static uv_rwlock_t guid_lock;
...
 static uv_rwlock_t global_lock;
 
+static void free_single_uuid(uuid_t *uuid)
+{
+    char uuid_str[37];
+    Pvoid_t *PValue, *PValue1;
+    char *existing_object;
+    Word_t  size;
+
+    uuid_unparse_lower(*uuid, uuid_str);
```
amoss, line 18: `uuid_str` doesn't get used?

stelfrag: blah.. will remove ; was part of debug code

stelfrag:
Missing the final judy array destruction now -- will fix

amoss:
It looks like there is a missing `uv_rwlock_rdunlock(&global_lock)` in `find_object_by_guid` https://github.com/netdata/netdata/blob/bca54e920c918bd0a8ce9edbdaad5e43c6aa4c42/database/engine/global_uuid_map/global_uuid_map.c#L172

amoss:
Same on https://github.com/netdata/netdata/blob/bca54e920c918bd0a8ce9edbdaad5e43c6aa4c42/database/engine/global_uuid_map/global_uuid_map.c#L178

amoss (approved):
The call-sites look correct and the assumptions about the memory layout match up with the contents of `guid_store_nolock` and ` find_object_by_guid` (although the difference of one byte at the beginning is confusing). Please check the missing unlocks and insert them if they are necessary.

stelfrag:
> The call-sites look correct and the assumptions about the memory layout match up with the contents of `guid_store_nolock` and ` find_object_by_guid` (although the difference of one byte at the beginning is confusing). Please check the missing unlocks and insert them if they are necessary.

Yes the locks are needed for completeness (all the calls to this unction provide enough space -- RRD_ID_LENGTH_MAX for the CHAR type)

I will add  the missing unlocks and rebase if CI/CD fails.

amoss (approved):
Reapproving

vlvkobal (approved):

PR #9738, opened 2020-08-13
Fix flushing errors

### LINKED ISSUE #8384: Repeated dbengine global flushing errors for statsd

### LINKED ISSUE #9730: using dbengine multihost results in connection/data loss to children as a result of "too slow disk"
##### Bug report summary

We have a Parent where 37 children are streaming around 3700 metrics each 2 seconds to parent. With the old dbengine setup per child everything is working fine. After upgrading to 1.24 and deleting the old db cache we now had problems and alerts lik "ipv4 tcphandshake last collected secs = 00:02:10 ago" for many children periodically after ~20 minutes. Looking into the dbengine netdata metrics revealed that the parent reported at the same time that the disk was to slow to write all metrics from the dbengin to disk. At the same time and during the time where data was not gather from children, netdata reported higher CPU usage.

My assumptions is now that before, each dbengin wrote to disk a different times smaller size of data and the disk was fast enough for netdata to write everything, and with the new single dbengine the data is now written at the same time which causes problems on netdata side.

To mitigate the data loss we reverted back to 1.23.2 on the parent. All children continue running on 1.24.0 (stable release channel)

##### OS / Environment

netdata in docker container on ubuntu server

##### Netdata version

1.24.0 docker
 

##### Component Name

dbengine

##### Steps To Reproduce

1. ...
2. ...

##### Expected behavior

Keep the same disk/io requirements for new dbengine which worked with old seperated dbengins.

mfundul:
##### Summary
Fixes  #8384
Fixes  #9730
##### Component Name
database
##### Test Plan

1. Make sure that `2 * 4096 * number-of-metrics-being-collected` bytes is more than the disk space quota. This should trigger the `Failed to flush dirty buffers quickly enough in dbengine instance` error message in the old code.
2. Make sure that a collector that has collected at least 34 minutes worth of metric data points stops collecting a specific metric (2 4K pages of metrics with 1 second collection interval). After a full database rotation you should start seeing the aforementioned error message.

### DIFF
--- a/database/engine/pagecache.c
+++ b/database/engine/pagecache.c
@@ -123,6 +123,24 @@ void pg_cache_wait_event_unsafe(struct rrdeng_page_descr *descr)
 }
 
 /*
+ * The caller must hold page descriptor lock.
+ * The lock will be released and re-acquired. The descriptor is not guaranteed
+ * to exist after this function returns.
+ * Returns UV_ETIMEDOUT if timeout_sec seconds pass.
+ */
+int pg_cache_timedwait_event_unsafe(struct rrdeng_page_descr *descr, uint64_t timeout_sec)
+{
+    int ret;
+    struct page_cache_descr *pg_cache_descr = descr->pg_cache_descr;
+
+    ++pg_cache_descr->waiters;
+    ret = uv_cond_timedwait(&pg_cache_descr->cond, &pg_cache_descr->mutex, timeout_sec * NSEC_PER_SEC);
+    --pg_cache_descr->waiters;
+
+    return ret;
+}
+
+/*
  * Returns page flags.
  * The lock will be released and re-acquired. The descriptor is not guaranteed
  * to exist after this function returns.
--- a/database/engine/pagecache.c
+++ b/database/engine/pagecache.c
@@ -444,6 +444,7 @@ uint8_t pg_cache_punch_hole(struct rrdengine_instance *ctx, struct rrdeng_page_d
             if (unlikely(debug_flags & D_RRDENGINE))
                 print_page_cache_descr(descr);
             pg_cache_wait_event_unsafe(descr);
+//metalog_print_dimension_by_uuid(ctx->metalog_ctx, page_index->id);
         }
     }
     if (remove_dirty) {
@@ -455,6 +456,7 @@ uint8_t pg_cache_punch_hole(struct rrdengine_instance *ctx, struct rrdeng_page_d
             if (unlikely(debug_flags & D_RRDENGINE))
                 print_page_cache_descr(descr);
             pg_cache_wait_event_unsafe(descr);
+//metalog_print_dimension_by_uuid(ctx->metalog_ctx, page_index->id);
         }
     }
     rrdeng_page_descr_mutex_unlock(ctx, descr);
--- a/database/engine/pagecache.c
+++ b/database/engine/pagecache.c
@@ -462,7 +462,6 @@ uint8_t pg_cache_punch_hole(struct rrdengine_instance *ctx, struct rrdeng_page_d
             if (unlikely(debug_flags & D_RRDENGINE))
                 print_page_cache_descr(descr);
             pg_cache_wait_event_unsafe(descr);
-//metalog_print_dimension_by_uuid(ctx->metalog_ctx, page_index->id);
         }
     }
     if (remove_dirty) {
@@ -474,7 +473,6 @@ uint8_t pg_cache_punch_hole(struct rrdengine_instance *ctx, struct rrdeng_page_d
             if (unlikely(debug_flags & D_RRDENGINE))
                 print_page_cache_descr(descr);
             pg_cache_wait_event_unsafe(descr);
-//metalog_print_dimension_by_uuid(ctx->metalog_ctx, page_index->id);
         }
     }
     rrdeng_page_descr_mutex_unlock(ctx, descr);
--- a/database/engine/rrdengine.c
+++ b/database/engine/rrdengine.c
@@ -670,7 +670,7 @@ void rrdeng_test_quota(struct rrdengine_worker_config* wc)
 
     out_of_space = 0;
     /* Do not allow the pinned pages to exceed the disk space quota to avoid deadlocks */
-    if (unlikely(ctx->disk_space > ctx->max_disk_space + 2 * ctx->metric_API_max_producers * RRDENG_BLOCK_SIZE)) {
+    if (unlikely(ctx->disk_space > MAX(ctx->max_disk_space, 2 * ctx->metric_API_max_producers * RRDENG_BLOCK_SIZE))) {
         out_of_space = 1;
     }
     datafile = ctx->datafiles.last;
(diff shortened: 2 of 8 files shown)


amoss:
I don't think that this looks like a fix, at best it seems to be a mitigation that hides the problem in the specific testing scenario that you have indicated. Why was the extra page pinned originally and what are we losing by disabling it?

stelfrag:
> I don't think that this looks like a fix, at best it seems to be a mitigation that hides the problem in the specific testing scenario that you have indicated. Why was the extra page pinned originally and what are we losing by disabling it?

As discussed with @mfundul -- it's not as a fix but disabling a feature that doesn't work correctly. The pinning was done for performance reasons so that the active page and the previous (pinned one) are available in memory for queries (for a maximum of ~34 minutes worth of data)

Unpinning will result in the page being brought in back from the disk if we get an incoming query that needs to access that data (and the page is not available)

Given the headaches it causes I think its best to disable it (at least for now until we figure out how it can be improved / reworked).

database/engine/rrdengine.c
```diff
@@ -669,7 +669,8 @@ void rrdeng_test_quota(struct rrdengine_worker_config* wc)
     int ret, error;
 
     out_of_space = 0;
-    if (unlikely(ctx->disk_space > ctx->max_disk_space)) {
+    /* Do not allow the pinned pages to exceed the disk space quota to avoid deadlocks */
+    if (unlikely(ctx->disk_space > MAX(ctx->max_disk_space, 2 * ctx->metric_API_max_producers * RRDENG_BLOCK_SIZE))) {
```
amoss, line 673: Do we still need this if pinning is disabled?

database/engine/pagecache.c
```diff
@@ -122,6 +122,24 @@ void pg_cache_wait_event_unsafe(struct rrdeng_page_descr *descr)
...
+    int ret;
+    struct page_cache_descr *pg_cache_descr = descr->pg_cache_descr;
+
+    ++pg_cache_descr->waiters;
+    ret = uv_cond_timedwait(&pg_cache_descr->cond, &pg_cache_descr->mutex, timeout_sec * NSEC_PER_SEC);
+    --pg_cache_descr->waiters;
+
+    return ret;
+}
+
```
amoss, line 142: This code does not get called.

amoss:
Removing the pinned page should not reduce the performance in any scenario other than querying data that is between 34 and 17 minutes old on a system that is loaded heavily enough that those pages are not left in the cache. This seems reasonable as a trade for fixing the db-rotation to remove the flushing errors.

Please clean up the commit - it adds code that is not used. I can understand leaving the debug in as it may be useful again, but why the new function that is unused?

database/engine/rrdengine.c
mfundul: Not really but the idea was that we'll re-enable pinning as soon as possible, since Kosta has explicitly asked for it, so it's good to have a solution in-place. Since it affects disk space usage for users it's a good idea to let them get used to the new disk space calculation early on.

database/engine/pagecache.c
mfundul: I know, I was experimenting with it when we discussed about timeouts and I left it there as a library function for future use. We have had other functions like that in the past, that we eventually used.

database/engine/rrdengine.c
mfundul: Moreover, it's a good idea that disk space is always bigger than cache space. It won't be pretty if the user only keeps 1% of the real time metrics because the disk space is 64MB but 1000 child nodes are streaming to it.

amoss: I think that we have to view re-enabling pinning as a piece of work to start from scratch. Pinned pages cannot prevent the rotation so there will need to be a rethink of how pinned pages work so that they can survive the cleanup and get moved to a later file / extent in the db.

mfundul: Agreed, right now the problem is that we don't remember which collectors came and went, and we don't have a strict reference counting model for them. In theory, if rotation could pre-empt a collectors reference without causing a crash, we could re-enable pinning.

PR #9751, opened 2020-08-14
Add check for spurious wakeups

vlvkobal:
##### Summary
@mfundul thanks for pointing this out

http://docs.libuv.org/en/v1.x/threading.html#conditions
> Callers should be prepared to deal with spurious wakeups on uv_cond_wait() and uv_cond_timedwait().

https://linux.die.net/man/3/pthread_cond_wait
> When using condition variables there is always a Boolean predicate involving shared variables associated with each condition wait that is true if the thread should proceed. Spurious wakeups from the pthread_cond_timedwait() or pthread_cond_wait() functions may occur. Since the return from pthread_cond_timedwait() or pthread_cond_wait() does not imply anything about the value of this predicate, the predicate should be re-evaluated upon such return.

##### Component Name
exporting engine

### DIFF
--- a/exporting/aws_kinesis/aws_kinesis.c
+++ b/exporting/aws_kinesis/aws_kinesis.c
@@ -103,7 +103,10 @@ void aws_kinesis_connector_worker(void *instance_p)
         struct stats *stats = &instance->stats;
 
         uv_mutex_lock(&instance->mutex);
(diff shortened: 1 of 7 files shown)


mfundul:
Code looks fine. The instance is implicitly zeroed during allocation, which is not immediately obvious.

PR #9240, opened 2020-06-01
python.d/isc_dhcpd: add support for ip range in pool

vsc55:
##### Summary
info package is needed

##### Component Name
collectors/python.d.plugin/isc_dhcpd/

PR #9755, opened 2020-08-14
python.d/isc_dhcpd: add support for ip ranges

vsc55:
##### Summary
info package is needed

##### Component Name
collectors/python.d.plugin/isc_dhcpd/

### DIFF
--- a/collectors/python.d.plugin/isc_dhcpd/isc_dhcpd.chart.py
+++ b/collectors/python.d.plugin/isc_dhcpd/isc_dhcpd.chart.py
@@ -45,6 +45,19 @@ CHARTS = {
     }
 }
 
+POOL_CIDR = "CIDR"
+POOL_IP_RANGE = "IP_RANGE"
+POOL_UNKNOWN = "UNKNOWN"
+
+def detect_ip_type(ip):
+    ip_type = ip.split("-")
+    if len(ip_type) == 1:
+        return POOL_CIDR
+    elif len(ip_type) == 2:
+        return POOL_IP_RANGE
+    else:
+        return POOL_UNKNOWN
+
 
 class DhcpdLeasesFile:
     def __init__(self, path):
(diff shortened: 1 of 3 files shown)


ilyam8:
tested in https://github.com/netdata/netdata/pull/9240#issuecomment-670953620

collectors/python.d.plugin/isc_dhcpd/isc_dhcpd.chart.py
```diff
@@ -95,6 +134,30 @@ def __contains__(self, item):
         return item.address in self.network
 
 
+class PoolIPRange:
+    def __init__(self, ip_range):
+        ip_range = ip_range.split("-")
+        self.networks = list(self._summarize_address_range(ip_range[0], ip_range[1]))
```
amoss, line 140: This is potentially very large and instantiating the list will cost at least 4-bytes + all the Python overhead per element. So for an address range in a `/8` this could be 100s of MB. The IPv4Address class support comparators so if you store the low and high values then you can test if it is in the range with `((x>=lo) and (x<=hi))` and it will be much faster and take no memory.

ilyam8: > This is potentially very large and instantiating the list will cost at least 4-bytes + all the Python overhead per element.

> ((x>=lo) and (x<=hi))

yes, i use this method in `dnsmasq_dhcp` collector:
 > https://github.com/netdata/go.d.plugin/blob/5f61c42fdd4e0e4b4801d4ac574cc78a5f9a0dec/modules/dnsmasq_dhcp/ip/range.go#L90-L93

---

However i think this is fine and can be improved later. I assume ip ranges usually not huge, it is more common to use cidr.

amoss: The CIDR case looks fine as it checks if the ip-address is inside the network range directly. If you are already handling this in the go code then it will get fixed during migration. LGTM

PR #9780, opened 2020-08-19
Fix UNIX socket access with kickstart-static64

### LINKED ISSUE #8962: Access control defaults incorrect for Netdata served over a UNIX socket.
##### Bug report summary
Per the [documentation](https://learn.netdata.cloud/docs/agent/web/server/#binding-netdata-to-multiple-ports), no access controls should be applied by default to connections over a UNIX socket. However, at least some calls to the agent over a UNIX socket re blocked as if access controls were in place.

##### OS / Environment

```
NAME=Gentoo
ID=gentoo
PRETTY_NAME="Gentoo/Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.gentoo.org/"
SUPPORT_URL="https://www.gentoo.org/support/"
BUG_REPORT_URL="https://bugs.gentoo.org/"
```

##### Netdata version
v1.22.0
 

##### Component Name
area/web

##### Steps To Reproduce

1. Configure Netdata to listen on a UNIX socket.
2. Attempt to fetch `netdata.conf` over the UNIX socket.

##### Expected behavior

Netdata should behave consistently with the documentation.

thiagoftsm:
##### Summary
Fixes #8962 

During the tests I made, I observed that Unix socket is presented with UNKNOWN client_ip, this value was not present in the default list of simple patterns associated to the variable `web_allow_netdataconf_from`.

I have to call attention for an important thing in this PR, I could never recreate the problem using nginx and doing normal compilation, this is one of the various problems that are appearing only on kickstart-static64.

I am also using this PR to remove an old comment that is not doing any contribution for the code.

##### Component Name
Web
kickstart-static64
##### Test Plan

First test, check if nothing was breaking:

1 - Compile this branch
2 - Start Netdata
3 - Access  http://localhost:19999/netdata.conf

Stop Netdata and do the next steps to confirm the problem was really fixed.

1 - Install docker on your host and start it.
2 - Clone this PR.
3 - Run  sh ./packaging/makeself/build-x86_64-static.sh
4 - Install the final binary that will be stored inside `./artifacts`.
5 - Stop Netdata
6  - Edit `/opt/netdata/etc/netdata/netdata.conf` and change the following option:
```
[web]
  bind to = unix:/tmp/fd1.sock
```
7  - Start Netdata
8 - Configure a nginx to use the unix socket:

```
    server {
        listen       8082;
        server_name  localhost;

        location / {
            proxy_redirect off;
            proxy_set_header Host $host;
            proxy_set_header X-Forwarded-Host $host:$server_port;
            proxy_set_header X-Forwarded-Server $host;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Proto https; #New
            proxy_http_version 1.1;
            proxy_pass_request_headers on;
            proxy_set_header Connection "keep-alive";
            proxy_store off;
            gzip on;
            gzip_proxied any;
            gzip_types *;
            rewrite ^/netdata(/.*)$ $1 break;
            proxy_pass http://unix:/tmp/fd1.sock;
         }

        location = /50x.html {
            root   /var/www/html;
        }
    }
```
9 - Start Nginx
10  - Access `http://localhost:8082/netdata.conf`, If you could access, this PR fixed the problem.

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -118,6 +118,7 @@ int make_dns_decision(const char *section_name, const char *config_name, const c
     if(strcmp("heuristic",value))
         error("Invalid configuration option '%s' for '%s'/'%s'. Valid options are 'yes', 'no' and 'heuristic'. Proceeding with 'heuristic'",
               value, section_name, config_name);
+
     return simple_pattern_is_potential_name(p);
 }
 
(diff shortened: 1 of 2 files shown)


vlvkobal (approved):

amoss (approved):
If the user reconfigures the DNS lookup to be active on the netdata.conf ACL then there is a possibility that the UNKNOWN could be matches against an incoming hostname, but I can't think of any cases were this would be dangerous.

PR #9779, opened 2020-08-18
Change timestamps for global variables in Prometheus output

vlvkobal:
##### Summary
Prometheus [drops global variables from responses](https://prometheus.io/docs/prometheus/latest/querying/basics/#staleness) by default
> If no sample is found (by default) 5 minutes before a sampling timestamp, no value is returned for that time series at this point in time. 

We will provide current time as a timestamp for the global variables to be sure they are available in Prometheus' responses.

Fixes #6470

##### Component Name
Web API

##### Test Plan
Query Netdata with the following URL
```
http://localhost:19999/api/v1/allmetrics?format=prometheus&source=raw&variables=yes&help=yes
```
Timestamps for global host and chart variables should represent the current time.

PR #9790, opened 2020-08-19
Fix redirect with parameters

thiagoftsm:
##### Summary
While I was testing #9779  I observed that when Netdata was doing redirects from `http` protocol to `https` protocol, it was not adding the query string when it was present.

This PR fixes this wrong redirect.

##### Component Name
Web
##### Test Plan

1 - Create a TLS certificate inside `/etc/netdata/ssl/`
2 - Configure your netdata.conf:

```
[web]
   bind to = *=dashboard|registry|streaming|management|netdata.conf|badges *:20000=dashboard|registry|streaming|netdata.conf|badges|management^SSL=optional
```

please, pay attention for the fact that the port 19999 will always redirect requests from `HTTP` while 20000 won't do this.

3 - Start Netdata
4 - Access the following URLS, with browser network visible for you see the response received:

```
http://localhost:19999/api/v1/allmetrics?format=prometheus&source=raw&variables=yes&help=yes
http://localhost:19999/api/v1/info
http://localhost:19999

https://localhost:19999
https://localhost:19999/api/v1/alarms?all
```

For the three first URLs, it is expected you receive a `301` and to be redirected for its respective `HTTPS` addresses. For the other two, you will have direct access.

5 - Now repeat the requests, but this time using port 20000 

```
http://localhost:20000/api/v1/allmetrics?format=prometheus&source=raw&variables=yes&help=yes
http://localhost:20000/api/v1/info
http://localhost:20000

https://localhost:20000
https://localhost:20000/api/v1/alarms?all
```

This time you won't be redirected for any request.

If you have some warnings when try to access the URL 4, this is expected due browser cache, please, press CTRL+ SHIFT+R to force the reload after you accept self-signed certificate.

### DIFF
--- a/web/server/web_client.c
+++ b/web/server/web_client.c
@@ -1165,7 +1165,7 @@ void web_client_build_http_header(struct web_client *w) {
     if (w->response.code == HTTP_RESP_MOVED_PERM) {
         buffer_sprintf(w->response.header_output,
                        "HTTP/1.1 %d %s\r\n"
-                       "Location: https://%s%s",
+                       "Location: https://%s%s\r\n",
                        w->response.code, code_msg,
                        w->server_host,
                        w->last_url);
--- a/web/server/web_client.c
+++ b/web/server/web_client.c
@@ -1162,34 +1162,7 @@ void web_client_build_http_header(struct web_client *w) {
         strftime(edate, sizeof(edate), "%a, %d %b %Y %H:%M:%S %Z", tm);
     }
 
-    //char headerlocation[8328];
     if (w->response.code == HTTP_RESP_MOVED_PERM) {
-        /*
-        char *move = headerlocation;
-        memcpy(move, "\r\nLocation: https://", 20);
-        move += 20;
-
-        size_t length = strlen(w->server_host);
-        memcpy(move, w->server_host, length);
-        move += length;
-
-        length = strlen(w->last_url);
-        memcpy(move, w->last_url, length);
-        move += length;
-        if (w->url_search_path && w->separator) {
-            char *s = url_find_protocol(w->url_search_path);
-            if (s)
-                *s = '\0';
-
-            length = strlen(w->url_search_path);
-            memcpy(move, w->url_search_path, length);
-            move += length;
-        }
-
-        memcpy(move, "\r\n",  2);
-        move += 2;
-        *move = '\0';
-    */
         buffer_sprintf(w->response.header_output,
                        "HTTP/1.1 %d %s\r\n"
                        "Location: https://%s%s",
@@ -1213,11 +1186,6 @@ void web_client_build_http_header(struct web_client *w) {
                        w->origin,
                        content_type_string,
                        date);
-
-        /*
-        memcpy(headerlocation,"\r\n",2);
-        headerlocation[2]=0x00;
-         */
     }
 
     if(unlikely(web_x_frame_options))
--- a/web/server/web_client.c
+++ b/web/server/web_client.c
@@ -1016,7 +1016,7 @@ static inline HTTP_VALIDATION http_request_validate(struct web_client *w) {
     encoded_url = s;
 
     //we search for the position where we have " HTTP/", because it finishes the user request
-    s = url_find_protocol(s);
+        s = url_find_protocol(s);
 
     // incomplete requests
     if(unlikely(!*s)) {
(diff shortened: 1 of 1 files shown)


amoss:
This code is unclear and complex. I have some questions about it. After reading the url parsing code it seems that the original URL is in `w->last_url` and it was stored as the first match in `GET .* HTTP/.*', i.e. it is the original URL as presented to the server after the %-decoding and unicode-expansion. All that needs to be done is change the `http://` to `https://`. 

This would be simple to do like this:
`buffer_sprintf(w->response.header_output, "https%s", w->last_url+4);`

* Why are you doing anything more complex than this?
* Part of the part removes the javascript redirect that is there as a fallback - why is this removed?

web/server/web_client.c
```diff
@@ -1581,9 +1596,6 @@ void web_client_process_request(struct web_client *w) {
 #ifdef ENABLE_HTTPS
         case HTTP_VALIDATION_REDIRECT:
         {
-            buffer_flush(w->response.data);
-            w->response.data->contenttype = CT_TEXT_HTML;
-            buffer_strcat(w->response.data, "<!DOCTYPE html><!-- SPDX-License-Identifier: GPL-3.0-or-later --><html><body onload=\"window.location.href ='https://'+ window.location.hostname + ':' + window.location.port +  window.location.pathname\">Redirecting to safety connection, case your browser does not support redirection, please click <a onclick=\"window.location.href ='https://'+ window.location.hostname + ':' + window.location.port +  window.location.pathname\">here</a>.</body></html>");
```
vlvkobal, line 1586: Why did you delete this part?

web/server/web_client.c
```diff
@@ -1016,7 +1016,7 @@ static inline HTTP_VALIDATION http_request_validate(struct web_client *w) {
     encoded_url = s;
 
     //we search for the position where we have " HTTP/", because it finishes the user request
-    s = url_find_protocol(s);
+        s = url_find_protocol(s);
```
vlvkobal, line 1019: You don't need an indentation here.

web/server/web_client.c
thiagoftsm: When the client receives the `301` as status and `Location: http://..` It redirects for the correct chart, as you can see in this [link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301), so I decided to clean up the response,

web/server/web_client.c
thiagoftsm: Thank you to call attention for this, I removed it now.

thiagoftsm:
> This code is unclear and complex. I have some questions about it. After reading the url parsing code it seems that the original URL is in `w->last_url` and it was stored as the first match in `GET .* HTTP/.*`, i.e. it is the original URL as presented to the server after the %-decoding and unicode-expansion. All that needs to be done is change the `http://` to `https://`.

You are right, we read from the request.

> This would be simple to do like this:
> `buffer_sprintf(w->response.header_output, "https%s", w->last_url+4);`
> 
>     * Why should we do anything more complex than this?
> 
>     * Part of the patch removes the javascript redirect that is there as a fallback - why is this removed?

Unfortunately, it cannot be only this, because the variable `last_url` that is filled in this line

https://github.com/netdata/netdata/blob/83e5cbe806bd81de526656df65ceb86bce17e392/web/server/web_client.c#L1081

does not have the protocol, the host and the port according the RFC:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html

when Netdata is parsing the request, we are setting `NULL` value, in the exact position that there is a separator between  the path and the query string, this is the motive we are also bringing this additional complexity for this case, I think we could reduce the complexity if we do not set the separator as NULL, but it cannot be only one line, @amoss do you agree with these changes suggested in this paragraph? 

Best regards!

amoss:
In the code that you've written the host field is copied in and then the `last_url` immediately after, why not like this?
```
buffer_sprintf(w->response.header_output, "https://%s%s", w->server_host, w->last_url);
```

thiagoftsm:
> ```
> buffer_sprintf(w->response.header_output, "https://%s%s", w->server_host, w->last_url);
> ```

Hi @amoss ,

Unfortunately we cannot do only  this, but I changed the code to have something like this, and the motives I am doing this are described in the next list:

1 - I changed the code for Netdata delivers the exact same request that was made, previously we were sending it decoded.
2 - We cannot do only `https://%s%s` inside the header output, because this would not following the [RFC specifications](https://tools.ietf.org/html/rfc7231#page-56) and browsers could ignore the request. I also considered these specifications to changed the response header, now Netdata sends what is really necessary.

Best regards!

web/server/web_client.c
thiagoftsm: Hey @vlvkobal , please, also take a look at [my latest answer](https://github.com/netdata/netdata/pull/9790#issuecomment-677964061), I gave more details there.

amoss:
> 1 - I changed the code for Netdata delivers the exact same request that was made, previously we were sending it decoded.

Yes, this is a good change.

> 2 - We cannot do only `https://%s%s` inside the header output, because this would not following the [RFC specifications]

Yes :) I expect you to put in the rest the header of the header around the piece that we are talking about. I don't expect you to just send `https:...` in the response :)

Good. The code is looking better, now let's check the last thing. The javascript redirect in the body is there to improve browser compatibility in case the `301` is not processed correctly. Is there a good reason to remove it?

vlvkobal:
When I try to access `http://arch-esxi:19999/api/v1/info` it redirects to `https://arch-esxi:19999/api/v1/infoCache-Control:%20no-cache,%20no-store,%20must-revalidate`, which is not correct.

amoss:
L1167 -> missing `\r\n`

amoss (changes requested):
The missing newline sequence that @vlvkobal spotted.

thiagoftsm:
Thank you @vlvkobal and @amoss to point the problem, I fixed it.

vlvkobal (approved):

amoss (approved):
Retested with variations on `curl -D - 'localhost:19999/foo?k=v;frag#anchor' | hexdump -C`

Issue #9626, opened 2020-07-29
Installer and updater should sanely handle `/tmp` being mounted with `noexec`.


Ferroin:
Currently we require a usable `/tmp` that is not mounted with `noexec` for our install, which causes otherwise sanely configured system to not work correctly.

Based on discussion between myself and @prologic, the plan is to update our installation and update tooling to fall back to `$PWD` if `/tmp` is mounted with `noexec` so that such systems work properly (or at least so that fewer of them fail).

prologic:
Thanks @Ferroin

PR #9797, opened 2020-08-20
Improved temporary directory checking in installer and updater.

Ferroin:
##### Summary

This updates the checks that our installer and updater make to choose what to use for a temporary directory to be both more robust and more portable.

The original logic directly checked the mount options for `/tmp`, which is unfortunately unreliable as it doesn't work at all on FreeBSD (no procfs), probably doesn't work on macOS (different paths), and fails in some configurations on Linux (rare cases where `/tmp` is not actually a mount point, such as systems where `/` is inherently on tmpfs). The new check instead explicitly writes a temporary file at the location under test, marks it executable, and tries to run it (similar to how Autoconf generated configure scripts handle this type of thing), which will always tell us if that location is usable.

The new checks use the following logic for deciding where to place temporary files:

* For each directory to be checked, verify that it is both writable by the current user, and that the current user can execute files they write there.
* If `$TMPDIR` is set, preferentially use that.
* If that fials, try `/tmp`.
* If that also fails, fall back to `$PWD`.
* If all checks fail, bail early with an explanation instead of failing when we first try to do things with the directory.

It also adds the same checks to the `netdata-installer.sh` script, which was previously completely missing them.

##### Component Name

area/packaging

##### Test Plan

Tested using local installs in Docker containers with various temporary directory setups.

##### Additional Information

Fixes: #9626

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -78,8 +78,6 @@ if [ -z "${TMPDIR}" ] || _cannot_use_tmpdir "${TMPDIR}" ; then
   fi
 fi
 
-echo >&2 "Using ${TMPDIR} as a temporary directory."
-
 # -----------------------------------------------------------------------------
 # set up handling for deferred error messages
 NETDATA_DEFERRED_ERRORS=""
@@ -362,6 +360,10 @@ netdata_banner "real-time performance monitoring, done right!"
 cat << BANNER1
 
   You are about to build and install netdata to your system.
+
+  The build process will use ${TPUT_CYAN}${TMPDIR}${TPUT_RESET} for
+  any temporary files. You can override this by setting \$TMPDIR to a
+  writable directory where you can execute files.
 
   It will be installed at these locations:
 
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -42,6 +42,45 @@ fi
 cd "${NETDATA_SOURCE_DIR}" || exit 1
 
 # -----------------------------------------------------------------------------
+# figure out an appropriate temporary directory
+_cannot_use_tmpdir() {
+  local testfile ret
+  testfile="$(TMPDIR="${1}" mktemp -q -t netdata-test.XXXXXXXXXX)"
+  ret=0
+
+  if [ -z "${testfile}" ] ; then
+    return "${ret}"
+  fi
+
+  if /bin/echo -e "#!/bin/sh\necho SUCCESS\n" > "${testfile}" ; then
+    if chmod +x "${testfile}" ; then
+      if [ "$("${testfile}")" = "SUCCESS" ] ; then
+        ret=1
+      fi
+    fi
+  fi
+
+  rm -f "${testfile}"
+  return "${ret}"
+}
+
+if [ -z "${TMPDIR}" ] || _cannot_use_tmpdir "${TMPDIR}" ; then
+  if _cannot_use_tmpdir /tmp ; then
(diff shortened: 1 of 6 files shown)


prologic (approved):

Ferroin:
@prologic Assuming this is still good after the rebase, feel free to merge it when you reapprove it.

prologic (approved):

netdata-installer.sh
```diff
@@ -41,6 +41,45 @@ fi
...
+_cannot_use_tmpdir() {
+  local testfile ret
+  testfile="$(TMPDIR="${1}" mktemp -q -t netdata-test.XXXXXXXXXX)"
+  ret=0
+
+  if [ -z "${testfile}" ] ; then
+    return "${ret}"
+  fi
+
+  if /bin/echo -e "#!/bin/sh\necho SUCCESS\n" > "${testfile}" ; then
```
amoss, line 55: This doesn't work if it is executed in bash as it tried to evaluate the `"!"` as a history event. It will work if you use `'` instead to make it literal.

packaging/installer/kickstart.sh
```diff
@@ -152,15 +153,43 @@ warning() {
...
+_cannot_use_tmpdir() {
+  local testfile ret
+  testfile="$(TMPDIR="${1}" mktemp -q -t netdata-test.XXXXXXXXXX)"
+  ret=0
+
+  if [ -z "${testfile}" ] ; then
+    return "${ret}"
+  fi
+
+  if /bin/echo -e "#!/bin/sh\necho SUCCESS\n" > "${testfile}" ; then
```
amoss, line 165: Same

amoss:
Why does kickstart use different case to the installer `TMPDIR` vs `tmpdir` ?

Ferroin:
> Why does kickstart use different case to the installer `TMPDIR` vs `tmpdir` ?

`TMPDIR` is a well known variable name used as a de-facto standard for a lot of tools. We _really_ should not be overwriting it for multiple reasons (not the least of which being that doing so has the potential to generate multiple layers of temporary directories). Because variable names are case sensitive, I just case-swapped our usage of `TMPDIR` for other things to preserve the name, though the name should indeed probably just be changed.

amoss:
I think that the same arguments apply to the installer as to the kickstart script - so if we should avoid setting `TMPDIR` in one then I assume that we need to avoid it in the other. Are we expecting users to `source` one of the scripts but execute the other, or is there something more complex here?

Ferroin:
> I think that the same arguments apply to the installer as to the kickstart script - so if we should avoid setting `TMPDIR` in one then I assume that we need to avoid it in the other. Are we expecting users to `source` one of the scripts but execute the other, or is there something more complex here?

There are two cases here to worry about. The first is the logic that this PR is introducing that ensures `$TMPDIR` is set to a sanevalue and then used by all invocations of `mktemp`. This _needs_ to manipulate `$TMPDIR` because that's the only way to get `mktemp` to use it (well, only portable way, because BSD/macOS `mktemp` does not support specifying a prefix on the CLI).

However, we also need to keep track of temporary paths that we produce at runtime in a couple of places in the kickstart and updater scripts, and those need to _not_ overwrite `$TMPDIR`, so we need to use a distinct variable name for those.

amoss:
Yes, that makes sense. Thanks for the explanation.

amoss (approved):

prologic (approved):

PR #9804, opened 2020-08-21
Stream Claim Id

underhood:
##### Summary
claimed_id of nodes is streamed to parents. For easy testing also added to `/api/v1/info` (where cloud will need it anyway sooner or later)
Example of 3-level streaming from the view of top-level parent:
Screenshot from 2020-08-25 11-38-41

##### Component Name
Streaming
##### Test Plan
Have some streaming setup and look at `/api/v1/info`. Check nothing breaks with Cloud by connecting this node to `staging.netdata.cloud`. Claim nodes and see if their claiming Id propagates up the network by checking `/api/v1/info` of the topmost node. All nodes have to be running this PR for it to work.

What was tested together with the cloud team:
1. Claim node to Staging and connect it. Node had 2 children.
1. Message as per screenshot above (with new `mirrored_hosts_status` field) is send to cloud as part of `onconnect` payload over ACLK
1. Checked Cloud GUI still works as expected with this new fields
1. Cloud team checked there are no errors in the background (not visible to us from GUI)

What was tested additionally:
- node that is claimed during runtime and then `netdatacli reload-claiming-state`
- up to 4 level hierarchy of streaming
    - `TopLevelMaster` (VM,not claimed)
    - `timowss` (real machine,claimed) -> `TopLevelMaster` (VM)
    - `Fed32-Layer1Host1` (VM,claimed) -> `timowss` (real machine, claimed)
    - `Fed32-Layer1Host2` (VM,not claimed) -> `timowss` (real machine, claimed)
    - `Fed32-Layer2Host1` (VM,not claimed) -> `Fed21-Layer1Host1` (VM, claimed)
    - this represents final test with biggest amount of hosts tested, during development and testing different combination of nodes and claiming states were used
- both claimed and unclaimed nodes have correct status reported

##### Additional Information
Codacy false triggers here. It doesn't seem to recognize `rrdhost_foreach_read` macro is a loop and as a result, thinks `if(count > 0) {` is always false (which is not true).

### DIFF
--- a/claim/claim.c
+++ b/claim/claim.c
@@ -35,7 +35,7 @@ char *is_agent_claimed()
 {
     char *result;
     netdata_mutex_lock(&localhost->claimed_id_lock);
-    result = (localhost->claimed_id == NULL) ? NULL : strdup(localhost->claimed_id);
+    result = (localhost->claimed_id == NULL) ? NULL : strdupz(localhost->claimed_id);
     netdata_mutex_unlock(&localhost->claimed_id_lock);
     return result;
 }
--- a/claim/claim.c
+++ b/claim/claim.c
@@ -154,6 +154,8 @@ void load_claiming_state(void)
 
     long bytes_read;
     claimed_id = read_by_filename(filename, &bytes_read);
+    //TODO: remove claimed_id global variable and cleanup
+    rrdhost_set_claimed_id(localhost, claimed_id);
     netdata_mutex_unlock(&claim_mutex);   // Only the main thread can call this function, safe to release and then read
     if (!claimed_id) {
         info("Unable to load '%s', setting state to AGENT_UNCLAIMED", filename);
@@ -162,6 +164,7 @@ void load_claiming_state(void)
 
     info("File '%s' was found. Setting state to AGENT_CLAIMED.", filename);
     netdata_cloud_setting = appconfig_get_boolean(&cloud_config, CONFIG_SECTION_GLOBAL, "enabled", 1);
+
 #endif
 }
 
--- a/claim/claim.c
+++ b/claim/claim.c
@@ -154,7 +154,7 @@ void load_claiming_state(void)
     long bytes_read;
     char *claimed_id = read_by_filename(filename, &bytes_read);
     if(claimed_id && uuid_parse(claimed_id, uuid)) {
-        error("claimed_id doesn't look like valid UUID");
+        error("claimed_id \"%s\" doesn't look like valid UUID", claimed_id);
         freez(claimed_id);
         claimed_id = NULL;
     }
--- a/claim/claim.c
+++ b/claim/claim.c
@@ -133,6 +133,7 @@ void load_claiming_state(void)
 #if defined( DISABLE_CLOUD ) || !defined( ENABLE_ACLK )
     netdata_cloud_setting = 0;
 #else
+    uuid_t uuid;
     netdata_mutex_lock(&localhost->claimed_id_lock);
     if (localhost->claimed_id) {
         freez(localhost->claimed_id);
@@ -152,6 +153,11 @@ void load_claiming_state(void)
 
     long bytes_read;
     char *claimed_id = read_by_filename(filename, &bytes_read);
+    if(claimed_id && uuid_parse(claimed_id, uuid)) {
+        error("claimed_id doesn't look like valid UUID");
+        freez(claimed_id);
+        claimed_id = NULL;
+    }
     localhost->claimed_id = claimed_id;
     netdata_mutex_unlock(&localhost->claimed_id_lock);
     if (!claimed_id) {
--- a/claim/claim.c
+++ b/claim/claim.c
@@ -27,8 +27,6 @@ static char *claiming_errors[] = {
         "Service Unavailable",                          // 17
         "Agent Unique Id Not Readable"                  // 18
 };
-static netdata_mutex_t claim_mutex = NETDATA_MUTEX_INITIALIZER;
-static char *claimed_id = NULL;
 
 /* Retrieve the claim id for the agent.
  * Caller owns the string.
@@ -36,9 +34,9 @@ static char *claimed_id = NULL;
 char *is_agent_claimed()
 {
     char *result;
-    netdata_mutex_lock(&claim_mutex);
-    result = (claimed_id == NULL) ? NULL : strdup(claimed_id);
-    netdata_mutex_unlock(&claim_mutex);
+    netdata_mutex_lock(&localhost->claimed_id_lock);
+    result = (localhost->claimed_id == NULL) ? NULL : strdup(localhost->claimed_id);
+    netdata_mutex_unlock(&localhost->claimed_id_lock);
     return result;
 }
 
@@ -135,10 +133,10 @@ void load_claiming_state(void)
 #if defined( DISABLE_CLOUD ) || !defined( ENABLE_ACLK )
     netdata_cloud_setting = 0;
 #else
-    netdata_mutex_lock(&claim_mutex);
-    if (claimed_id != NULL) {
-        freez(claimed_id);
-        claimed_id = NULL;
+    netdata_mutex_lock(&localhost->claimed_id_lock);
+    if (localhost->claimed_id) {
+        freez(localhost->claimed_id);
+        localhost->claimed_id = NULL;
     }
     if (aclk_connected)
     {
@@ -153,10 +151,9 @@ void load_claiming_state(void)
     snprintfz(filename, FILENAME_MAX, "%s/cloud.d/claimed_id", netdata_configured_varlib_dir);
 
     long bytes_read;
-    claimed_id = read_by_filename(filename, &bytes_read);
-    //TODO: remove claimed_id global variable and cleanup
-    rrdhost_set_claimed_id(localhost, claimed_id);
-    netdata_mutex_unlock(&claim_mutex);   // Only the main thread can call this function, safe to release and then read
+    char *claimed_id = read_by_filename(filename, &bytes_read);
+    localhost->claimed_id = claimed_id;
+    netdata_mutex_unlock(&localhost->claimed_id_lock);
     if (!claimed_id) {
         info("Unable to load '%s', setting state to AGENT_UNCLAIMED", filename);
         return;
@@ -164,7 +161,6 @@ void load_claiming_state(void)
 
     info("File '%s' was found. Setting state to AGENT_CLAIMED.", filename);
     netdata_cloud_setting = appconfig_get_boolean(&cloud_config, CONFIG_SECTION_GLOBAL, "enabled", 1);
-
 #endif
 }
 
--- a/streaming/receiver.c
+++ b/streaming/receiver.c
@@ -116,7 +116,7 @@ PARSER_RC streaming_claimed_id(char **words, void *user, PLUGINSD_ACTION *plugin
         error("The host GUID received doesn't have format of valid UUID!");
         return PARSER_RC_ERROR;
     }
-    if(uuid_parse(words[2], uuid)) {
+    if(uuid_parse(words[2], uuid) && strcmp(words[2], "NULL")) {
         error("The Claim ID received doesn't have format of valid UUID!");
         return PARSER_RC_ERROR;
     }
--- a/streaming/receiver.c
+++ b/streaming/receiver.c
@@ -106,24 +106,24 @@ PARSER_RC streaming_claimed_id(char **words, void *user, PLUGINSD_ACTION *plugin
 
     for (i = 0; words[i]; i++) ;
     if (i != CLAIMED_ID_MIN_WORDS) {
-        error("Command CLAIMED_ID came malformed %d parameters are compulsory", CLAIMED_ID_MIN_WORDS - 1);
+        error("Command CLAIMED_ID came malformed %d parameters are expected but %d received", CLAIMED_ID_MIN_WORDS - 1, i - 1);
         return PARSER_RC_ERROR;
     }
 
     // We don't need to parsed UUID
     // just do it to check the format
     if(uuid_parse(words[1], uuid)) {
-        error("The host GUID received doesn't have format of valid UUID!");
+        error("1st parameter (host GUID) to CLAIMED_ID command is not valid GUID. Received: \"%s\".", words[1]);
         return PARSER_RC_ERROR;
     }
     if(uuid_parse(words[2], uuid) && strcmp(words[2], "NULL")) {
-        error("The Claim ID received doesn't have format of valid UUID!");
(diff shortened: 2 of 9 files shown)


web/api/web_api_v1.c
```diff
@@ -775,13 +775,44 @@ static inline void web_client_api_request_v1_info_summary_alarm_statuses(RRDHOST
 
 static inline void web_client_api_request_v1_info_mirrored_hosts(BUFFER *wb) {
     RRDHOST *rc;
+    BUFFER *guids = buffer_create(2048);
```
amoss, line 778: This buffer is allocated and freed only so that there is a single loop over the hosts. But looping over the hosts is cheap - it is only walking down a linked list. Allocating and freeing the buffer is expensive as it involves syscalls. Why not two loops that both `buffer_sprintf(wb` instead of two buffers?

streaming/receiver.c
```diff
@@ -95,6 +95,54 @@ PARSER_RC streaming_timestamp(char **words, void *user, PLUGINSD_ACTION *plugins
...
+    netdata_mutex_lock(&host->claimed_id_lock);
+    if (host->claimed_id)
+        freez(host->claimed_id);
+    host->claimed_id = strcmp(words[2], "NULL") ? strdupz(words[2]) : NULL;
+    netdata_mutex_unlock(&host->claimed_id_lock);
+
+    rrdpush_claimed_id(host);
+
+    return PARSER_RC_OK;
+}
```
amoss, line 144: The error messages in this function are not useful. If this breaks then we need to diagnose it by checking logs, so at the point that an error is encountered there needs to be enough information in the log to pinpoint the problem.
* If the claim ID string is not a valid form then show the string.
* If the GUID is not valid then show the string.
* If the child is attempting to change the wrong UUID then show the incorrect UUIDS.
This does not jut help us - users with broken config will look in their logs to try to understand the problem and it should contain enough information.

The conversation above has been marked resolved - but the problem is not resolved. A child cannot change the claim id of an arbitrary host. This means that configuration errors propagate through the system and break the state on other nodes. Find another way to handle it,

streaming/sender.c
```diff
@@ -622,6 +622,15 @@ void *rrdpush_sender_thread(void *ptr) {
...
             }
+            if (s->version >= STREAM_VERSION_CLAIM) {
+                rrd_rdlock();
+                RRDHOST *h;
+                rrdhost_foreach_read(h) {
+                    if(h->claimed_id)
+                        rrdpush_claimed_id(h);
+                }
+                rrd_unlock();
+            }
```
amoss, line 633: When an agent is streaming data from multiple hosts each one has a separate connection and sender.

amoss (changes requested):
1. Improve the logging.
2. Fix the loop with the unnecessary buffer
3. Find a better solution for updating claim ids on other hosts that does not permit the system to be broken.
4. Either explicitly confirm with the cloud team that the changes do not break anything or put in the version negotiation first and change the protocol version.

web/api/web_api_v1.c
underhood: OK I will change it. Yep it was done to go trough the list only once while lock is held. There is therefore small improvement during lock but as you correctly pointed out at cost outside of the critical section. This would be improved also by mempooling the buffers but that is another topic

streaming/receiver.c
underhood: I will improve the logging. 
Host must be able to modify claim_id of another host than self (as it can itself have children). Not sure yet how to improve this other than adding visibility of structure of children below (as a tree instead of list)

ktsaou:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=6051564&bid=19824351)

ktsaou:


```diff

Issues
======
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=6051564&bid=19824351)

streaming/receiver.c
amoss: The terminology is a bit confusing here (historically I think the terms have shifted a bit). When a sender connects to a receiver it only transmits a single *host* - in this context the host is a particular database. When a proxy node is transmitting the data for multiple children to a parent there one stream per host. So the protocol that is coming over the wire is only describing a single host (even though the node may be streaming multiple hosts over multiple connections).

ktsaou:


```diff

Issues
======
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=6051564&bid=19824351)

underhood:
@amoss in regards to point 4 it is confirmed with cloud team now this can go in. the `hello` version negotiation message (this will be another PR) will need separate explicit GO from cloud team

amoss:
The test plan is insufficient. This needs to be tested it by connecting it to staging and checking with the cloud team that the data received is in a format that they can process without errors.

ktsaou:


```diff

Issues
======
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=6051564&bid=19824351)

streaming/sender.c
underhood: you are right done

ktsaou:


```diff

Issues
======
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=6051564&bid=19824351)

streaming/receiver.c
underhood: ah I see what you mean now. Knew there is connection per host but somehow misunderstood/didn't make a connection with what you try to say.

Changed it. Thanks

underhood:
@amoss all your points are resolved now.
1. done
2. I changed it to your wishes: so it makes critical section a bit longer (loop over hosts twice instead of once) but it removes the expense of Netdata BUFFER allocation
3. It would possibly break system only if child had a bug but it is good suggestion and I added additional check on what child sends.
4. clarified with cloud team and have explicit approval for `/api/v1/info` payload update that is done in this PR

amoss:
The test plan does not describe the testing in enough detail for a reviewer to decide if this is properly tested. Please fill it in properly:
1. What was the test scenario
2. What was sent to the cloud
3. What was checked on the far end

underhood:
made test plan more verbose

amoss:
What was the 4-level setup? Which nodes were the parents of which nodes? Which nodes were claimed and which nodes were not claimed?

underhood:
updated info on 4 layer stream test performed

web/api/web_api_v1.c
```diff
@@ -776,14 +776,57 @@ static inline void web_client_api_request_v1_info_summary_alarm_statuses(RRDHOST
...
     }
+
+    buffer_strcat(wb, "\n\t],\n\t\"mirrored_hosts_status\": [\n");
+    count = 0;
+    rrdhost_foreach_read(rc) {
+        if (rrdhost_flag_check(rc, RRDHOST_FLAG_ARCHIVED))
+            continue;
+        if (count > 0)
+            buffer_strcat(wb, ",\n");
+
```
mfundul, line 799: I don't think we should nest those locks, they are completely unrelated and I don't like the implications. We could just copy the strings somewhere or maybe use `buffer_strcat()`?

mfundul (changes requested):

underhood:
@amoss 
ZO_12i8m38RlVOgm-mwjDr7OIJpCcoyGjuYJr4gJAO8FhyLDN57myfU4VlnVAb6CsdisONUaHc4RpZLTY3SeIj5CHvX3mzXvQc49UsIX4IjMrCvSMUEDebq7KLi02h0zM9DYGchitw1sFbn5pF3c_tuNkB70cHJ_gRaFDVTKI_65pCzUtCioDtqjqstPPZ2omvXgAZgrlUS7

this is topology of the 4 layer test However during testing I tried different combinations and topologies

claim/claim.c
```diff
@@ -27,18 +27,16 @@ static char *claiming_errors[] = {
...
  * Caller owns the string.
 */
 char *is_agent_claimed()
 {
     char *result;
-    netdata_mutex_lock(&claim_mutex);
-    result = (claimed_id == NULL) ? NULL : strdup(claimed_id);
-    netdata_mutex_unlock(&claim_mutex);
+    netdata_mutex_lock(&localhost->claimed_id_lock);
+    result = (localhost->claimed_id == NULL) ? NULL : strdup(localhost->claimed_id);
```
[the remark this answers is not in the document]
underhood: thx, comes form pre-existing code but you are right and I will fix it

streaming/receiver.c
```diff
@@ -95,6 +95,53 @@ PARSER_RC streaming_timestamp(char **words, void *user, PLUGINSD_ACTION *plugins
...
+    uuid_t uuid;
+    RRDHOST *host = ((PARSER_USER_OBJECT *)user)->host;
+
+    for (i = 0; words[i]; i++) ;
+    if (i != CLAIMED_ID_MIN_WORDS) {
+        error("Command CLAIMED_ID came malformed %d parameters are expected but %d received", CLAIMED_ID_MIN_WORDS - 1, i - 1);
+        return PARSER_RC_ERROR;
+    }
+
+    // We don't need to parsed UUID
```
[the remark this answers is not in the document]
underhood: thanks

ktsaou:


```diff

Issues
======
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/netdata/netdata/pullRequest?prid=6051564&bid=19824351)

amoss:
Ok good, now the test plan is comprehensive. I'll let the other reviewer take a look before I go back for a second look at the code.

amoss:
I've disabled the Codacy warnings as they are just noise - it does not understand the loop macro for some reason.

web/api/web_api_v1.c
underhood: avoided holding both locks at expense of strdupz

web/api/web_api_v1.c
```diff
@@ -776,14 +776,57 @@ static inline void web_client_api_request_v1_info_summary_alarm_statuses(RRDHOST
...
+        netdata_mutex_unlock(&rc->claimed_id_lock);
+
+        count++;
+    }
+    rrd_unlock();
+
+    buffer_strcat(wb, "\n\t],\n");
+}
+
+static inline void web_client_api_request_v1_info_mirrored_hosts_uids(BUFFER *wb) {
```
mfundul, line 821: Is this called somewhere? Cause I can't find it.

underhood: aaah artifact from cherry-picked commit from another branch
removed, thanks

streaming/rrdpush.c
```diff
@@ -364,6 +364,21 @@ void rrdpush_send_labels(RRDHOST *host) {
...
+    if(unlikely(!host->rrdpush_send_enabled || !host->rrdpush_sender_connected))
+        return;
+
+    sender_start(host->sender);
+    netdata_mutex_lock(&host->claimed_id_lock);
+
+    buffer_sprintf(host->sender->build, "CLAIMED_ID %s %s\n", host->machine_guid, (host->claimed_id ? host->claimed_id : "NULL") );
+
+    netdata_mutex_unlock(&host->claimed_id_lock);
+    sender_commit(host->sender);
```
amoss, line 379: After writing into the sender buffer the sender thread should be signalled over the pipe so that it knows there is data waiting.

amoss: There is an example on https://github.com/netdata/netdata/blob/9ecb4756d7104a2dd11e74ecec31f413035c69bb/streaming/rrdpush.c#L362

streaming/receiver.c
```diff
@@ -95,6 +95,53 @@ PARSER_RC streaming_timestamp(char **words, void *user, PLUGINSD_ACTION *plugins
...
+        error("2nd parameter (Claim ID) to CLAIMED_ID command is not valid GUID. Received: \"%s\".", words[2]);
+        return PARSER_RC_ERROR;
+    }
+
+    if(strcmp(words[1], host->machine_guid)) {
+        error("Claim ID is for host \"%s\" but it came over connection for \"%s\"", words[1], host->machine_guid);
+        return PARSER_RC_OK; //the message is OK problem must be somewehere else
+    }
+
+    if (host == localhost) {
```
amoss, line 129: This check is redundant now, as it is captured by the check above.

web/api/web_api_v1.c
```diff
@@ -776,16 +776,46 @@ static inline void web_client_api_request_v1_info_summary_alarm_statuses(RRDHOST
...
         buffer_sprintf(wb, "\t\t\"%s\"", rc->hostname);
         count++;
     }
-    buffer_strcat(wb, "\n");
+
+    buffer_strcat(wb, "\n\t],\n\t\"mirrored_hosts_status\": [\n");
+    count = 0;
+    rrdhost_foreach_read(rc)
+    {
+        if (rrdhost_flag_check(rc, RRDHOST_FLAG_ARCHIVED))
```
amoss, line 796: Can you check with product if this is correct? Are they expecting child node databases to be available after a restart on the parent node?

web/api/web_api_v1.c
```diff
@@ -776,16 +776,46 @@ static inline void web_client_api_request_v1_info_summary_alarm_statuses(RRDHOST
...
+            buffer_strcat(wb, ",\n");
+
         buffer_sprintf(wb, "\t\t\"%s\"", rc->hostname);
         count++;
     }
-    buffer_strcat(wb, "\n");
+
+    buffer_strcat(wb, "\n\t],\n\t\"mirrored_hosts_status\": [\n");
+    count = 0;
+    rrdhost_foreach_read(rc)
```
amoss, line 794: This should not be called `rc` it should be called `host`. The part that you copied it from is wrong, the contributor was working on alarms so they reused the iterator name for alarm entries.

web/api/web_api_v1.c
underhood: This is unchanged in current PR archived hosts were skipped before as well.
In case we need archived hosts in info payload we will need to update also the metadatalog to keep and comapct the claim_id. That however would not be 1hr task

mfundul: It was specifically asked to not expose archived host by Chris.

web/api/web_api_v1.c
underhood: will change the variable name for clarity

streaming/rrdpush.c
underhood: added

underhood:
@Ferroin arch CI build started failing yesterday with ` configure: error: libuv required but not found. Try installing 'libuv1-dev' or 'libuv-devel'.` Can you take a look if you have 5'

mfundul (approved):

amoss (approved):

underhood:
I checked also with production that it doesnt kill something just to be paranoid before merge.

PR #9811, opened 2020-08-24
Remove dots in front of "oid" field in examples of SNMP collector

OdysLam:
According to the docs: 
```
Keep in mind that  `snmpwalk` outputs the OIDs with a dot in front them. You should remove this dot when adding OIDs to the configuration file of this collector.
```
The examples that we provide after this message have a dot in front of the OID. This commit removes that dot.

Reporter by user in Zendesk, ticket #328

### DIFF
--- a/collectors/node.d.plugin/snmp/README.md
+++ b/collectors/node.d.plugin/snmp/README.md
@@ -297,14 +297,14 @@ This switch has a very slow SNMP processors. To respond, it needs about 8 second
                     "family": "power",
                     "dimensions": {
                         "supply": {
-                            "oid": ".1.3.6.1.2.1.105.1.3.1.1.2.1",
+                            "oid": "1.3.6.1.2.1.105.1.3.1.1.2.1",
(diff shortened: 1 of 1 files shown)


ilyam8:
I dont see a leading dot is a problem, leading dot means it is an absolute OID and not relative (some tools could add some prefix to relative OIDs).

Our snmp collector doesnt add any prefixes, leading dot does nothing.

ilyam8:
> You should remove this dot when adding OIDs to the configuration file of this collector.

I think this statement is wrong and we should remove it.

---

```cmd
[ilyam@ilyam-pc node.d]$ cat snmp.conf | grep oid
                            "oid":".1.3.6.1.2.1.31.1.1.1.6.1",
                            "oid":".1.3.6.1.2.1.31.1.1.1.10.1",
```

works ok for me

OdysLam:
οκ @ilyam8. I will do a new PR with that change. Thanks.

PR #9813, opened 2020-08-24
Fix broken link and clean up frontmatter in health docs

joelhans:
##### Summary

The brackets around the GitHub edit URL in the reference doc are throwing errors over on Learn. This PR removes the brackets. I also removed some unnecessary frontmatter code from adjacent docs as well.

##### Component Name

area/docs
area/health

### DIFF
--- a/health/QUICKSTART.md
+++ b/health/QUICKSTART.md
@@ -1,9 +1,6 @@
 <!--
----
 title: "Health quickstart"
-date: 2020-04-27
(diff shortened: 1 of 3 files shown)


thiagoftsm (approved):
I observed that the PR is removing comments and bringing an update for an URL that will have the same pattern of the other files, so LGTM and thank you @joelhans !

ilyam8 (approved):

Issue #9035, opened 2020-05-14
Alarms dialog does not load correctly


djmattyg007:
##### Bug report summary
Loading the Alarms dialog by clicking on the Alarms button in the toolbar does not work correctly. I get four error messages, one after the other, informing me that various JS libraries failed to load.

Specifically, the following libraries fail to load:

- clipboard-polyfill-be05dad.js
- bootstrap-table-1.11.0.min.js
- bootstrap-table-export-1.11.0.min.js
- tableExport-1.6.0.min.js

I believe this is something to do with how I'm accessing Netdata. I have Nginx set up to reverse proxy to Netdata, following the instructions here, in the "As a subfolder for multiple Netdata servers, via one Nginx" section:

https://learn.netdata.cloud/docs/agent/running-behind-nginx/#as-a-subfolder-for-multiple-netdata-servers-via-one-nginx

For some reason, for the four libraries listed above, instead of trying to load URLs that look like this:

https://example.com/netdata/hostname/lib/library.js

The URLs look like this:

https://example.com/netdata/lib/library.js

To confirm, I do have the snippet in place that ensures there is a trailing slash, and that does appear to be working correctly.

##### OS / Environment
```
Linux shacklebolt 5.6.11-arch1-1 #1 SMP PREEMPT Wed, 06 May 2020 17:32:37 +0000 x86_64 GNU/Linux
/etc/os-release:NAME="Arch Linux"
/etc/os-release:PRETTY_NAME="Arch Linux"
/etc/os-release:ID=arch
/etc/os-release:BUILD_ID=rolling
/etc/os-release:ANSI_COLOR="38;2;23;147;209"
/etc/os-release:HOME_URL="https://www.archlinux.org/"
/etc/os-release:DOCUMENTATION_URL="https://wiki.archlinux.org/"
/etc/os-release:SUPPORT_URL="https://bbs.archlinux.org/"
/etc/os-release:BUG_REPORT_URL="https://bugs.archlinux.org/"
/etc/os-release:LOGO=archlinux

```

##### Netdata version
netdata v1.22.1
 

##### Component Name
Alarms dialog in dashboard

##### Steps To Reproduce
Click on the alarms button in the toolbar

##### Expected behavior
There should be no errors when loading the dialog.

sebastianopilla:
I have the same setup (netdata behind nginx, in a subdirectory) and I can confirm that I have the same problem with 1.22.1. The requests for the 4 above mentioned js files return a 404 error to the browser.

[...]

PR #9814, opened 2020-08-25
fix old dashboard 3rd party packaging

jacekkolasa:
##### Summary
relates to https://github.com/netdata/netdata/issues/9035

We host the old dashboard under /old suffix. To make it work, we had to change few paths in old js code. Netdata packaged by third party is sometimes using the old dashboard as standard, but our before mentioned changes have broken few functionalities, like opening the alarms modal.

I've added a detection to check if we are using /old suffix, or not, so 3rd party packages are not partially broken.

##### Component Name
web/dashboard

##### Test Plan

1. Install Netdata using external package, for example `yum install netdata` in CentOS
2. Configure Nginx so that Dashboard is hosted under subpath (like http://localhost/netdata), example configuration: https://learn.netdata.cloud/docs/agent/running-behind-nginx#as-a-subfolder-to-an-existing-virtual-host
3. Go to dashboard, open alarms modal. There should be no alert error.

### DIFF
--- a/web/gui/main.js
+++ b/web/gui/main.js
@@ -500,11 +500,25 @@ function toggleAgentItem(e, guid) {
 // has multiple host databases. It's own, and multiple mirrored. Mirrored databases
 // can be accessed with <http://localhost:19999/host/NAME/>
 const OLD_DASHBOARD_SUFFIX = "old"
+let isOldSuffix = true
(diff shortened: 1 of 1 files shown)


VLegakis (approved):
LGTM

PR #9819, opened 2020-08-26
ACLK Version Negotiation

underhood:
##### Summary

Merge is possible after following cloud PRs are released to production:
- [x] https://github.com/netdata/cloud-agent-metadata-ctrl-service/pull/72
- [x] https://github.com/netdata/cloud-agent-metadata-ctrl-service/pull/74
- [x] Final Test With Staging environment together with Cloud
- [x] Wait for Leonidas confirms https://github.com/netdata/product/issues/1382 is not caused by recent changes/unrelated or until it is fixed
- [x] Previous step result. Following bugfixes have to be merged to Cloud https://github.com/netdata/cloud-agent-metadata-ctrl-service/pull/73/files
- [x] Tested on Staging with Leonidas
- [x] Confirmed with Cloud Team agents with these modifications can connect to Cloud

The agent will send a `hello` message to the cloud as soon as the MQTT connection is made and the agent is subscribed.
The agent will wait up to 3 seconds for a reply from the cloud and if none arrives default to an oldest supported version _(this is done so we can merge this before cloud implements support for this messages)_.
If reply arrives _(in form of `version` message)_ the highest version supported by both Agent and Cloud is selected for communication.

Messages `hello`, `version`, and LWT are versioned separately as they are sent (in case of LWT set) before the negotiation was finished.

##### Test Plan
The first test is easy. Connect agent to staging and as it doesn't reply to `hello` message you should see it to revert to version 1 (oldest supported) of ACLK after the timeout. You should see a log like this: `ACLK version negotiation failed. No reply to "hello" with "version" from the cloud in time of 3s.  Reverting to the default version of 1.`

Other tests are a bit harder but I tried to make it easier by providing some helper script that emulates Cloud with support of version negotiation.

To test this you will have to temporarily disable challenge/response to allow the agent to connect to your local MQTT broker as follows:
```
static void aclk_try_to_connect(char *hostname, char *port, int port_num)
{
    UNUSED(port);
    if (!aclk_private_key) {
            error("Cannot try to establish the agent cloud link - no private key available!");
            return;
    }
    info("Attempting to establish the agent cloud link");
    /*aclk_get_challenge(hostname, port);
    if (aclk_password == NULL)
        return;*/
    aclk_password = strdupz("anon");
    int rc;
    aclk_connecting = 1;
...
```
Then set `cloud.conf` to connect to your local mosquito broker as follows:
```
[global]
  enabled = yes
  cloud base url = 127.0.0.1:9002
```
This will work with broker started as `mosquitto -c mosquitto.conf` with `mosquito.conf` contents:
```
port 1883
listener 9002
protocol websockets
certfile server.crt
keyfile server.key
log_dest stdout
log_type all
websockets_log_level 9999
```
you will of course need your own self-signed certificate `server.crt` and `server.key` and agent compiled with `-DACLK_SSL_ALLOW_SELF_SIGNED` _(unfortunately the flag `-DACLK_DISABLE_SSL` was removed from agent, this would be useful for testing)_

Then you can use my script from repo `https://github.com/netdata/internal/` path `underhood/cloud_helpers/aclk_reply_hello.rb`. You will need `ruby` interpreter and following ruby gems (packages/libraries):
```
gem install mqtt
gem install colorize
```

Then you can start it as follows: `./aclk_reply_hello.rb [MIN_VER] [MAX_VER]`. This will wait for agent to start and reply to `hello` message with minimum and maximum ACLK version specified by parameters. This allows to test this PR while cloud doesn't reply anything yet.

As currently there is `ACLK_VERSION_MIN == ACLK_VERSION_MAX == 1` only 1 version you will have to play with these defines in `aclk_common.h` to pretend there are other versions already.

Test following cases:
- Cloud doesn't reply to message. Agent should revert to default version
- Cloud min & max version are both > than agent max version -> Agent is way too old for the cloud
- Agent min & max version are both > than max version of cloud -> Cloud is way too old for the agent
- min & max of agent and cloud partially or fully overlap

### DIFF
--- a/aclk/aclk_common.h
+++ b/aclk/aclk_common.h
@@ -12,6 +12,11 @@ extern netdata_mutex_t aclk_shared_state_mutex;
 #define ACLK_VERSION_MIN 1
 #define ACLK_VERSION_MAX 1
 
+// Version negotiation messages have they own versioning
+// this is also used for LWT message as we set that up
+// before version negotiation
+#define ACLK_VERSION_NEG_VERSION 1
+
 // Maximum time to wait for version negotiation before aborting
 // and defaulting to oldest supported version
 #define VERSION_NEG_TIMEOUT 3
--- a/aclk/aclk_common.h
+++ b/aclk/aclk_common.h
@@ -7,6 +7,11 @@ extern netdata_mutex_t aclk_shared_state_mutex;
 #define ACLK_SHARED_STATE_LOCK netdata_mutex_lock(&aclk_shared_state_mutex)
 #define ACLK_SHARED_STATE_UNLOCK netdata_mutex_unlock(&aclk_shared_state_mutex)
 
+// minimum and maximum supported version of ACLK
+// in this version of agent
+#define ACLK_VERSION_MIN 1
+#define ACLK_VERSION_MAX 1
+
 typedef enum aclk_cmd {
     ACLK_CMD_CLOUD,
     ACLK_CMD_ONCONNECT,
@@ -31,6 +36,7 @@ extern struct aclk_shared_state {
     ACLK_METADATA_STATE metadata_submitted;
     ACLK_AGENT_STATE agent_state;
     time_t last_popcorn_interrupt;
+    int negotiated_version;
 } aclk_shared_state;
 
 typedef enum aclk_proxy_type {
--- a/aclk/aclk_common.h
+++ b/aclk/aclk_common.h
@@ -76,4 +76,6 @@ void safe_log_proxy_censor(char *proxy);
 int aclk_decode_base_url(char *url, char **aclk_hostname, char **aclk_port);
 const char *aclk_get_proxy(ACLK_PROXY_TYPE *type);
 
+extern int aclk_disable_runtime;
+
 #endif //ACLK_COMMON_H
--- a/aclk/aclk_common.h
+++ b/aclk/aclk_common.h
@@ -12,6 +12,19 @@ extern netdata_mutex_t aclk_shared_state_mutex;
 #define ACLK_VERSION_MIN 1
 #define ACLK_VERSION_MAX 1
 
+// Version negotiation messages have they own versioning
+// this is also used for LWT message as we set that up
+// before version negotiation
+#define ACLK_VERSION_NEG_VERSION 1
+
+// Maximum time to wait for version negotiation before aborting
+// and defaulting to oldest supported version
+#define VERSION_NEG_TIMEOUT 3
+
+#if ACLK_VERSION_MIN > ACLK_VERSION_MAX
+#error "ACLK_VERSION_MAX must be >= than ACLK_VERSION_MIN"
+#endif
+
 typedef enum aclk_cmd {
     ACLK_CMD_CLOUD,
     ACLK_CMD_ONCONNECT,
@@ -36,7 +49,11 @@ extern struct aclk_shared_state {
     ACLK_METADATA_STATE metadata_submitted;
     ACLK_AGENT_STATE agent_state;
     time_t last_popcorn_interrupt;
-    int negotiated_version;
+
+    // read only while ACLK connected
+    // protect by lock otherwise
+    volatile int version_neg;
+    usec_t version_neg_wait_till;
 } aclk_shared_state;
 
 typedef enum aclk_proxy_type {
--- a/aclk/aclk_common.h
+++ b/aclk/aclk_common.h
@@ -12,11 +12,6 @@ extern netdata_mutex_t aclk_shared_state_mutex;
 #define ACLK_VERSION_MIN 1
 #define ACLK_VERSION_MAX 1
 
-// Version negotiation messages have they own versioning
-// this is also used for LWT message as we set that up
-// before version negotiation
-#define ACLK_VERSION_NEG_VERSION 1
-
 // Maximum time to wait for version negotiation before aborting
 // and defaulting to oldest supported version
 #define VERSION_NEG_TIMEOUT 3
--- a/aclk/aclk_common.h
+++ b/aclk/aclk_common.h
@@ -52,7 +52,7 @@ extern struct aclk_shared_state {
 
     // read only while ACLK connected
     // protect by lock otherwise
-    volatile int version_neg;
+    int version_neg;
     usec_t version_neg_wait_till;
 } aclk_shared_state;
 
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -553,15 +553,14 @@ void *aclk_query_main_thread(void *ptr)
                 " Reverting to default ACLK version of %d.", VERSION_NEG_TIMEOUT, ACLK_VERSION_MIN);
             aclk_shared_state.version_neg = ACLK_VERSION_MIN;
         }
-        if (unlikely(!aclk_shared_state.metadata_submitted)) {
-            ACLK_SHARED_STATE_UNLOCK;
+        if (unlikely(aclk_shared_state.metadata_submitted == ACLK_METADATA_REQUIRED)) {
             if (unlikely(aclk_queue_query("on_connect", NULL, NULL, NULL, 0, 1, ACLK_CMD_ONCONNECT))) {
+                ACLK_SHARED_STATE_UNLOCK;
                 errno = 0;
                 error("ACLK failed to queue on_connect command");
                 sleep(1);
                 continue;
             }
-            ACLK_SHARED_STATE_LOCK;
             aclk_shared_state.metadata_submitted = ACLK_METADATA_CMD_QUEUED;
         }
         ACLK_SHARED_STATE_UNLOCK;
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -549,7 +549,8 @@ void *aclk_query_main_thread(void *ptr)
                 sleep(1);
                 continue;
             }
-            error("ACLK version negotiation failed. No reply to \"hello\" with \"version\" from cloud in time of %ds. "
+            errno = 0;
+            error("ACLK version negotiation failed. No reply to \"hello\" with \"version\" from cloud in time of %ds."
                 " Reverting to default ACLK version of %d.", VERSION_NEG_TIMEOUT, ACLK_VERSION_MIN);
             aclk_shared_state.version_neg = ACLK_VERSION_MIN;
         }
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -348,7 +348,7 @@ static int aclk_execute_query(struct aclk_query *this_query)
         buffer_flush(local_buffer);
         local_buffer->contenttype = CT_APPLICATION_JSON;
 
-        aclk_create_header(local_buffer, "http", this_query->msg_id, 0, 0, ACLK_VERSION);
+        aclk_create_header(local_buffer, "http", this_query->msg_id, 0, 0, aclk_shared_state.version_neg);
         buffer_strcat(local_buffer, ",\n\t\"payload\": ");
         char *encoded_response = aclk_encode_response(w->response.data->buffer, w->response.data->len, 0);
         char *encoded_header = aclk_encode_response(w->response.header_output->buffer, w->response.header_output->len, 1);
@@ -538,6 +538,15 @@ void *aclk_query_main_thread(void *ptr)
 
     while (!netdata_exit) {
         ACLK_SHARED_STATE_LOCK;
+        if (unlikely(!aclk_shared_state.version_neg)) {
+            if (!aclk_shared_state.version_neg_wait_till || aclk_shared_state.version_neg_wait_till > now_monotonic_usec()) {
+                ACLK_SHARED_STATE_UNLOCK;
+                info("Waiting for ACLK Version Negotiation message from Cloud");
+                sleep(1);
+                continue;
+            }
+            aclk_shared_state.version_neg = ACLK_VERSION_MIN;
+        }
         if (unlikely(!aclk_shared_state.metadata_submitted)) {
             ACLK_SHARED_STATE_UNLOCK;
             if (unlikely(aclk_queue_query("on_connect", NULL, NULL, NULL, 0, 1, ACLK_CMD_ONCONNECT))) {
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -348,7 +348,7 @@ static int aclk_execute_query(struct aclk_query *this_query)
         buffer_flush(local_buffer);
         local_buffer->contenttype = CT_APPLICATION_JSON;
 
-        aclk_create_header(local_buffer, "http", this_query->msg_id, 0, 0);
+        aclk_create_header(local_buffer, "http", this_query->msg_id, 0, 0, ACLK_VERSION);
         buffer_strcat(local_buffer, ",\n\t\"payload\": ");
         char *encoded_response = aclk_encode_response(w->response.data->buffer, w->response.data->len, 0);
         char *encoded_header = aclk_encode_response(w->response.header_output->buffer, w->response.header_output->len, 1);
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -537,6 +537,10 @@ void *aclk_query_main_thread(void *ptr)
     }
 
     while (!netdata_exit) {
+        if(aclk_disable_runtime) {
+            sleep(1);
+            continue;
+        }
         ACLK_SHARED_STATE_LOCK;
         if (unlikely(!aclk_shared_state.version_neg)) {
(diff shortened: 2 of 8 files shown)


underhood:
rebased on latest master as it was quite behind

aclk/agent_cloud_link.c
```diff
@@ -981,6 +1018,7 @@ void *aclk_main(void *ptr)
         // TODO: Move to on-connect
         if (unlikely(!aclk_subscribed)) {
             aclk_subscribed = !aclk_subscribe(ACLK_COMMAND_TOPIC, 1);
+            aclk_hello_msg();
```
amoss, line 1021: Remove the TODO comment as it is no longer accurate.

amoss:
Having a separate version numbering scheme for the negotiation messages seems clunky and complex. It does not give us a way to update the message in the future - it seems to rule it out completely.

Using the minimum support version number for the hello messages would solve the problem cleanly.

underhood:
well it allows us to update it if versioned separatelly but we would have to do it in lockstep with cloud, let me think a bit about your idea with minimum supported version... sounds like an option

underhood:
@amoss Not sure what will cloud do if the agent is too new for cloud. Example:
Cloud min 1, max 5
Agent min 6, max 9999999

then we will send `hello` message version 6 to cloud which might Kick us instead of sending the reply with `cloud min 1, max 5` back. This has a disadvantage of us not being sure why we got kicked. Whereas if we got reply we can know and log the reason nicely.

We could, in theory, make cloud ignore the version of `hello` message but then this key loses meaning completely.

having a separate version for only those messages makes it possible to upgrade in future although we would have to do it together with cloud. We have freedom of course to implement new versions of protocol for all other messages.

underhood:
rebased on latest master which should have fix for arch CI

amoss:
If the intersection is empty then it leads to the correct result. If the cloud kicks the agent then we should get a disconnection message with a reason.

If we have a separate versioning system for the message that does the versioning then we are unable to upgrade it at all - it just recreates the same problem that we are trying to solve.

underhood:
Well only for negotiation messages which will not change that often and all the main protocol is already negotiated.
I changed to your wish with MIN_VERSION instead of separate versioning for neg. messaged but I think I prefer former.

amoss:
> I changed to your wish with MIN_VERSION instead of separate versioning for neg. messaged but I think I prefer former.

That's not really how we work. If you have technical arguments for why the other approach is better then present them.

underhood:
I do not consider sending whatever the ACLK_MIN_VERSION is as a necessarily cleaner solution to the problem.
We send the `hello` message before the protocol version is negotiated. IMHO it makes sense that it is versioned separately. We have a way to change it if we need to but mostly it will as HELLO message stay the same for a long time.

> If the intersection is empty then it leads to the correct result. If the cloud kicks the agent then we should get a disconnection message with a reason.

With what protocol version the cloud disconnection message should come? Will the agent understand it if they have no common version?

I do however accept your solution as well, just consider separate versioning cleaner.

amoss:
Placing a separate version number on the hello message does not solve the problem. It takes the same problem and moves it one step earlier in the process without a solution.

The problem that we are trying to solve:
* The agent should not send a newer version of a message than the cloud can understand.
* We should not have to synchronize releases of code between the two teams to achieve this.

The solution that should be implemented here is negotiation:
* Before sending any (higher than v1) version of a message agree on a commonly understood version.

Separate versioning of the version negotiation message does not solve this problem. We then have exactly the same problem again:
* How do we agree the version of the version-negotiation message?

This is about more than the code. If we decide to put another version number scheme in place for that message what is the deployment process for the Cloud and the Agent, and how does it prevent a mismatch? We are simply back in the same place that we are trying to avoid where we need to synchronize the release steps between the two code-bases. This is not a solution at all - either cleaner or less clean, more or less elegant - as it does not solve the release problem.

So instead of debating whether the non-solution is cleaner or more elegant you should be checking that whatever is put in place fulfills the actual objective of the work. I can see different possibilities with different risks. You should be evaluating which of them is the cleanest solution to the problem with the lowest risk.

* Assume that we get the version negotiation message correct and it will not need to change.
* Use the oldest-supported version of the protocol in the agent.
* Embed the version in the type of the message (e.g. "hello1").
* Rely on the disconnection message to reject agents completely.

Each of these has a risk of going wrong, and an impact on the upgrade of agents if it does. Pick a valid solution to the problem and implement it. Putting a separate version numbering scheme on the initial message is not a valid solution as it does not de-synchronize the release schedules between the teams.

underhood:
> Putting a separate version numbering scheme on the initial message is not a valid solution as it does not de-synchronize the release schedules between the teams.

it does except for the `hello` message but all the rest of ACLK can be released without cloud release sync

underhood:
>   Assume that we get the version negotiation message correct and it will not need to change.
- This sounds dangerous
>    Use the oldest-supported version of the protocol in the agent.
- Has issues mentioned before with how the cloud will reply (whether `version` or `disconnect`) + AFAIK the `disconnect` message is not implemented yet (it was supposed to be with Unclaiming feature). This would have as a consequence problem with the client not knowing why it got disconnected.
>    Embed the version in the type of the message (e.g. "hello1").
- Not sure how much different it is from having separate versioning for `connect` message. If we would choose this way then why still not keep it as `hello` and then continue with `hello1` as second version (cloud already has some parsing for `hello` message done)
>    Rely on the disconnection message to reject agents completely.
- As per the previous discussion what versioning disconnection message has? How can we know agents can understand it....

amoss:
As we discussed in the meeting earlier this week there might not be a clean solution to the problem at all. I had another think about this approach:

> > Use the oldest-supported version of the protocol in the agent.

> Has issues mentioned before with how the cloud will reply (whether version or disconnect) + AFAIK the disconnect message is > not implemented yet (it was supposed to be with Unclaiming feature). This would have as a consequence problem with the 
> client not knowing why it got disconnected.

This would work slightly differently. To change the version of the hello message we would not be able to perform a single release / update in the agent. It would take at least two, but it may result in decoupled release cycles.

1. Update the agent from version `n-1` to `n` with support for the new hello message. It would not be sent as the oldest supported version would still be `n-1`.
2. Cloud adds support for version `n`.
3. Deprecate version `n-1` in the agent with a second release / update. The oldest minimum would then be `n` and so we would send the new version of the hello message.
4. Cloud deprecates support for version `n-1` to block old agents and force upgrade.

The ordering of the steps is still fixed, but the timing between them can vary and I think that the combinations of versions in the field at any time should work. What is the equivalent with separate versioning on the message - does it look simpler or more complex?

underhood:
Well with separate versioning it would look something like this I guess:
```
1. cloud releases V2 hello msg support first (they can support V1 as well)
2. we switch agent to V2 hello
3. cloud removes V1 hello support when there are no agents connecting with V1 anymore or their number is insignificant
```
all other messages (except hello/version and LWT) are versioned and can be updated anytime.

amoss:
That looks about right. We could skip step (1) in my list as it doesn't achieve much and then the two lists align.

So both approaches would work the same way, which implies that we get the same coupling effect from either. Which gives us simpler code?

underhood:
Well code is same (use one or other define). But wouldn't then using minimum ACLK version as `hello` version couple deprecating old versions of protocol and having cloud support new hello version first?

amoss:
If the new version of the message is the same as the old one then it doesn't add any work to deprecating the version. I thought that it simplified the code slightly because you can remove an argument to the create header procedure - this is not a big deal. It seems that ultimately each of the approaches is about the same - you pick.

underhood:
the create header procedure will still need argument because depending on negotiated version we will send different version in the header so this will stay in regardless.

I think I will stay with separate versioning for now.

aclk/aclk_common.h
```diff
@@ -31,6 +49,11 @@ extern struct aclk_shared_state {
     ACLK_METADATA_STATE metadata_submitted;
     ACLK_AGENT_STATE agent_state;
     time_t last_popcorn_interrupt;
+
+    // read only while ACLK connected
+    // protect by lock otherwise
+    volatile int version_neg;
```
amoss, line 55: I don't think that volatile is needed here as access is controlled by a lock.

underhood: yes thank you, I think you are right that volatile is not needed in the final solution.
it is protected by lock during the initialization phase (where we can write to it). Afterward (when ACLK is running) read-only access is allowed and thus lock is not necessary.

aclk/aclk_query.c
```diff
@@ -537,16 +537,30 @@ void *aclk_query_main_thread(void *ptr)
...
-        if (unlikely(!aclk_shared_state.metadata_submitted)) {
-            ACLK_SHARED_STATE_UNLOCK;
+        if (unlikely(!aclk_shared_state.version_neg)) {
+            if (!aclk_shared_state.version_neg_wait_till || aclk_shared_state.version_neg_wait_till > now_monotonic_usec()) {
+                ACLK_SHARED_STATE_UNLOCK;
+                info("Waiting for ACLK Version Negotiation message from Cloud");
+                sleep(1);
+                continue;
+            }
+            error("ACLK version negotiation failed. No reply to \"hello\" with \"version\" from cloud in time of %ds. "
```
stelfrag, line 552: You can set  `errno = 0`  o avoid misleading info e.g.

```
ACLK version negotiation failed. No reply to "hello" with "version" from cloud in time of 3s.  Reverting to default ACLK version of 1. (errno 22, Invalid argument)
```

underhood: thanks, will change.
I am thinking though since this is repeating a lot -> should we have some `error` equivalet that ignores `errno` to avoid having to set it to 0 in bazzilion places in the code...

underhood: `3s.  Reverting` removed the double space as well :D

aclk/agent_cloud_link.c
```diff
@@ -820,12 +829,35 @@ static void aclk_try_to_connect(char *hostname, char *port, int port_num)
...
     rc = mqtt_attempt_connection(hostname, port_num, aclk_username, aclk_password);
     if (unlikely(rc)) {
         error("Failed to initialize the agent cloud link library");
     }
 }
 
+// Sends "hello" message to negotiate ACLK version with cloud
+static inline void aclk_hello_msg()
+{
+    BUFFER *buf = buffer_create(NETDATA_WEB_RESPONSE_HEADER_SIZE);
```
stelfrag, line 845: You are not releasing the buffer

```
==208869==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==208869==    by 0x1420E1: callocz (libnetdata.c:183)
==208869==    by 0x13B75B: buffer_create (buffer.c:387)
==208869==    by 0x1DE6CC: aclk_hello_msg (agent_cloud_link.c:845)
==208869==    by 0x1DED60: aclk_main (agent_cloud_link.c:1020)
```

underhood: fixed, thx

aclk/agent_cloud_link.c
```diff
@@ -1443,75 +1481,174 @@ int aclk_update_alarm(RRDHOST *host, ALARM_ENTRY *ae)
...
+    if (aclk_stats_enabled) {
+        ACLK_STATS_LOCK;
+        aclk_metrics_per_sample.cloud_req_recvd++;
+        ACLK_STATS_UNLOCK;
+    }
+
     if (unlikely(!payload)) {
         debug(D_ACLK, "ACLK incoming message is empty");
-        return 0;
+        goto err_cleanup;
```
stelfrag, line 1606: note: this works only because the cloud_to_agent is zero-ed out. The structures are valid only after json parsing. You could save some processing time by jumping to a new label right before ACLK stats recording.

aclk/agent_cloud_link.c
```diff
@@ -1443,75 +1481,174 @@ int aclk_update_alarm(RRDHOST *host, ALARM_ENTRY *ae)
...
 
-        if (aclk_stats_enabled) {
-            ACLK_STATS_LOCK;
-            aclk_metrics_per_sample.cloud_req_err++;
-            ACLK_STATS_UNLOCK;
+    for (int i = 0; aclk_incoming_msg_types[i].name; i++) {
+        if (strcmp(cloud_to_agent.type_id, aclk_incoming_msg_types[i].name) == 0) {
+            if (likely(!aclk_incoming_msg_types[i].fnc(&cloud_to_agent))) {
+                freez(cloud_to_agent.type_id);
+                return 0;
```
stelfrag, line 1629: I think this will leak memory (not going through err_cleanup)

aclk/agent_cloud_link.c
```diff
@@ -1443,75 +1482,174 @@ int aclk_update_alarm(RRDHOST *host, ALARM_ENTRY *ae)
...
+    }
+    if(unlikely(!cloud_to_agent->min_version)) {
+        error("Min version missing or 0");
+        return 1;
+    }
+    if(unlikely(!cloud_to_agent->max_version)) {
+        error("Max version missing or 0");
+        return 1;
+    }
+    if(unlikely(cloud_to_agent->max_version < cloud_to_agent->max_version)) {
```
amoss, line 1542: Typo! max should be min

aclk/agent_cloud_link.c
underhood: Well it happens only on an error and would only skip 4 null checks at the end so the impact is minimal
but I will improve it to be perfectionist :)

aclk/agent_cloud_link.c
underhood: the handlers are supposed to clean after themselves. in case of cloud request, the `aclk_queue_query` will take ownership of pointers so we cannot free it! this was how it was before this PR

underhood: This is done to avoid need to copy the strings when we add them to queue!

aclk/agent_cloud_link.c
underhood: thanks this is a very very good catch... hard to spot

aclk/agent_cloud_link.c
underhood: What I can do though is to explicitly document this. I think that is good idea

amoss:
This looks bad
cloud-crash

underhood:
@amoss hmmm never saw that... any info how to reproduce? are we sure it is related to this PR?

amoss:
I'm going to wipe out the install directory and reinstall it to check:
1. Install with no CFLAGS (default options).
2. Create a new room on app.netdata.cloud.
3. Claim the node.

On the table-view the charts (sparklines) are all missing, click on the node view to see it blow up.

underhood:
Seems OK here
Screenshot from 2020-09-03 18-57-51
Screenshot from 2020-09-03 18-58-03

underhood:
@amoss thank you will try to reproduce your steps (reclaim the node).

amoss:
Log looks fine on the agent side:
```
2020-09-03 18:54:13: netdata INFO  : ACLK_Query_1 : Waiting for ACLK Version Negotiation message from Cloud
2020-09-03 18:54:14: netdata INFO  : ACLK_Query_0 : Waiting for ACLK Version Negotiation message from Cloud
2020-09-03 18:54:14: netdata INFO  : ACLK_Query_1 : Waiting for ACLK Version Negotiation message from Cloud
2020-09-03 18:54:15: netdata INFO  : ACLK_Query_0 : Waiting for ACLK Version Negotiation message from Cloud
2020-09-03 18:54:15: netdata INFO  : ACLK_Query_1 : Waiting for ACLK Version Negotiation message from Cloud
2020-09-03 18:54:16: netdata ERROR : ACLK_Query_0 : ACLK version negotiation failed. No reply to "hello" with "version" from cloud in time of 3s. Reverting to default ACLK version of 1.
```

underhood:
I see the same thing after reclaim.... very strange.. I think you found some cloud bug

underhood:
@amoss Tried with master now. Same bug.... IMHO we found critical cloud bug. Maybe related to changes done in Cloud for this PR but not by this PR directly.

amoss:
Happened again the second time, slightly more detailed steps:
```
./netdata-installer.sh --install /home/amoss/netdata/dev-install-review --disable-ebpf --disable-telemetry --dont-start-it
```
1. Go to `localhost:19999` in Chrome.
2. Click the sign in button
3. Go to room
4. Manage room / add nodes / copy and and paste claiming command.
5. Node is claimed, World is broken.

underhood:
@amoss Cloud team is investigating this. Doesn't seem to be caused by bug in this PR
@thiagoftsm has the issue as well in master (he never tried to run this PR) + some users report issues as well

amoss (approved):
Passes basic exploratory testing without problems (ignoring the concurrent incident in the cloud). Good looks fine, only minor problems spotted which are addressed.

stelfrag (approved):

underhood:
Thank you guys. Just to be paranoid, I will wait for Leonidas confirms Cloud is ready for this and then merge.

Issue #9824, opened 2020-08-26
Build failures on Arch Linux due to system not being up to date.


Ferroin:
Minimal reproducer:

1. From the Netdata source directory, start an Arch environment with `docker run -it --rm -v $PWD:/netdata archlinux:latest /bin/bash`
2. Run `/netdata/packaging/installer/install-required-packages.sh --dont-wait --non-interactive netdata-all`
3. Run `cd /netdata && autoreconf -ivf && configure`

The configure script will fail to find libuv, however `pkg-config --modversion libuv` properly finds it and manual inspection of the relevant paths show it is indeed there.

Relevant configure log output for the failed check is:

```
configure:7688: checking for uv_fs_scandir_next in -luv
configure:7713: gcc -o conftest -g -O2 -pthread   conftest.c -luv    >&5
/usr/sbin/ld: /usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../lib/libuv.so: undefined reference to `pthread_sigmask@GLIBC_2.32'
collect2: error: ld returned 1 exit status
configure:7713: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "netdata"
| #define PACKAGE_TARNAME "netdata"
| #define PACKAGE_VERSION "v1.24.0-125-ga305ee79"
| #define PACKAGE_STRING "netdata v1.24.0-125-ga305ee79"
| #define PACKAGE_BUGREPORT ""
| #define PACKAGE_URL ""
| #define PACKAGE "netdata"
| #define VERSION "v1.24.0-125-ga305ee79"
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define __EXTENSIONS__ 1
| #define _ALL_SOURCE 1
| #define _GNU_SOURCE 1
| #define _POSIX_PTHREAD_SEMANTICS 1
| #define _TANDEM_SOURCE 1
| #define HAVE_FUNC_ATTRIBUTE_RETURNS_NONNULL 1
| #define HAVE_FUNC_ATTRIBUTE_MALLOC 1
[...]

PR #9825, opened 2020-08-26
Ensure Arch Linux images are up to date in CI before running checks.

Ferroin:
##### Summary

SSIA

##### Component Name

area/ci

##### Test Plan

Build process tested locally in Arch Linux, checks verified on my fork.

##### Additional Information

Fixes: #9824

### DIFF
--- a/.github/workflows/build-and-install.yml
+++ b/.github/workflows/build-and-install.yml
@@ -59,7 +59,7 @@ jobs:
             rmjsonc: 'apk del json-c-dev'
 
           - distro: 'archlinux:latest'
-            pre: 'pacman --noconfirm -Sy grep libffi'
+            pre: 'pacman --noconfirm -Syu && pacman --noconfirm -Sy grep libffi'
 
             #- distro: 'centos:8'
             #rmjsonc: 'dnf remove -y json-c-devel'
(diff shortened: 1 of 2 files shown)


prologic (approved):

.github/workflows/build-and-install.yml
```diff
@@ -59,7 +59,7 @@ jobs:
             rmjsonc: 'apk del json-c-dev'
 
           - distro: 'archlinux:latest'
-            pre: 'pacman --noconfirm -Sy grep libffi'
+            pre: 'pacman --noconfirm -Syu && pacman --noconfirm -Sy grep libffi'
```
amoss, line 62: If you use `-Syyu` it will upgrade the package lists if necessary.

amoss (approved):
Please check the additional flag above, merge if it is irrelevant otherwise I can re-approve after a change.

.github/workflows/build-and-install.yml
Ferroin: Not quite. `-yy` forces a refresh of the lists, just `-y` will do one if needed, which will always happen here because Arch strips the package cache from their images before the publish them.

Issue #9796, opened 2020-08-20
REST API / Charts not displaying correctly


Unfairz:
##### Bug report summary

Hello, it appears that recently the netdata API isn't returning properly formed JSON and visualizing the charts is failing. I have also double-checked the code on here - https://learn.netdata.cloud/docs/agent/web/api and can confirm that the code, grabbing the charts is the same, however, in the console output, I am getting malformed JSON errors.

```
Error: Parse error on line 1:
{version:'0.6',reqId:
-^
Expecting 'STRING', '}'
```
##### OS / Environment

```
CloudLinux release 7.8 (Alexei Leonov) 
```

##### Netdata version
```
netdata v1.24.0
```
##### Component Name
```
Netdata REST API / Google Charts API
```
##### Steps To Reproduce

1. In order to test it, simply copy and pasted the code given in your Wiki page or directly call the API:
```
/api/v1/data?after=-60&format=datasource&options=nonzero&chart=system.load#
```
##### Expected behavior
JSON is formed correctly and charts are displayed on the page.

Thank you for your time and let me know if any further information is needed :sparkles:

ilyam8:
Hi

could you show https://github.com/netdata/netdata/issues/9745#issuecomment-674019270

could be the same issue

Unfairz:
Sure! Here is the output:

<details>
<summary>JSON Output</summary>

```
{
version: "v1.24.0",
uid: "3eec1832-831b-11ea-a9c9-42010a9a0022",
mirrored_hosts: [
"X"
],
alarms: {
normal: 187,
warning: 0,
critical: 1
},
os_name: "CloudLinux",
os_id: "cloudlinux",
os_id_like: "rhel fedora centos",
os_version: "7.8 (Alexei Leonov)",
os_version_id: "7.8",
os_detection: "/etc/os-release",
cores_total: "2",
total_disk_space: "111454736384",
cpu_freq: "2000000000",
ram_total: "3972739072",
container_os_name: "none",
container_os_id: "none",
container_os_id_like: "none",
container_os_version: "none",
container_os_version_id: "none",
kernel_name: "Linux",
kernel_version: "3.10.0-962.3.2.lve1.5.38.el7.x86_64",
architecture: "x86_64",
virtualization: "kvm",
virt_detection: "systemd-detect-virt",
container: "unknown",
container_detection: "none",
host_labels: {
_os_name: "CloudLinux",
_os_version: "7.8 (Alexei Leonov)",
[...]

PR #9826, opened 2020-08-26
Remove Google Charts info from API doc

joelhans:
##### Summary

As requested by @jacekkolasa in #9796—we're removing some of this information as it's quite outdated, and thus no one really knows how it works or why the example code in jsfiddle fails in the user's tests. They're moving to the officially-supported dashboard features anyway.

##### Component Name

area/docs
area/web

### DIFF
--- a/web/api/README.md
+++ b/web/api/README.md
@@ -1,8 +1,6 @@
 <!--
----
 title: "API"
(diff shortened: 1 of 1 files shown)


amoss (approved):

jacekkolasa (approved):
thanks @joelhans !

Issue #9829, opened 2020-08-27
Curl error while running netdata-updater


davidmehren:
##### Bug report summary
I'm running netdata nightly on a number of hosts. This morning, I got a cron mail from each of them, as `/etc/cron.daily/netdata-updater` seems to output either `curl: (3) <url> malformed
` or `curl: (3) URL using bad/illegal format or missing URL` to stderr.
Running `netdata-update` manually yields (the output seems to be the same on all hosts affected, just with the different error message from curl):
```
Thu 27 Aug 2020 09:38:15 AM CEST : INFO:  Running on a terminal - (this script also supports running headless from crontab)                                                          
curl: (3) URL using bad/illegal format or missing URL
Thu 27 Aug 2020 09:38:15 AM CEST : ERROR:  FAILED TO UPDATE NETDATA : Cannot download /latest-version.txt                                                                            
Thu 27 Aug 2020 09:38:15 AM CEST : INFO:  Current Version: 001024000128
Thu 27 Aug 2020 09:38:15 AM CEST : INFO:  Latest Version: 000000000000
7a58066f5c1b411321115578feb516e2e89513e8ca65aa75f298cfd0fdf92da5 *./netdata-latest.tar.gz
7a58066f5c1b411321115578feb516e2e89513e8ca65aa75f298cfd0fdf92da5 *./netdata-v1.24.0-128-g46e868f9.tar.gz                                                                             
Thu 27 Aug 2020 09:38:15 AM CEST : INFO:  Newest version is already installed
```

##### OS / Environment
This happened on my VM-hosts runnning Proxmox/Debian 10 and on both LXC containers and VMs on them running various versions of Ubuntu and Debian.

Detailed info of one of my Proxmox hosts:
```
Linux iceman 5.4.55-1-pve #1 SMP PVE 5.4.55-1 (Mon, 10 Aug 2020 10:26:27 +0200) x86_64 GNU/Linux
/etc/os-release:PRETTY_NAME="Debian GNU/Linux 10 (buster)"
/etc/os-release:NAME="Debian GNU/Linux"
/etc/os-release:VERSION_ID="10"
/etc/os-release:VERSION="10 (buster)"
/etc/os-release:VERSION_CODENAME=buster
/etc/os-release:ID=debian
/etc/os-release:HOME_URL="https://www.debian.org/"
/etc/os-release:SUPPORT_URL="https://www.debian.org/support"
/etc/os-release:BUG_REPORT_URL="https://bugs.debian.org/"
[...]

PR #8808, opened 2020-04-23
Fix the netdata-updater.sh to correctly pass REINSTALL_OPTIONS (finally)

prologic:
Fixes #8782
Fixes #9763 

##### Summary

ssia

##### Component Name

- area/packaging

##### Test Plan

__WARNING:__ Testing this is a _bit tricky_ to say the least. It's a bit of a chicken 'n egg type problem until this PR is merved.

1. First of all, you need to patch the `netdata-updater.sh` on your system with teh following patch:

```#!diff
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -123,13 +123,12 @@ set_tarball_urls() {
   if [ "$1" = "stable" ]; then
     local latest
     # Simple version
-    # latest="$(curl -sSL https://api.github.com/repos/netdata/netdata/releases/latest | grep tag_name | cut -d'"' -f4)"
     latest="$(download "https://api.github.com/repos/netdata/netdata/releases/latest" /dev/stdout | grep tag_name | cut -d'"' -f4)"
     export NETDATA_TARBALL_URL="https://github.com/netdata/netdata/releases/download/$latest/netdata-$latest.${extension}"
     export NETDATA_TARBALL_CHECKSUM_URL="https://github.com/netdata/netdata/releases/download/$latest/sha256sums.txt"
   else
-    export NETDATA_TARBALL_URL="https://storage.googleapis.com/netdata-nightlies/netdata-latest.${extension}"
-    export NETDATA_TARBALL_CHECKSUM_URL="https://storage.googleapis.com/netdata-nightlies/sha256sums.txt"
+    export NETDATA_TARBALL_URL="$NETDATA_NIGHTLIES_BASEURL/netdata-latest.${extension}"
+    export NETDATA_TARBALL_CHECKSUM_URL="$NETDATA_NIGHTLIES_BASEURL/sha256sums.txt"
   fi
 }

@@ -185,8 +184,8 @@ update() {
       do_not_start="--dont-start-it"
     fi

-    if [ -n "${NETDATA_SELECTED_DASHBOARD}" ] ; then
-        env="NETDATA_SELECTED_DASHBOARD=${NETDATA_SELECTED_DASHBOARD}"
+    if [ -n "${NETDATA_SELECTED_DASHBOARD}" ]; then
+      env="NETDATA_SELECTED_DASHBOARD=${NETDATA_SELECTED_DASHBOARD}"
     fi

     info "Re-installing netdata..."
[...]

PR #9831, opened 2020-08-27
Fixed bug in installer introduced by #8808

Ferroin:
##### Summary

In short, you can't do default value substitution in Bash if the default value has parameter expansion meta-characters in it. Simple local test that demonstrates this: `echo ${foo:-http://example.com}` will echo an empty string in Bash.

The result of this is that the updater tries to use a malformed URL for the version check, which in turn always fails.

Users will need to manually handle updates once the next nightly goes out after this is merged.

##### Component Name

area/packaging

##### Test Plan

Verified via local testing.

##### Additional Information

Fixes: #9829 

Originally introduced by #8808

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -272,7 +272,7 @@ export NETDATA_LIB_DIR="${NETDATA_LIB_DIR:-${NETDATA_PREFIX}/var/lib/netdata}"
 [[ -z "${NETDATA_TARBALL_CHECKSUM}" ]] && [[ -f ${NETDATA_LIB_DIR}/netdata.tarball.checksum ]] && NETDATA_TARBALL_CHECKSUM="$(cat "${NETDATA_LIB_DIR}/netdata.tarball.checksum")"
 
 # Grab the nightlies baseurl (defaulting to our Google Storage bucket)
-export NETDATA_NIGHTLIES_BASEURL="${NETDATA_NIGHTLIES_BASEURL:-https://storage.googleapis.com/netdata-nightlies}"
+[ -z "${NETDATA_NIGHTLIES_BASEURL}" ] && export NETDATA_NIGHTLIES_BASEURL="https://storage.googleapis.com/netdata-nightlies"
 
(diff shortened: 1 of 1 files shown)


amoss:
This example works for me in bash:
```
echo $foo

echo ${foo:-http://example.com}
http://example.com
```

amoss:
In case it is relevant:
```
> shopt
autocd         	off
assoc_expand_once	off
cdable_vars    	off
cdspell        	off
checkhash      	off
checkjobs      	off
checkwinsize   	on
cmdhist        	on
compat31       	off
compat32       	off
compat40       	off
compat41       	off
compat42       	off
compat43       	off
compat44       	off
complete_fullquote	on
direxpand      	off
dirspell       	off
dotglob        	off
execfail       	off
expand_aliases 	on
extdebug       	off
extglob        	on
extquote       	on
failglob       	off
force_fignore  	on
globasciiranges	on
globstar       	off
gnu_errfmt     	off
histappend     	on
histreedit     	off
histverify     	off
hostcomplete   	off
huponexit      	off
inherit_errexit	off
interactive_comments	on
lastpipe       	off
lithist        	off
localvar_inherit	off
localvar_unset 	off
login_shell    	on
mailwarn       	off
no_empty_cmd_completion	off
nocaseglob     	off
nocasematch    	off
nullglob       	off
progcomp       	on
progcomp_alias 	off
promptvars     	on
restricted_shell	off
shift_verbose  	off
sourcepath     	on
xpg_echo       	off
```

ilyam8:
> Thu 27 Aug 2020 09:38:15 AM CEST : ERROR:  FAILED TO UPDATE NETDATA : Cannot download /latest-version.txt                                                                            

https://github.com/netdata/netdata/blob/ea8b0e976bdd39849e0d91ae7d209c82bae79bcf/packaging/installer/netdata-updater.sh#L144

It looks like `$NETDATA_TARBALL_BASEURL` is not set

prologic (approved):

Ferroin:
This did not actually fix anything, and is reverted in #9834.

PR #9832, opened 2020-08-27
Replace streaming code with replication.

amoss:
##### Summary

Fixes: #7327

---

When both ends of a streaming connection negotiate a sufficient version the streaming code is disabled completely and replication is used to transfer metrics data.

* The standard BEGIN / SET / END is disabled.
* A new REPBEGIN / REPDIM / REPEND communicates windows of metric data.
* The data transferred is `stored_number`s taken from the sending nodes database.
* The data is stored directly in the receiving node's database.
* When gaps are detected (network drops or node restarts) explicit requests with window times are sent from receiver to sender.
* The sender fills the window with as much data as possible, the `window_end communicates the progress even if the window is empty.
* No `colllected_number` values are sent across the link.
* The data is only interpolated once on the original node.

##### Component Name

streaming

##### Test Plan

T.B.D.

##### Additional Information

It is currently unclear what effect this change will have on the health code and the evaluation of alarms on the receiving node.


streaming/sender.c
```diff
@@ -714,3 +733,128 @@ void *rrdpush_sender_thread(void *ptr) {
...
+                    debug(D_REPLICATION, "%s.%s query handle finished early @%ld", st->id, rd->id, sample_t);
+                    break;
+                }
+
+                storage_number n = rd->state->query_ops.next_metric(&handle, &metric_t);
+                if (n==SN_EMPTY_SLOT) {
+                    debug(D_REPLICATION, "%s.%s db empty in valid dimension range @ %ld", st->id, rd->id, sample_t);
+                }
+                if (sample_t != metric_t) {
+                    debug(D_REPLICATION, "%s.%s Sample mismatch during replication %ld vs %ld", st->id, rd->id,
```
mfundul, line 801: Data replication cannot depend on `update_every`. `dbengine` supports variable granularity in data collection. As a result, a query can return results that contain gaps and various frequencies, e.g.:

| | Data points |
| ------------- | ------------- |
| frequency |  1, 1, 1, 1, GAP, ... GAP, 2, 2, 2, 2, 2, 2, 2, GAP, ... GAP, 3 |

You should remove the `for` loop condition `sample_t < rd_end` since you can potentially lose data points this way. `sample_t` does not seem like it's giving useful information in general.

streaming/receiver.c
```diff
@@ -51,50 +51,230 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
...
+    //time_t st_last = rrdset_last_entry_t(user->st);  UNUSED?
+    if (rd == NULL) {
+        errno = 0;
+        error("Unknown dimension \"%s\" on %s during replication - ignoring", id, user->st->name);
         return PARSER_RC_OK;
     }
+
+    if (user->st->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE)
+    {
+        rd->state->collect_ops.store_metric(rd, timestamp * USEC_PER_SEC, value);
```
mfundul, line 204: You should detect `SN_EMPTY_SLOT` here. Whenever you detect a gap, you flush the metric and you don't call `store_metric()` anymore. This has 2 advantages:
- You don't write gaps to the database. Imagine a 1 month gap for instance.
  - You still transmit the 1-month gap though, which should also be fixed.
- By flushing at the gap, you align the first non-empty data point to a new page, thus supporting `update_every` changes. You must avoid having data points with variable granularity in the same page at all costs:
  - e.g. [1112221111] will just put the data points with the correct starting and ending timestamps, but all intermediate timestamps will be incorrectly inferred)

mfundul (changes requested):

streaming/sender.c
mfundul: Moreover, it is possible that the `next_metric()` call will jump in time, so the frequency change must be inferred.

streaming/receiver.c
mfundul: the frequency change must also be inferred by detecting time jumps.

streaming/sender.c
amoss: Fixed

amoss:
Latest commit will address @mfundul's concerns about gaps and changes in collector frequency on the sending side. This is currently being retested and a second commit will follow to handle page flushing on the receiver side.

amoss:
Ok, so the testing run shows that this breaks all of the non-dbengine configurations because `metric_next` has a different interface in the legacy modes. I will fix the interface after I check that it won't break anything that relies on it...

streaming/receiver.c
amoss: I don't think that we should detect SN_EMPTY_SLOT here. We are trying to replicate the state of the collecting node's db. If there is an empty slot at that point in the time-series (which can occur for several different reasons) then we should create one here. We should stick to the case that the `update_every` has changed, which we can infer:
* Inside `rrdeng_store_metric_next()`
*  `spacing = (descr->end_time - decr->start_time) / (descr->page_length / sizeof(number))`
* `if ((point_in_time - desc->end_time) % spacing != 0)` then there is a change.
On a change we should call `rrdeng_store_metric_flush_current_page(rd)` to force a new page after the change in time-frame alignment. This will collapse any gap immediately before a change in `update_every` - which probably does exist because the only valid scenario is restarting the agent.

amoss: Commit with this change incoming and then I will rerun the testing set in parallel to check it has broken any of the existing cases.

amoss:
Testing was not a huge success, although my crash wrapper works :) Fix incoming after lunch:
```
Thread 6 "PLUGIN[diskspac" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x7fb27db75700 (LWP 100)]
0x000055b6277d2b1a in rrdeng_store_metric_next (rd=0x55b62983f310, 
    point_in_time=1598955940000000, number=560039084)
    at database/engine/rrdengineapi.c:264
264	        if (0 != ((point_in_time - descr->end_time) % tf_spacing))
Stack level 0, frame at 0x7fb27db73870:
 rip = 0x55b6277d2b1a in rrdeng_store_metric_next
    (database/engine/rrdengineapi.c:264); saved rip = 0x55b6277c4a00
 called by frame at 0x7fb27db73970
 source language c.
 Arglist at 0x7fb27db73860, args: rd=0x55b62983f310, 
    point_in_time=1598955940000000, number=560039084
 Locals at 0x7fb27db73860, Previous frame's sp is 0x7fb27db73870
 Saved registers:
  rbp at 0x7fb27db73860, rip at 0x7fb27db73868
rd = 0x55b62983f310
point_in_time = 1598955940000000
number = 560039084
tf_spacing = 0
handle = 0x55b62983f700
ctx = 0x55b627964dc0 <multidb_ctx>
pg_cache = 0x55b627a6cf58 <multidb_ctx+1081752>
descr = 0x55b629880570
page = 0x55b6278e4002
must_flush_unaligned_page = 0 '\000'
perfect_page_alignment = 1 '\001'
tf_alignment_change = 0 '\000'
__FUNCTION__ = "rrdeng_store_metric_next"
#0  0x000055b6277d2b1a in rrdeng_store_metric_next (rd=0x55b62983f310, 
    point_in_time=1598955940000000, number=560039084)
    at database/engine/rrdengineapi.c:264
#1  0x000055b6277c4a00 in rrdset_done_interpolate (st=0x55b62983a010, 
    update_every_ut=1000000, last_stored_ut=1598955939000000, 
    next_store_ut=1598955940000000, last_collect_ut=1598955939356813, 
    now_collect_ut=1598955940352978, store_this_entry=1 '\001', 
    storage_flags=16777216) at database/rrdset.c:1351
#2  0x000055b6277c6bc4 in rrdset_done (st=0x55b62983a010)
    at database/rrdset.c:1877
#3  0x000055b6278231e3 in do_disk_space_stats (mi=0x55b6298200f0, 
    update_every=1) at collectors/diskspace.plugin/plugin_diskspace.c:287
#4  0x000055b627823831 in diskspace_main (
    ptr=0x55b62793d3d0 <static_threads+112>)
    at collectors/diskspace.plugin/plugin_diskspace.c:392
#5  0x000055b62776aa40 in thread_start (ptr=0x55b62971ebe0)
    at libnetdata/threads/threads.c:184
#6  0x00007fb27fd88fa3 in start_thread (arg=<optimized out>)
    at pthread_create.c:486
#7  0x00007fb27fcb94cf in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) quit
```

amoss:
Current test-run looks promising, over the roughly 80 integration tests it leaves a small gap in one chart in one test. I will look into what the cause is there as it is certainly some kind of race-hazard between the replication and the restart of the node in the test, but overall it seems quite stable.

There are two known limitations that we need to understand and discuss before any merge of the code (so I will leave it in [WIP] for now). (cc @cakrit )

1. We never considered how health should work with replicated blocks of data. It was outside the scope of the original issue, and now all data in the system is replicated.
2. I have not put in the pass-through of requests for multi-level hierarchies yet. So child-proxy-parent configurations will still see gaps in charts at the top-level. This is not a regression on what we have. This code improves the current situation dramatically and I don't think we should delay merging it until the next piece is ready.

amoss:
The last failure only occurs when the parent node is dbengine and it produces gaps in the time-series that are stored in the database. It would seem that there is still a problem with the (3 line!) addition to the dbengine that causes it to flush too many pages and this causes gaps in the data. I assume that a sequence of single-value pages was outside the design-spec for the dbengine so I will look at why that code is still causing too many flushes.

amoss:
Ok, I think that was the last issue that is now fixed and the tester will now re-run over all the integration tests. The only real thing left to do is we need to discuss the impact on health and how we handle merging this code.

amoss:
@thiagoftsm This is now ready for you to start reviewing.

database/engine/rrdengineapi.c
```diff
@@ -191,8 +191,7 @@ void rrdeng_store_metric_flush_current_page(RRDDIM *rd)
         return;
     }
     if (likely(descr->page_length)) {
-        int page_is_empty;
-
+        int /*ret, */page_is_empty;
```
thiagoftsm, line 194: @amoss the variable `ret` was removed with the PR https://github.com/netdata/netdata/pull/9845, please, rebase your PR.

database/rrdset.c
```diff
@@ -1089,6 +1091,64 @@ inline void rrdset_next_usec(RRDSET *st, usec_t microseconds) {
 // ----------------------------------------------------------------------------
 // RRDSET - process the collected values for all dimensions of a chart
 
+void rrdset_dump_debug_state(RRDSET *st) {
+#ifdef NETDATA_INTERNAL_CHECKS
+    if (debug_flags&D_REPLICATION) {
```
thiagoftsm, line 1096: please, add a space between the variables and the bitwise operator to be more readable.

streaming/receiver.c
```diff
@@ -51,47 +51,235 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
...
+    PARSER_USER_OBJECT *user = user_v;
+    UNUSED(plugins_action);
+    char *id = words[1];
+    char *start_txt = words[2];
+    char *first_txt = words[3];
+    char *end_txt = words[4];
+    if (!id || !start_txt || !end_txt)
+        goto disable;
+
+    RRDSET *st = NULL;
```
thiagoftsm, line 103: you are setting `NULL` for a variable that is assigned in the next line with a new value, so this value is never used.

streaming/receiver.c
```diff
@@ -51,47 +51,235 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
...
-                 host->hostname, cd->cmd, remote_time, now, now-remote_time);
-        else {
-            gap = now - prev;
-            info("STREAM %s from %s: Checking for gaps... remote=%ld local=%ld..%ld slew=%ld  %ld-sec gap",
-                 host->hostname, cd->cmd, remote_time, prev, now, remote_time - now, gap);
+#else
+    ret = send(host->receiver->fd, message, strlen(message), MSG_DONTWAIT);
+#endif
+    if (ret != (int)strlen(message))
+        error("Failed to send replication request for %s!", st_id);
```
thiagoftsm, line 70: We have a `void` function that is responsible to send data and when it fails it writes a message inside `error.log`, but the fail can happens, because the connection was lost. Instead `void`, I think this function needs to return a value for the caller indicates to the thread that a reconnection is necessary.

streaming/receiver.c
```diff
@@ -51,47 +51,235 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
...
+            st->state->ignore_block = 0;
+            return PARSER_RC_OK;
         } else {
-            ret = send(host->receiver->fd, message, strlen(message), MSG_DONTWAIT);
+            debug(D_REPLICATION, "Initial data for %s, asking for history window=%ld..%ld",
+                  st->name, (long)now - rpt->gap_history+1, now);
+            send_replication_req(user->host, st->id, now - rpt->gap_history + 1, now);
+            st->state->ignore_block = 1;
+            st->last_updated.tv_sec = now - rpt->gap_history;
+            return PARSER_RC_OK;
```
thiagoftsm, line 127: To complement what I wrote above in the previous function, I agree that when we reach this point the PARSER happened, but there was a problem to send data.

streaming/receiver.c
```diff
@@ -51,47 +51,235 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
...
+    return PARSER_RC_ERROR;
+}
+
+PARSER_RC streaming_rep_end(char **words, void *user_v, PLUGINSD_ACTION *plugins_action) {
+    PARSER_USER_OBJECT *user = user_v;
+    UNUSED(plugins_action);
+    char *num_points_txt = words[1];
+    char *col_total_txt  = words[2];
+    char *last_total_txt = words[3];
+    if (!num_points_txt)
```
thiagoftsm, line 237: You are not verifying the addresses for `col_total_txt` and `last_total_txt`, but you use them in the lines 251 and 252.

streaming/sender.c
```diff
@@ -715,3 +733,122 @@ void *rrdpush_sender_thread(void *ptr) {
...
+
+            for (time_t metric_t = rd_oldest; metric_t < rd_end; ) {
+
+                if (rd->state->query_ops.is_finished(&handle)) {
+                    debug(D_REPLICATION, "%s.%s query handle finished early @%ld", st->id, rd->id, metric_t);
+                    break;
+                }
+
+                storage_number n = rd->state->query_ops.next_metric(&handle, &metric_t);
+                if (n==SN_EMPTY_SLOT)
```
thiagoftsm, line 800: please add spaces between comparison and variables.

database/engine/rrdengineapi.c
amoss: I already rebased to remove the conflict. This PR will be adding it back in as comment. This is because the code that uses it is still there commented out until we decide on a more permanent fix for the flushing problems.

streaming/receiver.c
amoss: This is done, I'll leave you to resolve the conversation when you are happy with the change in `store_next`

database/rrdset.c
amoss: Fixed

streaming/receiver.c
amoss: Fixed.

streaming/receiver.c
amoss: The replication requests are pumped by an external event. Each incoming point on the stream causes a replication request to be sent if the windows do not match. So although we do not explicitly return an error, there is a retry that is caused because the system has not reached the target state. At this point we do not know if a failure is transient or permanent so we rely on the logic in the main loop for the socket to handle errors / closures.

streaming/receiver.c
amoss: If the data was not sent then it is retried the next the event happens - the overall design of the sender is idempotent (with respect to the changing state of the window). The return code indicates that the message was correctly parsed and the connection should be left open (if the parse is invalid then the stream is aborted).

streaming/receiver.c
amoss: Fixed.

streaming/sender.c
amoss: Fixed.

thiagoftsm:
I executed the following tests:

### First test

1 - I compiled this PR on `parent`, and I configured the following alarm:

```
 alarm: dev_dim_template
    on: system.cpu
    os: linux
 units: %
 every: 1s
  warn: $this > 1
  crit: $this > 4
```

After this I started Netdata.

2 - I compiled the current master on `child` (a VM with the same Slackware Linux current and two processors) and I started it.  After few second I ran on child the following : 

```
for i in `seq 1 1000000` ; do echo $i > /dev/null ; done
```

In this initial test everything worked as expected. I had alarms and the dashboards were synchronized.

### Second test

I kept Netdata `parent` running and I stopped the child, after this I got the version [1.12.2](https://github.com/netdata/netdata/releases/download/v1.12.2/netdata-v1.12.2.gz.run) and I installed it in the default directory `/opt/netdata`. After to configure the file `/opt/netdata/etc/netdata/stream.conf`, I started this child. As expected I had another child registered in my `parent`.

After this I ran again the next command on `child`:

```
for i in `seq 1 1000000` ; do echo $i > /dev/null ; done
```

And the alarms were raised. 

I would like to call attention for the fact that I had on my dashboard two links with the same text, the child name, but with different links. I think we can improve this, but it is subject for another issue.

### Third test

I stopped Netdata 1.12.2 on `child` and I compiled the current PR on child. After this I started the child, please, pay attention for the fact that I did stop the parent any time since the first test.

After the `child` starts the stream, I reload the `parent` dashboard and I observed that I did not have more two links for the `childs`, I was not hopping a behavior like this. When I clicked on link, the `child` dashboard did not open more. And I saw that I have the following error messages on child:

```
2020-09-02 11:07:38: netdata ERROR : STREAM_SENDER[zeus] : STREAM zeus [send to 192.168.0.12:19999]: connection closed by far end. Restarting connection
2020-09-02 11:07:38: netdata INFO  : STREAM_SENDER[zeus] : STREAM zeus [send to 192.168.0.12:19999]: connecting...
2020-09-02 11:07:38: netdata INFO  : STREAM_SENDER[zeus] : STREAM zeus [send to 192.168.0.12:19999]: initializing communication...
2020-09-02 11:07:38: netdata INFO  : STREAM_SENDER[zeus] : STREAM zeus [send to 192.168.0.12:19999]: waiting response from remote netdata...
2020-09-02 11:07:38: netdata INFO  : STREAM_SENDER[zeus] : STREAM zeus [send to 192.168.0.12:19999]: established communication with a parent using protocol version 4 - ready to send metrics...
```

This happens 82 times in an interval of 8 minutes and 42 seconds.

On the master side, I had the errors:

```
2020-09-02 11:07:32: netdata ERROR : STREAM_RECEIVER[zeus,[192.168.0.12]:48602] : STREAM zeus [receive from [192.168.0.12]:48602]: disconnected (completed 0 updates).
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48602] : STREAM zeus [receive from [192.168.0.12]:48602]: receive thread ended (task id 17259)
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48602] : thread with task id 17259 finished
2020-09-02 11:07:32: netdata INFO  : WEB_SERVER[static3] : clients wants to STREAM metrics.
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : thread created with task id 17327
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : set name of thread 17327 to STREAM_RECEIVER
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : STREAM zeus [192.168.0.12]:48620: receive thread created (task id 17327)
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : STREAM zeus [receive from [192.168.0.12]:48620]: client willing to stream metrics for host 'zeus' with machine_guid 'bf4ffeea-7e88-11ea-8e02-080027900376': update every = 1, history = 924, memory mode = dbengine, health enabled, SSL, tags ''
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : STREAM zeus [receive from [192.168.0.12]:48620]: initializing communication...
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : STREAM zeus [receive from [192.168.0.12]:48620]: Netdata is using the stream version 4.
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : Postponing health checks for 60 seconds, on host 'zeus', because it was just connected.
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : STREAM zeus [receive from [192.168.0.12]:48620]: receiving metrics...
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.web_thread1_cpu' on host 'zeus' already exists.
2020-09-02 11:07:32: netdata INFO  : MAIN : Finished metadata log compaction (id:14).
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.web_thread2_cpu' on host 'zeus' already exists.
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'system.idlejitter' on host 'zeus' already exists.
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.plugin_tc_cpu' on host 'zeus' already exists.
2020-09-02 11:07:32: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.plugin_tc_time' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : ACLK_Query_0 : Sending MQTT len=1066 mid=799 wq=0 (0-bytes) readq=0:       {"type": "status-change",       "msg-id": "c0956579-1541-461b-ad
2020-09-02 11:07:33: netdata INFO  : ACLK_Query_0 : Sending MQTT len=1050 mid=800 wq=1 (1132-bytes) readq=0:    {"type": "status-change",       "msg-id": "63dee807-460d-48c3-ad
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.plugin_cgroups_cpu' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.aclk_status' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.aclk_query_per_second' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.aclk_latency_mqtt' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.aclk_write_q' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'netdata.aclk_read_q' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'disk_space._run' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'disk_inodes._run' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'disk_space._dev' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'disk_inodes._dev' on host 'zeus' already exists.
2020-09-02 11:07:33: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : RRDSET: chart name 'disk_space._' on host 'zeus' already exists.

2020-09-02 11:07:39: netdata ERROR : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : STREAM zeus [receive from [192.168.0.12]:48620]: disconnected (completed 0 updates).
2020-09-02 11:07:39: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : STREAM zeus [receive from [192.168.0.12]:48620]: receive thread ended (task id 17327)
2020-09-02 11:07:39: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:48620] : thread with task id 17327 finished
```

It looks like that in conditions where the database is never reset before the  update, the new code is unstable.

amoss:
This is an interesting test case. What happens if you reset the parent after upgrading the child - do they successfully connect or does it continue to fail?

thiagoftsm:
> This is an interesting test case. What happens if you reset the parent after upgrading the child - do they successfully connect or does it continue to fail?

I reset completely the database on `parent` and `child` and I am having these messages on `parent`:

```
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : STREAM zeus [receive from [192.168.0.12]:49634]: initializing communication...
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : STREAM zeus [receive from [192.168.0.12]:49634]: Netdata is using the stream version 4.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : Postponing health checks for 60 seconds, on host 'zeus', because it was just connected.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : STREAM zeus [receive from [192.168.0.12]:49634]: receiving metrics...
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'netdata.apps_cpu' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'netdata.apps_sizes' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'netdata.apps_fix' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'netdata.apps_children_fix' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'apps.cpu' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'apps.cpu_user' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'apps.cpu_system' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'apps.threads' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'apps.processes' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'apps.uptime' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata ERROR : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : Failed to send replication request for apps.uptime! (errno 32, Broken pipe)
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'apps.uptime_min' on host 'zeus' already exists.
2020-09-02 12:53:06: netdata ERROR : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : Failed to send replication request for apps.uptime_min! (errno 32, Broken pipe)
2020-09-02 12:53:06: netdata INFO  : STREAM_RECEIVER[zeus,[192.168.0.12]:49634] : RRDSET: chart name 'apps.uptime_avg' on host 'zeus' already exists.
```
And thanks this the streaming is not happening.

thiagoftsm:
Another thing @amoss , while I was repeating the tests you required, I observed that this time I did not have alarms for `child` when I executed the `first` and `second` tests described [here](https://github.com/netdata/netdata/pull/9832#issuecomment-685675534) and I also observed that this time all hosts were kept on dashboard. I am thinking the problems probably have more relationship with the database than your PR (cc: @mfundul and @stelfrag )

database/rrddim.c
```diff
@@ -307,11 +309,20 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
                     memset(rd, 0, size);
                     reset = 1;
                 }
-                else if(dt_usec(&now, &rd->last_collected_time) > (rd->entries * rd->update_every * USEC_PER_SEC)) {
+                // Collected times are valid on the collecting node
+                else if(host == localhost &&
+                    dt_usec(&now, &rd->last_collected_time) > (rd->entries * rd->update_every * USEC_PER_SEC)) {
```
mfundul, line 314: You should probably deduplicate `dt_usec(&now, &rd->last_collected_time) > (rd->entries * rd->update_every * USEC_PER_SEC)` by putting its value in a variable.

mfundul: ah, scratch that, I see that it's different after all.

thiagoftsm:
I could replicate the same problems reported [here](https://github.com/netdata/netdata/pull/9832#issuecomment-685675534) with memory mode `save` and `ram`.

I also observed something that created concerns while I was testing another PR and it can have relationship with this PR, but I did not have time to test this today. When the `child` has memory mode set as `dbengine` and the plugin `example` is enabled, it always send the dimensions that were stored for the first time inside the database. If I stop Netdata completely and increase or decrease the dimensions inside the plugin, these new values are completely ignored.

thiagoftsm:
Finally, @amoss I observed that `Test plan` was not completely defined, please, can you give us the complete list of tests you made while was developing? Thank you!

amoss:
> Finally, @amoss I observed that `Test plan` was not completely defined, please, can you give us the complete list of tests you made while was developing? Thank you!

That's a good point! I'll make a second PR that adds the integration tests so that you can see them.

amoss:
The integration testing code is in https://github.com/netdata/netdata/pull/9890, instructions are in the test-plan for that PR.

thiagoftsm:
> The integration testing code is in #9890, instructions are in the test-plan for that PR.

Thank you, I will take a look on it.

mfundul:
Set-up:
- parent node: UTC+3, dbengine
- child node: UTC+1, ram
- alarms enabled

I get a lot of:
```
2020-09-07 13:49:47: netdata INFO  : STREAM_SENDER[localhost.localdomain] : STREAM localhost.localdomain [send to 192.168.56.164]: connecting...
2020-09-07 13:49:47: netdata INFO  : STREAM_SENDER[localhost.localdomain] : STREAM localhost.localdomain [send to 192.168.56.164]: initializing communication...
2020-09-07 13:49:47: netdata INFO  : STREAM_SENDER[localhost.localdomain] : STREAM localhost.localdomain [send to 192.168.56.164]: waiting response from remote netdata...
2020-09-07 13:49:47: netdata INFO  : STREAM_SENDER[localhost.localdomain] : STREAM localhost.localdomain [send to 192.168.56.164]: established communication with a parent using protocol version 4 - ready to send metrics...
2020-09-07 13:49:48: netdata ERROR : STREAM_SENDER[localhost.localdomain] : STREAM localhost.localdomain [send to 192.168.56.164]: buffer full (1048576-bytes) after 1048576 bytes. Restarting connection
```
in the child node.

I get a lot of:
```
2020-09-07 15:50:51: netdata INFO  : WEB_SERVER[static2] : Ignoring metric with out of order timestamp.
```
and a lot of:
```
2020-09-07 15:49:58: netdata ERROR : STREAM_RECEIVER[localhost.localdomain,[192.168.56.107]:50166] : Attempt to store older/duplicate point in db: 1599482973000000 < 1599482994000000
```
and:
```
2020-09-07 16:24:09: netdata ERROR : WEB_SERVER[static2] : INTERNAL ERROR: overwriting the timestamp of RRDR line 136 from 1599485022 to 1599485048, of chart 'system.cpu_pressure'
2020-09-07 16:24:09: netdata INFO  : WEB_SERVER[static2] : INTERNAL ERROR: rrd2rrdr() on system.cpu_pressure update every 2 with aligned grouping average (group: 1, resampling_time
: 0, resampling_group: 1), after (got: 1599484751, want: 1599484750, req: 1599484748, db: 1599484616), before (got: 1599485048, want: 1599485048, req: 1599485048, db: 1599485048), 
duration (got: 299, want: 300, req: 300, db: 434), points (got: 137, want: 150, req: 150, db: 5), stopped, because attempted to access the db after 'wanted before'
```
in the parent node.

Moreover, I can notice irregular updates in the live dashboard of the parent for the child, during the connection losses. During the timestamp overwrite events the charts for the child become corrupted.

Finally, while health is enabled for the child in the parent, there are no alarms whatsoever for some reason. I can't get any alarms even if I don't use dbengine, or if I synchronize the clocks.

A potential explanation for the `buffer full` situation is that gap filling tries to send too big a message?

amoss:
The default maximum gap size is 60 seconds, this turns into a few hundred bytes of message. Have you changed the default gap size, or are you running the child in a situation where it loses the CPU for long periods between iterations of the sender thread main loop?

mfundul:
No, I haven't done anything non standard for the child, other than the things mentioned above.

mfundul:
One of the reasons for the disappearing alarms is probably related to this:
https://github.com/netdata/netdata/blob/b5a3fadf68e754e85f5a9fda996839d2a2f90b0f/health/health_json.c#L298

amoss:
Yes, that is the kind of interaction between health and streaming that I was worried about. Tomorrow I'll look at whether it is easier to put a synthetic last_collected time in the chart (e.g. 0.5s into the interval) or update the health code...

So for the three things that you've found we need to isolate the difference between your setup and the testing that I did:
1. The socket dying and filling the buffer.
2. The timestamp difference between the two ends.
3. The case where the charts blow up after a restart.

If you look at the integration-testing PR case (3) should be ChildShortRestart and you can do the test directly using:
`python3 test_scenario.py ChildShortRestart --child=save --parent=dbengine`. I'll go through the logs tomorrow and see if it works differently for me in VMs to the behaviour in docker.

thiagoftsm:
I did two tests now and I observed something interesting with this PR:

### First test

On this test I had Slackware current 100% updated on both `parent` and `child`. On this scenario I tested:

####  `Parent` running master and child running `this PR`

Running on this situation everything worked as expected. I tested with TLS enabled and disabled and I always had the stream working.

####  Both parent and child running this PR

On this situation I never had streaming working, and I have the following logs on parent:

```
/var/log/netdata/error.log:2020-09-07 21:52:43: netdata INFO  : STREAM_RECEIVER[ceres,[192.168.0.17]:48818] : (0571@streaming/:rrdpush_receive): STREAM ceres [receive from [192.168.0.17]:48818]: initializing communication...
/var/log/netdata/error.log:2020-09-07 21:52:43: netdata INFO  : STREAM_RECEIVER[ceres,[192.168.0.17]:48818] : (0574@streaming/:rrdpush_receive): STREAM ceres [receive from [192.168.0.17]:48818]: Netdata is using the stream version 4.
/var/log/netdata/error.log:2020-09-07 21:52:43: netdata INFO  : STREAM_RECEIVER[ceres,[192.168.0.17]:48818] : (0628@streaming/:rrdpush_receive): Postponing health checks for 60 seconds, on host 'ceres', because it was just connected.
/var/log/netdata/error.log:2020-09-07 21:52:43: netdata INFO  : STREAM_RECEIVER[ceres,[192.168.0.17]:48818] : (0637@streaming/:rrdpush_receive): STREAM ceres [receive from [192.168.0.17]:48818]: receiving metrics...
/var/log/netdata/error.log:2020-09-07 21:53:44: netdata ERROR : STREAM_RECEIVER[ceres,[192.168.0.17]:48818] : (0647@streaming/:rrdpush_receive): STREAM ceres [receive from [192.168.0.17]:48818]: disconnected (completed 0 updates).
/var/log/netdata/error.log:2020-09-07 21:53:44: netdata INFO  : STREAM_RECEIVER[ceres,[192.168.0.17]:48818] : (0047@streaming/:rrdpush_receive): STREAM ceres [receive from [192.168.0.17]:48818]: receive thread ended (task id 30174)
```

and on child I had:

```
../../var/log/netdata/error.log:2020-09-07 22:30:11: netdata INFO  : STREAM_SENDER[ceres] : (0162@streaming/:rrdpush_sender_): STREAM ceres [send to 85.25.177.37:19999]: connecting...
../../var/log/netdata/error.log:2020-09-07 22:30:11: netdata INFO  : STREAM_SENDER[ceres] : (0178@streaming/:rrdpush_sender_): STREAM ceres [send to 85.25.177.37:19999]: initializing communication...
../../var/log/netdata/error.log:2020-09-07 22:30:11: netdata INFO  : STREAM_SENDER[ceres] : (0320@streaming/:rrdpush_sender_): STREAM ceres [send to 85.25.177.37:19999]: waiting response from remote netdata...
../../var/log/netdata/error.log:2020-09-07 22:30:12: netdata INFO  : STREAM_SENDER[ceres] : (0370@streaming/:rrdpush_sender_): STREAM ceres [send to 85.25.177.37:19999]: established communication with a parent using protocol version 4 - ready to send metrics...
../../var/log/netdata/error.log:2020-09-07 22:31:13: netdata ERROR : STREAM_SENDER[ceres] : (0649@streaming/:rrdpush_sender_): STREAM ceres [send to 85.25.177.37:19999]: could not send metrics for 60 seconds - closing connection - we have sent 939 bytes on this connection via 1 send attempts.
```

These messages are happening during the whole time that Netdata is running.

### Second test

I installed this PR on both parent and child. The parent is running `CentOS 8` while the child is running `FreeBSD 12.1-RELEASE-p9`. On this scenario the streaming started without problems. 

After I have the stream running, I stopped the parent 20 seconds and when the child reconnected, it could stream the data from the time that the parent restart, but I could not see the data that were stored before the reboot. The parent chart on the other hand I had the data before the reboot and after the reboot with the gap well specified.

On this test and the next, the parent is at a different timezone of the childs.

### Third test

This time I used the script `./packaging/makeself/build-x86_64-static.sh` to build a static binary to run on Slackware current. When the compilation finished, I used the same `CentOS 8` described in the previous test and I installed the static Netdata on Slackware child. This time I had the streaming working and I also have the same problem reported in the previous test after the parent restart.

### Conclusion

I need to continue investigating what is happening when parent and child are Slackware current, something is happening that does not allow me to do stream on this situation.

I have serious concerns with the fact that the dashboard is not showing data stored before the reboot and this is happening with any operate system I tested.

amoss:
I've done some investigation on case (3) with @mfundul and the root cause seems to be differing clocks between the two systems. The original scope was defined as assume that both ends of the system are using NTP so that the clocks are similar. In the test-case where the charts blow up the following case happens:

1. There is a 3.5 minute difference between the two clocks
2. The gap-filling limit is the default config of 60s
3. A three minute restart-gap is created on the child

As a result the replication request time-interval collides with a time-range that as valid data and the gap disappears. I'm tempted to say "notabugwontfix" but I think the original scope was wrong and this case seems realistic. I will setup the VM nodefarm that we used for ACLK testing so that I can play around with clock differences.

amoss:
We did some more testing and there are now two definite problems that I know about that I can fix:
1. The alarms
2. Replicating a gap that is longer than the max-replication period fails in a particular way.

I'll push commits for these later and then go back to the other problems, like the socket dying and the buffer filling,

mfundul:
`dbengine` changes look good.

mfundul:
I haven't tested this yet but it's possible that `memory mode = none` in the child nodes will not work at all because they only use collected values and have no database.

amoss:
I think your reasoning is correct. We should disable replication for that memory mode.

amoss:
The fresh bug in health (not the one that I am looking for) is: https://github.com/netdata/netdata/blob/f9f813cd463173acda2f64b275bb6f3f365e5964/health/health.c#L1039

The realtime clock can go backwards in some circumstances, in which case:
* The delta will be negative
* The health thread will sleep until a restart.

The obvious fix is to use the monotonic clock, but it has slightly different constraints on the delta values and we would need to do a lot of testing to see how it impacts the health loop.

amoss:
Had a chat offline with @mfundul and we've agreed the following list of changes to get this ready for merging:

1. Identify which of the recent commits has caused the memory corruption in the dimension handle with the `0xbebebebebebebebe` values.
2. Get `health_main` to access `last_stored_value` correctly so that calculation variables are shown correctly.
3. Investigate the transient connection problem with the dead socket.
4. Disable replication when `memory mode = none` or with a new config setting to allow for situations where replication cannot be used (e.g. different clocks on the parent and child node that would affect alarm windows).

database/rrdset.c
```diff
@@ -1089,6 +1091,65 @@ inline void rrdset_next_usec(RRDSET *st, usec_t microseconds) {
...
+                                 " col_counter=%zu"
+                                 " col_volume=" CALCULATED_NUMBER_FORMAT
+                                 " store_volume=" CALCULATED_NUMBER_FORMAT
+                                 " last_coll_time=%ld.%ld",
+                                 st->id, rd->id, rd->calculated_value, rd->last_calculated_value,
+                                 rd->last_stored_value, rd->collected_value, rd->last_collected_value,
+                                 rd->collections_counter, rd->collected_volume,
+                                 rd->stored_volume, rd->last_collected_time.tv_sec, rd->last_collected_time.tv_usec);
+            #ifdef ENABLE_DBENGINE
+            if (st->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE && rd->state->handle.rrdeng.descr) {
```
mfundul, line 1120:
```suggestion
            if (st->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE && !rrddim_flag_check(rd, RRDDIM_FLAG_ARCHIVED) && rd->state->handle.rrdeng.descr) {
```
I can't reproduce the crash for some reason, but this is one potential problem I can think of.

amoss: This was it, confirmed that it is is fixed.

amoss:
@mfundul These are the broken values that I see in the docker / bare-metal combo - is this similar to your test?

broken-alarm

amoss:
It's nearly there. The last commit fixes the calculation variables, but the alarms still do not trigger
broken-alarm

amoss:
Correction: it works fine. I misread the two thresholds in the expression. After throwing another CPU core on the bonfire it triggered correctly.

amoss:
I've rerun the integration test set. There are some isolated failures, but none of them are regressions. There are three kinds:

1. Occasionally the request for the load averages returns no data. This is a false positive caused by the delays being slightly unpredictable and the chart only updating every 5s.
2. Occasionally the latest sample is on the child node, but not on the parent. This is a false positive caused by the sample in the test (i.e. it has not been replicated yet when the sample is taken).
3. Occasionally the first sample is not on the parent node - this happens in some tests where the length of the maximum gap does not over the data in the child because the length of the testing sequence is slightly unpredictable (startup time of the agent in docker).

Overall the false positives can be ignored, the rest of the tests pass. None of the last set of changes have caused regressions.

mfundul:
It looks like the sender logic misses some bytes when it overflows, I can't see any retries. Is this the case?

amoss:
> It looks like the sender logic misses some bytes when it overflows, I can't see any retries. Is this the case?

The new code has the same underlying limitation as the old code: groups of REPBEGIN/REPDIM/REPEND messages should be considered as atomic but are not. This is for exactly the same reasons that BEGIN/SET/END were not atomic. It would require a major architectural change to the parser, how it accesses the internal data-model and how it interfaces with streaming.

When the buffer fills and causes a break in this sequence the protocol recovers as before, so there is no regression over the existing code. The buffer filling can be prevented either by increasing the buffer size to account for the 60s of all metrics that are replicated when the connection is brought up, or the block size for replication can be reduced.

The protocol does not allow for transactions (for the reasons given above) and so the replication is best-effort rather than guaranteed.

streaming/receiver.c
```diff
@@ -137,20 +137,21 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
         if (rpt->receiver_tx_spawn)
             receiver_tx_thread_stop(rpt);
 
+        netdata_mutex_lock(&rpt->host->receiver_lock);
```
amoss, line 140: Beware: rpt->host can be NULL

mfundul: True, nice catch.

amoss:
Another month swings past. Is this ever going to be merged?

splastunov:
I've been waiting for this feature since I started using netdata

underhood:
should we update the stock config to include those keys (maybe commented out) ?

mfundul:
> 
> 
> should we update the stock config to include those keys (maybe commented out) ?

Sounds useful, maybe not commented out but disabled instead.

underhood (changes requested):

sauber:
> This pull request has been mentioned on **Netdata Community**. There might be relevant details there:
> 
> https://community.netdata.cloud/t/is-stream-compression-possible/750/2

My request is reduce network bandwidth for STREAM for example by applying compression.

database/rrdhost.c
```diff
@@ -784,8 +784,11 @@ void rrdhost_free(RRDHOST *host) {
                 netdata_thread_cancel(host->receiver->thread);
             netdata_mutex_unlock(&host->receiver_lock);
             struct receiver_state *rpt = host->receiver;
-            while (host->receiver && !rpt->exited)
+            int i;
+            for (i = 0 ; host->receiver && !rpt->exited && i < 20 ; ++i)
                 sleep_usec(50 * USEC_PER_MS);
+            if (20 == i)
+                info("Deadlock resolved during shutdown of streaming receiver context for host %s.", host->hostname);
```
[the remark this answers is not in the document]
underhood: why not make the thread joinable and wait for it that way? Then destroying receiver state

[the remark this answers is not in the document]
mfundul: The problem is that `join()` will block for ever if we hit this specific race condition, because the thread will never be canceled. Fixing the race condition will probably take a lot of refactoring of the agent, it's not obvious to me.

database/engine/rrdengineapi.c
```diff
@@ -179,8 +179,7 @@ void rrdeng_store_metric_flush_current_page(RRDDIM *rd)
...
```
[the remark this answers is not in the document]
mfundul: Andrew kept it in a comment for the code in lines 203 - 206

streaming/README.md
```diff
@@ -227,6 +227,60 @@ The receiving end (proxy or parent) logs entries like these:
...
+node (receiver) replication request. Default value of `60`.
+
+`history gap replication` is the maximum number of seconds the receiver can go back in the past and request data points
+from the sender's database, when doing the initial synchronization per netdata chart. Default value of `60`.
+
+`max gap replication` is the maximum number of empty data points the receiver will attempt to fill with the
+corresponding data points from the sender's database. Default value of `60`.
+
+To enable explicit database replication you need to set `enable replication = yes` in the `stream` configuration
+section. Fore example:
```
thiagoftsm, line 247: We have typo error here, instead `Fore`, it is expected `For`.

streaming/receiver.c
```diff
@@ -51,47 +162,264 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
...
+    req.st_id = st_id;
+    req.start = start;
+    req.end = end;
+    receiver_tx_enq_cmd(rpt, &req);
+}
+
+
+void send_replication_req(RRDHOST *host, char *st_id, time_t start, time_t end) {
+    char message[RRD_ID_LENGTH_MAX+60];
+    snprintfz(message, RRD_ID_LENGTH_MAX + 60, "REPLICATE %s %ld %ld\n", st_id, start, end);
```
thiagoftsm, line 179: The second argument of this function is not using our pattern, it needs to be updated for `RRD_ID_LENGTH_MAX + 59`, or the variable declaration needs to have one more byte.

streaming/receiver.c
```diff
@@ -51,47 +162,264 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
...
         }
-        char message[128];
-        sprintf(message,"REPLICATE %ld %ld\n", remote_time - gap, remote_time);
-        int ret;
-#ifdef ENABLE_HTTPS
-        SSL *conn = host->stream_ssl.conn ;
-        if(conn && !host->stream_ssl.flags) {
-            ret = SSL_write(conn, message, strlen(message));
+    }
+    time_t expected_t = st->last_updated.tv_sec + st->update_every;
```
thiagoftsm, line 254: I think it would be better do not use suffix `_t` to name variables, this is expected for data type as you can see in its own declaration ( `time_t` ).

streaming/receiver.c
```diff
@@ -51,47 +162,264 @@ static void rrdpush_receiver_thread_cleanup(void *ptr) {
...
+    char *time_txt  = words[2];
+    char *value_txt = words[3];
+
+    if (user->st == NULL) {
+        errno = 0;
+        error("Received RRDDIM for %s out of sequence", id);
+        goto disable;
+    }
+
+
```
thiagoftsm, line 316: Please, remove this blank line.

ilyam8:
Closing in favor of #13048

PR #9834, opened 2020-08-27
Fixed updater bug introduced by incomplete variable rename in #8808.

Ferroin:
##### Summary

#8808 renamed a variable in `netdata-updater.sh`, but did not rename all instances of the variable, resulting in a failure in some cases for the update process. This finishes the rename so that things work properly again.

This also reverts the changes from #9831 which were intended to fix this issue but did so incorrectly due to not understanding the real underlying issue.

##### Component Name

area/packaging

##### Test Plan

Properly locally verified in a clean environment.

##### Additional Information

Fixes: #9829 
Supersedes: #9831 
Originally bug introduced by #8808 

The original testing I did for #9831 which suggested that that would fix it was in fact bogus (I had forgotten I had a really strange `.bashrc` left over from before I switched to ZSH for normal interactive usage, and that caused the failures that I was seeing in manual testing).

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -17,7 +17,7 @@
 # Optional environment options:
 #
 #  - TMPDIR (set to a usable temporary directory)
-#  - NETDATA_TARBALL_BASEURL (set the base url for downloading the dist tarball)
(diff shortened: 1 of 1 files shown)


ilyam8 (approved):

amoss (approved):

prologic:
Thanks @Ferroin !

Issue #9659, opened 2020-08-04
the dashboard scales units only when the values are being changed


ilyam8:
@jacekkolasa scale units works strange - the dashboard scales units only when the values are being changed (looks like)

- 2nd chart has `Hz`

<img width="1175" alt="Screenshot 2020-07-31 at 17 41 30" src="https://user-images.githubusercontent.com/22274335/89046046-1ec30680-d355-11ea-9576-d4ef0f1f852e.png">

- 3rd chart has `bytes`

<img width="1189" alt="Screenshot 2020-07-31 at 17 42 56" src="https://user-images.githubusercontent.com/22274335/89046170-4a45f100-d355-11ea-890a-50ca801798b0.png">

_Originally posted by @ilyam8 in https://github.com/netdata/netdata/pull/9639#issuecomment-667156144_

---

Script to reproduce the bug
<details>
<summary>script click click</summary>

```python
from random import SystemRandom
 
from bases.FrameworkServices.SimpleService import SimpleService
 
priority = 90000
 
ORDER = [
    'random_Hz_dynamic',
    'random_Hz_static',
    'random_bytes_dynamic',
    'random_bytes_static',
]
 
CHARTS = {
    'random_Hz_dynamic': {
        'options': [None, 'A random number', 'Hz', 'Hz', 'Hz', 'line'],
        'lines': [
            ['random1']
        ]
    },
    'random_Hz_static': {
        'options': [None, 'A random number', 'Hz', 'Hz', 'Hz', 'line'],
        'lines': [
            ['random2']
        ]
    },
    'random_bytes_dynamic': {
        'options': [None, 'A random number', 'bytes', 'bytes', 'bytes', 'line'],
        'lines': [
            ['random3']
        ]
    },
    'random_bytes_static': {
        'options': [None, 'A random number', 'bytes', 'bytes', 'bytes', 'line'],
        'lines': [
            ['random4']
        ]
    }
}
 
 
class Service(SimpleService):
    def __init__(self, configuration=None, name=None):
        SimpleService.__init__(self, configuration=configuration, name=name)
        self.order = ORDER
        self.definitions = CHARTS
        self.random = SystemRandom()
 
    @staticmethod
    def check():
        return True
 
    def get_data(self):
        data = dict()
 
        for i in range(1, 5):
            dimension_id = ''.join(['random', str(i)])
            if i % 2 == 0:
                data[dimension_id] = 1000 * 1000 * 1000 * self.random.randint(1, 100)
            else:
[...]

Issue #9386, opened 2020-06-21
Automatic Unit Conversion Does Not Work For d3pie Charts


NoahTroy:
I currently have a pull request open to add support for Vnstat to netdata. I've noticed, however, that the d3pie charts I'd like to use do not automatically convert all units, like the stacked charts do. This has been noticed by several testers across different OS's. You can find the full code/implementation in [this pull request](https://github.com/netdata/netdata/pull/8955).

If you do not want to set up your own server, feel free to [look at my test server](http://51.79.87.85:19999/#menu_vnstat_submenu_average_data_rate_ens3;theme=slate) as an example, which already has the Vnstat plugin installed, and is currently demonstrating the problem with certain units (for example, values provided in kilobits/s convert fine, but not values provided in bytes).

If you have any questions or need any more information, please feel free to let me know.

jacekkolasa:
@NoahTroy you were absolutely right, d3pie was not using auto-scaling properly, fix is merged, thanks!

NoahTroy:
I'm glad it's been fixed, thank you! :)

Issue #9371, opened 2020-06-18
Absence of browser preferred language setting breaks graphs


amichelic:
##### Bug report summary
When there's no preferred language for web pages set in the browser, charts stop working (JavaScript error). Nothing displayed at all.

##### OS / Environment
Netdata platform doesn't matter. Verified in Firefox: 68.8.0esr (Linux/amd64).

##### Netdata version
netdata v1.22.1-302-g5bc76f3b

##### Component Name
web (dashboard)

##### Steps To Reproduce
1. Open Firefox
2. about:preferences
3. "Choose your preferred language for displaying pages" - click "Choose"
4. remove all language settings so that the list is completely empty
5. click Ok
6. all charts disappear, JS console shows a few stack traces; first one:
```
RangeError: "invalid language tag: "
    xAxisTimeString date-time.ts:97
    axisLabelFormatter dygraph-chart.tsx:236
    x dygraph-tickers.js:435
    a dygraph-tickers.js:214
    addXTicks_ dygraph.js:1937
```

##### Expected behavior
Expect charts to still show data, regardless of language setting.

nodiscc:
This also happens if you set `user_pref("intl.accept_languages",				"en-US, en");` in a Firefox user.js file (such as https://github.com/pyllyukko/user.js/blob/master/user.js#L216). Disabling this preference makes the dashboard work again.

jacekkolasa:
@amichelic @nodiscc thanks for your input! The fix is merged and will be added in next release

PR #9837, opened 2020-08-27
dashboard v1.4.2

jacekkolasa:
##### Summary
Disable calls to netdata.cloud when `--disable-cloud` option is used during installation
netdata/dashboard#114
fixes: netdata/netdata#9043

Y-axis and auto-scaling for constant values fix
netdata/dashboard#115
netdata/dashboard#117
fixes: netdata/netdata#9659

Fix broken dashboard when browser is configured to have no preferred language
netdata/dashboard#118
fixes: netdata/netdata#9371

Fix d3-pie chart unit conversion on updates
netdata/dashboard#119
fixes: netdata/netdata#9386

##### Component Name
web/dashboard


VLegakis (approved):
Great Jacek! Only small detail is I’ve already published 1.4.2 on npm! It happened during that package lock problem we had the other day while I was trying to resolve it. Sorry about that. It won’t happen again.

prologic (approved):

jacekkolasa:
@VLegakis i think that's not a big problem, Dashboard App and Dashboard npm package can be treated as separate entities, we just use the same versioning to have easier time tracking changes :) But thanks for info!

Issue #6133, opened 2019-05-27
[feature] collect active processes limit


ilyam8:
#6116 uses hardcoded maximum number of active processes (default actually).

it is configurable value, we could collect it and calculate system active processes usage % and warn user based on it.

some info from @apardyl 

> Yes, we can do that. We would probably also need to read /proc/sys/kernel/threads-max - pid_max is the max PID that can be assigned, threads-max is the actual limit of threads/processes (by default higher than pid_max, but can be set lower that it), at least if I understand http://man7.org/linux/man-pages/man5/proc.5.html correctly.
I'm not sure if this is needed - 32768 processes is quite a number, I don't think many users get close to it on a single (virtual)machine. I can implement this as well, I see it would require adding one or two (if we want to check threads-max) new files to monitor to the proc.plugin.

___

Expected behaviour:
 - collect active processes limit and add it as a cvariable to the `system.active_processes` chart.
 - use collected value as limit.

cakrit:
`active_processes` is in `collectors/freebsd.plugin/freebsd_sysctl.c` and `collectors/proc.plugin/proc_loadavg.c`
cc @vlvkobal

Ferroin:
@apardyl is mostly correct, except that threads on Linux functionally have to have an assigned PID, so the highest possible PID value (the value listed in `/proc/sys/kernel/pid_max`) is a hard limit regardless of the maximum number of threads, while the `threads-max` file functionally places a hard limit on how many threads a single process can create.  So, the correct thing to track here is the max PID, not the max number of threads.

Also, just for the record, 32768 is only a lot of processes on small systems averaged over the long-term.  Because of that whole 'each thread must have a PID' thing, it can be really inadequate on large systems doing large scale parallel computation (or _lots_ of pthread-based asynchronous I/O).  Even on smaller systems though, it may not be unusual to get close to that limit.  When sitting idle, my laptop has about 1.2k assigned PID's at any given time, and it's not unusual for it to go above the 10k mark at times.  Also, it's not unusual for admins to bump up the max PID to mitigate the issues that could arise from frequent reuse of PID's.

mfundul:
> 
> 
[...]

PR #9838, opened 2020-08-28
collect active processes limit feature #6133

Ancairon:
##### Summary
Added feature mentioned on #6133 
##### Component Name
Collectors and Health

##### Additional Information
**Made it possible so we can read the value of `pid_max` for linux. After that, the notification calculates active processes percentage based on the given limit.** The limit was added as a parameter to the `active_processes` graph but then hidden so we can use it in the `processes.conf` file _(I hid it so we dont get a line pointing to a static 4 million value f.e)_. On freebsd the limit is static so made it to calculate % based on it

### DIFF
--- a/collectors/proc.plugin/proc_loadavg.c
+++ b/collectors/proc.plugin/proc_loadavg.c
@@ -46,6 +46,19 @@ int do_proc_loadavg(int update_every, usec_t dt) {
 
     //unsigned long long running_processes  = str2ull(procfile_lineword(ff, 0, 3));
     unsigned long long active_processes     = str2ull(procfile_lineword(ff, 0, 4));
+    
+    //open PID_MAX and get the number
+    FILE *fp;
+    int val;
+    char pid_max [TMP_MAX];
+    fp=fopen("/proc/sys/kernel/pid_max", "r");
+    fscanf(fp,"%d",&val);
+    //snprintf(pid_max,TMP_MAX,"%d",val);
+    fclose(fp);
+    //
+
+    unsigned long long max_processes        = (unsigned long long) val;
+   
     //unsigned long long next_pid           = str2ull(procfile_lineword(ff, 0, 5));
 
 
@@ -94,7 +107,8 @@ int do_proc_loadavg(int update_every, usec_t dt) {
 
     if(likely(do_all_processes)) {
         static RRDSET *processes_chart = NULL;
-        static RRDDIM *rd_active = NULL;
+        static RRDDIM *rd_active = NULL, *rd_pidmax;
+        
 
         if(unlikely(!processes_chart)) {
             processes_chart = rrdset_create_localhost(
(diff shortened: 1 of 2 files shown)


Ancairon:
hmmm interesting it says that there is no email on this commit but it seems to be under my account

Ancairon:
it seems that ghost took the commit because I had my email private... made it public now, what can I do in this situation? I have also signed the CLA but it shows unsigned

Ferroin:
> it seems that ghost took the commit because I had my email private... made it public now, what can I do in this situation? I have also signed the CLA but it shows unsigned

If you create a local clone of the repo, I believe you can use the `git change-commits` command to modify the `GIT_AUTHOR_EMAIL` filed on the commit itself. You'll probably have to force-push after doing so though.

Ancairon:
> > it seems that ghost took the commit because I had my email private... made it public now, what can I do in this situation? I have also signed the CLA but it shows unsigned
> 
> If you create a local clone of the repo, I believe you can use the `git change-commits` command to modify the `GIT_AUTHOR_EMAIL` filed on the commit itself. You'll probably have to force-push after doing so though.

ok I will look into that! So you mean I might have to make a new oull request after I modify the email?

collectors/proc.plugin/proc_loadavg.c
```diff
@@ -113,10 +127,13 @@ int do_proc_loadavg(int update_every, usec_t dt) {
             );
 
             rd_active = rrddim_add(processes_chart, "active", NULL, 1, 1, RRD_ALGORITHM_ABSOLUTE);
+            rd_pidmax = rrddim_add(processes_chart, "pidmax", NULL, 1, 1, RRD_ALGORITHM_ABSOLUTE);
         }
         else rrdset_next(processes_chart);
 
         rrddim_set_by_pointer(processes_chart, rd_active, active_processes);
+        rrddim_set_by_pointer(processes_chart, rd_pidmax,    max_processes);
+        rrddim_hide(processes_chart, "pidmax");
```
ilyam8, line 136: could be chart [VARIABLE](https://learn.netdata.cloud/docs/agent/collectors/plugins.d#variable)

i think it is better then a hidden dimension

Ancairon: But then should we implement another notification? Because the `$pidmax` will not be available to calculate on `processes.conf` otherwise

ilyam8: > VARIABLE defines a variable that can be used in alarms

If we dont need history but the current value only VARIABLE is the way to go

Ancairon:
> > it seems that ghost took the commit because I had my email private... made it public now, what can I do in this situation? I have also signed the CLA but it shows unsigned
> 
> If you create a local clone of the repo, I believe you can use the `git change-commits` command to modify the `GIT_AUTHOR_EMAIL` filed on the commit itself. You'll probably have to force-push after doing so though.

I have run `git config --global user.email` and set my email so in the next push probably there should be my email. Now the thing is that with this commit I can't sign the CLA as it doesn't recognise an author. It is my first pull request so I think I will figure it out!

collectors/proc.plugin/proc_loadavg.c
Ancairon: Okay, I will look into it and add it! (should I enter WIP in the title as I'm changing the code again?)

ilyam8: I made it `draft`, click on `Ready for review` button when the PR is done 👍

Ancairon: Perfect, thanks!

Ancairon:
made the change but will create another PR and close this one so I can adress the email issue!

Issue #9840, opened 2020-08-28
Installation script is broken on FreeBSD and macOS


vlvkobal:
##### Bug report summary
There is an error on FreeBSD after #9797 has been merged:
```
root@freebsd-11:/usr/home/vlad/netdata # ./netdata-installer.sh
/tmp/netdata-test.XXXXXXXXXX.apI3qOCP: line 1: -e: command not found
/usr/home/vlad/netdata/netdata-test.XXXXXXXXXX.ZoULjEiw: line 1: -e: command not found

Unable to find a usable temprorary directory. Please set $TMPDIR to a path that is both writable and allows execution of files and try again.
```

On macOS:
```
[vlad@mac-mini netdata]$ sudo ./netdata-installer.sh
/var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/netdata-test.XXXXXXXXXX.2ClUUahT: line 1: -e: command not found
/var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/netdata-test.XXXXXXXXXX.30xEcw02: line 1: -e: command not found

Unable to find a usable temprorary directory. Please set $TMPDIR to a path that is both writable and allows execution of files and try again.

```

##### OS / Environment
```
FreeBSD freebsd-11.3 11.3-RELEASE-p3 FreeBSD 11.3-RELEASE-p3 #0: Mon Aug 19 21:08:43 UTC 2019     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
```
```
Darwin mac-mini 19.6.0 Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64
```

##### Netdata version
netdata v1.24.0-73-g257856e7

##### Component Name
installation script

##### Steps To Reproduce
Run `./netdata-installer.sh` on FreeBSD or macOS.

##### Expected behavior
Netdata can be installed from the sources on FreeBSD and macOS.

Ferroin:
Huh, apparently `/bin/echo` on macOS and FreeBSD is not POSIX compliant...

Ferroin:
No, NVM, I misunderstood POSIX compliance requirements for `/bin/echo` when writing the code...

PR #9842, opened 2020-08-28
Fixed tmpdir handling failure on macOS/FreeBSD.

Ferroin:
##### Summary

When writing the new temporary directory handling code, I forgot that POSIX versions of the `echo` command don't take any options, and as a result ended up breaking the installer and updater for systems that use a POSIX version of `echo`. This fixes the issue by just using `printf` instead.

##### Component Name

area/packaging

##### Test Plan

Relevant code manually tested on macOS and FreeBSD.

##### Additional Information

Fixes: #9840

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -52,7 +52,7 @@ _cannot_use_tmpdir() {
     return "${ret}"
   fi
 
-  if /bin/echo -e '#!/bin/sh\necho SUCCESS\n' > "${testfile}" ; then
+  if printf '#!/bin/sh\necho SUCCESS\n' > "${testfile}" ; then
     if chmod +x "${testfile}" ; then
       if [ "$("${testfile}")" = "SUCCESS" ] ; then
         ret=1
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -130,7 +130,7 @@ _cannot_use_tmpdir() {
     return "${ret}"
   fi
 
-  if /bin/echo -e '#!/bin/sh\necho SUCCESS\n' > "${testfile}" ; then
+  if printf '#!/bin/sh\necho SUCCESS\n' > "${testfile}" ; then
     if chmod +x "${testfile}" ; then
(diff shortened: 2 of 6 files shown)


thiagoftsm:
@Ferroin I tested your PR now, and I verified that it is compiling without any problems with `netdata-installer.sh`, but when I tried to execute the `netdata-update.sh`, I got this error:

VirtualBox_FreeBSD_28_08_2020_14_14_29

I installed Netdata using the following command:

```
./netdata-installer.sh --dont-wait --dont-start-it --install /usr/local/
```

Should I open another issue? Or can you use this PR to fix this error?

Ferroin:
@thiagoftsm How exactly did you go about testing the PR? Was it from a checkout of the PR itself or did you just use the fixed version of `netdata-installer.sh` on a checkout of the master branch?

In the first case, that's an issue with the PR, while in the second it's expected behavior (because the same bug affects the current version of the updater, and this PR fixes that too).

thiagoftsm:
Before to compile your PR, I used the `master` branch to compile and to see the reported error, I was only able to compile on FreeBSD using your PR.

After this I removed the original installation that I had inside `/usr/local/netdata`, and I used this PR. As you can see in the last command I ran, I tried to confirm that the `echo -e` was removed to be sure that an old file was not present.

[netdata-installer.log](https://github.com/netdata/netdata/files/5142704/netdata-installer.log)

I am adding the compilation log to help you.

Ferroin:
Ah, no, I just realized what's wrong, that's not the updater script itself throwing the error, it's the downloaded `netdata-installer.sh` script that's throwing the error. That's expected behavior until this gets merged and goes out in the nightlies.

thiagoftsm (approved):
Considering the [latest comment](https://github.com/netdata/netdata/pull/9842#issuecomment-682658599) I am approving the PR, because it is fixing the error.

I did not test kickstart scripts, but I cannot find the same bug when I run a grep.

vlvkobal (approved):

prologic (approved):

PR #9843, opened 2020-08-28
collect active processes limit feature v2

Ancairon:
##### Summary
Added feature mentioned on #6133 and also applied changes mentioned on #9838 
##### Component Name
Collectors and Health

##### Additional Information
Made it possible so we can read the value of `pid_max` for linux. After that, the notification calculates active processes percentage based on the given limit. The limit was added as a `VARIABLE` to the `active_processes` graph so we can use it in the processes.conf file. On freebsd the limit is static so made it to calculate % based on it

### DIFF
--- a/collectors/proc.plugin/proc_loadavg.c
+++ b/collectors/proc.plugin/proc_loadavg.c
@@ -47,17 +47,9 @@ int do_proc_loadavg(int update_every, usec_t dt) {
     //unsigned long long running_processes  = str2ull(procfile_lineword(ff, 0, 3));
     unsigned long long active_processes     = str2ull(procfile_lineword(ff, 0, 4));
     
-    //open PID_MAX and get the number
-    FILE *fp;
-    int val;
-    char pid_max [TMP_MAX];
-    fp=fopen("/proc/sys/kernel/pid_max", "r");
-    fscanf(fp,"%d",&val);
-    fclose(fp);
+    //get system pid_max
+    unsigned long long max_processes        = get_system_pid_max();
     //
-
-    unsigned long long max_processes        = (unsigned long long) val;
-   
     //unsigned long long next_pid           = str2ull(procfile_lineword(ff, 0, 5));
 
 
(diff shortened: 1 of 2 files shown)


vlvkobal:
Please pay attention to the following points:
1. There is a function you should use: [`get_system_pid_max()`](https://github.com/netdata/netdata/blob/master/libnetdata/os.c#L69-L110).
2. No need to keep a separate alarm for FreeBSD.
3. Choose a better name for the alarm or keep the old one.

Ancairon:
> Please pay attention to the following points:
> 1. There is a function you should use: [`get_system_pid_max()`](https://github.com/netdata/netdata/blob/master/libnetdata/os.c#L69-L110).
> 2. No need to keep a separate alarm for FreeBSD.
> 3. Choose a better name for the alarm or keep the old one.

Perfect, I will look into the function. The alarm for FreeBSD was seperate from before so I didn't know if it was ok to scrap it. Should I remove it? Will also look on the name.

vlvkobal:
> The alarm for FreeBSD was seperate from before so I didn't know if it was ok to scrap it. Should I remove it?

Yes, I think you should remove it.

Ancairon:
> Yes, I think you should remove it.

Pushed another commit,
1) it utilizes the function `get_system_pid_max()`
2) it uses only one alarm 
3) it has a simpler name very close to the original, just the original was "**number** of active processes" , I changed it to "**percentage** of active processes"

vlvkobal (approved):

vlvkobal:
@ilyam8 pointed out that
```
    warn: $this > (($status >= $WARNING)  ? (75) : (80))
    crit: $this > (($status == $CRITICAL) ? (85) : (90))
```
prevents frequent status flipping when the variable value is around 80 or 90. Why have you removed the conditions?

Ancairon:
> @ilyam8 pointed out that
> ```
>     warn: $this > (($status >= $WARNING)  ? (75) : (80))
>     crit: $this > (($status == $CRITICAL) ? (85) : (90))
> ```
> prevents frequent status flipping when the variable value is around 80 or 90. Why have you removed the conditions?


You are right, I need to add them back in.

ilyam8:
@Ancairon Have tested the alarm? Does it work and you see it in the `alarms` section? i suppose so but you didnt mention it explicitly.

Ancairon:
> @Ancairon Have tested the alarm? Does it work and you see it in the `alarms` section? i suppose so but you didnt mention it explicitly.

Yes i have tested the alarm and it shows correctly in the alarm section, my mistake not mentioning I tested it, will do in future instances.

ilyam8 (approved):

Ancairon:
Perfect, thanks to everyone who helped and advised me through my first pull request!

vlvkobal (approved):

Issue #9605, opened 2020-07-25
CentOS 7 - No package Judy-devel available.


as1988:
#####Question summary

Hi, 
When I attempt to install via the kickstart method I get this error:

```
Press ENTER to run it >
yum install Judy-devel
Loaded plugins: fastestmirror, universal-hooks
Loading mirror speeds from cached hostfile
 * EA4: 91.197.228.252
 * cpanel-addons-production-feed: 91.197.228.252
 * cpanel-plugins: 91.197.228.252
 * base: mirrors.clouvider.net
 * extras: uk.mirror.nsec.pt
 * updates: mirror.sov.uk.goscomb.net
No package Judy-devel available.
Error: Nothing to do



We are very sorry!

Installation of required packages failed.

What to do now:

  1. Make sure your system is updated.
     Most of the times, updating your system will resolve the issue.

  2. If the error message is about a specific package, try removing
     that package from the command and run it again.
     Depending on the broken package, you may be able to continue.

  3. Let us know. We may be able to help.
     Open a github issue with the above log, at:

           https://github.com/netdata/netdata/issues


 FAILED

 WARNING  It failed to install all the required packages, but installation might still be possible.
```

It would appear the Judy is not available for install on my version of CentOS 7? How do I get around this?

Thank you!

##### OS / Environment
CentOS Linux release 7.8.2003 (Core) - 3.10.0-1062.9.1.el7.x86_64
cPanel v88.0.13

Ferroin:
You appear to not have the EPEL repo enabled, which is required for installation to work correctly. The installation process _should_ try to install this itself, so you most likely said no to the prompt asking about installing it.

as1988:
> You appear to not have the EPEL repo enabled, which is required for installation to work correctly. The installation process _should_ try to install this itself, so you most likely said no to the prompt asking about installing it.

It’s definitely installed, or at least I’ve ran the command for it to be installed and I’ve not said no to it being installed either. I presumably haven’t been asked as it too recognised that it is installed. Any ideas on how to resolve?

prologic:
There are open issues with Judy and Judy-devel currently. I'm working on bundling the Judy development libraries into our source tree. @Ferroin Perhaps you can assist me with this?

rdifrancesco:
@prologic Could you elaborate on what those issues are with Judy?  Could any of them be related to agents unable to connect to Netdata Cloud?

as1988:
Hi
I have managed to resolve this issue.
[...]

Issue #9356, opened 2020-06-17
Judy-devel installation error


DPR1604:
##### Bug report summary

When Attempting to install netdata cloud agent I'm currently getting errors installing Judy-devel when using the kickstart script.

I have narrowed the issue down to this output 

```
yum install http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm
Extra OKay Packages for Enterprise Linux - x86_64                                                                                                                                                             1.7 MB/s | 3.7 MB     00:02
Last metadata expiration check: 0:00:02 ago on Wed Jun 17 14:05:33 2020.
[MIRROR] Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm: Status code: 404 for http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm (IP: 109.228.54.129)
[MIRROR] Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm: Status code: 404 for http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm (IP: 109.228.54.129)
[MIRROR] Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm: Status code: 404 for http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm (IP: 109.228.54.129)
[MIRROR] Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm: Status code: 404 for http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm (IP: 109.228.54.129)
[FAILED] Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm: Status code: 404 for http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.1.0+217+4d875839.x86_64.rpm (IP: 109.228.54.129)
[...]

Issue #9267, opened 2020-06-04
libJudy cmake detection error in Fedora 31


underhood:
##### Bug report summary
![Screenshot from 2020-06-04 10-19-28](https://user-images.githubusercontent.com/6674623/83732875-fd8fb380-a64c-11ea-905b-dc6bc32d98db.png)

##### OS / Environment
FC31 AMD64

##### Component Name
cmake

Ferroin:
This seems to be solved by #9776 and #9859.

underhood:
@prologic this is error if we try to use cmake build system instead of default autotools and was not fixed by judy bundling

prologic:
@Ferroin Can you comment on this?

prologic:
@underhood I'm a bit out of context, sorry. Can you elaborate on your previous comment?

underhood:
@prologic the cmake doesn't detect libjudy-devel is installed on Fedora 31. This is not related to judy bundling. See following(latest master):

![Screenshot from 2020-09-14 15-26-13](https://user-images.githubusercontent.com/6674623/93091652-aadb8400-f69e-11ea-8115-a638b7d71765.png)

prologic:
> @prologic the cmake doesn't detect libjudy-devel is installed on Fedora 31. This is not related to judy bundling. autotools corectly detect judy is installed but cmake doesn't. See following(latest master):

Thanks! Do you have a proposed fix? Otherwise @Ferroin and I will look at this.

underhood:
@prologic this is low prio as it is not user-facing (all our build methods use autotools). I think there was a plan to move over to cmake eventually (and again I guess the reason is future plan to support MS Doors OS).

For these reasons I didn't investigate much into this. I suspect maybe a way how fedora packages libjudy makes it not detected by cmake but it is a guess.

prologic:
Thanks @netdata/cloud Let's ice this then.

wangpei-nice:
I have the same issue on debian10 when exec cmake cmd.  Actually, I've already installed libjudy-dev.
![image](https://user-images.githubusercontent.com/81410079/115003604-8c5d9580-9ed8-11eb-8244-3bba17bc3170.png)

ilyam8:
Closing because Fedora 31 is EOL.

cc @underhood

PR #9776, opened 2020-08-18
Added code to bundle libJudy on systems which do not provide a usable copy of it.

Ferroin:
##### Summary

This should 'fix' most of the issues we've had with availability of libJudy by sidestepping them completely.

This adds one new build time dependency (libtool).

##### Component Name

area/packaging

##### Test Plan

Basic build code for Judy itself tested independently from the script on numerous distros.

Actual Netdata build tested using `netdata-installer.sh` in a Docker container and then running the unit tests (which include stress testing of the dbengine).

##### Additional Information

Fixes: #8153 

This is a first step in fixing #9356 and #9605, I'll open a followup PR to stop trying to install libJudy on systems that are known not to have it once this one is merged, which will completely fix those two issues.

Relevant to but not a fix for: #8602

PR #9859, opened 2020-09-01
Removed dependency on libJudy for systems which don't have it.

Ferroin:
##### Summary

This updates how we handle libJudy in our dependency installation script. Instead of trying to install it everywhere and hoping it works, we now only install it on distros that we know both a system copy available that actually works.

##### Component Name

area/packaging

##### Test Plan

Verified that this behaves correctly in our CI.

##### Additional Information

Followup to: #9776 
Fixes: #9267 
Fixes: #9356 
Fixes: #9605 

This functionally adds CI for #9776 because we test on platforms that will now not have libJudy installed in the testing environment.

### DIFF
--- a/packaging/installer/install-required-packages.sh
+++ b/packaging/installer/install-required-packages.sh
@@ -1125,16 +1125,13 @@ declare -A pkg_openssl=(
 )
 
 declare -A pkg_judy=(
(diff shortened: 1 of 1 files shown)


prologic:
🎉🙇‍♂️

prologic (approved):

Issue #9827, opened 2020-08-27
Netdata still sends notifications to telegram, but notification is set to slient


joshuabeny1999:
##### Bug report summary

I enabled Netdata on my servers and disabled the disk_space time notification. Due it will fire every night for my backup disk.
![image](https://user-images.githubusercontent.com/55023662/91394672-fa1a5b80-e83a-11ea-8f69-2652ac2be56f.png)

But I still get the notifications via Telegram:
![image](https://user-images.githubusercontent.com/55023662/91396029-6006e300-e83b-11ea-9342-2d14be29dd25.png)

##### OS / Environment

```
joshua@joshuah:~$ uname -a; grep -Hv "^#" /etc/*release
Linux joshuah 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
/etc/lsb-release:DISTRIB_ID=Ubuntu
/etc/lsb-release:DISTRIB_RELEASE=20.04
/etc/lsb-release:DISTRIB_CODENAME=focal
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="20.04.1 LTS (Focal Fossa)"
/etc/os-release:ID=ubuntu
/etc/os-release:ID_LIKE=debian
/etc/os-release:PRETTY_NAME="Ubuntu 20.04.1 LTS"
/etc/os-release:VERSION_ID="20.04"
/etc/os-release:HOME_URL="https://www.ubuntu.com/"
/etc/os-release:SUPPORT_URL="https://help.ubuntu.com/"
/etc/os-release:BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
/etc/os-release:PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
/etc/os-release:VERSION_CODENAME=focal
/etc/os-release:UBUNTU_CODENAME=focal
```

##### Netdata version

```
joshua@joshuah:~$ $(ps aux | grep -E -o "[a-zA-Z/]+netdata ") -V
netdata v1.24.0
```
##### Component Name
/health/notifications

##### Steps To Reproduce

1. Setup netdata with kickstart.sh
2. Use `sudo /etc/netdata/edit-config health_alarm_notify.conf`and add Settings for telegram
[...]

PR #9869, opened 2020-09-02
Change instruction to reload HEALTH

thiagoftsm:
##### Summary
@amoss  wrote something that we missed when we brought [netdatacli](https://github.com/netdata/netdata/issues/9827#issuecomment-685421668), so I am bringing this PR to update our documentation related to health.

##### Component Name
Doc

### DIFF
--- a/docs/guides/monitor/stop-notifications-alarms.md
+++ b/docs/guides/monitor/stop-notifications-alarms.md
@@ -72,8 +72,8 @@ To silence this alarm, change `sysadmin` to `silent`.
  to: silent
 ```
 
-Use `netdatacli reload-health` to reload your health configuration and ensure you get no more notifications about that
-alarm.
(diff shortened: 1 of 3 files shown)


ilyam8:
`netdatacli` is not always in the `PATH`. Using `killall` is quite handy in that case.

amoss:
Do we know the absolute path of netdatacli when we generate notifications?

thiagoftsm:
@ilyam8 I talked with @joelhans and we are now centralizing the information to reload Health and creating links to it.

Yes @amoss , we always know its position, because it is inserted at `${NETDATA_PREFIX}/usr/bin`, but considering that this is documentation, @joelhans and I moved in a direction to show both options centralized to simplify user's life when they are looking for some information.

amoss (approved):

PR #9871, opened 2020-09-02
python.d/dnsdist: fix latency-avg chart units

scottymuse:
The latency-avg values are in microseconds, not milliseconds.
https://dnsdist.org/statistics.html#latency-avg100

##### Summary
The latency average graph is currently labeled with ms/query (milliseconds), but it should be us/query (microseconds). I got a little panicy when I was reading that my DNS queries were taking 5 seconds on average. 

##### Component Name
dnsdist collector

### DIFF
--- a/collectors/python.d.plugin/dnsdist/dnsdist.chart.py
+++ b/collectors/python.d.plugin/dnsdist/dnsdist.chart.py
@@ -105,7 +105,7 @@ CHARTS = {
         ]
     },
     'query_latency_avg': {
-        'options': [None, 'Average latency for the last N queries', 'us/query', 'latency',
+        'options': [None, 'Average latency for the last N queries', 'microseconds', 'latency',
                     'dnsdist.query_latency_avg', 'line'],
         'lines': [
             ['latency-avg100', '100', 'absolute'],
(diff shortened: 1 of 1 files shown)


collectors/python.d.plugin/dnsdist/dnsdist.chart.py
```diff
@@ -105,7 +105,7 @@
         ]
     },
     'query_latency_avg': {
-        'options': [None, 'Average latency for the last N queries', 'ms/query', 'latency',
+        'options': [None, 'Average latency for the last N queries', 'us/query', 'latency',
```
ilyam8, line 108: i think just `microsecond` or `us` will do, lets remove `/query` suffix. What do you think?

scottymuse: I like `microseconds` or even `useconds`. The `/query` probably isn't necessary, and just `us` could be confusing.

ilyam8: lest make `microseconds`

scottymuse: Done. Thank you!

ilyam8 (approved):

thiagoftsm (approved):
LGTM!

ilyam8:
@scottymuse thx for the contribution!

PR #9888, opened 2020-09-07
Fix lock order reversal. (Coverity defect CID 361629)

mfundul:
##### Summary

Fix the following coverity defect:

** CID 361629:  Program hangs  (ORDER_REVERSAL)

```
*** CID 361629:  Program hangs  (ORDER_REVERSAL)
/streaming/receiver.c: 464 in rrdpush_receive()
458                 rpt->host->senders_disconnected_time = now_realtime_sec();
459                 rrdhost_flag_set(rpt->host, RRDHOST_FLAG_ORPHAN);
460                 if(health_enabled == CONFIG_BOOLEAN_AUTO)
461                     rpt->host->health_enabled = 0;
462             }
463             if (rpt->host->receiver == rpt) {
>>>     CID 361629:  Program hangs  (ORDER_REVERSAL)
>>>     Calling "rrdpush_sender_thread_stop" acquires lock "sender_state.mutex" while holding lock "rrdhost.rrdhost_rwlock" (count: 1 / 2).
464                 rrdpush_sender_thread_stop(rpt->host);
465             }
466             netdata_mutex_unlock(&rpt->host->receiver_lock);
467             rrdhost_unlock(rpt->host);
468             rrd_unlock();
469         }
```

##### Component Name
Streaming
##### Test Plan
Start and Stop 1000 child nodes in a streaming set-up. Restart the agent.

### DIFF
--- a/streaming/receiver.c
+++ b/streaming/receiver.c
@@ -460,11 +460,11 @@ static int rrdpush_receive(struct receiver_state *rpt)
             if(health_enabled == CONFIG_BOOLEAN_AUTO)
                 rpt->host->health_enabled = 0;
         }
+        rrdhost_unlock(rpt->host);
         if (rpt->host->receiver == rpt) {
             rrdpush_sender_thread_stop(rpt->host);
         }
         netdata_mutex_unlock(&rpt->host->receiver_lock);
-        rrdhost_unlock(rpt->host);
         rrd_unlock();
     }
 


underhood (approved):

amoss (approved):
I think this is the correct fix. The two options are:
1. Changing the other locking points to hold the host lock
2. Removing the host lock from the call to match the other code.

In the sender thread main-loop the mutex is locked without the host lock for performance reasons (so that we do not get unpredictable delays between the TCP window opening and the data being sent). Adding the host lock at that point would cause a regression in the code.

underhood (approved):

PR #9904, opened 2020-09-09
Fixed another typo in the libJudy bundling code.

Ferroin:
##### Summary

SSIA

##### Component Name

area/packaging

##### Test Plan

Minimally verified locally.

##### Additional Information

This should properly fix the issues we thought were just a result of the misspelled option name.


amoss (approved):

amoss (changes requested):
I can't see the configure log in the CI/CD, but there is still a case-issue on some distros but not others, looking at the alpine 3.11 build:
```
checking for libJudy in externaldeps/libJudy... yes
```
the test program worked
```
./database/engine/rrdengine.h:11:10: fatal error: Judy.h: No such file or directory
   11 | #include <Judy.h>
      |  
```
The include path should be set to externaldeps/libJudy so I don't see how this could fail...
```
[/netdata]# mkdir -p /netdata/externaldeps/libJudy 
 OK   
[/netdata]# cp /tmp/netdata-judy-PBEfaP/libjudy-1.0.5-netdata1/src/libJudy.a /netdata/externaldeps/libJudy/libJudy.a 
 OK   
[/netdata]# cp /tmp/netdata-judy-PBEfaP/libjudy-1.0.5-netdata1/src/Judy.h /netdata/externaldeps/libJudy/Judy.h
```
Looks correct but the build is working on some distros but failing on others.

Ferroin:
OK, tracked down the issue, we're not actually setting any CFLAGS in the configure script for libJudy. I _think_ I can fix this myself, but I'm not sure.

mfundul:
> 
> 
> OK, tracked down the issue, we're not actually setting any CFLAGS in the configure script for libJudy. I _think_ I can fix this myself, but I'm not sure.

This:
https://github.com/netdata/netdata/blob/v1.24.0/configure.ac#L331
used to work so maybe it's simple.

Ferroin:
> > OK, tracked down the issue, we're not actually setting any CFLAGS in the configure script for libJudy. I _think_ I can fix this myself, but I'm not sure.
> 
> This:
> https://github.com/netdata/netdata/blob/v1.24.0/configure.ac#L331
> used to work so maybe it's simple.

It worked because we weren't previously using custom prefixes that aren't in the system include paths. The fix in this case appears to just be a 2-line change in the configure script so that we persist the `OPTIONAL_JUDY_CFLAGS` variable we're using in the script itself into the build environment and then have it included in the CFLAGS used in the Makefile. Testing this now and should hopefully have this PR updated with that fix shortly if the testing works.

prologic (approved):

prologic:
@mfundul @amoss Please re-review. Thanks! 🙇‍♂️

amoss (approved):

mfundul (approved):

PR #9908, opened 2020-09-10
dashboard v1.5.2

jacekkolasa:
##### Summary
New Dashboard version.

The only new change related to Agent Dashboard is https://github.com/netdata/dashboard/pull/124:
- add new social media links section below right hand sidebar
- move "Privacy policy" & "Terms and conditions" links from sidebar to the bottom of the dashboard
- leave only "Copyright 2020, Netdata, Inc" in Copyright section in the bottom of the dashboard


ilyam8:
I see the changes, i think the style a little bit strange, do we deliberately use that background to highlight social media links?

Screenshot 2020-09-10 at 15 31 09

thiagoftsm:
> I see the changes, i think the style a little bit strange, do we deliberately use that background to highlight social media links?
> 

I agree with all @ilyam8 words and also I have doubts:
1 -  why are we not adding all social network listed [here](https://github.com/netdata/netdata/#community)? 
2 - I think the second sentence `Give us a start?` should not have the interrogation. Am I correct @joelhans ?

jacekkolasa:
Yes, that's deliberate. The design looks like this: 
image

I agree it can be a little distracting, but maybe that's only because we are so used to the same sidebar? I'm not sure. @manos-saratsis what do you think?

ilyam8:
@jacekkolasa i see you the angles are roundish on your screenshot which looks better, indeed.

joelhans:
@thiagoftsm I don't think this is really my call but rather product's, but yes, I would end that sentence with an exclamation point or a period.

Give us a star!
Give us a star.

prologic (approved):

manos-saratsis:
Give us a star! It was a mistake on the design. Also we should make the Github star a bit more prominent,

jacekkolasa:
closing, will be replaced by newer version

PR #9911, opened 2020-09-10
Add HTTP and HTTPS support to the simple exporting connector

### LINKED ISSUE #9189: Reduce the number of HTTP headers in messages sent by OpenTSDB exporting connector
##### Feature idea summary
In order to avoid excessive buffer copying we attach HTTP headers to every metric in the OpenTSDB exporting connector. We can reduce the number of headers.

We can use connector specific data and add a message_header variable to the connector specific data for the simple connector and send it before the buffer with metrics if it's not NULL.

##### Expected behavior
HTTP header is sent just once for a batch.

### LINKED ISSUE #9607: Exporting module: differing behavior vs 'backend'
I'm unsure if this is related to #9512, but throwing it here as a more generic problem report.

##### Bug report summary

The exporting module has different connection behavior than the now deprecated backend environment.

The 'backend' connects to the destination when it has data queued to send, and upon a successful connection, immediately sends it.  In contrast, the exporting module connects to the destination, and sends data at the next "update every" interval.  The end result being that destinations that close the TCP connection between samples can lose a sample in-between updates.

This feels like surprising behavior, if the exporting module is intended as a drop-in replacement for backend.

##### OS / Environment

```
> uname -a; uname -K
FreeBSD dev 12.1-RELEASE FreeBSD 12.1-RELEASE r354233 GENERIC  amd64
1201000
```

##### Netdata version

```
> netdata -V
netdata v1.23.1
```

##### Component Name

exporting.

##### Steps To Reproduce

Minimal backend configuration:

```
[backend]
    hostname = dev
    enabled = yes
    data source = average
    type = json
    destination = localhost:2222
    prefix = netdata
    update every = 5
    buffer on failures = 10
    send charts matching = system.*
```

Minmal exporting configuration:

```
[exporting:global]
        enabled = yes

[json:netcat]
        hostname = dev
        enabled = yes
        data source = average
        destination = localhost:2223
        prefix = netdata
        update every = 5
        buffer on failures = 10
        send charts matching = system.*
 ```

Start a netcat listener on 2222, and one on 2223.   Observe the difference.

```
> nc -vkl 2222
Connection from localhost 62669 received!
{"prefix":"netdata","hostname":"dev",.....    <--- this is instantaneous after connection
```

```
> nc -vkl 2223
Connection from localhost 62878 received!
[ 5 second pause ]
{"prefix":"netdata","hostname":"dev",.....    <--- this is after the 'update every' pause
```

##### Expected behavior

I have many, many netdata clients all funneling their samples to a single destination, which is a traditional forking server.  To conserve resources (1000s of forked processes and database connections) on this destination, the TCP socket is closed between sends, letting netdata just merrily reconnect on its next interval.  Clearly, this is a tradeoff at the expense of establishling TCP connections for each sample, but for this environment that's an easy trade to make.

My expectation was that behavior would be identical between backend and exporting.   This appears to not be the case with both JSON and OpenTSDB, I didn't try with others.

Thanks all!

### LINKED ISSUE #9631: Secure OpenTSDB backend does not work with InfluxDb
##### Bug report summary
When turning on SSL for the opentsdb backend in the latest v23 netdata, a few measurements are created in the corresponding InfluxDB database, but only a few measurement points are written to it. No error messages on either netdata, or influxdb side.

This issue is already reported on GitHub: OpenTSDB and TLS #9068.
A test was performed, but the number of measurements, and the points therein were not checked.

We set up the latest netdata in a Kubernetes cluster with the helm chart provided at https://github.com/netdata/helmchart.
Configured the backend with 

```cmd
[backend]
          enabled = yes
          type = opentsdb:https
          destination = influxdb:4242
          send hosts matching = system. *  
```

We set up the influxdb with the following helm chart: https://influxdata.github.io/helm-charts
Created a self-signed certificate, mounted it, and turned on the opentsdb interface:

```yaml
  opentsdb: 
    enabled : true
    bind-address : ":4242"
    database : "netdata"
    tls-enabled : true
    certificate : "/etc/ssl/keycrt/influxdb.pem" 
```

##### OS / Environment

```cmd
Environment
Kubernetes and OS versions:
Master and Woker node OS: Ubuntu 18.04.4 LTS
kubeadm: v1.18.2
kubectl:
client: v1.18.2
server: v1.18.2
cni:
calico: v3.11
metrics-server: k8s.gcr.io/metrics-server-amd64:v0.3.6 (image)
kubernetes-dashboard: kubernetesui/dashboard:v2.0.0-rc6 (image)
```

##### Netdata version
v23
 

##### Component Name
backend

##### Steps To Reproduce
We set up the latest netdata in a Kubernetes cluster with the helm chart provided at https://github.com/netdata/helmchart.
Configured the backend with 

```cmd
[backend]
          enabled = yes
          type = opentsdb:https
          destination = influxdb:4242
          send hosts matching = system. *  
```

We set up the influxdb with the following helm chart: https://influxdata.github.io/helm-charts
Created a self-signed certificate, mounted it, and turned on the opentsdb interface:

```yaml
  opentsdb: 
    enabled : true
    bind-address : ":4242"
    database : "netdata"
    tls-enabled : true
    certificate : "/etc/ssl/keycrt/influxdb.pem" 
```

##### Expected behavior
All the measurements should be there, being continuously filled with data points. Only a subset of the collected metrics have their respective measurements, and only a few points are stored there, those that are collected right after the start.

### LINKED ISSUE #10057: Backend "send charts matching" doesn't filter out internal Netdata monitoring
##### Bug report summary

I'm trying to export Netdata metrics to prometheus. I'm interested in keeping Netdata internal monitoring enabled, but I don't want it to be exported to prometheus. I used the `send charts matching` key in `[backend]` to filter it out, but this config is ignored.

##### OS / Environment

Arch linux
`Linux ... 5.8.13-arch1-1 #1 SMP PREEMPT Thu, 01 Oct 2020 20:40:35 +0000 x86_64 GNU/Linux`

Netdata installed through `pacman`

##### Netdata version

`netdata v1.25.0`

##### Component Name

Not sure, `exporting` I guess ?

##### Steps To Reproduce

1. Install Netdata with pacman.
2. Edit netdata conf to set 
```
[backend]
        # host tags = 
        # enabled = no
        # data source = average
        # type = graphite
        # destination = localhost
        # prefix = netdata
        # hostname = ...
        # update every = 10
        # buffer on failures = 10
        # timeout ms = 20000
        # send names instead of ids = yes
        send charts matching = * !ipv6* !*tun0* !users* !netdata*
        # send hosts matching = localhost *
        # remote write URL path = /receive
```
3. Restart Netdata with systemd
4. Wait few seconds, and check if internal monitoring is sent: 
```
$ curl -s 'http://localhost:19999/api/v1/allmetrics?format=prometheus&help=no&source=average' | grep -o 'netdata_netdata_dbengine[a-zA-Z_]*'
netdata_netdata_dbengine_ram_MiB_average
netdata_netdata_dbengine_ram_MiB_average
netdata_netdata_dbengine_ram_MiB_average
netdata_netdata_dbengine_global_file_descriptors_descriptors_average
netdata_netdata_dbengine_global_file_descriptors_descriptors_average
netdata_netdata_dbengine_global_errors_errors_persec_average
netdata_netdata_dbengine_global_errors_errors_persec_average
netdata_netdata_dbengine_global_errors_errors_persec_average
netdata_netdata_dbengine_io_operations_operations_persec_average
netdata_netdata_dbengine_io_operations_operations_persec_average
netdata_netdata_dbengine_io_throughput_MiB_persec_average
netdata_netdata_dbengine_io_throughput_MiB_persec_average
netdata_netdata_dbengine_long_term_page_stats_pages_average
netdata_netdata_dbengine_long_term_page_stats_pages_average
netdata_netdata_dbengine_long_term_page_stats_pages_average
netdata_netdata_dbengine_long_term_page_stats_pages_average
netdata_netdata_dbengine_compression_ratio_percentage_average
```

##### Expected behavior

I believe the `send charts matching` key does not respect my `!netdata*`, as I expected to not have any Netdata internal monitoring to be exported

vlvkobal:
##### Summary
There is a group of exporting connectors that use the same base - the simple connector:
- Graphite
- JSON
- OpenTSDB
- Prometheus Remote Write

We are adding HTTP and HTTPS support to this group and fixing a bunch of bugs along the way.

- [x] Use a ring buffer for the simple connector
- [x] Support for HTTP protocol in the Graphite and JSON connectors
- [x] Support for HTTPS in the Graphite, JSON, and Prometheus Remote Write connectors
- [x] Send HTTP headers once per batch
- [x] Use a correct JSON list format
- [x] Compact OpenTSDB HTTP output
- [x] Use functions from the exporting engine for the allmetrics API call with Prometheus format

Fixes #9607
Fixes #9631
Closes #9189
Fixes #10057

##### Component Name
exporting engine

##### Test Plan
Test the exporting engine with every protocol (plain, http, https) for the Graphite, JSON, OpenTSDB, and Prometheus Remote Write connectors.

Check the output of `allmetrics?format=prometheus` API request.

### DIFF
--- a/exporting/clean_connectors.c
+++ b/exporting/clean_connectors.c
@@ -35,3 +35,46 @@ void clean_instance(struct instance *instance)
     uv_cond_destroy(&instance->cond_var);
     // uv_mutex_destroy(&instance->mutex);
 }
+
+/**
+ * Clean up a simple connector instance on Netdata exit
+ *
+ * @param instance an instance data structure.
+ */
+void simple_connector_cleanup(struct instance *instance)
+{
+    info("EXPORTING: cleaning up instance %s ...", instance->config.name);
+
+    struct simple_connector_data *simple_connector_data =
+        (struct simple_connector_data *)instance->connector_specific_data;
+
+    buffer_free(instance->buffer);
+
+    struct simple_connector_buffer *next_buffer = simple_connector_data->first_buffer;
+    for (int i = 0; i < instance->config.buffer_on_failures; i++) {
+        struct simple_connector_buffer *current_buffer = next_buffer;
+        next_buffer = next_buffer->next;
+
+        if (current_buffer->header)
+            buffer_free(current_buffer->header);
+        if (current_buffer->buffer)
+            buffer_free(current_buffer->buffer);
+        freez(current_buffer);
+    }
+
+#ifdef ENABLE_HTTPS
+    if (simple_connector_data->conn)
+        SSL_free(simple_connector_data->conn);
+#endif
+
+    freez(simple_connector_data);
+
+    struct simple_connector_config *simple_connector_config =
+        (struct simple_connector_config *)instance->config.connector_specific_config;
+    freez(simple_connector_config);
+
+    info("EXPORTING: instance %s exited", instance->config.name);
+    instance->exited = 1;
+
+    return;
+}
--- a/exporting/clean_connectors.c
+++ b/exporting/clean_connectors.c
@@ -49,16 +49,16 @@ void simple_connector_cleanup(struct instance *instance)
         (struct simple_connector_data *)instance->connector_specific_data;
 
     buffer_free(instance->buffer);
+    buffer_free(simple_connector_data->buffer);
+    buffer_free(simple_connector_data->header);
 
     struct simple_connector_buffer *next_buffer = simple_connector_data->first_buffer;
     for (int i = 0; i < instance->config.buffer_on_failures; i++) {
         struct simple_connector_buffer *current_buffer = next_buffer;
         next_buffer = next_buffer->next;
 
-        if (current_buffer->header)
-            buffer_free(current_buffer->header);
-        if (current_buffer->buffer)
-            buffer_free(current_buffer->buffer);
+        buffer_free(current_buffer->header);
+        buffer_free(current_buffer->buffer);
         freez(current_buffer);
     }
 
--- a/exporting/graphite/graphite.c
+++ b/exporting/graphite/graphite.c
@@ -17,6 +17,8 @@ int init_graphite_instance(struct instance *instance)
     connector_specific_config->default_port = 2003;
 
     struct simple_connector_data *connector_specific_data = callocz(1, sizeof(struct simple_connector_data));
+    instance->connector_specific_data = connector_specific_data;
+
 #ifdef ENABLE_HTTPS
     connector_specific_data->flags = NETDATA_SSL_START;
     connector_specific_data->conn = NULL;
@@ -24,7 +26,6 @@ int init_graphite_instance(struct instance *instance)
         security_start_ssl(NETDATA_SSL_CONTEXT_EXPORTING);
     }
 #endif
-    instance->connector_specific_data = connector_specific_data;
 
     instance->start_batch_formatting = NULL;
     instance->start_host_formatting = format_host_labels_graphite_plaintext;
--- a/exporting/graphite/graphite.c
+++ b/exporting/graphite/graphite.c
@@ -29,7 +29,7 @@ int init_graphite_instance(struct instance *instance)
     instance->end_host_formatting = flush_host_labels;
     instance->end_batch_formatting = simple_connector_update_buffered_bytes;
 
-    instance->send_header = NULL;
+    instance->prepare_header = NULL;
     instance->check_response = exporting_discard_response;
 
     instance->buffer = (void *)buffer_create(0);
--- a/exporting/graphite/graphite.c
+++ b/exporting/graphite/graphite.c
@@ -40,7 +40,11 @@ int init_graphite_instance(struct instance *instance)
     instance->end_host_formatting = flush_host_labels;
     instance->end_batch_formatting = simple_connector_end_batch;
 
-    instance->prepare_header = NULL;
+    if (instance->config.type == EXPORTING_CONNECTOR_TYPE_GRAPHITE_HTTP)
+        instance->prepare_header = graphite_http_prepare_header;
+    else
+        instance->prepare_header = NULL;
+
     instance->check_response = exporting_discard_response;
 
     instance->buffer = (void *)buffer_create(0);
@@ -200,4 +204,27 @@ int format_dimension_stored_graphite_plaintext(struct instance *instance, RRDDIM
         (unsigned long long)last_t);
 
     return 0;
+}
+
+/**
+ * Ppepare HTTP header
+ *
+ * @param instance an instance data structure.
+ * @return Returns 0 on success, 1 on failure.
+ */
+void graphite_http_prepare_header(struct instance *instance)
+{
+    struct simple_connector_data *simple_connector_data = instance->connector_specific_data;
+
+    buffer_sprintf(
+        simple_connector_data->last_buffer->header,
+        "POST /api/put HTTP/1.1\r\n"
+        "Host: %s\r\n"
+        "Content-Type: application/graphite\r\n"
+        "Content-Length: %lu\r\n"
+        "\r\n",
+        instance->config.destination,
+        buffer_strlen(simple_connector_data->last_buffer->buffer));
+
+    return;
 }
--- a/exporting/graphite/graphite.c
+++ b/exporting/graphite/graphite.c
@@ -37,6 +37,9 @@ int init_graphite_instance(struct instance *instance)
         error("EXPORTING: cannot create buffer for graphite exporting connector instance %s", instance->config.name);
         return 1;
     }
+
+    simple_connector_init(instance);
+
     if (uv_mutex_init(&instance->mutex))
         return 1;
     if (uv_cond_init(&instance->cond_var))
--- a/exporting/graphite/graphite.c
+++ b/exporting/graphite/graphite.c
@@ -27,7 +27,7 @@ int init_graphite_instance(struct instance *instance)
 
     instance->end_chart_formatting = NULL;
     instance->end_host_formatting = flush_host_labels;
-    instance->end_batch_formatting = simple_connector_update_buffered_bytes;
+    instance->end_batch_formatting = simple_connector_end_batch;
 
     instance->prepare_header = NULL;
     instance->check_response = exporting_discard_response;
--- a/exporting/graphite/graphite.c
+++ b/exporting/graphite/graphite.c
@@ -16,6 +16,16 @@ int init_graphite_instance(struct instance *instance)
     instance->config.connector_specific_config = (void *)connector_specific_config;
     connector_specific_config->default_port = 2003;
 
+    struct simple_connector_data *connector_specific_data = callocz(1, sizeof(struct simple_connector_data));
+#ifdef ENABLE_HTTPS
+    connector_specific_data->flags = NETDATA_SSL_START;
+    connector_specific_data->conn = NULL;
+    if (instance->config.options & EXPORTING_OPTION_USE_TLS) {
+        security_start_ssl(NETDATA_SSL_CONTEXT_OPENTSDB);
+    }
+#endif
+    instance->connector_specific_data = connector_specific_data;
+
     instance->start_batch_formatting = NULL;
     instance->start_host_formatting = format_host_labels_graphite_plaintext;
     instance->start_chart_formatting = NULL;
--- a/exporting/graphite/graphite.c
+++ b/exporting/graphite/graphite.c
@@ -21,7 +21,7 @@ int init_graphite_instance(struct instance *instance)
     connector_specific_data->flags = NETDATA_SSL_START;
     connector_specific_data->conn = NULL;
     if (instance->config.options & EXPORTING_OPTION_USE_TLS) {
-        security_start_ssl(NETDATA_SSL_CONTEXT_OPENTSDB);
+        security_start_ssl(NETDATA_SSL_CONTEXT_EXPORTING);
     }
 #endif
     instance->connector_specific_data = connector_specific_data;
--- a/exporting/json/json.c
+++ b/exporting/json/json.c
@@ -17,13 +17,6 @@ int init_json_instance(struct instance *instance)
     connector_specific_config->default_port = 5448;
 
     struct simple_connector_data *connector_specific_data = callocz(1, sizeof(struct simple_connector_data));
-#ifdef ENABLE_HTTPS
-    connector_specific_data->flags = NETDATA_SSL_START;
-    connector_specific_data->conn = NULL;
-    if (instance->config.options & EXPORTING_OPTION_USE_TLS) {
-        security_start_ssl(NETDATA_SSL_CONTEXT_EXPORTING);
-    }
-#endif
     instance->connector_specific_data = connector_specific_data;
 
     instance->start_batch_formatting = NULL;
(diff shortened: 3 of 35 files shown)


exporting/graphite/graphite.c
```diff
@@ -27,16 +38,23 @@ int init_graphite_instance(struct instance *instance)
 
     instance->end_chart_formatting = NULL;
     instance->end_host_formatting = flush_host_labels;
-    instance->end_batch_formatting = simple_connector_update_buffered_bytes;
+    instance->end_batch_formatting = simple_connector_end_batch;
+
+    if (instance->config.type == EXPORTING_CONNECTOR_TYPE_GRAPHITE_HTTP)
```
thiagoftsm, line 43: We could use a ternary operator here, but I suppose you opted for an `if ... else`, because  it is more readable. :thinking:

exporting/json/json.c
```diff
@@ -27,16 +30,78 @@ int init_json_instance(struct instance *instance)
...
+    instance->end_chart_formatting = NULL;
+    instance->end_host_formatting = flush_host_labels;
+    instance->end_batch_formatting = close_batch_json_http;
+
+    instance->prepare_header = json_http_prepare_header;
+
+    instance->check_response = exporting_discard_response;
+
+    instance->buffer = (void *)buffer_create(0);
+    if (!instance->buffer) {
```
thiagoftsm, line 98: The function [buffer_create](https://github.com/netdata/netdata/blob/a99a988d9d38e9cec4450d8a34c66f35a296904d/libnetdata/buffer/buffer.c#L387) calls `callocz` that will shutdown Netdata, so this test and the error message is not necessary.

exporting/json/json.c
```diff
@@ -189,6 +263,11 @@ int format_dimension_stored_json_plaintext(struct instance *instance, RRDDIM *rd
         }
     }
 
+    if (instance->config.type == EXPORTING_CONNECTOR_TYPE_JSON_HTTP) {
+        if (buffer_strlen((BUFFER *)instance->buffer) > 2)
+        buffer_strcat(instance->buffer, ",\n");
```
thiagoftsm, line 268: Missing tabulation.

exporting/opentsdb/opentsdb.c
```diff
@@ -240,26 +258,26 @@ int format_dimension_stored_opentsdb_telnet(struct instance *instance, RRDDIM *r
 }
 
 /**
- * Prepare an HTTP message for OpenTSDB HTTP connector
+ * Ppepare HTTP header
```
thiagoftsm, line 261: Typo error, instead `Prepare` we have `Ppepare`.

exporting/process_data.c
```diff
@@ -363,14 +363,66 @@ int flush_host_labels(struct instance *instance, RRDHOST *host)
...
+    }
+
+    // swap buffers
+    BUFFER *tmp_buffer = last_buffer->buffer;
+    last_buffer->buffer = instance_buffer;
+    instance->buffer = instance_buffer = tmp_buffer;
+
+    buffer_flush(instance_buffer);
+
+    BUFFER *header = last_buffer->header;
```
thiagoftsm, line 400: The `header` variable is useful on lines `401` and `402', but after these lines, it does not need to be used, so you can use direct `last_buffer->header` instead to create this new variable.

thiagoftsm:
@vlvkobal when  I am using this PR and instead to use `OpenTSDB` I use `InfluxDB`, I always receive the alarm:

erro

This is not happening with the current `master` branch.

I am using the same `exporting.conf` independent of the database:

``` 
[opentsdb:http:instanceopentsdb]
    enabled = yes
    destination = localhost:4242
    database = netdata
```

exporting/prometheus/remote_write/remote_write.c
```diff
@@ -40,17 +31,9 @@ int prometheus_remote_write_send_header(int *sock, struct instance *instance)
...
-    ssize_t written = send(*sock, buffer_tostring(header), header_len, flags);
+        buffer_strlen(simple_connector_data->last_buffer->buffer));
 
-    buffer_flush(header);
-
-    if (written != -1 && (size_t)written == header_len)
-        return 0;
-    else
-        return 1;
+    return;
```
thiagoftsm, line 36: This `return` line is always implict added when the compiler has a `void` function on `amd64` or `x86` computers, but in normal situation we are not adding this for our `void` functions, as you can see [here](https://github.com/netdata/netdata/blob/2e2a855935fce6ad9025e81ed479bdbc883d197d/daemon/main.c#L128)  and [here](https://github.com/netdata/netdata/blob/2e2a855935fce6ad9025e81ed479bdbc883d197d/daemon/main.c#L102), so I will suggest you remove this.

exporting/prometheus/remote_write/remote_write.c
```diff
@@ -115,22 +99,36 @@ int init_prometheus_remote_write_instance(struct instance *instance)
     instance->end_host_formatting = NULL;
     instance->end_batch_formatting = format_batch_prometheus_remote_write;
 
-    instance->send_header = prometheus_remote_write_send_header;
+    instance->prepare_header = prometheus_remote_write_prepare_header;
     instance->check_response = process_prometheus_remote_write_response;
 
     instance->buffer = (void *)buffer_create(0);
     if (!instance->buffer) {
```
thiagoftsm, line 106: Please, remove this unnecessary test as described before.

exporting/read_config.c
```diff
@@ -432,7 +439,14 @@ struct engine *read_exporting_config()
         tmp_instance->config.prefix = strdupz(exporter_get(instance_name, "prefix", "netdata"));
 
 #ifdef ENABLE_HTTPS
-        if (tmp_instance->config.type == EXPORTING_CONNECTOR_TYPE_OPENTSDB_USING_HTTP && !strncmp(tmp_ci_list->local_ci.connector_name, "opentsdb:https", 14)) {
+        if ((tmp_instance->config.type == EXPORTING_CONNECTOR_TYPE_GRAPHITE_HTTP &&
```
thiagoftsm, line 442: @ilyam8 take a look here! :smile:

exporting/send_data.c
```diff
@@ -132,52 +153,47 @@ void simple_connector_receive_response(int *sock, struct instance *instance)
...
         ERR_clear_error();
 #endif
 
     struct stats *stats = &instance->stats;
-
-    int ret = 0;
-    if (instance->send_header)
-        ret = instance->send_header(sock, instance);
-
     ssize_t written = -1;
```
thiagoftsm, line 173: It is not necessary to keep this initial value here, because it won't be used.

exporting/send_data.c
```diff
@@ -132,52 +153,47 @@ void simple_connector_receive_response(int *sock, struct instance *instance)
...
-            written = send(*sock, buffer_tostring(buffer), len, flags);
-        }
+    if (exporting_tls_is_enabled(instance->config.type, options) &&
+        connector_specific_data->conn &&
+        connector_specific_data->flags == NETDATA_SSL_HANDSHAKE_COMPLETE) {
+        written = (ssize_t)SSL_write(connector_specific_data->conn, buffer_tostring(header), header_len);
+        written += (ssize_t)SSL_write(connector_specific_data->conn, buffer_tostring(buffer), buffer_len);
+    } else {
+        written = send(*sock, buffer_tostring(header), header_len, flags);
+        written += send(*sock, buffer_tostring(buffer), buffer_len, flags);
```
thiagoftsm, line 185: You should not try to send the `buffer`, if there was problem to send the `header`, please, add a test for the first test.

exporting/send_data.c
```diff
@@ -132,52 +153,47 @@ void simple_connector_receive_response(int *sock, struct instance *instance)
...
-            connector_specific_data->flags == NETDATA_SSL_HANDSHAKE_COMPLETE) {
-            written = (ssize_t)SSL_write(connector_specific_data->conn, buffer_tostring(buffer), len);
-        } else {
-            written = send(*sock, buffer_tostring(buffer), len, flags);
-        }
+    if (exporting_tls_is_enabled(instance->config.type, options) &&
+        connector_specific_data->conn &&
+        connector_specific_data->flags == NETDATA_SSL_HANDSHAKE_COMPLETE) {
+        written = (ssize_t)SSL_write(connector_specific_data->conn, buffer_tostring(header), header_len);
+        written += (ssize_t)SSL_write(connector_specific_data->conn, buffer_tostring(buffer), buffer_len);
```
thiagoftsm, line 182: The `buffer` needs to be sent, if and only if we had success to send the `header`.

exporting/send_data.c
```diff
@@ -132,52 +153,47 @@ void simple_connector_receive_response(int *sock, struct instance *instance)
...
+    } else {
+        written = send(*sock, buffer_tostring(header), header_len, flags);
+        written += send(*sock, buffer_tostring(buffer), buffer_len, flags);
+    }
 #else
-        written = send(*sock, buffer_tostring(buffer), len, flags);
+    written = send(*sock, buffer_tostring(header), header_len, flags);
+    written += send(*sock, buffer_tostring(buffer), buffer_len, flags);
```
thiagoftsm, line 189: We are also missing a test here.

exporting/graphite/graphite.c
vlvkobal: Yes, I like `if` better 😉.

exporting/read_config.c
ilyam8: at least it is easy to follow 😉 

```c
 (a && b) ||
 (c && d) ||
 (e && f) ||
 ...
```

exporting/json/json.c
vlvkobal: Removed.

exporting/json/json.c
vlvkobal: Fixed.

exporting/opentsdb/opentsdb.c
vlvkobal: Fixed.

exporting/prometheus/remote_write/remote_write.c
vlvkobal: Removed.

exporting/prometheus/remote_write/remote_write.c
vlvkobal: Removed.

exporting/process_data.c
vlvkobal: It's also used in 404 and 405.

exporting/send_data.c
vlvkobal: Removed.

vlvkobal:
> when I am using this PR and instead to use OpenTSDB I use InfluxDB, I always receive the alarm

I confirm. Something is wrong with the statistics:
image

exporting/process_data.c
thiagoftsm: Yes, I agree, but on `404` and `405`, we could write only `last_buffer->header = buffer_create(0);`. :thinking:

vlvkobal: Removed the variable.

exporting/send_data.c
vlvkobal: Fixed.

exporting/send_data.c
vlvkobal: Fixed.

exporting/send_data.c
vlvkobal: Fixed.

vlvkobal:
The statistics should work correctly after the last fix.

thiagoftsm (approved):
With the last commits, all the problems were fixed and concerns removed. I did not identify any problem with the tests, but I was not able to test Graphana.

exporting/clean_connectors.c
```diff
@@ -35,3 +35,46 @@ void clean_instance(struct instance *instance)
...
+
+    buffer_free(instance->buffer);
+    buffer_free(simple_connector_data->buffer);
+    buffer_free(simple_connector_data->header);
+
+    struct simple_connector_buffer *next_buffer = simple_connector_data->first_buffer;
+    for (int i = 0; i < instance->config.buffer_on_failures; i++) {
+        struct simple_connector_buffer *current_buffer = next_buffer;
+        next_buffer = next_buffer->next;
+
```
stelfrag, line 59: Is it possible to next_buffer to be already null here?

vlvkobal: No, it's not possible. `next_buffer` can be NULL only if `instance->config.buffer_on_failures == 0`.

exporting/read_config.c
mfundul: I'm always afraid of this type of magic numbers. Why not `strlen()` instead?

vlvkobal: Fixed

thiagoftsm (approved):

mfundul (approved):

stelfrag (approved):

PR #9918, opened 2020-09-11
ACLK Child Availability Messages

underhood:
##### Summary
Implements messages which indicate the availability of children to the Cloud.
New messages implemented:
`child_connect` - indicates a child connected to the node sending the message. This is the equivalent of agent connecting on MQTT
`child_disconnect` - clear from the name
Additionally, popcorning is started (which is now per host), and `connect` payload is sent to Cloud upon timer expiration.

##### Component Name
Agent/ACLK

##### Test Plan
Test with streaming setup. Tested with 3 layered set up during development. Listen for messages mentioned above and check their validity

### DIFF
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -554,6 +554,14 @@ static int aclk_process_query(struct aclk_query_thread *t_info)
     switch (this_query->cmd) {
         case ACLK_CMD_ONCONNECT:
             ACLK_HOST_PTR_COMPULSORY("ACLK_CMD_ONCONNECT");
+#if ACLK_VERSION_MIN < ACLK_V_CHILDRENSTATE
+            if (host != localhost && aclk_shared_state.version_neg < ACLK_V_CHILDRENSTATE) {
+                error("We are not allowed to send connect message in ACLK version before %d", ACLK_V_CHILDRENSTATE);
+                break;
+            }
+#else
+#warning "This check became unnecessary. Remove"
+#endif
 
             debug(D_ACLK, "EXECUTING on connect metadata command for host \"%s\" GUID \"%s\"",
                 host->hostname,
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -567,7 +567,7 @@ static int aclk_process_query(struct aclk_query_thread *t_info)
             debug(D_ACLK, "EXECUTING a chart update command");
             if (!host)
                 fatal("Pointer to host compulsory");
-            aclk_send_single_chart(host->hostname, this_query->query);
+            aclk_send_single_chart(host, this_query->query);
             break;
 
         case ACLK_CMD_CHARTDEL:
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -581,7 +581,20 @@ static int aclk_process_query(struct aclk_query_thread *t_info)
             aclk_execute_query_v2(this_query);
             break;
 
+        case ACLK_CMD_CHILD_CONNECT:
+        case ACLK_CMD_CHILD_DISCONNECT:
+            debug(
+                D_ACLK, "Execution Child %s command",
+                this_query->cmd == ACLK_CMD_CHILD_CONNECT ? "connect" : "disconnect");
+
+            if (!host)
+                fatal("ACLK_CMD_CHILD_CONNECT/DISCONNECT needs host pointer");
+            aclk_send_info_child_connection(host, this_query->cmd);
+            break;
+
         default:
+            errno = 0;
+            error("Unknown ACLK Query Command");
             break;
     }
     debug(D_ACLK, "Query #%ld (%s) done", query_count, this_query->topic);
--- a/aclk/aclk_query.c
+++ b/aclk/aclk_query.c
@@ -512,6 +512,13 @@ cleanup:
     return retval;
 }
 
+#define ACLK_HOST_PTR_COMPULSORY(x) \
+    if (unlikely(!host)) { \
+        errno = 0; \
+        error(x " needs host pointer"); \
+        break; \
+    }
+
 /*
  * This function will fetch the next pending command and process it
  *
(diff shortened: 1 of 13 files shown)


aclk/aclk_query.c
```diff
@@ -638,6 +663,39 @@ void aclk_query_threads_start(struct aclk_query_threads *query_threads)
...
+        rrdhost_aclk_state_unlock(host);
+        return ret;
+    }
+
+    if (!host->aclk_state.t_last_popcorn_update){
+        rrdhost_aclk_state_unlock(host);
+        return ret;
+    }
+
+    time_t t_diff = now_realtime_sec() - host->aclk_state.t_last_popcorn_update;
```
mfundul, line 685: Shouldn't timeouts be monotonic or something? This means that a date change will make the diff wrong or even negative. I realize that this is a widespread problem in the agent though.

aclk/aclk_query.c
```diff
@@ -638,6 +663,39 @@ void aclk_query_threads_start(struct aclk_query_threads *query_threads)
     }
 }
 
+/**
+ * Checks and updates popcorning state of rrdhost
+ * returns actual/updated popcorning state
+ */
+
+ACLK_POPCORNING_STATE aclk_host_popcorning(RRDHOST *host)
```
mfundul, line 671: This function has the same name as the `#define` above, kind of confusing.

mfundul:
What are we testing here exactly? Are those messages supposed to do something, or are they no-op for now and we can't use any extra functionality?

underhood:
@mfundul currently they are not implemented on the cloud side (they are working on it). You must force the ACLK min version to 3 and connect the agent to your own broker so that you can see the messages.

mfundul:
Ah, so that's why I'm not seeing any messages.

aclk/aclk_query.c
underhood: yep will change

aclk/aclk_query.c
underhood: pushed a few commits that should make the names a bit more human friendly. hopefully :)

mfundul (dismissed):

aclk/aclk_query.c
```diff
@@ -551,25 +551,37 @@ static int aclk_process_query(struct aclk_query_thread *t_info)
...
+            meta_state = host->aclk_state.metadata;
+            host->aclk_state.metadata = ACLK_METADATA_SENT;
+            rrdhost_aclk_state_unlock(host);
+            aclk_send_metadata(meta_state, host);
             break;
 
         case ACLK_CMD_CHART:
+            if (!host)
+                fatal("ACLK_CMD_CHART needs host pointer");
+
```
stelfrag, line 571: You check for (!host) twice (a few lines below that)

- consider using unlikely
- How can we remove the fatal? When will that happen, can we ignore and continue?

underhood: hmmm, well if we don't have host here it is a logical programming error, not a runtime error. We could continue though and just ignore the message?

stelfrag: I would say log the error / ignore that message and continue instead of fatal

underhood: changed

mfundul (approved):

aclk/agent_cloud_link.c
```diff
@@ -434,23 +434,123 @@ static struct _collector *_add_collector(const char *hostname, const char *plugi
...
+    }
+    
+    info("Host Disconnected before ACLK popcorning finished. Canceling. Host \"%s\" GUID:\"%s\"", host->hostname, host->machine_guid);
+    host->aclk_state.t_last_popcorn_update = 0;
+    host->aclk_state.metadata = ACLK_METADATA_REQUIRED;
+    rrdhost_aclk_state_unlock(host);
+
+    if(host == aclk_shared_state.next_popcorn_host) {
+        aclk_shared_state.next_popcorn_host = NULL;
+        aclk_update_next_child_to_popcorn(host);
```
stelfrag, line 544: `aclk_update_next_child_to_popcorn` is not defined to take a parameter

underhood: thanks, fixed

mfundul (approved):

stelfrag:
I observe that the parent information is overwritten on the cloud. How to reproduce

1. Have a parent and child setup 
    - Both parent and child are claimed
2. Start the parent and child
3. Stop the parent and wait a few seconds
4. Restart the parent
    - Both nodes show the child's info

underhood:
@stelfrag this is with both nodes claimed in production cloud and agents on this PR?

stelfrag:
> @stelfrag this is with both nodes claimed in production cloud and agents on this PR?

Yes in this case both are claimed in production, but I have node info replacing the "parent" info when the child is claimed in staging as well.

underhood:
@stelfrag changed it a bit to prevent that problem + added an explicit safety check

aclk/agent_cloud_link.c
```diff
@@ -434,23 +434,135 @@ static struct _collector *_add_collector(const char *hostname, const char *plugi
...
+{
+    usec_t now = now_monotonic_sec();
+    info("Starting ACLK popcorn timer for host \"%s\" with GUID \"%s\"", host->hostname, host->machine_guid);
+    ACLK_SHARED_STATE_LOCK;
+    rrdhost_aclk_state_lock(host);
+    if (host == localhost && !ACLK_IS_HOST_INITIALIZING(host)) {
+        errno = 0;
+        error("Localhost is allowed to do popcorning only once after startup!");
+        rrdhost_aclk_state_unlock(host);
+        return;
```
stelfrag, line 527: You need to do a `ACLK_SHARED_STATE_UNLOCK` here

underhood:
force push after resigning commits with new GPG key

underhood:
rebase on latest master

stelfrag (approved):

mfundul (approved):

PR #9913, opened 2020-09-10
Added a way to get build configuration info from the agent.

Ferroin:
##### Summary

This adds a new option to the `-W` switch called 'buildinfo'. When invoked with this argument, Netdata will print it's version, the
configure options, and a list of optional features and whether they are enabled or not.

This is intended to serve three purposes:

* It allows developers to more quickly get an idea of how Netdata was built when triaging bug reports.
* It provides an easier way to validate changes to the build system that affect optional features during the development cycle.
* It provides an easier way to build CI workflows that validate that building under a given set of constraints results in a feature being enabled or not.

The actual implementation is a bit large but overall exceedingly simple, consisting of a set of preprocessor directives to extract optional feature state information from config.h and then a series of printf() calls to actually report this info (which should end up optimized by smart compilers due to all the arguments being compile-time constants).

The use of `-W` for this was chosen to avoid needing a new option name and also avoid polluting the `-V` output.

The output looks like this (from a local build on my development system):

```
Version: netdata v1.24.0-220-g16165edb8
Configure options:  'PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig'
Features:
    dbengine:                NO
    Native HTTPS:            NO
    Netdata Cloud:           NO
    TLS Host Verification:   YES
Libraries:
    jemalloc:                NO
    JSON-C:                  YES
    libcap:                  YES
    libcrypto:               NO
    libm:                    YES
    LWS:                     NO
    mosquitto:               NO
    tcalloc:                 NO
    zlib:                    YES
Plugins:
    apps:                    YES
    cgroup Network Tracking: YES
    CUPS:                    YES
    EBPF:                    NO
    IPMI:                    NO
    NFACCT:                  YES
    perf:                    YES
    slabinfo:                YES
    Xen:                     NO
    Xen VBD Error Tracking:  NO
Exporters:
    AWS Kinesis:             NO
    GCP PubSub:              NO
    MongoDB:                 NO
    Prometheus Remote Write: YES
```

Sample POSIX shell one-liner to test for a specific feature in the build using this output (in this case checking for Netdata Cloud support):

```
[ "$(./netdata -W buildinfo | grep 'Netdata Cloud' | awk '{print $2}')" = "YES" ]
```

##### Component Name

area/build
area/daemon

##### Test Plan

Testing is done by simply building the agent and validating the output of `netdata -W buildinfo` against the contents of `config.log`.

##### Additional Information

[...]

PR #9936, opened 2020-09-16
adds ACLK DISABLE_CLOUD to -W buildinfo

underhood:
##### Summary
I just noticed #9913 from @Ferroin got merged yesterday.

This PR improves by adding better info if ACLK was disabled by user request

##### Test Plan
build with `--disable-cloud`
build with ACLK

@Ferroin good idea to bring this feature as it allows us to diagnose issues that users might have much more easily.

### DIFF
--- a/daemon/buildinfo.c
+++ b/daemon/buildinfo.c
@@ -8,7 +8,11 @@
 #ifdef ENABLE_ACLK
 #define FEAT_CLOUD "YES"
 #else
+#ifdef DISABLE_CLOUD
+#define FEAT_CLOUD "NO (by user request e.g. '--disable-cloud')"
+#else
 #define FEAT_CLOUD "NO"
+#endif
 #endif
 
 #ifdef ENABLE_DBENGINE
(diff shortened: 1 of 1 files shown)


daemon/buildinfo.c
```diff
@@ -181,6 +181,14 @@ void print_build_info(void) {
     printf("    dbengine:                %s\n", FEAT_DBENGINE);
     printf("    Native HTTPS:            %s\n", FEAT_NATIVE_HTTPS);
     printf("    Netdata Cloud:           %s\n", FEAT_CLOUD);
+#ifndef ENABLE_ACLK
+#ifdef DISABLE_CLOUD
+    printf("        Cloud Disabled By User Request (e.g. installer flag --disable-cloud)\n");
+#else
+    printf("        Usable Libwebsockets:       %s\n", FEAT_LWS);
+    printf("        Usable Mosquitto:           %s\n", FEAT_MOSQUITTO);
```
Ferroin, line 189: Strictly speaking, we shouldn't need this block here, as we have both LWS and mosquitto listed in the libaries.

I do agree that we should have something special to indicate that the user explicitly disabled cloud support though.

underhood: yep you are right, I wanted to have it in one place but I can remove

underhood: I tried to keep it together also in configure script:
so you have all you need near the line that says `can't build ACLK` and no need to scroll and guess..
but you are right this is not user facing thing but more of a troubleshooting so I will remove it

underhood: changed it to your wishes

mfundul (approved):

Ferroin (approved):

PR #9963, opened 2020-09-21
Bumps ACLK version to 3

underhood:
##### Summary
ACLK v3 doesn't change ACLK itself but is used to signify data query for context is supported in this agent. This way Cloud can know which agents from the current room it can include in aggregate view,


underhood:
decision was made to not bump version for this feature due to v_neg not being implemented on cloud side in time

Issue #7822, opened 2020-01-22
Refactor scripts that publish to packagecloud for nightlies


prologic:

prologic:
@Ferroin I _know_ you are working on this (_and the other related issues_); How close are we?

Ferroin:
Reasonably close. I should hopefully hav ethe time to start actively testing the new packaging workflow today.

Issue #7346, opened 2019-11-21
Eliminate LXC from package generation pipeline


cosmix:
### Rationale

Currently LXC is used to build the agent packages for linux distributions. This has already proved problematic, necessitating several workarounds to make things work. More people are familiar with Docker (vs LXC), there is much more community support and documentation, let alone tooling to automate this, and it would make sense to move our package building to Docker.

prologic:
@Ferroin Can we turn this into an EPIC? I can't tell how far along we are with this and/or if you need any help? :) I'm pretty full up at the moment with eBPF but I'd love to help you here if you need it!

Ferroin:
> @Ferroin Can we turn this into an EPIC? I can't tell how far along we are with this and/or if you need any help? :) I'm pretty full up at the moment with eBPF but I'd love to help you here if you need it!

I don't know that an EPIC really fits in this case, as there are only two intermediary steps, both are done, and neither have a direct user-visible impact.

As far as progress, right now I'm just trying to find the time to actually test the GHA workflows I've put together for this. I'd be happy to push the branch to my fork if you want to look at it and try testing it though.

prologic:
> As far as progress, right now I'm just trying to find the time to actually test the GHA workflows I've put together for this. I'd be happy to push the branch to my fork if you want to look at it and try testing it though.

Please by all means :D early feedback is always better! I can help you test it :)

Ferroin:
> > As far as progress, right now I'm just trying to find the time to actually test the GHA workflows I've put together for this. I'd be happy to push the branch to my fork if you want to look at it and try testing it though.
> 
> Please by all means :D early feedback is always better! I can help you test it :)

OK, the branch is up. It's called `gha-packaging`. Right now it's just doing amd64 packages, but it covers all 5 distros we use with all versions we support (and takes significant advantage of YAML's anchor map merging to avoid duplication). It's also using a custom local action to do the commit filtering, because there not only isn't any support for this in GHA, but there isn't any published action I could find that does it the way we need to either.

prologic:
> OK, the branch is up. It's called `gha-packaging`. Right now it's just doing amd64 packages, but it covers all 5 distros we use with all versions we support (and takes significant advantage of YAML's anchor map merging to avoid duplication).

Thanks I'll check it out! Can you put the PR itself up too in Draft form?

> It's also using a custom local action to do the commit filtering, because there not only isn't any support for this in GHA, but there isn't any published action I could find that does it the way we need to either.

[...]

PR #9964, opened 2020-09-21
Docker-based packaging workflow in GitHub Actions.

Ferroin:
##### Summary

This implements a new Docker-based packaging CI workflow in GitHub Actions. It still uses the same trigger commits as the existing LXC-based workflow we have in Travis so that minimal changes are needed to have it work correctly.

This requires the addition of two secrets to the repository itself:

* `PACKAGE_CLOUD_REPO`: This specifies which PackageCloud repository to push the packages to.
* `PACKAGE_CLOUD_API_KEY`: This specifies the PackageCloud API key to use when uploading packages.

This leverages GitHub Actions' matrix build feature to keep overall code in the workflow file to a minimum.

##### Component Name

area/ci
area/packaging

##### Additional Information

Fixes: #7346 
Fixes: #7822


Ferroin:
For those who want to verify the packages that would be produced by this locally, here are the steps required to use the same package building process locally:

1. Pull the correct Docker image for your distribution and architecture from [the `netdata/package-builders` repository on Docker Hub](https://hub.docker.com/repository/docker/netdata/package-builders/general). There is one tag there for each supported distribution.
2. From your local copy of the Netdata sources, run: `docker run -it --rm -v $PWD:/netdata netdata/package-builders:<tag>`, replacing `<tag>` with the tag for your distribution (for example `debian10` for a Debian 10 package).

This will create an `artifacts` directory in the Netdata source tree on the host system which will contain the package files, which you can then manually install on your system by passing them directly to your package manager. The generated packages will list a version of `0.1` instead of the current version, but are otherwise the current version of the code provided by the Netdata sources used for the build.

The package building process itself takes about 5-10 minutes on a reasonably fast system.

Note that we currently only build for 64-bit x86 on most of these platforms (and the handful we don’t only include 32-bit x86 beyond that).

knatsakis (approved):

kaskavel (approved):

knatsakis (approved):

PR #9967, opened 2020-09-21
Fix setting for disabling eBPF-apps.plugin integration

joelhans:
##### Summary

We had our `yes` and `no` mixed up for this setting. This PR is a quick fix to flip the logic.

cc @OdysLam 

##### Component Name

area/docs

##### Test Plan

Verified on my local system that setting `disable apps = yes` removed the eBPF charts from the **Applications** section. They were enabled and visible by default.


thiagoftsm (approved):
LGTM!

prologic (approved):

Issue #8961, opened 2020-05-11
Allow use of system libwebsockets and mosquitto


candrews:
##### Feature idea summary
Currently, netdata builds bundled copies of libwebsockets and mosquitto. It would be better to (at least optionally) use the system provided libwebsockets and mosquitto libraries, just like netdata does for all of its other dependencies.

##### Expected behavior
In the autotools build system, `configure` should take `--enable-libwebsockets` and `--enable-mosquitto` arguments which would look for and use the system libraries.

Note that this issue is a blocker for enabling netdata cloud on distributions that prohibit library bundling, such as Gentoo.

amoss:
The version of mosquitto that we use is not standard so we cannot use the system copy. Now that we have got through the release we will talk to the maintainers and see if we can merge our modifications back upstream.

For libwebsockets we have a different problem - we need a specific version that has been compiled in a specific way. If one of the options was not specified in the build then we will link against the system copy successfully but the link will silently fail at runtime. We currently have no way around this so we build the libwebsockets library to ensure that the correct flags are set.

At this time we cannot switch to using the system libraries, even if they are installed.

candrews:
> The version of mosquitto that we use is not standard so we cannot use the system copy. Now that we have got through the release we will talk to the maintainers and see if we can merge our modifications back upstream.

That's good news - perhaps when you've submitted the change requests upstream, you could link to them here, and interested parties can follow along.



> For libwebsockets we have a different problem - we need a specific version that has been compiled in a specific way. If one of the options was not specified in the build then we will link against the system copy successfully but the link will silently fail at runtime. We currently have no way around this so we build the libwebsockets library to ensure that the correct flags are set.

Perhaps change requests can be submitted upstream to address this problem, just like what is planned for mosquitto?

amoss:
I'm not that we can do the same with libwebsockets. It is not a change to the code or the functionality of the library, the problem is the build configuration. In our case we had weird problems if libcap was not included in the libwebsockets build via a compile-time flag. This is not a change that we can push back upstream - they have support for it already, but it depends on how it is compiled for system packages.

candrews:
> In our case we had weird problems if libcap was not included in the libwebsockets build via a compile-time flag.

How about making it optional via a `configure` option? For example, `--enable-system-libwebsockets`? Each distro's packager would be responsible for making sure that the libwebsockets system library is compiled with `libcap` before they enable this option for their package.

Ferroin:
[...]

PR #9973, opened 2020-09-22
allow using LWS without SOCKS5

underhood:
##### Summary
As a first step towards allowing usage of system LWS we allow building Netdata with LWS that was built without SOCKS5.
In case the user tries to use SOCKS5 we will log appropriate error into the logfile and ignore the setting.

More detailed discussion #8961 

##### Component Name
ACLK

##### Test Plan
Build with LWS without SOCKS 5 support. Should build without error. Example how to do that: Cripple installer.sh (remove `LWS_WITH_SOCKS5:bool=ON`) here (in else branch):
```
build_libwebsockets() {
  local env_cmd=''

  if [ -z "${DONT_SCRUB_CFLAGS_EVEN_THOUGH_IT_MAY_BREAK_THINGS}" ]; then
    env_cmd="env CFLAGS= CXXFLAGS= LDFLAGS="
  fi

  pushd "${1}" > /dev/null || exit 1
  if [ "$(uname)" = "Darwin" ] && [ -d /usr/local/opt/openssl ]; then
    run ${env_cmd} cmake \
      -D OPENSSL_ROOT_DIR=/usr/local/opt/openssl \
      -D OPENSSL_LIBRARIES=/usr/local/opt/openssl/lib \
      -D LWS_WITH_SOCKS5:bool=ON \
      $CMAKE_FLAGS \
      .
  else
    run ${env_cmd} cmake -D LWS_WITH_SOCKS5:bool=ON $CMAKE_FLAGS .
```
Attempting to use SOCKS5 proxy will result in error message and such setting to be ignored.

### DIFF
--- a/aclk/aclk_common.c
+++ b/aclk/aclk_common.c
@@ -149,7 +149,7 @@ const char *aclk_lws_wss_get_proxy_setting(ACLK_PROXY_TYPE *type)
 
     if (strcmp(proxy, ACLK_PROXY_ENV) == 0) {
         if (check_socks_enviroment(&proxy) == 0) {
-#ifdef LWS_WITH_SOCKS
+#ifdef LWS_WITH_SOCKS5
             *type = PROXY_TYPE_SOCKS5;
             return proxy;
 #else
--- a/aclk/aclk_common.c
+++ b/aclk/aclk_common.c
@@ -2,6 +2,10 @@
 
 #include "../daemon/common.h"
 
+#ifdef ENABLE_ACLK
+#include <libwebsockets.h>
+#endif
+
 netdata_mutex_t aclk_shared_state_mutex = NETDATA_MUTEX_INITIALIZER;
 
 int aclk_disable_runtime = 0;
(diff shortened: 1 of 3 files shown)


mfundul (dismissed):

aclk/aclk_common.c
```diff
@@ -144,14 +148,31 @@ const char *aclk_lws_wss_get_proxy_setting(ACLK_PROXY_TYPE *type)
...
             *type = PROXY_TYPE_HTTP;
         return proxy;
     }
 
     *type = aclk_verify_proxy(proxy);
+#ifndef LWS_WITH_SOCKS5
+    if (*type == PROXY_TYPE_SOCKS5) {
+        char *log = strdupz(proxy);
+        safe_log_proxy_censor(log);
+        fatal("Config var \"" ACLK_PROXY_CONFIG_VAR "\" set to use SOCKS5 proxy but this agent "
```
mfundul, line 171: We shouldn't kill the agent, but we should drop ACLK instead.

mfundul (approved):

aclk/aclk_common.c
underhood: I changed it to behavior discussed on Slack.
Now proxy setting is checked at the beginning of the ACLK thread and will be disabled if sepcifically SOCKS5 is requested by config.

underhood: that means we will not `fatal` after last changes same as you suggested

mfundul (dismissed):

stelfrag:
I get 

```
ACLK_Main : socks_proxy environment variable set to use SOCKS5 proxy but Libwebsockets used doesn't have SOCKS5 support built in. Ignoring and checking for other options. Provided Value:"socks5://XXXXXXXXXXX@192.168.1.10:7000" (errno 2, No such file or directory)
```

But i do have it compiled

```
 LWS_WITH_RANGES = OFF
 LWS_PLAT_OPTEE = OFF
 LWS_WITH_ESP32 = OFF
 LWS_WITH_ZIP_FOPS = OFF
 LWS_AVOID_SIGPIPE_IGN = OFF
 LWS_WITH_STATS = OFF
 LWS_WITH_SOCKS5 = ON
 LWS_HAVE_SYS_CAPABILITY_H = 
 LWS_HAVE_LIBCAP = 
 LWS_WITH_PEER_LIMITS = OFF
```

underhood:
@stelfrag was caused by typo

stelfrag (approved):

mfundul (approved):

PR #9978, opened 2020-09-23
Improved the data query when using the context parameter

stelfrag:
##### Summary
Improved the processing of the data query when data for a `context` is requested

- Define a generic structure to pass along to functions (instead of RRDDIM linked list) to allow for easy changes in the future
- Collect the matching charts, dimensions and calculate first_entry_t / last_entry_t at the same time
- Speedup context match by using the context hash in RRDSET
- Add correct response `Chart is not found:` or `Context is not found:`

##### Component Name
database

##### Test Plan
Example context call

```
http://127.0.0.1:19999/api/v1/data?context=cpu.cpu&options=jsonwrap
```

### DIFF
--- a/web/api/formatters/rrd2json.c
+++ b/web/api/formatters/rrd2json.c
@@ -24,7 +24,7 @@ static inline void free_temp_rrddim(RRDDIM *temp_rd)
 
 void free_context_param_list(struct context_param **param_list)
 {
-    if (unlikely(!param_list  || !*param_list))
+    if (unlikely(!param_list || !*param_list))
         return;
 
     free_temp_rrddim(((*param_list)->rd));
@@ -34,8 +34,6 @@ void free_context_param_list(struct context_param **param_list)
 
 void build_context_param_list (struct context_param **param_list, RRDSET *st)
 {
-    st->last_accessed_time = now_realtime_sec();
-
     if (unlikely(!param_list || !st))
         return;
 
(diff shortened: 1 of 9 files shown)


web/api/formatters/rrd2json.c
```diff
@@ -22,6 +22,58 @@ static inline void free_temp_rrddim(RRDDIM *temp_rd)
...
+        (*param_list)->first_entry_t = LONG_MAX;
+        (*param_list)->last_entry_t = 0;
+        (*param_list)->last_accessed_time = now_realtime_sec();
+        (*param_list)->rd = NULL;
+    }
+
+    RRDDIM *rd1;
+    rrdset_rdlock(st);
+
+    st->last_accessed_time = (*param_list)->last_accessed_time;
```
mfundul, line 51: The loop that calls `build_context_param_list()` can take many seconds depending on the load of the computer. As a result, the `last_accessed_time` can be a minute old, and in the meantime this chart may have been queried many times since then.

mfundul (changes requested):

underhood (approved):

web/api/formatters/rrd2json.c
```diff
@@ -121,7 +171,7 @@ int rrdset2value_api_v1(
     long i = (!(options & RRDR_OPTION_REVERSED))?rrdr_rows(r) - 1:0;
     *n = rrdr2value(r, i, options, value_is_null);
 
-    free_temp_rrddim(temp_rd);
+    //free_temp_rrddim(temp_rd);
```
mfundul, line 174: You can remove this.

web/api/formatters/rrd2json.c
```diff
@@ -355,7 +372,7 @@ int rrdset2anything_api_v1(
         break;
     }
 
-    free_temp_rrddim(temp_rd);
+    //free_temp_rrddim(temp_rd);
```
mfundul, line 375: You can remove this.

web/api/formatters/rrd2json.c
```diff
@@ -22,6 +22,57 @@ static inline void free_temp_rrddim(RRDDIM *temp_rd)
...
+{
+    if (unlikely(!param_list || !*param_list))
+        return;
+
+    free_temp_rrddim(((*param_list)->rd));
+    freez((*param_list));
+    *param_list = NULL;
+}
+
+void build_context_param_list (struct context_param **param_list, RRDSET *st)
```
mfundul, line 35:
```suggestion
void build_context_param_list(struct context_param **param_list, RRDSET *st)
```

mfundul (changes requested):

mfundul (approved):

underhood (approved):

PR #9998, opened 2020-09-28
minor - removes leading whitespace before JSON in ACLK

underhood:
##### Summary
Each message was prepended by the leading tab. This typo causes Cloud to have bothersome additional processing step for agent messages that skips leading whitespace before parsing JSON message.

##### Component Name
ACLK
##### Test Plan
messages should not have leading space anymoar.

Before:
```
       {"type": "hello",
        "msg-id": "32a7e02e-8c68-498c-8f5c-a4e010e3f95c",
        "timestamp": 1601307084,
        "timestamp-offset-usec": 4898,
        "connect": 1601307083,
        "connect-offset-usec": 29813,
        "version": 1,
```
After:
```
{       "type": "hello",
        "msg-id": "32a7e02e-8c68-498c-8f5c-a4e010e3f95c",
        "timestamp": 1601307084,
        "timestamp-offset-usec": 4898,
        "connect": 1601307083,
        "connect-offset-usec": 29813,
        "version": 1,
```

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -1214,7 +1214,7 @@ inline void aclk_create_header(BUFFER *dest, char *type, char *msg_id, time_t ts
 
     buffer_sprintf(
         dest,
-        "\t{\"type\": \"%s\",\n"
+        "{\t\"type\": \"%s\",\n"
         "\t\"msg-id\": \"%s\",\n"
         "\t\"timestamp\": %ld,\n"
         "\t\"timestamp-offset-usec\": %llu,\n"


stelfrag (approved):

aclk/agent_cloud_link.c
```diff
@@ -1214,7 +1214,7 @@ inline void aclk_create_header(BUFFER *dest, char *type, char *msg_id, time_t ts
 
     buffer_sprintf(
         dest,
-        "\t{\"type\": \"%s\",\n"
+        "{\t\"type\": \"%s\",\n"
```
fracasula, line 1217: I wouldn't move the tab after the curly bracket though (i.e. just `{"type":`). If possible at all it would be great to just remove all unnecessary whitespaces, if that is not possible we can leave it I guess. It's not a big deal but the reason I'm asking is that I'm running a regular expression to detect whether a message is an `hello` one or not. I can account for all possible whitespaces in the regex of course but I thought it'd be better to ask for a change to save some bandwidth in the communication as well.

underhood: I am all for a minified JSON in ACLK. Actually this has been my stance since the beginning. There are plenty of tools to pretty-print JSON for humanz where needed.
Even during development, I use test server/tools in Ruby and they do pretty print for me where needed.

underhood: @fracasula This fixes the problem so that you can skip the whitespace processing step on your side.

If @stelfrag and @mfundul don't have an issue with sending minified JSON I will then think of minifying everything at once with one PR.
There is a bit of caveat on our side because we use `/api/v1/info` and embed it into our JSON and that is not minified as well.

fracasula: Let's wait for the minify then and close this PR? I don't mind having an extra check now if we can have a minified version later. Consider that all messages get debounced as well so they end up in Redis back and forth and then in Pulsar where we have considerable retention enabled as well. It would save us quite some space in the long run.

underhood: well, they are not mutually exclusive.
this removes the leading whitespace problem, next PR will minify

fracasula: True but the minification should remove the trailing whitespace too so I proposed we postpone since it's a minor issue like you presented it in the title of this PR. I'm fine either way.

underhood: Don't see any advantage in waiting. It will just delay when you can remove the needless code on the cloud side, doesn't really give us any advantage

thiagoftsm (approved):
LGTM!

PR #9999, opened 2020-09-28
Allow connecting to arbitrary MQTT WSS broker for devs

underhood:
##### Summary
For easier testing of ACLK, it is useful to connect to an arbitrary MQTT WSS brokers. This allows doing that easily by adding `-DACLK_DISABLE_CHALLENGE` to the compiler.
This will disable challenge-response functionality which is supported and specific only to netdata cloud.

During the implementation, I noticed we are unnecessarily passing the port as both string and integer and carrying that around the code. This fixes also that problem.

Review the 2 commits separately for a clearer picture as one does code cleanup and the second one actually implements the feature,

##### Component Name
ACLK

##### Test Plan
compile with `-DACLK_SSL_ALLOW_SELF_SIGNED -DACLK_DISABLE_CHALLENGE`, set up `cloud base url = 127.0.0.1:9002` in `cloud.conf`. Set mosquito broker (compiled with WSS support) to listen with WSS listener on port `9002`. Example mosquito config:
```
port 1883
listener 9002
protocol websockets
certfile server.crt
keyfile server.key
log_dest stdout
log_type all
websockets_log_level 9999
```

Agent should skip challenge-response and connect to the broker successfully.

### DIFF
--- a/aclk/agent_cloud_link.c
+++ b/aclk/agent_cloud_link.c
@@ -821,21 +821,36 @@ CLEANUP:
 
 static void aclk_try_to_connect(char *hostname, int port)
 {
+    int rc;
+
+// this is usefull for developers working on ACLK
+// allows connecting agent to any MQTT broker
+// for debugging, development and testing purposes
+#ifndef ACLK_DISABLE_CHALLENGE
     if (!aclk_private_key) {
-            error("Cannot try to establish the agent cloud link - no private key available!");
-            return;
+        error("Cannot try to establish the agent cloud link - no private key available!");
+        return;
     }
+#endif
+
     info("Attempting to establish the agent cloud link");
+#ifdef ACLK_DISABLE_CHALLENGE
+    if (aclk_password == NULL)
+        aclk_password = strdupz("anon");
+#else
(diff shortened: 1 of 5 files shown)


aclk/agent_cloud_link.c
```diff
@@ -819,24 +819,39 @@ void aclk_get_challenge(char *aclk_hostname, char *aclk_port)
...
 
-static void aclk_try_to_connect(char *hostname, char *port, int port_num)
+static void aclk_try_to_connect(char *hostname, int port)
 {
+    int rc;
+
+// this is usefull for developers working on ACLK
+// allows connecting agent to any MQTT broker
+// for debugging, development and testing purposes
+#ifndef ACLK_DISABLE_CHALLENGE
```
mfundul, line 829: You should add an `#else` clause that prints a huge `error()` message with all capitals warning the user that it's an insecure build and it's not supported by netdata.

underhood: Will add. No worries agent built like this is not able to connect to Cloud though.

mfundul (approved):

stelfrag (approved):

PR #10002, opened 2020-09-29
Fix Docs GitHub Action with ignore list and update

joelhans:
##### Summary

I've noticed that the Docs GitHub Action always fails due to failing to check the link `https://pi-hole.net`. There's two problems with this. First, it shouldn't always be checking links in the Pi-hole guide—that's a bug. Second, the link is functional, so I'm choosing to ignore the failure.

##### Component Name

area/ci

### DIFF
--- a/.mlc_config.json
+++ b/.mlc_config.json
@@ -1,7 +1,7 @@
 {
     "ignorePatterns": [
 		{
-			"pattern": "^https://pi-hole.net"
+			"pattern": "^https:\/\/pi-hole.net"
 		}
 	],
     "replacementPatterns": [
--- a/.mlc_config.json
+++ b/.mlc_config.json
@@ -1,7 +1,7 @@
 {
     "ignorePatterns": [
 		{
-			"pattern": "^https://pi-hole.net/"
+			"pattern": "^https://pi-hole.net"
 		}
 	],
     "replacementPatterns": [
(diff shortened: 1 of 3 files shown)


.mlc_config.json
```diff
@@ -1,4 +1,9 @@
 {
+    "ignorePatterns": [
+		{
+			"pattern": "^https://pi-hole.net"
```
prologic, line 4: Technically if this is a regex is should probably be `^https\:\/\/pi-hole\.net`

prologic (dismissed):

.mlc_config.json
joelhans: Cool, thanks! I saw some errors in VSCode related to having the escape in front of the `:` and `.`. Do you know if this matters? It stopped being mad at me if the regex is `^https:\/\/pi-hole.net`.

prologic (approved):

joelhans:
@Ferroin Mind looking at this? It would be nice to fix the **Docs** GHA so that we can more robustly ensure that we're not merging broken links into the docs.

Ferroin (approved):

Ferroin:
@joelhans Sorry about the delay in getting this approved, I seem to have missed it somehow...

joelhans:
@Ferroin No worries! I forgot about it myself until I was tracking down some broken links yesterday. Looks like James' approval isn't valid any more, so we need one more. Have anyone in mind?

Ferroin:
At this point there really isn’t anyone else other than me, so I’ll just merge it.

PR #10024, opened 2020-10-05
Fixed broken changelog link

danshilm:
##### Summary
Simple update of the link to the changelog when an update is available. I'm guessing it wasn't updated when the location of the docs was changed.

##### Component Name
Web GUI

### DIFF
--- a/docs/Running-behind-apache.md
+++ b/docs/Running-behind-apache.md
@@ -123,7 +123,6 @@ with this content:
 
 ```conf
 <VirtualHost *:80>
-	RewriteEngine On
 	ProxyRequests Off
 	ProxyPreserveHost On
 	


OdysLam:
Although it appears to redirect, so it doesn't any effect to the user experience, thanks of the fix @danshilm  and welcome to our community!

We are also running a community forum at https://community.netdata.cloud

cc @joelhans

joelhans (approved):
Thanks for the fix, @danshilm! We may need to make this change over in the netdata/dashboard repo as well.

danshilm:
> Although it appears to redirect, so it doesn't any effect to the user experience, thanks of the fix @danshilm and welcome to our community!
> 
> We are also running a community forum at https://community.netdata.cloud
> 
> cc @joelhans

I hadn't tried navigating to the page, but yes, it does indeed redirect when navigating to the page, but not when using the link that shows up on the dashboard when checking for updates. 


> Thanks for the fix, @danshilm! We may need to make this change over in the [netdata/dashboard](https://github.com/netdata/dashboard) repo as well.

Sure, I'll make a PR on the dashboard repo as well.

jacekkolasa (approved):

ilyam8:
was fixed in https://github.com/netdata/netdata/pull/10586

PR #10027, opened 2020-10-05
Completely hide SSO iframe

Jiab77:
##### Summary

The SSO `iframe` is not completely hidden and it appears as a small visible square.

##### Component Name

Web / Gui

##### Test Plan

Add / Remove `style="border: none;"` on the `iframe` with the web console

##### Additional Information

You can see the issue with the picture below:

image

After patch:

image

I've spotted the issue on the stable and nightly versions.

### DIFF
--- a/web/gui/index.html
+++ b/web/gui/index.html
@@ -1315,7 +1315,7 @@
             </div>
         </div>
     </div>
-    <iframe id="ssoifrm" width="0" height="0"></iframe>
+    <iframe id="ssoifrm" width="0" height="0" style="border: none;"></iframe>
     <div id="hiddenDownloadLinks" style="display: none;" hidden></div>
     <script type="text/javascript" src="dashboard.js?v20190902-0"></script>
 </body>


Jiab77:
Any updates on this pull request?

ilyam8:
@jacekkolasa please take a look

Jiab77:
@ilyam8 Thanks a lot!

jacekkolasa (approved):
that looks good, thanks @Jiab77 ! Apologies for the delays. We'll need to add the same change to https://github.com/netdata/dashboard because unfortunately right now that file is duplicated (here it's the old dashboard, which is accessible via /old path)

Jiab77:
Hello @jacekkolasa, before creating this pull request, I've tried to look at other files / pages that would probably contains the same issue but I just found this one.

Sorry to haven't found the other pages, I'm still lacking complete visibility of the project in my mind. Let me know if you need/want me to do other changes.

jacekkolasa:
No worries! I admit it is confusing, i'm a little afraid that removing that old code will result in third-party packaging problems, but we'll need to rethink that.

VLegakis (approved):

novykh (approved):

Jiab77:
@jacekkolasa I'm sorry for having raised an internal packaging issue... Let me know if I can help.

jacekkolasa:
@Jiab77 no no, you didn't break anything, i was just explaining why we keep the older dashboard code here. That change is very welcome, i think you can merge this PR :)

Jiab77:
@jacekkolasa Ok great! But I can't merge it myself 😅 

Anyway, while continuing to analyse the iframes, I've found some interesting stuffs in the `sign-in` one like this:

image

Which explains why the `favicon` is the one from the ReactJS project:

image

I guess I should create another pull request or at least an issue?

jacekkolasa:
I think that react favicon is not visible when Dashboard is accessed normally, but please correct me if i'm wrong. I don't see any harm in that if i can be honest. :)

Jiab77:
@jacekkolasa Nope, it's not visible from the dashboard. It's only visible when calling the URL directly. But you should probably change the `<meta name="description">` with something more related to the netdata cloud instead of the default `create-react-app` sample.

Anyway I've just reported what I've seen, only you can decide if it worth to be changed or not 😉

PR #10030, opened 2020-10-05
ACLK allow child query

underhood:
##### Summary
Enables querying children in ACLK Protocol v3 by their GUID.

##### Component Name
ACLK

##### Test Plan
Connect agent to your own broker. Send `http` messages to query data. Try for master and try for child with `/host/[GUID]/api/v1/info/data?`
If you can emulate `hello` response with `version` message. Otherwise you can build agent modified with `ACLK_VERSION_MIN 3` and thus forcing v3 to agent.

### DIFF
--- a/aclk/aclk_rx_msgs.c
+++ b/aclk/aclk_rx_msgs.c
@@ -15,18 +15,52 @@ static inline int aclk_extract_v2_data(char *payload, char **data)
     return 0;
 }
 
-static inline int aclk_v2_payload_get_query(const char *payload, struct aclk_request *req)
+#define ACLK_GET_REQ "GET "
+#define ACLK_CHILD_REQ "/host/"
+#define ACLK_CLOUD_REQ_V2_PREFIX "/api/v1/"
+#define STRNCMP_CONST(str, const) strncmp(str, const, strlen(const))
+static inline int aclk_v2_payload_get_query(struct aclk_cloud_req_v2 *cloud_req, struct aclk_request *req)
 {
-    const char *start, *end;
+    const char *start, *end, *ptr;
+    char uuid_str[UUID_STR_LEN];
+    uuid_t uuid;
 
-    if(strncmp(payload, ACLK_CLOUD_REQ_V2_PREFIX, strlen(ACLK_CLOUD_REQ_V2_PREFIX))) {
-        errno = 0;
+    errno = 0;
+
+    if(STRNCMP_CONST(cloud_req->data, ACLK_GET_REQ)) {
+        error("Only accepting GET HTTP requests from CLOUD");
+        return 1;
+    }
+    start = ptr = cloud_req->data + strlen(ACLK_GET_REQ);
+
+    if(!STRNCMP_CONST(ptr, ACLK_CHILD_REQ)) {
+        ptr += strlen(ACLK_CHILD_REQ);
+        if(strlen(ptr) < UUID_STR_LEN) {
+            error("the child id in URL too short \"%s\"", start);
+            return 1;
+        }
+
+        strncpyz(uuid_str, ptr, UUID_STR_LEN - 1);
+
+        if(ptr[0] && uuid_parse(uuid_str, uuid)) {
+            error("Got Child query (/host/XXX/...) host id \"%s\" doesn't look like valid GUID", uuid_str);
+            return 1;
+        }
+        ptr += UUID_STR_LEN - 1;
+
+        cloud_req->host = rrdhost_find_by_guid(uuid_str, 0);
+        if(!cloud_req->host) {
+            error("Cannot find host with GUID \"%s\"", uuid_str);
+            return 1;
+        }
+    }
+
+    if(strncmp(ptr, ACLK_CLOUD_REQ_V2_PREFIX, strlen(ACLK_CLOUD_REQ_V2_PREFIX))) {
         error("Only accepting requests that start with \"%s\" from CLOUD.", ACLK_CLOUD_REQ_V2_PREFIX);
         return 1;
     }
-    start = payload + 4;
 
-    if(!(end = strstr(payload, " HTTP/1.1\x0D\x0A"))) {
+    if(!(end = strstr(ptr, " HTTP/1.1\x0D\x0A"))) {
         errno = 0;
         error("Doesn't look like HTTP GET request.");
         return 1;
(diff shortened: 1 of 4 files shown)


underhood:
rebased on latest master after this has been idle long time

aclk/aclk_rx_msgs.c
```diff
@@ -15,18 +15,52 @@ static inline int aclk_extract_v2_data(char *payload, char **data)
...
-static inline int aclk_v2_payload_get_query(const char *payload, struct aclk_request *req)
+#define ACLK_GET_REQ "GET "
+#define ACLK_CHILD_REQ "/host/"
+#define ACLK_CLOUD_REQ_V2_PREFIX "/api/v1/"
+#define STRNCMP_CONST(str, const) strncmp(str, const, strlen(const))
+static inline int aclk_v2_payload_get_query(struct aclk_cloud_req_v2 *cloud_req, struct aclk_request *req)
 {
-    const char *start, *end;
+    const char *start, *end, *ptr;
+    char uuid_str[UUID_STR_LEN];
```
stelfrag, line 25: we use `GUID_LEN + 1` for this in a lot of places in the code

aclk/aclk_rx_msgs.c
```diff
@@ -15,18 +15,52 @@ static inline int aclk_extract_v2_data(char *payload, char **data)
...
+        strncpyz(uuid_str, ptr, UUID_STR_LEN - 1);
+
+        if(ptr[0] && uuid_parse(uuid_str, uuid)) {
+            error("Got Child query (/host/XXX/...) host id \"%s\" doesn't look like valid GUID", uuid_str);
+            return 1;
+        }
+        ptr += UUID_STR_LEN - 1;
+
+        cloud_req->host = rrdhost_find_by_guid(uuid_str, 0);
+        if(!cloud_req->host) {
```
stelfrag, line 52: The host guid id is stored in the lower case form by using `uuid_unparse_lower`. Although the cloud has no reason to change that, we may need to handle it so that `rrdhost_find_by_guid` works for upper and lower case uuid string representation.

aclk/aclk_rx_msgs.c
```diff
@@ -15,18 +15,52 @@ static inline int aclk_extract_v2_data(char *payload, char **data)
...
-    if(strncmp(payload, ACLK_CLOUD_REQ_V2_PREFIX, strlen(ACLK_CLOUD_REQ_V2_PREFIX))) {
-        errno = 0;
+    errno = 0;
+
+    if(STRNCMP_CONST(cloud_req->data, ACLK_GET_REQ)) {
+        error("Only accepting GET HTTP requests from CLOUD");
+        return 1;
+    }
+    start = ptr = cloud_req->data + strlen(ACLK_GET_REQ);
+
```
stelfrag, line 35: Should we avoid / optimize these strlen functions on the defined strings?

underhood: the compiler will turn that into constant AFAIK

underhood: yep I remember that correctly `strlen` on literal/constant string is resolved at compile time and turned into constant.

underhood: yep I checked you can make small program like this:
```
#include <stdio.h>
#include <string.h>

#define WAKA "WakaWaka"

int main() {
        printf("Len %d %s\n", strlen(WAKA), WAKA);
}
```
and compile with gcc -S test.c
From generated assembly output you will see `strlen(WAKA)` is replaced by constant 8

aclk/aclk_rx_msgs.c
underhood: I saw that in Netdata code but I prefer this one as it comes from UUID lib directly but can change it if you really insist

aclk/aclk_rx_msgs.c
underhood: Yep this makes sense I should sanitize for case even if we have control over both endpoints

underhood: done

stelfrag: It can also be a `uuid_unparse_lower(uuid, uuid_str)`;  just before L54

underhood: well same effect, slightly bigger computational overhead, but has advantage of making actual use of `uuid`...
Don't really have attachment to one way or the other

stelfrag (approved):

aclk/aclk_rx_msgs.c
```diff
@@ -15,18 +15,55 @@ static inline int aclk_extract_v2_data(char *payload, char **data)
...
-    const char *start, *end;
+    const char *start, *end, *ptr;
+    char uuid_str[UUID_STR_LEN];
+    uuid_t uuid;
 
-    if(strncmp(payload, ACLK_CLOUD_REQ_V2_PREFIX, strlen(ACLK_CLOUD_REQ_V2_PREFIX))) {
-        errno = 0;
+    errno = 0;
+
+    if(STRNCMP_CONST(cloud_req->data, ACLK_GET_REQ)) {
```
mfundul, line 30: I was thinking about `STRNCMP_CONST(str, const)` and I'm wondering. Why do we need `strncmp()` here?

If `str` is **NOT** a string, then `strlen(const)` does not really guarantee any buffer overflow protection, since we don't know the size of `str`.

If `str` **IS** a string, we don't need `strlen(const)` because they're both strings, and the comparison would stop at the first `\0` anyway.

Finally, maybe `const` should be something else, e.g. `constant` because it's confusing since it's a C language reserved word.

underhood: yep you are right it is not necessary, will remove

underhood: aaah it is because we do not want to compare whole string just checking beginning matches

underhood: meaning for example here:
```
    if(STRNCMP_CONST(cloud_req->data, ACLK_GET_REQ)) {
```
we check string starts with `GET ` not that it is equal to it

underhood: I do agree naming can be changed

stelfrag (approved):

mfundul (approved):

PR #10032, opened 2020-10-06
Fixed the data endpoint to prioritize chart over context if both are present

stelfrag:
##### Summary
When a chart and context is specified in a data query give priority to the chart

##### Component Name
database

##### Test Plan

Before the PR
`http://127.0.0.1:19999/api/v1/data?chart=system.cpu&options=jsonwrap&format=array&after=-10&context=cpu.cpu`

will return a response for the **cpu.cpu**

With the change in this PR the `context=cpu.cpu` will be ignored and return information for the **system.cpu** chart

### DIFF
--- a/web/api/web_api_v1.c
+++ b/web/api/web_api_v1.c
@@ -492,7 +492,7 @@ inline int web_client_api_request_v1_data(RRDHOST *host, struct web_client *w, c
     }
 
     struct context_param  *context_param_list = NULL;
-    if (context) {
+    if (context && !chart) {
         RRDSET *st1;
         uint32_t context_hash = simple_hash(context);
         rrdhost_rdlock(localhost);
--- a/web/api/web_api_v1.c
+++ b/web/api/web_api_v1.c
@@ -513,7 +513,7 @@ inline int web_client_api_request_v1_data(RRDHOST *host, struct web_client *w, c
     }
 
     if (!st && !context_param_list) {
-        if (context) {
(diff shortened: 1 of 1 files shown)


web/api/web_api_v1.c
```diff
@@ -492,7 +492,7 @@ inline int web_client_api_request_v1_data(RRDHOST *host, struct web_client *w, c
     }
 
     struct context_param  *context_param_list = NULL;
-    if (context) {
+    if (context && !chart) {
```
mfundul, line 495: You should do the same in line 516.

mfundul (approved):

underhood:
should we log a warning in this case? `Both Context and Chart specified. Ignoring context`?

stelfrag:
> should we log a warning in this case? `Both Context and Chart specified. Ignoring context`?

This is a bug, the cloud shouldn't specify both

underhood:
@stelfrag yep that is why I want to log warning or error if cloud passes such guery

stelfrag:
> @stelfrag yep that is why I want to log warning or error if cloud passes such guery

I wouldn't want to do it because this fix will probably go in before the cloud fixes the issue and it will flood the log file. 

Every query for every chart if someone uses the cloud will log a message. 

I agree it should eventually go in after the cloud issues a fix so we can catch the (rare) cases where it passes a context and chart.

underhood (approved):

PR #10045, opened 2020-10-08
Rename NETDATA_PORT to NETDATA_LISTENER_PORT

knatsakis:
Because it clashes with a kubernetes defined NETDATA_PORT variable: Our Helm
chart creates a service called netdata which in turn makes kubernetes export a
NETDATA_PORT environment variable with a value like tcp://\<ip\>:19999 which is
not parsable by the netdata agent as a value to the `-p` argument.

This PR should resolve the:
```
netdata ERROR : MAIN : LISTENER: Invalid listen port 0 given. Defaulting
to 19999. (errno 22, Invalid argument)
```

errors seen in Helm installations as reported in netdata/helmchart#161.

##### Additional Information

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -107,8 +107,8 @@ RUN \
 # Install any Python wheels
 RUN pip install /wheels/*
 
-ENV NETDATA_PORT 19999
-EXPOSE $NETDATA_PORT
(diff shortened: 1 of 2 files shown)


thiagoftsm (approved):
LGTM!

ncmans (approved):

Ferroin (approved):

PR #10078, opened 2020-10-15
Create clean

alexmyczko:
allow double builds...
fails otherwise as the clean target of upstream is not clean

##### Test Plan

run debuild, and then run it again

### DIFF
--- a/contrib/debian/clean
+++ b/contrib/debian/clean
@@ -1,0 +1,8 @@
+*/*/*/*/*/*.o
+*/*/*/*/*.o
+*/*/*/*.o
+*/*/*.o
+*/*.o
+tmp/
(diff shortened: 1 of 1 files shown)


ilyam8:
@Ferroin any thoughts on the changes? Do we need them?

alexmyczko:
@ilyam8 if you consider users and packages with "we", yes it's needed. maybe not often it will be built twice in a row, but when it is, this file cleans up leftovers so a second build is possible. if "we" refers to developers of the software only, not building debian packages, the answer would be no.

Ferroin:
Closing this as our official DEB package build infrastructure has been migrated to using CPack, thus this is no longer relevant.

PR #10080, opened 2020-10-15
Create include-binaries

alexmyczko:
goes together with my previous PR


ilyam8:
@Ferroin any thoughts on the changes? Do we need them?

alexmyczko:
@ilyam8 this pr just is an ugly workaround about a defect, including compressed tarball (not even in released source tarballs) but being downloaded at build time. building software is classically really only building software, assuming there is internet, and trying or doing downloads of dependent other sources should be avoided. i hope this can be fixed at some later point...

PR #10100, opened 2020-10-19
Added more robust doucmentation around updates.

Ferroin:
##### Summary

* Add a section to help users figure out which method they originally used to install Netdata.
* Update the description of how the manual update process works.
* Add a fallback command to help users deal with failures using the normal update process for the `kickstart.sh` and `kickstart-static64.sh` install methods.
* Don’t line wrap a long link that is causing problems with syntax highlighting in some editors when manually line-wrapped (it makes the whole document look like one big link).

##### Component Name

area/docs

##### Test Plan

n/a

### DIFF
--- a/packaging/installer/UPDATE.md
+++ b/packaging/installer/UPDATE.md
@@ -31,19 +31,14 @@ you installed it.** Choose from the following list to see the appropriate update
 
 #### How to determine which install method you used
 
-First, figure out where your user configuration files for Netdata are, normally this is `/etc/netdata`
-or `/opt/netdata/etc/netdata`. You can determine this from the system you installed by running the following command:
+First, see [here](https://learn.netdata.cloud/docs/configure/nodes#the-netdata-config-directory) to figure out
+where your user configuration files for Netdata are.
 
-```bash
-curl -sS http://localhost:19999/netdata.conf | grep 'config directory'
-```
-
-The location of your user configuration files is the path on the `config directory` line in the output of that command.
-
-Once you have figured this out, look for a file called `.environment` in this directory. If it is not there, you
-used a package manager to install Netdata, and need to update it through that package manager. If the `.environment` file is present, check
-the contents of the file. If `IS_NETDATA_STATIC_BINARY` is `"yes"`, then you installed using `kickstart-static64.sh`.
-Otherwise you installed using `kickstart.sh`.
+Once you have figured this out, look for a file called `.environment` in this directory (you will need to use
+`ls -a` to see it, as it will not be listed by default by a regular `ls` command). If it is not there, you used
+package manager to install netdata and need to update it through that package manager. If the `environment` file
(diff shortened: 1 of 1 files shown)


packaging/installer/UPDATE.md
```diff
@@ -24,28 +24,55 @@ you installed it.** Choose from the following list to see the appropriate update
...
 -   [macOS](#macos)
 -   [Manual installation from Git](#manual-installation-from-git)
 
+#### How to determine which install method you used
+
+First, figure out where your user configuration files for Netdata are, normally this is `/etc/netdata`
+or `/opt/netdata/etc/netdata`. You can determine this from the system you installed by running the following command:
+
+```bash
+curl -sS http://localhost:19999/netdata.conf | grep 'config directory'
```
manos-saratsis, line 38: @Ferroin if the dashboard is down this won't work but it's ok

Ferroin: Yeah, there’s unfortunately not any good way to do this without involving the dashboard or going through a long list of shell commands. I think it’s safe to assume that most users _probably_ don’t actually need this command, but among those who do they are probably using default settings so it _should_ work in most cases.

manos-saratsis (approved):

packaging/installer/UPDATE.md
joelhans: We could add sentence after this command saying something like, "If the `curl` command failed, try manually navigating to either `/etc/netdata` and `/opt/netdata/etc/netdata` and use `ls` to look for the presence of the `netdata.conf` file."

joelhans (changes requested):
Looking good! Just a few small requests.

packaging/installer/UPDATE.md
```diff
@@ -24,28 +24,55 @@ you installed it.** Choose from the following list to see the appropriate update
...
--   [Pre-built static binary for 64-bit systems
-    (`kickstart-static64.sh`)](#pre-built-static-binary-for-64-bit-systems-kickstart-static64sh)
+-   [Pre-built static binary for 64-bit systems (`kickstart-static64.sh`)](#pre-built-static-binary-for-64-bit-systems-kickstart-static64sh)
 -   [Docker](#docker)
 -   [macOS](#macos)
 -   [Manual installation from Git](#manual-installation-from-git)
 
+#### How to determine which install method you used
+
+First, figure out where your user configuration files for Netdata are, normally this is `/etc/netdata`
```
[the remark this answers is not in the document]
Ferroin: I actually did not know we had a section elsewhere dedicated to the config directory.

joelhans (dismissed):

packaging/installer/UPDATE.md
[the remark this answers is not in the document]
joelhans: It's one of the newer docs in my push from last month. Do let me know if you think there's any more important things to add to that doc (we'll make a separate PR, of course!).

joelhans (approved):

Issue #9202, opened 2020-05-28
netdata-claim.sh: command not found


manval70:
When creating a bug report please: When run script to claim a node

##### OS / Environment

```
Ubuntu Server 18.04 LTS
```

##### Netdata version

 1.9

##### Steps To Reproduce

1. ...
2. ...

prologic:
Looks like we don't include the script in the source dist:

```#!sh
~/Netdata/netdata
 (static64_plus_ebpf) 0
$ find . | grep 'netdata-claim.sh'
./claim/netdata-claim.sh
./claim/netdata-claim.sh.in

$ sift 'netdata-claim.sh'
netdata.spec.in:449:%{_sbindir}/netdata-claim.sh
packaging/docker/Dockerfile:49:    mv /usr/sbin/netdata-claim.sh    /app/usr/sbin/ && \
packaging/installer/netdata-uninstaller.sh:493:  rm_file "/usr/sbin/netdata-claim.sh"
claim/Makefile.am:7:    netdata-claim.sh \
claim/Makefile.am:14:    netdata-claim.sh \
claim/Makefile.am:18:    netdata-claim.sh.in \
claim/claim.c:78:              "exec netdata-claim.sh %s -hostname=%s -id=%s -url=%s -noreload %s",
claim/claim.c:86:    info("Executing agent claiming command 'netdata-claim.sh'");
claim/README.md:51:sudo netdata-claim.sh -token=TOKEN -rooms=ROOM1,ROOM2 -url=https://app.netdata.cloud
claim/README.md:78:netdata-claim.sh -token=TOKEN -rooms=ROOM1,ROOM2 -url=https://app.netdata.cloud
claim/README.md:96:docker exec -it netdata netdata-claim.sh -token=TOKEN -rooms=ROOM1,ROOM2 -url=https://app.netdata.cloud
claim/README.md:165:You can now move on to claiming. When you claim with the `netdata-claim.sh` script, add the `-proxy=` parameter and
claim/README.md:169:sudo netdata-claim.sh -token=MYTOKEN1234567 -rooms=room1,room2 -url=https://app.netdata.cloud -proxy=socks5h://203.0.113.0:1080
claim/README.md:283:sudo netdata-claim.sh -token=TOKEN -rooms=ROOM1,ROOM2 -url=https://app.netdata.cloud -id=$(uuidgen)
claim/README.md:305:A Space's administrator can claim an Agent by directly calling the `netdata-claim.sh` script either with root privileges
[...]

PR #10103, opened 2020-10-19
Remove Docker example from update docs and add section to claim troubleshooting

joelhans:
##### Summary

A mash-up of two small fixes to the docs.

First, a new section to address some [ongoing community feedback](https://github.com/netdata/netdata/issues/9202#issuecomment-712097853) for the `netdata-claim.sh` script not being in the user's PATH.

Second, removing the example `docker run` command from the update page, as it conflicts with the up-to-date examples found in the official Docker installation page. Let's direct users to that page instead of duplicating the command.

##### Component Name

area/docs
area/claim
area/packaging

### DIFF
--- a/claim/README.md
+++ b/claim/README.md
@@ -198,9 +198,11 @@ Use these keys and the information below to troubleshoot the ACLK.
 
 If you run the claiming script and see a `command not found` error, you either installed Netdata in a non-standard
 location or are using an unsupported package. If you installed Netdata in a non-standard path using the `--install`
-option, you need to update your `$PATH`, or `cd` to the path and run `netdata-claim.sh` there. If you are using an
-unsupported package, such as a third-party `.deb`/`.rpm` package provided by your distribution, please remove that
-package and reinstall using our [recommended kickstart script](/docs/get#install-the-netdata-agent).
+option, you need to update your `$PATH` or run `netdata-claim.sh` using the full path. For example, if you installed
+Netdata to `/opt/netdata`, use `/opt/netdata/bin/netdata-claim.sh` to run the claiming script.
+
+If you are using an unsupported package, such as a third-party `.deb`/`.rpm` package provided by your distribution,
+please remove that package and reinstall using our [recommended kickstart script](/docs/get#install-the-netdata-agent).
 
 #### Claiming on older distributions (Ubuntu 14.04, Debian 8, CentOS 6)
 
--- a/claim/README.md
+++ b/claim/README.md
@@ -194,6 +194,20 @@ might be having with the ACLK or claiming process.
 
 Use these keys and the information below to troubleshoot the ACLK.
 
+#### bash: netdata-claim.sh: command not found
+
+If you run the claiming script and see a `command not found` error, the claiming script is not in your system's `$PATH`.
+Netdata expects this file at `/usr/sbin/`, but some systems may have installed the script in `/opt/netdata/usr/sbin/` or
+`/opt/netdata/bin/`.
+
+Check those locations for the existence of the `netdata-claim.sh` file. When you find it, run the following commands to
+claim your node.
+
+```bash
+sudo chmod +x netdata-claim.sh
+sudo ./netdata-claim.sh -token=TOKEN -rooms=ROOM1,ROOM2 -url=https://app.netdata.cloud
+```
+
 #### Claiming on older distributions (Ubuntu 14.04, Debian 8, CentOS 6)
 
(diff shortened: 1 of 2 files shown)


claim/README.md
```diff
@@ -194,6 +194,20 @@ might be having with the ACLK or claiming process.
 
 Use these keys and the information below to troubleshoot the ACLK.
 
+#### bash: netdata-claim.sh: command not found
+
+If you run the claiming script and see a `command not found` error, the claiming script is not in your system's `$PATH`.
+Netdata expects this file at `/usr/sbin/`, but some systems may have installed the script in `/opt/netdata/usr/sbin/` or
```
Ferroin, line 200: Netdata Cloud does, Netdata itself should be looking in the right place and if it isn’t then that’s a bug in netdata.

claim/README.md
```diff
@@ -194,6 +194,20 @@ might be having with the ACLK or claiming process.
...
+`/opt/netdata/bin/`.
+
+Check those locations for the existence of the `netdata-claim.sh` file. When you find it, run the following commands to
+claim your node.
+
+```bash
+sudo chmod +x netdata-claim.sh
```
Ferroin, line 207: Marking it executable should never be needed, same as the other comment if this _is_ necessary, it’s a bug and should be handled as such.

claim/README.md
joelhans: Just read the long thread on Slack about this. Also related: https://github.com/netdata/netdata/issues/9202. I can't tell if these users are encountering a bug or not.

Perhaps the better route is to document that you may see this error if you installed Netdata in a non-standard location or are using an unsupported package. For the former, you can update your PATH or just find the script and run it from that location. For the latter, uninstall and reinstall using kickstart.

How does that sound?

claim/README.md
joelhans: Good call, thanks—I'll probably just remove this block entirely if we go the route I described in your other comment.

claim/README.md
Ferroin: That sounds like an excellent approach.

OdysLam: @Ferroin you have mentioned that you believe that we shouldn't makes changes to PATH and instead we should change the documents to point to absolute paths. Perhaps this is a good place to start?

Ferroin: Possibly. The problem is that right now we don’t have absolute paths that work everywhere.

And it’s not an issue if _users_ decide to update the contents of `$PATH`, the issue that potentially arises is _us_ modifying it, because we can’t count on the user not changing it in some other way.

joelhans:
@OdysLam && @Ferroin Any more changes to request here, or can we merge these changes?

OdysLam (dismissed):

claim/README.md
```diff
@@ -194,6 +194,14 @@ might be having with the ACLK or claiming process.
 
 Use these keys and the information below to troubleshoot the ACLK.
 
+#### bash: netdata-claim.sh: command not found
+
+If you run the claiming script and see a `command not found` error, you either installed Netdata in a non-standard
+location or are using an unsupported package. If you installed Netdata in a non-standard path using the `--install`
+option, you need to update your `$PATH`, or `cd` to the path and run `netdata-claim.sh` there. If you are using an
```
Ferroin, line 201: Strictly speaking, the user just has to use the absolute path to the command. For example, if they installed to `/opt/netdata` (as the static install does), they would need to run `/opt/netdata/bin/netdata-claim.sh` instead of `netdata-claim.sh`.

joelhans: Changed it. Thanks!

Ferroin (approved):

claim/README.md
```diff
@@ -194,6 +194,16 @@ might be having with the ACLK or claiming process.
...
 Use these keys and the information below to troubleshoot the ACLK.
 
+#### bash: netdata-claim.sh: command not found
+
+If you run the claiming script and see a `command not found` error, you either installed Netdata in a non-standard
+location or are using an unsupported package. If you installed Netdata in a non-standard path using the `--install`
+option, you need to update your `$PATH` or run `netdata-claim.sh` using the full path. For example, if you installed
+Netdata to `/opt/netdata`, use `/opt/netdata/bin/netdata-claim.sh` to run the claiming script.
+
+If you are using an unsupported package, such as a third-party `.deb`/`.rpm` package provided by your distribution,
```
OdysLam, line 204: Since we do provide binbary packages for these, why we don't nudge people to use those?

OdysLam (approved):

PR #10107, opened 2020-10-20
Kubernetes labels

ilyam8:
##### Additional Information

`cgroup-name.sh` return value (OR):

- `<NAME>`
- `<NAME><SPACE><LABELS>`, `<LABELS>` format is `name="value",name="value",...`

Pod/container level cgroups labels
- pod 

| Name                  |  Value                                                 |
| :-------------------- | :----------------------------------------------------- |
| `k8s_cluster_id`      | `kube_system` _namespace.metadata.uid_                 |
| `k8s_namespace`       | _pod.metadata.namespace_                               |
| `k8s_pod_name`        | _pod.metadata.name_                                    |
| `k8s_pod_uid`         | _pod.metadata.uid_                                     |
| ~~`pod_label_*`~~     | _pod.metadata.labels_                                  |
| `k8s_controller_kind` | _pod.metadata.ownerReferences[].kind_                  |
| `k8s_controller_name` | _pod.metadata.ownerReferences[].name_                  |
| `k8s_node_name`       | _pod.spec.nodeName_                                    |
| `k8s_kind`            | pod                                                    |

- container

**pod level cgroups labels +**

| Name             |  Value                                                |
| :--------------- | :---------------------------------------------------- |
| `k8s_container_name` | _pod.status.containerStatuses.name_                   |
| `k8s_container_id`   | _pod.status.containerStatuses.containerID_            |
| `k8s_kind`           | container                                             |

---

TODO:
  - [x] add `kind` and `name` values from `ownerReferences` if `controller: true` 

<details>
<summary>kubectl explain pod.metadata.ownerReferences</summary>

```cmd
0 node.d.plugin (master)$ kubectl explain pod.metadata.ownerReferences
KIND:     Pod
VERSION:  v1

RESOURCE: ownerReferences <[]Object>

DESCRIPTION:
     List of objects depended by this object. If ALL objects in the list have
     been deleted, this object will be garbage collected. If this object is
     managed by a controller, then an entry in this list will point to this
     controller, with the controller field set to true. There cannot be more
     than one managing controller.

     OwnerReference contains enough information to let you identify an owning
     object. An owning object must be in the same namespace as the dependent, or
     be cluster-scoped, so there is no namespace field.

FIELDS:
   apiVersion	<string> -required-
     API version of the referent.

   blockOwnerDeletion	<boolean>
     If true, AND if the owner has the "foregroundDeletion" finalizer, then the
     owner cannot be deleted from the key-value store until this reference is
     removed. Defaults to false. To set this field, a user needs "delete"
     permission of the owner, otherwise 422 (Unprocessable Entity) will be
     returned.

   controller	<boolean>
     If true, this reference points to the managing controller.

   kind	<string> -required-
     Kind of the referent. More info:
     https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds

   name	<string> -required-
     Name of the referent. More info:
     http://kubernetes.io/docs/user-guide/identifiers#names

   uid	<string> -required-
     UID of the referent. More info:
     http://kubernetes.io/docs/user-guide/identifiers#uids
```

</details>

### DIFF
--- a/database/rrdlabels.c
+++ b/database/rrdlabels.c
@@ -126,6 +126,17 @@ struct label *add_label_to_list(struct label *l, char *key, char *value, LABEL_S
     return lab;
 }
 
+void update_label_list(struct label **labels, struct label *new_labels)
+{
+    free_label_list(*labels);
+
+    while (new_labels != NULL)
+    {
+        *labels = add_label_to_list(*labels, new_labels->key, new_labels->value, new_labels->label_source);
+        new_labels = new_labels->next;
+    }
+}
+
 int label_list_contains(struct label *head, struct label *check)
 {
     while (head != NULL)
--- a/database/rrdlabels.c
+++ b/database/rrdlabels.c
@@ -129,6 +129,7 @@ struct label *add_label_to_list(struct label *l, char *key, char *value, LABEL_S
 void update_label_list(struct label **labels, struct label *new_labels)
 {
     free_label_list(*labels);
+    *labels = NULL;
 
     while (new_labels != NULL)
     {
--- a/database/rrdlabels.c
+++ b/database/rrdlabels.c
@@ -1,0 +1,159 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#define NETDATA_RRD_INTERNALS
+#include "rrd.h"
+
+char *translate_label_source(LABEL_SOURCE l) {
+    switch (l) {
+        case LABEL_SOURCE_AUTO:
+            return "AUTO";
+        case LABEL_SOURCE_NETDATA_CONF:
+            return "NETDATA.CONF";
+        case LABEL_SOURCE_DOCKER :
+            return "DOCKER";
+        case LABEL_SOURCE_ENVIRONMENT  :
+            return "ENVIRONMENT";
+        case LABEL_SOURCE_KUBERNETES :
+            return "KUBERNETES";
+        default:
+            return "Invalid label source";
+    }
+}
+
+int is_valid_label_value(char *value) {
+    while(*value) {
+        if(*value == '"' || *value == '\'' || *value == '*' || *value == '!') {
+            return 0;
+        }
+
+        value++;
+    }
+
+    return 1;
+}
+
+int is_valid_label_key(char *key) {
+    //Prometheus exporter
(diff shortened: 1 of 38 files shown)


database/rrdlabels.c
```diff
@@ -129,6 +129,7 @@ struct label *add_label_to_list(struct label *l, char *key, char *value, LABEL_S
 void update_label_list(struct label **labels, struct label *new_labels)
 {
     free_label_list(*labels);
+    *labels = NULL;
 
     while (new_labels != NULL)
```
ilyam8, line 134: is this line reachable?

ilyam8: oh, new_labels! 🙈

vlvkobal (dismissed):
I haven't reviewed it as a whole, but I reviewed it while writing my part and later reviewed separate commits. I also tested it a lot.

vlvkobal:
It seems that Travis and LGTM raised alarms that are not connected with the PR itself.

stelfrag:
When configuring
```
[exporting:global]
send configured labels = yes
send automatic labels = yes
update every = 3

[opentsdb:my_instance0]
enabled = yes
destination = 192.168.1.138:4242
```
 It will block during shutdown while reporting 

```
EXPORTING : RRDHOST 'futura' should be read-locked, but it is not, at function format_host_labels_opentsdb_telnet() at line 155 of file 'exporting/opentsdb/opentsdb.c'
```
in the logfile

mfundul (approved):

vlvkobal (approved):

thiagoftsm (approved):

database/rrdlabels.c
```diff
@@ -0,0 +1,181 @@
...
+
+    return str;
+}
+
+struct label *create_label(char *key, char *value, LABEL_SOURCE label_source)
+{
+    size_t key_len = strlen(key), value_len = strlen(value);
+    size_t n = sizeof(struct label) + key_len + 1 + value_len + 1;
+    struct label *result = callocz(1,n);
+    if (result != NULL) {
```
underhood, line 88: callocz checks for NULL already and fatals if OOM

stelfrag (approved):

PR #10110, opened 2020-10-21
Fixed compile error in CENTOS 6

### LINKED ISSUE #10106: Netdata update failed - Centos 6
Hi.

I tried to update netdata on centos 6, but process failed.

I attached a screenshot.

What now ?
Zrzut ekranu 2020-10-20 104246

stelfrag:
Fixes #10106
##### Summary
Fix the compile error in centos 6 (when trying to use a forward `typedef structure` definition)

##### Component Name
database

##### Test Plan
- Compiling in centos 6 is now successful

### DIFF
--- a/web/api/formatters/rrd2json.h
+++ b/web/api/formatters/rrd2json.h
@@ -53,12 +53,6 @@
 extern void rrd_stats_api_v1_chart(RRDSET *st, BUFFER *wb);
 extern void rrdr_buffer_print_format(BUFFER *wb, uint32_t format);
 
-typedef struct context_param {
-    RRDDIM *rd;
-    time_t first_entry_t;
-    time_t last_entry_t;
-} CONTEXT_PARAM;
-
 extern int rrdset2anything_api_v1(
           RRDSET *st
         , BUFFER *wb
--- a/database/rrd.h
+++ b/database/rrd.h
@@ -13,7 +13,6 @@ typedef struct rrddimvar RRDDIMVAR;
 typedef struct rrdcalc RRDCALC;
 typedef struct rrdcalctemplate RRDCALCTEMPLATE;
 typedef struct alarm_entry ALARM_ENTRY;
-typedef struct context_param CONTEXT_PARAM;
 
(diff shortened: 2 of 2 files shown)


vlvkobal (approved):

web/api/formatters/rrd2json.h
```diff
@@ -53,12 +53,6 @@
 extern void rrd_stats_api_v1_chart(RRDSET *st, BUFFER *wb);
 extern void rrdr_buffer_print_format(BUFFER *wb, uint32_t format);
 
-typedef struct context_param {
```
mfundul, line 56: I've played a bit with the code and you don't really have to move anything to `rrd.h`.

The only issue is the `typedef` redefinition, which is not compliant with anything < C11. The minimum fix is to remove the `typedef` from the `struct` definition here, and keep `rrd.h` as is:
```
struct context_param {
Attach files by dragging & dropping, selecting or pasting them.
    RRDDIM *rd;
    time_t first_entry_t;
    time_t last_entry_t;
}; 
```

mfundul (changes requested):

web/api/formatters/rrd2json.h
mfundul: Alternatively, if we should put the structure in `rrd.h`, then we should do it consistently as such:
-  keep `typedef struct context_param CONTEXT_PARAM;` in the forward typedefs.
- Add `struct context_param;` in the forward declarations but do not add the definition there.
- Skip ` #include <time.h>` as it will be included by ` #include "../daemon/common.h"`
- Add the definition in the main body of `rrd.h`:
```
struct context_param {
Attach files by dragging & dropping, selecting or pasting them.
    RRDDIM *rd;
    time_t first_entry_t;
    time_t last_entry_t;
}; 
```

stelfrag: Yes as discussed 2nd option is better will do that

mfundul (approved):

vlvkobal (approved):

PR #10112, opened 2020-10-21
Added new data query option "allow_past"

stelfrag:
##### Summary
Adds a new query option `allow_past` for the data endpoint that will allow the after parameter to extend before the first timestamp stored.  The entries that do not exist will be returned as null

##### Component Name
database

##### Test Plan
- Start agent will no history (e.g. if using dbengine memory mode, delete the dbengine files)
- Run a query such as: 
  `/api/v1/data?chart=system.cpu`
  - This should return just a few entries (the number of seconds the agent is up and running)
- Run the query 
  `/api/v1/data?chart=system.cpu&options=allow_past`
   - You should get 600 entries returned, most of them null

### DIFF
--- a/web/api/netdata-swagger.json
+++ b/web/api/netdata-swagger.json
@@ -140,6 +140,17 @@
             }
           },
           {
+            "name": "context",
+            "in": "query",
+            "description": "The context of the chart as returned by the /charts call.",
+            "required": false,
+            "allowEmptyValue": false,
+            "schema": {
+              "type": "string",
+              "format": "as returned by /charts"
+            }
+          },
+          {
             "name": "dimension",
             "in": "query",
             "description": "Zero, one or more dimension ids or names, as returned by the /chart call, separated with comma or pipe. Netdata simple patterns are supported.",
@@ -156,7 +167,7 @@
           {
             "name": "after",
             "in": "query",
-            "description": "This parameter can either be an absolute timestamp specifying the starting point of the data to be returned, or a relative number of seconds (negative, relative to parameter: before). Netdata will assume it is a relative number if it is less that 3 years (in seconds). Netdata will adapt this parameter to the boundaries of the round robin database. The default is the beginning of the round robin database (i.e. by default netdata will attempt to return data for the entire database).",
+            "description": "This parameter can either be an absolute timestamp specifying the starting point of the data to be returned, or a relative number of seconds (negative, relative to parameter: before). Netdata will assume it is a relative number if it is less that 3 years (in seconds). If not specified the default is -600 seconds. Netdata will adapt this parameter to the boundaries of the round robin database unless the allow_past option is specified.",
             "required": true,
             "allowEmptyValue": false,
             "schema": {
@@ -273,7 +284,8 @@
                   "unaligned",
                   "match-ids",
                   "match-names",
-                  "showcustomvars"
+                  "showcustomvars",
+                  "allow_past"
                 ]
               },
               "default": [
@@ -478,7 +490,7 @@
               "oneOf": [
                 {
                   "type": "string",
-                  "enum" : [
+                  "enum": [
                     "green",
                     "brightgreen",
                     "yellow",
@@ -520,7 +532,7 @@
               "oneOf": [
                 {
                   "type": "string",
-                  "enum" : [
+                  "enum": [
                     "green",
                     "brightgreen",
                     "yellow",
@@ -551,7 +563,7 @@
               "oneOf": [
                 {
                   "type": "string",
-                  "enum" : [
+                  "enum": [
                     "green",
                     "brightgreen",
                     "yellow",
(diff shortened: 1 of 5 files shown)


web/api/queries/query.c
```diff
@@ -1581,7 +1582,11 @@ RRDR *rrd2rrdr(
     rrd_update_every = st->update_every;
     absolute_period_requested = rrdr_convert_before_after_to_absolute(&after_requested, &before_requested,
                                                                       rrd_update_every, first_entry_t,
-                                                                      last_entry_t);
+                                                                      last_entry_t, options);
+    if (options & RRDR_OPTION_ALLOW_PAST)
+        if (first_entry_t > after_requested)
+            first_entry_t = after_requested;
+
```
mfundul, line 1589: This code should probably move after `rrdr_convert_before_after_to_absolute()` has been called, because only then is `after_requested` up-to-date for the comparison to make sense. `rrdr_convert_before_after_to_absolute()` can even swap `before_requested` and `after_requested` values.

mfundul (changes requested):

thiagoftsm:
Hi @stelfrag ,

Please, can you also change the [swagger file](https://github.com/netdata/netdata/blob/master/web/api/netdata-swagger.json) and [yaml file](https://github.com/netdata/netdata/blob/master/web/api/netdata-swagger.yaml) with this PR for we do not have options without documentation?

Thank you!

web/api/queries/query.c
stelfrag: It is called after rrdr_convert_before_after_to_absolute() in L1583. 

Now there are other calls to rrdr_convert_before_after_to_absolute() when dbengine is enabled, but I don't see those affecting the `after_requested` and `before_requested` once those are set to absolute values.

stelfrag:
> Hi @stelfrag ,
> 
> Please, can you also change the [swagger file](https://github.com/netdata/netdata/blob/master/web/api/netdata-swagger.json) and [yaml file](https://github.com/netdata/netdata/blob/master/web/api/netdata-swagger.yaml) with this PR for we do not have options without documentation?
> 
> Thank you!

good point !!! :+1:

web/api/queries/query.c
mfundul: Ah yes, funny how I missed that.

thiagoftsm:
```
Run a query such as:
/api/v1/data?chart=system.cpu

    This should return just a few entries (the number of seconds the agent is up and running)

Run the query
/api/v1/data?chart=system.cpu&options=allow_past

    You should get 600 entries returned, most of them null
```

@stelfrag I tested now with `memory mode = ram` and `memory mode = dbengine`, for this last I removed everything inside `/var/cache/netdata/dbengine` before to test, and I could not see the results that you described.  

I execute the following requests to simplify the comparison:

```
$ curl -k -o without.json https://localhost:19999/api/v1/data?chart=system.cpu ; curl -k -o with.json https://localhost:19999/api/v1/data?chart=system.cpu&options=allow_past
$ wc -l with*
```

stelfrag:
> ```
> Run a query such as:
> /api/v1/data?chart=system.cpu
> 
>     This should return just a few entries (the number of seconds the agent is up and running)
> 
> Run the query
> /api/v1/data?chart=system.cpu&options=allow_past
> 
>     You should get 600 entries returned, most of them null
> ```
> 
> @stelfrag I tested now with `memory mode = ram` and `memory mode = dbengine`, for this last I removed everything inside `/var/cache/netdata/dbengine` before to test, and I could not see the results that you described.
> 
> I executed the following requests to simplify the comparison:
> 
> ```
> $ curl -k -o without.json https://localhost:19999/api/v1/data?chart=system.cpu ; curl -k -o with.json https://localhost:19999/api/v1/data?chart=system.cpu&options=allow_past
> $ wc -l with*
> ```

Ah, yes I have been tricked quite a few times by this. 

Please escape the & because you sent the command to the background :smile:

web/api/netdata-swagger.json
```diff
@@ -156,7 +167,7 @@
           {
             "name": "after",
             "in": "query",
-            "description": "This parameter can either be an absolute timestamp specifying the starting point of the data to be returned, or a relative number of seconds (negative, relative to parameter: before). Netdata will assume it is a relative number if it is less that 3 years (in seconds). Netdata will adapt this parameter to the boundaries of the round robin database. The default is the beginning of the round robin database (i.e. by default netdata will attempt to return data for the entire database).",
+            "description": "This parameter can either be an absolute timestamp specifying the starting point of the data to be returned, or a relative number of seconds (negative, relative to parameter: before). Netdata will assume it is a relative number if it is less that 3 years (in seconds). If not specified the default is -600 seconds. Netdata will adapt this parameter to the boundaries of the round robin database unless the allow_past option is specified.",
```
mfundul, line 170: How is the default -600? I thought it was still the beginning of the database.

stelfrag: This changed back in March 2020 in the code.  The docs specified a default of -600, the code didn't.

mfundul: So then I guess we have to update README.md as well then?

mfundul (approved):

thiagoftsm (approved):
After to add quotes for curl :eyes: , everything worked as expected.
Thank you @stelfrag !

PR #10139, opened 2020-10-26
Migrate metadata log to SQLite

stelfrag:
##### Summary
Migrate metadata to SQLite

##### Component Name
database

##### Test Plan
- Use this PR on an agent with metadata log support
  - A new database file `netdata-meta.db` will be created under the default cache directory `var/cache/netdata`

- Use cases
  - Parent agent with no metadata support (version prior to 1.22)
    - metrics should be available after migration
  - Parent agent with metadata support (version 1.22 and later)
    - metrics should be available after migration
  - Parent agent any version
    - Children with  (version prior to 1.22)
    - Children with  (version 1.22 and later)


database/sqlite/sqlite_functions.c
```diff
@@ -0,0 +1,949 @@
...
+    //   sqlite3_create_function(db, "uncompress", 1, SQLITE_ANY , 0, _uncompress, 0, 0);
+
+    return 0;
+}
+
+/*
+ * Close the sqlite database
+ */
+
+int sql_close_database()
```
mfundul, line 135:
```suggestion
int sql_close_database(void)
```

database/sqlite/sqlite_functions.c
```diff
@@ -0,0 +1,949 @@
...
+        if (rc != SQLITE_OK)
+            error("Error %d while closing the SQLite database", rc);
+    }
+    return 0;
+}
+
+/*
+ * Return the database size in MiB
+ */
+int sql_database_size()
```
mfundul, line 164:
```suggestion
int sql_database_size(void)
```

database/sqlite/sqlite_functions.c
```diff
@@ -0,0 +1,949 @@
...
+        return 0;
+
+    if (!res) {
+        rc = sqlite3_prepare_v2(
+            db, "select chart_id, type, id, name from chart where host_id = @host;", -1, &res, 0);
+        if (rc != SQLITE_OK)
+            return 0;
+    }
+
+    rc = sqlite3_bind_blob(res, 1, &host->host_uuid, 16, SQLITE_TRANSIENT);
```
mfundul, line 288: Let's try and avoid the magic numbers from now on. We could use something like:
```suggestion
    rc = sqlite3_bind_blob(res, 1, &host->host_uuid, sizeof(host->host_uuid), SQLITE_TRANSIENT);
```
Or use a `#define` if we like it better.

database/sqlite/sqlite_functions.h
```diff
@@ -0,0 +1,48 @@
...
+
+#define SQL_STORE_CHART "insert or replace into chart (chart_id, host_id, type, id, " \
+    "name, family, context, title, unit, plugin, module, priority, update_every , chart_type , memory_mode , " \
+    "history_entries) values (?1,?2,?3,?4,?5,?6,?7,?8,?9,?10,?11,?12,?13,?14,?15,?16);"
+
+#define SQL_STORE_DIMENSION                                                                                           \
+    "INSERT OR REPLACE into dimension (dim_id, chart_id, id, name, multiplier, divisor , algorithm) values (?0001,?0002,?0003,?0004,?0005,?0006,?0007);"
+
+extern int sql_init_database();
+extern int sql_close_database();
```
mfundul, line 30:
```suggestion
extern int sql_init_database(void);
extern int sql_close_database(void);
```

mfundul (changes requested):

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -3,460 +3,479 @@
...
+//    UNUSED(ctx);
+//    RRDDIM *rd;
+//    RRDSET *st;
+//    RRDHOST *host;
+//    uint8_t empty_chart;
+//    rd = metalog_get_dimension_from_uuid(ctx, metric_uuid);
+//    if (!rd) { /* in the case of legacy UUID convert to multihost and try again */
+//        uuid_t multihost_uuid;
+//
+//        rrdeng_convert_legacy_uuid_to_multihost(ctx->rrdeng_ctx->machine_guid, metric_uuid, &multihost_uuid);
```
mfundul, line 417: This function is skipped, but it is essential otherwise the dbengine UUIDs from different legacy instances collide.

database/sqlite/sqlite_functions.c
```diff
@@ -0,0 +1,778 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include "sqlite_functions.h"
+
+#define ENABLE_CACHE_CHARTS 1
+#define ENABLE_CACHE_DIMENSIONS 1
+
+
+#define error_report(x, args...) { errno = 0; error(x, ##args); }
```
mfundul, line 9: We could put this in `libnetdata/log/log.h` and reuse it!

database/sqlite/sqlite_functions.c
```diff
@@ -0,0 +1,778 @@
...
+static uint32_t page_count;
+static uint32_t free_page_count;
+
+/*
+ * Database parameters
+ */
+uint32_t sqlite_disk_quota_mb; // quota specified in the database
+uint32_t desired_pages = 0;
+
+static int execute_insert(sqlite3_stmt *res)
```
mfundul, line 47: It appears we are randomly dropping "INSERT" statements if the DB is busy and the timeout is reached. This will corrupt the database. How do we fix this?

mfundul: I would say retry forever, this is what we were doing with netdata custom DBs before anyways.

CMakeLists.txt
```diff
@@ -597,6 +597,8 @@ set(RRD_PLUGIN_FILES
         database/rrdsetvar.h
         database/rrdvar.c
         database/rrdvar.h
+        database/sqlite/sqlite_functions.c
+        database/sqlite/sqlite_functions.h
```
mfundul, line 601: I think those are missing:
```
database/sqlite/sqlite3.c
database/sqlite/sqlite3.h
```

database/sqlite/sqlite_functions.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include "../../daemon/common.h"
+#include "sqlite3.h"
+
+
+/*
+ * Initialize a database
+ */
+
+#define SQLITE_WAL_SIZE (8 * 1024 * 1024 * 1024)
```
mfundul, line 14: This is unused, but if we ever use it, it can't be 8GiB :)

stelfrag: will remove

CMakeLists.txt
stelfrag: Added will appear in the next push

database/sqlite/sqlite_functions.h
```diff
@@ -0,0 +1,74 @@
...
+#include "../../daemon/common.h"
+#include "sqlite3.h"
+
+
+/*
+ * Initialize a database
+ */
+
+#define SQLITE_SELECT_MAX   (10)        // Max select retries
+#define SQLITE_INSERT_MAX   (10)        // Max select retries
```
underhood, line 15: minor comment copy&paste

database/sqlite/sqlite_functions.c
```diff
@@ -0,0 +1,1224 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include "sqlite_functions.h"
+
+#define ENABLE_CACHE_CHARTS 1
+#define ENABLE_CACHE_DIMENSIONS 1
+
+const char *database_config[] = {
+    "PRAGMA auto_vacuum=incremental; PRAGMA synchronous=1 ; PRAGMA journal_mode=WAL; PRAGMA temp_store=MEMORY;",
+    "PRAGMA journal_size_limit=17179869184;",
```
mfundul, line 10: Well, 17GB of Journal size is a little bit much. How can we make sure this works in embedded devices?

database/sqlite/sqlite_functions.h
stelfrag: Will fix

database/rrddim.c
```diff
@@ -232,21 +232,6 @@ RRDDIM *rrddim_add_custom(RRDSET *st, const char *id, const char *name, collecte
...
         rc += rrddim_set_multiplier(st, rd, multiplier);
         rc += rrddim_set_divisor(st, rd, divisor);
-        if (!is_archived && rrddim_flag_check(rd, RRDDIM_FLAG_ARCHIVED)) {
-            rd->state->collect_ops.init(rd);
-            rrddim_flag_clear(rd, RRDDIM_FLAG_ARCHIVED);
-            rrddimvar_create(rd, RRDVAR_TYPE_CALCULATED, NULL, NULL, &rd->last_stored_value, RRDVAR_OPTION_DEFAULT);
-            rrddimvar_create(rd, RRDVAR_TYPE_COLLECTED, NULL, "_raw", &rd->last_collected_value, RRDVAR_OPTION_DEFAULT);
-            rrddimvar_create(rd, RRDVAR_TYPE_TIME_T, NULL, "_last_collected_t", &rd->last_collected_time.tv_sec, RRDVAR_OPTION_DEFAULT);
-            calc_link_to_rrddim(rd);
-        }
```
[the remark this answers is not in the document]
stelfrag: True, needs to return to cover that case

database/sqlite/sqlite_functions.c
stelfrag: It's wrong -- I will switch to 16 or 32MB which is enough for the transactions we have at this stage for the WAL

database/sqlite/sqlite_functions.c
```diff
@@ -0,0 +1,1224 @@
...
+    "CREATE TABLE IF NOT EXISTS dimension_active(dim_id blob primary key, date_created int);",
+    "CREATE TABLE IF NOT EXISTS metadata_migration(filename text, file_size, date_created int);",
+    "CREATE INDEX IF NOT EXISTS ind_host on chart(host_id);",
+    "CREATE INDEX IF NOT EXISTS ind_chart on dimension(chart_id);",
+
+    "delete from chart_active;",
+    "delete from dimension_active;",
+
+    "delete from chart where chart_id not in (select chart_id from dimension);",
+    "delete from host where host_id not in (select host_id from chart);",
```
mfundul, line 28: Are these going to take a long time with a large database? Maybe we should use them as part of a consistency check operation when a problem is detected.

database/sqlite/sqlite_functions.c
mfundul: The worst case is probably going to be the huge transaction of parsing a metadata log file. Does this fit in 32MB?

database/sqlite/sqlite_functions.c
stelfrag: These are part of the cleanup due to rotation (cleanup during startup). I don't expect those to take a long time. 

I will run a test with an agent with 2 million dimensions and 200K charts to compare how the query plan behaves on

```
delete from chart where chart_id not in (select chart_id from dimension);

vs

delete from chart where chart_id not in (select distinct(chart_id) from dimension);
```
If both take a long time (as in more than a couple of seconds) we can test doing it when a dimension is deleted with a trigger. If that is heavy, we can just record chart_ids using a delete trigger on the dimension to a new table `chart_to_check` and then test if 

`delete from chart where chart_id not in (select c.chart_id from dimension d, chart_to_check c where d.chart_id = c.chart_id);`

behaves better. 

We will need to add vacuum logic at some point especially if we have a lot of hosts that appear just once and then deleted to reclaim space (it won't be a problem if they are deleted and reappear though as the space will be reused)

database/sqlite/sqlite_functions.c
stelfrag: I will set it to 16MB. The largest size I observed by processing a single metadata logfile is about 2MB (each metadata logfile is  one transaction)

database/engine/metadata_log/metadatalogapi.c
```diff
@@ -3,459 +3,10 @@
...
-    }
-    st = rd->rrdset;
-    host = st->rrdhost;
-
-    error_with_guid(metric_uuid, "Host - Chart - Dimension are the below:");
-    error("%s %s %s.", host->hostname, st->id, rd->id);
-
-    if (unlikely(host->rrd_memory_mode != RRD_MEMORY_MODE_DBENGINE))
-        error_with_guid(metric_uuid, "UUID does not belong to RRD_MEMORY_MODE_DBENGINE.");
+    delete_dimension_uuid(rd->state->metric_uuid);
```
underhood, line 8: why not call this directly?

libnetdata/log/log.h
```diff
@@ -44,6 +44,7 @@ extern "C" {
 #define D_ACLK              0x0000000200000000
 #define D_METADATALOG       0x0000000400000000
 #define D_GUIDLOG           0x0000000800000000
+#define D_SQLITE            0x0000001000000000
```
underhood, line 47: not a problem with this PR but just noticed is there any reason why we are skipping `0x0000000080000000` between:
```
#define D_STREAM            0x0000000040000000
#define D_RRDENGINE         0x0000000100000000
```

underhood: also wouldn't using `D_METADATALOG` serve the purpose? What in case in future netdata uses SQLite in more places will this D_LOG be for all SQLite errors regardles of Netdata part?

database/engine/metadata_log/metadatalogapi.c
stelfrag: True, I will change it

libnetdata/log/log.h
stelfrag: Hmm good point. Right now the D_SQLITE will report things related to metadata and nothing else (for example generating UUIDs etc)

SQLite errors will be reported in the `error.log` anyway.

database/sqlite/sqlite_functions.c
```diff
@@ -0,0 +1,1224 @@
...
+        freez(item);
+    }
+    return;
+}
+
+/*
+ * Find a type,id,name in cache -- remove if found
+ * Caller must lock the uuid_cache
+ */
+uuid_t *find_in_uuid_cache(struct uuid_cache **uuid_cache, const char *type, const char *id, const char *name)
```
underhood, line 279: Should we optimize with sth. like skip list?

database/rrddim.c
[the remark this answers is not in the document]
stelfrag: Changed, also added code in the chart case

database/sqlite/sqlite_functions.c
stelfrag: I don't see it as an urgent issue because that list is used only during init of charts / dimensions and it progressively shrinks as items are found. 

However because charts and dimensions are created in a more or less consistent order (when a collector starts) I will add a small change to the query itself which will result in the items stored in reverse order in the list.

Doing a quick test on an agent with metadata log to migrate, the item search was in the first entry of the list
- Before the query change --> 14% of the time
- After the query change --> 74% of the time

That jumps to 92% for an agent that starts fresh (dbengine / metadata log files deleted)

underhood:
some minor warnings:

```
database/sqlite/sqlite3.c: In function ‘statConnect’:
database/sqlite/sqlite3.c:201327:9: warning: unused parameter ‘pAux’ [-Wunused-parameter]
201327 |   void *pAux,
       |   ~~~~~~^~~~
database/sqlite/sqlite3.c: In function ‘statBestIndex’:
database/sqlite/sqlite3.c:201383:40: warning: unused parameter ‘tab’ [-Wunused-parameter]
201383 | static int statBestIndex(sqlite3_vtab *tab, sqlite3_index_info *pIdxInfo){
       |                          ~~~~~~~~~~~~~~^~~
database/sqlite/sqlite3.c: In function ‘statFilter’:
database/sqlite/sqlite3.c:201865:27: warning: unused parameter ‘idxStr’ [-Wunused-parameter]
201865 |   int idxNum, const char *idxStr,
       |               ~~~~~~~~~~~~^~~~~~
database/sqlite/sqlite3.c:201866:7: warning: unused parameter ‘argc’ [-Wunused-parameter]
201866 |   int argc, sqlite3_value **argv
       |   ~~~~^~~~
```

database/sqlite/sqlite_functions.c
```diff
@@ -231,170 +228,20 @@ int sql_database_size(void)
...
-    return 0;
-#endif
-}
-
 #define FIND_UUID_TYPE  "select 1 from host where host_id = @uuid union select 2 from chart where chart_id = @uuid union select 3 from dimension where dim_id = @uuid;"
 
 int find_uuid_type(uuid_t *uuid)
 {
-    sqlite3_stmt *res = NULL;
+    static __thread sqlite3_stmt *res = NULL;
```
underhood, line 235: is this not GCC specific?

underhood: Additionally if you use ptherads api: `In addition pthread_key_create can optionally accept a destructor function that will automatically be called at thread exit, if the thread-specific data is not NULL`

mfundul: The pthread thing has CPU overhead because it does allocations and a lookup. We already use thread local storage in netdata in the core:
https://github.com/netdata/netdata/blob/3f8e44d1a6c501feeeacab3b3bf86ec0e401d4a3/libnetdata/threads/threads.c#L19

underhood: Well there are multiple options here. Question is how much we care to clean up as per previous discussion.

mfundul: My main argument here is that there is no problem with GCC since this kind of code already exists in netdata, so we shouldn't worry about compatibility.

underhood: Yes that is certainly valid argument. I see 2 minor problems with this (don't worry won't block this PR on this though):
1. cleanup
2. portability

Is it worth killing both? How much is the performance impact?

stelfrag: For this specific function it may not matter because it will be used only during migration, but in general preparing the statement helps a lot.

Adding some quick logging to check

Preparing every time
```
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 49 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 31 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 26 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 24 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 23 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 23 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 23 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 23 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 23 usec
2020-11-23 12:04:02: netdata INFO  : MAIN : SQLITE: FIND UUID 23 usec
```

Preparing once
```
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 43 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 13 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 7 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 6 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 6 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 6 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 5 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 5 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 6 usec
2020-11-23 12:02:16: netdata INFO  : MAIN : SQLITE: FIND UUID 6 usec
```

We can refine the code in future PRs to properly release the memory allocated by having a generic prepare function that will keep track of the statements and release during shutdown.

mfundul: We can also resolve the memory leak by pushing in the pthread cleanup callback stack with via [pthread_cleanup_push()](https://man7.org/linux/man-pages/man3/pthread_cleanup_push.3.html).

mfundul:
We should do what @underhood suggested and make the warnings disappear so that we don't break any other build systems people are using, e.g. for packaging netdata into other Linux distributions.

stelfrag:
> We should do what @underhood suggested and make the warnings disappear so that we don't break any other build systems people are using, e.g. for packaging netdata into other Linux distributions.

There is a `#pragma GCC diagnostic ignored "-Wunused-parameter" ` that can be added to the top of the file. I need to check

mfundul (dismissed):

underhood (approved):

mfundul (approved):

underhood (approved):

PR #10142, opened 2020-10-26
Update proc_net_snmp.c

manos-saratsis:
I believe the -1 was the wrong multiplier. Also as the dimensions represent different errors the default chart type should be changed to stacked.


vlvkobal:
I'm closing this. The value is `-1` by design at the moment.

PR #10143, opened 2020-10-26
Update proc_net_snmp6.c

manos-saratsis:
Maybe the multiplier is wrong and the chart type should be stacked


ilyam8:
@manos-saratsis that pattern (read/write in/out receive/send, etc: positive/negative) is used all over the place.

See example

https://github.com/netdata/netdata/blob/8352554ff8121e8ad537098be5d5cf481c3d1107/collectors/proc.plugin/proc_net_snmp.c#L631-L650

ilyam8:
Good example

Screenshot 2020-10-27 at 18 12 03

So there is logic behind it, but

> the multiplier is wrong and the chart type should be stacked

is a valid representation too 👍

vlvkobal:
I'm closing this. The value is `-1` by design at the momet.

PR #10144, opened 2020-10-27
smartd_log: Change from absolute to incremental

### LINKED ISSUE #7388: smartd_log: wrong chart type for several attributes
Sorry, I was otherwise engaged. There are a few other entries which need fixing too.

Most of the "INCREMENTAL" entries should be looked at closely, however the important ones are:

read_total_err_corrected
read_total_unc_errors
write_total_err_corrected
write_total_unc_errors
verify_total_err_corrected
verify_total_unc_errors

spin_up_retries
calibration_retries
reallocated_sectors_count
program_fail_count
erase_fail_count
reallocation_event_count

When monitoring drives, it's easy to miss blips and what really matters is the _total_ health numbers over time (It's worth reading the smart attribute tables for explanations of each item)

If these traces sit at zero it's far too easy to get a false sense of security

_Originally posted by @Stoatwblr in https://github.com/netdata/netdata/issues/7383#issuecomment-559598122_

thiagoftsm:
##### Summary
Fixes #7388 

Considering the [comment](https://github.com/netdata/netdata/issues/7388#issuecomment-716874179) that @Ferroin made and comparing with what was discussed at [4285](https://github.com/netdata/netdata/issues/4285#issuecomment-425654443),  we should at least to change one algorithm inside the module.

##### Component Name
smartd_log
##### Test Plan

1 - Create the directory `/var/log/smartd/`.
2 - Start `smartd` running :

```bash
smartd -A /var/log/smartd/ -i 600
```

3 - Start Netdata

### DIFF
--- a/collectors/python.d.plugin/smartd_log/smartd_log.chart.py
+++ b/collectors/python.d.plugin/smartd_log/smartd_log.chart.py
@@ -262,7 +262,7 @@ CHARTS = {
                     'line'],
         'lines': [],
         'attrs': [ATTR5],
-        'algo': ABSOLUTE,
+        'algo': INCREMENTAL,
     },
     'reserved_block_count': {
         'options': [None, 'Reserved Block Count', 'percentage', 'wear', 'smartd_log.reserved_block_count', 'line'],


ilyam8:
This PR doesn't fix #7388.

I wrote there

> All attributes from OP should be absolute, not incremental. We need to check it and fix it if it is true.

> i tend to agree. incremental type for raw values is not good, because zero values (rate) on a chart are misleading, a user interpret them as all is ok, which is false

How come changing smartd 5 attribute algorithm from `absolute` to `incremental` fixes the issue?

---

There is `help wanted` label because i wanted some input to before making a decision. It didn't look clear for me at that time, we could revisit the issue..

ilyam8:
It is absolute according its description

> Count of reallocated sectors. The raw value represents a count of the bad sectors that have been found and remapped.[25] Thus, the higher the attribute value, the more sectors the drive has had to reallocate. This value is primarily used as a metric of the life expectancy of the drive; **a drive which has had any reallocations at all is significantly more likely to fail in the immediate months**.


It doesn't look like we are interested in rate of reallocated sectors, but existence of them.

Another problem with rate is

> incremental type for raw values is not good, because **zero values (rate) on a chart are misleading, a user interpret them as all is ok, which is false**

PR #10148, opened 2020-10-27
Split ReviewDog check to only run when relevant.

Ferroin:
##### Summary

This splits out the ReviewDog CI checks to only run on PRs that actually change files of types they are checking. This should keep the CI list cleaner, and make it less likely that issues with ReviewDog will affect PRs that are unrelated to the checks that are having issues.

##### Component Name

area/ci

##### Test Plan

Of the three review checks, only the yamllint one actually runs the check itself, with the others skipping the reviewdog step of their job.

##### Additional Information

Prompted by current issues with the eslint checks in ReviewDog.

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -3,6 +3,10 @@
 name: Review
 on:
   pull_request:
+env:
+  run_eslint: 0
+  run_shellcheck: 0
+  run_yamllint: 0
 jobs:
   eslint:
     name: eslint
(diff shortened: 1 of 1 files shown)


.github/workflows/reviewdog-yamllint.yml
```diff
@@ -0,0 +1,19 @@
+---
+name: ReviewDog yamllint
+on:
+  pull_equest:
+    paths:
+      - '**.yml'
+      - '**.yaml'
```
ilyam8, line 7: add `python.d` config files?

Ferroin: Oh yes, I forgot we have config files that are YAML but do not end in a proper YAML extension...

Ferroin: Looks like it’s not properly matching the GHA workflows either...

Ferroin:
OK, looks like this needs a completely different approach...

knatsakis (approved):

PR #10166, opened 2020-10-30
add explicit example of how to get to netdata.conf and open it in edi…

andrewm4894:
…t mode

This may be overkill but assuming this is one place a lot of new people will look i was wondering if worth explicitly adding step of how to get in and open netdata.conf in edit mode. 

Totally ok if you think this add's more noise than is worth. Just that one thing always struck me was everywhere we say "just edit netdata.conf" but for like the first few months that basically meant me opening up another chrome tab and searching for "edit netdata.conf" until it just came naturally for me to just do:

```bash
cd /etc/netdata/
sudo ./edit-config netdata.conf
```

##### Component Name

docs

### DIFF
--- a/docs/guides/longer-metrics-storage.md
+++ b/docs/guides/longer-metrics-storage.md
@@ -27,6 +27,12 @@ larger dataset than your system's available RAM.
 The database engine is currently the default method of storing metrics, but if you're not sure which database you're
 using, check out your `netdata.conf` file and look for the `memory mode` setting:
 
+```bash
+# navigate to and open your netdata.conf file to make changes if needed
+cd /etc/netdata/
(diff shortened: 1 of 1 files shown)


joelhans (commented):
Do you think we also could add a link in the paragraph above this code block to [this doc](https://learn.netdata.cloud/docs/configure/nodes), which is meant to be a source of truth for how to find and use `edit-config`? I don't mind if you add the steps here, but we should give users a reference. This isn't your fault, but rather mine! But it's a convenient time to add in the link :wink: 

Otherwise LGTM!

andrewm4894:
closing as stale and not really relevant any more

PR #10180, opened 2020-11-02
Fix an infinite loop in the statsd plugin

### LINKED ISSUE #10176: statsd_process() in dead loop
##### Bug report summary
static inline size_t statsd_process(char *buffer, size_t size, int require_newlines);
if buffer is ":", this function will never end.

I do not know why the buffer is ":" and it only happens on virtual machines. Anyway, there is a definitely bug in function statsd_process().

vlvkobal:
##### Summary
Fixes #10176

##### Component Name
statsd plugin

##### Test Plan
1) Insert `strcpy(buffer, ":");` in https://github.com/netdata/netdata/blob/master/collectors/statsd.plugin/statsd.c#L693
2) Run `echo "test:1|c" | nc -u localhost 8125`
3) Check if there is no infinite loop.

### DIFF
--- a/collectors/statsd.plugin/statsd.c
+++ b/collectors/statsd.plugin/statsd.c
@@ -698,7 +698,10 @@ static inline size_t statsd_process(char *buffer, size_t size, int require_newli
 
         s = name_end = (char *)statsd_parse_skip_up_to(name = s, ':', '|');
         if(name == name_end) {
-            s = statsd_parse_skip_spaces(s);
+            if (*s) {
+                s++;
+                s = statsd_parse_skip_spaces(s);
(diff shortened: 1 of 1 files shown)


ilyam8:
What i did:
 - (latest master): `echo ':' | nc -c -w 1 -u localhost 8125` => netdata process uses 100% cpu
 - (this PR): `echo ':' | nc -c -w 1 -u localhost 8125` => all is ok

Screenshot 2020-11-03 at 11 08 49

ilyam8 (approved):

stelfrag (approved):
Looks good

Issue #10178, opened 2020-11-02
libnetdata is not usable from c++ (cpp)


KickerTom:
##### Bug report summary
Attempting to write a new plugin for netdata in C++, I found that the headers are C only, unusable for C++ application. I understand netdata is (partially) C project, but interface libraries, and especially headers, could be C++ clean to make it more usable.
One of the most prominent problems is usage of _this_ as a struct member, while _this_ is reserved keyword in cpp.

##### Component Name
libnetdata

##### Steps To Reproduce
1. Create simple test.cpp file including libnetdata.h, e.g.:
```
#include "config.h"
#include "libnetdata/libnetdata.h"

int main(int argc, char **argv)
{
    return 0;
}
```
2. Try compiling the code with e.g.
`g++ -DTARGET_OS=OS_LINUX test.cpp`

3. You will receive many errors:
```
In file included from libnetdata/libnetdata.h:304,
                 from test.cpp:2:
libnetdata/avl/avl.h:43:14: error: declaration of ‘avl_tree avl_tree_lock::avl_tree’ [-fpermissive]
     avl_tree avl_tree;
              ^~~~~~~~
libnetdata/avl/avl.h:40:3: error: changes meaning of ‘avl_tree’ from ‘typedef struct avl_tree avl_tree’ [-fpermissive]
 } avl_tree;
   ^~~~~~~~
In file included from libnetdata/libnetdata.h:313,
                 from test.cpp:2:
libnetdata/config/appconfig.h:114:9: error: declaration of ‘avl config_option::avl’ [-fpermissive]
     avl avl;                // the index entry of this entry - this has to be first!
         ^~~
In file included from libnetdata/libnetdata.h:304,
                 from test.cpp:2:
libnetdata/avl/avl.h:34:3: error: changes meaning of ‘avl’ from ‘typedef struct avl avl’ [-fpermissive]
 } avl;
   ^~~
In file included from libnetdata/libnetdata.h:313,
                 from test.cpp:2:
libnetdata/config/appconfig.h:127:9: error: declaration of ‘avl section::avl’ [-fpermissive]
     avl avl;                // the index entry of this section - this has to be first!
         ^~~
In file included from libnetdata/libnetdata.h:304,
                 from test.cpp:2:
libnetdata/avl/avl.h:34:3: error: changes meaning of ‘avl’ from ‘typedef struct avl avl’ [-fpermissive]
 } avl;
   ^~~
In file included from libnetdata/libnetdata.h:316,
[...]

PR #10185, opened 2020-11-03
Make libnetdata headers compilable by C++.

KickerTom:
Minimal set of changes to libnetdata to allow including it from C++ code. Please note that I haven't even tried compiling netdata itself with these changes, nor I checked the code can be really used from C++ code. This only removes compilation errors and serves for demonstration purposes.
Please refer to issue #10178 for more details.


KickerTom:
Seems to be complete now. Fixes #10178.
Please note I have added the 'extern "C"' only in the main libnetdata.h and log/log.h headers. If some other headers are expected to be used alone, without including the whole lot, let me know, I will add it there too.

thiagoftsm (dismissed):
I tested your PR now and I could not identify any problem neither with the plugins nor with stream, that are direct affected from these changes.

Thank you @KickerTom  for your contribution!

libnetdata/dictionary/dictionary.h
```diff
@@ -13,7 +13,7 @@ struct dictionary_stats {
 };
 
 typedef struct name_value {
-    avl avl;                // the index - this has to be first!
+    avl avl_instance;       // the index - this has to be first!
```
mfundul, line 16:
```suggestion
    avl avl_node;           // the index - this has to be first!
```
Let's keep it consistent with the similar change above.

libnetdata/avl/avl.h
```diff
@@ -34,13 +34,13 @@ typedef struct avl {
...
-typedef struct avl_tree {
+typedef struct avl_tree_type {
     avl *root;
     int (*compar)(void *a, void *b);
-} avl_tree;
+} avl_tree_type;
 
 typedef struct avl_tree_lock {
-    avl_tree avl_tree;
+    avl_tree_type avl_tree;
```
mfundul, line 43:
```suggestion
    avl_tree avl_tree_instance;
```
I don't like how changes propagated outside of libnetdata. Let's go the other way around, similar to what you did with `struct avl`.

mfundul (changes requested):

libnetdata/avl/avl.h
KickerTom: I originally tried that as well. But the avl_tree member is being accessed on many places, so it propagates outside even more. Moreover, as part of the app config structure (appconfig.h, struct config) it is being used and accessed often by collector plugins, so changing the member name has bigger potential to break third-party plugins than the type name change.
But both variants are possible, so if you think changing member name is better, I can do that.

mfundul: No, you are right, there are a bunch of initializers that touch the `avl_tree` field at various places. It's probably not worth it.

mfundul (approved):

thiagoftsm (approved):

thiagoftsm:
Thank you very much @KickerTom !

Issue #10198, opened 2020-11-04
TrueNAS (FreeBSD) auto update does not work


abrbon:
On my TrueNAS 12 server (based on FreeBSD) the auto-update does not work.

If i run the "/etc/periodic/daily/netdata-updater" script manually it does work fine altough i have to answer 1 question during the script with "Y" (or "N") before it can continue and finish the update.

`[/tmp/netdata-updater-XXXXXXXXXX.SxF11L2O/netdata-v1.26.0]# mv go.d.plugin-v0.23.0.freebsd-amd64 /usr/local/netdata/usr/libexec/netdata/plugins.d/go.d.plugin
override rwxr-x---  root/netdata for /usr/local/netdata/usr/libexec/netdata/plugins.d/go.d.plugin? (y/n [n])`

I have checked if the netdata-updater actually runs via cron by editing the netdata-updater script with a simple 
1 line code "touch /root/test-netdata-updater-cron.txt" and ran the 'periodic daily' command to force the cron to 
execute the daily scripts and the test-netdata-updater-cron.txt is created in my /root/ folder as intended.

But i can not find any logs or anything why the netdata-updater script does not work.

I have created a new jail on TrueNAS and installed netdata v1.25 from source inside this new jail.

Any ideas?

Ferroin:
Most likely cause is a difference between GNU and FreeBSD userspace. I’ll try to make a point to look into this further either later today or early tomorrow.

ilyam8:
Default is no confirmation, `mv` asks on overwriting only when it is explicitly called with `-i` option. Same behavior for both GNU and bsd versions ([a funny related topic](https://forums.freebsd.org/threads/prompt-confirmation-y-n-to-move-a-file.70676/)).

ilyam8:
Likely not, but i have only this in mind - could it be an alias?

Ferroin:
It might be an alias, but I would be rather surprised if that was leaking into a shell script. Aliases should be defined in the profile files, not the shell rc files, and thus should not be visible in shell scripts.

abrbon:
I don't think it's an alias issue because there is a second 'mv' command i can see in the installer log which did not ask for confirmation during install.

`[root /root/netdata-v1.25.0]# mv go.d.plugin-v0.22.0.freebsd-amd64 /usr/local/netdata/usr/libexec/netdata/plugins.d/go.d.plugin  ... OK
[...]

PR #10204, opened 2020-11-06
Fixed bug in auto-updater for FreeBSD (#10198)

abrbon:
**Summary**

Fixes bug in auto-updater for FreeBSD systems (like FreeNAS/TrueNAS) #10198 

The netdata-updater.sh script which is used to automatically check for new versions of Netdata and install them if necessary
did not work in my TrueNAS jail (based on FreeBSD).

Problem is the RANDOM function which does not work in FreeBSD.
Build an extra check if running on FreeBSD system to use another method for the RANDOM function on FreeBSD.

**Component Name**

area/packaging

**Test plan**

Installed an older version of Netdata (v1.25) and ran the 'periodic daily' command to force the daily cron jobs to execute and checked if Netdata was automatically upgraded to latest version (v1.26 at time of writing).
All seemed to work fine.

**Additional Information**

@Ferroin  This is my first PR ever on Github so if i did anything wrong please let me know.

Fixes: #10198

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -256,7 +256,7 @@ done
 # But only we're not a controlling terminal (tty)
 # Randomly sleep between 1s and 60m
 if [ ! -t 1 ] && [ -z "${NETDATA_NOT_RUNNING_FROM_CRON}" ]; then
-  sleep $(((RANDOM % 3600) + 1))s
+    sleep $(((RANDOM % 3600) + 1))
 fi
 
 # Usually stored in /etc/netdata/.environment


ilyam8:
@abrbon 👋 


See the warnings

> https://github.com/netdata/netdata/pull/10204/files

Ferroin (commented):
Looks great other than the two shellcheck warnings (if you have questions about them I’ll be happy to explain). You can update the PR to fix those warnings by making the required changes to the file, creating a new commit on this branch in your local repository, and then pushing the branch again (GitHub will automatically detect the changes to the branch and update the PR accordingly).

knatsakis (commented):
Happy to approve as long as the comments in https://github.com/netdata/netdata/pull/10204#pullrequestreview-525092743 are addressed

abrbon:
Hi I did try to get rid off the back ticks but it didn’t work. I think it is FreeBSD specific.
I will try to figure out a solution.

ilyam8:
@abrbon did you try `sleep $(jot -r 1 1 3600)`?

And i still dont get why it is the problem - the issue was a confirmation dialog during updating/installation

> [/tmp/netdata-updater-XXXXXXXXXX.SxF11L2O/netdata-v1.26.0]# mv go.d.plugin-v0.23.0.freebsd-amd64 /usr/local/netdata/usr/libexec/netdata/plugins.d/go.d.plugin override rwxr-x--- root/netdata for /usr/local/netdata/usr/libexec/netdata/plugins.d/go.d.plugin? (y/n [n])

How come invalid random func is the cause?

And i think `RANDOM` is a bash function, so it shoudn't be os specific 🤔

abrbon:
@ilyam8 That was the first thing i tried (the $(jot ...) but it didn't work. 
Nothing happend, no error or anything it just doesn't work.

The issue was that that the auto-update didn't work.
But running the netdata-updater.sh manually works except for the confirmation dialog.

But i tried simple test with following line from the command line: `echo $((RANDOM))` 
but that gives me an error message: 

```
root@freenas:/ # echo $((RANDOM))
Illegal variable name.
root@freenas:/ # 
```
So i think it is FreeBSD specific since FreeNAS/TrueNAS is based on FreeBSD.

abrbon:
Finally found the issue, turned out to be very simple :)

The reason the auto-updater didn't work on my TrueNAS server turned out to be a simple fix.
All i needed to do is to remove the 's' at the end of the line: 

`sleep $(((RANDOM % 3600) + 1))s`

My apologies for this mistake on my side :(
Sometimes the solotion is really that simple!

I've updated the PR to correct for this.

ilyam8:
@abrbon interesting 🤔 

macOS

```
0 ~ $ sleep --help
Usage: sleep NUMBER[SUFFIX]...
  or:  sleep OPTION
Pause for NUMBER seconds.  SUFFIX may be 's' for seconds (the default),
'm' for minutes, 'h' for hours or 'd' for days.  NUMBER need not be an
integer.  Given two or more arguments, pause for the amount of time
specified by the sum of their values.
```

abrbon:
@ilyam8 On macOS or Linux you can use the sleep command with the 's' at the end.
However when i test this on my TrueNAS server (jail) where i installed my Netdata it fails with the extra 's' at the end.
Even if i switch to a bash env:
```
root@netdata-custom:~ # env bash
[root@netdata-custom ~]# sleep 2s
usage: sleep seconds
[root@netdata-custom ~]# sleep 2
[root@netdata-custom ~]# 

```
So that is interesting indeed 🤔

But it works on all platforms (macOS, Linux, FreeBSD) without the 's' as well.

ilyam8 (approved):

thiagoftsm (approved):
After to test this PR on FreeBSD-12.2, I confirm that the update happened without problems.

knatsakis (approved):

Ferroin (approved):

Ferroin:
@abrbon Thanks for this fix and all the awesome work you did debugging the issue!

abrbon:
> @abrbon Thanks for this fix and all the awesome work you did debugging the issue!

No problem @Ferroin, it was fun to do and i have learned something new as well!

PR #10217, opened 2020-11-09
Don't stop Prometheus remote write collector when data is not available for dimension formatting

### LINKED ISSUE #10152: Can't configure Netdata to export via prometheus_remote_write to Victoria Metrics
As I'm having trouble configuring Netdata with Victoria Metrics backend via prometheus_remote_write, I opened an issue on Victoria Metrics github.
Here's the link : [VictoriaMetrics/VictoriaMetrics/issues/867](https://github.com/VictoriaMetrics/VictoriaMetrics/issues/867)

Please tell me if I have to copy/paste the full content of the issue here, or if you need more information.

vlvkobal:
##### Summary
If `data source = as collected` and there is no data available, Prometheus remote write collector stops with the following error
```
/var/log/netdata/error.log:2020-10-29 13:58:54: netdata ERROR : EXPORTING : EXPORTING: cannot format metric for victoria_metrics
```
The connector should continue processing data instead of exiting.

Fixes #10152

##### Component Name
exporting engine

### DIFF
--- a/exporting/prometheus/remote_write/remote_write.c
+++ b/exporting/prometheus/remote_write/remote_write.c
@@ -242,7 +242,7 @@ int format_dimension_prometheus_remote_write(struct instance *instance, RRDDIM *
                     (unsigned long)rd->last_collected_time.tv_sec,
                     (unsigned long)instance->after,
                     (unsigned long)instance->before);
-                return 1;
+                return 0;
             }
 
             if (homogeneous) {


thiagoftsm (approved):
LGTM!

stelfrag:
Fix a bit the comments in the function definition that says

```
@return Always returns 0.
```

because it doesn't now and we are good to go  !!!

vlvkobal:
> because it doesn't now

On the contrary, it does now.

stelfrag (approved):
Looks good !

PR #10218, opened 2020-11-09
Fix chart filtering

vlvkobal:
##### Summary
When chart filtering is configured for both, the Prometheus Web API and a Prometheus remote write connector instance, the Web API adds charts that are configured in a `prometheus_remote_write` section to its own filter.

Related: #10057

##### Component Name
exporting engine

##### Test Plan
Configure chart filtering for the Prometheus Web API and for a Prometheus remote write connector instance. Check that `http://localhost:19999/api/v1/allmetrics?format=prometheus` doesn't return charts, configured for the Prometheus remote write instance.


thiagoftsm (approved):

stelfrag (approved):
Looks good

PR #10240, opened 2020-11-16
Fix streaming buffer size

### LINKED ISSUE #10220: Gaps in streamed charts / k8s helm deployment
##### Bug report summary
Investigating some random issue on a K8S cluster, trying to deploy netdata to have more visibility, the graphs are full of gaps/holes/discontinuities
image

##### OS / Environment
```
4.18.0-193.19.1.el8_2.x86_64
Red Hat Enterprise Linux 8.2 (Ootpa)
k8s v1.18.6
containerd 1.2.13
```

##### Netdata version
```
netdata v1.26.0
chart netdata-2.0.15
```

##### Component Name
?

##### Steps To Reproduce

Deploy using helm chart netdata-2.0.15

```
ingress:
  enabled: true
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /$2
    kubernetes.io/ingress.class: nginx
  path: /netdata(/|$)(.*) 
  hosts:
    - pilot
  tls:
    - secretName: pilot-tls
      hosts:
        - pilot

parent:
  env:
    DO_NOT_TRACK: 1

  database:
    persistence: false

  alarms:
    persistence: false

  nodeSelector:
    kubernetes.io/hostname: master-01

  tolerations:
    - operator: Exists
      effect: NoSchedule

  configs:
    netdata:
      enabled: true
      path: /etc/netdata/netdata.conf
      data: |
        [global]
          process scheduling policy = nice
          process nice level = -10
          memory mode = save

        [plugins]
          cgroups = no
          tc = no
          enable running new plugins = no
          check for new plugins every = 72000
          python.d = no
          charts.d = no
          go.d = no
          node.d = no
          apps = no
          proc = no
          idlejitter = no
          diskspace = no
    stream:
      enabled: true
      path: /etc/netdata/stream.conf
      data: |
        [11111111-2222-3333-4444-555555555555]
          enabled = yes
          history = 172800
          default memory mode = save
          health enabled by default = auto
          allow from = *

child:
  env:
    DO_NOT_TRACK: 1
  configs:
    netdata:
      enabled: true
      path: /etc/netdata/netdata.conf
      data: |
        [global]
          process scheduling policy = nice
          process nice level = -10
          memory mode = none
          update every = 5
        [health]
          enabled = no
    stream:
      enabled: true
      path: /etc/netdata/stream.conf
      data: |
        [stream]
          enabled = yes
          destination = netdata:{{ .Values.service.port }}
          api key = 11111111-2222-3333-4444-555555555555
          timeout seconds = 60
          buffer size bytes = 10485760
          reconnect delay seconds = 5
          initial clock resync iterations = 60
```

##### Expected behavior
Everything works

##### Comments
I've already tried to bump `buffer size bytes` on the childs, to set the nice value at -10 instead of 19, and increase `update every`
I've verified that the settings are applied via `curl IP:19999/netdata.conf` (don't know how to check for stream.conf)
Depending on the servers we have between 7.5k et 8.5k metrics
no buffer full logs (`kubectl logs -l role=child --all-containers | grep -i buf`)

vlvkobal:
##### Summary
Since #9214 was merged the `buffer size bytes` option didn't work for streaming.

Fixes #10220

##### Component Name
streaming

### DIFF
--- a/streaming/rrdpush.h
+++ b/streaming/rrdpush.h
@@ -43,7 +43,6 @@ struct sender_state {
     pid_t task_id;
     unsigned int overflow:1;
     int timeout, default_port;
-    size_t max_size;
     usec_t reconnect_delay;
     char connected_to[CONNECTED_TO_SIZE + 1];   // We don't know which proxy we connect to, passed back from socket.c
     size_t begin;
--- a/streaming/sender.c
+++ b/streaming/sender.c
@@ -584,10 +584,14 @@ void *rrdpush_sender_thread(void *ptr) {
 
     s->timeout = (int)appconfig_get_number(&stream_config, CONFIG_SECTION_STREAM, "timeout seconds", 60);
     s->default_port = (int)appconfig_get_number(&stream_config, CONFIG_SECTION_STREAM, "default port", 19999);
-    s->buffer->max_size = (size_t)appconfig_get_number(&stream_config, CONFIG_SECTION_STREAM, "buffer size bytes",
-                                                  1024 * 1024);
-    s->reconnect_delay = (unsigned int)appconfig_get_number(&stream_config, CONFIG_SECTION_STREAM, "reconnect delay seconds", 5);
-    remote_clock_resync_iterations = (unsigned int)appconfig_get_number(&stream_config, CONFIG_SECTION_STREAM, "initial clock resync iterations", remote_clock_resync_iterations); // TODO: REMOVE FOR SLEW / GAPFILLING
+    s->buffer->max_size =
+        (size_t)appconfig_get_number(&stream_config, CONFIG_SECTION_STREAM, "buffer size bytes", 1024 * 1024);
+    s->reconnect_delay =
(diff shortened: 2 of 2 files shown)


mfundul:
You should either remove the `max_size` field if it is to remain unused, or do it similarly to [here](https://github.com/netdata/netdata/pull/9832/commits/b1cefbb6f4bd0642a23d8cb336e4e894fcce6d54).

vlvkobal:
I don't understand. I removed it from the struct. Are there other places where I should remove it from?

mfundul:
> 
> 
> I don't understand. I removed it from the struct. Are there other places where I should remove it from?

Nope, I'm just blind apparently!

mfundul (approved):

stelfrag:
It works as it was changed -- just omitting the s->max_size altogether and setting directly to the buffer max size.

stelfrag (approved):

thiagoftsm (approved):
LGTM!

Issue #10249, opened 2020-11-17
Make netdata-updater self-update prior to updating Netdata.


Ferroin:
##### Feature idea summary

Put simply, our updater script should always try to run the latest version of itself so that any bugs in older versions are avoided. This translates to fetching the most recent version of itself and switching to that code prior to actually updating Netdata.

##### Expected behavior

The updater script properly recovers from a wider variety of updater bugs than it currently does.

PR #10261, opened 2020-11-19
Made the update script significantly more robust and user friendly.

### LINKED ISSUE #9863: [Bug]: the source directory does not include netdata-installer.sh
I run  `bash <(curl -Ss https://my-netdata.io/kickstart.sh)`

The first line output says `/dev/fd/63: line 167: /tmp/netdata-test.GTtLvJePOp: Permission denied`

And it ends:
`ABORTED Cannot install netdata from source (the source directory does not include netdata-installer.sh). Leaving all files in /tmp/netdata-kickstart-Q30do8Yo21`

**It's the first time I get an error when installing Netdata. Can anyone advise me how to solve this problem?**

Full output:

`[root@daisy-5 ~]# bash <(curl -Ss https://my-netdata.io/kickstart.sh)
/dev/fd/63: line 167: /tmp/netdata-test.GTtLvJePOp: Permission denied
System            : Linux
Operating System  : GNU/Linux
Machine           : x86_64
BASH major version:
 --- Fetching script to detect required packages... ---
[/tmp/netdata-kickstart-y8DEbXhhIM]# curl -q -sSL --connect-timeout 10 --retry 3 --output /tmp/netdata-kickstart-y8DEbXhhIM/install-required-packages.sh https://raw.githubusercontent.com/netdata/netdata/master/packaging/installer/install-required-packages.sh  OK

 --- Running downloaded script to detect required packages... ---
[/tmp/netdata-kickstart-y8DEbXhhIM]# /usr/bin/bash /tmp/netdata-kickstart-y8DEbXhhIM/install-required-packages.sh netdata Loading /etc/os-release ...
You should have EPEL enabled to install all the prerequisites.
Check: http://www.tecmint.com/how-to-enable-epel-repository-for-rhel-centos-6-5/
 > CentOS Version: 7 ...
 > Checking for epel ...

/etc/os-release information:
NAME            : CentOS Linux
VERSION         : 7 (Core)
ID              : centos
ID_LIKE         : rhel fedora
VERSION_ID      : 7

We detected these:
Distribution    : centos
Version         : 7
Codename        : 7 (Core)
Package Manager : install_yum
Packages Tree   : centos
Detection Method: /etc/os-release
Default Python v: 2 (will install python3 too)

Searching for distro_sdk ...
Searching for autoconf_archive ...
 > Checking if package 'autoconf-archive' is installed...
Searching for libz_dev ...
 > Checking if package 'zlib-devel' is installed...
Searching for libuuid_dev ...
 > Checking if package 'libuuid-devel' is installed...
Searching for libmnl_dev ...
 > Checking if package 'libmnl-devel' is installed...
Searching for json_c_dev ...
 > Checking if package 'json-c-devel' is installed...
Searching for libuv ...
 > Checking if package 'libuv-devel' is installed...
Searching for lz4 ...
 > Checking if package 'lz4-devel' is installed...
Searching for openssl ...
 > Checking if package 'openssl-devel' is installed...
Searching for judy ...
 > Checking if package 'Judy-devel' is installed...
Searching for libelf ...
 > Checking if package 'elfutils-libelf-devel' is installed...

All required packages are already installed. Now proceed to the next step.

 OK

[/tmp/netdata-kickstart-y8DEbXhhIM]# curl -q -sSL --connect-timeout 10 --retry 3 --output /tmp/netdata-kickstart-y8DEbXhhIM/sha256sum.txt https://storage.googleapis.com/netdata-nightlies/sha256sums.txt  OK

[/tmp/netdata-kickstart-y8DEbXhhIM]# curl -q -sSL --connect-timeout 10 --retry 3 --output /tmp/netdata-kickstart-y8DEbXhhIM/netdata-latest.tar.gz https://storage.googleapis.com/netdata-nightlies/netdata-latest.tar.gz  OK

[/tmp/netdata-kickstart-y8DEbXhhIM]# tar -xf netdata-latest.tar.gz  OK

 ABORTED  Cannot install netdata from source (the source directory does not include netdata-installer.sh). Leaving all files in /tmp/netdata-kickstart-y8DEbXhhIM

Ferroin:
##### Summary

This makes a number of changes to the updater to make it more user-friendly and more generally robust, including:

* If the installer is not where we expect it to be and there is _exactly_ one directory in the location we expected the installer to be, try changing to that directory and running it from there. This should fix #9863, as well as related issues with new installs.
* Each time the updater is run, check if a newer version exists in the repo, and if so download that and run it instead of using the current updater code. This will allow automatic recovery from a wide variety of bugs that may show up in the updater (especially since it runs very early in the update process). The actual check involves making a call to GitHub’s API to check the commit date on the master branch for the updater code, and only downloads it if the commit date is newer than the timestamp on the local copy. This intentionally _does not_ overwrite the local copy of the updater script, as that would actually make it easier to break things.

It also properly updates the copyright notice in the updater, which has gotten severely out of date.

##### Component Name

area/packaging

##### Test Plan

Behavior verified locally.

Testing of the auto-updating of the updater script can be verified by installing a local copy of Netdata from this branch, using `touch` to set the timestamp on the installed copy of the updater script to 1970-01-01 00:00:00 and adding a line just after line 362 to print something recognizable to the console, and then running the updater script. You should _not_ see output from the line added after 362 on the console if it works correctly.

##### Additional Information

Fixes: #9863 
Fixes: #10249

### DIFF
--- a/packaging/installer/kickstart.sh
+++ b/packaging/installer/kickstart.sh
@@ -471,12 +471,20 @@ cd netdata-* || fatal "Cannot cd to netdata source tree"
 # ---------------------------------------------------------------------------------------------------------------------
 # install netdata from source
 
-if [ -x netdata-installer.sh ]; then
+install() {
   progress "Installing netdata..."
   run ${sudo} ./netdata-installer.sh ${NETDATA_UPDATES} ${NETDATA_INSTALLER_OPTIONS} "${@}" || fatal "netdata-installer.sh exited with error"
   if [ -d "${ndtmpdir}" ] && [ ! "${ndtmpdir}" = "/" ]; then
     run ${sudo} rm -rf "${ndtmpdir}" > /dev/null 2>&1
   fi
+}
+
+if [ -x netdata-installer.sh ]; then
+  install "$@"
 else
-  fatal "Cannot install netdata from source (the source directory does not include netdata-installer.sh). Leaving all files in ${ndtmpdir}"
+  if [ "$(find . -mindepth 1 -maxdepth 1 -type d | wc -l)" -eq 1 ] && [ -x "$(find . -mindepth 1 -maxdepth 1 -type d)/netdata-installer.sh" ]; then
+    cd "$(find . -mindepth 1 -maxdepth 1 -type d)" && install "$@"
+  else
+    fatal "Cannot install netdata from source (the source directory does not include netdata-installer.sh). Leaving all files in ${ndtmpdir}"
+  fi
 fi
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -28,6 +28,8 @@
 
 set -e
 
+script_source="$("$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd -P)/netdata-updater.sh")"
+
 info() {
   echo >&3 "$(date) : INFO: " "${@}"
 }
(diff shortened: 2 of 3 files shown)


Ferroin:
@joelhans I’d like to add something to the documentation about how the self-updating updater script works so that we have something to point users at when they inevitably ask why auto-updates are now poking at GitHub. Any ideas where the best place to put this in the docs might be (long term should probably be included in the centralized privacy and security doc you were talking about in another PR the other day).

joelhans:
@Ferroin Great idea. What about under the **Automatic updates** section in `/packaging/installer/README.md`? We could create a new `###` section under that and explain the new behavior.

`/packaging/installer/UPDATE.md` is another obvious choice, but since that's more focused on helping users _manually_ update their installs, it wouldn't be very discoverable for the auto-updaters out there.

packaging/installer/kickstart.sh
```diff
@@ -471,12 +471,20 @@ cd netdata-* || fatal "Cannot cd to netdata source tree"
...
   if [ -d "${ndtmpdir}" ] && [ ! "${ndtmpdir}" = "/" ]; then
     run ${sudo} rm -rf "${ndtmpdir}" > /dev/null 2>&1
   fi
+}
+
+if [ -x netdata-installer.sh ]; then
+  install "$@"
 else
-  fatal "Cannot install netdata from source (the source directory does not include netdata-installer.sh). Leaving all files in ${ndtmpdir}"
+  if [ "$(find . -mindepth 1 -maxdepth 1 -type d | wc -l)" -eq 1 ] && [ -x "$(find . -mindepth 1 -maxdepth 1 -type d)/netdata-installer.sh" ]; then
```
dimko, line 485: @Ferroin Just to better understand how this should work.

Why is the line count check in the if condition required? Isn't existence and "executability" of the sh script, in combination with **find . -mindepth 1 -maxdepth 1 -type d** as script's parent directory, a sufficient check? 

If there are more than one parent directories for the script the condition will fail.

Is there some case that 
`find . -mindepth 1 -maxdepth 1 -type d/netdata-installer.sh` would be executed while `"$(find . -mindepth 1 -maxdepth 1 -type d | wc -l)" -eq 1` would fail? :thinking_face:

Ferroin: Find returns successfully as long as it finds any directories, while the case we are trying to account for here is a particular bug that causes our sources to be extracted into a sub-directory of the temporary working directory instead of directly into the temporary working directory. I’ve intentionally made the condition to check for this case as specific as possible to minimize the potential for this introducing other bugs.

dimko (approved):

packaging/installer/netdata-updater.sh
```diff
@@ -87,32 +89,85 @@ _cannot_use_tmpdir() {
...
+  fi
+}
+
+newer_commit_date() {
+  echo >&3 "Checking if a newer version of the updater script is available."
+
+  if command -v jq > /dev/null 2>&1; then
+    commit_date="$(_safe_download "https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1" /dev/stdout | jq '.[0].commit.committer.date')"
+  elif command -v python > /dev/null 2>&1;then
+    commit_date="$(_safe_download "https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1" /dev/stdout | python -c 'from __future__ import print_function;import sys,json;print(json.load(sys.stdin)["commit"]["committer"]["date"])')"
```
dimko, line 149: I guess, trying to install netdata and run the updater on a system that doesn't have support for either of these (jq or python) is an edge case and not possible.

Ferroin: No, it will work fine. We set `commit_date` to the start of the UNIX epoch below if it’s unset, so not having either of these conditions match just means that systems that have neither `jq` nor Python will always download the updater script instead of only doing it when there’s a newer version in the repo.

knatsakis (approved):

kaskavel (approved):

Issue #10264, opened 2020-11-19
nvidia-smi user statistics broken for LDAP users (not in /etc/passwd)


scatenag:
With latest updates to nvidia stack and netdata something is broken in nvidia-smi users charts:
nvidia_smi.gpu0_processes_mem shows correctly the process by the nvidia_smi.gpu0_user_mem and nvidia_smi.gpu0_user_num are empty.

![broken](https://user-images.githubusercontent.com/46522361/99726296-9bb9ed80-2ab6-11eb-8ab2-e7a9cb900e46.png)

netdata version: v1.26.0
NVIDIA-SMI 455.45.01    Driver Version: 455.45.01    CUDA Version: 11.1     
on CentOS Linux release 7.9.2009 (Core)

trying to debug it seems that 

- nvidia smi returns the correct pid
- `get_username_by_pid_safe` is not finding the username for pid,
that is, 
in ` /usr/libexec/netdata/python.d/nvidia_smi.chart.py`
running

```
for info in processes_info:
            pid = info.find('pid').text
            print(int(pid));
            print(info.find('process_name').text);
	    print(int(info.find('used_memory').text.split()[0]));
	    print(get_username_by_pid_safe(pid, passwd_file));
            processes.append({
                'pid': int(pid),
                'process_name': info.find('process_name').text,
                'used_memory': int(info.find('used_memory').text.split()[0]),
                'username': get_username_by_pid_safe(pid, passwd_file),
	    })
        return processes

```
I got no username

```
92144
python
1577

```

but ps is finding the right info

```
ps -p 92144 -o pid,vsz=MEMORY -o user,group=GROUP -o comm,args=ARGS                                                 1 ↵  22:34:57
   PID MEMORY USER     GROUP    COMMAND         ARGS
 92144 30673852 a.molin+ di_users python        python experiments_with_leam.py
```

Any thoughts to debug this ?

ilyam8:
Hi @scatenag 

check `passwd_file` content

scatenag:
Ok; I think I found the cause but not the solution:

the user is an LDAP user;

passwd_file does not contains it.

it's weird because I could have sworn it worked with LDAP users until recently.

ilyam8:
We look for a pid in the `/etc/passwd` file.

https://github.com/netdata/netdata/blob/149a7c3f3665545253cbe243afa642101764243f/collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py#L247-L269

[...]

PR #10268, opened 2020-11-20
python.d/nvidia_smi: use `pwd` lib to get username if not inside a container

ilyam8:
##### Summary

Fixes: #10264

[`pwd`](https://docs.python.org/3/library/pwd.html) is python standard library. But it doesn't allow to set _host_ prefix - that is why we don't want to use it inside a docker container.

This PR changes find username by uid logic to:
 - if installed directly on host (env var `NETDATA_HOST_PREFIX` not set): use `pwd`.
 - if not: `/etc/passwd` file using custom function that respects `NETDATA_HOST_PREFIX`.

If we fail (both cases): use process `st_uid` (User ID of owner) as username.

##### Component Name

`collectors/python.d/nvidia_smi`

##### Additional Information

- [x] test PR locally, ensure name resolution works.
- [x] test PR with LDAP users (LDAP is configured by sssd), ensure name resolution works (@scatenag )

### DIFF
--- a/collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
+++ b/collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
@@ -248,11 +248,12 @@ HOST_PREFIX = os.getenv('NETDATA_HOST_PREFIX')
 ETC_PASSWD_PATH = '/etc/passwd'
 PROC_PATH = '/proc'
 
-IS_INSIDE_DOCKER = HOST_PREFIX is not None
+IS_INSIDE_DOCKER = False
 
 if HOST_PREFIX:
     ETC_PASSWD_PATH = os.path.join(HOST_PREFIX, ETC_PASSWD_PATH[1:])
     PROC_PATH = os.path.join(HOST_PREFIX, PROC_PATH[1:])
+    IS_INSIDE_DOCKER = True
 
 
 def read_passwd_file():
@@ -294,7 +295,7 @@ def get_username_by_pid_safe(pid, passwd_file):
     try:
         return passwd_file[uid][0]
     except KeyError:
-        return uid
+        return str(uid)
 
 
 class GPU:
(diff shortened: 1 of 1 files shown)


scatenag:
Tested outside docker: it is running fine! 
@ilyam8 Great work!

andrewm4894 (commented):
Is it worth adding a note to readme for this? Or just an implementation detail that does not really need to be explicitly mentioned?

collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
```diff
@@ -247,9 +248,12 @@ def on_call(*args, **kwargs):
 ETC_PASSWD_PATH = '/etc/passwd'
 PROC_PATH = '/proc'
 
+IS_INSIDE_DOCKER = False
+
 if HOST_PREFIX:
```
andrewm4894, line 253: Is there ever a case where the HOST_PREFIX would be set but we not running in Docker, likewise not set but we actually are running in Docker? As much asking for my own curiosity.

andrewm4894 (approved):

ilyam8:
i consider them details

collectors/python.d.plugin/nvidia_smi/nvidia_smi.chart.py
ilyam8: Yes, but i see no reason to set `NETDATA_HOST_PREFIX` var for any other reason. Checking well-known env vars is common practice, for example see https://github.com/kubernetes/client-go/blob/30548acd0a9e231d42d406433aef55b111b36240/rest/config.go#L483-L491

ilyam8: > likewise not set but we actually are running in Docker

we mount host /proc, /sys, etc to /host/, see https://learn.netdata.cloud/docs/agent/packaging/docker#create-a-new-netdata-agent-container

thiagoftsm (approved):
LGTM!

PR #10287, opened 2020-11-25
web/gui: Added support for dynamic shading on dygraphs

dpsy4:
##### Summary
This change adds support for dynamic shading on a dygraph, based on the value of a dimension. The rationale for this change was to shade the area of a graph based on a dimension value, i.e. highlight when a value is below some target. 

To achieve this, the following new attributes are exposed on the graph component:
* `dygraph-shaderefdimension`: the reference dimension to apply shading
* `dygraph-shaderefthreshold`: a threshold value of the reference dimension to begin shading
* `dygraph-shaderefcomparewith`: the operation used to compare against the threshold. It accepts '<', '<=', '>', '>=', '=', '!=' and defaults to '<'
* `dygraph-shadecolor`: a color hex code for the shading

This works by hooking into the `underlayCallback` that is called before every re-draw of the graph. It checks to see if the reference dimension is defined in the dataset and compares that value to a defined threshold. If the comparison of the value and the threshold is truthy, it adds it to a record of shaded areas. Once all the data on the graph has been checked, it then draws a rectangle on the underlying canvas for each of the shaded regions that it collected earlier.

##### Component Name
web/gui
 
##### Test Plan
I created a graph with the following attributes set.
```
data-dygraph-shaderefdimension="actual"
data-dygraph-shaderefthreshold=2
data-dygraph-shaderefcomparewith=">"
data-dygraph-shadecolor="#0000FF1F"
```
I confirmed that the shading was applied when expected. I tried dragging the graphs back through time and confirmed that the shading was applied. Zooming in and out also scaled the shading appropriately. All other graphs on the page continued to be in in time sync with the graph.

Afterwards I tried the other comparison methods:  '<', '<=', '>', '>=', '=', '==', '===' , '!==',  '!=', '!' and confirmed that the shading applied as desired.

##### Additional Information
The comparison methods of '=', '==', '==='  are all compared using `===` and '!==',  '!=', '!' are all compared using `!==` under the hood. The additional formats were added to provide more flexibility to the input interface.

### DIFF
--- a/web/gui/dashboard.js
+++ b/web/gui/dashboard.js
@@ -3596,7 +3596,7 @@ NETDATA.gaugeChartCreate = function (state, data) {
         colorStart: startColor,     // Colors
         colorStop: stopColor,       // just experiment with them
         strokeColor: strokeColor,   // to see which ones work best for you
-        generateGradient: (generateGradient === true), // gmosx:
+        generateGradient: (generateGradient === true), // gmosx: 
         gradientType: 0,
         highDpiSupport: true        // High resolution support
     };
@@ -10104,7 +10104,7 @@ NETDATA.registry = {
                 }
                 NETDATA.registry.access(2, function (person_urls) {
                     NETDATA.registry.parsePersonUrls(person_urls);
-                });
+                });    
             }
         });
     },
@@ -10155,7 +10155,7 @@ NETDATA.registry = {
             // data.
             name = NETDATA.registry.hostname;
             url = NETDATA.serverDefault;
-        }
+        } 
 
         console.log("ACCESS", name, url);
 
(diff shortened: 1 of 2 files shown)


dpsy4:
Note, here is an image of it in practice.
Screen Shot 2020-11-25 at 2 48 37 PM

dpsy4:
I've noticed that the ESLint check is failing, but it does not seem to be related to the submitted code. Is that expected?

ilyam8:
@dpsy4 it is expected https://github.com/reviewdog/action-eslint/issues/56

dpsy4:
> @dpsy4 it is expected [reviewdog/action-eslint#56](https://github.com/reviewdog/action-eslint/issues/56)

Thanks.

dpsy4:
@jacekkolasa, how do the Netdata releases work? Before an upcoming release, is there a separate branch created or is master stabilized and used?

ilyam8:
@dpsy4, thanks for the contribution and apologies for the no response 😅.

A dashboard-related PR should be opened in https://github.com/netdata/dashboard.

PR #10293, opened 2020-11-27
dbengine extent cache

mfundul:
##### Component Name
database/engine
##### Test Plan

Choose a big enough time period in the netdata dashboard and create a snapshot.

Measure disk I/O and time to completion. Compare with the code in master.

You can measure disk I/O reads before and after the experiment by looking at `/proc/diskstats` and subtracting the values of the 6th column. Those are 512-byte sectors.

##### Additional Information
- The typical netdata query will only fetch 1 dimension from disk for a specific chart.
- Subsequent queries will fetch the rest of the dimensions.
- As a result, the extents will be read from disk multiple times.
A small extent cache will speed this up.

### DIFF
--- a/database/engine/rrdengine.c
+++ b/database/engine/rrdengine.c
@@ -108,6 +108,7 @@ static int try_insert_into_xt_cache(struct rrdengine_worker_config* wc, struct e
         xt_cache_elem = &xt_cache->extent_array[idx];
     }
     xt_cache_elem->extent = extent;
+    xt_cache_elem->fileno = extent->datafile->fileno;
     xt_cache_elem->inflight_io_descr = NULL;
     xt_cache_replaceQ_insert(wc, xt_cache_elem);
     modify_bit(&xt_cache->allocation_bitmap, idx, 1);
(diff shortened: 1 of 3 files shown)


database/engine/rrdengine.c
```diff
@@ -99,6 +273,34 @@ void read_extent_cb(uv_fs_t* req)
...
         /* care, we don't hold the descriptor mutex */
     }
+    {
+        uint8_t xt_is_cached = 0, xt_is_inflight = 0;
+        unsigned xt_idx;
+        struct extent_info *extent = xt_io_descr->descr_array[0]->extent;
+
+        xt_is_cached = !lookup_in_xt_cache(wc, extent, &xt_idx);
+        xt_is_inflight = check_bit(wc->xt_cache.inflight_bitmap, xt_idx);
+        if (xt_is_cached && xt_is_inflight) {
```
stelfrag, line 283: xt_idx is undefined if xt_is_cache is false

mfundul: It should be safe in practice but it's quite ugly so I'm fixing it.

stelfrag (approved):

database/engine/rrdengine.h
```diff
@@ -99,6 +100,27 @@ struct generic_io_descriptor {
     struct completion *completion;
 };
 
+struct extent_cache_element {
+    struct extent_info *extent; /* The ABA problem is avoided with the help of fileno below */
+    unsigned fileno;
+    struct extent_cache_element *prev; /* LRU */
+    struct extent_cache_element *next; /* LRU */
```
underhood, line 107: /* MRU */?

mfundul: Hmm, maybe I should just say `LRU list`?

underhood (approved):
all comments i would have are of "subjective" nature, so approving :)

PR #10296, opened 2020-11-30
python.d: always create a runtime chart on `create` call

ilyam8:
##### Summary

`alarms` collector has no charts after init/check.

As a result i get

```cmd
2020-11-30 17:30:50: netdata ERROR : PLUGINSD[python.d] : requested a BEGIN on chart 'netdata.runtime_alarms_local', which does not exist on host 'pc'. Disabling it. (errno 9, Bad file descriptor)
```

---

runtime chart is being created right after `check` and only if there are any charts 

https://github.com/netdata/netdata/blob/95407a625c6a55f157cdf8b0e59096d178098d41/collectors/python.d.plugin/python_modules/bases/FrameworkServices/SimpleService.py#L136-L178

https://github.com/netdata/netdata/blob/95407a625c6a55f157cdf8b0e59096d178098d41/collectors/python.d.plugin/python_modules/bases/charts.py#L43-L48

##### Component Name

`collectors/python.d/`

##### Test Plan

- [X] install the PR, ensure python.d.plugin is working

### DIFF
--- a/collectors/python.d.plugin/alarms/alarms.conf
+++ b/collectors/python.d.plugin/alarms/alarms.conf
@@ -20,7 +20,7 @@
 
 # update_every sets the default data collection frequency.
 # If unset, the python.d.plugin default is used.
-update_every: 5
+# update_every: 10
 
 # priority controls the order of charts at the netdata dashboard.
 # Lower numbers move the charts towards the top of the page.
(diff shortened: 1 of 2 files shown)


vlvkobal (approved):

thiagoftsm (approved):

andrewm4894 (commented):
lgtm

PR #10315, opened 2020-12-03
ACLK-NG

underhood:
##### Summary
This is the working implementation of ACLK-NG.For now a fallback for systems that can't use ACLK Legacy due to dependencies, etc.

Brings new ACLK implementation which requires only `OpenSSL` and `JSON-C` as external deps. JSON-C is compulsory for ACLK-NG.

don't forget to `git submodules update --init --recursive`

compile with `./netdata-installer.sh --aclk-ng`

verify all good after compile with: `./netdata -W buildinfo`. Should report:
```
Features:
    Netdata Cloud:           YES (Next Generation)
```
and `api/v1/info` should have new JSON key `aclk-implementation: "Next Generation"`

##### Component Name
ACLK and everything it touches

##### Test Plan
Absolute priority during the review is confirming ACLK Legacy still is built when it can and not broken in any way by this PR.
ACLK-NG due to its state is currently fallback only.

##### Additional Information
Coverity will always fail on this PR because it is from fork `underhood/netdata` do the Coverity credentials will fail (as only netdata/netdata has access). @Ferroin . El proof is here https://github.com/netdata/netdata/pull/10326

### DIFF
--- a/daemon/buildinfo.c
+++ b/daemon/buildinfo.c
@@ -8,7 +8,13 @@
 #ifdef ENABLE_ACLK
 #define FEAT_CLOUD 1
 #define FEAT_CLOUD_MSG ""
+#ifdef ACLK_NG
+#define ACLK_IMPL "Next Generation"
 #else
+#define ACLK_IMPL "Legacy"
+#endif
+#else
+#define ACLK_IMPL ""
 #ifdef DISABLE_CLOUD
 #define FEAT_CLOUD 0
 #define FEAT_CLOUD_MSG "(by user request)"
@@ -199,6 +205,9 @@ void print_build_info(void) {
     printf("    dbengine:                %s\n", FEAT_YES_NO(FEAT_DBENGINE));
     printf("    Native HTTPS:            %s\n", FEAT_YES_NO(FEAT_NATIVE_HTTPS));
     printf("    Netdata Cloud:           %s %s\n", FEAT_YES_NO(FEAT_CLOUD), FEAT_CLOUD_MSG);
+#if FEAT_CLOUD == 1
+    printf("    Cloud Implementation:    %s\n", ACLK_IMPL);
+#endif
     printf("    TLS Host Verification:   %s\n", FEAT_YES_NO(FEAT_TLS_HOST_VERIFY));
 
     printf("Libraries:\n");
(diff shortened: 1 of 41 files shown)


underhood:
@Ferroin is there easy way for CI to initialize git submodules? ACLK-NG uses git submodules and CI is failing here as the files are missing.

underhood:
rebased on latest aclk_2_aclk_legacy (on latest master) for CentOS 8 CI fixes

underhood:
@stelfrag rebased this on latest master to include your `chart` de-duplication fix

daemon/buildinfo.c
```diff
@@ -6,7 +6,11 @@
 // Optional features
 
 #ifdef ENABLE_ACLK
-#define FEAT_CLOUD "YES"
+#ifdef ACLK_NG
+#define FEAT_CLOUD "YES (Next Generation)"
```
Ferroin, line 10: We should probably come up with a name other than ‘next generation’ for this. Perhaps we should just start versioning the low-level transport implementation?

underhood: We could change that in the next PRs I guess. My plan was after a time Legacy will be removed completely and this just becomes "ACLK" without additional designations

Ferroin: Yeah, that does make sense. I just want to avoid having generic naming bite us in the arse down the road.

On a side note, the version that completely removes the old transport layer should probably be a major version bump. Regardless of whether we have an on-premises Cloud offering at that point or not, it’s a breaking change and should be handled as such.

underhood: I generally agree with you but would think about this as separate PR + we need to keep a way, not to only version ACLK but possibly have alternative implementation ACLK Legacy/NG Timo/NG Paho.... so it is not always a newer version only currently Legacy and NG are on-par feature-wise.

Ferroin: In that case, perhaps ‘ACLK (simplified)’ for this one? AFAICT this new implementation is significantly simpler than the legacy implementation that needs mosquitto and LWS.

underhood: Well, I do not object to renaming if consensus is found but would not like to mess up this PR as it is huge already. But open to doing that in the future.

underhood:
rebased to resolve conflicts after some recently merged PRs

stelfrag:
I passed `--disable-cloud --aclk-ng` by accident it turns out that produces compiler errors.

underhood:
added explicit check for that combination and a user-friendly error message will be printed by `configure` if that is used

MrZammler (approved):

Issue #10321, opened 2020-12-04
Haproxy python plugin


ashish1099:
Issue reported but closed https://github.com/netdata/netdata/issues/2189

##### OS / Environment
Xenial 16.04

##### Netdata version
1.26

##### Component Name
Haproxy python.d plugin

##### Steps To Reproduce
1. enable haproxy stats under any frontend
2. enable the haproxy plugin
```
http://127.0.0.1:63661/haproxy?stats;csv;norefresh
```
3. monitor the haproxy log and you would two entries 
```
2020-12-03T03:18:02+01:00 localhost haproxy[19080]: 127.0.0.1:28490 [03/Dec/2020:03:18:02.642] netdata-monitoring netdata-monitoring/<NOSRV> 0/-1/-1/-1/0 503 237 - - SC-- 182/1/0/0/0 0/0 "GET /haproxy HTTP/1.1"
2020-12-03T03:18:12+01:00 localhost haproxy[19080]: 127.0.0.1:28824 [03/Dec/2020:03:18:12.631] netdata-monitoring netdata-monitoring/<STATS> 0/0/0/0/1 200 21725 - - LR-- 180/1/0/0/0 0/0 "GET /haproxy?stats;csv;norefresh HTTP/1.1"
```

##### Expected behavior
when setting up the haproxy config to use via_url, it tries to hit the haproxy twice.

1. stats
2. info

but in case of `info` there is no **info** in the haproxy stats page, not sure why we are doing this and we can not get `info` stats from the stats page.
it's resulting in 503 all the time.

ilyam8:
@ashish1099 hi 👋  thx for reporting!

I checked haproxy code and to be honest a little bit confused 😄 


Could you try to change `_get_data` to

```python
    def _get_data(self):
        to_netdata = dict()
        self.request, self.url = self.options_['stat'], self.options_['stat']
        stat_data = self._get_stat_data()
        if self.poll is SocketService:
            self.request, self.url = self.options_['info'], self.options_['info']
            info_data = self._get_info_data(regex=self.options_['regex'])
            to_netdata.update(info_data)
        to_netdata.update(stat_data)
        return to_netdata or None
```

and check if it fixes the problem?

ilyam8:
Oh, i see

https://github.com/netdata/netdata/issues/2189#issuecomment-301779436

Module does HTP requests
 - `http://ip.address:port/haproxy_stats;csv`: stats in csv format.
[...]

PR #10322, opened 2020-12-04
haproxy pythond plugin doc fix when using the plugin `via_url`

ashish1099:
Fixes: #10321

### DIFF
--- a/collectors/python.d.plugin/haproxy/README.md
+++ b/collectors/python.d.plugin/haproxy/README.md
@@ -12,7 +12,8 @@ And health metrics such as backend servers status (server check should be used).
 Plugin can obtain data from url **OR** unix socket.
 
 **Requirement:**
-Socket MUST be readable AND writable by the `netdata` user.
+Socket: MUST be readable AND writable by the `netdata` user.
+URL: MUST have `stats uri <path>` present in the haproxy config, otherwise you will 503 in the haprooxy logs, more info [here](https://github.com/netdata/netdata/issues/10321)
 
 It produces:
 


thiagoftsm (approved):
Considering what was discussed [here](https://github.com/netdata/netdata/issues/10321#issuecomment-738738340), I am approving the PR. Thank you @ashish1099 !

collectors/python.d.plugin/haproxy/README.md
```diff
@@ -12,7 +12,8 @@ And health metrics such as backend servers status (server check should be used).
 Plugin can obtain data from url **OR** unix socket.
 
 **Requirement:**
-Socket MUST be readable AND writable by the `netdata` user.
+Socket: MUST be readable AND writable by the `netdata` user.
```
ilyam8, line 15: @joelhans please review, i am not sure about uppercase letters 🤔

collectors/python.d.plugin/haproxy/README.md
```diff
@@ -12,7 +12,8 @@ And health metrics such as backend servers status (server check should be used).
...
+URL: MUST have `stats uri <path>` present in the haproxy config, otherwise you will 503 in the haprooxy logs, more info [here](https://github.com/netdata/netdata/issues/10321)
```
runejuhl, line 16:
```suggestion
URL: MUST have `stats uri <path>` present in the haproxy config, otherwise you
will get HTTP 503 in the haproxy logs, more info [here](https://github.com/netdata/netdata/issues/10321)
```

collectors/python.d.plugin/haproxy/README.md
joelhans, line 15: I would prefer we not use uppercase for emphasis. I'll suggest we use no emphasis, but @ashish1099 if you believe emphasis is important for users reading this doc, you can use italics `_must_` or bold `**must**`.

```suggestion
Socket: Must be readable and writeable by the `netdata` user.
```

joelhans (changes requested):

collectors/python.d.plugin/haproxy/README.md
runejuhl: The casing by @ashish1099 follows the pattern established by RFCs (where `MUST`, `MAY`, `SHALL` etc. have particular meanings), and has the benefit over Markdown formatting that it works in plaintext (e.g. without a renderer) as well as in rendered form. Perhaps a compromise?

```suggestion
Socket: **MUST** be readable **AND** writable by the `netdata` user.
```

ilyam8: Why do you think we need to follow the pattern established by RFCs if the readme is not RFC?

joelhans: The Netdata [style guide](https://learn.netdata.cloud/contribute/style-guide#capitalization) explicitly says to not use capitalization for emphasis in documentation. I know there are places there the rules are still not followed, but let's say those are "grandfathered" in. All _new_ changes should follow the style guide.

runejuhl: > Why do you think we need to follow the pattern established by RFCs if the readme is not RFC?

Simply to follow an already established convention. I wasn't aware of the style guide, but that should of course be followed.

Issue #10165, opened 2020-10-30
exporting.conf seems not working 


mihailigoshin:
This is not working:

` /etc/netdata/edit-config exporting.conf
`
```

[graphite:my_graphite]
    enabled = yes
    destination = udp:influxdb.home:2003
    data source = average
    prefix = netdata
    hostname = lxc04
    #update every = 10
```

PS.  I've tried via tcp but for sure influxdb graphite is via udp

This is working:

`/etc/netdata/edit-config netdata.conf`

```
[backend]
        # host tags = 
         enabled = yes
        # data source = average
         type = graphite
         destination = udp:influxdb.home:2003
         prefix = netdata
         hostname = lxc04
```

Here is the exact configuration from influxdb.conf

```
[[graphite]]
  # Determines whether the graphite endpoint is enabled.
    enabled = true
    database = "graphite"
   #retention-policy = ""
    bind-address = ":2003"
    protocol = "udp"
   #consistency-level = "one"
```

Influxdb is on a remote machine 
BTW I've tried with opentsdb and got the same results

ilyam8:
Hi @mihailigoshin, netdata version?

mihailigoshin:
@ilyam8 Hi, 

netdata v1.26.0

vlvkobal:
I've tried with
```
[graphite:instance_graphite_udp]
enabled = yes
destination = udp:localhost:2003
data source = average
update every = 2
prefix = netdata
hostname = lsc04
send charts matching = example.random
```
and
```
    enabled = true
    database = "graphite"
    bind-address = ":2003"
    protocol = "udp"

```
everything is working as expected 🤷‍♂️.

vlvkobal:
Please check the Netdata part using `nc -l -p 2003`.

vlvkobal:
Well, I'm closing this. I can't reproduce the issue and there were quite considerable changes to the exporting engine during the last days. @mihailigoshin if you still face the problem after updating Netdata and you are able to provide more information, feel free to reopen the issue.

fayak:
I have the same issue. On archlinux, install netdata with `pacman` -> (`netdata v1.26.0`).

Go to `/etc/`, `./edit-config exporting.conf`, and set

```
[exporting:global]
    enabled = yes
    # send configured labels = yes
    # send automatic labels = no
    # update every = 10

[prometheus:exporter]
    # send names instead of ids = yes
    # send configured labels = yes
    send automatic labels = yes
    # send charts matching = *
[...]

PR #10323, opened 2020-12-04
Fix exporting config

vlvkobal:
##### Summary
The `prefix` and `send names instead of ids` options didn't work in the `prometheus:exporter` section of `exporting.conf`. The options should also inherit values from the `backend` section of `netdata.conf` if they are not set in `exporting.conf`. The PR fixes the issue.

##### Component Name
exporting engine, Web API

##### Test Plan
Check the `http://localhost:19999/api/v1/allmetrics?format=prometheus` endpoint with different settings of `prefix` and `send names instead of ids` in `exporting.conf` and `netdata.conf`.

### DIFF
--- a/exporting/read_config.c
+++ b/exporting/read_config.c
@@ -238,7 +238,8 @@ struct engine *read_exporting_config()
         prometheus_exporter_instance->config.update_every =
             prometheus_config_get_number(EXPORTING_UPDATE_EVERY_OPTION_NAME, EXPORTING_UPDATE_EVERY_DEFAULT);
 
-        if (prometheus_config_get_boolean("send names instead of ids", CONFIG_BOOLEAN_YES))
+        if (prometheus_config_get_boolean(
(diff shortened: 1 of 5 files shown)


fayak:
Will it also fix the lack of exported `host tags` / `send configured labels` as per discussed brievly in https://github.com/netdata/netdata/issues/10165 ?

vlvkobal:
I tested `send configured labels` and other options as well. I didn't find any problems with them.

thiagoftsm (approved):
I tested all combinations suggested now and I had `prefix` for all options.

fayak:
Awesome, thanks a lot !

pickfire:
Do we have to wait for the next version to get this?

OdysLam:
@pickfire welcome to the community!

A new release is just around the corner, but I am not sure if this PR will be included. Let's see what the rest of the team has to say. cheers :)

stelfrag (approved):

stevenans9859 (approved):

PR #10324, opened 2020-12-04
Improve bug_report.md template

OdysLam:

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,7 +8,7 @@ labels: bug, needs triage
 When creating a bug report please:
 - Verify first that your issue is not already reported on GitHub.
 - Verify that you are using the latest version of the Netdata Agent, and if not, update: https://learn.netdata.cloud/docs/agent/packaging/installer/update
-- If you are unsure if it's a bug, post a question on our support forums at https://community.netdata.cloud/c/support/agent-support/14
+- If you're not sure your issue is a bug, post a question on our support forums: https://community.netdata.cloud/c/support/agent-support/14
 -->
 
 ##### Bug report summary
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,7 @@ labels: bug, needs triage
 <!--
 When creating a bug report please:
 - Verify first that your issue is not already reported on GitHub.
-- Verify that you are using the latest version of the Netdata Agent and if not update. You can read more about it in our documentation at https://learn.netdata.cloud/docs/agent/packaging/installer/update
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,7 +7,8 @@ labels: bug, needs triage
 <!--
 When creating a bug report please:
 - Verify first that your issue is not already reported on GitHub.
-- Test if the latest release and master branch are affected too.
+- Verify that you are using the latest version of the Netdata Agent and if not update. You can read more about it in our documentation at https://learn.netdata.cloud/docs/agent/packaging/installer/update
```
ilyam8, line 10: That is slightly different sense for me. The gist is - if a user uses stable netdata we want him to verify that the bug is present in the master branch.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,7 +7,8 @@ labels: bug, needs triage
...
+- If you are unsure if it's a bug, post a question on our support forums at https://community.netdata.cloud/c/support/agent-support/14
```
joelhans, line 11:
```suggestion
- If you're not sure your issue is a bug, post a question on our support forums: https://community.netdata.cloud/c/support/agent-support/14
```

.github/ISSUE_TEMPLATE/bug_report.md
joelhans: Good point. Let's figure this out before following my suggestion below.

OdysLam: How can a user easily understand if they are on the master branch? I want to assume no prior knowledge of Git.

ilyam8: Checking netdata version via cmd line (netdata -v) or gui (there is update button on the top bar and if you click it there is Your `netdata version: ...`)

- stable version format: `v1.26.0` (major.minor.patch)
- nigthly version format: `v1.26.0-315-g6335324a` (major.minor.patch-numOfCommitsAfterTheTag-lastCommitHash)

OdysLam: So, we ask the user to verify that they are using the latest version of the Netdata Agent? My question is how to rephrase the "latest releast/master branch" into something simpler "check netdata version with netdata -v.

ilyam8: It is a little bit more complex. Using stable releases is a common pattern, but before submitting the bug it would be nice to ensure that the bug is not fixed in the master: **look for open/closed bugs in the repo**.

Optional for stable release users (if feels boring/adventurous): install master branch and reproduce the bug.

ilyam8: Which is covered in 

> Verify first that your issue is not already reported on GitHub.

So there is only optional part.

PR #10329, opened 2020-12-06
Added number of allocated/stored objects within each Varnish storage

ernestojpg:
##### Summary

Currently, for each of the Varnish Storages (SMF, SMA, or MSE) we are only providing the free and allocated space in KiB. This Pull Request is also adding a new graph with the number of allocated/stored objects within each storage.

##### Component Name

Affected component: Varnish collector (python.d plugin)

##### Test Plan

I have tested the change in Varnish-Cache 6.4.0 (Free version) and Varnish-Plus 6.0.6r10 (Commercial/Enterprise version).
The new added field `g_alloc` is available [at least from Varnish 5.0](https://varnish-cache.org/docs/5.0/reference/varnish-counters.html#per-malloc-storage-counters-sma).

##### Additional Information

This is an example of the new graphs for the `s0` and `Transient` (SMA) storages:

Screenshot 2020-12-06 at 19 08 41

### DIFF
--- a/collectors/python.d.plugin/varnish/varnish.chart.py
+++ b/collectors/python.d.plugin/varnish/varnish.chart.py
@@ -155,7 +155,7 @@ def backend_charts_template(name):
     return order, charts
 
 
-def storage_charts_template(name):
+def storage_usage_charts_template(name):
     order = [
         'storage_{0}_usage'.format(name),
     ]
@@ -172,6 +172,21 @@ def storage_charts_template(name):
 
     return order, charts
 
+def storage_alloc_objs_charts_template(name):
+    order = [
+        'storage_{0}_alloc_objs'.format(name),
+    ]
+
+    charts = {
+        order[0]: {
+            'options': [None, 'Storage "{0}" Allocated Objects'.format(name), 'objects', 'storage usage', 'varnish.storage_usage', 'line'],
+            'lines': [
+                ['{0}.g_alloc'.format(name), 'allocated', 'absolute']
+            ]
+        },
+    }
+
+    return order, charts
 
 VARNISHSTAT = 'varnishstat'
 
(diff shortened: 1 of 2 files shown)


collectors/python.d.plugin/varnish/varnish.chart.py
```diff
@@ -172,6 +172,21 @@ def storage_charts_template(name):
 
     return order, charts
 
+def storage_alloc_objs_charts_template(name):
+    order = [
```
ilyam8, line 176: How about to add this chart to the `storage_charts_template` so it returns both?

ernestojpg: Yes, I could also do that. Do you prefer in that way?

ernestojpg: Done

collectors/python.d.plugin/varnish/varnish.chart.py
```diff
@@ -168,11 +169,16 @@ def storage_charts_template(name):
                 ['{0}.g_bytes'.format(name), 'allocated', 'absolute', 1, 1 << 10]
             ]
         },
+        order[1]: {
+            'options': [None, 'Storage "{0}" Allocated Objects'.format(name), 'objects', 'storage usage', 'varnish.storage_usage', 'line'],
```
ernestojpg, line 173: I'm not sure about the `varnish.storage_usage` parameter. I used the same as in the previous chart, because both are in the "storage usage" section. Is it correct?

ilyam8: It is [chart `context`](https://learn.netdata.cloud/docs/agent/collectors/plugins.d#chart), i think it should be `varnish.storage_alloc_objs`

ernestojpg: Sorry for the delay @ilyam8 . It is updated now. Does this "chart context" need to be unique? Because if we have this chart for each of the storages, then we will have multiple `varnish.storage_alloc_objs`, right?

ilyam8: No, it suposed to be not unique so you can use it in an alarm template and apply alarms for all the chart with the same context.

ilyam8 (approved):

thiagoftsm (approved):

collectors/python.d.plugin/varnish/varnish.chart.py
ernestojpg: Cool, thanks @ilyam8

ilyam8:
Thx for contributing @ernestojpg 👍

PR #10344, opened 2020-12-08
Fixed handling of PowerTools repo on CentOS 8.

Ferroin:
##### Summary

They appear to have changed DNF to be case-sensitive for repo naming, and the repo name that is expected is all lower case.

##### Component Name

area/packaging

##### Test Plan

Verified through local testing in a CentOS 8 container. Without this change, installing Netdata from sources with `netdata-installer.sh` fails in a clean environment because it cannot install certain required dependencies. With this change, it works correctly.

### DIFF
--- a/packaging/installer/install-required-packages.sh
+++ b/packaging/installer/install-required-packages.sh
@@ -1550,7 +1550,7 @@ validate_tree_centos() {
     echo >&2 " > Checking for PowerTools ..."
     if ! run yum ${sudo} repolist | grep PowerTools; then
       if prompt "PowerTools not found, shall I install it?"; then
-        run ${sudo} yum ${opts} config-manager --set-enabled PowerTools
+        run ${sudo} yum ${opts} config-manager --set-enabled powertools
       fi
     fi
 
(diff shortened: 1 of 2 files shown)


packaging/installer/install-required-packages.sh
```diff
@@ -1550,7 +1550,7 @@ validate_tree_centos() {
     echo >&2 " > Checking for PowerTools ..."
     if ! run yum ${sudo} repolist | grep PowerTools; then
```
underhood, line 1551: no need to update grep as well or `grep -i PowerTools` ?

Ferroin: Nope, the full name is in the output of `yum repolist`, which includes the capitalized version. And even if it gets a false negative it, the config-manager command below is idempotent.

underhood (approved):

stelfrag (approved):
I tested and it now  matches `PowerTools` on the description

`powertools    CentOS Linux 8 - PowerTools` 

now

Issue #10325, opened 2020-12-04
Agent installation script fails to compile libJudy on OSX


krital:
##### Bug report summary

Agent installation script fails to compile libJudy on OSX because it tries to use libtool and libtoolize. As Apple bundles its own libtool, to avoid conflicts brew renames it to glibtool (and glibtoolize) on osx. So we need the following to happen:

1. Change the install script to call glibtool / glibtoolize when running OS=`Darwin'
2. Modify the documentation to include libtool / libtoolize in the brew dependencies needed on OSX before we execute the script.

##### OS / Environment

```
Darwin Alexs-MBP 19.6.0 Darwin Kernel Version 19.6.0: Thu Oct 29 22:56:45 PDT 2020; root:xnu-6153.141.2.2~1/RELEASE_X86_64 x86_64
ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H15
```

##### Component Name

Agent install script
Documentation for OSX page

##### Steps To Reproduce

1. sudo ./netdata-installer.sh --install /usr/local
2. libJudy compilation fails with the following error:

env: libtoolize: No such file or directory

##### Expected behavior

Expect to call glibtoolize when OS=Darwin
Prerequisites installable by brew before running the script should include libtool / libtoolize

krital:
any chance we can include this in the release? should be a very simple change only for OSX

PR #10346, opened 2020-12-08
Use `glibtoolize` on macOS instead of regular `libtoolize`.

Ferroin:
##### Summary

This ensures we get the version from Homebrew instead of the version from XCode.

##### Component Name

area/packaging

##### Test Plan

Without this change, Judy fails to build on macOS, and thus we have no dbengine. With this change, Judy builds correctly, and the dbengine is available.

##### Additional Information

Fixes: #10325

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -695,13 +695,18 @@ bundle_libwebsockets
 
 build_judy() {
   local env_cmd=''
+  local libtoolize="libtoolize"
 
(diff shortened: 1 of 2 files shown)


krital (approved):
This builds libJudy without problems now

kaskavel (approved):

krital (approved):
Looks great to me

joelhans (approved):
Docs change LGTM—thanks for handling that, @Ferroin!

kaskavel (approved):

PR #10349, opened 2020-12-09
Fix recursive invocation at declaration

tickez:
##### Summary
This creates an endless recursive invocation resulting in the machine becoming unusable.
If called from the /etc/cron.daily it luckily only results in
`/etc/cron.daily/netdata-updater: line 31: /etc/cron.daily/netdata-updater.sh: No such file or directory`

##### Component Name
Updater

##### Test Plan
None, just executed the updater after the fix.

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -28,7 +28,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


knatsakis (approved):

Ferroin:
The same issue has been fixed in https://github.com/netdata/netdata/pull/10352, alongside a couple of other bugs in the updater.

@tickez Thank you for this PR even though we did not end up actually using it.

Issue #10351, opened 2020-12-09
daily and manual netdata update fails


rbike:
##### Bug report summary
Daily netdata update cron job reports error 127
Manual netadata update with "bash <(curl -Ss https://my-netdata.io/kickstart.sh)" is not working (runs indefinitely)

##### OS / Environment
/etc/lsb-release:DISTRIB_ID=Ubuntu
/etc/lsb-release:DISTRIB_RELEASE=20.04
/etc/lsb-release:DISTRIB_CODENAME=focal
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="20.04.1 LTS (Focal Fossa)"
/etc/os-release:ID=ubuntu
/etc/os-release:ID_LIKE=debian
/etc/os-release:PRETTY_NAME="Ubuntu 20.04.1 LTS"
/etc/os-release:VERSION_ID="20.04"
/etc/os-release:HOME_URL="https://www.ubuntu.com/"
/etc/os-release:SUPPORT_URL="https://help.ubuntu.com/"
/etc/os-release:BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
/etc/os-release:PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
/etc/os-release:VERSION_CODENAME=focal
/etc/os-release:UBUNTU_CODENAME=focal

##### Netdata version
netdata v1.26.0-333-nightly

##### Component Name
Updater

##### Steps To Reproduce
1. I have installed netdata using the automatic online installation script "bash <(curl -Ss https://my-netdata.io/kickstart.sh)". 
2. Everything was running fine for weeks
3. Since this night I get the following error message from the daily netdata update cron job:
```
/etc/cron.daily/netdata-updater: line 31: /etc/cron.daily/netdata-updater.sh: No such file or directory
run-parts: /etc/cron.daily/netdata-updater exited with return code 127
```
4. I tried a manual update with "bash <(curl -Ss https://my-netdata.io/kickstart.sh)". This results in the following cmdline output:
```
 --- Found existing install of Netdata under: / ---
 --- Attempting to update existing install instead of creating a new one ---
[/root]# //usr/libexec/netdata/netdata-updater.sh --not-running-from-cron
```
and then nothing happens. I had to stop the process with Ctrl-C after some time.
[...]

PR #10352, opened 2020-12-09
Fixed handling of self-updating in updater script.

Ferroin:
##### Summary

There’s a bug in the code used to locate the script itself (used for determining whether to self-update or not) that causes either an infinite loop or an immediate failure to run right after invoking the script depending on how it’s invoked.

Also fixes a typo in one of the option names.

##### Component Name

area/packaging

##### Test Plan

Same testing regimen as #10261, which now seems to correctly fail for the broken code now. When the timestamp is explicitly overridden it should get stuck in an infinite loop _after_ checking for a new version of the updater script (because it will pull the version with the bug).

##### Additional Information

Fixes: #10351 

Not sure how this made it past manual testing. Checking now it seems to not work correctly at all (gets stuck in an infinite loop), and looking back at the code it was obviously broken from the start. Ironically though, this further highlights just how useless the lifecycle checks in Trais were, as they should have caught this too but did not.

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -150,7 +150,7 @@ newer_commit_date() {
   echo >&3 "Checking if a newer version of the updater script is available."
 
   if command -v jq > /dev/null 2>&1; then
-    commit_date="$(_safe_download "https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1" /dev/stdout | jq '.[0].commit.committer.date')"
+    commit_date="$(_safe_download "https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1" /dev/stdout | jq '.[0].commit.committer.date' | tr -d '"')"
   elif command -v python > /dev/null 2>&1;then
     commit_date="$(_safe_download "https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1" /dev/stdout | python -c 'from __future__ import print_function;import sys,json;print(json.load(sys.stdin)["commit"]["committer"]["date"])')"
   fi
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -176,6 +176,7 @@ self_update() {
     cd "$ndtmpdir" || exit 1
 
     if _safe_download "https://raw.githubusercontent.com/netdata/netdata/master/packaging/installer/netdata-updater.sh" ./netdata-updater.sh; then
+      chmod +x ./netdata-updater.sh || exit 1
       exec ./netdata-updater.sh --not-running-from-cron --no-self-update --tmpdir-path "$(pwd)"
     else
       echo >&3 "Failed to download newest version of updater script, continuing with current version."
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -156,7 +156,7 @@ newer_commit_date() {
   fi
 
   if [ -z "${commit_date}" ] ; then
-    commit_date="1970-01-01T00:00:00Z"
+    commit_date="9999-12-31T23:59:59Z"
   fi
 
   if [ -e "${script_source}" ]; then
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -152,7 +152,7 @@ newer_commit_date() {
   if command -v jq > /dev/null 2>&1; then
     commit_date="$(_safe_download "https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1" /dev/stdout | jq '.[0].commit.committer.date' | tr -d '"')"
   elif command -v python > /dev/null 2>&1;then
-    commit_date="$(_safe_download "https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1" /dev/stdout | python -c 'from __future__ import print_function;import sys,json;print(json.load(sys.stdin)["commit"]["committer"]["date"])')"
+    commit_date="$(_safe_download "https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1" /dev/stdout | python -c 'from __future__ import print_function;import sys,json;print(json.load(sys.stdin)[0]["commit"]["committer"]["date"])')"
   fi
 
   if [ -z "${commit_date}" ] ; then
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -28,7 +28,13 @@
 
 set -e
 
-script_source="$("$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd -P)/netdata-updater.sh")"
+script_dir="$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd -P)"
+
+if [ -x "${script_dir}/netdata-updater" ]; then
+  script_source="${script_dir}/netdata-updater"
+else
+  script_source="${script_dir}/netdata-updater.sh"
+fi
 
 info() {
   echo >&3 "$(date) : INFO: " "${@}"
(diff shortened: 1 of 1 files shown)


knatsakis (approved):

packaging/installer/netdata-updater.sh
```diff
@@ -153,7 +159,13 @@ newer_commit_date() {
     commit_date="1970-01-01T00:00:00Z"
```
mfundul, line 159: Shouldn't this be something very far in the future, so that the default behaviour is to always overwrite the updater if something went wrong?

thiagoftsm (dismissed):
I had warning messages related to Python, as we discussed on Slack, but after to remove my `/usr/sbin/netdata`, I observed that I could run the update script and it did not have an infinite loop. The installation was successful .

mfundul:
I run the installer in your fork, is this a problem?
```
./netdata-installer.sh: line 957: remote: No such file or directory
```

packaging/installer/netdata-updater.sh
Ferroin: Actually, yes, I somehow got the logic backwards here.

mfundul:
It doesn't seem to be working for me, `commit_date` is empty:
```
+ self_update
+ '[' -z '' ']'
+ newer_commit_date
+ echo 'Checking if a newer version of the updater script is available.'
+ command -v jq
+ command -v python
++ _safe_download 'https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1' /dev/stdout
++ url='https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1'
++ python -c 'from __future__ import print_function;import sys,json;print(json.load(sys.stdin)["commit"]["committer"]["date"])'
++ dest=/dev/stdout
++ command -v curl
++ curl -sSL --connect-timeout 10 --retry 3 'https://api.github.com/repos/netdata/netdata/commits?path=packaging%2Finstaller%2Fnetdata-updater.sh&page=1&per_page=1'
++ return 0
  File "<string>", line 1, in <module>
TypeError: list indices must be integers or slices, not str
+ commit_date=
+ '[' -z '' ']'
+ commit_date=1970-01-01T00:00:00Z
+ '[' -e /usr/libexec/netdata/netdata-updater.sh ']'
++ date -r /usr/libexec/netdata/netdata-updater.sh +%s
+ script_date=1607530781
++ date -d 1970-01-01T00:00:00Z +%s
+ '[' 0 -ge 1607530781 ']'
```

packaging/installer/netdata-updater.sh
Ferroin: Updated now to default to a date long past when we will all be dead for the last commit date. If Netdata is still around in seven thousand nine hundred and seventy nine years and this code is still being used, our distant descendants can deal with any issues this might cause then (alongside all the other issues that may exist at that time due to most ISO 8601 date parsing code assuming a four digit year).

Ferroin:
@mfundul there’s a bug in the handling of the Python code. I probably need to just restrict it to Python 3...

vlvkobal:
```
[vlad@arch-esxi netdata]$ sudo  /usr/libexec/netdata/netdata-updater.sh
  File "<string>", line 1, in <module>
TypeError: list indices must be integers or slices, not str
/usr/libexec/netdata/netdata-updater.sh: line 179: /tmp/netdata-updater-ZBSwPuTkVK/netdata-updater.sh: Permission denied
[vlad@arch-esxi netdata]$
```

mfundul:
Problem with permissions:
```
/usr/libexec/netdata/netdata-updater.sh: line 179: /tmp/netdata-updater-XZQ2N2xwlv/netdata-updater.sh: Permission denied
```
```
[root@localhost netdata]# ls -lht /tmp/netdata-updater-XZQ2N2xwlv/netdata-updater.sh
-rw-r--r--. 1 root root 13K Dec  9 18:38 /tmp/netdata-updater-XZQ2N2xwlv/netdata-updater.sh
```

Ferroin:
@mfundul you probably have `/tmp` mounted `noexec`. We should not be throwing an error in that case, but...

@vlvkobal Please check that you have the newest version of the branch, I’ve pushed some fixes that should prevent that error from happening.

thiagoftsm (dismissed):
I retest now and this time I did not have any warning or error messages.

mfundul:
> 
> 
> @mfundul you probably have `/tmp` mounted `noexec`. We should not be throwing an error in that case, but...
> 
It's standard vanilla Fedora 31:
```
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,seclabel)
```

underhood:
@mfundul @Ferroin same issue on ubuntu server 1604
Screenshot from 2020-12-09 17-55-42

cakrit:
We shouldn't be running it from `/tmp`. Respect the `noexec` please.

underhood:
LGTM after cc4161d

underhood (approved):
after cc4161d LGTM. CI failure is unrelated (I think).

mfundul:
It can't fork or something:
```
+ chmod +x ./netdata-updater.sh
++ pwd
+ exec ./netdata-updater.sh --not-running-from-cron --no-self-update --tmpdir-path /tmp/netdata-updater-25K19hIltG
/tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: retry: Resource temporarily unavailable
/tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: retry: Resource temporarily unavailable
/tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: retry: Resource temporarily unavailable
/tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: retry: Resource temporarily unavailable
/tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: Resource temporarily unavailable
/tmp/netdata-updater-25K19hIltG/netdata-updater.sh: line 31: /netdata-updater.sh: No such file or directory
```

Ferroin:
> It can't fork or something:
> 
> ```
> + chmod +x ./netdata-updater.sh
> ++ pwd
> + exec ./netdata-updater.sh --not-running-from-cron --no-self-update --tmpdir-path /tmp/netdata-updater-25K19hIltG
> /tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: retry: Resource temporarily unavailable
> /tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: retry: Resource temporarily unavailable
> /tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: retry: Resource temporarily unavailable
> /tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: retry: Resource temporarily unavailable
> /tmp/netdata-updater-25K19hIltG/netdata-updater.sh: fork: Resource temporarily unavailable
> /tmp/netdata-updater-25K19hIltG/netdata-updater.sh: line 31: /netdata-updater.sh: No such file or directory
> ```

Expected behavior. It’s pulling he version that has the bug when it tries to update itself.

mfundul (approved):

thiagoftsm (approved):
Everything is working on Slackware Current.

webdock-io:
How do we recover from this if we have a vast sleuth of servers with Netdata installed? 

Running `bash <(curl -Ss https://my-netdata.io/kickstart.sh) --no-updates` just invokes the same updater script.

Basically I am looking for a non-interactive way to fix this across our many many servers so we can get back to having updated versions of Netdata ... Would we need to patch the netdata-updater script ourselves? Remove Netdata entirely and start over on the install? This would require manual interaction and would take lots of time to do, so I'd like to avoid that if possible.

Ferroin:
> How do we recover from this if we have a vast sleuth of servers with Netdata installed?
> 
> Running `bash <(curl -Ss https://my-netdata.io/kickstart.sh) --no-updates` just invokes the same updater script.
> 
> Basically I am looking for a non-interactive way to fix this across our many many servers so we can get back to having updated versions of Netdata ... Would we need to patch the netdata-updater script ourselves? Remove Netdata entirely and start over on the install? This would require manual interaction and would take lots of time to do, so I'd like to avoid that if possible.

You can update existing systems by re-running the installer script with the same options you originally used, and adding `--reinstall` to the end of the list of options. This will cause it to skip trying to run the updater, and do a clean install while preserving your local data and configuration.

webdock-io:
@Ferroin That's great, and seems to work - it's not hands-free / non-interactive though. So I can't script this. And then I'd need to manually update many many hosts - which would take up the remainder of my day :) Any way around that chore?

Ferroin:
@webdock-io If you had manually set up all of those hosts, then there’s not really any good option here short of patching the updater with the changes in this script (but that _should_ be easily scriptable if they’re all up to date) and then update as normal.

Otherwise, you should be able to temporarily add the `--reinstall` option in whatever deployment/orchestration tool you originally used and re-run it against all the relevant hosts.

One quick note, there’s a bug right now in the option parsing code in the kickstart install script that results in it being sensitive to the order of the options passed to it when it should not be. Because of this, the `--reinstall` option will need to be the first option passed. See https://github.com/netdata/netdata/pull/10396 for more info on the issue itself.

webdock-io:
I'll script up a patch and then run the script in order to achieve a non-interactive upgrade which I can deploy across all our hosts.

Probably won't get around to it anytime soon - but thank you for offering a confirmation on a solution :)

PS: This bug really triggered a whole boatload of alarms for us as the script forked to infinity and made everything rather unstable - on all host, at the exact same time :D (# panic) I am glad to see you are implementing better testing, as this should not have made it into the wild.

PR #10354, opened 2020-12-09
Rename abs to ABS to avoid clash with standard definitions. Fixes #10353.

### LINKED ISSUE #10353: Redefinition of abs() macro causes problems with C++ plugins.
##### Bug report summary
Related to #10178. When using libnetdata from C++, it is easy to hit a problem with redefinition of abs macro in libnetdata.h, as C++ does not allow this kind of redefinitions. Also, it can be a bit misleading to redefine "standard" functions.

##### Netdata version
master from git 

##### Component Name
libnetdata, netdata core.

##### Steps To Reproduce
Include libnetdata header in C++ source along with some 'pure' C++ headers, e.g. boost/regex.hpp. Compilation of the C++ code fails with clashing names.

KickerTom:
##### Summary
Fixes: #10353
I have renamed the abs macro to ABS, similarly to already existing MIN and MAX macros. I have replaced all ocurences in the code (although it compiled fine with the "standard" macro in the gcc headers) to use this "internal" version. Resolves the issue described in #10353 for me.

##### Component Name
libnetdata, netdata core.

##### Test Plan
Code compiles natively on amd64 linux, cross compiles on aarch64. Native linux version tested fine.

### DIFF
--- a/daemon/unit_test.c
+++ b/daemon/unit_test.c
@@ -371,7 +371,7 @@ int unit_test_str2ld() {
                 return -1;
             }
         }
-        else if(mine != sys && abs(mine-sys) > 0.000001) {
+        else if(mine != sys && ABS(mine-sys) > 0.000001) {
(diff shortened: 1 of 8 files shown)


thiagoftsm (approved):
After to test this PR, I could not detect nothing anormal, thank you @KickerTom !

vlvkobal (approved):

stelfrag (approved):

PR #10381, opened 2020-12-14
Fix issue with chart metadata sent multiple times over ACLK

stelfrag:
##### Summary
Mark a chart as "need to send via ACLK" when a chart is created, has it's direct metadata changed, or a dimension is
added. 

Check if the chart needs to be sent via ACLK when metrics are collected.

##### Component Name
database

##### Test Plan
- Enable cgroups
- Start an agent from the master branch (compile with -DNETDATA_INTERNAL_CHECKS=1)
   - Notice the number of `Sending MQTT len=913 mid=438 wq=0 (0-bytes) readq=0: {	"type": "chart" ... `  msgs (**sample**)
- Apply the PR, compile and restart
   - Notice the reduced number of `Sending MQTT len=913 mid=438 wq=0 (0-bytes) readq=0: {	"type": "chart" ... `  msgs

The type `chart` messages are individual chart updates

### DIFF
--- a/database/rrddim.c
+++ b/database/rrddim.c
@@ -211,7 +211,7 @@ void rrdcalc_link_to_rrddim(RRDDIM *rd, RRDSET *st, RRDHOST *host) {
     }
 #ifdef ENABLE_ACLK
     if (netdata_cloud_setting)
-        aclk_update_chart(st->rrdhost, st->id, ACLK_CMD_CHART);
+        rrdset_flag_set(st, RRDSET_FLAG_ACLK);
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


underhood:
I think you can do this always (no need to check cloud enabled)
`         rrdset_flag_set(st, RRDSET_FLAG_ACLK);`

database/rrddim.c
```diff
@@ -211,7 +211,7 @@ void rrdcalc_link_to_rrddim(RRDDIM *rd, RRDSET *st, RRDHOST *host) {
     }
 #ifdef ENABLE_ACLK
     if (netdata_cloud_setting)
-        aclk_update_chart(st->rrdhost, st->id, ACLK_CMD_CHART);
+        rrdset_flag_set(st, RRDSET_FLAG_ACLK);
```
underhood, line 214: all the places like this can always set the flag (regardless of netdata_cloud_setting). This way we avoid branching

stelfrag: True yes, the flag is checked again in `aclk_update_chart`

underhood:
reduces number of chart messages in my case from 376 to 283 (saves 93 messages). Did multiple tests with same results +- 2 on each number.

We should merge this definitely but I think we should still hold back for second and then send multiple chart updates together.

underhood (approved):

thiagoftsm (approved):
I tested now and I did not have any problems with ACLK, all charts were sent.

Issue #10377, opened 2020-12-11
Persist `$TMPDIR` from environment on install to allow usage in updater.


Ferroin:
As in the title.

This would allow users who install Netdata and set a custom value for `$TMPDIR` to not need to change anything to use the same location for the updater script when using auto-updates.

Should still be overridden if `$TMPDIR` is set at updater runtime, but should take precedence over `/tmp` and `$PWD` when trying to find a location to place temporary files.

therealgilles:
One option would be to add the variable to /etc/netdata/netdata.conf or a separate config file, and have the updater read it.

PR #10384, opened 2020-12-14
Persist `$TMPDIR` from installer to updater.

Ferroin:
##### Summary

This updates the installer to persist the value of the `$TMPDIR` environment variable so that it will be used automatically by the updater.

By default this persisted value is used in preference to any setting for `$TMPDIR` at runtime unless the path is not usable (doesn’t exist, or doesn’t allow both writing and executing files).

##### Component Name

area/packaging

##### Test Plan

Behavior verified locally.

##### Additional Information

Fixes: #10377


knatsakis (approved):

stevenans9859 (approved):

therealgilles:
Does this fix mean we need to define NETDATA_TMPDIR in /etc/netdata/.environment when /tmp has the noexec atttribute, like this:
`NETDATA_TMPDIR=/root/tmp # or some other executable directory`
?

Ferroin:
@therealgilles The first successful update to a version with this fix _should_ set the variable correctly without needing any user intervention unless none of the default locations we check (`$TMPDIR`, `/tmp`, and then the current working directory) meet the requirements for our usage.

Manually setting the variable would work also though.

therealgilles:
Trying to run the cron job manually, I get this:

```
> /etc/cron.daily/netdata-updater 
/root/tmp/netdata-updater-ppQAIwvx9X/netdata-updater.sh: line 354: THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT: No such file or directory
```

therealgilles:
It's now working after re-running: `bash <(curl -Ss https://my-netdata.io/kickstart.sh) --stable-channel --reinstall`
Weird that it did not work after I ran it yesterday, unless there was a fix/update since then.

Ferroin:
The `ENVIRONMENT_FILE` thing is an unrelated bug that will be fixed by #10447.

therealgilles:
Thank you for the response. I would personally call that a related bug due to its effect.

PR #10387, opened 2020-12-14
Use remark-lint to check documentation changes on PRs.

Ferroin:
##### Summary

SSIA

##### Component Name

area/ci

##### Test Plan

n/a


joelhans:
@Ferroin I think the move to remove `package.json` entirely made a lot of checks angry :)

Including the `action-remark-lint` action itself, which I think is unable to install of the modules it needs.

```
BREAKING_CHANGES.md
  1:1  error  Error: Cannot parse file `.remarkrc.js`
Cannot find module 'remark-preset-lint-markdown-style-guide'
```

Maybe we can try restoring `package.json` to see what happens, then more selectively reduce the scope of the listed dependencies? That said, if we still need to declare these dependencies inside the repo, then we've added a ReviewDog check, which is great, but haven't solved the issue of having too many/outdated dependencies.

knatsakis (dismissed):

joelhans:
@Ferroin An update on this, because I saw a conversation on Slack and maybe it's a bigger priority to merge this and get rid of `package-lock.json` than I'd originally thought. I pushed some commits and got remark working, but it seems to run on every `.md` in the repo, not just those changed. I tried playing with `filter_mode:` a bit, but no luck. I'll try again later today, but if you have any ideas, I'm all ears.

Ferroin:
Closing this as it’s horribly out of date.

We will probably have a new PR some time within the next month or so doing the same thing.

Issue #10389, opened 2020-12-15
Install Netdata fails on proxmox


toxic0berliner:
##### Bug report summary
Suffering from a known bug I needed to reinstall (current issue is that cron netdata-updater.sh fails on line 31, I get a daily email about this failure to update). The proposed solution was to update, so I tried using the same method I used to install : the kickstarter.
The script is spawning so many threads that it overloads everything, I was thinking about increasing the limit for number of procs but something is obviously wrong...

##### OS / Environment
Proxmox (debian based)
Linux virgo 5.4.34-1-pve #1 SMP PVE 5.4.34-2 (Thu, 07 May 2020 10:02:02 +0200) x86_64 GNU/Linux
```
/etc/os-release:PRETTY_NAME="Debian GNU/Linux 10 (buster)"
/etc/os-release:NAME="Debian GNU/Linux"
/etc/os-release:VERSION_ID="10"
/etc/os-release:VERSION="10 (buster)"
/etc/os-release:VERSION_CODENAME=buster
/etc/os-release:ID=debian
/etc/os-release:HOME_URL="https://www.debian.org/"
/etc/os-release:SUPPORT_URL="https://www.debian.org/support"
/etc/os-release:BUG_REPORT_URL="https://bugs.debian.org/"
```

##### Netdata version
netdata v1.26.0-333-nightly
 

##### Component Name
Install script

##### Steps To Reproduce
1. Run this 
````
 # bash <(curl -Ss https://my-netdata.io/kickstart.sh) --disable-telemetry --reinstall
 --- Found existing install of Netdata under: / ---
 --- Attempting to update existing install instead of creating a new one ---
[/root]# //usr/libexec/netdata/netdata-updater.sh --not-running-from-cron //usr/libexec/netdata/netdata-updater.sh: fork: retry: Resource temporarily unavailable
//usr/libexec/netdata/netdata-updater.sh: fork: retry: Resource temporarily unavailable
//usr/libexec/netdata/netdata-updater.sh: fork: retry: Resource temporarily unavailable
//usr/libexec/netdata/netdata-updater.sh: fork: retry: Resource temporarily unavailable
//usr/libexec/netdata/netdata-updater.sh: fork: Resource temporarily unavailable
//usr/libexec/netdata/netdata-updater.sh: line 31: /netdata-updater.sh: No such file or directory
 FAILED

 ABORTED  Failed to update existing Netdata install

````

[...]

Issue #10372, opened 2020-12-11
netdata-updater.sh: No such file or directory


gaia:
##### Bug report summary
```/etc/cron.daily/netdata-updater: line 31: /etc/cron.daily/netdata-updater.sh: No such file or directory
run-parts: /etc/cron.daily/netdata-updater exited with return code 127```

##### OS / Environment
```Linux node2 5.4.60-1-pve #1 SMP PVE 5.4.60-2 (Fri, 04 Sep 2020 10:24:50 +0200) x86_64 GNU/Linux
/etc/os-release:PRETTY_NAME="Debian GNU/Linux 10 (buster)"
/etc/os-release:NAME="Debian GNU/Linux"
/etc/os-release:VERSION_ID="10"
/etc/os-release:VERSION="10 (buster)"
/etc/os-release:VERSION_CODENAME=buster
/etc/os-release:ID=debian
/etc/os-release:HOME_URL="https://www.debian.org/"
/etc/os-release:SUPPORT_URL="https://www.debian.org/support"
/etc/os-release:BUG_REPORT_URL="https://bugs.debian.org/"```

##### Netdata version
`netdata v1.26.0-333-nightly`

##### Component Name
Updater (?)

##### Steps To Reproduce
run cron to update daily

##### Expected behavior
no errors, successful update when available

**Comment by ilyam8** [comment]
Hi @gaia, thx for reporint! Unfortunately updater in `netdata v1.26.0-333-nightly` is broken. The problem is fixed in master.

To fix it you need to re-run the installer, see https://github.com/netdata/netdata/issues/10366#issuecomment-742699376

**Comment by gaia** [comment]
thanks, sorry i didn't see it before i posted!

**Comment by gaia** [comment]
still happening

```
/etc/cron.daily/netdata-updater:
/etc/cron.daily/netdata-updater: line 31: /etc/cron.daily/netdata-updater.sh: No such file or directory
run-parts: /etc/cron.daily/netdata-updater exited with return code 127
```

LordBrud:
I fixed it by changing Line 31 in `netdata-updater.sh`.
I changed `script_source="$("$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd -P)/netdata-updater.sh")"`  (https://github.com/netdata/netdata/commit/7a416ecdcba6c1f4dda389731404afa7cc62a3f3#diff-4ecececabbbf75e2b00768f09ba63bef87a6c323c8b1959ad9d79faca7b9f5c9)
[...]

PR #10396, opened 2020-12-15
Fixed option parsing in kickstart.sh.

Ferroin:
##### Summary

The existing code is dependent on options specifically for the kickstart script coming before any options to be passed to the installer, resulting in some options being ignored if the were ordered in certain ways.

This fixes the parsing to explicitly parse _all_ options so that we are not dependent on option order.

##### Component Name

area/packaging

##### Test Plan

1. Create a new install of Netdata in a test environment using the kickstart script with the `--disable-telemetry` option (any option that gets passed to the `netdata-installer.sh` script will work).
2. Try to reinstall in the same environment using the kickstart script by adding `--reinstall` to the end of the command you used in step 1 (note, option order is important here, the `--reinstall` _must_ come after the other options).
3. Without this PR, step 2 will result in the updater script being run (with a message indicating that in the output). With this kickstart script from this PR, it will instead print the message `User requested reinstall instead of update, proceeding.` and not run the updater script.

##### Additional Information

Fixes: #10389 
Relevant to: #10372

### DIFF
--- a/packaging/installer/kickstart.sh
+++ b/packaging/installer/kickstart.sh
@@ -374,7 +374,8 @@ while [ -n "${1}" ]; do
     export NETDATA_LOCAL_TARBALL_OVERRIDE_DEPS_SCRIPT="${1}"
     shift 1
   else
-    break
+    NETDATA_INSTALLER_OPTIONS="$NETDATA_INSTALLER_OPTIONS ${1}"
+    shift 1
   fi
 done
 
@@ -473,7 +474,7 @@ cd netdata-* || fatal "Cannot cd to netdata source tree"
 
 install() {
   progress "Installing netdata..."
-  run ${sudo} ./netdata-installer.sh ${NETDATA_UPDATES} ${NETDATA_INSTALLER_OPTIONS} "${@}" || fatal "netdata-installer.sh exited with error"
+  run ${sudo} ./netdata-installer.sh ${NETDATA_UPDATES} ${NETDATA_INSTALLER_OPTIONS} || fatal "netdata-installer.sh exited with error"
   if [ -d "${ndtmpdir}" ] && [ ! "${ndtmpdir}" = "/" ]; then
     run ${sudo} rm -rf "${ndtmpdir}" > /dev/null 2>&1
   fi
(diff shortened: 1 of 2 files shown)


thiagoftsm (approved):
I executed the same steps described and I also tried to install with other options, everything worked as expected. LGTM!

stelfrag:
I tried this on docker / alpine. The current step (2) to reproduce the message gives me

```
bash-5.0# bash <(curl -Ss https://my-netdata.io/kickstart.sh) --disable-telemetry --reinstall
 --- Found existing install of Netdata under: / --- 
 --- Attempting to update existing install instead of creating a new one --- 
[/]# //usr/libexec/netdata/netdata-updater.sh --not-running-from-cron Checking if a newer version of the updater script is available.
date: invalid date '9999-12-31T23:59:59Z'
//usr/libexec/netdata/netdata-updater.sh: line 168: [: : integer expression expected
Tue Dec 15 17:18:04 UTC 2020 : INFO:  Running on a terminal - (this script also supports running headless from crontab)
Tue Dec 15 17:18:05 UTC 2020 : INFO:  Current Version: 00102600000391
Tue Dec 15 17:18:05 UTC 2020 : INFO:  Latest Version: 00102600000391
Tue Dec 15 17:18:05 UTC 2020 : INFO:  Newest version (current=00102600000391 >= latest=00102600000391) is already installed
 OK  
```

stelfrag (approved):
This works as describe and correctly processes the --reinstall option

There is however an issue with the updater  -- the date check doesn't seem to work on a docker alpine environment (unknown option) BusyBox v1.31.1

Ferroin:
> This works as described and correctly processes the --reinstall option
> 
> There is however an issue with the updater -- the date check doesn't seem to work on a docker alpine environment (unknown option) BusyBox v1.31.1

I’m not 100% certain that it’s actually Alpine specific. We’re running in bash there just like everywhere else, and `[` is a builtin in bash, so it should work everywhere. The exact message is also usually indicative that _something_ is causing one of the two variables being compared on that line to not be integers. It’s probably just something that we have thus far only seen on Alpine.

Issue #10285, opened 2020-11-25
cgroups collection gaps in dynamic environments


cakrit:
##### Bug report summary
When cgroups get dynamically updated (e.g. our k8s infra), the cgroup collector is blocked and we lose data.

##### OS / Environment
Netdata production k8s infra

##### Netdata version
Latest

##### Component Name
cgroups

##### Steps To Reproduce
trigger running of cgroup-name.sh (new cgroup detected) in an infra with many containers 

##### Expected behavior
The bash scripts should be executed in parallel and not block data collection. This means that the cgroups collector should be multi-threaded.

cakrit:
cc @mfundul who helped with the triage

ilyam8:
`cgroup-name.sh` is indeed a problem and very slow/cpu expensive.

We added 2nd http call to get `cluster_id` in [Kubernetes labels PR](https://github.com/netdata/netdata/pull/10107)

i did some optimizations:
 - `/api/v1/pods` call: reduce number of returned data by [adding `fieldSelector` query parameters](https://github.com/netdata/netdata/pull/10107/commits/024d1538a9b7ab97ccf37d151e5cbe6f808192b9), so now it returns list of pods for the node netdata is running.
 - `api/v1/namespaces/kube-system`: it doesn't change, so i decided [to cache the response](https://github.com/netdata/netdata/pull/10107/commits/abf4e8af2c6d659ecacb9e83f1a556db6dd34fe8), so it is only one http call (`curl`) then we read from a file.

We could try to cache `/api/v1/pods` too and do a http call (`curl`) only if a cgroup is not in there. So `cgroup-name.sh` would be reading from a file, this would reduce number of http calls on start significantly. I am not sure this helps, but could try.

**That workaround is not good**, but we should have no problems with caching data since we do name resolution only on container creation.

Ideally we need some cgroup name resolution service or something 🤔

ilyam8:
`cgroups` collector checks for new cgroups every 10 by default.
If there is any it calls `cgroup-name.sh` script _NumberOfNewCgroups_ times.

I didnt check other sources, but for k8s we need only one HTTP call, so it would be much better to pass a list of ids and get a list of names.

Just thinking out loud here 🤔

ilyam8:
Fixed in #10399

PR #10399, opened 2020-12-16
Dispatch cgroup discovery into another thread

vlvkobal:
##### Summary
This is a quick fix for #10285. cgroup discovery will run in a separate thread and won't stop the main thread from collecting metrics and updating charts.

##### Component Name
cgroups plugin

##### Test Plan
1) Start a container that gives you live charts in Netdata
2) Add `sleep 1` to the end of `/usr/libexec/netdata/plugins.d/cgroup-name.sh`
```
fi

sleep 1
info "cgroup '${CGROUP}' is called '${NAME}'"
echo "${NAME}"

exit ${NAME_NOT_FOUND}
```
3) Run with `sudo` the following script
```
#!/bin/bash

CGROUP_PATH="/sys/fs/cgroup/cpu/test"

# number_of_containers=1000 for memory leak test
number_of_containers=10

mkdir "${CGROUP_PATH}"

for i in $(seq 0 9)
do

    echo "create cgroups, step ${i}"
    seq_boundaries="$((i * number_of_containers + 1)) $((i * number_of_containers + number_of_containers))"
    for j in $(seq ${seq_boundaries}); do mkdir "${CGROUP_PATH}/${j}"; done

    echo "sleep..."
    sleep 30

    echo "remove cgroups, step ${i}"
    for j in $(seq ${seq_boundaries}); do rmdir "${CGROUP_PATH}/${j}"; done

    echo "sleep..."
    sleep 10

done

rmdir "${CGROUP_PATH}"

```
4) Check that you don't have gaps for the container you ran in the first step.

### DIFF
--- a/collectors/cgroups.plugin/sys_fs_cgroup.c
+++ b/collectors/cgroups.plugin/sys_fs_cgroup.c
@@ -1704,106 +1704,8 @@ static inline void mark_all_cgroups_as_not_available() {
     }
 }
 
-static inline void cleanup_all_cgroups() {
-    struct cgroup *cg = discovered_cgroup_root, *last = NULL;
-
-    for(; cg ;) {
-        if(!cg->available) {
-            // enable the first duplicate cgroup
-            {
-                struct cgroup *t;
-                for(t = discovered_cgroup_root; t ; t = t->discovered_next) {
-                    if(t != cg && t->available && !t->enabled && t->options & CGROUP_OPTIONS_DISABLED_DUPLICATE && t->hash_chart == cg->hash_chart && !strcmp(t->chart_id, cg->chart_id)) {
-                        debug(D_CGROUP, "Enabling duplicate of cgroup '%s' with id '%s', because the original with id '%s' stopped.", t->chart_id, t->id, cg->id);
-                        t->enabled = 1;
-                        t->options &= ~CGROUP_OPTIONS_DISABLED_DUPLICATE;
-                        break;
-                    }
-                }
-            }
-
-            if(!last)
-                discovered_cgroup_root = cg->discovered_next;
-            else
-                last->discovered_next = cg->discovered_next;
-
-            cgroup_free(cg);
-
-            if(!last)
-                cg = discovered_cgroup_root;
-            else
-                cg = last->discovered_next;
-        }
-        else {
-            last = cg;
-            cg = cg->discovered_next;
-        }
-    }
-}
-
-static inline void copy_discovered_cgroups()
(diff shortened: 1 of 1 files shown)


collectors/cgroups.plugin/sys_fs_cgroup.c
```diff
@@ -1452,24 +1466,27 @@ static inline struct cgroup *cgroup_add(const char *id) {
...
+                //     t->enabled = 0;
+                //     t->options |= CGROUP_OPTIONS_DISABLED_DUPLICATE;
+                // }
+                // else {}
+                //
+                // https://github.com/netdata/netdata/issues/797#issuecomment-241248884
+                error("CGROUP: chart id '%s' already exists with id '%s' and is enabled and available. Disabling cgroup with id '%s'.",
+                        cg->chart_id, t->id, cg->id);
+                debug(D_CGROUP, "Control group with chart id '%s' already exists with id '%s' and is enabled and available. Disabling cgroup with id '%s'.",
+                        cg->chart_id, t->id, cg->id);
```
ilyam8, line 1487: seems like same message, why do we need that debug?

collectors/cgroups.plugin/sys_fs_cgroup.c
```diff
@@ -2046,10 +1976,137 @@ static inline void find_all_cgroups() {
...
+    uv_mutex_unlock(&cgroup_root_mutex);
 
     debug(D_CGROUP, "done searching for cgroups");
 }
 
+void cgroup_discovery_worker(void *ptr)
+{
+    UNUSED(ptr);
+
+    while (!netdata_exit) {
```
thiagoftsm, line 2094: @vlvkobal while we were working with `ebpf.plugin` we had a talk about to use or not `netdata_exit`, and we arrived in the conclusion we should not use it. Are you sure that this won't create an infinite loop?

vlvkobal: We still use `netdata_exit`. I don't see any alternatives.

collectors/cgroups.plugin/sys_fs_cgroup.c
vlvkobal: Removed.

ilyam8:
I **tested only functionality** and it works good, name resolution no longer affects data collection, good work @vlvkobal

ilyam8 (approved):

thiagoftsm (approved):
I tested now and everything looks OK.

PR #10400, opened 2020-12-17
python.d/fail2ban: Add handling "yes" and "no" as bool, match flexible spaces

grinapo:
##### Summary
1. `enable = yes` is handled by fail2ban, so we shall, too
2. `enable ________          = ________              true` (_ = space, somehow github eats multiple spaces!) is valid, however stupid it looks

##### Component Name
python.d/fail2ban.chart.py

##### Test Plan
The change is trivial, unless there's a syntax error it shall work as intended.

##### Additional Information
Curiously the 'yes' / 'no' isn't mentioned in the new fail2ban doc but my old files contained such settings and handled.

### DIFF
--- a/collectors/python.d.plugin/fail2ban/fail2ban.chart.py
+++ b/collectors/python.d.plugin/fail2ban/fail2ban.chart.py
@@ -50,7 +50,7 @@ def charts(jails):
     return ch
 
 
-RE_JAILS = re.compile(r'\[([a-zA-Z0-9_-]+)\][^\[\]]+?enabled\s+= (true|false)')
+RE_JAILS = re.compile(r'\[([a-zA-Z0-9_-]+)\][^\[\]]+?enabled\s+= +(true|yes|false|no)')
 
 # Example:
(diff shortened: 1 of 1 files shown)


ilyam8 (approved):

ilyam8:
@grinapo thx for contributing! lgtm 👍

vlvkobal (approved):

ilyam8:
@grinapo 

> Curiously the 'yes' / 'no' isn't mentioned in the new fail2ban doc but my old files contained such settings and handled.


[code](https://github.com/fail2ban/fail2ban/blob/960e30cfcdae7e2c81f74046efdbff1606f9774e/fail2ban/helpers.py#L120-L122):

```python
def _as_bool(val):
	return bool(val) if not isinstance(val, basestring) \
		else val.lower() in ('1', 'on', 'true', 'yes')
```

thiagoftsm (approved):

alivx (approved):

Issue #10420, opened 2020-12-23
/etc/cron.daily/netdata-updater: line 168: [: : integer expression expected


skibbipl:
##### Bug report summary
There is some problem with netdata updater. I have 2 raspberry pis and on both I'm getting the same error.
```
 # /etc/cron.daily/netdata-updater
Checking if a newer version of the updater script is available.
date: invalid date ‘9999-12-31T23:59:59Z’
/etc/cron.daily/netdata-updater: line 168: [: : integer expression expected
Wed Dec 23 15:36:43 CET 2020 : INFO:  Running on a terminal - (this script also supports running headless from crontab)
Wed Dec 23 15:36:44 CET 2020 : INFO:  Current Version: 00102800000010
Wed Dec 23 15:36:44 CET 2020 : INFO:  Latest Version: 00102800000010
Wed Dec 23 15:36:44 CET 2020 : INFO:  Newest version (current=00102800000010 >= latest=00102800000010) is already installed
```
##### OS / Environment
```
Linux rpi4b 5.4.79-v8+ #1373 SMP PREEMPT Mon Nov 23 13:32:41 GMT 2020 aarch64 GNU/Linux
/etc/os-release:PRETTY_NAME="Raspbian GNU/Linux bullseye/sid"
/etc/os-release:NAME="Raspbian GNU/Linux"
/etc/os-release:ID=raspbian
/etc/os-release:ID_LIKE=debian
/etc/os-release:HOME_URL="http://www.raspbian.org/"
/etc/os-release:SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
/etc/os-release:BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

##### Netdata version
netdata v1.28.0-10-nightly

##### Component Name
netdata-updater

##### Steps To Reproduce
run `/etc/cron.daily/netdata-updater`

##### Expected behavior
No errors

Lamera:
Same issue #10417

Lamera:
Duplicate of #10417

underhood:
@Lamera is correct this is duplicate.
@skibbipl closing this as duplicate of #10417 to track the issue in single place. Thank you reporting and contributing PR with possible fix.

Issue #10417, opened 2020-12-22
/etc/cron.daily/netdata-updater fails


bmatheny:
Found https://community.netdata.cloud/t/periodically-anacron-job-cron-daily-on-localhost-reports-a-problem-with-netdata-updater-sh/612 but this workaround does not fix my issue.

##### Bug report summary
cron has been failing to run netdata-updater (`/etc/cron.daily/netdata-updater`) for a while now:

```
root@admin00:~# /etc/cron.daily/netdata-updater
Checking if a newer version of the updater script is available.
Downloading newest version of updater script.
/tmp/netdata-updater-6YqW7XDJWd/netdata-updater.sh: line 348: THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT: No such file or directory
```

I found the above thread and tried to `--reinstall` but that didn't fix the issue. After running `--reinstall` and it not fixing the issue, I ran the uninstaller, deleted any netdata related files, and did a full reinstall via: `bash <(curl -Ss https://my-netdata.io/kickstart.sh) --stable-channel`

After doing that, I still get the same error when I run `netdata-updater` via cron (or via the CLI using `/etc/cron.daily/netdata-updater`).

It looks to me like the kickstart (or some other) installer does some sed magic on netdata-updater when dropping it in to `/etc/cron.daily/`, but that same magic isn't applied by `netdata-updater` when attempting to self update.

Possible fixes:

1. Set default for netdata-updater in cron to not self-update (`--no-updater-self-update` works)
2. Apply same logic from kickstart installer for `/etc/cron.daily/netdata-updater` on `/tmp/netdata-updater-XXXXXXXXXX/netdata-updater.sh` when `/etc/cron.daily/netdata-updater` attempts to self update.

##### OS / Environment

```
/etc/lsb-release:DISTRIB_ID=Ubuntu
/etc/lsb-release:DISTRIB_RELEASE=20.04
/etc/lsb-release:DISTRIB_CODENAME=focal
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="20.04.1 LTS (Focal Fossa)"
/etc/os-release:ID=ubuntu
/etc/os-release:ID_LIKE=debian
/etc/os-release:PRETTY_NAME="Ubuntu 20.04.1 LTS"
/etc/os-release:VERSION_ID="20.04"
```

[...]

PR #10424, opened 2020-12-26
Qick and dirty fix for #10420

skibbipl:
##### Summary
At least on my raspbian i do not have python but python 3. So Additional check is needed. Also this value `commit_date="9999-12-31T23:59:59Z"` is not valid for `date -d $commit_date`

##### Component Name
Updater

##### Test Plan
Ensure you have only python3 and not python available.


underhood:
@Ferroin seems to fix #10420 and #10417

underhood:
@Ferroin I gues also installing packages `python-is-python2` or `python-is-python3` works

Ferroin (approved):

PR #10425, opened 2020-12-26
Use bash shell as user netdata for debug

Steve8291:
##### Summary
Looked like this was just a typo but the command won't work as it was written.
##### Component Name
Docs for ebpf.plugin

##### Test Plan

Tested command in bash.

### DIFF
--- a/collectors/ebpf.plugin/README.md
+++ b/collectors/ebpf.plugin/README.md
@@ -252,7 +252,7 @@ output.
 
 ```bash
 cd /usr/libexec/netdata/plugins.d/
-sudo -s /bin/bash netdata
+sudo su -s /bin/bash netdata
 ./ebpf.plugin
 ```
 
--- a/collectors/ebpf.plugin/README.md
+++ b/collectors/ebpf.plugin/README.md
@@ -252,7 +252,7 @@ output.
 
 ```bash
 cd /usr/libexec/netdata/plugins.d/
-sudo -u netdata bash
+sudo -s /bin/bash netdata
 ./ebpf.plugin
 ```
 


joelhans:
On my system (Fedora 33), the existing command works fine, but `sudo -s /bin/bash netdata` gives me an error: `/usr/sbin/netdata: /usr/sbin/netdata: cannot execute binary file`. Not to say my system is the source of truth, but rather that it's probably important to investigate and implement what is, truly, the most universal way of switching to the `netdata` user.

In the Go collector docs, we even have an entirely different method: `sudo -u netdata -s`. @Steve8291 Could you test that and let me know what works best on your system?

I'd also like to hear from @ilyam8 and @Ferroin, who are most likely to know some best practices for this, rather than my speculation. If we can come to a consensus, it would be nice to push that here, and then I can also comb through the docs to make all the other references consistent.

ilyam8:
> sudo -u netdata bash

is correct, we invoke `bash` command as `netdata` user.

```bash
     -u user, --user=user
                 Run the command as a user other than the default target user (usually root).
```

---

> Looked like this was just a typo but the command won't work as it was written.

@Steve8291 could you show the error?

ilyam8:
> the most universal way of switching to the netdata user.

My personal expirience, i always use `su` 🤷‍♂️ 

```cmd
NAME
       su - change user ID or become superuser
SYNOPSIS
       su [options] [username
```

> sudo su -s /bin/bash netdata

---

On the other hand we want just to execute a command as `netdata` user, so `sudo` is enough

```cmd
NAME
     sudo, sudoedit — execute a command as another user

SYNOPSIS
     sudo -l [-AknS] [-a type] [-g group] [-h host] [-p prompt] [-U user] [-u user] [command]
```

> sudo -u netdata ./go.d.plugin -d -m example

Steve8291:
@joelhans @ilyam8 
Sorry guys I think I messed up here. I meant to write the pull request as @ilyam8 said above. That's the way I always do this as well: `sudo su -s /bin/bash netdata`
As @joelhans said, the code in the documentation is different in other collectors like Go. I was thinking it would be good to standardize it.

On Ubuntu: 
If I run `sudo -u netdata bash`
I get `bash: /home/user/.bashrc: Permission denied`

If I run `sudo -u netdata -s`
I get `bash: /home/user/.bashrc: Permission denied`

If I run `sudo -s /bin/bash netdata`
I get `/usr/sbin/netdata: /usr/sbin/netdata: cannot execute binary file`

If I run `sudo su -s /bin/bash netdata`
The command works as it should.

Steve8291:
Would it be cleaner to just have the whole thing as one command in the docs without changing dirs and switching users?
`sudo -u netdata /usr/libexec/netdata/plugins.d/ebpf.plugin`
As well as an example for debugging modules for plugins that have them:
`sudo -u netdata /usr/libexec/netdata/plugins.d/go.d.plugin -d -m <module>`

joelhans:
Seems like `sudo su -s /bin/bash netdata` is the best way. @Steve8291 can you make the change here?

> Would it be cleaner to just have the whole thing as one command in the docs without changing dirs and switching users?

I like the cleanliness of that, but not every installation will have those files at `/usr/libexec/netdata`. As you can see from an [example Go collector's](https://learn.netdata.cloud/docs/agent/collectors/go.d.plugin/modules/apache#troubleshooting) troubleshooting section:

> First, navigate to your plugins directory, usually at /usr/libexec/netdata/plugins.d/. If that's not the case on your system, open netdata.conf and look for the setting plugins directory

But, we could clarify that usage as well, asking users to replace the path in the command if theirs is different. I'll have to think about it some more, but for now, I'm going to do some investigation around how we talk about switching users across the docs and come up with a plan.

joelhans:
No need to worry about the failing **Docs** check. It has nothing to do with this PR, and the link works fine despite the failure.

Steve8291:
Sounds good @joelhans 
I've made the change.

ilyam8 (approved):

joelhans (approved):

PR #10441, opened 2021-01-03
Exclude autofs by default in diskspace plugin

nabijaczleweli:
##### Summary

autofs's point is to be mounted on explicit request, and "automatically by netdata" is not one.

##### Component Name

diskspace.plugin i guess

##### Test Plan

Add this to your existing netdata installation and see that autofs mounts aren't always instamounted.


vlvkobal:
I don't think `autofs` mount points should be excluded by default. Should they? @Ferroin, what do you think?

vlvkobal (approved):

vlvkobal:
Thank you @nabijaczleweli for your contribution!

Issue #10444, opened 2021-01-04
Updater doesn't message the installer correctly


nabijaczleweli:
##### Bug report summary
I've been getting root mail to this effect since December 17th:
```
● cron-daily.service - systemd-cron daily script service
     Loaded: loaded (/lib/systemd/system/cron-daily.service; static)
     Active: failed (Result: exit-code) since Mon 2021-01-04 07:17:45 CET; 73ms ago
       Docs: man:systemd.cron(7)
    Process: 17839 ExecStartPre=/lib/systemd-cron/boot_delay 5 (code=exited, status=0/SUCCESS)
    Process: 17840 ExecStart=/bin/run-parts --report /etc/cron.daily (code=exited, status=1/FAILURE)
   Main PID: 17840 (code=exited, status=1/FAILURE)
        CPU: 294ms

Jan 04 06:25:07 szarotka systemd[1]: Starting systemd-cron daily script service...
Jan 04 07:17:45 szarotka run-parts[17840]: /etc/cron.daily/netdata-updater:
Jan 04 07:17:45 szarotka run-parts[17840]: /tmp/netdata-updater-FmurKT367W/netdata-updater.sh: line 348: THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT: No such file or directory
Jan 04 07:17:45 szarotka run-parts[17840]: run-parts: /etc/cron.daily/netdata-updater exited with return code 1
Jan 04 07:17:45 szarotka systemd[1]: cron-daily.service: Main process exited, code=exited, status=1/FAILURE
Jan 04 07:17:45 szarotka systemd[1]: cron-daily.service: Failed with result 'exit-code'.
Jan 04 07:17:45 szarotka systemd[1]: Failed to start systemd-cron daily script service.
Jan 04 07:17:45 szarotka systemd[1]: cron-daily.service: Triggering OnFailure= dependencies.
```

Using kickstart manually does work, but this reappears the following run.

##### OS / Environment
```
Linux szarotka 5.9.0-5-amd64 #1 SMP Debian 5.9.15-1 (2020-12-17) x86_64 GNU/Linux
/etc/os-release:PRETTY_NAME="Debian GNU/Linux bullseye/sid"
/etc/os-release:NAME="Debian GNU/Linux"
/etc/os-release:ID=debian
/etc/os-release:HOME_URL="https://www.debian.org/"
/etc/os-release:SUPPORT_URL="https://www.debian.org/support"
[...]

Issue #10419, opened 2020-12-23
Autoupdate issue on Ubuntu 20.04


animosity22:
##### Bug report summary
After a fresh install and the update tries to run, it generates the error below.

##### OS / Environment
```
root@gemini:~# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.1 LTS
Release:        20.04
Codename:       focal
root@gemini:~# uname -a
Linux gemini 5.4.0-58-generic #64-Ubuntu SMP Wed Dec 9 08:16:25 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

##### Netdata version
```
root@gemini:~# /opt/netdata/bin/srv/netdata -V
netdata v1.28.0-13-gb170f860
```
 

##### Component Name
/etc/cron.daily/netdata-updater:

##### Steps To Reproduce
Let the cron job run and it produces:

```
/tmp/netdata-updater-4kjwwOmVpf/netdata-updater.sh: line 348: THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT: No such file or directory
run-parts: /etc/cron.daily/netdata-updater exited with return code 1
```

##### Expected behavior
Update works.

ChrisG661:
I also ran into this issue, except it fails on line 354.

PR #10447, opened 2021-01-04
Fixed handling of environment file in updater script.

Ferroin:
##### Summary

Currently, when the script auto-updates, it is unable to find the environment file in many cases due to a coding oversight, causing auto-updates to fail due to the automatic updating of the updater script.

This updates the script to pass the environment file path to the updated copy, as well as more robustly looking for the file itself instead of just failing if it was not substituted in during install, which both fixes the script going forwards and allows a significant majority of existing installs to update properly.

##### Component Name

area/packaging

##### Test Plan

The bug can be verified in existing installs by simply trying to update.

The fix can be verified by replacing the updater in a broken existing install with the copy from this PR, and then trying to update again.

##### Additional Information

Fixes: #10372 
Fixes: #10419 
Fixes: #10444

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -44,6 +44,26 @@ error() {
   echo >&3 "$(date) : ERROR: " "${@}"
 }
 
+: "${ENVIRONMENT_FILE:=THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT}"
+
+if [ "${ENVIRONMENT_FILE}" == "THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT" ]; then
+  if [ -r "${script_dir}/../../../etc/netdata/.environment" ]; then
+    ENVIRONMENT_FILE="${script_dir}/../../../etc/netdata/.environment"
+  elif [ -r "/etc/netdata/.environment" ]; then
+    ENVIRONMENT_FILE="/etc/netdata/.environment"
+  elif [ -r "/opt/netdata/etc/netdata/.environment" ]; then
+    ENVIRONMENT_FILE="/opt/netdata/etc/netdata/.environment"
+  else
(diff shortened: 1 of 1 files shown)


cepn-itadmin:
Any chance to have this useful PR merged soon? I have this bug every time the updater is self updated from repo.

packaging/installer/netdata-updater.sh
```diff
@@ -44,6 +44,26 @@ error() {
...
+
+if [ "${ENVIRONMENT_FILE}" == "THIS_SHOULD_BE_REPLACED_BY_INSTALLER_SCRIPT" ]; then
+  if [ -r "${script_dir}/../../../etc/netdata/.environment" ]; then
+    ENVIRONMENT_FILE="${script_dir}/../../../etc/netdata/.environment"
+  elif [ -r "/etc/netdata/.environment" ]; then
+    ENVIRONMENT_FILE="/etc/netdata/.environment"
+  elif [ -r "/opt/netdata/etc/netdata/.environment" ]; then
+    ENVIRONMENT_FILE="/opt/netdata/etc/netdata/.environment"
+  else
+    envpath="$(find / -type d \( -path /sys -o -path /proc -o -path /dev \) -prune -false -o -path '*netdata/.environment' -type f  2> /dev/null | head -n 1)"
```
vkalintiris, line 57: We should verify this works fine on MacOS and FreeBSD (ie. if we haven't already done so).

vkalintiris (approved):

packaging/installer/netdata-updater.sh
Ferroin: It _should_ work both places as the whole line is POSIX compliant in both syntax and options for the tools, but I will make a point to double-check it prior to merging.

Ferroin: Confirmed, it works on both macOS and FreeBSD.

Issue #10340, opened 2020-12-08
Tarball checksum validation failed


diegomontoya:
Ubuntu 20.04
X86-64

```
[/tmp/netdata-kickstart-PfyFGgU46M]# curl -q -sSL --connect-timeout 10 --retry 3 --output /tmp/netdata-kickstart-PfyFGgU46M/sha256sum.txt https://storage.googleapis.com/netdata-nightlies/sha256sums.txt  OK  

[/tmp/netdata-kickstart-PfyFGgU46M]# curl -q -sSL --connect-timeout 10 --retry 3 --output /tmp/netdata-kickstart-PfyFGgU46M/netdata-latest.tar.gz https://storage.googleapis.com/netdata-nightlies/netdata-latest.tar.gz  OK  

 ABORTED  Tarball checksum validation failed. Stopping netdata installation and leaving tarball in /tmp/netdata-kickstart-PfyFGgU46M 
```

Exactly same as https://github.com/netdata/netdata/issues/8967

OdysLam:
Hey @diegomontoya ,

Thanks for posting this issue. I think this is related to a problem we have identified with our updater. Please be patient and one of our engineers will come back with more information. Perhaps @Ferroin can provide more context?

lamka02sk:
This should be a priority, the installer is broken completely 😞

Ferroin:
Has anybody actually retried this? It’s a known issue if you try to update right around the same time that our nightly releases get published due to Google Cloud Storage not preemptively invalidating their edge caches when the build assets are updated (and we have no way to work around this), resulting in the files not matching on occasion when fetched around about 00:30-01:30 UTC.

If you are _still_ seeing issues despite testing outside of that time range, then I will be happy to look into them if you can provide more details about your specific setup, but I have been unable to reproduce any such issues locally on any of my test systems.

OdysLam:
Since this is a recurring theme, perhaps @manos-saratsis @dim08 you are interested in tackling this from a product perspective. Perhaps a different error message?

lamka02sk:
@Ferroin Today, I tried running the installer again and it worked. It looks like I hit the mentioned invalid cache window since I commented right around that time.

manos-saratsis:
@OdysLam I think the message is fine. An idea for improvement would be to ask the user to redo the installation @Ferroin what's your thoughts on this.

Ferroin:
[...]

PR #10448, opened 2021-01-04
Updated messages about checksum validation failures on install.

Ferroin:
##### Summary

They now explain the most common cause of the failures and suggest the most direct way to fix the situation.

##### Component Name

area/packaging

##### Test Plan

n/a

##### Additional Information

Fixes: #10340

### DIFF
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -350,7 +350,7 @@ else
 fi
 
 if ! grep netdata-latest.gz.run "${TMPDIR}/sha256sum.txt" | safe_sha256sum -c - > /dev/null 2>&1; then
-  fatal "Static binary checksum validation failed. Stopping netdata installation and leaving binary in ${TMPDIR}"
+  fatal "Static binary checksum validation failed. Stopping netdata installation and leaving binary in ${TMPDIR}\nUsually this is a result of an older copy of the file being cached somehere and can be resolved by simply retrying in an hour."
 fi
 
 # ----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


packaging/installer/kickstart-static64.sh
```diff
@@ -350,7 +350,7 @@ else
 fi
 
 if ! grep netdata-latest.gz.run "${TMPDIR}/sha256sum.txt" | safe_sha256sum -c - > /dev/null 2>&1; then
-  fatal "Static binary checksum validation failed. Stopping netdata installation and leaving binary in ${TMPDIR}"
+  fatal "Static binary checksum validation failed. Stopping netdata installation and leaving binary in ${TMPDIR}\nUsually this is a result of an older copy of the file being cached somehere and can be resolved by simply retrying in an hour."
```
joelhans, line 353: Just a few suggestions to copy-paste to the other blocks, too. Is there a more precise word than "somewhere" we could use? Personally, I can't tell whether that implies cached on my system or the platform where we host the kickstart files.

```suggestion
  fatal "Static binary checksum validation failed. Stopping Netdata Agent installation and leaving binary in ${TMPDIR}\nUsually this is a result of an older copy of the file being cached somewhere and can be resolved by retrying in an hour."
```

Ferroin: ‘somewhere upstream’ maybe? The problem is that we can’t be certain where. It’s most likely to be stale data on a GCS edge node, but if the user is connecting through a proxy it may be cached there instead

joelhans: "Somewhere upstream" is helpful, thanks.

joelhans (approved):

manos-saratsis (approved):

PR #10455, opened 2021-01-06
Make diskspace plugin multithreaded

### LINKED ISSUE #10150: NetData Disk Space Plots for all disks are missing values because of problems with a mounted CIFS
##### Bug report summary

NetData Disk Space Plugin is not showing values for any disks if CIFS share has a problem. We are not sure what the problem with the CIFS is but each time this mounted drive has problems we see that plots of other disks are affected and show missing values ( " - ") 

/mnt/SHARE is a cifs vers=3.0
/mnt/TMP is an internal SSD which is completely fine. The other internal disks are also fine and show this pattern of missing values. 

![image](https://user-images.githubusercontent.com/2061823/97413128-b27b8300-1902-11eb-86fb-4f6eb05863b5.png)

![image](https://user-images.githubusercontent.com/2061823/97413264-dccd4080-1902-11eb-94d7-1b54c78e37ae.png)


##### OS / Environment

```
Linux XXXXXXX 4.15.0-91-generic #92-Ubuntu SMP Fri Feb 28 11:09:48 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 18.04.4 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="18.04.4 LTS (Bionic Beaver)"
/etc/os-release:ID=ubuntu
/etc/os-release:PRETTY_NAME="Ubuntu 18.04.4 LTS"
```

##### Netdata version

```
Netdata version: v1.26.0-77-g5f0f0827
```

##### Component Name
netdata.plugin_diskspace

All instances of the following plots are affected and missing data:
* Disk Files (inodes) Usage for XXX
* Disk Space Usage for XXX
* NetData Disk Space Plugin CPU usage (netdata.plugin_diskspace)
* NetData Disk Space Plugin Duration (netdata.plugin_diskspace_dt)

##### Steps To Reproduce

not sure, we are trying to figure out what the problem with the CIFS Share is.

Not monitoring /mnt/SHARE by setting the following fixes the problem for the other disks

```
[plugin:proc:diskspace]                                                                                                                                                                                                                                       
  exclude space metrics on paths = /nmt/SHARE /proc/* /sys/* /var/run/user/* /run/user/* /snap/* /var/lib/docker/*     
```                                        

##### Expected behavior

Plots for the other disks should not be affected by /mnt/SHARE

vlvkobal:
##### Summary
Diskspace plugin uses `stat()` and `statvfs()` system calls to get information about used space and inodes for mounted filesystems. These calls are blocking and can cause interruptions in data collecting, especially if network mounts are used.

A libuv thread pool is used to dispatch data collecting to a separate thread for every mount point.

Fixes #10150

##### Component Name
diskspace plugin

##### Test Plan
1. To test the plugin with delays, insert `if (!strcmp(mi->mount_point, "/")) sleep(5);` before `stat()` and `statvfs()` calls in `plugin_diskspace.c` and `proc_self_mountinfo.c`.

2. Dynamic mount point creation and deletion:
```
sudo mkdir /mnt/ramdisk
for i in $(seq 1 10); do sudo mount -t tmpfs -o size=1g tmpfs /mnt/ramdisk; sleep 30; sudo umount /mnt/ramdisk; sleep 30; done
```

### DIFF
--- a/collectors/diskspace.plugin/plugin_diskspace.c
+++ b/collectors/diskspace.plugin/plugin_diskspace.c
@@ -455,6 +455,8 @@ struct work_data {
 
 void disk_space_stats_work(uv_work_t* req)
 {
+    // TODO: syncronize the thread with the main loop
+    
     struct work_data *d = req->data;
 
     do_disk_space_stats(d->mi, d->update_every);
@@ -489,6 +491,8 @@ void *diskspace_main(void *ptr) {
     fatal_assert(0 == uv_rwlock_init(&disk_mountinfo_lock));
     fatal_assert(0 == uv_rwlock_init(&dict_mountpoints_lock));
 
+
+    // TODO: create a thread pool for fast mountpoints
     int error = uv_thread_create(&loop_thread.thread, run_event_loop, NULL);
     if (error) {
         error("uv_thread_create(): %s", uv_strerror(error));
@@ -523,15 +527,35 @@ void *diskspace_main(void *ptr) {
 
             mi->busy = 1;
 
+            // TODO:
+            // if a mount point is fast - wait for an available thread - uv_sem_wait()
+            // if a mount point is slow
+            //   check for an available thread - uv_thread_trywait()
+            //   if no threads are available
+            //     create a new thread if possible - uv_sem_post(),uv_thread_create()
+            //     or add to the slow queue (increase the queue counter)
+            // pick a free thread - go throug the list and check for the free flag
+            
+            // prepare data to pass - assign a pointer to the mi data
             struct work_data *d = mallocz(sizeof(struct work_data));
             d->mi = mi;
             d->update_every = update_every;
             mi->work.data = d;
-
+            
+            // wake the thread - uv_cond_signal()
             fatal_assert(0 == uv_queue_work(&loop_thread.loop, &mi->work, disk_space_stats_work, disk_space_stats_done));
+            // if the thread is not busy - repeat the signal
+
+            // :TODO
             
             if(unlikely(netdata_exit)) break;
         }
+
+        // TODO:
+        // go through the slow queue
+        //   check for an available thread with small intervals 
+        //   dispatch jobs (decrease the queue counter)
+        //   end the loop if too few time left before the next heartbeat
 
         if(unlikely(netdata_exit)) break;
 
@@ -547,6 +571,10 @@ void *diskspace_main(void *ptr) {
 
             // ----------------------------------------------------------------
 
+            // TODO: more useful stats
+            // -time until all jobs are dispatched
+            // -total number of slow moutpoints
+            // -number of skipped mountpoints (the slow queue counter)
             getrusage(RUSAGE_THREAD, &thread);
 
             if(unlikely(!stcpu_thread)) {
--- a/collectors/diskspace.plugin/plugin_diskspace.c
+++ b/collectors/diskspace.plugin/plugin_diskspace.c
@@ -411,6 +411,14 @@ static void diskspace_main_cleanup(void *ptr) {
 
     info("cleaning up...");
 
+    /* stop event loop */
+    fatal_assert(0 == uv_async_send(&loop_thread.async));
+
+    int error = uv_thread_join(&loop_thread.thread);
+    if (error) {
+        error("uv_thread_join(): %s", uv_strerror(error));
+    }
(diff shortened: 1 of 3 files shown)


Ferroin:
This should probably be optional (though defaulting to this new behavior makes sense). On a system with a lot of volumes sharing the same persistent storage combined with slow persistent storage, this will result in a much higher impact on the rest of the system and in theory may actually slow down data collection.

mfundul:
We need to discuss this further since there is an inherent problem here. By looking [here](http://docs.libuv.org/en/v1.x/threadpool.html) we can see that "The threadpool is global and shared across all event loops.".

Increasing the maximum size of the threadpool can be done by changing the environment variable `UV_THREADPOOL_SIZE`, which again is not obvious how to do. We probably cannot do this from withing the C code early enough, and it could be necessary to be done from a script spawning the agent, which we don't always have.

Also, increasing the maximum size of the threadpool may not be sufficient if netdata is a very busy master server.

There are 2 cases:
1. If you have a lot of problematic mountpoints you starve the threadpool and no I/O can take place in the agent anymore.
2. If you have a lot of I/O (busy server) then you starve the collector even if there are no problematic mountpoints whatsoever.

The ideal solution would be to use real libuv threads instead of the workqueue API. However, spawning threads every second is not ideal, and implementing a custom thread pool is a lot of work.

mfundul:
So apparently, for now at least, the `UV_THREADPOOL_SIZE` environment variable is gotten in `init_threads()`:
https://github.com/libuv/libuv/blob/dec0723864c2a1d41cfbab8164c5683f5cffff14/src/threadpool.c#L194

which is eventually called by `uv__work_submit()` here:
https://github.com/libuv/libuv/blob/dec0723864c2a1d41cfbab8164c5683f5cffff14/src/threadpool.c#L261

So this probably means that the thread pool initialization is lazy and takes place when `uv_queue_work()` is called for the first time. This could mean that it's possible to increase the maximum threadpool size from within the C code.

mfundul:
Increasing the maximum threadpool size still doesn't resolve case 2 though.

Ferroin:
@mfundul Perhaps it might be best then to split the diskspace plugin out as a separate executable? That way it can just take the easy route for parallelization without also potentially blocking other IO in the main agent process.

ilyam8:
Another relatively easy workaround - skip slow/unresponsive mountpoints (measure `stat()`/`statfs()` execution time). Probably could be implemented as a temporary solution until we find a better way of fixing the issue.

mfundul:
> 
> 
> @mfundul Perhaps it might be best then to split the diskspace plugin out as a separate executable? That way it can just take the easy route for parallelization without also potentially blocking other IO in the main agent process.

It is an internal collector currently. For it to become a plugin, would mean a lot of work, at the cost of a lot of extra overhead. The internal collectors are very lean and it looks like a heavy compromise to me.

vlvkobal:
This is how the fix is implemented at the moment (using a libuv thread pool)
diskspace plugin-Multithreaded

> The ideal solution would be to use real libuv threads instead of the workqueue API. However, spawning threads every second is not ideal, and implementing a custom thread pool is a lot of work.

So, we need to implement our own thread pools
diskspace plugin-Thread pool

mfundul:
How do we identify a slow thread, and how do we guarantee that we never block before identifying that a thread is going to be slow?

vlvkobal:
> How do we identify a slow thread

We'll calculate the time spent while processing a mount point. For the first approach, it's enough to compare it to an empirically determined interval. Later we can make the interval depend on how busy is the fast thread pool.

> how do we guarantee that we never block before identifying that a thread is going to be slow

We do not guarantee that. It's not a big problem if in the first run there will be gaps when fast threads are busy with slow mount points. We can solve the problem by dispatching a separate thread for every new mount point, but I don't think it's worth implementing.

mfundul:
What if a mountpoint suddenly decides to time-out for 60 seconds. Doesn't this mean that the agent will have 60 second gaps for all internal collectors?

vlvkobal:
We have gaps only for the diskspace collector, not for all collectors. If happens something like you described, nothing will happen to the plugin - we still have 3 threads available 😉.

Issue #10456, opened 2021-01-06
Anomalies collector custom models update() unhandled exception: Found array with 0 feature(s) ... while a minimum of 1 is required.


andrewm4894:
##### OS / Environment

```
/etc/lsb-release:DISTRIB_ID=Ubuntu
/etc/lsb-release:DISTRIB_RELEASE=18.04
/etc/lsb-release:DISTRIB_CODENAME=bionic
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 18.04.4 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="18.04.4 LTS (Bionic Beaver)"
/etc/os-release:ID=ubuntu
/etc/os-release:ID_LIKE=debian
/etc/os-release:PRETTY_NAME="Ubuntu 18.04.4 LTS"
/etc/os-release:VERSION_ID="18.04"
/etc/os-release:HOME_URL="https://www.ubuntu.com/"
/etc/os-release:SUPPORT_URL="https://help.ubuntu.com/"
/etc/os-release:BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
/etc/os-release:PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
/etc/os-release:VERSION_CODENAME=bionic
/etc/os-release:UBUNTU_CODENAME=bionic
```

##### Netdata version

netdata v1.26.0-325-nightly
 

##### Component Name

collectors/python.d.plugin/anomalies

##### Steps To Reproduce

1. Uncomment out custom model example in `anomalies.conf` lines 106, 109, 110.
2. Restart Netdata agent.

##### Expected behavior

We should after a little while see an additional anomaly score dimension on our anomalies charts called "apps_python_d_plugin"

Instead we see errors like below begin to appear in the log.

```
2020-12-29 15:34:18: python.d DEBUG: anomalies[local] : update => [FAILED] (elapsed time: -, failed retries in a row: 5)
2020-12-29 15:34:22: python.d INFO: anomalies[local] : Found array with 0 feature(s) (shape=(14397, 0)) while a minimum of 1 is required.
2020-12-29 15:34:22: python.d INFO: anomalies[local] : training failed for user_netdata at run_counter 6, defaulting to hbos model.
[...]

PR #10459, opened 2021-01-06
Anomalies collector custom model bugfix for issue #10456

andrewm4894:
##### Summary

Adds small fix to anomalies collector for custom models.

- Add "|" as delimiter in custom_models_dims_renamed as opposed to ".". The "|" is how the collector knows what cols to pull from the dataframe for each model. Expected colnames in the df are "model_name|dim...." so it looks for "model_name" for example.
- Remove `[self.expected_cols]` check on df_allmetrics. This will avoid failing if you happen to use a dim in a custom model that belongs to a chart that has one or more `null` values from /api/v1/allmetrics. If you happen to use such a dim in your model it will still fail as we can't make a prediction when one of the features is `null`.

##### Component Name

collectors/python.d.plugin/anomalies

##### Test Plan

Testing by creating different `anomalies.conf` that do and do not use custom models. 

##### Additional Information

See related issue for additional background.


collectors/python.d.plugin/anomalies/anomalies.chart.py
```diff
@@ -287,7 +287,7 @@ def predict(self):
         df_allmetrics = get_allmetrics_async(
             host_charts_dict=self.host_charts_dict, host_prefix=True, host_sep='::', wide=True, sort_cols=True,
             protocol=self.protocol, numeric_only=True, float_size='float32', user=self.username, pwd=self.password
-            )[self.expected_cols]
+            )
```
ilyam8, line 290: `self.expected_cols` is no longer used, you can remove it.

```cmd
[ilyam@pc python.d]$ grep self.expected_cols anomalies.chart.py
        self.expected_cols = []
        self.expected_cols = list(df_train.columns)
            )[self.expected_cols]
```

ilyam8 (approved):

thiagoftsm (approved):

Issue #10457, opened 2021-01-06
Incomplete package from PackageCloud


CberYellowstone:
When I install netdata `v1.28.0` from the `packagecloud`,  Web UI send a massage like this
>The installer could not prepare the required dependencies to enable Netdata Cloud functionality
![image](https://user-images.githubusercontent.com/37031767/103771607-fefde000-5062-11eb-81e7-d9dc4de1e41b.png)

Both `.rpm` or `.deb` have this problem, And there is also a message from `netdata-claim.sh`:

>This agent was built without the dependencies for Cloud and cannot be claimed

I'm not the only one with this problem, @Habetdin has this problem, too

the system I use is `centos 7.9` , and  @Habetdin 's is `Ubuntu 18.04`

looking for a reply @prologic @underhood

underhood:
@CberYellowstone @Habetdin thanks for reporting the issue. There is a PR that should fix this already. Apologize for the delayed answer during the Christmas holidays.

x86txt:
Hi, this issue is affecting your .deb package too. I just installed it today and am seeing the same error in the GUI and when manually running netdata-claim.sh, I receive: 

    This agent was built without the dependencies for Cloud and cannot be claimed

This is on Debian Buster.

Issue #10403, opened 2020-12-18
1.27.0.deb built without Cloud dependencies


dtipro:
##### Bug report summary
After updating to 1.27.0 via the .deb package my existing nodes are no longer able to communicate with Netdata Cloud and new nodes are unable to be claimed.  When opening the dashboard I see a message stating:

> The installer could not prepare the required dependencies to enable Netdata Cloud functionality

When trying to claim a node I receive:

> This agent was built without the dependencies for Cloud and cannot be claimed

This was working with the 1.26.0 .deb package and works if 1.27.0 is installed via the kickstart script.

##### OS / Environment
Can reproduce on real hardware and in a VM
```
/etc/lsb-release:DISTRIB_ID=Ubuntu
/etc/lsb-release:DISTRIB_RELEASE=18.04
/etc/lsb-release:DISTRIB_CODENAME=bionic
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 18.04.5 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="18.04.5 LTS (Bionic Beaver)"
/etc/os-release:ID=ubuntu
/etc/os-release:ID_LIKE=debian
/etc/os-release:PRETTY_NAME="Ubuntu 18.04.5 LTS"
/etc/os-release:VERSION_ID="18.04"
/etc/os-release:HOME_URL="https://www.ubuntu.com/"
/etc/os-release:SUPPORT_URL="https://help.ubuntu.com/"
/etc/os-release:BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
/etc/os-release:PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
/etc/os-release:VERSION_CODENAME=bionic
/etc/os-release:UBUNTU_CODENAME=bionic
```

##### Netdata version
netdata v1.27.0
 

##### Component Name
packaging

##### Steps To Reproduce

1. Install netdata 1.27.0 using the .deb from packagecloud
2. Attempt to register with Netdata Cloud

##### Expected behavior
Netdata agent registers with Netdata Cloud

stevecorya:
I have the same issue with Debian 10; upgrade from 1.26.0 to 1.27.0 causes eight nodes to display "All nodes have errors" in the dashboard. Downgrading to 1.26.0 brings them all back up. 

I have not tried re-installing 1.27.0 with the kickstart script.

[...]

PR #9984, opened 2020-09-24
allows use of system libwebsockets instead of bundled one

underhood:
##### Summary
Allows using system `libwebsockets` instead of one bundled by the installer.

More detailed discussion #8961

If system LWS is used configure will check if the version is recent enough (min `3.2`)
Code has been updated to support up to the latest released LWS at the point of this PR creation (`4.1.4`)

##### Test Plan
Use configure with `--without-bundled-lws` on system with libwebsockets.
Use configure as normal on system without libwebsockets.
Everything should work as expected and/or appropriate error message should be printed.
Check with `ldd netdata` if output is correct.

PR #10460, opened 2021-01-06
Fixed bundling of libwebsockets in binary packages.

Ferroin:
##### Summary

Unintentionally broken by #9984. This re-enables cloud support in binary packages.

##### Component Name

area/packaging

##### Test Plan

Without this patch, our binary packages do not have cloud support.

With this patch, they have cloud support again.

##### Additional Information

Fixes: #10457 
Fixes: #10403 

Apologies to everyone affected by the bug this is fixing that it took so long to fix.

### DIFF
--- a/contrib/debian/rules
+++ b/contrib/debian/rules
@@ -39,7 +39,8 @@ override_dh_auto_configure:
 	packaging/bundle-libbpf.sh .
 	autoreconf -ivf
 	dh_auto_configure -- --prefix=/usr --sysconfdir=/etc --localstatedir=/var --libdir=/usr/lib \
-	--libexecdir=/usr/libexec --with-user=netdata --with-math --with-zlib --with-webdir=/var/lib/netdata/www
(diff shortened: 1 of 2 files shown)


underhood (approved):

underhood:
@Ferroin just as an idea? How difficult it would be to parse output of `./netdata -W buildinfo` and check all expected features are there?

underhood:
@Ferroin additional question why we not let it be shared lib and make it RPM dependency?

Ferroin:
> @Ferroin additional question why we not let it be shared lib and make it RPM dependency?

Long-term that’s the goal. For the moment though this is a fix that gets us back to a known working state without requiring the extensive cross-distro testing that should be done when we switch to using the native LWS packages instead of bundling our own.

Essentially I’m trying to just get this working again for our users ASAP, and will then worry about making it better.

mfundul (approved):

PR #10466, opened 2021-01-07
Update pfsense.md

OdysLam:
change netdata to newer version.

On a relevant note, @joelhans how hard/valuable would it be to have a GH action that combs our documentation with every netdata agent release and increments the version?

Would it be possible to have the version as a variable in the templating engine so whenever the docs are generated, the correct version is added? Thoughts?

### DIFF
--- a/packaging/installer/methods/pfsense.md
+++ b/packaging/installer/methods/pfsense.md
@@ -56,7 +56,7 @@ pkg add http://pkg.freebsd.org/FreeBSD:11:amd64/latest/All/py37-yaml-5.3.1.txz
 You can now install Netdata from the FreeBSD repository.
 
 ```bash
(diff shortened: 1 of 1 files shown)


joelhans (approved):
@OdysLam I searched the repo for `1.26` and `1.27`, and the only results are this doc and the FreeBSD doc, both of which suffer from the same problem of being consistently out-of-date. That's part of the reason I added the banner at the top to warn users or get more contributions.

I'm not sure how valuable a GHA is for two documents. There's no templating language or anything on the Docusaurus side to help make that any easier for us.

The most reliable action we could take is to add to our release checklist. For example, I edit the main README with news about the new release once its out. That process could also include incrementing the version in these two docs, or maybe a less-precise find-replace across the whole repo.

thiagoftsm (approved):

Issue #10421, opened 2020-12-24
rpm build faild


Jolin-blank:
```
In file included from ./database/rrd.h:37,
                 from ./daemon/common.h:36,
                 from ./web/api/web_api_v1.h:6,
                 from web/api/formatters/rrd2json.h:6,
                 from web/api/formatters/charts2json.h:6,
                 from web/api/formatters/charts2json.c:3:
./database/../aclk/aclk_common.h:71: error: redefinition of typedef 'RRDHOST'
./database/rrd.h:7: note: previous declaration of 'RRDHOST' was here
In file included from ./database/rrd.h:37,
                 from ./daemon/common.h:36,
                 from ./web/api/web_api_v1.h:6,
                 from web/api/formatters/rrd2json.h:6,
                 from web/api/formatters/json_wrapper.h:6,
                 from web/api/formatters/json_wrapper.c:3:
./database/../aclk/aclk_common.h:71: error: redefinition of typedef 'RRDHOST'
./database/rrd.h:7: note: previous declaration of 'RRDHOST' was here
In file included from ./database/rrd.h:37,
                 from ./daemon/common.h:36,
                 from ./web/api/web_api_v1.h:6,
                 from web/api/formatters/rrd2json.h:6,
                 from web/api/formatters/rrdset2json.h:6,
                 from web/api/formatters/rrdset2json.c:3:
./database/../aclk/aclk_common.h:71: error: redefinition of typedef 'RRDHOST'
./database/rrd.h:7: note: previous declaration of 'RRDHOST' was here
make[2]: *** [web/api/formatters/json_wrapper.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: *** [web/api/formatters/charts2json.o] Error 1
make[2]: *** [web/api/formatters/rrdset2json.o] Error 1
make[2]: Leaving directory `/root/rpmbuild/BUILD/netdata-v1.28.0'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/root/rpmbuild/BUILD/netdata-v1.28.0'
make: *** [all] Error 2
error: Bad exit status from /var/tmp/rpm-tmp.N2Mgf1 (%build)
```

dgsus:
im getting the same but via the kickstart script on centos 6

no problem on dependencies
Loading /etc/centos-release ...
You should have EPEL enabled to install all the prerequisites.
Check: http://www.tecmint.com/how-to-enable-epel-repository-for-rhel-centos-6-5/
[...]

PR #10470, opened 2021-01-08
Fix for older compilers

underhood:
##### Summary
Older compilers (PRE C11) will complain:
```
./database/../aclk/aclk_common.h:71: error: redefinition of typedef 'RRDHOST'
./database/rrd.h:7: note: previous declaration of 'RRDHOST' was here
```
The 2011 C standard allows redeclaration of typedef names. 6.7 3 says:
> typedef name may be redefined to denote the same type as it currently does, provided that type is not a variably modified type

This caused error on older systems like CentOS6 with old compiler version, see user reported issue:
Fixes https://github.com/netdata/netdata/issues/10421

##### Component Name
ACLK, Build System

##### Test Plan
Should build on older systems like CentOS 6 (tested)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -713,6 +713,7 @@ set(BACKENDS_PLUGIN_FILES
 set(CLAIM_PLUGIN_FILES
         claim/claim.c
         claim/claim.h
+        aclk/aclk_rrdhost_state.h
         aclk/aclk_common.c
         aclk/aclk_common.h
         )
--- a/Makefile.am
+++ b/Makefile.am
@@ -524,6 +524,7 @@ PARSER_FILES = \
     $(NULL)
 
 ACLK_FILES = \
+    aclk/aclk_rrdhost_state.h \
     aclk/aclk_common.c \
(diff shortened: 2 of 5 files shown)


Ferroin:
Are we certain we actually want to support older compilers here? The only platform we even remotely support that people are likely to be on a version old enough for this to be an issue is CentOS 6, and that’s EOL. Everything else we support has had a C11 compatible compiler for at least the past five years.

If the answer is yes, then we need to confirm that we do not violate C99 requirements anywhere else.

Either way, we need to start properly enforcing this (by passing the proper option to the compiler) _and_ document that we need a C compiler that supports whatever minimum C standard we want to require. We can then add a CI check to ensure we continue to meet that standard independent of what distro we are on.

underhood:
@Ferroin well this PR also cleans the whole thing up. The thing older compilers complained about actually also pointed to a bit of mess that was created.

By another words. This improves things even if we ignore the fixes CentOS 6 build aspect.

stelfrag (approved):

vlvkobal (approved):

PR #10428, opened 2020-12-27
Spelling build

jsoref:
##### Summary

This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

##### Component Name

> [stuff](https://github.com/netdata/netdata/pull/10428#issuecomment-760294967) that @Ferroin is responsible for

##### Test Plan

The misspellings have been reported at https://github.com/jsoref/netdata/commit/33c5b20ed9511dac0a23ee916fa247cc2f388304#commitcomment-45444734

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/netdata/commit/e2737c090e4b1d7b1bce405b9f702bcf526852a3

##### Additional Information

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

PR #10474, opened 2021-01-11
GitHub action markdown link check update

jsoref:
##### Summary
Tolerate GitHub `429`s as hit in https://github.com/netdata/netdata/pull/10428

This upgrades the action per gaurav-nelson/github-action-markdown-link-check#75

##### Component Name
area/ci

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Run link check
-        uses: gaurav-nelson/github-action-markdown-link-check@1.0.7
+        uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
           use-quiet-mode: 'no'
           use-verbose-mode: 'yes'
--- a/.mlc_config.json
+++ b/.mlc_config.json
@@ -1,7 +1,7 @@
 {
     "ignorePatterns": [
 		{
-			"pattern": "^https:\/\/pi-hole.net"
+			"pattern": "^https:\/\/pi-hole\.net\b"
 		}
 	],
     "replacementPatterns": [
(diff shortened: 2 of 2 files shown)


joelhans (approved):
Been wondering about how to get around the 429 errors before. Nice!

Pardon my GHA ignorance—does specifying `@v1` mean we're always running the latest `v.1.X` version of the action? As in, right now, [1.0.12](https://github.com/gaurav-nelson/github-action-markdown-link-check/releases/tag/1.0.12)?

ilyam8:
> does specifying @v1 mean we're always running the latest v.1.X version of the action?

Yes, to be specific - not version, but tag.

https://github.com/actions/toolkit/blob/master/docs/action-versioning.md#versioning

> Binding to a major version is the latest of that major version ( e.g. v1 == "1.*" )

jsoref:
@joelhans: Technically how `@v1` behaves is up to the project maintainer (@gaurav-nelson). But yes, the convention is that it's a tag which will float forward on its own within `1.*` releases.

Instead of using that, you could enable [dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) for GitHub Actions and have it try to drag you forward.

I picked `v1` because that's what the maintainer suggested.

I personally am manually bringing projects forwards for my own GitHub Action (and using a prerelease branch while developing spelling branches for others). It means upgrades are slower (as here), but it gives me a bit more control of rollout (and I don't have enough test coverage to know that I won't break something).

Ferroin (approved):
We’ve actually mostly been keying off of major versions for our own workflows, so I’m not 100% certain how this ended up specifying a full version in the first place. Definitely improves consistency with the rest of our workflows to fix this one.

Given that and that this appears to fix the issue correctly, I’m going to go ahead and approve and merge. Thank you @jsoref for this fix!

jsoref:
Oops. I just realized that the other PR _is_ dependabot, so you're already using it. -- Too many things in the air.

Ferroin:
@jsoref It’s OK, the other changes here actually fix a problem, and in general we do probably want this tracking major versions instead of an exact patch release.

PR #10477, opened 2021-01-11
Cleanup libnetdata headers

vkalintiris:
##### Summary

Include only what we use in the implementation files of libnetdata.

##### Component Name

libnetdata

##### Test Plan

Tested by building netdata on Linux with/without a combination of different preprocessor defs (eg. NETDATA_INTERNAL_CHECKS, NETDATA_VERIFY_LOCKS, etc).


vkalintiris:
FWIW, I tested this on FreeBSD 12.2 and it works fine AFAICT.

libnetdata/sysinc.h
```diff
@@ -0,0 +1,152 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
```
thiagoftsm, line 2: Please, add to this file these code:

```
#ifdef __cplusplus
extern "C" {
# endif
```
and

```
# ifdef __cplusplus
}
# endif
```

I am asking this, because we had an user contribution last year that added them, because our headers were not 100% compatible with c++ compilers. Considering that good part, if not all the code, is coming from `libnetdata.h` and it was present inside these conditions, it is possible that c++ developers could have the problem again.

I already tested this PR on `Slackware current` and `FreeBSD 12.2` and I did not observe any warning, after this change I will approve it.

thiagoftsm (dismissed):
The last commits removed my concerns, and I also could compile this PR on `Slackware current` and `Debian 10`.

vkalintiris:
Closing this because it's been inactive for quite a while and rebasing will generate far too many conflicts.

PR #10482, opened 2021-01-12
Claiming retry/backoff

underhood:
##### Summary
Due to various network conditions claiming can fail. This is problem especially if run automatically trough Docker etc. where user just manually running script again is no go.

This implements retry independent of retry logic of wget/curl and common for them.

Me and bash are not best of friends :) so feel free to make alternative PR/proposal. This was supposed to quick & dirty do this for testing purposes.

##### Component Name
Claiming

##### Test Plan
Block port temporarily and then unblock it.


Ferroin (approved):
Approving based on code inspection, still needs testing.

mfundul (approved):

PR #10479, opened 2021-01-11
Truncate excessive information from titles for apps and cgroups

vlvkobal:
##### Summary
There is excessive information in the titles of apps and group charts. We should truncate it.

##### Component Name
apps plugin, cgroups plugin

PR #10483, opened 2021-01-12
Support multiple chart label keys in data queries

stelfrag:
##### Summary
When a data query with context is used, support `chart_label_key` with multiple labels instead of just one. Only one of the keys needs to exist for the query to succeed. 

Keys that do not exist will return a null result set. 

example

executing a query with `chart_label_key=key1,key2,key_not_found`

will return 

``` 
     "chart_labels":  { "key1": ["val1", "val2"], "key2": ["val3", "val4"] , "key_not_found": [null, null] } 
```

##### Component Name
database

##### Test Plan
- Test on an agent with label support that monitors K8S 
- Execute a data query to a known `context` with charts that have labels e.g. something like
   - api/v1/data?context=target_context&after=-600&option=jsonwrap&chart_label_key=key1,key2

- You should see 
   - `"chart_labels": ` in the json header for both keys

### DIFF
--- a/database/rrdlabels.c
+++ b/database/rrdlabels.c
@@ -140,13 +140,21 @@ void update_label_list(struct label **labels, struct label *new_labels)
 
 struct label *label_list_lookup_key(struct label *head, char *key, uint32_t key_hash)
 {
+    SIMPLE_PATTERN *pattern = NULL;
+
+    if (!key_hash)
+        pattern = simple_pattern_create(key, ",|\t\r\n\f\v", SIMPLE_PATTERN_EXACT);
+
     while (head != NULL)
     {
-        if (head->key_hash == key_hash && !strcmp(head->key, key))
-            return head;
+        if ((key_hash && head->key_hash == key_hash && !strcmp(head->key, key)) ||
+            (pattern && simple_pattern_matches(pattern, head->key)))
+            break;
         head = head->next;
     }
-    return NULL;
+    if (pattern)
+        simple_pattern_free(pattern);
+    return head;
 }
 
 int label_list_contains_key(struct label *head, char *key, uint32_t key_hash)
--- a/database/rrdlabels.c
+++ b/database/rrdlabels.c
@@ -140,21 +140,13 @@ void update_label_list(struct label **labels, struct label *new_labels)
 
 struct label *label_list_lookup_key(struct label *head, char *key, uint32_t key_hash)
 {
-    SIMPLE_PATTERN *pattern = NULL;
-
-    if (!key_hash)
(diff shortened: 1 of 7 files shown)


vlvkobal (approved):

ilyam8:
@stelfrag please rebase on the master, we want to test this branch in k8s but we need to have https://github.com/netdata/netdata/pull/10479 fix in it.

database/rrdlabels.c
```diff
@@ -159,6 +159,31 @@ int label_list_contains(struct label *head, struct label *check)
     return label_list_contains_key(head, check->key, check->key_hash);
 }
 
+struct label *label_list_lookup_keylist(struct label *head, char *key)
+{
+    SIMPLE_PATTERN *pattern = NULL;
+
+    pattern = simple_pattern_create(key, ",|\t\r\n\f\v", SIMPLE_PATTERN_EXACT);
+
+    if (!pattern)
```
thiagoftsm, line 168: The function `simple_pattern_create` calls (`mallocz`)[https://github.com/netdata/netdata/blob/15a81d70d6ffa7b950c797ca8f1058b97d32495d/libnetdata/simple_pattern/simple_pattern.c#L92] that will shutdown everything when there is not memory available, so this test is not necessary.

stelfrag:
> @stelfrag please rebase on the master, we want to test this branch in k8s but we need to have #10479 fix in it.

rebased

libnetdata/simple_pattern/simple_pattern.c
```diff
@@ -354,3 +354,12 @@ char *simple_pattern_trim_around_equal(char *src) {
...
     return store;
 }
+
+char *simple_pattern_iterate(SIMPLE_PATTERN **p)
+{
+    struct simple_pattern *root = (struct simple_pattern *) *p;
+    struct simple_pattern **Proot = (struct simple_pattern **)p;
+
+    (*Proot) = (*Proot)->next;
+    return root->match;
```
vlvkobal, line 364:
```
libnetdata/simple_pattern/simple_pattern.c: In function 'simple_pattern_iterate':
libnetdata/simple_pattern/simple_pattern.c:364:16: warning: return discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  364 |     return root->match;
      |            ~~~~^~~~~~~
```

vlvkobal (dismissed):

database/rrdlabels.c
stelfrag: I have removed it. There are others checks in place to make sure this works fine even without this condition.

vlvkobal (approved):

thiagoftsm (approved):

Issue #8673, opened 2020-04-10
Mismatch of file permissions between spec file and installer


jorti:
##### Bug report summary
For the external plugins that require additional privileges, there is a mismatch of the file capabilities and permissions in the spec file and what the installer configures:
 
```
# netdata.spec.in
%caps(cap_dac_read_search,cap_sys_ptrace=ep) %attr(0750,root,netdata) %{_libexecdir}/%{name}/plugins.d/apps.plugin
%caps(cap_setuid=ep) %attr(4750,root,netdata) %{_libexecdir}/%{name}/plugins.d/cgroup-network
%caps(cap_setuid=ep) %attr(4750,root,netdata) %{_libexecdir}/%{name}/plugins.d/perf.plugin
%caps(cap_setuid=ep) %attr(4750,root,netdata) %{_libexecdir}/%{name}/plugins.d/slabinfo.plugin
%caps(cap_setuid=ep) %attr(4750,root,netdata) %{_libexecdir}/%{name}/plugins.d/freeipmi.plugin
```

```
# netdata-installer.sh
/usr/libexec/netdata/plugins.d/apps.plugin = cap_dac_read_search,cap_sys_ptrace+ep (root:netdata 0750)
/usr/libexec/netdata/plugins.d/cgroup-network (root:netdata 4750)
/usr/libexec/netdata/plugins.d/perf.plugin (root:netdata 4750)
/usr/libexec/netdata/plugins.d/slabinfo.plugin (root:netdata 4750)
/usr/libexec/netdata/plugins.d/freeipmi.plugin (root:netdata 4750)
```

##### OS / Environment
```
/etc/fedora-release:Fedora release 32 (Thirty Two)
/etc/os-release:NAME=Fedora
/etc/os-release:VERSION="32 (Workstation Edition)"
/etc/os-release:ID=fedora
/etc/os-release:VERSION_ID=32
/etc/os-release:VERSION_CODENAME=""
/etc/os-release:PLATFORM_ID="platform:f32"
/etc/os-release:PRETTY_NAME="Fedora 32 (Workstation Edition)"
/etc/os-release:ANSI_COLOR="0;34"
/etc/os-release:LOGO=fedora-logo-icon
/etc/os-release:CPE_NAME="cpe:/o:fedoraproject:fedora:32"
/etc/os-release:HOME_URL="https://fedoraproject.org/"
[...]

PR #10490, opened 2021-01-12
Fixed handling of permissions for some plugins.

Ferroin:
##### Summary

* Removed pointless capabilities in our RPM spec file.
* Shifted slabinfo and perf plugins to use proper capabilties instead of being SUID root.

##### Component Name

area/packaging

##### Test Plan

Verification is needed to ensure that the perf and slabinfo plugins still report data correctly with this change.

##### Additional Information

Fixes: #8673

### DIFF
--- a/contrib/debian/netdata.postinst
+++ b/contrib/debian/netdata.postinst
@@ -54,9 +54,9 @@ case "$1" in
     chown -R root:netdata /usr/libexec/netdata/plugins.d
     chown -R root:netdata /var/lib/netdata/www
     setcap cap_dac_read_search,cap_sys_ptrace+ep /usr/libexec/netdata/plugins.d/apps.plugin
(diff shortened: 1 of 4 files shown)


Saruspete:
Hello

Nice cleanup ! 

Just a note: capabilities are stored in the Extended attributes of the FS, more specifically in `security.capabilities`.
Maybe in the **very unlikely** event the FS doesn't support xattr (nfs, ext3...)  it would be wise to check if the destination folder support xattr (or is exported through NFS)

Cheers,

Ferroin:
@Saruspete Good point about xattrs. We kind of try to handle that in our own installer but it doesn’t really work correctly right now. Long term we should probably fix that (and add proper checking to our packages, but I think for the moment we’re probably safe just assuming they’re supported as only drastically non-standard setups are likely to not support security xattrs at all.

knatsakis (approved):

PR #10492, opened 2021-01-13
Gitignore with spell check

jsoref:


jsoref:
I did *not* mean to do this

PR #10507, opened 2021-01-14
Switched to using system libwebsockets for RPM builds.

Ferroin:
##### Summary

Also cleans up the configure statement in the RPM spec file.

##### Component Name

area/packaging
area/ci

##### Test Plan

Verified locally for CentOS 7 and 8, Fedora 32 and 33, and OpenSUSE 15.1 and 15.2 by building in Docker containers.

##### Additional Information

This also coincidentally fixes a bug introduced by #10460 that is causing RPM package builds to consistently fail.

We unfortunately cannot do this for DEB packages yet due to the fact that the Debian and Ubuntu environments provide versions of LWS that are too old.


vkalintiris (dismissed):
LGTM, but let's wait for @underhood's approval as well.

underhood:
@Ferroin do we have any control about min/max version of libwebsockets our package requires?
We are quite flexible but still we should at least check min version.

vkalintiris (approved):

underhood (approved):

Ferroin:
Merging as the CI failure is unrelated to the changes in the PR.

PR #10508, opened 2021-01-14
Spelling md

jsoref:
##### Summary
Split from #10428

### DIFF
--- a/collectors/python.d.plugin/anomalies/README.md
+++ b/collectors/python.d.plugin/anomalies/README.md
@@ -142,7 +142,7 @@ local:
     # just the average of all anomaly probabilities at each time step
     include_average_prob: true
 
-    # Define any custom models you would like to create anomaly probabilties for, some examples below to show how.
+    # Define any custom models you would like to create anomaly probabilities for, some examples below to show how.
     # For example below example creates two custom models, one to run anomaly detection user and system cpu for our demo servers
     # and one on the cpu and mem apps metrics for the python.d.plugin.
     # custom_models:
--- a/collectors/python.d.plugin/anomalies/README.md
+++ b/collectors/python.d.plugin/anomalies/README.md
@@ -134,7 +134,7 @@ local:
     diffs_n: 1
 
     # What is the typical proportion of anomalies in your data on average? 
-    # This paramater can control the sensitivity of your models to anomalies. 
+    # This parameter can control the sensitivity of your models to anomalies. 
     # Some discussion here: https://github.com/yzhao062/pyod/issues/144
     contamination: 0.001
 
--- a/collectors/python.d.plugin/anomalies/README.md
+++ b/collectors/python.d.plugin/anomalies/README.md
@@ -215,7 +215,7 @@ If you would like to go deeper on what exactly the anomalies collector is doing 
   - Typically ~3%-3.5% additional cpu usage from scoring, jumping to ~60% for a couple of seconds during model training.
   - About ~150mb of ram (`apps.mem`) being continually used by the `python.d.plugin`.
 - If you activate this collector on a fresh node, it might take a little while to build up enough data to calculate a realistic and useful model.
-- Some models like `iforest` can be comparatively expensive (on same n1-standard-2 system above ~2s runtime during predict, ~40s training time, ~50% cpu on both train and predict) so if you would like to use it you might be advised to set a relativley high `update_every` maybe 10, 15 or 30 in `anomalies.conf`.
+- Some models like `iforest` can be comparatively expensive (on same n1-standard-2 system above ~2s runtime during predict, ~40s training time, ~50% cpu on both train and predict) so if you would like to use it you might be advised to set a relatively high `update_every` maybe 10, 15 or 30 in `anomalies.conf`.
 - Setting a higher `train_every_n` and `update_every` is an easy way to devote less resources on the node to anomaly detection. Specifying less charts and a lower `train_n_secs` will also help reduce resources at the expense of covering less charts and maybe a more noisy model if you set `train_n_secs` to be too small for how your node tends to behave.
 
 ## Useful links and further reading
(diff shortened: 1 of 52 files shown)


collectors/tc.plugin/README.md
```diff
@@ -172,7 +172,7 @@ And this is what you are going to get:
 
 ## QoS Configuration with tc
 
-First, setup the tc rules in rc.local using commands to assign different DSCP markings to different classids. You can see one such example in [github issue #4563](https://github.com/netdata/netdata/issues/4563#issuecomment-455711973). 
+First, setup the tc rules in rc.local using commands to assign different DHCP markings to different classids. You can see one such example in [github issue #4563](https://github.com/netdata/netdata/issues/4563#issuecomment-455711973).
```
vlvkobal, line 175: It should be left as DSCP (stands for Differentiated Services Code Point).

Ferroin: Perhaps we should change it to ‘QoS markings’? It’s obviously less precise, but both looks less like a typo and is less likely to be misinterpreted by someone who does not have a background in low-level networking.

ilyam8: I think DSCP is well-known by people who know what QoS is. [COS](https://en.wikipedia.org/wiki/Class_of_service) is QoS markings too.

ilyam8: Example, we are referring to, has `dsfield`. 

```
dsfield VAL_MASK_8
  IPv4 only. Match the packet header's DSCP/ECN
  field. Synonyms to this are tos and precedence.
```

Actually i agree - `QoS markings` will do.

This always was confusing to me - QoS, TOS, COS, IP Precedence, DSCP, ECN 🤪

collectors/python.d.plugin/anomalies/README.md
```diff
@@ -206,7 +206,7 @@ If you would like to go deeper on what exactly the anomalies collector is doing
...
... (or each custom model). The default settings for these parameters aim to create a rolling matrix of recent smoothed [differenced](https://en.wikipedia.org/wiki/Autoregressive_integrated_moving_average#Differencing) values for each chart. The aim of the model then is to score how unusual this 'matrix' of features is for each chart based on what it has learned as 'normal' from the training data. So as opposed to just looking at the single most recent value of a dimension and considering how strange it is, this approach looks at a recent smoothed window of all dimensions for a chart (or dimensions in a custom model) and asks how unusual the data as a whole looks. This should be more flexible in capturing a wider range of [anomaly types](https://andrewm4894.com/2020/10/19/different-types-of-time-series-anomalies/) and be somewhat more robust to temporary 'spikes' in the data that tend to always be happening somewhere in your metrics but often are not the most important type of anomaly (this is all covered in a lot more detail in the [deepdive tutorial](https://nbviewer.jupyter.org/github/netdata/community/blob/main/netdata-agent-api/netdata-pandas/anomalies_collector_deepdive.ipynb)).
```
joelhans, line 209: Can't see the changed word here—not sure if it's GitHub's UI or some other issue. Any way to pick it out?

joelhans (commented):
Aside from my one unknown and the other "DSCP" item being discussed, these all appear to be good and valid changes. Should be able to approve and merge once we have those figured out.

collectors/python.d.plugin/anomalies/README.md
ilyam8: flexibile => flexible 🤓

joelhans: How?!? Your proficiency with all things Git continues to amaze...

jsoref: `git diff --word-diff` is your friend for some of this fwiw (if you have a local repo and are willing to pull in the branch -- there's magic for being able to do that).

Also, until I squash, you can look at the log for the file and see the commit, on average the commit message will at least give you the replacement string:

1. [View file]( https://github.com/netdata/netdata/blob/6a6c2bb33255ad9b919cd7bfba246160c8ba12ad/collectors/python.d.plugin/anomalies/README.md)
2. [History]( https://github.com/netdata/netdata/commits/59c30f63ae66262e3c8dcbf66aa8b308b7accfea/collectors/python.d.plugin/anomalies/README.md)
3. From here, you can just check each commit
4. Or, use [Blame]( https://github.com/netdata/netdata/blame/6a6c2bb33255ad9b919cd7bfba246160c8ba12ad/collectors/python.d.plugin/anomalies/README.md)
5. And scroll to the [line]( https://github.com/netdata/netdata/blame/6a6c2bb33255ad9b919cd7bfba246160c8ba12ad/collectors/python.d.plugin/anomalies/README.md#L209)
6. In this case, the UI clearly hints about the most recent change. You can then use the "[previous change](https://github.com/netdata/netdata/blame/aa8cf9aef6c4fac9a6520b5e9c8a7ae3ff0f4d78/collectors/python.d.plugin/anomalies/README.md)" thing in case you think there are multiple changes for the line.
7. Scrolling to the [same line]( https://github.com/netdata/netdata/blame/aa8cf9aef6c4fac9a6520b5e9c8a7ae3ff0f4d78/collectors/python.d.plugin/anomalies/README.md#L209) shows that it's unchanged from its surroundings.

collectors/tc.plugin/README.md
jsoref: I'd be more than happy to change it to QoS, that seems like a reasonable choice - fwiw, there are ~2x as many hits for `"QoS markings"` than for `"dscp markings"` and they appear to be reasonable hits.

---

Alternatively, I'd suggest adding a couple of other instances of the word (even in this section). That's probably achievable by inlining some content from the link instead of forcing people to follow the link.

I tend to leave things alone if there are multiple hits and thus hint at being terms of art. Here, this is the _only_ instance of the token which made it suspect. (But not apparently enough for me to Google it.)

---

Reading the sentence, `Setup` should actually be `Set up`, so if this project has a documentation team, I'd suggest flagging this document to them for review :-).

---

Let me know which way you want to go. I wouldn't want to spend a lot of time trying to fix this if someone else w/ more domain knowledge could improve the section.

The easiest thing for me to do would be to drop the commit and file an issue asking for someone to improve the section (citing this thread).

joelhans: > if this project has a documentation team

Hey, that's me! :stuck_out_tongue: I don't have enough domain expertise on QoS to put in a vote either way, but I trust both @Ferroin and @ilyam8 to have the reader's best understanding in mind with their suggestions.

I'll make a note in my backlog to look at this doc, and this section, for additional improvements as a separate action/PR.

jsoref: I've pushed `QoS`. I think that'll enable people to move forward with this later.

I won't have any more time until next week.

joelhans (approved):
I think, with the `QoS` change made, this is good to merge.

ilyam8 (approved):

vlvkobal (approved):

PR #10524, opened 2021-01-19
Reduce the number of alarm updates on ACLK

stelfrag:
##### Summary
Send alarm updates only if the new or old status is warning or critical. 

##### Component Name
aclk
health

##### Test Plan
- Compile agent with `-DNETDATA_INTERNAL_CHECKS=1` and cloud support. This will record and `Sending MQTT` messages as well as the PUBACK messages in the log file e.g. 
  `ACLK_Query_4 : Sending MQTT len=1121 mid=534 wq=0 (0-bytes) readq=0: {	"type": "status-change"`
  `ACLK_Main : Publish_callback: mid=5 latency=158ms`
  **Note:**  A `status-change` message refers to an alarm status update
- Start the agent and notice the total of PUBACKS as the agent initializes. 
- Apply the PR, recompile and observe the reduced number of messages (PUBACKS)

### DIFF
--- a/health/health_log.c
+++ b/health/health_log.c
@@ -153,8 +153,12 @@ inline void health_alarm_log_save(RRDHOST *host, ALARM_ENTRY *ae) {
         }
     }
 #ifdef ENABLE_ACLK
-    if (netdata_cloud_setting)
-        aclk_update_alarm(host, ae);
+    if (netdata_cloud_setting) {
+        if ((ae->new_status == RRDCALC_STATUS_WARNING || ae->new_status == RRDCALC_STATUS_CRITICAL) ||
+            ((ae->old_status == RRDCALC_STATUS_WARNING || ae->old_status == RRDCALC_STATUS_CRITICAL))) {
+            aclk_update_alarm(host, ae);
+        }
+    }
 #endif
 }
 


health/health_log.c
```diff
@@ -153,8 +153,12 @@ inline void health_alarm_log_save(RRDHOST *host, ALARM_ENTRY *ae) {
         }
     }
 #ifdef ENABLE_ACLK
-    if (netdata_cloud_setting)
-        aclk_update_alarm(host, ae);
+    if (netdata_cloud_setting) {
+        if ((ae->new_status == RRDCALC_STATUS_WARNING || ae->new_status == RRDCALC_STATUS_CRITICAL) ||
+            ((ae->old_status == RRDCALC_STATUS_WARNING || ae->old_status == RRDCALC_STATUS_CRITICAL))) {
```
underhood, line 158: how about WARNING->CLEAR and CRITICAL->CLEAR?

underhood:
you can also compile with ACLK_DISABLE_CHALLENGE connect to any broker and count the messages by mosq_sub

underhood:
## Agent startup
before change:
```
...
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Got status-change v2
Total ACLK messages: 262, Matching Filter: 260
```
after this change:
```
Total ACLK messages: 2, Matching Filter: 0
```

underhood:
Initial `connect_alarms` message sets all alarms as
`	"status": "UNINITIALIZED",`
your PR now will send update only if it changes to WARNING or CRITICAL or changes from WARNING or CRITICAL.
That means that until alarm fires all alarms will be forever `UNINITIALIZED` -> question is can it cause trouble for cloud?

stelfrag:
> Initial `connect_alarms` message sets all alarms as
> ` "status": "UNINITIALIZED",`
> your PR now will send update only if it changes to WARNING or CRITICAL or changes from WARNING or CRITICAL.
> That means that until alarm fires all alarms will be forever `UNINITIALIZED` -> question is can it cause trouble for cloud?
> Perhaps we should always send CLEAR... or hold initial `connect_alarms` till all alarms are in some valid for cloud state?

Good point. It is tricky especially if we want to try and "fix" broken states in the cloud. Let me describe the reasoning to verify this doesn't make things worse :smile: 

- If agent - cloud is in-sync, on agent start WARNING and CRITICAL will be sent (so last status will be WARN/CRIT). This is because the last status will be read from the health db file. 
   - If the alarm clears, the update will be sent because old status will be WARN / CLEAR
- If agent - cloud is not in sync (so cloud shows WARN/CRIT) but on agent restart the stored status is not WARN/CRIT then this doesn't help even today (we do send all the updates)

If an alarm reaches the CLEAR state the cloud needs to know it only if we had previously told it it was WARN/CRIT. We have told it either:
- On connect from the health db
- As a regular update message (so new status would be "anything" to WARN / CRIT)

There is another issue of stale status in the health - db, that will be addressed with another PR and has to do with alarms that will never fire again because the alarm doesn't exist anymore and needs to be handled on the "on connect" message.

underhood (approved):

thiagoftsm (approved):
I ran a test now with a High frequency alarm (27 alarms in 122 seconds), and  the changes send the alarms as expected. LGTM!

mfundul (approved):

PR #10531, opened 2021-01-20
fix postgres password bug and change default config

OdysLam:
##### Summary

Fix bug on postgres collector where a `None` password was output as `*****` to the user. 

Add default password to default configuration, so that netdata can collect from containerised postgresql instances, as the no-password `trust` configuration doesn't work for non-local connections.

##### Test Plan

- Run Postgres in container and see that it automatically connects to the instance.
- Comment-out the configuration jobs that contain the password and see that the debug trace outputs `password: None`

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -1162,7 +1162,7 @@ def zero_lock_types(databases):
 
 
 def hide_password(config):
-    return dict((k, v if k != 'password' else '*****') for k, v in config.items())
+    return dict((k, v if (k != 'password' or (k == 'password' and v == None) ) else '*****') for k, v in config.items())
 
 
 def add_database_lock_chart(order, definitions, database_name):
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -1162,7 +1162,7 @@ def zero_lock_types(databases):
 
 
 def hide_password(config):
-    return dict((k, v if (k != 'password' or (k == 'password' and v == None) ) else '*****') for k, v in config.items())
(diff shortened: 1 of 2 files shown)


collectors/python.d.plugin/postgres/postgres.conf
```diff
@@ -132,3 +136,26 @@ tcpipv6:
...
+    host     : '127.0.0.1'
+    port     : 5432
+
+tcpipv6:
+    name     : 'local'
+    database : 'postgres'
+    user     : 'postgres'
+    passowrd:  'postgres'    
+    host     : '::1'
+    port     : 5432
```
ilyam8, line 161: We decode configuration in a `dict`, so kyes should be unique.

Adding `password: postgres` to existine tcp* job will do, i think we don't need additional jobs.

collectors/python.d.plugin/postgres/postgres.chart.py
```diff
@@ -1162,7 +1162,7 @@ def zero_lock_types(databases):
 
 
 def hide_password(config):
-    return dict((k, v if k != 'password' else '*****') for k, v in config.items())
+    return dict((k, v if (k != 'password' or (k == 'password' and v == None) ) else '*****') for k, v in config.items())
```
ilyam8, line 1165:
```suggestion
    return dict((k, v if k != 'password' or not v else '*****') for k, v in config.items())
```

Simplified: we want password to be not `******` if None or ''.

collectors/python.d.plugin/postgres/postgres.conf
```diff
@@ -116,6 +120,7 @@ tcp:
     name     : 'local'
     database : 'postgres'
     user     : 'postgres'
+    password : 'postgres'
```
ilyam8, line 123: lets add password for tcpipv4 and tcpipv6 jobs

ilyam8 (approved):

thiagoftsm (approved):

Issue #10536, opened 2021-01-21
netdata-updater removes custom tmp dir


therealgilles:
##### Bug report summary
I use /root/tmp as NETDATA_TMPDIR_PATH.

/etc/netdata/.environment:
NETDATA_TMPDIR_PATH="/root/tmp"

After the /etc/cron.daily/netdata-updater runs, the /root/tmp directory is gone.

##### OS / Environment
```
/etc/lsb-release:DISTRIB_ID=Ubuntu
/etc/lsb-release:DISTRIB_RELEASE=20.04
/etc/lsb-release:DISTRIB_CODENAME=focal
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="20.04.1 LTS (Focal Fossa)"
/etc/os-release:ID=ubuntu
/etc/os-release:ID_LIKE=debian
/etc/os-release:PRETTY_NAME="Ubuntu 20.04.1 LTS"
/etc/os-release:VERSION_ID="20.04"
/etc/os-release:HOME_URL="https://www.ubuntu.com/"
/etc/os-release:SUPPORT_URL="https://help.ubuntu.com/"
/etc/os-release:BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
/etc/os-release:PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
/etc/os-release:VERSION_CODENAME=focal
/etc/os-release:UBUNTU_CODENAME=focal
```

##### Netdata version
netdata v1.28.0

##### Component Name
netdata-updater

##### Steps To Reproduce
Run:
```SHELL
> mkdir /root/tmp
> /etc/cron.daily/netdata-updater

Checking if a newer version of the updater script is available.
Downloading newest version of updater script.
Checking if a newer version of the updater script is available.
Wed 20 Jan 2021 04:01:28 PM PST : INFO:  Running on a terminal - (this script also supports running headless from crontab)
Wed 20 Jan 2021 04:01:28 PM PST : INFO:  Current Version: 00102800000000
Wed 20 Jan 2021 04:01:28 PM PST : INFO:  Latest Version: 00102800000000
Wed 20 Jan 2021 04:01:28 PM PST : INFO:  Newest version (current=00102800000000 >= latest=00102800000000) is already installed
```

Then on the 2nd run, I get this:
[...]

PR #10550, opened 2021-01-25
Properly handle saved temporary directory on updates.

Ferroin:
##### Summary

We should have been using the saved `TMPDIR` value as a prefix for a temporary directory, but where using it directly. This resulted in errors in the updater.

##### Component Name

area/packaging

##### Test Plan

Verify that the custom temporary directory is not removed by the script on the first run of the updater.

##### Additional Information

Fixes: #10536


knatsakis (approved):

kaskavel (approved):

therealgilles:
@Ferroin: I think the mktemp call on line 139 should be outside the if statement.

PR #10557, opened 2021-01-26
fix minor vulnerability alert, updating socket-io dependency

jacekkolasa:
##### Summary
Fix minor vulnerability alert, by updating socket-io dependency in package-lock.json.
This change does not have any impact on the Agent, the dependency was only used for automated testing one of the node collectors.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -289,6 +289,30 @@
         "defer-to-connect": "^1.0.1"
       }
     },
+    "@types/component-emitter": {
+      "version": "1.2.10",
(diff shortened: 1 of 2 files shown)


thiagoftsm (approved):
I tested the PR using Firefox private mode and I had one problem to load a JS file one time, but I could not replicate it. I am assuming that this is a problem related with my browser, except for this problem, all the other tests that I did was successful .

jacekkolasa:
Thanks @thiagoftsm for checking, but i'm sure this change has nothing to do with Dashboard, this file is mainly used by people who wish to contribute to building node.js collectors.

stelfrag (approved):

PR #10570, opened 2021-01-29
Update claiming docs for Docker containers.

Ferroin:
##### Summary

This adds information about claiming Docker containers using environment variables (the currently preferred claiming method) and removes info about invoking the claiming script by passing options for the agent to the `docker run` command.

##### Component Name

area/docs

##### Test Plan

n/a

##### Additional Information

Based on discussion with @manos-saratsis.

The diff is a bit disjoint because of how `git`’s diff algorithm works, the changes are really a lot less complicated than they look.

### DIFF
--- a/claim/README.md
+++ b/claim/README.md
@@ -80,12 +80,13 @@ you don't see the node in your Space after 60 seconds, see the [troubleshooting 
 
 ### Claim an Agent running in Docker
 
-The claiming process works with Agents running inside of Docker containers. You can either set a specific set
-of environment variables in the container to have it automatically claimed on startup or restart, or you can use
-`docker exec` to manually claim an already-running container.
+The claiming process works with the Netdata Agent running inside of a Docker container. You can either set a
+specific set of environment variables in the container to have it automatically claimed on startup or restart,
+or you can use `docker exec` to manually claim an already-running container.
 
-Note that for claiming to wokr, the contents of `/var/lib/netdata` _must_ be preserved across container restarts
-using a persistent volume.
+For claiming to work, the contents of `/var/lib/netdata` _must_ be preserved across container restarts
+using a persistent volume.  See our [recommended `docker run` and Docker Compose
+examples](/packaging/docker/README.md#create-a-new-netdata-agent-container) for details.
 
 #### Using environment variables
 
@@ -96,9 +97,9 @@ The Netdata Docker container looks for the following environment variables on st
 - `NETDATA_CLAIM_ROOMS`
 - `NETDATA_CLAIM_PROXY`
 
-If the token and URL are specified in their corresponding variables _and_ the container is not already claimed, it
-will use these values to attempt to claim the container, automatically adding the node to the specified rooms. If
-a proxy is specified, it will be used for the claiming process and for connecting to the cloud.
+If the token and URL are specified in their corresponding variables _and_ the container is not already claimed,
+it will use these values to attempt to claim the container, automatically adding the node to the specified War
+Rooms. If a proxy is specified, it will be used for the claiming process and for connecting to Netdata Cloud.
 
 These variables can be specified using any mechanism supported by your container tooling for setting environment
 variables inside containers. For example, when creating a new Netdata continer using `docker run`, the following
@@ -130,7 +131,7 @@ as it works in the widest variety of situations and simplifies configuration man
 
 #### Using docker exec
 
-Claim a _running Agent container_ by appending the script offered by Cloud to a `docker exec ...` command, replacing
+Claim a _running Netdata Agent container_ by appending the script offered by Cloud to a `docker exec ...` command, replacing
 `netdata` with the name of your running container:
 
 ```bash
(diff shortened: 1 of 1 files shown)


claim/README.md
```diff
@@ -80,29 +80,29 @@ you don't see the node in your Space after 60 seconds, see the [troubleshooting
 
 ### Claim an Agent running in Docker
 
-The claiming process works with Agents running inside of Docker containers. You can use `docker exec` to run the
-claiming script on containers already running, or append the claiming script to `docker run` to create a new container
-and immediately claim it.
+The claiming process works with Agents running inside of Docker containers. You can either set a specific set
+of environment variables in the container to have it automatically claimed on startup or restart, or you can use
+`docker exec` to manually claim an already-running container.
```
joelhans, line 85:
```suggestion
The claiming process works with the Netdat Agent running inside of a Docker container. Use a specific set
of environment variables in the container to have it automatically claimed on startup or restart, or use
`docker exec` to manually claim an already-running container.
```

claim/README.md
```diff
@@ -80,29 +80,29 @@ you don't see the node in your Space after 60 seconds, see the [troubleshooting
...
 
-#### Running Agent containers
+Note that for claiming to wokr, the contents of `/var/lib/netdata` _must_ be preserved across container restarts
+using a persistent volume.
```
joelhans, line 88: Perhaps we can link 

```suggestion
For claiming to work, the contents of `/var/lib/netdata` _must_ be preserved across container restarts
using a persistent volume. See our [recommended `docker run` and Docker Compose examples](/packaging/docker/README.md#create-a-new-netdata-agent-container) for details.
```

claim/README.md
```diff
@@ -80,29 +80,29 @@ you don't see the node in your Space after 60 seconds, see the [troubleshooting
...
-The script should return `Agent was successfully claimed.`. If the claiming script returns errors, or if
-you don't see the node in your Space after 60 seconds, see the [troubleshooting information](#troubleshooting).
+- `NETDATA_CLAIM_TOKEN`
+- `NETDATA_CLAIM_URL`
+- `NETDATA_CLAIM_ROOMS`
+- `NETDATA_CLAIM_PROXY`
 
-#### New/ephemeral Agent containers
+If the token and URL are specified in their corresponding variables _and_ the container is not already claimed, it
+will use these values to attempt to claim the container, automatically adding the node to the specified rooms. If
```
joelhans, line 100:
```suggestion
will use these values to attempt to claim the container, automatically adding the node to the specified War Rooms. If
```

claim/README.md
```diff
@@ -80,29 +80,29 @@ you don't see the node in your Space after 60 seconds, see the [troubleshooting
...
+a proxy is specified, it will be used for the claiming process and for connecting to the cloud.
```
joelhans, line 101:
```suggestion
a proxy is specified, it will be used for the claiming process and for connecting to Netdata Cloud.
```

claim/README.md
Ferroin: I’m curious about the reasoning behind shifting to a more imperative wording here.

joelhans: Lately I've been more critical about Netdata's writing—maybe it's the time on spent on the style guide—and feel we should be more active in our tone. I prefer "do X or Y" over "you can do X or Y" because it's shorter and feels more empowering and inclusive.

We're prolific in our use of "you can," from the beginning of the project. I don't know if I inherited it from the existing docs or brought more of the tendency in myself, but I'm trying to correct myself and move docs away from that style as I approach it. I even wrote a Go script to parse the docs and list every one that uses "you can," among other things...

joelhans: Also, I typo'd **Netdata** in my suggestion. Whoops.

claim/README.md
```diff
@@ -114,24 +115,32 @@ docker run -d --name=netdata \
   -v /proc:/host/proc:ro \
   -v /sys:/host/sys:ro \
   -v /etc/os-release:/host/etc/os-release:ro \
+  -e NETDATA_CLAIM_TOKEN=TOKEN \
+  -e NETDATA_CLAIM_URL=https://app.netdata.cloud \
```
manos-saratsis, line 119: @joelhans in all cases the claim URL needs to be in double quotes such as "https://app.netdata.cloud" Cc @Ferroin

Ferroin: This was actually my mistake.

claim/README.md
Ferroin: OK, the shorter part I get, but I personally feel that using `do X or Y` sounds out of place in documentation unless you use it in a way that doesn’t result in it sounding like a command.

IOW, this should probably be ‘To claim an instance of the Netdata Agent running in a Docker container, either do X or do Y’, otherwise it sounds like we’re commanding the reader to do something instead of explaining how to do it. The ‘either/or’ aspect is also kind of important here given how long each of the optional clauses is in the sentence, as it solidifies ahead of time that this is a set of choices.

joelhans (approved):
I believe this is ready—it concisely explains how this process works—unless @manos-saratsis has further requirements from the product side.

manos-saratsis (approved):

PR #10593, opened 2021-02-03
Add support for changing the number of pages per extent

mfundul:
##### Summary

Add support for changing the number of dbengine pages per disk extent in `netdata.conf`. Also support changing dynamically during runtime.

##### Component Name
database/engine

##### Test Plan

Play around with `netdata.conf`. Default and maximum value is (Supported range 1-64):
```
[global]
  dbengine extent pages = 64
```

You can change it dynamically by executing:
```
netdatacli write-config 'netdata|global|dbengine extent pages|32'
```

You can only observe the difference after a lot of extents have been written to disk, and most of the database files have rotated. Small extents will have significantly larger metadata:data ratio and as a result much smaller retention period (metric history).

You can play around with various workloads and compare performance with default settings (e.g. export a snapshot from the dashboard).

### DIFF
--- a/database/engine/rrdengine.c
+++ b/database/engine/rrdengine.c
@@ -9,6 +9,8 @@ rrdeng_stats_t rrdeng_reserved_file_descriptors = 0;
 rrdeng_stats_t global_pg_cache_over_half_dirty_events = 0;
 rrdeng_stats_t global_flushing_pressure_page_deletions = 0;
 
+static unsigned pages_per_extent = MAX_PAGES_PER_EXTENT;
+
 static void sanity_check(void)
 {
     /* Magic numbers must fit in the super-blocks */
@@ -666,7 +668,7 @@ static int do_flush_pages(struct rrdengine_worker_config* wc, int force, struct 
          PValue = JudyLFirst(pg_cache->committed_page_index.JudyL_array, &Index, PJE0),
          descr = unlikely(NULL == PValue) ? NULL : *PValue ;
 
-         descr != NULL && count != MAX_PAGES_PER_EXTENT ;
+         descr != NULL && count != pages_per_extent ;
 
          PValue = JudyLNext(pg_cache->committed_page_index.JudyL_array, &Index, PJE0),
          descr = unlikely(NULL == PValue) ? NULL : *PValue) {
@@ -1031,6 +1033,21 @@ struct rrdeng_cmd rrdeng_deq_cmd(struct rrdengine_worker_config* wc)
     return ret;
 }
 
+static void load_configuration_dynamic(void)
+{
+    unsigned read_num;
+    static int printed_error = 0;
+
+    read_num = (unsigned) config_get_number(CONFIG_SECTION_GLOBAL, "dbengine extent pages",
+                                                    MAX_PAGES_PER_EXTENT);
+    if (read_num > 0 && read_num <= MAX_PAGES_PER_EXTENT) {
+        pages_per_extent = read_num;
+    } else if (!printed_error) {
+        printed_error = 1;
+        error("Invalid dbengine extent pages %u given. Defaulting to %u.", read_num, pages_per_extent);
+    }
+}
+
 void async_cb(uv_async_t *handle)
 {
     uv_stop(handle->loop);
@@ -1084,6 +1101,7 @@ void timer_cb(uv_timer_t* handle)
             }
         }
     }
+    load_configuration_dynamic();
 #ifdef NETDATA_INTERNAL_CHECKS
     {
         char buf[4096];


thiagoftsm:
@mfundul  I tested your PR now, and I observed that when I ran it without to add the new configuration inside `netdata.conf` everything worked as expected, after this I stop Netdata and I set the variable `dbengine extent pages = 64`, after the restart I also had data collection.

After the first test I removed completely my dbengine files, and I started Netdata using `dbengine extent pages = 256`. Finally I ran more three tests using values `64`, `2` and `1024`.

All these tests happening in an interval of 3 hours, there is only one thing that I wanna to confirm, I observed that the dashboard was not keeping old chart values every time I restart the agent, I think this is expected, can you confirm this?

mfundul:
Losing data is not expected, check if your binary is crashing during shutdown. Maybe you're using address sanitizer? Also, the supported range for pages per extent is 1 - 64. You should get some errors for invalid values, check the logs for those as well.

database/engine/rrdengine.c
```diff
@@ -1031,6 +1033,21 @@ struct rrdeng_cmd rrdeng_deq_cmd(struct rrdengine_worker_config* wc)
...
+    unsigned read_num;
+    static int printed_error = 0;
+
+    read_num = (unsigned) config_get_number(CONFIG_SECTION_GLOBAL, "dbengine extent pages",
+                                                    MAX_PAGES_PER_EXTENT);
+    if (read_num > 0 && read_num <= MAX_PAGES_PER_EXTENT) {
+        pages_per_extent = read_num;
+    } else if (!printed_error) {
+        printed_error = 1;
+        error("Invalid dbengine extent pages %u given. Defaulting to %u.", read_num, pages_per_extent);
```
thiagoftsm, line 1047: Considering your last comment, please, add the expected range of values here.

thiagoftsm:
> Losing data is not expected, check if your binary is crashing during shutdown. Maybe you're using address sanitizer? Also, the supported range for pages per extent is 1 - 64. You should get some errors for invalid values, check the logs for those as well.

Yes, I always compile with address sanitizer, and this was the motive I was losing data, but I do not have any explicit error message related to this fact, I only have old reports:

```
    #0 0x4a3c67 in __asan::AsanCheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) (/usr/sbin/netdata+0x4a3c67)
    #1 0x4c136a in __sanitizer::CheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) (/usr/sbin/netdata+0x4c136a)
    #2 0x4a83e4 in __asan::AsanThread::GetStackFrameAccessByAddr(unsigned long, __asan::AsanThread::StackFrameAccess*) (/usr/sbin/netdata+0x4a83e4)
    #3 0x41f1d7 in __asan::GetStackAddressInformation(unsigned long, unsigned long, __asan::StackAddressDescription*) (/usr/sbin/netdata+0x41f1d7)
    #4 0x420558 in __asan::AddressDescription::AddressDescription(unsigned long, unsigned long, bool) (/usr/sbin/netdata+0x420558)
    #5 0x422990 in __asan::ErrorGeneric::ErrorGeneric(unsigned int, unsigned long, unsigned long, unsigned long, unsigned long, bool, unsigned long) (/usr/sbin/netdata+0x422990)
    #6 0x4a3304 in __asan::ReportGenericError(unsigned long, unsigned long, unsigned long, unsigned long, bool, unsigned long, unsigned int, bool) (/usr/sbin/netdata+0x4a3304)
    #7 0x451d8a in unpoison_tm(void*, __sanitizer::__sanitizer_tm*) (/usr/sbin/netdata+0x451d8a)
    #8 0x459d92 in localtime_r (/usr/sbin/netdata+0x459d92)
    #9 0x521ebf in log_date libnetdata/log/log.c:456
    #10 0x521ebf in info_int libnetdata/log/log.c:718
    #11 0x80af93 in proc_main_cleanup collectors/proc.plugin/plugin_proc.c:85
    #12 0x80af93 in proc_main collectors/proc.plugin/plugin_proc.c:91
    #13 0x53ad04 in thread_start libnetdata/threads/threads.c:184
    #14 0x7f9ff2f5a004 in start_thread (/lib64/libpthread.so.0+0x8004)
    #15 0x7f9ff2e6658e in __clone (/lib64/libc.so.6+0x10c58e)
```


When I compiled without the sanitizer, everything worked perfectly.

About the error messages, I had them as expected.

database/engine/rrdengine.c
```diff
@@ -1084,6 +1101,7 @@ void timer_cb(uv_timer_t* handle)
             }
         }
     }
+    load_configuration_dynamic();
```
stelfrag, line 1104: I think it can be expensive to load the config all the time -- perhaps only when `cmd_write_config_execute` is called?

mfundul: In theory you are correct. In practice I don't think it's going to be a problem to have 2 AVL extra lookups per second.

stelfrag (approved):

thiagoftsm (approved):

MrZammler (approved):

PR #10599, opened 2021-02-04
Add a new workflow to test that updater works as expected

kaskavel:
Add workflow for installing latest version of netdata, building from source of this branch & running the updater. Add script build-dist to create the artifacts used for the update.

##### Summary

We add a new pipeline, meant to run in PRs and merges to master, that uses the kickstart script to install the latest version of netdata, then checkout the branch of the PR, build netdata and update to this version of the agent. 

##### Test Plan

Currently the pipeline has been tested for latest Debian and Fedora, more distributions will be added in the near future, along with more checks regarding the updater's behavior.

### DIFF
--- a/.github/scripts/check-updater.sh
+++ b/.github/scripts/check-updater.sh
@@ -13,7 +13,7 @@ check_successfull_update() {
   (
     netdata_version=$(netdata -v | awk '{print $2}')
     updater_version=$(cat packaging/version)
-    if [ "$netdata_version" == "$netdata_version" ]; then
+    if [ "$netdata_version" == "$updater_version" ]; then
       echo "Update successfull!"
     else
       exit 1
--- a/.github/scripts/check-updater.sh
+++ b/.github/scripts/check-updater.sh
@@ -1,0 +1,42 @@
+#!/bin/sh
+#
+set -e
+# shellcheck source=.github/scripts/functions.sh
+. "$(dirname "$0")/functions.sh"
+
+NAME="${NAME:-netdata}"
+VERSION="${VERSION:-"$(git describe --always)"}"
+BASENAME="$NAME-$VERSION"
+
+check_successfull_update() {
+  progress "Check netdata version after update"
+  (
+    netdata_version=$(netdata -v | awk '{print $2}')
+    updater_version=$(cat packaging/version)
+    if [ "$netdata_version" == "$netdata_version" ]; then
+      echo "Update successfull!"
+    else
+      exit 1
+    fi
+  ) >&2
+}
+
+steps="check_successfull_update"
+
+_main() {
+  for step in $steps; do
+    if ! run "$step"; then
+      if [ -t 1 ]; then
+        debug
+      else
+        fail "Build failed"
+      fi
+    fi
+  done
+
+  echo "🎉 All Done!"
+}
+
+if [ -n "$0" ] && [ x"$0" != x"-bash" ]; then
+  _main "$@"
+fi
--- a/.github/scripts/check-updater.sh
+++ b/.github/scripts/check-updater.sh
@@ -4,16 +4,12 @@ set -e
 # shellcheck source=.github/scripts/functions.sh
 . "$(dirname "$0")/functions.sh"
 
-NAME="${NAME:-netdata}"
-VERSION="${VERSION:-"$(git describe --always)"}"
-BASENAME="$NAME-$VERSION"
-
 check_successfull_update() {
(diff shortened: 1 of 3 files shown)


.github/workflows/updater.yml
```diff
@@ -0,0 +1,41 @@
+---
+name: Updater
+on:
+  push:
+    branches: [ master ]
```
Ferroin, line 5: In most other places, we have specifically tended towards long-form lists for this type of thing in our workflow files, even if it’s just a single item

I _think_ you can just list the branch name directly though if it’s only one branch, without needing the list structure.

Ferroin (approved):

Ferroin:
@dimko @vkalintiris I would appreciate if one of the two of you could take a look at this to make sure I haven’t missed anything.

The shellcheck warning is expected but harmless (I’ll cover a fix for it in a follow-up PR, as I was planning on doing one anyway to sync up the list of distros for this with what we are using for regular build checks).

The two failing build checks are unrelated to this PR (it’s not touching any code that would cause those to fail, and they’re also failing elsewhere), and thus can be ignored for the purposes of reviewing this PR.

.github/workflows/updater.yml
```diff
@@ -0,0 +1,60 @@
...
+        distro:
+          - 'debian:10'
+          - 'fedora:33'
+          - 'ubuntu:20.10'
+          - 'alpine:3.13'
+          - 'archlinux:latest'
+          - 'centos:8'
+          - 'clearlinux:latest'
+          - 'opensuse/tumbleweed:latest'
+    runs-on: ubuntu-latest
```
dimko, line 26: I am a bit confused. How do you specify that the pipeline will run using a different distribution? shouldn't the **runs-on** be something like ${{ matrix.distro }}?

Ferroin: Actually, I somehow missed that we’re not handling the per-distro runs correctly. Nice catch.

@kaskavel To actually run cross-distro, we need to run the code to be tested inside a Docker container. So the install, script modifications, and updater run need to be bundled into a single script that can then run inside a Docker container.

kaskavel: This issue has now been addressed so feel free to take another look!

vkalintiris (dismissed):

.github/scripts/check-updater.sh
```diff
@@ -0,0 +1,42 @@
...
+NAME="${NAME:-netdata}"
+VERSION="${VERSION:-"$(git describe --always)"}"
+BASENAME="$NAME-$VERSION"
+
+check_successfull_update() {
+  progress "Check netdata version after update"
+  (
+    netdata_version=$(netdata -v | awk '{print $2}')
+    updater_version=$(cat packaging/version)
+    if [ "$netdata_version" == "$updater_version" ]; then
```
Ferroin, line 16: This needs to be `=`.

.github/scripts/check-updater.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+#
+set -e
+# shellcheck source=.github/scripts/functions.sh
+. "$(dirname "$0")/functions.sh"
+
+NAME="${NAME:-netdata}"
+VERSION="${VERSION:-"$(git describe --always)"}"
+BASENAME="$NAME-$VERSION"
```
Ferroin, line 9: Shellcheck is correct that this appears to be unused. It should either be removed, or if used externally exported.

Ferroin (changes requested):

Ferroin (approved):

vkalintiris (approved):

.github/scripts/check-updater.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+#
+set -e
+# shellcheck source=.github/scripts/functions.sh
+. "$(dirname "$0")/functions.sh"
```
dimko, line 5: Do double quotes inside double quotes work? Doesn't this need escaping?

dimko (approved):

.github/scripts/check-updater.sh
Ferroin: There’s a layer between the two sets of double quotes because of how the shell parsing works.

Effectively, anything inside a `$()` is parsed as regular shell code in a context completely isolated from the surrounding code (albeit inheriting the environment variables from the surrounding context), so something like `"$(echo ")` would fail to parse due to the unmatched double quote inside the `$()`, not due to the unmatched parenthesis.

PR #10602, opened 2021-02-05
add stats per cloud query type

underhood:
##### Summary
`http` cloud query can query different `/api/v1/info` endpoints. Add chart counting how many queries of each type have been requested by the cloud.

##### Component Name
ACLK

##### Test Plan
Look at the new charts through the cloud (you need to have through cloud traffic to see anything)


MrZammler (approved):

MrZammler:
Hi Timo! The only thing I could mention perhaps (although it's not that important I guess), is the same priority `200005` of the charts (perhaps it's intentional?).

underhood:
@MrZammler thanks that is indeed an oversight. I rebased this PR on latest master and updated the prios

stelfrag (approved):

MrZammler (approved):

PR #10616, opened 2021-02-08
add aggregator collector

andrewm4894:
aggregation collector to aggregate charts from multiple child nodes into one chart at each time step.

##### Summary

Add's an aggregator collector that can aggregate charts from multiple child nodes into one 'aggregated' chart on the parent node. 

The need for this arose out of dogfooding the anomalies collector on our parent netdata node where we have multiple child nodes subsets of which correspond to different production services or clusters we run. 

For example if we have a parent with 10 children 5 of which are from GKE cluster 1 while the other 5 are from GKE cluster 2. Then what this collector does is enable us to aggregate all charts from GKE cluster 1 into a set of aggregated charts for the whole GKE cluster 1 and likewise for GKE cluster 2. 

The idea extends to any situation where you have some children you would like to view in aggregate on your netdata parent as opposed to just being able to see the charts for each child in isolation on the parent hosts.  

##### Component Name

/collectors/python.d.plugin/aggregator

##### Test Plan

Tested locally with two dev nodes streaming to one parent.


andrewm4894:
Adding the example screenshot here so i can use this like in the README docs. 

netdata-aggregator-collector

ilyam8:
> The idea extends to any situation where you have some children you would like to view in aggregate on your netdata parent as opposed to just being able to see the charts for each child in isolation on the parent hosts.

Since this implemented on the Cloud and represented in a much better way

Screenshot 2021-03-05 at 15 28 32

Is it a temorary solution until claiming mvp0 implemented? 

Why do you think we need this collector?

ilyam8:
Anyway please apply code auto-formatting, there are some problems.

andrewm4894:
> Why do you think we need this collector?

- If i have a parent node and want to aggregate some of the children charts into one view on my parent without having to use cloud then i can use this. 
- If i want to do anomaly detection on that aggregated set of child nodes (or metrics correlation for that matter or alarms even) then i have to use something like this as i can not do that via composite charts. 
- Composite charts should be the better and more 'correct' way to do this assuming you are using cloud and we have more features built into composite charts (which we don't quite yet but should over next couple of months) but that is not an argument for not doing this. This just gives users a bit more flexibility and more options.

smhaller:
Is there a reason why this pull request got closed and not merged?

andrewm4894:
@smhaller nothing technical as such but more just prioritization of other work. Is this something you would find useful in some way? 

The best way to do this currently (and most flexible way) is by claiming the children themselves into a room in netdata cloud. For example below i have my "devml" space and in the default "General" room with the 3 nodes claimed to it and so i can get what i need via the composite charts capability in netdata cloud - aggregate each of my 3 nodes charts into one chart and explore it in various way, different aggregations on the fly and focus on individual dims etc. 

So this would be the best way to get similar but better functionality by using netdata cloud.

image

That said, if above not really an option for you or you could see yourself using this aggregator collector in some way in the netdata agent dashboard on the host (as opposed to via netdata cloud) then i'm sure it could be opened again and we see what might be involved in merging it (my original use case was a bit specific to a particular use case i had on a parent node so i closed it as was able to do more via composite charts in netdata cloud).

smhaller:
Yes, I see. Thank you for your explanation! At the moment I am not allowed to use anything else but inhouse storage. So the cloud in the current form is not an option - if there would be a possibility to self-host the cloud it would be different.

My solution for now is just to locally  pull your pr into the master branch - this works fine for me. More generally I would like to see the data aggregation directly in a netdata release.

andrewm4894:
Cool, I'll follow up with the team to resurrect this PR. Feel free to make any changes you find useful in your branch too that you think others might as you use it and we could include them too.

d235j:
@andrewm4894 Are there any plans to resurrect this PR? I am very interested in this functionality.

PR #10618, opened 2021-02-09
dashboard@v2.13.6

jacekkolasa:
##### Summary
Dashboard v2.13.6

Posthog test
netdata/dashboard#215
https://github.com/netdata/dashboard/issues/223

Test posthog on demo agents and on any agents when localStorage.setItem("posthog-dev", "true") flag is set.

##### Component Name
Dashboard


andrewm4894 (approved):
I have tested all the posthog related aspects of this on some dev nodes and all looks good to me in terms of functionality and what we are looking for on the PostHog side of things.

thiagoftsm (approved):
I confess I do not know exactly what is the posthog, but considering that @andrewm4894 approved this after some tests, I am approving, because everything I tested was working when the flag was set as true.

PR #10620, opened 2021-02-11
Posthog dev

MrZammler:
##### Summary
Provide analytics data to PostHog. Also provide data that are currently not sent to GA.

##### Component Name
Agent

##### Test Plan

It needs to run under a variety of test scenarios (i.e. agent streams, is a parent, etc) to make sure the data is correct.

##### Additional Information
I've moved the analytics related functions to a new file.

There's now 3 levels of setting env variables for analytics: The already existing `set_global_environment`(1), a new `set_late_global_environment`(2) that runs after `rrd_init`, and a new analytics thread(3), that will wait for a while (TBD), before setting up more variables (for things that are not yet available at startup).

(1) + (2) are sent via the existing START action.
(3) is sent via a new META action.

There are still attributes to be added, and some of these should be added to api/v1/info endpoint also.


MrZammler:
Perhaps the thread could be left up, and check for changes that don't require a restart and send them if any (i.e. when a node is claimed) ?

andrewm4894:
Thanks for this @MrZammler. From chatting to @Ferroin a bit about this i think the plan is to first merge in a more basic switch of current functionality from GA to the Netdata PostHog [here](https://github.com/netdata/netdata/pull/10636) and then once that is live we will merge something like this PR that captures a bit more of the useful data for the anonymous statistics. 

So idea being go live with PostHog for what we have first and then once we happy with it, merge in this PR which will be some better and more useful data for the anonymous statistics that will be going to the Netdata PostHog instance. 

So i think it makes sense this lives here as a draft PR for now and then my hope would be in a few weeks with can just merge it once we happy with the GA to posthog migration of the existing events.

MrZammler:
All right! Ok, so I will try and add more data to this PR in the mean time!

MrZammler:
So, as an update, this PR currently:

1. Has 3 places to collect data:

- On agent start, just like before.
- After `rrd_init` (late environment).
- In a thread.

What is collected from the first 2 points, will appear in the START event. (Note: All events carry the same attributes. On the START event, those attributes that haven't been collected yet will have a value of 'N/A'.).

Then, the analytics thread comes up, and waits a specified amount of time (to consider the system stable) and collects more data. It sends a META event with those data, and exits. It is up to see if some data should/can be collected in late or in the thread. It is useful to have as many data as possible in the START event, to have some information in case of a crash (where the META will never be sent).

Just before sending an EXIT event, it will try to gather data again (those that are counted from the thread) and send them on the EXIT event.

2. What it currently collects (only those added by this PR will be listed):

Posthog attribute | Description | Gathered at | Notes
--------------------- | ------------- | ------------------|-------
config_stream_enabled | If streaming is enabled on the host | late environment | default_rrdpush_enabled
config_memory_mode | Global memory mode | late environment | default_rrd_memory_mode
config_page_cache_size | Engine page cache size | late environment | default_rrdeng_page_cache_mb
config_multidb_disk_quota | Engine disk quota | late environment | default_multidb_disk_quota_mb
config_https_enabled | HTTPS enabled | late environment | #ifdef ENABLE_HTTPS
config_web_enabled | If web mode is enabled | late environment | web_server_mode
config_update_every | Update every X sec | late environment | default_rrd_update_every
host_collectors_plugins | List of collector's plugins | thread | 
host_collectors_modules | List of collector's modules | thread |
host_collectors_count | Count of collectors | thread
host_alarms_count | Count of all alarms | thread | Could be a list of warning/critical/normal instead of total
host_charts_count | Count of charts | thread
host_metrics_count | Count of metrics | thread
config_is_parent | If the host is a parent | thread |
config_hosts_available | Available hosts | thread | rrd_hosts_available
config_cloud_enabled | If cloud is enabled in the config | late environment | Config section global -> cloud_config -> enabled
host_aclk_enabled | If aclk is enabled | thread | #ifdef ENABLE_ACLK
host_claimed | If host is claimed | thread | is_agent_claimed()
host_aclk_connected | If aclk is connected | thread | aclk_connected
config_exporting_enabled | If exporting is enabled in the config | thread | Config section exporting -> enabled
host_notifications_methods | The enabled/available notification methods | thread | Internally runs the alarm-notify.sh script which is modified to accept a parameter that will dump the SEND_XXX set to YES variables.
host_allmetrics_prometheus_used | If the allmetrics page was called for prometheus | thread | If the allmetrics export page was called for prometheus at least `ANALYTICS_MAX_PROMETHEUS_HITS` times, it will be set to true.
host_allmetrics_shell_used | If the allmetrics page was called for shell | thread | If the allmetrics export page was called for shell at least `ANALYTICS_MAX_SHELL_HITS` times, it will be set to true.
host_allmetrics_json_used | If the allmetrics page was called for json | thread | If the allmetrics export page was called for json at least `ANALYTICS_MAX_JSON_HITS` times, it will be set to true. Also note for the three above: If the allmetrics hits don't get called enough to set to true for the META event, they will be still counted on the EXIT event.

3. Some of those are also set to the web/v1/info page as a test. We can set what's not already there too.
4. This is not a final list of attributes, just the current state. We should also add e.g. any set backend exporters too.

andrewm4894:
@MrZammler adding a few tasks or small changes - happy to add as separate tickets or some better way if needed 

- [ ] rename `host_claimed` to `host_agent_claimed`
- [ ] convert `host_alarms_count = 79|0|1` into `alarms_normal = 79`, `alarms_warning = 0`, `alarms_critical = 1`

MrZammler:
Ok, Andy, I'll also update the gsheet when I'm done with the renames, etc.

andrewm4894:
@MrZammler i may have asked this already but as part of this work are you also able to make all the backend attributes available via the `/api/v1/info` endpoint so that the frontend events can also just pick them up and include them?

MrZammler:
Yes! I know, it's on the todo!

I've also added a new column in the gsheet where I will update with the names the attributes will have on the info page.

MrZammler:
Added a log for registry?hello hit from the dashboard, to indicate if the dashboard has been used. This only happens though on a first connect of the dashboard to the agent. If the dashboard page is left open, and the agent restarts, the dashboard will reconnect without hitting registry?hello.

MrZammler:
I think we should close this PR, and start multiple smaller ones....

andrewm4894:
@MrZammler can you add me on the smaller PR's. I won't really be able to review them but would like to keep up to date on them as they go through review.

PR #10634, opened 2021-02-15
cpu stats per query thread

MrZammler:
##### Summary

Adds a cpu usage chart for each query thread.

cpu_per_query_thread

##### Component Name

ACLK

##### Test Plan

The agent node must be claimed. 
View the agent from the cloud to create queries to the agent.
Observe the charts by e.g. adding load when requesting non visible charts via the cloud.
Observe the charts also on idle.

##### Additional Information

Query threads are measured only when they are responding to queries.

### DIFF
--- a/aclk/legacy/aclk_query.c
+++ b/aclk/legacy/aclk_query.c
@@ -630,7 +630,7 @@ static int aclk_process_query(struct aclk_query_thread *t_info)
         aclk_queries_per_thread[t_info->idx]++;
         ACLK_STATS_UNLOCK;
 
-        if (likely(getrusage_called_this_tick[t_info->idx]) < MAX_GETRUSAGE_CALLS_PER_TICK) {
+        if (likely(getrusage_called_this_tick[t_info->idx] < MAX_GETRUSAGE_CALLS_PER_TICK)) {
             getrusage(RUSAGE_THREAD, &rusage_per_thread[t_info->idx]);
             getrusage_called_this_tick[t_info->idx]++;
         }
--- a/aclk/legacy/aclk_query.c
+++ b/aclk/legacy/aclk_query.c
@@ -628,8 +628,13 @@ static int aclk_process_query(struct aclk_query_thread *t_info)
         ACLK_STATS_LOCK;
         aclk_metrics_per_sample.queries_dispatched++;
         aclk_queries_per_thread[t_info->idx]++;
-        getrusage(RUSAGE_THREAD, &rusage_per_thread[t_info->idx]);
         ACLK_STATS_UNLOCK;
+
+        if (likely(getrusage_called_this_tick[t_info->idx]) < MAX_GETRUSAGE_CALLS_PER_TICK) {
+            getrusage(RUSAGE_THREAD, &rusage_per_thread[t_info->idx]);
+            getrusage_called_this_tick[t_info->idx]++;
+        }
+
     }
 
     aclk_query_free(this_query);
--- a/aclk/legacy/aclk_query.c
+++ b/aclk/legacy/aclk_query.c
@@ -628,6 +628,7 @@ static int aclk_process_query(struct aclk_query_thread *t_info)
         ACLK_STATS_LOCK;
         aclk_metrics_per_sample.queries_dispatched++;
         aclk_queries_per_thread[t_info->idx]++;
+        getrusage(RUSAGE_THREAD, &rusage_per_thread[t_info->idx]);
         ACLK_STATS_UNLOCK;
     }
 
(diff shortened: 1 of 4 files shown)


aclk/legacy/aclk_query.c
```diff
@@ -628,6 +628,7 @@ static int aclk_process_query(struct aclk_query_thread *t_info)
         ACLK_STATS_LOCK;
         aclk_metrics_per_sample.queries_dispatched++;
         aclk_queries_per_thread[t_info->idx]++;
+        getrusage(RUSAGE_THREAD, &rusage_per_thread[t_info->idx]);
```
underhood, line 631: do we know how expensive this is? Just wonder if a single thread is processing a lot of trivial/quick queries... calling this a lot in this case....

MrZammler: Hi Timo, yes, I will try to find a way to minimize calls to getrusage and see. Thanks!

aclk/legacy/aclk_stats.c
```diff
@@ -222,11 +230,43 @@ static void aclk_stats_mat_metric_process(struct aclk_metric_mat *metric, struct
     rrdset_done(metric->st);
 }
 
+static void aclk_stats_cpu_threads(void)
+{
+    char id[100 + 1];
+    char title[100 + 1];
+    char dim_name[MAX_DIM_NAME];
```
stelfrag, line 237: dim_name does not seem to be used

MrZammler: Removed, thanks!

aclk/legacy/aclk_query.c
underhood: also I wonder if it would make sense moving syscall out of that lock

MrZammler: Ok! Will move it outside of ACLK_STATS_LOCK.

I've also tried some tests by recording the last `now_realtime_sec()` of each call of `getrusage` per thread, and only allow it running when the last recorded sec is different than now. I then tried scrolling my cloud view to get some load on the queries. There where indeed some denied calls to `getrusage` (that would happen on the same second for the same thread), but I'm not sure that the load of calling `now_realtime_sec` everytime would offset the extra-same-second calls to `getrusage`...

[aclk_queries_timestamps.txt](https://github.com/netdata/netdata/files/5981070/aclk_queries_timestamps.txt)

There should/could be a better way to implement the only-run-once-per-second trick, but I'll need to think of it a bit more !

underhood: I see if too complicated or not easily solvable let it be.

underhood: you could also use the fact that ACLK_stats thread clears some structs to 0 every tick to detect second passed with much lower overhead?

underhood: in `aclk_stats_main_thread` we do:
```
        memcpy(aclk_queries_per_thread_sample, aclk_queries_per_thread, sizeof(uint32_t) * query_thread_count);
        memset(aclk_queries_per_thread, 0, sizeof(uint32_t) * query_thread_count);
```
so you can set just boolean flag -> `getrusage_called_this_tick` and `query_stats` will clear it on every tick

MrZammler: Ok! That was a good idea! I was only a bit worried that we might perhaps lose some chart "resolution", if it would only be 1 or 0, so I added a `MAX_GETRUSAGE_CALLS_PER_TICK` to limit to e.g. 5 calls (we can change it) (per thread, per tick), if you don't mind.

underhood: Well there is an alternative if you are worried about that. In function `aclk_query_main_thread` you have:
```
        while (aclk_process_query(info)) {
            // Process all commands
        };
```
So you can call it every X queries (to handle a hypothetical case where single thread has so much work that it stays active here for 100 queries without qoing to COND_WAIT.
Then ensure it gets called once after this cycle?

MrZammler: Hi! It's ok, I'm pretty ok with the max calls 5, the charts looks good! We could even lower it to 2 or 3, and still have plenty of data, and still prevent it from putting pressure on the query handling routine. Thanks a lot!

underhood: Yes, I see what I worry about here is if the single thread does 100 queries now the data will be wrong. This is the motive behind calling once per X and ensure once call just before the thread goes to sleep.
But OK we can merge it like this if you are not concerned about that :)

underhood (approved):

aclk/legacy/aclk_query.c
MrZammler: OK thanks Timo! We'll see (as I get more comfortable with the code), in a new PR perhaps if it shows anything off.

stelfrag (approved):

underhood (approved):

MrZammler:
Thank you both!

PR #10636, opened 2021-02-15
replace GA with PostHog for backend telemetry events.

andrewm4894:
- update anonymous-statistics.sh.in to use Netdata PostHog instead of GA
- update docs accordingly

##### Summary

Change the anonymous statistics script to use the Netdata hosted PostHog instead of Google Analytics. Update the docs to reflect this too.

##### Component Name

daemon/anonymous-statistics.sh.in

##### Test Plan

This has been tested locally on development machines. The next step is to implement the script on the Netdata demo agents and validate it sends events as expected.


andrewm4894:
I am making this a draft PR as we will manually update the script on our demo agents and validate the events before merging fully with master.

thiagoftsm:
@andrewm4894 , please, take a look at `shellcheck` report.

docs/anonymous-statistics.md
```diff
@@ -15,49 +16,48 @@ the statistics gathered from this information for two purposes:
...
+    distinct_id: machineGuid,
+    $ip: "127.0.0.1",
+    $current_url: "agent dashboard",
+    $pathname: "netdata-dashboard",
+    $host: "dashboard.netdata.io",
+  })
+}
+```
+
+In the above snippet a Netdata PostHog session is initialized and the `ip`, `current_url`, `pathname` and `host` attributes are set to constant values for all events that may be sent during the session. This way things like the ip or hostname of the agent will not be sent as part of the product usage event data.
```
joelhans, line 52:
```suggestion
In the above snippet a Netdata PostHog session is initialized and the `ip`, `current_url`, `pathname` and `host` attributes are set to constant values for all events that may be sent during the session. This way, information like the IP or hostname of the Agent will not be sent as part of the product usage event data.
```

docs/anonymous-statistics.md
```diff
@@ -103,7 +103,10 @@ Each of these opt-out processes does the following:
...
+
+## Migration from Google Analytics and Google Tag Manager.
+
+Prior to vX.XX we used Google Analytics to capture this information. This led to discomfot with some of our users in sending any product usage data to a third party like Google. It was also not even that useful in terms of generating the insights we needed to help catch bugs early and find opportunities for product improvement as Google Analytics does not allow its users access to the raw underlying data without paying a significant amount of money which would be infeasible for a project like Netdata.
 
-[![analytics](https://www.google-analytics.com/collect?v=1&aip=1&t=pageview&_s=1&ds=github&dr=https%3A%2F%2Fgithub.com%2Fnetdata%2Fnetdata&dl=https%3A%2F%2Fmy-netdata.io%2Fgithub%2Fdocs%2Fanonymous-statistics&_u=MAC~&cid=5792dfd7-8dc4-476b-af31-da2fdb9f93d2&tid=UA-64295674-3)]()
+While we migrate fully away from Google Analytics to PostHog there maybe be a small period of time where we run both in parallel before we remove all Google Analytics related code. This is to ensure we can fully test and validate the Netdata PostHog implementation before fully defaulting to it.
```
joelhans, line 112: Even though it's sort of ironic, I believe the GA tag at the end (the `[!analytics](...` stuff) should remain, since we're using it to track pageviews, and it's sort of orthogonal to the (awesome) work you're doing in migrating the Agent anonymous statistics.

joelhans (commented):
Excited to push this and share with users when the time is right!

docs/anonymous-statistics.md
andrewm4894: yep - have been seeing that in the collector docs too. I think we probably can/cloud move this stuff into its own PostHOg project at a later stage if we want to. But agree should stay for now.

andrewm4894:
i think shellcheck is confused by the fact that its multiple lines all wrapped in a ' so i think that is why the shellcheck is failing the way it is.

Ferroin:
Based on a cursory look at the changes here, it looks like the JSON request bdy for the two possible cURL commands is largely the same regardless of which command gets invoked.

Given this, I would suggest constructing it in a variable using a here-document and then using the variable to pass it to the cURL commands. Essentially, using something like this to assign the value of the variable:

```bash
REQ_BODY="$(cat << EOF
{
  "a": 1,
  "b": "${SOME_VALUE}"
}
EOF
)"
```

Because here-documents are interpreted literally other than variable substitution, you can use quotes within them all you want without needing special handling, meaning that this approach would avoid the need for complex nested quotes, which should in turn both make the code easier to read and resolve the complaints from shellcheck.

andrewm4894:
@Ferroin thanks that's much better!

Ferroin (approved):
Approving based on code inspection of the shell script. Ideally there should be at least one review  from someone who understands better than i do how the PostHog stuff is supposed to work.

thiagoftsm (approved):
I am in the same situation of @Ferroin , the code looks ok, but it is necessary a review from somebody with PostHog knowledge to verify that nothing is missing inside `REQ_BODY`.

PR #10642, opened 2021-02-16
Validate JSON values in health code

vkalintiris:
##### Summary

This will make sure we don't send unparseable JSON to consumers of the alarm endpoints.

The logic behind this change is that there are multiple different places were the JSON values can come from. If any of them is malformed we don't want to create a malformed JSON. Instead, we should "fail" gracefully by annotating the bad values with a self-explanatory message. 

##### Component Name

health

##### Test Plan

Compared the JSON output of alarm endpoints against master to make sure they are identical.

### DIFF
--- a/health/health_json.c
+++ b/health/health_json.c
@@ -2,6 +2,7 @@
 
 #include "health.h"
 #include "libnetdata/libnetdata.h"
+#include <stdbool.h>
 
 #define BOOL_STR(x)         (x) ? "true" : "false";
 #define DEFAULT_STR(x, y)   (x) ? (x) : (y);
--- a/health/health_json.c
+++ b/health/health_json.c
@@ -1,92 +1,129 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
 #include "health.h"
+#include "libnetdata/libnetdata.h"
 
-static inline void health_string2json(BUFFER *wb, const char *prefix, const char *label, const char *value, const char *suffix) {
-    if(value && *value) {
-        buffer_sprintf(wb, "%s\"%s\":\"", prefix, label);
-        buffer_strcat_htmlescape(wb, value);
-        buffer_strcat(wb, "\"");
-        buffer_strcat(wb, suffix);
+#define BOOL_STR(x)         (x) ? "true" : "false";
+#define DEFAULT_STR(x, y)   (x) ? (x) : (y);
+
+static void
+health_kv_html_escape(BUFFER *wb, const char *key, const char *value,
+               const char *prefix, const char *suffix)
+{
+    if (!value || !*value) {
+        buffer_sprintf(wb, "%s\"%s\": null%s", prefix, key, suffix);
+        return;
     }
-    else
-        buffer_sprintf(wb, "%s\"%s\":null%s", prefix, label, suffix);
+
+    if (utf8_check(value)) {
+        error("Found malformed utf8 string for key %s", key);
+        value = "malformed_utf8_string";
+    }
+
+    buffer_sprintf(wb, "%s\"%s\": \"", prefix, key);
+    buffer_strcat_htmlescape(wb, value);
+    buffer_strcat(wb, "\"");
+    buffer_strcat(wb, suffix);
 }
 
-inline void health_alarm_entry2json_nolock(BUFFER *wb, ALARM_ENTRY *ae, RRDHOST *host) {
-    buffer_sprintf(wb,
-            "\n\t{\n"
-                    "\t\t\"hostname\": \"%s\",\n"
-                    "\t\t\"unique_id\": %u,\n"
-                    "\t\t\"alarm_id\": %u,\n"
-                    "\t\t\"alarm_event_id\": %u,\n"
-                    "\t\t\"name\": \"%s\",\n"
-                    "\t\t\"chart\": \"%s\",\n"
-                    "\t\t\"family\": \"%s\",\n"
-                    "\t\t\"processed\": %s,\n"
-                    "\t\t\"updated\": %s,\n"
-                    "\t\t\"exec_run\": %lu,\n"
-                    "\t\t\"exec_failed\": %s,\n"
-                    "\t\t\"exec\": \"%s\",\n"
-                    "\t\t\"recipient\": \"%s\",\n"
-                    "\t\t\"exec_code\": %d,\n"
-                    "\t\t\"source\": \"%s\",\n"
-                    "\t\t\"units\": \"%s\",\n"
-                    "\t\t\"when\": %lu,\n"
-                    "\t\t\"duration\": %lu,\n"
-                    "\t\t\"non_clear_duration\": %lu,\n"
-                    "\t\t\"status\": \"%s\",\n"
-                    "\t\t\"old_status\": \"%s\",\n"
-                    "\t\t\"delay\": %d,\n"
-                    "\t\t\"delay_up_to_timestamp\": %lu,\n"
-                    "\t\t\"updated_by_id\": %u,\n"
-                    "\t\t\"updates_id\": %u,\n"
-                    "\t\t\"value_string\": \"%s\",\n"
-                    "\t\t\"old_value_string\": \"%s\",\n"
-                    "\t\t\"last_repeat\": \"%lu\",\n"
-                    "\t\t\"silenced\": \"%s\",\n"
-                   , host->hostname
-                   , ae->unique_id
-                   , ae->alarm_id
-                   , ae->alarm_event_id
-                   , ae->name
-                   , ae->chart
-                   , ae->family
-                   , (ae->flags & HEALTH_ENTRY_FLAG_PROCESSED)?"true":"false"
-                   , (ae->flags & HEALTH_ENTRY_FLAG_UPDATED)?"true":"false"
-                   , (unsigned long)ae->exec_run_timestamp
-                   , (ae->flags & HEALTH_ENTRY_FLAG_EXEC_FAILED)?"true":"false"
-                   , ae->exec?ae->exec:host->health_default_exec
-                   , ae->recipient?ae->recipient:host->health_default_recipient
-                   , ae->exec_code
-                   , ae->source
-                   , ae->units?ae->units:""
-                   , (unsigned long)ae->when
-                   , (unsigned long)ae->duration
-                   , (unsigned long)ae->non_clear_duration
-                   , rrdcalc_status2string(ae->new_status)
-                   , rrdcalc_status2string(ae->old_status)
-                   , ae->delay
-                   , (unsigned long)ae->delay_up_to_timestamp
-                   , ae->updated_by_id
-                   , ae->updates_id
-                   , ae->new_value_string
-                   , ae->old_value_string
-                   , (unsigned long)ae->last_repeat
-                   , (ae->flags & HEALTH_ENTRY_FLAG_SILENCED)?"true":"false"
-    );
-
-    health_string2json(wb, "\t\t", "info", ae->info?ae->info:"", ",\n");
-
-    if(unlikely(ae->flags & HEALTH_ENTRY_FLAG_NO_CLEAR_NOTIFICATION)) {
-        buffer_strcat(wb, "\t\t\"no_clear_notification\": true,\n");
+static void
+health_kv_str(BUFFER *wb, const char *key, const char *value,
+       const char *prefix, const char *suffix) {
+    if (utf8_check(value)) {
+        error("Found malformed utf8 string for key %s", key);
+        value = "malformed_utf8_string";
     }
 
-    buffer_strcat(wb, "\t\t\"value\":");
+    buffer_sprintf(wb, "%s\"%s\": \"", prefix, key);
+    buffer_strcat(wb, value);
+    buffer_strcat(wb, "\"");
+    buffer_strcat(wb, suffix);
+}
+
+static void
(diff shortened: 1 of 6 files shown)


health/health_json.c
```diff
@@ -1,92 +1,129 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
 #include "health.h"
+#include "libnetdata/libnetdata.h"
+
+#define BOOL_STR(x)         (x) ? "true" : "false";
+#define DEFAULT_STR(x, y)   (x) ? (x) : (y);
+
+static void
+kv_html_escape(BUFFER *wb, const char *key, const char *value,
```
thiagoftsm, line 10: Sorry, I did not understand prefix `kv`, what does it mean in health context?

thiagoftsm: If we are talking about the pair `key` and `value`, would it better to move it for `libnetdata/json` instead to keep a `kv` here inside `health` directory?

vkalintiris: You are right, `kv` stands for `key, value` because these set of functions accept a buffer, key, value, prefix and suffix (I'm open to suggestions if you find this overly succinct).

> would it better to move it for libnetdata/json instead to keep a kv here inside health directory?

I wouldn't mind doing so. I didn't already do so because we expect to replace this code soon with proper validation from `json-c` and we do not intend to have any other external users of these functions.

thiagoftsm: Perfect, so I will give a suggestion you append `health_`, in front of `kv`, because `kv` is generic to be inside health, be free to accept or reject this idea. I called your attention for this, because we do not have any `kv` prefix inside our health code.

> I wouldn't mind doing so. I didn't already do so because we expect to replace this code soon with proper validation from json-c and we do not intend to have any other external users of these functions.

Considering this, let us keep inside health only.

underhood:
why not just consume it with JSON-C and catch potential errors? Should check also another JSON formatting errors

health/health_json.c
```diff
@@ -1,92 +1,129 @@
...
+}
+
+static void
+health_kv_lu(BUFFER *wb, const char *key, long unsigned value,
+      const char *prefix, const char *suffix) {
+    buffer_sprintf(wb, "%s\"%s\": %lu%s", prefix, key, value, suffix);
+}
+
+static void
+health_kv_ld(BUFFER *wb, const char *key, long int value,
```
underhood, line 50: hmmm not sure this is the best way. this is defined in health_json.c but ideally we would have consistency with all the places where we generate JSON... what about chart/charts etc.

underhood (changes requested):

underhood:
checked in configure script and:
```
    # next 2 lines are just to have info for ACLK dependencies in common place
    AC_MSG_CHECKING([if json-c available for ACLK])
    AC_MSG_RESULT([${enable_jsonc}])

    test "${enable_cloud}" = "yes" -a "${enable_jsonc}" = "no" && \
        AC_MSG_ERROR([You have asked for ACLK to be built but no json-c available. ACLK requires json-c])
        ```
        json-c is compulsory for ACLK already (optional for the rest of the agent)

**Comment by underhood** [comment]
will `json_tokener_parse` of JSON-C catch such malformed string? if so we should use it to validate

**Comment by vkalintiris** [comment]
> why not just consume it with JSON-C and catch potential errors? Should check also another JSON formatting errors.
> The way you did it not all will get caught -> not full JSON output validation.

This is the next step but it will take a while to roll out the full JSON-C serialization thing. In the meantime, this PR will allow the backend to unmarshall JSON payloads instead of simply failing entire requests.

**Comment by underhood** [comment]
> > why not just consume it with JSON-C and catch potential errors? Should check also another JSON formatting errors.
> > The way you did it not all will get caught -> not full JSON output validation.
> 
> This is the next step but it will take a while to roll out the full JSON-C serialization thing. In the meantime, this PR will allow the backend to unmarshall JSON payloads instead of simply failing entire requests.

I think there is a misunderstanding here.  In ACLK before sending a message parse it with `json_tokener_parse` when that gives an error we have malformed JSON (confirmed by full JSON library), write an error to console. No custom checking necessary. We just need to validate that JSON-C will catch a malformed UTF-8 string with `json_tokener_parse` (which it should).

This way we no need to modify the JSON generation code at all for now. The only change would be ACLK.

**Comment by vkalintiris** [comment]
> I think there is a misunderstanding here.

Indeed, I think we have a different understanding of the scope of this PR vs. the final/ideal situation we want to end up.

From my perspective, ACLK is irrelevant here because the health code is a standalone module that can have multiple users other than ACLK. Thus, verifying that the serialized JSON is correct at the ACLK step is too late (and not enough for other consumers, unless you repeat whatever checks you end up doing in ACLK at each consumer).

> In ACLK before sending a message to Cloud we can parse it with `json_tokener_parse` when that gives an error we have malformed JSON (confirmed by full JSON library => means we are checking for all kinds of errors, not just this particular one), write an error to log. No custom checking necessary.

This would be too late as well, because once you have a malformed serialized JSON string, all you can do is either accept it or throw it away. With the current PR, you can fulfill each request and use whatever is _usable_ in the JSON response (e.g. you can ignore the `family` field and consume the rest of the payload).

> Also we should validate all ACLK messages like that not just `alarm` msg.

Indeed, but the scope of this PR is focused only on requests that hit the alarm code. We should do this as part of switching to the JSON-C library.

**Comment by underhood** [comment]
> From my perspective, ACLK is irrelevant here because the health code is a standalone module that can have multiple users other than ACLK. Thus, verifying that the serialized JSON is correct at the ACLK step is too late (and not enough for other consumers, unless you repeat whatever checks you end up doing in ACLK at each consumer).

Definitely a different perspective here. The goal I have in mind is agent should never send malformed JSON to Cloud. If malformed JSON is ever generated it should be logged and the developer should look at it ASAP.

> This would be too late as well, because once you have a malformed serialized JSON string, all you can do is either accept it or throw it away. With the current PR, you can fulfill each request and use whatever is usable in the JSON response (e.g. you can ignore the family field and consume the rest of the payload).

I see. Not sure alarm message without `family` field will not trip Cloud. Have you checked with them?

> Indeed, but the scope of this PR is focused only on requests that hit the alarm code. We should do this as part of switching to the JSON-C library.

This is what I do not like here. We have had in past problem with malformed `connect`, `disconnect` now `alarms`. My perspective here is I would like to see a PR that prevents agent to send malformed JSON once and for all.

**Comment by vkalintiris** [comment]
> Definitely a different perspective here. The goal I have in mind is agent should never send malformed JSON to Cloud. If malformed JSON is ever generated it should be logged and the developer should look at it ASAP.

Agreed. In terms of terminology, we need to make the distinction between a malformed JSON string vs a well-formed JSON string that might contain semantically invalid fields. The former is something that can not get parsed. The later is something that can get parsed but it won't necessarily contain useful information for some or all objects.

The intention is to (a) fix the former with this PR, (b) switch to JSON-C, and (c) fix all the different fields that might contain
invalid strings one by one (rinse and repeat once we are done with the health code).
 
> This is what I do not like here. We have had in past problem with malformed `connect`, `disconnect` now `alarms`. My perspective here is I would like to see a PR that prevents agents to send malformed JSON once and for all. In the future the problem will be solved by using JSON-C for JSON generation. I do not personally like the cherry-picking health code for Agent wide issue.
>
> But OK let it be your way (won't block you on this if other reviewers don't have the issue with this as well).
> 
> The functions `health_kv_XYZ` are step in good direction (until we do the proper JSON fix as it is better and more readable than huge printf), but again I do not like cherry picking only health code here when we do same things in another JSON message generators,

Indeed, the intention is to have a single, unified way of serializing JSON stuff in the entire agent and not just for health code (along with having fields that contain correct values in the 1st place). This PR is just the first step that will get us there.

**Comment by underhood** [comment]
> Agreed. In terms of terminology, we need to make the distinction between a malformed JSON string vs a well-formed JSON string that might contain semantically invalid fields. The former is something that can not get parsed. The later is something that can get parsed but it won't necessarily contain useful information for some or all objects.
> 
> The intention is to (a) fix the former with this PR, (b) switch to JSON-C, and (c) fix all the different fields that might contain
> invalid strings one by one (rinse and repeat once we are done with the health code).
> 

Well, `json_tokener_parse` will check for (a), as for (c) this is not something we should check with this PR and is a completely different topic.
The reason I mention missing `family` field is different:
1. Current situation -> cloud rejects message due to malformed JSON
2. Your PR -> will send it without the `family` field. My question is does this improve things compared to just not sending the message at all? Did you check with the Cloud team what will they do with such a message? Maybe the situation will be the same -> Error on cloud side and rejection of the message. **Therefore situation before and after this PR can possibly be the same.**

**Comment by underhood** [comment]
@vkalintiris what I worry about here is skipping "random" fields and sending such messages to the cloud can have unforeseen consequences the most benign of which is the cloud will throw the message away with error (which means this PR doesn't change anything).

**Comment by mfundul** [comment]
I have seen corrupted health log files, @underhood , so that means you cannot avoid throwing away messages. A corrupted health log file cannot be repaired by fixing the software.

**Comment by underhood** [comment]
> I have seen corrupted health log files, @underhood , so that means you cannot avoid throwing away messages. A corrupted health log file cannot be repaired by fixing the software.

yep my point is maybe it is better to throw away that message instead of replacing random field with `malformed_utf8_string` and sending (what if alarm name gets replaced or any other field.

**Comment by mfundul** [comment]
In this case though, it's problematic to throw away the message, because it happens during alarm entry removal. If we throw away the message, then the cloud side will never remove the alarms and it will fill up. In this particular scenario, it would be better if the cloud removed the alarm even if the "family" field is corrupted, as long as every other field is valid, which they happen to be.

**Comment by underhood** [comment]
> In this case though, it's problematic to throw away the message, because it happens during alarm entry removal. If we throw away the message, then the cloud side will never remove the alarms and it will fill up. In this particular scenario, it would be better if the cloud removed the alarm even if the "family" field is corrupted, as long as every other field is valid, which they happen to be.

should we make it specific for `family` then? (going more quick and dirty)

**Comment by stelfrag** [comment]
I would like to have a simpler approach than trying to verify everything we send (it's not **that** broken) :smiley: That being said, setting up a mechanism to validate all outbound json payloads should be planned and implemented.

The less intrusive way is I believe to try and handle the corrupted field which appears to be in the health-log.db file. This check can be *added to the code that "loads" the file*  and either

- "Fix" the string using "unknown" or something like that
- Ignore the entry
   - If the alarm entry was WARN/CRIT and not superseded it may cause incorrect alarm state in the cloud 
     - This may be minor compared to sending a json payload that is invalid

Keep in mind that

- The agents that have this bug (versions 1.27.0 - v1.28.0-110) may continue to send malformed json payloads
   - Agents after v.1.28-0-110 may continue to briefly send malformed payloads until the heath log files rotate
- Agents that will apply **this** fix will not generate any more corrupted entries in the health-log.db

**Comment by mfundul** [comment]
The actual range of the problematic agent versions is `v1.26.0-382-g0f8175dd` - `v1.28.0-122-g7f002c12`

**Comment by underhood** [comment]
@stelfrag agreed

**Comment by underhood** [comment]
I did some checks and the problem is bigger in scope. I made collector where chart name is "somethig\x00somethingelse" and that is passed to cloud and breaks metric correlations on that node, the chart itself is broken as well although present and who know what else

**Comment by papazach** [comment]
Hello guys 👋 , just wanted to jump in the conversation since we saw those `correlation` operation errors in the cloud and I have some extra clues to share.

When the user asks for metric correlations for a time-window, from the cloud side we do the following:

- Perform a `charts` request to the Agent to fetch all the chartIDs
- Then for each chartID we request data for two different time windows (the baseline and the highlight of the original time window)
- We feed the previous result to Andy's ML service to crunch the numbers and produce results.

So given the above the error message that we got was `ruby.collector: invalid character 'x' in string escape code`

I dug through and found the chart in our Mongo view:

```json
{
    "_id" : "6024232da395243b786aa30c",
    "id" : "ruby.collector",
    "nodeID" : "7366bd1b-3aef-4d42-bfd1-098f26f6817f",
    "charttype" : "line",
    "context" : "ruby",
    "dimensions" : [ 
        {
            "id" : "whateva",
            "name" : "whateva"
        }, 
        {
            "id" : "t\\x00est_1",
            "name" : "t\\x00est_1"
        }, 
        {
            "id" : "test_2",
            "name" : "test_2"
        }, 
        {
            "id" : "test_3",
            "name" : "test_3"
        }, 
        {
            "id" : "test_4",
            "name" : "test_4"
        }, 
        {
            "id" : "test_5",
            "name" : "test_5"
        }, 
        {
            "id" : "something_else",
            "name" : "something_else"
        }
    ],
    "family" : "r__________ro______uby",
    "module" : "",
    "name" : "ruby.testchart",
    "plugin" : "ruby.plugin",
    "priority" : 1000,
    "title" : "The JSON hard drive is down, generate the open-source bus so we can hack the AI application! (ruby.testchart)",
    "type" : "ruby",
    "units" : "pixel",
    "updateevery" : 1
}
```

It seems that we do not handle gracefully from the cloud side the unmarshalling error upon trying to unmarshal the second dimension with id `t\\x00est_1`.

We' re still investigating but the errors point to that.

underhood:
@papazach wow kudos for noticing that in the cloud... i mentioned this to @andrewm4894 only in the chat as I didn't know why one of my nodes doesn't work with correlations... then I remember I left the purposefully evil collector enabled...

Good to fix on cloud side as well

underhood:
@papazach and you are officially the winner of `suddenly a wild magician appeared out of nowhere` prize gold medal that includes 5 beers on the next meetup voucher.

PR #10648, opened 2021-02-16
Changed Docker image tagging to use semver tags for releases.

Ferroin:
##### Summary

This allows users to use image names like `netdata/netdata:1` or `netdata/netdata@1.29` and track the most up-to-date release that matches that version prefix.

Such usage is a common practice for projects using semantic versioning like we are.

##### Component Name

area/ci
area/packaging

##### Test Plan

The tag generations cript can be tested locally, it takes a single argument which is a version string as we normally use them (so `vX.Y.Z`) and outputs a list of version tags to push.

##### Additional Information

Short term, the `1` tag will be functionally identical to the `stable` tag. Long-term, we may eventually end up with a new major version of Netdata, at which point having the `1` tag to let users stick with the old version is important during the transition period.

### DIFF
--- a/packaging/docker/README.md
+++ b/packaging/docker/README.md
@@ -94,6 +94,22 @@ volumes:
   netdatacache:
 ```
 
+## Docker tags
+
+The official `netdata/netdata` Docker image provides the following named tags:
+
+* `stable`: The `stable` tag will always point to the most recently published stable build.
+* `edge`: The `edge` tag will always point ot the most recently published nightly build. In most cases, this is
+  updated daily at around 01:00 UTC.
+* `latest`: The `latest` tag will always point to the most recently published build, whether it’s a stable build
+  or a nightly build. THis is what Docker will use by default if you do not specify a tag.
+
+Additionally, for each stable release, three tags are pushed, one with the full version of the release (for example,
+`v1.30.0`), one with just the major and minor version (for example, `v1.30`), and one with just the major version
+(for example, `v1`). The tags for the minor versions and major versions are updated whenever a release is published
+that would match that tag (for example, if `v1.30.1` were to be published, the `v1.30` tag would be updated to
+point to that instead of `v1.30.0`).
+
 ## Health Checks
 
 Our Docker image provides integrated support for health checks through the standard Docker interfaces.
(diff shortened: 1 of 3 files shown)


knatsakis (approved):

vkalintiris (approved):

Ferroin:
I am intentionally holding off merging this until we know that we will not have another v1.29.x patch release. It should only go in for v1.30.0 because it introduces a breaking change in our handling of Docker image version tags.

Ferroin:
I just realized we can do this in a backwards-compatible way, and that doing so would actually make us more consistent with how other projects are actually doing things (they usually do in fact have a leading `v` in their version tags).

Given this, if I could get this re-approved, it can be merged now instead of having to wait for an indeterminate amount of time.

knatsakis (dismissed):

packaging/docker/README.md
```diff
@@ -94,6 +94,22 @@ volumes:
...
 
+## Docker tags
+
+The official `netdata/netdata` Docker image provides the following named tags:
+
+* `stable`: The `stable` tag will always point to the most recently published stable build.
+* `edge`: The `edge` tag will always point ot the most recently published nightly build. In most cases, this is
+  updated daily at around 01:00 UTC.
+* `latest`: The `latest` tag will always point to the most recently published build, whether it’s a stable build
+  or a nightly build. THis is what Docker will use by default if you do not specify a tag.
```
joelhans, line 105:
```suggestion
  or a nightly build. This is what Docker will use by default if you do not specify a tag.
```

joelhans (commented):
This looks good. You think most users who care about these tag options will know how to use them, or should we add a sentence or two, or a link, explaining to add the tag to the end of `netdata/netdata`?

Ferroin:
> This looks good. You think most users who care about these tag options will know how to use them, or should we add a sentence or two, or a link, explaining to add the tag to the end of `netdata/netdata`?

In general, if a user does not know how to select specific tags for a Docker image, they should not be using Docker, so I’d say it’s beyond the scope of _our_ documentation to try and help them there.

I also think it may come off as a bit patronizing to people who do know if we go through explaining how to use Docker tags.

OdysLam (approved):

joelhans:
> In general, if a user does not know how to select specific tags for a Docker image, they should not be using Docker,

:laughing: Fair enough!

joelhans (approved):
Docs LGTM.

This should still get a review/approval from >1 engineer.

kaskavel (approved):

PR #10652, opened 2021-02-17
Fix the context filtering on the data query endpoint

stelfrag:
##### Summary
Adjust the filtering on data endpoint when a context parameter is used. For a requested time range (t1, t2) filter out the matching charts when their  `last_entry_t < t1` is true

##### Component Name
database

### DIFF
--- a/web/api/formatters/rrd2json.c
+++ b/web/api/formatters/rrd2json.c
@@ -32,6 +32,30 @@ void free_context_param_list(struct context_param **param_list)
     *param_list = NULL;
 }
 
+void rebuild_context_param_list(struct context_param *context_param_list, time_t after_requested)
+{
+    RRDDIM *temp_rd = context_param_list->rd;
+    RRDDIM *new_rd_list = NULL, *t;
+    while (temp_rd) {
+        t = temp_rd->next;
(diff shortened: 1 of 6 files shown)


web/api/formatters/rrd2json.c
```diff
@@ -32,6 +32,30 @@ void free_context_param_list(struct context_param **param_list)
...
+        } else {
+            freez((char *)temp_rd->id);
+            freez((char *)temp_rd->name);
+#ifdef ENABLE_DBENGINE
+            if (temp_rd->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE)
+                freez(temp_rd->state->metric_uuid);
+#endif
+            freez(temp_rd->state);
+            freez(temp_rd);
+        }
```
vkalintiris, line 53: Pretty similar to what `free_temp_rrddim`. It'd be nice to post a follow up patch that would allow us to free temp RRDDIMs from one place.

vkalintiris (approved):

web/api/formatters/rrd2json.c
stelfrag: Correct this should be handled differently. In fact there is more internal booking that needs to be in place in a later fix

MrZammler (approved):

underhood (approved):

PR #10672, opened 2021-02-22
add python changefinder collector

andrewm4894:
##### Summary

- adds python 'changefinder' based collector for online changepoint detection.

##### Component Name

collectors/python.d/changefinder

##### Test Plan

Tested locally on development nodes running for 24-48 hours to find useful examples of changefinder finding actually useful changes and validate runtime performance characteristics. 

##### Additional Information

Much more detail in the README of the collector.


andrewm4894:
adding image here to link to in readme

image

collectors/python.d.plugin/changefinder/README.md
```diff
@@ -0,0 +1,154 @@
...
+- **ChangeFinder Flags** (`changefinder.flags`): This chart shows `1` or `0` if the latest score has a percentile value that exceeds the `cf_threshold` threshold. By default, any scores that are in the 99th or above percentile will raise a flag on this chart. The raw changefinder score itself can be a little noisey and so limiting ourselves to just periods where it surpasses the 99th percentile can help manage the "[signal to noise ratio](https://en.wikipedia.org/wiki/Signal-to-noise_ratio)" better. The `cf_threshold` paramater might be one you want to play around with to tune things a little more specifically for the workloads on your node and the specific charts you want to monitor. For example, maybe the 95th percentile might work better for you than the 99th percentile (likley at the cost of a little more noise on the chart over time).
```
ilyam8, line 20: That line is 852 characters. I suggest to apply markdown auto-format to the readme, there are a lot of problems.

collectors/python.d.plugin/changefinder/changefinder.chart.py
```diff
@@ -0,0 +1,175 @@
...
+from json import loads
+import re
+
+from bases.FrameworkServices.UrlService import UrlService
+
+import numpy as np
+import changefinder
+from scipy.stats import percentileofscore
+
+update_every = 1
```
ilyam8, line 15: That collector does quite heavy query - `/api/v1/allmetrics?format=json` - are you sure 1 is ok as default?

collectors/python.d.plugin/changefinder/changefinder.chart.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if self.show_scores:
+            data_score['average_score'] = sum(data_score.values()) / len(data_score)
+            self.update_chart('scores', data_score, divisor=100)
+
+        self.update_chart('flags', data_flag)
+
+        data = {**data_score, **data_flag}
+
+        return data
```
ilyam8, line 175: newline

collectors/python.d.plugin/changefinder/changefinder.chart.py
```diff
@@ -0,0 +1,175 @@
...
+        return score, flag
+
+    def update_chart(self, chart, data, algo='absolute', multiplier=1, divisor=1):
+        """Add dimensions found in data to chart if needed.
+        """
+        if not self.charts:
+            return
+
+        for dim in data:
+            if dim not in self.charts[chart]:
```
ilyam8, line 108: As you know charts dimension is a list, so if there is a lot of checks and we do them on every data collection we need some cache (`set`)

ilyam8:
@andrewm4894 

I have a feeling that every collector you add does `api/v1/allmetrics?format=json` query, which is quite heavy.

Is there a chance that it could one omega ML thing with different components (changefinder, etc.), so we do only one query? 🤔 Are there any other problems with that approach?

collectors/python.d.plugin/changefinder/changefinder.chart.py
andrewm4894: bumped it to 5 from 1.

andrewm4894: 1 would be best in terms of performance of the algo (in sense of detecting changes) but 5 should be fine. Better to have a safer default and people can reduce it to 1 then if they want.

collectors/python.d.plugin/changefinder/changefinder.chart.py
andrewm4894: Added a `self.collected_dims` which is a set for each chart and gets maintained by `update_chart()`. I think that is what we want here - similar to how it's done in the alarms collector.

andrewm4894:
> I have a feeling that every collector you add does api/v1/allmetrics?format=json query, which is quite heavy.

My understanding from talking to @mfundul a while back was the /allmetrics does not quite go all the way to the database since its sort of cached somewhere and so a lot cheaper than doing a /data type call for each chart.

But yes - getting the data into python is probably the biggest challenge with all of this. Interesting enough i think having an efficient and portable/flexible sort of storage that would give more options is part of the reasoning behind [influx iox](https://www.influxdata.com/blog/announcing-influxdb-iox/) would be cool if there was a way for me to just go and grab the data from arrow and not have to go near the rest api and query engine. Or some sort of python package that could interface more efficiently with the raw data. 

I was also thinking of some sort of ml redis queue or something where 'hot' data (in terms of likely to be needed by some ml stuff) could be sent and gotten at quickly if/as needed. But again i'm kind of at the edge of my comfort zone in if such ideas make sense or not. 

For now trying to squeeze as much as possible out of the REST API as its all i can do.

collectors/python.d.plugin/changefinder/README.md
andrewm4894: i see what you mean, have made the bullets into sub headers and broken it into a couple of paragraphs instead.

andrewm4894:
@ilyam8 @thiagoftsm hey - i was hoping to try get this reviewed this week or next if possible.

collectors/python.d.plugin/changefinder/changefinder.conf
```diff
@@ -0,0 +1,75 @@
...
+  # The percentile above which scores will be flagged.
+  cf_threshold: 99
+
+  # The number of recent scores to use when calculating the percentile of the changefinder score.
+  n_score_samples: 14400
+
+  # Set to true if you also want to chart the percentile scores in addition to the flags.
+  # Mainly useful for debugging or if you want to dive deeper on how the scores are evolving over time.
+  show_scores: false
+
```
ilyam8, line 75:
```suggestion

```

collectors/python.d.plugin/changefinder/README.md
```diff
@@ -0,0 +1,218 @@
...
+    - About ~85mb of ram (`apps.mem`) being continually used by the `python.d.plugin` under default configuration.
+
+## Useful links and further reading
+
+- [PyPi changefinder](https://pypi.org/project/changefinder/) reference page.
+- [GitHub repo](https://github.com/shunsukeaihara/changefinder) for the changefinder library.
+- Relevant academic papers:
+    - Yamanishi K, Takeuchi J. A unifying framework for detecting outliers and change points from nonstationary time
+      series data. 8th ACM SIGKDD international conference on Knowledge discovery and data mining - KDD ’02. 2002:
+      676. ([pdf](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.12.3469&rep=rep1&type=pdf))
```
ilyam8, line 186: something is wrong here, I see there is open `’`, but no close

andrewm4894: maybe a rogue copy paste from the paper i got the reference from - will clean.

collectors/python.d.plugin/changefinder/changefinder.chart.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # process each chart
+        for chart in charts_in_scope:
+
+            if self.mode == 'per_chart':
+
+                # average dims on chart and run changefinder on that average
+                x = [raw_data[chart]['dimensions'][dim]['value'] for dim in raw_data[chart]['dimensions']]
+                x = [x for x in x if x is not None]
+                x = sum(x) / len(x)
```
ilyam8, line 147: Could `len(x)` be a zero? If so do we need to handle it?

collectors/python.d.plugin/changefinder/changefinder.chart.py
```diff
@@ -0,0 +1,179 @@
...
+                    x = x if x else 0
+                    x = self.diff(x, chart_dim) if self.cf_diff else x
+
+                    score, flag = self.get_score(x, chart_dim)
+                    if self.show_scores:
+                        data_score['{}_score'.format(chart_dim)] = score * 100
+                    data_flag[chart_dim] = flag
+
+        if self.show_scores:
+            data_score['average_score'] = sum(data_score.values()) / len(data_score)
```
ilyam8, line 172: Could `len(data_score)` be a zero? If so do we need to handle it?

collectors/python.d.plugin/changefinder/changefinder.chart.py
andrewm4894: very unlikely but maybe technically possible if netdata rest api returns `null` for all dims on a chart which i guess maybe could technically happen (e.g. `apps.cpu` can often have some `null` although have never see all null) so worth a little check to handle - i have added it.

collectors/python.d.plugin/changefinder/changefinder.chart.py
andrewm4894: have added a condition just in case

ilyam8 (approved):

collectors/python.d.plugin/changefinder/README.md
```diff
@@ -0,0 +1,218 @@
...
+/usr/libexec/netdata/plugins.d/python.d.plugin changefinder debug trace nolock
+```
+
+## Notes
+
+- It may take an hour or two (depending on your choice of `n_score_samples`) for the collector to 'settle' into it's
+  typical behaviour in terms of the trained models and scores you will see in the normal running of your node. Mainly
+  this is because it can take a while to build up a proper distribution of previous scores in over to convert the raw
+  score returned by the ChangeFinder algorithim into a percentile based on the most recent `n_score_samples` that have
+  already been produced. So when you first turn the collector on, it will have a lot of flags in the begining and then
```
vlvkobal, line 168:
```suggestion
  already been produced. So when you first turn the collector on, it will have a lot of flags in the beginning and then
```

collectors/python.d.plugin/changefinder/README.md
```diff
@@ -0,0 +1,218 @@
...
+- [GitHub repo](https://github.com/shunsukeaihara/changefinder) for the changefinder library.
+- Relevant academic papers:
+    - Yamanishi K, Takeuchi J. A unifying framework for detecting outliers and change points from nonstationary time
+      series data. 8th ACM SIGKDD international conference on Knowledge discovery and data mining - KDD02. 2002:
+      676. ([pdf](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.12.3469&rep=rep1&type=pdf))
+    - Kawahara Y, Sugiyama M. Sequential Change-Point Detection Based on Direct Density-Ratio Estimation. SIAM
+      International Conference on Data Mining. 2009:
+      389–400. ([pdf](https://onlinelibrary.wiley.com/doi/epdf/10.1002/sam.10124))
+    - Liu S, Yamada M, Collier N, Sugiyama M. Change-point detection in time-series data by relative densityratio
+      estimation. Neural Netw. Jul.2013 43:72–83. [PubMed: 23500502] ([pdf](https://arxiv.org/pdf/1203.0453.pdf))
```
vlvkobal, line 191:
```suggestion
    - Liu S, Yamada M, Collier N, Sugiyama M. Change-point detection in time-series data by relative density-ratio
      estimation. Neural Networks. Jul.2013 43:72–83. [PubMed: 23500502] ([pdf](https://arxiv.org/pdf/1203.0453.pdf))
```

vlvkobal (dismissed):

ilyam8:
Approving under the assumption that

> Tested locally on development nodes running

was done with latest changes

ilyam8 (dismissed):

andrewm4894:
> Approving under the assumption that
> 
> > Tested locally on development nodes running
> 
> was done with latest changes

going to spin up a vm with latest changes and run that for a few hours to be sure.

andrewm4894:
i was missing a small check in `validate_charts()` that meant the initial check on the collector was failing. fixed now so is similar to what we do in the alarms collector.

andrewm4894:
just retesting - will re-request review once ready.

andrewm4894:
working now :)

ilyam8 (approved):

vlvkobal (approved):

PR #10675, opened 2021-02-23
Alarms collector add alarm values

andrewm4894:
##### Summary

Add option `show_alarm_values` to `alarms.conf` which will show the alarm values themselves over time below the alarm status chart. 

`show_alarm_values` will be set to `false` by default such that a user would have to explicitly enable this in `alarms.conf` if they want the "Values" chart.

For example: 
image

##### Component Name

/collectors/python.d.plugin/alarms

##### Test Plan

Tested locally for when both show_alarm_values was `true` and `false` in situations where i used `stress-ng --all 2` to trigger lots of alarms.

### DIFF
--- a/collectors/python.d.plugin/alarms/alarms.chart.py
+++ b/collectors/python.d.plugin/alarms/alarms.chart.py
@@ -29,9 +29,6 @@ def charts_template(sm, alarm_status_chart_type='line'):
         'values': {
             'options': [None, 'Alarm Values', 'value', 'value', 'alarms.value', 'line'],
             'lines': [],
-            'variables': [
-                [],
-            ]
         }
     }
     return order, charts
--- a/collectors/python.d.plugin/alarms/alarms.chart.py
+++ b/collectors/python.d.plugin/alarms/alarms.chart.py
@@ -64,7 +64,7 @@ class Service(UrlService):
         data['alarms_num'] = len(data)
 
         if self.show_alarm_values:
-            data_values = {a: alarms[a]['value'] * 100 for a in alarms if 'value' in alarms[a] and alarms[a]['value'] is not None}
+            data_values = {'{}_value'.format(a): alarms[a]['value'] * 100 for a in alarms if 'value' in alarms[a] and alarms[a]['value'] is not None}
             self.update_charts('values', alarms, data_values, 1, 100)
             data = {**data, **data_values}
 
--- a/collectors/python.d.plugin/alarms/alarms.chart.py
+++ b/collectors/python.d.plugin/alarms/alarms.chart.py
@@ -66,7 +66,7 @@ class Service(UrlService):
         if self.collect_alarm_values:
             data_values = {'{}_value'.format(a): alarms[a]['value'] * 100 for a in alarms if 'value' in alarms[a] and alarms[a]['value'] is not None}
             self.update_charts('values', data_values, divisor=100)
-            data = {**data, **data_values}
+            data.update(data_values)
 
         return data
 
--- a/collectors/python.d.plugin/alarms/alarms.chart.py
+++ b/collectors/python.d.plugin/alarms/alarms.chart.py
@@ -63,7 +63,7 @@ class Service(UrlService):
         self.update_charts('alarms', data)
         data['alarms_num'] = len(data)
 
-        if self.show_alarm_values:
+        if self.collect_alarm_values:
             data_values = {'{}_value'.format(a): alarms[a]['value'] * 100 for a in alarms if 'value' in alarms[a] and alarms[a]['value'] is not None}
             self.update_charts('values', data_values, divisor=100)
             data = {**data, **data_values}
--- a/collectors/python.d.plugin/alarms/alarms.chart.py
+++ b/collectors/python.d.plugin/alarms/alarms.chart.py
@@ -64,7 +64,7 @@ class Service(UrlService):
 
         if self.show_alarm_values:
             data_values = {'{}_value'.format(a): alarms[a]['value'] * 100 for a in alarms if 'value' in alarms[a] and alarms[a]['value'] is not None}
-            self.update_charts('values', alarms, data_values, multiplier=1, divisor=100)
+            self.update_charts('values', alarms, data_values, divisor=100)
             data = {**data, **data_values}
 
         return data
--- a/collectors/python.d.plugin/alarms/alarms.chart.py
+++ b/collectors/python.d.plugin/alarms/alarms.chart.py
@@ -20,14 +20,14 @@ def charts_template(sm):
     mappings = ', '.join(['{0}={1}'.format(k, v) for k, v in sm.items()])
     charts = {
         'alarms': {
-            'options': [None, 'Alarms ({0})'.format(mappings), 'status', 'alarms', 'alarms.status', 'line'],
+            'options': [None, 'Alarms ({0})'.format(mappings), 'status', 'Alarms', 'alarms.status', 'line'],
             'lines': [],
             'variables': [
                 ['alarms_num'],
             ]
         },
         'values': {
-            'options': [None, 'Alarm Values', 'value', 'values', 'alarms.values', 'line'],
+            'options': [None, 'Alarm Values', 'value', 'Values', 'alarms.values', 'line'],
             'lines': [],
             'variables': [
                 [],
@@ -48,7 +48,6 @@ class Service(UrlService):
         self.sm = self.configuration.get('status_map', DEFAULT_STATUS_MAP)
         self.order, self.definitions = charts_template(self.sm)
         self.url = self.configuration.get('url', DEFAULT_URL)
-        self.collected_alarms = set()
         self.show_alarm_values = bool(self.configuration.get('show_alarm_values', DEFAULT_SHOW_ALARM_VALUES))
 
     def _get_data(self):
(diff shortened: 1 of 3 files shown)


collectors/python.d.plugin/alarms/alarms.chart.py
```diff
@@ -14,24 +14,32 @@
...
     order = [
         'alarms',
+        'values'
     ]
 
     mappings = ', '.join(['{0}={1}'.format(k, v) for k, v in sm.items()])
     charts = {
         'alarms': {
-            'options': [None, 'Alarms ({0})'.format(mappings), 'status', 'alarms', 'alarms.status', 'line'],
+            'options': [None, 'Alarms ({0})'.format(mappings), 'status', 'Alarms', 'alarms.status', 'line'],
```
vlvkobal, line 23: Why did you capitalize the family?

collectors/python.d.plugin/alarms/alarms.chart.py
```diff
@@ -51,21 +59,20 @@ def _get_data(self):
...
 
-        for a in data:
-            if a not in self.collected_alarms:
-                self.collected_alarms.add(a)
-                self.charts['alarms'].add_dimension([a, a, 'absolute', '1', '1'])
-
-        for a in list(self.collected_alarms):
-            if a not in alarms:
-                self.collected_alarms.remove(a)
-                self.charts['alarms'].del_dimension(a, hide=False)
```
vlvkobal, line 71: Don't you need to delete dimensions in the `alarms` chart anymore?

collectors/python.d.plugin/alarms/alarms.chart.py
andrewm4894: oh yep - that's not like how we do it for others - yep - will revert.

collectors/python.d.plugin/alarms/alarms.chart.py
andrewm4894: I don't think so - if it just disappears from the `data` from the collector then i think it will disappear from the chart. The trouble i had was trying to manage the bookkeeping for `self.collected_alarms` in the situation where we are charting both `alarms` and `values` chart. It was confusing for me to consider all the combinations that could happen if we had a status for one alarm and not a value or vice versa. In such a case, removing the alarm from `self.collected_alarms` was breaking the other chart (and vice versa). 

So i was trying to just avoid having to track and manage `self.collected_alarms` as was not quite sure why we need it. 

I think by default such dimensions will be hidden. 

Or maybe i could just use both a `self.collected_alarms` and `self.collected_values` and then bring back this code and do the bookkeeping based on if trying to update the alarms or the values chart - kinda separate it out.

ilyam8: > as was not quite sure why we need it.

We need it because `charts.dimensions` is a `list` and it is O(n) to check `if dim in charts['id']`. If there are a lot of dimensions it is slow. So there are 2 ways:

 - refactor charts dimensions attr (make it ordered set or somthing)
 - use local cache for checks (self.collected_alarms)

---


> The trouble i had was trying to manage the bookkeeping for self.collected_alarms in the situation where we are charting both alarms and values chart.

Just add a second one

 - self.collected_status_alarms
 - self.collected_values_alarms

ilyam8: > I think by default such dimensions will be hidden.

No, there is auto-cleanup for (stale/not updated) charts and there is no such things for dimensions. Have to be removed manually via `self.charts['id'].dim_remove`

ilyam8: @andrewm4894 ofc i would prefer _refactor charts dimensions_, but let's keep using local cache until we have done it

andrewm4894: Yep, agree just adding second one is the best option here, will do

collectors/python.d.plugin/alarms/alarms.chart.py
```diff
@@ -40,7 +48,7 @@ def __init__(self, configuration=None, name=None):
         self.sm = self.configuration.get('status_map', DEFAULT_STATUS_MAP)
         self.order, self.definitions = charts_template(self.sm)
         self.url = self.configuration.get('url', DEFAULT_URL)
-        self.collected_alarms = set()
+        self.show_alarm_values = bool(self.configuration.get('show_alarm_values', DEFAULT_SHOW_ALARM_VALUES))
```
ilyam8, line 51:
```suggestion
        self.сollect_alarm_values = bool(self.configuration.get('collect_alarm_values', DEFAULT_SHOW_ALARM_VALUES))
```

I think `collect_` is better because it literally reflects the meaning, show is a little bit wrong. What do you think?

andrewm4894: agree, done

collectors/python.d.plugin/alarms/alarms.chart.py
andrewm4894: i have added dict of sets called `self.collected_dims` to keep track of the set of collected dims for each chart and maintain it in the same way as the current alarms collector does.

andrewm4894:
ok i think this is ready for another look whenever you get a chance. 

p.s As part of dogfooding this on our parent node @cakrit mentioned maybe being able to config the status charts to be stacked if the user likes. I think it's not a bad idea - we could easily add a `alarm_status_chart_type` config that defaults to `line` but can be changed to `stacked` if the user wants that.

I know stacking 1's and 2's which represent different status might sound a bit funny but i think it could be useful - if you see a "tower" of 2's you know you in trouble very obviously, whereas if you see some stacked 1's it might not be as important but still worth investigating etc. 

Any objections to adding such an option in this PR?

collectors/python.d.plugin/alarms/alarms.chart.py
```diff
@@ -51,21 +60,26 @@ def _get_data(self):
...
 
         data = {a: self.sm[alarms[a]['status']] for a in alarms if alarms[a]['status'] in self.sm}
-        self.update_charts(alarms, data)
+        self.update_charts('alarms', data)
         data['alarms_num'] = len(data)
 
+        if self.collect_alarm_values:
+            data_values = {'{}_value'.format(a): alarms[a]['value'] * 100 for a in alarms if 'value' in alarms[a] and alarms[a]['value'] is not None}
+            self.update_charts('values', data_values, divisor=100)
+            data = {**data, **data_values}
```
ilyam8, line 69:
```suggestion
            data.update(data_values)
```

py2 is dead, but it is still deafult on some old distros. How about to use dict.update?

andrewm4894:
Here is an example of how it looks when stacked with some alarms of different status over time. I actually think this is more useful and how i am planning to set it up on our parent node for dogfooding it once possible. 

image

ilyam8:
> when stacked

Why do you think it is useful? What do you try to achieve?

See y-axis, it looks strange for me 🤷‍♂️ 

Screenshot 2021-03-02 at 21 21 31

andrewm4894:
> Why do you think it is useful? What do you try to achieve?

a quick scroll down one of these for each of the nodes streaming to my parent really quickly gives me a feel for how many alarms on each. If i see a big tower of 1's and 2's i know i have a lot of either warning or critical alarms. From the width of each i can even get a feel for if its lots of 1's, lots of 2's or a mix.

ilyam8:
> e.g. in this real example there are actually 3 critical alarms but you can't tell that from looking at it as the 3 lines are on top of each other
> So not being able to distinguish between a period where you have 1 critical alarm vs 3 is not great.

Then it should be `numbers_of_alarms_in_state` metric/chart. The chart, you are referring to, doesn't supposed to give that information.

> My own personal preference would be for stacked in this case.

My problem with it that it makes `y-axis` senseless.

ilyam8:
And there are 2 cases:
 - you need `number_of_alarms_in_state` as a metric (doesn't sound useful)
 - you want to see the number of alarms in a state

If the latter then we need to request to add those number to the Agent dashboard. Like we see them on the Cloud dashboard

Screenshot 2021-03-02 at 23 48 45

andrewm4894:
> My problem with it that it makes y-axis senseless.

I don't think it does, you can think of it as a sum of scores based on the state of the alarm. You get a score of 1 for warning alarm and a score of 2 for critical alarm. Summing all this up is a score/measure of how messed up the system is over time. 

It might be that we could call it something better maybe. Eg if `alarm_status_chart_type='stacked'` then maybe "status sum" is better than just "status" when its a stacked chart.

ilyam8:
Well, i can only repeat that **sum of alarm statuses codes** (int values are configurable by a user) has no sense for me.

> Summing all this up is a score/measure of how messed up the system is over time.

As i see it you need a number of [alarms in a state](https://github.com/netdata/netdata/pull/10675#issuecomment-789206938), i dont get why you are trying to achieve it not using the most (imo) straightforward way.

ilyam8:
Key part is **for me**, was just sharing my opinion on the subject. Ofc it doesn't mean i want you to remove the option (ideally it (representation) should be configurable via gui).

collectors/python.d.plugin/alarms/alarms.chart.py
```diff
@@ -11,36 +11,47 @@
...
     order = [
         'alarms',
+        'values'
     ]
 
     mappings = ', '.join(['{0}={1}'.format(k, v) for k, v in sm.items()])
     charts = {
         'alarms': {
-            'options': [None, 'Alarms ({0})'.format(mappings), 'status', 'alarms', 'alarms.status', 'line'],
+            'options': [None, 'Alarms ({0})'.format(mappings), 'status', 'alarms', 'alarms.status', alarm_status_chart_type],
```
ilyam8, line 23:
```suggestion
            'options': [None, 'Alarms ({0})'.format(mappings), 'status', 'status', 'alarms.status', alarm_status_chart_type],
```

`status` as the family looks better, do you agree?

Screenshot 2021-03-04 at 13 59 26

collectors/python.d.plugin/alarms/alarms.chart.py
```diff
@@ -11,36 +11,47 @@
...
             'lines': [],
             'variables': [
                 ['alarms_num'],
             ]
+        },
+        'values': {
+            'options': [None, 'Alarm Values', 'value', 'values', 'alarms.values', 'line'],
```
ilyam8, line 30:
```suggestion
            'options': [None, 'Alarm Values', 'value', 'value', 'alarms.values', 'line'],
```

Singular seems better to me. What do you think?

collectors/python.d.plugin/alarms/alarms.chart.py
```diff
@@ -11,36 +11,47 @@
...
+            'lines': [],
+            'variables': [
+                [],
+            ]
```
ilyam8, line 34:
```suggestion
```

variables not needed here

ilyam8:
@andrewm4894 i have only few minor suggestion, overall looks good and ready to merge.

collectors/python.d.plugin/alarms/alarms.chart.py
ilyam8: if plural is better, then we need to make both families plural: statuses, values.

ilyam8 (approved):

thiagoftsm (approved):

PR #10676, opened 2021-02-23
Add new cookie to fix 8094

### LINKED ISSUE #8094: Registry person GUID cookie will eventually stop working in Chrome
Chrome displays the following, for cookies set by the current netdata.cloud and by the registry (my-netdata.io).:

A cookie associated with a cross-site resource at http://registry.my-netdata.io/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.

We'll need to get it fixed.

thiagoftsm:
##### Summary
Fixes #8094

This PR fixes the problem to access cloud with Netdata.

##### Component Name
Webserver

##### Test Plan

1 - Access Netdata dashboard and verify that you do not have the error `Some cookies are misusing the recommended “sameSite“ attribute 11` using browser developer tools

### DIFF
--- a/registry/README.md
+++ b/registry/README.md
@@ -176,6 +176,17 @@ There can be up to 2 files:
 
 Both files are machine readable text files.
 
+### How can I disable the SameSite and Secure cookies?
+
+Beginning with `v1.30.0`, when the Netdata Agent's web server processes a request, it delivers the `SameSite=none`
+and `Secure` cookies. If you have problems accessing the local Agent dashboard or Netdata Cloud, disable these
+cookies by [editing `netdata.conf`](/docs/configure/nodes.md#use-edit-config-to-edit-configuration-files):
+
+```conf
+[registry]
+    enable cookies SameSite and Secure = no
+```
+
 ## The future
 
 The registry opens a whole world of new possibilities for Netdata. Check here what we think:
--- a/registry/registry.c
+++ b/registry/registry.c
@@ -32,10 +32,12 @@ static void registry_set_cookie(struct web_client *w, const char *guid) {
 
     if(registry.registry_domain && registry.registry_domain[0])
         snprintfz(domain, 511, "Domain=%s", registry.registry_domain);
+    else
+        domain[0]='\0';
 
     snprintfz(w->cookie2, NETDATA_WEB_REQUEST_COOKIE_SIZE,
-              NETDATA_REGISTRY_COOKIE_NAME "=%s; Expires=%s; SameSite=None ; %s",
-              guid, registry.registry_domain, edate, domain);
+              NETDATA_REGISTRY_COOKIE_NAME "=%s; Expires=%s; SameSite=None; Secure; %s",
+              guid, edate, domain);
 }
 
 static inline void registry_set_person_cookie(struct web_client *w, REGISTRY_PERSON *p) {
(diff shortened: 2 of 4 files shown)


ilyam8 (approved):

jacekkolasa:
@thiagoftsm Thanks for taking the initiative!
Just to be clear, this has nothing to do with the cloud, but with registry, right?

Normally `SameSite=None` requires also `Secure` cookie, but I don't know yet if we can set this (we don't want to break cases where cookies work now). I think that if we need to find the best solution, we should make some experiments, but this specific PR should be OK if it's tested and doesn't break common use-case.

thiagoftsm:
> @thiagoftsm Thanks for taking the initiative!
> Just to be clear, this has nothing to do with the cloud, but with registry, right?
> 
> Normally `SameSite=None` requires also `Secure` cookie, but I don't know yet if we can set this (we don't want to break cases where cookies work now). I think that if we need to find the best solution, we should make some experiments, but this specific PR should be OK if it's tested and doesn't break common use-case.

I was reading about [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) and I have some concerns to bring this and we stop access to `http` requests. I will do some tests, @novykh can you help us here?

jacekkolasa:
@thiagoftsm my problem is that in Chrome registry via `http://` indeed doesn't seem to work, but in Firefox - it does. For example in http://frankfurt.my-netdata.io/ (see registry?action=access call sending cookies and returning personId). My worry is that after setting `Secure` flag in the cookie, it will break.

thiagoftsm:
@jacekkolasa the last commit is bringing the missing `Secure` and after a talk with @novykh we are also setting change the policy to set the cookies.

thiagoftsm:
After a talk @novykh , we decided to move this PR for draft to review, because there are different scenarios we have to test and there is the possibility of we need to change some cookies.

cakrit:
After this is merged, @Ferroin please update the registry server, so we can try it out there too. Of course tests should be done with a private registry, before the merge.

thiagoftsm:
After a talk with @novykh we arrived in the conclusion that the best approach to fix the issue to enable by default the cookies, and if users have problems, because they do not use HTTPS, we are giving an option to disable these cookies.

jacekkolasa (approved):
I only now realised that we are not *changing* any cookies, we are *adding* one. If it's http://, then nothing breaks because the second (new) cookie is just disregarded.

I tested it and cookies are properly set. The only thing I'm not sure is that when we disable this feature in `netdata.conf`, the cookie is duplicated:

image

This doesn't seem to cause any harm, so I'm approving it

cakrit:
@Ferroin said we can ignore the failing tests. Anyone else need to review/test this? If not, let's press the red button! :)

thiagoftsm:
> @Ferroin said we can ignore the failing tests. Anyone else need to review/test this? If not, let's press the red button! :)

@novykh, please, can you review it? We worked together on this, I think it will be easy for you.

novykh (approved):

thiagoftsm:
@jacekkolasa I will retest to verify if the reported problem is happening on all browsers, because I do not remember to see this, please, what was the browser you had this problem?

jacekkolasa:
@thiagoftsm i see it both on newest Firefox & Chrome

PR #10664, opened 2021-02-19
Update screenshots and text for new Cloud nav

joelhans:
##### Summary

This PR updates some of the screenshots and text that address Netdata Cloud's new navigation.

##### Component Name

area/docs

PR #10678, opened 2021-02-23
Fix broken links in active alarms doc

joelhans:
##### Summary

I neglected to properly link these in #10664.

##### Component Name

area/docs


ilyam8 (approved):

OdysLam (approved):
Why are we using absolute urls and not relative?

joelhans:
@OdysLam Those docs don't live in this repo—only in the Learn repo—so we have to link to them directly to avoid situations where a user is reading a README here on GitHub and wants to click a link. Relative links won't work in that case.

PR #10680, opened 2021-02-24
Ebpf support new collectors

### LINKED ISSUE #10679: Changes on ebpf.plugin
Netdata is bringing new  ebpf collectors and this requires few changes for ebpf.plugin:

1 - It needs to load different eBPF programs, right now it is only loading `kprobes`.
2 - The new collectors will have charts integrated with `apps.plugin` and we will need to simplify this integration.
3 - Right now we are only creating charts on `ebpf` and `apps` submenu, but ideally data related to hard disk should go to disk, data related to memory should go to mem, and so on. To fix this, it is necessary pass `contexts` as arguments for few functions.

thiagoftsm:
##### Summary
Fixes #10679 
##### Component Name
`ebpf.plugin`

##### Test Plan
1 - Install this PR.
2 - Take a look on ebpf charts on `ebpf submenu` and `apps submenu`.

##### Additional Information
This PR is not dealing with any changes at `kernel-collector` repo, but it is creating conditions for we bring features from it.


collectors/ebpf.plugin/ebpf.c
```diff
@@ -437,7 +442,7 @@ void ebpf_create_charts_on_apps(char *id, char *title, char *units, char *family
                                 char *algorithm, struct target *root)
 {
     struct target *w;
-    ebpf_write_chart_cmd(NETDATA_APPS_FAMILY, id, title, units, family, "stacked", order);
+    ebpf_write_chart_cmd(NETDATA_APPS_FAMILY, id, title, units, family, "stacked", "''", order);
```
underhood, line 445: I would find more natural to be able to set this parameter to NULL and then function `ebpf_write_chart_cmd` to output `''` when it sees that parameter is NULL.

collectors/ebpf.plugin/ebpf.c
```diff
@@ -343,22 +343,25 @@ void write_io_chart(char *chart, char *family, char *dwrite, long long vwrite, c
...
+ * @param charttype chart type
+ * @param context   chart context
+ * @param order     chart order
  */
-void ebpf_write_chart_cmd(char *type, char *id, char *title, char *units, char *family, char *charttype, int order)
+void ebpf_write_chart_cmd(char *type, char *id, char *title, char *units, char *family,
+                          char *charttype, char *context, int order)
 {
-    printf("CHART %s.%s '' '%s' '%s' '%s' '' %s %d %d\n",
+    printf("CHART %s.%s '' '%s' '%s' '%s' %s %s %d %d\n",
```
underhood, line 358: why other parameters are enclosed in single quotes `'`  but this one you must pass the parameter to it with the quotes

underhood (changes requested):
see inline comments above

collectors/ebpf.plugin/ebpf_process.c
```diff
@@ -786,11 +795,12 @@ static void ebpf_create_apps_charts(ebpf_module_t *em, struct target *root)
...
         return;
 
-    if (ebpf_modules[EBPF_MODULE_PROCESS_IDX].apps_charts)
-        ebpf_process_create_apps_charts(em, root);
-
-    if (ebpf_modules[EBPF_MODULE_SOCKET_IDX].apps_charts)
-        ebpf_socket_create_apps_charts(NULL, root);
+    int counter;
+    for (counter = 0; ebpf_modules[counter].thread_name; counter++) {
+        ebpf_module_t *ptr = &ebpf_modules[counter];
```
vlvkobal, line 800: Wouldn't it be better to use another name for the variable to avoid confusion with the second parameter?

thiagoftsm: I renamed the variable, but this specific function `ebpf_create_apps_charts` does not have this variable, I think you confused with the function above it (`ebpf_process_create_apps_charts`).

collectors/ebpf.plugin/ebpf.c
thiagoftsm: Internally we were never giving arguments with space, but I agree with you and I added them, it is more safe.

collectors/ebpf.plugin/ebpf.c
thiagoftsm: I agree 100% with you, but I thought it would be more readable.
After you wrote this, I am more comfortable to  use the NULL.

collectors/ebpf.plugin/ebpf_process.c
vlvkobal: Ah, yeah, I was confused by a gap in the diff 😁. Anyway, I think it is better now.

vlvkobal (approved):

collectors/ebpf.plugin/ebpf.c
```diff
@@ -343,22 +343,25 @@ void write_io_chart(char *chart, char *family, char *dwrite, long long vwrite, c
...
+                          char *charttype, char *context, int order)
 {
-    printf("CHART %s.%s '' '%s' '%s' '%s' '' %s %d %d\n",
+    printf("CHART %s.%s '' '%s' '%s' '%s' '%s' '%s' %d %d\n",
            type,
            id,
            title,
            units,
            family,
+           context,
```
underhood, line 364: should it be `(context ? context : "")` ?

thiagoftsm: I did not use this, because when a pointer with NULL value was given to Netdata, it could interpret what was given and the plugin runs without any error, so I considered this not necessary.

underhood: `it could interpret` you mean `printf` chagnes that to `"(null)"`? If so can we rely on this under all conditions?

thiagoftsm: I took a deeper look in the case now, and I observed that we are dealing with `"(null)"` only for `name` on `plugins.d`, initially I assumed that this was present for all fields, because I knew this was happened for `names`. Thanks to take a look on this, the latest commit is fixing.
Now I will extend a discussion, should we also do a special work for variables like `context`? Considering that by default Netdata makes a `context` when is not given, I think we are missing this.

underhood (approved):

vlvkobal (approved):

PR #10683, opened 2021-02-24
Add support for runtime configuration of UID/GID for Netdata user.

Ferroin:
##### Summary

This adds functionality to the Docker image allowing users to specify a custom UID and/or GID for Netdata to run as using the `NETDATA_UID` and `NETDATA_GID` environment variables.

The changes involve pushing the `netdata` user and group creation from the build process to the entrypoint script, and then adding some code to handle a handful of the most likely edge-cases, aswell as setting things up to automatically fix permissions on startup.

This new behavior is only exhibited if neither `DOCKER_USR` (kept for compatibility) nor `NETDATA_USER` (added for consistency) are set. This behavior was chosen on the assumption that a user who is specifically requesting that Netdata run as a particular user knows what they’re doing, and therefore we shouldn’t risk getting in their way.

This also allows some internal restructuring of the Dockerfile which lets us reduce the total number of layers, resulting in a marginally smaller image overall.

##### Component Name

area/packaging

##### Test Plan

Verified locally to behave identically to the current image after startup when the variables are not set, and to correctly handle updating the user and group when restarted with different values for the variables.

##### Additional Information

I’ve been looking at doing this for a long time now, but hadn’t gotten around to it yet. We need this indirectly for the Helm chart support for setup of persistent volumes though, so it got bumped up in priority.

### DIFF
--- a/packaging/docker/README.md
+++ b/packaging/docker/README.md
@@ -248,10 +248,9 @@ how you created the container.
 
 ### Custom agent UID/GID
 
-By default, the Netdata Agent in the container will run with a user ID and group ID of 201, matching the default
-IDs used on normal installations of Netdata. In the unlikely event that you need to use a different UID or GID
-for netdata for some reason, you can set the `NETDATA_UID` and/or `NETDATA_GID` environment variables for the
-container to thedesired UID/GID.
+By default, Netdata in the container will run with a user ID and group ID of `201`, matching the default IDs used
+on normal installations of Netdata. In the unlikely event that you need to use a different UID or GID for netdata,
+set the `NETDATA_UID` and/or `NETDATA_GID` environment variables for the container to the desired UID/GID.
 
 ### Add or remove other volumes
 
--- a/packaging/docker/README.md
+++ b/packaging/docker/README.md
@@ -246,6 +246,13 @@ If you don't want to destroy and recreate your container, you can edit the Agent
 above section on [configuring Agent containers](#configure-agent-containers) to find the appropriate method based on
 how you created the container.
 
+### Custom agent UID/GID
+
+By default, the Netdata Agent in the container will run with a user ID and group ID of 201, matching the default
+IDs used on normal installations of Netdata. In the unlikely event that you need to use a different UID or GID
+for netdata for some reason, you can set the `NETDATA_UID` and/or `NETDATA_GID` environment variables for the
+container to thedesired UID/GID.
+
 ### Add or remove other volumes
 
 Some of the volumes are optional depending on how you use Netdata:
--- a/packaging/docker/run.sh
+++ b/packaging/docker/run.sh
@@ -13,7 +13,6 @@ if [ ! "${DO_NOT_TRACK:-0}" -eq 0 ] || [ -n "$DO_NOT_TRACK" ]; then
   touch /etc/netdata/.opt-out-from-anonymous-statistics
 fi
 
-
 BALENA_PGID=$(ls -nd /var/run/balena.sock | awk '{print $4}')
 DOCKER_PGID=$(ls -nd /var/run/docker.sock | awk '{print $4}')
 
@@ -27,14 +26,77 @@ elif [[ $DOCKER_PGID =~ $re ]]; then
   DOCKER_HOST="/var/run/docker.sock"
   PGID=$(ls -nd /var/run/docker.sock | awk '{print $4}')
 fi
+
 export PGID
 export DOCKER_HOST
+
+create_group=
+remove_group=
+create_user=
+remove_user=
+user_in_group=
+
+if getent group netdata > /dev/null; then
+  existing_gid="$(getent group netdata | cut -d ':' -f 3)"
+
+  if [ "${existing_gid}" != "${NETDATA_GID}" ]; then
+    echo "Netdata group ID mismatch (expected ${NETDATA_GID} but found ${existing_gid}), the existing group will be replaced."
+    remove_group=1
+    create_group=1
+  fi
+else
+  echo "Netdata group not found, preparing to create one with GID=${NETDATA_GID}."
+  create_group=1
+fi
+
+if getent passwd netdata > /dev/null; then
+  existing_user="$(getent passwd netdata)"
+  existing_uid="$(echo "${existing_user}" | cut -d ':' -f 3)"
+  existing_primary_gid="$(echo "${existing_user}" | cut -d ':' -f 4)"
+
+  if [ "${existing_gid}" != "${NETDATA_UID}" ]; then
+    echo "Netdata user ID mismatch (expected ${NETDATA_UID} but found ${existing_uid}), the existing user will be replaced."
+    remove_user=1
+    create_user=1
+  fi
+
+  if [ "${existing_primary_gid}" = "${NETDATA_GID}" ]; then
+    user_in_group=1
+  else
+    echo "Netdata user is not in the correct primary group (expected ${NETDATA_GID} but found ${existing_primary_gid}), the user will be updated."
+  fi
+else
+  echo "Netdata user not found, preparing to create one with UID=${NETDATA_UID}."
+  create_user=1
+fi
+
+if [ -n "${remove_group}" ]; then
+  delgroup netdata netdata
+  delgroup netdata || exit 1
+fi
+
+if [ -n "${create_group}" ]; then
+  addgroup -g "${NETDATA_GID}" -S netdata || exit 1
+fi
+
+if [ -n "${remove_user}" ]; then
+  userdel netdata || exit 1
+fi
+
+if [ -n "${create_user}" ]; then
+  adduser -S -H -s /usr/sbin/nologin -u "${NETDATA_UID}" -h /etc/netdata -G netdata netdata
+elif [ -z "${user_in_group}" ]; then
+  usermod -a -G netdata netdata
+fi
+
+chown -R netdata:root /usr/lib/netdata /var/cache/netdata /var/lib/netdata /var/log/netdata
+chown -R netdata:netdata /var/lib/netdata/cloud.d
 
 if [ -n "${PGID}" ]; then
   echo "Creating docker group ${PGID}"
   addgroup -g "${PGID}" "docker" || echo >&2 "Could not add group docker with ID ${PGID}, its already there probably"
   echo "Assign netdata user to docker group ${PGID}"
-  usermod -a -G "${PGID}" "${DOCKER_USR}" || echo >&2 "Could not add netdata user to group docker with ID ${PGID}"
+  usermod -a -G "${PGID}" netdata || echo >&2 "Could not add netdata user to group docker with ID ${PGID}"
 fi
 
 if [ -n "${NETDATA_CLAIM_URL}" ] && [ -n "${NETDATA_CLAIM_TOKEN}" ] && [ ! -f /var/lib/netdata/cloud.d/claimed_id ]; then
--- a/packaging/docker/run.sh
+++ b/packaging/docker/run.sh
@@ -36,6 +36,10 @@ create_user=
 remove_user=
 user_in_group=
 
+if [ -n "${DOCKER_USR}" ]; then
+  NETDATA_USER="${DOCKER_USR}"
+fi
+
 if getent group netdata > /dev/null; then
   existing_gid="$(getent group netdata | cut -d ':' -f 3)"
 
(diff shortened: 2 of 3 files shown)


Ferroin:
@ilyam8 I’ve tagged you for review on this one as it0s related to the Helmchart persistence stuff you and @knatsakis have been working on.

ilyam8:
i run 2 containers using this PR and latest netdata/netdata

```cmd
[ilyam@pc netdata]$ ps faxu | grep netdata
ilyam     135427  0.0  0.0   6412  2296 pts/2    S+   22:22   0:00              \_ grep --colour=auto netdata
root      130541  2.5  0.0  69020 28708 ?        SNsl 22:15   0:10  \_ /usr/sbin/netdata -u  -D -s /host -p 19999 -W set web web files group root -W set web web files owner root
root      130584  0.1  0.0  27020  5184 ?        SNl  22:15   0:00      \_ /usr/sbin/netdata --special-spawn-server
root      130718  0.2  0.0  53980  2412 ?        SN   22:15   0:01      \_ /usr/libexec/netdata/plugins.d/apps.plugin 1
root      130725  0.1  0.0 725628 22908 ?        SNl  22:15   0:00      \_ /usr/libexec/netdata/plugins.d/go.d.plugin 1
root      130737  0.0  0.0   2428  1936 ?        SN   22:15   0:00      \_ bash /usr/libexec/netdata/plugins.d/tc-qos-helper.sh 1
root      130740  1.6  0.0  34208 30268 ?        SNl  22:15   0:06      \_ /usr/bin/python3 /usr/libexec/netdata/plugins.d/python.d.plugin 1

201       134461 11.8  0.0  69156 29012 ?        SNsl 22:21   0:07  \_ /usr/sbin/netdata -u netdata -D -s /host -p 19999 -W set web web files group root -W set web web files owner root
201       134493  0.6  0.0  27024  5148 ?        SNl  22:21   0:00      \_ /usr/sbin/netdata --special-spawn-server
root      134652  0.3  0.0  53964  2340 ?        SN   22:21   0:00      \_ /usr/libexec/netdata/plugins.d/apps.plugin 1
201       134670  2.8  0.0  34200 30280 ?        SNl  22:21   0:01      \_ /usr/bin/python3 /usr/libexec/netdata/plugins.d/python.d.plugin 1
201       134689  0.0  0.0   2428  1884 ?        SN   22:21   0:00      \_ bash /usr/libexec/netdata/plugins.d/tc-qos-helper.sh 1
201       134696  0.4  0.0 726140 22940 ?        SNl  22:21   0:00      \_ /usr/libexec/netdata/plugins.d/go.d.plugin 1
```

The difference, pay attention to `-u ...`

- this PR

```cmd
root      130541  2.5  0.0  69020 28708 ?        SNsl 22:15   0:10  \_ /usr/sbin/netdata -u  -D -s /host -p 19999 -W set web web files group root -W set web web files owner root
```

- latest

```cmd
201       134461 11.8  0.0  69156 29012 ?        SNsl 22:21   0:07  \_ /usr/sbin/netdata -u netdata -D -s /host -p 19999 -W set web web files group root -W set web web files owner root
```


Reason

https://github.com/netdata/netdata/blob/2bddc7a0f4ad406bf45f526a853b3d025839bc3b/packaging/docker/run.sh#L30

---

We still need to allow users to run netdata as `root`

Ferroin:
@ilyam8 Nice catch, I completely forgot about that use-case.

Ferroin:
@ilyam8 Fixed now.

knatsakis (approved):

Ferroin:
@ilyam8 @vkalintiris I would appreciate if I could get a second approval on this so we can finally merge it.

ilyam8:
We have [running-as-root-unsafe guide](https://learn.netdata.cloud/docs/agent/packaging/docker#running-as-root-unsafe). 

I create an image from this PR and tried the guide, it doesn't work. It works with the current run.sh. Is that expected? I re-read OP and it seems it is not.

- netdata/netdata

```cmd
[ilyam@pc ~]$ docker run -d --env DOCKER_USR=root netdata/netdata
4d73bc09620bb2b52d57e87b51b48ee545b00654b21a2317f8530a85a1f3d00f

[ilyam@pc ~]$ docker ps -a
CONTAINER ID   IMAGE             COMMAND              CREATED         STATUS                            
4d73bc09620b   netdata/netdata   "/usr/sbin/run.sh"   3 minutes ago   Up 3 minutes (healthy)     19999/tcp   clever_shaw
```

- this PR

```cmd

[ilyam@pc ~]$ docker run -d --env DOCKER_USR=root nd-test
28395800b70777fda8a2ab66a8ab6d68adf91e2fa07216e2247604b19f66ee77

[ilyam@pc ~]$ docker ps -a
CONTAINER ID   IMAGE             COMMAND              CREATED         STATUS                            PORTS       NAMES
28395800b707   nd-test           "/usr/sbin/run.sh"   3 minutes ago   Exited (1) 3 minutes ago               admiring_mendeleev

[ilyam@pc ~]$ docker logs admiring_mendeleev
Netdata group not found, preparing to create one with GID=201.
chown: unknown user/group root:netdata
```

ilyam8:
@Ferroin did you do any changes? Hard to understand because of force-push.

Ferroin:
> @Ferroin did you do any changes? Hard to understand because of force-push.

I had not yet, and probably will not get a chance to for a while as I’m dealing with issues with the new packaging code.

Ferroin:
@ilyam8 The case of running with a username other than `netdata` should be properly fixed now. The code was unconditionally expecting the `netdata` group to exist, but only creating it if the username was set to `netdata`.

I’ve also rebased onto the latest `master` branch so that all the CI checks are up to date.

Ferroin:
Closing and reopening to force checks to re-run properly...

packaging/docker/README.md
```diff
@@ -246,6 +246,13 @@ If you don't want to destroy and recreate your container, you can edit the Agent
 above section on [configuring Agent containers](#configure-agent-containers) to find the appropriate method based on
 how you created the container.
 
+### Custom agent UID/GID
+
+By default, the Netdata Agent in the container will run with a user ID and group ID of 201, matching the default
```
joelhans, line 251:
```suggestion
By default, Netdata in the container will run with a user ID and group ID of `201`, matching the default
```

packaging/docker/README.md
```diff
@@ -246,6 +246,13 @@ If you don't want to destroy and recreate your container, you can edit the Agent
...
+IDs used on normal installations of Netdata. In the unlikely event that you need to use a different UID or GID
+for netdata for some reason, you can set the `NETDATA_UID` and/or `NETDATA_GID` environment variables for the
+container to thedesired UID/GID.
```
joelhans, line 254:
```suggestion
container to the desired UID/GID.
```

packaging/docker/README.md
```diff
@@ -246,6 +246,13 @@ If you don't want to destroy and recreate your container, you can edit the Agent
 above section on [configuring Agent containers](#configure-agent-containers) to find the appropriate method based on
 how you created the container.
 
+### Custom agent UID/GID
+
+By default, the Netdata Agent in the container will run with a user ID and group ID of 201, matching the default
+IDs used on normal installations of Netdata. In the unlikely event that you need to use a different UID or GID
+for netdata for some reason, you can set the `NETDATA_UID` and/or `NETDATA_GID` environment variables for the
```
joelhans, line 253:
```suggestion
for Netdata, set the `NETDATA_UID` and/or `NETDATA_GID` environment variables for the
```

packaging/docker/README.md
Ferroin: I’m curious about the rationale behind this change. It’s specifically the agent that’s running in the container, and I was under the impression that we generally prefer to say ‘the Netdata Agent’ in such cases over just ‘Netdata’.

packaging/docker/run.sh
```diff
@@ -13,11 +13,79 @@ if [ ! "${DO_NOT_TRACK:-0}" -eq 0 ] || [ -n "$DO_NOT_TRACK" ]; then
...
+if [ -n "${remove_group}" ]; then
+  delgroup netdata netdata
+  delgroup netdata || exit 1
+fi
+
+if [ -n "${create_group}" ]; then
+  addgroup -g "${NETDATA_GID}" -S netdata || exit 1
+fi
+
+if [ "${NETDATA_USER}" = "netdata" ]; then
```
ilyam8, line 48: There is no _else_. Should it work if NETDATA_USER != netdata ?

```cmd
[ilyam@pc ~]$ docker run -d --env NETDATA_USER=justme nd_docker_runtime7ca0bc2f49d4a80a741d2b3a242a69c2d99bd510221bedf729cd8312d6b27a6b
[ilyam@pc ~]$ docker logs busy_darwin
Netdata group not found, preparing to create one with GID=201.
chown: unknown user/group justme:root
```

ilyam8:
@Ferroin that is a hard PR to review. Could be a lot of combinations: we have USER, UID, GID (not GROUP btw?), they could exist on the system or not, etc, etc. I feel like we need a table with all the combinations and their expected outcome.

Ferroin:
> @Ferroin that is a hard PR to review. Could be a lot of combinations: we have USER, UID, GID (no GROUP btw?), they could exist on the system or not, etc, etc. I feel like we need a table with all the combinations and their expected outcome.

Good point.

The implementation here is supposed to work such that:

* If the user has set `NETDATA_USER` to `root`, we behave just like the current code does when `DOCKER_USR` is set to `root`.
* If the user has set `NETDATA_USER` to something other than `root` or `netdata`, we assume they know what they’re doing and intentionally stay out of their way as much as possible. Essentially, your comment on the conditional above falls into this category. The expectation is that if the user _actually_ needs this (almost nobody should), they will be mounting a custom `/etc/passwd` and `/etc/group` into the container which will have the correct settings.
* We specifically do not handle custom group names because such a situation is even less likely than a custom user name, and at that point the user should _probably_ be using a custom entrypoint script anyway.
* Otherwise, we treat the UID and GID overrides as independently as possible. It should be possible to specify just a UID, just a GID, or both and should still function in the least surprising way possible in each of those cases.

packaging/docker/run.sh
```diff
@@ -13,11 +13,79 @@ if [ ! "${DO_NOT_TRACK:-0}" -eq 0 ] || [ -n "$DO_NOT_TRACK" ]; then
...
+    remove_group=1
+    create_group=1
+  fi
+else
+  echo "Netdata group not found, preparing to create one with GID=${NETDATA_GID}."
+  create_group=1
+fi
+
+if [ -n "${remove_group}" ]; then
+  delgroup netdata netdata
```
ilyam8, line 40: > Usage: delgroup [USER] GROUP

Is `netdata` user is the only option at this point?

ilyam8:
@Ferroin I am not sure we need `delgroup` and `userdel`, it may be risky to do that on container start. We can take into account that changing an env variable is deleting an existing container and creating a new one (can't change UID, GID, and USER for created container).

What about the following:
 - `USER` doesn't exist on the system: create `netdata` group (GID); create USER (UID); add USER to the group; start Netdata (**default case**).
 - `USER` exists on the system: ignore UID, GID; start Netdata as USER, user the user primary group id. 

Is having the `netdata` group a hard requirement?

packaging/docker/run.sh
Ferroin: No, but the group needs to be empty for `delgroup` to let us delete it. The only user that _should_ exist in the group should be `netdata`.

We actually do need an additional check here though I guess, to make sure that the group and password databases are writable before we try to do this.

Ferroin:
> @Ferroin I am not sure we need `delgroup` and `userdel`, it may be risky to do that on container start. We can take into account that changing an env variable is deleting an existing container and creating a new one (can't change UID, GID, and USER for created container).
> 
> What about the following:
> 
> * `USER` doesn't exist on the system: create `netdata` group (GID); create USER (UID); add USER to the group; start Netdata (**default case**).
> * `USER` exists on the system: ignore UID, GID; start Netdata as USER, use the user primary group id.

Semantically, that’s not really any different from what’s happening now. The group deletion logic is just to make it all idempotent and ensure we have a consistent state. Additionally though, changing environment variables does not inherently guarantee that the container will be recreated. The current OCI runtime implementations all behave this way AFAIK, but nothing explicitly requires that that be the case.

> Is having the netdata group a hard requirement?

Currently yes, the existing entrypoint code assumes that such a group exists.

Ferroin:
Rebased to pick up latest CI and added conditional to avoid trying to modify users or groups if account databases are read only (we just assume the user knows what they're doing in this case).

Ferroin:
Rebased to resolve conflicts.

ilyam8:
> Additionally though, changing environment variables does not inherently guarantee that the container will be recreated.

Changing env vars is not supported by Docker engine. I think you _hack_ around it by mounting some file (.bashrc or something) or modifying `/var/lib/docker/containers/[container-id]/config.json` 🤷‍♂️ 

I'm still [not sure we need delgroup/userdel](https://github.com/netdata/netdata/pull/10683#issuecomment-852887180). Will wait for @knatsakis review.

Ferroin:
> > Additionally though, changing environment variables does not inherently guarantee that the container will be recreated.
> 
> Changing env vars is not supported by Docker engine. I think you _hack_ around it by mounting some file (.bashrc or something) or modifying `/var/lib/docker/containers/[container-id]/config.json` 

It’s not supported by Docker, but Docker is not the only available OCI runtime, and I’m not certain about whether or not other OCI runtimes support it.

knatsakis:
> I'm still [not sure we need delgroup/userdel](https://github.com/netdata/netdata/pull/10683#issuecomment-852887180). Will wait for @knatsakis review.

I would have used `groupmod` + `usermod` to change gid/uid, but I don't see the risk of delgroup/userdel either. It does duplicate user/group creation code though, so if that ever needs to change it will have to be changed in two places.

Nevertheless, approving.

knatsakis (approved):

Ferroin:
@ilyam8 This still needs approval.

ilyam8 (approved):

PR #10692, opened 2021-02-25
packaging/docker: fix claiming via env vars

ilyam8:
##### Summary

- `netdata-claim.sh` script expects `-token`, `-url`, `-rooms` and `proxy` params to be `-param=`

https://github.com/netdata/netdata/blob/a3ada604962b126b20c86a41e8aac5ca451b8a33/claim/netdata-claim.sh.in#L163-L178

---

- we shouldn't run `netdata-claim.sh` script if there is no `/var/lib/netdata/registry/netdata.public.unique.id`. That is the case on [the first container start](https://learn.netdata.cloud/docs/agent/claim#using-environment-variables). `netdata.public.unique.id` is not generated at that point.

https://github.com/netdata/netdata/blob/a3ada604962b126b20c86a41e8aac5ca451b8a33/claim/netdata-claim.sh.in#L144-L151

---

Current logic is to execute `netdata-claim.sh` in the [`run.sh`](https://github.com/netdata/netdata/blob/2bddc7a0f4ad406bf45f526a853b3d025839bc3b/packaging/docker/run.sh#L23-L28). If the script fails netdata containers fails to start, docker engine keeps restarting it. 

This PR makes netdata daemon to execute `netdata-claim.sh` script. It means that container no longer fails to start if there are some problems with claiming.

If current behaviour (claim script fails => containers fails to start) is intentional, then it is buggy - there are a lot of different claim errors and we shouldn't exit with code != 0 on some of them, for instance

https://github.com/netdata/netdata/blob/a3ada604962b126b20c86a41e8aac5ca451b8a33/claim/netdata-claim.sh.in#L59-L60

It is not really a problem to do a claim being already claimed, Cloud handles it (that is our default helmchart setup btw - execute claim script on every container start/re-creation).

##### Component Name

`packaging/docker`

##### Test Plan

- [X] start a container **w/o claim env vars**, ensure
  - it works
  - doesn't try to claim

```cmd
[ilyam@pc ~]$ docker exec netdata ps -o args | grep netdata -m 1
/usr/sbin/netdata -u netdata -D -s /host -p 19999 -W set web web files group root -W set web web files owner root
```
 
- [X] start a container **w/ claim env vars** and **w/o claimed_id**, ensure
  - it works
  - it tries to claim
  - available on the Cloud 

```cmd
[ilyam@pc ~]$ docker exec netdata ps -o args | grep netdata -m 1
/usr/sbin/netdata -u netdata -D -s /host -p 19999 -W set web web files group root -W set web web files owner root -W set2 cloud global enabled true -W set2 cloud global cloud base url https://app.netdata.cloud -W claim -url=https://app.netdata.cloud -token=XXXX -rooms=XXXX -proxy=
```

- [X] start a container **w/ claim env vars** and **w/ claimed_id**, ensure
  - it works
  - doesn't try to reclaim
  - available on the Cloud 
  
```cmd
[ilyam@pc ~]$ docker exec netdata ps -o args | grep netdata -m 1
/usr/sbin/netdata -u netdata -D -s /host -p 19999 -W set web web files group root -W set web web files owner root
```

### DIFF
--- a/packaging/docker/run.sh
+++ b/packaging/docker/run.sh
@@ -20,15 +20,21 @@ if [ -n "${PGID}" ]; then
   usermod -a -G "${PGID}" "${DOCKER_USR}" || echo >&2 "Could not add netdata user to group docker with ID ${PGID}"
 fi
 
-if [ -n "${NETDATA_CLAIM_URL}" ] &&
-  [ -n "${NETDATA_CLAIM_TOKEN}" ] &&
-  [ ! -f /var/lib/netdata/cloud.d/claimed_id ] &&
-  [ -f /var/lib/netdata/registry/netdata.public.unique.id ]; then
-  /usr/sbin/netdata-claim.sh \
-    -token="${NETDATA_CLAIM_TOKEN}" \
-    -url="${NETDATA_CLAIM_URL}" \
-    ${NETDATA_CLAIM_ROOMS:+-rooms="${NETDATA_CLAIM_ROOMS}"} \
-    ${NETDATA_CLAIM_PROXY:+-proxy="${NETDATA_CLAIM_PROXY}"}
+if [ -f /var/lib/netdata/cloud.d/claimed_id ] || [ -z "${NETDATA_CLAIM_URL}" ] || [ -z "${NETDATA_CLAIM_TOKEN}" ]; then
+  exec /usr/sbin/netdata \
+    -u "${DOCKER_USR}" \
+    -D -s /host \
+    -p "${NETDATA_LISTENER_PORT}" \
+    -W set web "web files group" root -W set web "web files owner" root \
+    "$@"
+else
+  exec /usr/sbin/netdata \
+    -u "${DOCKER_USR}" \
+    -D -s /host \
+    -p "${NETDATA_LISTENER_PORT}" \
+    -W set web "web files group" root -W set web "web files owner" root \
+    -W set2 cloud global enabled true \
+    -W set2 cloud global "cloud base url" "${NETDATA_CLAIM_URL}" \
+    -W "claim -url=${NETDATA_CLAIM_URL} -token=${NETDATA_CLAIM_TOKEN} -rooms=${NETDATA_CLAIM_ROOMS} -proxy=${NETDATA_CLAIM_PROXY}" \
+    "$@"
 fi
-
-exec /usr/sbin/netdata -u "${DOCKER_USR}" -D -s /host -p "${NETDATA_LISTENER_PORT}" -W set web "web files group" root -W set web "web files owner" root "$@"
--- a/packaging/docker/run.sh
+++ b/packaging/docker/run.sh
@@ -20,7 +20,10 @@ if [ -n "${PGID}" ]; then
   usermod -a -G "${PGID}" "${DOCKER_USR}" || echo >&2 "Could not add netdata user to group docker with ID ${PGID}"
 fi
 
-if [ -n "${NETDATA_CLAIM_URL}" ] && [ -n "${NETDATA_CLAIM_TOKEN}" ] && [ ! -f /var/lib/netdata/claim.d/claimed_id ]; then
+if [ -n "${NETDATA_CLAIM_URL}" ] &&
+  [ -n "${NETDATA_CLAIM_TOKEN}" ] &&
+  [ ! -f /var/lib/netdata/claim.d/claimed_id ] &&
+  [ -f /var/lib/netdata/registry/netdata.public.unique.id ]; then
   /usr/sbin/netdata-claim.sh \
     -token="${NETDATA_CLAIM_TOKEN}" \
     -url="${NETDATA_CLAIM_URL}" \
--- a/packaging/docker/run.sh
+++ b/packaging/docker/run.sh
@@ -21,10 +21,11 @@ if [ -n "${PGID}" ]; then
 fi
 
 if [ -n "${NETDATA_CLAIM_URL}" ] && [ -n "${NETDATA_CLAIM_TOKEN}" ] && [ ! -f /var/lib/netdata/claim.d/claimed_id ]; then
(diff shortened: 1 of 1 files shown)


ilyam8:
@stelfrag @Ferroin somehow we missed that claiming via environment variables was broken. I fixed it in this PR, but still there is a problem: have to restart the container to get it claimed - on the very very first start `netdatavarlib` persistent volume is empty, so we have no `netdata.public.unique.id`.

Ferroin:
There are a few issues I see with this:

* While we should not be failing to start if we fail to claim (when writing this I forgot we had `set -e` at the top of the entrypoint script), the correct solution is to just add `|| true` to the end of the claiming command.
* The machine ID issue will be resolved when #10084 is finally merged (and that _needs_ to be merged, I’m going to pester the reviewers about it again shortly).
* The proposed change compromises on the maintainability of the non-claiming related options.
* We shouldn’t be forcibly enabling the cloud options only if we’re trying to claim. They default to enabled in all of our Docker images, and if the user is changing them and then trying to claim, it’s on them, not us.
* I disagree given the context of the script’s execution that getting back an ‘already claimed’ response is not an issue. We only run the claiming script if the node lacks the information that is required for it to connect to the cloud. If it gets run and we get back an ‘already claimed’ response, then _something_ is wrong which requires user intervention to fix.

ilyam8:
> We only run the claiming script if the node lacks the information that is required for it to connect to the cloud.

That is in theory, in practice we don't. Claiming via docker env is broken and helmchart default is to claim on every restart

https://github.com/netdata/helmchart/blob/5903a4d4e5343018b6789ed0b5355c1bcb557dc3/charts/netdata/templates/daemonset.yaml#L83-L93


---

Why executing `netdata-claim.sh` in `run.sh` when netdata daemon can do it, i see it is implemented?

https://github.com/netdata/netdata/blob/a65193925e6ba5fe67aa57024353d403d5abb144/claim/claim.c#L75-L85

ilyam8:
> The proposed change compromises on the maintainability of the non-claiming related options.

Doesn't sound good, but i am not clear what you mean. Can you provide some examples?

Ferroin:
> > We only run the claiming script if the node lacks the information that is required for it to connect to the cloud.
> 
> That is in theory, in practice we don't. Claiming via docker env is broken and helmchart default is to claim on every restart

But that’s the Helm chart handling it, not the Docker image itself. The Helm chart is not currently using the Docker code at all, and the Docker code _does_ refrain from claiming if the claim ID file is present.

> 
> https://github.com/netdata/helmchart/blob/5903a4d4e5343018b6789ed0b5355c1bcb557dc3/charts/netdata/templates/daemonset.yaml#L83-L93
> 
> Why executing `netdata-claim.sh` in `run.sh` when netdata daemon can do it, i see it is implemented?
> 
> https://github.com/netdata/netdata/blob/a65193925e6ba5fe67aa57024353d403d5abb144/claim/claim.c#L75-L85

Three reasons:

* To avoid needing conditionals for the handling of arguments for the daemon.
* To keep the command-line size down. We have a _lot_ of stuff users may want to configure through the command line arguments, so leaving them as much space as possible is generally a good thing.
* It helps keep the claiming script output all in one place in the logs.

> Doesn't sound good, but i am not clear what you mean. Can you provide some examples?

If we’re handling the agent invocation as is proposed in this PR, any changes to options that need to be passed regardless of the claiming state need to be made in two places instead of one.

This is mostly an argument about the exact implementation, not the possibility of handling claiming through the agent itself. I would much prefer to have the options for claiming as a separate environment variable that gets set or not set based on whether we need to claim or not, and then always include that environment variable in the options.

Essentially, from a maintainability perspective, it’s better to do something like the following for the agent invocation:

```bash
exec /usr/sbin/netdata -u "${DOCKER_USR}" -D -s /host -p "${NETDATA_LISTENER_PORT}" -W set web "web files group" root -W set web "web files owner" root ${CLAIMING_OPTS} "$@"
```

And then set `$CLAIMING_OPTS` only if we need to claim the node, instead of having two different invocations depending on whether or not we’re claiming.

ilyam8:
> It helps keep the claiming script output all in one place in the logs.

What i basically did is moved claiming script invocation to the netdata daemon making. All the logs go to error.log.

Why we need 2 options? 

- claiming from the entrypoint.
- claiming via netdata daemon.

That long line argument doesn't sound strong enough, especially if i remind that we add claiming opts **only once**, we don't claim again if the agent is claimed (claimed_id exists).


> exec /usr/sbin/netdata -u "${DOCKER_USR}" -D -s /host -p "${NETDATA_LISTENER_PORT}" -W set web "web files group" root -W set web "web files owner" root ${CLAIMING_OPTS} "$@"

Yes, i agree.

knatsakis (approved):

Ferroin:
With the merge of #10084, this needs a rebase, and some of the changes may not be needed any longer.

ilyam8:
Yes, this PR does two things:
 - fixes claiming script parameters names
 - moves claiming to the netdata daemon (passing appropriate parameters)

1st, i think was addressed. 2nd is not relevant, because @Ferroin prefers to keep it as it is - call claiming script from run.sh.

Closing this PR as wontfix then.

ilyam8:
@Ferroin 

i didn't really get why you kept `/var/lib/netdata/claim.d/claimed_id`, we know it is wrong.

https://github.com/netdata/netdata/blob/0a47bbdd6f44bf56e309c96d718467ef931a0b4f/packaging/docker/run.sh#L23

Ferroin:
> @Ferroin
> 
> i didn't really get why you kept `/var/lib/netdata/claim.d/claimed_id`, we know it is wrong.
> 
> https://github.com/netdata/netdata/blob/0a47bbdd6f44bf56e309c96d718467ef931a0b4f/packaging/docker/run.sh#L23

I plan to fix that in another PR later today.

PR #10693, opened 2021-02-26
Add Linux page cache metrics to eBPF

thiagoftsm:
##### Summary
When I finish the documentation I will give more details.


web/gui/dashboard_info.js
```diff
@@ -1010,6 +1010,22 @@ netdataDashboard.context = {
...
+    },
+
+    'mem.cachestat_hits': {
+        info: 'When the processor needs to read or write a location in main memory, it checks for a corresponding entry in the page cache. If the entry is there, a page cache hit has occurred and the read is from the cache. Hits show pages accessed that were not modified (we are excluding dirty pages), this counting also excludes the recent pages inserted for read.'
+    },
+
+    'mem.cachestat_misses': {
+        info: 'When the processor needs to read or write a location in main memory, it checks for a corresponding entry in the page cache. If the entry is there, a page cache hit has occurred and the read is from the cache. Hits show pages accessed that were not modified (we are excluding dirty pages), this counting also excludes the recent pages inserted for read.'
```
vlvkobal, line 1026: You forgot to change the chart info for misses.

vlvkobal (changes requested):
There is `ebpf_modules[1]` in L1676. Please use a macro instead of the number.

collectors/ebpf.plugin/ebpf.conf
```diff
@@ -19,11 +19,13 @@
 #
 # The eBPF collector enables and runs the following eBPF programs by default:
 #
+#  `cachestat`: Make charts for kernel functions related to page cache.
```
vlvkobal, line 22: Why do you use the `cachestat` word? I think we don't need to add the `stat` suffix. We can simply use `cache` or `pagecache`.

collectors/ebpf.plugin/ebpf_cachestat.h
```diff
@@ -0,0 +1,63 @@
...
+    long long hit;
+    long long miss;
+
+    netdata_cachestat_pid_t current;
+    netdata_cachestat_pid_t prev;
+} netdata_publish_cachestat_t;
+
+extern void *ebpf_cachestat_thread(void *ptr);
+
+#endif // NETDATA_EBPF_CACHESTAT_H
```
vlvkobal, line 63: Missing newline.

collectors/ebpf.plugin/ebpf_cachestat.c
```diff
@@ -0,0 +1,660 @@
...
+    ebpf_create_memory_charts();
+
+    pthread_mutex_unlock(&lock);
+
+    cachestat_collector(em);
+
+endcachestat:
+    netdata_thread_cleanup_pop(1);
+    return NULL;
+}
```
vlvkobal, line 660: Missing newline.

collectors/ebpf.plugin/ebpf_cachestat.c
```diff
@@ -0,0 +1,660 @@
...
+
+    // If hits are < 0, then its possible misses are overestimate due to possibly page cache read ahead adding
+    // more pages than needed. In this case just assume misses as total and reset hits.
+    calculated_number hits = total - misses;
+    if (hits < 0 ) {
+        misses = total;
+        hits = 0;
+    }
+
+    calculated_number ratio = (total > 0)?hits/total:0;
```
vlvkobal, line 143: Wrong formatting.

collectors/ebpf.plugin/ebpf_cachestat.c
```diff
@@ -0,0 +1,660 @@
...
+ */
+void cachestat_update_publish(netdata_publish_cachestat_t *out, uint64_t mpa, uint64_t mbd,
+                              uint64_t apcl, uint64_t apd)
+{
+    // Adapted algorithm from https://github.com/iovisor/bcc/blob/master/tools/cachestat.py#L126-L138
+    calculated_number total = (calculated_number) (((long long)mpa) - ((long long)mbd));
+    if (total < 0)
+        total = 0;
+
+    calculated_number misses = (calculated_number) (((long long)apcl) -((long long)apd));
```
vlvkobal, line 131: Wrong formatting

collectors/ebpf.plugin/ebpf.conf
thiagoftsm: We are using `cachestat` to keep compatibility with the names used by [`IOvisor`](https://github.com/iovisor/bcc/tree/master/tools), this was defined together our Product team (@manos-saratsis).

web/gui/dashboard_info.js
```diff
@@ -1010,6 +1010,22 @@ netdataDashboard.context = {
...
+    },
+
+    'mem.cachestat_hits': {
+        info: 'When the processor needs to read or write a location in main memory, it checks for a corresponding entry in the page cache. If the entry is there, a page cache hit has occurred and the read is from the cache. Hits show pages accessed that were not modified (we are excluding dirty pages), this counting also excludes the recent pages inserted for read.'
+    },
+
+    'mem.cachestat_misses': {
+        info: 'hen the processor needs to read or write a location in main memory, it checks for a corresponding entry in the page cache. If the entry is not there, a page cache miss has occurred and the cache allocates a new entry and copies in data for the main memory. Misses count page insertions to the memory not related to writing.'
```
vlvkobal, line 1026:
```suggestion
        info: 'When the processor needs to read or write a location in main memory, it checks for a corresponding entry in the page cache. If the entry is not there, a page cache miss has occurred and the cache allocates a new entry and copies in data for the main memory. Misses count page insertions to the memory not related to writing.'
```

thiagoftsm: Fixed with the latest commits.

vlvkobal:
Why are there data gaps in the `apps.cachestat_ratio` chart?

image

thiagoftsm:
> Why are there data gaps in the `apps.cachestat_ratio` chart?
> 
> image

This is a problem with the [original algorithm](https://github.com/iovisor/bcc/blob/master/tools/cachestat.py#L146-L154) and it is more visible when we take a look direct for application. This is an expected behavior, considering that we are sending data for Netdata only when ratio is different of zero. We took this decision to force a more smooth chart.

vlvkobal:
I wouldn't say it is smooth. Especially this one:

image

thiagoftsm:
> I wouldn't say it is smooth. Especially this one:
> 

I agree, I was trying to keep the same algorithm for `apps` and the `mem`, but I changed to have a better chart, at least on my environment it is smother now, please, test on yours.

underhood:
@thiagoftsm sorry have been away for few days. Will start reviewing immediately.

First comment PR name is misleading:
`Ebpf page cache` makes me think you are implementing cache inside eBPF collector to avoid some allocations or sth.
This PR should be named sth. like `add Linux page cache metrics to eBPF` ?

vlvkobal (approved):

PR #10697, opened 2021-03-01
Log ACLK cloud commands to access.log

stelfrag:
##### Summary
Logs ACLK requests from the cloud to the access.log
##### Component Name
database
aclk

##### Test Plan
- Start a cloud enabled / claimed agent
- Login to the cloud
- Go to the local agent and observe the `access.log`
   - You should see entries similar to 
      ```
      2021-03-01 15:42:55: 0: 196758 '[ACLK]:2' 'DATA' (sent/all = 3340/8971 bytes -63%, prep/sent/total = 0.04/1.01/1.05 ms) 200 '/api/v1/data'
      2021-03-01 15:42:55: 0: 196756 '[ACLK]:0' 'DATA' (sent/all = 3780/8299 bytes -54%, prep/sent/total = 0.02/0.41/0.43 ms) 200 '/api/v1/data'
       ```
This indicates that the /api/v1/data command is issued via ACLK and is processed by query thread 2 and query thread 0 respectively.


aclk/legacy/aclk_query.c
```diff
@@ -237,6 +239,7 @@ int aclk_queue_query(char *topic, void *data, char *msg_id, char *query, int run
     new_query->data = data;
     new_query->next = NULL;
     new_query->created = now_realtime_usec();
+    now_realtime_timeval(&new_query->tv_in);
```
underhood, line 242: If I am not mistaken one line above `new_query->created` can be calculated from this value as well thus avoiding calling `clock_gettime` twice (both `now_realtime_usec` and `now_realtime_timeval` call it internally). Also maybe the field `new_query->created` is obsoleted as it contains the same information just formated diferently

underhood (changes requested):

aclk/legacy/aclk_query.c
underhood: as this is done while holding `ACLK_QUEUE_LOCK` that means we are blocking every Query Thread here thus reducing 2 syscalls to 1 makes sense here

stelfrag: > If I am not mistaken one line above `new_query->created` can be calculated from this value as well thus avoiding calling `clock_gettime` twice (both `now_realtime_usec` and `now_realtime_timeval` call it internally). Also maybe the field `new_query->created` is obsoleted as it contains the same information just formated diferently

Looks like it can be calculated

`new_query->tv_in.tv_sec * USEC_PER_SEC + new_query->tv_in.tv_usec`

underhood (approved):

MrZammler (approved):

PR #10699, opened 2021-03-02
Read new variables from health alarm conf files, and pass them to v1 api and cloud.

MrZammler:
##### Summary

This PR reads 3 new variables (`class`, `component` and `type`) from the `health/health.d/` conf files, stores them, and passes them both to `/api/v1/alarm_log` and via ACLK to the cloud. (Under the payload `configured-alarms.alarms` of the `connect_alarms` message).

##### Component Name

Health

##### Test Plan

Add (`class`, `component`, `type`) to a `health/health.d/` conf file.
Observe `/api/v1/alarm_log`.
Observe the relevant aclk payload.

##### Additional Information

* When the new variables are missing it is ok to send an empty string to cloud ("")
* This is a draft PR, since we should wait for the xls at: https://docs.google.com/spreadsheets/d/1I-9RuX_Sr3bB-W-C2zzZ5TOkMgBBqbbjoNvApZIw2AY/ to be completed, and then store the variable values to all conf files via Vasilihs script.
* Currently only `load_average_15` contains the new variables as a proof-of-concept.
* It is also possible that I've missed some corner cases, so I will be reviewing the existing health code more.


MrZammler:
We need to send "Other" when a variable does not exist.

MrZammler:
Change Other to "Unknown" when a variable does not exist.

PR #10701, opened 2021-03-02
Add noauthcodecheck workaround flag to the freeipmi plugin

### LINKED ISSUE #7066: Workaround flags for freeipmi-plugin
Hello! Please add workaround command line options to freeipmi-plugin connect for example to OCP servers

    ipmi-dcmi -D LAN -h 192.168.1.1 -u admin -p password --get-system-power-statistic --workaround-flags=noauthcodecheck

Without this flag I get following error

    freeipmi.plugin ERROR : MAIN : ipmi_monitoring_sensor_readings_by_record_id(): password verification timeout

vlvkobal:
##### Summary
Fixes #7066

##### Component Name
freeipmi plugin

##### Test Plan
Run the freeipmi plugin with `noauthcodecheck` flag. Check if it works for protocol 1.5.

### DIFF
--- a/collectors/freeipmi.plugin/freeipmi_plugin.c
+++ b/collectors/freeipmi.plugin/freeipmi_plugin.c
@@ -1619,6 +1619,14 @@ int parse_outofband_driver_type (const char *str)
     return (-1);
 }
 
+int host_is_local(const char *hostname)
+{
+    if (hostname && (!strcmp(hostname, "localhost") || !strcmp(hostname, "127.0.0.1") || !strcmp(hostname, "::1")))
+        return (1);
+
+    return (0);
+}
+
 int main (int argc, char **argv) {
 
     // ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ilyam8 (approved):

collectors/freeipmi.plugin/freeipmi_plugin.c
```diff
@@ -1619,6 +1619,14 @@ int parse_outofband_driver_type (const char *str)
     return (-1);
 }
 
+int host_is_local(const char *host)
+{
+    if (host && (!strcmp(host, "localhost") || !strcmp(host, "127.0.0.1") || !strcmp(host, "::1")))
```
thiagoftsm, line 1624: We are using only one of the possible representations for IPV6 `loopback`  here, please, add also the [unicast representation](https://tools.ietf.org/html/rfc4291#section-2.5.3) and the [IPV6 notation](https://tools.ietf.org/html/rfc4291#section-2.4).

vlvkobal: Well, these checks are for the _most common_ cases, not for every corner case. If the user misconfigures the plugin he gets an error message from the `libipmimonitoring` library itself (but it's not as clear as our error message). My intention was to shorten [the list of addresses]( https://git.savannah.gnu.org/cgit/freeipmi.git/tree/common/miscutil/network.c#n271) to check.

thiagoftsm (approved):

PR #10706, opened 2021-03-03
Add JSON output option for buildinfo.

Ferroin:
##### Summary

This adds a variant of the `-W buildinfo` option called `-W buildinfojson` which outputs JSON instead of plain text, allowing for easier parsing and simpler handling in our anonymous statistics reporting.

Sample output (with JSON syntax hilighting):

```json
{
  "features": {
    "dbengine": "NO",
    "native-https": "YES",
    "cloud": "NO",
    "tls-host-verify": "YES"
  },
  "libs": {
    "jemalloc": "NO",
    "jsonc": "YES",
    "libcap": "YES",
    "libcrypto": "YES",
    "libm": "YES",
    "lws": "NO",
    "mosquitto": "NO",
    "tcmalloc": "NO",
    "zlib": "YES"
  },
  "plugins": {
    "apps": "YES",
    "cgroup-net": "YES",
    "cups": "YES",
    "ebpf": "NO",
    "ipmi": "NO",
    "nfacct": "YES",
    "perf": "YES",
    "slabinfo": "YES",
    "xen": "NO",
    "xen-vbd-error": "NO"
  },
  "exporters": {
    "kinesis": "NO",
    "pubsub": "NO",
    "mongodb": "NO",
    "prom-remote-write": "YES"
  }
}
```

##### Component Name

area/daemon

##### Test Plan

Manually tested by building with various options and verifying that the output from `netdata -W buildinfo` and `netdata -W buildinfojson` match correctly.

Also validated all the output to ensure it’s valid JSON.

##### Additional Information

The resultant output intentionally does not include the configure options. These are very likely to leak identifying information when dealing with custom builds, and would also require comprehensive escaping to be safe for use as part of  a JSON string.

This also intentionally does not use JSON-C. We need this to work even if we don’t have JSON-C, so a trivial static substitution approach like this is really the only option unless we want to include a different JSON implementation (and I strongly suspect we do not, as much as I would love to just vendor [cJSON](https://github.com/DaveGamble/cJSON) instead of depending on JSON-C as an external dependency).

### DIFF
--- a/daemon/buildinfo.c
+++ b/daemon/buildinfo.c
@@ -6,228 +6,237 @@
 // Optional features
 
 #ifdef ENABLE_ACLK
-#define FEAT_CLOUD "YES"
+#define FEAT_CLOUD 1
+#define FEAT_CLOUD_MSG ""
 #else
 #ifdef DISABLE_CLOUD
-#define FEAT_CLOUD "NO (by user request e.g. '--disable-cloud')"
+#define FEAT_CLOUD 0
+#define FEAT_CLOUD_MSG "(by user request)"
 #else
-#define FEAT_CLOUD "NO"
+#define FEAT_CLOUD 0
+#define FEAT_CLOUD_MSG ""
 #endif
 #endif
 
 #ifdef ENABLE_DBENGINE
-#define FEAT_DBENGINE "YES"
+#define FEAT_DBENGINE 1
 #else
-#define FEAT_DBENGINE "NO"
+#define FEAT_DBENGINE 0
 #endif
 
 #if defined(HAVE_X509_VERIFY_PARAM_set1_host) && HAVE_X509_VERIFY_PARAM_set1_host == 1
-#define FEAT_TLS_HOST_VERIFY "YES"
+#define FEAT_TLS_HOST_VERIFY 1
 #else
-#define FEAT_TLS_HOST_VERIFY "NO"
+#define FEAT_TLS_HOST_VERIFY 0
 #endif
 
 #ifdef ENABLE_HTTPS
-#define FEAT_NATIVE_HTTPS "YES"
+#define FEAT_NATIVE_HTTPS 1
 #else
-#define FEAT_NATIVE_HTTPS "NO"
+#define FEAT_NATIVE_HTTPS 0
 #endif
 
 // Optional libraries
 
 #ifdef ENABLE_JSONC
-#define FEAT_JSONC "YES"
+#define FEAT_JSONC 1
 #else
-#define FEAT_JSONC "NO"
+#define FEAT_JSONC 0
 #endif
 
 #ifdef ENABLE_JEMALLOC
-#define FEAT_JEMALLOC "YES"
+#define FEAT_JEMALLOC 1
 #else
-#define FEAT_JEMALLOC "NO"
+#define FEAT_JEMALLOC 0
 #endif
 
 #ifdef ENABLE_TCMALLOC
-#define FEAT_TCMALLOC "YES"
+#define FEAT_TCMALLOC 1
 #else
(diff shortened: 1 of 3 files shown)


underhood:
wouldn't it be better to have bools instead of strings "YES"/"NO"?

Ferroin:
> wouldn't it be better to have JSON bools (`"xyz": true`)instead of strings "YES"/"NO"?

Possibly, but that makes the code somewhat more complicated, and for at least the LWS thing we’d need much more complex handling on top of the extra complexity involved in just making the things boolean values.

andrewm4894 (dismissed):
The output of this looks good to me. I wonder if "true" / "false" might be better then "YES" / "NO" as it would be easier then to just convert them to booleans as needed. But at the same time maybe making a string called "true" has its own issues. 

No strong opinion from me anyway - either approach will work for what we need.

MrZammler:
Hi! I'm ok with this too (though I agree, it's better to use true/false).

However, unless I'm not up-to-date with something, if we want to include the buildinfo details on the statistics, we could avoid calling the binary with a `popen` to get the data, and instead get them from "inside" the agent code (and your code showed me how to do it! :-) ) and provide it to the anonymous-statistics script.

underhood:
@Ferroin take a look at my latest commit if disagree skip it and I approve (sorry for the direct push that was unintended)

daemon/buildinfo.c
```diff
@@ -6,224 +6,291 @@
...
+    printf("  },\n");
+
+    printf("  \"libs\": {\n");
+    printf("    \"jemalloc\": %s,\n",         FEAT_JSON_BOOL(FEAT_JEMALLOC));
+    printf("    \"jsonc\": %s,\n",            FEAT_JSON_BOOL(FEAT_JSONC));
+    printf("    \"libcap\": %s,\n",           FEAT_JSON_BOOL(FEAT_LIBCAP));
+    printf("    \"libcrypto\": %s,\n",        FEAT_JSON_BOOL(FEAT_CRYPTO));
+    printf("    \"libm\": %s,\n",             FEAT_JSON_BOOL(FEAT_LIBM));
+#if defined(ENABLE_ACLK)
+    printf("    \"lws\": \"%s\",\n", FEAT_JSON_BOOL(FEAT_LWS));
```
underhood, line 265: forgot to remove quotes here

daemon/buildinfo.c
```diff
@@ -6,224 +6,291 @@
...
+    printf("    \"lws-version\": \"%d.%d.%d\",\n", LWS_LIBRARY_VERSION_MAJOR, LWS_LIBRARY_VERSION_MINOR, LWS_LIBRARY_VERSION_PATCH);
+    printf("    \"lws-type\": %s,\n", FEAT_LWS_MSG);
```
underhood, line 267: here I forgot to add quotes

underhood (changes requested):

thiagoftsm:
I made some tests now and everything looks ok, but I will wait for @Ferroin to take decision about @underhood suggestions before to approve the PR.

Ferroin:
> @Ferroin take a look at my latest commit if disagree skip it and I approve (sorry for the direct push - that was unintended, planned to push to my fork... )

And this is part of why I’m not a C developer...

The changes you’re suggesting look good to me, I just never would have thought of such an approach myself.

underhood:
@Ferroin just trying to pay back your help with the installer and CI etc. :) I will fix my remaining 2 mistakes and we can go ahead with it. Can I push the fix directly to the branch?

Ferroin:
@underhood Yeah, just push the fixes directly.

thiagoftsm (approved):

MrZammler (approved):

PR #10709, opened 2021-03-04
Overhaul streaming documentation

joelhans:
##### Summary

No longer a WIP. 

This PR improves our streaming documentation by breaking up the monolithic `/stream/README.md` file into 3 different docs. 

- **How metrics streaming works** explains some key concepts & terminology around streaming. 
- **Enable streaming between nodes** is a quickstart for enabling a basic parent-child replication stream. 
- **Streaming reference** is—you guessed it—a new format for the streaming reference that's has a lot more detail while also being smaller than the original.

I still want to make some additional improvements, such as moving the content on enabling TLS/SSL into a separate document, but that will come in a future PR. Once this is merged, I'll also create another PR to remove the deprecated `/stream/README.md` and change all the links that point to it.

##### Component Name

area/docs
area/streaming


docs/store/stream-metrics.md
```diff
@@ -0,0 +1,116 @@
...
+
+The simplest streaming configuration is **replication**, in which a child node streams its metrics in real time to a
+parent node, and both nodes retain metrics in their own databases.
+
+See the [streaming reference doc](/streaming/README.md#supported-streaming-configurations) for details about other
+possible configurations.
+
+### Enable streaming on the parent node
+
+First, log onto the node that will act as the parent.
```
OdysLam, line 32:
```suggestion
First, log onto the machine which runs the node that will act as the parent.
```

docs/store/stream-metrics.md
```diff
@@ -0,0 +1,116 @@
...
+
+Run `uuidgen` to create a new API key, which is a randomly-generated machine GUID the Netdata Agent uses to identify
+itself while initiating a streaming connection. Copy that into a separate text file for later use.
```
OdysLam, line 35: I remember that the program does not exist by default on machines. Perhaps we can add something:
> In case you don't have it installed, you will need to install it. For example, on a Linux system that uses the `apt` package manager, you would run `apt-get install uuid-runtime`

docs/store/stream-metrics.md
```diff
@@ -0,0 +1,116 @@
...
+    enabled = yes
+    default memory mode = dbengine
+```
+
+Save the file and close it, then restart Netdata with `sudo systemctl restart netdata`, or the [appropriate
+method](/docs/configure/start-stop-restart.md) for your system.
+
+### Enable streaming on the child node
+
+Log into the node that will act as the child.
```
OdysLam, line 66:
```suggestion
Log onto the machine which runs the node that will act as the parent.
```

streaming/README.md
```diff
@@ -188,9 +153,9 @@ the unique id the Netdata generating the metrics (i.e. the Netdata that original
 them `/var/lib/netdata/registry/netdata.unique.id`). So, metrics for Netdata `A` that pass through
 any number of other Netdata, will have the same `MACHINE_GUID`.
 
-You can also use `default memory mode = dbengine` for an API key or `memory mode = dbengine` for
- a single host. The additional `page cache size` and `dbengine multihost disk space` configuration options
- are inherited from the global Netdata configuration.
+You can also use `default memory mode = dbengine` for an API key or `memory mode = dbengine` for a single host. The
```
OdysLam, line 156:
```suggestion
You can also use `default memory mode = dbengine` for an API key or `memory mode = dbengine` for a single host. In other words, the Parent can have a particular memory mode for every host that uses the same API key or different per-host.
```

OdysLam (dismissed):

docs/metrics-storage-management/reference-streaming.mdx
```diff
@@ -0,0 +1,481 @@
...
+| `initial clock resync iterations`               | `60`                      | Sync the clock of charts for how many seconds when starting.                                                                                                                                                                                                                                                                                              |
+| **`[API_KEY]` section**                         |                           | For configuring a parent node using an `API_KEY`.                                                                                                                                                                                                                                                                                                         |
```
thiagoftsm, line 48: I will give a suggestion, instead to have all parameters in an unique table, I think it will be more clear for users, if we treat the sections in different tables.

joelhans: Great suggestion—let me know what you think about the revision I just commited!

thiagoftsm: LGTM! :) 
I will approve your PR after this commit. Thank you @joelhans !

thiagoftsm (dismissed):

vlvkobal (changes requested):

docs/metrics-storage-management/reference-streaming.mdx
```diff
@@ -0,0 +1,486 @@
...
+> `netdata.conf`, be sure to add the IP address there so that it can access the API port.
+
+#### `default memory mode`
+
+The [database](/database/README.md) to use for all nodes using this `API_KEY`. Valid settings are `dbengine`, `ram`,
+`save`, `map`, or `none`.
+
+- `dbengine`: The default, recommended time-series database (TSDB) for Netdata. Stores recent metrics in memory, then
+  efficiently spills them to disk for long-term storage.
+- `ram`: Stores metrics _only_ in memory, which means metrics are when Netdata stops or restarts. Ideal for streaming
```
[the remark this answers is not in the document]
joelhans: Thanks very much @vlvkobal, for this and the other suggestions!

OdysLam:
Hey @joelhans,

As a first comment, perhaps we should have the SSL instructions inside the first guide and actively say that it's best to encrypt your data, as a way to be conscious and enforce security best practices. We could title the section as `optional` so that the user who is doing a homelab can quickly skip it.

joelhans:
@OdysLam Good idea. Let me see if I can distill the TLS instructions into something that's not very complicated and report back.

joelhans:
@OdysLam Done adding the TLS/SSL section!

docs/metrics-storage-management/enable-streaming.mdx
```diff
@@ -72,6 +72,45 @@ looks like the following:
...
+On the **parent** node, use OpenSSL to create the key and certificate, then use `chown` to make the new files readable
+by the `netdata` user.
+
+```bash
+sudo openssl req -newkey rsa:2048 -nodes -sha512 -x509 -days 365 -keyout /etc/netdata/ssl/key.pem -out /etc/netdata/ssl/cert.pem
+sudo chown netdata:netdata /etc/netdata/ssl/cert.pem k/etc/netdata/ssl/ey.pem
+```
+
+Next, enforce TLS/SSL on the web server. Open `netdata.conf`, scroll down to the `[web]` section, and look for the `bind
+to` setting. Add `^SSL=force` to turn on TLS/SSL.
```
thiagoftsm, line 92: I think it is a good idea to add a link here for the other options available for `^SSL=`.

vlvkobal (approved):

docs/metrics-storage-management/enable-streaming.mdx
joelhans: Good idea. Done!

thiagoftsm (approved):

vlvkobal (approved):

PR #10710, opened 2021-03-04
Add guide: Develop a custom data collector for Netdata in Python

joelhans:
##### Summary

This PR adds a new guide, written initially by a community member, about developing a custom Python collector.

@OdysLam Let's review this together at first, and fix up any loose ends, and then we'll bring in another reviewer or two.

##### Component Name

area/docs
area/collectors

### DIFF
--- a/docs/guides/python-collector.md
+++ b/docs/guides/python-collector.md
@@ -1,0 +1,421 @@
+<!--
+title: "Develop a custom data collector for Netdata in Python"
+description: "Learn how write a custom data collector in Python, which you'll use to collect metrics from and monitor any application that isn't supported out of the box."
+image: /img/seo/guides/python-collector.png
+author: "Panagiotis Papaioannou"
+author_title: "University of Patras"
+author_img: "/img/authors/panagiotis-papaioannou.jpg"
+custom_edit_url: https://github.com/netdata/netdata/edit/master/docs/guides/python-collector.md
+-->
+
+# Develop a custom data collector for Netdata in Python
+
+The Netdata Agent uses [data collectors](/docs/collect/how-collectors-work.md) to fetch metrics from hundreds of system,
+container, and service endpoints. While the Netdata team and community has built [powerful
+collectors](/collectors/COLLECTORS.md) for most system, container, and service/application endpoints, there are plenty
+of custom applications that can't be monitored by default.
+
+## Problem
+
+You have a custom application or infrastructure that you need to monitor, but no open-source monitoring tool offers a
+prebuilt method for collecting your required metric data.
+
+## Solution
+
+In this tutorial, you'll learn how to leverage the [Python programming language](https://www.python.org/) to build a
+custom data collector for the Netdata Agent. Follow along with your own dataset, using the techniques and best practices
+covered here, or use the included examples for collecting and organizing eithre random or weather data.
+
+## What you need to get started
+
+- A physical or virtual Linux system, which we'll call a _node_.
+- A working installation of the free, open-source [Netdata Agent](/docs/get/README.md).
+
+## Jobs and elements of a Python collector
+
+A Python collector for Netdata is a Python script that gathers data from an external source and transforms these data
+into charts to be displayed by Netdata dashboard. The basic jobs of the plugin are:
+
+- Gather the data from the service/application.
+- Create the required charts.
+- Parse the data to extract or create the actual data to be represented.
+- Assign the correct values to the charts
+- Set the order for the charts to be displayed.
+- Give the charts data to Netdata for visualization.
+
+The basic elements of a Netdata collector are:
+
+- `ORDER[]`: A list containing the charts to be displayed.
+- `CHARTS{}`: A dictionary containing the details for the charts to be displayed.
+- `data{}`: A dictionary containing the values to be displayed.
+- `get_data()`: The basic function of the plugin which will retrun to Netdata the correct values.
+
+Let's walk through these jobs and elements as independent elements first, then apply them to example Python code.
+
+### Determine how to gather metrics data
+
+Netdata can collect data from any program that can print to stdout. Common input sources for collectors can be logfiles,
+HTTP requests, executables, and more. While this tutorial will offer some example inputs, your custom application will
+have different inputs and metrics.
+
+A great deal of the work in developing a Netdata collector is investigating the target application and understanding
+which metrics it exposes and how to 
+
+### Create charts
+
+For the data to be represented in the Netdata dashboard, you need to create charts. Charts (in general) are defined by
+several characteristics: title, legend, units, type, and presented values. Each chart is represented as a dictionary
+entry:
+
+```python
+chart= {
+    "chart_name":
+        {
+            "options": [option_list],
+            "lines": [
+                [dimension_list]
+            ]
+        }
+    }
+```
+
+Use the `options` field to set the chart's options, which is a list in the form `options: [name, title, units, family,
+context, charttype]`, where:
+
+- `name`: The name of the chart.
+- `title` : The title to be displayed in the chart.
+- `units` : The units for this chart.
+- `family`: An identifier used to group charts together (can be null).
+- `context`: An identifier used to group contextually similar charts together (can be null).
+- `charttype`: Either `line`, `area`, or `stacked`. If null line is the default value.
+
+You can read more about `family` and `context` in the [web dashboard](/web/README.md#families) doc.
+
+Once the chart has been defined, you should define the dimensions of the chart. Dimensions are basically the metrics to
+be represented in this chart and each chart can have more than one dimension. In order to define the dimensions, the
+"lines" list should be filled in with the required dimensions. Each dimension is a list:
+
+`dimension: [id, name, algorithm, multiplier, divisor]`
+- `id` : The id of the dimension. Mandatory unique field (string) required in order to set a value.
+- `name`: The name to be presented in the chart. If null id will be used.
+- `algorithm`: Can be absolute or incremental. If null absolute is used. Incremental shows the difference from the
+  previous value.
+- `multiplier`: an integer value to divide the collected value, if null, 1 is used
+- `divisor`: an integer value to divide the collected value, if null, 1 is used
+
+The multiplier/divisor fields are used in cases where the value to be displayed should be decimal since Netdata only
+gathers integer values.
+
+### Parse the data to extract or create the actual data to be represented
+
+Once the data is received, your collector should process it in order to get the values required. If, for example, the
+received data is a JSON string, you should parse the data to get the required data to be used for the charts.
+
+### Assign the correct values to the charts
+
+Once you have process your data and get the required values, you need to assign those values to the charts you created.
+This is done using the `data` dictionary, which is in the form:
+
+`"data": {dimension_id: value }`, where:
+- `dimension_id`: The id of a defined dimension in a created chart.
+- `value`: The numerical value to associate with this dimension.
+
+### Set the order for the charts to be displayed
+
+Next, set the order of chart appearance with the `ORDER` list, which is in the form:
+
+`"ORDER": [chart_name_1,chart_name_2, …., chart_name_X]`, where:
+- `chart_name_x`: is the chart name to be shown in X order.
+
+### Give the charts data to Netdata for visualization
+
+Our plugin should just rerun the data dictionary. If everything is set correctly the charts should be updated with the
+correct values.
+
+## Framework classes 
+
+The `python.d` plugin has a number of framework classes that can be used to speed up the development of your python
+collector. Your class can inherit one of these framework classes, which have preconfigured methods.
+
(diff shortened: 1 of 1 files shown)


joelhans:
[Python collector guide (Panagiotis/Upatras)](https://trello.com/c/bapnPfHP/96-python-collector-guide-panagiotis-upatras)

docs/guides/python-collector.md
```diff
@@ -0,0 +1,421 @@
...
+    'humid_current'
+]
+```
+
+[Restart Netdata](/docs/configure/start-stop-restart.md) with `sudo systemctl restart netdata` to see the new
+min/max/average temperature chart with multiple dimensions:
+
+![A snapshot of the modified chart](https://i.imgur.com/g7E8lnG.png)
+
+## What's next?
```
OdysLam, line 412:
```suggestion
## What's next?

You can find the source code for the above examples on [GitHub](https://github.com/papajohn-uop/netdata). 
```

docs/guides/python-collector.md
```diff
@@ -0,0 +1,421 @@
...
+
+Now we you ready to start developing our Netdata python Collector and share it with the rest of the Netdata community.
+
```
OdysLam, line 415:
```suggestion
- If you need help while developing your collector, join our [Netdata
  Community](https://community.netdata.cloud/c/agent-development/9) to chat about it.
- Follow the
  [checklist](https://learn.netdata.cloud/docs/agent/collectors/python.d.plugin#pull-request-checklist-for-python-plugins)
  to contribute the collector to the Netdata Agent [repository](https://github.com/netdata/netdata).
- If this collector is a PoC, after validating the value of the collector in production, consider migrating the collector in Golang ([go.d.plugin](https://github.com/netdata/go.d.plugin)). Golang is more performant, easier to maintain, and simpler for users since it doesn't require a particular runtime on the node to execute (Python plugins require Python on the machine to be executed). Netdata uses go.d as the platform of choice for production-grade collectors.
- Celebrate! You have contributed to an open-source project with hundreds of thousands of users!

```

docs/guides/python-collector.md
```diff
@@ -0,0 +1,421 @@
...
+     }
+}
+```
+
+## Parse the data to extract or create the actual data to be represented
+
+A standard practice would be to either get the data on JSON format or transform them to JSON format. We use a dictionary
+to give this format and issue random values to simulate received data.
+
+We iterate through the names of the expected values and create a dictionary with key the name of each metric: and value
```
OdysLam, line 201:
```suggestion
We iterate through the names of the expected values and create a dictionary with the name of the value as `key` and a random value as `value`.

**Review by OdysLam — COMMENTED** [review]
I think we are good to go.

**Comment by andrewm4894** [comment]
I think it would be cool to actually use a real weather api like this https://openweathermap.org/api

Then you could configure it a bit more with actual params and maybe even do two jobs. One for some cities of interest in USA for example and then maybe another job for a different country or something.

Would not be that much extra work to adapt this a bit. I'd be happy to help in any way.

**Comment by OdysLam** [comment]
I think there is some merit on the multiple jobs part @andrewm4894, that's a nice idea. If you feel like it, go ahead and add a paragraph about jobs.

That being said, I believe that making more radical changes to include real API is outside of the scope of the guide. I feel that the goal is achieved without having to invest more time in it. The goal is to help a user understand the process to create a python collector. 

Any other feedback you may have is super welcome. It's useful to have as many pairs of eyes as possible.

**Comment by andrewm4894** [comment]
agree having to use an external api is messy and could just break in future. defo think it needs to have at least some sort of params in a conf file as a collector without a conf is not really useful. Jobs likewise although to a lesser extent. 

a nice little conf thing that could be added here maybe is a param for if you want celsius or fahrenheit. So a user could config if they want to display temperature as "Celsius" or "Fahrenheit".

**Comment by joelhans** [comment]
@OdysLam Addressed your comments/suggestions!

File: docs/guides/python-collector.md [hunk 4]
```diff
@@ -411,11 +411,20 @@ min/max/average temperature chart with multiple dimensions:
```
OdysLam, line 411:
```suggestion
## Add a configuration file

The last piece of the puzzle to create a fully robust python collector is the configuration file. Python.d uses configuration in [YAML](https://www.tutorialspoint.com/yaml/yaml_basics.htm) format and is used as follows:

- Create a configuration file in the same directory as the `<plugin_name>.chart.py`. Name it `<plugin_name>.conf`.
- Define a `job`, which is an instance of the collector. It is useful when you want to collect data from different sources with different attributes. For example, we could gather data from 2 different weather stations, which use different temperature measures: Fahrenheit and Celcius.
- You can define many different jobs with the same name, but with different attributes. Netdata will try each job serially and will stop at the first job that returns data. If multiple jobs have the same name, only one of them can run. This enables you to define different "ways" to fetch data from a particular data source so that the collector has more chances to work out-of-the-box. For example, if the data source supports both `HTTP` and `linux socket`, you can define 2 jobs named `local`, with each using a different method. 
- Check the `postgresql` collector configuration file on [GitHub](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/postgres/postgres.conf) to get a sense of the structure.

```yaml
weather_station_1:
    name: 'Greece'
    endpoint: 'https://endpoint_1.com'
    port: 67
    type: 'celcius'
weather_station_1:
    name: 'Florida: USA'
    endpoint: 'https://endpoint_2.com'
    port: 67 
    type: 'fahrenheit'
   ```  
   
Then,  you can access the above configuration variables in the `__init__` function:

```python
        self.endpoint = self.configuration.get('endpoint', <default_endpoint>)
```
It's worth noting that since you initiate the `framework class`,e.g `SimpleService.__init__`, the configuration will be available throughout the whole `Service` class of your module, as `self.configuration`. Finally, note that the `configuration.get` function takes 2 arguments, one with the name of the configuration field and one with a default value in case it doesn't find the configuration field. This allows you to define sane defaults for your collector.

Moreover, when creating the configuration file, create a large comment section that describes the configuration variables and inform the user about the defaults. For example, take a look at the `postgresql` collector on [GitHub](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/postgres/postgres.conf).

You can read more about the configuration file on the [Python.d documentation](https://learn.netdata.cloud/docs/agent/collectors/python.d.plugin). 
```

**Review by OdysLam — CHANGES_REQUESTED** [review]

File: docs/guides/python-collector.md (hunk 4 above)
Comment (OdysLam) [reply docs/guides/python-collector.md:411]: I added a mention about configuration files @joelhans  and @andrewm4894 . What do you think? I feel that I covered all the gotchas, but please feel free to suggest additions!

Comment (andrewm4894) [reply docs/guides/python-collector.md:411]: i like this - a nice example of multiple jobs and params. i think it's could be `weather_station_gr` and `weather_station_us` as the job names. Or at least looks like the both `weather_station_1` in the suggestion, maybe you meant _1 and _2? Either way, i think the names need to be unique for each job.

Comment (andrewm4894) [reply docs/guides/python-collector.md:411]: it would be nice to then mention or link to the fully done example collector in a repo somewhere that then uses the `type` param in the chart definition maybe. 

Or maybe we should just make a small PR to add a bit more to `example.conf` and then just link to that as another nice complete canonical example: https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/example/example.conf

I am actually going to make a PR to add some jobs and params to the example collector as it really should have some. https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/example

Comment (OdysLam) [reply docs/guides/python-collector.md:411]: I mention the postgresql example for the different jobs-same name example. You think I should add the example.conf as well and underline how the attribute can surface as a chart attribute?

(Nice PR on the example btw ✌️)

Comment (andrewm4894) [reply docs/guides/python-collector.md:411]: i made this PR so the example collector can be referenced when discussing jobs and params if useful in this guide: https://github.com/netdata/netdata/pull/10777

Comment (andrewm4894) [reply docs/guides/python-collector.md:411]: @OdysLam oh  i mean i'm just not sure this will work:

image

I don't think they can both be called `weather_station_1`. Unless maybe they can? But not sure. I think not allowed from reading [this](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/example/example.conf#L43)

Comment (OdysLam) [reply docs/guides/python-collector.md:411]: Fixed it @andrewm4894 , thanks! You were absolutely correct.

Comment (joelhans) [reply docs/guides/python-collector.md:411]: Nice work @OdysLam, thanks for adding this!

File: docs/guides/python-collector.md [hunk 5]
```diff
@@ -0,0 +1,481 @@
...
+
+Next, access the above configuration variables in the `__init__` function:
+
+```python
+def __init__(self, configuration=None, name=None):
+        SimpleService.__init__(self, configuration=configuration, name=name)
+        self.endpoint = self.configuration.get('endpoint', <default_endpoint>)
+```
+
+Because you initiate the `framework class` (e.g `SimpleService.__init__`)q, the configuration will be
```
OdysLam, line 451:
```suggestion
Because you initiate the `framework class` (e.g `SimpleService.__init__`), the configuration will be
```

docs/guides/python-collector.md
```diff
@@ -0,0 +1,481 @@
...
+available throughout the whole `Service` class of your module, as `self.configuration`. Finally, note that the
+`configuration.get` function takes 2 arguments, one with the name of the configuration field and one with a default
+value in case it doesn't find the configuration field. This allows you to define sane defaults for your collector.
+
+Moreover, when creating the configuration file, create a large comment section that describes the configuration
+variables and inform the user about the defaults. For example, take a look at the `postgresql` collector on
+[GitHub](https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/postgres/postgres.conf).
+
+You can read more about the confqiguration file on the [`python.d.plugin`
```
OdysLam, line 460:
```suggestion
You can read more about the configuration file on the [`python.d.plugin`
```

docs/guides/python-collector.md
joelhans: Thanks for catching these q's I added when I was trying to hit the keyboard shortcut to wrap lines! :laughing:

OdysLam (dismissed):

andrewm4894 (approved):
LGTM

docs/guides/python-collector.md
```diff
@@ -0,0 +1,481 @@
...
+Find the source code for the above examples on [GitHub](https://github.com/papajohn-uop/netdata). 
+
+Now we you ready to start developing our Netdata python Collector and share it with the rest of the Netdata community.
+
+- If you need help while developing your collector, join our [Netdata
+  Community](https://community.netdata.cloud/c/agent-development/9) to chat about it.
+- Follow the
+  [checklist](https://learn.netdata.cloud/docs/agent/collectors/python.d.plugin#pull-request-checklist-for-python-plugins)
+  to contribute the collector to the Netdata Agent [repository](https://github.com/netdata/netdata).
+- If you're developing a proof of concept (PoC), consider migrating the collector in Golang
```
andrewm4894, line 474:
```suggestion
- Check out the [example](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/example) python collector which is a minimal example collector you could also use as a starting point. Once comfortable with that then browse other [existing collectors](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin) that might have similarities to what you want to do.  
- If you're developing a proof of concept (PoC), consider migrating the collector in Golang
```

andrewm4894: just adding a link to look at the example collector too.

docs/guides/python-collector.md
```diff
@@ -0,0 +1,482 @@
...
+values for these metrics.
+
+### Chart creation
+
+First, create a single chart that shows the latest temperature metric:
+
+```python
+CHARTS = {
+    "temp_current": {
+         "options": ["my_temp", "Temperature", "Celsius", "TEMP", "weather_station", "line"],
```
ilyam8, line 188: context

> "weather_station"

During export we use context as a metrics name. Good naming is especially important here.

Context name common pattern is `A.B`

 - `A`: application/module (apache, cpu, etc.)
 - `B`: metric name.

In that example should be (assuming weather_station is the module name) `weather_station.temperature` i think.

joelhans: @ilyam8 Thanks for this. I updated all the lines in the doc that used `weather_station` as the context, and added a note about this naming convention up at the top where it explains the `options` field.

docs/guides/python-collector.md
```diff
@@ -0,0 +1,486 @@
+F<!--
```
ilyam8, line 1:
```suggestion
<!--
```

ilyam8: [Press F ...](https://dotesports.com/culture/news/the-origin-of-press-f-to-pay-respects)

ilyam8 (approved):

OdysLam (approved):

PR #10718, opened 2021-03-08
Add the ability to store chart labels in the database

stelfrag:
##### Summary
- This PR will create a new database table chart_label if it does not exist
- Store chart labels as they are attached to a chart

The information stored is
- chart_id blob (unique chart UUID)
- source_type int
- label_key text
- label_value text
- date_created (time since epoch)

##### Component Name
database

##### Test Plan
- Start an agent that define chart labels (e.g. k8s cluster)
- Connect to the database via `sqlite3`
  - Type `.schema` 
  - Observe the new table `chart_label` table

example:
```
fccdbc1e-52c6-411c-b8fa-fb5e27523356,4,container_name,qemu_ubuntu18.04
fccdbc1e-52c6-411c-b8fa-fb5e27523356,4,label,value
c4d3832a-5138-464b-9a2a-11147b0b3e8c,4,container_name,qemu_ubuntu18.04
c4d3832a-5138-464b-9a2a-11147b0b3e8c,4,label,value
902891fb-12dc-4f0c-b3bd-d2c1c242308f,4,container_name,qemu_ubuntu18.04
902891fb-12dc-4f0c-b3bd-d2c1c242308f,4,label,value
83a0f039-c47b-4e9c-968f-f569db3ab8b5,4,container_name,qemu_ubuntu18.04
83a0f039-c47b-4e9c-968f-f569db3ab8b5,4,label,value
```

### DIFF
--- a/database/rrdset.c
+++ b/database/rrdset.c
@@ -1933,7 +1933,7 @@ void rrdset_finalize_labels(RRDSET *st)
         replace_label_list(labels, new_labels);
     }
 #ifdef ENABLE_DBENGINE
-    if (likely(st->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE)) {
+    if (st->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE) {
         netdata_rwlock_wrlock(&labels->labels_rwlock);
         struct label *lbl = labels->head;
         while (lbl) {
--- a/database/rrdset.c
+++ b/database/rrdset.c
@@ -1932,6 +1932,18 @@ void rrdset_finalize_labels(RRDSET *st)
     } else {
         replace_label_list(labels, new_labels);
     }
+#ifdef ENABLE_DBENGINE
+    if (likely(st->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE)) {
+        netdata_rwlock_wrlock(&labels->labels_rwlock);
+        struct label *lbl = labels->head;
+        while (lbl) {
+            sql_store_chart_label(st->chart_uuid, (int)lbl->label_source, lbl->key, lbl->value);
+            lbl = lbl->next;
+        }
+        netdata_rwlock_unlock(&labels->labels_rwlock);
+    }
+#endif
+
     st->state->new_labels = NULL;
 }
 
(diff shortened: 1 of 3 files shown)


database/rrdset.c
```diff
@@ -1932,6 +1932,18 @@ void rrdset_finalize_labels(RRDSET *st)
     } else {
         replace_label_list(labels, new_labels);
     }
+#ifdef ENABLE_DBENGINE
+    if (likely(st->rrd_memory_mode == RRD_MEMORY_MODE_DBENGINE)) {
```
underhood, line 1936: thinking out loud... no sure `likely` is appropriate here. For people using DBENGINE, modern branch prediction will take the hint even without this. For people who use another memory mode, this will actually make things worse

stelfrag: No problem, I will remove it (besides the next PR down the road will remove this all together. It will work in any memory mode.

underhood: ah ok in such case no need to remove it from this PR ... up to you

database/sqlite/sqlite_functions.c
```diff
@@ -1020,3 +1022,53 @@ void add_migrated_file(char *path, uint64_t file_size)
...
+        goto failed;
+    }
+
+    rc = sqlite3_bind_text(res, 3, label, -1, SQLITE_STATIC);
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("Failed to bind label parameter to store label information");
+        goto failed;
+    }
+
+    rc = sqlite3_bind_text(res, 4, value, -1, SQLITE_STATIC);
```
underhood, line 1059: can we have empty values?

underhood: looks like it can be an empty string indeed but should not be NULL....

stelfrag: Well (for example) the `parse_k8s_data` function will not allow empty values. 

However there is no problem in the `value` being null here. Actually a null value binds to null properly and stores a null in the database in that case. 

So value = NULL would be equivalent to `sqlite3_bind_null(res, 4)`

underhood: what I have in mind is "" -> NULL in DB

stelfrag: The "" will be valid and stores a value of size 0 in the database. 

For the `create_label` maybe we need to recheck but most likely the caller does the extra checks needed (so it does not fail in unexpected ways)

PR #10720, opened 2021-03-08
Fixed detection of already claimed node in Docker images.

Ferroin:
##### Summary

The name of the path we are checking was incorrect, which resulted in the check not correctly skipping claiming on startup of containers that are already claimed.

##### Component Name

area/packaging

##### Test Plan

Verified that the check now correctly skips attempting to claim a node that is already claimed.

##### Additional Information

Credit to @ilyam8 for noticing this bug.

### DIFF
--- a/packaging/docker/run.sh
+++ b/packaging/docker/run.sh
@@ -20,7 +20,7 @@ if [ -n "${PGID}" ]; then
   usermod -a -G "${PGID}" "${DOCKER_USR}" || echo >&2 "Could not add netdata user to group docker with ID ${PGID}"
 fi
 
-if [ -n "${NETDATA_CLAIM_URL}" ] && [ -n "${NETDATA_CLAIM_TOKEN}" ] && [ ! -f /var/lib/netdata/claim.d/claimed_id ]; then
+if [ -n "${NETDATA_CLAIM_URL}" ] && [ -n "${NETDATA_CLAIM_TOKEN}" ] && [ ! -f /var/lib/netdata/cloud.d/claimed_id ]; then
   /usr/sbin/netdata-claim.sh -token "${NETDATA_CLAIM_TOKEN}" \
                              -url "${NETDATA_CLAIM_URL}" \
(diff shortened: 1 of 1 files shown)


vkalintiris (approved):

Ferroin:
@ilyam8 Could you please check and make sure this properly resolves your concerns about handling of claiming in the Docker images, and if so approve and merge?

ilyam8:
>  your concerns about handling of claiming

@Ferroin this change is correct. 

Another my concern is failing to start netdata if claiming script exits with code != 0. Is it what we want? If no, as you suggested, adding ` || true` will fix that issue.

Ferroin:
> > your concerns about handling of claiming
> 
> @Ferroin this change is correct.
> 
> Another my concern is failing to start netdata if claiming script exits with code != 0. Is it what we want? If no, as you suggested, adding ` || true` will fix that issue.

I’m of the opinion that we should fail (as a user, I would expect a container successfully starting with the claiming variables set to indicate that the container was successfully claimed), but we should probably see what the product team thinks. At minimum, we probably should fail for things that are likely to be transient networking issues (failed connection attempts, failed DNS lookups, etc) as doing so will cause the claiming to ultimately be re-attempted.

ilyam8:
> ’m of the opinion that we should fail 

I am emphasizing on it, because i had this experience - that is how i discovered that claiming via env vars didn't work. It was quite frustrating, especially taking into account that i 100% knew that the error is far from being critical and nothing stopped netdata from starting and connecting to the Cloud except claiming script exit code. The error was `Already claimed` (reason was wrong `[ -f .../claim.d/claiming_id` conditions) - it is not critical and there is no way to recover from it - so my containers stuck in a restart/crash loop.

ilyam8:
> should fail for things that are likely to be transient networking issues

Perhaps, i agree with that that it depends on an error. Until that _it depends_ implemented i think it is better to not fail - better user expirience.

PR #10723, opened 2021-03-08
Support VS Code container devenv

OdysLam:
##### Summary

With the existence of the `.devcontainer` directory, when a user opens the repository using VS Code, they will be automatically prompted to re-open the directory from inside the developer container. T

The VS Code window is connected to that container, being able to both use extensions to facilitate programming, tools. to check that the code is according to our standards (e.g flake8) and use VS Code sidebar as a GUI for file navigation inside the container. Finally, VS Code syncs the directory that we just opened inside the container, meaning that any file we place from our host machine into that directory (e.g netdata), it will be synced inside the container. 

The developer container contains all the required dependencies to build from source:
- netdata
- python.d.plugin
- go.d.plugin
- netdata dashboard

The new contribution workflow will be the following:
- Fork the netdata/netdata repository
- git clone the forked repository locally
- open the directory using VS Code
- reopen from inside the devenv, ready to develop!

This new **optional** workflow greatly simplifies the contribution process, since the user only needs to have VS Code installed. 

##### Test Plan

Download the repository and open the directory using VS Code

##### Additional Information
- https://code.visualstudio.com/docs/remote/containers
- https://github.com/netdata/community/tree/main/devenv


OdysLam:
Can we review this? @ilyam8 @Ferroin

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,14 @@
+{
+    "image": "netdata/devenv", 
+    "extensions":[
+		"golang.go",
```
ilyam8, line 4: Only this - if i want to develop in python i get goland extensions installed.

If that is ok and it is your plan i approve it

underhood:
I did test this on podman and LGTM. File sharing was not working so remote container in vscode was not possible but I think this is related to Fedora/Podman not directly to this PR

.devcontainer/devcontainer.json
OdysLam: The idea is to have a devenv for the whole suit of Netdata. I think it's more flexible and the overhead is truly minimum. All the extensions take a couple of seconds to be installed. The dependencies already exist inside the container since we have 1 container for both netdata, go.d and dashboard.

OdysLam: So that's OK from me :P

OdysLam:
> I did test this on podman and LGTM. File sharing was not working so remote container in vscode was not possible but I think this is related to Fedora/Podman not directly to this PR

That's a pity. In the docs it says that it *should* work, but no certainty. With podman thus the user has to run the container and ssh into it independently? If you run the container, is it possible to attach to it using VS code? (If you attach you get both access to the filesystem and a shell). It would be useful to see what level of support we have, so we can document it.

ilyam8 (approved):

Ferroin (approved):

underhood:
@OdysLam if i attach to container I get access to shell but no files there

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,14 @@
+{
```
rajsaurabhgupta, line 1: 2Zf

PR #10735, opened 2021-03-09
Rename struct avl to avl_element and the typedef to avl_t

vkalintiris:
Before:

```
struct foobar {
    avl avl;
    ...
}
```

After:

```
struct foobar {
    avl_t avl;
    ...
};
```

This will make the code compatible with C++ 11 without using the -fpermissive flag, which is
required when using the same name for both the type and the name of a field.

##### Component Name

Agent

##### Test Plan

CI


thiagoftsm (approved):
I did not receive any warning while I was compiling and Netdata performed as expected.

underhood (approved):

PR #10766, opened 2021-03-15
Fixed handling of perf.plugin capabilities.

Ferroin:
##### Summary

`CAP_PERFMON` is only available with Linux kernel 5.8 or newer, and needs a sufficiently up-to-date userspace to be able to set it.

With this change, we fall back to `CAP_SYS_ADMIN` (which `CAP_PERFMON` is a proper subset of) if we cannot set `CAP_PERFMON` for the perf plugin.

##### Component Name

area/packaging

##### Test Plan

Check that the script does not report any errors or failures on older Linux systems.


knatsakis (approved):

Ferroin:
Merging this without waiting for reviews as it fixes a significant bug in the installer and I’m confident it won’t make things any worse than they already are.

PR #10772, opened 2021-03-16
add a dump_methods parameter to alarm-notify.sh.in

MrZammler:
##### Summary

Add a dump_methods parameter to the alarm-notify.sh.in script. This will be used by the posthog analytics. The agent will execute the script from the user's directory, and will report back the available notification methods.

It also adds a check for `MSTEAM`, `PD` & `PROWL` methods. (Without them, they will be always set to `YES`). Presumably they are set somewhere later, but we need them to be there too for the dump_methods to be accurate.

Comments are very much encouraged, and if there is something I'm missing especially for the methods above.

One thing this PR can't detect is whether `SEND_CUSTOM` is set. Any ideas would be great!

##### Component Name

Health

##### Test Plan

When run with the dump_methods parameter, the script should just print the available methods and should NOT try to send anything. 
Normal execution should be possible without this parameter.

### DIFF
--- a/health/notifications/alarm-notify.sh.in
+++ b/health/notifications/alarm-notify.sh.in
@@ -209,6 +209,9 @@ if [[ ${1} = "unittest" ]]; then
   cfgfile="${3}"    # the location of the config file to use for unit testing
   status="${4}"     # the current status : REMOVED, UNINITIALIZED, UNDEFINED, CLEAR, WARNING, CRITICAL
   old_status="${5}" # the previous status: REMOVED, UNINITIALIZED, UNDEFINED, CLEAR, WARNING, CRITICAL
+elif [[ ${1} = "dump_methods" ]]; then
+    dump_methods=1
+    status="WARNING"
 else
   roles="${1}"               # the roles that should be notified for this event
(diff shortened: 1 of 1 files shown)


Ferroin:
This would actually be telling us what the user has enabled, not which methods are available. This distinction is kind of important, because:

* The ‘availability’ of a method is not very interesting information, as most methods just need a working `curl` command to be ‘available’ and we install cURL as part of the standard install, and the syslog method will be available on almost all modern Linux systems with no user setup, meaning that the only two methods in question in most cases are email (which needs a working `sendmail` command) and IRC (which requires some other command that I don’t remember the name of).
* It actually does not matter much to us what methods a user _could use_, but what methods they _are using_ is important in directing our choice of where to prioritize resources.

Ferroin (approved):

thiagoftsm (approved):
The script worked as expected and the alarms were sent with these modifications.

MrZammler:
Thanks guys!

Yes, Austin, you are correct: The purpose of this is to find out what notifications methods the user has enabled (by providing required parameters to some of them).

PR #10778, opened 2021-03-16
Adds _aclk_impl label

underhood:
##### Summary
1. Adds `_aclk_impl` label with values of `Legacy` or `Next Generation` or not present if ACLK not available
2. Removes unneeded include from aclk_otp which references file from legacy ACLK

##### Component Name
ACLK

##### Test Plan
build agent with aclk-ng and look at `/api/v1/info`, build with legacy and do the same.

### DIFF
--- a/aclk/aclk.c
+++ b/aclk/aclk.c
@@ -813,6 +813,7 @@ struct label *add_aclk_host_labels(struct label *label) {
             proxy_str = "none";
             break;
     }
+    label = add_label_to_list(label, "_aclk_impl", "Next Generation", LABEL_SOURCE_AUTO);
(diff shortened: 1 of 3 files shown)


aclk/aclk.c
```diff
@@ -813,6 +813,7 @@ struct label *add_aclk_host_labels(struct label *label) {
             proxy_str = "none";
             break;
     }
+    label = add_label_to_list(label, "_aclk_impl", "Next Generation", LABEL_SOURCE_AUTO);
```
stelfrag, line 816: I don't know if it would be problematic to have spaces in labels (in general)

underhood: should not be a problem if it is in value

stelfrag (approved):

MrZammler (approved):

PR #10788, opened 2021-03-17
Add a new parameter 'chart' to the /api/v1/alarm_log.

MrZammler:
##### Summary

Adds a `chart` parameter to `/api/v1/alarm_log`. When specified, only charts matching the parameter (to chart attribute) will be returned.

##### Component Name

Health

##### Test Plan

Call `/api/v1/alarm_log`.
Test by providing none of the available parameters
Test by providing either `after` or `chart` parameters
Test by providing both `after` and `chart` parameters

### DIFF
--- a/health/health_json.c
+++ b/health/health_json.c
@@ -100,9 +100,11 @@ void health_alarm_log2json(RRDHOST *host, BUFFER *wb, uint32_t after, char *char
 
     unsigned int max = host->health_log.max;
     unsigned int count = 0;
+    uint32_t hash_chart = 0;
+    if (chart) hash_chart = simple_hash(chart);
     ALARM_ENTRY *ae;
     for(ae = host->health_log.alarms; ae && count < max ; ae = ae->next) {
-        if( (ae->unique_id > after && !chart) || (ae->unique_id > after && chart && !strcmp(chart, ae->chart)) ) {
+        if( (ae->unique_id > after && !chart) || (ae->unique_id > after && chart && ae->hash_chart==hash_chart) ) {
             if(likely(count)) buffer_strcat(wb, ",");
             health_alarm_entry2json_nolock(wb, ae, host);
             count++;
--- a/health/health_json.c
+++ b/health/health_json.c
@@ -93,7 +93,7 @@ inline void health_alarm_entry2json_nolock(BUFFER *wb, ALARM_ENTRY *ae, RRDHOST 
     buffer_strcat(wb, "\t}");
 }
 
-void health_alarm_log2json(RRDHOST *host, BUFFER *wb, uint32_t after) {
+void health_alarm_log2json(RRDHOST *host, BUFFER *wb, uint32_t after, char *chart) {
     netdata_rwlock_rdlock(&host->health_log.alarm_log_rwlock);
 
     buffer_strcat(wb, "[");
@@ -101,10 +101,11 @@ void health_alarm_log2json(RRDHOST *host, BUFFER *wb, uint32_t after) {
     unsigned int max = host->health_log.max;
     unsigned int count = 0;
     ALARM_ENTRY *ae;
-    for(ae = host->health_log.alarms; ae && count < max ; count++, ae = ae->next) {
-        if(ae->unique_id > after) {
+    for(ae = host->health_log.alarms; ae && count < max ; ae = ae->next) {
+        if( (ae->unique_id > after && !chart) || (ae->unique_id > after && chart && !strcmp(chart, ae->chart)) ) {
             if(likely(count)) buffer_strcat(wb, ",");
             health_alarm_entry2json_nolock(wb, ae, host);
+            count++;
         }
     }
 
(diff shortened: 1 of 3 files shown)


health/health_json.c
```diff
@@ -93,18 +93,21 @@ inline void health_alarm_entry2json_nolock(BUFFER *wb, ALARM_ENTRY *ae, RRDHOST
...
 
     unsigned int max = host->health_log.max;
     unsigned int count = 0;
+    uint32_t hash_chart = 0;
+    if (chart) hash_chart = simple_hash(chart);
     ALARM_ENTRY *ae;
-    for(ae = host->health_log.alarms; ae && count < max ; count++, ae = ae->next) {
-        if(ae->unique_id > after) {
+    for(ae = host->health_log.alarms; ae && count < max ; ae = ae->next) {
+        if( ( !chart && ae->unique_id > after ) || ( chart && ae->unique_id > after && ae->hash_chart == hash_chart && !strcmp(ae->chart, chart) ) ) {
```
stelfrag, line 107: Isn't is simpler with 

 (ae->unique_id > after) && (!chart || (ae->hash_chart == hash_chart && !strcmp(ae->chart, chart))))

MrZammler: Yes it is!! https://github.com/netdata/netdata/pull/10788/commits/8dc332a6a5a03be7953f3c8a97a7823313dc3c59

stelfrag (approved):

web/api/web_api_v1.c
```diff
@@ -285,12 +286,13 @@ inline int web_client_api_request_v1_alarm_log(RRDHOST *host, struct web_client
         if(!name || !*name) continue;
         if(!value || !*value) continue;
 
-        if(!strcmp(name, "after")) after = (uint32_t)strtoul(value, NULL, 0);
+        if      (!strcmp(name, "after")) after = (uint32_t)strtoul(value, NULL, 0);
```
thiagoftsm, line 289: Hello,

I can see in this line a tab instead a simple space between `if` and the parenthesis, but this is not what called my attention, I saw that the base was set to 0, and I will suggest we use 10, because the `0` can mean different things according the official documentation:

```
 The strtoul() function converts the initial part of the string in nptr to an unsigned long value according to the given base, which must be between 2 and 36 inclusive, or be the special value 0.

       The  string  may  begin  with an arbitrary amount of white space (as determined by isspace(3)) followed by a single optional '+' or '-' sign.  If base is zero or 16, the string may then include a "0x" prefix, and the number
       will be read in base 16; otherwise, a zero base is taken as 10 (decimal) unless the next character is '0', in which case it is taken as 8 (octal).

       The remainder of the string is converted to an unsigned long value in the obvious manner, stopping at the first character which is not a valid digit in the given base.  (In bases above 10, the letter 'A' in either uppercase
       or lowercase represents 10, 'B' represents 11, and so forth, with 'Z' representing 35.)
```

When I accessed the URL on browser to get a specific timestamp, I selected the following value:

Screenshot_20210324_125853


But, when I ran another query I got a clean request. After to identify this problem I ran the following commands:

```bash
bash-5.1$ curl -k "https://localhost:19999/api/v1/alarm_log?chart=system.cpu" | wc -l
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  111k  100  111k    0     0  13.6M      0 --:--:-- --:--:-- --:--:-- 13.6M
4626
bash-5.1$ curl -k "https://localhost:19999/api/v1/alarm_log?chart=system.cpu&after=1616590251" 
[
]
bash-5.1$ curl -k "https://localhost:19999/api/v1/data?chart=system.cpu"  | wc -l
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 39257  100 39257    0     0  9584k      0 --:--:-- --:--:-- --:--:-- 12.4M
605
bash-5.1$ curl -k "https://localhost:19999/api/v1/data?chart=system.cpu&after=1616590535"  | wc -l
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  9072  100  9072    0     0  4429k      0 --:--:-- --:--:-- --:--:-- 4429k
144
```

As you can see it is not working as expected, please, compare with the `/api/v1/data`, because it looks like we are missing something here.

stelfrag: I hadn't noticed that but it seems to work as expected. It just means that the `after` can also be in hexadecimal or octal. So the following are identical (in hex, decimal and octal)

`http://127.0.0.1:19999/api/v1/alarm_log?after=0x5F4DF281`

and

`http://127.0.0.1:19999/api/v1/alarm_log?after=1598943873`

and 

`http://127.0.0.1:19999/api/v1/alarm_log?after=013723371201`

Note that the `after` is applied to the unique_id

thiagoftsm (approved):
After to talk with the team, we agreed that we won't change the variable names now, but there is a conflict between endpoints that we need to address in the future.

stelfrag (approved):

PR #10790, opened 2021-03-17
Properly prepare version for binary package builds.

Ferroin:
##### Summary

DEB package builds expect no leading letters in version numbers, and we don’t want them in our RPM builds either.

##### Component Name

area/packaging

##### Test Plan

Verified that the changes properly mangle the version information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -134,7 +134,7 @@ jobs:
 
     - name: Trigger DEB and RPM package build
       script:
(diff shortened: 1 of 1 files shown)


Ferroin:
Merging without waiting for approvals as this cannot break things worse than they currently are (even if it’s wrong, the worst that will happen is that we still don’t get binary packages published in tonight’s nightlies).

PR #10794, opened 2021-03-17
Update Coverity version so that scanning works again.

Ferroin:
##### Summary

The version served by the download link has changed, which means the filename has changed as well, which was causing the scan process to fail.

##### Component Name

area/ci

##### Test Plan

Verified that the version now matches the downloaded tarball name.

##### Additional Information

I’ll be opening a followup PR to add alerts to ensure this does not go unnoticed for so long again.

### DIFF
--- a/coverity-scan.sh
+++ b/coverity-scan.sh
@@ -40,7 +40,7 @@ set -e
 INSTALL_DIR="/opt"
 
 # the version of coverity to use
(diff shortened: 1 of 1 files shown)


thiagoftsm (approved):
After to update on my computer the coverity is running again and I am having correct reports.

underhood (approved):

PR #10798, opened 2021-03-17
Fix misleading long double typedef & macro.

vkalintiris:
##### Summary

`LONG_DOUBLE` and `LONG_DOUBLE_MODIFIER` names were misleading,
because they were operating on `doubles` instead of `long doubles`, when
Netdata is built without `long double` support.

This PR fixes this issue by:

1. Renaming the `LONG_DOUBLE_MODIFIER` macro to `CALCULATED_NUMBER_MODIFIER`, and
2. Deleting the `LONG_DOUBLE` macro and switching type references to `calculated_number`.

##### Component Name

agent

##### Test Plan

Built/Run an agent locally + green CI builds.

### DIFF
--- a/collectors/statsd.plugin/statsd.c
+++ b/collectors/statsd.plugin/statsd.c
@@ -39,7 +39,7 @@
 // data specific to each metric type
 
 typedef struct statsd_metric_gauge {
-    LONG_DOUBLE value;
+    calculated_number value;
 } STATSD_METRIC_GAUGE;
 
 typedef struct statsd_metric_counter { // counter and meter
@@ -68,7 +68,7 @@ typedef struct statsd_histogram_extensions {
 
     size_t size;
     size_t used;
-    LONG_DOUBLE *values;   // dynamic array of values collected
+    calculated_number *values;   // dynamic array of values collected
 } STATSD_METRIC_HISTOGRAM_EXTENSIONS;
 
 typedef struct statsd_metric_histogram { // histogram and timer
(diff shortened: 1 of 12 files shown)


vlvkobal:
While `LONG_DOUBLE` is misleading, using the word `calculated` in the new name might also be misleading 🤔. The variables that have been changed have nothing with `calculated_number` (which is a supplement to `collected_number`) - it just happened that their type is `(long) double`.

vkalintiris:
> While LONG_DOUBLE is misleading, using the word calculated in the new name might also be misleading thinking.

Totally agree. The underlying problem being that we treat the calculated_number type, not as something that specifies the range of possible values and the operations associated with them, but as something that signifies where values of that type came from (or where they'll end up). If you have any suggestions, I'm all ears. From my perspective, it would be better to replace `LONG_DOUBLE` with `long double`. Especially, when considering that building Netdata without `long double`s, breaks the agent (as I understand by inspecting the dashboard).

FWIW, the ultimate thing I want to investigate is using `doubles` instead of `long doubles` across the agent. `long doubles` occupy 16-bytes in most architectures which means that memory and CPU performance take a **significant hit** for no obvious advantage.

vlvkobal:
The same definitions you are working on are in `storage_number.h`. What about them?

> FWIW, the ultimate thing I want to investigate is using doubles instead of long doubles across the agent.

BTW, Prometheus uses doubles for metric values: [We restrained ourselves to 64-bit floats to simplify the design](https://prometheus.io/docs/introduction/faq/#why-are-all-sample-values-64-bit-floats-i-want-integers). I don't know about its internals. @ilyam8 could tell more.

We might revisit our `storage_number` as well. We have already run into trouble with precision because of our proprietary type.

Issue #9319, opened 2020-06-11
cannot claim node - code error 28


euggge:
##### Bug report summary

whenever i try to claim a node i get the following error:
```
Failed to connect to https://app.netdata.cloud, return code 28.
```
##### OS / Environment

```
Linux raspberrypi 4.19.118-v7l+ #1311 SMP Mon Apr 27 14:26:42 BST 2020 armv7l GNU/Linux
/etc/os-release:PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
/etc/os-release:NAME="Raspbian GNU/Linux"
/etc/os-release:VERSION_ID="10"
/etc/os-release:VERSION="10 (buster)"
/etc/os-release:VERSION_CODENAME=buster
/etc/os-release:ID=raspbian
/etc/os-release:ID_LIKE=debian
/etc/os-release:HOME_URL="http://www.raspbian.org/"
/etc/os-release:SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
/etc/os-release:BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

##### Netdata version

 
netdata v1.22.1-248-nightly

##### Steps To Reproduce

1. ... Install netdata
2. ... Claim node

##### Expected behavior

for the node to be claimed

underhood:
@euggge did you try again after some time? could it be some intermittent unavailability of the `app.netdata.cloud`?

euggge:
i doubt it, i tried it multiple times in the span of 3 days, both with the stable e and the nightlty version, same error

underhood:
are you able to to `wget https://app.netdata.cloud`?
I can try this on RPi tomorrow, I have one lying around (not sure it is 2nd or 3rd gen)

euggge:
yes wget works as expected 
```
--2020-06-11` 20:01:11--  https://app.netdata.cloud/
Resolving app.netdata.cloud (app.netdata.cloud)... 35.196.244.138
Connecting to app.netdata.cloud (app.netdata.cloud)|35.196.244.138|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 5238 (5.1K) [text/html]
Saving to: ‘index.html’

index.html          100%[===================>]   5.12K  --.-KB/s    in 0.002s  

2020-06-11 20:01:17 (2.92 MB/s) - ‘index.html’ saved [5238/5238]
```

[...]

PR #10800, opened 2021-03-17
claiming: increase curl connect-timeout and decrease number of claim attempts

ilyam8:
##### Summary

This PR attempts to fix slow DNS lookups, 5 seconds is not enough. I change it to 15 because we use 15 [when using wget](https://github.com/netdata/netdata/blob/e66b02ff6a936454acd63788c7b0e7953f0c1480/claim/netdata-claim.sh.in#L292).

see, sometimes it takes 5+ seconds
 - https://github.com/netdata/netdata/issues/9319#issuecomment-773674215
 - https://community.netdata.cloud/t/error-return-code-28-when-trying-to-claim-node/1047/19?u=ilyam8

The changes:
 - increate curl --connect-timeout from 5 to 30
 - decrease number of claim attempts from 5 to 3

⚠️ [We do 3 attempts](https://github.com/netdata/netdata/blob/e66b02ff6a936454acd63788c7b0e7953f0c1480/claim/netdata-claim.sh.in#L344-L355), so the maximum possible script execution time is up to 90 after this PR.

##### Component Name

`claim`

##### Test Plan

Nothing to test.

##### Additional Information

Difference between wget and curl timeouts

| Tool        | Option            | Description |
| ----------- | ----------------- | ----------- |
| wget        | -T                | Set the network timeout to seconds seconds.  This is equivalent to specifying --dns-timeout, -- connect-timeout, and --read-timeout, all at the same time. |
| curl        | --connect-timeout | Maximum  time  in  seconds that you allow curl's connection to take.  This only limits the connection phase. |

~~curl equivalent of wget's -T is --max-time~~

> -m, --max-time <seconds>
  Maximum time in seconds that you allow the whole operation to take.

~~Could switch to it 🤔 Just thinking out loud.~~ Wrong

### DIFF
--- a/claim/netdata-claim.sh.in
+++ b/claim/netdata-claim.sh.in
@@ -282,7 +282,7 @@ fi
 
 
 if [ "${URLTOOL}" = "curl" ] ; then
-        URLCOMMAND="curl --connect-timeout 15 --retry 0 -s -i -X PUT -d \"@${CLAIMING_DIR}/tmpin.txt\""
+        URLCOMMAND="curl --connect-timeout 30 --retry 0 -s -i -X PUT -d \"@${CLAIMING_DIR}/tmpin.txt\""
         if [ "${NOPROXY}" = "yes" ] ; then
                 URLCOMMAND="${URLCOMMAND} -x \"\""
         elif [ -n "${PROXY}" ] ; then
--- a/claim/netdata-claim.sh.in
+++ b/claim/netdata-claim.sh.in
@@ -282,7 +282,7 @@ fi
 
 
 if [ "${URLTOOL}" = "curl" ] ; then
-        URLCOMMAND="curl --connect-timeout 5 --retry 0 -s -i -X PUT -d \"@${CLAIMING_DIR}/tmpin.txt\""
+        URLCOMMAND="curl --connect-timeout 15 --retry 0 -s -i -X PUT -d \"@${CLAIMING_DIR}/tmpin.txt\""
         if [ "${NOPROXY}" = "yes" ] ; then
                 URLCOMMAND="${URLCOMMAND} -x \"\""
         elif [ -n "${PROXY}" ] ; then
--- a/claim/netdata-claim.sh.in
+++ b/claim/netdata-claim.sh.in
@@ -341,7 +341,7 @@ attempt_contact () {
         return 0
 }
 
-for i in {1..5}
+for i in {1..3}
 do
         if attempt_contact ; then
                 echo "Connection attempt $i successful"


Ferroin:
`--max-time` is not actually equivalent to wget’s `-T`. `--max-time` sets the total wall-clock time the whole operation can take (from the initial DNS resolution all the way through actually completely downloading the resource being fetched), whereas wget’s `-T` sets timeouts on each phase of the operation (IOW, if you use `-T 10`, then the DNS lookup times out after 10 seconds, the connection setup times out after 10 seconds, and the process of reading the response times out if there is a delay of 10 or more seconds between reads).

So with `wget` (which already uses `-T 15`), the max execution time is actually just over 225 seconds, not just over 75 seconds.

---

That said, I’d actually vote for reducing the number of retries but significantly increasing the timeout. Sitting there andconstantly retrying without showing the user the error is not good UX, and our timeouts are indisputably far too short.

I’d vote for dropping to only retrying twice (so we only attempt the call three times total), but bumping the cURL timeout to thirty seconds (and keeping the current wget timeout the same, as that _seems_ to work and results in equivalent behavior for most cases). If the network is usable at all, the first attempt should almost always work with that timeout, and if it doesn’t it’s not super likely that a retry will, but retrying once or twice is still worth it.

ilyam8:
Yeah, you are right, not equivalent at all!

ilyam8:
@Ferroin i changed number of tries to 3.

> but bumping the cURL timeout to thirty seconds

Perhaps go with 15 for now? Should be enough for establishing the connection.

Ferroin:
> > but bumping the cURL timeout to thirty seconds
> 
> Perhaps go with 15 for now? Should be enough for establishing the connection.

It needs to include both the time for DNS resolution and for actually establishing the connection. This is a case where it’s not likely to hurt UX much if we have a timeout that0s a bit longer than absolutely necessary, but having one that is too short will cause problems for some users, so I would personally tend towards over-estimating the upper bound here.

ilyam8:
I don't really like long timeouts because we start netdata process after executing the claim script. Could get some problems with readiness probes in k8s (and CrashLoopBackOff as the result).

Ferroin:
> I don't really like long timeouts because we start netdata process after executing the claim script. Could get some problems with readiness probes in k8s (and CrashLoopBackOff as the result).

I would normally agree, but this is a case where the norm is going to be finishing far sooner than the timeout, and the type of thing we’re doing here is not going to be significantly more likely to work if we fail fast and retry than if we just wait for it to work.

ilyam8:
@Ferroin i applied suggested changes

Ferroin (approved):

thiagoftsm (dismissed):
Sometimes I had problems with this, but fortunately Brazilian internet is not a pattern around the world, thank you @ilyam8 !

ilyam8:
had to rebase/force-push to fix ci checks

thiagoftsm (approved):

vlvkobal (approved):

PR #10801, opened 2021-03-18
Find and replace variables in health info strings

MrZammler:
##### Summary

Finds and replaces variables (`$this`, `$family` and `$units`) in health info strings.

`$this` is mapped to `ae->new_value_string`
`$family` is mapped to `ae->family`
`$units` is mapped to `ae->units`

##### Component Name

Health

##### Test Plan

Edit an alarm's conf file, and add variables, e.g.:

`info: number of seconds ($this) since the last successful buffering of exporting data is $this, in family $family and with units $units.`

This will result in an `ae->info` string:

`number of seconds (00:00:02 ago) since the last successful buffering of exporting data is 00:00:02 ago, in family exporting_my_graphite_instance and with units seconds ago.`

##### Additional Information

I've noticed that when e.g. `units` is missing on the conf file, it defaults to `KiB` ?

This PR is to be merged to a feature branch.


vlvkobal:
Info strings are used in the dashboard as well. Will it be handled by the dashboard itself?

libnetdata/libnetdata.c
```diff
@@ -1491,3 +1491,41 @@ char *read_by_filename(char *filename, long *file_size)
         *file_size = size;
     return contents;
 }
+
+//https://www.geeksforgeeks.org/c-program-replace-word-text-another-given-word/
+char *find_and_replace(const char* s, const char* find, const char* replace)
```
underhood, line 1496: what is the license here?

MrZammler:
@vlvkobal 

Not sure to be honest, will have to get back with this...

libnetdata/libnetdata.c
underhood: hmm we will have to figure this out before merge can take place

MrZammler: OK, their policy is a bit fuzzy.... Will have a look, maybe for something clearer (in licence) or from scratch.. Thanks Timo!

MrZammler:
Hi Vladimir!

I've added this: https://github.com/netdata/netdata/pull/10801/commits/457ff9966d48c9f601fed0511bc5f635c0961f76 commit, more like a proof of concept for the agent to make the info replacements for the alarms endpoint. If we want it, we can use it?

ilyam8:
@MrZammler do we support `${variable}` syntax in the info? 

It is a valid syntax

https://github.com/netdata/netdata/blob/ff29e685f3e3a3e9f897561437585bdbb6d687ba/health/health.d/tcp_mem.conf#L16-L17

MrZammler:
> @MrZammler do we support `${variable}` syntax in the info?
> 
> It is a valid syntax
> 
> https://github.com/netdata/netdata/blob/ff29e685f3e3a3e9f897561437585bdbb6d687ba/health/health.d/tcp_mem.conf#L16-L17
> 
> It is nice to have to handle potentially ambiguous cases (e.g $this% => ${this}%).

Hi Ilya!

Sure thing. 

I wasn't really sure what to use, saw in the xls you had something like: `average CPU utilization is {{$this}}% over the last 10 minutes (excluding nice)`, so for this example, the variable is `{$this}` and you want the final string to be `average CPU utilization is {92}% over the last 10 minutes (excluding nice)` ? (Or is the variable `{{$this}}`) ?

ilyam8:
> saw in the xls you had something like

i wasn't aware of the syntax and just used something that it stands out so it is easier to see for me - that is a temporary sytanx 😄 

It looks logical for me to support the following cases (same cases we support in other fields):
 - $variable => value
 - ${variable} => value (w/o any {})

MrZammler:
Sure, it's not a problem to support both cases, unless we choose not to do it for performance reasons I guess, or perhaps having 2 cases could confuse users, or there is some already established way with dealing with variables (maybe in other contexts of the code ${something} is different than $something.

So right now, $variable works, I guess we'll wait on whether the other guys have an opinion on this.

health/health_log.c
```diff
@@ -498,6 +497,32 @@ inline ALARM_ENTRY* health_create_alarm_entry(
     ae->old_value_string = strdupz(format_value_and_unit(value_string, 100, ae->old_value, ae->units, -1));
     ae->new_value_string = strdupz(format_value_and_unit(value_string, 100, ae->new_value, ae->units, -1));
 
+    if (likely(info)) {
+
+        ae->info = strdupz(info);
+
+        if ( strstr (ae->info, "$this") ) {
+            char *buf=NULL;
+            buf = find_and_replace(ae->info, "$this", ae->new_value_string);
```
ilyam8, line 506: `this` is `value units`, not just a value. It affects wording and makes `$units` senseless in most of the cases.

Perhaps have `$this` as a value and use some oher var for `new_value_string`? @vlvkobal

MrZammler: You are right, sorry I didn't discuss it with you before.

I think it's better to change it so that $this is always just the value, and if the user wants can use $units if he wants?

ilyam8: Yes, i agree, it is better to have `$this` as just a value.

health/health_json.c
```diff
@@ -142,6 +142,24 @@ static inline void health_rrdcalc2json_nolock(RRDHOST *host, BUFFER *wb, RRDCALC
     char value_string[100 + 1];
     format_value_and_unit(value_string, 100, rc->value, rc->units, -1);
 
+    char *replaced_info = NULL;
+    if (likely(rc->info)) {
+
+        replaced_info = strdupz(rc->info);
+
+        if ( strstr (replaced_info, "$this") ) {
```
thiagoftsm, line 150: I will give some suggestions.

In normal situation  the `strstr` implementation is slow, as you can see [here](https://opensource.apple.com/source/tcl/tcl-10/tcl/compat/strstr.c.auto.html), and [here](https://github.com/lattera/freebsd/blob/master/sys/libkern/strstr.c). You are calling this function many times here and also in find_and_replace, what is your opinion about instead to pass as argument the text we pass the pointer and an identification from an `enum`?

Now a doubt, your code is easy to keep with these three variables, but are we going to extend it with more variables? If we need to do this, probably we need to do something like we have for `libnetdata/eval/`.

Best regards,

Thiago

health/health_log.c
MrZammler: Hi Ilya,

I've made some changes to use the raw value. I should mention though, that we do lose some of the goodies that `format_value_and_unit` provides, such as e.g. pretty print of seconds (like `00:00:02 ago` instead of just `2`), some precision changes (i.e. `97.0107422` vs `97'), and sometimes the value will be null (instead of undefined - which I don't think it will reach an alarm anyway, so it shouldn't matter). It's not bad (it's actually more 'accurate'), I think it's just a matter of what we like most.

health/health_json.c
MrZammler: Hi Thiago, thank you very much for your input, I will look into it.

Not sure if we're planning for more than three variables...

vlvkobal:
There is a problem with value updates in the info string in the dashboard. @dim08 please take a look:

image

Wouldn't it be better if the same update mechanism will be used as in badges?

ilyam8:
@vlvkobal i guess that is expected. But yes, if you are watching Alarms menu the value on the badge != the value in the info string 🤷‍♂️

ilyam8:
Usually `every` is 10 seconds

```cmd
0 health.d (master)$ grep "every:.*" -R * | grep -o "every.*" | sort | uniq -c
    197 every: 10s
     79 every: 1m
      3 every: 20s
     12 every: 30s
      2 every: 5m
      3 every: 5s
      2 every: 60
      7 every: 60s
```

So if i open the Alarm menu and browse the alarms i get badge value != info value in (up to) 10 seconds.

vlvkobal:
> So if I open the Alarm menu and browse the alarms I get badge value != info value in (up to) 10 seconds.

The values in info strings don't change until you close and reopen the Alarm menu.

ilyam8:
I think we only use `$family` for now to keep it simple, sometimes it provides a value (e.g network interface name).

Example alarm:

```
template: 1h_traffic_average_rate_in
      on: net.net
  lookup: average -10s unaligned absolute of received
    calc: $this / 1000
   units: Megabits/s
   every: 1s
    info: average inbound rate for network interface $family over the last hour
      to: sysadmin
```

That is how it looks in the logs

Screenshot 2021-03-23 at 16 15 15

health/health_json.c
```diff
@@ -142,6 +142,37 @@ static inline void health_rrdcalc2json_nolock(RRDHOST *host, BUFFER *wb, RRDCALC
     char value_string[100 + 1];
     format_value_and_unit(value_string, 100, rc->value, rc->units, -1);
 
+    char *replaced_info = NULL;
+
+    if (likely(rc->info)) {
+        char *m;
+        replaced_info = strdupz(rc->info);
+
+        while ( m = strstr (replaced_info, "$this") ) {
```
thiagoftsm, line 151: Hello @MrZammler ,

A huge improvement :) .
I was studying your code, and I agree 100% that it is good to have the `while`, because users can set different `$this` in the message, but I will give another suggestion for you. When the `$this` is overwritten, we do not expect to have more `$this` before the address `m`, instead to pass as argument for `strstr`  the variable `replace_info`,  I think we can pass `m`. Of course, this implies that m needs to be initialize before any call for the loop. What do you think?

Best regards!

MrZammler: Yes! I was thinking about this too (honest :-) ), I will have a go I think!

I really appreciate your feedback, many thanks!

MrZammler:
So, it seems we have these following options with this:

1) Remove `$this`, & `$units` and keep only `$family`
2) Keep `$this` to point to formatted value and `$family`, and only expose `$family` in conf files (The user can choose to use it).
3) Keep `$this` to point to formatted value, `$units`, `$value` to point to raw value, and `$family`, and only expose `$family` in conf files (The user can choose to use the others - but they will consume cpu to check if there).

ilyam8:
@MrZammler from my point of view [having several options](https://github.com/netdata/netdata/pull/10801#issuecomment-805764850) indicates that there is uncertainty, which means it is better to not expose something we don't need.

tl;dr option 1

health/health_json.c
thiagoftsm: You are welcome, I am happy to help you! :)

MrZammler:
Closing this, a cleaner PR with changes from this, is: https://github.com/netdata/netdata/pull/10885

PR #10802, opened 2021-03-18
Alerts notify template

jacekkolasa:
##### Summary
Replaced old alarm template with the new one.
There are few small things to do:
1. We'll need also to display "Global time", in UTC:
image

2. i'm not sure how to fill Agent installation path, ie. in 
> sudo {/etc/netdata}/edit-config health.d/{disks.conf}

3. "Warning for 10 minutes" - needs to be dynamic. Minutes should be shortened to "m", and when numbers are big, we should probably use hours

### DIFF
--- a/health/notifications/alarm-notify.sh.in
+++ b/health/notifications/alarm-notify.sh.in
@@ -3392,7 +3392,7 @@ Content-Transfer-Encoding: 8bit
                   <div
                     style="font-family:Ubuntu, Helvetica, Arial, sans-serif;font-size:14px;line-height:1;text-align:left;color:#555555;"
                   >Global time:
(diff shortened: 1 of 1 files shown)


MrZammler:
For the agent path, i.e. to fill:

`>sudo {/etc/netdata}/edit-config health.d/{disks.conf}<br />
                    The alarm to edit is at line {160}</div>`

I think we will need to edit the agent, and pass the values to some new variables/parameters to the script.

The agent already sends an $src variable which is e.g. `25@/home/evas/stuff/netdata/install_netdata//netdata/usr/lib/netdata/conf.d/health.d/exporting.conf` (and was used as is in the old template) and we should keep that because it is included in other notification methods.

Issue #10776, opened 2021-03-16
Slack notifications for GitHub Actions failures.


Ferroin:
Primarily for restoring parity with our older Travis-based CI.

PR #10803, opened 2021-03-18
Fix typos in GHA workflows.

Ferroin:
##### Summary

This fixes a bug in our CI introduced by #10776. The syntax errors technically got flagged there in CI, but nobody noticed because they aren’t required checks (that will be changing shortly).

##### Component Name

area/ci

##### Test Plan

Manually validated the files this time.


Ferroin:
Merging this without waiting for reviews, as it can’t make things worse than they already are.

PR #10820, opened 2021-03-19
health: apply adapter_raid alarms for every logical/physical device

ilyam8:
##### Summary

I think it is better to to apply LD status/PD state alarms per every device. Back then [`foreach`](https://learn.netdata.cloud/docs/agent/health/reference#alarm-line-lookup) wasnt implemented.

cc @kozicpetar

i have a question, is having a LD/PD not in desired state is CRIT level alarm in general? Or we can change it to WARN?

##### Component Name

`health`

### DIFF
--- a/collectors/python.d.plugin/adaptec_raid/adaptec_raid.chart.py
+++ b/collectors/python.d.plugin/adaptec_raid/adaptec_raid.chart.py
@@ -23,11 +23,13 @@ ORDER = [
 
 CHARTS = {
     'ld_status': {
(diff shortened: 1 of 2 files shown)


vlvkobal (approved):

thiagoftsm (approved):

kozicpetar:
@ilyam8 

In general you can set alarm per device, but I think is CRITICAL.

ilyam8:
@kozicpetar thx for the quick answer!

PR #10821, opened 2021-03-19
Pass extra information to the alarm-notify.sh script

MrZammler:
##### Summary

This PR adds some new parameters to the alarm-notify.sh script to complete the data needed by the new email template.

It constructs an `edit-config` command which will be displayed in the email. The command can be copy-pasted by the user to a terminal to perform the edit. The string also contains the line number of the alarm. 
The same string will be used by the cloud to include in the email message send by the cloud.

It also passes the `class` attribute, pretty-prints the elapsed time of active alarms and adds the `$host` and `$roles` variables to appropriate places in the email template.

##### Component Name

Health

##### Test Plan

Have email sending enabled by the agent for the alarms.
Observe the email message sent.

##### Additional Information

Not sure if the edit_command needs to be stored to health log.

This PR is to be merged to a feature branch.


jacekkolasa (approved):

health/health_config.c
```diff
@@ -650,6 +663,7 @@ static int health_readfile(const char *filename, void *data) {
             rt->name = strdupz(value);
             rt->hash_name = simple_hash(rt->name);
             rt->source = health_source_file(line, filename);
+            rt->edit_command = health_edit_command(line, filename);
```
thiagoftsm, line 666: Hello @MrZammler ,

Please, add a call for `freez()` with this variable as argument inside `rrdcalctemplate_free()`, else we will have a memory leak when this PR is merged.

Thank you!

MrZammler: Hi @thiagoftsm !

Thank you very much, I've pushed https://github.com/netdata/netdata/pull/10821/commits/d03606ba6e6a9d244c432dc515240d5e6c96a7f9

health/health_config.c
```diff
@@ -471,6 +471,18 @@ static inline char *health_source_file(size_t line, const char *file) {
     return strdupz(buffer);
 }
 
+static inline char *health_edit_command(size_t line, const char *file) {
+    char buffer[FILENAME_MAX + 1];
+    char *file_no_path = strrchr(file, '/');
+
+    if (likely(file_no_path))
+        snprintfz(buffer, FILENAME_MAX, "sudo %s/edit-config health.d/%s=%zu", netdata_configured_user_config_dir, file_no_path+1, line);
```
thiagoftsm, line 479: Hello @MrZammler ,

After the last commit, I decided to test your PR to verify if I was not missing something, I observed that the string is exactly what was planned, as you can see:

```conf
                        "command": "sudo /etc/netdata/edit-config health.d/cpu.conf=4",
                        "command": "sudo /etc/netdata/edit-config health.d/cpu.conf=17",
                        "command": "sudo /etc/netdata/edit-config health.d/cpu.conf=30",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=7",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=17",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=31",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=75",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=89",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=162",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=172",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=55",
                        "command": "sudo /etc/netdata/edit-config health.d/net.conf=65",
                        "command": "sudo /etc/netdata/edit-config health.d/disks.conf=12",
                        "command": "sudo /etc/netdata/edit-config health.d/disks.conf=26",
                        "command": "sudo /etc/netdata/edit-config health.d/disks.conf=12",
                        "command": "sudo /etc/netdata/edit-config health.d/disks.conf=12",
                        "command": "sudo /etc/netdata/edit-config health.d/disks.conf=26",
                        "command": "sudo /etc/netdata/edit-config health.d/disks.conf=12",
                        "command": "sudo /etc/netdata/edit-config health.d/disks.conf=12",
```

but, if somebody runs this command on terminal, they can have errors depending of default editor, for example, I use `vi` and I had this error:

```bash
bash-5.1$ /etc/netdata/edit-config health.d/disks.conf=12
File 'health.d/disks.conf=12' is not found in '/usr/lib/netdata/conf.d'
```
I propose we add the line as a comment instead part of command line.

MrZammler: Hi Thiago, I'm sorry for the confusion, I didn't state it in the original comment.

The string to be sent, contains also the line number. Both the local `alarm-notify.sh` script, and the cloud, when they receive this string, they will split it on the `=` sign. The command then will be printed on the email just up to the point `sudo /etc/netdata/edit-config health.d/disks.conf` and the line number will appear in a different place on the email notification.

Thanks!

ilyam8: Interesting, how do you use that line number? And what line is that? 🤔

MrZammler: Hi Ilya!

In the `alarm-notify.sh` script, it gets split up (Line 2809 - 2812 in this commit for `alarm-notify.sh` script, sorry can't send a link to that), will split this to the `=` character.

It will then be rendered to the email as:

mail_edit_command

The line number is already read when reading the conf files, so I didn't mess with that. It is already used in the similar `source` string to emails (which is left too, because it's used in other notification methods).

stelfrag:
This needs to change a bit to handle the fact that the command is not saved in the `health-log,db` file

MrZammler:
> This needs to change a bit to handle the fact that the command is not saved in the `health-log,db` file

Ok, on it. It will be removed from rrd, etc, and will be constructed dynamically when needed (in json and when sent to the script).

health/health_config.c
ilyam8: Thanks for the clarification!

MrZammler: You are welcome!

thiagoftsm: Thank you for the answer, now it makes more sense.

health/health.c
```diff
@@ -344,7 +344,9 @@ static inline void health_alarm_execute(RRDHOST *host, ALARM_ENTRY *ae) {
         }
     }
 
-    snprintfz(command_to_run, ALARM_EXEC_COMMAND_LENGTH, "exec %s '%s' '%s' '%u' '%u' '%u' '%lu' '%s' '%s' '%s' '%s' '%s' '" CALCULATED_NUMBER_FORMAT_ZERO "' '" CALCULATED_NUMBER_FORMAT_ZERO "' '%s' '%u' '%u' '%s' '%s' '%s' '%s' '%s' '%s' '%d' '%d' '%s' '%s'",
+    char *edit_command = health_edit_command_from_source(ae->source?ae->source:"UNKNOWN");
```
thiagoftsm: Hello @MrZammler ,

Here the `UNKNOWN` is 100% in uppercase, but at like 377, we have `Unknown`. Is there a special motive for this difference?

thiagoftsm: I saw the motive in the next files, another doubt, instead to compare with `UNKNOWN`, why did you not give a `NULL` pointer?

MrZammler: Well, for the class attribute, the `Unknown` has been decided, however the source attribute is already sending `UNKNOWN`, so it will keep like this...

In any case, I will merge this PR to the branch (not the master), so the rest of the team can test the changes. Before this goes to master we will have another chance to review it.

Thank you Thiago!

thiagoftsm (approved):

PR #10833, opened 2021-03-22
implements ACLK env endpoint

underhood:
##### Summary
Implements getting data and parsing them from new `/env` endpoint. This will be used in later PRs

##### Component Name
ACLK

##### Test Plan
aclk_env should be properly populated by data from the response from the cloud. Can test with production cloud.
The agent should connect to the cloud successfully.

### DIFF
--- a/aclk/aclk.c
+++ b/aclk/aclk.c
@@ -483,13 +483,13 @@ static int aclk_attempt_to_connect(mqtt_wss_client client)
         }
         aclk_env = callocz(1, sizeof(aclk_env_t));
         aclk_get_env(aclk_env, aclk_hostname, aclk_port);
-        url_t url;
-        memset(&url, 0, sizeof(url_t));
-        url_parse(aclk_env->auth_endpoint, &url);
+        url_t auth_url;
+        memset(&auth_url, 0, sizeof(url_t));
+        url_parse(aclk_env->auth_endpoint, &auth_url);
 
-        aclk_get_mqtt_otp(aclk_private_key, &mqtt_otp_user, &mqtt_otp_pass, &url);
+        aclk_get_mqtt_otp(aclk_private_key, &mqtt_otp_user, &mqtt_otp_pass, &auth_url);
 
-        url_t_destroy(&url);
+        url_t_destroy(&auth_url);
         mqtt_conn_params.clientid = mqtt_otp_user;
         mqtt_conn_params.username = mqtt_otp_user;
         mqtt_conn_params.password = mqtt_otp_pass;
--- a/aclk/aclk.c
+++ b/aclk/aclk.c
@@ -421,6 +421,7 @@ static int aclk_block_till_recon_allowed() {
     return 0;
 }
 
+#ifndef ACLK_DISABLE_CHALLENGE
 /* Cloud returns transport list ordered with highest
  * priority first. This function selects highest prio
  * transport that we can actually use (support)
@@ -435,6 +436,7 @@ static int aclk_get_transport_idx(aclk_env_t *env) {
     }
     return -1;
 }
+#endif
 
 /* Attempts to make a connection to MQTT broker over WSS
  * @param client instance of mqtt_wss_client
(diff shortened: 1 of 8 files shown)


stelfrag:
Looks like it connects and builds the env
```
(gdb) p **env->transports
$7 = {endpoint = 0x5555618f9210 "https://app.netdata.cloud/mqtt", type = ACLK_TRP_MQTT_3_1_1}
```

I will check the code now

aclk/aclk_otp.c
```diff
@@ -253,9 +289,357 @@ void aclk_get_mqtt_otp(RSA *p_key, char *aclk_hostname, int port, char **mqtt_us
...
+    https_req_t req = HTTPS_REQ_T_INITIALIZER;
+    https_req_response_t resp = HTTPS_RES_RESPONSE_T_INITIALIZER;
+
+    req.request_type = HTTP_REQ_GET;
+
+    char *agent_id = is_agent_claimed();
+    if (agent_id == NULL)
+    {
+        error("Agent was not claimed - cannot perform challenge/response");
+        buffer_flush(buf);
```
stelfrag, line 610: Did you mean to free the buffer in all those cases? Because you have buffer create in L599

underhood: thx, you are right this is a mistake. I think originally buffer was given by caller function but then I changed it and forgot to update the flushes.... will fix this

underhood: fixed

underhood:
rebased after prereq PR merged to master

stelfrag:
The agent will not connect or drops the connection. Trying to replicate reliably

underhood:
@stelfrag if you are getting 409 that is the too many connection kick from cloud

stelfrag:
> @stelfrag if you are getting 409 that is the too many connection kick from cloud

I managed to have two instances running -- scratch that!

underhood:
> > @stelfrag if you are getting 409 that is the too many connection kick from cloud
> 
> I managed to have two instances running -- scratch that!

Yeah Up the stack we have the PR with TBEB and there we log nice error messages with cloud reasons etc.

aclk/aclk.c
```diff
@@ -473,29 +494,67 @@ static int aclk_attempt_to_connect(mqtt_wss_client client)
...
+        aclk_get_mqtt_otp(aclk_private_key, &mqtt_otp_user, &mqtt_otp_pass, &auth_url);
+        url_t_destroy(&auth_url);
+
         mqtt_conn_params.clientid = mqtt_otp_user;
         mqtt_conn_params.username = mqtt_otp_user;
         mqtt_conn_params.password = mqtt_otp_pass;
 #endif
 
+        // Do the MQTT connection
+        ret = aclk_get_transport_idx(aclk_env);
```
stelfrag, line 530: This will now cause a crash if the ACLK_DISABLE_CHALLENGE is used

underhood: good catch forgot to retest the `ACLK_DISABLE_CHALLENGE`

underhood: should be fixed now, this is developer convenience flag

aclk/aclk_otp.c
```diff
@@ -298,3 +298,335 @@ void aclk_get_mqtt_otp(RSA *p_key, char *aclk_hostname, int port, char **mqtt_us
...
+        return 2;
+    }
+    if (resp.http_code != 200) {
+        error("The HTTP code not 200 OK (Got %d)", resp.http_code);
+        https_req_response_free(&resp);
+        buffer_flush(buf);
+        return 3;
+    }
+
+    if (parse_json_env(resp.payload, &aclk_env)) {
```
vkalintiris: I might be overtly worried but if the remote's response is malformed, is it possible to end up with a palyoad that is NULL?

underhood: thats why we have this check few lines above:
`if (aclk_https_request(&req, &resp)) {`

underhood: ah I see what you mean now.... you mean due to some cloud bug they could return `200 OK` but the payload will be empty. Added additional check. Thx

vkalintiris (approved):
LGTM. A couple of non-blocking comments inline.

stelfrag (approved):

underhood:
@vkalintiris I will bring some of your comments in follow up (polish) PRs

Issue #10832, opened 2021-03-22
Netdata crashes on restart if the dashboard constantly updates charts


vlvkobal:
##### Bug report summary
If "When to refresh charts?" is set to `Always` and disk charts are shown in the dashboard, Netdata crashes on start after #10771 was merged:
```
2021-03-22 18:54:09: netdata FATAL : WEB_SERVER[static5] : RRDSET 'disk.sda2' should be read-locked, but it is not, at function rrdr2value() at line 7 of file 'web/api/formatters/value/value.c' # : Resource temporarily unavailable

2021-03-22 18:54:09: netdata INFO  : WEB_SERVER[static5] : EXIT: netdata prepares to exit with code 1...
```

##### OS / Environment
```
Linux arch-esxi 5.11.8-arch1-1 #1 SMP PREEMPT Sun, 21 Mar 2021 01:55:51 +0000 x86_64 GNU/Linux
/etc/lsb-release:LSB_VERSION=1.4
/etc/lsb-release:DISTRIB_ID=Arch
/etc/lsb-release:DISTRIB_RELEASE=rolling
/etc/lsb-release:DISTRIB_DESCRIPTION="Arch Linux"
/etc/os-release:NAME="Arch Linux"
/etc/os-release:PRETTY_NAME="Arch Linux"
/etc/os-release:ID=arch
/etc/os-release:BUILD_ID=rolling
/etc/os-release:ANSI_COLOR="38;2;23;147;209"
/etc/os-release:HOME_URL="https://www.archlinux.org/"
/etc/os-release:DOCUMENTATION_URL="https://wiki.archlinux.org/"
/etc/os-release:SUPPORT_URL="https://bbs.archlinux.org/"
/etc/os-release:BUG_REPORT_URL="https://bugs.archlinux.org/"
/etc/os-release:LOGO=archlinux
```

##### Netdata version
```
Version: netdata v1.29.3-180-g65bc43d9c
[...]

PR #10835, opened 2021-03-22
Add lock check to avoid shutdown when compiled with internal and locking checks

stelfrag:
#10832

##### Summary
Add a lock check to avoid agent shutdown when compiled with internal checks and verify lock checks. This is a quick 
fix and it will be improved in the future.

##### Component Name
database, web

##### Test Plan
-  Verify  as  described in #10832, then apply PR and verify that the agent does not shutdown

### DIFF
--- a/web/api/formatters/value/value.c
+++ b/web/api/formatters/value/value.c
@@ -4,7 +4,8 @@
 
 
 inline calculated_number rrdr2value(RRDR *r, long i, RRDR_OPTIONS options, int *all_values_are_null) {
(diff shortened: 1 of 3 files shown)


vlvkobal:
I've tested the PR. No crashes so far.

vlvkobal (approved):

underhood (approved):
OK considering it is to be improved in the future

thiagoftsm (approved):
I did not have the reported error message when I tested this PR.

PR #10837, opened 2021-03-23
Fix memory leak when archived data is requested

stelfrag:
##### Summary
- Properly release memory used for the charts when accessing archived data (charts are archived) 
- Properly shutdown the SQLite database connection
  - Database close was not issued during agent shutdown
- Store and finalize SQL statements to release memory during shutdown (Quick Fix)
  - Keep a list of statements to finalize and do sqlite3_finalize on shutdown
- Address coverity 
  - CID 369624:  Resource leaks  (RESOURCE_LEAK)
  - CID 369623:  API usage errors  (LOCK)

##### Component Name
database

##### Test Plan
- SQLite Reported memory leaks during shutdown should vanish (or greatly reduced)
  - e.g. run with `valgrind --leak-check=full --show-leak-kinds=all`

### DIFF
--- a/database/sqlite/sqlite_functions.c
+++ b/database/sqlite/sqlite_functions.c
@@ -1281,6 +1281,11 @@ void sql_build_context_param_list(struct context_param **param_list, RRDHOST *ho
         sprintf(id, "%s.%s", sqlite3_column_text(res, 3), sqlite3_column_text(res, 1));
 
         if (!st || uuid_compare(*(uuid_t *)sqlite3_column_blob(res, 7), chart_id)) {
+            if (unlikely(st && !st->counter)) {
+                freez(st->context);
+                freez((char *) st->name);
+                freez(st);
+            }
             st = callocz(1, sizeof(*st));
             char n[RRD_ID_LENGTH_MAX + 1];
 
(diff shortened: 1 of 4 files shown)


database/sqlite/sqlite_functions.c
```diff
@@ -313,6 +337,7 @@ void delete_dimension_uuid(uuid_t *dimension_uuid)
             error_report("Failed to prepare statement to delete a dimension uuid");
             return;
         }
+        add_stmt_to_list(res);
```
underhood, line 340:
```
        rc = sqlite3_prepare_v2(db_meta, DELETE_DIMENSION_UUID, -1, &res, 0);
        if (rc != SQLITE_OK) {
            error_report("Failed to prepare statement to delete a dimension uuid");
            return;
        }
        add_stmt_to_list(res);
```
This 4 lines keep repeating... I would probably make function netdata_sqlite_prepare_v2 or sth. that ensures correct handling

stelfrag: Yes I will improve on a next iteration

stelfrag: Fixing it

underhood (approved):

database/sqlite/sqlite_functions.c
```diff
@@ -48,6 +48,31 @@ static int execute_insert(sqlite3_stmt *res)
...
+    static int idx = 0;
+    static sqlite3_stmt *statements[MAX_OPEN_STATEMENTS];
+
+    if (unlikely(!res)) {
+        while (idx > 0)
+            sqlite3_finalize(statements[--idx]);
+        return;
+    }
+
+    if (unlikely(idx == MAX_OPEN_STATEMENTS))
```
vkalintiris, line 64: AFAICT, this should never happen in practice. Would it be okay to add an `info` when `idx == MAX_OPEN_STATEMENTS`?

vkalintiris (approved):

database/sqlite/sqlite_functions.c
stelfrag: True it may not reach it and it's best effort to release memory during shutdown. 

I didn't put an info message here because we can't really make any use of it. The proper way will be to release the resources associated with the statement when that thread terminates (and take advantage of the resources during runtime)

PR #10839, opened 2021-03-23
prevents mqtt connection attempt on OTP failure

underhood:
##### Summary
Fixes MQTT connection being attempted when we couldn't get the OTP.

##### Component Name
ACLK-NG

##### Test Plan
Fail OTP (block cloud IP) see that MQTT connection attempt is not being made until next reconnect attempt (after honoring TBEB)

### DIFF
--- a/aclk/aclk.c
+++ b/aclk/aclk.c
@@ -520,9 +520,12 @@ static int aclk_attempt_to_connect(mqtt_wss_client client)
             continue;
         }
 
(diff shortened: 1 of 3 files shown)


stelfrag:
This is good to go when the stacked PRs are merged ( I will approve at that point)

stelfrag (approved):

vkalintiris (approved):

PR #10843, opened 2021-03-23
Check device names in diskstats plugin

### LINKED ISSUE #10841: Incorrect disk identification in the diskstats plugin
##### Bug report summary
If disks/partitions have the same major/minor numbers, Netdata can't distinguish the devices and show the wrong charts. The most problematic devices are nvme disks: #5744, #9560, [a community discussion](https://community.netdata.cloud/t/strange-disk-utilization-alert/1072).

The problem is that disks are identified by Netdata using major/minor numbers themselves, but they are not always unique. We should take into account device names. `iostat`, for example, uses device names for identification instead of major/minor numbers. We can't fully replicate the iostat behavior because Netdata does the additional job as well (bcache stats, aliases and mount points as family names, filtering based on major numbers, names, aliases, mount points).

The idea is to add checks for device names while keeping major/minor number checks untouched. It will ensure we distinguish devices from each other. At the same time, existing plugin functionality will be hardly affected.

##### Netdata version
v1.29.3-192-g04ce5769b

##### Component Name
diskstats module of proc plugin

##### Expected behavior
Stats are available for all disks/partitions present in a system.

vlvkobal:
##### Summary
We are checking device names in addition to major/minor numbers to ensure we distinguish devices from each other. 

Fixes #10841

##### Component Name
diskstats module of proc plugin

##### Test Plan
Check if all diskstats charts are available for all disks in the system. Verify if the names of charts and families are correct.

### DIFF
--- a/collectors/proc.plugin/proc_diskstats.c
+++ b/collectors/proc.plugin/proc_diskstats.c
@@ -627,12 +627,12 @@ static struct disk *get_disk(unsigned long major, unsigned long minor, char *dis
     // check if we can find its mount point
 
     // mountinfo_find() can be called with NULL disk_mountinfo_root
-    struct mountinfo *mi = mountinfo_find(disk_mountinfo_root, d->major, d->minor);
+    struct mountinfo *mi = mountinfo_find(disk_mountinfo_root, d->major, d->minor, d->device);
     if(unlikely(!mi)) {
         // mountinfo_free_all can be called with NULL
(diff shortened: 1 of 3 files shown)


ilyam8 (approved):

thiagoftsm (approved):
I do not have the specified device, but during my tests the other devices ran as expected, and the code looks great.

PR #10844, opened 2021-03-23
Fix agent crash when executing data query with context and non-existing chart_label_key

stelfrag:
##### Summary
Fix an agent crash when executing a data query with a context and chart_label_key that does not exist.

The extended functionality of chart label key is not supported for archived queries for now. 

This PR will ignore non-matching chart labels keys that do not exist in active (in-memory charts) and do no
attempt to access the sqlite metadata

##### Component Name
database, web

##### Test Plan
- Compile agent from master and execute
  `http://localhost:19999/api/v1/data?context=system.cpu&options=jsonwrap&chart_label_key=this_breaks`
- Apply the PR and try again


underhood (approved):
Seems to do as the label advertises

MrZammler (approved):

Issue #10522, opened 2021-01-19
Segfault in health_alarm_log_read


KrisShannon:
##### Bug report summary
After a disk full issue the last line of `health-log.db` was truncated
After cleaning up some disk space starting Netdata failed with a segfault on [Line 359 of health/health_log.c in a call to str2l](https://github.com/netdata/netdata/blob/v1.28.0/health/health_log.c#L359)

```
#0  str2l.119441 (s=0xfffabb8a642 <Address 0xfffabb8a642 out of bounds>) at libnetdata/inlined.h:75
#1  0x0000000000499ad2 in health_alarm_log_read (host=0x203a790, fp=0x21313b0, filename=0x203b7f0 "/var/lib/netdata/health/health-log.db") at health/health_log.c:359
#2  0x000000000058ac5d in health_alarm_log_load (host=0x203a790) at health/health_log.c:422
```

Deleting the last line of `health-log.db` allowed netdata to start up succesfully.

##### OS / Environment
```
Linux w97a4b328e.temp 3.10.0-962.3.2.lve1.5.24.9.el7.x86_64 #1 SMP Wed Feb 13 08:24:50 EST 2019 x86_64 x86_64 x86_64 GNU/Linux
/etc/cloudlinux-release:CloudLinux release 7.9 (Boris Yegorov) 
/etc/os-release:NAME="CloudLinux"
/etc/os-release:VERSION="7.9 (Boris Yegorov)"
/etc/os-release:ID="cloudlinux"
/etc/os-release:ID_LIKE="rhel fedora centos"
/etc/os-release:VERSION_ID="7.9"
/etc/os-release:PRETTY_NAME="CloudLinux 7.9 (Boris Yegorov)"
/etc/os-release:ANSI_COLOR="0;31"
/etc/os-release:CPE_NAME="cpe:/o:cloudlinux:cloudlinux:7.9:GA:server"
/etc/os-release:HOME_URL="https://www.cloudlinux.com/"
/etc/os-release:BUG_REPORT_URL="https://www.cloudlinux.com/support"
/etc/os-release:
/etc/redhat-release:CloudLinux release 7.9 (Boris Yegorov) 
/etc/system-release:CloudLinux release 7.9 (Boris Yegorov) 
```

##### Netdata version
netdata v1.28.0
RPM: netdata-1.28.0-1.el7.x86_64 from https://packagecloud.io/netdata/netdata/el/7/x86_64

##### Component Name
health

##### Expected behavior
[...]

PR #10850, opened 2021-03-24
Store and read new health attributes from the health log

MrZammler:
##### Summary

Store and read back the new attributes `class`, `component` and `type` from the health log.

##### Component Name

Health

##### Test Plan

The agent should be able to read a health db log from a previous version which does not contain these 3 attributes.


health/health_log.c
```diff
@@ -145,6 +146,10 @@ inline void health_alarm_log_save(RRDHOST *host, ALARM_ENTRY *ae) {
                             , ae->new_value
                             , ae->old_value
                             , (uint64_t)ae->last_repeat
+
```
thiagoftsm, line 149: Please, remove this blank line.

health/health_log.c
```diff
@@ -145,6 +146,10 @@ inline void health_alarm_log_save(RRDHOST *host, ALARM_ENTRY *ae) {
...
+                            , (ae->class)?ae->class:"Unknown"
```
thiagoftsm, line 150: I saw in other PR the usage of the word `Unknown`, and I am seeing it here in this PR too, considering that we are using a lot this word, I will give a suggestion for we do like we do inside [`health_config.c`](https://github.com/netdata/netdata/blob/master/health/health_config.c#L7-L29), this will help the compiler to optimize the code and I also think we will have a better organization for keywords.

health/health_log.c
```diff
@@ -191,7 +196,7 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char
         host->health_log_entries_written++;
         line++;
 
-        int max_entries = 30, entries = 0;
+        int max_entries = 33, entries = 0; //why was it 30 ? There should have been 28 entries
```
thiagoftsm, line 199: About your question, please, take a look on this issue https://github.com/netdata/netdata/issues/10522, it will help you to understand the motive we are using a bigger value.

health/health_log.c
```diff
@@ -364,6 +370,20 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char
 
             ae->last_repeat = last_repeat;
 
+            if (likely(entries > 28)) {
+                freez(ae->class);
+                ae->class       = strdupz(pointers[28]);
+                if(!*ae->class) { freez(ae->class); ae->class = NULL; }
```
thiagoftsm, line 376: You do not need to do this test here when some `NAMEz` function is called, because when it is not possible to allocate memory, Netdata will shutdown itself, please, take a look [here](https://github.com/netdata/netdata/blob/7b913eaf8b1f94af1f99b3acfa90ae36e7c47706/libnetdata/libnetdata.c#L165-L169), [here](https://github.com/netdata/netdata/blob/7b913eaf8b1f94af1f99b3acfa90ae36e7c47706/libnetdata/log/log.h#L95), and  [here](https://github.com/netdata/netdata/blob/7b913eaf8b1f94af1f99b3acfa90ae36e7c47706/libnetdata/log/log.c#L795-L840).
Any doubt, please, let me know.

health/health_log.c
MrZammler: Oks! with https://github.com/netdata/netdata/pull/10850/commits/0918653d9dd04ac94ef0b8968bc10db6e7a61cce

health/health_log.c
MrZammler: Oks with https://github.com/netdata/netdata/pull/10850/commits/0918653d9dd04ac94ef0b8968bc10db6e7a61cce

health/health_log.c
MrZammler: Hi Thiago! Thanks! So, you're ok with the value now being 33, right? (We add here 3 more values) ?

health/health_log.c
MrZammler: Thanks Thiago. To be honest, I followed the pattern of reading the other values (i.e. `ae-exec`, `ae->recipient`, etc), a few lines above. I think the check is on the value itself, i.e. if the strdupz returns something but its empty, set the pointer to null... ?

thiagoftsm: Hello @MrZammler ,

The `strdupz` won't return `NULL`, because the `strdupz()` will call  `fatal()` that reports the problem and finally calls  `netdata_cleanup_and_exit(1);` . 

Our `NAMEz` functions differ of the traditional standard library, because it exits when memory is not allocated, while standard library returns NULL.

health/health_log.c
thiagoftsm: Your initial thought was not completely wrong when we take a look for the majority of error logs stored, but there are few exceptions that force us to use a bigger value, 33 makes more sense here.

health/health_log.c
stelfrag: Its `!*ae->class` (notice the *) -- checking for a zero length string

MrZammler:
Closing without merging, will create a clean PR

PR #10853, opened 2021-03-24
health: improve alarms infos

ilyam8:
##### Summary

First step in improving alarms info (description) direction. 

##### Component Name

`health`

##### Test Plan

- [x] install this PR, ensure there is no syntax errors and all alarms are loaded and working.

### DIFF
--- a/health/health.d/memcached.conf
+++ b/health/health.d/memcached.conf
@@ -48,6 +48,6 @@ template: memcached_out_of_cache_space_time
     warn: $this > 0 and $this < (($status >= $WARNING)  ? (48) : (8))
     crit: $this > 0 and $this < (($status == $CRITICAL) ? (24) : (2))
    delay: down 15m multiplier 1.5 max 1h
-    info: estimated time the cache will run out of space, \
-          if the system continues to add data with the rate of the last hour
+    info: estimated time the cache will run out of space \
+          if the system continues to add data at the same rate as the past hour
       to: dba
--- a/health/health.d/memcached.conf
+++ b/health/health.d/memcached.conf
@@ -23,7 +23,7 @@ template: memcached_cache_memory_usage
     warn: $this > (($status >= $WARNING)  ? (70) : (80))
     crit: $this > (($status == $CRITICAL) ? (80) : (90))
    delay: up 0 down 15m multiplier 1.5 max 1h
-    info: current cache memory usage
+    info: cache memory utilization
       to: dba
 
 
@@ -35,7 +35,7 @@ template: memcached_cache_fill_rate
     calc: ($this - $available) / (($now - $after) / 3600)
    units: KB/hour
    every: 1m
-    info: average rate the cache fills up (positive), or frees up (negative) space, for the last hour
+    info: average rate the cache fills up (positive), or frees up (negative) space over the last hour
 
 
 # find the hours remaining until memcached cache is full
@@ -48,5 +48,6 @@ template: memcached_out_of_cache_space_time
     warn: $this > 0 and $this < (($status >= $WARNING)  ? (48) : (8))
     crit: $this > 0 and $this < (($status == $CRITICAL) ? (24) : (2))
    delay: down 15m multiplier 1.5 max 1h
-    info: estimated time the cache will run out of space, if the system continues to add data with the rate of the last hour
+    info: estimated time the cache will run out of space, \
+          if the system continues to add data with the rate of the last hour
       to: dba
--- a/health/health.d/memory.conf
+++ b/health/health.d/memory.conf
@@ -10,7 +10,7 @@
    every: 1m
     warn: $this > 0
    delay: down 1h multiplier 1.5 max 1h
-    info: number of ECC correctable errors in the last hour
+    info: number of ECC correctable errors in the last 10 minutes
       to: sysadmin
 
    alarm: 1hour_ecc_memory_uncorrectable
--- a/health/health.d/memory.conf
+++ b/health/health.d/memory.conf
@@ -22,7 +22,7 @@
    every: 1m
     crit: $this > 0
    delay: down 1h multiplier 1.5 max 1h
-    info: number of ECC uncorrectable errors in the last hour
+    info: number of ECC uncorrectable errors in the last 10 minutes
       to: sysadmin
 
    alarm: 1hour_memory_hw_corrupted
--- a/health/health.d/memory.conf
+++ b/health/health.d/memory.conf
@@ -10,7 +10,7 @@
    every: 1m
     warn: $this > 0
    delay: down 1h multiplier 1.5 max 1h
-    info: number of ECC correctable errors during the last hour
+    info: number of ECC correctable errors in the last hour
       to: sysadmin
 
    alarm: 1hour_ecc_memory_uncorrectable
@@ -22,7 +22,7 @@
    every: 1m
     crit: $this > 0
    delay: down 1h multiplier 1.5 max 1h
-    info: number of ECC uncorrectable errors during the last hour
+    info: number of ECC uncorrectable errors in the last hour
       to: sysadmin
 
    alarm: 1hour_memory_hw_corrupted
--- a/health/health.d/net.conf
+++ b/health/health.d/net.conf
@@ -12,7 +12,7 @@
      calc: ( $nic_speed_max > 0 ) ? ( $nic_speed_max) : ( nan )
     units: Mbit
     every: 10s
-     info: The current speed of the physical network interface
+     info: network interface current speed
 
  template: 1m_received_traffic_overflow
        on: net.net
@@ -25,7 +25,7 @@
     every: 10s
      warn: $this > (($status >= $WARNING)  ? (85) : (90))
     delay: up 1m down 1m multiplier 1.5 max 1h
-     info: interface received bandwidth usage over net device speed max
+     info: average inbound utilization for the network interface over the last minute
        to: sysadmin
 
  template: 1m_sent_traffic_overflow
@@ -39,7 +39,7 @@
     every: 10s
      warn: $this > (($status >= $WARNING)  ? (85) : (90))
     delay: up 1m down 1m multiplier 1.5 max 1h
-     info: interface sent bandwidth usage over net device speed max
+     info: average outbound utilization for the network interface over the last minute
        to: sysadmin
 
 # -----------------------------------------------------------------------------
@@ -60,7 +60,7 @@ families: !net* *
   lookup: sum -10m unaligned absolute of inbound
    units: packets
    every: 1m
-    info: interface inbound dropped packets in the last 10 minutes
+    info: number of inbound dropped packets for the network interface in the last 10 minutes
 
 template: outbound_packets_dropped
       on: net.drops
@@ -70,7 +70,7 @@ families: !net* *
   lookup: sum -10m unaligned absolute of outbound
    units: packets
    every: 1m
-    info: interface outbound dropped packets in the last 10 minutes
+    info: number of outbound dropped packets for the network interface in the last 10 minutes
 
 template: inbound_packets_dropped_ratio
       on: net.packets
@@ -83,7 +83,7 @@ families: !net* !wl* *
    every: 1m
     warn: $this >= 2
    delay: up 1m down 1h multiplier 1.5 max 2h
-    info: the ratio of inbound dropped packets vs the total number of received packets of the network interface, during the last 10 minutes
+    info: ratio of inbound dropped packets for the network interface over the last 10 minutes
       to: sysadmin
 
 template: outbound_packets_dropped_ratio
@@ -97,7 +97,7 @@ families: !net* !wl* *
    every: 1m
     warn: $this >= 2
    delay: up 1m down 1h multiplier 1.5 max 2h
-    info: the ratio of outbound dropped packets vs the total number of sent packets of the network interface, during the last 10 minutes
+    info: ratio of outbound dropped packets for the network interface over the last 10 minutes
       to: sysadmin
 
 template: wifi_inbound_packets_dropped_ratio
@@ -111,7 +111,7 @@ families: wl*
    every: 1m
     warn: $this >= 10
    delay: up 1m down 1h multiplier 1.5 max 2h
-    info: the ratio of inbound dropped packets vs the total number of received packets of the network interface, during the last 10 minutes
+    info: ratio of inbound dropped packets for the network interface over the last 10 minutes
       to: sysadmin
 
 template: wifi_outbound_packets_dropped_ratio
@@ -125,7 +125,7 @@ families: wl*
    every: 1m
     warn: $this >= 10
    delay: up 1m down 1h multiplier 1.5 max 2h
-    info: the ratio of outbound dropped packets vs the total number of sent packets of the network interface, during the last 10 minutes
+    info: ratio of outbound dropped packets for the network interface over the last 10 minutes
       to: sysadmin
 
 # -----------------------------------------------------------------------------
@@ -141,7 +141,7 @@ families: *
    every: 1m
     warn: $this >= 5
    delay: down 1h multiplier 1.5 max 2h
-    info: interface inbound errors in the last 10 minutes
+    info: number of inbound errors for the network interface in the last 10 minutes
       to: sysadmin
 
 template: interface_outbound_errors
@@ -154,7 +154,7 @@ families: *
    every: 1m
     warn: $this >= 5
    delay: down 1h multiplier 1.5 max 2h
-    info: interface outbound errors in the last 10 minutes
+    info: number of outbound errors for the network interface in the last 10 minutes
       to: sysadmin
 
 # -----------------------------------------------------------------------------
@@ -175,7 +175,7 @@ families: *
    every: 1m
     warn: $this > 0
    delay: down 1h multiplier 1.5 max 2h
-    info: interface fifo errors in the last 10 minutes
+    info: number of FIFO errors for the network interface in the last 10 minutes
       to: sysadmin
 
 # -----------------------------------------------------------------------------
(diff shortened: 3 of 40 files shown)


ilyam8:
@vlvkobal @joelhans i feel like there are 2 things we want to describe:
 1. what's that alarm/value (lookup=>calc value result)
 2. why the alarm is active (warn/crit conditions)

[`disk_space_usage`](https://github.com/netdata/netdata/blob/aa390501a36a1433fbbd7025e349949e887c9023/health/health.d/disks.conf#L12-L24) example:
 - average disk space utilization
 - disk is almost full

At the moment we use info for 1st. I think @dim08 was mainly referring to 2nd, which makes sense. 

What we can do is:
 - use `info` field for 2nd and let user to extract 1st from lookup/calc (sometimes it is not easy). Try to improve alarms/templates names, we can make them more descriptive.
 - use `info` for both 1nd and 2nd. Downside of this is having infos with a lot of text and, to be honest, i don't really care if that is average for the last 5 or 10 minutes, or rate, or whatever. If the alarm is up, what i care is, `disk is almost full` sentence.


What do you think, guys @joelhans @vlvkobal

vlvkobal (changes requested):

health/health.d/pihole.conf
```diff
@@ -36,7 +36,7 @@ template: pihole_blocklist_last_update
     calc: $ago
     warn: $this > 60 * 60 * 24 * 8
     crit: $this > 60 * 60 * 24 * 8 * 2
-    info: blocklist last update time
+    info: gravity.list (blocklist) file last update seconds ago
```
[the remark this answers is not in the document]
ilyam8: the vaule actually shows not timestamp, but amount of seconds ago

health/health.d/memory.conf
```diff
@@ -22,7 +22,7 @@
    every: 1m
     crit: $this > 0
    delay: down 1h multiplier 1.5 max 1h
-    info: number of ECC uncorrectable errors during the last hour
+    info: number of ECC uncorrectable errors in the last hour
```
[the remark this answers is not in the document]
ilyam8: ahha the alarm name is `1hour_ecc_memory_uncorrectable

health/health.d/anomalies.conf
```diff
@@ -5,7 +5,7 @@ template: anomalies_anomaly_probabilities
   lookup: average -2m foreach *
    every: 1m
     warn: $this > 50
-    info: average anomaly probability > 50% for last 2 minutes
+    info: average anomaly probability over the last 2 minutes
```
thiagoftsm, line 8: Probably distribution have different mean/averag, I think to hide the 50% can conduct to different conclusions.  @vlvkobal and @andrewm4894 am I missing something?

health/health.d/backend.conf
```diff
@@ -6,7 +6,7 @@
    every: 1m
     warn: $this > 0
    delay: down 5m multiplier 1.5 max 1h
-    info: The backends subsystem is deprecated and will be removed soon. Migrate your configuration to exporting.conf.
+    info: the backends subsystem is deprecated and will be removed soon. Migrate your configuration to exporting.conf.
```
thiagoftsm, line 9: I prefer the uppercase here, because it is the beginning of the sentence.

health/health.d/dockerd.conf
```diff
@@ -4,5 +4,5 @@ template: docker_unhealthy_containers
    every: 10s
   lookup: average -10s
     crit: $this > 0
-    info: number of unhealthy containers
+    info: average number of unhealthy docker containers over the last 10 seconds
```
thiagoftsm, line 7: Should we add an explanation for the `unhealthy` here? :thinking: Because we are sending message because the system is with some problem, so it is implicit that the system is not health.

health/health.d/anomalies.conf
ilyam8: If the alarm is up there is value

`WARN | average anomaly probability over the last 2 minutes| 73`

health/health.d/backend.conf
ilyam8: i see no we follow that rule anywhere. Also there is no `.` at the end of infos.

health/health.d/dockerd.conf
ilyam8: It is https://docs.docker.com/engine/reference/builder/#healthcheck

> When a container has a healthcheck specified, it has a health status in addition to its normal status. This status is initially starting. Whenever a health check passes, it becomes healthy (whatever state it was previously in). After a certain number of consecutive failures, it becomes unhealthy.

health/health.d/web_log.conf
```diff
@@ -165,15 +165,15 @@ families: *
   lookup: average -5m at -5m unaligned of successful_requests
    units: requests/s
    every: 30s
-    info: average rate of successful HTTP requests over the last 5 minutes
+    info: average number of successful HTTP responses for the 5 minutes starting 10 minutes ago
```
vlvkobal, line 168:
```suggestion
    info: average number of successful HTTP requests for the 5 minutes starting 10 minutes ago
```

health/health.d/web_log.conf
```diff
@@ -165,15 +165,15 @@ families: *
...
 
 template: 5m_successful
       on: web_log.response_statuses
 families: *
   lookup: average -5m unaligned of successful_requests
    units: requests/s
    every: 30s
-    info: average successful HTTP requests over the last 5 minutes
+    info: average number of successful HTTP responses over the last 5 minutes
```
vlvkobal, line 176:
```suggestion
    info: average number of successful HTTP requests over the last 5 minutes
```

health/health.d/web_log.conf
```diff
@@ -185,7 +185,7 @@ families: *
     crit: ($5m_successful_old > 120) ? ($this > 400 OR $this < 25) : (0)
    delay: down 15m multiplier 1.5 max 1h
 options: no-clear-notification
-    info: the percentage of successful web requests over the last 5 minutes, \
+    info: ratio of successful HTTP responses over the last 5 minutes, \
```
vlvkobal, line 188:
```suggestion
    info: ratio of successful HTTP requests over the last 5 minutes, \
```

health/health.d/web_log.conf
```diff
@@ -359,15 +359,15 @@ families: *
   lookup: average -5m at -5m unaligned of success
    units: requests/s
    every: 30s
-    info: average rate of successful HTTP requests over the last 5 minutes
+    info: average number of successful HTTP responses for the 5 minutes starting 10 minutes ago
```
vlvkobal, line 362:
```suggestion
    info: average number of successful HTTP requests for the 5 minutes starting 10 minutes ago
```

health/health.d/web_log.conf
```diff
@@ -359,15 +359,15 @@ families: *
...
 
 template: web_log_5m_successful
       on: web_log.type_requests
 families: *
   lookup: average -5m unaligned of success
    units: requests/s
    every: 30s
-    info: average successful HTTP requests over the last 5 minutes
+    info: average number of successful HTTP responses over the last 5 minutes
```
vlvkobal, line 370:
```suggestion
    info: average number of successful HTTP requests over the last 5 minutes
```

health/health.d/web_log.conf
```diff
@@ -379,7 +379,7 @@ families: *
     crit: ($web_log_5m_successful_old > 120) ? ($this > 400 OR $this < 25) : (0)
    delay: down 15m multiplier 1.5 max 1h
 options: no-clear-notification
-    info: the percentage of successful web requests over the last 5 minutes, \
+    info: ratio of successful HTTP responses over over the last 5 minutes, \
```
vlvkobal, line 382:
```suggestion
    info: ratio of successful HTTP requests over over the last 5 minutes, \
```

vlvkobal (changes requested):

vlvkobal (dismissed):

joelhans:
> i don't really care if that is average for the last 5 or 10 minutes, or rate, or whatever. If the alarm is up, what i care is, disk is almost full sentence.

@ilyam8 I think that makes a lot of sense. I think it's more valuable if the `info` line helps explain why the user should care about the alarm that's just gone off.

> use info field for 2nd and let user to extract 1st from lookup/calc (sometimes it is not easy)

Don't our alarm notifications always show the current value? If so, we don't need to include the value calculation in `info`.

ilyam8:
> Don't our alarm notifications always show the current value?

It is just a value (e.g _44.4_). Almost always it is [lookup](https://learn.netdata.cloud/docs/agent/health/reference#alarm-line-lookup) and/or [calc](https://learn.netdata.cloud/docs/agent/health/reference#alarm-line-calc) (number/average/ratio/etc of something over some period of time).

joelhans:
> It is just a value (e.g _44.4_). Almost always it is [lookup](https://learn.netdata.cloud/docs/agent/health/reference#alarm-line-lookup) and/or [calc](https://learn.netdata.cloud/docs/agent/health/reference#alarm-line-calc) (number/average/ratio/etc of something over some period of time).

image

Right. So, pretend the alarm in this image had an `info` field. It wouldn't make sense to repeat the lookup/calc in `info`, since it's already displayed in that badge on the left.

I _think_ we agree on this, since I see you already removed most of the variables in `info`, just making sure.

thiagoftsm (dismissed):

health/health.d/dockerd.conf
ilyam8: i mean if you want to suggest something, please do and i submit changes 😉

ilyam8:
Well, that is `1.` in https://github.com/netdata/netdata/pull/10853#issuecomment-806614579

We assume user reads and understands lookup + calc. However, it is not a rare case when, let's be honest, it is pretty hard to follow it 😅  And you need to spend a decent amount of time to understand it.

Example. 

Some variable * other variable / another variable. Wait what? 😄 

https://github.com/netdata/netdata/blob/f3881e1cd1d2e67275ea4a66760c742b14ac430a/health/health.d/net.conf#L36-L42


So, we don't repeat it, but explain it in human-readable format.

ilyam8:
> I think it's more valuable if the info line helps explain why the user should care about the alarm that's just gone off.

Yes, we will work in that direction.

joelhans:
> So, we don't repeat it, but explain it in human-readable format.

I get it now: `info` explains the `calc`/`lookup` in a human-readable format.

Given that, I agree should probably have a second `description` field that explains _why_ someone should care, or propose what they do next. But we should only add a `description` if we have valid, useful information to include, and not just fill out every one with some random bit of templatized text.

I think a combination of `info` and `description` like this would be pretty valuable:

```
template: bind_rndc_stats_file_size
      on: bind_rndc.stats_size
   units: megabytes
   every: 60
    calc: $stats_size
    warn: $this > 512
    crit: $this > 1024
    info: BIND statistics-file size
      to: sysadmin
description: The BIND statistics-file is more than 512Mb. To reduce its size and stop this alert, consider creating a logrotate job for the file, which exists in /var/named/data/, or with chroot enabled, in /var/named/chroot/var/named/data.
```

Now, how do we go about implementing that...

ilyam8:
Yeah, explanation of the value is needed, because it is not only the web GUI, where you have db lookup/calc, but notifications too. I think in a lot notifications methods we don't send lookup and calc, but value/chart/info only.

health/health.d/memcached.conf
```diff
@@ -48,5 +48,6 @@ template: memcached_out_of_cache_space_time
     warn: $this > 0 and $this < (($status >= $WARNING)  ? (48) : (8))
     crit: $this > 0 and $this < (($status == $CRITICAL) ? (24) : (2))
    delay: down 15m multiplier 1.5 max 1h
-    info: estimated time the cache will run out of space, if the system continues to add data with the rate of the last hour
+    info: estimated time the cache will run out of space, \
+          if the system continues to add data with the rate of the last hour
```
joelhans, line 52:
```suggestion
    info: estimated time the cache will run out of space \
          if the system continues to add data at the same rate as the past hour
```

health/health.d/net.conf
```diff
@@ -209,5 +209,6 @@ families: *
     warn: $this > (($status >= $WARNING)?(200):(5000))
     crit: $this > (($status == $CRITICAL)?(5000):(6000))
  options: no-clear-notification
-    info: the % of the rate of received packets in the last 10 seconds, compared to the rate of the last minute (clear notification for this alarm will not be sent)
+    info: ratio of average number of received packets for the network interface over the last 10 seconds, \
+          compared to the rate of the last minute
```
joelhans, line 213:
```suggestion
    info: ratio of average number of received packets for the network interface over the last 10 seconds, \
          compared to the rate over the last minute
```

health/health.d/ram.conf
```diff
@@ -7,7 +7,8 @@
    hosts: *
     calc: ($zfs.arc_size.arcsz = nan)?(0):($zfs.arc_size.arcsz - $zfs.arc_size.min)
    every: 10s
-    info: the amount of memory that is reported as used, but it is actually capable for resizing itself based on the system needs (eg. ZFS ARC)
+    info: amount of memory that is reported as used, \
+          but it is actually capable for resizing itself based on the system needs (eg. ZFS ARC)
```
joelhans, line 11: Do we know what "it" refers to in the second line? What is actually capable of resizing itself? 

```suggestion
    info: amount of memory reported as used, \
          but it is actually capable for resizing itself based on the system needs (eg. ZFS ARC)
```

health/health.d/softnet.conf
```diff
@@ -12,7 +12,8 @@
    every: 10s
     warn: $this > (($status >= $WARNING) ? (0) : (10))
    delay: down 1h multiplier 1.5 max 2h
-    info: average number of packets dropped in the last 1min, because sysctl net.core.netdev_max_backlog was exceeded (this can be a cause for dropped packets)
+    info: average number of dropped packets in the last minute \
+          due to sysctl net.core.netdev_max_backlog was exceeded
```
joelhans, line 16:
```suggestion
    info: average number of dropped packets in the last minute \
          due to exceeded net.core.netdev_max_backlog
```

health/health.d/softnet.conf
```diff
@@ -36,5 +39,7 @@
    every: 10s
     warn: $this > (($status >= $WARNING) ? (0) : (10))
    delay: down 1h multiplier 1.5 max 2h
-    info: average number of drops in the last 1min, because sysctl net.route.netisr_maxqlen was exceeded (this can be a cause for dropped packets)
+    info: average number of drops in the last minute \
+          due to sysctl net.route.netisr_maxqlen was exceeded \
+          (this can be a cause for dropped packets)
```
joelhans, line 44:
```suggestion
    info: average number of drops in the last minute \
          due to exceeded sysctl net.route.netisr_maxqlen \
          (this can be a cause for dropped packets)
```

health/health.d/tcp_resets.conf
```diff
@@ -23,7 +23,10 @@
     warn: $this > ((($1m_ipv4_tcp_resets_sent < 5)?(5):($1m_ipv4_tcp_resets_sent)) * (($status >= $WARNING)  ? (1) : (20)))
    delay: up 20s down 60m multiplier 1.2 max 2h
  options: no-clear-notification
-    info: average TCP RESETS this host is sending, over the last 10 seconds (this can be an indication that a port scan is made, or that a service running on this host has crashed; clear notification for this alarm will not be sent)
+    info: average number of sent TCP RESETS over the last 10 seconds \
+          (this can be an indication that a port scan is made, \
+          or that a service running on this host has crashed; \
+          clear notification for this alarm will not be sent)
```
joelhans, line 29:
```suggestion
    info: average number of sent TCP RESETS over the last 10 seconds. \
          This can be indicate a port scan, \
          or that a service running on this host has crashed. \
          Netdata will not send a clear notification for this alarm.
```

health/health.d/tcp_resets.conf
```diff
@@ -48,5 +51,7 @@
     warn: $this > ((($1m_ipv4_tcp_resets_received < 5)?(5):($1m_ipv4_tcp_resets_received)) * (($status >= $WARNING)  ? (1) : (10)))
    delay: up 20s down 60m multiplier 1.2 max 2h
  options: no-clear-notification
-    info: average TCP RESETS this host is receiving, over the last 10 seconds (this can be an indication that a service this host needs, has crashed; clear notification for this alarm will not be sent)
+    info: average number of received TCP RESETS over the last 10 seconds \
+          (this can be an indication that a service this host needs, has crashed; \
+          clear notification for this alarm will not be sent)
```
joelhans, line 56:
```suggestion
    info: average number of received TCP RESETS over the last 10 seconds. \
          This can be an indication that a service this host needs has crashed. \
          Netdata will not send a clear notification for this alarm.
```

joelhans (changes requested):
Overall these all look like smart changes. Hoping we can do something about this `description` where we can have richer content.

health/health.d/ram.conf
ilyam8: > Do we know what "it" refers to

[you'd better dont touch this topic](https://static.wikia.nocookie.net/stephenking/images/b/b5/PennywiseTheClown2017.png/revision/latest?cb=20191128001428)

jokes aside, ARC stands for Adaptive Replacement Cache and we treat it (size minus min, which is amount of memory needed by ZFS thing itself) as available memory when calculating `ram_in_use` alarm.

ilyam8: To be honest i don't really know why this alarm name is not `zfs_arc_size`. Perhaps the idea was to have a generic name and add to it different stuff which is used, but reclaimable.

vlvkobal (approved):

MrZammler (approved):

ilyam8:
Did force-push because of stuck GHA jobs

ilyam8:
@joelhans my opinion on `info`/`description` terms:  using `description` for the value and `info` for the event looks better for me then vice versa. But, on the other hand, they are ~ synonyms...


Which one looks/feels better for you?

Example:

| Value  | Description                             | Info                      |
| ------ | --------------------------------------- | ------------------------- |
| 2      | number of disks in the down state       | the array X is degraded   |


vs

| Value  | Info                                    | Description               |
| ------ | --------------------------------------- | ------------------------- |
| 2      | number of disks in the down state       | the array X is degraded   |

PR #10856, opened 2021-03-25
Dashboard version 2.17.0

allelos:
Agent redesgin

Top bar, side bar and navigation redesign according to desings [here](https://www.figma.com/file/RgebwHKHvTCAlTmIxLsb6l/New-Cloud-and-Agent-navigation-%231629?node-id=1015%3A45)

Sidebar:
- Documentation button now lives on the vertical bar
- Settings button now lives on the vertical bar
- Introduced a UserSettings button that renders sign-in / sign-up option when not signed-in, sign out option when signed in along with an operational status option. 

Side panel:
- Nodes are now searchable when there are more than 4 present
- Each node now shows a status badge

Topbar:
- Now includes a light bulb icon which opens a modal that displays agent news & features. Icon is colored when it has unread entries 
- shows number of alerts with corresponding coloring
- includes the date time picker in horizontal format

image

**Agent News & Features Announcements**

**Designs**:
https://www.figma.com/file/RgebwHKHvTCAlTmIxLsb6l/New-Cloud-and-Agent-navigation-%231629?node-id=535%3A935

**Unseen**:
image

**Seen**:
image

image

**Minor**:
- Fixed word breaking on alarms modal. https://github.com/netdata/dashboard/pull/267


ilyam8:
> shows number of alerts with corresponding coloring

A little comment from me. I think it would be better if we dim corresponding counter if there are no active alarms. Having it always red makes me feeling that there is something wrong, makes me draw attention on it.  cc @dim08

allelos:
> > shows number of alerts with corresponding coloring
> 
> A little comment from me. I think it would be better if we dim corresponding counter if there are no active alarms. Having it always red makes me feeling that there is something wrong, makes me draw attention on it. cc @dim08

@manos-saratsis

ilyam8:
Few problems i spotted:
 - `sign in to cloud` button is not clickable
 - `sign in to cloud` button place, is it supposed to be there?
 - sidebar _jumps_ up and down every time i click anywhere


Screenshot 2021-03-26 at 14 57 29

ilyam8:
@allelos there is a problem with checksum validation (build ci checks fail for this reason)

```cmd
[/home/ilyam/src/nd_allelos:dashboard-v2.15.1-alpha]# curl -q -sSL --connect-timeout 10 --retry 3 --output /tmp/netdata-dashboard-7dlrAf/dashboard.tar.gz https://github.com/netdata/dashboard/releases/download/v2.15.3-alpha/dashboard.tar.gz
 OK

dashboard.tar.gz: FAILED
sha256sum: WARNING: 1 computed checksum did NOT match
 FAILED  dashboard files checksum validation failed.

 FAILED  Unable to fetch React dashboard. The install process will continue, but you will not be able to use the new dashboard.
```

ilyam8:
@allelos  i installed 2.15.3 and the sidebar problem is fixed 🎉  But, not clickable sign-in-to-cloud is not. It is clickable only for few seconds after refreshing the page.

> has an online status badge according to browsers connectivity

I see no online status

allelos:
> @allelos i installed 2.15.3 and the sidebar problem is fixed 🎉 But, not clickable sign-in-to-cloud is not. It is clickable only for few seconds after refreshing the page.
> 
> > has an online status badge according to browsers connectivity
> 
> I see no online status

@ilyam8 Many thanks for being on top of it and testing every time 🙏 

The online status as well as the Netdata logo on the top bar will be removed as you spotted. I will update the description when finished. I am keeping this branch as an alpha, let's say branch, so guys can test the progress.

As far as the sign in. The sign in doesn't have to do with the dashboard per say (in this case), but with the iframes inside the dashboard.

I am guessing the iframes on your build are hitting app.netdata.cloud where I haven't releaesed the new iframes version. If you want you can check by pointing the agent iframes to staging.netdata.cloud.

ilyam8:
Well, i am fast enough i can click on the button and it works.

Btw i see that button even if i've built netdata with `--disable-cloud` flag. Is it supposed to be so?

Just imagine you deliberately disable cloud finctionality, but that `sing in to cloud` button is on the dashboard! (and i m not sure about its place, it always covers some chart).

Screenshot 2021-03-29 at 18 19 52

ilyam8:
@allelos i tested the latest and overall looks good to me. There is no `sign in to cloud` if i compile netdata w/o cloud support 👍 

Can mention only two things:
 - `sign in to cloud` button place (we discussed it on slack)
 - perhaps it is me, but i'd prefer to open sing ing page in a separate browser tab (without leaving the current page)

Yannis-San:
Design QA

**Space nav bottom icons**
- less space between the icons
- hover and active state to be added

**Left nav main nav (open)**
- C2A button on left panel differs to design (should be full width)
- background colour on bottom panel (sign in/up) is not white but #536775
- Distance on copy too big
- copy at the bottom panel should be left align (both sign in/up)
- check the font type and size in the bottom panel
- When signed out, no need to show Operational Status of Cloud if you click on the avatar

**Top bar**
- Update icon does not change colour when new version is updated

Will add some more tommorow @allelos @manos-saratsis

allelos:
Just some questions and notes @manos-saratsis @Yannis-San 

* space between icons is the same as the cloud. Are we sure we want to make it less `20px` and not keep it consistent?

* By hover and active states I believe we are referring to the circle enclosing the icon, right? I believe we don't have this anywhere else, neither cloud or agent, are we sure want to add it here?
image

* Background color on bottom panel (sign in / sign up) is white, because `#536775` is nowhere found in our color palette (netdata-ui) and I would like to avoid hardcoding colors. Can we find a relatively close one for now, from our palette, and change to this one later when we update to new colors? If not I will hard code it.

* The font type should be consistent across, since it's directly from our ui kit.

Yannis-San:
> Just some questions and notes @manos-saratsis @Yannis-San
> 
> * space between icons is the same as the cloud. Are we sure we want to make it less (`20px`) and not keep it consistent? **Leave as is**
> * By hover and active states I believe we are referring to the circle enclosing the icon, right? I believe we don't have this anywhere else, neither cloud or dashboard, are we sure want to add it here?**Leave as is**
>   image
> * Background color on bottom panel (sign in / sign up) is white, because `#536775` is nowhere found in our color palette (netdata-ui) and I would like to avoid hardcoding colors. Can we find a relatively close one for now, from our palette, and change to this one later when we update to new colors? If not I will hard code it. **Send you a new one**
> * It would be nice if we could see the correct distance on the copy on the design, instead of being a complete block.
>   image**16px**
> * The font type should be consistent across, since it's directly from our ui kit.

packaging/dashboard.version
```diff
@@ -1 +1 @@
-v2.13.28_
+v2.17.0-beta
```
ilyam8, line 1: still beta

ilyam8: I see [v2.17.0 is released](https://github.com/netdata/dashboard/releases/tag/v2.17.0)

allelos: Oops! lost in rebase :P

ilyam8 (approved):

vlvkobal (approved):

PR #10857, opened 2021-03-25
Alerts template styling fixes

jacekkolasa:
##### Summary
- fixed most of the styling problems, especially for outlook (online) and thunderbird
- updated colors based on status (both for active alarms list and the main part of template)

Most of the differences between email and the design is caused by the fact, that it's hard to make sure the proper font is loaded. Most of the time the needed one (IBM Plex Sans) is not loaded, and Helvetica is used instead. Sometimes it results in *bold* fonts (font-weight: 500) to look differently than in design.

obraz

second part:

obraz

##### Component Name
Health / alarm template

### DIFF
--- a/health/notifications/alarm-notify.sh.in
+++ b/health/notifications/alarm-notify.sh.in
@@ -2664,24 +2664,36 @@ if [ -n "$total_warn_alarms" ]; then
                   >
 
                     <table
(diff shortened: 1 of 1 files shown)


thiagoftsm (approved):
I tested now and I can receive notifications on methods that were not changed. The `shellcheck` is also not reporting problems. And for methods that were modified, there are improvements.
Finally great work @jacekkolasa ! :)

MrZammler (approved):

PR #10859, opened 2021-03-25
ACLK Passwd endpoint update

underhood:
##### Summary
Updates password endpoint to match specs of New Architecture.

- Parse response of password endpoint in new format
- Use topics given to us from cloud instead of ones hardcoded in the agent.

##### Component Name
ACLK
##### Test Plan
Connects to cloud without issue,

### DIFF
--- a/aclk/aclk_util.c
+++ b/aclk/aclk_util.c
@@ -70,52 +70,156 @@ int aclk_get_conv_log_next()
 #define ACLK_TOPIC_PREFIX "/agent/"
 
 struct aclk_topic {
-    const char *topic_suffix;
+    enum aclk_topics topic_id;
+    // as received from cloud - we keep this for
+    // eventual topic list update when claim_id changes
+    char *topic_recvd;
+    // constructed topic
     char *topic;
 };
 
 // This helps to cache finalized topics (assembled with claim_id)
 // to not have to alloc or create buffer and construct topic every
 // time message is sent as in old ACLK
-static struct aclk_topic aclk_topic_cache[] = {
-    { .topic_suffix = "outbound/meta",   .topic = NULL }, // ACLK_TOPICID_CHART
-    { .topic_suffix = "outbound/alarms", .topic = NULL }, // ACLK_TOPICID_ALARMS
-    { .topic_suffix = "outbound/meta",   .topic = NULL }, // ACLK_TOPICID_METADATA
-    { .topic_suffix = "inbound/cmd",     .topic = NULL }, // ACLK_TOPICID_COMMAND
-    { .topic_suffix = NULL,              .topic = NULL }
-};
+static struct aclk_topic **aclk_topic_cache = NULL;
+static size_t aclk_topic_cache_items = 0;
 
 void free_topic_cache(void)
 {
-    struct aclk_topic *tc = aclk_topic_cache;
-    while (tc->topic_suffix) {
-        if (tc->topic) {
-            freez(tc->topic);
-            tc->topic = NULL;
+    if (aclk_topic_cache) {
+        for (size_t i = 0; i < aclk_topic_cache_items; i++) {
+            freez(aclk_topic_cache[i]->topic);
+            freez(aclk_topic_cache[i]->topic_recvd);
+            freez(aclk_topic_cache[i]);
         }
-        tc++;
+        freez(aclk_topic_cache);
+        aclk_topic_cache = NULL;
+        aclk_topic_cache_items = 0;
     }
 }
 
-static inline void generate_topic_cache(void)
-{
-    struct aclk_topic *tc = aclk_topic_cache;
-    char *ptr;
-    if (unlikely(!tc->topic)) {
-        rrdhost_aclk_state_lock(localhost);
-        while(tc->topic_suffix) {
-            tc->topic = mallocz(strlen(ACLK_TOPIC_PREFIX) + (UUID_STR_LEN - 1) + 2 /* '/' and \0 */ + strlen(tc->topic_suffix));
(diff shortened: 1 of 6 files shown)


vkalintiris (commented):
Overall, LGTM. Let me know your thoughts on the partial-updates comment I've made inline and we are good to go.

aclk/aclk_util.c
```diff
@@ -70,52 +70,156 @@ int aclk_get_conv_log_next()
...
+        if (json_object_get_type(obj) != json_type_object) {
+            error("expected json_type_object");
+            return 1;
+        }
+        aclk_topic_cache[i] = callocz(1, sizeof(struct aclk_topic));
+        if (topic_cache_add_topic(obj, aclk_topic_cache[i])) {
+            error("failed to parse topic @idx=%d", (int)i);
+            return 1;
+        }
+    }
```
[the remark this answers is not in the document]
underhood: I thought of that. I didn't do it in that way because if there was error you should not attempt connection at all and therefore not use any topic

aclk/aclk_util.c
```diff
@@ -70,52 +70,156 @@ int aclk_get_conv_log_next()
...
+    char *replace_tag = strstr(topic->topic_recvd, CLAIM_ID_REPLACE_TAG);
+    if (!replace_tag)
+        return;
+
+    topic->topic = mallocz(strlen(topic->topic_recvd) + 1 - strlen(CLAIM_ID_REPLACE_TAG) + strlen(localhost->aclk_state.claimed_id));
+    dest = topic->topic;
+    memcpy(topic->topic, topic->topic_recvd, replace_tag - topic->topic_recvd);
+    dest = topic->topic + (replace_tag - topic->topic_recvd);
+    // TODO check claimed_id NULL
+    memcpy(dest, localhost->aclk_state.claimed_id, strlen(localhost->aclk_state.claimed_id));
```
stelfrag, line 140: Probably need to lock here as in function `is_agent_claimed` -- `rrdhost_aclk_state_lock(localhost);`

aclk/aclk_util.c
```diff
@@ -125,9 +229,14 @@ static inline void generate_topic_cache(void)
  */
 const char *aclk_get_topic(enum aclk_topics topic)
 {
-    generate_topic_cache();
+    if (!aclk_topic_cache)
+        fatal("Topic cache not initialized");
```
stelfrag, line 233: we should fail the aclk connection not the agent

underhood: true, will change in a moment

aclk/aclk_util.c
```diff
@@ -70,52 +70,156 @@ int aclk_get_conv_log_next()
...
+}
+
+#define CLAIM_ID_REPLACE_TAG "#{claim_id}"
+static void topic_generate_final(struct aclk_topic *topic) {
+    char *dest;
+    char *replace_tag = strstr(topic->topic_recvd, CLAIM_ID_REPLACE_TAG);
+    if (!replace_tag)
+        return;
+
+    topic->topic = mallocz(strlen(topic->topic_recvd) + 1 - strlen(CLAIM_ID_REPLACE_TAG) + strlen(localhost->aclk_state.claimed_id));
```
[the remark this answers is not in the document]
underhood: renamed parameter to `t` for increased clarity

aclk/aclk_util.c
underhood: fixed

aclk/aclk_util.c
```diff
@@ -70,52 +70,156 @@ int aclk_get_conv_log_next()
...
+    json_object *obj;
+
+    size_t array_size = json_object_array_length(json);
+    if (!array_size) {
+        error("Empty topic list!");
+        return 1;
+    }
+
+    if (aclk_topic_cache)
+        free_topic_cache();
```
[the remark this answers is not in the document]
underhood: well empty topic list from cloud is an error in any case... but we could free the topic cache before. Not sure of what benefit that would give us

aclk/aclk_util.c
```diff
@@ -70,52 +70,189 @@ int aclk_get_conv_log_next()
...
+
+    rrdhost_aclk_state_lock(localhost);
+    if (unlikely(!localhost->aclk_state.claimed_id)) {
+        error("This should never be called if agent not claimed");
         rrdhost_aclk_state_unlock(localhost);
+        return;
     }
+
+    t->topic = mallocz(strlen(t->topic_recvd) + 1 - strlen(CLAIM_ID_REPLACE_TAG) + strlen(localhost->aclk_state.claimed_id));
+    dest = t->topic;
```
stelfrag, line 161: dest is reassigned in L163

stelfrag:
Not related to this PR but lwt is missing the `connect` and `connect-offset-usec` those are probably initialized when `mqtt_connected_actions` is called

```
"{\"type\":\"disconnect\",\"msg-id\":\"5f117db1-81b5-47f6-9fe1-d91b40ad8679\",\"timestamp\":1618950923,\"timestamp-offset-usec\":340571,\"connect\":0,\"connect-offset-usec\":0,\"version\":2,\"payload\":\"unexpected\"}"
```

underhood:
@stelfrag good catch. Will fix the timestamp thing in separate PR. Must have slipped our attention during original ACLK-NG review

aclk/aclk_util.c
underhood: yep i set it then copy there and move it

stelfrag (approved):

vkalintiris (approved):

PR #10860, opened 2021-03-25
Fix handling of binary package uploads.

Ferroin:
##### Summary

SSIA.

This is mostly switching to using the existing PackageCloud wrapper script to handle the uploads instead of doing it directly with cURL.

##### Component Name

area/ci

##### Test Plan

Tested locally on my fork of the Netdata repo.


Ferroin:
Merging without waiting for reviews or CI as this needs to be in for tonight’s nightlies and can’t make things worse than they already are WRT binary packages.

PR #10879, opened 2021-03-29
Change notifications email subject

MrZammler:
##### Summary

Change the email subject on alarm-notify.sh script. It should now read for example `CRITICAL: exporting metrics lost = 96.1 metrics, on winterland`

It also adds the `$info` variable to the `Details` part of the email.

##### Component Name

Health


jacekkolasa (approved):
good stuff

PR #10880, opened 2021-03-29
Add a chart for out of memory kills

### LINKED ISSUE #10875: collectors/proc/vmstat: add `oom_kill` metric collection
##### Feature idea summary

There is `oom_kill` metric in the `proc/vmstat`. It is number of [OOM](https://en.wikipedia.org/wiki/Out_of_memory) kills.

❗ Available on kernel 4.13+.

```cmd
[ilyam@pc ~]$ cat /proc/vmstat | grep oom
oom_kill 0
```

##### Expected behavior

[proc_vmstat](https://github.com/netdata/netdata/blob/master/collectors/proc.plugin/proc_vmstat.c) collects `oom_kill` metric.

---

Would be nice to add an alarm to notify users about OOM events (system memory saturation, cgroup memory limits).

There is no details about killed processes, so manual investigation is needed.

```cmd
[ilyam@pc ~]$ cat /proc/vmstat | grep oom
oom_kill 1

[ilyam@pc ~]$ sudo dmesg | egrep -i 'killed process'
[ 1836.276684] Memory cgroup out of memory: Killed process 4971 (netdata) total-vm:47820kB, anon-rss:7560kB, file-rss:0kB, shmem-rss:0kB, UID:201 pgtables:100kB oom_score_adj:1000
```

vlvkobal:
##### Summary
SSIA

Closes #10875

##### Component Name
proc plugin

##### Test Plan
1. Check the `mem.oom_kill` chart.
2. Run a container with memory limit `docker run -m 6m mongo`.
3. Check if the OOM kill was displayed in the chart.


collectors/all.h
```diff
@@ -82,6 +82,7 @@
 #define NETDATA_CHART_PRIO_MEM_SYSTEM_AVAILABLE       1010
 #define NETDATA_CHART_PRIO_MEM_SYSTEM_COMMITTED       1020
 #define NETDATA_CHART_PRIO_MEM_SYSTEM_PGFAULTS        1030
+#define NETDATA_CHART_PRIO_MEM_SYSTEM_OOM_KILL        1040
```
ilyam8, line 85: How about to put the chart right after `NETDATA_CHART_PRIO_MEM_SYSTEM_AVAILABLE`? So you don't need to scroll every time you want to check it?

vlvkobal: Yes, it makes sense. Moved.

health/health.d/ram.conf
```diff
@@ -37,6 +37,18 @@
...
+      on: mem.oom_kill
+      os: linux
+   hosts: *
+  lookup: sum -10m unaligned
+   units: kills
+   every: 10s
+    warn: $this > 0
+   delay: down 1h multiplier 1.5 max 1h
+    info: number of out of memory kills in the last 10 minutes
+      to: sysadmin
```
ilyam8, line 50:
```suggestion
   alarm: oom_kill
      on: mem.oom_kill
      os: linux
   hosts: *
  lookup: sum -1m unaligned
   units: kills
   every: 30s
    warn: $this > 0
   delay: down 5m
    info: number of out of memory kills in the last 10 minutes
      to: sysadmin
```

Some thoughts:
 - we don't want to lose oom kills (`every` needs to be < then `lookup`)
 - we don't want to have this alarm to be active forever (down 1h is too much, as you know there is no way to _acknowledge_ alarms)
 - since it is sum and warn > 0 we don't want big lookup and small every, because the alarm will be active for the whole lookup period.

What do you think?

vlvkobal: Discussed and agreed.

thiagoftsm (approved):

ilyam8 (approved):

collectors/proc.plugin/proc_vmstat.c
```diff
@@ -43,6 +59,21 @@ int do_proc_vmstat(int update_every, usec_t dt) {
         arl_expect(arl_base, "pswpin", &pswpin);
         arl_expect(arl_base, "pswpout", &pswpout);
 
+        int has_oom_kill = 0;
+        
+        for (l = 0; l < lines; l++) {
+            info ("OOM %s", procfile_lineword(ff, l, 0));
+            if (!strcmp(procfile_lineword(ff, l, 0), OOM_KILL_STRING)) {
```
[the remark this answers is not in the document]
vlvkobal: Fixed.

ilyam8 (dismissed):

thiagoftsm (dismissed):

collectors/proc.plugin/README.md
```diff
@@ -291,6 +291,28 @@ each state.
...
 
+## Monitoring memory
+
+### Monitored memory metrics
+
+-   **amount of memory swapped in/out**
+-   **amount of memory paged from/to disk**
+-   **number of memory page faults**
+-   **number of out of memory kills**
+-   **number of NUMA events**
```
joelhans, line 302: I think these can be safely un-bolded.

```suggestion
-  Amount of memory swapped in/out
-  Amount of memory paged from/to disk
-  Number of memory page faults
-  Number of out of memory kills
-  Number of NUMA events
```

ilyam8 (approved):

thiagoftsm (approved):

joelhans (approved):

PR #10885, opened 2021-03-30
Find and replace $family variable in health info strings.

MrZammler:
##### Summary

Finds and replaces `$family` and `${family}` in health info strings. Continues from https://github.com/netdata/netdata/pull/10801. 

##### Component Name

Health

##### Test Plan

Insert a `$family` or `${family}` variable in an alarm's configuration info string. Observe it to be replaced with the family value in dashboard, aclk, v1/api and notification emails.

##### Additional Information

We should merge this in the feature branch so it can be tested generally. When it goes into master we should have decided if we're going to support both variable formats, or just one of them, and whether we need more (i.e. `$this`, `$units`, etc).


MrZammler:
I think I will merge this in the feature branch, and it will be reviewed again before going to master... I think we need to give them something for the variables, and then decide if it will have both versions of the variables, or any more of them....

PR #10887, opened 2021-03-31
Provide more agent analytics to posthog

MrZammler:
##### Summary

This is a clean PR, derived from https://github.com/netdata/netdata/pull/10620.

The purpose is to provide more data to posthog analytics, from various agent information.

All the current statistics (START/EXIT, etc) will remain, and this PR will add a new one (META) that will be sent X minutes after the agent has started, to allow for gathering data.

##### Component Name

Analytics

##### Test Plan

Compile an agent with this PR. 
Observe that statistics are sent to posthog.

##### Additional Information

As the PR progresses (I will try to have clean commits), more data will be added. I think having one PR will aid in testing, although if it becomes messy, there can be multiple.


andrewm4894 (approved):
Looks good to me in terms of what's being sent to PostHog. 

Here is machine guid i used to validate in PostHog. https://posthog.netdata.cloud/person/a998d6cc-921d-11eb-9f71-42010a8e0fc8

daemon/analytics.c
```diff
@@ -0,0 +1,195 @@
+#include "common.h"
+
+static const char *verify_required_directory(const char *dir) {
+    if(chdir(dir) == -1)
+        fatal("Cannot cd to directory '%s'", dir);
```
thiagoftsm, line 5: Please, instead to have the message "Cannot cd", could we have something like this "Cannot change to directory ..."? I think the current message is not clear for users.

daemon/analytics.c
```diff
@@ -0,0 +1,195 @@
...
+            }
+        }
+    }
+
+    if(timezone && *timezone) {
+        // make sure it does not have illegal characters
+        // info("TIMEZONE: fixing '%s'", timezone);
+
+        size_t len = strlen(timezone);
+        char tmp[len + 1];
```
thiagoftsm, line 82: Did you test your code with old compilers? I am almost sure that I already saw problems with declarations like this while I was at college, but this happened more than 10 years ago.

daemon/analytics.c
```diff
@@ -0,0 +1,195 @@
...
+        }
+        *d = '\0';
+        strncpyz(buffer, tmp, len);
+        timezone = buffer;
+        // info("TIMEZONE: fixed as '%s'", timezone);
+    }
+
+    if(!timezone || !*timezone)
+        timezone = "unknown";
+
```
thiagoftsm, line 100: This function is using different methods to define the timezone, please, add a debug message that will give us conditions to identify the road used to get the timezone.

daemon/analytics.c
```diff
@@ -0,0 +1,195 @@
...
+    if (netdata_anonymous_statistics_enabled == -1) {
+        char *optout_file = mallocz(sizeof(char) * (strlen(netdata_configured_user_config_dir) +strlen(".opt-out-from-anonymous-statistics") + 2));
+        sprintf(optout_file, "%s/%s", netdata_configured_user_config_dir, ".opt-out-from-anonymous-statistics");
+        if (likely(access(optout_file, R_OK) != 0)) {
+            as_script = mallocz(sizeof(char) * (strlen(netdata_configured_primary_plugins_dir) + strlen("anonymous-statistics.sh") + 2));
+            sprintf(as_script, "%s/%s", netdata_configured_primary_plugins_dir, "anonymous-statistics.sh");
+            if (unlikely(access(as_script, R_OK) != 0)) {
+               netdata_anonymous_statistics_enabled=0;
+               info("Anonymous statistics script %s not found.",as_script);
+               freez(as_script);
```
thiagoftsm, line 168: This calll for `freez` is not in the best place, please, move outside the conditions.

MrZammler: Hi Thiago!

I agree, it is perhaps not the best place, but the `as_script` I think is supposed to be there (defined as `static`) all the time...

On the first call to `send_statistics`, if we don't know if we should send statistics or not (`if (netdata_anonymous_statistics_enabled == -1)`) the variable is setup with the script path, and sets up `netdata_anonymous_statistics_enabled` to either 0 or 1.

On subsequent calls to `send_statistics`, it has already decided that we should send or not, so the `as_script` is not defined/malloced again. If we have freed it after the if, then it will not be available to include in the `command_to_run`.

Ideally it should be freed after the last call to `send_statistics`, I will think about it....

daemon/analytics.c
```diff
@@ -0,0 +1,195 @@
...
+            } else {
+               netdata_anonymous_statistics_enabled=1;
+            }
+        } else {
+            netdata_anonymous_statistics_enabled = 0;
+            as_script = NULL;
+        }
+        freez(optout_file);
+    }
+    if(!netdata_anonymous_statistics_enabled) return;
```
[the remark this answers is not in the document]
MrZammler: Ok, thanks, fixed with https://github.com/netdata/netdata/pull/10887/commits/0f5189aa3341b6377d662165de84f3b15e64dd80

daemon/analytics.c
MrZammler: Ok, thanks! Fixed with https://github.com/netdata/netdata/pull/10887/commits/0f5189aa3341b6377d662165de84f3b15e64dd80

daemon/analytics.c
MrZammler: Well I haven't no, but considering this has been there (in `main.c`, I just moved it to `analytics.c`) for some time, perhaps we can consider it safe by now.... ?

thiagoftsm: All right, Netdata is newer than my college years, so it is safe to assume compilers already fixed this.

andrewm4894 (approved):

daemon/analytics.c
thiagoftsm: I understood your point of view, my unique concern is with the possibility this variable will never be cleaned, because this variable is kept inside the function.

MrZammler: Yes, indeed you are right, it will stay there until the agent exits... We'll see if we can up with a certain point perhaps to free it.

daemon/analytics.c
```diff
@@ -1,5 +1,41 @@
...
+}
+
+void *analytics_main(void *ptr) {
+    netdata_thread_cleanup_push(analytics_main_cleanup, ptr);
+    int sec = 0;
+
+    debug(D_ANALYTICS, "Analytics thread starts");
+
+    while(!netdata_exit) {
+        sleep(1);
```
thiagoftsm, line 19: A suggestion, please, use Netdata `heartbeat`  instead the sleep.

daemon/analytics.c
```diff
@@ -1,5 +1,41 @@
 #include "common.h"
 
+void analytics_main_cleanup(void *ptr) {
+    struct netdata_static_thread *static_thread = (struct netdata_static_thread *)ptr;
+    static_thread->enabled = NETDATA_MAIN_THREAD_EXITING;
+
+    debug(D_ANALYTICS, "Cleaning up...");
+
```
thiagoftsm, line 8: Another suggestion, I think you can clean here the thread allocated inside `daemon/main.c` and also the variable that you used in another function.

daemon/analytics.c
MrZammler: Thiago! We actually had a look at this function, for a different reason, and I'd like your input (or anyone else's!). Please have in mind that this function was already there.

It seems to follow this path to get the timezone: `TZ` variable -> `/etc/timezone` -> `/etc/localtime` (and `strftime`, etc). (In the end of course everything is overwritten if the user has set the `timezone` option in `netdata.conf`.

I'm not sure if it is ok, mainly in the sense of whether `/etc/localtime` should be picked before `/etc/timezone`. `/etc/timezone` is in essence a file that could contain almost garbage, and it wouldn't really affect the local machines date. *Most* applications do not care of what is inside `/etc/timezone`. (And in e.g. macos, I think the file is not there).

The code to also pick up `/etc/localtime` expects it to be a link (to an entry in `/usr/share/zoneinfo`) but it's also not always the case. It could be a real file copied from there, so it won't work either then.

Plus, setting the `timezone` to `unknown`, I'm not sure it's ok. It should perhaps be set to `UTC`, which is also kinda the default for no timezone set.

daemon/analytics.c
MrZammler: Done! With https://github.com/netdata/netdata/pull/10887/commits/45c2923382fee1289c7a33f585f1973f895ba5c7

daemon/analytics.c
thiagoftsm: It looks like `/etc/timezone` is inside the function to keep compatibility with all possible operate systems, on Slackware and on FreeBSD we do not have this file.

When I gave the suggestion to add a debug message, I had in mind that some of these methods do not apply to some systems and it would be good to know which method was used on my computer, because sometimes users can adjust some group of variables or files, and forget others. My concern is Netdata to show different timezone when compared with other servers running on the same computer.

About the default timezone, like you said, I remember that PostgreSQL also uses UTC as [default timezone](https://www.postgresql.org/docs/13/datatype-datetime.html). I do not have any objection to use UTC as default, and I think it is the best option.

Ferroin: IMO, the correct handling here should be, in order:

1. Any `timezone` value in `/etc/netdata.conf`. This obviously overrides any othe rsettings, so we should check it _first_
2. Any value in the `TZ` environment variable. This will be the norm for Docker containers, as well as being the de-facto standard for overriding whatever is set at the system level.
3. Any value in `/etc/timezone`. Only some Linux systems have this file, but those that do _should_ have it in sync with what `/etc/localtime` has, and it0s easier to parse (it will point to a path in the tzdata database).
4. If `/etc/localtime` is a link, the path it points to. This will usually only be the case if `/etc/timezone` _is not present_ (systems that have `/etc/timezone` tend to copy the tzdata entry, and use `/etc/timezone` to know which one to copy when the database gets updated).
5. Use `date '+%Z %z'` to get the timezone abbreviation and offset. This will (almost) always work, including if none of the above configuration is set up and `/etc/localtime` is an actual file instead of a link, and will return UTC if no other configuration exists.
6. In the exceedingly unusual case that none of the above works, continue to return ‘unknown’ as a way to indicate that something is wrong.

I agree with @thiagoftsm that we should indicate which method was used to determine the timezone, it’s useful debugging information.

MrZammler: Hi guys, thanks for your input.

The `/etc/netdata.conf` does take precedence (if it's there, it will overwrite whatever it was picked up from the whole function, even if it's in the last line). From @Ferroin 's answer, I guess it already then works as it is. We could just re-enable the commented `info` there to have a log of what method was used.

daemon/analytics.h
```diff
@@ -0,0 +1,53 @@
...
+        .config_section = NULL, \
+        .config_name = NULL, \
+        .enabled = 0, \
+        .thread = NULL, \
+        .init_routine = NULL, \
+        .start_routine = analytics_main \
+    },
+
+struct analytics_data {
+    char *NETDATA_CONFIG_STREAM_ENABLED;
```
thiagoftsm, line 28: Why are you using uppercase variable? I remember we are using this pattern for constants, but I do not remember to see variables in uppercase.

MrZammler: Hi Thiago! Yes, you are right. I was likely carried away because they mapped to env variables which are uppercase... Changed with https://github.com/netdata/netdata/pull/10887/commits/e57d39cf78eab69e61aedbe5d51692203f7d0112

daemon/analytics.c
thiagoftsm: Hello @MrZammler ,

Are you going to add a message indicating the method used to define the timezone? 

Best regards!

MrZammler: Yes! I've re-enabled the info messages with https://github.com/netdata/netdata/pull/10887/commits/c694b0f35be0c02ac6ef1966fd7450c9f629d730

Thanks!!

daemon/buildinfo.c
```diff
@@ -312,3 +313,49 @@ void print_build_info_json(void) {
     printf("  }\n");
     printf("}\n");
 };
+
+//return a list of enabled features for use in analytics
+//find a way to have proper |
+void analytics_build_info(BUFFER *b) {
+    if(FEAT_DBENGINE)        buffer_strcat (b, "dbengine");
```
thiagoftsm, line 320: Considering what I saw in this function, we are having pipe as separator, and we are assuming that `dbengine` will be always enabled, and thanks this, this is the unique keyword without a pipe. What will happen on webserver side when `dbengine` is not enabled and the message starts with a pipe `|JSON-C`, for example?

daemon/main.c
```diff
@@ -1497,8 +1305,23 @@ int main(int argc, char **argv) {
     info("netdata initialization completed. Enjoy real-time performance monitoring!");
     netdata_ready = 1;
 
+    set_late_global_environment();
+
     send_statistics("START", "-",  "-");
 
+    //check if ANALYTICS needs to start
+    if (netdata_anonymous_statistics_enabled==1) {
```
thiagoftsm, line 1313: please add space before and after `==`.

exporting/exporting_engine.c
```diff
@@ -4,6 +4,64 @@
...
+
+        switch (instance->config.type) {
+            case EXPORTING_CONNECTOR_TYPE_GRAPHITE:
+                buffer_strcat(b, "Graphite");
+                break;
+            case EXPORTING_CONNECTOR_TYPE_GRAPHITE_HTTP:
+                buffer_strcat(b, "GraphiteHTTP");
+                break;
+            case EXPORTING_CONNECTOR_TYPE_JSON:
+                buffer_strcat(b, "Json");
```
thiagoftsm, line 26: Considering that JSON means` JavaScript Object Notation`,  I suggest to use `JSON`, instead `Json`.

exporting/exporting_engine.c
```diff
@@ -4,6 +4,64 @@
...
+                break;
+            case EXPORTING_CONNECTOR_TYPE_JSON_HTTP:
+                buffer_strcat(b, "JsonHTTP");
```
thiagoftsm, line 29: the same suggestion for this text.

daemon/buildinfo.c
MrZammler: Hi Thiago. Yes, I know it's not optimal, but I believe posthog will handle it (I was going to  talk to @andrewm4894 about this). Basically from what I know, posthog can build it's own internal list separating by the pipe. An initial empty entry (i.e. without `dbengine`) could be handled by posthog. If not, I will add a counter there.

daemon/main.c
MrZammler: Fixed, thanks Thiago!

exporting/exporting_engine.c
MrZammler: Fixed!

exporting/exporting_engine.c
MrZammler: Fixed!

MrZammler:
So, this is what we have right now. We can add more data if anyone wants.

The thread will come up 2 minutes after the agent start, send a META event and then will exit. @andrewm4894 we can consider keeping the thread alive and report every X hours or something if you deem it has value.

Ferroin:
> So, this is what we have right now. We can add more data if anyone wants.
> 
> The thread will come up 2 minutes after the agent start, send a META event and then will exit. @andrewm4894 we can consider keeping the thread alive and report every X hours or something if you deem it has value.

I can see some value in having it regularly report in that the agent is still running. That will let us have a much better view of agent versions for users who have turned off auto-updates and are not otherwise regularly restarting Netdata.

daemon/analytics.c
```diff
@@ -0,0 +1,811 @@
...
+        analytics_set_data (&analytics_data.netdata_dashboard_used, b);
+    }
+}
+
+void analytics_mirrored_hosts( void ) {
+    RRDHOST *host;
+    int count = 0;
+    int reachable = 0;
+    int unreachable = 0;
+    char b[10];
```
thiagoftsm, line 208: Assuming that we will never have 2147483647 `count`, `reachable` and `unreachable`.  The size `10` is enough to store data, but to work with all possible integer values, we would need one more byte in this vector.

daemon/buildinfo.c
thiagoftsm: I wrote a consideration for integer size, but it is something small, this is the greatest concern I have right now, I will wait for @andrewm4894 confirmation before to do the last test and to approve your PR.

andrewm4894:
> > So, this is what we have right now. We can add more data if anyone wants.
> > The thread will come up 2 minutes after the agent start, send a META event and then will exit. @andrewm4894 we can consider keeping the thread alive and report every X hours or something if you deem it has value.
> 
> I can see some value in having it regularly report in that the agent is still running. That will let us have a much better view of agent versions for users who have turned off auto-updates and are not otherwise regularly restarting Netdata.

i was actually going to ask for it to function almost like a heartbeat every hour or so. would be a lot more useful like that especially for spotting issues around a deploy and stuff like that or any other changes.

daemon/analytics.c
MrZammler: Thank you my friend! Did it `11` and changed from `9` -> `10` the limits in the later lines.

MrZammler:
> > > So, this is what we have right now. We can add more data if anyone wants.
> > > The thread will come up 2 minutes after the agent start, send a META event and then will exit. @andrewm4894 we can consider keeping the thread alive and report every X hours or something if you deem it has value.
> > 
> > 
> > I can see some value in having it regularly report in that the agent is still running. That will let us have a much better view of agent versions for users who have turned off auto-updates and are not otherwise regularly restarting Netdata.
> 
> i was actually going to ask for it to function almost like a heartbeat every hour or so. would be a lot more useful like that especially for spotting issues around a deploy and stuff like that or any other changes.
> 
> so if not too much effort then sending META event every hour would be great. But if awkward or more work we could always look at adding it later.

Just pushed a change to do that! (Although I did put every 2 hours.... I think every 1 hour might be too much?, or even every 2 hours....) ?

MrZammler:
Some ideas for more attributes (probably in future PRs)... Check the `.environment` file, to try and figure out how the agent was installed (suggested by Vasilis). Check a directory (maybe /var/lib/netdata) to figure out when the user first installed the agent? I.e. check how longs has been using it? Perhaps statistics on what alerts are the most "common"... etc...

MrZammler:
If anyone wants to test it more thoroughly, compiling and running it with debug flags `0x0000000080000000` will print to the debug.log the data it has collected.

daemon/buildinfo.c
andrewm4894: yep something like "|foo|bar||blah|" would be ok by me - the posthog plugin will then downstream use javascript to split the string by "|" and just ignore the nulls in the resulting array. 

there is an example [here](https://github.com/netdata/posthog-netdata-event-processing/blob/main/process_properties.js#L25) that shows what i mean in the data processing plugin code.

andrewm4894:
@MrZammler would a `netdata_uptime` property be possible do you know? Being able to spot changes in a property like `netdata_uptime` would be another useful way to quickly and easily spot issues. So something like rough number of seconds since last start. Or even just something like `netdata_start_time` and then we could work out the `netdata_uptime` on the fly in the posthog plugin maybe that might be more efficient for the agent, just store time of startup as a event property. Thoughts?

daemon/main.c
thiagoftsm: Hello @MrZammler ,

I cloned your PR now and I cannot see the space, please, take a look on this one more time.

Thanks!

daemon/analytics.c
```diff
@@ -0,0 +1,833 @@
...
+        sprintf(optout_file, "%s/%s", netdata_configured_user_config_dir, ".opt-out-from-anonymous-statistics");
+        if (likely(access(optout_file, R_OK) != 0)) {
+            as_script = mallocz(sizeof(char) * (strlen(netdata_configured_primary_plugins_dir) + strlen("anonymous-statistics.sh") + 2));
+            sprintf(as_script, "%s/%s", netdata_configured_primary_plugins_dir, "anonymous-statistics.sh");
+            if (unlikely(access(as_script, R_OK) != 0)) {
+               netdata_anonymous_statistics_enabled=0;
+               info("Anonymous statistics script %s not found.",as_script);
+               freez(as_script);
+            } else {
+               netdata_anonymous_statistics_enabled=1;
```
thiagoftsm, line 808: Please, add space between equal, name and value.

daemon/analytics.c
```diff
@@ -0,0 +1,833 @@
...
+            }
+        } else {
+            netdata_anonymous_statistics_enabled = 0;
+            as_script = NULL;
+        }
+        freez(optout_file);
+    }
+    if (!netdata_anonymous_statistics_enabled) return;
+    if (!action) return;
+    if (!action_result) action_result="";
```
thiagoftsm, line 818: Please, also add here. and in the next lines. Are you using our `.clang-format`?

MrZammler: Hi Thiago. Both this and the previous case, were old code which I thought I shouldn't mess with (formatting wise).

I've applied clang-format to the whole file, and I'll be more careful to these issues from now on, so to not waste your time. Thank you again.

daemon/analytics.c
MrZammler: Ok, done!

daemon/main.c
MrZammler: Done, thanks and sorry for the mixup!

MrZammler:
> @MrZammler would a `netdata_uptime` property be possible do you know? Being able to spot changes in the distribution of a property like `netdata_uptime` would be another useful way to quickly and easily spot issues. So something like rough number of seconds since last start. Or even just something like `netdata_start_time` and then we could work out the `netdata_uptime` on the fly in the posthog plugin maybe that might be more efficient for the agent, just store time of startup as a event property. Thoughts?

Hi! Yes, it's doable, I think the `netdata_start_time` is the best option, I'll look into it.

daemon/analytics.c
```diff
@@ -0,0 +1,874 @@
...
+ * Set a string data with a value
+ */
+void analytics_set_data_str(char **name, char *value)
+{
+    size_t value_string_len;
+    if (*name)
+        freez(*name);
+    value_string_len = strlen(value) + 3;
+    *name = mallocz(sizeof(char) * value_string_len);
+    snprintfz(*name, value_string_len, "\"%s\"", value);
```
thiagoftsm, line 152: In normal situation when a vector has a size `N`, we pass to functions like `sprintfz` a smaller size `N - 1` to keep code safe, are you sure that it is better to give as argument the same size here?

daemon/analytics.c
```diff
@@ -0,0 +1,874 @@
...
+
+    info("Executing %s", script);
+
+    BUFFER *b = buffer_create(1000);
+    int cnt = 0;
+    FILE *fp = mypopen(script, &command_pid);
+    if (fp) {
+        char line[200 + 1];
+
+        while (fgets(line, 200, fp) != NULL) {
```
thiagoftsm, line 344: We use an approach a little bit different in [health_config.c](https://github.com/netdata/netdata/blob/d7212e031da9d1f827a64b7ba9e22c30df72b8e5/health/health_config.c#L548), because the manual gives ( `man 3 fgets` gives the following information:

```
fgets() reads in at most one less than size characters from stream and stores them into the buffer pointed to by s.  Reading stops after an EOF or a newline.
```

I think we need to rework the logic here.

daemon/main.c
```diff
@@ -1497,8 +1305,23 @@ int main(int argc, char **argv) {
...
     netdata_ready = 1;
 
+    set_late_global_environment();
+
     send_statistics("START", "-",  "-");
 
+    //check if ANALYTICS needs to start
+    if (netdata_anonymous_statistics_enabled == 1) {
+        for (i = 0; static_threads[i].name != NULL; i++) {
+            if (!strncmp(static_threads[i].name, "ANALYTICS", 9)) {
```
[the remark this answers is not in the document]
MrZammler: Hi Thiago, it intends to start the analytics thread only if the user has not opted out from analytics. There's no use in gathering data if they will never be sent.

daemon/analytics.c
MrZammler: Ok, modified a bit to provide `-1`.

daemon/main.c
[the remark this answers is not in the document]
thiagoftsm: I think we have a problem related to this, I compiled your code using :

```
 CFLAGS="-O1 -ggdb -Wall -Wextra -fno-omit-frame-pointer -Wformat-signedness -fstack-protector-all -DNETDATA_INTERNAL_CHECKS=1 -D_FORTIFY_SOURCE=2 -DNETDATA_VERIFY_LOCKS=1" ./netdata-installer.sh --disable-lto --dont-wait --dont-start-it
```

and I changed the function `analytics_main` adding a call for the function `error()`,  I cannot see the error message that I added, and I also do not see any report for the thread `ANALYTICS` inside my file `/var/log/netdata/error.log`. I remember that for we disable the analytics, it was necessary to give as argument `--disable-telemetry`, did we change this?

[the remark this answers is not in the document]
MrZammler: I was missing a lock, thanks Thiago.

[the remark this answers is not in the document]
thiagoftsm: I am glad to help you :) .
Unfortunately I tested now the latest commit, and I did not have the thread running again.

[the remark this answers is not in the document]
MrZammler: Hmm, weird.... Do you by any chance have a `.opt-out-from-anonymous-statistics` in `/etc/netdata` ? Can you check when it reaches this point `netdata_anonymous_statistics_enabled` is indeed `1` ? Thank you!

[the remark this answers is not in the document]
thiagoftsm: I made tests weeks ago with `--disable-telemetry` and I forgot to remove the file, your PR is working as expected, I will approve it.
Thank you to remember me this file.

thiagoftsm (approved):
LGTM!

daemon/analytics.c
```diff
@@ -0,0 +1,881 @@
...
+
+    buffer_free(b);
+}
+
+void analytics_charts(void)
+{
+    RRDSET *st;
+    int c = 0;
+    int show_archived = 0;
+    rrdset_foreach_read(st, localhost)
```
stelfrag, line 373: show_archived should be removed

daemon/analytics.c
```diff
@@ -0,0 +1,881 @@
...
+    // use the contents of /etc/timezone
+    if (!timezone && !read_file("/etc/timezone", buffer, FILENAME_MAX)) {
+        char *c = buffer;
+        while (*c) {
+            if (unlikely(*c == '\n'))
+                *c = '\0';
+            c++;
+        }
+        timezone = buffer;
+        info("TIMEZONE: using the contents of /etc/timezone: '%s'", timezone);
```
stelfrag, line 671: This appears now on startup. Do we really need to show it?

thiagoftsm: @stelfrag netdata uses different possible methods to select the timezone, if this is removed from the startup message, I think it would be good to insert this information inside our `error.log` for we have conditions to debug the origin of the timezone.

stelfrag: I guess it will be helpful for figuring out timezone issues to have it logged

daemon/analytics.c
MrZammler: Done (https://github.com/netdata/netdata/pull/10887/commits/67833e82af990879167d5ae97a6e75f8fef0070b)! Thanks Stelio.

andrewm4894 (approved):
LGTM - i have validated the events appear in PostHog as we want.

stelfrag (approved):

PR #10893, opened 2021-04-01
Add a CRASH event when the agent fails to properly shutdown

stelfrag:
##### Summary
- Store a "START" entry in the database (agent_log)
- Delete the entry during shutdown
- If the entry is detected on startup, the agent is assumed to have terminated abnormally e.g. crashed, killed etc

##### Component Name
code

##### Test Plan
- Start an agent
- Use `kill -9` to kill the agent process
- Restart the agent
  - CRASH event should appear in the posthog analytics

### DIFF
--- a/database/sqlite/sqlite_functions.c
+++ b/database/sqlite/sqlite_functions.c
@@ -19,6 +19,7 @@ const char *database_config[] = {
     "CREATE INDEX IF NOT EXISTS ind_c1 on chart (host_id, id, type, name);",
     "CREATE TABLE IF NOT EXISTS chart_label(chart_id blob, source_type int, label_key text, "
     "label_value text, date_created int, PRIMARY KEY (chart_id, label_key));",
+    "CREATE TABLE IF NOT EXISTS agent_log (status, date_created);"
 
     "delete from chart_active;",
     "delete from dimension_active;",
@@ -205,6 +206,13 @@ void sql_close_database(void)
     info("Closing SQLite database");
 
     add_stmt_to_list(NULL);
+
+    char *err_msg = NULL;
+    rc = sqlite3_exec(db_meta, "delete from agent_log;", 0, 0, &err_msg);
+    if (rc != SQLITE_OK) {
+        info("Failed to record that the agent is shutting down properly, rc = %d (%s)", rc, err_msg);
+        sqlite3_free(err_msg);
+    }
 
     rc = sqlite3_close_v2(db_meta);
     if (unlikely(rc != SQLITE_OK))
@@ -1337,4 +1345,37 @@ failed:
     UNUSED(chart);
 #endif
     return;
+}
+
+
+#define SQL_CHECK_SHUTDOWN      "select status from agent_log;"
+#define SQL_RECORD_AGENT_START  "insert into agent_log (status, date_created) values ('STARTED',  strftime('%s'));"
+
+int is_restart_after_crash()
+{
+    int rc;
+    int crash_rc;
+    char *err_msg = NULL;
+
+    sqlite3_stmt *res = NULL;
+
+    rc = sqlite3_prepare_v2(db_meta, SQL_CHECK_SHUTDOWN, -1, &res, 0);
+    if (unlikely(rc != SQLITE_OK)) {
+       error_report("Failed to prepare statement to check crash status");
+       return 0;
+    }
+    crash_rc = sqlite3_step(res);
+
+    if (unlikely(sqlite3_finalize(res) != SQLITE_OK))
+        error_report("Failed to finalize the prepared statement when checking crash status");
+
+    // Try to insert only if it doesn't exist
+    if (likely(crash_rc != SQLITE_ROW)) {
+        rc = sqlite3_exec(db_meta, SQL_RECORD_AGENT_START, 0, 0, &err_msg);
+        if (rc != SQLITE_OK) {
+            error_report("SQLite error during database setup, rc = %d (%s)", rc, err_msg);
+            sqlite3_free(err_msg);
+        }
+    }
+    return (crash_rc == SQLITE_ROW);
 }
--- a/database/sqlite/sqlite_functions.c
+++ b/database/sqlite/sqlite_functions.c
@@ -19,7 +19,6 @@ const char *database_config[] = {
     "CREATE INDEX IF NOT EXISTS ind_c1 on chart (host_id, id, type, name);",
     "CREATE TABLE IF NOT EXISTS chart_label(chart_id blob, source_type int, label_key text, "
     "label_value text, date_created int, PRIMARY KEY (chart_id, label_key));",
-    "CREATE TABLE IF NOT EXISTS agent_log (status, date_created);"
 
     "delete from chart_active;",
     "delete from dimension_active;",
@@ -206,13 +205,6 @@ void sql_close_database(void)
     info("Closing SQLite database");
 
     add_stmt_to_list(NULL);
-
-    char *err_msg = NULL;
-    rc = sqlite3_exec(db_meta, "delete from agent_log;", 0, 0, &err_msg);
-    if (rc != SQLITE_OK) {
-        info("Failed to record that the agent is shutting down properly, rc = %d (%s)", rc, err_msg);
-        sqlite3_free(err_msg);
-    }
 
     rc = sqlite3_close_v2(db_meta);
     if (unlikely(rc != SQLITE_OK))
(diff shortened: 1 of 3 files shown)


ilyam8:
> Use kill -9 to kill the agent process

Docker sends SIGTERM and then wait a short period so the process can exit gracefully. If the process doesn't terminate within a grace period it sends SIGKILL. It is not a crash, but it looks like it would be according _Test Plan_. Can this be a problem?

stelfrag:
> > Use kill -9 to kill the agent process
> 
> Docker sends SIGTERM and then wait a short period so the process can exit gracefully. If the process doesn't terminate within a grace period it sends SIGKILL. It is not a crash, but it looks like it would be according _Test Plan_. Can this be a problem?

This might be an issue if the netdata exit code takes a while. The "EXIT" event will be sent right away but the "cleanup" for the CRASH detection happens a bit later (after shutdown dbengine, cancel of main threads etc). 

However if we cleanup too soon and there is a crash in the cleanup sequence we will never know.

underhood (dismissed):
I would probably do this just by some file (using SQLite for this seems a bit overkill to me (too complex stuff going down internally in sqlite)) but that is just a personal opinion.

Ferroin:
It0s not just Docker that does the whole TERM then KILL methodology. That’s standard behavior in almost _every_ process management tool in existence (systemd does it, OpenRC does it, runit does it, upstart does it, etc).

We _do_ need to know if we take too long to shut down though.

I also agree with @underhood, I do not think we should be involving the database code here, it should be a lockfile of some sort as that’s both lighter weight and less likely to be influenced by third-party code (we currently cant catch crashes that occur when opening the database file, which is a huge blind spot IMO).

Of course, I’d also argue we should try and do _some_ handling on the exit side of things when possible (if we get a SIGABRT for example, we can catch that and report an actual backtrace in the event we eventually send, instead of just blindly stating that we crashed).

database/sqlite/sqlite_functions.c
```diff
@@ -1338,3 +1346,36 @@ void sql_build_context_param_list(struct context_param **param_list, RRDHOST *ho
 #endif
     return;
 }
+
+
+#define SQL_CHECK_SHUTDOWN      "select status from agent_log;"
+#define SQL_RECORD_AGENT_START  "insert into agent_log (status, date_created) values ('STARTED',  strftime('%s'));"
+
+int is_restart_after_crash()
```
thiagoftsm, line 1354: This functions is doing what is expected, and the `CRASH` is sent, I have only one doubt, please, take a look in the commands:

```
bash-5.1# sqlite3 /var/cache/netdata/netdata-meta.db
SQLite version 3.35.3 2021-03-26 12:12:52
Enter ".help" for usage hints.
sqlite> SELECT * from agent_log ;
STARTED|1617284132
sqlite> ^D
bash-5.1# ps aux| grep netdata
root     18559  0.0  0.0   6780  2712 pts/1    S+   13:44   0:00 grep netdata
bash-5.1# netdata
2021-04-01 13:44:52: netdata INFO  : MAIN : SIGNAL: Not enabling reaper
bash-5.1# ps aux| grep netdata
thiago   18567  0.8  2.9 21475259940 238908 ?  Sl   13:44   0:07 netdata
thiago   18571  0.0  0.2 21474972656 21340 ?   Sl   13:44   0:00 /usr/sbin/netdata --special-spawn-server
thiago   18773  0.0  0.0   6992  3736 ?        S    13:44   0:00 bash /usr/libexec/netdata/plugins.d/tc-qos-helper.sh 1
thiago   18799  0.0  0.2 723924 21200 ?        Sl   13:44   0:00 /usr/libexec/netdata/plugins.d/go.d.plugin 1
thiago   18804  0.0  0.0 21474908288 7856 ?    S    13:44   0:00 /usr/libexec/netdata/plugins.d/perf.plugin 1 cycles instructions
thiago   18811  1.3  1.6 21474908908 135460 ?  S    13:44   0:11 /usr/libexec/netdata/plugins.d/apps.plugin 1
root     18813  0.5  0.7 21474967784 63676 ?   Sl   13:44   0:04 /usr/libexec/netdata/plugins.d/ebpf.plugin 1
thiago   18815  0.3  0.3  51732 29040 ?        Sl   13:44   0:02 /usr/bin/python /usr/libexec/netdata/plugins.d/python.d.plugin 1
root     18819  0.0  0.0 21474908340 7668 ?    S    13:44   0:00 /usr/libexec/netdata/plugins.d/nfacct.plugin 1
root     21679  0.0  0.0   6780  2756 pts/1    S+   13:59   0:00 grep netdata
bash-5.1# kill -9 18567
bash-5.1# ps aux| grep netdata
thiago   18804  0.0  0.0 21474908288 7856 ?    S    13:44   0:00 /usr/libexec/netdata/plugins.d/perf.plugin 1 cycles instructions
thiago   18815  0.2  0.3  51732 29044 ?        Sl   13:44   0:02 /usr/bin/python /usr/libexec/netdata/plugins.d/python.d.plugin 1
root     21727  0.0  0.0   6780  2700 pts/1    S+   13:59   0:00 grep netdata
bash-5.1# kill -9 18804 18815
bash-5.1# sqlite3 /var/cache/netdata/netdata-meta.db
SQLite version 3.35.3 2021-03-26 12:12:52
Enter ".help" for usage hints.
sqlite> SELECT * from agent_log ;
STARTED|1617284132
sqlite> ^D
bash-5.1# ps aux| grep netdata
root     21805  0.0  0.0   6780  2748 pts/1    S+   13:59   0:00 grep netdata
bash-5.1# ccnd 2> err >out
bash-5.1# less err 
bash-5.1# netdata
2021-04-01 14:04:34: netdata INFO  : MAIN : SIGNAL: Not enabling reaper
bash-5.1# sqlite3 /var/cache/netdata/netdata-meta.db
SQLite version 3.35.3 2021-03-26 12:12:52
Enter ".help" for usage hints.
sqlite> SELECT * from agent_log ;
STARTED|1617284132
``` 

The `ccnd` is a script that I use to compile Netdata with all flags. And let us ignore the fact that `perf.plugin` and `python.d.plugin` stay alive after the crash, my doubt is: why are we not updating the STARTED? I think it would be good to update it every time we start the agent.

stelfrag: You are correct it's not updated every time  -- its not even used at this point. I will go ahead and simplify the PR to use a file at this point since it is not doing something intelligent and it is an overkill.

stelfrag:
> I would probably do this just by some file (using SQLite for this seems a bit overkill to me (too complex stuff going down internally in sqlite)) but that is just a personal opinion.

True, although the reasoning was to eventually store more stuff in there, a file would be a better choice at this point and we can revisit with a more complete solution. I will switch to a file.

stelfrag:
> I also agree with @underhood, I do not think we should be involving the database code here, it should be a lockfile of some sort as that’s both lighter weight and less likely to be influenced by third-party code (we currently cant catch crashes that occur when opening the database file, which is a huge blind spot IMO).

Good point!

I will switch to a file and also try to avoid the process management kills by removing the file as soon as a shutdown is detected. So the main focus will be just regular crashes that will give us an idea of what's happening out there.

database/sqlite/sqlite_functions.c
stelfrag: Switched to file

stelfrag:
> I would probably do this just by some file (using SQLite for this seems a bit overkill to me (too complex stuff going down internally in sqlite)) but that is just a personal opinion.

switched to file

stelfrag:
> > Use kill -9 to kill the agent process
> 
> Docker sends SIGTERM and then wait a short period so the process can exit gracefully. If the process doesn't terminate within a grace period it sends SIGKILL. It is not a crash, but it looks like it would be according _Test Plan_. Can this be a problem?

Switched to file that is removed fast enough so the "kill by delay" will have less chances to be considered a crash.

thiagoftsm (approved):
Everything worked as expected.

Ferroin:
> > > Use kill -9 to kill the agent process
> > 
> > 
> > Docker sends SIGTERM and then wait a short period so the process can exit gracefully. If the process doesn't terminate within a grace period it sends SIGKILL. It is not a crash, but it looks like it would be according _Test Plan_. Can this be a problem?
> 
> Switched to file that is removed fast enough so the "kill by delay" will have less chances to be considered a crash.

While not counting it as a crash is good, I _do_ think we probably want to track that kind of thing. If Netdata is consistently overshooting the shutdown delay, then that probably means something is broken (and I do not mean the length of the delay). In fact, I strongly suspect that a lot of nodes are taking long enough to shut down that they get killed, given that I regularly see Netdata taking multiple minutes to shut down on some systems.

ilyam8:
> given that I regularly see Netdata taking multiple minutes to shut down on some systems.

Yeah, i too, it happens from time to time on my _localhost_, but can't really reproduce it.

underhood (approved):

Issue #10891, opened 2021-03-31
netdata-1.30.0/mqtt_websockets: No rule to make target 'all'.


candrews:
##### Bug report summary

Trying to build netdata 1.3.0 with the configure options `--enable-cloud --with-aclk-ng` results in this error:
```
Making all in mqtt_websockets
make[2]: Entering directory '/var/tmp/portage/net-analyzer/netdata-1.30.0/work/netdata-1.30.0/mqtt_websockets'
make[2]: *** No rule to make target 'all'.  Stop.
make[2]: Leaving directory '/var/tmp/portage/net-analyzer/netdata-1.30.0/work/netdata-1.30.0/mqtt_websockets'
make[1]: *** [Makefile:5743: all-recursive] Error 1
```
[build.log](https://github.com/netdata/netdata/files/6237199/build.log)

##### OS / Environment

```
$ uname -a; grep -Hv "^#" /etc/*release
Linux irrational 5.11.0-gentoo-x86_64 #1 SMP Mon Feb 15 09:44:11 EST 2021 x86_64 Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz GenuineIntel GNU/Linux
/etc/gentoo-release:Gentoo Base System release 2.7
/etc/lsb-release:DISTRIB_ID="Gentoo"
/etc/os-release:NAME=Gentoo
/etc/os-release:ID=gentoo
/etc/os-release:PRETTY_NAME="Gentoo/Linux"
/etc/os-release:ANSI_COLOR="1;32"
/etc/os-release:HOME_URL="https://www.gentoo.org/"
/etc/os-release:SUPPORT_URL="https://www.gentoo.org/support/"
/etc/os-release:BUG_REPORT_URL="https://bugs.gentoo.org/"
```

##### Netdata version

1.3.0

##### Installation method

Using `configure` and `make`

##### Component Name

Build system

##### Steps To Reproduce

[...]

PR #10894, opened 2021-04-01
Fixed bundling of ACLK-NG components in dist tarballs.

Ferroin:
##### Summary

This ensures we actually bundle the `mqtt_websockets` code into our nightly and release tarballs so that people can properly build it without needing to build from a git clone of the repo.

##### Component Name

area/packaging

##### Test Plan

Locally verified that the tarball produced by `autoreconf -ivf && ./configure && make dist` can be used to successfully build the agent with the new ACLK code.

The Tavis CI change can be verified by looking at the build logs for this PR.

##### Additional Information

This also sorts the `dist_noinst_DATA` list in the top level Makefile so that future changes are cleaner.

Fixes: #10891

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ install:
   # These are release-related artifacts and have to be evaluated before we start doing conditional checks inside stages
   - source ".travis/tagger.sh"
   - export GIT_TAG="$(git tag --points-at)"
+  - git submodule update --init --recursive
 
 
 # Setup notification system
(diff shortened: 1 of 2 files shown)


underhood (approved):
generated tar with make dist and then used that tar to build netdata on another machine with `--aclk-ng` the `-W buildinfo` correctly reports ACLK NG is available

vkalintiris (approved):

PR #10895, opened 2021-04-01
Add a plugin for the system clock synchronization state

### LINKED ISSUE #10889: Add kernel clock sync status and time offset in between local system and reference clock metrics
##### Feature idea summary

It is important to have system clock synchronized to a reliable server. 

We can gather this data using [adjtimex](https://man7.org/linux/man-pages/man2/adjtimex.2.html) system call (linux only).

It returns a lot of metrics. I am not a specialist, but several of them looks worth to collect. Likely there is more, but i am not sure atm 🤷‍♂️ 

Related article: http://www.ntp.org/ntpfaq/NTP-s-algo-kernel.htm

I think we need:
 - clock sync status (is system clock synchronized to a reliable server or not)
 - time offset in between local system and reference clock (if synced)

`node_exporter` has [timex collector](https://github.com/prometheus/node_exporter/blob/master/collector/timex.go), it does exactly it: executes `adjtimex` system call.

##### Expected behavior

Netdata collects kernel clock sync status and time offset in between local system and reference clock.

vlvkobal:
##### Summary
SSIA

Charts added:
- System Clock Synchronization State
- Time Offset

Closes #10889

##### Component Name
internal plugins

##### Test Plan
Check `system.clock_sync_state` and `system.clock_sync_offset` charts in a `System Overview->clock synchronization` menu section.


ilyam8:
Good work @vlvkobal 👍 

Few suggestions:
 - move system clock charts under System Overview section (timex.state => system.clock_sync_state)
 - set default update every to 10
 - convert ns to ms (in the code, don't use div/mul), so we have offset chart with milliseconds

collectors/timex.plugin/plugin_timex.c
```diff
@@ -0,0 +1,178 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include "plugin_timex.h"
+#include "sys/timex.h"
```
ilyam8, line 4: can this be a problem on non linux systems? should we compile this code only on linux?

collectors/timex.plugin/plugin_timex.c
```diff
@@ -0,0 +1,178 @@
...
+        return NULL;
+
+    usec_t step = update_every * USEC_PER_SEC;
+    heartbeat_t hb;
+    heartbeat_init(&hb);
+    while (!netdata_exit) {
+        usec_t duration = heartbeat_monotonic_dt_to_now_usec(&hb);
+        heartbeat_next(&hb, step);
+
+        if (unlikely(netdata_exit))
```
vkalintiris, line 43: Why do we need to check `netdata_exit` in the loop's body?

collectors/REFERENCE.md
```diff
@@ -96,6 +96,7 @@ This section features a list of Netdata's plugins, with a boolean setting to ena
 	# PYTHONPATH environment variable = 
 	# proc = yes
 	# diskspace = yes
+    # timex = yes
```
thiagoftsm, line 99: Please, take a look in the tabulation here,

collectors/timex.plugin/plugin_timex.c
thiagoftsm: FreeBSD also supports `timex`,  I am not sure about Mac.

collectors/timex.plugin/plugin_timex.c
vlvkobal: It's the method we use to shut down Netdata gracefully.

collectors/REFERENCE.md
vlvkobal: Fixed

collectors/timex.plugin/plugin_timex.c
```diff
@@ -0,0 +1,182 @@
...
+
+    int update_every = (int)config_get_number(CONFIG_SECTION_TIMEX, "update every", 10);
+    if (update_every < localhost->rrd_update_every)
+        update_every = localhost->rrd_update_every;
+
+    int do_sync = config_get_boolean(CONFIG_SECTION_TIMEX, "clock synchronization state", CONFIG_BOOLEAN_YES);
+    int do_offset = config_get_boolean(CONFIG_SECTION_TIMEX, "time offset", CONFIG_BOOLEAN_YES);
+
+    if (unlikely(do_sync == CONFIG_BOOLEAN_NO && do_offset == CONFIG_BOOLEAN_NO))
+        return NULL;
```
thiagoftsm, line 34: We have a call for `netdata_thread_cleanup_pop` at the end of this thread, is it not better to add a `goto` here and call it instead to return  `NULL`, considering that the `netdata_thread_cleanup_push` was called at beginning?

collectors/timex.plugin/plugin_timex.c
ilyam8: yeah, i found it http://fxr.watson.org/fxr/source/sys/timex.h

Linux's adjtimex(2) is called ntp_adjtime(2) in FreeBSD.

My point is that perhaps we need to update make instructions to not build this collector on non linux systems for now. Not a specialist, just highlighting a possible problem.

collectors/timex.plugin/plugin_timex.c
vkalintiris: AFAICT, this variable should be checked before/after getting into a piece of code that will take a _long_ time to complete **and** doesn't contain any cancellation points. The specific check is right after `heartbeat_next` which contains such a cancellation point. Similarly, I'd expect most of the function calls in the loop body to (a) contain more cancellation points, and/or (b) finish quickly enough so that the loop's condition check suffices.

collectors/timex.plugin/plugin_timex.c
vlvkobal: Done.

collectors/timex.plugin/plugin_timex.c
vlvkobal: Well, I don't know why we have this pattern in all plugins. Maybe historical reasons? 🤷‍♂️

collectors/timex.plugin/plugin_timex.c
vlvkobal: Please look at https://github.com/netdata/netdata/blob/40e247375b6bc64399a8455a89c44cbb66767e04/Makefile.am#L717-L721

collectors/timex.plugin/plugin_timex.c
vkalintiris: That's what I believe as well. I asked just to make sure there's no some other weird reason. In this case, I'd suggest that we remove the checks that are in the body loop and keep the main one in the loop's header.

thiagoftsm: Probably we are doing this, because users can specify a longer period to sleep, and when the plugin wakes up, instead to run everything one more time, it checks that the correct is to leave.

vkalintiris: `netdata_exit` is set from `cancel_main_threads()`. Prior to setting `netdata_exit` it will call `netdata_thread_cancel` (ie. `pthread_cancel`). All sleep functions that we use are cancellation points. Given this, I think it doesn't matter what is the duration of a sleep call.

collectors/timex.plugin/plugin_timex.c
ilyam8: Cool 😄

collectors/timex.plugin/README.md
```diff
@@ -0,0 +1,25 @@
...
+-->
+
+# timex.plugin
+
+This plugin monitors the system clock synchronization state, under Linux.
+
+Two charts are available:
+
+-   Clock Synchronization State
+-   Time Offset
```
ilyam8, line 14:
```suggestion
-   System Clock Synchronization State
-   Time Offset Between Local System and Reference Clock
```

Ferroin: I’d go a step further and make it ‘computed time offset’. Sometimes that computation will be _very_ accurate (for example on a system syncing off of a number of high quality stratum 1 or 2 NTP servers), sometimes it will be a rough estimate at best (for example, if syncing from a system that is not, itself, tied to a specific reference clock).

collectors/timex.plugin/plugin_timex.c
stelfrag: In most cases the thread will not even see it coming :smiley: The pthread_cancel will kill it (in fact it may be the case that even in the cleanup code the netdata_exit may or may not be set yet)

(some of the loops we use do disable the cancelability and in that case the netdata_exit will help)

We need to address this at some point

collectors/timex.plugin/README.md
vlvkobal: Updated.

collectors/timex.plugin/plugin_timex.c
vlvkobal: Removed the checks.

thiagoftsm (approved):
I tested on Slackware current and the charts worked as expected.

ilyam8 (approved):

collectors/timex.plugin/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+This plugin monitors the system clock synchronization state, under Linux.
+
+Two charts are available:
+
+-   System Clock Synchronization State
+-   Computed Time Offset Between Local System and Reference Clock
+
+## configuration
+
```
joelhans, line 17: It would be great if we could add our little bit of boilerplate here about configuration.

```
Edit the `netdata.conf` configuration file using [`edit-config`](/docs/configure/nodes.md#use-edit-config-to-edit-configuration-files) from the [Netdata config directory](/docs/configure/nodes.md#the-netdata-config-directory), which is typically at `/etc/netdata`.

Scroll down to the `[plugin:timex]` section to find the available options:
```

vlvkobal: Done.

ilyam8 (approved):

thiagoftsm (approved):

joelhans (approved):

vkalintiris (approved):
LGTM!

collectors/timex.plugin/plugin_timex.c
```diff
@@ -0,0 +1,176 @@
...
+                rd_sync_state = rrddim_add(st_sync_state, "state", NULL, 1, 1, RRD_ALGORITHM_ABSOLUTE);
+            } else {
+                rrdset_next(st_sync_state);
+            }
+
+            rrddim_set_by_pointer(st_sync_state, rd_sync_state, sync_state != TIME_ERROR ? 1 : 0);
+            rrdset_done(st_sync_state);
+        }
+
+        if (do_sync) {
```
vkalintiris, line 83: Did you mean to check `do_offset`?

PR #10897, opened 2021-04-02
Change the html email subject and add a link to Edit part

MrZammler:
##### Summary

Changes format for the html email subject, and makes the `Edit` a link to the documentation.

##### Component Name

Health

##### Test Plan

When an alarm is raised, and an html mail is sent, the subject should be in the format:

`Warning|Critical|Clear, alarm_name = value, on host`
The `Edit` word in `Edit Edit this alert's configuration...` should be a link to netdata docs part for editing alarms.


MrZammler:
This will be merged to the feature branch for testing.

PR #10899, opened 2021-04-02
Get the local utc offset, and pass it to alarms

MrZammler:
##### Summary

Get the local utc offset from strftime, and pass it to alarms. This is for the cloud to produce a local time for the email notification from utc time.

##### Component Name

Health

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -696,7 +696,7 @@ static void get_system_timezone(void) {
 
     //get the utc offset
     //could be merged with the above strftime call
-    //Stelios note: This will need an agent restart to get new offset on time change (dst, etc).
+    //Note: This will need an agent restart to get new offset on time change (dst, etc).
     {
         time_t t;
         struct tm *tmp, tmbuf;
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -700,20 +700,25 @@ static void get_system_timezone(void) {
     {
         time_t t;
         struct tm *tmp, tmbuf;
-        char helper[6];
+        char zone[FILENAME_MAX + 1]; //can be really less...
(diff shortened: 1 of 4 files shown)


daemon/main.c
```diff
@@ -693,6 +693,25 @@ static void get_system_timezone(void) {
...
+        char helper[6];
+
+        t = now_realtime_sec();
+        tmp = localtime_r(&t, &tmbuf);
+
+        if (tmp != NULL) {
+            if(strftime(helper, 6, "%z", tmp) == 0)
+                buffer[0] = '\0';
+
+            netdata_utc_offset = strdupz(helper);
```
stelfrag, line 712: This is undefined on strftime failure

MrZammler: I've changed it with https://github.com/netdata/netdata/pull/10899/commits/e8638fe0cfebe1a3509447e9588b5072f0980f7c and added +0000 if it fails....

daemon/main.c
```diff
@@ -693,6 +693,28 @@ static void get_system_timezone(void) {
...
+        tmp = localtime_r(&t, &tmbuf);
+
+        if (tmp != NULL) {
+            if(strftime(helper, 6, "%z", tmp) == 0)
+                netdata_utc_offset = strdupz("+0000");
+            else
+                netdata_utc_offset = strdupz(helper);
+        }
+
+        info ("[%ld]", tmbuf.tm_gmtoff);
```
thiagoftsm, line 715: please, add more information in this info message to simplify the identification of the value. Netdata stores a lot of information inside the `error.log`.

MrZammler: Hi Thiago, sorry for the mess... this is changing at this moment, so that piece of code will go away.

Issue #10905, opened 2021-04-05
poor Dashboard->Alarms performance when there are a lot of alarms under the same section (family)


ilyam8:
##### Bug report summary

It takes 6-7 seconds to load netdata alarms when there a lot of alarms under the same section.

<img width="1059" alt="Screenshot 2021-04-05 at 15 39 57" src="https://user-images.githubusercontent.com/22274335/113574645-301b8b80-9625-11eb-8862-e6e62303550a.png">

Exact number of alarms in my test is 100

```cmd
[pc ilyam]# curl -s http://127.0.0.1:19999/api/v1/alarms?all | jq ".alarms" | jq length
100
```

---

I was trying to add alarms for systemd units state. [systemdunits collector](https://github.com/netdata/go.d.plugin/tree/master/modules/systemdunits#systemd-units-state-monitoring-with-netdata) creates a chart per unit type. Number of dimensions in charts equal to number of units per the type (up to 90 on my server).

I added the following alarm

```cmd
template: systemd_service_units_state
      on: systemd.service_units_state
  lookup: max -1s foreach *
   units: ok/failed
   every: 10s
    warn: $this != nan AND $this == 5
   delay: down 5m multiplier 1.5 max 1h
    info: systemd service crashed
      to: sysadmin
```

And noticed performance problems.

##### OS / Environment

```
Linux pc 5.10.23-1-MANJARO #1 SMP PREEMPT Thu Mar 11 18:47:18 UTC 2021 x86_64 GNU/Linux
```

<details>
<summary>release [Click]</summary>

```cmd
/etc/arch-release:Manjaro Linux
/etc/lsb-release:DISTRIB_ID=ManjaroLinux
/etc/lsb-release:DISTRIB_RELEASE=21.0
/etc/lsb-release:DISTRIB_CODENAME=Ornara
/etc/lsb-release:DISTRIB_DESCRIPTION="Manjaro Linux"
/etc/manjaro-release:Manjaro Linux
/etc/os-release:NAME="Manjaro Linux"
/etc/os-release:ID=manjaro
/etc/os-release:ID_LIKE=arch
/etc/os-release:BUILD_ID=rolling
/etc/os-release:PRETTY_NAME="Manjaro Linux"
/etc/os-release:ANSI_COLOR="32;1;24;144;200"
/etc/os-release:HOME_URL="https://manjaro.org/"
/etc/os-release:DOCUMENTATION_URL="https://wiki.manjaro.org/"
[...]

PR #10906, opened 2021-04-05
health: add systemdunits alarms

ilyam8:
##### Summary

This PR adds alarms for the [systemdunits collector](https://github.com/netdata/go.d.plugin/tree/master/modules/systemdunits#systemd-units-state-monitoring-with-netdata) charts.

States:
 - `active`: started, bound, plugged in, etc. depending on the unit type. 
 - `inactive`: stopped, unbound, unplugged, etc.
 - `activating`: in the process of being activated.
 - `deactivating`: in the process of being deactivated.
 - `failed`: very similar to inactive and is entered when the service failed in some way (process returned error code on exit, or crashed, an operation timed out, or after too many restarts).

New alarms check if there are any systemd units in the failed state.

I tried to add those alarms using [foreach](https://learn.netdata.cloud/docs/agent/health/reference#alarm-line-lookup) (for every unit), but faced https://github.com/netdata/netdata/issues/10905 (likely there is more problems).

##### Component Name

`health`

##### Test Plan

- [x] install this PR, and ensure
  - there is no errors during install
  - systemdunits.conf file in the stock health dir
  - the alarms are in the Dashboard->Alarms->All and their values are not `-`

### DIFF
--- a/health/health.d/systemdunits.conf
+++ b/health/health.d/systemdunits.conf
@@ -9,7 +9,7 @@ template: systemd_service_units_state
    every: 10s
     warn: $this != nan AND $this == 5
    delay: down 5m multiplier 1.5 max 1h
-    info: several systemd service units are in the failed state
+    info: one or more systemd service units are in the failed state
       to: sysadmin
 
 ## Socket units
@@ -20,7 +20,7 @@ template: systemd_socket_units_state
    every: 10s
     warn: $this != nan AND $this == 5
    delay: down 5m multiplier 1.5 max 1h
-    info: several systemd socket units are in the failed state
+    info: one or more systemd socket units are in the failed state
       to: sysadmin
 
 ## Target units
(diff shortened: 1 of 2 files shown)


health/health.d/systemdunits.conf
```diff
@@ -0,0 +1,112 @@
+## Check if the are any systemd units in the failed state (crashed).
+## States: 1 - active, 2 - inactive, 3 - activating, 4 - deactivating, 5 - failed.
+
+## Service units
+template: systemd_service_units_state
+      on: systemd.service_units_state
+  lookup: max -1s min2max
```
ilyam8, line 7: > max -1s

Intention is to get the latest value (current state).

Read it as: 

> Notify if there are any units in the failed state now

Since we `every: 10s` it is possible to miss a unit that crashed, but was restarted by systemd during `every`. 

It is _by design_.

thiagoftsm (approved):
I tested on CentOS 8 and everything worked as expected.

vlvkobal (approved):

thiagoftsm (approved):
I am approving without test it again, because the latest commit changed only the description.

PR #10688, opened 2021-02-25
health: make alarms less sensitive

ilyam8:
##### Summary

Fixes: #10687

The goal of this PR is to adapt our ~top noisy system alarms to general cases:
 - relax trigger conditions.
 - remove `crit` level if the event is not really critical for a system.

Working on this i realized that `warn` and `crit` levels require you to be really careful and doesn't allow to notify about some events you want to notify - using `warn` level for them makes Netdata health kind of annoying and noisy and, as a result, you stop pay attention to all those warnings - having `notice` level would be nice!

| Alarm       | Change      | Reason      |
| ----------- | ----------- | ----------- |
| [used_file_descriptors](https://github.com/netdata/netdata/blob/db9db41bc8a3234b24f1468cd7172b719505cfa0/health/health.d/apps_plugin.conf#L5-L15)      | Removed       | Cleanup. Was commented out.             |
| [bcache_cache_errors](https://github.com/netdata/netdata/blob/db9db41bc8a3234b24f1468cd7172b719505cfa0/health/health.d/bcache.conf#L2-L11)   | Tuned `lookup`, added `delay up`        | Not errors, but read races. Not harmful. Meaning is number of times data was being read from the cache, the bucket was reused and invalidated => increase read races counter, reread data. ❗ **The metric doesn't look useful, the alarm could be deleted**...  |
| [cluster_space_usage](https://github.com/netdata/netdata/blob/db9db41bc8a3234b24f1468cd7172b719505cfa0/health/health.d/ceph.conf#L3-L12)      | Changed calculation from avail to used, tune `every`       | Kind of cleanup. Make it consistent with other disk usage alarms.             |
| [out_of_disk_space_time](https://github.com/netdata/netdata/blob/db9db41bc8a3234b24f1468cd7172b719505cfa0/health/health.d/disks.conf#L68-L80)      | Commented out       | High chance to be false positive (netowrks and disks are fast these days, making assumption based on write rate (if that write rate last for N hours...) seems not good **in general**). Disk purpose is to store data. **If there is enough space we are ok**. We have `disk_space_usage` alarm. ❗ Not removed, so users can use it as an example.  |
| [out_of_disk_inodes_time](https://github.com/netdata/netdata/blob/db9db41bc8a3234b24f1468cd7172b719505cfa0/health/health.d/disks.conf#L109-L121)      | Commented out       | Same arguments as above. We have `disk_inode_usage` alarm. ❗ Not removed, so users can use it as an example. |
[...]

PR #10926, opened 2021-04-06
health: use $family variable in the alarms info

ilyam8:
##### Summary

This PR adds $family variables to the following sets of alarms:
 - net ($family is a network device name)
 - disks ($family is a device or partition name)
 - mdstat ($family is an array name)

##### Component Name

`health`

##### Test Plan

- [x] Install this PR and check net/disks/mdstat alarms info on the dashboard and in the `api/v1/alarms?all`

### DIFF
--- a/health/health.d/disks.conf
+++ b/health/health.d/disks.conf
@@ -23,7 +23,7 @@ component: Disk
      warn: $this > (($status >= $WARNING ) ? (80) : (90))
      crit: $this > (($status == $CRITICAL) ? (90) : (98))
     delay: up 1m down 15m multiplier 1.5 max 1h
-     info: disk space utilization
+     info: disk $family space utilization
        to: sysadmin
 
  template: disk_inode_usage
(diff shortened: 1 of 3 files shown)


health/health.d/disks.conf
```diff
@@ -150,7 +150,7 @@ component: Disk
      warn: $this > $green * (($status >= $WARNING)  ? (0.7) : (1))
      crit: $this > $red   * (($status == $CRITICAL) ? (0.7) : (1))
     delay: down 15m multiplier 1.2 max 1h
-     info: average percentage of time the disk was busy over the last 10 minutes
+     info: average percentage of time $family disk was busy over the last 10 minutes
        to: silent
```
MrZammler, line 154: Hi @ilyam8 , is this one supposed to have `to: silent` by default?

ilyam8: yes

ilyam8: > If this is 100%, it doesn’t always mean that the device is saturated. Depends on a number of oustanding requests and the number of concurrent requests the device can handle. For instance, if the device is able to handle 8 concurrent requests and there are 2 we would see 100% untilization, but there is no problem at all.

from https://github.com/netdata/netdata/pull/10688

MrZammler: Many thanks man!

Issue #10931, opened 2021-04-07
/tmp/netdata-updater-XXX/netdata-updater.sh: line 298: cd: too many arguments


nabijaczleweli:
##### Bug report summary
```
Date: Wed,  7 Apr 2021 07:16:58 +0200 (CEST)
From: systemd-cron <root@tarta.nabijaczleweli.xyz>
To: root@tarta.nabijaczleweli.xyz
Subject: [tarta] job cron-daily failed

● cron-daily.service - systemd-cron daily script service
   Loaded: loaded (/lib/systemd/system/cron-daily.service; static; vendor preset: enabled)
   Active: failed (Result: exit-code) since Wed 2021-04-07 07:16:55 CEST; 120ms ago
     Docs: man:systemd.cron(7)
  Process: 5609 ExecStartPre=/lib/systemd-cron/boot_delay 5 (code=exited, status=0/SUCCESS)
  Process: 5610 ExecStart=/bin/run-parts --report /etc/cron.daily (code=exited, status=1/FAILURE)
 Main PID: 5610 (code=exited, status=1/FAILURE)

Apr 07 07:16:55 tarta run-parts[5610]: /etc/cron.daily/netdata-updater:
Apr 07 07:16:55 tarta run-parts[5610]: /tmp/netdata-updater-QPYlo9rBYN/netdata-updater.sh: line 298: cd: too many arguments
Apr 07 07:16:55 tarta run-parts[5610]: Wed  7 Apr 07:16:55 CEST 2021 : INFO:  Current Version: 00103000000003
Apr 07 07:16:55 tarta run-parts[5610]: Wed  7 Apr 07:16:55 CEST 2021 : INFO:  Latest Version: 00103000000011
Apr 07 07:16:55 tarta run-parts[5610]: ./netdata-latest.tar.gz: OK
Apr 07 07:16:55 tarta run-parts[5610]: run-parts: /etc/cron.daily/netdata-updater exited with return code 1
Apr 07 07:16:55 tarta systemd[1]: cron-daily.service: Main process exited, code=exited, status=1/FAILURE
Apr 07 07:16:55 tarta systemd[1]: cron-daily.service: Failed with result 'exit-code'.
Apr 07 07:16:55 tarta systemd[1]: Failed to start systemd-cron daily script service.
Apr 07 07:16:55 tarta systemd[1]: cron-daily.service: Triggering OnFailure= dependencies.
```
and
```
[...]

PR #10936, opened 2021-04-08
Don’t use glob expansion in argument to `cd` in updater.

Ferroin:
##### Summary

As in the summary. We should never be using glob expansion in arguments for commands that expect a single argument.

In this case, we can use a combination of `find` and `head` to extract exactly one matching directory based on some additional info we already have, and then pass just that path to `cd`.

##### Component Name

area/packaging

##### Test Plan

Verified using CI.

##### Additional Information

Fixes: #10931

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -234,13 +234,11 @@ parse_version() {
 }
 
 get_latest_version() {
-  local latest
   if [ "${RELEASE_CHANNEL}" == "stable" ]; then
-    latest="$(download "https://api.github.com/repos/netdata/netdata/releases/latest" /dev/stdout | grep tag_name | cut -d'"' -f4)"
+    download "https://api.github.com/repos/netdata/netdata/releases/latest" /dev/stdout | grep tag_name | cut -d'"' -f4
   else
-    latest="$(download "$NETDATA_NIGHTLIES_BASEURL/latest-version.txt" /dev/stdout)"
+    download "$NETDATA_NIGHTLIES_BASEURL/latest-version.txt" /dev/stdout
   fi
-  parse_version "$latest"
 }
 
 set_tarball_urls() {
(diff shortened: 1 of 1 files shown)


vkalintiris (approved):
Not related to this PR, but it would be awesome if we could start using `set -u` in our functions/scripts.

knatsakis (approved):

vkalintiris (approved):

Steve8291:
The `find` command looks like it will work to fix #10937
Closing #10937 in favor of this fix.
I also agree about using `set -u` just need to check all the scripts to make sure that they don't expect to continue if a variable isn't set. I've been caught by that before.

Steve8291:
One last thought on this - Will users have to manually run the `kickstart.sh` again if they have the older non-working code from this `netdata-updater.sh` installed? I know my installation had stopped updating altogether once it downloaded the non-functioning `cd` command.

Ferroin:
> One last thought on this - Will users have to manually run the `kickstart.sh` again if they have the older non-working code from this `netdata-updater.sh` installed? I know my installation had stopped updating altogether once it downloaded the non-functioning `cd` command.

The `netdata-updater.sh` script will attempt to run the newest code from the repo if it’s newer than the local copy, and that check and handoff happens before the problematic piece of code gets run. As a result, as long as this is not masking some other issue (and for all I know it might be), things should start working as soon as this gets merged.

packaging/installer/netdata-updater.sh
```diff
@@ -295,7 +295,7 @@ update() {
     NEW_CHECKSUM="$(safe_sha256sum netdata-latest.tar.gz 2> /dev/null | cut -d' ' -f1)"
     tar -xf netdata-latest.tar.gz >&3 2>&3
     rm netdata-latest.tar.gz >&3 2>&3
-    cd netdata-* || exit 1
+    cd "$(find . -maxdepth 1 -name "netdata-${path_version}*" | head -n 1)" || exit 1
```
nabijaczleweli, line 298: `-print -quit`

PR #10937, opened 2021-04-08
Fix quoting on wildcard glob

Steve8291:
This script is failing because of incorrect quoting on a wildcard glob for bash.
Here is the output from running it:

```
+ sudo //usr/libexec/netdata/netdata-updater.sh --not-running-from-cron
Thu Apr  8 08:21:03 EDT 2021 : INFO:  Running on a terminal - (this script also supports running headless from crontab)
Thu Apr  8 08:21:03 EDT 2021 : INFO:  Current Version: 00103000000003
Thu Apr  8 08:21:03 EDT 2021 : INFO:  Latest Version: 00103000000016
./netdata-latest.tar.gz: OK
/tmp/netdata-updater-lz3VVkHsCz/netdata-updater.sh: line 298: cd: too many arguments
+ local ret=1
+ '[' 1 -ne 0 ']'
+ run_failed
+ printf ' FAILED  \n\n'
 FAILED  

+ printf 'FAILED with exit code 1\n'
+ return 1
+ fatal 'Failed to update existing Netdata install'
+ printf ' ABORTED  Failed to update existing Netdata install \n\n'
 ABORTED  Failed to update existing Netdata install
```

##### Summary
Added quotes to `cd netdata-* || exit 1`
`cd "netdata-"* || exit 1`

##### Component Name
Packaging/Installer

##### Test Plan
Tested on Ubuntu 18.04.5 LTS

### DIFF
--- a/packaging/installer/netdata-updater.sh
+++ b/packaging/installer/netdata-updater.sh
@@ -295,7 +295,7 @@ update() {
     NEW_CHECKSUM="$(safe_sha256sum netdata-latest.tar.gz 2> /dev/null | cut -d' ' -f1)"
     tar -xf netdata-latest.tar.gz >&3 2>&3
     rm netdata-latest.tar.gz >&3 2>&3
-    cd netdata-* || exit 1
+    cd "netdata-"* || exit 1
     RUN_INSTALLER=1
     cd "${NETDATA_LOCAL_TARBALL_OVERRIDE}" || exit 1
   fi


Ferroin:
AIUI, the issue here is not the quoting, but the fact that the glob expression is matching multiple valid file/directory names, which is causing `cd` to fail.

Steve8291:
> AIUI, the issue here is not the quoting, but the fact that the glob expression is matching multiple valid file/directory names, which is causing `cd` to fail.

I think it will still fail because it won't be expanded correctly unless there is some `shopt` being set somewhere. But I agree, it would be better to make sure it didn't match multiple directories. Although I only ever get the single directory downloaded to /tmp when running the `kickstart.sh` or `netdata-updater.sh` and it cleans up after itself. What if we set a variable to the path of the tmp dir that is being created and then referred to that like: `UPDATER_DIR=$(mktemp -d -t netdata-updater-XXXXXXXXXX)` at line 139 and then used that variable in place of the wildcard?

Ferroin:
Relevant as an alternative ‘fix’ for the issue here: https://github.com/netdata/netdata/pull/10936

Steve8291:
@Ferroin 
Yep. Looks like #10936 will fix it too. I think just setting it as a variable is simpler but `find` is more robust. I can close this in favor of #10936 if you like.

Ferroin:
> @Ferroin
> Yep. Looks like #10936 will fix it too. I think just setting it as a variable is simpler but `find` is more robust. I can close this in favor of #10936 if you like.

Sounds good to me.

Also, if you could comment on #10936 indicating that you’ve looked at it and it looks like it should fix things, that would be awesome. We unfortunately are a bit short on people who have a solid understanding of shell script right now, so getting a proper review for stuff like this is somewhat challenging. Having a vote of confidence from interested community members that something at least looks correct is always appreciated on such PRs.

Steve8291:
@Ferroin 
Will do. I'd be happy to test it quick in my environment. Is there an easy way to force the script to run the update when you already have the most current version of netdata installed? Otherwise it will miss the code because I just updated.

Ferroin:
> @Ferroin
> Will do. I'd be happy to test it quick in my environment. Is there an easy way to force the script to run the update when you already have the most current version of netdata installed? Otherwise it will miss the code because I just updated.

Unfortunately it’s a bit complicated. The easiest approach is probably to adapt the code we currently have in CI for testing this (the overall process can be found in https://github.com/netdata/netdata/blob/master/.github/workflows/updater.yml) to your own environment. I use a similar process myself for testing, but the exact details are heavily influenced by the way I have my development system set up.

Steve8291:
Closing in favor of fix #10936

PR #10980, opened 2021-04-15
remove vneg from ACLK-NG

underhood:
##### Summary
Removes version negotiation code `hello`/`version` messages which have been obsoleted in favor of feature flags/capabilities mechanism in new cloud architecture

##### Component Name
ACLK-NG
##### Test Plan
Should still work OK with cloud as before

### DIFF
--- a/aclk/aclk.c
+++ b/aclk/aclk.c
@@ -38,8 +38,6 @@ netdata_mutex_t aclk_shared_state_mutex = NETDATA_MUTEX_INITIALIZER;
 struct aclk_shared_state aclk_shared_state = {
     .agent_state = AGENT_INITIALIZING,
     .last_popcorn_interrupt = 0,
-    .version_neg = 0,
-    .version_neg_wait_till = 0,
     .mqtt_shutdown_msg_id = -1,
     .mqtt_shutdown_msg_rcvd = 0
 };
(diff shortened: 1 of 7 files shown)


underhood:
I am keeping the inbound msgtype resolution there as there will be new message types in future PRs.
```
    for (int i = 0; aclk_incoming_msg_types[i].name; i++) {
        if (strcmp(cloud_to_agent.type_id, aclk_incoming_msg_types[i].name) == 0) {
            if (likely(!aclk_incoming_msg_types[i].fnc(&cloud_to_agent, payload))) {
                // in case of success handler is supposed to clean up after itself
                // or as in the case of aclk_handle_cloud_request take
                // ownership of the pointers (done to avoid copying)
                // see what `aclk_queue_query` parameter `internal` does

                // NEVER CONTINUE THIS LOOP AFTER CALLING FUNCTION!!!
                // msg handlers (namely aclk_handle_version_responce)
                // can freely change what aclk_incoming_msg_types points to
                // so either exit or restart this for loop
                freez(cloud_to_agent.type_id);
                return 0;
            }
            goto err_cleanup;
        }
    }
    ```

stelfrag (dismissed):
Looks good -- needs to go after the new https client otherwise will deadlock

vkalintiris (dismissed):
LGTM. FYI, there's a comment reference to `version_neg` in aclk_tx_msgs.c

underhood:
@vkalintiris thx I removed the reference in the comment

stelfrag (approved):

vkalintiris (approved):

PR #10961, opened 2021-04-12
Provide new attributes in health conf files

MrZammler:
##### Summary

This PR adds 3 new attributes (`class`, `component` and `type`) to health configuration files, as speficied in https://docs.google.com/spreadsheets/d/1I-9RuX_Sr3bB-W-C2zzZ5TOkMgBBqbbjoNvApZIw2AY/edit#gid=50781198&fvid=275594596

The attributes are read, stored in the health-log.db, and added to health related json structures (to be sent to the cloud). When an attribute is missing, it is sent as `Unknown`.

It also parses and replaces a `$family` variable from info files.

##### Component Name

Health

##### Test Plan

Check that the new attributes appear in alarm related jsons.
The agent should not crash if reading a health-log.db file from a previous agent version (which lacks the new attributes).
Check that the $family variable is correctly replaced in info strings that contain it.
Check the new conf files for any regressions! They should contain the latest changes from @ilyam8.

##### Additional Information

This is PR 1 out of (probably 3) to bring the changes from https://github.com/netdata/netdata/tree/alerts-changes-notify-attributes to the master.

PR #10984, opened 2021-04-16
Add sections for class, component and type.

MrZammler:
##### Summary

Add initial sections in the documentation for the new `class`, `component` and `type` attributes, that the PR https://github.com/netdata/netdata/pull/10961 adds to the stock alarms configuration files.

##### Component Name

Health

##### Test Plan

Not needed.

##### Additional Information

The documentation can be perhaps extended to provide information on how those attributes are used in the cloud.

### DIFF
--- a/health/REFERENCE.md
+++ b/health/REFERENCE.md
@@ -136,15 +136,43 @@ If you create a template using the `disk.io` context, it will apply an alarm to 
 
 Specify the classification of the alarm or template. 
 
-Class can be used to indicate the broader area of the system that the alarm applies to. For example, under the general `Database` class, you can group together alarms that operate on various database systems, like `MySQL`, `CockroachDB`, `CouchDB` etc.
+Class can be used to indicate the broader area of the system that the alarm applies to. For example, under the general `Database` class, you can group together alarms that operate on various database systems, like `MySQL`, `CockroachDB`, `CouchDB` etc. Example:
 
 ```yaml
 class: Database
 ```
 
+Netdata's stock alarms use the following `class` attributes by default, but feel free to adjust for your own requirements:
+
+| Class                    | Groups alarms for:                              |
+| ------------------------ | ----------------------------------------------- |
+| Ad Filtering             | Services related to Ad Filtering (like pi-hole) |
+| Certificates             | 
+| Cgroups                  |
+| Computing                | 
+| Containers               |
+| Database                 | Database systems (e.g. MySQL, Postgress, etc)   |
+| Data Sharing             |
+| DHCP                     |
+| DNS                      | 
+| Kubernetes               |
+| KV Storage               |
+| Linux                    | Services specific to Linux (e.g. systemd)       |
+| Messaging                |
+| Netdata                  | Internal Netdata components monitoring          |
+| Other                    |
+| Power Supply             |
+| Search engine            |
+| Storage                  |
+| System                   | General system alarms (e.g. cpu, network, etc.) |
+| Virtual Machine          | Virtual Machine software                        |
+| Web Proxy                |
+| Web Server               | Web server software (e.g. Apache, ngnix, etc.)  |
+| Windows                  |                                                 |
+
 #### Alarm line `component`
 
-Component can be used to narrow down what the previous `class` value specifies for each alarm or template. Continuing from the previous example, `component` might include `MySQL`, `CockroachDB`, `MongoDB`, all under the same `Database` classification.
+Component can be used to narrow down what the previous `class` value specifies for each alarm or template. Continuing from the previous example, `component` might include `MySQL`, `CockroachDB`, `MongoDB`, all under the same `Database` classification. Example:
 
 ```yaml
 component: MySQL
@@ -152,7 +180,7 @@ component: MySQL
 
 #### Alarm line `type`
 
-This indicates the type of error (or general problem area) that the alarm or template applies to. For example, `Latency` can be used for alarms that trigger on latency issues, be it in network interfaces, web servers, or database systems.
+This indicates the type of error (or general problem area) that the alarm or template applies to. For example, `Latency` can be used for alarms that trigger on latency issues, be it in network interfaces, web servers, or database systems. Example:
 
 ```yaml
 type: Latency
--- a/health/REFERENCE.md
+++ b/health/REFERENCE.md
@@ -141,7 +141,8 @@ Class can be used to indicate the broader area of the system that the alarm appl
 ```yaml
 class: Database
 ```
-Netdata's stock alarms use the following `class` attributes by default, but feel free to adjust for your own requirements:
+<details>
+<summary>Netdata's stock alarms use the following `class` attributes by default, but feel free to adjust for your own requirements:</summary>
 
 | Class                    | Description                                                                                      |
 | ------------------------ | ------------------------------------------------------------------------------------------------ |
@@ -168,6 +169,8 @@ Netdata's stock alarms use the following `class` attributes by default, but feel
 | Web Proxy                | Web proxy software (e.g. squid)                                                                  |
 | Web Server               | Web server software (e.g. Apache, ngnix, etc.)                                                   |
 | Windows                  | Alerts for monitor of wmi services                                                               |
+
+</details>
 
 If an alarm configuration is missing the `class` line, it's value will default to `Unknown`.
 
--- a/health/REFERENCE.md
+++ b/health/REFERENCE.md
@@ -172,7 +172,7 @@ class: Database
 
 </details>
 
-If an alarm configuration is missing the `class` line, it's value will default to `Unknown`.
+If an alarm configuration is missing the `class` line, its value will default to `Unknown`.
 
 #### Alarm line `component`
 
--- a/health/REFERENCE.md
+++ b/health/REFERENCE.md
@@ -58,6 +58,9 @@ Netdata parses the following lines. Beneath the table is an in-depth explanation
 | --------------------------------------------------- | --------------- | ------------------------------------------------------------------------------------- |
 | [`alarm`/`template`](#alarm-line-alarm-or-template) | yes             | Name of the alarm/template.                                                           |
 | [`on`](#alarm-line-on)                              | yes             | The chart this alarm should attach to.                                                |
+| [`class`](#alarm-line-class)                        | no              | The general classification of the alarm.                                              |
+| [`component`](#alarm-line-component)                | no              | Specify the component of the class of the alarm.                                      |
+| [`type`](#alarm-line-type)                          | no              | The type of error the alarm monitors.                                                 |
 | [`os`](#alarm-line-os)                              | no              | Which operating systems to run this chart.                                            |
 | [`hosts`](#alarm-line-hosts)                        | no              | Which hostnames will run this alarm.                                                  |
 | [`plugin`](#alarm-line-plugin)                      | no              | Restrict an alarm or template to only a certain plugin.                                             |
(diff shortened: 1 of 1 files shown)


health/REFERENCE.md
```diff
@@ -129,6 +132,32 @@ You're interested in what comes after the comma: `disk.io`. That's the name of t
...
 If you create a template using the `disk.io` context, it will apply an alarm to every disk available on your system.
 
+#### Alarm line `class`
+
+Specify the classification of the alarm or template. 
+
+Class can be used to indicate the broader area of the system that the alarm applies to. For example, under the general `Database` class, you can group together alarms that operate on various database systems, like `MySQL`, `CockroachDB`, `CouchDB` etc.
+
+```yaml
+class: Database
```
ilyam8, line 142: What about allowed characters? Are there any restrictions? Are spaces allowed?

thiagoftsm: Considering the current code in https://github.com/netdata/netdata/pull/10961, I think we would have the same restrictions that JSON has, except for quotes that we are stripping, @MrZammler can you confirm this?

health/REFERENCE.md
```diff
@@ -129,6 +132,32 @@ You're interested in what comes after the comma: `disk.io`. That's the name of t
...
+```
+
+#### Alarm line `component`
+
+Component can be used to narrow down what the previous `class` value specifies for each alarm or template. Continuing from the previous example, `component` might include `MySQL`, `CockroachDB`, `MongoDB`, all under the same `Database` classification.
+
```
thiagoftsm, line 148: Can `component` and `class` to have the exact same values? If it is not possible, we need to update the PR #10961.

health/REFERENCE.md
```diff
@@ -129,6 +132,32 @@ You're interested in what comes after the comma: `disk.io`. That's the name of t
...
+component: MySQL
+```
+
+#### Alarm line `type`
+
+This indicates the type of error (or general problem area) that the alarm or template applies to. For example, `Latency` can be used for alarms that trigger on latency issues, be it in network interfaces, web servers, or database systems.
+
+```yaml
+type: Latency
+```
```
thiagoftsm, line 159: Right now our current code gives a big freedom for users, I do not see this as a problem, on the other hand this freedom also brings the necessity to start thinking on binary trees to get data related to these possible values, I did not see them in the previous PR.

health/REFERENCE.md
MrZammler: Hi Thiago, yes, I will do some tests and report back.

thiagoftsm: I will wait for your tests before to proceed with the review, the question @ilyam8 raised is very important for that PR.
Please, also take a look in the next question I wrote, because it also impacts the PR.

MrZammler: The final values of these strings are consumed by the cloud. I think they're used to group together alarms. That's why I've added @dim08 (the docs should reflect that too - that's their main use after all). 

So I think if the user chooses to have a `class` value of `System 1 - my great machine!!^@&#` to some alarms, and the cloud is able to consume it, we should allow it to be read and pass through the agent. We do strip quotes (both single and double), so this could go in the docs maybe? (But again if grouping is the target, then it should work).

The same could be true for the info field? I think any character that can be passed through JSON should be allowed, unless the cloud has other restrictions that I'm not aware of. Of course some kind of non-utf valid character, will likely break things, so if it's needed we should sanitize the fields (and the info one too).

health/REFERENCE.md
MrZammler: I don't recall that being a requirement... or even discussed at all for that matter. I think a requirement like that would have been explicitly stated.

ilyam8: > Can component and class to have the exact same values?

Yes, "Unknown" 😄

health/REFERENCE.md
thiagoftsm: According your previews [answer](https://github.com/netdata/netdata/pull/10984#discussion_r614888815) this is liked to be used on cloud only, so I will mark this as resolved, but case we start using this on agent, it will be necessary to add more control on agent side.

health/REFERENCE.md
thiagoftsm: > What about allowed characters? Are there any restrictions? Are spaces allowed?

I do not know exactly what the product team wants, but I made some tests on Friday and today, and I saw that we can use any character for these events, on the other hand I discovered a bug thanks these tests, but I will bring a complete different PR that will address it.

Please, talk with the cloud team to confirm they are expecting any ASCII character.

joelhans (commented):
No specific changes for me to request at the moment, however I'm wondering if it would be helpful to list the `class`, `component`, and `type` values that are already being used by our preconfigured alarms. For example, the `class` section might have a paragraph like:

> If you want to use an existing class already defined in Netdata's preconfigured alarms, use one of the following: `Database`, `Web`, `RAM`, ...

MrZammler:
> No specific changes for me to request at the moment, however I'm wondering if it would be helpful to list the `class`, `component`, and `type` values that are already being used by our preconfigured alarms. For example, the `class` section might have a paragraph like:
> 
> > If you want to use an existing class already defined in Netdata's preconfigured alarms, use one of the following: `Database`, `Web`, `RAM`, ...

Hi Joel! I've added a table of all the stock `class` attributes we use. Not sure if we want it like this (a table), have a description or not of each attribute?

The `component` and `type` attributes can get really long for a table though, and I don't know if it makes sense to describe them all ?

OdysLam:
What's the status of this PR? @MrZammler 

Let's try to merge this since the relevant code is already merged (I believe).

MrZammler:
Well, considering it's my first document-related PR, I'm ok if you guys are too. Yes, the code is already commited (along with the additions of those fields in the alarms configuration).

health/REFERENCE.md
thiagoftsm: @MrZammler the unique doubt I have to approve the PR is this discussion, do we arrive in the conclusion that cloud is allowing any ASCII character?

MrZammler: Hi Thiago, I'm looking into this (the characters), thanks.

health/REFERENCE.md
```diff
@@ -129,6 +132,60 @@ You're interested in what comes after the comma: `disk.io`. That's the name of t
...
+| Netdata                  | Internal Netdata components monitoring          |
+| Other                    |
+| Power Supply             |
+| Search engine            |
+| Storage                  |
+| System                   | General system alarms (e.g. cpu, network, etc.) |
+| Virtual Machine          | Virtual Machine software                        |
+| Web Proxy                |
+| Web Server               | Web server software (e.g. Apache, ngnix, etc.)  |
+| Windows                  |                                                 |
```
ilyam8, line 171: A lot of classes have no _Groups alarms for_ descriptions. 

- Are we done with `class`ification? if not perhaps it makes sense to not add this table? @joelhans 

- And what's Other? do we use it?

- And i think we need to mentions that if not set we use `Unknown` for all the new attributes.

health/REFERENCE.md
MrZammler: Sorry Thiago for being late to address this.... 

So, I did a test on the info field (I can't find the class/component/type strings on the cloud...), and the cloud will display an info string with e.g. `number !@#$%^&*()_+=-}{][;:?/><,. of metrics lost ....`. I've asked on cloud-ops, will update.

health/REFERENCE.md
MrZammler: Hi Ilya. Yes, we do have some `Other` as class in some alerts. I will add the `Unknown` case in the docs, thanks.

MrZammler: I will add the rest of the descriptions, and go from there with Joel if needed. I suppose it makes sense to keep the table.

ilyam8: Ok, just an idea - consider using _collapsible section_ feature for the tables to keep the doc tidy. Could be just me, feel free to ignore the suggestion if you don't like it.

<details>
<summary>default class attributes</summary>


| Class                    | Groups alarms for:                              |
| ------------------------ | ----------------------------------------------- |
| Ad Filtering             | Services related to Ad Filtering (like pi-hole) |
| Certificates             | 
| Cgroups                  |
| Computing                | 
| Containers               |
| Database                 | Database systems (e.g. MySQL, Postgress, etc)   |
| Data Sharing             |
| DHCP                     |
| DNS                      | 
| Kubernetes               |
| KV Storage               |
| Linux                    | Services specific to Linux (e.g. systemd)       |
| Messaging                |
| Netdata                  | Internal Netdata components monitoring          |
| Other                    |
| Power Supply             |
| Search engine            |
| Storage                  |
| System                   | General system alarms (e.g. cpu, network, etc.) |
| Virtual Machine          | Virtual Machine software                        |
| Web Proxy                |
| Web Server               | Web server software (e.g. Apache, ngnix, etc.)  |
| Windows                  |                                                 |

</details>

MrZammler: Whoa! Neat! Added!

MrZammler:
So I think we're missing an ok or changes from @joelhans and an ok from cloud on the characters.

health/REFERENCE.md
```diff
@@ -129,6 +132,67 @@ You're interested in what comes after the comma: `disk.io`. That's the name of t
...
+
+Specify the classification of the alarm or template. 
+
+Class can be used to indicate the broader area of the system that the alarm applies to. For example, under the general `Database` class, you can group together alarms that operate on various database systems, like `MySQL`, `CockroachDB`, `CouchDB` etc. Example:
+
+```yaml
+class: Database
+```
+<details>
+<summary>Netdata's stock alarms use the following `class` attributes by default, but feel free to adjust for your own requirements:</summary>
```
ilyam8, line 145:
```suggestion
<summary>Netdata's stock alarms use the following `class` attributes by default, but feel free to adjust for your own requirements.</summary>
```

ilyam8: i am wrong, nvm 😄

health/REFERENCE.md
```diff
@@ -129,6 +132,67 @@ You're interested in what comes after the comma: `disk.io`. That's the name of t
...
+| Storage                  | Class for alerts dealing with storage services (storage devices typically live under `System`)   |
+| System                   | General system alarms (e.g. cpu, network, etc.)                                                  |
+| Virtual Machine          | Virtual Machine software                                                                         |
+| Web Proxy                | Web proxy software (e.g. squid)                                                                  |
+| Web Server               | Web server software (e.g. Apache, ngnix, etc.)                                                   |
+| Windows                  | Alerts for monitor of wmi services                                                               |
+
+</details>
+
+If an alarm configuration is missing the `class` line, it's value will default to `Unknown`.
```
joelhans, line 175:
```suggestion
If an alarm configuration is missing the `class` line, its value will default to `Unknown`.
```

health/REFERENCE.md
```diff
@@ -129,6 +132,67 @@ You're interested in what comes after the comma: `disk.io`. That's the name of t
...
+
+#### Alarm line `component`
+
+Component can be used to narrow down what the previous `class` value specifies for each alarm or template. Continuing from the previous example, `component` might include `MySQL`, `CockroachDB`, `MongoDB`, all under the same `Database` classification. Example:
+
+```yaml
+component: MySQL
+```
+As with the `class` line, if `component` is missing from the configuration, it's value will default to `Unknown`.
```
joelhans, line 184:
```suggestion
As with the `class` line, if `component` is missing from the configuration, its value will default to `Unknown`.
```

health/REFERENCE.md
```diff
@@ -129,6 +132,67 @@ You're interested in what comes after the comma: `disk.io`. That's the name of t
...
+
+#### Alarm line `type`
+
+This indicates the type of error (or general problem area) that the alarm or template applies to. For example, `Latency` can be used for alarms that trigger on latency issues, be it in network interfaces, web servers, or database systems. Example:
```
joelhans, line 188:
```suggestion
This indicates the type of error (or general problem area) that the alarm or template applies to. For example, `Latency` can be used for alarms that trigger on latency issues in network interfaces, web servers, or database systems. Example:
```

joelhans (commented):
Just a few small tweaks. Overall it looks good.

joelhans (approved):

ilyam8 (approved):

health/REFERENCE.md
MrZammler: Furthermore, I've received an alert email from the cloud, after having set a class attribute to ` Classification: This is a class !@#$%^*()_&+=-}{][:;?><,./ ` and it correctly appeared in the email message:

thiagoftsm (approved):
After last explanation from @MrZammler , LGTM!

Issue #10968, opened 2021-04-13
Add netdata uptime metric and export it as a special (global) variable (health)


ilyam8:
##### Feature idea summary

Idea is to add netdata process uptime chart to the `Netdata Monitoring->netdata`. It will allow us to see if the netdata process was restarted.

In addition i suggest to export netdata process uptime as [a health special variable](https://learn.netdata.cloud/docs/agent/health/reference#variables) (e.g `$uptime` or `$nd_uptime`). It would be very handy in some alarms (e.g: don't trigger for N minutes after netdata restart). [Netdata update triggers ipv4.tcphandshake alerts](https://github.com/netdata/netdata/issues/9799) issue is a good example.

##### Expected behavior

- new chart `Netdata Monitoring->netdata->uptime`
- new health global variable `$uptime` or `$nd_uptime`.

Keep in mind that this is just an idea, if there are any problems with it feel free to comment.

ilyam8:
@vlvkobal @thiagoftsm what do you think, guys?

thiagoftsm:
> @vlvkobal @thiagoftsm what do you think, guys?

I think it makes sense, because we also have a chart monitoring how long the computer is up `system.uptime`, this additional information would give for users how long their system ran without monitoring.

Ferroin:
Agreed, being able to determine downtime on the Netdata agent itself is exceedingly useful.

Additionally, being able to have alarms behave differently just after a restart is exceedingly useful (perhaps we should also export the system updtime as a special variable for alarms to look at for similar reasons), and it would also be useful to be able to warn on a restart of the agent without an associated system restart (though I do not think we should provide such an alarm by default).

stelfrag:
This is good to have !!! We will need to have that information also on the parent for all the children.

vlvkobal:
The idea is to add a separate internal plugin for Netdata uptime and make it impossible to disable it. We don't need the health variable because we can access the value from the `netdata.uptime` chart in alarms.

ilyam8:
> we can access the value from the netdata.uptime chart

That sounds good!

ilyam8:
Ok, looks like netdata global statistics is actually a module of the proc plugin

https://github.com/netdata/netdata/blob/27552389682e7162b52ea1a69683f2e457ad2dff/collectors/proc.plugin/plugin_proc.c#L189

We decided to try to move it out from proc and make it a separate plugin. Then we can easily add netdata process uptime to it.

[...]

PR #10991, opened 2021-04-19
Move global stats to a separate thread

### LINKED ISSUE #10510: lack of dbengine statistics if proc plugin is disabled (`Netdata Monitoring->dbengine`)
##### Bug report summary

Discovered in #10423, see [the comment](https://github.com/netdata/netdata/issues/10423#issuecomment-760223846).

Use case: netdata in k8s - Netdata parent has all collectors disabled, because children instances are responsible for data collection. However parent instance maintains the database.

I think the only requirement for `Netdata Monitoring->dbengine` is enabled dbegine memory mode. 

Currently those statistics populated by [global_statistics_charts](https://github.com/netdata/netdata/blob/de6035c543296de179e359a415664106d66c3878/daemon/global_statistics.c#L174) function which seems doesnt executed if there are no active collectors.

##### OS / Environment

any

##### Netdata version

`netdata v1.28.0-134-g6dade0cd0`
 

##### Component Name

`daemon`

##### Steps To Reproduce

1. Parent-child setup, parent memory mode is dbengine, all collectors on the parent instance are disabled.

##### Expected behavior

Always have dbengine statistics if memory mode is dbengine.

vlvkobal:
##### Summary
Global Netdata statistics was run from `proc`, `freebsd`, or `macos` plugins, but we need it to be independent and be present if the main system stats plugin is disabled.

The PR is the first step towards the implementation of Netdata uptime chart #10968.

Fixes #10510

##### Component Name
`proc`, `freebsd`, `macos`, and `global statistics` plugins.

##### Test Plan
Check charts in the "Netdata Monitoring" menu section. 
Check "plugin CPU usage" and "plugin modules durations" charts on Linux, FreeBSD, and macOS.


ilyam8 (approved):

ilyam8:
@vlvkobal thanks, good work 👍 global statistics no longer depend on proc plugin!

thiagoftsm (approved):
I tested on `Slackware current` and `FreeBSD 13.0`, and everything worked as expected.
I cannot test on MAC, but considering what I saw on code, I expect it will run there.

PR #11003, opened 2021-04-20
improvements to anomalies collector following dogfooding

andrewm4894:
##### Summary

- add check for python 3.
- add ability to reinitialize models and data regulary if needed (off by default).
- change check() to be less error prone when running on a parent node (direct learning from dogfooding on netdata parent).
- add some model initialization functions to enable reinitialization if model gets into a "bad state".
- add some logic to check for models that have not been trained and reinitialize and train them as needed.
- add logic to `validate_charts()` to maintain dimensions by adding and removing as needed.

##### Component Name
collectors/python.d.plugin/anomalies

##### Test Plan

Tested on local development servers and dogfooding the branch on netdata ml parent node.


collectors/python.d.plugin/anomalies/anomalies.conf
```diff
@@ -44,6 +44,9 @@ local:
     # Use http or https to pull data
     protocol: 'http'
 
+    # SSL verify parameter for requests.get() calls
+    verify: true
```
ilyam8, line 48: perhaps `tls_verify`?

ilyam8: so the name reflects the option meaning

ilyam8 (approved):

thiagoftsm (approved):

joelhans (approved):

PR #11009, opened 2021-04-21
Framework and basic UI test covering presence of elements for a generic installation

kaskavel:
##### Summary

We add support for running UI tests for the agent dashboard using Cypress. The tests are running independently as part of a new workflow.


novykh (approved):

dimko (dismissed):

.github/workflows/review.yml
```diff
@@ -23,13 +23,11 @@ jobs:
           if git diff --name-only origin/${{ github.base_ref }} HEAD | grep -Eq '*\.js|node\.d\.plugin\.in' ; then
             echo 'run_eslint=1' >> $GITHUB_ENV
           fi
-      - name: Run eslint
+      - name: Install eslint
+        run: npm install eslint -D
```
Ferroin, line 27: This needs the same `if: env.run_eslint == 1` conditional as the next step has, otherwise we’ll always install eslint instead of only doing so when we actually need to.

.github/workflows/review.yml
```diff
@@ -23,13 +23,11 @@ jobs:
...
+      - name: Run ESLint
         if: env.run_eslint == 1
-        uses: reviewdog/action-eslint@v1
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          reporter: github-pr-check
-          eslint_flags: '.'
+        run: ./node_modules/.bin/eslint . --ext .js,.jsx,.ts,.tsx
```
Ferroin, line 30: If possible I’d like to keep using the ReviewDog actions, as they provide proper annotations on the diff itself for any warnings. Simply ensuring ESLint is installed and properly updating the flags we are passing in _should_ be enough to make it work correctly.

.github/workflows/review.yml
```diff
@@ -23,13 +23,11 @@ jobs:
           if git diff --name-only origin/${{ github.base_ref }} HEAD | grep -Eq '*\.js|node\.d\.plugin\.in' ; then
```
Ferroin, line 23: If we’re going to be linting other file types, we need to update this regex to match those files.

.eslintignore
```diff
@@ -1 +1,2 @@
 **/*{.,-}min.js
+tests/cypress/**/*.js
```
Ferroin, line 2: Are we sure we do not want to lint our test infrastructure?

.github/workflows/review.yml
kaskavel: I decided after all, not to mix the eslint changes needed with this PR, as the purpose of this addition is to add a simple UI test and an action running it. Regarding linting, we will treat it in a subsequent PR, after discussing it with the FE team.

dimko (dismissed):

kaskavel:
@Ferroin any objections on merging this one? (more complex tests will follow)

.github/workflows/ui_tests.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git clone repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Install the agent
+        run: >-
+          sudo apt-get install zlib1g-dev uuid-dev libuv1-dev liblz4-dev libjudy-dev libssl-dev libelf-dev libmnl-dev
+          gcc make git autoconf autoconf-archive autogen automake pkg-config curl python cmake &&
```
Ferroin, line 23: We almost certainly want to be running `packaging/installer/install-required-packages.sh netdata --non-interactive--dont-wait` here instead of this, as that will ensure we properly track dependency changes.

Ferroin (approved):

.github/workflows/ui_tests.yml
kaskavel: Done :)

dimko (approved):

.github/workflows/ui_tests.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git clone repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Install the agent
+        run: |
+          ./packaging/installer/install-required-packages.sh netdata --non-interactive --dont-wait
+          sudo ./netdata-installer.sh --dont-wait
```
ilyam8, line 23: Should we add `--disable-telemetry`? Otherwise, if I understand it right, we will send statistics to the PostHog on every job run.

ilyam8:
@Ferroin perhaps we always need to disable telemetry in ci jobs if it is not explicitly enabled. We can use `CI` environment variable (i checked [GHA](https://docs.github.com/en/actions/reference/environment-variables) and [Travis CI](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables), both of them always have `CI` set to `true`)

.github/workflows/ui_tests.yml
Ferroin: Excellent catch.

Ferroin:
> @Ferroin perhaps we always need to disable telemetry in ci jobs if it is not explicitly enabled (optional, probably we never need it enabled in ci). We can use `CI` environment variable (i checked [GHA](https://docs.github.com/en/actions/reference/environment-variables) and [Travis CI](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables), both of them always have `CI` set to `true`)

Yeah, this sounds like a good idea in general. I _think_ we already have most of the CI runs that would be sending telemetry opted out, but not all of them.

Alternatively, perhaps globally set `DO_NOT_TRACK=1` on all of our CI? That’s a bit quicker to implement and avoids the need for special handling for the case of CI in the scripts and tooling.

ilyam8:
> Alternatively, perhaps globally set DO_NOT_TRACK=1 on all of our CI?

This will do, but basically, there is the same problem as with `--disable-telemetry` - if you forgot to add it...

If we never ever need to send analytics when running in ci we can hardcode it in the `c` code (check `CI` env var, if set and `true` - disable).

I suspect it is very easy to implement. @MrZammler can help for sure.

Ferroin:
> > Alternatively, perhaps globally set DO_NOT_TRACK=1 on all of our CI?
> 
> This will do, but basically, there is the same problem as with `--disable-telemetry` - if you forgot to add it...
> 
> If we never ever need to send analytics when running in ci we can hardcode it in the `c` code (check `CI` env var, if set and `true` - disable).
> 
> I suspect it is very easy to implement. @MrZammler can help for sure.

It’s more than just the C code that will need it though, the new updater and installer code will also be handling telemetry.

ilyam8:
Hm, I briefly checked the code and I see that if we set `netdata_anonymous_statistics_enabled` to 0 there is no way we send any statistics:
-  [we don't start analytics thread](https://github.com/netdata/netdata/blob/9f1c1bddf40d175619cb73102ab880f80cd4e6bf/daemon/main.c#L1298-L1308)
-  [`send_statistics` function just returns](https://github.com/netdata/netdata/blob/5de2432537ba4300c2c2160efdd017d828f23937/daemon/analytics.c#L835-L863) (not start, crash, etc reports)
- we execute `anonymous-statistics.sh` script only from `daemon/analytics.c:send_statistics()`

Why do we need to update the installer/updater code? (you said the **new** updater, I am not aware of it)

Ferroin:
> Hm, I briefly checked the code and I see that if we set `netdata_anonymous_statistics_enabled` to 0 there is no way we send any statistics:
> 
> * [we don't start analytics thread](https://github.com/netdata/netdata/blob/9f1c1bddf40d175619cb73102ab880f80cd4e6bf/daemon/main.c#L1298-L1308)
> * [`send_statistics` function just returns](https://github.com/netdata/netdata/blob/5de2432537ba4300c2c2160efdd017d828f23937/daemon/analytics.c#L835-L863) (no start, crash, etc reports)
> * we execute `anonymous-statistics.sh` script only from `daemon/analytics.c:send_statistics()`
> 
> Why do we need to update the installer/updater code? (you said the **new** updater, I am not aware of it)

There’s ongoing work on a new installer and updater that rolls support for binary packages, static installs, and local builds into a single installation script and single update script, and there are plans to have those report telemetry about the installation and update processes.

ilyam8:
Thanks for the explanation, didn't know that.

Does disabling telemetry in the `c` code when `CI` env var set sound good for you? For me, it looks ok because it is easy to implement (read as can be done and not forgotten) and it helps us to avoid problems with analytics. My assumption that we never want any analytics from ci.

Ferroin:
> Thanks for the explanation, didn't know that.
> 
> Does disabling telemetry in the `c` code when `CI` env var set sound good for you? For me, it looks ok because it is easy to implement (read as can be done and not forgotten) and it helps us to avoid problems with analytics. My assumption that we never want any analytics from ci.

The issue I see is that while we never want analytics from _our_ CI, we probably do want analytics if anybody else is using Netdata in CI as part of testing other things.

.github/workflows/ui_tests.yml
Ferroin: See the changes in https://github.com/netdata/netdata/pull/11233 for an example of what to change here to just disable telemetry for the whole workflow.

cakrit:
- Disable telemetry for now. 
- Put an option _somewhere_ which will allow the messages to have a special flag, indicating the events and the pageviews come from us. When that's in place, let me and Andy know and we'll show those events separately.

PR #11010, opened 2021-04-21
anonymous-statistics: add a timeout when using `curl`

ilyam8:
##### Summary

This PR adds `--max-time=2` timeout.

I set it to 2 because 
- it [was 2](https://github.com/netdata/netdata/blob/9580ed4d13125adac01483bac14a18c952afd13e/daemon/anonymous-statistics.sh.in#L33) before switching to [PostHog](https://github.com/netdata/netdata/pull/10636)
- @andrewm4894 didn't change `wget` timeout ([it is 1](https://github.com/netdata/netdata/blob/b6d729f96a0b1e17abffb2cf69d1ff29e62004f4/daemon/anonymous-statistics.sh.in#L85-L87))

##### Component Name

`anonymous-statistics`

### DIFF
--- a/daemon/anonymous-statistics.sh.in
+++ b/daemon/anonymous-statistics.sh.in
@@ -80,7 +80,7 @@ EOF
 
 # send the anonymous statistics to the Netdata PostHog
 if [ -n "$(command -v curl 2> /dev/null)" ]; then
-  curl -X POST --header "Content-Type: application/json" -d "${REQ_BODY}" https://posthog.netdata.cloud/capture/ > /dev/null 2>&1
+  curl -X POST --max-time 2 --header "Content-Type: application/json" -d "${REQ_BODY}" https://posthog.netdata.cloud/capture/ > /dev/null 2>&1
 else
   wget -q -O - --no-check-certificate \
   --method POST \


thiagoftsm (approved):
LGTM!

andrewm4894 (approved):
LGTM

ilyam8:
@andrewm4894 @stelfrag @Ferroin we may need to reconsider anon stats timeouts.

It is 1 for `wget` and 2 for `curl`.

andrewm4894:
> Could be ok for GA and not ok for PostHog 🤷‍♂️

I'll double check with the PostHog guys to see what they think. Its a [post only public endpoint](https://posthog.com/docs/api/post-only-endpoints) that does not really return anything so i guess should be ok but will confirm, and see if we can monitor latency on it in someway.

ilyam8:
> I'll double check with the PostHog guys to see what they think.

Well, it always depends on a server's distance from the service, internet connection quality, etc.

andrewm4894:
> Well, it always depends on a server's distance from the service, internet connection quality, etc.

Yep - they said pretty much same thing:

> On posthog cloud the endpoint has a 20ms p90 latency (10ms p50 and 100ms p99)...so unless network traffic takes more than a second, for the API these timeouts are fine.

PR #11012, opened 2021-04-22
Functions to store and retrieve node_id for a host

stelfrag:
##### Summary
- Update the node id for a host
- Check or retrieve the node id for a host

##### Component Name
database
aclk

##### Test Plan
- TBA


stelfrag:
Due to freq changes, closing for now to avoid triggering the CI all the time

Issue #10842, opened 2021-03-23
ACLK-NG Mac OS compatibility


underhood:
##### Bug report summary

ACLK-NG can't build on Mac OS due to few small issues:
-> `endian.h` missing including necessary functions like `be16toh`
-> use of undeclared identifier 'CLOCK_BOOTTIME'
-> implicit declaration of function 'pipe2' (mac knows only `pipe`)

##### OS / Environment
```
La wonderful fruity OS
```

##### Component Name
ACLK-NG
##### Steps To Reproduce
`netdata-installer --aclk-ng` on Mac OS

##### Expected behavior
It builds and works automagically

OdysLam:
Do we have any update on this @underhood ? It's a blocker to enable netdata-cloud in homebrew installations.

underhood:
@OdysLam I have been asking around about getting Mac VM for development...

OdysLam:
We had internal chat with @underhood and @kaskavel. Let's keep this thread to sync about any macOS VM and testing. 

Looking forward for the fixes ✌️

chenrui333:
Any update on this thread?

Conna576:
I honestly don't think I have this installed right.  It all makes no sense.  😞

underhood:
@chenrui333 we are working on this. There is already one PR that improves things considerably and another one incoming. Currently, we are in code freeze ahead of release. They will go in ASAP as the code freeze is lifted.

PR #11013, opened 2021-04-22
Add installation instructions for MacPorts

gpanders:
##### Summary

Netdata was recently added to MacPorts! This PR adds installation instructions for MacPorts users.

##### Component Name

Website.

##### Test Plan

N/A

##### Additional Information

None

### DIFF
--- a/packaging/installer/methods/macos.md
+++ b/packaging/installer/methods/macos.md
@@ -14,6 +14,7 @@ We recommend installing Netdata with the community-created and -maintained [**Ho
 package**](#install-netdata-with-the-homebrew-package). 
 
 -   [Install Netdata via the Homebrew package](#install-netdata-with-the-homebrew-package)
+-   [Install Netdata via MacPorts](#install-netdata-with-macports)
 -   [Install Netdata from source](#install-netdata-from-source)
 
 ## Install Netdata with the Homebrew package
@@ -33,6 +34,25 @@ brew install netdata
 > Homebrew will place your Netdata configuration directory at `/usr/local/etc/netdata/`. Use the `edit-config` script
 > and the files in this directory to configure Netdata. For reference, you can find stock configuration files at
 > `/usr/local/Cellar/netdata/{NETDATA_VERSION}/lib/netdata/conf.d/`.
+
+Skip on ahead to the [What's next?](#whats-next) section to find links to helpful post-installation guides.
+
+## Install Netdata with MacPorts
+
+To install Netdata with MacPorts, simply run:
+
+```bash
+sudo port install netdata
+```
+
+You can have MacPorts automatically run the `netdata` service in the background and on restarts with:
+
+```bash
+sudo port load netdata
+```
+
+Your Netdata config file will be placed at `/opt/local/etc/netdata/netdata.conf`. You can find stock configuration
+files at `/opt/local/lib/netdata/conf.d/`.
 
 Skip on ahead to the [What's next?](#whats-next) section to find links to helpful post-installation guides.
 


joelhans:
Thanks for this! The docs LGTM me at first glance, but I'm pinging some extra people here in case they want to chime in with feedback or test out the process themselves. @OdysLam is very invested in the macOS installation story, so he'll want to check this out, and @Ferroin handles our packaging overall.

Lately we've been discussing how to handle official vs. community-driven installation methods in the documentation, so this PR will be a good forcing function for us to decide on a solution.

OdysLam:
Hey @gpanders, 

Before we move forward, I want to test if the Agent illustrates 2 of the problems that exist with brew, namely:
1) Packaged with old dashboard (newer is downloaded from netdata/dashboard)
2) Doesn't work with Netdata Cloud. 

Your contribution is not only awesome but very timely since as @joelhans mentioned is a subject that we are discussing internally at the moment. Please bear with us as we try to chart a strategy. You rock eitherway ✌️

gpanders:
> Before we move forward, I want to test if the Agent illustrates 2 of the problems that exist with brew, namely:
> 
>     1. Packaged with old dashboard (newer is downloaded from netdata/dashboard)
> 
>     2. Doesn't work with Netdata Cloud.

Yes I believe both of these are the case with MacPorts as well. Like brew, MacPorts is affected by #10842 which means we also are currently providing version 1.29.3. We are also unable to compile with Netdata Cloud support because the `configure` script [insists on using a static library for `libmosquitto`](https://github.com/netdata/netdata/blob/master/configure.ac#L653). MacPorts provides `libmosquitto` as a shared library but currently there is no mechanism to provide libraries statically.

We could work around this by compiling libmosquitto statically as one of the build steps in our Netdata port, but that is unconventional and we wouldn't go down that route unless 1) we knew for certain that Netdata *needs* libmosquitto to be statically compiled for some reason and 2) we had enough users requesting that feature to make it worth it.

OdysLam:
@gpanders -- Thanks for the detailed response. You greatly saved me a bit of time from manually testing it myself. 

For the Cloud, we have released a version of the component that required `libmosquitto` that no longer requires it. While the new version can be built on Linux successfully, on MacOS we have a couple of bugs that we are root causing (cc @underhood ). At the moment we can't offer a timeframe.

For the dashboard, check our [FAQ](https://community.netdata.cloud/t/i-have-an-old-version-of-the-netdata-agent-dashboard/1207). You only need to download the latest release from netdata/dashboard and place the files into the correct directory. We are working on this front as well in order to make a long-term solution. cc @Ferroin for this. 

Again, we are humbled by your contribution and for taking the time to discuss these issues. Rest assured that they are top of mind, amongst a myriad other product and engineering-related matters.

gpanders:
> For the dashboard, check our [FAQ](https://community.netdata.cloud/t/i-have-an-old-version-of-the-netdata-agent-dashboard/1207). You only need to download the latest release from netdata/dashboard and place the files into the correct directory. We are working on this front as well in order to make a long-term solution. cc @Ferroin for this.

Thanks for pointing this out -- I wasn't aware that the dashboard was now hosted separately. Our port should be updated [soon](https://github.com/macports/macports-ports/pull/10763).

OdysLam:
Thanks @gpanders for improving your port. Rest assured that we are working towards solving both.

packaging/installer/methods/macos.md
```diff
@@ -36,6 +37,25 @@ brew install netdata
...
+```
+
+You can have MacPorts automatically run the `netdata` service in the background and on restarts with:
+
+```bash
+sudo port load netdata
+```
+
+Your Netdata config file will be placed at `/opt/local/etc/netdata/netdata.conf`. You can find stock configuration
+files at `/opt/local/lib/netdata/conf.d/`.
```
ilyam8, line 55: Is this path valid for m1 devices?

packaging/installer/methods/macos.md
```diff
@@ -36,6 +37,25 @@ brew install netdata
 
 Skip on ahead to the [What's next?](#whats-next) section to find links to helpful post-installation guides.
 
+## Install Netdata with MacPorts
+
+To install Netdata with MacPorts, simply run:
```
ilyam8, line 42: Let's add a link to the https://ports.macports.org/port/netdata/ where users can check the Agent version/additional info.

iigorkarpov (commented):
I'd propose to make both descriptions similar:
```
-   [Install Netdata via the Homebrew package](#install-netdata-with-the-homebrew-package)
 -   [Install Netdata via MacPorts](#install-netdata-with-macports)
```
 Also, bash is not needed for Macports. Not sure about Homebrew though.

ilyam8:
@iigorkarpov I'm not quite clear what you mean

- They are similar, see https://github.com/gpanders/netdata/blob/5dffa80ca115032bbdebce0ddc660a4f6b33c4da/packaging/installer/methods/macos.md?plain=1#L16-L17.
- I don't see "bash is required" anywhere.

packaging/installer/methods/macos.md
gpanders: Yes, MacPorts uses the same prefix on all devices.

ilyam8:
Closing because [MacPorts Netdata](https://ports.macports.org/port/netdata/) is outdated (v 1.29.3). I think it is not an option for now.

PR #11020, opened 2021-04-23
Provide more agent analytics to posthog

MrZammler:
##### Summary

The purpose of this PR is to provide more data to posthog analytics, from various agent information.

All the current statistics (START/EXIT, etc) will remain, and this PR will add a new one (META) that will be sent 2 minutes after the agent has started, to allow for gathering data. After the initial META event, the agent will continue to send META events every 2 hours.

##### Component Name

Analytics

##### Test Plan

Compile an agent with this PR.
Observe that statistics are sent to posthog.

##### Additional Information

This PR is mostly the same as https://github.com/netdata/netdata/pull/10887. @ilyam8 and @vlvkobal though found an error when the agent would not be able to `popen` and run plugins. The `error.log` would list multiple `Cannot popen("exec...` errors.

It appears that doing `setenv` (which was the previous PR way of passing data to the `anonymous-statistics.sh`) was messing with the threads. This PR now passes those data as parameters to the script, in order to avoid setting environmental variables, and should not create such problems, but it would be good to test it against them again.

New commits from the last PR are https://github.com/netdata/netdata/commit/06211195af235470d8b976653b5823a099bc891e and https://github.com/netdata/netdata/commit/28e7ea2e4306fe53cb2d1cf8e517a2ec0609f2f2


daemon/analytics.h
```diff
@@ -0,0 +1,89 @@
...
+#define ANALYTICS_MAX_SHELL_HITS 255
+#define ANALYTICS_MAX_JSON_HITS 255
+#define ANALYTICS_MAX_DASHBOARD_HITS 255
+
+#define NETDATA_PLUGIN_HOOK_ANALYTICS \
+    { \
+        .name = "ANALYTICS", \
+        .config_section = NULL, \
+        .config_name = NULL, \
+        .enabled = 0, \
```
vlvkobal, line 25: Should it be disabled by default?

daemon/main.c
```diff
@@ -1497,8 +1305,23 @@ int main(int argc, char **argv) {
...
+        for (i = 0; static_threads[i].name != NULL; i++) {
+            if (!strncmp(static_threads[i].name, "ANALYTICS", 9)) {
+                struct netdata_static_thread *st = &static_threads[i];
+                st->thread = mallocz(sizeof(netdata_thread_t));
+                st->enabled = 1;
+                debug(D_SYSTEM, "Starting thread %s.", st->name);
+                netdata_thread_create(st->thread, st->name, NETDATA_THREAD_OPTION_DEFAULT, st->start_routine, st);
+            }
+        }
+    }
```
vlvkobal, line 1323: What for do you need this part? `netdata_anonymous_statistics_enabled` is always -1.

MrZammler: Hi Vlad! Well, `netdata_anonymous_statistics_enabled` starts initially as `0` (defined as a global variable, on top of main.c). At some point in main.c (https://github.com/MrZammler/netdata/blob/28e7ea2e4306fe53cb2d1cf8e517a2ec0609f2f2/daemon/main.c#L1260), it gets the value of `-1`, i.e. we don't know if we're gonna send analytics. The first call to `send_statistics` will decide if the value will be `1` (i.e. send) or `0` (dont send).

The point of this (and having the analytics thread as `enabled = 0` is to only run the thread if we are indeed going to send analytics.

vlvkobal: Oh, sorry, my bad! I didn't notice that it is changed in `netdata_anonymous_statistics_enabled`.

Then I have another question. Why did you add `NETDATA_PLUGIN_HOOK_ANALYTICS` to `static_threads`? It won't start anyway.

MrZammler: Well, the first iterations of the PR always started the analytics thread... But, It's there also because the agent (if the thread eventually runs) will handle it the same way as other threads (send cancel thread, etc).

vlvkobal (approved):

ilyam8:
> multiple Cannot popen("exec... errors.

It no longer happens, good work @MrZammler 👍

thiagoftsm (approved):
I did not retest, but the code is ok according my previous reviews.

stelfrag (approved):

MrZammler:
Thank you all!

PR #11035, opened 2021-04-26
Improve installation method for Alpine

tiramiseb:
##### Summary

Improve installation method for Alpine

##### Component Name

Documentation

##### Additional Information

- the `python` package doesn't exist since Alpine 3.5 (dec 2016), use `python3` instead
- create a proper start script instead of using `/etc/local.d`

### DIFF
--- a/packaging/installer/methods/alpine.md
+++ b/packaging/installer/methods/alpine.md
@@ -9,7 +9,7 @@ Execute these commands to install Netdata in Alpine Linux 3.x:
 
 ```sh
 # install required packages
-apk add alpine-sdk bash curl libuv-dev zlib-dev util-linux-dev libmnl-dev gcc make git autoconf automake pkgconfig python logrotate
+apk add alpine-sdk bash curl libuv-dev zlib-dev util-linux-dev libmnl-dev gcc make git autoconf automake pkgconfig python3 logrotate
 
 # if you plan to run node.js Netdata plugins
 apk add nodejs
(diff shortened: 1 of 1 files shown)


packaging/installer/methods/alpine.md
```diff
@@ -21,16 +21,20 @@ cd netdata
...
-rc-update add local
+depend() {
+        need net localmount
+        after firewall
+}
+EOF
 ```
 
+If you have installed Netdata in another directory, you have to change the content of the `command` variable in that script.
+
```
Ferroin, line 39: Actually, none of this should be needed. We include an OpenRC init script as part of the regular install that _should_ be getting installed correctly on Alpine (I have not checked recently) as `/etc/init.d/netdata`.

Ferroin: To clarify, I’m arguing that the whole section about manually setting up the startup/shutdown should be removed, not that we shouldn’t update what’s already there to use a proper init script instead of a local stub.

tiramiseb: There was no `/etc/init.d/netdata` file after following these instructions (installation done less than 1/2h before the PR), that's why I suggested the change...

tiramiseb:
hah! I just noticed there is a netdata package in the community repo. It would be great to mention it on this page...

packaging/installer/methods/alpine.md
Ferroin: Ah, OK, that’s a bug then.

In that case, I’m all for adding this for now, but we should also open an issue about the fact that the OpenRC initscript is not being installed on Alpine.

Ferroin (approved):

joelhans (approved):

packaging/installer/methods/alpine.md
ilyam8: > In that case, I’m all for adding this for now, but we should also open an issue about the fact that the OpenRC initscript is not being installed on Alpine.

Kind reminder to do it if you think we need it!

PR #11048, opened 2021-04-26
Improve storage number unpacking by using a lookup table.

vkalintiris:
The LUT contains precomputed values of multipliers/divisors that are
used to unpack storage numbers into calculated numbers efficiently.

##### Summary

Performance improves by at least 35% when testing with random integers
in `uint32_t`s entire range. Unpacking smaller random values (eg. in the
range [0,1000]) is faster by 200-300%.

##### Test Plan

1. Compare calculated_numbers in old vs new version,
2. Run agent and check dashboards for any irregularities.

### DIFF
--- a/libnetdata/storage_number/storage_number.c
+++ b/libnetdata/storage_number/storage_number.c
@@ -91,15 +91,32 @@ RET_SN:
     return r;
 }
 
+// Lookup table to make storage number unpacking efficient.
+static calculated_number lut10x[4 * 8];
+
+__attribute__((constructor)) void initialize_lut(void) {
+    // The lookup table is partitioned in 4 subtables based on the
+    // values of the factor and exp bits.
+    for (int i = 0; i < 8; i++) {
+        // factor = 0
+        lut10x[0 * 8 + i] = 1 / pow(10, i);    // exp = 0
+        lut10x[1 * 8 + i] = pow(10, i);        // exp = 1
+
+        // factor = 1
+        lut10x[2 * 8 + i] = 1 / pow(100, i);   // exp = 0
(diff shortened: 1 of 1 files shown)


vkalintiris:
Benchmark source code: https://gist.github.com/vkalintiris/2fc64b8b444dcbba2216e363d6121a87

libnetdata/storage_number/storage_number.c
```diff
@@ -81,23 +81,40 @@ storage_number pack_storage_number(calculated_number value, uint32_t flags) {
     return r;
 }
 
+// Lookup table to make storage number unpacking efficient.
+static calculated_number lut10x[4 * 8];
+
+__attribute__((constructor)) void initialize_lut(void) {
```
vkalintiris, line 87: I don't think there exists a compiler out there that manages to build netdata successfully without already supporting the constructor attribute. If you don't like this I can make this function visible in a header and call it from main.

vlvkobal (dismissed):

libnetdata/storage_number/storage_number.c
underhood: does this include MS compiler? we plan in future to move to Win

vkalintiris: `msvc` won't compile this code whereas `clang-cl` will.

underhood: ok keep it as it is for now I guess. we will deal with this later

underhood (dismissed):

vlvkobal (approved):

stelfrag (approved):

PR #11051, opened 2021-04-27
Provide UTC offset in seconds and edit health config command

MrZammler:
##### Summary

For the cloud to be able to include an agent local time in the alerts email notification, the agent needs to send the UTC offset in seconds for each host (parent and/or children), together with the abbreviated time zone format. It also constructs and sends an  `edit_command` string, which when parsed by the cloud, will appear on the email message to provide the user with a copy-paste command to use to edit the alert's configuration:

Screenshot from 2021-04-27 16-55-56

The UTC offset is calculated based on the string returned from `strftime` using the `%z` field, i.e. from a string with format `+hhmm` or `-hhmm`. For example, for `+0300` the UTC offset will be `10800`.

The `edit_command` string is constructed from the `source` string, and includes the line number of the configuration file to edit. Example: `sudo /etc/netdata/edit-config health.d/exporting.conf=15`. The string will be split at the `=` character.

The `edit_command` will also be used by the agent in it's own email alarm messages (will come in next PR).

##### Component Name

Health

##### Test Plan

The `status-change` aclk message for an alarm, should contain `utc_offset`, `timezone` and `edit_command`.

To make sure that e.g. a child's `utc-offset` and `timezone` is correctly included in the parent's `host` entry, the child's `/etc/localtime` could be temporarly changed to point to e.g. `/usr/share/zoneinfo/America/Havana`.

If both a child and and the parent are running an agent with this PR, then each of the hosts on the parent will have it's own `abbrev_timezone` and `utc_offset` (if they are in different timezones of course).

If a child is running an older version of the agent, then the parent should set `UTC` as the abbreviated timezone, and `0` as the utc_offset for that child.

If the parent is running an older version of the agent, then in the `error.log` it will list `request has parameter 'abbrev_timezone' = 'CDT', which is not used.` but streaming should work as usual.

##### Additional Information

The json entry is called `timezone` even though it contains the abbreviated one; it is already agreed upon with the cloud so it can not change now, perhaps it could be changed with the new architecture messages.

The `edit_command` string includes `sudo`. It is not required if the agent is installed and run under another user and not root, but it doesn't harm either. Even new configurations created with sudo, will have the correct permissions. However, if in some cases it is a problem (maybe when the agent runs as an X user, and the config must be edited from another?), then the function should check and maybe add sudo or not.

The UTC offset is calculated on agent start. That means it will have wrong value when daylight savings time kicks in, and will need an agent restart to re-calculate.

### DIFF
--- a/daemon/analytics.c
+++ b/daemon/analytics.c
@@ -696,6 +696,46 @@ static void get_system_timezone(void)
         timezone = "unknown";
 
     netdata_configured_timezone = config_get(CONFIG_SECTION_GLOBAL, "timezone", timezone);
+
+    //get the utc offset, and the timezone as returned by strftime
+    //will be sent to the cloud
+    //Note: This will need an agent restart to get new offset on time change (dst, etc).
+    {
+        time_t t;
+        struct tm *tmp, tmbuf;
+        char zone[FILENAME_MAX + 1]; //can be really less, but %z could be abbriviated or long
+        char sign[2], hh[3], mm[3];
+
+        t = now_realtime_sec();
+        tmp = localtime_r(&t, &tmbuf);
+
+        if (tmp != NULL) {
+            if (strftime(zone, FILENAME_MAX, "%Z", tmp) == 0) {
+                netdata_configured_abbrev_timezone = strdupz("UTC");
+            } else
+                netdata_configured_abbrev_timezone = strdupz(zone);
+
+            if (strftime(zone, FILENAME_MAX, "%z", tmp) == 0) {
+                netdata_configured_utc_offset = 0;
+            } else {
+                sign[0] = zone[0] == '-' || zone[0] == '+' ? zone[0] : '0';
+                sign[1] = '\0';
+                hh[0] = isdigit(zone[1]) ? zone[1] : '0';
+                hh[1] = isdigit(zone[2]) ? zone[2] : '0';
+                hh[2] = '\0';
+                mm[0] = isdigit(zone[3]) ? zone[3] : '0';
+                mm[1] = isdigit(zone[4]) ? zone[4] : '0';
+                mm[2] = '\0';
+
+                netdata_configured_utc_offset = (atoi(hh) * 60 * 60) + (atoi(mm) * 60);
+                netdata_configured_utc_offset =
+                    sign[0] == '-' ? -netdata_configured_utc_offset : netdata_configured_utc_offset;
+            }
+        } else {
+            netdata_configured_abbrev_timezone = strdupz("UTC");
+            netdata_configured_utc_offset = 0;
+        }
+    }
 }
 
 void set_global_environment()
--- a/daemon/analytics.c
+++ b/daemon/analytics.c
@@ -703,7 +703,7 @@ static void get_system_timezone(void)
     {
         time_t t;
         struct tm *tmp, tmbuf;
-        char zone[FILENAME_MAX + 1]; //can be really less, but %z could be abbriviated or long
+        char zone[FILENAME_MAX + 1];
         char sign[2], hh[3], mm[3];
 
         t = now_realtime_sec();
--- a/daemon/analytics.c
+++ b/daemon/analytics.c
@@ -727,7 +727,7 @@ static void get_system_timezone(void)
                 mm[1] = isdigit(zone[4]) ? zone[4] : '0';
                 mm[2] = '\0';
 
-                netdata_configured_utc_offset = (atoi(hh) * 60 * 60) + (atoi(mm) * 60);
+                netdata_configured_utc_offset = (str2i(hh) * 3600) + (str2i(mm) * 60);
                 netdata_configured_utc_offset =
                     sign[0] == '-' ? -netdata_configured_utc_offset : netdata_configured_utc_offset;
             }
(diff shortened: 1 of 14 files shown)


daemon/analytics.c
```diff
@@ -688,6 +688,46 @@ static void get_system_timezone(void)
...
+                sign[0] = zone[0] == '-' || zone[0] == '+' ? zone[0] : '0';
+                sign[1] = '\0';
+                hh[0] = isdigit(zone[1]) ? zone[1] : '0';
+                hh[1] = isdigit(zone[2]) ? zone[2] : '0';
+                hh[2] = '\0';
+                mm[0] = isdigit(zone[3]) ? zone[3] : '0';
+                mm[1] = isdigit(zone[4]) ? zone[4] : '0';
+                mm[2] = '\0';
+
+                netdata_configured_utc_offset = (atoi(hh) * 60 * 60) + (atoi(mm) * 60);
```
thiagoftsm, line 722: By steps, `atoi` is an API for `strtol`, according documentation, but I suggest do not use any of these functions in `netdata`. Please, use [str2i](https://github.com/netdata/netdata/blob/2cee1476f21d2472c4339e4a6e180b20856e469d/libnetdata/inlined.h#L58).

thiagoftsm: Another thing, please, multiply by `3600` instead `60 * 60`

database/rrd.h
```diff
@@ -764,6 +764,8 @@ struct rrdhost {
     const char *os;                                 // the O/S type of the host
     const char *tags;                               // tags for this host
     const char *timezone;                           // the timezone of the host
+    const char *abbrev_timezone;                    // the abbriviated timezone of the host
+    int32_t utc_offset;                             // the offset in seconds from utc
```
thiagoftsm, line 768: `rrdhost` is the main structure of our data model, did you test this PR  on a machine that had an older netdata installed? I have the feeling that an update will break compatibility. :thinking:

streaming/sender.c
```diff
@@ -214,7 +214,7 @@ static int rrdpush_sender_thread_connect_to_parent(RRDHOST *host, int default_po
 
     char http[HTTP_HEADER_SIZE + 1];
     int eol = snprintfz(http, HTTP_HEADER_SIZE,
-            "STREAM key=%s&hostname=%s&registry_hostname=%s&machine_guid=%s&update_every=%d&os=%s&timezone=%s&tags=%s&ver=%u"
+            "STREAM key=%s&hostname=%s&registry_hostname=%s&machine_guid=%s&update_every=%d&os=%s&timezone=%s&abbrev_timezone=%s&utc_offset=%d&tags=%s&ver=%u"
```
thiagoftsm, line 217: Our stream protocol has `versions`, considering that you are modifying the protocol, this means we need to bring a new version for it.

daemon/analytics.c
MrZammler: Hi Thiago, thanks, changed with https://github.com/netdata/netdata/pull/11051/commits/bf473262fcebc9061b500249f848160273323383

database/rrd.h
MrZammler: Hi Thiago.

Yes, I did some tests (streaming) with old (i.e. without this PR) agents. 

If an old agent is connected as child to a new parent  (with this PR), then for the child `rrdhost` we will have `UTC` as an abbreviated timezone, and `0` as `utc_offset`. (i.e. the child does not send these new attributes).

If a new agent is connected as a child to an old parent, then the new attributes that the child sends will be ignored by the parent.

I've documented this behaviour in the Test Plan. Is that what you're asking? Perhaps you mean something else?

thiagoftsm: Hello @MrZammler ,

My concern in this PR and PR #11059 (ping @stelfrag ) is when users start using netdata after these PRs are merged, we load host from our database and instead to get the stored values in `RRDHOST_FLAGS flags;` and ` RRDHOST_FLAGS *exporting_flags;`, they will have values stored in these new variables. 

I am already months without to see what is happening in our dbengine, so it is possible I am missing something and my concern is not necessary.

MrZammler: Hi Thiago, yes I see what you mean. Yes, the new `utc_offset` and `abbrev_timezone` are not stored in the sqlite. At this point I think it's not needed (but @stelfrag will of course decide).

There is a change in `database/sqlite/sqlite_functions.c` (line 987), where when a host is read from the sqlite, in `set_host_properties` it will pass `0` for `utc_offset` and `NULL` for `abbrev_timezone` (which will be `UTC` eventually).

I think it is safe for now, and if at a later date we decide to also store these attributes in sqlite, it can be done.

stelfrag: The system info is not currently stored in the database so it is not affected

streaming/sender.c
stelfrag: This doesn't seem to affect us since this is in the header (the version checks happen after this is parsed)

thiagoftsm (approved):
After some tests I observed that stream was executing without any problems, I could send data  to cloud, and I also did not have any report from valgrind. LGTM!

stelfrag (approved):
Looks Ok!

MrZammler:
Thanks guys!

PR #11070, opened 2021-04-30
Adds New Cloud Arch Connectivity Msgs

underhood:
##### Summary
This might temporarily not work due to latest commit. Cloud needs to update to match. This will be done any moment now.

Adds New connectivity messages for new cloud Arch.

Currently in Dev only mode

##### Component Name
ACLK-NG
##### Test Plan
use staging not production
compile with `-DACLK_NEWARCH_DEVMODE`
can compile with `#define ACLK_LOG_CONVERSATION_DIR "/home/[SOMETHING]/MSG_CAPTURE"` to get all messages sent and received recorded
You can view the recording using this internal tool https://github.com/netdata/aclk_log_viewer


aclk/aclk.c
```diff
@@ -240,6 +241,44 @@ static void msg_callback(const char *topic, const void *msg, size_t msglen, int
...
+    if (unlikely(!*msgtype)) {
+        error_report("Message type empty. Ignoring message from topic \"%s\"", topic);
+        return;
+    }
+
+#ifdef ACLK_LOG_CONVERSATION_DIR
+#define FN_MAX_LEN 512
+    char filename[FN_MAX_LEN];
+    int logfd;
+    snprintf(filename, FN_MAX_LEN, ACLK_LOG_CONVERSATION_DIR "/%010d-rx-%s.bin", ACLK_GET_CONV_LOG_NEXT(), msgtype);
```
thiagoftsm, line 271: @underhood, please, either subtract one byte from `FN_MAX_LEN` or add one byte to filename declaration, for we keep compatibility with the rest of the code and keep this call safe.

aclk/aclk.c
```diff
@@ -240,6 +241,44 @@ static void msg_callback(const char *topic, const void *msg, size_t msglen, int
...
+    logfd = open(filename, O_CREAT | O_TRUNC | O_WRONLY, S_IRUSR | S_IWUSR );
+    if(logfd < 0)
+        error("Error opening ACLK Conversation logfile \"%s\" for RX message.", filename);
+    write(logfd, msg, msglen);
```
thiagoftsm, line 275: The `write` syscall will return the number of bytes written, considering that we are reporting error when when we try to open, my suggestion is also to verify possible errors for `write`.

aclk/aclk_tx_msgs.c
```diff
@@ -32,6 +32,36 @@ static void aclk_send_message_subtopic(mqtt_wss_client client, json_object *msg,
...
+    }
+
+    mqtt_wss_publish_pid(client, topic, msg, msg_len,  MQTT_WSS_PUB_QOS1, &packet_id);
+#ifdef NETDATA_INTERNAL_CHECKS
+    aclk_stats_msg_published(packet_id);
+#endif
+#ifdef ACLK_LOG_CONVERSATION_DIR
+#define FN_MAX_LEN 1024
+    char filename[FN_MAX_LEN];
+    snprintf(filename, FN_MAX_LEN, ACLK_LOG_CONVERSATION_DIR "/%010d-tx-%s.bin", ACLK_GET_CONV_LOG_NEXT(), msgname);
```
thiagoftsm, line 55: I will do the same request here, please, either subtract a byte from `FN_MAX_LEN`, or add one more byte to `filename` declaration.

aclk/aclk_tx_msgs.c
```diff
@@ -32,6 +32,36 @@ static void aclk_send_message_subtopic(mqtt_wss_client client, json_object *msg,
...
+    FILE *fptr;
+    fptr = fopen(filename,"w");
```
thiagoftsm, line 57: Please, add a test to check that you did not received `NULL` as result. Another thing, considering that you are using here `fopen(3)`, `fwrite(3)`, and `fclose(3)`, instead to call the syscalls `open(2)`, `write(2)`, and `close(2)`. I will give a suggestion for you to use the same functions in other file, and of course, this means that we can create an unique function for both.

aclk/schema-wrappers/node_creation.cc
```diff
@@ -0,0 +1,37 @@
...
+    nodeinstance::create::v1::CreateNodeInstance msg;
+
+    if (data->claim_id)
+        msg.set_claim_id(data->claim_id);
+    msg.set_machine_guid(data->machine_guid);
+    msg.set_hostname(data->hostname);
+    msg.set_hops(data->hops);
+
+    *len = msg.ByteSizeLong();
+    char *bin = (char*)malloc(*len);
```
thiagoftsm, line 19: Considering that here we are having a situation that can create an exception, why are we not writing it inside an `try ... catch` block?

thiagoftsm:
I finished the tests of your PR and I could see improvement compared with the previous ACLK.
When you address my review, please, also take a look at [LGTM warnings ](https://github.com/netdata/netdata/pull/11070#issuecomment-829920607). Thank you!

aclk/aclk.c
underhood: OK, please note all under `#ifdef ACLK_LOG_CONVERSATION_DIR` is meant to be only for developer use during development/debugging so not that much attention is payd for that

underhood:
@thiagoftsm every comment about code in `#ifdef ACLK_LOG_CONVERSATION_DIR` is valid it is developer only debug feature which is quick and dirty. I will improve that.

LGTM warnings were cleared already in recent commits

I will address your other comments ASAP. Thanks for the review

thiagoftsm (approved):
Considering the last comment I was addressed, I am approving this PR, but I will not mark my comments as resolved, for we remember it will be necessary to do small adjusts in this code.

thiagoftsm (dismissed):
I retest after the last commits and everything worked when I worked with dbengine, but considering that the problem with RAM mode does not have relationship with this PR, I am approving it.

thiagoftsm:
@stelfrag and @underhood when I tesed running `memory mode = RAM`, I had problems to send data for the cloud and I observed these message:

```
2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 874, expected slot: 439
2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 875, expected slot: 440
2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 876, expected slot: 441
2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 877, expected slot: 442
2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 878, expected slot: 443

```

This is not the first time that the problem happened, I could send data during two minutes and after this the cloud did not collect more data from my host. After to move for `dbengine`, I did not have the same problem.

stelfrag:
> @stelfrag and @underhood when I tesed running `memory mode = RAM`, I had problems to send data for the cloud and I observed these message:
> 
> ```
> 2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 874, expected slot: 439
> 2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 875, expected slot: 440
> 2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 876, expected slot: 441
> 2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 877, expected slot: 442
> 2021-05-07 12:03:53: netdata ERROR : ACLK_Query_0 : INTERNAL CHECK: Unaligned query for iowait, database slot: 878, expected slot: 443
> ```
> 
> This is not the first time that the problem happened, I could send data during two minutes and after this the cloud did not collect more data from my host. After to move for `dbengine`, I did not have the same problem.

This is unrelated to this PR and it has to do with the way the query works in RAM mode (as it wraps around the allocated memory page)

thiagoftsm:
> This is unrelated to this PR and it has to do with the way the query works in RAM mode (as it wraps around the allocated memory page)

I agree, but it is something that probably we need to take a look, or to bring an additional information in our documentation that cloud works when Netdata runs using `dbengine`, but some problems are expected with other memory mode.

aclk/schema-wrappers/node_creation.cc
underhood: malloc doesn't throw exceptions

aclk/aclk_tx_msgs.c
underhood: as pre previous comments please ignore all under #ifdef ACLK_LOG_CONVERSATION_DIR for now
it is meant to be only for developer use during development/debugging and will be reworked/improved in future PRs

underhood:
@stelfrag can we get this in to not create conflict with the switch ACLK-NG to default PR?

stelfrag:
> @stelfrag can we get this in to not create conflict with the switch ACLK-NG to default PR?

Sounds good -- lets review !

underhood:
note as @thiagoftsm correctly pointed out in his review stuff in `#ifdef ACLK_LOG_CONVERSATION_DIR` needs some work, but this is developer only debugging feature. This will be reworked/improved in future commits

aclk/aclk.c
thiagoftsm, line 271: Please, add one character for `filename` or subtract one byte here inside the argument. I prefer the first option.

aclk/aclk.c
thiagoftsm, line 275: The current algorithm will try to write to `logfd` when it returns error, please, add an `else` before `write` to be called.

aclk/aclk.c
underhood: `The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.`

```
int snprintf ( char * s, size_t n, const char * format, ... );
Write formatted output to sized buffer
Composes a string with the same text that would be printed if format was used on printf, but instead of being printed, the content is stored as a C string in the buffer pointed by s (taking n as the maximum buffer capacity to fill).

If the resulting string would be longer than n-1 characters, the remaining characters are discarded and not stored, but counted for the value returned by the function.

A terminating null character is automatically appended after the content written.

After the format parameter, the function expects at least as many additional arguments as needed for format.
```

if you do program like following:
```C
        char buf[5];
        snprintf(buf, 5, "0123456789");
```
and open it in GDB  you can see result is properly terminated string "0123" not exceeding 5 bytes

thiagoftsm: I called your attention for this, because we are following this pattern on other files like [this](https://github.com/netdata/netdata/blob/37bee1d197229a4de1d6eac9942fc626f1da8ee3/collectors/proc.plugin/plugin_proc.c#L248-L250).

The unique real problem that we will have with this, it is to have two different patterns inside our code.

aclk/aclk.c
```diff
@@ -895,3 +972,92 @@ struct label *add_aclk_host_labels(struct label *label) {
...
+        rrdhost_aclk_state_unlock(localhost);
+        create_query->data.node_creation.hops = 1; //TODO
+        create_query->data.node_creation.hostname = strdupz(host->hostname);
+        create_query->data.node_creation.machine_guid = strdupz(host->machine_guid);
+        aclk_queue_query(create_query);
+        return;
+    }
+
+    aclk_query_t query = aclk_query_new(NODE_STATE_UPDATE);
+    query->data.node_update.hops = 1; //TODO
```
thiagoftsm, line 1004: Please, can you add more information for what is necessary `TODO`.

aclk/schema-wrappers/node_connection.cc
```diff
@@ -0,0 +1,35 @@
...
+    timestamp->set_seconds(tv.tv_sec);
+    timestamp->set_nanos(tv.tv_usec * 1000);
+
+    *len = msg.ByteSizeLong();
+    char *bin = (char*)malloc(*len);
+    if (bin)
+        msg.SerializeToArray(bin, *len);
+
+    return bin;
+}
```
thiagoftsm, line 35: Please, add a new line at the end of this file.

underhood: added thx

aclk/aclk.c
underhood: added clarification - this will be changed when streaming hops will be implemented (e.g. when we will have information from streaming how far the children is from parent)

aclk/aclk.c
underhood: I prefer correctness over patterns. Adding byte here will just add useless byte of memory and confusion for whoever reads the code.

thiagoftsm: You are right, this is not a source of bug. 

I will consider that it is not possible to reach 512 bytes and I will start the tests of this PR now.

aclk/aclk.c
thiagoftsm: @underhood I tested your PR and everything looks ok, please, take a look on this call for `write', and after this I will approve your PR.

underhood:
closing in favor of ACLK-NG as default as we changed the plan. This will be reopened in some form afterwards

PR #10870, opened 2021-03-27
Improvements to dash-example.html

tnyeanderson:
##### Summary

* Fixes a race condition that caused graphs to not load
* Moves configuration to the top of the file
* Improves docs
* Other small fixes

**NOTE: This file is not maintained by Netdata, it is just meant to be an example of multi-host streaming capabilities**

##### Component Name

`dash-example.html`

##### Test Plan

**NOTE: This file is not maintained by Netdata, it is just meant to be an example of multi-host streaming capabilities**

PR #11077, opened 2021-04-30
Remove dash-example, place in community repo

tnyeanderson:
See #10870 for reasoning.

PR for community: netdata/community#31


joelhans (dismissed):
Not sure what's going on with the Actions, but in principle this LGTM. Pretty simple change.

Ferroin:
> Not sure what's going on with the Actions

They’re failing because removing this file without also removing it from `web/gui/Makefile.am` breaks the build.

@tnyeanderson The fix for the failing PR checks is to remove the reference to `dash-example.html` from the `dist_web_DATA` variable in `web/gui/Makefile.am` at line 70. Just deleting the whole line should do it.

tnyeanderson:
Thank you for the tip @Ferroin ... Done!

Ferroin (approved):

joelhans (approved):

OdysLam:
Hey @jacekkolasa, since we are removing the community-driven dashboards. I think we should also transfer to the community repo the following files:
```
demo.html
demo2.html
demosites.html
demosites2.html
tv.html
```

As far as I can tell they serve illustrative purposes and most of them haven't been updated for a very long period of time.

jacekkolasa:
@OdysLam We can remove them, they are anyway replaced by the files from https://github.com/netdata/dashboard/tree/master/public during installation. @Ferroin is making changes to the way Dashboard is build, so that will change, too

jacekkolasa (approved):

OdysLam:
Okay, merging this for now, but my comment stands even for the dashboard repo as well. I think we should remove the same files from the dashboard repo and add them to the community repo. 

What do you say about this? cc @novykh

tnyeanderson:
@OdysLam does this mean I'm a [Netdata Core member](https://github.com/netdata/netdata/pull/10870#issuecomment-829130283)? :)

PR #11078, opened 2021-05-02
Add an event when an incomplete agent shutdown is detected

stelfrag:
##### Summary
A small improvement on the previous agent crash detection

If the shutdown function is called, rename the generated file from `.agent_crash` to `.agent_incomplete_shutdown`. 
If the shutdown is interrupted, on restart the `agent_incomplete_shutdown` will be detected and an INCOMPLETE_SHUTDOWN event will be generated instead of a CRASH event

Note: this only applies when agent anonymous statistics is enabled

##### Component Name
agent

##### Test Plan
- The CRASH event is generated when the agent is killed or terminated abnormally
- Try to kill the agent if it takes a long time to shutdown
  - The file `var/lib/netdata/.agent_incomplete_shutdown` will be created and detected on the next agent restart

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -31,8 +31,10 @@ void netdata_cleanup_and_exit(int ret) {
     analytics_free_data();
 
     char agent_crash_file[FILENAME_MAX + 1];
+    char agent_incomplete_shutdown_file[FILENAME_MAX + 1];
     snprintfz(agent_crash_file, FILENAME_MAX, "%s/.agent_crash", netdata_configured_varlib_dir);
-    (void) unlink(agent_crash_file);
+    snprintfz(agent_incomplete_shutdown_file, FILENAME_MAX, "%s/.agent_incomplete_shutdown", netdata_configured_varlib_dir);
+    (void) rename(agent_crash_file, agent_incomplete_shutdown_file);
 
     // cleanup/save the database and exit
     info("EXIT: cleaning up the database...");
@@ -68,6 +70,7 @@ void netdata_cleanup_and_exit(int ret) {
     security_clean_openssl();
 #endif
     info("EXIT: all done - netdata is now exiting - bye bye...");
+    (void) unlink(agent_incomplete_shutdown_file);
     exit(ret);
 }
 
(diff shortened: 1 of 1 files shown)


daemon/main.c
```diff
@@ -31,8 +31,10 @@ void netdata_cleanup_and_exit(int ret) {
     analytics_free_data();
 
     char agent_crash_file[FILENAME_MAX + 1];
+    char agent_incomplete_shutdown_file[FILENAME_MAX + 1];
     snprintfz(agent_crash_file, FILENAME_MAX, "%s/.agent_crash", netdata_configured_varlib_dir);
-    (void) unlink(agent_crash_file);
+    snprintfz(agent_incomplete_shutdown_file, FILENAME_MAX, "%s/.agent_incomplete_shutdown", netdata_configured_varlib_dir);
```
thiagoftsm, line 36: I am not sure if the name is proper for what is happening :thinking:  . I sent a `SIGTERM` to finish the agent, and everything described on this PR happened, but a `SIGTERM  is not something expected to generate an incomplete shutdown. My suggestion is we rename the file. If I misunderstood the main goal of the PR, please, can you explain why a normal end means an "incomplete shutdown"? Thanks.

Ferroin: My understanding is that the goal here is to detect cases where the agent was asked to shut down gracefully, but is then not given sufficient time to do so. ‘incomplete shutdown’ sounds like a correct description to me provided that that situation is what we are actually checking for.

thiagoftsm: I got your point @Ferroin , but considering the fact that the agent was running,  my understanding is the fact that an incomplete shutdown happens only when the shutdown process starts, but somehow it cannot go until the end. This was the motive I required to rename the file.

stelfrag: The `incomplete` shutdown will happen if a SIGTERM is received and the agent fails (crashes or externally killed before it gracefully exits). In the happy path that it exits gracefully it will remove the file so it will not be detected upon restart.

We need to further improve this in subsequent PRs to measure the time it takes to shutdown etc

MrZammler:
Tested, and seems to work as expected. It also generated an INCOMPLETE_SHUTDOWN and sent to posthog. It's okay, as long as we keep in mind in posthog that these events (and the CRASH one) refer to a previous agent run. Perhaps also these events can have an 'ERROR' in `action_result` to indicate that these events signify a problem (so Andy perhaps can group/query them).

daemon/main.c
MrZammler: I think what Thiago is saying, is that you can keep the logic the same, but change the filename to perhaps `.agent_shutting_down`...

MrZammler (approved):

daemon/main.c
thiagoftsm: @MrZammler is right, I do not think the name is the best option to describe what is happening right now.
Considering that we will have another PRs to improve this, I will approve this PR for we do not postpone more the feature. Please, think about the possibility to rename the file.

thiagoftsm (approved):

PR #11083, opened 2021-05-03
build mqtt_websockets with netdata autotools

### LINKED ISSUE #10971: Netdata 1.30.1 - mqtt_websockets compilation issue (gcc dep hardcoded and openssl pc requirement)
The following can be found in ```mqtt_websockets/Makefile``` 
```
CC = gcc -std=gnu99
CFLAGS = -Wextra -Wall `pkg-config --cflags openssl`
```
This breaks compilation on any platform that doesn't ship gcc (FreeBSD for example) or doesn't call it gcc.
Regarding openssl, why doesn't it use parameters set by configure script?

underhood:
##### Summary
Don't use handwritten makefiles from git submodules. Build it by netdata build system for netdata.
Fixes #10971

##### Component Name
autotools build
##### Test Plan
CI is pretty much the best test. Build with Legacy / NG

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -117,11 +117,7 @@ SUBDIRS += \
     spawn \
     $(NULL)
 
-if ACLK_NG
-SUBDIRS += \
-    mqtt_websockets \
-    $(NULL)
-else
+if !ACLK_NG
 SUBDIRS += \
     aclk/legacy \
     $(NULL)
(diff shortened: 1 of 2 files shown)


vkalintiris (approved):
@underhood Do you have a rough estimate on when we will be able to delete the old ACLK stuff?

underhood:
@vkalintiris I hope ASAP. I would like that to be default by next release at minimum. We can talk with @stelfrag about this and make a more firm plan

stelfrag (approved):

PR #11084, opened 2021-05-03
Automatically trigger Helmchart PR on agent release.

Ferroin:
##### Summary

This will automatically trigger the workflow in the netdata/helmchart repo that generates a PR to bump the agent version bundled with the helmchart when we bpulish a docker image for a release.

##### Component Name

area/ci

##### Test Plan

Needs to be actually merged to be able to properly verify that it correctly triggers the workflow in netdata/helmchart, but there is no risk of it breaking the existing build infrastructure.

##### Additional Information

Actually having this integration was the roiginal intent of automating the process of generating the PRs in the netdata/helmchart repo, though it got kind of forgotten about until just recently.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -75,4 +75,4 @@ jobs:
           repo: netdata/helmchart
           workflow: Agent Version PR
           ref: refs/heads/master
-          inputs: {"agent_version": "${{ github.event.inputs.version }}"}
+          inputs: '{"agent_version": "${{ github.event.inputs.version }}"}'
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -67,3 +67,11 @@ jobs:
             && github.event_name != 'pull_request'
             && startsWith(github.ref, 'refs/heads/master')
           }}
+      - name: Trigger Helmchart PR
+        if: github.event_name == 'workflow_dispatch' && github.event.inputs.version != 'nightly'
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -67,3 +67,11 @@ jobs:
...
             && startsWith(github.ref, 'refs/heads/master')
           }}
+      - name: Trigger Helmchart PR
+        if: github.event_name == 'workflow_dispatch' && github.event.inputs.version != 'nightly'
+        with:
+          token: ${{ secrets.NETDATABOT_GITHUB_TOKEN }}
+          repo: netdata/helmchart
+          workflow: Agent Version PR
+          ref: refs/heads/master
+          inputs: {"agent_version": "${{ github.event.inputs.version }}"}
```
ilyam8, line 77: Can you give a link to the docs?

I wonder how it works - I assume this step suppose to do an HTTP request using GitHub API, but there is no `uses`. And is it only `repo`? or should be `owner` and `repo`?

Ferroin: Oh, nice catch, I forgot to add the `uses` line...

I’ll have a fix pushed shortly for that.

Ferroin: Updated now to properly include the `uses` key.

For reference, the action being used here is: https://github.com/benc-uk/workflow-dispatch

ilyam8: Thanks!

ilyam8 (approved):

knatsakis (approved):

PR #11100, opened 2021-05-06
Add host_cloud_enabled attribute to analytics

MrZammler:
##### Summary

Adds a new attribute `host_cloud_enabled` to posthog analytics. The attribute will be false if `--disable-cloud` was used during install, or `enabled = no` is defined in `/var/lib/netdata/cloud.d/cloud.conf`

##### Component Name

Analytics

##### Test Plan

To test, enable debug with analytics debug flag `0x0000000080000000`, or observe the host's page on posthog.
If `--disable-cloud` is used, the attribute should be `false`.
If `enabled = no` is specified in `/var/lib/netdata/cloud.d/cloud.conf`, the attribute should also be `false`.
Otherwise, it should be `true`

### DIFF
--- a/daemon/analytics.c
+++ b/daemon/analytics.c
@@ -569,14 +569,14 @@ void set_late_global_environment()
 {
     analytics_set_data(&analytics_data.netdata_config_stream_enabled, default_rrdpush_enabled ? "true" : "false");
     analytics_set_data_str(&analytics_data.netdata_config_memory_mode, (char *)rrd_memory_mode_name(default_rrd_memory_mode));
-    analytics_set_data(&analytics_data.netdata_config_exporting_enabled, appconfig_get_boolean(&exporting_config, CONFIG_SECTION_EXPORTING, "enabled", 1) ? "true" : "false");
+    analytics_set_data(&analytics_data.netdata_config_exporting_enabled, appconfig_get_boolean(&exporting_config, CONFIG_SECTION_EXPORTING, "enabled", CONFIG_BOOLEAN_NO) ? "true" : "false");
 
 #ifdef DISABLE_CLOUD
     analytics_set_data(&analytics_data.netdata_host_cloud_enabled, "false");
 #else
     analytics_set_data(
         &analytics_data.netdata_host_cloud_enabled,
-        appconfig_get_boolean(&cloud_config, CONFIG_SECTION_GLOBAL, "enabled", 1) ? "true" : "false");
+        appconfig_get_boolean(&cloud_config, CONFIG_SECTION_GLOBAL, "enabled", CONFIG_BOOLEAN_YES) ? "true" : "false");
 #endif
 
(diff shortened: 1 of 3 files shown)


daemon/analytics.c
```diff
@@ -569,6 +571,14 @@ void set_late_global_environment()
     analytics_set_data_str(&analytics_data.netdata_config_memory_mode, (char *)rrd_memory_mode_name(default_rrd_memory_mode));
     analytics_set_data(&analytics_data.netdata_config_exporting_enabled, appconfig_get_boolean(&exporting_config, CONFIG_SECTION_EXPORTING, "enabled", 1) ? "true" : "false");
 
+#ifdef DISABLE_CLOUD
+    analytics_set_data(&analytics_data.netdata_host_cloud_enabled, "false");
+#else
+    analytics_set_data(
+        &analytics_data.netdata_host_cloud_enabled,
+        appconfig_get_boolean(&cloud_config, CONFIG_SECTION_GLOBAL, "enabled", 1) ? "true" : "false");
```
ilyam8, line 579: I think you need to use 

https://github.com/netdata/netdata/blob/1b819afdbba067ae6abbeac20fb8805cd1e4e3c4/libnetdata/config/appconfig.h#L152-L153

instead of 1.

I see it is the pattern, for instance

https://github.com/netdata/netdata/blob/ac3f9b171754014126edf6235d7d751f9aaa4ea5/collectors/ebpf.plugin/ebpf.c#L830-L835

MrZammler: Correct, I will, thanks (I will use `CONFIG_BOOLEAN_YES`).

Btw, I also use it here, and I think it's wrong:

`analytics_set_data(&analytics_data.netdata_config_exporting_enabled, appconfig_get_boolean(&exporting_config, CONFIG_SECTION_EXPORTING, "enabled", 1) ? "true" : "false");`

I'm not sure if the default value of exporting is enabled or not (I think it's not, so if it is not found in the config, it should be disabled?) I.e. the above:

`analytics_set_data(&analytics_data.netdata_config_exporting_enabled, appconfig_get_boolean(&exporting_config, CONFIG_SECTION_EXPORTING, "enabled", CONFIG_BOOLEAN_NO) ? "true" : "false");`

ilyam8: @vlvkobal please halp

vlvkobal: Yes, it is disabled by default.

MrZammler: Ok, thanks both guys, pushed: https://github.com/netdata/netdata/pull/11100/commits/a078fe63a226ec847cd95016541b09132f80918f

ilyam8 (approved):

thiagoftsm (approved):

vlvkobal (approved):

MrZammler:
Thanks guys!

PR #11108, opened 2021-05-06
Create installation reference doc

joelhans:
##### Summary

I'm working to improve our installation story by consolidating and moving docs into the `/docs/installation` folder and having a more central reference doc for how `netdata-installer.sh` works. I decided to start with that and moving the reinstallation doc.

A major improvement here is to remove the `/packaging/installer/README.md` doc in favor of `/docs/get-started.mdx`, which is the much newer source of truth.

##### Component Name

area/packaging


docs/installation/reference-installer.mdx
```diff
@@ -111,21 +75,48 @@ operating systems.
     svg="linux" />
 </Install>
 
+## `netdata-installer.sh`
+
+The following installation methods are a wrapper around a single script,
```
OdysLam, line 80:
```suggestion
The following installation methods use the same single script,
```

docs/installation/reference-installer.mdx
```diff
@@ -111,21 +75,48 @@ operating systems.
...
+| :-------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `--dont-wait`         | Skip prompting for permission to install any required packages. Useful for automated installation, such as with [Ansible](/docs/guides/ansible.md).                                 |
+| `--dont-start-it`     | Prevent the installer from starting Netdata automatically.                                                                                                                          |
+| `--stable-channel`    | Automatically update only on the release of new major versions. See [nightly vs. stable releases](#nightly-vs-stable-releases).                                                                                                                    |
+| `--nightly-channel`   | Automatically update on every new nightly build. See [nightly vs. stable releases](#nightly-vs-stable-releases).                                                                                                                                   |
```
OdysLam, line 103: I think we should mention at this point of the guide that this is the default behaviour of the Agent. So in reality it's about either using `--stable-channel` flag or no flag at all.

docs/installation/reference-installer.mdx
```diff
@@ -1,76 +1,40 @@
...
+- Alpine `edge`, `3.13`, `3.12`, `3.11`, `3.10`
+- Arch Linux `latest`
+- Fedora `34`, `33`, `32`
+- CentOS `8`, `7`
+- OpenSUSE `tumbleweed:latest`, `leap:15.2`
+- Clear Linux `latest`
 
-The usage statistics are _vital_ for us, as we use them to discover bugs and prioritize new features. We thank you for
-_actively_ contributing to Netdata's future.
+Netdata runs on other distributions and operating systems as well, but they aren't officially supported by our team.
```
OdysLam, line 33: This is a bold claim. We need to update the installation platform modals with a disclaimer (e.g macos).

OdysLam (approved):

docs/installation/reference-installer.mdx
```diff
@@ -111,21 +75,48 @@ operating systems.
...
 </Install>
 
+## `netdata-installer.sh`
+
+The following installation methods are a wrapper around a single script,
+[`netdata-installer.sh`](https://raw.githubusercontent.com/netdata/netdata/master/netdata-installer.sh), which handles
+the installation process itself:
+
+- [One-line installer (`kickstart.sh`)](/packaging/installer/methods/kickstart.md)
+- [Pre-compiled static binary (`kickstart-static64.sh`)](/packaging/installer/methods/kickstart64.md)
```
Ferroin, line 85: Strictly speaking, the static installer is not a wrapper around the `netdata-installer.sh` script. That script is used to _create_ the static install bundle, but none of that code actually gets run on the user system when they use the static installer (the same applies to the Docker images as well actually).

docs/installation/reference-installer.mdx
Ferroin: We’ve been discussing our formal support policy some recently. There’s a meeting about this scheduled for next Monday. Assuming we end up with a concrete and well defined support policy there, the plan is t then update `packaging/DISTRIBUTIONS.md` accordingly and make an effort to keep it up to date, in which case this whole section should probably just be a link to that page.

docs/installation/reference-installer.mdx
joelhans: Thanks for that clarification! Some people might like using the static binary in that case.

docs/installation/reference-installer.mdx
joelhans: @OdysLam I guess I'm not sure how this is a bold claim. Is it not true that Netdata runs on FreeBSD, Alpine, etc etc?

But @Ferroin, I'm looking forward to seeing what you come up with about the support policy and the `DISTRIBUTIONS.md` doc. I'll very happily streamline the structure of the docs around installation when there's more clarity there.

docs/installation/reference-installer.mdx
joelhans: Maybe an enhancement for the next "pass" on this front would be to use the individual method pages to explain the default. For example, the `kickstart.sh` page would explain that by running `bash <(curl -Ss https://my-netdata.io/kickstart.sh)`, you get:

- Automatic updates
- Nightly updates
- Anonymous statistics on

Then the page can reference back to this table to help users who want to override those.

joelhans:
@OdysLam @Ferroin Friendly ping on this one, fellas.

Ferroin:
It looks like the file moves have broken the build, though I’m not 100% certain why (I think we try to include those files in the dist tarball we generate for releases?).

joelhans:
@Ferroin Yep, you're right about the file being included somewhere. I added it back it empty and the builds work again. What do you think—move forward with this "dummy" file or figure out how to remove the "dependency"? I'm afraid I don't know enough about the build process to know where to even start.

OdysLam (dismissed):

Ferroin:
> @Ferroin Yep, you're right about the file being included somewhere. I added it back it empty and the builds work again. What do you think—move forward with this "dummy" file or figure out how to remove the "dependency"? I'm afraid I don't know enough about the build process to know where to even start.

Fixing the build system so we don’t need the dummy file would be preferable. The fix _should_ just consist of removing `packaging/installer/README.md` from the `dist_noinst_DATA` variable in `Makefile.am` (should just be deleting the whole line).

joelhans:
@Ferroin Thanks—I think I finally got it figured out. For some reason `Makefile.ac` isn't showing up in my VS Code, which meant it took me forever to realize it was always in the filesystem. Weird. Must have hidden it or something.

kickoke:
@Ferroin 
In terms of housekeeping: This PR has been dormant for quite a while. Given that we reworked the installation sections recently, do I have your permission to close this PR?

Ferroin:
> @Ferroin In terms of housekeeping: This PR has been dormant for quite a while. Given that we reworked the installation sections recently, do I have your permission to close this PR?

Fine by me.

PR #11123, opened 2021-05-10
Build netdata without long double by default.

vkalintiris:
##### Summary

This will (1) improve performance without any loss of precission, and (2) allow CGO 
code to access all struct fields without having to write pass-through wrapper functions.
We use `AM_CPPFLAGS` to define `NETDATA_WITHOUT_LONG_DOUBLE`, in order
to allow overriding the default if necessary.

##### Test Plan

Built netdata under many different configuration options, verified that we get no new warnings/errors, and checked the dashboard's charts to make sure we don't get any weird numbers.


thiagoftsm (approved):
I tested the PR using `Slackware current` and `CentOS 8`. I ran Netdata with memory mode `dbengine` and `ram`. On both systems everything worked as expected.

I also used the following alarm to confirm I was receiving notifications:

``` 
 alarm: dev_dim_template
    on: system.cpu
    os: linux
 units: %
 every: 1s
  warn: $this > 1
  crit: $this > 4
```

And I did not observe problems with health system.

vlvkobal (approved):

stelfrag:
Note: This is a breaking change -- if you use memory mode `save`, you will lose all metrics

vkalintiris:
> Note: This is a breaking change -- if you use memory mode `save`, you will lose all metrics

@stelfrag Is it okay to merge, or should we wait to do so right before the next release? FYI, either way, I'll do a final manual verification on a rebased branch before merging this.

PR #11124, opened 2021-05-10
Remove unecessary relative paths when including headers.

vkalintiris:
Currently, we add the repository's top-level dir in the compiler's
header search path. This means that code in every top-level directory
within the repo can include headers sibling top-level directories.

This patch makes header inclusion consistent when it comes to files
that are included from sibling top-level directories within the repo.


thiagoftsm (approved):
I compiled this PR using `gcc 10.3.0` on Slackware current, and after this I compiled the PR with `clang 11.0.1` on FreeBSD 13.0` and I did not observe any warnings.

vlvkobal (approved):

Issue #7150, opened 2019-10-22
Converting the build-system to CMake


amoss:
## Summary
Wrapper for the separate issues in converting the build-system.

vlvkobal:
@stelfrag @vkalintiris we still support both - Autotools and CMake build systems, which is not wise. Can we plan further steps on the Epic? I believe that getting rid of Autotools will be of great benefit for us.

vlvkobal:
BTW, we broke our CMake build configuration with the last commits. At least, ACLK is misconfigured.

ilyam8:
@amoss ❤️

PR #11129, opened 2021-05-10
Compile/Link with absolute paths for bundled/vendored deps.

vkalintiris:
##### Summary

The modified options are set only by the netdata-installer.sh script. The paths always point to the correspondent `externaldeps/xxx` directory when the options are specified.

##### Test Plan

CI builds

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -174,12 +174,8 @@ AC_ARG_ENABLE(
 )
 AC_ARG_WITH(
     [bundled-lws],
-    [AS_HELP_STRING([--with-bundled-lws=DIR], [Use a specific Libwebsockets static library @<:@default use system library@:>@])],
-    [
-        with_bundled_lws="yes"
-        bundled_lws_dir="${withval}"
-    ],
-    [with_bundled_lws="no"]
+    [AS_HELP_STRING([--with-bundled-lws], [Use the bundled version of libwebsockets library @<:@default use system library@:>@])],
+    [with_bundled_lws="yes"], [with_bundled_lws="no"]
 )
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


underhood:
To provide more info about this PR for reviewers: I assume the motive is to allow `VPATH` automake builds and ultimate goal of this is the GO integration (fixing VPATH builds is of course correct thing with or without golang long term goal)?

configure.ac
```diff
@@ -671,9 +666,9 @@ if test "$enable_cloud" != "no" -a "$aclk_ng" != "yes"; then
 
     if test "${with_bundled_lws}" = "yes"; then
         AC_MSG_CHECKING([if libwebsockets static lib is present])
-        if test -f "${bundled_lws_dir}/libwebsockets.a"; then
-            LWS_CFLAGS="-I ${bundled_lws_dir}/include"
-            OPTIONAL_LWS_LIBS="${bundled_lws_dir}/libwebsockets.a"
+        if test -f "externaldeps/libwebsockets/libwebsockets.a"; then
+            LWS_CFLAGS="-I \$(abs_top_srcdir)/externaldeps/libwebsockets/include"
+            OPTIONAL_LWS_LIBS="externaldeps/libwebsockets/libwebsockets.a"
```
underhood, line 671: why `OPTIONAL_LWS_LIBS` has neigher `abs_top_srcdir` or build dir?

vkalintiris:
> To provide more info about this PR for reviewers: I assume the motive is to allow `VPATH` automake builds and ultimate goal of this is the GO integration (fixing VPATH builds is of course correct thing with or without golang long term goal)?

The primary goal is to support [cgo](https://golang.org/cmd/cgo/). A secondary goal is to standardize the way we are building compilation/linking flags, which in turn will make testing other build systems easier (eg. CMake, Bazel and/or Meson).

Fixing `VPATH` (assuming it's already broken) is outside of the scope of the work I plan to do. Although, in theory this change should make this easier in the future for anyone brave enough to give it a try.

configure.ac
vkalintiris: I plan on doing this in a follow-up PR. I wanted to get a thumbs up first, to make sure everyone's fine with the approach I'm taking. Considering that making the internal library deps use an absolute path is a small change, I can update this PR if you think it's easier/better to review at once.

vlvkobal:
If it's easy to integrate `cgo` to our existing build system then, of course, it's wise doing it that way. But if it's comparable with moving to CMake completely, we should get back to #7150.

vlvkobal (approved):

vkalintiris:
> If it's easy to integrate `cgo` to our existing build system then, of course, it's wise doing it that way. But if it's comparable with moving to CMake completely, we should get back to #7150.

Thankfully, it's nowhere near as complicated as moving to a new build system :smile: The remaining bits are mostly about calling the go toolchain with the appropriate `CFLAGS` and `LDFLAGS`.

vlvkobal:
> Thankfully, it's nowhere near as complicated as moving to a new build system 😄

BTW, we should still support our CMake build configuration. Will that build work with `cgo` as well as Autotools build?

vkalintiris:
> > Thankfully, it's nowhere near as complicated as moving to a new build system smile
> 
> BTW, we should still support our CMake build configuration. Will that build work with `cgo` as well as Autotools build?

Why should we support CMake?

vkalintiris:
@Ferroin ok to merge this?

underhood (approved):

PR #11134, opened 2021-05-11
Fix autogeneration of dashboard update PRs.

Ferroin:
##### Summary

* Use the netdatabot PAT instead of the default Github token for creating the commit and PR so that we properly trigger CI checks. Without this we don’t get GHA workflows running on the PR due to the restriction that the default token cannot trigger other worklows.
* Properly follow redirects when downloading the release artifacts. Without this we get the incorrect hash saved, as for some reason Github sends the link access through a redirect before giving you the actual file.

##### Component Name

area/ci

##### Test Plan

cURL fix verified locally.

Token fix will need to be verified by actual usage.

### DIFF
--- a/.github/workflows/dashboard-pr.yml
+++ b/.github/workflows/dashboard-pr.yml
@@ -27,7 +27,7 @@ jobs:
           git push -u origin dashboard-${{ github.event.inputs.dashboard_version }}
       - name: Update Files
         run: |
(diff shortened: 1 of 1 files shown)


jacekkolasa (approved):
looks good, i got the proper file now and sha256sum matches, thanks @Ferroin !

Ferroin:
Superseded by #11139 (one of the two fixes here is made redundant by that PR, and the other is included in that PR)

Issue #11075, opened 2021-04-30
Remove legacy dashboard code and add react dashboard code to repo in it’s place.


Ferroin:
This will consist of:

* Removing all of the legacy dashboard code that is not required either by the new dashboard code or for custom dashboards.
* Adding the (pre-built) code from the current release from the most recent full release of https://github.com/netdata/dashboard.
* Modifying the build system accordingly.
* Removing the bundle-on-install code for the react dashboard from `netdata-installer.sh`
* Removing the bundling code for the react dashboard in our binary packages.
* Updating the workflow created in https://github.com/netdata/netdata/issues/11073 to update the react dashboard code in this repo.

Issue #11072, opened 2021-04-30
Bundle new dashboard code into the agent repository instead of fetching it on install.


Ferroin:
This will provide a couple of concrete benefits:

* It will eliminate a large chunk of complexity from our install and packaging code.
* It will make life easier for anyone building and installing by hand from a git checkout.
* It will mean that third-party packages (such as Homebrew) will finally have the new dashboard code.

The general approach that will be taken here is as follows:

1. Create a new CI workflow here in the agent repository that automates creating a PR to update `packaging/dashboard.version` and `packaging/dashboard.checksums` based simply on a release tag.
2. Update the release workflow in https://github.com/netdata/dashboard to automatically trigger the aforementioned workflow in the agent repo every time a new release is created for the dashboard code. This gets us to near full automation of the _current_ approach used for handling the new dashboard code.
3. Remove the legacy dashboard code from the agent repository, add the (at the time) current version of the new dashboard code, update the build system and installer appropriately, and update the workflow from step 1 to handle automatically updating this code when new dashboard releases occur.

PR #11139, opened 2021-05-12
Bundle the react dashboard code into the agent repo directly.

Ferroin:
##### Summary

This updates our handling of the dashboard code found in https://github.com/netdata/dashboard. Instead of pulling the release in on install, we will now be vendoring the released code in the agent repo itself. This will make the install process simpler and faster for all users, and should allow users who are using third-party packages to use the new dashboard and finally take advantage of the features that depend on it.

This change looks a lot more complicated than it actually is, and is logically split into four parts consisting of one commit each.

The first commit removes the installer code that pulls in the dashboard release, as well as the associated packaging code. This was done as a separate commit to make review easier, because while it0s needed for the other changes to work, it is functionally independent of them.

The second commit adds a `web/gui/dashboard` directory to store the vendored dashboard code, and removes any duplicate files from `web/gui` that are included with the vendored dashboard code. It also adds a Python script to generate a `Makefile.am` for the vendored code, and includes the required changes to the build system to properly install said code. The current version uses the code from the v2.17.1 release of the dashboard code. The overall structure was designed to make it as easy as possible to generate clean updates to this vendored code automatically.

The third commit cleans up CI checks that are obsoleted or misconfigured as a result of the changes in the second commit.

The fourth commit updates the GHA workflow that is used to generate dashboard update PRs in the agent repo to properly update the vendored code instead of updating the files that were used by the installer code to pull in the correct dashboard version.

##### Component Name

area/web
area/packaging
area/ci

##### Test Plan

Code changes verified locally by building normally and attempting to use the dashboard of the built version.

The automation tooling can be mostly verified by running `web/gui/bundle_dashboard.py v2.17.1` to rerun the code to generate the state of the repository and looking at `git diff` output to verify that it did not actually change anything.

##### Additional Information

Closes: #11075 

This is the final step of #11072


Ferroin:
@joelhans You got automatically pulled in for review on this because it adds a new `README.md` file. The file in question is not actually user documentation, but is intended to warn people off of changing stuff in the newly added directory. It _shouldn’t_ have any actual impact on generating the documentation because nothing links to it.

vkalintiris (approved):
LATM, ie. looks awesome to me!

One thing that we should keep in mind is that committing minified stuff in git will blow up the repo size. However, (1) we are small enough for this not to be a significant issue in the near future, (2) I don't expect us to update minified sources very frequently.

@Ferroin can you confirm or provide more info on the 2nd point?

Ferroin:
> LATM, ie. looks awesome to me!
> 
> One thing that we should keep in mind is that committing minified stuff in git will blow up the repo size. However, (1) we are small enough for this not to be a significant issue in the near future, (2) I don't expect us to update minified sources very frequently.
> 
> @Ferroin can you confirm or provide more info on the 2nd point?

Updates would be whenever we have a dashboard release. We’re only keying off of actual releases though, not pre-releases. In practice this appears to be about once a month unless a bug fix release occurs. Additionally, AFAICT, _most_ of the cases of things changing will be whole files being removed and added instead of changes happening in minified files.

That said, a large repo is IMO the least objectionable option here. We _cannot_ force a build-time dependency on Node.js, and bundling at install time makes the installer more complicated and beaks things for users using third-party packages.

vkalintiris:
Thanks for the detailed response!

> That said, a large repo is IMO the least objectionable option here. We _cannot_ force a build-time dependency on Node.js, and bundling at install time makes the installer more complicated and beaks things for users using third-party packages.

Agreed. I asked because we've already started depending on git submodules and I was wondering if it would make sense to import the dashboard files as an additional submodule to avoid growing the size of the main `netdata` repo. However, like you said, it's not going to be an issue any time soon.

knatsakis (approved):

Ferroin:
> Thanks for the detailed response!
> 
> > That said, a large repo is IMO the least objectionable option here. We _cannot_ force a build-time dependency on Node.js, and bundling at install time makes the installer more complicated and beaks things for users using third-party packages.
> 
> Agreed. I asked because we've already started depending on git submodules and I was wondering if it would make sense to import the dashboard files as an additional submodule to avoid growing the size of the main `netdata` repo. However, like you said, it's not going to be an issue any time soon.

Doing it with submodules just moves where the repository bloat happens. We need the compiled JS code regardless, and we still would not be forcing a build-time Node.JS dependency on our users,therefore we would have to commit the minified files to the dashboard repo instead, and woud have the same issue there.

kaskavel:
Checked this branch in on local MacOS agent installation. Dashboard works fine, no issues found.

kaskavel (approved):

OdysLam:
@kaskavel how did you install the agent in the macos installation? Can you also verify that netdata-cloud continues not to play ball with macos?

kaskavel:
> @kaskavel how did you install the agent in the macos installation? Can you also verify that netdata-cloud continues not to play ball with macos?

@OdysLam issuing `sudo ./netdata-installer.sh --install /usr/local` from a local clone of this branch. 
Actually I was able to claim the node in production and I am able to sign in to cloud from the agent dashboard with no apparent problems.

Ferroin:
@OdysLam Cloud support should work fine with this on macOS, but only using the legacy ACLK code, which means that the Homebrew packages will still not work correctly until the `mqtt_websockets` fixes get merged. They will, however, get all the other features that are in the React dashboard but not the legacy dashboard without having to do anything.

jacekkolasa:
@OdysLam @Ferroin FTR - it my Intel Mac it works OK, including Cloud - but i haven't been able to install it on M1 Mac Mini. But that's a different discussion i guess

jacekkolasa (approved):
Amazing work! Thanks for the detailed per-commit descriptions, it made CR indeed much easier. I haven't seen any issues, tested new/old dashboard, and static custom dashboards, all seems present and working.

Issue #7947, opened 2020-02-02
Pfsense\freebsd sysctl(vfs.bufspace...) failed: Cannot allocate memory (errno 12, Cannot allocate memory) 


kdy1968:
##### Bug report summary

```bash
system.ram chart not available
2020-02-02 07:46:13: netdata ERROR : PLUGIN[freebsd] : FREEBSD: sysctl(vfs.bufspace...) failed: Cannot allocate memory (errno 12, Cannot allocate memory)
2020-02-02 07:46:13: netdata ERROR : PLUGIN[freebsd] : DISABLED: system.ram chart
2020-02-02 07:46:13: netdata ERROR : PLUGIN[freebsd] : DISABLED: system.ram module
```

##### OS / Environment
```
FreeBSD venus.mecury.local 11.3-STABLE FreeBSD 11.3-STABLE #80 54b0ad59490(RELENG_2_4_5): Sat Feb  1 17:22:58 EST 2020     root@buildbot2-nyi.netgate.com:/build/ce-crossbuild-245/obj/amd64/IuZS8Bew/build/ce-crossbuild-245/sources/FreeBSD-src/sys/pfSense  amd64
1103507
```

##### Netdata version

netdata v1.19.0

 

##### Steps To Reproduce

1. ...removed and re-added netdata, upgraded from freebsd 11.2 to 11.3
2. ...

##### Expected behavior
system.ram chart to be displayed in console

ilyam8:
i see the problem in the freebsd.plugin @vlvkobal

amoss:
@ilyam8 When you say that you can see the problem could you please provide a little more information - can you see a bug in the code or do you mean that you can replicate the problem on a similar system?

ilyam8:
@amoss, no i cant replicate the problem (didnt try). @thiagoftsm assigned sre team guys.

I see in the OP

> 2020-02-02 07:46:13: netdata ERROR : PLUGIN[freebsd] : FREEBSD: sysctl(vfs.bufspace...) failed: Cannot allocate memory (errno 12, Cannot allocate memory)


and assume that correct person is @vlvkobal

amoss:
I think the question is who can try to replicate the problem most effectively - @vlvkobal is more familiar with that part of the code but may not have a pfsense configuration for testing. We need to replicate the problem to check that it is a bug and not part of the configuration on the original system. Who can try a pfsense configuration most effectively (because they already have one available) ?

kdy1968:
[...]

PR #11142, opened 2021-05-13
Use size_t instead of int for vfs_bufspace_count in FreeBSD plugin

diizzyy:
This fixes memory allocation errors on arm64
sysctl -t vfs.bufspace
vfs.bufspace: long integer
Same output on both amd64 and arm64
Thanks to Mikael Urankar for troubleshooting and providing a fix

Signed-off-by: Daniel Engberg <daniel.engberg.lists@pyret.net>


ilyam8:
@diizzyy thanks for the contribution! There is CLA that needs to be signed 😉

vlvkobal:
I believe the PR should fix #7947 👍

vlvkobal (approved):

diizzyy:
Done

thiagoftsm (approved):
I agree with comments from @vlvkobal . LGTM!

PR #11144, opened 2021-05-13
Add info on other memory modes to performance.md

cakrit:
##### Summary

Added use cases in performance optimization that can warrant using memory modes `ram` or `none`. 

##### Component Name

Docs

##### Test Plan

See it on the generated site.

##### Additional Information

I tried to add a link to https://learn.netdata.cloud/docs/agent/database for more info at the very end that would work both from github and on the site, but I don't think that's possible any more. @joelhans please free to improve.

### DIFF
--- a/docs/guides/configure/performance.md
+++ b/docs/guides/configure/performance.md
@@ -155,6 +155,12 @@ All the settings are found in the `[global]` section of `netdata.conf`:
     dbengine multihost disk space = 256
 ```
 
+Metric retention is not important in certain use cases, such as:
+ - Data collection nodes stream collected metrics collected to a centralization point.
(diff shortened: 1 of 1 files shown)


cakrit:
Actually, streaming and exporting could also be links to the relevant guides or reference docs. You can improve it after merging @joelhans

thiagoftsm (approved):

joelhans (approved):

joelhans:
@cakrit FYI, if you create links between docs that start from the root of the repo, they'll always work here and on Learn. For example:

`[database](/database/README.md)` resolves as expected in GitHub, and then when it's ingested into Learn it's converted into the `/docs/agent/database` format.

PR #11149, opened 2021-05-15
Remove deprecated options.

vkalintiris:
##### Summary

These options have been deprecated for a long time and they are probably not used anywhere any more.

##### Test Plan

CI

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -388,20 +388,6 @@ int help(int exitcode) {
     return exitcode;
 }
 
-// TODO: Remove this function with the nix major release.
-void remove_option(int opt_index, int *argc, char **argv) {
-    int i;
-
-    // remove the options.
-    do {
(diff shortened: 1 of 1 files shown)


thiagoftsm:
@Ferroin, please, can you confirm we are not using these options to make packages?

Ferroin:
> @Ferroin, please, can you confirm we are not using these options to make packages?

AFAICT, none of these are in use in any of our packages.

The `-pidfile` option _may_ be in use in some third-party packages, but given that these all appear to have been deprecated almost four years ago, any third-party packages still using that option are just asking for trouble at this point and thus I would not consider them worth worrying about.

thiagoftsm (approved):
After @Ferroin's comment, and some tests I think we can merge this PR. LGTM!

vlvkobal (approved):

PR #11150, opened 2021-05-15
add python.d.plugin/watchtower

tknobi:
##### Summary

A module for monitoring [Watchtower](https://github.com/containrrr/watchtower), a container-based solution for
automating Docker container base image updates.

The modules uses UrlService to collect data from watchtower's [metrics](https://containrrr.dev/watchtower/metrics/).
Also, a health check is added in case updates fail.

##### Component Name

python.d.plugin/watchtower

##### Test Plan

* compare logs of a running watchtower instance with produced charts

##### Additional Information

See [watchtower's docs](https://containrrr.dev/watchtower/metrics/).

I still have two unanswered questions:
1. in the graphs the entries with the values 0 are hidden. Is there a possibility to show them? Or is it always like this in netdata and thus a desired behavior?
2. is it possible to set up a heath alarm in case the API is not reachable (e.g. when watchtower crashed).

Looking forward to your feedback :)

Note: I am not affiliated with watchtower, I am just a user.

### DIFF
--- a/collectors/python.d.plugin/Makefile.am
+++ b/collectors/python.d.plugin/Makefile.am
@@ -108,6 +108,7 @@ include traefik/Makefile.inc
 include uwsgi/Makefile.inc
 include varnish/Makefile.inc
 include w1sensor/Makefile.inc
+include watchtower/Makefile.inc
 include web_log/Makefile.inc
 
 pythonmodulesdir=$(pythondir)/python_modules
--- a/collectors/python.d.plugin/python.d.conf
+++ b/collectors/python.d.plugin/python.d.conf
@@ -106,4 +106,5 @@ nginx_log: no
 # uwsgi: yes
 # varnish: yes
 # w1sensor: yes
+# watchtower: yes
 # web_log: yes
--- a/collectors/python.d.plugin/watchtower/Makefile.inc
+++ b/collectors/python.d.plugin/watchtower/Makefile.inc
@@ -1,0 +1,13 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# THIS IS NOT A COMPLETE Makefile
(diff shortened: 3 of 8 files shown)


andrewm4894:
@tknobi thanks for this, just curious if any particular reason you closed this or just that it's been sitting here for a few months?

I had not heard of it but watchtower looks interesting so would be a shame to lose the effort you made here.

On your questions above:

1. There is a toggle in the settings on the dashboard that will display 0 values, it just defaults to not showing them to try keep things cleaner when lots of dims on a chart and many might be all 0 in visible window.
2. I believe this would be possible in fact I think there might already be similar default alarms sort of set up like this for other api based collectors.

tknobi:
Hi @andrewm4894 ,

thanks for asking and sorry for closing without explanation.

When writing the plugin I was not yet familiar with the prometheus concept and how well it is already integrated with netdata.
Therefore, another (far from generic) prometheus parser is not necessary :)

If one wants to monitor watchtower, it is easy with the [go.d/prometheus](https://learn.netdata.cloud/docs/agent/collectors/go.d.plugin/modules/prometheus) plugin via the following configuration (`/etc/netdata/go.d/prometheus.conf`):

```yaml
jobs:
  - name: watchtower_local
    url: 'http://127.0.0.1:8099/v1/metrics'
    headers:
      Authorization: Bearer XXXXXXXXXXXXXX
    expected_prefix: 'watchtower_'
```

tknobi

andrewm4894:
Ah i see, cool makes sense. Yeah prometheus almost like a standard now. I'm actually trying to make sure our prometheus format plays nice with Grafana Cloud right now. 

Anyway, glad to hear you got sorted and agree makes sense to use something like prometheus standard when possible/supported.

(Hopefully this thread could be a useful pointer for any future watchtower users who land here :) )

PR #11152, opened 2021-05-17
Move parser from children to main thread

### LINKED ISSUE #11136: kernel crash when set "ebpf load mode = return" in /usr/lib/netdata/conf.d/ebpf.d/network.conf
##### Bug report summary
kernel crash when set "ebpf load mode = return" in /usr/lib/netdata/conf.d/ebpf.d/network.conf

Here is the serial port log.
[minicom-log.txt](https://github.com/netdata/netdata/files/6463743/minicom-log.txt)

##### OS / Environment

OS: Debian 10
Kernel: 4.19
Arch: X86

##### Netdata version

Netdata Version: 1.30.1

##### Installation method

curl -s https://packagecloud.io/install/repositories/netdata/netdata/script.deb.sh | sudo bash

##### Component Name

ebpf.plugin

##### Steps To Reproduce

1. Install netdata(1.30.1)
2. Modify the /etc/netdata/netdata.conf to enable the ebpf.plugin     ---> add  "[plugins]  ebpf = yes"
3. Modify the /usr/lib/netdata/conf.d/ebpf.d.conf   ---> network connections = yes
4. Modify the /usr/lib/netdata/conf.d/ebpf.d/network.conf    --->  ebpf load mode = return
5. run "systemctl restart netdata.service", then kernel crash

##### Expected behavior

The netdata.service works well, kernel would not crash.

thiagoftsm:
##### Summary
Fixes #11136 

This PR is fixing the reported bug moving the parser and also fixing a wrong pointer given to a function

##### Component Name
ebpf.plugin
##### Test Plan

1 - Compile this PR on Debian 10.9, and follow user's instruction:
```
Modify the /etc/netdata/netdata.conf to enable the ebpf.plugin ---> add "[plugins] ebpf = yes"
Modify the /usr/lib/netdata/conf.d/ebpf.d.conf ---> network connections = yes
Modify the /usr/lib/netdata/conf.d/ebpf.d/network.conf ---> ebpf load mode = return
run "systemctl restart netdata.service", 
```
2 - This branch won't crash. On the other hand if you try the same scenario with master we will have a crash.


vlvkobal (approved):

underhood (approved):

libnetdata/ebpf/ebpf.c
```diff
@@ -507,20 +507,22 @@ void ebpf_update_module_using_config(ebpf_module_t *modules, struct config *cfg)
...
+        read_from_user--;
+        if (!ebpf_load_config(em->cfg, filename)) {
+            error("Cannot load the ebpf configuration file %s", em->config_file);
             return;
         }
     }
 
-    ebpf_update_module_using_config(em, cfg);
+    // We only update when information is read from /etc/netdata/
+    if (read_from_user)
```
vlvkobal, line 526: Why do you need it at all? I don't think you should ignore stock configuration.

thiagoftsm: I decided to add this condition, because I thought the following scenario:

-  User changes only `ebpf.d.conf`  to have `return` as default mode.
-  Whether the condition is not present, when we load the stock files, the `return` mode won't be applied. Thanks this condition we are using default values set inside `ebpf.d.conf`.

underhood:
````
Conflicting files
collectors/ebpf.plugin/ebpf.c
```

libnetdata/ebpf/ebpf.c
vlvkobal: I think this is incorrect. If you don't want any options from the stock file to be applied, comment them out.

underhood (dismissed):

vlvkobal:
I'm approving the PR, but don't you want to comment out all global options in `ebpf.d/*.conf`? Your concern about `ebpf.d.conf` is valid.

vlvkobal (approved):

thiagoftsm:
> I'm approving the PR, but don't you want to comment out all global options in `ebpf.d/*.conf`? Your concern about `ebpf.d.conf` is valid.

You are right, but I figure out this needs to come with a new PR, this was the motive I did not bring a new commit  here.

PR #11154, opened 2021-05-17
Make storage_number symmetric and free SN_EXISTS bit flag.

vkalintiris:
##### Summary

This patch updates the storage_number format and the associated
packing/unpacking operations to:

    1. use the same exponent for min/max positive and negative values,
    2. free the SN_EXISTS bit.

With the 1st change, the smallest possible value becomes (+/-) 1e-14 from
(+/-) 1e-7.

With the 2nd change, we free-up a bit in the storage_number format which
will be reused for other purposes in follow-up PRs.

##### Test Plan

Run the tests under `libnetdata/storage_number/tests` and the agent with `-W unittest`.


vkalintiris:
@stelfrag: Should we also `memset` to `SN_EMPTY_SLOT` instead of `0`, [here](https://github.com/netdata/netdata/blob/master/database/engine/rrdengine.c#L338-L339)?

vkalintiris:
@Ferroin `exporting`'s unit tests pass on my machine (both with CMake and automake):


```
#!/usr/bin/env bash

autoreconf -ivf
./configure --without-aclk-ng

rm -rf build-tmp && mkdir -p build-tmp && pushd build-tmp

cmake -DUNIT_TESTING=1 -DBUILD_TESTING=1 -DCMAKE_BUILD_TYPE="Debug" ..
make && ctest

popd
```

exporting/tests/test_exporting_engine.c
```diff
@@ -468,7 +468,7 @@ static void test_format_dimension_stored_json_plaintext(void **state)
         "{\"prefix\":\"netdata\",\"hostname\":\"test-host\",\"host_tags\":\"TAG1=VALUE1 TAG2=VALUE2\","
         "\"chart_id\":\"chart_id\",\"chart_name\":\"chart_name\",\"chart_family\":\"(null)\"," \
         "\"chart_context\": \"(null)\",\"chart_type\":\"(null)\",\"units\": \"(null)\",\"id\":\"dimension_id\","
-        "\"name\":\"dimension_name\",\"value\":690565856.0000000,\"timestamp\": 15052}\n");
+        "\"name\":\"dimension_name\",\"value\":673788640.0000000,\"timestamp\": 15052}\n");
```
thiagoftsm, line 471: Why are you changing the constant here?

vkalintiris: Because, in the original version of my PR, I changed the representation of storage_number to consider the `SN_EXISTS_100` bit flag during unpacking. The updated PR contains only the necessary changes to free the SN_EXISTS bit. I'll post a separate PR to fix the SN_EXISTS_100 issue, because it make review (and my understanding) more complex than what it has to be.

vkalintiris:
@stelfrag + @thiagoftsm 

I marked the PR as draft because more work is needed to fix the exporting tests.

I modified the PR and - at the time of writing this comment - it contains *only* changes that make our code not use the SN_EXISTS bit.

However, this breaks the exporting tests. The reason is that the exporting code prints *raw* `storage_numbers` as `calculated_number`s without unpacking them first.

The number we are packing in the exporting tests is `27`. I wrote a small program to test every bit before/after my changes:

```
With the SN_EXISTS flag
-----------------------

[1] Will pack calculated number:
        27
[2] Output storage number:
        690565856 0x292932e0 00101001|001010010011001011100000
        Sign: 0, DivMul: 0, Exp: 101, Base_100: 0, Reset: 0, Bit24: 1
[3] Will unpack storage number:
        690565856
[4] Output calculated_number:
        27

Without the SN_EXISTS flag
--------------------------
[1] Will pack calculated number:
        27
[2] Output storage number:
        673788640 0x282932e0 00101000|001010010011001011100000
        Sign: 0, DivMul: 0, Exp: 101, Base_100: 0, Reset: 0, Bit24: 0
[3] Will unpack storage number:
        673788640
[4] Output calculated_number:
        27
```

You can see that the only difference is the bit 24, which is unset after my changes.

Apologies for the confusion/noise, I should have kept the PR simple and use a binary converter that didn't throw out leading zeros.

exporting/tests/test_exporting_engine.c
thiagoftsm: Thank you for the explanation. I will review this PR again, when you finish the work and remove the draft.

PR #11155, opened 2021-05-17
add disclaimer for macos problems

OdysLam:

### DIFF
--- a/packaging/installer/methods/macos.md
+++ b/packaging/installer/methods/macos.md
@@ -5,6 +5,10 @@ custom_edit_url: https://github.com/netdata/netdata/edit/master/packaging/instal
 
 # Install Netdata on macOS
 
+**Disclaimer:** Currently the Netdata Agent can't connect to the Netdata Cloud when used in MacOS. To experience the full feature-set of Netdata:
+1) Download [Docker Desktop for Mac](https://hub.docker.com/editions/community/docker-ce-desktop-mac)
+2) Follow the instructions for [installing Netdata using Docker](../docker/../../../README.md).
+
 Netdata works on macOS, albeit with some limitations. The number of charts displaying system metrics is limited, but you
 can use any of Netdata's [external plugins](../../../collectors/plugins.d/README.md) to monitor any services you might
 have installed on your macOS system. You could also use a macOS system as the parent node in a [streaming
@@ -13,8 +17,10 @@ configuration](/streaming/README.md).
 We recommend installing Netdata with the community-created and -maintained [**Homebrew
 package**](#install-netdata-with-the-homebrew-package). 
 
--   [Install Netdata via the Homebrew package](#install-netdata-with-the-homebrew-package)
--   [Install Netdata from source](#install-netdata-from-source)
+- [Install Netdata on macOS](#install-netdata-on-macos)
+  - [Install Netdata with the Homebrew package](#install-netdata-with-the-homebrew-package)
+  - [Install Netdata from source](#install-netdata-from-source)
+  - [What's next?](#whats-next)
 
 ## Install Netdata with the Homebrew package
 


packaging/installer/methods/macos.md
```diff
@@ -5,6 +5,10 @@ custom_edit_url: https://github.com/netdata/netdata/edit/master/packaging/instal
 
 # Install Netdata on macOS
 
+**Disclaimer:** Currently the Netdata Agent can't connect to the Netdata Cloud when used in MacOS. To experience the full feature-set of Netdata:
+1) Download [Docker Desktop for Mac](https://hub.docker.com/editions/community/docker-ce-desktop-mac)
+2) Follow the instructions for [installing Netdata using Docker](../docker/../../../README.md).
```
ilyam8, line 10: In that case, you effectively monitor Linux virtual machine, not the host itself.

OdysLam: We could say something like for testing purposes and for real production value you need linux machine. Either way, we should say that currently it's broken or remove it alltogether.

Ferroin: The key here is that Netdata Cloud does not work. As of the release of 1.31.0 though (planned for tomorrow), we should otherwise be as feature complete as reaslistically possible on macOS  compared to Linux.

And even then, it’s only installs with Homebrew where it does not work, installs with our kickstart script do (AFAIK) have working cloud support.

I think the correct way to handle this is to say that the Homebrew package does not currently have working Cloud support, but we are working on fixing it and anyone who wants to just test out Netdata Cloud can either use our Docker images with Docker Desktop for Mac, or install using our kickstart installer.

packaging/installer/methods/macos.md
```diff
@@ -13,8 +17,10 @@ configuration](/streaming/README.md).
 We recommend installing Netdata with the community-created and -maintained [**Homebrew
 package**](#install-netdata-with-the-homebrew-package). 
 
--   [Install Netdata via the Homebrew package](#install-netdata-with-the-homebrew-package)
--   [Install Netdata from source](#install-netdata-from-source)
+- [Install Netdata on macOS](#install-netdata-on-macos)
+  - [Install Netdata with the Homebrew package](#install-netdata-with-the-homebrew-package)
+  - [Install Netdata from source](#install-netdata-from-source)
+  - [What's next?](#whats-next)
```
joelhans, line 23: This TOC isn't necessary.

packaging/installer/methods/macos.md
```diff
@@ -13,8 +17,10 @@ configuration](/streaming/README.md).
 We recommend installing Netdata with the community-created and -maintained [**Homebrew
 package**](#install-netdata-with-the-homebrew-package).
```
joelhans, line 18: If we're going to recommend using Docker (or kickstart), then we should remove this sentence.

packaging/installer/methods/macos.md
joelhans: I concur with @Ferroin. If kickstart works and provides Cloud support, that should 100% be the recommendations, not Docker. We should, actually, remove all the other options if kickstart does indeed work at this point.

joelhans (changes requested):

packaging/installer/methods/macos.md
Ferroin: In terms of user experience, the Homebrew method is _far_ superior to the kickstart script or Docker, the only caveat is the lack of cloud support.

packaging/installer/methods/macos.md
Ferroin: Kickstart still has some significant UX issues, most notably that it currently requires you to explicitly specify an install prefix.

MrZammler: Just a note, and we should check more, I'm not sure if kickstart works ok on M1 macs...

Ferroin: If our installation _shell script_ does not work correctly on Apple M1 hardware, then I would argue that that is a bug in macOS, not our code.

If it’s something in the build system though, then that affects the Homebrew formula as well, not just the kickstart installer.

OdysLam: Last time I checked, the kickstart script failed to build for netdata cloud.

vkalintiris: Drive-by comment: if documentation contains disclaimers about things that are not working, they should also have a github issue link to let users get updates about the latest status.

OdysLam: This is a good comment @vkalintiris

ilyam8:
Closing because we can't merge the PR as it is (installing using Docker Desktop is very questionable - we don't monitor the host but Linux VM).

But updating the "Install Netdata on macOS" guide is a good idea, for instance:
 - be more explicit about what doesn't work when using homebrew, mac ports.
 - intel/m1 kickstart.sh compatibility

Issue #11156, opened 2021-05-17
Save information about what installation method was used when installing Netdata.


Ferroin:
This will both let us collect this data in our anonymous statistics, and provide a way for users to definitively tell us how they installed Netdata without having to remember it or look through shell history.

The current plan is to store this in a way that we can easily retrieve it with the anonymous statistics script and that users can easily check it themselves. Most likely this will be a file that gets installed somewhere under `/usr/share/netdata` containing just the tag indicating the type of install.

The intent is to identify the following types of install:

* `custom`: Anything not covered by the other identifiers, including manual builds and third-party packages.
* `kickstart-build`: Normal `kickstart.sh` install.
* `kickstart-static-<arch>`: Installation of a static build using `kickstart-static64.sh`, built for the specified architecture. The architecture should be embedded here so we can catch cases of users trying to install for the wrong CPU architecture (the install itself will succeed if they do, they just won’t be able to actually run anything).
* `manual-static-<arch>`: Installation of a static build by manually executing the self-extracting archive for the static install, built for the specified architecture.
* `oci-<arch>`: Installed using our official Docker images, though not necessarily running on Docker, built for the specified architecture.
* `binpkg-deb-<distro>-<release>-<arch>`: Installed using our official DEB packages, built for the given distro, release, and architecture. The distro and release info here is relevant to help diagnose issues if users are using the packages on ‘compatible’ systems that are not an exact match for the distro the package was built for (for example, using our Debian 10 packages on LMDE 4, which _should_ work, but may not). The architecture info is relevant for differentiating multiarch installs (they’ll report as the host CPU arch, even if the code is built for a different supported architecture).
* `binpkg-rpm-<distro>-<release>-<arch>`: Installed using our official RPM packages, built for the given distro, release, and architecture.

PR #11157, opened 2021-05-18
Store info about the installation type for later retrieval.

Ferroin:
##### Summary

This adds a file called `.install-type` in the system configuration directory for Netdata (usually `/etc/netdata` or `/opt/netdata/etc/netdata`), which will contain a couple of key/value pairs (one per line, in a format suitable for sourcing from a shell script) identifying properties of the installation method. The following key names are defined for this file:

* `INSTALL_TYPE`: Will always be present, and will consist of a single string identifying the installation method. See below for possible values.
* `PREBUILT_ARCH`: Will be present for any prebuilt installation method (Docker images, static builds, and binary packages), and will contain the CPU architecture that Netdata was built for. This makes it easier to identify cases of users installing for the wrong CPU architecture, as well as letting us see how many people are using alternate architecture installs (for example, running a 32-bit build on a 64-bit OS).
* `PREBUILT_DISTRO`: Will be present for official binary packages, and will contain a string identifying the distribution and release of that distribution that the package was built for. This will help us diagnose any issues that arise from using a package on a ‘compatible’ distro (for example, using a Debian 10 package on LMDE 4, or a CentOS 8 package on RHEL 8).

Possible values added by this PR are:

* `custom`: Anything not covered by the other identifiers, including manual builds, manually running `netdata-installer.sh`, and third-party packages.
* `kickstart-build`: Normal `kickstart.sh` install.
* `legacy-build`: Used for pre-existing `kickstart.sh` or `netdata-installer.sh` installs when they have recieved this update. This exist because we cannot determine how the install originally happened.
* `kickstart-static-<arch>`: Installation of a static build using `kickstart-static64.sh`, built for the specified architecture.
* `manual-static-<arch>`: Installation of a static build by manually executing the self-extracting archive for the static install.
* `legacy-static`: Same as `legacy-build`, but for static installs.
* `oci`: Installed using our official Docker images, though not necessarily running on Docker.
* `binpkg-deb`: Installed using one of our official DEB packages.
* `binpkg-rpm`: Installed using one of our official RPM packages.

This information will be used for three purposes:

* Reporting the exact installation type as part of our anonymous statistics. This will help us identify issues with specific types of installations, as well as helping us decide which installation methods to focus our efforts on improving. The actual changes required for this will be added in a separate PR.
* Making the install scripts better at catching and properly handling cases of attempting to install over an existing installation of a different type. Currently we have very heavy-handed checks to prevent this situation from happening because it is very likely to break things. With these changes though we should be able to update the installer to handle certain cases transparently to the user without breaking anything.
* Providing a way for users to verify which installation method was used to install Netdata. This is not especially significant at the moment, but will become much more significant in the future when we update the installer to default to using pre-built packages whenever possible.

For pre-existing installs:

* Locally built installs updated through `netdata-updater.sh` will be identified as `legacy-build`.
* Static build installs updated through `netdata-updater.sh` will be identified as `legacy-static`, but will not have a `PREBUILT_ARCH` value.
* All other types will automatically end up with the correct identifier for their install method.

##### Component Name

area/packaging

##### Test Plan

The build changes are minimal and can be tested trivially by just building Netdata. The actual logic handling of the various possible values for the `.install-type` file is also pretty easy, with most of the possible cases being a simple matter of using the code from this branch for the install. The only tricky case is the `kickstart-static-*` values, which will only work correctly if the `kickstart-static64.sh` script from this PR is used to install a static build which was also created from the PR (this is needed to ensure we properly report the architecture the static build was created for, instead of the architecture that it’s being installed on).

##### Additional Information

Fixes: #11156 

The changes for the `binpkg-*` values will is handled in https://github.com/netdata/helper-images/pull/118.


MrZammler:
Hi @Ferroin ! Looks nice! Just a question: 

Would for analytics purposes require some parsing of the value in the file? Or would it just reporting the exact value be ok ? i.e. we'll have a value of `kickstart-static-<arch>` . If that's reported as is, then to answer the question "How many use kickstart static regardless of arch", would require to either know every possible value of `<arch>`, or make a query with `like kickstart-static-%`.

Same e.g. for `binpkg-deb-<distro>-<release>-<arch>`. Would it maybe make sense to have 2 rows in the file? The first being the broad category (e.g. `kickstart-static`, `manual-static`, `binpkg-deb`, etc), and the second `<arch>`, etc ?

Or we can add a pre-defined field delimiter, and analytics code or agent can split. Again, it depends on what we want to have in posthog, @andrewm4894 any thoughts?

andrewm4894:
> Or we can add a pre-defined field delimiter, and analytics code or agent can split. Again, it depends on what we want to have in posthog, @andrewm4894 any thoughts?

We could use, simple enough, json and handle it in the posthog plugin then or even some string format like "key1=val1,key2=val2" and we could then wrangle that as needed in the posthog plugin (assuming data not going to get too messy and have keys or values that have "," or "=" in them).

andrewm4894:
Or if we have a very controlled vocab of strings (like in issue description above) then we can just handle that in the posthog plugin too. Might be a bit more fragile in some sense if things ever change. I think a json string would be best and most future proof in the even if that was to change posthog would just evolve with it but if thats too messy in terms of getting the info in the file then i'm happy to work with whatever is easiest and/or most efficient way to put some sort of info in the file and just send that raw string on to posthog.

MrZammler:
Also, a question: Will this file be created from an e.g. agent on nightlies that will be updated when a new nightly is out, or will it appear only for fresh/new installations?

Ferroin:
The caveat to using JSON here is that all of this is being generated using shell script, and constructing JSON properly in that case (when we _cannot_ count on having `jq`) is non-trivial and error prone.

The actual list of possible values is not likely to change significantly over time, but will change, and it’s important, especially for the binary packages) that that change does not need to be coordinated across multiple repos.

I think a multi-line approach is probably our best option here, ideally three lines, the first being the install type, the second being the build architecture for pre-built stuff (that is, the `<arch>` value in the existing one-line strings) and the third bieing a distro and release if it’s a binary package.

Ferroin:
> Also, a question: Will this file be created from an e.g. agent on nightlies that will be updated when a new nightly is out, or will it appear only for fresh/new installations?

It depends.

Docker images and our official binary packages will get the correct value automatically when they update.

Existing static installs will end up with the appropriate `manual-static-*` on update (we can’t detect if they were originally installed by hand or using the kickstart script, so this is the only option).

The tricky case is updates to installs that were done using the kickstart installer. Those will at the moment end up with `custom` because we cannot detect that it was installed using the kickstart installer instead of a manual build or manually running `netdata-installer.sh` (and we can’t mark `netdata-installer.sh` installs, because that will interfere with our docker images and static builds). I can easily change that though so that they get some other value (perhaps `legacy` or `unknown`?) if so desired.

andrewm4894:
@Ferroin is it possible to also add a timestamp of install to this file perhaps? Could be useful to know the age of the install to understand a bit better about the velocity of updates if this was to change regular enough or even if not then time of install could be useful for understanding the 'age' of the install for various other reasons. 

I think having a timestamp of install could help us understand more about the velocity of updates and agent version changes which could be useful. Thoughts?

Ferroin:
> @Ferroin is it possible to also add a timestamp of install to this file perhaps? Could be useful to know the age of the install to understand a bit better about the velocity of updates if this was to change regular enough or even if not then time of install could be useful for understanding the 'age' of the install for various other reasons.
> 
> I think having a timestamp of install could help us understand more about the velocity of updates and agent version changes which could be useful. Thoughts?

Do you mean when the original install happened, or when the last update happened? The second case is something we don’t actually need any new code to save, because we can just check the creation time of the `netdata` binary itself to determine that relatively reliably.

andrewm4894:
> Do you mean when the original install happened, or when the last update happened? 

Both ideally. so one line for first install and another line for last install.

MrZammler:
Perhaps we can check for `first install` by the creation date of e.g. `/etc/netdata/` ?

Ferroin:
> Perhaps we can check for `first install` by the creation date of e.g. `/etc/netdata/` ?

That may work. I’m not sure if `mkdir -p` messes with ctime for directories that already exist though...

Ferroin:
Closing and reopening to forcibly re-run checks...

Ferroin:
@MrZammler Actually, for the case of checking the install time, the better option is probably `/var/lib/netdata/registry/netdata.public.unique.id`. `/etc/netdata` will report the build time for docker images, not the initial ‘install’ time, but the machine ID will report when that instance was first run.

Ferroin:
@MrZammler @andrewm4894 I’ve updated the handling of the install type file (and updated the description of the PR to match).

Short version is that you can now just source the `.install-type` file and it will set an `INSTALL_TYPE` variable identifying the type of install, and possibly `PREBUILT_ARCH` and `PREBUILT_DISTRO` for the other parts that had been in the tag if they’re relevant.

MrZammler:
Ok, @Ferroin ! I think it looks good! And since these are now env variables, perhaps `.install-type` could be sourced from `system-info.sh` (but it doesn't know where etc/ is...), assign those env variables to `NETDATA_...` ones, and use those in `anonymous-statistics.sh` script :-) Voila, no agent change required! (Lazy mode on :-) ).But, ok, we'll see if it's possible, we'll see how we'll get them there.

MrZammler:
(Btw, probably related, but might not - general idea). Since some times when a user has a problem, we ask for things such as the install type, environment, or some lines from his logs. Now that this is here, could there be another script which the user runs when we ask him (even when submitting a bug report) that would gather some information (like this, the install method) and provide it to us for debugging?

Ferroin:
> Ok, @Ferroin ! I think it looks good! And since these are now env variables, perhaps `.install-type` could be sourced from `system-info.sh` (but it doesn't know where etc/ is...), assign those env variables to `NETDATA_...` ones, and use those in `anonymous-statistics.sh` script :-) Voila, no agent change required! (Lazy mode on :-) ).But, ok, we'll see if it's possible, we'll see how we'll get them there.

Actually, that’s probably the best way to handle this overall, as this is kind of stuff that we probably want exposed through other places that the `system-info.sh` data gets shown.

> (Btw, probably related, but might not - general idea). Since some times when a user has a problem, we ask for things such as the install type, environment, or some lines from his logs. Now that this is here, could there be another script which the user runs when we ask him (even when submitting a bug report) that would gather some information (like this, the install method) and provide it to us for debugging?

The problem is that we can’t really rely unconditionally on stuff pulled in by the install itself, because that obviously won’t be helpful if it’s an install issue.

I’ve honestly been meaning to just write a script for this specific purpose though.

Ferroin:
@netdata/agent Does anybody know why the changes here might be causing `system/.install-type` to have it’s permissions changed in the source tree when `make install` is run? That’s what’s causing the `.gitignore` check in CI to fail, and we really do not want this file to be executable (because it is, quite simply, not an executable or a script).

andrewm4894:
@Ferroin can you add a sort of spec for the `.install-type` file in the issue description? 

Just so i'm sure i understand it. I think what would be great would be something like:

```
<first_install_timestamp>
<last_install_timestamp>
<install_type>
<arch>
<release>
```

i think if we add something like above to the description then we can evolve and maintain it as we finalize how we want it. 

once we agree on this file then i can work with @MrZammler to consume this into into the telemetry.

Ferroin:
> @Ferroin can you add a sort of spec for the `.install-type` file in the issue description?
> 
> Just so i'm sure i understand it. I think what would be great would be something like:
> 
> ```
> <first_install_timestamp>
> <last_install_timestamp>
> <install_type>
> <arch>
> <release>
> ```
> 
> i think if we add something like above to the description then we can evolve and maintain it as we finalize how we want it.
> 
> once we agree on this file then i can work with @MrZammler to consume this into into the telemetry.

For the install timestamps, it’s going to be more reliable to look at specific files on the filesystem. The most recent install can be reasonably approximated by checking the ctime of the `netdata` binary itself, and the time of first install can be reasonably accurately approximated by checking the ctime of `/var/lib/netdata/registry/netdata.public.unique.id` (this is created on first run, which is the only way to get this info accurately for a Docker container).

The general format of the rest of the data though is roughly as follows:

* The first line will _always_ start with `INSTALL_TYPE=`, followed directly by the installation type as outlined in the PR description, wrapped in single quotes.
* The second line, if it exists, will always start with `PREBUILT_ARCH=`, and list the architecture name Netdata was built for (as would be output by `uname -m` on the build machine), wrapped in single quotes. This will only be present for prebuilt binary distributions of Netdata.
* The third line, if it exists, will always start with `PREBUILT_DISTRO=`, and will list the distribution name and release version that the Netdata package was built for, wrapped in single quotes. This will only be present for binary packages.

So, a regular kickstart install will give a file containing:

```bash
INSTALL_TYPE='kickstart-build'
```

A static build for 64-bit x86 that was installed by directly running the self-extracting archive will give:

```bash
INSTALL_TYPE='manual-static'
PREBUILT_ARCH='x86_64'
```

An ARM docker image will give something like:

```bash
INSTALL_TYPE='oci'
PREBUILT_ARCH='armhf'
```

And a 32-bit x86 binary package for Debian 9 will give somethinglike:

```bash
INSTALL_TYPE='binpkg-deb'
PREBUILT_ARCH='i686'
PREBUILT_DISTRO='Debian 9.0'
```

knatsakis (approved):

andrewm4894 (approved):
lgtm in terms of the information we want to add to telemetry

MrZammler (approved):
Looks good!

joelhans:
@Ferroin Sorry I didn't dig into this PR earlier, but does this have any immediate implications for our recommendations around [upgrading](https://learn.netdata.cloud/docs/agent/packaging/installer/update#determine-which-installation-method-you-used)? We have a lot of checks in that doc to help users find the right installation method and retain their installation options.

Ferroin:
> @Ferroin Sorry I didn't dig into this PR earlier, but does this have any immediate implications for our recommendations around [upgrading](https://learn.netdata.cloud/docs/agent/packaging/installer/update#determine-which-installation-method-you-used)? We have a lot of checks in that doc to help users find the right installation method and retain their installation options.

Hmm, that’s actually a good point. This should, in theory, make it much easier to determine your install method. In general, checking the type this way will only reliably identify new installs, but for those with new installs it0s a significant step up from the current situation.

nothub:
it seems like this broke updates...

```
/etc/cron.daily/netdata-updater:
cat: /opt/netdata/etc/netdata/.install-type: No such file or directory
```

ilyam8:
Thanks for reporting @nothub

PR #11177, opened 2021-05-23
health: add system clock synchronization state alarm

ilyam8:
##### Summary

This PR adds an alarm for the system clock synchronization state.

It takes into account that `ntpd` needs several minutes to select a reliable time server and sync with it.

##### Component Name

`health`

##### Test Plan

- [x] install the PR, ensure the alarm is in the stock config dir.
- [x] test the alarm triggers (not in `WARN` during select/sync period, in `WARN` after select/sync period if the clock is not synchronized).

##### Additional Information

Requested in https://community.netdata.cloud/t/add-timex-alerts/1339

### DIFF
--- a/health/health.d/timex.conf
+++ b/health/health.d/timex.conf
@@ -1,0 +1,17 @@
+
+# It can take several minutes before ntpd selects a server to synchronize with;
+# try checking after 17 minutes (1024 seconds).
+
+    alarm: system_clock_sync_state
+       on: system.clock_sync_state
+       os: linux
+    class: System
+component: Clock
+     type: Error
+     calc: $state
+    units: synchronization state
+    every: 10s
+     warn: $netdata.uptime.uptime > 17 * 60 AND $this == 0
+    delay: down 5m
+     info: the system time is not synchronized to a reliable server
+       to: silent
--- a/health/health.d/timex.conf
+++ b/health/health.d/timex.conf
@@ -11,7 +11,7 @@ component: Clock
      calc: $state
     units: synchronization state
     every: 10s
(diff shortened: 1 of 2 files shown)


health/health.d/timex.conf
```diff
@@ -0,0 +1,17 @@
...
+    class: System
+component: Clock
+     type: Error
+     calc: $state
+    units: synchronization state
+    every: 10s
+     warn: $netdata.uptime.uptime > 17 * 60 AND $this == 0
+    delay: down 5m
+     info: the system time is not synchronized to a reliable server
+       to: silent
```
ilyam8, line 17: @yasharne do you think we shouldn't make it _silent_ by default?

yasharne: Making it `silent` makes sense e.g. in our scenario only some of the servers need to have their `clock` monitored

vlvkobal (approved):

thiagoftsm (approved):

fti-cneve:
@ilyam8 
Is it specific to ntpd? Or will work with systemd-timesyncd as well?

ilyam8:
@fti-cneve **I'm not certain**, but I think yes, it will.

[timex.plugin](https://github.com/netdata/netdata/tree/master/collectors/timex.plugin) gets kernel lock sync state doing [adjtimex](https://man7.org/linux/man-pages/man2/adjtimex.2.html) call. So it shouldn't matter what you use to keep your clocks synchonized with NTP servers (e.g.: systemd-timesyncd).

fti-cneve:
@ilyam8 
I see. Thanks for the information !

Issue #11115, opened 2021-05-07
Netdata doesn't log an error when sending analytics if any


ilyam8:
##### Bug report summary

- we don't log an error if `mypopen` fails.
- we don't check `anonymous-statistics.sh` return code (`mypclose`).

https://github.com/netdata/netdata/blob/aa42f9eecd2cfb6273ee296d67ee0e0544e620f7/daemon/analytics.c#L846-L854

- we don't check HTTP response status code (`curl`/`wget` exit code could be `0`, but status code != 200/whatever expected from PostHog).

https://github.com/netdata/netdata/blob/aa42f9eecd2cfb6273ee296d67ee0e0544e620f7/daemon/anonymous-statistics.sh.in#L147-L157

##### OS / Environment

Not relevant.

##### Netdata version

```CMD
Version: netdata v1.30.1-113-g037584392
```

##### Installation method

Not relevant.

##### Component Name

`daemon/analytics`

##### Steps To Reproduce

Just check the code.

##### Expected behavior

Netdata logs an error if it fails to successfully send analytics.

PR #11188, opened 2021-05-26
Check return status of execution of anonymous statistics script

MrZammler:
##### Summary

This PR adds some checks on the execution of `anonymous-statistics.sh` script as discussed in #11115. It adds a check on the exit_code of `mypclose`, a check on `fp`, and tries to force both `curl` and `wget` to output the http response code (and check that too).

##### Component Name

Analytics

##### Test Plan

On normal cases, no error should be reported in the log file.
If an error is found in either the exit code of `mypclose`, or an http response other than `200` is produced, it should be printed in the error.log

##### Additional Information

`wget`'s way of producing the http code is hacky... I don't know if we're comfortable forcing them to give us the http response and parsing it... If not, we can only keep the checks on `mypopen` and `fp`.

### DIFF
--- a/daemon/analytics.c
+++ b/daemon/analytics.c
@@ -876,10 +876,17 @@ void send_statistics(const char *action, const char *action_result, const char *
 
     FILE *fp = mypopen(command_to_run, &command_pid);
     if (fp) {
-        char buffer[100 + 1];
-        while (fgets(buffer, 100, fp) != NULL)
-            ;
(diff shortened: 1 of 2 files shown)


vlvkobal (approved):

ilyam8:
Good work @MrZammler, thanks!

thiagoftsm (approved):
LGTM!

PR #11193, opened 2021-05-26
Check for non ascii characters in health log

MrZammler:
##### Summary

It appears that in some `health-log.db` files, there are random characters in the `family` field of the lines. Sometimes those random characters will contain a tab character, so the count of fields will be wrong. When this happens, the code that reads `class`, `component` and `type` gets executed, and result in a crash... This is fixed by reading these 3 items when at least 31 entries are in a line (instead of previously 29).

However, at that point, because of the extra tab, fields have been most likely assigned to wrong variables, and those characters even if read in correct variables will break json. So a check for non-ascii characters is added, and when it detects one, it will ignore the whole line.

We could just keep the change in line `388`, and it will prevent a crash, but there will be problems with the alarm entries.

##### Component Name

Health

##### Test Plan

[health-log.db.old.zip](https://github.com/netdata/netdata/files/6548132/health-log.db.old.zip)

Attached `health-log.db.old` file will cause a core dump, and an alarm-log with json errors. After this PR, only "correct" items should be loaded.

##### Additional Information

Need to see why those entries appear on the `family` field in the first place. The check for non-ascii characters of course only makes sense for ascii files, not sure if we/will do UTF... Comments are welcomed!!

### DIFF
--- a/health/health_log.c
+++ b/health/health_log.c
@@ -195,7 +195,7 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char 
         host->health_log_entries_written++;
         line++;
 
-        int max_entries = 33, entries = 0, non_ascii = 0;
+        int max_entries = 33, entries = 0;
         char *pointers[max_entries];
 
         char *c = buf;
@@ -206,13 +206,12 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char 
                     host->hostname,
                     line,
                     filename);
-                non_ascii = 1;
                 break;
             }
             c++;
         }
 
-        if (unlikely(non_ascii == 1))
+        if (unlikely(c))
             continue;
 
         pointers[entries++] = s++;
--- a/health/health_log.c
+++ b/health/health_log.c
@@ -211,8 +211,10 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char 
             c++;
         }
 
-        if (unlikely(c))
+        if (unlikely(*c)) {
+            errored++;
             continue;
+        }
 
         pointers[entries++] = s++;
         while(*s) {
--- a/health/health_log.c
+++ b/health/health_log.c
@@ -198,24 +198,6 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char 
         int max_entries = 33, entries = 0;
         char *pointers[max_entries];
 
-        char *c = buf;
-        while (*c) {
-            if (unlikely(!isascii((unsigned char)*c))) {
-                error(
-                    "HEALTH [%s]: line %zu of file '%s' contains a non-ascii character, ignoring.",
-                    host->hostname,
-                    line,
-                    filename);
-                break;
-            }
-            c++;
-        }
-
-        if (unlikely(*c)) {
-            errored++;
-            continue;
-        }
-
         pointers[entries++] = s++;
         while(*s) {
             if(unlikely(*s == '\t')) {
--- a/health/health_log.c
+++ b/health/health_log.c
@@ -195,8 +195,25 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char 
         host->health_log_entries_written++;
         line++;
 
-        int max_entries = 33, entries = 0;
+        int max_entries = 33, entries = 0, non_ascii = 0;
         char *pointers[max_entries];
+
+        char *c = buf;
+        while (*c) {
+            if (unlikely(!isascii(*c))) {
+                error(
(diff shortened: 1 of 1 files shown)


health/health_log.c
```diff
@@ -195,9 +195,26 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char
...
+                    host->hostname,
+                    line,
+                    filename);
+                non_ascii = 1;
+                break;
+            }
+            c++;
+        }
+
+        if (unlikely(non_ascii == 1))
```
thiagoftsm, line 215: Considering that `*c` will have `NULL` when success cases happen, I think we could remove the `non_ascii` variable.

MrZammler: Ah, yes, you are of course correct! Pushed it in https://github.com/netdata/netdata/pull/11193/commits/783ebf79d2857e6411202cad55c7d0c835af5147

thiagoftsm (approved):
Everything works on local tests. LGTM!

MrZammler:
Thank you @thiagoftsm !

health/health_log.c
```diff
@@ -198,6 +198,24 @@ static inline ssize_t health_alarm_log_read(RRDHOST *host, FILE *fp, const char
         int max_entries = 33, entries = 0;
         char *pointers[max_entries];
 
+        char *c = buf;
+        while (*c) {
+            if (unlikely(!isascii(*c))) {
```
stelfrag, line 203: This probably needs to change from `(*c)` to  `((unsigned char) *c)`

MrZammler: Hi @stelfrag, thanks! Pushed 8ec9012c1ae2896c34545312ce3b280da1a2dc1f

stelfrag:
I believe the "ascii" check is risky. 

For example changing the info description of an alarm and adding a greek description breaks things:

```
2021-06-02 10:11:07: netdata INFO  : HEALTH : thread with task id 541329 finished
2021-06-02 10:11:09: netdata ERROR : MAIN : HEALTH [futura]: line 17 of file '/home/stelios/netdata-inst/netdata/var/lib/netdata/health/health-log.db.old' contains a non-ascii character, ignoring.
2021-06-02 10:11:10: netdata ERROR : MAIN : HEALTH [futura]: line 1072 of file '/home/stelios/netdata-inst/netdata/var/lib/netdata/health/health-log.db' contains a non-ascii character, ignoring.
2021-06-02 10:11:10: netdata ERROR : MAIN : HEALTH [futura]: line 1327 of file '/home/stelios/netdata-inst/netdata/var/lib/netdata/health/health-log.db' contains a non-ascii character, ignoring.
2021-06-02 10:11:10: netdata ERROR : MAIN : HEALTH [futura]: line 1328 of file '/home/stelios/netdata-inst/netdata/var/lib/netdata/health/health-log.db' contains a non-ascii character, ignoring.
2021-06-02 10:11:10: netdata ERROR : MAIN : HEALTH [futura]: line 1482 of file '/home/stelios/netdata-inst/netdata/var/lib/netdata/health/health-log.db' contains a non-ascii character, ignoring.
2021-06-02 10:11:10: netdata ERROR : MAIN : HEALTH [futura]: line 1896 of file '/home/stelios/netdata-inst/netdata/var/lib/netdata/health/health-log.db' contains a non-ascii character, ignoring.
2021-06-02 10:11:10: netdata ERROR : MAIN : HEALTH [futura]: line 1897 of file '/home/stelios/netdata-inst/netdata/var/lib/netdata/health/health-log.db' contains a non-ascii character, ignoring.
```

MrZammler:
Yes, I know, it will break for UTF... It was one of the concerns.... I'll remove the ascii check and re-think it...

stelfrag (approved):

thiagoftsm (approved):

ilyam8:
A quick question

> However, at that point, because of the extra tab, fields have been most likely assigned to wrong variables, and those characters even if read in correct variables will break json.

Could fixing this crash introduce more problems? If I understand it right, after the fix, the agent no longer crashes (which is nice), but something else doesn't work because of that broken JSON you mention @MrZammler

MrZammler:
Yes, and that's what I tried to do with this PR.... I did have in mind that e.g. greek, or russing letters would be skipped and yes, skipping those lines is not acceptable... We had I think a while ago some similar situation...

Basically it crashes because of my reading of the new variables (class, component, type). We now fix that. The problem could have been there before those new variables, and it wouldn't crash back then, and still mess with jsons (so I don't think we introduced any more problems than what was already there. Just skipped the crash)

ilyam8:
Thanks for the explanation @MrZammler I have only one minor request - let's make the PR title reflect the changes 😄

MrZammler:
You're right, done! :-)

PR #11201, opened 2021-05-28
rename default alarms job to be "Alarms"

andrewm4894:
##### Summary

rename default alarms job to be "Alarms"

##### Component Name

collectors/python.d.plugin/alarms


joelhans (approved):
LGTM, sorry for the delay on this. FYI, there's been some conversation to change **Alarms** -> **Alerts** across all of Netdata, so be ready if and when that happens.

PR #11204, opened 2021-05-28
health: fix alarm-line-charts matching

ilyam8:
##### Summary

This PR changes [alarm-line-charts](https://learn.netdata.cloud/docs/agent/health/reference#alarm-line-charts) matching from `chart.id OR chart.name` to `chart.name`.

<details>
<summary>Chart name is on the dashboard and as a user I want the pattern to match against it.</summary>

Screenshot 2021-05-28 at 19 56 00

</details>

We convert some characters to `_`, so id sometimes != name.

```JSON
"id": "cgroup_bla-cloud-crdb-cdc-streamer-service-bla.mem_utilization",
"name": "cgroup_bla_cloud_crdb_cdc_streamer_service_bla.mem_utilization",
```

In that case, I have to **write 2 patterns to exclude the char**t. That is pretty confusing.

##### Component Name

`health/`

##### Test Plan

- ensure `charts` pattern matches against chart name only.

### DIFF
--- a/database/rrdcalctemplate.c
+++ b/database/rrdcalctemplate.c
@@ -45,9 +45,7 @@ static int rrdcalctemplate_is_there_label_restriction(RRDCALCTEMPLATE *rt,  RRDH
 }
 
 static inline int rrdcalctemplate_test_additional_restriction(RRDCALCTEMPLATE *rt, RRDSET *st) {
-    if (rt->charts_pattern &&
-        !(simple_pattern_matches(rt->charts_pattern, st->id) ||
-         simple_pattern_matches(rt->charts_pattern, st->name)))
+    if (rt->charts_pattern && !simple_pattern_matches(rt->charts_pattern, st->name))
         return 0;
 
     if (rt->family_pattern && !simple_pattern_matches(rt->family_pattern, st->family))


database/rrdcalctemplate.c
```diff
@@ -45,9 +45,7 @@ static int rrdcalctemplate_is_there_label_restriction(RRDCALCTEMPLATE *rt,  RRDH
 }
 
 static inline int rrdcalctemplate_test_additional_restriction(RRDCALCTEMPLATE *rt, RRDSET *st) {
-    if (rt->charts_pattern &&
-        !(simple_pattern_matches(rt->charts_pattern, st->id) ||
-         simple_pattern_matches(rt->charts_pattern, st->name)))
+    if (rt->charts_pattern && !simple_pattern_matches(rt->charts_pattern, st->name))
```
thiagoftsm, line 48: I think it will be better to remove the exclamation point in front of parentheses, instead to remove the ID.

ilyam8: You mean make it `if (matches(id) OR matches(name)) return 0`?

thiagoftsm: Yes, I was thinking to use:

```
    if (rt->charts_pattern &&
        (simple_pattern_matches(rt->charts_pattern, st->id) ||
        simple_pattern_matches(rt->charts_pattern, st->name)))
```

This way we would keep `name` and `id`.

ilyam8: That is opposite to our global pattern - we filter if a pattern doesn't match and you are suggesting to filter if it matches.

`simple_pattern_matches` returns:
 - 0: if it doesn't match or the match is negative (! prefix).
 - 1: only when it matches and the match is positive.

Am i missing something?

thiagoftsm: Sorry, my fault.

thiagoftsm (approved):
LGTM!

vlvkobal (approved):

PR #11210, opened 2021-05-30
Ignore dbengine journal files that can not be read

stelfrag:
##### Summary
When a read fails when processing dbengine journal files, the agent terminates by calling fatal.

This will ignore the error skip the rest of the transactions in that journal file and attempt to continue.

##### Component Name
database

##### Test Plan
- This is hard to reproduce (probably caused by disk/file corruption). However any attempt to alter
  the journal file is almost always gracefully handled with a CRC error warning as in
  `netdata ERROR : MAIN : Transaction 16 was read from disk. CRC32 check: FAILED`


thiagoftsm (approved):
I tried to replicate during four hours, I ran kernel compilation and I kept the whole hardware busy during the time, but I could not recreate. 

Taking a look on code, it is possible to verify that the goal will be reached.

vkalintiris (approved):

PR #11212, opened 2021-05-31
Reduce memory needed per dimension

stelfrag:
##### Summary
- The metric_uuid is now stored as `uuid_t` in the dimension state instead of `uuid_t *`
   It used to be compiled only if dbengine was available but now it is used in all memory modes so no need to do the extra malloc
- Remove `create_dimension_uuid` and include the functionality in `find_dimension_uuid`
##### Component Name
database

##### Test Plan
- There shouldn't be any change in the functionality
- Start the agent with memory mode dbengine (the metric UUID is directly linked with stored metrics in dbengine)
- Observe metrics being collected
- Build an agent with the PR
- Stop old agent, restart with the PR applied
  - You should have access to the old metrics

### DIFF
--- a/web/api/formatters/rrd2json.c
+++ b/web/api/formatters/rrd2json.c
@@ -18,7 +18,6 @@ static inline void free_single_rrdrim(RRDDIM *temp_rd, int archive_mode)
             freez(temp_rd->rrdset);
         }
     }
-    freez(temp_rd->state->metric_uuid);
     freez(temp_rd->state);
     freez(temp_rd);
 }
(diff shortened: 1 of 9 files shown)


underhood (approved):

thiagoftsm (approved):
I ran a PR that I am develoing using `dbengine`, I stopped it and I compile this PR. After to start the new version, I did not detect problemas, as it was suggested in the description. LGTM!

web/api/formatters/rrd2json.c
```diff
@@ -95,8 +94,6 @@ void build_context_param_list(struct context_param **param_list, RRDSET *st)
         memcpy(rd->state, rd1->state, sizeof(*rd->state));
         memcpy(&rd->state->collect_ops, &rd1->state->collect_ops, sizeof(struct rrddim_collect_ops));
         memcpy(&rd->state->query_ops, &rd1->state->query_ops, sizeof(struct rrddim_query_ops));
-        rd->state->metric_uuid = mallocz(sizeof(uuid_t));
-        uuid_copy(*rd->state->metric_uuid, *rd1->state->metric_uuid);
```
vkalintiris, line 99: Not sure I understand why we need to delete this.

vkalintiris (approved):
LGTM, with one quick question in line.

web/api/formatters/rrd2json.c
stelfrag: The `memcpy` will copy the `rd->state` which includes the `state->metric_uuid`

PR #11216, opened 2021-06-01
Switches ACLK-NG to be the default ACLK impl.

underhood:
##### Summary
The installer will now try to build ACLK-NG as default.
ACLK Legacy will be built only if requested by the user specifically (`--aclk-legacy` to the installer) or will be tried as fallback if NG could not be built for some reason

##### Component Name
autotools build

##### Test Plan
test with default params -> NG should be built
remove dependency ACLK-NG needs but Legacy doesn´t (e.g. delete mqtt_websockets submodule folder) -> Legacy should be built
make all available for both ACLK implementations and pass `--aclk-legacy` to the installer -> Legacy should be built despite NG being possible

CI will also be useful for this PR


underhood:
Closed in favor of #11223 after a slight deployment plan change

PR #11217, opened 2021-06-01
python.d: merge user/stock plugin configuration files

ilyam8:
##### Summary

This PR changes loading stock/user `python.d.plugin` configuration files logic (python.d.conf).

❗ yes, the changes concern only plugin configuration loading, not its modules (e.g.: apache.conf, mysql.conf) 

Current:
 - load user config if it exists.
 - if not load stock config.

After this PR:
 - load user config if it exists.
 - load stock config file.
 - merge user into stock.

In addition, `python.d.plugin` no longer fails to start if there are any problems on loading configuration files (e.g.: invalid YAML syntax), it starts and uses internal defaults.

---

Motivation: the feature/change was requested on our forum. See the following topics: 
- [override-config-files](https://community.netdata.cloud/t/override-config-files/1342)
- [local-support-for-on-off-plugins](https://community.netdata.cloud/t/local-support-for-on-off-plugins/1003).

##### Component Name

`collectors/python.d.plugin`

##### Test Plan

Manual tests 😐 

- check that user config get merged into stock
- check the plugin doesn't stop if there are any errors during config loading (invalid YAML syntax)

### DIFF
--- a/collectors/python.d.plugin/python.d.plugin.in
+++ b/collectors/python.d.plugin/python.d.plugin.in
@@ -500,27 +500,31 @@ class Plugin:
         self.saver = None
         self.runs = 0
 
-    def load_config(self):
-        paths = [
-            DIRS.plugin_user_config,
-            DIRS.plugin_stock_config,
-        ]
-        self.log.debug("looking for '{0}' in {1}".format(self.config_name, paths))
-        abs_path = multi_path_find(self.config_name, *paths)
-        if not abs_path:
-            self.log.warning("'{0}' was not found, using defaults".format(self.config_name))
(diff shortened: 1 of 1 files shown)


thiagoftsm (approved):

k0ste:
Sorry for late reply, just confirms that feature works as expected 🤝
Support for '.local' `pythond.d.conf` also added in netdata ansible-role here: https://github.com/k0ste/ansible-role-netdata/commit/ed04600fb5d151cb87250851c570ae724fe38fd1
Works fine with CentOS8-Stream EPEL packages 🎉

PR #11224, opened 2021-06-04
eBPF ext4 (new thread for collector)

thiagoftsm:
##### Summary
This PR is bringing the first latency charts related to filesystem for Netdata, the selected filesystem is EXT4.

Our plan is to bring all filesystems that we will monitor with an unique thread, thanks this we are calling the new thread `filesystem, instead `ext4`.

##### Component Name
ebpf.plugin
##### Test Plan

1 - Create a disk and format with mkfs.ext4.
2 - Enable the new thread inside your `ebpf.d.conf`:

```conf
[ebpf programs]
    filesystem = yes
```
3 - Compile this branch

##### Additional Information
This PR was tested on the following distributions

| Linux Distribution | kernel version |
|-------------------|----------------|
| Slackware Current | 5.12.0  |
| Slackware Current | 5.10.41 |
| Slackware Current | 5.9.16 |
| Ubuntu 18.0.4 | 5.4.106 |
| Manjaro 21.04 | 5.4.106 |
| Ubuntu 18.0.4 | 4.15.18 |
| Slackware Current | 4.14.231 |
| CentOS 7.9.2009 | 3.10.0-1160 |

PR #11225, opened 2021-06-04
Allows ACLK NG and Legacy to coexist

underhood:
##### Summary
This allows both ACLK-NG and Legacy to coexist together and agent decides at startup which one will be used based on `aclk implementation` config param of `cloud` section of `netdata.conf`

This has origin in good idea from @Ferroin (asking why can't we have both :D)

Some of the things can be done nicer... e.g. aclk_stats of `NG` and `Legacy` could be merged. This is not done as this should be done with minimal amount of effort as `Legacy` will be removed soon (would be lot of wasted effort) + time pressures.

##### Component Name
ACLK, ACLK-NG


underhood:
@Ferroin question here would be what do we cover with CI after such a drastic change... + does it make sense to do the work (update the CI) considering in few weeks if everything goes well Legacy ACLK will not exist anymore

underhood:
CI failure on CentOS 8 seems unrelated and also breaks on #11224

Ferroin:
> @Ferroin question here would be what do we cover with CI after such a drastic change... + does it make sense to do the work (update the CI) considering in few weeks if everything goes well Legacy ACLK will not exist anymore

How likely are to leverage this for future revisions to the ACLK functionality? That’s really the deciding factor here IMO.

underhood:
> > @Ferroin question here would be what do we cover with CI after such a drastic change... + does it make sense to do the work (update the CI) considering in few weeks if everything goes well Legacy ACLK will not exist anymore
> 
> How likely are to leverage this for future revisions to the ACLK functionality? That’s really the deciding factor here IMO.

Not at all. The idea is to get rid of legacy ASAP. We are keeping it in case some users have issues we are not aware of yet and which might take longer time to fix/implement. (in which case we can tell to user... we are working on it please use Legacy till then)

Ferroin:
> > > @Ferroin question here would be what do we cover with CI after such a drastic change... + does it make sense to do the work (update the CI) considering in few weeks if everything goes well Legacy ACLK will not exist anymore
> > 
> > 
> > How likely are to leverage this for future revisions to the ACLK functionality? That’s really the deciding factor here IMO.
> 
> Not at all. The idea is to get rid of ACLK-Legacy ASAP. We are keeping it in case some users have issues we are not aware of yet and which might take longer time to fix/implement. (in which case we can tell to user... we are working on it please use Legacy till then)

In that case I see little to no value in special CI for checking the runtime selection.

We obviously ideally do want CI to provide some basic verification of the ACLK code itself though, but that should probably be in the unit tests.

stelfrag (approved):

MrZammler (approved):

PR #11245, opened 2021-06-14
Improve visibility of security info

joelhans:
##### Summary

This PR adds a new "overview" security doc with information that was previously in the ACLK doc and otherwise scattered throughout the repo. The goal here is to condense the information into a single place and make it more discoverable for users.

##### Component Name

area/docs


OdysLam (approved):

kickoke:
@DShreve2 If we merged this PR, 2 additional documents regarding security would come on board to the docs. I think that Joel had a specific idea how to reconcile all the privacy docs into one place, but I'm not sure about the ultimate goal here. 
I will create a topic in our backlog (trello) so we investigate further when the time allows it.

PR #11246, opened 2021-06-15
Fixed crash trying to access install-type before created

### LINKED ISSUE #11228: BUG: kickstart-static64.sh - Doesn't work after latest commit
Latest update to the kickstart-static64.sh references a file that does not exist and hence the install script will fail every time. This script does not create that file as other install scripts does, so should be very easy to reproduce.

Error: kickstart-static64.sh: 212: .: Can't open /opt/netdata/etc/netdata/.install-type

tvarsis:
Fixes #11228

### DIFF
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -208,8 +208,6 @@ safe_sha256sum() {
 
 mark_install_type() {
     install_type_file="/opt/netdata/etc/netdata/.install-type"
-    # shellcheck disable=SC1090
-    . "${install_type_file}"
     cat > "${install_type_file}" <<-EOF
 	INSTALL_TYPE='kickstart-static'
 	PREBUILT_ARCH='${PREBUILT_ARCH}'


ilyam8:
@tvarsis thank you for fixing the issue 👍 Please, sign the CLA. That is required for a PR to get merged.

Ferroin (changes requested):
While this does get rid of the error message, it creates other issues (namely, nightly static builds don’t properly record the build architecture that’s being installed).

The correct fix here is to wrap the whole function body in a conditional checking for the existence of the file referenced by `${install_type_file}`.

Dexus:
why removing it, and not use `[ -e "${install_type_file}" ] && . "${install_type_file}" ` to fix it?!

packaging/installer/kickstart-static64.sh
```diff
@@ -208,8 +208,6 @@ safe_sha256sum() {
 
 mark_install_type() {
     install_type_file="/opt/netdata/etc/netdata/.install-type"
-    # shellcheck disable=SC1090
-    . "${install_type_file}"
```
Dexus, line 212: `[ -e "${install_type_file}" ] && . "${install_type_file}"` would be the best fix here!

Dexus (changes requested):

PR #11247, opened 2021-06-15
Send correct aclk implementation used by agent to posthog.

MrZammler:
##### Summary

Send correct `aclk` implementation to posthog, since now both variants are built together.

##### Component Name

Analytics

##### Test Plan

Use the configuration variable `aclk implementation` to switch between `legacy` and `ng`, and observe that it get's picked by analytics correctly. If none of the aclk implementation is present, it will send an empty string.

### DIFF
--- a/daemon/analytics.c
+++ b/daemon/analytics.c
@@ -411,13 +411,13 @@ void analytics_misc(void)
 {
 #ifdef ENABLE_ACLK
     analytics_set_data(&analytics_data.netdata_host_cloud_available, "true");
-#ifdef ACLK_NG
-    analytics_set_data_str(&analytics_data.netdata_host_aclk_implementation, "Next Generation");
-#else
-    analytics_set_data_str(&analytics_data.netdata_host_aclk_implementation, "legacy");
-#endif
+    if (aclk_ng)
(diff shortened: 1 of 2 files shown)


daemon/analytics.c
```diff
@@ -411,13 +411,13 @@ void analytics_misc(void)
 {
 #ifdef ENABLE_ACLK
     analytics_set_data(&analytics_data.netdata_host_cloud_available, "true");
-#ifdef ACLK_NG
-    analytics_set_data_str(&analytics_data.netdata_host_aclk_implementation, "Next Generation");
-#else
-    analytics_set_data_str(&analytics_data.netdata_host_aclk_implementation, "legacy");
-#endif
+    if (aclk_ng)
```
underhood, line 414: will this always come after `aclk_starter`?

MrZammler: Yes, it will actually be called 2 minutes after the agent start (it's the amount we let to make sure things have settled a bit...). I assume it should be enough?

stelfrag (approved):

underhood (approved):

thiagoftsm (approved):
LGTM!

MrZammler:
Thanks guys!

PR #11160, opened 2021-05-18
Remove old streaming reference doc

joelhans:
##### Summary

Follow-up to #10709

Now that the [new streaming reference](/docs/metrics-storage-management/refrence-streaming.mdx) is published, we can remove almost all of the old content. I left the file in place with a little bit of information about the new user documentation in case there's a need to create _developer_ documentation in `/streaming/README.md`, which wouldn't be published on [Netdata Learn](https://learn.netdata.cloud).

> For some context on this and similar PRs, I'm slowly moving all user documentation to the `/docs` folder. This should improve how we organize and publish the docs on [Netdata Learn](https://learn.netdata.cloud), while also leaving the option to create developer documentation in various `README.md` files scattered across the codebase. By separating user vs. developer docs, we should improve the experience for both.

##### Component Name

area/streaming

PR #11249, opened 2021-06-15
[docs] streaming: add a note that streaming protocol is not HTTP

ilyam8:
##### Summary

ssia

https://community.netdata.cloud/t/properly-running-parent-child-nodes/1417 follow up

##### Component Name

`streaming/`

##### Test Plan

Not needed.

### DIFF
--- a/streaming/README.md
+++ b/streaming/README.md
@@ -20,6 +20,8 @@ capable of:
 The nodes that send metrics are called **child** nodes, and the nodes that receive metrics are called **parent** nodes.
 There are also **proxies**, which collects metrics from a child and sends it to a parent.
 
+:exclamation: This communication is not HTTP (it cannot be proxied by web proxies).
+
 ## Supported configurations
 
 ### Netdata without a database or web API (headless collector)


vlvkobal (approved):

joelhans (approved):
I'm OK with this, but the canonical doc about streaming is now here: https://learn.netdata.cloud/docs/metrics-storage-management/how-streaming-works

And the reference: https://learn.netdata.cloud/docs/metrics-storage-management/reference-streaming

The first doc includes the following statement about the protocol:

> It's important to note that the streaming connection uses TCP, UDP, or Unix sockets, not HTTP. To proxy streaming metrics, you need to use a proxy that tunnels OSI layer 4-7 traffic without interfering with it, such as SOCKS or Nginx's TCP/UDP load balancing.

I actually have a stalled out PR that's designed to remove this file since it's now been replaced: #11160

ilyam8:
Closed in favor of https://github.com/netdata/netdata/pull/11160

PR #11259, opened 2021-06-17
Add openSUSE 15.3 package builds.

Ferroin:
##### Summary

SSIA

##### Component Name

area/ci

##### Test Plan

CI checks pass.

### DIFF
--- a/packaging/scripts/install.sh
+++ b/packaging/scripts/install.sh
@@ -60,7 +60,7 @@ install_suse_like() {
 
   # Install testing tools
   zypper install -y --no-recommends \
-    curl gnu-netcat jq
+    curl netcat-openbsd jq
 }
 
 case "${DISTRO}" in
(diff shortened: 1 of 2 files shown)


kaskavel (dismissed):

packaging/scripts/install.sh
```diff
@@ -60,7 +60,7 @@ install_suse_like() {
 
   # Install testing tools
   zypper install -y --no-recommends \
-    curl gnu-netcat jq
+    curl netcat-openbsd jq
```
vkalintiris, line 63: @Ferroin Out of curiosity, why is this change required?

vkalintiris (approved):

packaging/scripts/install.sh
Ferroin: OpenSUSE has apparently stopped shipping the GNU version of netcat for some reason. We end up using netcat when checking the packages as the last part of the CI checks (and as a final layer of insurance before uploading them during nightly builds and release builds), so we need a working copy of netcat.

PR #11262, opened 2021-06-18
fix kickstart-static64.sh install script fail when trying to access `.install-type` before it is created

ilyam8:
##### Summary

Fixes: #11228

Making this PR because #11246 was closed by its author.

using @Ferroin comment as a fix

> The correct fix here is to wrap the whole function body in a conditional checking for the existence of the file referenced by ${install_type_file}

##### Component Name

`packaging/`

##### Test Plan

@Ferroin a reminder - upload kickstart filed needed after the merge.

### DIFF
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -208,7 +208,7 @@ safe_sha256sum() {
 
 mark_install_type() {
   install_type_file="/opt/netdata/etc/netdata/.install-type"
-  if [ -e "${install_type_file}" ]; then
+  if [ -f "${install_type_file}" ]; then
     # shellcheck disable=SC1090
     . "${install_type_file}"
     cat > "${install_type_file}" <<- EOF
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -207,13 +207,15 @@ safe_sha256sum() {
 }
 
 mark_install_type() {
-    install_type_file="/opt/netdata/etc/netdata/.install-type"
+  install_type_file="/opt/netdata/etc/netdata/.install-type"
+  if [ -e "${install_type_file}" ]; then
     # shellcheck disable=SC1090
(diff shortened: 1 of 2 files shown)


packaging/installer/kickstart-static64.sh
```diff
@@ -207,13 +207,15 @@ safe_sha256sum() {
 }
 
 mark_install_type() {
-    install_type_file="/opt/netdata/etc/netdata/.install-type"
+  install_type_file="/opt/netdata/etc/netdata/.install-type"
+  if [ -e "${install_type_file}" ]; then
```
Ferroin, line 211: This should be `-f` to make sure it’s actually a file. Otherwise we’re just checking that the path exists (it could be a directory or even a socket for all we know).

Ferroin (changes requested):

packaging/installer/kickstart-static64.sh
ilyam8: Thanks, fixed.

Ferroin (approved):

thiagoftsm (approved):

tvarsis:
@ilyam8 @Ferroin I closed the original pull request since I don't want to sign any agreement, even though I'm sure it is fine. If I read the code correctly, the merged fix will never create the install-type file at all since it won't exist in the first place and we only update it if it exists, hence it will never be created right? But maybe I'm missing something.

Ferroin:
> If I read the code correctly, the merged fix will never create the install-type file at all since it won't exist in the first place and we only update it if it exists, hence it will never be created right? But maybe I'm missing something.

The self-extracting archive that gets run just before this function gets called includes the file (at least, in recent nightlies it does). The only purpose of calling this function is to mark the install as having been done with with the kickstart script instead of by just manually running the self-extracting archive (this doesn’t actually change any code that gets run by the updater or anything, it’s just a way for us to be able to point users at a specific place to look when asking how they installed Netdata so we can provide better support if they run into problems).

PR #11264, opened 2021-06-18
Explicitly update libarchive on CentOS 8 when installing dependencies.

Ferroin:
##### Summary

The version of CMake currently shipped in CentOS 8 has an incorrect version limit in it0s dependency on libarchive. As a result, building LWS without an up to date copy of libarchive on such a system will result in a build failure.

This explicitly updates libarchive so that CMake will work properly. It can be removed again when we finally end support or the legacy ACLK code that depends on LWS.

##### Component Name

area/packaging

##### Test Plan

CentOS 8 CI checks pass again.

### DIFF
--- a/packaging/installer/install-required-packages.sh
+++ b/packaging/installer/install-required-packages.sh
@@ -1562,6 +1562,9 @@ validate_tree_centos() {
       fi
     fi
 
+    echo >&2 " > Updating libarchive ..."
+    run ${sudo} yum ${opts} install libarchive
+
     echo >&2 " > Installing Judy-devel directly ..."
     run ${sudo} yum ${opts} install http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/Judy-devel-1.0.5-18.module_el8.3.0+757+d382997d.x86_64.rpm
 


packaging/installer/install-required-packages.sh
```diff
@@ -1562,6 +1562,9 @@ validate_tree_centos() {
       fi
     fi
 
+    echo >&2 " > Updating libarchive ..."
```
vkalintiris, line 1565: Are we really _updating_ `libarchive` or _installing_ it?

vkalintiris (approved):
Confused by this one. IIUC, CMake references a symbol that's missing from the `libarchive` installed by default. While I'm approving this PR because its changes are trivial, I don't understand how `yum install libarchive` will fix the underlying problem? In other words, what is the after/before version of `libarchive` with/without this PR?

underhood (approved):

packaging/installer/install-required-packages.sh
Ferroin: libarchive is part of the base instal, but will not update unless you manually ‘install’ it like this.

vkalintiris: Thanks for the clarification, this answers my question above as well.

PR #11267, opened 2021-06-19
Added new postgres charts and updated standby charts to include slot_…

mjtice:
##### Summary
I've added three new charts.

The first chart adds to the autovacuum context. It measures the current percentage until a forced autovac process is kicked off.
The second and third are under a new tx wraparound context. These measure values related to the dreaded 'transaction wraparound' in postgres.
Additionally I added some naming context around the standby databases. The application_name is pretty generic (in my case, they were all called walreceiver). This makes it hard to have meaningful charts and alarms. I joined the standby query against pg_replication_slots to get the slot_name and use that if it exists, otherwise fall back on the application_name.

##### Component Name
`collectors/python.d.plugin/postgres/postgres.chart.py`
`web/gui/dashboard_info.js`

##### Test Plan
For the naming changes to the standby charts:
My test involved standing up two standby databases. The first database is tied to a replication_slot, the second is not.
I expect the output of the sql (and consequently the chart) to show both the slot_name for the standby associated with a replication slot and the application_name for the standby not associated with a replication_slot.

This has been tested on psql v. 9.6, 10, 11, 12, 13

##### Additional Information
This is a clone of PR # 11241 which needed to be closed because of mistakes with the github user doing the committing (see the comments in that PR for details).

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -55,6 +55,8 @@ QUERY_NAME_REPSLOT_FILES = 'REPSLOT_FILES'
 QUERY_NAME_IF_SUPERUSER = 'IF_SUPERUSER'
 QUERY_NAME_SERVER_VERSION = 'SERVER_VERSION'
 QUERY_NAME_AUTOVACUUM = 'AUTOVACUUM'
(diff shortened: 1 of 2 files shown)


ilyam8 (approved):

thiagoftsm (approved):
Thank you @mjtice !
LGTM!

PR #11272, opened 2021-06-22
makes ACLK-NG the default if available

underhood:
##### Summary
Switches NG to be the default.
##### Component Name
ACLK
##### Test Plan
If both ACLKs available NG will be chosen if user doesn't specify otherwise in `netdata.conf`


Ferroin:
Given that this change functionally deprecates the legacy code, we should ideally also publish a timetable for the expected removal of that code when we publish a release including this change.

MrZammler (approved):

stelfrag (approved):

PR #11274, opened 2021-06-22
Auto-detect PGID in Dockerfile's ENTRYPOINT script

OdysLam:
##### Summary

Improve the `ENTRYPOINT` script so that it auto-detects the correct PGID for the virtualization technology. This is required for correct name resolution. 

Beforehand, the user must supply the correct PGID as an environment variable. Now it detects the PGID by reading the appropriate `.sock` file, e.g `/var/run/docker.sock`. 

Ideally, we should add support for the most widely used virt engines (e.g podman, LXD containers, etc.)

##### Component Name

`/packaging/docker/run.sh`

##### Test Plan

1. Create a container with a new dockerfiler, based on `netdata/netdata` and supply a new ENTRYPOINT, using the edited `run.sh` script. Example that I am running ATM: https://github.com/OdysLam/home-urbit/tree/master/netdata

### DIFF
--- a/packaging/docker/run.sh
+++ b/packaging/docker/run.sh
@@ -13,6 +13,23 @@ if [ ! "${DO_NOT_TRACK:-0}" -eq 0 ] || [ -n "$DO_NOT_TRACK" ]; then
   touch /etc/netdata/.opt-out-from-anonymous-statistics
 fi
 
+
+BALENA_PGID=$(ls -nd /var/run/balena.sock | awk '{print $4}')
+DOCKER_PGID=$(ls -nd /var/run/docker.sock | awk '{print $4}')
+
+re='^[0-9]+$'
+if [[ $BALENA_PGID =~ $re ]]; then
+  echo "Netdata detected balena-engine.sock"
+  DOCKER_HOST='/var/run/balena-engine.sock'
+  PGID="$BALENA_PGID"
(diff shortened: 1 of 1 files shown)


packaging/docker/run.sh
```diff
@@ -13,6 +13,23 @@ if [ ! "${DO_NOT_TRACK:-0}" -eq 0 ] || [ -n "$DO_NOT_TRACK" ]; then
...
+
+BALENA_PGID=$(ls -nd /var/run/balena.sock | awk '{print $4}')
+DOCKER_PGID=$(ls -nd /var/run/docker.sock | awk '{print $4}')
+
+re='^[0-9]+$'
+if [[ $BALENA_PGID =~ $re ]]; then
+  echo "Netdata detected balena-engine.sock"
+  DOCKER_HOST='/var/run/balena-engine.sock'
+  PGID="$BALENA_PGID"
+else if [[ $DOCKER_PGID =~ $re ]]; then
```
Ferroin, line 25: This needs to be `elif` not `else if`. Fixing that should resolve most of the shellcheck errors.

OdysLam:
@Ferroin should we move forward with this? 

I tried to look into the `cgroups-name` script to find info about the other virt options and where the sockets might live.

Ferroin (approved):

vkalintiris (approved):

PR #11281, opened 2021-06-24
Fix boolean value for ProtectControlGroups

didier13150:
Hi,

In systemd service unit file, change ProtectControlGroups from true to on

On CentOS 7 , each systemd reload write the following line in /var/log/messages (`systemctl daemon-reload`)
`systemd: [/usr/lib/systemd/system/netdata.service:72] Unknown lvalue 'ProtectControlGroups' in section 'Service'`

No error was reported with new value (on or yes)

Original bug report: https://bugzilla.redhat.com/show_bug.cgi?id=1975343

### DIFF
--- a/system/netdata.service.in
+++ b/system/netdata.service.in
@@ -66,7 +66,7 @@ ProtectSystem=full
 ProtectHome=read-only
 # PrivateTmp break netdatacli functionality. See - https://github.com/netdata/netdata/issues/7587
 #PrivateTmp=true
-ProtectControlGroups=true
+ProtectControlGroups=on
 # We whitelist this because it's the standard location to listen on a UNIX socket.
 ReadWriteDirectories=/run/netdata
 # This is needed to make email-based alert deliver work if Postfix is the email provider on the system.


Ferroin:
Closing and reopening to try and force CI to actually run (something seems to be broken with it).

Ferroin:
@didier13150 I think the CLA check is confused here because of a mismatch between the email on the commit and the emails on your GitHub account. If you could look into that it would be appreciated (if that is the problem, fixing it should amount to just adding the same email you used on the commit to GitHub).

didier13150:
@Ferroin Austin I check it

didier13150:
@Ferroin I add my email addr (which I made the PR) to my GH account, Hope all is right now. If not, please report.

Ferroin (approved):
LGTM.

Failed CI jobs are either known to not be working in master, or appear to be completely unrelated to this change.

ilyam8:
@didier13150 should we also change

https://github.com/netdata/netdata/blob/6f554dc15db57da985b7ba1cfcad1a9497f33f7c/system/netdata.service.in#L20

From https://www.freedesktop.org/software/systemd/man/systemd.syntax.html

> Boolean arguments used in configuration files can be written in various formats. For positive settings the strings 1, yes, true and on are equivalent. For negative settings, the strings 0, no, false and off are equivalent.

Src code: https://github.com/systemd/systemd/blob/1160267a980a410515d8396871d16e94eefe92a0/src/basic/parse-util.c#L22-L45

ilyam8:
I don't get how come that is possible, `true` had been accepted before `ProtectControlGroups` was added.

> Unknown lvalue 'ProtectControlGroups' in section 'Service'. @didier13150 what is systemd version on Centos7?

Looks like an unknown option name (lvalue, not rvalue).

ilyam8:
ProtectControlGroups was added in v232

```cmd
systemd ((v228)) $ git checkout v231
Previous HEAD position was dd050decb6 build: bump version numbers
HEAD is now at 38b383d9fe build-sys: metadata updates for v231 (#3803)
systemd ((v231)) $ git grep ProtectControlGroup

systemd ((v231)) $ git checkout v232
Previous HEAD position was 38b383d9fe build-sys: metadata updates for v231 (#3803)
HEAD is now at a1e2ef7ec9 Revert "sd-bus: use PRIu64 instead of casting" (#4556)
systemd ((v232)) $ git grep ProtectControlGroup
NEWS:        * The new ProtectControlGroups= option can be used to disable write
...
```

ilyam8 (approved):

ilyam8 (approved):

ilyam8:
The issue is systemd versions before v232 - ProtectControlGroups was added in v232. The log message says that the key (lvalue) is unknown. The error is cosmetic because systemd just ignores the value. 

v231 - https://github.com/systemd/systemd/blob/38b383d9fe0f5c4e987c1e01136ae6073076fee3/src/shared/conf-parser.c#L157-L161

main - https://github.com/systemd/systemd/blob/f2d26cd89b195e53f184387f1a5b97a98512c82a/src/shared/conf-parser.c#L156-L160


I 🤷‍♂️ why there is no log line when changing the value to `on`, but ok 😄

vlvkobal (approved):

PR #11282, opened 2021-06-24
Properly handle the file list for updating the dashboard.

Ferroin:
##### Summary

SSIA

##### Component Name

area/ci

##### Test Plan

n/a

### DIFF
--- a/web/gui/bundle_dashboard.py
+++ b/web/gui/bundle_dashboard.py
@@ -2,7 +2,9 @@
 #
 # Copyright: © 2021 Netdata Inc.
 # SPDX-License-Identifier: GPL-3.0-or-later
-'''Bundle the dashboard code into the agent repo.'''
+'''Bundle the dashboard code into the agent repo.
+
+   This is designed to be run as part of a GHA workflow, but will work fine outside of one.'''
 
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


web/gui/bundle_dashboard.py
```diff
@@ -101,5 +111,15 @@ def write_makefile():
     BASEPATH.joinpath('Makefile.am').write_text(MAKEFILEDATA)
 
 
+def list_changed_files():
+    '''Create a list of changed files, and set it in an environment variable.'''
```
vkalintiris, line 115: I take it that the list of files is and will remain small enough so that we can stay below the system's environment size limit.

vkalintiris (approved):

web/gui/bundle_dashboard.py
Ferroin: For the moment, it seems exceedingly unlikely that we will ever have a list of file names that takes up more than 128 kiB given that the list is currently at approximately 5.2 kiB.

Ferroin:
@kaskavel @jacekkolasa it would be appreciated if I could get a second review on this so we can merge it and have working automatic PRs for dashboard releases again.

kaskavel (approved):

PR #11289, opened 2021-06-29
fix `mdstat` current operation charts context/title

ilyam8:
##### Summary

Notices that both "Approximate Time Until Finish" and "Operation Speed" have the same context ("md.rate"). 

##### Component Name

`collectors/proc/mdstat`

##### Test Plan

I checked:
 - we have no stock alarms for those charts
 - we have no info in the dashboard_info.js for those charts

Not tests needed.


thiagoftsm (approved):
LGTM!

vlvkobal (approved):

PR #11308, opened 2021-07-01
add example of how to install a third party collector

andrewm4894:
##### Summary
add example of how to install a third party collector

##### Component Name
docs
collectors

### DIFF
--- a/collectors/COLLECTORS.md
+++ b/collectors/COLLECTORS.md
@@ -534,6 +534,16 @@ default. To use a third-party collector, visit their GitHub/documentation page a
     servers.
 -   [SSH](https://github.com/Yaser-Amiri/netdata-ssh-module): Monitor failed authentication requests of an SSH server.
 
+If you wanted to install a third party collecotr into your Netdata agent you would do something like this (using [hamedbrd/netdata/tree/logparser](https://github.com/hamedbrd/netdata/tree/logparser) as an example):
+
+```bash
+# get the collector code
+sudo wget "https://raw.githubusercontent.com/hamedbrd/netdata/logparser/collectors/python.d.plugin/logparser/logparser.chart.py" -P /usr/libexec/netdata/python.d/
+
+# get the collector conf file (which you can then edit as needed)
+sudo wget "https://raw.githubusercontent.com/hamedbrd/netdata/logparser/collectors/python.d.plugin/logparser/logparser.conf" -P /etc/netdata/python.d/
+```
+
 ## Etc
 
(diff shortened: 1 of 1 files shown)


collectors/COLLECTORS.md
```diff
@@ -534,6 +534,16 @@ default. To use a third-party collector, visit their GitHub/documentation page a
     servers.
 -   [SSH](https://github.com/Yaser-Amiri/netdata-ssh-module): Monitor failed authentication requests of an SSH server.
 
+If you wanted to install a third party collecotr into your Netdata agent you would do something like this (using [hamedbrd/netdata/tree/logparser](https://github.com/hamedbrd/netdata/tree/logparser) as an example):
+
+```bash
+# get the collector code
+sudo wget "https://raw.githubusercontent.com/hamedbrd/netdata/logparser/collectors/python.d.plugin/logparser/logparser.chart.py" -P /usr/libexec/netdata/python.d/
```
ilyam8, line 541: it is not always `/usr/libexec/netdata/python.d/`

collectors/COLLECTORS.md
```diff
@@ -534,6 +534,16 @@ default. To use a third-party collector, visit their GitHub/documentation page a
...
+
+# get the collector conf file (which you can then edit as needed)
+sudo wget "https://raw.githubusercontent.com/hamedbrd/netdata/logparser/collectors/python.d.plugin/logparser/logparser.conf" -P /etc/netdata/python.d/
```
ilyam8, line 544: it is not always `/etc/netdata/python.d/`

collectors/COLLECTORS.md
```diff
@@ -534,6 +534,16 @@ default. To use a third-party collector, visit their GitHub/documentation page a
     servers.
 -   [SSH](https://github.com/Yaser-Amiri/netdata-ssh-module): Monitor failed authentication requests of an SSH server.
 
+If you wanted to install a third party collecotr into your Netdata agent you would do something like this (using [hamedbrd/netdata/tree/logparser](https://github.com/hamedbrd/netdata/tree/logparser) as an example):
```
ilyam8, line 537: Not really a good example, it is a Netdata fork branch that can be deleted at any time. Wording it differently - an example of a specific use case.

andrewm4894:
yeah - i think we need a place in here maybe too add such collectors and then can update the docs accordingly and have it be more future proof. 

@OdysLam what do you think about adding a dir in netdata/community where useful but not fully supported third party collectors can live? Then we could add example of how as a user to 'get' or 'install' such a collector. 

Just feels like we could maybe make it easier for community driven collectors to be more of a 'thing'.

PR #11311, opened 2021-07-02
Add hop count for children

stelfrag:
##### Summary
Adds a hop count for children in the system_info of the host structure. 
The local agent is hop 0 and each time a child connects to a parent it will send a `hops` variable that will +1 of its hop count.

The `hops` parameter is added to the initial STREAM connect command (child --> parent)

A parent that accepts a STREAM command (child) will initially assume that the hop count will be 1 and then:
-  Remain 1 (if the child runs an older version of the agent and therefore cannot supply a hops parameter)
- Update it to the value supplied in the STREAM command

If a child running a version with this PR connects to a parent running an older version, then the parents error.log will
have the following

```
STREAM [receive from XXXX]:nnnnnn]: request has parameter 'hops' = '1', which is not used.
STREAM [receive from XXXX]:nnnnnn]: request has parameter 'hops' = '2', which is not used.
```

for children connecting that supply a `hops` parameter with a value of 1 and 2 respectively

##### Component Name
streaming

##### Test Plan
- The agent should compile and run (no visible changes)

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -1230,6 +1230,7 @@ int main(int argc, char **argv) {
     netdata_anonymous_statistics_enabled=-1;
     struct rrdhost_system_info *system_info = calloc(1, sizeof(struct rrdhost_system_info));
     get_system_info(system_info);
+    system_info->hops = 0;
 
     if(rrd_init(netdata_configured_hostname, system_info))
         fatal("Cannot initialize localhost instance with name '%s'.", netdata_configured_hostname);
--- a/database/rrd.h
+++ b/database/rrd.h
@@ -740,6 +740,7 @@ struct rrdhost_system_info {
     char *container;
     char *container_detection;
     char *is_k8s_node;
+    uint16_t hops;
 };
 
 struct rrdhost {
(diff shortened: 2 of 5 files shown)


Ferroin:
Do we possibly want to cleanly differentiate children that are not providing this parameter? IOW, perhaps something like `-1` instead of `1` for that case, and then pass that on unmodified so we can tell the user we don’t know how many hops away the child is.

stelfrag:
> Do we possibly want to cleanly differentiate children that are not providing this parameter? IOW, perhaps something like `-1` instead of `1` for that case, and then pass that on unmodified so we can tell the user we don’t know how many hops away the child is.

The cloud will use this "hop count" to perform query routing if the child is directly claimed and streams to a parent. In this case the parent (hop count 1) will be preferred for the query to access the childs metrics -- so with that in mind, we need at least 1 there.

thiagoftsm:
@stelfrag my tests are showing that everything is OK, but before to approve I will give a suggestion, right now we have the following information when we access `/api/v1/info`:

```
"mirrored_hosts": [
		"hades",
		"zeus"
	],
	"mirrored_hosts_status": [
		{ "guid": "3", "reachable": true, "claim_id": "3" },
		{ "guid": "a", "reachable": true, "claim_id": null }
	],
```

my suggestion is we also add this parameter for this endpoint.

thiagoftsm (approved):
The tests ran fine again, LGTM!

underhood (approved):

PR #11314, opened 2021-07-05
Log message when the page cache manager sleeps for more than 1 second.

vkalintiris:
If the page cache manager can not reserve the requested number of
pages, it sleeps based on an exponential-backoff value. Normally, this
value is a couple milliseconds. However, in rare cases the sleep duration
can grow up to several seconds. This patch logs an error message
whenever the sleep period is more than 1 second.

### DIFF
--- a/database/engine/pagecache.c
+++ b/database/engine/pagecache.c
@@ -299,10 +299,15 @@ static void pg_cache_reserve_pages(struct rrdengine_instance *ctx, unsigned numb
             destroy_completion(&compl);
 
             if (unlikely(failures > 1)) {
-                unsigned long slots;
+                unsigned long slots, usecs_to_sleep;
                 /* exponential backoff */
                 slots = random() % (2LU << MIN(failures, FAILURES_CEILING));
-                (void)sleep_usec(slots * exp_backoff_slot_usec);
+                usecs_to_sleep = slots * exp_backoff_slot_usec;
+
+                if (usecs_to_sleep >= USEC_PER_SEC)
+                    error("Page cache is full. Sleeping for %llu second(s).", usecs_to_sleep / USEC_PER_SEC);
+
+                (void)sleep_usec(usecs_to_sleep);
             }
             uv_rwlock_wrlock(&pg_cache->pg_cache_rwlock);
         }


database/engine/pagecache.c
```diff
@@ -299,10 +299,15 @@ static void pg_cache_reserve_pages(struct rrdengine_instance *ctx, unsigned numb
...
             if (unlikely(failures > 1)) {
-                unsigned long slots;
+                unsigned long slots, usecs_to_sleep;
                 /* exponential backoff */
                 slots = random() % (2LU << MIN(failures, FAILURES_CEILING));
-                (void)sleep_usec(slots * exp_backoff_slot_usec);
+                usecs_to_sleep = slots * exp_backoff_slot_usec;
+
+                if (usecs_to_sleep >= USEC_PER_SEC)
+                    error("Page cache is full. Sleeping for %llu second(s).", usecs_to_sleep / USEC_PER_SEC);
```
thiagoftsm, line 308: I think I am missing something, so I have to do a question, we are not expecting wait few seconds before to proceed, and we are showing this error when this happens, is it possible to create a limit and adjust the variable `usecs_to_sleep` to avoid what is not expected? On the other hand I think we need to keep the message, because it is showing an important information.

stelfrag: I hit "request review' again by mistake :smile:

ilyam8: Is that an `error` level event? Or `info`? How often the case can occur?

vkalintiris: > is it possible to create a limit and adjust the variable usecs_to_sleep to avoid what is not expected

@thiagoftsm, my motivation for writing this patch was an agent that was getting stuck infinitely in a sleeping loop because there were no available pages to evict. The agent was built with the anomaly detection changes I'm working on, and I had to spend considerable time to figure out what was the issue. So, I think we can always fine-tune the (maximum) sleeping period, but we also want to log an error message when that is above a threshold. This will help both us (eg, when developing a new feature that might trigger this behaviour) and the users (who right now do not see any message and their report simply mentions "my dashboard does not get updated"). The reason behind picking 1 second as the threshold, is because this is the minimum update period of the agent, ie. if we block for more than 1 sec we will definitely lose values.

> Is that an error level event? Or info? How often the case can occur?

@ilyam8, it's definitely an error-level worthwhile event if the duration is more than 1 second. I don't have any idea how often this event occurs out in the wild.

thiagoftsm: I agree we need to log the information, my concern is with the fact agent can stay seconds sleeping when we expect milliseconds. 
I will approve the PR, because I already tested it, whether you bring a new commit I will retest, but whether you decide do not use this PR to create this hard limit, please, create an issue for we do not forget this possible problem.

thiagoftsm (approved):
LGTM without to forget the limit we need to create!

stelfrag (approved):

Issue #10747, opened 2021-03-10
NG per cloud request charts


underhood:
create chart per cloud req. type in NG (this chart is present in ACLK Legacy already)

UmanShahzad:
For reference, here's a full list of charts (in pictures) available for ACLK-legacy currently (in reverse order):

![Screenshot from 2021-07-07 10-38-08](https://user-images.githubusercontent.com/10018065/124712673-0ff7b300-df19-11eb-9c58-6a02cf8c37aa.png)
![Screenshot from 2021-07-07 10-38-03](https://user-images.githubusercontent.com/10018065/124712684-11c17680-df19-11eb-877f-92adc0028e86.png)
![Screenshot from 2021-07-07 10-37-54](https://user-images.githubusercontent.com/10018065/124712687-125a0d00-df19-11eb-95fe-1f13ff462571.png)
![Screenshot from 2021-07-07 10-37-50](https://user-images.githubusercontent.com/10018065/124712690-12f2a380-df19-11eb-9abe-4ee95ee34627.png)
![Screenshot from 2021-07-07 10-37-46](https://user-images.githubusercontent.com/10018065/124712691-138b3a00-df19-11eb-9b7e-2156338eb33e.png)
![Screenshot from 2021-07-07 10-37-38](https://user-images.githubusercontent.com/10018065/124712692-14bc6700-df19-11eb-84ad-0da908892e12.png)
![Screenshot from 2021-07-07 10-37-34](https://user-images.githubusercontent.com/10018065/124712696-1554fd80-df19-11eb-9437-e5312fb7874d.png)
![Screenshot from 2021-07-07 10-37-28](https://user-images.githubusercontent.com/10018065/124712699-16862a80-df19-11eb-8821-e09cde8048ca.png)
![Screenshot from 2021-07-07 10-37-22](https://user-images.githubusercontent.com/10018065/124712703-171ec100-df19-11eb-9e8b-dd50ff1f67d0.png)
![Screenshot from 2021-07-07 10-37-18](https://user-images.githubusercontent.com/10018065/124712706-17b75780-df19-11eb-99c1-e274fc8072e6.png)
![Screenshot from 2021-07-07 10-37-07](https://user-images.githubusercontent.com/10018065/124712708-184fee00-df19-11eb-9ee9-2d2e8b1800fe.png)
[...]

PR #11326, opened 2021-07-09
Add ACLK-NG cloud request type charts

UmanShahzad:
##### Summary

Fixes https://github.com/netdata/netdata/issues/10747

This PR adds 2 charts:

- `netdata.aclk_cloud_req_http_type` will show a breakdown of HTTP query types, e.g. `/api/v1/data` vs. `/api/v1/chart` etc.
- `netdata.aclk_cloud_req_type` will show a breakdown at a higher-level, e.g. `http` vs `alarm_update` vs `metadata_alarms` etc. This is based off of the entries found in `aclk/aclk_query.c#aclk_query_handlers`.

##### Component Name

aclk

##### Test Plan

Simply go to the ACLK chart submenu and see these new charts.

### DIFF
--- a/aclk/aclk_query_queue.c
+++ b/aclk/aclk_query_queue.c
@@ -49,10 +49,36 @@ int aclk_queue_query(aclk_query_t query)
 {
     int ret = _aclk_queue_query(query);
     if (!ret) {
+        // local cache of query type before we wake up query thread, which may
+        // free the query in a race.
+        aclk_query_type_t qtype = query->type;
         QUERY_THREAD_WAKEUP;
+
         if (aclk_stats_enabled) {
+            // get target query type metric before lock so we keep lock for
+            // minimal time.
+            volatile uint32_t *metric = NULL;
+            if (qtype == HTTP_API_V2)
+                metric = &aclk_metrics_per_sample.cloud_req_type_http;
+            else if (qtype == ALARM_STATE_UPDATE)
+                metric = &aclk_metrics_per_sample.cloud_req_type_alarm_upd;
+            else if (qtype == METADATA_INFO)
+                metric = &aclk_metrics_per_sample.cloud_req_type_metadata_info;
+            else if (qtype == METADATA_ALARMS)
+                metric = &aclk_metrics_per_sample.cloud_req_type_metadata_alarms;
+            else if (qtype == CHART_NEW)
+                metric = &aclk_metrics_per_sample.cloud_req_type_chart_new;
+            else if (qtype == CHART_DEL)
+                metric = &aclk_metrics_per_sample.cloud_req_type_chart_del;
+            else if (qtype == REGISTER_NODE)
+                metric = &aclk_metrics_per_sample.cloud_req_type_register_node;
+            else if (qtype == NODE_STATE_UPDATE)
+                metric = &aclk_metrics_per_sample.cloud_req_type_node_upd;
+
             ACLK_STATS_LOCK;
             aclk_metrics_per_sample.queries_queued++;
+            if (metric)
+                *metric += 1;
             ACLK_STATS_UNLOCK;
         }
     }
--- a/aclk/aclk_query_queue.c
+++ b/aclk/aclk_query_queue.c
@@ -45,6 +45,31 @@ static inline int _aclk_queue_query(aclk_query_t query)
 
 }
 
+// Gets a pointer to the metric associated with a particular query type.
+// NULL if the query type has no associated metric.
+static inline volatile uint32_t *aclk_stats_qmetric_for_qtype(aclk_query_type_t qtype) {
+    switch (qtype) {
+        case HTTP_API_V2:
+            return &aclk_metrics_per_sample.cloud_req_type_http;
+        case ALARM_STATE_UPDATE:
+            return &aclk_metrics_per_sample.cloud_req_type_alarm_upd;
+        case METADATA_INFO:
+            return &aclk_metrics_per_sample.cloud_req_type_metadata_info;
+        case METADATA_ALARMS:
+            return &aclk_metrics_per_sample.cloud_req_type_metadata_alarms;
+        case CHART_NEW:
+            return &aclk_metrics_per_sample.cloud_req_type_chart_new;
+        case CHART_DEL:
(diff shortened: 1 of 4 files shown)


UmanShahzad:
Here are sample images of the new charts:

image

image

aclk/aclk_query_queue.c
```diff
@@ -49,10 +49,36 @@ int aclk_queue_query(aclk_query_t query)
...
     if (!ret) {
+        // local cache of query type before we wake up query thread, which may
+        // free the query in a race.
+        aclk_query_type_t qtype = query->type;
         QUERY_THREAD_WAKEUP;
+
         if (aclk_stats_enabled) {
+            // get target query type metric before lock so we keep lock for
+            // minimal time.
+            volatile uint32_t *metric = NULL;
```
underhood, line 60: could we perhaps turn this into inline func and change to `switch` just for readability?

UmanShahzad: Yes that may be useful in other places indeed.

aclk/aclk_stats.c
```diff
@@ -110,6 +110,86 @@ static void aclk_stats_cloud_req(struct aclk_metrics_per_sample *per_sample)
...
+    static RRDDIM *rd_type_metadata_info = NULL;
+    static RRDDIM *rd_type_metadata_alarms = NULL;
+    static RRDDIM *rd_type_chart_new = NULL;
+    static RRDDIM *rd_type_chart_del = NULL;
+    static RRDDIM *rd_type_register_node = NULL;
+    static RRDDIM *rd_type_node_upd = NULL;
+
+    if (unlikely(!st)) {
+        st = rrdset_create_localhost(
+            "netdata", "aclk_cloud_req_type", NULL, "aclk", NULL, "Requests received from cloud via their type", "req/s",
```
underhood, line 127: i think copy paste typo.... `via` should be `by`? (it makes sense in the other chart)

UmanShahzad: In this case it wasn't actually a typo, because I think `via` can *technically* make some sense here, but definitely `by` is much better - thanks!

underhood (approved):

aclk/aclk_query_queue.c
```diff
@@ -45,14 +45,49 @@ static inline int _aclk_queue_query(aclk_query_t query)
...
     if (!ret) {
+        // local cache of query type before we wake up query thread, which may
+        // free the query in a race.
+        aclk_query_type_t qtype = query->type;
         QUERY_THREAD_WAKEUP;
+
         if (aclk_stats_enabled) {
+            // get target query type metric before lock so we keep lock for
+            // minimal time.
+            volatile uint32_t *metric = aclk_stats_qmetric_for_qtype(qtype);
```
vkalintiris, line 85: I have so many questions about the use of `volatile`. Why do we use it throughout the code? Why and/or where do we not update stats in locked-regions? Did we profile the code to check that `aclk_stats_qmetric_for_qtype`'s latency (which should be in the order of nanoseconds) justifies moving this outside of the lock?

I won't block the review because of this, but I think we really should post a follow up patch to use either locks or (builtin even) atomics when updating stats.

UmanShahzad: Those are reasonable questions @vkalintiris.

I'd imagine `volatile` is being used to avoid dealing with stale reads in a multi-threaded setting when the compiler tries to pull tricks. But you're right, we take locks, obviously for writes, and in fact even for reads when reading to insert into the timeseries db. In that case we cannot possibly be dealing with subtle data races across threads, so in the future usage of `volatile` here could be removed.

I did not profile the code to check the latency of `aclk_stats_qmetric_for_qtype`, which you're right would be often just a few cmp instructions and unnoticeable. But then again, there is no justification to move it back into the lock either - it's a free win with no effort.

I agree we should probably use atomics to update stats. I don't think the cost of a pthread mutex, which could sleep, is reasonable to pay for doing arithmetic instructions safely.

vkalintiris: Thanks @UmanShahzad!

AFAIK, ops on volatiles are not atomic and their ordering is defined relative to each other only within the context of a single thread (unless they are guarded by a lock). I asked those questions mostly to find out if we don't really care about displaying 100% accurate information 100% of the time. If this is the case:
- I guess we can get away with it for the time being, but we still should fix this at some point, either by moving these ops in locked-region or by using atomic ops.
- I'm curious if there is a quick/easy way to verify that the gcc/clang thread sanitizer does not complain. If it does, we can either prioritize a little bit more the aforementioned changes, or add some kind of suppression file.

UmanShahzad: Oh, I think there may be a misunderstanding here @vkalintiris : all changes to the stats are already in locked regions. This particular line we're commenting on is just getting the address of a particular stat variable, but updating it only later in the lock. The point is that which stat to update depends upon the query type, so we pick the right one by address outside of the lock region, then update it while locked.

So we're definitely getting 100% accurate info 100% of the time right now, whether here or elsewhere. But long-term, given that a pthread lock is almost always >= cost of a single atomic update (since it too is just an atomic CAS), we should probably move to atomics for updating/reading simple stats.

underhood: it is not so already to not have both implemented `#if defined(HAVE_C___ATOMIC) && !defined(NETDATA_NO_ATOMIC_INSTRUCTIONS)` but yeah atomics are much better and faster way to deal with this if available

vkalintiris: Thanks for the feedback, folks! I have to admit I'm a little bit confused. If read/writes are synchronized, why do we need `volatile` at all? Can't we just get a non-volatile object's address outside of locks?

UmanShahzad: @vkalintiris For this particular line, we need `volatile`, because the data being pointed to is `volatile`.

Longer term we need to change the underlying stat data to not be volatile. We can probably do that right now actually but definitely out of this PR!

(We might also be missing something because I see other parts of the codebase do similar things with `volatile`. I don't know, this is something to ask someone who knows the reasoning for the original cause of this; `git blame` shows that it might have been Costa).

stelfrag (approved):

PR #11328, opened 2021-07-10
add postgresql version to requirements section

charoleizer:
According to [this](https://github.com/netdata/netdata/blob/77c07842b1111576f521e8ee23fe0494202f31d6/collectors/python.d.plugin/postgres/postgres.chart.py#L172), postgres at version 11 is compatible

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.conf
+++ b/collectors/python.d.plugin/postgres/postgres.conf
@@ -97,14 +97,6 @@
 # the client (Netdata) is not considered local, unless it runs from inside
 # the same container. 
 #
-# Postgres supported versions are : 
-# - 9.3 (without autovacuum)
-# - 9.4
-# - 9.5
-# - 9.6
-# - 10
-# - 11
-#
 # Superuser access is needed for theses charts:
 #   Write-Ahead Logs
 #   Archive Write-Ahead Logs
--- a/collectors/python.d.plugin/postgres/postgres.conf
+++ b/collectors/python.d.plugin/postgres/postgres.conf
@@ -103,6 +103,7 @@
 # - 9.5
 # - 9.6
 # - 10
+# - 11
 #
 # Superuser access is needed for theses charts:
 #   Write-Ahead Logs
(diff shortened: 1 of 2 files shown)


ilyam8 (dismissed):

ilyam8:
@charoleizer [V11](https://github.com/netdata/netdata/blob/77c07842b1111576f521e8ee23fe0494202f31d6/collectors/python.d.plugin/postgres/postgres.chart.py#L172) variable is not used. And lack of `V*` parameter doesn't mean it doesn't work with the version. We use them for compatibility with older versions. So it works with v11, v12, and v13.

charoleizer:
> @charoleizer [V11](https://github.com/netdata/netdata/blob/77c07842b1111576f521e8ee23fe0494202f31d6/collectors/python.d.plugin/postgres/postgres.chart.py#L172) variable is not used. And lack of `V*` parameter doesn't mean it doesn't work with the version. We use them for compatibility with older versions. So it works with v11, v12, and v13.

So, can we add to documentation that it is compatible with newers version ?

At first sight, it doesn't seem to be compatible to me, and then I tried with 11.4 and it works like a charm

collectors/python.d.plugin/postgres/postgres.conf
```diff
@@ -103,6 +103,7 @@
 # - 9.5
 # - 9.6
 # - 10
+# - 11
```
charoleizer, line 106: Maybe something like # - from 9.4 to 13.3

ilyam8: Or

 - remove [_Postgres supported versions ..._](https://github.com/netdata/netdata/blob/512f9228ecdd7078f3b26b85700f32553908c522/collectors/python.d.plugin/postgres/postgres.conf#L100-L105) from postgres.conf.
 - add `Postgress v9.4+` to the readme `Requirements` section.

ilyam8 (approved):

thiagoftsm (approved):
Thank you @charoleizer!  LGTM!

PR #11335, opened 2021-07-13
Allows bundled protobuf

underhood:
##### Summary
This eases build for systems that do not have google protobuf available.
By default bundled lib will be used if present and system one as fallback. This behaviour can be changed by:
```
--with-bundled-protobuf -> only use bundled one and fail if cannot be used
--without-bundled-protobuf -> use only system shared lib
```

##### Component Name
build system

##### Test Plan
Provide protobuf built from source to `externaldeps/protobuf` folder. Should be used by default if present instead of shared lib.
Can verify by `ldd ./netdata | grep proto`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -177,6 +177,13 @@ AC_ARG_WITH(
     [AS_HELP_STRING([--with-bundled-lws], [Use the bundled version of libwebsockets library @<:@default use system library@:>@])],
     [with_bundled_lws="yes"], [with_bundled_lws="no"]
 )
+AC_ARG_WITH(
+    [bundled-protobuf],
+    [AS_HELP_STRING([--with-bundled-protobuf],
+                    [Uses the bundled version of Google Protocol Buffers @<:@default bundled if present@:>@])],
+    [with_bundled_protobuf="$withval"],
+    [with_bundled_protobuf="detect"]
+)
 
 # -----------------------------------------------------------------------------
 # Enforce building with C99, bail early if we can't.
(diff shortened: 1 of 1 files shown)


Ferroin (approved):

MrZammler:
Ok, tested:

1. Without system protobuf, and without bundled protobuf -> Builds legacy ACLK
2. Without system protobuf and with bundled protobuf -> Builds NG ACLK (and resulting binary contains google protobuf symbols)
3. Without system protobuf and without bundled protobuf and with `--with-bundled-protobuf` build fails with appropriate message. (OK)
4. Without system protobuf and without bundled protobuf and with `--without-bundled-protobuf` builds legacy ACLK.
5. With system protobuf and with bundled protobuf and with `--with-bundled-protobuf` builds NG ACLK and `ldd` and `nm` show protobuf symbols from static library
6. With system protobuf and with bundled protobuf and with `--without-bundled-protobuf` builds NG ACLK and `ldd` and `nm` show protobuf is through shared library.
7. With system protobuf and with bundled protobuf, without any parameters, it builds NG ACLK and uses the static library.

A note: netdata binary with statically libprotobuf will be ~30MB bigger than the one without.... I tried using `libprotobuf-lite.a` but it doesn't work... Don't know if it's an issue enough to change to prefer the system one if both are present ? (Of course, probably when using the system one, if it get's upgraded, netdata might need to be re-compiled....)

Ferroin:
> (Of course, probably when using the system one, if it get's upgraded, netdata might need to be re-compiled....)

_Will_ need to be recompiled. Protobuf has a strictly versioned API/ABI, and all the major distros properly reflect this in their packaging (that is, each version of protobuf is it’s own package).

The static linking will only be needed for local builds though, our static build would be getting it regardless so that doesn’t matter, the Docker image will have a consistent environment and therefore won’t need to worry about static linking, and the binary packages will just use the distro version with proper dependency management through the distro package manager.

vkalintiris:
> netdata binary with statically libprotobuf will be ~30MB bigger than the one without.... I tried using `libprotobuf-lite.a` but it doesn't work...

The stripped down version of libprotobuf on my system is 2.5 MB and the lite version is 0.5MB. We are probably building with debug info and without stripping the final binaries.

MrZammler (approved):

PR #11340, opened 2021-07-14
Allows ACLK-NG to grow MQTT buffer

underhood:
##### Summary
With Legacy Cloud, some big single JSON payloads can be bigger than the buffer. This allows the buffer to grow on the next connection with a maximum limit.
This is a workaround for now. The final solution will not need this buffer at all (its size will be 0Bytes). Also, the new cloud architecture doesn't try to do crazy things like send 10MB single message to the cloud.

##### Component Name
ACLK
##### Test Plan
Make your charts list `/api/v1/charts` sth like 5MB long. After some time agent should connect to the cloud.
Don't `ACLK_NEWARCH_DEVMODE`

### DIFF
--- a/aclk/aclk.c
+++ b/aclk/aclk.c
@@ -562,7 +562,8 @@ static int aclk_attempt_to_connect(mqtt_wss_client client)
             .will_topic = "lwt",
             .will_msg   = NULL,
             .will_flags = MQTT_WSS_PUB_QOS2,
-            .keep_alive = 60
+            .keep_alive = 60,
+            .drop_on_publish_fail = 1
         };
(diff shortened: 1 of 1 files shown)


stelfrag (approved):

MrZammler:
Couldn't try with a big enough `/api/v1/charts` so I started with a small initial `MQTT_BUFFER_SIZE`. Saw that `client->mqtt_send_buf_size` doubled each time `MQTT_WSS_ERR_TX_BUF_TOO_SMALL` was raised.

MrZammler (approved):

PR #11341, opened 2021-07-15
add geth default config

OdysLam:

### DIFF
--- a/health/health.d/geth.conf
+++ b/health/health.d/geth.conf
@@ -20,6 +20,6 @@ component: geth
     every: 10s
      calc: $goroutines
      warn: $this > (($geth.p2p_peers < 100) ? (600) : (1500))
-     warn: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
+     crit: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
     delay: up 5s
 
--- a/health/health.d/geth.conf
+++ b/health/health.d/geth.conf
@@ -1,25 +1,25 @@
- template: chainhead_diff_between_header_block
+#chainhead_header is expected momenterarily to be ahead. If its considerably ahead (e.g more than 5 blocks), then the node is definetely out of sync. 
+ template: geth_chainhead_diff_between_header_block
        on: geth.chainhead
     class: Workload
      type: ethereum_node
-component: Geth
-    every: 5s
+component: geth
+    every: 10s
      calc: $chain_head_block -  $chain_head_header
     units: blocks
      warn: $this != 0
      crit: $this > 5
     delay: up 5s
 
- template: goroutines
+# For about 50 peers, ~500 goroutines are expected. For 200 peers, ~1500 goroutines is expected. 
+ template: geth_goroutines
        on: geth.goroutines
     class: Workload
      type: ethereum_node
-component: Geth
-    every: 5s
+component: geth
+    every: 10s
      calc: $goroutines
      warn: $this > (($geth.p2p_peers < 100) ? (600) : (1500))
      warn: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
     delay: up 5s
 
-#chainhead_header is expected momenterarily to be ahead. If its considerably ahead (e.g more than 5 blocks), then the node is definetely out of sync. 
-# For about 50 peers, ~500 goroutines are expected. For 200 peers, ~1500 goroutines is expected. 
--- a/health/health.d/geth.conf
+++ b/health/health.d/geth.conf
@@ -1,23 +1,25 @@
  template: chainhead_diff_between_header_block
        on: geth.chainhead
     class: Workload
-     type: ethereum node
+     type: ethereum_node
 component: Geth
     every: 5s
      calc: $chain_head_block -  $chain_head_header
     units: blocks
      warn: $this != 0
-     crit: $this > 20
-   repeat: warning 120s critical 10s
-
+     crit: $this > 5
+    delay: up 5s
 
  template: goroutines
        on: geth.goroutines
     class: Workload
-     type: ethereum node
+     type: ethereum_node
 component: Geth
     every: 5s
      calc: $goroutines
      warn: $this > (($geth.p2p_peers < 100) ? (600) : (1500))
      warn: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
-   repeat: warning 120s critical 10s
+    delay: up 5s
+
+#chainhead_header is expected momenterarily to be ahead. If its considerably ahead (e.g more than 5 blocks), then the node is definetely out of sync. 
+# For about 50 peers, ~500 goroutines are expected. For 200 peers, ~1500 goroutines is expected. 
--- a/health/health.d/geth.conf
+++ b/health/health.d/geth.conf
@@ -9,4 +9,4 @@ component: geth
     units: blocks
      warn: $this != 0
      crit: $this > 5
-    delay: up 5s
+    delay: down 1m multiplier 1.5 max 1h
--- a/health/health.d/geth.conf
+++ b/health/health.d/geth.conf
@@ -10,16 +10,3 @@ component: geth
      warn: $this != 0
      crit: $this > 5
     delay: up 5s
-
-# For about 50 peers, ~500 goroutines are expected. For 200 peers, ~1500 goroutines is expected. 
- template: geth_goroutines
-       on: geth.goroutines
-    class: Workload
-     type: ethereum_node
-component: geth
-    every: 10s
-     calc: $goroutines
-    warn: $this > (($geth.p2p_peers_number.p2p_peers < 100) ? (550) : (1600))
-    crit: $this > (($geth.p2p_peers_number.p2p_peers < 100) ? (800) : (1700))               
-    delay: up 5s
-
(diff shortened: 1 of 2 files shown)


health/health.d/geth.conf
```diff
@@ -0,0 +1,17 @@
+template: chainhead_diff_between_header_block
+on: geth.chainhead
+every: 5s
```
thiagoftsm, line 3: Hello @OdysLam , as you can see [here](https://github.com/netdata/netdata/blob/master/health/health.d/backend.conf#L4-L6) we inserted new components for our alarms, and this template does not have them, please, can you add for us?

ilyam8: Also, consider formatting the alarms like

https://github.com/netdata/netdata/blob/03002eb22812ca342a591e9b961821e2c0afdcd1/health/health.d/adaptec_raid.conf#L4-L15

ilyam8: Still _invalid_ formatting. If you check other alarms you notice that we don't do _align left_, but _align center_.

health/health.d/geth.conf
```diff
@@ -0,0 +1,23 @@
+ template: chainhead_diff_between_header_block
+       on: geth.chainhead
+    class: Workload
+     type: Go.d plugin;
```
thiagoftsm, line 4: @ilyam8 , please, take a look here. I do not remember to see go.d.plugin written this way. :thinking:

OdysLam: Just let me merge😔

ilyam8: @OdysLam please, check [alarm-line-type](https://learn.netdata.cloud/docs/agent/health/reference#alarm-line-type) description, it is 

> Type can be used to indicate the broader area of the system that the alarm applies to. For example, under the general Database type, you can group together alarms that operate on various database systems, like MySQL, CockroachDB, CouchDB etc.

@thiagoftsm is correct.

ilyam8: Also, whatever the `type` is, don't add `;`

ilyam8:
@OdysLam you don't add this alarm to [health/Makefile.am](https://github.com/netdata/netdata/blob/35ce95760503d074057414008217a7b3e03220ef/health/Makefile.am#L26) - it won't be copied to the stock folder during installation.

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
+ template: chainhead_diff_between_header_block
```
ilyam8, line 1:
```suggestion
 template: geth_chainhead_diff_between_header_block
```

it is a common pattern to use the module name as a prefix for alarms names. This guarantees alarms names uniqueness.

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
...
+       on: geth.chainhead
+    class: Workload
+     type: ethereum_node
+component: Geth
+    every: 5s
```
ilyam8, line 6:
```suggestion
    every: 10s
```
Usually, we use 10s as the default minimum.

<details>
  <summary>stock alarms every [click!]</summary>
  
```cmd
[ilyam@pc health.d]$ grep -E -o -h  "every.*" -R * | sort | uniq -c | sort -n -r
    173 every: 10s
     79 every: 1m
     12 every: 30s
     10 every) : (60 * $update_every))
     10 every) : ( 5 * $update_every))
      5 every: 60s
      3 every: 5s
      3 every: 5m
      3 every: 20s
      3 every 1 minute
      2 every: 60
      1 every 10 seconds
```

</details>

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
...
+     calc: $goroutines
+     warn: $this > (($geth.p2p_peers < 100) ? (600) : (1500))
+     warn: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
+    delay: up 5s
+
+#chainhead_header is expected momenterarily to be ahead. If its considerably ahead (e.g more than 5 blocks), then the node is definetely out of sync.
```
ilyam8, line 24: let's put it above geth_chainhead_diff_between_header_block alarm

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
...
+     type: ethereum_node
+component: Geth
+    every: 5s
+     calc: $goroutines
+     warn: $this > (($geth.p2p_peers < 100) ? (600) : (1500))
+     warn: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
+    delay: up 5s
+
+#chainhead_header is expected momenterarily to be ahead. If its considerably ahead (e.g more than 5 blocks), then the node is definetely out of sync. 
+# For about 50 peers, ~500 goroutines are expected. For 200 peers, ~1500 goroutines is expected.
```
ilyam8, line 25: let's put it above geth_goroutines alarm

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
...
+     type: ethereum_node
+component: Geth
+    every: 5s
+     calc: $chain_head_block -  $chain_head_header
+    units: blocks
+     warn: $this != 0
+     crit: $this > 5
+    delay: up 5s
+
+ template: goroutines
```
ilyam8, line 13:
```suggestion
 template: geth_goroutines
```

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
...
+       on: geth.goroutines
+    class: Workload
+     type: ethereum_node
+component: Geth
+    every: 5s
```
ilyam8, line 18:
```suggestion
    every: 10s
```

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
+ template: chainhead_diff_between_header_block
+       on: geth.chainhead
+    class: Workload
+     type: ethereum_node
+component: Geth
```
ilyam8, line 5: What do you think about type: Etherium, component: Node?

OdysLam: That doesn't make sense. Let's keep it that way

ilyam8: > Type can be used to indicate the broader area of the system that the alarm applies to. For example, under the general Database type, you can group together alarms that operate on various database systems, like MySQL, CockroachDB, CouchDB etc.

> Ethereum is a distributed network of computers running software (known as nodes) that can verify blocks and transaction data. You need an application, known as a client, on your computer to "run" a node.

For me, it looks like the general type is Etherium. And specific component is Node (client, if it is running it makes your server a Node => part of the Etherium network).

ilyam8: Just thinking out loud 🤷‍♂️

ilyam8: [src](https://ethereum.org/en/developers/docs/nodes-and-clients/)

> "Node" refers to a running piece of client software. A client is an implementation of Ethereum that verifies all transactions in each block, keeping the network secure and the data accurate.

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
...
+# For about 50 peers, ~500 goroutines are expected. For 200 peers, ~1500 goroutines is expected. 
+ template: geth_goroutines
+       on: geth.goroutines
+    class: Workload
+     type: ethereum_node
+component: geth
+    every: 10s
+     calc: $goroutines
+     warn: $this > (($geth.p2p_peers < 100) ? (600) : (1500))
+     warn: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
```
ilyam8, line 23:
```suggestion
     warn: $this > (($geth.p2p_peers < 100) ? (600) : (1500))
     crit: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
```

Is the number of goroutines > N a critical event? Do we need `crit` at all?

health/health.d/geth.conf
```diff
@@ -0,0 +1,25 @@
...
+ template: geth_goroutines
+       on: geth.goroutines
+    class: Workload
+     type: ethereum_node
+component: geth
+    every: 10s
+     calc: $goroutines
+     warn: $this > (($geth.p2p_peers < 100) ? (600) : (1500))
+     crit: $this > (($geth.p2p_peers < 100) ? (1000) : (2000))
+    delay: up 5s
```
ilyam8, line 24: This alarm won't work. The reason is using `geth.p2p_peers`.

ilyam8: Somehow i missed that in initial review 😔

OdysLam: it works

ilyam8: Unfortunately, no. `$geth.p2p_peers < 100` is False, because there is no such chart as `geth.p2p_peers`

health/health.d/geth.conf
```diff
@@ -0,0 +1,12 @@
...
+       on: geth.chainhead
+    class: Workload
+     type: ethereum_node
+component: geth
+    every: 10s
+     calc: $chain_head_block -  $chain_head_header
+    units: blocks
+     warn: $this != 0
+     crit: $this > 5
+    delay: up 5s
```
ilyam8, line 12:
```suggestion
    delay: down 1m multiplier 1.5 max 1h
```

ilyam8 (approved):

vlvkobal (approved):

thiagoftsm (approved):
LGTM!

PR #11342, opened 2021-07-15
Assorted kickstart install fixes.

Ferroin:
##### Summary

* Use the correct path for the claiming script when installing to a non-default installation prefix with `kickstart.sh`.
* Just assume that `install-required-packages.sh` knows if it can install packages on a given OS or not instead of trying to decide that in `kickstart.sh`.

##### Component Name

area/packaging

##### Test Plan

Both of these issues broke things on macOS. The version of `kickstart.sh` from this PR should now properly handle dependencies there and properly claim the node.

### DIFF
--- a/packaging/installer/kickstart.sh
+++ b/packaging/installer/kickstart.sh
@@ -259,36 +259,32 @@ dependencies() {
   echo "Machine           : ${MACHINE}"
   echo "BASH major version: ${BASH_MAJOR_VERSION}"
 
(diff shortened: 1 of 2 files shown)


dimko (approved):

kaskavel (approved):
Only issue found is that in the case of MacOS in order to have autoclaiming working, we need to restart the agent. Perhaps we should document this.

Issue #11346, opened 2021-07-16
Running kickstart script with claim arguments when netdata is already installed, doesn't claim netdata to the cloud


dimko:
Running the following command (or executing the kickstart.sh  directly from netdata dir) when there is an already present netdata installation on user's machine doesn't claim agent to the cloud.

`bash <(curl -Ss https://my-netdata.io/kickstart.sh) --dont-wait --claim-token <token> --claim-rooms <rooms> --claim-url https://app.netdata.cloud` 

Command output:

````-- Found existing install of Netdata under: / ---
 --- Attempting to update existing install instead of creating a new one ---
[/]# //usr/libexec/netdata/netdata-updater.sh --not-running-from-cron Checking if a newer version of the updater script is available.
Fri Jul 16 15:06:05 UTC 2021 : INFO:  Running on a terminal - (this script also supports running headless from crontab)
Fri Jul 16 15:06:06 UTC 2021 : INFO:  Current Version: 00103100000124
Fri Jul 16 15:06:06 UTC 2021 : INFO:  Latest Version: 00103100000124
Fri Jul 16 15:06:06 UTC 2021 : INFO:  Newest version (current=00103100000124 >= latest=00103100000124) is already installed`

PR #11350, opened 2021-07-19
Added support for claiming existing installs via kickstarter scripts.

Ferroin:
##### Summary

This adds code to allow for using the kickstart script to claim existing installs of Netdata.

For this code to be run, the following conditions must be met:

* The user must supply correct claiming options.
* The user must not request a reinstall.

Under such circumstances, if the script can find an existing install of the correct type (using the same logic used to prevent accidental duplicate installs), it will attempt to run the claiming script from that install with the correct claiming arguments and then exit.

##### Component Name

area/packaging

##### Test Plan

See the description. When invoked correctly on a system with an existing install, the kickstart scripts from this PR should attempt to run the claiming script from that install.

##### Additional Information

Fixes: #11346 

A sizable chunk of the changes here are just moving the claiming code in the kickstart scripts into it’s own function to avoid code duplication.

### DIFF
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -218,6 +218,19 @@ mark_install_type() {
   fi
 }
 
+claim() {
+  progress "Attempting to claim agent to ${NETDATA_CLAIM_URL}"
+  NETDATA_CLAIM_PATH=/opt/netdata/bin/netdata-claim.sh
+
+  if "${NETDATA_CLAIM_PATH}" -token=${NETDATA_CLAIM_TOKEN} -rooms=${NETDATA_CLAIM_ROOMS} -url=${NETDATA_CLAIM_URL} ${NETDATA_CLAIM_EXTRA}; then
+    progress "Successfully claimed node"
+    return 0
+  else
+    run_failed "Unable to claim node, you must do so manually."
+    return 1
+  fi
+}
+
 # ----------------------------------------------------------------------------
 umask 022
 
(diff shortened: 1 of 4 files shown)


vkalintiris (approved):

kaskavel:
Hi @Ferroin, 

In the following scenario, what is the expected behavior: 

- User has a local installation of the agent
- There is a newer agent version available
- User executes [kickstart script](https://github.com/netdata/netdata/issues/11346) with claiming parameters

Apart from the agent getting claimed, should the agent get updated to the newer version or should it stay in the version it was before executing the kickstart script. We observe that it gets updated.

Ferroin:
> Hi @Ferroin,
> 
> In the following scenario, what is the expected behavior:
> 
> * User has a local installation of the agent
> * There is a newer agent version available
> * User executes [kickstart script](https://github.com/netdata/netdata/issues/11346) with claiming parameters
> 
> Apart from the agent getting claimed, should the agent get updated to the newer version or should it stay in the version it was before executing the kickstart script. We observe that it gets updated.

If the code as written is doing that, then something is very much wrong. As it is now, it should just claim and then exit in those circumstances.

kaskavel:
The previous run was done upon a local installation of an agent-ng agent. Since there are other issues to solve for these ones, I 've switched to testing this PR with legacy agents alone. In this case, it works fine. 

A possible improvement, would be to add the -id parameter in the script arguments to cover cases of reclaiming, since if the agent is already claimed somewhere else, we get the known error message.

kaskavel (approved):

Issue #11097, opened 2021-05-05
Reorganize submenu related to block device


thiagoftsm:
Right now Netdata has different menus bring information about hard disks on our dashboard and this information is spread around our menu.

![Screenshot_20210505_175833](https://user-images.githubusercontent.com/49162938/117190344-d4345600-adce-11eb-9131-6c4fcd47731f.png)

We are given suggestion to reorganize our menu and put this information near each other to simplify the user visualization. We also give the suggestion instead to have only one `Disks` submenu, we will rename it for `Block devices` and split it in two new menus. Our new organization will be:

-  Block devices: Inside this submenu only hard disks will be displayed ( `sda`, `sda1` ). Do we need to separate the partitions from the other devices?
- Mount points: Inside this submenu we will move all mount points that are now inside `Disks` for it.
- MD stat: Information about raid devices
- Filesystems: Information for all file systems, inside it we will have `btrfs`(already merged) and the new charts that we will bring with eBPF (`ext4`, `xfs`, `nfs`)

We will also move nearest the network submenu the information for `NFS server` and `NFS client`, the idea is to move below the current `Network Interface`

Ferroin:
I’m strongly of the opinion that this is a case where we would benefit from having more than just a two-level hierarchy for the side menu. Essentially, I’d love to see things structured so that partitions are clearly grouped under the associated device (and this gets more important for some cases, NVMe devices have another layer between the raw device and the partitions (namespaces) that we would want to group by as well).

As far as sorting, I’d actually put MD statistics before mount points, and we probably want to group the hardware RAID adapter collectors near the disk submenu as well.

thiagoftsm:
@Ferroin this is our main goal, but right now we as a quick win are moving these families nearest each other.

thiagoftsm:
After to have a talking with @jacekkolasa we observed that we will also need to change React dashboard to bring everything necessary to fix this issue. To avoid postpone what is possible to bring right now, I will split this issue in small issues.

Conna576:
Im so lost with all this, i do feel there is only harm with my phone so I should just throe away

ilyam8:
@thiagoftsm I suggest doing the following:
 - find all related to disk/filesystem collectors
 - propose a hierarchy for them
 - do the changes (separate issues, PRs) following the hierarhy

Otherwise, it is hard to understand/review PRs like https://github.com/netdata/netdata/pull/11119

[...]

PR #11358, opened 2021-07-22
eBPF mount

thiagoftsm:
##### Summary
This PR is bringing monitoring for `mount` and `umount` syscalls.

This PR is also bringing necessary changes inside `main.js` for we move ahead https://github.com/netdata/netdata/issues/11097 .
##### Component Name
ebpf.plugin
##### Test Plan

1 - Compile this branch.
2 - Start Netdata
3 - Mount a device and check chart `mount_points.call` you will have a peak for `mount` syscall.
4 - Umount the device now, take a look again on chart, it needs to show a peak for `umount` syscall.

##### Additional Information

This PR was tested on:

| Linux Distribution | kernel version |
|-------------------|----------------|
| Slackware Current | 5.13.4  |
| Manajaro | 5.10.49 |
| Ubuntu 18.04 | 5.4.124 |
| Debian 10.10 | 4.19.171 |
| Ubuntu 18.04 | 4.15.18 |
| Slackware Current | 4.14.239 |


vlvkobal (dismissed):

collectors/ebpf.plugin/ebpf_mount.c
```diff
@@ -0,0 +1,257 @@
...
+    netdata_thread_cleanup_push(ebpf_mount_cleanup, ptr);
+
+    ebpf_module_t *em = (ebpf_module_t *)ptr;
+    em->maps = mount_maps;
+    fill_ebpf_data(&mount_data);
+
+    if (!em->enabled)
+        goto endmount;
+
+    if (ebpf_update_kernel(&mount_data))
```
UmanShahzad, line 235: Not sure how important this is to note but, `ebpf_update_kernel` always returns 0. We may have discussed this before, but I assume this function is gonna change later? (As it currently is, it seems to just overwrite memory in `mount_data.kernel_string`, which points to a global shared 64-byte array - so all ebpf init threads are writing over each other here).

collectors/ebpf.plugin/ebpf_mount.c
```diff
@@ -0,0 +1,257 @@
...
+ * @param ptr thread data.
+ */
+static void ebpf_mount_cleanup(void *ptr)
+{
+    ebpf_module_t *em = (ebpf_module_t *)ptr;
+    if (!em->enabled)
+        return;
+
+    freez(mount_threads.thread);
+    freez(mount_values);
```
UmanShahzad, line 52: These two can be NULL frees if the function is executed before `mount_collector` is called in the mount init thread, which can happen in some error cases.

collectors/ebpf.plugin/ebpf_mount.c
```diff
@@ -0,0 +1,257 @@
...
+    }
+    read_thread_closed = 1;
+
+    return NULL;
+}
+
+/**
+ * Send data to Netdata calling auxiliar functions.
+ *
+ * @param em the structure with thread information
```
UmanShahzad, line 130: This param does not exist.

collectors/ebpf.plugin/ebpf_mount.c
```diff
@@ -0,0 +1,257 @@
...
+*/
+static void ebpf_mount_send_data()
+{
+    int i, j;
+    int end = NETDATA_EBPF_MOUNT_SYSCALL;
+    for (i = NETDATA_KEY_MOUNT_CALL, j = NETDATA_KEY_MOUNT_ERROR; i < end; i++, j++) {
```
UmanShahzad, line 136: Maybe just my opinion but, this loop is really hard to read, even though it's doing something simple. If it was just looping through an enum from start to end as in `for (i = NETDATA_KEY_MOUNT_CALL; i < NETDATA_MOUNT_END; i++)`, it would be alright, but here it's confusing.

Can we just unroll the loop and inline it with explicit enum values? e.g.

```c
// update call count
mount_publish_aggregated[NETDATA_KEY_MOUNT_CALL].ncall = mount_hash_values[NETDATA_KEY_MOUNT_CALL];
mount_publish_aggregated[NETDATA_KEY_UMOUNT_CALL].ncall = mount_hash_values[NETDATA_KEY_UMOUNT_CALL];

// update err count
mount_publish_aggregated[NETDATA_KEY_MOUNT_CALL].nerr = mount_hash_values[NETDATA_KEY_MOUNT_ERROR];
mount_publish_aggregated[NETDATA_KEY_UMOUNT_CALL].nerr = mount_hash_values[NETDATA_KEY_UMOUNT_ERROR];
```

collectors/ebpf.plugin/ebpf_mount.c
```diff
@@ -0,0 +1,257 @@
...
+
+static int read_thread_closed = 1;
+static netdata_idx_t *mount_values = NULL;
+
+static struct bpf_link **probe_links = NULL;
+static struct bpf_object *objects = NULL;
+
+static netdata_idx_t mount_hash_values[NETDATA_MOUNT_END];
+
+struct netdata_static_thread mount_threads = {"MOUNT KERNEL",
```
UmanShahzad, line 30: Since this is only a single thread, should it not be called `mount_thread` instead of `mount_threads`? When I was reading it in the code, I thought it may have referred to an array of threads.

collectors/ebpf.plugin/ebpf.d/mount.conf
```diff
@@ -0,0 +1,13 @@
+# The `ebpf load mode` option accepts the following values :
+#  `entry` : The eBPF collector only monitors calls for the functions, and does not show charts related to errors.
+#  `return : In the `return` mode, the eBPF collector monitors the same kernel functions as `entry`, but also creates
+#            new charts for the return of these functions, such as errors.
+#
+# The eBPF collector also creates charts for each running application through an integration with the `apps plugin`.
+# If you want to disable the integration with `apps.plugin` along with the above charts, change the setting `apps` to
+# 'no'.
```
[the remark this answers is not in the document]
thiagoftsm: Good point, I will push a commit removing these lines, we do not have integration with apps here, so it does not make sense to keep these lines.

collectors/ebpf.plugin/ebpf_mount.c
thiagoftsm: This function will be removed soon, I am using it here, because we need update kernel version before to load, but your conclusions are correct.

collectors/ebpf.plugin/ebpf_mount.c
thiagoftsm: This is the motive we set a test to verify whether the collector was enabled before. To call `freez` with NULL value is not a problem, observe that here we are not calling `free(3)`.

web/gui/dashboard_info.js
```diff
@@ -3563,6 +3569,14 @@ netdataDashboard.context = {
         info: 'Latency is the time it takes for an event to be completed. Netdata is attaching a kprobe for when the function <code>btrfs_sync_file</code> is called and another for when it finishes the execution. We calculate the difference between the calling and return times, we get the logarithmic for the final result and we sum one value to the respective bin. Based on the eBPF <a href="https://github.com/iovisor/bcc/blob/master/tools/btrfsdist_example.txt" target="_blank">btrfsdist</a> from BCC tools.'
     },
 
+    'mount_points.call': {
+        info: 'Monitor calls for syscalls <code>mount(2)</code> and <code>umount(2)</code> that are responsible to attach or remove file systems given as argument.'
```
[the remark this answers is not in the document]
thiagoftsm: I updated both dashboard descriptions, thanks!

collectors/ebpf.plugin/ebpf_mount.c
```diff
@@ -0,0 +1,257 @@
...
+/*****************************************************************
+ *
+ *  MAIN THREAD
+ *
+ *****************************************************************/
+
+/**
+ * Directory Cache thread
+ *
+ * Thread used to make dcstat thread
```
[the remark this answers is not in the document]
thiagoftsm: Thanks!

collectors/ebpf.plugin/ebpf_mount.c
thiagoftsm: I understand your point of view, but other threads that have more than two dimensions are using loop, so I brought loop here to keep one unique pattern.

I will give a suggestion, for this specific PR we will keep this pattern, but let us make another PR doing something like `go.d.plugin`, where there is a specific file describing how to test and develop with some `module`. I do not want to have different patterns in different threads to avoid confusion.

collectors/ebpf.plugin/ebpf_mount.c
thiagoftsm: I renamed.

UmanShahzad (approved):
image

vlvkobal (approved):

PR #11362, opened 2021-07-23
Properly handle eBPF plugin in RPM packages.

### LINKED ISSUE #10200: netdata 1.26 rpm contain a empty ebpf.plugin
##### Bug report summary
RPM for netdata 1.24 install file `/usr/libexec/netdata/plugins.d/ebpf.plugin` that is working fine but when I use the RPM for 1.26 I get an empty file (that do not work).

##### OS / Environment
RHEL (RedHat) 7.9

##### Netdata version
1.26.0 

##### Component Name
ebpf

##### Steps To Reproduce
Install netdata-1.26.0-1.el7.x86_64.rpm

Ferroin:
##### Summary

This stops trying to make the eBPF build conditional in the RPM spec file, as all the distros we’re building for are supported by our code. This, in turn, results in actually including eBPF support in our RPM packages.

A check is also added to the script used for bundling the eBPF programs to avoid installing them if the eBPF plugin did not get built for some reason.

##### Component Name

area/packaging

##### Test Plan

The packages can be built for testing by running `docker run -it --rm -v $PWD:/netdata -e VERSION=0.1 netdata/package-builders:fedora34` in the root of the Netdata source directory (you can replace `fedora34` with another distro and version number if desired, though do note that our CentOS 8 package builds are currently broken for reasons unrelated to this PR), which will put the built packages in the `artifacts` directory under the source tree. When build without this change, the package will not include `/usr/libexec/netdata/plugins.d/ebpf.plugin`, while with it it will include that file.

##### Additional Information

Fixes #10200 (and possibly other issues).

### DIFF
--- a/packaging/bundle-ebpf.sh
+++ b/packaging/bundle-ebpf.sh
@@ -6,12 +6,14 @@ PLUGINDIR="${2}"
 EBPF_VERSION="$(cat "${SRCDIR}/packaging/ebpf.version")"
 EBPF_TARBALL="netdata-kernel-collector-glibc-${EBPF_VERSION}.tar.xz"
 
-mkdir -p "${SRCDIR}/tmp/ebpf"
-curl -sSL --connect-timeout 10 --retry 3 "https://github.com/netdata/kernel-collector/releases/download/${EBPF_VERSION}/${EBPF_TARBALL}" > "${EBPF_TARBALL}" || exit 1
-grep "${EBPF_TARBALL}" "${SRCDIR}/packaging/ebpf.checksums" | sha256sum -c - || exit 1
-tar -xaf "${EBPF_TARBALL}" -C "${SRCDIR}/tmp/ebpf" || exit 1
-if [ ! -d "${PLUGINDIR}/ebpf.d" ];then
-    mkdir "${PLUGINDIR}/ebpf.d"
+if [ -x "${PLUGINDIR}/ebpf.plugin" ] ; then
+    mkdir -p "${SRCDIR}/tmp/ebpf"
+    curl -sSL --connect-timeout 10 --retry 3 "https://github.com/netdata/kernel-collector/releases/download/${EBPF_VERSION}/${EBPF_TARBALL}" > "${EBPF_TARBALL}" || exit 1
+    grep "${EBPF_TARBALL}" "${SRCDIR}/packaging/ebpf.checksums" | sha256sum -c - || exit 1
+    tar -xaf "${EBPF_TARBALL}" -C "${SRCDIR}/tmp/ebpf" || exit 1
+    if [ ! -d "${PLUGINDIR}/ebpf.d" ];then
+        mkdir "${PLUGINDIR}/ebpf.d"
+    fi
+    # shellcheck disable=SC2046
+    cp -a $(find "${SRCDIR}/tmp/ebpf" -mindepth 1 -maxdepth 1) "${PLUGINDIR}/ebpf.d"
 fi
-# shellcheck disable=SC2046
-cp -a $(find "${SRCDIR}/tmp/ebpf" -mindepth 1 -maxdepth 1) "${PLUGINDIR}/ebpf.d"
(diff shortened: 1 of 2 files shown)


thiagoftsm (approved):
After to test on CentOS 8.x I observed that eBPF programs were inside the RPM package.

packaging/bundle-ebpf.sh
```diff
@@ -6,12 +6,14 @@ PLUGINDIR="${2}"
 EBPF_VERSION="$(cat "${SRCDIR}/packaging/ebpf.version")"
 EBPF_TARBALL="netdata-kernel-collector-glibc-${EBPF_VERSION}.tar.xz"
 
-mkdir -p "${SRCDIR}/tmp/ebpf"
-curl -sSL --connect-timeout 10 --retry 3 "https://github.com/netdata/kernel-collector/releases/download/${EBPF_VERSION}/${EBPF_TARBALL}" > "${EBPF_TARBALL}" || exit 1
-grep "${EBPF_TARBALL}" "${SRCDIR}/packaging/ebpf.checksums" | sha256sum -c - || exit 1
-tar -xaf "${EBPF_TARBALL}" -C "${SRCDIR}/tmp/ebpf" || exit 1
-if [ ! -d "${PLUGINDIR}/ebpf.d" ];then
-    mkdir "${PLUGINDIR}/ebpf.d"
+if [ -x "${PLUGINDIR}/ebpf.plugin" ] ; then
```
vkalintiris, line 9: Should we print an info message if this condition is not true?

vkalintiris (approved):

packaging/bundle-ebpf.sh
Ferroin: I think we’re fine for now without it.

Saruspete:
Hello,

I don't understand how the removal of conditional code would fix anything here ?
When we edit the specfile to explicitely disable ebpf in configure step (for performance, security, and many other impact) the rpm was using the configure status, thus what we wanted.

Now, it'll just fail to build. it would have been better as a [weak dependency](https://rpm.org/user_doc/dependencies.html#weak-dependencies), wouldi't it ?

Saruspete:
For example, ebpf is either directly implied, or is used to allow to trigger flaws:
- [CVE-2021-22555 Score: 8.3](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-22555) heap out-of-bounds write vulnerability in Linux Netfilter 
- [CVE-2021-31440 Score: 8.8](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-31440) Incorrect Bounds Calculation in the Linux Kernel eBPF Verifier
- [CVE-2021-29154 Score: 7.8](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-29154) BPF JIT compilers in the Linux kernel through 5.11.12 have incorrect computation of branch displacements, allowing them to execute arbitrary code within the kernel context
- [CVE-2021-3489 Score: 7.2](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-3489) no check that the allocated size was smaller than the ringbuf size, allowing an attacker to perform out-of-bounds writes within the kernel 
- [CVE-2020-27170 Score: 4.7](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-27170)  undesirable out-of-bounds speculation on pointer arithmetic, leading to side-channel attacks that defeat Spectre mitigations
- [CVE-2020-27171 Score: 6.0](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-27171) 

Forcing eBPF to be build with netdata is a very very bad idea, production wise.

Ferroin:
The long term plan is to split stuff like this out as it’s own package (essentially the same as what we’re doing with the IPMI and CUPS plugins, albeit for different reasons in those cases), In fact, at this point I’ll probably just be working on that as a FOSS contribution on my free time as it’s been something I’ve been trying to push for pretty much since I started working here and have gotten no traction on whatsoever...

PR #11348, opened 2021-07-19
Ebpf btrfs

thiagoftsm:
##### Summary
I am waiting for https://github.com/netdata/kernel-collector/pull/243 to be merged before to give instructions to test.

##### Component Name
ebpf.plugin
##### Test Plan
1 - Format a partition with ` mkfs.btrfs` to create a `btrfs` partition
2 - Mount your new partition
3 - Configure your `/etc/netdata/ebpf.d.conf`:

```conf
[ebpf programs]
     filesystem = yes
```

You do not need to to configure `/etc/netdata/ebpf.d/filesystem.conf`, because all filesystem by default are monitored.

4 - Compile this branch
5 - Copy data to directory you mount file system and edit them. This will be enough to generate data for all charts.

##### Additional Information
This PR was tested on:

| Linux Distribution | kernel version |
|-------------------|----------------|
| Slackware Current | 5.13.2  |
| Manajaro | 5.10.49 |
| Ubuntu 18.04 | 5.4.119 |
| Debian 10.10 | 4.19.171 |
| Ubuntu 18.04 | 4.15.18 |
| Slackware Current | 4.14.239 |

PR #11363, opened 2021-07-23
Ebpf latency description

thiagoftsm:
##### Summary
We are accepting Ilya's [suggestion](https://github.com/netdata/netdata/pull/11348#issuecomment-882898320) and this PR is centralizing Latency description for eBPF latency charts.

##### Component Name
ebpf.plugin
##### Test Plan

1 - Enable filesystem inside `/etc/netdata/ebpf.d.conf`:

```
[ebpf programs]
    filesystem = yes
```
2 - Compile this Branch
3 - Access dashboard using `private mode` or forcing browser reload to see new description.

##### Additional Information
It is not necessary to test on different kernels, because we are changing only dashboard.

### DIFF
--- a/web/gui/dashboard_info.js
+++ b/web/gui/dashboard_info.js
@@ -814,6 +814,11 @@ netdataDashboard.submenu = {
     'filesystem.ZFS_latency': {
         title: 'ZFS Latency',
         info: 'Latency is the time it takes for an event to be completed. We calculate the difference between the calling and return times, we get the logarithmic for the final result and we sum one value to the respective bin. Based on the eBPF <a href="https://github.com/iovisor/bcc/blob/master/tools/zfsdist_example.txt" target="_blank">zfsdist</a> from BCC tools.'
+    },
+
+    'filesystem.BTRFS_latency': {
+        title: 'BTRFS Latency',
+        info: 'Latency is the time it takes for an event to be completed. We calculate the difference between the calling and return times, we get the logarithmic for the final result and we sum one value to the respective bin. Based on the eBPF <a href="https://github.com/iovisor/bcc/blob/master/tools/btrfsdist_example.txt" target="_blank">btrfsdist</a> from BCC tools.'
     }
 };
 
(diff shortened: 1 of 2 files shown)


web/gui/dashboard_info.js
```diff
@@ -794,6 +794,31 @@ netdataDashboard.submenu = {
     'ntpd.peers': {
         title: 'peers',
         info: 'Statistics of the peer variables for each peer configured in <code>/etc/ntp.conf</code> as shown by the readvar billboard <code>ntpq -c "rv &lt;association&gt;"</code>, while each peer is assigned a nonzero association ID as shown by <code>ntpq -c "apeers"</code>. The module periodically scans for new/changed peers (default: every 60s). <b>ntpd</b> selects the best possible peer from the available peers to synchronize the clock. A minimum of at least 3 peers is required to properly identify the best possible peer.'
+    },
+
+    'filesystem.EXT4_latency': {
+        title: 'EXT4 Latency',
+        info: 'Latency is the time it takes for an event to be completed. We calculate the difference between the calling and return times, we get the logarithmic for the final result and we sum one value to the respective bin. Based on the eBPF <a href="http://www.brendangregg.com/blog/2016-10-06/linux-bcc-ext4dist-ext4slower.html" target="_blank">ext4dist</a> from BCC tools.'
```
ilyam8, line 801: I see that the link _contains_ latency description. In case of ext4 it is

> This "latency" is measured from when the operation was issued from the VFS interface to the file system, to when it is completed. This spans everything: block device I/O (disk I/O), file system CPU cycles, file system locks, run queue latency, etc. 

Do we get the same numbers as `ext4dist` provides?

ilyam8: My point is that _difference between the calling and return times_ sounds like implementation details and not latency description (in the general sense). I could be wrong, not sure 🤷‍♂️

hugovalente-pm: If our goal is to provide more a description of latency I agree with what @ilyam8 mentions and we should write something closer to what is on the link for latency description. If the goal is to provide the link and add to that how we calculate it seem ok.

From my pov, a person who doesn't know much about these tech details, it would be more helpful having something more close to a latency description, but not sure that is the intended goal.

thiagoftsm: > Do we get the same numbers as ext4dist provides?

@ilyam8  we do not show the exact same numbers of `ext4dist`, because we are using `INCREMENTAL` algorithm, but we collect like them.

I will bring more information for our dashboard related to latency in the next commit.

thiagoftsm: @ilyam8  and @hugovalente-pm I did not copy and paste texts, because I think this is not our goal, and probably this could create problems, so I adjusted the text that we are showing on our dashboard, please, take a look when you have a time.

UmanShahzad (approved):

ilyam8 (approved):

hugovalente-pm (approved):

PR #11368, opened 2021-07-26
Allow bundled proto

underhood:
##### Summary
SRE guys need it because they have only CentOS7 available for testing on high power VMs. CentOS 7 unfortunatelly have only protobuf 2.5 bundled


underhood:
Accidental
This was planned to be opened on `stelfrag/netdata` branch `agent-new-architecture`.

PR #11370, opened 2021-07-26
Update configure-alarms.md

samrafalowski:
Updating to use "alerts" term


docs/monitor/configure-alarms.md
```diff
@@ -31,7 +31,7 @@ For example, to edit the `cpu.conf` health configuration file, run:
 sudo ./edit-config health.d/cpu.conf
 ```
 
-Each health configuration file contains one or more health _entities_, which always begin with `alarm:` or `template:`.
+Each health configuration file contains one or more health _entities_, which always begin with `alert:` or `template:`.
```
MrZammler, line 34: Hi Samantha!

On this particular place, we can't change `alarm` to `alert` since that is a keyword present in health configuration files and is required by the agent when it reads configurations. (And if we decide to change the keyword, we'll need to do some work in agent).

Thanks!

ilyam8:
@samrafalowski 

If this change is a part of an Epic please add a link to this epic, so we see the context of _Updating to use "alerts" term_.

(And please sign the CLA.)

PR #11371, opened 2021-07-27
Add installer log

### LINKED ISSUE #9348: Persist installation logs even when the install was successful.
##### Feature idea summary

We currently remove our installation logs after a successful install, even if there were non-fatal errors during the install. This makes it difficult in some cases to figure out what might have gone wrong when optional components didn't get built.

##### Expected behavior

Users are able to view installation logs even for successful installs.

@manos-saratsis

raven428:
##### Summary
Add persistent logging to `kickstart.sh` and `kickstart-static64.sh` fixes #9348:
- Create `/var/log/netdata` if it doesn’t exist.
- Rotate any existing install log as the absolute first thing it does.
- If installing under a prefix, ideally place logs where Netdata would place them by default under that prefix
- Fall back to putting the log in the current working directory if and only if the log directory cannot be accessed and/or cannot be created.
- Rotation should not involve compression (we’re already way too heavy when it comes to installs and updates).
- I would default to saving the logs for the last five installs.

##### Component Name
Install scripts `kickstart.sh` and `kickstart-static64.sh`

##### Test Plan
1. Run `./kickstart.sh` or `./kickstart-static64.sh`.
2. Check out `/var/log/netdata/installer.log` for first and `/opt/netdata/var/log/installer.log` for log records.
3. After running install scripts, again and again, the old log inside directories above should be renamed to `installer.log.1`, `.2`, and so on, limited by `.5` as described in summary.

##### Additional Information
Mentioning @Ferroin as negotiated [here](https://github.com/netdata/netdata/issues/9348#issuecomment-886831404) 😉

### DIFF
--- a/packaging/installer/kickstart-static64.sh
+++ b/packaging/installer/kickstart-static64.sh
@@ -306,7 +306,7 @@ if [ -n "${NETDATA_DISABLE_CLOUD}" ]; then
   fi
 fi
 
-# shellcheck disable=SC2235,SC2030
+# shellcheck disable=SC2235,SC2030,SC2031
 if ( [ -z "${NETDATA_CLAIM_TOKEN}" ] && [ -n "${NETDATA_CLAIM_URL}" ] ) || ( [ -n "${NETDATA_CLAIM_TOKEN}" ] && [ -z "${NETDATA_CLAIM_URL}" ] ); then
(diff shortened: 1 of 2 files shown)


raven428:
Replacing the commits owner.

PR #11372, opened 2021-07-27
Add installer log

### LINKED ISSUE #9348: Persist installation logs even when the install was successful.

raven428:
##### Summary
Add persistent logging to `kickstart.sh`, `kickstart-static64.sh`, and `netdata-installer.sh` fixes #9348:
- Create `/var/log/netdata` if it doesn’t exist.
- Rotate any existing install log as the absolute first thing it does.
- If installing under a prefix, ideally place logs where Netdata would place them by default under that prefix
- Fall back to putting the log in the current working directory if and only if the log directory cannot be accessed and/or cannot be created.
- Rotation should not involve compression (we’re already way too heavy when it comes to installs and updates).
- I would default to saving the logs for the last five installs.

##### Component Name
Install scripts `kickstart.sh`, `kickstart-static64.sh`, and `netdata-installer.sh`.

##### Test Plan
1. Run `./kickstart.sh` or `./kickstart-static64.sh`.
2. Check out `/var/log/netdata/{kickstart,installer}.log` for first and `/opt/netdata/var/log/kickstart-static64.log` for second script log records.
3. After running install scripts, again and again, the old log inside directories above should be renamed to `installer.log.1`, `.2`, and so on, limited by `.5` as described in summary.

##### Additional Information
Mentioning @Ferroin as negotiated [here](https://github.com/netdata/netdata/issues/9348#issuecomment-886831404) 😉

### DIFF
--- a/netdata-installer.sh
+++ b/netdata-installer.sh
@@ -110,6 +110,7 @@ else
   # shellcheck source=packaging/installer/functions.sh
   source "${NETDATA_SOURCE_DIR}/packaging/installer/functions.sh" || exit 1
 fi
+log_create
 
 download_go() {
   download_file "${1}" "${2}" "go.d plugin" "go"
@@ -1588,7 +1589,7 @@ remove_old_ebpf() {
     mv "${NETDATA_PREFIX}/etc/netdata/ebpf_process.conf" "${NETDATA_PREFIX}/etc/netdata/ebpf.d.conf"
   fi
 
-  # Added to remove eBPF programs with name pattern: NAME_VERSION.SUBVERSION.PATCH 
+  # Added to remove eBPF programs with name pattern: NAME_VERSION.SUBVERSION.PATCH
   if [ -f "${NETDATA_PREFIX}/usr/libexec/netdata/plugins.d/pnetdata_ebpf_process.3.10.0.o" ]; then
     echo >&2 "Removing old eBPF programs with patch."
     rm -f "${NETDATA_PREFIX}/usr/libexec/netdata/plugins.d/rnetdata_ebpf"*.?.*.*.o
--- a/packaging/installer/functions.sh
+++ b/packaging/installer/functions.sh
@@ -88,8 +88,49 @@ setup_terminal() {
 }
 setup_terminal || echo > /dev/null
 
+logpath_suffix='var/log/netdata/installer.log'
+
+# shellcheck disable=SC2031
+state=$(set +o)
+set +u
+if [ -z "${NETDATA_PREFIX}" ] ; then
+  logpath="/${logpath_suffix}"
+else
+  logpath="${NETDATA_PREFIX}/${logpath_suffix}"
+fi
+eval "${state}"
+
+log_create() {
+  logdir="$(dirname "${logpath}")"
+  if [ ! -d "${logdir}" ] ; then
+    echo >&2 "Creating [${logdir}] for installer log file:"
+    mkdir -vp "$(dirname "${logpath}")" >&2
+  fi
+  if ! touch "${logpath}" ; then
+    echo >&2 "Unable to create [$logpath], falling back to current directory"
+    logpath="${PWD}/$(basename "${logpath}")"
+  fi
+  if [ -e "${logpath}" ] ; then
+    echo >&2 "Rotating [${logpath}] log file:"
+    i=4
+    while [ ${i} -gt 0 ] ; do
+      if [ -e "${logpath}.${i}" ] ; then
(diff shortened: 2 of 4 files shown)


Ferroin:
From a cursory look, the changes here are good (code is well written, etc), there are only three issues I see:

* There are a couple of places where `bash`-specific syntax has been added, but both of the scripts here are run using `/bin/sh`. The shellcheck CI job has flagged these with warnings, so you can check there to see the exact places.
* The check for a custom prefix needs to look at `${NETDATA_INSTALL_PREFIX}`, not just `${ndprefix}`. `${ndprefix}` only gets set if we detect an _existing_ install and need to decide whether to update it, reinstall over top of it, or bail out. In the case of a new install in a custom prefix, this will never get set, but `${NETDATA_INSTALL_PREFIX}` will instead be set to the equivalent value (specifically, set by the `--install` option on the command line). Also, handling for this shouldn’t be needed for the `kickstart-static64.sh` case, as that is only used for static installs which can only ever be under `/opt/netdata`.
* A large part of the actual installation process for the regular `kickstart.sh` case is actually handled by the `netdata-installer.sh` script found in the root of the repository. This generates it’s own logs already, but we ideally need to copy the contents of those to the logs being generated by the `kickstart.sh` script.

On a side note, I recommend ignoring the ‘Checks / Checksums’ CI error until we’ve got everything else finalized. It’s much easier to just update the checksums in the documentation after you’ve got everything else working than to have to update them each time you make a change (and, for the record, I won’t hold the fact that you did not know about this check against you, it’s non-obvious that we have it, and I actually regularly forget about it myself until I actually open a PR).

raven428:
As [asked](https://github.com/netdata/netdata/pull/11372#issuecomment-888273979), I performed the next changes:
* I rewrote bash-specific code by 5002679, but not sure it fixes all of the added syntaxes. Still didn't have a chance yet to dig into shellcheck CI job and am not sure, because my VSCode shellcheck extension may not detect some issues.
* Probably, you mean `${NETDATA_PREFIX}`. I added in 346f3f1 considering this variable from command-line option `--install` for choosing the log location.
* I changed the log name for both kickstart scripts to `kickstart.log` (`kickstart-static64.log` for `kickstart-static64.sh`) and added logging to `netdata-installer.sh` by 96029c9. Also, I tried to find a way to make joined `kickstart.log` with new-generated `installer.log`, but have yet not forged a good solution.

Ferroin:
Looks like there’s a build failure in the static build check related to this (see https://github.com/netdata/netdata/pull/11372/checks?check_run_id=3280903212). You can test that locally by running `packaging/makeself/build-x86_64-static.sh` from the root of a clean checkout of the Netdata repository on a system with a working Docker install.

The eslint, CentOS 8 package build, and Fedora Kinesis build can be safely ignored, all three are known broken right now.

The checksums thing will need to be addressed, but as mentioned before that should wait until everything else is good.

raven428:
Thank you for your follow-up, I also already noticed it a couple of days ago. I'll check and will fix it tommorow.

raven428:
Fixed as promised. Now writing to log file be skipped if the logging directory is not created by `log_create` as usually happens at test builds. `log_create` is explicitly called only by an installer.

raven428:
Looks like some jobs are still queued. Also, failed jobs are for external reasons. @Ferroin could you please check this?

Ferroin:
> Looks like some jobs are still queued. Also, failed jobs are for external reasons. @Ferroin could you please check this?

The Fedora AWS Kinesis build can be ignored, as can the ClearLinux stuff (or alternatively, if you rebase your branch to the latest `master`, we’ve actually removed the ClearLinux checks).

The builds being stuck queued are something we’ve been having intermitent issues with (seems to be a bug in GitHub’s infra from what I can tell). I’ve queued a rerun of the affected workflows which _should_ resolve this.

As far as the static build check, that still seems to be an issue with the changes here as it’s not failing in the `master` branch right now.

Ferroin:
@raven428 Any chance you could rebase this onto the latest master? I suspect the CI failures we were seeing will simply evaporate, at which point it should be relatively quick and easy to finally get this merged.

ilyam8:
Hey, @raven428. Do you plan to continue work on this PR? If yes - there are a lot of merge conflicts that need to be resolved.

PR #11383, opened 2021-07-30
add Travis ctrl file for checking if changes happened

### LINKED ISSUE #10298: Travis: Don't trigger nightly update if no new commits have been made
It seems like you guys are using a cron job with travis to trigger a new nightly version daily _even if there are no new changes_ (as seen in https://github.com/netdata/netdata/commit/c9fea34095b2035e43c8e4c05d89ee272640b523, for reference [this](https://travis-ci.com/github/netdata/netdata/jobs/449994456#L13456) should be the travis "build").

I suggest adding some kind of check to if make sure it only runs if necessary to prevent unnecessary updates for the consumer.

rgiovanardi:
##### Summary
Fixes #10298
`nightlies.sh` already checks if changes happened since last nightly. When changes happens, it write a new `current_build_status` file and the subsequent `.travis/generate_changelog_for_nightlies.sh` script commits it and push into repository.

On the next Travis stage, "Nightly Release", the job "Create nightly release artifacts, publish to GCS" will check the previous control file. If the `TRAVIS_BUILD_NUMBER` differs from the one written in the file, it means that `nightlies.sh` previously saw that no changes happened since last nightly release, hence no new release is needed and it exits.

##### Component Name
Packaging

##### Test Plan
1. Commit a new release and ensure Travis CI builds and push new releases to GCS.  
2. Without any new commit, wait the scheduled nightly Travis CI builds or run a Travis build manually and ensure no new releases are pushed.

##### Additional Information
The two trigger jobs of "Nightly Release" stage will run anyway: "Trigger Docker image build and publish" and "Trigger DEB and RPM package build".

### DIFF
--- a/.travis/create_artifacts.sh
+++ b/.travis/create_artifacts.sh
@@ -38,7 +38,7 @@ if [ -f .travis/current_build_status ]; then
 	  exit 0
   fi
 else
-  echo "File .travis/current_build_status, probably this is the very first build, let's continue"
+  echo "File .travis/current_build_status doesn't exist, probably this is the very first build, let's continue"
 fi
 
 echo "--- Initialize git configuration ---"
--- a/.travis/create_artifacts.sh
+++ b/.travis/create_artifacts.sh
@@ -26,6 +26,21 @@ if [ ! "${TRAVIS_REPO_SLUG}" == "netdata/netdata" ]; then
   exit 0
 fi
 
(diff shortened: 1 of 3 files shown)


Ferroin (approved):
We don’t really have any good way to test this without just merging it, but it looks to me like it should work correctly, and I see no obvious issues.

PR #11392, opened 2021-08-03
Update perf.events and add new charts

thiagoftsm:
##### Summary
This PR is bringing [Instruction per Cycle](https://en.wikipedia.org/wiki/Instructions_per_cycle) for Netdata using `perf.plugin`. 

##### Component Name
perf.plugin
##### Test Plan

1 - Set your `/etc/netdata/netdata.conf` as :

```conf
[plugins]
    perf = yes

[plugin:perf]
        update every = 1
        command options = cycles instructions emulation L1I
```

2 - Compile this Branch and take a look at  `Perf counters` section.
3 - Access `http://localhost:1999/api/v1/info` and verify that we have the following modules set:

```
		{
			"plugin": "perf.plugin",
			"module": "hardware"
		},
                {
			"plugin": "perf.plugin",
			"module": "cache"
		},
                {
			"plugin": "perf.plugin",
			"module": "software"
		},
```


collectors/perf.plugin/perf_plugin.c
```diff
@@ -527,6 +529,38 @@ static void perf_send_metrics() {
...
+
+            printf("CHART %s.%s '' '%s' 'instructions/cycle' %s '' line %d %d %s\n"
+                   , RRD_TYPE_PERF
+                   , "instructions_per_cycle"
+                   , "Instructions per Cycle(IPC). An IPC < 1.0 likely means memory bound, and an IPC > 1.0 likely means instruction bound. "
+                   , RRD_FAMILY_HW
+                   , NETDATA_CHART_PRIO_PERF_IPC
+                   , update_every
+                   , PLUGIN_PERF_NAME
+            );
```
ilyam8, line 544: It is `[update_every [options [plugin [module]]]`. It looks like you are missing options (`''`) before "plugin".

thiagoftsm: You are right, we are missing this not only for this chart, but for the whole plugin. 
I think this is subject for another PR, because I copied and changed from other chart that the plugin plots.

ilyam8: I don't get it. Why do not fix it now if you agree?

ilyam8: > An IPC < 1.0 likely means memory bound, and an IPC > 1.0

What about 1.0 btw? 😄

ilyam8 (dismissed):

collectors/perf.plugin/perf_plugin.c
thiagoftsm: >  I don't get it. Why do not fix it now if you agree?

I will rename the PR name, update the description and update it here  :smile: .


> What about 1.0 btw? smile

Good point, this was not defined in the [source](https://www.brendangregg.com/blog/2017-05-09/cpu-utilization-is-wrong.html). I will add it here for users get more details.

ilyam8: > I will rename the PR name, update the description and update it here

Ok, let's do not do it. Perhaps I couldn't word my requests properly, I don't think it requires renaming. Let's do that change in another PR.

thiagoftsm: @ilyam8 as I said to you, I brought more commits and I updated the `title`, I added the missing `options` and finally I added the missed `module`. I will update the issue with that has missing `modules` for charts.

ilyam8:
I don't think `perf.plugin` has modules. It collects different types of events (hardware, software, hardware cache), yes, but it doesn't make them automatically modules. I didn't request those changes, sorry if I wasn't clear 😅 That is only my opinion, I am not saying revert them.

thiagoftsm:
> I don't think `perf.plugin` has modules. It collects different types of events (hardware, software, hardware cache), yes, but it doesn't make them automatically modules. I didn't request those changes, sorry if I wasn't clear sweat_smile That is only my opinion, I am not saying revert them.

I understand your point of view, on the other hand I think we can group them as `module` :thinking: , I applied here the same logic that I used to define `eBPF.plugin` modules. Let us wait for @vlvkobal opinion.

vlvkobal:
I agree with @ilyam8 there are different types of events in the plugin, not modules.

underhood:
changed PR name as on first look on PR list I thought it mean Inter-process Communication

thiagoftsm:
> I agree with @ilyam8, there are different types of events in the plugin, not modules.

All right, I will remove the module with next commit.

collectors/perf.plugin/perf_plugin.c
```diff
@@ -527,11 +529,43 @@ static void perf_send_metrics() {
...
     // ------------------------------------------------------------------------
 
+    if(likely(perf_events[EV_ID_INSTRUCTIONS].updated) && likely(perf_events[EV_ID_CPU_CYCLES].updated)) {
+        if(unlikely(!ipc_chart_generated)) {
+            ipc_chart_generated = 1;
+
+            printf("CHART %s.%s '' '%s' 'instructions/cycle' %s '' line %d %d '' %s\n"
+                   , RRD_TYPE_PERF
+                   , "instructions_per_cycle"
+                   , "Instructions per Cycle(IPC). An IPC < 1.0 likely means memory bound, and an IPC > 1.0 likely means instruction bound. For more details about the metric take a lookt at this <a href=\"https://www.brendangregg.com/blog/2017-05-09/cpu-utilization-is-wrong.html\" target=\"_blank\">blog post</a>."
```
ilyam8, line 539: Usually, we use dashboard_info.js for that. Let's move everything after `Instructions per Cycle(IPC)` to it.

thiagoftsm: I push another commit moving content for dashboard_info.js.

web/gui/dashboard_info.js
```diff
@@ -3467,6 +3467,13 @@ netdataDashboard.context = {
         info: 'Size of metric samples written to disk.'
     },
 
+    // ------------------------------------------------------------------------
+    // Perf
+
+    'perf.instructions_per_cycle': {
+        info: 'An IPC < 1.0 likely means memory bound, and an IPC > 1.0 likely means instruction bound. For more details about the metric take a lookt at this <a href="https://www.brendangregg.com/blog/2017-05-09/cpu-utilization-is-wrong.html" target="_blank">blog post</a>.'
```
ilyam8, line 3474:
```suggestion
        info: 'An IPC < 1.0 likely means memory bound, and an IPC > 1.0 likely means instruction bound. For more details about the metric take a look at this <a href="https://www.brendangregg.com/blog/2017-05-09/cpu-utilization-is-wrong.html" target="_blank">blog post</a>.'
```

thiagoftsm: Typo fixed with last commit.

ilyam8 (approved):

vlvkobal (approved):

PR #11399, opened 2021-08-04
Use sqlite to store the health log and alert configurations.

MrZammler:
##### Summary

This PR uses sqlite to store the health log instead of files. It also stores the alert configurations to sqlite. Both these changes are required to support the new architecture.

The basic functionality is that on startup, if the health-log.db files have not been migrated to sqlite, it will do so then, and from there on it will load and save health log items to sqlite. A different table for each of the parent and/or children will be used (it will contain the machine guid of each on the table name).

It will also read and store the alert configurations as found on the disk to sqlite. For every alert or template configuration, it will create a unique config hash id, which will then be passed to rrdcalc and rrdcalctemplate structures (and from there on to alert events too).

The PR description will be updated with more information, as it progresses.

##### Component Name

Health/Database

##### Test Plan

* For the config hash:

Check that the url `http://192.168.0.1:19999/api/v1/alert_config` has the whole configuration as in files inside `health/health.d`. The url will contain also the `lookup` field, broken into tokens (needed to be pushed to cloud).

Check that editing an alert configuration produces a new hash entry.

* Health log

TBA

Since it makes use of sqlite, an sqlite client like https://sqlitestudio.pl can be handy for checks.

##### Additional Information

* By using sqlite for the health log, `health_label_log_save` is not used any more.
* Most of the health code remains the same, in later steps we could refactor for e.g. to keep less alert items in memory, skip some code that goes through the structure, etc. Also in case a user downgrades, the health-log.db files are not deleted atm.
* There is room to improve in cleanup. We can probably keep more than 2000 lines which is currently the default. When new architecture is introduced, it will also affect the cleanup policy.


database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1367 @@
...
+    uuid_t hash_id,
+    struct alert_config *cfg)
+{
+    EVP_MD_CTX *evpctx;
+    unsigned char hash_value[EVP_MAX_MD_SIZE];
+    unsigned int hash_len;
+    evpctx = EVP_MD_CTX_create();
+    EVP_DigestInit_ex(evpctx, EVP_sha256(), NULL);
+
+    error("HASH asked to hash [%s]", cfg->alarm ? cfg->alarm : cfg->template_key);
```
stelfrag, line 1297: We should probably remove this or report with a suitable debug flag

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1367 @@
...
+    snprintfz(command, MAX_HEALTH_SQL_SIZE, SQL_COUNT_HEALTH_LOG(uuid_str));
+
+    rc = sqlite3_prepare_v2(db_meta, command, -1, &res, 0);
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("Failed to prepare statement to count health log entries from db");
+        return;
+    }
+    rc = sqlite3_step(res);
+
+    host->health_log_entries_written = sqlite3_column_int(res, 0);
```
stelfrag, line 651: Need to check `rc` for error

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1367 @@
...
+                // find the latest last_repeat timestamp. Otherwise,
+                // there is no need to keep them in memory.
+                continue;
+            }
+        }
+
+        freez(alarm_name);
+
+        // check for a possible host mismatch
+        if(strcmp((char *) sqlite3_column_text(res, 0), host->hostname)) {
```
stelfrag, line 759: Should this be done earlier?  Maybe line 728? To avoid code in L734-L744

database/sqlite/sqlite_health.c
MrZammler: Ah, removed, was left there for some tests....

database/sqlite/sqlite_health.c
MrZammler: Ok, thanks, added check in https://github.com/netdata/netdata/pull/11399/commits/050fd0defdccc233c45916afedfd43e2d72b69f1

database/rrdhost.c
```diff
@@ -313,6 +310,22 @@ RRDHOST *rrdhost_create(const char *hostname,
...
+        if (host->health_enabled) {
+            if (!file_is_migrated(host->health_log_filename)) {
+                int rc = sql_create_health_log_table(host);
+                if (unlikely(rc)) {
+                    error_report("Failed to create health log table in the database");
+                    health_alarm_log_load(host);
+                    health_alarm_log_open(host);
+                }
+                else {
+                    health_alarm_log_load(host);
```
MrZammler, line 322: Do you think we should add a check here if the table really exists, and if not re-create it, if it has been deleted by the user? Now, if the user has deleted the health log table, the only re-create will happen if he deletes also the entry in metadata_migration

database/rrdhost.c
```diff
@@ -313,6 +310,22 @@ RRDHOST *rrdhost_create(const char *hostname,
...
+                    add_migrated_file(host->health_log_filename, 0);
+                }
+            } else {
+                sql_health_alarm_log_load(host);
```
MrZammler, line 326: Do you think we should add a check here if the table really exists, and if not re-create it, if it has been deleted by the user? Now, if the user has deleted the health log table, the only re-create will happen if he deletes also the entry in metadata_migration

database/sqlite/sqlite_health.c
MrZammler: I don't know if it's even needed at all.... it is also disabled in old health log code... I will remove....

MrZammler: Removed in https://github.com/netdata/netdata/pull/11399/commits/9aa506af892add8c4d87681abc9040e0ee6f3413

database/rrdhost.c
MrZammler: Ok, added in https://github.com/netdata/netdata/pull/11399/commits/97ed0ef80e4009f9f62b33bdeb294c51ff8198e0

Doesn't check return, not sure if it makes sense at that point....

database/sqlite/sqlite_functions.c
```diff
@@ -20,6 +20,12 @@ const char *database_config[] = {
     "CREATE TABLE IF NOT EXISTS chart_label(chart_id blob, source_type int, label_key text, "
     "label_value text, date_created int, PRIMARY KEY (chart_id, label_key));",
     "CREATE TABLE IF NOT EXISTS node_instance (host_id blob PRIMARY KEY, claim_id, node_id, date_created);",
+    "CREATE TABLE IF NOT EXISTS alert_hash(hash_id blob PRIMARY KEY, alarm text, template text, "
```
thiagoftsm, line 23: This table has only one index, and the rest of the code is not creating indexes for it, on the other hand fields like `plugin`, `module`, and `alarm` are fields we could o queries to show some report from our endpoints, why are we not creating indexes for other fields?

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    }
+
+    char uuid_str[GUID_LEN + 1];
+    uuid_unparse_lower(host->host_uuid, uuid_str);
+    uuid_str[8] = '_';
+    uuid_str[13] = '_';
+    uuid_str[18] = '_';
+    uuid_str[23] = '_';
+
+    sprintf(command, SQL_UPDATE_HEALTH_LOG(uuid_str));
```
stelfrag, line 67: Need to pass the max size of command (snprintf or something like that)

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    char uuid_str[GUID_LEN + 1];
+    uuid_unparse_lower(host->host_uuid, uuid_str);
+    uuid_str[8] = '_';
+    uuid_str[13] = '_';
+    uuid_str[18] = '_';
+    uuid_str[23] = '_';
+
+    snprintfz(command, MAX_HEALTH_SQL_SIZE, SQL_CREATE_HEALTH_LOG_TABLE(uuid_str));
+
+    if (unlikely(!db_meta)) {
```
stelfrag, line 25: if the test needs to exist it is better to be at the very top

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    }
+
+    guid = strdupz(host->machine_guid);
+
+    guid[8]='_';
+    guid[13]='_';
+    guid[18]='_';
+    guid[23]='_';
+
+    sprintf(command, SQL_SELECT_HEALTH_LOG(guid));
```
stelfrag, line 377: snprintf?

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    }
+
+    guid = strdupz(host->machine_guid);
+
+    guid[8]='_';
+    guid[13]='_';
+    guid[18]='_';
+    guid[23]='_';
+
+    sprintf(command, SQL_SELECT_ALL_HEALTH_LOG(guid, host->health_log.max));
```
stelfrag, line 552: snprintf

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    char uuid_str[GUID_LEN + 1];
+    uuid_unparse_lower(host->host_uuid, uuid_str);
+    uuid_str[8] = '_';
+    uuid_str[13] = '_';
+    uuid_str[18] = '_';
+    uuid_str[23] = '_';
+
+    snprintfz(command, MAX_HEALTH_SQL_SIZE, SQL_CLEANUP_HEALTH_LOG(uuid_str, uuid_str, host->health_log_entries_written - rotate_every));
+
+    rc = sqlite3_prepare_v2(db_meta, command, -1, &res, 0);
```
stelfrag, line 606: There is no finalize for this statement

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("HEALTH [%s]: Failed to prepare sql statement to load health log.", host->hostname);
+        return;
+    }
+
+    netdata_rwlock_rdlock(&host->health_log.alarm_log_rwlock);
+
+    while (sqlite3_step(res) == SQLITE_ROW) {
+        errno = 0;
```
stelfrag, line 708: why set errno?

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+        uint32_t alarm_id = (uint32_t) sqlite3_column_int64(res, 2);
+        if(!alarm_id) {
+            error_report("HEALTH [%s]: Got invalid alarm id. Ignoring it.", host->hostname);
+            errored++;
+            continue;
+        }
+
+        // Check if we got last_repeat field
+        time_t last_repeat = 0;
+        char* alarm_name = strdupz((char *) sqlite3_column_text(res, 14));
```
stelfrag, line 730: need to allocate memory? (and free it in L755?)

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+            }
+        }
+
+        freez(alarm_name);
+
+        ae = callocz(1, sizeof(ALARM_ENTRY));
+
+        ae->unique_id = unique_id;
+
+        if (!is_valid_alarm_id(host, (const char *) sqlite3_column_text(res, 14), (const char *) sqlite3_column_text(res, 13), alarm_id))
```
stelfrag, line 761: We may not need this check at all

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+            alarm_id = rrdcalc_get_unique_id(host, (const char *) sqlite3_column_text(res, 14), (const char *) sqlite3_column_text(res, 13), NULL);
+        ae->alarm_id = alarm_id;
+
+        uuid_copy(ae->config_hash_id, *((uuid_t *) sqlite3_column_blob(res, 4)));
```
stelfrag, line 765: Should probably check for corrupted entry that will have null in that field

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    uuid_str[13] = '_';
+    uuid_str[18] = '_';
+    uuid_str[23] = '_';
+
+    sprintf(command, SQL_UPDATE_HEALTH_LOG(uuid_str));
+
+    rc = sqlite3_prepare_v2(db_meta, command, -1, &res, 0);
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("HEALTH [%s]: Failed to prepare statement for SQL_UPDATE_HEALTH_LOG", host->hostname);
+        goto failed;
```
stelfrag, line 72: This must not go to finilize if the prepare has failed

database/rrdcalc.c
```diff
@@ -398,6 +400,7 @@ inline RRDCALC *rrdcalc_create_from_template(RRDHOST *host, RRDCALCTEMPLATE *rt,
     rc->hash = simple_hash(rc->name);
     rc->chart = strdupz(chart);
     rc->hash_chart = simple_hash(rc->chart);
+    uuid_copy(rc->config_hash_id, *((uuid_t *) &rt->config_hash_id));
```
stelfrag, line 403: Wouldn't just `uuid_copy(rc->config_hash_id,  rt->config_hash_id)`  work?

health/health_config.c
```diff
@@ -503,6 +503,68 @@ static inline void strip_quotes(char *s) {
     }
 }
 
+static inline void alert_config_free(struct alert_config *cfg)
+{
+    freez(cfg->alarm);
+    cfg->alarm = NULL;
```
stelfrag, line 509: Why do you need to set everything to NULL?

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    guid[13]='_';
+    guid[18]='_';
+    guid[23]='_';
+
+    sprintf(command, SQL_SELECT_HEALTH_LOG(guid));
+
+    rc = sqlite3_prepare_v2(db_meta, command, -1, &res, 0);
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("Failed to prepare statement SQL_SELECT_HEALTH_LOG");
+        return;
```
stelfrag, line 382: guid memory leak

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+    guid[13]='_';
+    guid[18]='_';
+    guid[23]='_';
+
+    sprintf(command, SQL_SELECT_ALL_HEALTH_LOG(guid, host->health_log.max));
+
+    rc = sqlite3_prepare_v2(db_meta, command, -1, &res, 0);
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("Failed to prepare statement for SQL_SELECT_ALL_HEALTH_LOG");
+        return;
```
stelfrag, line 557: guid memory leak

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+        buffer_strcat(wb, "\t\t\"old_value\":");
+        if (sqlite3_column_type(res, 26) == SQLITE_NULL)
+            buffer_strcat(wb, "null");
+        else
+            buffer_rrd_value(wb, (calculated_number) sqlite3_column_double(res, 26));
+        buffer_strcat(wb, "\n");
+
+        buffer_strcat(wb, "\t}");
+
+        freez(replaced_info);
```
stelfrag, line 524: edit_command memory leak?

health/health_config.c
```diff
@@ -652,25 +717,30 @@ static int health_readfile(const char *filename, void *data) {
             rc->old_status = RRDCALC_STATUS_UNINITIALIZED;
             rc->warn_repeat_every = host->health_default_warn_repeat_every;
             rc->crit_repeat_every = host->health_default_crit_repeat_every;
+            alert_cfg = callocz(1, sizeof(struct alert_config));
```
stelfrag, line 720: memory leak if alert_cfg was allocated already

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1357 @@
...
+#define SQL_SELECT_ALERT_CONFIG_WITH_HASH "select hash_id, alarm, template, on_key, class, component, type, os, hosts, lookup, every, units, calc, families, plugin, module, green, red, warn, crit, exec, to_key, info, delay, options, repeat, host_labels, p_db_lookup_dimensions, p_db_lookup_method, p_db_lookup_options, p_db_lookup_after, p_db_lookup_before, p_update_every from alert_hash where hash_id = @hash_id;"
+#define SQL_SELECT_ALERT_CONFIG "select hash_id, alarm, template, on_key, class, component, type, os, hosts, lookup, every, units, calc, families, plugin, module, green, red, warn, crit, exec, to_key, info, delay, options, repeat, host_labels, p_db_lookup_dimensions, p_db_lookup_method, p_db_lookup_options, p_db_lookup_after, p_db_lookup_before, p_update_every from alert_hash;"
+void sql_select_alert_config(char *hash_str, BUFFER *wb)
+{
+    int rc;
+    sqlite3_stmt *res_alert = NULL;
+
+    uuid_t hash_id;
+    if (hash_str)
+        uuid_parse(hash_str, hash_id);
```
stelfrag, line 1126: hash_str may not be a valid uuid so uuid_parse will fail in this case

database/sqlite/sqlite_health.c
MrZammler: Ok, thanks! Done.

database/sqlite/sqlite_health.c
MrZammler: Ok, thanks! Done, moved.

database/sqlite/sqlite_health.c
MrZammler: Ok, thanks! Done.

database/sqlite/sqlite_health.c
MrZammler: Ok, thanks! Done.

database/sqlite/sqlite_health.c
MrZammler: Thanks! Added.

database/sqlite/sqlite_health.c
MrZammler: Removed, thanks!

database/sqlite/sqlite_health.c
MrZammler: Will use the the sqlite_column_text by itself, thanks!

database/sqlite/sqlite_health.c
MrZammler: Ok, added a check before this if the field is null.

database/sqlite/sqlite_health.c
MrZammler: Ok, thanks, it will return instead.

database/sqlite/sqlite_health.c
MrZammler: Thanks, replaced with static variable as with other to keep similar.

database/sqlite/sqlite_health.c
MrZammler: Thanks, replaced with static variable as with other to keep similar.

database/sqlite/sqlite_health.c
MrZammler: Thanks. Added free. Also removed replaced_info from there, not needed.

health/health_config.c
MrZammler: Thanks, not needed, removed.

database/rrdcalc.c
MrZammler: Yes, it does, changed, thanks.

database/sqlite/sqlite_health.c
MrZammler: Ok, added a check, thanks!

health/health_config.c
MrZammler: Ok, added free of `cfg` inside `alert_config_free`

database/sqlite/sqlite_functions.c
stelfrag: @thiagoftsm you are correct -- we may need this in the future but for now it is ok.

underhood:
@MrZammler can we resolve the merge conflicts here?

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1347 @@
...
+        goto failed;
+    }
+
+    rc = sqlite3_bind_int64(res, 4, (sqlite3_int64) ae->alarm_event_id);
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("Failed to bind alarm_event_id parameter for SQL_INSERT_HEALTH_LOG");
+        goto failed;
+    }
+
+    rc = sqlite3_bind_blob(res, 5, ae->config_hash_id, sizeof(ae->config_hash_id), SQLITE_STATIC);
```
stelfrag, line 171: In most cases we'd use & here `&ae->config_hash_id`

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1347 @@
...
+}
+
+
+#define SQL_SELECT_HEALTH_LOG(guid) "SELECT hostname, unique_id, alarm_id, alarm_event_id, config_hash_id, updated_by_id, updates_id, when_key, duration, non_clear_duration, flags, exec_run_timestamp, delay_up_to_timestamp, name, chart, family, exec, recipient, source, units, info, exec_code, new_status, old_status, delay, new_value, old_value, last_repeat, class, component, type FROM health_log_%s where unique_id = ? and alarm_id = ?;", guid
+void health_alarm_entry_sql2json(BUFFER *wb, uint32_t unique_id, uint32_t alarm_id, RRDHOST *host) {
+    sqlite3_stmt *res = NULL;
+    int rc;
+    char command[MAX_HEALTH_SQL_SIZE + 1];
+
+    if (unlikely(!db_meta)) {
```
stelfrag, line 364: The caller has already checked, so you probably do not need this

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1347 @@
...
+    rc = sqlite3_prepare_v2(db_meta, command, -1, &res, 0);
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("Failed to prepare statement for SQL_SELECT_ALL_HEALTH_LOG");
+        return;
+    }
+
+    int count=0;
+    while (sqlite3_step(res) == SQLITE_ROW) {
+        if (count) buffer_strcat(wb, ",");
+        health_alarm_entry_sql2json(wb, (uint32_t) sqlite3_column_int64(res, 0), (uint32_t) sqlite3_column_int64(res, 1), host);
```
stelfrag, line 546: Its probably better to prepare the statement before calling the function and pass it as a parameter and reset the parameters as needed

database/sqlite/sqlite_health.c
MrZammler: Ok, fixed, thanks!

database/sqlite/sqlite_health.c
MrZammler: Actually, removed both `health_alarm_entry_sql2json` and `sql_health_alarm_log_select_all`, where not used. If required in later point will be re-added.

database/sqlite/sqlite_health.c
MrZammler: Actually, removed both `health_alarm_entry_sql2json` and `sql_health_alarm_log_select_all`, where not used. If required in later point will be re-added.

database/sqlite/sqlite_health.c
MrZammler: Yes, removed.

health/health_json.c
```diff
@@ -187,9 +191,13 @@ static inline void health_rrdcalc2json_nolock(RRDHOST *host, BUFFER *wb, RRDCALC
         }
     }
 
+    char hash_id[37];
```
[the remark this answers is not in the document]
MrZammler: Correct, thanks! Replaced with `GUID_LEN + 1` (GUID_LEN = 36).

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1130 @@
...
+    if (cfg->repeat)
+        EVP_DigestUpdate(evpctx, cfg->repeat, strlen(cfg->repeat));
+    if (cfg->host_labels)
+        EVP_DigestUpdate(evpctx, cfg->host_labels, strlen(cfg->host_labels));
+
+    EVP_DigestFinal_ex(evpctx, hash_value, &hash_len);
+    EVP_MD_CTX_destroy(evpctx);
+    fatal_assert(hash_len > sizeof(uuid_t));
+
+    char uuid_str[36 + 1];
```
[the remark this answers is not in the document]
MrZammler: Correct, thanks! Replaced with `GUID_LEN + 1` (GUID_LEN = 36).

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1130 @@
...
+        /* char id[512]; */
+        /* sprintf(id, "%s.%s", sqlite3_column_text(res_alert, 3), sqlite3_column_text(res_alert, 1)); */
+
+        if (c)
+            buffer_strcat(wb, ",\t\t\n");
+        else
+            buffer_strcat(wb, "\t\n");
+        c++;
+
+        char uuid_str[36 + 1];
```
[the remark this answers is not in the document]
MrZammler: Correct, thanks! Replaced with `GUID_LEN + 1` (GUID_LEN = 36).

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,1130 @@
...
+
+    snprintfz(command, MAX_HEALTH_SQL_SIZE, SQL_LOAD_HEALTH_LOG(uuid_str, host->health_log.max));
+
+    rc = sqlite3_prepare_v2(db_meta, command, -1, &res, 0);
+    if (unlikely(rc != SQLITE_OK)) {
+        error_report("HEALTH [%s]: Failed to prepare sql statement to load health log.", host->hostname);
+        return;
+    }
+
+    netdata_rwlock_rdlock(&host->health_log.alarm_log_rwlock);
```
vkalintiris, line 482: Should this be a write lock instead of a read lock?

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,958 @@
...
+    unsigned int hash_len;
+    evpctx = EVP_MD_CTX_create();
+    EVP_DigestInit_ex(evpctx, EVP_sha256(), NULL);
+
+    if (cfg->alarm) {
+        EVP_DigestUpdate(evpctx, cfg->alarm, strlen(cfg->alarm));
+    } else if (cfg->template_key)
+        EVP_DigestUpdate(evpctx, cfg->template_key, strlen(cfg->template_key));
+
+    if (cfg->os)
```
underhood, line 895: I wonder if we should add "NULL" in case it is NULL... Lets consider 2 cases:
1. `cfg->os == "tralalala"  && cfg->host == NULL`
2. `cfg->os == NULL && cfg->host == "tralalala"` 
I think current code will result in same hash for both cases (dangerous)?

underhood: I think the `else` case should be something like `EVP_DigestUpdate(evpctx, `\0`, 1);`

MrZammler: Good point, thanks Timo!

It will fail, for example in this scenario of two "blocks" of configs:

```
 template: fping_host_latency
 families: *
       on: fping.latency
    class: Latency
     type: Other
component: Network
   lookup: average -10s unaligned of average
    units: ms
    every: 10s
    green: 500
     warn: $this > $green OR $max > $red
     crit: $this > $red
    delay: down 30m multiplier 1.5 max 2h
     info: average latency to the network host over the last 10 seconds
       to: sysadmin

 template: fping_host_latency
 families: *
       on: fping.latency
    class: Latency
     type: Other
component: Network
   lookup: average -10s unaligned of average
    units: ms
    every: 10s
      red: 500
     warn: $this > $green OR $max > $red
     crit: $this > $red
    delay: down 30m multiplier 1.5 max 2h
     info: average latency to the network host over the last 10 seconds
       to: sysadmin
```

Taking advantage of the ordering of calling, we need to hash something if the value is null. I tried `EVP_DigestUpdate(evpctx, '\0', 1);` and `EVP_DigestUpdate(evpctx, '\0', 0);` but won't work, so I went with hashing the word `NULL` instead in https://github.com/netdata/netdata/pull/11399/commits/8b0257421c90d5237b3af68955d9df55d78b500e

We might be able to use something else instead of `NULL`, or a different approach ?

database/sqlite/sqlite_health.c
```diff
@@ -0,0 +1,931 @@
...
+    EVP_MD_CTX *evpctx;
+    unsigned char hash_value[EVP_MAX_MD_SIZE];
+    unsigned int hash_len;
+    evpctx = EVP_MD_CTX_create();
+    EVP_DigestInit_ex(evpctx, EVP_sha256(), NULL);
+
+    cfg->alarm ? EVP_DigestUpdate(evpctx, cfg->alarm, strlen(cfg->alarm)) : EVP_DigestUpdate(evpctx, "NULL", 4);
+    cfg->template_key ? EVP_DigestUpdate(evpctx, cfg->template_key, strlen(cfg->template_key)) : EVP_DigestUpdate(evpctx, "NULL", 4);
+
+    cfg->os ? EVP_DigestUpdate(evpctx, cfg->os, strlen(cfg->os)) : EVP_DigestUpdate(evpctx, "NULL", 4);
```
underhood, line 893: don't want to be nit picking here but would still prefer to add single byte that is 0 here instead of `NULL` to not create clash in case the pointer != NULL and case where there is a string that has value "NULL".

Additionally for readability I would macroize this or do it as static function

underhood: Meaning if I make my hostname "NULL" the hash should be different from case where there is no hostname or we couldnt read it or dont know it (pointer is NULL)

underhood: e.g.:
```
uint_8t x = 0
EVP_DigestUpdate(evpctx, &x, 1);
```

database/sqlite/sqlite_health.c
underhood: see my comments below (used IDE to comment on new code so it started new thread)

underhood: EVP_DigestUpdate(evpctx, '\0', 0); yeah that wont work because we need pointer to 0 byte instead of 0 byte itself. See my example below or maybe `EVP_DigestUpdate(evpctx, "", 1);` should work? (empty string is just 0 byte)

MrZammler: Maybe we can use a control character? e.g. `EVP_DigestUpdate(evpctx, "\n", 1);`  ... ?

underhood: I strongly prefer 0 byte for following reasons:
- the hashing function itself can hash any data (it doesn't care if it is string or whatever)
- by using `EVP_DigestUpdate(evpctx, cfg->template_key, strlen(cfg->template_key))` you are hashing the strings without trailing 0 byte. By definition then the valid hashed data never contain 0 byte *(while ASCII and UTF8 encodings are used, in case of UTF16 we have bigger problems everywhere else in agent)*.
- therefore 0 byte can be safely used to mean missing member in hashed data (there will ever be clash)

MrZammler: Ok thanks, pushed https://github.com/netdata/netdata/pull/11399/commits/8b43102095009381bde28939f4fa9aaa16c13dee

underhood (approved):

thiagoftsm (approved):
I tested this PR and I also tested the current master branch with the same alarm, and I could notice difference when I made requests for our `api`. LGTM!

MrZammler:
Thanks all!

PR #11410, opened 2021-08-09
eBPF Hard IRQ latency

UmanShahzad:
##### Summary

This PR brings eBPF-based hard IRQ latency tracking.

##### Component Name

ebpf.plugin

##### Test Plan

- Compile this branch
- Check that a hard IRQ latency graph is showing in the system overview section.


collectors/ebpf.plugin/ebpf.c
```diff
@@ -1544,6 +1567,8 @@ int main(int argc, char **argv)
             NULL, NULL, ebpf_modules[EBPF_MODULE_DISK_IDX].start_routine},
         {"EBPF MOUNT" , NULL, NULL, 1,
             NULL, NULL, ebpf_modules[EBPF_MODULE_MOUNT_IDX].start_routine},
+        {"EBPF HARDIRQ" , NULL, NULL, 1,
```
thiagoftsm, line 1570: Hello,

Are you sure we will have one for `hardirq` and another for `softirq`? Can we have both in one unique thread?

UmanShahzad: @thiagoftsm It all comes down to how we look at our architecture. Theoretically we could combine every thread we currently have into one thread and find a way to make it work. So the question is, how is the border decided?

If we base it off of unique kernel collectors in the kernel-collections repo, then it should be 1 thread for each, which is the model I've followed here.

thiagoftsm: My question was based in the fact that both are working with IRQ, take a look at `filesystem` thread, for example, there we are monitoring `ext4`, `nfs`, `xfs`, `zfs` and `btrfs`. 

I did not take a look inside both eBPF programs that you are working yet, my doubt has relationship with the `IRQ` that both has in its name. If they are not bringing the same charts, so we must keep them separated.

UmanShahzad: Ah I see, okay!

The main difference between the two, is that hard IRQ IDs are not set in stone, but the software IRQs are. The actual charts for each will be separate charts, but the parameters and type of data of the chart will be the same. Would that warrant making a single thread do the collection then? (my own opinion is that there is no issue here, so I'm happy to structure this so it combines them).

thiagoftsm: Considering your description, to unify both is the simplest way to bring both and save your time. :)

UmanShahzad: Alright sounds good!

UmanShahzad: Ok after looking at this again, I honestly don't think we should combine them, as they are only coincidentally similar in name, but for example their tracepoints are different, the labels are static for one and dynamic for the other, and there will likely be more differences I encounter as I develop them.

UmanShahzad:
So now just need to:

- Have reader thread publish aggregated data across CPUs to collector thread.
- Have collector thread write out the published data to netdata agent.

UmanShahzad:
Alright last thing before this is ready for review is to properly add dimensions only when they actually don't exist, and only once.

UmanShahzad:
For reference, here's how the chart looks so far though:

image

UmanShahzad:
This is ready for review. Example of the graph:

image

collectors/ebpf.plugin/ebpf_hardirq.c
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    read_thread_closed = 1;
+    return NULL;
+}
+
+static void hardirq_create_charts()
+{
+    ebpf_create_chart(
+        "system",
```
thiagoftsm, line 342: Instead to use `"system"` here, I suggest you to use this [defintion](https://github.com/netdata/netdata/blob/f2e95f70819a70c6cffacb42784285f0e54e0bb4/collectors/ebpf.plugin/ebpf.h#L106), because  whether new menu comes with a different name for everything grouped inside the current `system`, this will simplify our life when we need to do modifications.

collectors/ebpf.plugin/ebpf_hardirq.c
```diff
@@ -0,0 +1,503 @@
...
+    hardirq_create_static_dims();
+    pthread_mutex_unlock(&lock);
+
+    // loop and read from published data until ebpf plugin is closed.
+    while (!close_ebpf_plugin) {
+        pthread_mutex_lock(&collect_data_mutex);
+        pthread_cond_wait(&collect_data_cond_var, &collect_data_mutex);
+        pthread_mutex_lock(&lock);
+
+        // write dims now for all hitherto discovered IRQs.
```
thiagoftsm, line 439: A suggestion, the next four function calls are made to write charts for Netdata, I think they could go for a `static inline` function.
Like I said before, please, use the previous cited definitions instead to repeat the word "system" here.

collectors/ebpf.plugin/ebpf_hardirq.c
```diff
@@ -0,0 +1,503 @@
...
+
+    if (!em->enabled) {
+        goto endhardirq;
+    }
+
+    if (ebpf_update_kernel(&hardirq_data)) {
+        goto endhardirq;
+    }
+
+    bool one_tp_succeeded = false;
```
thiagoftsm, line 477: Another suggestion here, the whole block to enable the IRQ can go inside a specific function.

collectors/ebpf.plugin/ebpf_hardirq.c
```diff
@@ -0,0 +1,503 @@
...
+        .internal_input = 0,
+        .user_input = 0,
+        .type = NETDATA_EBPF_MAP_CONTROLLER,
+        .map_fd = ND_EBPF_MAP_FD_NOT_INITIALIZED
+    }
+};
+
+static ebpf_data_t hardirq_data;
+
+static ebpf_tracepoint_t hardirq_tracepoints[] = {
```
thiagoftsm, line 41: When we have a big sample of repeated strings like we have here, old compilers could create a lot of garbage, I suggest to create definitions for `irq` and 'irq_vectors` and use these defines here.

UmanShahzad: Do you mean to use the C preprocessor? `#define HARDIRQ_TP_CLASS_IRQ "irq"`? Will this not just paste the same exact string and be effectively the same thing?

Or do you mean to use `const char *`? That would likely guarantee deduplication in old compilers that behave like this.

collectors/ebpf.plugin/ebpf_hardirq.c
UmanShahzad: 👍

collectors/ebpf.plugin/ebpf_hardirq.c
UmanShahzad: 👍

UmanShahzad: Just to be clear, the `write_begin_chart`, `write_end_chart` and `avl_traverse_lock` cannot be made `static inline` unless we moved them all to headers, since they come from other compilation units currently. So I assume you only meant to static inline `hardirq_write_static_dims`, which I'll do - let me know if I misunderstood.

UmanShahzad: (Doing an inline on `hardirq_write_dims` is not going to do anything because it's being used as a callback at runtime).

collectors/ebpf.plugin/ebpf_hardirq.c
UmanShahzad: In this particular case I don't know if it's worth it - the function will never be re-used, and just causes the reader to have to jump somewhere else.

UmanShahzad: Actually, I think I can make it reusable by making it operate on `ebpf_tracepoint_t[]` and return how many tracepoint enables succeeded.

Thanks for the idea!

collectors/ebpf.plugin/ebpf_hardirq.c
```diff
@@ -0,0 +1,503 @@
...
+    return NULL;
+}
+
+static void hardirq_create_charts()
+{
+    ebpf_create_chart(
+        "system",
+        "hardirq_latency",
+        "Hardware IRQ latency",
+        "latency (milliseconds)",
```
ilyam8, line 345:
```suggestion
        "milliseconds",
```

I'd rather use just `milliseconds` as units. I don't think we need "latency" there.

UmanShahzad: I think this is referring to the X-axis label. See the photo below.

UmanShahzad: I think it'd be fine to make it just "milliseconds" though, let me know if given the above photo it still makes sense to you too.

ilyam8: Yes, was referring to the Y-axis (units).

collectors/ebpf.plugin/ebpf_hardirq.c
UmanShahzad: FYI I tried doing it with `const char *` but had way too many issues with the compiler complaining e.g. `"initializer element is not constant"` and other const-related things. It was easier to just leave it as is.

thiagoftsm: > Do you mean to use the C preprocessor? #define HARDIRQ_TP_CLASS_IRQ "irq"? Will this not just paste the same exact string and be effectively the same thing?

Yes, I was thinking to use preprocessors, this is a pattern we are using in different plugins like [perf.plugin](https://github.com/netdata/netdata/blob/ade4b8798173142fd7cf70be949343217dbe72b0/collectors/perf.plugin/perf_plugin.c#L63-L66).

collectors/ebpf.plugin/ebpf_hardirq.c
thiagoftsm: You are right, let us move it inside a specific function, I forgot where they were declared.

thiagoftsm:
I tested this PR on kernels `5.13.8`, `5.10.56`, 4.14.239 and 3.10.0-1160, this last using CentOS 7.9.2009, everything worked, I am only waiting for actions on this [comment](https://github.com/netdata/netdata/pull/11410#discussion_r689867752) before to approve PR.
Congratulations @UmanShahzad ! :smile:

collectors/ebpf.plugin/ebpf_hardirq.c
UmanShahzad: Alright, addressed in https://github.com/netdata/netdata/pull/11410/commits/72f4700c1fd75d7132f9a38ebef5e0212d6d6f3a

UmanShahzad:
Thank you @thiagoftsm ! I addressed that comment now in https://github.com/netdata/netdata/pull/11410/commits/72f4700c1fd75d7132f9a38ebef5e0212d6d6f3a

thiagoftsm (approved):
As described [here](https://github.com/netdata/netdata/pull/11410#issuecomment-899936226) I tested previously in different kernels, and now I also tested on `4.18` running `CentOS 8.4.2015`, everything worked as expected. LGTM!

UmanShahzad:
@thiagoftsm thank you!

@vlvkobal or @underhood can one of you also review? Thanks!

vlvkobal (approved):

PR #11374, opened 2021-07-27
Added support for bundling protobuf as part of the install.

Ferroin:
##### Summary

This adds basic support for bundling a statically linked version of libprotobuf as part of the regular install process. This is currently used by the Prometheus remote write exporter, and will be used in the near future for a new version of the protocol being used to communicate with Netdata Cloud.

If both Netdata Cloud _and_ the Prometheus remote write exporter are disabled, then we will completely skip bundling of protobuf.

The user can pass the option `--with-system-protobuf` to the installer to use a system-provided copy of protobuf. If doing so, we will skip bundling of protobuf, but will print a deferred error at the end of the install warning about how fragile such a setup is.

##### Component Name

area/packaging

##### Test Plan

The build itself seems to work, and from inspection of the resulting binary it looks like it properly includes libprotobuf. Ideally the Prometheus remote writ exporter should be tested with this, though I lack the setup required to do that properly.

##### Additional Information

This adds a non-trivial amount of extra time to the build process, largely due to how big the `protoc` command is (and the fact that it is unconditionally built with support for all languages for which there are protobuf bindings in the official repo). There is unfortunately not much that can be done about this (it’s already set to parallelize the build).

PR #11432, opened 2021-08-12
Default to not using LTO for builds.

Ferroin:
##### Summary

This significantly speeds up the build process (cutting build times roughly in half) and avoids strange linking errors such as what we’re seeing with Debian 9 in #11374.

##### Component Name

area/build

##### Test Plan

Verified that the build still works correctly without LTO and that the CI runs much faster.

##### Additional Information

We _might_ consider enabling this again at some time in the future, but right now we can’t even quantify the performance benefits of using LTO.

Users who _really_ want this can still enable it by passing `--enable-lto` to the installer or configure script.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -144,9 +144,9 @@ AC_ARG_ENABLE(
 )
 AC_ARG_ENABLE(
     [lto],
-    [AS_HELP_STRING([--disable-lto], [Link Time Optimizations @<:@default autodetect@:>@])],
+    [AS_HELP_STRING([--enable-lto], [Link Time Optimizations @<:@default disabled@:>@])],
     ,
-    [enable_lto="detect"]
+    [enable_lto="no"]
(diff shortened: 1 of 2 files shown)


diizzyy:
Are there any noticable changes in terms of size and performance? This approach also seems to hide underlying issues rather than fixing them,

Ferroin:
> Are there any noticable changes in terms of size and performance? This approach also seems to hide underlying issues rather than fixing them,

We have no way currently to properly quantify any performance impact. The thing is though, the difference in build times and resource usage when building is _huge_, and large segment of users are building locally for installs and updates. So we have no real proof that there’s any actual benefit to using LTO, but very concrete proof that it is causing us problems at least in some cases, and that it’s got a significant cost both for our developers and our users.

stelfrag (approved):

Issue #11434, opened 2021-08-13
apcupsd status check inconsistent


0x3333:
##### Bug report summary

In the apcupsd script, it has been changed in PR #8688 to replace a check if the `ups` is online from "ONLINE" and "ONBATT" to "COMMLOST" or "SHUTTING DOWN", but the `awk` script has not been changed accordingly. 

In my case, when the nobreak is on status "TRIM ONLINE", the script thinks it is not online and raises an alarm.

The alarm:
![Screen Shot 2021-08-13 at 08 28 27](https://user-images.githubusercontent.com/289289/129350790-61638a05-0c8e-4017-965c-5ec177edb043.png)

The status:
![Screen Shot 2021-08-13 at 08 28 47](https://user-images.githubusercontent.com/289289/129350808-ce2ef196-4772-44f0-b15e-510eeb4007d0.png)

##### OS / Environment

```
$ uname -a; grep -Hv "^#" /etc/*release
Linux cube 5.10.0-0.bpo.7-amd64 #1 SMP Debian 5.10.40-1~bpo10+1 (2021-06-04) x86_64 GNU/Linux
/etc/os-release:PRETTY_NAME="Debian GNU/Linux 10 (buster)"
/etc/os-release:NAME="Debian GNU/Linux"
/etc/os-release:VERSION_ID="10"
/etc/os-release:VERSION="10 (buster)"
/etc/os-release:VERSION_CODENAME=buster
/etc/os-release:ID=debian
/etc/os-release:HOME_URL="https://www.debian.org/"
/etc/os-release:SUPPORT_URL="https://www.debian.org/support"
/etc/os-release:BUG_REPORT_URL="https://bugs.debian.org/"
```

##### Netdata version

```
$ netdata -W buildinfo
Version: netdata v1.31.0-190-nightly
Configure options:  '--prefix=/usr' '--sysconfdir=/etc' '--localstatedir=/var' '--libexecdir=/usr/libexec' '--libdir=/usr/lib' '--with-zlib' '--with-math' '--with-user=netdata' '--with-bundled-lws' 'CFLAGS=-O2' 'LDFLAGS='
Features:
    dbengine:                YES
    Native HTTPS:            YES
    Netdata Cloud:           YES
    ACLK Next Generation:    NO
    ACLK Legacy:             YES
    TLS Host Verification:   YES
Libraries:
    jemalloc:                NO
    JSON-C:                  YES
    libcap:                  NO
    libcrypto:               YES
[...]

PR #8688, opened 2020-04-12
charts.d/apcupsd: fix ups status check

ilyam8:
##### Summary

Fixes: #8687
Fixes: https://github.com/netdata/netdata/issues/5428#issuecomment-589941219

Current implementation treats all statuses except `ONLINE` and `ONBATT` as failed statuses.

https://github.com/netdata/netdata/blob/231d19351d0ea20aa66fd204215cbb24f34d7575/collectors/charts.d.plugin/apcupsd/apcupsd.chart.sh#L50-L51

All available statuses [according docs](https://www.systutorials.com/docs/linux/man/8-apcaccess/):

- One or more of the following (space-separated): CAL TRIM BOOST ONLINE ONBATT OVERLOAD LOWBATT REPLACEBATT NOBATT SLAVE SLAVEDOWN
- COMMLOST
- SHUTTING DOWN

We have 2 reported problems:
 - add `CAL ONBATT` status ([link](https://github.com/netdata/netdata/issues/5428#issuecomment-589941219))
 - add `TRIM ONLINE` status ([link](https://github.com/netdata/netdata/issues/5428#issuecomment-612635833))

I agree with @Mic-Pod that it is safe to treat all statuses except `COMMLOST` and `SHUTTING DOWN` as ok and try to collect data from upses. I am not 100% sure but it looks like a better alternative then keep adding statuses one by one. 

##### Component Name

`collectors/charts.d`

##### Test Plan

- [X] tested new logic with the following script

<details>
<summary>click click</summary>

```sh
is_ups_alive() {
  case "$1" in
    "" | "COMMLOST" | "SHUTTING DOWN") return 1 ;;
    *) return 0 ;;
  esac
}

extract_ups_status() {
  sed -e 's/STATUS.*: //' -e 't' -e 'd' <<< "$1"
}

tests=(
  "before\nSTATUS : ONBATT\n after"
  "before\nSTATUS : ONBATT SOMETHING\n after"
  "before\nSTATUS : ONLINE\n after"
  "before\nSTATUS : ONLINE SOMETHING\n after"
  "before\nSTATUS : TRIM ONLINE\n after"
  "before\nSTATUS : COMMLOST\n after"
  "before\nSTATUS : SHUTTING DOWN\n after"
  ""
)

for test in "${tests[@]}"; do
  if is_ups_alive "$(extract_ups_status "$(echo -e "$test")")"; then
    echo "OK: $test"
  else
    echo "FAIL: $test"
  fi
done
```

</details>

Result:

[...]

PR #11435, opened 2021-08-13
Fix issue #11434 regarding inconsistent status check on  component.

0x3333:
##### Summary

Fixes: #11434

Fix an inconsistent status check on `apcupsd`. After PR #8688, it checks for "COMMLOST" or "SHUTTING DOWN" to check if ups is online, but the PR missed the same check on the `apcupsd_update()` function.

##### Component Name

`apcupsd`

##### Test Plan

I created a file with output from `apcaccess` and run through all possible values for status. Now it consider online all options for status except "COMMLOST" or "SHUTTING DOWN".

##### Additional Information

I got these alarms(`apcupsd_last_collected_secs`) for several months, but I didn't dig, just ignored them. Today I got upset and started looking what was happening, for my surprise, it was indeed a bug.


ilyam8 (approved):

ilyam8:
Thanks for the fix, @0x3333.

vlvkobal (approved):

PR #11442, opened 2021-08-18
Installation review

hugovalente-pm:
Summary
Review the documentation to give more emphasis on using kickstart as part of the work on netdata/product#1531

Component Name
Documentation

Test Plan
Review docs after merge

Additional Information


claim/README.md
```diff
@@ -102,45 +122,104 @@ it will use these values to attempt to connect the container, automatically addi
...
+    - netdatacache:/var/cache/netdata
+    - /etc/passwd:/host/etc/passwd:ro
+    - /etc/group:/host/etc/group:ro
+    - /proc:/host/proc:ro
+    - /sys:/host/sys:ro
+    - /etc/os-release:/host/etc/os-release:ro
+  environment:
+    - NETDATA_CLAIM_TOKEN=TOKEN
+    - NETDATA_CLAIM_URL="https://app.netdata.cloud"
+    - NETDATA_CLAIM_ROOMS=ROOM1,ROOM2
```
manos-saratsis, line 187: @hugovalente-pm not sure if we should have some draft values for the rooms and token

manos-saratsis (approved):
Only unsure if we shall have some draft values on token and room ids

claim/README.md
hugovalente-pm: @manos-saratsis if you want to place some draft values we could, I just kept the original _values_. I personally I don't see the added value of putting some dummy values there and think that could lead to confusion, since values like ROOM1 and ROOM2 and TOKEN user (should) immediately see that these are to be replaced, if we put a dummy value it could be not so clear

claim/README.md
```diff
@@ -41,44 +38,67 @@ There are two important notes regarding connecting nodes:
 
 ## How to connect a node
 
+There will be two main flows from where you might want to connect a node to Netdata Cloud.
```
kaskavel, line 41: There is also the option of adding nodes from table view (nodes tab), where a user can press the (+) button. Should we mention this as well?

kaskavel (approved):

claim/README.md
hugovalente-pm: I see what you mean it opens the "Manage War Room / Add Nodes" tab on the right... that one skipped me 😓
For consistency I believe it is best to include it, will add this

kaskavel (dismissed):

kaskavel (approved):

manos-saratsis (approved):

Issue #11443, opened 2021-08-18
netdata-updater failing on all of my systems with /lib/cpp failes sanity check


pnunn:
##### Bug report summary

netdata is trying to run the updater from cron and failing.

Large amounts of output that seems to work, stuff built OK etc.
Gets to protobuf and fails with the error 
checking how to run the C preprocessor... gcc -E
checking how to run the C++ preprocessor... /lib/cpp
configure: error: in `/tmp/netdata-protobuf-kTxGOH/protobuf-3.17.3':
configure: error: C++ preprocessor "/lib/cpp" fails sanity check
See `config.log' for more details
 FAILED   

This is happening on multiple systems.

##### OS / Environment

To get this information, execute the following commands based on your operating system:
Linux cwprfw 5.4.0-67-generic #75-Ubuntu SMP Fri Feb 19 18:03:38 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
/etc/lsb-release:DISTRIB_ID=Ubuntu
/etc/lsb-release:DISTRIB_RELEASE=20.04
/etc/lsb-release:DISTRIB_CODENAME=focal
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 20.04.2 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:VERSION="20.04.2 LTS (Focal Fossa)"
/etc/os-release:ID=ubuntu
/etc/os-release:ID_LIKE=debian
/etc/os-release:PRETTY_NAME="Ubuntu 20.04.2 LTS"
/etc/os-release:VERSION_ID="20.04"
/etc/os-release:HOME_URL="https://www.ubuntu.com/"
/etc/os-release:SUPPORT_URL="https://help.ubuntu.com/"
/etc/os-release:BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
/etc/os-release:PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
/etc/os-release:VERSION_CODENAME=focal
/etc/os-release:UBUNTU_CODENAME=focal

Place the output from the command in the code section below. 
```
[...]

PR #11446, opened 2021-08-20
Don’t bail early if we fail to build cloud deps with required cloud.

Ferroin:
##### Summary

Due to having two ACLK implementations, failing to build any arbitrary set of cloud dependencies is not guaranteed to prevent us from building a working cloud implementation, so we should fall back to relying on the configure script for determining if the cloud can be built.

This also ensures that the configure script properly bails if we can’t enable either ACLK implementation _and_ were asked to require the cloud.

##### Component Name

area/packaging

##### Test Plan

CI passes

##### Additional Information

Fixes: #11443

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -879,6 +879,10 @@ if test "$enable_cloud" != "no" -a "$aclk_legacy" != "no"; then
     fi
 fi
 
+if test "$enable_cloud" = "yes" -a "$enable_aclk" != "yes"; then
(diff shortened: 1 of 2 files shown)


underhood:
ignore AWS kinesis errors on this PR. This is going to be fixed soon, not an issue with this PR

MrZammler (approved):

vkalintiris (approved):

PR #11450, opened 2021-08-23
Check for failed protobuf configure or make

MrZammler:
##### Summary

This PR adds a check in `netdata-installer.sh` for the status of configure and make phases of protobuf. If any of those fails, it makes sure it returns back to original directory to continue with the rest of the installer.

It also does not bail out from the configure phase of the agent if bundled protobuf is not found. ACLK Legacy should be available in this case, since the builds with NG are separate.

##### Component Name

Installer

##### Test Plan

If dependencies are there (i.e. g++), ACLK NG should be built with protobuf.
If configure phase (uninstall g++) or build of protobuf (`ctrl-z` after configure, break a protobuf file, `fg`), observe that installer continues with the rest of components, and does not build ACLK NG.

##### Additional Information

This does not solve the big problem, which is how to eventually "force" use of ACLK NG with protobuf.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -657,8 +657,7 @@ if test "${with_bundled_protobuf}" != "no"; then
     fi
     AC_MSG_RESULT([${bundled_proto_avail}])
     if test "${with_bundled_protobuf}" == "yes" -a "${bundled_proto_avail}" != "yes"; then
-        AC_MSG_WARN([Bundled protobuf requested using --with-bundled-protobuf but it cannot be used/found, skipping ACLK NG])
-        aclk_ng="no"
+        AC_MSG_ERROR([Bundled protobuf requested using --with-bundled-protobuf but it cannot be used/found])
     fi
     if test "${with_bundled_protobuf}" == "detect" -a "${bundled_proto_avail}" == "yes"; then
         with_bundled_protobuf="yes"
--- a/configure.ac
+++ b/configure.ac
@@ -657,7 +657,8 @@ if test "${with_bundled_protobuf}" != "no"; then
     fi
     AC_MSG_RESULT([${bundled_proto_avail}])
     if test "${with_bundled_protobuf}" == "yes" -a "${bundled_proto_avail}" != "yes"; then
-        AC_MSG_ERROR([Bundled protobuf requested using --with-bundled-protobuf but it cannot be used/found])
+        AC_MSG_WARN([Bundled protobuf requested using --with-bundled-protobuf but it cannot be used/found, skipping ACLK NG])
+        aclk_ng="no"
     fi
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -657,7 +657,8 @@ if test "${with_bundled_protobuf}" != "no"; then
     fi
     AC_MSG_RESULT([${bundled_proto_avail}])
     if test "${with_bundled_protobuf}" == "yes" -a "${bundled_proto_avail}" != "yes"; then
-        AC_MSG_ERROR([Bundled protobuf requested using --with-bundled-protobuf but it cannot be used/found])
+        AC_MSG_WARN([Bundled protobuf requested using --with-bundled-protobuf but it cannot be used/found, skipping ACLK NG])
```
Ferroin, line 660: This should still be an error. The install script _only_ passes this configure flag if the protobuf build succeeded, and if it got passed but we can’t find the files, it’s very much an error.

Ferroin:
We probably need something like this for the other bundling cases (except libJudy). They technically do return to the correct directory, but they also don’t properly return a non-zero exit status for build failures, and don’t bail early if configuration fails.

configure.ac
MrZammler: I think you're right.... In the morning, I did have problem with this.... Might have been some switch I did some tries... Will look into it and revert probably. Thanks.

underhood: Agreed with @ferroin here the flag is supposed to be 3state. If nothing given try use bundled and if not use system. This place checks case where `--with-bundled-protobuf`  was used. In such case user specifically wants to use bundled version of protobuf. In such case we need to fail with error if it is not present

MrZammler: Ok, I could have sworn I had a case with it failing, without (I think) issuing `--with-bundled-protobuf`... Reverted.

MrZammler:
> We probably need something like this for the other bundling cases (except libJudy). They technically do return to the correct directory, but they also don’t properly return a non-zero exit status for build failures, and don’t bail early if configuration fails.

Agree, do you think we need to do it on this PR, or keep them separate?

vkalintiris:
> Agree, do you think we need to do it on this PR, or keep them separate?

I'd prefer separate PRs for this in order to keep things as simple as possible (and have PRs that are easily reversible).

vkalintiris (approved):
LGTM. I did not run this manually on my end, because I came up with a similar patch that worked correctly before I saw your PR.

Ferroin:
> Agree, do you think we need to do it on this PR, or keep them separate?

I would normally argue for having them all in one PR, but in this case getting this merged is a high priority and adding them here would slow things down, so I think separate PRs would be best.

Ferroin (approved):

MrZammler:
Thanks guys!

PR #11451, opened 2021-08-23
Remove AWS Kinesis CI.

Ferroin:
##### Summary

We’re constantly seeing issues with upstream code on these checks that have nothing to do with netdata, so there0s no real point in us having them in the first place.

##### Component Name

area/ci

##### Test Plan

n/a

##### Additional Information

Once this is merged, we should (finally) see all our CI passing for the first time in quite some time.


underhood:
#11424 also fixes those checks apart from FC32 one. But yep agree on removing it

vlvkobal (approved):

stelfrag (approved):

Issue #11453, opened 2021-08-24
Problem with Netdata on the PostgreSQL database with cascading replication


vitabaks:
##### Bug report summary

When installing and configuring Netdata for monitoring the local postgres database configured in the cascade replication mode, there are simply no postgres charts in the netdata dashboard.

When debugging the postgres module, we observe the following error:
```
sudo /usr/libexec/netdata/plugins.d/python.d.plugin postgres debug trace nolock
2021-08-24 12:54:09: python.d INFO: plugin[main] : using python v2
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : looking for 'python.d.conf' in ['/etc/netdata', '/usr/lib/netdata/conf.d']
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : loading '/usr/lib/netdata/conf.d/python.d.conf'
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : '/usr/lib/netdata/conf.d/python.d.conf' is loaded
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : looking for 'pythond-jobs-statuses.json' in /var/lib/netdata
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : loading '/var/lib/netdata/pythond-jobs-statuses.json'
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : '/var/lib/netdata/pythond-jobs-statuses.json' is loaded
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : [postgres] looking for 'postgres.conf' in ['/etc/netdata/python.d', '/usr/lib/netdata/conf.d/python.d']
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : [postgres] loading '/usr/lib/netdata/conf.d/python.d/postgres.conf'
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : [postgres] '/usr/lib/netdata/conf.d/python.d/postgres.conf' is loaded
2021-08-24 12:54:09: python.d INFO: plugin[main] : [postgres] built 4 job(s) configs
2021-08-24 12:54:09: python.d DEBUG: plugin[main] : postgres[socket] was previously active, applying recovering settings
2021-08-24 12:54:09: python.d DEBUG: postgres[socket] : server version: 130004
2021-08-24 12:54:09: python.d DEBUG: postgres[socket] : superuser: True
[...]

PR #11455, opened 2021-08-25
Fix postgres replication_slot chart on standby

anayrat:
##### Summary

Fixes: #11453

##### Component Name

Postgres plugin

##### Test Plan

Create a replication slot on a standby server. Before this PR you got this error in postgres logs:

> ERROR:  recovery is in progress
> HINT : WAL control functions cannot be executed during recovery.

After this PR the error should not appear and the charts will be created.

##### Additional Information

I applied the same fix from https://github.com/OPMDG/check_pgactivity/pull/246 
So thanks for check_pgactivity project and maintainers.

### DIFF
--- a/collectors/python.d.plugin/postgres/postgres.chart.py
+++ b/collectors/python.d.plugin/postgres/postgres.chart.py
@@ -576,8 +576,20 @@ FROM
         slot_type,
         COALESCE (
           floor(
-            (pg_wal_lsn_diff(pg_current_wal_lsn (),slot.restart_lsn)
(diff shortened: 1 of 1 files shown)


ilyam8:
@anayrat thanks for the fix!

ilyam8:
@vitabaks, thanks for [testing the fix](https://github.com/netdata/netdata/issues/11453#issuecomment-905370198)!

ilyam8 (approved):

vlvkobal (approved):

PR #11462, opened 2021-08-27
Adds aclk/cloud state command to netdatacli

underhood:
##### Summary
Implements product ask to have current aclk/cloud state available trough `netdatacli`. This is useful for example in cases where the local dashboard is not accessible or is disabled.

Us still having the ACLK Legacy here adds some awkwardness to the thing.

##### Test Plan
Try with various ACLK implementations. (Hopefully we will get rid of ACLK legacy soon)

run
`netdatacli aclk-state` and/or `netdatacli aclk-state json`

### DIFF
--- a/aclk/aclk_api.c
+++ b/aclk/aclk_api.c
@@ -197,7 +197,7 @@ struct label *add_aclk_host_labels(struct label *label) {
 
 char *aclk_state(void) {
 #ifndef ENABLE_ACLK
-    return strdupz("ACLK Available: No\n");
+    return strdupz("ACLK Available: No");
 #else
 #ifdef ACLK_NG
     if (aclk_ng)
@@ -213,7 +213,7 @@ char *aclk_state(void) {
 
 char *aclk_state_json(void) {
 #ifndef ENABLE_ACLK
-    return strdupz("{\"aclk-available\": false}\n");
+    return strdupz("{\"aclk-available\": false}");
 #else
 #ifdef ACLK_NG
     if (aclk_ng)
--- a/aclk/aclk_api.c
+++ b/aclk/aclk_api.c
@@ -194,3 +194,35 @@ struct label *add_aclk_host_labels(struct label *label) {
 #endif
     return label;
 }
+
+char *aclk_state(void) {
+#ifndef ENABLE_ACLK
+    return strdupz("ACLK Available: No\n");
+#else
+#ifdef ACLK_NG
+    if (aclk_ng)
+        return ng_aclk_state();
+#endif
+#ifdef ACLK_LEGACY
+    if (!aclk_ng)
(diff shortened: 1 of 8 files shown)


aclk/aclk_api.c
```diff
@@ -194,3 +194,35 @@ struct label *add_aclk_host_labels(struct label *label) {
 #endif
     return label;
 }
+
+char *aclk_state(void) {
+#ifndef ENABLE_ACLK
+    return strdupz("ACLK Available: No\n");
```
MrZammler, line 200: Hey Timo.

Very little thing: when compiled with `--disable-cloud`, there is only `ACLK Available: No` printed. There is however an extra newline added when the message is actually printed I guess, so there is an empty line before returning to prompt, just a visual thing (or maybe when something else needs to parse it). Thanks! (Also same in the json equivalent).

aclk/legacy/agent_cloud_link.c
```diff
@@ -1450,3 +1450,53 @@ int legacy_aclk_update_alarm(RRDHOST *host, ALARM_ENTRY *ae)
...
+        "ACLK Available: Yes\n"
+        "ACLK Implementation: Legacy\n"
+        "Claimed: "
+    );
+
+    char *agent_id = is_agent_claimed();
+    if (agent_id == NULL)
+        buffer_strcat(wb, "No\n");
+    else {
+        buffer_sprintf(wb, "Yes\nClaimed Id: %s\n", agent_id);
```
MrZammler, line 1469: In it's json counterpart, `claimed-id` will be printed (`"claimed-id":null`) even if not claimed, while here it will be only visible if `agent_id` is not null. Don't know if we should also include it here too even when no claim_id exists? (Or remove it also from json when it's not claimed) ? Just to have similar output for both. i.e.:

```
# ../install_netdata_11462/netdata/usr/sbin/netdatacli aclk-state
ACLK Available: Yes
ACLK Implementation: Next Generation
Claimed: No
Online: No
# ../install_netdata_11462/netdata/usr/sbin/netdatacli aclk-state-json
{"aclk-available":true,"aclk-implementation":"Next Generation","agent-claimed":false,"claimed-id":null,"online":false}
```

underhood: I wanted JSON part to have fixed list of keys on purpose to aid using JSON versions in user scripts etc.
Also the JSON version will be used as new `/api/v1/aclk-state` HTTP API in future PRs (after more info is added, e.g. chart streaming state, alerts streaming state etc.)

aclk/aclk_api.c
underhood: thx removed

stelfrag (approved):

MrZammler (approved):

PR #11470, opened 2021-08-29
eBPF OOM kill tracking

UmanShahzad:
##### Summary

This PR brings eBPF-based OOM kill tracking. https://github.com/netdata/product/issues/2158

##### Component Name

ebpf.plugin

##### Test Plan

- Compile this branch.
- Turn off swap, e.g. `sudo swapoff -a`.
- Run netdata with `sudo choom -n -1000 -- sudo netdata -D` (so it doesn't get OOM killed).
- Run something that will exhaust memory quickly, e.g. `sudo choom -n 1000 -- tail /dev/zero`.
- Check that the chart shows up (`apps.oomkills`).

Tested with:

- Ubuntu 18.04.1 (Linux 4.15)
- Ubuntu 21.04 (Linux 5.11.0-31)


collectors/ebpf.plugin/ebpf_oomkill.c
```diff
@@ -0,0 +1,203 @@
...
+        "oomkills",
+        "OOM kills",
+        EBPF_COMMON_DIMENSION_KILLS,
+        "mem",
+        NULL,
+        NETDATA_EBPF_CHART_TYPE_LINE,
+        20020,
+        NULL, NULL, 0,
+        NETDATA_EBPF_MODULE_NAME_OOMKILL
+    );
```
thiagoftsm, line 126: When we are creating charts here, we are not adding any dimensions, and this is completely different of all other charts we have right now in `apps` menu. I think this will be strange for our users. Following the current logic that we have, I think the users will expect that all  dimensions defined at `apps_groups.conf` will have `0` when an `oom` did not happen, and they will have a value different of zero when an event happens. 

We also could have a global chart show total of `oom` that are happening on machine. :thinking:

UmanShahzad: I think the best plan is then to just move this chart next to the existing OOM chart `mem.oom_kill`, which only shows the raw count of OOM kills on the machine.

> We also could have a global chart show total of oom that are happening on machine. thinking

This chart already exists:

image

What do you think?

thiagoftsm: This is a good approach, and I think we can move forward with this, and if it is necessary to bring something integrated with `apps`, we follow the pattern that we already have.

UmanShahzad: Ok perfect, I'll just move the chart then and it should be good. This is actually a really useful chart, but only in those very rare & desperate times!

UmanShahzad:
Here's an example of the chart sitting below the global oomkill one:

image

thiagoftsm:
> Here's an example of the chart sitting below the global oomkill one:
> 
> image

I have more two concerns with this PR:

1 -  To have a chart that is not showing any data until a process raises an `oom`, probably we need to have an additional dimension to count numbers of `oom` for we have some data on this chart.
2 - Another thing that is calling my attention is the fact that we are not obsoleting the dimension on this chart. I think we could obsolete the dimension in the next iteration and recreate it when another event happens.

 @ilyam8 and @vlvkobal  what do you think about these concerns?

UmanShahzad:
@thiagoftsm I think there are actually many issues here as well. For one, whenever data for the chart first comes through (i.e. there is 1 OOM kill), you cannot see the dimension on the actual chart except as a label only, because it seems to require multiple entries before a line or similar is drawn. This makes it impossible to know *when* the OOM kill actually happened.

The only way to fix this seems to be to continuously write `SET <process comm>=0` for a while. But for how long? And should we consider writing this before the OOM kill happens too? If yes then we would need to do that for *all* PIDs!

I think we need a clear idea of what this chart needs to actually display and how it should display it, before moving forward on any other work on this.

UmanShahzad:
@thiagoftsm I think there are actually many issues here as well. For one, whenever data for the chart first comes through (i.e. there is 1 OOM kill), you cannot see the dimension on the actual chart except as a label only, because it seems to require multiple entries before a line or similar is drawn. This makes it impossible to know *when* the OOM kill actually happened.

The only way to fix this seems to be to continuously write `SET <process comm>=0` for a while. But for how long? And should we consider writing this before the OOM kill happens too? If yes then we would need to do that for *all* PIDs!

I think we need a clear idea of what this chart needs to actually display and how it should display it, before moving forward on any other work on this.

ilyam8:
> For one, whenever data for the chart first comes through (i.e. there is 1 OOM kill), you cannot see the dimension on the actual chart except as a label only, because it seems to require multiple entries before a line or similar is drawn.

Yes, we have that problem. It ~makes sense for `incremental` dimensions (you need 2 points to calc per second rate), but not for `absolute`.

There is a hack/workaround - [`store_first`](https://learn.netdata.cloud/docs/agent/collectors/plugins.d#chart) chart option. But it has other problems.

> store_first to make Netdata store the first collected value, assuming there was an invisible previous value set to zero (this is used by statsd charts - if the first data collected value of incremental dimensions is not zero based, unrealistic spikes will appear with this option set)

ilyam8:
> I think we need a clear idea of what this chart needs to actually display and how it should display it, before moving forward on any other work on this.

Can we start with integrating this module with the `apps.plugin`? (num of oom kills for groups in `apps_groups.conf`).  That would be awesome (and I think we do the same for all `ebpf.plugin` modules).

UmanShahzad:
@ilyam8 Thank you, @thiagoftsm & I talked about it and confirm that your recommendation is the best option currently. We'll basically do `SET <proc_comm>=0` for all processes discovered via `apps_groups.conf`, and a `SET <proc_comm>=1` only when we see the OOM kill appear for the PID in the eBPF map.

UmanShahzad:
Alright updated it according to our discussions.

### Normal situation (no OOM kill)

image

### With an OOM kill

image

collectors/ebpf.plugin/ebpf_oomkill.c
```diff
@@ -0,0 +1,198 @@
...
+void ebpf_oomkill_create_apps_charts(struct ebpf_module *em, void *ptr)
+{
+    UNUSED(em);
+
+    struct target *root = ptr;
+    ebpf_create_charts_on_apps("oomkills",
+                               "OOM kills",
+                               EBPF_COMMON_DIMENSION_KILLS,
+                               "mem",
+                               NETDATA_EBPF_CHART_TYPE_LINE,
```
ilyam8, line 153: Usually, we use "stacked" for apps charts when the sum of dimensions makes sense.

collectors/ebpf.plugin/ebpf.d/oomkill.conf
```diff
@@ -0,0 +1,7 @@
+# The `ebpf load mode` option accepts the following values :
+#  `entry` : The eBPF collector only monitors calls for the functions, and does not show charts related to errors.
+#  `return : In the `return` mode, the eBPF collector monitors the same kernel functions as `entry`, but also creates
+#            new charts for the return of these functions, such as errors.
+[global]
+    ebpf load mode = entry
+    update every = 1
```
thiagoftsm, line 7: When there is integration with `apps`, this file needs to have more line, please, take a look on `vfs.conf` for the missing lines.

collectors/ebpf.plugin/ebpf.h
```diff
@@ -113,6 +114,7 @@ typedef struct ebpf_tracepoint {
 #define NETDATA_EBPF_CHART_TYPE_STACKED "stacked"
 #define NETDATA_EBPF_MEMORY_GROUP "mem"
 #define NETDATA_EBPF_SYSTEM_GROUP "system"
+#define NETDATA_EBPF_APPS_GROUP "apps"
```
thiagoftsm, line 117: You started to use the function `ebpf_create_charts_on_apps` that uses `apps` by default, so this line is not more necessary, please, remove it.

collectors/ebpf.plugin/ebpf_oomkill.c
```diff
@@ -0,0 +1,198 @@
...
+            // impossible to get this condition.
+            error("key unexpectedly not available for deletion.");
+        }
+    }
+
+    // for each app, see if it was OOM killed. record as 1 if so otherwise 0.
+    struct target *w;
+    for (w = apps_groups_root_target; w != NULL; w = w->next) {
+        if (likely(w->exposed && w->processes)) {
+            bool was_oomkilled = false;
```
thiagoftsm, line 95: I do not have objections with usage of `bool`, but I do not remember functions where we are using this data type, I will suggest you use `CONFIG_BOOLEAN_YES` and `CONFIG_BOOLEAN_NO` to keep pattern with the rest of the code.

ilyam8: I think "CONFIG" prefix indicates that it is only for configuration.

<details>
<summary>grep "bool " --include "*.c" --include "*.h" -R *</summary>

```cmd
collectors/ebpf.plugin/ebpf.h:    bool enabled;
collectors/ebpf.plugin/ebpf_hardirq.c:        bool v_is_new = false;
collectors/ebpf.plugin/ebpf_hardirq.c:        bool name_saved = false;
collectors/ebpf.plugin/ebpf_hardirq.h:    bool dim_exists; // keep this after `int irq` for alignment byte savings.
collectors/proc.plugin/proc_net_wireless.c:    bool configured;
exporting/tests/test_exporting_engine.h:bool __wrap_mongoc_uri_set_option_as_int32(const mongoc_uri_t *uri, const char *option, int32_t value);
exporting/tests/test_exporting_engine.h:bool __wrap_mongoc_client_set_appname(mongoc_client_t *client, const char *appname);
exporting/tests/test_exporting_engine.h:bool __wrap_mongoc_collection_insert_many(
exporting/tests/exporting_doubles.c:bool __wrap_mongoc_uri_set_option_as_int32 (const mongoc_uri_t *uri, const char *option, int32_t value)
exporting/tests/exporting_doubles.c:bool __wrap_mongoc_client_set_appname (mongoc_client_t *client, const char *appname)
exporting/tests/exporting_doubles.c:bool __wrap_mongoc_collection_insert_many(
libnetdata/ebpf/ebpf.h:    bool retprobe;
web/api/tests/web_api.c:static void build_request(struct web_buffer *wb, const char *url, bool use_cr, size_t num_headers)
web/api/tests/web_api.c:    bool completed, use_cr;
```

</details>

collectors/ebpf.plugin/ebpf.d/oomkill.conf
UmanShahzad: In this case we actually always need apps integration, because that's the only chart we create - the one that requires apps group information.

UmanShahzad: So in other words, disabling apps integration here would be equivalent to disabling the whole thing.

collectors/ebpf.plugin/ebpf.h
UmanShahzad: Good point!

collectors/ebpf.plugin/ebpf_oomkill.c
UmanShahzad: Right, `bool` is a very standard data type so I don't see any issue with that, even if it isn't used that much already. It better represents the data we're trying to store here.

collectors/ebpf.plugin/ebpf_oomkill.c
UmanShahzad: :+1:

UmanShahzad: https://github.com/netdata/netdata/pull/11470/commits/40c2ddbdf57c0184eff3cc4625202c37654106e6

collectors/ebpf.plugin/ebpf.h
UmanShahzad: https://github.com/netdata/netdata/pull/11470/commits/21397ffa67451420bfe1d8a0f57f6865578141b6

collectors/ebpf.plugin/ebpf.d/oomkill.conf
thiagoftsm: All right, I understood, but this is not explicit in this file and I also did not see any information on documentation related to this chart, please, can you add them for us?

UmanShahzad: Added in https://github.com/netdata/netdata/pull/11470/commits/d9fc39769bbf9493a7c1c8dc3a6bc3fce2728869

ilyam8:
My test

- create a container

```cmd
docker run -d --name ubuntu --memory=100M ubuntu sleep infinity
```

- check oom kills

```cmd
grep oom /proc/vmstat
oom_kill 0
```

- force some oom kills

```cmd
for _ in $(seq 1 5); do docker exec ubuntu sh -c "tail /dev/zero"; sleep 1; done
Killed
Killed
Killed
Killed
Killed
```

- check oom kills

```cmd
grep oom /proc/vmstat
oom_kill 5

sudo dmesg | grep oom-killer
[  179.692923] tail invoked oom-killer: gfp_mask=0x100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
[  180.938345] tail invoked oom-killer: gfp_mask=0x100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
[  182.163180] tail invoked oom-killer: gfp_mask=0x100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
[  183.439132] tail invoked oom-killer: gfp_mask=0x100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
[  184.664986] tail invoked oom-killer: gfp_mask=0x100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
```

=> For some reason `apps.oomkills` chart doesn't show them (dimension's values are 0), I expected `other` dimension to be no 0.

UmanShahzad:
@ilyam8 This is happening because of a race: we depend upon the ebpf `process` thread to parse the proc filesystem so we can get PID data, but the processes you created in the test run are turned on and then OOM killed so fast that we never get a chance to record them properly in the procfs parsing step, and thus we can't find the PID's command/etc.

The best thing we can do I think, is to default to writing to `other` in such a case. This is a problem we cannot really solve unless we gathered the command name inside the eBPF program itself on the physical point of OOM kill, but because of some kernel limitations we cannot do this, although we tried a lot.

In realistic settings though, the process most likely to be OOM killed will already have been running for a while, and thus picked up by procfs parsing. So I don't think this is going to be an issue users ever encounter in reality.

After making the above fix and testing with commands like the following, here's an example chart:

```bash
$ for _ in $(seq 1 50); do docker exec ubuntu sh -c "tail /dev/zero" & done
$ sudo choom -n 1000 -- tail /dev/zero
```

image

ilyam8:
@UmanShahzad I see, thanks for the explanation!

_writing missed PIDs to "other" dim_ solves the issue, but I see several problems with that:
 - it is not oom specific and we need to apply it to all other modules too.
 - using "other" is not really correct, because the process could be in one of the apps groups - we just don't know.

Perhaps we need to add "short-lived" dimension. I think we can discuss how to deal with short-lived processes and then adjust all ebpf modules.

UmanShahzad:
@ilyam8 yes it's probably even an issue beyond ebpf: anything that has a PID through some means, but relies on procfs parsing for further details, will have an issue for short-lived PIDs.

Whatever strategy we end up applying, I think we should at least merge this (as it won't be an issue real users will almost ever experience), and then make a separate issue to investigate a proper strategy.

vlvkobal (approved):

thiagoftsm (approved):
LGTM!

ilyam8:
@UmanShahzad just in case, you forgot to update the OP: some TODO's are still there.

ilyam8 (approved):

PR #11476, opened 2021-09-01
Announce proto capability and enable if cloud supports

underhood:
##### Summary
Announces to cloud this agent is capable of proto.
In case the cloud then returns encoding proto we will switch to the new protocol.

##### Component Name
ACLK

##### Test Plan
Test together with the cloud team. Test without `#define ACLK_NEWARCH_DEVMODE`.
At the time of my last testing, not every cloud environment behaved correctly (testing/staging/prod). You might have to check with the cloud on which one is the latest code.

### DIFF
--- a/aclk/aclk.c
+++ b/aclk/aclk.c
@@ -30,8 +30,6 @@ int aclk_alert_reloaded = 1; //1 on startup, and again on health_reload
 
 time_t aclk_block_until = 0;
 
-aclk_env_t *aclk_env = NULL;
-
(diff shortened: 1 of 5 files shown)


thiagoftsm (dismissed):
During the tests on my host the data was sent normally to cloud. LGTM!

underhood:
rebased on latest master to resolve merge conflicts... retesting the code in 5'

stelfrag (approved):

MrZammler (approved):

PR #11479, opened 2021-09-02
Allow arbitrary options to be passed to make from netdata-installer.sh.

Ferroin:
##### Summary

This adds support to the installer for using the `MAKEOPTS` environment variable to specify arbitrary options to be passed to `make` when building Netdata.

The primary use of this is the `-j` and `-l` options, which will allow users a greater degree of control over resource utilization during the install process.

By default, if there is no `-j` option, we will still add a `-j` option with a value equal to the number of logical CPU cores on the system (usually determined by running `nproc`).

These settings will be persisted by default so they get used during system updates as well.

##### Component Name

area/packaging

##### Test Plan

This can be tested easily by running `./netdata-installer.sh` with arbitrary values for the `MAKEOPTS` environment variable. The interesting case for testing is `MAKEOPTS='-j1'`, which should serialize the build process for each component (and make the build take significantly longer on large systems).

##### Additional Information

Created in response to https://community.netdata.cloud/t/installation-crashes-the-entire-machine-ubuntu-raspberry-pi-4/1665 on the forum.

The generic `MAKEOPTS` approach was chosen here for three reasons:

* It’s consistent with how most source-based distros handle the same functionality, and thus statistically more likely to be intuitive to users.
* It’s exponentially simpler than adding options for this functionality.
* Other options may be of use to some users (such as the `--silent` option for gmake) or developers (such as possibly the `--trace` option).


netdata-installer.sh
```diff
@@ -118,6 +118,13 @@ download_go() {
 # make sure we save all commands we run
 run_logfile="netdata-installer.log"
 
+# set default make options
+if [ -z "${MAKEOPTS}" ]; then
+  MAKEOPTS="-j$(find_processors)"
+elif echo "${MAKEOPTS}" | grep -vqF "-j"; then
```
MrZammler, line 124: Should we add `--` before the `"-j"` ? In a test I get: `grep: invalid option -- 'j'` when I try e.g. `MAKEOPTS="-j3" ./netdata-installer.sh ....`

Ferroin: Ah, yeah, we do need that...

Ferroin: Though for `grep`, it’s `-e` for the pattern. `--` is used to mark the start of a file list to search through.

Ferroin:
> Looks good! Small detail, but do you think we should check if `-j` is followed by a number? Don't know if there is such a case, i.e. if there is a valid case of someone passing a parameter that starts with `-j`

In theory we probably should, but the check as it is now is easier to quickly understand, and the user specifying `-j` without a number is unlikely to ever happen except by accident (and I’m pretty sure `make` will complain about a missing argument in that case).

ilyam8 (approved):

ilyam8:
LGTM, did a quick test using `MAKEOPTS="-j1" ./netdata-installer.sh --install /opt` and it worked great 👍

PR #11490, opened 2021-09-07
Added static builds for ARMv7l and ARMv8a

Ferroin:
##### Summary

This adds static builds for ARMv7 and 64-bit ARMv8 platforms. These will not be used directly yet (though adventurous users can manually download and run the installer archives), but will enable usage of static builds on these platforms when we release the new kickstarter script.

##### Component Name

area/ci
area/packaging

##### Test Plan

Verified in CI and through some manual testing.

##### Additional Information

We may be adding ARMv6 (for RPi 1 hardware) as well.


Ferroin:
OpenSUSE 15.2 build failure is due to upstream issues with OpenSUSE.

The shellcheck error is bogus (the line it’s alerting on is incorrect, and the one it logically should alert on is a case where we _want_ word splitting).

vkalintiris (approved):
LGTM. However, I'm curious what is the build time improvement with all the extra options that we'll have to maintain.

Ferroin:
> LGTM. However, I'm curious what is the build time improvement with all the extra options that we'll have to maintain.

Prior to this, it’s been normal for the single Docker build check to take more than an hour. With this, the whole set complete in about 35 minutes.

Ferroin:
Disregard my previous (now deleted) comment, it was on the wrong PR.

Ferroin:
Rebased to pick up some CI improvements that will make testing this easier (check the ‘Artifacts’ dropdown on the Static Build CI check details pages).

